Issue #822, opened 2014-03-19
Identify bottleneck in lighting code on ARMs ($5)


madmaxoft:
The lighting code runs very slow on RasPi and possibly other ARMs, running at half the chunk-speed of the chunk generator. Find out why - use a profiler, a cache grinder or whatever tools are available to find out why the code is so slow.

bearbin:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=1544639)](https://www.bountysource.com/issues/1544639-identify-bottleneck-in-lighting-code-on-arms?utm_source=1544639&utm_medium=shield&utm_campaign=ISSUE_BADGE)

worktycho:
To anyone looking to claim the bounty: The main issue is that the lighting code has to copy a large amount of memory (500kB) before starting. Reducing the time taken by this copy would massively reduce the total time taken.

PR #883, opened 2014-04-11
Lighting reads blocktypes only for blocks under heightmap.

madmaxoft:
This is a preliminary "fix" for #822, but it needs verification first.

It should theoretically speed lighting up, since less data is copied back and forth.

Also implemented a possibly more cache-friendly blocklight starter algorithm (`PrepareBlockLight2()`), is disabled by default, needs perf testing.

I don't have the tools needed for cache performance testing. @worktycho could you run a comparison between this and the master, and then again between this and when you change `src/LightingThread.cpp:240` to `PrepareBlockLight2();`? Each should provide a performance boost, but I'm not sure how big and if the other effects cancel it out.

### DIFF
--- a/src/ChunkDef.cpp
+++ b/src/ChunkDef.cpp
@@ -1,0 +1,9 @@
+
+#include "Globals.h"
+
+#include "ChunkDef.h"
+
+// It appears that failing to have this definition causes link errors as cChunkDef::Height is not
+// defined. It also appears that we can have the initalizer in the declaration so it can be inlined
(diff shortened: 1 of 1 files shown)


tommysanterre:
I'm looking for the benchmark you create @worktycho but I can't find the lightingbenchmark branch and it doesn't seem to have been merged.

worktycho:
It probably got deleted. What do you want it for?

tommysanterre:
I want to run it and work on the issue.

PR #1136, opened 2014-06-28
Send statistics to the player on login.

Howaner:


tigerw:
@Howaner do you remember why this was needed?

PR #1596, opened 2014-11-15
Rewrote HandleRightClick()

Howaner:


tigerw:
@Howaner, can the branch be deleted?

NiLSPACE:
Btw, I don't think the packet is removed in 1.8, because WorldEdit's compass tool is working just fine. It's just that the animation variable was changed from `1` to `0`

PR #1621, opened 2014-11-29
Minecart Derail resolved (Resolves #1588)

### LINKED ISSUE #1588: Powered Rails give too much power
So.. unlike the original Vanilla server, in MCServer, the powered rails, when powered with Redstone Torch, don't maintain a constant speed(acceletaring every block of powered rail) and the Cart goes so fast and derails.

sirturtletheknight:
Let's add this to main shall we?


SphinxC0re:
Any update on this one?

PR #1630, opened 2014-11-30
Block fixes

Howaner:
Do not merge.
We will test some more blocks.


tigerw:
@Howaner?

tigerw:
Though there are some potentially good bugfixes here.

worktycho:
Yes, but the branch would need a lot of work to be mergable, and no one is currently working on it.

Issue #1482, opened 2014-09-29
Ghost Items


SphinxC0re:
Ghost Items are created when crafting, with shift - click, something, in the ninth slot of the hotbar.
![2014-09-29_19 44 48](https://cloud.githubusercontent.com/assets/3702016/4446222/8177e08e-4800-11e4-8f8e-2250b39919f6.png)
![2014-09-29_19 44 51](https://cloud.githubusercontent.com/assets/3702016/4446221/8177d076-4800-11e4-995e-60e60ea04d43.png)

SphinxC0re:
Fixed. Finally :dancer:

PR #1665, opened 2014-12-13
Own classes for all windows and implemented vanilla-like shift click.

### LINKED ISSUE #1669: Inventory crafting box, shift click creates a double stack
So far I have only tested this with spruce wood -> spruce planks
when I shift click the planks to create as many as I can, let say I have enough for 40 planks, it will create 40 in my inventory and 40 in one of my hand slots, then if i try to move the inventory one it will dissapear.
The console error is
Window lost sync at item 8 in SlotArea 9 with items
My item: -1:0:0 * 0
Their item: spruceplanks:1 * 40

Howaner:
Fixes #1669 and #1482


madmaxoft:
@Howaner any chance of getting this polished, so that it can be merged?

src/UI/Window.h
```diff
@@ -154,14 +153,17 @@ class cWindow
 	
 	/** Called on shift-clicking to distribute the stack into other areas; Modifies a_ItemStack as it is distributed!
 	if a_ShouldApply is true, the changes are written into the slots;
-	if a_ShouldApply is false, only a_ItemStack is modified to reflect the number of fits (for fit-testing purposes)
-	*/
-	void DistributeStack(cItem & a_ItemStack, cPlayer & a_Player, cSlotArea * a_ExcludeArea, bool a_ShouldApply);
+	if a_ShouldApply is false, only a_ItemStack is modified to reflect the number of fits (for fit-testing purposes) */
+	virtual void DistributeStack(cItem & a_ItemStack, int a_Slot, cPlayer & a_Player, cSlotArea * a_ClickedArea, bool a_ShouldApply) = 0;
+
+	/** Called from DistributeStack() to distribute the stack into a_AreasInOrder; Modifies a_ItemStack as it is distributed!
```
Howaner: You use it in the crafting slot area. I don't know why a_ShouldApply is needed ..
https://github.com/mc-server/MCServer/blob/Inventory/src/UI/SlotArea.cpp#L662

madmaxoft: It is used so that any code can test how much of the stack will be left after distributing (like in the example you linked). The crafting code uses this to prevent shift-click crafting of too many items.

By using this code we actually fixed a slight bug in vanilla - for example if you shift-click-craft torches, but there's space only for two of them in your inventory, vanilla will still craft them and you will lose those two extra torches that won't fit. We won't craft, so you don't lose items.

My question is whether this behavior has been changed in your changes (the comment suggests that `a_ShouldApply` is more or less ignored).

Howaner:
Is this merge-able?

madmaxoft:
Is it really necessary to put the method implementation into the header files? That is really bad practice - whenever you need to change anything in the implementation, half the project will need to recompile because the change is in the header that is included into other files. It should only be used for three things:
- templates
- simple getters and setters
- (maybe) forwarders

Howaner:
So I should create cpp classes for every window?

madmaxoft:
Yes please.

SphinxC0re:
Status?

Howaner:
Finished.

Issue #1411, opened 2014-09-10
Add sponge interaction


sphinxc0re:
http://minecraft.gamepedia.com/Sponge

The sponge should remove water in its vicinity.

ard1998:
sponges still won't get wet

earboxer:
Sponges have a limit to the number of water blocks they can absorb (65), and the distance that they can absorb out from (taxicab distance of 7 in all directions).

I'm thinking to do this properly:

1. Generate a list of connected waterblocks with taxicab distance 7 (recursive function, also note the distance, If you hit a block twice, store the smaller distance)
2. do a breadth first traversal, removing water until 65 blocks have been removed,
    or the list is empty. (Go through the list 7 times, each time removing the items that are closest)

peterbell10:
@earboxer you might want to look into [flood fill](https://en.wikipedia.org/wiki/Flood_fill). I think this is only a slight variant of that algorithm.

A seed should track the remaining distance, starting at 7 and decresing by one each time. Then for each seed you check the 6 nearest neighbours for water and if so mark them as a seed. By using a queue to store seeds, you would get breadth first search by default.

PR #1678, opened 2014-12-17
Added sponge.

Howaner:
Implements #1411


Howaner:
Is this merge-able?

madmaxoft:
This is somewhat unfortunate - the sponge as implemented here doesn't make the water-absorption count as the player-placement. This means that plugins such as grief prevention couldn't prevent protected water being absorbed by unprotected sponges. Is there any way this could be worked around?

tigerw:
Would you suggest implementing a new hook for sponges or reusing the current player-break-block hook? The first one is more maintainable.

madmaxoft:
It should reuse the player-place infrastructure, placing air blocks. This will ensure that the same mechanism is used everywhere; if we decide to change the `OnPlayerPlacingBlock` hook (which I'm considering, changing it to report all the blocks at once), it will get changed automatically, too.

tigerw:
If only the hook could be renamed without regard for thinking about plugins.

SphinxC0re:
Any update on this one?

Howaner:
So, what should I do.
Should I create a new BLOCK_SPONGE_HOOK hook?

madmaxoft:
Instead of changing the blocks to air in the `cWorld`, it should call the `cPlayer::PlaceBlocks()` instead. That's all.

SphinxC0re:
Does this implement the *new* or the *old* sponge mechanics?
[This](https://www.youtube.com/watch?v=VfOuiPOIJA0) will only work with the new mechanics

SphinxC0re:
What's up with this PR??

SphinxC0re:
Closing due to no response

PR #1697, opened 2015-01-02
Updated copyright/license date to 2015 for new year.

benhc123:
Updated copyright/license date to 2015 for new year.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
 MCServer: A performant C++ Minecraft Server
 www:      http://mc-server.org/
  
-Copyright 2014 MCServer Team
+Copyright 2015 MCServer Team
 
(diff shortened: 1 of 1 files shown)


bearbin:
I think this should really be changed so it has copyright xxxx to xxxx rather than just 2015, as most of the code hasn't actually been modified in 2015

madmaxoft:
And do we need the date at all? It's something that will only need changing, can't we remove the year altogether?

bearbin:
I'm petty sure that it's better to have no year than an incorrect one, so
removing it would be fine with me.

tigerw:
@benhc123, want to update the PR?

bearbin:
Fixed in master, x to x.

PR #1698, opened 2015-01-04
More user friendly log file names

planetx:
Changed the format of the log file names to LOG-YYYY-MM-DD.txt for easier use by humans and scripts.


madmaxoft:
Not good - all log files from a single day are joined into one file.

bearbin:
BTW, you need to link your github account to your git email address.

planetx:
You only generate a single log file anyways.  This would make log file management easier.

madmaxoft:
No, a new file is generated for each server run, and that's intentional.

PR #1701, opened 2015-01-09
LibEvent-based socket API

### LINKED ISSUE #1352: Implement a high-level "server" and "client" class abstractions.
These should provide an  easy interface for making a TCP server and client. The interface should be based on callbacks that the class calls for various events - new connection, available incoming data, connection closed etc.
The classes should use whatever underlying OS-dependent API they need to provide reasonable responses with little to no limits (such as the `select`'s fd limit on *nix).

Then, we can export this interface to Lua to provide the much needed async networking to the plugins.

madmaxoft:
DO NOT MERGE YET!

This is a work-in-progress to implement a LibEvent-based socket API that both the C++ part and Lua plugins could use. The PR has been created as a platform for (early) code review; for discussion please use the forum thread:
http://forum.mc-server.org/showthread.php?tid=1700

Will fix #1352


madmaxoft:
The build fails because of `ASSERT(!"Not implemented yet")`; but otherwise should be alright.

worktycho:
I don't think running tests during global constructors is the right idea. I'll see if I can convert them to external tests.

madmaxoft:
It's a quick and easy way to test stuff; the tests will either be commented out or under an #ifdef in the final version. If you pull it out into a separate test, that should be fine, too.

worktycho:
I'm reverting e66c71a as this was an intentional design design decision. The problem with logs in global constructors is the order of global constructors is undefined so there is the potential for calling methods on an uninitiated  object.

madmaxoft:
That's why there's the pseudo-singleton, the order is well defined then.

worktycho:
There's also the fact that you shouldn't be doing real work in global constructors anyway because of the issue. And its not needed now I've extracted the test.

madmaxoft:
I'll rebase the whole thing, dropping the unused commits, okay?

worktycho:
Fine.

worktycho:
Network.cpp seems a little large. Shall I refactor TCPLink into its own file?

madmaxoft:
I think it's alright for now. Let's wait until the server part is implemented, then we'll see if a split is useful.
The cpp file is hiding most of the underlying implementation; if you wanted to split it, you'd need to expose a lot in a header file.

madmaxoft:
Weird, MSVC is happy with a class being forward-declared in the `friend class` declaration; gcc and clang need an explicit forward declaration.

worktycho:
That's an MSVC bug I believe.

madmaxoft:
We have a problem, LibEvent doesn't support being built out-of-source while being a direct dependency - it generates some of its config headers into the build dir and so such headers are then not found by the dependent project. I'll try to fix this.

madmaxoft:
I think this is now ready for a major review - is this API sufficient, does it perform well, and is it documented enough?

src/OSSupport/Network.h
```diff
@@ -0,0 +1,215 @@
...
+
+
+#pragma once
+
+
+
+
+
+/** Interface that provides the methods available on a single TCP connection. */
+class cTCPLink
```
worktycho: This class should probably be in a separate file, as should cServerHandle.

worktycho: Also is cTCPLick intended to be used as a value type as it only contains a pointer?

src/OSSupport/Network.h
```diff
@@ -0,0 +1,215 @@
...
+	virtual ~cServerHandle() {}
+
+	/** Stops the server, no more incoming connections will be accepted.
+	All current connections will be shut down (cTCPLink::Shutdown()). */
+	virtual void Close(void) = 0;
+
+	/** Returns true if the server has been started correctly and is currently listening for incoming connections. */
+	virtual bool IsListening(void) const = 0;
+};
+typedef SharedPtr<cServerHandle> cServerHandlePtr;
```
worktycho: Is it really necessary for this to be a shared_ptr rather than a unique_ptr?

src/OSSupport/Network.h
```diff
@@ -0,0 +1,215 @@
...
+		virtual void OnSuccess(cTCPLink & a_Link) = 0;
+
+		/** Called when the Connect call fails. */
+		virtual void OnError(int a_ErrorCode) = 0;
+	};
+	typedef SharedPtr<cConnectCallbacks> cConnectCallbacksPtr;
+
+
+	/** Callbacks used when listening for incoming connections as a server. */
+	class cListenCallbacks
```
worktycho: Does it really make sense to use a special functor type here and not just use lambdas?

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+public:
+	/** Creates a new link to be queued to connect to a specified host:port.
+	Used for outgoing connections created using cNetwork::Connect().
+	Call Connect() first, before using the link. */
+	cTCPLinkImpl(const cCallbacksPtr a_LinkCallbacks);
+
+	/** Creates a new link based on the given socket.
+	Used for connections accepted in a server using cNetwork::Listen().
+	a_Address and a_AddrLen describe the remote peer that has connected. */
+	cTCPLinkImpl(evutil_socket_t a_Socket, cCallbacksPtr a_LinkCallbacks, cServerHandleImpl * a_Server, const sockaddr * a_Address, int a_AddrLen);
```
worktycho: It would probably be better if this constructor was private and cNetwork::Connect was moved to this class.

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+
+	/** Destroys the LibEvent handle representing the link. */
+	~cTCPLinkImpl();
+
+	/** Queues a connection request to the specified host.
+	a_ConnectCallbacks must be valid.
+	The object must have been constructed by the right constructor (without the Socket param). */
+	bool Connect(const AString & a_Host, UInt16 a_Port, cNetwork::cConnectCallbacksPtr a_ConnectCallbacks);
```
worktycho: This method just seems unnecessary, what is the intended use case for it?

worktycho: Why is this a separate method from the constructor rather than moving the logic into factory method and moving the factory method into the class?

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+		cNetwork::cListenCallbacksPtr a_ListenCallbacks,
+		cTCPLink::cCallbacksPtr a_LinkCallbacks
+	);
+
+	/** Closes the server, dropping all the connections. */
+	~cServerHandleImpl();
+
+	/** Starts listening on the specified port.
+	Both IPv4 and IPv6 interfaces are used, if possible. */
+	bool Listen(UInt16 a_Port);
```
worktycho: Again why is this a separate method rather than moving the logic into the factory method and moving the factory method into this class?

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+
+	// cServerHandle overrides:
+	virtual void Close(void) override;
+	virtual bool IsListening(void) const override { return m_IsListening; }
```
worktycho: Under what situation would you expect to have a not-listening server handle?

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+
+	/** Container for all currently active connections on this server. */
+	cTCPLinkImplPtrs m_Connections;
+
+	/** Mutex protecting m_Connections againt multithreaded access. */
+	cCriticalSection m_CS;
+
+
+	/** The callback called by LibEvent upon incoming connection. */
+	static void Callback(evconnlistener * a_Listener, evutil_socket_t a_Socket, sockaddr * a_Addr, int a_Len, void * a_Self);
```
worktycho: Are you sure that this is the right calling convention for passing to libevent?

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+				evutil_inet_ntop(AF_INET, &(sin->sin_addr), IP, sizeof(IP));
+				break;
+			}
+			case AF_INET6:  // IPv6
+			{
+				sockaddr_in6 * sin = reinterpret_cast<sockaddr_in6 *>(a_Addr->ai_addr);
+				evutil_inet_ntop(AF_INET6, &(sin->sin6_addr), IP, sizeof(IP));
+				break;
+			}
+			default:
```
worktycho: Might be worth putting in either an assert of a LOGD in here.

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+			break;
+		}
+	}  // switch (address family)
+}
+
+
+
+
+
+void cIPLookup::Callback(int a_Result, char a_Type, int a_Count, int a_Ttl, void * a_Addresses, void * a_Self)
```
worktycho: Again is this the right calling convention?

src/OSSupport/Network.cpp
madmaxoft: When the `listen` syscall fails (most likely another socket already listening on that port), we still need to return a valid `cServerHandle` instance (so that the Lua API can read the error code and message from it).

src/OSSupport/Network.cpp
madmaxoft: If it werent' the compiler would complain.

src/OSSupport/Network.cpp
madmaxoft: Not really. If a hostname resolves to IPv4, IPv6 and additionally, say, and IPX/SPX address, we want to ignore the IPX/SPX address silently.

src/OSSupport/Network.h
madmaxoft: Why put an interface definition that has some 30 lines into a separate file, just for the sake of it?

`cCallbacksPtr` is shared between multiple links that are created by the `cServerHandle` for incoming connections.

src/OSSupport/Network.h
madmaxoft: The instance is returned by the `cNetwork::Listen()` function and will most likely in later versions be also stored within the `cNetworkSingleton` class.

src/OSSupport/Network.h
madmaxoft: We use functors for everything else, and this allows us to later add more callbacks - such as an error one, or a pre-accept address check.

src/OSSupport/Network.cpp
madmaxoft: That's a good idea.

src/OSSupport/Network.h
worktycho: Its more about splitting up the cpp file than the h. If you split off the interface, you can split the implementation off in to the same cpp as the interface.

src/OSSupport/Network.cpp
```diff
@@ -0,0 +1,1259 @@
...
+
+
+
+
+/** Implements the cTCPLink details so that it can represent the single connection between two endpoints. */
+class cTCPLinkImpl:
+	public cTCPLink
+{
+	typedef cTCPLink super;
+	friend class cServerHandleImpl;
```
worktycho: What private data does cServerHandle need access to?

madmaxoft: None, this seems to be a historical remnant.

madmaxoft:
Shall we merge now and open a new PR for the actual rewrite of the server internals (so that the Lua API can be developed in parallel in a separate PR)?

STRWarrior:
I agree for a merge.

madmaxoft:
After reading the `cServer` code for a while, I think the API is not good yet. (more in the forum)

PR #1702, opened 2015-01-11
Started to convert more of MCServer to std::chrono

worktycho:


madmaxoft:
Just out of curiosity, what is `sizeof(std::chrono::milliseconds)`? Is it an object small enough to pass by value?

worktycho:
On any sensible implementation of the standard library it should just be a single 64 bit integer field. I have checked this is true for the libstdc++4.8. It was designed to be used in exactly this way.

madmaxoft:
Sounds good enough.

tigerw:
Presumably you've agreed not to use the user defined literals?

worktycho:
I wanted to maximise the chance of the PR getting accepted. I might add them in a future PR.

src/World.h
```diff
@@ -1051,13 +1049,13 @@ class cWorld :
 	cWorld(const AString & a_WorldName, eDimension a_Dimension = dimOverworld, const AString & a_OverworldName = "");
 	virtual ~cWorld();
 
-	void Tick(float a_Dt, int a_LastTickDurationMSec);
+	void Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_LastTickDurationMSec);
```
madmaxoft: Now the real question: What's the difference between `a_Dt` and `a_LastTickDurationMSec`?

worktycho: a_Dt is how long between the start of the last tick and this on, a_LastTickDurationMSec is how long the last tick took to execute. I'm not sure anything uses a_LastTickDurationMSec.

worktycho:
Would someone mind reviewing this code?

src/World.cpp
```diff
@@ -869,30 +865,27 @@ void cWorld::Tick(float a_Dt, int a_LastTickDurationMSec)
...
 
-	m_WorldAgeSecs  += (double)a_Dt / 1000.0;
-	m_WorldAge  = (Int64)(m_WorldAgeSecs  * 20.0);
+	m_WorldAge += a_Dt;
 
 	if (m_IsDaylightCycleEnabled)
 	{
 		// We need sub-tick precision here, that's why we store the time in seconds and calculate ticks off of it
-		m_TimeOfDaySecs += (double)a_Dt / 1000.0;
+		m_TimeOfDay += a_Dt;
```
madmaxoft: The comment above seems somewhat misleading now. Isn't `m_TimeOfDay` in msec?

madmaxoft:
Apart from the comment issue this seems alright.

worktycho:
I could reduce some of the chattyness by using user defined literals though I'm not sure what support for them is like.

PR #1707, opened 2015-01-17
Fixed some problems found by CppCheck

UltraCoderRU:
1. Function parameter should be passed by reference.
2. Possible inefficient checking for emptiness.
3. Prefer prefix ++/-- operators for non-primitive types.


madmaxoft:
Good changes, thanks.

PR #1713, opened 2015-01-24
Use libevent

madmaxoft:
DO NOT MERGE YET, TESTING IN PROGRESS.

Migrates the internals from `cSocketThreads` and `cListenThread` to the new LibEvent-based `cNetwork` API.


worktycho:
Just to say that I'm currently working on cServer.

src/RCONServer.cpp
```diff
@@ -39,6 +39,43 @@ enum
...
+	// cNetwork::cListenCallbacks overrides:
+	virtual cTCPLink::cCallbacksPtr OnIncomingConnection(const AString & a_RemoteIPAddress, UInt16 a_RemotePort) override
+	{
+		LOG("RCON Client \"%s\" connected!", a_RemoteIPAddress.c_str());
+		return std::make_shared<cRCONServer::cConnection>(m_RCONServer, a_RemoteIPAddress);
+	}
+	virtual void OnAccepted(cTCPLink & a_Link) override {}
+	virtual void OnError(int a_ErrorCode, const AString & a_ErrorMsg) override
+	{
+		LOGWARNING("RCON server error on port %d: %d (%s)", m_Port, a_ErrorCode, a_ErrorMsg);
```
worktycho: You forgot to convert a_ErrorMsg to a c string.

madmaxoft:
Hmm, I already have like 90 % of cServer converted.

src/RCONServer.cpp
madmaxoft: As usual.

worktycho:
How about we both post what we've done so far then see what we can combine. I've also done some work on improving the cServer API, so that might be useful anyway.

madmaxoft:
Okay, I'll commit to this branch, you make a new branch with your changes.

I'll commit only when I have things working, which might take a while longer, though.

worktycho:
I've committed what I've got so far to libEventServer. It doesn't compile at the moment because the server callback functions aren't defined yet.

madmaxoft:
Turning the `cServer` instance into a unique ptr seems like a good idea, but rewriting so much code just to make a factory seems redundant. Why not just make the factory function use the old functions without changes?

worktycho:
It was more that I felt I was going to a major refactor of cServer so I might as well move the code.

madmaxoft:
Any change that breaks apart the logical grouping of code into functions is a bad change. Your factory function was too long and had too many responsibilities; the INI-file reading should have stayed a separate function (possibly `protected` so that only the factory can call it).

madmaxoft:
So many "side quests" along the way. Just found out that logging the client communication fails when the client is using an IPv6 address (default in new `cNetwork` API) because the logfile fails to open (contains invalid chars).

worktycho:
I know. Its a horrible commit that needed spliting up. Its just what I had it a working directory in the middle of development.

madmaxoft:
This is it - now we've gotten rid of the old networking code and we're depending on LibEvent. This branch now needs lots and lots of testing before merging.

src/RCONServer.cpp
```diff
@@ -112,24 +149,28 @@ void cRCONServer::Initialize(cIniFile & a_IniFile)
...
-	if (m_ListenThread6.Initialize(Ports6))
+	// Read the listening ports for RCON from config:
+	AStringVector Ports = ReadUpgradeIniPorts(a_IniFile, "RCON", "Ports", "PortsIPv4", "PortsIPv6", "25575");
+
+	// Start listening on each specified port:
+	for (auto port: Ports)
 	{
-		HasAnyPorts = true;
-		m_ListenThread6.Start();
+		UInt16 PortNum = static_cast<UInt16>(atol(port.c_str()));
```
worktycho: any reason to use `atol` rather than `atoi` for a 16 bit number?

worktycho: Also this seems like a good case to use `StringToInteger<UInt16>` which checks for bounds as well.

src/RCONServer.cpp
```diff
@@ -184,49 +218,45 @@ bool cRCONServer::cConnection::DataReceived(const char * a_Data, size_t a_Size)
...
-			m_RCONServer.m_SocketThreads.RemoveClient(this);
-			m_Socket.CloseSocket();
-			delete this;
-			return false;
+			m_Link->Close();
+			m_Link.reset();
+			return;
 		}
-		if (Length > (int)(m_Buffer.size() + 4))
+		if (Length > static_cast<int>(m_Buffer.size() + 4))
```
worktycho: If you're going to correct the cast, you might want to correct the type as well. Length should really be an Int32.

src/ClientHandle.cpp
```diff
@@ -689,6 +693,47 @@ void cClientHandle::HandleCreativeInventory(short a_SlotNum, const cItem & a_Hel
...
+		(m_Player->GetWindow()->GetWindowID() != a_WindowID) ||
+		(m_Player->GetWindow()->GetWindowType() != cWindow::wtEnchantment)
+	)
+	{
+		return;
+	}
+	
+	cEnchantingWindow * Window = (cEnchantingWindow*) m_Player->GetWindow();
+	cItem Item = *Window->m_SlotArea->GetSlot(0, *m_Player);
+	int BaseEnchantmentLevel = Window->GetPropertyValue(a_Enchantment);
```
worktycho: This should be a short

src/OSSupport/File.cpp
```diff
@@ -453,6 +453,29 @@ AString cFile::ReadWholeFile(const AString & a_FileName)
...
 
+AString cFile::ReplaceFileNameInvalidChars(const AString & a_FileName, char a_Replacement)
+{
+	AString res(a_FileName);
+	for (auto & ch: res)
+	{
+		switch (ch)
+		{
+			case ':':
+			case '\\':
```
worktycho: It is possible to create a file on linux that contains `\` or `:`. I think removing invalid chars is the wrong approach.

worktycho: If you want to go for the minimum cross platform set you also need to exclude everything other than A-Z. 
For example their would be an issue with IPv4 addresses on FAT32, mounted on a system that doesn't long filenames.

madmaxoft: If you have a better solution, I'm all ears.

worktycho: I think we should either use an ifdef, or just substitute the colons out rather then trying to build a generic solution.

madmaxoft:
There's something seriously wrong in the `cNetwork` code. It keeps crashing the server or disconnecting the client. I usually get disconnected within a minute, sometimes even before actually spawning in the game.

src/RCONServer.h
```diff
@@ -35,72 +33,61 @@ class cRCONServer :
...
+		virtual void OnError(int a_ErrorCode, const AString & a_ErrorMsg) override;
 		
-		/// Processes the given packet and sends the response; returns true if successful, false if the connection is to be dropped
-		bool ProcessPacket(int a_RequestID, int a_PacketType, int a_PayloadLength, const char * a_Payload);
+		/** Processes the given packet and sends the response; returns true if successful, false if the connection is to be dropped */
+		bool ProcessPacket(UInt32 a_RequestID, UInt32 a_PacketType, UInt32 a_PayloadLength, const char * a_Payload);
 		
-		/// Reads 4 bytes from a_Buffer and returns the int they represent
-		int IntFromBuffer(const char * a_Buffer);
+		/** Reads 4 bytes from a_Buffer and returns the LE UInt32 they represent */
```
worktycho: This comment is ambiguous. It can mean that the value read from the buffer is a LE value or that the value is returned in LE order.

worktycho:
I've fixed the travis build error.

src/RCONServer.h
madmaxoft: I don't think so. The value is represented LE in the buffer; storing the value in a specific endian in the `UInt32` is something no-one would really do.

But if you have a better wording, go ahead and change it by all means. I'm not a native English speaker, after all.

madmaxoft:
Okay, now we need to find the crashing problem. Can't merge before that.

madmaxoft:
Strangely enough, when I turn authentication off, those crashes almost disappear. They show up from time to time, but really scarce.

madmaxoft:
The `BEV_OPT_THREADSAFE` flag seems to have helped, I am now unable to reproduce the crash. Only when stopping the server I get a crash inside the CRT, no idea where that comes from.

madmaxoft:
And this might be the cause: https://connect.microsoft.com/VisualStudio/feedback/details/786016/creating-a-global-c-object-that-used-thread-join-in-its-destructor-causes-a-lockup

madmaxoft:
Yes, moving the termination code to the end of `main()` seems to have helped.
Now I need to solve a deadlock.

madmaxoft:
I think this is now ready for proper testing.

madmaxoft:
No-one reported any problems, so I merged.

SphinxC0re:
Does this make the server any faster?

worktycho:
Should reduce context switches in the networking, but the main purpose was to build an API we could expose to lua as the old one involved creating two threads per socket whilst this one has a single global network marshaling thread. Also the new API deals with the problem that some platforms have limits in how many sockets could be passed to select at once.

PR #1716, opened 2015-01-25
filter out leaves meta changes from being sent to the client (resolves #1714)

### LINKED ISSUE #1714: Too many leave block changes
Currently the server sends too many leave block changes.

The server updates the meta from a leave block every tick.
It's important to change this. (Don't change the meta or don't send it to the client)

This generates many lags in 1.7 and in 1.8 with disabled VBO.

Raekye:
If I understand the issue, it's that the leave meta changes (e.g. decay) don't need to be sent to the client.
So in `cChunk::FastSetBlock(...)`, if the block is a leaf and hasn't changed, then the meta must have changed, so we skip it

I wasn't entirely sure how to run the tests, so I just ran `CIbuild.sh`

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1570,6 +1570,10 @@ void cChunk::FastSetBlock(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_BlockT
 	if (
 		a_SendToClients &&                  // ... we are told to do so AND ...
 		(
+			!(                                // ... the old and new blocktypes AREN'T leaves (because the client doesn't need meta updates)
+				((OldBlockType == E_BLOCK_LEAVES) && (a_BlockType == E_BLOCK_LEAVES)) ||
+				((OldBlockType == E_BLOCK_NEW_LEAVES) && (a_BlockType == E_BLOCK_NEW_LEAVES))
+			) ||                              // ... OR ...
 			(OldBlockMeta != a_BlockMeta) ||  // ... the meta value is different OR ...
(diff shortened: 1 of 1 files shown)


worktycho:
The tests don't matter since they don't test anything that would be affected by this change at the moment.

src/Chunk.cpp
```diff
@@ -1570,6 +1570,10 @@ void cChunk::FastSetBlock(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_BlockT
 	if (
 		a_SendToClients &&                  // ... we are told to do so AND ...
 		(
+			!(                                // ... the old and new blocktypes AREN'T leaves (because the client doesn't need meta updates)
+				((OldBlockType == E_BLOCK_LEAVES) && (a_BlockType == E_BLOCK_LEAVES)) ||
+				((OldBlockType == E_BLOCK_NEW_LEAVES) && (a_BlockType == E_BLOCK_NEW_LEAVES))
+			) ||                              // ... OR ...
```
worktycho: Doesn't this need to be an and sat the end? This would cause the server to alway update blocks that aren't leaves rather that not updating leaves on meta changes.

Raekye: Ah true... stupid me

madmaxoft:
Thanks.

madmaxoft:
Incidentally, the condition is getting out of hands. We might want to split it into a separate function, "ShouldSendBlockChange()". That would make it easier to maintain.

PR #1724, opened 2015-02-04
Changed Noise to not check for repeated values

worktycho:
Also changed NoiseTest to test if this is an improvement.

Do not merge! This is an experimental micro-optimisation that needs benchmarking on more systems to verify an improvement.

If any one would like to try the NoiseTest benchmark please report the results.


tigerw:
Merge? It's probably beneficial.

worktycho:
I'd like a second opinion on this. xoft was against it because the tests a couple of years back had shown it to be advantageous to cache results. I suspect improved compilers have changed the situation, but I don't want to merge this on my own.

bearbin:
Your email isn't associated with the one used to push this commit.

tigerw:
@madmaxoft?

madmaxoft:
@worktycho Your new noise is much faster because it is incorrect - it doesn't calculate the same values as the old one. I added an explicit comparison between the noises, step through the `CompareNoises()` function to debug.
I'm suspicious of this line in the new noise: `for (int y = 0; y < a_EndY - a_StartY; y++)`
You're comparing integral `y` with floating-point `a_EndY - a_StartY`, which doesn't make much sense. I'd expect the comparison to `a_SizeY` instead.

worktycho:
Yes that probably does make a lot more sense. I'll update this to try and get it working.

tigerw:
Query of status.

worktycho:
Close if you want. If probably not going to be able to work on this until September.

tigerw:
Closing for now, then.

PR #1729, opened 2015-02-07
Set reuse flag to sockets

### LINKED ISSUE #1726: HTTP cannot listen on port after server restart
@XsFrozdY initially reported:
Also sometimes when restarting the server I get: `HTTP server error on port 8080:10048 <cannot bind IPv6 socket to port 8080  <normally only one user allowed for each socket adress<protocol/networkadress/port>` or something like that, I translated some of it from swedish

It seems that either the HTTP server doesn't have the REUSE flag set, or there is some other difference from the main server, which didn't get reported.

Howaner:
Should fix #1726


madmaxoft:
I think you should use the LibEvent function that exists exactly for this purpose:
`int evutil_make_listen_socket_reuseable(evutil_socket_t sock);`
Note that this operation is needed only for Linux; Windows closes the socket immediately and the `SO_REUSEADDR` option means something completely different (allow multiple **active** listening sockets to coexist on the same port) Read through http://www.wangafu.net/~nickm/libevent-book/Ref5_evutil.html for more details.

PR #1732, opened 2015-02-08
Changed style of landing page a bit ;)

SphinxC0re:
please review this. I mostly used bootstrap because I wanted it to be a bit more responsive than our current site and use less JavaScript
http://sphinxc0re.github.io/MCServer/


bearbin:
What does the change look like? That's the most important bit really.

SphinxC0re:
Okay I made a total mess but.. is this mergable?

tigerw:
Are you absolutely sure it's necessary? Instead of using less JS there is now an entire folder for scripts and 14000 extra lines of code :P

SphinxC0re:
Yes but I can mostly remove the JavaScript and .less files. Did you take a look at it?

SphinxC0re:
Okay that didn't do anything but does size really matter at this point???

tigerw:
That's definitely better. May I suggest some further points:

* Re-add the navigation links (API Documentation, Forums, and etc.)
* Re-add the build status and tip status SVGs
* (Would highly appreciate) if you could make the font first Segoe UI Light for those on Windows then the present font as a fallback?
* There are two horizontal rules between the end of the Title/Subtitle and the start of the downloads table. Perhaps you could add a little bit of padding between them to reduce the impression of missing content
* You might also want to consider making the max-width of the table less, since personally it appears unnecessarily wide and may impact searchability
* 'Other installation Methods' to 'Other installation methods'?

:)

tigerw:
Not sure if `<meta http-equiv="X-UA-Compatible" content="IE=edge">`; IE doesn't seem to cause any problems and apparently it isn't standards compliant.

----


    
    
Surely we don't need to support older browsers :P

LO1ZB:
Is the MacOS 64bit link broken or did he remove the folder?
And what about an link to the build server?

SphinxC0re:
Yeah tw already told me about a lot of these things. I am not responsable for the MacOS link but it should be no problem to fix this. I am working on the most of this. The font thingy is nto doable though.

tigerw:
I guess that's alright. Improvements can be added in bits after merge.

SphinxC0re:
I'm not quite good with the position of the status SVGs. Any other suggestions?

tigerw:
Displace status banners to new line?

SphinxC0re:
Merge?

madmaxoft:
http://forum.mc-server.org/showthread.php?tid=1767 ?

SphinxC0re:
Does it keep us from merging?

madmaxoft:
Not really, it just means that the files in this change will be forever present in this repo, even if we move the web part out. And we already are at several tens of megabytes of data when cloning the repo.

SphinxC0re:
So then just make another repo for the web part an we'll merge it there

worktycho:
Since the gh-pages branch is a separate branch we can purge the whole branch. Git only has issues purging commits if something you want to keep has that commit in its history. Since we won't want to keep any of the gh-pages branch in this repo after the split we will be able to purge it.

SphinxC0re:
Okay, so...?

madmaxoft:
I like the design, but do we really need all that JS in order to produce a **single static page**?

jan64:
Imo, headers above tables are kind of pointless, and take up space. 
Also, the tip4commit/build status would look better if placed somewhere below download links. 
https://imgur.com/gwc12oI
And, i'm not sure if we really want these gray lines.

bearbin:
I'd also prefer that the little tags for version information be removed as they clutter up the download buttons.

Instead, perhaps you could make outdated download buttons orange or something like that?

jan64:
I think just removing the MC text from each of those would help a lot.

jan64:
https://i.imgur.com/Q4VOqpL.png

bearbin:
I think that looks much better, although not having the info at all wouldn't be that much of a usability downgrade.

I agree with your suggestion for hover animations on the buttons, but the gray lines make the table much more readable for me so I'd prefer they stay.

jan64:
Okay, imo the lines should be added both below and above the table - otherwise it looks inconsistently.

bearbin:
Yeah, above and below is fine, just none at all makes the table much harder to read.

SphinxC0re:
@madmaxoft the doesn't use any javascript
@everyone I will try to use another theme for usability but btw: On a touch device there is no "hover" j/s

jan64:
Doesn't mean other users should be deprived of that. ;)

SphinxC0re:
I also removed the status badges. I don't think anyone will care :dancer:

bearbin:
I disagree - the status badges need to be on the page. I don't really mind where you put them though.

SphinxC0re:
Give me a good reason for that. Someone who is really interested in how many bitcoins we get for each commit or wants to know if the build of our project on travis failed could also go to our github page. The 'normal' user doesn't care

bearbin:
The front page isn't just for "normal users". It's other purpose is to attract developers - that's why we have the information about the GitHub repo and the buildserver.

Having those status badges on the front page of the website signals that the project is developed in the open and also shows that MCServer is a project that uses best practices - we are testing our builds and making sure that they pass.

SphinxC0re:
Mind if I put them into the big source button?

jan64:
The Source button would look better if it was smaller, and on the bottom right of the page. ;<

SphinxC0re:
We want to attract developers, so I made it huge

madmaxoft:
How about splitting the source information off from the downloads table (it's not a download anyway) - that way you can include the various badges with it.

SphinxC0re:
This might be ready for merge, or isn't it?

madmaxoft:
Now the link to source is missing. The github fork ribbon is a terrible UX, no-one will notice it.

SphinxC0re:
It is pretty common and highly recognized in many software products on Github

SphinxC0re:
So make a suggestion on what I should do and where I should place the button

bearbin:
It may be that my screen contrast is terrible, but the lines around the boxes at the bottom are almost invisible.

I think this is good for a merge after that is fixed though, all the other changes are minor enough to do later.

SphinxC0re:
Is it okay now?

bearbin:
A lot better, thanks. Everyone else, good to merge?

SphinxC0re:
I tested the border width with like five different widths but they were all rubbish. Like 5px would be the last one that I would choose for that

NiLSPACE:
It's a little too flashy for my taste, but I don't mind it.

But I thought we were going to make a new project where we will have the website like mc-server/cuberite.github.io

bearbin:
I think that is still the plan, but we can copy it over after the merge.

jan64:
From IRC:
<jan64_afk> SphinxC0re, the colors are too bland, imo. 
<jan64_afk> and, the borders around bottom tables actually looked nice when they were visible
<jan64_afk> + Ofc remove the MC text from version tags
<jan64_afk> and, remove the Downloads! / More! h1 tags
<jan64_afk> it's obvious those are downloads
<jan64_afk> and, the more text is kind of pointless, too.
<jan64_afk> it'd look better, if that was just empty space
<jan64_afk> and, you can actually place the build passing tab on the buildserver button
<jan64_afk> and, the tip4commit on the github one
<jan64_afk> + I'd add more colors to the buttons
<jan64_afk> maybe change linux to orange(or something else), windows to blue, and macos to black
<jan64_afk> just to match the os's logos
<jan64_afk> and, change the colors for each config, too.
<jan64_afk> + Change the inactive buttons color to grey
<jan64_afk> that's kind of industry standard

SphinxC0re:
Okay. I tried all these things and they look terrible and are not contributing to usability

bearbin:
I'd still say merge, because although those are problems they are easy for anyone to change later.

Issue #277, opened 2013-10-21
Mob code too slow


madmaxoft:
The new mob code (cMobCollector et al) is very slow, with debug versions of MCServer the ticks take as long as 1 sec each. This is not acceptable.

std::set and std::list are not good structures, we need more speed!

tigerw:
It's still kinda slow there, probably could do with some optimisation.

SafwatHalaby:
@SphinxC0re I prefer to assign myself to things I'm actually working on, this allows others to actually work on things I'm not working on. So not every AI thing should be auto-assigned to me.

LogicParrot:
I is possible fact that monsters in unloaded chunks are not destroyed properly amplifies this issue. #2973

PR #1733, opened 2015-02-08
Added basic activation range

tonibm19:
Let's save some CPU!

I dont really know if this can help to reduce CPU usage or not.

Activation range set to 48. Tell me if I should lower or increase it or maybe I should try to make it configurable.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -95,7 +95,8 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_CanPickUpLoot(true)
 	, m_BurnsInDaylight(false)
 	, m_RelativeWalkSpeed(1.0)
-	, m_ActivationRange(48)
+	, m_TickSearchPlayer(0)
+	, m_PlayerDistance()
 {
 	if (!a_ConfigName.empty())
 	{
@@ -255,11 +256,44 @@ bool cMonster::ReachedFinalDestination()
 
 void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
-	cPlayer * a_Closest_Player = m_World->FindClosestPlayer(GetPosition(), float(m_ActivationRange));
-	if (a_Closest_Player != nullptr)
+	if ((m_PlayerDistance < 48) && (m_PlayerDistance > 0))
 	{
 		super::Tick(a_Dt, a_Chunk);
 	}
+	if (m_TickSearchPlayer == 5)
+	{
+		// Search for players inside the mob sight
+		cPlayer * a_Closest_Player = m_World->FindClosestPlayer(GetPosition(), float(m_SightDistance));
+		// Make sure we've got a player.
+		if (a_Closest_Player != nullptr)
+		{
+			m_PlayerDistance = (a_Closest_Player->GetPosition() - GetPosition()).Length();
+			// Check if the player has the item the mob wants
+			if (m_PlayerDistance < 20)
+			{
+				cItem FollowedItem = GetFollowedItem();
+				if (FollowedItem.IsEmpty())
+				{
+					return;
+				}
+				if (a_Closest_Player->GetEquippedItem().IsEqual(FollowedItem))
+				{
+					Vector3d PlayerPos = a_Closest_Player->GetPosition();
+					MoveToPosition(PlayerPos);
+				}
+			}
+			m_TickSearchPlayer = 0;
+		}
+		else
+		{
+			m_PlayerDistance = 0;
+		}
+	}
+	else
+	{
+		m_TickSearchPlayer++;
+	}
+
 
 	if (m_Health <= 0)
 	{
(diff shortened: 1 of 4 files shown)


NiLSPACE:
Vanilla has something simular if you're wondering.

tonibm19:
Yeah, and bukkit too.

bearbin:
Initialisation order is wrong, so clang build failed.

tigerw:
I think this is currently what's happening: https://github.com/mc-server/MCServer/blob/master/src/World.cpp#L1018

madmaxoft:
I don't think this change is worth it. Instead of doing relatively simple AI, you're searching for a nearby player, which is a heavy operation itself, not to mention that the function is marked as unsafe and should be avoided.

A better option would be to store the distance in a member variable and only update it occasionally. (This coincides with #277 - we want the mob spawning code to cache the distances, too)

tonibm19:
Update 1 time per second then?
I could also move the mob following player when it has wheat or so code there.

tigerw:
Pues, do what xoft said and add a cached value to the current code :)

tonibm19:
Didn't you say its already done?

tigerw:
The present code still does what xoft says shouldn't happen I think - collecting distances every tick.

tonibm19:
What do you think?

tonibm19:
?

src/Mobs/Monster.cpp
```diff
@@ -95,6 +95,8 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_CanPickUpLoot(true)
 	, m_BurnsInDaylight(false)
 	, m_RelativeWalkSpeed(1.0)
+	, m_TickSearchPlayer(0)
+	, m_PlayerDistance()
```
madmaxoft: Move this line up, in front of `m_TickSearchPlayer`, to fix compilation.

madmaxoft:
Instead of counting up to a constant number of ticks (`m_TickSearchPlayer`), count down towards zero and update when zero. This will allow you later to change the interval dynamically, such as mobs closer to players will get updated more often than mobs further away. Rename to `m_TicksUntilPlayerSearch`.

Afterwards it'd be really nice if you rebased this entire branch onto master, squashing all the commits into one with a reasonable commit message.

tonibm19:
I really like your idea, dynamically changing search interval. I think it would be fine if search interval was = distance with a minimum value of 5.
There is one problem thought, related to pathfinding. If a mob is following you (holding wheat or so) and you dont search for the player every tick, pathfinding/AI acts and changes mob destination. In 4 ticks interval is almost unoticeable, but what if it was 16? Maybe it wouldnt be enough to make the mob follow you. Any idea about how to fix this?

madmaxoft:
Quite the contrary. Calling `MoveToPosition` sets the target for the mob, so it will walk there; the further away the mob is from a player, the further the player has to move for the targetting to be completely off. It doesn't matter if the mob is targetting something two blocks away from the current player position if it is some 40 blocks away.

I think a good function would be `m_TicksUntilPlayerSearch = 4 + static_cast<int>(m_DistanceToPlayer) / 4`: for each 4 blocks of distance it adds a single tick of delay, so a mob 4 blocks away from a player will be re-targetted 4 times per second, while a mob 64 blocks away from a player will be re-targetted once a second.

tonibm19:
It seems there are some problems with mob movement. After a while, they start stopping moving, and then moving away +6 blocks and "reviving". Also, wolves do not follow you anymore when holding bones. I'll look into it.
Any idea about what's wrong? Latest MCS without this changes works fine.

tigerw:
> What do you think?
> ?

Sorry, had no Internet access. Looks good, but put it in World.cpp (L1018), since it already looks for mob-player distances there.

----

> It seems there are some problems with mob movement.

I'm not certain, but,
```cpp
cItem FollowedItem = GetFollowedItem();
if (FollowedItem.IsEmpty())
{
	return;
}
```
would mean that for entities without GetFollowedItem, the rest of cMonster::Tick (the pathfinding code) would not be executed (one in five?) ticks, which could cause jittering. Wolves don't have GetFollowedItem set, so that may be the cause of the issues with them not following bones.

----

```cpp
if ((m_PlayerDistance < 48) && (m_PlayerDistance > 0))
{
	super::Tick(a_Dt, a_Chunk);
}
```
might cause problems since if an entity occupies the exact spot as a player, the vector length, and so m_PlayerDistance, should be zero, and so super::Tick(), which does physics handling, wouldn't run.

----

Please correct me if I've made a mistake :P

tonibm19:
In my current code (not updated here yet) whay you said about an entity occupying the exact spot as a player is fixed. But the problem is there, happening with mobs which have a followeditem set.

tigerw:
How strange. Try with a debugger?

tonibm19:
Now I'll try your with the wolf. Let's see if it works.

tonibm19:
Yay!! Wolf fixed!! First I thought there was something wrong because it was always following me but then I realized it was tamed.
Will see if I can do something about the other problem.
Also, I'll try to schedule the wolf code.

SphinxC0re:
clang build fails because of wrong initialization order

tonibm19:
I know. It should be fixed in my local version.

tonibm19:
It's been a long time... How have you been?

I've been busy with exams (and CS:GO, let's be honest) and I haven't looked deeper into this, my repository is someway messed up too.

I'm thinking on closing, as I don't realize why are the mobs not acting well. If anyone realizes why, I'll probably close it soon.

PR #1734, opened 2015-02-08
Handle client 'leave bed' request

### LINKED ISSUE #1728: Server doesn't recognizes when players leave beds before it gets day
I discoverd some weird bed behavior:
First off, you can break a single piece of a bed with a piston but the other bed-part is still placed (kinda expected that to happen).
But where it gets weird is, when sleeping in it and leaving it before it gets day, the server seems to think that a player is still in that bed, so it recognizes that bed as "occupied". But only the part of the bed you clicked on when you went to sleep is occupied, not the other one. When you break one of the pieces, the other will be still placed (maybe because they are now to diffrent blocks that do not belong to each other).
Here is a video that should explain it a bit better:
https://www.youtube.com/watch?v=_BXrDC1oQKk&feature=youtu.be

tigerw:
* Fixes #1728

### DIFF
--- a/src/Blocks/BlockBed.cpp
+++ b/src/Blocks/BlockBed.cpp
@@ -1,13 +1,9 @@
 #include "Globals.h"
 #include "BlockBed.h"
 
-
-
-
 #include "BroadcastInterface.h"
 #include "Entities/../World.h"
 #include "Entities/Player.h"
-#include "WorldInterface.h"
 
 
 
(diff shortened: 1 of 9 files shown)


tigerw:
I think there will be a compilation error, but I am not sure why. VS told me that the virtual ScheduleTask() function in cWorld didn't override anything, even though its corresponding pure-virtual function in cWorldInterface was defined. It might be to do with the cTask class not being visible in cWorldInterface? I don't know. Halp.

worktycho:
The problem is that the definition in cWorldInterface specifies cTask in the global namespace whereas the definition in cWorld Takes a cWorld::cTask. If you want to move ScheduleTask up to WorldInterface you need to move cTask up as well.

tigerw:
I have attempted to move cTask and its descendants into the world interface, but since the other half of those classes are defined in cWorld, it will likely be a very large change. I think I'll skip that part for now.

Are the interfaces presently being used?

worktycho:
Sort of. A number of the block handlers take them. However I've given up on the project to decouple the code due to just how heavily coupled the code is. So long as we don't mind outright rejecting the possibility of other applications using significant portions of MCServers code like http://forum.mc-server.org/showthread.php?tid=1763 then you should be able to remove it.

bearbin:
Any update?

tigerw:
I had no internets :P

tigerw:
'Google-test' failed on one for some reason. Is this normal or indicative of issues?

tigerw:
Google test excepted. Is this mergeable?

tigerw:
The big green buttons calls out to me...

SphinxC0re:
Come on @madmaxoft ! Don't you see how much he is holding himself back? :smile:

NiLSPACE:
xoft might not have the time, because he's moving out around this time.

tigerw:
:P

In that case, nevermind.

PR #1741, opened 2015-02-15
Fixed Server.cpp

Creaprog:
Performance improvement.

### DIFF
--- a/Tools/ProtoProxy/Server.cpp
+++ b/Tools/ProtoProxy/Server.cpp
@@ -85,7 +85,7 @@ int cServer::Init(short a_ListenPort, short a_ConnectPort)
 
 void cServer::Run(void)
 {
-	LOG("Server running.");
+	LOGINFO("Server running.");
 	while (true)
 	{
 		sockaddr_in Addr;
--- a/Tools/ProtoProxy/Server.cpp
+++ b/Tools/ProtoProxy/Server.cpp
@@ -33,7 +33,7 @@ int cServer::Init(short a_ListenPort, short a_ConnectPort)
 		}
 	#endif  // _WIN32
 	
-	printf("Generating protocol encryption keypair...\n");
+	puts("Generating protocol encryption keypair...");
 	m_PrivateKey.Generate();
 	m_PublicKeyDER = m_PrivateKey.GetPubKeyDER();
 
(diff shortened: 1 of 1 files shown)


NiLSPACE:
Is there a reason why LOG isn't used here?

madmaxoft:
ProtoProxy had no logging at the time this piece of code was written. I guess we can now use it, since the logging framework has been pulled in already.

SphinxC0re:
Wait..! We have a logging framework? Where is it documented?

Creaprog:
Puts is simpler than printf but be aware that the former automatically appends a newline.

madmaxoft:
@SphinxC0re the LOG family of functions that are normally used by the server, they write to the console and into a logfile. Yes, we've had that since before I joined the project, which means long ago :)

@Creaprog change the calls to `LOG`, for style unification.

Creaprog:
I replaces puts by LOG ?

SphinxC0re:
Yes :dancer:

Creaprog:
Log is auto newline ?

NiLSPACE:
Yeah

NiLSPACE:
Anyone any idea why the build failed?

SphinxC0re:
It says
```error: ‘LOG’ was not declared in this scope```

**Comment by Creaprog** [comment]
Yes replaces puts by LOG is error. Why use function LOG ?
The function log is no write, where is the function LOG ?

**Comment by worktycho** [comment]
LOG should be declared in Logger.h. The main point of LOG is that it can be forwarded to places other than stdout. At the moment in outputs to a file as well but there is the potential to output LOG to plugins or hooks for android rather than reading the log file if we get that working again.

**Comment by Creaprog** [comment]
Ok, i add Logger.h.

**Comment by worktycho** [comment]
Logger.h should already have been included through globals.h. Also you're breaking protoproxy which defines the logging functions differently. You're problem is that the protoproxy globals only define LOGINFO, LOGWARNING and LOGERROR. You'll need to add a define to tools/protoproxy/Globals.h

File: Tools/ProtoProxy/Server.cpp [hunk 1]
```diff
@@ -85,7 +85,7 @@ int cServer::Init(short a_ListenPort, short a_ConnectPort)
 
 void cServer::Run(void)
 {
-	printf("Server running.\n");
+	LOG("Server running.");
```
bearbin: You missed one log here, so the build failed.

Creaprog:
@bearbin Sorry, i have forget 1 LOG.

bearbin:
Build worked, merging.

Creaprog:
Yes, thanks :)

PR #1742, opened 2015-02-15
Fixed guest.html

Creaprog:
Correction syntaxe html5.


bearbin:
The `<p>` tags are correct, but the other changes are HTML4, and HTML5 style rejects them.

Creaprog:
1) Where is the tag <p> and </p> ?
2) The problem <br> is incorrect. <br /> is correct.
3) Tag <strong> is better.

bearbin:
Fixed in 41e2eaa9

PR #1745, opened 2015-02-15
Terrainheight fixes and utility validation methods

tommysanterre:
Addesses issue https://github.com/mc-server/MCServer/issues/1606.
This pull request is not finished. I submitted it before going any further to give everyone an idead of the possible usage of the validation methods I propose.
I'm looking for validation on the new method before going ahead and replacing test on cChunkDef::Height and cChunkDef::Width everywhere else with call to those methods.

### DIFF
--- a/src/BlockEntities/HopperEntity.cpp
+++ b/src/BlockEntities/HopperEntity.cpp
@@ -127,7 +127,7 @@ void cHopperEntity::OpenNewWindow(void)
 /// Moves items from the container above it into this hopper. Returns true if the contents have changed.
 bool cHopperEntity::MoveItemsIn(cChunk & a_Chunk, Int64 a_CurrentTick)
 {
-	if (m_PosY >= cChunkDef::Height)
+	if (cChunkDef::IsRelCoordMoreThanChunkHeight(m_PosY))
 	{
 		// This hopper is at the top of the world, no more blocks above
 		return false;
--- a/src/ChunkData.cpp
+++ b/src/ChunkData.cpp
@@ -150,9 +150,7 @@ cChunkData::~cChunkData()
 
 BLOCKTYPE cChunkData::GetBlock(int a_X, int a_Y, int a_Z) const
 {
-	ASSERT((a_X >= 0) && (a_X < cChunkDef::Width));
-	ASSERT((a_Y >= 0) && (a_Y < cChunkDef::Height));
-	ASSERT((a_Z >= 0) && (a_Z < cChunkDef::Width));
+    cChunkDef::AssertCoordinatesIsValid(a_X, a_Y, a_Z);
 	int Section = a_Y / SectionHeight;
 	if (m_Sections[Section] != nullptr)
 	{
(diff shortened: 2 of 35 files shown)


madmaxoft:
You're using spaces for indent, which is wrong and the build will fail because of this.

src/ChunkData.cpp
```diff
@@ -150,9 +150,7 @@ cChunkData::~cChunkData()
 
 BLOCKTYPE cChunkData::GetBlock(int a_X, int a_Y, int a_Z) const
 {
-	ASSERT((a_X >= 0) && (a_X < cChunkDef::Width));
-	ASSERT((a_Y >= 0) && (a_Y < cChunkDef::Height));
-	ASSERT((a_Z >= 0) && (a_Z < cChunkDef::Width));
+    cChunkDef::AssertCoordinatesIsValid(a_X, a_Y, a_Z);
```
madmaxoft: Coordinates Is? I'd use "Are" instead.

src/ChunkDef.h
```diff
@@ -85,6 +85,50 @@ class cChunkDef
...
+    {
+        return !IsMoreThanChunkWidth(v) && !IsLessThanChunkWidth(v);
+    }
+
+    static bool IsValidCoordinates(int x, int y, int z)
+    {
+        return IsValidWidth(x) && IsValidHeight(y) && IsValidWidth(z);
+    }
+
+    static void AssertCoordinatesIsValid(int x, int y, int z)
```
madmaxoft: We use the `a_` prefix for all function parameters. Also, the coords are expected relative, so I'd suggest `a_RelX` and so on.
Similarly, the function name should probably reflect the fact that it operates on relative coords, so the `Rel` keyword should be included in the name. How about `AssertRelCoordsAreValid()`?

Similarly for the other functions.

src/ChunkData.cpp
tommysanterre: I agree

src/ChunkDef.h
tommysanterre: No problem, I will make those change. I had trouble finding a variable name for the width methods. That's why I simply used v. The Width methods can be called with an x or a z. I didn't want to use x as a parameter as I tought user of the code would look for the method with a z. Do you have a suggestion for the parameter name?

tommysanterre:
I really like the integration with Travis CI.

tommysanterre:
Do you prefer that I squash my commits before I submit my next pull request?

src/ChunkDef.h
madmaxoft: A generic `a_Coord` should be alright, can't think of anything better.

madmaxoft:
I usually squash my commits if they are part of a single change, such as trying to find a solution, then deciding on a different solution in the following commit. There's no point in keeping the old attempt in the repository.
On the other hand, I prefer to keep commits small, so that each individual change can be inspected and traced back, or even reverse-merged.

For this particular case, you should probably squash the commits so far with the one that will fix the issues into a single commit - it's a single logical change.

tommysanterre:
I fear this was bigger than I expected and i'm worried I might introduce regression. The version I currently have trigger assert that were not previously triggered. I fear this change would require a lot of review and testing.

I still pushed my change so you can see what kind of readability we can get from the change. I still believe it's an improvement on this front.

Overall I feel I under estimated the time required to make the change. However, I you feel that it's worth it and can take the time to review the change once i'm ready and help with the testing I could continue.

Instead I believe it would be better to look for the specific error cases.

madmaxoft:
I don't know about you, but for me the new style is much less readable than the old comparisons. The function names take a lot of time to decipher whether they are "less" or "more". I'd vote against this change.

src/BlockEntities/HopperEntity.cpp
```diff
@@ -127,7 +127,7 @@ void cHopperEntity::OpenNewWindow(void)
 /// Moves items from the container above it into this hopper. Returns true if the contents have changed.
 bool cHopperEntity::MoveItemsIn(cChunk & a_Chunk, Int64 a_CurrentTick)
 {
-	if (m_PosY >= cChunkDef::Height)
+	if (cChunkDef::IsRelCoordMoreThanChunkHeight(m_PosY))
```
madmaxoft: This change doesn't seem equivalent. The old code was checking for equality as well.

madmaxoft:
There's gonna be a lot more mistakes like this.

tommysanterre:
Yes that's what i'm worried about. I will take another approach.

PR #1750, opened 2015-02-18
Add files generated on build in gitignore

tommysanterre:
Some files were always showing up on git status. We should ignore them to avoid committing them by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,12 @@ nbproject/
 ipch/
 Win32/
 MCServer/MCServer
+MCServer/buildinfo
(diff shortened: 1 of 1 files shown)


madmaxoft:
What build tools are you using? These files don't come up for me with MSVC. I haven't tried for some time, but as far as I remember, gcc toolchain didn't produce any of those either.

tommysanterre:
Maybe it's when i ran the cibuild script. I will verify tomorrow.

madmaxoft:
That's possibly, the CI build could probably do that.

PR #1758, opened 2015-02-21
Added cCryptoHash namespace to Lua API

madmaxoft:
The new namespace provides access to the crypto hashing functions for Lua plugins. Each function has two variants, one returns the raw binary hash, the other returns the hash hex-encoded.


NiLSPACE:
Shall we merge this?

NiLSPACE:
Thank you :)

PR #1760, opened 2015-02-21
Update COMPILING.md

Samvbnm:
Tell the reader to make sure they have the latest buildtools installed.


bearbin:
The text should be outside of the code block.

madmaxoft:
It's not "the latest" that you need, it's "g++ 4.8 or newer". For some platforms, the latest is 4.6, which just won't make it.

Samvbnm:
Does the gcc version also matter or is that irrelevant?

bearbin:
Nothing else really matters apart from gcc and g++.

PR #1761, opened 2015-02-22
cNetwork: Added EnumLocalIPAddresses() function.

madmaxoft:
Added code to enumerate all local IP addresses on the machine.

I need someone with Linux expertise to fill in the Linux-specific code, I have no idea how to do this there.


SphinxC0re:
http://stackoverflow.com/questions/212528/get-the-ip-address-of-the-machine
This thread pretty much explains how it works

src/OSSupport/NetworkInterfaceEnum.cpp
```diff
@@ -0,0 +1,129 @@
...
+		if (dwRetVal != ERROR_SUCCESS)
+		{
+			LOG("GetAdaptersAddresses() failed: %u", dwRetVal);
+			return res;
+		}
+
+		// Enumerate all active adapters
+		for (auto pCurrAddresses = pAddresses; pCurrAddresses != nullptr; pCurrAddresses = pCurrAddresses->Next)
+		{
+			if (pCurrAddresses->OperStatus != 1)
```
tigerw: Could 1 be replaced with `IfOperStatusUp`?

https://msdn.microsoft.com/en-us/library/windows/desktop/aa366054(v=vs.85).aspx

src/OSSupport/NetworkInterfaceEnum.cpp
```diff
@@ -0,0 +1,172 @@
...
+			{
+				continue;
+			}
+
+			switch (ifa->ifa_addr->sa_family)
+			{
+				case AF_INET:
+				{  // IPv4
+					char AddressBuffer[INET_ADDRSTRLEN];
+					inet_ntop(AF_INET, TmpAddrPtr, &((struct sockaddr_in *)ifa->ifa_addr)->sin_addr, INET_ADDRSTRLEN);
```
worktycho: Clang is complaining about alignment issues in this cast

src/OSSupport/NetworkInterfaceEnum.cpp
```diff
@@ -0,0 +1,172 @@
...
+					inet_ntop(AF_INET6, &(((struct sockaddr_in6 *)ifa->ifa_addr)->sin6_addr), AddressBuffer, INET6_ADDRSTRLEN);
+					res.emplace_back(AddressBuffer);
+					break;
+				}
+				default:
+				{
+					LOG("Unknown address family: %i", ifa->ifa_addr->sa_family);
+					break;
+				}
+			}
```
madmaxoft: Move this entire `switch` into a separate function (just like the Windows version does), and replace the C-style casts with proper C++ casts.

src/OSSupport/NetworkInterfaceEnum.cpp
tigerw: I adopt the mentality to not think about alignment unless I must, so I require enlightenment 😃

madmaxoft: Alignment: Some CPUs will refuse to read memory if the read address is not a multiple of N bytes ("aligned"); usually N equals the size of memory to be read, so for example to read 4 bytes, they need to be at an address divisible by 4. If this condition is broken, the CPU can do many things; most of them raise a predefined exception, or trap; sometimes the OS handles that exception transparently so the program needn't care.

Consider this code:
```cpp
Int32 value;
// C++ guarantees that "value" is at a properly aligned address

Int32 * ptr = &value;
// Pointer to the aligned address, okay

char someArray[8];
// No guarantees about alignment for char (single-byte) datatype

char * ptr2 = &(someArray[2]);
// Pointer is valid and can retrieve chars from the address

Int32 * ptr3 = (Int32 *)ptr2;
// Pointer is valid, but most likely UNaligned - cannot read Int32s from that address
```
Unfortunately (or fortunately), both x86 and x64 don't care much about alignment, the reads and writes will work, although with some performance loss.
However, on ARM CPUs alignment must be watched, the CPU will trigger an exception for unaligned reads and writes.

tigerw:
This works. From research, there appear to be two methods to work around alignment errors:

Cast to void pointer and then to the required type.
Have all three structures, base, IPv4 and 6, in a union.

Which do you suggest?

src/OSSupport/NetworkInterfaceEnum.cpp
tigerw: Thanks, very useful information!

madmaxoft:
Still, you should restructure the code as I suggested, and possibly squash all your commits into a single one (we don't want those "compile fix" commits everywhere).

tigerw:
Of course, though I'm currently outside.

Could you suggest which solution you want? Union versus void pointer cast, or perhaps a better solution.

madmaxoft:
Have a look at the windows method, doesn't something similar work?

tigerw:
Windows directly casts to a sockaddr_in6, which in Linux works fine with GCC but makes Clang explode with improper alignment.

madmaxoft:
The proper solution is to declare a variable of the destination type, and `memcpy` the data into it:
```cpp
sockaddr_in sin;
memcpy(&sin, ifa->ifa_addr, sizeof(sin));
```

tigerw:
Merge? Example self test output on Travis:

```
Enumerating all IP addresses...
  127.0.0.1
  127.0.0.2
  172.30.174.223
  ::1
  2607:f700:8000:12d:5f08:89df:c0ae:6149
Done.
```

tigerw:
Merge?

madmaxoft:
Looks good, but I'd prefer someone with Linux expertise to look at this, too.

tigerw:
@bearbin, @worktycho, @Howaner and others - your services are requested.

worktycho:
Looks good to me.

PR #1762, opened 2015-02-22
Lua string compress

NiLSPACE:
Exported CompressString and UncompressString to the Lua API

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -115,10 +115,10 @@ static int tolua_CompressStringZLIB(lua_State * tolua_S)
 {
 	cLuaState S(tolua_S);
 	if (
-		!lua_isstring(tolua_S, 1) ||
+		!S.CheckParamString(1) &&
 		(
-		!lua_isnumber(tolua_S, 2) &&
-		!lua_isnil(tolua_S, 2)
+		!S.CheckParamNumber(2) ||
+		!S.CheckParamEnd(2)
 		)
 	)
 	{
@@ -146,8 +146,8 @@ static int tolua_UncompressStringZLIB(lua_State * tolua_S)
 {
 	cLuaState S(tolua_S);
 	if (
-		!lua_isstring(tolua_S, 1) ||
-		!lua_isnumber(tolua_S, 2)
+		!S.CheckParamString(1) &&
+		!S.CheckParamNumber(2)
 	)
 	{
 		cLuaState::LogStackTrace(tolua_S);
@@ -173,7 +173,10 @@ static int tolua_UncompressStringZLIB(lua_State * tolua_S)
 static int tolua_CompressStringGZIP(lua_State * tolua_S)
 {
 	cLuaState S(tolua_S);
-	if (!lua_isstring(tolua_S, 1))
+	if (
+		!S.CheckParamString(1) &&
+		!S.CheckParamEnd(2)
+	)
 	{
 		cLuaState::LogStackTrace(tolua_S);
 		return 0;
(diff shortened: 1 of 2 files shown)


NiLSPACE:
I noticed there are more functions like CompressStringGZIP, UncompressStringGZIP and InflateString. Should I export those as well?

madmaxoft:
These bindings are not good. they should not have the `Length` parameter, because that one is implicit in Lua strings (and C++ `std::string`).

You should use something like this:
```cpp
static int tolua_CompressString(lua_State * tolua_S)
{
  cLuaState S(tolua_S);
  // TODO: Check param types

  // Get the params:
  AString ToCompress;
  int CompressionLevel;
  S.GetStackValues(1, ToCompress, CompressionLevel);

  // Compress the string:
  AString res;
  CompressString(ToCompress.data(), ToCompress.size(), res, CompressionLevel);
  S.Push(res);
  return 1;
}
```

Then the plugins can use code like this:
```lua
local FileContents = cFile:ReadWholeFile(...)
local Compressed = CompressString(FileContents)
```

NiLSPACE:
I asume from looking at the code that there is an optional second parameter for the compressionlevel?

But how would the uncompress function look like? Would that look like this?:
```lua
UncompressString(a_CompressedString, a_UncompressedSize)
```

madmaxoft:
Yes, unfortunately the compression algorithm doesn't store the uncompressed length, so it needs it as an input.

How about putting all those functions into a `cStringCompression` namespace (similar to `cCryptoHash` namespace)? We are likely to add a few more in the future, so it would make sense.

NiLSPACE:
Good idea

NiLSPACE:
Also, how about we rename the current exported functions to CompressStringZLIB so they match the GZIP functions?

madmaxoft:
Makes sense.

NiLSPACE:
How is this?

madmaxoft:
We've got `CheckParamXYZ()` functions in `cLuaState` that check the param type and output an error if the param doesn't match, you should use those. Right now an erroneous parameter will cause a stack trace with no info whatsoever.

NiLSPACE:
Like this?

src/Bindings/ManualBindings.cpp
```diff
@@ -110,6 +111,146 @@ static int tolua_Clamp(lua_State * tolua_S)
...
 
 
+static int tolua_CompressStringZLIB(lua_State * tolua_S)
+{
+	cLuaState S(tolua_S);
+	if (
+		!S.CheckParamString(1) &&
+		(
+		!S.CheckParamNumber(2) ||
+		!S.CheckParamEnd(2)
```
madmaxoft: Extra indent should go here.

madmaxoft:
I'm wondering, don't we have the opposite function to Inflate? Makes no sense to have just one direction.

src/Bindings/ManualBindings.cpp
```diff
@@ -110,6 +111,146 @@ static int tolua_Clamp(lua_State * tolua_S)
 
 
 
+static int tolua_CompressStringZLIB(lua_State * tolua_S)
+{
+	cLuaState S(tolua_S);
+	if (
+		!S.CheckParamString(1) &&
+		(
+			!S.CheckParamNumber(2) ||
```
NiLSPACE: I'm confusing myself. Should that really be OR?

madmaxoft: Nope, this should be an AND, and all the others should be ORs. You want to fail if any of the top-level tests fail.

madmaxoft:
And that's why you should write a testing plugin for each new API that you make.

NiLSPACE:
I usualy test new API's using my Execute plugin

NiLSPACE:
The build failed because the Google test failed.

NiLSPACE:
Again the google test failed

NiLSPACE:
Hmm. Currently there is no way to know if the compression failed or succeeded. Maybe we should add another return value?

madmaxoft:
If the compression fails, the returned string is empty. Similar with decompression.

NiLSPACE:
So is this mergeable?

PR #1769, opened 2015-02-25
Snow weather

NiLSPACE:
As of MC 1.7 snow isn't bound the specific biomes, but rather the height of specific biomes.

### DIFF
--- a/src/BiomeDef.cpp
+++ b/src/BiomeDef.cpp
@@ -223,7 +223,7 @@ bool IsBiomeCold(EMCSBiome a_Biome)
 
 
 
-int GetSnowStartHeight(EMCSBiome a_Biome)
+unsigned GetSnowStartHeight(EMCSBiome a_Biome)
 {
 	switch (a_Biome)
 	{
@@ -237,7 +237,7 @@ int GetSnowStartHeight(EMCSBiome a_Biome)
 		case biColdTaigaM:
 		{
 			// Always snow
-			return -1;
+			return 0;
 		}
 	
 		case biExtremeHills:
@@ -341,7 +341,7 @@ int GetSnowStartHeight(EMCSBiome a_Biome)
 
 		default:
 		{
-			return -1;
+			return 0;
 		}
 	}
 }
--- a/src/BiomeDef.cpp
+++ b/src/BiomeDef.cpp
@@ -223,7 +223,7 @@ bool IsBiomeCold(EMCSBiome a_Biome)
 
 
 
-unsigned GetSnowStartHeight(EMCSBiome a_Biome)
+int GetSnowStartHeight(EMCSBiome a_Biome)
 {
 	switch (a_Biome)
 	{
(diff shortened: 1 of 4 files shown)


NiLSPACE:
Something that still has to be done is make the heights variate a little. You can now clearly see a line of snow on top of the mountains.

NiLSPACE:
Google test failed again.

src/BiomeDef.cpp
```diff
@@ -222,3 +222,130 @@ bool IsBiomeCold(EMCSBiome a_Biome)
...
+		case biIcePlains:
+		case biIceMountains:
+		case biFrozenRiver:
+		case biColdBeach:
+		case biColdTaiga:
+		case biColdTaigaHills:
+		case biColdTaigaM:
+		{
+			// Always snow
+			return -1;
```
madmaxoft: You may want to return 0, and return an `unsigned` from this function, so that it compares with `HEIGHTTYPE` without clang/gcc warnings.

NiLSPACE: done

NiLSPACE:
Clang failed :S

src/Chunk.cpp
```diff
@@ -874,80 +874,71 @@ void cChunk::ApplyWeatherToTop()
...
-			// TODO: Check light levels, don't snow over when the BlockLight is higher than (7?)
-			int Height = GetHeight(X, Z);
-			BLOCKTYPE TopBlock = GetBlock(X, Height, Z);
-			NIBBLETYPE TopMeta = GetMeta (X, Height, Z);
-			if (m_World->IsDeepSnowEnabled() && (TopBlock == E_BLOCK_SNOW))
+
+	// TODO: Check light levels, don't snow over when the BlockLight is higher than (7?)
+	int Height = GetHeight(X, Z);
+
+	if (GetSnowStartHeight(GetBiomeAt(X, Z)) > Height)
```
tigerw: unsigned-signed integer comparison?

src/Generating/FinishGen.cpp
```diff
@@ -470,30 +470,22 @@ void cFinishGenSnow::GenFinish(cChunkDesc & a_ChunkDesc)
 	{
 		for (int x = 0; x < cChunkDef::Width; x++)
 		{
-			switch (a_ChunkDesc.GetBiome(x, z))
+			int Height = a_ChunkDesc.GetHeight(x, z);
+			if (GetSnowStartHeight(a_ChunkDesc.GetBiome(x, z)) > Height)
```
tigerw: unsigned-signed integer comparison?

src/Chunk.cpp
madmaxoft: That would be because `GetHeight()` has a wrong return type, it should return a `HEIGHTTYPE`

NiLSPACE: So should I change GetHeight to return a `HEIGHTTYPE`?

madmaxoft: You know what, not now in this PR, it might create more changes in unrelated code. Revert back to `GetSnowStartHeight()` returning an int (but still keep the `0` return) and we can change the types later in another PR.

NiLSPACE: Done

madmaxoft:
Looks alright, merge if it compiles.

PR #1775, opened 2015-02-27
Correct all world height validations.

tommysanterre:
Addesses issue https://github.com/mc-server/MCServer/issues/1606.

No more assertion triggered.

### DIFF
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -571,7 +571,7 @@ void cFinishGenSnow::GenFinish(cChunkDesc & a_ChunkDesc)
 				continue;
 			}
 
-			if (!cBlockInfo::IsSnowable(a_ChunkDesc.GetBlockType(x, Height, z)) && (Height < cChunkDef::Height - 1))
+			if (!cBlockInfo::IsSnowable(a_ChunkDesc.GetBlockType(x, Height, z)) || (Height >= cChunkDef::Height - 1))
 			{
 				// The top block can't be snown over.
 				continue;
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -80,7 +80,7 @@ void cChestEntity::UsedBy(cPlayer * a_Player)
 void cChestEntity::OpenNewWindow(void)
 {
 	// TODO: cats are an obstruction
-	if ((GetPosY() + 1 < cChunkDef::Height) && cBlockInfo::IsSolid(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())))
+	if ((GetPosY() < cChunkDef::Height - 1) && cBlockInfo::IsSolid(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())))
 	{
 		// Obstruction, don't open
 		return;
@@ -99,7 +99,7 @@ void cChestEntity::OpenNewWindow(void)
 		
 		virtual bool Item(cChestEntity * a_Chest) override
 		{
-			if ((a_Chest->GetPosY() + 1 < cChunkDef::Height) && cBlockInfo::IsSolid(a_Chest->GetWorld()->GetBlock(a_Chest->GetPosX(), a_Chest->GetPosY() + 1, a_Chest->GetPosZ())))
+			if ((a_Chest->GetPosY() < cChunkDef::Height - 1) && cBlockInfo::IsSolid(a_Chest->GetWorld()->GetBlock(a_Chest->GetPosX(), a_Chest->GetPosY() + 1, a_Chest->GetPosZ())))
 			{
 				// Obstruction, don't open
 				return false;
--- a/src/Blocks/BlockBigFlower.h
+++ b/src/Blocks/BlockBigFlower.h
@@ -81,7 +81,7 @@ public:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, int a_RelX, int a_RelY, int a_RelZ, const cChunk & a_Chunk) override
 	{
-		return ((a_RelY > 0) && (a_Chunk.GetBlock(a_RelX, a_RelY - 1, a_RelZ) != E_BLOCK_AIR) && (a_RelY < cChunkDef::Height) && ((a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_AIR) || (a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_BIG_FLOWER)));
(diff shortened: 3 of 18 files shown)


src/BlockArea.cpp
```diff
@@ -441,10 +441,10 @@ bool cBlockArea::Read(cForEachChunkProvider * a_ForEachChunkProvider, int a_MinB
 		LOGWARNING("%s: MaxBlockY less than zero, adjusting to zero", __FUNCTION__);
 		a_MaxBlockY = 0;
 	}
-	else if (a_MaxBlockY > cChunkDef::Height)
+	else if (a_MaxBlockY >= cChunkDef::Height)
 	{
 		LOGWARNING("%s: MaxBlockY more than chunk height, adjusting to chunk height", __FUNCTION__);
-		a_MaxBlockY = cChunkDef::Height;
+		a_MaxBlockY = cChunkDef::Height - 1;
```
madmaxoft: This breaks stuff. `a_MaxBlockY` is supposed to be "one block past the maximum" at this point (see lines 423 - 425), so with your change there's no way to read the top of the chunk

src/Generating/ChunkDesc.cpp
```diff
@@ -336,10 +336,10 @@ void cChunkDesc::ReadBlockArea(cBlockArea & a_Dest, int a_MinRelX, int a_MaxRelX
 		LOGWARNING("%s: MaxRelY less than zero, adjusting to zero", __FUNCTION__);
 		a_MaxRelY = 0;
 	}
-	else if (a_MaxRelY > cChunkDef::Height)
+	else if (a_MaxRelY >= cChunkDef::Height)
 	{
 		LOGWARNING("%s: MaxRelY more than chunk height, adjusting to chunk height", __FUNCTION__);
-		a_MaxRelY = cChunkDef::Height;
+		a_MaxRelY = cChunkDef::Height - 1;
```
madmaxoft: Again, the max is supposed to be "one past", so the original test was correct.

src/Generating/FinishGen.cpp
```diff
@@ -124,9 +124,9 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
 	}
 
 	int MaxY = a_RelY + 2;
-	if (MaxY > cChunkDef::Height)  // Check if the coordinate is outside the chunk. If it it then adjust it.
+	if (MaxY >= cChunkDef::Height)  // Check if the coordinate is outside the chunk. If it it then adjust it.
 	{
-		MaxY = cChunkDef::Height;
+		MaxY = cChunkDef::Height - 1;
```
madmaxoft: The original code was correct, `MaxY` is used as an non-inclusive bound in the for loop below.

src/Generating/RoughRavines.cpp
```diff
@@ -139,7 +139,7 @@ class cRoughRavine :
 			int NumBlocks = (rnd % 3) + 2;
 			rnd = rnd / 4;
 			float Val = (float)(rnd % 256) / 128 - 1;  // Random float in range [-1, +1]
-			if (h + NumBlocks > cChunkDef::Height)
+			if (h + NumBlocks >= cChunkDef::Height)
```
madmaxoft: Wrong change.

src/World.cpp
```diff
@@ -533,7 +533,7 @@ void cWorld::InitializeSpawn(void)
 void cWorld::Start(void)
 {
 	m_SpawnX = 0;
-	m_SpawnY = cChunkDef::Height;
+	m_SpawnY = cChunkDef::Height - 1;
```
madmaxoft: Nope. We don't want to spawn *in* the terrain, but above it.

src/BlockArea.cpp
tommysanterre: Thank you, I didn't notice this. I could rename the variables to make it clearer we're working with intervals. Maybe using iterators like naming.
Example: 
    a_MinBlockX becomes a_BeginBlockX
    a_MaxBlockX becomes a_EndBlockX

src/Generating/ChunkDesc.cpp
tommysanterre: I could rename the variable here as mentioned above.

src/Generating/FinishGen.cpp
tommysanterre: And here as well.

src/BlockArea.cpp
madmaxoft: No, I don't like that naming, it doesn't convey that the coords need to be sorted, and it indicates no inclusion either. Additionally, the original parameters are inclusive, but the maxima are made exclusive within the function, so either way you name them, one instance will be wrong. I prefer the interface part to be right.

tommysanterre: Oh I see that's the goal of that block. Make them exclusive.

	// Include the Max coords:
	a_MaxBlockX += 1;
	a_MaxBlockY += 1;
	a_MaxBlockZ += 1;

My proposition doesn't make sense then, sorry.

src/BlockArea.cpp
```diff
@@ -441,7 +441,7 @@ bool cBlockArea::Read(cForEachChunkProvider * a_ForEachChunkProvider, int a_MinB
 		LOGWARNING("%s: MaxBlockY less than zero, adjusting to zero", __FUNCTION__);
 		a_MaxBlockY = 0;
 	}
-	else if (a_MaxBlockY > cChunkDef::Height)
+	else if (a_MaxBlockY >= cChunkDef::Height)
```
madmaxoft: I still think this is a wrong change, we need to be able to read areas with full chunk height (`a_MaxBlockY` is exclusive at this point due to line 425)

src/BlockArea.cpp
```diff
@@ -517,7 +517,7 @@ bool cBlockArea::Write(cForEachChunkProvider * a_ForEachChunkProvider, int a_Min
 		LOGWARNING("%s: MinBlockY less than zero, adjusting to zero", __FUNCTION__);
 		a_MinBlockY = 0;
 	}
-	else if (a_MinBlockY > cChunkDef::Height - m_Size.y)
+	else if (a_MinBlockY >= cChunkDef::Height - m_Size.y)
```
madmaxoft: I'm not sure about this change either; if it is indeed needed, then the assignment on line 523 is wrong.

src/BlockArea.cpp
tommysanterre: You're right.

src/BlockArea.cpp
tommysanterre: Here I thought it was possible to call FastSetBlock with an invalid value but I can't reproduce it even when entering incorrect values with the debugger. I will revert this change also.

tommysanterre:
@madmaxoft I really appreciate the reviews you have done for me. I've revised my changes once more and corrected some mistake I also fixed the issue highlighted by your last review. Could you review the code once more?

madmaxoft:
Thanks for all the hard work :)

Issue #1538, opened 2014-10-15
Server-side teleportation shouldn't call "HOOK_PLAYER_MOVING"


LO1ZB:
When a player got teleported, it shouldn't call "HOOK_PLAYER_MOVING" because it makes it impossible to know, if the player is moving normal, or if he got teleported. Couldn't a teleportation call a hook like "HOOK_PLAYER_TELEPORTING" or something simulator instead?

joshi07:
i implemented the teleport hook, its calling both hooks (move + teleport) when you teleport. but only the moving hook when you move. but i dont know hot to do a pull request from my fork.

joshi07:
there is a hook now #1790 for the teleport of all entities.

avarun42:
Is this issue solved already then?

Cl1608Ho:
Well, there is a hook for teleporting, but I'm not sure if a teleport still calls moving or not.

PR #1776, opened 2015-02-27
Added Teleport Hook, Updated HookNotify Plugin

joshi07:
i added a Hook for the teleport. 

Plugins may or may not deny any teleportation.

Updated HookNotify plugin with it.

as requested in  #1538

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -2392,10 +2392,10 @@ static int tolua_md5HexString(lua_State * tolua_S)
 
 	// Convert the md5 checksum to hex string:
 	std::stringstream Output;
-	Output << std::hex << std::setw(2) << std::setfill('0');
+	Output << std::hex << std::setfill('0');
 	for (size_t i = 0; i < ARRAYCOUNT(md5Output); i++)
 	{
-		Output << static_cast<unsigned short>(md5Output[i]);  // Need to cast to a number, otherwise a char is output
+		Output << std::setw(2) << static_cast<unsigned short>(md5Output[i]);  // Need to cast to a number, otherwise a char is output
 	}
 	lua_pushlstring(tolua_S, Output.str().c_str(), Output.str().size());
 	return 1;
@@ -2438,10 +2438,10 @@ static int tolua_sha1HexString(lua_State * tolua_S)
 
 	// Convert the sha1 checksum to hex string:
 	std::stringstream Output;
-	Output << std::hex << std::setw(2) << std::setfill('0');
+	Output << std::hex << std::setfill('0');
 	for (size_t i = 0; i < ARRAYCOUNT(sha1Output); i++)
 	{
-		Output << static_cast<unsigned short>(sha1Output[i]);  // Need to cast to a number, otherwise a char is output
+		Output << std::setw(2) << static_cast<unsigned short>(sha1Output[i]);  // Need to cast to a number, otherwise a char is output
 	}
 	lua_pushlstring(tolua_S, Output.str().c_str(), Output.str().size());
 	return 1;


NiLSPACE:
I don't think this is right. You can also teleport entities, so I think you should either rename this to OnPlayerTeleport, or make it accept all entities. I vote for accept all entities ;)

joshi07:
you're right, i will look over this to accept all entities. :) 

why did the Travis CI thingy fail ? it just compiled fine on my Visual express ?!

i hoped my first contribution would go through in first try xD *dreaming*

src/Entities/Player.cpp
```diff
@@ -1274,13 +1274,19 @@ unsigned int cPlayer::AwardAchievement(const eStatistic a_Ach)
...
 void cPlayer::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
 {
-	SetPosition(a_PosX, a_PosY, a_PosZ);
-	m_LastGroundHeight = (float)a_PosY;
-	m_LastJumpHeight = (float)a_PosY;
-	m_bIsTeleporting = true;
 
-	m_World->BroadcastTeleportEntity(*this, GetClientHandle());
-	m_ClientHandle->SendPlayerMoveLook();
+	if (!cRoot::Get()->GetPluginManager()->CallHookPlayerTeleport(*this, m_LastPos, GetPosition())) // if the Plugin doesn't denie the teleport
```
NiLSPACE: You need two spaces before you start a comment
```
MyFunction)  // My comment
```

That's why the build failed

joshi07:
Thank you very much, i will think on that next time :)

joshi07:
is it possible to call the hook from multiple funktions? 

like from this places ?

cEntity::TeleportToCoords
and
cPlayer::TeleportToCoords 

so the hook would accept all entities ?

NiLSPACE:
Yes, you can call it from wherever you want.

NiLSPACE:
You do have to change the parameters in the hook itself though ;)

joshi07:
how can i add changes from my rebo to this PR ?

joshi07:
Yay it passed :) if you want to hook for player teleports, you have to use the IsPlayer funktion to find out if the Entity is a player or some other stuff.

src/Entities/Entity.cpp
```diff
@@ -1632,8 +1632,11 @@ void cEntity::TeleportToEntity(cEntity & a_Entity)
 
 void cEntity::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
 {
-	SetPosition(a_PosX, a_PosY, a_PosZ);
-	m_World->BroadcastTeleportEntity(*this);
+	if (!cRoot::Get()->GetPluginManager()->CallHookEntityTeleport(*this, m_LastPos, GetPosition()))  // if the Plugin doesn't denie the teleport
```
madmaxoft: `deny` is the proper spelling. Also, it might be a good idea to put the comment above the condition, so that the line is not that long. I'd write it like this:
```cpp
// If plugins agree with the teleport, set the new position:
if (!cRoot::Get()->...)
```

madmaxoft: Does the `GetPosition()` return the new position already? I don't think so.

src/Entities/Player.cpp
```diff
@@ -1274,13 +1274,19 @@ unsigned int cPlayer::AwardAchievement(const eStatistic a_Ach)
...
 void cPlayer::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
 {
-	SetPosition(a_PosX, a_PosY, a_PosZ);
-	m_LastGroundHeight = (float)a_PosY;
-	m_LastJumpHeight = (float)a_PosY;
-	m_bIsTeleporting = true;
 
-	m_World->BroadcastTeleportEntity(*this, GetClientHandle());
-	m_ClientHandle->SendPlayerMoveLook();
+	if (!cRoot::Get()->GetPluginManager()->CallHookEntityTeleport(*this, m_LastPos, GetPosition()))  // if the Plugin doesn't denie the teleport
```
madmaxoft: Same as in cEntity - put the comment above and check that `GetPosition()` is the right thing to call.

joshi07:
thank you, i will look into that.

joshi07:
now this works, it returns the new position, so the plugin can decide to allow the teleport.

madmaxoft:
Could you rebase onto current master, and squash those "fix spaces" commits?

joshi07:
i can and will tomorrow! And with "return", i didn't mean the hook in C++  but the funktion in LUA :)

madmaxoft:
The Lua function doesn't return anything except for the `bool` either, does it?

joshi07:
You're right, i need to work on my writing, so my writing matches my thougts!

It was late and i was tired yesterday.

What is it called, if the funktion call comes with arguments for the function? (Thats what i meant with return for the LUA) 

Does it "pass" the arguments? I've never talked in english about programming.

madmaxoft:
I'd use "function takes the new position as an argument", but I'm not a native English speaker either, so it needn't be the best one.
I think that all those four lines of comments are unneeded, only the first line, "Ask plugins if they allow the teleport" should be kept, remove the rest.

worktycho:
I am a native speaker and in my experience you'd use "function takes the new position as an argument". Generally I use takes for describing the interface, and passes for describing the action of calling the function.

joshi07:
im sorry to keep asking stupid question. But how can i rebase my fork and add my changes again? do i have to delete my fork and refork? i tried some stuff but i coudn't resolve it. Im pretty new to Github.

joshi07:
i think i f*cked up :o how can i clean up that mess ? maybe its best to delete my fork and readd my changes ?so i get a clean hystory ?

madmaxoft:
Normally I work with command-line git, so here's what I'd do:
```
git remote add upstream https://github.com/mc-server/MCServer.git
git fetch --all
git rebase --interactive upstream/master
```
At this point, git will open your favorite text editor to ask you to edit a file that has a list of all the commits, along with instructions; you can select an action for each commit. Save the file and exit, git will continue with the rebase. If there are any conflicts, you will need to fix them, commit and continue with the rebase.

joshi07:
for every commit which is not mine i get this error with untracked files.

am i doing something wrong? I corrected the conflicts by changing the conflicting files content and used 
`git add *filefapth*` 
to add the file.
then i put
`git rebase --continue`
After that when my commits are done and he tries to replay the other commits (none of mine) this error pops up.
```
rebase in progress; onto d4b505d
You are currently rebasing branch 'master' on 'd4b505d'.

Untracked files:
        sfmtcpoy.cqq.txt

nothing added to commit but untracked files present
The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git reset'
Could not apply cfce882e2f003979728da7fa77c2c5df43b26cb8... Fixed door placement.
```
i don't know what to do there

madmaxoft:
I guess in this situation the easiest would be to store your changes elsewhere, reset the git repo back to upstream/master and re-do the changes.

PR #1780, opened 2015-02-28
Modified Player class to use int data type for experience

DevToaster:
The old MC protocal no longer uses short int for experience.

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -105,15 +105,15 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 		SetPosX(World->GetSpawnX());
 		SetPosY(World->GetSpawnY());
 		SetPosZ(World->GetSpawnZ());
-		SetBedPos(Vector3i((int)World->GetSpawnX(), (int)World->GetSpawnY(), (int)World->GetSpawnZ()));
+		SetBedPos(Vector3i(static_cast<int>(World->GetSpawnX()), static_cast<int>(World->GetSpawnY()), static_cast<int>(World->GetSpawnZ())));
 		
 		LOGD("Player \"%s\" is connecting for the first time, spawning at default world spawn {%.2f, %.2f, %.2f}",
 			a_PlayerName.c_str(), GetPosX(), GetPosY(), GetPosZ()
 		);
 	}
 
-	m_LastJumpHeight = (float)(GetPosY());
-	m_LastGroundHeight = (float)(GetPosY());
+	m_LastJumpHeight = static_cast<float>(GetPosY());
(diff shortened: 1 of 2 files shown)


src/Entities/Player.cpp
```diff
@@ -382,19 +382,19 @@ bool cPlayer::SetCurrentExperience(short int a_CurrentXp)
...
-		LOGWARNING("Attempt was made to increment Xp by %d, which overflowed the short datatype. Ignoring.", a_Xp_delta);
+		LOGWARNING("Attempt was made to increment Xp by %d, which overflowed the int datatype. Ignoring.", a_Xp_delta);
 		return -1;  // Should we instead just return the current Xp?
 	}
 
 	m_CurrentXp += a_Xp_delta;
 
 	// Make sure they didn't subtract too much
-	m_CurrentXp = std::max<short>(m_CurrentXp, 0);
+	m_CurrentXp = std::max<int>(m_CurrentXp, 0);
```
madmaxoft: The `<int>` specification may be unnecessary now, did you check?

src/Entities/Player.cpp
```diff
@@ -1725,8 +1725,8 @@ bool cPlayer::LoadFromFile(const AString & a_FileName, cWorldPtr & a_World)
 	m_FoodSaturationLevel = root.get("foodSaturation", MAX_FOOD_LEVEL).asDouble();
 	m_FoodTickTimer       = root.get("foodTickTimer",  0).asInt();
 	m_FoodExhaustionLevel = root.get("foodExhaustion", 0).asDouble();
-	m_LifetimeTotalXp     = (short) root.get("xpTotal", 0).asInt();
-	m_CurrentXp           = (short) root.get("xpCurrent", 0).asInt();
+	m_LifetimeTotalXp     = (int) root.get("xpTotal", 0).asInt();
+	m_CurrentXp           = (int) root.get("xpCurrent", 0).asInt();
```
madmaxoft: No need to cast, `asInt()` returns an `int` already.

src/Entities/Player.cpp
```diff
@@ -307,18 +307,18 @@ short cPlayer::CalcLevelFromXp(short a_XpTotal)
 	// level 30+
 	if (a_XpTotal > XP_TO_LEVEL30)
 	{
-		return (short) (151.5 + sqrt( 22952.25 - (14 * (2220 - a_XpTotal)))) / 7;
+		return (int) (151.5 + sqrt( 22952.25 - (14 * (2220 - a_XpTotal)))) / 7;
```
madmaxoft: Could you change all the casts to `static_cast<int>` as well? We're trying to get rid of the C-style casts :)

DevToaster:
Changed casts and removed unneeded casts.

PR #1784, opened 2015-03-01
Added Steppy height generator.

madmaxoft:
Creates biome-agnostic landscapes like this:
steppy2

Ref.: http://forum.mc-server.org/showthread.php?tid=1801


src/Generating/HeiGen.cpp
```diff
@@ -10,6 +10,66 @@
...
+			std::make_shared<cProtIntGenRndBetween>         (a_Seed + 6, 60,
+			std::make_shared<cProtIntGenRndChoice>          (a_Seed + 7, 10, 50, 50,
+			std::make_shared<cProtIntGenSmooth>             (a_Seed + 8,
+			std::make_shared<cProtIntGenZoom>               (a_Seed + 9,
+			std::make_shared<cProtIntGenRndChoice>          (a_Seed + 1, 10, 50, 50,
+			std::make_shared<cProtIntGenAddRnd>             (a_Seed + 2, 2,
+			std::make_shared<cProtIntGenZoom>               (a_Seed + 3,
+			std::make_shared<cProtIntGenZoom>               (a_Seed + 4,
+			std::make_shared<cProtIntGenChoice>             (a_Seed + 5, 10)
+		)))))))))))))))))))))));
```
worktycho: This has reminded me that I'm still not happy about this pipeline syntax, but that's a separate issue.

madmaxoft:
So how would you propose to change it?

worktycho:
I want to be able to do 
```
MakeIntGen<cIntGenChoice> (a_Seed + 9, 10)
>>
>> MakeIntGen<cIntGenAddRnd> (a_Seed + 2, 2)

+ std::make_shared<cProtIntGenZoom> (a_Seed + 3,
+ std::make_shared<cProtIntGenZoom> (a_Seed + 4,
+

madmaxoft:
That looks even worse than my "pipelines", imho.

worktycho:
I think its clearer, and you don't have to worry about parenthesis matching. I can make it look like a unix pipeline if you'd rather. If you'd rather not then close the PR.

madmaxoft:
Using a pipe character could work. However, your change still reverts the logical flow; I guess I'll survive that. But the templating code needs improvement, it was completely unreadable.

worktycho:
Really, I always thought of the flow in you're code as backward. By code follows the convention in most European human languages and shells that data flows from left to right.

madmaxoft:
It follows normal math notation: "apply function f to the result of applying function g to x" is noted as "f(g(x))"

worktycho:
or in point-free style (g . f) (x)

SphinxC0re:
I do agree with @worktycho here. We are using the decorator pattern anyway so why not making it easier and more readable?

tigerw:
So many undocumented generators everywhere!

madmaxoft:
So, can we merge this and then modify the syntax once @worktycho figures it all out? :)

tigerw:
So many generators...

SphinxC0re:
Hey! :boom: @worktycho wake up! Wake up! It's your time to shine :laughing:

worktycho:
I'm fine to merge. My PR is a separate but related issue to the new generator.

PR #1786, opened 2015-03-01
Added pipelining syntax for generators

worktycho:
Also converted part of cBioGenGrown as an example.

### DIFF
--- a/src/Generating/IntGen.h
+++ b/src/Generating/IntGen.h
@@ -70,14 +70,19 @@ public:
 	
 };
 
-template<size_t size, class... Args>
-struct PackToInt
-{
-	enum
-	{
-		value = size - sizeof...(Args),
-	};
+// Code adapted from http://stackoverflow.com/questions/7858817/unpacking-a-tuple-to-call-a-matching-function-pointer
+
+template<int ...>
+struct seq { };
+
+template<int N, int ...S>
+struct gens : gens<N-1, N-1, S...> { };
+
+template<int ...S>
+struct gens<0, S...> {
+  typedef seq<S...> type;
 };
+
 
 template<class Gen, class... Args>
 class cIntGenFactory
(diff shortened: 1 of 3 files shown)


madmaxoft:
I have to say, I don't like this code at all. It is extremely twisted, very hard to understand. And the original code requires the Underlying generator assignment within constructor, which the new code doesn't have.

worktycho:
If you look, I haven't changed the constructors of the generators. Instead, I created a wrapper type cIntGenFactory to hold the args until we can construct the generator. As for the twistedness, it came out a lot more twisted than I expected and I am going to see if I can clean it.

worktycho:
Also check basic style fails for `std::shared_ptr<Gen> construct(LhsGen&& lhs)`. This is a general problem for declarations using rvalue references and templates that do not contain `T` in them.

worktycho:
I tried to simplify this then found a bug and ended up increasing the twistyness.

madmaxoft:
I thought we agreed on changing the operator to pipe. As it is now, there's just way too many anglebrackets.

I suppose style checks are failing for you as well, you have curly braces at end of lines.

worktycho:
I compleatly forgot about changing the operator change. The style checks are partially failing on brackets and partially on false positive.

worktycho:
What do you think of the changes now?

src/Generating/IntGen.h
```diff
@@ -62,9 +67,70 @@ class cIntGen
...
+
+// Code adapted from http://stackoverflow.com/questions/7858817/unpacking-a-tuple-to-call-a-matching-function-pointer
+
+template<int ...>
+struct seq
+{
+};
+
+template<int N, int ...S>
+struct gens : gens<N-1, N-1, S...>
```
madmaxoft: Spaces around the operators

madmaxoft:
Functionally, it's okay. Stylistically it's rather poor. Spaces are missing around most symbols, `a_CamelCase` is not used etc.

worktycho:
I've fixed the style issues I can identify.

madmaxoft:
Looks good enough, merge if it compiles.

PR #1790, opened 2015-03-05
Added OnTeleportEntity hook for plugins.

joshi07:
Plugins may or may not allow teleport to the new position.
Updated the HookNotify plugin with it.


joshi07:
i reset my fork and readded my changes. This was indeed the easiest way of solving my problem.

Thanks madmaxoft.

joshi07:
how can i update the wiki for the API ?

madmaxoft:
The Lua API has its description in the APIDump plugin in the repository, so have a look there. The webpages with the API are built and uploaded once a day by a script.

PR #1804, opened 2015-03-10
Streamline Travis testing

### LINKED ISSUE #1795: Somewhat inefficient continuous integration?
Currently builds run for pull requests and a simulation of a merge of that into master (?).

----

Travis does all of these:

1. Release, x64, GCC
2. Debug, x64, GCC
3. Release, x86, GCC
4. Debug, x86, GCC
5. Release, x64, Clang
6. Debug, x64, Clang
7. Release, x86, Clang
8. Debug, x86, Clang
9. Coverage, GCC

twice (pr and push).

----

Suggestions:

* Clang's always more picky, so there's no need to test GCC separately if it works on the former.
* If it works for 64 bit, it's unlikely not to work for 32.
* What does the coverage build do?
* What does the `uploadCoverage.sh` script do? An excerpt from its output:
```sh
StringUtils.cpp.gcno:version '408*', prefer '406*'
Segmentation fault
TCPLinkImpl.cpp.gcno:version '408*', prefer '406*'
Segmentation fault
ServerHandleImpl.cpp.gcno:version '408*', prefer '406*'
Segmentation fault
```

----

Suggested remedy:

1. Release, x64, Clang
2. Debug, x64, Clang
3. Coverage, GCC

tigerw:
* Fixes #1795
- Reduced matrix build count from 9 to 3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: cpp
-compiler:
-  - gcc
-  - clang
+compiler: clang
 
 before_install:
   - if [ "$TRAVIS_MCSERVER_BUILD_TYPE" == "COVERAGE" ]; then sudo pip install cpp_coveralls; fi
@@ -29,8 +27,6 @@ after_success:
 env:
   - TRAVIS_MCSERVER_BUILD_TYPE=RELEASE MCSERVER_PATH=./MCServer
   - TRAVIS_MCSERVER_BUILD_TYPE=DEBUG   MCSERVER_PATH=./MCServer_debug
-  - TRAVIS_MCSERVER_BUILD_TYPE=RELEASE TRAVIS_MCSERVER_FORCE32=1 MCSERVER_PATH=./MCServer
-  - TRAVIS_MCSERVER_BUILD_TYPE=DEBUG   TRAVIS_MCSERVER_FORCE32=1 MCSERVER_PATH=./MCServer_debug
 
 matrix:
   include:
--- a/CIbuild.sh
+++ b/CIbuild.sh
@@ -10,8 +10,8 @@ cmake . -DBUILD_TOOLS=1 -DSELF_TEST=1;
 cd src
 lua CheckBasicStyle.lua
 cd ..
-make -j 4;
-make -j 4 test;
+make -j 2;
+make -j 2 test;
 cd MCServer/;
 if [ "$TRAVIS_MCSERVER_BUILD_TYPE" != "COVERAGE" ]
 	then echo stop | $MCSERVER_PATH;
--- a/CIbuild.sh
+++ b/CIbuild.sh
@@ -10,8 +10,8 @@ cmake . -DBUILD_TOOLS=1 -DSELF_TEST=1;
 cd src
 lua CheckBasicStyle.lua
 cd ..
-make -j 2;
-make -j 2 test;
+make -j 4;
+make -j 4 test;
 cd MCServer/;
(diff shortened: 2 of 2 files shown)


bearbin:
If it works I'd have no problem with merging.

worktycho:
I think we need to do a comparison of speeds between 2 and 4 jobs. I would not expect travis vms to have enough parallel resources to run 4 jobs. They only have 1.5 CPU cores.

bearbin:
I didn't notice that. Yeah 2 would be better, it would reduce the amount the logs get jumbled up.

tigerw:
I was under the impression that logs only became misaligned with really large thread numbers. Haven't observed any disordering with four threads (on a physical computer) yet.

Isn't the general consensus twice the number of cores available? So maybe three threads. Or four for luck. One can't to too much harm, surely.

Anyway, the build doesn't seem to be starting whatever the concurrency settings are.

tigerw:
If Travis ever decides to upgrade their OS version to one that comes with GCC 4.8, speeds could be farther improved by using their 'container based infrastructure', which has ~~over 9000~~ 2 cores and 4GB of memory. It doesn't allow su elevation, so at present, it cannot be activated as the apt GCC upgrade commands won't work.

tigerw:
Oh look, it does get jumbled. Fancy that.

tigerw:
It inexplicably fails when linking polarssl on GCC. Clang builds *look* faster though.

.travis.yml
```diff
@@ -16,9 +14,6 @@ install:
 
   # lua, needed for style checking and possibly later on for bindings generation
   - sudo apt-get install -qq lua5.1
-  
-  # g++4.8
-  - if [ "$CXX" == "g++" ]; then export CXX="g++-4.8"; export CC="gcc-4.8"; fi
```
worktycho: If there are any build using gcc we need this

tigerw: Whewps. Thanks!

tigerw:
Many thanks for the diagnosis, @worktycho.

With 4 threads:
`5299.1  4 min 30 sec  about 7 hours ago  clang TRAVIS_MCSERVER_BUILD_TYPE=RELEASE MCSERVER_PATH=./MCServer Linux` (4m 30s)
`5299.2  4 min 14 sec  about 7 hours ago  clang TRAVIS_MCSERVER_BUILD_TYPE=DEBUG MCSERVER_PATH=./MCServer_debug Linux` (4m 14s)

Without:
`5298.5  5 min 41 sec  about 8 hours ago  clang TRAVIS_MCSERVER_BUILD_TYPE=RELEASE MCSERVER_PATH=./MCServer Linux` (5m 41s)
`5298.6  5 min 17 sec  about 8 hours ago  clang TRAVIS_MCSERVER_BUILD_TYPE=DEBUG MCSERVER_PATH=./MCServer_debug Linux` (5m 17s)

The experimental procedure is somewhat flawed, since the second test is for the NetworkIPEnum branch, but it seems faster.

tigerw:
Merge?

worktycho:
My main problem with the experiment is that in values are within normal variations. For example job 5304 has builds lasting between 8 min 22 sec and 3 min 55 sec.

madmaxoft:
For me the main point is that there are many less CI build jobs, so the overall time is much shorter - ~12 minutes instead of ~40. I'd say merge.

worktycho:
My view is similar but I think we should remove the change to the number of parallel make jobs then merge. That can be done in a seperate PR.

tigerw:
Muchas gracias!

Issue #490, opened 2013-12-31
Change command string-splitting


madmaxoft:
Currently string splitting (StringSplit()) is implemented so that it splits at each occurrence of any of the delimiting characters. This is not well-suited for commands that may take a file path as a parameter, because a file path may contain spaces, but that would split the argument in two.
Therefore I'd like to change the way commands are split. To minimize the impact of this change on other parts of MCS, I'll implement it as a new function, StringSplitWithQuotes(). It would take quotes and doublequotes into account. Compare:
```cpp
StringSplit("/cmd param1 param2");
// 3 items: {"/cmd", "param1", "param2"}

StringSplit("/cmd c:\My Documents\My server");
// 4 items: {"/cmd", "c:\My", "Documents\My", "server"}

StringSplit("/cmd 'c:\My Documents\My server'");
// 4 items: {"/cmd", "'c:\My", "Documents\My", "server'"}

StringSplitWithQuotes("/cmd param1 param2");
// 3 items: {"/cmd", "param1", "param2"}

StringSplitWithQuotes("/cmd c:\My Documents\My server");
// 4 items: {"/cmd", "c:\My", "Documents\My", "server"}

StringSplitWithQuotes("/cmd 'c:\My Documents\My server'");
// 2 items (!): {"/cmd", "c:\My Documents\My server"}
```

This shouldn't break any existing plugins, since none of them actually handle the space-related issues, and it would allow us to write better plugins - for example the WorldEdit plugin could load schematic files from disk files, or the Gallery plugin could allow spaces in area names.
Is there anything obvious that I've missed?

madmaxoft:
Reopening, the `StringSplitWithQuotes()` as a separate function still seems like a good idea. It needn't be used anywhere in the server yet, but the plugins could make use of it after #493 is implemented.

madmaxoft:
Implemented.

PR #1807, opened 2015-03-11
Fixes #493 and #490

### LINKED ISSUE #493: Add "full text" parameter to command handlers API
If #490 is implemented, it would be non-trivial for plugins to go from the split array back to original command text, which sometimes is more useful than the split. Therefore it would be nice if the command handlers received the full text of the given command as a third parameter. Since Lua doesn't enforce parameter counts in function calls, this wouldn't break any existing plugin, but will make life easier later on. This would come in handy especially if we ever implement a "user command aliases" plugin.

flx5:
I hope this should fix the referenced issues properly?

(Tested only on Windows; Linux to be done)


madmaxoft:
Thanks for this, finally someone who can understand the Lua bindings a bit and can help with good code! :)

I do have some concerns about the code, I'll comment on it in a moment.

src/Bindings/ManualBindings.cpp
```diff
@@ -266,6 +266,21 @@ static int tolua_StringSplit(lua_State * tolua_S)
 
 
 
+static int tolua_StringSplitWithQuotes(lua_State * tolua_S)
+{
+	cLuaState LuaState(tolua_S);
+	std::string str = (std::string)tolua_tocppstring(LuaState, 1, 0);
+	std::string delim = (std::string)tolua_tocppstring(LuaState, 2, 0);
```
madmaxoft: We have a nice template function `cLuaState::GetStackValues()` that is better at doing this - it can handle strings with embedded NULs etc. I'd prefer if you used it instead.

src/Bindings/ManualBindings.cpp
```diff
@@ -3663,6 +3678,7 @@ void ManualBindings::Bind(lua_State * tolua_S)
 		// Globals:
 		tolua_function(tolua_S, "Clamp",              tolua_Clamp);
 		tolua_function(tolua_S, "StringSplit",        tolua_StringSplit);
+		tolua_function(tolua_S, "StringSplitWithQuotes", tolua_StringSplitWithQuotes);
```
madmaxoft: Normally we try to keep the vertical alignment in cases like these - so add the three spaces in the lines above and below this one.

src/Bindings/Plugin.h
```diff
@@ -111,12 +111,12 @@ class cPlugin
 	Command permissions have already been checked.
 	Returns true if command handled successfully
 	*/
-	virtual bool HandleCommand(const AStringVector & a_Split, cPlayer & a_Player) = 0;
+	virtual bool HandleCommand(const AStringVector & a_Split, cPlayer & a_Player, const AString & fullCommand) = 0;
```
madmaxoft: We use `a_ParamName` naming convention for function parameters, please do so as well, for consistency.

src/StringUtils.cpp
```diff
@@ -140,6 +140,48 @@ AStringVector StringSplit(const AString & str, const AString & delim)
...
+
+		results.push_back(current);
+		Prev = cutAt + 1;
+	}
+
+	if (Prev < str.length())
+	{
+		AString current = str.substr(Prev);
+
+		if ((current.length() >= 2) && ((current.front() == '"') || (current.front() == '\'')) && (current.front() == current.back()))
```
madmaxoft: This condition should be probably split onto multiple lines, along the `&&` 'separators'. Also a comment would be nice:
```cpp
// If the remnant is wrapped in matching quotes, remove them:
if (
  (current.length() >= 2) && 
  ((current.front() == '"') || (current.front() == '\'')) && 
  (current.front() == current.back())
)
{
  ...
```

flx5:
Thanks for the feedback.
The new commit should match the coding conventions :)

src/StringUtils.cpp
```diff
@@ -140,6 +140,54 @@ AStringVector StringSplit(const AString & str, const AString & delim)
...
+	AStringVector results;
+
+	size_t cutAt = 0;
+	size_t Prev = 0;
+	size_t cutAtQuote = 0;
+
+	while ((cutAt = str.find_first_of(delim, Prev)) != str.npos)
+	{
+		AString current = str.substr(Prev, cutAt - Prev);
+		if ((current.at(0) == '"') || (current.at(0) == '\''))
```
tigerw: Is std::string::at preferred, as opposed to unchecked [] operators? I understand that since it throws exceptions and no exception handlers exist around here, it provides no benefits whilst negatively impacting speed.

src/StringUtils.cpp
```diff
@@ -140,6 +140,54 @@ AStringVector StringSplit(const AString & str, const AString & delim)
...
+		if (
+			(current.length() >= 2) &&
+			((current.front() == '"') || (current.front() == '\'')) &&
+			(current.front() == current.back())
+		)
+		{
+			current = current.substr(1, current.length() - 2);
+		}
+
+		results.push_back(current);
```
tigerw: Will `emplace_back` have any effect here?

worktycho: It wouldn't but an improvement would be to use std::move here to remove the unnessicary buffer copy.

flx5:
What about std::string::front instead of [0]? (simply to have the same syntax as in src/StringUtils.cpp Line 176.
Should I create another pull request for these enhancements?

tigerw:
If you want, though I wouldn't be surprised if the compiler identified it for optimisation anyway.

PR #1810, opened 2015-03-13
Paintings

tigerw:
* Implemented painting saving
* Streamlined hanging entity model
* Fixed reported issues about item frame orientation

### DIFF
--- a/src/Entities/HangingEntity.h
+++ b/src/Entities/HangingEntity.h
@@ -27,7 +27,7 @@ public:
 	eBlockFace GetFacing() const { return cHangingEntity::ProtocolFaceToBlockFace(m_Facing); }
 
 	/** Set the direction in which the entity is facing. */
-	void SetFacing(eBlockFace a_Facing)	{ m_Facing = cHangingEntity::BlockFaceToProtocolFace(a_Facing); }
+	void SetFacing(eBlockFace a_Facing) { m_Facing = cHangingEntity::BlockFaceToProtocolFace(a_Facing); }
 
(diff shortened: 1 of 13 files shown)


src/Entities/HangingEntity.h
```diff
@@ -24,28 +24,82 @@ class cHangingEntity :
...
+		{
+			case BLOCK_FACE_ZP: Dir = 0; break;
+			case BLOCK_FACE_ZM: Dir = 2; break;
+			case BLOCK_FACE_XM: Dir = 1; break;
+			case BLOCK_FACE_XP: Dir = 3; break;
+			default:
+			{
+				// Uncomment when entities are initialised with their real data, instead of dummy values:
+				// LOGINFO("Invalid facing (%d) in a cHangingEntity, adjusting to BLOCK_FACE_XP.", a_BlockFace);
+				// ASSERT(!"Tried to convert a bad facing!");
```
madmaxoft: What is this about, why can't we log / assert on invalid values?

tigerw: A BlockFace parameter is present in the constructor of all HangingEntities, which is converted through this function into protocol-wise values. However, when loading entities from NBT, WSSAnvil constructs it with dummy values (face  = -1, position = (0, 0, 0), etc.) and then sets it afterwards, thus causing mini-explosions if the lines were uncommented.

Since explosions are bad, I've begun to change it (https://github.com/mc-server/MCServer/tree/fixesandfeatures), but until then, //

PR #1811, opened 2015-03-14
Fixed grass spread, closes #1743

### LINKED ISSUE #1743: Grass doesn't spread.
Forum thread: http://forum.mc-server.org/showthread.php?tid=1782

tigerw:
- Removed the salt parameter in cFastRandom functions, it wasn't doing
what we thought it was following the move to C++11. Constant salts are not good for seeds.

### DIFF
--- a/src/Blocks/BlockDirt.h
+++ b/src/Blocks/BlockDirt.h
@@ -69,9 +69,9 @@ public:
 		cFastRandom rand;
 		for (int i = 0; i < 2; i++)  // Pick two blocks to grow to
 		{
-			int OfsX = rand.NextInt(3, a_RelX) - 1;  // [-1 .. 1]
-			int OfsY = rand.NextInt(5, a_RelY) - 3;  // [-3 .. 1]
-			int OfsZ = rand.NextInt(3, a_RelZ) - 1;  // [-1 .. 1]
+			int OfsX = rand.NextInt(3) - 1;  // [-1 .. 1]
+			int OfsY = rand.NextInt(5) - 3;  // [-3 .. 1]
+			int OfsZ = rand.NextInt(3) - 1;  // [-1 .. 1]
 	
 			BLOCKTYPE  DestBlock;
 			NIBBLETYPE DestMeta;
--- a/src/FastRandom.cpp
+++ b/src/FastRandom.cpp
@@ -6,6 +6,12 @@
 #include "Globals.h"
 #include "FastRandom.h"
 
+#ifdef _WIN32
+	#define thread_local __declspec(thread)
+#endif
+
+thread_local unsigned int m_Counter = 0;
+
 
 
 
@@ -86,7 +92,7 @@ public:
 
 
 cFastRandom::cFastRandom(void) :
-	m_LinearRand(static_cast<unsigned>(std::chrono::system_clock::now().time_since_epoch().count()))
+	m_LinearRand(m_Counter++)
 {
 }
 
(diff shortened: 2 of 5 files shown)


worktycho:
Bad change. The problem with this is that cFastRandom gets its seed from the system_clock time. There is no guarantee about the step on this clock. This is a minor problem if we call it multiple time in the same tick and the clock does not advance. It is a major problem if the step is greater than a tick.

worktycho:
Although If I'm correct this code has always been broken and we've only just noticed it because the c++11 port made it more broken.

worktycho:
I'm currently researching JVM memory allocation techniques to attempt to find a solution to the problem of seeds.

tigerw:
Eh, I thought the generator was seeded using the time at initialisation, after which it does magic/distributions using that seed?

worktycho:
And it being initialised every time OnUpdate is called.

tigerw:
Oh, I see. high_resolution_clock then?

worktycho:
High_resolution clock doesn't guarantee to tick quickly enough. I'm thinking we should use either a std::atomic<int> as a global counter or stick counters in thread_local storage.

tigerw:
Okay, for atomic, is `long long` a sufficiently large type?

worktycho:
32 bits is long enough so we should just use an long. The problem with atomic is that some of the implementations implement it using mutex's, which is exactly what we're trying to avoid with fast random. Thread_local is a possibility but I can't find enough information on its performance to be able to judge whether its a good idea.

worktycho:
One option I'm favouring at the moment is a method with selection based on types/Macros. For example: ATOMIC_LONG_LOCK_FREE which specifies whether the system supports lock free atomic's. Otherwise fall back the thread_local.

worktycho:
```
long getTrueRandomLong() {
    std::random_device rd();
    std::uniform_distribution<long> dist();
    return dist(rd);
}

#ifdef ATOMIC_LONG_LOCK_FREE && ATOMIC_LONG_LOCK_FREE == 1
cFastRandom MakeFastRandom()
{
    static std::atomic<long> seed{getTrueRandomLong()};
    return cFastRandom{seed++};
}
#else
cFastRandom MakeFastRandom()
{
    static thread_local long seed{getTrueRandomLong()};
    return cFastRandom{seed++};
}
#endif

tigerw:
That's quite complex. Are you sure that it's worth the trouble for a random generator that's not meant to be secure anyway?

Besides, if it is found to cause problems, it could always be worked around by storing a FastRandom instance, or using cWorld's.

worktycho:
cFastRandom is not supposed to be secure but it is supposed to be non-deterministic. It should not give the same sequence of numbers on two different ticks regularly.

Storing a cFastRandom is something that might work for the first two objects but for MobSpawner its still a problem as a new MobSpawner object is created every time the mobs are ticked. The problem with using cWorld's is that it is not thread safe.

madmaxoft:
Why not just use a simple int value, and increment it on usage. Even if we get a thread collision on this, we don't really mind, it's random, after all :)

worktycho:
The problem with a simple int is that on the C++ Virtual Machine defined by the standard the risks of a thread collision are potentially enormous. The classic issue is the ABA problem but it would be personally reasonable for an architecture to fault on a collision. Clang has a flag (-fsanitize=thread) to enable faulting on data races. This is one of those things where we may have to decide use a complex standard complaint solution or a simple solution that breaks the standard but runs on current architectures.

worktycho:
Having looked a bit further into the x86 ABI I think that thread_local is the way to go, TLS access should be as cheap as a array access on modern ABI's. I'll merge this if it compiles.

tigerw:
Okay, cFastRandom now uses transport layer security.

tigerw:
What happens when the int runs out though?

worktycho:
Thread local storage, there's only ~17000 TLA's so there are a few collisions. The overflow is defined to wrap as it's an unsigned int.

tigerw:
Nah, transport layer security sounds better :)

Okay, that makes sense. Merge?

PR #1813, opened 2015-03-16
Added console input indicator

SphinxC0re:
I'm open for a discussion whether it should be a '>' or a '$' or something else.
It's currently a $

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -70,10 +70,11 @@ cRoot::~cRoot()
 void cRoot::InputThread(cRoot & a_Params)
 {
 	cLogCommandOutputCallback Output;
-	
+	std::cout << std::endl;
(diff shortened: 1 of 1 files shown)


worktycho:
This will rapidly disappear under all of the console logs.

SphinxC0re:
Yeah it's only for input, you are right

madmaxoft:
I don't think it's worth it - whenever the server outputs a new log to the console, the indicator line stays and a new one is appended below. Try running a debug-mode server, then walking past the edge of generated chunks. The input indicator gets mixed with generator logs.

SphinxC0re:
That can all be fixed. The indicator is more about a good UX. I mention a point where the user can see that he can type something in that really does something would be pretty nice.

Issue #1715, opened 2015-01-25
Endianness.h uses implemention reserved identifiers


worktycho:
Endianness.h uses identifiers which begin with a double underscore. These identifiers are reserved for implementation use.

PR #1816, opened 2015-03-17
Bugfix 1715

bibo38:
To getting started, I fixed the Issue #1715 by replcaing some variable names.

### DIFF
--- a/src/Endianness.h
+++ b/src/Endianness.h
@@ -11,10 +11,10 @@
 // Changes endianness
 inline UInt64 HostToNetwork8(const void * a_Value)
 {
-	unsigned long long __HostToNetwork8;
-	memcpy( &__HostToNetwork8, a_Value, sizeof( __HostToNetwork8));
-	__HostToNetwork8 = (( ( (unsigned long long)htonl((u_long)__HostToNetwork8)) << 32) + htonl(__HostToNetwork8 >> 32));
-	return __HostToNetwork8;
+	unsigned long long buf;
+	memcpy( &buf, a_Value, sizeof( buf));
+	buf = (( ( (unsigned long long)htonl((u_long)buf)) << 32) + htonl(buf >> 32));
+	return buf;
 }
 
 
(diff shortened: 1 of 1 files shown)


src/Endianness.h
```diff
@@ -11,10 +11,10 @@
 // Changes endianness
 inline UInt64 HostToNetwork8(const void * a_Value)
 {
-	unsigned long long __HostToNetwork8;
-	memcpy( &__HostToNetwork8, a_Value, sizeof( __HostToNetwork8));
-	__HostToNetwork8 = (( ( (unsigned long long)htonl((u_long)__HostToNetwork8)) << 32) + htonl(__HostToNetwork8 >> 32));
-	return __HostToNetwork8;
+	unsigned long long buf;
```
worktycho: This should probably be changed to a UInt64.

bibo38:
Shouldn't the Globals.h also be included when using the UInt64 typedefs?

worktycho:
This project has a confusing convention of including Globals.h in the cpp files only. It has something to do with the MSVC build.

madmaxoft:
The reason for Globals.h is Precompiled headers - MSVC uses that file to generate the precompiled header, which speeds up the compilation considerably. But then it needs to be the first #include in each and every .cpp file and it has to be the exact same path in each .cpp file.

I wish we had something similar for gcc, I have too little knowledge to set that up, but I believe it would speed up the compilation quite a lot, too.

bibo38:
Maybe **[Cotire](https://github.com/sakra/cotire)** is the right way, which is mentioned in [this Stackoverflow question](http://stackoverflow.com/a/9838635): 

It should automaticly create Precompiled headers and a [Single Compilation Unit](http://en.wikipedia.org/wiki/Single_Compilation_Unit). Or you can use your own Precompiled Headers file.
All in all it is compatible with all common used compilers, so it should speed gcc and clang up.
I've never used it, but it seems to work really good.

PR #1818, opened 2015-03-18
Enabled output of individual tests.

madmaxoft:
This should give us an idea of why the google test fails randomly.

### DIFF
--- a/CIbuild.sh
+++ b/CIbuild.sh
@@ -11,7 +11,7 @@ cd src
 lua CheckBasicStyle.lua
 cd ..
 make -j 2;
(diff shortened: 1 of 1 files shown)


madmaxoft:
Is anyone against merging this?

It does produce slightly longer build logs, but on the other hand if something goes wrong, we'll have some details instead of just "failed".

NiLSPACE:
No problem here

bearbin:
Merge?

tigerw:
Yeah!

tigerw:
We're literally all owners here, but delegation!

Issue #1783, opened 2015-03-01
Flowing water to still water "endless Well"


Kellerkind85:
If you take a Waterblock out of a 1Block deep Sea, normally the other blocks around should fill up the missing water and become standing water.

Without this also the "endless Well" is not working.

bearbin:
What is the contents of your world.ini file?

Kellerkind85:
Oh.. ok perhaps its my bad. I am Sorry. But thanks for your help. here it is

-------------------------------
-------------------------------
[SpawnProtect]
ProtectRadius=10

[WorldLimit]
LimitRadius=0

[Difficulty]
WorldDifficulty=1

[General]
Dimension=Overworld
IsDaylightCycleEnabled=1
Gamemode=0
Weather=0
TimeInTicks=13403

[Broadcasting]
BroadcastDeathMessages=1
BroadcastAchievementMessages=1

[SpawnPosition]
MaxViewDistance=12
PregenerateDistance=20
X=-611.09742876027
Y=85
Z=1011.5583653016

[Storage]
Schema=Default
CompressionFactor=0

[Plants]
MaxCactusHeight=3
MaxSugarcaneHeight=3
IsCactusBonemealable=0
IsCarrotsBonemealable=1
IsCropsBonemealable=1
IsGrassBonemealable=1
IsMelonStemBonemealable=1
IsMelonBonemealable=0
IsPotatoesBonemealable=1
IsPumpkinStemBonemealable=1
IsPumpkinBonemealable=0
IsSaplingBonemealable=1
IsSugarcaneBonemealable=0

[Physics]
DeepSnow=1
ShouldLavaSpawnFire=1
TNTShrapnelLevel=2
WaterSimulator=Vanilla
LavaSimulator=Vanilla
SandInstantFall=0
RedstoneSimulator=Incremental

[Mechanics]
CommandBlocksEnabled=0
PVPEnabled=1
UseChatPrefixes=1

[Monsters]
VillagersShouldHarvestCrops=1
AnimalsOn=1
Types=bat, cavespider, chicken, cow, creeper, enderman, horse, mooshroom, ocelot, pig, sheep, silverfish, skeleton, slime, spider, squid, wolf, zombie

[LinkedWorlds]
NetherWorldName=world_nether
EndWorldName=world_end

[Generator]
Generator=Composable
BiomeGen=MultiStepMap
ShapeGen=DistortedHeightmap
CompositionGen=DistortedHeightmap
[...]

PR #1819, opened 2015-03-19
Fixed creation of Fluid-Source-Blocks

bibo38:
The else-Statement blocked the CheckNeighboursForSource function to get executed. This bricked the creation of new Fluid-Source-Blocks. Bugfix #1783 .


tigerw:
Subtle!

madmaxoft:
Yeah, exactly. I didn't catch this while reviewing the PR that fixed fluids spreading below `Y=0`, which is when this bug was introduced.

Issue #1765, opened 2015-02-24
Piston Block drops untextured and unstackable block when broken in survival.


Samvbnm:
Title says it all
Also if it helps;
MCServer was built on 23 Feb 2015.
OS: Linux 3.13.0-45-generic #74-Ubuntu 14.04 x86_64. 
heres some screenshots:
![2015-02-23_20 01 37](https://cloud.githubusercontent.com/assets/11139686/6340755/0d98decc-bb99-11e4-9658-442d94583a1f.png)
![2015-02-23_20 01 39](https://cloud.githubusercontent.com/assets/11139686/6340758/163c7926-bb99-11e4-80c5-64a78563cb77.png)
![2015-02-23_20 01 42](https://cloud.githubusercontent.com/assets/11139686/6340763/1d9cba0a-bb99-11e4-8cf1-d155561cf0ae.png)
![2015-02-23_20 01 49](https://cloud.githubusercontent.com/assets/11139686/6340767/23636ae2-bb99-11e4-9e09-d0b74615fcc3.png)

PR #1821, opened 2015-03-20
Removed Metavalue from Piston drops.

bibo38:
Overridden `ConvertToPickups` function to prevent creation of an item with metavalue. Bugfix #1765


madmaxoft:
Nice fix, thanks.

Usually we put the `Fixes #1765` part of the commit message into the "comments" section - after two newlines.

PR #1822, opened 2015-03-20
Added cPluginLua::cResettable interface, used for scheduled tasks.

### LINKED ISSUE #1556: A scheduled task can crash server on reload command
Hey,

I started an scheduler that runs 600 ticks later. If I do an reload, the server crashes directly (relase version) or in  a few seconds (debug version) with that error:

[7f1b379f5700|22:36:52] Assertion failed: IsValid(), file /home/lukas/cpp/MCServer/src/Bindings/LuaState.cpp, line 290
MCServer_debug: /home/lukas/cpp/MCServer/src/Bindings/LuaState.cpp:290: bool cLuaState::PushFunction(int): Assertion `0' failed.

Reproduce:
1) Create a scheduled code that will run for example 600 ticks later
2) Run the command /reload directly after step 1
3) Server should crash (can take a few seconds in debug version) 

Good luck by fixing.

regards,
Seadragon91

madmaxoft:
This allows plugins to register objects that can "survive" the plugin unloading - they will simply bail out if the plugin is already unloaded, instead of referencing bad plugin data.
Fixes #1556.

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -1299,23 +1299,27 @@ tolua_lerror:
 
 
 class cLuaWorldTask :
-	public cWorld::cTask
+	public cWorld::cTask,
+	public cPluginLua::cResettable
 {
 public:
 	cLuaWorldTask(cPluginLua & a_Plugin, int a_FnRef) :
-		m_Plugin(a_Plugin),
+		cPluginLua::cResettable(a_Plugin),
 		m_FnRef(a_FnRef)
 	{
 	}
 
 protected:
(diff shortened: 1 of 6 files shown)


madmaxoft:
Well, this is a major let-down: it doesn't work. The `cResettable`'s `m_CSPlugin` is shared with the `cPluginLua`, so once the `cPluginLua` instance goes away, the mutex is invalid and the scheduled task still fails.

Making it a separate mutex would create the possibility of deadlocks.

madmaxoft:
I might have fixed it, the deadlock with separate mutexes need not be possible.

PR #1823, opened 2015-03-20
Call hook HOOK_PLAYER_SPAWNED, if player has been moved to a other world...

### LINKED ISSUE #1435: HOOK_PLAYER_SPAWNED doesn't get called if player has been moved to world
Hey,

I am not sure if that is a bug.
There is a hook named HOOK_PLAYER_SPAWNED, that gets called if the player (re)spawns in the world where they are.

My thought is, if I move the player with the function Player:MoveToWorld("world"), then the hook HOOK_PLAYER_SPAWNED should be called, but that doesn't happen.

There is no other hook with that I can check if the player gets added to the world, for the function Player:MoveToWorld("world")

Thanks,
Seadragon91

Seadragon91:
This fixes #1435.

### DIFF
--- a/src/Blocks/BlockOre.h
+++ b/src/Blocks/BlockOre.h
@@ -50,6 +50,11 @@ public:
 				a_Pickups.push_back(cItem(E_ITEM_COAL));
 				break;
 			}
+			case E_BLOCK_NETHER_QUARTZ_ORE:
+			{
+				a_Pickups.push_back(cItem(E_ITEM_NETHER_QUARTZ));
+				break;
+			}
 			default:
 			{
 				a_Pickups.push_back(cItem(m_BlockType));


madmaxoft:
Doesn't this get called from elsewhere as well?

Seadragon91:
I tested it. The hook gets called if the player is authenticated or respawns, but not if he is moved to a other world.

Seadragon91:
Oh, should have taked a closer look at the api. The method `MoveToWorld` from cPlayer accepts, the `boolean ShouldSendRespawn`. Well then I think the issue and pull request, can be both closed.

madmaxoft:
Not necessarily - the respawn sending doesn't seem to trigger the hook.

Seadragon91:
Okay, forgot that pull request.

To topic.
1) The option `ShouldSendRespawn` from MoveToWorld, doesn't call the hook.
2) Whats about adding a hook, if the player has changed the world? Example name PLAYER_CHANGED_WORLD?

PR #1824, opened 2015-03-21
made underlying size variables constexpr

worktycho:
This makes the code standard compliant.
Previously we had got away with the non-compliant code because gcc and clang have an extension that allows the use of VLA's in c++ and visual studio has a bug/misfeature that allows const variables to be used in array declarations if they are initialized by a constant expression or by a const variable that is initialized in the same manner.


madmaxoft:
MSVC2013 doesn't support `constexpr` yet.

worktycho:
So we either need to require the Nov 2013 CTP or add the support for const vars in array initializes as a required extension.

worktycho:
Or remove the variables or use an enum value instead of a constvar.

madmaxoft:
I think I'd prefer the enum-based solution.

worktycho:
I've implemented the enum based solution.

madmaxoft:
I still don't understand what's wrong with the original code. I believe array bounds specified as const integers are perfectly valid C++ code.

worktycho:
Looks like const variables are valid so long as they are initialised with a constant expression. However the constexpr change would of clarified things if MSVC had supported it.

tigerw:
Can the branch be deleted?

PR #1826, opened 2015-03-22
Unify packetizer

madmaxoft:
Unified the two different `cPacketizer` classes inside the `cProtocol` descendants into a single shared class. The packetizer only writes the low-level datatypes, the high-level information (such as `cItem` data or mob metadata) is handled by the `cProtocol` descendants.

### DIFF
--- a/Tools/ProtoProxy/Connection.cpp
+++ b/Tools/ProtoProxy/Connection.cpp
@@ -214,7 +214,11 @@ cConnection::cConnection(SOCKET a_ClientSocket, cServer & a_Server) :
 	Printf(m_LogNameBase, "Logs/Log_%d_%d", (int)time(NULL), a_ClientSocket);
 	AString fnam(m_LogNameBase);
 	fnam.append(".log");
-	m_LogFile = fopen(fnam.c_str(), "w");
+	#ifdef _WIN32
+		fopen_s(&m_LogFile, fnam.c_str(), "w");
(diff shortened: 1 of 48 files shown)


NiLSPACE:
This could mean we could possible add custom packets for Lua. Something like
```lua
ClientHandle:SendCustomPacket(
   function(a_Packetizer, a_ProtocolVer)
      -- Packet Data
   end
)

SphinxC0re:
That would be a huge step in case of Forge mod compatibility

jan64:
Or, plugin-based protocol support.

madmaxoft:
Nah, I'm not willing to go there. And neither is Mojang - after all, we have the plugin channel packets.

NiLSPACE:
But it would allow plugins to use more then just the "broadcastXXX" functions.

jan64:
Of course, it could end up being a big mess, but i think it's worth it for the few creative uses that will most likely emerge.

jan64:
Also, adding a raw incoming packet hook would be nice.
But, that could cause performance issues.

madmaxoft:
Anyone against a merge?

NiLSPACE:
I don't have any objections.

tigerw:
Oh no, not the American 'z'!

:)

SphinxC0re:
I like the 'z' :D

PR #1843, opened 2015-03-30
A step towards implementing Vanilla gravity

DevToaster:
I created an "air drag" property using values from here (http://minecraft.gamepedia.com/Entity#Motion_of_entities) to affect the flight of entities in the air to behave more like vanilla.  Gravity is also tweaked for minecrafts, boats, items, and pawns (players, mobs).  I believe that the collision has to be redone though, as mobs falling at a high enough speed will be partially submerged in the ground.  This behavior is also viewable before these changes to an extent.  Also, sorry for the random 6x commit.

### DIFF
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -21,7 +21,7 @@ cArrowEntity::cArrowEntity(cEntity * a_Creator, double a_X, double a_Y, double a
 {
 	SetSpeed(a_Speed);
 	SetMass(0.1);
-	SetGravity(-20.f);
+	SetGravity(-20.0f);
 	SetAirDrag(0.2);
 	SetYawFromSpeed();
 	SetPitchFromSpeed();
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -21,8 +21,11 @@ cArrowEntity::cArrowEntity(cEntity * a_Creator, double a_X, double a_Y, double a
 {
 	SetSpeed(a_Speed);
 	SetMass(0.1);
+<<<<<<< HEAD
 	SetGravity(-20.0f);
 	SetAirDrag(0.2);
+=======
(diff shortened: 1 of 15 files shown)


src/Entities/Entity.h
```diff
@@ -504,6 +508,10 @@ class cEntity
 	For realistic effects, this should be negative. For spaaaaaaace, this can be zero or even positive */
 	float m_Gravity;
 	
+	/** Stores the air drag that is applied to the entity every tick, measured in speed ratio per second
+	Acts as air friction and slows down flight */
```
madmaxoft: `per second` will be quite interesting to implement in the actual gravity code. `per tick` might be a bit easier, although all the values would need recalculating.

madmaxoft:
Please don't use `1.f` format, use `1.0f` instead. The former is quite misleading for us "foreign" guys.

DevToaster:
Oh sorry about the "1.f" format.  I'll change it.  Per second right now seems to work very much like vanilla from what I've tested.

src/Entities/ArrowEntity.cpp
```diff
@@ -21,6 +21,11 @@ cArrowEntity::cArrowEntity(cEntity * a_Creator, double a_X, double a_Y, double a
 {
 	SetSpeed(a_Speed);
 	SetMass(0.1);
+<<<<<<< HEAD
+	SetGravity(-20.0f);
+	SetAirDrag(0.2);
+=======
+>>>>>>> parent of 5d9eb71... Gave entities more vanilla-like gravity and created 'air drag' property
```
worktycho: You left some merge markers in.

DevToaster: Ah yes, don't know how those even got there.  :(  That's why the build failed.  Time to spam more commits.

SphinxC0re:
Any updates on this one?

DevToaster:
Nope, this one is basically the same as the other one that just got merged.

Issue #1300, opened 2014-08-10
Player speed doesn't change


NiLSPACE:
Currently the speed of a player doesn't change (cPlayer:GetSpeed() returns a Vector3d(0, 0, 0)).
We should be able to set the speed somewhere [Here](https://github.com/mc-server/MCServer/blob/master/src/Entities/Player.cpp#L219) by using something like `Speed = NewPos - OldPos`

papatwika:
Hi, I just started working on the project and I had a question regarding the issue.

I tried to keep track of the position of the player (positon at the last tick or last second) to compare it with the current position, something like NiLSPACE proposed. The problem is that I recalculate the positions so often it makes the player laggy.

I was wondering if I can do the calculations somewhere else than cPlayer::Tick so it doesn't lag the player when he moves. Also, would it be a good practice to calculate the speed only when the method getSpeed() is called ?

madmaxoft:
Checking the speed in every tick shouldn't produce any visible lag at all, unless you're using some sort of a monster check. I think a more subtle thing happened - when you tried to set the calculated speed for the player, you called the `SetSpeed` method, which in turn sent a packet to the client informing them of their speed, thus moving the client.

papatwika:
You were right, the lag was caused by SetSpeed!

Does the speed value needs to be calculated by second ?

Does: Speed = distance / time
distance =  | oldPos - newPos |
time = sec?tick?

NiLSPACE:
Time is in ticks I believe.

madmaxoft:
The `cEntity::m_Speed` is in meters per second (in each direction) - it's in the doxy comments for it.

LO1ZB:
I try to fix this

PR #1844, opened 2015-03-30
Modified physics for more vanilla-like behavior

DevToaster:
I'm very sorry for the commit / pull request spam.  I was having trouble with Git not syncing correctly.  I fixed all of the previous issues and removed those mysterious merge tags and it now builds fine.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -36,6 +36,7 @@ cEntity::cEntity(eEntityType a_EntityType, double a_X, double a_Y, double a_Z, d
 	m_bHasSentNoSpeed(true),
 	m_bOnGround(false),
 	m_Gravity(-9.81f),
+	m_AirDrag(0.4f),
 	m_LastPos(a_X, a_Y, a_Z),
 	m_IsInitialized(false),
 	m_WorldTravellingFrom(nullptr),
@@ -943,6 +944,7 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		{
 			// Normal gravity
 			fallspeed = m_Gravity * DtSec.count();
+			NextSpeed -= NextSpeed * m_AirDrag * DtSec.count();
 		}
 		NextSpeed.y += static_cast<float>(fallspeed);
 	}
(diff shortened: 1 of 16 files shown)


src/Entities/Entity.cpp
```diff
@@ -943,6 +944,7 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		{
 			// Normal gravity
 			fallspeed = m_Gravity * DtSec.count();
+			NextSpeed -= NextSpeed * m_AirDrag * DtSec.count();
```
madmaxoft: This calculation means that `m_AirDrag` is measured neither in "ratio per second" nor "ratio per tick".

DevToaster: I see, it's not exactly m_AirDrag / sec when compounded.  I'll modify it so it's m_AirDrag per interpolated tick.

SphinxC0re:
any update on this one?

madmaxoft:
Code-wise, I think this is good.

SphinxC0re:
Well, then why not merge this?

madmaxoft:
Has anyone else actually tried this branch, if it doesn't break anything? I'm a bit cautious about any physics code lately.

DevToaster:
I've tested this myself.  It even fixes Blaze fireball projectiles slowing down and freezing midair in the overworld when shot upwards.

madmaxoft:
And it doesn't mess with player movement?

DevToaster:
Nope, not at all.
It seems that player movement is handled entirely client side anyways.  Mobs seem to have the same aerodynamics as players now.  The only thing I can think of is the "floaty" behavior of hobs when hit, but that's due to the raytrace/collision handling within the HandlePhysics function.

madmaxoft:
Actually, if you happen to send a SetSpeed packet to the client, it starts moving in that direction (see #1300). But if you say it doesn't happen, then it's safe to merge.

DevToaster:
Nothing involving SetSpeed has been changed from the master branch, just NextSpeed is subtracted by the air drag * NextSpeed.  It appears to work flawlessly.

Issue #1360, opened 2014-08-31
Allow server to be started as a Windows Service


madmaxoft:
Currently the executable can be run in the interactive session only, there's no way to run it as a service directly, one needs a wrapper program for that and there are problems with that (see http://forum.mc-server.org/showthread.php?tid=1560 ). We should add the support for running as a service when there's a special cmdline parameter added, and possibly another cmdline parameter to install the service, to make the setup easier.

birkett:
Is there still interest in improving Service support?

I have created a service wrapper as a drop in replacement for srvany, and use this to run an MCServer instance: https://github.com/birkett/srvany-ng

I have also been looking at adding native service support to MCServer - if it's something that would be accepted in, I will continue working on it.

madmaxoft:
It would be great to improve the windows service support. I'd love to have the main executable be able to be run as a service, and to have an extra pair of scripts for registering and unregistering the service (so that our users don't need to edit anything in regedit, as required by srvany).

I suggest using a new commandline parameter that would signalize to the executable that it should run as a service, and wrapping all of that functionality there.

birkett:
Ill see what I can do.

The actual code to get it running as a service is pretty short and simple, and the stdin issues don't seem to be a problem any more.

PR #1845, opened 2015-03-31
Enable MCServer to run as a service on Windows.

birkett:
This allows MCServer to run as a service on Windows, when started from the Service Control Manager with the "/service" command line switch.

### DIFF
--- a/src/BlockID.cpp
+++ b/src/BlockID.cpp
@@ -26,8 +26,18 @@ class cBlockIDMap
 	typedef std::map<AString, std::pair<short, short>, Comparator> ItemMap;
 	
 public:
+	static bool m_bHasRunInit;
+
 	cBlockIDMap(void)
 	{
+		// Dont load items.ini on construct, this will search the wrong path when running as a service.
+	}
+
+
+	void init()
+	{
+		m_bHasRunInit = true;
+
 		cIniFile Ini;
 		if (!Ini.ReadFile("items.ini"))
 		{
@@ -174,7 +184,7 @@ protected:
 
 
 
-
+bool cBlockIDMap::m_bHasRunInit = false;
 static cBlockIDMap gsBlockIDMap;
 
 
@@ -209,6 +219,10 @@ int BlockStringToType(const AString & a_BlockTypeString)
 		return res;
 	}
 	
+	if (!gsBlockIDMap.m_bHasRunInit)
+	{
+		gsBlockIDMap.init();
+	}
 	return gsBlockIDMap.Resolve(TrimString(a_BlockTypeString));
 }
 
@@ -222,6 +236,11 @@ bool StringToItem(const AString & a_ItemTypeString, cItem & a_Item)
 	{
 		ItemName = ItemName.substr(10);
 	}
+
+	if (!gsBlockIDMap.m_bHasRunInit)
+	{
+		gsBlockIDMap.init();
+	}
 	return gsBlockIDMap.ResolveItem(ItemName, a_Item);
 }
 
@@ -231,6 +250,10 @@ bool StringToItem(const AString & a_ItemTypeString, cItem & a_Item)
 
 AString ItemToString(const cItem & a_Item)
 {
+	if (!gsBlockIDMap.m_bHasRunInit)
+	{
+		gsBlockIDMap.init();
+	}
 	return gsBlockIDMap.Desolve(a_Item.m_ItemType, a_Item.m_ItemDamage);
 }
 
@@ -240,6 +263,10 @@ AString ItemToString(const cItem & a_Item)
 
 AString ItemTypeToString(short a_ItemType)
 {
+	if (!gsBlockIDMap.m_bHasRunInit)
+	{
+		gsBlockIDMap.init();
+	}
 	return gsBlockIDMap.Desolve(a_ItemType, -1);
 }
 
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -84,7 +84,10 @@ void cRoot::InputThread(cRoot & a_Params)
 	if (m_TerminateEventRaised || !std::cin.good())
 	{
 		// We have come here because the std::cin has received an EOF / a terminate signal has been sent, and the server is still running; stop the server:
-		a_Params.m_bStop = true;
+		if (m_RunAsService)  // HACK: Dont kill if running as a service
+		{
+			a_Params.m_bStop = true;
+		}
 	}
 }
 
@@ -265,6 +268,13 @@ void cRoot::Start(void)
 	delete fileLogListener;
 }
 
+
+
+
+void cRoot::SetStopping(bool a_Stopping)
+{
+	m_bStop = a_Stopping;
+}
 
 
 
--- a/src/Root.h
+++ b/src/Root.h
@@ -46,12 +46,16 @@ public:
 	// tolua_end
 
 	static bool m_TerminateEventRaised;
+	static bool m_RunAsService;
 
 
 	cRoot(void);
 	~cRoot();
 
 	void Start(void);
+
+	// Added so the service handler can request a stop
+	void SetStopping(bool a_Stopping);
 
 	// tolua_begin
 	cServer * GetServer(void) { return m_Server; }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -15,6 +15,8 @@
 
 
 
+/** Make the Root instance global, so it can be terminated from the worker threads */
+cRoot Root;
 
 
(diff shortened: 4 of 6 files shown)


birkett:
See issue #1360 for previous discussion.

src/main.cpp
```diff
@@ -179,6 +189,165 @@ BOOL CtrlHandler(DWORD fdwCtrlType)
...
+        serviceSetState(0, SERVICE_STOPPED, GetLastError());
+        return;
+    }
+
+    serviceSetState(SERVICE_ACCEPT_STOP, SERVICE_RUNNING, 0);
+
+    g_ServiceThread = CreateThread(NULL, 0, serviceWorkerThread, NULL, 0, NULL);
+    if (g_ServiceThread == NULL)
+    {
+        LOGERROR("CreateThread() failed\n");
```
worktycho: This won't log anywhere because the logger listeners aren't attached yet. I'm not sure where you can log to though.

worktycho:
Also this project uses tabs for indenting not spaces and that's what's causing the CI builds to fail.

birkett:
Thanks worktycho. Note to self, set tab width on a fresh VS install.

I've replaced those calls to LOGERROR with OutputDebugString. That will give some debug output to the debugger console if that section of code fails, and the debugger is attached.

src/BlockID.cpp
```diff
@@ -222,6 +236,11 @@ bool StringToItem(const AString & a_ItemTypeString, cItem & a_Item)
 	{
 		ItemName = ItemName.substr(10);
 	}
+
+	if (gsBlockIDMap.m_bHasRunInit != true)
```
NiLSPACE: You can just use `if (!gsBlockIDMap.m_bHasRunInit)`. It should be faster as well.

birkett: This is an artefact of some coding standards I normally use. 

For the sake of readability, I usually do the above, and let the compiler optimise it out. Is the rest of the codebase using the short form?

NiLSPACE: Yes, we use it everywhere else.

madmaxoft: Actually, never ever write `if (anything == true)` in C++, it may result in big surprises. `true` is a fixed constant, while the compiler is free to evaluate any true condition to **any** nonzero value, which will then compare as **not** equal to `true`.

birkett: I figured it was safe against that, being a bool. I've sorted it with 7fc9fbc.

madmaxoft:
This PR has become too overgrown with weed - unrelated changes in whitespace all over the place, some of them even changing stuff (such as in `cRoot::LogChunkStats`).

Could you rebase-and-edit the commits, squash them into a single one and edit out the changes to whitespace?

src/Root.cpp
```diff
@@ -268,6 +271,13 @@ void cRoot::Start(void)
 
 
 
+void cRoot::SetStopping(bool stopping)
```
madmaxoft: We use `a_ArgName` for function arguments.

src/Root.h
```diff
@@ -129,15 +133,15 @@ class cRoot
...
-	
+
 	// tolua_begin
-	
+
 	/// Sends a chat message to all connected clients (in all worlds)
-	void BroadcastChat       (const AString & a_Message, eMessageType a_ChatPrefix = mtCustom);
-	void BroadcastChat       (const cCompositeChat & a_Message);
+	void BroadcastChat	   (const AString & a_Message, eMessageType a_ChatPrefix = mtCustom);
+	void BroadcastChat	   (const cCompositeChat & a_Message);
```
madmaxoft: This should be failing the CheckBasicStyle.lua script.

src/Root.h
```diff
@@ -204,24 +208,20 @@ class cRoot
...
+
 	/// Does the actual work of executing a command
 	void DoExecuteConsoleCommand(const AString & a_Cmd);
-	
+
 	static cRoot* s_Root;
 
 	static void InputThread(cRoot & a_Params);
 };  // tolua_export
-
```
madmaxoft: Do keep the empty lines at the end of each file, please.

src/main.cpp
```diff
@@ -29,8 +30,16 @@ bool g_ShouldLogCommIn;
 /** If set to true, the protocols will log each player's outgoing (S->C) communication to a per-connection logfile */
 bool g_ShouldLogCommOut;
 
+/** If set to true, binary will attempt to run as a service on Windows */
+bool g_RunAsService = false;
+bool cRoot::m_RunningAsService = false;
```
madmaxoft: Why two variables for the same thing?

src/main.cpp
```diff
@@ -179,6 +188,168 @@ BOOL CtrlHandler(DWORD fdwCtrlType)
...
+// universalMain - Main startup logic for both standard running and as a service
+
+void universalMain()
+{
+#ifdef _WIN32
+	if (!SetConsoleCtrlHandler((PHANDLER_ROUTINE)CtrlHandler, TRUE))
+	{
+		LOGERROR("Could not install the Windows CTRL handler!");
+	}
+#endif
```
madmaxoft: We indent the preprocessor directives along the normal code.

src/main.cpp
```diff
@@ -179,6 +188,168 @@ BOOL CtrlHandler(DWORD fdwCtrlType)
...
+	switch (CtrlCode)
+	{
+	case SERVICE_CONTROL_STOP:
+		Root.SetStopping(true);
+		serviceSetState(0, SERVICE_STOP_PENDING, 0);
+		break;
+
+	default:
+		break;
+	}
```
madmaxoft: We use braces for the insides of each `case`, and indent the `case` labels one level more.

src/main.cpp
```diff
@@ -179,6 +188,168 @@ BOOL CtrlHandler(DWORD fdwCtrlType)
...
+
+void WINAPI serviceMain(DWORD argc, TCHAR *argv[])
+{
+#if defined(_DEBUG) && defined(DEBUG_SERVICE_STARTUP)
+	// Sleep on startup to all a debugger to manually attach.
+	Sleep(10000);
+#endif
+
+	char* applicationFilename = (char*)calloc(MAX_PATH, sizeof(char));
+	char* applicationDirectory = (char*)calloc(MAX_PATH, sizeof(char));
```
madmaxoft: Why alloc something of fixed max size? Simply use a stack-based array.

src/main.cpp
```diff
@@ -280,45 +444,38 @@ int main( int argc, char **argv)
...
-	try
-	#endif
-	{
-		cRoot Root;
-		Root.Start();
-	}
-	#if !defined(ANDROID_NDK)
-	catch (std::exception & e)
+
+#if defined(_WIN32) || defined(_WIN64)
```
madmaxoft: `_WIN32` is enough, it is defined for both 32-bit and 64-bit builds.

birkett:
Issues addressed, and squashed into one, much cleaner, commit.

src/main.cpp
```diff
@@ -179,6 +188,165 @@ BOOL CtrlHandler(DWORD fdwCtrlType)
...
+	serviceSetState(SERVICE_ACCEPT_STOP, SERVICE_RUNNING, 0);
+	
+	g_ServiceThread = CreateThread(NULL, 0, serviceWorkerThread, NULL, 0, NULL);
+	if (g_ServiceThread == NULL)
+	{
+		OutputDebugString("CreateThread() failed\n");
+		serviceSetState(0, SERVICE_STOPPED, GetLastError());
+		return;
+	}
+	WaitForSingleObject(g_ServiceThread, INFINITE);  // Wait here for a stop signal.
```
madmaxoft: Why create a separate thread just to wait for its termination anyway? Couldn't you just call `universalMain` from here instead?

birkett: Yes, that can be done. 

Usually with a service though, the worker thread is started to provide some monitoring of the process. Having a separate worker thread will allow cRoot to be redactored later - to add calls like beginLoop() and isQuitting() - which the service worker thread can act on instead of just waiting for the Root to return.

The second potential problem, is that service threads expect a void* parameter, used to pass data into the thread. This is a Windows specific thing, so I thought it best to leave universalMain completely cross-platform, instead of adding a parameter that would only be used on Windows.

If both of those are not of concern, I will re-write it.

madmaxoft:
Oops, this is bad, this PR is causing the server to crash on exit - the `cRoot` destructor calls into several other destructors and those use the logging framework, which is already unloaded by that time.

madmaxoft:
Should be fixed.

PR #1846, opened 2015-04-03
Fixed player switching world bug

### LINKED ISSUE #1842: cPlayer:MoveToWorld does not broadcast event
When I was testing out a few of my plugins with some friends over LAN a day ago, we couldn't see each other when we were both moved to the same world using the cPlayer:MoveToWorld function.  I narrowed it down to this by using /portal.  Along with that, we could not pick up items after being moved.

NiLSPACE:
Fixed players not appearing/disappearing when traveling to a different world.
Fixes #1842

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1603,6 +1603,9 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn)
 		m_ClientHandle->SendRespawn(a_World->GetDimension());
 	}
 
+	// Broadcast for other people that the player is gone.
+	GetWorld()->BroadcastDestroyEntity(*this);
+
 	// Remove player from the old world
 	SetWorldTravellingFrom(GetWorld());  // cChunk handles entity removal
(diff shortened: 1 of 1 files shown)


NiLSPACE:
Merge?

tigerw:
If it works, why not.

NiLSPACE:
It works for me. Maybe @DevToaster can confirm?

DevToaster:
I'll do a compile and test run and see what happens.

DevToaster:
Now I just noticed another problem.  :(  When a player dies and respawns, occasionally they may respawn as an invisible player.

madmaxoft:
Is this thread-safe? This code is rather fragile in this sense.

NiLSPACE:
@madmaxoft Should I try to move the `BroadcastSpawnEntity` function to the place where the player actualy is put into the world?

NiLSPACE:
So it was thread-safe?

madmaxoft:
I guess. No thread-related problems have been observed; you'd most likely hit the deadlock detector if there was a problem.

Issue #1806, opened 2015-03-10
Add /title command


sphinxc0re:
I only found the german wiki page for this (http://minecraft-de.gamepedia.com/Befehl/title) I hope you don't mind.
Also:
http://wiki.vg/Protocol#Title

madmaxoft:
I've seen this used in the Plots plugin on some servers, it is quite a useful feature - automatically listing the plot owners, or other useful information, without having to query.

This should be rather easy to add:
- add the packet sending function(s) in the protocol class (need to research about when the packet was added; implement as NOOP for protocols that don't support it)
- add the packet sending function(s) to a `cClientHandle` class 
- export the `cClientHandle` functions to Lua (should be possible to do this automatically)

I think it'd be better to implement the interface as multiple functions rather than one single one:
```cpp
void SendSetTitle(const cCompositeChat & a_Title, const cCompositeChat & a_SubTitle);
void SendSetTitle(const AString & a_Title, const AString & a_SubTitle);
void SendTitleTimes(int a_FadeInTicks, int a_DisplayTicks, int a_FadeOutTicks);
void SendHideTitle();
void SendResetTitle();
```

flx5:
It was added with 14w20a (http://wiki.vg/Protocol_History#14w20a) which equals protocol version 18 (http://wiki.vg/Protocol_version_numbers).

scottwillmoore:
I decided to attempt working on this feature as it appears fairly simple to implement and you have given me all the neccesary steps to implement it! I do not have much experience working with C++, however I am fairly experienced in Java and have a fairly solid understanding of the Minecraft internals as I have explored much of the Minecraft source code.

So far I have implemented the functions you suggested above in the base `cProtocol` class, then I have also overriden the functions in the implementing headers `cProtocol172` and `cProtocol180` and have now began implementing the functionality. As mentioned above it was implemented in a 1.8 snapshot and is therefore not avaliable in the 1.7 released, so I have implemented a NOOP for each of the new functions.

I have read through your source code and the implementation appears fairly simple I just have a few simple questions. 

Firstly when setting up the environment in the `GETTING-STARTED` guide it mentions to run `src/Bindings/AllToLua.bat`, however it appears I do not have the `tolua++` executable? I have followed the instructions so should I manually install the `tolua++` or is it supposed to build with the files. I ended up skipping this step for the time being.

[...]

PR #1848, opened 2015-04-07
Implement backend for /title command

scottwillmoore:
My current progress of the implementation of the /title command (#1806), also noticed while implementing it that the document is not in complete alphabetical order as the command `virtual void SendDisplayObjective` appears near the bottom.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2320,6 +2320,15 @@ void cClientHandle::SendHealth(void)
 
 
 
+void cClientHandle::SendHideTitle(void)
+{
+	m_Protocol->SendHideTitle();
+}
+
+
+
+
+
 void cClientHandle::SendInventorySlot(char a_WindowID, short a_SlotNum, const cItem & a_Item)
 {
 	m_Protocol->SendInventorySlot(a_WindowID, a_SlotNum, a_Item);
@@ -2514,6 +2523,15 @@ void cClientHandle::SendRemoveEntityEffect(const cEntity & a_Entity, int a_Effec
 
 
 
+void cClientHandle::SendResetTitle()
+{
+	m_Protocol->SendResetTitle();
+}
+
+
+
+
+
 void cClientHandle::SendRespawn(eDimension a_Dimension, bool a_ShouldIgnoreDimensionChecks)
 {
 	m_Protocol->SendRespawn(a_Dimension, a_ShouldIgnoreDimensionChecks);
@@ -2562,6 +2580,42 @@ void cClientHandle::SendScoreUpdate(const AString & a_Objective, const AString &
 void cClientHandle::SendDisplayObjective(const AString & a_Objective, cScoreboard::eDisplaySlot a_Display)
 {
 	m_Protocol->SendDisplayObjective(a_Objective, a_Display);
+}
+
+
+
+
+
+void cClientHandle::SendSetSubTitle(const cCompositeChat & a_SubTitle)
+{
+	m_Protocol->SendSetSubTitle(a_SubTitle);
+}
+
+
+
+
+
+void cClientHandle::SendSetRawSubTitle(const AString & a_SubTitle)
+{
+	m_Protocol->SendSetRawSubTitle(a_SubTitle);
+}
+
+
+
+
+
+void cClientHandle::SendSetTitle(const cCompositeChat & a_Title)
+{
+	m_Protocol->SendSetTitle(a_Title);
+}
+
+
+
+
+
+void cClientHandle::SendSetRawTitle(const AString & a_Title)
+{
+	m_Protocol->SendSetRawTitle(a_Title);
 }
 
 
@@ -2652,6 +2706,15 @@ void cClientHandle::SendTeleportEntity(const cEntity & a_Entity)
 void cClientHandle::SendThunderbolt(int a_BlockX, int a_BlockY, int a_BlockZ)
 {
 	m_Protocol->SendThunderbolt(a_BlockX, a_BlockY, a_BlockZ);
+}
+
+
+
+
+
+void cClientHandle::SendTitleTimes(int a_FadeInTicks, int a_DisplayTicks, int a_FadeOutTicks)
+{
+	m_Protocol->SendTitleTimes(a_FadeInTicks, a_DisplayTicks, a_FadeOutTicks);
 }
 
 
--- a/src/Protocol/Protocol.h
+++ b/src/Protocol/Protocol.h
@@ -86,6 +86,7 @@ public:
 	virtual void SendExplosion                  (double a_BlockX, double a_BlockY, double a_BlockZ, float a_Radius, const cVector3iArray & a_BlocksAffected, const Vector3d & a_PlayerMotion) = 0;
 	virtual void SendGameMode                   (eGameMode a_GameMode) = 0;
 	virtual void SendHealth                     (void) = 0;
+	virtual void SendHideTitle                  (void) = 0;
 	virtual void SendInventorySlot              (char a_WindowID, short a_SlotNum, const cItem & a_Item) = 0;
 	virtual void SendKeepAlive                  (int a_PingID) = 0;
 	virtual void SendLogin                      (const cPlayer & a_Player, const cWorld & a_World) = 0;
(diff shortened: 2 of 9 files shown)


scottwillmoore:
The integration tests have failed, so I will have a look at fixing those up tomorrow. Is this because I have used tabs for the indentation of my code? I thought in the CONTRIBUTING guide it mentions to use tabs for indentation? It also says I need two spaces in front of a style comment? `// comment >> //  comment`, however many of the comments I just copied and pasted from other sections of the source such as `// In game mode?`.

src/Protocol/Protocol.h
```diff
@@ -86,6 +86,7 @@ class cProtocol
 	virtual void SendExplosion                  (double a_BlockX, double a_BlockY, double a_BlockZ, float a_Radius, const cVector3iArray & a_BlocksAffected, const Vector3d & a_PlayerMotion) = 0;
 	virtual void SendGameMode                   (eGameMode a_GameMode) = 0;
 	virtual void SendHealth                     (void) = 0;
+	virtual void SendHideTitle					(void) = 0;
```
NiLSPACE: Only the at the start of a line you align the code using tabs. After that you use spaces.

src/Protocol/Protocol18x.cpp
```diff
@@ -542,6 +542,18 @@ void cProtocol180::SendHealth(void)
 
 
 
+void cProtocol180::SendHideTitle(void)
+{
+	ASSERT(m_State == 3); // In game mode?
```
NiLSPACE: You need to have 2 spaces after `//`. That's why the build is failing.

scottwillmoore:
Yeah just looking through some of it now, sorry about that I will fix it up tomorrow.

scottwillmoore:
Ah I see thanks, you can tell it is late because I thought CheckBasicStyle.lua was used to check lua styles! That will help heaps, thanks for that. Decided to fix it up tonight, enjoy. :)

src/Protocol/Protocol18x.cpp
```diff
@@ -1123,6 +1147,68 @@ void cProtocol180::SendDisplayObjective(const AString & a_Objective, cScoreboard
...
+
+void cProtocol180::SendSetSubTitle(const AString & a_SubTitle)
+{
+	ASSERT(m_State == 3);  // In game mode?
+
+	cPacketizer Pkt(*this, 0x45);  // Title packet
+	Pkt.WriteVarInt32(1);  // Set action as, set subtitle
+
+	Pkt.WriteString(Printf("{\"text\":\"%s\"}", EscapeString(a_SubTitle).c_str()));
+}
```
madmaxoft: We might want to rename this to `SendSetRawSubtitle()` and send the input string as-is, so that the `/title` command can send the entire string through here unparsed unmodified.

src/Protocol/Protocol18x.cpp
```diff
@@ -542,6 +542,18 @@ void cProtocol180::SendHealth(void)
 
 
 
+void cProtocol180::SendHideTitle(void)
+{
+	ASSERT(m_State == 3);  // In game mode?
+
+	cPacketizer Pkt(*this, 0x45);  // Title packet
+	Pkt.WriteVarInt32(3);  // Set action as, hide title
```
madmaxoft: The comment is very weird, why the comma in the middle?

scottwillmoore:
Hello, yep I agree with the name changes, they make the function easier to understand, and also you specified in the issue that it was `SendSetSubTitle(AString & a_SubTitle)` however now you are using `SendSetSubtitle(AString & a_Subtitle)`? Would you like me to change to using the second form? Subtitle is an actual word so it would make sense to display it the second way in the functions.

With the actual functionality of the `SendSetRawSubtitle(AString & a_Subtitle)` is there a reason you would want to send the string unformatted with json, as I am pretty sure if the text is not in the json format it will not work when it reaches the client. The `cCompositeChat` function allows you to send a fully formatted json text to the client and I just viewed the `AString` version as a shorthand to sending an unformatted, plain text title which gets wrapped in a basic json structure, whithout requiring the user to create a `cCompositeChat` item. I guess an advantage of not wrapping the string in the json structure would be that users could send their own json string through the function, rather than sending a `cCompositeChat` structure through the other function? I doubt many people would do this however as it is probably easier to create a `cCompositeChat` object.

Ah, yeah the comment is a bit weird, I think I put the comma there to separate the `hide title` which is the action specified by the protocol, from the other text. I can probably just remove the comma or replace it with a colon, which would make more grammatical sense.

tigerw:
About refactoring: http://stackoverflow.com/questions/363292/why-is-visual-c-lacking-refactor-functionality

madmaxoft:
If you're asking about the uppercase `T` vs lowercase `t` in `SubTitle`, I don't really care about those. I care more about the `Raw` being present.

The point is, how will we implement the `title <player> subtitle <raw json title>` command handler? Without the `Raw` variant of the call, it would have to parse the raw json title into a `cCompositeChat` (possibly losing things) and then the call would re-serialize that into a json string? That seems somewhat unnecessary.

If we wanted to have a simple interface for sending one single string, we could do that on the `cClientHandle` level (which gets exported to Lua API) and just make it call the `Raw` cProtocol function.

src/Protocol/Protocol18x.cpp
```diff
@@ -1123,6 +1147,68 @@ void cProtocol180::SendDisplayObjective(const AString & a_Objective, cScoreboard
...
+{
+	ASSERT(m_State == 3);  // In game mode?
+
+	cWorld * World = m_Client->GetPlayer()->GetWorld();
+	bool ShouldUseChatPrefixes = (World == nullptr) ? false : World->ShouldUseChatPrefixes();
+
+	cPacketizer Pkt(*this, 0x45);  // Title packet
+	Pkt.WriteVarInt32(1);  // Set subtitle
+
+	Pkt.WriteString(a_SubTitle.CreateJsonString(ShouldUseChatPrefixes));
```
madmaxoft: I think neither titles nor subtitles should ever have chat prefixes, or should they? @tigerw what do you think?
Also, how about just calling `SendSetRawSubTitle(a_SubTitle.CreateJsonString(false))`? That way the actual packet format is centralized to a single location and it's easier to clone+upgrade the protocol to a newer version.

src/Protocol/Protocol18x.cpp
```diff
@@ -1123,6 +1147,68 @@ void cProtocol180::SendDisplayObjective(const AString & a_Objective, cScoreboard
...
+{
+	ASSERT(m_State == 3);  // In game mode?
+
+	cWorld * World = m_Client->GetPlayer()->GetWorld();
+	bool ShouldUseChatPrefixes = (World == nullptr) ? false : World->ShouldUseChatPrefixes();
+
+	cPacketizer Pkt(*this, 0x45);  // Title packet
+	Pkt.WriteVarInt32(0);  // Set title
+
+	Pkt.WriteString(a_Title.CreateJsonString(ShouldUseChatPrefixes));
```
madmaxoft: Again, chat prefixes are questionable.

madmaxoft:
Sorry about the late reaction. Please add a comment to the PR when you think we should have a look, otherwise it doesn't pop up and we tend to forget :)

src/Protocol/Protocol18x.cpp
tigerw: I would agree that titles shouldn't be prefixed.

scottwillmoore:
Oh sorry about that. I wasn't aware that you wouldn't be notified if I didn't comment. Well I have fixed up the changes mentioned above and now `SendSetSubTitle(...)` and `SendSetTitle(...)` both use the raw implementation instead. Also removed chat prefixes.

madmaxoft:
This looks good now. Would you be willing to squash all the commits into one? We prefer not to have the "fix formatting" commits in the repo.

scottwillmoore:
Ok, you may have to check (I have never squashed commits into one before), I squashed them on my local repo which worked perfectly. I then went to push them to my github repo and it wouldn't let me. So I ended up using this command `git push origin +ImplementTitleCommand` which appears to have worked. On my github fork there is only one commit called `Implement backend for /title command` however in this pull request it still shows 4 commits in the commits tab above. Is this ok?

scottwillmoore:
Also we havn't tested to see if these functions will actually work? Is that what we will do now, so we can see if they work when sent to a client? Or does a lua wrapper for these functions need to be created so  the testing of these functions can be easily tested from a lua script.

madmaxoft:
That is correct, after squashing commits locally you need to force-push to the repo (because you're effectively changing the history). If multiple people are working on the repo, this gets dangerous, but right now it's only you so you're fine.

The GitHub page needs reloading, then it shows only the single commit.

In order to test the functions, you need to export them to the Lua API, then the easiest (and recommended) way to test is to write some test code in the Debuggers plugin. This will serve both as a test, and as example code.

To export a simple function to the API, you just need to add a `// tolua_export` comment after it in the header file. Namely, export the `cClientHandle` functions, as that's the proper interface to use them. When you build, you'll notice that ToLua is rebuilding the bindings as part of the build. Then you can use those functions from the plugins.

src/ClientHandle.cpp
```diff
@@ -2320,6 +2320,15 @@ void cClientHandle::SendHealth(void)
 
 
 
+void cClientHandle::SendHideTitle(void)
```
Howaner: Maybe remove this method and if SendSetSubtitle() is called with a empty string, hide the title.

Howaner: Oh, no. Its better to remove the SendResetTitle() and call it if SendSubtitle() is called with an empty string.

src/ClientHandle.cpp
```diff
@@ -2568,6 +2586,42 @@ void cClientHandle::SendDisplayObjective(const AString & a_Objective, cScoreboar
 
 
 
+void cClientHandle::SendSetSubTitle(const cCompositeChat & a_SubTitle)
```
Howaner: I don't like this name. Why not ```SendSubtitle(...)``|

src/ClientHandle.cpp
madmaxoft: There is a slight possibility that you'd want to display the subtitle, but hide the main title, that's why a separate method is still a good idea.

src/ClientHandle.cpp
madmaxoft: To be consistent with `SendSetTitle()` and `SendHideTitle()`

NiLSPACE: Would `SendShowTitle` not make more sense then?

madmaxoft: Not really, because the packet doesn't show the title, it only sets the title and you need to send TitleTimes to actually show the title.

scottwillmoore: That is actually incorrect. I managed to export the functions to the lua API and I was testing some of them the other day. It appears that `SendSetTitle(...)` and `SendSetSubTitle(...)` both when called (doesn't matter which one) will display the specifed type. `SendTitleTimes(...)` is just a preliminary step to displaying the title or subtitle. If both the title and the subtitle packet are sent at the same time they will both appear together.

bearbin:
@scottwillmoore Any update?

scottwillmoore:
Yeah, sorry for being so late, I have just been very busy lately and the laptop I was developing just recently broke and I have yet to replace.

The actual code appears to work and function how it should. When testing the functions in lua they all appeared to work and display the title correctly, however further testing should be conducted as I am just trying to recall this from memory.

At this stage the backend (C++) code is basically finished from my understanding, unless there are changes to the alogrithm or naming of the current functions that have been implemented. This could probably even be merged currently into the project at this stage, or into separate feature branch as it is functional code and works as expected.

A lua plugin still needs to be created which allows users to use commands to use this functionality ingame following the current syntax for the command as seen on the [wiki](http://minecraft.gamepedia.com/Commands#title).

Unfortunately I will not be currently able to work on the rest of the features until I have more time and my laptop has been repaired (hopefully under warranty!) Is anyone else keen on working on the lua plugin?

bearbin:
Well then I think this can be merged. @madmaxoft ?

madmaxoft:
Okay, merging.
@scottwillmoore could you please add the Lua API export as a separate PR? Thanks.

PR #1852, opened 2015-04-07
Fix slab placing

### LINKED ISSUE #1837: Slab placing
There are a few bugs when placing slabs.

If I place a slab beneath another slab which is on half block height two slabs were built. One to complete the Block and one more under the full block.

If I want to place a slab in the half block space between a slab on half block height with a full block under it and point the upper side of the block below, the slab will not appear.

And if I want to put a slab on the top of another slab no sound is played and this slaps sometimes disappear (maybe 1 of 20).

wraith11:
fix #1837

### DIFF
--- a/src/Items/ItemSlab.h
+++ b/src/Items/ItemSlab.h
@@ -39,10 +39,15 @@ public:
 		int a_CursorX, int a_CursorY, int a_CursorZ
 	) override
 	{
+		// Prepare sound effect
+		AString PlaceSound = cBlockInfo::GetPlaceSound(m_ItemType);
+		float Volume = 1.0f, Pitch = 0.8f;
+
 		// Special slab handling - placing a slab onto another slab produces a dblslab instead:
 		BLOCKTYPE ClickedBlockType;
 		NIBBLETYPE ClickedBlockMeta;
 		a_World.GetBlockTypeMeta(a_BlockX, a_BlockY, a_BlockZ, ClickedBlockType, ClickedBlockMeta);
+		// If clicked on a half slab directly
 		if (
 			(ClickedBlockType == m_ItemType) &&                         // Placing the same slab material
 			((ClickedBlockMeta & 0x07) == a_EquippedItem.m_ItemDamage)  // Placing the same slab sub-kind (and existing slab is single)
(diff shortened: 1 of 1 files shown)


src/Items/ItemSlab.h
```diff
@@ -63,11 +69,28 @@ class cItemSlabHandler:
 				((ClickedBlockMeta & 0x08) != 0)
 			)
 			{
+				a_World.BroadcastSoundEffect(PlaceSound, a_BlockX + 0.5, a_BlockY + 0.5, a_BlockZ + 0.5, Volume, Pitch);
 				return a_Player.PlaceBlock(a_BlockX, a_BlockY, a_BlockZ, m_DoubleSlabBlockType, ClickedBlockMeta & 0x07);
 			}
 		}
 
+		// Checking the type of block that should be placed
```
worktycho: Theres one extra space one the end of this line which is being flagged by the build server.

wraith11: A space in a comment, really? Think this is a bit strict.

worktycho: The rule used is very simple, white space followed by a new line because lua patterns can't tell if you're in a comment.

src/Items/ItemSlab.h
```diff
@@ -63,11 +69,28 @@ class cItemSlabHandler:
...
+		AddFaceDirection(a_BlockX, a_BlockY, a_BlockZ, a_BlockFace);
```
madmaxoft: I'd have preferred if you used a separate set of coords, `PlacedBlockX`, `PlacedBlockY` and `PlacedBlockZ`, and keep the values of `a_BlockX` et al.

src/Items/ItemSlab.h
```diff
@@ -63,11 +69,28 @@ class cItemSlabHandler:
...
+		BLOCKTYPE PlaceBlockType;
+		NIBBLETYPE PlaceBlockMeta;
+		a_World.GetBlockTypeMeta(a_BlockX, a_BlockY, a_BlockZ, PlaceBlockType, PlaceBlockMeta);
+		// If it's a slab combine into a doubleslab (means that clicked on side, top or bottom of a block adjacent to a half slab)
+		if (
+			(PlaceBlockType == m_ItemType) &&                         // Placing the same slab material
+			((PlaceBlockMeta & 0x07) == a_EquippedItem.m_ItemDamage)  // Placing the same slab sub-kind (and existing slab is single)
```
madmaxoft: This doesn't check the single-slabbed-ness at all. The `PlaceBlockType` check (above) does that, because double-slabs have a different block type

madmaxoft:
It'd be great if you could squash all of the commits into a single one, so that we don't have the "deleted whitespace" commits in the main history.

wraith11:
done

PR #1855, opened 2015-04-11
Issue 1300

papatwika:
Adresses this issue : https://github.com/mc-server/MCServer/issues/1300


madmaxoft:
Again, why aren't we doing the calculation on each tick? This way we present information that is almost always wrong. Was there a performance problem with a per-tick calculation? I have a feeling that the `chrono` methods you use eat up way more CPU than the simple calculation would have.

Also, your solution hides the `GetSpeed()` and `SetSpeed()` functions from `cEntity`, which is rather unfortunate, especially when it comes to plugins - they expect consistent behavior.

Not to mention that your calculation is wrong, you're doing `abs()` on the difference in positions (and an integral one), thus mis-representing the speed vector.

src/Entities/Player.cpp
```diff
@@ -209,6 +213,15 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		}
 	}
 
+	Vector3d CurrPos = GetPosition();
+	std::chrono::steady_clock::duration timeDiff = std::chrono::steady_clock::now() - m_LastTickTime;
```
worktycho: If you're going to calculate the value every tick then the a_Dt argument to this method will be equal to timeDiff, but with millisecond precision.

src/Entities/Player.cpp
```diff
@@ -209,6 +213,15 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+	double timeRatio = 1 / (double(timeDiff.count()) * std::chrono::steady_clock::period::num / std::chrono::steady_clock::period::den);
```
worktycho: I really don't like this code as one of the purpose's of std::chrono is to handle this for you. I suspect the best solution is to convert the following code to work with the millisecond count directly.

src/Entities/Player.cpp
```diff
@@ -209,6 +213,15 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+
+	SetSpeedValue(new Vector3d((m_LastPosition.x - CurrPos.x)*timeRatio, (m_LastPosition.y - CurrPos.y)*timeRatio, (m_LastPosition.z - CurrPos.z)*timeRatio));
```
worktycho: This is a memory leak. The new is constructing a pointer which is then copy constructed into a value type. @madmaxoft Is there any reason we can't delete the `Vector3(const Vector3<_T> * a_Rhs)` constructor? It is pretty dangerous.

Also this code looks generally weird. The Vector3 family of types provides arithmetic methods. Is there any reason to do this rather than `(m_LastPosition - CurrPos) * timeRatio`?

src/Entities/Player.cpp
```diff
@@ -95,7 +95,11 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 	SetMaxHealth(MAX_HEALTH);
 	m_Health = MAX_HEALTH;
 	
+	m_LastPosition = new Vector3d(0.0, 0.0, 0.0);
```
worktycho: This leaks memory.

worktycho:
@papatwika Any updates on this?

SphinxC0re:
This is totally outdated. Closing for now

PR #1860, opened 2015-04-14
Removed cBlockPluginInterface in favor of unified cPluginInterface.

madmaxoft:
This allows the BlockHandlers to call all plugins defined by cPluginManager, and simplifies the class hierarchy because the cPluginManager implements the cPluginInterface directly.


madmaxoft:
@worktycho I'd like your review of this.

worktycho:
I think you might as well remove the entire interface. I added that when I was trying to decouple the code for testing so that that I could mock the plugin interface without sucking cWorld in. With you're interface definition there are so many concrete types in the method parameters that this interface won't help with that. I've pretty much given up on unit testing so you can remove the interface if its getting in the way, but don't unify it. That was part of the point of it.

madmaxoft:
Okay, I'll remove all three interfaces, they are quite getting in the way and there's no added benefit to them anyway.

PR #1864, opened 2015-04-16
Recipes match zero DamageValue strictly.

### LINKED ISSUE #1859: Crafted special block variants revert to plain or other versions
Dyed carpets revert to their default colors right after being crafted.

Stained baked clay cannot be crafted. The desired block shows up in the inventory, but seems non-existent.

Polished diorite and polished andesite become plain stone bricks when trying to craft them.

Trying to create chiseled stone made of two slabs doesn't work. The new block appears for a split second; then vanishes.

madmaxoft:
To match any DamageValue for the ingredient, an explicit "-1" as the DamageValue must be specified.
Fixes #1859.

### DIFF
--- a/MCServer/crafting.txt
+++ b/MCServer/crafting.txt
@@ -14,7 +14,10 @@
 #
 # <ItemType> can be either a number, or an item name (checked against items.ini)
 #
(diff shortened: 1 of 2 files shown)


madmaxoft:
I *think* this is ready for a merge, but I'd prefer if someone else checked, too.

PR #1869, opened 2015-04-17
Added MobTypeName for Giant

mathias-gh:
Added missing MobTypeName for Giant. This allows plugins like Essentials to spawn Giants.


NiLSPACE:
Nice catch :)

PR #1870, opened 2015-04-17
Fixed IronGolem sounds

mathias-gh:
I changed the sound names for IronGolems to the correct ones, to make IronGolem sounds work in-game.

### DIFF
--- a/src/Mobs/IronGolem.cpp
+++ b/src/Mobs/IronGolem.cpp
@@ -8,7 +8,7 @@
 
 
 cIronGolem::cIronGolem(void) :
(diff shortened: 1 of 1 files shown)


madmaxoft:
Nice, thank you!

Who would've thought the sound names are case-sensitive

jammet:
Could this help make the sounds of sheep, cows and all the other animals work? The world is a pretty quiet world at the moment.

mathias-gh:
Only hurt and death sounds are played for mobs right now, afaik, so this method won't help much. However, MagmaCubes have some invalid sounds right now, which this method could fix.

Guardians have a similar problem. A sound called "minecraft:damage.thorns" plays when Guardians get hurt, but this sound doesn't exist. I don't have any idea why it plays, since the code says that "mob.guardian.idle" should play.

PR #1873, opened 2015-04-18
Use variadic functions for hook calling

tigerw:
* Better readability
* Simplified cPluginLua inheritance


src/Bindings/PluginLua.h
```diff
@@ -95,89 +98,43 @@ class cPluginLua :
...
 	bool CanAddOldStyleHook(int a_HookType);
 	
-	// cWebPlugin override
-	virtual const AString GetWebTitle(void) const {return GetName(); }
+	// cWebPlugin
+	const AString GetWebTitle(void) const {return GetName(); }
 
 	// cWebPlugin and WebAdmin stuff
-	virtual AString HandleWebRequest(const HTTPRequest * a_Request) override;
+	AString HandleWebRequest(const HTTPRequest * a_Request);
```
madmaxoft: Why remove the `virtual` and `override` specifiers?

madmaxoft:
Instead of merging you should have rebased.

Why are you removing the `cPlugin` class?

I don't like the fact that there are two hooks that need different calls than other hooks.

tigerw:
About cPlugin, since Squirrel isn't used anymore, what function does it serve?

madmaxoft:
It's in the Lua API, removing it will break existing plugins.

madmaxoft:
This has one more unfortunate side-effect: There's no type-checking for the parameters to `CallHook`. You could practically do `CallHook(HOOK_LOGIN, SomeWorld, 10, "some string", cItem(E_ITEM_TNT));` and the compiler won't complain at all. This is, in my opinion, a serious problem.

tigerw:
There could be some manual checking if hooks are changed in a pull request, which shouldn't happen often.

If hooks are added, they'll be cross examined with their documentation (anyway).

tigerw:
Since cPlugin and all of its functions was merged into cPluginLua, nothing should break.

worktycho:
Its still potentially a breaking change as it will cause any code that checks the type name to break.

worktycho:
I agree with xoft. This removes all type checking on calling hooks so should not be accepted. However I also agree with tiger that the current list of hooks has some serious issues with redundancy. This may be a place for either macros or a code generation script.

PR #1874, opened 2015-04-18
Simplified inheritance of cPluginLua

tigerw:
Nothing breaks!


madmaxoft:
Why are you so intent on this? Did you decide that you just need to create a merge conflict for me? I say I'll work on the plugin manager, and suddenly everyone starts messing with it, rather needlessly, too.

madmaxoft:
This makes it seem that plugins are allowed to add webadmin tabs to other plugins (`AddWebTab` is available after using `cPluginManager:GetPlugin()` )

tigerw:
My apologies; I would be happy for you to merge your changes and then resolve the conflicts here, if you want.

Since LoadPlugin and GetPlugin are both marked deprecated and unsafe, should I remove them?

madmaxoft:
Those two have already been dealt with in my PR. Have a look there if there's anything you disagree with; if not, we can merge it and then you can have a go at this (although I still don't think this is a good idea)

tigerw:
Tada!

src/Bindings/PluginManager.cpp
```diff
@@ -78,9 +77,9 @@ void cPluginManager::RefreshPluginList(void)
 	// Set all plugins with invalid folders as psNotFound:
 	for (auto & plugin: m_Plugins)
 	{
-		if (std::find(Folders.cbegin(), Folders.cend(), plugin->GetFolderName()) == Folders.end())
+		if (std::find(Folders.cbegin(), Folders.cend(), plugin->GetLocalFolder()) == Folders.end())
```
madmaxoft: Wrong, `GetLocalFolder` returns the path including the "Plugins/", `Folders` never contains the "Plugins/" prefix.
You need to call `GetFolderName`

src/Generating/ChunkGenerator.h
```diff
@@ -177,7 +177,7 @@ class cChunkGenerator :
 	cGenerator * m_Generator;
 	
 	/** The plugin interface that may modify the generated chunks */
-	cPluginInterface * m_PluginInterface;
+	cPluginLuaInterface * m_PluginInterface;
```
madmaxoft: I don't think this file needed any of the changes.

tigerw:
Tada!

madmaxoft:
I still don't like this change. It forever breaks the opportunity to have different language plugins, and it exposes some functions to the API that shouldn't be exposed. Is it really worth the few lines you have saved?

tigerw:
There have been requests to add new scripting languages recently, but those have been turned down for one reason or another, and surely the exposed functions can be unexported again.

tigerw:
Merge?

madmaxoft:
I'm gonna say this one more time: I don't like this change and I don't see any benefit in making it.
If there are others who support this change, okay; if not, then I vote DON'T MERGE at all.

NiLSPACE:
I think we need some more voices here, since it makes it really hard to add more scripting languages. @bearbin @worktycho @Seadragon91, @SafwatHalaby @SphinxC0re?

As for me, I don't really like the fact that it would get harder to implement new languages, but on the other hand thus far we rejected every other language. I'm a little 50/50.

madmaxoft:
It's not only about the other languages, this also changes the API in a bit. Suddenly each plugin is capable of calling other plugin's API methods, which hasn't been possible before. Think `OtherPlugin:AddWebTab()` (this one is actually explicitly protected in the manual bindings, but you get the point)

bearbin:
I don't think that I really have the knowledge to say whether this is a good idea. But I think it's quite unlikely that we'll ever have another scripting language, and I don't think there's much of a problem with plugins calling other plugins' API methods - only the plugins that really need it will do it.

SafwatHalaby:
Like @bearbin , I still don't have enough knowledge on this subject to give a worthy opinion. Though I don't imagine we'll have any other scripting languages.

Currently my understanding of MCServer is very narrowly focused, I only learned what I need to learn to get the Pathfinder going, so my knowledge is limited to cChunk and cEntity and its descendants. I'll study the project thoroughly when I have time.

tigerw:
No one's sure. Benefit of the doubt, and revert if problems in the future?

worktycho:
I've got another problem. By allowing plugins to call other plugins API methods you break thread safety. The methods are written to assume that the Plugin Lock is held, but under this code, they could be called from other plugins in which case the lock would not be held.

tigerw:
Guess not then, dam threading. Dam locks. Dam puns.

PR #1875, opened 2015-04-19
Split the plugin names and plugin folders to avoid confusing them.

madmaxoft:
Ref.: http://forum.mc-server.org/showthread.php?tid=1877

Note that this needs changes in the Core, implemented in the PluginFolders branch there.


madmaxoft:
Changes in the Core plugin published as PR mc-server/Core#131

src/Bindings/ManualBindings.cpp
```diff
@@ -1493,6 +1482,18 @@ static int tolua_cPluginManager_GetCurrentPlugin(lua_State * S)
 
 
 
+static int tolua_cPluginManager_GetPlugin(lua_State * tolua_S)
```
NiLSPACE: A little unrelated, but maybe a `cPluginManager::PluginExists(a_PluginName)` could be usefull. Previously `cPluginManager::GetPlugin` could be used for that.

madmaxoft: I suppose we need different levels of "exists" - "exists and is loaded", "exists but don't care about status"

src/Bindings/Plugin.cpp
```diff
@@ -28,9 +28,33 @@ cPlugin::~cPlugin()
 
 
 
+void cPlugin::Unload(void)
+{
+	auto pm = cPluginManager::Get();
+	pm->RemovePluginCommands(this);
+	pm->RemovePluginConsoleCommands(this);
+	pm->RemoveHooks(this);
```
NiLSPACE: Are the webpages removed somewhere else?

madmaxoft: You're right, they aren't being removed. This doesn't make a problem when reloading, since the tab names are overwritten, but it does make a problem when unloading a plugin via the `unload` console command and then going through webadmin.

NiLSPACE:
Looks great :+1: :)

madmaxoft:
Webpages are now properly cleared when a plugin is unloaded. Also took the liberty to convert the `cWebPlugin` to C++11 style and added proper locking.

madmaxoft:
Added the `IsPluginLoaded()` API function.

I believe this is now merge-worthy.

NiLSPACE:
I think it is. Merge if compile?

NiLSPACE:
I think this has broken the front page of the webadmin. It no longer shows any plugins.

madmaxoft:
I works alright for me. Do you get any error in the console?

NiLSPACE:
I haven't compiled the server for a while, but the gallery server doesn't show anything anymore
Img

madmaxoft:
Ah, that would be because I forgot to update the `template.lua` file in Gallery's `webadmin` folder.

madmaxoft:
Fixed

NiLSPACE:
Great :)

PR #1879, opened 2015-04-20
Changed Nether composition to change the threshold from a cubic noise

NiLSPACE:
This will case some dense and open areas in the nether.

### DIFF
--- a/src/Generating/CompoGen.cpp
+++ b/src/Generating/CompoGen.cpp
@@ -218,7 +218,7 @@ void cCompoGenClassic::InitializeCompoGen(cIniFile & a_IniFile)
 cCompoGenNether::cCompoGenNether(int a_Seed) :
 	m_Noise1(a_Seed + 10),
 	m_Noise2(a_Seed * a_Seed * 10 + a_Seed * 1000 + 6000),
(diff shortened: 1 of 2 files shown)


madmaxoft:
I'd expect this to create somewhat visible boundaries between chunks, since the threshold is applied per-chunk. Is that so?

NiLSPACE:
I didn't notice any since a difference of 100 threshold is hardly noticeable, but I'll do some further testing.

I did consider calculating the threshold for each column in the chunk, but I thought I Could save some performance by doing it this way.

NiLSPACE:
After searching I found a few chunk borders, but they were really small. It did look a big awkward in some cases, so I guess I'll calculate the threshold for each column.

NiLSPACE:
I'm now searching a better default value for `m_MaxThreshold`.  10000 doesn't seem to make big differences in the terrain.

NiLSPACE:
I think this is done now

madmaxoft:
I'll try this out

madmaxoft:
Looks good enough. A bit too dense for my taste, but that can probably be tweaked further.

PR #1880, opened 2015-04-20
Update submodules

tigerw:
Testing, just in case something goes awry.


madmaxoft:
There are also PolarSSL targets that are explicitly disabled in the cmakefiles, are they still disabled in the new versions? We don't want to build PolarSSL test suite.

tigerw:
The tests remain disabled, I believe.

madmaxoft:
This seems to have broken my RasPi builds, for some reason mbedtls won't compile and then MCS fails to link. I'll investigate further.

madmaxoft:
False alert, I forgot to run `git submodule update` in the build script.

madmaxoft:
For some reason on Windows CMake was complaining while configuring embedtls that there is a duplicate library embedtls being declared. I had to remove all cmake-generated files from my MCS build and re-run cmake from scratch.

madmaxoft:
And it still wouldn't work

tigerw:
Ah; does it work now?

madmaxoft:
Works for me now, both on RasPi and Windows.

PR #1883, opened 2015-04-21
Narrowed WindowID to use only 7 bits.

madmaxoft:
This fixes window issues reported on the forum, players are unable to interact with any windows after a while.

Ref.: http://forum.mc-server.org/showthread.php?tid=1876


worktycho:
Merge if complies.

PR #1896, opened 2015-04-24
Eliminate race condition in cRoot::FindAndDoWithPlayer

worktycho:
Avoid a race condition where a player gets deleted during the FindAndDoWithPlayer callback by moving the callback into DoWithPlayer.

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -607,7 +607,7 @@ bool cRoot::FindAndDoWithPlayer(const AString & a_PlayerName, cPlayerListCallbac
 			size_t Rating = RateCompareString (m_PlayerName, a_pPlayer->GetName());
 			if ((Rating > 0) && (Rating >= m_BestRating))
 			{
-				m_BestMatch = a_pPlayer;
+				m_BestMatch = a_pPlayer->GetName();
 				if (Rating > m_BestRating)
 				{
 					m_NumMatches = 0;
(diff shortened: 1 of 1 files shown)


src/Root.cpp
```diff
@@ -607,7 +607,7 @@ bool cRoot::FindAndDoWithPlayer(const AString & a_PlayerName, cPlayerListCallbac
 			size_t Rating = RateCompareString (m_PlayerName, a_pPlayer->GetName());
 			if ((Rating > 0) && (Rating >= m_BestRating))
 			{
-				m_BestMatch = a_pPlayer;
+				m_BestMatch = a_pPlayer->GetName();
```
madmaxoft: You should probably store the EntityID instead - for when there are two players of the same name (happens a lot when debugging).

worktycho: If there are two players with the same name it will fail the if condition on line 639 and the callback just won't be called. This might be a bug in the existing implementation as the documentation implies that the callback is called on all players in that scenario, where as the current and proposed implementations will just fail.

worktycho: Unless both player have the exact name that is specified in which case the server will just pick one of the players. And it doesn't matter which one, so the fact that this code moves the choice into DoWithPlayer, rather than FindAndDoWithPlayer is irrelevant.

The only scenario in which using the name could fail is if a player name changes whilst the server is running without logging out and back in again, something which I think is not going to happen.

worktycho: Also using the ID results in the rather unintuitive situation where if two players with the same name are connected and one leaves whilst FindAndDoWithPlayer is executing, the call fails as despite there never being a point in time during the execution where there are no players with that name connected.

The linerization of the operations that results in that condition is:

1. Two players with name "Foo"  and IDs 1 and 2 connect
2. Start FindAndDoWithPlayer("Foo")
3. ForEachPlayer() locates player with ID 1.
4. Player with ID 1 disconnects.
5. DoWIthEntityByID Fails as no entity with ID 1.
6. FindAndDoWithPlayer("Foo") fails.

When the code uses the name it automatically fails over the the player with ID 2.

worktycho: Do you still think it should keep the entityID rather than name?

madmaxoft: I guess it's good enough as it is, this API is still pretty useless in my opinion, so what that it's half broken :P

worktycho: Its only real use is command matching which core uses it a lot for. For example the kick command which is how I found this.

PR #1907, opened 2015-04-27
Use libc++ on all OS X compliers.

worktycho:


madmaxoft:
Why close?

worktycho:
Unnecessary, given that we've obsoleted gcc<4.8, and apple haven't ported higher gcc versions.

PR #1911, opened 2015-04-27
Cxx11upgrade

### LINKED ISSUE #1910: Use standard library <cstdint> types rather than self defined integer types
No need to define your own integer types when the standard library already does that. 
(Integer typedefs in Globals.h)
http://www.cplusplus.com/reference/cstdint/

4c370798d52b1b7c055c49ff4a38f73cf720c82e
b1d4b3bb96629b3624e8328d7b1a0bce5333bb7d

worktycho:
A couple of c++11 migration fixes.
Fixes #1910.

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -124,41 +124,22 @@
 
 
 #include <stddef.h>
+#include <stdint.h>
 
 
 // Integral types with predefined sizes:
-typedef signed long long Int64;
-typedef signed int       Int32;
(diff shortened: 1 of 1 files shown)


tigerw:
Why close?

worktycho:
Its dependent on https://github.com/SRombauts/SQLiteCpp/issues/34. Once that's fixed and we've updated the a fixed version this can be reopened.

PR #1916, opened 2015-04-28
A* Pathfinding

wiseoldman95:
Hurray :)
Please see http://forum.mc-server.org/showthread.php?tid=1571&pid=19781#pid19781


src/Mobs/CMakeLists.txt
```diff
@@ -62,6 +63,7 @@ SET (HDRS
 	Ocelot.h
 	PassiveAggressiveMonster.h
 	PassiveMonster.h
+        Path.h
```
tigerw: Arbitrary tabulation is arbitrary?

worktycho: Its just github's bad formatting of diffs with leading tabs.

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,56 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
+	if (m_Path == nullptr)
 	{
-		// Too low/high, can't really do anything
+		/*printf("%d %d %d > %d %d %d\n", floor(GetPosition().x), floor(GetPosition().y), floor(GetPosition().z), floor(m_FinalDestination.x), floor(m_FinalDestination.y), floor(m_FinalDestination.z));*/
+		Vector3d position=GetPosition();
+		Vector3d Dest = m_FinalDestination;
+		
+		// Can someone explain why are these two NOT THE SAME???
+		// m_Path = new cPath(GetWorld(), GetPosition(), m_FinalDestination, 30);
+		m_Path = new cPath(GetWorld(), Vector3d(floor(position.x), floor(position.y), floor(position.z)), Vector3d(floor(Dest.x), floor(Dest.y), floor(Dest.z)), 90);
```
tigerw: I think GetPosition() etc. can be replaced with POS_TOINT.

You can then do Dest.Floor() to floor its values using the proper method (FloorC).

(I'm not sure, but it seems like the commented-out bit should be the same? Maybe something to do with copy vs reference of vector object.)

src/Mobs/Monster.h
```diff
@@ -162,6 +162,10 @@ class cMonster :
 
 	/** A pointer to the entity this mobile is aiming to reach */
 	cEntity * m_Target;
+	cPath * m_Path;
+	ePathFinderStatus m_PathStatus;
+	bool m_IsFollowingPath;
+	int m_GiveUpCounter;  // If 0, will give up reaching the next m_Dest and will re-compute path.
```
tigerw: Probably best to do the comment:

```
/** If 0 ... */
int m_GiveUp...;
```

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,386 @@
...
+	#endif
+	return Ret;
+}
+
+
+
+
+
+// Add the next point in the final path.
+void cPath::addPoint(Vector3d a_Vector)
```
tigerw: Perhaps consider capitalisation to AddPoint?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,386 @@
...
+				}
+			}
+		}
+	}
+	return m_Status;
+}
+
+
+
+void cPath::processIfWalkable(const Vector3d & a_Location, cPathCell * a_Parent, int a_Cost)
```
tigerw: Perhaps consider capitalisation to ProcessIfWalkable?

src/Mobs/CMakeLists.txt
tigerw: My apologies.

src/Mobs/Path.h
```diff
@@ -0,0 +1,168 @@
...
+#ifdef COMPILING_PATHFIND_DEBUGGER
+/* Note: the COMPILING_PATHFIND_DEBUGGER flag is used by Native/WiseOldMan95 to debug this class outside of MCServer.
+This preprocessor flag is never set when compiling MCServer. */
+#include "PathFinderIrrlicht_Head.h"
+#endif
+
+#include <vector>
+#include <queue>
+#include <unordered_map>
+using namespace std;
```
worktycho: This is problematic because its importing std::make_unique into the global namespace on systems with c++14 support which conflicts with our polyfill method in globals.h

src/Mobs/Path.h
```diff
@@ -0,0 +1,168 @@
...
+// fwd: Vector3.h
+template <typename T> class Vector3;
+typedef Vector3<double> Vector3d;
+
+// fwd: cChunkMap.h
+typedef cItemCallback<cChunk> cChunkCallback;
+#endif
+
+/* Various little structs and classes */
+enum ePathFinderStatus {CALCULATING,  PATH_FOUND,  PATH_NOT_FOUND};
```
worktycho: Since this is new code you might want to use an enum class to avoid namespace pollution.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,390 @@
...
+#include <stdio.h>
+#include <exception>
+#include <cmath>
+
+#define DISTANCE_MANHATTEN 0  // 1: More speed, a bit less accuracy 0: Max accuracy, less speed.
+#define HEURISTICS_ONLY 0  // 1: Much more speed, much less accurate.
+#define CALCULATIONS_PER_STEP 1  // Higher means more CPU load but faster path calculations.
+// The only version which guarantees the shortest path is 0, 0.
+
+enum eCellStatus {OPENLIST,  CLOSEDLIST,  NOLIST};
```
worktycho: Consider using an enum class?

src/Mobs/Path.h
```diff
@@ -0,0 +1,168 @@
...
+	cPathCell * OpenListPop();
+	void processIfWalkable(const Vector3d &a_Location, cPathCell *a_Parent, int a_Cost);
+	
+	/* Map management */
+	void ProcessCell(cPathCell * a_Cell,  cPathCell * a_Caller,  int a_GDelta);
+	cPathCell* GetCell(const Vector3d & a_location);
+	
+	/* Pathfinding fields */
+	priority_queue<cPathCell *,  vector<cPathCell *>,  compareHeuristics> m_OpenList;
+	unordered_map<Vector3d,  cPathCell *> m_Map;
```
worktycho: If this map owns the cells it should be a map of std::unique_ptr's

src/Mobs/Monster.h
```diff
@@ -162,6 +162,10 @@ class cMonster :
 
 	/** A pointer to the entity this mobile is aiming to reach */
 	cEntity * m_Target;
+	cPath * m_Path;
```
worktycho: This should be a unique_ptr.

src/Mobs/Path.h
Woazboat: Indiscriminately importing the whole standard library namespace is Bad in pretty much all cases

src/Mobs/Path.h
wiseoldman95: Some of them are shared between m_OpenList and m_Map.

wiseoldman95:
Can someone give me a hand here? https://github.com/wiseoldman95/MCServer...rtPointers
I am trying to convert cPath to use smart pointers but it's compiler errors all over the place. I'll later merge SmartPointers with this pull request.

src/Mobs/CMakeLists.txt
madmaxoft: Not really. The CMakeLists.txt uses tabs, but the commit has spaces.

wiseoldman95:
Also, what's wrong with the CI? This branch builds properly on my machine.

src/Mobs/Monster.cpp
```diff
@@ -95,12 +95,16 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_DropChanceBoots(0.085f)
 	, m_CanPickUpLoot(true)
 	, m_BurnsInDaylight(false)
-	, m_RelativeWalkSpeed(1.0)
+	, m_RelativeWalkSpeed(0.5)
```
madmaxoft: The default should be 1; descendants should change this value in their constructors.

src/Mobs/Monster.cpp
```diff
@@ -95,12 +95,16 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
...
 {
 	if (!a_ConfigName.empty())
 	{
 		GetMonsterConfig(a_ConfigName);
 	}
+	m_Path = nullptr;
+	m_PathStatus = ePathFinderStatus::PATH_NOT_FOUND;
+	m_IsFollowingPath = false;
+	m_GiveUpCounter=0;
```
madmaxoft: Spaces around `=`

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,56 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
+		
+		// Can someone explain why are these two NOT THE SAME???
+		// m_Path = new cPath(GetWorld(), GetPosition(), m_FinalDestination, 30);
+		m_Path = new cPath(GetWorld(), Vector3d(floor(position.x), floor(position.y), floor(position.z)), Vector3d(floor(Dest.x), floor(Dest.y), floor(Dest.z)), 20);
+		
+		
+		m_IsFollowingPath = false;
+	}
+	m_PathStatus = m_Path->Step();
+	if (m_PathStatus == ePathFinderStatus::PATH_NOT_FOUND)
```
madmaxoft: Why not use a `switch` instead of the `if`s?

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,56 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
-			(!cBlockInfo::IsSolid(BlockAtYP)) &&
-			(!cBlockInfo::IsSolid(BlockAtYPP)) &&
-			(BlockAtY != E_BLOCK_FENCE) &&
-			(BlockAtY != E_BLOCK_FENCE_GATE)
-			)
-		{
-			m_PotentialCoordinates.push_back(Vector3d((gCrossCoords[i].x + PosX), PosY + 1, gCrossCoords[i].z + PosZ));
-		}
+		// printf("calculating...\n");
+		m_Destination=GetPosition();
```
madmaxoft: Spaces

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,56 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 	}
-
-	if (!m_PotentialCoordinates.empty())
+	if (m_PathStatus == ePathFinderStatus::PATH_FOUND)
 	{
-		Vector3f ShortestCoords = m_PotentialCoordinates.front();
-		for (std::vector<Vector3d>::const_iterator itr = m_PotentialCoordinates.begin(); itr != m_PotentialCoordinates.end(); ++itr)
+		if (ReachedDestination() || (m_IsFollowingPath == false))
```
madmaxoft: Never EVER compare a bool variable to a bool literal. Always use `if (var)` or `if (!var)`

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,56 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
+		{
+			// printf("Getting next point...\n");
+			m_Destination = m_Path->GetNextPoint();
+			m_IsFollowingPath = true;
+			m_GiveUpCounter=40;  // Give up after 2 seconds if failed to reach m_Dest
```
madmaxoft: Spaces

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,56 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
+			if (!ReachedDestination())
 			{
-				ShortestCoords = *itr;
+				printf("Did not reach destination\n");
 			}
+			if (!m_IsFollowingPath)
+			{
+				printf("Not following.\n");
+			}
+		}*/
```
madmaxoft: Remove the commented-out debug code

src/Mobs/Monster.cpp
```diff
@@ -292,48 +269,48 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
-			if (m_bOnGround)
-			{
-				Distance *= 2.5f;
-			}
-			else if (IsSwimming())
-			{
-				Distance *= 1.3f;
-			}
-			else
+			if (m_Destination != GetPosition())
```
madmaxoft: Does this work properly? Aren't the points specified in floating-point types?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
+// TODO uniq_ptr instead of ptr whenever possible
+
+#ifndef COMPILING_PATHFIND_DEBUGGER
```
madmaxoft: MSVC builds will not work unless the very first preprocessor instruction in each CPP file is the `#include "Globals.h"` line. You need to work around that in your PathFinderDebugger project - provide an empty Globals.h, for example.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+#define DISTANCE_MANHATTEN 0  // 1: More speed, a bit less accuracy 0: Max accuracy, less speed.
+#define HEURISTICS_ONLY 0  // 1: Much more speed, much less accurate.
+#define CALCULATIONS_PER_STEP 5  // Higher means more CPU load but faster path calculations.
+// The only version which guarantees the shortest path is 0, 0.
+
+enum class eCellStatus {OPENLIST,  CLOSEDLIST,  NOLIST};
+struct cPathCell
+{
+	Vector3d m_Location;   // Location of the cell in the world
+	int m_F, m_G, m_H;     // F, G, H as defined in regular A*.
```
madmaxoft: I'm pretty sure they have some friendly names in the algorithm description, too - such as current distance, heuristic or something like that.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+#define DISTANCE_MANHATTEN 0  // 1: More speed, a bit less accuracy 0: Max accuracy, less speed.
+#define HEURISTICS_ONLY 0  // 1: Much more speed, much less accurate.
+#define CALCULATIONS_PER_STEP 5  // Higher means more CPU load but faster path calculations.
+// The only version which guarantees the shortest path is 0, 0.
+
+enum class eCellStatus {OPENLIST,  CLOSEDLIST,  NOLIST};
+struct cPathCell
+{
+	Vector3d m_Location;   // Location of the cell in the world
```
madmaxoft: I expect the pathfinder is working with integral coords only? Why not use the `Vector3i` instead of the `Vector3d` then?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+
+
+
+
+cPath::cPath(cWorld * a_World,
+const Vector3d & a_StartingPoint, const Vector3d & a_EndingPoint, int a_MaxSteps,
+double a_BoundingBoxWidth, double a_BoundingBoxHeight,
+int a_MaxUp, int a_MaxDown)
+{
```
madmaxoft: Add indent for the params, the closing parenthesis should be on a separate line (one less indent) and the first param should be on a separate line.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+		a_Cell->m_F = a_Cell->m_H + a_Cell->m_G;  // Regular A*.
+		#endif
+		
+		OpenListAdd(a_Cell);
+		return;
+	}
+	
+
+	
+	// Case 3: Cell is in the open list, check if G and H need an update.
```
madmaxoft: Remove the excessive empty lines

src/Mobs/Path.cpp
wiseoldman95: Correct, int coords only right now. I used Vector3d because non ints are in the roadmap. But now that I think of it, some parts of the code will always be integral either way.

src/Mobs/Path.cpp
wiseoldman95: H is sometimes called Heuristics.
I believe F and G are derived from the math functions f(x) and g(x). I'll see if they have better names.

src/Mobs/Monster.cpp
wiseoldman95: It is possible that this is always returning true, this might not be needed at all, I'll check it out.

src/Mobs/Monster.cpp
wiseoldman95: Alright. Is this convention stuff or is there a technical reason?

src/Mobs/Path.cpp
NiLSPACE: Unless it's after the end of a function.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+void cPath::ProcessIfWalkable(const Vector3d & a_Location, cPathCell * a_Parent, int a_Cost)
+{
+	cPathCell * cell =  GetCell(a_Location);
+	if (!cell->m_IsSolid && GetCell(a_Location+Vector3d(0, -1, 0))->m_IsSolid && !GetCell(a_Location+Vector3d(0, 1, 0))->m_IsSolid)
+	{
+		ProcessCell(cell, a_Parent, a_Cost);
+	}
+}
+
+bool cPath::Step_Internal()
```
madmaxoft: Add empty lines above the function header.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+		return true;
+	}
+	
+	// Path found.
+	if (CurrentCell->m_Location == m_Destination)
+	{
+		// printf("cPath::Step_Internal() - Destination in closed list. Path Found.\n");
+		do
+		{
+			AddPoint(CurrentCell->m_Location+Vector3d(0.5, 0, 0.5));  // Populate the cPath with points.
```
madmaxoft: Spaces around `+`

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+			CurrentCell = CurrentCell->m_Parent;
+		}
+		while (CurrentCell != NULL);
```
madmaxoft: Merge this line with the previous one, otherwise it's really misleading (I first thought this was an endless loop)

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+		m_CurrentPoint = -1;
+		FinishCalculation(ePathFinderStatus::PATH_FOUND);
+		return true;
+	}
+	
+	// Calculation not finished yet, process a currentCell by inspecting all 8 neighbors.
```
madmaxoft: 8 neighbors?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(-1, i, 0), CurrentCell, 10);
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(0, i, 1), CurrentCell, 10);
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(0, i, -1), CurrentCell, 10);
+	}
+	
+	// Diagonals
+	int x, z;
+	for (x=-1; x<=1; x+=2)
+	{
+		for (z=-1; z<=1; z+=2)
```
madmaxoft: Spaces, spaces, spaces...

src/Mobs/Path.h
```diff
@@ -0,0 +1,167 @@
+#pragma once
+
+#ifdef COMPILING_PATHFIND_DEBUGGER
+/* Note: the COMPILING_PATHFIND_DEBUGGER flag is used by Native/WiseOldMan95 to debug this class outside of MCServer.
+This preprocessor flag is never set when compiling MCServer. */
+#include "PathFinderIrrlicht_Head.h"
+#endif
+
+#include <vector>
+#include <queue>
```
madmaxoft: vector and queue are included in Globals.h, no need to reinclude them.

src/Mobs/Path.h
```diff
@@ -0,0 +1,167 @@
...
+typedef cItemCallback<cChunk> cChunkCallback;
+#endif
+
+/* Various little structs and classes */
+enum class ePathFinderStatus {CALCULATING,  PATH_FOUND,  PATH_NOT_FOUND};
+struct cPathCell;  // Defined inside Path.cpp
+class compareHeuristics
+{
+public:
+	bool operator()(cPathCell * & a_V1,  cPathCell * &  a_V2);
```
madmaxoft: Use a typedef for `cPathCell *` to avoid the pointer-to-reference / reference-to-pointer symbol goulash.

src/Mobs/Path.h
```diff
@@ -0,0 +1,167 @@
...
+	{
+		ASSERT(m_Status == ePathFinderStatus::PATH_FOUND);
+		return m_PathPoints[m_PointCount - 1 - (++m_CurrentPoint)];
+	}
+	
+	
+	
+	
+	
+	/** Checks whether this is the last point or not. Never call getnextPoint when this is true. */
```
madmaxoft: Remove the excessive empty lines. 1 empty line between function declarations. 2 empty lines between member variables and functions.

src/Vector3.h
```diff
@@ -380,7 +381,27 @@ typedef Vector3<int>    Vector3i;
...
+	{
+		// Guaranteed to have no hash collisions for any 128x128x128 area suitable for pathfinding.
+		// TODO check the bit length thing.
+		size_t t=0;
+		t+=(char)v2.x;
+		t=t << 8;
+		t+=(char)v2.y;
+		t=t << 8;
+		t+=(char)v2.z;
+		t=t << 8;
```
madmaxoft: Spaces, c++-casts

worktycho:
Part of the reason CI is failing is that the libstdc++ we use for CI has a bug which only shows in clang. libstdc++ uses a friend class declaration in std::vector rather than a struct declaration. So in order to work with clang and this libstdc++, you need to include <functional> before including <vector>. That may mean modifying Globals.h

src/Mobs/Path.cpp
wiseoldman95: This is an outdated comment.
We check 4 neighbors on y-1 (NSEW)
8 neighbors on y (Diagonals + NSEW)
4 neighbors on y+1 (NSEW)

src/Mobs/Path.cpp
wiseoldman95: Gotcha, sorry :P
I'll eventually get used to it.

src/Mobs/Path.h
wiseoldman95: I'm trying to ditch pointers all together in the SmartPointers branch, not going well as of now.

src/Mobs/Path.cpp
wiseoldman95: Note that the chunk is queried far less than that number thanks to caching.

src/Mobs/Path.h
wiseoldman95: Including Globals.h in all files spawned hundreds of errors, mainly INSIDE globals.h,. Any idea?

madmaxoft: What all files? You should only include it from CPP files, which in your case is a single file - Path.cpp

wiseoldman95: Got it, no Globals.h in headers. Correct?

src/Mobs/Monster.cpp
wiseoldman95: I'd like to keep it just for a bit more, might need it. I will remove all the printfs just before we merge.

src/Vector3.h
wiseoldman95: Will do spaces. Regarding casts: What casts?
I'm probably going to make some changes to the types later (see  Woazboat's comment).

src/Mobs/Path.h
wiseoldman95: Thought you wanted 5 because the implementation is in the header.

src/Mobs/Path.cpp
wiseoldman95: Added to my to-do list.

src/Mobs/Monster.cpp
wiseoldman95: Good point.

src/Mobs/Monster.cpp
wiseoldman95: Why should the default be 1?  Sometimes the momentum causes mobs to fall off cliffs and miss their pathfinding waypoints.

Can we keep it 0.5 for now? The AI needs some more thorough revision later, but for now, this is the fastest fix.

wiseoldman95:
#include <Functional> didn't help the CI, so I reverted that.

src/Mobs/Path.cpp
wiseoldman95: I didn't find any common names. Even if there are non-common ones, I think that F, G, and H are much more recognizable and in fact any other name would be misleading.

wiseoldman95:
Can we do the merging now? I see no serious issues. The little issues can be resolved later.

(Sorted by order of urgency)
*Current issues visible to the players:*
- Fix mob daylight behavior
- Jump helper waypoints

*Stuff to do that are invisible to the player:*
- smartPointers
- Vector3i wherever Vector3d will never be needed
- Check if if (m_Destination != GetPosition()) is needed in monster.cpp
- check the size_t issue in the Vector3d hash function (I think that most systems have a good enough size_t, besides, some hash collisions are tolerable)

madmaxoft:
Would it be possible to define the hash functional not globally, but only within the template where it is needed? Something like:
```cpp
struct hash
{
  // hashing support
}
std::unordered_map<Vector3i, cPathCell *, hash> m_Map
```

madmaxoft:
We can't merge because the builds are still failing.

src/Mobs/Monster.cpp
madmaxoft: it's called **relative**, so 1 means "the default". If you want a different default, change the base then.

wiseoldman95:
I'll see what I can do that with the hash.

What can be done about the CI? I have no idea what's wrong.

madmaxoft:
Also, before a merge I'd like you to squash everything into a single commit, the partial commits have unneeded files in them, unneeded changes in them and don't compile (which will break `git bisect`)

madmaxoft:
The hash is what's wrong, that's why I'm trying to find a workaround - if you can't use a struct and can't use a class, then don't use it at all (as the global object, but rather locally)

wiseoldman95:
Ok. Do I also need the #include <Functional>?

wiseoldman95:
Also, what's the exact error? Is it something we could resolve by modifying the hash function itself? Or is it just a scope thing?

jan64:
https://travis-ci.org/mc-server/MCServer/jobs/60535182

wiseoldman95:
Thanks, Let's see if that fixes it...

wiseoldman95:
The latest one is throwing errors even in GCC, but it's exactly like this SO question: http://stackoverflow.com/questions/15595804/c-hash-function-for-string-in-unordered-map

What have I done differently?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+	}
+}
+
+
+
+
+
+void cPath::FinishCalculation()
+{
+	for (std::unordered_map<Vector3d, cPathCell *>::iterator it = m_Map.begin(); it != m_Map.end(); ++it)
```
worktycho: This is still using std::hash. It should probably be a range for loop, and use auto to avoid this, if you can't get the smart pointers working:
```
for (auto&& pair : m_Map)
{
	delete pair.second;
}
```

wiseoldman95: I would have never found this...

wiseoldman95:
Travis looks happy, I'll prepare the squashed commit. Is it possible to squash the same branch? I only know how to do it with a separate new branch and I failed to find the right magic Google words.

wiseoldman95:
oepend squashed version of the pr ^

src/Mobs/Monster.cpp
wiseoldman95: Oh, I see. Is pushing another commit to squashed PR ok?

wiseoldman95: If 1 means the default, then the default should be lower than it currently is.

PR #1917, opened 2015-04-28
Small code cleanup

Woazboat:
Just some small code cleanup.

Possible div by 0 in Tracer.cpp not yet fixed because I don't understand enough about what the code is supposed to do. 
(Div by 0 also listed in https://gist.github.com/tommysanterre/59350aa90a9bbc2082cf)

### DIFF
--- a/src/Tracer.cpp
+++ b/src/Tracer.cpp
@@ -16,14 +16,14 @@ const float FLOAT_EPSILON = 0.0001f;  // TODO: Stash this in some header where i
 
 
 const std::array<const Vector3f, 6> cTracer::m_NormalTable =
-{
+{{
 	Vector3f(-1,  0,  0),  // 1: -x
 	Vector3f( 0,  0, -1),  // 2: -z
 	Vector3f( 1,  0,  0),  // 3: +x
 	Vector3f( 0,  0,  1),  // 4: +z
 	Vector3f( 0,  1,  0),  // 5: +y
 	Vector3f( 0, -1,  0)   // 6: -y
-};
+}};
 
 
 
--- a/src/Tracer.cpp
+++ b/src/Tracer.cpp
@@ -86,7 +86,7 @@ void cTracer::SetValues(const Vector3f & a_Start, const Vector3f & a_Direction)
 	// same but y-direction
 	if (dir.x != 0.f)
 	{
-		tDelta.x = 1 / fabs(dir.x);
+		tDelta.x = 1 / std::abs(dir.x);
 	}
 	else
 	{
@@ -94,7 +94,7 @@ void cTracer::SetValues(const Vector3f & a_Start, const Vector3f & a_Direction)
 	}
 	if (dir.y != 0.f)
 	{
-		tDelta.y = 1 / fabs(dir.y);
+		tDelta.y = 1 / std::abs(dir.y);
 	}
 	else
 	{
@@ -102,7 +102,7 @@ void cTracer::SetValues(const Vector3f & a_Start, const Vector3f & a_Direction)
 	}
 	if (dir.z != 0.f)
 	{
-		tDelta.z = 1 / fabs(dir.z);
+		tDelta.z = 1 / std::abs(dir.z);
 	}
 	else
 	{
@@ -312,7 +312,7 @@ int cTracer::intersect3D_SegmentPlane(const Vector3f & a_Origin, const Vector3f 
 	float     D = a_PlaneNormal.Dot(u);      // dot(Pn.n, u);
 	float     N = -(a_PlaneNormal.Dot(w));  // -dot(a_Plane.n, w);
 
-	if (fabs(D) < FLOAT_EPSILON)
+	if (std::abs(D) < FLOAT_EPSILON)
 	{
 		// segment is parallel to plane
 		if (N == 0.0)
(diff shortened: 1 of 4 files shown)


src/Tracer.cpp
```diff
@@ -295,8 +310,7 @@ int cTracer::intersect3D_SegmentPlane(const Vector3f & a_Origin, const Vector3f
 	float     D = a_PlaneNormal.Dot(u);      // dot(Pn.n, u);
 	float     N = -(a_PlaneNormal.Dot(w));  // -dot(a_Plane.n, w);
 
-	const float EPSILON = 0.0001f;
-	if (fabs(D) < EPSILON)
+	if (fabs(D) < FLOAT_EPSILON)
```
madmaxoft: How about `std::abs`?

src/Tracer.cpp
```diff
@@ -12,17 +12,24 @@
 
 
 
+const float FLOAT_EPSILON = 0.0001f;  // TODO: Stash this in some header where it can be reused
+
+
+const std::array<const Vector3f, 6> cTracer::m_NormalTable =
+{{
```
madmaxoft: You need to put each brace on a separate line. The basic style checker has a false positive for this occurrence.

madmaxoft:
You can run the code style checker yourself manually, it's in the `$/src/CheckBasicStyle.lua` file. You need to have Lua (5.1) installed for it to work, though.

src/Vector3.h
```diff
@@ -121,11 +126,7 @@ class Vector3
 	{
 		// Perform a bitwise comparison of the contents - we want to know whether this object is exactly equal
 		// To perform EPS-based comparison, use the EqualsEps() function
-		return (
-			(memcmp(&x, &a_Rhs.x, sizeof(x)) == 0) &&
-			(memcmp(&y, &a_Rhs.y, sizeof(y)) == 0) &&
-			(memcmp(&z, &a_Rhs.z, sizeof(z)) == 0)
-		);
+		return !((x != a_Rhs.x) || (y != a_Rhs.y) || (z != a_Rhs.z));
```
madmaxoft: This unfortunately triggers warnings in clang (https://travis-ci.org/mc-server/MCServer/jobs/60462042), I'd prefer to keep it as it was.

Woazboat: It does, but the behavior it warns about (bitwise/strict comparison of floats) is exactly what is wanted in this case, so the warnings are false positives.
The original version is way unsafer and does the same (bitwise/strict comparison of floats). The only difference is that clang doesn't recognize it as such and thus doesn't trigger a warning.

madmaxoft: How is a memcmp unsafer?

Unless you know of a way to explicitly turn off the warning for this specific instance (and not for the rest of the code), then make the change, otherwise please change this back to the original code. We want as few warnings as possible.

jan64:
https://stackoverflow.com/questions/7460790/controlling-clang-warnings

src/Vector3.h
jan64: https://stackoverflow.com/questions/7460790/controlling-clang-warnings

Woazboat: By using memcmp you are assuming that the object you are comparing is POD. Since Vector3 is a template, you can't guarantee that. 
Then there's things like positive zero and negative zero for floats which should compare equal, but have a different bit representation.

madmaxoft: The types we're comparing ARE POD - they're `int`, `float` or `double`, based on the template param.
Positive and negative zero most likely won't compare well with `!=` either, since clang basically warns that it's using the bitwise comparison. I'm still not convinced.

Woazboat: You still have a template, which should work no matter what type it's used with. If someone uses it with a non-POD type it'll fail without warning.

IEEE-754 positive and negative zero are required to compare equal, which they do.
```
    std::cout << -0.0f << " == " << 0.0f <<std::endl;
    if(-0.0f == 0.0f)
    {
        std::cout << "true" << std::endl;
    }
    else
    {
        std::cout << "false" <<std::endl;
    }
```
Output:
```
-0 == 0
true
```

worktycho: The compiler can convert a -0.0f literal into a positive zero, so that isn't definite evidence, but I agree with the rgument.

Woazboat:
Uhh... the clang errors weren't fixed yet in the version you just merged. I just wanted to push that a minute ago. The commits also aren't squashed yet.

worktycho:
I've reverted.
However tthere's no point squashing now its in the public history. Also, I'm in support of the changes that caused the clang warnings for the reasons you gave. It would be much better to use pragma's to say that we're explicitly ignoring the warnings than that horrible hack to trick the compiler.

Woazboat:
That's what I meant by "fixing" the clang errors. Adding the override pragmas.
I'm going to have to create a new pull request anyway because this one is already closed, so I might as well refactor the commits into pieces that make sense while I'm at it. (And 100% fewer one line commits just because I forgot to insert a space somewhere  ;)

worktycho:
I wouldn't worry about it unless you're getting warnings off the explicit instantiation. There's another 7 of the warnings scattered around the code base, and it would be batter to fix the calls so the method is not instantiated for floating points in the first place.

Woazboat:
As far as I can tell it generates an error for every float instantiation, which seems to be happening twice. So, no warning overrides?
I just realized though that I need to add warning overrides to the HasNonZeroLength member function I made, because that's generating warnings as well and those really are false positives.

Woazboat:
New version of this: #1945

PR #1919, opened 2015-04-29
Added Guardian and Rabbit to DefaultMonsters in Overworld

mathias-github:
I added Guardian and Rabbit to the default mobs in world.ini.


NiLSPACE:
Are there classes for them already?

tigerw:
I think so.

PR #1920, opened 2015-04-29
Implemented A* Pathfinding and updated monster.cpp to use it. 

wiseoldman95:
Squashed version of https://github.com/mc-server/MCServer/pull/1916

### DIFF
--- a/src/Mobs/CMakeLists.txt
+++ b/src/Mobs/CMakeLists.txt
@@ -24,6 +24,7 @@ SET (SRCS
 	Mooshroom.cpp
 	PassiveAggressiveMonster.cpp
 	PassiveMonster.cpp
+	Path.cpp
 	Pig.cpp
 	Rabbit.cpp
 	Sheep.cpp
@@ -62,6 +63,7 @@ SET (HDRS
 	Ocelot.h
 	PassiveAggressiveMonster.h
 	PassiveMonster.h
+	Path.h
 	Pig.h
 	Rabbit.h
 	Sheep.h
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -13,7 +13,7 @@
 #include "../Chunk.h"
 #include "../FastRandom.h"
 
-
+#include "Path.h"
 
 
 
@@ -74,6 +74,10 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_EMState(IDLE)
 	, m_EMPersonality(AGGRESSIVE)
 	, m_Target(nullptr)
+	, m_Path(nullptr)
+	, m_PathStatus(ePathFinderStatus::PATH_NOT_FOUND)
+	, m_IsFollowingPath(false)
+	, m_GiveUpCounter(0)
 	, m_bMovingToDestination(false)
 	, m_LastGroundHeight(POSY_TOINT)
 	, m_IdleInterval(0)
@@ -94,8 +98,9 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_DropChanceLeggings(0.085f)
 	, m_DropChanceBoots(0.085f)
 	, m_CanPickUpLoot(true)
+	, m_TicksSinceLastDamaged(100)
 	, m_BurnsInDaylight(false)
-	, m_RelativeWalkSpeed(1.0)
+	, m_RelativeWalkSpeed(1)
 {
 	if (!a_ConfigName.empty())
 	{
@@ -118,87 +123,52 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
 
 void cMonster::TickPathFinding()
 {
-	const int PosX = POSX_TOINT;
-	const int PosY = POSY_TOINT;
-	const int PosZ = POSZ_TOINT;
 
-	std::vector<Vector3d> m_PotentialCoordinates;
-	m_TraversedCoordinates.push_back(Vector3i(PosX, PosY, PosZ));
+	if (m_Path == nullptr)
+	{
+		Vector3d position = GetPosition();
+		Vector3d Dest = m_FinalDestination;
 
-	static const struct  // Define which directions to try to move to
-	{
-		int x, z;
-	} gCrossCoords[] =
-	{
-		{ 1, 0},
-		{-1, 0},
-		{ 0, 1},
-		{ 0, -1},
-	} ;
-	
-	if ((PosY - 1 < 0) || (PosY + 2 >= cChunkDef::Height) /* PosY + 1 will never be true if PosY + 2 is not */)
-	{
-		// Too low/high, can't really do anything
-		FinishPathFinding();
-		return;
+		// Can someone explain why are these two NOT THE SAME???
+		// m_Path = new cPath(GetWorld(), GetPosition(), m_FinalDestination, 30);
+		m_Path = new cPath(GetWorld(), Vector3d(floor(position.x), floor(position.y), floor(position.z)), Vector3d(floor(Dest.x), floor(Dest.y), floor(Dest.z)), 20);
+
+
+		m_IsFollowingPath = false;
 	}
-
-	for (size_t i = 0; i < ARRAYCOUNT(gCrossCoords); i++)
+	m_PathStatus = m_Path->Step();
+	switch (m_PathStatus)
 	{
-		if (IsCoordinateInTraversedList(Vector3i(gCrossCoords[i].x + PosX, PosY, gCrossCoords[i].z + PosZ)))
+
+		case ePathFinderStatus::PATH_NOT_FOUND:
 		{
-			continue;
+			FinishPathFinding();
+			break;
 		}
 
-		BLOCKTYPE BlockAtY = m_World->GetBlock(gCrossCoords[i].x + PosX, PosY, gCrossCoords[i].z + PosZ);
-		BLOCKTYPE BlockAtYP = m_World->GetBlock(gCrossCoords[i].x + PosX, PosY + 1, gCrossCoords[i].z + PosZ);
-		BLOCKTYPE BlockAtYPP = m_World->GetBlock(gCrossCoords[i].x + PosX, PosY + 2, gCrossCoords[i].z + PosZ);
-		int LowestY = FindFirstNonAirBlockPosition(gCrossCoords[i].x + PosX, gCrossCoords[i].z + PosZ);
-		BLOCKTYPE BlockAtLowestY = (LowestY >= cChunkDef::Height) ? E_BLOCK_AIR : m_World->GetBlock(gCrossCoords[i].x + PosX, LowestY, gCrossCoords[i].z + PosZ);
 
-		if (
-			(!cBlockInfo::IsSolid(BlockAtY)) &&
-			(!cBlockInfo::IsSolid(BlockAtYP)) &&
-			(!IsBlockLava(BlockAtLowestY)) &&
-			(BlockAtLowestY != E_BLOCK_CACTUS) &&
-			(PosY - LowestY < FALL_DAMAGE_HEIGHT)
-			)
+		case ePathFinderStatus::CALCULATING:
 		{
-			m_PotentialCoordinates.push_back(Vector3d((gCrossCoords[i].x + PosX), PosY, gCrossCoords[i].z + PosZ));
+			m_Destination = GetPosition();
+			break;
 		}
-		else if (
-			(cBlockInfo::IsSolid(BlockAtY)) &&
-			(BlockAtY != E_BLOCK_CACTUS) &&
-			(!cBlockInfo::IsSolid(BlockAtYP)) &&
-			(!cBlockInfo::IsSolid(BlockAtYPP)) &&
-			(BlockAtY != E_BLOCK_FENCE) &&
-			(BlockAtY != E_BLOCK_FENCE_GATE)
-			)
-		{
-			m_PotentialCoordinates.push_back(Vector3d((gCrossCoords[i].x + PosX), PosY + 1, gCrossCoords[i].z + PosZ));
-		}
-	}
 
-	if (!m_PotentialCoordinates.empty())
-	{
-		Vector3f ShortestCoords = m_PotentialCoordinates.front();
-		for (std::vector<Vector3d>::const_iterator itr = m_PotentialCoordinates.begin(); itr != m_PotentialCoordinates.end(); ++itr)
+
+		case ePathFinderStatus::PATH_FOUND:
 		{
-			Vector3f Distance = m_FinalDestination - ShortestCoords;
-			Vector3f Distance2 = m_FinalDestination - *itr;
-			if (Distance.SqrLength() > Distance2.SqrLength())
+			if (ReachedDestination() || !m_IsFollowingPath)
 			{
-				ShortestCoords = *itr;
+				m_Destination = m_Path->GetNextPoint();
+				m_IsFollowingPath = true;
+				m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_Dest.
 			}
-		}
+			if (m_Path->IsLastPoint())
+			{
+				FinishPathFinding();
+			}
+			break;
 
-		m_Destination = ShortestCoords;
-		m_Destination.z += 0.5f;
-		m_Destination.x += 0.5f;
-	}
-	else
-	{
-		FinishPathFinding();
+		}
 	}
 }
 
@@ -206,20 +176,47 @@ void cMonster::TickPathFinding()
 
 
 
+/* Currently, the mob will only start moving to a new position after the position it is
+currently going to is reached. */
 void cMonster::MoveToPosition(const Vector3d & a_Position)
 {
-	FinishPathFinding();
-
 	m_FinalDestination = a_Position;
 	m_bMovingToDestination = true;
-	TickPathFinding();
 }
+
+
+
+
+
+void cMonster::StopMovingToPosition()
+{
+	m_bMovingToDestination = false;
+	FinishPathFinding();
+}
+
+
 
 
 
 bool cMonster::IsCoordinateInTraversedList(Vector3i a_Coords)
 {
 	return (std::find(m_TraversedCoordinates.begin(), m_TraversedCoordinates.end(), a_Coords) != m_TraversedCoordinates.end());
+}
+
+
+
+
+
+/* No one should call this except the pathfinder orthe monster tick or StopMovingToPosition.
+Resets the pathfinder, usually starting a brand new path, unless called from StopMovingToPosition. */
+void cMonster::FinishPathFinding(void)
+{
+	if (m_Path != nullptr)
+	{
+		delete m_Path;
+		m_Path = nullptr;
+
+	}
 }
 
 
@@ -239,13 +236,14 @@ bool cMonster::ReachedDestination()
 
 
 
+
 bool cMonster::ReachedFinalDestination()
 {
 	if ((GetPosition() - m_FinalDestination).Length() <= m_AttackRange)
 	{
 		return true;
 	}
-	
+
 	return false;
 }
 
@@ -268,6 +266,10 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		return;
 	}
 
+	if (m_TicksSinceLastDamaged < 100)
+	{
+		++m_TicksSinceLastDamaged;
+	}
 	if ((m_Target != nullptr) && m_Target->IsDestroyed())
 	{
 		m_Target = nullptr;
@@ -275,6 +277,8 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 	// Burning in daylight
 	HandleDaylightBurning(a_Chunk);
+
+
 
 	if (m_bMovingToDestination)
 	{
@@ -289,52 +293,50 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 			}
 		}
 
+		TickPathFinding();
+
 		Vector3d Distance = m_Destination - GetPosition();
 		if (!ReachedDestination() && !ReachedFinalDestination())  // If we haven't reached any sort of destination, move
 		{
-			Distance.y = 0;
-			Distance.Normalize();
-
-			if (m_bOnGround)
-			{
-				Distance *= 2.5f;
-			}
-			else if (IsSwimming())
-			{
-				Distance *= 1.3f;
-			}
-			else
-			{
-				// Don't let the mob move too much if he's falling.
-				Distance *= 0.25f;
-			}
-
-			// Apply walk speed:
-			Distance *= m_RelativeWalkSpeed;
-
-			AddSpeedX(Distance.x);
-			AddSpeedZ(Distance.z);
-
-			// It's too buggy!
-			/*
-			if (m_EMState == ESCAPING)
-			{
-				// Runs Faster when escaping :D otherwise they just walk away
-				SetSpeedX (GetSpeedX() * 2.f);
-				SetSpeedZ (GetSpeedZ() * 2.f);
-			}
-			*/
-		}
-		else
-		{
-			if (ReachedFinalDestination())  // If we have reached the ultimate, final destination, stop pathfinding and attack if appropriate
+			if (--m_GiveUpCounter == 0)
 			{
 				FinishPathFinding();
 			}
 			else
 			{
-				TickPathFinding();  // We have reached the next point in our path, calculate another point
+				Distance.y = 0;
+				Distance.Normalize();
+
+				if (m_bOnGround)
+				{
+					Distance *= 2.5f;
+				}
+				else if (IsSwimming())
+				{
+					Distance *= 1.3f;
+				}
+				else
+				{
+					// Don't let the mob move too much if he's falling.
+					Distance *= 0.25f;
+				}
+
+				// Apply walk speed:
+				Distance *= m_RelativeWalkSpeed;
+
+				/* Reduced default speed.
+				Close to Vanilla, easier for mobs to follow m_Destinations, hence
+				better pathfinding. */
+				Distance *= 0.5;
+
+				AddSpeedX(Distance.x);
+				AddSpeedZ(Distance.z);
+
 			}
+		}
+		else if (ReachedFinalDestination())
+		{
+			StopMovingToPosition();
 		}
 	}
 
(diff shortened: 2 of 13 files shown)


wiseoldman95:
Is it ok to push some trivial commits which will definitely compile? (Changing relative speed, removing spacings before Path.h in CmakeLists, and things like that.)

NiLSPACE:
Of course ^^

wiseoldman95:
Despite this?
> Also, before a merge I'd like you to squash everything into a single commit, the partial commits have unneeded files in them, unneeded changes in them and don't compile (which will break git bisect)

wiseoldman95:
Despite this?
> Also, before a merge I'd like you to squash everything into a single commit, the partial commits have unneeded files in them, unneeded changes in them and don't compile (which will break git bisect)

wiseoldman95:
Merging is welcome.
Any further code remarks are also welcome.

wiseoldman95:
SmartPointers will be implemented later in a different PR.

src/Mobs/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ SET (SRCS
 	Mooshroom.cpp
 	PassiveAggressiveMonster.cpp
 	PassiveMonster.cpp
+        Path.cpp
```
madmaxoft: Use a tab here

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,43 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
-			(BlockAtY != E_BLOCK_CACTUS) &&
-			(!cBlockInfo::IsSolid(BlockAtYP)) &&
-			(!cBlockInfo::IsSolid(BlockAtYPP)) &&
-			(BlockAtY != E_BLOCK_FENCE) &&
-			(BlockAtY != E_BLOCK_FENCE_GATE)
-			)
-		{
-			m_PotentialCoordinates.push_back(Vector3d((gCrossCoords[i].x + PosX), PosY + 1, gCrossCoords[i].z + PosZ));
-		}
+		Vector3d position=GetPosition();
```
madmaxoft: Spaces

src/Mobs/CMakeLists.txt
wiseoldman95: I could swear I already did that.

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,43 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 	{
-		Vector3f ShortestCoords = m_PotentialCoordinates.front();
-		for (std::vector<Vector3d>::const_iterator itr = m_PotentialCoordinates.begin(); itr != m_PotentialCoordinates.end(); ++itr)
-		{
-			Vector3f Distance = m_FinalDestination - ShortestCoords;
-			Vector3f Distance2 = m_FinalDestination - *itr;
-			if (Distance.SqrLength() > Distance2.SqrLength())
+		case ePathFinderStatus::PATH_NOT_FOUND:
+			FinishPathFinding();
+			break;
```
madmaxoft: Each `case` block should have a brace around it.

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,43 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
+		
+		case ePathFinderStatus::CALCULATING:
+			m_Destination = GetPosition();
+			break;
+		
+		case ePathFinderStatus::PATH_FOUND:
+			if (ReachedDestination() || (!m_IsFollowingPath))
```
madmaxoft: No need to parenthesize a simple `!var`

src/Mobs/Monster.cpp
```diff
@@ -118,87 +122,43 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 			{
-				ShortestCoords = *itr;
+				m_Destination = m_Path->GetNextPoint();
+				m_IsFollowingPath = true;
+				m_GiveUpCounter=40;  // Give up after 2 seconds if failed to reach m_Dest.
```
madmaxoft: Spaces

src/Mobs/Monster.h
```diff
@@ -203,8 +208,13 @@ class cMonster :
...
 	inline void FinishPathFinding(void)
 	{
-		m_TraversedCoordinates.clear();
 		m_bMovingToDestination = false;
+		if (m_Path != nullptr)
+		{
+			delete m_Path;
+			m_Path = nullptr;
+			
+		}
```
madmaxoft: This is already a large function, consider moving it into CPP

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
+#include "Globals.h"
+#ifndef COMPILING_PATHFIND_DEBUGGER
+/* MCServer headers */
+#include "../World.h"
+#include "../Chunk.h"
+#endif
```
madmaxoft: Use indentation in preprocessor commands:
```cpp
#if ...
  #something
#endif
```

src/Mobs/CMakeLists.txt
wiseoldman95: This was done in "Fixed RelativeSpeed".

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+
+#include "Path.h"
+
+#include <stdio.h>
+#include <exception>
```
madmaxoft: Is `<exception>` still needed?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+#include <cmath>
+
+#define DISTANCE_MANHATTEN 0  // 1: More speed, a bit less accuracy 0: Max accuracy, less speed.
```
madmaxoft: The correct spelling is `MANHATTAN`

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+bool compareHeuristics::operator()(cPathCell * & a_V1, cPathCell * & a_V2)
+{
+	return a_V1->m_F > a_V2->m_F;
+}
+
+
+
+
+
+/* cPath implementation */
```
madmaxoft: Move constructor to the top of implementation, then destructor, then the rest. The order should match the declaration order in the header file.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+	}
+}
+
+
+
+
+
+void cPath::FinishCalculation()
+{
+	for (auto&& pair : m_Map)
```
madmaxoft: Spaces
```cpp
for (auto && pair: m_Map)
```

madmaxoft: ALso I'm not sure about why the rvalue ref is used instead of a regular ref, is there any specific reason?

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+		Cell->m_IsSolid = IsSolid(a_Location);
+		Cell->m_Status = eCellStatus::NOLIST;
+		#ifdef COMPILING_PATHFIND_DEBUGGER
+		#ifdef COMPILING_PATHFIND_DEBUGGER_MARK_UNCHECKED
+		si::setBlock(a_Location.x, a_Location.y, a_Location.z, debug_unchecked, Cell->m_IsSolid ? NORMAL : MINI);
+		#endif
+		#endif
+		return Cell;
+	}
+	else return m_Map[a_Location];
```
madmaxoft: Make this a separate braced block.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+			a_Cell->m_G = a_Caller->m_G + a_GDelta;
+		}
+		else
+		{
+			a_Cell->m_G = 0;
+		}
+		// Calculate H. This is A*'s Heuristics value.
+		#if DISTANCE_MANHATTEN == 1
+		// Manhatten distance. DeltaX + DeltaY + DeltaZ.
+		a_Cell->m_H = 10 * (abs(a_Cell->m_Location.x-m_Destination.x) + abs(a_Cell->m_Location.y-m_Destination.y) + abs(a_Cell->m_Location.z-m_Destination.z));
```
madmaxoft: spaces

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+		#else
+		// Euclidian distance. sqrt(DeltaX^2 + DeltaY^2 + DeltaZ^2), more precise.
+		a_Cell->m_H = std::sqrt( (a_Cell->m_Location.x-m_Destination.x) * (a_Cell->m_Location.x-m_Destination.x) * 100+ (a_Cell->m_Location.y-m_Destination.y) * (a_Cell->m_Location.y-m_Destination.y) * 100 + (a_Cell->m_Location.z-m_Destination.z) * (a_Cell->m_Location.z-m_Destination.z) * 100);
```
madmaxoft: Spaces. Add around `-`, remove between `( (`

wiseoldman95:
Xoft, you're annoying :P
But I have to thank you for that, you' made me wanna get used to conventions, which would make me a better programmer on the long run.

Fixing in a moment.

src/Mobs/Path.cpp
madmaxoft: How about using the math already present in the `Vector3` class? 
```cpp
(a_Cell->m_Location - m_Destination) * 100).Length()
```
should do the trick.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+	
+}
+
+
+
+
+
+ePathFinderStatus cPath::Step()
+{
+	if (m_Status == ePathFinderStatus::CALCULATING)
```
madmaxoft: Invert the condition and return early:
```
if (m_Status != CALCULATING)
{
    return m_Status;
}
```
You'll have one less indent level, much easier to read

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+}
+
+
+
+
+
+void cPath::ProcessIfWalkable(const Vector3d & a_Location, cPathCell * a_Parent, int a_Cost)
+{
+	cPathCell * cell =  GetCell(a_Location);
+	if (!cell->m_IsSolid && GetCell(a_Location+Vector3d(0, -1, 0))->m_IsSolid && !GetCell(a_Location+Vector3d(0, 1, 0))->m_IsSolid)
```
madmaxoft: spaces

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+
+
+
+bool cPath::Step_Internal()
+{
+	cPathCell * CurrentCell = OpenListPop();
+	
+	// Path not reachable, open list exauhsted.
+	if (CurrentCell == NULL)
```
madmaxoft: Use `nullptr` instead of `NULL` everywhere. We're C++11.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+		m_CurrentPoint = -1;
+		FinishCalculation(ePathFinderStatus::PATH_FOUND);
+		return true;
+	}
+	
+	// Calculation not finished yet, process a currentCell by inspecting all neighbors.
+	
+	// Check North, South, East, West on all 3 different heights.
+	int i;
+	for (i=-1; i <= 1; ++i)
```
madmaxoft: Spaces

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,382 @@
...
+	{
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(1, i, 0), CurrentCell, 10);
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(-1, i, 0), CurrentCell, 10);
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(0, i, 1), CurrentCell, 10);
+		ProcessIfWalkable(CurrentCell->m_Location + Vector3d(0, i, -1), CurrentCell, 10);
```
madmaxoft: Align the last four lines, using spaces

src/Mobs/Path.h
```diff
@@ -0,0 +1,158 @@
+#pragma once
+
+#ifdef COMPILING_PATHFIND_DEBUGGER
+/* Note: the COMPILING_PATHFIND_DEBUGGER flag is used by Native/WiseOldMan95 to debug
+this class outside of MCServer. This preprocessor flag is never set when compiling MCServer. */
+#include "PathFinderIrrlicht_Head.h"
```
madmaxoft: Indent

src/Vector3.h
```diff
@@ -379,10 +380,6 @@ typedef Vector3<float>  Vector3f;
 typedef Vector3<int>    Vector3i;
 // tolua_end
 
-
-
-
-
```
madmaxoft: Get rid of the changes in this file.

src/Server.cpp
```diff
@@ -437,7 +439,7 @@ void cServer::ExecuteConsoleCommand(const AString & a_Cmd, cCommandOutputCallbac
 	{
 		return;
 	}
-
+
```
madmaxoft: Get rid of changes in this file.

madmaxoft:
Yes, I know I'm annoying, but imagine what the code would have looked like if I weren't, and everybody used their own idea of writing code...

src/Mobs/Path.cpp
wiseoldman95: How come lua isn't complaining? It used to bash me with errors when it comes to spaces.

src/Mobs/Path.cpp
wiseoldman95: I never even used `auto` and `pair` before and honestly I cannot answer you. Worktycho suggested that particular line as a workaround for the hash issue.

src/Mobs/Path.cpp
wiseoldman95: Nope.

src/Mobs/Monster.h
wiseoldman95: I probably misunderstood you,  I don't see what's the relationship between a large function and moving it into CPP. (I assume CPP stands for cPlusPlus?)

I plan to migrate everything to smart pointers pretty soon.

src/Mobs/Path.cpp
madmaxoft: Some of those errors cannot be checked in a simple way. We cannot check for a space after a minus, because there are things like `a -= 1` or `b = -a` which are legal.

src/Mobs/Monster.h
madmaxoft: The .cpp file

src/Mobs/Path.cpp
wiseoldman95: *Weeps*

src/Mobs/Path.cpp
madmaxoft: @worktycho can you enlighten me? I think in this particular case a const ref (`for auto &`) would be a better choice, since we don't want to do any moving.

Of course, all of this would be irrelevant if the map used `UniquePtr`s

src/Mobs/CMakeLists.txt
wiseoldman95: My apologies, it was not done before.

src/Mobs/Path.cpp
wiseoldman95: Note that return m_Status is triggered anyways, always. This is not an else.
But yes, it can be done.

src/Mobs/Path.cpp
wiseoldman95: Assuming you meant last two columns. Correct me otherwise.

wiseoldman95:
Hope you realized it was an indirect compliment.

src/Mobs/Path.cpp
madmaxoft: I meant the last four lines, all columns, including the vector's coords. Everything where alignment makes sense (similar items)

madmaxoft:
```cpp
ProcessIfWalkable(CurrentCell->m_Location + Vector3d( 1, i,  0), CurrentCell, 10);
ProcessIfWalkable(CurrentCell->m_Location + Vector3d(-1, i,  0), CurrentCell, 10);
ProcessIfWalkable(CurrentCell->m_Location + Vector3d( 0, i,  1), CurrentCell, 10);
ProcessIfWalkable(CurrentCell->m_Location + Vector3d( 0, i, -1), CurrentCell, 10);
```

src/Server.cpp
wiseoldman95: Seems that my IDE tabs all newlines to make them match the current indentation level... *Weeps*

src/Mobs/Path.cpp
wiseoldman95: Done.

wiseoldman95:
All comments have been treated.
Awaiting merge / further comments.

wiseoldman95:
Perhaps the issue with frozen zombies should be fixed before this merge.

wiseoldman95:
If anyone can find out why this happens, it'd be appreciated. It's a problem with monster.cpp which freezes mobs sometimes, I can't find a clear pattern.

src/Mobs/Monster.cpp
```diff
@@ -95,12 +95,16 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
...
+	, m_RelativeWalkSpeed(1)
 {
 	if (!a_ConfigName.empty())
 	{
 		GetMonsterConfig(a_ConfigName);
 	}
+	m_Path = nullptr;
+	m_PathStatus = ePathFinderStatus::PATH_NOT_FOUND;
+	m_IsFollowingPath = false;
+	m_GiveUpCounter = 0;
```
Woazboat: This should probably be in the initializer list

wiseoldman95:
Mob freezing / daylight awkwardness fixed.

wiseoldman95:
Not adding any more new features in this merge. Only critical bugs or syntax / codestyle stuff.

wiseoldman95:
@tigerw , for some reason, the comments you make are different from @madmaxoft, his comments disappear when I fix the issue and " madmaxoft commented on an outdated diff" shows instead, yours never do that.

madmaxoft:
That's because @tigerw is commenting directly on the individual commits, while I'm commenting in the code diff for the PR.

wiseoldman95:
Oh, I see.
Will I need to re-squash this?

madmaxoft:
Yes, once it is finished.

You don't have to make a new PR, though, you can simply force-push the branch to your repo after squashing, overwriting this branch, and it will get updated properly.

wiseoldman95:
Wouldn't that ruin it for those who fetched the branch?

madmaxoft:
Not much, they'd need to merge their changes onto the new branch, but it's pretty much unlikely that anyone else than you is making changes to this.

wiseoldman95:
Could you give me the exact git commands? I currently create a new branch then do git merge --squash, this forces me to do it on a new branch.

madmaxoft:
`git rebase -i master`

wiseoldman95:
Alright, we should wait a bit for more comments.
Tell me when you want to merge and I'll squash it.

src/Mobs/Monster.cpp
```diff
@@ -1126,6 +1115,28 @@ void cMonster::HandleDaylightBurning(cChunk & a_Chunk)
...
+{
+	int RelX = FloorC(a_Location.x) - a_Chunk.GetPosX() * cChunkDef::Width;
+	int RelY = FloorC(a_Location.y);
+	int RelZ = FloorC(a_Location.z) - a_Chunk.GetPosZ() * cChunkDef::Width;
+	if (
+			(a_Chunk.GetSkyLight(RelX, RelY, RelZ) == 15) &&             // In the daylight
+			(a_Chunk.GetBlock(RelX, RelY, RelZ) != E_BLOCK_SOULSAND) &&  // Not on soulsand
+			(GetWorld()->GetTimeOfDay() < (12000 + 1000)) &&             // It is nighttime
+			!IsOnFire() &&                                               // Not already burning
+			GetWorld()->IsWeatherSunnyAt(POSX_TOINT, POSZ_TOINT)         // Not raining
```
madmaxoft: Too much indent here, use one level less

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+#ifndef COMPILING_PATHFIND_DEBUGGER
+	/* MCServer headers */
+	#include "../World.h"
+	#include "../Chunk.h"
+#endif
+
+
+#include "Path.h"
+
+#include <stdio.h>
```
madmaxoft: Get rid of stdio and use `LOGD` for debugging, instead of `printf`.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+
+
+#ifndef COMPILING_PATHFIND_DEBUGGER
+bool cPath::Item(cChunk * a_Chunk)  // returns FALSE if there's a solid or if we failed.
+{
+	int RelX = m_Item_CurrentBlock.x - a_Chunk->GetPosX() * cChunkDef::Width;
+	int RelZ = m_Item_CurrentBlock.z - a_Chunk->GetPosZ() * cChunkDef::Width;
+
+	// If it's setmode, Sets a block to cobblestone and exits, will be removed eventually.
```
madmaxoft: How about remove now?

src/Mobs/Monster.cpp
wiseoldman95: Done.

src/Mobs/Path.cpp
wiseoldman95: Done.

src/Mobs/Path.cpp
wiseoldman95: Perfect idea :P

wiseoldman95: Done.

src/Mobs/Path.cpp
```diff
@@ -0,0 +1,391 @@
...
+	{
+		return false;
+	}
+	BLOCKTYPE BlockType;
+	NIBBLETYPE BlockMeta;
+	a_Chunk->GetBlockTypeMeta(RelX, m_Item_CurrentBlock.y, RelZ, BlockType, BlockMeta);
+	return (!cBlockInfo::IsSolid(BlockType));
+
+	// TODO Maybe I should queue several blocks and call item() at once for all of them for better performance?
+	// I think Worktycho said each item() call needs 2 locks.
```
madmaxoft: There is an even better way:
Make `Step()` call the `cWorld::DoWithChunk()` function, then in the `Item()` function call the `SingleStep()` several times, passing the `cChunk` over to it for direct querying. Note that once you have a `cChunk` instance, you can simply get its neighboring chunks and thus query the neighborhood rather efficiently. I believe this would be such an improvement that it would invalidate the need for a cache completely.

wiseoldman95: Yes, much better indeed. But is this performance critical or can I do it later? doWithChunk will be called at most 100 times per 2 seconds for each mob, and typically much less than that. If it's critical, I'll do it now, otherwise, I'll add it to my to do list.

madmaxoft: I guess we can do without, but make it top priority on that list :)

wiseoldman95: Alright :)

src/Server.cpp
```diff
@@ -21,6 +21,8 @@
 #include "IniFile.h"
 #include "Vector3.h"
 
+
+
```
madmaxoft: Get rid of these when rebasing

madmaxoft:
I think this is good enough, squash them and we can merge.

wiseoldman95:
Squash complete.

SphinxC0re:
DO IT!!! I love this!! :dancer:

wiseoldman95:
Hold on, rebasing again and applying your last comment.

src/Server.cpp
wiseoldman95: Done.

wiseoldman95:
Second rebase complete, OK to merge.

wiseoldman95:
Let's tease @SphinxC0re by waiting for next week :stuck_out_tongue:

SphinxC0re:
AWWWWWWWWWWWWWWWWWWWWWWWW!!!

madmaxoft:
I'll wait for the builds to confirm.

tonibm19:
Yay, correct pahfinding! :D

wiseoldman95:
Is there any particular reason why both Travis and Appveyor are used, can a build fail on only one of them? Appveyor seems much slower, btw.

madmaxoft:
Travis compiles using gcc and clang on Linux, Appveyor compiles using MSVC2013 on Windows.

SphinxC0re:
WOOOHOHOHOHOOOO!!!!

tonibm19:
:D

src/Mobs/Path.h
```diff
@@ -0,0 +1,161 @@
...
+	/** Creates a pathfinder instance. A Mob will probably need a single pathfinder instance for its entire life.
+
+	Note that if you have a man-sized mob (1x1x2, zombies, etc), you are advised to call this function without parameters
+	because the declaration might change in later version of the pathFinder, and a parameter-less call always assumes a man-sized mob.
+
+	If your mob is not man-sized, you are advised to use cPath(width, height), this would be compatible with future versions,
+	but please be aware that as of now those parameters will be ignored and your mob will be assumed to be man sized.
+
+	@param a_BoundingBoxWidth the character's boundingbox width in blocks. Currently the parameter is ignored and 1 is assumed.
+	@param a_BoundingBoxHeight the character's boundingbox width in blocks. Currently the parameter is ignored and 2 is assumed.
```
Woazboat: Afaik the standard human height is ~1.85 blocks
http://i.imgur.com/5Ky2G.png

worktycho: From the point of view of the pathfinder thats 2 blocks high as you cannot easily create a space that's between 1.5 and 2 blocks high.

Woazboat: Easily done with trapdoors or carpet on the floor
http://minecraft.gamepedia.com/Solid_block#Heights

worktycho: Not when the pathfinder considers all blocks as one block high.

worktycho: In order for this to matter, MCServer would need to have a more detailed specification for block solidity than Full, Partial, or Empty.

Woazboat: Well, in that case...

wiseoldman95: The PathFinder currently does not deal with partials and assumes they're full solids.

madmaxoft: I believe vanilla doesn't consider partials when pathfinding either. Check the "mob-safe doors" video that shows how to take advantage of that:
https://www.youtube.com/watch?v=sQWHkUUqVdU

wiseoldman95: Speaking of that, I've been thinking that "cheating" could be better:
I could assume that all bounding box dimensions are integers, it would work with almost all use cases, then I could hack my way around the exotics, e.g. completely ignoring open doors and pushing all the waypoints half a block up when a ground-facing slab is detected.

It keeps the PathFinder codebase much simpler and is much easier to implement than real non-int bounding boxes.
What do you think?

wiseoldman95: Xoft, it seems that the Vanilla's PathFinding indeed doesn't take that into account, but I guess the Vanilla's version of MoveToPosition works better and it doesn't have trouble "rounding corners" and walking on e.g. halfslabs. perhaps I only need to improve that.

madmaxoft:
I think that's a much better way to go.

PR #1921, opened 2015-04-30
Revised README file

BijanKarrobi:
Updated the file so it is correct with the current details


madmaxoft:
You've lost most of the links

PR #1922, opened 2015-04-30
cBlockArea extensions

### LINKED ISSUE #1915: Add cBlockArea:GetNonAirCropRelCoords()
Add a new API function that returns the minimum and maximum coords in each direction for the first non-air block in each direction.

It is possible to calculate this in Lua, but is quite slow, a native implementation would be much better.

madmaxoft:
Adds `cBlockArea:GetNonAirCropRelCoords` API function, and `cBlockArea.msSimpleCompare` merge strategy.

Fixes #1915.


NiLSPACE:
Nice, with these we can fix some of WorldEdit's issues as well.

PR #1924, opened 2015-04-30
Added cBlockArea:CountNonAirBlocks API function.

madmaxoft:


worktycho:
Why size_t?

madmaxoft:
Why not? Is there a type that is more suited?

worktycho:
int? Its not like this needs to be a 64 bit type on 64 bit systems.

madmaxoft:
Just like the amount of data in the packet buffer is never gonna need more than 32-bits, yet still we use `size_t` for it.
Also, unsigned.

worktycho:
Packet buffer code interacts with sizeof a lot so it make sense to have the same type as sizeof(). size_t is for size's of c++ objects. This is a count.

PR #1925, opened 2015-05-01
Pathfinder optimization

SafwatHalaby:
Got rid of DoWithChunk, an order of magnitude less locks.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -1,12 +1,9 @@
 #include "Globals.h"
-#ifndef COMPILING_PATHFIND_DEBUGGER
-	/* MCServer headers */
-	#include "../World.h"
-	#include "../Chunk.h"
-#endif
 
 #include <cmath>
+
 #include "Path.h"
+#include "../Chunk.h"
 
 #define DISTANCE_MANHATTAN 0  // 1: More speed, a bit less accuracy 0: Max accuracy, less speed.
 #define HEURISTICS_ONLY 0  // 1: Much more speed, much less accurate.
@@ -38,18 +35,17 @@ bool compareHeuristics::operator()(cPathCell * & a_Cell1, cPathCell * & a_Cell2)
 
 /* cPath implementation */
 cPath::cPath(
-	cWorld * a_World,
+	cChunk * a_Chunk,
 	const Vector3d & a_StartingPoint, const Vector3d & a_EndingPoint, int a_MaxSteps,
 	double a_BoundingBoxWidth, double a_BoundingBoxHeight,
 	int a_MaxUp, int a_MaxDown
 )
 {
+	ASSERT(m_Chunk != nullptr);
 	// TODO: if src not walkable OR dest not walkable, then abort.
 	// Borrow a new "isWalkable" from ProcessIfWalkable, make ProcessIfWalkable also call isWalkable
 
-	m_World = a_World;
-	// m_World = cRoot::Get()->GetDefaultWorld();
-
+	m_Chunk = a_Chunk;
 	m_Source = a_StartingPoint.Floor();
 	m_Destination = a_EndingPoint.Floor();
 
@@ -65,6 +61,7 @@ cPath::cPath(
 	m_PointCount = 0;
 
 	ProcessCell(GetCell(a_StartingPoint), nullptr, 0);
+	m_Chunk = nullptr;
 }
 
 
@@ -83,8 +80,10 @@ cPath::~cPath()
 
 
 
-ePathFinderStatus cPath::Step()
+ePathFinderStatus cPath::Step(cChunk * a_Chunk)
 {
+	m_Chunk = a_Chunk;
+	ASSERT(m_Chunk != nullptr);
 	if (m_Status != ePathFinderStatus::CALCULATING)
 	{
 		return m_Status;
@@ -106,6 +105,7 @@ ePathFinderStatus cPath::Step()
 			}
 		}
 	}
+	m_Chunk = nullptr;
 	return m_Status;
 }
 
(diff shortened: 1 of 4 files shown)


madmaxoft:
I think the `cPath` constructor doesn't need the chunk either, only the `Step` function does, so how about losing the `m_Chunk` member variable, too, and propagating the chunk via parameters?

wiseoldman95:
But the cPath constructor performs the "initializer step" and checks the starting position and ending positions, it needs interactions with the world.

Regarding propagating via parameters: It's technically possible, but we'll needlessly pass that parameter everywhere, I find that a bit dirty, don't you think? 
Also, if the old chunk must be discarded when GetNeighborChunk is called, it is impossible to use parameters only.

wiseoldman95:
I could move the "initializer step" to the step function and then add:
`if (ThisIsTheFirstStep) { PerformInitializerStep(); return; }`
But that's an extra if, 20 times per sec per mob :P

wiseoldman95:
Awaiting comments / merge.
I'll rebase it when you want to merge.

wiseoldman95:
Is it legal to do it the other way around: only passing the chunk via the constructor and using that chunk till the path calculation ends?

wiseoldman95:
This isn't exactly what you asked for, but I feel it's cleaner. Tell me what you think.

madmaxoft:
No, that won't work. If the chunk gets unloaded between the constructor call and the path finish, this will fail horribly. You need to pass a valid chunk in each step.

I didn't see the initialization step in the constructor, that makes it necessary to pass the chunk there as well.

It isn't actually necessary to get rid of the "old" chunk when moving to a neighbor. The only time you have to consider your chunks invalid is when you exit out of the Tick function - at that moment every `cChunk` instance is in danger of being unloaded.

wiseoldman95:
Got it. Is it ok to keep the chunk as a member function?

wiseoldman95:
Rebased.

madmaxoft:
I guess so. Perhaps add a comment to its declaration, say that it's valid only within the `Step` call.

wiseoldman95:
Awaiting comments / Rebase and merge.

wiseoldman95:
I'd like to fix the swimming / fence issue by editing Issolid, I'm going to include it in this PR because otherwise a merge conflict will occur. To avoid confusion I'll rebase and name the commit "Issolid rewritten" or something like that.

NiLSPACE:
If you change the IsSolid function you might break other code that uses it.

wiseoldman95:
No, it's my PathFinder's Issolid, a different function.

wiseoldman95:
Fixed fence issue, swimming for later, awaiting comments / rebase and merge

Seadragon91:
I think for the fence, you need to also check for E_BLOCK_FENCE_GATE.

jan64:
Nether brick, spruce, birch, jungle, dark_oak, acacia.
Fence gate comes in all those flavors too.

Seadragon91:
That doesn't matter, we don't need to check the data value of the fence. Only the block type itself.

wiseoldman95:
@Seadragon91 , you are right.
I observed that my fence fix isn't perfect, I'll do it thoroughly later in another pull request.

wiseoldman95:
Fencegate added.

src/Mobs/Path.cpp
```diff
@@ -113,15 +109,24 @@ ePathFinderStatus cPath::Step()
...
+		return true;
+	}
+	BLOCKTYPE BlockType;
+	NIBBLETYPE BlockMeta;
+	int RelX = a_Location.x - m_Chunk->GetPosX() * cChunkDef::Width;
+	int RelZ = a_Location.z - m_Chunk->GetPosZ() * cChunkDef::Width;
+	m_Chunk->GetBlockTypeMeta(RelX, a_Location.y, RelZ, BlockType, BlockMeta);
+	if ((BlockType == E_BLOCK_FENCE) || (BlockType == E_BLOCK_FENCE_GATE))
+	{
+		GetCell(a_Location+Vector3d(0, 1, 0))->m_IsSolid = true;  // Mobs will always think that the fence is 2 blocks high and therefore won't jump over.
```
madmaxoft: spaces

src/Mobs/Path.cpp
```diff
@@ -113,15 +109,24 @@ ePathFinderStatus cPath::Step()
...
+	if (!m_Chunk->IsValid())
+	{
+		return true;
+	}
+	BLOCKTYPE BlockType;
+	NIBBLETYPE BlockMeta;
+	int RelX = a_Location.x - m_Chunk->GetPosX() * cChunkDef::Width;
+	int RelZ = a_Location.z - m_Chunk->GetPosZ() * cChunkDef::Width;
+	m_Chunk->GetBlockTypeMeta(RelX, a_Location.y, RelZ, BlockType, BlockMeta);
+	if ((BlockType == E_BLOCK_FENCE) || (BlockType == E_BLOCK_FENCE_GATE))
```
madmaxoft: Rather than depending on "someone else" to set the solidity for you, how about checking the block below yourself instead?

madmaxoft: Also, fence gate is completely non-solid when open. You need to check the meta value as well.

wiseoldman95: "Rather than depending on "someone else" to set the solidity for you, how about checking the block below yourself instead?", I just wanted to squeeze some performance by not calling GetCell unless really needed. You may think I'm micro-optimizing, but this function is the most-called function in the PathFinder (and possibly in the whole game, given enough mobs).

wiseoldman95: Also, this is a temporary, it doesn't handle all situations well.
Regarding meta, you're right, I'll  change that.

wiseoldman95: How do I check if it's closed?

jan64:
@Seadragon91 Different kinds of fences are defined as different blocks, they don't only differ in data values.

wiseoldman95:
@jan64 Looks like you're right, BlockID has lots . I really don't want to mix pull requests so let's leave this at the current imperfect state, I'll need to take a different approach in dealing with fences anyways.

wiseoldman95:
Any non-fence related comments? I'm not pushing the issue aside, I just don't want to invest time on something I'm going to rewrite anyways.

src/Mobs/Path.cpp
madmaxoft: A closed fence-gate has a different meta than an open one.

As for the fences, consider the following situation:

Your code would allow the zombie to pathfind through to get me, simply because the cell on top of the fence is processed earlier than the cell with the fence.

wiseoldman95: Is that a bad thing? :P

wiseoldman95: Do they not do that in Vanilla?

madmaxoft:
I personally would reset `m_Chunk` to `nullptr` at the end of the constructor and the `Step` function, and add `ASSERT(m_Chunk != nullptr)` to each function accessing the chunk, so that any mis-use is detected early.

src/Mobs/Path.cpp
madmaxoft: They can't get through, there's a plank block above the fence so the hole is actually only 1.5 blocks high.

wiseoldman95:
How about `ASSERT(m_Chunk = nullptr || 1==1);` or something like that? to get rid of setting the NULL needlessly in production?

madmaxoft:
Not worth it. It took me a whole minute to understand what that assert is doing (that it's setting, not comparing, the m_Chunk), so it's not good for code clarity. Just put the assignment there, it's not worth micro-optimizing this much.

wiseoldman95:
Got some equavilant which is clearer?

wiseoldman95:
Perhaps there should be a Macro DoInDebugModeonly or something like that.

wiseoldman95:
Ok to squash?

madmaxoft:
Fix the parentheses and squash.

wiseoldman95:
Squashed and fixed extra `(`.

SafwatHalaby:
Note that my Github username has been changed from WiseOldMan95 to SafwatHalaby.

PR #1927, opened 2015-05-01
AI - Much better burning/daylight behavior

SafwatHalaby:
Mobs will stand their ground if it means they won't burn there and they'll burn if they chase you.

Mobs WONT stand their ground if you are in a shade and they are in a shade and they can walk to you.

Mobs WONT stand their ground if  they are going to burn to death anyways.

Mobs will attack regardless of the light condition if they were provoked recently.

No more zombie freezing at twilight.

When very close to them, they sometimes step out of the shade accidentally, this was something I was planning to fix but it's actually a neat feature :wink:

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -142,7 +142,7 @@ void cMonster::TickPathFinding()
 
 		case ePathFinderStatus::PATH_NOT_FOUND:
 		{
-			FinishPathFinding();
+			ResetPathFinding();
 			break;
 		}
 
@@ -164,7 +164,7 @@ void cMonster::TickPathFinding()
 			}
 			if (m_Path->IsLastPoint())
 			{
-				FinishPathFinding();
+				ResetPathFinding();
 			}
 			break;
 
@@ -191,7 +191,7 @@ void cMonster::MoveToPosition(const Vector3d & a_Position)
 void cMonster::StopMovingToPosition()
 {
 	m_bMovingToDestination = false;
-	FinishPathFinding();
+	ResetPathFinding();
 }
 
 
@@ -209,7 +209,7 @@ bool cMonster::IsCoordinateInTraversedList(Vector3i a_Coords)
 
 /* No one should call this except the pathfinder orthe monster tick or StopMovingToPosition.
 Resets the pathfinder, usually starting a brand new path, unless called from StopMovingToPosition. */
-void cMonster::FinishPathFinding(void)
+void cMonster::ResetPathFinding(void)
 {
 	if (m_Path != nullptr)
 	{
@@ -254,6 +254,7 @@ bool cMonster::ReachedFinalDestination()
 void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	super::Tick(a_Dt, a_Chunk);
+	GET_AND_VERIFY_CURRENT_CHUNK(Chunk, POSX_TOINT, POSZ_TOINT);
 
 	if (m_Health <= 0)
 	{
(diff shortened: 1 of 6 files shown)


src/Mobs/Monster.cpp
```diff
@@ -300,7 +300,12 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		{
 			if (--m_GiveUpCounter == 0)
 			{
-				FinishPathFinding();
+				FinishPathFinding();  // Not to be confused with StopMovingToPosition, this just discards the current path and calculates another.
```
madmaxoft: How about calling it `RestartPathFinding`, then? Or something along these lines.

src/Mobs/Monster.cpp
```diff
@@ -1129,7 +1134,7 @@ bool cMonster::WouldBurnAt(Vector3d a_Location, cChunk & a_Chunk)
 		(a_Chunk.GetSkyLight(RelX, RelY, RelZ) == 15) &&             // In the daylight
 		(a_Chunk.GetBlock(RelX, RelY, RelZ) != E_BLOCK_SOULSAND) &&  // Not on soulsand
 		(GetWorld()->GetTimeOfDay() < (12000 + 1000)) &&             // It is nighttime
-		!IsOnFire() &&                                               // Not already burning
+
```
madmaxoft: Empty line inside a condition? That doesn't look too good.

src/Mobs/Monster.cpp
wiseoldman95: It's not always used for restarting though, I guess `ResetPathFinding` is best.

wiseoldman95: Got a name which means "Either finish or restart, depending on whether we're done or not"?

src/Mobs/Monster.cpp
wiseoldman95: Done and rebased.

wiseoldman95:
Standing by for comments / Rebase and merge.

src/Mobs/Monster.cpp
```diff
@@ -1129,7 +1135,6 @@ bool cMonster::WouldBurnAt(Vector3d a_Location, cChunk & a_Chunk)
 		(a_Chunk.GetSkyLight(RelX, RelY, RelZ) == 15) &&             // In the daylight
```
tigerw: Unrelated, but the usage of chunks are causing issues, since I believe mobs can move to a different chunk (invalidating the parameter) in super::Tick, and then this function is called.

Fix is to use the GET_AND_VERIFY_CURRENT_CHUNK macro and place it under super::Tick in this file, I think.

wiseoldman95: Super already has this line: `GET_AND_VERIFY_CURRENT_CHUNK(NextChunk, POSX_TOINT, POSZ_TOINT)`, wouldn't that solve it?

wiseoldman95: I fixed it in this particular case, but this could happen ANYWHERE, lots of stuff, including many mob types, inhert the tick() function.

tigerw: Yeah. We have many bugs :P

wiseoldman95:
Squash?

src/Mobs/Monster.cpp
wiseoldman95: We could prevent this with a macro, instead of calling super::tick, make two macros such as:
`SUPER_TICK_AT_START();`
and 
`SUPER_TICK_END()`
The former first does the chunk update thing, and only then calls super, while the latter just calls super.

wiseoldman95: The first one is used when you want to call SUPER at the begining of your tick, the latter when you want to call it at the end of your tick.

wiseoldman95: Or have a single one called Super_safeTick or something

wiseoldman95:
Squashed.

src/Mobs/Monster.cpp
tigerw: Macros everywhere! The main thing would be to remember to highlight to contributors that it needs to be done if they are using the chunk parameter. I suppose cMonster::Tick could call GET_AND_VERIFY... before it does work and after to account for all situations, but for some mobs the extra processing is needless since they don't override.

wiseoldman95: Alright, but that'll be in another PR. This one is ready for merging.

SafwatHalaby:
To avoid confusion: My username has been changed from WiseOldMan95 to SafwatHalaby.

PR #1930, opened 2015-05-02
AI - Improved Mob Jumping

SafwatHalaby:
Mobs now have a decent jump ability, and they go up blocks much more efficiently.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -80,6 +80,7 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_GiveUpCounter(0)
 	, m_bMovingToDestination(false)
 	, m_LastGroundHeight(POSY_TOINT)
+	, m_JumpCoolDown(0)
 	, m_IdleInterval(0)
 	, m_DestroyTimer(0)
 	, m_MobType(a_MobType)
@@ -286,12 +287,21 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	{
 		if (m_bOnGround)
 		{
-			if (DoesPosYRequireJump((int)floor(m_Destination.y)))
+			if (m_JumpCoolDown == 0)
 			{
-				m_bOnGround = false;
-
-				// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
-				AddPosY(1.2);  // Jump!!
+				if (DoesPosYRequireJump(static_cast<int>(floor(m_Destination.y))))
+				{
+					m_bOnGround = false;
+					m_JumpCoolDown = 20;
+					// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
+					AddPosY(1.6);  // Jump!!
+					SetSpeedX(3.2 * (m_Destination.x - GetPosition().x));  // Move forward in a preset speed.
+					SetSpeedZ(3.2 * (m_Destination.z - GetPosition().z));  // The numbers were picked based on trial and error and 1.6 and 3.2 are perfect.
+				}
+			}
+			else
+			{
+				--m_JumpCoolDown;
 			}
 		}
 
--- a/src/Mobs/Monster.h
+++ b/src/Mobs/Monster.h
@@ -217,6 +217,7 @@ protected:
 
 	virtual void HandleFalling(void);
 	int m_LastGroundHeight;
+	int m_JumpCoolDown;
 
 	/* =========================== */
 


src/Mobs/Monster.h
```diff
@@ -217,6 +217,7 @@ class cMonster :
 
 	virtual void HandleFalling(void);
 	int m_LastGroundHeight;
+	int m_JumpCoolDown;
```
tigerw: unsigned int?

tigerw:
Still have to fix collision detection though.

src/Mobs/Monster.cpp
```diff
@@ -286,12 +287,18 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 				m_bOnGround = false;
-
+				m_JumpCoolDown = 20;
 				// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
-				AddPosY(1.2);  // Jump!!
+				AddPosY(1.6);  // Jump!!
+				SetSpeedX(3.2 * (m_Destination.x - GetPosition().x));  // Move forward in a preset speed.
+				SetSpeedZ(3.2 * (m_Destination.z - GetPosition().z));  // The numbers were picked based on trial and error and 1.6 and 3.2 are perfect.
+			}
+			else if (m_JumpCoolDown > 0)
```
NiLSPACE: Is this check needed? If it isn't `0` we can be sure it's above it right?

SafwatHalaby: Good question, and no, we can't, with your proposal:
If m_JumpCoolDown is 0 and DoesPosYRequireJump returns 0, we'll get an m_JumpCoolDown of -1.

src/Mobs/Monster.h
SafwatHalaby: Why?

SafwatHalaby: It's a value which is between 20 and 0, I don't see the merit of an unsigned int here. I could even use an int_8 (perhaps I should?)

src/Mobs/Monster.cpp
SafwatHalaby: You're right though, the code is confusing, I'll push a little improvement.

src/Mobs/Monster.h
madmaxoft: No, use a regular `int` or `unsigned` for normal stuff. Architectures such as ARM prefer ints to bytes.

src/Mobs/Monster.cpp
SafwatHalaby: Turns out I didn't need the extra check if the code is written a bit differently, done.

src/Mobs/Monster.cpp
```diff
@@ -286,12 +287,21 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
-				// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
-				AddPosY(1.2);  // Jump!!
+				if (DoesPosYRequireJump((int)floor(m_Destination.y)))
+				{
+						m_bOnGround = false;
+						m_JumpCoolDown = 20;
+						// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
+						AddPosY(1.6);  // Jump!!
+						SetSpeedX(3.2 * (m_Destination.x - GetPosition().x));  // Move forward in a preset speed.
+						SetSpeedZ(3.2 * (m_Destination.z - GetPosition().z));  // The numbers were picked based on trial and error and 1.6 and 3.2 are perfect.
```
madmaxoft: Too much indent here.

SafwatHalaby:
Fixed and squashed.

src/Mobs/Monster.h
SafwatHalaby: No need for unsigned then.

src/Mobs/Monster.cpp
```diff
@@ -286,12 +287,21 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 		if (m_bOnGround)
 		{
-			if (DoesPosYRequireJump((int)floor(m_Destination.y)))
+			if (m_JumpCoolDown == 0)
 			{
-				m_bOnGround = false;
-
-				// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
-				AddPosY(1.2);  // Jump!!
+				if (DoesPosYRequireJump((int)floor(m_Destination.y)))
```
madmaxoft: Use a `static_cast<int>(...)`
We're trying to get rid of the C-style casts.

SafwatHalaby:
Looks good to me. Squash?

madmaxoft:
Yup

SafwatHalaby:
Squashed, feel free to merge.

SafwatHalaby:
This is offtopic but: Aren't the CI's stupid? Why do they have to build it all from scratch every time? Can't they cache their object files or something?

madmaxoft:
You want them to check that you haven't forgotten any files, or that there aren't any unwanted dependencies.

PR #1932, opened 2015-05-03
PathFinder - Fixed Howaner's bug

SafwatHalaby:
Fixes [this](http://forum.mc-server.org/showthread.php?tid=1571&pid=19891#pid19891).


madmaxoft:
I'd leave the assert at the top of the function, but check the `a_Chunk` instead. Or even better, pass the chunk as a reference, that can never be `nullptr`; then assign as `m_Chunk = &a_Chunk`

SafwatHalaby:
Will do. Last time I tried this, there was a weird issue with Chunk references and the = operator. (But the ref might have been on the left hand, I'm not sure.)

madmaxoft:
This looks good, merge if it compiles.

SafwatHalaby:
Let me squash it first

SafwatHalaby:
OK to merge.

PR #1934, opened 2015-05-03
PathFinder - Fixed nullptr chunk bug

SafwatHalaby:
I fixed this:
```
        At Path.cpp. line 118:
    ASSERT(m_Chunk != nullptr);
    m_Chunk = m_Chunk->GetNeighborChunk(a_Location.x, a_Location.z);
        //Now m_Chunk is nullptr
    if (!m_Chunk->IsValid())   // CRASH
```
I still don't understand when does `m_Chunk->GetNeighborChunk` return a `nullptr `and I'd appreciate an explanation.
I didn't really test if it returned a `nullptr`, but it's unlikely that something's wrong with Chunk passed to the PathFinder by `tick(...)`, and this is the only other explanation I can think of.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -117,7 +117,7 @@ bool cPath::IsSolid(const Vector3d & a_Location)
 {
 	ASSERT(m_Chunk != nullptr);
 	m_Chunk = m_Chunk->GetNeighborChunk(a_Location.x, a_Location.z);
-	if (!m_Chunk->IsValid())
+	if ((m_Chunk == nullptr) || (!m_Chunk->IsValid()))
 	{
 		return true;
 	}


src/Mobs/Path.cpp
```diff
@@ -117,7 +117,7 @@ bool cPath::IsSolid(const Vector3d & a_Location)
 {
 	ASSERT(m_Chunk != nullptr);
 	m_Chunk = m_Chunk->GetNeighborChunk(a_Location.x, a_Location.z);
-	if (!m_Chunk->IsValid())
+	if (m_Chunk == nullptr || (!m_Chunk->IsValid()))
```
NiLSPACE: Use `((m_Chunk == nullptr) || (!...`

tigerw:
Ah, I fixed this in a branch I was working on :P

If the neighbouring chunk doesn't exist (not loaded or generated), it will return null.

src/Mobs/Path.cpp
SafwatHalaby: Done.

SafwatHalaby:
Oh, I see. MERGE IT BEFORE IT LAYS EGGS.

tigerw:
> MERGE IT BEFORE IT LAYS EGGS.

That's one way of putting it!

SafwatHalaby:
This compiled well, I suggest you merge it. I know it's included in your PR but I wanna fully check that one out first.

PR #1935, opened 2015-05-03
Entity improvements

### LINKED ISSUE #1217: 3 Knockback issues
I thought i would try to fix the knockback but there are 3 main issues that prevent me to do the knockback vanilla like. (i made it nearly vanilla like, issue 2 makes it not vanilla like...)

1. When entity are flying they go in a mode that i call "anti-gravity" because every move they do is much slower.

2. Vector3d has some serious issues when y is to small. I don't know the exact number but 4 is okay. Normally  x and y are "performed" at the same time but when y is too small the vector will first "perform" x and when it's done y.

(x and y represent the ingame coordinates)
vector3d_y
3. When the Vector is done the mob still has the velocity
example: I write for example: y= 0.5 and x = 2.
That means when i hit it, it flys 0.5 blocks up and 2 blocks forward.
When there is no ground it still has the velocity (speed) and it still flys forward even if it only should fly 2 blocks.

velocity

that paint skills, i know :D

### LINKED ISSUE #1933: Segfault
Stackstrace:
```
[15:34:55]   D:    | MCServer has encountered an error and needs to close
[15:34:55] Details | SIGSEGV: Segmentation fault
./MCServer(_Z15PrintStackTracev+0x23)[0x998153]
./MCServer(_Z14NonCtrlHandleri+0x67)[0x98be17]
/lib64/libc.so.6(+0x35650)[0x7fb9f68c8650]
./MCServer(_ZNK6cChunk7IsValidEv+0xc)[0x8d0a38]
./MCServer(_ZN5cPath7IsSolidERK7Vector3IdE+0x5c)[0xbce8a8]
./MCServer(_ZN5cPath7GetCellERK7Vector3IdE+0xa9)[0xbcf493]
./MCServer(_ZN5cPathC2ER6cChunkRK7Vector3IdES5_iddii+0x1db)[0xbce615]
./MCServer(_ZN8cMonster15TickPathFindingER6cChunk+0x245)[0xbc950b]
./MCServer(_ZN8cMonster4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEER6cChunk+0x3eb)[0xbc9c8f]
./MCServer(_ZN15cPassiveMonster4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEER6cChunk+0x41)[0xbce203]
./MCServer(_ZN6cWorld8TickMobsENSt6chrono8durationIlSt5ratioILl1ELl1000EEEE+0x329)[0x974539]
./MCServer(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x5a8)[0x9740e4]
./MCServer(_ZN6cWorld11cTickThread7ExecuteEv+0xad)[0x96bdf3]
./MCServer(_ZN9cIsThread9DoExecuteEv+0x2f)[0x98e369]
./MCServer(_ZNKSt7_Mem_fnIM9cIsThreadFvvEEclIIEvEEvPS0_DpOT_+0x65)[0x98f881]
./MCServer(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EE9_M_invokeIILm0EEEEvSt12_Index_tupleIIXspT_EEE+0x43)[0x98f7d1]
./MCServer(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EEclEv+0x1b)[0x98f6d9]
./MCServer(_ZNSt6thread5_ImplISt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS3_EEE6_M_runEv+0x1c)[0x98f672]
/lib64/libstdc++.so.6(+0xb51e0)[0x7fb9f72211e0]
/lib64/libpthread.so.0(+0x7df5)[0x7fb9f747adf5]
/lib64/libc.so.6(clone+0x6d)[0x7fb9f69891ad]
[15:34:55]   D:    | MCServer has encountered an error and needs to close
[15:34:55] Details | SIGABRT: Server self-terminated due to an internal fault
./MCServer(_Z15PrintStackTracev+0x23)[0x998153]
./MCServer(_Z14NonCtrlHandleri+0x9e)[0x98be4e]
/lib64/libc.so.6(+0x35650)[0x7fb9f68c8650]
/lib64/libc.so.6(gsignal+0x37)[0x7fb9f68c85d7]
/lib64/libc.so.6(abort+0x148)[0x7fb9f68c9cc8]
./MCServer(_Z14NonCtrlHandleri+0x6c)[0x98be1c]
/lib64/libc.so.6(+0x35650)[0x7fb9f68c8650]
./MCServer(_ZNK6cChunk7IsValidEv+0xc)[0x8d0a38]
./MCServer(_ZN5cPath7IsSolidERK7Vector3IdE+0x5c)[0xbce8a8]
./MCServer(_ZN5cPath7GetCellERK7Vector3IdE+0xa9)[0xbcf493]
./MCServer(_ZN5cPathC2ER6cChunkRK7Vector3IdES5_iddii+0x1db)[0xbce615]
./MCServer(_ZN8cMonster15TickPathFindingER6cChunk+0x245)[0xbc950b]
./MCServer(_ZN8cMonster4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEER6cChunk+0x3eb)[0xbc9c8f]
./MCServer(_ZN15cPassiveMonster4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEER6cChunk+0x41)[0xbce203]
./MCServer(_ZN6cWorld8TickMobsENSt6chrono8durationIlSt5ratioILl1ELl1000EEEE+0x329)[0x974539]
./MCServer(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x5a8)[0x9740e4]
./MCServer(_ZN6cWorld11cTickThread7ExecuteEv+0xad)[0x96bdf3]
./MCServer(_ZN9cIsThread9DoExecuteEv+0x2f)[0x98e369]
./MCServer(_ZNKSt7_Mem_fnIM9cIsThreadFvvEEclIIEvEEvPS0_DpOT_+0x65)[0x98f881]
./MCServer(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EE9_M_invokeIILm0EEEEvSt12_Index_tupleIIXspT_EEE+0x43)[0x98f7d1]
./MCServer(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EEclEv+0x1b)[0x98f6d9]
./MCServer(_ZNSt6thread5_ImplISt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS3_EEE6_M_runEv+0x1c)[0x98f672]
/lib64/libstdc++.so.6(+0xb51e0)[0x7fb9f72211e0]
/lib64/libpthread.so.0(+0x7df5)[0x7fb9f747adf5]
/lib64/libc.so.6(clone+0x6d)[0x7fb9f69891ad]
[1]    5762 abort      ./MCServer

### PR DESCRIPTION
**Pull request by tigerw**
* Pathfinder improvements
* Fixes #1217
* Fixes #1933

(It laid an egg - #1934)

Could @SafwatHalaby check this?

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -75,10 +75,8 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_EMPersonality(AGGRESSIVE)
 	, m_Target(nullptr)
 	, m_Path(nullptr)
-	, m_PathStatus(ePathFinderStatus::PATH_NOT_FOUND)
 	, m_IsFollowingPath(false)
 	, m_GiveUpCounter(0)
-	, m_bMovingToDestination(false)
 	, m_LastGroundHeight(POSY_TOINT)
 	, m_JumpCoolDown(0)
 	, m_IdleInterval(0)
@@ -122,54 +120,109 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
 
 
 
-void cMonster::TickPathFinding(cChunk & a_Chunk)
+bool cMonster::TickPathFinding(cChunk & a_Chunk)
 {
+	if (!m_IsFollowingPath)
+	{
+		return false;
+	}
+
+	if (ReachedFinalDestination())
+	{
+		StopMovingToPosition();
+		return false;
+	}
 
 	if (m_Path == nullptr)
 	{
-		Vector3d position = GetPosition();
-		Vector3d Dest = m_FinalDestination;
-
-		// Can someone explain why are these two NOT THE SAME???
-		// m_Path = new cPath(GetWorld(), GetPosition(), m_FinalDestination, 30);
-		m_Path = new cPath(a_Chunk, Vector3d(floor(position.x), floor(position.y), floor(position.z)), Vector3d(floor(Dest.x), floor(Dest.y), floor(Dest.z)), 20);
-
-
-		m_IsFollowingPath = false;
+		m_Path = new cPath(a_Chunk, GetPosition().Floor(), m_FinalDestination.Floor(), 20);
 	}
-	m_PathStatus = m_Path->Step(a_Chunk);
-	switch (m_PathStatus)
-	{
 
+	switch (m_Path->Step(a_Chunk))
+	{
 		case ePathFinderStatus::PATH_NOT_FOUND:
 		{
-			ResetPathFinding();
+			StopMovingToPosition();  // Give up pathfinding to that destination
 			break;
 		}
-
-
 		case ePathFinderStatus::CALCULATING:
 		{
-			m_Destination = GetPosition();
+			// Pathfinder needs more time
 			break;
 		}
-
-
 		case ePathFinderStatus::PATH_FOUND:
 		{
-			if (ReachedDestination() || !m_IsFollowingPath)
+			if (--m_GiveUpCounter == 0)
 			{
-				m_Destination = m_Path->GetNextPoint();
-				m_IsFollowingPath = true;
-				m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_Dest.
+				ResetPathFinding();  // Try to calculate a path again
+				return false;
 			}
-			if (m_Path->IsLastPoint())
+			else if (!m_Path->IsLastPoint() && (m_Path->IsFirstPoint() || ReachedDestination()))  // Have we arrived at the next cell, as denoted by m_Destination?
 			{
-				ResetPathFinding();
+				m_Destination = Vector3d(0.5, 0, 0.5) + m_Path->GetNextPoint();
+				m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_Destination.
 			}
-			break;
-
+			return true;
 		}
+	}
+
+	return false;
+}
+
+
+
+
+
+void cMonster::MoveToWayPoint(cChunk & a_Chunk)
+{
+	if (m_JumpCoolDown == 0)
+	{
+		// We're not moving and waypoint is above us, it means we are hitting something and we should jump.
+		if ((GetSpeedX() < 0.1) && (GetSpeedY() < 0.1) && DoesPosYRequireJump(FloorC(m_Destination.y)))
+		{
+			if (IsOnGround() || IsSwimming())
+			{
+				m_bOnGround = false;
+				m_JumpCoolDown = 20;
+				// TODO: Change to AddSpeedY once collision detection is fixed - currently, mobs will go into blocks attempting to jump without a teleport
+				AddPosY(1.6);  // Jump!!
+				SetSpeedX(3.2 * (m_Destination.x - GetPosition().x));  // Move forward in a preset speed.
+				SetSpeedZ(3.2 * (m_Destination.z - GetPosition().z));  // The numbers were picked based on trial and error and 1.6 and 3.2 are perfect.
+			}
+		}
+	}
+	else
+	{
+		--m_JumpCoolDown;
+	}
+
+	Vector3d Distance = m_Destination - GetPosition();
+	if ((Distance.x != 0) || (Distance.z != 0))
+	{
+		Distance.y = 0;
+		Distance.Normalize();
+
+		if (m_bOnGround)
+		{
+			Distance *= 2.5f;
+		}
+		else if (IsSwimming())
+		{
+			Distance *= 1.3f;
+		}
+		else
+		{
+			// Don't let the mob move too much if he's falling.
+			Distance *= 0.25f;
+		}
+		// Apply walk speed:
+		Distance *= m_RelativeWalkSpeed;
+		/* Reduced default speed.
+		Close to Vanilla, easier for mobs to follow m_Destinations, hence
+		better pathfinding. */
+		Distance *= 0.5;
+		AddSpeedX(Distance.x);
+		AddSpeedZ(Distance.z);
 	}
 }
 
@@ -177,12 +230,15 @@ void cMonster::TickPathFinding(cChunk & a_Chunk)
 
 
 
-/* Currently, the mob will only start moving to a new position after the position it is
-currently going to is reached. */
 void cMonster::MoveToPosition(const Vector3d & a_Position)
 {
-	m_FinalDestination = a_Position;
-	m_bMovingToDestination = true;
+	if ((m_FinalDestination - a_Position).Length() > 0.25)
+	{
+		ResetPathFinding();
+
+		m_FinalDestination = a_Position;
+		m_IsFollowingPath = true;
+	}
 }
 
 
@@ -191,7 +247,7 @@ void cMonster::MoveToPosition(const Vector3d & a_Position)
 
 void cMonster::StopMovingToPosition()
 {
-	m_bMovingToDestination = false;
+	m_IsFollowingPath = false;
 	ResetPathFinding();
 }
 
@@ -199,53 +255,13 @@ void cMonster::StopMovingToPosition()
 
 
 
-bool cMonster::IsCoordinateInTraversedList(Vector3i a_Coords)
-{
-	return (std::find(m_TraversedCoordinates.begin(), m_TraversedCoordinates.end(), a_Coords) != m_TraversedCoordinates.end());
-}
-
-
-
-
-
-/* No one should call this except the pathfinder orthe monster tick or StopMovingToPosition.
-Resets the pathfinder, usually starting a brand new path, unless called from StopMovingToPosition. */
 void cMonster::ResetPathFinding(void)
 {
 	if (m_Path != nullptr)
 	{
 		delete m_Path;
 		m_Path = nullptr;
-
 	}
-}
-
-
-
-
-
-bool cMonster::ReachedDestination()
-{
-	if ((m_Destination - GetPosition()).Length() < 0.5f)
-	{
-		return true;
-	}
-
-	return false;
-}
-
-
-
-
-
-bool cMonster::ReachedFinalDestination()
-{
-	if ((GetPosition() - m_FinalDestination).Length() <= m_AttackRange)
-	{
-		return true;
-	}
-
-	return false;
 }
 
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by SafwatHalaby** [comment]
Will do.

File: src/Mobs/Monster.cpp [hunk 1]
```diff
@@ -122,67 +120,68 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 	{
+		m_Path = new cPath(a_Chunk, GetPosition().Floor(), m_FinalDestination.Floor(), 200);
+	}
 
+	switch (m_Path->Step(a_Chunk))
+	{
 		case ePathFinderStatus::PATH_NOT_FOUND:
 		{
-			ResetPathFinding();
+			StopMovingToPosition();  // Try to calculate a path again
```
SafwatHalaby: Used RestPathFinding() instead

SafwatHalaby: Why?

SafwatHalaby: On a second thought, this is the right approach, it's just the comment that's misleading. It should be "Give up PathFinding to that destination".

tigerw: Changed.

SafwatHalaby:
All in all, looks well. I like it :)

I was planning to switch everything to `vector3i` except where something else would be better, you switched it all to `vector3i` and this forced you to move `vector(0.5...` out. We've got two choiced here.
1. Switching back m_PathPoints to Vector3d
2. Removing 0.5 and just adding MY_OFFSET_WIDTH/2 (not sure what's the variable name)

2 sounds better to me.

SafwatHalaby:
Regarding steps, I would keep a middle ground for now (Say 10 internal steps and 100 max steps)
Other than that, go ahead and merge, and thanks for the cleanup :+1:

src/Mobs/Monster.h
```diff
@@ -185,42 +182,40 @@ class cMonster :
...
-		return (
-			(a_PosY <= POSY_TOINT) ||
-			DoesPosYRequireJump(a_PosY)
-			);
-	}
+	/** Returns if the ultimate, final destination has been reached */
+	bool ReachedFinalDestination(void) { return ((m_FinalDestination - GetPosition()).SqrLength() < (m_AttackRange * m_AttackRange)); }
+
+	/** Returns if the intermediate waypoint of m_Destination has been reached */
+	bool ReachedDestination(void) { return ((m_Destination - GetPosition()).SqrLength() < 0.25); }
```
SafwatHalaby: Please put this back to 0.5, 0.25 is too precise and the mob will sometimes circle around a waypoint, insisting on landing exactly.

SafwatHalaby:
I've based a branch on this, it's best to wait for my PR and merge it instead to avoid conflicts.

SafwatHalaby:
I am not sure this is identical in functionality to the previous code, the zombies sometimes wait too long before starting to chase, and they still circle around themselves even after I set the ReachedDestination value to 0.5

SafwatHalaby:
> For the steps per call, I thought it was a little slow, so it was increased.

I am starting to suspect that this is because you introduced a bug. They sometimes take too long to start moving.

SafwatHalaby:
Alright, I've almost got this fixed, I suggest you do not modify `path.cpp, path.h, monster.cpp, monster.h` to avoid merge conflicts.

tigerw:
Thanks for the help :)

SafwatHalaby:
You're welcome :)
BTW, beefing up the max-steps is part of the trouble. You made the PathFinder too stubburn, so if for a fraction of a moment there was no path t o th player (e.g. when jumping), The pathfinder would keep scanning for 200 ticks. Hence freezing the zombie for 200 ticks.

src/Mobs/Monster.cpp
```diff
@@ -122,67 +120,68 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
+			if (--m_GiveUpCounter == 0)
 			{
-				m_Destination = m_Path->GetNextPoint();
-				m_IsFollowingPath = true;
-				m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_Dest.
+				ResetPathFinding();  // Try to calculate a path again
+				return false;
 			}
-			if (m_Path->IsLastPoint())
+			else if (!m_Path->IsLastPoint() && (m_Path->IsFirstPoint() || ReachedDestination()))  // Have we arrived at the next cell, as denoted by m_Destination?
```
SafwatHalaby: When reaching the las tpoint, this is never triggered, and the zombie circles the path for 20 ticks, until ResetPath is triggered. Hmm....

tigerw: On reaching the last point, ReachedFinalDest should be true above?

SafwatHalaby:
You should merge https://github.com/SafwatHalaby/MCServer/tree/water2 into this branch.

src/Mobs/Monster.cpp
SafwatHalaby:
```
			else if (m_Path->IsFirstPoint() || ReachedDestination())  // Have we arrived at the next cell, as denoted by m_Destination?
			{
				m_Destination = Vector3d(0.5, 0, 0.5) + m_Path->GetNextPoint();
				m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_Destination.
				if (m_Path->IsLastPoint())
				{
					ResetPathFinding();
					return true;
				}
			}
			return true;
```
You should merge https://github.com/SafwatHalaby/MCServer/tree/water2 into this branch, it has this fix, among things.

SafwatHalaby:
That branch is based on yours, and the following changes were made:
-cPath's max steps was set back to 20, this is to prevent delayed mob freezes when the path is not reachable.
-Swimming was added.
-Jumping perfected.
-Fixed IsLastPoint (Mobs circling around) bug.
-Added some extra functions to tidy tick().

SafwatHalaby:
Sorry for cramming seemingly unrelated things into one commit, I might not be able to work on this for the next few days and I wanted to finish swimming NOW, and I couldn't risk adding swimming to a separate master branch because @tigerw  overhauled the Pathfinder/Monsters and not basing my work on his would have caused collisions.

src/Mobs/Monster.cpp
SafwatHalaby: >On reaching the last point, ReachedFinalDest should be true above?

No. The last point is never fetched in your code. (Once we detect it's the last point, we never enter), so ReachedFinalDest is never true above, and we have to wait for 20 ticks till the mob navigation gives up.

SafwatHalaby:
...Or I could submit it as a PR. Your choice.

tigerw:
It's okay, it's sort of a generic pull request anyway.

Is there some way to pathfind to a challenging location without freezing the mob?

SafwatHalaby:
Currently? Increasing max CALCULATIONS_PER_STEP without increasing a_MaxSteps should do that. 

Right now the problem with high a_MaxSteps  is that if a path is not present, the mob will wait as long as MaxSteps, I should come up with a "abort early" mechanism, I'll think of it later. For now, never increase a_MaxSteps and keep it at 20.
You could try putting a CALCULATIONS_PER_STEP  of 40 or 60.

tigerw:
I'll try that, thanks!

SafwatHalaby:
Are you planning to merge /water2 into this branch? or should I open a PR?

tigerw:
I'll merge it.

SafwatHalaby:
I also did some minor change: removed `auto chunk` and put `cChunk * chunk` instead. I prefer explicit code, I hope you don't mind that.

src/Mobs/Monster.h
SafwatHalaby: Nevermind that comment ^

SafwatHalaby:
Is this going to be merged soon? I cannot make any changes to PathFinding without causing conflicts until this is merged.

src/Mobs/Monster.cpp
```diff
@@ -122,101 +120,124 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 
 
 
-/* No one should call this except the pathfinder orthe monster tick or StopMovingToPosition.
-Resets the pathfinder, usually starting a brand new path, unless called from StopMovingToPosition. */
-void cMonster::ResetPathFinding(void)
+void cMonster::MoveToPosition(const Vector3d & a_Position)
 {
-	if (m_Path != nullptr)
+	if ((m_FinalDestination - a_Position).Length() > 0.25)
```
SafwatHalaby: Is this an alternative to the `case PATH_FOUND` changes?
Looks like one, just making sure.

src/Mobs/Monster.cpp
```diff
@@ -122,101 +120,124 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 	{
-		delete m_Path;
-		m_Path = nullptr;
+		ResetPathFinding();
```
SafwatHalaby: This fixes path recalculation in a brute-force way, not really a good idea.

SafwatHalaby: Also, if we ever increase a_MaxSteps, the path calculation would never make it, ResetPathFinding would be called again and again before calculation is done.

src/Mobs/Monster.cpp
```diff
@@ -397,32 +347,29 @@ void cMonster::SetPitchAndYawFromDestination()
 	{
 		if (m_Target->IsPlayer())
 		{
-			FinalDestination.y = ((cPlayer *)m_Target)->GetStance();
+			FinalDestination.y = static_cast<cPlayer *>(m_Target)->GetStance() - 1;
```
SafwatHalaby: I'm not sure what this does, but the FinalDestination.y should be the the block where the "player's knees" are at, is this the case?

src/Mobs/Monster.cpp
```diff
@@ -122,101 +120,124 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
-			m_Destination = GetPosition();
+			// Pathfinder needs more time
 			break;
 		}
-
-
 		case ePathFinderStatus::PATH_FOUND:
 		{
-			if (ReachedDestination() || !m_IsFollowingPath)
+			if (--m_GiveUpCounter == 0)
```
SafwatHalaby: This wasn't fixed yet, there's a fix in my branch.
Or perhaps you fixed it in another way I didn't catch?

src/Mobs/Monster.cpp
SafwatHalaby: Actually, any path that requires more than 1 step to finish calculation will have this problem.

tigerw: Well, if the final destination has changed (i.e. player moved), what's the point of finishing a computation to the old position, where the target is no longer at?

src/Mobs/Monster.cpp
tigerw: It is set to where the target's eyes are, so that attacking mobs' heads will face the target.

src/Mobs/Monster.cpp
tigerw: Would you mean the mob circling issue? I think I've tried to fix it another way, if you have time, would you like to confirm this?

src/Mobs/Monster.cpp
SafwatHalaby: Did you test if that works? The pathfinder will not work if its destination is in the air.

src/Mobs/Monster.cpp
SafwatHalaby: At the very least, you should add a `if old position is new position, return`. 

But there's more:
Think of a very complex path to a target which is constantly moving; the mob will just stand there, because its path is being reset again and again before calculation is done. On the other hand, if it goes to the old pos, it all works fine. You're not feeling this with a big CALCULATIONS_PER_STEP, but as worktycho mentioned, that value might be turned into a compile-time dependent thing.

Currently, there's this weird effect of mobs going to your old position, I understand you tried to fix it, but this is not the right way, it's just working in the specific case. I intend to fix this by completely ignoring the pathfinder if there's a clear line of sight between the mob and the player. For now, I suggest you revert the changes to `MoveToPosition`.
If you've got a better suggestion, I'm all ears :)

src/Mobs/Monster.cpp
SafwatHalaby: Yes, I mean that issue. I'll test it later.

SafwatHalaby: You might have "fixed" it by over-calling resetPath in `moveToPosition`

src/Mobs/Monster.cpp
SafwatHalaby: Your PR is pretty good, I could take it from here and take care of those little things myself if you merge it. If it's currently pathing, you can merge it.

src/Mobs/Monster.cpp
SafwatHalaby: One alternative is this
```
MoveToPosition(bool OverrideOldPosition = true)
{
    if  (VOerrideOldPosition) ResetPathFinding();
    ....
}
```
Then, all normal uses will call `MoveToPosition()`, whose parameter is true by default, with the exception of the mob's pathfinding part, which calls `MoveToPosition(false)` to FORCE it to go to the old path. This , in combination with "Ignore pathfinder if there's a clear line of sight" looks like the ultimate solution. Someone who's using the mob's interface will get a proper `MoveToPosition`, while the PathFinding would also be happy.

tigerw: But not having a recalculation will also mean that if a player does move to a place where a different path is needed, for example, up a ladder where mobs can't follow, and there is no easy line-of-sight route, the mob will first go to the old position, realise it's wrong, and potentially have to go a long way back to a correct route if it's a complex maze and such.

Meanwhile, the player can simply move back down the ladder, and the mob will not realise, since its pathfinding isn't being reset. It will instead head down the route to get to where the player was when they headed up the ladder.

Perhaps a way to fix the 'mob stands there while path is calculated' problem is to have two Path variables. When a mob spots a player, it remains in the IDLE state, and tasks one cPath to pathfind to the player, whilst the other handles normal idling behaviour; random walking and such. If the player could be navigated to, the state is set to ATTACKING and the mob moves along the calculated route. If no route was found, the mob sets a variable indicating that the current target cannot be reached (continuing with idling behaviour), which is not cleared until the target moves position.

What would you suggest?

src/Mobs/Monster.cpp
tigerw: Thanks! I think the FinalDestination variable is badly named in that it is separate from m_FinalDestination - it's used purely in this function for mob rotation/pitch calculations.

src/Mobs/Monster.cpp
worktycho: I suggest that you follow down that route but rather than having an IDLE/ATTACKING path its more of a double buffer. So one pathfinder is calculating whilst the other is following, then when the second finishes you swap. This means that you're continuously recalculating.

tigerw: Okay, I suppose this PR can be merged in its current form and improvements done later?

SafwatHalaby: @worktycho , Yes, sorry if I was vague before, I said I will implement proper recalculation later, (possibly with double buffers), but for now, going to the old position and then correcting self is better than this hacky fix.

SafwatHalaby: By the way, swapping isn't as easy as it seems because the older path has your OLD destination. But it's possible,.

SafwatHalaby: One alternative to double buffers is having a re-calculation timer which is proportional to distance, the further you are from target, the less often you calculate.

SafwatHalaby: >Okay, I suppose this PR can be merged in its current form and improvements done later?

Other than the fact that you should remove resetPath from MoveToPosition, merging is fine by me.

SafwatHalaby:
Could you please merge it? I'm locked till you do.

bearbin:
I can merge it it you want it done now.

SafwatHalaby:
Thanks, but I'm not sure whether TigerW is still making changes or not.

tigerw:
I haven't changed anything, so merge if you want to :)

SafwatHalaby:
Ok, merge.

PR #1936, opened 2015-05-03
Crafting - Allow to craft torches with charcoals

Seadragon91:
Ignore the damage value from coal and this will allow to create torches with charcoals.


madmaxoft:
Nice catch. I knew there would be more of these left when I "strictified" the crafting recipe matcher.

PR #1938, opened 2015-05-04
Change visual protocol name to include MCServer

jan64:
http://forum.mc-server.org/showthread.php?tid=1909


NiLSPACE:
Does this actualy work? Did you try it?

jan64:
Given the simplicity of the change, and the fact that other server software uses this idea i believe this should work. 
But, as i don't have the setup necessary to compile mcserver from source atm, i didn't test this on a live server.

jan64:
I did test this with 1.8, and it seems to work.

PR #1940, opened 2015-05-05
Chmod +x CheckBasicStyle.lua

SafwatHalaby:
Allows the pre-commit hook to work.


SafwatHalaby:
Warning - this failed to compile in Appveyor, this is very strange since the change is pretty basic. But it might cause a chain reaction of commits not compiling.

SafwatHalaby:
Oh, perhaps it simply aborted because you merged early.

bearbin:
It aborts if you merge early, yes. I don't think there's really a reason that it even needs to be enabled - it gives too many false positives.

PR #1946, opened 2015-05-06
PathFinder - Couple of bugfixes

SafwatHalaby:
Couple of fixes, each in its own commit.

-Min recalculate interval is now half second (There was no minimum limit, it was CPU-wasteful)
-Fixed mobs not reaching a player when the player's bounding box is in the air (e.g. leaning with SHIFT)
-Fixed cChunk issues with swimming.


bearbin:
Is this tested and works?

If it is, LGTM.

SafwatHalaby:
I renamed the PR.
The changes are trivial, and I tested stuff for half an hour.
I'm pretty sure it's safe merge.

PR #1947, opened 2015-05-06
AI - Livestock escape fixed, water jumping fixed

SafwatHalaby:


SafwatHalaby:
With the exception of `bool cMonster::EnsureProperDestination(cChunk & a_Chunk)`, the changes are trivial.

I did my best to not misuse chunks in that function, but can anyone double-check that?

src/Mobs/Monster.cpp
```diff
@@ -252,6 +259,64 @@ void cMonster::MoveToWayPoint(cChunk & a_Chunk)
...
 
 
+bool cMonster::EnsureProperDestination(cChunk & a_Chunk)
+{
+	cChunk * Chunk = a_Chunk.GetNeighborChunk(m_FinalDestination.x, m_FinalDestination.z);
+	BLOCKTYPE BlockType;
+	NIBBLETYPE BlockMeta;
+	int RelX = m_FinalDestination.x - Chunk->GetPosX() * cChunkDef::Width;
+	int RelZ = m_FinalDestination.z - Chunk->GetPosZ() * cChunkDef::Width;
+	if ((Chunk != nullptr) && Chunk->IsValid())
```
madmaxoft: Use an early bail-out:
```cpp
if (!cond)
{
  return false;
}

// Do whatever the original if block did
```

src/Mobs/Monster.cpp
```diff
@@ -252,6 +259,64 @@ void cMonster::MoveToWayPoint(cChunk & a_Chunk)
...
+	{
+		// If destination in the air, go down to the lowest air block.
+		for (;;)
+		{
+			Chunk->GetBlockTypeMeta(RelX, m_FinalDestination.y - 1, RelZ, BlockType, BlockMeta);
+			if (cBlockInfo::IsSolid(BlockType))
+			{
+				break;
+			}
+			m_FinalDestination += Vector3d(0, -1, 0);
```
madmaxoft: I think it'd be better to use `m_FinalDestination.y += 1` instead, as that will not use a needless new object.

src/Mobs/Monster.cpp
```diff
@@ -252,6 +259,64 @@ void cMonster::MoveToWayPoint(cChunk & a_Chunk)
...
+			else if (cBlockInfo::IsSolid(BlockType))
+			{
+				InWater = false;
+			}
+			else
+			{
+				break;
+			}
+			m_FinalDestination += Vector3d(0, 1, 0);
+		}
```
madmaxoft: Both loops will fail in a world where the void below is accessible - if there is no solid block under the mob. Use `while (m_FinalDestination.y < cChunkDef::Height)`

src/Mobs/Monster.h
```diff
@@ -204,6 +204,7 @@ class cMonster :
 	*/
 	bool TickPathFinding(cChunk & a_Chunk);
 	void MoveToWayPoint(cChunk & a_Chunk);
+	bool EnsureProperDestination(cChunk & a_Chunk);
```
madmaxoft: Add doxy-comments about what the functions do.

src/Mobs/Monster.cpp
SafwatHalaby: Good point, I have lots of those all over the place, I'll submit a change to all of them in a different PR.

src/Mobs/Monster.cpp

src/Mobs/Monster.h
SafwatHalaby: Done.

src/Mobs/Monster.cpp

madmaxoft:
I'd have preferred a doxy-comment without empty lines, but an empty lines between the previous function and the next function's doxy-comment.

SafwatHalaby:
Done.

SafwatHalaby:
Appveyor finished compiling way back but it's stuck. It might have to do with the Github downtime that occured. Nevertheless, all CI's are successful.

SafwatHalaby:
Rebased to shake the CI.

SafwatHalaby:
Things are acting weird, re-opening a new branch.

SafwatHalaby:
#1950

PR #1949, opened 2015-05-06
AI - Better shade cover

SafwatHalaby:
Changed the conditions of when the AI should remain in shade (doesn't fix the mushroom issue). Specifically, made the zombies light-sane even when idle and not chasing player.

The change is pretty trivial, it is very unlikely to break anything.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -383,9 +383,9 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		2. I was not hurt by a player recently.
 		Then STOP. */
 		if (
-			m_BurnsInDaylight && ((m_TicksSinceLastDamaged == 100) || (m_EMState == IDLE)) &&
-			WouldBurnAt(m_NextWayPointPosition, *Chunk->GetNeighborChunk(FloorC(m_NextWayPointPosition.x), FloorC(m_NextWayPointPosition.z))) &&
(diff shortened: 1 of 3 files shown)


src/Mobs/Monster.cpp
```diff
@@ -314,7 +314,15 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	HandleDaylightBurning(*Chunk, WouldBurnAt(GetPosition(), *Chunk));
 	if (TickPathFinding(*Chunk))
 	{
-		if (m_BurnsInDaylight && WouldBurnAt(m_NextWayPointPosition, *Chunk->GetNeighborChunk(FloorC(m_NextWayPointPosition.x), FloorC(m_NextWayPointPosition.z))) && !IsOnFire() && (m_TicksSinceLastDamaged == 100))
+		/* If I burn in daylight, and I won't burn where I'm standing, and I'll burn in my next position, and at least one of those is true:
+		1. I am idle
+		2. I was not hurt by a player recently.
+		Then STOP. */
+		if (
+			m_BurnsInDaylight && ((m_TicksSinceLastDamaged == 100) || (m_EMState == IDLE)) &&
```
bearbin: Should this be >= 100 rather than == 100?

SafwatHalaby: m_TicksSinceLastDamaged maxes at100 and doesn't increase beyond that. But `>=` is better, just in case this is ever changed.

SafwatHalaby:
Merge?

PR #1952, opened 2015-05-06
Actually empty the open list

worktycho:


src/Mobs/Path.cpp
```diff
@@ -245,7 +245,7 @@ void cPath::FinishCalculation()
 	}
 
 	m_Map.clear();
-	m_OpenList.empty();
+	m_OpenList = std::priority_queue<cPathCell *, std::vector<cPathCell *>, compareHeuristics>{};
```
SafwatHalaby: Isn't it supposed to be `( )`?

worktycho: its a new c++11 syntax. It avoids issues with most vexing parse.

SafwatHalaby: Oh, thanks.

SafwatHalaby:
I'd say it's pretty trivial yet important, and should be merged right away.

PR #1954, opened 2015-05-06
More style checking.

madmaxoft:
Spaces around some operators are checked.

### DIFF
--- a/src/Entities/Minecart.cpp
+++ b/src/Entities/Minecart.cpp
@@ -905,7 +905,7 @@ bool cMinecart::TestEntityCollision(NIBBLETYPE a_RailMeta)
 			}
 
 			/* Check to which side the minecart is to be pushed.
-			Let's consider a z-x-coordinate system where the minecart is the center (0/0).
+			Let's consider a z-x-coordinate system where the minecart is the center (0, 0).
 			The minecart moves along the line x = -z, the perpendicular line to this is x = z.
 			In order to decide to which side the minecart is to be pushed, it must be checked on what side of the perpendicular line the pushing entity is located. */
 			if (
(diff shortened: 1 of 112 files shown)


madmaxoft:
Unfortunately the spaces-around-operators is quite difficult and had a few false positives, so I had to work around them. Spaces around minus cannot be checked at all, due to minus being used as a hyphen between words.

src/Entities/Minecart.cpp
```diff
@@ -954,7 +954,7 @@ bool cMinecart::TestEntityCollision(NIBBLETYPE a_RailMeta)
 			}
 
 			/* Check to which side the minecart is to be pushed.
-			Let's consider a z-x-coordinate system where the minecart is the center (0/0).
+			Let's consider a z-x-coordinate system where the minecart is the center (0 / 0).
```
bearbin: This should be (0,0) anyway?

bearbin:
There are a huge amount of false positives for "/", so many I don't think it's really worth it to check for spaces there.

madmaxoft:
It's not that bad, and I worked around all of them anyway.

tigerw:
Those are quite a lot of somewhat unnecessary changes; could you get the script to ignore lines that are comments?

madmaxoft:
Oh I don't mind the changes in the comments at all, all of those are good. The only thing I don't quite like is the `" HTTP/1.0" needing to be split into two strings. If you can write a Lua pattern that matches the others but not this, be my guest.

tigerw:
Though it's still an additional (comment) style for people to conform to, and not out of necessity either, but a side-effect from something else. Since comments should follow language, not code conventions, it would be better for the style checker to ignore them. @bearbin also recommends that similarly strings be ignored also, which also fixes your problem.

worktycho:
The problem with comments is that its difficult to write a pattern to check is something is in a comment. I could probably manage a perl regex to do it, but even that would be a mess. Its certainly problematic with patterns to avoid: `/* a b c */ a/a /* a */ ` getting ignored.

worktycho:
The regex is: `^(\/\*([^*]|\*[^/])*\*\/[^op]*)*(([^ ]op)|op[^ *])`  where op is the operator. I don't know how you'ld write that in lua patterns.

madmaxoft:
I've fixed the operator-space patterns, now they ignore everything near doublequotes. So I got rid of the unwanted changes.

worktycho:
This is being got by the broken master. Can you rebase on to the fix?

madmaxoft:
The Appveyor build has been stuck so I cancelled it, the other builds passed.

PR #1955, opened 2015-05-07
Fixed Magma cube sound and spawn size

Tri125:
Based from the Slime itself. 

I spawned a few of them to test the sound and realized the big one couldn't spawn. The size seems to be on par with vanilla Minecraft.


madmaxoft:
This looked good, why did you close? Squash the commits into a single one and it can be merged.

PR #1956, opened 2015-05-07
Fixed Magma cube sound and spawn size

Tri125:
Take two, first time working with Travis sorry about that.

Based from the Slime itself.

I spawned a few of them to test the sound and realized the big one couldn't spawn. The size now seems to be on par with vanilla Minecraft.

### DIFF
--- a/src/Mobs/MagmaCube.h
+++ b/src/Mobs/MagmaCube.h
@@ -26,7 +26,7 @@ public:
 	
 protected:
 
-	/// Size of the MagmaCube, 1 .. 3, with 1 being the smallest
+	/// Size of the MagmaCube, 1, 2 and 4, with 1 being the smallest
 	int m_Size;
 } ;
 
--- a/src/Mobs/MagmaCube.h
+++ b/src/Mobs/MagmaCube.h
@@ -22,7 +22,7 @@ public:
 
 	/** Returns the text describing the slime's size, as used by the client's resource subsystem for sounds.
 	Returns either "big" or "small". */
-	const AString GetSizeName(int a_Size) const;
+	static AString GetSizeName(int a_Size);
 	
 protected:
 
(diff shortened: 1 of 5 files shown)


SafwatHalaby:
In the future, you can use `src/CheckBasicStyle.lua` to check if your commits are styled correctly before pushing the commit, just execute that script and it'll notify you of any syntax errors.

Or even better, you can create a [pre-commit hook](http://forum.mc-server.org/showthread.php?tid=434&pid=19872#pid19872) which prevents you from committing as long as there are syntax errors, and instead asks you to fix them first.

src/Mobs/MagmaCube.h
```diff
@@ -19,10 +19,14 @@ class cMagmaCube :
 
 	virtual void GetDrops(cItems & a_Drops, cEntity * a_Killer = nullptr) override;
 	int GetSize(void) const { return m_Size; }
+
+	/** Returns the text describing the slime's size, as used by the client's resource subsystem for sounds.
+	Returns either "big" or "small". */
+	const AString GetSizeName(int a_Size) const;
```
madmaxoft: This could be a `static` function, since it doesn't use any members.

src/Mobs/MagmaCube.h
```diff
@@ -19,10 +19,14 @@ class cMagmaCube :
...
 	
 protected:
 
-	/// Size of the MagmaCube, 1 .. 3, with 1 being the smallest
+	/// Size of the MagmaCube, 1, 2 and 4, with 1 being the smallest
```
madmaxoft: Even in vanilla the size can be really any number, it just happens that vanilla spawns only sizes 1, 2 and 4.

src/Mobs/MagmaCube.h
Tri125: Would you like me to create a commit making GetSizeName of Slime and MagmaCube static?

madmaxoft: Yes, please.

madmaxoft:
The builds are failing due to an unrelated reason, we need to fix those first, then re-evaluate this PR.

worktycho:
I think I've fixed master, and I've restarted the build, so hopefully it will compile now.

SafwatHalaby:
Travis failed.

PR #1985, opened 2015-05-10
PathFinder - Less calculations per tick

SafwatHalaby:
People are complaining that the PathFinder lags when there are many mobs. This should fix it, at the cost of a bit worse PathFinding. On the long run, path recalculation should be tweaked so that it works good even with low calculations per tick.


SafwatHalaby:
This can be merged with testing because it's very trivial.

madmaxoft:
If you want this in testing, you should make a PR against the testing branch as well.

PR #1986, opened 2015-05-10
Testing - Less path calcs

SafwatHalaby:


SafwatHalaby:
No, no, no, bad, bad, bad. :P

madmaxoft:
You need to update your `testing` branch, then cherry-pick the commit you want there from `master` and finally submit a PR.

SafwatHalaby:
I already created a different commit with the same changes and merged it with `testing`, is that ok?

madmaxoft:
I guess it is for now, although we'll see what it does when we finally want to merge master to testing, it might create a conflict.

SafwatHalaby:
It's identical, I think git can handle it.

PR #1987, opened 2015-05-10
PathFinder - Less calcs per tick (testing)

SafwatHalaby:


SafwatHalaby:
I should have based my initial commit on "testing" to avoid the situation that just happened. Will do that next time.

PR #2001, opened 2015-05-11
Bindings cleanup

madmaxoft:
Removed unused stuff from ToLua, and added generating some glue code instead of manually written functions.


madmaxoft:
@tigerw I know you are against *.inc files, but I think this change is beneficial. Do you think you can survive this? It's rather unlikely that you'd ever need to use a debugger on those generated functions anyway, they're just glue code.

madmaxoft:
@worktycho you mentioned some code generator, would this method (piggybacking on the bindings generator) be sufficient for you? Just provide another function next to `OutputLuaStateHelpers` in BindingsProcessor.lua.

worktycho:
Not really. It would need another run of the parser as the methods generated aren't really connected to the lua bindings.

tigerw:
No objections here.

madmaxoft:
Ha, I was chasing the CI error in the CMake script, and it was in the CheckBasicStyle the whole time.

PR #2007, opened 2015-05-13
All in-game commands go through the OnExecuteCommand hook.

madmaxoft:
Everything that has a slash at the beginning, doesn't matter if it is a registered command or not, is sent through the OnExecuteCommand hook.


bearbin:
Could this code be simplified? 

Just always ask plugins if they want to handle commands if they begin with slash rather than having the call in two places with nested ifs?

madmaxoft:
It can't really, because the default return values are different for each case.

bearbin:
Just merge it then :P

PR #2010, opened 2015-05-14
Update World.cpp

TwiddleStick:
Fixed minor typos in comments, simply trying to understand the code base to fix issue regarding no sound when lighting is produced, no change to actual code.

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -1019,7 +1019,7 @@ void cWorld::TickWeather(float a_Dt)
 		// 0.5% chance per tick of thunderbolt
 		if (m_TickRand.randInt() % 199 == 0)
 		{
-			CastThunderbolt(0, 0, 0);  // TODO: find random possitions near players to cast thunderbolts.
+			CastThunderbolt(0, 0, 0);  // TODO: find random positions near players to cast thunderbolts.
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


bearbin:
This PR is ready to merge, but could you add your name to the CONTRIBUTORS.md file?

madmaxoft:
@bearbin It's not a requirement, it's an optional priviledge :)

bearbin:
It's just better from a copyright perspective to have a list of all the contributors. But this fix is probably so minor that it isn't entirely necessary.

TwiddleStick:
I'm not sure if this is the right place to ask but, looking over the code I have specifically where the lighting event/weather events are produced, however I'm confused how sound is added to specific events i.e. the rain sound when it starts raining, or when TNT explodes. It seems the functions broadcast the event and handle the specific thunderbolt given the coordinates, etc. Could someone give me a function where the sound is produced in an event so I could mabye better understand how sound is produced. I looked over the Noise but am not sure what its relation/or if there is one is to events. Thanks in advance! Also sorry if this is the wrong place to ask for advice.

worktycho:
@TwiddleStick The Forum http://forum.mc-server.org/index.php is usually the best place to ask. 

To answer your question, sounds are usually created when a event is created, for example in cWorld::DoExplosionAt. The broadcast functions are simply about broadcasting an event to the clients, and don't know anything about sounds, so that plugins can do things like silent explosions. cNoise has nothing to do with sound, and is actually a pseudo-random number generator.

zackp30:
@TwiddleStick the better place would be the forums, and sorry I don't know the answer to the question, I just saw this in my GH feed. :smile:

@bearbin `git log --format="%aN <%aE>" | sort -u` :smile:

bearbin:
@zackp30 Yes, you can do that, but it's entirely possible that the code will get separated from the git repo at some point - what if somebody redistributes the code as some sort of package. They still need the contributor info, but git can't provide it as it's not there.

zackp30:
@bearbin Ahh, fair enough.

PR #2018, opened 2015-05-14
Fixed possible crash in cProbabDistrib

NiLSPACE:
It could divide through 0 which causes a crash.

I especialy noticed this in the nether


worktycho:
Might it be worth replacing cProbabDistrib with std::discrete_distribution?

NiLSPACE:
Would that be able to produce the same result each time we regenerate a chunk?

worktycho:
It would depend on what the generator passed is. std::discrete_distribution is just a functor from a generator to a number. It can use anything that fulfills these requirements: http://en.cppreference.com/w/cpp/concept/UniformRandomNumberGenerator

PR #2019, opened 2015-05-14
Initial Implementation of command line arguments

worktycho:
I am posting this for design review, and have not tested it yet.

### DIFF
--- a/src/IniFile.cpp
+++ b/src/IniFile.cpp
@@ -931,7 +931,18 @@ AStringVector ReadUpgradeIniPorts(
 )
 {
 	// Read the regular value, but don't use the default (in order to detect missing value for upgrade):
-	AStringVector Ports = StringSplitAndTrim(a_Settings.GetValue(a_KeyName, a_PortsValueName), ";,");
+
+	AStringVector Ports;
+
+	for (auto pair : a_Settings.GetValues(a_KeyName))
+	{
+		if (pair.first != a_PortsValueName)
+		{
+			continue;
+		}
+		AStringVector temp = StringSplitAndTrim(pair.second, ";,");
+		Ports.insert(Ports.end(), temp.begin(), temp.end());
+	}
 
 	if (Ports.empty())
 	{
--- a/src/IniFile.cpp
+++ b/src/IniFile.cpp
@@ -49,6 +49,9 @@ cIniFile::cIniFile(void) :
 
 bool cIniFile::ReadFile(const AString & a_FileName, bool a_AllowExampleRedirect)
 {
+
+	m_Filename = a_FileName;
+
 	// Normally you would use ifstream, but the SGI CC compiler has
 	// a few bugs with ifstream. So ... fstream used.
 	fstream f;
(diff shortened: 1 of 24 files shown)


madmaxoft:
You commit under yet another email which is not linked to your account

src/IniFile.cpp
```diff
@@ -49,6 +49,9 @@ cIniFile::cIniFile(void) :
 
 bool cIniFile::ReadFile(const AString & a_FileName, bool a_AllowExampleRedirect)
 {
+
+	m_Filename = a_FileName;
+
```
madmaxoft: I specifically removed the filename storage from the class, so that each write operation needs to specify what filename to write to. Why are you adding it back in?

worktycho:
No its the same email as last time, I just forgot to change the email globally.

src/IniFile.cpp
worktycho: Because I need the cSettingsRepositoryInterface::Flush method to have the semantics of flush to backing storage, where backing storage may not be a file. Requiring this method to take a filename seriously messes up the semantics of cOverridesSettingsRepository::Flush().

worktycho:
This version should pass CI build.

src/IniFile.cpp
madmaxoft: And why do you need the entire repository to `Flush`, can't you `Write` the individual underlying INI files instead?

worktycho: Because to do so the object constructing the ini file needs to retain ownership of the inifile, complicating memory management.

worktycho:
Some how this has fallen foul of ADL in visual studio.

worktycho:
Now VS should compile.

worktycho:
Does anyone have any other comments before I start implementing the rest of the arguments?

src/IniFile.cpp
madmaxoft: I still don't like this and would prefer if you changed the pointer to shared and let the files be written explicitly. Because now there are two functions that dump the contents to a file, and their interaction is not exactly clear. If you call WriteFile(), will it overwrite the filename stored for Flush()?

worktycho: And I really don't like how much the code is dependent of Inifiles. What if we want to change the config file? The way I view WriteFile, is that it is a separate dump to this file, don't change the backing store.

madmaxoft:
Okay, have it your way. But you should squash all those "fix smth" commits into their previous commits, we don't really want those in the history (it makes git bisect ineffective, among other things).

And you should doxy-comment your code. Basically you could move the doxy-comments from `cIniFile`'s methods to their `cSettingsRepositoryInterface` declarations.

worktycho:
I've kept the three commits because the make_unique thing is a logically different change.

madmaxoft:
Yes, that's what I meant - keep the logically-separate commits, but squash the "fix" ones. I'd say this is good, do you want to merge or will you keep implementing stuff on top of this PR?

worktycho:
I don't mind. The rest of the changes should be trivial, but I don't want to guarantee that.

worktycho:
Can we merge now? Ports is a key feature.

bearbin:
Did you just fix the build?

worktycho:
Should be fixed now.

bearbin:
Have you properly addressed xoft's concern about file name storage?

worktycho:
We agreed that what I was doing was fine.

PR #2023, opened 2015-05-15
PathFinder - smart pointers

SafwatHalaby:
This is not complete yet.
I am new to smart pointers so I won't be surprised if there are bad practices.
If you're wondering why there are still raw pointers around (and why I use get()), that's because I followed the guidelines in this [stack exchange answer] (http://stackoverflow.com/questions/8338570/c11-replace-all-non-owning-raw-pointers-with-stdshared-ptr). "Delete" should never be called on those raw pointers.
I would have preferred using a reference instead, but some functions need to return a nullptr, so I didn't.

Like I said, new to this. All advices welcome.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -215,11 +215,6 @@ bool cPath::Step_Internal()
 
 void cPath::FinishCalculation()
 {
-	for (auto && pair : m_Map)
-	{
-		delete pair.second;
-	}
-
 	m_Map.clear();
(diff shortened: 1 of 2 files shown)


bearbin:
> "This is not complete yet."

worktycho:
Reverted the merge.

SafwatHalaby:
It's actually complete if everyone is fine with it.
I wanted advice regarding the use of raw pointers and the use of `get()`.

tigerw:
When this is okay and if nothing else is added, revert the revert reverting the merge?

worktycho:
@tigerw you do that then.

SafwatHalaby:
That was messy :package:

PR #2028, opened 2015-05-15
GetBlockTypeMeta overloads

SafwatHalaby:


SafwatHalaby:
Looks like my IDE insisted on removing some TABS...

src/Chunk.cpp
```diff
@@ -2637,6 +2637,32 @@ void cChunk::GetBlockTypeMeta(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE & a_
...
+	int RelZ = a_Location.z - GetPosZ() * cChunkDef::Width;
+
+	GetBlockTypeMeta(RelX, a_Location.y, RelZ, a_BlockType, a_BlockMeta);
+}
+
+
+
+
+
+void cChunk::GetBlockTypeMeta(Vector3d a_Location, BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta) const
```
worktycho: Do we really need this overload?

SafwatHalaby: They're not essential, but seeing the Relx, RelZ duplication all over the project so I made them.

worktycho: I agree with the Vector3i overload, I'm doing the same thing. Its a second vector3d overload which I think is unnecessary.

SafwatHalaby: It just saves calling `floor()` when dealing the monster, whose vectors are mostly Vector3d. It is *that* necessary but why not?

worktycho: Other than the fact cChunk has a 500 line header? I just think we shouldn't add code to our god objects where possible.

madmaxoft:
guru tip: use https://github.com/mc-server/MCServer/pull/2028/files?w=0 to view diff without whitespace.

I'm with @worktycho  here, I don't think the floating-point functions should be added.
Additionally, the parameters should by by-const-ref, instead of by-value, to avoid needless copying.

SafwatHalaby:
Alright

SafwatHalaby:
I'll removed the Vector3d verison.
But there's more that can be done here.

I observed that almost always, we have this preparation code:
```
cChunk * Chunk = a_Chunk.GetNeighborChunk(blockStuff);
	BLOCKTYPE BlockType;
	NIBBLETYPE BlockMeta;

	if ((Chunk == nullptr) || !Chunk->IsValid())
	{
		return false;
	}
     .Only now we call GetBlockTypeMeta
```

In some cases, that preparation code isn't needed, e.g. you are checking a block beneath you and it has to be on the same chunk.

That's ugly code duplication.
One solution is putting that preparation code inside GetBlockTypeMeta. But the few cases that don't need the preparation code will needlessly call GetNeighborChunk. If calling GetNeighborChunk in such a way that a chunk returns itself if cheap, then this is fine, and it eliminates all those ugly chunk crashes we sometimes get.

madmaxoft:
Actually the preparation code is only needed if you're not sure of which chunk you should query, and that happens only in the `cEntity`-related code. All the other parts of MCS know exactly which chunk they're working with and don't need the preparation code. I believe the `GetBlockTypeMeta()` is used in some tight loops so I wouldn't add any unnecessary code to it.

SafwatHalaby:
I see. How about a variation instead of a modification: `GetBlockTypeMetaSafe()` which calls the safe code first, and then calls `GetBlockTypeMeta()`

madmaxoft:
How about restructuring the cEntity code so that it doesn't need the prep code all over the place? Just move the cChunk switching to the place that actually moves the entity, so that the correct chunk is assigned right after the coords change. This could also check for and prevent bad things such as mobs moving out into an unloaded chunk etc.

SafwatHalaby:
Good idea. But I think that would require changing the tick() functions to having a chunk pointer rather than a chunk reference, because this pattern is all over the place:
```
// a child entitity
child::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
{
super::Tick(a_Dt, a_Chunk); //super might move the entity, we want our a_Chunk to be affected
...
}
```

madmaxoft:
Yes, that's true. I'm okay with that (as long as you use a typedef for `cChunk *` to avoid the `*&`).

I have a feeling we're having yet another issue here - if an entity moves across chunk borders, it may be ticked twice - once in the old chunk (which moves it to new chunk) and then again in the new chunk.

SafwatHalaby:
Closing this. Will make another one with more drastic changes to `tick()` later.

PR #2029, opened 2015-05-15
Bed uses bottom two bits for direction not three

worktycho:


madmaxoft:
There are more places in the cpp where the code checks three bits instead of two, you should fix all of them

worktycho:
Fixed. Also I'm not sure what the &0xd is is doing in the set at the end.

SafwatHalaby:
Offtopic but a bit relevant: 

>No magic numbers, use named constants

There are lots of magic numbers when it comes to meta values, and it's kind of ugly.

jan64:
i believe these are not magic numbers, just bitmasks.

worktycho:
I agree with that, but we've never worked out a good way of dealing with all the magic numbers. If only C++ bit-fields weren't terrible. It might be possible to do something with templates to improve the situation though.

SafwatHalaby:
> i believe these are not magic numbers but bitmasks. 

Magic bitmasks then :P

jan64:
if you convert them to binary, they aren't magic - they are actually readable. :P

Issue #1984, opened 2015-05-09
Players cannot see when another player opens any wooden door, except standard oak


jammet:
Players cannot see when another player opens any wooden door, except standard oak. The old standard door opens and closes for everyone to see. The acacia one and any other, doesn't make a sound or looks like it opens and closes.

jan64:
i believe this can be closed

jammet:
Tested. Works fine. :)

Issue #1868, opened 2015-04-17
No sound for lightning


mathias-github:
Currently, there is no sound for lightning.

pillsbury7901:
Hello, I would like to contribute. I am new to GitHub and would like to know how to get sarted reproducing this bug. I would very much appreiate any feedback. Thanks in advance.

bearbin:
@pillsbury7901 If you haven't already, you should read this document: https://github.com/mc-server/MCServer/blob/master/CONTRIBUTING.md

The easiest way to reproduce this bug would be to start a server and change the weather to stormy, so that a lightning strike will occur every so often.

bearbin:
Sorry, wrong document. You should read that one too, but this one is more important for a newcomer to the project: 

https://github.com/mc-server/MCServer/blob/master/GETTING-STARTED.md

pillsbury7901:
Thank you for the response. I have been trying to compile MCServer for the past few days to no avail. Do you have any suggestions for compiling on a linux box? I believe the error is related to gcc but I have not been able to remedy this issue. Thanks again.

worktycho:
Its probably a gcc version that does not support c++11. Check gcc --version, and if its less than gcc 4.8 then you need to install a more recent gcc.

pillsbury7901:
This is what i feared. When I attempt to update with apt-get or aptitude it says that I have the latest version installed. I have downloaded gcc-5.1.0 and followed the installation instructions from the Documentation precisely but I am still getting the same error.

worktycho:
If you've got multiple versions of gcc installed it may be that you're calling the wrong one. Try `which gcc` and if thats wrong, specify the path to the correct one when calling cmake, like `CC=<path/to/gcc> CXX=<path/to/g++> cmake`

pillsbury7901:
Thank you. I will try this now.

pillsbury7901:
Question, must I buy a Minecraft Client to debug and fix code for this project?

madmaxoft:
There are tasks that don't require any interaction with the client; as far as I know @worktycho isn't using a client at all. But of course majority of tasks will require some form of verification through the client interface. So it really depends on what you want to be doing.
This particular task requires a client to verification that a fix is indeed working.

PR #2030, opened 2015-05-16
Issue #1868 and Issue #1984

beeduck:
Addressing issue #1868, I've added the sound effect for thunder. It is created with a fixed volume to be realistically heard over the sound of the rain.

I noticed that lightning is currently created at a fixed location (0, 0, 0), I was wondering if anyone is currently working on the TODO of changing lightning to be cast randomly near player locations. If not, I'd like to try and fix that as well. This would also probably change the fixed volume of the thunder sound effect.

I've also addressed issue #1984. Door open/close and sound effect handling was only looking for basic wooden door types.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -2,6 +2,7 @@ Many people have contributed to MCServer, and this list attempts to broadcast at
 
 BasedDoge (Donated AlchemistVillage prefabs)
 bearbin (Alexander Harkness)
+beeduck
 derouinw
 Diusrex
 Duralex
--- a/src/Blocks/BlockDoor.cpp
+++ b/src/Blocks/BlockDoor.cpp
@@ -50,10 +50,24 @@ void cBlockDoorHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterfac
 	UNUSED(a_CursorY);
 	UNUSED(a_CursorZ);
 
-	if (a_ChunkInterface.GetBlock(a_BlockX, a_BlockY, a_BlockZ) == E_BLOCK_WOODEN_DOOR)
+	switch (a_ChunkInterface.GetBlock(a_BlockX, a_BlockY, a_BlockZ))
 	{
-		ChangeDoor(a_ChunkInterface, a_BlockX, a_BlockY, a_BlockZ);
-		a_Player->GetWorld()->BroadcastSoundParticleEffect(1003, a_BlockX, a_BlockY, a_BlockZ, 0, a_Player->GetClientHandle());
+		default:
+		{
+			ASSERT(!"Unhandled door block type");
+		}
+		case E_BLOCK_ACACIA_DOOR:
+		case E_BLOCK_BIRCH_DOOR:
+		case E_BLOCK_DARK_OAK_DOOR:
+		case E_BLOCK_JUNGLE_DOOR:
+		case E_BLOCK_SPRUCE_DOOR:
+		case E_BLOCK_IRON_DOOR:
+		case E_BLOCK_WOODEN_DOOR:
+		{
+			ChangeDoor(a_ChunkInterface, a_BlockX, a_BlockY, a_BlockZ);
+			a_Player->GetWorld()->BroadcastSoundParticleEffect(1003, a_BlockX, a_BlockY, a_BlockZ, 0, a_Player->GetClientHandle());
+			break;
+		}
 	}
 }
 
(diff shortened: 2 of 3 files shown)


bearbin:
Looks great!

If you could add yourself to the CONTRIBUTORS file, then I think this could be merged.

https://github.com/mc-server/MCServer/blob/master/CONTRIBUTORS

src/Blocks/BlockDoor.cpp
```diff
@@ -50,10 +50,20 @@ void cBlockDoorHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterfac
 	UNUSED(a_CursorY);
 	UNUSED(a_CursorZ);
 
-	if (a_ChunkInterface.GetBlock(a_BlockX, a_BlockY, a_BlockZ) == E_BLOCK_WOODEN_DOOR)
+	switch (a_ChunkInterface.GetBlock(a_BlockX, a_BlockY, a_BlockZ))
```
worktycho: This switch should have a default case, to cover future new door blocks.

src/World.cpp
```diff
@@ -2452,6 +2452,7 @@ void cWorld::BroadcastTeleportEntity(const cEntity & a_Entity, const cClientHand
 void cWorld::BroadcastThunderbolt(int a_BlockX, int a_BlockY, int a_BlockZ, const cClientHandle * a_Exclude)
 {
 	m_ChunkMap->BroadcastThunderbolt(a_BlockX, a_BlockY, a_BlockZ, a_Exclude);
+	BroadcastSoundEffect("ambient.weather.thunder", a_BlockX, a_BlockY, a_BlockZ, 50, 1);
```
worktycho: This should probably be in CastThunderbolt, not broadcast thunderbolt because the broadcasts are tied to the packets.

NiLSPACE: I disagree. Some plugins might want it to just make a bright flash in the sky (for example for some kind of minigame). Not to make the sound with it.

NiLSPACE: I think there should be a way to make the lightning flash without the sound. It could be used in for example minigames.

worktycho: Which is why it should be in CastThunderbolt. CastThunderbolt is for is a plugin wants to simulate a minecraft thunderbolt with everything that comes with that. BroadcastThunderbolt should just be to send the flash.

NiLSPACE: Yeah, I didn't make it clear that I agreed with you. We do have to export the BroadcastThunderbolt function as well though.

worktycho: I'm happy exporting it. But not until after I've moved it to cBroadcaster.

src/Blocks/BlockDoor.cpp
madmaxoft: With an `ASSERT(!"Unhandled door block type");` in it so that it is quickly found when such a case happens.

beeduck:
Okay I'll make the default door case and thunder sound effect change a little later today

beeduck:
Are there any other changes you need me to make to this before it can be merged?

CONTRIBUTORS
```diff
@@ -35,5 +35,6 @@ WebFreak001
 worktycho
 xoft
 Yeeeeezus (Donated AlchemistVillage prefabs)
+beeduck
```
bearbin: You should be near the front, this list is alpha-sorted.

bearbin:
I don't think you've addressed @worktycho and @xoft's comments yet?

worktycho:
@bearbin, there addressed, The only issue is the style of the default. I like it, but I suspect @madmaxoft is going to want braces. So I'm waiting on his comments.

src/Blocks/BlockDoor.cpp
```diff
@@ -50,10 +50,22 @@ void cBlockDoorHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterfac
...
 	{
-		ChangeDoor(a_ChunkInterface, a_BlockX, a_BlockY, a_BlockZ);
-		a_Player->GetWorld()->BroadcastSoundParticleEffect(1003, a_BlockX, a_BlockY, a_BlockZ, 0, a_Player->GetClientHandle());
+		default:
+			ASSERT(!"Unhandled door block type");
```
madmaxoft: Braces: yes please :)

madmaxoft:
Fix the braces and squash the unneeded commits, and we can merge.

beeduck:
Alright, all the commits should be in order now

PR #2032, opened 2015-05-16
Broadcast refactor

worktycho:
Moved BroadcastChunkData over to cBroadcaster.

Also some how have some cherry picks from testing in the history.


madmaxoft:
That would be because you based your branch on `testing` instead of `master`. Could you rebase (or simply cherrypick the commit and overwrite the branch)?

PR #2034, opened 2015-05-16
Moved BroadcastChunkData over to cBroadcaster

worktycho:


madmaxoft:
Do we really need a new type for the two coords? Especially when it is only used to store them, there are no operations defined.

madmaxoft:
You need to add the BroadcasterForwards.cpp to the list of ignored files in CheckBasicStyle.lua

worktycho:
Theres a lot methods that work with such pairs, so I think a two element vector is worthwhile.

worktycho:
Fixed check basic style

madmaxoft:
Still not good. Can't the source file be generated in current source dir, instead of the build dir?

worktycho:
I've found a better way of handling BroadcastChunkData, so I'm going to leave this for now.

PR #2036, opened 2015-05-16
Should Fix sign flipping.

### LINKED ISSUE #1118: Sign block meta-data mirroring is wrong
Mirroring the signs using the `//mirror xy` or `//mirror yz` WorldEdit commands is buggy, because the sign blockhandler's meta mirroring is wrong. Fix the `cBlockSignHandler::MetaMirrorXY()` and `cBlockSignHandler::MetaMirrorYZ()` functions.

bearbin:
Fixes #1118

### DIFF
--- a/src/Blocks/BlockSignPost.h
+++ b/src/Blocks/BlockSignPost.h
@@ -71,7 +71,7 @@ public:
 
 		// There are 16 meta values which correspond to different directions.
 		// These values are equated to angles on a circle; 0x08 = 180 degrees.
-		return (a_Meta < 0x08) ? (0x08 + a_Meta) : (0x08 - a_Meta);
+		return (a_Meta > 0x08) ? (0x08 - a_Meta) : (0x18 - a_Meta);
 	}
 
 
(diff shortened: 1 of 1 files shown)


madmaxoft:
Have you tested this?

bearbin:
No. I will now.

bearbin:
I've now tested it, but mirroring doesn't seem to be working on anything now :(

bearbin:
I tested this, and the flipping didn't work using WorldEdit - but when I tested vines they didn't work either, so I think it may just be the worldedit flip function that's broken.

Otherwise, this should work, as it matches up with the wiki information.

madmaxoft:
I have a feeling vines don't work either. Try a wall-mounted torch, that should work 100%

bearbin:
Nope, torch didn't work.

tigerw:
nothing works

bearbin:
I'm pretty sure that this code is correct, and since the mirroring using worldedit is not working at all and there's no way to test, should I just merge this?

NiLSPACE:
It's not working in WorldEdit? What? Why isn't that reported? How did you try. I believe you have to copy it first, mirror it, and then paste it again.

bearbin:
I selected an area and did `//mirror`. The blocks in the area switched around, but they did not rotate.

SphinxC0re:
That's another problem. May someone remove this coveralls job?

PR #2037, opened 2015-05-16
Pathfinder - approximated paths when original destination unreachable

### LINKED ISSUE #1963: Escaping mobs sometimes do not escape
When hitting a passive mob, it is supposed to escape, sometimes it does not.

I know what the underlying cause is. I'm reporting it to remind myself to fix it and to let others know that the problem is known.

The problem is as follows: When hit, a mob will set a far away point as its destination, if that destination happens to be unreachable (usually happens to be a tree top), the mob will not run away.

SafwatHalaby:
- Mobs now move to a nearby place if no path to target
- Fixed bugs when leaning
- Temporarily fixed Attacking through walls by lowering attackDistance to 1.5 (Someone should check out why the trace at cAggressiveMonster::Tick is failing)
- Fixed mob head shakes

Fixes #1963

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -89,7 +89,7 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_SoundDeath(a_SoundDeath)
 	, m_AttackRate(3)
 	, m_AttackDamage(1)
-	, m_AttackRange(2)
+	, m_AttackRange(1)
 	, m_AttackInterval(0)
 	, m_SightDistance(25)
 	, m_DropChanceWeapon(0.085f)
@@ -147,10 +147,15 @@ bool cMonster::TickPathFinding(cChunk & a_Chunk)
 		(Recalculate lots when close, calculate rarely when far) */
 		if (
 			((GetPosition() - m_PathFinderDestination).Length() < 0.25) ||
-			((m_TicksSinceLastPathReset > 10) && (m_TicksSinceLastPathReset > (0.15 * (m_FinalDestination - GetPosition()).SqrLength())))
+			((m_TicksSinceLastPathReset > 10) && (m_TicksSinceLastPathReset > (0.4 * (m_FinalDestination - GetPosition()).SqrLength())))
 		)
 		{
-			ResetPathFinding();
+			/* Re-calculating is expensive when there's no path to target, and it results in mobs freezing very often as a result of always recalculating.
+			This is a workaround till we get better path recalculation. */
+			if (!m_NoPathToTarget)
+			{
+				ResetPathFinding();
+			}
 		}
 	}
 
@@ -161,12 +166,21 @@ bool cMonster::TickPathFinding(cChunk & a_Chunk)
 			StopMovingToPosition();  // Invalid chunks, probably world is loading or something, cancel movement.
 			return false;
 		}
+		m_NoPathToTarget = false;
+		m_NoMoreWayPoints = false;
 		m_PathFinderDestination = m_FinalDestination;
 		m_Path = new cPath(a_Chunk, GetPosition().Floor(), m_PathFinderDestination.Floor(), 20);
 	}
 
 	switch (m_Path->Step(a_Chunk))
 	{
+		case ePathFinderStatus::NEARBY_FOUND:
+		{
+			m_NoPathToTarget = true;
+			m_Path->AcceptNearbyPath();
+			break;
+		}
+
 		case ePathFinderStatus::PATH_NOT_FOUND:
 		{
 			StopMovingToPosition();  // Give up pathfinding to that destination.
@@ -179,15 +193,22 @@ bool cMonster::TickPathFinding(cChunk & a_Chunk)
 		}
 		case ePathFinderStatus::PATH_FOUND:
 		{
-			if (--m_GiveUpCounter == 0)
+			if (m_NoMoreWayPoints || (--m_GiveUpCounter == 0))
 			{
 				ResetPathFinding();  // Try to calculate a path again.
 				return false;
 			}
-			else if (!m_Path->IsLastPoint() && (m_Path->IsFirstPoint() || ReachedNextWaypoint()))  // Have we arrived at the next cell, as denoted by m_NextWayPointPosition?
+			else if (!m_Path->IsLastPoint())  // Have we arrived at the next cell, as denoted by m_NextWayPointPosition?
 			{
-				m_NextWayPointPosition = Vector3d(0.5, 0, 0.5) + m_Path->GetNextPoint();
-				m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_NextWayPointPosition.
+				if ((m_Path->IsFirstPoint() || ReachedNextWaypoint()))
+				{
+					m_NextWayPointPosition = Vector3d(0.5, 0, 0.5) + m_Path->GetNextPoint();
+					m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_NextWayPointPosition.
+				}
+			}
+			else
+			{
+				m_NoMoreWayPoints = true;
 			}
 			return true;
 		}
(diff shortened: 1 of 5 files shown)


src/Mobs/Path.cpp
```diff
@@ -150,20 +143,26 @@ bool cPath::Step_Internal()
 	cPathCell * CurrentCell = OpenListPop();
 
 	// Path not reachable, open list exauhsted.
-	if (CurrentCell == nullptr)
+	if ((CurrentCell == nullptr) || (m_StepsLeft == 0))
```
worktycho: `<=` as a precaution?

SafwatHalaby:
Do not merge yet, I may have commited an old code. Checking.

src/Mobs/Path.cpp
```diff
@@ -150,20 +143,26 @@ bool cPath::Step_Internal()
...
 	{
-		FinishCalculation(ePathFinderStatus::PATH_NOT_FOUND);
-		ASSERT(m_Status == ePathFinderStatus::PATH_NOT_FOUND);
-		return true;
+		printf("Steps left %d\n", m_StepsLeft);
```
worktycho: Use LOGD, then it will only output in debug, and will go into the logfile.

src/Mobs/Path.cpp
SafwatHalaby: I love crashing stuff. Will add an ASSERT as a precaution.

src/Mobs/Path.cpp
```diff
@@ -150,20 +143,26 @@ bool cPath::Step_Internal()
...
+		// Try to find an alternative nearby path, which doesn't exactly match the destination.
+		if (!AttemptToFindAlternative())
+		{
+			printf("Failed to find alternative\n");
```
worktycho: Again LOGD

SafwatHalaby:
Yes, this is an old code, I have no idea what I messed up /:

src/Mobs/Path.cpp
worktycho: That's an alternative.

SafwatHalaby:
Recovered the right code and rebased.

SafwatHalaby:
The `AttemptToFindAlternative` is undergoing an overhaul, you may disregard the current version.

src/Mobs/Path.cpp
```diff
@@ -149,29 +155,17 @@ bool cPath::Step_Internal()
 {
 	cPathCell * CurrentCell = OpenListPop();
 
-	// Path not reachable, open list exauhsted.
-	if (CurrentCell == nullptr)
+	// Path not reachable.
+	if ((CurrentCell == nullptr))
```
worktycho: Why the double brackets?

src/Mobs/Path.cpp
```diff
@@ -213,6 +207,76 @@ bool cPath::Step_Internal()
...
+	z = OriginalZ;
+	// Scan a 5x3x5 cube around the original path, starting with a random spot.
+	// TODO: scan a 5x1x5 and If a block in the scan area is air, we scan the solid below instead.
+	for (;;)
+	{
+		for (;;)
+		{
+			for (y = -3; y <= 3; y+= 1)
+			{
+				if ((x == 0) && (y == 0) && (z == 0))
```
worktycho: Is it problematic to check the original destination? It seems that this only increases the complexity of the code unnecessarily.

src/Mobs/Path.cpp
```diff
@@ -213,6 +207,76 @@ bool cPath::Step_Internal()
...
+	int x, y, z;
+	x = OriginalX;
+	z = OriginalZ;
+	// Scan a 5x3x5 cube around the original path, starting with a random spot.
+	// TODO: scan a 5x1x5 and If a block in the scan area is air, we scan the solid below instead.
+	for (;;)
+	{
+		for (;;)
+		{
+			for (y = -3; y <= 3; y+= 1)
```
worktycho: Why not declare y here?

src/Mobs/Path.cpp
```diff
@@ -213,6 +207,76 @@ bool cPath::Step_Internal()
...
+	int OriginalX = (rand() % 5) - 2;
+	int OriginalZ = (rand() % 5) - 2;  // A random value between -2 and 2.
+	int x, y, z;
+	x = OriginalX;
+	z = OriginalZ;
+	// Scan a 5x3x5 cube around the original path, starting with a random spot.
+	// TODO: scan a 5x1x5 and If a block in the scan area is air, we scan the solid below instead.
+	for (;;)
+	{
+		for (;;)
```
worktycho: I think making this `for(int z = OriginalZ; z != OriginalZ; z = (z == 3 ? -2 : z + 1))` would be clearer

worktycho: And the same for the outer loop.

SafwatHalaby:
Travis failed due to parts of the code I didn't touch...

src/Mobs/Path.cpp
SafwatHalaby: I'm going to completely re-write this function. I'll keep track of the best cell so far and return that instead.

src/Mobs/Path.cpp
SafwatHalaby: I wrote: 

>The AttemptToFindAlternative function is undergoing an overhaul, you may disregard the current version.

SafwatHalaby:
It's as if @worktycho is focusing `AttemptToFindAlternative` on purpose :P

SafwatHalaby:
`printf` will be eventually removed.

SafwatHalaby:
This is not a minor change, I tested it thoroughly but I may have missed things. Please double check me, especially on CHUNK querying stuff.

src/Mobs/Monster.cpp
```diff
@@ -447,21 +510,32 @@ void cMonster::SetPitchAndYawFromDestination()
...
-		SetYaw(Rotation);
+		VectorToEuler(Distance.x, Distance.y, Distance.z, HeadRotation, HeadPitch);
+		if (abs(Rotation-HeadRotation) < 120)  // We're not an owl. If it's more than 90, don't look behind and instead look at the next waypoint.
+		{
+			SetHeadYaw(HeadRotation);
+			SetPitch(-HeadPitch);
+		}
+		else
+		{
+			Distance = m_NextWayPointPosition - GetPosition();
```
tigerw: These operations look like what the body rotation calculations above are doing, maybe just reuse BodyDistance if so?

SafwatHalaby: Yeah, makes perfect sense.

SafwatHalaby:
Is Coveralls.io stuck?

bearbin:
Yep. I'm going do disable it because it keeps getting stuck and is pretty much useless.

SafwatHalaby:
Squashing to see if Coveralls was disabled.

SafwatHalaby:
Yes, Coveralls is now disabled.

src/Mobs/Monster.cpp
```diff
@@ -269,22 +290,64 @@ bool cMonster::EnsureProperDestination(cChunk & a_Chunk)
...
+				m_FinalDestination.y -= 1;
+			}
 		}
-		m_FinalDestination.y -= 1;
 	}
 
 
+
+
+
```
madmaxoft: Don't add so many empty lines within a function body

src/Mobs/Monster.h
```diff
@@ -180,6 +180,13 @@ class cMonster :
 	/** Coordinates for the ultimate, final destination last given to the pathfinder. */
 	Vector3d m_PathFinderDestination;
 
+	/** True if there's no path to target and we're walking to an approximated location. */
+	bool m_NoPathToTarget;
+
+	/** Usually equals m_PathFinderDestination. If there isn't a path to m_PathFinderDestination,
+	Will be the coordinates of a nearby path. */
+	bool m_NoMoreWayPoints;
```
madmaxoft: Coordinates in a `bool`?

SafwatHalaby: Outdated comment for some other idea I had and discarded.

src/Mobs/Monster.cpp
SafwatHalaby: Fixed.

src/Mobs/Monster.h
SafwatHalaby: Fixed.

SafwatHalaby:
@bearbin, Travis is complaining about you:
```
'__APPLE__' is not defined, evaluates to 0 [-Werror,-Wundef]

#elif __APPLE__
```

SafwatHalaby:
I think this was already fixed before my build started, will re-shake Travis and see if it helps.

Issue #1995, opened 2015-05-10
Creative Mode drop bug


falkseidl:
If a player in gamemode 1 trys to move items from his inventory out, to drop them, then nothing happens.

Video: http://youtu.be/uOow4PNxnwc

NiLSPACE:
I always thought this was intentional. Doesn't vanilla do the same?

falkseidl:
Nope, i tried it.

PR #2038, opened 2015-05-16
Fixed inventory handling.

### LINKED ISSUE #2035: Crafting issues.
I encountered various issues with crafting. 

I first tried drag-to spread and spread out my wood planks around the mini crafting slot in the player window. On the client side this appeared to work, but upon trying to get the crafting result it failed.

Then I tried manually right-clicking with the planks in all of the boxes. Upon clicking on the result, it produced 4 sticks instead of the expected crafting table, while still using the original amount of wood.

Finally, I tried the method in the paragraph above again, and it did produce a crafting table and I placed the table in my inventory. When I pressed escape to exit the window, another crafting table was dropped on the ground in front of me, but upon a relog the dupe was fixed.

madmaxoft:
Fixes #2035.

### DIFF
--- a/src/Protocol/Protocol18x.cpp
+++ b/src/Protocol/Protocol18x.cpp
@@ -51,7 +51,7 @@ Implements the 1.8.x protocol classes:
 
 
 /** The slot number that the client uses to indicate "outside the window". */
-static const Int16 SLOT_NUM_OUTSIDE = -1;
+static const Int16 SLOT_NUM_OUTSIDE = -999;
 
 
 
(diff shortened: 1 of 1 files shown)


worktycho:
Will this mean we need to reopen #1995?

madmaxoft:
No, I fixed both at the same time. This is how it should have been fixed in the first place.

worktycho:
Merge if CI succeeds then

src/Protocol/Protocol18x.cpp
```diff
@@ -2265,7 +2265,7 @@ void cProtocol180::HandlePacketCreativeInventoryAction(cByteBuffer & a_ByteBuffe
 	{
 		return;
 	}
-	m_Client->HandleCreativeInventory(SlotNum, Item, (SlotNum == SLOT_NUM_OUTSIDE) ? caLeftClickOutside : caLeftClick);
+	m_Client->HandleCreativeInventory(SlotNum, Item, (SlotNum == -1) ? caLeftClickOutside : caLeftClick);
```
tigerw: But values are meant to be named?

PR #2039, opened 2015-05-17
Fix creepers not exploding

### LINKED ISSUE #708: Creepers don't explode when a player is within range but vertically one below
Being in a range of 1 horizontally but 1 vertically below the creeper will make him charge completly but he will never explode.

SafwatHalaby:
I wouldn't say this is fit to merging yet. I thought I'd PR it to share the issue.

Fixes #708


SafwatHalaby:
A proper PR coming soon.

PR #2040, opened 2015-05-17
Fix FastRandom seeding

### LINKED ISSUE #1815: Build is broken on OSX 10.10
As of commit a8a2d3d504a84f26d71aa364a169963a70ce6288, the build fails on OSX with the error:

MCServer/src/FastRandom.cpp:13:1: error: thread-local storage is unsupported for the current target
thread_local unsigned int m_Counter = 0;
^

Reversing the referenced commit allows it to build and run without issue.

### LINKED ISSUE #2021: Seed the MTRand/cFastRandom seed counter with a true random number
At the moment their is very little entropy in our world seeds because it is based off this counter.

worktycho:
Fixes #1815
Fixes #2021

### DIFF
--- a/src/FastRandom.cpp
+++ b/src/FastRandom.cpp
@@ -6,12 +6,28 @@
 #include "Globals.h"
 #include "FastRandom.h"
 
+#include <random>
+
 #ifdef _WIN32
-	#define thread_local __declspec(thread)
+	#define thread_local static __declspec(thread)
+#elseif __OSX__
+	#define thread_local static __thread
 #endif
 
-thread_local unsigned int m_Counter = 0;
-
+static unsigned int GetRandomSeed()
+{
+	thread_local bool SeedCounterInitialized = 0;
+	thread_local unsigned int SeedCounter = 0;
+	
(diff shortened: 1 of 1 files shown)


bearbin:
This compiles, but have you tested running this and are the seeds random?

worktycho:
I haven't tested it running, but It can't be any worse then the old seeding code.

src/FastRandom.cpp
```diff
@@ -6,12 +6,28 @@
...
+	#define thread_local static __declspec(thread)
+#elseif __OSX__
+	#define thread_local static __thread
 #endif
 
-thread_local unsigned int m_Counter = 0;
-
+static unsigned int GetRandomSeed()
+{
+	thread_local bool SeedCounterInitialized = 0;
```
tigerw: Aren't bools normally set to false/true?

worktycho: Yes, but I wanted a initialization value I was sure all three different thread_local systems would accept, and zero initialization is the safest there is.

Issue #2046, opened 2015-05-17
Trees drop too many saplings.


bearbin:
They should drop slightly fewer.

NiLSPACE:
Should be easy to fix. Simply increase the maximum number of https://github.com/mc-server/MCServer/blob/master/src/Blocks/BlockLeaves.h#L44

PR #2049, opened 2015-05-17
Fix sapling probability.

bearbin:
Probability is 5% of leaves drop a sapling.

This should really be different for jungle leaves (2.5%) and the Fortune enchantment should increase the probability.

### DIFF
--- a/src/Blocks/BlockLeaves.h
+++ b/src/Blocks/BlockLeaves.h
@@ -41,7 +41,7 @@ public:
 		cFastRandom rand;
 
 		// Old leaves - 3 bits contain display; new leaves - 1st bit, shifted left two for saplings to understand
-		if (rand.NextInt(6) == 0)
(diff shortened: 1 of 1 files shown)


NiLSPACE:
Does fortune really improve the probability? I thought it only worked on ores.

bearbin:
This should fix #2046. And see the wiki for more info: http://minecraft.gamepedia.com/Sapling#Obtaining

SafwatHalaby:
For [reference](url=http://minecraft.gamepedia.com/Sapling):
>When leaves are broken by any tool except shears, they will have a 5% chance of dropping a sapling. Jungle leaves instead have a 2.5% chance of dropping a sapling.

SafwatHalaby:
I accidentally merged, this is not that much of a problem, jungle can be fixed on a seperate PR.

PR #2060, opened 2015-05-18
Improved leaves a bit more.

### LINKED ISSUE #2058: Jungle saplings drop too often.
Jungle should have a 2.5% chance drop, and not 5%.

bearbin:
Should fix #2058.

### DIFF
--- a/src/Blocks/BlockLeaves.h
+++ b/src/Blocks/BlockLeaves.h
@@ -40,29 +40,41 @@ public:
 	{
 		cFastRandom rand;
 
-		// Old leaves - 3 bits contain display; new leaves - 1st bit, shifted left two for saplings to understand
-		if (rand.NextInt(20) == 0)
+		// There is a chance to drop a sapling that varies depending on the type of leaf broken.
+		// TODO: Take into account fortune for sapling drops.
+		int chance;
+		if ((m_BlockType == E_BLOCK_LEAVES) && ((a_BlockMeta & 0x03) == E_META_LEAVES_JUNGLE))
+		{
+			// Jungle leaves have a 2.5% chance of dropping a sapling.
+			chance = rand.NextInt(40);
+		}
+		else
+		{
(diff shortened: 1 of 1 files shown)


worktycho:
Merge on CI success

bearbin:
I think there's also something odd with shears, saplings sometimes drop when you break leaves with them, according to @NiLSPACE.

src/Blocks/BlockLeaves.h
```diff
@@ -40,22 +40,33 @@ class cBlockLeavesHandler :
...
-					E_BLOCK_SAPLING,
-					1,
-					(m_BlockType == E_BLOCK_LEAVES) ? (a_BlockMeta & 0x03) : (2 << (a_BlockMeta & 0x01))
-				)
-			);
+			cItem(
+				E_BLOCK_SAPLING,
+				1,
+				(m_BlockType == E_BLOCK_LEAVES) ? (a_BlockMeta & 0x03) : (4 + (a_BlockMeta & 0x01))
+			));
```
madmaxoft: The indent is wrong here. Either merge the `a_Pickups.push_back` and `cItem(` lines, or separate the two parentheses here (and indent the internals by one more level)

bearbin: Yes, that has got messed up. I'll fix it.

NiLSPACE:
Yes, didn't you have the shears problem as well?

bearbin:
I think the problem is that the shears handler is calling ConvertToPickups.

https://github.com/mc-server/MCServer/blob/814e4404bb428a0d825f12e513562d4e43e9e6dc/src/Items/ItemShears.h#L29

Or maybe is the code that;s currently in ConvertToPickups supposed to be in OnBlockDestroyed for the leaves?

bearbin:
Can anybody comment about the shears problem?

bearbin:
Merging, the shears issue can be fixed later.

PR #2064, opened 2015-05-18
Added hooks HOOK_ENTITY_CHANGE_WORLD and HOOK_ENTITY_CHANGED_WORLD

### LINKED ISSUE #1435: HOOK_PLAYER_SPAWNED doesn't get called if player has been moved to world

Seadragon91:
This fixes #1435.
I wanted to add the lua description for the hooks. How can I add them?

### DIFF
--- a/MCServer/Plugins/APIDump/Hooks/OnEntityChangeWorld.lua
+++ b/MCServer/Plugins/APIDump/Hooks/OnEntityChangeWorld.lua
@@ -1,0 +1,29 @@
+return
+{
+	HOOK_ENTITY_CHANGE_WORLD =
+	{
+		CalledWhen = "Before a entity is changing the world.",
+		DefaultFnName = "OnEntityChangeWorld",  -- also used as pagename
+		Desc = [[
+			This hook is called before the server moves the {{cEntity|entity}} to the given world. Plugins may
+			refuse the changing of the entity to the new world.<p>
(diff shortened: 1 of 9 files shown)


bearbin:
The description is in MCServer/Plugins/APIDump/apidesc.lua

Also, your build failed.

Seadragon91:
To the hook files, is that really correct?
`The following paths are ignored by one of your .gitignore files:
MCServer/Plugins/APIDump/Hooks/OnEntityChangeWorld.lua
Use -f if you really want to add them.`

bearbin:
Oh, sorry. The hook files are in the Hooks directory.

Your git change seems correct to me.

Seadragon91:
Okay, I have boths hook files should I add then to the same commit? Or both files to a own commit?

bearbin:
I don't think the commits need squashing, how they are at the moment is fine.

madmaxoft:
Someone was too trigger-happy again.

The hooks are named badly. We use an `-ING` and `-ED` pair for pre- and post- hooks.

PR #2066, opened 2015-05-18
Fixed compile error in Monster.cpp

jasperarmstrong:
On line 522 of Monster.cpp, abs() was being used instead of fabs() for variables of type double. I guess this compiled fine on other platforms, but on OS X it caused the build to fail.


bearbin:
I think std::abs would be better than fabs. C++ style, vs C-style.

bearbin:
Thanks for the contribution. After the build succeeds I'll merge.

bearbin:
The build failed because you indented with spaces, manually merging.

jasperarmstrong:
oops sorry about that

Issue #2051, opened 2015-05-17
Furnaces stay lit


NiLSPACE:
When using a furnace the furnace stays lit, even after it finished. When relogging it returns to unlit, so it is client-side.

tigerw:
A brief test seemed to show the block changing correctly. Could it be that the fuel wasn't all consumed yet, so it remained lit?

bearbin:
No, all the fuel was consumed when I tested and it still stayed lit.

NiLSPACE:
I noticed that when I remove the fuel the furnaces becomes unlit.

bearbin:
I experienced this bug again, but this time it stayed lit even after a relog.

PR #2069, opened 2015-05-19
Fixes lit furnace issue in #2051

HaoTNN:
m_FuelBurnTime never got set to 0 after fuel was used up, so it'd just increment and never got to the code block that set it to an unlit furnace.


bearbin:
Has this been tested and does it work?

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -189,6 +189,7 @@ void cFurnaceEntity::BurnNewFuel(void)
 	// Is the input and output ready for cooking?
 	if (!CanCookInputToOutput())
```
tigerw: Maybe merge this condition with the one above, so that both BurnTimes and IsCooking are set?

HaoTNN:
@tigerw That works, I'll add that in in the next commit.

I've tested it out and it seems to get rid of the issue referenced. However, there's some other issues with the furnace I found (unrelated to my changes) given the scenarios:

- Furnace is empty and not burning. If you drop in a valid input item, it'll light up. Also occurs when output slot contains same item as input slot.
E.g, Furnace has cooked pork in output slot with no fuel and isn't cooking. If you drop in raw pork as input, it'll light up.

- Furnace has fuel but not burning, has an item in output, and has different item in input. If you remove the output item, the input item will not start cooking.
E.g, Furnace has cooked pork in output slot with fuel but isn't cooking, and has iron ore in the input slot. If you remove the cooked pork, the iron ore won't start cooking.

I'm thinking I should find a fix for this and squash it down to a single commit since it's still furnace related.

madmaxoft:
Yes please.
Just in case, make sure you test furnaces saving and loading from world storage, there have been several issues there as well and fixing them is not always straightforward.

HaoTNN:
Apologies for late update, still getting to know the code base.

I moved some code around and got rid of problems mentioned earlier about the furnace, plus some. Set up some test cases and everything seems to be working fine, including saving/loading from world storage.

About the problems from loading and saving: When the furnace is being loaded from world storage, OnSlotChanged is firing when the items are loaded in, causing a few issues. Ended up fixing it by introducing a "LoadingState" boolean so that when the furnace is being loaded, OnSlotChanged won't call any of the update on the inventory. Would this be the appropriate way to fix it?

src/BlockEntities/FurnaceEntity.h
```diff
@@ -129,6 +134,9 @@ class cFurnaceEntity :
 
 	/** Amount of ticks that the current fuel has been burning */
 	int m_TimeBurned;
+
+	/** Is the block currently being loaded into the world? */
+	bool m_LoadingState;
```
madmaxoft: `m_IsLoading` would be a better name.

HaoTNN:
Okay, just made the change.

worktycho:
The only thing I'm worried about is the other world storage schemes

madmaxoft:
Next time please rebase your commits, instead of merging master into them. It will simplify history and also allow you to later squash unneeded multiple commits into a single one.

Ad other world storage schemas: The `Forgetful` one doesn't load any data at all and the `Classic` one was removed some time ago, so there are no more schemas to worry about :)

HaoTNN:
I see; where do you recommended I go from here? Can I still rebase at this point? Still learning how to use GitHub properly.

madmaxoft:
I think this PR is beyond rebasing, so we'll merge it as-is (if there's no-one opposed, I'm satisfied with the current state). Just use the rebasing next time :)

bearbin:
Merge it in then :)

madmaxoft:
Sir yes sir! :)

SuperSpyTX:
Thought I comment on the fact I've never seen a commit hash where the first 7 characters in the hash are just pure numbers.

tigerw:
One of a kind

PR #2073, opened 2015-05-19
Path recalculation improvements

SafwatHalaby:
Mobs are now much less dull when there is no path to the player.

I also added an element of randomness, if you're unreachable, the mobs will not just stack up on the closest block to you. This partially fixes the mob stacking issue, when that is fixed we can remove the randomness, but I think it's actually awesome and can be kept anyways.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -6,6 +6,7 @@
 #include "Path.h"
 #include "../Chunk.h"
 
+
 #define DISTANCE_MANHATTAN 0  // 1: More speed, a bit less accuracy 0: Max accuracy, less speed.
 #define HEURISTICS_ONLY 0  // 1: Much more speed, much less accurate.
 #define CALCULATIONS_PER_STEP 10  // Higher means more CPU load but faster path calculations.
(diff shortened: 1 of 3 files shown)


src/Mobs/Path.cpp
```diff
@@ -185,11 +186,18 @@ bool cPath::Step_Internal()
...
+		{
+			if (m_Rand.NextInt(4) == 0)
+			{
+				m_NearestPointToTarget = CurrentCell;
+			}
+		}
+		else if (CurrentCell->m_H < m_NearestPointToTarget->m_H)
+		{
+			m_NearestPointToTarget = CurrentCell;
+		}
```
bearbin: Indentation wat?

SafwatHalaby: Fixed.

src/Mobs/Path.cpp
```diff
@@ -185,11 +186,18 @@ bool cPath::Step_Internal()
 
 	// Calculation not finished yet.
 	// Check if we have a new NearestPoint.
-	if (CurrentCell->m_H < m_NearestPointToTarget->m_H)
+
+	if ((m_Destination-CurrentCell->m_Location).Length() < 5)
```
madmaxoft: Spaces around `-`

SafwatHalaby: Fixed.

SafwatHalaby:
Squashed.

SphinxC0re:
Looks good to me. May I merge?

PR #2074, opened 2015-05-19
Removed UniquePTR from PathFinder

SafwatHalaby:
@Woazboat noticed it's not needed, so I removed it.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -11,15 +11,8 @@
 #define CALCULATIONS_PER_STEP 10  // Higher means more CPU load but faster path calculations.
 // The only version which guarantees the shortest path is 0, 0.
 
-enum class eCellStatus {OPENLIST,  CLOSEDLIST,  NOLIST};
-struct cPathCell
-{
-	Vector3i m_Location;   // Location of the cell in the world.
-	int m_F, m_G, m_H;  // F, G, H as defined in regular A*.
-	eCellStatus m_Status;  // Which list is the cell in? Either non, open, or closed.
-	cPathCell * m_Parent;  // Cell's parent, as defined in regular A*.
-	bool m_IsSolid;	   // Is the cell an air or a solid? Partial solids are currently considered solids.
-};
+
+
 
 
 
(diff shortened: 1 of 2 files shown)


src/Mobs/Path.h
```diff
@@ -24,7 +24,15 @@ class cChunk;
...
-struct cPathCell;  // Defined inside Path.cpp
+enum class eCellStatus {OPENLIST,  CLOSEDLIST,  NOLIST};
+struct cPathCell
+{
+	Vector3i m_Location;   // Location of the cell in the world.
+	int m_F, m_G, m_H;  // F, G, H as defined in regular A*.
+	eCellStatus m_Status;  // Which list is the cell in? Either non, open, or closed.
+	cPathCell * m_Parent;  // Cell's parent, as defined in regular A*.
+	bool m_IsSolid;	   // Is the cell an air or a solid? Partial solids are currently considered solids.
+};
```
NiLSPACE: I think we want 5 empty spaces after this

SphinxC0re: You mean lines, don't you?

NiLSPACE: Yes, sorry ;)

SafwatHalaby: Fixed.

src/Mobs/Path.cpp
```diff
@@ -388,23 +381,20 @@ void cPath::ProcessCell(cPathCell * a_Cell, cPathCell * a_Caller, int a_GDelta)
...
 				si::setBlock(a_Location.x, a_Location.y, a_Location.z, debug_unchecked, Cell->m_IsSolid ? NORMAL : MINI);
 			#endif
 		#endif
-		return Cell;
+		return &m_Map[a_Location];
 	}
 	else
 	{
-		return m_Map[a_Location].get();
+		return &m_Map[a_Location];
```
SphinxC0re: What does the else do here? you can just return  &m_Map[a_Location]; after closing the if

SafwatHalaby: True, it's logically pointless and it's just there for code clarity.
I wonder if compilers optimize it away...

SphinxC0re: I don't think so. Compilers aren't that intelligent

worktycho: They will, its just basic hoisting.

SphinxC0re: Okay then I think it's ready for merge

SafwatHalaby:
You should have let me squash it first.

Issue #2072, opened 2015-05-19
"Can't connect to sessionserver.mojang.com: SSL handshake failed" causes segfault


wizardfrag:
The following are logs from the latest nightly build on Ubuntu 14.04 x86_64
```
     [15:35:20] --- Started Log ---

     [15:35:20] MCServer Jenkins gcc x64 Release (master) build id: #559
     [15:35:20] from commit id: 007bac638bb732aad7ca63dd3e7a79f5dbb9d2b9 built at: Mon May 18 11:01:06 EDT 2015
     [15:35:20] Creating new server instance...
     [15:35:20] Reading server config...
     [15:35:20] Starting server...
info [15:35:20] Compatible clients: 1.7.x, 1.8.x
info [15:35:20] Compatible protocol versions 4, 5, 47
Warn [15:35:20] WebServer: The server is running in unsecured HTTP mode.
info [15:35:20] Put a valid HTTPS certificate in file 'webadmin/httpscert.crt' and its corresponding private key to 'webadmin/httpskey.pem' (without any password) to enable HTTPS support
     [15:35:20] Loaded 407 crafting recipes
     [15:35:20] Loaded 20 furnace recipes and 56 fuels
     [15:35:20] -- Loading Plugins --
     [15:35:20] Initialised Core v.15
     [15:35:20] Initialized TransAPI v.1
     [15:35:20] Initialized ChatLog v.3
     [15:35:20] -- Loaded 3 Plugins --
     [15:35:21] Startup complete, took 721ms!
Warn [15:35:47] SecureRequest: Can't connect to sessionserver.mojang.com: SSL handshake failed: -0x2700
info [15:35:47] Kicking player wizardfrag for "Failed to authenticate account!"
Err  [15:36:48]   D:    | MCServer has encountered an error and needs to close
Err  [15:36:48] Details | SIGSEGV: Segmentation fault
Err  [15:36:48] MCServer Jenkins gcc x64 Release (master) build id: #559
Err  [15:36:48] from commit id: 007bac638bb732aad7ca63dd3e7a79f5dbb9d2b9 built at: Mon May 18 11:01:06 EDT 2015
Err  [15:36:48]   D:    | MCServer has encountered an error and needs to close
Err  [15:36:48] Details | SIGABRT: Server self-terminated due to an internal fault
Err  [15:36:48] MCServer Jenkins gcc x64 Release (master) build id: #559
[...]

PR #2076, opened 2015-05-19
Don't verify certificate

worktycho:
Temporary Hot-fix for #2072

### DIFF
--- a/src/PolarSSL++/SslContext.cpp
+++ b/src/PolarSSL++/SslContext.cpp
@@ -152,7 +152,7 @@ void cSslContext::SetCACerts(const cX509CertPtr & a_CACert, const AString & a_Ex
 	m_CACerts = a_CACert;
 	
 	// Set the trusted CA root cert store:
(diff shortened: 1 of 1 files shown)


bearbin:
Looks good. We need a proper fix but this should do for now.

madmaxoft:
This is stupid, why not simply change the root cert?

madmaxoft:
Also, it affects all SSL connections (even those by plugins), not only the Mojang one.

SphinxC0re:
I agree. This is not even a *real* fix. We're just abandoning the real problem

bearbin:
The problem is that the new certificate is a wildcard cert, so simply changing the root to the correct one does nothing. 

I agree that this isn't a proper fix, but nobody was going to get one done last night.

worktycho:
This was a emergency fix, so that people could start using the server again. I agree it is compleatly broken, but we'd been trying to get a proper fix out for most of the afternoon, and it didn't look like we'd get one soon, so I put this in as a quick change that would allow the server to accept logins until we got the wildcard cert working.

PR #2091, opened 2015-05-21
Placing buckets up to 25 blocks away #2059

### LINKED ISSUE #2059: Bucket operations can be done from a distance.
If you right-click a bucket while looking at a faraway block (up to about 25 blocks), you will place what you have in the bucket or suck up the liquid. You should have to be in touch range to do this.

Rakete1111:
Sorry for my misunderstanding,
this fixed the issue #2059

### DIFF
--- a/src/Items/ItemBucket.h
+++ b/src/Items/ItemBucket.h
@@ -80,6 +80,13 @@ public:
 			return false;
 		}
 		
+		// Check to see if destination block is too far away
+		// Reach Distance Multiplayer = 5 Blocks
+		if ((BlockPos.x - a_Player->GetPosX() > 5) || (BlockPos.z - a_Player->GetPosZ() > 5))
+		{
+			return false;
+		}
+
 		// Remove water / lava block (unless plugins disagree)
 		if (!a_Player->PlaceBlock(BlockPos.x, BlockPos.y, BlockPos.z, E_BLOCK_AIR, 0))
 		{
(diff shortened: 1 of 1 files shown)


src/Items/ItemBucket.h
```diff
@@ -80,6 +80,13 @@ class cItemBucketHandler :
 			return false;
 		}
 		
+		// Check to see if destination block is too far away
+		// Reach Distance Multiplayer = 5 Blocks
+		if ((BlockPos.x - a_Player->GetPosX() > 5) || (BlockPos.z - a_Player->GetPosZ() > 5))
```
bearbin: More brackets here would be good.

madmaxoft:
I'm not sure if this is the right way to check. Doesn't Vanilla use regular distance? You're using NY distance, so the reach is longer in the diagonals (and potentially infinite in the Y axis).

Also I remember something about Creative mode having a bit longer reach than Survival mode.

tigerw:
In the function `bool GetBlockFromTrace(cWorld * a_World, cPlayer * a_Player, Vector3i & a_BlockPos)` there is a `Vector3d End()`. It might be better to play with that value to control where the trace ends to prevent duplicate checks.

Rakete1111:
@madmaxoft Forgot the Y axis, you're right :)
@tigerw I'll do that
Thanks guys
@bearbin Were could I insert more brackets?

src/Items/ItemBucket.h
bearbin: Like so: 

    if (((BlockPos.x - a_Player->GetPosX()) > 5) || ((BlockPos.z - a_Player->GetPosZ()) > 5))

Rakete1111: OK thanks

madmaxoft: No need to, really

madmaxoft:
Note that the `Vector3<>` classes have a `Length` and `SqrLength` functions on them, that should simplify your math regarding the distance calculation.

If you decide to edit this PR, do so either by amending the existing commit, or in the end squash all the commits into a single one. You should also change the commit message to something more descriptive - "Added a limit to bucket reach" or something along those lines.

worktycho:
@bearbot signoff

tigerw:
Wait, what? Our suggestions haven't been realised yet :P

Rakete1111:
Still working on it :)
I'll probably create a new pull request for it

Issue #1980, opened 2015-05-09
Spectate gamemode head visible


FalkAlexander:
If a player goes into gamemode 3, the head of him is visible.

Screenshots: 
https://minefight.org/owncloud/index.php/s/uTbglnL6rYzkFGz
https://minefight.org/owncloud/index.php/s/Lk9uJv88yX8xkl8

PR #2103, opened 2015-05-22
Fixed players head visible when in spectator mode

beeduck:
Fix for #1980. It should be noted that my server will occasionally crash when BroadcastSpawnEntity() and BroadcastDestroyEntity() are called, which SetVisible() calls. I know these functions are used when monsters are killed as well.


bearbin:
Have you reported the crash in a separate issue?

beeduck:
I have now, #2119

Issue #2053, opened 2015-05-17
Player falls through the world if they connect to the server already in spectator mode


beeduck:
Found this bug while fixing issue #1980. I already have a fix implemented, just waiting to make a pull request for this issue after I've fixed #1980.

jammet:
Yeah, I've had this happen to me a few times, and only switching to creative mode and then /spawn'ing helped.

PR #2104, opened 2015-05-22
Fix for players falling through world when loading in as spectator

beeduck:
Fix for issue #2053. There were no checks for a player connecting to a server when they are already in spectator mode.


bearbin:
Looks good :)

PR #2107, opened 2015-05-23
Cows no longer spawn over water.

### LINKED ISSUE #2080: Cows spawn in oceans
It should probably be squids and other sea mobs, but when I fly across large ocean biomes, I see the server spawning groups of cows all over the place, instead.

bearbin:
Fixes #2080

### DIFF
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -1259,7 +1259,7 @@ bool cFinishGenPassiveMobs::TrySpawnAnimals(cChunkDesc & a_ChunkDesc, int a_RelX
 	}
 	if (
 		(BlockUnderFeet != E_BLOCK_GRASS) &&
-		((AnimalToSpawn == mtSheep) || (AnimalToSpawn == mtChicken) || (AnimalToSpawn == mtPig))
+		((AnimalToSpawn == mtRabbit) || (AnimalToSpawn == mtCow) || (AnimalToSpawn == mtSheep) || (AnimalToSpawn == mtChicken) || (AnimalToSpawn == mtPig))
 	)
 	{
 		return false;


NiLSPACE:
Maybe also rabbits?

bearbin:
Fixed.

src/Generating/FinishGen.cpp
```diff
@@ -1259,7 +1259,7 @@ bool cFinishGenPassiveMobs::TrySpawnAnimals(cChunkDesc & a_ChunkDesc, int a_RelX
 	}
 	if (
 		(BlockUnderFeet != E_BLOCK_GRASS) &&
-		((AnimalToSpawn == mtSheep) || (AnimalToSpawn == mtChicken) || (AnimalToSpawn == mtPig))
+		((AnimalToSpawn == mtRabbit) || (AnimalToSpawn == mtCow) || (AnimalToSpawn == mtSheep) || (AnimalToSpawn == mtChicken) || (AnimalToSpawn == mtPig))
```
worktycho: This is getting very long. Perhaps this needs to be refactored out into a properties file similar to BlockInfo?

NiLSPACE: Maybe some code from the mobspawner can be shared? (not the blockentity, but the global mob spawner https://github.com/mc-server/MCServer/blob/master/src/MobSpawner.cpp)

worktycho: That seems like a good idea, but I would probably refactor that method out a bit. Its ridiculously long.

bearbin: You could refactor it into a case statement. Would that be good?

worktycho: I was referring to MobSpawner::CanSpawnHere which is 200 lines long.

bearbin: Ah ok. Sharing some code might be good, and refactoring would certainly be a good idea.

bearbin:
Can this be merged, and another issue opened to refactor both methods?

PR #2108, opened 2015-05-23
TryGetHeight returns a bool

### LINKED ISSUE #2099: cWorld:TryGetHeight IsValid is a number
cWorld:TryGetHeight returns 1 when valid, and 0 when false. Because of this you can't use
```lua
if (IsValid) then
   -- stuff
end
```

bearbin:
should fix #2099


madmaxoft:
This must be an oversight on my part, sorry. Merge if compiles (it should :)

PR #2110, opened 2015-05-23
Fixed wrong copy/paste

worktycho:
Fixes CID 110962

### DIFF
--- a/src/OverridesSettingsRepository.cpp
+++ b/src/OverridesSettingsRepository.cpp
@@ -260,7 +260,7 @@ bool cOverridesSettingsRepository::DeleteValue(const AString & a_KeyName, const 
 	}
 	else
 	{
-		return m_Overrides->DeleteValue(a_KeyName, a_ValueName);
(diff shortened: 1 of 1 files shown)


src/OverridesSettingsRepository.cpp
```diff
@@ -260,7 +260,7 @@ bool cOverridesSettingsRepository::DeleteValue(const AString & a_KeyName, const
 	}
 	else
 	{
-		return m_Overrides->DeleteValue(a_KeyName, a_ValueName);
+		return m-Main->DeleteValue(a_KeyName, a_ValueName);
```
NiLSPACE: I assume it's `a_Main` not `a-Main`?

worktycho: Yep, fixing.

worktycho:
I think I need to give up using the web editor to make changes, I always make a smaller error that needs a new commit to be squashed in.

worktycho:
Fixed.

PR #2112, opened 2015-05-23
Add `reload <pluginname>`

### LINKED ISSUE #365: Add "reload <pluginname>" console command ($5)
The reload console command reloads all plugins. It is, however, possible to reload a single plugin, either by unloading and loading it, or by creating a special plugin-reload function.

This will be useful for developing plugins, since only the relevant plugin can be reloaded, thus making the operation faster and possibly automate-able through RCON.

bearbin:
Fixes #365

### DIFF
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -449,7 +449,16 @@ void cServer::ExecuteConsoleCommand(const AString & a_Cmd, cCommandOutputCallbac
 	}
 	else if (split[0] == "reload")
 	{
-		cPluginManager::Get()->ReloadPlugins();
+		if (split.size() > 1)
+		{
+			cPluginManager::Get()->UnloadPlugin(split[1]);
+			a_Output.Out("Plugin unloaded scheduled");
+			a_Output.Out(cPluginManager::Get()->LoadPlugin(split[1]) ? "Plugin loaded" : "Error occurred loading plugin");
+		}
+		else
+		{
+			cPluginManager::Get()->ReloadPlugins();
+		}
 		a_Output.Finished();
 		return;
 	}


src/Server.cpp
```diff
@@ -449,7 +449,16 @@ void cServer::ExecuteConsoleCommand(const AString & a_Cmd, cCommandOutputCallbac
 	}
 	else if (split[0] == "reload")
 	{
-		cPluginManager::Get()->ReloadPlugins();
+		if (split.size() > 1)
+		{
+			cPluginManager::Get()->UnloadPlugin(split[1]);
+			a_Output.Out("Plugin unloaded scheduled");
+			a_Output.Out(cPluginManager::Get()->LoadPlugin(split[1]) ? "Plugin loaded" : "Error occurred loading plugin");
```
tigerw: I'm not sure, but it seems that since UnloadPlugin is non-blocking, LoadPlugin may be called before its actually unloaded, and so the plugin would end up unloaded when the queue is processed, but not loaded again.

tigerw: Does anyone know why UnloadPlugin must use a queue?

bearbin:
This comment should be ignored.

src/Server.cpp
madmaxoft: Because unloading a plugin requires locking a mutex and since the console handler is already holding another mutex, it resulted in deadlocks in certain scenarios.

madmaxoft:
You need to go deeper. Add a `cPluginManager:ReloadPlugin()` function that adds the plugin to a queue of plugins to reload, and in the `cPluginManager:Tick()` function unload and then re-load those plugins.
Closing this PR, since it will have almost no code in common with the recommended implementation.

PR #2118, opened 2015-05-23
Fix gamemode not changing properly.

### LINKED ISSUE #604: Players can be left floating if gameMode changes from creative to survival
1. A player is flying in creative mode
2. The server is shut down and switched to survival
3. The player will still be flying and cannot disable flymode

The player is stuck in flymode forever.

bearbin:
Should fix #604

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -94,7 +94,7 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 
 	SetMaxHealth(MAX_HEALTH);
 	m_Health = MAX_HEALTH;
-	
+
 	m_LastPlayerListTime = std::chrono::steady_clock::now();
 	m_PlayerName = a_PlayerName;
 
@@ -106,7 +106,7 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 		SetPosY(World->GetSpawnY());
 		SetPosZ(World->GetSpawnZ());
 		SetBedPos(Vector3i(static_cast<int>(World->GetSpawnX()), static_cast<int>(World->GetSpawnY()), static_cast<int>(World->GetSpawnZ())));
-		
+
 		LOGD("Player \"%s\" is connecting for the first time, spawning at default world spawn {%.2f, %.2f, %.2f}",
 			a_PlayerName.c_str(), GetPosX(), GetPosY(), GetPosZ()
 		);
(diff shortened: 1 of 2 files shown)


src/Entities/Player.cpp
```diff
@@ -1231,6 +1231,21 @@ void cPlayer::SetGameMode(eGameMode a_GameMode)
...
 {
 	m_GameMode = a_GameMode;
+	
+	if (!(IsGameModeCreative() || IsGameModeSpectator()))
+	{
+		SetFlying(false);
+		SetCanFly(false);
+	}
+
+	if (IsGameModeSpectator() && IsVisible())
```
tigerw: I'm not sure if this is how spectator works, but if a player is in spectator mode and they aren't visible (normal presumably?), does that mean it will fall through to the else if and set them to be visible?

bearbin: You're right. This is a logic error.

bearbin: Fixed.

bearbin:
It seems my editor has automatically cleaned up some whitespace but it shouldn't be an issue.

src/Entities/Player.cpp
```diff
@@ -1231,6 +1218,30 @@ void cPlayer::SetGameMode(eGameMode a_GameMode)
...
+
+void cPlayer::SetCapabilities()
+{
+	if (!IsGameModeCreative() || IsGameModeSpectator())
+	{
+		SetFlying(false);
+		SetCanFly(false);
+	}
+
+	if (!IsGameModeSpectator() && !IsVisible())
```
tigerw: Would:

```cpp
if (IsGameModeSpectator())
{
    SetVisible(false);
}
else
{
    SetVisible(true);
}
```

work?

bearbin: Probably, I'm just not sure of the side-effects of calling SetVisible and calling it again for a player who's already in the state they are being set to may have unintended consequences (performance, client bugs, etc).

bearbin: I just checked and that code would work. Amending the commit now.

src/Entities/Player.cpp
```diff
@@ -1196,30 +1196,17 @@ void cPlayer::SetGameMode(eGameMode a_GameMode)
...
-	
+
 	if (m_GameMode == a_GameMode)
 	{
 		// Gamemode already set
 		return;
 	}
-	
+
 	m_GameMode = a_GameMode;
```
SafwatHalaby: beep

PR #2132, opened 2015-05-24
Do not merge - Incomplete optimization

SafwatHalaby:
Just an optimization that might turn out to be useless. Do not merge.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -159,11 +159,13 @@ bool cPath::IsSolid(const Vector3i & a_Location)
 	m_Chunk->GetBlockTypeMeta(RelX, a_Location.y, RelZ, BlockType, BlockMeta);
 	if ((BlockType == E_BLOCK_FENCE) || (BlockType == E_BLOCK_FENCE_GATE))
 	{
-		GetCell(a_Location + Vector3i(0, 1, 0))->m_IsSolid = true;  // Mobs will always think that the fence is 2 blocks high and therefore won't jump over.
+		a_Location.y += 1;
+		GetCell(a_Location)->m_IsSolid = true;  // Mobs will always think that the fence is 2 blocks high and therefore won't jump over.
 	}
-	if (BlockType == E_BLOCK_STATIONARY_WATER)
+	else if (BlockType == E_BLOCK_STATIONARY_WATER)
 	{
-		GetCell(a_Location + Vector3i(0, -1, 0))->m_IsSolid = true;
+		a_Location.y -= 1;
+		GetCell(a_Location)->m_IsSolid = true;
 	}
 
 	return cBlockInfo::IsSolid(BlockType);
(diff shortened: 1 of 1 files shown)


worktycho:
I can't say for certain without looking at the output, but I would expect the compiler to optimize that. If you're using clang the --emit-llvm flag can be interesting reading.

SafwatHalaby:
I am almost sure it does not. I can't imagine it's intelligent enough to decide that it does not need to construct a new object. It's a logic-level thing.

worktycho:
For clang at least, once it reaches the optimizer, the constructor is just a function. And so if you inline the constructor, it becomes a load of moves, which can be canceled, and some zero adds, which are easily removed.

SafwatHalaby:
Yes, zero adds are removed, but 3 brand new x,y,z's are needlessly allocated.

worktycho:
Its a stack allocation. And most optimizers are capable of removing unused stack allocations, in order to deal with register promotions.

SafwatHalaby:
Interesting. What about copying?
In the old version, Whenever we do CurrentCell->m_Location + vector(...), We create 1 new vector which houses the result (and 1 new vector which is the right hand side of the +).

worktycho:
Its the same thing, once its inlined its just a bunch of moves.

SafwatHalaby:
Yes but is it smart enough to optimize a bunch of needless but intertwined copy moves?
I guess we should look at the output.

worktycho:
Yep, reading the llvm IR can be interesting when optimizing.

src/Mobs/Path.cpp
```diff
@@ -159,11 +159,13 @@ bool cPath::IsSolid(const Vector3i & a_Location)
 	m_Chunk->GetBlockTypeMeta(RelX, a_Location.y, RelZ, BlockType, BlockMeta);
 	if ((BlockType == E_BLOCK_FENCE) || (BlockType == E_BLOCK_FENCE_GATE))
 	{
-		GetCell(a_Location + Vector3i(0, 1, 0))->m_IsSolid = true;  // Mobs will always think that the fence is 2 blocks high and therefore won't jump over.
+		a_Location.y += 1;
+		GetCell(a_Location)->m_IsSolid = true;  // Mobs will always think that the fence is 2 blocks high and therefore won't jump over.
 	}
-	if (BlockType == E_BLOCK_STATIONARY_WATER)
+	else if (BlockType == E_BLOCK_STATIONARY_WATER)
```
worktycho: This changes the semantic meaning of the code, and is definitely worth having.

worktycho:
I double checked this. Because of the way register machines work, on x86 with clang 3.6 at least, they both lower to assembly that is very similar, but the current version results in an extra stack allocation. This means it has slightly worse cache usage, but executes the exact same number/sequence of instructions.

tigerw:
> register machines work

Who knew tills could be so advanced?

worktycho:
As apposed to stack, accumulator, or belt computers.

SafwatHalaby:
"The current version", as in the Master version?

worktycho:
Actually by current version I meant a slightly modified version of master, that has the control flow changes, and takes the parameter by value. This was so I could actually look at the changes caused by the objects only.

SafwatHalaby:
How could the non-master version produce *MORE* stack allocations?

SafwatHalaby:
So, in conclusion. Is this PR is useless? -with the exception of the `else if` change-

worktycho:
master produces more, but its not really worth worrying about. ANd yes the PR is useless unless this function shows up in a profiler.

PR #2136, opened 2015-05-25
Fixed ClientHandle not getting properly removed from the world

### LINKED ISSUE #1969: Server crashes if cPlayer is deleted during the adding procedure.
There is a race condition in the server that for a sufficiently loaded server, if the connection is closed during the player spawn process the server will crash as cWorld::HasEntity returns true for Entitys in the queue, but cWorld::Player does not remove it from. This causes the server to crash due to an ASSERT in cEntity.

I achieved this situation with a c++ program that simulates the bare minimum of a client to login, and attempts to run a login against the server every 10 ms.

worktycho:
when socket is closed whilst the client is being added to the world
Fixes #1969

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2957,13 +2957,20 @@ void cClientHandle::PacketError(UInt32 a_PacketType)
 void cClientHandle::SocketClosed(void)
 {
 	// The socket has been closed for any reason
-	
+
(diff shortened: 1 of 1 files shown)


worktycho:
Removed the lines

madmaxoft:
Thanks, could you squash the commits?

worktycho:
Squashed

Issue #1650, opened 2014-12-05
Add API to query server uptime


madmaxoft:
It would be nice to be able to query the server's uptime platform-independently. Plugins could then display it on the webpage or let the player use a command for the query.

Of course, plugins could already measure the uptime with rough precision by counting server ticks, but that's an ugly workaround.

PR #2138, opened 2015-05-25
Added method to get the uptime of the server in seconds. 

Seadragon91:
This implements #1650. All other changes are from my atom editor. It trims all lines who are empty, containing spaces/tabs, if that is not okay, I can correct it.

### DIFF
--- a/src/Root.h
+++ b/src/Root.h
@@ -70,22 +70,31 @@ public:
 	a_OverworldName should be set for non-overworld dimensions if one wishes that world to link back to an overworld via portals
 	*/
 	cWorld * CreateAndInitializeWorld(const AString & a_WorldName, eDimension a_Dimension = dimOverworld, const AString & a_OverworldName = "");
+
+	/** Returns the up time of the server in seconds */
+	int GetServerUpTime(void)
+	{
+		return static_cast<int>(std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - m_StartTime).count());
+	}
 	// tolua_end
-	
+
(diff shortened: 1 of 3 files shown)


SphinxC0re:
Looks good to me. Any complains?

src/Root.h
```diff
@@ -70,22 +70,31 @@ class cRoot
 	a_OverworldName should be set for non-overworld dimensions if one wishes that world to link back to an overworld via portals
 	*/
 	cWorld * CreateAndInitializeWorld(const AString & a_WorldName, eDimension a_Dimension = dimOverworld, const AString & a_OverworldName = "");
+
+	/** Returns the up time of the server in seconds */
+	long int GetServerUpTime(void)
+	{
+		return static_cast<long int>(std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - m_StartTime).count());
```
worktycho: I'd rather you didn't use long as it causes too many issues with differences between windows and linux, choose either long long, or int.

Seadragon91:
A `int` should be enough. I don't think that someone can run the server with a up time of 68 years :-)

madmaxoft:
I never understood the zealous IDEs that try to remove all the whitespace

Seadragon91:
@madmaxoft Should I disabled the option, that trims the lines of spaces/tabs? It's no problem for me.

madmaxoft:
I'd prefer that, the commits are a mess of useless whitespace changes this way and it's difficult to navigate useful code. Github does have the option to ignore whitespace in the diffs, but then I cannot add comments in such view, so it's pain to use.

PR #2144, opened 2015-05-26
Redstone improvements [SEE DESC]

### LINKED ISSUE #1083: Redstone wires arent updated within 1 Tick
As i tested the redstone simulator, i found the bug, that a line from 15 redstone isn't udated within 1 tick, see the following gif. (this isnt a copy from #860, because this bug is that redstone isnt updated within 1 tick and #860 is that redstone isnt updated properly)
9hseg

### LINKED ISSUE #1763: Redstone dust doesn't work properly.
When you have a redstone torch powering a line of redstone dust, after disabling the torch, the dust flashes on and off really fast untill it's nearly shut off, but it doesn't shut off because there's still some dust being powered somehow.

It's hard to explain, you need to try it yourself.

Edit:
the Windows x64 build 178 of the server, and I'm running Windows 7 ulti. x64

### LINKED ISSUE #1887: Trapdoors "forget" open/closed state upon server restart
Many times they are used for decoration -

 

... and this is what it looks like when the state they were left in wasn't saved, or saved correctly.
Look at that sad little lantern.... ;)

### LINKED ISSUE #2009: Crossing redstone dust chunk bug
If redstone dust paths are crossing each other, then the dust isn't switching off correctly after turning off the lever. This bug is happening if the redstone dust path crosses chunk borders. Maybe it is related to the direction, so we've found out it's often not working with: south-north; east-west; west-east.

Video: http://youtu.be/d4GnuUxfcZE

### LINKED ISSUE #2191: Redstone not gates not working properly with repeaters nearby
These two are perfectly valid clocks in Vanilla. In MCServer, they simply won't work. I recall a similar issue a very long time ago. Back then, Tiger fixed it, I think.

tigerw:
+ Partially implemented comparators
+ Wires now power on/off simultaneously
* Potential performance improvements
* Improvements to linked powering
* Fixes #1887
* Fixes #1763
* Fixes #1083
* Fixes #2009
* Fixes #2191

### DIFF
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -96,6 +96,19 @@ public:
 		a_Z = a_Z - a_ChunkZ * Width;
 	}
 
+	inline static Vector3i AbsoluteToRelative(Vector3i a_BlockPosition)
+	{
+		int ChunkX, ChunkZ;
+		BlockToChunk(a_BlockPosition.x, a_BlockPosition.z, ChunkX, ChunkZ);
+
+		return {a_BlockPosition.x - ChunkX * Width, a_BlockPosition.y, a_BlockPosition.z - ChunkZ * Width};
+	}
+
+	inline static Vector3i AbsoluteToRelative(Vector3i a_BlockPosition, int a_ChunkX, int a_ChunkZ)
+	{
+		return {a_BlockPosition.x - a_ChunkX * Width, a_BlockPosition.y, a_BlockPosition.z - a_ChunkZ * Width};
+	}
+
 	/** Converts relative block coordinates into absolute coordinates with a known chunk location */
 	inline static Vector3i RelativeToAbsolute(const Vector3i & a_RelBlockPosition, int a_ChunkX, int a_ChunkZ)
 	{
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -96,6 +96,12 @@ public:
 		a_Z = a_Z - a_ChunkZ * Width;
 	}
 
+	/** Converts relative block coordinates into absolute coordinates with a known chunk location */
+	inline static Vector3i RelativeToAbsolute(const Vector3i & a_RelBlockPosition, int a_ChunkX, int a_ChunkZ)
+	{
+		return Vector3i(a_RelBlockPosition.x + a_ChunkX * Width, a_RelBlockPosition.y, a_RelBlockPosition.z + a_ChunkZ * Width);
+	}
+
 
 	/// Converts absolute block coords to chunk coords:
 	inline static void BlockToChunk(int a_X, int a_Z, int & a_ChunkX, int & a_ChunkZ)
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -514,32 +514,3 @@ typedef cCoordWithData<BLOCKTYPE>  cCoordWithBlock;
 
 typedef std::list<cCoordWithInt>   cCoordWithIntList;
 typedef std::vector<cCoordWithInt> cCoordWithIntVector;
-
-
-
-
-
-/** Generic template that can store two types of any kind of data together with a triplet of 3 coords */
-template <typename X, typename Z> class cCoordWithDoubleData
-{
-public:
-	int x;
-	int y;
-	int z;
-	X   Data;
-	Z   DataTwo;
-
-	cCoordWithDoubleData(int a_X, int a_Y, int a_Z) :
-		x(a_X), y(a_Y), z(a_Z)
-	{
-	}
-
-	cCoordWithDoubleData(int a_X, int a_Y, int a_Z, const X & a_Data, const Z & a_DataTwo) :
-		x(a_X), y(a_Y), z(a_Z), Data(a_Data), DataTwo(a_DataTwo)
-	{
-	}
-};
-
-typedef cCoordWithDoubleData <BLOCKTYPE, bool> cCoordWithBlockAndBool;
-
-typedef std::vector<cCoordWithBlockAndBool> cCoordWithBlockAndBoolVector;
--- a/src/Simulator/IncrementalRedstoneSimulator.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator.cpp
@@ -640,16 +640,26 @@ void cIncrementalRedstoneSimulator::FindAndPowerBorderingWires(std::vector<std::
 		auto Neighbour = a_Entry.second->GetRelNeighborChunkAdjustCoords(AdjustedPos.x, AdjustedPos.z);
 		auto MyPower = IsWirePowered(a_Entry.first, a_Entry.second);
 
+		auto & PoweredBlocks = ((cIncrementalRedstoneSimulatorChunkData *)Neighbour->GetRedstoneSimulatorData())->m_PoweredBlocks;  // We need to insert the value into the chunk who owns the block position
+		sPoweredBlocks RC;
+		RC.a_BlockPos = AdjustedPos;
+		RC.a_SourcePos = a_Entry.first + Vector3i((a_Entry.second->GetPosX() - Neighbour->GetPosX()) * cChunkDef::Width, 0, (a_Entry.second->GetPosZ() - Neighbour->GetPosZ()) * cChunkDef::Width);
+		RC.a_PowerLevel = MyPower - 1;
+
 		if ((Neighbour->GetBlock(AdjustedPos) == E_BLOCK_REDSTONE_WIRE) && (MyPower > 1) && (MyPower > IsWirePowered(AdjustedPos, Neighbour)))
 		{
-			sPoweredBlocks RC;
-			RC.a_BlockPos = AdjustedPos;
-			RC.a_SourcePos = a_Entry.first + Vector3i((a_Entry.second->GetPosX() - Neighbour->GetPosX()) * cChunkDef::Width, 0, (a_Entry.second->GetPosZ() - Neighbour->GetPosZ()) * cChunkDef::Width);
-			RC.a_PowerLevel = MyPower - 1;
-			((cIncrementalRedstoneSimulatorChunkData *)Neighbour->GetRedstoneSimulatorData())->m_PoweredBlocks.emplace_back(RC);  // We need to insert the value into the chunk who owns the block position
+			auto Position = std::find_if(PoweredBlocks.begin(), PoweredBlocks.end(), [RC](const sPoweredBlocks & itr) { return itr.a_BlockPos == RC.a_BlockPos && itr.a_SourcePos == RC.a_SourcePos; });
+			if (Position != PoweredBlocks.end())
+			{
+				Position->a_PowerLevel = RC.a_PowerLevel;
+			}
+			else
+			{
+				PoweredBlocks.emplace_back(RC);
 
-			Neighbour->SetIsRedstoneDirty(true);
-			m_Chunk->SetIsRedstoneDirty(true);
+				Neighbour->SetIsRedstoneDirty(true);
+				m_Chunk->SetIsRedstoneDirty(true);
+			}
 
 			Neighbour->SetMeta(AdjustedPos.x, AdjustedPos.y, AdjustedPos.z, MyPower - 1);
 			a_PotentialWireList.emplace_back(std::make_pair(AdjustedPos, Neighbour));
--- a/src/Simulator/IncrementalRedstoneSimulator.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator.cpp
@@ -24,41 +24,42 @@
 
 
 
-void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk, cChunk * a_OtherChunk)
+void cIncrementalRedstoneSimulator::AddBlock(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk)
 {
-	if ((a_Chunk == nullptr) || !a_Chunk->IsValid())
+	AddBlock(cChunkDef::AbsoluteToRelative({ a_BlockX, a_BlockY, a_BlockZ }, a_Chunk->GetPosX(), a_Chunk->GetPosZ()), a_Chunk, nullptr);
+}
+
+
+
+
+
+void cIncrementalRedstoneSimulator::AddBlock(const Vector3i & a_RelBlockPosition, cChunk * a_OriginalChunk, cChunk * a_NeighborChunk)
+{
+	if ((a_OriginalChunk == nullptr) || !a_OriginalChunk->IsValid())
 	{
 		return;
 	}
-	else if ((a_BlockY < 0) || (a_BlockY >= cChunkDef::Height))
+	else if ((a_RelBlockPosition.y < 0) || (a_RelBlockPosition.y >= cChunkDef::Height))
 	{
 		return;
 	}
 
-	// We may be called with coordinates in a chunk that is not the first chunk parameter
-	// In that case, the actual chunk (which the coordinates are in), will be passed as the second parameter
-	// Use that Chunk pointer to get a relative position
+	// The relative block position is relative to the neighboring chunk should it be passed as an argument
 
-	int RelX = 0;
-	int RelZ = 0;
 	BLOCKTYPE Block;
 	NIBBLETYPE Meta;
 
-	if (a_OtherChunk != nullptr)
+	if (a_NeighborChunk != nullptr)
 	{
-		RelX = a_BlockX - a_OtherChunk->GetPosX() * cChunkDef::Width;
-		RelZ = a_BlockZ - a_OtherChunk->GetPosZ() * cChunkDef::Width;
-		a_OtherChunk->GetBlockTypeMeta(RelX, a_BlockY, RelZ, Block, Meta);
+		a_NeighborChunk->UnboundedRelGetBlock(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, Block, Meta);
 
-		// If a_OtherChunk is passed (not nullptr), it is the chunk that had a block change, and a_Chunk will be the neighbouring chunk of that block
+		// If a_OtherChunk is passed (not nullptr), it is the neighbouring chunk of a_Chunk, which itself is the chunk with the block change
 		// Because said neighbouring chunk does not know of this change but still needs to update its redstone, we set it to dirty
-		a_Chunk->SetIsRedstoneDirty(true);
+		a_NeighborChunk->SetIsRedstoneDirty(true);
 	}
 	else
 	{
-		RelX = a_BlockX - a_Chunk->GetPosX() * cChunkDef::Width;
-		RelZ = a_BlockZ - a_Chunk->GetPosZ() * cChunkDef::Width;
-		a_Chunk->GetBlockTypeMeta(RelX, a_BlockY, RelZ, Block, Meta);
+		a_OriginalChunk->GetBlockTypeMeta(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, Block, Meta);
 	}
 
 	// Every time a block is changed (AddBlock called), we want to go through all lists and check to see if the coordiantes stored within are still valid
@@ -76,39 +77,53 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
 		)
 	)
 	{
-		SetSourceUnpowered(RelX, a_BlockY, RelZ, a_OtherChunk != nullptr ? a_OtherChunk : a_Chunk);
+		SetSourceUnpowered(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_NeighborChunk != nullptr ? a_NeighborChunk : a_OriginalChunk);
 	}
 
 	if (!IsViableMiddleBlock(Block))
 	{
-		SetInvalidMiddleBlock(RelX, a_BlockY, RelZ, a_OtherChunk != nullptr ? a_OtherChunk : a_Chunk);
+		SetInvalidMiddleBlock(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_NeighborChunk != nullptr ? a_NeighborChunk : a_OriginalChunk);
 	}
 
-	auto & SimulatedPlayerToggleableBlocks = ((cIncrementalRedstoneSimulator::cIncrementalRedstoneSimulatorChunkData *)a_Chunk->GetRedstoneSimulatorData())->m_SimulatedPlayerToggleableBlocks;
-	SimulatedPlayerToggleableBlocks.erase(std::remove_if(SimulatedPlayerToggleableBlocks.begin(), SimulatedPlayerToggleableBlocks.end(), [RelX, a_BlockY, RelZ, Block, this](const sSimulatedPlayerToggleableList & itr)
-		{
-			return itr.a_RelBlockPos.Equals(Vector3i(RelX, a_BlockY, RelZ)) && !IsAllowedBlock(Block);
-		}
-	), SimulatedPlayerToggleableBlocks.end());
-
-	
-	auto & RepeatersDelayList = ((cIncrementalRedstoneSimulator::cIncrementalRedstoneSimulatorChunkData *)a_Chunk->GetRedstoneSimulatorData())->m_RepeatersDelayList;
-	RepeatersDelayList.erase(std::remove_if(RepeatersDelayList.begin(), RepeatersDelayList.end(), [RelX, a_BlockY, RelZ, Block](const sRepeatersDelayList & itr)
-		{
-			return itr.a_RelBlockPos.Equals(Vector3i(RelX, a_BlockY, RelZ)) && (Block != E_BLOCK_REDSTONE_REPEATER_ON) && (Block != E_BLOCK_REDSTONE_REPEATER_OFF);
-		}
-	), RepeatersDelayList.end());
-
-	if (a_OtherChunk != nullptr)
+	if (a_NeighborChunk != nullptr)
 	{
 		// DO NOT touch our chunk's data structure if we are being called with coordinates from another chunk - this one caused me massive grief :P
 		return;
 	}
 
-	cCoordWithBlockAndBoolVector & RedstoneSimulatorChunkData = ((cIncrementalRedstoneSimulator::cIncrementalRedstoneSimulatorChunkData *)a_Chunk->GetRedstoneSimulatorData())->m_ChunkData;
+	auto & SimulatedPlayerToggleableBlocks = static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_SimulatedPlayerToggleableBlocks;
+	if (
+		DoesIgnorePlayerToggle(Block) &&
+		std::find_if(SimulatedPlayerToggleableBlocks.begin(),SimulatedPlayerToggleableBlocks.end(),
+			[a_RelBlockPosition](const sSimulatedPlayerToggleableList & itr){ return itr.a_RelBlockPos == a_RelBlockPosition; }) == SimulatedPlayerToggleableBlocks.end()
+		)
+	{
+		// We have arrived here; no block must be in list - add one
+		sSimulatedPlayerToggleableList RC;
+		RC.a_RelBlockPos = a_RelBlockPosition;
+		RC.WasLastStatePowered = AreCoordsDirectlyPowered(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_OriginalChunk) || AreCoordsLinkedPowered(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_OriginalChunk);
+
+		// Initialise the toggleable blocks list so that trapdoors etc. aren't reset on restart (#1887)
+		SimulatedPlayerToggleableBlocks.emplace_back(RC);
+	}
+
+	SimulatedPlayerToggleableBlocks.erase(std::remove_if(SimulatedPlayerToggleableBlocks.begin(), SimulatedPlayerToggleableBlocks.end(), [a_RelBlockPosition, Block, this](const sSimulatedPlayerToggleableList & itr)
+		{
+			return (itr.a_RelBlockPos == a_RelBlockPosition) && !DoesIgnorePlayerToggle(Block);
+		}
+	), SimulatedPlayerToggleableBlocks.end());
+	
+	auto & RepeatersDelayList = static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_RepeatersDelayList;
+	RepeatersDelayList.erase(std::remove_if(RepeatersDelayList.begin(), RepeatersDelayList.end(), [a_RelBlockPosition, Block](const sRepeatersDelayList & itr)
+		{
+			return (itr.a_RelBlockPos == a_RelBlockPosition) && (Block != E_BLOCK_REDSTONE_REPEATER_ON) && (Block != E_BLOCK_REDSTONE_REPEATER_OFF);
+		}
+	), RepeatersDelayList.end());
+
+	auto & RedstoneSimulatorChunkData = static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_ChunkData;
 	for (auto & itr : RedstoneSimulatorChunkData)
 	{
-		if ((itr.x == RelX) && (itr.y == a_BlockY) && (itr.z == RelZ))  // We are at an entry matching the current (changed) block
+		if (a_RelBlockPosition == Vector3i(itr.x, itr.y, itr.z))  // We are at an entry matching the current (changed) block
 		{
 			if (!IsAllowedBlock(Block))
 			{
@@ -128,16 +143,16 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
 		return;
 	}
 
-	cCoordWithBlockAndBoolVector & QueuedData = ((cIncrementalRedstoneSimulator::cIncrementalRedstoneSimulatorChunkData *)a_Chunk->GetRedstoneSimulatorData())->m_QueuedChunkData;
+	auto & QueuedData = static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_QueuedChunkData;
 	for (const auto & itr : QueuedData)
 	{
-		if ((itr.x == RelX) && (itr.y == a_BlockY) && (itr.z == RelZ))
+		if (a_RelBlockPosition == Vector3i(itr.x, itr.y, itr.z))
 		{
 			// Can't have duplicates in here either, in case something adds the block again before the structure can written to the main chunk data
 			return;
 		}
 	}
-	QueuedData.emplace_back(cCoordWithBlockAndBool(RelX, a_BlockY, RelZ, Block, false));
+	QueuedData.emplace_back(cCoordWithBlockAndBool(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, Block, false));
 }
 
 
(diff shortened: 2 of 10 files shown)


NiLSPACE:
And we're back with the `[SEE DESC]` PR's ;)

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
+	} CCB;
+
+	auto AbsoluteComparatorCoords = cChunkDef::RelativeToAbsolute(Vector3i(a_RelBlockX, a_RelBlockY, a_RelBlockZ), m_Chunk->GetPosX(), m_Chunk->GetPosZ());
+	auto AbsoluteRearCoords = cBlockComparatorHandler::GetRearCoordinate(AbsoluteComparatorCoords.x, AbsoluteComparatorCoords.y, AbsoluteComparatorCoords.z, Meta & 0x3);
+
+	m_Chunk->DoWithBlockEntityAt(AbsoluteRearCoords.x, AbsoluteRearCoords.y, AbsoluteRearCoords.z, CCB);
+	auto RearPower = std::max(CCB.m_SignalStrength, IsWirePowered(AdjustRelativeCoords(cBlockComparatorHandler::GetRearCoordinate(a_RelBlockX, a_RelBlockY, a_RelBlockZ, Meta & 0x3)), m_Chunk));
+
+	if ((Meta & 0x4) == 0x4)
+	{  // Subtraction mode
```
NiLSPACE: Are comments on the braces allowed?

src/ChunkDef.h
```diff
@@ -96,6 +96,25 @@ class cChunkDef
...
+
+	inline static Vector3i AbsoluteToRelative(Vector3i a_BlockPosition, int a_ChunkX, int a_ChunkZ)
+	{
+		return {a_BlockPosition.x - a_ChunkX * Width, a_BlockPosition.y, a_BlockPosition.z - a_ChunkZ * Width};
+	}
+
+	/** Converts relative block coordinates into absolute coordinates with a known chunk location */
+	inline static Vector3i RelativeToAbsolute(const Vector3i & a_RelBlockPosition, int a_ChunkX, int a_ChunkZ)
+	{
+		return Vector3i(a_RelBlockPosition.x + a_ChunkX * Width, a_RelBlockPosition.y, a_RelBlockPosition.z + a_ChunkZ * Width);
```
worktycho: Why does this function use function style initialization when the other two use uniform initialization?

src/Simulator/IncrementalRedstoneSimulator.cpp
madmaxoft: Not really

src/Simulator/IncrementalRedstoneSimulator.h
```diff
@@ -348,47 +391,30 @@ class cIncrementalRedstoneSimulator :
...
-		while (RelPos.z < 0)
-		{
-			RelPos.z += cChunkDef::Width;
-		}
-
-		return RelPos;
+	inline static Vector3i AdjustRelativeCoords(const Vector3i & a_RelPosition)
+	{
+		return Vector3i((a_RelPosition.x % cChunkDef::Width + cChunkDef::Width) % cChunkDef::Width, a_RelPosition.y, (a_RelPosition.z % cChunkDef::Width + cChunkDef::Width) % cChunkDef::Width);
 	}
```
worktycho: What does this do exactly? `x % y + y % y == x %y`

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -77,39 +77,53 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
...
+		SetInvalidMiddleBlock(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_NeighborChunk != nullptr ? a_NeighborChunk : a_OriginalChunk);
+	}
+
+	if (a_NeighborChunk != nullptr)
+	{
+		// DO NOT touch our chunk's data structure if we are being called with coordinates from another chunk - this one caused me massive grief :P
+		return;
+	}
+
+	auto & SimulatedPlayerToggleableBlocks = static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_SimulatedPlayerToggleableBlocks;
```
worktycho: Since you only every query it for blocks, should m_SimulatedPlayerToggleableBlocks be a map of some kind?

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -25,41 +24,42 @@ typedef cItemCallback<cChestEntity> cChestCallback;
...
 {
-	if ((a_Chunk == nullptr) || !a_Chunk->IsValid())
+	AddBlock(cChunkDef::AbsoluteToRelative({ a_BlockX, a_BlockY, a_BlockZ }, a_Chunk->GetPosX(), a_Chunk->GetPosZ()), a_Chunk, nullptr);
+}
+
+
+
+
+
+void cIncrementalRedstoneSimulator::AddBlock(const Vector3i & a_RelBlockPosition, cChunk * a_OriginalChunk, cChunk * a_NeighborChunk)
```
worktycho: This method is getting too big. Break it up into smaller submethods.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
 
 void cIncrementalRedstoneSimulator::HandleRedstoneWire(int a_RelBlockX, int a_RelBlockY, int a_RelBlockZ)
```
worktycho: This method is too big, so if you're going to make significant changes, split it up

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
-			continue;
+			return;
 		}
-		if (Type == E_BLOCK_REDSTONE_WIRE)
+
+		if ((Neighbour->GetBlock(AdjustedPos) == E_BLOCK_REDSTONE_WIRE) && (MyPower > 1) && (MyPower > IsWirePowered(AdjustedPos, Neighbour)))
 		{
-			SetBlockPowered(a_RelBlockX + gCrossCoords[i].x, a_RelBlockY + gCrossCoords[i].y, a_RelBlockZ + gCrossCoords[i].z, a_RelBlockX, a_RelBlockY, a_RelBlockZ, MyPower);
+			PowerBorderingWires(a_PotentialWireList, a_Entry, AdjustedPos, Neighbour, MyPower);
 		}
```
worktycho: This bock should be moved into a seperate method, so you don't repeat yourself for all three loops

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
-		case REDSTONE_Z_POS:
+	}
+}
+
+
+
+
+
+void cIncrementalRedstoneSimulator::PowerBorderingWires(std::vector<std::pair<Vector3i, cChunk *>> & a_PotentialWireList, const std::pair<Vector3i, cChunk *> & a_Entry, const Vector3i & a_AdjustedPos, cChunk * a_NeighbourChunk, unsigned char a_MyPower)
```
worktycho: a_Entry has a weird type. Why not separate params?

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
+{
+	sPoweredBlocks RC;
```
worktycho: Why construct RC if you never use it as an object, only a collection of fields?

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
+}
+
+
+
+
+
+void cIncrementalRedstoneSimulator::HandleRedstoneComparator(int a_RelBlockX, int a_RelBlockY, int a_RelBlockZ)
+{
+	auto Meta = m_Chunk->GetMeta(a_RelBlockX, a_RelBlockY, a_RelBlockZ);
+	auto HighestSidePower = std::max(IsWirePowered(AdjustRelativeCoords(cBlockComparatorHandler::GetSideCoordinate(a_RelBlockX, a_RelBlockY, a_RelBlockZ, Meta & 0x3, false)), m_Chunk), IsWirePowered(AdjustRelativeCoords(cBlockComparatorHandler::GetSideCoordinate(a_RelBlockX, a_RelBlockY, a_RelBlockZ, Meta & 0x3, true)), m_Chunk));
```
worktycho: This line is too long, split it up into separate vars.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
+			return false;
 		}
+
+		unsigned char m_SignalStrength;
+	} CCB;
+
+	auto AbsoluteComparatorCoords = cChunkDef::RelativeToAbsolute(Vector3i(a_RelBlockX, a_RelBlockY, a_RelBlockZ), m_Chunk->GetPosX(), m_Chunk->GetPosZ());
+	auto AbsoluteRearCoords = cBlockComparatorHandler::GetRearCoordinate(AbsoluteComparatorCoords.x, AbsoluteComparatorCoords.y, AbsoluteComparatorCoords.z, Meta & 0x3);
+
+	m_Chunk->DoWithBlockEntityAt(AbsoluteRearCoords.x, AbsoluteRearCoords.y, AbsoluteRearCoords.z, CCB);
```
worktycho: I think it would be better to refactor DoWithBlockEntityAt to have an overload that takes a lambda, than to write yet another functor object.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +514,280 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
+	auto RearPower = std::max(CCB.m_SignalStrength, IsWirePowered(AdjustRelativeCoords(cBlockComparatorHandler::GetRearCoordinate(a_RelBlockX, a_RelBlockY, a_RelBlockZ, Meta & 0x3)), m_Chunk));
```
worktycho: I think the container fetching logic should be moved out into a separate method.

src/Simulator/IncrementalRedstoneSimulator.h
tigerw: Isn't it supposed to be ( x % y + y) % y? Since normal mod doesn't produce the wanted result on negative numbers (i.e. -5 % 16 should be 11 for relative adjustment but instead is -5), this is done.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Then it should be.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: But I have already, with the FindAndPowerBorderingWires and PowerBorderingWires, no?

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Does the emplace_back(RC) not count? Or should I use emplace_back to construct an RC?

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: What would you mean by separate?

src/Simulator/IncrementalRedstoneSimulator.h
worktycho: That's what I meant. But you don't need the inner mod, and you should put a comment about negative numbers.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Its still 70 lines of actual code, which is a bit on the large size.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Pass a vector3i, and a cChunk * as individual parameters to the function.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Given that its one emplace in an if, I would go for using emplace_back to construct it.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -680,19 +819,50 @@ void cIncrementalRedstoneSimulator::HandleRedstoneRepeater(int a_RelBlockX, int
...
 		else if (!IsSelfPowered && IsOn)  // Queue a power change if unpowered, on, and not locked.
 		{
-			QueueRepeaterPowerChange(a_RelBlockX, a_RelBlockY, a_RelBlockZ, a_Meta, false);
+			QueueRepeaterPowerChange(a_RelBlockX, a_RelBlockY, a_RelBlockZ, Meta, false);
+		}
+	}
+
+	if (IsOn)
+	{
+		switch (Meta & 0x3)  // We only want the direction (bottom) bits
```
worktycho: This block should be in its own method

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -1702,35 +1832,35 @@ bool cIncrementalRedstoneSimulator::IsPistonPowered(int a_RelBlockX, int a_RelBl
...
 
 
-bool cIncrementalRedstoneSimulator::IsWirePowered(int a_RelBlockX, int a_RelBlockY, int a_RelBlockZ, unsigned char & a_PowerLevel)
+unsigned char cIncrementalRedstoneSimulator::IsWirePowered(Vector3i a_RelBlockPosition, cChunk * a_Chunk)
 {
-	a_PowerLevel = 0;
+	unsigned char PowerLevel = 0;
 
-	for (const auto & itr : *m_PoweredBlocks)  // Check powered list
+	for (const auto & itr : static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_Chunk->GetRedstoneSimulatorData())->m_PoweredBlocks)  // Check powered list
```
worktycho: You should change this list to support querying for a source better.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -1702,35 +1832,35 @@ bool cIncrementalRedstoneSimulator::IsPistonPowered(int a_RelBlockX, int a_RelBl
...
+		if (itr.a_BlockPos != a_RelBlockPosition)
 		{
 			continue;
 		}
-		a_PowerLevel = std::max(itr.a_PowerLevel, a_PowerLevel);  // Get the highest power level (a_PowerLevel is initialised already and there CAN be multiple levels for one block)
+		PowerLevel = std::max(itr.a_PowerLevel, PowerLevel);  // Get the highest power level (a_PowerLevel is initialised already and there CAN be multiple levels for one block)
 	}
 
-	for (const auto & itr : *m_LinkedPoweredBlocks)  // Check linked powered list
+	for (const auto & itr : static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_Chunk->GetRedstoneSimulatorData())->m_LinkedBlocks)  // Check linked powered list
```
worktycho: Again this should be changed for better querying.

worktycho:
@tigerw Can you write up a post on the forum about the general architecture of the simulator. Because no one else really understands it, so we can't review your code.

tigerw:
😜

I will try.





From: worktycho
Sent: ‎Thursday‎, ‎28‎ ‎May‎ ‎2015 ‎12‎:‎59
To: mc-server/MCServer
Cc: Tiger Wang





@tigerw Can you write up a post on the forum about the general architecture of the simulator. Because no one else really understands it, so we can't review your code.

tigerw:
Do you think it is a good idea, to reduce the size of IncrementalSim.cpp, to place the various redstone handler functions in their respective BlockHandlers?





From: worktycho
Sent: ‎Thursday‎, ‎28‎ ‎May‎ ‎2015 ‎12‎:‎59
To: mc-server/MCServer
Cc: Tiger Wang





@tigerw Can you write up a post on the forum about the general architecture of the simulator. Because no one else really understands it, so we can't review your code.

worktycho:
I think its a good idea to move things out of IncrementalSim, but I would not move them to the blockHandlers. I would create a separate set of redstoneHandler classes.

worktycho:
@tigerw Can you rebase this onto master?

tigerw:
Yes, though I must need address your comments first.

src/Simulator/IncrementalRedstoneSimulator.h
madmaxoft: The inner mod is needed, otherwise it doesn't get rid of the negative number. I think the code's correct and it's also the minimum needed.

madmaxoft: You might want to consider splitting the long line, each param on a separate line (plus separate line for the closing parenthesis, of course)

tigerw:
Ah, encountering the same problem as #1916. Is there no way no specialise std::hash with GCC/Clang?

worktycho:
Not in the current releases. Use a custom hash object instead.

MCServer/Plugins/MagicCarpet
```diff
@@ -1 +1 @@
-Subproject commit 94da343b62f0498a5843247f36d6ee00cbeb8f21
```
worktycho: Why

worktycho:
What are all the note blocks binary files? And if you delete them please remove them from the history compleatly.

src/Vector3.h
```diff
@@ -235,6 +235,12 @@ class Vector3
 		return *this;
 	}
 
+	/** Provides a hash of a vector's contents */
+	size_t operator()(const Vector3<T> & a_Vector) const
```
worktycho: Don't make this part of vector3. Either refactor VectorHasher out of aPath as aseperate object, or create one yourself at global scope.

madmaxoft:
@tigerw wants to archive his world and stats on GitHub :)

src/Vector3.h
tigerw: Why? Who knows when someone might want to hash a vector.

tigerw:
It's my legacy, xoft.

src/Vector3.h
worktycho: Because why should operator() return the has of a vector? It doesn't make sense. The std lib API is designed so that the hash object should be a seperate object from the data object, so that std::hash works.

worktycho: What you should do is create a seperate object, either at global scope, or a sub object of vector3 and use that.

src/Vector3.h
```diff
@@ -235,6 +235,12 @@ class Vector3
...
+	{
+		return ((std::hash<T>()(a_Vector.x) ^ (std::hash<T>()(a_Vector.y) << 1)) ^ std::hash<T>()(a_Vector.z));
```
worktycho: This has function is really bad for our dataset, as there will be lots of collisions within a chunk sized area. @SafwatHalaby had a much better hash function for our needs.

SafwatHalaby: It's still there, in Path.h

tigerw: How do you get a figure for the number of collisions?

worktycho: Reasoning about the entropy and trying to force a collision. Within a chunk, the four bits of entropy in each of the X and Z direction is concentrated in the lower order bits. You squash all that entropy into 5 bits.

tigerw: Can I improve this by shifting more?

worktycho: And then you end up with the hash function safwat used, just use the existing hash function.

SafwatHalaby: Perhaps we should migrate it from Path.h to Vector.h so that everybody uses it. It originally belonged there. (But clang complained about something, I think?)

worktycho: That's what I  think, but as a separate object, not part of cVector3. The clang error was from specializing std::hash which we wouldn't be doing.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -77,68 +77,60 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
...
-				itr.DataTwo = false;
-				itr.Data = Block;  // Update block information
-			}
-			return;
-		}
+		// Initialise the toggleable blocks list so that trapdoors etc. aren't reset on restart (#1887)
+		SimulatedPlayerToggleableBlocks.emplace(
+			a_RelBlockPosition,
+			AreCoordsDirectlyPowered(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_OriginalChunk) || AreCoordsLinkedPowered(a_RelBlockPosition.x, a_RelBlockPosition.y, a_RelBlockPosition.z, a_OriginalChunk)
+			);  // This map won't insert if key already present, so no need to check
```
worktycho: You should make AreCoordsDirectlyPowered and AreCoordsLinkedPowered take a vector.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -77,68 +77,60 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
...
+	{
+		SimulatedPlayerToggleableBlocks.erase(a_RelBlockPosition);
+	}
+
+	if ((Block != E_BLOCK_REDSTONE_REPEATER_ON) && (Block != E_BLOCK_REDSTONE_REPEATER_OFF))
+	{
+		static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_RepeatersDelayList.erase(a_RelBlockPosition);
 	}
 
+	auto & RedstoneSimulatorChunkData = static_cast<cIncrementalRedstoneSimulatorChunkData *>(a_OriginalChunk->GetRedstoneSimulatorData())->m_ChunkData;
```
worktycho: You should move the ChunkData get above the if, so you don't have to fetch it twice.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: In another PR I will aim to convert the simulator to use `Vector3i` :)

src/Simulator/IncrementalRedstoneSimulator.h
```diff
@@ -107,126 +98,171 @@ class cIncrementalRedstoneSimulator :
...
 	/** Handles tripwires */
 	void HandleTripwire(int a_RelBlockX, int a_RelBlockY, int a_RelBlockZ);
-	/* ===================== */
 
-	/* ====== Helper functions ====== */
+
 	/** Marks a block as powered */
-	void SetBlockPowered(int a_RelBlockX, int a_RelBlockY, int a_RelBlockZ, int a_RelSourceX, int a_RelSourceY, int a_RelSourceZ, unsigned char a_PowerLevel = MAX_POWER_LEVEL);
+	void SetBlockPowered(Vector3i a_RelBlockPosition, Vector3i a_RelSourcePosition, unsigned char a_PowerLevel = MAX_POWER_LEVEL);
+	void SetBlockPowered(int a_RelBlockX, int a_RelBlockY, int a_RelBlockZ, int a_RelSourceX, int a_RelSourceY, int a_RelSourceZ, unsigned char a_PowerLevel = MAX_POWER_LEVEL) { SetBlockPowered(Vector3i(a_RelBlockX, a_RelBlockY, a_RelBlockZ), Vector3i(a_RelSourceX, a_RelSourceY, a_RelSourceZ), a_PowerLevel); }
```
worktycho: Why do you still have this overload? How many clients does this function have?

tigerw:
About the style checker, I disagree that

```cpp
return
{
   1, 2, 3
}
```
should be favoured over
```cpp
return {
    1, 2, 3
}
```
since we use
```cpp
if (
   (1 == 1) &&
   (2 == 2)
)
{
}
```
as opposed to
```cpp
if
(
   ...
)
{
}
```

src/Simulator/IncrementalRedstoneSimulator.h
```diff
@@ -107,126 +98,171 @@ class cIncrementalRedstoneSimulator :
...
+
+	/** Recursively searches for a wire path and powers everything that should be powered */
+	void FindAndPowerBorderingWires(std::vector<std::pair<Vector3i, cChunk *>> & a_PotentialWireList, const Vector3i & a_EntryRelBlockPosition, cChunk * a_EntryChunk);
```
worktycho: What exactly do these pairs hold?

src/Simulator/IncrementalRedstoneSimulator.h
tigerw: The int overload? I will check.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Well you should still change it to use AreCoordsPowered.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -77,68 +77,60 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
...
-	cCoordWithBlockAndBoolVector & QueuedData = ((cIncrementalRedstoneSimulator::cIncrementalRedstoneSimulatorChunkData *)a_Chunk->GetRedstoneSimulatorData())->m_QueuedChunkData;
-	for (const auto & itr : QueuedData)
+	else
 	{
-		if ((itr.x == RelX) && (itr.y == a_BlockY) && (itr.z == RelZ))
+		if (Iterator != RedstoneSimulatorChunkData.end())
 		{
-			// Can't have duplicates in here either, in case something adds the block again before the structure can written to the main chunk data
+			Iterator->second.second = false;
+			Iterator->second.first = Block;  // Update block information
```
worktycho: I have no idea what this code is doing, which is why you should avoid std::pair here.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -185,24 +169,24 @@ void cIncrementalRedstoneSimulator::SimulateChunk(std::chrono::milliseconds a_Dt
...
 	{
-		if (dataitr->DataTwo)
+		if (dataitr->second.second)
 		{
 			dataitr = m_RedstoneSimulatorChunkData->m_ChunkData.erase(dataitr);
 			continue;
 		}
 
-		switch (dataitr->Data)
+		switch (dataitr->second.first)
```
worktycho: This huge switch should really be replaced with a lookup table, but I suspect that's out of scope for this PR.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Ah, but that doesn't take a cChunk.

worktycho: There is way too much complexity here then.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -77,68 +77,60 @@ void cIncrementalRedstoneSimulator::RedstoneAddBlock(int a_BlockX, int a_BlockY,
...
-				itr.DataTwo = true;  // The new blocktype is not redstone; it must be queued to be removed from this list
-			}
-			else
-			{
-				itr.DataTwo = false;
-				itr.Data = Block;  // Update block information
-			}
-			return;
-		}
+		// Initialise the toggleable blocks list so that trapdoors etc. aren't reset on restart (#1887)
```
worktycho: Anytime you write a comment: this block does x, it indicates the block should be a separate method.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Must you know? Not using a pair means another struct to wrap the two members for map storage. I have too many structs already :/

Would a comment where the map is defined satisfy you?

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: I am trying to keep the size of IncrementalSim.cpp down, especially with the requirement of 5 lines between functions, and not having much time to split the file.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: You have too many structs because you have too much logic in one class. std::pair only hides the root problem which is that cIncrementalRedstoneSimulator is a good object, whilst making the code even more confusing.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: That is a bad goal. If you don't have time to split the object, don't try to keep the size down, as that just leads to more confusing code. The only way to reduce the size of cIncrementalSim is to split the object.

tigerw: But I don't really see how in this case, adding another function will affect clarity. The block isn't long, nor is it used anywhere else. Moving the block into a method for the sake of it will at most replace this comment.

worktycho: Then either the code is unclear and should be rewritten for increased clarity, or the comment is unnecessary. I suspect that in this case its the first one due to the data-structures of the sim being too complicated.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: But comments de-confuse pairs?

cIncrementalRedstoneSim is a god object because it's a redstone simulator, no? Would you have separate classes for every redstone device it simulates, and 10+ additional files?

worktycho: No comments do not. And yes having 10 additional files is superior to the current situation, because when working out whats going on, you only have to look at a subset of the code.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Then how would you uncomplicate it?

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Why do comments not? `Iterator->second.second = false; // De-schedule entry removal`

Having more files arguably doesn't reduce complexity, good commenting does.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Split the object.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Its still completely unclear how the code is doing that, comments do not substitute clean code.

worktycho: No, but good commenting doesn't reduce complexity either. What multiple files do is allow you to ignore complexity that is not relevant to the problem you are working on.

tigerw: Okay, but I do not understand how it is unclear how the code is working, nor how adding files instead of comments will explain the iterator invalidation problems that necessitate using std::pair.

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Or for now another method?

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: Its the fact that the pair fields have no name, that Iterator->second.second = false; means nothing.

src/Simulator/IncrementalRedstoneSimulator.cpp
worktycho: I think we're either going to have to leave it for now, or draw a line and say we're not accepting any more feature PR's until the sim is split. I can't make that decision on my own @madmaxoft Whats your opinion?

tigerw: Unless a comment goes along with it?

src/Simulator/IncrementalRedstoneSimulator.cpp
tigerw: Unless a comment goes along with it?

worktycho: In every single location, which is much more work then just creating a struct with names.

tigerw: And then it becomes a matter of preference, especially since it's only used in two places.

src/Simulator/IncrementalRedstoneSimulator.h
tigerw: Around 10.

tigerw:
Merge?

src/Simulator/IncrementalRedstoneSimulator.h
worktycho: You should refactor them to use the vector version.

worktycho:
I'm not going to merge this because you haven't addressed all my comments.

src/Simulator/IncrementalRedstoneSimulator.h
tigerw: In another PR?

tigerw:
About std::pair:
I still don't fully agree with your recommendations. If a structure is used, the descriptiveness of the members would be that of `m_BlockType` and `m_Chunk`. In a pair, this would be given as BLOCKTYPE and cChunk by either reading the declaration or by the IDE. If one wanted to know what they did, one would still need to explore further.

About splitting large functions and the file in general:
I agree this is a good idea, but this pull request is already quite large and isn't relevant to it. I will be sure to do so sometime in the future.

src/Simulator/IncrementalRedstoneSimulator.h
worktycho: That wont happen. Seriously you need to finish any refactors in this PR, unless its a public method thats used in lots of places.

worktycho:
I'm just not going to merge this. But I won't object if someone else does.

tigerw:
:/

tigerw:
I view of the guideline not to add unrelated changes to requests, what would you like me to do? Neither std::pair nor splitting have anything to do with performance/bug fixes, which are the primary concern of this request.

Are you sure this can't go in? None of the issues mentioned are being caused by these changes, merely raised by them.

worktycho:
If making a refactor would improve the code you're changing, make it part of the PR. Otherwise the refactors never happen. The problem is adding multiple feature/bug fixes to a PR, which you keep doing.

worktycho:
Also, whilst the changes do not cause a lot of the problems, they exacerbate them, which brings those problems in scope.

tigerw:
Could I sway you if I promise to refactor after exams?

worktycho:
No

tigerw:
Well.

That's that.

No

bearbin:
Merging, as I don't think it's likely tiger is going to refactor for a few weeks, due to exams. A proper refactoring of the redstone code should hopefully follow.

src/Simulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -503,152 +486,278 @@ void cIncrementalRedstoneSimulator::HandleRedstoneButton(int a_RelBlockX, int a_
...
+				case E_BLOCK_REDSTONE_REPEATER_OFF:
+				{
+					BorderingMechanismCount++;
+					if (!RepeaterPresent)
+					{
+						// TODO: only if wire is actually connected to repeater (repeater facing right way)
+						RepeaterPresent = true;
+						OffsetToPower = { -Offset.x, Offset.y, -Offset.z };  // Negate to obtain offset in opposite direction since wire powers that way
+					}
+					SetBlockPowered(AdjustedOffset, RelBlock, MyPower);
```
madmaxoft: Shouldn't there be a `break` here?

tigerw: Looks like it, though (thankfully) it hasn't manifested as a bug yet.

PR #2148, opened 2015-05-27
AI - Saner Livestock

SafwatHalaby:
Fixes https://github.com/mc-server/MCServer/issues/2147 and makes chicken heads act less weirdly. Livestock should be much more Vanilla-like now.


SafwatHalaby:
signoff.

Issue #2106, opened 2015-05-23
Players become invisible when entering the Nether


jammet:
All the time, when I join the server, and at that time it happens that one or more other players are connected to the Nether they:
1. do not show up on the player list
2. they chat does isn't shown to people who're not in the nether
3. more often than not, when they return from the nether, you will not be able to see them, and vice versa.

beeduck:
Have you tried this with the End as well?

jammet:
No. I haven't found the end in my server world yet. Cannot say if the behavior is the the same.

PR #2150, opened 2015-05-28
Fix for Issue2106

beeduck:
Issue #2106
I've changed the way clients connecting to the server receive player lists. Before this change, the connecting client only requested the player list from the world they were initially connecting in. This behavior caused players in other worlds to not appear on the newly connected player's player list as well as appearing invisible when they finally see each other. I changed this by adding a function to cRoot that iterates through all current worlds sending their player list to the requesting client.

I've also made a change to broadcasting chat. Originally chat was only broadcast to the world of the sender, it is not broadcast server wide to all worlds.


madmaxoft:
You should rebase this PR onto current master, so that the CIs can have a go at it. Perhaps even squash the commits. Thanks.

PR #2157, opened 2015-05-28
Fix for chat and player lists with players in separate worlds

beeduck:
Issue #2106
I've changed the way clients connecting to the server receive player lists. Before this change, the connecting client only requested the player list from the world they were initially connecting in. This behavior caused players in other worlds to not appear on the newly connected player's player list as well as appearing invisible when they finally see each other. I changed this by adding a function to cRoot that iterates through all current worlds sending their player list to the requesting client.

I've also made a change to broadcasting chat. Originally chat was only broadcast to the world of the sender, it is not broadcast server wide to all worlds.

There is another pull request for this issue that I closed. I had to force push and clean up my fork.

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -558,6 +558,23 @@ void cRoot::SaveAllChunks(void)
 
 
 
+void cRoot::SendPlayerLists(cPlayer * a_DestPlayer)
+{
+	for (const auto & itr : m_WorldsByName)
+	{
+		itr.second->SendPlayerList(a_DestPlayer);
+	}  // for itr - m_WorldsByName[]
+}
+
+
+
+void cRoot::BroadcastPlayerListsAddPlayer(const cPlayer & a_Player, const cClientHandle * a_Exclude)
+{
+	for (const auto & itr : m_WorldsByName)
+	{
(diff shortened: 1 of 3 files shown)


src/Root.cpp
```diff
@@ -558,6 +558,14 @@ void cRoot::SaveAllChunks(void)
 
 
 
+void cRoot::SendPlayerLists(cPlayer * a_DestPlayer)
+{
+	for (WorldMap::iterator itr = m_WorldsByName.begin(), end = m_WorldsByName.end(); itr != end; ++itr)
```
tigerw: Could you do a `for (const auto & itr : m_WorldsByName)` here?

madmaxoft: Please don't use `itr` for things that are not iterators

tigerw: WorldMap::iterator isn't an iterator?

madmaxoft: A `WorldMap::iterator` is, but `world` in `for (const auto & world: m_WorldsByName)` isn't - it is a `std::pair<AString, cWorld *>`.
Ref.: http://stackoverflow.com/questions/6963894/c11-how-to-use-range-based-for-loop-with-stdmap
For this reason I think it's not worth the change.

Issue #1856, opened 2015-04-12
Nether: spawners, dungeons, portals and water


cafehaine:
With default generator:
Blaze spawners are pig spawners
Dungeons don't have any chests/rooms with nether wart
Portals are not created when we enter from overworld
We can place water in nether

NiLSPACE:
If you change the water simulator to `vaporize` then you can't place water. This should be part of the default configuration though.

NiLSPACE:
Dungeons do have nether wart rooms and chests. The chest do not have any content though.

madmaxoft:
Additional bug in portals: they don't link together. When you create a portal in the Nether and travel back to the overworld, you will not teleport to the original overworld portal
(reported in http://forum.mc-server.org/showthread.php?tid=1923 )

SphinxC0re:
Also the travel through a newly created portal should create another portal on the other side

NiLSPACE:
I've been wondering how we're going to detect nether portals. Maybe instead of scanning a whole chunk we could check in a checkerboard method, since a portal is at least 3 blocks wide (in MCServer, Vanilla has the minimal set to 4).
```
X: Checked
Z: Not Checked
X Z X Z X Z X Z X Z X Z X Z X
Z X Z X Z X Z X Z X Z X Z X Z
X Z X Z X Z X Z X Z X Z X Z X
Z X Z X Z X Z X Z X Z X Z X Z
X Z X Z X Z X Z X Z X Z X Z X
etc
```
That way you would always find the portals, but with half the amount of blocks checked. Or is there a different method to detect netherportals?

lkolbly:
I'm new to MCServer, just poking through it a little bit.

What if the chunks (or maybe the world?) stored a list of nether portals? It wouldn't be too terrible to add them in Blocks/BlockFire.h's FindAndSetPortalFrame, since it already knows where portals are. Then cEntitity::DoMoveToWorld could validate the portal and make sure the portal's still there, and if it isn't it could remove it from the list. (Vanilla Minecraft seems to do a similar thing, according to the minecraft wiki there's a 60-second cache of nether portal targets - so presumably they're caching these linkages somewhere)

Another thing that needs to be done is to multiply the entity's coordinates by 8 when switching between the overworld and the nether. Would this be best done as a function in the cWorld which translates coordinates from a world to its own coordinate space?

I have some downtime this week, I can take a look at this if nobody else already has plans to do so.

NiLSPACE:
[...]

PR #2162, opened 2015-05-29
Nether Portal Jumping & Creation Code

lkolbly:
Per #1856, this code implements a portal scanner which finds and creates nether portals when entities move between the overworld and the nether.

It also changes https://github.com/mc-server/MCServer/commit/dae9e579#diff-97bcf8a96a22aec81982926aa1462363R298 because StringToInteger returns true if the conversion was successful. I think.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -640,6 +640,7 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 		else if ((*itr)->IsWorldTravellingFrom(m_World))
 		{
 			// Remove all entities that are travelling to another world
+			LOGD("Removing entity from [%d, %d] that's travelling between worlds.", m_PosX, m_PosZ);
 			MarkDirty();
 			(*itr)->SetWorldTravellingFrom(nullptr);
 			itr = m_Entities.erase(itr);
@@ -695,7 +696,6 @@ void cChunk::MoveEntityToNewChunk(cEntity * a_Entity)
 	}
 
 	ASSERT(Neighbor != this);  // Moving into the same chunk? wtf?
-	
 	Neighbor->AddEntity(a_Entity);
 
 	class cMover :
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -13,6 +13,7 @@
 #include "Player.h"
 #include "Items/ItemHandler.h"
 #include "../FastRandom.h"
+#include "../NetherPortalScanner.h"
 
 
 
@@ -42,6 +43,7 @@ cEntity::cEntity(eEntityType a_EntityType, double a_X, double a_Y, double a_Z, d
 	m_WorldTravellingFrom(nullptr),
 	m_EntityType(a_EntityType),
 	m_World(nullptr),
+	m_IsWorldChangeScheduled(false),
 	m_IsFireproof(false),
 	m_TicksSinceLastBurnDamage(0),
 	m_TicksSinceLastLavaDamage(0),
@@ -1260,9 +1262,26 @@ void cEntity::DetectCacti(void)
 
 
 
+void cEntity::ScheduleMoveToWorld(cWorld * a_World, Vector3d a_NewPosition)
+{
+	m_NewWorld = a_World;
+	m_NewWorldPosition = a_NewPosition;
+	m_IsWorldChangeScheduled = true;
+}
+
+
+
 
 bool cEntity::DetectPortal()
 {
+	// If somebody scheduled a world change with ScheduleMoveToWorld, change worlds now.
+	if (m_IsWorldChangeScheduled)
+	{
+		m_IsWorldChangeScheduled = false;
+		MoveToWorld(m_NewWorld, false, m_NewWorldPosition);
+		return true;
+	}
+
 	if (GetWorld()->GetDimension() == dimOverworld)
 	{
 		if (GetWorld()->GetLinkedNetherWorldName().empty() && GetWorld()->GetLinkedEndWorldName().empty())
@@ -1312,8 +1331,15 @@ bool cEntity::DetectPortal()
 						// Send a respawn packet before world is loaded / generated so the client isn't left in limbo
 						((cPlayer *)this)->GetClientHandle()->SendRespawn(dimOverworld);
 					}
-					
-					return MoveToWorld(cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedOverworldName()), false);
+
+					Vector3d TargetPos = GetPosition();
+					TargetPos.x *= 8.0;
+					TargetPos.z *= 8.0;
+
+					cWorld * TargetWorld = cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedOverworldName(), dimNether, GetWorld()->GetName());
+					LOGD("Jumping nether -> overworld");
+					new cNetherPortalScanner(this, TargetWorld, TargetPos, 256);
+					return false;
 				}
 				else
 				{
@@ -1329,8 +1355,15 @@ bool cEntity::DetectPortal()
 						((cPlayer *)this)->AwardAchievement(achEnterPortal);
 						((cPlayer *)this)->GetClientHandle()->SendRespawn(dimNether);
 					}
-					
-					return MoveToWorld(cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedNetherWorldName(), dimNether, GetWorld()->GetName()), false);
+
+					Vector3d TargetPos = GetPosition();
+					TargetPos.x /= 8.0;
+					TargetPos.z /= 8.0;
+
+					cWorld * TargetWorld = cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedNetherWorldName(), dimNether, GetWorld()->GetName());
+					LOGD("Jumping overworld -> nether");
+					new cNetherPortalScanner(this, TargetWorld, TargetPos, 128);
+					return false;
 				}
 			}
 			case E_BLOCK_END_PORTAL:
@@ -1392,7 +1425,7 @@ bool cEntity::DetectPortal()
 
 
 
-bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn)
+bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d a_NewPosition)
 {
 	UNUSED(a_ShouldSendRespawn);
 	ASSERT(a_World != nullptr);
@@ -1413,6 +1446,8 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn)
 	// Remove all links to the old world
 	SetWorldTravellingFrom(GetWorld());  // cChunk::Tick() handles entity removal
 	GetWorld()->BroadcastDestroyEntity(*this);
+
+	SetPosition(a_NewPosition);
 
 	// Queue add to new world
 	a_World->AddEntity(this);
@@ -1438,7 +1473,7 @@ bool cEntity::MoveToWorld(const AString & a_WorldName, bool a_ShouldSendRespawn)
 		return false;
 	}
 
-	return DoMoveToWorld(World, a_ShouldSendRespawn);
+	return DoMoveToWorld(World, a_ShouldSendRespawn, GetPosition());
 }
 
 
--- a/src/Entities/Entity.h
+++ b/src/Entities/Entity.h
@@ -392,15 +392,20 @@ public:
 	/// Teleports to the coordinates specified
 	virtual void TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ);
 
+	/// Schedules a MoveToWorld call to occur on the next Tick of the entity
+	void ScheduleMoveToWorld(cWorld * a_World, Vector3d a_NewPosition);
+
+	bool MoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d a_NewPosition) { return DoMoveToWorld(a_World, a_ShouldSendRespawn, a_NewPosition); }
+
 	/** Moves entity to specified world, taking a world pointer */
-	bool MoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn = true) { return DoMoveToWorld(a_World, a_ShouldSendRespawn); }
+	bool MoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn = true) { return MoveToWorld(a_World, a_ShouldSendRespawn, GetPosition()); }
 
 	/** Moves entity to specified world, taking a world name */
 	bool MoveToWorld(const AString & a_WorldName, bool a_ShouldSendRespawn = true);
 	
 	// tolua_end
 
-	virtual bool DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn);
+	virtual bool DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d a_NewPosition);
 
 	/** Returns if the entity is travelling away from a specified world */
 	bool IsWorldTravellingFrom(cWorld * a_World) const { return (m_WorldTravellingFrom == a_World); }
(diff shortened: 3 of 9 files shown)


src/BlockID.cpp
```diff
@@ -295,7 +295,7 @@ eDimension StringToDimension(const AString & a_DimensionString)
 {
 	// First try decoding as a number
 	int res;
-	if (!StringToInteger(a_DimensionString, res))
```
worktycho: Thanks, I got a few of these wrong.

src/Chunk.cpp
```diff
@@ -640,6 +640,7 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 		else if ((*itr)->IsWorldTravellingFrom(m_World))
 		{
 			// Remove all entities that are travelling to another world
+			LOGD("Removing entity from %d,%d that's travelling between worlds.", m_PosX, m_PosZ);
```
madmaxoft: Usually we log chunk coords enclosed in `[]` and block coords enclosed in `{}`. Also, add a space after the comma.

src/Entities/Entity.cpp
```diff
@@ -1259,10 +1259,349 @@ void cEntity::DetectCacti(void)
...
+	}
+	else
+	{
+		m_PortalLoc.x += AcrossOffset;
+		m_PortalLoc.z += OutOffset;
+	}
+
+	m_Entity->ScheduleMoveToWorld(m_World, m_PortalLoc);
+	delete this;
+}
```
madmaxoft: This entire class should be moved into a separate file (Entity.cpp is already huge without it). Also, separate each function by 5 empty lines.

lkolbly: Alrighty. I noticed you guys have a vcproj file in src/ - I don't have Visual Studio, do I need to do anything to update that file or can I just create the file and add it to CMakeLists? (and the AllFiles.lst)

madmaxoft:
Please try to use the same coding style. We use a space after all commas, and a space around pointer and reference declarations (`cWorld * a_World` etc.) There's too many of these to report each one. Unfortunately the basic stylechecking script cannot detect these reliably, so you'll need to do it by hand.

lkolbly:
Okay, I'll go back through as I move it to a new file.

src/Entities/Entity.cpp
madmaxoft: There's no vcproj file, there's only a .vcproj.user file which is a template for per-user settings in VS. You only need to add the file to CMakeLists (it will be added to AllFiles.lst automatically)

src/Entities/Entity.cpp
```diff
@@ -1259,10 +1259,349 @@ void cEntity::DetectCacti(void)
...
+	return true;
+}
+
+
+
+bool cNetherPortalScanner::OnAllChunksAvailable(void)
+{
+	if (m_FoundPortal)
+	{
+		m_PortalLoc = m_PortalLoc;
```
NiLSPACE: I doubt this is needed.

lkolbly: Good catch, thanks.

src/BlockID.cpp
worktycho: I'm going to manually apply this patch to master, because its an important fix.

lkolbly:
Okay, I think I got it compliant with the coding convention in bad47a8, except maybe local variable names, I don't know what the convention is for those. I didn't see any commas without spaces, though, so I'm probably looking at the wrong commas?

src/Entities/Entity.cpp
```diff
@@ -1260,9 +1261,25 @@ void cEntity::DetectCacti(void)
...
+
 
 bool cEntity::DetectPortal()
 {
+	if (m_ReadyToChangeWorlds)
+	{
+		m_ReadyToChangeWorlds = false;
+		MoveToWorld(m_NewWorld, false, m_NewWorldPosition);
+		return true;
+	}
```
madmaxoft: This block deserves a comment above it explaining what it is doing.

src/Entities/Entity.cpp
```diff
@@ -1312,8 +1329,15 @@ bool cEntity::DetectPortal()
...
 						((cPlayer *)this)->GetClientHandle()->SendRespawn(dimOverworld);
 					}
-					
-					return MoveToWorld(cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedOverworldName()), false);
+
+					Vector3d TargetPos = GetPosition();
+					TargetPos.x *= 8.0;
+					TargetPos.z *= 8.0;
+
+					cWorld *TargetWorld = cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedOverworldName(), dimNether, GetWorld()->GetName());
```
madmaxoft: space after *

src/Entities/Entity.cpp
```diff
@@ -1329,8 +1353,15 @@ bool cEntity::DetectPortal()
...
 						((cPlayer *)this)->GetClientHandle()->SendRespawn(dimNether);
 					}
-					
-					return MoveToWorld(cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedNetherWorldName(), dimNether, GetWorld()->GetName()), false);
+
+					Vector3d TargetPos = GetPosition();
+					TargetPos.x /= 8.0;
+					TargetPos.z /= 8.0;
+
+					cWorld *TargetWorld = cRoot::Get()->CreateAndInitializeWorld(GetWorld()->GetLinkedNetherWorldName(), dimNether, GetWorld()->GetName());
```
madmaxoft: space after *

src/Entities/Entity.h
```diff
@@ -392,15 +392,19 @@ class cEntity
 	/// Teleports to the coordinates specified
 	virtual void TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ);
 
+	void ScheduleMoveToWorld(cWorld *a_World, Vector3d a_NewPosition);
```
madmaxoft: space after *

src/Entities/Entity.h
```diff
@@ -530,6 +534,10 @@ class cEntity
 	eEntityType m_EntityType;
 	
 	cWorld * m_World;
+
+	bool m_ReadyToChangeWorlds = false;
+	cWorld *m_NewWorld;
```
madmaxoft: space after *

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,288 @@
...
+	m_BuildPlatform = true;
+	m_Dir = 0;
+	m_Position = a_DestPosition;
+	m_PortalLoc = a_DestPosition;//TODO: Cast!
+	m_MaxY = a_MaxY;
+
+	int MinX = (int)floor((m_Position.x - SearchRadius) / cChunkDef::Width);
+	int MinZ = (int)floor((m_Position.z - SearchRadius) / cChunkDef::Width);
+	int MaxX = (int)ceil((m_Position.x + SearchRadius) / cChunkDef::Width);
+	int MaxZ = (int)ceil((m_Position.z + SearchRadius) / cChunkDef::Width);
```
madmaxoft: Use `FloorC` and `CeilC` instead - they do the static cast to `int` for you.

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,288 @@
...
+			m_Dir = 1;
+		}
+	}
+	else
+	{
+		// Scan the area for a suitable location
+		int minx = (int)floor(m_Position.x) - BuildSearchRadius;
+		int minz = (int)floor(m_Position.z) - BuildSearchRadius;
+		int maxx = (int)floor(m_Position.x) + BuildSearchRadius;
+		int maxz = (int)floor(m_Position.z) + BuildSearchRadius;
```
madmaxoft: `FloorC`

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+
+
+void cNetherPortalScanner::OnDisabled(void)
+{
+	// Now we actually move the player
+	if (!m_FoundPortal)
+	{
+		// Build a new nether portal.
+		LOGD("Building nether portal at {%d,%d,%d}", m_PortalLoc.x, m_PortalLoc.y, m_PortalLoc.z);
```
madmaxoft: Spaces after commas

src/NetherPortalScanner.h
```diff
@@ -0,0 +1,69 @@
...
+
+
+// This is the chunk stay which finds nearby nether portals
+class cNetherPortalScanner : public cChunkStay
+{
+public:
+	cNetherPortalScanner(cEntity * a_MovingEntity, cWorld * a_DestinationWorld, Vector3d a_DestPosition, int a_MaxY);
+	void OnChunkAvailable(int a_ChunkX, int a_ChunkY);
+	bool OnAllChunksAvailable(void);
+	void OnDisabled(void);
```
madmaxoft: These functions are `virtual` and should have an `override` specifier.

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,288 @@
...
+		}
+	}
+	return true;
+}
+
+
+
+
+
+void cNetherPortalScanner::BuildNetherPortal(Vector3i a_Location, int a_Direction, int a_IncludePlatform)
```
madmaxoft: `a_IncludePlatform` should be a `bool`

src/NetherPortalScanner.h
```diff
@@ -0,0 +1,69 @@
...
+	/** The world we're moving the entity to. */
+	cWorld * m_World;
+
+	/** Whether we found a portal during the loading of the chunks. */
+	bool m_FoundPortal;
+
+	/** Whether to build a platform. True if we couldn't build the portal on solid ground */
+	bool m_BuildPlatform;
+
+	/** The direction of the portal. 0=long axis along X, 1=long axis along Z */
```
madmaxoft: Spaces around `=`

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,288 @@
...
+		}
+
+		if (Possibilities.size() > 0)
+		{
+			m_BuildPlatform = false;
+
+			// Find the nearest
+			double DistanceToClosest = (Possibilities[0] - m_Position).SqrLength();
+			Vector3i Closest = Possibilities[0];
+			for (std::vector<Vector3i>::iterator itr = Possibilities.begin(); itr != Possibilities.end(); itr++)
```
tigerw: `for (const auto & itr : Possibilities)`?

lkolbly:
What's the preferred way to convert between Vector types, e.g. from Vector3i to a Vector3d?

src/NetherPortalScanner.cpp
madmaxoft: Not an `itr` :P

madmaxoft:
I don't think there's any preference. There's a simple copy-constructor provided within the vector template, and for converting to a `Vector3i` there's also the `Floor()` function defined on the class (`Ceil()` is missing right now, but could be added with minimal effort).

lkolbly:
Okay. I think I've fixed the convention errors.

src/Entities/Entity.cpp
madmaxoft: I think a simple `/* If a move to another world was scheduled via ScheduleMoveToWorld(), move the entity now. */` would have been enough. This is too wordy and still doesn't explain much. Also note that comments inside functions shouldn't use the doxy-comment format (`/**`), only comments above declarations should use it.

Also the `m_ReadyToChangeWorlds` is a somewhat misleading name, `m_IsWorldChangeScheduled` might have been better.

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,291 @@
...
+cNetherPortalScanner::cNetherPortalScanner(cEntity * a_MovingEntity, cWorld * a_DestinationWorld, Vector3d a_DestPosition, int a_MaxY)
+{
+	m_Entity = a_MovingEntity;
+	m_World = a_DestinationWorld;
+	m_FoundPortal = false;
+	m_BuildPlatform = true;
+	m_Dir = 0;
+	m_Position = a_DestPosition;
+	m_PortalLoc = a_DestPosition.Floor();
+	m_MaxY = a_MaxY;
```
madmaxoft: These should go in the initialization section of the constructor

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,291 @@
...
+	int z = a_Location.z;
+
+	// Clear a 3x4x4 area starting right above the base
+	for (int i = 0; i < SearchSolidBaseWidth; i++)
+	{
+		for (int j = 0; j < PortalLength; j++)
+		{
+			for (int k = 1; k < PortalHeight; k++)
+			{
+				if (a_Direction)
```
madmaxoft: Don't use an `int` like a `bool`. Ints should be compared to explicit values, bools should be used directly.
`if (intValue == 4)` vs `if (boolValue)`

lkolbly:
Alright, I just refactored m_Dir to use a Direction type which is either X or Y (I considered NORTH_SOUTH and EAST_WEST, but that didn't seem to map to the code as well).

madmaxoft:
Is there any chance you could get this rebased onto current master and squashed into a single commit?

lkolbly:
Sure thing. Done.

madmaxoft:
I think this is merge-worthy. Opinions?

src/Entities/Entity.h
```diff
@@ -530,6 +535,11 @@ class cEntity
 	eEntityType m_EntityType;
 	
 	cWorld * m_World;
+
+	/// State variables for ScheduleMoveToWorld.
+	bool m_IsWorldChangeScheduled = false;
```
NiLSPACE: Shouldn't this be initialized in the construcor? That is where we usualy initialize member variables right?

NiLSPACE:
Apart from the question I have about the `m_IsWorldChangeScheduled` initialization it looks really good.

src/Entities/Entity.h
tigerw: And also `/** */` for comment style.

Seadragon91:
Please add yourself to the contributors file.

src/NetherPortalScanner.h
```diff
@@ -0,0 +1,75 @@
...
+// This is the chunk stay which finds nearby nether portals
+class cNetherPortalScanner : public cChunkStay
+{
+public:
+	cNetherPortalScanner(cEntity * a_MovingEntity, cWorld * a_DestinationWorld, Vector3d a_DestPosition, int a_MaxY);
+	virtual void OnChunkAvailable(int a_ChunkX, int a_ChunkY) override;
+	virtual bool OnAllChunksAvailable(void) override;
+	virtual void OnDisabled(void) override;
+
+	enum Direction
```
tigerw: Maybe `enum class Direction`?

src/Entities/Entity.h
lkolbly: Yes, it should be initialized in the constructor. For the comment, there was a previous block of several comments there that were triple slashes, so I changed the whole block in the commit I just now pushed.

SphinxC0re:
Looks like it's ready for merge

src/Entities/Entity.h
madmaxoft: Yeah, sorry for the comment style, we couldn't decide on the format beforehands, so one half of the code uses `///` and the other half uses `/** */`. We've since then decided to use the latter, but didn't fix the former yet. Good call about replacing them all.

NiLSPACE:
After almost a month I finally updated my MCServer and tried this, and I noticed that when teleporting to an portal I get teleported 1 block next to it. That probably isn't intended right? You're supposed to teleport in another portal.

SphinxC0re:
Also, you get teleported instantly but I think this is another issue

NiLSPACE:
That's only in creative gamemode right? Because in vanilla you also get teleported instantly in creative mode.

src/NetherPortalScanner.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+
+
+void cNetherPortalScanner::OnChunkAvailable(int a_ChunkX, int a_ChunkZ)
+{
+	cChunkDef::BlockTypes blocks;
+	m_World->GetChunkBlockTypes(a_ChunkX, a_ChunkZ, blocks);
+
+	// Iterate through all of the blocks in the chunk
+	for (unsigned int i = 0; i < cChunkDef::NumBlocks; i++)
```
NiLSPACE: Looking at this again, shouldn't this be `i += 2` as suggested by @madmaxoft [here](https://github.com/cuberite/cuberite/issues/1856#issuecomment-105779965)? It would speed up the search for a nether portal by quite allot since it wouldn't have to check every block in the chunk.

Seadragon91:
I tested this, I can not get back to the world from the nether trough the portal. Used the FAQ for creating a nether world.

NiLSPACE:
That's probably because you haven't configured the world linking in the world.ini. The FAQ only describes how you can generate such a world.

SphinxC0re:
Yes, this is a problem. You *have* to stand right in the portal to teleport. in vanilla you only need partial contact with the portal.

tigerw:
> Yes, this is a problem. You have to stand right in the portal to teleport. in vanilla you only need partial contact with the portal.

This is to do with lack of bounding box detection in MCS at the present.

src/NetherPortalScanner.cpp
lkolbly: Because of the way the chunks are stored (XZY?), doing i+2 here would result in a pattern that skips every other row. Note that it's iterating from 0 to NumBlocks, not 0 to 16 each axis - this results in a speedup in not having to call CoordinateToIndex.

lkolbly:
@NiLSPACE 
> ... and I noticed that when teleporting to an portal I get teleported 1 block next to it.
That was intended, but can be changed easily.

NiLSPACE:
I think it should be changed, because I got teleported in the portal frame once. It **could** also be possible to be teleported inside a wall if the portal is next to it.

lkolbly:
I'm not sure you can since it places the portal in a 3x4x5 empty space, so there's always air on either side of the portal. But I can change it anyway.

NiLSPACE:
But if I build a portal against a wall then there isn't an empty space around it. ^^

lkolbly:
Oh, that's true. I hadn't thought about manually created portals. I will change that & the Y limit and make a PR after #2237 is closed. Having multiple forks is so much effort.

src/NetherPortalScanner.cpp
madmaxoft: It would have been possible to have two loops, the inner one iterating over the 256 blocks in a single Y slice, and the outer one iterating over Y slices, skipping 2 slices out of each 3:
```cpp
for (unsigned y = 0; y < cChunkDef::Height; y += 3)
{
  unsigned BaseY = cChunkDef::Width * cChunkDef::Width * y;
  for (unsigned i = 0; i < 256; i++)
  {
    // Check block at index [BaseY + i]
  }
}
```

PR #2163, opened 2015-05-29
Clean up Spawn Prepare

worktycho:
Made cSpawnPrepare execute on the same thread since it is a syncronous operation, and most of the code happens on the lighting thread.
Also moved cSpawnPrepare into its own file


madmaxoft:
Looks good.

PR #2164, opened 2015-05-30
Fix FreeBSD/clang errors caused by -Werror

linnemannr:
With FreeBSD/clang, -Werror combined with the configured warning flags yields
some fatal errors, specifically related to signed conversion, 64 to 32 bit
conversion, and tautological compares.

CONTRIBUTORS

	Add myself to the contributor list

src/Generating/FinishGen.cpp

	In cFinishGenPassiveMobs::GetRandomMob(), change the type of RandMob
	from size_t to the difference_type of the ListOfSpawnables iterator
	MobIter. Using size_t triggers a 64 bit to 32 bit conversion if the
	difference_type of the iterator class is 64 bit

	Also explicitly cast the noise expression to unsigned long so we don't
	get a signed conversion warning from the modulo against
	ListOfSpawnables.size()

src/OSSupport/StackTrace.cpp

	FreeBSD 10 and above includes a non glibc implementation of benchmark()
	for which size_t, not int, is the return type. To account for this and
	prevent a signed conversion warning, abstract the type for numItems with
	a macro btsize

src/StringUtils.h

	In StringToInteger(), correct a tautological compare warning for
	unsigned types with the template. If T is unsigned, comparing
	std::numeric_limits<T>::min() to the unsigned result is always
	false. That control can enter this branch in an evaluated template with
	an unsigned type T may also permit a signed number to be parsed and
	erroneously stripped of its signedness at runtime. To guard against this
	and avoid the warning in the case that the number parsed from the string
	is non-positive, return false and don't try to parse if T is unsigned
	and control enters the non-positive branch

### DIFF
--- a/src/Mobs/Guardian.cpp
+++ b/src/Mobs/Guardian.cpp
@@ -37,9 +37,10 @@ void cGuardian::GetDrops(cItems & a_Drops, cEntity * a_Killer)
 
 void cGuardian::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
+	m_IsFollowingPath = false;  // Disable Pathfinding until it's fixed. TODO
+
 	// We must first process current location, and only then tick, otherwise we risk processing a location in a chunk
 	// that is not where the entity currently resides (FS #411)
-	
 	Vector3d Pos = GetPosition();
 
 	// TODO: Not a real behavior, but cool :D
--- a/src/Mobs/Squid.cpp
+++ b/src/Mobs/Squid.cpp
@@ -35,9 +35,10 @@ void cSquid::GetDrops(cItems & a_Drops, cEntity * a_Killer)
 
 void cSquid::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
+	m_IsFollowingPath = false;  // Disable Pathfinding until it's fixed. TODO
+
 	// We must first process current location, and only then tick, otherwise we risk processing a location in a chunk
 	// that is not where the entity currently resides (FS #411)
-	
 	Vector3d Pos = GetPosition();
 
 	// TODO: Not a real behavior, but cool :D


src/Generating/FinishGen.cpp
```diff
@@ -1375,8 +1375,12 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
 		return mtInvalidType;
 	}
 
-	size_t RandMob = static_cast<size_t>((m_Noise.IntNoise2DInt(chunkX - chunkZ + 2, chunkX + 5) / 7)) % ListOfSpawnables.size();
 	auto MobIter = ListOfSpawnables.begin();
+	using diff_type =
+	    std::iterator_traits<decltype(MobIter)>::difference_type;
+	diff_type RandMob = static_cast<diff_type>
+	    ((unsigned long)(m_Noise.IntNoise2DInt(chunkX - chunkZ + 2, chunkX + 5) / 7)
+	    % ListOfSpawnables.size());
```
madmaxoft: Use tabs for indent

src/StringUtils.h
```diff
@@ -168,6 +168,10 @@ bool StringToInteger(const AString & a_str, T & a_Num)
 	}
 	else
 	{
+		// Unsigned result cannot be signed!
+		if (!std::numeric_limits<T>::is_signed)
+			return false;
```
madmaxoft: Add braces for the block, even if it is a single expression.

linnemannr:
Style notes addressed

madmaxoft:
You should rebase, instead of merge master

src/OSSupport/StackTrace.cpp
```diff
@@ -12,8 +12,12 @@
...
 #endif
 
-
-
+// FreeBSD uses size_t for the return type of backtrace()
+#if defined(__FreeBSD__) && __FreeBSD__ >= 10
+	#define btsize size_t
+#else
+	#define btsize int
+#endif
```
madmaxoft: Keep the 5-empty-lines spacing please

linnemannr:
that was a fatfinger mistake, that wasn't actually supposed to commit. I'm unexperienced with git. Is there any way to destroy that change?

madmaxoft:
Other than the style issues, this is good. I'd prefer if you squashed all the commits into a single one, though.

linnemannr:
Sure, can you direct me on how to do that?

madmaxoft:
I suppose with changes this small, it would be easiest to copy the modified files out, then checkout a fresh new copy of the upstream/master branch and copy the modified files back in. Commit and force-push over the current branch. This will make it a single commit atop the current master branch and this PR will get updated automatically.

madmaxoft:
(Once you merge, rebasing is extremely difficult. For future reference, usually this set of commands is used:
```
git remote add upstream https://github.com/mc-server/MCServer.git
git fetch --all
git rebase --interactive upstream/master
```

linnemannr:
Let me make sure I have this right - checkout my upstream/master remote (mc-server/MCServer) and make the changes there, then git push --force to my origin remote (linnemannr/MCServer)?

madmaxoft:
After copying the files out:
```
git checkout -- \*
git checkout master
git fetch --all
git merge upstream/master
```
This will reset your working folder to the current upstream/master. Now you can copy back the files, commit and force-push your working copy (as origin/master)

madmaxoft:
Wait, no, that won't work because you made the PR on your master branch.

linnemannr:
Haven't developed good github habits yet. Can I rebase origin/master as shown in https://help.github.com/articles/changing-a-commit-message/ to accomplish what we want?

madmaxoft:
This will get interesting
```
git checkout -b helper   (creates a helper branch so that master can be removed)
git branch -D master   (removes local master branch)
git checkout upstream/master    (resets the local folder to clean upstream master)
git checkout -b master   (recreates local master branch from the clean contents)
git branch -D helper   (removes the helper branch)
git branch --set-upstream-to=origin/master   (re-attaches the local master branch to origin/master)
git push --force   (force-pushes the clean contents to your repo)
```
Now copy back the modified files and commit and normal-push

PR #2167, opened 2015-05-30
PF - Fixed diagonal cutting

### LINKED ISSUE #2146: Pathfinder Diagonal cutting
@jammet noticed that 1-block high mobs get stuck sometimes. The underlying reason is that they try to move diagonally in places where they shouldn't. I thought I've fixed this in https://github.com/mc-server/MCServer/commit/7021547e99e1e169aff1ff0fada6f7e1e5349bcf but it seems I haven't.

SafwatHalaby:
Fixes #2146.


SafwatHalaby:
lgtm.

PR #2172, opened 2015-05-30
Made cLightingThread own its callbacks

worktycho:


madmaxoft:
We do have a `UniquePtr` that should be used instead of the lengthy and underscore-y `std::unique_ptr`

worktycho:
I personally thing the project typedefs are a barrier to newcomers. At least with the std types every c++ project has them. But with our typedefs, they need to be learned specially for our project.

worktycho:
Merge or are you going to insist on UniquePtr.

madmaxoft:
I'm too soft :P

PR #2173, opened 2015-05-30
Moved shared cmake flags to the CMAKE_C_FLAGS var

### LINKED ISSUE #2168: x86 Cross-compilation does not work.
Here's an error log: http://bernard.bearbin.net/job/MCServer%20Linux%20x86%20Master/14/

This problem also occurs on the old buildserver, but the old one was misconfigured so the problem did not appear.

worktycho:
Fixes #2168

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -16,14 +16,8 @@ macro (add_flags_lnk FLAGS)
 endmacro()
 
 macro(add_flags_cxx FLAGS)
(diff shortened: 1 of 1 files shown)


bearbin:
Huh? GitHub showed all builds as successful when I merged, and now it shows appveyor as failed.

madmaxoft:
I quite remember this exact trick not working for some reason - of course I had tried this before.

PR #2175, opened 2015-05-30
PF - Iron bars support

SafwatHalaby:


bearbin:
Iron bars aren't fences though - they're only 1 block high.

SafwatHalaby:
Don't they block mobs though?

bearbin:
And if iron bars are included, glass panes should be also. 

And yes, they do block mobs, i didn't realise what function this was in.

SafwatHalaby:
Thanks. I'll add glass panes too.

I'm planning to rewrite isSolid, by the way. It is not the right place for it and it can cause some issues in some fence setups.

SafwatHalaby:
Wrong, wrong.
I just tested this on Vanilla, neither glass nor iron acts like a fence. Mobs can freely jump on them.

PR #2178, opened 2015-05-31
Fix Daylightsensor recipe

jammet:
Something like Woodslap^-1 didn't work here, so I split it up into individual recipes.


madmaxoft:
That's weird, it should have worked.

madmaxoft:
(Also, what's with the "slap"s? :)

madmaxoft:
Ah, that would be because there are two wooden slabs in MC. The old one (`woodslab`) shared the block type with the stone and brick slabs, the new one (`woodenslab`) had different wood types. If you change the recipe to use the new one, it should work with `^1`.

jammet:
I don't know, sometimes my grammar is just way off, and turns what would otherwise be completely boring into something stupid! XD

I'll test the woodenslaps thing on my server and if it works, I'll change it around here, too.

jammet:
Nice! :)

PR #2179, opened 2015-05-31
Chunk queue collapsing

worktycho:
This change should provide major performance increases to larger servers.

### DIFF
--- a/src/ChunkMap.h
+++ b/src/ChunkMap.h
@@ -302,7 +302,7 @@ public:
 	Returns false if there's no command block at those coords or callback returns true, returns true if found. */
 	bool DoWithCommandBlockAt(int a_BlockX, int a_BlockY, int a_BlockZ, cCommandBlockCallback & a_Callback);  // Lua-accessible
 
-	/** Calls the callback for the mob head block at the specified coords.chu
+	/** Calls the callback for the mob head block at the specified coords.
 	Returns false if there's no mob head block at those coords or callback returns true, returns true if found. */
 	bool DoWithMobHeadAt(int a_BlockX, int a_BlockY, int a_BlockZ, cMobHeadCallback & a_Callback);  // Lua-accessible
 
--- a/src/ChunkMap.h
+++ b/src/ChunkMap.h
@@ -73,7 +73,6 @@ public:
 	void BroadcastBlockAction(int a_BlockX, int a_BlockY, int a_BlockZ, char a_Byte1, char a_Byte2, BLOCKTYPE a_BlockType, const cClientHandle * a_Exclude = nullptr);
 	void BroadcastBlockBreakAnimation(UInt32 a_EntityID, int a_BlockX, int a_BlockY, int a_BlockZ, char a_Stage, const cClientHandle * a_Exclude = nullptr);
 	void BroadcastBlockEntity(int a_BlockX, int a_BlockY, int a_BlockZ, const cClientHandle * a_Exclude);
-	void BroadcastChunkData(int a_ChunkX, int a_ChunkZ, cChunkDataSerializer & a_Serializer, const cClientHandle * a_Exclude = nullptr);
 	void BroadcastCollectEntity(const cEntity & a_Entity, const cPlayer & a_Player, const cClientHandle * a_Exclude = nullptr);
 	void BroadcastCollectPickup(const cPickup & a_Pickup, const cPlayer & a_Player, const cClientHandle * a_Exclude = nullptr);
 	void BroadcastDestroyEntity(const cEntity & a_Entity, const cClientHandle * a_Exclude = nullptr);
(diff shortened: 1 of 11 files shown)


src/ChunkSender.cpp
```diff
@@ -13,7 +13,7 @@
 #include "BlockEntities/BlockEntity.h"
 #include "Protocol/ChunkDataSerializer.h"
 #include "ClientHandle.h"
-
+#include "Chunk.h"
```
NiLSPACE: This needs an empty line afterwards.

worktycho: Why?

src/ChunkSender.cpp
```diff
@@ -247,64 +217,62 @@ void cChunkSender::Execute(void)
...
-		return;
+		if(!(*itr)->WantsSendChunk(a_ChunkX, a_ChunkZ))
+		{
+			auto toremove = itr;
+			itr++;
+			a_Clients.erase(toremove);
+		}
+		else
+		{
+			itr++;
```
tigerw: Apparently, you can erase normally when iterating a set or map as all implementations ensure order is unchanged anyway (as does C++14).

src/ChunkSender.cpp
```diff
@@ -247,64 +217,62 @@ void cChunkSender::Execute(void)
...
-	// Send block-entity packets:
-	for (sBlockCoords::iterator itr = m_BlockEntities.begin(); itr != m_BlockEntities.end(); ++itr)
-	{
-		if (a_Client == nullptr)
-		{
-			m_World->BroadcastBlockEntity(itr->m_BlockX, itr->m_BlockY, itr->m_BlockZ);
-		}
-		else
+		// Send block-entity packets:
+		for (auto Pos : m_BlockEntities)
```
tigerw: People recommend to use constant references to avoid copying. Does this apply?

src/ClientHandle.cpp
```diff
@@ -2907,9 +2904,9 @@ void cClientHandle::AddWantedChunk(int a_ChunkX, int a_ChunkZ)
 	
 	LOGD("Adding chunk [%d, %d] to wanted chunks for client %p", a_ChunkX, a_ChunkZ, this);
 	cCSLock Lock(m_CSChunkLists);
-	if (std::find(m_ChunksToSend.begin(), m_ChunksToSend.end(), cChunkCoords(a_ChunkX, a_ChunkZ)) == m_ChunksToSend.end())
+	if (m_ChunksToSend.find(cChunkCoords(a_ChunkX, a_ChunkZ)) == m_ChunksToSend.end())
```
tigerw: For all the finds, you could probably use the count() member for more concision.

src/ChunkMap.h
```diff
@@ -303,7 +302,7 @@ class cChunkMap
 	Returns false if there's no command block at those coords or callback returns true, returns true if found. */
 	bool DoWithCommandBlockAt(int a_BlockX, int a_BlockY, int a_BlockZ, cCommandBlockCallback & a_Callback);  // Lua-accessible
 
-	/** Calls the callback for the mob head block at the specified coords.
+	/** Calls the callback for the mob head block at the specified coords.chu
```
madmaxoft: leftovers?

src/ChunkSender.cpp
NiLSPACE: Because we always have 5 empty lines right?

tigerw:
How much of a performance improvement does it provide? Could you explain how it does it?

src/ClientHandle.cpp
madmaxoft: Why would you want to use `count()` when all you care about is whether it is found or not?

tigerw: I understand that since normal maps have unique keys, count returns 0 or 1, which may look better in comparison to comparing to containername.end().

However, some advise that find can be faster, even when count is aliased to find in sets and maps, because of less function calls.

src/ChunkSender.cpp
worktycho: Its ~16 bytes, less than one cache line. Its not going to make much difference.

worktycho:
For my benchmark, which is a extreme scenario, orders of magnitude improvements. It does it by changing the data structures, so that most of lookup's can be done in O(1) time.

madmaxoft:
I like where this change is headed. Fix up the details so it can get merged soon.

worktycho:
Fixed most of the issues

madmaxoft:
Rebase onto master? I suppose the first commit has already been merged in the other PR

worktycho:
I forgot about the first commit, I needed to squash a fix into the penultimate commit.

worktycho:
Rebased.

bearbin:
This pull request broke chunk sending. Chunks are now randomly not sent and do not display on the client.

worktycho:
Is there a pattern to it? Because off the top of my head I can't think of anything.

worktycho:
Is it inital sending or broadcasts that are the problem?

bearbin:
I think initial sending. 

There's no pattern to it really, it's just random. After a server restart the missing chunks change.

worktycho:
I can't see anything, and I can't really debug this this weekend. Could you have a look at it? or should I revert the merge.

bearbin:
Revert, and then investigate.

PR #2180, opened 2015-05-31
Fixed crafting recipe matching.

### LINKED ISSUE #2096: After crafting a hoe, the ingredients for a shovel are left in the crafting window.

madmaxoft:
Fixes #2096.

This was a rather big oversight, wonder how it could go unnoticed for so long.


bearbin:
Not enough testing in survival I guess. And it only manifests on a few crafting recipies.

Issue #2181, opened 2015-06-01
Service support broken on Windows


birkett:
With the addition of the TCLAP library to handle command line parameters, Windows service support is broken.

Starting the executable with "/service" on the command line, results in an unhanded exception while parsing the command line.

I'm working on a fix, and will issue a PR later today.

bearbin:
What happens if you use linux-style switches?

birkett:
TCLAP doesn't seem to like anything other than a single character in the switch names. Swapping the service to use -d instead of /service.

Also, lets ignore unmatched parameters with CmdLine::ignoreUnmatched. That will stop the exceptions being thrown with unknown parameters, and just ignore them instead.

PR #2182, opened 2015-06-01
Use the new style command line parameters for Windows Service support.

birkett:
Fixes service support on Windows by swapping to the "-d" (daemon) switch instead of "/service". Plays nicely with TCLAP.

Also, ignore unmatched parameters, preventing unhanded exceptions.

See issue #2181

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -382,6 +382,10 @@ std::unique_ptr<cMemorySettingsRepository> parseArguments(int argc, char **argv)
 
 		TCLAP::SwitchArg noBufArg("", "no-output-buffering", "Disable output buffering", cmd);
 
+		TCLAP::SwitchArg runAsServiceArg("d", "run-as-service", "Run as a service on Windows", cmd);
+
+		cmd.ignoreUnmatched(true);
+
 		cmd.parse(argc, argv);
 
 		auto repo = cpp14::make_unique<cMemorySettingsRepository>();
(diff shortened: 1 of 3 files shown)


madmaxoft:
Looks good. Not sure about the unmatched parameters, we do want some feedback on unknown parameters.

birkett:
Agreed.
Right now, TCLAP is throwing an exception, then calling exit(). I have no idea why a library would ever want to do that.

It does provide a method CmdLine::setExceptionHandling(bool), to disable the internal exception handling, and just re-throw. I'll put together some code to handle the exceptions outside the library.


Would you want invalid arguments to kill the execution, or print a warning, and continue with default values?

src/main.cpp
```diff
@@ -382,6 +382,10 @@ std::unique_ptr<cMemorySettingsRepository> parseArguments(int argc, char **argv)
 
 		TCLAP::SwitchArg noBufArg("", "no-output-buffering", "Disable output buffering", cmd);
 
+		TCLAP::SwitchArg runAsServiceArg("d", "run-as-service", "Run as a service on Windows", cmd);
```
worktycho: If you make the first parameter an empty string and then the second parameter service, you should be able to get the old behaviour of only allowing long form.

worktycho:
I think the idea of TCLAPs error handling is that for a basic app,TCLAP has built in feedback that is sufficient, so doing that means you don't have to worry about invalid args. If you want anything more complex you have to write your own, so the library simplifies the simple case.

Issue #1608, opened 2014-11-22
Anvil and enchantments not working


tggtggt:
Anvils are working to combine to combine things.
Ex: sword + enchanted book: looting = sword
Same thing goes for enchantments

parlock:
Any word on when these will be implemented in game, missing anvils and enchantments limits what one can do in game.  Can't even enchant armor or weapons.

Wasn't able to get these either to work at all in 1.8.* of the client.

PR #2183, opened 2015-06-01
Fixed Anvil Mechanics with Enchantments #1608

Rakete1111:
This fixes anvil behavior when fusing/enchanting items and books.
#1608

### DIFF
--- a/src/UI/SlotArea.cpp
+++ b/src/UI/SlotArea.cpp
@@ -1140,30 +1140,33 @@ void cSlotAreaAnvil::UpdateResult(cPlayer & a_Player)
 			int FirstLevel = Input.m_Enchantments.GetLevel(CurrentEnchantment);
 
 			// The enchantment could be illegal (Example: Sharpness II on an iron helmet) - Should happen only with Books
-			if ((((Input.m_ItemType == E_ITEM_LEATHER_CAP) || (Input.m_ItemType == E_ITEM_CHAIN_HELMET) || (Input.m_ItemType == E_ITEM_IRON_HELMET) || (Input.m_ItemType == E_ITEM_DIAMOND_HELMET) || (Input.m_ItemType == E_ITEM_GOLD_HELMET))
+			if ((ItemCategory::IsHelmet(Input.m_ItemType)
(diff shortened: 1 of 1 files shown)


Rakete1111:
Why did Travis (clang) fail?? It doesn't seen to be coding conventions :)

bearbin:
https://travis-ci.org/mc-server/MCServer/jobs/64961027#L1441

src/UI/SlotArea.cpp
```diff
@@ -1126,8 +1127,137 @@ void cSlotAreaAnvil::UpdateResult(cPlayer & a_Player)
...
+				|| ((Input.m_ItemType == 398)
+				&& (CurrentEnchantment != 34))
+				/* Carrot on a Stick */
+				|| ((Input.m_ItemType == 359)
+				&& ((CurrentEnchantment != 34) && (CurrentEnchantment != 32) && (CurrentEnchantment != 33)))
+				/* Shears */
+				|| ((Input.m_ItemType == 259)
+				&& (CurrentEnchantment != 34))
+				/* Flint and Steel */
+				)
```
madmaxoft: This is just TERRIBLE!
Use symbolic names (`E_ITEM_ABC`) instead of the numbers. Format the conditions properly, use indent to indicate the depth of the comparisons.

Rakete1111: I didn't know there were symbolic names....
I'll make another commit ;D

NiLSPACE: You can even use the ItemCategory functions.

Rakete1111: The clang compiler gives an error, as I try to compare unsigned and signed variables. But if it explicitly convert them to unsigned int, then the VS compiler gives me a warning, what should I do??

bearbin: You should use the symbolic names.

Rakete1111: This has nothing to do with symbolic names, is it possible that Clang sees const int variables as unsigned? VS doesn't say so.
I already added the symbolic names, but the build will fail Travis again if this isn't fixed :(
This is really weird

madmaxoft:
It's still terrible, the condition is completely unreadable.

Rakete1111:
It's readable if the code editor is a bit larger than the one github uses.

madmaxoft:
No, I mean the condition is unreadable because different depth levels use the same indent level, so it's become a mash of parenthesis and comparisons.

I think most of the enchantments is applicable to groups of similar items (helmets, boots, shovels etc) so it would make sense to abstract the itemtype comparisons into categories. There are `ItemCategory::IsShovel()` and similar functions that should be used for that.

parlock:
Any idea of this will get in the main code?  Would be nice to be able to use anvils to enchant things.

worktycho:
the easiest way is to fork the project on github and push your changes to a branch there. you should then be able to open a pull request with the changes against the main cuberite master, and we can review the changes their.

SphinxC0re:
Any updates on this one?

SphinxC0re:
Closing due to inactivity

PR #2188, opened 2015-06-01
Skip unknown cflag for Apple clang & remove cSemaphore

cengizIO:
Greetings,
Two issues were addressed in this PR.

I) I have no idea why the https://github.com/mc-server/MCServer/blob/master/SetFlags.cmake#L266 was needed (Perhaps not anymore?). 

It's throwing an error with my clang
````
$ clang -v
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.3.0
Thread model: posix
````
````
[ 29%] Building CXX object src/Blocks/CMakeFiles/Blocks.dir/BlockBed.cpp.o
error: unknown warning option '-Wno-reserved-id-macro' [-Werror,-Wunknown-warning-option]
````
II) `sem_init` and `sem_destroy` was throwing compiler errors because they are marked as `__deprecated` in OSX `/usr/include/sys/semaphore.h`. Here's a SO question about them: http://stackoverflow.com/questions/1413785/sem-init-on-os-x. 

Thus, I had to force named semaphores to be used for best Unix portability. It compiles fine, but I didn't test it yet.

### CODE REVIEW

**Comment by worktycho** [comment]
The compiler error is problematic. Because apple do not use official clang releases, it appears that the flag was introduced at some point between apples fork, and clang 3.6 being officially released. So we need add `-Wno-reserved-id-macro` for official clang 3.6 and later.

File: src/OSSupport/Semaphore.cpp [hunk 1]
```diff
@@ -6,32 +6,22 @@
...
-#ifndef _WIN32
-	: m_bNamed( false)
-#endif
 {
 #ifndef _WIN32
 	(void)a_MaxCount;
 	m_Handle = new sem_t;
-	if (sem_init( (sem_t*)m_Handle, 0, 0))
+	AString Name;
+	Printf(Name, "cSemaphore%p", this);
```
worktycho: This is broken. Semaphore names are system global, so this could cause issues running multiple copies of MCServer at once.

worktycho:
I suspect that the semaphore code is going to have to special case OS X. It looks like apple don't want to duplicate the Cocoa API's for unnamed semaphores, and the IPC use case complicates the POSIX semaphores. So on OS X the preferred way of handling this use case is to use the Cocoa GCD API.

madmaxoft:
It might be better to get rid of the semaphores completely. It should be possible with the C++11 synchronization primitives.

worktycho:
There isn't one in c++11 thread lib. So we'd need to build one out of a condition var and an atomic counter.

cengizIO:
### Unknown Compiler Flag

@worktycho yes I have confirmed that clang which is shipped with the latest XCode 6.3.2 does not contain `reserved-id-macro`.

screenshot 2015-06-03 11 05 33

Side | URL
------------ | -------------
Left | http://www.opensource.apple.com/source/clang/clang-600.0.54/src/tools/clang/include/clang/Basic/DiagnosticGroups.td
Right | http://llvm.org/svn/llvm-project/cfe/tags/RELEASE_360/final/include/clang/Basic/DiagnosticGroups.td

I don't know how can we let `cmake` know which `clang` flags are supported. Version numbers are same. 

Is it acceptable if I add cmake `APPLE` variable to the condition in SetFlags?

cengizIO:
### Deprecated Semaphore Functions

Would it solve semaphore name uniqueness problem if we prepended MCServer port to the semaphore name? Does it sound hacky?

Otherwise, I'll research the GCD API and add a separate condition to cSemaphore

madmaxoft:
I think appending a PID to the semaphore name would be a good enough fix. Of course, getting rid of them would be the best :)

cengizIO:
I updated `cSemaphore` as @madmaxoft suggested and added an availability check on clang compiler flag. 

Also, `cSemaphore` is never used in `master` branch. I updated it anyhow.

tigerw:
cSemaphore is never used?

Couldn't it be removed then?

cengizIO:
@tigerw I've went back in history and found duplicated code in `cEvent`.

`cSemaphore` is never used and `cEvent` currently uses C++11 primitives.

- Removed `cSemaphore`
- Updated documentation

madmaxoft:
Looks good, merge if it compiles.

PR #2193, opened 2015-06-02
Added HOOK_KILLED

### LINKED ISSUE #1197: Make death messages overridable by a plugin
It should be possible for a plugin to provide custom death messages, or to disable death messages altogether.

nounoursheureux:
Fix #1197

### DIFF
--- a/MCServer/Plugins/APIDump/APIDesc.lua
+++ b/MCServer/Plugins/APIDump/APIDesc.lua
@@ -2683,7 +2683,7 @@ Parser:close();
 				},
 			},  -- AdditionalInfo
 		},  -- lxp
-		
+
 		sqlite3 =
 		{
 			Desc = [[
--- a/MCServer/Plugins/APIDump/APIDesc.lua
+++ b/MCServer/Plugins/APIDump/APIDesc.lua
@@ -2683,7 +2683,7 @@ Parser:close();
 				},
 			},  -- AdditionalInfo
 		},  -- lxp
-
+		
 		sqlite3 =
 		{
 			Desc = [[
--- a/MCServer/Plugins/APIDump/Hooks/OnKilled.lua
+++ b/MCServer/Plugins/APIDump/Hooks/OnKilled.lua
@@ -1,0 +1,22 @@
+return 
+{
+	HOOK_KILLED = 
+	{
+		CalledWhen = "A player or a mob died.",
+		DefaultFnName = "OnKilled",
+		Desc = [[
+			This hook is called whenever player or a mob dies. It can be used to change the death message.
+		]]
+		Params = 
+		{
+			{ Name = "Victim", Type = "{{cEntity}}", Notes = "The player or mob that died" },
+			{ Name = "TDI", Type = "{{TakeDamageInfo}}", Notes = "Informations about the death" },
+			{ Name = "DeathMessage", Type = "string", Notes = "The default death message. An empty string if the victim is not a player" },
+		},
+		Returns = [[
+			The function may return two values. The first value is a boolean specifying whether other plugins should be called. If it is true, the other plugins won't get notified of the death. If it is false, the other plugins will get notified.</p>
+			<p>The second value is a string containing the death message. If the victim is a player, this death message is broadcasted instead of the default death message. If it is empty, no death message is broadcasted. If it is nil, the message is left unchanged. If the victim is not a player, the death message is never broadcasted.</p>
+			<p>In either case, the victim is dead.
+		]],
+	},  -- HOOK_KILLED
+}
--- a/MCServer/Plugins/APIDump/Hooks/OnKilling.lua
+++ b/MCServer/Plugins/APIDump/Hooks/OnKilling.lua
@@ -7,11 +7,7 @@ return
 		Desc = [[
 			This hook is called whenever a {{cPawn|pawn}}'s (a player's or a mob's) health reaches zero. This
 			means that the pawn is about to be killed, unless a plugin "revives" them by setting their health
-			back to a positive value.</p>
-			<p>
-			FIXME: There is no HOOK_KILLED notification hook yet; this is deliberate because HOOK_KILLED has
-			been recently renamed to HOOK_KILLING, and plugins need to be updated. Once updated, the HOOK_KILLED
-			notification will be implemented.
+			back to a positive value.
 		]],
 		Params =
 		{
--- a/src/Bindings/Plugin.h
+++ b/src/Bindings/Plugin.h
@@ -64,6 +64,7 @@ public:
 	virtual bool OnHandshake                (cClientHandle & a_Client, const AString & a_Username) = 0;
 	virtual bool OnHopperPullingItem        (cWorld & a_World, cHopperEntity & a_Hopper, int a_DstSlotNum, cBlockEntityWithItems & a_SrcEntity, int a_SrcSlotNum) = 0;
 	virtual bool OnHopperPushingItem        (cWorld & a_World, cHopperEntity & a_Hopper, int a_SrcSlotNum, cBlockEntityWithItems & a_DstEntity, int a_DstSlotNum) = 0;
+	virtual bool OnKilled                   (cEntity & a_Victim, TakeDamageInfo & a_TDI, AString & a_DeathMessage) = 0;
 	virtual bool OnKilling                  (cEntity & a_Victim, cEntity * a_Killer, TakeDamageInfo & a_TDI) = 0;
 	virtual bool OnLogin                    (cClientHandle & a_Client, int a_ProtocolVersion, const AString & a_Username) = 0;
 	virtual bool OnPlayerAnimation          (cPlayer & a_Player, int a_Animation) = 0;
--- a/src/Bindings/PluginLua.cpp
+++ b/src/Bindings/PluginLua.cpp
@@ -750,6 +750,30 @@ bool cPluginLua::OnHopperPushingItem(cWorld & a_World, cHopperEntity & a_Hopper,
 
 
 
+bool cPluginLua::OnKilled(cEntity & a_Victim, TakeDamageInfo & a_TDI, AString & a_DeathMessage)
+{
+	cCSLock Lock(m_CriticalSection);
+	if (!m_LuaState.IsValid())
+	{
+		return false;
+	}
+	bool res = false;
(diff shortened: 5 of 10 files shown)


nounoursheureux:
Yay, it's finally working !

madmaxoft:
I think it would have been better to make without the extra structure. Just have the hook return an extra value for the string to be used as the message (similar to how `OnChat` hook works)

nounoursheureux:
I didn't know it was possible, but it is a great idea !

nounoursheureux:
Thanks, this is much cleaner now.

madmaxoft:
I think you didn't commit the entire thing, the PDI is still used in the code.

nounoursheureux:
I forgot to add the files ><

madmaxoft:
Looks good. But before we merge, there are two things to do: Squash the commits into a single one, and then add another commit with the APIDump documentation for the hook (in the $/MCServer/Plugins/APIDump/Hooks folder)

Howaner:
I don't like this hook.
Why not create a PLAYER_DEATH hook with the death message as a return value?

nounoursheureux:
Well, it is basically the same. Why don't you like it ?

Howaner:
I don't like the name.
Maybe rename the hook to PLAYER_DEATH_HOOK ?

nounoursheureux:
I think both names are OK 
@madmaxoft which one do you prefer ?

jan64:
Maybe HOOK_KILLED? That way, it'll fit in with the KILLING one.

nounoursheureux:
Not exactly, because this hook is only called when a player dies and HOOK_KILLING is called for all entites.

jan64:
There is no hook specific to players only when they are about to die, so i don't think this should work like that.

nounoursheureux:
This hook sends the default death message as a parameter, and non-player entites don't have death messages so it will not work. I think it is better to have a hook specific to players deaths

jan64:
http://mc-server.xoft.cz/LuaAPI/OnKilling.html , we need a hook that's useful as a notifications when something dies. And i think it could be used to pass the death message. That way, it'll be more unified.

jan64:
Also, i don't think it's necessary to pass the server-provided death message at all. 
If plugin gets the same data as in the KILLING hook, and have a way to block the server death message (return true, or smthing).

jan64:
Sorry for saying that your idea is wrong, but if we are gonna do this we should do it the right way.

nounoursheureux:
I understand that it is more consistent, but what should happen when a non-player entity dies and a plugin returns a death message ? Ignore the message ? 
Debating is precisely the point of pull requests :P

jan64:
I don't think there should be a return message, just a way to cancel the built-in one (this would get ignored for non-players). 
That way, hook callback could check if the entity killed is a player and if so, do something (be it broadcasting a message or banning the player).

nounoursheureux:
OK, I understand. But there is a default death message, so the callback should return something to indicate it doesn't want the default message to be broadcasted.

jan64:
Yes, it could be done the same way as in HOOK_PLAYER_JOINED.

jan64:
As a side note, how regularly are the API docs updated?

madmaxoft:
Frankly, I think the `HOOK_PLAYER_JOINED`-way is the wrong one. What if the plugin doesn't want the join message to be broadcasted, but also doesn't want to block other plugins from receiving the hook callback? There's no way of doing that now. I like this PR's way much more.
I'd be okay if the hook was renamed to HOOK_KILLED and the returned death message would be ignored for non-player entities. That sounds completely reasonable API-wise.
Side note: The API docs are generated and uploaded daily, unless my build scripts failed.

madmaxoft:
(Of course if a plugin returned an empty death message, the server should not send such a message)

nounoursheureux:
If an empty string is returned, the server should send the default message or nothing at all ?

nounoursheureux:
I don't think HOOK_KILLED could return a death message because this would mean that it is passed as an argument, and non-player entities don't have death messages.

jan64:
they can have empty ones, another solution is we return two bools - one as an indications of blocking death message, and another one as a way to prevent other plugins from receiving this event.

madmaxoft:
But then you need a plugin to broadcast the message.
The join message is broadcast from the server itself as well.

jan64:
Yes, that's the point.
I think broadcasting message using a call from plugin to relevant functions looks cleaner/more apparent in plugin code than returning a string. And a plugin can suppress the built-in message with one of the two returned values (`return true, true` in lua)

madmaxoft:
So you want the broadcasting code BOTH in the server and in a plugin?
With your setup, there's no way to have a plugin that just adds a prefix or suffix to the default messages generated by the server, or display the server-generated message in some other way (log to file or send as hotbar message, whatever).
No, I still think passing the default message in and receiving the potentially modified message out of the plugin is the best way to go.

jan64:
Yes. Plugins can still log all the relevant info, even without receiving the death message contents.
Your solution sounds like an invitation for ugly hacks that parse server-provided death message.

madmaxoft:
Plugins can't log one relevant info in your scenario - the actual death message. Especially if there is one plugin that modifies death messages (MockPlugin - laughs at players for dying) and another one that logs them.
In my scenario, there's no need to parse anything, since all data included in the message will be already included in the TDI structure.

jan64:
What was i trying to say was that this creates room for hackish solutions (doesn't force them, but still) and makes the plugin code less readable. I figured out one scenario where your way is preferable (i don't really see the point of logging death messages), and that is using this as a way to broadcast death messages on the IRC. You've got me convinced, but this leads me to believe that we need 3 hooks. HOOK_BROADCASTING_DEATH_MESSAGE, HOOK_BROADCASTED_DEATH_MESSAGE and normal HOOK_KILLED. (first for plugins to change dm, second for plugins to log dm, and the last one for the death of all entities)

madmaxoft:
We already do have plugin load sequence dependency, so that's no problem. A single hook is still enough, the admin will have to configure correctly if they want multiple DM plugins.

nounoursheureux:
So, should I change the hook to HOOK_KILLED or not ?

madmaxoft:
My opinion is, change to `HOOK_KILLED`, make it called for all entities, make it receive the death message (even for mobs, empty message is ok in such a case) and have the hook return a second value indicating the new value of the message (already done). If the message is empty after the call, don't broadcast. If the victim was an entity, don't broadcast.

nounoursheureux:
Ok, I changed it to `HOOK_KILLED`

madmaxoft:
Looks good to me. Squash into a single commit and it can get merged.

madmaxoft:
You should add the API documentation for the OnKilled hook. It's okay to have it as a separate commit.

nounoursheureux:
Thanks, I wrote it but I forgot to add it :P

MCServer/Plugins/APIDump/Hooks/OnKilled.lua
```diff
@@ -0,0 +1,22 @@
...
+		]]
+		Params = 
+		{
+			{ Name = "Victim", Type = "{{cEntity}}", Notes = "The player or mob that died" },
+			{ Name = "TDI", Type = "{{TakeDamageInfo}}", Notes = "Informations about the death" },
+			{ Name = "DeathMessage", Type = "string", Notes = "The default death message. An empty string if the victim is not a player" },
+		},
+		Returns = [[
+			The function may return two values. The first value is a boolean specifying whether others plugins should be called. If it is true, the others plugins won't get notified of the death. If it is false, the other plugins will get notified.</p>
+			<p>The second value is a string containing the death message. If the victim is a player, this death message is broadcasted instead of the default death message. If it is empty, no death message is broadcasted. If it is nil, the default message is broadcasted. If the victim is not a player, the death message is never broadcasted.</p>
```
madmaxoft: I don't think this is exactly true. If the plugin returns `nil` as the message (or doesn't return the second value, which is equivalent in Lua), the message is left unchanged.

This differs in cases where plugin A modifies the message and then plugin B returns `nil`.

MCServer/Plugins/APIDump/Hooks/OnKilled.lua
```diff
@@ -0,0 +1,22 @@
...
+			This hook is called whenever player or a mob dies. It can be used to change the death message.
+		]]
+		Params = 
+		{
+			{ Name = "Victim", Type = "{{cEntity}}", Notes = "The player or mob that died" },
+			{ Name = "TDI", Type = "{{TakeDamageInfo}}", Notes = "Informations about the death" },
+			{ Name = "DeathMessage", Type = "string", Notes = "The default death message. An empty string if the victim is not a player" },
+		},
+		Returns = [[
+			The function may return two values. The first value is a boolean specifying whether others plugins should be called. If it is true, the others plugins won't get notified of the death. If it is false, the other plugins will get notified.</p>
```
madmaxoft: `others` -> `other`

madmaxoft:
@nounoursheureux Could you fix the two minor problems with the documentation, so that we can get this merged?

nounoursheureux:
Thanks for the reminder, I fixed the problems :ok_hand:

MCServer/Plugins/APIDump/APIDesc.lua
```diff
@@ -2683,7 +2683,7 @@ Parser:close();
 				},
 			},  -- AdditionalInfo
 		},  -- lxp
-
+
```
bearbin: Why has this indentation been added?

nounoursheureux:
Sorry, fixed it

NiLSPACE:
It seems that the APIDump plugin doesn't find the documentation for `HOOK_KILLED`.

nounoursheureux:
Do you know who worked on the API-Dump plugin ? Maybe I need to register the hook somewhere ?

NiLSPACE:
@madmaxoft Any idea if something was forgotten? It should've seen the OnKilled file right?

nounoursheureux:
I compared with [this](https://github.com/cuberite/cuberite/pull/2064/files) pull request but I still don't see what I forgot.

madmaxoft:
```
Cannot load API descriptions from Plugins/APIDump/Hooks/OnKilled.lua, Lua error 'Plugins/APIDump/Hooks/OnKilled.lua:10: '}' expected (to close '{' at line 4) near 'Params''.
```

Fixed in `master`

nounoursheureux:
Thanks :+1:

PR #2199, opened 2015-06-02
Added system and above action bar chat messages

jan64:
Addresses https://github.com/mc-server/MCServer/issues/1803
The only thing missing atm for this to be done is documentation.


jan64:
Also, i don't really like splitting that function into that many others. It creates redundant code.

jan64:
As a side note, i rewrote the CombatLog plugins to use this: https://github.com/jan64/AntiCombatLog

Issue #2170, opened 2015-05-30
Some characters aren't shown correctly on signs


mathias-github:
A few characters aren't shown correctly on signs. Some of them are:

& - Shows as \u0026
= - Shows as \u003d
' - Shows as \u0027
< - Shows as \u003c
.>  - Shows as \u003e

jammet:
Yeah, things like that have been a bit of a pest for a while now.
I also noticed that sometimes, the text written on signs in worlds that used to be first made with the Mojang generators, are out in quotes for some reason.

PR #2201, opened 2015-06-02
Fix for special characters showing incorrectly on signs

beeduck:
Fix for issue #2170
For some reason the buffer for signs is receiving some special characters in JSON code. I checked the Minecraft protocol - signs receive four chat fields, which are the same field that are used for chat messages. For signs specific characters & ' " < > = are still encoded for some reason, but the error is not occurring for chat messages.
I implemented a parser to handle the codes in the mean time.

### DIFF
--- a/src/Protocol/Protocol18x.cpp
+++ b/src/Protocol/Protocol18x.cpp
@@ -2548,10 +2548,32 @@ void cProtocol180::HandlePacketUpdateSign(cByteBuffer & a_ByteBuffer)
 		return;
 	}
 
+	std::unordered_map<AString, AString> SpecialCharacters;
+	SpecialCharacters["\\u0026"] = "&";
+	SpecialCharacters["\\u0027"] = "'";
+	SpecialCharacters["\\\""] = "\"";
+	SpecialCharacters["\\u003c"] = "<";
+	SpecialCharacters["\\u003e"] = ">";
+	SpecialCharacters["\\u003d"] = "=";
+
 	AString Lines[4];
 	for (int i = 0; i < 4; i++)
 	{
 		HANDLE_READ(a_ByteBuffer, ReadVarUTF8String, AString, Line);
+
+		// Buffer receives special characters in JSON code for sign messages, use SpecialCharacters map to parse them
+		for (const auto & itr : SpecialCharacters)
+		{
+			size_t Index = 0;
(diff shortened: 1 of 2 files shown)


src/Protocol/Protocol18x.cpp
```diff
@@ -2548,10 +2548,33 @@ void cProtocol180::HandlePacketUpdateSign(cByteBuffer & a_ByteBuffer)
...
 	AString Lines[4];
 	for (int i = 0; i < 4; i++)
 	{
 		HANDLE_READ(a_ByteBuffer, ReadVarUTF8String, AString, Line);
+
+		// Buffer receives special characters in JSON code for sign messages, use SpecialCharacters map to parse them
+		size_t loc = 0;
+		for (const auto & itr : SpecialCharacters)
+		{
+			while (loc < Line.length())
```
SafwatHalaby: Minor comments: 
1. Putting `loc = 0` on top of the while loop and replacing `size_t loc = 0;` with `size_t loc;` is a bit cleaner in my opinion.
2. Using `Location` or `Index` is less vague than `loc`. Also, we use CamelCase, so `Loc`.

src/Protocol/Protocol18x.cpp
```diff
@@ -2548,10 +2548,33 @@ void cProtocol180::HandlePacketUpdateSign(cByteBuffer & a_ByteBuffer)
 		return;
 	}
 
+	std::map<std::string, std::string> SpecialCharacters;
```
tigerw: I *think* `unordered_set` may be suitable as well.

NiLSPACE: And we use AString instead of std::string

beeduck:
Alright, cleaned up the index counter, switched std::string to  AString and switched std::map to std::unordered_map, which is faster.

madmaxoft:
This is not a good solution. The client could use an encoding for more characters (such as language-specific `ěščřž` etc.) and we should be able to decode all of those. Also, an unordered map is not a good storage for something that needs to be iterated over. And last but not least, the backslash needs to be replaced as the last one, otherwise it would interfere with compounds: `\\\u003e` would get decoded as `>` instead of the correct `\u003e`

I want a more systematic approach, a full json parser with utf-8 support.

PR #2202, opened 2015-06-03
Cmake cleanup

madmaxoft:
Organized the projects into solution folders and removed a legacy target that's not needed (should speed up the builds a bit)


tigerw:
I don't see any indentation, do you see any indentation?

madmaxoft:
What indentation?

tigerw:
Difference view wasn't showing tabulation, sorry.

PR #2203, opened 2015-06-03
Change SendChatType behavior for < 1.8 clients, update docs

jan64:


jan64:
This should fix the TravisCI build.

PR #2205, opened 2015-06-04
Daemon support on Linux

birkett:
Enable MCServer to run as a daemon on *nix like systems. This will make the GSP's happy down the road.

This uses the same "-d" switch, as introduced originally with the Windows service support.

I don't consider this ready to merge yet. It needs testing on Mac and RaspberryPi. I would also like to create a set of example init scripts (Debian, sysvinit and systemd).

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -106,7 +106,7 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> overridesRepo)
 	EnableMenuItem(hmenu, SC_CLOSE, MF_GRAYED);  // Disable close button when starting up; it causes problems with our CTRL-CLOSE handling
 	#endif
 
-	cLogger::cListener * consoleLogListener = MakeConsoleListener();
+	cLogger::cListener * consoleLogListener = MakeConsoleListener(m_RunAsService);
 	cLogger::cListener * fileLogListener = new cFileListener();
 	cLogger::GetInstance().AttachListener(consoleLogListener);
 	cLogger::GetInstance().AttachListener(fileLogListener);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -250,7 +250,7 @@ void universalMain(std::unique_ptr<cSettingsRepositoryInterface> overridesRepo)
 
 
 
-#if defined(_WIN32)
+#if defined(_WIN32)  // Windows service support.
 ////////////////////////////////////////////////////////////////////////////////
 // serviceWorkerThread: Keep the service alive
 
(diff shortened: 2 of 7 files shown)


worktycho:
If youre closing stdout you should probably disable the console log listener, so that it doesnt try to print to closed filehandles.

bearbin:
I don't really think this is necessary, there are better ways to handle persistent processes.

madmaxoft:
I'm not against this change

src/main.cpp
```diff
@@ -17,7 +17,9 @@
 
 #include "MemorySettingsRepository.h"
 
-
+#ifndef _WIN32
+	#include <unistd.h>
```
cengizIO: I think this should be in [Globals.h#L207](https://github.com/mc-server/MCServer/blob/master/src/Globals.h#L207) instead

birkett: Agreed. Moved as of the next commit / squash.

birkett:
The console logging is now skipped when running as a service @worktycho.

@bearbin While there are a whole bunch wrappers and workarounds that can be used, nothing quite beats the simplicity of built-in support. I completely agree that this isn't necessary, but it's polish that many would miss.

madmaxoft:
Me likes. Merge if no-one else complains :)

src/Root.cpp
```diff
@@ -106,11 +106,16 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> overridesRepo)
...
 	#endif
 
-	cLogger::cListener * consoleLogListener = MakeConsoleListener();
+	cLogger::cListener * consoleLogListener = nullptr;
 	cLogger::cListener * fileLogListener = new cFileListener();
-	cLogger::GetInstance().AttachListener(consoleLogListener);
 	cLogger::GetInstance().AttachListener(fileLogListener);
 
+	if (!m_RunAsService)
+	{
```
worktycho: I feel this logic should go in MakeConsoleListerner but I'm not sure how

madmaxoft: `MakeConsoleListener(bool a_IsService)` ?

birkett: Seems like the cleanest way. I've looked into ways of checking if stdout is pointing to a valid handle, but it's all messy and not very cross-platform.

birkett:
I've added cNullConsoleListener instead of just returning null from MakeConsoleListener().

Also went down the route mentioned by @madmaxoft of passing MakeConsoleListener() a bool.

madmaxoft:
You'll need to update the tools, too. MCADefrag uses the logging, and I think ProtoProxy does as well.

cengizIO:
This is very well done I think. Waiting to try this.

PR #2206, opened 2015-06-05
APIDump time stamp

### LINKED ISSUE #2204: APIDump: Include a timestamp in the generated HTML docs
The APIDump plugin should include a timestamp in each of the generated HTML pages, to make it simpler to check if the docs are up to date.
When it is possible to get server commit hash and other build parameters, those would be nice to have in the pages, too.

madmaxoft:
Added timestamp and commit info to the APIDump-generated pages.
In order to do so, the build info was exported to the Lua API and extended so that it is generated even for local builds.

Fixes #2204


worktycho:
Do we want to add a build time dependency on git? That would break bearbins source builds.

madmaxoft:
That's true, I didn't think of that. How does this cope if git is not installed?

bearbin:
We should just add timestamp data to the build, because even if commit information was added, there's no guarantee that the APIDump data has not been modified between compilation and output generation.

I think a better option would be to add timestamp and an info hash, which is just a sha1 of the output files.

worktycho:
At the moment I think configure fails if git is not installed, or the folder is not a git repo.

madmaxoft:
I've added some error handling so that when git is not available or the source folder is not a git repo, the commit ID and datetime are set to "unknown".

I believe this is now a reasonable state. We mostly need this for official builds (for which it worked even before) and for some semi-official builds, such as the one through which LuaAPI docs get updated. Both of these work with reasonable amounts of information being available, and regular builds even on clean systems should still be possible, while returning no information.

tigerw:
Reasonable.

PR #2207, opened 2015-06-05
Fixed overshoot in cLineBlockTracer.

### LINKED ISSUE #2161: cLineBlockTracer::Trace doesn't respect input vector boundaries
If I understand the Trace function properly, you give it 2 vectors and then it does something based on OnNextBlock. But the vector boundaries given in the Trace function aren't being respected, which is the cause of #2059. There is also a lot of unused variables in the function.
Is this intentional?
EDIT: added info

madmaxoft:
Fixes #2161.


madmaxoft:
Note that this change may produce different trace results for lines that go exactly through block's edge-lines or corners. The previous code favored transitions in the XYZ order, current code favors transitions in the ZYX order. It doesn't affect transitions through regular block walls.

PR #2209, opened 2015-06-05
Refactoring PathFinding.

SafwatHalaby:
DO NOT MERGE YET.

This is a work in progress, thought I'd share to get a second opinion. Especially about the PathFinder.h interface. `PathFinder.cpp` Is just a dirty barely working implementation intended for development only. Do not take it seriously.

### DIFF
--- a/src/Mobs/PathFinder.h
+++ b/src/Mobs/PathFinder.h
@@ -1,0 +1,16 @@
+#pragma once
+class cPathFinder;
+class cPath;
+class cChunk;
+
+class cPathFinder
+{
+public:
+	cPathFinder(double a_BoundingBoxWidth, double a_BoundingBoxHeight);
+	Vector3d GetWayPoint(cChunk & a_Chunk, const Vector3d & a_CurrentLocation, Vector3d & a_Destination);
+private:
+	bool EnsureProperDestination(cChunk & a_Chunk, Vector3d & a_Destination);
+	std::unique_ptr<cPath> m_Path;
+	int m_BoundingBoxWidth;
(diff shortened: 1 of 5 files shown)


src/Mobs/PathFinder.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+class cPathFinder;
+class cPath;
+class cChunk;
+
+class cPathFinder
+{
+public:
+	cPathFinder(double a_BoundingBoxWidth, double a_BoundingBoxHeight);
+	Vector3d GetWayPoint(cChunk & a_Chunk, const Vector3d & a_CurrentLocation, Vector3d & a_Destination);
```
SafwatHalaby: a_Destination is not a const because PathFinder may decide that it's a bad destination and pick a close one. This is probably not the best way to do it and should be changed. PathFinder should probably signal that it only found a nearby path.

SafwatHalaby:
The idea is this: `cPath` is a dumb device which takes 2 points and outputs a path.
`cPathFinder` is a cPathWrapper, which you tell it where you are and where you want to go. It smartly calls cPath as little as possible and takes care of path recalculation when needed.

worktycho:
I like where this is going. I suspect the interface is going to get more complex, but moving stuff out of cMonster is always good.

SafwatHalaby:
Tycho: I'll do my best to keep the interface simple. It needs a slight modification so that when nearby paths are calculated, the interface user can know that. Other than that' I'll try to keep it simple.

Due to lack of time, I'm closing this for now. I'll re-submit the complete PR later.

PR #2211, opened 2015-06-07
Fix incorrent experience sending for 1.7 clients

jan64:
Fixes https://github.com/mc-server/MCServer/issues/2079


tigerw:
I suspect what xoft would say is...

"Good catch!"
    - xoft

madmaxoft:
Yeah, I'd say that, but we've already caught it once, it just was a different-topic PR so I put it in the issue instead. The issue was very specific about these two lines.

PR #2212, opened 2015-06-07
Revert "Chunk queue collapsing"

worktycho:
Reverts mc-server/MCServer#2179


bearbin:
This was needed because there is a major bug with chunk sending introduced with the PR this reverts.

PR #2220, opened 2015-06-08
Namechange to Cuberite

SafwatHalaby:
Do not merge yet. I changed the occurrences of MCServer to Cuberite using a regex. It caught a low amount of occurrences, so please double-check my regex. Also, since this is an automatic change, all changed lines should be manually inspected.
The new code compiles and runs fine.
Here are the used bash commands:

```
#Directory is MCServer/src
find . -type f | grep .cpp | while read line; do perl -pe 's/(mc ?server)/Cuberite/gi' -i $line; done
find . -type f | grep .h | while read line; do perl -pe 's/(mc ?server)/Cuberite/gi' -i $line; done
```

~~Edit: Now I also executed the same command, but with this regex:
`s/(mc\-server)/Cuberite/gi`~~ Edit2: This was reverted.

### DIFF
--- a/src/Bindings/PluginLua.h
+++ b/src/Bindings/PluginLua.h
@@ -14,8 +14,8 @@
 #include "LuaState.h"
 
 // Names for the global variables through which the plugin is identified in its LuaState
-#define LUA_PLUGIN_NAME_VAR_NAME     "_MCServerInternal_PluginName"
-#define LUA_PLUGIN_INSTANCE_VAR_NAME "_MCServerInternal_PluginInstance"
+#define LUA_PLUGIN_NAME_VAR_NAME     "_CuberiteInternal_PluginName"
+#define LUA_PLUGIN_INSTANCE_VAR_NAME "_CuberiteInternal_PluginInstance"
 
 
 
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -67,8 +67,8 @@ public:
 	/// The type used for any heightmap operations and storage; idx = x + Width * z; Height points to the highest non-air block in the column
 	typedef HEIGHTTYPE HeightMap[Width * Width];
 
-	/** The type used for any biomemap operations and storage inside MCServer,
-	using MCServer biomes (need not correspond to client representation!)
+	/** The type used for any biomemap operations and storage inside Cuberite,
+	using Cuberite biomes (need not correspond to client representation!)
 	idx = x + Width * z  // Need to verify this with the protocol spec, currently unknown!
 	*/
 	typedef EMCSBiome BiomeMap[Width * Width];
--- a/src/Protocol/Authenticator.cpp
+++ b/src/Protocol/Authenticator.cpp
@@ -149,7 +149,7 @@ bool cAuthenticator::AuthWithYggdrasil(AString & a_UserName, const AString & a_S
 	AString Request;
 	Request += "GET " + ActualAddress + " HTTP/1.0\r\n";
 	Request += "Host: " + m_Server + "\r\n";
-	Request += "User-Agent: MCServer\r\n";
+	Request += "User-Agent: Cuberite\r\n";
 	Request += "Connection: close\r\n";
 	Request += "\r\n";
 
@@ -223,7 +223,7 @@ bool cAuthenticator::GetPlayerProperties(const AString & a_UUID, Json::Value & a
 	AString Request;
 	Request += "GET " + ActualAddress + " HTTP/1.0\r\n";
 	Request += "Host: " + m_Server + "\r\n";
-	Request += "User-Agent: MCServer\r\n";
+	Request += "User-Agent: Cuberite\r\n";
 	Request += "Connection: close\r\n";
 	Request += "\r\n";
 
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -187,7 +187,7 @@ void cServer::PlayerDestroying(const cPlayer * a_Player)
 
 bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_ShouldAuth)
 {
-	m_Description = a_Settings.GetValueSet("Server", "Description", "MCServer - in C++!");
+	m_Description = a_Settings.GetValueSet("Server", "Description", "Cuberite - in C++!");
 	m_MaxPlayers  = a_Settings.GetValueSetI("Server", "MaxPlayers", 100);
 	m_bIsHardcore = a_Settings.GetValueSetB("Server", "HardcoreEnabled", false);
 	m_bAllowMultiLogin = a_Settings.GetValueSetB("Server", "AllowMultiLogin", false);
--- a/src/WorldStorage/WSSAnvil.cpp
+++ b/src/WorldStorage/WSSAnvil.cpp
@@ -898,7 +898,7 @@ cBlockEntity * cWSSAnvil::LoadChestFromNBT(const cParsedNBT & a_NBT, int a_TagId
 	// Check if the data has a proper type:
 	// TODO: Does vanilla use "TrappedChest" or not? MCWiki says no, but previous code says yes
 	// Ref.: http://minecraft.gamepedia.com/Trapped_Chest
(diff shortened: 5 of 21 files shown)


src/ChunkDef.h
```diff
@@ -67,8 +67,8 @@ class cChunkDef
 	/// The type used for any heightmap operations and storage; idx = x + Width * z; Height points to the highest non-air block in the column
 	typedef HEIGHTTYPE HeightMap[Width * Width];
 
-	/** The type used for any biomemap operations and storage inside MCServer,
-	using MCServer biomes (need not correspond to client representation!)
+	/** The type used for any biomemap operations and storage inside Cuberite,
+	using Cuberite biomes (need not correspond to client representation!)
```
SafwatHalaby: Cuberite biomes. Is this ok?

src/Server.cpp
```diff
@@ -187,7 +187,7 @@ void cServer::PlayerDestroying(const cPlayer * a_Player)
 
 bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_ShouldAuth)
 {
-	m_Description = a_Settings.GetValueSet("Server", "Description", "MCServer - in C++!");
+	m_Description = a_Settings.GetValueSet("Server", "Description", "Cuberite - in C++!");
```
SafwatHalaby: Perhaps `Cuberite - MC in C++`?

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -898,7 +898,7 @@ cBlockEntity * cWSSAnvil::LoadChestFromNBT(const cParsedNBT & a_NBT, int a_TagId
 	// Check if the data has a proper type:
 	// TODO: Does vanilla use "TrappedChest" or not? MCWiki says no, but previous code says yes
 	// Ref.: http://minecraft.gamepedia.com/Trapped_Chest
-	//       https://github.com/mc-server/MCServer/blob/d0551e2e0a98a28f31a88d489d17b408e4a7d38d/src/WorldStorage/WSSAnvil.cpp#L637
+	//       https://github.com/mc-server/Cuberite/blob/d0551e2e0a98a28f31a88d489d17b408e4a7d38d/src/WorldStorage/WSSAnvil.cpp#L637
```
SafwatHalaby: Will need changing once the organization is moved.

SafwatHalaby:
Executed `s/(mc\-server)/Cuberite/gi`

Does any of this break LUA stuff?

src/Chunk.cpp
```diff
@@ -828,7 +828,7 @@ void cChunk::CheckBlocks()
 void cChunk::TickBlocks(void)
 {
 	// Tick dem blocks
-	// _X: We must limit the random number or else we get a nasty int overflow bug - http://forum.mc-server.org/showthread.php?tid=457
+	// _X: We must limit the random number or else we get a nasty int overflow bug - http://forum.Cuberite.org/showthread.php?tid=457
```
SafwatHalaby: Bad link.

src/ChunkMap.cpp
```diff
@@ -3130,7 +3130,7 @@ void cChunkMap::cChunkLayer::UnloadUnusedChunks(void)
 		{
 			// The cChunk destructor calls our GetChunk() while removing its entities
 			// so we still need to be able to return the chunk. Therefore we first delete, then nullptrify
-			// Doing otherwise results in bug http://forum.mc-server.org/showthread.php?tid=355
+			// Doing otherwise results in bug http://forum.Cuberite.org/showthread.php?tid=355
```
SafwatHalaby: Bad link.

src/ClientHandle.cpp
```diff
@@ -1369,7 +1369,7 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 	{
 		// Only compare ItemType, not meta (torches have different metas)
 		// The -1 check is there because sometimes the client sends -1 instead of the held item
-		// Ref.: http://forum.mc-server.org/showthread.php?tid=549&pid=4502#pid4502
+		// Ref.: http://forum.Cuberite.org/showthread.php?tid=549&pid=4502#pid4502
```
SafwatHalaby: Bad link.

SafwatHalaby:
Removing all my comments and re-commenting on the diff.

SafwatHalaby:
Added the wrong type of comments again. Ah well.

I've reverted `s/(mc\-server)/Cuberite/gi` because all it did was break mc-server.org links.

SafwatHalaby:
Just realized `-i` accepts multiple files, my bash was overkill.

src/Protocol/Authenticator.cpp
```diff
@@ -223,7 +223,7 @@ bool cAuthenticator::GetPlayerProperties(const AString & a_UUID, Json::Value & a
 	AString Request;
 	Request += "GET " + ActualAddress + " HTTP/1.0\r\n";
 	Request += "Host: " + m_Server + "\r\n";
-	Request += "User-Agent: MCServer\r\n";
+	Request += "User-Agent: Cuberite\r\n";
```
bearbin: @worktycho Said that we should use something like

    Cuberite (MCServer)

here. I'm not quite sure of the reason behind it though.

src/WorldStorage/WSSAnvil.cpp
bearbin: It should also be a lower case "c" for the repo name.

bearbin:
Looks good apart from the places that you commented.

SafwatHalaby:
Do you think my regex caught em all?

src/Bindings/PluginLua.h
```diff
@@ -14,8 +14,8 @@
 #include "LuaState.h"
 
 // Names for the global variables through which the plugin is identified in its LuaState
-#define LUA_PLUGIN_NAME_VAR_NAME     "_MCServerInternal_PluginName"
-#define LUA_PLUGIN_INSTANCE_VAR_NAME "_MCServerInternal_PluginInstance"
+#define LUA_PLUGIN_NAME_VAR_NAME     "_CuberiteInternal_PluginName"
```
SafwatHalaby: I have no clue abot Lua. Will this break anything?

NiLSPACE: No idea what plugin is using this, but yes it could break some things.

src/Protocol/Authenticator.cpp
worktycho: General convention is that when changing name you keep the old name in you're user agent so that anything that matched the old useragent continues to match.

madmaxoft: Shouldn't it be `MCServer (Cuberite)` then?

src/Bindings/PluginLua.h
madmaxoft: No, it won't break things. Plugins are NOT supposed to access this, it is only used internally by MCS to identify the `cPluginLua` instance from within the `cLuaState`

src/Protocol/Authenticator.cpp
worktycho: I'm not sure.

SafwatHalaby: Regardless of the old name, it's wise to give the server an idea of what you are.

SafwatHalaby: We're still young. I don't think that changing to pure `Cuberite` at this point will break anyone's things.

bearbin: I agree - `Cubeite (MCServer)` or just `Cuberite` would be the right option here.

src/Bindings/PluginLua.h
NiLSPACE: It is in the [Global](http://mc-server.xoft.cz/LuaAPI/Globals.html) documentation, so I guess it is exported.

madmaxoft: Then we have to remove it from documentation, this is really something a plugin should never touch.

SafwatHalaby:
I'll treat the comments and merge it this weekend.

SphinxC0re:
any updates on this one?

madmaxoft:
Closing, I rebased the changes and pushed them as #2490

PR #2221, opened 2015-06-08
cBlockArea: Added CountSpecificBlocks() API function.

madmaxoft:


src/BlockArea.cpp
```diff
@@ -1665,6 +1665,65 @@ size_t cBlockArea::CountNonAirBlocks(void) const
...
+
+
+
+
+size_t cBlockArea::CountSpecificBlocks(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) const
+{
+	// If blocktypes are not valid, log a warning and return zero occurences:
+	if (m_BlockTypes == nullptr)
+	{
+		LOGWARNING("%s: BlockTypes not available!");
```
worktycho: What are you trying to do with this string format?

madmaxoft: Forgot to add the function name. I shouldn't be coding this late :P

SphinxC0re:
Looks good

PR #2224, opened 2015-06-10
Chunk queue collapsing: MK II

tigerw:
@worktycho:
> This change should provide major performance increases to larger servers.

Everything appears to work now :)

### DIFF
--- a/src/ChunkSender.h
+++ b/src/ChunkSender.h
@@ -1,4 +1,4 @@
-
+﻿
 // ChunkSender.h
 
 // Interfaces to the cChunkSender class representing the thread that waits for chunks becoming ready (loaded / generated) and sends them to clients
@@ -61,8 +61,8 @@ public:
 
 	enum eChunkPriority
 	{
-		E_CHUNK_PRIORITY_HIGH   = 0,
-		PRIORITY_BROADCAST,
+		E_CHUNK_PRIORITY_HIGH = 0,
+		E_CHUNK_PRIORITY_MIDHIGH,
 		E_CHUNK_PRIORITY_MEDIUM,
 		E_CHUNK_PRIORITY_LOW,
 		
@@ -86,7 +86,16 @@ protected:
 		eChunkPriority m_Priority;
 		cChunkCoords m_Chunk;
 
-		bool operator <(const sChunkQueue & a_Other) const { return this->m_Priority < a_Other.m_Priority; }
+		bool operator <(const sChunkQueue & a_Other) const
+		{
+			/* The Standard Priority Queue sorts from biggest to smallest
+			return true here means you are smaller than the other object, and you get pushed down.
+
+			The priorities go from HIGH (0) to LOW (2), so a smaller priority should mean further up the list
+			therefore, return true (affirm we're "smaller", and get pushed down) only if our priority is bigger than theirs (they're more urgent)
+			*/
+			return this->m_Priority > a_Other.m_Priority;
+		}
 	};
 
 	/// Used for sending chunks to specific clients
@@ -107,9 +116,9 @@ protected:
 	cCriticalSection  m_CS;
 	std::priority_queue<sChunkQueue> m_SendChunks;
 	std::unordered_map<cChunkCoords, sSendChunk, cChunkCoordsHash> m_ChunkInfo;
-	cEvent            m_evtQueue;  // Set when anything is added to m_ChunksReady
-	cEvent            m_evtRemoved;  // Set when removed clients are safe to be deleted
-	int               m_RemoveCount;  // Number of threads waiting for a client removal (m_evtRemoved needs to be set this many times)
+	cEvent m_evtQueue;  // Set when anything is added to m_ChunksReady
+	cEvent m_evtRemoved;  // Set when removed clients are safe to be deleted
+
 	// Data about the chunk that is being sent:
 	// NOTE that m_BlockData[] is inherited from the cChunkDataCollector
 	unsigned char m_BiomeMap[cChunkDef::Width * cChunkDef::Width];
--- a/src/ChunkSender.h
+++ b/src/ChunkSender.h
@@ -29,6 +29,9 @@ Note that it may be called by world's BroadcastToChunk() if the client is still 
 #include "ChunkDef.h"
 #include "ChunkDataCallback.h"
 
+#include <unordered_set>
+#include <unordered_map>
+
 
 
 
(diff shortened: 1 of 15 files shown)


bearbin:
What was the problem in tycho's code?

tigerw:
Now:
```cpp
		while (m_SendChunks.empty())
		{
			int RemoveCount = m_RemoveCount;
			m_RemoveCount = 0;
			cCSUnlock Unlock(Lock);
			for (int i = 0; i < RemoveCount; i++)
			{
				m_evtRemoved.Set();  // Notify that the removed clients are safe to be deleted
			}
			m_evtQueue.Wait();
			if (m_ShouldTerminate)
			{
				return;
			}
		}
```
Original:
```cpp
		do
		{
			int RemoveCount = m_RemoveCount;
			m_RemoveCount = 0;
			cCSUnlock Unlock(Lock);
			for (int i = 0; i < RemoveCount; i++)
			{
				m_evtRemoved.Set();  // Notify that the removed clients are safe to be deleted
			}
			m_evtQueue.Wait();
			if (m_ShouldTerminate)
			{
				return;
			}
		} while (m_SendChunks.empty());
```
I believe it's because m_evtQueue.Wait() can block, and if it is called and does block before m_SendChunks has been processed, some chunks aren't sent.

worktycho:
However, either the queue should be empty, or the event should be set when it enters that loop, so the do while was intended.

tigerw:
What if the queue isn't empty but it started waiting on the condition variable because it checks later, and nothing signals the variable to resume? The thread stops executing doesn't it?

worktycho:
If the queue isn't empty the cond var should be in the signaled state, or the code adding to the queue is broken.

tigerw:
Well, it isn't necessarily going to process one by one - queuing is fast, but the sending is slow, and by the time the server had queued all chunks (and stops notifying) the thread is still sending them, and so gets blocked.

Besides, it didn't work before :P

worktycho:
The cond var stays in the set state until its waited on. So that shouldn't cause the problem. And I want to understand why it stopped working, because I want to be sure it is properly fixed.

tigerw:
Right, but unless notify() counts how many times it is called and then unblocks wait() until that many times, the posting thread can finish its work before the sending thread has, no? So the next time the send thread waits, nothing wakes it up.

worktycho:
I think I see the problem now. The issue was that the old code had a separate check for removed clients at the end of each loop. This check did not wait on the queue event which the top one does, so removing the bottom one is invalid. So the correct fix is to also revert removing the check on m_RemoveCount at the end, or else we could get a deadlock on remove client.

tigerw:
The problem with the missing chunks was due to m_evtQueue.Wait() blocking, but this old code which you mention doesn't touch m_evtQueue at all, so it must be a separate problem.

worktycho:
That's the point, I tried to merge a piece of code that didn't mention m_evtQueue, with the loop which did. This resulted in an extra call to wait, causing the bug.

tigerw:
Okay. Now?

worktycho:
No. Still wrong. I'll fix it tomorrow.

tigerw:
How?

worktycho:
I think I've fixed it now.

madmaxoft:
CIs think something else :D

tigerw:
I don't understand - why is this m_RemoveCount thing needed? When a client is removed in cChunkSender::RemoveClient, all presence of that a_Client is removed from m_ChunkInfo. Why wait for the Execute thread to acknowledge this, perform no processing on it, and *then* resume execution?

Example:
```cpp
::Execute()
/*
			int RemoveCount = m_RemoveCount;
			m_RemoveCount = 0;
			cCSUnlock Unlock(Lock);
			for (int i = 0; i < RemoveCount; i++)
			{
*/
				m_evtRemoved.Set();  // Notify that the removed clients are safe to be deleted
				ButForWhat?();
/*
			}
			m_evtQueue.Wait();
			if (m_ShouldTerminate)
			{
				return;
			}
*/
```
Because here:
```cpp
::RemoveClient()
/*
{
		cCSLock Lock(m_CS);
		for (auto && pair : m_ChunkInfo)
		{
			auto && clients = pair.second.m_Clients;
			clients.erase(a_Client); <-- NOTE
		}

		m_RemoveCount++;
	}
	m_evtQueue.Set();
*/
	m_evtRemoved.Wait();  // By this point, a_Client has already been removed, what is this for? Why wait?
}
```
It seems like legacy code.

tigerw:
Anyone?

worktycho:
I need to think about it.

tigerw:
```
Thinking...    59%
[========       ]
```

worktycho:
The wait is necessary so that if a chunk is being sent to the client when they are removed, it waits until that has finished before leaving the method. This guarantees that cChunkSender has no remaining references to the client, after the method is called.

tigerw:
Evidently, I'm not altogether thorough at the subtleties of threading, but...

How about now?

worktycho:
Anyone want to review this?

src/ChunkSender.h
```diff
@@ -53,87 +56,73 @@ class cChunkSender:
...
-			return (
-				(a_Other.m_ChunkX == m_ChunkX) &&
-				(a_Other.m_ChunkZ == m_ChunkZ) &&
-				(a_Other.m_Client == m_Client)
-			);
+			/* The Standard Priority Queue sorts from biggest to smallest
+			return true here means you are smaller than the other object, and you get pushed down.
+
+			The priorities go from HIGH (0) to LOW (2), so a smaller priority should mean further up the list
+			∴ return true (affirm we're "smaller", and get pushed down) only if our priority is bigger than theirs (they're more urgent)
```
madmaxoft: Please remove the weird character here, we want our sources to be 7-bit clean

src/ChunkSender.h
```diff
@@ -53,87 +56,73 @@ class cChunkSender:
...
-	cChunkSender(void);
+	cChunkSender(cWorld & a_World);
 	~cChunkSender();
 
 	enum eChunkPriority
 	{
-		E_CHUNK_PRIORITY_HIGH   = 0,
-		E_CHUNK_PRIORITY_MEDIUM = 1,
-		E_CHUNK_PRIORITY_LOW    = 2,
+		E_CHUNK_PRIORITY_LEGENDARY = 0,
```
madmaxoft: Come on!

src/OSSupport/Event.h
```diff
@@ -37,7 +42,7 @@ class cEvent
 private:
 
 	/** Used for checking for spurious wakeups. */
-	bool m_ShouldWait;
+	std::atomic<bool> m_ShouldContinue;
 
 	/** Mutex protecting m_ShouldWait from multithreaded access. */
```
madmaxoft: Fix the comment

src/OSSupport/IsThread.h
```diff
@@ -51,13 +47,18 @@ class cIsThread
 
 	/** Returns true if the thread calling this function is the thread contained within this object. */
 	bool IsCurrentThread(void) const { return std::this_thread::get_id() == m_Thread.get_id(); }
+	
+private:
+	/** Wrapper for Execute() that waits for the initialization event, to prevent race conditions in thread initialization. */
+	void DoExecute(void);
```
madmaxoft: We usually sort things types first, then variables and last functions.

src/OSSupport/IsThread.h
```diff
@@ -51,13 +47,18 @@ class cIsThread
...
 
-protected:
 	AString m_ThreadName;
 	std::thread m_Thread;
 
 	/** The event that is used to wait with the thread's execution until the thread object is fully initialized.
-	This prevents the IsCurrentThread() call to fail because of a race-condition. */
+	This prevents the IsCurrentThread() call to fail because of a race-condition where the thread starts before m_Thread has been fully assigned.
+	Note: dependant on the cEvent behaviour of a call to Wait() not blocking if Set() was called prior with no threads waiting
+	m_Thread can initialise faster than (and call Set() before) the child thread can be created. */
```
madmaxoft: Is the note still valid? It doesn't make much sense to me.

src/ChunkSender.h
tigerw: Are you objecting against the humour?

src/ChunkSender.h
tigerw: Okay, but is there a reason not to use Unicode files?

src/OSSupport/IsThread.h
tigerw: I meant that the behaviour of cEvent (the next Wait won't block if Set called whilst nothing is waiting) is important since the Set call in the caller can be reached faster than the Wait call in the created thread.

src/ChunkSender.h
madmaxoft: Yes, people without the cultural references will not "get" it and it makes the source needlessly difficult to maintain.

src/ChunkSender.h
madmaxoft: Some platforms / IDEs / compilers don't support them. Not sure what the standard says about this.

src/OSSupport/IsThread.h
madmaxoft: But the behavior is consistent everywhere, so there's no need to make this note here.

src/ChunkSender.h
worktycho: The standard doesn't even give you all of ASCII, you get a-z, A-Z, 0-9 and _ { } [ ] # ( ) < > % : ; . ? * + - / ^ & | ~ ! = , \ " ’ plus space, hroizontal and vertical tabs, newline and form-feed.

src/ChunkSender.h
tigerw: About this, it was added to push the original HIGH higher, to fit a priority between high and medium - this was the position where the old m_ChunksReady list was processed, which was where cNotifyChunkSender put its chunks. Is it possible to just have three priorities, and give cNotifyChunkSender a MEDIUM priority?

worktycho: Or you can do what I did, and call it broadcast priority.

tigerw: That doesn't fit well into the E_CHUNK_PRIORITY naming scheme though.

worktycho: I think it seems to fit in well enough, since you are going to have to go to the enum to work out relative priority anyway most of the time. What would be better would be if we could associate the priority of the original request, but I didn't want to have to work out the race conditions of that.

worktycho:
@madmaxoft Do you have any more comments, or can we merge this?

madmaxoft:
To be honest I don't have the energy to review it again. If it is tested and it works, why not. I'd appreciate it if someone rebased it onto current master and squashed the commits, though.

PR #2231, opened 2015-06-11
SelfTests are registered and executed after logging framework init.

### LINKED ISSUE #2228: Change self-tests to execute after logging has been initialized
Currently the self-tests are executed in the constructors of global objects. This has the unfortunate side-effect that the logging framework is not initialized with console output at that time.

I propose we change the way self-tests are done. The global constructor should register the self-test for later execution, and then once logging is initialized, the queued tests should get executed.

madmaxoft:
Fixes #2228.

### DIFF
--- a/src/BoundingBox.cpp
+++ b/src/BoundingBox.cpp
@@ -5,6 +5,7 @@
 #include "Globals.h"
 #include "BoundingBox.h"
 #include "Defines.h"
+#include "SelfTests.h"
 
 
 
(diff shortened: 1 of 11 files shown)


tigerw:
Would it be possible to initialise the logging framework just for testing?

madmaxoft:
Not really, because one needs `cRoot::m_IsService` in order to initialize the logging, so you need to read the cmdline params first, and it gets hairier and hairier. This way is much simpler.

madmaxoft:
I wonder why AppVeyor is not testing the PR

worktycho:
I think that the self tests frame work should be outside cRoot, but really these tests should be in separate executables.

madmaxoft:
I'm moving it out of Root.h, otherwise it gets included in ProtoProxy (hence the CI failures).
Moving them out of the executable is out of scope for now.

madmaxoft:
I think this is now ready for a merge. Anyone opposed?

SphinxC0re:
Looks good to me

Issue #1661, opened 2014-12-13
Bad door generation in villages.


SamJBarney:
The image speaks for itself. All of the houses have this same problem:
![2014-12-12_20 02 31](https://cloud.githubusercontent.com/assets/1827860/5422314/414bee5a-823a-11e4-86a5-ede1fc4bae34.png)

mathiascode:
Can't reproduce anymore.

PR #2237, opened 2015-06-13
Fixing how village doors are generated in 1.8.

lkolbly:
Fix for #1661 in Minecraft 1.8 (removed bit 3 in the door's upper block and fixed typo in BlockDoor.h).

I'm not sure if manually applying the fix to the prefab files is the proper way to do this, since they're apparently generated, but they were committed so I changed them in a commit.

### DIFF
--- a/src/Blocks/BlockDoor.h
+++ b/src/Blocks/BlockDoor.h
@@ -242,7 +242,7 @@ public:
 		if ((Meta & 0x08) != 0)
 		{
 			// The coords are pointing at the top part of the door
-			if (a_BlockX > 0)
+			if (a_BlockY > 0)
 			{
 				NIBBLETYPE DownMeta = a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY - 1, a_BlockZ);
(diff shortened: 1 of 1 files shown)


madmaxoft:
This is probably not enough to fix it. The prefabs are generated from the Gallery server ( http://forum.mc-server.org/showthread.php?tid=1372 ) and if the bits are wrong there, it means the server has some broken code that generates the wrong bits. Your BlockDoor.h fix doesn't do anything bit-wise, so I assume the core issue is not yet fixed.
As for the prefabs, it'd be better to regenerate them after the door issue is fixed on the Gallery server. Your current changes would get erased with the next reexport anyway.

lkolbly:
@madmaxoft Yeah, the BlockDoor.h change doesn't fix the entire issue. It surprises me that bit 3 matters at all, since bit 3 on doors is allegedly unused on the upper part.

I'll take a look at the Gallery server, and see what I can find. In the meantime I've changed this pull request to be super insanely small (fun sized pull request?).

tigerw:
Fun sized sounds good.

PR #2239, opened 2015-06-13
Added a size check for portals.

NiLSPACE:
The minimum size is 3x2, and the maximum is 23x23


tigerw:
Are you sure? I intended portals purposefully to be any size, unlike Vanilla.

NiLSPACE:
I thought it was cool as well, but since we're trying to be vanilla like I thought I'd add these checks. Anyone else who has an opinion?

bearbin:
Vanilla size restrictions would probably be a good idea, portals larger than 23x23 probably aren't actually that useful.

I don't think there's a point in adding minimum sizes though.

tigerw:
But they could look cool.

NiLSPACE:
@bearbin the minimum size is also vanilla.

bearbin:
I know, but I don't really see a reason for it. A maximum is reasonable, because of technical limitations, but I don't see a reason for a minimum.

NiLSPACE:
My bad, I thought since you said `Vanilla size restrictions would probably be a good idea` you didn't know the minimum restrictions were also vanilla.

tigerw:
How about a maximum twice as large as Vanilla, and no minimum?

NiLSPACE:
But if we don't have a minimum, then we can't implement [this](https://github.com/cuberite/cuberite/issues/1856#issuecomment-105779965) right?

tigerw:
Hm. Neat little small portals for item delivery versus potential speed improvements. It's a dilemma!

Example:

lkolbly:
What if you made it configurable in world.ini or something? Then it's not a dilemma. :-)

madmaxoft:
Configurable in world.ini, with vanilla-based defaults sounds good to me.

SphinxC0re:
Update?

NiLSPACE:
Whoops, I forgot about this, sorry.

All I have to do is put Get/Set functions in `cWorldInterface` and override them in `cWorld` right?

madmaxoft:
To make the size configurable, you need to store the values in the `cWorld`, add reading and writing them in the initialization, and of course modify the check to actually use them. I don't think there's any need to add them to `cWorldInterface`.

NiLSPACE:
But the check [receives a `cWorldInterface` ](https://github.com/cuberite/cuberite/pull/2239/files#diff-5a7e76bfee5ab82d3fb92e5065bbef08R106)

NiLSPACE:
**bumb**

NiLSPACE:
**bumb2**

lkolbly:
I agree with @NiLSPACE that you would have to modify cWorldInterface, since you can't directly get a cWorld from the OnPlaced function. Would it be crazy to have a GetWorld() function in the cWorldInterface, so that you don't have to add a new function for every config parameter?

tigerw:
I believe worktycho agreed that the interfaces weren't really being useful and could be removed?

In the meantime, I think it is possible to do the normal procedure of adding functions to cWorld - adding a virtual in the Interface and overriding it in World.

tigerw:
Also, NiL is **bumb**ling around the place!

madmaxoft:
I didn't check, if it receives an interface, then it would need to be extended (or removed).

NiLSPACE:
I added them to `cWorldInterface`. I think it's too unrelated to this PR to do.

tigerw:
Okay, looks good.

One final point: I read somewhere that getters and setters are not ideal. Instead, there should be a more holistic method which, in this case, would set the min and max, width and height, based on an object/a set a parameters.

Thoughts?

madmaxoft:
Getters and setters are not ideal mostly in interpreted languages. In languages compiled to native code, the compiler optimizes most of them away anyway, so performance-wise they're good.

madmaxoft:
Fix, rebase and squash and we can merge :)

SphinxC0re:
@madmaxoft has to restart the first clang build

worktycho:
@SphinxC0re I've restarted the build.

NiLSPACE:
My god git is frustrating....... I just can't get it to work.

tigerw:
Okay, looks good.

One final point: I read somewhere that getters and setters are not ideal. Instead, there should be a more holistic method which, in this case, would set the min and max, width and height, based on an object/a set a parameters.

Thoughts?

tigerw:
Sorry, repost.

SphinxC0re:
@NiLSPACE just ```git rebase master``` then pick the first one and squash all the commits below. After that, force-push to the branch

NiLSPACE:
> `git rebase -i master`

```
fatal: Needed a single revision
invalid upstream master
```

I don't think git likes me :|

NiLSPACE:
I did it! With TortoiseGit I finaly managed to do it.

SphinxC0re:
@NiLSPACE great :+1:

PR #2241, opened 2015-06-13
Work on NetherPortalScanner. Setup portal scanner to reset PortalCool…

lkolbly:
…down. Changed where player is spawned. Added a_InitSpawn flag to CreateAndInitializeWorld.

The CreateAndInitializeWorld flag was created because as far as I know, the Nether does not require a spawn area (**correct me if I'm wrong, this seems like a big but valid assumption to me**). DetectPortal sends a player respawn packet, but this packet is not actually sent until DetectPortal returns, which is after CreateAndInitializeWorld is called. Creating a spawn area is time-consuming, so sometimes the packet can be delayed for a while (many seconds).

This PR also contains minor stylistic changes to some comments I found. Also ScheduleMoveToWorld can set the PortalCooldown. Also actually checks the MaxY limit.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1269,11 +1269,12 @@ void cEntity::DetectCacti(void)
 
 
 
-void cEntity::ScheduleMoveToWorld(cWorld * a_World, Vector3d a_NewPosition)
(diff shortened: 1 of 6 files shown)


lkolbly:
Could I get a yay/nay on this please?

PR #2243, opened 2015-06-14
Add new "ForEach" function: ForEachLoadedChunk.

warmist:
My first attempt at adding a new foreach function. This would come in handy in world modifying functions.

### DIFF
--- a/src/Bindings/ManualBindings_World.cpp
+++ b/src/Bindings/ManualBindings_World.cpp
@@ -110,6 +110,73 @@ static int tolua_cWorld_ChunkStay(lua_State * tolua_S)
 
 
 
+static int tolua_cWorld_ForEachLoadedChunk(lua_State * tolua_S)
+{
+    // Exported manually, because i don't know what i'm doing...
+    // Function signature: ForEachLoadedChunk(callback)
+
+    cLuaState L(tolua_S);
+    if (
+        !L.CheckParamUserType(1, "cWorld") ||
+        !L.CheckParamFunction(2)
+        )
+    {
+        return 0;
+    }
+
+    cPluginLua * Plugin = cManualBindings::GetLuaPlugin(tolua_S);
+    if (Plugin == nullptr)
+    {
+        return 0;
+    }
+
+    // Read the params:
+    cWorld * World = (cWorld *)tolua_tousertype(tolua_S, 1, nullptr);
+    if (World == nullptr)
+    {
+        LOGWARNING("World:ForEachLoadedChunk(): invalid world parameter");
+        L.LogStackTrace();
+        return 0;
+    }
+    cLuaState::cRef FnRef;
+    L.GetStackValues(2, FnRef);
+    if (!FnRef.IsValid())
+    {
+        return cManualBindings::lua_do_error(tolua_S, "Error in function call '#funcname#': Could not get function reference of parameter #2");
+    }
+
+    class cLuaCallback :public cChunkDataCallback
+    {
+    public:
+        cLuaCallback(cLuaState & a_LuaState, cLuaState::cRef & a_FnRef) :
+            m_LuaState(a_LuaState),
+            m_FnRef(a_FnRef)
+        {
+        }
+
+    private:
+        cLuaState & m_LuaState;
+        cLuaState::cRef & m_FnRef;
+
+        virtual bool Coords(int a_ChunkX, int a_ChunkZ) override
+        {
+            bool res = false;  // By default continue the enumeration
+            m_LuaState.Call(m_FnRef, a_ChunkX,a_ChunkZ, cLuaState::Return, res);
+            return res;
+        }
+    } Callback(L, FnRef);
+
+    // Call the enumeration:
+    bool res = World->ForEachLoadedChunk(Callback);
+
+    // Push the return value:
+    L.Push(res);
+    return 1;
+}
+
+
+
+
 
 static int tolua_cWorld_GetBlockInfo(lua_State * tolua_S)
 {
@@ -570,6 +637,7 @@ void cManualBindings::BindWorld(lua_State * tolua_S)
 			tolua_function(tolua_S, "ForEachEntityInChunk",      ForEachInChunk<cWorld, cEntity,        &cWorld::ForEachEntityInChunk>);
 			tolua_function(tolua_S, "ForEachFurnaceInChunk",     ForEachInChunk<cWorld, cFurnaceEntity, &cWorld::ForEachFurnaceInChunk>);
 			tolua_function(tolua_S, "ForEachPlayer",             ForEach<       cWorld, cPlayer,        &cWorld::ForEachPlayer>);
+            tolua_function(tolua_S, "ForEachLoadedChunk",        tolua_cWorld_ForEachLoadedChunk);
 			tolua_function(tolua_S, "GetBlockInfo",              tolua_cWorld_GetBlockInfo);
 			tolua_function(tolua_S, "GetBlockTypeMeta",          tolua_cWorld_GetBlockTypeMeta);
 			tolua_function(tolua_S, "GetSignLines",              tolua_cWorld_GetSignLines);
--- a/src/Bindings/ManualBindings_World.cpp
+++ b/src/Bindings/ManualBindings_World.cpp
@@ -113,7 +113,7 @@ static int tolua_cWorld_ChunkStay(lua_State * tolua_S)
 static int tolua_cWorld_ForEachLoadedChunk(lua_State * tolua_S)
 {
 	// Exported manually, because tolua doesn't support converting functions to functor types.
-	// Function signature: ForEachLoadedChunk(callback)
+	// Function signature: ForEachLoadedChunk(callback) -> bool
 
 	cLuaState L(tolua_S);
 	if (
@@ -146,7 +146,7 @@ static int tolua_cWorld_ForEachLoadedChunk(lua_State * tolua_S)
 	}
 
 	// Call the enumeration:
-	World->ForEachLoadedChunk(
+	bool ret = World->ForEachLoadedChunk(
 		[&L, &FnRef](int a_ChunkX, int a_ChunkZ) -> bool
 		{
 			bool res = false;  // By default continue the enumeration
@@ -155,7 +155,9 @@ static int tolua_cWorld_ForEachLoadedChunk(lua_State * tolua_S)
 		}
 	);
 
-	return 0;
+	// Push the return value:
+	L.Push(ret);
+	return 1;
 }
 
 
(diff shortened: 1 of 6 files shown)


worktycho:
@madmaxoft This bring up a point. Do we want to use std::function for future callbacks, which allows the use of lambda methods which are nicer to write, or do we want to keep the current practice of using custom functor classes?

src/ChunkMap.cpp
```diff
@@ -2593,6 +2593,30 @@ bool cChunkMap::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinCh
...
 
 
+bool cChunkMap::ForEachLoadedChunk(cChunkDataCallback & a_Callback)
+{
+    bool Result = true;
+    cCSLock Lock(m_CSLayers);
+    for (cChunkLayerList::const_iterator itr = m_Layers.begin(); itr != m_Layers.end(); ++itr) //iterate over ALL loaded layers
+    {
+        cChunkLayer* layer = *itr;
+        for (int x = 0; x < LAYER_SIZE; x++)
```
NiLSPACE: We always use brackets 
```cpp
for (;;)
{
}

if (true)
{
}

switch (true)
{
   case true:
   {
   }
}
```

src/World.h
```diff
@@ -416,6 +416,8 @@ class cWorld :
 	/** Calls the callback for each chunk in the coords specified (all cords are inclusive). Returns true if all chunks have been processed successfully */
 	virtual bool ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinChunkZ, int a_MaxChunkZ, cChunkDataCallback & a_Callback) override;
 
+    /** Calls the callback for each loaded chunk. Returns true if all chunks have been processed successfully */
+    virtual bool ForEachLoadedChunk(cChunkDataCallback & a_Callback);
```
NiLSPACE: Indentation is wrong here

NiLSPACE: In fact is wrong everywhere. Please use tabs in front of each line, and spaces afterwards.

warmist:
Sorry i'll fix it quick.

src/World.cpp
```diff
@@ -3025,6 +3025,14 @@ bool cWorld::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinChunk
...
 
 
+bool cWorld::ForEachLoadedChunk(cChunkDataCallback & a_Callback)
+{
+	return m_ChunkMap->ForEachLoadedChunk(a_Callback);
+}
+
+
+
+
```
bearbin: One too few newlines here.

src/World.h
```diff
@@ -416,6 +416,8 @@ class cWorld :
 	/** Calls the callback for each chunk in the coords specified (all cords are inclusive). Returns true if all chunks have been processed successfully */
 	virtual bool ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinChunkZ, int a_MaxChunkZ, cChunkDataCallback & a_Callback) override;
 
+	/** Calls the callback for each loaded chunk. Returns true if all chunks have been processed successfully */
+	virtual bool ForEachLoadedChunk(cChunkDataCallback & a_Callback);
```
bearbin: To keep this how it was before there should be a newline after this.

src/ChunkMap.cpp
```diff
@@ -2593,6 +2593,34 @@ bool cChunkMap::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinCh
...
+				}
+			}
+		}
+	}
+	return Result;
+}
+
+
+
+
```
bearbin: One too few newlines.

src/ChunkMap.cpp
```diff
@@ -2593,6 +2593,34 @@ bool cChunkMap::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinCh
 
 
 
+bool cChunkMap::ForEachLoadedChunk(cChunkDataCallback & a_Callback)
```
bearbin: What's the point of Result in this function? It just seems to be an alias for true to me.

warmist: I was not sure about this. At first i wanted to maintain same form like ForEachChunkInRect. However i'm not sure if it can fail? Is there logic to have it bool return?

src/Bindings/ManualBindings_World.cpp
```diff
@@ -110,6 +110,73 @@ static int tolua_cWorld_ChunkStay(lua_State * tolua_S)
 
 
 
+static int tolua_cWorld_ForEachLoadedChunk(lua_State * tolua_S)
+{
+	// Exported manually, because i don't know what i'm doing...
```
bearbin: What's up with this comment?

src/ChunkMap.cpp
bearbin: What I was saying is that you have a variable (that implies there is some logic beind it) "Result" that is literally used for nothing apart from to store the value true.

src/Bindings/ManualBindings_World.cpp
warmist: Wanted a comment from someone who knows how it works: can it be exported automatically? Does it need to be exported automatically? Is manual export okay?

worktycho: It has to be exported manually because tolua doesn't support converting functions to functor types.

madmaxoft:
@worktycho I'd go with `std::function` in these cases, but what about things that need multiple callbacks? Those will still need to stay as interface-classes, is it worth having two different approaches to callbacks?

worktycho:
There's something to be said for going for multiple function objects in those cases, but how many of them are there?

src/ChunkMap.cpp
```diff
@@ -2593,6 +2593,33 @@ bool cChunkMap::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinCh
...
+{
+	cCSLock Lock(m_CSLayers);
+	for (cChunkLayerList::const_iterator itr = m_Layers.begin(); itr != m_Layers.end(); ++itr)  // iterate over ALL loaded layers
+	{
+		cChunkLayer * layer = *itr;
+		for (int x = 0; x < LAYER_SIZE; x++)
+		{
+			for (int z = 0; z < LAYER_SIZE; z++)
+			{
+				cChunkPtr p = layer->FindChunk(layer->GetX()*LAYER_SIZE + x, layer->GetZ()*LAYER_SIZE + z);  // A bit stupid to have local->global->local chain...
```
madmaxoft: Can't it be done without the `FindChunk` call?

src/ChunkMap.cpp
```diff
@@ -2593,6 +2593,33 @@ bool cChunkMap::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinCh
 
 
 
+void cChunkMap::ForEachLoadedChunk(std::function<bool(int, int)> a_Callback)
```
madmaxoft: All our `ForEach` functions return a bool, specifying if they enumerated everything, or the callback aborted. You should use it, too.

src/ChunkMap.cpp
```diff
@@ -2593,6 +2593,33 @@ bool cChunkMap::ForEachChunkInRect(int a_MinChunkX, int a_MaxChunkX, int a_MinCh
...
+	cCSLock Lock(m_CSLayers);
+	for (cChunkLayerList::const_iterator itr = m_Layers.begin(); itr != m_Layers.end(); ++itr)  // iterate over ALL loaded layers
+	{
+		cChunkLayer * layer = *itr;
+		for (int x = 0; x < LAYER_SIZE; x++)
+		{
+			for (int z = 0; z < LAYER_SIZE; z++)
+			{
+				cChunkPtr p = layer->FindChunk(layer->GetX()*LAYER_SIZE + x, layer->GetZ()*LAYER_SIZE + z);  // A bit stupid to have local->global->local chain...
+				if (p && p->IsValid())  // if chunk is loaded
```
madmaxoft: We compare pointers explicitly to `nullptr`. Only bools have the privilege of being used without comparison.
`if ((p != nullptr) && p->IsValid())`

src/ChunkMap.cpp
warmist: Not sure. Adding a function to look up a chunk in local coordinates just for this looks silly. Adding "naked code" for that would mean that if it ever changes there is chance that someone will forget to update this. So no really good way. 
About spaces around operators: can clang style checker check these? Or in general be more exact? It's really annoying when the checker misses them and then waste I other people time.

madmaxoft: Usually we do this by calling into an internal function of the `cChunkLayer` class. This way it stays encapsulated and isolated, yet keeps the raw performance. It is quite a lot of extra code, though.

NiLSPACE:
@worktycho Currently only the `cBlockLineTracer`, `cNetwork` and `cChunkstay` I believe.

madmaxoft:
It seems the build failure is the weird random google test fail. I restarted the affected builds.

You should rebase your code as a single commit on top of current master.

src/Bindings/ManualBindings_World.cpp
```diff
@@ -110,6 +110,59 @@ static int tolua_cWorld_ChunkStay(lua_State * tolua_S)
...
+	}
+
+	cPluginLua * Plugin = cManualBindings::GetLuaPlugin(tolua_S);
+	if (Plugin == nullptr)
+	{
+		return 0;
+	}
+
+	// Read the params:
+	cWorld * World = (cWorld *)tolua_tousertype(tolua_S, 1, nullptr);
```
madmaxoft: Use `reinterpret_cast`

NiLSPACE:
@worktycho and @madmaxoft Is it possible to use a `unordered_map<AString, std::function>` for callbacks with multiple functions?

worktycho:
It is, but I wouldn't recommend it. You'd be better off using a class with std::function members, at which point its debatable as to whether a strategy class is a better option.

tigerw:
What do you mean by multiple callbacks?
```cpp
void DoWith(int Thing, cCallback C1, cCallback C2, cCallback C3);
```
If so, why not:
```cpp
void DoWith(int Thing, std::function<> C1, std::function<> C2, std::function<> C3);
```
If not, nevermind.

NiLSPACE:
No, for example the [cLineBlockTracer callbacks](https://github.com/cuberite/cuberite/blob/master/src/BlockTracer.h#L28)
```cpp
class MyCallback : public cBlockTracer::cCallbacks
public:
   virtual bool OnNextBlock(int a_X, int a_Y, int a_Z, BLOCKTYPE a_BlockType NIBBLETYPE a_BlockMeta, char a_Face) override
   {
   }
...
```

tigerw:
Okay. So then, could one do this?

```cpp
cLineBlockTracer::Trace(std::function<bool(int, int, int, BLOCKTYPE, NIBBLETYPE)> OnNextBlock, std::function<bool(...)> OnOutOfWorld, ...)

...

Tracer.Trace(
   [](int, int, int, ...) { /* Next block */ };
   [](...) { /* Out of world */ };
   ...
);
```

SphinxC0re:
Any updates on this one?

worktycho:
I'm going to clean up this PR, and get it mergable.

worktycho:
Manually merged

PR #2248, opened 2015-06-15
Fixed override

worktycho:
This was causing a compile error in clang 3.6


Seadragon91:
Ty very much, hmm... . I had to fix a few other things, too. Small question do we really need the flag WError?

worktycho:
We need Werror because otherwise people ignore warnings. And most of the warnings are things you shouldn't be doing. What would be better would be if we could CI with 3.6 as well as 3.4, but we don't have the resources to do that quick enough.

bearbin:
Wow, appveyor finished first on this one 0.0

PR #2254, opened 2015-06-17
Implemented brewingstand

Seadragon91:
This implements brewing in cuberite.

There are a few things to be done, before this can be merged:
* Code needs to be reviewed, specially the `BrewingRecipe.cpp`


src/Chunk.cpp
```diff
@@ -2278,6 +2299,38 @@ bool cChunk::DoWithBeaconAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBeaconCal
 
 
 
+bool cChunk::DoWithBrewingstandAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBrewingstandCallback & a_Callback)
+{
```
worktycho: Why is this necessary? Can't you just use DoWithBlockEntityAt and do the the checks in the callback?

Seadragon91: I have done it like all other block entities, they all have a own method for the callback.

worktycho: Right, and this gives me the opportunity to bring up the fact that I don't think we should be doing that. It adds to much complexity, and redundant code to cChunk/cWorld and cChunkMap.

madmaxoft: I agree that it's become too much. However, I don't think that this PR is the right place to fix it. Let's accept it for now and refactor the entire thing later in a separate PR.
Just a note, the Lua API already has these and they're used a lot, so we'll need compatibility bindings, possibly templatized.

worktycho <notifications@github.com> wrote:

>> @@ -2278,6 +2299,38 @@ bool cChunk::DoWithBeaconAt(int a_BlockX, int
>a_BlockY, int a_BlockZ, cBeaconCal
>>  
>>  
>>  
>> +bool cChunk::DoWithBrewingstandAt(int a_BlockX, int a_BlockY, int
>a_BlockZ, cBrewingstandCallback & a_Callback)
>> +{
>
>Right, and this gives me the opportunity to bring up the fact that I
>don't think we should be doing that. It adds to much complexity, and
>redundant code to cChunk/cWorld and cChunkMap.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/cuberite/cuberite/pull/2254/files#r32614669

-- 

worktycho: I agree that refactoring the existing code is out of scope for this PR, but I think that we should use what ever new method we decide on for accessing brewing stands.

worktycho: I would like to use DoWithChunkAt, and GetBlockEntity. They provide enough info to get the blockentity.

Seadragon91:
The progress bars works now fine, the problem were in only sending an update very 10th ticks, had overlooked that.

Seadragon91:
Rebased the pull request. Can't do the part with the SlotArea.cpp, anyone better in it?

Seadragon91:
The slot handling is one of the difficult part `Window lost sync`...

Seadragon91:
I fixed the slot handling problem. The last thing now are the recipes. I first have to look and play around with it.

Seadragon91:
It's possible to use bit checks to check if the potion is a base, level 2 or extended.
http://minecraft.gamepedia.com/Potion#Data_values

My question, what do you think is better? Should I use bit checks and a file that contains the recipes, that can not be checked over that or should I create a file with all possible recipes.

Seadragon91:
I have created the brewing.txt file with the base drinkable recipes.

In `BrewingRecipe.cpp` I check if the ingredient is a gunpowder, then I create a new recipe and add it to the recipe list. If the potion is a splash potion, I search for the drinkable potion and create a new recipe and add it to the list too.

I think this needs to be reworked...

PR #2255, opened 2015-06-17
Loadable pieces

### LINKED ISSUE #2247: Add runtime loading possibility to cPrefabPiecePool
Make it possible to load pieces into `cPrefabPiecePool` at runtime. This will make it possible to make more configurable villages, trees and other structure generators.

Also, this way we can get rid of the hard-coded prefabs and instead use prefabs prepackaged with the server (but changeable by the admin).

madmaxoft:
Added support for loading pieces into the `cPrefabPiecePool` class from a cubeset file.

Fixes #2247.


madmaxoft:
The cubeset file format documentation, included in this PR, can be previewed at http://mc-server.xoft.cz/cubeset/cubeset%20file%20format.html

worktycho:
Build failed.

madmaxoft:
Yeah, and I'm not exactly sure why. These templates with rvalue references are still a bit over my head.

src/Generating/PrefabPiecePool.cpp
```diff
@@ -5,6 +5,81 @@
...
+		cPrefabPiecePool test;
+		auto res = test.LoadFromFile("test.cubeset", true);
+		if (!res)
+		{
+			LOGWARNING("Loading from file \"test.cubeset\" failed.");
+			return;
+		}
+		LOG("Loaded %u pieces and %u starting pieces", static_cast<unsigned>(test.m_AllPieces.size()), static_cast<unsigned>(test.m_StartingPieces.size()));
+	}
+} g_Test;
```
worktycho: This should be written as a separate test executable. I've made the global constructors warning an error for most files.

worktycho:
The problem is that because PushCallPop doesn't use perfect forwarding for a_Param, a copy is created before calling it. This is a problem because cRef isn't copyable. You need to change PushCallPop to use perfect forwarding.

SamJBarney:
I believe std::move/std::forward might help as well. Can't remember which at the moment.

worktycho:
Std::forward is relevant, because we want perfect forwarding.

madmaxoft:
I couldn't get it to work, the required syntax for these is terrible. I think I got it working in MSVC, so I'll force-push over this soon.

src/Generating/PrefabPiecePool.cpp
madmaxoft: In an ideal world, yes, it should be outside. In our world, it can't be because of dependencies.

worktycho: Then perhaps modification to improve testability could be included in this PR. Or are we giving up on tests completely?

worktycho:
Perfect forwarding is working, the only issue now is the self test.

src/Generating/PrefabPiecePool.cpp
madmaxoft: I'm giving up on external tests here. The loading requires LuaState and BlockArea, and those bring in most of the server with them - plugin manager, block handlers...

worktycho: BlockArea could be changed to use dependency injection for the block handlers, and it doesn't look like it has much else in the way of external decencies. The LuaState's plugin manager dependency can be eliminated by by moving Command Result out of PluginManager, and into its own enum class. I'll do that if you want.

madmaxoft: I think we should merge this and then play around with the tests. I'd like to continue the work on the actual loading, so that the generators use the loaded prefabs instead of hardcoded ones. If this is alright with you, then review the code as is and let's merge.

worktycho: Then I'm going to take the same position I took on the redstone PR. I'm not going to merge this with the tests as they are, but I won't object if someone else does. Also, to get this to pass Travis, you need to add a flag to disable the error in the CMakeList.

madmaxoft: I tried even more rigorously to externalize the test, I've got the linker not complain about anything, but then the test crashes because it's actually dependent on the `cBlockInfo` being available.

worktycho: Well, if you did anything to reduce coupling post that, otherwise we'll have to merge as is.

madmaxoft:
There, I managed to do it, the test is now external.

worktycho:
Merge if CI succeeds.

madmaxoft:
Wow, even @worktycho accepts my `Stubs.cpp`, didn't expect that. I really expected resistance.

Seadragon91:
Ci failed...

worktycho:
Stubs.cpp seems a good way to handle it. Not sure what the failure is.

Seadragon91:
AppVeyor failed too, in boths case there is a file missing `LuaState_Typedefs.inc`.
Forgot to do `git add .` @madmaxoft? :-)

madmaxoft:
That one is auto-generated, either the test is compiled too soon, or the path setup is wrong. I'll have a look tomorrow.

madmaxoft:
This is just so impossible to get right! :(

Seadragon91:
I tested your branch with clang 3.4/3.6 and it succeeds in boths.

madmaxoft:
I guess you didn't test with `cmake -DSELF_TEST`

madmaxoft:
I've been editing this so much I'll be seeing the git rebase in my dreams :P

madmaxoft:
Woo, finally it works. Anyone want to review this beast?

tests/LoadablePieces/LoadablePieces.cpp
```diff
@@ -0,0 +1,52 @@
...
+static int DoTest(void)
+{
+	cPrefabPiecePool test;
+	auto res = test.LoadFromFile("Test.cubeset", true);
+	if (!res)
+	{
+		LOGWARNING("Loading from file \"Test.cubeset\" failed.");
+		return 1;
+	}
+	LOG("Loaded %u regular pieces and %u starting pieces", static_cast<unsigned>(test.GetAllPiecesCount()), static_cast<unsigned>(test.GetStartingPiecesCount()));
```
worktycho: It might be worth putting an assert on these values, to check it loaded everything.

madmaxoft: Done.

worktycho:
Looks Good.

NiLSPACE:
I believe the Prefabs folder isn't provided when downloading from cuberite.org

madmaxoft:
Fixed, thanks for the heads up.

PR #2261, opened 2015-06-18
Grass Growth Fix

SamJBarney:
Grass now turns to dirt when light levels are below 5.
Grass will not spread to blocks where the light level is below 5.
Grass no longer grows under farmland.

### DIFF
--- a/src/Blocks/BlockDirt.h
+++ b/src/Blocks/BlockDirt.h
@@ -94,10 +94,9 @@ public:
 				continue;
 			}
 
-			BLOCKTYPE AboveDest;
-			NIBBLETYPE AboveMeta;
-			Chunk->GetBlockTypeMeta(BlockX, BlockY + 1, BlockZ, AboveDest, AboveMeta);
-			if (cBlockInfo::GetHandler(AboveDest)->CanDirtGrowGrass(AboveMeta))
+			NIBBLETYPE light = std::max(a_Chunk.GetBlockLight(BlockX, BlockY + 1, BlockZ), a_Chunk.GetTimeAlteredLight(a_Chunk.GetSkyLight(BlockX, BlockY + 1, BlockZ)));
+			// Grass does not spread to blocks with a light level less than 5
+			if (light > 4)
 			{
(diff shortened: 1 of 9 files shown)


src/Blocks/BlockDirt.h
```diff
@@ -96,10 +94,9 @@ class cBlockDirtHandler :
 				continue;
 			}
 
-			BLOCKTYPE AboveDest;
-			NIBBLETYPE AboveMeta;
-			Chunk->GetBlockTypeMeta(BlockX, BlockY + 1, BlockZ, AboveDest, AboveMeta);
-			if (cBlockInfo::GetHandler(AboveDest)->CanDirtGrowGrass(AboveMeta))
```
bearbin: Why can't we use CanDirtGrowGrass?

SamJBarney: Grass growth is based on light levels, and all CanDirtGrowGrass does is check whether the block is listed as transparent, or if it is a one-hit break type, such as ferns.
So that method was completely wrong in the first place.
By checking the light levels, we are ensuring that we are more Vanilla.

tigerw: Can the functionality of CanDirt... be changed?

SamJBarney: It could, if we passed it the light level, or had it use the chunk to calculate the light level.

Passing the light level wouldn't do any good, as all the method would do is wrap a check on the light level.

Passing the chunk would be inefficient as I would have to make 4 calls to find light levels instead of 3 (2 checking to see if grass can spread, 1 for checking if the current grass block can spread or if it should turn to dirt).

SamJBarney:
Can I get a Yes/No on this?

SamJBarney:
Anyone had a chance to review this?

worktycho:
could you update CanDirtGrowGrass to take a light level? Yes it just wraps a comparison, but it means the level compared against is only in one place.

SamJBarney:
CandDirtGrowGrass is a virtual function. I would have to put it in multiple places.

SamJBarney:
It really should just be stripped out.

worktycho:
That's just what I was about to say, can you remove it.

SamJBarney:
I guess I need to rebase my code and remove that reference that @madmaxoft added. https://github.com/cuberite/cuberite/blob/master/tests/LoadablePieces/Stubs.cpp#L196
Oh well. A new version should be up soon.

PR #2269, opened 2015-06-20
Added basic support for loading village prefabs from files.

madmaxoft:
The village generator can now be customized to generate villages from any `.cubeset` files. THe built-in villages are no longer hard-coded into cpp code, but rather exported as .cubeset files.

This may require a change in the packaging scripts on the buildserver to include the `MCServer/Prefabs` folder.


MCServer/Prefabs/Villages/AlchemistVillage.cubeset
```diff
@@ -0,0 +1,3647 @@
...
+
+
+Cubeset =
+{
+	Metadata =
+	{
+		CubesetFormatVersion = 1,
+		ExportDate = "2015-06-20 10:17:14",
+		["AllowedBiomes"] = "Desert, DesertM",
+		["IntendedUse"] = "Village",
```
NiLSPACE: Why are `AllowedBiomes` and `IntendedUse` enclosed in brackets, while `ExportDate` and `CubesetFormatVersion` aren't?

madmaxoft: This is exported by the GalExport plugin.
The first two values are always added by the plugin, so it knows that it doesn't need to quote the values. All the remaining values (`AllowedBiomes` and `IntendedUse` here, but anything is supported) are copied directly from the DB and there's no guarantee they won't need quoting.

NiLSPACE: Yeah, I know they are the same, but I just noticed that they were both used here ;)

madmaxoft:
@worktycho Do you have any hint why the build in clang is failing? It seems to complain about missing values in the `switch` statement, but there's a `default` case, so what's wrong this time?

madmaxoft:
If no-one's opposed, I'd like to merge this.

PR #2270, opened 2015-06-20
Button fixes

Haxi52:
Buttons no longer attempt to turn on if already on.
Buttons now play sound when turning off.

Also created cTaskPlaySound, for anyone trying to queue a sound to be played.

### DIFF
--- a/src/Blocks/BlockButton.h
+++ b/src/Blocks/BlockButton.h
@@ -2,6 +2,7 @@
 
 #include "BlockHandler.h"
 #include "Chunk.h"
+#include "World.h"
 #include "MetaRotator.h"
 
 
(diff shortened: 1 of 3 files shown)


src/World.h
```diff
@@ -142,6 +142,30 @@ class cWorld :
 		std::vector<Vector3i> m_SendQueue;
 	};
 
+	class cTaskPlaySound :
```
worktycho: cWorld does not seem the right place for this. If it really is generic it should be in a seperate file. However I think we should make cWorld::ScheduleTask support lambdas and then this would be a trivial task.

Haxi52:
Right, I put it with all the other cTask derivatives, though it does seem like World.h is getting quite large. I'll see what I can do about the lambdas.

worktycho:
Those cTask derivatives should be limited to those used by cWorld its self. But the current architecture is a hangover form before lambda's where available.

Haxi52:
Ok, I think I got it.
Created cTaskLambda, and added method ScheduleLambdaTask to cWorld.
Let me know what you think.

src/Blocks/BlockButton.h
```diff
@@ -61,22 +79,22 @@ class cBlockButtonHandler :
...
-			case BLOCK_FACE_ZP: return 0x3;
-			case BLOCK_FACE_XM: return 0x2;
-			case BLOCK_FACE_XP: return 0x1;
-			case BLOCK_FACE_YM: return 0x0;
-			case BLOCK_FACE_NONE:
-			{
-				ASSERT(!"Unhandled block face!");
-				return 0x0;
-			}
+		case BLOCK_FACE_YP: return 0x5;
```
Seadragon91: Please correct the indentation.

src/Blocks/BlockButton.h
```diff
@@ -61,22 +79,22 @@ class cBlockButtonHandler :
...
+		}
 		}
-		#if !defined(__clang__)
-			ASSERT(!"Unknown BLOCK_FACE");
-			return 0;
-		#endif
+#if !defined(__clang__)
+		ASSERT(!"Unknown BLOCK_FACE");
+		return 0;
+#endif
```
madmaxoft: Please don't change these

src/World.h
```diff
@@ -708,6 +725,9 @@ class cWorld :
 	/** Queues a task onto the tick thread, with the specified delay. */
 	void ScheduleTask(int a_DelayTicks, cTaskPtr a_Task);
 
+	/** Queues a task onto the tick thread, with the specified delay. */
+	void cWorld::ScheduleLambdaTask(int a_DelayTicks, std::function<void(cWorld&)> a_Func);
```
madmaxoft: This won't compile, remove the `cWorld::` part

worktycho: Also, there's no need to create a separate name, when an overload would do.

Haxi52: The indentation is visual studio formatting. I'll dig around the settings to see if I can disable it.

Ah, cWorld was left from me moving it around, It compiled fine for me, but I'll remove it.

I tried creating an overload to ScheduleTask, but the compiler complained the call was ambiguous, I'm too new to C++ to understand why the typedef cTaskPtr could be implicitly cast to std::function...

madmaxoft:
Before you go any further, rebase this code on top of current master and squash into a single commit.

Haxi52:
Ok, I've gone and completely mucked this up. Going to close this and start over.

PR #2274, opened 2015-06-21
Refactored block placement workflow.

### LINKED ISSUE #1700: Placing some blocks doesn't remove them from player inventory
I placed a chest in survival mode and it don't get removed from player inventory.

madmaxoft:
Multi-blocks can now use the default OnPlayerPlaced() callback in cItemHandler.

Fixes #1700.

### DIFF
--- a/src/Items/ItemBed.h
+++ b/src/Items/ItemBed.h
@@ -25,10 +25,11 @@ public:
 	}
 
 
-	virtual bool OnPlayerPlace(
+	virtual bool GetBlocksToPlace(
 		cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
 		int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
-		int a_CursorX, int a_CursorY, int a_CursorZ
+		int a_CursorX, int a_CursorY, int a_CursorZ,
+		sSetBlockVector & a_BlocksToPlace
 	) override
 	{
 		// Can only be placed on the floor:
@@ -36,12 +37,10 @@ public:
 		{
 			return false;
 		}
-		AddFaceDirection(a_BlockX, a_BlockY, a_BlockZ, a_BlockFace);
 
 		// The "foot" block:
-		sSetBlockVector blks;
 		NIBBLETYPE BlockMeta = cBlockBedHandler::RotationToMetaData(a_Player.GetYaw());
-		blks.emplace_back(a_BlockX, a_BlockY, a_BlockZ, E_BLOCK_BED, BlockMeta);
+		a_BlocksToPlace.emplace_back(a_BlockX, a_BlockY, a_BlockZ, E_BLOCK_BED, BlockMeta);
 		
 		// Check if there is empty space for the "head" block:
 		// (Vanilla only allows beds to be placed into air)
@@ -50,10 +49,8 @@ public:
 		{
 			return false;
 		}
-		blks.emplace_back(a_BlockX + Direction.x, a_BlockY, a_BlockZ + Direction.z, E_BLOCK_BED, BlockMeta | 0x08);
-
-		// Place both bed blocks:
-		return a_Player.PlaceBlocks(blks);
+		a_BlocksToPlace.emplace_back(a_BlockX + Direction.x, a_BlockY, a_BlockZ + Direction.z, E_BLOCK_BED, BlockMeta | 0x08);
+		return true;
 	}
 } ;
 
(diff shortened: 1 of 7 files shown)


src/Items/ItemBed.h
```diff
@@ -25,10 +25,11 @@ class cItemBedHandler :
 	}
 
 
-	virtual bool OnPlayerPlace(
+	virtual bool GetBlocksToPlace(
 		cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
 		int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
-		int a_CursorX, int a_CursorY, int a_CursorZ
+		int a_CursorX, int a_CursorY, int a_CursorZ,
+		sSetBlockVector & a_BlocksToPlace
```
tigerw: It doesn't matter, but just to note that BlocksToSet is used everywhere else.

src/Items/ItemChest.h
```diff
@@ -27,6 +27,8 @@ class cItemChestHandler:
 	}
 
 
+	/** We need an OnPlayerPlace override because we're processing neighbor chests and changing their metas,
```
tigerw: The sSetBlockVector in GetBlocksToPlace can't set neighbour chests?

madmaxoft: It uses `SetBlock` which deletes and re-creates the block entity, so it would destroy the chest contents.

src/Items/ItemHandler.cpp
```diff
@@ -411,6 +425,27 @@ bool cItemHandler::OnPlayerPlace(
...
+bool cItemHandler::GetBlocksToPlace(
+	cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
+	int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
+	int a_CursorX, int a_CursorY, int a_CursorZ,
+	sSetBlockVector & a_BlocksToSet
+)
+{
+	BLOCKTYPE BlockType;
+	NIBBLETYPE BlockMeta;
+	if (!GetPlacementBlockTypeMeta(&a_World, &a_Player, a_BlockX, a_BlockY, a_BlockZ, a_BlockFace, a_CursorX, a_CursorY, a_CursorZ, BlockType, BlockMeta))
```
tigerw: I notice that this isn't called in places where its overridden. Maybe it should be placed somewhere which makes sure that it is, like a non-overridable caller that calls all of the virtual functions?

src/Items/ItemMobHead.h
```diff
@@ -30,34 +32,36 @@ class cItemMobHeadHandler :
 		{
 			return true;
 		}
-		AddFaceDirection(a_BlockX, a_BlockY, a_BlockZ, a_BlockFace);
+		auto placedX = a_BlockX, placedY = a_BlockY, placedZ = a_BlockZ;
```
tigerw: thisCoding
styleLooks
reallyStrange;
:P

src/Items/ItemMobHead.h
```diff
@@ -30,34 +32,36 @@ class cItemMobHeadHandler :
...
-			if (TrySpawnWitherAround(a_World, a_Player, a_BlockX, a_BlockY, a_BlockZ))
+			if (TrySpawnWitherAround(a_World, a_Player, placedX, placedY, placedZ))
 			{
 				return true;
 			}
 			// Wither not created, proceed with regular head placement
 		}
 
-		return PlaceRegularHead(a_World, a_Player, a_EquippedItem, a_BlockX, a_BlockY, a_BlockZ, a_BlockFace);
+		cItem itemCopy(a_EquippedItem);
```
tigerw: Variable redundant?

src/Items/ItemChest.h
tigerw: How unfortunate.

src/Items/ItemMobHead.h
madmaxoft: Right you are, I made a copy because I knew I would need it, but then forgot to use it a few lines below.

src/Items/ItemHandler.cpp
madmaxoft: If `GetBlocksToPlace` is overriden, it already returns the correct blocks to place, so there isn't really a need to call `GetBlockPlacementTypeMeta` anymore.

tigerw:
How curious, my comment on the return value of HeadPlaced disappeared, yet it appears to have been invalidated anyway.

madmaxoft:
I suppose you made the comment directly in the commit; I amended the commit and force-pushed over this PR, so the commit itself disappeared.

PR #2282, opened 2015-06-23
Button fixes

Haxi52:
Buttons no longer click on when already on. Buttons now play sound when clicking off.

In addition added an overload to cWorld::ScheduleTask to accept lambdas.

### DIFF
--- a/src/Bindings/ManualBindings_World.cpp
+++ b/src/Bindings/ManualBindings_World.cpp
@@ -490,7 +490,7 @@ static int tolua_cWorld_ScheduleTask(lua_State * tolua_S)
 
 	auto task = std::make_shared<cLuaScheduledWorldTask>(*Plugin, FnRef);
 	Plugin->AddResettable(task);
-	World->ScheduleTask(DelayTicks, task);
+	World->ScheduleTask(DelayTicks, static_cast<cWorld::cTaskPtr>(task));
 	return 0;
 }
 
--- a/src/Blocks/BlockButton.h
+++ b/src/Blocks/BlockButton.h
@@ -19,15 +19,35 @@ public:
 	
 	virtual void OnUse(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cPlayer * a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ) override
 	{
+		NIBBLETYPE Meta = a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ);
+		
+		double x(a_BlockX);
+		double y(a_BlockY);
+		double z(a_BlockZ);
+
+		// If button is already on do nothing
+		if (Meta & 0x08)
+		{
+			return;
+		}
+
 		// Set p the ON bit to on
-		NIBBLETYPE Meta = (a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ) | 0x08);
+		Meta |= 0x08;
 
 		a_ChunkInterface.SetBlockMeta(a_BlockX, a_BlockY, a_BlockZ, Meta);
 		a_WorldInterface.WakeUpSimulators(a_BlockX, a_BlockY, a_BlockZ);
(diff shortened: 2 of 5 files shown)


src/Bindings/ManualBindings_World.cpp
```diff
@@ -490,7 +490,7 @@ static int tolua_cWorld_ScheduleTask(lua_State * tolua_S)
 
 	auto task = std::make_shared<cLuaScheduledWorldTask>(*Plugin, FnRef);
 	Plugin->AddResettable(task);
-	World->ScheduleTask(DelayTicks, task);
+	World->ScheduleTask(DelayTicks, (cWorld::cTaskPtr)task);
```
worktycho: We don't use c style casts

src/Blocks/BlockButton.h
```diff
@@ -19,15 +19,31 @@ class cBlockButtonHandler :
...
+		}
+
 		// Set p the ON bit to on
-		NIBBLETYPE Meta = (a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ) | 0x08);
+		Meta |= 0x08;
 
 		a_ChunkInterface.SetBlockMeta(a_BlockX, a_BlockY, a_BlockZ, Meta);
 		a_WorldInterface.WakeUpSimulators(a_BlockX, a_BlockY, a_BlockZ);
-		a_WorldInterface.GetBroadcastManager().BroadcastSoundEffect("random.click", (double)a_BlockX, (double)a_BlockY, (double)a_BlockZ, 0.5f, (Meta & 0x08) ? 0.6f : 0.5f);
+		a_WorldInterface.GetBroadcastManager().BroadcastSoundEffect("random.click", (double)a_BlockX, (double)a_BlockY, (double)a_BlockZ, 0.5f, 0.5f);
```
worktycho: please don't use c style cast in new code.

Haxi52: Ok, No problem,
Just copied from the line above.
I take it this is not something the lua script can pickup/report?

madmaxoft: It doesn't, and generally can't. We might be able to detect the simple cases such as these, but since they're still used in the legacy code, we'd need to clean everything up before making that change to the script.

Haxi52: On that note. This builds fine for me in VS, but the build server failed using clang.
Trying to see about getting clang to run on my machine, until then do I just need to add a destructor to cTaskLambda? I don't understand the error its throwing.

madmaxoft: Basically it's saying "you have defined a destructor, and an explicit constructor for a derived class, so it would quite make sense to have an explicit constructor for the base, too."
No idea where this idea came from or why it's an error, adding an empty constructor to `cTask` should fix it.

src/World.cpp
```diff
@@ -3051,7 +3051,11 @@ void cWorld::QueueTask(cTaskPtr a_Task)
 }
 
 
-
+void cWorld::ScheduleTask(int a_DelayTicks, std::function<void (cWorld&)> a_Func)
+{
+	cTaskLambda task = cTaskLambda(a_Func);
```
worktycho: This is causing the error with clang, because it indirectly invokes the implicitly auto-generated cTask copy constructor, which is depreciated because of the destructor. You can fix that by either eliminating the copy, by doing `cTaskLambda task(a_Func);` or adding `cTask (const cTask & other) = default;` to cTask.

src/Blocks/BlockButton.h
worktycho: I've commented on the line that caused the problem, with the options to fix it. As for the old casts, the clang is set to error for all files, except those which have old code with casts.

src/World.cpp
Haxi52: Ah ha! Makes sense. There's my C# background shining through.
Thanks for the help!

Haxi52: I did some more digging around, it looks like the actual implementation of SendSound uses an int, instead of double, and almost every upstream caller is casting int to double. It would make sense to refactor all the sound calls in between to accept int instead of double, but I didn't want to make this pull request about refactoring the sound system.

Haxi52:
I did some more digging around, it looks like the actual implementation of SendSound uses an int, instead of double, and almost every upstream caller is casting int to double. It would make sense to refactor all the sound calls in between to accept int instead of double, but I didn't want to make this pull request about refactoring the sound system.

madmaxoft:
You should rebase onto current master and squash into a single commit.

Haxi52:
Did I do this correctly?
Still sorta new to git in this respect.

worktycho:
Perfect.

PR #2284, opened 2015-06-23
Generate biomes when pregenerating heights through CompositedHeiGen.

### LINKED ISSUE #2283: Bad biomes generated under village when joining on far coords
If a client joins the server at a place there the world hasn't been generated (most likely due to server resettting world), and the server generates a village around the player, some of the chunks receive wrong biomes.

This is caused by village generator querying height, which in turn asks `cCompositedHeiGen` to generate composition for a chunk, but it doesn't provide the biomes, so the biomes field is uninitialized. This composition is then saved into the composition cache and is reused when generating the chunk for real.

madmaxoft:
Fixes #2283.

### DIFF
--- a/src/Generating/CompositedHeiGen.h
+++ b/src/Generating/CompositedHeiGen.h
@@ -20,7 +20,8 @@ class cCompositedHeiGen:
 	public cTerrainHeightGen
 {
 public:
-	cCompositedHeiGen(cTerrainShapeGenPtr a_ShapeGen, cTerrainCompositionGenPtr a_CompositionGen):
+	cCompositedHeiGen(cBiomeGenPtr a_BiomeGen, cTerrainShapeGenPtr a_ShapeGen, cTerrainCompositionGenPtr a_CompositionGen):
+		m_BiomeGen(a_BiomeGen),
 		m_ShapeGen(a_ShapeGen),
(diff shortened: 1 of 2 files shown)


src/Generating/CompositedHeiGen.h
```diff
@@ -28,18 +29,20 @@ class cCompositedHeiGen:
...
 
 
-	// cTerrainheightGen overrides:
+	// cTerrainHeightGen overrides:
 	virtual void GenHeightMap(int a_ChunkX, int a_ChunkZ, cChunkDef::HeightMap & a_HeightMap) override
 	{
 		cChunkDesc::Shape shape;
 		m_ShapeGen->GenShape(a_ChunkX, a_ChunkZ, shape);
 		cChunkDesc desc(a_ChunkX, a_ChunkZ);
+		m_BiomeGen->GenBiomes(a_ChunkX, a_ChunkZ, desc.GetBiomeMap());		// Need to initialize biomes for the composition gen
```
worktycho: There's a tab before the comment.

madmaxoft:
How did that happen? Fixed.

src/Generating/CompositedHeiGen.h
madmaxoft: How did that happen? Fixed.

Issue #2213, opened 2015-06-07
Segfault when melons are harvested


jammet:
Just discovered that when any player uses any tool to harvest melons in the current version of the server, the server segfaults immediately.

bearbin:
Can you give the stacktrace that was produced?

jammet:
Sadly no, it just did Sefault (in line 24741), and the number kept changing.
Interestingly, this happened every single time yesterday, and now I cannot reproduce it. I don't what it is happening here, to be honest.

bearbin:
Interesting. There was a crash related to pumpkins previously: https://github.com/mc-server/MCServer/issues/1970

But nobody actually got anywhere with solving it. Perhaps this issue is related.

What gamemode where you in, and if possible could you provide a video of the crash? (If not, try running the server under gdb

jammet:
When I encountered this yesterday, I tried destroying pumpkins too, and
also made sure to see if other blocks using orehandler made the server
crash, but those were alright. It was just melons, and the segfault never
gave me a trace. With all the other, earlier segfaults, I got a trace, but
with this, I'm just not getting any. If I could reproduce this, I would
have used gdb or strace to get more information already, but without having
the slightest clue how to get the crash, .... no can do.

On Mon, Jun 8, 2015 at 6:41 PM, Alexander Harkness <notifications@github.com

bearbin:
Was the melon you broke a naturally occurring melon, a placed one, or one that grew on a plant?

Did you always break the same melon, or did you try ones in different places?

jammet:
It's a plantation I created, with seeds I planted several months ago. Three
of us were connected when BlazeToaster wanted to harvest a melon below his
new construction. The server died, we all reconnected and nobody had an
idea what just happened. Seconds later, it happened again. BlazeToaster
mentioned he was trying to get a melon, so I teleported to him, and picked
one, randomly from another spot, and the server crashed. I remembered that
I did that, on that very same spot just hours earlier and everything was
fine. So I told everyone to disconnect and tried it as the only person
connected to the server. It crashed. Then we all tried it using with
various tools and just hands. It crashed. Always. I wrote the bug report,
went to sleep, connected 10 hours later and tried again, picking a random
melon.

The server didn't crash. The confusion is complete.


On Mon, Jun 8, 2015 at 6:50 PM, Alexander Harkness <notifications@github.com

SafwatHalaby:
[...]

PR #2288, opened 2015-06-23
Call Neighbour changed handlers when a melon is created

worktycho:
This is a partial fix for #2213, which should prevent the conditions which cause the crash from being created.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1035,6 +1035,9 @@ void cChunk::GrowMelonPumpkin(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_Bl
 				Meta
 			);
 			VERIFY(UnboundedRelFastSetBlock(a_RelX + x, a_RelY, a_RelZ + z, ProduceType, Meta));
+			auto Absolute = RelativeToAbsolute(Vector3i{a_RelX + x, a_RelY, a_RelZ + z}, m_PosX, m_PosZ);
+			cChunkInterface ChunkInterface(this->GetWorld()->GetChunkMap());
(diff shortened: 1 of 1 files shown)


src/Chunk.cpp
```diff
@@ -1035,6 +1035,9 @@ void cChunk::GrowMelonPumpkin(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_Bl
 				Meta
 			);
 			VERIFY(UnboundedRelFastSetBlock(a_RelX + x, a_RelY, a_RelZ + z, ProduceType, Meta));
+			auto Absloute = RelativeToAbsolute(Vector3i{a_RelX + x, a_RelY, a_RelZ + z}, m_PosX, m_PosZ);
```
madmaxoft: Fix the typo in the variable name, `Absolute`

worktycho: Not a typo, I just can't spell.

Seadragon91:
Tested the pull request, still the same problem.

madmaxoft:
This only prevents the situation from being created, but it doesn't fix the already created ones.

worktycho:
So if you create a new melon farm, it should no longer crash.

Seadragon91:
Yes, sorry shouldn't comment on pull request / issues in the earlier morning. I have tested it now, can destroy new created melon blocks without any problems.

worktycho:
Is this merge-able now?

Seadragon91:
The fix works fine. Any good idea to fix already corrupted worlds that have that problem?

worktycho:
In the server, not really. That involves some significant changes to how cascaded neighbor updates are handled. What might work is using worldedit to delete or convert to dirt all of the farmland under the melons in one go.

Seadragon91:
@worktycho Yes tested it on jammet's world, the replace command works.

PR #2290, opened 2015-06-25
Fixed Middle-Click Button in the ClickWindow-Packet.

### LINKED ISSUE #2260: Dropper Inventory Item Glitch
Some weird things only when in dropper inventory.
I'm middle mouse clicking to get a full stack.
When dropping in player inventory I lose the previously created stack.
If the created stack is halved (right click) I lose the half left in inventory.
Any stacks dropped into the dropper inventory are lost. (If created with middle or right clicks)
dropperitems

Here is the server output while doing the above clip. (everything after server time changed)
serveroutput-dropperglitch

bibo38:
See http://wiki.vg/Protocol#Click_Window
The MiddleClick has the button 2 and not 0.
Fixes #2260  

The middle mouse button is already implemented in `cSlotArea::MiddleClicked` (line 256), but was not read correctly from the packet.

### DIFF
--- a/src/Protocol/Protocol17x.cpp
+++ b/src/Protocol/Protocol17x.cpp
@@ -2374,7 +2374,7 @@ void cProtocol172::HandlePacketWindowClick(cByteBuffer & a_ByteBuffer)
 		case 0x0206: Action = caNumber7;         break;
 		case 0x0207: Action = caNumber8;         break;
 		case 0x0208: Action = caNumber9;         break;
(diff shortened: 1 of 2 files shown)


madmaxoft:
Looks good, I verified the protocol docs for both 1.8 and 1.7. Wonder why I got the constants wrong in the first place, there's no protocol version that would use `0x300`.

tigerw:
Remember when you kept changing VarInts to normal ints? :P

Issue #2259, opened 2015-06-18
Invisible Horses


SamJBarney:
I was  testing my code to kill grass if there is a block on top of it, and I saw a shadow moving around on the ground, not associated with any entity. So I right clicked it, and I mounted it, so I assume it is a horse.

I just tested using the spawn egg for a horse, and I was able to get one of them trapped. 
![2015-06-18_10 00 28](https://cloud.githubusercontent.com/assets/1827860/8235750/f6533fe8-15a0-11e5-8c61-557ddcbeb682.png)

madmaxoft:
Save the world and upload it somewhere. Or doesn't it survive world unloading?
I suppose this is caused by some horse metadata being sent wrong, or an off-limits value, so the client has no idea how to render such a horse.

SamJBarney:
Here it is: https://www.dropbox.com/s/u6p888nbwlmr6ab/world.zip?dl=0

Seadragon91:
Why not invisible creepers? hehehe

SamJBarney:
At least we know the client can handle invalid metadata when it comes to horses :+1: 
Makes me wonder if you could do that with other entities as well :)

jan64:
I like the invisible creeper idea, very much.

jammet:
We already had invisible creepers. I "saw" them not, in January. It was hilarious. XD Not.

Anyway. Yesterday -- the invisible horse happened to me. I used spawn eggs to repopulate the world. Of of them .. simply was completely invisible, except for the shadow. I gave it a quick, merciful death.

SamJBarney:
![Feature Bug]
(http://cdn.meme.am/instances/60835984.jpg)

tigerw:
A perfectly valid way to deal with issues.

bibo38:
Should now not be possible to spawn a invisible horse.
But maybe this could be "feature", so existing worlds won't be fixed.
I think this Bug can be closed.

Seadragon91:
@bibo38 It's possible to close the issue trough a pull request if merged. Here a list of [keywords](https://help.github.com/articles/closing-issues-via-commit-messages/).

bibo38:
@Seadragon91 Thank you. Nice thing.

PR #2294, opened 2015-06-26
Now Horses spawn with correct data values.

bibo38:
The max-value of style is 4, of color is 6.
See http://wiki.vg/Entities#Horse
Also based on the if, the max value of HorseType is 7.
Bugfix #2259 

This doesn't fix wrong values in existing Worlds, because invisible horses could be a "feature", so I didn't checked the values on Horse-Creation.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -1073,9 +1073,9 @@ cMonster * cMonster::NewMonsterFromType(eMonsterType a_MobType)
 		case mtHorse:
 		{
 			// Horses take a type (species), a colour, and a style (dots, stripes, etc.)
-			int HorseType = Random.NextInt(7);
-			int HorseColor = Random.NextInt(6);
-			int HorseStyle = Random.NextInt(6);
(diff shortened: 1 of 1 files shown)


src/Mobs/Monster.cpp
```diff
@@ -1073,9 +1073,9 @@ cMonster * cMonster::NewMonsterFromType(eMonsterType a_MobType)
 		case mtHorse:
 		{
 			// Horses take a type (species), a colour, and a style (dots, stripes, etc.)
-			int HorseType = Random.NextInt(7);
-			int HorseColor = Random.NextInt(6);
-			int HorseStyle = Random.NextInt(6);
+			int HorseType = Random.NextInt(8);
```
worktycho: Type only appears to go up to 4?

bibo38: Yeah and the types 5, 6, 7 are used to increase spawning rate of the normal horses.
In the next lines is an if, which maps the placeholder values to 0(normal horse).

PR #2295, opened 2015-06-26
Added Rabbit Metadata values.

### LINKED ISSUE #1867: Only brown rabbits spawn
When spawning rabbits, only brown colored rabbits spawn, instead of rabbits in different colors and patterns.

bibo38:
Added the Rabbit Types and the MoreCarrotTicks value. Types are selected randomly on creation.
Fixes #1867

### DIFF
--- a/src/Mobs/Rabbit.cpp
+++ b/src/Mobs/Rabbit.cpp
@@ -10,7 +10,20 @@
 
 
 cRabbit::cRabbit(void) :
-	super("Rabbit", mtRabbit, "mob.rabbit.idle", "mob.rabbit.death", 0.82, 0.68)
+	cRabbit(static_cast<eRabbitType>(cFastRandom().NextInt(
+		static_cast<UInt8>(eRabbitType::SaltAndPepper) + 1  // Max possible Rabbit-Type
+	)), 0)
+{
+}
+
+
+
+
+
+cRabbit::cRabbit(eRabbitType Type, int MoreCarrotTicks) :
+	super("Rabbit", mtRabbit, "mob.rabbit.idle", "mob.rabbit.death", 0.82, 0.68),
+	m_Type(Type),
+	m_MoreCarrotTicks(MoreCarrotTicks)
 {
 }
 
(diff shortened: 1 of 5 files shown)


src/Mobs/Rabbit.cpp
```diff
@@ -9,8 +9,27 @@
 
 
 
+static cFastRandom Random;
```
worktycho: Are you sure this is safe? The constructors can be called from plugins, and cFastRandom is not thread safe.

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -610,6 +610,13 @@ void cNBTChunkSerializer::AddMonsterEntity(cMonster * a_Monster)
 				m_Writer.AddByte("IsConverting", (((const cZombie *)a_Monster)->IsConverting() ? 1 : 0));
 				break;
 			}
+			case mtRabbit:
+			{
+				const cRabbit & Rabbit = *((const cRabbit *) a_Monster);
```
worktycho: reinterpert_cast?

worktycho:
@madmaxoft Do you know what the problem with appveyor is? Its complaining about platform tools.

src/Mobs/Rabbit.cpp
NiLSPACE: cFastRandom isn't exported to the API.

worktycho: No, I meant cRabbit's constructors. They can be called through cWorld::SpawnMob

NiLSPACE: Oh, right, I didn't think of that.

src/WorldStorage/NBTChunkSerializer.cpp
bibo38: I wanted to keep consistent with the surrounding code. None of them uses reinterpretet casts.

worktycho: Then change the surrounding code. We are moving over from c style casts, its just there are a lot of them to onvert.

SphinxC0re:
Any updates on this?

bibo38:
If everything is ok, I would squash the PR into a single commit. Then it should be able getting merged.

PR #2298, opened 2015-06-27
First attempt of annotations

worktycho:
An experiment with clang thread annotations.

Checks that cEntity::TIck is only called form a tick thread, and that all locks are unlocked.

Enable checking with -DCLANG_THREAD_ANALYSIS=1 to cmake
May not work in clang <3.6

Does anyone have any suggestions about additional checks that would be useful?


src/Bindings/PluginLua.cpp
```diff
@@ -72,12 +72,13 @@ void cPluginLua::Close(void)
...
 	std::swap(m_Resettables, resettables);
 	{
-		cCSUnlock Unlock(Lock);
+		m_CriticalSection.Unlock();
 		for (auto resettable: resettables)
 		{
 			resettable->Reset();
 		}
 		m_Resettables.clear();
+		m_CriticalSection.Lock();
```
madmaxoft: This will fail hard if anything between `Unlock()` and `Lock()` exits via exception. The RAII for unlocking is needed.

src/ChunkMap.cpp
```diff
@@ -1088,7 +1088,7 @@ int  cChunkMap::GetHeight(int a_BlockX, int a_BlockZ)
 		}
 
 		// The chunk is not valid, wait for it to become valid:
-		cCSUnlock Unlock(Lock);
+		m_CSLayers.Unlock();
 		m_evtChunkValid.Wait();
```
madmaxoft: This doesn't even re-lock the `Lock`, which is clearly broken.

madmaxoft:
I don't quite like the naming. `REQUIRES(WorldTick)` doesn't say much, could you use `REQUIRES_THREAD` instead?

src/Bindings/PluginLua.cpp
worktycho: Then either we can't have the annotations, or we need to use noexcept and try catch blocks as an alternatives. The annotations don't understand aliasing, so they can't tell that the inner reference is the same as the one passed in.

worktycho:
The complication is that it can require a role object or a lock. I just haven't used the REQUIRES(Lock) form yet. Do you think we should have two identical macros, REQUIRES_ROLE and REQUIRES_LOCK?

src/ChunkMap.cpp
worktycho: This is  weird bit of code, because what the original bit of code was doing was locking when Unlock was released, then immediately unlocked when Lock is released. I suspect the real answer is to not use cCSLock here.

src/Bindings/PluginLua.cpp
worktycho: Also, the way the servers set up at the moment, if an exception is triggered, the server is about to stop, so there's no point unlocking a lock that's about to be destroyed.

src/ChunkMap.cpp
worktycho: I've changed the code above to lock the mutex, rather than using cCSLock, so this code should now be fixed.

madmaxoft: No way, now you leave the function (line 1082 or 1087) with the chunkmap still closed. There's a reason we use RAII for locking, and we're not removing the RAII just to be able to use some weird annotations.

If anything, make the RAII Lock local to the needed block (make it go out of scope before reaching the `m_evtChunkValid.Wait()` line)

worktycho: That was stupid of me, and yes I've changed it to use an extra scope.

SphinxC0re:
@worktycho are you going to resume your work on this?

src/Bindings/PluginLua.cpp
worktycho: I'd like to try this again. There are two places in the server that use cCSUnlock around something other than a cEvent::Wait(). Would you mind if I had try catch blocks in three places to make this exception safe, and annotatible?

PR #2304, opened 2015-06-30
BlockArea: Fixed a crash with areas higher than chunk height.

### LINKED ISSUE #2302: Writing cBlockArea 256 blocks high triggers assert
Writing a `cBlockArea` that is 256 blocks high into the world will correctly report area origin problems in the console, but will then trigger an assert in Chunk.cpp (and most likely crash in release mode).

madmaxoft:
Fixes #2302.


worktycho:
Looks good, merge on CI success.

PR #2309, opened 2015-07-01
Fixed warnings generated by 64-bit MSVC.

madmaxoft:


src/Generating/BioGen.cpp
```diff
@@ -87,7 +87,7 @@ void cBioGenCache::GenBiomes(int a_ChunkX, int a_ChunkZ, cChunkDef::BiomeMap & a
 	if (((m_NumHits + m_NumMisses) % 1024) == 10)
 	{
 		LOGD("BioGenCache: %d hits, %d misses, saved %.2f %%", m_NumHits, m_NumMisses, 100.0 * m_NumHits / (m_NumHits + m_NumMisses));
-		LOGD("BioGenCache: Avg cache chain length: %.2f", (float)m_TotalChain / m_NumHits);
+		LOGD("BioGenCache: Avg cache chain length: %.2f", static_cast<double>(m_TotalChain) / m_NumHits);
```
worktycho: Can you remove the old style cast warning exception for BioGen.cpp from the CMakeList?

src/Protocol/Protocol18x.cpp
```diff
@@ -747,7 +747,7 @@ void cProtocol180::SendMapDecorators(int a_ID, const cMapDecoratorList & a_Decor
 	cPacketizer Pkt(*this, 0x34);
 	Pkt.WriteVarInt32(a_ID);
 	Pkt.WriteBEUInt8(m_Scale);
-	Pkt.WriteVarInt32(a_Decorators.size());
+	Pkt.WriteVarInt32(static_cast<UInt32>(a_Decorators.size()));
```
worktycho: Again can you remove the old-style-cast warning exception?

src/Generating/BioGen.cpp
madmaxoft: I don't think so, there are plenty of those casts still left.

worktycho: Not in BioGen.cpp. And the Werror exceptions are per file now, so we can turn off no-error=old-style-cast for BioGen.cpp individually.

PR #2312, opened 2015-07-02
Fixes grass dying at night

SamJBarney:
* Grass death is now only decided by whether the block above it is opaque
* Grass spread now requires the block above to be transparent

### DIFF
--- a/src/Blocks/BlockDirt.h
+++ b/src/Blocks/BlockDirt.h
@@ -3,6 +3,7 @@
 
 #include "BlockHandler.h"
 #include "../FastRandom.h"
+#include "../BlockInfo.h"
 #include "Root.h"
 #include "Bindings/PluginManager.h"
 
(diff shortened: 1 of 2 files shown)


src/Blocks/BlockDirt.h
```diff
@@ -48,13 +49,15 @@ class cBlockDirtHandler :
 		}
 		else if ((a_RelY < cChunkDef::Height - 1))
 		{
+			BLOCKTYPE above = a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ);
 			NIBBLETYPE light = std::max(a_Chunk.GetBlockLight(a_RelX, a_RelY + 1, a_RelZ), a_Chunk.GetTimeAlteredLight(a_Chunk.GetSkyLight(a_RelX, a_RelY + 1, a_RelZ)));
```
worktycho: Move this after the transparency check.

SamJBarney: Except I need it for the transparency check.

worktycho: Why? The variable isn't used until line 62. I was suggesting that you put it between the two ifs.

SamJBarney: It is used on line 55

PR #2317, opened 2015-07-06
Bonemeal Plant Growth Update

SamJBarney:
* Fixed Crops growing to maturity with only 1 bonemeal
* Trees now go through growth stages with bonemeal, as they do in Vanilla Minecraft

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -1455,9 +1455,12 @@ bool cWorld::GrowRipePlant(int a_BlockX, int a_BlockY, int a_BlockZ, bool a_IsBy
 			}
 			if (BlockMeta < 7)
 			{
-				if (!a_IsByBonemeal) {
+				if (!a_IsByBonemeal)
+				{
 					++BlockMeta;
-				} else {
+				}
+				else
+				{
 					BlockMeta += random.NextInt(4) + 2;
 					BlockMeta = std::min(BlockMeta, 7);
 				}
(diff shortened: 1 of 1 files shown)


SphinxC0re:
Test

src/World.cpp
```diff
@@ -1481,10 +1492,19 @@ bool cWorld::GrowRipePlant(int a_BlockX, int a_BlockY, int a_BlockZ, bool a_IsBy
...
 			{
-				FastSetBlock(a_BlockX, a_BlockY, a_BlockZ, BlockType, 7);
+				if (!a_IsByBonemeal)
+				{
+					++BlockMeta;
+				}
+				else
+				{
+					BlockMeta += random.NextInt(4) + 2;
+					BlockMeta = std::min(BlockMeta, (NIBBLETYPE)7);
```
worktycho: c style cast

src/World.cpp
```diff
@@ -1495,7 +1515,17 @@ bool cWorld::GrowRipePlant(int a_BlockX, int a_BlockY, int a_BlockZ, bool a_IsBy
...
-				FastSetBlock(a_BlockX, a_BlockY, a_BlockZ, BlockType, 7);
+
+				if (!a_IsByBonemeal)
+				{
+					++BlockMeta;
+				}
+				else
+				{
+					BlockMeta += random.NextInt(4) + 2;
+					BlockMeta = std::min(BlockMeta, (NIBBLETYPE)7);
```
worktycho: c style cast

SamJBarney:
It has been tested. The thing that broke it was I switched to using std::max instead of the ternary operator. It compiled fine on my machine, but travis required a cast to be there.
Also, I just had to move a few braces to match the style.

src/World.cpp
SamJBarney: Ok. I wasn't sure if I needed to use the c++ cast templates for basic types. I know std::reinterpret_cast is used for pointer types, but I was unsure about basic types such as ints and unsigned chars.

SamJBarney:
@worktycho  c-casts have been replaced

SphinxC0re:
Looks good. Could you squash the commits onto the current master?

SamJBarney:
Sure

Issue #1871, opened 2015-04-18
Issues with Chests


default0:
Chests and trapped chests cannot be opened if a non-solid block is on top of them (such as another chest, a lever, a trapdoor, a slab, a leaf, glass, etc).
Ender Chests can always be opened, even if a solid block is on top of them.

Further, placing down a chest or a trapped chest does not remove it from your inventory (it reappears the next time your inventory re-syncs).

nounoursheureux:
@madmaxoft @tigerw @SphinxC0re it is fixed now so the issue should be closed :+1:

PR #2318, opened 2015-07-06
(Ender) Chests are obstructed by opaque blocks.

nounoursheureux:
See #1871 
Trapped chests, normal chests and ender chests now open if there is a transparent block above, and don't if it is an opaque block.


madmaxoft:
I merged this manually after rebasing onto current master.

tigerw:
I still don't agree that chest's can open when glowstone or glass is above them, but not when stone is above them. Makes no sense.

NiLSPACE:
Glowstone is transparent. Vanilla allows chest to be opened when a transparent block is above it.

nounoursheureux:
> I still don't agree that chest's can open when glowstone or glass is above them, but not when stone is above them. Makes no sense.

I think the goal of Cuberite/MCServer is to have the same behaviour as the vanilla server. I understand that this makes no sense, but the vanilla server allows it so we should allow it. This is my opinion.

tigerw:
That seems to be the general consensus, le sigh.

PR #2330, opened 2015-07-09
Library update

tigerw:
* Updated luaexpat and expat
* Updated zlib
* Updated JsonCPP


SphinxC0re:
It works? Then go and squash them and rebase them onto the current master

tigerw:
It works. They're already rebased, but since I would say they're logically separate, would it be okay if they weren't squashed into one?

SphinxC0re:
You're right. Since they are indeed logically separate you can leave them as they are

.gitmodules
```diff
@@ -22,12 +22,21 @@
...
+	ignore = dirty
+[submodule "lib/jsoncpp"]
+	path = lib/jsoncpp
+	url = https://github.com/open-source-parsers/jsoncpp.git
+	ignore = dirty
 [submodule "lib/TCLAP"]
 	path = lib/TCLAP
 	url = https://github.com/mc-server/TCLAP.git
+	ignore = dirty
+
```
SphinxC0re: missing newline

Issue #2341, opened 2015-07-12
Build Failure with Xcode


hallucino:
Xcode 6.1.1
Build version 6A2008a

Still some errors due to Polarssl in:
src/Bindings
src/HTTPServer
src/Protocol
src/PolarSSL++

peterbell10:
This issue is rather vague but the cmake SYSTEM issue from the PR should be fixed.

PR #2342, opened 2015-07-12
Fix syntax errors with xcode

hallucino:
#2341


worktycho:
That flag is off globally for a reason. What should be handling those errors is the source file properties on a per file basis. Does xcode not support that?

SamJBarney:
@hallucino did you try to fix the old c-style casts in the codebase (not polarssl)? I've gone through and checked each one, and they are all fixable with the right kind of c++ cast.
Please go through and fix the c-style casts, minus the ones in polarssl, instead of adding in the extra flags.

SamJBarney:
@worktycho Xcode does not support that, unfortunately.

worktycho:
I can't find any documentation on internet about per file flags and xcode. Does cmake ignore the declarations completely, or does xcode ignore the flags?

hallucino:
@samjbarney: yes this error are related due to cast style in the polarssl code directly so I can't do anything.

I could change and add the no-style flag for particular files if you want but each time someone add a new file that depends on polarssl, the osx build will failed.

SphinxC0re:
For me it looks fine. Merge?

SamJBarney:
@hallucino You can fix the ones in the SslHttpConnection.cpp file, and the file in the binding file, without dropping in the flag for those.

worktycho:
@SphinxC0re I object to merge in the current state. If XCode absoluly cannot do per file flags, which I have seen no evidence of, then the additional flags should only be added in xcode not across all generators which use clang.

SamJBarney:
@SphinxC0re I agree with @worktycho . This should not be merged until it works with minimal addition of flags. Most of the errors can be fixed with the right c++ casts.

SphinxC0re:
Oh, I see. Then I see no point in merging this at any point, since it would be the right thing to change all c-style casts into cpp casts

hallucino:
For SslHttpConnection.cpp I got the errors:
src/../lib/polarssl/include/polarssl/cipher.h:443:16: error: use of old-style cast [-Werror,-Wold-style-cast]
        return (int) ctx->iv_size;
src/../lib/polarssl/include/polarssl/ssl.h:1990:30: error: use of old-style cast [-Werror,-Wold-style-cast]
    const unsigned char *A = (const unsigned char *) a;
src/../lib/polarssl/include/polarssl/ssl.h:1991:30: error: use of old-style cast [-Werror,-Wold-style-cast]
    const unsigned char *B = (const unsigned char *) b;

I could fix the errors in polarssl of course. Do you prefer this solution ?

SamJBarney:
@SphinxC0re Don't fix polarssl. Use the flag on that one since it is a library.

SphinxC0re:
@SamJBarney @hallucino Yeah, since it is an external library, that we don't maintain ourselves

worktycho:
I still want to know where the flags are disappearing. We pass the SYSTEM flag to the cmake include_directories, which should result in isystem being passed to clang. That should suppress the warnings in polarssl. Where is that not being passed through?

hallucino:
I changed my PR to only fix errors for particular files.

SamJBarney:
@hallucino I think you are misunderstanding what we are saying. You need to go into the source files and fix the c-style casts. If they cannot be fixed, such as a pointer to an object being cast to a char pointer, that is when you use the flags.

hallucino:
@SamJBarney nop I understood. But all these errors are in polarssl and previous you said that we can't do some fix in the lib.

hallucino:
s/previous/previously

SamJBarney:
@hallucino What about [src/Bindings/CMakeLists.txt](https://github.com/hallucino/cuberite/commit/c652b42b1f742babbd7d849f32c5b2e6bc4251ea#diff-9160665890b02ec2d5b5d73f518f339cL146)?
Are all those Polaris related? I'm pretty certain that LuaChunkStay.cpp has nothing to do with polaris.

hallucino:
@SamJBarney yes you re right for LuaChunkStay.cpp, but errors are related to the lib/lua:
src/Bindings/LuaChunkStay.cpp:43:18: error: use of old-style cast [-Werror,-Wold-style-cast]
        int NumChunks = luaL_getn(L, a_ChunkCoordTableStackPos);

src/../lib/lua/src/lauxlib.h:22:34: note: expanded from macro 'luaL_getn'
#define luaL_getn(L,i)          ((int)lua_objlen(L, i))

SamJBarney:
So don't use the macro.

tigerw:
Maybe an exception this time to the no-modify-libraries is okay? It seems unlikely that there's going to be the programmer-power needed to upgrade Lua anytime soon.

worktycho:
The warnings should not be firing. The SYSTEM flag for the includes should prevent that. That is what iwe need to be debugging. What is the command line that clang is actually executing?

SphinxC0re:
Any updates on this one?

madmaxoft:
I don't think this is even needed anymore, the old-style casts have been fixed (almost) everywhere.

worktycho:
Its the cmake SYSTEM issue that needs resolving. But again no mac to test on.

Issue #2329, opened 2015-07-09
In Generating/BioGen.cpp, convert max(min()) calls to Clamp()


SamJBarney:
Can be found on lines 664 and 645

PR #2347, opened 2015-07-13
Replace min-max combo with Clamp in BioGen

SamJBarney:
* Fixes https://github.com/cuberite/cuberite/issues/2329


SphinxC0re:
Nice catch!

PR #2348, opened 2015-07-13
Added CustomName and Health to sended Metadata of Monsters.

### LINKED ISSUE #2113: cMonster:SetCustomName not working correctly.
If I spawn a sheep and set the custom name of it. The code is called. Tested a name  tag and this works correctly.

TestPlugin
```lua
function Initialize(a_Plugin)
	a_Plugin:SetName("TestPlugin");
	a_Plugin:SetVersion(1);

    cPluginManager.BindCommand("/create", "testplugin.create", SpawnCustomMob, " - spawns a mob with at custom name")
    
	return true;
end

function SpawnCustomMob(a_Split, a_Player)
    local world = a_Player:GetWorld()
    local mobX = a_Player:GetPosition().x
    local mobY = a_Player:GetPosition().y
    local mobZ = a_Player:GetPosition().z
    local mob = world:SpawnMob(mobX, mobY, mobZ, mtSheep)
    
    if (mob == -1) then
        a_Player:SendMessage("Sheep couldn't be spawned.")
        return true
    end
    world:DoWithEntityByID(
        mob,
        function(a_Entity)
            if (a_Entity:IsMob()) then
                local sheep = tolua.cast(a_Entity, "cMonster")
                sheep:SetCustomName("MySheep")
                sheep:SetCustomNameAlwaysVisible(true)
            end
        end)
    
    a_Player:SendMessage("Sheep spawned with custom name")
    return true
end
```

bibo38:
Fixes #2113.

### DIFF
--- a/src/Protocol/Protocol17x.cpp
+++ b/src/Protocol/Protocol17x.cpp
@@ -3015,6 +3015,19 @@ void cProtocol172::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_En
 
 void cProtocol172::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob)
 {
+	// Living Enitiy Metadata
+	if (a_Mob.HasCustomName())
+	{
+		a_Pkt.WriteBEUInt8(0x8a);
+		a_Pkt.WriteString(a_Mob.GetCustomName());
+
+		a_Pkt.WriteBEUInt8(0x0b);
+		a_Pkt.WriteBool(a_Mob.IsCustomNameAlwaysVisible());
+	}
+
+	a_Pkt.WriteBEUInt8(0x66);
+	a_Pkt.WriteBEFloat(a_Mob.GetHealth());
+
 	switch (a_Mob.GetMobType())
 	{
 		case mtBat:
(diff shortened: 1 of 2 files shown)


src/Protocol/Protocol17x.cpp
```diff
@@ -3015,6 +3015,19 @@ void cProtocol172::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_En
 
 void cProtocol172::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob)
 {
+	// Living Enitiy Metadata
+	if(a_Mob.HasCustomName())
```
NiLSPACE: I believe you need to have a space after `if`. That's why the build is failing.

SamJBarney:
@bibo38 You're missing some spaces after 'if' statements.

src/Protocol/Protocol17x.cpp
```diff
@@ -3015,6 +3015,19 @@ void cProtocol172::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_En
...
+	{
+		a_Pkt.WriteBEUInt8(4 << 5 | 10);
```
tigerw: Eh, since everywhere else writes the hexadecimal computed value, do you think for consistency you could do it here too?

bibo38: I could do it. Or maybe change all Hex-Values to this format to allow easier verification.
Or maybe use a macro/inline function to make it easier to understand it.

bibo38:
Now it should be fine getting it merged.

PR #2349, opened 2015-07-13
Improved maps

tigerw:
Tada!


src/BlockID.h
```diff
@@ -62,7 +62,7 @@ enum ENUM_BLOCK_ID  // tolua_export
 	E_BLOCK_TORCH = 50,
 	E_BLOCK_FIRE = 51,
 	E_BLOCK_MOB_SPAWNER = 52,
-	E_BLOCK_WOODEN_STAIRS = 53,
+	E_BLOCK_OAK_WOOD_STAIRS = 53,
```
NiLSPACE: This could possibly break plugins.

src/BlockID.h
```diff
@@ -120,7 +120,7 @@ enum ENUM_BLOCK_ID  // tolua_export
 	E_BLOCK_PUMPKIN_STEM = 104,
 	E_BLOCK_MELON_STEM = 105,
 	E_BLOCK_VINES = 106,
-	E_BLOCK_FENCE_GATE = 107,
+	E_BLOCK_OAK_FENCE_GATE = 107,
```
NiLSPACE: Same here.

NiLSPACE:
The item renaming can break plugins. Perhaps add the old names at the end of the list with a comment telling they are deprecated?

tigerw:
Pfft, backwards compatibility is overrated.

Is there some way to tell plugins that they're using an outdated constant?

SamJBarney:
No there isn't.

NiLSPACE:
Yes there is, but you'll have to change the `__index` metamethod on the global environment (_G) ;)

tigerw:
Gentlemen, behold! An expert at work.

tigerw:
...how might I do this?

NiLSPACE:
I doubt you'd want to do that. By doing that you'd make a plugin call your function every time you'd try to get a global variable.

tigerw:
I added the aliases for the values I thought would be currently in use, so I omitted the (DOUBLE_)NEW_STONE_STONE_SLAB ones.

Mergeable?

NiLSPACE:
None of the GetMapBaseColourID functions are overridden. Is that intended?

tigerw:
Whoops, thanks.

Interesting: `override` isn't actually needed to override virtual functions but allows the compiler to verify that you've overridden correctly...

SamJBarney:
@tigerw the number that you are returning when getting the ColourID, can you turn them into an enum so we can see which color is actually being returned?

tigerw:
I thought about that, however, the problem is that, as can be seen from [here](http://minecraft.gamepedia.com/Map_item_format), the base colours are various shades of brown/grey/green/etc. that really have no good textual description. I would say that looking at the colours in the wiki is better than reading identifiers such as "light brown", "brown", and "dark brown".

SamJBarney:
I see. NVM.

tigerw:
Merge?

SphinxC0re:
Looks good. Maxim + merge :D

madmaxoft:
Why did you remove all the empty lines between functions? They make the code more readable by making the function headers stand out visually.

PR #2352, opened 2015-07-13
Fix duping doors

mmdk95:
No more door duplicating when door gets destroyed


mmdk95:
Oh rebase I reckon?

tigerw:
Try doing a `git rebase -i HEAD~2`, and then removing entirely the second commit from the editor that opens, and then save. That might fix it - if it does, then do a `git push -f`.

mmdk95:
Thanks

Issue #1903, opened 2015-04-26
Dyeing leather clothes is not working


jammet:
You cannoy dye leather tunics, helmets, boots or pants. I'm also not sure if this is done via crafting.txt. Minecraft doesn't call those "somecolor dyed leather someclothing", just "Leather tunic - dyed".

madmaxoft:
This might need a bit more research about how the dyed leather armor is actually stored. I don't think it's possible to represent this particular crafting operation by a crafting recipe (same as merging two half-broken items into a single repaired item cannot be reasonably done by a crafting recipe), it might be necessary to actually write this in a plugin or directly in the server C++ code.

tigerw:
I expect something similar to how fireworks are done, with a template recipe combined with internal code to handle NBT population.

SamJBarney:
[Algorithm](http://minecraft.gamepedia.com/Dyeing#Dyeing_armor)
[NBT Storage Example](http://minecraft.gamepedia.com/Player.dat_format#Display_Properties)

SamJBarney:
The recipe is a position-less recipe: you can place the leather armor and the dye wherever you want within the crafting grid.

jammet:
I don't know why or when this happened, but today the dye information
somehow got lost and my leather clothes reverted to their undyed look. I
fell to my death twice and picked the stuff up, but when I did this
intentionally to see if that was the reason, the dye was still there, so ...

 If this happens again and I find out why, I can post some info.

On Wed, Jul 15, 2015 at 10:31 PM, Julian Laubstein <notifications@github.com

PR #2353, opened 2015-07-14
Implements Dyed Leather Armor

SamJBarney:
* Fixes https://github.com/cuberite/cuberite/issues/1903

### DIFF
--- a/src/CraftingRecipes.cpp
+++ b/src/CraftingRecipes.cpp
@@ -787,6 +787,9 @@ cCraftingRecipes::cRecipe * cCraftingRecipes::MatchRecipe(const cItem * a_Crafti
 	// We use Recipe instead of a_Recipe because we want the wildcard ingredients' slot numbers as well, which was just added previously
 	HandleFireworks(a_CraftingGrid, Recipe.get(), a_GridStride, a_OffsetX, a_OffsetY);
 
+	// Handle Dyed Leather
+	HandleDyedLeather(a_CraftingGrid, Recipe.get(), a_GridStride, a_GridWidth, a_GridHeight);
+
 	return Recipe.release();
 }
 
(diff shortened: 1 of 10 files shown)


worktycho:
It might be clearer to use the std::max initalizer list overload, rather than calling max twice.

worktycho:
Also on item.cpp line 182, you cast to unsigned, then compare to minus one. The comparison will always be true.

SamJBarney:
@worktycho can you show me an example of using the initializer list with std::max?

tigerw:
Now you know the maxim...

SamJBarney:
All I need to do now is get the correct colors for the dyes, and this should be ready.

src/CraftingRecipes.cpp
```diff
@@ -874,3 +877,182 @@ void cCraftingRecipes::HandleFireworks(const cItem * a_CraftingGrid, cCraftingRe
...
+			{
+				int GridIdx = x + a_GridStride * y;
+				if ((a_CraftingGrid[GridIdx].m_ItemType == result_type) && (found == false))
+				{
+					found = true;
+					enchants = a_CraftingGrid[GridIdx].m_Enchantments;
+					name = a_CraftingGrid[GridIdx].m_CustomName;
+					lore = a_CraftingGrid[GridIdx].m_Lore;
+					damage = a_CraftingGrid[GridIdx].m_ItemDamage;
+					repair = a_CraftingGrid[GridIdx].m_RepairCost;
```
madmaxoft: This is not a good way. If there ever is an update that adds yet another thing to `cItem`, this will need updating. Why don't you instead make a copy of the entire `cItem` object.

SamJBarney: True. Let me fix that.

src/Item.cpp
```diff
@@ -172,6 +179,21 @@ void cItem::FromJson(const Json::Value & a_Value)
 		m_CustomName = a_Value.get("Name", "").asString();
 		m_Lore = a_Value.get("Lore", "").asString();
 
+		int red = a_Value.get("Color_Red", -1).asInt();
```
worktycho: Where does the JSON come from? We should probably check for a value outside the range 0-255.

src/Item.h
```diff
@@ -206,6 +222,7 @@ class cItem
 
 	int            m_RepairCost;
 	cFireworkItem  m_FireworkItem;
+	unsigned int m_ItemColor;
```
worktycho: I think if you're going to compress the colour into an int, you should use an object to abstract accessing the various fields.

src/Item.cpp
SamJBarney: Good point.

src/Item.h
SamJBarney: I took this approach because sending an item across the network will happen more frequently than changing an item color.  As it stands right now, I don't have to compress the color into an int every single time the item is sent to a client, and I can still alter the rgb values with minimal manipulation.

worktycho: And that's fine, I'm just saying that you should create a class to handle that.

SamJBarney: I think a class might be a little over the top for this.

worktycho: I just don't think it should be it cItem, because we will add more complex logic involving colors.

SphinxC0re:
Maxim? :smile:

SamJBarney:
I'm going to make one more change and go with what @worktycho suggested and move the color into its own class.

SamJBarney:
Any more changes?

worktycho:
Looks good. Squash and rebase?

SphinxC0re:
Or just the maxim :smile:

SamJBarney:
Done!

madmaxoft:
You've merged this without checking Appveyor, now there's a ton of MSVC warnings.

SphinxC0re:
Oh, sorry I did only check whether the CI failed or not

PR #2356, opened 2015-07-14
Fix food drain on movement.

### LINKED ISSUE #2044: Food drains too quickly.
Food appears to be mostly working, but it drains much more quickly than in the vanilla server.

LO1ZB:
Fixes #2044

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -16,6 +16,7 @@
 #include "../Items/ItemHandler.h"
 #include "../Vector3.h"
 #include "../FastRandom.h"
+#include <cmath>
 
 #include "../WorldStorage/StatSerializer.h"
 #include "../CompositeChat.h"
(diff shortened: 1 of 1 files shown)


src/Entities/Player.cpp
```diff
@@ -2178,20 +2179,18 @@ void cPlayer::ApplyFoodExhaustionFromMovement()
...
-	if (GetWorld()->GetWorldAge() % 2 != 0)
-	{
-		return;
-	}
-
 	// Calculate the distance travelled, update the last pos:
-	Vector3d Movement(GetPosition() - m_LastPos);
-	Movement.y = 0;  // Only take XZ movement into account
+	double SpeedX = abs(m_Speed.x);
+	double SpeedZ = abs(m_Speed.z);
```
worktycho: The abs is unnecessary.

SphinxC0re: Why? Is the speed always positive?

src/Entities/Player.cpp
```diff
@@ -2178,20 +2179,18 @@ void cPlayer::ApplyFoodExhaustionFromMovement()
...
-	Vector3d Movement(GetPosition() - m_LastPos);
-	Movement.y = 0;  // Only take XZ movement into account
+	double SpeedX = m_Speed.x;
+	double SpeedZ = m_Speed.z;
+	double BaseExhaustion(sqrt((SpeedX * SpeedX) + (SpeedZ * SpeedZ)));
 
 	// Apply the exhaustion based on distance travelled:
-	double BaseExhaustion = Movement.Length();
-	if (IsSprinting())
+	if (IsFlying() || IsClimbing())
```
SphinxC0re: Are you sure, that no exhaustion is applied, when climbing up a ladder?

src/Entities/Player.cpp
worktycho: no, but a number squared is, which is all you use it for.

src/Entities/Player.cpp
LO1ZB: yes http://minecraft.gamepedia.com/Hunger#Exhaustion_level_increase

PR #2362, opened 2015-07-16
Fixed redstone issue

scribblemaniac:
Fixed issue where glowstone blocks would block power transmission.

Glowstone is a transparent material according to the [wiki](http://minecraft.gamepedia.com/Glowstone#In_redstone_circuits) as well as in MC's source code (probably because it is a glass based block).

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator.cpp
@@ -645,7 +645,8 @@ void cIncrementalRedstoneSimulator::FindAndPowerBorderingWires(std::vector<std::
 
 		if (
 			(Neighbour->GetBlock(AdjustedPos) == E_BLOCK_REDSTONE_WIRE) &&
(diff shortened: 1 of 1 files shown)


tigerw:
This could potentially break other parts which depend on glowstone not being transparent, which it technically isn't. Since the simulator is undergoing refactoring anyway, do you think you could postpone this change, or add an explicit check for glowstone here?

scribblemaniac:
I'll just make it an explicit check for now I guess, but like I said before it *is a tranparent block* as far as code is concerned.

tigerw:
Mojang code is strange :P

scribblemaniac:
Yeah :P
My guess as to why is because both block breaking sounds and transparency are defined by the material type. Rather than making a new material type, they used  the glass material for the breaking sound. The transparency of the block was just a side effect of using that transparent glass material.

scribblemaniac:
Updated.

PR #2363, opened 2015-07-16
Improved spider AI

### LINKED ISSUE #2335: Spiders start attacking in daylight
"Spiders are hostile to the player and to iron golems as long as the light level immediately around them is 9 or less; otherwise they won't attack unless a player attacks first. Hostile spiders will continue to chase the player even if they are exposed to well-lit locations. If a spider sustains damage from a source other than the player, such as falling, its hostility will be reset." - http://minecraft.gamepedia.com/Spider

scribblemaniac:
Fixes #2335

### DIFF
--- a/src/Mobs/Spider.cpp
+++ b/src/Mobs/Spider.cpp
@@ -3,7 +3,8 @@
 
 #include "Spider.h"
 
-
+#include "../World.h"
+#include "../Entities/Player.h"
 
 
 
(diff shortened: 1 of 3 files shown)


src/Mobs/Spider.cpp
```diff
@@ -33,3 +34,37 @@ void cSpider::GetDrops(cItems & a_Drops, cEntity * a_Killer)
 
 
 
+
+void cSpider::EventSeePlayer(cEntity * a_Entity)
+{
+	if (!static_cast<cPlayer *>(a_Entity)->IsGameModeCreative() && (this->GetWorld()->GetBlockBlockLight(this->GetPosition()) <= 9))
```
tigerw: For consistency with the rest of the code, perhaps remove the `this` keyword?

I suspect that before you get the block light, you will also need to check cWorld::IsChunkLighted and call cWorld::QueueLightChunk and abort if it isn't. You would also have to get the current chunk position through cChunkDef::BlockToChunk(POSX_TOINT, ...).

scribblemaniac:
Implemented proposed changes and retested. Please review.

tigerw:
Completely forgot about GetChunkX/Z(), sorry.

scribblemaniac:
No problem, I got it anyway :wink:

Issue #2326, opened 2015-07-09
Support ageable mob


hallucino:
I have not found any opened issue to tha.

We could have some ageable mobs for:
pig
rabbit
sheep
villager
horse
ocelot
wolf

If nobody is working on that, I can send a pull request.

NiLSPACE:
Go ahead, I believe nobody is working on it :)

jan64:
Related: #2195

hallucino:
@NiLSPACE I will add the export of the following lua functions (cMonster):
GetAge
SetAge
IsBaby

Is it ok ?

If this modification + the usage of the HOOK_SPAWNING_MONSTER hook, we will be able to spawn baby mobs directly from lua.

SphinxC0re:
Yep :+1:

hallucino:
@SphinxC0re @NiLSPACE I don't know if the previous method is the best. I could also change SpawnMob function ? What do you think ?

NiLSPACE:
Changing the SpawnMob function should work fine, though you'll have to check if the type of monster actually has a baby version.

Sorry that I didn't respond to you, the internet on our vacation suddenly shutdown, which is obviously even worse then the 1300ms ping we had previously (and currently have back).

hallucino:
Ok I will close this CL, you can use baby mobs with Essentials:
https://github.com/mc-server/Essentials/issues/35

Have fun !

thx @NiLSPACE @SphinxC0re

PR #2364, opened 2015-07-16
Spawn baby mobs

hallucino:
#2326


NiLSPACE:
You could've saved yourself allot of time by giving the IsBaby parameter a default value of `false` ;) That would make plugins using the function work just fine as well.

hallucino:
like that ? I could remove the changement in others files or it is still more clear like that ?

NiLSPACE:
I suppose it makes sense to keep them :)

tigerw:
Why bother with optional values at all if they're filled in anyway?

scribblemaniac:
@tigerw Optional values are still a good idea so that any external plugins that use that function are not broken and still behave the same.

NiLSPACE:
@scottwillmoore Exactly :)

The documentation has to be updated though.

hallucino:
I must change something ?

NiLSPACE:
In the APIDump plugin the spawnmob function is still documented without the IsBaby parameter, You should add that with brackets around it since it's an optional parameter.

tigerw:
Apparently Lua ignores any extra parameters at the end that aren't filled though.

NiLSPACE:
Yes, that's true simply because it has a default value. Otherwise it would give an error because of too many parameters. This is simply because the cpp code that exports it checks for the amount of parameters.

hallucino:
Not sure about the doc for SpawnMob, if you have a better idea I will use it :+1:

hallucino:
Ok fixed. Sorry

SphinxC0re:
Looks good. Squash and rebase. When it compiles, merge.

PR #2365, opened 2015-07-16
Updated the Core submodule reference to the latest commit.

bibo38:
Without this, the Core module always fails on start and cannot be used.
Further reading: Pull Request mc-server/core#144.


hallucino:
I got:

Warning: [41344ba983db1729|10:33:29] LUA: Plugins/Core/give.lua:487: table index is nil
Warning: [41344ba983db1729|10:33:29] Stack trace:
Warning: [41344ba983db1729|10:33:29]   Plugins/Core/give.lua(487): IntializeItemBlacklist
Warning: [41344ba983db1729|10:33:29]   Plugins/Core/main.lua(84): (no name)
Warning: [41344ba983db1729|10:33:29] Stack trace end
Log: [41344ba983db1729|10:33:29] LUA: BreakIntoDebugger() not found / not a function
Warning: [41344ba983db1729|10:33:29] Error in plugin Core calling function Initialize()
Warning: [41344ba983db1729|10:33:29] Error in plugin Core: Cannot call the Initialize() function. Plugin is temporarily disabled.


Is it related ?

bibo38:
Yeah, exactly. The Enum-names have changed with 8656e149, e.g. E_BLOCK_DOUBLE_NEW_STONE_SLAB is now called E_BLOCK_DOUBLE_RED_SANDSTONE_SLAB.
In [line 487](https://github.com/scribblemaniac/Core/commit/0cc25e114e062188482c2cc144c0c3edbf62d427#diff-05389cd90cecaf846c5eb27526f20c5eL487) of the give script is the old enum name used. But this doesn't exists anymore and the initialisation crashes.

This has been fixed in the core plugin, but Cuberite did not updated the submodule commit, so that it checks out the broken version.
After updating it, the Core plugin doesn't crash anymore and all commands are available again.

hallucino:
Yes exactly, so the webadmin interface is broken too.

Issue #2195, opened 2015-06-02
Eggs spawn adult chickens.


SafwatHalaby:
Should be babies.

jan64:
This is a bit harder, because i don't think we have any support for spawning/creating baby chickens atm.

hallucino:
I can do it if nobody want to work on it.

NiLSPACE:
Go ahead, you're the first ^^

hallucino:
More generally spawn eggs don't work, right ?

hallucino:
oh no ok it works :)

SphinxC0re:
Isn't this solved

NiLSPACE:
Yup

PR #2366, opened 2015-07-17
Add ageable chicken + spawn baby chicken with the egg

hallucino:
#2195

### DIFF
--- a/src/Protocol/Protocol18x.cpp
+++ b/src/Protocol/Protocol18x.cpp
@@ -3336,6 +3336,14 @@ void cProtocol180::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob)
 			break;
 		}  // case mtBat
 
+		case mtChicken:
+		{
+			auto & Chicken = reinterpret_cast<const cChicken &>(a_Mob);
+			a_Pkt.WriteBEUInt8(0x0c);
+			a_Pkt.WriteBEInt8(Chicken.GetAge());
+			break;
+		}  // case mtChicken
+
 		case mtCreeper:
 		{
 			auto & Creeper = reinterpret_cast<const cCreeper &>(a_Mob);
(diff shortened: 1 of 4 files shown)


src/Protocol/Protocol18x.cpp
```diff
@@ -3336,6 +3336,14 @@ void cProtocol180::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob)
 			break;
 		}  // case mtBat
 
+		case mtChicken:
```
NiLSPACE: Is the 1.7 protocol updated as well?

SamJBarney: Just looked at his commits. There isn't any changes to the 1.7 protocol

src/Protocol/Protocol18x.cpp
```diff
@@ -3336,6 +3336,14 @@ void cProtocol180::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob)
...
+		{
+			auto & Chicken = reinterpret_cast<const cChicken &>(a_Mob);
+			a_Pkt.WriteBEUInt8(0x0c);
+			a_Pkt.WriteBEInt8(Chicken.GetAge());
```
madmaxoft: I think this is wrong. the age should be an int (see #2398) and the protocol only specifies a bool-like value, "IsChild"

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -531,6 +531,11 @@ void cNBTChunkSerializer::AddMonsterEntity(cMonster * a_Monster)
 				m_Writer.AddByte("BatFlags", reinterpret_cast<const cBat *>(a_Monster)->IsHanging());
 				break;
 			}
+			case mtChicken:
+			{
+				m_Writer.AddByte("Age", reinterpret_cast<const cChicken *>(a_Monster)->GetAge());
```
madmaxoft: Check the MC wiki if this should be a byte or an int.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2084,6 +2084,13 @@ void cWSSAnvil::LoadChickenFromNBT(cEntityList & a_Entities, const cParsedNBT &
 		return;
 	}
 
+	int AgeableIdx  = a_NBT.FindChildByName(a_TagIdx, "Age");
+	if (AgeableIdx > 0)
+	{
+		Byte Age = a_NBT.GetByte(AgeableIdx);
```
madmaxoft: byte or int?

SphinxC0re:
Any updates on this?

PR #2370, opened 2015-07-18
Implemented eating behaviour of "new" fish types

### LINKED ISSUE #250: Add new types of fish.
Pufferfish
Restores Half Hunger
If eaten, you will be given Poison IV for 1 minute, Hunger III and Nausea II for 15 seconds.
Used to brew a water breathing potion
Salmon
Can be cooked and eaten
Uncooked: Restores Hunger
Cooked: Restores 3 x Hunger
Clownfish
Usage still unknown besides being eaten and used to tame Ocelots
Restores Half Hunger

Ties in with #249

bibo38:
This fixes #250 
I slacked the ItemHandler, which now needs the ItemDamage to create a fitting ItemHandler.
Also all ItemHandler acceses are going through cItem.
Also I changed the signature of the GetEffect to AddEffects, to allow easily adding multiple effects. I don't know, if this is important for plugins. Otherwise a list will be needed.

### DIFF
--- a/src/Items/ItemFood.h
+++ b/src/Items/ItemFood.h
@@ -40,7 +40,8 @@ public:
 				if (m_ItemDamage == E_META_COOKED_FISH_SALMON)
 				{
 					return FoodInfo(6, 9.6);
-				} else
+				}
+				else
 				{
 					return FoodInfo(5, 6);
 				}
--- a/src/Items/ItemFood.h
+++ b/src/Items/ItemFood.h
@@ -13,8 +13,9 @@ class cItemFoodHandler :
 	typedef cItemHandler super;
 	
 public:
-	cItemFoodHandler(int a_ItemType)
-		: super(a_ItemType)
+	cItemFoodHandler(int a_ItemType, short a_ItemDamage = 0) :
+		super(a_ItemType),
+		m_ItemDamage(a_ItemDamage)
 	{
 	}
 
@@ -34,7 +35,16 @@ public:
 			case E_ITEM_BREAD:            return FoodInfo(5, 6);
 			// Carrots handled in ItemSeeds
 			case E_ITEM_COOKED_CHICKEN:   return FoodInfo(6, 7.2);
-			case E_ITEM_COOKED_FISH:      return FoodInfo(5, 6);  // TODO: Add other fish types
+			case E_ITEM_COOKED_FISH:
+			{
+				if (m_ItemDamage == E_META_COOKED_FISH_SALMON)
+				{
+					return FoodInfo(6, 9.6);
+				} else
+				{
+					return FoodInfo(5, 6);
+				}
+			}
 			case E_ITEM_COOKED_MUTTON:    return FoodInfo(6, 9.6);
 			case E_ITEM_COOKED_PORKCHOP:  return FoodInfo(8, 12.8);
 			case E_ITEM_COOKED_RABBIT:    return FoodInfo(5, 6);
@@ -49,7 +59,16 @@ public:
 			case E_ITEM_RED_APPLE:        return FoodInfo(4, 2.4);
 			case E_ITEM_RAW_BEEF:         return FoodInfo(3, 1.8);
 			case E_ITEM_RAW_CHICKEN:      return FoodInfo(2, 1.2);
-			case E_ITEM_RAW_FISH:         return FoodInfo(2, 1.2);
+			case E_ITEM_RAW_FISH:
+			{
+				switch (m_ItemDamage)
+				{
+					default:
+					case E_META_RAW_FISH_SALMON:     return FoodInfo(2, 0.4);
+					case E_META_RAW_FISH_CLOWNFISH:
+					case E_META_RAW_FISH_PUFFERFISH: return FoodInfo(1, 0.2);
+				}
+			}
 			case E_ITEM_RAW_MUTTON:       return FoodInfo(2, 1.2);
 			case E_ITEM_RAW_PORKCHOP:     return FoodInfo(3, 1.8);
 			case E_ITEM_RAW_RABBIT:       return FoodInfo(3, 1.8);
@@ -97,10 +116,19 @@ public:
 				a_Chance = 0.6f;
 				return true;
 			}
+			case E_ITEM_RAW_FISH:
+			{
+				if (m_ItemDamage == E_META_RAW_FISH_PUFFERFISH)
+				{
+					// TODO Add Pufferfish effects
+				}
+			}
 		}
 		return false;
 	}
 
+	protected:
+		short m_ItemDamage;
 };
 
 
--- a/src/Items/ItemFood.h
+++ b/src/Items/ItemFood.h
@@ -80,47 +80,46 @@ public:
 		return FoodInfo(0, 0.f);
 	}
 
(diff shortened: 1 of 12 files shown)


bibo38:
Can somebody explain me, whats wrong with the static variable?
>ItemHandler.cpp:66:87: error: 
>      declaration requires a global destructor [-Werror,-Wglobal-constructors]
>  ...E_ITEM_LAST + 1> cItemHandler::m_ItemHandler

Normally this is a warning, but why it is changed to an error?

src/Items/ItemFood.h
```diff
@@ -34,7 +35,16 @@ class cItemFoodHandler :
 			case E_ITEM_BREAD:            return FoodInfo(5, 6);
 			// Carrots handled in ItemSeeds
 			case E_ITEM_COOKED_CHICKEN:   return FoodInfo(6, 7.2);
-			case E_ITEM_COOKED_FISH:      return FoodInfo(5, 6);  // TODO: Add other fish types
+			case E_ITEM_COOKED_FISH:
+			{
+				if (m_ItemDamage == E_META_COOKED_FISH_SALMON)
+				{
+					return FoodInfo(6, 9.6);
+				} else
```
tigerw: woah there

src/Items/ItemHandler.cpp
```diff
@@ -63,16 +63,15 @@
 
 
 
-bool cItemHandler::m_HandlerInitialized = false;
-cItemHandler * cItemHandler::m_ItemHandler[2268];
+std::array<std::vector<std::unique_ptr<cItemHandler>>, E_ITEM_LAST + 1> cItemHandler::m_ItemHandler;
```
tigerw: I thought std::array is fixed-sized and defined at compile time? Was its use intended?

tigerw: Nevermind, I'm wrong on the second count.

src/Items/ItemFood.h
bibo38: whats the problem?

tigerw:
See [this StackOverflow question](http://stackoverflow.com/questions/15708411/how-to-deal-with-global-constructor-warning-in-clang).

src/Items/ItemFood.h
tigerw: Should be:
```
}
else
{
```

src/Items/ItemHandler.cpp
```diff
@@ -82,76 +81,77 @@ cItemHandler * cItemHandler::GetItemHandler(int a_ItemType)
...
 }
 
 
 
 
 
-cItemHandler * cItemHandler::CreateItemHandler(int a_ItemType)
+void cItemHandler::CreateItemHandler(int a_ItemType)
 {
+	std::vector<std::unique_ptr<cItemHandler>> &handler = m_ItemHandler[a_ItemType];
```
tigerw: Minor, but should be `>> & handler`;

src/Items/ItemHandler.cpp
tigerw: May I suggest that instead of creating a vector of pointers to item handlers to accommodate for metadata, damage values are simply passed to GetFoodInfo and whatever requires them?

src/Items/ItemHandler.h
```diff
@@ -154,18 +157,11 @@ class cItemHandler
...
-	static void Deinit();
-	
 protected:
 	int m_ItemType;
-	static cItemHandler * CreateItemHandler(int m_ItemType);
+	static void CreateItemHandler(int a_ItemType);
 
-	static cItemHandler * m_ItemHandler[E_ITEM_LAST + 1];
-	static bool m_HandlerInitialized;  // used to detect if the itemhandlers are initialized
+	static std::array<std::vector<std::unique_ptr<cItemHandler>>, E_ITEM_LAST + 1> m_ItemHandler;
```
worktycho: It would probably be useful to have a comment here about how this is basically a map of type -> { meta -> handler }.

src/Items/ItemHandler.cpp
bibo38: Yeah, I also thought of this. But I think both versions have their advantages/disadvantages.
The "problem" with passing the datavalues to the function is, that you have such constructs:
> Item.getHandle().EatFood(Item.m_Damage);

This looks silly in my eyes, so that you would need to implement all cItemHandler functions into cItem to remove passing already avaible data.

If the vectors seem a to high performance implication, we could use unique_ptr<cItemHandler[]> and implement something to track max ItemData values...
But I think the performance shouldn't get worse, because we do only allocate memory 1-4 times, when the ItemHandlers are created and pushed.

worktycho: @bibo38 I personally think that you're going down the right lines, as for all intents and purposes, fish and cooked fish are different items.

src/Items/ItemHandler.cpp
```diff
@@ -251,7 +262,24 @@ cItemHandler * cItemHandler::CreateItemHandler(int a_ItemType)
...
+			handler.push_back(cpp14::make_unique<cItemFoodHandler>(a_ItemType, E_META_COOKED_FISH_SALMON));
+			break;
+		}
+
+		case E_ITEM_RAW_FISH:
+		{
+			handler.push_back(cpp14::make_unique<cItemFoodHandler>(a_ItemType, E_META_RAW_FISH_FISH));
+			handler.push_back(cpp14::make_unique<cItemFoodHandler>(a_ItemType, E_META_RAW_FISH_SALMON));
+			handler.push_back(cpp14::make_unique<cItemFoodHandler>(a_ItemType, E_META_RAW_FISH_CLOWNFISH));
+			handler.push_back(cpp14::make_unique<cItemFoodHandler>(a_ItemType, E_META_RAW_FISH_PUFFERFISH));
```
bibo38: What do you think about subclassing the cItemFoodHandler for each fish, so that the Metavalue doesn't need to be passed and saved?
Or is it better to keep the number of files/classes low?

tigerw: In my opinion, subclassing is better than the array of mostly one item vectors, and the handler.push_back(); break; method.

bibo38: yeah, didn't mean it this way. I only meant subclassing to reduce the m_ItemDamage variable in cItemFoodHandler.
But how you want to solve the one item vectors thingy with subclassing?

bibo38:
So, it works.
Should I squash the PR or should I wait for some other reviews?

src/Items/ItemHandler.cpp
tigerw: I was thinking like cItemCookedFish, ItemCookedSalmon, ItemRawFish, etc.

Could you just do it with meta as an argument though? The de-facto method used everywhere is for the handlers not to store anything but block/item type (i.e. see `cBlockHandler::GetMapBaseColour(NIBBLETYPE a_Meta)`) and really, `Item.GetHandler().EatFood(Item.m_Damage);` is much better than all those `push_back`s and dealing unnecessarily with vector overhead.

If you want, have separate handlers for fish and cooked fish, as tycho has suggested, but I would recommend the above approach to metadata.

madmaxoft:
I don't think this is the right way to go. The item handlers should handle all the items with any damage value (or else there could be as many as 65536 (itemtype) * 65536 (damage) handlers). Instead, pass the damage values to any function that needs it.

tigerw:
Precisely.

SphinxC0re:
Any updates on this?

SphinxC0re:
Closing due to no response

PR #2371, opened 2015-07-18
Warning fixes

SamJBarney:
I got tired of the long list of warnings that kept popping up throughout the build process, so I went through and fixed most of them. Please mark things that you would like changed.

PR #2376, opened 2015-07-20
Fix FreeBSD build

mjhanninen:
Fix the FreeBSD build that was broken by a too strict compiler check. Downgraded the offending error to warning.

This PR may give a merge conflict with #2371 although there is no real conflict.


madmaxoft:
What was the exact error message? Wouldn't it be better to fix the actual offending source code?

mjhanninen:
Can't remember the exact error but the gist of it was that `T = unsigned int` specialization of `StringToInteger<T>` (in `StringUtils.h`) refused to compile failing at the following spot (see lines 183-186):

    if (std::numeric_limits<T>::min() / 10 > result)
    {
        return false;
    }

When `T == unsigned int` the condition `std::numeric_limit<T>::min() / 10 === 0 / 10 === 0 > result` is always `false` (and its negation tautologically true). This triggers the `-Wtautological-compare` on Clang 3.3 which in turn is promoted to an error due to the `-Werror` compilation flag.

Looking at the code around the problematic condition it seems that it never gets evaluated for unsigned types. So the compiler warning/error is spurious. In fact the latest Clang on OS X (which corresponds, I guess, to the version 3.6 of the community Clang) doesn't trip on this so maybe the compiler just got smarter.

So, I wouldn't regard the source code as offending in this particular case. However I could try to permute the code a little bit to make it go away that way.

PR #2379, opened 2015-07-21
Hopefully fixed food drain bugs

### LINKED ISSUE #2044: Food drains too quickly.

tigerw:
Tada! Fixes #2044, #2414.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -20,7 +20,6 @@
 #include "ChatColor.h"
 #include "Items/ItemHandler.h"
 #include "Blocks/BlockHandler.h"
-#include "Blocks/BlockSlab.h"
 #include "Blocks/BlockBed.h"
 #include "Blocks/ChunkInterface.h"
 #include "BlockInServerPluginInterface.h"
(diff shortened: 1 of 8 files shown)


tigerw:
Fancy new integration status is fancy.

tigerw:
mergey mergey?

src/ClientHandle.cpp
```diff
@@ -760,40 +759,31 @@ void cClientHandle::HandlePlayerPos(double a_PosX, double a_PosY, double a_PosZ,
 		return;
 	}
 
-	/*
-	// TODO: Invalid stance check
-	if ((a_PosY >= a_Stance) || (a_Stance > a_PosY + 1.65))
+	Vector3d NewPosition(a_PosX, a_PosY, a_PosZ);
+	Vector3d OldPosition = GetPlayer()->GetPosition();
+	auto PreviousIsOnGround = GetPlayer()->IsOnGround();
```
SphinxC0re: isn't this supposed to be a bool? Also do generics get evaluated at compile time?

madmaxoft: `auto` is better, we should use it whenever possible. Among others it makes maintenance easier, in case you're refactoring, changing the return type of a function, `auto` allows you to not care about the exact type.

src/Color.h
```diff
@@ -15,7 +15,7 @@ class cColor
 {
 public:
 
-	enum
+	enum : unsigned int
```
madmaxoft: Yeah, because it wouldn't be @tigerw if he didn't touch something unrelated :)

src/Entities/Player.cpp
```diff
@@ -474,61 +451,87 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
...
+				(
+					cBlockSlabHandler::IsAnySlabType(GetWorld()->GetBlock((GetPosition() + Vector3d(0, 0, 0)).Floor())) ||
+					cBlockSlabHandler::IsAnySlabType(GetWorld()->GetBlock((GetPosition() + Vector3d(HalfWidth, 0, 0)).Floor())) ||
+					cBlockSlabHandler::IsAnySlabType(GetWorld()->GetBlock((GetPosition() + Vector3d(-HalfWidth, 0, 0)).Floor())) ||
+					cBlockSlabHandler::IsAnySlabType(GetWorld()->GetBlock((GetPosition() + Vector3d(0, 0, HalfWidth)).Floor())) ||
+					cBlockSlabHandler::IsAnySlabType(GetWorld()->GetBlock((GetPosition() + Vector3d(0, 0, -HalfWidth)).Floor()))
+				)
+			) ||
+			(
+				(fmod(GetPosY(), 0.125) <= std::numeric_limits<double>::epsilon()) &&
```
madmaxoft: I think you're using `epsilon` for something it wasn't meant to be used for.
A hard-coded constant, `0.0001` or such, would have been better.

src/Entities/Player.cpp
```diff
@@ -474,61 +451,87 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
...
+		m_bTouchGround = false;
+	}
+
+	if (IsFlying() || IsSwimming() || IsClimbing())
+	{
+		m_LastGroundHeight = GetPosY();
 	}
+
+	UNUSED(a_bTouchGround);
+	/* Unfortunately whatever the reason, there are still desyncs in on-ground status between the client and server. Por ejemplo:
```
madmaxoft: Please use English only.

src/Entities/Player.cpp
SphinxC0re: But that's a *magic number*. We should at least define it somewhere.

madmaxoft: That's why I said "named" - name it whatever, `EPS`, I don't care, but make the value reasonable, it's definitely not a `std::numeric_limits::epsilon`

src/Color.h
tigerw: Tee hee

tigerw:
Now?

SphinxC0re:
Try to leave out the *Hopefully* from the CM

tigerw:
But then I shall be held responsible and burnt at the stake if it doesn't work! :P

SphinxC0re:
Yeah, that's the thing

SamJBarney:
@tigerw I'll provide the stake if it doesn't work, steak if it was properly tested and does :P

SphinxC0re:
:clap:

tigerw:
Let's see then:

PR #2380, opened 2015-07-21
Use container-based infrastructure

tigerw:
The problem here is that, since Travis container doesn't yet support [update-alternatives](https://github.com/travis-ci/travis-ci/issues/3668), `export CXX` will set the compiler to GCC 5 even if it's a Clang build.

But it works, amiright?


tigerw:
Travis fails due to Google-test.

worktycho:
How about only changing the CC and CXX cars if we are in a gcc build?

tigerw:
Good idea. Any idea why Google-test is throwing an exception, but only on one build?

tigerw:
Google test still randomly excepting for reasons unknown. Would this be mergeable?

madmaxoft:
I think the google test has something to do with LibEvent deinitialization being done at a wrong moment, but I have no idea on how to ascertain that, let alone fix it.

tigerw:
So merge?

madmaxoft:
Rebase onto current master (and get rid of your non-working google fix) to see if it actually compiles well.

tigerw:
Google was returning a 302 without the www prefix so that commit is still valid. However, since I'm in 'murica I can't do anything fancy :/

tigerw:
If you examine the logs, the failure is always during the test - builds start and run fine.

madmaxoft:
I've rebased onto current master and the builds fail.

madmaxoft:
Seems like `clang-3.7` is not available. Perhaps in another repository? Or use another version?

tigerw:
Still not sure which repository it's in though.

worktycho:
clang 3.7 doesn't exist yet. Testing against 3.4 is important for warnings. Is it possible to test against 3.4 and 3.6?

tigerw:
I was sure it managed to install 3.7 before, how curious.

Testing against both sounds like it would require further extensinos to the condition in CIBuild.sh - I guess a test for Clang and Clang version. If 3.7 has less warnings than 3.4 (amazing), how about I remain on 3.4?

worktycho:
3.7 is currently the trunk provisional version, so maybe you installed trunk?

tigerw:
yes!

tigerw:
I mean, yes it compiled! as opposed to yes I installed trunk!
I'm not sure what I installed.

madmaxoft:
I thought we wanted to use clang 3.6? Or am I just too old and forgetful?

madmaxoft:
Also, any chance of getting rid of that `gcc --version` in favor of `gcc-4.8 --version` in the gcc builds?

src/Bindings/LuaState.cpp
```diff
@@ -868,7 +868,7 @@ cLuaState::cStackValue cLuaState::WalkToValue(const AString & a_Name)
 		// Remove the previous value from the stack (keep only the new one):
 		lua_remove(m_LuaState, -2);
 	}  // for elem - path[]
-	return std::move(cStackValue(*this));
+	return cStackValue(*this);
```
madmaxoft: I'm not too confident about this change.

CIbuild.sh
```diff
@@ -6,6 +6,9 @@ export MCSERVER_BUILD_SERIES_NAME="Travis $CC $TRAVIS_MCSERVER_BUILD_TYPE"
 export MCSERVER_BUILD_ID=$TRAVIS_JOB_NUMBER
 export MCSERVER_BUILD_DATETIME=`date`
 
+if [ "$CXX" == "g++" ]; then
+	export CXX="/usr/bin/g++-4.8"
+fi
```
madmaxoft: This could use a comment saying why you're overriding the CXX variable. Also, shouldn't CC be overridden as well?

src/Bindings/LuaState.cpp
tigerw: Clang 3.7 said to remove std::move else it wouldn't compile.

CIbuild.sh
tigerw: Okay.
It works without CC or an upgrade to gcc-4.8, why bother :P

tigerw:
Tycho suggested testing with both clang 3.4 and 3.6, but I'm having difficulty getting a clang upgrade to work, so I just left it at 3.4 + gcc 4.8.

However, it would be nice to do something with clang since gcc 5 libraries don't appear to work with older clang versions.

src/Bindings/LuaState.cpp
tigerw: What's the difference?

tigerw:
I'm not precisely sure how to change the redirection of `gcc` to `gcc-4.8` without (sudo) update-alternatives. Any ideas?

src/Bindings/LuaState.cpp
madmaxoft: I think the original code used move-construction, while the new code could be using copy-construction (which isn't exactly supported by the `cStackValue` class)

tigerw: The compiler was complaining though. See first comment for error: https://news.ycombinator.com/item?id=9915300

```
    foo.cpp:44:25: error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]
        auto result = std::move(worker.run());

    foo.cpp:44:25: note: remove std::move call here
        auto result = std::move(worker.run());
```

SphinxC0re:
What about *Use container-based infrastructure with Travis*?

tigerw:
On [this](https://github.com/travis-ci/travis-ci/issues/3668) issue it is reported that they're looking into it.

tigerw:
oops, I made if "for" instead of "with". Is this okay?

tigerw:
I guess it is! meeeeeeeeerge --->

SetFlags.cmake
```diff
@@ -266,11 +266,18 @@ macro(set_exe_flags)
...
+				check_cxx_compiler_flag(-Wno-documentation-unknown-command HAS_NO_DOCUMENTATION_UNKNOWN)
 				if (HAS_NO_RESERVED_ID_MACRO)
 					# Use this flag to ignore error for a reserved macro problem in sqlite 3
 					add_flags_cxx("-Wno-reserved-id-macro")
 				endif()
+				if (HAS_NO_DOCUMENTATION_UNKNOWN)
+					# Ignore another problem in sqlite
+					add_flags_cxx("-Wno-documentation-unknown-command")
+				endif()
+				add_flags_cxx("-Wno-format-pedantic")
```
worktycho: I didn't even notice this. Why did you add it?

tigerw: Ref: https://github.com/paul99/v8m-rb/issues/128

3.7 builds failed without it. Clang complains tye %p specifier expects `void *` but that rarely happened.

worktycho: Are we supporting pre-release compilers now?

tigerw: yes :)

worktycho: Then can we define what we mean by c++? because technically clang is right that passing anything other than a `void *` to %p is UB.

tigerw: C++ is a programming language based off C, no?

If UggBoots are the case, then it is of course necessary to fix those instances. I shall inform #2441 to remove the flag after all.

.travis.yml
```diff
@@ -1,20 +1,17 @@
...
-
-  # lua, needed for style checking and possibly later on for bindings generation
-  - sudo apt-get install -qq lua5.1
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - g++-4.8
+    - lua5.1
```
worktycho: Another thing I've just noticed. Why are we installing lua, when we don't use the system lua in the build?

madmaxoft: We use Lua sources and headers, don't we?

worktycho: The ones we bundle. This asks for their to be a system lua, which we then don't use.

tigerw: All I did was convert the previous `apt-get install lua5.1` to this.

tigerw: Apparently I did remove a comment though: https://github.com/cuberite/cuberite/commit/e8c8aca7791a598244aa9dcc9a0992d2c249652d

worktycho: Ah, its used for the style check in the CI script.

PR #2385, opened 2015-07-24
Update CONTRIBUTING.md

Wolftein:
thisIsCamelCase ThisIsTitleCase.


madmaxoft:
Actually, camelcase can be used for both, and is such a common word that everyone knows it, unlike TitleCase. Paired with the actual visual representation, I think the current text is adequate and doesn't require a change.

/me votes NO merge.

SphinxC0re:
I agree. TitleCase is very uncommon

madmaxoft:
I would accept a change to UpperCamelCase, though - this would make things clear and still use the common term.

SphinxC0re:
https://en.wikipedia.org/wiki/CamelCase

SphinxC0re:
https://en.wikipedia.org/wiki/Letter_case#Title_case

PR #2400, opened 2015-07-28
Offload bad chunks

madmaxoft:
Chunks that fail to load are stored in separate files in a `badchunks` folder and regenerated.

### DIFF
--- a/src/OSSupport/File.cpp
+++ b/src/OSSupport/File.cpp
@@ -147,7 +147,7 @@ bool cFile::IsEOF(void) const
 
 
 
-int cFile::Read (void * iBuffer, size_t iNumBytes)
+int cFile::Read (void * a_Buffer, size_t a_NumBytes)
 {
 	ASSERT(IsOpen());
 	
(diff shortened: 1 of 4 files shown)


SphinxC0re:
This might collide with #2371

madmaxoft:
It will most likely create one or two merge conflicts, but those should be easy to fix.

src/OSSupport/File.cpp
```diff
@@ -172,7 +193,7 @@ int cFile::Write(const void * iBuffer, size_t iNumBytes)
 		return -1;
 	}
 
-	int res = (int)fwrite(iBuffer, 1, (size_t)iNumBytes, m_File);  // fwrite() returns the portion of Count parameter actually written, so we need to send iNumBytes as Count
+	int res = static_cast<int>(fwrite(a_Buffer, 1, a_NumBytes, m_File));  // fwrite() returns the portion of Count parameter actually written, so we need to send iNumBytes as Count
```
NiLSPACE: Comment on this line is wrong now. `iNumBytes` => `a_NumBytes`

madmaxoft: Good catch

madmaxoft:
Done, rebased, I think this is now ready for merging.

PR #2402, opened 2015-07-28
Compile.sh was outcommented, due to it being broken

RincewindsHat:
The compile.sh script produced an old build.
To prevent this I added a comment to that and made a link to the
new wiki, where a working tutorial is.
As soon as the script is fixed, this can be reverted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,9 @@ For Windows, you just need to download a file and extract it:
 
 For other operating systems you need to download and compile yourself. This can be done either manually, or with this automatic script:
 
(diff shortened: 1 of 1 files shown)


SphinxC0re:
I disagree. The builds are up-to-date. https://github.com/mc-server/MCServer is redirecting to cuberite. If noone is objecting this, I'm for closing

RincewindsHat:
ok with me, but than somehow old bugs get back into the build

worktycho:
The build server has been having some problems recently. So the script was probably fine, just pulling down older builds.

SphinxC0re:
Why would the *compile script* do that. It clones the repo and starts building

worktycho:
Because I can't read?

Issue #2377, opened 2015-07-20
Warning Fix TODO List


SamJBarney:
* https://github.com/cuberite/cuberite/pull/2371#discussion-diff-35019270R306
* https://github.com/cuberite/cuberite/pull/2371#discussion-diff-35019489R1332

PR #2410, opened 2015-07-30
Refactored code to use cLuaState

SamJBarney:
* Fixes https://github.com/cuberite/cuberite/issues/2377, as requested by @madmaxoft

### DIFF
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -281,32 +281,29 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
 	cLuaState LuaState(tolua_S);
 
 	#ifndef TOLUA_RELEASE
-	tolua_Error tolua_err;
 	if (
-		!tolua_isusertype (LuaState, 1, "cWorld", 0, &tolua_err) ||
-		!tolua_isnumber   (LuaState, 2, 0, &tolua_err) ||
-		!tolua_isnumber   (LuaState, 3, 0, &tolua_err) ||
-		!tolua_isnumber   (LuaState, 4, 0, &tolua_err) ||
-		!tolua_iscppstring(LuaState, 5, 0, &tolua_err) ||
-		!tolua_iscppstring(LuaState, 6, 0, &tolua_err) ||
-		!tolua_iscppstring(LuaState, 7, 0, &tolua_err) ||
-		!tolua_iscppstring(LuaState, 8, 0, &tolua_err) ||
-		!tolua_isusertype (LuaState, 9, "cPlayer", 1, &tolua_err) ||
-		!tolua_isnoobj    (LuaState, 10, &tolua_err)
+		!LuaState.CheckParamUserType(1, "cWorld") ||
+		!LuaState.CheckParamNumber(2, 4) ||
+		!LuaState.CheckParamString(5, 8) ||
+		!LuaState.CheckParamUserType(9, "cPlayer") ||
+		!LuaState.CheckParamEnd(10)
 		)
-		goto tolua_lerror;
+		return 0;
 	else
 	#endif
 	{
-		cWorld * self       = reinterpret_cast<cWorld *>(tolua_tousertype (LuaState, 1, nullptr));
-		cPlayer * Player    = reinterpret_cast<cPlayer*>(tolua_tousertype (LuaState, 9, nullptr));
-		int BlockX          = static_cast<int>(tolua_tonumber(LuaState, 2, 0));
-		int BlockY          = static_cast<int>(tolua_tonumber(LuaState, 3, 0));
-		int BlockZ          = static_cast<int>(tolua_tonumber(LuaState, 4, 0));
-		const AString Line1 = tolua_tocppstring(LuaState, 5, 0);
-		const AString Line2 = tolua_tocppstring(LuaState, 6, 0);
-		const AString Line3 = tolua_tocppstring(LuaState, 7, 0);
-		const AString Line4 = tolua_tocppstring(LuaState, 8, 0);
+		cWorld * self = nullptr;
+		cPlayer * Player = nullptr;
+		int BlockX = 0;
+		int BlockY = 0;
+		int BlockZ = 0;
+		AString Line1;
+		AString Line2;
+		AString Line3;
+		AString Line4;
+
+		LuaState.GetStackValues(1, self, BlockX, BlockY, BlockZ, Line1, Line2, Line3, Line4, Player);
+
 		#ifndef TOLUA_RELEASE
 		if (self == nullptr)
 		{
@@ -321,12 +318,6 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
 	LOGWARNING("Warning in function call 'UpdateSign': UpdateSign() is deprecated. Please use SetSignLines()");
 	LuaState.LogStackTrace(0);
 	return 1;
-	
-	#ifndef TOLUA_RELEASE
-tolua_lerror:
-	tolua_error(LuaState, "#ferror in function 'UpdateSign'.", &tolua_err);
-	return 0;
-	#endif
 }
 
 
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -1313,23 +1313,27 @@ static int tolua_SetObjectCallback(lua_State * tolua_S)
 
 static int tolua_cPluginLua_AddWebTab(lua_State * tolua_S)
 {
-	cPluginLua * self = reinterpret_cast<cPluginLua *>(tolua_tousertype(tolua_S, 1, nullptr));
+	cLuaState LuaState(tolua_S);
+	cPluginLua * self = nullptr;
 
+	if (!LuaState.GetStackValue(1, self))
+	{
+		LOGWARNING("cPluginLua:AddWebTab: invalid self as first argument");
+		return 0;
+	}
+	
 	tolua_Error tolua_err;
 	tolua_err.array = 0;
 	tolua_err.index = 3;
 	tolua_err.type = "function";
 
-	std::string Title = "";
+	std::string Title;
 	int Reference = LUA_REFNIL;
 
-	if (
-		tolua_isstring(tolua_S, 2, 0, &tolua_err) &&
-		lua_isfunction(tolua_S, 3)
-	)
+	if (LuaState.CheckParamString(2) && LuaState.CheckParamFunction(3))
 	{
 		Reference = luaL_ref(tolua_S, LUA_REGISTRYINDEX);
-		Title = (static_cast<std::string>(tolua_tocppstring(tolua_S, 2, 0)));
+		LuaState.GetStackValue(2, Title);
 	}
 	else
 	{


src/Bindings/DeprecatedBindings.cpp
```diff
@@ -293,20 +299,32 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
...
-		int BlockX          = static_cast<int>(tolua_tonumber(LuaState, 2, 0));
-		int BlockY          = static_cast<int>(tolua_tonumber(LuaState, 3, 0));
-		int BlockZ          = static_cast<int>(tolua_tonumber(LuaState, 4, 0));
-		const AString Line1 = tolua_tocppstring(LuaState, 5, 0);
-		const AString Line2 = tolua_tocppstring(LuaState, 6, 0);
-		const AString Line3 = tolua_tocppstring(LuaState, 7, 0);
-		const AString Line4 = tolua_tocppstring(LuaState, 8, 0);
+		cWorld * self = nullptr;
+		cPlayer * Player = nullptr;
+		int BlockX = 0;
```
NiLSPACE: Can't these be one line?
```cpp
int BlockX, BlockY, BlockZ
```
Or do we not accept that?

SamJBarney: I honestly don't know what the coding style is on that, but I do know they need to be initialized to 0.

SphinxC0re: I personally find this more readable

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -293,20 +299,32 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
...
+
+		LuaState.GetUserType(1, self, "cWorld");
+		LuaState.GetUserType(9, Player, "cPlayer");
+		LuaState.GetStackValue(2, BlockX);
+		LuaState.GetStackValue(3, BlockY);
+		LuaState.GetStackValue(4, BlockZ);
+		LuaState.GetStackValue(5, Line1);
+		LuaState.GetStackValue(6, Line2);
+		LuaState.GetStackValue(7, Line3);
+		LuaState.GetStackValue(8, Line4);
```
NiLSPACE: Perhaps we can create a cLuaState::GetStackValue**s** function
It would look like this:
```cpp
cLuaState::GetStackValues(int a_MinStackPos, int a_MaxStackPos, ...)
{
   va_list argList;
   va_start(argList, a_MaxStackPos);
   for (int i = a_MinStackPos; i <= a_MaxStackPos; i++)
   {
      // Is it possible to have a variable number of parameters as references? If so, then set them here.
   }
   va_end(argList)
}
```

SamJBarney: Either a var arg, or a variadic template

SphinxC0re: That's a different issue. We have to focus on, not cascade through problems

NiLSPACE: Well, otherwise it would require another refactor.

SphinxC0re: Yeah but that's waht I would prefer: Doing it layer by layer. Refactory should take a long time and many steps

SamJBarney:
Well, other than the debate about style and the creation of cLuaState::GetStackValues, does this look good for squash?

NiLSPACE:
@madmaxoft probably wants to take a look as well.

src/Bindings/DeprecatedBindings.cpp
madmaxoft: There already is a `cLuaState::GetStackValues()` function. Its first parameter is the staring index, and then it's just a list of variables which should be read. It's templated so that any number and any type of parameters are supported; it calls `GetStackValue` internally.

src/Bindings/DeprecatedBindings.cpp
madmaxoft: With initialization, use one line per variable. Without initialization (like the Line1, Line2 etc below), I prefer to merge them into a single line. But that's just personal opinion.

src/Bindings/LuaState.cpp
```diff
@@ -845,6 +843,21 @@ bool cLuaState::GetStackValue(int a_StackPos, float & a_ReturnedVal)
...
+bool cLuaState::GetStackValue(int a_StackPos, int & a_Value)
+{
+	double tmp;
+	if (GetStackValue(a_StackPos, tmp))
+	{
+		a_Value = FloorC(tmp);
+		return true;
+	}
+	return false;
+}
```
madmaxoft: This is not needed at all, there's already a templated function handling `int`s:
https://github.com/cuberite/cuberite/blob/master/src/Bindings/LuaState.h#L268-L287

src/Bindings/LuaState.h
```diff
@@ -263,6 +266,19 @@ class cLuaState
...
+	// If not, a_UserType remains unchanged.
+	template <typename T>
+	bool GetUserType(int a_StackPos, T * & a_UserType, const char * a_Type)
+	{
+		if (CheckParamUserType(a_StackPos, a_Type))
+		{
+			a_UserType = static_cast<T *>(tolua_tousertype(m_LuaState, a_StackPos, nullptr));
+		}
+		return false;
+	}
```
madmaxoft: This is not needed. Client code should first check all parameters, and then use `GetStackValue()` or `GetStackValues()`.

I'm considering a rewrite of the `CheckParamXYZ` handling so that it would better support overloaded functions, and for that I need the current usage idiom.

src/Bindings/ManualBindings.cpp
```diff
@@ -1313,41 +1313,48 @@ static int tolua_SetObjectCallback(lua_State * tolua_S)
...
+		else
 		{
-			luaL_unref(tolua_S, LUA_REGISTRYINDEX, Reference);
+			LOGWARNING("cPluginLua:AddWebTab: invalid function reference in 2nd argument (Title: \"%s\")", Title.c_str());
 		}
 	}
 	else
 	{
-		LOGWARNING("cPluginLua:AddWebTab: invalid function reference in 2nd argument (Title: \"%s\")", Title.c_str());
+		LOGWARNING("cPluginLua::AddWebTab: invalid self as first argument");
```
madmaxoft: Lua uses a single colon for class-function calls, so we report it that way.

src/Bindings/LuaState.h
SamJBarney: @madmaxoft I think that GetUserType is necessary, because there are no GetStackValue methods that would handle user types correctly.

How are you planning on rewriting the GetParamXYZ handling? Maybe I can update my code based on that.

madmaxoft: The usertype-based `GetStackValue` functions are generated into the $/src/Bindings/LuaState_Declaration.inc and _implementation.cpp files automatically during the build process.

SamJBarney:
Well, at least I know more about how the bindings are being generated. Though they shouldn't need to be generated at buildtime with a proper template method.

madmaxoft:
Templates can't turn class names into strings, which is needed for the custom types.

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -293,20 +299,32 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
 		!tolua_iscppstring(LuaState, 8, 0, &tolua_err) ||
 		!tolua_isusertype (LuaState, 9, "cPlayer", 1, &tolua_err) ||
 		!tolua_isnoobj    (LuaState, 10, &tolua_err)
+		*/
 		)
 		goto tolua_lerror;
```
madmaxoft: Return immediately, the `CheckParamXYZ()` functions already do output the error message so there's no need to do anything more.

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -293,20 +299,32 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
...
+		int BlockX = 0;
+		int BlockY = 0;
+		int BlockZ = 0;
+		AString Line1;
+		AString Line2;
+		AString Line3;
+		AString Line4;
+
+		LuaState.GetStackValue(1, self);
+		LuaState.GetStackValue(9, Player);
```
madmaxoft: Sort by stack index and use `GetStackValues`

src/Bindings/LuaState.h
```diff
@@ -35,6 +35,9 @@ extern "C"
 #include "PluginManager.h"
 #include "LuaState_Typedefs.inc"
 
+#undef TOLUA_TEMPLATE_BIND
+#include "tolua++/include/tolua++.h"
+
```
madmaxoft: Revert this change, move the inclusion back into the cpp file. We don't want to include the tolua library into each compilation unit including LuaState.h

src/Bindings/ManualBindings.cpp
```diff
@@ -1313,41 +1313,48 @@ static int tolua_SetObjectCallback(lua_State * tolua_S)
...
 {
-	cPluginLua * self = reinterpret_cast<cPluginLua *>(tolua_tousertype(tolua_S, 1, nullptr));
+	cLuaState LuaState(tolua_S);
+	cPluginLua * self = nullptr;
 
-	tolua_Error tolua_err;
-	tolua_err.array = 0;
-	tolua_err.index = 3;
-	tolua_err.type = "function";
+	if (LuaState.GetStackValue(1, self))
```
madmaxoft: Use opposite condition and return out of the function, to avoid nesting too deep.

SamJBarney:
@madmaxoft any more changes?

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -281,32 +281,32 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
...
+		int BlockZ = 0;
+		AString Line1;
+		AString Line2;
+		AString Line3;
+		AString Line4;
+
+		LuaState.GetStackValue(1, self);
+		LuaState.GetStackValues(2, BlockX, BlockY, BlockZ);
+		LuaState.GetStackValues(5, Line1, Line2, Line3, Line4);
+		LuaState.GetStackValue(9, Player);
```
madmaxoft: Why not just a single call to `GetStackValues()`? It does support variable length of varying parameters, you know :)
https://github.com/cuberite/cuberite/blob/master/src/Bindings/ManualBindings_World.cpp#L41

SamJBarney: I think it is just better semantically to get similar stack values with a single method call than all stack values with a single method call.

Plus, GetStackValues causes a bit of binary bloat. Not a lot, but some. And by grouping like I am, I can hopefully cut down on it a little bit.

madmaxoft: I think it's much better to use just one call, because it makes it easy to see the function's parameters, from the API standpoint this is excellent.

I don't think there's much difference between the amount of code generated for the two.

worktycho: There's some code bloat, though not much but it would be far more effective to optimize that explicitly if necessary, rather than on an ad-hoc basis.

src/Bindings/ManualBindings.cpp
```diff
@@ -1348,6 +1352,7 @@ static int tolua_cPluginLua_AddWebTab(lua_State * tolua_S)
 		LOGWARNING("cPluginLua:AddWebTab: invalid function reference in 2nd argument (Title: \"%s\")", Title.c_str());
 	}
 
+
```
madmaxoft: Get rid of this extra line when rebasing.

madmaxoft:
Looks good, rebase and squash (plus edit out the extra newline)

SamJBarney:
Squashed

madmaxoft:
I merged this manually, because I didn't like the commit message :P

It would have been much easier if you developed on a branch inside the main repo.

Issue #2404, opened 2015-07-28
Lilly pad doesn't get destroyed when block is placed under it


mmdk95:
Well title say it all

PR #2411, opened 2015-07-31
Fix for #2404

yo4you:
Fixes issue #2404: "Lilly pad doesn't get destroyed when block is placed under it"


madmaxoft:
I believe lilypads can be placed on non-stationary water as well. I'm not sure about non-source water, though.

yo4you:
This works for non stationary water, non source doesn't work (like it should).

madmaxoft:
There's no way to know whether the source in the screenshot is stationary or not. Breaking the tallgrass next to it should convert it to non-stationary, until the fluid simulator converts it back to stationary; I'm not sure if block checks on neighbors are performed with such a change, though.

There's nothing in the code that prevents lilypads on non-source water blocks; being unable to place them there may be a client-side check.

yo4you:
Updated the blocks around it and the lily pad seems to stay on until I remove the water.  Turning a source into a non-source does make it pop off. Also, removing a source next to another source makes both lilies pop off, I'm not sure if that's related to the water system.
(unrelated: Heh sorry, I have still to figure out how to make commands work in my debug world)

madmaxoft:
The neighbor source removal that pops off both lilypads is what I was referring to with stationary / nonstationary. Removing the neighboring source turns the source into a non-stationary block so that the fluid simulator checks it again, and that triggers the lilypad check. So that's why you need to check for both stationary and non-stationary source block.

yo4you:
I see, interesting. Changed the commit works fine now.

madmaxoft:
You seem to be doing unrelated changes on this branch. This way the PR will never get merged.

I'll try cherry-picking and squashing your commits relevant to the lilypad change.

yo4you:
Oh sorry my bad (didn't know that would affect his PR), those were for a different issue. The only file related is lilypad.hpp I'll put the other ones in a seperate PR once I'm done messing with it.

SphinxC0re:
Update?

madmaxoft:
I have cherry-picked the lilypad-related changes into a separate PR, and updated the changes to match vanilla behavior. Closing this PR.

PR #2415, opened 2015-08-01
Fixes #2413

### LINKED ISSUE #2413: Iron door can be opened with hand
iron door can be opened by right clicking

nichwall:
Iron doors can no longer be opened by right clicking. Buttons/redstone still work when I tested it. #2413

### DIFF
--- a/src/Blocks/BlockDoor.cpp
+++ b/src/Blocks/BlockDoor.cpp
@@ -61,11 +61,16 @@ void cBlockDoorHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterfac
 		case E_BLOCK_DARK_OAK_DOOR:
 		case E_BLOCK_JUNGLE_DOOR:
 		case E_BLOCK_SPRUCE_DOOR:
-		case E_BLOCK_IRON_DOOR:
 		case E_BLOCK_OAK_DOOR:
 		{
 			ChangeDoor(a_ChunkInterface, a_BlockX, a_BlockY, a_BlockZ);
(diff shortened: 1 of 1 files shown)


src/Blocks/BlockDoor.cpp
```diff
@@ -61,13 +61,18 @@ void cBlockDoorHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterfac
...
 		case E_BLOCK_SPRUCE_DOOR:
-		case E_BLOCK_IRON_DOOR:
 		case E_BLOCK_OAK_DOOR:
 		{
 			ChangeDoor(a_ChunkInterface, a_BlockX, a_BlockY, a_BlockZ);
 			a_Player->GetWorld()->BroadcastSoundParticleEffect(1003, a_BlockX, a_BlockY, a_BlockZ, 0, a_Player->GetClientHandle());
 			break;
 		}
+		// Prevent iron door from opening on player click
+		case E_BLOCK_IRON_DOOR:
```
tigerw: I don't think handling here is required as the client knows not to open an iron door.

NiLSPACE: I had the same thought.

nichwall:
Apparently not. I tried right clicking on both iron and wooden doors with Minecraft client 1.8.7, and it opened both of them. Could just be me, though

NiLSPACE:
Really? That's weird. I guess it has to stay then :)

nichwall:
Alrighty. When I added that, buttons and redstone were still working, so hopefully this doesn't break something else.

tigerw:
The client be buggy.

madmaxoft:
It's good to have this as an anti-cheat, in case a modified malicious client sent the wrong packet, anyway.

PR #2417, opened 2015-08-05
Fix compilation warnings related to implicit and old-style casts

mjhanninen:
The compilation output on OS X is rather noisy. This PR tackles some of this by fixing old-style casts and changing some implicit casts into explicit ones.

### DIFF
--- a/src/Bindings/LuaFunctions.h
+++ b/src/Bindings/LuaFunctions.h
@@ -2,11 +2,13 @@
 
 #include "Logger.h"
 #include <time.h>
+#include <chrono>
 // tolua_begin
 
 inline unsigned int GetTime()
 {
-	return (unsigned int)time(0);
+	// NB: For caveats, please see http://stackoverflow.com/a/14505248
+	return static_cast<unsigned int>(std::chrono::seconds(time(0)).count());
 }
 
 inline std::string GetChar( std::string & a_Str, unsigned int a_Idx)
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1668,7 +1668,7 @@ void cClientHandle::HandleRespawn(void)
 
 
 
-void cClientHandle::HandleKeepAlive(int a_KeepAliveID)
+void cClientHandle::HandleKeepAlive(UInt32 a_KeepAliveID)
 {
 	if (a_KeepAliveID == m_PingID)
 	{
--- a/src/Generating/DistortedHeightmap.cpp
+++ b/src/Generating/DistortedHeightmap.cpp
@@ -129,13 +129,13 @@ cDistortedHeightmap::cDistortedHeightmap(int a_Seed, cBiomeGenPtr a_BiomeGen) :
 	m_HeightGen(m_UnderlyingHeiGen, 64),
 	m_IsInitialized(false)
 {
-	m_NoiseDistortX.AddOctave((NOISE_DATATYPE)1,    (NOISE_DATATYPE)0.5);
-	m_NoiseDistortX.AddOctave((NOISE_DATATYPE)0.5,  (NOISE_DATATYPE)1);
-	m_NoiseDistortX.AddOctave((NOISE_DATATYPE)0.25, (NOISE_DATATYPE)2);
+	m_NoiseDistortX.AddOctave(static_cast<NOISE_DATATYPE>(1),    static_cast<NOISE_DATATYPE>(0.5));
+	m_NoiseDistortX.AddOctave(static_cast<NOISE_DATATYPE>(0.5),  static_cast<NOISE_DATATYPE>(1));
+	m_NoiseDistortX.AddOctave(static_cast<NOISE_DATATYPE>(0.25), static_cast<NOISE_DATATYPE>(2));
 
-	m_NoiseDistortZ.AddOctave((NOISE_DATATYPE)1,    (NOISE_DATATYPE)0.5);
-	m_NoiseDistortZ.AddOctave((NOISE_DATATYPE)0.5,  (NOISE_DATATYPE)1);
-	m_NoiseDistortZ.AddOctave((NOISE_DATATYPE)0.25, (NOISE_DATATYPE)2);
+	m_NoiseDistortZ.AddOctave(static_cast<NOISE_DATATYPE>(1),    static_cast<NOISE_DATATYPE>(0.5));
+	m_NoiseDistortZ.AddOctave(static_cast<NOISE_DATATYPE>(0.5),  static_cast<NOISE_DATATYPE>(1));
+	m_NoiseDistortZ.AddOctave(static_cast<NOISE_DATATYPE>(0.25), static_cast<NOISE_DATATYPE>(2));
 }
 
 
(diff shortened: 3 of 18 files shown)


src/Protocol/Protocol18x.cpp
```diff
@@ -2197,13 +2197,13 @@ void cProtocol180::HandlePacketLoginEncryptionResponse(cByteBuffer & a_ByteBuffe
 		m_Client->Kick("Hacked client");
 		return;
 	}
-	if (ntohl(DecryptedNonce[0]) != (unsigned)(uintptr_t)this)  // Cannot convert to c++ casts
+	if (ntohl(DecryptedNonce[0]) != static_cast<unsigned>(reinterpret_cast<std::uintptr_t>(this)))
```
SamJBarney: Remove the `std::`

src/Bindings/LuaFunctions.h
```diff
@@ -6,7 +6,7 @@
 
 inline unsigned int GetTime()
 {
-	return (unsigned int)time(0);
+	return static_cast<unsigned int>(time(0));
```
worktycho: This needs to be rewriten to std::chrono, because this cast is undefined in general.

src/ClientHandle.cpp
```diff
@@ -1924,7 +1924,7 @@ void cClientHandle::Tick(float a_Dt)
 		{
 			m_PingID++;
 			m_PingStartTime = std::chrono::steady_clock::now();
-			m_Protocol->SendKeepAlive(m_PingID);
+			m_Protocol->SendKeepAlive(static_cast<UInt32>(m_PingID));
```
worktycho: Why can't we change m_PingId to be UInt32

src/Generating/DistortedHeightmap.cpp
```diff
@@ -263,9 +263,9 @@ void cDistortedHeightmap::InitializeShapeGen(cIniFile & a_IniFile)
 
 int cDistortedHeightmap::GetHeightmapAt(NOISE_DATATYPE a_X, NOISE_DATATYPE a_Z)
 {
-	int RelX = (int)std::floor(a_X);
+	int RelX = static_cast<int>(std::floor(a_X));
```
worktycho: @madmaxoft Should this be FloorC?

SamJBarney: Yes it should. @madmaxoft drilled that into me :P

src/ClientHandle.cpp
SamJBarney: It should be converted into UInt32

madmaxoft:
@mjhanninen Set your editor not to remove the five lines at the bottom of each file.

I don't know where the editor authors get those ideas...

SphinxC0re:
One line would also do the trick

src/ClientHandle.cpp
mjhanninen: Or should `m_PingID` and `SendKeepAlive` both be changed to `Int32`? After all the protocol seems to be using signed 32-bit integers for the ID ... at least according to [this source](http://wiki.vg/Protocol#Keep_Alive), that is.

SamJBarney: As it doesn't matter what the value is, it would be easier to change m_PingID than go through and change cProtocol::SendKeepAlive() and any overloaded methods, along with cByteBuffer::WriteVarInt32().

mjhanninen: Curious. According to [the Minecraft protocol](http://wiki.vg/Protocol#Data_types) the VarInt type should be that same as [the Protobuf Varint](https://developers.google.com/protocol-buffers/docs/encoding#varints). However looking at [the encoding code](https://github.com/cuberite/cuberite/blob/6e4122e/src/ByteBuffer.cpp#L743-L759) it seems that the protocol doesn't employ [the ZigZag encoding](https://developers.google.com/protocol-buffers/docs/encoding#signed-integers). So it is the two's complement then, right?

madmaxoft: When varints were first implemented in Cuberite, the documentation only supported unsigned varints, so that's what was implemented. Signed ones were documented much later and so far are not implemented.

Since the actual value of the ping ID is not relevant, only that it is the same as the one sent, I suggest we convert it to `UInt32`.

mjhanninen: Fair, I'll do that.

SphinxC0re:
Looks good. Squash the commits into a single one and rebase onto the current master

mjhanninen:
Squashed. Turned out to be sitting on the latest master already.

src/PolarSSL++/RsaPrivateKey.cpp
```diff
@@ -100,7 +100,7 @@ AString cRsaPrivateKey::GetPubKeyDER(void)
 	{
 		return AString();
 	}
-	return AString((const char *)(buf + sizeof(buf) - res), (size_t)res);
+	return AString(static_cast<const char *>(static_cast<const void*>(buf + sizeof(buf) - res)), static_cast<size_t>(res));
```
madmaxoft: This should probably be a `reinterpret_cast`:
```cpp
return AString(reinterpret_cast<const char *>(buf + sizeof(buf) - res), static_cast<size_t>(res));
```

src/Protocol/Protocol17x.cpp
```diff
@@ -2145,7 +2145,7 @@ void cProtocol172::HandlePacketEntityAction(cByteBuffer & a_ByteBuffer)
 void cProtocol172::HandlePacketKeepAlive(cByteBuffer & a_ByteBuffer)
 {
 	HANDLE_READ(a_ByteBuffer, ReadBEInt32, Int32, KeepAliveID);
-	m_Client->HandleKeepAlive(KeepAliveID);
+	m_Client->HandleKeepAlive(static_cast<UInt32>(KeepAliveID));
```
madmaxoft: Change `KeepAliveID` to an `UInt32`, by changing the `HANDLE_READ`:
```cpp
HANDLE_READ(a_ByteBuffer, ReadBEUInt32, UInt32, KeepAliveID);
```
Then no cast will be needed.

src/Root.cpp
```diff
@@ -756,7 +756,7 @@ int cRoot::GetVirtualRAMUsage(void)
 		if (KERN_SUCCESS == task_info(
 			mach_task_self(),
 			TASK_BASIC_INFO,
-			(task_info_t)&t_info,
+			static_cast<task_info_t>(static_cast<void *>(&t_info)),
```
madmaxoft: Shouldn't this be a single `reinterpret_cast`?

src/Root.cpp
```diff
@@ -808,7 +808,7 @@ int cRoot::GetPhysicalRAMUsage(void)
...
```
madmaxoft: `reinterpret_cast`?

src/PolarSSL++/RsaPrivateKey.cpp
mjhanninen: Both should work and I'll change it to `reinterpret_cast`; however in my opinion using the weakest cast that does the job (however ugly it was) would be the conservative choice.

mjhanninen:
I'll squash the latest changes and rebase once they're okayed.

src/PolarSSL++/RsaPrivateKey.cpp
madmaxoft: I wasn't aware there was a "weakness" comparison between casts.
In my opinion, `reinterpret_cast` means no change to the data, while `static_cast` means "change the data if needed", so `reinterpret_cast` should be used when converting between raw pointer types.

madmaxoft:
Looks good to me; still I'd appreciate another set of eyes checking it out.

SphinxC0re:
Looks good on my side

mjhanninen:
Squashed & rebased.

madmaxoft:
Merge if build succeeds.

PR #2420, opened 2015-08-07
Flush log file when a warning or error message is output.

### LINKED ISSUE #2419: Flush log file after each warning and error
If the server outputs a warning or error message to the log, it should flush the file, so that all the log contents are synced to the disk. Otherwise error reports get lost because the server is terminated before it manages to flush the log.

madmaxoft:
Fixes #2419.

### DIFF
--- a/src/LoggerListeners.cpp
+++ b/src/LoggerListeners.cpp
@@ -317,6 +317,7 @@ cFileListener::cFileListener(void)
 void cFileListener::Log(AString a_Message, cLogger::eLogLevel a_LogLevel)
 {
 	const char * LogLevelPrefix = "Unkn ";
+	bool ShouldFlush = false;
 	switch (a_LogLevel)
 	{
 		case cLogger::llRegular:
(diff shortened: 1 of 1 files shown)


SphinxC0re:
Wouldn't it be good, if the logs contained every logline? For error reporting this would be a good thing to get a context

madmaxoft:
The problem is that flushing the file is a rather expensive operation, so you don't want to do it too often. Especially with the debug output, there's no point, really. I think having the flush for warnings and errors is a reasonable compromise.

SphinxC0re:
So does every log message get fushed into a file? If not we could solve the problem by caching the log output, flushing it into a file when warning or error mesages come

madmaxoft:
Right now the messages are cached by the C runtime and flushed out to file in 4-KiB blocks. This PR changes that so that after each warning or error message the buffers are flushed.

madmaxoft:
(rebased onto current master)

PR #2424, opened 2015-08-10
Compile MSVC2015

madmaxoft:
Fixed compilation in MSVC2015


tonibm19:
Wow, can you try android compilation?
I think vs2015 supports it

LO1ZB:
I don't think thats possibile for us right now, becuase CMake is not able to create Androit/iOS projekt for visual studio
http://public.kitware.com/Bug/view.php?id=15579

PR #2426, opened 2015-08-13
Fixed a few missing MCServers in APIDump

SphinxC0re:
There are **still** some inconsistencies about this. We have to push the renaming of the project way more than that


SphinxC0re:
I don't like breaking the rules, but ...

PR #2430, opened 2015-08-14
Readded plugin article

SphinxC0re:


madmaxoft:
Ah, this should have been handled differently, now we've lost the history for the file. Luckily it doesn't matter much for this file, but next time please don't do this.

SphinxC0re:
I renamed the other file, loosing it. And the new file hasn't been saved

PR #2433, opened 2015-08-14
Some warning fixes

SphinxC0re:


src/Bindings/DeprecatedBindings.cpp
```diff
@@ -30,7 +30,7 @@ static int tolua_get_AllToLua_g_BlockLightValue(lua_State* tolua_S)
 	}
 	#endif
 	BlockType = static_cast<BLOCKTYPE>(tolua_tonumber(tolua_S, 2, 0));
-	if ((BlockType < 0) || (BlockType > E_BLOCK_MAX_TYPE_ID))
+	if (BlockType > E_BLOCK_MAX_TYPE_ID)
```
NiLSPACE: Why?

src/Bindings/ManualBindings.h
```diff
@@ -263,7 +263,7 @@ class cManualBindings
 
 		// Get parameters:
 		Ty1 * Self = nullptr;
-		int BlockX, BlockY, BlockZ;
+		int BlockX, BlockY, BlockZ = 0;
```
NiLSPACE: I believe you now only have set `BlockZ` to 0.
You'd have to use this
```cpp
int BlockX = 0, BlockY = 0, BlockZ = 0;
```
or
```cpp
int BlockX(0), BlockY(0), BlockZ(0);
```

src/Bindings/DeprecatedBindings.cpp
SphinxC0re: because BLOCKTYPE is unsigned so comparisons against it being lesser than 0 are always true

madmaxoft:
I don't think this is a good way of fixing this.
We should `static_cast` the number received from Lua into an `int`, then compare it to both zero and max, and then use it for indexing.

With the current code, consider this plugin code:
```lua
local lightValue = g_BlockLightValue[-120]
```
The current code translates this to block type 136 and returns some valid number, instead of issuing a warning.

worktycho:
The best way would just to use the same approach as cLuaState::GetStackValue, which is to do the comparisons on floats, then static cast afterwards.

madmaxoft:
Rewriting it all to use `cLuaState` would be the best course of action anyway.

SphinxC0re:
Merge?

PR #2434, opened 2015-08-15
Update compile.sh

artur9010:


SphinxC0re:
This is not really complete. The compile script should be changed as a whole

SphinxC0re:
closed because of  #2436

PR #2437, opened 2015-08-17
Output directory change; cmake project name change

SphinxC0re:
I need a little bit of help with this one. I don't know why the Linux *debug* builds fail. Also I don't know why win builds fail. I don't know much about Windows builds


SphinxC0re:
I need a little bit of help with this one. I don't know why the Linux debug builds fail. Also I don't know why win builds fail. I don't know much about Windows builds

worktycho:
For windows builds, by renaming the project, you also rename the solution. So you need to update any scripts referencing the old solution name.

SphinxC0re:
I can't find the solution name in any script

worktycho:
As for linux builds, something is going wrong with postfix's. You might want to run CIBuild.sh locally to see what the name looks like.

worktycho:
Regarding the solution, I think cmake uses the master project, so you'll need to reconfigure appvayour to use the new solution through the web UI. (if you can, otherwise poke @madmaxoft)

SphinxC0re:
I don't have access to appveyor. @madmaxoft  could you adjust it?

src/CMakeLists.txt
```diff
@@ -331,12 +331,12 @@ add_executable(${EXECUTABLE} ${SOURCE})
...
+set(EXECUTABLE_OUTPUT_PATH ${CMAKE_SOURCE_DIR}/Server)
 SET_TARGET_PROPERTIES(${EXECUTABLE} PROPERTIES
-	RUNTIME_OUTPUT_DIRECTORY_DEBUG          ${CMAKE_SOURCE_DIR}/MCServer
-	RUNTIME_OUTPUT_DIRECTORY_RELEASE        ${CMAKE_SOURCE_DIR}/MCServer
-	RUNTIME_OUTPUT_DIRECTORY_DEBUGPROFILE   ${CMAKE_SOURCE_DIR}/MCServer
-	RUNTIME_OUTPUT_DIRECTORY_RELEASEPROFILE ${CMAKE_SOURCE_DIR}/MCServer
+        RUNTIME_OUTPUT_DIRECTORY_DEBUG          ${CMAKE_SOURCE_DIR}/Server
+        RUNTIME_OUTPUT_DIRECTORY_RELEASE        ${CMAKE_SOURCE_DIR}/Server
+        RUNTIME_OUTPUT_DIRECTORY_DEBUGPROFILE   ${CMAKE_SOURCE_DIR}/Server
+        RUNTIME_OUTPUT_DIRECTORY_RELEASEPROFILE ${CMAKE_SOURCE_DIR}/Server
```
madmaxoft: Don't change tabs to spaces, please.

madmaxoft:
If we change Appveyor settings, all the other builds will fail there.

I'll add the .yml file with Appveyor settings into the repo, so that the settings can be changes in PRs.
And you should have access to Appveyor, I did set the project so that all Github members can edit it.

madmaxoft:
Rebase (and squash) onto current master, change the appveyor.yml file to edit the settings.

madmaxoft:
Don't merge master, rebase onto it.

SphinxC0re:
Okay merge?

CMakeLists.txt
```diff
@@ -75,188 +75,187 @@ endif()
...
+		message(FATAL_ERROR "The project requires C++11 features. "
+			"You need at least Visual Studio 11 (Microsoft Visual Studio 2012), "
+			"with Microsoft Visual C++ Compiler Nov 2012 CTP (v120_CTP_Nov2012).")
+		elseif( MSVC_VERSION EQUAL 1700 )  # VC11 / VS2012
+			message( "VC11: using Microsoft Visual Studio 2012 "
+				"with Microsoft Visual C++ Compiler Nov 2012 CTP (v120_CTP_Nov2012)" )
+				set(CMAKE_GENERATOR_TOOLSET "v120_CTP_Nov2012" CACHE STRING "Platform Toolset" FORCE)
+			else() # VC12+, assuming C++11 supported.
+			endif()
+		endif()
```
madmaxoft: The indent gets wrong here. The `if()`, `elseif()`, `else()` and `endif()` should all be on the same level.

SphinxC0re:
Fixed. Merge?

madmaxoft:
There's still a few things missing - the profiling scripts for example. I'll fix those or remove them (they are kinda obsolete, having lastly worked on VS2008 + WinXP)

SphinxC0re:
Wouldn't this be part of another PR? It has near to nothing to do with this PR

madmaxoft:
This PR breaks those scripts, so either they should be removed (if the functionality is not needed anymore) or fixed.

madmaxoft:
Remove the scripts, MSVC2015 Community Edition has the profiler built in, it's no longer necessary to use external scripts.
($/Server/profile_run.cmd, $/Server/profile_run_debug.cmd, $/Server/profile_run_x64.cmd)

madmaxoft:
The valgrind and helgrind scripts ($/Server/hg, $/Server/vg) should be updated to use the proper executable name, too)

madmaxoft:
The install_windows_service script should get updated as well.

madmaxoft:
And I think you should add the original MCServer folder into the .gitignore, so that plugin developers don't suddenly try to commit their plugins from therein :)

SphinxC0re:
:pray: is it okay now?

SphinxC0re:
Isn't this ready for merge? @worktycho @madmaxoft @tigerw

tigerw:
Looks good. I think everyone's very bust at the moment.

SphinxC0re:
Okay, then I'll merge

madmaxoft:
Would you wait for me to review once more? No net in new home ,tomorrow

Julian Laubstein <notifications@github.com> wrote:

>Okay, then I'll merge
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/cuberite/cuberite/pull/2437#issuecomment-133923041

-- 

SphinxC0re:
Okay

CIbuild.sh
```diff
@@ -21,7 +21,7 @@ cd ..
 echo "Building..."
 make -j 2;
 make -j 2 test ARGS="-V";
-cd MCServer/;
+cd Server/;
 if [ "$TRAVIS_MCSERVER_BUILD_TYPE" != "COVERAGE" ]; then
 	echo restart | $MCSERVER_PATH;
 	echo stop | $MCSERVER_PATH;
```
madmaxoft: These three should be changed as well - how can it work without the change?

MCServer/delete_windows_service.cmd
```diff
@@ -1,4 +0,0 @@
-@echo off
-set SERVICENAME="MCServer"
-
-sc delete %SERVICENAME%
```
madmaxoft: This script should have been moved, not deleted.

MCServer/install_windows_service.cmd
```diff
@@ -1,7 +0,0 @@
-rem Alter this if you need to install multiple instances.
-@echo off
-set SERVICENAME="MCServer"
-
-set CURRENTDIR=%CD%
-sc create %SERVICENAME% binPath= "%CURRENTDIR%\MCServer.exe -d" start= auto DisplayName= %SERVICENAME%
-sc description %SERVICENAME% "Minecraft server instance"
```
madmaxoft: This script should have been moved, not deleted.

Server/install_windows_service.cmd
```diff
@@ -0,0 +1,7 @@
+rem Alter this if you need to install multiple instances.
+@echo off
+set SERVICENAME="Cuberite"
+
+set CURRENTDIR=%CD%
+sc create %SERVICENAME% binPath= "%CURRENTDIR%\Cuberite.exe -d" start= auto DisplayName= %SERVICENAME%
+sc description %SERVICENAME% "Minecraft server instance"
```
SphinxC0re: it isn't :dancer:

madmaxoft:
This change behaves poorly when applied to an already-initialized repository. The changed submodules require further manual steps ( http://stackoverflow.com/questions/20991138/moving-git-submodule-to-another-directory-within-the-same-repository#20991964 ) or else the `git submodule update` command fails - it tries to update the submodules in their original locations.
Did you move the submodules using `git mv`, or did you just rename the folder using your OS command? You may need to re-do that.

madmaxoft:
I'll try to fix it.

madmaxoft:
Fixed; if you want, amend the commit (so that it becomes yours again) and force-push the branch.

.gitmodules
```diff
@@ -1,23 +1,23 @@
 [submodule "MCServer/Plugins/Core"]
-	path = MCServer/Plugins/Core
+	path = Server/Plugins/Core
 	url = https://github.com/mc-server/Core.git
 [submodule "MCServer/Plugins/ProtectionAreas"]
```
madmaxoft: Note: the submodule headers will have to stay like this (`MCServer/...`) in order to work with already-initialized repos.

SphinxC0re: And when will we be able to change that?

madmaxoft: Never. Once the submodules have been named like this, they need to stay or else you break everyone's checked-out repo.

Since this is only a git-internal name for the submodule and not the actual path, it's of no consequence, really. On the disk the files will be checked out to the proper path.

SphinxC0re: What about this
http://stackoverflow.com/questions/4526910/rename-a-git-submodule
it tells us that we have to *sync* the submodules

madmaxoft: Yes, that would work, but everyone would have to execute that in their repo everytime they checked out something across this commit.

SphinxC0re: I don't think so. Git modules get saved into the ```.git/modules``` folder. So they should be committed.

madmaxoft:
Yes, they are committed, but they need to have the same git-internal name in order for those changes to apply correctly. Which is exactly what I've done - I've changed the paths, but kept the git-internal names.

SphinxC0re:
But the git internals are also updated and committed with the *sync*

SphinxC0re:
I get an error while trying to add these submodules:
```
error: unable to index file Server/Plugins/ChatLog
fatal: Aktualisierung der Dateien fehlgeschlagen
```

SphinxC0re:
I don't really mind whose commit this will be. The change is important

madmaxoft:
I'd like to get this merged (I rebased onto current master, no other changes from previous versions).
Is anyone against? @NiLSPACE @worktycho @tigerw ?

worktycho:
No objections.

NiLSPACE:
Looks good to me. The only thing I could think of is that if the favicon is changed that the executable icon could be changed to the new one as well. 

Or **is** that the executable icon? If that's the case then that isn't the one we agreed on.

madmaxoft:
`Server/favicon.png` is the icon that is sent to the client for the server list. It's true that the Windows executable icon should be changed, too; I don't see it as necessary within this PR, though.

If that's the only objection, how about merging this and handling the icons later on with a full-blown graphic review?

NiLSPACE:
I can live with that :) I just thought since the favicon icon is already changed in this PR why not do the executable icon as well.

madmaxoft:
Note that for existing repositories, you will need to do a `git submodule update --init` to rebuild the submodules, and then manually move all unversioned files (world, players, logs etc.) from the MCServer folder to the Server folder.

Issue #2440, opened 2015-08-21
OS X build broken due to new compiler flag


mjhanninen:
OS X build fails due to new compiler flag `-Wno-format-pedantic` introduced in the PR #2380:

    [ 26%] Built target event_core
    [ 26%] Building CXX object src/OSSupport/CMakeFiles/OSSupport.dir/CriticalSection.cpp.o
    error: unknown warning option '-Wno-format-pedantic' [-Werror,-Wunknown-warning-option]
    src/OSSupport/CMakeFiles/OSSupport.dir/build.make:62: recipe for target 'src/OSSupport/CMakeFiles/OSSupport.dir/CriticalSection.cpp.o' failed
    make[2]: *** [src/OSSupport/CMakeFiles/OSSupport.dir/CriticalSection.cpp.o] Error 1
    CMakeFiles/Makefile2:1090: recipe for target 'src/OSSupport/CMakeFiles/OSSupport.dir/all' failed
    make[1]: *** [src/OSSupport/CMakeFiles/OSSupport.dir/all] Error 2
    Makefile:127: recipe for target 'all' failed
    make: *** [all] Error 2

The compiler in question is:

    Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
    Target: x86_64-apple-darwin14.4.0
    Thread model: posix

PR #2441, opened 2015-08-21
Revert Clang compiler flag that breaks OS X build

mjhanninen:
This is probably not the right way to fix the problem reported in #2440 but let's OS X people to build the software in the interim.

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -277,7 +277,6 @@ macro(set_exe_flags)
 					# Ignore another problem in sqlite
 					add_flags_cxx("-Wno-documentation-unknown-command")
 				endif()
-				add_flags_cxx("-Wno-format-pedantic")
 			endif()
 		endif()
 	endif()


tigerw:
Could you do what the other compiler flags do in this block and do a check for existence before adding? That's probably a better fix.

tigerw:
Also, ermagherd I broke something; @SamJBarney, pitchfork please.

mjhanninen:
Indeed, that's a lot better solution. Duh.

SamJBarney:
@tigerw  how about an angry mob instead?
Grab a Pitchfork

tigerw:
Aaaaah! *♪[The Simpsons...](https://www.youtube.com/watch?v=DX1iplQQJTo)♪*

tigerw:
Squash into one commit with 2nd commit's message?

SphinxC0re:
Also rebase onto current master

tigerw:
I've always wondered [@SphinxC0re](http://www.crystalinks.com/sphinx3.gif), why rebase onto master if GitHub can merge into master as-is, i.e. no merge conflicts?

worktycho:
its because with long running feature branches the merge results in a complex tree if you don't rebase first. however for a two commit PR I think we can merge this.

tigerw:
Even if it's squashed into one commit?

worktycho:
Yes.

Also, we really need better testing against apples clang.

tigerw:
When I say apples you say clang!
apples clang!
apples clang!

:DD

tigerw:
In fact, as per https://github.com/cuberite/cuberite/pull/2380/files#r37696134, the issues which the flag warns about are technically issues. Just remove the lines completely :)

mjhanninen:
Done, squashed, and rebased.

PR #2445, opened 2015-08-22
Reduce precision on monster targets

### LINKED ISSUE #2444: cAggressiveMonster's target arrival detection possibly broken
Can [this](https://github.com/cuberite/cuberite/blob/a341419b04/src/Mobs/AggressiveMonster.cpp#L111-L124) be right? To my eyes it looks like the arrival to the target position `(x0, z0)` is detected through condition `|x -x0| < tolerance && |z - z0| > tolerance` which in my opinion is incorrect (note the direction of the latter inequality).

Relatedly `std::numeric_limits<float>::epsilon()` is a rather tight tolerance. I wonder if the tolerance should be derived from the unit of length as opposed to float precision?

worktycho:
fixes incorrect z comparison.
Fixes #2444

### DIFF
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -110,13 +110,13 @@ void cAggressiveMonster::Attack(std::chrono::milliseconds a_Dt)
 
 bool cAggressiveMonster::IsMovingToTargetPosition()
 {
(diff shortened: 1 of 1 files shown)


tigerw:
I believe this function isn't used.

worktycho:
okay, so it should be removed. i can't do that through the web interface, so I'll have to do it later.

PR #2446, opened 2015-08-22
Fixed a position bug in the pathfinder

tigerw:
@SafwatHalaby?

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -17,9 +17,6 @@
 
 
 
-
-
-
 bool compareHeuristics::operator()(cPathCell * & a_Cell1, cPathCell * & a_Cell2)
 {
 	return a_Cell1->m_F > a_Cell2->m_F;
@@ -36,7 +33,7 @@ cPath::cPath(
 	double a_BoundingBoxWidth, double a_BoundingBoxHeight,
 	int a_MaxUp, int a_MaxDown
 ) :
-
+	m_StepsLeft(a_MaxSteps),
 	m_CurrentPoint(0),  // GetNextPoint increments this to 1, but that's fine, since the first cell is always a_StartingPoint
 	m_Chunk(&a_Chunk),
 	m_BadChunkFound(false)
@@ -67,22 +64,8 @@ cPath::cPath(
 
 	m_NearestPointToTarget = GetCell(m_Source);
 	m_Status = ePathFinderStatus::CALCULATING;
-	m_StepsLeft = a_MaxSteps;
 
-	ProcessCell(GetCell(a_StartingPoint), nullptr, 0);
-	m_Chunk = nullptr;
-}
-
-
-
-
-
-cPath::~cPath()
-{
-	if (m_Status == ePathFinderStatus::CALCULATING)
-	{
-		FinishCalculation();
-	}
+	ProcessCell(GetCell(m_Source), nullptr, 0);
 }
 
 
@@ -113,7 +96,7 @@ ePathFinderStatus cPath::Step(cChunk & a_Chunk)
 		int i;
 		for (i = 0; i < CALCULATIONS_PER_STEP; ++i)
 		{
-			if (Step_Internal())  // Step_Internal returns true when no more calculation is needed.
+			if (StepOnce())  // StepOnce returns true when no more calculation is needed.
 			{
 				break;  // if we're here, m_Status must have changed either to PATH_FOUND or PATH_NOT_FOUND.
 			}
@@ -179,7 +162,7 @@ bool cPath::IsSolid(const Vector3i & a_Location)
 
 
 
-bool cPath::Step_Internal()
+bool cPath::StepOnce()
 {
 	cPathCell * CurrentCell = OpenListPop();
 
--- a/src/Mobs/Path.h
+++ b/src/Mobs/Path.h
@@ -72,10 +72,7 @@ public:
 		double a_BoundingBoxWidth, double a_BoundingBoxHeight,
 		int a_MaxUp = 1, int a_MaxDown = 1
 	);
-
-	/** Destroys the path and frees its memory. */
-	~cPath();
-
+	
 	/** Performs part of the path calculation and returns the appropriate status.
 	If NEARBY_FOUND is returned, it means that the destination is not reachable, but a nearby destination
 	is reachable. If the user likes the alternative destination, they can call AcceptNearbyPath to treat the path as found,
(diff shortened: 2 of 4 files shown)


SphinxC0re:
You should have waited for @SafwatHalaby to review this

jammet:
Hope Safwat is alright, haven't seen him around lately.

SafwatHalaby:
I'm back :) I'll check it out.

jammet:
:dancers: Welcome back! :]

SafwatHalaby:
Am I wrong to deduce that this calculates a path every tick regardless of *anything*?

SafwatHalaby:
Also, what bug does this fix?

src/Mobs/Path.cpp
```diff
@@ -67,22 +64,8 @@ cPath::cPath(
...
-
-
-
-cPath::~cPath()
-{
-	if (m_Status == ePathFinderStatus::CALCULATING)
-	{
-		FinishCalculation();
-	}
+	ProcessCell(GetCell(m_Source), nullptr, 0);
```
tigerw: Here is the bug:

`ProcessCell(GetCell(a_StartingPoint)`
instead of
`ProcessCell(GetCell(m_Source)`

Where m_Source is a floored and adjusted a_StartingPoint.

SafwatHalaby: The pathfinder needs the adjusted version.

tigerw:
Looks like it looks for a path until is finds one, at which point the route is followed until the final destination is reached (and it stops), or it times out, and restarts.

src/Mobs/Path.cpp
SafwatHalaby: From the Pathfinder's perspective, the bounding box is at one of the bottom corners (I don't remember which one right now), the game's bounding box is centered, so I adjust the point before feeding it to the Pathfinder.

SafwatHalaby:
You calculate a new path inside PATH_FOUND, and you neutralized the cooldown (TicksSinceLastPathReset), I think it's simply calculating every single tick.

src/Mobs/Path.cpp
SafwatHalaby: Oh, sorry, I thought you replaced m_Source but you actually added m_Source. Good catch.

SafwatHalaby:
The point I had about recalculation is invalid, but there are some issues:

1. Due to the removal of TicksSinceLastPathReset,  the mob never recalculates unless the final destination is reached, this is funny on long distances, you pass by the mob and they keep going to where you've been 10 seconds ago.

2. Due to the removal of NoMoreWayPoints, the mob circles around the final waypoint for 40 ticks before recalculating.

src/Mobs/Path.h
```diff
@@ -99,45 +96,27 @@ class cPath
...
 
-	/** Returns the total number of points this path has. */
-	inline size_t GetPointCount()
-	{
-		if (m_Status != ePathFinderStatus::PATH_FOUND)
-		{
-			return 0;
-		}
-		return m_PathPoints.size();
-	}
```
SafwatHalaby: These two methods are unused now, but we might use them later.

SafwatHalaby:
Also, there *is* a difference between FinalDestination and PathFinderDestination, the former is where the mob wants to go, the latter is where it's going right now.

jammet:
(Very complicated stuff... wow..)

SafwatHalaby:
I like the `Path.cpp` changes, but I think `monster.cpp` needs a revert because of the issues mentioned above. Can I do that?

SafwatHalaby:
I'll keep the `PathfinderActivated` change though.

tigerw:
> Due to the removal of NoMoreWayPoints, the mob circles around the final waypoint for 40 ticks before recalculating.

Indeed. Do put the reset code into a function and call it from both places though, instead of a variable.

> Due to the removal of TicksSinceLastPathReset, the mob never recalculates unless the final destination is reached, this is funny on long distances, you pass by the mob and they keep going to where you've been 10 seconds ago.

Looks like it. Since mobiles call MoveToPosition constantly, instead of counting ticks/distances, have the implementation of MoveToPosition reset if the current and new destinations are too far apart.

> Also, there is a difference between FinalDestination and PathFinderDestination, the former is where the mob wants to go, the latter is where it's going right now.

Is a distinction needed though?

SafwatHalaby:
> Indeed. Do put the reset code into a function and call it from both places though, instead of a variable.

I don't follow. Which two places are you referring to? And what do you mean by "reset code"?

> Is a distinction needed though?

Yes, in the previous code there was a clear distinction:
PathFinderDestination = Where we are going.
FinalDestination = Where we should go.

In the code you've deleted, `PathFinderDestination` and `FinalDestination` are compared, if the distance between them is too big, and if `ticksSinceLastReset` is large enough, we reset and recalculate. This is why I need `ticksSinceLastReset` in the first way, by the way, to tame recalcs.

SafwatHalaby:
In other words, this piece of code should have either stayed or replaced with some better recalc logic.

```
	if ((m_FinalDestination - m_PathFinderDestination).Length() > 0.25)  // if the distance between where we're going and where we should go is too big.
	{
		/* If we reached the last path waypoint,
		Or if we haven't re-calculated for too long.
		Interval is proportional to distance squared, and its minimum is 10.
		(Recalculate lots when close, calculate rarely when far) */
		if (
			((GetPosition() - m_PathFinderDestination).Length() < 0.25) ||
			((m_TicksSinceLastPathReset > 10) && (m_TicksSinceLastPathReset > (0.4 * (m_FinalDestination - GetPosition()).SqrLength())))
		)
		{
			/* Re-calculating is expensive when there's no path to target, and it results in mobs freezing very often as a result of always recalculating.
			This is a workaround till we get better path recalculation. */
			if (!m_NoPathToTarget)
			{
				ResetPathFinding();
			}
		}
	}
```

tigerw:
It seems to me there that its just recalculating for no reason?

SafwatHalaby:
It seems your missing the point. It is recalculating for a very good reason: `FinalDestination` (The spot where the player is at) and `PathFinderDestination` (The spot where the mob is heading) have deviated too far apart.

In an ideal world, we recalculate every time the player moves, hence `FinalDestination` and `PathFinderDestination` would be identical. But in reality we must save some recalcs by approximating; only calculate when those two vars are too far apart / have been unequal for too long. This is why the distinction is needed.

tigerw:
> I don't follow. Which two places are you referring to? And what do you mean by "reset code"?

NoMoreWayPoints aims to recalc if the end was reached but not the final destination by resetting the pathfinder. I am saying that instead of having a variable that is checked on the next tick, do the reset where the variable was originally set. The other place is the reset that occurs when the entity timed out reaching the next waypoint.

SafwatHalaby:
> do the reset where the variable was originally set.

The reason I originally did it that way is because otherwise the mob would miss the last waypoint (Which is fine for attacking mobs, the last waypoint is where the player is standing, but it's not fine for generic pathfinding).

SafwatHalaby:
I wonder if that's still the case though ... (regarding missing last point)

SafwatHalaby:
>have been unequal for too long. This is why the distinction is needed.

...And this is why the ticks var is needed.

tigerw:
> It seems your missing the point. It is recalculating for a very good reason:  FinalDestination  (The spot where the player is at) and  PathFinderDestination  (The spot where the mob is heading) have deviated too far apart.

> In an ideal world, we recalculate every time the player moves, hence  FinalDestination  and  PathFinderDestination  would be identical. But in reality we must save some recalcs by approximating; only calculate when those two vars are too far apart / have been unequal for too long. This is why the distinction is needed.

I mentioned something about this previously:
> Looks like it. Since mobiles call MoveToPosition constantly, instead of counting ticks/distances, have the implementation of MoveToPosition reset if the current and new destinations are too far apart.

If performance really is a problem, increase the distance at which MoveToPosition decides to reset.

tigerw:
> The reason I originally did it that way is because otherwise the mob would miss the last waypoint (Which is fine for attacking mobs, the last waypoint is where the player is standing, but it's not fine for generic pathfinding).

They *shouldn't* miss the last waypoint, otherwise ReachedFinalDestination would never return true and the mobiles would never stop searching, but I dunno.

SafwatHalaby:
>They shouldn't miss the last waypoint, otherwise ReachedFinalDestination would never return true and the mobiles would never stop searching, but I dunno.

For attacking mobs:  `ReachedFinalDestination` would never be true anyways because there's this code in the `chasing` state which is somewhat like: `if distance to player is small enough, stopMovingToposition()`

For non attacking mobs: Yes, eachedFinalDestination would never return true and they get stuck. *This* is why I put it. 

Some changes have been made though, I'm not sure the statements I made are still true. Will need to check.

tigerw:
Then looks like both of those places need to be fixed, can't have code vying for control over who gets to call what :P

SafwatHalaby:
Well, here's the deal:

As I mentioned elsewhere, I'd like to get all that path recalc logic out of `monster.cpp` and move it into a new class because `monster.cpp` is getting fat and bloated. Recalc is really not worth fixing because I'm going to rewrite lots of it anyways.

However, I think I should apply my revert PR in the meantime. yes, it's far from perfect, but it will be a temporary fix. I'd like to do this because the current recalc is completely broken: It only recalc when it reaches the last waypoint. By that time the player might be long gone. And does that only  after a 40-tick weird dance.

What say you?

SafwatHalaby:
I won't revert `path.cpp`, those were good changes.

tigerw:
If you want, but I'm just saying that, again, if you want, you could move the recalculation into MoveToPosition and do what I said about NoMoreWaypoints etc.

It seems that you're intending to change it all anyway though, so I would say either don't bother or revert if it please you.

SafwatHalaby:
>If you want, but I'm just saying that, again, if you want, you could move the recalculation into MoveToPosition and do what I said about NoMoreWaypoints etc.

That would have been a neat change, but since I'm starting to move the recalc code out immediately, I won't bother.

SafwatHalaby:
>If performance really is a problem, increase the distance at which MoveToPosition decides to reset.

It is worth mentioning that even after moving the recalc code to MoveToPosition, one may still need two dest vars and/or a time keeping variable (TicksSinceLastPathReset), depending on the exact recalculation algorithm. A simple algorithm of "if delta is larger than x" requires neither but may be suboptimal.

PR #2427, opened 2015-08-13
Fixed and improved restarting

tigerw:
Restarts are now an actual, close-as-possible to application exit+reopen.

Fixes #2232
Fixes #2094

PR #2447, opened 2015-08-23
Fix FreeBSD build by including signal.h

mjhanninen:
On FreeBSD the build dies due to missing `signal.h`:

    [ 97%] Building CXX object src/CMakeFiles/MCServer.dir/Root.cpp.o
    /home/matti/cuberite/src/Root.cpp:301:51: error: use of undeclared identifier 'SIGKILL'
                    if (pthread_kill(m_InputThread.native_handle(), SIGKILL) != 0)

Dependency on `signal.h` was introduced by #2427.

This PR fixes the build. However I wonder whether these POSIX headers should be included in less a case-by-case manner. I'm not a pro on that front so I couldn't tell.

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -25,13 +25,15 @@
 
 #include <iostream>
 
-#ifdef _WIN32
+#if defined(_WIN32)
 	#include <psapi.h>
-#elif defined(__linux__)
-	#include <fstream>
+#elif defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
 	#include <signal.h>
-#elif defined(__APPLE__)
-	#include <mach/mach.h>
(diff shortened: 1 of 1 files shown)


worktycho:
I agree with you that a more general solution is necessary. How about using the  `__unix__` macro instead?

mjhanninen:
Yep, I guess it is less wrong. According to [this SO answer](http://stackoverflow.com/a/11351171) the following is a reasonable way of detecting a POSIX system:

    #if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
    ...
    #endif

worktycho:
That seems reasonable.

madmaxoft:
Looks good, squash and rebase onto current master and we can merge.

PR #2450, opened 2015-08-24
Update cRankManager::SetRankVisuals

nicodinh:
delete LOGINFO message

### DIFF
--- a/src/RankManager.cpp
+++ b/src/RankManager.cpp
@@ -1846,7 +1846,7 @@ void cRankManager::SetRankVisuals(
 		stmt.bind(2, a_MsgSuffix);
 		stmt.bind(3, a_MsgNameColorCode);
 		stmt.bind(4, a_RankName);
-		if (!stmt.executeStep())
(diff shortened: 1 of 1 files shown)


madmaxoft:
Why would you want to remove the message? Setting visuals for a non-existent rank should give some kind of an error.

nicodinh:
Currently the message is displayed even if the update is done.

madmaxoft:
Ah, of course, the update statement doesn't return a row, so the condition was wrong.
Still, I think the log message should be present for the failure case.

madmaxoft:
Okay, can you squash the two commits into a single one?

SphinxC0re:
Also, rebase onto current master

nicodinh:
i'm lost, i don't know what i'm doing...

SphinxC0re:
NOoooooooooooo, don't close

nicodinh:
@SphinxC0re  Thanks a lot !
It's better with -f.

PR #2451, opened 2015-08-24
Changed mob age from char to int

### LINKED ISSUE #2398: Change cMonster age from char to int
Current code represents the mob age as a char; this is clearly wrong, the age should be a full int, incremented at each server tick.
Ref.: http://minecraft.gamepedia.com/Chunk_format#Breedable, "Age" is a TAG_Int

SphinxC0re:
Fixes #2398

### DIFF
--- a/src/WorldStorage/WSSAnvil.cpp
+++ b/src/WorldStorage/WSSAnvil.cpp
@@ -2344,8 +2344,14 @@ void cWSSAnvil::LoadHorseFromNBT(cEntityList & a_Entities, const cParsedNBT & a_
 	int AgeableIdx  = a_NBT.FindChildByName(a_TagIdx, "Age");
 	if (AgeableIdx > 0)
 	{
-		Byte Age = a_NBT.GetByte(AgeableIdx);
-		Monster->SetAge(static_cast<char>(Age));
+		int Age;
+		switch (a_NBT.GetType(AgeableIdx))
+		{
+			case TAG_Byte: Age = static_cast<int>(a_NBT.GetByte(AgeableIdx)); break;
+			case TAG_Int:  Age = a_NBT.GetInt (AgeableIdx); break;
+			default: Age = 0; break;
+		}
+		Monster->SetAge(Age);
 	}
 	
 	a_Entities.push_back(Monster.release());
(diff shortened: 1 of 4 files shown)


madmaxoft:
You need to fix the NBT reading and the protocol, too

SphinxC0re:
Where is the reading part located?

SphinxC0re:
Is this okay now?

madmaxoft:
You didn't test-play after the changes, did you? I think you broke the protocol. The values there are single bytes, they only specify whether the mob is a baby or not, rather than the full age.
http://wiki.vg/Entities#Ageable

SphinxC0re:
Okay, now I'm confused. Is it a char, an int or a byte? The issue is about changing the mob age from char to int

madmaxoft:
Yes, the age itself should be an int. But in the client protocol, only a byte flag is sent, whether the mob is a baby or not, so it should stay a char.
`a_Pkt.WriteBEInt8(Horse.IsBaby() ? -1 : 0);`

madmaxoft:
We need to get this fixed. Worlds generated by Vanilla assert in Cuberite because Cuberite tries to read the age as a byte, while it is indeed an int in the NBT.

bibo38:
@SphinxC0re Any updates on this?
In short terms:

- NBT stores this as an int
- The protocol only uses a byte, because the client only needs to know, that it is a baby

SphinxC0re:
I halted the work on this for a while. But it shouldn't be too hard to fix this

SphinxC0re:
This should be okay to merge now

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2344,8 +2344,8 @@ void cWSSAnvil::LoadHorseFromNBT(cEntityList & a_Entities, const cParsedNBT & a_
 	int AgeableIdx  = a_NBT.FindChildByName(a_TagIdx, "Age");
 	if (AgeableIdx > 0)
 	{
-		Byte Age = a_NBT.GetByte(AgeableIdx);
-		Monster->SetAge(static_cast<char>(Age));
+		int Age = a_NBT.GetInt(AgeableIdx);
```
madmaxoft: This will now assert on maps created by the previous code, because it used to store the age as a char, but now it's reading an int. You need to detect the item's datatype and load accordingly.

SphinxC0re: How do I do that?

madmaxoft:
```cpp
int Age;
switch (a_NBT.GetType(AgeIdx))
{
  case TAG_Byte: Age = a_NBT.GetByte(AgeIdx); break;
  case TAG_Int:  Age = a_NBT.GetInt (AgeIdx); break;
  default: Age = 0; break;
}
```

SphinxC0re:
Is this okay now?

SphinxC0re:
@cuberite/core-team

PR #2457, opened 2015-08-30
Increase robustness of the logging subsystem

worktycho:
Note: This PR does not address the issue of failing to write to the log, only creating the log listener.

### DIFF
--- a/Tools/MCADefrag/MCADefrag.cpp
+++ b/Tools/MCADefrag/MCADefrag.cpp
@@ -22,25 +22,25 @@ static const Byte g_Zeroes[4096] = {0};
 
 int main(int argc, char ** argv)
 {
-	cLogger::cListener * consoleLogListener = MakeConsoleListener(false);
-	cLogger::cListener * fileLogListener = new cFileListener();
-	cLogger::GetInstance().AttachListener(consoleLogListener);
-	cLogger::GetInstance().AttachListener(fileLogListener);
+	auto consoleLogListener = MakeConsoleListener(false);
+	auto consoleAttachment = cLogger::GetInstance().AttachListener(std::move(consoleLogListener));
+	auto fileLogListenerRet = MakeFileListener();
+	if (!fileLogListenerRet.first)
+	{
+		LOGERROR("Failed to open log file, aborting");
+		return EXIT_FAILURE;
+	}
+	auto fileAttachment = cLogger::GetInstance().AttachListener(std::move(fileLogListenerRet.second));
 	
 	cLogger::InitiateMultithreading();
 	
 	cMCADefrag Defrag;
 	if (!Defrag.Init(argc, argv))
 	{
-		return 1;
+		return EXIT_FAILURE;
 	}
 	
 	Defrag.Run();
-	
-	cLogger::GetInstance().DetachListener(consoleLogListener);
-	delete consoleLogListener;
-	cLogger::GetInstance().DetachListener(fileLogListener);
-	delete fileLogListener;
 	
 	return 0;
 }
--- a/Tools/ProtoProxy/ProtoProxy.cpp
+++ b/Tools/ProtoProxy/ProtoProxy.cpp
@@ -15,11 +15,17 @@
 int main(int argc, char ** argv)
 {
 	// Initialize logging subsystem:
-	cLogger::InitiateMultithreading();
 	auto consoleLogListener = MakeConsoleListener(false);
-	auto fileLogListener = new cFileListener();
-	cLogger::GetInstance().AttachListener(consoleLogListener);
-	cLogger::GetInstance().AttachListener(fileLogListener);
+	auto consoleAttachment = cLogger::GetInstance().AttachListener(std::move(consoleLogListener));
+	auto fileLogListenerRet = MakeFileListener();
+	if (!fileLogListenerRet.first)
+	{
+		LOGERROR("Failed to open log file, aborting");
+		return EXIT_FAILURE;
+	}
+	auto fileAttachment = cLogger::GetInstance().AttachListener(std::move(fileLogListenerRet.second));
+
+	cLogger::InitiateMultithreading();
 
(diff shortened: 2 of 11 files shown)


Tools/MCADefrag/MCADefrag.cpp
```diff
@@ -22,10 +22,15 @@ static const Byte g_Zeroes[4096] = {0};
...
-	cLogger::cListener * fileLogListener = new cFileListener();
-	cLogger::GetInstance().AttachListener(consoleLogListener);
-	cLogger::GetInstance().AttachListener(fileLogListener);
+	auto consoleLogListener = MakeConsoleListener(false);
+	auto consoleAttachment = cLogger::GetInstance().AttachListener(std::move(consoleLogListener));
+	auto fileLogListenerRet = MakeFileListener();
+	if (!fileLogListenerRet.first)
+	{
+		LOGERROR("Failed to open log file, aborting");
+		return 1;
```
tigerw: EXIT_FAILURE my good sir.

Tools/ProtoProxy/ProtoProxy.cpp
```diff
@@ -15,11 +15,17 @@
...
 	auto consoleLogListener = MakeConsoleListener(false);
-	auto fileLogListener = new cFileListener();
-	cLogger::GetInstance().AttachListener(consoleLogListener);
-	cLogger::GetInstance().AttachListener(fileLogListener);
+	auto consoleAttachment = cLogger::GetInstance().AttachListener(std::move(consoleLogListener));
+	auto fileLogListenerRet = MakeFileListener();
+	if (!fileLogListenerRet.first)
+	{
+		LOGERROR("Failed to open log file, aborting");
+		return 1;
```
tigerw: EXIT_FAILURE?

src/Logger.cpp
```diff
@@ -86,7 +91,16 @@ void cLogger::AttachListener(cListener * a_Listener)
...
-	m_LogListeners.erase(std::remove(m_LogListeners.begin(), m_LogListeners.end(), a_Listener));
+	m_LogListeners.erase(
+		std::remove_if(
+			m_LogListeners.begin(),
+			m_LogListeners.end(),
+			[=](std::unique_ptr<cListener> & a_OtherListener) -> bool
+			{
+				return a_OtherListener.get() == a_Listener;
+			}
+		)
```
tigerw: Don't you normally specify a m_LogListeners.end() here?

worktycho: Where? remove_if is just remove with  custom comparison function, and the erase call is the same as before.

tigerw: Oh, so if you give .erase a position and nothing else, it defaults to removing from there 'till the end? I always thought you had to specify a stop point.

Tools/MCADefrag/MCADefrag.cpp
```diff
@@ -201,7 +201,7 @@ void cMCADefrag::cThread::ProcessFile(const AString & a_FileName)
 	// Write a copy of the Timestamps into the Out file:
 	if (Out.Write(Timestamps, sizeof(Timestamps)) != sizeof(Timestamps))
 	{
-		LOGWARNING("Cannot write Timestamps to file %s, skipping file.", OutFileName.c_str());
+		LOGWARN("Cannot write Timestamps to file %s, skipping file.", OutFileName.c_str());
```
NiLSPACE: According to the Lua API documentation `LOGWARN` is obsolete, and we should use `LOGWARNING` instead.

worktycho: Right, I'll change the macro and function around.

worktycho:
Fixed comments.

madmaxoft:
Still failing

worktycho:
Fixed CI.

src/Logger.cpp
tigerw: How strange. When I was doing #2463, I tried this with a std::list<>::erase and ...vector... and both crashed if remove_if found nothing to remove. Yet it works here :/

Also, is it necessary to specify -> bool and must you capture everything by value? :P
Why not just put a_Listener in the capture list?

worktycho: The `-> bool` is not necessary, but it adds a little clarity, and with the lambda style we use, its not overly verbose.

As for by value captures, the only thing this lambda captures is a pointer, which is usually more efficient to capture by value.

madmaxoft:
Wouldn't it be easier to just use a `SharedPtr` for the listeners? It's not like we'll be removing listeners explicitly, so all this needs is a bit of implicit lifetime management.

src/Logger.h
```diff
@@ -23,13 +23,36 @@ class cLogger
 		virtual ~cListener(){}
 	};
 
+	class cAttachment
+	{
+		public:
+
+		cAttachment(cAttachment & other) = default;
```
NiLSPACE: `a_Other`?

worktycho:
I used a unique_ptr, because it simplifies the memory ownership, and having the separate attachment class makes it clear when a listener is attached.

madmaxoft:
Just a quick thought: Will this support plugin-based listeners in the future? Say, like a plugin registers a log listener that calls a plugin-defined callback function for each message output; we will need to be able to remove the listener somehow and yet there's no good way of saying when exactly to delete the pointer (plugin is unloading in one thread, logs are being produced by another thread etc.)

worktycho:
The pointer is deleted when the handle falls out of scope. So I don't think it would be anymore difficult to handle than any uses of unique_ptr, and if it is, I can easily add an explicit deletion function back in.

worktycho:
Merge?

madmaxoft:
I still think using a `SharedPtr` would have been better for the plugins' case

worktycho:
Yet at the moment we have no plugins, and I don't want to add the complexity of yet another shared pointer unnecessarily. It would be trivial to convert this to a shared ptr when we add plugin listeners.

PR #2459, opened 2015-09-02
Fixes #2458

### LINKED ISSUE #2458: Massive amount of cPickup messages problematic
While it would be great to get to the root of the real problem, the server should not flood error messages by the dotzens every few seconds. Some mechanism to reduce verbosity of error reporting, via loglevel, disabling logging altogether, or piping it all to /dev/null (which sadly cannot be done for a directiory, such as /logs -- only individual files), or another method, would go a long way. 

Right now I cannot run the server because the logfiles generated are so large that I would have to constantly be on my toes and check if enough drive space is still available. 

Example:
`Warn [12:42:00] Entity at 0x717d3908 (cPickup, ID 133) spawning in a non-existent chunk, the entity is lost.`

The underlying, real problem is a mystery to me, but the amount of these error messages alone is a problem all in itself right now.

tigerw:
* Simplified entity adding

@jammet, so what happened was.
1. I fixed a typo where chunk unloading would only happen once every 5 minutes. It was intended to be once every 10 seconds.
2. In your world, at (148 77 103) and (103 90 120) you have jukeboxes with discs in them.
3. These happened to be in chunks that got unloaded at the end of this 10 second interval.
4. Jukeboxes eject their discs on unload, so this queued a new entity to be added to the unloading chunk
5. The server gets round to adding the queued entity, and promptly finds the chunk unloaded. It consequently tries to load the chunk again, and thus the jukebox, which saved with the disc still inside, since saving happens before the disc is ejected.
6. This loaded chunk is the unloaded 10 seconds later, and thus repeat *ad infinitum*, spamming those 'cPickup spawned in non-existent chunk' messages.

~~Now the pickup entities are added immediately to the [still alive](https://www.youtube.com/watch?v=Y6ljFaKRTrI) chunk, and are destroyed when it is [nullified](https://www.youtube.com/watch?v=uqH_Y1TupoQ). There is still the issue that jukeboxes eject a disc, effectively into the ether, when they unload, since everything's discarded afterwards, but oh well.~~

Okay, as suspected the above has concurrency issues. New fix is to fix the bug reported in the second part of the above. Jukeboxes don't throw a disc into the air when they're about to ~~die~~ be unloaded.

### DIFF
--- a/src/BlockEntities/JukeboxEntity.cpp
+++ b/src/BlockEntities/JukeboxEntity.cpp
@@ -21,7 +21,6 @@ cJukeboxEntity::cJukeboxEntity(int a_BlockX, int a_BlockY, int a_BlockZ, cWorld 
 
 cJukeboxEntity::~cJukeboxEntity()
 {
-	EjectRecord();
 }
 
 


src/World.cpp
```diff
@@ -3202,8 +3178,8 @@ void cWorld::ScheduleTask(int a_DelayTicks, cTaskPtr a_Task)
 
 void cWorld::AddEntity(cEntity * a_Entity)
```
worktycho: What are the threads that this is called from?

worktycho:
This may fix this immediate issue, but introduces a lot of potential corruption bugs as the entity lists can now be written to from multiple chunks. I do not think this can be merged in its current form.

This one needs considerably more thought on the API's. If we can add a guarantee that the jukebox blockentity is **only** called form the world thread, and the only whilst still attached to the world, then it can add an entity directly. This guarantee needs to be backed up with asserts. I also suggest you create an additional function, AddEntityFromWorldTickThread, with an assert that it is being called the tick thread, so that you don't need to modify the safety of the existing function, and thus all the code that depends on it.

tigerw:
I knew it was too good to be true :)

In that case, I'll just remove the EjectDisc from the jukebox destructor since the ejected disc is lost anyway and the box still saves with a disc inside.

Meanwhile, to eliminate the need for two additional objects (queue + lock) to guard an object accessed from multiple threads, do you think wrapping a container with locks to make it concurrency-aware is a good idea? Or will that be bad for performance?

worktycho:
The problem with that is that entities containers are used incredibly heavily by the tick thread, so using a thread safe container will probably significantly increase that. However without profile evidence I can't be sure either way.

Another thing we could do is make the chunk unload process check the queue and remove entities, bu ultimately there is a race condition in spawning entities, which we might have to ignore.

jammet:
Gosh, I am so glad you're on the right track! Thank you *so* much, guys! So - I gather, from what you're discussing here, it is not safe yet to get a new build and use that now, because of file corruption risk, right?

Should I just - for now - destroy every jukebox in the game?

tigerw:
The newest build is safe, we were just talking about the problem with my initially overly optimistic fix for this issue :P

The problem should hopefully be fixed once this change is accepted and merged, please stand by.

jammet:
Fantastic! It's working! I'm so relieved! And I had to laugh out loud just now because the first new bug I noticed *immediately* after connecting to my good ol' server is ... the squid ... their AI wasn't working properly, remember? They were de-activated so they were completely still.

Now they move again.
They move.
Onto dry land ... burning away... XD.

PR #2462, opened 2015-09-04
Now the player always respawns at the overworld, when he dies.

### LINKED ISSUE #2454: Dying in Nether
When dying in the Nether, you don't respawn in the overworld. You stay in the Nether (and in worst cases you'll keep falling into lava when respawning).

bibo38:
Fixes #2454

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1046,12 +1046,21 @@ void cPlayer::Respawn(void)
 	m_LifetimeTotalXp = 0;
 	// ToDo: send score to client? How?
 
-	m_ClientHandle->SendRespawn(GetWorld()->GetDimension(), true);
+	m_ClientHandle->SendRespawn(dimOverworld, true);
 
 	// Extinguish the fire:
 	StopBurning();
 
-	TeleportToCoords(GetLastBedPos().x, GetLastBedPos().y, GetLastBedPos().z);
+	cWorld *world = GetWorld();
+	if (world->GetDimension() != dimOverworld)
+	{
+		world = cRoot::Get()->CreateAndInitializeWorld(world->GetLinkedOverworldName(), dimOverworld, world->GetLinkedOverworldName(), false);
+		MoveToWorld(world, false, GetLastBedPos());
+	}
+	else
+	{
+		TeleportToCoords(GetLastBedPos().x, GetLastBedPos().y, GetLastBedPos().z);
+	}
 
 	SetVisible(true);
 }


src/Entities/Player.cpp
```diff
@@ -1046,12 +1046,21 @@ void cPlayer::Respawn(void)
...
+	m_ClientHandle->SendRespawn(dimOverworld, true);
 
 	// Extinguish the fire:
 	StopBurning();
 
-	TeleportToCoords(GetLastBedPos().x, GetLastBedPos().y, GetLastBedPos().z);
+	cWorld *world = GetWorld();
+	if (world->GetDimension() != dimOverworld)
+	{
+		world = cRoot::Get()->CreateAndInitializeWorld(world->GetLinkedOverworldName(), dimOverworld, world->GetLinkedOverworldName(), false);
```
NiLSPACE: We should really move away from `CreateAndInitializeWorld` because it's really unsafe. The only reason it's still used is because there is no alternative. I'd say, if there is no proper world then just ignore the action and bail out.

bibo38: What makes the ***CreateAndInitializeWorld*** function unsafe? Should it be possible to disable the overworld?

NiLSPACE: It's unsafe because it can cause a deadlock. @madmaxoft or @worktycho could possibly shed some light on this.

worktycho: As far as I can tell the main issue is that it does a huge amount of work on the thread that calls it. If its called from a tick thread that causes it to block for a couple of minutes triggering the deadlock detector.

bibo38: @NiLSPACE I'm against this bail out idea, because it's inconsistent with the nether portal code and it would only cause trouble, when the server has not loaded the overworld(endless lava deaths...).
Instead a new issue referencing the core <code>CreateAndInitializeWorld</code> problem should be created to fix all calls to this function.

NiLSPACE: But if an admin purposely didn't configure an overworld world linked to the nether world you wouldn't want to see that the server created one when someone died.

madmaxoft: I think the proper solution here would be to simply get the linked world by name, and only move to that world if available. If the world is not available, then it hasn't been configured by the admin (imagine a Nether-only server, for example) and the player should then stay in the Nether.

bibo38: Then we need to change all world moves to cRoot::GetWorld instead of cRoot::CreateAndInitializeWorld.

SphinxC0re:
Any updates on this?

src/Entities/Player.cpp
SafwatHalaby: The conversation @ #2810 sounds related.

LogicParrot:
I'm going to close this and redo it as part of the "phantom worlds" issue.

PR #2463, opened 2015-09-04
Prepared ScheduleTask for threading refactor

tigerw:
How badly did I break it :)

### DIFF
--- a/src/Bindings/ManualBindings_World.cpp
+++ b/src/Bindings/ManualBindings_World.cpp
@@ -374,7 +374,6 @@ static int tolua_cWorld_PrepareChunk(lua_State * tolua_S)
 
 
 class cLuaWorldTask :
-	public cWorld::cTask,
 	public cPluginLua::cResettable
 {
 public:
@@ -384,11 +383,7 @@ public:
 	{
 	}
 
-protected:
-	int m_FnRef;
-	
-	// cWorld::cTask overrides:
-	virtual void Run(cWorld & a_World) override
+	void Run(cWorld & a_World)
 	{
 		cCSLock Lock(m_CSPlugin);
 		if (m_Plugin != nullptr)
@@ -396,7 +391,10 @@ protected:
 			m_Plugin->Call(m_FnRef, &a_World);
 		}
 	}
-} ;
+
+protected:
+	int m_FnRef;
+};
 
 
 
@@ -433,9 +431,9 @@ static int tolua_cWorld_QueueTask(lua_State * tolua_S)
 		return cManualBindings::lua_do_error(tolua_S, "Error in function call '#funcname#': Could not get function reference of parameter #1");
 	}
 
-	auto task = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
-	Plugin->AddResettable(task);
-	self->QueueTask(task);
+	auto ResettableTask = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
+	Plugin->AddResettable(ResettableTask);
+	self->QueueTask(std::bind(&cLuaWorldTask::Run, ResettableTask, std::placeholders::_1));
 	return 0;
 }
 
@@ -483,35 +481,6 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
 
 
 
-class cLuaScheduledWorldTask :
-	public cWorld::cTask,
-	public cPluginLua::cResettable
-{
-public:
-	cLuaScheduledWorldTask(cPluginLua & a_Plugin, int a_FnRef) :
-		cPluginLua::cResettable(a_Plugin),
-		m_FnRef(a_FnRef)
-	{
-	}
-
-protected:
-	int m_FnRef;
-	
-	// cWorld::cTask overrides:
-	virtual void Run(cWorld & a_World) override
-	{
-		cCSLock Lock(m_CSPlugin);
-		if (m_Plugin != nullptr)
-		{
-			m_Plugin->Call(m_FnRef, &a_World);
-		}
-	}
-};
-
-
-
-
-
 static int tolua_cWorld_ScheduleTask(lua_State * tolua_S)
 {
 	// Binding for cWorld::ScheduleTask
@@ -548,11 +517,9 @@ static int tolua_cWorld_ScheduleTask(lua_State * tolua_S)
 		return cManualBindings::lua_do_error(tolua_S, "Error in function call '#funcname#': Could not get function reference of parameter #1");
 	}
 	
-	int DelayTicks = static_cast<int>(tolua_tonumber(tolua_S, 2, 0));
-
-	auto task = std::make_shared<cLuaScheduledWorldTask>(*Plugin, FnRef);
-	Plugin->AddResettable(task);
-	World->ScheduleTask(DelayTicks, static_cast<cWorld::cTaskPtr>(task));
+	auto ResettableTask = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
+	Plugin->AddResettable(ResettableTask);
+	World->ScheduleTask(static_cast<int>(tolua_tonumber(tolua_S, 2, 0)), std::bind(&cLuaWorldTask::Run, ResettableTask, std::placeholders::_1));
 	return 0;
 }
 
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -839,7 +839,6 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
 	TickClients(static_cast<float>(a_Dt.count()));
 	TickQueuedBlocks();
 	TickQueuedTasks();
-	TickScheduledTasks();
 	
 	GetSimulatorManager()->Simulate(static_cast<float>(a_Dt.count()));
 
@@ -962,55 +961,39 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
 
 void cWorld::TickQueuedTasks(void)
 {
-	// Make a copy of the tasks to avoid deadlocks on accessing m_Tasks
-	cTasks Tasks;
+	// Move the tasks to be executed to a seperate vector to avoid deadlocks on accessing m_Tasks
+	decltype(m_Tasks) Tasks;
 	{
 		cCSLock Lock(m_CSTasks);
-		std::swap(Tasks, m_Tasks);
-	}
-
-	// Execute and delete each task:
-	for (cTasks::iterator itr = Tasks.begin(), end = Tasks.end(); itr != end; ++itr)
-	{
-		(*itr)->Run(*this);
-	}  // for itr - m_Tasks[]
-}
-
-
-
-
-
-void cWorld::TickScheduledTasks(void)
-{
-	// Move the tasks to be executed to a seperate vector to avoid deadlocks on accessing m_Tasks
-	cScheduledTasks Tasks;
-	{
-		cCSLock Lock(m_CSScheduledTasks);
-		auto WorldAge = m_WorldAge;
-
-		// Move all the due tasks from m_ScheduledTasks into Tasks:
-		for (auto itr = m_ScheduledTasks.begin(); itr != m_ScheduledTasks.end();)  // Cannot use range-based for, we're modifying the container
+		if (m_Tasks.empty())
 		{
-			if ((*itr)->m_TargetTick < std::chrono::duration_cast<cTickTimeLong>(WorldAge).count())
-			{
-				auto next = itr;
-				++next;
-				Tasks.push_back(std::move(*itr));
-				m_ScheduledTasks.erase(itr);
-				itr = next;
-			}
-			else
-			{
-				// All the eligible tasks have been moved, bail out now
-				break;
-			}
+			return;
 		}
+
+		// Partition everything to be executed by returning false to move to end of list if time reached
+		auto MoveBeginIterator = std::partition(m_Tasks.begin(), m_Tasks.end(), [this](const decltype(m_Tasks)::value_type & a_Task)
+			{
+				if (a_Task.first < std::chrono::duration_cast<cTickTimeLong>(m_WorldAge).count())
+				{
+					return false;
+				}
+				return true;
+			}
+		);
+
+		// Cut all the due tasks from m_Tasks into Tasks:
+		Tasks.insert(
+			Tasks.end(),
+			std::make_move_iterator(MoveBeginIterator),
+			std::make_move_iterator(m_Tasks.end())
+		);
+		m_Tasks.erase(MoveBeginIterator, m_Tasks.end());
 	}
 
-	// Execute and delete each task:
-	for (cScheduledTasks::iterator itr = Tasks.begin(), end = Tasks.end(); itr != end; ++itr)
+	// Execute each task:
+	for (const auto & Task : Tasks)
 	{
-		(*itr)->m_Task->Run(*this);
+		Task.second(*this);
 	}  // for itr - m_Tasks[]
 }
 
@@ -2662,7 +2645,7 @@ void cWorld::UnloadUnusedChunks(void)
 
 void cWorld::QueueUnloadUnusedChunks(void)
 {
-	QueueTask(cpp14::make_unique<cWorld::cTaskUnloadUnusedChunks>());
+	QueueTask([](cWorld & a_World) { a_World.UnloadUnusedChunks(); });
 }
 
 
@@ -3161,42 +3144,32 @@ void cWorld::SaveAllChunks(void)
 
 void cWorld::QueueSaveAllChunks(void)
 {
-	QueueTask(std::make_shared<cWorld::cTaskSaveAllChunks>());
+	QueueTask([](cWorld & a_World) { a_World.SaveAllChunks(); });
 }
 
 
 
 
 
-void cWorld::QueueTask(cTaskPtr a_Task)
+void cWorld::QueueTask(std::function<void(cWorld &)> a_Task)
 {
 	cCSLock Lock(m_CSTasks);
-	m_Tasks.push_back(std::move(a_Task));
+	m_Tasks.emplace_back(0, a_Task);
 }
 
 
-void cWorld::ScheduleTask(int a_DelayTicks, std::function<void (cWorld&)> a_Func)
-{
-	cTaskLambda task(a_Func);
-	ScheduleTask(a_DelayTicks, static_cast<cTaskPtr>(std::make_shared<cTaskLambda>(task)));
-}
 
 
-void cWorld::ScheduleTask(int a_DelayTicks, cTaskPtr a_Task)
+
+void cWorld::ScheduleTask(int a_DelayTicks, std::function<void (cWorld &)> a_Task)
 {
 	Int64 TargetTick = a_DelayTicks + std::chrono::duration_cast<cTickTimeLong>(m_WorldAge).count();
-	
-	// Insert the task into the list of scheduled tasks, ordered by its target tick
-	cCSLock Lock(m_CSScheduledTasks);
-	for (cScheduledTasks::iterator itr = m_ScheduledTasks.begin(), end = m_ScheduledTasks.end(); itr != end; ++itr)
+
+	// Insert the task into the list of scheduled tasks
 	{
-		if ((*itr)->m_TargetTick >= TargetTick)
-		{
-			m_ScheduledTasks.insert(itr, cScheduledTaskPtr(new cScheduledTask(TargetTick, a_Task)));
-			return;
-		}
+		cCSLock Lock(m_CSTasks);
+		m_Tasks.emplace_back(TargetTick, a_Task);
 	}
-	m_ScheduledTasks.push_back(cScheduledTaskPtr(new cScheduledTask(TargetTick, a_Task)));
 }
 
 
@@ -3622,75 +3595,6 @@ void cWorld::AddQueuedPlayers(void)
 
 
 
-
-
-////////////////////////////////////////////////////////////////////////////////
-// cWorld::cTaskSaveAllChunks:
-
-void cWorld::cTaskSaveAllChunks::Run(cWorld & a_World)
-{
-	a_World.SaveAllChunks();
-}
-
-
-
-
-
-////////////////////////////////////////////////////////////////////////////////
-// cWorld::cTaskUnloadUnusedChunks
-
-void cWorld::cTaskUnloadUnusedChunks::Run(cWorld & a_World)
-{
-	a_World.UnloadUnusedChunks();
-}
-
-
-
-
-
-////////////////////////////////////////////////////////////////////////////////
-// cWorld::cTaskSendBlockToAllPlayers
-
-cWorld::cTaskSendBlockToAllPlayers::cTaskSendBlockToAllPlayers(std::vector<Vector3i> & a_SendQueue) :
-	m_SendQueue(a_SendQueue)
-{
-}
-
-void cWorld::cTaskSendBlockToAllPlayers::Run(cWorld & a_World)
-{
-	class cPlayerCallback :
-		public cPlayerListCallback
-	{
-	public:
-		cPlayerCallback(std::vector<Vector3i> & a_SendQueue, cWorld & a_CallbackWorld) :
-			m_SendQueue(a_SendQueue),
-			m_World(a_CallbackWorld)
-		{
-		}
-
-		virtual bool Item(cPlayer * a_Player)
-		{
-			for (std::vector<Vector3i>::const_iterator itr = m_SendQueue.begin(); itr != m_SendQueue.end(); ++itr)
-			{
-				m_World.SendBlockTo(itr->x, itr->y, itr->z, a_Player);
-			}
-			return false;
-		}
-
-	private:
-
-		std::vector<Vector3i> m_SendQueue;
-		cWorld & m_World;
-
-	} PlayerCallback(m_SendQueue, a_World);
-
-	a_World.ForEachPlayer(PlayerCallback);
-}
-
-void cWorld::cTaskLambda::Run(cWorld & a_World)
-{
-	m_func(a_World);
-}
 
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/Blocks/BlockPiston.cpp
+++ b/src/Blocks/BlockPiston.cpp
@@ -27,7 +27,6 @@
 		} \
 	}
 
-#define PISTON_TICK_DELAY 1
 #define PISTON_MAX_PUSH_DISTANCE 12
 
 
@@ -156,15 +155,12 @@ void cBlockPistonHandler::ExtendPiston(int a_BlockX, int a_BlockY, int a_BlockZ,
 	// Push blocks, from the furthest to the nearest:
 	int oldx = a_BlockX, oldy = a_BlockY, oldz = a_BlockZ;
 	NIBBLETYPE currBlockMeta;
-	std::vector<Vector3i> ScheduledBlocks;
-	ScheduledBlocks.reserve(PISTON_MAX_PUSH_DISTANCE);
 
 	for (int i = dist + 1; i > 1; i--)
 	{
 		AddPistonDir(a_BlockX, a_BlockY, a_BlockZ, pistonMeta, -1);
 		a_World->GetBlockTypeMeta(a_BlockX, a_BlockY, a_BlockZ, currBlock, currBlockMeta);
-		a_World->SetBlock(oldx, oldy, oldz, currBlock, currBlockMeta, false);
-		ScheduledBlocks.push_back(Vector3i(oldx, oldy, oldz));
+		a_World->SetBlock(oldx, oldy, oldz, currBlock, currBlockMeta);
 		oldx = a_BlockX;
 		oldy = a_BlockY;
 		oldz = a_BlockZ;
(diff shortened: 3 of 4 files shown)


worktycho:
Are we really ready for c++14? Given that it is only supprted in gcc >4.9 and clang  >3.4 I think we need to wait until the compilers are more widely deployed.

tigerw:
Well, are there any stats on how many people who wish to compile need to manually install a newer compiler anyway? Debian wheezy at least has clang 3.0 and gcc 4.7 (so it'll just be a version difference for them) and jessie meets both requirements. I guess the main problem is people who have already set up their build environment with something other than the latest having to upgrade.

For example, @madmaxoft, are you willing to update to VS 2015 or be unable to compile the server?

madmaxoft:
I think VS2015 should be able to install on Win7, so I should be ok. Not sure about the compilers for the more exotic architectures, though - Android, RasPi etc.

Still, I think moving to C++14 would be too rushed at this point.

tigerw:
Alright. So, do people think this will work?

src/Entities/Entity.cpp
```diff
@@ -1451,7 +1450,7 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	}
 
 	// Remove all links to the old world
-	SetWorldTravellingFrom(GetWorld());  // cChunk::Tick() handles entity removal
+	GetWorld()->GetChunkMap()->RemoveEntity(this);
```
worktycho: What thread does this get called from?

worktycho:
What are the contention levels like on the locks?

src/Entities/Entity.cpp
tigerw: I believe the world tick thread and whatever thread lets plugins run binding code.

tigerw:
I'm not quite sure, wouldn't that depend on server load? How would you suggest I go about collecting this data?

worktycho:
You'd need a profilier that can tell you it, but I'm not sure what the best one to do that is, as I've only used gprof which doesn't tell you that. However Intel VTune gets a lot of good coverage, claims to support lock profiling, and has an free linux version for open source projects.

tigerw:
I did a concurrency profile with VS, and the locks in question didn't even show up. That seems to suggest that its effects are very small, but it might be that 1) I'm using the profiler wrong or 2) There's not enough load on the server.

Can you open VS profile files?

worktycho:
Not at the moment. Did you connect to the server, or just start it?

tigerw:
I connected and went through a portal a couple of times both ways.

worktycho:
Ok, but I think we need to keep an eye on the locks for performance issues.

tigerw:
So is that a yes?

worktycho:
I still haven't properly reviewed it for race's. `cEntity *`'s can be held by other parts of the code at the moment so long as they only access it through the tick thread. This breaks the safety of that so you need to change all of those pointers to go through the lock.

tigerw:
I believe all manipulation of retrieved entity pointers are done within the lock acquired before cChunk calls Tick() on each, so surely other threads can't delete an entity whilst they are being used?

worktycho:
That is the the state before this patch. I assumed that you added the ability to manipulate entity pointers outside the cChunk lock, because otherwise, why do you need a separate lock for the list?

tigerw:
Because I was thinking of moving across worlds, where one world thread adds an entity to another world thread, which needs synchronisation.

madmaxoft:
Moving across worlds was done through buffers - the entity was moved from the original world to destination world's "add" buffer within original world's tick thread, and then the destination world's tick thread would move the entity from its "add" buffer into the world. It cannot be done in one step, because that would produce a lock dependency and thus a deadlock possibility.

tigerw:
nooooooooooo

I'm debating two solutions:
1. To extend manual_lock_container with a queue. Inserts to the underlying container should go through a staged_push_back method. If a try_lock call on the mutex guarding the underlying container acquired it, the new object is pushed into the container. Otherwise, it is pushed onto the internal queue. This queue is processed by a separate thread, which, when alerted by a condition_variable set by the thread pushing onto the queue, calls std::lock on both mutexes (guarding the queue and underlying container), and when the locks are both successfully acquired, inserts the queue into the relevant data structure. The problem with this approach is that it adds a lot of complexity. The benefits are faster (but not guaranteed) pushing of entities into cChunk, and a reduction in work by the tick thread.
2. To changed manual_lock_container to a simple class encapsulating a container, a queue, and a lock guarding the queue, similar to what we have now, just as a class. The adder calls a method which pushes to the queue, and the tick thread calls a method which processes the queue.

I'm torn. What would you suggest? I guess there is really no way to avoid the deadlock of the current attempt?

tigerw:
[Concurrency...](http://treasure.diylol.com/uploads/post/image/336787/resized_y-u-no-meme-generator-concurrency-y-u-so-hard-78948c.jpg)

worktycho:
Its not difficult if you write pure functions, or keep state per thread. Its only concurrent access to shared state that's difficult to work with. :) But seriously, shared state is hard, so we should try and avoid sharing things where possible. (Queues are relatively simple, and we have a generic thread safe queue).

tigerw:
Okay, I have another idea. Since problems are arising due to the need for other threads to concurrently access one world tick thread, it makes sense therefore to limit direct access to a cWorld, and any variables that are a member of it, to solely one thread. This could be partially enforced by not allowing cRoot to provide any world object. If one world wished to modify something in another, it must post a lambda (via cRoot) to the world in question (with variables capture by-value), which is then executed in the context of the tick thread.

Therefore, if everything is done right, I think this will eliminate the need completely for individual locks, except those which are directly exposed to concurrency.

This should work in theory, right? If so, does anyone know which threads currently interface with a world thread? Plugin hooks are executed synchronously, and network threads post to a queue, which is processed when clienthandles are ticked. So maybe the webadmin thread?

worktycho:
No as per the following hole:

    postFunction([=worldPtr] (cWorld a_World) { /* copy from worldPtr to a_World */});

As far as I know, there is no way of using compile time interfaces to solve this problem in c++.

tigerw:
What do you mean?

worktycho:
as in whats to stop you posting a world pointer to another thread in a lambda.

tigerw:
Whomever reviews the PR? :P
What stops me from forgetting to guard a new concurrently accessed vector with a lock? I can't really protect the code against the programmer, but if they try to pass a pointer in that fashion, at least it becomes obvious that something's wrong.

worktycho:
And what about entity transfers? You need to ensure that thy are removed from on world before the post. We already review for thread safety, and c++ cannot provide the guarantees needed for static checks. There are scenarios where you don't have access to a cWorld object, but know you're on the right thread. So ScheduleTask should handle the cross thread execution issues.

tigerw:
So for entity transfers, remove the entity entry from the old world and post the pointer onto the task queue of the new one, or, if cChunk owns smart pointers of entities, call a function which moves the ownership to the task lambda and deletes the entry afterwards.

Well, it would be a very interesting situation if one were processing on the world tick thread without any way to get a world pointer. If that is the case, one should really modify the arguments of whatever they're in to pass one.

madmaxoft:
Also, we can't really lock the world access to the tick thread:
- it would create deadlocks with plugins executing world functions in the webadmin
- it would break the API for all plugins

tigerw:
It would break plugins, since cRoot:GetWorld() can't exist, but this in theory is only currently used in the webadmin handler, where no world object is available, so changing it in every plugin is always do-able.

With the webadmin handler having to do work by passing a lambda to ScheduleTask, this will be guarded by a lock. But, since this won't depend on any other locks, where is the deadlock?

(forgive me if it's obvious, I'm still learning :P)

madmaxoft:
It's used in most plugins, notably in the webadmin and console command handlers.

The deadlock is caused by the plugin code itself being guarded by a lock. Thus, a webadmin page goes to a Lua plugin, which tries to execute task on a world, but the world tick thread is waiting for the plugin to execute its hook.

Note that in order for most things to continue working, you need not only `ScheduleTask`-like function, but also `DoWith`-like function (immediate execution).

tigerw:
I don't see how there could be a deadlock. A webadmin thread executing plugin code will not be given direct access to a cWorld, and therefore won't try to lock it. Instead, it should schedule a task (which depends on only one lock) via cRoot to execute within the context of the world thread. I just realised that I'm describing [this](http://mc-server.xoft.cz/LuaAPI/WebWorldThreads.html).

Meanwhile, plugins which are handling a hook will be doing so in the context of the world tick thread. By routing all extra-worldthread modifications through a schedule, all locks + the DoWith functions can be removed, since everything (entity creation/destruction, data (structure) changes, block changes, etc.) is serialised and confined to one thread.

In short, what I'm proposing is that:
* Only one thread changes a world directly
* Everything else is queued
* No more DoWith or providing world pointers
* No more locks in a world or anything that it contains, since everything is queued

Is it gonna work?

madmaxoft:
It's not gonna work for plugins that need to interact with the world immediately (`DoWith`-style function). It would work only for postponed access (`ScheduleTask`-style function), but that would severely limit the abilities of plugins. For example webadmin functions need to return the entire HTML code immediately, they can't "schedule".a query and produce HTML based on that query.

tigerw:
I think it will work for every place that's using DoWith now, since
* They are in the tick thread
* They have a world object
* So, they can access the world directly.

For the webadmin thread, according to [this](http://mc-server.xoft.cz/LuaAPI/WebWorldThreads.html) again, they aren't supposed to do anything related to unqueued access (ForEach, DoWith) because of what you said about deadlocks:
*  Webadmin: Plugin locked -> plugin incorrectly does a ForEach -> thread switch
*  World: Layers locked -> Tick -> hook handling -> plugin lock **blocked** -> thread switch
*  Webadmin: Layers lock **blocked**

madmaxoft:
True, those deadlocks are a PITA. Before that could be done, though, we need a "resumable webadmin request" and "resumable command handler". That is, allow the handler to say "I have scheduled some work, the request / command will be finished later on from a callback" and the server will keep the request / command dangling until that callback finishes it.

tigerw:
So does that mean the single thread world access is going to work? :D

I don't know about delaying the webadmin until a scheduled task finishes, but I'm just trying to argue that I think my proposals won't break anything.

madmaxoft:
It could be made to work, yes. Yes, it will break things, but there is a workaround for those. The question is, do you want to invest that much work into something when you have very little expectancy of gains?

tigerw:
Well, hopefully we can:

1. Eliminate all locks in cWorld (apart from the ScheduleTask bit) since nothing should be accessed out-of-thread
2. Eliminate all queuing, again, since other threads aren't trying to read/write concurrently.

I think even if there aren't many performance benefits, at least it makes the code clearer and smaller. And there won't be as many iterations of queues, so there's that.

tigerw:
Why doesn't GitHub do automatic merges if the master is ahead but non-conflicted any more?

madmaxoft:
Because we have required CI builds for the master branch.
I personally like that, because it prevents the tangled history. You should rebase before merging.

worktycho:
Its an option I've turned on. I think it comes with not letting you merge unless CI has passed the build.

SetFlags.cmake
```diff
@@ -62,6 +62,9 @@ macro(set_flags)
 		set(CMAKE_EXE_LINKER_FLAGS_RELEASE    "${CMAKE_EXE_LINKER_FLAGS_RELEASE}    /LTCG")
 		set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /LTCG")
 		set(CMAKE_MODULE_LINKER_FLAGS_RELEASE "${CMAKE_MODULE_LINKER_FLAGS_RELEASE} /LTCG")
+
+		# Make build use Unicode:
+		add_definitions(-DUNICODE -D_UNICODE)
```
madmaxoft: Again: WHY? This has the potential to break many things and is not strictly needed for your goals.

src/Bindings/ManualBindings_World.cpp
```diff
@@ -380,9 +378,9 @@ static int tolua_cWorld_QueueTask(lua_State * tolua_S)
 		return cManualBindings::lua_do_error(tolua_S, "Error in function call '#funcname#': Could not get function reference of parameter #1");
 	}
 
-	auto task = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
-	Plugin->AddResettable(task);
-	self->QueueTask(task);
+	auto ResettableTask = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
+	Plugin->AddResettable(ResettableTask);
+	self->QueueTask(std::bind(&cLuaWorldTask::Run, ResettableTask, std::placeholders::_1));
```
madmaxoft: I still prefer the old code style. It was clear and readable, unlike this mess of quadra-dots and a `std::bind` which is always a bad sign.

SetFlags.cmake
worktycho: And if anything it should be in Globals.h along with all the other windows flags.

src/Bindings/ManualBindings_World.cpp
tigerw: I understand your concern, but wouldn't you say that one messy line is better than one class in World.h for every action which you want to perform with QueueTask? `std::bind` is the proper way to achieve this, despite that it doesn't look nice.

worktycho: I agree that this weird hybrid of lambda's and strategy objects is a bad idea. Particularly as your taking a member function pointer and treating it as a function pointer.

What would be a much better way of doing this would be to take advantage of the fact that std::function will take anything that meets the Callable concept. So if you replace the Run method with an override of operator(), you should be able to construct a std::function from it.

SetFlags.cmake
tigerw: Nothing appears to be untowardly broken, so why not? Unicode all the way, right?

I don't know enough to be able to respond to the Globals.h comment though, so okay, I guess.

src/Bindings/ManualBindings_World.cpp
tigerw: Really? I'll try that.

tigerw:
you guys are such nitpickers :)

SetFlags.cmake
worktycho: The big problem is that windows uses UCS-2(ish) whereas we use UTF-8. Now UTF-8 can be munged to windows codepage and back fairly safely. But UCS-2(ish) is really easy for UTF-8 systems to munge.

madmaxoft: Turning to unicode (especially on a single platform out of several) is a huge undertaking, it's not something you can just flip a switch and fix the few compiler warnings. The errors will be abundant and they will be subtle. Better not to do it in such a hurry.

src/Bindings/ManualBindings_World.cpp
madmaxoft: You still have one class for each action that is to be scheduled, so there's no real improvement here - only that your construct doesn't inherit from a common predecessor.

src/OSSupport/Errors.cpp
```diff
@@ -5,11 +5,11 @@
...
 AString GetOSErrorString( int a_ErrNo)
 {
-	char buffer[ 1024 ];
 	AString Out;
 
 	#ifdef _WIN32
 
+	wchar_t buffer[1024];
 	FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, nullptr, a_ErrNo, 0, buffer, ARRAYCOUNT(buffer), nullptr);
 	Printf(Out, "%d: %s", a_ErrNo, buffer);
```
worktycho: This is broken because %s is for single byte encoding unit strings, not multi-byte encoding unit strings.

src/Bindings/ManualBindings_World.cpp
tigerw: Yes, lambdas still do create a class, but I was intending to convey that it is much less effort now that before, and World.h is that much smaller.

worktycho: The improvement is that it lets us use lambda methods elsewhere.

SetFlags.cmake
tigerw: I don't quite understan the munging bit. Why would "other systems" be affected by a flag and changes limited to Windows compilation?

Also, what do you mean by "we use UTF-8"? Do you mean the encoding of the source files?

worktycho: By we use utf-8, I'm referring to what our strings are encoded with. So all of our code expects a std::string to be in an encoding that uses single byte encoding units. In practice this means utf-8 or a windows code-page, but for a lot of string manipulation it doesn't actually matter which. 

UCS-2/UTF-16 is different enough that it can cause serious issues if it is passed to a function that expects a single byte encoding unit encoding. For example UCS-2/UTF-16 can have single byte embedded zero bytes in the middle of text, which breaks strcpy.

worktycho: And if you don't understand what an encoding unit is, you shouldn't be making this change.

tigerw: Well, I don't understand how std::function works either, doesn't mean I shouldn't experiment :)

Am I correct in thinking that, even under UNICODE, strings, both literal and dynamic, will only become wide if prefixed by `L` or given a type `wchar_t`? Because the server ran, printed messages, parsed recipes, etc. perfectly fine.

src/Bindings/ManualBindings_World.cpp
tigerw: I can't quite get std::fuction to work. Do you have any ideas that are one-liners?

It still seems to me the std::bind is the best way. The complaints thus far has been entirely of aesthetics, convention, and a fear of bind - do correct me if I'm wrong here. If it works, and it does excellently, why not use it?

SetFlags.cmake
worktycho: Not understanding how std::function works is fine so long as you understand the API. encoding units is part of the API text encodings present.

And you are correct about the L prefix. The problem is that we don't use the windows API much, so testing this involves things like purposeful crashes.

src/Bindings/ManualBindings_World.cpp
worktycho: Whats code you've written and the error? The signature of operator() should be `void operator()(cWorld & a_World);`

-----
My main worry with std::bind is that member function pointers are a mess of UB, and like varargs, often appear to work when UB.

tigerw: But I think the current state is correct? It's unlikely to change also.

Attempt:
`cpp
class cLuaWorldTask :
	public cPluginLua::cResettable
{
public:
	cLuaWorldTask(cPluginLua & a_Plugin, int a_FnRef) :
		cPluginLua::cResettable(a_Plugin),
		m_FnRef(a_FnRef)
	{
	}

	void operator()(cWorld & a_World)
	{
		cCSLock Lock(m_CSPlugin);
		if (m_Plugin != nullptr)
		{
			m_Plugin->Call(m_FnRef, &a_World);
		}
	}

protected:
	int m_FnRef;
};

Method1
{
	auto ResettableTask = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
	std::function<void(cWorld &)> CallableTask(std::placeholders::_1);
	CallableTask = ResettableTask;

	Plugin->AddResettable(ResettableTask);
	self->QueueTask(CallableTask);
}

Method2
{
	auto ResettableTask = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
	std::function<void(cWorld &)> CallableTask = ResettableTask;

	Plugin->AddResettable(ResettableTask);
	self->QueueTask(CallableTask);
}
```

`Severity	Code	Description	Project	File	Line
Error	C2893	Failed to specialize function template 'unknown-type std::invoke(_Callable &&,_Types &&...)' (compiling source file C:\Users\Tiger\Documents\GitHub\cuberite\src\Bindings\ManualBindings_World.cpp)	MCServer	C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\type_traits	1494
`

Comment (worktycho) [inline src/Bindings/ManualBindings_World.cpp]: Ah. The problem is that you're passing a shared_ptr, not the cLuaWorldTask. And looking into it, fixing this is more complicated because of cResettable is a non-copyable type. A simple workaround is to use lambda for the bind, which is probably clearer.

Comment (tigerw) [inline src/Bindings/ManualBindings_World.cpp]: I can haz example?

Comment (worktycho) [inline src/Bindings/ManualBindings_World.cpp]: `self->QueueTask([&ResettableTask] (cWorld & a_World) {ResettableTask->Run(a_World)});`
Or we can make cRessetable more supportive of the functional style, by adding:
```
template<class F>
void TryDoWithPlugin(F callback)
{
     cCSLock Lock(m_CSPlugin);
     if (m_Plugin != nullptr)
     {
          F(m_Plugin);
     }
}
```

Comment (tigerw) [inline src/Bindings/ManualBindings_World.cpp]: \*sigh\*

You and @madmaxoft sure? Bind was exactly meant for this and the llama introduces an extra intermediate class.

Comment (worktycho) [inline src/Bindings/ManualBindings_World.cpp]: The point of TryDoWithPlugin would be to move the logic of cLuaWorldTask into the lambda. eg: 
```
self->QueueTask(
    [&Resettable, =FnRef] (cWorld & a_World)
    {
        Resettable->TryDoWithPlugin([&a_World, =FnRef] (cPlugin & a_Plugin)
        {
             a_Plugin->Call(FnRef, a_World);
        })
    });
```
But maybe this is too far down the road to callback hell.

Comment (tigerw) [inline src/Bindings/ManualBindings_World.cpp]: Yeah, probably. Are you like, absolutely sure about bind? I'm still not convinced.

Comment (worktycho) [inline src/Bindings/ManualBindings_World.cpp]: This is the problem with functional code in imperative languages.
Maybe as SphinxC0re suggested on IRC what we really want is to be able to go full FRP with something like:
```
auto Task = self->QueueTask();
Task.flat_map([&Resettable](cWorld & a_World)
    {
         return Resettable.TryDoWithPlugin();
    })
    .zip(Task)
    .then([=FnRef](std::tuple<cPlugin, cWorld & a_args)
    {
         (a_args.first)->Call(FnRef, a_args.second);
    });

worktycho: Or another option:
Resurface locked_ptr, which should be simple to implement with a custom deleter and unique_ptr.
Then we can get rid of the inner callback in the callback hell one.

tigerw: Perhaps we are over-engineering this somewhat. Could we hold the idea for another time and just use std::bind? It seems the easiest and most performant option.

SetFlags.cmake
tigerw: I can confirm that a simulated crash writes a core dump file correctly. I can further confirm that on changing the format string in Errors.cpp, errors are printed correctly.

This good?

tigerw:
No farther communication, assumed acceptable?

worktycho:
I was hoping to get @madmaxoft 's opinion on the bind alternatives.

src/World.cpp
```diff
@@ -962,55 +961,39 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
...
-			else
+			return;
+		}
+
+		// Partition everything to be executed by returning false to move to end of list if time reached
+		auto MoveBeginIterator = std::partition(m_Tasks.begin(), m_Tasks.end(), [this](const decltype(m_Tasks)::value_type & a_Task)
 			{
-				// All the eligible tasks have been moved, bail out now
-				break;
+				if (a_Task.first <= std::chrono::duration_cast<cTickTimeLong>(m_WorldAge).count())
```
worktycho: Should Tasks execute after x Ticks time or after x Ticks have been executed? The current code is executed after x Ticks time. It can cause subtle differences when lagging and I suspect that the current cod is not the expected behavior.

tigerw: Indeed, should have thought about that more thoroughly. Many thanks!

madmaxoft: I believe we should aim to process all game logic in logical ticks instead of wall clock time
If the ticks take too long, the server could limit the actions that are processed somehow, but delaying tasks is not a good way of doing so, especially if such tasks belong to core logic.

madmaxoft:
As for the `bind` alternatives: I personally really dislike the "full FRP" style. It's too wordy and yet it doesn't easily say what it's doing.

So far I like this code the most:
```cpp
self->QueueTask(
    [&Resettable, =FnRef] (cWorld & a_World)
    {
        Resettable->TryDoWithPlugin([&a_World, =FnRef] (cPlugin & a_Plugin)
        {
             a_Plugin->Call(FnRef, a_World);
        })
    });
```
Since the call into a plugin would be one of the more common operations that will be queued / scheduled, it would make sense to get rid of the inner lambda by writing a templated `cResettable::Call(...)` function.

src/World.cpp
worktycho: The reason the code is like this is we don't have a good measure of world age in ticks at the moment. m_WorldAge is the only measure we have and its wall clock age.

worktycho:
Something along the following lines would work:
```
template <class... Args>
void cResetable::CallOnWorld(cWorld & a_World, Args&&... a_args)
{
	a_world.QueueTask(
		[=this, =FnRef, &a_args...] (cWorld & a_World)
		{
			cCSLock Lock(m_CSPlugin);
			if (m_Plugin != nullptr)
			{
				m_Plugin->Call(m_FnRef, &a_World, std::forward(a_args)...);
			}
		});
}
```
By moving the method into cResetable, we remove the need for TryDoWithPlugin, meaning we only need one lambda.

madmaxoft:
Okay, that looks reasonable.

tigerw:
We unfortunately have a bit of an impasse here because this
```cpp
template <class... Args>
void cResetable::CallOnWorld(cWorld & a_World, Args&&... a_args)
{
	a_world.QueueTask(
		[=this, =FnRef, &a_args...] (cWorld & a_World)
		{
			cCSLock Lock(m_CSPlugin);
			if (m_Plugin != nullptr)
			{
				m_Plugin->Call(m_FnRef, &a_World, std::forward(a_args)...);
			}
		});
}
```
was not a change I had in mind to make when I decided upon this
```cpp
std::bind(&cLuaWorldTask::Run, ResettableTask, std::placeholders::_1)
```

I still don't have a reason why bind shouldn't be used.

worktycho:
@tigerw Sometimes (actually quite frequently in this code base) you find that one refactor needs another to work well. I sometime find myself so deep in dependent refactors that I forget what the original refactor was. For example #2488 was required for a protocol refactor I'm working on, which is required for #2477.
I personally think xofts suggestion is the best way of handling this. If you don't think it should go in this PR, I'll put it in another one. We can then rebase this PR on to it.

tigerw:
I see what you're saying, but in this case, *why* are the suggestions necessarily better?

worktycho:
I'm looking at the code again, and fundamentally the problem is that cRessettable's interface doesn't work well with the functional style. In the functional style you're going for cLuaWorldTask shouldn't exist. So the functional way would be to go with my TryDoWithPlugin method.

It doesn't look to me like CallOnWorld belongs in cResettable. I'm wondering if a better way to handle this is to refactor cRessable into a non-owning atomic smart_ptr. I'm just going to look at the concurrency TS to see what atomic_weak_ptr is going to look like.

worktycho:
Atomic_shared_ptr did not give me any insperation, and I'm thinking the locked pointer option is the best. It's the only one I haven't written up yet, so I'll try and remember to do that tomorrow.

tigerw:
Do I detect reservations? :P

The thought of introducing atomicity or further locking into this, or even, as you say, contaminating cResettable with task scheduling code, is not filling me with confidence of the practicality of any of this.

If cLuaWorldTask isn't functional, then fine. Should we really stick vigorously to convention and ideas about what code *should* be, at the expense of clarity?

worktycho:
The reason I care about cLuaWorldTask not being functional, is that the whole problem here is that you're trying to fit a strategy object into a space for function. If we can fix that the whole complication with bind disappears.

tigerw:
Is fitting a strategy object into the space for a function bad? And bind isn't really that complicated.

worktycho:
its messy, and the reason you're doing it is a pattern that comes up frequently, so its worth sorting out.

tigerw:
Well, alright.

So merge first and do whatever you want later?

worktycho:
I think we should clean the code up as it comes in, because otherwise cleanup PR's don't happen that much.

tigerw:
But [this guy (point 3)](http://hintjens.com/blog:95) says that open source projects should merge first and ask questions later!

worktycho:
Yes, and that was how we ended up with god objects in the first place.

tigerw:
You could also argue that a strict approvals process is how there are 21 open pull requests, most abandoned, and that this isn't exactly beneficial to a open source project aiming to replicate a server for a rapidly-evolving game, as can be seen by our very erratic new changes/features rate.

Please let's merge if there isn't anything really serious?

worktycho:
And Neither is that fact that we have a number of crash bugs originating in race conditions. Most of which are really hard to find due to confusing control flow or ownership. Its a matter of which is more important: feature parity, or the ability to run for a length of time without crashing.

tigerw:
Well, this PR and subsequent others are planning to simplify and hopefully fix all of that :)

worktycho:
Good luck sorting that. Also at this point I think I'd merge this if you replaced std::bind with a lambda that did the same thing, as personally I think its clearer. Otherwise, I think we need @madmaxoft's opinion on what to do as our discussion is just degenerating into an argument about the importance of code quality.

madmaxoft:
Get rid of the Unicode change, I'm not letting that in without serious testing.

As for the bind, I think I can live with it, despite grinding my teeth about it. The code could use a bit more commenting, though.

And change the commit message header to something more descriptive.

SphinxC0re:
:pray: thanks for that

PR #2465, opened 2015-09-06
Added BungeeCord support for Protocol18x

electromatter:
I copied the changes from Protocol17x to Protocol18x to support BungeeCord with 1.8.x clients.


madmaxoft:
Can anyone with BungeeCoord test this? It seems okay, but I'd still like a confirmation.

SphinxC0re:
Any update on this. I think PRs shouldn't be lying around for so long. Be should have a close policy

PR #2473, opened 2015-09-13
APIDump: Fixed ZBS integration filenames after project rename.

madmaxoft:
The API filename has been changed, a well as the article for setting ZBS up.


NiLSPACE:
Looks good to me.

PR #2477, opened 2015-09-15
Always tick clients in the server

worktycho:
Always call cClientHandle's Tick methods from cServer, and never the world.

This simplifies the threading of cClientHandle.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -175,6 +175,7 @@ void cClientHandle::Destroy(void)
 	{
 		m_Player->RemoveClientHandle();
 	}
+	m_Player = nullptr;
 	m_State = csDestroyed;
 }
 
@@ -1679,27 +1680,12 @@ bool cClientHandle::CheckMultiLogin(const AString & a_Username)
 	}
 	
 	// Check if the player is waiting to be transferred to the World.
-	if (cRoot::Get()->GetServer()->IsPlayerInQueue(a_Username))
+	if (cRoot::Get()->GetServer()->IsClientOnServer(a_Username))
 	{
 		Kick("A player of the username is already logged in");
 		return false;
 	}
 
-	class cCallback :
-		public cPlayerListCallback
-	{
-		virtual bool Item(cPlayer * a_Player) override
-		{
-			return true;
-		}
-	} Callback;
-
-	// Check if the player is in any World.
-	if (cRoot::Get()->DoWithPlayer(a_Username, Callback))
-	{
-		Kick("A player of the username is already logged in");
-		return false;
-	}
 	return true;
 }
 
@@ -1887,6 +1873,17 @@ void cClientHandle::Tick(float a_Dt)
 			Link->Send(OutgoingData.data(), OutgoingData.size());
 		}
 	}
+
+
+	if (m_State == csAuthenticated)
+	{
+		StreamNextChunk();
+		
+		// Add the player to the world (start ticking from there):
+		m_State = csDownloadingWorld;
+		m_Player->GetWorld()->AddPlayer(m_Player);
+		return;
+	}
 	
 	m_TicksSinceLastPacket += 1;
 	if (m_TicksSinceLastPacket > 600)  // 30 seconds time-out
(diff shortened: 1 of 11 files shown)


tigerw:
-218 lines: YES @worktycho!

worktycho:
I'm confused. Why would checkBasicStyle only fail in some jobs?

tigerw:
CheckBasicStyle ain't failing; Travis is failing sometimes to locate gcc-4.8, and so either cmake fails at configuring, or lua doesn't get installed because the whole apt command failed.

Why that's happening?

I'm confused. Why would apt in Travis containers only fail in some jobs?

src/ClientHandle.cpp
```diff
@@ -2163,7 +2112,7 @@ void cClientHandle::SendChatSystem(const cCompositeChat & a_Message)
 
 void cClientHandle::SendChunkData(int a_ChunkX, int a_ChunkZ, cChunkDataSerializer & a_Serializer)
 {
-	ASSERT(m_Player != nullptr);
+	if (m_Player != nullptr) return;
```
madmaxoft: Bad style. Add braces and put on separate lines.

madmaxoft:
This produces an even bigger problem: The packet decoding is done in the `cClientHandle::Tick()` function, which means that you've now moved all packet decoding into the server tick thread. Unfortunately, typically the packet processing accesses the world *a lot*, so it produces a lot of contention on the tick thread.

Not to mention that from the game logic point of view, all changes in the game state should occur within the single tick. Now the packet processing breaks that, because it can change the player state in a non-determinate moment inside the game tick.

I'm against this change at any form.

worktycho:
I didn't realise about the performance impact of protocol decoding.
This is problematic as we need to get rid of accessing clienthandles by the world thread in order to make their destruction understandable.

worktycho:
The only solution I can think of at the moment is to split world decoding into a separate  object.

worktycho:
@madmaxoft Would you accept this PR if I moved packet decoding out of cClientHandle, and into a separate object?

madmaxoft:
How do you propose to do that and what would the threads accesses be? I don't see a way to untangle this reasonably.

worktycho:
how about:
```
Network data comes in -> Decode to cPacket objects
 -> Queue in a cStateUpdater, which is owned by cPlayer -> Applied in cPlayers Tick Method.
```

By applying in a different thread from decoding, we remove the issue of moving the decoder across threads. (And as for having a large number of cPacket Classes, I have a solution involving higher order macros).

tigerw:
As per my idea in #2463, you will do whatever decoding you want in another thread, and just ScheduleTask the results onto the tick thread to be processed. So it satisfies that:

1. All changes happen in a tick
2. There is no contention for locks

madmaxoft:
/me likes, except for the class name `cPacket`. Use something else. How about a `cClientAction`?

worktycho:
That's fine. cPacket was only a provisional name anyway.

madmaxoft:
That way you can say the `cProtocol` descendants parse network packets into `cClientAction` instances that get executed in the world thread. Sounds nice and everyone should know what's going on :)

tigerw:
So current cProtocol::DataReceived both decodes the data string into packets and executes the actions they represent. Are you saying you're gonna split cProtocol180 into something which creates a cClientAction for every packet, which stores (a lambda?) to execute?

madmaxoft:
That's more or less the idea. I'm a bit afraid of the allocation side of this, what if it becomes a bottleneck?

worktycho:
Then we use an allocation pool. Since most of the actions will be destroyed at similar times, this seems lie a good candidate for a slab allocator.

tigerw:
Oh noes, not the roll-our-own movement! :P

If the cClientAction class sizes are going to be more or less the same size, does that mean the allocator will have no problem finding a suitable place in memory to allocate a new one once an old one is gone?

worktycho:
Whoops, I confused a slab allocator with an arena allocator. What I meant was that we could have arenas which are either reused or just used to bunch the allocations.

worktycho:
I've just discovered cProtocol172::DataRecived and cProtocol180::DataRecived have identical implementations. I'd like to unify them along with some of the other initial packet splitting code. Should I create a new cEncryptedProtocol class to hold the logic, or move it up to cProtocol. At the moment I'm leaning on moving to cProtocol, but would that cause problems for future protocols?

madmaxoft:
It used to be in a special mid-descendant, `cLengthedProtocol` that serviced the protocols that use packet lengths (1.7+), unlike the previous versions, where each packet had implicit length and had to be decoded on its own before the next packet could be started.

I'd be okay with the move to `cProtocol`, since it's quite unlikely that Mojang would change this any time soon, and if they do, we can still create the mid-descendant again.

worktycho:
That's where I was going. I just wanted confirmation from someone else before sinking hours into rewriting cProtocol.

SphinxC0re:
@worktycho are you still working on this?

worktycho:
I'm working on the protocol refactor that needs to happen before this can be merged.

PR #2479, opened 2015-09-16
Removed a significant performance issue.

worktycho:
Iterating through the list of clients in chunks was taking up a significant
amount of time with larger numbers of clients due to processor stalls.
Changing the data structure to a vector fixed the issue.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -199,9 +199,9 @@ void cChunk::MarkRegenerating(void)
 	SetPresence(cpQueued);
 
 	// Tell all clients attached to this chunk that they want this chunk:
-	for (cClientHandleList::iterator itr = m_LoadedByClient.begin(); itr != m_LoadedByClient.end(); ++itr)
+	for (auto ClientHandle : m_LoadedByClient)
 	{
-		(*itr)->AddWantedChunk(m_PosX, m_PosZ);
+		ClientHandle->AddWantedChunk(m_PosX, m_PosZ);
 	}  // for itr - m_LoadedByClient[]
 }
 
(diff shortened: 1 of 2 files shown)


src/Chunk.cpp
```diff
@@ -474,25 +474,25 @@ void cChunk::Stay(bool a_Stay)
 void cChunk::CollectMobCensus(cMobCensus & toFill)
 {
 	toFill.CollectSpawnableChunk(*this);
-	std::list<const Vector3d *> playerPositions;
-	cPlayer * currentPlayer;
-	for (auto itr = m_LoadedByClient.begin(), end = m_LoadedByClient.end(); itr != end; ++itr)
+	std::vector<Vector3d> playerPositions;
+	playerPositions.reserve(m_LoadedByClient.size());
```
NiLSPACE: I believe we start with upper case letters.

worktycho:
I've updated the style issue.

src/Chunk.cpp
```diff
@@ -1887,31 +1885,24 @@ bool cChunk::AddClient(cClientHandle * a_Client)
 
 void cChunk::RemoveClient(cClientHandle * a_Client)
 {
-	for (cClientHandleList::iterator itrC = m_LoadedByClient.begin(); itrC != m_LoadedByClient.end(); ++itrC)
+	std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client);
```
tigerw: I thought std::remove doesn't actually remove anything?

src/Chunk.h
```diff
@@ -439,7 +439,12 @@ class cChunk :
 	void SetAlwaysTicked(bool a_AlwaysTicked);
 
 	// Makes a copy of the list
-	cClientHandleList GetAllClients(void) const {return m_LoadedByClient; }
+	cClientHandleList GetAllClients(void) const
+	{
+		cClientHandleList copy;
+		std::copy(m_LoadedByClient.begin(), m_LoadedByClient.end(), std::back_inserter(copy));
+		return copy;
```
tigerw: Could someone perhaps explain the difference to me between std::copy and returning a copy of the list? Does std::copy make copies of the clienthandles?

src/Chunk.cpp
madmaxoft: Right, if this isn't misleading naming, then nothing is. C++ standard authors should be ashamed.

tigerw: ... :P

Quick, fix it before anyone notices!

worktycho: Whoops. Why did no one catch this before merge?

tigerw: I forgot to comment :(

I sowwy

worktycho: #2484 has the fix.

src/Chunk.h
worktycho: The difference is that using the copy constructor only works if you're returning the same type of list. In order to restrict the scope of this change, I didn't expose the change to a vector publicly. Which means it needs std::copy to copy from a vector to a list.

tigerw: Ohhh.

I remember there being something about initialising containers based on ranges? So could you do `cClientHandleList copy(m_Loaded.begin(), m_Loaded.end())`?

worktycho: That would be a better way of doing it. Shall I add that to the improvements PR I'll write tomorrow, along with and better handling for duplicate clienthandles?

tigerw: Of course do, but it's up to you in any case.

madmaxoft: Isn't the iterator range initialization limited to iterators of the container type? That wouldn't work here, the iterators are of a different container type.
-- 

PR #2480, opened 2015-09-17
Added CircleCI for stylechecking.

madmaxoft:
This will allow us to remove the stylecheck from Travis builds, making them a bit faster, and having fast style checks.

There were header files missing from the cmake file list, resulting in broken style in some of them. Fixed those as part of this PR.

`OSSupport/Socket.*` files were present in the repo although unused, removed them.


src/OSSupport/Socket.cpp
```diff
@@ -1,377 +0,0 @@
-
```
NiLSPACE: I don't think this was meant to be in this PR is it?

madmaxoft: It is. The files are not used by any project, and since the stylechecking CI doesn't use cmake to produce `AllFiles.lst`, and rather scans all .cpp and .h files, this was in the way.

NiLSPACE: Oh, I misinterpreted you then. I was expecting a file to be added to the cmake file list, not the missing file to be removed ;)

worktycho:
Good idea for faster feedback.

PR #2484, opened 2015-09-18
Fixed erase

worktycho:


src/Chunk.cpp
```diff
@@ -1885,7 +1885,7 @@ bool cChunk::AddClient(cClientHandle * a_Client)
 
 void cChunk::RemoveClient(cClientHandle * a_Client)
 {
-	std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client);
+	m_LoadedByClient.erase(std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client));
```
tigerw: Are you, like, *sure* that you don't need a m_LoadedByClient.end() in the erase call? [This page](http://www.cplusplus.com/reference/vector/vector/erase/) suggests that only giving one position means it will only erase one thing.

worktycho: And if there are two copies of a pointer in that list then we've got bigger problems. But I'll create another PR that asserts if we need to delete more than one item.

tigerw: That makes sense.

> we've got bigger problems
The story of MCS/Cuberite :/

PR #2486, opened 2015-09-19
Fixed wrong assumtion regarding RemoveClient

worktycho:
Remove client can be called with a client that is not present in the chunk

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1885,7 +1885,11 @@ bool cChunk::AddClient(cClientHandle * a_Client)
 
 void cChunk::RemoveClient(cClientHandle * a_Client)
 {
-	m_LoadedByClient.erase(std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client));
+	auto itr = std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client);
+	// We should always remove at most one client.
+	ASSERT(std::distance(itr, m_LoadedByClient.end()) <= 1);
+	// Note: itr can equal m_LoadedByClient.end()
+	m_LoadedByClient.erase(itr, m_LoadedByClient.end());
 
 	if (!a_Client->IsDestroyed())
 	{
--- a/src/Chunk.h
+++ b/src/Chunk.h
@@ -441,9 +441,7 @@ public:
 	// Makes a copy of the list
 	cClientHandleList GetAllClients(void) const
 	{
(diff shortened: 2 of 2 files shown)


src/Chunk.cpp
```diff
@@ -1885,7 +1885,13 @@ bool cChunk::AddClient(cClientHandle * a_Client)
 
 void cChunk::RemoveClient(cClientHandle * a_Client)
 {
-	m_LoadedByClient.erase(std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client));
+	auto itr = std::remove(m_LoadedByClient.begin(), m_LoadedByClient.end(), a_Client);
+	// We should always remove at most one client.
+	ASSERT(std::distance(itr, m_LoadedByClient.end()) <= 1);
+	if (itr != m_LoadedByClient.end())
```
tigerw: Since you're doing a debug check, is there any reason for the check, in the form of this condition, effectively to be in release too? Why not just erase from itr to end straight up?

worktycho: There doing different things. The debug check ensures we never have more than one client to remove. The condition is to handles when m_LoadedByClient doesn't contain a_Client at all.

tigerw: I understand that you can do `m_Loaded.erase(itr, m_Loaded.end())` as-is?

worktycho: I could, but by putting in the if, I'm reminding myself that them being equal is an issue.

Personally, given it has to be zero or one, forcing it rather than a range, ensures that we have better handling for unexpected values.

tigerw: Surely though, if there is, heaven forbid, in release, more than one client handle to be removed, removing both is better than removing one and having the server crash later?

worktycho: If in release we have a duplicated client handle, we are probably going to crash anyway.

tigerw: Agreed, but there still isn't a reason for an explicit check. Assuming that the server *isn't* super-buggy (ah, what an assumption :P), it's not needed.

worktycho: The range erase is also a more complex code path, as it needs a loop, for a case that shouldn't happen.

tigerw: I guess, but can't make too much of difference, right?

worktycho:
I've given up on the argument over the range erase.

tigerw:
I feel guilty now.

PR #2487, opened 2015-09-19
Ignore Sigpipe

worktycho:


src/main.cpp
```diff
@@ -472,6 +472,10 @@ int main(int argc, char **argv)
 		#endif  // SIGABRT_COMPAT
 	#endif
 
+
+	#ifdef __unix__
+		std::signal(SIGPIPE, SIG_IGN);
```
tigerw: You could always set its handler to NonCtrl, and in there do the ignore...?

worktycho: SIGINT starts a shutdown.

tigerw: Right, but its action is set to ignore in NonCtrl. For consistency, would you want to do the same for SIGPIPE?

worktycho: No. SIGINT fires the first time calling NonCtrl. We then initiate a shutdown, and change the handler to ignore future SIGINT signals.

WIth SIGPIPE, we want to ignore all SIGPIPE signals.

tigerw: Oh, I see. I guess then you could move the "stop" command call, but nevermind.

madmaxoft:
(rebased onto current master)

PR #2497, opened 2015-09-25
Implemented brewing

Seadragon91:
~~DO NOT MERGE YET!~~

This implements brewing.
- Brewing stand and window
- [Recipes](https://github.com/Seadragon91/cuberite/blob/brewing/Server/brewing.txt), this file only contains the regular recipes. because the ingredient for a regular and a splash potion (level II -> extended) is the same.

TODO:
 - [x] Code needs to be reviewed

Discussion can be here and in the [forum](http://forum.mc-server.org/showthread.php?tid=1983)


src/BrewingRecipe.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+#include <fstream>
+
+#define BREWING_RECIPE_FILE "brewing.txt"
+
+
+
+
+
+typedef std::list<cBrewingRecipe::cRecipe> RecipeList;
```
worktycho: Why a list? It general the default container should be a vector.

src/BrewingRecipe.h
```diff
@@ -0,0 +1,49 @@
...
+
+	/** Parses the recipe contained in the line, adds it to m_pState's recipes.
+	Logs a warning to the console on input error. */
+	void AddRecipeFromLine(const AString & a_Line, unsigned int a_LineNum);
+
+	/** Parses an item string, returns true if successful. */
+	bool ParseItem(const AString & a_String, cItem & a_Item);
+
+	struct sBrewingRecipeState;
+	sBrewingRecipeState * m_pState;
```
worktycho: This should be a unique_ptr, not a raw pointer.

src/BrewingRecipe.h
```diff
@@ -0,0 +1,49 @@
...
+	cBrewingRecipe(void);
+	~cBrewingRecipe();
+
+	void ReloadRecipes(void);
+
+	struct cRecipe
+	{
+		cItem * Input;
+		cItem * Output;
+		cItem * Ingredient;
```
worktycho: All these fields should be unique_ptr's

src/BrewingRecipe.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+
+
+
+void cBrewingRecipe::ReloadRecipes(void)
+{
+	ClearRecipes();
+	LOGD("Loading brewing recipes...");
+
+	std::ifstream f(BREWING_RECIPE_FILE, std::ios::in);
```
worktycho: Generally this project uses our own cFile class. I don't know why, but @madmaxoft doesn't like the streams apis.

src/BrewingRecipe.cpp
```diff
@@ -0,0 +1,324 @@
...
+		{
+			continue;
+		}
+
+		// Remove comments from the line:
+		size_t FirstCommentSymbol = ParsingLine.find('#');
+		if ((FirstCommentSymbol != AString::npos) && (FirstCommentSymbol != 0))
+		{
+			ParsingLine.erase(ParsingLine.begin() + static_cast<const long>(FirstCommentSymbol), ParsingLine.end());
+		}
```
worktycho: If you move this if above the empty line check, and remove the comparison against a first char comment, then you can remove the following switch.

src/BrewingRecipe.cpp
```diff
@@ -0,0 +1,324 @@
...
+		if (ParsingLine.empty())
+		{
+			continue;
+		}
+
+		// Remove comments from the line:
+		size_t FirstCommentSymbol = ParsingLine.find('#');
+		if ((FirstCommentSymbol != AString::npos) && (FirstCommentSymbol != 0))
+		{
+			ParsingLine.erase(ParsingLine.begin() + static_cast<const long>(FirstCommentSymbol), ParsingLine.end());
```
worktycho: I think what you want here is `std::advance(ParsingLine.begin(), FirstCommentSymbol)` rather than the cast and addition.

src/BrewingRecipe.cpp
Seadragon91: I am not really sure how I should change it.

src/BrewingRecipe.cpp
Seadragon91: This shows an error: `no matching function for call to 'advance'`. I think he don't like the iterator.

src/BrewingRecipe.cpp
Seadragon91: I think you mean the cIniFile. The brewing.txt is not really a ini file, it's like the furnace.txt.

src/BrewingRecipe.cpp
worktycho:
```
// Remove comments from the line:
size_t FirstCommentSymbol = ParsingLine.find('#');
if ((FirstCommentSymbol != AString::npos))
{
	ParsingLine.erase(ParsingLine.begin() + static_cast<const long>(FirstCommentSymbol), ParsingLine.end());
}

if (ParsingLine.empty())
{
	continue;
}
```

src/BrewingRecipe.cpp
worktycho: My mistake, std::advance increments an lvalue. But its probably still worth using even with the extra two statements to first declare and the increment the iterator, given that the correct form of that cast is `static_cast<std::iterator_traits<AString::iterator>::difference_type(FirstCommentSymbol)`

src/BrewingRecipe.cpp
madmaxoft: Use `auto`, no need to `static_cast`.

src/BrewingRecipe.cpp
worktycho: No I meant cFile, the headers at src/OSSupport/File.h.

madmaxoft: Nah, using streams for file IO is okay. I had some reservations for using them to convert between strings and numbers.

worktycho: So why does cFile exist? Is it just legacy?

src/BrewingRecipe.cpp
```diff
@@ -0,0 +1,299 @@
...
+		return nullptr;
+	}
+
+	// Check for splash potion
+	if (a_Input.m_ItemDamage & 0x4000)
+	{
+		// Search for the drinkable potion, the ingredients are the same
+		short SplashItemDamage = a_Input.m_ItemDamage - 8192;
+
+		for (RecipeList::const_iterator itr = m_pState->Recipes.begin(); itr != m_pState->Recipes.end(); ++itr)
```
worktycho: Range for loop?

src/BrewingRecipe.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+
+
+
+
+const cBrewingRecipe::cRecipe * cBrewingRecipe::GetRecipeFrom(const cItem & a_Input, const cItem & a_Ingredient) const
+{
+	const cRecipe * BestRecipe = nullptr;
+
+	for (RecipeList::const_iterator itr = m_pState->Recipes.begin(); itr != m_pState->Recipes.end(); ++itr)
```
worktycho: range for loop

src/BlockEntities/BrewingstandEntity.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+
+
+
+
+void cBrewingstandEntity::UpdateProgressBars(bool a_ForceUpdate)
+{
+	/** Sending an update every 3th tick, using a higher value lets look the progressbar ugly */
+	if (!a_ForceUpdate && (m_World->GetWorldAge() % 3 != 0))
```
worktycho: This updates on wallclock time. Not sure how to fix it though.

Seadragon91:
Okay I am not really sure, what the error is. I think it has to do with `cBrewingRecipe::cRecipe`.

src/BrewingRecipe.cpp
NiLSPACE: Even if cFile wouldn't be used in the cpp code it's still convenient to have it for the Lua API.

worktycho:
I don't think it does. MSVC is complaining about a non-existent control flow  issue in a method you haven't touched. This sort of brokenness is good example of why we need CI to compile against the exact result of the merge.

Seadragon91:
I get that in visual stuido
`Error	37	error C2280: 'std::unique_ptr<cItem,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)' : attempting to reference a deleted function	C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xmemory0	577	1	Cuberite`

I really think I forgot something...

worktycho:
You're right, I'm used to make stopping on errors so I misread it. MSVC is still broken though as its trying to generate a copy constructor for a type with move only members. Try explicitly deleting the cRecipe copy constructor.

tigerw:
According to [here](http://stackoverflow.com/questions/30120573/visual-studio-2013-does-not-delete-the-copy-constructor-when-a-user-defined-move), the bug is fixed in VS2015.

Seadragon91:
@tigerw I installed visual studio 2015 and this has been fixed there. Appveyor would be needed to upgraded, before the workaround can be removed.

Seadragon91:
@worktycho  After change to `std::unique_ptr`,  the recipe list contains only empty recipes. I debugged that and tried to find the cause, but had no success. Do you have an idea, whats wrong?

Seadragon91:
Found it. Had to change `typedef std::vector<cBrewingRecipe::cRecipe> RecipeList;` to `typedef std::vector<std::unique_ptr<cBrewingRecipe::cRecipe>> RecipeList;` and then fixing all the errors.

Seadragon91:
In the brewing stand I store the current recipes that has been found (can be 3) in a pointer array named `m_CurrentBrewingRecipes`. Also simplified and fixed the return of a Recipe in GetRecipeFrom.

Seadragon91:
Hooks should be added, too. The question is now what hooks. For started brewing process, brewing completed and a hook for a not found brewing recipe?

Seadragon91:
I think a hook for a completed brewing process should be enough.

Seadragon91:
Added hooks:
- Brewing process is completing
- Brewing process is completed

Seadragon91:
I could add a array that has the result item for every bottle slot and it could be accessed from lua. My question is do we want that?

Seadragon91:
Are the hooks good?
Only thing is now, if we want access to the current brewing result from lua side.

Seadragon91:
I fixed a few potions and added a array that contains the result items, for the brewing recipes, that can be accessed from to lua.

I think I have everything for the brewing. Only a good review would be needed.

Seadragon91:
How should be proceed with this pull request?

SphinxC0re:
rebase an squash onto current master, then we will see

Seadragon91:
If no one is against a merge, then I will rebase and merge this in the next days.

SphinxC0re:
You should rebase again

SphinxC0re:
also @cuberite/server-contributors nayone against the merge?

SphinxC0re:
@cuberite/maintainers

src/BrewingRecipes.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+
+
+
+void cBrewingRecipes::ReloadRecipes(void)
+{
+	ClearRecipes();
+	LOGD("Loading brewing recipes...");
+
+	std::ifstream f(BREWING_RECIPE_FILE, std::ios::in);
```
NiLSPACE: Don't we use cFile for reading/writing files?

NiLSPACE:
If not using cFile is not a problem, then I have no objection.

Seadragon91:
We alread discussed about that [here](https://github.com/cuberite/cuberite/pull/2497#discussion_r40496296).

NiLSPACE:
Whoops, then I have no objection :)

src/Root.h
```diff
@@ -208,6 +210,7 @@ class cRoot
 
 	cCraftingRecipes * m_CraftingRecipes;
 	cFurnaceRecipe *   m_FurnaceRecipe;
+	cBrewingRecipes *  m_BrewingRecipes;
```
tigerw: Break the cycle and make it a unique_ptr :)

src/ChunkMap.h
```diff
@@ -246,6 +248,10 @@ class cChunkMap
 	Returns true if all block entities processed, false if the callback aborted by returning true. */
 	bool ForEachBlockEntityInChunk(int a_ChunkX, int a_ChunkZ, cBlockEntityCallback & a_Callback);  // Lua-accessible
 
+	/** Calls the callback for brewingstand in the specified chunk.
+	Returns true if all brewingstands processed, false if the callback aborted by returning true. */
+	bool ForEachBrewingstandInChunk(int a_ChunkX, int a_ChunkZ, cBrewingstandCallback & a_Callback);  // Lua-accessible
```
tigerw: Do you need this? @worktycho, did you say something about just using the generic ForEachBlockEntity function?

src/OSSupport/File.cpp
```diff
@@ -18,7 +18,7 @@
 cFile::cFile(void) :
 	m_File(nullptr)
 {
-	// Nothing needed yet
+	// Nothing needed ye
```
madmaxoft: Get rid of this change

src/BlockEntities/BrewingstandEntity.h
```diff
@@ -0,0 +1,164 @@
...
+				continue;
+			}
+			Recipe = BR->GetRecipeFrom(GetSlot(i), GetSlot(bsIngredient));
+			if (Recipe != nullptr)
+			{
+				m_CurrentBrewingRecipes[i] = Recipe;
+				m_Results[i] = Recipe->Output->CopyOne();
+			}
+		}
+	}
```
madmaxoft: The implementation should go into the cpp file.

src/BlockEntities/BrewingstandEntity.h
```diff
@@ -0,0 +1,164 @@
...
+
+	/** Starts the brewing proccess. Will be called if the brewing stand gets loaded from the world. */
+	void ContinueBrewing(void)
+	{
+		// Continue brewing if number is greater than 0
+		if (m_TimeBrewed > 0)
+		{
+			m_IsBrewing = true;
+		}
+	}
```
madmaxoft: The implementation should go into the cpp file.

src/UI/SlotArea.cpp
```diff
@@ -2338,7 +2575,3 @@ cItem * cSlotAreaTemporary::GetPlayerSlots(cPlayer & a_Player)
 	}
 	return itr->second.data();
 }
-
-
-
-
```
madmaxoft: Keep the final 5 empty lines

src/Root.h
Seadragon91: All 3? :)

src/UI/SlotArea.cpp
SphinxC0re: Why?

src/ChunkMap.h
madmaxoft: Keep it for now, we can handle that in another PR.

Seadragon91: I think all the callbacks should be reworked, but that should be done in a  own pull request.

src/OSSupport/File.cpp
Seadragon91: Ops :)

src/UI/SlotArea.cpp
madmaxoft: It makes MSVS jump around the file when editing near the end and starting the debug session

src/ChunkMap.h
tigerw: Alright, but we are committing to a documented new API, and with all of your insistence of backwards compatibility...

...that DeprecatedBindings file doesn't want to get larger :P

madmaxoft: This will only be an API function, but not an actual call in `cWorld` / `cChunkMap`, the API functions will use a common template-based framework over `ForEachEntityInChunk()`

src/BrewingRecipes.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+void cBrewingRecipes::AddRecipeFromLine(const AString & a_Line, unsigned int a_LineNum)
+{
+	AString Line(a_Line);
+	Line.erase(std::remove_if(Line.begin(), Line.end(), isspace), Line.end());
+
+	short InputDamage;
+	short OutputDamage;
+
+	std::unique_ptr<cItem> InputItem(new cItem());
```
tigerw: Make these a cpp14::make_unique?

src/BrewingRecipes.cpp
```diff
@@ -0,0 +1,282 @@
...
+		}
+	}
+
+	// Check for gunpowder
+	if (a_Ingredient.m_ItemType == E_ITEM_GUNPOWDER)
+	{
+		if (a_Input.m_ItemDamage & 0x2000)
+		{
+			// Create new recipe and add it to list
+			std::unique_ptr<cItem> InputItem(new cItem());
```
tigerw: make_unique?

src/ChunkMap.h
tigerw: Oh. Why not have authors check for themselves the block entity they want in the generic function? Its a lot of effort to remember/implement new templates for each new block entity.

madmaxoft: Because the check is trivial in C++, but rather costly when done in Lua (each C++-to-Lua call having a penalty hit for each entity, then each check calls Lua-to-C++ virtual function just to check type)

src/Root.h
Seadragon91: Okay there are more lines, I think this should be done in own pull request. Or should I change only my line? :)

madmaxoft: Change only yours, we can adapt the rest later(tm)

Seadragon91: Okay this result in changing around 10 lines and still not works. Let it for a own pull request...

Seadragon91:
Okay with this unique_ptr change? :)

Seadragon91:
Are the changes okay? Can I squash and rebase?

src/Root.h
```diff
@@ -208,6 +210,7 @@ class cRoot
 
 	cCraftingRecipes * m_CraftingRecipes;
 	cFurnaceRecipe *   m_FurnaceRecipe;
+	std::unique_ptr<cBrewingRecipes>  m_BrewingRecipes;
```
madmaxoft: Remove the extra space - you're not aligning anymore, so a single space is enough.

Seadragon91: Removed it.

SphinxC0re:
@cuberite/core-team again, is anyone against a merge?

tigerw:
🎵No not really 'cos oh! I think that you've found yourself a merged PR, etc. etc. 🎶

SphinxC0re:
What?

tigerw:
https://en.wikipedia.org/wiki/Cheerleader_(song)

PR #2498, opened 2015-09-25
Enabled Unicode support for Windows

tigerw:
* Enabled Unicode support on Windows

Testing completed:

Test | Status
-------|----------
Crash dump generation| Success
~~GetOSErrorString output~~ | ~~Success~~


madmaxoft:
How about the `cFile::GetFolderContents()` function? It still uses the old MBCS interface functions.

src/LeakFinder.cpp
```diff
@@ -875,7 +875,7 @@ static int MyAllocHook(int nAllocType, void *pvData,
 				printf("********************************************\n");
 				printf("Please wait\n");
 				
-				LeakFinderXmlOutput Output("memdump.xml");
+				LeakFinderXmlOutput Output(L"memdump.xml");
```
madmaxoft: Instead of hard-coding string width, you should use the `TEXT()` macro and the `LPTSTR` or `LPCTSTR` typedefs for pointers.

src/OSSupport/Errors.cpp
```diff
@@ -5,13 +5,13 @@
...
 {
-	char buffer[ 1024 ];
 	AString Out;
 
 	#ifdef _WIN32
 
+	wchar_t buffer[1024];
 	FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, nullptr, a_ErrNo, 0, buffer, ARRAYCOUNT(buffer), nullptr);
-	Printf(Out, "%d: %s", a_ErrNo, buffer);
+	Printf(Out, "%d: %S", a_ErrNo, buffer);
```
madmaxoft: What happens if the message comes in Chinese from the OS and the `Printf` fails because it cannot convert it into MBCS?

worktycho:
GetFolderContents uses the explicit Ansi versions, so is unaffected by the change.

src/main.cpp
```diff
@@ -131,7 +131,7 @@ typedef BOOL  (WINAPI *pMiniDumpWriteDump)(
 
 pMiniDumpWriteDump g_WriteMiniDump;  // The function in dbghlp DLL that creates dump files
 
-char g_DumpFileName[MAX_PATH];  // Filename of the dump file; hes to be created before the dump handler kicks in
+wchar_t g_DumpFileName[MAX_PATH];  // Filename of the dump file; hes to be created before the dump handler kicks in
```
madmaxoft: Use `TCHAR` types

src/main.cpp
```diff
@@ -310,14 +310,14 @@ void WINAPI serviceCtrlHandler(DWORD CtrlCode)
 
 void WINAPI serviceMain(DWORD argc, TCHAR *argv[])
 {
-	char applicationFilename[MAX_PATH];
-	char applicationDirectory[MAX_PATH];
+	wchar_t applicationFilename[MAX_PATH];
+	wchar_t applicationDirectory[MAX_PATH];
```
madmaxoft: `TCHAR`

src/main.cpp
```diff
@@ -310,14 +310,14 @@ void WINAPI serviceCtrlHandler(DWORD CtrlCode)
...
 
 	GetModuleFileName(nullptr, applicationFilename, sizeof(applicationFilename));  // This binary's file path.
 
 	// Strip off the filename, keep only the path:
-	strncpy_s(applicationDirectory, sizeof(applicationDirectory), applicationFilename, (strrchr(applicationFilename, '\\') - applicationFilename));
-	applicationDirectory[strlen(applicationDirectory)] = '\0';  // Make sure new path is null terminated
+	wcsncpy_s(applicationDirectory, sizeof(applicationDirectory), applicationFilename, (wcsrchr(applicationFilename, '\\') - applicationFilename));
```
madmaxoft: `tcsncpy` or whatever it's called nowadays

madmaxoft:
@worktycho yes, but it's inconsistent now. Even worse with Lua, it uses the MBCS versions, too, and so suddenly cannot open some files that Cuberite can.

This inconsistency is one of the main reasons I didn't want to go through with this at all.

worktycho:
What this really needs is test on a machine set up in Chinese before you'll accept this?

madmaxoft:
Seriously, I don't understand why everyone's pushing this change anyway. No matter what you do, Lua will still stay MBCS, so you still won't have the desired effect anyway. I'm just trying to list the reasons why I think this is foolish.

worktycho:
I'm just putting that up as something I think tigers unlikely to do. Personally I'd like unicode, but I think it sits beyond sorting out the god objects in difficulty. So I'm just going to throw up test case after test case until tiger recognizes the difficulty.

src/OSSupport/Errors.cpp
tigerw: Well, I've installed the Chinese (Simplified) language pack on me computer, and have founds that in short, getting Unicode to print on a console window is for me impossible, since writing to the console window apparently isn't truly done in Unicode but is instead dependant on the system locale.

The GUI part of things (even MessageBox called from a console app) fully supports Unicode, but alas, we're not a graphical application.

Of course, tests show that the ANSI version doesn't work with Chinese, displaying all '?'s in the IDE, even when the Wide version produced a sentence correctly.

Since there doesn't appear to be any way to fix this, I'll revert the Wide changes to this file, since that's more correct in context of the rest of the server.

tigerw: I can't actually find where GetOSErrorString() is used though.

tigerw:
In conclusion, the console does not support Unicode fully, so it looks like this change is going to be a minor speedup in the dump-file initialisation portion, and that only.

Anyone willing to convert us to WinAPI-based WinForms? :P

src/main.cpp
tigerw: A question was posed [here](http://stackoverflow.com/questions/234365/is-tchar-still-relevant). The outcome is generally that unless you want to be backwards compatible, it's best to stick with Wide and be explicit about it.

src/OSSupport/Errors.cpp
worktycho: It is possible to fix the console with more custom win32 code, as utf8 is a valid codepage (65001). We could set the codepage to that on startup, and then restore on shutdown. But yes, this is a big change.

worktycho:
As I mentioned in my comment on the line, windows does support unicode on the console, but it is a different encoding form  from the api unicode, so we would probably need to translate. Making a win32 subsystem app is unnecessary.

tigerw:
It printed, as an example, Chinese, with the Wide-style code, but only when the "locale for non-Unicode programs" was set to 中文(PRC). If it was set to English(UK)/英语(英国), no matter what I did, whether it was getting a wide string from FormatMessageW, or getting a normal one from FormatMessageA and calling MultiByteToWideChar on it, none of the method WriteConsole, std::wcout, or wprintf would work. I could only consistently get Chinese from a combination of Wide and MessageBoxW.

`_setmode(_fileno(stdout), _O_U16TEXT);` was used before all printing.

worktycho:
But did you try `SetConsoleCP((UINT)65001);`? Setting stdout to utf16 need only set the input to the console out to utf 16, it need not set the actual console output output to a unicode code page.

tigerw:
confounded unicode

Am I doing it right?

worktycho:
Its got the right number of box's, and there all box's which suggests a font issue. What happens if you go into the console preferences and set the font to Lucida Console?

tigerw:
The font is presently set to Lucida Console. I don't expect it to support the entirety of Unicode though, and therein lies the problem, from what it seems.

worktycho:
H'm. Can you find out what font the dialog box is using and set it to that? Also try copying and pasting the box's to github to see what they result in.

tigerw:
SimSun or NSimSun?

According to research, yes, the console knows what the text is, it just can't display it.

worktycho:
That's outside where I can help then. Xoft might be able to help you through. If not there's always stackoverflow.

tigerw:
In the meantime, merge?

madmaxoft:
If I merge, will you consider this closed and won't poke in this further? :P

tigerw:
Looks like it for the near future.

madmaxoft:
And I've got the first victim of this mega-bullet: LibEvent's error reporting reports single character error strings ever since this change.

PR #2500, opened 2015-09-26
Extended and fixed the cFile API.

madmaxoft:
Added a few common operations (CreateFolderRecursive, DeleteFolderContents) and fixed the API function signatures (tolua would return the strings in parameters as an extra return value)

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -482,8 +482,258 @@ cPluginLua * cManualBindings::GetLuaPlugin(lua_State * L)
 
 
 
+static int tolua_cFile_ChangeFileExt(lua_State * tolua_S)
+{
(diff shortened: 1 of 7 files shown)


src/Bindings/ManualBindings.cpp
```diff
@@ -508,8 +758,119 @@ static int tolua_cFile_GetFolderContents(lua_State * tolua_S)
...
+		!L.CheckParamEnd(3)
+	)
+	{
+		return 0;
+	}
+
+	// Execute:
+	AString Path;
+	ASSERT(L.GetStackValues(2, Path));
+	L.Push(cFile::GetSize(Path));
```
worktycho: There isn't a push overload for long, windows it converts to int so this works, but on linux its a long long, so this is ambiguous.

PR #2501, opened 2015-09-27
Api dump check

madmaxoft:
This allows the CI builds to check for newly exported API symbols without documentation.

### DIFF
--- a/CIbuild.sh
+++ b/CIbuild.sh
@@ -21,8 +21,19 @@ cd ..
 echo "Building..."
 make -j 2;
 make -j 2 test ARGS="-V";
+
+echo "Testing..."
 cd Server/;
 if [ "$TRAVIS_CUBERITE_BUILD_TYPE" != "COVERAGE" ]; then
-	echo restart | $CUBERITE_PATH;
-	echo stop | $CUBERITE_PATH;
+	$CUBERITE_PATH << EOF
+load APIDump
+apicheck
+restart
+stop
+EOF
+	if [ -f ./NewlyUndocumented.lua ]; then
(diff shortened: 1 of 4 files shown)


madmaxoft:
This looks merge-able now.

This will refuse PRs that add new API symbols but fail to document it in the APIDump plugin, hence forcing all new API to be documented.

Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -1616,9 +1615,22 @@ local function DumpApi()
...
+local function DumpApi()
+	LOG("Dumping the API...")
+
+	-- Match the currently exported API with the available documentation:
+	local API, Globals = PrepareApi()
+	
 	-- Check that the API lists the inheritance properly, report any problems to a file:
 	CheckAPIDescendants(API)
-
+
```
SphinxC0re: Could you remove this

worktycho:
It would also be good to check the other way, for documentation for functions that aren't exposed.

CIbuild.sh
```diff
@@ -21,8 +21,19 @@ cd ..
...
+
+echo "Testing..."
 cd Server/;
 if [ "$TRAVIS_CUBERITE_BUILD_TYPE" != "COVERAGE" ]; then
-	echo restart | $CUBERITE_PATH;
-	echo stop | $CUBERITE_PATH;
+	$CUBERITE_PATH << EOF
+load APIDump
+apicheck
+restart
```
madmaxoft: For some reason the `restart` command seems to be ignored by the server. I don't care much at this moment, but we might want to look at it later.

Issue #2460, opened 2015-09-02
Squids go on dry land and die, burning


jammet:
I didn't see this until just now, so it must have been a very recent change.

Squids used to stand completely still before, because their AI wasn't working properly. Safwat, I think, wanted to take care of these at a later point. Now they've started to move again, and they all seem to think they're land creatures, so they go on dry land and burn...

SafwatHalaby:
I thought I completely froze squids as a temporary solution.

jammet:
I know you did, but that changed a while ago. Now they move around again. ^^' Some other change must have interfered.

![2015-09-25_13 32 38](https://cloud.githubusercontent.com/assets/11940735/10099442/f9b5b850-6389-11e5-8feb-8740248d1173.png)

SafwatHalaby:
Found it. Guilty commit: https://github.com/cuberite/cuberite/pull/2446. @tigerw 's commit set the squids loose again because it changed the way `m_IsFollowingPath` behaves.

SafwatHalaby:
Fixed in #2502.

PR #2502, opened 2015-09-27
Disabled squids and Guardians pathfinding, #2460

SafwatHalaby:
Fixed Issue: #2460
Introduced at: #2446


SafwatHalaby:
I typically wait for 2 guys, but the change is rather trivial. Merging.

PR #2508, opened 2015-09-28
Added LuaState tracker and memory stats logging.

madmaxoft:
Added a new built-in command `luastats` that lists the memory used by each `cLuaState` instance currently active:
```
[6fcee5d7|21:29:14] Executing console command: "luastats"
[6fcee5d7|21:29:14] State "<webadmin_template>" is using 653 KiB of memory
[6fcee5d7|21:29:14] State "plugin APIDump" is using 1238 KiB of memory
[6fcee5d7|21:29:14] State "plugin ChunkWorx" is using 592 KiB of memory
[6fcee5d7|21:29:14] State "plugin Core" is using 1272 KiB of memory
[6fcee5d7|21:29:14] State "plugin GalExport" is using 889 KiB of memory
[6fcee5d7|21:29:14] State "plugin Gallery" is using 850 KiB of memory
[6fcee5d7|21:29:14] State "plugin LastSeen" is using 597 KiB of memory
[6fcee5d7|21:29:14] State "plugin NetworkTest" is using 625 KiB of memory
[6fcee5d7|21:29:14] State "plugin PluginMemory" is using 1452 KiB of memory
[6fcee5d7|21:29:14] State "plugin ProtectionAreas" is using 680 KiB of memory
[6fcee5d7|21:29:14] State "plugin WorldEdit" is using 815 KiB of memory
[6fcee5d7|21:29:14] Total memory used by Lua: 9663 KiB
```

This will allow us to see if increased memory usage reported in Linux is due to Lua or our memory leaks. Also, is cool :)

### DIFF
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -43,6 +43,76 @@ const cLuaState::cRet cLuaState::Return = {};
 
 
 ////////////////////////////////////////////////////////////////////////////////
+// cLuaStateTracker:
+
+void cLuaStateTracker::Add(cLuaState & a_LuaState)
+{
+	auto & Instance = Get();
+	cCSLock Lock(Instance.m_CSLuaStates);
+	Instance.m_LuaStates.push_back(&a_LuaState);
+}
+
+
+
+
+void cLuaStateTracker::Del(cLuaState & a_LuaState)
+{
+	auto & Instance = Get();
+	cCSLock Lock(Instance.m_CSLuaStates);
+	Instance.m_LuaStates.erase(
+		std::remove_if(
+			Instance.m_LuaStates.begin(), Instance.m_LuaStates.end(),
+			[&a_LuaState](cLuaStatePtr a_StoredLuaState)
+			{
+				return (&a_LuaState == a_StoredLuaState);
+			}
+		),
+		Instance.m_LuaStates.end()
+	);
+}
+
+
+
+
+
+AString cLuaStateTracker::GetStats(void)
+{
+	auto & Instance = Get();
+	cCSLock Lock(Instance.m_CSLuaStates);
+	AString res;
+	int Total = 0;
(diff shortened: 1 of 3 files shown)


Seadragon91:
Would it be possible to use this also from the plugin side? Something like looping over all plugins getting the memory and dumping it to a file over a scheduled task?

NiLSPACE:
You could use cPluginManager:CallPlugin(pluginname, "collectgarbage", "count")

madmaxoft:
That would work for plugins, but not for the other created cLuaState instances - webadmin template or prefab generator piecepools.

NiLSPACE:
True, but that's what Seadragon91 asked for ^^

I do like the idea where plugins can interact with the other Lua states a little more. Now the plugins only know it exists simply because the developer knows it's in there somewhere.  Perhaps we could even add an [`ExecuteString`](https://github.com/mc-server/WorldEdit/blob/master/Api/Manage.lua#L281) function like in WorldEdit. That would allow plugins to edit, remove or add structures to existing prefabs.

Something like this:
```lua
cPrefab:ExecuteString("SandFlatRoofVillage", [[
   function()
      for Idx, PrefabData in ipairs(Cubeset.Pieces) do
         if (PrefabData.OriginData.Name == "Desert 25") then
            table.remove(CubeSet.Pieces, Idx)
            return true
         end
      end
      return false
   end
]])
```

It could even work with string.dump so you don't have to write the function as a string.

madmaxoft:
@Seadragon91 My PluginMemory plugin already does that: https://github.com/madmaxoft/PluginMemory

@NiLSPACE I don't think that's a good idea, breaking sandboxes like that. And anyway it won't work with the prefabs, because they're loaded only once and then the LuaState is destroyed.

NiLSPACE:
Oh, so you used Lua only as a parser. For some reason I expected the generator to use the state the whole time. No wonder I didn't see a drop in performance, it all makes sense now :P

PR #2511, opened 2015-09-29
The bottom layer cannot hold a valid spawn point

### LINKED ISSUE #2510: Debug mode - Assertion error
Hey,

I if generate a air world in debug mode, the server will crash at startup on this [line](https://github.com/cuberite/cuberite/blob/master/src/Chunk.cpp#L2612).

In [World.cpp](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L649) we decrease y by one and then we have `y = -1` and that's why the server crashes at line above.

Removing the assertion line, fixes this crash at startup.

Not sure if there will be other assertions that could occur in a air world with a server running in debug mode.

regards,
Seadragon91

worktycho:
Fixes #2510

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -624,6 +624,12 @@ void cWorld::GenerateRandomSpawn(void)
 
 bool cWorld::CheckPlayerSpawnPoint(int a_PosX, int a_PosY, int a_PosZ)
 {
+	// The bottom layer cannot hold a valid spawn point
+	if (a_PosY <= 0)
+	{
+		return false;
+	}
+
 	// Check that spawnblock and surrounding blocks are neither solid nor water / lava
 	static const struct
 	{


src/World.cpp
```diff
@@ -624,6 +624,12 @@ void cWorld::GenerateRandomSpawn(void)
 
 bool cWorld::CheckPlayerSpawnPoint(int a_PosX, int a_PosY, int a_PosZ)
 {
+	// The bottom layer cannot hold a valid spawn point
+	if (a_PosY <= 0)
+	{
+		return false;
+	}
+
```
Seadragon91: Works fine, remove the tab and then you can merge it. Thanks you.

NiLSPACE: I believe we keep the tabs on those lines. At least it's still [here](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L2556), [here](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L2733), and [here](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L3089)

madmaxoft: We haven't specified a policy for that yet, I personally try to avoid trailing whitespace on empty lines, but don't want to edit it out of the entire code-base, since that would totally destroy `git blame`

worktycho:
Fixed whitespace. Merge?

madmaxoft:
Merge when it builds.

SphinxC0re:
The branch is ood with master. Would you please rebase onto it again

PR #2513, opened 2015-09-29
Fixed a missing cast

SphinxC0re:
My IDE told me, that this section would cause the method to return an int instead of a NIBBLETYPE


Seadragon91:
I have 600 warnings in vs 2015, over 500 of them are false warnings and the last are about conversion, buffer overrun, "consider moving some data to heap", and deprecated posix names.

SphinxC0re:
So...?

PR #2516, opened 2015-09-30
Fixed a few warnings

SphinxC0re:
I'm not sure if this is right. Shouldn't the enum class for blocks return a BLOCKTYPE?


madmaxoft:
There has to be a better way of doing it, rather than dropping casts everywhere

SphinxC0re:
Like I said, just let the enum array return a BLOCKTYPE

madmaxoft:
How about re-typing `E_BLOCK_XYZ` constants from an `enum` into a `static const UInt8` or something like that?

SphinxC0re:
This might be a better idea. I recognized that the enum already returns a BLOCKTYPE

SphinxC0re:
But why not re-typing them all into a ```static const BLOCKTYPE``` ?

madmaxoft:
(That's what I meant, edited my comment)

PR #2517, opened 2015-09-30
ReTyped block ids

SphinxC0re:
As discussed in #2516

### DIFF
--- a/src/BlockID.h
+++ b/src/BlockID.h
@@ -3,12 +3,13 @@
 // The following hackery is to allow typed C++ enum for C++ code, yet have ToLua process the values.
 // ToLua doesn't understand typed enums, so we use preprocessor to hide it from ToLua.
 
-enum ENUM_BLOCK_ID : BLOCKTYPE
+
+static const BLOCKTYPE
 #if 0
-enum ENUM_BLOCK_ID  // tolua_export
-#endif
 // tolua_begin
+enum BLOCKTYPE
 {
+#endif
 	E_BLOCK_AIR = 0,
 	E_BLOCK_STONE = 1,
 	E_BLOCK_GRASS = 2,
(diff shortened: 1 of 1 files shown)


SphinxC0re:
@madmaxoft HELP! :sob:

worktycho:
tolua doesn't support assignments in declarations. We either need to patch tolua, or do something clever with macros, or use two sets of declarations.

tigerw:
OMG.

This is *waaaay* too verbose, xoft, c'mon.

NiLSPACE:
Would :
```cpp
static const BLOCKTYPE 
   E_BLOCK_AIR = 0,
   E_BLOCK_STONE = 1,
   // ....

```

or something simular work?

tigerw:
What was the warning that this fixed? The original issue casts ::GetBlock to a BLOCKTYPE, which I thought it already returned?

worktycho:
It wont work. If you want to use a single declaration in c++ we need to patch tolua to ignore the equals.

SphinxC0re:
We have to patch tolua anyway

worktycho:
What might work is:
```
static const BloCKTYPE
#ifdef 0
enum BLOCKTYPE // tolua_export
{
#endif
   E_BLOCK_AIR = 0,
   E_BLOCK_STONE = 1,
   // ....
#ifdef 0
}
#endif
;

**Comment by SphinxC0re** [comment]
Doesn't work. Any other suggestions?

**Comment by madmaxoft** [comment]
You didn't use @worktycho's suggestion, it should have worked.

**Comment by SphinxC0re** [comment]
Oh, I did try it locally and it failed

**Comment by madmaxoft** [comment]
Try the updated one, with `tolua_begin` instead

**Comment by SphinxC0re** [comment]
I get an error:
```
Error: macro names must be identifiers
 #ifdef 0
        ^
```

madmaxoft:
Okay, let me try locally. I'll rebase this onto current master, too.

madmaxoft:
There, this should work.

SphinxC0re:
looks good

SphinxC0re:
circle seems to be stuck

madmaxoft:
Yes, they seem to be experiencing some problems. I've verified that the basic style checks out. Merging manually.

SphinxC0re:
And there go another three warnings :smile:

tigerw:
Does nobody care what I say? Why was there a change from enumerations, what should be used, to something that is essentially a ton of static constants, with *less* type safety, all in the name of "fixing compiler warnings"?

worktycho:
Because this change has to happen if we want to support mods that add blocktypes. Enum have undefined behavior for values that aren't in the enum, so we can't use other values. And we have better type safety for how we want to use blocktypes in the future.

tigerw:
Assuming that the extensibility was an intended side-effect of this PR, I argue that you could do it perfectly well having a separate dynamic structure for custom handlers and their ID mappings.

After all, what could this change have possibly achieved? Comparing custom-defined BLOCKTYPEs within the core server? Dynamically creating a new static constant? The server should have its own, unchangeable and inextensible enumeration of basic blocks, and this didn't exactly further that intention.

worktycho:
And what about the block array which also uses blocktype? That needs to store custom blocks.

tigerw:
Damn, you have a point. It occurred to me that storing blocks necessitates mixing the classed enum values and plain unsigned chars. So it looks like extensible handlers demands no enumeration usage. That's really unsatisfying :(

PR #2523, opened 2015-10-04
Decoupled cMonster and path recalculation logic, created cPathFinder

SafwatHalaby:
cMonster had too much path-finding code. I've moved the path recalculation logic into a new class called cPathFinder.

Advantages:
 - Cleaner cMonster, seperation of concerns.
 - The path recalculation algorithm can be swapped without having to edit the AI code simply by reimplementing by  `cPathFinder`.

**Some background**

**cPath**: A device which takes two points as an input, and returns a path.

The most accurate pathfinding would be calling cPath every single step. But that's very resource-intensive. A saner approach would be calling cPath less often, but that's a delicate task: If we call cPath too little, the mob would move along stalled paths and it ends going to where the player had been long ago. Until now,  cMonster had this path recalculation logic, but that made it too tightly coupled with pathfinding and too messy. This commit moves that recalculation logic to a new class called cPathFinder.

**cPathFinder**: Wraps cPath, takes two points as an input each tick, returns the next way point the mob should move to. It internally decides whether to use an old cPath or to recalculate. That decision is based on both the time since the last recalculation and the distance to the target.

In most cases, no one should use cPath directly. The only exception I can think of is when one needs to pre-calculate static paths, but perhaps that should also be wrapped in some class like `cStaticPathFinder`.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -239,6 +239,7 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		{
 			case ePathFinderStatus::PATH_FOUND:
 			case ePathFinderStatus::NEARBY_FOUND:
+			{
 				/* If I burn in daylight, and I won't burn where I'm standing, and I'll burn in my next position, and at least one of those is true:
 				1. I am idle
 				2. I was not hurt by a player recently.
@@ -258,7 +259,7 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 					a_IsFollowingPath = true;  // Used for proper body / head orientation only.
 					MoveToWayPoint(*Chunk);
 				}
-				break;
+			}
 			default:
 			{
 
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -13,7 +13,7 @@
 #include "../Chunk.h"
 #include "../FastRandom.h"
 
-#include "Path.h"
+#include "PathFinder.h"
 
 
 
@@ -74,11 +74,8 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_EMState(IDLE)
 	, m_EMPersonality(AGGRESSIVE)
 	, m_Target(nullptr)
-	, m_Path(nullptr)
-	, m_IsFollowingPath(false)
+	, m_PathFinder(a_Width, a_Height)
 	, m_PathfinderActivated(false)
-	, m_GiveUpCounter(0)
-	, m_TicksSinceLastPathReset(1000)
 	, m_LastGroundHeight(POSY_TOINT)
 	, m_JumpCoolDown(0)
 	, m_IdleInterval(0)
@@ -123,118 +120,6 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
 
 
 
-bool cMonster::TickPathFinding(cChunk & a_Chunk)
-{
-	if (!m_PathfinderActivated)
-	{
-		return false;
-	}
-	if (m_TicksSinceLastPathReset < 1000)
-	{
-		// No need to count beyond 1000. 1000 is arbitary here.
-		++m_TicksSinceLastPathReset;
-	}
-
-	if (ReachedFinalDestination())
-	{
-		StopMovingToPosition();
-		return false;
-	}
-
-	if ((m_FinalDestination - m_PathFinderDestination).Length() > 0.25)  // if the distance between where we're going and where we should go is too big.
-	{
-		/* If we reached the last path waypoint,
-		Or if we haven't re-calculated for too long.
-		Interval is proportional to distance squared, and its minimum is 10.
-		(Recalculate lots when close, calculate rarely when far) */
-		if (
-			((GetPosition() - m_PathFinderDestination).Length() < 0.25) ||
-			((m_TicksSinceLastPathReset > 10) && (m_TicksSinceLastPathReset > (0.4 * (m_FinalDestination - GetPosition()).SqrLength())))
-		)
-		{
-			/* Re-calculating is expensive when there's no path to target, and it results in mobs freezing very often as a result of always recalculating.
-			This is a workaround till we get better path recalculation. */
-			if (!m_NoPathToTarget)
-			{
-				ResetPathFinding();
-			}
-		}
-	}
-
-	if (m_Path == nullptr)
-	{
-		if (!EnsureProperDestination(a_Chunk))
-		{
-			StopMovingToPosition();  // Invalid chunks, probably world is loading or something, cancel movement.
-			return false;
-		}
-		m_GiveUpCounter = 40;
-		m_NoPathToTarget = false;
-		m_NoMoreWayPoints = false;
-		m_PathFinderDestination = m_FinalDestination;
-		m_Path = new cPath(a_Chunk, GetPosition(), m_PathFinderDestination, 20, GetWidth(), GetHeight());
-	}
-
-	switch (m_Path->Step(a_Chunk))
-	{
-		case ePathFinderStatus::NEARBY_FOUND:
-		{
-			m_NoPathToTarget = true;
-			m_PathFinderDestination = m_Path->AcceptNearbyPath();
-			break;
-		}
-
-		case ePathFinderStatus::PATH_NOT_FOUND:
-		{
-			StopMovingToPosition();  // Try to calculate a path again.
-			// Note that the next time may succeed, e.g. if a player breaks a barrier.
-			break;
-		}
-		case ePathFinderStatus::CALCULATING:
-		{
-			// Pathfinder needs more time
-			break;
-		}
-		case ePathFinderStatus::PATH_FOUND:
-		{
-			if (m_NoMoreWayPoints || (--m_GiveUpCounter == 0))
-			{
-				if (m_EMState == ATTACKING)
-				{
-					ResetPathFinding();  // Try to calculate a path again.
-					// This results in mobs hanging around an unreachable target (player).
-				}
-				else
-				{
-					StopMovingToPosition();  // Find a different place to go to.
-				}
-				return false;
-			}
-			else if (!m_Path->IsLastPoint())  // Have we arrived at the next cell, as denoted by m_NextWayPointPosition?
-			{
-				if ((m_Path->IsFirstPoint() || ReachedNextWaypoint()))
-				{
-					m_NextWayPointPosition = m_Path->GetNextPoint();
-					m_GiveUpCounter = 40;  // Give up after 40 ticks (2 seconds) if failed to reach m_NextWayPointPosition.
-				}
-			}
-			else
-			{
-				m_NoMoreWayPoints = true;
-			}
-
-			m_IsFollowingPath = true;
-			return true;
-		}
-	}
-
-	return false;
-}
-
-
-
-
-
 void cMonster::MoveToWayPoint(cChunk & a_Chunk)
 {
 	if (m_JumpCoolDown == 0)
@@ -242,8 +127,7 @@ void cMonster::MoveToWayPoint(cChunk & a_Chunk)
 		if (DoesPosYRequireJump(FloorC(m_NextWayPointPosition.y)))
 		{
 			if (
-				(IsOnGround() && (GetSpeedX() == 0.0f) && (GetSpeedY() == 0.0f)) ||
-				(IsSwimming() && (m_GiveUpCounter < 15))
+				(IsOnGround() && (GetSpeedX() == 0.0f) && (GetSpeedY() == 0.0f))  // TODO water handling?
 			)
 			{
 				m_bOnGround = false;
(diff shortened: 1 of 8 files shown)


lib/SQLiteCpp
```diff
@@ -1 +1 @@
-Subproject commit 7e9983b00fb44c2e569ba54bfee6f548496d7f61
+Subproject commit af78d59349fd013341746b345961e3f19ecae60d
```
SafwatHalaby: What's this?

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,226 @@
...
+	, m_IsFollowingPath(false)
+	, m_GiveUpCounter(0)
+	, m_TicksSinceLastPathReset(1000)
+{
+	m_Width = a_MobWidth;
+	m_Height = a_MobHeight;
+
+}
+
+
```
SafwatHalaby: 5 spaces

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,226 @@
...
+			return ePathFinderStatus::PATH_FOUND;
+		}
+
+	default:  // Dummy to get rid of compiler warning.
+		return ePathFinderStatus::PATH_FOUND;
+		break;
+	}
+}
+
+
```
SafwatHalaby: 5 spaces

lib/SQLiteCpp
NiLSPACE: 5b196ef: The library was updated.

src/Mobs/PathFinder.cpp
NiLSPACE: And add braces around it.

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,226 @@
...
+				}
+				RelX = FloorC(m_FinalDestination.x+x) - Chunk->GetPosX() * cChunkDef::Width;
+				RelZ = FloorC(m_FinalDestination.z+z) - Chunk->GetPosZ() * cChunkDef::Width;
+				Chunk->GetBlockTypeMeta(RelX, FloorC(m_FinalDestination.y) - 1, RelZ, BlockType, BlockMeta);
+				if (cBlockInfo::IsSolid(BlockType))
+				{
+					m_FinalDestination.x += x;
+					m_FinalDestination.z += z;
+					InTheAir = false;
+					goto breakBothLoops;
```
NiLSPACE: Do we use goto's?

Though I don't have a problem with it here, since breaking multiple loops is pretty much the only thing it's good for ;)

lib/SQLiteCpp
SafwatHalaby: I'm unfamiliar with submodules. How do I apply the update to the branch?

Seadragon91: Try `git submodule update`

NiLSPACE: @Seadragon91 wouldn't it still use the version from af78d59349fd013341746b345961e3f19ecae60d?

Seadragon91: Yes you are right. @SafwatHalaby if you do git status, does the file still appears? If yes add and commit it and rebase your commit and then the file should be removed from your commit.

SphinxC0re:
clang build fails
```bash
/home/travis/build/cuberite/cuberite/src/Mobs/PathFinder.cpp:115:2: error: 

      default label in switch which covers all enumeration values

      [-Werror,-Wcovered-switch-default]

        default:  // Dummy to get rid of compiler warning.

        ^

/home/travis/build/cuberite/cuberite/src/Mobs/PathFinder.cpp:117:3: error: will

      never be executed [-Werror,-Wunreachable-code]

                break;

                ^~~~~

/home/travis/build/cuberite/cuberite/src/Mobs/PathFinder.cpp:116:10: error: 

      will never be executed [-Werror,-Wunreachable-code]

                return ePathFinderStatus::PATH_FOUND;

                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~

/home/travis/build/cuberite/cuberite/src/Mobs/PathFinder.cpp:78:4: error: will

      never be executed [-Werror,-Wunreachable-code]

                        break;

                        ^~~~~

In file included from /home/travis/build/cuberite/cuberite/src/Mobs/PathFinder.cpp:4:

/home/travis/build/cuberite/cuberite/src/Mobs/PathFinder.h:36:6: error: private

      field 'm_JumpCoolDown' is not used [-Werror,-Wunused-private-field]

        int m_JumpCoolDown;

            ^

5 errors generated.
```

lib/SQLiteCpp
SafwatHalaby: Did I fix this?

SafwatHalaby:
Thanks. I'll take care of the clang issues tomorrow.

Seadragon91:
Rebase the 2 commits with `git rebase -i HEAD~2`, then the file should disappear.

SafwatHalaby:
What file are you referring to? Did it disappear already?

Seadragon91:
You have 2 commits, the second named test has the change about the submdodule `lib/SQLiteCpp`. If the 2 commits get rebased together, then the change about `lib/SQLiteCpp`, won't show anymore in the rebased commit.

SafwatHalaby:
Ok. I think there'll be several more commits so I'll rebase later, just before merging to master.

SphinxC0re:
Rebase onto master:
```bash
git fetch
git rebase -i origin/master
```

madmaxoft:
@SafwatHalaby If you developed this on a branch in the main repo, I could have rebased and fixed the submodule for you. Please do that next time, it's much easier both for the review and for fixes like these :)

madmaxoft:
As for the clang errors, the current solution we use is
```cpp
#ifndef __clang__
  default:
  {
    ...
  }
#endif
```

src/Mobs/PathFinder.cpp
SafwatHalaby: ...or for [stack unwinding](http://programmers.stackexchange.com/a/154980/131807) ;)

I think double-breaking is a perfectly valid use case. Does anyone have an objection?

madmaxoft: I'd really prefer to live without them, if at all possible.

In this case, I'd consider replacing the two `for`-loops with a single one by listing all the coords it checks in a `static const` array of 2D coords and iterating over that array, using a regular `break` instead of the `goto`.

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,221 @@
...
+	Chunk->GetBlockTypeMeta(RelX, FloorC(m_FinalDestination.y) - 1, RelZ, BlockType, BlockMeta);
+	if (!cBlockInfo::IsSolid(BlockType))
+	{
+		bool InTheAir = true;
+		int x, z;
+		for (z = -1; z <= 1; ++z)
+		{
+			for (x = -1; x <= 1; ++x)
+			{
+				if ((x==0) && (z==0))
```
madmaxoft: StyleCheck should have picked these comparisons

SafwatHalaby:
@madmaxoft - Clang was complaining about some unreachable code paths. It doesn't seem to be complaining about the blank `default`. Should I add the preprocessor wrapper anyways?

src/Mobs/PathFinder.cpp
SafwatHalaby: It's worth noting that these had been lying in cMonster since forever.

SafwatHalaby:
Regarding developing on the cuberite branch: Will do :)

SafwatHalaby:
Would it be safe if I remove my cloned repo from Github?

src/Mobs/PathFinder.cpp
SafwatHalaby: That sounds dirter than the `goto` to me. What would you've done if the iteration was much longer?

SphinxC0re:
Nope. If you do that, all your changes included in this PR will be removed as well

SafwatHalaby:
I mean for future PRS and such, not now.

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "../Chunk.h"
+
+cPathFinder::cPathFinder(double a_MobWidth, double a_MobHeight)
+	: m_Path(nullptr)
+	, m_GiveUpCounter(0)
+	, m_TicksSinceLastPathReset(1000)
+{
+	m_Width = a_MobWidth;
+	m_Height = a_MobHeight;
+
```
tigerw: Extra carriage return?

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+#include "Globals.h"
+#include "PathFinder.h"
+#include "../Chunk.h"
+
+cPathFinder::cPathFinder(double a_MobWidth, double a_MobHeight)
+	: m_Path(nullptr)
+	, m_GiveUpCounter(0)
```
tigerw: Consider comma use in constructor variable initialisation to be placed after variables, and colon use to be placed after the constructor definition as:

```cpp
cClass:cClass(...) :
   m_A(...),
   m_B(...),
   m_C(...)
```

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,233 @@
...
+	, m_TicksSinceLastPathReset(1000)
```
tigerw: Any chance of resetting when new path is wanted, instead of counting ticks? :)

src/Mobs/Monster.cpp
```diff
@@ -453,29 +230,43 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
-			StopMovingToPosition();
-			m_GiveUpCounter = 40;  // This doesn't count as giving up, keep the giveup timer as is.
-		}
-		else
+		// Note that m_NextWayPointPosition is actually returned by GetNextWayPoint)
+		switch (m_PathFinder.GetNextWayPoint(*Chunk, GetPosition(), m_FinalDestination, m_NextWayPointPosition))
 		{
-			MoveToWayPoint(*Chunk);
+			case ePathFinderStatus::PATH_FOUND:
+			case ePathFinderStatus::NEARBY_FOUND:
```
tigerw: Missed some braces.

src/Mobs/Monster.cpp
```diff
@@ -453,29 +230,43 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+				}
+				else
+				{
+					a_IsFollowingPath = true;  // Used for proper body / head orientation only.
+					MoveToWayPoint(*Chunk);
+				}
+				break;
+			default:
+			{
+
```
tigerw: Is this a compiler warning thing?

src/Mobs/PathFinder.cpp
SafwatHalaby: You mean recalculating solely based on distance as you proposed earlier in `MoveToPosition`?

src/Mobs/Monster.cpp
SafwatHalaby: I don't remember, will check.

src/Mobs/PathFinder.cpp
SafwatHalaby: Would you mind explaining how exactly that'd be done? I find counting ticks essential:
 - It limits the frequency of how many times a new cPath is created, which is good for performance.
 - If recalc is based solely on distance between the final location and the destination last given to the pathfinder (those two were once called `FinalDestination` and `PathFinderDestination`), recalculation may not happen when it should. (I got an example if you'd like)

SafwatHalaby: The current algorithm is somewhat like this:
- If the destination changes:
  - If TicksSinceLastPathReset > X
       Recalculate
X is a number whose value is dependent on the distance to target: The further we are from target, the larger the X. In other words: When near, recalculate often, when far, recalculate less often.

Do you have a better proposal which doesn't need `TicksSinceLastPathReset`?

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,233 @@
...
+		++m_TicksSinceLastPathReset;
+	}
+
+	/* if (ReachedFinalDestination())
+	{
+		StopMovingToPosition();
+		return false;
+	} */
+
+	if ((m_FinalDestination - m_PathDestination).Length() > 0.25)  // if the distance between where we're going and where we should go is too big.
```
worktycho: Why not use SqrLength? `(m_FinalDestination - m_PathDestination).SqrLength() > 0.25 * 0.25` is faster and just as clear.

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,233 @@
...
+	{
+		/* If we reached the last path waypoint,
+		Or if we haven't re-calculated for too long.
+		Interval is proportional to distance squared, and its minimum is 10.
+		(Recalculate lots when close, calculate rarely when far) */
+		if (
+			((m_Source - m_PathDestination).Length() < 0.25) ||
```
worktycho: SqrLength?

src/Mobs/PathFinder.h
```diff
@@ -0,0 +1,92 @@
...
+private:
+
+	/** The width of the Mob which owns this PathFinder. */
+	double m_Width;
+
+	/** The height of the Mob which owns this PathFinder. */
+	double m_Height;
+
+	/** The current cPath instance we have. This is discarded and recreated when a path recalculation is needed. */
+	cPath *  m_Path;  // TODO unique ptr
```
worktycho: Make this unique_ptr at the same time?

src/Mobs/Villager.cpp
```diff
@@ -158,7 +158,7 @@ void cVillager::HandleFarmerPrepareFarmCrops()
 void cVillager::HandleFarmerTryHarvestCrops()
 {
 	// Harvest the crops if the villager isn't moving and if the crops are closer then 2 blocks.
-	if (!m_IsFollowingPath && (GetPosition() - m_CropsPos).Length() < 2)
+	if (!m_PathfinderActivated && (GetPosition() - m_CropsPos).Length() < 2)
```
worktycho: SqrLength?

src/Mobs/PathFinder.cpp
SafwatHalaby: That's a good idea.

src/Mobs/PathFinder.h
SafwatHalaby: How about I just use a `cPath m_Path`?
To re-initialize, I'd just do:
`m_Path = cPath(...);`

worktycho: Thats better. I just thought you had a reason for making it a pointer.

src/Mobs/Monster.cpp
SafwatHalaby: Yes, this gets rid of some warnings.

src/Mobs/PathFinder.cpp
SafwatHalaby: I've simply mimicked the cMonster style. Should I change both?

src/Mobs/PathFinder.h
SafwatHalaby: I thought so too. It seems a pointer is needed for line 57, but I'll rewrite that.

SafwatHalaby: I changed m_Path from a pointer to a normal class, but I had to do a little hack to accomplish this. Is there a better way?

SafwatHalaby:
One of the commits ruined idle mobs' pathfinding, chasing monsters are also pausing every now and then. Looking into it.

src/Mobs/Monster.cpp
SafwatHalaby: Fixed.

SphinxC0re:
Try bisecting the commit history using ```git bisect```

SafwatHalaby:
[This](https://github.com/SafwatHalaby/cuberite/commit/de76c864d1450a1ec5942af1c4f3d0bf7d10a1b4) is hacky, it aims to turn cPathFinder's m_Path from a pointer to a regular class variable. Is there a better way?

SafwatHalaby:
>Try bisecting the commit history using `git bisect`

Thanks, didn't know of this before.

src/Mobs/PathFinder.cpp
tigerw: My suggestion remains the same, to recalculate if there is a change in where the entity should be going.

I suppose it comes down to a matter of performance versus accuracy. For selective recalculation, you'll definitely encounter terrain layouts where even a one block change in destination could mean a completely different path, and so a player would see a monster walking towards them, stop, and turn back to head along a different route. This may or may not be noticeable depending on how large X is set to be. For immediate recalculation, the monster may not move at all if the pathfinder is particularly slow and the player is in constant motion, but that really depends on exactly how fast cPath currently is, which I have no idea.

So in summary, I haven't enough data to really know. :/
Perhaps you could do some tests? :)

SafwatHalaby: >I suppose it comes down to a matter of performance versus accuracy.

It is, up to a certain point. 

Beyond that point, you're just killing performance without gaining noticable accuracy. For instance, there is very little visible difference between the current ticking method and your proposed 1-block recalc method. But there is a huge performance difference. Also, once I make the mobs ditch the pathfinder and just move in a straight line towards target when there is a clear line of sight to target, you'd have to be really sharp eyed to see any difference at all. So if we go the 1-block-recalc route, we trade performance away and practically gain nothing.

>For selective recalculation, you'll definitely encounter terrain layouts where even a one block change in destination could mean a completely different path, and so a player would see a monster walking towards them for X ticks, stop, and turn back to head along a different route

Ah, yes. I like to call it the "meta-stable-spot problem": The mob is at a special point where an inch of player movement changes the shortest path completely. That problem is present in both methods of recalculation in different forms. In some cases the delayed recalc is actually an advantage here: The mob just moves along one path (and doesn't recalc), making it leave the meta-stable-spot faster. In other terrain cases 1-block-recalc may behave better. You can't make a case in favor of 1-block-recalc here. I think the problem is too rare and too harmless to be worth taking into account anyways.

I see no single rational reason in favor of 1-block-recalc. If you do have one, please enlighten me.

Why not to use 1-block-recalc:
 - It calls cPath every single tick. That's much slower and not justifiable. 
 - We already have a working code for the tick+distance method.
 - The accuracy advantage is barely noticbale, will be virtually unnoticable when line-of-sight optimization is done.
 - That meta-stable spot issue is present in both methods.
 - Why would a mob which is 40-block away from a player recalculate every inch of player movement? That doesn't make sense. 
 - The current algorithm increases the recalc frequency significantly as the player moves closer, up to a certain max frequency, we can increase that max frequency if needed. If necessary, we could  remove the max frequency restriction altogether, making it calculate as frequent as 1-block-recalc on close range, and less often on long range. But even that much accuracy is unneeded.

src/Mobs/PathFinder.cpp
SafwatHalaby: Perhaps the best fix would be moving the two nested loops into a function, and using `return` instead of a `goto`.

SafwatHalaby:
cPathFinder is too buggy. Closing this, will re-implement recalc from scratch once we settle #2536

PR #2525, opened 2015-10-05
Improved the data structure for storing loaded chunks

worktycho:
Should significantly increase the performance of deciding which chunks to stream

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -456,7 +456,7 @@ bool cClientHandle::StreamNextChunk(void)
 				// If the chunk already loading / loaded -> skip
 				if (
 					(m_ChunksToSend.find(Coords) != m_ChunksToSend.end()) ||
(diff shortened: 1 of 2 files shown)


madmaxoft:
Looks good to me.

src/ClientHandle.cpp
```diff
@@ -525,7 +525,7 @@ void cClientHandle::UnloadOutOfRangeChunks(void)
 	cChunkCoordsList ChunksToRemove;
 	{
 		cCSLock Lock(m_CSChunkLists);
-		for (cChunkCoordsList::iterator itr = m_LoadedChunks.begin(); itr != m_LoadedChunks.end();)
+		for (auto itr = m_LoadedChunks.begin(); itr != m_LoadedChunks.end();)
```
SphinxC0re: why can't this be a c++11 loop?
```cpp
for (auto chunk : m_LoadedChunks)
{
..
}
```

worktycho: Because it does an erase in the loop body.

PR #2526, opened 2015-10-05
Converted to android-cmake for APK creation

tigerw:
* Various changes to allow for Android to compile successfully


Android/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+cmake_minimum_required (VERSION 2.8.7)
+project(MCServer)
+include(ExternalProject)
+
+externalproject_add(lua_native
+	SOURCE_DIR ../lib/lua
+	CMAKE_GENERATOR "Visual Studio 14 2015"
```
madmaxoft: I don't think this is the right way to go - what about building this on Linux?

tigerw:
So with regards to this, I'm planning to merge the core server related changes in, then remove the Android folder and place that into cuberite-android, and add the main repository as a submodule.

Android/CMakeLists.txt
tigerw: It's definitely not right, but conversely, I have no idea how to get the autodetection working as it does when you run `cmake .`

Android/.gitignore
```diff
@@ -0,0 +1,9 @@
+/bin/
+/libs/
+/MCServer/
+/CMakeFiles/
+/gen/
+/lua_native-prefix/
+/tolua_native-prefix/
+CMakeCache.txt
+local.properties
```
SphinxC0re: missing newline

Android/AndroidNdkModules.cmake
```diff
@@ -0,0 +1,58 @@
...
+        android_ndk_import_module_native_app_glue()
+        
+        include_directories(${ANDROID_NDK}/sources/android/ndk_helper)
+        file(GLOB _NDK_HELPER_SRCS ${ANDROID_NDK}/sources/android/ndk_helper/*.cpp ${ANDROID_NDK}/sources/android/ndk_helper/gl3stub.c)
+        add_library(ndk_helper ${_NDK_HELPER_SRCS})
+        target_link_libraries(ndk_helper log android EGL GLESv2 cpufeatures native_app_glue)
+        
+        unset(_NDK_HELPER_SRCS)
+    endif()
+endmacro()
```
SphinxC0re: Missing newline

Android/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+cmake_minimum_required (VERSION 2.8.7)
+project(MCServer)
```
SphinxC0re: shouldn't this be
```cmake
project(Cuberite)
```
?

SphinxC0re:
You should really update your branch

tigerw:
I have updated my branch - if you're referring to `project (MCServer)`, I did it before the rename en-masse.

worktycho:
What about warnings? It looks like you've left it a gcc's defaults, which is particularly problematic for non-x86 platforms like android where assumptions are more likely to be violated.

Android/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+	BUILD_COMMAND
+		${CMAKE_COMMAND} --build <BINARY_DIR> --config Release
+	INSTALL_COMMAND ""
+)
+
+# Tell libevent that Android has pthread (not auto detected due to NDK implementation problems)
+# Tell Lua to not use unimplemented Android locale features
+add_definitions(-Dgetlocaledecpoint\(\)="'.'")
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14") # Add c++14 support here since we're not calling set_flags()
```
worktycho: We are using c++11. Theres no point asking for the extra features we can't use.

worktycho:
What compile options does this build with? Experience of the pi has shown that cuberite requires several non-standard compile options on arm.

- `-fomit-frame-pointer` : mbedtls miscompiles on armv6 without this
- `-fsigned-char` : various bugs in cuberite's core are exposed if this is not enabled.
- `-ffast-math` : improves performance, by allowing the compiler to break the IEEE 754 standard

tigerw:
Have a look at the android cmake toolchain file, I think that the first two are present. I'm not sure about the last, but I'm assuming that since soft-float is so slow, there won't be a particularly noticeable speed increase.

Android/CMakeLists.txt
tigerw: I think it's okay to use C++14 here since Android builds are using a downloaded compiler set which is certain to include the relevant support.

tigerw:
About warnings; what are good warnings? Since the plan is to include the core/desktop repository as a submodule, perhaps warnings would only really be effective when generated compiling the desktop version?

Android/CMakeLists.txt
tigerw: Would anyone have any idea about this? It could be a showstopper for integration into @bearbin's Jenkins server.

worktycho:
-fomit-frame-pointer is only passed in release, we need it passed in debug as well.
fast-math is especially important on soft float systems, because optimizations will have more of an effect on slower operations.

As for warnings, they are the real problem with the submodule strategy. We don't want warnings passed to library code but we do for our code. Why can you not have non-platform specific flags handled by setflags?

tigerw:
Okay, I'll try to SetFlags then. I didn't use it since I really couldn't be bothered to trawl through the dense file and correct the "If not Windows, is Linux" and "Linux is all the same" behaviour.

worktycho:
Theres very little that assumes Linux. It does tend to assume a POSIXish environment if not windows, but android should be POSIXish. And we have enough special cases for POSIX, so android should be simpler. Android is closer to Linux than FreeBSD and OS X. Look at the FreeBSD build cases for examples.

SphinxC0re:
Any updates?

SphinxC0re:
Is this still in progress or did you drop this? @tigerw

Android/CMakeLists.txt
worktycho: Asking for c++14 can also result in broken code, if someone wants to copy android specific code to general codebase.

tigerw:
Still in progress.

tigerw:
Now?

SetFlags.cmake
```diff
@@ -86,7 +86,11 @@ macro(set_flags)
 		else()
 			add_flags_cxx("-pthread")
 		endif()
-
+	elseif(ANDROID)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
worktycho: c++11 only. I don't want to have to deal with more compiler variation than we have to. And yes, c++14 has caused breaking changes for template free functions with the same name as a new standard library function. (ADL makes it impossible to add template free functions without causing breaking changes).

lib/jsoncpp
```diff
@@ -1 +1 @@
-Subproject commit 81cf237917b6873decd27e15b7fe8473003a2762
+Subproject commit beae99924f29d88405576d5b9787bd494ec13e7a
```
worktycho: Why?

src/Bindings/ManualBindings_Network.cpp
```diff
@@ -878,11 +878,11 @@ static int tolua_cUDPEndpoint_Send(lua_State * L)
 
 	// Get the data to send:
 	AString Data, RemotePeer;
-	int RemotePort = 0;
+	UInt16 RemotePort = 0;
 	S.GetStackValues(2, Data, RemotePeer, RemotePort);
 
 	// Check the port:
-	if ((RemotePort < 0) || (RemotePort > USHRT_MAX))
+	if ((RemotePort < 0) || (RemotePort > std::numeric_limits<UInt16>::max()))
```
worktycho: This is never true. Use the checks in GetStackValues.

Android/CMakeLists.txt
worktycho: Looking at this again, is their any reason not to use the CMAKE_GENERATOR variable? You're always going to want to use the same generator as the main project.

Android/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+externalproject_add(lua_native
+	SOURCE_DIR ../lib/lua
+	CMAKE_GENERATOR "Visual Studio 14 2015"
+	CMAKE_ARGS -DANDROID_CROSSCOMPILE=TRUE
+	BUILD_COMMAND
+		${CMAKE_COMMAND} --build <BINARY_DIR> --config Release
+	INSTALL_COMMAND ""
+)
+externalproject_add(tolua_native
+	SOURCE_DIR ../lib/tolua++
```
worktycho: How is this referencing the lua target?

src/Bindings/ManualBindings_Network.cpp
tigerw: Oh yeah!

Then in that case, when are the checks in GetStackValues triggered? Is it something to do with a difference between lua_Number and the template type T? Is lua_Number bigger?

Android/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+	../lib/
+	../src/
+	../lib/jsoncpp/include/
+	../lib/polarssl/include/
+	../lib/sqlitecpp/include/
+	../lib/sqlitecpp/sqlite3/
+	../lib/libevent/include/
+)
+
+# Probably not the recommended method, but other ways don't seem to generate libraries in libs/armeabi
```
worktycho: Move libs after build?

Android/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+# Tell libevent that Android has pthread (not auto detected due to NDK implementation problems)
+# Tell Lua to not use unimplemented Android locale features
+add_definitions(-Dgetlocaledecpoint\(\)="'.'")
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
+set(CROSSCOMPILE 1)
+
+add_library(AndroidNative jni/app-android.cpp)
+add_dependencies(tolua_native lua_native)
+add_dependencies(AndroidNative tolua_native)
```
worktycho: Why does the androidNative lib depend on tolua_native?

src/ClientHandle.h
```diff
@@ -453,7 +453,12 @@ class cClientHandle  // tolua_export
 		// TODO: Add Kicking here as well
 	} ;
 	
+#ifdef ANDROID
+#pragma message "Android <atomic> support is incomplete or broken - using plain variable (subject to data races!)"
+	eState m_State;
```
worktycho: This is not an acceptable solution. If we can't use std::atomic, then we need to setup typedefs to a custom std::atomic implementation using cCriticalSection.

src/FastRandom.cpp
```diff
@@ -8,8 +8,11 @@
 
 #include <random>
 
-#if defined (__GNUC__)
-	#define ATTRIBUTE_TLS static __thread
+#if defined (ANDROID)
+	// Tycho: I'm sorry, I'm sorry, I'm so, so, sorry, but please blame the NDK developers for lack of, thread_local
+	#define ATTRIBUTE_TLS static
```
worktycho: Really? What about __thread?

src/Bindings/ManualBindings_Network.cpp
worktycho: GetStackValues will return false if the input value is not a number or falls outside the range of the type you have requested conversion to.

src/FurnaceRecipe.cpp
```diff
@@ -70,6 +70,10 @@ void cFurnaceRecipe::ReloadRecipes(void)
 		LineNum++;
 		if (ParsingLine.empty())
 		{
+			// There is a problem here on Android. Text files transferred from another OS may have a newline representation Android's implementation of getline doesn't expect
+			// Thus, part of a newline may be left in ParsingLine. ::empty() thus thinks the string isn't empty, and the below code outputs interesting errors since it was passed a nearly empty string
+			// Ref: http://stackoverflow.com/questions/6089231/getting-std-ifstream-to-handle-lf-cr-and-crlf
+			// TODO: There is a solution in the above reference, but it isn't very pretty. Fix it somehow.
```
worktycho: Solution: Don't build on windows :smile: , but seriously, I'd say that we should use the wrapper from the question if it wasn't for licensing issues. The other short term option is to just tell anyone building the android app on window to set git to use unix line endings. Most text editors can handle it.

src/LoggerListeners.cpp
```diff
@@ -295,6 +297,27 @@ std::unique_ptr<cLogger::cListener> MakeConsoleListener(bool a_IsService)
 
 ////////////////////////////////////////////////////////////////////////////////
 // cFileListener:
+#if defined (ANDROID)
```
worktycho: Create a separate listener.

src/OSSupport/CMakeLists.txt
```diff
@@ -47,10 +47,13 @@ if(NOT MSVC)
...
 
 	if(UNIX)
-		if(NOT APPLE)
+		if(APPLE)
+			target_link_libraries(OSSupport pthread event_core event_extra)
+		elseif(ANDROID)
+			# Android has neither pthread nor rt
+			target_link_libraries(OSSupport)
+		else()
 			target_link_libraries(OSSupport rt)
```
worktycho: Everyone needs event_core and event_extra

worktycho: And not apple needs pthread.

src/OSSupport/File.cpp
```diff
@@ -639,9 +639,13 @@ unsigned cFile::GetLastModificationTime(const AString & a_FileName)
...
 		return 0;
 	}
-	#ifdef _WIN32
+	#if defined(_WIN32)
 		// Windows returns times in local time already
 		return static_cast<unsigned>(st.st_mtime);
+	#elif defined(ANDROID)
+		// Identical to Linux below, but st_mtime is an unsigned long, so cast is needed:
+		auto Time = static_cast<time_t>(st.st_mtime);
+		return static_cast<unsigned>(mktime(localtime(&Time)));
```
worktycho: Why not just change the general linux code to be this?

src/OSSupport/NetworkInterfaceEnum.cpp
```diff
@@ -8,10 +8,10 @@
 #include "event2/util.h"
 #include "../SelfTests.h"
 
-#ifdef _WIN32
+#if defined(_WIN32)
 	#include <IPHlpApi.h>
 	#pragma comment(lib, "IPHLPAPI.lib")
-#else  // _WIN32
+#elif !defined(ANDROID)  // _WIN32
 	#include <sys/types.h>
```
worktycho: How are you including networking headers on andriod?

src/OSSupport/NetworkInterfaceEnum.cpp
```diff
@@ -121,7 +121,7 @@ AStringVector cNetwork::EnumLocalIPAddresses(void)
 {
 	AStringVector res;
```
worktycho: So you're just breaking this function on android?

worktycho: If you are going to break it, at least update the API docs.

src/Protocol/MojangAPI.cpp
```diff
@@ -584,7 +584,7 @@ void cMojangAPI::SaveCachesToDisk(void)
 	try
 	{
 		// Open up the SQLite DB:
-		SQLite::Database db("MojangAPI.sqlite", SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE);
+		SQLite::Database db(FILE_IO_PREFIX "MojangAPI.sqlite", SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE);
```
worktycho: What about the SQLite lua bindings? Don't they need the prefix as well?

src/OSSupport/NetworkInterfaceEnum.cpp
tigerw: Liberal application of magic.

src/OSSupport/File.cpp
tigerw: Do compilers complain if you cast to the same type?

src/LoggerListeners.cpp
tigerw: awww, do I have to

src/FastRandom.cpp
tigerw: Apparently only GCC emulates thread local storage and not Clang (yet), but GCC crashes when compiling Lua.

Also, imagining your expression when you read the downgrading of TLS to just the plain static, and then saying, "really?", makes me laugh :P

src/ClientHandle.h
tigerw: nooooooooooooo

src/OSSupport/NetworkInterfaceEnum.cpp
worktycho: Meaning? Is it just that this file isn't actually doing anything?

src/OSSupport/File.cpp
worktycho: I don't think so. But test it.

src/LoggerListeners.cpp
worktycho: Yes.

src/FastRandom.cpp
worktycho: Just leaving it as a static is UB so not really an option. So we either need to use atomics (is this possible on android?), the pthread thread local api (slow), or work out some way of using a mutex (slow).

src/ClientHandle.h
worktycho: Unless you want to review all the code that uses std::atomic, and come up with alternative implementations, I think a custom minimal atomic implementation is easier.

SetFlags.cmake
tigerw: You won’t be dealing with it, considering this will be in a separate repository. It’ll probably end up being me dealing with it :P

Android/CMakeLists.txt
tigerw: I’m not sure.
•	If you mean how it’s doing the referencing, I’m still not sure.
•	If you mean it’s not referencing anything but that I’m saying that it is, then I’m confused.

Android/CMakeLists.txt
tigerw: > “is there any reason not to use...”
Did you mean to use? If you did, I think it didn’t work if I didn’t specify that variable, but I’ll try again soon.

Android/CMakeLists.txt
tigerw: I thought it was more like tolua_native depended on AndroidNative. My intention was to get tolua_native and lua_native to actually build.

Android/CMakeLists.txt
tigerw: ehhh naah 😊

src/ClientHandle.h
tigerw: I don't think you should be bothered about this, considering it may well be fixed anyway in future NDK releases and Android runs at *0.8 chunks per second* and so is unlikely to be used as an actual server. I definitely don't have the time to try and re-implement atomic :P

src/FastRandom.cpp
tigerw: They are working on it. Unless someone is actually really in need of running a reliable server on their phone, it doesn't matter in my humble opinion.

worktycho: Why not just replace cCriticalSection with a noop? It will offer a major performance boost on these low power devices? Because we need thread safety. Any instance of a race condition is grounds for the compiler to do whatever it feels like.

src/ClientHandle.h
worktycho: As I said, if thread safety is less important than performance, you should make cCriticalSection a noop. A minimal std::atomic implantation using cCriticalSection would not be much work. It doesn't need to be fully standard complaint, just provide the interfaces we use.

Android/CMakeLists.txt
worktycho: Then you should add a dependency on tolua_native in src/Bindings/CmakeLists.txt

SetFlags.cmake
worktycho: Then let me advise you to stick to c++11 to avoid having a similar problem to when a PR wouldn't compile on windows because msvc exposes std::make_unique, which was causing issues with my make_unique if it was in the global namespace.

src/FastRandom.cpp
tigerw: I am intending for this to be an experimental thing. Of course there will be bugs, but there are loads of them anyway, and the compiler doesn't seem to be doing anything untoward.

In an ideal situation, everything would be standards compliant, but since I've no time to reimplement <atomic> nor wish to pollute the main repository with more Android artefacts, it's gonna be a case of buggy mobile builds versus none at all.

Does that sit comfortable on your conscience??

...unless you want to do something code-y about it?

Android/CMakeLists.txt
worktycho: I've only just seen this after your reference in #3161. But what I meant was why cant you do `CMAKE_GENERATOR ${CMAKE_GENERATOR}`?

tigerw: > Would anyone have any idea about this? It could be a showstopper for integration...

...and @worktycho comes to the rescue. :)

Schwertspize:
@tigerw I would recommend that you keep this one closed, re-write your branch to not generate an apk, instead binaries, the NDK can do that 😛. Then open a new pr

PR #2533, opened 2015-10-08
Fixed bindings for cTCPLink:StartTLSClient().

madmaxoft:
The bindings used to enforce client cert, now it's optional.

The only downside is that there's no error message in case the params are of a completely wrong type.


NiLSPACE:
Looking at the code I'm guessing this wasn't the reason why connecting to https servers did not work.

madmaxoft:
No, this was just a side quest I stumbled upon while investigating the https issue.

PR #2538, opened 2015-10-09
Exported cClientHandle:SendParticleEffect and SendSoundParticleEffect

NiLSPACE:
This Would make it possible to implement something like [this](http://forum.mc-server.org/showthread.php?tid=870&pid=22839#pid22839).

### DIFF
--- a/src/ClientHandle.h
+++ b/src/ClientHandle.h
@@ -180,8 +180,8 @@ public:  // tolua_export
 	void SendInventorySlot              (char a_WindowID, short a_SlotNum, const cItem & a_Item);
 	void SendMapData                    (const cMap & a_Map, int a_DataStartX, int a_DataStartY);
 	void SendPaintingSpawn              (const cPainting & a_Painting);
-	void SendParticleEffect             (const AString & a_ParticleName, float a_SrcX, float a_SrcY, float a_SrcZ, float a_OffsetX, float a_OffsetY, float a_OffsetZ, float a_ParticleData, int a_ParticleAmount);
-	void SendParticleEffect             (const AString & a_ParticleName, const Vector3f a_Src, const Vector3f a_Offset, float a_ParticleData, int a_ParticleAmount, std::array<int, 2> a_Data);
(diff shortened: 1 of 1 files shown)


worktycho:
I'm not conformable that this is the right interface, and you need documentation to export the functions.

NiLSPACE:
I knew I forgot something :/ I'll wait with the documentation in case a better idea for this arrives or if this is alright after all.

src/ClientHandle.h
```diff
@@ -180,8 +180,8 @@ class cClientHandle  // tolua_export
 	void SendInventorySlot              (char a_WindowID, short a_SlotNum, const cItem & a_Item);
 	void SendMapData                    (const cMap & a_Map, int a_DataStartX, int a_DataStartY);
 	void SendPaintingSpawn              (const cPainting & a_Painting);
-	void SendParticleEffect             (const AString & a_ParticleName, float a_SrcX, float a_SrcY, float a_SrcZ, float a_OffsetX, float a_OffsetY, float a_OffsetZ, float a_ParticleData, int a_ParticleAmount);
-	void SendParticleEffect             (const AString & a_ParticleName, const Vector3f a_Src, const Vector3f a_Offset, float a_ParticleData, int a_ParticleAmount, std::array<int, 2> a_Data);
+	void SendParticleEffect             (const AString & a_ParticleName, float a_SrcX, float a_SrcY, float a_SrcZ, float a_OffsetX, float a_OffsetY, float a_OffsetZ, float a_ParticleData, int a_ParticleAmount);  // tolua_export
+	void SendParticleEffect             (const AString & a_ParticleName, const Vector3f a_Src, const Vector3f a_Offset, float a_ParticleData, int a_ParticleAmount, std::array<int, 2> a_Data);  // tolua_export
```
madmaxoft: How does tolua export even handle the `std::array`? You might want to try the exported function in a sample, perhaps in the Debuggers plugin, to see if it even works at all.

worktycho: It exports it as a user type, but doesn't provide any constructors. So the function is impossible to use.

NiLSPACE:
I'm closing this for now. I don't really have the time to work on it.

PR #2540, opened 2015-10-09
Do not try to path to and from locations outside the world

### LINKED ISSUE #2539: Server crash in the PathFinder
The PathFinder trys to get a block that is out of boundarys of m_Map.
Stack Trace:
http://pastebin.com/rB3T0jNL
World:
https://www.dropbox.com/sh/7upzmuqo4xkhakj/AAArZGAyy0xTQR2ygbjkPrexa?dl=0
edit: (faster mirror):
https://mega.nz/#!9pZzxIiZ!vcxiLu2kKXDgrh30u7MYFH2LUuNrcGH4POYo1YPfevI

worktycho:
May Fix #2539

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -51,6 +51,14 @@ cPath::cPath(
 	m_Source.x = FloorC(a_StartingPoint.x - HalfWidthInt);
 	m_Source.y = FloorC(a_StartingPoint.y);
 	m_Source.z = FloorC(a_StartingPoint.z - HalfWidthInt);
+	
+	if ((m_Source.y <= 0) || (m_Destination.y <= 0)
+		|| (m_Source.y > 256) || (m_Destination.y > 256))
+	{
+		// The source or destination is outside the world, so cannot path
+		m_Status = ePathFinderStatus::PATH_NOT_FOUND;
+		return;
+	}
 
 	m_Destination.x = FloorC(a_EndingPoint.x - HalfWidthInt);
 	m_Destination.y = FloorC(a_EndingPoint.y);


SafwatHalaby:
To make this future-proof, we need a similar check at `ProcessIfWalkable`'s `a_Location` variable, that would prevent this from happening when we implement jumping / climbing
 
```
	if ((a_Location.y <= 0) || (a_Location.y > 256))
	{
		return;
	}
```

SafwatHalaby:
After a second thought, it'd be unclean to put the same check several times, and we may miss some places.

Putting it once in `IsSolid(...)` should globally fix this. 

```
bool cPath::IsSolid(const Vector3i & a_Location)
{
   if Vector3i is outside the world, return true
}
```

This makes the new check you added at `cPath::cPath(...)` redundant. because line 55 would catch it.

src/Mobs/Path.cpp
```diff
@@ -51,6 +51,14 @@ cPath::cPath(
 	m_Source.x = FloorC(a_StartingPoint.x - HalfWidthInt);
 	m_Source.y = FloorC(a_StartingPoint.y);
 	m_Source.z = FloorC(a_StartingPoint.z - HalfWidthInt);
+	
+	if ((m_Source.y <= 0) || (m_Destination.y <= 0)
+		|| (m_Source.y > 256) || (m_Destination.y > 256))
```
madmaxoft: Proper formatting:
```cpp
if (
	(m_Source.y <= 0) || (m_Destination.y <= 0) ||
	(m_Source.y > 256) || (m_Destination.y > 256)
)
{
	...
```

SafwatHalaby: I think this should be moved to `IsSolid(...)` as my other comment suggests.

madmaxoft: I don't think that would be proper, that's making the function do something that it isn't supposed to do.

Instead, `IsSolid` should assert that the coords given to it are valid, and anything that calls `IsSolid` should make sure the coords are valid. Ideally, paths for invalid coords shouldn't be created at all.

SafwatHalaby: That would add extra checks which are spread in cPath, imho it's much more error prone and less future-proof. Why not simply redefine what `IsSolid` does and avoid code dupes?

> that's making the function do something that it isn't supposed to do.

I like to look at it differently: That makes the function assume there's a solid roof just above the world and a solid ground just below the world.

SafwatHalaby: We could rename it to `IsSolidOrOutSideworld` or something.

madmaxoft: That's a bad assumption - it would mean that mobs would try to walk on top of the void in worlds where there's no bedrock (End, SkyBlock etc.)

It's exactly future-proof, by stating that only valid coords can be given to it.

SafwatHalaby: >That's a bad assumption - it would mean that mobs would try to walk on top of the void in worlds where there's no bedrock (End, SkyBlock etc.)

That's a very good point.

SafwatHalaby:
Just a followup: This PR isn't enough and neither are my two suggestions.

src/Mobs/Path.cpp
worktycho: One option is to treat outside the world as always empty. For the bottom it might try to path in to holes, but give up when it can't find a ground to walk on.

SafwatHalaby: Sounds good to me.

SafwatHalaby:
```
bool cPath::IsSolidAndValid(const Vector3i & a_Location)
{
   if Vector3i is outside the world, return false
   ...
}
```

SafwatHalaby:
```
//previously called IsSolid
bool cPath::IsSolidAndValid(const Vector3i & a_Location)
{
   if Vector3i is outside the world, return false
   ...
}
```

SafwatHalaby: It won't even try to path through the holes.

SafwatHalaby: @worktycho This is problamatic for the top of the world, consider a block built on the topmost valid Y coord, it may try to go on top of it.

SafwatHalaby: Perhaps a hybrid solution: if >255 then consider it a solid, if <0 consider it empty.

worktycho: Its should not problematic for a mob to sit on top of the world. The entities should be able to be outside the world. Players can certainly pass over the top of such blocks, and the code for mobs and players are the same in this instance.

SafwatHalaby: Can instances be below 0 as well? If so, why did it even crash?

worktycho: It crashed because the pathfinder asked what a block outside the world was. The crash was nowhere near any entity code.

madmaxoft:
I still think the proper way to go around this would be to make the pathfinder assume (and assert) that the coords given to it are always valid, and any external code calling into the pathfinder must make sure this holds.
Because for mobs outside the world, there will need to be special handling anyway.

worktycho:
I don't think so. For mobs outside the world, they are always falling and thus cannot path. If we just treat all bocks outside the world as air, I think that will cause the pathfinder to behave exactly as we want.

madmaxoft:
Exactly - they fall, and therefore need no pathfinding at all. So not calling the pathfinder is an optimization.

SafwatHalaby:
@madmaxoft , according to what @worktycho , blocks outside the world are in fact air, so we aren't even cheating here. So, `IsSolid(x,-10,z)` **SHOULD** return false. Am I missing something?

SafwatHalaby:
>SHOULD return false

...Rather than fail an assert.

SafwatHalaby:
Regarding optimization: We can keep the new `cPath::cPath()` check regardless of the change to `IsSolid`.

SafwatHalaby:
>Exactly - they fall, and therefore need no pathfinding at all. So not calling the pathfinder is an optimization.

A mob will not calculate a path if it's in the air anyways. So considering the void an air actually disables the pathfinder.

SafwatHalaby:
>I don't think so. For mobs outside the world, they are always falling and thus cannot path. If we just treat all bocks outside the world as air, I think that will cause the pathfinder to behave exactly as we want.

I think this is the best approach too. Treating all blocks outside as air is the right way to treat them. They **are** air. And we aren't making `IsSolid` doing stuff it isn't intended to do.

SphinxC0re:
I hope I'm not pushing this too hard but is there still work done on this?

worktycho:
I need to change the PR.

NiLSPACE:
You can now also use the `cChunkDef::IsValidHeight` function

worktycho:
Replacement PR is #2641

Issue #97, opened 2013-08-15
Add a ServerSetBlock function


tigerw:
After pulling my hair out thrice about the piston issue, I have come to the conclusion that it is best for the server to simply not send the arm set packet to the client. The client behaves as expected and everything is fine. The server can send the block to new clients entering the vicinity or logging in.

I have tried tracing the setblock / fastsetblock functions to see what they do but they go through a myriad of files and I gave up eventually. I guess one could just remove the line which sends a block update to the client?

On a more positive note, I fixed the sticky piston not giving a sticky arm bug!

PR #2542, opened 2015-10-10
Reorganised the redstone simulator

### LINKED ISSUE #1416: add block classes
Following on from the addition of more types of door and slab I propose we create a set of block classes for blocks that act the same way. For example one class would be all slabs. This would make code that wants to check for particular types simpler.

### LINKED ISSUE #1589: Redstone doesn't propagate through power rails
When powering a powered rail, the power should propagate through the rail through several blocks; currently only one rail block is powered.

### LINKED ISSUE #2234: Redstone wire update when powered from both sides
Redstone wire powering off does not properly re-power from opposite side.
redstone001

### LINKED ISSUE #2235: Placing redstone torch against powered block should create 1 tick pulse
In vanilla this turns on the right lamp for 1 tick.

redstone003

the lamp under the lever to show it is in the "on" state.

tigerw:
-> Many thanks to @worktycho for the idea, and @Haxi52 for the implementation plan!

* Uses classes and inheritance now
* Speed should be improved

Original PR: #2313
Refactor request reference: #2144 

Fixes #1416 (it mentioned the redstone simulator as its target)
*Nearly* fixes #1589, but not quite because of bugs
Fixes #2234
Fixes #2235

See Tycho? I *did* keep that promise of refactoring :)
And admittedly, this new structure is generally better.

Demonstration: https://youtu.be/5q-nmkelEDI


src/Blocks/BlockButton.h
```diff
@@ -128,6 +128,13 @@ class cBlockButtonHandler :
 		UNUSED(a_Meta);
 		return 0;
 	}
+
+
+	static const bool IsButtonOn(NIBBLETYPE a_BlockMeta)
```
worktycho: did you mean `static bool IsButtonOn(...) const` because currently th const is applied to the bool, not the this pointer.

src/Blocks/BlockLever.h
```diff
@@ -155,6 +155,13 @@ class cBlockLeverHandler :
 		UNUSED(a_Meta);
 		return 0;
 	}
+
+
+	static const bool IsLeverOn(NIBBLETYPE a_BlockMeta)
```
worktycho: Same const issue again

src/Simulator/IncrementalRedstoneSimulator/RedstoneSimulatorChunkData.h
```diff
@@ -0,0 +1,70 @@
...
+
+	void SetCachedPowerData(const Vector3i & a_Position, cRedstoneHandler::PoweringData a_PoweringData)
+	{
+		m_CachedPowerLevels[a_Position] = a_PoweringData;
+	}
+
+	std::tuple<int, int, bool> * GetMechanismDelayInfo(const Vector3i & a_Position)
+	{
+		auto Result = m_MechanismDelays.find(a_Position);
+		return (Result == m_MechanismDelays.end()) ? nullptr : &Result->second;
```
worktycho: So you're using a raw pointer as an optional return. We really need an optional type for situations like this.

src/Blocks/BlockButton.h
```diff
@@ -128,6 +128,13 @@ class cBlockButtonHandler :
 		UNUSED(a_Meta);
 		return 0;
 	}
+
+
+	static bool IsButtonOn(NIBBLETYPE a_BlockMeta)
```
worktycho: Why not const?

tigerw: It's considered unnecessary I think.

worktycho: Its useful to have.

tigerw: But I don't think it's used in other places as a rule.

src/Simulator/IncrementalRedstoneSimulator/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+	RedstoneToggleHandler.h
+	PistonHandler.h
+	SmallGateHandler.h
+	NoteBlockHandler.h
+	TNTHandler.h
+	TrappedChestHandler.h
+	TripwireHookHandler.h
+	PoweredRailHandler.h
+	PressurePlateHandler.h
+)
```
worktycho: You aren't actually compiling these files.

tigerw: They're still needed though, not least so that everything appears in Visual Studio's Solution Explorer.

worktycho: No your not compiling the cpp files either.

tigerw: I realised, but now I am! :D

src/Simulator/IncrementalRedstoneSimulator/RedstoneRepeaterHandler.h
```diff
@@ -0,0 +1,74 @@
...
+{
+	typedef cRedstoneHandler super;
+public:
+
+	cRedstoneRepeaterHandler(cWorld & a_World) :
+		super(a_World)
+	{
+	}
+
+	inline static const bool IsOn(BLOCKTYPE a_Block)
```
worktycho: Useless const

src/Simulator/IncrementalRedstoneSimulator/RedstoneTorchHandler.h
```diff
@@ -0,0 +1,99 @@
...
+class cRedstoneTorchHandler : public cRedstoneHandler
+{
+public:
+
+	cRedstoneTorchHandler(cWorld & a_World) :
+		cRedstoneHandler(a_World)
+	{
+	}
+
+	inline static const bool IsOn(BLOCKTYPE a_Block)
```
worktycho: useless const

Seadragon91:
Do you also add the slime block?

SphinxC0re:
That would be good. I already made a prototype on the logic:
https://github.com/SphinxC0re/SlimeBlockInteraction

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+
+
+
+
+void cIncrementalRedstoneSimulator::Simulate(float a_dt)
+{
+	for (auto & DelayInfo : m_Data.m_MechanismDelays)
+	{
+		std::get<1>(DelayInfo.second)++;
```
worktycho: Move this increment after the tie, so its clearer which field you're accessing.

worktycho: Also, if you change elapsed ticks into ticks left and count down, you can get rid of the Delayticks field.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.h
```diff
@@ -0,0 +1,19 @@
...
+#include "IncrementalRedstoneSimulator.h"
+
+
+
+
+
+class cRedstoneHandlerFactory
+{
+public:
+	static ComponentPtr GetComponent(cWorld & a_World, BLOCKTYPE a_BlockType, cIncrementalRedstoneSimulatorChunkData * a_Data);
```
worktycho: Why use a class for this rather than free functions?

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandler.h
```diff
@@ -0,0 +1,127 @@
...
+				{ 1, 0, 0 },
+				{ -1, 0, 0 },
+				{ 0, 0, 1 },
+				{ 0, 0, -1 },
+			}
+		};
+	}
+};
+
+typedef std::shared_ptr<cRedstoneHandler> ComponentPtr;
```
worktycho: Why shared_ptr? use a unique_ptr until you know that the code has to use a shared_ptr.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.cpp
```diff
@@ -0,0 +1,99 @@
...
+ComponentPtr cRedstoneHandlerFactory::GetComponent(cWorld & a_World, BLOCKTYPE a_BlockType, cIncrementalRedstoneSimulatorChunkData * a_Data)
+{
+	auto Result = a_Data->m_Handlers.find(a_BlockType);
+	if (Result == a_Data->m_Handlers.end())
+	{
+		auto Handler = cRedstoneHandlerFactory::CreateComponent(a_World, a_BlockType, a_Data);
+		a_Data->m_Handlers[a_BlockType] = Handler;
+		return Handler;
+	}
+	return Result->second;
```
worktycho: Do the handlers really need caching? its probably more expensive to retrieve them from the cache then it is to recreate them each time.

NiLSPACE:
I'm assuming that the cVector3iArray is a list of blocks that should be updated. I'm wondering if those coordinates are relative. If so then perhaps we can turn it into 4 bits instead of an int. Is that even possible? Depending on how much this is used it could save a little RAM. I doubt it would be used allot, but I'm just throwing it in here. :)

I read [here](http://www.cplusplus.com/forum/general/51911/) that you will always use 1 byte though, but you never know.

worktycho:
You could make a 2 byte 3d vector by storing the XZ coordinates in a single byte using either bit-fields, or a bit shifts on access.

worktycho:
Would save 10 bytes a vector compared to `Vector3<int>` and 2 bytes compared to `Vector3<Byte>`

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandler.h
```diff
@@ -0,0 +1,127 @@
...
+
+	cRedstoneHandler(cWorld & a_World) :
+		m_World(a_World)
+	{
+	}
+
+public:
+
+	// Disable the copy constructor and assignment operator
+	DISALLOW_COPY_AND_ASSIGN(cRedstoneHandler);
```
worktycho: Why not use deleted constructors? DISALLOW_COPY_AND_ASSIGN is an obsolete macro form before c++11.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandler.h
```diff
@@ -0,0 +1,127 @@
...
+
+	// Force a virtual destructor in descendants
+	virtual ~cRedstoneHandler() {}
+
+protected:
+
+	cWorld & m_World;
+
+	template <class Container>
+	Container StaticAppend(const Container & a_Lhs, const Container & a_Rhs)
```
worktycho: This function would be far better written with rvalue references:
```
template <class Container>
Container StaticAppend(Container && a_Lhs, Container && a_Rhs)
{
	a_Lhs.insert(a_Lhs.end(), a_Rhs.begin(), a_Rhs.end());
	return a_Lhs;
}
```
This way, if a_Lhs has the capacity, or is a node based container, you don't end up copying the elements of a_Rhs.

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		auto CurrentHandler = cRedstoneHandlerFactory::GetComponent(m_World, CurrentBlock, &m_Data).get();
+		if (CurrentHandler == nullptr)
+		{
+			continue;
+		}
+
+		cRedstoneHandler::PoweringData Power;
+		for (const auto & Location : CurrentHandler->GetValidSourcePositions(CurrentLocation, CurrentBlock, CurrentMeta))
```
worktycho: Is there a way to avoid the dynamic allocation in using a vector for GetValidSourcePostions's return. Perhaps a std::array of bools which can used against a static list?

tigerw: naaah too complicated for me.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.h
tigerw: Because you're all about more classes?

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.cpp
tigerw: How much more expensive is probably?

tigerw:
With regards to cVector3iArray and GetValidSourcePositions: maybe in the future compilers will optimise it. There should be a focus on ease-of-understanding really, not making optimisations.

With regards to slimeblocks; unfortunately no, still no implementation. Was mainly concerned with porting the simulator over to classes - sorry. Soon though, soon.

SphinxC0re:
@tigerw could you use my prototype for that? I spend too much time on it to just drop it.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.cpp
worktycho: No real idea, I just don't think the additional complexity of caching is worthwhile.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.h
worktycho: No, I don't want large classes. Free functions are not in classes, so do not result in large classes.

tigerw:
> @tigerw could you use my prototype for that? I spend too much time on it to just drop it.

Sure. I can't guarantee that it'll look exactly the same though, or that it'll be me who implements it.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandlerFactory.cpp
tigerw: I dunno either, and it's certainly simpler without, but look at this:

```cpp
#include <iostream>
#include <memory>
#include <unordered_map>
#include <chrono>
using namespace std;
using namespace chrono;

class lol
{
    public:
    long long d=7;
    long double e=5;
    bool f=8;
    int derp()
    {
        return (int)(d*e*(int)f);
        
    }
};

int main()
{
 unordered_map<int, unique_ptr<lol>> mpp;
 for (int i = 0; i != 10000; i++) mpp[i] = make_unique<lol>();
 
 
 
 auto t2 = high_resolution_clock::now();
 for (int i = 0; i != 10000; i++) volatile auto lel = make_unique<lol>()->derp();
 cout<<duration_cast<nanoseconds>(high_resolution_clock::now()-t2).count()<<endl;
 
 auto t1 = high_resolution_clock::now();
 for (int i = 0; i != 10000; i++) volatile auto lel = mpp[i]->derp();
 cout<<duration_cast<nanoseconds>(high_resolution_clock::now()-t1).count()<<endl;
 
 return 0;
}
```

```
375301
173165
 ```

File: src/BlockID.h [hunk 15]
```diff
@@ -486,13 +486,13 @@ enum
...
-	E_BLOCK_BUTTON_ZM      = 4,
-	E_BLOCK_BUTTON_YP      = 5,
-	E_BLOCK_BUTTON_PRESSED = 8,
+	E_META_BUTTON_YM      = 0,
+	E_META_BUTTON_XP      = 1,
+	E_META_BUTTON_XM      = 2,
+	E_META_BUTTON_ZP      = 3,
+	E_META_BUTTON_ZM      = 4,
+	E_META_BUTTON_YP      = 5,
+	E_META_BUTTON_PRESSED = 8,
```
worktycho: This is a breaking change for lua. So you either need to revert the change or add manual bindings for the old names.

SphinxC0re:
Travis is failing:
```
Globals.E_META_BUTTON_PRESSED (constant)
Globals.E_META_BUTTON_XM (constant)
Globals.E_META_BUTTON_XP (constant)
Globals.E_META_BUTTON_YM (constant)
Globals.E_META_BUTTON_YP (constant)
Globals.E_META_BUTTON_ZM (constant)
Globals.E_META_BUTTON_ZP (constant)
```

src/BlockID.h
tigerw: Aah, I say forget backwards compatibility :P

tigerw:
Meeerge?

SphinxC0re:
I don't see why not. @worktycho ?

SphinxC0re:
@madmaxoft ?

Seadragon91:
I could do the changes in a other pull requests, after this has been merged, there are a few other things too.

src/BlockEntities/CommandBlockEntity.cpp
```diff
@@ -114,11 +113,10 @@ void cCommandBlockEntity::Activate(void)
 
 void cCommandBlockEntity::SetRedstonePower(bool a_IsPowered)
 {
-	if (a_IsPowered && !m_IsPowered)
+	if (a_IsPowered)
 	{
 		Activate();
 	}
-	m_IsPowered = a_IsPowered;
```
madmaxoft: This function has changed its contract. It used to mean "set the redstone power to either on or off, and if it is going from off to on, call Activate". Calling it to set "on" when the state was already "on" didn't call `Activate`. Now it calls `Activate` for each setting to "on". Is that intentional? If so, update the docs for the function everywhere.

src/Blocks/BlockLever.h
```diff
@@ -155,6 +155,13 @@ class cBlockLeverHandler :
 		UNUSED(a_Meta);
 		return 0;
 	}
+
+
+	static bool IsLeverOn(NIBBLETYPE a_BlockMeta)
+	{
+		// Extract the ON bit from metadata and return if true if it is set:
```
madmaxoft: This comment should be the doxy-comment for the function

src/ChunkMap.cpp
```diff
@@ -3154,28 +3078,14 @@ void cChunkMap::cChunkLayer::UnloadUnusedChunks(void)
...
-	this->FastSetBlocks(FastSetBlockQueueCopy);
-	if (!FastSetBlockQueueCopy.empty())
+	cCSLock Lock(m_CSLayers);
+	cChunkPtr Chunk = GetChunk(ChunkX, ChunkZ);
+	if ((Chunk != nullptr) && Chunk->IsValid())
 	{
-		// Some blocks failed, store them for next tick:
-		cCSLock Lock(m_CSFastSetBlock);
-		m_FastSetBlockQueue.splice(m_FastSetBlockQueue.end(), FastSetBlockQueueCopy);
+		Chunk->FastSetBlock(X, Y, Z, a_BlockType, a_BlockMeta);
```
madmaxoft: Why did you change the asynchronous set to a synchronous one? There must have been a reason for the asynchronicity, and git blame should have helped you pinpoint it.

src/ClientHandle.cpp
```diff
@@ -926,6 +926,12 @@ void cClientHandle::HandleBeaconSelection(int a_PrimaryEffect, int a_SecondaryEf
 
 void cClientHandle::HandleCommandBlockBlockChange(int a_BlockX, int a_BlockY, int a_BlockZ, const AString & a_NewCommand)
 {
+	if (a_NewCommand.empty())
+	{
+		Kick("Command block string unexpectedly empty - hacked client?");
+		return;
+	}
+
```
madmaxoft: Why? What does the client send when you update the command to an empty string in the UI?

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -0,0 +1,184 @@
...
+		case E_BLOCK_STONE_PRESSURE_PLATE:
+		case E_BLOCK_WOODEN_PRESSURE_PLATE:
+		case E_BLOCK_HEAVY_WEIGHTED_PRESSURE_PLATE:
+		case E_BLOCK_LIGHT_WEIGHTED_PRESSURE_PLATE: return cpp14::make_unique<cPressurePlateHandler>(a_World);
+		case E_BLOCK_TRIPWIRE_HOOK: return cpp14::make_unique<cTripwireHookHandler>(a_World);
+		case E_BLOCK_ACTIVE_COMPARATOR:
+		case E_BLOCK_INACTIVE_COMPARATOR: return cpp14::make_unique<cRedstoneComparatorHandler>(a_World);
+		case E_BLOCK_DROPPER:
+		case E_BLOCK_DISPENSER: return cpp14::make_unique<cDropSpenserHandler>(a_World);
+		case E_BLOCK_TRAPPED_CHEST: return cpp14::make_unique<cTrappedChestHandler>(a_World);
```
madmaxoft: Style! This switch is completely unreadable. Either sort it alphabetically and return in each case, or make each case a curly-brace block and add newlines to separate the multi-case blocks.

src/World.cpp
```diff
@@ -3502,6 +3500,7 @@ cRedstoneSimulator * cWorld::InitializeRedstoneSimulator(cIniFile & a_IniFile)
 
 	if (NoCaseCompare(SimulatorName, "Incremental") == 0)
 	{
+		// res = new cIncrementalRedstoneSimulator(*this);
```
madmaxoft: What's this? A commented-out line followed by the same line? Some debugging leftovers?

madmaxoft:
I'm not reviewing the redstone code, not mycuppatea, I expect you to have tested it thoroughly. But I could use an *overview* about how it works. And I'm not letting this in until you explain the async-to-sync change in `FastSetBlock`.

Seadragon91:
Okay first test failed. A simple on/off not working correctly. Switching it on works, but switching it off, doesn't change the redstone wire to off.

src/ChunkMap.cpp
tigerw: I was trying to get piston animations working (remember that fiasco? #97), but it's clear that it isn't going to look nice without proper usage of the PistonMoving tile entity - in 1.8 I am now experiencing client-side block duplication, and retraction was always hit-and-miss.

src/ClientHandle.cpp
tigerw: Nothing; the button is disabled.

tigerw:
Oh yeah, levers/buttons are buggy.

tigerw:
Test & merge?

src/ClientHandle.cpp
madmaxoft: I still think an empty command should be allowed. There's no reason not to, is there? Empty commands still need to be supported because that's the default state of a newly created command block. If Mojang ever decides to support empty commands, this would get in our way.

madmaxoft:
You still haven't properly addressed my concerns about changing the synchronicity of `FastSetBlock`. Did you investigate into why it was asynchronous, and did you find any countermeasures for the deadlocks it had avoided?

src/ChunkMap.cpp
tigerw: @madmaxoft.

Seadragon91:
If something basic redstone mechanic, doesn't work, would you fix them in this pull request, too. Or should I create a own issue for that?

tigerw:
If its something that's currently implemented, I'll try to fix it if its straightforward :P

tigerw:
Still Alive with the current implementation: https://youtu.be/5q-nmkelEDI

Seadragon91:
Played around with redstone machines. Had then downloaded and tested one that crosses a few chunks. The server [deadlocked](https://gist.githubusercontent.com/Seadragon91/1ca19ba287000d428f1d/raw/58216ac6a043450e93656e2782ee7b08e4f8a83a/gistfile1.txt).

worktycho:
I think what happened was not an actual deadlock. The server merely lagged so much in ticking that it assumed it was deadlocked. If the machine can run without player intervention, I'd like to profile the server running it.

Seadragon91:
You can use the world provided in this [youtube video](https://www.youtube.com/watch?v=Ns4f8wqo8eQ). I only had to teleport to the coordinates `-600 150 -500`, then the server crash after around 20-30 seconds.

LO1ZB:
I think its the AI thats deathlocking this world, not the redstone simulator.

src/Simulator/IncrementalRedstoneSimulator/TNTHandler.h
```diff
@@ -0,0 +1,58 @@
...
+		UNUSED(a_Meta);
+		return 0;
+	}
+
+	virtual cVector3iArray Update(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, PoweringData a_PoweringData) override
+	{
+		LOGD("Evaluating explodinator the trinitrotoluene (%d %d %d)", a_Position.x, a_Position.y, a_Position.z);
+		if (a_PoweringData.PowerLevel != 0)
+		{
+			m_World.BroadcastSoundEffect("game.tnt.primed", (double)a_Position.x, (double)a_Position.y, (double)a_Position.z, 0.5f, 0.6f);
```
worktycho: Old style casts

worktycho:
The server is currently using almost exactly 100% of a cpu core, on that world, and does not crash if I turn of the deadlock detector. This looks like the redstone simulator is dong too much work on the tick thread.

tigerw:
Debug or release?

At the expense of performance, this new model achieves relative simplicity by generally simulating a block multiple times unnecessarily, treating solid blocks as redstone devices to simplify linked powering, and not implementing (as of yet) chunked data storage.

However, I did a profiler run on VS as you suggested on the forum, and found that CPU usage on the noteblocks world was tiny on Release mode. Try release?

worktycho:
I've just run a release profile against the deadlocking world. It generated plenty of data.

tigerw:
What did it say? I probably can't do anything about whatever it highlights though unless its really obvious.

worktycho:
A lot of time in simulate. I'm looking at the annotations to see why. The main issue seems to be you're using so much memory, you're causing a lot of cache issues.

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
worktycho: I suggest you take closer look at this loop. Loading data from the stack in this loop is where the simulator spends most of its time.

src/Simulator/IncrementalRedstoneSimulator/SolidBlockHandler.h
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	virtual unsigned char GetPowerDeliveredToPosition(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, const Vector3i & a_QueryPosition, BLOCKTYPE a_QueryBlockType) override
+	{
+		// TODO: wire isn't linked powered only if the source was a wire, not just because it is a wire
+		return (
+			!cIncrementalRedstoneSimulator::IsRedstone(a_QueryBlockType) ||
+			(
+				(a_QueryBlockType == E_BLOCK_REDSTONE_WIRE) &&
+				(static_cast<cIncrementalRedstoneSimulator *>(m_World.GetRedstoneSimulator())->GetChunkData()->GetCachedPowerData(a_Position).PoweringBlock == E_BLOCK_REDSTONE_WIRE)
```
worktycho: See if you can get rid of some of these calls by passing in the cached power data. At some point in the chain, this is hiting cold data, and spending ages waiting on retrieving it from RAM.

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -0,0 +1,170 @@
...
+			}
+			m_World.GetBlockTypeMeta(Location.x, Location.y, Location.z, PotentialBlock, PotentialMeta);
+
+			auto PotentialSourceHandler = std::move(cIncrementalRedstoneSimulator::CreateComponent(m_World, PotentialBlock, &m_Data));
+			if (PotentialSourceHandler == nullptr)
+			{
+				continue;
+			}
+
+			decltype(Power) PotentialPower(PotentialBlock, PotentialSourceHandler->GetPowerDeliveredToPosition(Location, PotentialBlock, PotentialMeta, CurrentLocation, CurrentBlock));
```
worktycho: It might be worth adding a manual devirtualisation here. Yes it would be ugly but it might give a significant performace boost.

Something along the lines of:
```
auto Delivered = IsRedstone(PotentialBlock) 
    ? PotentialSourceHandler->GetPowerDeliveredToPosition(Location, PotentialBlock, PotentialMeta, CurrentLocation, CurrentBlock) 
    : PotentialSourceHandler->SolidBlockHandler::GetPowerDeliveredToPosition(Location, PotentialBlock, PotentialMeta, CurrentLocation, CurrentBlock) ;
decltype(Power) PotentialPower(PotentialBlock, Delivered);
```

tigerw:
I just commented out the code in cSolidBlock::Update and guess what, it actually works faster than the normal Java client. The calculator doesn't actually work (hah, fat chance) but it runs and doesn't deadlock.

Looks like I may need to rework the idea of using solid blocks as link-powering. Whether its just a subtle infinite loop bug or an actual performance problem I don't know.

worktycho:
Its a performance problem. The server respond to commands, it just ticks slowly. You can tell that by running the world with the deadlock detector off.

SphinxC0re:
Any updates?

tigerw:
@tigerw vacillates, citing insurmountable technical challenges

I'm not really sure about what to do with the slowness.

SphinxC0re:
@worktycho any suggestions? I'm out of expertise here

worktycho:
I've pointed out two places in the code where you could improve performance, particularly getting rid of the cWorld accesses in the solid block handler.

I would generally try and refactor the code to avoid external references in the handler code, try to make the functions pure where possible.

SphinxC0re:
@tigerw are you still working on this one?

tigerw:
Yes, but at an average pace of 1 character per day.

SphinxC0re:
Okay, seems like you abandoned this PR for a whole month now, @tigerw. What about the performance suggestions made by @worktycho?

tigerw:
Evidently, I don't have the time to keep up with my changes. It does have issues, but what about merge first? No point leaving code stagnant after all.

worktycho:
I think we need to fix the major performance regression first.

tigerw:
I somehow don't think the current code can handle the calculator world either.

src/Simulator/IncrementalRedstoneSimulator/DoorHandler.h
```diff
@@ -0,0 +1,59 @@
...
+
+	virtual cVector3iArray Update(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, PoweringData a_PoweringData) override
+	{
+		LOGD("Evaluating dori the door (%d %d %d)", a_Position.x, a_Position.y, a_Position.z);
+
+		if (a_PoweringData != static_cast<cIncrementalRedstoneSimulator *>(m_World.GetRedstoneSimulator())->GetChunkData()->ExchangeUpdateOncePowerData(a_Position, a_PoweringData))
+		{
+			cChunkInterface ChunkInterface(m_World.GetChunkMap());
+			cBlockDoorHandler::SetOpen(ChunkInterface, a_Position.x, a_Position.y, a_Position.z, (a_PoweringData.PowerLevel != 0));
+			m_World.BroadcastSoundParticleEffect(1003, a_Position.x, a_Position.y, a_Position.z, 0);
```
worktycho: This functions signature has changed.

src/Simulator/IncrementalRedstoneSimulator/PistonHandler.h
```diff
@@ -0,0 +1,68 @@
...
+		return 0;
+	}
+
+	virtual cVector3iArray Update(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, PoweringData a_PoweringData) override
+	{
+		LOGD("Evaluating pisty the piston (%d %d %d)", a_Position.x, a_Position.y, a_Position.z);
+
+		if (a_PoweringData.PowerLevel > 0)
+		{
+			cBlockPistonHandler::ExtendPiston(a_Position.x, a_Position.y, a_Position.z, &m_World);
```
worktycho: This functions signature has changed.

src/Simulator/IncrementalRedstoneSimulator/PistonHandler.h
```diff
@@ -0,0 +1,68 @@
...
+		}
+		else
+		{
+			cBlockPistonHandler::RetractPiston(a_Position.x, a_Position.y, a_Position.z, &m_World);
```
worktycho: This function's signiture has changed.

worktycho:
You need to get this to compile against the current maaster, before we can even think about merging.

tigerw:
Obviously, but it seems that I can't git.

tigerw:
Can you think about merging now?

tigerw:
I guess! :D

PR #2544, opened 2015-10-12
Removed unused typedef

worktycho:


SphinxC0re:
In my opinion, we should never use c++ lists ever. The std::vector is very powerful and won't have any cachemisses

madmaxoft:
There are some situations where lists are just better than vectors, but they are truly rare. I agree that we should use vectors as much as possible, but just don't close your mind.

Merge if this builds.

PR #2551, opened 2015-10-19
Update JsonCPP

Seadragon91:
This updates JsonCpp to latest version. The version we currently using has a problem that crashes cmake 3.4 under windows. Updating JsonCpp  fixes the problem.
I had to fix a few errors from JsonCpp. Please look at them and if they are okay then it can be merged.

Edit: Not worth struggling with this anymore.
If someone has an error from CMake under windows. You can install CMake 3.4.0-rc2, that contains the fix for the JsonCpp problem.

### DIFF
--- a/CIbuild.sh
+++ b/CIbuild.sh
@@ -14,7 +14,7 @@ fi
 cmake . -DBUILD_TOOLS=1 -DSELF_TEST=1;
 
 echo "Building..."
-make -j 2;
+VERBOSE=1 make -j 2;
 make -j 2 test ARGS="-V";
 
 echo "Testing..."


src/BlockEntities/EnderChestEntity.cpp
```diff
@@ -75,7 +75,7 @@ void cEnderChestEntity::OpenNewWindow()
 void cEnderChestEntity::LoadFromJson(const Json::Value & a_Value, cItemGrid & a_Grid)
 {
 	int SlotIdx = 0;
-	for (Json::Value::iterator itr = a_Value.begin(); itr != a_Value.end(); ++itr)
+	for (Json::Value::const_iterator itr = a_Value.begin(); itr != a_Value.end(); ++itr)
```
worktycho: If you're going to touch the loop, try to convert it to a c++11 range loop where possible.

src/WorldStorage/StatSerializer.cpp
```diff
@@ -102,7 +102,7 @@ bool cStatSerializer::LoadStatFromJSON(const Json::Value & a_In)
 {
 	m_Manager->Reset();
 
-	for (Json::ValueIterator it = a_In.begin() ; it != a_In.end() ; ++it)
+	for (Json::Value::const_iterator it = a_In.begin() ; it != a_In.end() ; ++it)
 	{
 		AString StatName = it.key().asString();
```
Seadragon91: I can not change the for loop here, because I need access to method key() from iterator.

Seadragon91:
@worktycho  Updated for loops. Please check the changes.

worktycho:
Something has gone weird with include paths. I'll try and remember to take a closer look this afternoon.

Seadragon91:
Really strange, it builds find under linux...

Seadragon91:
Did an force pull requests, thought that something could be wrong on travis. Still not working.

Seadragon91:
I will test something about that now.

Seadragon91:
@worktycho Okay I let this now to you. Really strange now...

worktycho:
libstdc++ might of done something weird. htonl is giving you warnings.

Seadragon91:
Okay. That was my last try in this pull request. If anyone wants to try it. Good luck.

worktycho:
I've found the problem: https://github.com/open-source-parsers/jsoncpp/commit/d1a2b94d5b1adceb7d6a9f8c0f40c95dd7e64980 is messing up our builds by adding ccache to everything. I think we need to use a cmake package to spit jsoncpp off from the rest of the build to fix it.

SphinxC0re:
@Seadragon91 are you still working on this?

Seadragon91:
I haven't work with cmake packages, therefore can't fix that problem with ccache. The best solution in the meantime is to use the latest version of cmake 3.4 that fixes the crash under windows.

PR #2567, opened 2015-10-27
Added cave spider to MobSpawner

Seadragon91:
Did no one missed the small, dangerous, poisonous spiders? :wink:
Had missed them in my [mazes](https://github.com/Seadragon91/MazeGenerator) :smile:

### DIFF
--- a/src/MobSpawner.cpp
+++ b/src/MobSpawner.cpp
@@ -236,6 +236,17 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
 				}
 				return CanSpawn && HasFloor && (SkyLight <= 7) && (BlockLight <= 7);
 			}
+
+			case mtCaveSpider:
+			{
+				return (
+					(TargetBlock == E_BLOCK_AIR) &&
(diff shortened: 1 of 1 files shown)


Seadragon91:
Used the code from the spider, just changed the x and z from 2 to 1.

src/MobSpawner.cpp
```diff
@@ -236,6 +236,32 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
...
 			}
+
+			case mtCaveSpider:
+			{
+				bool CanSpawn = true;
+				bool HasFloor = false;
+				for (int x = 0; x < 1; ++x)
+				{
+					for (int z = 0; z < 1; ++z)
+					{
```
madmaxoft: These two loops are not needed at all, they basically just say 'int x = 0;` and `int z = 0`, since they are never repeated.

madmaxoft:
It would have been better if you copied the creeper / skeleton / zombie code and removed the `BlockAbove` checks from it.

Seadragon91:
Corrected it. Yes the code from zombie is better.

Seadragon91:
Rebased on master.

PR #2576, opened 2015-10-28
Added myself to list, ran it through `sort -f`

Schwertspize:


bearbin:
Just making sure that you are aware:

by adding yourself to the list you are aware that you are licensing all present and future contributions to Cuberite under the Apache License 2.0 so that anyone may use them freely.

PR #2577, opened 2015-10-28
Chicken egg throw patch

rhamilton1415:
If an egg hits an ender crystal the crystal will be destroyed


Schwertspize:
first, circleci has not passed, that's the style check. see CONTRIBUTING.md
secondly, those other CI did not pass, have a look at the errors and fix them.
thirdly, next time test locally.

PR #2574, opened 2015-10-28
Pressure plate fix

Schwertspize:
Dont know the first commit, but the second one is right, tested.
it fixes the sound for stone plate and sets a delay on leaving the pressure plate (all ones, including weighted)

PR #2578, opened 2015-10-29
Revert "Pressure plate fix"

madmaxoft:
Reverts cuberite/cuberite#2574

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator.cpp
@@ -1160,33 +1160,17 @@ void cIncrementalRedstoneSimulator::HandlePressurePlate(int a_RelBlockX, int a_R
 		{
 			// MCS feature - stone pressure plates can only be triggered by players :D
 			cPlayer * a_Player = this->m_World.FindClosestPlayer(Vector3f(BlockX + 0.5f, static_cast<float>(a_RelBlockY), BlockZ + 0.5f), 0.5f, false);
-			static Int64 m_PlayerLastOnPlate = 0;
-			static bool m_WasPlayerOnPlate = false;
 
 			if (a_Player != nullptr)
 			{
-				if (m_WasPlayerOnPlate == false)
-				{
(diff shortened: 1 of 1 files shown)


bearbin:
Did you mean to merge this?

Sorry for merging the first broken PR.

SphinxC0re:
Oh! Got it! With *reverse-merging* you meat just merging the reversed changes.

Schwertspize:
why? . 1 why not just press the revert button, secondly what is wrong with my pr

tigerw:
No idea about the revert button, it probably was used, but the reversion reason was that the use of `static` variables meant that for more than one pressure plate, since the handler function is reused, the states would have been overwritten by each plate. I believe an appropriate implementation would be to store the pressure plate delay datum in the chunk's redstone simulator data object, so it will be distinct for each plate.

Schwertspize:
no. not for every plate. just if you move to another plate while the first one is activated, those keep activated. but I have talked with @bearbin about what I missed and he said I should make It static.

madmaxoft:
You're thinking single-player. Think 100 players each stepping on various pressure plates. The variable will get overwritten.

Schwertspize:
Oh... but how to solve it? I mean we can't make it non static because that would break the whole code...

tigerw:
I believe an appropriate implementation would be to store the pressure plate delay datum in the chunk's redstone simulator data object, so it will be distinct for each plate.

However, the redstone sim is currently undergoing a remodel. This issue still persists in the new layout, so do you think once that PR is merged (#2542) you could revisit the bug?

Schwertspize:
yeah but where to save it? as a variable in cChunk and a getter and setter? 
I'm new to this project..

tigerw:
So for the current model, there's a structure in cIncrementalRedstoneSimulator.h called cIncrementalRedstoneSimulatorChunkData; you could either add a new map of coordinate to pressure plate delay data, or adapt the existing repeater delay storage to pressure plates.

However, it is recommended to wait until #2542 is merged, since everything changes.

Schwertspize:
yeah.... I think I will wait (but please, merge soon) and save the delay in  a (static)  3 dimensional. map.

bearbin:
Is that not a memory leak - deleted pressure plates are never removed from the map?

Schwertspize:
yeah. thanks for pointing that out, we should create a method to delete them if they exist and run it I'm block destroy.

PR #2585, opened 2015-11-01
Added gdb to travis

Seadragon91:
Added gdb and script commands.
Also code that crashes the server after initialized, will be removed after success.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
 language: cpp
-sudo: false
+sudo: required
 
 compiler:
 - clang
 - gcc
+
+before_script:
+- ulimit -c unlimited -S    #enable core dumps
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


worktycho:
Why does this need the old arcitecture? the new one is faster, and tavis have indecated they will eventually close the old one.

Seadragon91:
According to this https://github.com/travis-ci/travis-ci/issues/3754#issuecomment-95629153

The problem is we cannot change the core-pattern to core and in containers the core-pattern is changed.

Then we need to wait until travis fixes the container.

worktycho:
Then do we really need to have core dumps on travis? It rarely dumps core in a way that cannot be dumped locally, and in my opinon we lose more from giving up container builds than we gain for getting core dumps.

Seadragon91:
I think so, too. Just close this and if there is a fix that works in containers, then we can reopen this.

PR #2591, opened 2015-11-02
Fix deadlock, resolves #2521

### LINKED ISSUE #2521: [Generator].PregenerateDistance set to zero deadlocks server on startup
When the world.ini file specifies `[Generator].PregenerateDistance = 0`, the server deadlocks on startup, while waiting for the spawn loader to finish loading chunks.

Ref.: #1656

bearbin:


bearbin:
I just tested and I'm pretty sure that this fixes the deadlock.

PR #2593, opened 2015-11-02
Destroy an ender crystal when hit by an egg.

bearbin:
Fixed version of #2577.

### DIFF
--- a/src/Entities/Entity.h
+++ b/src/Entities/Entity.h
@@ -84,16 +84,16 @@ public:
 		etFloater,
 		etItemFrame,
 		etPainting,
-		
+
 		// Common variations
 		etMob = etMonster,  // DEPRECATED, use etMonster instead!
 	} ;
-	
+
 	// tolua_end
-	
+
 	enum eEntityStatus
 	{
-		// TODO: Investiagate 0, 1, and 5 as Wiki.vg is not certain
+		// TODO: Investigate 0, 1, and 5 as Wiki.vg is not certain
 
 		// Entity becomes coloured red
 		esGenericHurt            = 2,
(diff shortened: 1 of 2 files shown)


SphinxC0re:
@cuberite/core-team any objections?

tigerw:
No objections, but *how* did the branch become out of date? Nobody committed anything relevant to master.

bearbin:
I'm not exactly sure, I think it might be that any new change to master could break the build, and you have to re-merge every time there is a new commit?

bearbin:
Just to make sure that the PR still works and doesn't break the build.

src/Entities/ThrownEggEntity.cpp
```diff
@@ -32,11 +32,15 @@ void cThrownEggEntity::OnHitSolidBlock(const Vector3d & a_HitPos, eBlockFace a_H
 void cThrownEggEntity::OnHitEntity(cEntity & a_EntityHit, const Vector3d & a_HitPos)
 {
 	int TotalDamage = 0;
-	// TODO: If entity is Ender Crystal, destroy it
-	
+	// If entity is an Ender Crystal, destroy it
+	if (a_EntityHit.IsEnderCrystal())
+	{
+		a_EntityHit.Destroy();
```
tigerw: Oh, perhaps a `return` here?

NiLSPACE: That means we wouldn't spawn any chickens. True, they'd explode right afterwards, but still.

tigerw: Alright.

src/Entities/ThrownEggEntity.cpp
```diff
@@ -32,11 +32,15 @@ void cThrownEggEntity::OnHitSolidBlock(const Vector3d & a_HitPos, eBlockFace a_H
...
+	}
+
 	TrySpawnChicken(a_HitPos);
 	a_EntityHit.TakeDamage(dtRangedAttack, this, TotalDamage, 1);
```
tigerw: I wonder if this should kill the crystal anyway. It should if the crystal only has one health. Hm.

madmaxoft:
I think the check and destruction should be inserted *after* the `TakeDamage` call. Or even better, shouldn't the `TakeDamage` call for the ender crystal destroy the crystal after any damage?

bearbin:
It seems like the takedamage call should destroy the ender crystal anyway. Has anybody actually tested this?

If the crystal is destroyed with the old code, all that needs doing is the comment removing.

madmaxoft:
Now waitasecond, do eggs really cause any damage in Vanilla? I thought they didn't.

madmaxoft:
No, eggs do damage only to the enderdragon: http://minecraft.gamepedia.com/Egg#Combat

bearbin:
OK, I will fix the PR :P

bearbin:
The PR should be fixed now. Ender dragons and eggs are damaged when hit, and other mobs are simply knocked back.

src/Entities/Entity.h
```diff
@@ -630,7 +630,3 @@ class cEntity
 } ;  // tolua_export
 
 typedef std::list<cEntity *> cEntityList;
-
-
-
-
```
madmaxoft: Keep the end lines

src/Entities/ThrownEggEntity.cpp
```diff
@@ -32,11 +32,16 @@ void cThrownEggEntity::OnHitSolidBlock(const Vector3d & a_HitPos, eBlockFace a_H
 void cThrownEggEntity::OnHitEntity(cEntity & a_EntityHit, const Vector3d & a_HitPos)
 {
 	int TotalDamage = 0;
-	// TODO: If entity is Ender Crystal, destroy it
-	
+	// If entity is an Ender Dragon or Ender Crystal, it is damaged.
+	if ((a_EntityHit.IsMob() && (static_cast<cMonster &>(a_EntityHit).GetMobType() == mtEnderDragon)) ||
+		a_EntityHit.IsEnderCrystal())
```
madmaxoft: Formatting:
```cpp
if (
  (isA() && (a == b)) ||  // Is an ender dragon
  isB()
)
{
```

bearbin:
I can't fix the issues now, I need to fix my editor for the lines at the end, but I fixed the other formatting issue.

Seadragon91:
I would suggest deactivating the package whitespace in atom, that makes it unreadable..

bearbin:
Whitespace is fixed now.

SphinxC0re:
Looks good. Merge if compiles

PR #2594, opened 2015-11-02
Fix shell location.

bearbin:


Schwertspize:
thanks, I had to fix it locally for me to work.

PR #2595, opened 2015-11-02
Use an unreserved qualifier for thread local storage

### LINKED ISSUE #1906: Won't run nor compile on Mac OS X or Debian
Hello,

I have a VPS running Debian and a Mac, and I can't see to make MCServer run on either of those. 

Here are the logs for the various attempts, named appropriately.

https://www.dropbox.com/sh/je8t2nnfdwqhccd/AADvW3fwj7U5EyOh1QdGQLf0a?dl=0

Note about attempting to execute the downloaded version of MCServer on Debian: I attempted to update libstdc++, but no source online seemed to do anything. 😕

If this could be fixed soon (or I'm doing something wrong), that would be wonderful, as I want to get started on my next server soon.

Thanks,
Nathan

cengizIO:
Fixes #1906 

This was never fixed in master branch and it was FTC 

First I tried using `thread_local` directly but apparently it's not supported by Apple clang out of the box. 

```
Apple LLVM version 7.0.0 (clang-700.1.76)
Target: x86_64-apple-darwin15.0.0
Thread model: posix
```

Fixed compiler error:
```
/Users/canc/Code/cpp/cuberite/src/FastRandom.cpp:14:10: error: keyword is hidden by macro definition [-Werror,-Wkeyword-macro]
        #define thread_local static __thread
                ^
```

Although it's not supported, it's still a valid reserved keyword in clang. That was causing this.

### DIFF
--- a/src/FastRandom.cpp
+++ b/src/FastRandom.cpp
@@ -8,17 +8,19 @@
 
 #include <random>
 
-#ifdef _WIN32
-	#define thread_local static __declspec(thread)
-#elif defined __APPLE__
-	#define thread_local static __thread
+#if defined (__GNUC__)
+	#define ATTRIBUTE_TLS static __thread
+#elif defined (_MSC_VER)
+	#define ATTRIBUTE_TLS static __declspec(thread)
+#else
+	#error "Unknown thread local storage qualifier"
 #endif
 
 static unsigned int GetRandomSeed()
 {
-	thread_local bool SeedCounterInitialized = 0;
-	thread_local unsigned int SeedCounter = 0;
-	
+	ATTRIBUTE_TLS bool SeedCounterInitialized = 0;
+	ATTRIBUTE_TLS unsigned int SeedCounter = 0;
+
 	if (!SeedCounterInitialized)
 	{
 		std::random_device rd;
@@ -47,8 +49,8 @@ public:
 		TestInts();
 		TestFloats();
 	}
-	
-	
+
+
 	void TestInts(void)
 	{
 		printf("Testing ints...\n");


cengizIO:
This also fixes #1815

src/FastRandom.cpp
```diff
@@ -8,17 +8,19 @@
 
 #include <random>
 
-#ifdef _WIN32
-	#define thread_local static __declspec(thread)
-#elif defined __APPLE__
-	#define thread_local static __thread
+#if defined (__GNUC__)
+	#define ATTRIBUTE_TLS static __thread
```
tigerw: I think all supported Linux compilers have plain `thread_local` available.

cengizIO: @tigerw Yes, it's available since GCC 4.8. The point is Apple's clang does not support it. And that clang is the default in OSX when you get Xcode tools.

madmaxoft:
Could you rebase onto current master, rather than merge it?

SphinxC0re:
Also, squash all commits into a single one

cengizIO:
@madmaxoft @SphinxC0re I just did. 

Thank you.

madmaxoft:
Thanks :)

src/FastRandom.cpp
tigerw: As in, you should add a separate definition for normal Linux systems for the standard attribute, in addition to the current Mac + Windows definitions, but nevermind :P

cengizIO: @tigerw You're right. 

I just chose the alternative with compiler identifiers.

PR #2601, opened 2015-11-04
Fix being set on fire while directly being under lava

Gargaj:
Right now you will be on fire if there is a block of lava right above you, even if it doesn't pour down.

Simulation:
- Player is on Y=0, Lava is on Y=2.
- Player is 1.8 tall
- MinY = floor(PosY) = 0
- MaxY = ceil(PosY + 1.8) = ceil(1.8) = 2
- The loop runs MinY->MaxY inclusive -> tests 0 to 2 -> tests lava block that the player is not in actual contact with.

Either my solution, or change the loop to be exclusive to MaxY.


bearbin:
This PR is wrong, the floor has a 0 included, so the value of MinY will always be 0.

Gargaj:
The change involves MaxY, not MinY

bearbin:
Sorry, I meant MaxY. Reread the line you changed.

bearbin:
Derp, misread the code.

PR #2602, opened 2015-11-04
Chickens shouldn't take fall damage

Gargaj:
Also needs an exception at https://github.com/cuberite/cuberite/blob/master/src/Entities/Entity.cpp#L931 for the falling speed to be exempted from gravity, but I don't know the actual numbers.


NiLSPACE:
I could be completely wrong, but doesn't the `HandleFalling` handle the actual falling of an entity, not the damage?

Gargaj:
Not as far as I can tell? https://github.com/cuberite/cuberite/blob/master/src/Mobs/Monster.cpp#L565

madmaxoft:
Looks good to me. Could be extended to ocelots, too.

SphinxC0re:
That would be another PR just for structure. Please rebase onto current master and squash the commits so we can merge

Issue #1404, opened 2014-09-10
Add slime block effects


archshift:
http://minecraft.gamepedia.com/Slime_Block

SphinxC0re:
I created a little prototype https://github.com/SphinxC0re/SlimeBlockInteraction

SphinxC0re:
@archshift could you set up a checkable list of features that work and ones that don't

mathwhiz1212:
A player on my server is complaining that slime blocks don't prevent fall damage.

tigerw:
No problem - this is where we direct complaining ~~prisoners~~ players:

![Solitary](http://vignette2.wikia.nocookie.net/prison-architect/images/8/85/Solitary-PA.png/revision/latest/scale-to-width-down/220?cb=20151230102024)

PR #2609, opened 2015-11-05
Slimeblock implementation

### LINKED ISSUE #2530: Slime block never drop itself when destroy

bibo38:
This PR implements the slimeblocks and the piston mechanics with them.
This fixes #2530, but not fully #1404, because the bouncing/item mechanics are not implemented.

Please be aware, that the client does the piston calculation client side, so that you should relog to check, if it really worked.

### DIFF
--- a/src/Blocks/BlockPiston.cpp
+++ b/src/Blocks/BlockPiston.cpp
@@ -107,6 +107,12 @@ bool cBlockPistonHandler::CanPushBlock(
 		return true;
 	}
 
+	if(!a_RequirePushable && cBlockInfo::IsPistonBreakable(currBlock))
+	{
+		// Block should not be broken, when it's not in the pushing direction
+		return true;
+	}
+
 	if (!CanPush(currBlock, currMeta))
 	{
 		return !a_RequirePushable;
@@ -273,7 +279,7 @@ void cBlockPistonHandler::RetractPiston(int a_BlockX, int a_BlockY, int a_BlockZ
 	}
 
 	std::unordered_set<Vector3i, VectorHasher<int>> pushedBlocks;
-	if (!CanPushBlock(a_BlockX, a_BlockY, a_BlockZ, a_World, true, pushedBlocks, pistonMeta))
+	if (!CanPushBlock(a_BlockX, a_BlockY, a_BlockZ, a_World, false, pushedBlocks, pistonMeta))
 	{
 		// Not pushable, bail out
 		return;
@@ -300,8 +306,19 @@ void cBlockPistonHandler::RetractPiston(int a_BlockX, int a_BlockY, int a_BlockZ
 
 		AddPistonDir(moveX, moveY, moveZ, pistonMeta, 1);
 
-		// TODO Do not allow pisons to pull breakable blocks
-		a_World->SetBlock(moveX, moveY, moveZ, moveBlock, moveMeta);
+		if (cBlockInfo::IsPistonBreakable(moveBlock))
+		{
+			cBlockHandler * Handler = BlockHandler(moveBlock);
+			if (Handler->DoesDropOnUnsuitable())
+			{
+				cChunkInterface ChunkInterface(a_World->GetChunkMap());
+				cBlockInServerPluginInterface PluginInterface(*a_World);
+				Handler->DropBlock(ChunkInterface, *a_World, PluginInterface, nullptr, moveX, moveY, moveZ);
+			}
+		} else
+		{
+			a_World->SetBlock(moveX, moveY, moveZ, moveBlock, moveMeta);
+		}
 	}
 }
 
--- a/src/Blocks/BlockPiston.cpp
+++ b/src/Blocks/BlockPiston.cpp
@@ -99,22 +99,26 @@ void cBlockPistonHandler::PushBlocks(const std::unordered_set<Vector3i, VectorHa
 	cWorld * a_World, const Vector3i & a_PushDir
 )
 {
+	// Sort blocks to move the blocks first, which are farest away from the piston
+	// This prevents the overwriting of existing blocks
 	std::vector<Vector3i> sortedBlocks(a_BlocksToPush.begin(), a_BlocksToPush.end());
 	std::sort(sortedBlocks.begin(), sortedBlocks.end(), [a_PushDir](const Vector3i & a, const Vector3i & b)
 	{
 		return a.Dot(a_PushDir) > b.Dot(a_PushDir);
 	});
 
+	// Move every block
 	BLOCKTYPE moveBlock;
 	NIBBLETYPE moveMeta;
 	for (Vector3i & moveBlockPos : sortedBlocks)
-	{	
+	{
 		a_World->GetBlockTypeMeta(moveBlockPos.x, moveBlockPos.y, moveBlockPos.z, moveBlock, moveMeta);
 		a_World->SetBlock(moveBlockPos.x, moveBlockPos.y, moveBlockPos.z, E_BLOCK_AIR, 0);
 
 		moveBlockPos += a_PushDir;
 		if (cBlockInfo::IsPistonBreakable(moveBlock))
 		{
+			// Block is breakable, drop it
 			cBlockHandler * Handler = BlockHandler(moveBlock);
 			if (Handler->DoesDropOnUnsuitable())
 			{
@@ -126,6 +130,7 @@ void cBlockPistonHandler::PushBlocks(const std::unordered_set<Vector3i, VectorHa
 			}
 		} else
 		{
+			// Not breakable, just move it
 			a_World->SetBlock(moveBlockPos.x, moveBlockPos.y, moveBlockPos.z, moveBlock, moveMeta);
 		}
 	}
@@ -153,7 +158,7 @@ bool cBlockPistonHandler::CanPushBlock(
 		return true;
 	}
 
-	if(!a_RequirePushable && cBlockInfo::IsPistonBreakable(currBlock))
+	if (!a_RequirePushable && cBlockInfo::IsPistonBreakable(currBlock))
 	{
 		// Block should not be broken, when it's not in the pushing direction
 		return true;
@@ -161,11 +166,13 @@ bool cBlockPistonHandler::CanPushBlock(
 
 	if (!CanPush(currBlock, currMeta))
 	{
+		// When it's not required to push this block, don't fail
 		return !a_RequirePushable;
 	}
 
 	if (a_BlocksPushed.size() >= PISTON_MAX_PUSH_DISTANCE)
 	{
+		// Do not allow to push too much blocks
 		return false;
 	}
 
@@ -174,18 +181,20 @@ bool cBlockPistonHandler::CanPushBlock(
 		return true;  // Element exist already
 	}
 
-	if(currBlock == E_BLOCK_SLIME_BLOCK)
+	if (currBlock == E_BLOCK_SLIME_BLOCK)
 	{
 		// Try to push the other directions
-		for(const Vector3i & testDir : pushingDirs)
+		for (const Vector3i & testDir : pushingDirs)
 		{
-			if(!CanPushBlock(a_BlockX + testDir.x, a_BlockY + testDir.y, a_BlockZ + testDir.z, a_World, false, a_BlocksPushed, a_PushDir))
+			if (!CanPushBlock(a_BlockX + testDir.x, a_BlockY + testDir.y, a_BlockZ + testDir.z, a_World, false, a_BlocksPushed, a_PushDir))
 			{
+				// When it's not possible for a direction, then fail
 				return false;
 			}
 		}
 	}
 
+	// Try to push the block in front of this block
 	return CanPushBlock(a_BlockX + a_PushDir.x, a_BlockY + a_PushDir.y, a_BlockZ + a_PushDir.z, a_World, true, a_BlocksPushed, a_PushDir);
 }
 
@@ -221,6 +230,7 @@ void cBlockPistonHandler::ExtendPiston(int a_BlockX, int a_BlockY, int a_BlockZ,
 
 	PushBlocks(blocksPushed, a_World, pushDir);
 
+	// Set the extension and the piston base correctly
 	a_World->SetBlock(a_BlockX, a_BlockY, a_BlockZ, pistonBlock, pistonMeta | 0x8);
 	a_World->SetBlock(a_BlockX + pushDir.x, a_BlockY + pushDir.y, a_BlockZ + pushDir.z,
 		E_BLOCK_PISTON_EXTENSION, pistonMeta | (IsSticky(pistonBlock) ? 8 : 0)
@@ -259,12 +269,13 @@ void cBlockPistonHandler::RetractPiston(int a_BlockX, int a_BlockY, int a_BlockZ
 	a_World->BroadcastBlockAction(a_BlockX, a_BlockY, a_BlockZ, 1, pistonMeta & ~(8), pistonBlock);
 	a_World->BroadcastSoundEffect("tile.piston.in", static_cast<double>(a_BlockX), static_cast<double>(a_BlockY), static_cast<double>(a_BlockZ), 0.5f, 0.7f);
 
-	if(!IsSticky(pistonBlock))
+	if (!IsSticky(pistonBlock))
 	{
 		// No need for block pulling, bail out
 		return;
 	}
 
+	// Get the block to pull
 	a_BlockX += 2 * pushDir.x;
 	a_BlockY += 2 * pushDir.y;
 	a_BlockZ += 2 * pushDir.z;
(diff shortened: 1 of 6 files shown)


SphinxC0re:
Wow! I like the commit count :) don't squash yet, makes it easier to review

SphinxC0re:
Did you use my prototype?

SphinxC0re:
Looks good to me. @cuberite/core-team any objections?

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
 
 
 
-int cBlockPistonHandler::FirstPassthroughBlock(int a_PistonX, int a_PistonY, int a_PistonZ, NIBBLETYPE pistonmeta, cWorld * a_World)
+Vector3i cBlockPistonHandler::GetDirectionVec(int a_PistonMeta)
```
SphinxC0re: I'm sorry, but I would like some more documentation on this one

bibo38:
Thank you!
Yeah, I've looked at your prototype and found, that it's the best way to make it, because you'll need recursion. Also I have some thoughts about this code, which I am not quite sure:

- In the CanPull method I retrieve every block, which needs to be pulled and put it into the set, but in the PushBlocks method I retrieve the blockdata again to check for breaking blocks. Maybe creating a special struct to carry this extra data is better to reduce the GetBlock calls to cWorld.
- To move every block, I set the previous block to air. But because of this pushing this block is overwritten again. Maybe do a little lookup, before calling the SetBlock method to reduce this calls.

This thoughts are mainly about the performance of calling cWorld methods too often. I don't know, how efficient this is implemented, but the current design is the simplest approach I can think of(Better to understand instead of (nonsense) optimisations).

SphinxC0re:
Reducing the cWorld calls would be good, but think about many pistons pushing many slimeblocks against each other. There could be a race condition if you'd put everything into a separate struct

src/Blocks/BlockPiston.cpp
bibo38: Maybe changing the GetDirectionVec method name into MetaDataToVector name?
P.S. I've documented it.

SphinxC0re: Good idea, what about ```VectorFromMetaData``` ?
You know what? Decide for yourself. I will take a look at the result

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
 
 
 
-int cBlockPistonHandler::FirstPassthroughBlock(int a_PistonX, int a_PistonY, int a_PistonZ, NIBBLETYPE pistonmeta, cWorld * a_World)
+Vector3i cBlockPistonHandler::VectorFromMetaData(int a_PistonMeta)
```
tigerw: Suggested name "MetadataToOffset"

tigerw: a_PistonMeta of type NIBBLETYPE?

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
...
+		case 2: return Vector3i( 0,  0, -1);
+		case 3: return Vector3i( 0,  0,  1);
+		case 4: return Vector3i(-1,  0,  0);
+		case 5: return Vector3i( 1,  0,  0);
+		default:
 		{
-			// This block breaks when pushed, extend up to here
-			return ret;
+			LOGWARNING("%s: invalid direction %d, ignoring", __FUNCTION__, a_PistonMeta & 0x07);
+			return Vector3i();
```
tigerw: Suggested return format `return {x, y, z}`

src/Blocks/BlockPiston.cpp
tigerw: Suggested return value `const Vector3i` and possibly make the function const too.

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
...
+
+
+
+
+bool cBlockPistonHandler::CanPushBlock(
+	int a_BlockX, int a_BlockY, int a_BlockZ, cWorld * a_World, bool a_RequirePushable,
+	std::unordered_set<Vector3i, VectorHasher<int>> & a_BlocksPushed, const Vector3i & a_PushDir
+)
+{
+	const static std::array<Vector3i, 6> pushingDirs = {{ Vector3i(-1, 0, 0), Vector3i(1, 0, 0), Vector3i(0, -1, 0), Vector3i(0, 1, 0),
```
tigerw: Suggested format:
```cpp
static const ... =  {
    {1, 2, 3},
    ...
};
```

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
...
+
+	if (!a_BlocksPushed.emplace(a_BlockX, a_BlockY, a_BlockZ).second || cBlockInfo::IsPistonBreakable(currBlock))
+	{
+		return true;  // Element exist already
+	}
+
+	if (currBlock == E_BLOCK_SLIME_BLOCK)
+	{
+		// Try to push the other directions
+		for (const Vector3i & testDir : pushingDirs)
```
tigerw: Suggested Vector3i -> auto

tigerw:
Just minor cosmetics. I'm sure the algorithm's fine.

src/Blocks/BlockPiston.cpp
SphinxC0re: Why should the function made const?

madmaxoft: The function should be `static`, because it doesn't depend on any state of the instance. No need to make the return value `const` if it is a by-value return.

src/Blocks/BlockPiston.cpp
madmaxoft: Not really, that would hide the fact that it's returning a `Vector3i`

madmaxoft: I think adding an `ASSERT` would be a good choice here. We want any code calling this with bad parameters to fail hard in debug mode.

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
...
+	}
+}
+
+
+
+
+
+void cBlockPistonHandler::PushBlocks(const std::unordered_set<Vector3i, VectorHasher<int>> & a_BlocksToPush,
+	cWorld * a_World, const Vector3i & a_PushDir
+)
```
madmaxoft: Style: Put the starting parameters on a separate line as well:
```cpp
void cBlockPistonHandler::PushBlocks(
	const std::unordered_set<Vector3i, VectorHasher<int>> & a_BlocksToPush,
	cWorld * a_World, const Vector3i & a_PushDir
)
```

src/Blocks/BlockPiston.cpp
madmaxoft: Nope. Make it apparent that it's a class.

src/Blocks/BlockPiston.cpp
```diff
@@ -86,28 +73,129 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
...
+
+
+
+bool cBlockPistonHandler::CanPushBlock(
+	int a_BlockX, int a_BlockY, int a_BlockZ, cWorld * a_World, bool a_RequirePushable,
+	std::unordered_set<Vector3i, VectorHasher<int>> & a_BlocksPushed, const Vector3i & a_PushDir
+)
+{
+	const static std::array<Vector3i, 6> pushingDirs = {{ Vector3i(-1, 0, 0), Vector3i(1, 0, 0), Vector3i(0, -1, 0), Vector3i(0, 1, 0),
+		Vector3i(0, 0, -1), Vector3i(0, 0, 1) }};
```
madmaxoft: Style: respect curly braces:
```cpp
const static std::array<Vector3i, 6> pushingDirs = {{
	Vector3i(-1, 0, 0), Vector3i(1, 0, 0),
	Vector3i(0, -1, 0), Vector3i(0, 1, 0),
	Vector3i(0, 0, -1), Vector3i(0, 0, 1)
}};
```

src/Blocks/BlockPiston.cpp
```diff
@@ -126,54 +214,27 @@ void cBlockPistonHandler::ExtendPiston(int a_BlockX, int a_BlockY, int a_BlockZ,
 		return;
 	}
 
-	int dist = FirstPassthroughBlock(a_BlockX, a_BlockY, a_BlockZ, pistonMeta, a_World);
-	if (dist < 0)
+	Vector3i pushDir = VectorFromMetaData(pistonMeta);
+
+	std::unordered_set<Vector3i, VectorHasher<int>> blocksPushed;
```
madmaxoft: The ridiculous type name should be `typedef`ed

src/Blocks/BlockPiston.cpp
```diff
@@ -126,54 +214,27 @@ void cBlockPistonHandler::ExtendPiston(int a_BlockX, int a_BlockY, int a_BlockZ,
...
+	if (!CanPushBlock(a_BlockX + pushDir.x, a_BlockY + pushDir.y, a_BlockZ + pushDir.z,
+		a_World, true, blocksPushed, pushDir)
+	)
```
madmaxoft: Again, starting parameters on a separate line

src/Blocks/BlockPiston.cpp
```diff
@@ -266,8 +331,3 @@ void cBlockPistonHeadHandler::OnDestroyedByPlayer(cChunkInterface & a_ChunkInter
 		a_WorldInterface.SpawnItemPickups(Pickups, a_BlockX + 0.5, a_BlockY + 0.5, a_BlockZ + 0.5);
 	}
 }
-
-
-
-
-
```
madmaxoft: Keep the empty lines

src/Blocks/BlockPiston.h
```diff
@@ -79,6 +81,11 @@ class cBlockPistonHandler :
 		}
 	}
 
+	/** This method converts the magic piston metadata into a direction vector.
+		This vector has a length of 1 and points into the direction, in which the piston will extend.
+	*/
```
madmaxoft: -> `/** Converts piston block's metadata into a unit vector representing the direction in which the piston will extend. */

src/Blocks/BlockPiston.cpp
bibo38: Yeah, I'll change it. But the function is just a helper and does not modify the cPistonHandler class. So I'll make this function const too.

src/Blocks/BlockPiston.h
bibo38: Nice and short :+1: . Thank you!

src/Blocks/BlockPiston.cpp
bibo38: @madmaxoft This code fails with the following error:

    Blocks/BlockPiston.cpp(149): 2 .. 55: Brace should be on a separate line

Should I really indent the second brace? Or what would you choose?

madmaxoft: Sorry, the double-brace should be on a separate line as well, forgot about that.

src/Blocks/BlockPiston.cpp
madmaxoft: `static` functions cannot be `const` - they don't have an object to not modify :) Keep it `static` then, just rename and change the param type..

src/Blocks/BlockPiston.cpp
bibo38: This would look like this:

    ... = 
    {
    {
        ....
    }}

Or similar to this(I could indent the inner braces). But the check code doesn't allow to have two opening braces in the same line.

src/Blocks/BlockPiston.cpp
bibo38: Yeah, I always thought it would allow to not modify the class parameters, but it only protects the this pointer ;D.

src/Blocks/BlockPiston.cpp
bibo38: @tigerw @madmaxoft Doesn't this defeat the readability of the code, because you don't know the type?
My view of auto is, that it helps to avoid complex types for variables, like iterators:

    auto it = x.begin();
    // std::map<int, std::string>::const_iterator it = x.begin();

But if auto is preferred, should I also replace the other foreach loop(s)?

madmaxoft: `auto` is preferred because of easy refactoring. All new loops should be written with `auto` and range-based, if possible.

src/Blocks/BlockPiston.cpp
tigerw: It's the return style the redstone sim uses, so consistency?

src/Blocks/BlockPiston.cpp
```diff
@@ -45,10 +32,10 @@ void cBlockPistonHandler::OnDestroyed(cChunkInterface & a_ChunkInterface, cWorld
...
 	NIBBLETYPE OldMeta = a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ);
 
-	int newX = a_BlockX;
-	int newY = a_BlockY;
-	int newZ = a_BlockZ;
-	AddPistonDir(newX, newY, newZ, OldMeta, 1);
+	const Vector3i pushDir = MetadataToOffset(OldMeta);
+	int newX = a_BlockX + pushDir.x;
+	int newY = a_BlockY + pushDir.y;
+	int newZ = a_BlockZ + pushDir.z;
```
worktycho: I would suggest refactoring this code to use the vector form only.

src/Blocks/BlockPiston.cpp
```diff
@@ -192,39 +268,43 @@ void cBlockPistonHandler::RetractPiston(int a_BlockX, int a_BlockY, int a_BlockZ
 		return;
 	}
 
+	Vector3i pushDir = MetadataToOffset(pistonMeta);
+
 	// Check the extension:
-	AddPistonDir(a_BlockX, a_BlockY, a_BlockZ, pistonMeta, 1);
-	if (a_World->GetBlock(a_BlockX, a_BlockY, a_BlockZ) != E_BLOCK_PISTON_EXTENSION)
+	if (a_World->GetBlock(a_BlockX + pushDir.x, a_BlockY + pushDir.y, a_BlockZ + pushDir.z) != E_BLOCK_PISTON_EXTENSION)
```
worktycho: WOuld it be worth refactoring this function to only use vectors?

bibo38: I could add a overloading, which uses vectors and calls the other function, but if you want to remove the GetBlock(int, int, int) function, then this should be done in annother PR.

worktycho: Theres already an overload of getBlock to use vectors, I was referring to RetractPiston

bibo38:
Any other recommendations?

src/Blocks/BlockPiston.cpp
```diff
@@ -86,105 +72,187 @@ bool cBlockPistonHandler::GetPlacementBlockTypeMeta(
...
+	cWorld * a_World, const Vector3i & a_PushDir
+)
+{
+	// Sort blocks to move the blocks first, which are farest away from the piston
+	// This prevents the overwriting of existing blocks
+	std::vector<Vector3i> sortedBlocks(a_BlocksToPush.begin(), a_BlocksToPush.end());
+	std::sort(sortedBlocks.begin(), sortedBlocks.end(), [a_PushDir](const Vector3i & a, const Vector3i & b)
+	{
+		return a.Dot(a_PushDir) > b.Dot(a_PushDir);
+	});
```
worktycho: I've no idea if this is it, but if you make a_BlocksToPush a std::map of distance to blocks, you can eliminate the sort.

worktycho:
Other than the one comment I'm not sure about, I think this is mergable.

src/Blocks/BlockPiston.cpp
bibo38: I've created a `std::unordered_set` to allow a O(1) insertion and finding performance. This is interesting for the CanPushBlock method to allow easy insertation and duplicate checking.
A `std::map<int, std::unordered_set>` would need O(log N) performance for every insertation and duplicate checking.

I don't know, if this is better than the current one time O(N*log N) time complexity(sorting) and the O(N) extra space(the vector to sort).

worktycho: Thats why I wasn't sure. I'm going to merge as is, and if we find its a perfomance issue, we can fix it later.

Issue #2613, opened 2015-11-06
Cave spiders can poison without hitting


Gargaj:
```cAggressiveMonster::Attack``` is ```void```, so there's no feedback on whether the attack has succeeded or if it's still in cooldown. Because of this, ```cCaveSpider::Attack``` calls the ```super``` but poisons the target regardless.

Gargaj:
(I can try doing a pull on this.)

bibo38:
Great :smile: !
Better we have a dirty solution, than a bug more.
Looks like the Attack method is bad implemented, because the name suggests, that the monster attacks something.

PR #2615, opened 2015-11-07
Fix cavespider poisoning even if attack is in cooldown

Gargaj:
cf. https://github.com/cuberite/cuberite/issues/2613

### DIFF
--- a/src/Mobs/AggressiveMonster.h
+++ b/src/Mobs/AggressiveMonster.h
@@ -20,7 +20,7 @@ public:
 	virtual void InStateChasing(std::chrono::milliseconds a_Dt) override;
 	
 	virtual void EventSeePlayer(cEntity *) override;
-	virtual void Attack(std::chrono::milliseconds a_Dt);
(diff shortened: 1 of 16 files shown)


SphinxC0re:
I don't have any objections. @cuberite/core-team ?

src/Mobs/AggressiveMonster.h
```diff
@@ -20,7 +20,7 @@ class cAggressiveMonster :
 	virtual void InStateChasing(std::chrono::milliseconds a_Dt) override;
 	
 	virtual void EventSeePlayer(cEntity *) override;
-	virtual void Attack(std::chrono::milliseconds a_Dt);
+	virtual bool Attack(std::chrono::milliseconds a_Dt);
```
madmaxoft: This could use some commenting - when it's called, what it's supposed to do, what the default implementation does and what the return value means.

Gargaj: I thought about it but I wasn't sure how to explain it; essentially it returns true if the attack was "successful" (even if it did no damage).

I'll add a comment later today

Issue #2614, opened 2015-11-06
Implement pushing / shoving


Gargaj:
Potential logic: if two mobs are within a certain distance (in 2D) to each other or to a player, increase the speed of the stationary one away from the other.

I looked into the source but haven't really figured out a way to do the distance check to several entities efficiently. (```ForEachEntityInBox``` maybe?)

NiLSPACE:
You mean pushing a mob? That shouldn't be too hard to implement. Something like this should work: 
```cpp
void cPawn::Tick(...)
{
	class Pusher : public cEntityCallback
	{
	public:
		cEntity * m_Pusher;
		
		Pusher(cEntity a_Pusher) :
			m_Pusher(a_Pusher)
		{
		}
		
		virtual bool Item(cEntity * a_Entity) override
		{
			if (a_Entity->GetUniqueID() == m_Pusher->GetUniqueID())
			{
				return false;
			}
			
			if ((a_Entity->GetPosition() - m_Pusher->GetPosition()).SqrLength() > /* Some size */)
			{
				return false;
			}
			
			a_Entity->AddSpeed(a_Entity->GetPosition() - m_Pusher->GetPosition());
			return false;
		}
	} Callback(/*Current entity*/)
	
	GetWorld().ForEachEntityInBox(new cBoundingBox(GetPosition(), GetWidth(), GetHeight()), Callback);
}
```

Gargaj:
I think the speed should be reciprocal to the distance i.e. you push them more when you are up close.

Gargaj:
Also, should only push mobs, but not other players. (It used to be both but got changed around 1.5 i think?)

PR #2617, opened 2015-11-07
Implement pushing/shoving

Gargaj:
Based on @NiLSPACE's suggestion
(https://github.com/cuberite/cuberite/issues/2614)

### DIFF
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -4,6 +4,7 @@
 #include "Pawn.h"
 #include "../World.h"
 #include "../Bindings/PluginManager.h"
+#include "BoundingBox.h"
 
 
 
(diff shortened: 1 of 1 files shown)


NiLSPACE:
I assume you did test if it actually works?

Gargaj:
Yeah I did. 
* Minecarts work although they start floating for some reason.
* Boats work but only if they're on water because on land you cant walk into them.
* Mobs work, tested both with me pushing them off trees with glee and organizing cow-moshpits in fenced off areas.

NiLSPACE:
Oh, it's already merged

source83:
How is this working with powered rails so far?

src/Entities/Pawn.cpp
```diff
@@ -43,6 +44,41 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+
+			// we only push other mobs, boats and minecarts
+			if ((a_Entity->GetEntityType() != etMonster) && (a_Entity->GetEntityType() != etMinecart) && (a_Entity->GetEntityType() != etBoat))
+			{
+				return false;
+			}
+
+			Vector3d v3Delta = a_Entity->GetPosition() - m_Pusher->GetPosition();
+			v3Delta.y = 0.0;  // we only push sideways
+			v3Delta *= 1.0 / v3Delta.Length();  // we push harder if we're close
```
madmaxoft: What if two entities are created at exactly the same spot? Division by zero?

NiLSPACE:
@Zee1234 reported a really weird bug on [reddit/admincraft](https://www.reddit.com/r/admincraft/comments/3tfjx7/i_want_to_set_up_a_small_server_for_friendsfamily/cx691m6). I believe it's caused by this PR.

SphinxC0re:
This PR causes a lot of trouble and will cause trouble in the future. I'm for revert. And for retry ;)

Zee1234:
Bug was NOT present a month ago (just checked on an October 19th compile).

Bug IS present for ALL mobs (just tested) and works with snowballs as well. Cuberite was compiled at 2015-11-15 3 AM GMT-5. Chickens are the most pronounced, although villagers have some crazy hang time. Found some other weird things just now, but they are completely unrelated. Going to the forum now before I forget them.

PR #2619, opened 2015-11-07
Use .travis.yml to whitelist coverity_scan for push builds.

worktycho:
Travis will only run coverity on a push build, but we want to disable push builds for most of the project. This whitelist should be a good compromise.


SphinxC0re:
Please rebase and squash onto current master

worktycho:
Rebased.

PR #2621, opened 2015-11-07
Fix pressure plate power levels

Gargaj:
Hilariously, the block type was being passed as a power level.
This made the block underneath the pressure plate have the wrong powerlevel -> current only travelled one block. Current on the level of the _pressure plate_ travelled correctly, however.


worktycho:
@tigerw

tigerw:
Its fine, yes.

PR #2624, opened 2015-11-07
Only send creeper metadata updates on change

Gargaj:
Otherwise this keeps spamming the metadata packet for all the creepers on the server.


Gargaj:
There's still a problem with creepers staying swollen if they don't explode, but I don't know what the server should do to fix that, since there's no "explosion timer" being sent out.

PR #2625, opened 2015-11-08
Neutral mobs should be neutral

Gargaj:
Right now endermen / pigmen attack immediately because they inherited ```cAggressiveMonster::EventSeePlayer```. (Wolves are an exception because ```cWolf::Tick```.)


madmaxoft:
Looks innocent enough for me, merge if no objections.

PR #2626, opened 2015-11-08
Fix memory leak

worktycho:
Fixes CID 132249


madmaxoft:
How could this have slipped through? Our code reviews are weak.

SphinxC0re:
There have been so many good code submissions in th  last view weeks. I think that is what weakened our reviews

Seadragon91:
My pull request was around 30 days.

PR #2630, opened 2015-11-09
Fix damage in shallow water (fixes #2622)

### LINKED ISSUE #2622: Taking damage in shallow water
This happens because the damage dealt in ```cPlayer::SetTouchGround``` is called by ```cClientHandle::HandlePlayerPos``` (which is a network reaction), but the swimming flag is set in ```cEntity::SetSwimState``` (which is called in a tick). This asynchronism allows the player to register to have landed on the ground but not yet registered as swimming.

Here's a bit of extended logging of me jumping into a single block deep pool of water from a height:
```
[ddbc180b|17:00:40] Y: 74.249, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 74.171, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 74.016, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 73.785, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 73.481, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 73.104, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 72.657, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 72.140, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 71.555, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 70.903, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 70.186, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 69.405, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 68.561, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 67.655, IsSwimming: 0, TouchGround: 0
[ddbc180b|17:00:40] Y: 67.000, IsSwimming: 0, TouchGround: 1
[ddbc180b|17:00:40] Dealing 3 damage
[ddbc180b|17:00:41] Y: 67.000, IsSwimming: 1, TouchGround: 1
```

As you can see, the waterblock at Y=67 is registered as swimming, but one packet too late which deals me damage even though I landed in water.

Gargaj:
Also add future code for liquid fall dampening.

This might need a bit of review and it doesn't fix a higher level problem: A player dropping through a single block of lava or water should get their ```m_LastGroundHeight``` reset, but since the client only sends updates every so often, it is entirely possible to to "phase through" that single block and get more fall damage.

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -461,8 +461,13 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
 
 	static const auto HalfWidth = GetWidth() / 2;
 	static const auto EPS = 0.0001;
+
+	BLOCKTYPE BlockAtFoot = GetWorld()->GetBlock(POS_TOINT);
+	bool IsFootInWater = IsBlockWater(BlockAtFoot);
+	bool IsFootInLiquid = IsFootInWater || IsBlockLava(BlockAtFoot) || (BlockAtFoot == E_BLOCK_COBWEB);  // okay so cobweb is not _technically_ a liquid...
+
 	if (
-		!IsSwimming() && !IsFlying() &&
+		!IsFlying() &&
 		(
 			(
 				((GetPosY() >= 1) && (GetPosY() <= 255) && ((GetPosY() - POSY_TOINT) <= EPS)) &&
@@ -498,7 +503,7 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
 	)
 	{
 		auto Damage = static_cast<int>(m_LastGroundHeight - GetPosY() - 3.0);
-		if (Damage > 0)
+		if ((Damage > 0) && !IsFootInWater)
 		{
 			// cPlayer makes sure damage isn't applied in creative, no need to check here
 			TakeDamage(dtFalling, nullptr, Damage, Damage, 0);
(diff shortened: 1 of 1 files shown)


SphinxC0re:
Looks good to me. Merge if CI succeeds

madmaxoft:
As soon as we got rid of one trigger-happy bear, he's been replaced by a trigger-happy sphinx. Oh my...

src/Entities/Player.cpp
```diff
@@ -461,8 +461,13 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
 
 	static const auto HalfWidth = GetWidth() / 2;
 	static const auto EPS = 0.0001;
+
+	BLOCKTYPE BlockAtFoot = GetWorld()->GetBlock(POS_TOINT);
```
madmaxoft: This will crash the server when the player is above or below the world

Seadragon91: Looking at `GetBlock` it will return 0, if there is no match found in the Chunk.

madmaxoft: It will assert, because you're supposed to check the Y coord before calling it.

PR #2631, opened 2015-11-10
Avoid division by zero

Gargaj:
It won't do much because direction will still be a nullvector, but it at least avoids a CPU spike.

### DIFF
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -70,7 +70,7 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 			Vector3d v3Delta = a_Entity->GetPosition() - m_Pusher->GetPosition();
 			v3Delta.y = 0.0;  // we only push sideways
-			v3Delta *= 1.0 / v3Delta.Length();  // we push harder if we're close
+			v3Delta *= 1.0 / (v3Delta.Length() + 0.01);  // we push harder if we're close
 			// QUESTION: is there an additional multiplier for this? current shoving seems a bit weak
 
 			a_Entity->AddSpeed(v3Delta);


src/Entities/Pawn.cpp
```diff
@@ -70,7 +70,7 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 			Vector3d v3Delta = a_Entity->GetPosition() - m_Pusher->GetPosition();
 			v3Delta.y = 0.0;  // we only push sideways
-			v3Delta *= 1.0 / v3Delta.Length();  // we push harder if we're close
+			v3Delta *= 1.0 / (v3Delta.Length() + 0.01);  // we push harder if we're close
```
NiLSPACE: Is it possible for the length function to return negative numbers? If so it might return -0.01 which would make this 0 again.

Gargaj: Can a square root be negative?

Schwertspize: I don't know but if so, you have to wrap the length() into an std::abs() which returns the positive value. also you don't wanna gave a negative v3Delta which would result in pulling...

Gargaj: I was being somewhat sarcastic - length cannot be negative.

madmaxoft: A square root is always non-negative. I think this is the proper way to fix the issue.

NiLSPACE: Right.. I keep forgetting that Length is just the square root.

Schwertspize: Oh yes... we are talking about vectors so the length is (like Pythagoras said) length l^2 = a^2 + b^2 + c^3 where a,b,c are the coordinates..

worktycho:
Rebase to master.

src/Entities/Pawn.cpp
tigerw: I guess you can get √(|*x*|)**i**

PR #2632, opened 2015-11-10
Avoid reading out of bounds

Gargaj:
as per https://github.com/cuberite/cuberite/pull/2630#discussion_r44377411

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -462,7 +462,17 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
 	static const auto HalfWidth = GetWidth() / 2;
 	static const auto EPS = 0.0001;
 
-	BLOCKTYPE BlockAtFoot = GetWorld()->GetBlock(POS_TOINT);
+	/* Since swimming is decided in a tick and is asynchronous to this, we have to check for dampeners ourselves.
+	The behaviour as of 1.8.8 is the following:
+	- Landing in water alleviates all fall damage
+	- Passing through any liquid (water + lava) and cobwebs "slows" the player down,
+		i.e. resets the fall distance to that block, but only after checking for fall damage
+		(this means that plummeting into lava will still kill the player via fall damage, although cobwebs
+		will slow players down enough to have multiple updates that keep them alive)
+
+	With this in mind, we first check the block at the player's feet, and decide which behaviour we want to go with.
+	*/
+	BLOCKTYPE BlockAtFoot = (cChunkDef::IsValidHeight(GetPosY())) ? GetWorld()->GetBlock(POS_TOINT) : E_BLOCK_AIR;
 	bool IsFootInWater = IsBlockWater(BlockAtFoot);
 	bool IsFootInLiquid = IsFootInWater || IsBlockLava(BlockAtFoot) || (BlockAtFoot == E_BLOCK_COBWEB);  // okay so cobweb is not _technically_ a liquid...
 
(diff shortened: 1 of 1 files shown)


src/Entities/Player.cpp
```diff
@@ -462,7 +462,7 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
 	static const auto HalfWidth = GetWidth() / 2;
 	static const auto EPS = 0.0001;
 
-	BLOCKTYPE BlockAtFoot = GetWorld()->GetBlock(POS_TOINT);
+	BLOCKTYPE BlockAtFoot = ((GetPosY() >= 1) && (GetPosY() <= 255)) ? GetWorld()->GetBlock(POS_TOINT) : E_BLOCK_AIR;
```
SphinxC0re: Could you add a little more documentation to this line

Gargaj: Sure

madmaxoft: Use `cChunkDef::Height` instead of `255`.

madmaxoft:
Other than naming the magic number, this looks good and is good for a merge after a fix.

Gargaj:
What magic number?

src/Entities/Player.cpp
```diff
@@ -462,7 +462,17 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
...
+	The behaviour as of 1.8.8 is the following:
+	- Landing in water alleviates all fall damage
+	- Passing through any liquid (water + lava) and cobwebs "slows" the player down,
+		i.e. resets the fall distance to that block, but only after checking for fall damage
+		(this means that plummeting into lava will still kill the player via fall damage, although cobwebs
+		will slow players down enough to have multiple updates that keep them alive)
+
+	With this in mind, we first check the block at the player's feet, and decide which behaviour we want to go with.
+	*/
+	BLOCKTYPE BlockAtFoot = ((GetPosY() >= 1) && (GetPosY() <= 255)) ? GetWorld()->GetBlock(POS_TOINT) : E_BLOCK_AIR;
```
madmaxoft: Use `cChunkDef::Height` instead of `255`.

Gargaj: In all fairness I just copied the code that was underneath it :D

madmaxoft: Gotta clean it up at some point :)

NiLSPACE: We were talking about a IsValidHeight function at some point right?

Gargaj: I did a little work on this bit that cleans the if-nado underneath it up slightly (cf. the fall-damage-on-stairs issue) but that'll take a bit of time to get right since it needs a BlockHandler-level change.
Until then, this'll have to do.

SphinxC0re:
#2634

src/Entities/Player.cpp
```diff
@@ -462,7 +462,17 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
 	static const auto HalfWidth = GetWidth() / 2;
 	static const auto EPS = 0.0001;
 
-	BLOCKTYPE BlockAtFoot = GetWorld()->GetBlock(POS_TOINT);
+	/* Since swimming is decided in a tick and is asynchronous to this, we have to check for dampeners ourselves.
```
tigerw: Its not async is it? The tick thread at present parses packets.

Gargaj: I might have misread. Either way even in the same thread the swimming state is set AFTER the damage is dealt, which won't work.

Gargaj:
I rebased this patch on top of https://github.com/cuberite/cuberite/pull/2634

madmaxoft:

PR #2633, opened 2015-11-10
Added brewing.txt to list of files packed with the windows build.

madmaxoft:
The file was missing from windows builds.


Seadragon91:
Ah didn't know that, thank you.

PR #2634, opened 2015-11-10
Added cChunkDef::IsValidHeight()

SphinxC0re:


src/ChunkDef.h
```diff
@@ -115,6 +115,11 @@ class cChunkDef
 		return Vector3i(a_RelBlockPosition.x + a_ChunkX * Width, a_RelBlockPosition.y, a_RelBlockPosition.z + a_ChunkZ * Width);
 	}
 
+	inline static bool IsValidHeight(int a_height)
+	{
+		return ((a_height >= 1) && (a_height <= Height));
```
madmaxoft: Capitalization is wrong. `a_Height`

madmaxoft:
See how easy coding is once you get the rage going? :D

SphinxC0re:
I'm sick of inconsistency in height checks! Sick of it!

madmaxoft:
Yessssss. Let the hate flow :D

NiLSPACE:
Good Good

Gargaj:
Shouldn't that be "< Height", since Height is 256?

madmaxoft:
Yup. And that's why we're refactoring it out, because every third instance of this check is wrong.

tigerw:
You've all gone mad.

worktycho:
Isn't 0 a valid height?

SphinxC0re:
Okay what do I have to check for? @madmaxoft

SphinxC0re:
Is this mergeable? @cuberite/core-team

NiLSPACE:
Perhaps we should export cChunkDef to Lua as well. For plugins that implement their own generator.

PR #2637, opened 2015-11-11
Made IsThread terminate flag into atomic running flag

### LINKED ISSUE #2636: cIsThread::m_ShouldTerminate synchronisation
The variable is being set in the world thread when Stop() is called but accessed in Execute() from another thread. The variable is declared volatile but this provides no atomicity guarantees.

worktycho:
Fixes #2636

### DIFF
--- a/src/ChunkSender.cpp
+++ b/src/ChunkSender.cpp
@@ -81,7 +81,7 @@ cChunkSender::~cChunkSender()
 
 bool cChunkSender::Start()
 {
-	m_ShouldTerminate = false;
+	m_KeepRunning.test_and_set();
 	return super::Start();
 }
 
@@ -91,7 +91,7 @@ bool cChunkSender::Start()
 
 void cChunkSender::Stop(void)
 {
-	m_ShouldTerminate = true;
+	m_KeepRunning.clear();
 	m_evtQueue.Set();
 	Wait();
 }
@@ -184,7 +184,7 @@ void cChunkSender::RemoveClient(cClientHandle * a_Client)
 
 void cChunkSender::Execute(void)
 {
-	while (!m_ShouldTerminate)
+	while (m_KeepRunning.test_and_set())
 	{
 		m_evtQueue.Wait();
 
--- a/src/DeadlockDetect.cpp
+++ b/src/DeadlockDetect.cpp
@@ -64,7 +64,7 @@ bool cDeadlockDetect::Start(int a_IntervalSec)
 void cDeadlockDetect::Execute(void)
 {
 	// Loop until the signal to terminate:
-	while (!m_ShouldTerminate)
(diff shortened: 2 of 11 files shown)


madmaxoft:
Can't you use something that doesn't have a `test_and_set`, because it's misleading that a test actually sets a value?

worktycho:
Theres atomic_bool, but its not guarenteed to be lock free, as not all platforms have support for an atomic load instruction.

madmaxoft:
What is the danger of using a non-lock-free variant? If it's a (reasonable) performance hit, I think we can live with that, since it's just a flag to tell the thread to terminate, after all.

worktycho:
The problem is that any platform that doesn't have an atomic load instruction is probably not going to have fast mutexes, so thats exactly when you don't want to be using mutex's.

worktycho:
So it would probably involve at least two syscalls to on load an atomic bool on a platform that doesn't have lock free atomic_bool.

Seadragon91:
@worktycho forgot to update submodules before commit? :)

worktycho:
Probably, I'm working on several pre-submodulisation branches,

madmaxoft:
I still don't like this. Now the flag's description is kinda wrong. The flag is set to false to terminate the thread, but the thread then resets it back to true after the first check. We can't have multiple places checking the flag because of the reset.

worktycho:
You can have multiple checks, its just a risk they might not notice because theres a gap between one thread reading it and resting it. Look at what I did with the lighting thread.

madmaxoft:
Which is another reason why I don't like the change - when reading such a piece of code, you go just "wtf?!"

SphinxC0re:
Do you have a better idea or would you go with "keep as it is"

worktycho:
We can't really keep as is.
The other two options are: wrap atomic_flag so that it auto reclears when reading, or risk performance issues with atomic_bool.

SphinxC0re:
What is the difference between ```std::atomic<bool>``` and ```std::atomic_bool``` ?

worktycho:
nothing, std::atomic_bool is just a typedef for std::atomic<bool>

madmaxoft:
I haven't found much on the performance, but all info I found suggested that there's not much difference between `atomic_flag` and `atomic_bool` performance-wise.

worktycho:
I depends on the implementations. pretty much all major implementations do have a lock free atomic bool, I'm arguing from a purely standard based point of view.

madmaxoft:
If they don't have a lock-free atomic bool, we'll be having much worse performance problems than threads checking whether they should exit. Please do rewrite this to use the `atomic_bool`.

tigerw:
(xoft being difficult again :P)

PR #2623, opened 2015-11-07
Prevent fall damage, when falling on slimeblocks

bibo38:
**NOT COMPLETE YET**
See: http://minecraft.gamepedia.com/Slimeblock#Usage

This is interesting for #1404.
I've merged the HandleFalling method from cMonster into cPawn to reduce code duplication.
But I have no idea, how to implement the fall damage, when sneaking and falling on slimeblocks.

PR #2638, opened 2015-11-11
Implement block heights + adapt ground checks

### LINKED ISSUE #2582: Walking down long sets of stairs is seen as falling.
When walking down a long stairs downward, the server sees this as the player falling.
So if you walk or run down a long enough set of stairs, the player is killed.

Gargaj:
Fixes #2582 but do please review

The meat of the change is at https://github.com/Gargaj/cuberite/blob/61bd3b615fe2ddf37040ba062ccb3e12c6eea8b9/src/Entities/Player.cpp#L479 and basically simplifies the previous giant tree of conditions (https://github.com/cuberite/cuberite/blob/5e840cf86f4c55f6c4da457a7a49b1ba4b50b516/src/Entities/Player.cpp#L479) down to a single loop + per-block override possibilities.

In my experience the actual per-block overrides aren't necessary, either: if we assume blocks are taller than what they actually are, damage is dealt adequatly (erring towards "yes there is a block" vs the current behaviour where half-blocks suspend the player in mid-air since the client keeps the player "mid-air").

### DIFF
--- a/src/BlockInfo.cpp
+++ b/src/BlockInfo.cpp
@@ -584,6 +584,27 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
 	a_Info[E_BLOCK_STONE               ].m_CanBeTerraformed = true;
 
 
+	// Block heights:
+	a_Info[E_BLOCK_BED                 ].m_BlockHeight = 0.5625;  // 9 pixels
+	a_Info[E_BLOCK_CAKE                ].m_BlockHeight = 0.5;     // 8 pixels
+	a_Info[E_BLOCK_ENCHANTMENT_TABLE   ].m_BlockHeight = 0.75;    // 12 pixels
+	a_Info[E_BLOCK_STONE_SLAB          ].m_BlockHeight = 0.5;
+	a_Info[E_BLOCK_WOODEN_SLAB         ].m_BlockHeight = 0.5;
+	a_Info[E_BLOCK_SNOW                ].m_BlockHeight = 0.125;   // one layer is 1 / 8 (2 pixels) tall
+	a_Info[E_BLOCK_ACACIA_FENCE        ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_ACACIA_FENCE_GATE   ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_BIRCH_FENCE         ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_BIRCH_FENCE_GATE    ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_DARK_OAK_FENCE      ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_DARK_OAK_FENCE_GATE ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_FENCE               ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_OAK_FENCE_GATE      ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_JUNGLE_FENCE        ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_JUNGLE_FENCE_GATE   ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_SPRUCE_FENCE        ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_SPRUCE_FENCE_GATE   ].m_BlockHeight = 1.5;
+
+
 	// Block place sounds:
 	a_Info[E_BLOCK_STONE               ].m_PlaceSound = "dig.stone";
 	a_Info[E_BLOCK_GRASS               ].m_PlaceSound = "dig.grass";
--- a/src/Blocks/BlockHandler.cpp
+++ b/src/Blocks/BlockHandler.cpp
@@ -553,6 +553,16 @@ bool cBlockHandler::DoesDropOnUnsuitable(void)
 
 
 
+/* default functionality: only test for height, since we assume full voxels with varying height */
+bool cBlockHandler::IsInsideBlock(const Vector3d & a_Position, const BLOCKTYPE a_BlockType, const NIBBLETYPE a_BlockMeta)
+{
+	return a_Position.y < cBlockInfo::GetBlockHeight(a_BlockType);
+}
+
+
+
+
+
 void cBlockHandler::Check(cChunkInterface & a_ChunkInterface, cBlockPluginInterface & a_PluginInterface, int a_RelX, int a_RelY, int a_RelZ, cChunk & a_Chunk)
 {
 	if (!CanBeAt(a_ChunkInterface, a_RelX, a_RelY, a_RelZ, a_Chunk))
--- a/src/Blocks/BlockHandler.h
+++ b/src/Blocks/BlockHandler.h
@@ -124,7 +124,11 @@ public:
 	/** Returns if this block drops if it gets destroyed by an unsuitable situation.
 	Default: true */
 	virtual bool DoesDropOnUnsuitable(void);
-	
+
+	/** Tests if a_Position is inside the block where a_Position is relative to the origin of the block
+	Note that this is considered from a "top-down" perspective i.e. empty spaces on the bottom of a block don't matter */
+	virtual bool IsInsideBlock(const Vector3d & a_Position, const BLOCKTYPE a_BlockType, const NIBBLETYPE a_BlockMeta);
+
 	/** Called when one of the neighbors gets set; equivalent to MC block update.
 	By default drops if position no more suitable (CanBeAt(), DoesDropOnUnsuitable(), Drop()),
 	and wakes up all simulators on the block. */
--- a/src/Blocks/BlockSnow.h
+++ b/src/Blocks/BlockSnow.h
@@ -88,6 +88,11 @@ public:
 		UNUSED(a_Meta);
 		return 14;
 	}
+
+	virtual bool IsInsideBlock(const Vector3d & a_Position, const BLOCKTYPE a_BlockType, const NIBBLETYPE a_BlockMeta) override
+	{
+		return a_Position.y < (cBlockInfo::GetBlockHeight(a_BlockType) * (a_BlockMeta & 7));
+	}
 } ;
 
 
--- a/src/Blocks/BlockStairs.h
+++ b/src/Blocks/BlockStairs.h
@@ -115,6 +115,38 @@ public:
 			}
 		}
 	}
+
+	/** EXCEPTION a.k.a. why is this removed:
+	This collision-detection is actually more accurate than the client, but since the client itself
+	sends inaccurate / sparse data, it's easier to just err on the side of the client and keep the
+	two in sync by assuming that if a player hit ANY of the stair's bounding cube, it counts as the ground. */
+	#if 0
+	bool IsInsideBlock(const Vector3d & a_Position, const BLOCKTYPE a_BlockType, const NIBBLETYPE a_BlockMeta)
+	{
+		if (a_BlockMeta & 0x4)  // upside down
+		{
+			return true;
+		}
+		else if ((a_BlockMeta & 0x3) == 0)  // tall side is east (+X)
+		{
+			return a_Position.y < ((a_Position.x > 0.5) ? 1.0 : 0.5);
+		}
+		else if ((a_BlockMeta & 0x3) == 1)  // tall side is west (-X)
+		{
+			return a_Position.y < ((a_Position.x < 0.5) ? 1.0 : 0.5);
+		}
+		else if ((a_BlockMeta & 0x3) == 2)  // tall side is south (+Z)
+		{
+			return a_Position.y < ((a_Position.z > 0.5) ? 1.0 : 0.5);
+		}
+		else if ((a_BlockMeta & 0x3) == 3)  // tall side is north (-Z)
+		{
+			return a_Position.y < ((a_Position.z < 0.5) ? 1.0 : 0.5);
+		}
+		return false;
+	}
+	#endif
+
 } ;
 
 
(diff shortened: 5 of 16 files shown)


src/Blocks/BlockEnchantmentTable.h
```diff
@@ -34,6 +34,11 @@ class cBlockEnchantmentTableHandler :
 		UNUSED(a_Meta);
 		return 29;
 	}
+
+	virtual double GetBlockHeight(void) override
+	{
+		return 12.0 / 16.0;
```
SphinxC0re: So every time, you use this method, this is being calculated instead of putting this into a constant. Generally what about making a constant list of all the block heights?

src/Entities/Player.cpp
```diff
@@ -476,41 +476,45 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
...
+		{ 0, 0 },
+		{ 1, 0 },
+		{ -1, 0 },
+		{ 0, 1 },
+		{ 0, -1 },
+	} ;
+
+	/* Here's the rough outline of how this mechanism works:
+		We take the player's pointlike position (sole of feet), and expand it into a crosslike shape.
+		If any of the five points hit a block, we consider the player to be "on" (or "in") the ground. */
```
SphinxC0re: bad formatting

src/Entities/Player.cpp
```diff
@@ -476,41 +476,45 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
...
-					(GetWorld()->GetBlock((GetPosition() + Vector3d(0, 0, 0)).Floor()) == E_BLOCK_SNOW) ||
-					(GetWorld()->GetBlock((GetPosition() + Vector3d(HalfWidth, 0, 0)).Floor()) == E_BLOCK_SNOW) ||
-					(GetWorld()->GetBlock((GetPosition() + Vector3d(-HalfWidth, 0, 0)).Floor()) == E_BLOCK_SNOW) ||
-					(GetWorld()->GetBlock((GetPosition() + Vector3d(0, 0, HalfWidth)).Floor()) == E_BLOCK_SNOW) ||
-					(GetWorld()->GetBlock((GetPosition() + Vector3d(0, 0, -HalfWidth)).Floor()) == E_BLOCK_SNOW)
-				)
-			)
-		)
-	)
+	static const struct
```
SphinxC0re: I would like a little bit more documentation on this

src/Blocks/BlockEnchantmentTable.h
Gargaj: Most compilers evaluate static constants during compile time.

src/Entities/Player.cpp
Gargaj: Which part?

src/Entities/Player.cpp
```diff
@@ -476,41 +476,45 @@ void cPlayer::SetTouchGround(bool a_bTouchGround)
...
+
+	/* Here's the rough outline of how this mechanism works:
+		We take the player's pointlike position (sole of feet), and expand it into a crosslike shape.
+		If any of the five points hit a block, we consider the player to be "on" (or "in") the ground. */
+	bool OnGround = false;
+	for (int i = 0; i < ARRAYCOUNT(gCoords); i++)
+	{
+		/* We calculate from the player's position, one of the cross-offsets above, and we move it down slightly so it's beyond inaccuracy.
+			The added advantage of this method is that if the player is simply standing on the floor,
+			the point will move into the next block, and the floor() will retrieve that instead of air. */
```
SphinxC0re: bad formatting

src/Entities/Player.cpp
Gargaj: I copied it straight from https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L634

src/Blocks/BlockCake.h
```diff
@@ -49,6 +49,11 @@ class cBlockCakeHandler :
 		UNUSED(a_Meta);
 		return 14;
 	}
+
+	virtual double GetBlockHeight(void) override
+	{
+		return 8.0 / 16.0;
```
SphinxC0re: this is basically 0.5

src/Blocks/BlockEnchantmentTable.h
SphinxC0re: *most* compilers. Because we cannot be sure of this and because of general structure I would like to have a constant list of all block heights

src/Entities/Player.cpp
SphinxC0re: I mean the docs

src/Entities/Player.cpp
SphinxC0re: the struct

src/Blocks/BlockStairs.h
```diff
@@ -115,6 +115,43 @@ class cBlockStairsHandler :
...
+	virtual double GetBlockHeight(void) override
+	{
+		return 1.0;
+	}
+
+	/** EXCEPTION a.k.a. why is this removed:
+		This collision-detection is actually more accurate than the client, but since the client itself
+		sends inaccurate / sparse data, it's easier to just err on the side of the client and keep the
+		two in sync by assuming that if a player hit ANY of the stair's bounding cube, it counts as the ground. */
+#if 0
```
madmaxoft: Style:
- comments should all start at the same indent level
- preprocessor directives should be indented the same as the code around them.

src/Blocks/BlockSnow.h
```diff
@@ -88,6 +88,16 @@ class cBlockSnowHandler :
...
 	}
+
+	virtual double GetBlockHeight(void) override
+	{
+		return 1.0 / 8.0;  // one layer of snow
+	}
+
+	virtual bool IsInsideBlock(const Vector3d & a_Position, const NIBBLETYPE & a_BlockMeta) override
+	{
+		return a_Position.y < (GetBlockHeight() * (a_BlockMeta & 7));
```
madmaxoft: Style: Add parenthesis around the comparison

src/Blocks/BlockEnchantmentTable.h
madmaxoft: I don't think we need to worry about compilers that don't optimize this away, because such compilers won't optimize anything at all and the performance would be terrible anyway. All modern compilers (and we use C++11, so we need a modern compiler) do the optimization.

src/Blocks/BlockHandler.cpp
```diff
@@ -549,6 +549,18 @@ bool cBlockHandler::DoesDropOnUnsuitable(void)
 	return true;
 }
 
+/* default functionality: assume all blocks are full 1 * 1 * 1 voxels */
+double cBlockHandler::GetBlockHeight(void)
+{
+	return 1.0;
+}
+
+/* default functionality: only test for height, since we assume full voxels with varying height */
```
madmaxoft: Style: 5 empty lines between function bodies in a cpp file.

src/Blocks/BlockHandler.h
```diff
@@ -123,7 +123,14 @@ class cBlockHandler
...
 	Default: true */
 	virtual bool DoesDropOnUnsuitable(void);
-	
+
+	/** Returns the block's height */
+	virtual double GetBlockHeight(void);
+
+	/** Tests if a_Position is inside the block where a_Position is always [0..1[
+	Note that this is considered from a "top-down" perspective i.e. a top-half slab will always return true */
+	virtual bool IsInsideBlock(const Vector3d & a_Position, const NIBBLETYPE & a_BlockMeta);
```
madmaxoft: Why a reference to blockmeta? it's a simple 8-bit value, it's better to pass it by-value.

madmaxoft: Also should be noted that `a_Position` is assumed to be within the block's grid boundaries.

src/Blocks/BlockSnow.h
```diff
@@ -88,6 +88,16 @@ class cBlockSnowHandler :
 		UNUSED(a_Meta);
 		return 14;
 	}
+
+	virtual double GetBlockHeight(void) override
+	{
+		return 1.0 / 8.0;  // one layer of snow
```
madmaxoft: The function might need a `a_BlockMeta` param in order to be able to handle snow properly - multiple layers of snow have a different height.

src/Blocks/BlockHandler.h
Gargaj: Fences are unaccounted for, I admit to that much, but that can be solved by simply sampling the block one block lower, but leaving the sampling point unchanged. That'd make the range [0..1,0..2,0..1[
Notably this can basically extend to any sampled blocks in the vicinity if there are ever blocks that have a single anchor but expand to multiple physical blocks.

src/Blocks/BlockSnow.h
Gargaj: Note the ```IsInsideBlock``` overload underneath.

SphinxC0re:
Implementing blockheights is a good way of handling collision from the top of objects. but what about collision in general? Wouldn't it be good to have a kind of bounding boy for each block?

Gargaj:
...oops. I'll fix this when I get home.

src/BlockInfo.cpp
```diff
@@ -584,6 +584,27 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
 	a_Info[E_BLOCK_STONE               ].m_CanBeTerraformed = true;
 
 
+	// Block heights:
+	a_Info[E_BLOCK_BED                 ].m_BlockHeight = 9.0 / 16.0;
+	a_Info[E_BLOCK_CAKE                ].m_BlockHeight = 8.0 / 16.0;
```
SphinxC0re: this is still 0.5

src/BlockInfo.cpp
```diff
@@ -584,6 +584,27 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
...
+	a_Info[E_BLOCK_BIRCH_FENCE         ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_BIRCH_FENCE_GATE    ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_DARK_OAK_FENCE      ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_DARK_OAK_FENCE_GATE ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_FENCE               ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_OAK_FENCE_GATE      ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_JUNGLE_FENCE        ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_JUNGLE_FENCE_GATE   ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_SPRUCE_FENCE        ].m_BlockHeight = 1.5;
+	a_Info[E_BLOCK_SPRUCE_FENCE_GATE   ].m_BlockHeight = 1.5;
```
worktycho: Do these need to be doubles, or could we make them 16th's of a block?

src/BlockInfo.cpp
Gargaj: Compilers still do constantfolding and now this value is actually stored in a variable once during initialization.

src/BlockInfo.cpp
Gargaj: I'm not sure. Logic would dictate that 16ths are not an unreasonable metric either, given that Minecraft objects are usually texel-aligned, but that's not a guarantee for the hitboxes of future objects (e.g. the fence hitbox is just arbitrarily 1.5, but could just be anything from 1.3 or 1.123), and there's no real performance-gain given that the actual testing is still going to be a floating point comparison.

src/BlockInfo.cpp
SphinxC0re: Still inconsistent. Please change this

Gargaj: Now it's "consistent" and looks awful. I added comments as reminders to where the magic values supposedly come from.

src/BlockInfo.cpp
SphinxC0re: What about mods that implement their own blocks with their own specific bounding box?

worktycho: I'm happy to leave it as doubles, based on Gargaj's arguments.

madmaxoft: My vote is for floats / doubles, too. I'd prefer using fractions (`1.0 / 16.0`) to represent them, rather than decimal numbers, since fractions are somewhat easier to read in this case - most, if not all, of the numbers will be `N / 16.0`.

Gargaj: I tried to argue for readability but at this point I'm willing to represent them as hexadecimal if it speeds up the process of getting the fix integrated.

Gargaj:
I've looked at this again and I think it'd be a good idea to re-build this on top of https://github.com/cuberite/cuberite/pull/2623 once @bibo38 finishes it.

Gargaj:
Or alternatively I could try to adapt what he did and merge it into this manually?

SphinxC0re:
Yup :+1:

Gargaj:
I updated this to integrate https://github.com/cuberite/cuberite/pull/2623 + worked some more on the slimeblock handling. The actual rebounding of the slimeblock is not done properly yet because it'd need to be (partly) in ```HandlePhysics```; I also had to change a bit of the code there, to actually clamp the Y position down from e.g. 123.45 to 123.0 because currently mobs indefinitely just hover slightly above the ground.

From what I can tell this has been working well (I spent ~15 minutes pushing mobs off cliffs into various things), but it'd be nice if someone could actually try it as well.

SafwatHalaby:
I wonder how/if this affects the PathFinder.

Gargaj:
Without knowing much of the pathfinder, my assumptions is that it shouldn't at all - this only deals with how the environment reacts to the entities moving.

src/BlockInfo.cpp
SafwatHalaby: I vote for base64 encoding. And if that's too unreadible, Then [Dvorak Encoding](https://en.wikipedia.org/wiki/Dvorak_encoding) is a good compromise.

*This post is not to be taken seriously.

Gargaj:
Is anyone taking a look at this?

Issue #2548, opened 2015-10-16
Using /spawn in the Nether returns to you a flat nether.


jammet:
When using /spawn in the nether to leave it, you are not transported to the normal world. Instead of that, I end up in a place of endless bedrock in a world that looks like the nether, or at least has those colors.

madmaxoft:
That would be the Nether's bedrock roof. The spawn coordinates generated at server first startup are wrong for Nether worlds, they always point to the highest non-air non-water non-lava block, which in the Nether is always the rooftop. We need to fix the algorithm to take Nether worlds into account. However, this won't fix existing worlds, they will need manual fixing. Go to the Nether world's world.ini and change the spawn coordinates manually (you'll notice that the Y coord is >127, you want below 125).

jammet:
Ah, I see, I have adjusted the height. Thanks for the help :).

PR #2639, opened 2015-11-12
Improve world spawn location generator.

samoatesgames:
- Supports spawning under tree spawning.
 - Uses a radial search pattern, improving distribution.
 - Doesn't spawn in ocean biomes.
 - Supports spawning in the Nether.

The search first finds a valid biome.
Then starts at Y 96 and works down till it finds ground.
If it finds some it uses that as the spawn, else it searches in a radial manor about the origin.
Checks adjoining blocks.

Call InitializeSpawn after creating a new world via a portal. This means the spawn will be correctly set in Nether worlds.

Fixes https://github.com/cuberite/cuberite/issues/2548

### DIFF
--- a/src/BlockEntities/BrewingstandEntity.cpp
+++ b/src/BlockEntities/BrewingstandEntity.cpp
@@ -29,7 +29,7 @@ cBrewingstandEntity::cBrewingstandEntity(int a_BlockX, int a_BlockY, int a_Block
 	m_Contents.AddListener(*this);
 	for (int i = 0; i < 3; i++)
 	{
-		m_Results[i] = *(new cItem());
+		m_Results[i] = cItem{};
 	}
 }
 


src/World.cpp
```diff
@@ -624,48 +721,39 @@ void cWorld::GenerateRandomSpawn(void)
 
 bool cWorld::CheckPlayerSpawnPoint(int a_PosX, int a_PosY, int a_PosZ)
 {
-	// The bottom layer cannot hold a valid spawn point
-	if (a_PosY <= 0)
+	// Check height bounds
+	if ((a_PosY <= 0) || (a_PosY > 255))
```
NiLSPACE: We now have `cChunkDef::IsValidHeight(a_PosX)`

src/World.cpp
```diff
@@ -572,50 +572,147 @@ void cWorld::Start(void)
...
+		E_BLOCK_DIRT,
+		E_BLOCK_SAND,
+		E_BLOCK_SNOW,
+		E_BLOCK_SNOW_BLOCK,
+		E_BLOCK_NETHERRACK
+	};
+
+	static const int ValidSpawnBlocksCount = ARRAYCOUNT(ValidSpawnBlocks);
+
+	for (int PotentialY = 96; PotentialY > 32; --PotentialY)
```
NiLSPACE: Can't you just use `cWorld:GetHeight(X, Z)` or even better `cWorld:TryGetHeight(X, Z)`

samoatesgames: GetHeight() will return the highest block. 
The Nether has a roof, meaning you will spawn on top of that.
It also means you can spawn on top of trees (or other things which aren't really the floor), where as the manual search can be more fine grain and only allows you to spawn on 'the ground' (which is set to the array ValidSpawnBlocks, maybe this should be configurable?).

src/World.cpp
samoatesgames: I didn't have this when I wrote it, though at the time I remember thinking 'I wish there was a method for this, magic numbers suck'.

I shall change it later today.

src/World.cpp
NiLSPACE: Maybe, but as far as I know the nether doesn't have a spawn point. Sure, currently Cuberite acts as if a Nether world has a spawn point, but I believe we should step away from that. 

Also, you are checking if the block is valid, so if the block is a tree the server searches for a different spot right?

src/World.cpp
NiLSPACE: Yes, it was added a few days ago :)

src/World.cpp
samoatesgames: Technically you are correct, the Nether does not have a 'spawn', however it is possible to do /spawn in the Nether which previously put you on the Nethers roof. Meaning an Admin had to set the spawn for the Nether.

It could be confusing for players if they do /spawn and it puts you on the roof with no way down until an Admin is around.

If the block is a tree leaf it will search down under the leaves to see if there is a suitable place. If there is not a suitable place it will search in a different chunk, using a radial search about the 'origin'. The origin is simply a point in a valid biome.

The previous spawn code randomly increased x and z chunks meaning it just searched diagonally. Meaning it missed many potentially valid spawn chunks closer to 0,Y,0 (which is where I presume most people think the default spawn would be).

NiLSPACE: If a user would use /spawn he should be send back to the linked overworld, or the default world. That's something the Core should fix.

Keep in mind that the height of the nether can be changed in Cuberite. By default it's 128, but you can set it as high or low as you want.

I believe vanilla picks a "random" (based on the seed) location between something like 500 and -500 for both X and Z coordinates. From there it searches for a valid spot. Or at least that's how I always thought it worked ;)

With this there is a chance that you can spawn in a cave isn't there?

samoatesgames: Technically yes. If there was a 'ground' block there. So the only block which can cause an issue is the Dirt block as no other blocks should exist in a cave.

The problem with cave support again comes down to the Nether being a massive cave, so I had to support it to support Nether spawns.
If that is something that isn't wanted then I can add cave protection back in.

Personally a spawn point in the Nether is something I wanted on my server (thus this change :D) but I understand it might not be wanted for the project overall.

src/World.cpp
```diff
@@ -572,50 +572,147 @@ void cWorld::Start(void)
...
+		if (BodyBlock != E_BLOCK_AIR)
+		{
+			continue;
+		}
+
+		BLOCKTYPE FloorBlock = GetBlock(static_cast<int>(a_X), PotentialY - 2, static_cast<int>(a_Z));
+
+		// Is the floor block ok
+		bool ValidSpawnBlock = false;
+		for (int BlockIndex = 0; BlockIndex < ValidSpawnBlocksCount; ++BlockIndex)
```
worktycho: Clang assumes that this loop is more expensive than the CheckPlayerSpawnPoint call, and reorders the loop and call.

src/World.cpp
```diff
@@ -572,50 +572,147 @@ void cWorld::Start(void)
...
 		{
-			if (CheckPlayerSpawnPoint(SpawnX, GetHeight(SpawnX, SpawnZ), SpawnZ))
+			if (ValidSpawnBlocks[BlockIndex] == FloorBlock)
```
worktycho: Replace this if with a ternary operator which allows the compiler to unroll the loop. This loop body is going to be executed 61440 times.

samoatesgames: Would the following be better?

ValidSpawnBlock |= (ValidSpawnBlocks[BlockIndex] == FloorBlock);

worktycho: It works without the early exit condition.

worktycho:
Why close?

samoatesgames:
I was re-basing to latest and adding the suggested changes.
I'm guessing me pushing to my fork removed my changes on the repo (I still have them locally), which made this close itself? 
I'm not a git expert (we use Perforce at work) so I probably messed something up.

The changes will be back in my repo at the end of my lunch break :D

PR #2640, opened 2015-11-12
Tried mapping of nether coords

### LINKED ISSUE #2541: Map nether coordinates to overworld coordinates and reverse when using portals
> Locations in the Nether correlate to Overworld coordinates, but horizontal Overworld distances are scaled down by a ratio of 8:1 for travel in the Nether. Therefore, traveling one block in the Nether means traveling eight blocks in the Overworld.

– from http://minecraft.gamepedia.com/The_Nether

So the teleport function should take the type of world the player is currently in plus his position. Then the position gets mapped onto the destination world. After this procedure, the player gets teleported

SphinxC0re:
I don't know if this is in any way the right way to do this, but I hope that a review will do

fixes #2541

### DIFF
--- a/src/NetherPortalScanner.cpp
+++ b/src/NetherPortalScanner.cpp
@@ -20,6 +20,17 @@ cNetherPortalScanner::cNetherPortalScanner(cEntity * a_MovingEntity, cWorld * a_
 	m_Position(a_DestPosition),
 	m_MaxY(a_MaxY)
 {
+	if (m_World->GetDimension() == eDimension::dimNether)
+	{
+		m_Position.x = round(m_Position.x / 8);
+		m_Position.z = round(m_Position.z / 8);
+	}
+	else if (m_World->GetDimension() == eDimension::dimOverworld)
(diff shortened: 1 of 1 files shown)


worktycho:
What about overworld to overworld portals? or Nether to Nether portals?

SphinxC0re:
these doesn't exist in Vanilla

SphinxC0re:
Oh :smile: I think this already happens in cEntity

worktycho:
They may not exist in vanilla, but they are possible in cuberite.

Seadragon91:
Well then I think, we could implement something into the api, for creating and linking portals in same worlds or other worlds. One point would be to allow to use portals to change the server (bungecoord)

SphinxC0re:
What I meant was that this PR is obsolete due to the fact that this is already done in cEntity. I just didn't recognize

SphinxC0re:
@Seadragon91 exporting portal linkage to the API might be a good idea though

PR #2642, opened 2015-11-13
Improved spawn location calculations.

samoatesgames:
Note: Sorry for messing up the other PR. I rebased my fork, pushed and it auto-closed the other PR :(

 - Supports Overworld and Nether spawns.
 - Supports spawning under objects, but still above ground (e.g. under the leaves of a tree).
 - Protects against spawning in oceans.
 - Protects against spawning in water.
 - Uses a radial search about the origin, rather than a linear.
 - Correctly calculates Nether spawn on spawn world generation (fixes: https://github.com/cuberite/cuberite/issues/2548)
 - Fixes a bug in CheckPlayerSpawnPoint() where the X offset was used in both the X and Z coordinates (BLOCKTYPE BlockType = GetBlock(a_PosX + Coords[i].x, a_PosY, a_PosZ + Coords[i].x);)

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -325,10 +325,17 @@ void cWorld::SetNextBlockTick(int a_BlockX, int a_BlockY, int a_BlockZ)
 
 void cWorld::InitializeSpawn(void)
 {
+	// For the debugging builds, don't make the server build too much world upon start:
+	#if defined(_DEBUG) || defined(ANDROID_NDK)
+	const int DefaultViewDist = 9;
+	#else
+	const int DefaultViewDist = 20;  // Always prepare an area 20 chunks across, no matter what the actual cClientHandle::VIEWDISTANCE is
+	#endif  // _DEBUG
+
 	if (!m_IsSpawnExplicitlySet)
 	{
-		// Spawn position wasn't already explicitly set, enerate random solid-land coordinate and then write it to the world configuration:
-		GenerateRandomSpawn();
+		// Spawn position wasn't already explicitly set, enumerate random solid-land coordinate and then write it to the world configuration:
+		GenerateRandomSpawn(DefaultViewDist);
 		cIniFile IniFile;
 		IniFile.ReadFile(m_IniFileName);
 		IniFile.SetValueF("SpawnPosition", "X", m_SpawnX);
@@ -337,20 +344,13 @@ void cWorld::InitializeSpawn(void)
 		IniFile.WriteFile(m_IniFileName);
 	}
 
-	int ChunkX = 0, ChunkZ = 0;
-	cChunkDef::BlockToChunk(FloorC(m_SpawnX), FloorC(m_SpawnZ), ChunkX, ChunkZ);
-	
-	// For the debugging builds, don't make the server build too much world upon start:
-	#if defined(_DEBUG) || defined(ANDROID_NDK)
-		const int DefaultViewDist = 9;
-	#else
-		const int DefaultViewDist = 20;  // Always prepare an area 20 chunks across, no matter what the actual cClientHandle::VIEWDISTANCE is
-	#endif  // _DEBUG
 	cIniFile IniFile;
 	IniFile.ReadFile(m_IniFileName);
 	int ViewDist = IniFile.GetValueSetI("SpawnPosition", "PregenerateDistance", DefaultViewDist);
 	IniFile.WriteFile(m_IniFileName);
 
+	int ChunkX = 0, ChunkZ = 0;
+	cChunkDef::BlockToChunk(FloorC(m_SpawnX), FloorC(m_SpawnZ), ChunkX, ChunkZ);
 	cSpawnPrepare::PrepareChunks(*this, ChunkX, ChunkZ, ViewDist);
 	
 	#ifdef TEST_LINEBLOCKTRACER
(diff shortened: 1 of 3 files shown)


src/World.cpp
```diff
@@ -572,50 +572,158 @@ void cWorld::Start(void)
...
+
+		// Early out - Is the floor block air
+		if (FloorBlock == E_BLOCK_AIR)
+		{
+			continue;
+		}
+
+		// Is the floor block ok
+		bool ValidSpawnBlock = false;
+		for (int BlockIndex = 0; BlockIndex < ValidSpawnBlocksCount && !ValidSpawnBlock; ++BlockIndex)
```
worktycho: This won't unroll unless you remove the && !ValidSpawnBlock condition

SphinxC0re:
There seems to be a style error somewhere

samoatesgames:
@SphinxC0re haha around the bit that @worktycho  suggests removing :D Two birds, One push.

src/World.cpp
```diff
@@ -325,10 +325,17 @@ void cWorld::SetNextBlockTick(int a_BlockX, int a_BlockY, int a_BlockZ)
 
 void cWorld::InitializeSpawn(void)
 {
+	// For the debugging builds, don't make the server build too much world upon start:
+#if defined(_DEBUG) || defined(ANDROID_NDK)
+	const int DefaultViewDist = 9;
+#else
+	const int DefaultViewDist = 20;  // Always prepare an area 20 chunks across, no matter what the actual cClientHandle::VIEWDISTANCE is
+#endif  // _DEBUG
```
madmaxoft: Style: Preprocessor directives should be indented like normal code

samoatesgames: Gah that's visual studio auto indenting when I moved. Damn you Microsoft!!!

src/World.cpp
```diff
@@ -572,50 +572,158 @@ void cWorld::Start(void)
...
+		E_BLOCK_GRASS,
+		E_BLOCK_SAND,
+		E_BLOCK_SNOW,
+		E_BLOCK_SNOW_BLOCK,
+		E_BLOCK_NETHERRACK
+	};
+
+	static const int ValidSpawnBlocksCount = ARRAYCOUNT(ValidSpawnBlocks);
+
+	static const int HighestSpawnPoint = std::min((cChunkDef::Height >> 1) - 1, GetHeight(static_cast<int>(a_X), static_cast<int>(a_Z) + 16));
```
madmaxoft: Prefer to use arithmetic operations instead of bit-shifts.

src/World.cpp
```diff
@@ -572,50 +572,158 @@ void cWorld::Start(void)
...
+
+	m_SpawnY = GetHeight(static_cast<int>(m_SpawnX), static_cast<int>(m_SpawnZ));
+	LOGWARNING("Did not find an acceptable spawnpoint. Generated a random spawnpoint position at {%.2f, %.2f, %.2f}", m_SpawnX, m_SpawnY, m_SpawnZ);
+}
+
+
+
+
+
+bool cWorld::CanSpawnAt(double a_X, double a_Z, double &a_Y)
```
madmaxoft: Style: Space after `&`

src/World.h
```diff
@@ -1042,7 +1042,10 @@ class cWorld :
 	void UpdateSkyDarkness(void);
 
 	/** Generates a random spawnpoint on solid land by walking chunks and finding their biomes */
-	void GenerateRandomSpawn(void);
+	void GenerateRandomSpawn(int a_MaxSpawnRadius);
+
+	/** Can the specified coordinates be used as a spawn point? */
+	bool CanSpawnAt(double a_X, double a_Z, double &a_Y);
```
madmaxoft: All our functions use X, Y, Z, this one should, too.
Document the Y being an out-param.

src/World.cpp
```diff
@@ -572,50 +572,158 @@ void cWorld::Start(void)
...
-			(biome != biOcean) && (biome != biFrozenOcean) &&  // The biome is acceptable (don't want a small ocean island)
-			!IsBlockWaterOrIce(GetBlock(SpawnX, GetHeight(SpawnX, SpawnZ), SpawnZ))  // The terrain is acceptable (don't want to spawn inside a lake / river)
-		)
+		// Found a usable biome
+		// Spawn chunks so we can find a nice spawn.
+		int ChunkX = 0, ChunkZ = 0;
+		cChunkDef::BlockToChunk(BiomeOffset.x, BiomeOffset.z, ChunkX, ChunkZ);
+		cSpawnPrepare::PrepareChunks(*this, ChunkX, ChunkZ, a_MaxSpawnRadius);
+		break;
+	} while ((--BiomeCheckCount) > 0);
```
madmaxoft: Prefer a regular `for` loop

src/World.cpp
```diff
@@ -572,50 +572,158 @@ void cWorld::Start(void)
...
+		Vector3i(HalfChunk, 0, -HalfChunk),
+		Vector3i(0, 0, -HalfChunk),
+		Vector3i(-HalfChunk, 0, -HalfChunk),
+		Vector3i(-HalfChunk, 0, 0),
+		Vector3i(-HalfChunk, 0, HalfChunk),
+	};
+
+	static const int PerRadiSearchCount = ARRAYCOUNT(ChunkOffset);
+
+	for (int RadiusOffset = 1; RadiusOffset < (a_MaxSpawnRadius << 1); ++RadiusOffset)
```
madmaxoft: Prefer arithmetic to bitshift operations

madmaxoft:
I have an issue with the new search code. It searches near chunks first, then extends further and further away. However, suppose the initial choice is in an ocean. The general assumption is that the nearby chunks will all be ocean, rather than land, so it makes sense to go as far away as possible as fast as possible. That's why the current search code actually walks away from the initial choice. I think that's a much better approach than generating lots of needless ocean chunks.

samoatesgames:
It has a check for ocean biomes and first moves out of them.
This means the search origin is in a valid biome, searching around that is less likely to move into another invalid biome, where as the linear search could easily move into a different ocean biome.

SphinxC0re:
But what about custom generators? Do they have to bring their own spawngenerator?

samoatesgames:
The case of custom generators would have to be supported by having a config option for invalid spawn biomes.

The current linear search would still have issues with custom generators.

madmaxoft:
Ah, I've missed that. It really does try to get out of ocean fast.

@SphinxC0re custom generators such as in plugin-provided generators? Those haven't worked with the old code either and they don't work with the new code, but there's currently no real way to get them to cooperate. We'd need to break the generator-to-plugin bindings into a bit smaller facets in order to achieve that.

samoatesgames:
All style issues should have been corrected.

samoatesgames:
Are there any other improvements people want made?
Or are people generally not happy with the algorithm used?
:)

SphinxC0re:
Thank you so much for the change @samoatesgames

PR #2645, opened 2015-11-14
Allow multiple followable items

Gargaj:
This will be needed for breeding (see rabbits), especially with the upcoming 1.9 changes.

### DIFF
--- a/src/Item.h
+++ b/src/Item.h
@@ -239,6 +239,8 @@ public:
 	void    Clear (void) {clear(); }
 	size_t  Size  (void) const { return size(); }
 	void    Set   (int a_Idx, short a_ItemType, char a_ItemCount, short a_ItemDamage);
+	bool    Contains(const cItem & a_Item);
+	bool    ContainsType(const cItem & a_Item);
 
 	void    Add   (short a_ItemType, char a_ItemCount, short a_ItemDamage)
 	{
--- a/src/Mobs/Cow.h
+++ b/src/Mobs/Cow.h
@@ -20,7 +20,10 @@ public:
 	virtual void GetDrops(cItems & a_Drops, cEntity * a_Killer = nullptr) override;
 	virtual void OnRightClicked(cPlayer & a_Player) override;
 
-	virtual const cItem GetFollowedItem(void) const override { return cItem(E_ITEM_WHEAT); }
+	virtual void GetFollowedItems(cItems & a_Items) override
+	{
+		a_Items.Add(E_ITEM_WHEAT);
+	}
 
 } ;
 
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -1550,6 +1550,8 @@ end
 					{ Params = "Index, ItemType, ItemCount, ItemDamage", Return = "", Notes = "Sets the item at the specified index to the specified item" },
 				},
 				Size = { Params = "", Return = "number", Notes = "Returns the number of items in the collection" },
+				Contains = { Params = "{{cItem|cItem}}", Return = "bool", Notes = "Returns true if the collection contains an item that is fully equivalent to the parameter" },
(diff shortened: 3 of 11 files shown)


src/Item.h
```diff
@@ -239,6 +239,28 @@ class cItems  // tolua_export
...
+	{
+		for (const_iterator itr = begin(); itr != end(); ++itr)
+		{
+			if (a_Item.IsSameType(*itr))
+			{
+				return true;
+			}
+		}
+		return false;
+	}
```
bearbin: There should be newlines around these functions.

bearbin:
This looks like a good change to me, although I probably missed a few minor issues. :)

src/Mobs/Cow.h
```diff
@@ -20,7 +20,10 @@ class cCow :
 	virtual void GetDrops(cItems & a_Drops, cEntity * a_Killer = nullptr) override;
 	virtual void OnRightClicked(cPlayer & a_Player) override;
 
-	virtual const cItem GetFollowedItem(void) const override { return cItem(E_ITEM_WHEAT); }
+	virtual void GetFollowedItems(cItems & a_Items) override
```
NiLSPACE: Why not return a `cItems` instead?

Gargaj: I haven't seen examples of it in the source, so I just followed what appeared to be the convention.

NiLSPACE: Something like this:
```cpp
virtual cItems GetFollowedItems() override
{
	cItems FollowedItems;
	FollowedItems.Add(E_ITEM_MY_ITEM1);
	FollowedItems.Add(E_ITEM_MY_ITEM2);
	FollowedItems.Add(E_ITEM_MY_ITEM3);
	return FollowedItems;
}
```

But you might want to wait with it until more people agree.

src/Item.h
```diff
@@ -240,6 +240,30 @@ class cItems  // tolua_export
 	size_t  Size  (void) const { return size(); }
 	void    Set   (int a_Idx, short a_ItemType, char a_ItemCount, short a_ItemDamage);
 
+	bool    Contains(const cItem & a_Item)
+	{
+		for (const_iterator itr = begin(); itr != end(); ++itr)
```
worktycho: Range for loop?

src/Item.h
```diff
@@ -240,6 +240,30 @@ class cItems  // tolua_export
...
+			{
+				return true;
+			}
+		}
+		return false;
+	}
+
+	bool    ContainsType(const cItem & a_Item)
+	{
+		for (const_iterator itr = begin(); itr != end(); ++itr)
```
worktycho: Range for loop?

src/Mobs/Cow.h
Gargaj: Not saying it wouldn't work, but wouldn't it need to call the copy-constructor on exit to create a new copy of the same array every time? Gets even worse once you start having more overrides underneath it.

worktycho: It should be move-constructed, so only the array pointer needs to be copied. And we might need it is never a good argument.

Gargaj: Why should it be different than how ```ConvertToPickups``` works then?

worktycho: Now thats a better reason. And my answer is that ConvertToPickups should also work like this, but because the signature predates the adoption of c++11 in this project, it made sense to do it by reference at the time.

Seadragon91:
Please move the implementation into the cpp file. Have no other objections :)

Gargaj:
There's no indication on the convention of which functions should be in the header and which should be in source, so it's really hard to judge. :/

worktycho:
Thats because there isn't really a convention, and just depends on whos reviewing the code.

Seadragon91:
I just mentioned it because @madmaxoft wanted something like that in my brewing pull request.

Gargaj:
So are we waiting for something here...?

SafwatHalaby:
Regarding breeding: That would be great. You may want to check out the [mob todo list](http://forum.mc-server.org/showthread.php?tid=1895) for a general overview of mob status.

PR #2656, opened 2015-11-18
Added MAINTAINERS file

SphinxC0re:


bearbin:
What is the point of this?

SphinxC0re:
I though it would be useful. Other projects also maintain such a file. Why do you think it wouldn't?

SphinxC0re:
I'm adding all the core team members while we should discuss about the use of this

bearbin:
I guess I don't see the point of this file any more than the existing CONTRIBUTORS file. We have a group email support@cuberite.org, and the core team is listed in the github org.

SphinxC0re:
Okay :disappointed:

SafwatHalaby:
Can I be added to the group mail?

SphinxC0re:
@madmaxoft could do that. Also what about a policy to set the support mail address as cc when answering a support mail?

bearbin:
I don't think the group mail is working at the moment. But that would be a good policy, so we can know what the others are saying.

SafwatHalaby:
Only 4 people are publicly listed as the Cuberite team at Github. Check out https://github.com/orgs/cuberite/people without logging in. The unlisted ones are all those who have a "collaborator" status, and those who have "owner" status but have their privacy setting set to "private".

PR #2661, opened 2015-11-19
Musl fixes

tboerger:
With these fixes we are able to build cuberite on Alpine linux and musl libc.


worktycho:
The build failure looks like an unrelated issue.

tboerger:
Yes, now it's green

PR #2668, opened 2015-11-20
Decoupled cMonster and path recalc logic, re-implemented recalc.

### LINKED ISSUE #2536: Find a better path recalc algorithm
@tigerw thinks the current recalc algorithm is needlessly complex. This is probably true.

I should find a simpler algorithm, but that simplicitly shouldn't come at the cost of performance/flexibility (so no 1-block-recalc).

References:
[Tiger's PR discussion](https://github.com/cuberite/cuberite/pull/2446)

[Recalc PR diff discussion](https://github.com/cuberite/cuberite/pull/2523#discussion-diff-41150490)

SafwatHalaby:
~~Do not merge yet. This needs review. This is in pretty good shape, but not yet complete.~~

Fixes #2536

Known issues:
 - **Fixed**, Mobs sometimes do not detect that they've reached their final destination. The result is that passive mobs sometimes just sit still. 
 - **Fixed**, Needs to conform to the lua style check script.
 - **Fixed**, It appears I've reverted some commits made by others,  need to fix that.
 - **Fixed**, Need to review and update any out-of-date AI-related doxy docs in the headers.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -14,7 +14,7 @@
 #include "../Chunk.h"
 #include "../FastRandom.h"
 
-#include "Path.h"
+#include "PathFinder.h"
 
 
 
@@ -75,11 +75,8 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	, m_EMState(IDLE)
 	, m_EMPersonality(AGGRESSIVE)
 	, m_Target(nullptr)
-	, m_Path(nullptr)
-	, m_IsFollowingPath(false)
+	, m_PathFinder(a_Width, a_Height)
 	, m_PathfinderActivated(false)
-	, m_GiveUpCounter(0)
-	, m_TicksSinceLastPathReset(1000)
 	, m_LastGroundHeight(POSY_TOINT)
 	, m_JumpCoolDown(0)
 	, m_IdleInterval(0)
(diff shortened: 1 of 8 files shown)


src/Mobs/Monster.cpp
```diff
@@ -123,127 +120,20 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
 
-void cMonster::MoveToWayPoint(cChunk & a_Chunk)
-{
 	if (m_JumpCoolDown == 0)
 	{
 		if (DoesPosYRequireJump(FloorC(m_NextWayPointPosition.y)))
 		{
 			if (
-				(IsOnGround() && (GetSpeedX() == 0.0f) && (GetSpeedY() == 0.0f)) ||
-				(IsSwimming() && (m_GiveUpCounter < 15))
```
SafwatHalaby: I don't remember touching these two lines.

src/Mobs/Monster.h
```diff
@@ -117,8 +112,8 @@ class cMonster :
 
 	// tolua_begin
 	bool IsBaby (void) const { return m_Age < 0; }
-	int GetAge (void) const { return m_Age; }
-	void SetAge(int a_Age)  { m_Age = a_Age; }
```
SafwatHalaby: Umm, I think I've overridden someone's commit/s somehow. Did I copy-paste a whole file at some point?

SafwatHalaby:
Clang error:
```
     conversion loses integer precision: 'unsigned long' to 'int'

      [-Werror,-Wshorten-64-to-32]

                return m_PathPoints.size() - m_CurrentPoint;

                ~~~~~~ ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
```

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,237 @@
...
+	m_Path(),
+	m_GiveUpCounter(0),
+	m_NotFoundCooldown(0)
+{
+	m_Width = a_MobWidth;
+	m_Height = a_MobHeight;
+}
+
+// TODO fix monster heads
+// TODO why does PATH_NOT_FOUND even occur?
```
SafwatHalaby: Irrelevant comments, already fixed in this commit

tigerw:
D'you know what? Why don't you just merge yourself after you've finished your own review? :P

src/Mobs/Monster.cpp
madmaxoft: This looks like the fix to avoid taking fall damage while swimming in water

src/Mobs/Monster.cpp
```diff
@@ -123,127 +120,20 @@ void cMonster::SpawnOn(cClientHandle & a_Client)
...
-		return false;
-	}
-	if (m_TicksSinceLastPathReset < 1000)
-	{
-		// No need to count beyond 1000. 1000 is arbitary here.
-		++m_TicksSinceLastPathReset;
-	}
-
-	if (ReachedFinalDestination())
+	if ((m_NextWayPointPosition - GetPosition()).Length() < GetWidth()/2)
```
NiLSPACE: Space around the `/`

Also, add extra parentheses to conditions. `(GetWidth() / 2)`

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,237 @@
...
+		return ePathFinderStatus::PATH_FOUND;
+	}
+
+	// If m_Path has not been initialized yet, initialize it.
+	if (!m_Path.IsValid())
+	{
+		ResetPathFinding(a_Chunk);
+	}
+
+	switch (m_Path.Step(a_Chunk))
```
NiLSPACE: I believe some people were not happy with the naming of `Step`. Perhaps this is the time to get a proper name for it?

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if (InWater)
+	{
+		m_FinalDestination.y -= 1;
+	}
+
+
+	return true;
+}
+
```
NiLSPACE: Enters after methods/functions.

src/Mobs/Monster.cpp
madmaxoft: Nah, the extra parenthesis is not really needed.

src/Mobs/PathFinder.cpp
SphinxC0re: "Empty lines"

src/Mobs/PathFinder.cpp
SphinxC0re: What about **find* instead?

src/Mobs/PathFinder.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+#include "Globals.h"
+#include "PathFinder.h"
+#include "../Chunk.h"
+
```
SphinxC0re: add 5 empty lines between the includes and the rest of the code

src/Mobs/Monster.cpp
SphinxC0re: Why not? The convention says that it needs to be like this

madmaxoft: The convention says "parenthesis around comparisons", not "parenthesis around comparison operands"
Similarly, we write `if (a > b + 2)` instead of `if (a > (b + 2))`

SphinxC0re: But isn't that bad? It looks awful

src/Mobs/PathFinder.cpp
SafwatHalaby: Suggestions are welcome. I can't find a better name.

SphinxC0re: What does it do?

SafwatHalaby: To prevent blocking for hard path calculations, the calculation is split into ticks. Each tick does a portion of the calculation. However, the word `tick` can be confused with Minecraft ticks, so I called it `step`.

Each call calculates a small part of the path, until the calculation is done.

SphinxC0re: ```CalculationStep``` ?

SafwatHalaby:
It took me a while to realize that I'm actually using `PathDestination` for two distinct purposes. I split that into two variables and that solved most issues.

src/Mobs/PathFinder.cpp
SafwatHalaby: Is that less vague than step?

src/Mobs/Monster.cpp
SafwatHalaby: The lua script missed this one.

src/Mobs/Monster.cpp
SafwatHalaby: TODO investigate this

src/Mobs/PathFinder.cpp
SafwatHalaby: Fixed.

src/Mobs/Monster.cpp
SafwatHalaby: Monster.cpp no longer has a `GiveUpCounter`, this is why I edited this line. It breaks jumping in water, but AI water handling is pretty broken altogether, so this isn't an issue. We'll get back to this when water handling is fixed.

SafwatHalaby:
Do not merge yet, even if CI succeeds.

src/Mobs/PathFinder.cpp
SphinxC0re: I guess so

SafwatHalaby:
Rebased.
Looking good, but I'm still not satisfied with idle mob behavior. I'll need to rethink a bit.

SafwatHalaby:
The last commit fixes idle mob issues.
I changed the `PathFinder.h` header a bit.
I Added a `dontCare` bool. If set to true, it means the monster doesn't care where it would end up going, and the PathFinder  may change the monster's final destination.

SafwatHalaby:
Also, I used a pointer because it explicitly indicates that the Pathfinder may modify the finalDestination. WIth references, I found myself putting comments all over the place to prevent confusion.

SafwatHalaby:
lgtm

src/Mobs/PathFinder.h
```diff
@@ -0,0 +1,96 @@
...
+
+	/** Updates the PathFinder's internal state and returns a waypoint.
+	A waypoint is a coordinate which the mob can safely move to from its current position in a straight line.
+	The mob is expected to call this function tick as long as it is following a path.
+	@param a_Chunk The chunk in which the mob is currently at.
+	@param a_Source The mob's position. a_Source's coordinates are expected to be within the chunk given in a_Chunk.
+	@param a_Destination The position the mob would like to reach. If a_ExactPath is true, the PathFinder may modify this.
+	@param a_OutputWaypoint An output parameter: The next waypoint to go to.
+	@param a_DontCare If true, the mob doesn't care where to go, and the Pathfinder may modify a_Destination.
+	This should usually be true. An exception is a wandering idle mob which doesn't care about its final destination.
```
SafwatHalaby: This should usually be **false**.

Seadragon91:
There are still a few TODOs. Are they treated in a other pull request, someday?

SafwatHalaby:
@Seadragon91 Which TODOs are you referring to?

SafwatHalaby:
@Seadragon91 
Monster give up timer - No longer relevant
monster.h TODO - Done
goto TODO - Will do later

lgtm.

PR #2670, opened 2015-11-20
Added OrePockets and DirtPockets finish gens.

madmaxoft:
These generate a more vanilla-like pockets of ores and dirt / gravel / granite / ...

TODO:
 - [x] add a `cFinishGenOrePockets` generator, reuse base code with `cStructGenOreNests`
 - [x] use the new generator class for actual world generators - `OrePockets` and `DirtPockets`
 - [x] exact ore pocket and seed configuration in world.ini


madmaxoft:
(I needed a break from PieceStructures :)

SphinxC0re:
I like this one a lot. Thx :+1:

src/Generating/FinishGen.cpp
```diff
@@ -1395,3 +1395,369 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
...
+
+
+
+
+const cFinishGenOres::OreInfos & cFinishGenOres::DefaultOverworldOres(void)
+{
+	static OreInfos res
+	{
+		// OreType,            OreMeta, MaxHeight, NumNests, NestSize
+		{E_BLOCK_COAL_ORE,     0,       127,       20,       16},
```
SphinxC0re: these should all be configurable since vanilla allows it to be configurable

src/Generating/FinishGen.cpp
```diff
@@ -1395,3 +1395,369 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
...
+				int BlockX = BaseX + x;
+				if ((BlockX < 0) || (BlockX >= cChunkDef::Width))
+				{
+					Num++;  // So that the cycle finishes even if the base coords wander away from the chunk
+					continue;
+				}
+				for (int y = ysize; y >= 0; --y)
+				{
+					int BlockY = BaseY + y;
+					if ((BlockY < 0) || (BlockY >= cChunkDef::Height))
```
worktycho: !IsValidHeight?

src/Generating/FinishGen.cpp
```diff
@@ -1395,3 +1395,369 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
...
+			// Put a cuboid around [BaseX, BaseY, BaseZ]
+			int rnd = m_Noise.IntNoise3DInt(chunkX + 64 * i, 2 * a_Seq + Num, chunkZ + 32 * i) / 8;
+			int xsize = rnd % 2;
+			int ysize = (rnd / 4) % 2;
+			int zsize = (rnd / 16) % 2;
+			rnd >>= 8;
+			for (int x = xsize; x >= 0; --x)
+			{
+				int BlockX = BaseX + x;
+				if ((BlockX < 0) || (BlockX >= cChunkDef::Width))
```
worktycho: Should we create an IsValidWidth function to match IsValidHeight?

src/Generating/FinishGen.cpp
```diff
@@ -1395,3 +1395,369 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
...
+					{
+						int BlockZ = BaseZ + z;
+						if ((BlockZ < 0) || (BlockZ >= cChunkDef::Width))
+						{
+							Num++;  // So that the cycle finishes even if the base coords wander away from the chunk
+							continue;
+						}
+
+						int Index = cChunkDef::MakeIndexNoCheck(BlockX, BlockY, BlockZ);
+						if (cBlockInfo::CanBeTerraformed(blockTypes[Index]))
```
worktycho: We need a way of hoisting the array init check in cBlockInfo::Get out of the loop. If you split the array fetch and  block lookup in cBlockInfo::Get into separate functions, then we should be able to hoist the lookup.

src/Generating/FinishGen.cpp
```diff
@@ -1395,3 +1395,369 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
...
+				{
+					// outside the sphere, bail out
+					continue;
+				}
+				int bX = x - baseX;
+				int bZ = z - baseZ;
+				auto blockType = a_ChunkDesc.GetBlockType(bX, y, bZ);
+				if ((blockType == E_BLOCK_STONE) || (blockType == E_BLOCK_NETHERRACK))
+				{
+					a_ChunkDesc.SetBlockTypeMeta(bX, y, bZ, a_OreType, a_OreMeta);
```
worktycho: Is it useful to be able to inline these methods, given they are called in a deeply nested loop?

src/Generating/FinishGen.cpp
SphinxC0re: I thought about implementing that a few days ago but I didn't find any occurrence of code where such a check would be useful

src/Generating/FinishGen.cpp
madmaxoft: Yeah, I did this on a branch prior to that function. I'll change that.

src/Generating/FinishGen.cpp
madmaxoft: I expect the compiler to inline them

src/Generating/FinishGen.cpp
madmaxoft: These are the defaults, the configurability will come in the parent class (not implemented yet) - it will parse a string and replace these defaults with whatever it parses. For the initial version, though, I'll skip :)

SphinxC0re: Okay

src/Generating/FinishGen.cpp
worktycho: It can't because they aren't inline methods.

madmaxoft: The compiler is free to inline *any* function call that it deems inlinable. These should qualify perfectly.

worktycho: Only if it can access the code. If you're not doing LTO, because the state of LTO on *nix is abysmal, then any function not in the same translation unit, can't be inlined. cChunkDesc::GelBlockType is in a different translation unit from FinishGen.cpp

madmaxoft: That makes sense. I wouldn't want to do it in this PR, though.

src/Generating/FinishGen.cpp
madmaxoft: Since this doesn't work too well (ores overwrite even dirt and grass), I'm considering rewriting this check to actually check only for stone and netherrack (just like the `cFinishGenOrePockets` does). Since this PR already changes the ore generation in the old generators a bit, I don't think this is as a big problem as a generic generator change.
@NiLSPACE what do you think?

NiLSPACE: It does make sense to do that. I wouldn't mind a little difference in world generation. I doubt anyone would actually notice. It's not like you get massive chunk borders. like in the old days.

src/Generating/FinishGen.cpp
SphinxC0re: Please use ```cChunkDef::IsValidWidth()```

madmaxoft:
I think this is now ready for a merge, unless someone disagrees?

SphinxC0re:
LGTM. merge if CI succeeds

NiLSPACE:
Should the default world.ini parameters be updated, or does it work the same?

madmaxoft:
This generator isn't used by default, I'd like to give it some time before deciding whether it's worthy of being the default.

PR #2671, opened 2015-11-21
compile.sh - Automatically choose master branch

SafwatHalaby:
Since the testing branch was [removed](http://forum.mc-server.org/showthread.php?tid=434&pid=23415#pid23415) by @worktycho , `compile.sh` was updated.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -84,6 +84,13 @@ if [[ $MISSING_PROGRAMS != "" ]]; then
 	missingDepsExit
 fi
 
+# Bypass Branch choice and choose master. Because it's the only branch right now.
+BRANCH="master"
+
+### Inactive code start. ###
+function inactiveCode
+{
+
 # Echo: Branch choice.
 echo
 echo "You can choose between 2 branches:"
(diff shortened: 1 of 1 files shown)


compile.sh
```diff
@@ -84,6 +84,13 @@ if [[ $MISSING_PROGRAMS != "" ]]; then
 	missingDepsExit
 fi
 
+# Bypass Branch choice and choose master. Because it's the only branch right now.
+BRANCH="master"
+
+##Inactive code start
```
madmaxoft: I do prefer a space between the comment-sign and the comment text itself, in C++ we have it as a hard requirement, but we should do it in scripts, too

SafwatHalaby: Fixed.

madmaxoft:
Still, shouldn't we instead make a new `testing` branch currently pointing to `master`?

SafwatHalaby:
I made this as a quick workaround because currently people will encounter a broken script, which isn't nice.

I'll easily modify it if/when a new testing branch is created.

SafwatHalaby:
Squashed.

madmaxoft:
Fair enough. Merging.

PR #2672, opened 2015-11-21
Readme update - organized the installation methods

SafwatHalaby:
The "Installation" section was a bit messy. I think this is cleaner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,24 +11,32 @@ Subscribe to [the newsletter](http://newsletter.cuberite.org/subscribe.htm) for 
 
 Installation
 ------------
-Hosted Cuberite is available DIY on DigitalOcean: [![Install on DigitalOcean](http://doinstall.bearbin.net/button.svg)](http://doinstall.bearbin.net/install?url=https://github.com/cuberite/cuberite) and [Gamocosm](https://gamocosm.com) also offers Cuberite support.
 
-For Linux there is an easy installation method, just run this in your terminal:
+There are several ways to obtain Cuberite.
 
-    curl -s https://raw.githubusercontent.com/cuberite/cuberite/master/easyinstall.sh | sh
+#### Binaries
+ - The easiest method is downloading for Windows or Linux from the [Project site](http://cuberite.org/).
+ - You can use the EasyInstall script for Linux, which automatically downloads the correct binary. The script is described below.
+ - You can also obtain a binary from the [buildserver archive](http://builds.cuberite.org).
 
-For Windows, you just need to download a file and extract it:
+##### The EasyInstall script
+This Linux script will download the correct binary from the project site.
 
- - [Windows 32 bit](http://builds.cuberite.org/job/Cuberite%20Windows%20x86%20Master/lastSuccessfulBuild/artifact/Install/Cuberite.zip)
- - [Windows 64 bit](http://builds.cuberite.org/job/Cuberite%20Windows%20x64%20Master/lastSuccessfulBuild/artifact/Install/Cuberite.zip)
+    bash -c "$(wget -O - https://raw.githubusercontent.com/cuberite/cuberite/master/easyinstall.sh)"
 
-For other operating systems you need to download and compile yourself. This can be done either manually, or with this automatic script:
+#### Compiling
+ - You can compile automatically for Linux with the `compile.sh` script. The script is described below.
+ - You can also compile manually.
+
+Compiling may provide better performance performance (1.5-3x as fast) and it supports more operating systems. See the [COMPILING.md](https://github.com/cuberite/cuberite/blob/master/COMPILING.md) file for more details.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,24 +11,32 @@ Subscribe to [the newsletter](http://newsletter.cuberite.org/subscribe.htm) for
...
 
-    bash -c "$(wget -O - https://raw.githubusercontent.com/cuberite/cuberite/master/compile.sh)"
+###### The EasyInstall script
+This Linux script will download the correct binary from the project site. 
 
-There is also an archive of binary builds on the buildserver: http://builds.cuberite.org
+    curl -s https://raw.githubusercontent.com/cuberite/cuberite/master/easyinstall.sh | sh
 
-Compiling the server yourself has other benefits: you may get better performance performance (1.5-3x as fast) and it supports more operating systems. See the [COMPILING.md](https://github.com/cuberite/cuberite/blob/master/COMPILING.md) file for more details.
+###### The compile compile.sh script
```
madmaxoft: compile compile?

bearbin:
I think the easyinstall script should be at the top, not at the bottom - it's the easiest way for beginners on linux.

SafwatHalaby:
The easiest install method for beginners is to hit download at the website. But I agree easyinstall shouldn't be that far down. Is the new order better?

worktycho:
Can we also make the scripts more consistent? easyinstall uses `curl -s ... | sh bash -c` whilst compile.sh uses `bash -c $(wget -O - ...)`

Schwertspize:
I don't think we should provide a "download" button on the homepage, instead a short copy-paste line (with a button to copy to clipboard) and an url shortener would be better

madmaxoft:
Depends. Windows users are used to downloading installers. Linux guys prefer compilation scripts.

bearbin:
We should have a system that detects the most appropriate installation type for your system and displays it, and then have a link to a "other platforms" page for downloads/alternative install techniques.

Schwertspize:
right. that's why I would say move the install button (or replace it with easy install) and provide the compile script first

Schwertspize:
@bearbin because many people who want to run a server do not use screen sharing from. their server

bearbin:
@Schwertspize That's why you provide a link to the "other platforms page". Most recreational server owners use their own computer or a pre-packaged server host, and we should optimise for them. "power-users" can be trusted to click a link :P

@SafwatHalaby The new order is better.

SafwatHalaby:
Agree with @bearbin 
@worktycho Done. I picked `wget` because I think it's more widespread.

README.md
```diff
@@ -11,24 +11,32 @@ Subscribe to [the newsletter](http://newsletter.cuberite.org/subscribe.htm) for
...
+ - You can use the EasyInstall script for Linux, which automatically downloads the correct binary. The script is described below.
+ - You can also obtain a binary from the [buildserver archive](http://builds.cuberite.org).
 
-For Windows, you just need to download a file and extract it:
+###### The EasyInstall script
+This Linux script will download the correct binary from the project site.
 
- - [Windows 32 bit](http://builds.cuberite.org/job/Cuberite%20Windows%20x86%20Master/lastSuccessfulBuild/artifact/Install/Cuberite.zip)
- - [Windows 64 bit](http://builds.cuberite.org/job/Cuberite%20Windows%20x64%20Master/lastSuccessfulBuild/artifact/Install/Cuberite.zip)
+    bash -c "$(wget -O - https://raw.githubusercontent.com/cuberite/cuberite/master/easyinstall.sh)"
```
bearbin: It would be better if this is:

    sh -c "$(wget -O - https://raw.githubusercontent.com/cuberite/cuberite/master/easyinstall.sh)"

Not all systems have bash, and the easyinstall script works on systems that use other shells.

SafwatHalaby:
Squashed.

@bearbin: Are you sure that the script is fully sh-compatible?

SafwatHalaby:
I just tested this. The script is not sh-compatible. Bash is an `sh` superset and the script uses some bash extensions. I'll address this in a seperate PR and make the script `sh` compatible.

worktycho:
@bearbin We should not do automatic platform detection on the downloads page. Doing it from javascript is a real mess.

SafwatHalaby:
lgtm

bearbin:
@worktycho There are libraries for platform direction, which make it quite simple.

README.md
SafwatHalaby: Fixed.

worktycho:
If you can tell me one that will detect Linux with JavaScript, I would be grateful, because the lib I'm using at the moment for a uni project has a habit of spitting out distro names, rather than saying linux. Thats just my experience though.

SafwatHalaby:
Regarding sh and bash, see #2673

PR #2673, opened 2015-11-21
compile.sh is now sh-compatible. (Unix, etc.)

SafwatHalaby:


bearbin:
Is this tested to work? If it is, LGTM.

PR #2686, opened 2015-11-24
Fix apple unused define bug

### LINKED ISSUE #2684: Can't compile latest master on OSX
```
Scanning dependencies of target OSSupport
[ 45%] Building CXX object src/OSSupport/CMakeFiles/OSSupport.dir/StackTrace.cpp.o
/Users/dave/dev/cuberite/src/OSSupport/StackTrace.cpp:21:10: error: macro is not used [-Werror,-Wunused-macros]
        #define btsize int
                ^
1 error generated.
make[2]: *** [src/OSSupport/CMakeFiles/OSSupport.dir/StackTrace.cpp.o] Error 1
make[1]: *** [src/OSSupport/CMakeFiles/OSSupport.dir/all] Error 2
make: *** [all] Error 2
```

Looks like some of the fixes for MUSL may have caused this :panda_face: 
https://github.com/cuberite/cuberite/pull/2661

worktycho:
Fixes #2684

### DIFF
--- a/src/OSSupport/StackTrace.cpp
+++ b/src/OSSupport/StackTrace.cpp
@@ -14,11 +14,13 @@
 	#include <unistd.h>
 #endif
 
-// FreeBSD uses size_t for the return type of backtrace()
-#if defined(__FreeBSD__) && (__FreeBSD__ >= 10)
-	#define btsize size_t
-#else
-	#define btsize int
+#ifdef __GLIBC__
+	// FreeBSD uses size_t for the return type of backtrace()
+	#if defined(__FreeBSD__) && (__FreeBSD__ >= 10)
+		#define btsize size_t
+	#else
+		#define btsize int
+	#endif
 #endif
 
 


SphinxC0re:
LGTM @cuberite/core-team ?

src/OSSupport/StackTrace.cpp
```diff
@@ -14,17 +14,17 @@
...
+		#define btsize size_t
+	#else
+		#define btsize int
+	#endif
 #endif
 
 
 
-
-
```
madmaxoft: Why remove the empty lines?

worktycho: Acident.

dave-tucker:
Would it be possible to get [Multiple OS Support](https://docs.travis-ci.com/user/multi-os) enabled on Travis? IIRC, the repo owner needs to drop an email over to travis CI support asking for it. This way, it'd be easy to include OSX in the build matrix and prevent issues (like the one fixed by this bug) occurring again in future!

madmaxoft:
I think a better solution would be to get rid of those things altogether.

Ref.: http://forum.mc-server.org/showthread.php?tid=2230&pid=23674#pid23674

PR #2687, opened 2015-11-24
Fix escaped characters in signs

Gargaj:
Use JSON parser instead.


SphinxC0re:
Does this actually work? Did you test it? Could someone else test this?

Gargaj:
Sure it works, per http://wiki.vg/Protocol#Update_Sign_2 the protocol sends 4 JSON strings as the 4 lines of text but the strings may contain escaped characters which the previous implementation simply ignores. (See screenshots in the tagged issues.)

One thing I'm not sure about is whether the JSON data sent is always strings, or if it's sometimes a complete chat message.

PR #2693, opened 2015-11-27
Normalized OrePockets generator config names.

madmaxoft:
All INI values use AllInitialCapsStyle, no underscores.


worktycho:
What about the old names? They are in master, so you need to migrate them.

madmaxoft:
The generator is so new that I don't think it's necessary.

worktycho:
They made it to master.

madmaxoft:
Anyone who was knowledgable enough to activate and reconfigure the generator at this point is also knowledgable enough to change the parameter names. I don't think the few testers who actually did it are worth the extra code just for the migration.

This seems like an opinion-vs-opinion battle, would anyone else chime in with theirs?

NiLSPACE:
I don't think it's going to be a problem this soon after the merge. I agree with @madmaxoft on this one.

PR #2694, opened 2015-11-27
Update COMPILING.md

SafwatHalaby:
- ~~Replaced `apt-get clang` with `apt-get gcc`, because beginners are better off with GCC, because it's more widely used in Linux and is more likely to be installed already. Advanced users know what they're doing.~~
 - Copied over the build flag list from the book.
 - Several minor changes.

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -27,7 +27,7 @@ To contribute your changes to the source back to the repository, you need a Git 
 
 Alternatively, if you want only to compile the source, without contributing, you can [download the sources in a ZIP file directly from GitHub](https://github.com/cuberite/cuberite/archive/master.zip).
 
-If you're using Git to get the source, use the following set of commands to set up the local workspace correctly:
+If you're using Git to get the source, use the following command to set up the local workspace correctly:
 
 ```
 git clone --recursive https://github.com/cuberite/cuberite.git
@@ -87,9 +87,17 @@ git clone --recursive https://github.com/cuberite/cuberite.git
 
 ### Building ###
 
-Follow the instructions at [CMake on Unix-based platforms](#cmake-on-unix-based-platforms).
+Run the following commands to build Cuberite:
 
-After doing so, run the command `make` in the build directory, and Cuberite will build.
+```
+mkdir Release
+cd Release
+cmake -DCMAKE_BUILD_TYPE=RELEASE ..
+make -j`nproc`
+```
+
+This will build Cuberite in release mode, which is better for almost all cases. For more `cmake` options, or for building in debug mode, see the section below.
+
 
 CMake on Unix-based platforms
 -----------------------------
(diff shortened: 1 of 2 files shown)


Schwertspize:
compiling.MD5

Seadragon91:
@SafwatHalaby Can you fix this line, too? `If you're using Git to get the source, use the following set of commands to set up the local workspace correctly:`. Because this is only 1 command.

SafwatHalaby:
I have no idea why I said "md5". rebasing.
@Seadragon91 will do.

Schwertspize:
just look at the headline  and your commits

SafwatHalaby:
@Schwertspize , rebased and fixed.

SafwatHalaby:
is `git clone --recursive <url>` totally equavilant to `git clone <url> && git submodule init --update`?

Seadragon91:
@SafwatHalaby Yes

SafwatHalaby:
lgtm

COMPILING.md
```diff
@@ -87,9 +87,16 @@ git clone --recursive https://github.com/cuberite/cuberite.git
 
 ### Building ###
 
-Follow the instructions at [CMake on Unix-based platforms](#cmake-on-unix-based-platforms).
+Run the following commands to build for Linux:
```
SafwatHalaby: This is a mistake. The main heading says "Linux, FreeBSD etc".

SafwatHalaby:
Is there a *nix `nproc` alternative?

bearbin:
I am opposed to the replacement of clang with the inferior and slower gcc.

SafwatHalaby:
I suspected there might be an opposition. Perhaps that should have been a different PR. I have no strong opinion on compiler choice. I'll just re-clang it.

SafwatHalaby:
Re-added clang as a default. Rebased.

SafwatHalaby:
lgtm

madmaxoft:
Looks good to me. Rebase onto current master and let's merge.

COMPILING.md
```diff
@@ -87,9 +87,17 @@ git clone --recursive https://github.com/cuberite/cuberite.git
...
 ### Building ###
 
-Follow the instructions at [CMake on Unix-based platforms](#cmake-on-unix-based-platforms).
+Run the following commands to build Cuberite:
+
+```
+mkdir Release
+cd Release
+cmake -DCMAKE_BUILD_TYPE=RELEASE ..
+make -j`nproc`
```
Schwertspize: doesn't it have to be

    make -j `nproc`

SphinxC0re: That's okay as long as you are not adding any value to it

madmaxoft: How about simple `make -j`, without a value?

SphinxC0re: What does it do?

madmaxoft: It runs as many jobs as possible at once. With current hardware, the bottleneck is seldom the CPU, so the number of cores is no longer that important anyway.

SphinxC0re: sounds good. How does our CI do it?

SafwatHalaby: The [manual](https://www.gnu.org/software/make/manual/html_node/Parallel.html) says "By default, there is no load limit.". Wouldn't `make -j` be dangerous on small machines like the Pi?

madmaxoft: I don't think so, because it's still limited by the number of jobs that can actually run together; only jobs that don't depend on each other can be run in parallel. In Cuberite, this mostly means the external libraries, so it's like 10 jobs max, give or take. Should be okay.

SafwatHalaby: I think this should be tested first. I remember my laptop used to run out of ram and resort to swap with `-j4` when compiling some program (PhantomJS?), drastically increasing compilation time. with `-j3` all was good.

SafwatHalaby: @Schwertspize `-j3` is legal. Many *Nix commands allow you to omit the space with numbers. e.g. `head -n15`.

bearbin: Make -j3 causes pretty much any machine to run out of RAM. -j2 or -j3 is probably optimal for most machines (IO wait is minimised on single threaded machines, and becomes the limiting factor with more threads).

SafwatHalaby: Let's just assume that a person who has no `nproc` knows what they're doing.

SafwatHalaby:
Apparently, the CI has deadlocked due to an issue unrelated to this PR. Can someone merge/rerun CI?

madmaxoft:
Restarted the build.

It seems that builds fail if the spawnpoint generation fails to find a suitable spawnpoint.

PR #2696, opened 2015-11-29
Implement aging + breeding

Gargaj:
Notes:
- Tested and works with pigs, sheep, cows/mooshrooms, rabbits and chicken
- Works with ocelots although taming for ocelots isn't implemented yet.
- Doesn't work yet with wolves: they are not a passive but a neutral mob. This was an oversight in my part, but I didn't want to move the behaviour into ```cMonster``` just yet because it would introduce a whole new set of problems.
- Horses need taming first and that behaviour seems dodgy / broken, so I didn't yet try that.

Another important note: There's a change in the 1.8 protocol implementation when it comes to communicating age: @rom1504 and I did a bit of snooping in the less-documented parts of the protocol (http://wiki.vg/Entities#Ageable) and found out that the only way to make the love-mode hearts disappear is to set the Age to x>0. This introduces 3 distinct states for Age: x<0 when baby, x==0 when adult and ready for mating, x>0 when exhausted from mating.

### DIFF
--- a/src/Mobs/PassiveMonster.cpp
+++ b/src/Mobs/PassiveMonster.cpp
@@ -4,12 +4,17 @@
 #include "PassiveMonster.h"
 #include "../World.h"
 #include "../Entities/Player.h"
+#include "BoundingBox.h"
 
 
 
 
 cPassiveMonster::cPassiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height) :
-	super(a_ConfigName, a_MobType, a_SoundHurt, a_SoundDeath, a_Width, a_Height)
+	super(a_ConfigName, a_MobType, a_SoundHurt, a_SoundDeath, a_Width, a_Height),
+	m_LovePartner(nullptr),
+	m_LoveTimer(0),
+	m_LoveCooldown(0),
+	m_MatingTimer(0)
 {
 	m_EMPersonality = PASSIVE;
 }
@@ -35,6 +40,31 @@ bool cPassiveMonster::DoTakeDamage(TakeDamageInfo & a_TDI)
 
 
 
+void cPassiveMonster::EngageLoveMode(cPassiveMonster * a_Partner)
+{
+	m_LovePartner = a_Partner;
+	m_MatingTimer = 50;  // about 3 seconds of mating
+}
+
+
+
+
+
+void cPassiveMonster::ResetLoveMode()
+{
+	m_LovePartner = nullptr;
+	m_LoveTimer = 0;
+	m_MatingTimer = 0;
+	m_LoveCooldown = 20 * 60 * 5;  // 5 minutes
+
+	// when an animal is in love mode, the client only stops sending the hearts if we let them know it's in cooldown, which is done with the "age" metadata
+	m_World->BroadcastEntityMetadata(*this);
+}
+
+
+
+
+
 void cPassiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	super::Tick(a_Dt, a_Chunk);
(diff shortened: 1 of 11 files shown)


src/Mobs/PassiveMonster.cpp
```diff
@@ -43,25 +73,122 @@ void cPassiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+				}
+
+				virtual bool Item(cEntity * a_Entity) override
+				{
+					// if we're the same species as someone around and they dont have a partner, swipe right
+					if ((a_Entity->GetEntityType() == m_Me->GetEntityType()) && (a_Entity != m_Me))
+					{
+						cPassiveMonster * Me = static_cast<cPassiveMonster*>(m_Me);
+						cPassiveMonster * Partner = static_cast<cPassiveMonster*>(a_Entity);
+						if (Partner->IsInLove() && (Partner->GetPartner() == nullptr))
```
SafwatHalaby: `....&& me->getPartner() == nullptr`?

SafwatHalaby:
Minor remark: At https://github.com/Gargaj/cuberite/blob/breeding/src/World.cpp#L3485 
It's probably best to have it that way:
```
if (a_Baby)
{
	Monster->SetAge(-1);
}
else
{
   mob->stopAgingTimer();
}
```

Gargaj:
Probably true, although in the grand scheme of things it doesn't change much if it spawns as an adult, it just counts the aging timer down and then doesn't do anything.

SafwatHalaby:
True as of now. But the value may be used in the future in ways we can't anticipate. Suppose some Lua API wants to query how long it'd take for a mob to grow up. Currently it would return a false answer for adult mobs.

Gargaj:
It's a bit of a mess because "age" seems to have two uses: http://minecraft.gamepedia.com/Chunk_format/Breedable

src/Mobs/PassiveMonster.cpp
```diff
@@ -43,25 +73,122 @@ void cPassiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			ResetLoveMode();
 		}
 	}
+	else
+	{
+		cItems FollowedItems;
+		GetFollowedItems(FollowedItems);
+		if (FollowedItems.Size() > 0)
+		{
+			cPlayer * a_Closest_Player = m_World->FindClosestPlayer(GetPosition(), static_cast<float>(m_SightDistance));
```
SafwatHalaby: This code means that the player holding the bait must be the nearest player, is this the case in Vanilla? This is a minor issue though, and can be resolved after this merge.

SafwatHalaby:
- Does the love cooldown reset on a failed attempt? The only call to `ResetLoveMode` is when breeding succeeds. Shouldn't it be called also when `m_LoveCooldown` reaches zero?
- I am pretty ignorant with regards to the chunk saving aspect of Cuberite. Are the timers saved properly on chunk loads / unloads? @worktycho @madmaxoft @tigerw

src/Mobs/PassiveMonster.cpp
Gargaj: That's not my code, I just worked around it.

Gargaj:
1. From what I can tell trying it in vanilla, no, you can re-feed the animals once the initial burst of love expires.
2. I didn't update the NBT writer yet.

rom1504:
You need to send the entity status "love mode" every n ms (got to figure out n) to near players if you want them to see the hearts for 30s. Only the client of the player that give the wheat predicts the "love mode" itself. Not sure if you already do that ?

SafwatHalaby:
@Gargaj You are correct about point 1.

SafwatHalaby:
Shouldn't the partner be nullified when `m_LoveTimer` hits zero to avoid chasing the partner after the hearts go away?

Gargaj:
There's no such state. If the mob has a partner, ```m_LoveTimer``` doesn't matter anymore.

SafwatHalaby:
Consider a case where you isolate the two partners just after they lock in to each other. Shouldn't they give up at some point?

Gargaj:
Right now it doesn't actually matter that they can't reach each other. :)
I don't know how vanilla deals with that, if at all.

SafwatHalaby:
I just tested this, two chickens had bred through a 2-block thick wall.  Once they lock in, their physical position doesn't matter any more, a baby will be born. I never knew that.

SafwatHalaby:
-Tested with Vanilla-

Gargaj:
There are so many jokes to be made about that.

SafwatHalaby:
Lol.
The code looks great to me. Thanks for this PR.
Is it ok for me to merge?

SafwatHalaby:
Vanilla does care about mob reaching each other. Apparently ~2 blocks is the maximum distance in which the mobs can still breed. Seperating them with a 3-block wall prevents breeding and makes them give up after a while.

src/Mobs/PassiveMonster.h
```diff
@@ -23,7 +24,22 @@ class cPassiveMonster :
...
+	void ResetLoveMode();
+
+	bool IsInLove() const { return (m_LoveTimer > 0); }
+	bool IsInLoveCooldown() const { return (m_LoveCooldown > 0); }
+
+protected:
+	cPassiveMonster * m_LovePartner;
+	int m_LoveTimer;
+	int m_LoveCooldown;
+	int m_MatingTimer;
```
madmaxoft: These should have been documented - what they do, how they interact with each other etc.

SafwatHalaby: https://github.com/cuberite/cuberite/pull/2699/files

SafwatHalaby: https://github.com/cuberite/cuberite/pull/2699

src/Mobs/PassiveMonster.cpp
NiLSPACE: What about a DoWithClosestPlayers(Position, Callback)?
It would sort the players based on how far they are away from the position. The closest player first, and the furthest as last. When the callback returns true it stops.

SafwatHalaby: Shouldn't there be an equiavilant with a max-distance cap? You don't want to sort all the players in the world every tick for every mob.

madmaxoft: This code is just ripe for a rewrite. We don't want to be finding the closes player all the time, instead, we should store the followed player and only update it from time to time. The variable name is super-confusing, too, since we use `a_` prefix for function **a**rgs. And of course we want to react to non-closest players, too, if applicable.

Gargaj:
Question: If you said above 2 blocks distance they dont breed, do they go into post-breed cooldown or can you re-feed them?

SafwatHalaby:
I will test that as soon as I can. But I think this is the least urgent of the AI issues.

Gargaj:
Sure, but also reasonably easy to fix :)

jammet:
I think they should be in love mode for a few more seconds, -- right now it's rather short, but I'm so glad this is being tackled! 

Maybe it was just by chance but I noticed that sheep do not inherit their fur/wool color yet, and that we don't get domesticated cats when ocelots breed.

Thanks for all of this!! :)

Outfenneced:
Looks like there's currently no way to pass down information from parents. Might even just be a good idea to pass the parents to the baby's constructor and have each baby mob do their own thing with the information. For instance, a baby sheep will find the wool color of the parents and use that to determine their wool color.

jammet:
On a technical side, I don't know how this works at all. But I'm glad you're figuring it out. Thank you very much :).

Gargaj:
The only reason I haven't added the wool color cos I haven't figured out how the colors blend when they're different color. Otherwise it's just some additional conditions right after the baby spawn.

PR #2704, opened 2015-12-01
Fix wolf collars

Gargaj:
I don't know why this had an inversion but it consistently made the wrong collar color (i.e. yellow dye made the collar blue)

### DIFF
--- a/src/Mobs/Wolf.cpp
+++ b/src/Mobs/Wolf.cpp
@@ -18,7 +18,7 @@ cWolf::cWolf(void) :
 	m_IsBegging(false),
 	m_IsAngry(false),
 	m_OwnerName(""),
-	m_CollarColor(14)
+	m_CollarColor(E_META_DYE_ORANGE)
 {
 	m_RelativeWalkSpeed = 2;
 }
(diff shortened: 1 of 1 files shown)


src/Mobs/Wolf.cpp
```diff
@@ -18,7 +18,7 @@ cWolf::cWolf(void) :
 	m_IsBegging(false),
 	m_IsAngry(false),
 	m_OwnerName(""),
-	m_CollarColor(14)
+	m_CollarColor(E_META_DYE_ORANGE)
```
NiLSPACE: By default they're red right, not orange?

Gargaj: According to the changelog on http://minecraft.gamepedia.com/Wolf they were red, then orange, then red again in 1.9 - it's a mess really.

NiLSPACE: Make up your mind @mojang ^ ^

Gargaj: Either way the other change (the 15-x one) is more important, and it'd be nice if someone would confirm why it was like that.

madmaxoft:
Rebase onto current master and let's merge this. I don't care about the default color.

Gargaj:
Done.

Gargaj:
Done.

madmaxoft:
Thanks

src/Mobs/Wolf.cpp
tigerw: I seem to recall doing that. If it was me, either I had no idea what I was doing, or it was because it worked in whatever MC version the changes were implemented in.

PR #2706, opened 2015-12-01
Stop creepers from exploding even if they just died

Gargaj:
Should mobs even ```Tick()``` once they died?

### DIFF
--- a/src/Mobs/Creeper.cpp
+++ b/src/Mobs/Creeper.cpp
@@ -43,7 +43,7 @@ void cCreeper::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 			m_ExplodingTimer += 1;
 		}
 
-		if (m_ExplodingTimer == 30)
+		if ((m_ExplodingTimer == 30) && (GetHealth() > 0.0))  // only explode when not already dead
 		{
 			m_World->DoExplosionAt((m_bIsCharged ? 5 : 3), GetPosX(), GetPosY(), GetPosZ(), false, esMonster, this);
 			Destroy();  // Just in case we aren't killed by the explosion


SphinxC0re:
Looks okay to me. Pleas add some documentation for what you just did

Gargaj:
Sigh, okay - question still stands though, is there any reason why entities who are dead should ```Tick()```? I can't seem to find any.

NiLSPACE:
Physics is also done in the Tick thread I believe. Can dead entities cause other things to activate like pressure plates? Just before they disappear perhaps?

Gargaj:
No, I think in vanilla once health is 0 nothing is calculated, that's why the falling-over animation stays in one place even if it's in mid-air.

Schwertspize:
I think we should not limitate to creeper. just check in tick if the entities has health >0 otherwise break (or continue or whatever)

madmaxoft:
We need to tick the mobs, because without it they won't get destroyed - the final destruction code gets called from the tick after a predefined timeout (for the smoke particle effect).

Yes, ticking physics and other behavior doesn't make sense in that scenario, so we could turn those off.

Gargaj:
So what's the preferred solution?

SafwatHalaby:
@madmaxoft merging a single PR out all other PRs until the people behind them rebase. Is this on purpose? It's rather limiting.

madmaxoft:
Yes, it's becoming rather limiting. On the other hand, not following this workflow makes the git history really tangled and difficult to follow. I suppose we need to find a reasonable compromise. Up until now most PRs were submitted by people working already in the canon repo, so one could rebase the PR before merging easily. Now that we get so many external contributors, it's getting out of hand. I'd say a compromise of "no need to rebase if not too old" would be good, the question is how old is old. A few commits is okay, 30 commits old is too old.

Gargaj:
I would say that should depend on the scope of the PR - most of my PR's so far have been single-line logic fixes that rarely depend on other parts of the code (i.e. merging them in rarely depended on whatever changed around it), but often they had to be rebased because someone changed a README or something. I don't mind rebasing as such, but since there's no notification mechanism, often I don't realize when a PR becomes stale.

As far as this specific PR goes, I tried looking at the code, but once again the only thing I can think of as a reasonable "global" solution is changing the return value from ```void``` to ```bool``` where returning false that the tick should not propagate to virtuals (i.e. ```if (!super::Tick(dt)) return false;```) That would allow the super tick to bail early in case the entity died, but still take care of the one-second delay mechanism.

SafwatHalaby:
> On the other hand, not following this workflow makes the git history really tangled and difficult to follow

How so?

worktycho:
Merges cause a lot more diamonds in the commit graph, making it a lot less linear. That makes the history much harder to understand in any tools that attempt to linearize the history.

Gargaj:
Soooooo...? Are we gonna do some more bikeshedding over this or...?

PR #2708, opened 2015-12-01
Allow use failures to propagate from the entity/block to the player

### LINKED ISSUE #2573: Torches placed on mob spawner not removed, if spawner destroyed
If I place a torch on a mob spawner and destroy the mob spawner the torch not falls down.

Gargaj:
This fixes #2573; currently if a block is "usable", it will never fail the use, so interacting with it as if it was a normal block will not happen - the client does some client-side prediction so e.g. placing a torch on a spawner will visually appear, but it won't actually be there and breaking it will result in a "tried to dig air" warning. Since the torches are not actually there, breaking the spawner (or e.g. a jukebox) will not update the blocks around it and the phantom-torches will keep floating in the air and will never pop off (since they're not really there).

### DIFF
--- a/src/BlockEntities/BlockEntity.h
+++ b/src/BlockEntities/BlockEntity.h
@@ -102,8 +102,9 @@ public:
 	
 	// tolua_end
 	
-	/** Called when a player uses this entity; should open the UI window */
-	virtual void UsedBy( cPlayer * a_Player) = 0;
+	/** Called when a player uses this entity; should open the UI window.
+	returns true if the use was successful, return false to use the block as a "normal" block */
+	virtual bool UsedBy( cPlayer * a_Player) = 0;
 	
 	/** Sends the packet defining the block entity to the client specified.
 	To send to all eligible clients, use cWorld::BroadcastBlockEntity() */
--- a/src/ChunkMap.h
+++ b/src/ChunkMap.h
@@ -100,8 +100,9 @@ public:
 	/** Sends the block entity, if it is at the coords specified, to a_Client */
 	void SendBlockEntity(int a_BlockX, int a_BlockY, int a_BlockZ, cClientHandle & a_Client);
 	
-	/** a_Player rclked block entity at the coords specified, handle it */
-	void UseBlockEntity(cPlayer * a_Player, int a_X, int a_Y, int a_Z);
+	/** a_Player rclked block entity at the coords specified, handle it
+	returns true if the use was successful, return false to use the block as a "normal" block */
+	bool UseBlockEntity(cPlayer * a_Player, int a_X, int a_Y, int a_Z);
 	
 	/** Calls the callback for the chunk specified, with ChunkMapCS locked; returns false if the chunk doesn't exist, otherwise returns the same value as the callback */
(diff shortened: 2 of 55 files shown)


src/ClientHandle.cpp
```diff
@@ -1383,15 +1383,16 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 
 		if (BlockHandler->IsUseable() && !m_Player->IsCrouched())
 		{
-			if (PlgMgr->CallHookPlayerUsingBlock(*m_Player, a_BlockX, a_BlockY, a_BlockZ, a_BlockFace, a_CursorX, a_CursorY, a_CursorZ, BlockType, BlockMeta))
+			if (!PlgMgr->CallHookPlayerUsingBlock(*m_Player, a_BlockX, a_BlockY, a_BlockZ, a_BlockFace, a_CursorX, a_CursorY, a_CursorZ, BlockType, BlockMeta))
```
NiLSPACE: I'd prefer to keep the  
```cpp
if (val)
{
   return;
}

//Other code
```

src/ChunkMap.h
```diff
@@ -101,7 +101,7 @@ class cChunkMap
 	void SendBlockEntity(int a_BlockX, int a_BlockY, int a_BlockZ, cClientHandle & a_Client);
 	
 	/** a_Player rclked block entity at the coords specified, handle it */
-	void UseBlockEntity(cPlayer * a_Player, int a_X, int a_Y, int a_Z);
+	bool UseBlockEntity(cPlayer * a_Player, int a_X, int a_Y, int a_Z);
```
madmaxoft: Add documentation for the return value

src/ClientHandle.cpp
Gargaj: That's exactly the crux of that change - if you return immediately, you're not leaving the "default" code below it a chance to run. I don't know if that's by design when it comes to plugins (i.e. should a plugin be able to block any interaction with a block, like placing a torch on it), but certainly when it comes to the non-plugin behaviour, my fix is necessary the way it is.

src/ChunkMap.h
Gargaj: Done

src/BlockEntities/BlockEntity.h
```diff
@@ -103,7 +103,7 @@ class cBlockEntity
 	// tolua_end
 	
 	/** Called when a player uses this entity; should open the UI window */
-	virtual void UsedBy( cPlayer * a_Player) = 0;
+	virtual bool UsedBy( cPlayer * a_Player) = 0;
```
madmaxoft: Again, document the return value. Need to do that for all non-overriding functions.

Gargaj: Done

madmaxoft:

madmaxoft:
Wait, we might want to consider documenting this behavior in the HOOK_PLAYER_USING_BLOCK documentation - what exactly happens if the plugin rejects the usage. Please edit the corresponding file in the APIDocs plugin.

Gargaj:
Like so?

PR #2711, opened 2015-12-03
Use C++11 auto instead of per-platform types.

madmaxoft:
Ref.: http://forum.mc-server.org/showthread.php?tid=2230&pid=23674#pid23674


SphinxC0re:
lgtm

madmaxoft:
There's plenty of places where `auto` could be used, but this one was painful because it broke the FreeBSD builds. As reported in the linked forum thread, it was the only thing keeping Cuberite from compiling.

PR #2713, opened 2015-12-05
Tilling dirt should send block update

Gargaj:
There are BUD switches that are built on this (https://www.youtube.com/watch?v=Bk_ChvmtCNY) but more prominently, torches placed on dirt blocks should pop off when the dirt block is tilled.

### DIFF
--- a/src/Items/ItemHoe.h
+++ b/src/Items/ItemHoe.h
@@ -57,7 +57,7 @@ public:
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


tigerw:
hoe it must be said that, till it becomes absolutely necessary, any redstone update bugs will be torched, since leaving them in makes dirty code and hidden openings for more issues to emerge.

----

Bad attempts at punning are bad.

SphinxC0re:
Oh maaaan! So many bad jokes these days :laughing:

Gargaj:
:neutral_face:

PR #2726, opened 2015-12-10
Proposed fixes to pistons.

lighth7015:
These changes are my proposed fixes for the piston blocks. I have some more work to do to these, however they seem to work well enough on my test server.


src/Blocks/BlockPiston.cpp
```diff
@@ -10,29 +9,28 @@
 #include <vector>
 #include <array>
 
-
-
-
-
```
SphinxC0re: Please keep all these lines

PR #2744, opened 2015-12-13
Strip meta from hopper on converting to pickup

SphinxC0re:


SphinxC0re:
Actually, this is wrong. The  hoppers rotation should be reset to default, but on destruction, the  hopper should spit out all of its items. I'll work on this

SphinxC0re:
No :smile: this works

PR #2748, opened 2015-12-15
Changed raw cPath to an unique_ptr, fixes memory leak (Fixes #2745)

### LINKED ISSUE #2745: Memory leak
I tested the windows builds from `http://builds.cuberite.org/` and have problem with the memory usage of them. I tested this on 2 computers (Win 7 x64)

The x86 build crashes seconds after I joined the server.
The x64 build increases the memory up to max RAM and later it crashes, too.

Self compiled builds works fine. Another thing is, the problem occurs after I join the server. I can pre-generate a world with over 10k chunks without any problems.

 Anyone can test this? Be careful it could freezes your computer and then it can be necessary to force a shutdown.

Edit: Any of the commit starting at this commit 6995af0 are the cause for a memory leak.

Seadragon91:
This fixes the memory leak (fixes #2745) for builds who will be compiled with Visual Studio 2013. Builds compiled with Visual Studio 2015, had no problem with this.

### DIFF
--- a/src/Mobs/PathFinder.cpp
+++ b/src/Mobs/PathFinder.cpp
@@ -46,17 +46,17 @@ ePathFinderStatus cPathFinder::GetNextWayPoint(cChunk & a_Chunk, const Vector3d 
 	}
 
 	// If m_Path has not been initialized yet, initialize it.
-	if (!m_Path.IsValid())
+	if (!m_Path->IsValid())
 	{
 		ResetPathFinding(a_Chunk);
 	}
 
-	switch (m_Path.CalculationStep(a_Chunk))
+	switch (m_Path->CalculationStep(a_Chunk))
 	{
 		case ePathFinderStatus::NEARBY_FOUND:
 		{
 			m_NoPathToTarget = true;
-			m_PathDestination = m_Path.AcceptNearbyPath();
+			m_PathDestination = m_Path->AcceptNearbyPath();
 			if (a_DontCare)
 			{
 				m_FinalDestination = m_PathDestination;
@@ -89,7 +89,7 @@ ePathFinderStatus cPathFinder::GetNextWayPoint(cChunk & a_Chunk, const Vector3d 
 				return ePathFinderStatus::CALCULATING;
 			}
 
-			if (m_Path.NoMoreWayPoints())
+			if (m_Path->NoMoreWayPoints())
 			{
 				// We're always heading towards m_PathDestination.
 				// If m_PathDestination is exactly m_FinalDestination, then we're about to reach the destination.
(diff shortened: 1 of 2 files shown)


SafwatHalaby:
Hold on a second. How would the existing code ever leak anything? It's not even a pointer.

Seadragon91:
I hope that worktycho can shed light on that, but this fixes the memory leak problem. This could be also a bug in the Visual Studio 2013 compiler, not sure.

src/Mobs/PathFinder.cpp
```diff
@@ -142,7 +142,7 @@ void cPathFinder::ResetPathFinding(cChunk &a_Chunk)
 	m_NoPathToTarget = false;
 	m_PathDestination = m_FinalDestination;
 	m_DeviationOrigin = m_PathDestination;
-	m_Path = cPath(a_Chunk, m_Source, m_PathDestination, 20, m_Width, m_Height);
```
SafwatHalaby: Doesn't this call the destructor of m_Path?

madmaxoft: No, it doesn't, it calls the assignment operator.

SafwatHalaby: Sorry, my bad.
Ok, it calls the default assignment oprerator, which calls the assignment operator for all the fields of cPath, which, as far as I can tell, shouldn't leak. I think we should pinpoint the issue, which is possible in cPath, before merging this, to avoid further trouble.

SafwatHalaby:
I merged this, but I think we should still find the root cause.

worktycho:
I think the issue is due to the pointers into m_Map. When the class is copied, moved or the assignment operators were used, these pointers are not updated, causing dangling pointers.

SafwatHalaby:
The implicit assignment operator does something like this:

```
cPath::operator = (cPath const & other)
{
    m_Map = other.m_Map; // other.m_Map is empty.
    m_OpenList = other.m_OpenList; // other.m_OpenList is empty.
    //etc.
}
```

So, the pointers you talk about should be gone. Also, if this is a logical error, why isn't it present in VS 2015?

worktycho:
Actually its
```
cPath & cPath::operator = (cPath && other)
{
    m_Map = std::move(other.m_Map); // As soon as this line is executed, what the pointers in m_OpenList point to is undefined.
    m_OpenList = std::move(other.m_OpenList);
    //etc.
}
```
As soon as you call cPath(a_Chunk, m_Source, m_PathDestination, 20, m_Width, m_Height) you fill m_Map and m_OpenList.
As for why its not present in VS 2015, I would guess that `std::map<T>::operator=(std::map<T>&&)` changed between versions.

SafwatHalaby:
That's definitely the problem. My assumption that other.m=M_List is empty is incorrect.

PR #2750, opened 2015-12-15
Fixed warning regarding an uninitialized value

SphinxC0re:
I should really collect more of those


Seadragon91:
Oh again a deadlock by the spawn point... Had three on my computer of them. Should really look what there is wrong...

PR #2751, opened 2015-12-15
PieceStructures generator: Fixed wrong merge conflict resolution.

madmaxoft:
Also improved reporting messages.
Ref.: http://forum.mc-server.org/showthread.php?tid=2256


Seadragon91:
Again a deadlock, hehe nice hit.

madmaxoft:
Weird thing is, when I copy the seed to my local instance of the server, it works without any problems. I wonder what's going on there.

Seadragon91:
I had around 3 pull requests, who deadlocked. Tested all the seeds local and had no problem with them. I just say, it's something like a race-condition.

madmaxoft:
Rebased and merged.

PR #2754, opened 2015-12-15
Allow horse control

Gargaj:
Still quite buggy, but at least it fixes 1. the protocol (which was totally wrong) and 2. the horse going willy-nilly if someone is saddled upon it 3. the fact that you literally couldn't control the horse :D

What remains: There's still no proper speed mechanism, the horse for some reason glides along randomly once you moved with it, jumping doesn't work, and the single-block climb isn't there. (For whoever would like to give it a shot.)


madmaxoft:

SafwatHalaby:
Cuberite: Where horses walk diagonally and can go reverse full speed. :tm:

SphinxC0re:
Definitely a feature

PR #2755, opened 2015-12-15
Working coverage build

worktycho:


madmaxoft:
I'd appreciate some more information on the coverage builds - what they can be used for, what tools work with them, what web pages to check out etc. Could you write some tiny guide as a forum post, for us dumb ones?

worktycho:
It uses lcov.

If you use this setup with gcc as a compiler: `cmake -DCMAKE_BUILD_TYPE=COVERAGE -DSELF_TEST=1`
Then run `make Cuberite_coverage` it generates a html report in build dir/coverage which says what code was run during the tests.

The idea is to provide guidance for what we are and aren't testing, in order to start getting some actual unit testing.

PR #2757, opened 2015-12-15
Preinitialized variables

SphinxC0re:


madmaxoft:
Do not merge yet!

All these values should be initialized properly in the `CalcFloorFrac()` function, yet both clang and coverity complain that they aren't. This needs further investigation, rather than slapping zeroes all over the place. If there is indeed a possibility of uninitialized value getting in there, then we have a real problem.

SafwatHalaby:
A general remark regarding intialization:
I personally dislike initializing variables with placeholder values (like zeros) which will be overridden later before ever being used. Because it reduces the likelihood of crashing when there is an error. Failing noisily is a good thing. It is also a waste of cpu cycles.

tigerw:
I remember reading somewhere that just passing uninitialised variables can constitute undefined behaviour, but the article concerned C.

madmaxoft:
If you pass them by-value, it is UB. If you pass them by-ref, it's a valid scenario (you want the function to fill them in)

SphinxC0re:
I'm so sorry for all the whitespace deletions. Atom does some nasty stuff to the code.

I removed most of the unnecessary initializations but I found some more places where uninitialized values throw a warning

src/Bindings/LuaJson.cpp
```diff
@@ -134,7 +134,7 @@ static Json::Value JsonSerializeTable(cLuaState & a_LuaState)
 	{
 		if (lua_type(a_LuaState, -2) == LUA_TNUMBER)
 		{
-			int idx;
+			int idx = 0;
 			a_LuaState.GetStackValue(-2, idx);
```
worktycho: This call to GetStackValues should check the return code for an error. Then the uninitialized value would never be read.

SphinxC0re: This should be handled in a different PR

madmaxoft: I don't think it's needed, because we've already checked that it's a number, so it's always read successfully (unless someone else bombs our memory).
I guess I could rewrite this to instead depend on the `GetStackValue()`'s return as to whether the value is an int or a string

madmaxoft: Still, another PR.

worktycho: It can fail. Lua numbers are doubles, and this is an int. So if I write plugin code to pass 1e40 to the function, its a number but its too big to fit in an int so get stack values should fail.

SphinxC0re:
Merge?

SphinxC0re:
Rebased again. Merge?

madmaxoft:
I'm against a merge of this kind of code. It hides the problems, instead of solving them.

We need to find out *why* the variables aren't initialized in the functions that are expected to assign values to them, instead of slapping the "don't care" plaster all over the place.

Also, set your editor not to remove the final empty lines of the files; those unrelated whitespace changes are not nice either.

PR #2758, opened 2015-12-16
Fixed String Parsing crash bug

worktycho:
Check string length against actual remaining data, not an arbitrary constant

### DIFF
--- a/src/WorldStorage/FastNBT.cpp
+++ b/src/WorldStorage/FastNBT.cpp
@@ -91,11 +91,7 @@ bool cParsedNBT::ReadString(size_t & a_StringStart, size_t & a_StringLen)
 	NEEDBYTES(2);
 	a_StringStart = m_Pos + 2;
 	a_StringLen = static_cast<size_t>(GetBEShort(m_Data + m_Pos));
-	if (a_StringLen > 0xffff)
-	{
-		// Suspicious string length
-		return false;
-	}
+	NEEDBYTES(a_StringLen);
 	m_Pos += 2 + a_StringLen;
 	return true;
 }


src/WorldStorage/FastNBT.cpp
```diff
@@ -91,9 +91,9 @@ bool cParsedNBT::ReadString(size_t & a_StringStart, size_t & a_StringLen)
 	NEEDBYTES(2);
 	a_StringStart = m_Pos + 2;
 	a_StringLen = static_cast<size_t>(GetBEShort(m_Data + m_Pos));
-	if (a_StringLen > 0xffff)
+	if (a_StringLen > m_Length - m_Pos - 2)
```
madmaxoft: Both `m_Length` and `m_Pos` are unsigned; an underflow can occur here when reading too close to the end of the data.

worktycho: No it can't without UB. If `m_Pos > m_Length` then we hit UB at line 93.
And if `2 < m_Length - m_Pos` then we should of exited in NEEDBYTES.

madmaxoft: Can't you just change it to `if (a_StringLen + 2 > m_Length - m_Pos)` ? It's clean and it's robust.

worktycho:
I've swapped it out for the NEEDBYTES macro.

worktycho:
Just discovered theres a bug in this fix. Because we don't increment m_Pos before the second `NEEDBYTES`, we should `NEEDBYTES(2 + a_StringLen)`.

madmaxoft:
Good catch

worktycho:
It was the fuzzer that found it. It causes my test program to crash if it hits a string tag with a length one longer than the remaining data.

PR #2760, opened 2015-12-16
Baby chickens don't lay eggs

Gargaj:


madmaxoft:
How about early bail-out:
```cpp
if (IsBaby())
{
  return;
}
```

Gargaj:
Does it matter?

madmaxoft:
It improves readability by reducing the nesting level

Gargaj:
I would argue ```if (this isn't true) { do something; }``` is easily more readable than ```if (this is true) nothing to do here;```

NiLSPACE:
It's more like `if (this is true) nothing left do do for us. We can leave`

madmaxoft:
Compare:
```cpp
if (something)
{
  ...
  if (somethingelse)
  {
    ...
    if (yetanother)
    {
      ...
    }
  }
}
```
with
```cpp
if (!something)
{
  // Nothing needed
  return;
}

...
if (!somethingelse)
{
  // We can't do this now
  return;
}

...
if (!yetanother)
{
  // Access denied
  return;
}
...
```

PR #2761, opened 2015-12-16
Moved variables into scope, removed unused variables and fixed variables

Seadragon91:
Playing around with static analyzers :)
* Moved variables into scope
* Removed unused variables
* Fixed use of wrong variable

### DIFF
--- a/src/BlockEntities/BrewingstandEntity.cpp
+++ b/src/BlockEntities/BrewingstandEntity.cpp
@@ -92,10 +92,10 @@ bool cBrewingstandEntity::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		return false;
 	}
 
-	const cBrewingRecipes::cRecipe * Recipe = nullptr;
 	// The necessary brewing time, has been reached
 	if (m_TimeBrewed >= m_NeedBrewingTime)
 	{
+		const cBrewingRecipes::cRecipe * Recipe = nullptr;
 		BroadcastProgress(0, 0);
 		m_IsBrewing = false;
 		m_TimeBrewed = 0;
--- a/src/Blocks/BlockCrops.h
+++ b/src/Blocks/BlockCrops.h
@@ -75,14 +75,7 @@ public:
 	
 	virtual void OnUpdate(cChunkInterface & cChunkInterface, cWorldInterface & a_WorldInterface, cBlockPluginInterface & a_PluginInterface, cChunk & a_Chunk, int a_RelX, int a_RelY, int a_RelZ) override
 	{
-		NIBBLETYPE Meta     = a_Chunk.GetMeta      (a_RelX, a_RelY, a_RelZ);
-		NIBBLETYPE Light    = a_Chunk.GetBlockLight(a_RelX, a_RelY, a_RelZ);
-		NIBBLETYPE SkyLight = a_Chunk.GetSkyLight  (a_RelX, a_RelY, a_RelZ);
-
-		if (SkyLight > Light)
-		{
-			Light = SkyLight;
-		}
+		NIBBLETYPE Meta = a_Chunk.GetMeta(a_RelX, a_RelY, a_RelZ);
 
 		// Check to see if the plant can grow
 		auto Action = CanGrow(a_Chunk, a_RelX, a_RelY, a_RelZ);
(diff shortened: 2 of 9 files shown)


src/PolarSSL++/BlockingSslClientSocket.cpp
```diff
@@ -189,7 +189,7 @@ bool cBlockingSslClientSocket::Send(const void * a_Data, size_t a_NumBytes)
 	size_t NumBytes = a_NumBytes;
 	for (;;)
 	{
-		int res = m_Ssl.WritePlain(a_Data, a_NumBytes);
+		int res = m_Ssl.WritePlain(Data, a_NumBytes);
```
Seadragon91: Fixed use of wrong variable. The method `Send` is not used, but better fix that.

src/Entities/Entity.cpp
```diff
@@ -513,7 +513,7 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
 			RemovedDamage += CeilC(EPFProtection * 0.04 * a_TDI.FinalDamage);
 		}
 
-		if ((a_TDI.DamageType == dtFalling) || (a_TDI.DamageType == dtFall) || (a_TDI.DamageType == dtEnderPearl))
+		if ((a_TDI.DamageType == dtFalling) || (a_TDI.DamageType == dtEnderPearl))
```
Seadragon91: dtFalling and dtFall have the same int value. No need to check twice :)

src/Blocks/BlockCrops.h
```diff
@@ -75,14 +75,7 @@ class cBlockCropsHandler :
...
-		NIBBLETYPE SkyLight = a_Chunk.GetSkyLight  (a_RelX, a_RelY, a_RelZ);
-
-		if (SkyLight > Light)
-		{
-			Light = SkyLight;
-		}
+		NIBBLETYPE Meta = a_Chunk.GetMeta(a_RelX, a_RelY, a_RelZ);
 
 		// Check to see if the plant can grow
 		auto Action = CanGrow(a_Chunk, a_RelX, a_RelY, a_RelZ);
```
Seadragon91: No need for the light variables, is checked in `CanGrow`.

src/BlockEntities/BrewingstandEntity.cpp
```diff
@@ -92,10 +92,10 @@ bool cBrewingstandEntity::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		return false;
 	}
 
-	const cBrewingRecipes::cRecipe * Recipe = nullptr;
 	// The necessary brewing time, has been reached
 	if (m_TimeBrewed >= m_NeedBrewingTime)
 	{
+		const cBrewingRecipes::cRecipe * Recipe = nullptr;
```
Seadragon91: This one was my fault...

Seadragon91:
Question about the `Rnd` [variables](https://github.com/cuberite/cuberite/blob/master/src/Generating/StructGen.cpp#L333#L335)

The variables `Rnd` aren't used in the for loop, could it be possible that `Rnd` should be [`BubbleRnd`](https://github.com/cuberite/cuberite/blob/master/src/Generating/StructGen.cpp#L329)?

src/PolarSSL++/BlockingSslClientSocket.cpp
madmaxoft: Actually this is wrong. `Data` is the correct variable to use - the `WritePlain` method could consume less than `a_NumBytes` bytes, in which case the loop continues, but only the remaining portion of the data is to be sent (line 202).

madmaxoft:
Good catch, it should indeed be `BubbleRnd`

Seadragon91:
Squashed the commits, Anyone against a merge?

The static analyzer I used, also reports the (possible) use of uninitialized variables, should I open a own issue for that?

madmaxoft:
Looks good to me.

SafwatHalaby:
Out of curiosity, which static analyzer did you use?

Seadragon91:
I used [CppCheck](http://cppcheck.sourceforge.net/)

worktycho:
If your looking static analyzers, you might want to request access to the coverity scan system https://scan.coverity.com/projects/mc-server-mcserver

Seadragon91:
Thank you, have requested access to it. Let's hunt bugs :)

PR #2764, opened 2015-12-17
Implemented new method for creating a chat message

Seadragon91:
Do not merge yet!

This is a start to implement a new way of creating a chat message.

Questions:
* ~~Should the constructor from cChatMessageBuilder, accepts a text for start~~
* Should achievement part be added
* SetColor/SetStyle or better only with cChatColor

Todos:
* ~~A check for the size of m_Parts should be added, to avoid crash if no text has been appended. Where in the code?~~
* ~~Add method BroadcastChat in cRoot/cWorld for cChatMessageBuilder~~


Seadragon91:
Currently for m_Parts, a crash would occur. If the chat message is build this way.
```
local cm = cChatMessageBuilder()
cm:SetClickEvent("suggest_command", "/help")
```

The size of m_Parts is 0 here. My question is where should we add the check? In [`ManualBindings.cpp`](https://github.com/cuberite/cuberite/blob/chat_message/src/Bindings/ManualBindings.cpp) or in [`ChatMessageBuilder.cpp`](https://github.com/cuberite/cuberite/blob/chat_message/src/ChatMessageBuilder.cpp)

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -672,6 +672,69 @@ end</pre>
 				},
 			},  -- AdditionalInfo
 		},  -- cCompositeChat
+		
+		cChatMessageBuilder =
+		{
+			Desc = [[
+				Encapsulates a chat message that can contain various formatting, URLs, commands executed on click,
```
madmaxoft: `Encapsulates` -> `Builds`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -672,6 +672,69 @@ end</pre>
...
+				commands suggested on click, text inserted on shift click and text to show if the mouse is over text. The chat
+				message can be sent by the regular chat-sending functions,
+				{{cPlayer}}:SendMessage(), {{cWorld}}:BroadcastChat() and {{cRoot}}:BroadcastChat().</p>
+				<p>
+				Note that most of the functions in this class are so-called chaining modifiers - they modify the
+				object and then return the object itself, so that they can be chained one after another. See the
+				Chaining example below for details.</p>
+				<p>
+				Colors and Style can be directly used in the text with {cChatColor|cChatColor}. See the Chaining example.
```
madmaxoft: Double-braces are needed for proper links.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -672,6 +672,69 @@ end</pre>
...
+	-- Send an example chat message to the player:
+		a_Player:SendMessage(cChatMessageBuilder()
+		:AppendPart("Hello, " .. cChatColor.Green .. cChatColor.Underlined .. a_Player:GetName())
+		:SetClickEvent("open_url", "http://www.cuberite.org")
+		:AppendPart(cChatColor.Underlined .. ", and welcome.")
+		:SetClickEvent("suggest_command", "/help")
+		:AppendPart(" SetDay")
+		:SetClickEvent("run_command", "/time set 0")
+		:SetMessageType(mtJoin)
+	)
```
madmaxoft: I'd change the indent this way:
```lua
	-- Send an example chat message to the player:
	a_Player:SendMessage(cChatMessageBuilder()
		:AppendPart("Hello, " .. cChatColor.Green .. cChatColor.Underlined .. a_Player:GetName())
			:SetClickEvent("open_url", "http://www.cuberite.org")
		:AppendPart(cChatColor.Underlined .. ", and welcome.")
			:SetClickEvent("suggest_command", "/help")
		:AppendPart(" SetDay")
			:SetClickEvent("run_command", "/time set 0")
			:SetMessageType(mtJoin)
	)
```
It makes it clear what is a new part and what is the formatting for it.

src/Bindings/ManualBindings.cpp
```diff
@@ -2968,6 +2969,182 @@ static int tolua_cBlockArea_SaveToSchematicString(lua_State * tolua_S)
...
+	if (self == nullptr)
+	{
+		tolua_error(tolua_S, "invalid 'self' in function 'cChatMessageBuilder:SetClickEvent'", nullptr);
+		return 0;
+	}
+
+	// Add the part:
+	AString Action, Text;
+	L.GetStackValue(2, Action);
+	L.GetStackValue(3, Text);
```
madmaxoft: -> `L.GetStackValues(2, Action, Text);`

src/Bindings/ManualBindings.cpp
```diff
@@ -2968,6 +2969,182 @@ static int tolua_cBlockArea_SaveToSchematicString(lua_State * tolua_S)
...
+	if (self == nullptr)
+	{
+		tolua_error(tolua_S, "invalid 'self' in function 'cChatMessageBuilder:SetHoverEvent'", nullptr);
+		return 0;
+	}
+
+	// Add the part:
+	AString Action, Text;
+	L.GetStackValue(2, Action);
+	L.GetStackValue(3, Text);
```
madmaxoft: Again, combine into a single call.

src/Bindings/ManualBindings.cpp
```diff
@@ -2968,6 +2969,182 @@ static int tolua_cBlockArea_SaveToSchematicString(lua_State * tolua_S)
...
+		return 0;
+	}
+	cChatMessageBuilder * self = reinterpret_cast<cChatMessageBuilder *>(tolua_tousertype(tolua_S, 1, nullptr));
+	if (self == nullptr)
+	{
+		tolua_error(tolua_S, "invalid 'self' in function 'cChatMessageBuilder:SetClickEvent'", nullptr);
+		return 0;
+	}
+
+	// Add the part:
```
madmaxoft: Wrong comment (copypasta)

src/Bindings/ManualBindings.cpp
```diff
@@ -2968,6 +2969,182 @@ static int tolua_cBlockArea_SaveToSchematicString(lua_State * tolua_S)
...
+		return 0;
+	}
+	cChatMessageBuilder * self = reinterpret_cast<cChatMessageBuilder *>(tolua_tousertype(tolua_S, 1, nullptr));
+	if (self == nullptr)
+	{
+		tolua_error(tolua_S, "invalid 'self' in function 'cChatMessageBuilder:SetHoverEvent'", nullptr);
+		return 0;
+	}
+
+	// Add the part:
```
madmaxoft: Copypasta comment

src/Bindings/ManualBindings.cpp
```diff
@@ -2968,6 +2969,182 @@ static int tolua_cBlockArea_SaveToSchematicString(lua_State * tolua_S)
...
+{
+	// function cChatMessageBuilder:AppendPart(Message)
+	// Exported manually to support call-chaining (return *this)
+
+	// Check params:
+	cLuaState L(tolua_S);
+	if (
+		!L.CheckParamUserType(1, "cChatMessageBuilder") ||
+		!L.CheckParamString(2)
+		)
```
madmaxoft: Wrong indent, the closing parenthesis should be one level less. Same in all the other binding functions.

src/Bindings/ManualBindings.cpp
```diff
@@ -3261,6 +3438,14 @@ void cManualBindings::Bind(lua_State * tolua_S)
 			tolua_function(tolua_S, "SendPluginMessage", tolua_cClientHandle_SendPluginMessage);
 		tolua_endmodule(tolua_S);
 
+		tolua_beginmodule(tolua_S, "cChatMessageBuilder");
+			tolua_function(tolua_S, "AppendPart", tolua_cChatMessageBuilder_AppendPart);
+			tolua_function(tolua_S, "SetClickEvent", tolua_cChatMessageBuilder_SetClickEvent);
+			tolua_function(tolua_S, "SetHoverEvent", tolua_cChatMessageBuilder_SetHoverEvent);
+			tolua_function(tolua_S, "SetInsertionText", tolua_cChatMessageBuilder_SetInsertionText);
+			tolua_function(tolua_S, "SetMessageType", tolua_cChatMessageBuilder_SetMessageType);
```
madmaxoft: Align the lines vertically - `tolua_cChatMessageBuilder_` in the same column

src/ChatMessageBuilder.cpp
```diff
@@ -0,0 +1,116 @@
...
+	m_Parts.push_back(new cChatMessageBuilder::cChatMessagePart(a_Text));
+}
+
+
+
+
+
+void cChatMessageBuilder::SetClickEvent(const AString & a_Action, const AString & a_Text)
+{
+	cChatMessagePart * Part = m_Parts.back();
```
madmaxoft: Use `auto`

src/ClassDiagram.cd
```diff
@@ -0,0 +1,2 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<ClassDiagram />
```
madmaxoft: Remove this file

src/ChatMessageBuilder.h
```diff
@@ -0,0 +1,90 @@
...
+	eMessageType GetMessageType(void) const { return m_MessageType; }
+
+	/** Returns additional data pertaining to message type, for example, the name of a mtPrivateMsg sender */
+	AString GetAdditionalMessageTypeData(void) const { return m_AdditionalMessageTypeData; }
+
+	/** Sets the message type, which is indicated by prefixes added to the message when serializing
+	Takes optional AdditionalMessageTypeData to set m_AdditionalMessageTypeData. See said variable for more documentation. */
+	void SetMessageType(eMessageType a_MessageType, const AString & a_AdditionalMessageTypeData = "");
+
+	// tolua_end
```
madmaxoft: This `tolua_end` should be moved just after `Clear()`, because the functions are exported manually.

src/ChatMessageBuilder.h
```diff
@@ -0,0 +1,90 @@
...
+
+		AString m_Text;
+		AString m_InsertionText;
+		AString m_HoverText;
+		AString m_ClickText;
+		AString m_HoverAction;
+		AString m_ClickAction;
+	} ;
+
+	std::vector<cChatMessagePart *> m_Parts;
```
madmaxoft: Use a `unique_ptr` instead of naked pointers. typedef the ptr first.

madmaxoft:
Have the constructor take one optional parameter, the initial text; use that to add the first part. That way, there's always at least one part in the builder, and no need for a check, really.

tigerw:
"new method"

Does that mean there will now be two methods of creating a chat message, with the old one kept for reasons of 'backwards compatibility'?

NiLSPACE:
I think together with this we should also allow a method to send raw json to the client.

madmaxoft:
Yes, there will be now three methods:
- sending a plain string
- sending a `cCompositeChat`
- (new) sending a `cChatMessageBuilder`

We'll most likely phase the `cCompositeChat` out in the future, because this new builder is superior. We'll keep the plain string in order to support simple messages.

madmaxoft:
I thought we already had the raw json sending, but can't find it now. Must've been dreaming :P

Seadragon91:
Newer did much work with typedef. Have currently this.
```
typedef std::unique_ptr<cChatMessagePart> cChatMessagePartPrt;
typedef std::vector<cChatMessagePartPrt> m_Parts;
```

madmaxoft:
Fix `Prt` to `Ptr` and it's good :)

Seadragon91:
Sorry, this was a mistype from me here.

madmaxoft:
Right, I overlooked that. It should be:
```cpp
typedef std::unique_ptr<cChatMessagePart> cChatMessagePartPtr;
typedef std::vector<cChatMessagePartPtr> cChatMessagePartPtrs;

cChatMessagePartPtrs m_Parts;
```

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -672,6 +672,69 @@ end</pre>
...
+function OnPlayerJoined(a_Player)
+	-- Send an example chat message to the player:
+		a_Player:SendMessage(cChatMessageBuilder()
+			:AppendPart("Hello, " .. cChatColor.Green .. cChatColor.Underlined .. a_Player:GetName())
+				:SetClickEvent("open_url", "http://www.cuberite.org")
+			:AppendPart(cChatColor.Underlined .. ", and welcome.")
+				:SetClickEvent("suggest_command", "/help")
+			:AppendPart(" SetDay")
+				:SetClickEvent("run_command", "/time set 0")
+				:SetMessageType(mtJoin)
```
madmaxoft: I still think this entire thing, line 724 - 731, has one more indent level than necessary - the `a_Player:SendMessage` line should be the same indent level as the comment above it.

src/Bindings/ManualBindings.cpp
```diff
@@ -2968,6 +2969,180 @@ static int tolua_cBlockArea_SaveToSchematicString(lua_State * tolua_S)
...
+static int tolua_cChatMessageBuilder_AppendPart(lua_State * tolua_S)
+{
+	// function cChatMessageBuilder:AppendPart(Message)
+	// Exported manually to support call-chaining (return *this)
+
+	// Check params:
+	cLuaState L(tolua_S);
+	if (
+		!L.CheckParamUserType(1, "cChatMessageBuilder") ||
+		!L.CheckParamString(2)
```
madmaxoft: Check param end, too, to avoid weirdnes when calling with more params than needed.

src/ChatMessageBuilder.cpp
```diff
@@ -0,0 +1,115 @@
...
+AString cChatMessageBuilder::CreateJsonString(bool a_ShouldUseChatPrefixes) const
+{
+	Json::Value msg;
+	msg["text"] = cClientHandle::FormatMessageType(a_ShouldUseChatPrefixes, GetMessageType(), GetAdditionalMessageTypeData());  // The client crashes without this field being present
+	for (auto ChatMessagePart : m_Parts)
+	{
+		Json::Value Part;
+		Part["text"] = ChatMessagePart->m_Text;
+
+		if (ChatMessagePart->m_ClickAction != "")
```
madmaxoft: Use `!str.empty()` instead of `str != ""`, it's a tiny bit faster.

src/ChatMessageBuilder.h
```diff
@@ -0,0 +1,107 @@
...
+{
+public:
+	// tolua_end
+
+	/** This class is used to store the text, the click event, hover event and insertion text. */
+	class cChatMessagePart
+	{
+	public:
+		cChatMessagePart() :
+			m_Text("")
```
madmaxoft: No need to initialize to an empty string

src/Bindings/ManualBindings.cpp
Seadragon91: To the other methods, too?

madmaxoft: Yes, all of them.

Seadragon91:
Looks much better now :)
Adding now the overloads in `BroadcastChat`.

src/ChatMessageBuilder.cpp
```diff
@@ -0,0 +1,115 @@
...
+			Json::Value Hover;
+			Hover["action"] = ChatMessagePart->m_HoverAction;
+			Hover["value"] = ChatMessagePart->m_HoverText;
+			Part["hoverEvent"] = Hover;
+		}
+		if (!ChatMessagePart->m_HoverAction.empty())
+		{
+			Part["insertion"] = ChatMessagePart->m_InsertionText;
+		}
+		msg["extra"].append(Part);
```
madmaxoft: Would `std::move(Part)` work here? It'd make sense.

src/ChatMessageBuilder.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+
+
+
+void cChatMessageBuilder::Clear(void)
+{
+	m_Parts.clear();
+
+	// Add a empty part
+	m_Parts.push_back(std::move(cChatMessagePartPtr(new cChatMessagePart(""))));
```
madmaxoft: I think `cpp14::make_unique<cChatMessagePart>("")` would be better and shouldn't need moving.

src/ChatMessageBuilder.h
```diff
@@ -0,0 +1,107 @@
...
+		AString m_HoverText;
+		AString m_ClickText;
+		AString m_HoverAction;
+		AString m_ClickAction;
+	} ;
+
+	typedef std::unique_ptr<cChatMessagePart> cChatMessagePartPtr;
+	typedef std::vector<cChatMessagePartPtr> cParts;
+
+	cParts m_Parts;
```
madmaxoft: This should probably be `protected`, too. And you should add a note that it is never empty, so that none of the modifiers need to check.

madmaxoft:
I think colors, styles and scores should be included in the API. I'd like to phase the old-style colors out completely, it is quite possible the vanilla client / server will phase them out in the future, too.

Basically include everything the JSON message can contain:
http://wiki.vg/Chat

madmaxoft:
I've rebased this onto current master.

Seadragon91:
This needs more tough. For the hover part, we need color and style, too. That means we need the message part for it.

madmaxoft:
Yea, I just didn't want this to become too old to work on.

Seadragon91:
In the example from `http://wiki.vg/Chat`, is no extra part for hover event. I tested this and the extra part is not recognized. The color / style only works for the text part.

Issue #2263, opened 2015-06-18
Deadlock


Seadragon91:
Was flying around in the world, lots of chunks has been generated, did something else for a few minutes, coming back and seeing that the server crashed, deadlock.

```
Currently logging to "gdb.txt".
Logs will be appended to the log file.
Output is being logged and displayed.
Starting program: /home/lukas/games/MCServer_brewing/MCServer 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff6b51700 (LWP 30905)]
[New Thread 0x7ffff6350700 (LWP 30906)]
[New Thread 0x7ffff53d4700 (LWP 30907)]
[New Thread 0x7ffff4bd3700 (LWP 30908)]
[New Thread 0x7ffff43d2700 (LWP 30909)]
[New Thread 0x7ffff33cc700 (LWP 30910)]
[New Thread 0x7ffff2bcb700 (LWP 30911)]
[New Thread 0x7ffff23ca700 (LWP 30912)]
[New Thread 0x7ffff1bc9700 (LWP 30913)]
[New Thread 0x7ffff13c8700 (LWP 30914)]
[New Thread 0x7ffff0bc7700 (LWP 30915)]

Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff1bc9700 (LWP 30913)]
0x00007ffff6b88cc9 in raise () from /lib/x86_64-linux-gnu/libc.so.6
  Id   Target Id         Frame 
  12   Thread 0x7ffff0bc7700 (LWP 30915) "MCServer" 0x00007ffff6c3d82d in read
    () from /lib/x86_64-linux-gnu/libc.so.6
  11   Thread 0x7ffff13c8700 (LWP 30914) "MCServer" 0x00007ffff74b9b9d in nanosleep () from /lib/x86_64-linux-gnu/libpthread.so.0
* 10   Thread 0x7ffff1bc9700 (LWP 30913) "MCServer" 0x00007ffff6b88cc9 in raise
    () from /lib/x86_64-linux-gnu/libc.so.6
  9    Thread 0x7ffff23ca700 (LWP 30912) "MCServer" 0x000000000052bab3 in cListAllocationPool<cChunkData::sChunkSection, 1600ul>::Free(cChunkData::sChunkSection*) ()
[...]

PR #2765, opened 2015-12-17
Changed the deadlock detector interval to 120 seconds.Closes #2550

### LINKED ISSUE #2550: Disable deadlock detector by default.
It is now far more common for a detected deadlock to be caused by a long computation on the tick thread causing lag, then it is from an actual deadlock. We should disable the detector by default as the server can recover in these situations, if it is given long enough to finish its computation.

Seadragon91:
Changed the default deadlock detector interval to 120 seconds.

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -193,7 +193,7 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	if (settingsRepo->GetValueSetB("DeadlockDetect", "Enabled", true))
 	{
 		LOGD("Starting deadlock detector...");
-		dd.Start(settingsRepo->GetValueSetI("DeadlockDetect", "IntervalSec", 20));
+		dd.Start(settingsRepo->GetValueSetI("DeadlockDetect", "IntervalSec", 120));
 	}
 
 	settingsRepo->Flush();


Seadragon91:
If 120 seconds are okay, please merge.

madmaxoft:
I'm not too happy about this change, 120 seconds per tick is really bad, but what can I do.

SafwatHalaby:
What kind of computations that long in the first place?

Seadragon91:
Currently is the problem, that the nether world will be generated if you go into the portal. On a raspberry pi, it takes longer than the 20 seconds, then the deadlock detector assumes freeze and aborts the server.

madmaxoft:
So we're fixing the result, instead of the cause...

Seadragon91:
Another problem is if you fly trough the world, where it also can deadlock, issue #2263

Seadragon91:
I know this is only a workaround...

SafwatHalaby:
Assuming the only computation-heavy thing on the tick thread is world generation, the idea propose at #2810 is a better solution than this PR.

SafwatHalaby:
The idea proposed at #2810 is good mitigation for the issue. (Though it won't solve the flight case).

Edit: rewording.

SafwatHalaby:
Just some proposals:
 - Solve traveling through portals issue via #2810
- When the chunk generator is overloaded (due to flight / fast movement), prevent players from venturing further into ungenerated areas by creating a temporarily invisible wall.

SafwatHalaby:
Just some proposals:
- Solve traveling through portals issue via #2810
- When the chunk generator is overloaded (due to flight / fast movement), prevent players from venturing further into ungenerated areas by creating a temporary invisible wall. 

This should eliminate the need for this PR and fix the actual issues rather than the symptoms.

PR #2769, opened 2015-12-17
Check for invalid tags when reading a compound tag

worktycho:


madmaxoft:
Looks good. Will you be doing anything more with this, or is this finished?

worktycho:
That depends if my fuzzing finds any more crashes. But it hasn't found any crashes in the last four hours of running, so I don't think so.

madmaxoft:
Wow, you're bringing some heavy guns in :)

worktycho:
I've been wanting to try fuzzing something for a while, and the NBT parser turned out to be easy to split. I'll upload the NBT validation program I wrote that I'm actually fuzzing over the weekend. It returns a zero exit code for valid NBT, and non-zero for invalid.

Seadragon91:
Thank you doing that :smile:

worktycho:
If theres any other parsers that are easy to split, I'll try and fuzz them. I'd like to fuzz the protocol, but I'm not sure I can split it into a separate small program.

madmaxoft:
Would it make sense to put the test programs into the `tests` folder and add it to the unit testing?

worktycho:
It might do. Just take all the testcases the fuzzer has generated, and test the parser against them to ensure it doesn't crash.

madmaxoft:
I rebased this thing on top of current master, merge if it compiles.

PR #2770, opened 2015-12-18
Removed unused class cChunkDataCollector.

madmaxoft:
Not needed for anything, and generating some (useless) MSVC warnings.

### DIFF
--- a/src/ChunkDataCallback.h
+++ b/src/ChunkDataCallback.h
@@ -52,26 +52,6 @@ public:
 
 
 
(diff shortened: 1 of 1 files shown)


SphinxC0re:
You should really wait for the builds to finish

madmaxoft:
I waited for Travis; I knew Appveyor would work because I'm working on MSVC as well; the failure is only because the build was started later than the PR was merged.

SphinxC0re:
Are you using the same *version* of MSVC as Appveyor?

madmaxoft:
For a change like this, don't care :)

PR #2774, opened 2015-12-18
Fix off by two error in ReadString

worktycho:


madmaxoft:
Merge after it builds.

PR #2776, opened 2015-12-18
Use atomic bool for isthread terminate

### LINKED ISSUE #2636: cIsThread::m_ShouldTerminate synchronisation

worktycho:
Fixes #2636
Closes #2637


madmaxoft:
Rebased onto master, merge if it compiles.

PR #2779, opened 2015-12-19
Fixes for clang-3.7

Seadragon91:


worktycho:
Merge if complies

madmaxoft:
Please wait for me, I'd like to see this first. (github won't work on my phone)
-- 

worktycho:
Its just removing some unnecessary std::moves for things that are implicitly convertible to rvalues, a c-style case to static_cast conversion, and using a cont reference rather than a copy for iterating through an array of vectors.

PR #2780, opened 2015-12-19
Fix race condition

worktycho:
Fixes a race condition from incrementing the number of chunks prepared on multiple threads.


madmaxoft:
Is that really a race condition if it's just in a (debug) logging?

worktycho:
Its potential UB. So it should really be fixed.

madmaxoft:
I was kinda hoping that this would fix the deadlock we're sometimes getting in the CI builds.

Anyway, rebased onto master, merge if it compiles.

PR #2783, opened 2015-12-20
Fixed spawn point generation. Wrong number for divisions and wrong highest spawn point

### LINKED ISSUE #2762: Spawnpoint
Hi,

no matter how often I create a new world, I always get the error "Did not find an acceptable spawnpoint. Generated a randowm spawnpoint position..". I am using the latest Windows version of Cuberite. Can you help me out please?

### LINKED ISSUE #2781: Spawn point generation not working correctly
This is about the spawn generation, at startup of the server.

~~I debugged this and here [GetHeight()](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L676) returns always 0. The chunk is not null and is valid. Here [GetHeight()](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L653) returns the correct y coordinate.~~

Edit: GetHeight works okay.

Seadragon91:
Corrected this, works good now :)
* Fixed wrong number, used for division
* Increased highest spawn point by two, because of body part and head

Fixes #2781 
Closes #2762

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -613,7 +613,7 @@ void cWorld::GenerateRandomSpawn(int a_MaxSpawnRadius)
 	}
 
 	// A search grid (searches clockwise around the origin)
-	static const int HalfChunk = static_cast<int>(cChunkDef::Width / 0.5f);
+	static const int HalfChunk = static_cast<int>(cChunkDef::Width / 2.0f);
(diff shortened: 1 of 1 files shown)


Seadragon91:
I had added block stone, but removed it. Don't want to spawn in a cave :)

Issue #2771, opened 2015-12-18
Implement path truncation


SafwatHalaby:
Occasionally, mobs will inevitably skip ahead of their path, skipping one waypoint but landing on another one. Currently, when this happens, mobs will go back to the waypoint they skipped, and resume their path. This looks weird, and is especially visible in chickens: After they jump, they'll always turn around before continuing to their destination.

Path truncation, suggested by @Gargaj, is an easy way to resolve this.
Pseudocode:

```
var currentWaypoint = getNextWaypoint();
var nextWaypoint = getNextWaypoint();
while (distanceTo(nextWaypoint) < distanceTo(currentWaypoint))
{
    currentWaypoint = nextWaypoint;
    nextWayPoint = getNextWaypoint();
}
```

madmaxoft:
The distance evaluation might get a bit too CPU-intensive, if calculated over and over again.

worktycho:
You can compare the squared distance to avoid square roots. Then its only 4 multiplications, two adds and a compare.

SafwatHalaby:
The distance evaluation is O(1), in 99% of the cases it'll not even enter the while loop. In some cases it'll loop once or twice.

SafwatHalaby:
More precisely: In normal circumnstances it won't enter the loop. But if you push a mob forward into its path, it'll enter the loop X times, where X is the amount of waypoints the mob skipped. It'll enter the loop once or twice when mobs jump and skip a waypoint. It'll enter it several times when a chicken hovers and skips some waypoints.

madmaxoft:
Then it won't fix cases such as trying to walk on a ledge towards a staircase down and falling down by accident instead.

SafwatHalaby:
Why wouldn't it? In such a case it would loop several times. And truncate old waypoints.

I should probably cap it to 3 or 4. Because if more waypoints were missed, a full recalc is probably needed.

SafwatHalaby:
Note that if the mob falls too far off and is unable to proceed with the old path, cPathFinder would trigger a recalc.

PR #2788, opened 2015-12-21
PF - Improved mob jumping

### LINKED ISSUE #2646: Allow mobs to jump down up to 3 or 4 blocks
I have built a few grinders on my skyblock server, but although the mobs spawn normally, they never walk off the edge into the water unless pushed by other mobs.  Spiders are an exception, and seem to walk over the edge almost as the first thing they do after spawning.  I would say that this is an error in the 4 block entity-jump-limit, but mobs that spawn on top of the grinder seem to be more than willing to jump off the top and make feeble attempts at trying to ambush me. :P

### LINKED ISSUE #2787: PF - Mob bugs when going up a block
A mob will ignore the upper green block and manage to get inside the glass structure. I'm reorganizing path.cpp, that should fix this and #2646

SafwatHalaby:
This PR drastically improves mob jumping.
Fixes #2787.
Closes #2646.

Makes the problems discussed in #2771 less severe. Specifically, chickens flight is now better handled, and zombies will not weirdly look back after going down a block.

Mobs no longer act weird when the player jumps.

`StepOnce` is now more readable.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -42,7 +42,7 @@ cPath::cPath(
 	// TODO: if src not walkable OR dest not walkable, then abort.
 	// Borrow a new "isWalkable" from ProcessIfWalkable, make ProcessIfWalkable also call isWalkable
 
-	a_BoundingBoxWidth = 1;  // Until we improve physics, if ever.
+	a_BoundingBoxWidth = 1;  // Treat all mobs width as 1 until physics is improved. This would also require changes to stepOnce to work.
 
 	m_BoundingBoxWidth = CeilC(a_BoundingBoxWidth);
 	m_BoundingBoxHeight = CeilC(a_BoundingBoxHeight);
@@ -57,7 +57,7 @@ cPath::cPath(
 	m_Destination.y = FloorC(a_EndingPoint.y);
 	m_Destination.z = FloorC(a_EndingPoint.z - HalfWidthInt);
 
-	if (GetCell(m_Source)->m_IsSolid || GetCell(m_Destination)->m_IsSolid)
+	if (!IsWalkable(m_Source))
 	{
 		m_Status = ePathFinderStatus::PATH_NOT_FOUND;
 		return;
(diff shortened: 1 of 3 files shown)


src/Mobs/Path.cpp
```diff
@@ -57,7 +57,7 @@ cPath::cPath(
 	m_Destination.y = FloorC(a_EndingPoint.y);
 	m_Destination.z = FloorC(a_EndingPoint.z - HalfWidthInt);
 
-	if (GetCell(m_Source)->m_IsSolid || GetCell(m_Destination)->m_IsSolid)
+	if (!IsWalkable(m_Source))
```
SafwatHalaby: Explanation: The mob should pathfind even if target is not reachable because the pathfinder can pick a nearby spot.

worktycho:
Looks good. Merge if CI.

src/Mobs/Path.cpp
NiLSPACE: Isn't it better to add that in the source?

Seadragon91:
Looks like AppVeyor, has problems.

SafwatHalaby:
Leap of faith. Report of Windows has trouble compiling this (it shouldn't).

src/Mobs/Path.cpp
SafwatHalaby: Mind explaining?

src/Mobs/Path.cpp
```diff
@@ -207,46 +206,103 @@ bool cPath::StepOnce()
...
+	// Now we start checking adjacent cells.
 
-	// Check diagonals on XY plane.
-	// x = -1: west, x = 1: east.
-	for (int x = -1; x <= 1; x += 2)
+
+	bool done_east = false,
+	done_west = false,
+	done_north = false,
+	done_south = false;  // If true, no need to do more checks in that direction
```
SafwatHalaby: Proper CamelCase needed here. Will do that in my followup PR.

PR #2789, opened 2015-12-21
Added analytics to APIDump

bearbin:


madmaxoft:
hmm, NO!

I don't want any tracking in my project

bearbin:
Huh? How come the main page needs tracking but not the API docs?

madmaxoft:
There's tracking in the main page, too? Get rid of it

bearbin:
Why have you changed your mind? You could have objected when I asked for the DNS, when you logged in to check it out, when I submitted the PR for the main site, etc.

The tracking really isn't that objectionable, and helps users by allowing us to improve our websites' designs. It's not like Google Analytics or ad networks, and all user data is anonymised.

bearbin:
If you wanted we could also provide an opt-out mechanism for users who don't want to be tracked.

36451:
Afair, piwik has an opt-out page. 
All you have to do, is link to it.

madmaxoft:
I'm very much against active tracking. But I don't have anything against passive tracking. With the domain, I thought you had access to server logs from cuberite.org, which you could parse for the information.

Opt-out is not good enough. I don't want us to be even technically able to track.

bearbin:
What is the aspect of the "active" tracking that you dislike? What this is doing is functionally identical to the server log parsing, no meaningful extra information is collected from the browser.

This is a full list of data that it would collect (I emboldened the ones not collected by server logs): 

 * Anonymised IP Address
 * Date and time of the request
 * URL of the page being viewed (Page URL)
 * URL of the page that was viewed prior to the current page (Referrer URL)
 * **Screen resolution being used**
 * Time in local user’s timezone
 * **Links to an outside domain that were clicked (Outlinks)**
 * **Pages generation time (the time it takes for webpages to be generated by the webserver and then downloaded by the user: Page speed)**
 * Main Language of the browser being used (Accept-Language header) (location is derived from this)
 * User Agent of the browser being used (User-Agent header)

madmaxoft:
It's not about what's there now, the problem I have with this is that it's really easy to add any kind of tracking later on, without anyone else knowing. Not to mention malware exploits, if they pwnd the analytics domain webserver, they'll serve malicious analytics to all of our pages.

Remember Battlestar Galactica's commander Adama and his decision to limit technology interconnectivity? That's me :)
Yes, I know, I'm paranoid, it's one of the side-effects of my job :P

bearbin:
Well, some level of paranoia is always sensible :p

What if instead of the page downloading the code from a remote server and executing it, the piwik code is embedded directly into the page and it can be audited to make sure that it is only doing what we tell it to. (And a github commit replacing the analytics might be more noticeable than a hack of the server).

madmaxoft:
That could be workable. Store the JS files in the Static subfolder, reference it in APIDesc so that the plugin copies it all into the dest folder, and use it using a relative path. Similar handling is already used for the syntax highlighting JS code.

bearbin:
Thank you. I will work on this and see if it is possible.

bearbin:
I think that it is mostly working, but I can't figure out why piwik.js is not copied along with the rest of the files. Could anybody help?

madmaxoft:
It should be copied, lines 1231 - 1237 in `main_APIDump.lua` should take care of that. Do you have the file permissions right (assuming Linux)?
If all else fails, add a logging statement to show each file being copied in that loop, and check if the file is found and if it's copied.

bearbin:
I realised what the issue was and I think I've fixed it.

madmaxoft:
That should work. Can you rebase and squash?

PR #2790, opened 2015-12-21
Add handling of water bottle

Gargaj:
Currently using an empty bottle on a water source block creates a phantom water bottle that disappears on inventory sync.

I copied the ```GetBlockFromTrace``` function from ```ItemBucket.h``` - probably should refactor it out into a ```World``` function or somewhere else.

### DIFF
--- a/src/Items/ItemBottle.h
+++ b/src/Items/ItemBottle.h
@@ -1,0 +1,95 @@
+
+#pragma once
+
+#include "ItemHandler.h"
+#include "../World.h"
+
+
+
+
+
+class cItemBottleHandler :
+	public cItemHandler
+{
+public:
+	cItemBottleHandler() :
+		cItemHandler(E_ITEM_GLASS_BOTTLE)
+	{
+	}
+
+
+	bool GetBlockFromTrace(cWorld * a_World, cPlayer * a_Player, Vector3i & a_BlockPos)
+	{
+		class cCallbacks :
+			public cBlockTracer::cCallbacks
+		{
+		public:
+			Vector3i m_Pos;
+			bool     m_HasHitFluid;
+
+
+			cCallbacks(void) :
+				m_HasHitFluid(false)
+			{
+			}
+
+			virtual bool OnNextBlock(int a_BlockX, int a_BlockY, int a_BlockZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, char a_EntryFace) override
+			{
+				if (IsBlockWater(a_BlockType))
+				{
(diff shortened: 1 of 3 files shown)


src/Items/ItemBottle.h
```diff
@@ -0,0 +1,107 @@
...
+
+		Vector3i BlockPos;
+		if (!GetBlockFromTrace(a_World, a_Player, BlockPos))
+		{
+			return false;  // Nothing in range.
+		}
+
+		if (a_World->GetBlockMeta(BlockPos.x, BlockPos.y, BlockPos.z) != 0)
+		{
+			// Not a source block
```
worktycho: What should happen if a player is looking at a source block through a non-source block? Because this condition is never true, as the block trace will only return a source block position.

src/Items/ItemBottle.h
```diff
@@ -0,0 +1,107 @@
...
+			return false;
+		}
+
+		BLOCKTYPE Block = a_World->GetBlock(BlockPos.x, BlockPos.y, BlockPos.z);
+
+		if (IsBlockWater(Block))
```
worktycho: This condition will only be false in rare race conditions, you can probably remove it.

src/Items/ItemBottle.h
Gargaj: Unsure what's the behaviour in vanilla, should check.

Gargaj: According to my tests it's the correct behaviour: clicking through non source blocks if there's a source block behind it still works to fill a bottle. Now, on that thought, the lava condition probably should be removed.

src/Items/ItemBottle.h
Gargaj: Actually no, see my lava comment above - but I'll fix it in a second.

Gargaj:
Updated the branch cf. tycho's comments, but haven't actually tested it in-game yet.

worktycho:
Merge if CI succeeds.

SafwatHalaby:
I think you have something wrong with your submodules.

Gargaj:
I'm pretty sure the OSX build failing isn't my fault :O

bearbin:
No, we're still getting it set up ATM.

SafwatHalaby:
I am talking about the change made in ` lib/cmake-coverage`. I think you'll need `git submodule update`.

Gargaj:
Good lord how I hate Git's submodule handling.

SafwatHalaby:
Compiles on Linux, and some light in game testing worked well. Appveyor and Travis succeeded, tycho approves. Merging.

PR #2791, opened 2015-12-21
cTeam:GetName returning wrong name

then0rTh:


worktycho:
Looks good. Merge if CI succeeds.

PR #2796, opened 2015-12-22
Added GeoTrust Root Certificate

worktycho:


madmaxoft:
Did Mojang change their signer again? Ugh.

Seems like you'll have to update CheckBasicStyle.lua so that it accepts the double-equal-sign in base64 encoding of the cert.

bearbin:
The build failing on macs is normal, until we can get the mac machine all up and running.

worktycho:
Fixed the checkbasicstyle script by ignoring ==\

worktycho:
Actually fixed now.

bearbin:
After this is pushed out, I will send a update newsletter.

bearbin:
Merge?

madmaxoft:
Looks good to me

PR #2798, opened 2015-12-22
Fast nbt test

worktycho:
Some tests for FastNBTReader


bearbin:
Travis failed :(

worktycho:
Forgot some untracked files

tests/WorldStorage/FastNBT/Parse.cpp
```diff
@@ -0,0 +1,39 @@
+#include "Globals.h"
+#include "WorldStorage/FastNBT.h"
+
+int main() {
```
madmaxoft: Brace on separate line (multiple occurences)

tests/WorldStorage/FastNBT/Parse.cpp
```diff
@@ -0,0 +1,39 @@
...
+		testassert(!nbt.IsValid());
+	}
+	{
+		cParsedNBT nbt{nullptr, 3};
+
+		testassert(!nbt.IsValid());
+	}
+
+	{
+		char data[] = {10, 0, 0, 0x7F, -0x7F};
```
madmaxoft: I'd prefer to have unsigned hex constants, it's rather mind-taxing having to convert the negative ones in the brain.

madmaxoft: Feel free to change the `cParsedNBT` constructor to accept `const void *` to avoid the need for casts.

worktycho: I'll change to const byte *.

madmaxoft: I think `const void *` is better for the interface, so that it can parse anything, and internally cast to whatever datatype is best.

worktycho: Its taking an array of bytes, not an untyped array. So it should be `Byte *`

madmaxoft: It's taking a pointer to a buffer containing data. There's not much sense in forcing the type of the pointer.

PR #2800, opened 2015-12-22
Removed offending tr1 references.

bearbin:
This should fix the mac build.

### DIFF
--- a/Tools/ProtoProxy/Globals.h
+++ b/Tools/ProtoProxy/Globals.h
@@ -12,26 +12,26 @@
 #if defined(_MSC_VER)
 	// MSVC produces warning C4481 on the override keyword usage, so disable the warning altogether
 	#pragma warning(disable:4481)
-	
+
 	// Disable some warnings that we don't care about:
 	#pragma warning(disable:4100)
 
 	#define OBSOLETE __declspec(deprecated)
-	
+
 	// No alignment needed in MSVC
 	#define ALIGN_8
 	#define ALIGN_16
-	
+
 	#define FORMATSTRING(formatIndex, va_argsIndex)
 
 #elif defined(__GNUC__)
 
 	// TODO: Can GCC explicitly mark classes as abstract (no instances can be created)?
 	#define abstract
-	
+
 	// TODO: Can GCC mark virtual methods as overriding (forcing them to have a virtual function of the same signature in the base class)
 	#define override
-	
+
 	#define OBSOLETE __attribute__((deprecated))
 
 	#define ALIGN_8 __attribute__((aligned(8)))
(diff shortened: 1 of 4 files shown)


Tools/ProtoProxy/Globals.h
```diff
@@ -223,13 +220,9 @@ typedef unsigned char Byte;
 #if (defined(_MSC_VER) && (_MSC_VER < 1600))
 	// MSVC before 2010 doesn't have std::shared_ptr, but has std::tr1::shared_ptr, defined in <memory> included earlier
 	#define SharedPtr std::tr1::shared_ptr
```
worktycho: Remove this, we don't build on MSVC < 2010 any more

Tools/QtBiomeVisualiser/Globals.h
```diff
@@ -335,13 +335,9 @@ void inline LOGERROR(const char* a_Format, ...)
 #if (defined(_MSC_VER) && (_MSC_VER < 1600))
 	// MSVC before 2010 doesn't have std::shared_ptr, but has std::tr1::shared_ptr, defined in <memory> included earlier
 	#define SharedPtr std::tr1::shared_ptr
```
worktycho: Remove this

Tools/ToLuaDoxy/Globals.h
```diff
@@ -43,10 +43,6 @@
 	#include <semaphore.h>
 	#include <errno.h>
 	#include <fcntl.h>
-	
-	#if !defined(ANDROID_NDK)
-		#include <tr1/memory>
```
worktycho: We need <memory> now.

Tools/ProtoProxy/Globals.h
```diff
@@ -12,26 +12,26 @@
...
 	#define FORMATSTRING(formatIndex, va_argsIndex)
 
 #elif defined(__GNUC__)
 
 	// TODO: Can GCC explicitly mark classes as abstract (no instances can be created)?
 	#define abstract
-	
+
 	// TODO: Can GCC mark virtual methods as overriding (forcing them to have a virtual function of the same signature in the base class)
 	#define override
```
SphinxC0re: isn't gcc totally capable of interpreting the override symbol?

worktycho: Thats ancient. I think it might cause an error in clang now to define this macro.

SphinxC0re: It's even in the standard: http://en.cppreference.com/w/cpp/language/override

Tools/ToLuaDoxy/Globals.h
bearbin: Memory is already included in that file, does it need to be included again?

worktycho: No. I just can see it.

bearbin:
The mac build is confirmed fixed.

PR #2801, opened 2015-12-22
Added a Json parser and serializer to Lua API.

madmaxoft:
This simplifies Lua table <-> Json conversions to oneliners. It is the basic requirement of accessing or implementing a JSON-based web API through plugins.

### DIFF
--- a/src/Bindings/LuaJson.cpp
+++ b/src/Bindings/LuaJson.cpp
@@ -1,0 +1,315 @@
+
+// LuaJson.cpp
+
+// Implements the Json exposure bindings to Lua
+
+#include "Globals.h"
+#include <sstream>
+#include "LuaJson.h"
+#include "LuaState.h"
+#include "tolua++/include/tolua++.h"
+#include "json/json.h"
+
+
+
+
+
+// fwd:
+
+static void PushJsonValue(const Json::Value & a_Value, cLuaState & a_LuaState);
+static Json::Value JsonSerializeValue(cLuaState & a_LuaState);
+
+
+
+
+
+/** Pushes the specified Json array as a table on top of the specified Lua state.
+Assumes that a_Value is an array. */
+static void PushJsonArray(const Json::Value & a_Value, cLuaState & a_LuaState)
+{
+	// Create the appropriately-sized Lua table:
+	lua_createtable(a_LuaState, static_cast<int>(a_Value.size()), 0);
+
+	// Insert each value to the appropriate index (1-based):
+	int idx = 1;
+	for (const auto & v: a_Value)
+	{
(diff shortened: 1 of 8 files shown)


bearbin:
Could you push this again? The mac build got stuck and I think I have fixed it.

madmaxoft:
We have a bit too many CIs, they keep scrolling. I have no idea how to turn Appveyor's branch builder off, though.

SphinxC0re:
I think we can merge this. There hasn't been changed anything on master since this change

madmaxoft:
I'd prefer more people to actually look at the code.

bearbin:
There are only 5 now.

What we could do is merge appveyor or travis into cuberite-buildserver, as I have the spare capacity for Linux builds and hopefully you would have the capacity for windows builds.

madmaxoft:
Fixed some typos in comments and improved the comments overall. Now it's ready for review.

SphinxC0re:
Looks good to me

src/Bindings/LuaJson.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+/** Pushes the specified Json value as an appropriate type on top of the specified Lua state.
+Assumes that the Json value is not a null. */
+void PushJsonValue(const Json::Value & a_Value, cLuaState & a_LuaState)
+{
+	switch (a_Value.type())
+	{
+		case Json::nullValue:
+		{
+			ASSERT(!"Cannot push null json value onto Lua stack");
```
worktycho: This seems vulnerable to DOS attacks on any plugin that uses this for a server. I would suggest returning an error code, then translating that to a lua error in the bindings.

NiLSPACE:
I'm wondering how well this perform. Could plugins use this to send tables to other plugins?

Also, how does the serializer react to meta tables?

src/Bindings/LuaJson.cpp
madmaxoft: Good thinking.

madmaxoft:
I think the performance is not too great, it should be okay for basic web API consumption, but not for a high-level high-volume protocol.

There's no special interaction for meta-tables. The serializer can use the `tostring` meta-methods for the table keys, but it only serializes numbers, strings and tables as values.

src/Bindings/LuaJson.cpp
madmaxoft: I decided to convert `Json::nullValue` to a Lua `nil`, it doesn't seem that would break anything, other than arrays having holes in them, which is acceptable I guess.

madmaxoft: It's even better this way, because with the old parser the array indices got shifted down.

PR #2802, opened 2015-12-22
Added 1.7 item names

mathias-github:
- Added item names introduced in Minecraft 1.7 to items.ini. /give commands using the minecraft: format for items will now work properly. I've also added 1.7 item names without underscores, for convenience and consistency.
- The IDs for banners are corrected.


SphinxC0re:
looks good to me

madmaxoft:
I'm a bit concerned with some of the renames, but they're all in the name of vanilla compatibility, so if it breaks anything, it will need to be fixed, instead of sticking to non-compliant item names.

PR #2813, opened 2015-12-24
Prevent player falling through unloaded chunks

### LINKED ISSUE #2316: Implement vanilla style falling prevention when the current chunk is not loaded
Cuberite should prevent entities from falling through blocks in unloaded chunks.
This is because entities, or at least players slowly fall when the chunks they are in are not loaded.
Vanilla does this. It is possible to save the location of an entity when entering an unloaded chunk, and teleporting him back when the chunk gets loaded.

This will fix a lot of nether portal problems we currently have, but it's not just portals, it's always when an entity enters an unloaded chunk.

http://forum.mc-server.org/showthread.php?tid=2044&pid=21678#pid21678

SafwatHalaby:
INCOMPLETE
closes #2316

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -79,7 +79,7 @@ cEntity::~cEntity()
 {
 	// Before deleting, the entity needs to have been removed from the world, if ever added
 	ASSERT((m_World == nullptr) || !m_World->HasEntity(m_UniqueID));
-	
+
 	/*
 	// DEBUG:
 	LOGD("Deleting entity %d at pos {%.2f, %.2f, %.2f} ~ [%d, %d]; ptr %p",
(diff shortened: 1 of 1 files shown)


src/Entities/Entity.cpp
```diff
@@ -849,13 +849,19 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	int BlockZ = POSZ_TOINT;
 
 	// Position changed -> super::HandlePhysics() called
-	GET_AND_VERIFY_CURRENT_CHUNK(NextChunk, BlockX, BlockZ)
+	cChunk * NextChunk = a_Chunk.GetNeighborChunk(BlockX, BlockZ);
+
+	if ((NextChunk == nullptr) || !ChunkVarName->IsValid())
+	{
+		// FREEZE THE CLIENT and DONT TRUST ITS position updates until chunk is loaded
```
SafwatHalaby: Am I in the right direction?

Issue #2810, opened 2015-12-23
"Phantom worlds"


LogicParrot:
- Currently, when going to the nether/end for the first time, it is generated at runtime. This is heavy, and can freeze the tick thread and trigger the deadlock detector on weak devices like Rpis.
 - The newly generated worlds are "phantom worlds" - never mentioned in `settings.ini`.
 - Nether and End should be in `settings.ini` by default so that they aren't "phantom worlds" and are generated at first startup to avoid the deadlock detector issue.
 - Worlds that are referenced in `world.ini` but aren't in `settings.ini` (The so-called phantom worlds) should cause an error when server reads the config, and not generate a new world.
 - The above point means deprecating `cRoot::Get()->CreateAndInitializeWorld(...)`  (or at least using it only at server startup when a world is missing) and redoing #2462.

**Todo:**
 - [x] 1. Nether and End in `settings.ini` by default, and created at startup. #2815
 - [x] 2. Fix server configs created prior to the #2815 update. Done in #2821
 - [x] 3. Prevent generating phantom worlds in default configs. Done in #2953
 - [x] 4. Reject the existence of phantom worlds by config check at startup. Done in #2961
 - [x] 5. Remove `CreateAndInitializeWorld`. Done in #2962, unmerged.

**Original post:**
When traveling to the Nether for the first time, the folder `world_nether` is created. But the newly created world is never added in `settings.ini`. Is this a bug or intended?

NiLSPACE:
The world should never be created in the first place. It's unsafe to add a world while the server is running.

SafwatHalaby:
Would it be hard to generate the Nether an End on startup just like the overworld? Then we can have them in `settings.ini` by default.

NiLSPACE:
I don't think Cuberite should generate a world if it's not defined by the admin himself in the `settings.ini`

If the admin defined a world in a `world.ini` while that world doesn't exist we should simply show an error message if you'd ask me.

SafwatHalaby:
My test suggests that adding `World=world_nether` to `settings.ini` prior to first server startup is enough to resolve this. It generates both worlds properly. That should be in the ini file by default.

SafwatHalaby:
This also prevents a world being generated at runtime, which you mentioned is unsafe @ #2462

SafwatHalaby:
>I don't think Cuberite should generate a world if it's not defined by the admin himself in the settings.ini
[...]

PR #2815, opened 2015-12-24
Nether and End in settings.ini by default, and created at startup

SafwatHalaby:
Fixes part 1 of #2810

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -143,13 +143,16 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	LOG("Reading server config...");
 
 	auto IniFile = cpp14::make_unique<cIniFile>();
-	if (!IniFile->ReadFile("settings.ini"))
+	bool IsNewIniFile = !IniFile->ReadFile("settings.ini");
+
+	if (IsNewIniFile)
 	{
 		LOGWARN("Regenerating settings.ini, all settings will be reset");
 		IniFile->AddHeaderComment(" This is the main server configuration");
 		IniFile->AddHeaderComment(" Most of the settings here can be configured using the webadmin interface, if enabled in webadmin.ini");
 		IniFile->AddHeaderComment(" See: http://wiki.mc-server.org/doku.php?id=configure:settings.ini for further configuration help");
 	}
+
 	auto settingsRepo = cpp14::make_unique<cOverridesSettingsRepository>(std::move(IniFile), std::move(a_OverridesRepo));
 
 	LOG("Starting server...");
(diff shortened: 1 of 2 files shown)


SafwatHalaby:
Apparently this is not enough.

SafwatHalaby:
Fixed and rebased.

src/Root.cpp
```diff
@@ -143,13 +143,20 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	LOG("Reading server config...");
 
 	auto IniFile = cpp14::make_unique<cIniFile>();
+	bool NewIniFile;
+
 	if (!IniFile->ReadFile("settings.ini"))
 	{
+		NewIniFile = true;
```
worktycho: I feel this might be better written as:
```
bool NewIniFile = !IniFile->ReadFile("settings.ini");
if (NewIniFile)

SafwatHalaby:
I think that before deprecating CreateAndInitializeWorld, we should make sure all those who have generated their world prior to this PR get some treatment. Otherwise many will end up without a Nether and an End without any obvious reason to them.

src/Root.cpp
SafwatHalaby: Fixed.

madmaxoft:
The code looks okay to me, except that I'd prefer naming bool values `IsXYZ`, `HasXYZ` etc. instead of simple `XYZ` - to make it clear that it's a flag, not the contents.

I assume you've tested this on a clean config?

There's still the matter of handling the linkage to non-existent worlds, but we can address that in a separate PR.

SafwatHalaby:
I renamed that variable and rebased.

Yes, I have tested that in various configs.
Before throwing errors on linkages to non-existent worlds, we must fix the servers who generated the world before this PR using this procedure:

If:
 - These folders exist: `world`, `world_nether`, `world_end` &&
 - `settings.ini` contains `world` but doesn't contain `world_nether` or `world_end`

Then:
 - Add `world_nether` and `world_end` to the ini.

SafwatHalaby:
Not happy with this at all. I think multiworld teleportation is too unstable and shouldn't be the default.

SafwatHalaby:
There shouldn't be a default Nether and End.

SafwatHalaby:
Perhaps the Nether and End shouldn't be traversable at all by default. Discussion here:
http://forum.mc-server.org/showthread.php?tid=2272

tigerw:
Hold on, my original intent was to have the format be \*current world name\*_end/_nether, rather than a preprogrammed world(_end/_nether) prefix.

SafwatHalaby:
But the code I added is triggered when NO INI FILE is present, as the default.

Issue #2817, opened 2015-12-24
Cannot load Prefabs/PieceStructures/NetherForts.cubeset.gz


SafwatHalaby:
The console warns me of this:

```
[3653c5da85050f72|18:02:47] Cannot load PieceStructures cubeset file Prefabs/PieceStructures/NetherForts.cubeset.gz
[3653c5da85050f72|18:02:47] The PieceStructures generator was asked to generate "NetherForts", but none of the prefabs are valid.
```

NiLSPACE:
In the world.ini change `PieceGenerator: NetherForts ` to `PieceGenerator: NetherFort`

madmaxoft:
There was a bug in the initial PieceStructures rewrite that suggested the wrong name (`NetherForts`), I have fixed it since to the correct `NetherFort`.

I suppose I should still change the error message to account for both possible filenames- with and without the `.gz` extension.

PR #2820, opened 2015-12-25
Added the cUrlParser class, exported to Lua API.

madmaxoft:
This will simplify the web API interaction with `cNetwork`.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -2286,7 +2286,50 @@ local CompressedString = cStringCompression.CompressStringGZIP("DataToCompress")
 				SetFuseTicks = { Return = "number", Notes = "Set the fuse ticks until the tnt will explode." },
 			},
 			Inherits = "cEntity",
-		},
+		},  -- cTNTEntity
+
+		cUrlParser =
+		{
+			Desc = [[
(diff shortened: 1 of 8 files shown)


worktycho:
What about protocol relative URLs? For example, the nasty test case: `//::1/foo`

madmaxoft:
Don't wanna do that. We don't need it to be that much universal.
And that URL of yours is invalid, IPv6 should be in brackets: `//[::1]/foo`

madmaxoft:
Uhhh.. I need assistance with the Linux build, it seems the Bindings library would need dependency on the HTTPServer library.

SphinxC0re:
The OS X build is failing:
```sh
Warning: [19:07:34] Cannot load PieceStructures cubeset file Prefabs/PieceStructures/NetherForts.cubeset.gz
Warning: [19:07:34] The PieceStructures generator was asked to generate "NetherForts", but none of the prefabs are valid.
```

SafwatHalaby:
@SphinxC0re That would be issue #2817.

SafwatHalaby:
@SphinxC0re the real failure seems to be a `Segmentation fault`. This is happening randomly on any PR.

madmaxoft:
I don't think the segfault is caused by this PR. Are there any more objections, or can we merge? I've already got two branches dependent on this PR :)

PR #2821, opened 2015-12-25
Fix pre - #2815 default world linkage configs

LogicParrot:
Fixes part 2 of #2810

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -347,7 +347,6 @@ void cRoot::LoadGlobalSettings()
 void cRoot::LoadWorlds(cSettingsRepositoryInterface & a_Settings, bool a_IsNewIniFile)
 {
 	// First get the default world
-	AString DefaultWorldName = a_Settings.GetValueSet("Worlds", "DefaultWorld", "world");
 	if (a_IsNewIniFile)
 	{
 		a_Settings.AddValue("Worlds", "World", "world_nether");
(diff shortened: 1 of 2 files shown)


madmaxoft:
I don't think this is necessary. If the server aborts due to bad linkage, the error message should include instructions on how to fix that and the admin should be left to fix it themselves. I don't like having these "compatibility fixed" around too much.

SafwatHalaby:
Your call.
But once the linkage rules are enforced, almost everyone will get an error without this.
With this merged, only a minority will get an error. The latter situation sounds much better to me.

SafwatHalaby:
Consider the fact that some people may have never bothered editing a `world.ini` beyond changing the gamemode and just want to play.

SafwatHalaby:
There's also a deeper issue which I've been wondering about:
After linkage rules are enforced, how exactly would I create a new Nether? If I add a new world in `settings.ini`, a regular world would be created upon startup.

madmaxoft:
Without this patch, everyone will get an error.
With this patch, everyone will suddenly have two extra worlds loaded that they never configured.

Not sure which is better. @cuberite/core-team ?

Your last issue is an old one - there was never a way to specify the world just in settings.ini and have it created as Nether or End. You always had to add the world to settings.ini, create the folder and supply a world.ini containing at least the nether dimension.

SphinxC0re:
I think having a default configuration with prelinked dimensions is good for starting a server

LogicParrot:
@madmaxoft: That's a bit inaccurate.
Without this patch, everyone will get an error.
With this patch, those who **never changed the defaults** will suddenly have two extra worlds in settings.ini. Those people may have already visited those two extra worlds and created Phantom Worlds (existing worlds which aren't in settings.ini).

LogicParrot:
@madmaxoft: That's a bit inaccurate.

Without this patch, those who **never changed the default linkages** or **have bad linkages** will get an error.

With this patch, those who **never changed the default linkages** will suddenly have two extra worlds in settings.ini. Those people may have already visited those two extra worlds and created Phantom Worlds (existing worlds which aren't in settings.ini). Those who have **have bad linkages** will get an error.

Edit: fixed.

LogicParrot:
Ping. Close or merge?

worktycho:
I vote merge. You've clearly marked it as a backcompat hack, and its nessicary to migrate most worlds.

madmaxoft:
Alright then, rebase and merge.

LogicParrot:
I sincerely apologize, this fix actually made it worse. People who ran a fresh server after this fix was merged got a `world_nether` and a `world_end`, which are both **overworlds**, which both have bad linkages. `world_nether`'s linkages are `world_nether_nether` and `world_nether_end`. and `world_end`'s  linkages are `world_end_nether` and `world_end_end`.

There is no need to revert, because with the new commits in place, this now works as intended.

LogicParrot:
In summary, this was supposed to autofix everyone, but due to the issue above it will not autofix people that downloaded/updated a server between `Jan 12, 2016` and `Feb 8, 2016` and then generated worlds. That's not too bad, considering the fact that we were thinking of not fixing it automatically for anyone at all.

LogicParrot:
In summary, this was supposed to autofix everyone, but due to the issue above it will not autofix people that downloaded/updated a server between `Jan 12, 2016` and `Feb 8, 2016` and then generated worlds while already having a settings.ini (completely fresh installs got things configured properly). That's not too bad, considering the fact that we were thinking of not fixing it automatically for anyone at all.

Edit: update info.

LogicParrot:
https://forum.cuberite.org/thread-2366-post-24992.html

PR #2825, opened 2015-12-26
fix crafting recipe for snow layers

Gargaj:
clipboard01


madmaxoft:
You do know that you have write-access to the main repo, do you? You can develop on a branch in here, which is much easier to merge then, because anyone can rebase the branch at any time.

Gargaj:
It is true that I've been offered write access, but I didn't accept it. I don't want to get more involved than necessary.

SphinxC0re:
What a brave man

Issue #2830, opened 2015-12-27
cMojangAPI segmentation faults and asserts


tigerw:
We are currently experiencing technical difficulties with regards to connections to Mojang; references are as follows:
* #2826
* #2827
* #2818
* http://forum.mc-server.org/showthread.php?tid=2154

Potential analysis:
----
### Problem 1
Mojang made their servers close the connection quicker. This means that `cBlockingSslClientSocket::OnDisconnected` is often called whilst cMojangAPI is still reading data from the SSL context (whatever that is :P). In Debug builds, this causes further calls to `cBlockingSslClientSocket::Receive` to assert that `m_IsConnected` is unexpectedly false. [This is known](https://girlsgonegeek.files.wordpress.com/2013/11/it-is-known.jpg).

### Problem 2
`OnDisconnected` runs on *a separate thread* from the thread cMojangAPI calls `Receive` from. `OnDisconnect` also will reset the `m_Server` *pointer* of the above class. Thus, in Release builds, I suspect, but am not sure, that a concurrency issue occurs where after `m_Server` has been reset but before `m_IsConnected` has been set to false by `OnDisconnect`, `cBlockingSslClientSocket::Disconnect` is called automatically by the destruction of the `cBlockingSslClientSocket` object (because cMojangAPI has completed its work super quickly, yeah) which attempts to dereference the `m_Server` to call `cTCPLink::Close`, but crashes since `m_Server` was already reset.

SphinxC0re:
@Dinnerbone could you give us a hint on this? :smile:

madmaxoft:
That wasn't nice, there's a problem in our code, no need to bother Mojang people about it.

We clearly have a race condition in the networking code.

Seadragon91:
Looked with clang thread sanitizer. There is a data race on variable `m_IsConnected` [BlockingSslClientSocket.cpp:341](https://github.com/cuberite/cuberite/blob/master/src/PolarSSL%2B%2B/BlockingSslClientSocket.cpp#L341) (write) and [BlockingSslClientSocket.cpp:255](https://github.com/cuberite/cuberite/blob/master/src/PolarSSL%2B%2B/BlockingSslClientSocket.cpp#L255) (read). Not sure if that is the problem.

tigerw:
Xoft, please try not to add too many queues and locks, I don't like queues and locks... :P

SphinxC0re:
I made a joke... just a joke

worktycho:
[...]

PR #2831, opened 2015-12-28
Changed m_IsConnected to an atomic bool, added logging.

madmaxoft:
This is the hunt for the cause of #2830, requiring more testing.

### DIFF
--- a/src/PolarSSL++/BlockingSslClientSocket.cpp
+++ b/src/PolarSSL++/BlockingSslClientSocket.cpp
@@ -70,7 +70,9 @@ class cBlockingSslClientSocketLinkCallbacks:
 	{
 		m_Socket.OnDisconnected();
 	}
+
 public:
+
 	cBlockingSslClientSocketLinkCallbacks(cBlockingSslClientSocket & a_Socket):
 		m_Socket(a_Socket)
 	{
@@ -143,7 +145,6 @@ bool cBlockingSslClientSocket::Connect(const AString & a_ServerName, UInt16 a_Po
 		return false;
 	}
 	
-	m_IsConnected = true;
 	return true;
 }
 
@@ -182,7 +183,11 @@ bool cBlockingSslClientSocket::SetTrustedRootCertsFromString(const AString & a_C
 
 bool cBlockingSslClientSocket::Send(const void * a_Data, size_t a_NumBytes)
 {
-	ASSERT(m_IsConnected);
+	if (!m_IsConnected)
+	{
+		m_LastErrorText = "Socket is closed";
+		return false;
+	}
 	
 	// Keep sending the data until all of it is sent:
 	const char * Data = reinterpret_cast<const char *>(a_Data);
@@ -216,8 +221,7 @@ bool cBlockingSslClientSocket::Send(const void * a_Data, size_t a_NumBytes)
 
 int cBlockingSslClientSocket::Receive(void * a_Data, size_t a_MaxBytes)
 {
-	ASSERT(m_IsConnected);
-	
+	// Even if m_IsConnected is false (socket disconnected), the SSL context may have more data in the queue
 	int res = m_Ssl.ReadPlain(a_Data, a_MaxBytes);
 	if (res < 0)
 	{
@@ -239,9 +243,16 @@ void cBlockingSslClientSocket::Disconnect(void)
 	}
 	
 	m_Ssl.NotifyClose();
-	m_Socket->Close();
-	m_Socket.reset();
 	m_IsConnected = false;
+
+	// Grab a copy of the socket so that we know it doesn't change under our hands:
+	auto socket = m_Socket;
+	if (socket != nullptr)
+	{
+		socket->Close();
+	}
+
+	m_Socket.reset();
 }
 
 
@@ -293,6 +304,7 @@ int cBlockingSslClientSocket::SendEncrypted(const unsigned char * a_Buffer, size
 
 
 
+
 void cBlockingSslClientSocket::OnConnected(void)
 {
 	m_IsConnected = true;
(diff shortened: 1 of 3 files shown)


src/PolarSSL++/BlockingSslClientSocket.cpp
```diff
@@ -305,7 +306,7 @@ void cBlockingSslClientSocket::OnConnected(void)
 
 void cBlockingSslClientSocket::OnConnectError(const AString & a_ErrorMsg)
 {
-	LOG("Cannot connect to %s: %s", m_ServerName.c_str(), a_ErrorMsg.c_str());
+	LOG("Cannot connect to %s: \"%s\"", m_ServerName.c_str(), a_ErrorMsg.c_str());
```
tigerw: naaah, quotes not needed :)

madmaxoft: They are, indeed, needed. If the error message is empty for any reason, it will not produce a misleading output.

madmaxoft:
The current state of this branch (781bf56) seems to have helped the crashing, at least on my Linux. However, now no-one can connect because the response is not parsed properly. Investigating more.

madmaxoft:
Commit e96d599 is fully working for me, both on Windows and Linux. @Seadragon91 , @Gargaj , @lstep , @psilohanf could you verify?

If this indeed fixes the issue, I'll rebase and squash the PR, please don't merge it as-is.

psilohanf:
Hi,

  I cloned the resp. branch but I cannot compile it. I am missing some dependencies (starting with lib/SQLiteCpp). The instructions for manual compilation do not cover this and the compile.sh is always referring to the „master“ branch :-(. I am not familiar enough with the project to solve this.

  Best regards,

	Christoph

>  Am 30.12.2015 um 00:54 schrieb Mattes D <notifications@github.com>:
> 
> Commit e96d599 <https://github.com/cuberite/cuberite/commit/e96d599ac0c53bf3bea3b9b5643526f6f612f18d> is fully working for me, both on Windows and Linux. @Seadragon91 <https://github.com/Seadragon91> , @Gargaj <https://github.com/Gargaj> , @lstep <https://github.com/lstep> , @psilohanf <https://github.com/psilohanf> could you verify?
> 
> If this indeed fixes the issue, I'll rebase and squash the PR, please don't merge it as-is.
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/cuberite/cuberite/pull/2831#issuecomment-167901888>.
> 

_________________
Christoph Krüger
Christoph@Krueger.name
www.christoph-krueger.de

SphinxC0re:
You have to ```git submodule update```

Seadragon91:
Tested release mode under windows, linux and it's works fine. In debug mode under linux, I had a access violation. Was unable to reproduce it, will look if I can reproduce it.

This is ready for merge, from my side.

tigerw:
Ah, the glories and wonders of multithreading and networking!

psilohanf:
Hi,

  did that … and some other stuff :-) (couldn’t accept defeat).
  Et voilá ... Cuberite is running again. No problems on my server. I could login without problems. Thank you very much. Well done!

  Best regards,

	Christoph

> Am 30.12.2015 um 01:25 schrieb Julian Laubstein <notifications@github.com>:
> 
> You have to git submodule update
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/cuberite/cuberite/pull/2831#issuecomment-167904294>.
> 

_________________
Christoph Krüger
Christoph@Krueger.name
www.christoph-krueger.de

SphinxC0re:
The OS X build is okay btw. The connection just broke up

src/PolarSSL++/BlockingSslClientSocket.cpp
```diff
@@ -239,9 +243,16 @@ void cBlockingSslClientSocket::Disconnect(void)
...
 	m_IsConnected = false;
+
+	// Grab a copy of the socket so that we know it doesn't change under our hands:
+	auto socket = m_Socket;
+	if (socket != nullptr)
+	{
+		socket->Close();
+	}
+
+	m_Socket.reset();
```
tigerw: With this arrangement, is the `!m_IsConnected` check needed any more, considering that the lack of locking means that the variable and the state it's associated with could change however it wants after that line?

Also, would it be valid to move the `::reset` call into the `if`'s codeblock for performances and stuff?

Also also, `socket` -> `Socket`? :P

src/PolarSSL++/BlockingSslClientSocket.cpp
```diff
@@ -182,7 +183,11 @@ bool cBlockingSslClientSocket::SetTrustedRootCertsFromString(const AString & a_C
 
 bool cBlockingSslClientSocket::Send(const void * a_Data, size_t a_NumBytes)
 {
-	ASSERT(m_IsConnected);
+	if (!m_IsConnected)
+	{
+		m_LastErrorText = "Socket is closed";
+		return false;
+	}
```
tigerw: Does it matter if the socket suddenly closes anywhere in this loop, after the check?

Gargaj:
Can confirm that this branch works 100% of the time vs current ```master``` only working 1 out of 10 times. GG!

madmaxoft:
@tigerw The `m_IsConnected` flag is there mostly to protect the clients of this code against bad usage - we don't want them to send data before opening the socket. Other than that, it is not used for the socket's logic anymore.

src/PolarSSL++/BlockingSslClientSocket.cpp
madmaxoft: Socket closure will be detected from within the `m_Ssl.WritePlain()`, it will return an error code, thus taking the branch on line 202. The error message will be different, but it will still be a failure.
This here is only so that the error message is obvious in the obvious cases.

madmaxoft:
I'm gonna merge this, sans the Debuggers change; I'll be refactoring most of this code away soon anyway.

PR #2832, opened 2015-12-28
Fence update

LogicParrot:
Prior to this PR, mobs would get confused when a player is next to a fence. This is no longer the case.

This is definitely not permanent, I'd like to fix it on a more fundemental level later.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -34,6 +34,7 @@ cPath::cPath(
 	const Vector3d & a_StartingPoint, const Vector3d & a_EndingPoint, int a_MaxSteps,
 	double a_BoundingBoxWidth, double a_BoundingBoxHeight
 ) :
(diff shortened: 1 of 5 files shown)


src/Mobs/Path.cpp
```diff
@@ -56,6 +57,11 @@ cPath::cPath(
 	m_Destination.y = FloorC(a_EndingPoint.y);
 	m_Destination.z = FloorC(a_EndingPoint.z - HalfWidthInt);
 
+	if (IsBlockFence(GetCell(m_Destination)->m_BlockType))
+	{
+		m_IsWideDestination = true;  // Any location in a 3x3 area centered on m_Destination is a good enough destination.
```
LogicParrot: Comment is inaccurate.

LogicParrot:
Not working as intended

PR #2833, opened 2015-12-28
Fence update

LogicParrot:
Take 2. This should work.

Prior to this PR, mobs would get confused when a player is next to a fence. This is no longer the case.
This is definitely not permanent, I'd like to fix it on a more fundemental level later.

### DIFF
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -34,6 +34,7 @@ cPath::cPath(
 	const Vector3d & a_StartingPoint, const Vector3d & a_EndingPoint, int a_MaxSteps,
 	double a_BoundingBoxWidth, double a_BoundingBoxHeight
 ) :
+	m_IsWideDestination(false),
 	m_StepsLeft(a_MaxSteps),
 	m_IsValid(true),
 	m_CurrentPoint(0),  // GetNextPoint increments this to 1, but that's fine, since the first cell is always a_StartingPoint
@@ -56,6 +57,11 @@ cPath::cPath(
 	m_Destination.y = FloorC(a_EndingPoint.y);
 	m_Destination.z = FloorC(a_EndingPoint.z - HalfWidthInt);
 
+	if (IsBlockFence(GetCell(m_Destination)->m_BlockType))
+	{
+		m_IsWideDestination = true;
+	}
+
 	if (!IsWalkable(m_Source, m_Source))
 	{
 		m_Status = ePathFinderStatus::PATH_NOT_FOUND;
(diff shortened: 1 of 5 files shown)


src/Mobs/Path.cpp
```diff
@@ -139,11 +149,25 @@ bool cPath::StepOnce()
...
 		FinishCalculation(ePathFinderStatus::PATH_FOUND);
 		return true;
 	}
 
+	if (m_IsWideDestination)
+	{
+		Vector3i Delta = CurrentCell->m_Location - m_Destination;
+		int DeltaX = abs(Delta.x);
+		int DeltaZ = abs(Delta.z);
+		if ((Delta.y == 0) && (((DeltaX == 0) && (DeltaZ == 1)) || ((DeltaX == 1) && (DeltaZ == 0))))
```
worktycho: This is a nasty condition. It might be worth breaking out the `(DeltaX == 0) && (DeltaZ == 1)` clauses on to separate lines, and use indentation to make it clear the `(Delta.y == 0)` clause applies to both.

LogicParrot: Agreed.

In general, it's a nasty PR which solves the fact that mobs go totally insane near fences. And it should be solved differently later.

madmaxoft:
I'm missing some comments that would guide me through what's going on - what's the problem and what's the solution.

LogicParrot:
The problem: When a player is touching a fence, the pathfinder thinks the player is standing inside a solid, because the player is occupying the same block the fence is occupying. This makes cPath return `NEARBY_FOUND`, which makes the mob think the player is not reachable and enter "hang around mode", just like it does when the player is really not reachable. It may randomly get close to the player while "hanging around", which triggers attacking.

The solution: This is a quick solution because I had no time to write a proper one. The idea is this: If `EnsureProperDestination` detects a fence, then set a flag in `cPath`. Normally. cPath's destination is the original block. With the flag set, the destination is inflated into a "cross", whose arms are each 1 block long.  So, if a player is touching a fence at `5, 5, 5`, then `4,5,5` is considered a valid destination, and so is `5,5,6`, etc.

worktycho:
Would it be a good idea to try and ensure the approaches from the same side as the player is?

LogicParrot:
@worktycho: Yes. This is one of the reasons I consider this incomplete. It's a "hotfix" for an ugly bug, which should be better treated later.

bibo38:
Maybe adding an Issue/TODO to not forget about this "hotfix"?

LogicParrot:
@bibo38 Fence handling is marked incomplete @ the [forum post](http://forum.mc-server.org/showthread.php?tid=1571). That should be enough to remind me.

PR #2836, opened 2015-12-30
Fixed default Nether generator params.

madmaxoft:
The prefab file is "NetherFort", not "NetherForts"


SphinxC0re:
again, the connection of the OS X node broke off

SphinxC0re:
I know, this is a small change but could you rebase one more time?

PR #2837, opened 2015-12-30
Fixed the monster burn code.

madmaxoft:
The old code failed when the monster was out of Y range.
The new code causes undead mobs to always burn above the world and never burn under the world.


SphinxC0re:
Looks good to me

PR #2840, opened 2016-01-01
Renamed leftover strings to Cuberite / Server, as needed.

madmaxoft:
Also upgraded the user setting file for MSVC to 2013.


SphinxC0re:
You missed a few newlines on some files

madmaxoft:
The newlines weren't there in the first place, so I didn't add them. The `Cuberite.vcxproj.user` file is even created by MSVC itself, so I don't think it'd be a good idea to edit it manually.

Server/Plugins/ChatLog
```diff
@@ -1 +1 @@
-Subproject commit 983d23ca37baa89f7e4dc11d71502d9c059f6376
+Subproject commit cc04da381b9aa683ee3bbd7013f9d3fd909db039
```
SphinxC0re: Is this change intentional?

madmaxoft: Yes, all submodule changes are intentional, there were substitutions in the plugins as well.

PR #2844, opened 2016-01-03
Added blocks that can't be mined with silktouch

### LINKED ISSUE #2843: Silk touch pickaxe creates drop that crashes client
When a player mines a block that also has an item variant (e.g. a redstone repeater), with a silk touch pickaxe, an invisible drop is created. Whenever the player picks up the drop, its client crashes.

In Spigot (most likely in vanilla as well), the item variant of the block is dropped when the block is mined with a silk touch pickaxe. If it's a technical block that doesn't have an item variant, no drop is created.

mathias-github:
Fixes #2843


madmaxoft:
Thanks.

PR #2849, opened 2016-01-08
Exported GetSeed to lua and added file extension opendb to .gitignore.

### LINKED ISSUE #2767: Export the seed of the world to lua
This would be a good idea. In vanilla minecraft you can get the seed with `/seed`. The command itself, could be implemented in Core.

Seadragon91:
Added `GetSeed` to `cWorld` from `cChunkGenerator`. Also added a file to .gitnore, that is created on startup of Visual Studio 2015, that makes `git add .` not usable. Current workaround `git add -u`.

Closes #2767.

### DIFF
--- a/src/World.h
+++ b/src/World.h
@@ -757,6 +757,9 @@ public:
 		return (IsWeatherWet() && !IsBiomeNoDownfall(GetBiomeAt(a_BlockX, a_BlockZ)));
 	}
 
+	/** Returns the seed of the world. */
+	int GetSeed(void) { return m_Generator.GetSeed(); }
+
 	// tolua_end
 
 	cChunkGenerator & GetGenerator(void) { return m_Generator; }
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,7 @@ src/AllFiles.lst
 *.opensdf
 *.sdf
 *.sln
+*.opendb
 *.idb
 
 # cmake output folders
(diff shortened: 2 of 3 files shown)


src/World.h
```diff
@@ -757,6 +757,9 @@ class cWorld :
 		return (IsWeatherWet() && !IsBiomeNoDownfall(GetBiomeAt(a_BlockX, a_BlockZ)));
 	}
 
+	/** Returns the seed of the world. */
+	int GetSeed(void) { return m_Generator.GetSeed(); }
```
worktycho: I don't like the fact that this is in cWorld. The Seed is not cWorld's responsibility.

worktycho: I'd rather you exported the generator with only this method.

madmaxoft:
I don't think exporting a class just so that we get one function out to it is worth it. I like the current solution and vote for merging this as-is.

The seed itself needn't be used only for the generator, for example the client uses it to generate random rotation when displaying lilypads.

NiLSPACE:
I believe we also need it for when it's snowing. Currently the snow line is fixed per-biome. In vanilla it can range between a few values.

worktycho:
But it doesn't belong in cWorld. cWorld has a thousand line header, because we refuse to make the decision to stop adding to cWorld. If you don't want to export the generator, use a free function.

PR #2851, opened 2016-01-10
Fixed not working packet decompression

### LINKED ISSUE #2467: Player kicked for using modded client
Just thought I'd put this out there...

Minecraft 1.8 with the Voxelmap plugin.

[12:02:34] Kicking player ShadowBudgie for "Bad compression"

### LINKED ISSUE #2841: Disconnected for "Bad compression"
Whenever you add much data to an item, you get disconnected for "Bad compression". I tried to add  more than 14 different enchantments to an item using a command, but got disconnected with this error. The same error also appeared when pasting a large command in a command block.

bibo38:
A compressed packed contains always the uncompressed size, but in the code this size was threatened as the compressed data size. This created a failure, when the client compressed a packet.
See: http://wiki.vg/Protocol#With_compression
Fixes #2841
Fixes #2467

### DIFF
--- a/src/Protocol/Protocol18x.cpp
+++ b/src/Protocol/Protocol18x.cpp
@@ -1834,39 +1834,44 @@ void cProtocol180::AddReceivedData(const char * a_Data, size_t a_Size)
 		}
 		
 		// Check packet for compression:
-		UInt32 CompressedSize = 0;
+		UInt32 UncompressedSize = 0;
 		AString UncompressedData;
 		if (m_State == 3)
 		{
 			UInt32 NumBytesRead = static_cast<UInt32>(m_ReceivedData.GetReadableSpace());
-			m_ReceivedData.ReadVarInt(CompressedSize);
-			if (CompressedSize > PacketLen)
+
+			if (!m_ReceivedData.ReadVarInt(UncompressedSize))
 			{
-				m_Client->Kick("Bad compression");
+				m_Client->Kick("Compression packet incomplete");
 				return;
 			}
-			if (CompressedSize > 0)
+
+			NumBytesRead -= static_cast<UInt32>(m_ReceivedData.GetReadableSpace());  // How many bytes has the UncompressedSize taken up?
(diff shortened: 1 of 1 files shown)


src/Protocol/Protocol18x.cpp
```diff
@@ -1834,39 +1834,40 @@ void cProtocol180::AddReceivedData(const char * a_Data, size_t a_Size)
...
 		{
 			UInt32 NumBytesRead = static_cast<UInt32>(m_ReceivedData.GetReadableSpace());
-			m_ReceivedData.ReadVarInt(CompressedSize);
-			if (CompressedSize > PacketLen)
-			{
-				m_Client->Kick("Bad compression");
-				return;
-			}
-			if (CompressedSize > 0)
+			m_ReceivedData.ReadVarInt(UncompressedSize);
```
madmaxoft: Shouldn't we check for failure here?

bibo38: How do you want to know, if the decompressed size really matches?
Unless you don't try to decompress it, you cannot know, how big it is...

madmaxoft: No, I meant if there are enough bytes in `m_ReceivedData` for the `UncompressedSize` value.
```cpp
if (!m_ReceivedData.ReadVarInt(...))
```

bibo38: Isn't this checked [here](https://github.com/bibo38/Cuberite/blob/decompressfix/src/Protocol/Protocol18x.cpp#L1829), since the package length is tested to be ready to read, which includes the uncompressed size?

madmaxoft: I think we shouldn't assume, because malicious clients could send a malformed packet with `PacketLength` set to, say, 1, but `UncompressedSize` set to 800000; this code would then probably fail.

bibo38: Maybe... I've just fixed it.

SphinxC0re:
LGTM

Issue #2834, opened 2015-12-28
Spectator mode teleportation doesn't work


mathiascode:
Whenever you try to teleport to another player using the teleportation feature in spectator mode (while in spectator mode, press any number to bring up the inventory, then select which player to teleport to), you get disconnected with the following message:

"Unknown [C->S] PacketType: 0x18"

Seadragon91:
See this issue #1772. Spectator mode is not really existent.

tigerw:
A new packet named “Spectate”...

Issue #2531, opened 2015-10-07
Spectator mode: 1st person view of other players not working


jammet:
Aside from flying around and observing the world, a spectator should be able to see through the eyes of another player, or any mob. This does not seem to work with Cuberite, now.

From the Wiki:
A spectator can left click on an entity, to take its view. Their position is locked to that of the entity, and their POV moves in accordance with the entity. Viewing oneself while spectating from an entity's point-of-view using F5 shows the entity. Pressing the dismount key (⇧ Shift by default) returns the player to flying mode.

Issue #1772, opened 2015-02-26
spectator mode bugs / missing features


Zauberfisch:
I have just finished installing and imported my existing world. 
It seems that npcs are aware of me even though I am in spectator mode. They follow me and creepers even explode.

details that might be relevant:
- Gamemode=3 (spectator mode, in world.ini)
- world folder is copied from vanilla server
- running windows x64
-  MCServer  build id: 182 (from commit a56b6906d60f5b511aeb93cf765359634e4bcf6e)

---

EDIT: I found a lot of other issues, so I collected them into a list:
- [ ] NPCs follow a spectator
- [ ] creepers will explode near a spectator
- [ ] spectator can attack/kill animals
- [ ] spectator can pick up items
- [ ] spectator can not spectate npcs (when you left click an npc as spectator, you should see the world through the eyes of the npc)
- [ ] spectator can interact with items as well (chests, doors)
- [ ] spectator can use inventory
- [ ] spectator can craft
- [ ] after re-aranging items in the inventory, a spectator is able to build (unlimited) blocks that he have in his inventory on slot 1. (eg 3 stone blocks in slot 1 => build unlimited stone blocks (seems to work with all items, but the item has to be moved to slot 1, other slots do not work))

madmaxoft:
Nice catch, thank you.

Zauberfisch:
do you guys prefer individual tickets or collections of related issues?

Because on the topic of spectator mode I noticed the following other issues:
- I can use my inventory
- I can attack/kill animals
- I can pick up items
- I can not spectate npcs (when you left click an npc as spectator, you should see the world through the eyes of the npc)

madmaxoft:
In this case we'll use a single issue, because the entire spectator mode is kinda not implemented. It's there so you can choose it, but there's not much handling for it.

Zauberfisch:
yeah, I am getting that feeling. will rename the issue and create a full list in the original message.

Zauberfisch:
done.

Also, let me point out that I actually think spectator is an important feature.
We use spectator as a sort of soft white-listing (instead of whitelist or build protection) to allow others to connect to our server and take a look, but stop them from doing anything unless they receive build permissions from the admins.

PR #2852, opened 2016-01-10
Spectators added

bibo38:
I've added the spectator gamemode completely to Cuberite.
This includes #1772, #2834, #2531.

Btw.: This also fixes some mobs targeting creative players(like endermen)


src/ClientHandle.cpp
```diff
@@ -1532,6 +1539,27 @@ void cClientHandle::HandleSlotSelected(Int16 a_SlotNum)
 
 
 
+void cClientHandle::HandleSpectate(const AString & a_PlayerUUID)
```
worktycho: Do we have to add another method to cClientHandle to handle this? I'm not sure this is the best long term decision.

src/ClientHandle.cpp
```diff
@@ -1591,7 +1619,26 @@ void cClientHandle::HandleUpdateSign(
...
+			bool Item(cEntity * a_Entity) override
+			{
+				m_Player.AttachTo(a_Entity);
+				return true;
+			}
+		public:
+			cSpectateEntity(cPlayer & a_Player) : m_Player(a_Player) {}
+		} Callback(*m_Player);
+
+		m_Player->GetWorld()->DoWithEntityByID(a_TargetEntityID, Callback);
```
worktycho: Does DoWithEntityByID not support lambdas?

src/ClientHandle.cpp
```diff
@@ -2082,6 +2141,15 @@ void cClientHandle::SendBlockChanges(int a_ChunkX, int a_ChunkZ, const sSetBlock
 
 
 
+void cClientHandle::SendCameraSetTo(const cEntity & a_Entity)
```
worktycho: Again, is there a good way to avoid new cClientHandle methods?

src/ClientHandle.cpp
bibo38: What's the better approach? Handeling the things in every Packet*.cpp?

src/ClientHandle.cpp
bibo38: I've always seen this kind of Callback-thing. Should I only change my callbacks or any other callbacks?

src/ClientHandle.cpp
bibo38: This are new packets... Unless you mix them with something else, but I have no idea about that...

src/ClientHandle.cpp
worktycho: No. But we need to think of a better way. The one I tried/am trying to implement is have a generic event object, which is dispatched to the relevant entities. Possibly with some form of observer/subscriber system.

src/ClientHandle.cpp
worktycho: Mostly its legacy code. We have a few hundred thousand lines of it, so updating everything to the latest practices is problematic. But lambdas are preferred for new code.

src/ClientHandle.cpp
bibo38: Wouldn't this just create for every Packet a new Class for the event?
Also it's harder to find bugs, when you use a Event queue to handle all the code.

src/ClientHandle.cpp
bibo38: Should I add the lambdas to the necessary functions in this PR?

src/ClientHandle.cpp
worktycho: Yes, which would split the code up. The problem with cClientHandle is that there is too much code in one module. And as for bugs, splitting the code up would make writing tests easier.

src/ClientHandle.cpp
worktycho: I think that would be a good idea.

bibo38:
@madmaxoft @NiLSPACE @bearbin @tigerw Have you looked at this or can it be squashed/merged?

bibo38:
Yeah, this was implemented in 598d90268f295e1666b2e626d3a2aabfd178b8b1. I'm just waiting for a feedback from the cuberite team.

src/Protocol/Protocol18x.cpp
```diff
@@ -247,6 +257,16 @@ void cProtocol180::SendBlockChanges(int a_ChunkX, int a_ChunkZ, const sSetBlockV
 
 
 
+void cProtocol180::SendCameraSetTo(const cEntity & a_Entity)
```
NiLSPACE: This packet allows the spectator to look through the eyes of a mob right? I'm not sure if this is the right name. Perhaps something like `SendSetMobView`

bibo38: You're right. According to [wiki.vg](http://wiki.vg/Protocol#Camera) this packet is just named `Camera`. But be aware, that this packet works with **every entity**, e.g. you can also look through the "eyes" of a minecart. Maybe name it `SendSetCameraView` or `SendSetEntityView` if you find the current text misleading?

LogicParrot:
Did anyone look at this?

jammet:
As of now, spectator-clicking on a mob or player actually attacks them, no camera mode yet.

SphinxC0re:
Any updates on this one?

LogicParrot:
Did anyone review this?

jammet:
Can't verify anything else at this moment, because no one else was on, today, but left clicking mobs in spectator still just attacks them instead of switching over to their camera.

LogicParrot:
@jammet this isn't merged with master yet. Unless you've compiled yourself, you won't see the changes.

jammet:
Ah, I see.

bibo38:
I've updated this PR to the latest master to allow a merge of this PR.
I would be happy, if someone could test the functionality, to be sure, that I've not missed something :smirk: (just clone my [branch](https://github.com/bibo38/Cuberite/tree/spectator) and compile it).

SphinxC0re:
Please rebase and squash your commits onto the current master

bibo38:
@SphinxC0re Done :yum:

mathias-github:
I will test the functionality, so we can get the PR merged as soon as possible.

mathias-github:
These are the issues I found:
- You can't see other players who have enabled spectator mode, if you have it enabled too. Players in spectator mode should look translucent.
- When in spectator mode, your head is not translucent.
- Cannot use spectator mode's teleport feature to teleport to other players in spectator mode. (not quite sure if this is the behaviour in vanilla as well)

mathias-github:
Before we merge this PR in the future, code for 1.9 would be good to have.

bibo38:
yep, I've tried to get this PR merged before the 1.9 thing...
I hope I get updated this soon.

src/Protocol/Protocol18x.cpp
```diff
@@ -2489,6 +2510,28 @@ void cProtocol180::HandlePacketSlotSelect(cByteBuffer & a_ByteBuffer)
...
+	{
+		return;
+	}
+
+	playerUUID.resize(32);
+	for (unsigned int i = 15; i < 16; i--)
+	{
+		playerUUID[i * 2 + 1] = ValueToHexDigit(playerUUID[i] & 0xf);
+		playerUUID[i * 2] = ValueToHexDigit(static_cast<UInt8>(playerUUID[i]) >> 4);
+	}
```
madmaxoft: This is a good candidate for a separate `cByteBuffer::ReadUUID()` function

bibo38: yep, you're right :yum:

SphinxC0re:
Don't merge. Rebase.
`git fetch`
`git rebase origin/master` or `git rebase upstream/master` if that is the remote name of the canon repo. After doing that, you have to force-push: `git push --force`

tigerw:
Why does it matter if squashing pushes it into one commit?

LogicParrot:
It does not.

madmaxoft:
It depends. Sometimes it makes sense to preserve the original commits, since it makes the history easier to understand in case we need to go back (bisect, blame etc.)

LogicParrot:
>It depends. Sometimes it makes sense to preserve the original commits,

We mean that @sphinxc0re  is asking OP to squash commits, but Github can do that automatically upon merge, so that request does not matter.

bibo38:
@LogicParrot Did I missed a Github feature? Is it not required for a PR to be manually squashed?

sphinxc0re:
AFAIK the sqash-merge does only that: Merging. Since we want a clean history, we require everyone to rebase onto the current master

madmaxoft:
The squash-merge does actually squash all the commits into a single one and rebases it onto the current master, so it is useful for those scenarios when we want the PR to be merged as a single commit. If we want to keep it in several commits, we still need to manually rebase and then merge.

PureTryOut:
I assume this has been merged, or at least partially? I can succesfully run `/gamemode spectator` in-game and I can noclip fine, I'm even hidden from other players. The only issue I see is that mobs still target you, which seems to be fixed in this PR.

jammet:
You also can't right click another player or mob to see through their eyes.

jammet:
Disregard my last message, I thought I was responding to the top of the
thread.

bibo38:
> ERROR: Newly undocumented API symbols found:
> cFloater.CanPickup (function)
> cFloater.GetAttachedMobID (function)
> cFloater.GetOwnerID (function)
> 
> The command "./CIbuild.sh" exited with 1.
(https://travis-ci.org/cuberite/cuberite/jobs/150280996#L2581)

Is this a CI problem? I don't see the problem with my PR.

bearbin:
Yes, this is a CI problem.

bibo38:
Currently everything works except the [issues reported by @mathiascode](https://github.com/cuberite/cuberite/pull/2852#issuecomment-218860366) (teleporting to other spectators is not possible in vanilla). I would be pleased if somebody knows how to fix these issues(missing packets, ...).

mathiascode:
Since everything works fine already, maybe we should merge this as it is, and fix minor bugs later?

madmaxoft:
Okay, I agree with a merge. @bibo38 would you mind if I squash-merged all your commits into a single one? That would be preferable to having "merged master" intermediate commits. Also, I need you to fix the merge conflict first.

bibo38:
@madmaxoft No problem. But the merging could take some time.

bibo38:
@madmaxoft Now it is merged to the current master. Only the ApiDesc.lua needed a manual merge.
Get it merged :wink:

madmaxoft:
Nooo, not a merge! **Rebase**

PR #2853, opened 2016-01-11
Added detailed logging to cBlockArea::Read() bad Y coords.

madmaxoft:
Ref.: http://forum.mc-server.org/showthread.php?tid=2307


NiLSPACE:
Perhaps we should hide these warnings in Release mode? Or we should add a debug mode for plugins where these warnings are shown. Not per-plugin, because that would be too difficult, but as a setting in the server.

madmaxoft:
If we hid them, we'd never really get the reports and never would fix such issues. I prefer people complaining to having hidden bugs.

NiLSPACE:
True, but if the problem is in a badly written plugin rather than the server it could get pretty annoying.

madmaxoft:
A badly written plugin can always be fixed

PR #2859, opened 2016-01-12
Fix mob attack interval

LogicParrot:
Before this commit, mobs' attack cooldown only ticks when they are stationary. This made them attack moving targets far less often. You had to stand still for a couple of ticks before a zombie attacked you.

### DIFF
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -75,7 +75,6 @@ void cAggressiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	if (TargetIsInRange() && !LineOfSight.Trace(MyHeadPosition, AttackDirection, static_cast<int>(AttackDirection.Length())) && (GetHealth() > 0.0))
 	{
 		// Attack if reached destination, target isn't null, and have a clear line of sight to target (so won't attack through walls)
-		StopMovingToPosition();
 		Attack(a_Dt);
 	}
 }
@@ -86,14 +85,13 @@ void cAggressiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 bool cAggressiveMonster::Attack(std::chrono::milliseconds a_Dt)
 {
-	m_AttackInterval += (static_cast<float>(a_Dt.count()) / 1000) * m_AttackRate;
-	if ((m_Target == nullptr) || (m_AttackInterval < 3.0))
+	if ((m_Target == nullptr) || (m_AttackCoolDownTicksLeft != 0))
 	{
 		return false;
 	}
 
 	// Setting this higher gives us more wiggle room for attackrate
-	m_AttackInterval = 0.0;
+	ResetAttackCooldown();
 	m_Target->TakeDamage(dtMobAttack, this, m_AttackDamage, 0);
 
 	return true;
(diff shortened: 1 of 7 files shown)


src/Mobs/AggressiveMonster.cpp
```diff
@@ -86,12 +85,13 @@ void cAggressiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 bool cAggressiveMonster::Attack(std::chrono::milliseconds a_Dt)
 {
-	m_AttackInterval += (static_cast<float>(a_Dt.count()) / 1000) * m_AttackRate;
+
 	if ((m_Target == nullptr) || (m_AttackInterval < 3.0))
 	{
 		return false;
 	}
 
+
```
madmaxoft: Any particular reason for a double empty line? A single would have sufficed, in my opinion.

madmaxoft:
I think the `m_AttackInterval` variable is poorly named, it's not clear what it does and what it contains. How about renaming it to something more reasonable, possibly with changing the logic behind it a bit? I was thinking, `m_AttackCooldownTicksLeft`, counting from the value set in the config down to zero and only attacking when it reaches 0. Also, it should count in ticks, not in wall-clock time.

LogicParrot:
I completely agree. I didn't know how exactly is `m_AttackInterval` being calculated (e.g. what's so special about 3.0?) so I left the calculations as they are. I'm afraid I won't have time to study this and apply the fixes any time soon. How about we merge this and apply your proposals in a later seperate PR?

madmaxoft:
It's just a constant, I guess.

Pity, I'd think this PR was the ideal chance of getting rid of this cruft, but if you don't think you can do it, then let's just postpone it.

LogicParrot:
The config only affect `m_AttackRate`, right?

madmaxoft:
Yes, the only point of all this code should be to configure the interval between mob attacks.

LogicParrot:
Could you double check `cMonster::ResetAttackCooldown()`? I think I've done it in a way that does not change the previous attack rates at all.

src/Mobs/AggressiveMonster.cpp
```diff
@@ -86,14 +85,13 @@ void cAggressiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
-	m_AttackInterval += (static_cast<float>(a_Dt.count()) / 1000) * m_AttackRate;
-	if ((m_Target == nullptr) || (m_AttackInterval < 3.0))
+	if ((m_Target == nullptr) || (m_AttackCoolDownTicksLeft != 0))
 	{
 		return false;
 	}
 
 	// Setting this higher gives us more wiggle room for attackrate
-	m_AttackInterval = 0.0;
+	m_AttackCoolDownTicksLeft = 20 * m_AttackRate * 3;
```
LogicParrot: Replace this with a call to  `cMonster::ResetAttackCooldown()`

src/Mobs/Monster.cpp
```diff
@@ -690,6 +695,15 @@ void cMonster::InStateEscaping(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
+void cMonster::ResetAttackCooldown()
+{
+	m_AttackCoolDownTicksLeft = 3 * 20 * m_AttackRate;  // A second has 20 ticks, an attack rate of 1 means 1 hit every 3 seconds
```
madmaxoft: I think this needs casts

madmaxoft: Yup, Clang complains.

LogicParrot:
Do not merge yet, this breaks skeletons.

LogicParrot:
Rebased, fixed skeleton issue, added @madmaxoft 's proposals.

madmaxoft:
Looks good to me.

PR #2860, opened 2016-01-12
Friendly wolf attack safety checks

LogicParrot:


src/Mobs/Wolf.cpp
```diff
@@ -74,9 +74,21 @@ bool cWolf::Attack(std::chrono::milliseconds a_Dt)
 
 void cWolf::NearbyPlayerIsFighting(cPlayer * a_Player, cEntity * a_Opponent)
 {
-	if ((m_Target == nullptr) && (a_Player->GetName() == m_OwnerName) && !IsSitting())
+	if ((m_Target == nullptr) && (a_Player->GetName() == m_OwnerName) && !IsSitting() && (a_Opponent->IsPawn()))
 	{
 		m_Target = a_Opponent;
+		if (m_Target->IsPlayer() && static_cast<cPlayer *>(m_Target)->GetName() == m_OwnerName)
+		{
+			m_Target = nullptr;  // Our owner must has self-inflicted damage. Abort attacking owner.
```
madmaxoft: `// Our owner has hurt himself, avoid attacking them`

madmaxoft:
Does that mean that hurting self will abort wolves attacking previous opponent? Not sure if that's what we want.

LogicParrot:
If we have a previous opponent then the first if block will not be entered thanks to `if (m_Target == nullptr)`.

madmaxoft:
Right, I missed that. Okay then. Fix the comment and in you go :)

PR #2862, opened 2016-01-12
Fixed creeper calling TargetIsInRange with null m_Target

LogicParrot:


madmaxoft:
Looks good to me.

LogicParrot:
I detected this thanks to the newly added assert @ `cMonster::TargetIsInRange`, illegally calling TargetIsInRange may still be a problem in other parts of the code. Those should assert now.

PR #2863, opened 2016-01-12
Do not spawn mobs too close to player

LogicParrot:
Mobs shouldn't spawn in a 24-block radius around the player. This commit takes care of that.

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -109,7 +109,7 @@ void cWorld::cTickThread::Execute(void)
 		auto WaitTime = std::chrono::duration_cast<std::chrono::milliseconds>(NowTime - LastTime);
 		m_World.Tick(WaitTime, TickTime);
 		TickTime = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - NowTime);
-		
+
 		if (TickTime < cTickTime(1))
 		{
 			// Stretch tick time until it's at least 1 tick
@@ -216,7 +216,7 @@ cWorld::~cWorld()
 	delete m_RedstoneSimulator;  m_RedstoneSimulator = nullptr;
 
 	UnloadUnusedChunks();
-	
+
 	m_Storage.WaitForFinish();
 
 	// Unload the scoreboard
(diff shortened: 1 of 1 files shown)


LogicParrot:
I used `cWorld::FindClosestPlayer`. Is it safe? a comment in its code warns about it.

src/Mobs/Creeper.cpp
```diff
@@ -27,7 +27,7 @@ void cCreeper::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	super::Tick(a_Dt, a_Chunk);
 
-	if (!TargetIsInRange() && !m_BurnedWithFlintAndSteel)
+	if ((m_Target == nullptr) || (!TargetIsInRange() && !m_BurnedWithFlintAndSteel))
```
LogicParrot: stray code

madmaxoft:
You should rebase onto current master (and work in the main repo instead of your own)

LogicParrot:
Rebased.

LogicParrot:
Should we have a policy regarding tabs? Different IDEs treat them differently and we often git mass tab changes when different devs work on one file.

madmaxoft:
Historically, we kept tabs on otherwise empty lines, because the then-used IDE of choice, MSVC2008, made it so. Nowadays we prefer no tabs on empty lines in new code, but don't want to touch old code just for this change, so we prefer to keep the formatting as-is, but accept PRs that change the formatting in files they touch.

madmaxoft:
The `FindClosestPlayer()` has a threading problem potential - as soon as it returns a `cPlayer` object, you cannot know if it hasn't been freed in the meantime from another thread.
Since you're only using it to check *whether* there is a player, you're on the safe side.

There is a plan to rewrite the function into a callback-based `DoWithClosestPlayer()`, it should be trivial to convert your code when it comes to this.

LogicParrot:
We have a problem. This neutralizes all egg spawns because they are too close to a player.

SphinxC0re:
So, this only applies for passive-aggressive and aggressive mobs?

LogicParrot:
@sphinxc0re I believe it applies to all mobs (it broke passive breeding at some point and then we fixed it)

PR #2868, opened 2016-01-13
Limited the username length to 16 characters.

### LINKED ISSUE #2545: Long username kicks the whole server
Whenever someone with a username longer than 16 characters joins the server (authentication off), all players online disconnect with the following message:
screenshot from 2015-10-13 19 42 07

and sometimes this one:
screenshot from 2015-10-13 19 39 42

bibo38:
This fixes a client crash, because Minecraft requires that a username is not longer than 16 characters.
See also: http://minecraft.gamepedia.com/The_Player#Name
Fixes #2545

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -89,6 +89,8 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 	m_UUID((a_Client != nullptr) ? a_Client->GetUUID() : ""),
 	m_CustomName("")
 {
+	ASSERT(a_PlayerName.length() <= 16);  // Otherwise this player could crash many clients...
+
 	m_InventoryWindow = new cInventoryWindow(*this);
 	m_CurrentWindow = m_InventoryWindow;
 	m_InventoryWindow->OpenedByPlayer(*this);
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1719,6 +1719,12 @@ bool cClientHandle::CheckMultiLogin(const AString & a_Username)
 
 bool cClientHandle::HandleHandshake(const AString & a_Username)
 {
+	if (a_Username.length() > 16)
+	{
+		Kick("Your username is too long(>16 characters)");
+		return false;
+	}
+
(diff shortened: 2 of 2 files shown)


src/Entities/Player.cpp
```diff
@@ -89,6 +89,12 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 	m_UUID((a_Client != nullptr) ? a_Client->GetUUID() : ""),
 	m_CustomName("")
 {
+	if (a_PlayerName.empty() || a_PlayerName.length() > 16)
+	{
+		LOGERROR("Playername of %s is too long(>16) or empty", a_PlayerName.c_str());
+		VERIFY(false);
+	}
```
madmaxoft: This is the worst construct you could have used.
We use `LOGERROR` only for stuff that the server cannot recover from and terminates. A long username should definitely NOT be such a case.
And `VERIFY(false)` is the stupidest construct ever. You're basically saying "I'm not gonna continue, but I'm not telling you why."

If the playername is supposed to already be truncated before getting here, use an ASSERT, if it isn't, then truncate it silently here.

madmaxoft: Also, with this change, a malicious client could crash the server at any time by sending an empty username.

bibo38: Asserting something on an input variable is a not a good practise, so I only asserted it in debug mode with `VERIFY` and gave an fatal error in the log. In my opinion this should never happen, but I don't want to just crash the server because somebody created an invalid cPlayer for something.

madmaxoft: You've got `ASSERT` and `VERIFY` mixed up. `ASSERT` only fires in debug mode and is compiled out in release mode, `VERIFY` is the same as `ASSERT` in debug mode and keep the code in its parameter in Release mode.
In this case it evaluates to `false;` which is a valid C++ expression saying "no operation".

Still, the proper thing to do is to assert if you expect this condition already handled elsewhere, or fix it silently if you expect to actually receive the bad data.

bibo38:
Better?

madmaxoft:
Thanks

src/Entities/Player.cpp
juniormatcha: I don't think there has any problems

PR #2869, opened 2016-01-13
compile.sh - 70 columns print limit

LogicParrot:


compile.sh
```diff
@@ -121,18 +123,18 @@ inactiveCode ()
...
-echo "		but using it and finding and reporting bugs helps us a lot!"
-echo
-echo "* (D)Dev:	The least stable of the three. (Master branch)"
-echo "		Choose the development branch if you are feeling adventurous and"
-echo "		want to try new, bleeding edge features."
-echo
+echo "
+You can choose between 3 branches:
+* (S)Stable:	Choose the stable branch if you want the most
+		reliable server.
```
madmaxoft: You shouldn't be using tabs for vertical alignment

LogicParrot: Done.

PR #2871, opened 2016-01-13
Fixed race conditions

Seadragon91:
Fixed a few race conditions.

`m_TicksSinceLastPacket`is accessed on line [ClientHandle.cpp:2018](https://github.com/cuberite/cuberite/blob/master/src/ClientHandle.cpp#L2018) and line [ClientHandle.cpp:3063](https://github.com/cuberite/cuberite/blob/master/src/ClientHandle.cpp#L3063).

Reverted: ~~[`LogSimple`](https://github.com/cuberite/cuberite/blob/master/src/Logger.cpp#L33) is called from different threads on startup, moving the lock to the top of the method fixes the race.~~

`m_bRestarting` is accessed fom `Server::Tick`, [Server.cpp:639](https://github.com/cuberite/cuberite/blob/master/src/Server.cpp#L639) and [Server.cpp:335](https://github.com/cuberite/cuberite/blob/master/src/Server.cpp#L639)

### DIFF
--- a/src/ClientHandle.h
+++ b/src/ClientHandle.h
@@ -415,7 +415,7 @@ private:
 	int m_LastStreamedChunkZ;
 
 	/** Number of ticks since the last network packet was received (increased in Tick(), reset in OnReceivedData()) */
-	int m_TicksSinceLastPacket;
+	std::atomic<int> m_TicksSinceLastPacket;
 	
 	/** Duration of the last completed client ping. */
 	std::chrono::steady_clock::duration m_Ping;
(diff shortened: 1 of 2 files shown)


src/Logger.cpp
```diff
@@ -52,7 +53,6 @@ void cLogger::LogSimple(AString a_Message, eLogLevel a_LogLevel)
 	#endif
 
 
-	cCSLock Lock(m_CriticalSection);
```
madmaxoft: This change is not needed, logging is supposed to be thread-unsafe on the line-level - unrelated threads have no proper ordering.
You're now creating more load and more contention on the logging mutex, and there's no real benefit, threads' output can still be mixed line-wise in any order.

Seadragon91: Reverted this change.

PR #2872, opened 2016-01-15
Lua callback

madmaxoft:
This implements a class for easy implementation of callbacks to Lua. A simple `GetStackValue` can be used to remember a callback and use it any time later on, even after the plugin has been unloaded (the callback reports this, instead of calling invalid Lua state).

Reimplemented the WebAdmin tab callbacks to use this new class. Used the opportunity to clean up the WebAdmin API, differentiate between the tab's title and URL, removed the `cWebPlugin` altogether, and added API docs regarding the webadmin.

Progress:
  - [x] Implement `cLuaState::cCallback`
  - [x] Change `cWebAdmin` to use the new Lua callbacks
  - [x] Change plugin hooks to use the new Lua callbacks
  - [x] Change world task scheduling to use the new Lua callbacks
  - [x] Change `cLuaWindow` to use the new Lua callbacks


LogicParrot:
```
/home/travis/build/cuberite/cuberite/src/WebAdmin.cpp:53:2: error: field

      'm_IsRunning' will be initialized after field 'm_TemplateScript'

      [-Werror,-Wreorder]

        m_IsRunning(false),

        ^

1 error generated.

make[2]: *** [src/CMakeFiles/Cuberite.dir/WebAdmin.cpp.o] Error 1

make[2]: *** Waiting for unfinished jobs....

make[1]: *** [src/CMakeFiles/Cuberite.dir/all] Error 2

make: *** [all] Error 2
```

src/Bindings/LuaState.h
```diff
@@ -112,6 +115,63 @@ class cLuaState
...
+		~cCallback()
+		{
+			Clear();
+		}
+
+		/** Calls the Lua callback, if still available.
+		Returns true if callback has been called.
+		Returns false if the Lua state isn't valid anymore. */
+		template <typename... Args>
+		bool Call(Args &&... args)
```
worktycho: Why not implement operator()? Then we can pass this to functions taking std::function

madmaxoft: Two reasons:
- I don't want to run into the same problem as `cWorld::ScheduleTask()`, which now accepts a `std::function` and so can accept lambdas, but if you pass in a lambda, the server will crash because the lambda object is no longer valid when the scheduled callback is called
- I'm using a `SharedPtr<cCallbackContainer>` for storage, so actually caling the callback is via `container->m_Callback->Call(...)`; if I had `operator ()`, the call would look weird: `container->m_Callback->operator()(...)`

worktycho: The lambda object should be valid for as long as the std::function object is, because its a value type. If its the members that have insufficient lifetimes, then I don't see what lambdas have to do with it, you get exactly the same problems from manual functor classes. 
Also I don't see any references to cCallbackContainer in the PR, so I don't really understand the second point.

tigerw: TL;DR: xoft, you have no idea what you are talking about :P

madmaxoft: Yes, and that's why I wrote a little experiment to help me understand that better: https://github.com/madmaxoft/LambdaLifetimeTest

I've found that the lambdas do stay valid, because the code uses excessive copying. Two things that are somewhat troubling is that the lambda's by-reference captures go out of scope, yet the lambda can access them (-> possible crash with accessing invalid memory) and we can't pass in non-copyable objects (such as Lua `cCallback`)

worktycho: The best way of thinking about lambdas is that they are syntax sugar. Each lambda is converted to an anonymous class, with members for each capture, and an `operator()` for the code. So 
```
auto x = [=foo, &bar](int baz) {return foo + bar + baz; };
```
is equivalent to:
```
class _AJFLKD
{
public:
    _AJFLKD(decltype(foo) a_foo, decltype(bar)& a_bar) : m_foo(a_foo), m_bar(a_bar) {}
    int operator()(int baz)
    {
        return m_foo + m_bar + baz;
    }
private:
    decltype(foo) m_foo;
    decltype(bar)& m_bar;
}
```

Admittedly some of the mistakes with references are slightly easier to make with that manual functors, but they are just as possible.

madmaxoft: Ah, sorry, I forgot to push the code to that repo, I was on a train and the net wasn't working.

I wasn't sure if we were storing the lambdas as values, as references or as pointers.

Anyway, as for this PR, I don't think it's worth it, because we're not using the class as an interface of any kind, it's a regular value class and I don't think it's gonna be used directly as a callback. I could change it if you really want me to, I've tested it yesterday and it didn't seem to produce anything awkward and it's just some 4 lines of changes.

worktycho: Leave it. It doesn't really matter.

src/Bindings/LuaState.h
```diff
@@ -112,6 +115,63 @@ class cLuaState
...
+		/** Reference to the Lua callback */
+		cRef m_Ref;
+
+
+		/** Invalidates the callback, without untracking it from the cLuaState.
+		Called only from cLuaState when closing the Lua state. */
+		void Invalidate(void);
+
+		/** This class cannot be copied, you can only std::move() it. */
+		cCallback(const cCallback &) = delete;
```
worktycho: Do we really want to be able to move this class? You're using address equality in the track functions, so wouldn't moving the callback break that?

madmaxoft: Good catch, I need to examine that more thoroughly.

madmaxoft:
This is now ready for a full review.

The main reasoning for the existence of the new `cLuaState::cCallback` class and for the extreme work it does in avoiding mutexes is that we want to support multithreading in plugins in the future - one plugin could create multiple Lua states, either dependent (soft-threads) or even independent (separate states); and we want to have a mechanism allowing callback registration that is not per-plugin, but rather per-state. In order to avoid deadlock possibilities while two such threads compete over the (re-)registration of the callbacks, I had to disable both the copy-constructor and the move-constructor.

src/Bindings/PluginLua.h
```diff
@@ -270,6 +222,28 @@ class cPluginLua :
 
 	/** Releases all Lua references, notifies and removes all m_Resettables[] and closes the m_LuaState. */
 	void Close(void);
+
+	/** Removes all WebTabs currently registered for this plugin from the WebAdmin. */
+	void ClearWebTabs(void);
+
+	/** Calls a hook that has the simple format - single bool return value specifying whether the chain should continue. */
```
NiLSPACE: Is it perhaps usefull to note here how the other hooks are called? Like, do they have a specific method here, or is that done manually somewhere else?

src/WebAdmin.cpp
```diff
@@ -467,20 +449,24 @@ AString cWebAdmin::GetContentTypeFromFileExt(const AString & a_FileExtension)
 		ContentTypeMap["jpe"]  = "image/jpeg";
 		ContentTypeMap["tiff"] = "image/tiff";
 		ContentTypeMap["ico"]  = "image/ico";
-		ContentTypeMap["csv"]  = "image/comma-separated-values";
+		ContentTypeMap["csv"]   = "text/csv";
```
NiLSPACE: Is it just me or are the tabs here wrong?

src/WebAdmin.cpp
```diff
@@ -489,86 +475,93 @@ AString cWebAdmin::GetContentTypeFromFileExt(const AString & a_FileExtension)
...
 			}
 		}
 	}
 
-	// Return the page contents
-	return Page;
+	// If a WebTab handler was found, call it:
+	if (tab != nullptr)
+	{
+		page.ContentType = "text/html";  // Default to HTML content type, unless overridden by a plugin
```
NiLSPACE: Does this mean plugins can change the content type even for 'normal' pages - eg, not with `~`?

src/WebAdmin.cpp
```diff
@@ -489,86 +475,93 @@ AString cWebAdmin::GetContentTypeFromFileExt(const AString & a_FileExtension)
...
+		if (!tab->m_Callback->Call(a_Request, split[1], page.Content, page.ContentType))
+		{
+			page.Content = GetHTMLEscapedString(Printf(
+				"WebTab callback for plugin %s, page %s has failed.",
+				tab->m_PluginName.c_str(), tab->m_Title.c_str()
+			));
 }
```
NiLSPACE: Indentation

src/WebAdmin.cpp
```diff
@@ -489,86 +475,93 @@ AString cWebAdmin::GetContentTypeFromFileExt(const AString & a_FileExtension)
...
+void cWebAdmin::AddWebTab(
+	const AString & a_Title,
+	const AString & a_UrlPath,
+	const AString & a_PluginName,
+	SharedPtr<cWebAdmin::cWebTabCallback> a_Callback
+)
 	{
-		World->ForEachPlayer(PlayerAccum);
-		Content.append(PlayerAccum.m_Contents);
+	cCSLock lock(m_CS);
```
NiLSPACE: Indentation

src/WebAdmin.cpp
```diff
@@ -489,86 +475,93 @@ AString cWebAdmin::GetContentTypeFromFileExt(const AString & a_FileExtension)
...
 {
-	return GetBaseURL(StringSplit(a_URL, "/"));
+	cCSLock lock(m_CS);
+	for (auto itr = m_WebTabs.begin(), end = m_WebTabs.end(); itr != end; ++itr)
+	{
+		if ((*itr)->m_UrlPath == a_UrlPath)
+		{
+			m_WebTabs.erase(itr);
+			return true;
+}
```
NiLSPACE: Indentation

NiLSPACE:
Perhaps we can move the webtab registration to the `Info.lua` file as well.

Other than a few questions and indentation errors this looks great :)

src/WebAdmin.cpp
```diff
@@ -255,17 +280,20 @@ void cWebAdmin::HandleWebadminRequest(cHTTPServerConnection & a_Connection, cHTT
 	}
 
 	// Check auth:
+	{
+		cCSLock Lock(m_CS);
 	AString UserPassword = m_IniFile.GetValue("User:" + a_Request.GetAuthUsername(), "Password", "");
```
NiLSPACE: Indentation is wrong here now

madmaxoft:
Huh, so many indent errors, it must have been the merge conflicts I had while rebasing this after the big tab rewrite. Hopefully fixed all of them.

src/WebAdmin.cpp
madmaxoft: Yes, plugins can override the ContentType of any page. Doesn't make much sense for the regular pages, but it's not really worth the trouble distinguishing between the two.

madmaxoft:
I don't want to do much about the webadmin tabs just yet, I have a complete redesign of the webadmin in my head so it would be kinda useless to change things now only to have them change completely again soon.

But yes, Info.lua is part of the new plans.

src/Bindings/LuaState.cpp
```diff
@@ -39,6 +43,10 @@ extern "C"
 
 const cLuaState::cRet cLuaState::Return = {};
 
+/** Each Lua state stores a pointer to its creating cLuaState in Lua globals, under this name.
+This way any cLuaState can reference the main cLuaState's TrackedCallbacks, mutex etc. */
+static const char * g_CanonLuaStateGlobalName = "__Cuberite_CanonLuaState";
```
NiLSPACE: We also have `_CuberiteInternal_PluginName`, but that was/is exported by accident. I assume this one isn't exported?

NiLSPACE:
This looks great, no objection here.

src/Bindings/LuaState.cpp
madmaxoft: Both are exported, they need to be - the bindings code needs to read them from the Lua engine in order to know which Plugin / LuaState is currently calling the function.

They should be explicitly ignored in the LuaAPI documentation, so that there's no mention of them in the API docs.

I should probably unify the names, though.

src/Bindings/PluginLua.h
```diff
@@ -270,6 +222,29 @@ class cPluginLua :
...
+		for (auto & hook: hooks)
+		{
+			hook->Call(std::forward<Args>(a_Args)..., cLuaState::Return, res);
+			if (res)
+			{
+				// Hook wants to terminate the chain processing
+				return true;
+			}
+		}
+		return true;
```
madmaxoft: Oy, a glaring error here, should return false.

madmaxoft:
Okay, finally done the last part, `cLuaWindow` now uses the `cLuaState::cCallback`, too.

I have checked the basic functionality with a client, it seems to work alright.

PR #2876, opened 2016-01-16
Fix egg spawning

### LINKED ISSUE #2873: Egg spawning was broken by #2863
#2863

LogicParrot:
[Obsolete]
Do not merge yet.
This is supposed to fix #2873.
However, the new chunk assert I added is surely to fail after just a few ticks. Which makes me wonder why `cMobSpawner::CanSpawnHere`  didn't crash before.
 
The code which is nulling the chunk is probably `a_Chunk = a_Chunk->GetRelNeighborChunkAdjusCoords` at line 364 in MobSpawner.cpp. 

Should I just bailout if the chunk is a nullptr? Or is there a deeper problem here?

### DIFF
--- a/src/MobSpawner.cpp
+++ b/src/MobSpawner.cpp
@@ -3,7 +3,7 @@
 
 #include "MobSpawner.h"
 #include "Mobs/IncludeAllMonsters.h"
-
+#include "World.h"
 
 
 
@@ -84,7 +84,7 @@ eMonsterType cMobSpawner::ChooseMobType(EMCSBiome a_Biome)
 		addIfAllowed(mtCreeper, allowedMobs);
 		addIfAllowed(mtSquid, allowedMobs);
 		addIfAllowed(mtGuardian, allowedMobs);
-		
+
 		if ((a_Biome != biDesert) && (a_Biome != biBeach) && (a_Biome != biOcean))
 		{
 			addIfAllowed(mtSheep, allowedMobs);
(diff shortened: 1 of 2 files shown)


LogicParrot:
I now believe that it didn't crash before because the code path at `cMobSpawner::canSpawnHere` luckily happens to never touch any of the a_Chunk's fields if a_Chunk is a nullptr, and bails out early at the first if block in that function. So this seems like a dormant ticking bomb. I'm going to add a bailout of a_Chunk is a nullptr at  `cMobSpawner::canSpawnHere`  and I think it's fine. Can someone analyze this and confirm?

LogicParrot:
I now believe that it didn't crash before because the code path at `cMobSpawner::canSpawnHere` luckily happens to never touch any of the a_Chunk's fields if a_Chunk is a nullptr. 

So this seems like a dormant ticking bomb. I'm going to replace the assert with a bailout at  `cMobSpawner::canSpawnHere`  and I think it's fine. Can someone analyze this and confirm?

LogicParrot:
I now believe that it didn't crash before because the code path at `cMobSpawner::canSpawnHere` luckily happens to never touch any of the a_Chunk's fields if a_Chunk is a nullptr. 

So this seems like a dormant ticking bomb. I'm going to replace the assert with a bailout at  `cMobSpawner::canSpawnHere`  and I think that should fix it all. Can someone analyze this and confirm?

LogicParrot:
I have submitted the latest commits to #2877 to keep my comments relevant to the code in this PR.

PR #2878, opened 2016-01-16
Fix crash when tamed wolf is hit by arrows

LogicParrot:


LogicParrot:
Projectiles are buggy, the problem is that an entity's `doTakeDamage` can be called with a non-pawn TDI.attacker (typically the arrow entity). This patches it and fixes a crash by nullifying TDI.attacker if it's not a cPawn, but a better way to do this should be implemetented.

LogicParrot:
I typically wait for atleast one reviewer, but I want to stabilize master asap. Merging. 

Some followup work should be done:
 - Investigate whether holding a raw entity in cMonster::m_Target is safe.
 - Investigate whether TDI.Attacker should be converted to a cPawn
 - When `cEntity::doTakeDamage` is triggered by a projectile, make the projectile pass a proper `cEntity*` or perhaps a proper entity ID.

src/Entities/Entity.cpp
```diff
@@ -242,10 +242,17 @@ void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_R
...
 	TDI.DamageType = a_DamageType;
-	TDI.Attacker = a_Attacker;
+	if ((a_Attacker != nullptr) && a_Attacker->IsPawn())
+	{
+		TDI.Attacker = a_Attacker;
+	}
+	else
+	{
+		TDI.Attacker = nullptr;
+	}
```
madmaxoft: This change potentially breaks plugins that assume that projectile damage comes through the hook as well. There's now no way to report projectile damage to plugins.

We should keep the projectile as the attacker and fix whatever problem that causes.

LogicParrot: In #2882, TDI.Attacker will always point to the actual attacker. If it's a projectile, then the attacker is the shooter of the projectile. There is no sense at all in TDI.Attacker pointing at the projectile. But perhaps TDI should have an extra field, damageType or something, which indicates how the damage occured.

LogicParrot: >We should keep the projectile as the attacker and fix whatever problem that causes.

The problem that originally caused is that wolves "attacked" arrows because they thought they were the attackers, and they never had access to the original attacker. As to why that crashes, I'm not sure, it could be wolves holding pointers to long-gone projectiles. I am currently investigating m_Target in general, it looks a bit dangerous the way it currently works.

Assuming that TDI.Attacker is always a cPawn really simplifies things. Do you have a proposal which keeps things simple while still retaining plugin compatiblity?

LogicParrot: Maybe I could do it that way:
```
TDI.Attacker - The entity that caused the damage, could be a projectile or a mob or a nullptr
TDI.Shooter - If TDI.Attacker is a projectile, this is the shooter of that projectile. Otherwise this is null.
```

Then in cMonster:
```
if (TDI.shooter != nullptr) m_Target = TDI.shooter;
else m_Target = TDI.Attacker;
```

That shouldn't harm plugins.

LogicParrot:
Also, how come plugins have access to a raw entity pointer? (TDI.Attacker)
Isn't that very dangerous?

src/Entities/Entity.cpp
LogicParrot: Also, how come plugins have access to a raw entity pointer? (TDI.Attacker)
Isn't that very dangerous?

madmaxoft: I think keeping the `TDI.Attacker` as the projectile should work, and whoever needs the shooter could do just:
```cpp
if (TDI.Attacker->IsProjectile())
{
  // Set target to TDI.Attacker's Creator
}
```

LogicParrot: Yes, that would work.

But the whole story is a bit inelegant. 
 1. This will force all users of TDI to come up with such checks, creating dupes. 
 2. Most callers will never care about the projectile entity that hit them and just want to reach the attacker.
 3. The name TDI.Attacker implies an **attacker**.

Can't we break compatibility this one time and fix this? We don't have too many plugins anyways. Fixing now is easier.

If you disagree on breaking compatilbility, I still think TDI.Attacker should directly contain the shooter for convenience.

LogicParrot:
Yes, that would work.

But the whole story is a bit inelegant. 
 1. This will force all users of TDI to come up with such checks, creating dupes. 
 2. Most callers will never care about the projectile entity that hit them and just want to reach the attacker.
 3. The name TDI.Attacker implies an **attacker**.

Can't we break compatibility this one time and fix this? We don't have too many plugins anyways. Fixing now is easier.

If you disagree on breaking compatilbility, I still think TDI should directly contain the shooter for convenience.

Edit: `s/TDI.attacker/TDI`

LogicParrot:
Yes, that would work.

But the whole story is a bit inelegant. 
 1. This will force all users of TDI to come up with such checks, creating dupes. 
 2. Most callers will never care about the projectile entity that hit them and just want to reach the attacker.
 3. The name TDI.Attacker implies an **attacker**.

Can't we break compatibility this one time and fix this? We don't have too many plugins anyways. Fixing now is easier.

If you disagree on breaking compatilbility, I still think TDI should directly contain the shooter for convenience and preventing dupes.

LogicParrot:
Yes, that would work.

But the whole story is a bit inelegant. 
 1. This will force all users of TDI to come up with such checks, creating dupes. 
 2. Most callers will never care about the projectile entity that hit them and just want to reach the attacker.
 3. The name TDI.Attacker implies an **attacker**.

Can't we break compatibility this one time and fix this? We don't have too many plugins anyways. Fixing now is easier.

If you disagree on breaking compatilbility, I still think TDI should directly contain the shooter for convenience and preventing dupes, under a different variable (e.g. TDI.Shooter). That would at least treat point 1 and 2.

src/Entities/Entity.cpp
LogicParrot: Yes, that would work.

But the whole story is a bit inelegant. 
 1. This will force all users of TDI to come up with such checks, creating dupes. 
 2. Most callers will never care about the projectile entity that hit them and just want to reach the attacker.
 3. The name TDI.Attacker implies an **attacker**.

Can't we break compatibility this one time and fix this? We don't have too many plugins anyways. Fixing now is easier.

If you disagree on breaking compatilbility, I still think TDI should directly contain the shooter for convenience and preventing dupes, under a different variable (e.g. TDI.Shooter). That would at least treat point 1 and 2.

LogicParrot: @madmaxoft I just realized having a raw pointer to the initial shooter is dangerous and could leave a dangling pointer if the attacker is destroyed before the arrow hits someone.

LogicParrot: I deleted two comments which were irrelevant.

LogicParrot: @madmaxoft The proposed method produces code dupes, and forces TDI users to worry about things they shouldn't worry about. For instance, I'd need this in `cMonster.cpp::doTakeDamage`:

```
		if (a_TDI.Attacker->IsProjectile())
		{
			m_Target = nullptr;
			UInt32 AttackerId = static_cast<cProjectileEntity*>(a_TDI.Attacker)->GetCreatorUniqueID();

			class cCallback :
				public cEntityCallback
			{
			public:
				cPawn * m_PawnAttacker;
				virtual bool Item(cEntity * a_PawnAttacker) override
				{
					m_PawnAttacker = static_cast<cPawn*>(a_PawnAttacker);
					return m_PawnAttacker;
				}
			} Callback;
			Callback.m_PawnAttacker = nullptr;
			m_World->DoWithEntityByID(AttackerId, Callback);  // If the projectile shooter still exists, Callback.m_PawnAttacker will point to it
			m_Target = Callback.m_PawnAttacker;
		}
```

I'd need something similar in the player's `doTakeDamage` so that the player asks the wolves for help. And all plugins would need some similar piece of code.

If TDI.Attacker contained the actual attacker, the users of TDI wouldn't have to care about *how* the attacker dealt damage, unless they are actually interested in that.

I propse this:
```
TDI.PawnAttacker - The player/Mob who caused the damage. nullptr if no player or mob were involved. If this is projectile damage, PawnAttacker is the shooter, and Attacker is the projectile. If this is melee damage, PawnAttacker and Attacker are identical.
TDI.Attacker - The entity which dealt the damage. Might be a projectile, a mob, or a null. See TDI.PawnAttacker.
```

LogicParrot: This preserves pulgin compatibility.

LogicParrot: Alternativaly:
```
cPawn * TDI.PawnAttacker - The pawn (player or mob) which caused the damage. nullptr if no pawn is involved.
cProjectile * TDI.ProjectileAttacker - The projectile which caused the damage. nullptr if no projectile is involved.
cEntity * TDI.Attacker - DEPRECATED. If melee damage, points to the pawn. If projectile damage, points to the projectile. Otherwise a nullptr. DO NOT USE IN NEW PLUGINS.
```
And we can phase out `TDI.Attacker` in the future.

LogicParrot: This also make perfect sense for projectiles that do not have a cPawn (e.g. dispenser projectile).

PR #2880, opened 2016-01-16
Improved tamed wolf pack cooperation

LogicParrot:
Tamed wolves now:
 - Defend each other, not just the player.
 - Spread their targeting more evenly. Previously they all often attacked the same target.


LogicParrot:
Do not merge yet, I found an extra tweak which should improve this.

PR #2882, opened 2016-01-16
Improved tamed wolf pack cooperation and projectile reactions

LogicParrot:
Also fixes the projectile issues mentioned in #2878
Wolves now:
- Defend each other, not just the player
- Never do nothing when a fellow wolf is fighting
- Spread their targeting more evenly. Previously they all often attacked the same target.
- Respond to arrow hits targeted at their owner or at their allied wolves
- Respond to arrow hits initiated by their owner

I also doubled wolf damage because currently wolves are rather week in comparison to Vanilla.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -238,6 +238,47 @@ void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_R
 
 
 
+void cEntity::TakeDamage(eDamageType a_DamageType, UInt32 a_AttackerID, int a_RawDamage, double a_KnockbackAmount)
+{
+	class cNotifyWolves : public cEntityCallback
+	{
+	public:
+
+		cEntity * m_Entity;
+		eDamageType m_DamageType;
+		int m_RawDamage;
+		double m_KnockbackAmount;
+
+		virtual bool Item(cEntity * a_Attacker) override
+		{
+			cPawn * Attacker;
+			if (a_Attacker->IsPawn())
+			{
+				Attacker = static_cast<cPawn*>(a_Attacker);
+			}
+			else
+			{
+				Attacker = nullptr;
+			}
+
+
+			int FinalDamage = m_RawDamage - m_Entity->GetArmorCoverAgainst(Attacker, m_DamageType, m_RawDamage);
+			m_Entity->TakeDamage(m_DamageType, Attacker, m_RawDamage, FinalDamage, m_KnockbackAmount);
+			return true;
+		}
+	} Callback;
+
+	Callback.m_Entity = this;
+	Callback.m_DamageType = a_DamageType;
+	Callback.m_RawDamage = a_RawDamage;
+	Callback.m_KnockbackAmount = a_KnockbackAmount;
+	m_World->DoWithEntityByID(a_AttackerID, Callback);
+}
+
+
+
+
+
 void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_RawDamage, int a_FinalDamage, double a_KnockbackAmount)
 {
 	TakeDamageInfo TDI;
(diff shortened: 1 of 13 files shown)


src/Entities/Entity.cpp
```diff
@@ -238,6 +238,37 @@ void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_R
...
+			{
+				m_Attacker = nullptr;
+			}
+			return true;
+		}
+	} Callback;
+
+	m_World->DoWithEntityByID(a_AttackerID, Callback);
+
+	int FinalDamage = a_RawDamage - GetArmorCoverAgainst(Callback.m_Attacker, a_DamageType, a_RawDamage);
```
LogicParrot: Is it safe to extract a field out of the callback and use it?

src/Entities/Entity.cpp
```diff
@@ -238,6 +238,40 @@ void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_R
...
+			{
+				m_Attacker = nullptr;
+			}
+			return true;
+		}
+	} Callback;
+
+
+	m_World->DoWithEntityByID(a_AttackerID, Callback);
+	Attacker = Callback.m_Attacker;
```
LogicParrot: Is it safe to extract a field out of the callback and use it?

worktycho: Maybe. It depends on the thread and lock guarantees that DoTakeDamage gets from its caller.
But it would be safer to do the cEntity::TakeDamage call from the callback. It might be easier to add a std::function overload to cWorld::DoWithEntityByID, and use lambdas than do manual captures in cNotifyWolves.

LogicParrot: Should be definitely easier. Is there any objection to eventually getting rid of all Callback classes project-wide, replacing them with lambdas?

LogicParrot: Should definitely be easier. Is there any objection/problem with eventually getting rid of all Callback classes project-wide, replacing them with lambdas?

LogicParrot:
I doubled wolf damage because currently wolves are rather week in comparison to Vanilla.

src/Mobs/Wolf.cpp
```diff
@@ -29,15 +30,37 @@ cWolf::cWolf(void) :
...
 {
+	cEntity * PreviousTarget = m_Target;
 	if (!super::DoTakeDamage(a_TDI))
 	{
 		return false;
 	}
 
-	if (!m_IsTame)
+
+	cPawn * Pawn = static_cast<cPawn*>(m_Target);
```
Seadragon91: I tested your code. The server crashed here and m_Target was nullptr. The wolf wasn't tamed,

LogicParrot: Thank you. Fixed in the latest commit.

LogicParrot: This did not fix it thoroughly. One moment.

LogicParrot: Should be working now.

Seadragon91: Works fine now, thanks.

src/Mobs/Wolf.h
```diff
@@ -64,6 +66,7 @@ class cWolf :
 	AString m_OwnerName;
 	AString m_OwnerUUID;
 	int     m_CollarColor;
+	int	    m_NotificationCooldown;
```
NiLSPACE: You've used a tab here

LogicParrot: Fixed.

LogicParrot:
Rebased.

Seadragon91:
Interesting crash, I copied it out.

LogicParrot:
ftr: Not PR-related.

LogicParrot:
Rebased to re-trigger CI.

src/Entities/Entity.cpp
worktycho: @madmaxoft has some objections in #2872, but I'm not sure whether those are lambda specific, or just general lifetime issues.

LogicParrot:
Do not merge before we finish discussing TDI.Attacker in #2878

LogicParrot:
This does not make TDI.Attacker any worse than #2878, merging. (And we'll possibly undo the TDI.Attacker changes according to the discussiong at #2878)

PR #2884, opened 2016-01-17
Sorted .ini and .txt files alphabetically

LogicParrot:
Experimental auto sort. Looks good, but please review before merging.


LogicParrot:
This looks harmless enough. Any objections to merge? Does anything depend on ordering in these files?

madmaxoft:
Looks good to me, there shouldn't be any dependency on the order.

PR #2885, opened 2016-01-17
No second spawn broadcast necessary

SteveWinfield:
These entities are already getting broadcasted in cEntity::Initialize


LogicParrot:
Thank you :)

PR #2894, opened 2016-01-18
Spiders now friendly at daylight, new cChunk functions

LogicParrot:
1. Spiders are now friendly in high light levels and aggressive otherwise.

2. I noticed there's much duplicate boilerplate code when it comes to entities querying chunks. So I added some cChunk functions which should eliminate the dupes. Previously, one needed this code, or some variant, in order to query a block for light / metadata:

```
	cChunk * Chunk = m_Chunk->GetNeighborChunk(GetPosition().x, GetPosition().z);
	if ((Chunk == nullptr) || !Chunk->IsValid())
	{
		return;
	}

	BLOCKTYPE BlockType;
	NIBBLETYPE BlockMeta;
	RelX = Location.x - Chunk->GetPosX() * cChunkDef::Width;
	RelZ = Location.z - Chunk->GetPosZ() * cChunkDef::Width;
	RelY = Location.y

	m_Chunk->GetBlockTypeMeta(RelX, Location.y, RelZ, BlockType, BlockMeta);
```

That's too much code duplication, and it can be found in various forms all over the place.
With the added functions, this is how it looks like:

```
	PREPARE_REL_AND_CHUNK(GetPosition(), a_Chunk);
	PREPARE_BLOCKDATA; 
	if (!RelSuccess)
	{
		return;
	}
	Chunk->GetBlockTypeMeta(Rel.x, Rel.y, Rel.z, BlockType, BlockMeta);
```

The two convenience macros are:
```
#define PREPARE_REL_AND_CHUNK(Position, OriginalChunk) cChunk * Chunk; Vector3i Rel; bool RelSuccess = (OriginalChunk).GetChunkAndRelByAbsolute(Position, &Chunk, Rel);
#define PREPARE_BLOCKDATA BLOCKTYPE BlockType; NIBBLETYPE BlockMeta;
```

This eliminates Rel/chunk calculation dupes. I'd like to later make all entity queries use the new form. Any objections?

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -123,9 +123,9 @@ cChunk::cChunk(
 cChunk::~cChunk()
 {
 	cPluginManager::Get()->CallHookChunkUnloaded(*m_World, m_PosX, m_PosZ);
-	
+
 	// LOGINFO("### delete cChunk() (%i, %i) from %p, thread 0x%x ###", m_PosX, m_PosZ, this, GetCurrentThreadId());
-	
+
 	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(); itr != m_BlockEntities.end(); ++itr)
 	{
 		delete *itr;
@@ -143,7 +143,7 @@ cChunk::~cChunk()
 			delete *itr;
 		}
 	}
-	
+
 	if (m_NeighborXM != nullptr)
 	{
 		m_NeighborXM->m_NeighborXP = nullptr;
@@ -284,12 +284,12 @@ void cChunk::GetAllData(cChunkDataCallback & a_Callback)
 	a_Callback.LightIsValid(m_IsLightValid);
 
 	a_Callback.ChunkData(m_ChunkData);
-	
+
 	for (cEntityList::iterator itr = m_Entities.begin(); itr != m_Entities.end(); ++itr)
 	{
 		a_Callback.Entity(*itr);
 	}
-	
+
 	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(); itr != m_BlockEntities.end(); ++itr)
 	{
 		a_Callback.BlockEntity(*itr);
@@ -305,7 +305,7 @@ void cChunk::SetAllData(cSetChunkData & a_SetChunkData)
 	ASSERT(a_SetChunkData.IsHeightMapValid());
 	ASSERT(a_SetChunkData.AreBiomesValid());
 	ASSERT(IsQueued());
-	
+
 	memcpy(m_BiomeMap, a_SetChunkData.GetBiomes(), sizeof(m_BiomeMap));
 	memcpy(m_HeightMap, a_SetChunkData.GetHeightMap(), sizeof(m_HeightMap));
 
@@ -339,19 +339,19 @@ void cChunk::SetAllData(cSetChunkData & a_SetChunkData)
 			ASSERT(WorldBlockType == EntityBlockType);
 		}  // for itr - m_BlockEntities
 	#endif  // _DEBUG
-	
+
 	// Set all block entities' World variable:
 	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(); itr != m_BlockEntities.end(); ++itr)
 	{
 		(*itr)->SetWorld(m_World);
 	}
-	
+
 	// Create block entities that the loader didn't load; fill them with defaults
 	CreateBlockEntities();
-	
+
 	// Set the chunk data as valid. This may be needed for some simulators that perform actions upon block adding (Vaporize)
 	SetPresence(cpPresent);
-	
+
 	// Wake up all simulators for their respective blocks:
 	WakeUpSimulators();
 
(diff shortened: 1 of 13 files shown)


src/Mobs/Spider.cpp
```diff
@@ -43,9 +43,15 @@ void cSpider::EventSeePlayer(cEntity * a_Entity)
 		return;
```
LogicParrot: Does it hurt performance the way it is? Lighting for the sake of spiders is not urgent. If I bailout without calling `GetWorld()->QueueLightChunk(GetChunkX(), GetChunkZ());`, would it be eventually be lighted?

src/Chunk.cpp
```diff
@@ -2600,6 +2600,31 @@ void cChunk::GetBlockInfo(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE & a_Bloc
...
+bool cChunk::GetChunkAndRelByAbsolute(const Vector3d & a_Position, cChunk ** a_Chunk, int & a_RelX, int & a_RelY, int & a_RelZ)
+{
+	return GetChunkAndRelByAbsolute(Vector3i(FloorC(a_Position.x), FloorC(a_Position.y), FloorC(a_Position.z)), a_Chunk, a_RelX, a_RelY, a_RelZ);
+}
+
+
+
+
+
+bool cChunk::GetChunkAndRelByAbsolute(const Vector3i & a_Position, cChunk ** a_Chunk, int & a_RelX, int & a_RelY, int & a_RelZ)
```
worktycho: Is there a reason that you are inconsistently using a vector for position, and components for relative position?

src/Mobs/Spider.cpp
worktycho: QueueLightChunk should be a lightweight operation, as it only adds the chunk to the queue.

src/Chunk.cpp
LogicParrot: I used components for relative position to stay consistent with the rest of the class. I used a vector for the new addition because it makes sense. Is there any reason the rest of the class is using components for relative position?

worktycho: Legacy code is the main reason.

LogicParrot: >I used a vector for the new addition because it makes sense

To further elaborate: entities have a vector already, no sense in bisecting it whenever we want to query cChunk.

LogicParrot: >I used a vector for the new addition because it makes sense

To further elaborate: entities have a vector already, no sense in bisecting it every time we want to query cChunk.

LogicParrot: @worktycho Okay, I will have a Rel vector instead of this mess. This also eliminates the need for the PREPARE_RELS macro.

src/Mobs/Spider.cpp
LogicParrot: Does it make sure it's not already there?

src/Chunk.cpp
LogicParrot: Done. I replaced `PREPARE_RELS` with a different macro. I think this is as convenient as it can get.

LogicParrot: Done. I also replaced `PREPARE_RELS` with a different macro. I think this is as convenient as it can get.

src/Mobs/Spider.cpp
worktycho: It checks when it reaches the head of the queue, if the chunk needs lighting, so there is a risk a chunk will be kept loaded for longer then necessary.

LogicParrot: I meant to ask if calling `QueueLightChunk` twice would add duplicates in the queue.

worktycho: It will add duplicates, but that is handled at the head of the queue.

LogicParrot: Oh.
Should it be worrying that all spiders adding dupes every single tick?

worktycho: They only do it if its not lit, but how many spiders will there be in the chunk?

LogicParrot: But what if the server is under heavy load and it takes, say, 3 seconds until the chunk is processed and lighted? During that time, each spider in the chunk will add the chunk to the queue 60 times. Is my understanding of the threads correct? I'm assuming lots of ticks can occur before a chunk is lighted and removed from the queue.

> how many spiders will there be in the chunk?
It depends. I'd say 0 - 5 on average.

worktycho: Yes that might actually be a problem.

LogicParrot: The simplest solution is bailing out without pushing anything to the queue. But is there a guarantee that the chunk will eventually light up at some point?

LogicParrot: I tried a bailout without a queue. Nothing changed in-game. I guess I'll go for that. In the very worst case scenario, a spider will not attack a player in some edge case even though it's dark. This is not worth worrying about.

LogicParrot:
Rebased, lgtm.

PR #2897, opened 2016-01-19
LuaTCPLink: Report data received via SSL just before connection closed.

madmaxoft:
DO NOT MERGE YET, this is for shared testing.

@LogicParrot edit:
Binaries for testing
 - [Windows x64](https://builds.cuberite.org/job/Cuberite%20Windows%20x64%20pr-testing/1/artifact/Install/Cuberite.zip)
 - [Gnu/Linux x64](https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20pr-testing/1/artifact/Cuberite.tar.gz)

### DIFF
--- a/src/Bindings/LuaTCPLink.cpp
+++ b/src/Bindings/LuaTCPLink.cpp
@@ -416,9 +416,10 @@ void cLuaTCPLink::OnReceivedData(const char * a_Data, size_t a_Length)
 	}
 
 	// If we're running in SSL mode, put the data into the SSL decryptor:
-	if (m_SslContext != nullptr)
+	auto sslContext = m_SslContext;
+	if (sslContext != nullptr)
 	{
-		m_SslContext->StoreReceivedData(a_Data, a_Length);
+		sslContext->StoreReceivedData(a_Data, a_Length);
 		return;
 	}
 
@@ -440,6 +441,13 @@ void cLuaTCPLink::OnRemoteClosed(void)
 	if (!m_Callbacks.IsValid())
 	{
 		return;
+	}
+
+	// If running in SSL mode and there's data left in the SSL contect, report it:
+	auto sslContext = m_SslContext;
+	if (sslContext != nullptr)
+	{
+		sslContext->FlushBuffers();
 	}
 
 	// Call the callback:


madmaxoft:
@NiLSPACE does this fix your HTTPS download issues? And does it output the log message `Leftover data in SSL context, flushing.`? (debug-mode only)

NiLSPACE:
I've tried it, but it doesn't receive any data:
```
     [dec3d1ad2a06c22|15:46:03] Executing console command: "load WorldEdit"
     [dec3d1ad2a06c22|15:46:03] [WorldEdit] Enabling WorldEdit v0.1.1
     [dec3d1ad2a06c22|15:46:03] Plugin loaded
Info [9a36935c4af72867|15:46:03] Connecting
Info [9a36935c4af72867|15:47:04] Connection closed
Info [9a36935c4af72867|15:47:04] Connecting
Info [9a36935c4af72867|15:48:05] Connection closed
Info [9a36935c4af72867|15:48:05] Connecting
Info [9a36935c4af72867|15:49:06] Connection closed
Info [9a36935c4af72867|15:49:06] Connecting
Info [9a36935c4af72867|15:50:07] Connection closed
Warn [9a36935c4af72867|15:50:07] Could not connect to Github to check for a newer version for WorldEdit
```

I've noticed this before at my internship (where I've done this test). Only sometimes it actually gets through. This also happens allot back home, but it happens way more here. As you can see the connection is closed exactly one minute after it is started.

Here is the used connection code just to see where I got the LOGINFO's:
```lua
OnConnected = function(a_TCPLink)
		local res, msg = a_TCPLink:StartTLSClient("", "", "")
		LOGINFO("Connecting")
		a_TCPLink:Send("GET /cuberite/WorldEdit/master/Info.lua HTTP/1.1\r\nHost: raw.githubusercontent.com\r\nConnection: Close\r\n\r\n")
	end,
	
	OnError = function(a_TCPLink, a_ErrorCode, a_ErrorMsg) 
		LOGINFO("Error")
		LOGWARNING("Error while checking for newer WorldEdit version: " .. a_ErrorMsg)
	end,
	
	OnReceivedData = function(a_TCPLink, a_Data)
		LOGINFO("Receiving Data")
		g_PluginInfo = g_PluginInfo .. a_Data
	end,
	
	OnRemoteClosed = function(a_TCPLink)
		LOGINFO("Connection closed")
		cUpdater:ParseResults()
	end,
```

I'll reload the plugin a few times to see if I can get it to connect at least once.

madmaxoft:
Looks as if the problem was elsewhere - with an exact minute before the disconnect, I'd venture a guess that Cuberite isn't actually sending the HTTP request, so the Mojang server just times out the connection without ever replying anything.

Might be interesting to replicate this behavior against a HTTPS server for which we have a certificate and dissect the traffic with wireshark, to see if it really is the case.

NiLSPACE:
You could use the gallery's webadmin to do that.

NiLSPACE:
Interesting. I decided to speed things up by putting the request in a for-loop that is 15 long. I finally got it to receive something back, but apparently it decided to merge two requests in one.
[RawOutput.txt](https://github.com/cuberite/cuberite/files/96084/RawOutput.txt)

madmaxoft:
You're not resetting the `g_PluginInfo` variable, so it's possible it got data from multiple requests.

NiLSPACE:
But it's only writing to `RawOutput.txt` if the file doesn't exist yet, so only the first request that receives data should be written.

LogicParrot:
I added the relevant CI binaries so that others can test without compiling.

madmaxoft:
This is not really testable in-game.
Anyway, I believe this change is valid, even if it doesn't fix the immediate problem, it makes the code more resistant. I have removed the debug logging and rebased onto master, I suggest we merge this in.

NiLSPACE:
Yes, I agree.

PR #2898, opened 2016-01-20
fix url parsing

marvinkopf:
This fixes a bug where http requests with long urls would crash the server with a std::length_error exception. 

[url.txt](https://github.com/cuberite/cuberite/files/96980/url.txt)


madmaxoft:
Good catch, thanks for the report; however, your fix was not ideal, so I fixed it properly in #2899.

PR #2899, opened 2016-01-20
Fixed HTTP parsing when in insecure mode.

madmaxoft:
Parsing would ignore the size of data already buffered, resulting in bad_alloc exception.
Fixes #2898.

### DIFF
--- a/src/HTTPServer/HTTPMessage.cpp
+++ b/src/HTTPServer/HTTPMessage.cpp
@@ -94,11 +94,13 @@ size_t cHTTPRequest::ParseHeaders(const char * a_Data, size_t a_Size)
 	{
 		// The first line hasn't been processed yet
 		size_t res = ParseRequestLine(a_Data, a_Size);
+		ASSERT((res == AString::npos) || (res <= a_Size));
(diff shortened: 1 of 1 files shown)


src/HTTPServer/HTTPMessage.cpp
```diff
@@ -210,7 +214,7 @@ size_t cHTTPRequest::ParseRequestLine(const char * a_Data, size_t a_Size)
 				}
 				m_Method = m_IncomingHeaderData.substr(LineStart, MethodEnd - LineStart);
 				m_URL = m_IncomingHeaderData.substr(MethodEnd + 1, URLEnd - MethodEnd - 1);
-				return i + 1;
+				return i + 1 - inBufferSoFar;
```
madmaxoft: This is the main point of the change - `i` counts over the buffer, but we want to return the number of bytes consumed from the current call only, hence the subtraction of the old buffer size.

SphinxC0re:
LGTM

madmaxoft:
@worktycho @tigerw Have I missed anything?

PR #2901, opened 2016-01-21
Fix running as a service.

### LINKED ISSUE #2900: Service support broken on all platforms.
Running as a service has been broken since 4315a11
PR incoming with a fix.

birkett:
Two problems identified, both traced to commit 4315a11
Fixes #2900 

(main.cpp) ParseArguments() was no longer being called before checking cRoot::m_RunAsService.
This led to m_RunAsService never being set (we use -d on the command line to instigate a service start).

(Root.cpp) Terminate events were being ignored when running as a service.
We only need to ignore the cin.good() call, not terminate events as a whole.

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -92,7 +92,8 @@ void cRoot::InputThread(cRoot & a_Params)
 	}
 
 	// We have come here because the std::cin has received an EOF / a terminate signal has been sent, and the server is still running
-	if (!std::cin.good())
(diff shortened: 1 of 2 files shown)


worktycho:
CI failed.

birkett:
Sorry about that, first version broke the restart command, causing the build to hang at the end. Fixed now.

madmaxoft:
Looks good to me.

SphinxC0re:
Looks good on my side, too

PR #2903, opened 2016-01-21
Fix TabAutoComplete

### LINKED ISSUE #2747: Chat auto complete -spaced commands duplicating first command when tabbing.
Issue:

When you press tab to auto-complete a multi-spaced command you will get initial command duplicated.

To duplicate this issue:
Cuberite server using Essentials plugin

open chat do "/time "  <-with space
press tab a couple of times and you will start getting "/time /time /time /time"

marvinkopf:
For tab on `/command s` the server now returns `subcommand` instead of `/command subcommand`.

Fixes #2747

### DIFF
--- a/src/Bindings/PluginManager.cpp
+++ b/src/Bindings/PluginManager.cpp
@@ -1894,7 +1894,27 @@ void cPluginManager::TabCompleteCommand(const AString & a_Text, AStringVector & 
 			// Player doesn't have permission for the command
 			continue;
 		}
-		a_Results.push_back(itr->first);
+
+		/*  Client expects to only get back the last part of a space separated command.
+		Find the position of the beginning of the last part:
+		Position of last space + 1 for space separated commands
+		string::npos + 1 = 0 for commands that are not separated
+
+		Then skip all commands that have too many subcommands.
+		When the client asks for suggestions for "/time s"
+		the server must skip all commands that consist of more than 2 words just as
+		"/time set day". Or in other words, the position of the last space (separator)
+		in the strings must be equal or string::npos for both. */
+		size_t LastSpaceInText = a_Text.find_last_of(' ') + 1;
+		size_t LastSpaceInSuggestion = itr->first.find_last_of(' ') + 1;
+
+		if (LastSpaceInText != LastSpaceInSuggestion)
+		{
+			// Suggestion has more subcommands than a_Text
+			continue;
+		}
+
+		a_Results.push_back(itr->first.substr(LastSpaceInSuggestion));
 	}
 }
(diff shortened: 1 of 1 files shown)


SphinxC0re:
No objections so far. @cuberite/core-team

SphinxC0re:
Oh, btw, shouldn't this loop be refactorable to
```cpp
for(auto command : m_commands)
{
...
}
```
?

madmaxoft:
I'd prefer a solution that was clearer to read. Right now it's a character soup, very difficult to parse for humans.
Add comments as to what it's doing, add temporary variables for the values if it makes it easier to understand the code.

marvinkopf:
Should have been continue if they are not equal. 

The client expects suggestions for the position he is currently at. When the client asks for suggestions for "/time q" the server must skip all commands that consist of more than 2 words just as "/time query daytime". Or in other words, the position of the last space (separator) in the strings must be equal.

madmaxoft:
That's exactly the kind of documentation that's needed there. By all means add the example into the comment to make it absolutely clear (but use `/time set day` command instead :)
Make the comment same-indent-level on all lines.

src/Bindings/PluginManager.cpp
```diff
@@ -1894,7 +1894,27 @@ void cPluginManager::TabCompleteCommand(const AString & a_Text, AStringVector &
...
 			continue;
 		}
-		a_Results.push_back(itr->first);
+
+		/*  Client expects to only get back the last part of a space separated command.
+			Find the position of the beginning of the last part:
+			Position of last space + 1 for space separated commands
+			string::npos + 1 = 0 for commands that are not separated
+
+			Then skipp all commands that have too many subcommands.
```
LogicParrot: Typo: Should be "skip".

LogicParrot:
```
	/* Client expects to only get back the last part of a space separated command.
	Find the position of the beginning of the last part:
	Position of last space + 1 for space separated commands
	string::npos + 1 = 0 for commands that are not separated

	Then skip all commands that have too many subcommands.
	When the client asks for suggestions for "/time s"
	the server must skip all commands that consist of more than 2 words just as
	"/time set day". Or in other words, the position of the last space (separator)
	in the strings must be equal or string::npos for both. */
```

SphinxC0re:
I would like to see this loop refactored as part of our c++11 conversion

LogicParrot:
I'll include that in a "refactor-spree" PR seperately later, along with lots of similar loop changes in monster code.

LogicParrot:
@marvinkopf thanks :)

PR #2908, opened 2016-01-25
Fix cPawn push logic

### LINKED ISSUE #2886: Minecarts are too fast, uncontrollable/go crazy
Right now, minecarts will speed up way too fast, even when there absolutely is no energy source used tp sufficiently propelling them forward. All sorts of redstone track parts are ignored by it, and it has strong tendency to fly off the tracks, doing health damage to the player riding it.

marvinkopf:
This adds a boolean IsAttachedTo method to cEntity and cPawn uses it to perform a check before a cPawn instance pushes another entity. fixes #2886

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1864,6 +1864,19 @@ bool cEntity::IsA(const char * a_ClassName) const
 
 
 
+bool cEntity::IsAttachedTo(const cEntity * a_Entity) const
+{
+	if ((m_AttachedTo != nullptr) && (a_Entity->GetUniqueID() == m_AttachedTo->GetUniqueID()))
+	{
+		return true;
+	}
+	return false;
+}
+
+
+
+
+
 void cEntity::SetHeadYaw(double a_HeadYaw)
 {
 	m_HeadYaw = a_HeadYaw;
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -73,6 +73,12 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 				return false;
 			}
 
+			// do not push a boat / minecart you're sitting in
+			if (m_Pusher->IsAttachedTo(a_Entity))
+			{
+				return false;
+			}
+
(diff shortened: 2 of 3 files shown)


SphinxC0re:
Does this apply to all entities or just to Minecarts? If it only applies to Minecarts, you should consider moving the code into the Minecart entity handler

marvinkopf:
This applies to boats & minecarts. 

But it's a bug in the Item method in cPawn and got nothing to do with the boat or minecarts class. I changed the title to make a bit more clear.

SphinxC0re:
Yeah since not every entity needs it, that would be good

madmaxoft:
No, keep the attachment methods together in `cEntity`. I like this change, but it still needs a few minor style tweaks.

src/Entities/Entity.cpp
```diff
@@ -1864,6 +1864,20 @@ bool cEntity::IsA(const char * a_ClassName) const
 
 
 
+bool cEntity::IsAttachedTo(const cEntity *a_Entity) const
```
madmaxoft: Space after `*` as well.

src/Entities/Entity.cpp
```diff
@@ -1864,6 +1864,20 @@ bool cEntity::IsA(const char * a_ClassName) const
...
+{
+	if ((m_Attachee != NULL) &&
```
madmaxoft: Move the condition to a separate line; don't use `NULL`, use `nullptr` instead.

src/Entities/Pawn.cpp
```diff
@@ -73,6 +73,11 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 				return false;
 			}
 
+			if (a_Entity->IsAttachedTo(m_Pusher))
```
madmaxoft: Add a comment about what's going on.

marvinkopf:
Moved the if condition into one single line, wasn't really that long after all.

LogicParrot:
LGTM. @madmaxoft , any more comments?

SphinxC0re:
Rebase and squash the commits onto the current master

PR #2939, opened 2016-02-01
cMonster::m_target safety across worlds

### LINKED ISSUE #2896: Potential use-after-free in mob code
If an entity is killed while a monster is targeting it, it will be freed in:
https://github.com/cuberite/cuberite/blob/85e534b745c96985d9132fc6f05b2e0fa67048fd/src/Chunk.cpp#L629

And the targeting monster may execute this after the free:
https://github.com/cuberite/cuberite/blob/85e534b745c96985d9132fc6f05b2e0fa67048fd/src/Mobs/Monster.cpp#L239

Is my reasoning correct?
Slightly related forum thread: [safety of cMonster::m_Target](http://forum.mc-server.org/showthread.php?tid=2324&pid=24663#pid24663)

LogicParrot:
Fixes #2896.
I used @madmaxoft 's "solution 3".

Summary of what this is all about:
 - Every cPawn(player or mob) keeps a list of monsters targeting it.
 - cMonster is responsible for notifying cPawn of target changes to keep the list up to date. See `cMonster::setTarget`
 - When a cPawn is destroyed, it tells all monsters to stop targeting it.
 - When a cPawn travels to a new world, it tells all monsters to stop targeting it.
 - When a monster is destroyed or moves to a new world, the previous two points are  done (because all monsters are cPawn), but additionally, the monster sets the target to null, thus notifying its target that it is no longer targeting it and keeping the list up to date.
 - The cWorld / Chunk `doWith...` functions must never do stuff with Pawns that are destroyed.
 - Added lots of asserts.

### DIFF
--- a/src/Entities/Pawn.h
+++ b/src/Entities/Pawn.h
@@ -4,6 +4,9 @@
 #include "Entity.h"
 #include "EntityEffect.h"
 
+// fwd cMonster
+class cMonster;
+
 
 
 
@@ -14,21 +17,28 @@ class cPawn :
 {
 	// tolua_end
 	typedef cEntity super;
-	
+
 public:
 	CLASS_PROTODEF(cPawn)
 
 	cPawn(eEntityType a_EntityType, double a_Width, double a_Height);
-	
+	~cPawn();
+	virtual void Destroyed() override;
+
 	virtual void Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
 	virtual void KilledBy(TakeDamageInfo & a_TDI) override;
-	
+
 	virtual bool IsFireproof(void) const override;
 	virtual void HandleAir(void) override;
 	virtual void HandleFalling(void);
 
+	/** Tells all pawns which are targeting us to stop targeting us. */
+	void StopEveryoneFromTargetingMe();
+
+
+
 	// tolua_begin
-	
+
 	/** Applies an entity effect
 	Checks with plugins if they allow the addition.
 	@param a_EffectType          The entity effect to apply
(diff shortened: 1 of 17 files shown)


src/Entities/Pawn.h
```diff
@@ -37,28 +47,39 @@ class cPawn :
...
 	bool HasEntityEffect(cEntityEffect::eType a_EffectType) const;
-	
+
 	/** Removes all currently applied entity effects (used when drinking milk) */
 	void ClearEntityEffects(void);
 
 	// tolua_end
 
+	/** Add the monster to the list of monsters targeting this pawn. (Does not check if already added!) */
+	void NoLongerTargetingMe(cMonster * a_Monster);
```
LogicParrot: The comments are flipped.

LogicParrot:
Summary of what this is all about:
 - Every cPawn(player or mob) keeps a list of monsters targeting it.
 - cMonster is responsible for notifying cPawn of target changes to keep the list up to date. See `cMonster::setTarget`
 - When a cPawn is destroyed, it tells all monsters to stop targeting it.
 - When a monster is destroyed, the above is done (because all monsters are cPawn), but additionally, the monster sets the target to null, thus notifying its target that it is no longer targeting it.

LogicParrot:
Summary of what this is all about:
 - Every cPawn(player or mob) keeps a list of monsters targeting it.
 - cMonster is responsible for notifying cPawn of target changes to keep the list up to date. See `cMonster::setTarget`
 - When a cPawn is destroyed, it tells all monsters to stop targeting it.
 - When a cPawn travels to a new world, it tells all monsters to stop targeting it.
 - When a monster is destroyed or moves to a new world, the previous two points are  done (because all monsters are cPawn), but additionally, the monster sets the target to null, thus notifying its target that it is no longer targeting it and keeping the list up to date.
 - The cWorld / Chunk `doWith...` functions must never do stuff with Pawns that are destroyed.
 - Added lots of asserts.

LogicParrot:
Todo: Did I miss any `forEach / doWith`?

LogicParrot:
This should be merged into testing.

SphinxC0re:
Then you should merge this directly into testing. Also testing should be propagated to release soon @worktycho

LogicParrot:
Okay. Let's do that once this (and #2940) are reviewed and rebased.

How would you like to do this? I think a good idea is merging into master and then cherry-picking.

SphinxC0re:
I would like it more, if we'd merge this into testing and then merge testing into master

LogicParrot:
I think it would be more natural for everyone to preserve their normal workflow and merge to master as usual. And we'd cherry-pick into `testing` whenever needed.

tigerw:
I can't help but feel that it may be a better idea to change entity ownership to smart pointers and have Target be a weak pointer, or entity ID, in the near future. I suspect doing manual invalid pointer cleanup on entity destruction isn't particularly robust.

Of course though, it's a hotfix, so anything goes :P

LogicParrot:
@tigerw I figured we'd need some manual management anyways; even with a smart pointer, we'd have to broadcast to other entities to let us go when world travelling to avoid threading issues. Don't we? So I've gone manual all the way :D

I've carefully made sure nothing is leaked, but I'm very open to safer proposals. But I'd say merge now, discuss and improve later, because the current master world travel issues are severe.

tigerw:
The referenced shared pointer would be deleted and a new one constructed based on a copy/move of the data contained in the old, therefore invalidating the referencing weak pointers in the previous world.

But yeah, it looks quite bad, but merge I guess.

LogicParrot:
Alright. could you review my conflict fix before we merge? (Not much changed)

LogicParrot:
Hold on, I think there's another potential threading issue in `DoMoveWithWorld`: Joining the new thread should be the very last thing that method does, right?

LogicParrot:
Or more precisely, the last field-modifying thing.

LogicParrot:
I solve that by putting setTarget(nullptr) before the world change code.

PR #2940, opened 2016-02-02
Fixed cChunk::m_Entities corruption upon world travel

LogicParrot:
Fixes #2724
possibly fixes or completely resolves other [world travel issues](https://github.com/cuberite/cuberite/milestones/Stable%20world%20travel) as well.

Issue #2839, opened 2015-12-31
Segfault cTCPLinkImpl::EventCallback


Seadragon91:
I had yesterday a few segfaults at this position [TCPLinkImpl.cpp:L265](https://github.com/cuberite/cuberite/blob/master/src/OSSupport/TCPLinkImpl.cpp#L265). The windows crashdump that I have, shows that Self was empty.

Tried to reproduce it in a debug build, had no luck so far.

Seadragon91:
Started the server and wanted to join it, the server crashed. Have attached the crashdump. Build id [220](https://builds.cuberite.org/job/Cuberite%20Windows%20x64%20Master/220/)

[Cuberite.exe.3192.dmp.zip](https://github.com/cuberite/cuberite/files/90314/Cuberite.exe.3192.dmp.zip)

Seadragon91:
I get this crash now 1 of 4 times on login to the server.

Seadragon91:
Would be enough to check if [`Self`](https://github.com/cuberite/cuberite/blob/8fc9cfa41101b6b99cbf1dda21be1a22980aab0f/src/OSSupport/TCPLinkImpl.cpp#L216) is nullptr and then only return?
The value of `a_What` is 0x11.

madmaxoft:
Not really, this is a much deeper issue, because `Self` is supposed to be a `shared_ptr` and there's already a check for an empty one, so this really means we've hit undefined behavior somewhere along the way, most likely due to a data race.

PR #2941, opened 2016-02-02
[Attempt 2] Fixed cChunk::m_Entities corruption upon world travel

### LINKED ISSUE #2724: Crash on multiworld travel
I was in the Nether. I executed `/portal world`

```
[70153d24f3fd4a92|15:28:10] Assertion failed: std::find(m_Entities.begin(), m_Entities.end(), a_Entity) == m_Entities.end(), file /home/cuberite/cuberite/src/Chunk.cpp, line 1944
./Cuberite_debug(_Z15PrintStackTracev+0x23)[0xb0c901]
./Cuberite_debug(_ZN6cChunk9AddEntityEP7cEntity+0xce)[0xa0e5f4]
./Cuberite_debug(_ZN6cChunk20MoveEntityToNewChunkEP7cEntity+0xfe)[0xa0af96]
./Cuberite_debug(_ZN6cChunk4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEE+0x3e6)[0xa0ac24]
./Cuberite_debug(_ZN9cChunkMap11cChunkLayer4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEE+0x86)[0xa23ecc]
./Cuberite_debug(_ZN9cChunkMap4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEE+0x58)[0xa2347a]
./Cuberite_debug(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x3f6)[0xad8a12]
./Cuberite_debug(_ZN6cWorld11cTickThread7ExecuteEv+0xad)[0xacf425]
./Cuberite_debug(_ZN9cIsThread9DoExecuteEv+0x2f)[0xb01d8d]
./Cuberite_debug(_ZNKSt7_Mem_fnIM9cIsThreadFvvEEclIIEvEEvPS0_DpOT_+0x65)[0xb0338d]
./Cuberite_debug(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EE9_M_invokeIILm0EEEEvSt12_Index_tupleIIXspT_EEE+0x43)[0xb032dd]
./Cuberite_debug(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EEclEv+0x1d)[0xb031d3]
./Cuberite_debug(_ZNSt6thread5_ImplISt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS3_EEE6_M_runEv+0x1c)[0xb03150]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xb6970)[0x7f3acdeb1970]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x80a4)[0x7f3ace10e0a4]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d)[0x7f3acd62104d]
Cuberite_debug: /home/cuberite/cuberite/src/Chunk.cpp:1944: void cChunk::AddEntity(cEntity*): Assertion `0' failed.

```

LogicParrot:
**Attempt 2, previous attempt: #2940**
Fixes #2724
possibly fixes or completely resolves other world travel issues as well.

### DIFF
--- a/src/Chunk.h
+++ b/src/Chunk.h
@@ -260,6 +260,9 @@ public:
 
 	void AddEntity(cEntity * a_Entity);
 	void RemoveEntity(cEntity * a_Entity);
+	/** RemoveEntity is dangerous if the chunk is inside the tick() method because it invalidates the iterator.
+	This will safely remove an entity. */
+	void SafeRemoveEntity(cEntity * a_Entity);
 	bool HasEntity(UInt32 a_EntityID);
 
 	/** Calls the callback for each entity; returns true if all entities processed, false if the callback aborted by returning true */
@@ -502,7 +505,7 @@ private:
 
 	/** If the chunk fails to load, should it be queued in the generator or reset back to invalid? */
 	bool m_ShouldGenerateIfLoadFailed;
-
+	bool m_IsInTick;       // True if the chunk is executing the tick() method.
 	bool m_IsLightValid;   // True if the blocklight and skylight are calculated
 	bool m_IsDirty;        // True if the chunk has changed since it was last saved
 	bool m_IsSaving;       // True if the chunk is being saved
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -603,6 +603,7 @@ void cChunk::SpawnMobs(cMobSpawner & a_MobSpawner)
 
 void cChunk::Tick(std::chrono::milliseconds a_Dt)
 {
+	m_IsInTick = true;
 	BroadcastPendingBlockChanges();
 
 	CheckBlocks();
@@ -637,7 +638,7 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 		else if ((*itr)->IsWorldTravellingFrom(m_World))
 		{
 			// Remove all entities that are travelling to another world
-			LOGD("Removing entity from [%d, %d] that's travelling between worlds.", m_PosX, m_PosZ);
+			LOGD("Removing entity from [%d, %d] that's travelling between worlds. (Scheduled)", m_PosX, m_PosZ);
 			MarkDirty();
 			(*itr)->SetWorldTravellingFrom(nullptr);
 			itr = m_Entities.erase(itr);
@@ -659,6 +660,7 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 	}  // for itr - m_Entitites[]
 
 	ApplyWeatherToTop();
+	m_IsInTick = false;
 }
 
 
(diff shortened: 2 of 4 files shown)


src/Chunk.h
```diff
@@ -260,6 +260,9 @@ class cChunk :
 
 	void AddEntity(cEntity * a_Entity);
 	void RemoveEntity(cEntity * a_Entity);
+	/** RemoveEntity is dangerous if the chunk is inside the tick() method because it invalidates the iterator.
+	This will safely remove an entity. Returns true on success. */
```
LogicParrot: returns nothing.

tigerw:
How does this work around the stated diagnosis of cChunk not removing anything at all because it doesn't tick? Iterators shouldn't be invalidated since the Entities container is I believe a list, not vector.

LogicParrot:
@tigerw Because of Diagnoses Continued :tm: 

There are two code paths that lead to teleport (cEntity::doMoveWithWorld).

Code path 1:
 - cChunk::tick() ticks all the entities [here](https://github.com/cuberite/cuberite/pull/2941/files#diff-4b84dbafe20b4c913f541b6369929de5R627), and one of the entities decides to teleport inside its own tick method. In that case, the entity marks itself for removal, and is properly removed inside [this removal if block](https://github.com/cuberite/cuberite/pull/2941/files#diff-4b84dbafe20b4c913f541b6369929de5R638) right after it ticks. This was done properly even before my changes.

Code path 2:
- Some force outside the cChunk::tick, like a plugin, asks for a teleport. In that case:
 A: The chunk may not be ticking at all, and the "removal if block" may never get executed.
 B: Direct removal is fine precisely because the chunk is not in the midst of a tick and there is no risk of iterator corruption.

My changes essentially say:
"If it's code path 2, remove directly. If it's code path 1, schedule for removal as we always have."

tigerw:
I guess that's fine. Probably something in the future along the lines of queueing a task to move the entity would look better and be easier to follow, even if it introduces a one-tick delay.

LogicParrot:
@tigerw It's true that it's a list, but this code causes trouble nevertheless:
` cChunk::tick() -> (*it)->tick() -> cEntity::tick > Tell cChunk to delete (it)`

Is there an easy way to do this? Maybe that would be more readable.
` cChunk::tick() -> (*it)->tick() -> cEntity::tick > Tell cChunk to hop to next iterator and then remove (it)`

LogicParrot:
@Seadragon91 would you mind testing this one? Preferably both netherportal travel and `/portal` travel.

Seadragon91:
This works fine now. Had 5 times the crash described in issue #2839.

tigerw:
So I guess merge, it will hopefully be all refactored away soon whatever the case.

PR #2953, opened 2016-02-03
[PhantomWorlds - 3] Regenerate worlds with proper dimension types and linkages

LogicParrot:
Fixes part 3 of #2810
When a nether world folder is removed, an overworld is created instead. This is unexpected. I fixed it. Also, previously newly created overworlds always linked to two worlds, even if these worlds do not exist. I fixed that too. Here is exactly what happens now:

- If a world exists in settings.ini but has no world.ini, then:
  - If the world name is x_nether, create a world.ini with the a dimension type "nether"
    - If a world called x exists, set it as x_nether's overworld
    - Otherwise set the default world as x_nether's overworld

  - If the world name is x_end, create a world.ini with the a dimension type "end"
    - If a world called x exists, set it as x_end's overworld
    - Otherwise set the default world as x_end's overworld

 - If the world name is x (and doesn't end with _end or _nether)
    - Create a world.ini with a dimension type of "overworld"
    - If a world called x_nether exists, set it as x's nether world, otherwise set x's nether world to blank.
    - If a world called x_end  exists, set it as x's end world, otherwise set x's nether world to blank.

**Note:** This is all about best-guessing how a newly created INI file should look like. 
 - It will never touch an already existing world.ini file at all. 
 - It will never put linkages to a nonexisting world in the newly created `world.ini`.
(Both points are deducable from the rules above, but I thought I should mention them explicitly)

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -409,6 +409,28 @@ void cRoot::LoadWorlds(cSettingsRepositoryInterface & a_Settings, bool a_IsNewIn
 		return;
 	}
 
+	/* Here are the world creation rules. Note that these only apply for a world which is in settings.ini but has no world.ini file.
+	If an ini file is present, it overrides the world linkages and the dimension type in cWorld::start()
+	The creation rules are as follows:
+
+	- If a world exists in settings.ini but has no world.ini, then:
+		- If the world name is x_nether, create a world.ini with the dimension type "nether".
+			- If a world called x exists, set it as x_nether's overworld.
+			- Otherwise set the default world as x_nether's overworld.
+
+		- If the world name is x_end, create a world.ini with the dimension type "end".
+			- If a world called x exists, set it as x_end's overworld.
+			- Otherwise set the default world as x_end's overworld.
+
+		- If the world name is x (and doesn't end with _end or _nether)
+			- Create a world.ini with a dimension type of "overworld".
+			- If a world called x_nether exists, set it as x's nether world.
+			- Otherwise set x's nether world to blank.h
+			- If a world called x_end  exists, set it as x's end world.
+			- Otherwise set x's nether world to blank.
+
+	*/
+
 	bool FoundAdditionalWorlds = false;
 	for (auto WorldNameValue : Worlds)
 	{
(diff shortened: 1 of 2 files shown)


LogicParrot:
Originally reported by @Seadragon91

src/Root.cpp
```diff
@@ -423,7 +423,20 @@ void cRoot::LoadWorlds(cSettingsRepositoryInterface & a_Settings, bool a_IsNewIn
 			continue;
 		}
 		FoundAdditionalWorlds = true;
-		cWorld * NewWorld = new cWorld(WorldName.c_str());
+		cWorld * NewWorld;
+		// Note that if there's a world.ini file present, the world will override the following worldTypes in cWorld::start
+		if (WorldName.find("_nether") != std::string::npos)  // If the world has the word nether, mark it as a nether world. (will be overridden by setting  if ini present)
```
LogicParrot: Do we have a project-specific convention for string comparison? Also, can I easily make this case-insensitive?

SphinxC0re: I think we could implement a method like this:
```cpp
bool AString::contains(char*);
```

tigerw: AString is an xoft-ism (i.e. completely unnecessary type definition) for the standard string type.

tigerw: Ha! Project convention! Very funny comrade.

You could do a C and (nocasecompare == 0), I suppose.

LogicParrot:
I realized I can catch a slightly bigger fish here. Do not merge yet.

LogicParrot:
Done. Rebased.

LogicParrot:
I propose that this makes it into testing. @worktycho ?
Also, did the auth problems make it into testing? If not, someone should cherry-pick those before release.

LogicParrot:
Updated initial post significantly.

src/Root.cpp
```diff
@@ -423,7 +445,44 @@ void cRoot::LoadWorlds(cSettingsRepositoryInterface & a_Settings, bool a_IsNewIn
 			continue;
 		}
 		FoundAdditionalWorlds = true;
-		cWorld * NewWorld = new cWorld(WorldName.c_str());
+		cWorld * NewWorld;
+		AString LowercaseName = WorldName;
+		std::transform(LowercaseName.begin(), LowercaseName.end(), LowercaseName.begin(), ::tolower);
```
madmaxoft: We've got `StrToLower()`

LogicParrot: Done.

PR #2961, opened 2016-02-04
[PhantomWorlds - 4] Prevent server from starting if there's bad world linkage

LogicParrot:
Fixes part 4 of #2810

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -327,6 +327,17 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 
 
 
+void cRoot::StopServer()
+{
+	m_TerminateEventRaised = true;
+	m_StopEvent.Set();
+	m_InputThreadRunFlag.clear();
+}
+
+
+
+
+
 void cRoot::LoadGlobalSettings()
 {
 	// Nothing needed yet
@@ -597,9 +608,7 @@ void cRoot::ExecuteConsoleCommand(const AString & a_Cmd, cCommandOutputCallback 
 	// Some commands are built-in:
 	if (a_Cmd == "stop")
 	{
-		m_TerminateEventRaised = true;
-		m_StopEvent.Set();
-		m_InputThreadRunFlag.clear();
+		StopServer();
 		return;
 	}
 	else if (a_Cmd == "restart")
--- a/src/Root.h
+++ b/src/Root.h
@@ -57,6 +57,9 @@ public:
 
 	void Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo);
 
+	/** Stops the server, as if "/stop" was typed in the console. */
+	void StopServer();
+
 	// tolua_begin
 	cServer * GetServer(void) { return m_Server; }
 	cWorld *  GetDefaultWorld(void);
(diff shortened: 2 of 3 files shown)


src/Root.h
```diff
@@ -57,6 +57,9 @@ class cRoot
 
 	void Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo);
 
+	/** Stops the server, as is /stop was typed in the console. */
```
LogicParrot: as if

src/Root.cpp
```diff
@@ -327,6 +327,13 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
...
 
 
+void cRoot::StopServer()
+{
+	m_TerminateEventRaised = true;
+	m_StopEvent.Set();
+	m_InputThreadRunFlag.clear();
+}
+
 void cRoot::LoadGlobalSettings()
```
madmaxoft: 5 empty lines between functions

src/World.cpp
```diff
@@ -520,6 +520,40 @@ void cWorld::Start(void)
...
+		}
+	}
+	else
+	{
+		if ((m_LinkedOverworldName != "") && (Root->GetWorld(m_LinkedOverworldName) == nullptr))
+		{
+			BadWorld = m_LinkedOverworldName;
+		}
+	}
+	if (BadWorld != "")
```
madmaxoft: Use `.empty()` instead

LogicParrot:
I'll do it better later.

src/World.cpp
```diff
@@ -520,6 +520,40 @@ void cWorld::Start(void)
 		m_LinkedOverworldName = IniFile.GetValueSet("LinkedWorlds", "OverworldName", GetLinkedOverworldName());
 	}
 
+	// If we are linked to one or more worlds that do not exist, ask the server to stop.
+	AString BadWorld = "";
```
madmaxoft: How about `BadWorlds`, since it may contain multiple worlds?

src/World.cpp
```diff
@@ -520,6 +520,40 @@ void cWorld::Start(void)
...
+	{
+		if ((m_LinkedOverworldName != "") && (Root->GetWorld(m_LinkedOverworldName) == nullptr))
+		{
+			BadWorld = m_LinkedOverworldName;
+		}
+	}
+	if (BadWorld != "")
+	{
+		const char * WorldName = m_WorldName.c_str();
+		LOGD("\n######\nERROR: %s is linked to one or more invalid worlds: %s\nPlease edit %s/world.ini and fix this.\n######\n",
```
madmaxoft: `LOGD` is debug-mode-only, use `LOGERROR` here (since you're stopping the server, this is one of the rare cases when this is appropriate)

LogicParrot: Okay. Can I somehow make it show up last in the console?

src/World.cpp
LogicParrot: Alright.

LogicParrot:
@madmaxoft fixes applied.

LogicParrot:
The server differentiates between `world` and `world ` and ` world`. I don't think this is right. Shouldn't we trim outer spaces when reading the ini?

src/World.cpp
madmaxoft: I don't think it's possible, and also not exactly necessary, the error message is colored so it stands out in the console. Having the log sequential makes things easier to understand, too, when reading the log, you know exactly what went wrong where.

LogicParrot: It wasn't colored back when I used `LOGD` so didn't know this. Yes, the coloring is enough.

LogicParrot:
Note: Do not merge before merging #2953, otherwise the server may refuse to start if players remove a world and then restart the server in order to regenerate it.

PR #2962, opened 2016-02-04
[PhantomWorlds - 5] Remove CreateAndInitializeWorld

LogicParrot:
Fixes part 5 of #2810.


NiLSPACE:
The documentation can be removed as well.

LogicParrot:
@NiLSPACE Done.

NiLSPACE:
Great :). No objections here anymore.

LogicParrot:
Note to everyone, do NOT merge this before #2961. If a player teleports to a non existing world, this will assert. #2961 prevents the server from starting if there are invalid worlds in the INI.

LogicParrot:
I've modified `cRoot::GetWorld`, which is an API function. Is this OK? @madmaxoft

LogicParrot:
Nevermind. The exported version ignores the second parameter, which I've removed. lgtm.

PR #2968, opened 2016-02-05
C++ 14 Deprecated Tag

bearbin:
Fixes error in https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/384/console

### DIFF
--- a/src/Entities/Player.h
+++ b/src/Entities/Player.h
@@ -218,6 +218,7 @@ public:
 
 	/** Forces the player to move in the given direction.
 	@deprecated Use SetSpeed instead. */
+	[[deprecated("Use SetSpeed instead.")]]
 	void ForceSetSpeed(const Vector3d & a_Speed);  // tolua_export
 
 	cWindow * GetWindow(void) { return m_CurrentWindow; }  // tolua_export


bearbin:
Should this be merged or is there something broken with my ccache implementation?

LogicParrot:
Appveyor failed. I presume this has nothing to do with ccache.

src/Entities/Player.h
```diff
@@ -218,6 +218,7 @@ class cPlayer :
 
 	/** Forces the player to move in the given direction.
 	@deprecated Use SetSpeed instead. */
+	[[deprecated("Use SetSpeed instead.")]]
```
LogicParrot: Is this valid c++? o.O

bearbin: http://josephmansfield.uk/articles/marking-deprecated-c++14.html

LogicParrot: But I thought we're C++11.

bearbin: That's what I thought. It's just that I got an error because of the lack of this, and I thought maybe we changed.

SphinxC0re:
We should really consider to use this feature. After all c++14 is two years old :-)

LogicParrot:
Stable > Cutting edge.

SphinxC0re:
c++17 is cutting edge

LogicParrot:
That's just terminology. It's bleeding-edge by my standards :P

The fact is, for portability and stability's sake, we shouldn't rush unless there's a serious incentive.

bearbin:
Thinking over this again, I don't think that this should be merged, instead we should find out what's going wrong with ccache such that it is using C++14 warnings.

LogicParrot:
Note that ccache is supposed to just pass stuff to the compiler, it does not generate warnings internally.

tigerw:
By the way, I'm all for C+14. I think it might just be a flag change.

PR #2969, opened 2016-02-05
Ignore trailing and leading spaces in INI values

LogicParrot:
Becase it's not nice for the server to generate a new world just because you typed "hello " and not "hello".

### DIFF
--- a/src/IniFile.cpp
+++ b/src/IniFile.cpp
@@ -146,7 +146,7 @@ bool cIniFile::ReadFile(const AString & a_FileName, bool a_AllowExampleRedirect)
 			case '=':
 			{
 				valuename = line.substr(0, pLeft);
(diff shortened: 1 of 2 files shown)


src/IniFile.cpp
```diff
@@ -893,6 +893,26 @@ void cIniFile::RemoveBom(AString & a_line) const
 
 
 
+AString cIniFile::TrimSpaces(const AString & a_String)
```
NiLSPACE: You can just use the [TrimString](https://github.com/cuberite/cuberite/blob/master/src/StringUtils.cpp#L219) function.

LogicParrot: Thanks.

LogicParrot:
Note that with this change, the server would clear bad spaces when it stops (it'll save to the INI file without the spaces.)

PR #2980, opened 2016-02-06
Proper entity destruction in non ticking chunks

### LINKED ISSUE #2731: `/butcher` command reports too many mobs
The butcher command seems to be lying about the number of mobs killed. This is either some internal issue or a plugin issue.

### LINKED ISSUE #2973: Investigate entity removal in unloaded chunks
Removing entities in unloaded chunks was the prime cause of world instability. It delayed the removal until the chunk loads, often causing m_Entities corruption in the process. #2941 resolves this by adding a function called  [cChunk::SafeRemoveEntity](https://github.com/cuberite/cuberite/commit/07b7fd4ad3fef714399009efd5d4a903a561a6f2#diff-af2ddf3aaa057bf2bbd7dbfe509c9918R265). We should investigate if more code paths lead to entity destruction in unloaded chunks, and we should make those code paths use some variant / modification of [cChunk::SafeRemoveEntity](https://github.com/cuberite/cuberite/commit/07b7fd4ad3fef714399009efd5d4a903a561a6f2#diff-af2ddf3aaa057bf2bbd7dbfe509c9918R265). #2731 is caused by such a code path.

LogicParrot:
Work in progress.
The destruction code is a mess, by the way.

Closes #2731
Closes #2973
May close or alleviate #277

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -129,7 +129,6 @@ cClientHandle::~cClientHandle()
 			World->RemovePlayer(m_Player, true);  // Must be called before cPlayer::Destroy() as otherwise cChunk tries to delete the player, and then we do it again
 			m_Player->Destroy();
 		}
-		delete m_Player;
 		m_Player = nullptr;
 	}
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -126,7 +126,7 @@ cClientHandle::~cClientHandle()
 				World->BroadcastPlayerListRemovePlayer(*m_Player, this);
 			}
 
-			World->RemovePlayer(m_Player, true);  // Must be called before cPlayer::Destroy() as otherwise cChunk tries to delete the player, and then we do it again
+			World->RemovePlayer(m_Player, true);
 			m_Player->Destroy();
 		}
 		m_Player = nullptr;
--- a/src/Entities/Entity.h
+++ b/src/Entities/Entity.h
@@ -481,6 +481,9 @@ public:
 	/** Sets the internal world pointer to a new cWorld, doesn't update anything else. */
 	void SetWorld(cWorld * a_World) { m_World = a_World; }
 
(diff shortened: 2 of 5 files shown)


LogicParrot:
Do not merge. There's a leak.

LogicParrot:
Todo:
- fix leak
- Proper function names to avoid confusion between move and delete, etc.

src/Entities/Entity.h
```diff
@@ -481,6 +481,9 @@ class cEntity
 	/** Sets the internal world pointer to a new cWorld, doesn't update anything else. */
 	void SetWorld(cWorld * a_World) { m_World = a_World; }
 
+	/** Sets the owning chunk. Should only be calld by cChunk::AddEntity */
```
Seadragon91: calld -> called

LogicParrot:
This turned out to be far tangled than anticipated. Closing for now.

src/ClientHandle.cpp
```diff
@@ -129,7 +129,6 @@ cClientHandle::~cClientHandle()
 			World->RemovePlayer(m_Player, true);  // Must be called before cPlayer::Destroy() as otherwise cChunk tries to delete the player, and then we do it again
 			m_Player->Destroy();
 		}
-		delete m_Player;
```
LogicParrot: Is there a reason the client handle should be the one deleting the player?

LogicParrot:
I believe this is stable now.

src/ClientHandle.cpp
```diff
@@ -129,7 +129,6 @@ cClientHandle::~cClientHandle()
 			World->RemovePlayer(m_Player, true);  // Must be called before cPlayer::Destroy() as otherwise cChunk tries to delete the player, and then we do it again
```
LogicParrot: What's wrong with cChunk deleting the player?

LogicParrot:
Todo: I think this code is problamatic if an entity calls the `destroy` of another entity which is before it in `m_Entities`.

PR #2981, opened 2016-02-06
Fix Pickups of BlockFlower

### LINKED ISSUE #2906: Every red_flower drops poppy
Every red_flower block (block_id: 38, e.g. Blue Orchid, Allium, Azure Bluet etc.) drops Poppy. => Data value is ignored. This makes difficult creating some dyes in survival.

Drahoslav7:
BlockMeta is now handled correctly, when converting to pickups, so flowers drops right flower type, not only poppy.

It closes #2906.

Also this is my first PR. I hope I'm doing this correctly.


bearbin:
@Drahoslav7 Good first PR!

Thanks for your contribution :)

PR #2985, opened 2016-02-07
Update RedstoneSimulator to delete unused cached PowerData

### LINKED ISSUE #2946: Minecarts don't accelerate properly on powered rails
I've tested various tracks with powered rails, and the Minecart doesn't seem to use them at all. In fact, whenever it's supposed to accelerate, it slows down and usually stops completely, even when the rails are powered up.

marvinkopf:
The RedstoneSimulator now erases cached PowerData for a position whenever an update is requested for a postion where the block doesn't have a corresponding RedstoneHandler. 

The PowerData for a position gets cached when there's a block (i.e. powered rail) with a corresponding redstone handler. When an updated is initiated for the position, the redstone handler will check if the new power is different from the cached one and will skip the update if the answer is no.

The problem is, when the block gets destroyed the cache remains unchanged, an update for the position gets initiated which the RedstoneSimulator skips* because the new block (Air) doesn't have a redstone handler. So, when an activated block is destroyed the cached power for that position still says it is activated. A new block at that position that's supposed to get activated will skip its update because the cached data wrongfully shows it is already activated.

*This is where the pull request will erase the cache for the position, before the update skips.

Reproduce the bug: Place red stone torch, place powered rail next to it (its activated), destroy the rail, place a new powered rail at the position (it is not activated). 

closes #2946 (I guess)

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
@@ -125,8 +125,11 @@ void cIncrementalRedstoneSimulator::Simulate(float a_dt)
 		}
 
 		auto CurrentHandler = cIncrementalRedstoneSimulator::CreateComponent(m_World, CurrentBlock, &m_Data);
-		if (CurrentHandler == nullptr)
+		if (CurrentHandler == nullptr)  // Block at CurrentPosition doesn't have a corresponding redstone handler
 		{
+			// Clean up cached PowerData for CurrentPosition
+			static_cast<cIncrementalRedstoneSimulator *>(m_World.GetRedstoneSimulator())->GetChunkData()->ErasePowerData(CurrentLocation);
+
 			continue;
 		}
 
--- a/src/Simulator/IncrementalRedstoneSimulator/RedstoneSimulatorChunkData.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/RedstoneSimulatorChunkData.h
@@ -41,6 +41,12 @@ public:
 		return (Result == m_MechanismDelays.end()) ? nullptr : &Result->second;
 	}
 
+	/** Erase cached PowerData for position */
+	void ErasePowerData(const Vector3i & a_Position)
+	{
+		m_CachedPowerLevels.erase(a_Position);
+	}
+
 	cRedstoneHandler::PoweringData ExchangeUpdateOncePowerData(const Vector3i & a_Position, cRedstoneHandler::PoweringData a_PoweringData)
 	{
 		auto Result = m_CachedPowerLevels.find(a_Position);


marvinkopf:
Any thoughts on this?

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -125,8 +125,11 @@ void cIncrementalRedstoneSimulator::Simulate(float a_dt)
 		}
 
 		auto CurrentHandler = cIncrementalRedstoneSimulator::CreateComponent(m_World, CurrentBlock, &m_Data);
-		if (CurrentHandler == nullptr)
+		if (CurrentHandler == nullptr)  // Block at CurrentPosition doesn't have a corresponding redstone handler
 		{
+			// Clean up cached PowerData for CurrentPosition
+			static_cast<cIncrementalRedstoneSimulator *>(m_World.GetRedstoneSimulator())->GetChunkData()->ErasePowerData(CurrentLocation);
```
tigerw: This may (or may not be) quite a hot path, so they could be some performance issues with this. However, I realise that I did not provide a better place to put it when doing the simulator, so I guess it will have to do.

src/Simulator/IncrementalRedstoneSimulator/RedstoneSimulatorChunkData.h
```diff
@@ -41,6 +41,12 @@ class cIncrementalRedstoneSimulatorChunkData : public cRedstoneSimulatorChunkDat
 		return (Result == m_MechanismDelays.end()) ? nullptr : &Result->second;
 	}
 
+	/** Erase cached PowerData for position */
+	void ErasePowerData(const Vector3i a_Position)
```
tigerw: Don't forget the reference :)

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
marvinkopf: Hm... chunk loading forces this to be called **a lot**, but it's a cheap call and the complexity should be constant. Don't think that will be a problem.

tigerw: We haven't learnt about big O notation yet, but I understand this constant may be constant but doesn't have to be a small constant? :P

madmaxoft:
Hmm, merging from deep history again, yuck :P

LogicParrot:
@madmaxoft Think we should re-enforce rebasing before merging for cleaner history? I didn't like it [earlier](https://forum.cuberite.org/thread-2252.html), but apparantly that was just a period of an abnormal amount of PRs, I don't mind it with the current PR volume.

madmaxoft:
I would very much appreciate that. We don't like spaghetti code, why should we like spaghetti history?

LogicParrot:
@NiLSPACE @bearbin To the rescue? :P

PR #2986, opened 2016-02-07
[Queue approach] Fix entity destruction in unticking chunks

### LINKED ISSUE #2731: `/butcher` command reports too many mobs

### LINKED ISSUE #2973: Investigate entity removal in unloaded chunks

LogicParrot:
Closes #2731
Closes #2973

@tigerw Is that the vision you had in mind?


LogicParrot:
I had to have a single new entity variable: m_ParentChunk. Because the parentChunk is needed for efficient queueing.

LogicParrot:
This can be used as a basis for a problem @madmaxoft once mentioned: Entities may tick twice when moving between two chunks. But I want the basic thing to be merged first.

LogicParrot:
@tigerw Is that the vision you had in mind?

src/ClientHandle.cpp
```diff
@@ -129,7 +129,6 @@ cClientHandle::~cClientHandle()
 			World->RemovePlayer(m_Player, true);  // Must be called before cPlayer::Destroy() as otherwise cChunk tries to delete the player, and then we do it again
 			m_Player->Destroy();
 		}
-		delete m_Player;
```
tigerw: You sure about this? In the current system, the ClientHandle manages the player pointer's life.

tigerw:
Could you not use DoWithChunk() as you have done before, without storing a parent chunk?

src/World.h
```diff
@@ -819,6 +822,15 @@ class cWorld :
...
 
+	struct EntityRemovalQueueItem
+	{
+		EntityRemovalQueueItem(cChunk * a_Chunk, cEntity * a_Entity, bool a_FreeMemory);
+		cChunk * m_Chunk;
+		cEntity * m_Entity;
+		bool m_FreeMemory;
+	};
+
+	std::list<EntityRemovalQueueItem> m_EntityRemovalQueue;
```
tigerw: cWorld already has a task system, I was thinking of something like:

```cpp
cWorld::QueueTask([...](...)
{
    DoWithChunk(..., ..., [...](...)
    {
        a_Chunk->RemoveEntity(); // New function to be added; delete if not player
    }
}
```

src/ClientHandle.cpp
LogicParrot: Well, no, I am not sure. `m_Player->Destroy();` will add the player destruction to the queue. The problem I can see is that `m_Player->Destroy();` is inside a conditional "if" so it's not guaranteed to free it. Guess I'll change this and make the player destruct here as usual.

LogicParrot:
@tigerw That technique was dangerous. It returns the chunk the player is **currently standing at**. At the start of each tick this is exactly the same as the **parent chunk**, but this will change if the entity crosses a border mid-tick, (and will re-converge to the same chunk when cChunk::tick calls MoveEntityToNewChunk. So, if a player teleports while walking, it may do nasty things.

The only thing guaranteed to do it right is the parent chunk (the curretly owning chunk). I am actually considering removing the `a_Chunk` parameter in the cEntity::tick method and its overrides, as this is always equal to m_ParentChunk. Note that using `a_Chunk` directly is impossible, because sometimes teleportation /destruction is called outside of a tick and therefore `a_Chunk` is not available.

src/World.h
LogicParrot: Is the task guaranteed to be executed just after the mob tick and before the next tick cycle?

LogicParrot:
Hmm, There's a chance I'm wrong in my last post. I'll need to re-check this.

LogicParrot:
Not wrong.

LogicParrot:
Also, regardless of my last argument, I think it'd be wise to have the chunk stored for convenience, rather than passing around `a_Chunk` all over the place.

src/World.h
LogicParrot: I just checked. The tasks are done before the mobs are ticked.

tigerw: Oh yeah, mobs are ticked separately (ugh). In that case, move the TickMobs call to after m_ChunkMap->Tick, and now its guaranteed :)

LogicParrot: TickMobs is already after m_ChunkMap->Tick. How does that help?

tigerw: As in directly afterwards/before TickQueuedTasks.

tigerw:
> Not wrong.

Are you certain? GetChunkX/Z() returns the current chunk based on the entity's position, DoWithChunk gets that chunk. The task will execute after Tick (after our fix); and Tick will have ensured that the entity is in the chunk that its position says it is in.

tigerw:
(this is all *so* confusing, I've already forgotten why this whole thing is needed about four times, and had to re-examine your explanation of it)

LogicParrot:
>Are you certain? GetChunkX/Z() returns the current chunk based on the entity's position, DoWithChunk gets that chunk. The task will execute after Tick (after our fix); and Tick will have ensured that the entity is in the chunk that its position says it is in.

Oh, yes, you're correct, I forgot that this will happen *after* the tick now that there's a fix. Confusing indeed!

But my guts tell me this is dangerous futurwise. It's the Cuberite equavilant of tinkering with undefined compiler behavior. Any future change may break this guarantee. m_ParentChunk seems a lot safer to me.

tigerw:
But then again, some future change may accidentally try to update ParentChunk or some future chunk refactor that completely misses the fact that there's an additional variable which needs maintenance.

I personally would do the "as little amount of variables as possible" route but I think both have pitfalls, so you decide.

LogicParrot:
Well, I've explicitly mentioned in the docs that only cChunk::addEntity has the right to update it.
I can't deny there are still dangers, but I think the dangers of your proposals are more real. For instance, the code that moves entities to neighboring chunks might one day be offloaded to the world task system to solve @madmaxoft 's double-tick issue. This will instantly break the following guarantee: ` and Tick will have ensured that the entity is in the chunk that its position says it is in.`, there are probably a million more ways this could break when doing seemingly unrelated changes. On the other hand, the only way m_ParentChunk could only break if someone sets it explicitly which is a much more verbose, much less sneaky error.

tigerw:
Very well, go for it then :)

LogicParrot:
One more question first: Should I really go through the cWorld's existing task system? How will it fare performance-wise? Perhaps it'd be wise to have a second, entity-specific task system instead? Entities will produce a lot of tasks if chunk switching is ever made a task, and maybe it'd be significantly faster without those capture lambdas? Just wondering.

LogicParrot:
My guess is that the current roll-your-own entity destructor queue is faster than the cWorld task system. I don't have any real numbers though.

tigerw:
Dunno. Maybe, maybe not, maybe the compiler does magic to it, @worktycho knows. Everything else is using QueueTask though - imagine what it would be like if they all added their own queues structures to the world! Entities shouldn't be an exception.

LogicParrot:
Done conversion to cWorld's tasks. Rebased and squashed. I like this.

src/Chunk.cpp
```diff
@@ -139,6 +139,7 @@ cChunk::~cChunk()
 	{
 		if (!(*itr)->IsPlayer())
 		{
+			(*itr)->SetParentChunk(nullptr);  // To prevent the entity from queueing a future chunk removal.
```
LogicParrot: change comment

src/ClientHandle.cpp
```diff
@@ -127,6 +127,7 @@ cClientHandle::~cClientHandle()
 			}
 
 			World->RemovePlayer(m_Player, true);  // Must be called before cPlayer::Destroy() as otherwise cChunk tries to delete the player, and then we do it again
+			m_Player->SetParentChunk(nullptr);  // Ensures the player will not call queue a delete task on itself, we want to delete it here.
```
LogicParrot: change comment

src/Entities/Entity.cpp
```diff
@@ -1522,13 +1550,25 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
 	cWorld * OldWorld = cRoot::Get()->GetWorld(GetWorld()->GetName());  // Required for the hook HOOK_ENTITY_CHANGED_WORLD
 	SetWorld(a_World);
 
 	// Entity changed the world, call the hook
 	cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*this, *OldWorld);
 
+
+	OldWorld->QueueTask([this, ParentChunk, a_World](cWorld & a_OldWorld)
+	{
+
```
tigerw: (•.•)
( •.•)>⌐■-■
(⌐■.■)
...greenspace.

LogicParrot:
Doubletick prevention seems to be working :)

LogicParrot:
I got a "moving to same chunk" error. I guess I'll leave the doubletick thing to a seperate PR ^_^

LogicParrot:
@tigerw explain what?

LogicParrot:
When you move an entity to a new chunk which hasn't yet ticked, that entity will tick twice. The solution is moving it post-tick, preventing such doubleticks. Hence the cChunk::tick task queue. But it seems to be messing something.

LogicParrot:
Rebased and removed doubletick prevention attempt.

src/Entities/Player.cpp
```diff
@@ -1746,6 +1739,16 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	// Player changed the world, call the hook
 	cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*this, *OldWorld);
 
+	OldWorld->QueueTask([this, ParentChunk, a_World](cWorld & a_OldWorld)
+	{
+		LOGD("Warping player (%s) from world (%s) to (%s). Source chunk: (%d, %d) ",
+			this->GetName().c_str(),
+			a_OldWorld.GetName().c_str(), a_World->GetName().c_str(),
+			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
```
LogicParrot: Is markDirty required for players?

LogicParrot:
I realized this commit name is misleading. It's not about unloaded chunks, it's about loaded, nonticking chunks. There's no trivial way to access entities in unloaded chunks. (You'd need to load and check all unloaded chunks from storage).

src/Chunk.cpp
```diff
@@ -621,31 +622,37 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
...
+			ASSERT((*itr)->GetParentChunk() == nullptr);
+			++itr;
+			continue;
+		}
+		ASSERT((*itr)->GetParentChunk() == this);
+
+
 		if (!((*itr)->IsMob()))  // Mobs are ticked inside cWorld::TickMobs() (as we don't have to tick them if they are far away from players)
 		{
+
```
madmaxoft: Get rid of this extra line

src/Chunk.cpp
```diff
@@ -621,31 +622,37 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
...
+		// Do not tick destroyed mobs.
+		if ((*itr)->IsDestroyed())
+		{
+			ASSERT((*itr)->GetParentChunk() == nullptr);
+			++itr;
+			continue;
+		}
+		ASSERT((*itr)->GetParentChunk() == this);
+
+
```
madmaxoft: Use only single-empty-line dividers

src/Chunk.cpp
```diff
@@ -621,31 +622,37 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
...
+			ASSERT((*itr)->GetParentChunk() == nullptr);
+			++itr;
+			continue;
 		}
-		else if (
-			((*itr)->GetChunkX() != m_PosX) ||
-			((*itr)->GetChunkZ() != m_PosZ)
+		ASSERT((*itr)->GetParentChunk() == this);
+
+
```
madmaxoft: single empty line

src/Entities/Entity.cpp
```diff
@@ -1522,13 +1550,23 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
-	a_World->AddEntity(this);
 	cWorld * OldWorld = cRoot::Get()->GetWorld(GetWorld()->GetName());  // Required for the hook HOOK_ENTITY_CHANGED_WORLD
 	SetWorld(a_World);
 
 	// Entity changed the world, call the hook
 	cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*this, *OldWorld);
 
+	OldWorld->QueueTask([this, ParentChunk, a_World](cWorld & a_OldWorld)
+	{
+		LOGD("Warping entity #%i (%s) from world (%s) to (%s). Source chunk: (%d, %d) ",
```
madmaxoft: I'd prefer world names in dblquotes `\"%s\"`; the server uses kinda convention of listing block coords in curly braces and chunk coords in square brackets.

src/Entities/Entity.cpp
```diff
@@ -1522,13 +1550,23 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
+			this->GetUniqueID(), this->GetClass(),
+			a_OldWorld.GetName().c_str(), a_World->GetName().c_str(),
+			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
+		);
+		ParentChunk->RemoveEntity(this);
+		ParentChunk->MarkDirty();
+		a_World->AddEntity(this);
+	});
```
madmaxoft: The hook call for EntityChangedWorld should be moved here, the hook description says "the entity is already in the new world"

src/Entities/Player.cpp
madmaxoft: Better call it, so that it's consistent with the other entities.

src/Entities/Player.cpp
```diff
@@ -1746,6 +1739,16 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
+		);
+		ParentChunk->RemoveEntity(this);
+		a_World->AddPlayer(this);
```
madmaxoft: Again, move the EntityChangedWorld hook invocation here.

src/Entities/Entity.cpp
LogicParrot: Wouldn't "Already in new world" mean it must be moved to a_World's addEntity's internals?

src/Entities/Player.cpp
LogicParrot: Same question as above.

src/Entities/Player.cpp
LogicParrot: But I see `if entity is not a player then call markDirty` everywhere.

src/Entities/Entity.cpp
LogicParrot: Okay, I'll move it. Note that it'll be in the new world's `m_ClientsToAdd` when the hook is called. I guess that's fine.

src/Entities/Player.cpp
LogicParrot: Done.

LogicParrot:
Rebased.

SphinxC0re:
Please squash the commits onto the current master

LogicParrot:
@SphinxC0re done.

LogicParrot:
Please do not merge yet, I want to be completely sure I got all pointers right in all cases first.

src/Chunk.cpp
```diff
@@ -621,37 +622,44 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
...
 			// The entity moved out of the chunk, move it to the neighbor
-			MarkDirty();
+
+			(*itr)->SetParentChunk(nullptr);
 			MoveEntityToNewChunk(*itr);
 			itr = m_Entities.erase(itr);
+			// Mark as dirty if it was a server-generated entity:
+			if (!(*itr)->IsPlayer())
+			{
+				MarkDirty();
```
LogicParrot: I think this line will save everyone thousands of false dirties a day. It seems that previously, anywhere you walk into would be marked dirty because the `if player` condition was missing.

src/Chunk.cpp
```diff
@@ -621,37 +622,44 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 
 	for (cEntityList::iterator itr = m_Entities.begin(); itr != m_Entities.end();)
 	{
+		// Do not tick mobs that are detached from the world. They're either schedule for teleportation or for removal.
```
LogicParrot: scheduled

src/Chunk.cpp
```diff
@@ -621,37 +622,44 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
...
 		if (!((*itr)->IsMob()))  // Mobs are ticked inside cWorld::TickMobs() (as we don't have to tick them if they are far away from players)
 		{
 			// Tick all entities in this chunk (except mobs):
+			ASSERT((*itr)->GetParentChunk() == this);
 			(*itr)->Tick(a_Dt, *this);
+			ASSERT((*itr)->GetParentChunk() == this);
 		}
 
-		if ((*itr)->IsDestroyed())  // Remove all entities that were scheduled for removal:
+		// Do not move mobs that are detached from the world to neighbors. They're either schedule for teleportation or for removal.
```
LogicParrot: scheduled

LogicParrot:
I have replaced `IsDestroyed` with `IsTicking`.  `IsDestroyed` returns true only when an entity is about to be deleted. `!IsTicking()` returns true if the entity is either about to teleport or about to be deleted.

LogicParrot:
I put all the player destruction logic in `cClientHandle::~cClientHandle:`. Previously, it was dispersed into 3 functions for no apparent reason: `cClientHandle::Destroy()`, `cClientHandle::ClosedSocket()`, `cClientHandle::~cClientHandle()`. Was there a reason for that? Did I miss anything here?

src/Entities/Entity.cpp
```diff
@@ -196,20 +193,59 @@ void cEntity::WrapSpeed(void)
...
 {
-	if (!m_IsInitialized)
+	ASSERT(IsTicking());
+	SetIsTicking(false);
+
+	if (a_ShouldBroadcast)
 	{
-		return;
+		m_World->BroadcastDestroyEntity(*this);
 	}
```
LogicParrot: assert parent chunk not null

src/Entities/Entity.h
```diff
@@ -252,11 +252,18 @@ class cEntity
...
-	/** Schedules the entity for destroying; if a_ShouldBroadcast is set to true, broadcasts the DestroyEntity packet */
+	/** Returns true if the entity is valid and ticking. Returns false if the entity is not ticking.
+	When the entity is not ticking, it is either scheduled for teleportation or for destruction.
+	Either way, using it is dangerous. */
+	bool IsTicking(void) const;
+
+	/** Destroys the entity and schedules it for memory freeing; if a_ShouldBroadcast is set to true, broadcasts the DestroyEntity packet */
 	void Destroy(bool a_ShouldBroadcast = true);
 
+	/** Destroy the entity without scheduling memory freeing. This should only be used by cChunk for internal memory management. */
```
LogicParrot: by cChunk and cClientHandler

LogicParrot:
What is wrong with the CI?

SphinxC0re:
Nothing. GCC is throwing so many errors, that it stops counting and aborts

LogicParrot:
```
CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.

Missing variable is:

CMAKE_CXX_COMPILER_ENV_VAR

CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.

Missing variable is:

CMAKE_CXX_COMPILER

CMake Error: Could not find cmake module file:/home/travis/build/cuberite/cuberite/CMakeFiles/CMakeCXXCompiler.cmake

CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.

Missing variable is:

CMAKE_C_COMPILER_ENV_VAR

CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.

Missing variable is:

CMAKE_C_COMPILER

CMake Error: Could not find cmake module file:/home/travis/build/cuberite/cuberite/CMakeFiles/CMakeCCompiler.cmake

CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage

CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage

-- Configuring incomplete, errors occurred!
```

LogicParrot:
It is throwing errors in unchanged code.

src/ClientHandle.cpp
```diff
@@ -164,18 +165,19 @@ void cClientHandle::Destroy(void)
...
 
-	if ((m_Player != nullptr) && (m_Player->GetWorld() != nullptr))
-	{
-		RemoveFromAllChunks();
-		m_Player->GetWorld()->RemoveClientFromChunkSender(this);
-	}
 	if (m_Player != nullptr)
 	{
+		cWorld * World = m_Player->GetWorld();
+		ASSERT(World != nullptr);
```
LogicParrot: Should there be an assert here?

tigerw:
great, IsTicking came back...

LogicParrot:
@tigerw `IsDestroyed` must be replaced. We need something which returns true for either destruction or teleportation, not just destruction.

tigerw:
Could you explain a bit more the problem? Why does anything need to distinguish between these states?

LogicParrot:
@tigerw Okay.

> Why does anything need to distinguish between these states?

Actually, we do NOT need to distinguish between teleportation and destruction, and this is why I made this change.

For instance, cChunk previously had this: 
```
for each entity
  if not IsDestroyed() then tick
```

The problem here is that teleporting entities will get ticked. So I replaced it with

```
for each entity
  if IsTicking() then tick
```

`IsTicking()` plays the role of the previous `!IsDestroyed()`, but now it returns a false for teleporting entities too, not just dying ones.

LogicParrot:
@tigerw Okay.

> Why does anything need to distinguish between these states?

Actually, we do NOT need to distinguish between teleportation and destruction, and this is why I made this change.

For instance, cChunk previously had this: 
```
for each entity
  if not IsDestroyed() then tick
```

The problem here is that teleporting entities will get ticked. So I replaced it with

```
for each entity
  if IsTicking() then tick
```

`IsTicking()` plays the role of the previous `!IsDestroyed()`, but now it returns a false for teleporting entities too, not just dying ones. In other words, I eliminated distinguishing between the two states.

src/Entities/Entity.h
```diff
@@ -285,6 +284,14 @@ class cEntity
 
 	// tolua_end
 
+	/** Returns true if the entity is valid and ticking. Returns false if the entity is not ticking.
+	When the entity is not ticking, it is either scheduled for teleportation or for destruction.
+	Either way, using it is dangerous. */
+	bool IsTicking(void) const;
```
LogicParrot: I removed `IsDestroyed`. How do I deal with the LUA API here?

Perhaps I should export `IsTicking`, and make `IsDestroyed` a deprecated alias for `!IsTicking()`?.

src/Entities/Player.cpp
```diff
@@ -146,6 +146,22 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
...
+void cPlayer::InitializePlayer()
+{
+	ASSERT(GetWorld() != nullptr);
+	ASSERT(GetParentChunk() == nullptr);
+	GetWorld()->AddPlayer(this);
+
+	cPluginManager::Get()->CallHookSpawnedEntity(*GetWorld(), *this);
+
+	// Spawn the entity on the clients:
+	GetWorld()->BroadcastSpawnEntity(*this);
```
tigerw: You sure about this? This could spawn the player on every client in the world, regardless of visibility.

src/Entities/Player.cpp
```diff
@@ -146,6 +146,22 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 
 
 
+void cPlayer::InitializePlayer()
```
tigerw: Reason for distinction? cEntity::Initialize also adds the entity to the chunk, is this done somewhere?

src/Entities/Player.cpp
tigerw: Though to be honest the old code did this too, so nevermind.

src/Chunk.cpp
```diff
@@ -139,7 +139,8 @@ cChunk::~cChunk()
 	{
 		if (!(*itr)->IsPlayer())
 		{
-			(*itr)->Destroy(false);
+			// Scheduling a normal destruction is neither possible (Since this chunk will be gone till the schedule occurs) nor necessary.
+			(*itr)->DestroyNoScheduling(false);  // No point in broadcasting in an unloading chunk. Chunks unload when no one is nearby.
```
tigerw: Should've let DoWithChunk handle it for you ;)

src/Entities/Entity.h
tigerw: ...don't bother...

(don't tell anyone I said that)

src/Entities/Player.cpp
LogicParrot: The distinction is because we need to call call `GetWorld()->AddPlayer(this);` instead of calling `GetWorld()->AddEntitythis);`.
The the old code did something hacky instead:
1. For every player, it calls `initialize` which calls `world->addEntity`
2. Then, it calls `world->AddPlayer`, but to avoid the player being added twice, there's a `addEntityIfNotPresent` inside `world->AddPlayer`.

So basically the old code adds the player twice, but aborts the second add.

src/Entities/Player.cpp
LogicParrot: On a slightly related note, I think the teleport packets could do some tweaking. Sometimes I get nether fog in overworlds or vice versa. Someone reported skeleton horses, I suspect this could be wrong spawn packets or something.

LogicParrot: But that deserves a seperate PR of course :)

src/Chunk.cpp
LogicParrot: But `DoWithChunk` is both slower and not future-safe!

LogicParrot: Actually, how exactly would `DoWithChunk` help in this scenario?

src/Entities/Player.cpp
tigerw: Why not utilise the magic of vtables and overload Init for players?

src/Chunk.cpp
tigerw: Ah, whilst I concur that pointers are indeed faster, I am of the humble opinion that their complex manipulation is plainly unsafe, speed notwithstanding ;)

tigerw: It would perform no action if the addressed chunk were not valid, ideally.

src/Entities/Player.cpp
LogicParrot: Will do.

src/Chunk.cpp
LogicParrot: I am still concerned about using `DoWithChunk`, the reasoning was mentioned earlier in this PR.

LogicParrot: Specifically, any future code shuffling may break the assumption that an entity is always within its parent chunk's coordinates when `DoWithChunk` is called. I am not even sure this is always the case in the current code.

src/Entities/Player.cpp
LogicParrot: @tigerw Now I see why I haven't done this before. It's because the player version does not need a world parameter.

LogicParrot:
Todo: Document the new API function`IsTicking`, mark `IsDestroyed` as deprecated, and thats it, I guess.

src/Entities/Player.cpp
tigerw: Doesn't matter, just don't use the parameter.

LogicParrot:
Done.

tigerw:
Looks good.

I shall, of course, take every available opportunity to tease you about ParentChunk :D

LogicParrot:
@tigerw is evil!

Issue #2959, opened 2016-02-04
Update GETTING-STARTED.md


LogicParrot:
I created a new tag. ingame-testing-needed. The name is self explanatory. ~~It is used for PRs only~~. It should be explained in "getting started" as it is great for beginners, just like the "easy" tag.

Edit: The idea is that maintainers put it on PRs, and essentially it's a "please test this, people" request. non-members can pull it, test it, and report back.

LogicParrot:
Preferably, we should assume they don't know git well, and briefly explain how to pull the PR they'd like to test.

Schwertspize:
@LogicParrot non-members of the project/group can't use tags.

But a git explanation would be nice

LogicParrot:
@Schwertspize The idea is that maintainers put it on PRs, and essentially it's a "please test this, people" request. non-members can pull it, test it, and report back.

Ajusa:
I like this idea. I would be fine with testing and uploaded video of the thing that was tested. If it speeds up development, I'll do it.
I also have a few computers to test it one and report back, so that is useful.

LogicParrot:
Tester feedback significantly speeds up development. I'll be putting those tags whenever I can. You can keep an eye on them [here](https://github.com/cuberite/cuberite/issues?utf8=%E2%9C%93&q=label%3Aingame-testing-needed+).

PR #2989, opened 2016-02-08
Prevent player falling when block breaking is denied

### LINKED ISSUE #2669: Block breaking denied, but player can slip trough
Example
Block breaking denied with plugin.
Player can still fall trough the ground, if he break the block under his feet.

LogicParrot:
Closes #2669

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1033,6 +1033,7 @@ void cClientHandle::HandleLeftClick(int a_BlockX, int a_BlockY, int a_BlockZ, eB
 	{
 		// A plugin doesn't agree with the action, replace the block on the client and quit:
 		m_Player->GetWorld()->SendBlockTo(a_BlockX, a_BlockY, a_BlockZ, m_Player);
+		SendPlayerPosition();  // Prevents the player from falling through the block that was temporarily broken client side.
 		return;
 	}
 
@@ -1229,6 +1230,7 @@ void cClientHandle::HandleBlockDigFinished(int a_BlockX, int a_BlockY, int a_Blo
 	{
 		// A plugin doesn't agree with the breaking. Bail out. Send the block back to the client, so that it knows:
 		m_Player->GetWorld()->SendBlockTo(a_BlockX, a_BlockY, a_BlockZ, m_Player);
+		SendPlayerPosition();  // Prevents the player from falling through the block that was temporarily broken client side.
 		return;
 	}
 


NiLSPACE:
Is this tested? I believe you have to teleport a player .001 above a block in order to prevent them from falling down.

madmaxoft:
This should be enough, but I'd still like a confirmation from someone who actually tested this in-game.

We should add instructions for testing whenever the `ingame-testing-needed` label is used.

Simplest test case: go to server spawn area, make sure you are not in admin rank, break a block (the default configuration should prevent you from breaking it). Now try breaking the block directly beneath you; before this fix you'd fall into the ground, after this fix you should stay on top of the ground (possibly sinking for small amount of time but always being teleported back to the top).

Seadragon91:
This works fine :)

LogicParrot:
> We should add instructions for testing whenever the ingame-testing-needed label is used.

@madmaxoft Agreed, hence #2959

@Seadragon91 Yay :)

SphinxC0re:
Looks good

tigerw:
woah woah woah, what happens if a player happens to be walking in some direction and breaks a block, say, beside them? Will they get unexpectedly teleported back despite not actually going through anything? Isn't this more of a task for the physics engine?

SphinxC0re:
Yeah it should be. But for now, this is an okayish fix

LogicParrot:
@tigerw Physics engine? [KISS](https://en.wikipedia.org/wiki/KISS_principle). If the block breaking is **denied**, then yes, a player walking in some direction will be teleported back. But typically block breaks are not denied.

LogicParrot:
Why okayish? This is a cool fix. The simplest thing possible, a fully-fledged anti-cheat system server-side player physics system should be very low in the Cuberite priorities right now.

tigerw:
Rare bugs are still bugs though, unless you call them features (we have a lot of those instances :P) Let's call this a stopgap solution.

> KISS
Says the guy of twenty extra member variables. Also, I think I was just subtly insulted via initialism ;)

LogicParrot:
@tigerw Yes but this is an extremely minor bug. A player will jerk a bit when breaking an unbreakable block.

LogicParrot:
@tigerw Yes but this is an extremely minor bug. A player will jerk a bit only when breaking an **unbreakable** block while **walking**.

tigerw:
Could be the difference between escaping and getting ~~360noscope'd~~ eliminated in a PvP match :)

LogicParrot:
@tigerw Why are you ignoring the "unbreakable" part? When was the last time anyone decided to break an unbreakable block in a PvP match? These code paths will not even trigger for say dirt or cobblestone or whatever regular block there is. The block has to be break-denied by a plugin in order for the jerk to occur.

bearbin:
Well, in a PvP arena, all blocks could easily be unbreakable due to the plugin being used. But I agree, this is not really a major issue. We should keep around an issue/reminder to fix it in future, but it's good enough for the moment.

LogicParrot:
@bearbin Good point, but they'd still have to press and hold for a while, a highly unlikely thing to do in PvP.

tigerw:
Snow + shovels?

PR #2990, opened 2016-02-08
Add DefaultWorld to newly generated INI

LogicParrot:
Fixes a bug introduced in #2815 - DefaultWorld was not in the default generated INI file.


LogicParrot:
Merge asap if compiles.

LogicParrot:
Slight correction: Due to this bug, DefaultWorld will not appear in the default world.ini initially. But it will be added later, once the server saves to the INI files.

PR #2992, opened 2016-02-08
Remove nproc to allow BSD compilation

LogicParrot:
Reference: https://forum.cuberite.org/thread-2332.html


madmaxoft:
CI builds failed, how could you merge!?!

SphinxC0re:
@LogicParrot that's what I told you before...

Seadragon91:
Looking at the change, it replaces `nproc` with a 2. I don't think, that there is a need to run the whole CI for this simple change...

LogicParrot:
CI did not fail. It's because I clicked merge before it completed. This is a 1-line bash change. as @Seadragon91 said.

Issue #2993, opened 2016-02-08
cChunk::MobCensus's intentions are not clear


LogicParrot:
cChunk::MobCensus appears to fill a list with N*M entries. Where N is the number of mobs in chunk and M is the number of clients that loaded that chunk. Why?

I believe the original intention was to compare every mob with every single player on the server and fill out the distances between them.

It seems this is what the author of cWorld::TickMobs had assumed. This wrong assumption led to the writing of a malfunctioning mob destroyer. The mob destroyer is supposed to destroy faraway mobs, but in reality the destroyer works on teleports, but doesn't work when players walk away from mobs. That's because walking away has time to remove the player from `m_LoadedByClient` before the destroyer code runs, but teleportation does not.

madmaxoft:
Closed without a reason?

LogicParrot:
Closed because we talked about iterating all `m_LoadedByClient` vs iterating all online players in the current world in #2994.

PR #2994, opened 2016-02-08
Rewrite mob tick code

### LINKED ISSUE #277: Mob code too slow

### LINKED ISSUE #2991: Mob cleanup code is potentially removing livestock
Livestock should not be removed if the player is too far. The mob code [here](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L1144) seems to be doing just that. 

@jammet and others, did you ever experience livestock disappearance?

LogicParrot:
- I've rewritten mob tick code.
- cMobCensus is no longer used.
- Closes #2991
- I believe the performance is better. May close #277
- Solves the teleportation issue mentioned in #2993

Note: Only cWorld was actually changed by this PR. The rest of the file changes are from the #2986 dependency.

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -1135,28 +1135,51 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
 		}  // for i - AllFamilies[]
 	}  // if (Spawning enabled)
 
-	// move close mobs
-	cMobProximityCounter::sIterablePair allCloseEnoughToMoveMobs = MobCensus.GetProximityCounter().getMobWithinThosesDistances(-1, 64 * 16);// MG TODO : deal with this magic number (the 16 is the size of a block)
-	for (cMobProximityCounter::tDistanceToMonster::const_iterator itr = allCloseEnoughToMoveMobs.m_Begin; itr != allCloseEnoughToMoveMobs.m_End; ++itr)
+	class cCallback : public cEntityCallback
 	{
-		cEntity & Entity = itr->second.m_Monster;
-		cChunk * Chunk = Entity.GetParentChunk();
-		if (!Entity.IsTicking())
+		virtual bool Item(cEntity * a_Entity) override
 		{
-			++itr;
-			continue;
-		}
-		ASSERT(Chunk == &(itr->second.m_Chunk));
-		Entity.Tick(a_Dt, *Chunk);
-		ASSERT(Chunk == &(itr->second.m_Chunk));
-	}
+			if (!a_Entity->IsMob())
+			{
+				return false;
+			}
+			if (!a_Entity->IsTicking())
+			{
+				return false;
+			}
 
-	// remove too far mobs
-	cMobProximityCounter::sIterablePair allTooFarMobs = MobCensus.GetProximityCounter().getMobWithinThosesDistances(128 * 16, -1);// MG TODO : deal with this magic number (the 16 is the size of a block)
-	for (cMobProximityCounter::tDistanceToMonster::const_iterator itr = allTooFarMobs.m_Begin; itr != allTooFarMobs.m_End; ++itr)
-	{
-		itr->second.m_Monster.Destroy(true);
-	}
+			auto Monster = static_cast<cMonster *>(a_Entity);
+			ASSERT(Monster->GetParentChunk() != nullptr);  // A ticking entity must have a valid parent chunk
+
+			// Tick close mobs
+			if (Monster->GetParentChunk()->HasAnyClients())
+			{
+				Monster->Tick(m_Dt, *(a_Entity->GetParentChunk()));
+			}
+			// Destroy far hostile mobs
+			else if ((Monster->GetMobFamily() == cMonster::eFamily::mfHostile))
+			{
+				if (Monster->GetMobType() != eMonsterType::mtWolf)
+				{
+					Monster->Destroy(true);
+				}
+				else
+				{
+					auto Wolf = static_cast<cWolf *>(Monster);
+					if (Wolf->IsAngry())
+					{
+						Monster->Destroy(true);
+					}
+				}
+			}
+			return false;
+		}
+	public:
+		std::chrono::milliseconds m_Dt;
+	} Callback;
+
+	Callback.m_Dt = a_Dt;
+	ForEachEntity(Callback);
 }
 
 


LogicParrot:
Future todo: Optimize this by first checking out the parent chunk's m_LoadedByClient list and only afterwords appealing to FindClosestPlayer.

src/Chunk.cpp
```diff
@@ -621,31 +622,37 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 
 	for (cEntityList::iterator itr = m_Entities.begin(); itr != m_Entities.end();)
 	{
```
SphinxC0re: Is it possible for you to refactor this loop into an iterative c++11 loop

LogicParrot: But we'd need the iterator for the erase below.

SphinxC0re: Oh! I didn't see that.

tigerw: @SphinxC0re I essentially mirror your exact thought process every time I see this.

@LogicParrot Perhaps you could auto-ify the type if you do a rebase later? There's no need to inconvenience yourself by doing this individually.

src/World.cpp
```diff
@@ -1130,19 +1130,46 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
...
-		itr->second.m_Monster.Tick(a_Dt, itr->second.m_Chunk);
-	}
+		virtual bool Item(cEntity * a_Entity) override
+		{
+			if (!a_Entity->IsMob())
+			{
+				return false;
+			}
+			cMonster * Monster = static_cast<cMonster*>(a_Entity);
+			cPlayer * Player = a_Entity->GetWorld()->FindClosestPlayer(a_Entity->GetPosition(), 128, false);
```
madmaxoft: So we're calculating the closest player `<NumMobs>` times here, won't this be too much for the server?

madmaxoft:
Finally someone tackled the part of the code I grew to hate the most, the dreaded MobCollector.

Still, I think there could be an even better solution to all this. Not sure if this PR would be the place to implement this, but at least to give you an idea: Do mobs really need to keep track of the closest player all the time? I don't think so. I think we should tick all mobs in all loaded chunks, but instead the AI should decide whether to perform any player-based checks or not, based on an approximate distance to "probably the nearest" player. Instead of checking the player on each tick, we should have an `ApproxNearestPlayerDistance` value stored for each mob that is updated once every, say, 20 ticks, and based on this value the mob can decide whether to do in-depth AI processing or not. Mobs with high value could even be frozen to avoid any pathfinding performance penalties.
The distance needn't even be a number, it could be an enum of `short`, `medium` and `far`, this would simplify updating even further (for medium+ distance, update all mobs in a chunk at once, assigning the same value to them)

LogicParrot:
@madmaxoft There are lots of ways to optimize this, I'll definitely tackle that and consider your proposals. But I really wanna close #2991 and #2993 asap first. Especially #2991; undead entities are really bad. So I propose we merge this and improve performance later.

src/World.cpp
LogicParrot: Yes. There's lots of room for improvement, but I want animals to stop disappearing asap (#2991), and only optimize later.

LogicParrot:
@madmaxoft There are lots of ways to optimize this, I'll definitely tackle that and consider your proposals. But I really wanna close #2991 and #2993 asap first. Especially #2991; unexpectedly destroyed animals are really bad. So I propose we merge this and improve performance later.

LogicParrot:
@madmaxoft  A slightly different approach:
Each mob has an `ApproxNearestPlayerPosition`, an `InvalidationTimer` which represents freshness of the value and keeps ticking down to zero., and a `GetApproxNearestPlayerPosition()` which does this:
```
   if (InvalidationTimer > 0)  // Cached is approximation is not too old, we can still use it.
   {
      return ApproxNearestPlayerPosition;
   }
   if (one of the mobs in the same chunk have an approximation which is not too old)
   {
      ApproxNearestPlayerPosition = them.ApproxNearestPlayerPosition;
      InvalidationTimer = them.InvalidationTimer;
      return ApproxNearestPlayerPosition;
   }
   if (there are players in our parent chunk's m_LoadedClient)
   {
      // Iterate only those players and get the closest one
      ApproxNearestPlayerPosition = them.GetPosition();
      InvalidationTimer = 40;  // Or whatever amount of ticks we'd like to cache this
      return ApproxNearestPlayerPosition;
   } 
   // As a last resort, iterate every player in the world
```

LogicParrot:
I used a position and not a distance because it is shared among mobs and they can do `myPosition - playerposition` to deduce distance.

madmaxoft:
The last resort is not really needed, if there are no players in the `m_LoadedClient` list, then all players are too far from this mob anyway. That's why I proposed to use an enum instead of numerical values.

LogicParrot:
@madmaxoft Incorrect. This assumption is in part why the current mob ticking is messed up. Depending on the client view distance, the player may be, say, 100 blocks away, making them not in m_LoadedClient, but the mob would need to track that player, so that if the player reaches the 128 blocks away mark the mob self-destructs. (vanilla mobs self-destruct at the 128 mark).

LogicParrot:
@madmaxoft By the way, this is still problematic even after this PR, because the chunk may unload before the mob gets the chance to self destruct.

LogicParrot:
(I was referring to hostile mobs in the self destruct part)

madmaxoft:
Note that I'm splitting the mob ticking and the AI processing in my thoughts. Mobs should always tick, no matter the distance from a player, but their AI should not perform extensive tasks if they are too far away - and exactly then is this optimization useful.

LogicParrot:
@madmaxoft I see. Then we'd still need the last resort for the generic ticking.

As mentioned earlier, hostiles have the additional special trait of needing to tick even in **unloaded chunks**, so that they self-destruct if no player is nearby. Any thoughts on how to accomplish this?

LogicParrot:
We could prevent chunks that have hostiles from ever unloading. That would give the hostiles time to self-destruct, and then the chunk would unload.

madmaxoft:
When the `m_LoadedClient` list gets empty, destroy all hostiles.

I assume unloaded -> un-playered :)

LogicParrot:
@madmaxoft sounds simple, but that would mean mobs can get destroyed far closer than 128 chunks, depending on player view distance. Is that ok?

madmaxoft:
If the players can't see them, there's no harm in destroying them.

I'm more concerned about the save/load scenario - a player leaves the server, the server should save the chunks and unload them, and restore them, including the hostiles, when the player re-logs. Otherwise you could simply cheat by disconnecting when being overrun by hostiles.

LogicParrot:
@madmaxoft Yes, that's an issue. You could simply cheat by lowering view distance as well.

LogicParrot:
@madmaxoft Hmm, you could **always** cheat by disconnecting, regardless.

LogicParrot:
I remember a plugin which killed the player if they disconnect mid combat.

LogicParrot:
@madmaxoft if player position are saved in NBT, then it'd be as simple as: `If !chunk->HasClients() && !chunk->HasOfflineClients() then destroy mobs. `

LogicParrot:
I've pushed a new commit, tell me what you think.

src/World.cpp
```diff
@@ -1130,19 +1130,36 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
...
+		{
+			if (!a_Entity->IsMob())
+			{
+				return false;
+			}
+
+			cMonster * Monster = static_cast<cMonster*>(a_Entity);
+
+			// Tick close mobs
+			if (Monster->GetParentChunk()->HasAnyClients())
```
madmaxoft: Is there any chance the parent chunk can be `nullptr`?

src/World.cpp
```diff
@@ -1130,19 +1130,36 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
...
-		itr->second.m_Monster.Tick(a_Dt, itr->second.m_Chunk);
-	}
+		virtual bool Item(cEntity * a_Entity) override
+		{
+			if (!a_Entity->IsMob())
+			{
+				return false;
+			}
+
+			cMonster * Monster = static_cast<cMonster*>(a_Entity);
```
madmaxoft: `auto`, space between `cMonster` and `*`

src/World.cpp
LogicParrot: That depends on which version of my commit this is, I don't remember for sure anymore.
I want to thoroughly review all cases for Get/SetParentChunk uses before merging #2986 , I'll change this accordingly, if needed, once done.

tigerw: (should've used GetChunkX() ;) )

LogicParrot: As discussed earlier, I still believe this is the cleanest approach, it's just that:
A: I want to be completely sure it's all good, because I slightly changed the way parentChunk is used mid-PR.
B: I am seeing potential redundancy here and I'd like to investigate if `isDestroyed` and `GetParentChunk` can be combined.

LogicParrot: ftr: Got rid of IsDestroyed in #2986

LogicParrot: This was resolved. The parentChunk should will never be nullptr for a ticking entity. An assert was added just in case.

LogicParrot:
@tigerw Thanks for `ParentChunk` which you so intensly despise, `cMobCensus` was eliminated from the tick code ;)

LogicParrot:
I think this is finished. Monster destruction is non-optimal,  but that's better than non-working. Destruction should be later tweaked to handle some cases (e.g. the cheat by logout @madmaxoft mentioned),

tigerw:
All roads lead to Rome...

PR #2995, opened 2016-02-08
Improve world linkage error message

LogicParrot:
Now that linkage rules are enforced, many will get a startup error. To prevent worldwide panic, I decided to tell them nothing is terribly wrong.


LogicParrot:
Merged (Don't worry, CI is fine, just a string change)

PR #2996, opened 2016-02-08
Add forum link to linkage error message

LogicParrot:


src/World.cpp
```diff
@@ -558,7 +558,7 @@ void cWorld::Start(void)
 	if (!BadWorlds.empty())
 	{
 		const char * WorldName = m_WorldName.c_str();
-		LOGERROR("\n###### ERROR: %s is linked to the following nonexisting world/s:\n%s\n\nPlease edit %s/world.ini and fix this.\n\nNote that the server started enforcing proper world linkages recently. And people with older configs may naturally get this error. If you just want a working default config and don't mind losing this world, delete the folder \"%s\" and the server will receate one for you. Otherwise edit the world.ini file and fix the invalid linkages.\n######\n",
+		LOGERROR("\n###### ERROR: \"%s\" is linked to the following nonexisting world/s:\n%s\n\nPlease edit %s/world.ini and fix this.\n\nNote that the server started enforcing proper world linkages recently. And people with older configs may naturally get this error. If you just want a working default config and don't mind losing this world, delete the folder \"%s\" and the server will receate one for you. Otherwise edit the world.ini file and fix the invalid linkages.\n######\n",
```
tigerw: Ah, about this, can you remove the slightly random newlines and the number sign (hash)? It literally occurs nowhere else, even when the server crashes spectacularly (i.e. often).

LogicParrot: I wanted to make it as different as possible because this is an error lots of people are going to get when starting their servers in the next few days, and I didn't want them to confuse it with a crash. The new lines are for tidyness's sake, unlike stacktraces, lots of people will read this. I wanted to explain the issue in detail.

tigerw: Why do so many have broken linkage anyway?

LogicParrot: Because the server happily generated worlds that were never added to `settings.ini` when people visited nethers or ends.

LogicParrot: Oh and https://github.com/cuberite/cuberite/pull/2821 seems to have made it worse in some cases, by generating a `world_nether` whose linkages are `world_nether_nether` and `world_nether_end`. and a `world_end` whose linkages are `world_end_nether` and `world_end_end`

LogicParrot:
I rebased this and added a forum link.

PR #3009, opened 2016-02-16
[WIP] Fix minecart entity collision

### LINKED ISSUE #2944: Pushing Minecarts is slow
Walking into a Minecart to push it very often has no, or just a very weak effect. You walk all the way through it sometimes, and it never moved.

marvinkopf:
* Minecarts no longer handle a collision if the entity is behind them.
* Minecarts will leave the pushing after a collision on a straight rail to the entity.

Minecarts would handle it if there was a "collision" with an entity which was in the opposite direction they are moving by slowing down, aka player running into the minecart from behind. I saw no reason why they would do that, the case is no longer handled. 

On straight rails minecarts will just stop now on collision with an entity, if the entity can push it will push the minecart.

fixes #2944

### DIFF
--- a/src/Entities/Minecart.cpp
+++ b/src/Entities/Minecart.cpp
@@ -845,24 +845,16 @@ bool cMinecart::TestEntityCollision(NIBBLETYPE a_RailMeta)
 		{
 			if (MinecartCollisionCallback.GetCollidedEntityPosition().z >= GetPosZ())
 			{
-				if ((-GetSpeedZ() * 0.4) < 0.01)
+				if (GetSpeedZ() > 0)  // True if minecart is moving into the direction of the entity
 				{
-					AddSpeedZ(-4);
-				}
-				else
-				{
-					SetSpeedZ(-GetSpeedZ() * 0.4);
+					SetSpeedZ(0);  // Entity handles the pushing
 				}
 			}
-			else
+			else  // if (MinecartCollisionCallback.GetCollidedEntityPosition().z < GetPosZ())
 			{
-				if ((GetSpeedZ() * 0.4) < 0.01)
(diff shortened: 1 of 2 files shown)


src/Entities/Minecart.cpp
```diff
@@ -845,24 +845,16 @@ bool cMinecart::TestEntityCollision(NIBBLETYPE a_RailMeta)
 		{
 			if (MinecartCollisionCallback.GetCollidedEntityPosition().z >= GetPosZ())
 			{
-				if ((-GetSpeedZ() * 0.4) < 0.01)
+				if (GetSpeedZ() > 0)  // Check if entity is in front of minecart
```
madmaxoft: This needs more explanation. How does `GetSpeedZ` relate to an entity being in front of the minecart?

marvinkopf: The condition here didn't change (much). GetSpeedZ can be positive or negative depending on the direction the minecart is moving to. We already know in which direction the entity is due to the outer if-clause, so we now need to check if the minecart is moving into that direction.

LogicParrot:
lgtm

marvinkopf:
Made the comments a bit more clear

Woazboat:
Seems to work fairly well.
Pushing minecarts up steep (multiple block high) slopes is not really possible though. Minecarts also sometimes derail and pop off the sides of the tracks. (Not sure if this was introduced by this change or if it already happened before.)

SphinxC0re:
lgtm

marvinkopf:
I'll see if I can do something about the up steep pushing.

madmaxoft:
Looks good to me. Let's merge this and handle the up steep pushing in a separate PR, since it is most likely unrelated to this code.

PR #3010, opened 2016-02-16
Fix unicode handling in StringUtils::TrimString

### LINKED ISSUE #3008: Changes in settings.ini ignored/reverted
Yesterday I noticed that for some reason, my server's name was altered somehow. The color code was missing. settings.ini used to look like this:
Description=§1L§9o§1n§9e§1l§9y §eI§2s§el§6a§2n§ed §1M§9i§1n§9e§1c§9r§1a§9f§1t

That says "Lonely Island Minecraft" and has nice colors to go along with it.

Now it says: 
Description=1L§9o§1n§9e§1l§9y §eI§2s§el§6a§2n§ed §1M§9i§1n§9e§1c§9r§1a§9f§1t

The first color code is truncated.

Now. If I make ANY changes to this, it's immediately overwritten back to the truncated version without the color code again. Also, I wrote entire sentences, and they were  #gone; the only trace left would be that settings.ini~ file in the same directory, which still showed my changes. Something keeps overwriting the file with the old information, but where else is that stored?

LogicParrot:
closes #3008.

### DIFF
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -222,7 +222,7 @@ AString TrimString(const AString & str)
 	size_t start = 0;
 	while (start < len)
 	{
-		if (str[start] > 32)
+		if (static_cast<unsigned char>(str[start]) > 32)
 		{
 			break;
 		}
@@ -236,7 +236,7 @@ AString TrimString(const AString & str)
 	size_t end = len;
 	while (end >= start)
 	{
-		if (str[end] > 32)
+		if (static_cast<unsigned char>(str[end]) > 32)
 		{
 			break;
 		}


SphinxC0re:
What a lovely name for a pr :smile: merge if CI succeeds

LogicParrot:
I have no idea how that got there :P

src/StringUtils.cpp
```diff
@@ -236,7 +236,7 @@ AString TrimString(const AString & str)
 	size_t end = len;
```
Woazboat: Could just as well be `size_t end = len - 1;` to avoid unnecessarily comparing the null character at the end of the string. 
http://www.cplusplus.com/reference/string/string/operator%5B%5D/

worktycho:
So we require utf-8 or ascii chars. Might want to document that.

worktycho:
Can we just replace it with std::isspace?

worktycho:
This covers a fraction of Unicodes spaces:https://www.cs.tut.fi/~jkorpela/chars/spaces.html

madmaxoft:
Cuberite tries to be encoding-agnostic throughout the codebase implicitly, unless noted otherwise; this is one of the places where this philosophy is somewhat dangerous (we could cut in the middle of an UTF8 char), but since we really only cut away chars in the 0-32 range, I don't care.

The PR's title is misleading, there's no handling of unicode actually.

LogicParrot:
As implied in #3008, this PR is guaranteed to never cut in the middle of a UTF-8 char.
However, it will not trim exotic whitespace mentioned by @worktycho, but it trims regular tabs and spaces properly. 
Should we bother trimming the rest of the UTF-8 whitespace?

LogicParrot:
> The PR's title is misleading, there's no handling of unicode actually.

@madmaxoft It is related to unicode. Perhaps I should have been more specific, e.g. "Prevent trimming any non-whitespace unicode chars"

worktycho:
Only its not Unicode. It's utf-8. Feed it utf-7,utf-16, utf-32, or ucs-2, and it will corrupt the string. It also won't work for anything not ascii based, like EBCDIC.

madmaxoft:
Let's keep it as-is for now.

LogicParrot:
Sorry, I used Unicode and utf-8 interchangebly. I meant utf-8 whenever I said unicode. 
Doesn't the server always create default configs with utf-8/ascii? Which means the rest of the encodings are unsupported anyways? Does std::string play os-specific encoding tricks?

Issue #3018, opened 2016-02-17
Destroyed ice block doesn't create water


Seadragon91:
Have tried to find the problem, no success...

Seadragon91:
Okay the problem is [here](https://github.com/cuberite/cuberite/blob/f4049d8eaa2860fd583ab5965fa840cf510ccb91/src/ClientHandle.cpp#L1247L1249).

This happens: ice block destroyed -> water placed -> replaced by air

Would it be okay to add a return value to `OnDestroyedByPlayer` and then check with it, if the block should be replaced by air?

Seadragon91:
Would it be enough to add change this line
`World->DigBlock(a_BlockX, a_BlockY, a_BlockZ);`
to
```
if (a_OldBlock != E_BLOCK_ICE)
{
	World->DigBlock(a_BlockX, a_BlockY, a_BlockZ);
}
```

PR #3020, opened 2016-02-18
Don't remove water after ice block has been destroyed

Seadragon91:
This would fix issue #3018. The question is now, if this is good enough or not.


SphinxC0re:
Looks good to me. And well documented as well :wink:

SphinxC0re:
Merge if CI succeeds

SphinxC0re:
This is okay for now but when more of these 'special cases' come up, we have to think about a better way to handle them

madmaxoft:
I'm not too happy about block-exceptions being stuffed into `cClientHandle`. Ideally, the client handle would know nothing about blocks and everything would be handled by the handlers. This points at the blockhandlers having the wrong interface.

SphinxC0re:
I totally agree with you, but for now, this fixed a real issue for the survival game. So we should start worry about this when more of these *special cases* pop up

Seadragon91:
@madmaxoft Would the idea to add a boolean value as return in `OnDestroyedByPlayer` fit for you?

NiLSPACE:
@Seadragon91 That was exactly what I had in mind.

Seadragon91:
bump
@madmaxoft return value okay for you?

madmaxoft:
How about the digging be moved into the `OnBlockDestroyed()` handler itself?

Issue #3011, opened 2016-02-16
General OnBlockBreaking lua callback


flx5:
Hey,
I wanted to write a plugin that would stop creepers from destroying the world, yet hurt the entities.

Sadly I had to find out that there is no hook acceptable for doing this.

Obviously changing the flag in the sourcecode would do the job:

[src/ChunkMap.cpp#L1774](https://github.com/cuberite/cuberite/blob/7d09fcbd2f87857abad4ea9fbb4f386aa0b44d89/src/ChunkMap.cpp#L1774)

As far as I am concerned the lazy solution would be to add a hook that influences that flag.
Yet that wouldn't be a very generic approach which is why I'd suggest to handle block destruction centrally. (Probably a method in the block class with a signature like Destroy(int dropRateInPercent))
There is already a PLAYER_BREAKING_BLOCK hook, but I'd suggest adding a general ON_BLOCK_BREAKING hook.

LogicParrot:
Relevant plugin request: https://forum.cuberite.org/thread-2323.html

LogicParrot:
Perhaps this should be configurable in `world.ini`.

```
AllowExplosionBlockDamage=true ; If false, rest of the options are irrelevant
AllowCreeperBlockDamage=true
AllowTNTBlockDamage=true
...etc...
```

jammet:
That would be a nice thing to have. :)

madmaxoft:
I don't think having a hook for each broken block is the best idea. In case of large explosions, the plugins would become the bottleneck for the server performance.

What might be a better solution for this specific case is an explosion hook that had a list of all the blocks it would be destroying as a (modifiable) parameter. That way plugins could either stop the entire damage, or remove some specific blocks from the damage (such as area protection) etc.

jammet:
Would it be possible to simplify such a thing, by ignoring all structural damage in the affected chunk entirely, when explosions go off?

madmaxoft:
@jammet that wouldn't work too well, what if the explosion occurs on a chunk border, which chunk gets the protection and which one doesn't? And why granularise on chunks when we can have any granularity we want? I think the plugin-editable list of destroyed blocks is a better approach. Of course, a plugin could then do a simple thing such as
```
function OnExploding(..., a_DestroyedBlocks)
  a_DestroyedBlocks:ClearAll()
  return true
end
```
and nothing will be destroyed by the server.

flx5:
@madmaxoft You're right, that would probably be a better solution.
[...]

PR #3022, opened 2016-02-18
Create hook for explosion breaking blocks

flx5:
Hey, this should fix the issue  #3011.

I could not find out how documentation for lua is generated.

It would be nice to get a hint for that :) 

Example usage:

```
function Initialize(Plugin)
	Plugin:SetName("Example")
	Plugin:SetVersion(1)

	cPluginManager.AddHook(cPluginManager.HOOK_EXPLOSION_BREAKING_BLOCKS, OnExplosion)

	LOG("Initialised " .. Plugin:GetName() .. " v." .. Plugin:GetVersion())
	return true
end


function OnExplosion(World, DestroyedBlocks, X, Y, Z, Source, SourceData)
	if Source == esMonster then
		DestroyedBlocks:Clear()
	end
	return true
end

```

### DIFF
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -575,6 +575,18 @@ void cLuaState::Push(const cLuaState::cRef & a_Ref)
 
 
 
+void cLuaState::Push(const cVector3Container<int> & a_VectorArray)
+{
+	ASSERT(IsValid());
+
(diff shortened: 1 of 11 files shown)


NiLSPACE:
I think it's better to simply add the needed parameter to the OnExplosion hook.

madmaxoft:
This is not a good PR, its title is nondescriptive and the implementation doesn't have the features we need.

flx5:
@NiLSPACE  You're right, I thought adding a parameter would break existing plugins. Testing proved otherwise.
 @madmaxoft Sorry, been in a hurry. Renamed + Closed

PR #3029, opened 2016-02-19
Added BACKERS to list of files in Windows build

mathias-github:


tigerw:
I feel strongly that only the bare minimum required to bring the server up to normal functionality should be included. Therefore there should be *reductions* in what is bundled in the install package, rather than additions.

Who will be the first to bring up the licensing? :P

mathias-github:
On the Bountysource page, the rewards section says that this file will be distributed with the binaries (it's currently included in the binaries for other OS:es). If people are against this, we should modify the text in the rewards section.

madmaxoft:
I actually think it's only fair to add the mentions to the distributions. And there's no questions about the licenses, we are using the libraries, so the least we could do to show our appreciation is to include their license files.
It's not like those files blow up the size considerably, they're text files, after all.

bearbin:
I don't know why the CIs didn't start, but it shouldn't have failed :P

tigerw:
Where then, would you put this file if the server were to be installed? In the user's Documents folder? Shown at startup? Shouldn't those who wish to know consult the repository?

PR #3031, opened 2016-02-20
Refactor HTTP parsing

madmaxoft:
HTTP parsing has been refactored into a stream parser, so that it is exportable to Lua API, as well as usable in the WebAdmin.
As a part of the rewrite, the HTTP server also accepts `chunked` transfer encoding now.
Added some basic unit tests for the parser.


src/HTTP/EnvelopeParser.cpp
```diff
@@ -28,12 +28,12 @@ size_t cEnvelopeParser::Parse(const char * a_Data, size_t a_Size)
 	}
 
 	// Start searching 1 char from the end of the already received data, if available:
-	size_t SearchStart = m_IncomingData.size();
-	SearchStart = (SearchStart > 1) ? SearchStart - 1 : 0;
+	auto searchStart = m_IncomingData.size();
```
tigerw: eh. *Java* style names. I mean, it's meant to be verbNoun as in `getThing()` or `resetObject` but it doesn't even *work* for everything. Why is the 'search' part of 'searchStart' so special?

And why did we change style conventions with a couple thousand lines of code already written? :P

tigerw:
What level of bug free guarantee can you provide with this again and how does that compare with a pre-existing library? :P

madmaxoft:
I provide a "works on my computer" certification, and there are unit tests to verify that the parser works against exemplary requests and responses.

Comparison with pre-existing code is that this is more modular, more independent and will be available for API export to Lua, as well as reuse in other parts. The previous code was hard-coded into the webadmin.

tigerw:
Just couldn't help but wonder whether a pre-existing library could have been used.

This can be merged as it stands right? It would be a shame to lose all of those nicely named and capitalised commits.

madmaxoft:
I haven't found a library that would follow our usage pattern - feed data to it in bits and have it call functions based on the parsed data. If you have found any, feel free to share the knowledge.

Issue #3030, opened 2016-02-20
After I join, the server crashes


Arken2013:
LOG:
```
     [11:48:30] --- Started Log ---
     [11:48:30] Cuberite  build id: 254
     [11:48:30] from commit id: 4495dd62b90e8f3a405ae7d9f464a574831aac91 built at: 254
     [11:48:30] Creating new server instance...
     [11:48:30] Reading server config...
Warn [11:48:30] Regenerating settings.ini, all settings will be reset
     [11:48:30] Starting server...
Info [11:48:30] Compatible clients: 1.7.x, 1.8.x
Info [11:48:30] Compatible protocol versions 4, 5, 47
Warn [11:48:30] Regenerating webadmin.ini, all settings will be reset
Warn [11:48:30] WebServer: The server is running in unsecured HTTP mode.
Info [11:48:30] Put a valid HTTPS certificate in file 'webadmin/httpscert.crt' and its corresponding private key to 'webadmin/httpskey.pem' (without any password) to enable HTTPS support
Info [11:48:31] There are no ranks, migrating old-style INI files to new DB ranks...
Info [11:48:31] (This might take a while)
Info [11:48:31] Rank migration failed, creating default ranks...
Info [11:48:31] Default ranks created.
     [11:48:31] Loaded 449 crafting recipes
     [11:48:31] Loaded 20 furnace recipes and 56 fuels
     [11:48:31] Loaded 88 brewing recipes
     [11:48:31] -- Loading Plugins --
     [11:48:31] Initialized ChatLog v.3
     [11:48:31] Initialised Core v.15
     [11:48:31] Initialized TransAPI v.1
     [11:48:31] -- Loaded 3 Plugins --
Info [11:48:31] Chosen a new random seed for world: 644403895
     [11:48:32] Preparing spawn (world): 2.75% (11/400; 10.87 chunks / sec)
     [11:48:33] Chunk generator performance: 41.29 ch / sec (83 ch total)
     [11:48:33] Preparing spawn (world): 8.00% (32/400; 20.98 chunks / sec)
     [11:48:34] Preparing spawn (world): 16.75% (67/400; 34.86 chunks / sec)
     [11:48:35] Preparing spawn (world): 26.00% (104/400; 37.00 chunks / sec)
     [11:48:35] Chunk generator performance: 40.57 ch / sec (163 ch total)
     [11:48:36] Preparing spawn (world): 35.75% (143/400; 38.96 chunks / sec)
[...]

PR #3035, opened 2016-02-20
Hotfix Add check if entity is in chunk before changing tick status

marvinkopf:
World tick would set the tick status of an entity to true after TRYING to add it to a chunk, this would lead to a failed assertion because a entity without a parentchunk was ticking.

Only set the tick status to true if the entity was added successful. 
<s>Remove IsTicking Assert before setting the tick status, this assertion should be (and is) done inside SetIsTicking.</s>

See my last comment in #3030

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -1019,7 +1019,10 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
 			(*itr)->SetWorld(this);
 			m_ChunkMap->AddEntity(*itr);
 			ASSERT(!(*itr)->IsTicking());
-			(*itr)->SetIsTicking(true);
+			if (m_ChunkMap->HasEntity((*itr)->GetUniqueID()))
+			{
+				(*itr)->SetIsTicking(true);
+			}
 		}
 		m_EntitiesToAdd.clear();
 	}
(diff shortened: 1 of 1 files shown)


tigerw:
In what circumstances would HasEntity return false after a call to AddEntity? I presume if the entity goes into an unloaded chunk or something?

tigerw:
Anyway, can't hurt. Thanks for triage :)

marvinkopf:
Yes, if the entity goes into an unloaded chunk.

LogicParrot:
Thanks for the hotfix, this seems to fix a crash. But I think this should be investigated further.

LogicParrot:
Some self-notes and thought-sharing:
 - How should we handle entities being added into unloaded chunks? Previously it'd just print a warning:
   `spawning in a non-existent chunk, the entity is lost.`. This was basically an ignored memory leak.
   After my PR, lots of asserts were added, which made this more pedantic, which caused an assert failure at
   at `cEntity::SetIsTicking()`. @marvinkopf 's PR prevents the assert failure (thanks!), but the root 
   problem is adding entities into unloaded chunks.
 - What called cWorld::addEntity with an invalid chunk in the first place? Is it the mob spawn code 
   (@marvinkopf's crash was probably caused by a bat).
 - Can this happen if players teleport to worlds with unloaded chunks under some cirumstances?
   What happens if we lose the player entity?
 - Possible resolutions:
    - If a chunk is unloaded, load it and then add the entity
    - Prevent all code paths which try to add entities to unloaded chunks (and bring back the assert failure)
    - other?

tigerw:
The player entity is (barring bugs) never lost, with the exception if when no memory is available to allocate chunks.

PR #3036, opened 2016-02-21
Fix false positives for minecart entity collision

marvinkopf:
Minecarts would only look for one (the first found) entity in their bounding box to handle a collision, but that entity could have been a false positive / in the wrong direction of the minecart. A minecart would ignore entities in front of it while the player was pushing it.

### DIFF
--- a/src/Entities/Minecart.cpp
+++ b/src/Entities/Minecart.cpp
@@ -25,8 +25,7 @@ class cMinecartCollisionCallback :
 {
 public:
 	cMinecartCollisionCallback(Vector3d a_Pos, double a_Height, double a_Width, UInt32 a_UniqueID, UInt32 a_AttacheeUniqueID) :
-		m_DoesInteserct(false),
-		m_CollidedEntityPos(0, 0, 0),
+		m_DoesIntersect(false),
 		m_Pos(a_Pos),
 		m_Height(a_Height),
 		m_Width(a_Width),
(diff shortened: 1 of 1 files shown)


marvinkopf:
added more comments and renamed some variables to make it a bit more clear that an entity inside the minecarts bounding box can but not must have a collision with the minecart and that there can be multiple entities of which exactly only one or none have a collision with the minecart.

NiLSPACE:
@tigerw you did minecart physics right? Could you review these changes?

SphinxC0re:
I'm not sure this is the right approach to the problem. The code is good but it would be much nicer to have a physics engine to handle the entity collision for minecarts and boats and essentially every other entity in the world

LogicParrot:
@tigerw ping.

LogicParrot:
@SphinxC0re The mechanics of boats, minecarts, horses, and mob knockback are all very different. I don't know if an engine would be of much use. It would still internally contain seperate code for each, wouldn't it?

tigerw:
> The mechanics of boats, minecarts, horses, and mob knockback

Knockback?

I can envision the current collision system for mobile entities (trace a line until hit, then move there) being applied to all entities. It would need to be adapted to look for entities too, for cart-cart collision, but I guess that base can be considered the engine.

Heh, engine. It really has nothing to do with engines.

LogicParrot:
Knockback as in e.g. arrows hitting players and applying some force to them.

marvinkopf:
I can't figure out any more what history state this PR was in when I tested it, and it doesn't work now. Also, there are some obvious mistakes in the code. Maybe I did a wrong force push and forgot to test it again at some point. 

At some point there was a bug like this in my history and this PR did have a solution for it. But right now this patch does not make that much sense ~closed

LogicParrot:
@marvinkopf Why was this closed?

PR #3037, opened 2016-02-21
Add speed limit enforcement for minecarts

### LINKED ISSUE #3000: Limit minecart max speed and powered rails not reducing speed
Minecarts are supposed not to go beyond 8 block per second even when being powered, right now it has no limit(or at least it seems like it) and every powered rail you put increases the minecart speed even more.

I mean is pretty funny like it is i've made a human rocket with a minecart, but the main side effect is that you end inside a wall, the minecart fly over the rail and when taking corners step out of the rails.

That plus powered rails not working also when are are not powered, make the minecart use pretty hard, when not powered the rails are supposed to reduce the speed to almost stop the minecart in one block. 

I mention this two things together because I think may be related to one another.

marvinkopf:
fixes #3000

### DIFF
--- a/src/Entities/Minecart.cpp
+++ b/src/Entities/Minecart.cpp
@@ -277,11 +277,8 @@ void cMinecart::HandleRailPhysics(NIBBLETYPE a_RailMeta, std::chrono::millisecon
 			if (GetSpeedZ() >= 0)
 			{
 				// SpeedZ POSITIVE, going SOUTH
-				if (GetSpeedZ() <= MAX_SPEED)  // Speed limit
-				{
-					AddSpeedZ(0.5);  // Speed up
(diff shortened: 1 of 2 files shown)


LogicParrot:
Can I trust you on this one? If so, rebase and let's merge.

LogicParrot:
Can I trust you tested this one? If so, rebase and let's merge.

SphinxC0re:
Sounds good

tigerw:
I think Mr Kopf may be a little miffed at the timing of this.

marvinkopf:
Ready to merge :-)

LogicParrot:
>I think Mr Kopf may be a little miffed at the timing of this.

Better late than never!

rekuen:
Could this PR be the reason of #3147 ?

marvinkopf:
That's very unlikely.

PR #3043, opened 2016-02-22
[WIP/RFC] Configurable block destruction for explosions

### LINKED ISSUE #3011: General OnBlockBreaking lua callback

Woazboat:
Added CanDestroyBlocks and CanDamageEntities parameters to the DoExplosionAt function, Lua bindings and plugin hooks to allow blocking creeper explosion damage, etc.... Breaks the plugin API because of the added parameters.

This is just a request for comments and is not really the final pull request. Are the changes I made okay or did I miss something important? As it is now it runs fine on my machine, although I'm not entirely sure if the Lua bindings and hooks are correct.

Fixes #3011

### DIFF
--- a/src/Bindings/ManualBindings_World.cpp
+++ b/src/Bindings/ManualBindings_World.cpp
@@ -114,16 +114,16 @@ static int tolua_cWorld_ChunkStay(lua_State * tolua_S)
 static int tolua_cWorld_DoExplosionAt(lua_State * tolua_S)
 {
 	/* Function signature:
-	World:DoExplosionAt(ExplosionSize, BlockX, BlockY, BlockZ, CanCauseFire, SourceType, [SourceData])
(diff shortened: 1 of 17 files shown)


madmaxoft:
I don't think this is the right way to do this. As I already said, it would be much better to have a list of affected blocks and a way to manipulate that list easily (such as clearing all items, or removing all specific blocktypes, removing items in a specific coords range etc.)

Issue #3054, opened 2016-02-28
cPlayer: Using TeleportToCoords inflicts damage


Fighter19:
Looking at https://github.com/cuberite/Essentials/issues/53
I found that TeleportToCoords inflicts damage while the player is moving.

Schwertspize:
you mean, if you teleport to a specific coordinate, you should ignore fall damage, or what else?

Fighter19:
Yes. Although it's not only "fall damage" for example if you teleport up while moving you get damage. (Also if moving horizontally only)

Schwertspize:
well, ok, this must definitely be fixed.... would you mind posting the "death reason", that would probably make fixing easier

Anyway, thanks

Fighter19:
I hope this is what you meant:
"[name] fell to death"

Fighter19:
Maybe make m_bIsTeleporting available and add something like into line 376 in Pawn.cpp ?:
FallDamageAbsorbed |= IsFlying || Player->IsClimbing() || Player->IsTeleporting();

PR #3056, opened 2016-02-28
Fixed teleporting dealing damage.

Fighter19:
This solves the issue https://github.com/cuberite/cuberite/issues/3054
by checking whether the current speed is 0 before setting IsTeleporting/m_bIsTeleporting to false.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -761,6 +761,13 @@ void cClientHandle::HandlePlayerPos(double a_PosX, double a_PosY, double a_PosZ,
 		return;
 	}
 
+	if (m_Player->IsTeleporting())
+	{
+		m_Player->SetIsTeleporting(false);
+		// Skip this once for Teleporting
+		return;
+	}
+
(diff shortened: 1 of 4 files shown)


Fighter19:
Is currently broken.......

Fighter19:
Concern: Disables HandlePlayerPos for the tick after the teleport.

Fighter19:
Cleaned it up (merged multiple commits into one, as I wasn't expecting the checks to fail)

Fighter19:
Well, this doesn't really fix it for all cases but only makes it less common. For example if you run before clicking on the sign, it will still kill you / inflict damage so another approach has to be found.

bearbin:
The travis CI build failed: 

You need to document the new functions you created in APIDump, since they are exported to the Lua API.

The documentation is stored in Server/Plugins/APIDump/

Fighter19:
I know. However this doesn't even fix the problem, so I see no need it fixing the documentation.

PR #3057, opened 2016-02-29
Mobs no longer spawn at the top of the nether.

### LINKED ISSUE #3055: Entities spawning on nether above Y 128 (nether's top), and mushrooms too.
It's just that I've seen some ghasts and pigman over there.
There are also mushrooms.

Shouldn't there be no monsters or mushrooms there?

tonibm19:
I don't really know if this is the right place for the check, but it
works.
Fixes #3055

### DIFF
--- a/src/MobSpawner.cpp
+++ b/src/MobSpawner.cpp
@@ -134,6 +134,11 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
 		return false;
 	}
 
+	if (cChunkDef::IsValidHeight(a_RelY - 1) && (a_Chunk->GetBlock(a_RelX, a_RelY - 1, a_RelZ) == E_BLOCK_BEDROCK))
+	{
+		return false;   // Make sure mobs do not spawn on bedrock.
+	}
+
 	cFastRandom Random;
 	BLOCKTYPE TargetBlock = a_Chunk->GetBlock(a_RelX, a_RelY, a_RelZ);
 


madmaxoft:
Actually in vanilla mobs do spawn above 127, there are several farm designs based on that:
https://www.youtube.com/watch?v=6W8YgdRnHy8
https://www.youtube.com/watch?v=fPTA6gBoVO8

tonibm19:
what about this wat?

madmaxoft:
I think we should instead focus on mobs not spawning on top of bedrock (anywhere) and that should get us consistent with Vanilla - if someone does get above the bedrock ceiling in the Nether, they can start a farm just like in those videos, simply by covering the bedrock with other blocks.

NiLSPACE:
I believe mobs don't spawn on bedrock everywhere, so it could be replaced with this:
```cpp
if (a_Chunk->GetBlock(a_RelX, a_RelY - 1, a_RelZ) == E_BLOCK_BEDROCK))
{
	return false;   // Make sure mobs do not spawn on bedrock.
}
```

madmaxoft:
Don't forget the Y bounds check

tonibm19:
It seems to work with @NiLSPACE code.
What Y bound check are you talking about? Height is already checked.

madmaxoft:
I don't think the `IsValidHeight()` check was there at first, was it? Or am I blind? :P

NiLSPACE:
No, I changed it afterwards. Forgot to comment that I changed it, sorry.

madmaxoft:
Remove the unrelated horse code modifications, squash into a single commit and we can merge.

tonibm19:
Why remove the horse code? It works.
Anyway, if you want that, i'll remove.

madmaxoft:
Because we don't want unrelated changes in a single PR, let alone a single commit.

NiLSPACE:
Great, can you squash all the commits?

tonibm19:
I'm trying it but... problems.
I've never squashed before D:

tonibm19:
I think I got it

SphinxC0re:
Uhh, a little squash-virgin :laughing: how did it feel?

tonibm19:
weird

SphinxC0re:
Don't worry it will get better over time :smile:

madmaxoft:
Now you can make that horse change, in a separate PR :)

PR #3068, opened 2016-03-03
WebAdmin: Finish HTTP responses.

madmaxoft:
Fixes a problem that the HTML pages were never terminated, and, on some browsers, failed to display.

Ref.: https://forum.cuberite.org/thread-2392.html


SphinxC0re:
Looks good

PR #3069, opened 2016-03-03
HTTP: Fixed header parsing.

madmaxoft:
Now forms are parsed correctly.

Hopefully fixes https://forum.cuberite.org/thread-2392.html for good.


Schwertspize:
working. (confirmed)

we can merge that i think

PR #3072, opened 2016-03-05
Implemented OverworldClumpFlowers

NiLSPACE:
I should probably add a few comments, but the overal code is done.

### DIFF
--- a/src/Generating/ComposableGenerator.cpp
+++ b/src/Generating/ComposableGenerator.cpp
@@ -438,6 +438,11 @@ void cComposableGenerator::InitFinishGens(cIniFile & a_IniFile)
 				m_FinishGens.push_back(gen);
 			}
 		}
+		else if (NoCaseCompare(finisher, "OverworldClumpFlowers") == 0)
+		{
+			auto flowers = cFinishGenClumpTopBlock::ParseIniFile(a_IniFile, "OverworldClumpFlowers");
+			m_FinishGens.push_back(cFinishGenPtr(new cFinishGenClumpTopBlock(Seed, flowers)));
+		}
 		else if (NoCaseCompare(finisher, "PieceStructures") == 0)
 		{
 			if (split.size() < 2)
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -180,6 +180,210 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
 
 
 ////////////////////////////////////////////////////////////////////////////////
+// cFinishGenClumpTopBlock
+
+void cFinishGenClumpTopBlock::GenFinish(cChunkDesc & a_ChunkDesc)
+{
+	int ChunkX = a_ChunkDesc.GetChunkX();
+	int ChunkZ = a_ChunkDesc.GetChunkZ();
+
+	int NoiseVal = m_Noise.IntNoise2DInt(ChunkX, ChunkZ);
+	EMCSBiome Biome = a_ChunkDesc.GetBiome(cChunkDef::Width / 2, cChunkDef::Width / 2);
+	BiomeInfo info = m_FlowersPerBiome[static_cast<size_t>(Biome)];
+
+	const auto & PossibleBlocks = info.m_Blocks;
+	if (PossibleBlocks.empty())
+	{
+		// No need to go any further. This biome can't generate any blocks.
+		return;
+	}
+
+	int NumClumps = info.m_MaxNumClumpsPerChunk - info.m_MinNumClumpsPerChunk;
+	if (NumClumps == 0)
+	{
+		NumClumps = 1;
+	}
+
+	NumClumps = NoiseVal % NumClumps + info.m_MinNumClumpsPerChunk;
+	for (int i = 0; i < NumClumps; i++)
+	{
+		int Val1 = m_Noise.IntNoise2DInt(ChunkX * ChunkZ * i, ChunkZ + ChunkX + i);
+		int Val2 = m_Noise.IntNoise2DInt(ChunkZ * ChunkX + i, ChunkZ - ChunkX * i);
+		int BlockVal = m_Noise.IntNoise2DInt(Val1, Val2);
+
+		int PosX = Val1 % (cChunkDef::Width - RANGE_FROM_CENTER * 2) + RANGE_FROM_CENTER;
+		int PosZ = Val2 % (cChunkDef::Width - RANGE_FROM_CENTER * 2) + RANGE_FROM_CENTER;
+
+		int TotalWeight = 0;
+		for (const auto & Block : PossibleBlocks)
+		{
+			TotalWeight += Block.m_Weight;
+		}
+
+		// Prevent division by 0
+		TotalWeight = (TotalWeight != 0) ? TotalWeight : 1;
+		int Weight = BlockVal % TotalWeight;
+		for (const auto & Block : PossibleBlocks)
+		{
+			Weight -= Block.m_Weight;
+			if (Weight < 0)
+			{
+				TryPlaceFoliageClump(a_ChunkDesc, PosX, PosZ, Block.m_BlockType, Block.m_BlockMeta, Block.m_BlockType == E_BLOCK_BIG_FLOWER);
+				break;
+			}
+		}
+	}
+}
+
+
+
+
+
+void cFinishGenClumpTopBlock::TryPlaceFoliageClump(cChunkDesc & a_ChunkDesc, int a_CenterX, int a_CenterZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, bool a_IsDoubleTall)
+{
+	int ChunkX = a_ChunkDesc.GetChunkX();
+	int ChunkZ = a_ChunkDesc.GetChunkZ();
+
+	int NumBlocks = m_Noise.IntNoise2DInt(a_CenterX + ChunkX * 16, a_CenterZ + ChunkZ * 16) % (MAX_NUM_FOLIAGE - MIN_NUM_FOLIAGE) + MIN_NUM_FOLIAGE + 1;
+	for (int i = 1; i < NumBlocks; i++)
+	{
+		int rnd = m_Noise.IntNoise2DInt(ChunkX + ChunkZ + i, ChunkX - ChunkZ - i) / 59;
+		int x = a_CenterX + (((rnd % 256) % RANGE_FROM_CENTER * 2) - RANGE_FROM_CENTER);
+		int z = a_CenterZ + (((rnd / 256) % RANGE_FROM_CENTER * 2) - RANGE_FROM_CENTER);
+		int Top = a_ChunkDesc.GetHeight(x, z);
+
+		if (a_ChunkDesc.GetBlockType(x, Top, z) != E_BLOCK_GRASS)
+		{
+			continue;
+		}
+
+		a_ChunkDesc.SetBlockTypeMeta(x, Top + 1, z, a_BlockType, a_BlockMeta);
+		if (a_IsDoubleTall)
+		{
+			a_ChunkDesc.SetBlockTypeMeta(x, Top + 2, z, E_BLOCK_BIG_FLOWER, 8);
+		}
+	}
+
+}
+
+
+
+
+
+void cFinishGenClumpTopBlock::ParseConfigurationString(AString a_RawClumpInfo, std::vector<BiomeInfo> & a_Output)
+{
+	// Initialize the vector for all biomes.
+	for (int i = static_cast<int>(a_Output.size()); i < static_cast<int>(biMaxVariantBiome); i++)
+	{
+		a_Output.push_back(BiomeInfo());
+	}
+
+	AStringVector ClumpInfo = StringSplitAndTrim(a_RawClumpInfo, "=");
+
+	// Information about a clump is divided in 2 parts. The biomes they can be in and the blocks that can be placed.
+	if (ClumpInfo.size() != 2)
+	{
+		LOGWARNING("OverworldClumpFoliage: Data missing for \"%s\". Please divide biome and blocks with a semi colon", a_RawClumpInfo.c_str());
+		return;
+	}
+
+	AStringVector Biomes = StringSplitAndTrim(ClumpInfo[0], ";");
+	AStringVector Blocks = StringSplitAndTrim(ClumpInfo[1], ";");
+
+	for (const auto & RawBiomeInfo : Biomes)
+	{
+		AStringVector BiomeInfo = StringSplitAndTrim(RawBiomeInfo, ",");
+		AString BiomeName = BiomeInfo[0];
+		EMCSBiome Biome = StringToBiome(BiomeName);
+		if (Biome == biInvalidBiome)
+		{
+			LOGWARNING("Biome \"%s\" is invalid.", BiomeName.c_str());
+			continue;
+		}
+
+		if (BiomeInfo.size() == 2)
+		{
+			// Only the minimum amount of clumps per chunk is changed.
+			int MinNumClump = 1;
+			if (!StringToInteger(BiomeInfo[1], MinNumClump))
+			{
+				LOGWARNING("OverworldClumpFoliage: Invalid data in \"%s\". Second parameter is either not existing or a number", RawBiomeInfo.c_str());
+				continue;
+			}
+			a_Output[static_cast<size_t>(Biome)].m_MinNumClumpsPerChunk = MinNumClump;
+
+			// In case the minimum number is higher than the current maximum value we change the max to the minimum value.
+			a_Output[static_cast<size_t>(Biome)].m_MaxNumClumpsPerChunk = std::max(MinNumClump, a_Output[static_cast<size_t>(Biome)].m_MaxNumClumpsPerChunk);
+		}
+		else if (BiomeInfo.size() == 3)
+		{
+			// Both the minimum and maximum amount of clumps per chunk is changed.
+			int MinNumClumps = 0, MaxNumClumps = 1;
+			if (!StringToInteger(BiomeInfo[1], MinNumClumps) || !StringToInteger(BiomeInfo[2], MaxNumClumps))
+			{
+				LOGWARNING("Invalid data in \"%s\". Second parameter is either not existing or a number", RawBiomeInfo.c_str());
+				continue;
+			}
+
+			a_Output[static_cast<size_t>(Biome)].m_MaxNumClumpsPerChunk = MaxNumClumps + 1;
+			a_Output[static_cast<size_t>(Biome)].m_MinNumClumpsPerChunk = MinNumClumps;
+		}
+
+		// TODO: Make the weight configurable.
+		for (const auto & BlockName : Blocks)
+		{
+			cItem Block = cItem();
+			if (!StringToItem(BlockName, Block) && IsValidBlock(Block.m_ItemType))
+			{
+				LOGWARNING("Block \"%s\" is invalid", BlockName.c_str());
+				continue;
+			}
+
+			FoliageInfo info = FoliageInfo(static_cast<BLOCKTYPE>(Block.m_ItemType), static_cast<NIBBLETYPE>(Block.m_ItemDamage), 100);
+			a_Output[static_cast<size_t>(Biome)].m_Blocks.push_back(info);
+		}
+	}
+}
+
+
+
+
+
+std::vector<cFinishGenClumpTopBlock::BiomeInfo> cFinishGenClumpTopBlock::ParseIniFile(cIniFile & a_IniFile, AString a_ClumpPrefix)
+{
+	// Also check dashes in case we will get more configuration options with the same prefix.
+	a_ClumpPrefix += "-";
+
+	std::vector<cFinishGenClumpTopBlock::BiomeInfo> foliage;
+	int NumGeneratorValues = a_IniFile.GetNumValues("Generator");
+	int GeneratorKeyId = a_IniFile.FindKey("Generator");
+	for (int i = 0; i < NumGeneratorValues; i++)
+	{
+		AString ValueName = a_IniFile.GetValueName("Generator", i);
+		if (ValueName.substr(0, a_ClumpPrefix.size()) == a_ClumpPrefix)
+		{
+			AString RawClump = a_IniFile.GetValue(GeneratorKeyId, i);
+			cFinishGenClumpTopBlock::ParseConfigurationString(RawClump, foliage);
+		}
+	}
+
+	if (foliage.size() == 0)
+	{
+		cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", a_ClumpPrefix + "-1", "Forest, -2, 2; ForestHills, -3, 2; FlowerForest = yellowflower, redflower, lilac, rosebush"), foliage);
+		cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", a_ClumpPrefix + "-2", "Plains, -2, 1; SunflowerPlains = yellowflower, redflower, azurebluet, oxeyedaisy"), foliage);
+		cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", a_ClumpPrefix + "-3", "SunflowerPlains, 1, 2 = sunflower"), foliage);
+		cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", a_ClumpPrefix + "-4", "FlowerForest, 2, 5 = allium, redtulip, orangetulip, whitetulip, pinktulip, oxeyedaisy"), foliage);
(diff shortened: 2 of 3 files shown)


src/Generating/ComposableGenerator.cpp
```diff
@@ -438,6 +438,11 @@ void cComposableGenerator::InitFinishGens(cIniFile & a_IniFile)
 				m_FinishGens.push_back(gen);
 			}
 		}
+		else if (NoCaseCompare(finisher, "OverworldClumpFlowers") == 0)
+		{
+			AString flowers = a_IniFile.GetValueSet("Generator", "OverworldClumpFlowers", "Forest+-2+2,ForestHills+-3+2,FlowerForest; yellowflower,redflower,lilac,rosebush | plains+-2+1,SunflowerPlains; yellowflower,redflower,azurebluet,oxeyedaisy | SunflowerPlains+1+2; sunflower | FlowerForest+2+5; allium,redtulip,orangetulip,whitetulip,pinktulip,oxeyedaisy | Swampland,SwamplandM; brownmushroom,redmushroom,blueorchid");
```
NiLSPACE: CheckBasicStyle doesn't like the format I'm using here with the spaces and plus signs. Any idea what I could use instead?

tigerw: @madmaxoft again, the style checker really shouldn't look at comments or string literals.

But, my God is that a horrible configuration string. I mean, you might as well not have one considering nobody is likely to touch that. :P

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,187 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+		int Val2 = m_Noise.IntNoise2DInt(ChunkZ * ChunkX / seq, ChunkZ - ChunkX * seq);
+
+		int PosX = Val1 % (cChunkDef::Width - RANGE_FROM_CENTER * 2) + RANGE_FROM_CENTER;
+		int PosZ = Val2 % (cChunkDef::Width - RANGE_FROM_CENTER * 2) + RANGE_FROM_CENTER;
+
+		int BlockVal = m_Noise.IntNoise2DInt(Val1, Val2);
+
+
+		int TotalWeight = 0;
+		for (auto Block : PossibleBlocks)
```
tigerw: You wish to make copies of BlockInfo?

src/Generating/ComposableGenerator.cpp
NiLSPACE: I was thinking on how to split it up, but I have no idea how. :(

But I think it's really hard to determine if something is a comment or not. What if the comment started on a different line?

src/Generating/FinishGen.cpp
NiLSPACE: I'm not quite following you :|

Seadragon91: `auto Block` -> `auto & Block`

NiLSPACE: Oh, thanks :)

madmaxoft: Perhaps even a `const auto &`, since you're not modifying the object at all.

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,71 @@ class cFinishGenNetherClumpFoliage :
...
+	public cFinishGen
+{
+public:
+	// Contains the meta, type and weight for a clump block
+	struct BlockInfo
+	{
+		BLOCKTYPE m_BlockType;
+		NIBBLETYPE m_BlockMeta;
+		int m_Weight;
+		BlockInfo(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, int a_Weight) :
```
madmaxoft: Style: separate functions from data members by an empty line

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,71 @@ class cFinishGenNetherClumpFoliage :
...
+			m_Blocks(a_Blocks)
+		{}
+	};
+
+
+	cFinishGenClumpTopBlock(int a_Seed, std::vector<BiomeInfo> a_BlockList) :
+		m_Noise(a_Seed),
+		m_FlowersPerBiome(a_BlockList)
+	{
+		m_FlowersPerBiome = a_BlockList;
```
madmaxoft: Two copy-assignments?
Use move semantics for `a_BlockList` to avoid so much copying.

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,71 @@ class cFinishGenNetherClumpFoliage :
...
+		{}
+	};
+
+	// Contains the minimum and maximum amount of clumps for a biome and it's blocks.
+	struct BiomeInfo
+	{
+		int m_MinNumClumpsPerChunk;
+		int m_MaxNumClumpsPerChunk;
+		std::vector<BlockInfo> m_Blocks;
+		BiomeInfo() :
```
madmaxoft: Style: empty line

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,187 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+void cFinishGenClumpTopBlock::GenFinish(cChunkDesc & a_ChunkDesc)
+{
+	int ChunkX = a_ChunkDesc.GetChunkX();
+	int ChunkZ = a_ChunkDesc.GetChunkZ();
+
+	int NoiseVal = m_Noise.IntNoise2DInt(ChunkX, ChunkZ);
+	EMCSBiome Biome = a_ChunkDesc.GetBiome(cChunkDef::Width / 2, cChunkDef::Width / 2);
+	BiomeInfo info = m_FlowersPerBiome[static_cast<int>(Biome)];
+
+	std::vector<BlockInfo> PossibleBlocks = info.m_Blocks;
```
madmaxoft: `const auto &`, to avoid copying and explicit types.

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,187 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+
+	AStringVector ClumpList = StringSplitAndTrim(a_String, "|");
+	for (auto RawClumpInfo : ClumpList)
+	{
+		AStringVector ClumpInfo = StringSplitAndTrim(RawClumpInfo, ";");
+
+		// Information about a clump is divided in 2 parts. The biomes they can be in and the blocks that can be placed.
+		if (ClumpInfo.size() != 2)
+		{
+			LOGWARNING("Data missing for \"%s\". Please divide biome and blocks with a semi colon", RawClumpInfo);
```
madmaxoft: All warnings displayed to the console should indicate what part of the configuration is wrong (`OverworldClumpFoliage: ...`)
`divide ... and ...` -> `separate ... from ...`

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,187 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+			continue;
+		}
+
+		AStringVector Biomes = StringSplitAndTrim(ClumpInfo[0], ",");
+		AStringVector Blocks = StringSplitAndTrim(ClumpInfo[1], ",");
+
+		for (auto RawBiomeInfo : Biomes)
```
madmaxoft: `const auto &`

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,187 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+					LOGWARNING("Invalid data in \"%s\". Second parameter is either not existing or a number", RawBiomeInfo);
+					continue;
+				}
+
+				BiomeList[Biome].m_MaxNumClumpsPerChunk = MaxNumClumps + 1;
+				BiomeList[Biome].m_MinNumClumpsPerChunk = MinNumClumps;
+			}
+
+			// TODO: Make the weight configurable.
+			for (auto BlockName : Blocks)
```
madmaxoft: `const auto &`

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,71 @@ class cFinishGenNetherClumpFoliage :
 
 
 
+class cFinishGenClumpTopBlock :
+	public cFinishGen
+{
+public:
+	// Contains the meta, type and weight for a clump block
+	struct BlockInfo
```
madmaxoft: Perhaps choose a different name than `Block`, to avoid confusion. How about `FoliageInfo`?

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,71 @@ class cFinishGenNetherClumpFoliage :
...
+
+	cFinishGenClumpTopBlock(int a_Seed, std::vector<BiomeInfo> a_BlockList) :
+		m_Noise(a_Seed),
+		m_FlowersPerBiome(a_BlockList)
+	{
+		m_FlowersPerBiome = a_BlockList;
+	}
+
+	/** Parses a string and returns a vector with a length of biMaxVariantBiome.
+		The format of the string is "<Biomes separated with a comma>;<Blocks separated with a comma>". This can also be repeated with a | */
```
madmaxoft: Too much indent here.

src/Generating/ComposableGenerator.cpp
madmaxoft: Am I reading this right that you're setting the minima to `-2` and `-3` in the first clump? How does that work?

src/Generating/FinishGen.h
NiLSPACE: Well, at first it was all FoliageXYZ instead of Block, but then I rebuild it in a way that it can be used for other things than flowers.

madmaxoft: Still, Foliage is the primary function, so it might be worth using it as the name.

The point is, when you later reference this as `block`, it creates a lot of confusion because it's unclear whether a `block` is this foliage info, or a topmost block in the chunk, or whatever else.

tigerw: > or whatever else

Whether or not xoft has a blocked drain.

src/Generating/ComposableGenerator.cpp
NiLSPACE: It picks a random number between the minimum and maximum amount of clumps. If it's negative no clumps will generate.

NiLSPACE:
I've adressed the issues, but unfortunately Visual Studio refuses to compile after I've reïnstalled Windows. It now returns with a weird error `"cmd.exe" exited with code 3`.

src/Generating/ComposableGenerator.cpp
```diff
@@ -438,6 +438,32 @@ void cComposableGenerator::InitFinishGens(cIniFile & a_IniFile)
...
+			if (flowers.size() == 0)
+			{
+				cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", "OverworldClumpFlowers-1", "Forest+-2+2,ForestHills+-3+2,FlowerForest; yellowflower,redflower,lilac,rosebush"));
+				cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", "OverworldClumpFlowers-2", "Plains+-2+1,SunflowerPlains; yellowflower,redflower,azurebluet,oxeyedaisy"));
+				cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", "OverworldClumpFlowers-3", "SunflowerPlains+1+2; sunflower"));
+				cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", "OverworldClumpFlowers-4", "FlowerForest+2+5; allium,redtulip,orangetulip,whitetulip,pinktulip,oxeyedaisy"));
+				cFinishGenClumpTopBlock::ParseConfigurationString(a_IniFile.GetValueSet("Generator", "OverworldClumpFlowers-5", "Swampland,SwamplandM; brownmushroom,redmushroom,blueorchid"));
+			}
+			
+			m_FinishGens.push_back(cFinishGenPtr(new cFinishGenClumpTopBlock(Seed, flowers)));
```
madmaxoft: This is getting too complicated and too specific. How about wrapping it all inside a `cFinishGenClumpTopBlock::Initialize()` function?

NiLSPACE: Did you mean having the Initialize method to be called after the cFinishGenClumpTopBlock was constructed?

I now created a ParseIniFile method that looks in the inifile for the the proper value keys. It's supposed to be called before constructing the object and giving the return value to the constructor.

I'm also not sure how to do the default values. I suppose I should keep them in the ComposableGenerator.cpp file right?

madmaxoft: All the other generators that have non-trivial initialization had used the construct-initialize-use sequence, where the constructor is trivial and initialization reads from the ini file. That way, we can make the `Initialize()` function virtual in the base class and just call it on all generators after they are constructed.

I think the best place for the default values is in the generator implementation itself. If there ever is a need to edit the defaults, it usually means the need to edit the generator anyway, so that it's in one place. Searching for the defaults in a completely different file seems counter-intuitive for me.

tigerw:
> I've adressed the issues, but unfortunately Visual Studio refuses to compile after I've reïnstalled Windows. It now returns with a weird error `"cmd.exe" exited with code 3`.

Hast thou reinstallificated CMake?

madmaxoft:
Usually the `cmd exited with error code N` error message is preceded by a more detailed error message from CMake,you need to look at the build log in the Output pane, rather than just the Error list. Most likely there's some problem with CMake generator.

Schwertspize:
Style issues and type errors.... Looks like @NiLSPACE should review that another time

NiLSPACE:
I'm finally working on this again. I'm currently changing the format of the configuration strings. What should I replace the plus, comma and semocolon with? I was thinking replacing the semicolon with a pipe, the plus with a slash, but I'm not sure what to do with the comma. Anyway, with those changes it would look like this:
`Forest/-2/2,ForestHills/-3/2,FlowerForest | yellowflower,redflower,lilac,rosebush`
Or I keep the commas in place and put a space after it so the style checker won't see it.

madmaxoft:
I think the pipe should be the topmost separator, equivalent to having multiple lines. The rest could be more or less anything; if whitespace could be supported, it would be great for readability.
How about this:
`Forest, -2, 2;  ForestHills, -3, 2;  FlowerForest = yellowflower, redflower, lilac, rosebush`
So, a pipe for toplevel, `=` between biomes and blocks, semicolon between individual biomes and comma on the lowest level. Whitespace optional anywhere (`StringSplitAndTrim()` helps a lot)

NiLSPACE:
Done.

NiLSPACE:
Whatcha think?

tigerw:
I think xoft has been consumed by Factorio. I dare say you can merge yourself when Travis is happy :P

NiLSPACE:
I'd like to get xoft's consent first though :P

LogicParrot:
@tigerw That sentence should make it into the Factorio homepage. Under the title "Geeks Love it, testimonies:"

NiLSPACE:
(**bump**)

madmaxoft:
Travis is failing (and a rebase?)

src/Generating/ComposableGenerator.cpp
```diff
@@ -438,6 +438,11 @@ void cComposableGenerator::InitFinishGens(cIniFile & a_IniFile)
 				m_FinishGens.push_back(gen);
 			}
 		}
+		else if (NoCaseCompare(finisher, "OverworldClumpFlowers") == 0)
+		{
+			std::vector<cFinishGenClumpTopBlock::BiomeInfo> flowers = cFinishGenClumpTopBlock::ParseIniFile(a_IniFile, "OverworldClumpFlowers");
```
madmaxoft: `auto`?

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,210 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+		int BlockVal = m_Noise.IntNoise2DInt(Val1, Val2);
+
+
+		int TotalWeight = 0;
+		for (const auto & Block : PossibleBlocks)
+		{
+			TotalWeight += Block.m_Weight;
+		}
+
+		int Weight = BlockVal % TotalWeight;
```
madmaxoft: If `TotalWeight` == 0 (bad config?), this will crash. Add a special case, `0` -> `1`

NiLSPACE:
I suppose I can't simply press the `Update branch` button. *sigh* now I have to look up how to rebase on master :/

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,210 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+
+std::vector<cFinishGenClumpTopBlock::BiomeInfo> cFinishGenClumpTopBlock::ParseIniFile(cIniFile & a_IniFile, AString a_ClumpPrefix)
+{
+	std::vector<cFinishGenClumpTopBlock::BiomeInfo> foliage;
+	int NumGeneratorValues = a_IniFile.GetNumValues("Generator");
+	int GeneratorKeyId = a_IniFile.FindKey("Generator");
+	for (int i = 0; i < NumGeneratorValues; i++)
+	{
+		AString ValueName = a_IniFile.GetValueName("Generator", i);
+		if (ValueName.substr(0, a_ClumpPrefix.size()) == a_ClumpPrefix)
```
madmaxoft: I think the dash should be checked as well, just in case there's a new parameter in the future that starts with the prefix but does something else (`OverworldClumpFlowersMaxHeight`, for example)

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,75 @@ class cFinishGenNetherClumpFoliage :
...
+	cFinishGenClumpTopBlock(int a_Seed, std::vector<BiomeInfo> a_BlockList) :
+		m_Noise(a_Seed),
+		m_FlowersPerBiome()
+	{
+		std::swap(a_BlockList, m_FlowersPerBiome);
+	}
+
+	/** Parses a string and returns a vector with a length of biMaxVariantBiome.
+	The format of the string is "<Biomes separated with a comma>;<Blocks separated with a comma>". This can also be repeated with a | */
+	static void ParseConfigurationString(AString a_String, std::vector<BiomeInfo> & a_Output = std::vector<BiomeInfo>());
```
madmaxoft: I don't think having a default value for the output parameter is a good idea.

madmaxoft: The doxycomment for the format seems wrong, at least compared to the default values inserted in `ParseIniFile()`

src/Generating/FinishGen.h
```diff
@@ -70,6 +70,75 @@ class cFinishGenNetherClumpFoliage :
...
+	/** Parses an inifile in search for all clumps */
+	static std::vector<BiomeInfo> ParseIniFile(cIniFile & a_IniFile, AString a_ClumpPrefix);
+protected:
+
+	cNoise m_Noise;
+	std::vector<BiomeInfo> m_FlowersPerBiome;
+
+	const int MAX_NUM_FOLIAGE = 8;
+	const int MIN_NUM_FOLIAGE = 4;
+	const int RANGE_FROM_CENTER = 5;
```
madmaxoft: These deserve a doxycomment, each.

madmaxoft:
While in this PR's branch:
```
git fetch origin master:master
git rebase origin/master
```

tigerw:
'Course you can press update branch.

src/Generating/FinishGen.h
NiLSPACE: I'm really wondering why I did that. I'm not even using it.

NiLSPACE:
I think I've addressed all issues except the branch out-of-date.

src/Generating/FinishGen.cpp
NiLSPACE: I've fixed this by first adding a dash to the prefix before doing anything else.

NiLSPACE:
I've fixed clang compilation and rebased onto master.

sphinxc0re:
Looks good to me

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,215 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+
+
+		int TotalWeight = 0;
+		for (const auto & Block : PossibleBlocks)
+		{
+			TotalWeight += Block.m_Weight;
+		}
+
+		// Prevent division by 0
+		TotalWeight = TotalWeight != 0 ? TotalWeight : 1;
```
madmaxoft: Add parentheses around the comparison

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,215 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+	{
+		NumClumps = 1;
+	}
+
+	NumClumps = NoiseVal % NumClumps + info.m_MinNumClumpsPerChunk;
+	int seq = 1;
+	for (int i = 0; i < NumClumps; i++)
+	{
+		int Val1 = m_Noise.IntNoise2DInt(ChunkX * ChunkZ * seq, ChunkZ + ChunkX / seq);
+		int Val2 = m_Noise.IntNoise2DInt(ChunkZ * ChunkX / seq, ChunkZ - ChunkX * seq);
```
madmaxoft: Why not use `+ i` instead of `/ seq`? Then you can get rid of `seq` completely.

NiLSPACE: Great idea

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,215 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+void cFinishGenClumpTopBlock::TryPlaceFoliageClump(cChunkDesc & a_ChunkDesc, int a_CenterX, int a_CenterZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, bool a_IsDoubleTall)
+{
+	int ChunkX = a_ChunkDesc.GetChunkX();
+	int ChunkZ = a_ChunkDesc.GetChunkZ();
+
+	int NumBlocks = m_Noise.IntNoise2DInt(a_CenterX + ChunkX * 16, a_CenterZ + ChunkZ * 16) % (MAX_NUM_FOLIAGE - MIN_NUM_FOLIAGE) + MIN_NUM_FOLIAGE + 1;
+	for (int i = 1; i < NumBlocks; i++)
+	{
+		int x = a_CenterX + ((m_Noise.IntNoise2DInt(i * ChunkX * a_CenterX, i * ChunkZ) % RANGE_FROM_CENTER * 2) - RANGE_FROM_CENTER);
+		int z = a_CenterZ + ((m_Noise.IntNoise2DInt(i * ChunkZ, i * ChunkX * a_CenterZ) % RANGE_FROM_CENTER * 2) - RANGE_FROM_CENTER);
```
madmaxoft: Use a single random integer, prefer `+ i` to `* i` (multiplication is much slower on some architectures), divide the random number by a small prime number to make the distribution more flat, then use `(rnd % 256)` and `((rnd / 256) % 256)` as the bases for each direction.

src/Generating/FinishGen.cpp
```diff
@@ -180,6 +180,215 @@ void cFinishGenNetherClumpFoliage::TryPlaceClump(cChunkDesc & a_ChunkDesc, int a
...
+		int Top = a_ChunkDesc.GetHeight(x, z);
+
+		if (a_ChunkDesc.GetBlockType(x, Top, z) != E_BLOCK_GRASS)
+		{
+			continue;
+		}
+
+		a_ChunkDesc.SetBlockTypeMeta(x, Top + 1, z, a_BlockType, a_BlockMeta);
+
```
madmaxoft: Get rid of the extra empty lines, they serve no real purpose here

src/Generating/FinishGen.cpp
NiLSPACE: Something like this?:
```cpp
int rnd = m_Noise.IntNoise2D(ChunkX + ChunkZ + i, ChunkX - ChunkZ - i) / 59;
int x = a_CenterX + (((rnd % 256) % RANGE_FROM_CENTER * 2) - RANGE_FROM_CENTER);
int z = a_CenterZ + (((rnd / 256) % RANGE_FROM_CENTER * 2) - RANGE_FROM_CENTER);
```

NiLSPACE:
@xoft I made some changes.

sphinxc0re:
Clang is failing because of some issue with implicit floating point conversion

NiLSPACE:
I believe I've fixed it.

NiLSPACE:
@madmaxoft were these the changes you wanted?

madmaxoft:
I guess this is workable. Could you rebase and squash (to make sure it still works on top of recent code changes)?

NiLSPACE:
Other than `continuous-integration/travis-ci` which isn't really going for some reason it looks good.

PR #3076, opened 2016-03-11
HTTP: Added missing HasAuth flag initialization.

madmaxoft:


tigerw:
such security
much determinism
wow

PR #3086, opened 2016-03-21
Revert "Lua callback"

madmaxoft:
Reverts cuberite/cuberite#2872

#2872 caused data races because the `cLuaState::cCallback` class doesn't lock the `cPluginLua::m_CriticalSection`, so multiple threads are allowed to interact with the same Lua state.

Reverting for now, the fix will be a difficult one.


madmaxoft:
Shame on me for not testing the PR thoroughly.

madmaxoft:
Merged even after Travis failure, it was only complaining about undocumented Lua API functions, which was to be expected.

PR #3090, opened 2016-03-24
Out of world

### LINKED ISSUE #2539: Server crash in the PathFinder

### LINKED ISSUE #3047: Crash on trying to use MagicCarpet
Hello guys, I got a crash when I try to use the magic carpet, first I run load MagicCarpet on server console, then ingame I use /mc ... the server crash intantly after fire an assertion trigger

1. [e3419425e791adbd|18:50:57] Assertion failed: !"GetBlock(x, y, z) out of bounds!", file /home/andre/Desktop/Projetos/Minecraft/cuberite/src/Chunk.cpp, line 2591
2. ./Cuberite_debug(_Z15PrintStackTracev+0x23)[0xae0765]
3. ./Cuberite_debug(_ZNK6cChunk8GetBlockEiii+0x5f)[0x9ea08b]
4. ./Cuberite_debug(_ZN9cChunkMap8GetBlockEiii+0xae)[0x9f625a]
5. ./Cuberite_debug(_ZN6cWorld8GetBlockEiii+0x38)[0x9fce86]
6. ./Cuberite_debug[0xb67784]
7. ./Cuberite_debug[0xb68088]
8. ./Cuberite_debug[0xf058bd]
9. ./Cuberite_debug[0xf19a46]
10. ./Cuberite_debug[0xf05b62]
11. ./Cuberite_debug[0xf11546]
12. ./Cuberite_debug[0xf04b91]
13. ./Cuberite_debug[0xf05f58]
14. ./Cuberite_debug(lua_pcall+0x168)[0xf116b0]
15. ./Cuberite_debug(_ZN9cLuaState12CallFunctionEi+0x16b)[0xaee2ab]
16. ./Cuberite_debug(_ZN9cLuaState11PushCallPopIIRbEEEbNS_4cRetEDpOT_+0x28)[0xb07b7e]
17. ./Cuberite_debug(_ZN9cLuaState11PushCallPopIRK7Vector3IdEIRKNS_4cRetERbEEEbOT_DpOT0_+0x67)[0xb33081]
18. ./Cuberite_debug(_ZN9cLuaState11PushCallPopIRK7Vector3IdEIS4_RKNS_4cRetERbEEEbOT_DpOT0_+0x7c)[0xb306d8]
19. ./Cuberite_debug(_ZN9cLuaState11PushCallPopIP7cPlayerJRK7Vector3IdES6_RKNS_4cRetERbEEEbOT_DpOT0_+0x94)[0xb2ce22]
20. ./Cuberite_debug(_ZN9cLuaState4CallIiJP7cPlayerRK7Vector3IdES6_RKNS_4cRetERbEEEbRKT_DpOT0_+0xb1)[0xb28fb3]
21. ./Cuberite_debug(_ZN10cPluginLua14OnPlayerMovingER7cPlayerRK7Vector3IdES5_+0xf7)[0xb2208f]
22. ./Cuberite_debug(_ZN14cPluginManager20CallHookPlayerMovingER7cPlayerRK7Vector3IdES5_+0xbb)[0xb3db13]
23. ./Cuberite_debug(_ZN13cClientHandle15HandlePlayerPosEddddb+0x1dc)[0xa0d920]
24. ./Cuberite_debug(_ZN12cProtocol18021HandlePacketPlayerPosER11cByteBuffer+0x120)[0xc2c9ae]
25. ./Cuberite_debug(_ZN12cProtocol18012HandlePacketER11cByteBufferj+0x17e)[0xc2adea]
26. ./Cuberite_debug(_ZN12cProtocol18015AddReceivedDataEPKcm+0x926)[0xc2a63c]
27. ./Cuberite_debug(_ZN12cProtocol18012DataReceivedEPKcm+0xb7)[0xc2159b]
28. ./Cuberite_debug(_ZN19cProtocolRecognizer12DataReceivedEPKcm+0x17a)[0xc34a8e]
29. ./Cuberite_debug(_ZN13cClientHandle4TickEf+0xd4)[0xa11820]
30. ./Cuberite_debug(_ZN6cWorld11TickClientsEf+0x37a)[0xaad49c]
31. ./Cuberite_debug(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x433)[0xaac6f9]
32. Cuberite_debug: /home/andre/Desktop/Projetos/Minecraft/cuberite/src/Chunk.cpp:2591: BLOCKTYPE cChunk::GetBlock(int, int, int) const: Assertion `0' failed.

The same assertion is triggered when I try to explode a cube with 10k+ tnt, probably physics send a block out of bound

### LINKED ISSUE #3087: Pathfinding: Unread variables
@LogicParrot 
I see that the variables here are assigned again and later not used anymore.
[DoneEast ](https://github.com/cuberite/cuberite/blob/56f3fc954e02236522df7d7edde7273be387d21f/src/Mobs/Path.cpp#L221)
[DoneWest](https://github.com/cuberite/cuberite/blob/56f3fc954e02236522df7d7edde7273be387d21f/src/Mobs/Path.cpp#L237)
[DoneWest](https://github.com/cuberite/cuberite/blob/56f3fc954e02236522df7d7edde7273be387d21f/src/Mobs/Path.cpp#L253) <-- duplicated, do you mean DoneSouth?
[DoneNorth](https://github.com/cuberite/cuberite/blob/56f3fc954e02236522df7d7edde7273be387d21f/src/Mobs/Path.cpp#L269)

Looking at the code, I think they can be removed there, or?

### LINKED ISSUE #3089: Assertion failure when falling to the void
When any entity, including players, dropped items, and mobs, fall below bedrock, an assertion fails. I will post more details soon. Related: #2539

LogicParrot:
Please discuss this before merging.

This PR entirely eliminates the good old "block out of bounds" assertion failures by actually removing these assertions. Whenever a blocks with a negative y value or a too high y value is requested, it is considered an air block (because it is actually an air block) rather than asserting.
Also, non player entities immediately self destruct when their y becomes negative.

**Edit:** I've decided to sum up all my comments here:

- Currently, calling Getblock with a negative (or Too big) height causes an assertion failure.
- In my opinion, calling `GetBlock` with a negative height is not an error at all. It should simply return an air block. All heights should valid for reading (But not for writing). A pathfinder **should** treat a hole in the bedrock just like any other hole. In other words, treat it as air. This applies to all other parts of Cuberite that I've dealt with. So, returning air should be safe for negative heights.
- Currently, most of the GetBlock calls have guarding code: `if Y < 0 || Y too big do not call GetBlock`. The pathfinder even has this:  `if Y < 0 || Y too big do returned air; else return whatever getblock() returns;`. The only purpose of all that guard code is evading an assert which, in my opinion, shouldn't have been there.

Closes #3089
Closes #3087
Closes #2539
Closes #3047
...and possibly others.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -2594,16 +2594,6 @@ bool cChunk::GetSignLines(int a_BlockX, int a_BlockY, int a_BlockZ, AString & a_
 
 BLOCKTYPE cChunk::GetBlock(int a_RelX, int a_RelY, int a_RelZ) const
 {
-	if (
-		(a_RelX < 0) || (a_RelX >= Width) ||
-		(a_RelY < 0) || (a_RelY >= Height) ||
-		(a_RelZ < 0) || (a_RelZ >= Width)
-	)
-	{
-		ASSERT(!"GetBlock(x, y, z) out of bounds!");
-		return 0;  // Clip
-	}
-
 	return m_ChunkData.GetBlock(a_RelX, a_RelY, a_RelZ);
 }
 
--- a/src/ChunkData.cpp
+++ b/src/ChunkData.cpp
@@ -150,9 +150,14 @@ cChunkData::~cChunkData()
 
 BLOCKTYPE cChunkData::GetBlock(int a_X, int a_Y, int a_Z) const
 {
-	ASSERT((a_X >= 0) && (a_X < cChunkDef::Width));
-	ASSERT((a_Y >= 0) && (a_Y < cChunkDef::Height));
-	ASSERT((a_Z >= 0) && (a_Z < cChunkDef::Width));
+	if (
+		(a_X < 0) || (a_X >= cChunkDef::Width) ||
+		(a_Y < 0) || (a_Y >= cChunkDef::Height) ||
+		(a_Z < 0) || (a_Z >= cChunkDef::Width)
+	)
+	{
+		return E_BLOCK_AIR;  // Coordinates are outside outside the world, so this must be an air block
+	}
 	int Section = a_Y / SectionHeight;
 	if (m_Sections[Section] != nullptr)
 	{
@@ -222,7 +227,7 @@ NIBBLETYPE cChunkData::GetMeta(int a_RelX, int a_RelY, int a_RelZ) const
 			return 0;
 		}
 	}
-	ASSERT(!"cChunkData::GetMeta(): coords out of chunk range!");
+	// Coordinates are outside outside the world, so it must be an air block with a blank meta
 	return 0;
 }
 
--- a/src/Mobs/Path.cpp
+++ b/src/Mobs/Path.cpp
@@ -250,7 +250,7 @@ bool cPath::StepOnce()
 		{
 			if (ProcessIfWalkable(CurrentCell->m_Location + Vector3i(0, y, 1),  CurrentCell, NORMAL_G_COST))
 			{
-				DoneWest = true;
+				DoneSouth = true;
 				if (y == 0)
 				{
 					WalkableSouth = true;
--- a/tests/ChunkData/Coordinates.cpp
+++ b/tests/ChunkData/Coordinates.cpp
@@ -13,7 +13,7 @@ int main(int argc, char** argv)
 		{
 			return new cChunkData::sChunkSection();
 		}
-		
+
 		virtual void Free(cChunkData::sChunkSection * a_Ptr)
 		{
 			delete a_Ptr;
@@ -36,7 +36,7 @@ int main(int argc, char** argv)
 		testassert(buffer.GetBlock(0, 32, 0) == 0x0);
 		testassert(buffer.GetMeta(0, 48, 0) == 0x0);
 
-		// Out of Range
+		// Out of range SetBlock
 		CheckAsserts(
 			buffer.SetBlock(-1, 0, 0, 0);
 		);
@@ -55,28 +55,7 @@ int main(int argc, char** argv)
 		CheckAsserts(
 			buffer.SetBlock(0, 0, 256, 0);
 		);
-
-		// Out of Range
-		CheckAsserts(
-			buffer.GetBlock(-1, 0, 0);
-		);
-		CheckAsserts(
-			buffer.GetBlock(0, -1, 0);
-		);
-		CheckAsserts(
-			buffer.GetBlock(0, 0, -1);
-		);
-		CheckAsserts(
-			buffer.GetBlock(256, 0, 0);
-		);
-		CheckAsserts(
-			buffer.GetBlock(0, 256, 0);
-		);
-		CheckAsserts(
-			buffer.GetBlock(0, 0, 256);
-		);
-
-		// Out of Range
+		// Out of range SetMeta
 		CheckAsserts(
 			buffer.SetMeta(-1, 0, 0, 0);
 		);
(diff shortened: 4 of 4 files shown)


LogicParrot:
Closes #3089, #3087, #2539 and possibly others.

LogicParrot:
CI Failure:
```Test failure: assert didn't fire for buffer.GetMeta(-1, 0, 0);, file /home/travis/build/cuberite/cuberite/tests/ChunkData```

If this PR is acceptable, this shouldn't be a problem, I'll just need to modify the tests.

tigerw:
The ::Get class of functions are understandably quite a hot path and used in various loops where more performance is good. The idea was to eventually remove conditional branching so there wouldn't be the need to involve branch prediction and mispredictions. Are we considering that an unreachable goal?

LogicParrot:
Can boundary checking ever be safely removed?

LogicParrot:
>Can boundary checking ever be safely removed?

Almost all boundary checking bugs occur at heights -1 or 256. At the cost of some more memory, boundary checks can be removed by creating a 1 block thick sentinel at the top and bottom of each chunk, always filled with air.

LogicParrot:
Is branching really a concern in this particular case? in almost all cases, a branch predictor will correctly predict-not-taken, because almost all GetBlocks will involved a valid height.

LogicParrot:
What's wrong with my new tests?

LogicParrot:
This PR is also in line with the consensus that plugins should be prevented from crashing the server. This function had been a common way for plugins to crash it.

NiLSPACE:
This looks good to me. I'm not sure why the test fails though :/

tigerw:
> Is branching really a concern in this particular case? in almost all cases, a branch predictor will correctly predict this, because almost all  GetBlock  calls will involve a valid height.

No clue, since nobody's done any testing.

> This PR is also in line with the consensus that plugins should be prevented from crashing the server. This function had been a common way for plugins to crash it.

The checks were there before though, so it's only relaxing Debug sanity checking, right?

-----

Still not wholly convinced. The amount of assertion failures here despite years of effort ~~is testament to how badly we code~~ makes me worry that

a) there could be bugs (remaining despite multiple 'fixes') that are not caught now that were before, and
b) duplicate checking of height validity will be harder to remove


buuut, as always, idk

LogicParrot:
>The checks were there before though, so it's only relaxing Debug sanity checking, right?

Sort of. The consideration is that Asking for the block at (0,-5,0) is not a logical error and there is no reason that should assert. It's simply an air block.

LogicParrot:
And by insisting that it should for some reason assert, you gotta put edge-handling code all over the place. (e.g. pathfinder, entities falling to the void, etc. they all must make careful checks to avoid this assert)

LogicParrot:
>there could be bugs (remaining despite multiple 'fixes') that are not caught now that were before, and

That's worth thinking about. Note that the assert for setting the blocks remains. It's as if all the blocks at Y<0 and Y>maxHeight are read-only air.

I am curious: Did this assert ever catch a legitimate bug? by "legitimate", I mean a bug in the logic. A pathfinder calling GetBlock(42,-1,30) is not a legitimate bug; returning air would make it work perfectly fine.

How about asserting for y < -100 and y > maxHeight + 100? That assert would not catch queries like (3, -1, 7), but if some code calls (142, -12951, 142), then probably something is messed up.

LogicParrot:
I may have said this before but: The main idea here is that what we previously considered bugs aren't really bugs. It's perfectly fine, at least in all the parts of the code I've dealt with, to ask "What's at (15, -5, 200)". The answer would be "air".

>duplicate checking of height validity will be harder to remove
The idea is not checking for height validity because all heights are valid. (Or at least only checking them in the extremes to catch memory corruption bugs, as explained in the previous comment)

LogicParrot:
@madmaxoft opinion?

LogicParrot:
@madmaxoft @worktycho opinion?

worktycho:
I think this is probably a good idea. If we find a hot loop where this is actually an impact, we could then add an internal UnsafeGetBlock for intenal use, but frankly GetBlock is now complicated enough with a branch and a dereference, that its probably not going to make too much difference.

LogicParrot:
@worktycho @tigerw I think we got this wrong. This change means *less*  branches at best and the same amount of branches at worst. Why? Because we already perform boundary checks, but they're spread all around the code. Those checks would be eliminated after this PR. e.g. see the diff at https://github.com/cuberite/cuberite/pull/3148

worktycho:
There are less or equal branches in the code, but more or equal on any one path. Whether that is a improvement or not is difficult to predict, because if the check is duplicated on a path, then that path could have more branch prediction stalls, but if they are not duplicated just moved, than the smaller code will result in better icache utilisation, and more focused branch prediction in a single place.

LogicParrot:
@worktycho Also note that:
- We probablycurrently perform redundant checks (functions that have checks and call functions that have checks)
- Lua plugins currently have to perform checks. They won't have to after this. Lua is slower.

tigerw:
> ...then that path could have more branch prediction stalls, but if they are not duplicated just moved, than the smaller code will result in better icache utilisation, and more focused branch prediction in a single place.

^ dis guy ^ knows what he's talkin' about.

worktycho:
If there are duplicate checks asserting conditions, then the performance can only improve, if we remove them. And getting common code out of lua is definitely an improvement.

LogicParrot:
>There are less or equal branches in the code, but more or equal on any one path.

Under what condition would the new code spawn more branches? The way I see it is that we're merely moving the checks from everywhere to `GetBlock`, so the new code should either have an equal amount of branches, or *less* (In the case of redundancy, functions performing checks and then calling functions that perform checks).

worktycho:
The condition it would produce more is if you have missed a check somewhere.

LogicParrot:
Merge?

worktycho:
I say merge.

PR #3092, opened 2016-03-24
Never put player inside a solid when moving to a new world

### LINKED ISSUE #3088: /portal command can teleport the player to non-air blocks, making them stuck inside

LogicParrot:
Closes #3088

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1790,7 +1790,50 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 		);
 		ParentChunk->RemoveEntity(this);
 		a_World->AddPlayer(this);
+		AString PlayerUUID = this->GetUUID();
 		cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*this, a_OldWorld);
+
+		// If the player teleported into a solid, schedule moving them out
+		a_World->QueueTask([PlayerUUID](cWorld & a_NewWorld)
+		{
+			class cCallback : public cPlayerListCallback
+			{
+				virtual bool Item(cPlayer * a_Player) override
+				{
+					Vector3d PlayerPos = a_Player->GetPosition();
+					PREPARE_REL_AND_CHUNK(PlayerPos, *(a_Player->GetParentChunk()));
+					if (RelSuccess)
+					{
+						int NewY = Rel.y;
+						if (NewY < 0)
+						{
+							NewY = 0;
+						}
+						while (NewY < cChunkDef::Height - 2)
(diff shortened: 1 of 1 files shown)


src/Entities/Player.cpp
```diff
@@ -1762,7 +1762,48 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
+						int NewY = Rel.y;
+						if (NewY < 0)
+						{
+							NewY = 0;
+						}
+						while (NewY < cChunkDef::Height - 2)
+						{
+							// If we find a position with enough space for the player
+							if ((Chunk->GetBlock(Rel.x, NewY, Rel.z) == E_BLOCK_AIR) &&
+							(Chunk->GetBlock(Rel.x, NewY + 1, Rel.z) == E_BLOCK_AIR))
```
NiLSPACE: Is this valid or should the line have one more indent?

LogicParrot: Fixed.

Schwertspize:
In my opinion moving to a new world should always teleport to the world's spawn if not explicitly stated otherwise (by the plugin)

LogicParrot:
@Schwertspize This PR does not prevent moving to the world's spawn point.

If you are referring to the way the `/portal` command works, then I agree. But this would require changes to the `core` plugin and is a seperate issue.

If you are referring to the API's `MoveToWorld` function, then I also agree, and it should be done in a seperate PR.

NiLSPACE:
I disagree with the `MoveToWorld` method. I think it should calculate the proper coordinates automatically unless stated otherwise.

LogicParrot:
@NiLSPACE I may have been a bit vague. I meant this: `MoveToWorld` with a specified position vector should teleport you to where you ask it to. `MoveToWorld` without a specified position should teleport to the spawn point. Currently the latter teleports you to a new world while retaining the coordinates you had in the old world. Is there any disagreement here?

NiLSPACE:
I hade the opposite in mind, teleport to calculated area when none is given, but after giving it a second thought I think your way makes more sense. Perhaps we should make a few static functions as well to calculate the proper coordinates for overworld <-> nether changes?

LogicParrot:
What do you mean by "calculated area"?

NiLSPACE:
For example when going to the overworld from the nether every block counts as 8.

When I'm in 4, 64, 9 in the nether it means I'm in 32, 64, 72 in the overworld.

LogicParrot:
@Schwertspize Sorry, I misunderstood your initial comment.

I agree that #3109 makes this redundant. Closing.

LogicParrot:
Merged with #3097

PR #3096, opened 2016-03-27
Updated icons and logo

mathias-github:
- Updated the icon used in Windows. The .ico file includes the same resolutions as the old one (16, 32, 48, 128 and 256)
- Updated the WebAdmin logo and favicon


NiLSPACE:
Could you post a picture on how the webadmin looks with that logo?

mathias-github:
screenshot from 2016-03-27 22 37 42
screenshot from 2016-03-27 22 37 17

NiLSPACE:
Oh, that looks better than I expected. :)

Schwertspize:
Looks good.... But maybe combine it with a complete rewrite once we are so far

tigerw:
Make sure that the rewrite has a counter in the top corner for "generations". Once we get to approx. [600 generations](https://www.youtube.com/watch?v=pgaEE27nsQw), the webadmin will start to develop itself.

NiLSPACE:
A rewrite of the webadmin? Again!? This would be like the 4th rewrite already. In my opinion this is the best layout we've ever had, no need to change it.

tigerw:
[Bit rot is real - embrace the rewrites!](https://en.wikipedia.org/wiki/Bit_rot)

109C:
@tigerw http://www.catb.org/jargon/html/B/bit-rot.html

SphinxC0re:
I agree that the current layout is good. But shouldn't the colors get a rewrite to fit the website?

mathias-github:
Perhaps, but I will not include anything else in this PR. Maybe someone can open an issue or a forum thread about the WebAdmin layout/design?

mathias-github:
Rebased.

PR #2814, opened 2015-12-24
Freeze player until chunk is loaded

SafwatHalaby:
This should prevent players sinking into ground on portal teleportation on slow devices like the RaspberryPi. I did not test this. Can someone try it out? @jammet ?

closes #2316

PR #3097, opened 2016-03-27
Never fall through unloaded or unsent chunks

### LINKED ISSUE #2964: Investigate unloaded chunk fall
One can fall through unloaded chunks upon world travel, which means #2814 is incomplete.

### LINKED ISSUE #3088: /portal command can teleport the player to non-air blocks, making them stuck inside

LogicParrot:
Players now freeze when going to chunks that are not yet sent to the client. This typically happens when flying fast, running, teleporting to other worlds, or teleporting to far coordinates on the same world.

Also, players never get stuck inside solids on teleport.

Closes #2964 
Closes #3088

Todo:
 - ~~For the world travel case, also send the proper packets for the teleport loading screen. As suggested by @tigerw in #2814~~
 - Done: Optimize the clientHandle changes. Currently the code is slow, written quickly just for testing.

Binaries:
- [Gnu/Linux](https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20ingame-testing-needed/7/artifact/Cuberite.tar.gz)
- [Raspberry Pi](https://builds.cuberite.org/job/Cuberite%20Linux%20raspi-armhf%20ingame-testing-needed/6/artifact/Cuberite.tar.gz)
- [Windows](https://builds.cuberite.org/job/Cuberite%20Windows%20x64%20ingame-testing-needed/5/artifact/Install/Cuberite.zip)

Note: Chunk generator's logging was supressed because it adds noise when testing this PR by flying fast. This will be undone in the final version.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -3,6 +3,7 @@
 #include "ClientHandle.h"
 #include "Server.h"
 #include "World.h"
+#include "Chunk.h"
 #include "Entities/Pickup.h"
 #include "Bindings/PluginManager.h"
 #include "Entities/Player.h"
@@ -65,6 +66,7 @@ cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
 	m_RequestedViewDistance(a_ViewDistance),
 	m_IPString(a_IPString),
 	m_Player(nullptr),
+	m_CachedSentChunk(0, 0),
 	m_HasSentDC(false),
 	m_LastStreamedChunkX(0x7fffffff),  // bogus chunk coords to force streaming upon login
 	m_LastStreamedChunkZ(0x7fffffff),
@@ -336,6 +338,7 @@ void cClientHandle::Authenticate(const AString & a_Name, const AString & a_UUID,
 
 	// Spawn player (only serversided, so data is loaded)
 	m_Player = new cPlayer(m_Self, GetUsername());
+	InvalidateCachedSentChunk();
 	m_Self.reset();
 
 	cWorld * World = cRoot::Get()->GetWorld(m_Player->GetLoadedWorldName());
@@ -1869,6 +1872,18 @@ void cClientHandle::RemoveFromWorld(void)
 
 
 
+void cClientHandle::InvalidateCachedSentChunk()
+{
+	ASSERT(m_Player != nullptr);
+	// Sets this to a junk value different from the player's current chunk, which invalidates it and
+	// ensures its value will not be used.
+	m_CachedSentChunk = cChunkCoords(m_Player->GetChunkX() + 500, m_Player->GetChunkZ());
+}
+
+
+
+
+
 bool cClientHandle::CheckBlockInteractionsRate(void)
 {
 	ASSERT(m_Player != nullptr);
@@ -1926,6 +1941,36 @@ void cClientHandle::Tick(float a_Dt)
 		return;
 	}
 
+
+	// Freeze the player if it is standing on a chunk not yet sent to the client
+	{
+		bool PlayerIsStandingAtASentChunk = false;
+		// If the chunk is invalid, do not bother checking if it's sent to the client, it is definitely not
+		if (m_Player->GetParentChunk()->IsValid())
+		{
+			// Before iterating m_SentChunks, see if the player's coords equal m_CachedSentChunk
+			// If so, the chunk has been sent to the client. This is an optimization that saves an iteration of m_SentChunks.
+			if (cChunkCoords(m_Player->GetChunkX(), m_Player->GetChunkZ()) == m_CachedSentChunk)
+			{
+				PlayerIsStandingAtASentChunk = true;
+			}
+			else
+			{
+				// This block is entered only when the player moves to a new chunk, invalidating the cached coords.
+				// Otherwise the cached coords are used.
+				cCSLock Lock(m_CSChunkLists);
+				auto itr = std::find(m_SentChunks.begin(), m_SentChunks.end(), cChunkCoords(m_Player->GetChunkX(), m_Player->GetChunkZ()));
+				if (itr != m_SentChunks.end())
+				{
+					m_CachedSentChunk = *itr;
+					PlayerIsStandingAtASentChunk = true;
+				}
+			}
+		}
+		// The player will freeze itself if it is standing on a chunk not yet sent to the client
+		m_Player->TickFreezeCode(PlayerIsStandingAtASentChunk);
+	}
+
 	// If the chunk the player's in was just sent, spawn the player:
 	if (m_HasSentPlayerChunk && (m_State == csDownloadingWorld))
 	{
@@ -1957,7 +2002,7 @@ void cClientHandle::Tick(float a_Dt)
 			}
 		}
 
-		// Unload all chunks that are out of the view distance (all 5 seconds)
+		// Unload all chunks that are out of the view distance (every 5 seconds)
 		if ((m_Player->GetWorld()->GetWorldAge() % 100) == 0)
 		{
 			UnloadOutOfRangeChunks();
--- a/src/ClientHandle.h
+++ b/src/ClientHandle.h
@@ -363,6 +363,8 @@ public:  // tolua_export
 	/** Returns the protocol version number of the protocol that the client is talking. Returns zero if the protocol version is not (yet) known. */
 	UInt32 GetProtocolVersion(void) const { return m_ProtocolVersion; }  // tolua_export
 
+	void InvalidateCachedSentChunk();
+
 private:
 
 	friend class cServer;  // Needs access to SetSelf()
@@ -407,6 +409,13 @@ private:
 	Vector3d m_ConfirmPosition;
 
 	cPlayer * m_Player;
+
+	/** This is an optimization which saves you an iteration of m_SentChunks if you just want to know
+	whether or not the player is standing at a sent chunk.
+	If this is equal to the coordinates of the chunk the player is currrently standing at, then this must be a sent chunk
+	and a member of m_SentChunks.
+	Otherwise, this contains an arbitrary value which should not be used. */
+	cChunkCoords m_CachedSentChunk;
 
 	bool m_HasSentDC;  ///< True if a Disconnect packet has been sent in either direction
 
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -251,35 +251,10 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	// Handle a frozen player
 	if (m_IsFrozen)
 	{
-		m_FreezeCounter += 1;
-		if (!m_IsManuallyFrozen && a_Chunk.IsValid())
-		{
-			// If the player was automatically frozen, unfreeze if the chunk the player is inside is loaded
-			Unfreeze();
-		}
-		else
-		{
-			// If the player was externally / manually frozen (plugin, etc.) or if the chunk isn't loaded yet:
-			// 1. Set the location to m_FrozenPosition every tick.
-			// 2. Zero out the speed every tick.
-			// 3. Send location updates every 60 ticks.
-			SetPosition(m_FrozenPosition);
-			SetSpeed(0, 0, 0);
-			if (m_FreezeCounter % 60 == 0)
-			{
-				BroadcastMovementUpdate(m_ClientHandle.get());
-				m_ClientHandle->SendPlayerPosition();
-			}
-			return;
-		}
-	}
-
-	if (!a_Chunk.IsValid())
-	{
-		FreezeInternal(GetPosition(), false);
-		// This may happen if the cPlayer is created before the chunks have the chance of being loaded / generated (#83)
 		return;
 	}
+
+	ASSERT(a_Chunk.IsValid());
 
 	super::Tick(a_Dt, a_Chunk);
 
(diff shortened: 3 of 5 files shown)


src/Entities/Player.cpp
```diff
@@ -337,6 +317,51 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
+void cPlayer::TickFreezeCode(bool a_MyChunkIsSent)
+{
+	// This function is ticked by the player's client handle. This ensures it always ticks, even if the player
+	// is standing in an unloaded chunk, unlike cPlayer::Tick. We need this because the freeze handling code must tick
+	// in unloaded chunk.
```
LogicParrot: grammar

tigerw:
aaaaah more merge conflicts for me :P

tigerw:
But I like jumping into unloaded chunks and exploring caves! I demand to know if you are attempting to limit my fun!

jammet:
Amazing! :) I've been flying and running around, throwing and plunging myself into suicidal, maniacal pits of nothingness, and the server kept me afloat, teleported me back to the topmost block once it was actually there. I'm not sure what else exactly to test, but this did work NICELY now! :]

LogicParrot:
@bearbin the builds are huge in the ingame-testing-needed branches at the build server. Any idea why?

LogicParrot:
Sorry. It's simply because it's a debug build, which is desired for that branch.

jammet:
I noticed you were trying to use some commands from the server log, this
morning, and I just ranked you to Admin so you do what you need when I'm
not around at all. =^_^= Sorry I wasn't there.

LogicParrot:
Can the code be simplified and performance be improved by simply calling `if !GetParentChunk()->IsValid() then freeze`?

src/Entities/Player.cpp
```diff
@@ -337,6 +317,51 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+	if (m_IsFrozen)
+	{
+		m_FreezeCounter += 1;
+		if ((!m_IsManuallyFrozen) && (a_MyChunkIsSent))
+		{
+			LOGD("UNFREEZE\n");
```
LogicParrot: remove logd

LogicParrot:
I've finished the optimization. I will implement the loading screen later in a seperate PR, there are more urgent issues right now. lgtm.

jammet:
=^_^= As long as the loading screen does not say "PRESS PLAY ON CASSETTE
TAPE AND PRESS ANY KEY" it's gonna be awesome ;).

tigerw:
I was wondering if you could do it another way, by storing the last position of the player when they went into an unloaded chunk, and teleporting them there again when the chunk loads. I was thinking that this wouldn't introduce the jerky view that arises with trying to freeze the player in one position.

LogicParrot:
@tigerw I'll see if I can do something similar or reduce jerkiness without adding too much code. Note that your proposal wouldn't work for moving to other worlds or logging in though without some modification.

LogicParrot:
Do we really need to teleport them again when it loads?

LogicParrot:
I found a far better optimization which also greatly simplifies this code.e Do not merge yet.

LogicParrot:
@tigerw You mentioned earlier that `A respawn packet when sent causes the client to exit the holding screen and start rendering the world.`. Where is the code that triggers the loading screen in the first place?

LogicParrot:
- The "far better optimization" turned out to be flawed.
 - I tried making the loading screen pop up as the chunk loads to avoid the jitter, but that didn't work. 
 - I think I'll leave it as is for now and fix jitter later. There are far more urgent issues on my list.

lgtm :P

tigerw:
The loading screen is on world change (including login). I was not clear; one does not simply trigger a loading screen arbitrarily.

LogicParrot:
@tigerw Alright. Do you have any objections to merge the current changes?

src/Entities/Player.cpp
```diff
@@ -339,6 +314,79 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
+void cPlayer::TickFreezeCode(bool a_MyChunkIsSent)
+{
+	// This function is ticked by the player's client handle. This ensures it always ticks, even if the player
+	// is standing in an unloaded chunk, unlike cPlayer::Tick. We need this because the freeze handling code must
```
tigerw: They player doesn't tick when standing in an unloaded chunk? That sounds like a bug if it isn't already...

src/ClientHandle.h
```diff
@@ -408,6 +410,13 @@ class cClientHandle  // tolua_export
 
 	cPlayer * m_Player;
 
+	/** This is an optimization which saves you an iteration of m_SentChunks if you just want to know
+	whether or not the player is standing at a sent chunk.
+	If this is equal to the coordinates of the chunk the player is currrently standing at, then this must be a sent chunk
+	and a member of m_SentChunks.
+	Otherwise, this contains an arbitrary value which should not be used. */
+	cChunkCoords m_CachedSentChunk;
```
tigerw: Don't bother. It's not worth it for a new variable, and besides, bugs will manifest when teleporting to ChunkX + 500.

src/Entities/Player.cpp
```diff
@@ -339,6 +314,79 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+	// also tick in unloaded chunks.
+	if (m_IsFrozen)
+	{
+		m_FreezeCounter += 1;
+		if ((!m_IsManuallyFrozen) && (a_MyChunkIsSent))
+		{
+			cWorld::cLock Lock(*GetWorld());
```
tigerw: Isn't cClientHandle::Tick called in the world tick thread? Not after #3115, but more merge conflicts for me :P

src/Entities/Player.cpp
```diff
@@ -339,6 +314,79 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			// If the player was automatically frozen, unfreeze if the chunk the player is inside is loaded
+			Unfreeze();
+
+			// Pull the player out of any solids that might have loaded on them.
```
tigerw: Isn't there already a FindNextHighestNonAirBlock in cMonster or somewhere?

src/Entities/Player.cpp
```diff
@@ -339,6 +314,79 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			}
+		}
+		else
+		{
+			// If the player was externally / manually frozen (plugin, etc.) or if the chunk isn't loaded yet:
+			// 1. Set the location to m_FrozenPosition every tick.
+			// 2. Zero out the speed every tick.
+			// 3. Send location updates every 60 ticks.
+
+			if ((m_FreezeCounter % 60 == 0) || ((m_FrozenPosition - GetPosition()).SqrLength() > 2 * 2))
```
tigerw: Is there really any need? Frozen means all position and speed updates are ignored, right? Just let them fall slowly. That m_FreezeCounter variable is looking awfully extraneous.

src/ClientHandle.cpp
```diff
@@ -1926,6 +1941,36 @@ void cClientHandle::Tick(float a_Dt)
 		return;
 	}
 
+
+	// Freeze the player if it is standing on a chunk not yet sent to the client
+	{
+		bool PlayerIsStandingAtASentChunk = false;
```
tigerw: No need, simply call TickFreezeCode in the remaining two paths.

src/Entities/Player.cpp
```diff
@@ -339,6 +314,79 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+				m_ClientHandle->SendPlayerPosition();
+			}
+			return;
+		}
+	}
+	else
+	{
+		if (!a_MyChunkIsSent)
+		{
+			FreezeInternal(GetPosition(), false);
```
tigerw: And so without FreezeCounter you can pull this up to the top of the function.

src/Entities/Player.cpp
jammet: If by that you mean, let a player still slowly sink, then I'd really prefer to be frozen entirely. I cannot remember how many times people and I suffocated underground. Being frozen in position would help it so much. Then again, just guessing it's that what you mean.

tigerw:
...and, speaking of [Frozen](https://www.youtube.com/watch?v=L0MK7qz13bU)...

jammet:
Aaah! Overrated movie! XD Let it go!

src/Entities/Player.cpp
tigerw: I am thinking that since the chunk hasn't been loaded, they would not suffocate. Once it is loaded, their position would be reset to a non-fatal one.

jammet: Oh, alright then. (zzzzzzzzzzip!)

tigerw: there's no need to zip :P

tigerw:
But yeah, as I said, I like to be pedantic.

src/Entities/Player.cpp
LogicParrot: Frozen means you keep sending them back to the same spot. `m_FrozenPosition` is needed for when plugins decide to manually freeze a player.

When, precisely, goes the client decide to go into "sink mode"?

LogicParrot: @tigerw Note that the freeze code is also used for freezing players on fully loaded chunks by plugins calling `Freeze`. I continously send the position to the player to prevent slow sinking **and** to prevent player movement in loaded chunks when a plugin ask forthat. I do not employ any client-side freezing.

But perhaps a better technique should be employed. Can we, in any way, really freeze the player client-side? If I set their speed to 0 and set their isFlying to true, I wonder if that would practically be a clientside freeze?

If a clientside freeze is possible, then I could do it like this:
 - Freeze clientside
 - Ignore position updates to avoid hacking
 - Upon unfreeze, unfreeze clientside, stop ignoring position updates

I also don't like slow sink because:
 - The player could sink into a cave by the time the chunk loads, Unless we resend original Position after the chunk is sent, but that would feel like a tiny jump, the above method, if it works, would be superior.
 - Things get strange at y<=0

tigerw: > I continously send the position to the player to prevent slow sinking and to prevent player movement in loaded chunks when a plugin asks for that. I do not employ any client-side freezing.

Forgot about plugins, whoops. Nevertheless, you can still drop both FrozenPosition and FreezeCounter. Use the world age to periodically resend position, and check in cPlayer::SetPosition/SetSpeed that the player is not frozen before setting.

I'm not sure about client freezing, but if someone knows how we can get rid of periodic resending.

nah, slow sink is fine; we both agree on sending the correct position (I propose topmost block) upon unfreeze due to chunk loaded, so no stuck in caves. And if there is a Y height bug, well, let's fix it :D

tigerw:
Also @jammet, is this [urgent](https://forum.cuberite.org/thread-2410.html)? It could be merged in its current state if players dying is becoming annoying, and fixed up later.

jammet:
It would be a big deal. It's one of the changes that make a huge difference
for many players. I'll let you decide if that's worth merging it soon, but
I'm all for it.

LogicParrot:
@tigerw Mysteriously falling to the void is probably the #1 most urgent issue for the Raspberry Pi.

src/Entities/Player.cpp
LogicParrot: Entities do not tick in unloaded chunks. I don't know if this is fine.

src/ClientHandle.h
LogicParrot: Are you sure? This is not just a micro optimization. Without this, an extra lock will be held per player per tick, and every tick, `m_SentChunks` would be fully iterated per player.

If you don't like the extra variable, we could turn `m_SentChunks` into an unordered map and get rid of this. I need an O(1) answer to the question of `is this chunk sent?` and an unordered map provides that.

src/Entities/Player.cpp
LogicParrot: I blindly followed the advice at `cWorld::TickMobs` and locked it just like it is locked there. It had to do with having to lock chunkmap before accessing the world.

src/Entities/Player.cpp
LogicParrot: >Isn't there already a FindNextHighestNonAirBlock in cMonster or somewhere?

Yes, there are probably similar functions several times. `EnsureProperPosition` also comes to mind. I was planning to unify them and have a generic function in a seperate PR. Because preventing chunk fall seemed more urgent to me. That function would accept mob height, block type, and perhaps some other stuff like direction of scanning.

LogicParrot: > Or just go to the highest block, easy for everyone.

If I have a giant hall and a spawn point inside, and then I decide to elevate the floor where the spawn point is at, I wouldn't want my players to spawn on the rooftop of the hall.

src/Entities/Player.cpp
LogicParrot: > Nevertheless, you can still drop both FrozenPosition and FreezeCounter.

Good point. I can drop FrozenPosition if I ignore the clientside position updates.
Anyways, I will see if I can clientside freeze. That sounds optimal.

src/ClientHandle.cpp
LogicParrot: Good point.

LogicParrot:
>Also @jammet, is this urgent? It could be merged in its current state if players dying is becoming annoying, and fixed up later.

I agree with that. If you see no critical problems with the current state, let's merge. I assure you I will treat the issues we discussed asap.

src/Entities/Player.cpp
LogicParrot: > Nevertheless, you can still drop both FrozenPosition and FreezeCounter.

Good point. I can drop FrozenPosition if I ignore the clientside position updates and FreezeCounter can be trivially dropped.

Anyways, I will see if I can clientside freeze. That sounds optimal.

src/Entities/Player.cpp
LogicParrot: Similarly, if someone covers a spawnpoint which resides at the nether, you wouldn't want them top spawn on the top of the nether.

LogicParrot: Similarly, if someone covers a spawnpoint which resides at the nether, you wouldn't want players to spawn on the top of the nether.

src/Entities/Player.cpp
LogicParrot: Isn't that fine though? Do we want a floating pack of skeletons to shoot at a slowly sinking player while a chunk is loading?

jammet: About if it would help, and if it would be be a good thing to have it as
soon as possible? Yes :). A big, blinking, singing and dancing YES.
Connectivity/lag spike/void issues have been a thorn in everyone's side for
far too long, and some players thought it pointless to even try and connect
after a while, because sometimes they fell from the sky, sometimes they
suffocate down below, or fell through the world, and when they were just
stuck in a way they had to practically take the spawn point apart by
desperately digging themselves out. :) It's been a bit stressy as of late.

LogicParrot:
@jammet The latest download now has this fix :)

src/Entities/Player.cpp
LogicParrot: Clientside freeze can be attained by setting flight and walk speed to 0, then setting the player to flying (to avoid falling).

src/ClientHandle.h
tigerw: I did the map thing in #3115 :P

LogicParrot:
Reminder to self. Did I treat the remaining issues?

PR #3099, opened 2016-03-28
Fix passive mob destruction crashes

LogicParrot:


LogicParrot:
I owe @jammet an apology for crashing his server twice. 
On the bright side, the cause of the crash has been found and fixed.

Issue #283, opened 2013-10-21
cWorld::TickMobs() doesn't lock chunkmap


madmaxoft:
The cWorld::TickMobs() function doesn't lock the chunkmap, yet it calls the monster's Tick() method which assumes the chunkmap is locked (because it has direct cChunk access)

LogicParrot:
I'd like to move the mob tick back to cChunk. It should be safe for me to remove this lock, right?

PR #3100, opened 2016-03-28
Move mob tick code to cChunk

LogicParrot:
After years, the mob tick code moves back to cChunk.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -40,6 +40,7 @@
 #include "SetChunkData.h"
 #include "BoundingBox.h"
 #include "Blocks/ChunkInterface.h"
+#include "Mobs/Wolf.h"
 
 #include "json/json.h"
 
@@ -641,13 +642,11 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 			continue;
 		}
 
-		if (!((*itr)->IsMob()))  // Mobs are ticked inside cWorld::TickMobs() (as we don't have to tick them if they are far away from players)
-		{
-			// Tick all entities in this chunk (except mobs):
-			ASSERT((*itr)->GetParentChunk() == this);
-			(*itr)->Tick(a_Dt, *this);
-			ASSERT((*itr)->GetParentChunk() == this);
-		}
+		// Tick all entities in this chunk - including mobs.
+		// In the past Cuberite ticked mobs in cWorld. This is no longer the case.
+		ASSERT((*itr)->GetParentChunk() == this);
+		(*itr)->Tick(a_Dt, *this);
+		ASSERT((*itr)->GetParentChunk() == this);
 
 		// Do not move mobs that are detached from the world to neighbors. They're either scheduled for teleportation or for removal.
 		// Because the schedulded destruction is going to look for them in this chunk. See cEntity::destroy.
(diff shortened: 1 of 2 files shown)


NiLSPACE:
What are the advantages of having the code in cChunk?

LogicParrot:
- Simplicity. All entities ticks in the chunk, why should mobs be any different?
 - Performance. Every tick, cWorld must hold a lock. Also, it must iterate all entities, which means iterating all chunks for the second time (it has already iterated them for the first time to tick them). Why not take advantage of the already existing chunk tick iteration? 

The real question is: Why is the mob tick code in cWorld? Well, it was put there as a hack to achieve proper mob destruction. This PR can destroy mobs properly.

As for the lock in cWorld::TickMobs, I still need to remove or rewrite MobCensus before removing it.

LogicParrot:
Relevant issues: #277 #283

src/Chunk.cpp
```diff
@@ -1864,13 +1863,41 @@ void cChunk::RemoveClient(cClientHandle * a_Client)
 		}
 	}
 
+	// All clients were removed. No one sees this chunk any more.
+	// Delete all hostiles inside it.
+	if (!HasAnyClients())
+	{
+		for (cEntityList::iterator itr = m_Entities.begin(); itr != m_Entities.end(); ++itr)
```
tigerw: The C++ standards committee expresses its dissatisfaction at the non-use of C++11 features here.

src/Chunk.cpp
```diff
@@ -1864,13 +1863,41 @@ void cChunk::RemoveClient(cClientHandle * a_Client)
...
+		{
+			cEntity * Entity = *itr;
+			if (Entity->IsTicking() && Entity->IsMob())
+			{
+				auto Monster = static_cast<cMonster *>(Entity);
+				if (
+					(Monster->GetMobFamily() == cMonster::eFamily::mfHostile) &&
+					(Monster->GetCustomName() == "")  // Monsters that have name tags are never removed
```
tigerw: The writers of `std::string::empty` express their dissatisfaction at the non-use of this function here.

src/Chunk.cpp
```diff
@@ -1864,13 +1863,41 @@ void cChunk::RemoveClient(cClientHandle * a_Client)
...
+					}
+				}
+			}
+		}
+	}
 	return;
 }
 
 
-
```
tigerw: @madmaxoft expresses his dissatisfaction at your non-compliance with the style guide here.

src/Chunk.cpp
```diff
@@ -1864,13 +1863,41 @@ void cChunk::RemoveClient(cClientHandle * a_Client)
...
+						auto Wolf = static_cast<cWolf *>(Monster);
+						if (Wolf->IsAngry())
+						{
+							Monster->Destroy(true);
+						}
+					}
+				}
+			}
+		}
+	}
```
tigerw: Braces here, braces there, braces everywhere!

LogicParrot:
@tigerw No more conflicts any time soon. Promise :)

LogicParrot:
Sorry, merged this on accident. I want to make sure all pointers behave right first.

Issue #3045, opened 2016-02-22
Investigate addition of entities to invalid chunks


LogicParrot:
#3044 deals with #3042 and #3035 deals with #3030, however, both are quick fixes, and we should check if there are more fundamental problems with dealing with entities going into invalid chunks, and there probably are such problems.

tigerw:
Normally such entities are simply lost.
 
Date: Mon, 22 Feb 2016 00:43:10 -0800

madmaxoft:
This is a scenario that should be completely avoided. Entities shouldn't ever be moved to an unloaded chunk.

My initial idea on how to solve this was to introduce a `m_HasAllNeighbors` flag for each chunk that would be true if all 8 neighbors of this chunk were valid; and all entities in such a chunk would be ticked only if the flag is true. This would mean that entities on loaded world's border won't be ticked, which seems like a good idea to me - minecarts would stop moving when they reach the edge of the world, rather than when leaving the loaded world completely etc.

Just haven't got around to actually implementing this yet, and I don't currently know how it would work with the ParentChunk rewrite.

LO1ZB:
Please do not forget, that some (block) entities are capable of loading/generation chunks without a player near by, like PrimedTnT and Minecarts in vanilla Minecraft. I dont have time to search for a list, will do that later today.

Schwertspize:
Do we really want that tnt minecarts generate a new chunk? (obviously that seems strange because they can't be on the track any longer. Maybe anotherr possibility instead of just not ticking would be ticking them 100 times slower.

A way better thing would be if we handle unloaded or invalid chunks like world border in Vanilla. It's basically a border between blocks which are handled LIKE solid blocks or something I think

NiLSPACE:
If an entity other than a player goes out of the loaded chunk can't we:
* load or generate the chunk 
* save the entity in the newly loaded chunk
* unload the chunk
?

I believe Vanilla does that as well.

PR #3105, opened 2016-03-29
Entities are never lost

### LINKED ISSUE #3104: Spawning in an ungenerated chunk causes fatal problems
When spawning in an ungenerated chunk, cWorld::3836 is not executed, causing instability in cPlayer.

LogicParrot:
Closes #3104 #3045

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -693,10 +693,9 @@ void cChunk::MoveEntityToNewChunk(cEntity * a_Entity)
 	cChunk * Neighbor = GetNeighborChunk(a_Entity->GetChunkX() * cChunkDef::Width, a_Entity->GetChunkZ() * cChunkDef::Width);
 	if (Neighbor == nullptr)
 	{
-		Neighbor = m_ChunkMap->GetChunkNoLoad(a_Entity->GetChunkX(), a_Entity->GetChunkZ());
-		if (Neighbor == nullptr)
+		Neighbor = m_ChunkMap->GetChunk(a_Entity->GetChunkX(), a_Entity->GetChunkZ());
+		if (Neighbor == nullptr)  // This will assert inside GetChunk in debug builds
 		{
-			// TODO: What to do with this?
 			LOGWARNING("%s: Failed to move entity, destination chunk unreachable. Entity lost", __FUNCTION__);
 			return;
 		}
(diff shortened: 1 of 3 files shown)


tigerw:
Wouldn't this mean entities will forcibly load chunks?

LogicParrot:
@tigerw Yes.

This solution is very simple, previously suggested by @NiLSPACE:
If a mob is being sent to an unloaded chunk, load it or generate it, and put the mob there as usual. If the chunk is too far, it will not tick and will unload quickly. This is very much desired: If I send an entity to a far chunk, I'd like the server to load it and add the entity to it, then save and unload it.

SphinxC0re:
Could you convert [this](https://github.com/LogicParrot/cuberite/blob/chunkStability/src/World.cpp#L1017) into a C++11 loop

LogicParrot:
@SphinxC0re will do.

SphinxC0re:
Also why are you still not working in the canon repo?

SphinxC0re:
Also (sorry for comment chaining) there is another change on line 3000 or so where you could easily convert such a loop to C++11

LogicParrot:
@SphinxC0re You could easily comment on the diff though.

>Also why are you still not working in the canon repo?

1. Habit. Maybe I should do `git remote remove parrot` to force-break it.
2. It's a bit tidier. Look at how many branches you guys have here :P

SphinxC0re:
I think we'll be okay with the branch count :wink:

LogicParrot:
In a followup PR, I think I'll also implement @madmaxoft 's idea:

> My initial idea on how to solve this was to introduce a m_HasAllNeighbors flag for each chunk that would be true if all 8 neighbors of this chunk were valid; and all entities in such a chunk would be ticked only if the flag is true. This would mean that entities on loaded world's border won't be ticked, which seems like a good idea to me - minecarts would stop moving when they reach the edge of the world, rather than when leaving the loaded world completely etc.

I could then ake advantage of that flag to reimplement #3097 more efficiently and remove m_CachedSentChunk: `if !GetParentChunk()->HasAllNeighbords then freeze player`.

LogicParrot:
@SphinxC0re There are lots of them. But I prefer not to mix PRs. We need a seperate PR which refactors all loops.

LogicParrot:
This makes Cuberite unstable after removing a world and then logging in, unless one is near the spawn point.
It also makes it unstable on world travel when warping to an ungenerated chunk.
It makes it a total mess in the Forgetful schema.

I am going to merge this despite the lack of review. Please do review it.

SphinxC0re:
Why are you merging this despite the proposed instability of the builds. You shouldn't.

LogicParrot:
The build is unstable **without** this.

LogicParrot:
I edited my post to avoid confusion.

PR #3109, opened 2016-03-30
MoveToWorld defaults to spawnpoint

### LINKED ISSUE #3088: /portal command can teleport the player to non-air blocks, making them stuck inside

### LINKED ISSUE #3094: MoveToWorld is not fully documented
It seems that `MoveToWorld` is not fully documented. The version with the destination position specified does not appear in the [cEntity api docs](http://api-docs.cuberite.org/cEntity.html). The Corresponding c++ code is [here](https://github.com/cuberite/cuberite/blob/master/src/Entities/Entity.h#L409). The relevant docs line is [here](https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/APIDesc.lua#L931).

LogicParrot:
>In my opinion moving to a new world should always teleport to the world's spawn if not explicitly stated otherwise (by the plugin) 

*-@Schwertspize in #3092*

Closes #3088
Closes #3094


bearbin:
Seems good to me :)

PR #3111, opened 2016-03-30
Never put player inside a solid when moving to a new world

LogicParrot:
This is a reopened clone of #3092.

I thought this is no longer needed, but this is needed; when the spawn point is filled with solids for some reason, this will prevent `/portal` from spawning the player underground.


LogicParrot:
Merged with #3097

Issue #2547, opened 2015-10-16
Teleportation - Server sends wrong chunks


jammet:
Using portals in the nether to return you to the normal world is delayed. When stepping into a portal on the nether side, you are seemingly first transported to the coordinates of the "other portal in the real world" without switching to the real world just yet. Instead Cuberite starts sending you the chunks that would be in the actual Nether instead of the normal world. So you'll be where you would be in the normal world, but it wastes time and bandwith by loading nether chunks of that location instead.

Edit: I split this into 3 issues. - @LogicParrot

SafwatHalaby:
I often get stuck in a teleport loop on my laptop. One portal leads to another, and before I get a chance to move, I'm teleported back to where I came from, and so on.

LogicParrot:
>Also, you still fall when the chunk isn't loaded, yet, and suffocate in some blocks. I've started to hollow out areas beneath portals to try and accommodate but it only works half the time. Nilspace tested this with me some time ago, for a while it seemed to work better, but I guess that was by chance.

Opened this in a a seperate issue: #2964

LogicParrot:
>Then, once you have arrived in the normal world, the game will act as if you had been standing in the nether gate of that side for some time (loading time), and quite often, you are transported back before you even see the normal world. This can create an infinite loop. 

Opened this in a seperate issue: #2963

jammet:
Thank you for re-organising those. :)

LogicParrot:
You're welcome :) 
I'll be working on #2963 and #2964 soon.

LogicParrot:
@jammet By editing this post you've removed the one issue I left here. (It had to do with server sending chunks of the wrong world).

LogicParrot:
Edit: My bad, all good, nothing is wrong with the issue :)

LogicParrot:
It appears that #3109 makes this more common. I believe this issue only occurs (or occurs more often) if the source and destination coords are not identical. This could indicate some stale values in cClientHandler's `m_sentChunks`. Just a wild guess.

LogicParrot:
Might be somewhat related to #3182 , to `cChunkSender`, specifically.

Issue #2405, opened 2015-07-28
Walls of random blocks in loading chunks


bramvdbogaerde:
When chuncks are loading it displays walls of leaves. Is the normal behaviour? 

This is how it looks like:
![2015-07-28_21 26 03](https://cloud.githubusercontent.com/assets/2016763/8941420/b17ff53e-356f-11e5-974e-c9ff1dbce0dc.png)

Could this be caused by a high load on the server CPU?

LO1ZB:
Yes, it could be, I got that error with some crazy debug configuration too a few weeks ago.

worktycho:
That is a bug if high cpu can cause that.

madmaxoft:
That's weird, we already had this once and it had been fixed. It had been caused by sending SetBlock / SetBlocks packets for chunks that haven't been sent to the client yet.

bramvdbogaerde:
What commit was that? I could be possible that i'm running a slightly older version.

NiLSPACE:
It's been a while, maybe end of 2014, so that's probably not the issue,

bramvdbogaerde:
Yes indeed, my build was from a week ago.

SphinxC0re:
Can we close this. The issue seems to be solved

NiLSPACE:
Are you sure? For me it usually happens when I just transported to a different world (and back).

SphinxC0re:
Do we know where this comes from?

madmaxoft:
It might be caused by teleporting to another world of the same dimension.

The previous cause of the leaves walls was that blockchange packets were sent to a client even when they didn't have the respective chunk loaded. We added a check that only sends those packets for chunks already sent to the client; perhaps this check doesn't get reset when teleporting to another world?

LO1ZB:
I had this after I rejoint after I timed out due to a exreamly overloaded world-thread.

jammet:
I've had this several times, too. Usually when using nethergates to travel back and forth. That is still a sluggish process (which reported about sometime before). NilSPACE might have seen it, too, when he was on my server to test it.

![2015-07-06_13 39 30](https://cloud.githubusercontent.com/assets/11940735/9658231/33e0e1c4-5248-11e5-97b7-e9b6617f6074.png)
![2015-07-05_10 26 57](https://cloud.githubusercontent.com/assets/11940735/9658232/370ece92-5248-11e5-9810-d93d7bb4c8e1.png)
[...]

PR #3113, opened 2016-03-30
Improve cPlayer::DoMoveToWorld

LogicParrot:
- Sometimes, `DoMoveToWorld` is called outside the tick thread. For thread safety, I moved everything to the world's task queue. This allows thread protection via the `cWorld::m_CSTasks` lock. 

- ~~I also removed a line in the protocol which has a comment about client stability. It eliminated the wild player jitter that sometimes occurs when calling `/portal`.~~ Replaced by #3351.

- ~~I added a call to `cPlayer::GetClientHandle()->RemoveFromAllChunks()`. I believe this, in combination with thread safety, might have eliminated #2547 and/or #2405, but please test this.~~

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1780,64 +1780,71 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 		return false;
 	}
 
-	//  Ask the plugins if the player is allowed to changing the world
+	//  Ask the plugins if the player is allowed to change the world
 	if (cRoot::Get()->GetPluginManager()->CallHookEntityChangingWorld(*this, *a_World))
 	{
-		// A Plugin doesn't allow the player to changing the world
+		// A Plugin doesn't allow the player to change the world
 		return false;
 	}
 
-	// The clienthandle caches the coords of the chunk we're standing at. Invalidate this.
-	GetClientHandle()->InvalidateCachedSentChunk();
-
-	// Prevent further ticking in this world
-	SetIsTicking(false);
-
-	// Tell others we are gone
-	GetWorld()->BroadcastDestroyEntity(*this);
-
-	// Remove player from world
-	GetWorld()->RemovePlayer(this, false);
-
-	// Set position to the new position
-	SetPosition(a_NewPosition);
-	FreezeInternal(a_NewPosition, false);
-
-	// Stop all mobs from targeting this player
-	StopEveryoneFromTargetingMe();
-
-	// Send the respawn packet:
-	if (a_ShouldSendRespawn && (m_ClientHandle != nullptr))
+	GetWorld()->QueueTask([this, a_World, a_ShouldSendRespawn, a_NewPosition](cWorld & a_OldWorld)
 	{
-		m_ClientHandle->SendRespawn(a_World->GetDimension());
-	}
+		// The clienthandle caches the coords of the chunk we're standing at. Invalidate this.
+		GetClientHandle()->InvalidateCachedSentChunk();
 
-	// Update the view distance.
-	m_ClientHandle->SetViewDistance(m_ClientHandle->GetRequestedViewDistance());
+		// Prevent further ticking in this world
+		SetIsTicking(false);
 
-	// Send current weather of target world to player
-	if (a_World->GetDimension() == dimOverworld)
-	{
-		m_ClientHandle->SendWeather(a_World->GetWeather());
-	}
+		// Tell others we are gone
+		GetWorld()->BroadcastDestroyEntity(*this);
 
-	// Broadcast the player into the new world.
-	a_World->BroadcastSpawnEntity(*this);
+		// Remove player from world
+		GetWorld()->RemovePlayer(this, false);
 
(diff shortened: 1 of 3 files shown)


tigerw:
The fixes branch, despite the disapproval of the anti-huge-commits committee, has relevance to this. *If* I can get it merged, could you have a look at this issue based on those changes? I'm dreading the merge conflicts.

LogicParrot:
Okay. Could you open a PR for the sake of convenient commenting?

LogicParrot:
@tigerw Our changes definitely overlap. Give me some time. I'll thoroughly review it :)

LogicParrot:
But please try splitting the changes next time. Maybe I over-split, but you cerainly over-combine. That's precisely why you keep getting conflicts.

LogicParrot:
Closing. #3115 Drastically changes this code anyways.

LogicParrot:
@tigerw I think I still want this merged. Your threading PR will take a while, and I'd like maximum stability in the meantime.  Feel free to just override the whole changes when you merge your PRs, so conflicts shouldn't be a problem.

src/Protocol/Protocol18x.cpp
```diff
@@ -1075,7 +1075,7 @@ void cProtocol180::SendRespawn(eDimension a_Dimension, bool a_ShouldIgnoreDimens
 	if ((m_LastSentDimension == a_Dimension) && !a_ShouldIgnoreDimensionChecks)
 	{
 		// Must not send a respawn for the world with the same dimension, the client goes cuckoo if we do (unless we are respawning from death)
-		return;
```
worktycho: Based on @Pokechu22's comment in #3114, https://github.com/cuberite/cuberite/pull/3114#discussion_r60810597, I don't think we can remove this line. However I do think we should follow his advice, and send two respawns in this situation, one to a different dimension, and one back. We should also rename `a_ShouldIgnoreDimensionChecks` to `a_SpawningInSameWorld` and make the change in cProtocol17x.cpp as well.

worktycho: How about you change this line to `SendRespawn(a_Dimension != eDimension::dimOverworld ? eDimension::dimOverworld : eDimension::dimNether, false);` That should handle #3114.

LogicParrot:
Todo update this to take #3351 into account

src/Protocol/Protocol18x.cpp
LogicParrot: Problem solved in #3351

src/Entities/Player.cpp
```diff
@@ -1771,64 +1771,79 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
+		// Send the respawn packet:
+		if (a_ShouldSendRespawn && (m_ClientHandle != nullptr))
+		{
+			m_ClientHandle->SendRespawn(a_World->GetDimension());
+		}
+		// Send the respawn packet:
+		cClientHandle * ch = this->GetClientHandle();
+		if (ch != nullptr)
+		{
+			ch->RemoveFromAllChunks();
```
LogicParrot: Should I really call this?

LogicParrot:
This has been updated and awaits review.

tigerw:
Note that if DoMoveToWorld is being called from another thread (ugh), then neither IsTicking nor GetWorld are atomic operations.

As always, #you-know-what

LogicParrot:
>Note that if DoMoveToWorld is being called from another thread (ugh), then neither IsTicking nor GetWorld are atomic operations. 

Indeed. I worked under the imperfect assumption that the only code paths that modify GetWorld or IsTicking are entered *after or during *doMoveToWorld*. The assumption holds in almost all (or possibly all), so thread issues are mostly not present.

There could still be some pathological cases (destruction and DoMoveToWorld at the same time?) that can cause changes to world/isTicking at the same time from two threads. In any case, this is safer than before, at least it resolves sane cases.

>Additionally, if plugins get that HookChangedWorld with a world that they aren't currently executing in, they can do even more undefined things.

I don't think this ever happens here.

LogicParrot:
>Note that if DoMoveToWorld is being called from another thread (ugh)

For the record, I think it was a plugin-related thread when invoking teleport via command.

LogicParrot:
Removed call to `RemoveFromAllChunks`, it is redundant. `RemoveFromWorld` does this job and is called by `cWorld::RemovePlayer` eventually.

src/Entities/Player.cpp
```diff
@@ -1780,64 +1780,78 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
-	StopEveryoneFromTargetingMe();
+		// Stop all mobs from targeting this player
+		StopEveryoneFromTargetingMe();
 
-	// Send the respawn packet:
-	if (a_ShouldSendRespawn && (m_ClientHandle != nullptr))
-	{
-		m_ClientHandle->SendRespawn(a_World->GetDimension());
-	}
+		// Send the respawn packet:
```
LogicParrot: Dupe

PR #3114, opened 2016-03-31
Delay chunk send on multiworld travel

LogicParrot:
The chunks of the old world fail to unload client side if chunks in the new world which have the same coords are sent too fast right after the unload. I don't know if this is a client or server issue.

Because of that, players often see chunks of an old world when traveling to a new world.

This mostly fixes it, but it uses a timer, which means it may not work in high-latency scenarios. In the future, the root cause should be found and a more elegant solution should be made.

Do not merge before #3097 ; since this properly unloads chunks, players are very likely to fall into the void while teleporting and #3097 prevents that.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -63,6 +63,7 @@ int cClientHandle::s_ClientCount = 0;
 cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
 	m_CurrentViewDistance(a_ViewDistance),
 	m_RequestedViewDistance(a_ViewDistance),
+	m_ChunkSendDelay(0),
 	m_IPString(a_IPString),
 	m_Player(nullptr),
 	m_HasSentDC(false),
@@ -95,7 +96,6 @@ cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
 	s_ClientCount++;  // Not protected by CS because clients are always constructed from the same thread
 	m_UniqueID = s_ClientCount;
 	m_PingStartTime = std::chrono::steady_clock::now();
-
 	LOGD("New ClientHandle created at %p", static_cast<void *>(this));
 }
 
@@ -1863,6 +1863,12 @@ void cClientHandle::RemoveFromWorld(void)
 	m_LastStreamedChunkZ = 0x7fffffff;
 
 	m_HasSentPlayerChunk = false;
+
+	// Do not send chunks for 1.5 seconds.
+	// Apparently, the chunks of the old world fail to unload if
+	// chunks in the new world which have the same coords are sent too fast after the unload.
+	// this is a temp fix, find the root cause!
+	m_ChunkSendDelay = 30;
 }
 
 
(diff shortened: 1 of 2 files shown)


tigerw:
The fixes branch, despite the disapproval of the anti-huge-commits committee, has relevance to this. If I can get it merged, could you have a look at this issue based on those changes? I'm dreading the merge conflicts.

LogicParrot:
Closing, as @tigerw 's `fixes` branch drastically changes this code anyways.

src/ClientHandle.cpp
```diff
@@ -1863,6 +1863,12 @@ void cClientHandle::RemoveFromWorld(void)
 	m_LastStreamedChunkZ = 0x7fffffff;
 
 	m_HasSentPlayerChunk = false;
+
+	// Do not send chunks for 1.5 seconds.
+	// Apparently, the chunks of the old world fail to unload if
+	// chunks in the new world which have the same coords are sent too fast after the unload.
+	// this is a temp fix, find the root cause!
+	m_ChunkSendDelay = 30;
```
LogicParrot: @Pokechu22 perhaps you can shed lights on what's going on here? You seem to be pretty knowlegable in Vanilla behavior.

LogicParrot: Without this PR, when teleporting to a world of the same type and to the same coords, the chunks of the old world remain visible, until they are slowly replaced by the chunks of the new world, despite the fact that Cuberite tells the client to unload those chunks. Waiting a bit prior to sending the new chunks resolves it.

Pokechu22: Well...  The client's supposed to unload the chunks when it changes dimensions or when respawning.  However, there _is_ a special case, which occurs with multiworld, regarding the [respawn packet](http://wiki.vg/Protocol#Respawn), given that players neither die nor respawn:

> <h3>Respawn</h3>
> 
> To change the player's dimension (overworld/nether/end), send them a respawn packet with the appropriate dimension, followed by prechunks/chunks for the new dimension, and finally a position and look packet. You do not need to unload chunks, the client will do it automatically. 
> 
> > **Warning**: Avoid changing player's dimension to same dimension they were already in unless they are dead. If you change the dimension to one they are already in, weird bugs can occur, such as the player being unable to attack other players in new world (until they die and respawn). 
> 
> > If you must respawn a player in the same dimension without killing them, send two respawn packets, one to a different world and then another to the world you want. You do not need to complete the first respawn; it only matters that you send two packets. 

Code-wise, the client only unloads chunks and resets the world if the dimension changes.  From `NetHandlerPlayClient`:

```java
    @Override
    public void handleRespawn(S07PacketRespawn packetIn) {
        PacketThreadUtil.checkThreadAndEnqueue(packetIn, this, this.gameController);

        if (packetIn.getDimensionID() != this.gameController.thePlayer.dimension) {
            this.doneLoadingTerrain = false;
            Scoreboard var2 = this.clientWorldController.getScoreboard();
            this.clientWorldController = new WorldClient(this, new WorldSettings(0L, packetIn.getGameType(), false, this.gameController.theWorld.getWorldInfo().isHardcoreModeEnabled(), packetIn.getWorldType()), packetIn.getWorldType(), packetIn.getDifficulty(), this.gameController.mcProfiler);
            this.clientWorldController.setWorldScoreboard(var2);
            this.gameController.loadWorld(this.clientWorldController);
            this.gameController.thePlayer.dimension = packetIn.getWorldType();
            this.gameController.displayGuiScreen(new GuiDownloadTerrain(this));
        }

        this.gameController
        .setDimensionAndSpawnPlayer(packetIn.getWorldType());
        this.gameController.playerController.setGameType(packetIn.getGameType());
    }
```

worktycho: Well, we aren't sending a respawn packet at all in same dimension multiworld, so that might have something to do with it. What we do instead is send unload chunk packets, then reload chunk packets. Clearly the client is not processing them in a queue like we expect.

worktycho: Perhaps changing the logic that filters out same dimension re-spawns to respawn to a different dimension and back might be a better approach.

LogicParrot: @Pokechu22 thank you for clearing things up. I'll re-do this properly. 

>Perhaps changing the logic that filters out same dimension re-spawns to respawn to a different dimension and back might be a better approach.

Probably.

LogicParrot:
Replaced by #3351

Issue #2835, opened 2015-12-29
Segfault cProtocol180::SendChunkData


Seadragon91:
I get sometimes this segfault on the raspberry pi 2. The memory usage was only around 50 MB. I self compiled the build on the raspberry pi.

```
(gdb) bt
#0  0x00000000 in ?? ()
#1  0x00848784 in cProtocol180::SendChunkData (this=0x712bc3b0, a_ChunkX=2,
    a_ChunkZ=0, a_Serializer=...)
    at /home/lukas/cpp/cuberite/src/Protocol/Protocol18x.cpp:287
#2  0x0085b300 in cProtocolRecognizer::SendChunkData (this=0x6f815ee0,
    a_ChunkX=2, a_ChunkZ=0, a_Serializer=...)
    at /home/lukas/cpp/cuberite/src/Protocol/ProtocolRecognizer.cpp:164
#3  0x00608918 in cClientHandle::SendChunkData (this=0x6f815c40, a_ChunkX=2,
    a_ChunkZ=0, a_Serializer=...)
    at /home/lukas/cpp/cuberite/src/ClientHandle.cpp:2214
#4  0x005f4a70 in cChunkSender::SendChunk (this=0x75525f98, a_ChunkX=2,
    a_ChunkZ=0, a_Clients=std::unordered_set with 1 elements = {...})
    at /home/lukas/cpp/cuberite/src/ChunkSender.cpp:265
#5  0x005f46dc in cChunkSender::Execute (this=0x75525f98)
    at /home/lukas/cpp/cuberite/src/ChunkSender.cpp:209
#6  0x006e2d08 in cIsThread::DoExecute (this=0x75525f98)
    at /home/lukas/cpp/cuberite/src/OSSupport/IsThread.cpp:74
#7  0x006e46e0 in std::_Mem_fn<void (cIsThread::*)()>::operator()<, void>(cIsThr                                   ead*) const (this=0x24c1104, __object=0x75525f98)
    at /usr/include/c++/4.9/functional:569
#8  0x006e4610 in std::_Bind_simple<std::_Mem_fn<void (cIsThread::*)()> (cIsThre                                   ad*)>::_M_invoke<0u>(std::_Index_tuple<0u>) (this=0x24c1100)
    at /usr/include/c++/4.9/functional:1700
[...]

PR #3115, opened 2016-03-31
Improved threading model

SafwatHalaby:
This is a branch by @tigerw.
Opening a PR for the sake of convenient commenting. See #3113
- Hopefully alleviated many deadlocks/crashes/unclear ownership relations/etc.
- Miscellaneous fixes

---

LogicParrot comments (Continously updated):

Code comments:
- DoMoveToWorld API call - You have deprecated this and now it shows a warning. In the new version, the "destination position is mandatory". Why so? In current master, not specifying destination position leads to the spawnpoint, a perfectly useful use case.
- Potentially too many world change related variables - See my first comment in this thread. This is a burden from the current master, but you've added an additional variable, and there's a chance we can actually eliminate some redundancy.

Ingame testing comments:
- The branch fails the "wolf test"; tamed Wolves have intricate interactions with mobs and players, and trying them is a somewhat strict test. Playing around with tamed wolves currently crashes the server.
- The server crashes when breaking a block in survival mode.
- The server crashes when a skeleton kills a zombie.
- Nether portals do not work.
- Logging out in survival mode while targeted by a zombie crashes the server.

*Note that crashes include debugmode assertion failures.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -41,7 +41,6 @@ cEntity::cEntity(eEntityType a_EntityType, double a_X, double a_Y, double a_Z, d
 	m_LastPosition(a_X, a_Y, a_Z),
 	m_EntityType(a_EntityType),
 	m_World(nullptr),
-	m_IsWorldChangeScheduled(false),
 	m_IsFireproof(false),
 	m_TicksSinceLastBurnDamage(0),
 	m_TicksSinceLastLavaDamage(0),
@@ -52,6 +51,7 @@ cEntity::cEntity(eEntityType a_EntityType, double a_X, double a_Y, double a_Z, d
 	m_IsSubmerged(false),
 	m_AirLevel(0),
 	m_AirTickTimer(0),
+	m_PortalCooldownData({0, false, true}),
 	m_TicksAlive(0),
 	m_IsTicking(false),
 	m_ParentChunk(nullptr),
@@ -77,9 +77,6 @@ cEntity::cEntity(eEntityType a_EntityType, double a_X, double a_Y, double a_Z, d
 
 cEntity::~cEntity()
 {
-
-	// Before deleting, the entity needs to have been removed from the world, if ever added
-	ASSERT((m_World == nullptr) || !m_World->HasEntity(m_UniqueID));
 	ASSERT(!IsTicking());
 
 	/*
@@ -201,7 +198,7 @@ void cEntity::SetParentChunk(cChunk * a_Chunk)
 
 
 
-cChunk * cEntity::GetParentChunk()
+cChunk * cEntity::GetParentChunk() const
 {
 	return m_ParentChunk;
 }
@@ -1369,36 +1366,13 @@ void cEntity::DetectCacti(void)
 
 
 
-void cEntity::ScheduleMoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown)
-{
-	m_NewWorld = a_World;
-	m_NewWorldPosition = a_NewPosition;
-	m_IsWorldChangeScheduled = true;
-	m_WorldChangeSetPortalCooldown = a_SetPortalCooldown;
-}
-
-
-
-
 bool cEntity::DetectPortal()
 {
-	// If somebody scheduled a world change with ScheduleMoveToWorld, change worlds now.
-	if (m_IsWorldChangeScheduled)
+	if (!m_PortalCooldownData.m_PositionValid)
 	{
-		m_IsWorldChangeScheduled = false;
-
-		if (m_WorldChangeSetPortalCooldown)
-		{
-			// Delay the portal check.
-			m_PortalCooldownData.m_TicksDelayed = 0;
-			m_PortalCooldownData.m_ShouldPreventTeleportation = true;
-		}
-
-		MoveToWorld(m_NewWorld, false, m_NewWorldPosition);
-		return true;
+		return false;
 	}
-
-	if (GetWorld()->GetDimension() == dimOverworld)
+	else if (GetWorld()->GetDimension() == dimOverworld)
 	{
 		if (GetWorld()->GetLinkedNetherWorldName().empty() && GetWorld()->GetLinkedEndWorldName().empty())
 		{
@@ -1412,7 +1386,7 @@ bool cEntity::DetectPortal()
 		return false;
 	}
 
-	int X = POSX_TOINT, Y = POSY_TOINT, Z = POSZ_TOINT;
+	const int X = POSX_TOINT, Y = POSY_TOINT, Z = POSZ_TOINT;
 	if ((Y > 0) && (Y < cChunkDef::Height))
 	{
 		switch (GetWorld()->GetBlock(X, Y, Z))
@@ -1425,62 +1399,56 @@ bool cEntity::DetectPortal()
 					return false;
 				}
 
-				if (IsPlayer() && !(reinterpret_cast<cPlayer *>(this))->IsGameModeCreative() && (m_PortalCooldownData.m_TicksDelayed != 80))
+				if (
+					IsPlayer() &&
+					// !reinterpret_cast<cPlayer *>(this)->IsGameModeCreative() &&  // TODO: fix portal travel prevention - client sends outdated position data throwing off checks
+					(m_PortalCooldownData.m_TicksDelayed != 80)
+				)
 				{
 					// Delay teleportation for four seconds if the entity is a non-creative player
 					m_PortalCooldownData.m_TicksDelayed++;
 					return false;
 				}
-				m_PortalCooldownData.m_TicksDelayed = 0;
+
+				m_PortalCooldownData.m_ShouldPreventTeleportation = true;  // Stop portals from working on respawn
+				m_PortalCooldownData.m_PositionValid = false;
 
 				if (GetWorld()->GetDimension() == dimNether)
 				{
-					if (GetWorld()->GetLinkedOverworldName().empty())
+					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedOverworldName());
+					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
+
+					if (GetWorld()->GetLinkedOverworldName().empty() || !OnPreWorldTravel(*TargetWorld))
 					{
 						return false;
-					}
-
-					m_PortalCooldownData.m_ShouldPreventTeleportation = true;  // Stop portals from working on respawn
-
-					if (IsPlayer())
-					{
-						// Send a respawn packet before world is loaded / generated so the client isn't left in limbo
-						(reinterpret_cast<cPlayer *>(this))->GetClientHandle()->SendRespawn(dimOverworld);
 					}
 
 					Vector3d TargetPos = GetPosition();
 					TargetPos.x *= 8.0;
 					TargetPos.z *= 8.0;
 
-					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedOverworldName());
-					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
 					LOGD("Jumping nether -> overworld");
-					new cNetherPortalScanner(this, TargetWorld, TargetPos, 256);
+					new cNetherPortalScanner(GetUniqueID(), GetWorld()->GetDimension(), TargetWorld, TargetPos, 256);
+
 					return true;
 				}
 				else
 				{
-					if (GetWorld()->GetLinkedNetherWorldName().empty())
+					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedNetherWorldName());
+					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
+
+					if (GetWorld()->GetLinkedNetherWorldName().empty() || !OnPreWorldTravel(*TargetWorld))
 					{
 						return false;
-					}
-
-					m_PortalCooldownData.m_ShouldPreventTeleportation = true;
-
-					if (IsPlayer())
-					{
-						reinterpret_cast<cPlayer *>(this)->AwardAchievement(achEnterPortal);
-						reinterpret_cast<cPlayer *>(this)->GetClientHandle()->SendRespawn(dimNether);
 					}
 
 					Vector3d TargetPos = GetPosition();
 					TargetPos.x /= 8.0;
 					TargetPos.z /= 8.0;
 
-					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedNetherWorldName());
-					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
 					LOGD("Jumping overworld -> nether");
-					new cNetherPortalScanner(this, TargetWorld, TargetPos, 128);
+					new cNetherPortalScanner(GetUniqueID(), GetWorld()->GetDimension(), TargetWorld, TargetPos, 128);
+
 					return true;
 				}
 			}
@@ -1491,6 +1459,9 @@ bool cEntity::DetectPortal()
 					return false;
 				}
 
+				m_PortalCooldownData.m_ShouldPreventTeleportation = true;
+				m_PortalCooldownData.m_PositionValid = false;
+
 				if (GetWorld()->GetDimension() == dimEnd)
 				{
 
@@ -1499,18 +1470,10 @@ bool cEntity::DetectPortal()
 						return false;
 					}
 
-					m_PortalCooldownData.m_ShouldPreventTeleportation = true;
-
-					if (IsPlayer())
-					{
-						cPlayer * Player = reinterpret_cast<cPlayer *>(this);
-						Player->TeleportToCoords(Player->GetLastBedPos().x, Player->GetLastBedPos().y, Player->GetLastBedPos().z);
-						Player->GetClientHandle()->SendRespawn(dimOverworld);
-					}
-
 					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedOverworldName());
 					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
-					return MoveToWorld(TargetWorld, false);
+
+					return MoveToWorld(*TargetWorld, Vector3d(TargetWorld->GetSpawnX(), TargetWorld->GetSpawnY(), TargetWorld->GetSpawnZ()));
 				}
 				else
 				{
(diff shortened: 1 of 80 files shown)


LogicParrot:
I've been thinking: We have m_ParentChunk, m_IsTicking, and now, m_IsChangingWorlds. Aren't these all essentially the same?

I remember that in my initial code, I did not need m_IsTicking. When a client is travelling worlds, I just set m_ParentChunk to null. But for some reason I later needed m_IsTicking. I cannot remember why right now.

LogicParrot:
Note that `m_IsTicking` and `m_IsChanging` worlds seem to always be opposities, except, perhaps, in the short transition period when travelling worlds.

Also, since entities in unloaded chunks do not tick, `m_IsTicking` is actually `m_IsInsideAWorldAndHaveAParentChunk` or so (except, perhaps, in the short transition period when travelling worlds :P ).

LogicParrot:
And sorry if it's messy. I duct taped world travel into stability with the little time I had, but perhaps it's time to do it properly now.

src/Entities/Entity.cpp
```diff
@@ -1835,6 +1702,131 @@ void cEntity::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
...
+				{
+					if (!a_Entity->IsMob())
+					{
+						return false;
+					}
+
+					auto Monster = static_cast<cMonster *>(a_Entity);
+					if (std::equal_to<std::weak_ptr<cPawn>>()(Monster->GetWeakTargetPtr(), m_Pawn))
+					{
+						Monster->SetTarget(nullptr);
```
LogicParrot: - This stops everyone from targeting this entity. I think you need something similar at `TeleportToCoords` and at `cEntity::Destroy `.
 - You got rid of the `StopEveryoneFromTargetingMe` and friends due to safety concerns, and that's understandable. But now entity  destruction costs O(Number of entities in world) rather than the previous O(Number of entities targeting me). But I guess `Stability and readability > Performance`, so I suppose it's fine for now.

LogicParrot: Sorry. My first point is invalid since this is a `weak_ptr`.

LogicParrot: This method can be dangerous, `ForEachEntity` is not guaranteed to catch all entities (unless you changed that?), an entity in an invalid chunk won't be caught.

How about copying the object to the new world instead?

tigerw: > You got rid of the  StopEveryoneFromTargetingMe  and friends due to safety concerns, and that's understandable. But now entity destruction costs O(Number of entities in world) rather than the previous O(Number of entities targeting me). But I guess  Stability and readability > Performance , so I suppose it's fine for now.

I was thinking that since entity world moving is typically rare, more benefits would be seen by not needing to maintain a list of targeting entities, and moving that processing to a one-time iteration when moving worlds.

Dunno about ForEachEntity - if it doesn't catch all then I have a bug :)

Where should I perform this copy, and to what object?

LogicParrot: >I was thinking that since entity world moving is typically rare, more benefits would be seen by not needing to maintain a list of targeting entities, and moving that processing to a one-time iteration when moving worlds.

You are correct. When I first made my comment, I was still fixated on the old code, in which destruction also required iterating pointers. World travel is rare, destruction is not.

> Dunno about ForEachEntity - if it doesn't catch all then I have a bug :)
> Where should I perform this copy, and to what object?

Entities in invalid chunks are missed. See `cChunkMap::cChunkLayer::ForEachEntity`, not the `&& m_Chunks[i]->IsValid();`. There are two possible resolutions:

1. Establishing a rigid policy of what we expect of invalid chunks to do and not to do, and consistently change all relevant functions such as `cChunkMap::cChunkLayer::ForEachEntity`.

2. I was thinking that perhaps copying the entity to the new world and then destroying the entity in the old world rids you of this whole iteration.

LogicParrot: >I was thinking that since entity world moving is typically rare, more benefits would be seen by not needing to maintain a list of targeting entities, and moving that processing to a one-time iteration when moving worlds.

You are correct. When I first made my comment, I was still fixated on the old code, in which destruction also required iterating pointers. World travel is rare, destruction is not.

> Dunno about ForEachEntity - if it doesn't catch all then I have a bug :)
> Where should I perform this copy, and to what object?

Entities in invalid chunks are missed. See `cChunkMap::cChunkLayer::ForEachEntity`, and note the `&& m_Chunks[i]->IsValid();`. There are two possible resolutions:

1. Establishing a rigid policy on what we expect of invalid chunks to do and not to do, and consistently change all relevant functions such as `cChunkMap::cChunkLayer::ForEachEntity`, and make `ForEachEntity` iterate everything.

2. I was thinking that perhaps copying the entity to the new world and then destroying the entity in the old world rids you of this whole iteration.

tigerw:
Ah cmake, you can go put your custom commands in a pipe, and smoke on it. I'm *done*.

tigerw:
I really gotta get this merged soon. It's blocking stuff.

...halp cmake hates me :(

worktycho:
I can't comment on the line because of the size of this commit, but World.h line 376 looks very fishy, as you are opening a cIniFile which is already opened in the method which called it. Multiple cInIFile instances pointing at the same file is extremely dangerous.

tigerw:
ty...tycho? Is that you? You're back? You're not dead‽ Where have you been! We were lost without your expertise and standards compliance enforcement! Lost!

tigerw:
What do you think are the chances that everything goes horribly wrong if I thrown caution to the wind and just merge it?

Seadragon91:
Take a look into the first log of travis, something bad happened there
```
[21:59:51] Cannot serialize chunk [-10, 5] into data
[21:59:51] Cannot get chunk [-9, 5] data for NBT saving
```

Seadragon91:
Rebased and tried to compile it with clang 3.5, doesn't compile.

LogicParrot:
>What do you think are the chances that everything goes horribly wrong if I thrown caution to the wind and just merge it?

DONT. :fearful:

tigerw:
> Rebased and tried to compile it with clang 3.5, doesn't compile.

I think the rebase did a false merge and pulled in some changes which were incompatible. Try now?

tigerw:
> DON'T. :fearful: 

ohhh I'm gonna do it...!

Seadragon91:
Deleted the branch, pulled the branch fixes and tried to compile it with clang/gcc. Something hates me...
```
** tolua internal error: [string "tolua embedded: src/bin/lua/compat-5.1.lua"]:31: error loading file BindingsProcessor.lua.
```

tigerw:
Is there a `BindingsProcessor.lua` file in `src/Bindings`?

Also, people are testing my changes :D
Perhaps I shall not be so hasty as to click the enticing green button.

SetFlags.cmake
```diff
@@ -74,10 +74,10 @@ macro(set_flags)
...
 		endif()
 
-		set(CMAKE_CXX_FLAGS          "${CMAKE_CXX_FLAGS}          -std=c++11")
-		set(CMAKE_CXX_FLAGS_DEBUG    "${CMAKE_CXX_FLAGS_DEBUG}    -std=c++11")
-		set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_COVERAGE} -std=c++11")
-		set(CMAKE_CXX_FLAGS_RELEASE  "${CMAKE_CXX_FLAGS_RELEASE}  -std=c++11")
+		set(CMAKE_CXX_FLAGS          "${CMAKE_CXX_FLAGS}          -std=c++14")
+		set(CMAKE_CXX_FLAGS_DEBUG    "${CMAKE_CXX_FLAGS_DEBUG}    -std=c++14")
+		set(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_COVERAGE} -std=c++14")
+		set(CMAKE_CXX_FLAGS_RELEASE  "${CMAKE_CXX_FLAGS_RELEASE}  -std=c++14")
```
worktycho: I'd like to see the discussion where c++14 was agreed. It took nearly 3 years before we agreed to require c++11, and its only just 3 years since c++14 was standardized.

Seadragon91:
Testing this currently under windows...
1) The worlds are now generated all simultaneously, instead of one after the other? Is that a good idea?

tigerw:
> The worlds are now generated all simultaneously, instead of one after the other. Is that a good idea?

I adopt the stance that if one has multiple processor cores, one should make use of it.

> The spawnpoint search doesn't wok correctly.

Definitely an issue here.

> This nice error comes, while the server is running and you are connected

I dunno about this one. I will work on it more.

> In lua  MoveToWorld  now wants a  cWorld  instead of the  world name

I'll change this.

> Sorry to say. but there are many scary changes. I think the best would be to recreate that pull request and split it in smaller parts.

This exactly, has always been my problem. :(

Seadragon91:
Okay, I will dive deeper into your changes :). There are more changes that breaks lua api methods, for example `TeleportToCoords` now wants a `Vector`.

tigerw:
I wonder if I should just go through everyone's plugins and change their method calls.

SetFlags.cmake
tigerw: There has been no discussion unfortunately. However, Debian includes Clang 3.5 and GCC 4.9 as the default compilers, so there's that. It would certainly be nice to be able to initialise variables in the capture list, as offered by C++14...

Seadragon91:
Okay you changed the method, I would say just add the method TeleportToCoords with x, y, z. Then it should work :)

Seadragon91:
I still was not able to compile it under debian, tried it on two different computers :(

Seadragon91:
Here the few lines containing the error
```
[ 45%] Generating ../../../src/Bindings/Bindings.cpp, ../../../src/Bindings/Bindings.h,
../../../src/Bindings/LuaState_Declaration.inc, ../../../src/Bindings/LuaState_Implementation.cpp,
../../../src/Bindings/LuaState_Typedefs.inc

** tolua internal error: [string "tolua embedded: src/bin/lua/compat-5.1.lua"]:31: error loading file BindingsProcessor.lua.


stack traceback:
        [string "tolua: embedded Lua code 23"]:6: in main chunk
[ 45%] Built target BindingsGeneration
Scanning dependencies of target HTTPServer
[ 46%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/EnvelopeParser.cpp.o
[ 46%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/HTTPFormParser.cpp.o
[ 46%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/HTTPMessage.cpp.o
[ 46%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/HTTPMessageParser.cpp.o
[ 46%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/HTTPServer.cpp.o
[ 47%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/HTTPServerConnection.cpp.o
[ 47%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/MultipartParser.cpp.o
[ 47%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/NameValueParser.cpp.o
[ 47%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/SslHTTPServerConnection.cpp.o
[ 47%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/TransferEncodingParser.cpp.o
[ 48%] Building CXX object src/HTTP/CMakeFiles/HTTPServer.dir/UrlParser.cpp.o
[ 48%] Linking CXX static library libHTTPServer.a
[ 48%] Built target HTTPServer
Scanning dependencies of target Items
[ 48%] Building CXX object src/Items/CMakeFiles/Items.dir/ItemHandler.cpp.o
In file included from /home/lukas/cpp/cuberite/src/Items/ItemHandler.cpp:8:
In file included from /home/lukas/cpp/cuberite/src/Items/../BlockInServerPluginInterface.h:14:
In file included from /home/lukas/cpp/cuberite/src/./Root.h:9:
In file included from /home/lukas/cpp/cuberite/src/./WebAdmin.h:8:
/home/lukas/cpp/cuberite/src/./Bindings/LuaState.h:36:10: fatal error: 'LuaState_Typedefs.inc' file not found
         ^
1 error generated.
src/Items/CMakeFiles/Items.dir/build.make:62: die Regel für Ziel „src/Items/CMakeFiles/Items.dir/ItemHandler.cpp.o“ scheiterte
make[2]: *** [src/Items/CMakeFiles/Items.dir/ItemHandler.cpp.o] Fehler 1
CMakeFiles/Makefile2:1234: die Regel für Ziel „src/Items/CMakeFiles/Items.dir/all“ scheiterte
make[1]: *** [src/Items/CMakeFiles/Items.dir/all] Fehler 2
Makefile:127: die Regel für Ziel „all“ scheiterte
make: *** [all] Fehler 2
```

Seadragon91:
Okay one step further, If I use CIbuild.sh then it compiles successfully.

SetFlags.cmake
worktycho: Debain may but visual studio still only has partial support. And Centos is still shipping gcc 4.8.

worktycho: And to use capture initialisations means requiring VS 2015, which is only 6 months old.

LogicParrot:
We should not break the API.

SetFlags.cmake
worktycho: You also haven't updated the visual studio version check to reflect the change to require VS2015.

worktycho:
If you want to change methods which are exposed to lua, then you should patch the calls in manual bindings to still expose the old api.

Seadragon91:
The startup complete message should come after everything has been started/loaded.
`[13:46:15] Startup complete, took 3784ms!`

I tried to find the cause for my problem posted above. Looked for flags, into cmake list files. Have no more ideas.

SetFlags.cmake
tigerw: > You also haven't updated the visual studio version check to reflect the change to require VS2015.

I will do this if this is agreed.

-----

> visual studio still only has partial support

Update 2 says it has full support plus ratified '17 supports too. (It's great :D)

-----

My opinion is that development should not be limited by perceived availability of compiler toolsets. We have a small developer base that is unencumbered by restrictions which say, a company faces, and therefore should not try to artificially apply those restrictions unto ourselves.

Would those who wish to compile it themselves object to requiring a newer version of their compiler, that is, if they don't already have it?

tigerw:
>tolua: error loading file BindingsProcessor.lua

@Seadragon91, could you make sure that this file exists in `src/Bindings` on whatever folder you've the repository in?

tigerw:
> ...break the API.

Alright alright.

------

...this sort of mentality is what got Windows into a mess of compatibility shims and "emulated bugs" and what got companies into using software last developed ten years ago... :(

Seadragon91:
Yes the file is existent  and the permissions are okay. As i wrote above, it compiles successfully, if I use `CIbuild.sh`.

SetFlags.cmake
LogicParrot: I think almost everyone should be able to easily compile Cuberite.

Do all major Linux distros support the new compilers right through the package manager?
Do Raspberry Pis, C.H.I.Ps, and all similar popular devices support the new compilers right through the package manager?

Seadragon91:
@tigerw It's a path problem. If I run cmake inside a folder named Release, then it will fails. Running it directly in cuberite, will work fine.

tigerw:
> @tigerw It's a path problem. If I run cmake inside a directory named Release, then it will fails. Running it in directory cuberite, will work fine.

oh. *ohhh*.

You're way better at this than me, why am I doing this?

worktycho:
@tigerw Regarding backcompat, we should follow the linux model of maintaining backwards compatibility according to documented behavior, and not maintaining undocumented behavior or bugs.

tigerw:
Alright.

Seadragon91:
Not sure, if you already fixed my cmake problem. After the changes below, it compiles fine.

File `/src/CMakeLists.txt`
Lines: `196 - 200`:
Changed `${CMAKE_SOURCE_DIR}/src` to `PROJECT_SOURCE_DIR`.

Lines: `212 and 223`:
Changed `WORKING_DIRECTORY Bindings/` to `WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/Bindings/`.

Seadragon91:
Tested the kick problem on your branch, with debug build and authentication off.
```
[9a8e239deea5486a|22:52:53] Assertion failed: !IsInTickThread(), file /home/lukas/cpp/cuberite/src/Server.cpp, line 690
./Cuberite_debug(_Z15PrintStackTracev+0x1c)[0xa8df7c]
./Cuberite_debug(_ZN7cServer16AuthenticateUserEiRKSsS1_RKN4Json5ValueE+0x6d)[0xa11d4d]
./Cuberite_debug(_ZN5cRoot16AuthenticateUserEiRKSsS1_RKN4Json5ValueE+0x45)[0x9f8f05]
./Cuberite_debug(_ZN14cAuthenticator12AuthenticateEiRKSsS1_+0x94)[0xbfc1b4]
./Cuberite_debug(_ZN13cClientHandle11HandleLoginEjRKSs+0x101)[0x949721]
./Cuberite_debug(_ZN12cProtocol18022HandlePacketLoginStartER11cByteBuffer+0x2d7)[0xc1a337]
./Cuberite_debug(_ZN12cProtocol18012HandlePacketER11cByteBufferj+0x101)[0xc19561]
./Cuberite_debug(_ZN12cProtocol18015AddReceivedDataEPKcm+0xd9e)[0xc0b2de]
./Cuberite_debug(_ZN12cProtocol18012DataReceivedEPKcm+0xf4)[0xc0a534]
./Cuberite_debug(_ZN19cProtocolRecognizer12DataReceivedEPKcm+0x172)[0xc3eef2]
./Cuberite_debug(_ZN13cClientHandle25ProcessQueuedIncomingDataEv+0xae)[0x94fa0e]
./Cuberite_debug(_ZN13cClientHandle4TickEf+0x25)[0x94ffb5]
./Cuberite_debug(_ZN7cServer11TickClientsEf+0x91)[0xa100d1]
./Cuberite_debug(_ZN7cServer4TickEf+0xb5)[0xa0e555]
./Cuberite_debug(_ZN7cServer11cTickThread7ExecuteEv+0x92)[0xa0e412]
./Cuberite_debug(_ZN9cIsThread9DoExecuteEv+0x37)[0xa80ed7]
./Cuberite_debug(_ZNKSt7_Mem_fnIM9cIsThreadFvvEEclIJEvEEvPS0_DpOT_+0x69)[0xa822f9]
./Cuberite_debug(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EE9_M_invokeIJLm0EEEEvSt12_Index_tupleIJXspT_EEE+0x42)[0xa82262]
./Cuberite_debug(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EEclEv+0x15)[0xa82215]
./Cuberite_debug(_ZNSt6thread5_ImplISt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS3_EEE6_M_runEv+0x19)[0xa81f99]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8d390)[0x7fb5927ad390]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x8182)[0x7fb592a35182]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d)[0x7fb59223e47d]
Cuberite_debug: /home/lukas/cpp/cuberite/src/Server.cpp:690: void cServer::AuthenticateUser(int, const AString &, const AString &, const Json::Value &): Assertion `0' failed.
```

tigerw:
aargh

Seadragon91:
I miss the line numbers, a missing flag for debug build?

tigerw:
?

I'm lamenting how bad I am at programming in general.

Hey. At least there is an assertion for improper thread access and it didn't go on to corrupt some variables silently.

tigerw:
@Seadragon91, was that done with authentication off? For the python client?

Seadragon91:
With authentication on, the assertion doesn't occur :)

LogicParrot:
@tigerw do you think you can split this to smaller digestable pieces?

tigerw:
Do you insist, or can I make a promise that I will next time?

tigerw:
sooooooo...

On a scale from 0 to 1, where P ⊆ ℝ

1. Find P(something goes horribly wrong | the request was merged). [4]
2. Draw a probability tree of all possible outcomes. [5]

bearbin:
1. P(something goes horribly wrong) == 1 ∴∴∴∴∴∴&#8756; P(something goes horribly wrong | the request was merged) == 1

tigerw:
Basically S1. I need to revise conditional probabilities though

Seadragon91:
Compiled the branch with clang threadsanitizer, looks like there are some data races. For example in chunk generator.

LogicParrot:
>Do you insist, or can I make a promise that I will next time?

I don't insist.

LogicParrot:
Are you planning to deal with the overuse of `m_X` world change variables? (Mentioned above)

SetFlags.cmake
LogicParrot: So, as per your discussions with @worktycho elsewhere (lost the link), no C++14 for now, right?

LogicParrot:
This will be a live comment, I'll constantly edit it as I test things.

Code comments:
- DoMoveToWorld - You have deprecated this and now it shows a warning. In the new version, the "destination position is mandatory". Why so? In current master, not specifying destination position leads to the spawnpoint, a perfectly useful use case.

Ingame testing comments:
 - Stay tuned.

Seadragon91:
@LogicParrot I would suggest adding this into the first post.

LogicParrot:
Okay. will do.

Seadragon91:
@tigerw 
Please add in file [ChunkSender.cpp](https://github.com/cuberite/cuberite/blob/e57898cda75ed27757c534c55e1cc1816347f32f/src/ChunkSender.cpp) in the lines 94 and 143 the &. To avoid copy and compile errors.

SetFlags.cmake
tigerw: Conceding ground in discussions? That doesn't sound like me :P

> I think almost everyone should be able to easily compile Cuberite.

Direct them to the compile script, which does it all for them.

Without C++14, basically this entire request becomes impossible without much extra copying in inter-thread lambda captures, and thus reduced performance.

worktycho: Where would the copies be? I can't find them.

Seadragon91:
@tigerw I see that you have added a new commit, but this doesn't compile.

tigerw:
Still working on it. If I seem to keep mentioning this, I'm only having fun with LogicParrot :P

LogicParrot:
Evil.

LogicParrot prays the branch will self-split into sensible commits.

tigerw:
I removed the changes to cRoot which made a bunch of objects gain automatic duration rather than manual, since that was unexpectedly hitting a limitation of our build system, so you got part of your prayer already :)

Seadragon91:
@tigerw After changing c++11 to c++14 and a few cast fixes, I was able to compile it. I run my script and it crashed with the same problem of #2835.

https://gist.github.com/Seadragon91/59f27e0918680fcf239cbf72cce77478

LogicParrot:
I don't think tiger was serious about this branch resolving #2835 at the present time. This is still a branch in progress. He meant that when it's done, threading issues in general should be less common.

tigerw:
Still, that was a bug. @Seadragon91 would you like to try again?

Seadragon91:
Well done @tigerw :+1:. Have run my script now for 30 mins, no crash in `SendChunkData`. Can this fix be backported into master? :smiley:

LogicParrot:
Any insight that could help fix this upstream?

tigerw:
Don't celebrate yet, it may well still break somewhere else :P

The part which I think pertains to cChunkSender was to make it take std::weak_ptr<cClientHandle>. That way it can atomically acquire (or fail to acquire) an owning pointer without locks. However, I believe that required changes to how many parts of the server store clienthandles, so the reason the changeset is so utterly big was because Chunks also now store weak_ptrs, etc.

LogicParrot:
>...this sort of mentality is what got Windows into a mess of compatibility shims and "emulated bugs" and what got companies into using software last developed ten years ago... :(

In many cases we can keep the old exported functions and mark them "deprecated" and internally point them to the new functions. It gives plugin developers some time to update their code and doesn't incur maintenance cost. We can remove the deprecated functions after some time.

I'd be careful to change the *behavior* of a function without changing its name though, because that alters plugin behavior.

If there is a fundemental change for the better, and it is hard to maintain a deprecated function that uses the new function, then I personally would consider API breaking a sane option, at least at this stage of development, where we don't have a million plugins.

LogicParrot:
I don't think issues should be closed simply because of age. Last I checked, @tigerw is still working on this one. (He split some of it into #3489)

tigerw:
Plan:

1. Revert changes moving protocol actions out of the tick thread (a la #2477). The added complexity was too much.
2. Unique-ptr entities was merged into master, no need for it here.
3. Embarassingly, the cNetherPortalScanner crash fix was sitting in this branch. Tracked in #4623.
4. Move the LightingThread changes to another branch.

Issue #3110, opened 2016-03-30
Reduce RAM and disk usage by minimizing the use of cChunk::MarkDirty


LogicParrot:
When a block in a chunk changes, the chunk is marked dirty. Dirty chunks are saved to disk within 5 minutes. During those 5 minutes, a dirty chunk cannot unload, which makes it sit in RAM. For maximum performance, especially on IO-limited and/or RAM-limited devices like the Raspberry Pi, we should minimize the use of `cChunk::MarkDirty`. Below is a list of places where I have identified a potential overuse of `cChunk::MarkDirty`.

- [ ] [BlockLeaves.h#L120](https://github.com/cuberite/cuberite/blob/master/src/Blocks/BlockLeaves.h#L120).
  Leaves occasionally call this. It sets a meta, which marks the chunk dirty without blocks actually changing.
  Also, it sends an update to the client. Do we really need to send this meta update to the client?
  Possible solution: Creating a new Method: `cChunk::SetMeta_WithoutMarkingDirty` which is used for setting metas whose values are not too important and do not have to be saved to disk. or perhaps `cChunk::SetMeta_WithoutMarkingDirty_AndWithSendingToClient`. Feel free to find shorter names ;)
- [ ] Grass is moving underground. As the grass decays or spawns, `MarkDirty` is called. #2850 should be fixed.
- [ ] [FloodyFluidSimulator.cpp#L130](https://github.com/cuberite/cuberite/blob/master/src/Simulator/FloodyFluidSimulator.cpp#L130) calls markDirty without any blocks visually changing. Again, should this meta update really be saved to disk? Should it be sent to the client?
- [ ] Ghost pickups, described in #3106, may be calling `MarkDirty`.

LogicParrot:
I will update the list as I find more.

tigerw:
Entities that are part of a chunk that is loaded from disk, still end up calling Initialize, which will queue an add one tick later. This results in SetDirty on every newly loaded chunk.

We need to rethink this whole two-part entity initialisation. An entity should be created with all the properties it needs, either defaulted or loaded from Anvil. A normal spawn still goes through the cWorld::AddEntity queue, a SetChunkData "spawn" should be special and the entity just added to the chunk's m_Entities directly.

PR #3129, opened 2016-04-06
Updated cChunk::SetMeta, fixed grass growth

### LINKED ISSUE #2850: Grass spreads underground
This only happens when it's day (not with torches). You can see it happening by executing a simple script like this:
```lua
function OnBlockSpread(...)
	print(...)
end

cPluginManager:AddHook(cPluginManager.HOOK_BLOCK_SPREAD, OnBlockSpread)
```

LogicParrot:
This PR resolves the grass underground spread issue  (#2850), also, it stops leaves decay checks from sending needless packets to the client and needless writing to disk. This is part of the solution for #3110.

Closes #2850.

### DIFF
--- a/src/BlockInfo.cpp
+++ b/src/BlockInfo.cpp
@@ -250,7 +250,6 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
 	a_Info[E_BLOCK_REDSTONE_WIRE       ].m_Transparent = true;
 	a_Info[E_BLOCK_SAPLING             ].m_Transparent = true;
 	a_Info[E_BLOCK_SIGN_POST           ].m_Transparent = true;
-	a_Info[E_BLOCK_SNOW                ].m_Transparent = true;
 	a_Info[E_BLOCK_SPRUCE_DOOR         ].m_Transparent = true;
 	a_Info[E_BLOCK_SPRUCE_FENCE        ].m_Transparent = true;
 	a_Info[E_BLOCK_SPRUCE_FENCE_GATE   ].m_Transparent = true;
--- a/src/Blocks/BlockDirt.h
+++ b/src/Blocks/BlockDirt.h
@@ -51,8 +51,8 @@ public:
 		{
 			BLOCKTYPE above = a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ);
 
-			// Grass turns back to dirt when the block above is not transparent
-			if (!cBlockInfo::IsTransparent(above))
+			// Grass turns back to dirt when the block above it is not transparent or water
+			if (!cBlockInfo::IsTransparent(above) || IsBlockWater(above))
 			{
 				a_Chunk.FastSetBlock(a_RelX, a_RelY, a_RelZ, E_BLOCK_DIRT, E_META_DIRT_NORMAL);
 				return;
@@ -97,10 +97,13 @@ public:
 				// Not a regular dirt block
 				continue;
 			}
-			BLOCKTYPE above = a_Chunk.GetBlock(BlockX, BlockY + 1, BlockZ);
-			NIBBLETYPE light = std::max(a_Chunk.GetBlockLight(BlockX, BlockY + 1, BlockZ), a_Chunk.GetTimeAlteredLight(a_Chunk.GetSkyLight(BlockX, BlockY + 1, BlockZ)));
-			// Grass does not spread to blocks with a light level less than 5
-			if ((light > 4)  && cBlockInfo::IsTransparent(above))
+			BLOCKTYPE above = Chunk->GetBlock(BlockX, BlockY + 1, BlockZ);
+			NIBBLETYPE light = std::max(Chunk->GetBlockLight(BlockX, BlockY + 1, BlockZ), Chunk->GetTimeAlteredLight(Chunk->GetSkyLight(BlockX, BlockY + 1, BlockZ)));
+			if ((light > 4)  &&
+				cBlockInfo::IsTransparent(above) &&
+				(!IsBlockLava(above)) &&
+				(!IsBlockWaterOrIce(above))
+			)
 			{
 				if (!cRoot::Get()->GetPluginManager()->CallHookBlockSpread(*Chunk->GetWorld(), Chunk->GetPosX() * cChunkDef::Width + BlockX, BlockY, Chunk->GetPosZ() * cChunkDef::Width + BlockZ, ssGrassSpread))
 				{
--- a/src/Chunk.h
+++ b/src/Chunk.h
@@ -392,15 +392,24 @@ public:
 	{
 		return m_ChunkData.GetMeta(a_RelX, a_RelY, a_RelZ);
 	}
-	inline void       SetMeta(int a_RelX, int a_RelY, int a_RelZ, NIBBLETYPE a_Meta)
+
+	/** Set a meta value, with the option of not informing the client and / or not marking dirty.
+	Used for setting metas that are of little value for saving to disk and / or for sending to the client,
+	such as leaf decay flags. */
+	inline void SetMeta(int a_RelX, int a_RelY, int a_RelZ, NIBBLETYPE a_Meta, bool a_ShouldMarkDirty = true, bool a_ShouldInformClient = true)
 	{
 			bool hasChanged = m_ChunkData.SetMeta(a_RelX, a_RelY, a_RelZ, a_Meta);
 			if (hasChanged)
 			{
-				MarkDirty();
 				m_IsRedstoneDirty = true;
-
-				m_PendingSendBlocks.push_back(sSetBlock(m_PosX, m_PosZ, a_RelX, a_RelY, a_RelZ, GetBlock(a_RelX, a_RelY, a_RelZ), a_Meta));
+				if (a_ShouldMarkDirty)
+				{
+					MarkDirty();
+				}
+				if (a_ShouldInformClient)
+				{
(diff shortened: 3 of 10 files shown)


src/Chunk.h
```diff
@@ -392,18 +392,37 @@ class cChunk :
 	{
 		return m_ChunkData.GetMeta(a_RelX, a_RelY, a_RelZ);
 	}
-	inline void       SetMeta(int a_RelX, int a_RelY, int a_RelZ, NIBBLETYPE a_Meta)
+	inline void  SetMeta(int a_RelX, int a_RelY, int a_RelZ, NIBBLETYPE a_Meta)
```
sgtbigman: With the added `SetMetaQuiet` method, would it be worth having the `SetMeta` method just call `SetMetaQuiet` with the boolean arguments set to `true`?

LogicParrot: I've been wondering about this too. My reasoning was that `SetMeta` is a heavily used function, and for the sake of speed, one should skip the two extra ifs of `SetMetaQuiet` when possible. I have no idea if this argument holds in practice.

LogicParrot: >With the added SetMetaQuiet method, would it be worth having the SetMeta method just call SetMetaQuiet with the boolean arguments set to true?

- Or, Just add the two extra parameters to the regular `SetMeta` and make them default to true, this eliminates the need for metaQuiet -

sgtbigman: Good point. It may not affect speed, but something worth considering.

sgtbigman: > -Or, Just add the two extra parameters to the regular SetMeta and make them default to true, this eliminates the need for metaQuiet-

Right, I was thinking of that when I made my original comment. It may be the better route in this case.

LogicParrot: I'll do it your way. Readability is preferred, especially if it turns out to be a micro optimization or if compilers optimize this away. If someone comes up with benchmarks in favor of a dedicated function, we'll later just use function overloading instead of a different brand new function. This keeps the Lua Api clean and saves an export of an additional function.

LogicParrot:
Todo update `APIDesc` for cWorld et al.

LogicParrot:
Potential todo: Replace the bools with named consts or with a bitmask to avoid parameter ordering confusion? I got confused myself already.

LogicParrot:
I am hesitant to change the API without consulting @madmaxoft . So, for now, I did not document the new optional parameters at the API.

tigerw:
The client uses this metadata for leaf decay indication in the debug screen. This is a hotfix only.

src/Blocks/BlockDirt.h
```diff
@@ -51,8 +51,8 @@ class cBlockDirtHandler :
 		{
 			BLOCKTYPE above = a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ);
 
-			// Grass turns back to dirt when the block above is not transparent
-			if (!cBlockInfo::IsTransparent(above))
+			// Grass turns back to dirt when the block above it is not transparent or water
+			if (!cBlockInfo::IsTransparent(above) || IsBlockWater(above))
```
tigerw: hmmm. Should water be considered transparent?

LogicParrot:
@tigerw I could easily send that to the client without marking dirty chunks by modifying a single line. But is it worth the bandwidth?

src/Blocks/BlockDirt.h
SphinxC0re: Isn't water already considered transparent? Wow!

src/BlockInfo.cpp
```diff
@@ -250,7 +250,6 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
 	a_Info[E_BLOCK_REDSTONE_WIRE       ].m_Transparent = true;
 	a_Info[E_BLOCK_SAPLING             ].m_Transparent = true;
 	a_Info[E_BLOCK_SIGN_POST           ].m_Transparent = true;
-	a_Info[E_BLOCK_SNOW                ].m_Transparent = true;
```
LogicParrot: Should snowing destroy grass?

Pokechu22:
@tigerw Actually, the notchian server also doesn't send this data.  For leaf decay, a special "Update but don't tell the client" mode is used.  From `BlockLeaves`:

```java
worldIn.setBlockState(blockpos, iblockstate.withProperty(CHECK_DECAY, Boolean.valueOf(true)), 4);
```

and 

```java
worldIn.setBlockState(pos, state.withProperty(CHECK_DECAY, Boolean.valueOf(false)), 4);
```

Here's the header for `World.setBlockState`:

```java
/**
 * Sets the block state at a given location. Flag 1 will cause a block update. Flag 2 will send the change to
 * clients (you almost always want this). Flag 4 prevents the block from being re-rendered, if this is a client
 * world. Flags can be added together.
 */
public boolean setBlockState(BlockPos pos, IBlockState newState, int flags)
```

Note how flags 1 and 2 are unset (in almost all cases they are set).

Pokechu22:
To clarify, the server doesn't send a block change packet when leaves decay, but if the player relogs then it does send the changed version.

LogicParrot:
@Pokechu22 this is the exact behavior after this PR.

tigerw:
How many of these packets are sent? It seems cleaner to tell clients what's happening, and less clean to add a bunch of fields to SetMeta (:P). I am quite surprised that its allegedly causing so much network utilisation.

src/BlockInfo.cpp
tigerw: It shouldn't.

src/Blocks/BlockDirt.h
tigerw: w...was that a good wow or a bad wow?

LogicParrot:
Multiple packets per second in a dense forest. Also, completely non dirty chunks are being marked dirty and saved to disk all the time. That is very serious. (They are also kept longer in ram when dirty, and that is very significant, see #3142).

Note that with the idea I had in mind, `SetMeta` flags are general purpose and should be used elsewhere whenever the meta update is not too important. E.g.:
- Pressing a button or opening a chest shouldn't mark the chunk dirty but should send the meta update to the client.
- Some non visually changing block change should save its meta to disk without telling the client.

tigerw:
Even multiple packets per second doesn't sound like much. In any case, why are so many leaves being generated decaying?

LogicParrot:
@tigerw Would you like to revert and discuss this?

tigerw:
Not at all, just asking about the original problem.

LogicParrot:
The original concern was the insanely high amount of dirty chunks on a Raspberry Pi. I found out that leaves and grass were behaving wildly and decided to fix them.

I thought that leaf decay meta updates, as well as many other meta updates (e.g. button pressing) are too unimportant for a disk write and/or a sent packet, so I made those optional at `setMeta`. I thought that perhaps many of code paths could utilize these flags in the future to reduce resource usage.

The immediate benefit is the reduced dirties and packets. I now realize this could have also been achieved by fixing the generator.

src/BlockInfo.cpp
LogicParrot: This has created a bug, then.

src/Blocks/BlockDirt.h
LogicParrot: I *think* that zombies will not burn when water is above.

LogicParrot:
The original concern was the insanely high amount of dirty chunks on a Raspberry Pi. I found out that leaves and grass were behaving wildly and decided to fix them.

I thought that leaf decay meta updates, as well as many other meta updates (e.g. button pressing, water/lava simulator bookeeping) are too unimportant for a disk write and/or a sent packet, so I made those optional at `setMeta`. I thought that perhaps many of the code paths could utilize these flags in the future to reduce resource usage.

The immediate benefit is the reduced dirties and packets. I now realize this could have also been achieved by fixing the generator.

LogicParrot:
>Even multiple packets per second doesn't sound like much.

I'll show you the log in a moment. You decide.

LogicParrot:
Here's a log of a STATIONARY player. (Only Moved a few meters upon startup, then stopped).
[decay.txt](https://github.com/cuberite/cuberite/files/222361/decay.txt)

LogicParrot:
We've got 610 updates at `22:05:46`. And this is probably not the peak.

LogicParrot:
@tigerw It seems that those flags have little use in any other code path. I think the generator is creating all leaves with the decay flag on, a fix should be easy. I'll fix the generator and revert this, except for the grass changes. 

Also, it seems someone took the leaf update into account in `FastSetBlock` already, but the leaves do not use that function.

tigerw:
SetBlock goes to FastSetBlock. Not sure about SetMeta.

I'm wondering if the generators are creating leaves with metadata zero? There are a lot of "leaf marked as checked" outputs which probably can be set during generation.

LogicParrot:
>SetBlock goes to FastSetBlock. Not sure about SetMeta.

Oh, sorry. I confused metas and blocks. I take the 2nd paragraph of my last comment back.

>I'm wondering if the generators are creating leaves with metadata zero?

Apparently so. all generaterd leaves can be safely marked as checked on generation.

tigerw:
progress :D

LogicParrot:
So, the plan for handling non important meta updates:
- For leaves: Fix generator to eradicate the updates. 
- For liquid simulators: Useless updates are probably already not sent to the client thanks to code in`setFastBlock`. Modify that code slightly so that it additionally does not call `markDirty` when nothing actually changes.
- For other blocks like buttons: The metas change too rarely to worry about.

It seems that no other use cases require my newly added meta flags, so that's all.

tigerw:
Plan is a good plan.

Issue #3136, opened 2016-04-10
segmentation fault on startup when -p or --port specified when ports= set in settings.ini


johnmccabe:
Hi,
I'm running Cuberite on Ubuntu 15.10 and get the crash at the end of the issue when I attempt to start the server with the port specified on the commandline when there is already a port specified in `settings.ini`.

To reproduce:
1. freshly install Cuberite `https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/lastSuccessfulBuild/artifact/Cuberite.tar.gz`
2. start using the `-p` arg to set a port, this creates the the `settings.ini` file with the port you've specified
     ```
    cd Server; ./Cuberite -p 25565
    ```
3. kill the server, `ctrl+c`
4. attempt to start again specifying the port on the command line, this should crash.
    ```
    ./Cuberite -p 25565
    ```

You can see the following, if you omit the `-p` arg then it starts and uses the port from `settings.ini`, if you delete `ports=` from `setttings.ini` then you can start when passing the `-p` arg (however this adds `ports=` back to `settings.ini` so subsequent starts fail.

Regards,
John

```
[13:12:48] --- Started Log ---
[13:12:48] Cuberite Jenkins clang x64 Release (master) build id: #454
[13:12:48] from commit id: 900d8e78c1691a8906d18b27414f5723cf17dad3 built at: Thu  7 Apr 17:44:20 CEST 2016
[13:12:48] Creating new server instance...
[13:12:48] Reading server config...
[13:12:48] Starting server...
[13:12:48] Compatible clients: 1.7.x, 1.8.x


; This is the main server configuration
[13:12:48] Compatible protocol versions 4, 5, 47
[13:12:48]   D:    | Cuberite has encountered an error and needs to close
[13:12:48] Details | SIGSEGV: Segmentation fault
[13:12:48] Cuberite Jenkins clang x64 Release (master) build id: #454
[13:12:48] from commit id: 900d8e78c1691a8906d18b27414f5723cf17dad3 built at: Thu  7 Apr 17:44:20 CEST 2016
./Cuberite(_Z15PrintStackTracev+0x19)[0x5eb4d9]
./Cuberite[0x5d54a5]
/lib/x86_64-linux-gnu/libc.so.6(+0x352f0)[0x7fdd9141a2f0]
./Cuberite(_ZN28cOverridesSettingsRepository9GetValuesESs+0x320)[0x57fa20]
[...]

PR #3137, opened 2016-04-10
Issue3136

bearbin:
This fixes the crash portion of #3136 

I couldn't get the ports set on the command line to actually override those set in the ini file though, maybe someone else could give me a hand?

### DIFF
--- a/src/OverridesSettingsRepository.cpp
+++ b/src/OverridesSettingsRepository.cpp
@@ -108,31 +108,26 @@ std::vector<std::pair<AString, AString>> cOverridesSettingsRepository::GetValues
 {
 	auto overrides = m_Overrides->GetValues(a_keyName);
 	auto main = m_Main->GetValues(a_keyName);
-	std::sort(overrides.begin(), overrides.end(), [](std::pair<AString, AString> a, std::pair<AString, AString> b) -> bool { return a < b ;});
-	std::sort(main.begin(), main.end(), [](std::pair<AString, AString> a, std::pair<AString, AString> b) -> bool { return a < b ;});
 
-	std::vector<std::pair<AString, AString>> ret;
+	auto ret = overrides;
 
-
-	size_t overridesIndex = 0;
-	for (auto pair : main)
+	for (const auto & mainpair : main)
 	{
-		if (overridesIndex >= overrides.size())
+		bool found = false;
+		for (const auto & returnpair : ret)
 		{
-			ret.push_back(pair);
-			continue;
+			if (returnpair.first == mainpair.first)
+			{
+				found = true;
+				break;
+			}
 		}
-		if (pair.first == overrides[overridesIndex].first)
+		if (found == false)
 		{
-			continue;
+			ret.push_back(mainpair);
 		}
-		while (pair.first > overrides[overridesIndex].first)
-		{
-			ret.push_back(overrides[overridesIndex]);
-			overridesIndex++;
-		}
-		ret.push_back(pair);
 	}
+
 	return ret;
 }
 
@@ -270,4 +265,3 @@ bool cOverridesSettingsRepository::Flush()
 {
 	return m_Overrides->Flush() && m_Main->Flush();
 }
-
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -371,7 +371,7 @@ static std::unique_ptr<cMemorySettingsRepository> ParseArguments(int argc, char
 		// Parse the comand line args:
 		TCLAP::CmdLine cmd("Cuberite");
 		TCLAP::ValueArg<int> slotsArg    ("s", "max-players",         "Maximum number of slots for the server to use, overrides setting in setting.ini", false, -1, "number", cmd);
-		TCLAP::MultiArg<int> portsArg    ("p", "port",                "The port number the server should listen to", false, "port", cmd);
+		TCLAP::MultiArg<int> portsArg    ("p", "ports",                "The port numbers the server should listen to", false, "port", cmd);
```
johnmccabe: This should probably stay as `--port` as its a `MultiArg` - ie `./Cuberite --port 25565 --port 25566`, would need something added to the description to indicate it can be set multiple times.

bearbin:
@johnmccabe I fixed your comments.

tigerw:
> I couldn't get the ports set on the command line to actually override those set in the ini file though, maybe someone else could give me a hand?

It certainly looks like it should work: the default is `overridesrepository`; but if something is present in `main` but not in `overrides`, add it. If this analysis is correct then something's probably subtly wrong.

src/OverridesSettingsRepository.cpp
```diff
@@ -108,31 +108,25 @@ std::vector<std::pair<AString, AString>> cOverridesSettingsRepository::GetValues
...
-			continue;
-		}
-		if (pair.first == overrides[overridesIndex].first)
+		bool found = false;
+		for (auto returnpair : ret)
 		{
-			continue;
+			if (returnpair.first == mainpair.first)
+			{
+				found = true;
```
tigerw: It looks like you can do the breakficiation here?

src/OverridesSettingsRepository.cpp
```diff
@@ -108,31 +108,25 @@ std::vector<std::pair<AString, AString>> cOverridesSettingsRepository::GetValues
...
-	std::sort(overrides.begin(), overrides.end(), [](std::pair<AString, AString> a, std::pair<AString, AString> b) -> bool { return a < b ;});
-	std::sort(main.begin(), main.end(), [](std::pair<AString, AString> a, std::pair<AString, AString> b) -> bool { return a < b ;});
 
-	std::vector<std::pair<AString, AString>> ret;
+	auto ret = overrides;
 
-
-	size_t overridesIndex = 0;
-	for (auto pair : main)
+	for (auto mainpair : main)
```
tigerw: If you want, perhaps a const auto &, just to not do copying :)

src/OverridesSettingsRepository.cpp
```diff
@@ -108,31 +108,25 @@ std::vector<std::pair<AString, AString>> cOverridesSettingsRepository::GetValues
...
 	{
-		if (overridesIndex >= overrides.size())
-		{
-			ret.push_back(pair);
-			continue;
-		}
-		if (pair.first == overrides[overridesIndex].first)
+		bool found = false;
+		for (auto returnpair : ret)
```
tigerw: Likewise

tigerw:
Any crash fix is a good fix. Merge while the sun still shines! 🌞

bearbin:
> It certainly looks like it should work: the default is overridesrepository; but if something is present in main but not in overrides, add it. If this analysis is correct then something's probably subtly wrong.

I recommend you look at the old code and try and draw some sense from it. I have absolutely no idea what it was trying to do.

bearbin:
In any case, I responded to your comments.

bearbin:
@LogicParrot Could you review the changes so as to see what went wrong here? In particular the old code, which I couldn't understand at all (I just replaced it as it seemed to make no sense).

LogicParrot:
I did not review the code yet. This is what I know so far:
- You modified `OverridesSettingsRepository.cpp`, and that made `cRoot::LoadWorlds` malfunction.
- I think that your changes made identical INI keys collide.
e.g. If we have such a config:
```
World=Hello1
World=Hello2
```
Only one of the worlds would be read (I do not recall which) and the other is completely ignored.

bearbin:
@LogicParrot If it was simply doubled INI values being ignored, then the fix is rather simple and I can push it back in a few mins if you want.

LogicParrot:
Sure. Please do not push it straight to master though. Let us review it, and let us try to understand what the old code does and see if the new code replicates the full functionality.

Issue #3130, opened 2016-04-06
Sometimes, the player is frozen at login


LogicParrot:
The changes in #3097 freeze players upon log in, and unfreeze them when the chunk loads. It seems that on rare occasions, the unfreeze fails, and the player will remain frozen even after the chunk has been loaded. Logging out and re-logging in resolves it.

It seems that this failure only happens on login, and not on other freeze triggers, such as venturing to an unloaded chunk. However, I am not completely sure about this.

- Why does it happen?
- Can anyone find a reliable reproduction method?

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=32674940)](https://www.bountysource.com/issues/32674940-sometimes-the-player-is-frozen-at-login?utm_source=32674940&utm_medium=shield&utm_campaign=ISSUE_BADGE)

popsUlfr:
Hi, I've been encountering this issue consistently on my server. The first time I connect to a freshly started server, there are no issues. The next time I connect, the player is frozen in place and I can not move at all.

Sometimes upon fast disconnect and reconnect I can walk a few steps at which point I'm "catapulted" back to my starting position and frozen in place.

The client throws this exception when it happens (Minecraft Client 1.8.9) :
```
[13:45:52] [Client thread/INFO]: [CHAT] §6Welcome to the TuxHomeNet server!
[13:45:52] [Client thread/INFO]: [CHAT] §6Type /help for all commands
[13:45:53] [Client thread/FATAL]: Error executing task
java.util.concurrent.ExecutionException: java.lang.IndexOutOfBoundsException: readerIndex(2) + length(14988) exceeds writerIndex(8): UnpooledHeapByteBuf(ridx: 2, widx: 8, cap: 8)
        at java.util.concurrent.FutureTask.report(FutureTask.java:122) ~[?:1.8.0_77]
        at java.util.concurrent.FutureTask.get(FutureTask.java:192) ~[?:1.8.0_77]
        at net.minecraft.util.Util.func_181617_a(SourceFile:45) [g.class:?]
        at net.minecraft.client.Minecraft.func_71411_J(Minecraft.java:1014) [ave.class:?]
        at net.minecraft.client.Minecraft.func_99999_d(Minecraft.java:349) [ave.class:?]
        at net.minecraft.client.main.Main.main(SourceFile:124) [Main.class:?]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_77]
[...]

PR #3139, opened 2016-04-13
Fix occasional freeze bug

LogicParrot:
This might close the player freeze issue (#3130). I could not reproduce the problem locally, so I did not test this. Can someone test this and report back? @popsUlfr @Pokechu22

Binaries are available below. You can also fetch the `freezeFix` branch and compile this.
- [Linux x64](https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20ingame-testing-needed/11/)
- [Raspberry Pi](https://builds.cuberite.org/job/Cuberite%20Linux%20raspi-armhf%20ingame-testing-needed/10/)
- [Windows x64](https://builds.cuberite.org/job/Cuberite%20Windows%20x64%20ingame-testing-needed/10/)

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -65,6 +65,7 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 	m_GameMode(eGameMode_NotSet),
 	m_IP(""),
 	m_ClientHandle(a_Client),
+	m_IsFrozen(false),
 	m_NormalMaxSpeed(1.0),
 	m_SprintingMaxSpeed(1.3),
 	m_FlyingMaxSpeed(1.0),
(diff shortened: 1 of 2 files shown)


LogicParrot:
I managed to test this. It works.

src/Entities/Player.cpp
```diff
@@ -65,6 +65,7 @@ cPlayer::cPlayer(cClientHandlePtr a_Client, const AString & a_PlayerName) :
 	m_GameMode(eGameMode_NotSet),
 	m_IP(""),
 	m_ClientHandle(a_Client),
+	m_IsFrozen(false),
```
LogicParrot: This is the only change. The rest is minor code tidying.

popsUlfr:
Thanks @LogicParrot, I can't reproduce the issue on my end anymore ;)

src/Entities/Player.cpp
tigerw: m_IsManuallyFrozen?

LogicParrot: The value of `m_IsManuallyFrozen` is irrelevant when `m_IsFrozen` is `false`.

LogicParrot:
Glad it works :)

For the record: `m_IsManuallyFrozen` and `m_IsFrozen` were both unitialized prior to this. So they are randomly true or false initially. The bug triggers when both `m_IsFrozen` and `m_IsManuallyFrozen` are initialized to `true`. Assuming equal chance for true and false, this bug would have manifested itself in 25% of all logins.

src/Entities/Player.cpp
LogicParrot: Hmm. Perhaps it would be wiser to have an enum. `m_FrozenState { None, Manual, Auto }`

tigerw: An excellent idea.

PR #3142, opened 2016-04-14
Save chunks more often to reduce RAM usage

LogicParrot:
Please discuss!

In Cuberite, a dirty chunk is a chunk which has changes that are yet unsaved to storage.
Every 300 seconds, there is a save cycle: chunks are scanned, dirty chunks are saved to disk and become non dirty.
Every 10 seconds, there is an unload cycle: chunks are scanned, unused, non-dirty chunks are freed from ram.

There's a problem here: An **unused dirty** chunk that wants to unload has to wait in RAM for up to 300 seconds until the save cycle, only then it may unload.

I made a simple change:
- Used chunks are saved every 300 seconds in the save cycle, just like before.
- Unused chunks are saved immediately in the unload cycle and then get unloaded, rather than waiting for the save cycle.

The result was astounding. I had a player fly in a straight line in standard for a while and test both cases.

**Case 1: The master branch. (Debug mode)**
Upon login, there were ~300 loaded chunks. As I flied in a straight line, loaded chunks accumilated until reaching ~3000, then 5 minutes passed and the save cycle triggered, my pc lagged for 20 seconds, and then there was a rapid drop in loaded chunks until settling back at ~300.
Ram usage peak was ~320 MB.

Here is a gif of the accumilation, and an image of the sudden drop.
bad
thedrop

**Case 2: My PR. (Debug mode)**
Upon login, there were ~300 loaded chunks. Flying never changed that significantly, the amount of loaded chunks never surpassed 630. Ram usage peaked at about 100MB.

Here is a gif of the lack of accumilation.
good

- Note that in both test cases, RAM usage would have been far less in release mode.
- The difference would have been bigger for faster flying speeds.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -3066,6 +3066,7 @@ void cChunkMap::cChunkLayer::Save(void)
 
 void cChunkMap::cChunkLayer::UnloadUnusedChunks(void)
 {
+	cWorld * World = m_Parent->GetWorld();
 	for (size_t i = 0; i < ARRAYCOUNT(m_Chunks); i++)
 	{
 		if (
@@ -3074,11 +3075,20 @@ void cChunkMap::cChunkLayer::UnloadUnusedChunks(void)
 			!cPluginManager::Get()->CallHookChunkUnloading(*(m_Parent->GetWorld()), m_Chunks[i]->GetPosX(), m_Chunks[i]->GetPosZ())  // Plugins agree
 		)
 		{
-			// The cChunk destructor calls our GetChunk() while removing its entities
-			// so we still need to be able to return the chunk. Therefore we first delete, then nullptrify
-			// Doing otherwise results in bug https://forum.cuberite.org/thread-355.html
-			delete m_Chunks[i];
-			m_Chunks[i] = nullptr;
+			if (m_Chunks[i]->IsDirty() && m_Chunks[i]->IsValid())
+			{
+				// The chunk is ready to unload, except that it's dirty. Let's save it first.
+				// It will be unloaded in the next unload cycle.
+				World->GetStorage().QueueSaveChunk(m_Chunks[i]->GetPosX(), m_Chunks[i]->GetPosZ());
+			}
+			else
+			{
+				// The cChunk destructor calls our GetChunk() while removing its entities
+				// so we still need to be able to return the chunk. Therefore we first delete, then nullptrify
+				// Doing otherwise results in bug https://forum.cuberite.org/thread-355.html
+				delete m_Chunks[i];
+				m_Chunks[i] = nullptr;
+			}
 		}
 	}  // for i - m_Chunks[]
 }
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -215,7 +215,6 @@ bool cChunk::CanUnload(void)
 {
 	return
 		m_LoadedByClient.empty() &&  // The chunk is not used by any client
-		!m_IsDirty &&                // The chunk has been saved properly or hasn't been touched since the load / gen
 		(m_StayCount == 0) &&        // The chunk is not in a ChunkStay
 		(m_Presence != cpQueued) ;   // The chunk is not queued for loading / generating (otherwise multi-load / multi-gen could occur)
 }


Server/Plugins/Core
```diff
@@ -1 +1 @@
-Subproject commit 5df413231a918070adc15dfe36630ffba06e15dd
+Subproject commit 0ab749631fcd744e6713af2499046a9feed04daf
```
LogicParrot: Oops.

LogicParrot:
This may require an `ifQueuedForSave` flag to avoid double saving, since saving now happens more often.

LogicParrot:
**More considerations**
- There is more room for improvement. In some cases, keeping the chunk in RAM is a good thing. Consider a scenario where a player is constantly visiting and then leaving a chunk. The performance would be better if we keep that chunk in RAM. This is the downside of this PR. In the future, some heuristics could be of use. e.g. a Branch-target-predictor-style mechanism which would detect chunks that are being unloaded / loaded too many times and keep them in RAM longer.

- This PR is about the classical balance problem between RAM and disk writes. Currently things are overly biased toward the hard drive. Also, the sudden save burst is unacceptable, in my opinion.

src/ChunkMap.cpp
```diff
@@ -3074,11 +3075,20 @@ void cChunkMap::cChunkLayer::UnloadUnusedChunks(void)
...
-			// The cChunk destructor calls our GetChunk() while removing its entities
-			// so we still need to be able to return the chunk. Therefore we first delete, then nullptrify
-			// Doing otherwise results in bug https://forum.cuberite.org/thread-355.html
-			delete m_Chunks[i];
-			m_Chunks[i] = nullptr;
+			if (m_Chunks[i]->IsDirty() && m_Chunks[i]->IsValid())
+			{
+				// The chunk is ready to unload, except that it's dirty. Let's save it first.
+				// It will be unloaded in the next unload cycle.
+				World->GetStorage().QueueSaveChunk(m_Chunks[i]->GetPosX(), m_Chunks[i]->GetPosZ());
```
LogicParrot: Should I check if it's already queued?

SphinxC0re: Probably

tonibm19:
Wow, really nice, good job.

worktycho:
I think that we should avoid unloading chunks unless we definitely need the space. Free RAM is wasted RAM. On option would be to only trigger an unload if used RAM reached a certain value. Perhaps if we aimed to keep chunk memory at around 80% active, 18% easily unloadable, 2% free. (Actual values subject to tuning and config options), or alternatively a fixed numerical amount in use (active/unloadable).

It should be fairly easy to keep such a balance by adding code to the cAllocationPool to track used memory, and trigger unloads when necessary. Even random unloading would eventually unload less used chunks, whilst more used one would be reloaded. Alternately, we could store ticks since last loaded, or store the Age at last loaded, to prioritize least recently used chunks.

Schwertspize:
I would recommend a config value defining the ~amount of good ram usage. If the user says, one gigabyte is completely fine (not used ram is wasted ram) we can store chunks much longer and/or pre-load chunks for better performance. If, on the other hand, the user says less ram, the chunk cleaner could be more aggressive. 

I think setting a default value of 150-200mb would be good (the server woild have to monitor the own memory usage and execute an unload chunk task if it's peaking). Another advantage is we don't need to code advanced heuristics.

Pokechu22:
Wouldn't those chunks also cause other performance issues due to needing to be ticked, beyond the RAM issues?

Pokechu22:
Ignore the previous comment, I see that chunks check whether they are loaded before a player before being ticked.

Schwertspize:
@Pokechu22 i saw that you stroke that. Anyway, I think this is about chunks which don't get ticked anyway.

LogicParrot:
> I think that we should avoid unloading chunks unless we definitely need the space.

This issue easily chokes Raspberry Pis with not too wild flying.


> Free RAM is wasted RAM. On option would be to...

Shouldn't the OS take care of most of that for us?If we free a chunk and a player re-visits it too soon, it is probably already cached in RAM thanks to the OS's disk IO cache. By the way, we already unload non-dirty chunks asap. The dirties are the problem here.

#3110 aims to minimize dirty chunks, if I do it well enough, this PR might become unneeded.

LogicParrot:
>I would recommend a config value defining the ~amount of good ram usage...

Configurability sounds like a good idea.

>It should be fairly easy to keep such a balance by adding code to the cAllocationPool to track used memory, and trigger unloads when necessary....

I'd like to add that if we decide to create such a thing we should priortize unloading non-dirty chunks.

worktycho:
My point is that we should keep non-dirty, inactive chunks around until we run out of RAM. The disk cache is still slower than keeping it in RAM, so unless the os is likely to page the chunk out, we want to keep the non-dirty chunks in RAM. So I'm proposing we should eliminate regular unload cycles, so a pi would run unloads frequently, due to low RAM, whilst a desktop/server would run unloads much less frequently. I'm also proposing that rather than unloading all inactive chunks, we should only unload some of them.

Certainly unloading non-dirty chunks first would be a good idea, as they could be freed immediately.

LogicParrot:
I see.

>Certainly unloading non-dirty chunks first would be a good idea, as they could be freed immediately.

It's also a good idea because if you write a dirty chunk to disk and then it gets dirty again (pretty likely), you had to perform 2 disk writes rather than 1.

src/ChunkMap.cpp
worktycho: We should handle that in cWorldStorage. At the moment its not possible to eliminate the race condition.

worktycho:
An inactive chunk is unlikely to get dirty, and if its been written too again, it probably shouldn't be unloaded. Possibly when a chunk becomes inactive (last player leaves and last chunkstay removed) we should add it to the save queue without waiting for a save cycle.

LogicParrot:
@worktycho a dirty chunk is likely to be become dirty again. For instance, a miner going back and forth between a mine and his house will constantly re-trigger dirtiness while mining.

Also, I think I now see the rationale between the save cycle; if the game ever crashes, we want the chunks to be in a consistent state, meaning we want them to have been saved at roughly the same time. So it is probably wrong to save chunks routinely, and it should only done via a save cycle where all saves happen roughly at the same time.

LogicParrot:
>Possibly when a chunk becomes inactive (last player leaves and last chunkstay removed) we should add it to the save queue without waiting for a save cycle.

Essentially, that's what this PR does.

LogicParrot:
@worktycho a dirty inactive chunk is likely to be become dirty again. For instance, a miner going back and forth between a mine and his house will constantly re-trigger dirtiness while mining.

Also, I think I now see the rationale of the save cycle; if the game ever crashes, we want the chunks to be in a consistent state, meaning we want them to have been saved at roughly the same time. So it is probably wrong to save chunks routinely, and it should only done via a save cycle where all saves happen roughly at the same time.

LogicParrot:
Even with the current save cycle, things are not atomic though; if it happens to crash in the midst of the cycle, the next time the game runs some chunks would be 5 minutes old and some wouldn't.

LogicParrot:
Proposed simple algorithm:
```
function cWorld::tick:
    //max_chunk_ram is configurable in settings.ini
    while LoadedChunks * sizeOf(oneChunk) > max_chunk_ram &&
              we have unused non dirty chunks 
    {
        unload one non dirty chunk
    }
    // If we make it into this loop, then unloading non dirties was not enough
   while LoadedChunks * sizeOf(oneChunk) > max_chunk_ram &&
    we have unused dirty chunks
    {
        save and unload one dirty chunk
    }

function cChunk::tick:
    if dirty and last saved > 5 minutes, then save. // chunks shouldn't remain dirty for too long, we want the chunk to be somewhat recent if there's a crash.
```

LogicParrot:
Multiworld variant:
```
function cWorld::tick:
     // Minimize locks by doing this every 2 seconds
     if (time % 40 == 0 && nothing is queued for unload)
    {
        //obtain lock from cRoot or cServer. 
        if (cRoot:GetCurrentUnloadingWorld() == this)
        {
            ChunksLoaded = cRoot:GetLoadedChunksOfAllWorlds()
            MaxChunks = cRoot:GetMaxChunks()
            if ChunksLoaded > MaxChunks
            {
                DesiredUnloadCount = MaxChunks - ChunksLoaded
                ActualUnloadCount = QueueUnloadChunks(DesiredUnloadCount)
                if (ActualUnloadCount < DesiredUnloadCount)
                {
                    if (nothing Is Queued For Save)
                    {
                      QueueSaveDirtyChunks(DesiredUnloadCount - ActualUnloadCount);
                    }
               }
            }
           cRoot:SetNextUnloadingWorld();
       }
    }
```

LogicParrot:
I removed my second algorithm because I need to think about it a bit.

Schwertspize:
@LogicParrot you told me you like the configurability. Would you comment on my suggestion to handle it based on the actual ram usage?

LogicParrot:
@Schwertspize Specifying the maximum RAM to allocate for chunks is a great idea. I've done something similar in #3151, but currently one has to specify the maxmum amount of chunks instead.

src/ChunkMap.cpp
LogicParrot: #3151 deals with this.

@worktycho What race condition?

worktycho:
The problem with that algorithm is that sizeof(Chunk) is not a fixed value and defiantly not equal to `sizeof(cChunk)`. In a normal world memory used by a chunk can vary by ~10KB. In something like a floating islands world, the chunk memory usage can vary by 100KB.

For calculating RAM usage, you need something along the lines of `LoadedChunks * (sizeof(cChunk)  + (Internal Allocations Factor))+ Allocated Segments * sizeof(cChunkData)`
You may also want to include the fact that we keep 1600 chunk segments around per world to protect against malloc failure. (I have no idea why I wrote the allocator the way I did, we really should look at that, and possibly use the MaxRamChunksSetting as a guide to how many spare segments to have around).

I would suggest an algorithm along the lines of:
```
function cWorld::tick:
    // GetMaxRamChunks()'s value is configurable in settings.ini
     if (nothing is queued for unload)
    {
        LoadedUsage = GetTotalChunkMemoryUsage() //Returns approximate amount of memory used for chunk storage based on above paragraph
        AverageUsage = LoadedUsage/GetLoadedChunks()
        MaxUsage = GetMaxRamChunks()
        if LoadedUsage > MaxUsage
        {
            DesiredUnloadCount = (MaxUsage - LoadedUsage)/AverageUsage 
            ActualUnloadCount = QueueUnloadChunks(DesiredUnloadCount)
            if (ActualUnloadCount < DesiredUnloadCount)
            {
                if (nothing Is Queued For Save)
                {
                  QueueSaveDirtyChunks(DesiredUnloadCount - ActualUnloadCount);
                }
            }
        }
    }
```
It approximates the number to unload based on total memory usage, (an approximate value unless we send all the STL stuff + dynamic allocations through a custom allocator that tracks total usage).

worktycho:
Also regarding dirty inactive chunks, what I meant was an inactive chunk is unlikely to become dirty without becoming an active chunk again. So in the house + mine example, when the miner would require the chunk to become active again. However, I think your point is that dirty chunks are more likely to become active, which seems extremely likely.

And what I meant by adding to the save queue immediately was add hooks into player and chunk-stay removal, so that we can add the chunk to the save queue as soon as it becomes inactive. Combined with the caching algorithm proposed, and a couple of hooks into chunk/chunksegment allocation, and we could eliminate the unload cycle completely.

LogicParrot:
Closing in favor of #3151

PR #3143, opened 2016-04-14
Proper fix for #3136

bearbin:
@LogicParrot the code is available for review.


SphinxC0re:
Could you please add some more documentation to the code, since the changes aren't that trivial? If we carry on with that behavior, we will eventually end up with a better documented source than before

PR #3144, opened 2016-04-14
Create ISSUE_TEMPLATE

SphinxC0re:
This issue template is a boilerplate that we should probably adjust in the future to fit our needs. I think issues might be much easier to read in the future :stuck_out_tongue_winking_eye:


bearbin:
Looks good to me.

Issue #2766, opened 2015-12-17
Chunk generator very slow in biome Roofed Forest


Seadragon91:
I have the problem if I enter a biome with Roofed Forest, the fps drops from 60 below 10. Also the chunk generator is slow, there.

Example seed: 19477766
Coord: -50 85 -70

![2015-12-17_20 22 16](https://cloud.githubusercontent.com/assets/14818618/11879623/c9f968aa-a4fc-11e5-8232-3a097cc4fa0b.png)

Please test this seed.

madmaxoft:
I expect this to be caused by way too many leaf-block updates.

Seadragon91:
Used worldedit to test that with the leaves, yes that could be it.

worktycho:
The algorithm used for handling leaf decay checks is very inefficient.

If we could improve it to avoid rechecking blocks it would probably improve efficiency.

LogicParrot:
The algorithm is not the bottleneck, it's the amount of times the algorithm is executed that's the problem. #3149 resolves this, but after visiting an affected area, it can take a couple of minutes for existing worlds' leaf metas to be properly sit, and during that time there would still be some performance issues.

Issue #3021, opened 2016-02-18
cuberite uses lots of bandwidth much more than java server


xiaooloong:
only one player in the game, it could take all my server's out going bandwidth 1mbps

SphinxC0re:
Wow! I never though this would be a problem. In 6 Years o development, this has never been a problem to *anyone*.

@cuberite/core-team @cuberite/server-contributors

xiaooloong:
我们这种破地方网速慢真是对不起了
之前用cauldron开服，装了一大堆plugin和mod，1M的带宽几个哥们玩儿一点儿事儿没有
今天试了一下cuberite，真是不给cent6活路，把服务器换成ubuntu14.04还得换源升级gcc，你说游戏服务器用新版本开发赶时髦就算了，还不体谅一下我们这些个老古董。
当然这不是正事儿，问题在于，服务器刚开好，就进去俩人，服务器1M的带宽就给你飙满了，走个路都要耗半天，说句话猴年马月才出的来。去后台iftop一看，带宽爆炸，吓得我赶紧退了游戏，带宽马上减半，然而并没有什么卵用，cuberite马上又把带宽用满，继续卡成狗。
我就不继续撕了，今天有点小情绪，出来爆粗口挺丢人的，真是对不住了

bearbin:
I feel like some process where we're re-sending data unnecessarily compared to vanilla could have crept in. Maybe we need to revive protoproxy?

SphinxC0re:
@xiaooloong huh?

bearbin:
@SphinxC0re Google Translate gives a reasonable translation.

madmaxoft:
@bearbin not really

tigerw:
He says:

-----

(Preamble) We live in a place with really slow internet, really sorry about this.

Before using cauldron (cuberite?), we installed a ton of plugins and mods, and a 1M connection was fine for a couple of mates playing.

We tried cuberite today, and changed centos6 to Ubuntu 14.04 and upgraded gcc, etc. - that was hard!

That doesn't matter, but the problem was that once the server started, two people came on, and the 1M connection was immediately saturated. Walking and chatting took ages.

Looking at `iftop`, utilisation was going through the roof, so I immediately left the game. Utilisation cut in half. Joining again, it went back up to full, but nobody was doing anything intensive on the server.

Sorry about the rant, very embarrassing.

xiaooloong:
@tigerw thank you very much
[...]

PR #3149, opened 2016-04-18
Reduced unnecessary block updates and some code cleanup

### LINKED ISSUE #3145: The generator creates grass in undesired areas
### Expected behavior
The generator is expected to create grass on a block only when air is above.

### Actual behavior
The generator creates grass under tree trunks and snow. that grass decays into dirt, sending a non-negligible amount needless packets for several minutes upon entering a newly generated area.

#### Benchmark
I created a fresh world and logged in. about 32 block updates were being sent per second! This number gradually decreased until reaching zero. The log is attached here:
[out.txt](https://github.com/cuberite/cuberite/files/222337/out.txt)

Related: #3129 , #3142 , #3110 , #3021

LogicParrot:
- Chunk generator no longer generates grass under tree trunks. This reduces dirty chunks. Closes #3145
- Water simulator no longer marks chunks dirty for no good reason.
- Leaves are now created with the proper metas, no hundreds of updates are sent to the client (#3129 had a hotfix for this, but here it is done properly)
- Closes a portion of https://github.com/cuberite/cuberite/issues/3110


Seadragon91:
Would this also fix #2766?

LogicParrot:
Very possibly. And it may fix this too: #3021 well, at least for stationary players. (I don't know well/bad sending full chunks is).

src/Chunk.cpp
```diff
@@ -1536,8 +1535,17 @@ void cChunk::FastSetBlock(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_BlockT
 		return;
 	}
 
-	MarkDirty();
-	m_IsRedstoneDirty = true;
+	bool ReplacingLiquids = ( \
```
tigerw: > \

o.0

src/Entities/Entity.cpp
```diff
@@ -228,6 +228,7 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
 			this->GetUniqueID(), this->GetClass(),
 			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
 		);
+		LOGD("%g %g %g", GetPosX(), GetPosY(), GetPosZ());
```
tigerw: And the comment to self in...

src/Chunk.cpp
109C: >(\
/)

src/Entities/Entity.cpp
LogicParrot: Lol !!

src/Generating/Trees.cpp
```diff
@@ -341,24 +341,24 @@ void GetSmallAppleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a
 	// 2 BigO2 + corners layers:
 	for (int i = 0; i < 2; i++)
 	{
-		PushCoordBlocks (a_BlockX, Hei, a_BlockZ, a_OtherBlocks, BigO2, ARRAYCOUNT(BigO2), E_BLOCK_LEAVES, E_META_LEAVES_APPLE);
-		PushCornerBlocks(a_BlockX, Hei, a_BlockZ, a_Seq, a_Noise, 0x5000000 - i * 0x10000000, a_OtherBlocks, 2, E_BLOCK_LEAVES, E_META_LEAVES_APPLE);
+		PushCoordBlocks (a_BlockX, Hei, a_BlockZ, a_OtherBlocks, BigO2, ARRAYCOUNT(BigO2), E_BLOCK_LEAVES, E_META_LEAVES_APPLE_CHECK_DECAY);
```
tigerw: Hmmm. For this and for all below, the MC wiki states that "if the check decay bit is set, the leaf is checked for decay. The bit is set when adjacent blocks are changed." Am I misunderstanding, or is our leaf decay code broken, or was the generator fine in the first place?

src/Entities/Entity.cpp
tigerw: ...now.

src/Generating/Trees.cpp
LogicParrot: We might be having some flipped roles... Let me check this.

LogicParrot: We had it flipped. I fixed it. I propose that we should not inform the clients of these meta updates, at least until all forests "normalize" for most people, and possibly forever.

src/Generating/StructGen.cpp
```diff
@@ -183,6 +183,14 @@ void cStructGenTrees::ApplyTreeImage(
 				CASE_TREE_OVERWRITTEN_BLOCKS:
 				{
 					a_ChunkDesc.SetBlockTypeMeta(itr->m_RelX, itr->m_RelY, itr->m_RelZ, itr->m_BlockType, itr->m_BlockMeta);
+					// If grass is below our tree, turn it to dirt
```
tigerw: Tree! Tree? My dear LogicParrot, this is a *StructGen*.

Still, I dislike the whole commit-fetch-rebase-push, especially for little things like comments, so I shan't mind if you don't bother.

src/Generating/StructGen.cpp
```diff
@@ -183,6 +183,14 @@ void cStructGenTrees::ApplyTreeImage(
...
+					if (
+						(cBlockInfo::IsSolid(itr->m_BlockType)) &&
```
tigerw: Could it be a good idea to pull the grass-to-dirt check out of BlockDirt and use that? I remember something about there being water checks too?

src/Generating/StructGen.cpp
```diff
@@ -183,6 +183,14 @@ void cStructGenTrees::ApplyTreeImage(
 				CASE_TREE_OVERWRITTEN_BLOCKS:
```
tigerw: I'd just like to take a moment to appreciate the peculiarity of this statement. It is truly a *case* of unnecessary macro use.

src/Generating/Trees.cpp
tigerw: I remain sceptical about meta updates. The client seems to send a position update once every tick from previous observation. If I am not wrong in this regard, then the connection can handle it fine. I would be interested to see a breakdown of all sent packet types, somehow. But, what do I know :P

src/Generating/StructGen.cpp
LogicParrot: What's wrong? This is a tree.

src/Generating/StructGen.cpp
LogicParrot: Water can decay grass. But since this is a tree being generated on grass, there is no water here.

LogicParrot:
Can we merge this?
@tigerw I don't quite understand what you meant by the tree comment. Would you mind explaining? Is there anything wrong?

tigerw:
Of couse you can merge it. Never mind about the comment about tree comments, it was just a thought.

PR #2565, opened 2015-10-26
Simplified cChunkMap chunk storage

tigerw:
Funnily enough, on my system, on this particular world, std::map is at least as fast, sometimes faster than the array splitting approach we use:

Current:
```
     [19:55:03] Preparing spawn (world): 3.50% (14/400; 13.78 chunks / sec)
     [19:55:04] Preparing spawn (world): 13.00% (52/400; 37.25 chunks / sec)
     [19:55:04] Chunk generator performance: 53.68 ch / sec (108 ch total)
     [19:55:05] Preparing spawn (world): 26.75% (107/400; 53.45 chunks / sec)
     [19:55:06] Preparing spawn (world): 40.75% (163/400; 54.90 chunks / sec)
     [19:55:06] Chunk generator performance: 55.86 ch / sec (226 ch total)
     [19:55:07] Preparing spawn (world): 51.50% (206/400; 42.83 chunks / sec)
     [19:55:08] Preparing spawn (world): 64.50% (258/400; 51.74 chunks / sec)
     [19:55:08] Chunk generator performance: 54.90 ch / sec (333 ch total)
     [19:55:09] Preparing spawn (world): 77.75% (311/400; 52.68 chunks / sec)
     [19:55:10] Preparing spawn (world): 92.00% (368/400; 56.44 chunks / sec)
     [19:55:10] Chunk generator performance: 56.02 ch / sec (453 ch total)
     [19:55:11] Startup complete, took 10432ms!
Info [19:57:12] User cheesedummy10 authenticated with UUID 91c40b1eb8954561a6aa735b05632cc9
     [19:57:12] Player data file not found for cheesedummy10 (91c40b1eb8954561a6aa735b05632cc9, offline 18446de9338f3b28859a231072cb0568 (unused)), will be reset to defaults.
Info [19:57:12] Player cheesedummy10 has joined the game
     [19:57:15] Chunk generator performance: 40.03 ch / sec (48 ch total)
     [19:57:20] Chunk generator performance: 38.33 ch / sec (77 ch total)
     [19:57:22] Chunk generator performance: 41.86 ch / sec (168 ch total)
     [19:57:37] Chunk generator performance: 43.91 ch / sec (89 ch total)
     [19:57:40] Chunk generator performance: 46.36 ch / sec (93 ch total)
```

New:
```
     [19:40:25] Preparing spawn (world): 4.75% (19/400; 18.52 chunks / sec)
[...]

PR #3153, opened 2016-04-20
ChunkLayers now stored in std::map

LogicParrot:
This is just an exploratory alternative to #2565. I have no idea if it's better, it might be not.
- #2565 has a map of chunks in cChunkMap
- This has a map of chunkLayers in cChunkMap. It means there are a thousand (64*64) times less items at best in the map at any given time.

If #2565 uses an unordered_map instead, it *might* outperform both. But we may need to tweak it to minimize rehashing and collisions.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -51,24 +51,10 @@ cChunkMap::cChunkMap(cWorld * a_World) :
 
 
 
-cChunkMap::~cChunkMap()
-{
-	cCSLock Lock(m_CSLayers);
-	while (!m_Layers.empty())
-	{
-		delete m_Layers.back();
-		m_Layers.pop_back();  // Must pop, because further chunk deletions query the chunkmap for entities and that would touch deleted data
-	}
-}
-
-
-
-
-
 void cChunkMap::RemoveLayer(cChunkLayer * a_Layer)
 {
 	cCSLock Lock(m_CSLayers);
-	m_Layers.remove(a_Layer);
+	m_Layers.erase(std::pair<int, int>(a_Layer->GetX(), a_Layer->GetZ()));
 }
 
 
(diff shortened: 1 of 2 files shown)


src/ChunkMap.h
```diff
@@ -8,7 +8,7 @@
 
 #include "ChunkDataCallback.h"
 #include "EffectID.h"
-
+#include <unordered_map>
```
worktycho: Why add the unordered_map header if your going to use a map?

LogicParrot: Leftover from some other experiment.

LogicParrot:
CI is complaining about copying a layer. Where am I copying a layer?

LogicParrot:
CI issues fixed.
Intuitively, this feels faster than #2565 (assuming an std::map implementation); Since chunk loads are not completely random and chunks tend to be loaded at the same area, there are far fewer elements in the `std::map`.  I wonder what the benchmarks say.

src/ChunkMap.cpp
```diff
@@ -51,24 +51,10 @@ cChunkMap::cChunkMap(cWorld * a_World) :
 
 
 
-cChunkMap::~cChunkMap()
-{
-	cCSLock Lock(m_CSLayers);
```
LogicParrot: Perhaps I should keep this lock?

worktycho: The problem is that if the Lock is protecting us against anything, then we are already in UB land.

LogicParrot:
This looks merge-ready.

tigerw:
Performance testing?

LogicParrot:
Yeah, merge ready except for the performance testing :P
Though technically, this can already be merged:
 - Definitely faster than master for any non trivial amount of loaded layers.
 - Changes are minimal and the chances of a bug are nill.
 - #2565 can still be merged once CI is happy and performance tests are made. Conflicts are minimal, and this PR's changes can simply be overwritten.

tigerw:
okay then.

LogicParrot:
It is safe to have references into an `std::map` variable, right? Internal re-balancing of trees and such should not break any references, right?

LogicParrot:
Funnily enough, this *did* break something. Reverting.

tigerw:
oh. derp.

LogicParrot:
Mobs are immobile with this. I assumed the change is too trivial to break things and did not do enough testing, sorry.

tigerw:
Nevermind; doubt anyone noticed. Besides, it's going to be exactly what happens with either #2565 or #3115.

LogicParrot:
Minor correction: The multiverse will collapse once ##3115 is merged.

LogicParrot:
So, how could this have affected mobs? Bad `GetLayer`?

src/ChunkMap.cpp
```diff
@@ -1677,9 +1660,10 @@ void cChunkMap::RemoveEntity(cEntity * a_Entity)
 bool cChunkMap::ForEachEntity(cEntityCallback & a_Callback)
 {
 	cCSLock Lock(m_CSLayers);
-	for (cChunkLayerList::const_iterator itr = m_Layers.begin(); itr != m_Layers.end(); ++itr)
+	for (auto & itr : m_Layers)
 	{
-		if (!(*itr)->ForEachEntity(a_Callback))
+		cChunkLayer & Layer = itr.second;
+		if (Layer.ForEachEntity(a_Callback))
```
LogicParrot: Aha! I missed a `!`.

PR #3156, opened 2016-04-20
Updated vector hashing bit operations

LogicParrot:
@worktycho Can you confirm this was an issue?

### DIFF
--- a/src/Vector3.h
+++ b/src/Vector3.h
@@ -396,13 +396,11 @@ public:
 	/** Provides a hash of a vector's contents */
 	size_t operator()(const Vector3<What> & a_Vector) const
 	{
-		// Guaranteed to have no hash collisions for any 128x128x128 area
-		size_t Hash = 0;
+		// Guaranteed to have non repeating hashes for any 128x128x128 area
+		size_t Hash = static_cast<size_t>(a_Vector.y);
+		Hash <<= 16;
 		Hash ^= static_cast<size_t>(a_Vector.x);
-		Hash <<= 8;
-		Hash ^= static_cast<size_t>(a_Vector.y);
-		Hash <<= 8;
-		Hash ^= static_cast<size_t>(a_Vector.z);
+		Hash ^= static_cast<size_t>(a_Vector.z) << 8;
 		return Hash;
 	}
 };


worktycho:
It shouldn't have been. And if we are going to change the hash function, we should think about whether we can improve the function. This hash wastes a lot of entropy.

LogicParrot:
It does waste entropy indeed, but it is only used by the Pathfinder in a local area and no entropy is needed beyond this. So this is not a problem at all. even uniqueness in a 64x64x64 area would have been fine. If you propose a hash function which is just as simple and fast, I wouldn't mind it. I am just saying we don't really need more entropy.

As for the issue: The idea is to simply take the first 8 bits of coordinate vector and put those 24 bits next to each other. The way I see it is: Since there was no `& 0x7`, this wasn't really happening, we were taking more than 8 bits from each coordinate and therefore the xor was probably doing something weird.

tigerw:
If it is intended to be confined to Pathfinder, perhaps you could move the hash there? It may cause confusion if left with more general codebase accessibility.

LogicParrot:
@tigerw It is probably useful for the rest of the codebase, since vectors are always used in a local area. (Except when some people use them to represent chunks, for some enigmatic reason :P )

LogicParrot:
Edit: Swapped 0x7 with 0xFF

worktycho:
As I suggested on #2565, I think we should look at this function:
```
size_t Hash = 0;
Hash ^= static_cast<size_t>(a_Vector.x);
Hash <<= sizeof(size_t) * CHAR_BIT / 2;
Hash ^= static_cast<size_t>(a_Vector.z);
Hash ^= (static_cast<size_t>(a_Vector.y) & 0x0F)<< (CHAR_BIT *sizeof(size_t) - 4);
Hash ^= (static_cast<size_t>(a_Vector.y) & static_cast<size_t>(~static_cast<What>(0x0F))) << (CHAR_BIT *sizeof(size_t) / 2 -4);
```

Its slightly more complicated, and is 9 operations after constant folding as opposed to 5 for the proposed hash here, but it adapts to make maximum use of size of size_t across all systems, improving performance on 64 bit systems significantly.

LogicParrot:
- Does it have the no 256x256x256 collision guarantee ?
- Since it has more operations, why do you think it'd improve performance? Remember, we only use the hash in spacially close areas, so having less collisions in the general case is useless.

Pokechu22:
If it helps, the normal server hashes positions to a 32-bit int with this code:

```java
return (this.getY() + this.getZ() * 31) * 31 + this.getX();
```

worktycho:
Thats a pretty bad hash function. Its clearly written by someone who just copied a generic hash combining function.

worktycho:
Not only is it worse in distribution, its also slower than mine, taking 10 cycles, excluding loads) on Skylake and Sandy Bridge compared to 4/8 for mine, and 4/4 for the current hash. (Mine and the current uses simpler instructions + ILP, so manages to fit twice as many instructions in less time.).

worktycho:
@LogicParrot:7
 - It depends on size_t. If size_t is at least 24 bits long, than it will provide a no collision guarentee for any 256x256x256 block starting on a multiple of (256,256,256). If size_t is at least 27 bits long, then it won't collide in any 256x128x256 block, or any 256x256x256 block where the y range is aligned with a multiple of 256. At 32 bits its any 2048x128x2048 block, or 2048x256x2048 with y range alignment. At 64 bits, its any 134217728x126x134217728 block or 134217728x256x134217728 block with y alignment.

worktycho:
So it provides equivelent or better collision guarantees that the current one on any system where size_t at least 24 bits long. And better than the propesed modification on  any system where size_t at least 27 bits long.

worktycho:
As for speed, assuming the data is already in the registers, mine the same speed as the current one on Intel Skylake, half the speed of current one on Intel Sandy Bridge, and as far as I can tell has just over half the speed (8 cycles vs 5) on ARM11. 
Compared to your proposal, its slightly faster on Skylake(4 vs 5 cycles),  slightly slower on Sandy Bridge (8 vs y cycles), and slightly slower on ARM11 (8 vs 7 cycles)

worktycho:
If you want max performance on your proposal, try:
size_t Hash = 0;
Hash ^= static_cast<size_t>(a_Vector.x) & 0xFF;
Hash <<= 16;
Hash ^= static_cast<size_t>(a_Vector.y) & 0xFF;
Hash ^= static_cast<size_t>(a_Vector.z) & 0xFF << 8;
This saves one cycle on most intel processors, as they can execute shift on z in parallel with the and on y.

worktycho:
It appears the Intel timings for the two proposals and current covers all Intel Core 2 and later processors. On Pentium II, III and M, Core Solo and Core Duo, its 5 cycles for mine, 5 for LogicParrots proposal, and 4 cycles for the current version. On Pentium 4 its 8 cycles for mine, 13 for LogicParrots, and 12 for current (Pentium 4 has a really slow shift right op). On 64 bit capable Pentium 4, its 4 cycles for mine, 5 for Logic Parrots, and 4 for current.

If anyone has any other processors they'd like me to do cycle timings for I'll look into it, but generally mine is faster on OO processors than this proposal, due to better ILP and shorter critical path, and slightly slower on non-OO processors, due to the additional instructions.

LogicParrot:
Sorry, this PR is completely useless. I forgot that shift-left does zeroing out and for some reason thought the bits wrap around. So this adds no benefit over the original, at all.

LogicParrot:
@worktycho I'm interested

LogicParrot:
>This saves one cycle on most intel processors, as they can execute shift on z in parallel with the and on y.

That's deep.

@worktycho I'm interested in seeing how my next commit compares to the original, do you think it's practically the same?

LogicParrot:
There. It's functionally the same, isn't it?

LogicParrot:
" its 5 cycles for mine" mine as in the original or the one with the comment "This saves one cycle on most intel processors, as they can execute shift on z in parallel with the and on y."?

worktycho:
I was assuming that the first xor gets removed, as `0 ^ x => x` is a fairly easy optimization. So my cycle counts still stand.

LogicParrot:
I see. Well, I'll leave this to you. Merge whatever you think is suitable, noting that the only measure of a good hash for our purposes is: The fastest hash that works without collisions for a small area like 128x128x128, we literally don't care if it's better or worse on larger areas.

LogicParrot:
Is replacing xor with an or impactful in any way? I'd guess not, but you're the compiler expert.

worktycho:
Ors run in the same time as xor on pretty much all processors. But if we want maximum speed, how about this variation on the current:
```
size_t Hash = 0;
Hash ^= static_cast<size_t>(a_Vector.x);
Hash <<= 16;
Hash ^= static_cast<size_t>(a_Vector.z);
Hash ^= static_cast<size_t>(a_Vector.y) <<8;
```
No collisions in any 256x128x128 area, or 256x256x256 on if the y is 0-255, and it executes in three cycles on modern Intel processors and 4 on ARM11, saving a cycle over our current version in both cases.

LogicParrot:
Performance aside, the problem with the current hash that this PR aimed to fix is that it didn't really meeet the "doesn't collide in a 128x128x128" area condition, and so does your last proposal. Counterexample:
Vector1: (0,0,0), Hash is (0x000000)
Vector2: (8,1,0), Hash2:(0x000000)

LogicParrot:
For your last proposal the counterexample would be:
Vector1: (0,0,0), Hash is (0x000000)
Vector2: (0,1,8), Hash is (0x000000)

worktycho:
(0, 1,8) hashes to 0x0108? Both with current and my alternative.

LogicParrot:
And I have a counter example for your initial proposal:

```
size_t Hash = 0;
Hash ^= static_cast<size_t>(a_Vector.x);
Hash <<= sizeof(size_t) * CHAR_BIT / 2;
Hash ^= static_cast<size_t>(a_Vector.z);
Hash ^= (static_cast<size_t>(a_Vector.y) & 0x0F)<< (CHAR_BIT *sizeof(size_t) - 4);
Hash ^= (static_cast<size_t>(a_Vector.y) & static_cast<size_t>(~static_cast<What>(0x0F))) << (CHAR_BIT *sizeof(size_t) / 2 -4)
```
Vector1: (0,0,0), Hash1: 0x000000
Vector2: (1, 0, M), where M is a number having one bit sit to 1 such that when x is shifted, they align. Hash2: 0x000000.

LogicParrot:
>(0, 1,8) hashes to 0x0108? Both with current and my alternative.

How? Where did I get my reasoning wrong?
```
size_t Hash = 0;
Hash ^= static_cast<size_t>(a_Vector.x);  // Does nothing.
Hash <<= 16;   // Does nothing
Hash ^= static_cast<size_t>(a_Vector.z);  // Hash is now 0x000800
Hash ^= static_cast<size_t>(a_Vector.y) <<8;  // right hand side is now 0x000800,
//we just xored 0x000800 with 0x000800, result is zero.
```

LogicParrot:
Sorry, this example is flawed, I see it. Give me a moment.

worktycho:
Hash ^= static_cast<size_t>(a_Vector.z);

worktycho:
The worst case for my proposal is (1, 0, 0) and (0, 256, 0). Whilst these two values conflict, it requires a y of 256.

LogicParrot:
Proper counterexample: Vector1(0,0,0). Vector2(0, 1, 0x000800)

LogicParrot:
Ah, I see, now Y is too high.

LogicParrot:
Okay, these are too far apart. Sorry for the mess.

LogicParrot:
@worktycho I've updated this PR to reflect your last proposal, with a minor change: Y is the coordinate that is shifted left by 16, and not X. The reason is that on machines with a size_t of 16, we'd rather lose the Y and not the X, since in practice, Y provides the least entropy  (Pathfinder is often on a uniform height).

worktycho:
Looks good, merge.

PR #3164, opened 2016-04-24
Fixed some logout crashes

LogicParrot:


SphinxC0re:
Looks good to me

src/ClientHandle.cpp
```diff
@@ -174,6 +173,8 @@ void cClientHandle::Destroy(void)
 		cWorld * World = m_Player->GetWorld();
 		if (World != nullptr)
 		{
+			m_Player->StopEveryoneFromTargetingMe();
```
LogicParrot: The key change is this line.

SphinxC0re: Because when a player is logged out, atm mobs are still targeting him?

tigerw: #weakptrmasterrace

LogicParrot: @SphinxC0re yes.

PR #3165, opened 2016-04-24
ChunkLayers now stored in std::map [revised]

LogicParrot:
A fixed version of #3153 , which was merged and reverted.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -51,24 +51,10 @@ cChunkMap::cChunkMap(cWorld * a_World) :
 
 
 
-cChunkMap::~cChunkMap()
-{
-	cCSLock Lock(m_CSLayers);
-	while (!m_Layers.empty())
-	{
-		delete m_Layers.back();
-		m_Layers.pop_back();  // Must pop, because further chunk deletions query the chunkmap for entities and that would touch deleted data
-	}
-}
-
-
-
-
-
 void cChunkMap::RemoveLayer(cChunkLayer * a_Layer)
 {
 	cCSLock Lock(m_CSLayers);
-	m_Layers.remove(a_Layer);
+	m_Layers.erase(std::pair<int, int>(a_Layer->GetX(), a_Layer->GetZ()));
 }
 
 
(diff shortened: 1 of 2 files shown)


SphinxC0re:
Looks good to me

src/ChunkMap.cpp
```diff
@@ -1677,9 +1660,10 @@ void cChunkMap::RemoveEntity(cEntity * a_Entity)
 bool cChunkMap::ForEachEntity(cEntityCallback & a_Callback)
 {
 	cCSLock Lock(m_CSLayers);
-	for (cChunkLayerList::const_iterator itr = m_Layers.begin(); itr != m_Layers.end(); ++itr)
+	for (auto & itr : m_Layers)
 	{
-		if (!(*itr)->ForEachEntity(a_Callback))
+		cChunkLayer & Layer = itr.second;
+		if (!Layer.ForEachEntity(a_Callback))
```
LogicParrot: For the record, the problem was that I forgot a `!` in #3163 , this prevented mobs from ticking.

SphinxC0re: I know. Isn't that ridiculous :smile:

LogicParrot:
Merging, but I just noticed another problem in master: `RemoveLayer` is never called! This means that `m_Layer` can only grow. I'll treat this in another PR.

src/ChunkMap.cpp
```diff
@@ -78,23 +64,19 @@ void cChunkMap::RemoveLayer(cChunkLayer * a_Layer)
...
 	// Not found, create new:
-	cChunkLayer * Layer = new cChunkLayer(a_LayerX, a_LayerZ, this, *m_Pool);
-	if (Layer == nullptr)
-	{
-		LOGERROR("cChunkMap: Cannot create new layer, server out of memory?");
-		return nullptr;
-	}
-	m_Layers.push_back(Layer);
-	return Layer;
+	m_Layers.emplace(std::piecewise_construct,
```
LogicParrot: Emplace return an iterator, I can get rid of the m_Layers.find  below, and gain some more performance.

src/ChunkMap.cpp
worktycho: Its also ridiculous that it wasn't picked up in CI. We need better testing.

PR #3176, opened 2016-05-01
Fixed crash on stopping the server

LogicParrot:


src/World.cpp
```diff
@@ -215,8 +215,6 @@ cWorld::~cWorld()
 	delete m_LavaSimulator;      m_LavaSimulator     = nullptr;
 	delete m_RedstoneSimulator;  m_RedstoneSimulator = nullptr;
 
-	UnloadUnusedChunks();
```
LogicParrot: Can anyone confirm that this line is completely unnecessary?

LogicParrot: - All chunks are already saved
- the chunkmap is about to be reset, which will destroy all chunklayers, which all unload/destroy all chunks. 

Seems redundant to have this line.

tigerw: It's unnecessary.

tigerw:
For the record, #3115.

PR #3183, opened 2016-05-10
Removed interface cast

estradjm:
Removed the cast from interface to specific class in BlockTrapdoor.h


worktycho:
Downcasting like that implicitly is not allowed by the language.

PR #3188, opened 2016-05-14
Update ISSUE_TEMPLATE

SphinxC0re:
I hope I didn't forget anything @cuberite/core-team what do you guys think?

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,3 +1,7 @@
+Client version:   x.x.x
+Commit id:        (Found at the top of the server's console output)
+
+
 ### Expected behavior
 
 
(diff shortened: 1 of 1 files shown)


LogicParrot:
`Client version` and `Operating system` are rarely important. I wouldn't want to make people type uneeded info. For the few issues in which these matter (Some protocol issues need client version, Some compile issues need OS), we can simply ask. 

`Commit id` is a good idea. But placing it in `Steps to reproduce the behavior` forces you to have a `procedure` subsection. I would have put it on top of the template or somewhere outside `Steps to reproduce the behavior` and avoid the subsection.

SphinxC0re:
There might be a lot of Issues just because someone is using the 1.7 client or a new version of Minecraft comes out and we aren't compatible yet. So I believe that the client version *is* important to eliminate some false friends. The OS is not really necessary, I agree

PR #3199, opened 2016-05-16
Some 1.9 Crafting Recipes and Items.

bearbin:
I haven't added all of the crafting recipes but I tried to add most of them. 

Some of the bits which were out of order have been reordered.

PR #3202, opened 2016-05-17
Clean up items.ini

bearbin:
underscores in the names of items are now ignored, and any items with that name are stripped from the file.

This pull request depends on https://github.com/cuberite/cuberite/pull/3199


mathias-github:
I added the underscores, because vanilla Minecraft uses them. Otherwise, give commands taken from the internet would be broken.

madmaxoft:
I'm not sure this is the right approach - what if future MC actually requires the use of underscored names for some reason? Such as advanced NBT tags for composite items, or whatever.

bearbin:
Would you rather items.ini contained only underscored names, and the server just did a underscore-insensitive compare to those?

madmaxoft:
I'm not sure. Is there a possibility that two underscored names can compare to the same underscore-less one?
Also, I think commands such as `/give o_akdo_or` should fail.

bearbin:
So in your eyes the correct solution would be to just edit items.ini so it contained both underscored and non-underscored names?

Does anybody else have an opinion on this matter?

madmaxoft:
Yes, I think having both is the way to go for now.

tigerw:
That's just inviting de-synchronisation. Recommended solution is to remove non-underscored names.

SphinxC0re:
Yeah, just let the server load all items at server start and for every item:
 - save it as it is 
 - remove the underscores and save it again

madmaxoft:
That would be a workable solution.

SphinxC0re:
Also, are the names check case-sensitive? Because I think this would be bad UX

madmaxoft:
@bearbin Any chance of finalizing this, based on @SphinxC0re's suggestions? (https://github.com/cuberite/cuberite/pull/3202#issuecomment-221207962)

SphinxC0re:
About that: By "saving" I didn't mean to actually save the results to disk. Just replace it with "caching" 😉

Issue #739, opened 2014-03-01
The build system should use FindLua51 to load the system lua if present


worktycho:
Cmake provides a function FindLua51 to locate the system lua install if installed. We should use this function to find the system lua and use that instead of our own if installed. This should be bypassed when the CROSS_COMPILING flag is set so as to avoid dependencies on the build machine.

PaddyMac:
I was thinking about opening a new bug, but then I found this open bug. I'd actually like to see the option to use system libraries for all of the dependencies if possible - maybe with a cmake commandline switch.

PR #3204, opened 2016-05-17
Added a Pure-Lua script for bindings generation.

madmaxoft:
This script can be opened in ZeroBraneStudio and debugged from there.


tigerw:
What does BindingsProcessor do? It doesn't look like a replacement for tolua.exe.

madmaxoft:
Originally the file provided class-access recognition (`public`, `protected`, `private`) and was loaded from inside ToLua++ to do so.

Now, in addition to that, it generates "documentation" for all the symbols (more like description, really - it lists all the classes, functions, variables, enums and constants, their types, function parameter types, function return types, and all the associated doxycomments) and even mimics the tolua++.exe's behavior, so that we can drop the executable from builds (as long as system Lua is present).

The purpose of tolua++.exe is to parse the commandline into a Lua table of parameters and then invoke Lua scripts for the actual parsing. The same thing is now achieved in the BindingsProcessor.lua script file - it fills the parameters (because we use constant values, there's no need for cmdline parsing) and invokes the ToLua++'s Lua scripts for parsing.

This is the source for the tolua++.exe: https://github.com/cuberite/toluapp/blob/44cd937e7687dadec80068116a2e1d683b139d06/src/bin/tolua.c
This is the `invokeToLua` function in this PR that replaces it: https://github.com/cuberite/cuberite/pull/3204/files#diff-04f8dfdd8705bba70347171e8618f444R37
These are the Lua scripts that are being invoked (the parser): https://github.com/cuberite/toluapp/tree/44cd937e7687dadec80068116a2e1d683b139d06/src/bin/lua

worktycho:
So if we combined this with #739 then we would not need to build a host lua when cross compiling. Which would probaby simplify cross-compiling significantly.

madmaxoft:
Exactly. We could make it a precondition - for cross-compiling, you need to have system Lua available.

Before merging this, I'd like to explore a few more options, so please don't merge this in yet.
Here's a forum discussion for this topic: https://forum.cuberite.org/thread-2448.html

madmaxoft:
I now consider this thing complete.
- DoxyComments have been extracted and added to the LuaAPI docs
- Fixed a few Lua APIs
- System Lua interpreter is used instead of ToLua++, if available
- Silenced an obnoxious CMake warning

PR #3211, opened 2016-05-22
compile.sh: added thread selection; added command line switches

### LINKED ISSUE #3014: compile.sh todo
Todo:
- [ ] Ask the user about how many threads they'd like to use.  If the user does not make a choice, resort to a sane default. Perhaps using some variant [of this](https://forum.cuberite.org/thread-2332-page-2.html#pid25009).
- [ ] Add commandline arguments. This allows convenient automatic unattended updates.

Schwertspize:
closes #3014

For convinience i used shell variables.

for example:

    BUILDTYPE=Debug BRANCH=master THREADS=4 ./compile.sh

all of these can be ommitted. if THREADS is not set, it will automagically be set. if BRANCH is not set, it will be asked, but because we lack of other branches, choosen automatically. if BUILDTYPE is not set, it will be asked. if both BUILDTYPE and BRANCH are set, it will omit the greeting, because the script will then run without user input and we dont need to spam logs with a greeting right? (I assume that people who use that use it in scripts).


bearbin:
compile.sh: 6: compile.sh: Syntax error: newline unexpected

Please test your code.

mathwhiz1212:
For me, if I run the suggested command: `BUILDTYPE=Debug BRANCH=master THREADS=4 ./compile.sh` or `BUILDTYPE=Debug BRANCH=master ./compile.sh` It will ask me for the "compile mode" and will not compile if executed from a script.

It looks way better than what I was doing :) 

```
Choose compile mode:
* (N)Normal:  Compiles normally.
              Generates the fastest build.

* (D)Debug:   Compiles in debug mode.
              Makes your console and crashes more verbose.  
              A bit slower than Normal mode. If you plan to help
              development by reporting bugs, this is preferred.
		

Note that the script will connect to the internet in order to fetch
code after this step. It will then compile your program.
```

Schwertspize:
Dah I noticed myself yesterday. I messed with the variables. 

You might also see that I messed with shell compatibility, I am not sure if everything is posix compatible

Schwertspize:
well, i fixed them, i think, and afaik this **is** POSIX compatible.

compile.sh
```diff
@@ -33,96 +25,190 @@ missingDepsExit ()
...
+	$GCC_EXE_NAME --help > /dev/null 2> /dev/null && GCC_EXISTS=1
+	$CLANG_EXE_NAME --help > /dev/null 2> /dev/null && CLANG_EXISTS=1
+	if [ "$GCC_EXISTS" -eq 0 ] && [ "$CLANG_EXISTS" -eq 0 ]; then
+		MISSING_PACKAGES=" $COMPILER_PACKAGE_NAME"
+	fi
 
-# Depdendency check.
-checkPackages ()
+	# Depdendency check.
+	checkPackages ()
```
LogicParrot: This function in nested. Is this POSIX?

compile.sh
```diff
@@ -33,96 +25,190 @@ missingDepsExit ()
...
-Compiling from source takes time, but it usually generates faster
-executables. If you prefer ready-to-use binaries or if you want
-more info, please visit:  http://cuberite.org/"
+greeting ()
+{
+	# Do not greet when running with arguments.
+	if [ "$BRANCH" = "" -a "$BUILDTYPE" = "" ]; then
+		# Do we already have a repo?
+		if [ -d .git ] && [ -f easyinstall.sh ] && [ -f src/BlockArea.cpp ]; then # A good enough indicator that we're in the Cuberite git repo.
+			cd ../
```
LogicParrot: This `cd` is needed even when there is no greeting.

Schwertspize: Thanks! I didn't notice that, I just thought it was a message.

compile.sh
Schwertspize: Should be POSIX because such functions are basically just the definition of a 'comand' (like a subscript with a subshell) 

Please correct me if it's not POSIX

LogicParrot: Okay. But I think it should be un-nested anyways. Nesting seems to serve no purpose here.

Schwertspize:
I added $DIRECTORY with which you can specify either the path to a cuberite directory or, as git accepts it, the path to a nonexisting or empty directory. it will perform any actions to there, like cloning there. unnested the dependency check function and fixed the directory proof checks. added an additionally check before trying to git checkout (the same as we already have).

Schwertspize:
@LogicParrot any review? Is this mergable? 

And if this get merged, we should include documentation about this somewhere in the book and in here that says explains the usage for scripts and maybe for normal usage too (THREADS?).

LogicParrot:
Let me check it this weekend.

Most Gnu/Linux scripts have built-in documentation, triggered when passing a `--help` / `-h` parameter or passing the wrong parameters.

Schwertspize:
I know but that way we would need to parse arguments.... Anyway, should I include that if $1 is nonzero (the user wanted to Parse arguments, which could be `./compile.sh anything` or `./compile.sh -h`) it outputs a documentation?

Schwertspize:
@LogicParrot any news?

tigerw:
LogicParrot said that he'll be busy, so if it works we could merge now. We do have a very enlivening [thread](https://forum.cuberite.org/thread-2363.html) about stagnant PRs.

Schwertspize:
I know there is this thead, I followed it and lastly posted a note ;) 

Well, so tomorrow I will update the branch and do some basic help if you pass an argument.

LogicParrot:
I apologize for the delay. I have not abandoned this. I will check it as soon as I finish my exams. @sphinxc0re @tigerw please do not merge impulsively without good testing :P

I glanced this *very* quickly, and there is some code duplication, I don't like that and it is avoidable.

compile.sh
```diff
@@ -33,96 +25,194 @@ missingDepsExit ()
...
+	fi
+}
+
+selectThreadsToUse ()
 {
-	# note that IFS is a TAB here!
-	echo "$PROGRAMS" | while IFS='	' read EXE_NAME PACKAGE_NAME __trash__; do
-		command -v $EXE_NAME > /dev/null 2> /dev/null || printf %s " $PACKAGE_NAME"
-	done
+	osName=`uname`
```
LogicParrot: is "Uname" POSIX? If not, this breaks non Linux compilations.

LogicParrot: Perhaps 
```
osNamew=`uname || echo other`
```

compile.sh
```diff
@@ -33,96 +25,194 @@ missingDepsExit ()
...
+
+	if [ "$osName" = "FreeBSD" ]; then
+		THREADS=`sysctl hw.ncpu | awk '{print $2}'`;
+	elif [ "$osName" = "Linux" ]; then
+		THREADS=`nproc`;
+	else
+		wCount=`whereis nproc | awk '{print $2}' | wc -m`;
```
LogicParrot: Overkill.
```
THREADS=`nproc 2>/dev/null || echo 2`
``

File: compile.sh [hunk 5]
```diff
@@ -33,96 +25,194 @@ missingDepsExit ()
...
-	echo "You have missing compilation dependencies:"
-	echo $MISSING_PACKAGES
+compile ()
+{
+	# Echo: Downloading began.
 	echo
+	echo " --- Downloading Cuberite's source code from the $BRANCH branch..."
+
+
+	if [ -d $DIRECTORY ] && [ -d $DIRECTORY/.git ] && [ -f $DIRECTORY/easyinstall.sh ] && [ -f $DIRECTORY/src/BlockArea.cpp ]; then # A good enough indicator that we're dealing with the Cuberite git repo.
```
LogicParrot: Please avoid this dupe.

LogicParrot:
This uses shell variables and not command line arguments, this is very non standard. Is there a reason for that?

Schwertspize:
It's way easier to parse and way easier to use :) 

But that's just my opinion 

As for most of your comments: That was original code, I did change nothing. For example, the dupe was very early in the history, I just moved it away (and copied it). Same applies to nproc or uname. (well, we don't even need uname if we handle the nproc differently. Let's see what I can do.

compile.sh
```diff
@@ -33,96 +25,194 @@ missingDepsExit ()
...
+	echo "If you're updating, you should run: <Path to Cuberite>/compile.sh"
+	echo "Compiling from source takes time, but it usually generates faster"
+	echo "executables. If you prefer ready-to-use binaries or if you want"
+	echo "more info, please visit:  http://cuberite.org/"
+}
+
+checkRepository ()
+{
+	# Do we already have a repo?
+	if [ -d .git ] && [ -f easyinstall.sh ] && [ -f src/BlockArea.cpp ]; then # A good enough indicator that we're in the Cuberite git repo.
```
LogicParrot: No, the dupe is new. This is the second time this line is present.

LogicParrot:
I submitted a PR with command line arguments. See #3255

LogicParrot:
>It's way easier to parse and way easier to use :) 

Parsing may be slightly easier, but I don't see how this easier **for users**.  The problem is that this is extremely uncommon and users will be confused. I prefer #3255

LogicParrot:
>For example, the dupe was very early in the history, I just moved it away (and copied it). Same applies to nproc or uname. (well, we don't even need uname if we handle the nproc differently. Let's see what I can do.

I do not see `nproc`, `uname`, or a douplication of the following line, in the current `compile.sh`.

```if [ -d .git ] && [ -f easyinstall.sh ] && [ -f src/BlockArea.cpp ]```

Schwertspize:
Do as you want to. 

These things were what was left from open issues.

LogicParrot:
Thanks for the effort, though :)

PR #3217, opened 2016-05-28
Add getTeamNames to cScoreboard

### LINKED ISSUE #3210: Lua function to get list of names of all the teams
Currently I'm trying to get rid of teams that do not appear in my database as well. As `cScoreboard:RemoveTeam()` requires a name of the team, which I don't know because it doesn't exist in my database, I have no way to remove said team. A way to retrieve a list of all the names of the teams would allow me to compare it to my database and thus fix the problem.

PlasmaPower:
Resolves #3210

### DIFF
--- a/Server/Plugins/Debuggers/Debuggers.lua
+++ b/Server/Plugins/Debuggers/Debuggers.lua
@@ -2184,3 +2184,16 @@ end
 
 
 
+
+function HandleTeamsCmd(a_Split, a_Player)
+	local Scoreboard = a_Player:GetWorld():GetScoreBoard()
+
+	a_Player:SendMessage("Teams: " .. table.concat(Scoreboard:GetTeamNames(), ", "))
+
+	return true
+end
+
+
+
+
+
--- a/src/Scoreboard.cpp
+++ b/src/Scoreboard.cpp
@@ -416,6 +416,22 @@ cTeam * cScoreboard::GetTeam(const AString & a_Name)
 
 
 
+AStringVector cScoreboard::GetTeamNames()
+{
+	AStringVector TeamNames;
+
+	for (const auto & Team: m_Teams)
+	{
+		TeamNames.push_back(Team.first);
+	}
+
+	return TeamNames;
+}
+
+
+
+
+
 cTeam * cScoreboard::QueryPlayerTeam(const AString & a_Name)
 {
 	cCSLock Lock(m_CSTeams);
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -2196,6 +2196,7 @@ end
 				GetObjective = { Params = "string", Return = "{{cObjective}}", Notes = "Returns the objective with the specified name." },
 				GetObjectiveIn = { Params = "DisplaySlot", Return = "{{cObjective}}", Notes = "Returns the objective in the specified display slot. Can be nil." },
 				GetTeam = { Params = "string", Return = "{{cTeam}}", Notes = "Returns the team with the specified name." },
+				GetTeamNames = { Params = "", Return = "array table of strings", Notes = "Returns the names of all teams" },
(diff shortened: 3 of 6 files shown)


bearbin:
Your build failed, but I'm not exactly sure why.

madmaxoft:
Travis seems offline for maintenance, and Appveyor seems to have had some connection problems. I've restarted the Appveyor build

Server/Plugins/Debuggers/Debuggers.lua
```diff
@@ -2184,3 +2184,16 @@ end
 
 
 
+
+function HandleTeamsCmd(a_Split, a_Player)
+	local Scoreboard = a_Player:GetWorld():GetScoreBoard()
+
+	a_Player:SendMessage("Teams: " .. table.concat(Scoreboard:GetTeamNames(), ","))
```
madmaxoft: Add a space after the concatenating comma, `", "))`

src/Scoreboard.cpp
```diff
@@ -416,6 +416,22 @@ cTeam * cScoreboard::GetTeam(const AString & a_Name)
 
 
 
+AStringVector cScoreboard::GetTeamNames()
+{
+	AStringVector Teams;
+
+	for (auto it : m_Teams)
```
madmaxoft: Don't use `it` for things that are not iterators. I think `for (const auto & Team: m_Teams)` would be better.

madmaxoft:
Except for the trivial style issues, this seems good. Fix those and we can merge.

Thanks :)

PureTryOut:
@PlasmaPower thanks for this, I'm eagerly waiting for it to be merged!

PlasmaPower:
@madmaxoft I've pushed the new styling. As for the travis build, it looks like clang is treating an unused function as an error, whereas gcc is treating it as a warning. That's not in the PR though.

madmaxoft:
Ah, right, an unused function. That is an error we don't want, so it's set as an error in clang.

The problem is that you put the new function in `Scoreboard.h` into a block that is exported to Lua automatically, but the automatic export doesn't understand `AStringVector`. Either move it out of a `// tolua_begin ... // tolua_end` block, or add a `// tolua_end` above the function and `// tolua_begin` below it.

PlasmaPower:
@madmaxoft Even when I move it out of the tolua block, `tolua_collect_AStringVector` is still defined but never used. Could it be a problem with the incremental build?

madmaxoft:
When you move it out of the block, it should trigger a Bindings.cpp regeneration, check your build log that it did. If it didn't, start it manually by deleting the Bindings.cpp file.

I'll try locally to see where the problem is.

PlasmaPower:
@madmaxoft Well Travis is succeeding, so it looks like the problem is gone (probably something with the build system).

madmaxoft:
Works for me locally.

PlasmaPower:
@madmaxoft I'm still getting this warning locally, but it shouldn't matter since the problem would only occur on my machine. It still might be helpful to fix for future occurrences though:

```
/home/lee/cuberite/src/Bindings/Bindings.cpp:190:12: warning: ‘int tolua_collect_AStringVector(lua_State*)’ defined but not used [-Wunused-function]
 static int tolua_collect_AStringVector (lua_State* tolua_S)
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

madmaxoft:
Did you try deleting the file before building?

PlasmaPower:
@madmaxoft That resolves the issue, but I'd expect the root cause still remains. Just something to look into I guess. It only caused a warning in this case though, I'm not sure if it could cause an error.

madmaxoft:
I think I found the cause of why the bindings weren't regenerated. It seems the `AllToLua.pkg` file contains references to headers that are not listed in `CMakeLists.txt` as the dependencies of the bindings. Thanks for the report.

PlasmaPower:
@madmaxoft Glad I could help!

PR #3218, opened 2016-05-29
Fixed explosions not causing some blocks to drop their broken variants (such as stone dropping cobblestone)

### LINKED ISSUE #3194: Creeper explosions cause grass blocks to drop themselves rather than dirt blocks
Client version:   1.9.4
Commit id: a4f327118b06ced1cd4510b7d20d34da83aa78a3
### Expected behavior

Grass drops will turn to dirt blocks in an explosion
### Actual behavior

They drop grass blocks

Have a creeper blow up on some grass and dirt

ahyattdev:
Fixes #3194 

Blocks included:
Stone -> Cobblestone
Grass Blocks -> Dirt

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1853,7 +1853,18 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 							// No pickups for air
 							break;
 						}
-
+						case E_BLOCK_GRASS:
+						{
+							// Change the grass block to dirt
+							area.SetBlockType(bx + x, by + y, bz + z, E_BLOCK_DIRT);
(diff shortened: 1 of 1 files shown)


bearbin:
Have you tested this? Reading the code, I don't think this does what you think it does.

Also, line 1875 should already be doing this, maybe that function needs to be updated instead?

madmaxoft:
@bearbin is right, line 1875 should be doing it, so if it isn't, it's a bug there.

ahyattdev:
I did test it, and thought it worked because grass then dropped dirt. 

The real reason it worked was because someone had fixed the issue, but did not report it on the GitHub issue for this bug. However, I still should have tested it on the latest version of Cuberite to see if it was still an issue. 

Lesson learned.

Issue #3077, opened 2016-03-11
Dispenser not placing boat/minecart/spawnegg


Schwertspize:
Note: this is not the current master, Its at lest one commit behind (that one was a webadmin fix I think, here commit sha: 39144480f588b60c246fbe0451b8002eaa6a0797) 

Dispensers should place some items like boats, minecarts or monsters (when dropping spawn eggs).

QUSpilPrgm:
This is fixed with #3219

PR #3219, opened 2016-05-29
Update dispenser code to let them act more like in Vanilla

QUSpilPrgm:
- Added code to make bonemeal, potions, minecarts, XP bottles and boats work inside dispensers.
- Dispensers are now able to place TNT if the block is transparent but not air. (like in Vanilla)
- Added return value that indicates the success of pumpkin, melon, sugarcane and cactus growing functions.
- Changed return value of "GrowRipePlant" so that it actually indicates if the block was able to grow. 
  This is important now because otherwise dispensers lose bonemeal even if the plant is already fully grown
- Fixed "GrowSugarcane" and "GrowCactus" in "GrowRipePlant" so that it only grows them a single. block.  Why did it grow them to max blocks before? I guess it could be anything since this is not a Vanilla feature but that still seemed very unlogical to me.
  If there is a desire to keep it the way it was please let me know.

This fixes issue #3077.

Since I am not familiar with this code base there might be style issues or code constraints I don't know about.
I will happily to fix everything. I hope it is not too bad :)


QUSpilPrgm:
The build failed because "IsMinecart" and "SpawnBoat" is undocumented.
In the source code they are just like all the other functions. Do I need to document them somewhere else?

bearbin:
The functions you created are exported to the Lua API. They need to be documented in APIDump. The documentation is located in Server/Plugins/APIDump/APIDesc.lua

src/BlockEntities/DispenserEntity.cpp
```diff
@@ -39,177 +41,244 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
...
+	if (ItemCategory::IsMinecart(SlotItem.m_ItemType) && IsBlockRail(DispBlock))  // only actually place the minecart if there are rails!
 	{
-		case E_ITEM_BUCKET:
+		if (m_World->SpawnMinecart(BlockX + 0.5, DispY + 0.5, BlockZ + 0.5, SlotItem.m_ItemType) != cEntity::INVALID_ID)
 		{
-			LOGD("Dispensing empty bucket in slot %d; DispBlock is \"%s\" (%d).", a_SlotNum, ItemTypeToString(DispBlock).c_str(), DispBlock);
-			switch (DispBlock)
+			m_Contents.ChangeSlotCount(a_SlotNum, -1);
+		}
+	}
```
madmaxoft: There's no code following this, so how about bailing out with a `return`, so that the nesting level for the rest of the code is lower?

madmaxoft:
Looks like we got a new interested coder, welcome aboard :)

QUSpilPrgm:
Thanks :)
I tried using Curerite the other day and ran into a large amount of issues especially related to redstone. I have compiled a list of all things that didn't work or need to be changed. Many of them seemed to be rather low priority but not too complicated to fix. So I decided to try doing it myself.

If my patches are welcome I will probably upload more fixes over the next weeks.

SphinxC0re:
Looks good to me.

Could you please squash and rebase your commits onto the current master (e.g. origin/master)

QUSpilPrgm:
I have squashed the old commits but I have also added new code to make big flowers and tall grass bonemealable (like they are in Vanilla).

madmaxoft:
I'll accept it for now, but next time please put unrelated code changes each in its separate branch.
Now you've introduced a some new code that needs new reviewing and has new style issues, so it makes for more work in the end.

src/World.cpp
```diff
@@ -1940,9 +1945,56 @@ bool cWorld::GrowRipePlant(int a_BlockX, int a_BlockY, int a_BlockZ, bool a_IsBy
...
+			{
+				GetBlockTypeMeta(a_BlockX, a_BlockY - 1, a_BlockZ, BlockType, BlockMeta);
+				if (BlockType != E_BLOCK_BIG_FLOWER)
+				{
+					return false;
+				}
+			}
+			if ((BlockMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) ||
+				(BlockMeta == E_META_BIG_FLOWER_LARGE_FERN))  // tall grass and fern do not work
+			{
```
madmaxoft: Bad condition formatting. If you're splitting an `if` line, use this template:
```
if (
  condition1 ||
  condition2
)
{
  ...
```

QUSpilPrgm:
Thats actually what I wanted to do but for some reason it ended up being in here too.
I am still learning how to use Git.

SphinxC0re:
Could you please squash and rebase your commits onto the current master?

madmaxoft:
Don't squash, keep the two commits separate.

SphinxC0re:
Oh I'm sorry, I didn't see the two changes

SphinxC0re:
After that you do `git push -f`

src/BlockEntities/DispenserEntity.cpp
```diff
@@ -178,6 +189,64 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
...
+		case E_ITEM_BOAT:
+		{
+			Vector3d SpawnPos(BlockX, DispY, BlockZ);
+			// Vanilla allows the water to be in front of the dispenser or one block below
+			if (IsBlockWater(DispBlock))
+			{
+				SpawnPos += Vector3d(0.0, 1.0, 0.0);
+			}
+			else if (IsBlockWater(DispChunk->GetBlock(DispX, DispY - 1, DispZ)))
+			{}
```
madmaxoft: Isn't some code missing here?

src/BlockEntities/DispenserEntity.cpp
```diff
@@ -178,6 +189,64 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
...
+			if (SlotItem.m_ItemDamage != E_META_DYE_WHITE)
+			{
+				DropFromSlot(a_Chunk, a_SlotNum);
+				break;
+			}
+			if (m_World->GrowRipePlant(BlockX, DispY, BlockZ, true))
+			{
+				m_Contents.ChangeSlotCount(a_SlotNum, -1);
+			}
+			break;
```
madmaxoft: Doesn't bonemeal ever drop as a regular pickup in vanilla?

QUSpilPrgm: Bonemeal does indeed never drop in Vanilla from a dispenser.

Bonemeal is one of the many dyes. The bonemeal functions should only be used, if the item is actually bonemeal. If the item metadata is not "E_META_DYE_WHITE", than it is another dye which is not bonemeal.

src/BlockEntities/DispenserEntity.cpp
madmaxoft: It is somewhat misleading, having an empty code block. How about this?
```
Vector3d SpawnPos;
if (IsBlockWater(DispBlock))
{
  // Water next to the dispenser, spawn a boat above the water block
  SpawnPos.Set(BlockX, DispY + 1, BlockZ);
}
else if (IsBlockWater(DispChunk->GetBlock(DispX, DispY - 1, DispZ)))
{
  // Water one block below the dispenser, spawn a boat at the dispenser's Y level
  SpawnPos.Set(BlockX, DispY, BlockZ);
}
else
{
  // There's no eligible water block, drop the boat as a pickup
  DropFromSlot(a_Chunk, a_SlotNum);
  break;
}
```

src/BlockEntities/DispenserEntity.cpp
madmaxoft: Ok, just checking. It feels wrong to me that one can dispense any dye other than bonemeal, but if Mojang decided that way...

src/BlockEntities/DispenserEntity.cpp
QUSpilPrgm: Sounds reasonable

madmaxoft:
Thanks :)

PR #3220, opened 2016-05-29
LuaState: Inter-plugin calls now support simple tables.

madmaxoft:
Tables can contain strings, numbers, booleans, nils and other tables as both values and keys. In order to avoid an infinite loop when copying tables with loops, a fixed maximum nesting level is selected, tables that go over that depth fail to get copied.


tigerw:
No unexpected inter-thread non-atomic access this time? :)

madmaxoft:
Not unless they've been there all along before, the code path only builds upon the pre-existing code to copy simple values.

NiLSPACE:
Do we also want to support functions using string.dump? We can't bring the local variables with it though.

madmaxoft:
I can't imagine a scenario for which functions (without upvalues) would be the best solution. And moving functions around this way would be rather expensive, I think.

No, if you want to inject code into the destination plugin, there's a much easier way - have a look at hov the Debuggers plugin does it in this PR - it calls the `dofile` function via `cPluginManager:Call()` to load a file into the plugin.

Oh god, now that I look at it, I've forgotten to commit the new file that gets injected. I'll fix that asap.

madmaxoft:
File added in beb3660c42d93ef7386bb5158a65cb9b4fa62549

NiLSPACE:
dofile most likely also calls `loadstring` right? If so, then the performance would be similar if something like [WorldEdit's ExecuteString function](https://github.com/cuberite/WorldEdit/blob/master/Api/Manage.lua#L277-L288)

madmaxoft:
Yes, `dofile` calls `loadstring` internally.

The difference is that when you decide to inject the code, you only do it once. The code stays there until a plugin reload, so you can then make as many inter-plugin calls to that injected code as you want without the `loadstring` penalty.

NiLSPACE:
Anyway, we're going off track. I was wondering if this also transports the metatable.

madmaxoft:
No, it doesn't. I hadn't thought about those. And anyway they're usually full of functions, so that would be quite a performance hit.

NiLSPACE:
`__index` can also be a table though. Same as `__newindex` I believe.

PR #3224, opened 2016-06-04
Add an activation flag to the block metadata of droppers and dispensers

QUSpilPrgm:
This fixes an issue were droppers and dispensers would shoot an item on every block update they receive. For example previously a powered dispenser with a water bucket inside would repeatedly place the water and pick it up again while being continuously powered. The newly added metadata flag also exists in Vanilla minecraft.

While looking at the code I found a probably incorrect range check inside "cIncrementalRedstoneSimulator::Simulate". I have fixed this as well. Maybe someone who is more familiar with the code can confirm that it was indeed broken before.


QUSpilPrgm:
Why does it complain about the new undocumented "E_META_DROPSPENSER_ACTIVATED" metadata? As far as I can tell all the other metadata values are defined in the exact same way and are undocumented too.

src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -198,7 +198,7 @@ void cIncrementalRedstoneSimulator::Simulate(float a_dt)
 		{
 			BLOCKTYPE PotentialBlock;
 			NIBBLETYPE PotentialMeta;
-			if ((Location.y < 0) || (Location.y > cChunkDef::Height))
+			if ((Location.y < 0) || (Location.y >= cChunkDef::Height))
```
SphinxC0re: What about `IsValidHeight()` ?

Pokechu22:
Note that it _was_ vanilla behavior for them to shoot multiple times, although they patched that in 1.5 (see [MC-846](https://bugs.mojang.com/browse/MC-846)).  Still should be removed from cuberite now since it's no longer vanilla behavior, but it's worth noting.

Regarding `E_META_DROPSPENSER_ACTIVATED`, the other ones are currently in the [official undocumented list](http://apidocs.cuberite.org/_undocumented.lua), meaning that they're already known to be undocumented.  What you can do is document all of them (including the `E_META_DROPSPENSER_FACING` constants) on line 2990 of Server/Plugins/APIDump/APIDesc.lua.

QUSpilPrgm:
Thanks for your comments. :)
I have updated the code accordingly.

QUSpilPrgm:
Is this waiting for anything specific?
I think I have incorporated all suggestions.

src/BlockEntities/DispenserEntity.cpp
```diff
@@ -283,7 +283,7 @@ UInt32 cDispenserEntity::SpawnProjectileFromDispenser(int a_BlockX, int a_BlockY
 
 Vector3d cDispenserEntity::GetShootVector(NIBBLETYPE a_Meta)
 {
-	switch (a_Meta & 0x7)
+	switch (a_Meta & ~E_META_DROPSPENSER_ACTIVATED)
```
madmaxoft: I'd appreciate having a specific name for `~E_META_DROPSPENSER_ACTIVATED`, such as `E_META_DROPSPENSER_FACING_MASK`, this will make the code much easier to understand at a glance.

src/Simulator/IncrementalRedstoneSimulator/DropSpenserHandler.h
```diff
@@ -18,6 +18,20 @@ class cDropSpenserHandler : public cRedstoneHandler
...
+	{
+		return (a_Meta & E_META_DROPSPENSER_ACTIVATED) != 0;
+	}
+	inline static NIBBLETYPE SetActivationState(NIBBLETYPE a_Meta, bool IsOn)
+	{
+		a_Meta &= ~E_META_DROPSPENSER_ACTIVATED;  // clear the bit
+		if (IsOn)
+		{
+			a_Meta |= E_META_DROPSPENSER_ACTIVATED;
+		}
```
madmaxoft: Do we really want to twiddle the bit off and on? I think the clearing should be in an `else` branch of the `if` statement instead; and you could actually `return` the value directly.

src/Simulator/IncrementalRedstoneSimulator/DropSpenserHandler.h
```diff
@@ -39,8 +53,9 @@ class cDropSpenserHandler : public cRedstoneHandler
 	virtual cVector3iArray Update(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, PoweringData a_PoweringData) override
 	{
 		// LOGD("Evaluating spencer the dropspenser (%d %d %d)", a_Position.x, a_Position.y, a_Position.z);
-
-		if (a_PoweringData.PowerLevel > 0)
+		bool IsPoweredNow = a_PoweringData.PowerLevel > 0;
```
madmaxoft: Use parentheses around the comparison please.

madmaxoft:
Sorry for the late comments, I should have reviewed this earlier.

Please fix those, and ideally rebase onto current master:
```
git remote add upstream https://github.com/cuberite/cuberite.git
git fetch upstream
git rebase upstream/master
```
You will then need to force-push your changes to your repo, either with `git push -f` or an equivalent GUI.

QUSpilPrgm:
Thanks for the review. I have applied all suggestions.

madmaxoft:
Looks good to me; I'd appreciate someone else reviewing this as well.

PR #3226, opened 2016-06-05
Redstone fixes

### LINKED ISSUE #3168: Crashing cuberite with a sticky piston, redstone, and a redstone block
### Expected behavior
Piston should oscillate at 1 tick, just like on bukkit, spigot, etc

### Actual behavior
Piston oscillates at intervals below 1 tick, resulting in the server crashing.
https://www.youtube.com/watch?v=h65EHBxXC2k

Shown in the youtube video.

### Build + Commit ID
Cuberite build id: 295
commit id: 16d53a039b2bd590989477c0f8b25acecd988417

### Similar issues
Similar to issue #3080

tigerw:
* Various improvements
* Fixes #3168

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,11 @@
 version: 1.0.{build}
 configuration: Debug
-clone_depth: 50
+clone_depth: 1
 before_build:
 - echo %TIME%
 - git submodule update --init
 - echo %TIME%
-- cmake -G "Visual Studio 12" .
+- cmake .
 - echo %TIME%
 build:
   project: Cuberite.sln
--- a/src/Blocks/BlockComparator.h
+++ b/src/Blocks/BlockComparator.h
@@ -60,6 +60,11 @@ public:
 		return true;
 	}
 
+	inline static bool IsInSubtractionMode(NIBBLETYPE a_Meta)
+	{
+		return ((a_Meta & 0x4) == 0x4);
+	}
+
 	inline static bool IsOn(NIBBLETYPE a_Meta)
 	{
 		return ((a_Meta & 0x8) == 0x8);
--- a/src/Blocks/BlockComparator.h
+++ b/src/Blocks/BlockComparator.h
@@ -60,6 +60,11 @@ public:
 		return true;
 	}
 
+	inline static bool IsOn(NIBBLETYPE a_Meta)
+	{
+		return ((a_Meta & 0x8) == 0x8);
+	}
+
 	inline static Vector3i GetSideCoordinate(const Vector3i & a_Position, NIBBLETYPE a_Meta, bool a_bInverse)
 	{
 		auto Position = a_Position;
--- a/src/Blocks/BlockPiston.h
+++ b/src/Blocks/BlockPiston.h
@@ -93,15 +93,15 @@ public:
 		return 11;
 	}
 
+	/** Returns true if the piston (with the specified meta) is extended */
+	static inline bool IsExtended(NIBBLETYPE a_PistonMeta) { return ((a_PistonMeta & 0x8) != 0x0); }
+
 private:
 
 	typedef std::unordered_set<Vector3i, VectorHasher<int>> Vector3iSet;
 
 	/** Returns true if the piston (specified by blocktype) is a sticky piston */
 	static inline bool IsSticky(BLOCKTYPE a_BlockType) { return (a_BlockType == E_BLOCK_STICKY_PISTON); }
-
-	/** Returns true if the piston (with the specified meta) is extended */
-	static inline bool IsExtended(NIBBLETYPE a_PistonMeta) { return ((a_PistonMeta & 0x8) != 0x0); }
 
 	/** Returns true if the specified block can be pushed by a piston (and left intact) */
 	static inline bool CanPush(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta)
--- a/src/Simulator/IncrementalRedstoneSimulator/PistonHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/PistonHandler.h
@@ -39,14 +39,37 @@ public:
 	virtual cVector3iArray Update(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, PoweringData a_PoweringData) override
 	{
 		// LOGD("Evaluating pisty the piston (%d %d %d)", a_Position.x, a_Position.y, a_Position.z);
+		auto Data = static_cast<cIncrementalRedstoneSimulator *>(m_World.GetRedstoneSimulator())->GetChunkData();
+		auto DelayInfo = Data->GetMechanismDelayInfo(a_Position);
 
-		if (a_PoweringData.PowerLevel > 0)
+		// Delay is used here to prevent an infinite loop (#3168)
+		if (DelayInfo == nullptr)
 		{
-			cBlockPistonHandler::ExtendPiston(a_Position, &m_World);
+			bool ShouldBeExtended = (a_PoweringData.PowerLevel != 0);
+			if (ShouldBeExtended != cBlockPistonHandler::IsExtended(a_Meta))
+			{
+				Data->m_MechanismDelays[a_Position] = std::make_pair(1, ShouldBeExtended);
+			}
 		}
 		else
 		{
-			cBlockPistonHandler::RetractPiston(a_Position, &m_World);
+			int DelayTicks;
+			bool ShouldBeExtended;
+			std::tie(DelayTicks, ShouldBeExtended) = *DelayInfo;
+
+			if (DelayTicks == 0)
+			{
+				if (ShouldBeExtended)
+				{
+					cBlockPistonHandler::ExtendPiston(a_Position, &m_World);
+				}
+				else
+				{
+					cBlockPistonHandler::RetractPiston(a_Position, &m_World);
+				}
+
+				Data->m_MechanismDelays.erase(a_Position);
+			}
 		}
 
 		return {};
--- a/src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
@@ -99,22 +99,42 @@ public:
 
 	virtual cVector3iArray Update(const Vector3i & a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta, PoweringData a_PoweringData) override
 	{
+		// Note that a_PoweringData here contains the maximum *side* power level, as specified by GetValidSourcePositions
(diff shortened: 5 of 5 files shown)


tigerw:
@madmaxoft, @NiLSPACE and other Windows guys, may I ask, what compiler versions are you using?

madmaxoft:
MSVS 2013 (the one you removed in this PR)

madmaxoft:
And the code is still wrong, the `static` variables cannot have `cWorld` parameters in their constructors.

NiLSPACE:
I'm using VS 2015

tigerw:
Surely thanks to PerWorldThreads, thread_local means that the handlers will be constructed with the correct world when that world's redstone simulator calls the factory?

------

@madmaxoft I don't suppose that you will be willing to use a newer version because of some redstone code improvements. (will you?) so hopefully 2013 can use `__declspec(thread)`.

QUSpilPrgm:
Why not create the object in a buffer on the stack?
I think that would be faster and less ugly than globals or thread local memory.

You can even make it 100% safe, using a small helper class like this one:
```
#include <cstddef>
#include <new>
#include <utility>
#include <type_traits>

template<typename BaseT, std::size_t BufferSize>
class cAnyOnStack
{
private:
#if defined(_MSC_VER) && (_MSC_VER < 1900)  // Compatiblity with Visual Studio 2013
    __declspec(align(8)) char Buffer[BufferSize];
#else
    alignas(alignof(std::max_align_t)) char Buffer[BufferSize];
#endif
    bool Used = false;
public:
    template<typename T, typename...  TArgs>
    T* Create(TArgs&&... a_Args)
    {
        static_assert(BufferSize >= sizeof(T), "Not enough space for T");
        return new(Buffer) T(std::forward<TArgs>(a_Args)...);
    }
    cAnyOnStack(const  cAnyOnStack&) = delete;
    cAnyOnStack& operator =(const cAnyOnStack&) = delete;
    ~cAnyOnStack()
    {
        static_assert(!std::has_virtual_destructor<BaseT>::value, "BaseT is required to have a virtual destructor");
        if (Used)
            reinterpret_cast<BaseT*>(Buffer)->~BaseT();
    }
};
```
Usage example:
```
typedef cAnyOnStack<cRedstoneHandler, 32> AnyRedstoneHandlerBuffer;
cRedstoneHandler * cIncrementalRedstoneSimulator::CreateComponent(AnyRedstoneHandlerBuffer& a_Any, cWorld & a_World, BLOCKTYPE a_BlockType, cIncrementalRedstoneSimulatorChunkData * a_Data)
{
   switch
   [...]
       return a_Any.Create<cPoweredRailHandler>(a_World);
   [...]
}

```

tigerw:
I wouldn't agree that it's less ugly :P

Research on the InternetⓇ (Your Infallible Source for Accurate Information™) suggests that TLS is not much slower at all than plain variables with a modern compiler, which would suggest that the added complexity is unnecessary.

QUSpilPrgm:
My worries about thread local memory are not only about speed, but also functionality and code quality.

Redstone in the nether is currently still broken. You could reassign the World at every invokation to avoid that, but I don't think that would be a good idea. Modifiable global state is _generally_ problematic because it adds an extra state dependency that is not documented in the argument list of a function. For example, in this case it would be a possibly very hard to detect error if the thread invoked the CreateComponent function a second time, while later again using an old cRedstoneHandler* pointer. In very rare circumstances this could refer to the wrong world. I think this kind of errors are hard to catch.

If you dislike the stack approach, there are possibly other ways.
How about just making the globals "const" and handing over the World as an extra argument?

tigerw:
I was hoping that with TLS and per world threads, it would be guaranteed that CreateComponent returned handlers for the current (correct) world.

I considered passing cWorld in, but decided that it would be so much tedious search and replace, and change so many files, that it wouldn't be suitable as a quick fix I could slip in to this PR.

Also, you know cAnyOnStack - I always thought that `new` created things on the heap; am I mistaken in that thought?

QUSpilPrgm:
This kind of new is called [placement new](https://en.wikipedia.org/wiki/Placement_syntax). (notice the brackets before the type)
It basicly just invokes the constructor on an arbitary buffer. In this case the buffer is on the stack so there is no memory allocation on the heap. You can compare the generated assembly on Godbolt.org: [Thread_Local](http://gcc.godbolt.org/#compilers:!((compiler:clang380,options:'-O3+-std%3Dc%2B%2B11+-DNDEBUG',sourcez:MYGwhgzhAEwOoHsBOIAmBvAvgbgFCkhmACUBTVCAFwQDtSAJMG1EUpXdXABwFcAjEAEtgALlzQJsRClQAqaNLR5JsMhWp1GzVkgAU8ZGnlgA%2BotQBKcSugiFh1LtPmrKrNYkA3QUko8wINAAfiTkVLQMTCxsuhZ4OLj4auGaUTryAMJIpGCUkdoxBjLGZg5WnCpUucLQlAAW2WCoJiAIwAGqYRr50UjQWr1OpTJxHv1pbAB05tAAvNDODsqS2X5INNAAZAM68UAAAA%3D%3D)),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3) and [cAnyOnStack](http://gcc.godbolt.org/#compilers:!((compiler:clang380,options:%27-O3+-std%3Dc%2B%2B11+-DNDEBUG%27,sourcez:PTCCDsE8HlwZQC4EMDGBrAdACwFAGIBLcFAGwFcATAUwAIAeFJAZyaoCcEA%2BfI0y2hkwQVqAM26Fi5avXBUA7hN7SBZBARIEEkJVP71tAByoB9BGyRam3HAioBbQySR26RquCT3aAIWZUAFQAaGiEKAC5wpgIAL1MEGh8yUVF2OFiqblJmJhoUCBh4ZHQcAG8cQzYCADcXKnCeURoxIioKAAoTAFk4AGETADUAUQAlAEoaADJJmk6e/uGR%2BhoARgBOAAYNiZoQGl6Ae0cXAgAjTW0aeS0sGgGCJjIkEhpESgIDmgAmDZWAZhwNCBNBMJmopCYxhQ7WeBAA5uB2gAOMYTFBYJBsRLJVJsADaSRSaQyAF0ANz4KgkViA4GwhHMGGaBEHUTtMKRexIAAeJnp4DMqLyGKxhNxBJxxLi5Mp4AoBFEtKBpwOBxeAFVWBQaABeGiiZ6sCmGMjnAgoBrAmh2RzOVzuTzeGjBa2QYyOqgYL1AgKgNhw6xK50AKn2bCodXavv9TGmXowNCQJj9AbGQfKVqtQhOKD5LHYCHaYqltE4euicVZUbGIQARAA5A4JDwHMhw26Q1C0UQHLEBWtjCmZun5jjtACEmrag6DVvDCDIbHANDk8iLkrYEwC7OEkR7bHkmIodGjAc4MOTMbG8ZnVoAvkH8lBYIhUGh2igDuAhECn4VX%2BgkwTHq1AkFQdhDsCf4vsUaAzAcxgWAgva6h%2BX4/tBRRvkBurNFS4FUJBQIAH6YQB75plaGbDqEyDqLmOQFhOHLhBiTAmNUBAcE8JBglQQhsGQKDIWwdB%2BKwAScJEtTkFQdbiYENAPDQ4YAI5kFxbTWp8GLVLQSA0Jx3HPHhAlCSJA5EZmCqzFOFCUTRc5UEQdhsJU4EmIwQhif4ATBuexabgAtJwxEKdut7Ag%2Bd4UjgcUgCMbRCF%2BVAABJIHKYFsNgODZCweQAOq9iQFClDFuXOPlKCJRQyVyOlmXsGUFSmpoFqPkVbAlaGnUlVZ1VJch9UZRQWUfr1FChkmE0OVa4Q0BNF4zemD5WkZC4maRNV1WlI1jTO5XuGIeQFDBb4MNtQ27Y1bAhH8XycDQBSXSlDWjewgUUgNtVXW9WWhr04Z1H97DtM9g2vXtH0bomJiBSEKATVNJjLdRQLzouy5JoFGCAxGrjfTtINsOe03FfZFJ3kAAA)),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3)

If you want to quick fix it for the moment, I would suggest going back to dynamic memory alloction like it was not too long ago. That is not the fastest solution but it is 100% safe and correct and there are currently plenty dynamic memory alloctions in that code anyway.

tigerw:
Well. Now I know, thanks! Fancy C++ is fancy.

------

I'm just curious as to why TLS can't be 100% correct.

QUSpilPrgm:
It can be correct but would always tend to be fragile as I already described earlier.

The most obvious problem at the moment is that thread_local variables are only constructed once. If you call `CreateComponent` again with a different world, the pointer to the old world won't be updated. 

As I just noticed it isnt even as easy to fix as just reassigning the world pointer, because it is actually a protected reference inside `cRedstoneHandler`.

tigerw:
But what's the point of thread_local static then, if constructing from different threads doesn't give you a per-thread static? Are you sure that's the case? Worlds are ticking in their own threads.

QUSpilPrgm:
Oh I didn't know that each World has its own thread. In that case it should work at the moment.

I still think it is very fragile long time solution. This interaction with threading is far from being obvious or easy to reason about.

worktycho:
The redstone simulator as a whole is very fragile. The basic issue is representing the FSM of each redstone component in a way that is as simple and as efficient as possible. The handlers approach is a significant improvement on the original single function approach, but there are definitely better ways to simulate out there.

tigerw:
Thank you for your vote of confidence. :P

QUSpilPrgm:
Wouldn't it be a good idea to try to move in a direction of less fragile code instead of the opposite?
This seemed to be a regression to me.

If there is a consensus to keep it `thread_local` for now, is there support for the idea of switching to a more clean and stable method in a future patch? (e.g. passing the world as a function argument instead of modifying globals)

worktycho:
@tigerw I blame mojang, not you. The redstone rules are a mess.

worktycho:
@QUSpilPrgm I support the idea of world as a function argument, but we do need a per world handler cache. @tigerw Have you considered putting handlers in the simulator data?

tigerw:
Maybe, but this works for now.

@madmaxoft, you still on 2013?

madmaxoft:
Yes, I'm still on VS2013, but even if I weren't, I'm strongly opposing using anything thread-local if we don't have to. What if we find a way, in the future, to have multiple world tick threads? Do you have the knowledge of the thread local implementation on all relevant platforms to be sure if it's faster than any other solution?

As @worktycho said, put the handlers in the world's simulator data.

tigerw:
They once said, progress stops for no man. No more, I guess.

I am of the opinion that handler objects don't belong there, and it just adds unnecessary confusion. Alternatively, I suppose the classes can be made entirely static, but in the meantime, no enhancements to performance. :(

Thanks Obama.

madmaxoft:
Reverting, because it doesn't compile in VS2013.

tigerw:
Ah, I knew I shouldn't have left those AppVeyor changes in, apologies.

PR #3227, opened 2016-06-09
Normalized Vector3 API to use the same capitalization as everything else

madmaxoft:
The `Abs` and `Clamp` functions are now capitalized just like all the other functions in the API; old variants are kept in DeprecatedBindings.cpp.


src/Bindings/DeprecatedBindings.cpp
```diff
@@ -340,6 +340,49 @@ static int tolua_cWorld_SetSignLines(lua_State * tolua_S)
...
+template <typename T>
+int tolua_Vector3_Clamp(lua_State * a_LuaState)
+{
+	// Retrieve the params, including self:
+	cLuaState L(a_LuaState);
+	Vector3<T> * self;
+	T min, max;
+	if (!L.GetStackValues(1, self, min, max))
+	{
+		tolua_error(a_LuaState, "invalid 'self' in function 'Vector3<T>:Clamp'", nullptr);
```
madmaxoft: Copy-pasta error in the error message, any param could be wrong, not just `self`

madmaxoft:
@NiLSPACE this is mostly your code being changed, are you against this change?

NiLSPACE:
Yes, absolutely

madmaxoft:
Wait, what? Yes, absolutely against this change, and a thumbs-up?

NiLSPACE:
I must've been more tired than I thought yesterday :|

I got no problem with merging this.

PR #3228, opened 2016-06-10
Separate self tests

madmaxoft:
Moved all self-tests into separate projects, rather than globally-created classes.


madmaxoft:
The Travis build seems to have stalled, I restarted it.

madmaxoft:
How am I supposed to fix a test that is failing before it even enters its `main` function? What's going on here? And it's random, too. Sometimes only the debug clang build fails, sometimes more of them. And the tests do work on my testing Linux environment :(

tigerw:
Is cNetworkSingleton::Initialise blocking something?

madmaxoft:
Nothing that I know of; and anyway there's a message supposed to be printed to screen just at the start of the `main` function, and it doesn't seem to get even there.

Seadragon91:
Running EnumInterfaces-exe in a loop freezes after a few runs.
```
counter=1
while [ true ]
do
        #./CIbuild.sh
        echo $counter
        ./tests/Network/EnumInterfaces-exe
        let "counter++"
done
```

Create the file, `chmod +x run.sh` and then start it.

Seadragon91:
Have added LOG messages and it freezes at this [line](https://github.com/cuberite/cuberite/blob/SeparateSelfTests/src/OSSupport/NetworkSingleton.cpp#L108)
`m_EventLoopThread.join();`

madmaxoft:
@Seadragon91 That might be something else, because that's the termination, not the initialization. The Travis builds seem to hang at the startup of the test executable: https://travis-ci.org/cuberite/cuberite/jobs/138118603 (note that the first thing the test does is print a log message, the message is not in the output.

madmaxoft:
Ah, the logging could be at co-fault here, it might not flush.

madmaxoft:
Right, so it hangs at the end, the output was missing because the logging functions weren't flushing the IO buffers.

madmaxoft:
Wow, it actually is a problem in the underlying `cNetwork` implementation that this has found. Quite a corner-case: if the `cNetwork` singleton is started and then terminated too quickly, it actually can hang. Fixing...

madmaxoft:
This should fix the `cNetwork` issue, but I'm still getting a dead-hang when trying to repeatedly run the EnumNetworkInterfaces-exe test. Before the fix it took like 500 repetitions, now it takes like 20.000 repetitions, and the hang is in the internals of the `cEvent` implementation. I guess my next move will be to stress-test that.

tigerw:
Before it took 500, now only 20 to 3 decimal places; that's good right?

madmaxoft:
That's actually a thousands separator, it's 20k. And the cause of that is a very different problem, it seems, so I decided to merge these despite the problem still being there.

It seems that our `cEvent` implementation is not robust enough, I'm stress-testing it currently and it deadlocks each time.

tigerw:
A thousands separator! Perhaps your locale settings may be the cause of cEvent instability? Perhaps it is interpreting numbers incorrectly.

...

:D

tigerw:
Is there no prettier networking library? Having to write wrappers around things is wasted effort, I think. And what was it you did to fix the deadlocking with the ev_timer thing? Surely libevent can handle rapid termination without resorting to unclear workarounds.

Seadragon91:
@tigerw I found a discussion about a method named`bufferevent_free` from libevent, it can result into a deadlock, if called from a other thread.

madmaxoft:
The problem was that the termination was called before the worker thread for LibEvent actually had a chance to start. The `Initialise` method created the worker thread but the scheduler decided to continue the Initialise thread, rather than the new thread, so all the test code got executed until `Terminate`, and only then the `RunEventLoop` function was executed in the worker thread, but it doesn't check the LibEvent termination flag before entering the loop, so the request for termination was essentially lost.

What I did is that the `Initialise` thread waits until the `RunEventLoop` starts up the LibEvent loop; and the only way to do that is to actually schedule a task onto that loop that notifies somehow that the loop is running; I use a timer with a zero timeout for that. So now when the `Initialise` method exits, it is clear that the worker thread is already running and processing stuff.

As for another networking library, there aren't actually that many. When we decided to drop our own network code in favor of a library, we did some research and only two libraries actually supported all the stuff we needed - LibEvent and Asio. Unfortunately, Asio is too much intertwined into Boost, which would complicate compiling, so LibEvent was chosen instead. It's not that bad, and we'd need some kind of a wrapper for any library anyway.

madmaxoft:
Well, I really have no idea what the problem is now. The EnumInterfaces test is still hanging for me from time to time (about once in 40000 attempts) with these two stacktraces:

```
(gdb) info threads
  Id   Target Id         Frame
  2    Thread 0x7f6c9c869700 (LWP 23525) "EnumInterfaces-" __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
* 1    Thread 0x7f6c9d8f6740 (LWP 23524) "EnumInterfaces-" 0x00007f6c9cc3c8ed in pthread_join (threadid=140104459261696, thread_return=0x0) at pthread_join.c:90

(gdb) bt
#0  0x00007f6c9cc3c8ed in pthread_join (threadid=140104459261696, thread_return=0x0) at pthread_join.c:90
#1  0x00007f6c9d429f47 in std::thread::join() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x000000000041cd65 in cNetworkSingleton::Terminate (this=0x664ea0 <cNetworkSingleton::Get()::Instance>) at /home/xoft/cuberite/src/OSSupport/NetworkSingleton.cpp:110
#3  0x000000000041b08b in main (argc=1, argv=0x7ffe538a35a8) at /home/xoft/cuberite/tests/Network/EnumInterfaces.cpp:31

(gdb) thread 2
[Switching to thread 2 (Thread 0x7f6c9c869700 (LWP 23525))]
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
135     ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S: No such file or directory.

(gdb) bt
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007f6c9cc414bd in pthread_cond_signal@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_signal.S:144
#2  0x00007f6c9d424cf9 in std::condition_variable::notify_one() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x000000000042650f in cEvent::Set (this=0x17cd500) at /home/xoft/cuberite/src/OSSupport/Event.cpp:59
#4  0x000000000041ced8 in cNetworkSingleton::SignalizeStartup (a_Socket=-1, a_Events=1, a_Self=0x664ea0 <cNetworkSingleton::Get()::Instance>) at /home/xoft/cuberite/src/OSSupport/NetworkSingleton.cpp:173
#5  0x0000000000434547 in event_process_active_single_queue (base=0x17cccf0, activeq=0x17cd140, max_to_process=2147483647, endtime=0x0) at /home/xoft/cuberite/lib/libevent/event.c:1642
#6  0x000000000042cacd in event_process_active (base=0x17cccf0) at /home/xoft/cuberite/lib/libevent/event.c:1734
#7  0x000000000042b84c in event_base_loop (base=0x17cccf0, flags=4) at /home/xoft/cuberite/lib/libevent/event.c:1957
#8  0x000000000041ccdf in cNetworkSingleton::RunEventLoop (a_Self=0x664ea0 <cNetworkSingleton::Get()::Instance>) at /home/xoft/cuberite/src/OSSupport/NetworkSingleton.cpp:161
#9  0x0000000000424f12 in std::_Bind_simple<void (*(cNetworkSingleton*))(cNetworkSingleton*)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x17cccd8)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.2.1/../../../../include/c++/5.2.1/functional:1530
#10 0x0000000000424ec5 in std::_Bind_simple<void (*(cNetworkSingleton*))(cNetworkSingleton*)>::operator()() (this=0x17cccd8)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.2.1/../../../../include/c++/5.2.1/functional:1520
#11 0x0000000000424e9c in std::thread::_Impl<std::_Bind_simple<void (*(cNetworkSingleton*))(cNetworkSingleton*)> >::_M_run() (this=0x17cccc0)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/5.2.1/../../../../include/c++/5.2.1/thread:115
#12 0x00007f6c9d42a030 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#13 0x00007f6c9cc3b6aa in start_thread (arg=0x7f6c9c869700) at pthread_create.c:333
#14 0x00007f6c9c970e9d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
(gdb) thread 1
[Switching to thread 1 (Thread 0x7f6c9d8f6740 (LWP 23524))]
#0  0x00007f6c9cc3c8ed in pthread_join (threadid=140104459261696, thread_return=0x0) at pthread_join.c:90
90      pthread_join.c: No such file or directory.
```

So thread 1 (main thread) has finished everything and is waiting for thread 2 (LibEvent loop thread) to terminate, but thread 2 is still stuck in setting the startup `cEvent`. What's going on here? Isn't `std::condition_variable::notify_one()` supposed to atomically unlock the CV and return?

PR #3230, opened 2016-06-11
Converted to android-cmake for Android building

tigerw:
Continuation of #2526 after it was, ahem, prematurely closed.


Schwertspize:
Some issues with Travis, looks like it's failing everywhere... 

I'm not sure if this is the right approach, in my opinion android code (=java) should be located in the Android project and this is just for cross-compiling the binaries. 

Thank you very much for reopening this!

tigerw:
Yes I agree, which is why I created the android repository, but I just want to get this merged as quickly as possible so that it isn't left stale, again, and you can massage it however you need afterwards.

tigerw:
And yes indeed, Travis is failing all over the place.

worktycho:
This still has a bunch of the same issues as reported in #2526. Any reason we can't just reopen #2526?

CMakeLists.txt
```diff
@@ -105,18 +105,20 @@ if(${BUILD_UNSTABLE_TOOLS})
 	add_subdirectory(Tools/GeneratorPerformanceTest/)
 endif()
 
-include(CheckCXXCompilerFlag)
+# The Android toolchain sets its own flags; interference is bad
 include(SetFlags.cmake)
-set_flags()
-set_lib_flags()
-enable_profile()
+if (NOT ANDROID)
```
worktycho: I'm still not happy with this. There is no way the android toolchain can inject different flags for libraries, and executables, which means this will break things.

src/CMakeLists.txt
```diff
@@ -332,51 +329,60 @@ if (MSVC)
...
-# Output the executable into the $/Server folder, so that it has access to external resources:
-set(EXECUTABLE_OUTPUT_PATH ${CMAKE_SOURCE_DIR}/Server)
-SET_TARGET_PROPERTIES(${EXECUTABLE} PROPERTIES
-	RUNTIME_OUTPUT_DIRECTORY_DEBUG          ${CMAKE_SOURCE_DIR}/Server
-	RUNTIME_OUTPUT_DIRECTORY_RELEASE        ${CMAKE_SOURCE_DIR}/Server
-	RUNTIME_OUTPUT_DIRECTORY_DEBUGPROFILE   ${CMAKE_SOURCE_DIR}/Server
-	RUNTIME_OUTPUT_DIRECTORY_RELEASEPROFILE ${CMAKE_SOURCE_DIR}/Server
-)
+if (NOT ANDROID)
+	add_executable(${CMAKE_PROJECT_NAME} ${SOURCE})
```
worktycho: Why abolish the executable variable?

Schwertspize: Naturally I have been in contact with tiger

He said he wanted to make sure it compiles. The current situation in cuberite/master is an Android App whicu provides an interface to a native library which provides an interface to the Android source. In my opinion this is just too complicated and it's much easier to just generate executables, but tiger stated on hangouts that this may take time and another pr. 'it doesn't fit the "make sure it builds" approach' 

As I am understanding this diff it's just disabling the executable for android which is at the moment the best we can to keep any usable wax for android and a base to build our new code on. I think it will take some time until android is finished in a way that we can use it like any other platform (running the actual binaries and such us a different problem, then, but one that is not hard to solve)

tigerw: What @Schwertspize said.

Also, for the variable itself, I thought, since project_name = executable, we might as well make it more succinct. One less thing to replace I guess if someone proposes another name change.

CMakeLists.txt
tigerw: Alright, I'll try to unify it.

src/CMakeLists.txt
```diff
@@ -332,51 +329,60 @@ if (MSVC)
...
+	SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME} PROPERTIES DEBUG_POSTFIX "_debug")
 
-# Make the profiled executables have a "_profile" postfix
-SET_TARGET_PROPERTIES(${EXECUTABLE} PROPERTIES DEBUGPROFILE_POSTFIX "_debug_profile")
-SET_TARGET_PROPERTIES(${EXECUTABLE} PROPERTIES RELEASEPROFILE_POSTFIX "_profile")
+	# Make the profiled executables have a "_profile" postfix
+	SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME} PROPERTIES DEBUGPROFILE_POSTFIX "_debug_profile")
+	SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME} PROPERTIES RELEASEPROFILE_POSTFIX "_profile")
+else()
+	add_library(${CMAKE_PROJECT_NAME} SHARED ${SOURCE})
```
worktycho: Why are you linking main.cpp into the lib? I feel that it might be better to restructure the main project to use a library, then an executable which is the lib + main.cpp.

tigerw:
> This still has a bunch of the same issues as reported in #2526. Any reason we can't just reopen #2526?

I force pushed the branch and now GitHub says nope.

worktycho:
Alright, but I'm going to go for no.

Schwertspize:
@worktycho why were you closing it? I don't think this is considered as finished yet

worktycho:
I was trying to reopen #2526.

tigerw:
Well, tycho, you surely have nothing to complain about now? Sooo... merge?

Cl1608Ho:
Uhm are the build issues fixed already? I don't think so (CI only builds for windows/linux) :o

madmaxoft:
I don't like the changes to the non-Android-related stuff, such as `cLuaState`. Pull those out to a separate PR with a proper description.

tigerw:
Android clang is super strict and won't build without fixing warnings tho'.

madmaxoft:
Then fix those in a separate PR and build this PR on top of it

Cl1608Ho:
Uhm this won't build for mips and mips64 due to this error 
```
/home/cuberite/cuberite/src/ByteBuffer.cpp:43:3: error: #error Cannot determine endianness of this platform
 #error Cannot determine endianness of this platform
  ^
```

Additionally, DNS isn't working on Android and so player authentication, you need to disable it manually.

Cl1608Ho:
Compiling tolua++ fails. Using Ubuntu 16.04.3 and cmake 3.7.0 (and standard build tools)

madmaxoft:
We did say that `tolua` wouldn't be needed for cross-compiling, you can use the native lua to process the bindings. Only the `tolualib` is then needed as a static library.

Cl1608Ho:
Oh sorry, something strange happened. Building it again solved it

madmaxoft:
@Cl1608Ho that might still be a problem. Try `make clean && make`, if it fails, it means that dependencies are wrong.

Cl1608Ho:
`make clean` is actually not enough, because it only reverts what make did, but not what cmake did. I don't know if tigerw already included my build script, it deletes everything from cmake cache to makefiles. The "all" target also creates all necessary zip files that the build server need to provide. 

I couldn't test yet if dns is working, because the app is momentarily in an not usable state, but I think I will provide that information soon

Cl1608Ho:
Seems like dns is still not working.... @tigerw ?

madmaxoft:
I also expect some form of documentation on how to actually build for Android, as part of this PR.

Cl1608Ho:
Well, there is a windows build command in the cmake file, but I hope tigerw will include my compile script as well

madmaxoft:
It's not the command itself, it's the list of tools that need installing, libraries that need downloading, etc.

Cl1608Ho:
Well there are comments in my shell script. 

Anyway, I send it here too. Cmake 3.7+. Everything else that cuberite needs.

Cl1608Ho:
Bump. Tigerw and I have discussed that although this seems to contains a DNS error (or one of the underlying socket) we want to get this merged and fix that issue later (because it's not that easily fixable and it doesn't completely disable the server, you need to try reconnecting.) as "hot fix" I will add a message that it's broken and a button to disable authentification in the Android app. 

Apart from that, this builds fine and runs fine (with quite some performance on my phone, a oneplus one), it's time to have this officially supported (with a few bugs as always) and get it onto the build servers.

Squash merge!

Note; please wait until tigerw pushes an updated version of the compile script. Thanks

Cl1608Ho:
Looks good. Can we now merge?

sphinxc0re:
No. First you have to rebase the branch and squash the commits. And I want to take a closer look at the changes first

android/.gitignore
```diff
@@ -0,0 +1,5 @@
+/Cuberite/
+/CMakeFiles/
+/lua_native-prefix/
+/tolua_native-prefix/
+CMakeCache.txt
```
sphinxc0re: missing newline

sphinxc0re:
Shouldn't the submodule changes go into a separate PR?

Cl1608Ho:
I dont know..... I guess they are required for Android to compile successfully.

However, I would like to see some changes to COMPILING.md under the headline Android :)

SetFlags.cmake
```diff
@@ -86,7 +86,13 @@ macro(set_flags)
 		else()
 			add_flags_cxx("-pthread")
 		endif()
-
+	elseif (ANDROID)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
madmaxoft: Why `c++14`? We only use c++11 features in the project.

madmaxoft:
I'm not too happy about the two-step build for tolua. We did change the build process so that it can use system-provided Lua for the bindings generation, and we did say this could be a precondition for cross-compiling.

android/.gitignore
tigerw: Just for the record, Linux requirements for newlines at the end of a file is very disagreeable to me.

tigerw:
> I'm not too happy about the two-step build for tolua. We did change the build process so that it can use system-provided Lua for the bindings generation, and we did say this could be a precondition for cross-compiling.

Does the new system Lua build completely replace tolua? You said tolualib was still needed above, so the 'two-step' build is still necessary unless you can load an Android library on Windows. If this is the case, I see no particular reason for mandating an extra installation of a Lua (5.1? Will 5.3+ work?) interpreter.

madmaxoft:
`tolualib` is only needed for Cuberite itself, it's a static library built for the target platform, it takes no part in the build process. Lua can completely generate the bindings (and cmake will already use system-wide Lua executable as long as it's in PATH).

Lua 5.1 is tested, but I see no reason why the build shouldn't work with any later Lua version.

I think having Lua installed system-wide is a rather reasonable precondition for cross-compiling, when it makes the build process much easier and less error-prone.

tigerw:
Alright, so what you're saying is that BindingsProcessor can generate everything without tolua.exe?

madmaxoft:
CMake, when it detects Lua in PATH, will arrange for all the bindings generation. It will use the `BindingsProcessor.lua`, but you don't actually have to do anything, it already happens. Note that it might not happen if you already ran CMake without having Lua, you may need to clean up CMakeCache and re-run CMake.

tigerw:
And does it like Lua 5.3.3? Or are modifications required?

madmaxoft:
It should work, but I haven't tested anything else than 5.1. Give it a try.

android/.gitignore
yangm97: I guess an [editorconfig](http://editorconfig.org) would take care of newlines and the such.

tigerw:
It seems to work after a great deal of changes.

For all my disparaging of backwards compatibility, I am surprised about the number of breaking changes between 5.0 and 5.2/5.3, but also how tolua++ managed to make use of so many of them in so many places :P

Cl1608Ho:
I have improved documentation about compiling it for android a lot,  let's wait until tigerw includes this commit. 

Apart from that, the dns issue appears to be a device specific one (specific to my device) as it didn't happen on the test device of a friend of mine.

@sphinxc0re would you approve?

tigerw:
boop

boop

Cl1608Ho:
For me everything is OK. Merge ASAP :P

bearbin:
Merged on @tigerw's instruction.

madmaxoft:
You've broken the Windows builds, because you've changed tolua in a way not compatible with Lua 5.1. I told you so, but still you didn't seem to care.
https://github.com/cuberite/toluapp/commit/c4cb9a5c6079ffa2e03b8e7533dc7175a19f25a3#commitcomment-20148897

sphinxc0re:
So, do we have to update Lua to 5.3 or revert this first?

madmaxoft:
It's not about our internal Lua, it's about the external interpreter used for generating bindings. Ideally, tolua should be compatible with 5.1 and all higher versions (it is possible to write the code that way).

sphinxc0re:
So it's about the machines building the binary? Why not just update Lua on these?

NiLSPACE:
I'd think it's a bad idea to use multiple versions of Lua throughout the project.

madmaxoft:
I run other Lua projects on my machine, so I won't be upgrading my Lua any time soon. I favor code that is universal.

sphinxc0re:
Isn't there a version manager for Lua? If you are running several projects, it might be a good idea to use one.

NiLSPACE:
I still think it's best to keep using one version of Lua throughout the project.

madmaxoft:
There, I fixed it for you, now would that have been that difficult?

https://github.com/cuberite/toluapp/commit/80c2548acafc8d26c2612a33305f5e2f07fe9d34

tigerw:
Apologies, I didn't see any of your comments because there were so many other notifications coming from GitHub. Thanks for fixing my mistake!

> now would that have been that difficult?

Probably; I don't really know anything about Lua bar the surface stuff. What exactly is load/loadstring, and why do you need to pass the enclosing function's name as a string to it?

android/compile.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
tigerw: > twisted sense of humor

:(

tigerw: like this?

madmaxoft:
They both convert a string into an executable code; `load` is from Lua 5.2+, `loadstring` is Lua 5.1. The other string that is passed to them is the "name" of the code blob, used for error-logging (`... in file "xyz" line N`). `load` additionally does the sandboxing, while `loadstring` doesn't; in Lua 5.2+ there is no way to sandbox later than when loading the code (which is what `setfenv()` from Lua 5.1 does).

PR #3232, opened 2016-06-12
Lua callback

madmaxoft:
`cLuaState::cCallback` is back, and this time care has been taken to make it thread-safe and everything.

This change defines and uses a new type to represent callbacks into Lua code (Lua code says "call this function when ..."). The callbacks are thread-safe (they lock the correct LuaState properly) and unload-safe (even if the plugin unloads, the callback object still exists and reports failure upon trying to invoke it).

I have rewritten most, if not all, of the old code using `cRef` and even `int` for callback storage, to use the new `cCallback` class. Therefore it has been tested extensively - it is now used, among others, for command handlers, hook registrations, webamin and scheduled world tasks.


src/Bindings/ManualBindings_World.cpp
```diff
@@ -466,67 +468,41 @@ static int tolua_cWorld_PrepareChunk(lua_State * tolua_S)
...
+	if (!Task->IsValid())
 	{
-		return cManualBindings::lua_do_error(tolua_S, "Error in function call '#funcname#': Could not get function reference of parameter #1");
+		return cManualBindings::lua_do_error(tolua_S, "Error in function call '#funcname#': Could not store the callback parameter");
 	}
 
-	auto ResettableTask = std::make_shared<cLuaWorldTask>(*Plugin, FnRef);
-	Plugin->AddResettable(ResettableTask);
-	self->QueueTask(std::bind(&cLuaWorldTask::Run, ResettableTask, std::placeholders::_1));
+	World->QueueTask([Task](cWorld & a_World)
```
tigerw: So I guess you and Tycho finally got your wishes... ;)

Server/Plugins/APIDump/Classes/Plugins.lua
```diff
@@ -22,10 +22,10 @@ return
 
 	cPluginLua =
 	{
-		Desc = "",
+		Desc = "(<b>OBSOLETE</b>) This class is no longer useful in the API and will be removed as soon as all core plugins are migrated away from it. The {{cPlugin}} class serves as the main plugin instance's interface.",
```
tigerw: And I guess I got mine too.

:)

-----

How did you solve the problem of the web tab implementation leaks, which was your concern when I proposed to remove cPluginLua?

tigerw:
+1,602   −2,290

No objections by default.

src/Bindings/ManualBindings_World.cpp
madmaxoft: Which wish exactly would that be?

Server/Plugins/APIDump/Classes/Plugins.lua
madmaxoft: I can't really remember the exact details of what you're referring to. I guess my concern was that once a plugin registered a web tab, the handler for the web tab would need to be valid for all webadmin calls, which happen in different threads. This would be solved by `cCallback` being properly thread-locked and staying valid even after its original `cLuaState` is gone.

src/Bindings/ManualBindings_World.cpp
tigerw: https://github.com/cuberite/cuberite/pull/2463#discussion_r39923458

madmaxoft:
Any chance of getting a review from @worktycho ?

worktycho:
I'm happy with the concept, but won't have the time/energy to do a full review until the weekend due to starting a summer job this week. I'll try then though.

src/Bindings/LuaState.h
```diff
@@ -112,6 +143,82 @@ class cLuaState
...
+
+		/** This class cannot be copied, because it is tracked in the LuaState by-ptr.
+		Use cCallbackPtr for a copyable object. */
+		cCallback(const cCallback &) = delete;
+
+		/** This class cannot be moved, because it is tracked in the LuaState by-ptr.
+		Use cCallbackPtr for a copyable object. */
+		cCallback(cCallback &&) = delete;
+	};
+	typedef SharedPtr<cCallback> cCallbackPtr;
```
worktycho: SharedPtr should not be the default. Ref-counting adds overhead, and makes it harder to change later.

src/Bindings/ManualBindings.cpp
```diff
@@ -1002,7 +1052,14 @@ static int tolua_cPluginManager_AddHook_FnRef(cPluginManager * a_PluginManager,
 	}
 
 	// Add the hook to the plugin
-	if (!Plugin->AddHookRef(HookType, a_ParamIdx + 1))
+	auto callback = std::make_shared<cLuaState::cCallback>();
```
worktycho: Why does this need to be a shared pointer? Its only owned by one thread then copied into the hookmap. We could just pass my reference into GetStackValue + AddHookCallback, then move into the hookmap.

worktycho: No, my misunderstanding.

worktycho: Still it seems excessive to use shared pointers here.

src/Bindings/PluginLua.h
```diff
@@ -231,45 +162,50 @@ class cPluginLua :
...
-	typedef std::map<AString, int> CommandMap;
-
 	/** Provides an array of Lua function references */
-	typedef std::vector<cLuaState::cRef *> cLuaRefs;
+	typedef std::vector<cLuaState::cCallbackPtr> cLuaCallbacks;
 
 	/** Maps hook types into arrays of Lua function references to call for each hook type */
-	typedef std::map<int, cLuaRefs> cHookMap;
-
+	typedef std::map<int, cLuaCallbacks> cHookMap;
```
worktycho: Does this need to be a shared pointer? If a cLuaCallback outlives the hookmap, something is probably wrong.

madmaxoft:
I have tried converting the `cCallbackPtr` from a `SharedPtr` to a `UniquePtr`, but ran into problems. The most cumbersome of those was that the `cCallbackPtr` was captured into a lambda (`cWorld:QueueTask` binding in `tolua_cWorld_QueueTask`), which is impossible with a `UniquePtr`.

Perhaps more fundamental reason is that I don't want to restrict the callbacks to be non-copyable, but on the other hand I don't want to have two separate ways of storing them. Anything that takes a `cCallbackPtr` should be able to take any callback, not only of the correct "copyability" type. Therefore I settled for the more universal of the types.

Refcounting is not that much of an overhead to be the killer here. And it only makes difference when shuffling the callback instance around (which is done once), not with the actual calls (which may be on the performance-heavy paths).

worktycho:
I think that fundamentally, having two types of pointers is not much of an overhead compared to the rest of c++, and if thats a problem we really need to think about the other issues c++ brings. This is also the reason I dislike the idea of a `cCallbackPtr` typedef in the first place.
Regarding refcounting, its less about the performance, but more that everytime I see a shared pointer, I have to assume I'm dealing with shared state, whereas with a unique pointer, I know I'm dealing with the master source of this pointer. This reduces the cognative load when dealing with the code.

madmaxoft:
I managed to convert the callbacks into `UniquePtr`s, with the only exception of the `QueueTask` binding, which uses a `SharedPtr` path of the code.

Ran into a really bad case of MSVC error reporting with this, an error was reported on a completely unrelated line (PluginLua.cpp reported template error on line 263 when a `std::move` was missing on line 1101, do look the code up, it's worth it).

I'm not too happy about there being two functions for reading the callback (`cLuaState::GetStackValue`), but I can live with it, it's not so bad as I thought it would be.

tigerw:
How do arguments to push_back affect a loop? Am confused :P
What would you say if I asked you for an upgrade to VS 2015 to test its behaviour?

Date: Thu, 30 Jun 2016 01:42:59 -0700

madmaxoft:
VS2015 is marginally better, it lists both lines under the problem:
```
(...)
1>  Cuberite\src\Bindings\PluginLua.cpp(1101): note: see reference to function template instantiation 'void std::vector<cLuaState::cCallbackPtr,std::allocator<_Ty>>::push_back(const std::unique_ptr<cLuaState::cCallback,std::default_delete<cLuaState::cCallback>> &)' being compiled
1>          with
1>          [
1>              _Ty=cLuaState::cCallbackPtr
1>          ]
1>  Cuberite\src\Bindings\PluginLua.cpp(263): note: see reference to class template instantiation 'std::vector<cLuaState::cCallbackPtr,std::allocator<_Ty>>' being compiled
1>          with
1>          [
1>              _Ty=cLuaState::cCallbackPtr
1>          ]
```

But it still does report line 263, which has no relation in the problem.

PR #3240, opened 2016-06-23
added Grown visualizer to the Generator documentation

NiLSPACE:


docs/Generator.html
```diff
@@ -429,9 +431,41 @@
...
 </p>
 
+<br />
+
+Here is a visualizer where you can play with the algorithm that the Grown biome generator uses:
+<table>
+	<tr>
+		<td rowspan="5">
+			<canvas id="stage" width="400" height="400"></canvas>
```
madmaxoft: Make it larger, at least 600 * 600. And limit the zoom so that one cannot zoom in when the array size is larger than the canvas size.

docs/js/ValueMap.js
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}
+
+		context.save();
+		context.globalCompositeOperation = 'difference';
+		context.fillStyle = 'white';
+		context.fillText("Size: " + (this.sizeX - 1) + "x" + (this.sizeZ - 1), 5, 10);
+		context.restore();
+	}
+}
```
madmaxoft: missing newline

docs/js/grown.js
```diff
@@ -0,0 +1,70 @@
...
+		g_ValueMap.visualize(g_Context, g_Canvas);
+
+		actions.splice(0, 1);
+		setTimeout(update, animationTimeBetween);
+	};
+	setTimeout(update, animationTimeBetween + 500);
+}
+
+
+window.onload = init;
```
madmaxoft: Missing newline

docs/Generator.html
NiLSPACE: How about I give a warning that rendering larger than 600x600 can lag the browser?

madmaxoft: That's pretty useless, reading the warning and clicking OK will require more effort than just waiting for the render. No, a simple disabling the button would be much better, in my opinion.

NiLSPACE:
I've added the missing newlines, made the canvas bigger and added a warning if the map gets too big. Also, I've added some some more information about the visualizer that rendering the map takes way longer than the actual zooming/smoothing.

NiLSPACE:
Done.

docs/Generator.html
```diff
@@ -429,9 +431,41 @@
...
 </p>
 
+<br />
+
+Here is a visualizer where you can play with the algorithm that the Grown biome generator uses. I'd like to point out that rendering the map takes way longer than the zoom/smooth operation.
```
madmaxoft: I don't think using "I'd like" is good in this text - no other part of the text uses that form. How about "Note that ..."?

docs/js/grown.js
```diff
@@ -0,0 +1,79 @@
...
+function btnReset() {
+	g_ValueMap.reset();
+	g_ValueMap.visualize(g_Context, g_Canvas);
+	
+	document.getElementById("grownZoomButton").disabled = false;
+}
+
+
+function btnAutomatic(target) {
+	target.disabled = true;
```
madmaxoft: Does this disable the Auto button? If so, I don't see any code to re-enable it.

NiLSPACE: It does at line 65 :)

docs/Generator.html
NiLSPACE: Yeah, that sounds much better.

docs/js/grown.js
madmaxoft: Yea, as I said, I'm blind in JS

madmaxoft:
/me likes, no objections from me towards a merge.

NiLSPACE:
Great, then I'll merge this :)

SphinxC0re:
I recognized, your are using es6 syntax. As long as you aren't using any transpiler this is not compatible with all systems.

NiLSPACE:
It should be compatible with all modern browsers. I've tried it on Windows with Firefox, Edge, Chrome. Linux on Firefox and Windows Phone with Edge.

SphinxC0re:
https://babeljs.io/

PR #3248, opened 2016-07-02
Use player UUIDs for cTeam

### LINKED ISSUE #3234: Make cTeam use UUID's instead of player names
Currently, as explained by @tigerw [here](https://forum.cuberite.org/thread-2466-post-26041.html#pid26041), cTeam manages it's members by player names. As Minecraft currently uses UUID's, this should be used instead.

PlasmaPower:
Resolves #3234 

Note: for those wondering, a UUID cannot also be a MC username.

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1232,14 +1232,17 @@ void cPlayer::SetTeam(cTeam * a_Team)
 
 	if (m_Team)
 	{
-		m_Team->RemovePlayer(GetName());
+		if (!m_Team->RemovePlayer(GetUUID()))
+		{
+			m_Team->RemovePlayer(GetName());
+		}
 	}
 
 	m_Team = a_Team;
 
 	if (m_Team)
 	{
-		m_Team->AddPlayer(GetName());
+		m_Team->AddPlayer(GetUUID());
 	}
 }
 
@@ -1256,8 +1259,14 @@ cTeam * cPlayer::UpdateTeam(void)
 	else
 	{
 		cScoreboard & Scoreboard = m_World->GetScoreBoard();
+		cTeam * Team = Scoreboard.QueryPlayerTeam(GetUUID());
 
-		SetTeam(Scoreboard.QueryPlayerTeam(GetName()));
+		if (Team == nullptr)
+		{
+			Team = Scoreboard.QueryPlayerTeam(GetName());
+		}
+
+		SetTeam(Team);
 	}
 
 	return m_Team;


madmaxoft:
This doesn't handle the loading and saving of teams, which still uses names even in Vanilla

PlasmaPower:
@madmaxoft Why doesn't it? If instead of names we send cTeam UUIDs, then it will save UUIDs instead of names.

madmaxoft:
That's exactly the problem. Vanilla saves player names into the same place, so this would mean we would become data-incompatible. If we decide to store UUIDs, we should change the storage so that the UUIDs get stored in a different NBT tag, and we should also provide compatibility with Vanilla - store playernames as well and when loading, if there are no UUIDs, use playernames.

PlasmaPower:
@madmaxoft That would also provide a backward compatible API, for better or for worse. I'll do that.

PlasmaPower:
@madmaxoft However, doing it that way would break scoreboards from both vanilla and older versions of cuberite, whereas without that it would only break saves *to* vanilla and older versions of cuberite (and moving back in cuberite versions is a lot more unlikely than moving forward).

PureTryOut:
Any progress on this?

PlasmaPower:
@PureTryOut I'm assuming you're talking to the maintainers, but I've totally forgotten about this. If someone with push access says it's good, I'll rebase it.

PureTryOut:
Well I was talking to anyone interested really :stuck_out_tongue:

I gladly see this merged so I can properly and reliable add players to teams in my [plugin](https://github.com/linuxmasterrace/Townvalds).

tigerw:
Could the team function be moved from cPlayer to cTeam (?). Was there any particular reason for the current design?.

PlasmaPower:
@PureTryOut Hmm, I wrote that comment while writing an implementation. Thinking back on it, I'm not sure why I thought that. Also, next time please put that edit as a separate comment so I'll get a notification for it.

PlasmaPower:
@tigerw Not sure why it's there. If I go with the store both UUIDs and player names method like I'm planning, then I'll make sure to deprecate the cPlayer method and keep it just as a wrapper for a cTeam method.

PlasmaPower:
Oh, [this is why](https://github.com/PlasmaPower/cuberite/blob/c81ce69e9a917cd3fdf9aacf42c4772d0d9214fa/src/Entities/Player.cpp#L934). Plus that's probably how vanilla does it (don't have deobf source on hand right now though, can't check easily).

PlasmaPower:
Hmm, using both UUIDs and player names separates the API into two independent sections: one that stores and uses UUIDs, and one that stores and uses player names. I could try to keep them linked, but because of worries about storage format compatibility there couldn't easily be a link between items on the two lists. Even if there was, implementations not caring about UUIDs could easily mess everything up.

This whole thing is starting to sound like a mess of compatibility and fragile systems. I doubt I'll be up to doing a much more complicated implementation than the current patchset.

PureTryOut:
I would think that plugins from now on can only use the UUIDs methods. I don't think much plugins use this function right now anyway, and it's good for them to change anyway, since it's more robust. API changes happen, that's just how it is.

I would say store both the UUID for newer Cuberite versions, and playernames for vanilla compatibility and older Cuberite versions. Then read it out again when needed by UUID.

PlasmaPower:
@PureTryOut But what happens when a mod requests a list of UUIDs, but the team was made in Vanilla and only has names?

tigerw:
If the issue is that Vanilla stores team data as names on disk, then a shim should be inserted into whatever functions read and write the team data to perform the necessary conversion - I believe we have routines for that, to answer your question. I suggest not to go through the hassle of storing two fields because we'd have to go through another whole deprecation process for one of them, sometime down the line. External (API) and internal code should then be converted to use UUIDs exclusively.

As a bonus (20% off your purchase), the team storage logic/API could also be separated from cPlayer and handed off to cTeam, if there is no reason not to.

PureTryOut:
Good call, I guess Cuberite would see there are no UUIDs but there are playernames, and them converts them to use UUIDs from that point onwards?

PureTryOut:
Yes I guess just converting UUIDs and storing them as playernames internally would be the best for now. That way if Mojang finally decides to do it right somewhere in the future, Cuberite would only need to remove the conversion and plugins will keep working like normal.

tigerw:
Yeah, something like that. Sounds like work though.

PlasmaPower:
What if we load a world, add UUIDs, but then the admin downgrades cuberite, then removes one player and ads another. The admin then upgrades cuberite (say the bug is fixed) and now, even though the list length is the same, things are broken as the UUIDs aren't matched up with the names. That means we have to re-fetch UUIDs every startup even if they already exist.

PlasmaPower:
Also, note to myself if I end up implementing this, there isn't a strict 1:1 correlation of names:UUIDs because a player might be registered on a team with both their old and current username.

PureTryOut:
>What if we load a world, add UUIDs, but then the admin downgrades cuberite, then removes one player and adds another to a team. The admin then upgrades cuberite

If you start thinking like this, it becomes one hell of a job to do. Should you really account for such things? Versions move so quickly here, I think after a week of having this new implementation, people will be downgrading to versions that already have it...

Seriously I would just convert UUIDs to names internally and back and be done with it.

PlasmaPower:
@PureTryOut Think about an admin switching back to a MC server then back to cuberite. Same thing.

Chances might be low, but it would cause a big mess if something goes wrong. Besides, I don't like hacky implementations like this. I'd like either a separate cuberite team storage (probably with a script to convert the traditional MC team storage to this, and maybe back too), or the current patch.

tigerw:
I do not understand the scenario. Proposed was a system to store only player names, converting to identifiers when the server uses them during operation, and saving to names when needed. The system should be linear with no duplication of data.

PureTryOut:
@tigerw indeed. @PlasmaPower not sure what would break if an admin switches to vanilla and back to Cuberite, if Cuberite just translates UUIDs internally to usernames and back.

When saving, convert the by the plugin given UUID to playername and store it like it does now. When reading, convert the username to UUID and give that to the plugin. This should be compatible with older Cuberite versions *and* vanilla MC.

PlasmaPower:
@tigerw Oh, I see. What if you want to remove a player from a large team? On average, half of the team will have to be queried. Are we okay with that?

tigerw:
Well, there is a cache, so at least the request won't be going through the network.

tigerw:
And anyway, the translation is done on load and save ideally, which should happen infrequently - maybe only on server start and stop (but I'm not sure about that).

PureTryOut:
@tigerw what if the server crashes? You would lose all the team changes you made.

PlasmaPower:
@PureTryOut The plan is only on load and save (not startup and shutdown).

PureTryOut:
Seems fine to me then.

PureTryOut:
I'm assuming you will go with the afore mentioned method?

PlasmaPower:
I'm busy, and unfortunately I don't think I'll have time to implement this anytime soon. If someone else wants to pick up this PR I'd be fine with that (although the current path won't be useful in the planned implementation).

PR #3252, opened 2016-07-05
Fixed piston destroying other pistons

### LINKED ISSUE #3251: Destroying a piston facing another piston destroys the extended part of the other piston
Client version: 1.8.9
Commit id: approx: 510614e8eff130839ab85f398317d307469f564f built at: approx: 2016-07-05 17:10:58 -0400

### Expected behavior
When an unextended piston is facing another pistons extended block, destroying the first piston should not destroy the second pistons extension.

### Actual behavior

Destroying the second piston will destroy the first pistons extension.

Set up two pistons so one is extended and the other is unextended like this:

Destroy the unextended piston and the extended part of the other piston will also be destroyed.

Altenius:
Fixes #3251

### DIFF
--- a/src/Blocks/BlockPiston.cpp
+++ b/src/Blocks/BlockPiston.cpp
@@ -32,13 +32,17 @@ void cBlockPistonHandler::OnDestroyed(cChunkInterface & a_ChunkInterface, cWorld
 {
 	Vector3i blockPos(a_BlockX, a_BlockY, a_BlockZ);
 
-	// Get the extension of the piston
 	NIBBLETYPE OldMeta = a_ChunkInterface.GetBlockMeta(blockPos.x, blockPos.y, blockPos.z);
-	blockPos += MetadataToOffset(OldMeta);
-
-	if (a_ChunkInterface.GetBlock(blockPos) == E_BLOCK_PISTON_EXTENSION)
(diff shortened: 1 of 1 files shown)


src/Blocks/BlockPiston.cpp
```diff
@@ -32,13 +32,17 @@ void cBlockPistonHandler::OnDestroyed(cChunkInterface & a_ChunkInterface, cWorld
 {
 	Vector3i blockPos(a_BlockX, a_BlockY, a_BlockZ);
 
-	// Get the extension of the piston
```
SphinxC0re: Keep this comment

madmaxoft:
Add back the comment, as SphinxC0re said, then rebase and squash all commits into a single one:
```
git fetch upstream
git rebase -i upstream/master
```

SphinxC0re:
@madmaxoft depending on the remotes' name

src/Blocks/BlockPiston.cpp
madmaxoft: This comment is no longer valid, I think a more general one would be better:
`If the piston is extended, destroy the extension as well`
And get rid of the empty line between setting `OldMeta` and testing it.

Altenius:
Updated

PR #3259, opened 2016-07-10
Fixed Lua Json serializing a bool.

### LINKED ISSUE #3257: bug with cJson??
Client version:   latest
Server OS:        macOS
Commit id:        bd2fc7716c4571dfd8840923b17d363dd19146d1

### Expected behavior
to serialize a lua table with a boolean to JSON

### Actual behavior
getting `null` on the bool field

ini looks like this
```ini
[section]
foo=1  
; since the cIniFile will only parse 1 or 0 as booleans :(
```
I serialize the iniFile to a table like so
```lua
ini = cIniFile()
ini:ReadFile('filepath')

val = ini:GetValueB( key , "foo")
table = {}
table.foo = val
```
and use cJson like this:
```lua
cJson:Serialize(table) -- results in "{foo: null}" 
```
When I log the type of `foo` it is `boolean`

madmaxoft:
Fixes #3257.

### DIFF
--- a/Server/Plugins/Debuggers/Debuggers.lua
+++ b/Server/Plugins/Debuggers/Debuggers.lua
@@ -2026,22 +2026,24 @@ end
 
 function HandleConsoleTestJson(a_Split, a_EntireCmd)
 	LOG("Testing Json parsing...")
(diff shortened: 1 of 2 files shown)


bearbin:
LGTM

PR #3265, opened 2016-07-17
Boat fixes

beeduck:
**Following fixes for boats:**
- Protocol for 0x10 is now handled (vehicle movement).
- Protocol for 0x11 is now handled (boat steering).
- Boats can only be spawned in water now. Similarly, boats can no longer spawn underwater, they can only be spawned at water level.
- Boats now properly float in water.
- Boat metadata now properly broadcasts, updating its position and paddles for other clients.

Once this pull is accepted I intend to work on handling boat spawning for all types of boats. Following that I will continue work on other vehicle types (horses and pigs).

### DIFF
--- a/src/Entities/Boat.h
+++ b/src/Entities/Boat.h
@@ -31,8 +31,29 @@ public:
 	virtual void HandleSpeedFromAttachee(float a_Forward, float a_Sideways) override;
 
 	cBoat(double a_X, double a_Y, double a_Z);
+
+	int GetLastDamage(void) const { return m_LastDamage; }
+	int GetForwardDirection(void) const { return m_ForwardDirection; }
+
+	float GetDamageTaken(void) const { return m_DamageTaken; }
+
+	int GetType(void) const { return m_Type; }
+
+	bool IsRightPaddleUsed(void) const { return m_RightPaddleUsed; }
+	bool IsLeftPaddleUsed(void) const { return m_LeftPaddleUsed; }
+
+	void SetLastDamage(int TimeSinceLastHit);
+
+	void UpdatePaddles(bool rightPaddleUsed, bool leftPaddleUsed);
+
+private:
+	int m_LastDamage;
+	int m_ForwardDirection;
+
+	float m_DamageTaken;
+
+	int m_Type;
+
+	bool m_RightPaddleUsed;
+	bool m_LeftPaddleUsed;
 } ;
-
-
-
-
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1038,6 +1038,20 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 			NextSpeed -= NextSpeed * (m_AirDrag * 20.0f) * DtSec.count();
 		}
 		NextSpeed.y += static_cast<float>(fallspeed);
+
+		// A real boat floats
+		if (IsBoat())
+		{
+			// Find top water block and sit there
+			int NextBlockY = BlockY;
+			BLOCKTYPE NextBlock = NextChunk->GetBlock(RelBlockX, NextBlockY, RelBlockZ);
+			while (IsBlockWater(NextBlock))
+			{
+				NextBlock = NextChunk->GetBlock(RelBlockX, ++NextBlockY, RelBlockZ);
+			}
+			NextPos.y = NextBlockY - 0.5;
+			NextSpeed.y = 0;
+		}
 	}
 	else
 	{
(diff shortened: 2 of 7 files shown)


src/Entities/Boat.h
```diff
@@ -31,8 +31,27 @@ class cBoat :
...
+	int GetForwardDirection(void) const { return m_ForwardDirection; }
+
+	float GetDamageTaken(void) const { return m_DamageTaken; }
+
+	int GetType(void) const { return m_Type; }
+
+	bool GetRightPaddle(void) const { return m_RightPaddle; }
+	bool GetLeftPaddle(void) const { return m_LeftPaddle; }
 
+	void UpdatePaddles(const bool & rightPaddle, const bool & leftPaddle);
```
madmaxoft: Why pass `bool`s by-reference, instead of directly?

src/Protocol/Protocol19x.cpp
```diff
@@ -2318,6 +2319,28 @@ void cProtocol190::HandlePacketBlockPlace(cByteBuffer & a_ByteBuffer)
...
+
+	// Get the players vehicle
+	cPlayer * Player = m_Client->GetPlayer();
+	cEntity * Vehicle = Player->GetAttached();
+
+	if (Vehicle)
+	{
+		if (Vehicle->GetEntityType() == cEntity::etBoat)
+		{
+			((cBoat*) Vehicle)->UpdatePaddles(RightPaddle, LeftPaddle);
```
madmaxoft: Don't use C-style cast.

src/Entities/Boat.h
```diff
@@ -31,8 +31,27 @@ class cBoat :
 	virtual void HandleSpeedFromAttachee(float a_Forward, float a_Sideways) override;
 
 	cBoat(double a_X, double a_Y, double a_Z);
-} ;
 
+	int GetTimeSinceLastHit(void) const { return m_TimeSinceLastHit; }
```
madmaxoft: `TimeSince` is neither a good name nor a good measure.
What are the units? Why not call it `Ticks` instead to make it clear?
Why make a variable that needs updating each tick to maintain it? Make it `LastHitWorldAge`, specifying the world's age (in ticks) when the boat was last hit, so that it needs no updating, yet it conveys the same information.

Also, I don't see the code that actually does update this field.

src/Entities/Entity.cpp
```diff
@@ -1038,6 +1038,20 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+		if (IsBoat())
+		{
+			// Find top water block and sit there
+			int NextBlockY = BlockY;
+			BLOCKTYPE NextBlock = NextChunk->GetBlock(RelBlockX, NextBlockY, RelBlockZ);
+			while (IsBlockWater(NextBlock))
+			{
+				NextBlock = NextChunk->GetBlock(RelBlockX, ++NextBlockY, RelBlockZ);
+			}
+			NextPos.y = NextBlockY - .5;
```
madmaxoft: Use full number format, `0.5`

src/Entities/Boat.h
```diff
@@ -31,8 +31,27 @@ class cBoat :
...
+private:
+	int m_TimeSinceLastHit;
+	int m_ForwardDirection;
 
+	float m_DamageTaken;
+
+	int m_Type;
+
+	bool m_RightPaddle;
+	bool m_LeftPaddle;
```
madmaxoft: Use a better name. `Paddle` what, Present? Moved? Used? Crafted?

src/Entities/Boat.h
beeduck: Just an old habit. I'll change it to direct pass if that's the preferred style.

src/Entities/Boat.h
beeduck: This is one of the variables specified in the boat metadata. I'm not sure where exactly it's used or why they've named it as such, as I couldn't find any protocols that pass it other than when the metadata is being passed to other clients for boats.

I'll look a bit more into it and see if I can figure out its purpose.

beeduck:
Responded to a few comments, I'll fix that others that have been noted. Do you want one total commit once the requested changes are made?

src/Entities/Boat.h
Pokechu22: It's used to control how the boat wobbles visually.  It's set to 10 when the boat is hit (attacked), and decremented by 1 each tick if it's greater than 0.

Since it's used as a countdown clientside, decreasing it each tick probably would be easier to handle than using the time it was hit.  Even still, `TimeSinceLastHit` really doesn't describe how it is used clientside - it's more how much the boat is wobbling (MCP does use `TIME_SINCE_HIT` to describe it, which is probably why wiki.vg calls it that).  

Minecarts have the exact same type of code for their "Shaking power" value (set to 10 on hit, decremented each tick).

beeduck: I've changed time since last hit to follow the model of minecarts which used the name "LastDamage"

I've also set LastDamage to 10 upon taking damage and decrement it by 1 each tick, per Pokechu22's recommendation.

beeduck:
Made recommended changes, let me know if there's anything else I should adjust.

madmaxoft:
Thanks

PR #3268, opened 2016-07-20
Fix endianness issue?

bearbin:
This fixes endianness on ODROID-c2 devices.


madmaxoft:
Looks good to me.

bearbin:
Someone should make sure that the endianness check includes all the architectures currently in use - it is most likely incomplete ATM.

Issue #2563, opened 2015-10-25
Playername tab completion issue


mathias-github:
Tab completion for playernames doesn't work if the player is located in another world than you are.

PR #3270, opened 2016-07-20
Possible actual fix for #2563

bearbin:
Add tab completion to names across world boundaries.

For some reason tolua fails for this on my machine so I have to create a PR.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1831,7 +1831,33 @@ void cClientHandle::HandleUnmount(void)
 void cClientHandle::HandleTabCompletion(const AString & a_Text)
 {
 	AStringVector Results;
-	m_Player->GetWorld()->TabCompleteUserName(a_Text, Results);
+	if (cRoot::Get()->GetServer()->ShouldAllowMultiWorldTabCompletion())
+	{
+		class cWorldCallback : public cWorldListCallback
+		{
+		public:
+			cWorldCallback(AStringVector & a_Results, const AString & a_Search) :
+				m_Results(a_Results),
+				m_Search(a_Search)
+			{
+			}
+
+			virtual bool Item(cWorld * a_World) override
+			{
+				a_World->TabCompleteUserName(m_Search, m_Results);
+				return true;
+			}
+		private:
+			AStringVector & m_Results;
+			const AString & m_Search;
+		} WC(Results, a_Text);
+
+		cRoot::Get()->ForEachWorld(WC);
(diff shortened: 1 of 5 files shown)


bearbin:
@madmaxoft Could you review this please?

src/ClientHandle.cpp
```diff
@@ -1831,7 +1831,33 @@ void cClientHandle::HandleUnmount(void)
...
+			{
+				a_World->TabCompleteUserName(m_Search, m_Results);
+				return true;
+			}
+		private:
+			AStringVector & m_Results;
+			const AString & m_Search;
+		} WC(Results, a_Text);
+
+		cRoot::Get()->ForEachWorld(WC);
```
madmaxoft: Extract this into a separate function in `cRoot`, so that we can export it to Lua API later on when we open up the tab completion to plugins completely.

bearbin:
@madmaxoft done

mathias-github:
Something's not right in this PR. Tab completion now only works on one of the players online.

bearbin:
@mathias-github Are you sure that the issue wasn't present on a previous commit? Can you provide a screenshot/video?

mathias-github:
I compiled Cuberite a few minutes ago. The commit ID is 56a0c48580f231aa73b0a4a577fe048afcbcafac. You can test it on my server flame.ga.

bearbin:
Huh, I couldn't get tab completion to work at all.

tigerw:
The change looks innocuous enough... Was it working before this PR?

mathias-github:
Yes, tab completion worked before this PR.

src/Root.cpp
```diff
@@ -1006,3 +1006,33 @@ int cRoot::GetFurnaceFuelBurnTime(const cItem & a_Fuel)
...
+		cWorldCallback(AStringVector & a_Results, const AString & a_Search) :
+			m_Results(a_Results),
+			m_Search(a_Search)
+		{
+		}
+
+		virtual bool Item(cWorld * a_World) override
+		{
+			a_World->TabCompleteUserName(m_Search, m_Results);
+			return true;
```
tigerw: I suppose the first thing that comes to mind is that returning true [here](https://github.com/cuberite/cuberite/blob/master/src/Root.cpp#L580) will stop iteration, so if you don't get the right world first time, no tabcomplete.

bearbin: Well, that would explain it then.. I will repair it.

PR #3271, opened 2016-07-21
CMake: Fix system Lua usage for non-5.1 versions.

### LINKED ISSUE #3269: Tolua internal error when compiling
```
** tolua internal error: [string "../../lib/tolua++/src/bin/lua/package.lua"]:341: attempt to call global 'setfenv' (a nil value)
stack traceback:
	[string "../../lib/tolua++/src/bin/lua/package.lua"]:341: in function 'prep'
	[string "../../lib/tolua++/src/bin/lua/package.lua"]:253: in function 'Package'
	[string "../../lib/tolua++/src/bin/lua/doit.lua"]:55: in function <[string "../../lib/tolua++/src/bin/lua/doit.lua"]:27>
	[C]: in function 'xpcall'
	../../lib/tolua++/src/bin/lua/all.lua:25: in main chunk
	[C]: in function 'dofile'
	BindingsProcessor.lua:55: in function 'invokeToLua'
	BindingsProcessor.lua:990: in main chunk
	[C]: in ?.
```

Is this the fault of my code or tolua?

madmaxoft:
Fixes #3269 .


madmaxoft:
Also added some documentation to the top of the main CMakeLists.txt file, listing the options that the script supports.

bearbin:
Should CROSSCOMPILE actually be mutually exclusive with using the provided tolua? CROSSCOMPILE only removes optimisations for your architecture, but you should still be able to run the output.

Actually cross-compiling for another architecture needs another method that doesn't use CROSSCOMPILE.

bearbin:
Also, how does this PR actually fix the issue - is it just assuming that you will use the DISABLE_SYSTEM_LUA flag? It would be better if it just detected the correct lua if it was installed.

madmaxoft:
The consensus was that for cross-compiling you will have to have Lua interpreter installed, so that the crosscompiling doesn't need to build anything for the actual host platform (before #3204 , crosscompiling required that the `tolua` executable was built for the host platform, executed, and then the rest of the build for the target platform is run. This is rather difficult, it's much easier to enforce system Lua usage for this scenario)

The fix in this PR is that the missing or bad-version Lua unsets an internal flag, `USE_SYSTEM_LUA`, which is then used for the actual decision whether to use system lua or not. Before, the code checked the presence (`HAS_LUA_INTERPRETER`) and version, but only used the presence flag for the decision (so it tried to use Lua 5.2 for the build, which does not work).

The `DISABLE_SYSTEM_LUA` is a slightly unrelated parameter, you can use it on the cmake commandline to actually disable the system Lua detection and usage, should you wish to do so. It doesn't have any relevance to the bug.

bearbin:
> The consensus was that for cross-compiling you will have to have Lua interpreter installed, so that the crosscompiling doesn't need to build anything for the actual host platform (before #3204 , crosscompiling required that the tolua executable was built for the host platform, executed, and then the rest of the build for the target platform is run. This is rather difficult, it's much easier to enforce system Lua usage for this scenario)

This is absolutely true, but the CROSSCOMPILE flag has nothing to do with actually crosscompiling. It just optimises for the lowest common denominator architecture rather than CPU specific.

> The fix in this PR is that the missing or bad-version Lua unsets an internal flag, USE_SYSTEM_LUA, which is then used for the actual decision whether to use system lua or not. Before, the code checked the 

Great, thanks for clearing that up.

madmaxoft:
In that case, the `CROSSCOMPILE` flag is badly named and should be renamed to something more clearly representing its purpose. I don't think it should be done in this PR, though, this is a hotfix for the Pure-Lua changes.

Issue #3275, opened 2016-07-25
Add cuberite.org links for download/compile


bearbin:
i.e.

    wget download.cuberite.org | sh

bearbin:
Service is now set up on the server:

https://download.cuberite.org
https://compile.cuberite.org

madmaxoft:
I think the `download.cuberite.org` webserver should serve a 302 redirect to the HTML download page; the download script should have a (small) postfix to the URL. There is a possibility of people trying to visit the `download.cuberite.org` domain.

bearbin:
GitHub pages doesn't do 302 redirects though, it would make it significantly more complicated. 

If you're worried about people being confused, adding a comment to the top of the script could be a solution?

tigerw:
it's awfully Linux centric... Is this common practice?

bearbin:
Yes, this is Linux only, but how often do you come across a Linux user who wants to download the server from the command line?

madmaxoft:
How about using `dl.cuberite.org` for the script, then? And point `download.cuberite.org` to the `cuberite.org` frontpage.

LogicParrot:
Relevant: https://github.com/cuberite/cuberite/pull/3278

>After this PR, An adversary can either comprimise Github.com or cuberite.org to infect downloaders by serving bad code, so by relying on two servers to serve the shell code, rather than one, we double the attack surface. Is this worth the aesthetical change?

bearbin:
@madmaxoft download.cuberite.org never existed before, so I don't understand how users would be confused as to the purpose of the domain. And how would dl.cuberite.org be any better? 

I answered @"LogicParrot"'s concern in #3278.

sphinxc0re:
Pointing `download.cuberite.org` to `cuberite.org` would be bad SEO because of duplicated content

PR #3278, opened 2016-07-25
Switch out long download/compile links.

bearbin:
The links have been replaced with short links that forward to the
long links.

Fixes issue #3275

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ There are several ways to obtain Cuberite.
 ##### The EasyInstall script
 This Linux script will download the correct binary from the project site.
 
-    sh -c "$(wget -O - https://raw.githubusercontent.com/cuberite/cuberite/master/easyinstall.sh)"
+    sh -c "$(wget -O - https://download.cuberite.org)"
 
 #### Compiling
  - You can compile automatically for Linux / *nix with the `compile.sh` script. The script is described below.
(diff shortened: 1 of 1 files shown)


LogicParrot:
After this PR, An adversary can either comprimise `Github.com` or `cuberite.org` to infect downloaders, so by relying on two servers to serve the shell code, rather than one, we double the attack surface. Is this worth the aesthetical change?

bearbin:
@LogicParrot This is not intended to be a high-use endpoint, more just for people to easily try out the server the first time. 

In that case, the attack surface is not actually increased, since they would have been downloading the binaries from https://cuberite.org instead. It think this is worth the additional risk if there is any.

tigerw:
I suppose the idea is fine. Is there a way to make the server behind download/compile detect the platform and serve the correct whatever?

fralix:
I think not, because the syntax of linux / windows scripts are different. But you could check the user agent like the user agent of curl is "curl/7.35.0" and for wget "Wget/1.15" (linux-gnu) (check for the curl /wget and differ now - but you should add a comment if the script doesn't work correctly and may be an option to select the right version like download.cuberite.org/?os=windows )

But the idea sounds nice

bearbin:
It is possible to write shell scripts that work on both windows and linux you know?

https://github.com/cuberite/cuberite/blob/master/compile.sh

bearbin:
or people can just use bash on linux now as well.

tigerw:
Alright. You've already done it anyway, so I have no further input.

LogicParrot:
>In that case, the attack surface is not actually increased, since they would have been downloading the binaries from https://cuberite.org instead. It think this is worth the additional risk if there is any.

I was talking about the compile link.

Schwertspize:
Ehm just asking, our cuberite.org is hosted on github right? cuberite.github.io with a dns record cuberite.org

LogicParrot:
>Ehm just asking, our cuberite.org is hosted on github right? cuberite.github.io with a dns record cuberite.org

Not anymore. Because that wouldn't provide https.
`Cuberite.org` has a proxy server to `cuberite.github.io` with a long cache.

PR #3279, opened 2016-07-25
Change CROSSCOMPILE flag to NO_NATIVE_OPTIMIZATION

bearbin:


madmaxoft:
Don't let the natives march! (@tigerw -style pun :)

I don't like the new name of the flag. How about `NO_NATIVE_OPTIMIZATION` instead? Other compilers could have a similar flag and this says what it does, instead of what it is.

bearbin:
Does anybody else have an opinion on this? Otherwise I'll change it.

tigerw:
> Don't let the natives march! (@tigerw -style pun, so proud of myself :)

Hehe, my plans are working >:D

bearbin:
@madmaxoft I have changed the name to NO_NATIVE_OPTIMIZATION as you requested.

CMakeLists.txt
```diff
@@ -3,11 +3,11 @@
...
 #
 # This script supports some configuration through CMake arguments (-Dparam=val syntax):
-#   BUILD_TOOLS=1          sets up additional executables to be built along with the server (ProtoProxy, GrownBiomeGenVisualiser, MCADefrag)
-#   BUILD_UNSTABLE_TOOLS=1 sets up yet more executables to be built, these can be broken and generally are obsolete (GeneratorPerformanceTest)
-#   CROSSCOMPILE=1         marks the build as a crosscompile (doesn't include optimizations for current platform)
-#   DISABLE_SYSTEM_LUA=1   disables the use of system Lua interpreter; the tolua executable will be built and used instead. Incompatible with cross-compiling
-#   SELF_TEST=1            enables testing code to be built
+#   BUILD_TOOLS=1            sets up additional executables to be built along with the server (ProtoProxy, GrownBiomeGenVisualiser, MCADefrag)
+#   BUILD_UNSTABLE_TOOLS=1   sets up yet more executables to be built, these can be broken and generally are obsolete (GeneratorPerformanceTest)
+#   NO_NATIVE_OPTIMIZATION=1 disables CPU-specific optimisations, allows use on other CPUs of the same platform
```
madmaxoft: How about `disables CPU-specific optimisations for current machine` to make it blindingly obvious for us non-Linuxers? :)

COMPILING.md
```diff
@@ -144,31 +144,27 @@ to your cmake command and 32 bit will be forced.
 
 ### Compiling for another computer ###
 
-When cross-compiling for another computer it is important to set cross compiling mode. This tells the compiler not to optimise for your machine. This switch can be used with debug or release mode. To enable, simply add:
+When cross-compiling for another computer of the same architechture it is important to set the NO_NATIVE_OPTIMIZATION flag. This tells the compiler not to optimise for your machine. This switch can be used with debug or release mode. To enable, simply add:
```
madmaxoft: Add `of the same architecture` to the heading as well?

CMakeLists.txt
bearbin: Your suggestion here just makes it more confusing for me, I would be inclined not to include it.

COMPILING.md
bearbin: Done.

CMakeLists.txt
madmaxoft: It's just that `CPU-specific` sounds too general to me, I'd understand it as "disable optimizations that are x86-specific", rather than "for this particular CPU flavor".

bearbin: Done.

madmaxoft: Well you don't need to take me as the absolute. After all, English is not my native language, so perhaps you can come up with an even better description?
How about this: `disables optimisations specific for the CPU in this machine. Without this flag, the executable need not work on any other CPU, even of the same platform` (I'm a bit confused about "may not" and "need not", which one is better here?)

Issue #3138, opened 2016-04-12
Tab list sometimes doesn't update


mathiascode:
Sometimes, when a player leaves the server, their name stays in the online list that appears when pressing the tab key. This seems to happen randomly.

bearbin:
I think this is fixed with the latest build. Could you confirm @mathias-github ?

mathias-github:
This is not completely fixed yet. A player who joined and disconnected quickly didn't disappear from the tab list.

PR #3280, opened 2016-07-25
Fix tab list sometimes not updating after a player leaves.

bearbin:
When a player leaves the game, their client handle has its state
changed to csDestroyed.

This should fix issue #3138 

@madmaxoft how did this even work in the first place? Is there an error in the logic somewhere else?


bearbin:
@madmaxoft Is it possible this could get reviewed?

tigerw:
Might have been just a copy-paste error. Looks fine.

PR #3281, opened 2016-07-26
Fixed RasPi builds of unit tests.

madmaxoft:
On RasPi with gcc 4.8.2, the asserts wouldn't compile when tests were enabled.

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -370,15 +370,22 @@ template class SizeChecker<UInt8,  1>;
 			fflush(stdout); \
 		}
 	#endif
-	#define ASSERT(x) do { if (!(x)) { throw cAssertFailure();} } while (0)
-	#define testassert(x) do { if (!(x)) { REPORT_ERROR("Test failure: %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } } while (0)
-	#define CheckAsserts(x) do { try {x} catch (cAssertFailure) { break; } REPORT_ERROR("Test failure: assert didn't fire for %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } while (0)
+
+	#ifdef _DEBUG
+		#define ASSERT(x) do { if (!(x)) { throw cAssertFailure();} } while (0)
+		#define testassert(x) do { if (!(x)) { REPORT_ERROR("Test failure: %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } } while (0)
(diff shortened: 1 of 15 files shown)


madmaxoft:
@worktycho could you have a look at this, please?

There are two main points to this change:
- `cLuaState::CopyTableFrom()` was using `assert` instead of `ASSERT`
- The `ASSERT` macro contained code that compiled into Release-mode executables (when self-tests were enabled)

I've changed the logic, so that anything inside an `ASSERT` produces code only in Debug builds. Self-tests in Release mode cannot therefore check for asserts, and instead log an info message about this to.
There's a problem with any other approach that sometimes the `ASSERT`ed expressions use variables that are declared only for that `ASSERT` - so they need to be guarded with an `#ifdef _DEBUG`, which interfered with the Release mode self-tests.

The rest of the changes are the results of clang complaining about dead code.

tigerw:
Thank thee for not introducing HAS_ASSERT :)

madmaxoft:
Seems @worktycho is busy; no-one else has voiced their concerns, and I really need this to go in to be able to compile on the RasPi. Will merge within a few hours.

madmaxoft:
@LogicParrot what are you doing? We DON'T mege master into branches, we rebase (or squash-merge)

LogicParrot:
I thought Github now supports automatically squashing it all into 1 commit on merging the PR?

tigerw:
Can confirm it worked for redstone changes.

src/Globals.h
```diff
@@ -370,15 +370,22 @@ template class SizeChecker<UInt8,  1>;
...
 		}
 	#endif
-	#define ASSERT(x) do { if (!(x)) { throw cAssertFailure();} } while (0)
-	#define testassert(x) do { if (!(x)) { REPORT_ERROR("Test failure: %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } } while (0)
-	#define CheckAsserts(x) do { try {x} catch (cAssertFailure) { break; } REPORT_ERROR("Test failure: assert didn't fire for %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } while (0)
+
+	#ifdef _DEBUG
+		#define ASSERT(x) do { if (!(x)) { throw cAssertFailure();} } while (0)
+		#define testassert(x) do { if (!(x)) { REPORT_ERROR("Test failure: %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } } while (0)
+		#define CheckAsserts(x) do { try {x} catch (cAssertFailure) { break; } REPORT_ERROR("Test failure: assert didn't fire for %s, file %s, line %d\n", #x, __FILE__, __LINE__); exit(1); } while (0)
```
tigerw: I always though asserting firing was bad. It's confusing the asserts not firing are errors.

madmaxoft: This is for tests that actually verify that when they pass bad data, an assert is fired.

PR #3292, opened 2016-08-02
Channel management

SamJBarney:
Allows lua plugins to register handles for channel messages.
- Only one handle can be registered for one channel at a time.
- Plugins can also add and remove clients from channels, sending the appropriate packet to the client

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -437,6 +437,22 @@ return
 			},
 		}, -- cBlockInfo
 
+		cChannelManager =
+		{
+			Desc = [[
+				This class allows plugins to register handlers for specific plugin channels.
+			]],
+
+			Functions =
+			{
+				AddClientToChannel = { Params = "{{cClientHandle}}, Channel Name", Notes = "Adds and registers a client on the specified channel." },
+				RemoveClientFromChannel = { Params = "{{cClientHandle}}, Channel Name", Notes = "Removes and unregisters a client on the specified channel." },
+				RegisterChannel = { Params = "Channel Name, Callback", Return = "bool", Notes = "Registers a handler for a specific channel. Returns true if it was able to be registered." },
+				RemoveChannel = { Params = "Channel Name", Return = "bool", Notes = "Removes the handler for the specified channel. Returns true if the handler was removed." },
+				BroadcastChannelMessage = { Params = "Channel Name, Message, {{cWorld|World}}", Notes = "Sends the message on the specified channel. Limits the message to the specified world, if provided." }
+			}
+		}, -- cChannelManager
+
 		cChatColor =
 		{
 			Desc = [[
@@ -2288,6 +2304,7 @@ end
 			Functions =
 			{
 				DoesAllowMultiLogin = { Params = "", Return = "boolean", Notes = "Returns true if players can log in multiple times from the same account (normally used for debugging), false if only one player per name is allowed." },
+				GetChannelManager = { Return = "{{cChannelManager}}", Notes = "Returns the {{cChannelManager|Channel Manager}}." },
 				GetDescription = { Return = "string", Notes = "Returns the server description set in the settings.ini." },
 				GetMaxPlayers = { Return = "number", Notes = "Returns the max amount of players who can join the server." },
 				GetNumPlayers = { Return = "number", Notes = "Returns the amount of players online." },
--- a/src/Bindings/CMakeLists.txt
+++ b/src/Bindings/CMakeLists.txt
@@ -84,6 +84,8 @@ set(BINDING_DEPENDENCIES
 	../BlockID.h
 	../BlockInfo.h
 	../BoundingBox.h
+	../ByteBuffer.h
+	../ChannelManager.h
 	../ChatColor.h
 	../ChunkDef.h
 	../ClientHandle.h
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -18,6 +18,7 @@
 #include "../Entities/Player.h"
 #include "../WebAdmin.h"
 #include "../ClientHandle.h"
+#include "../ChannelManager.h"
 #include "../BlockArea.h"
 #include "../BlockEntities/BeaconEntity.h"
 #include "../BlockEntities/BrewingstandEntity.h"
@@ -39,6 +40,7 @@
 #include "../BuildInfo.h"
 #include "../HTTP/UrlParser.h"
 #include "../BoundingBox.h"
+#include "../ChannelCallback.h"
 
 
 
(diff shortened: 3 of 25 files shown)


bearbin:
CircleCI build failed, also the commits don't have your email address associated with them.

SamJBarney:
Hmmmm... I need to fix that. One sec

SamJBarney:
CircleCI is failing, but the line it is asking for exists :/

SamJBarney:
Oh. Right. I need to add documentation for the Lua Bindings. lol

Pokechu22:
> Only one handle can be registered for one channel at a time.

I would say that this isn't a good idea.  There are valid situations where multiple plugins will want to listen to the same channel (and more commonly send on the same channel) - blocking them from doing so doesn't seem necessary.  IMO the general hook should also be called even if a plugin is listening to a specific channel.

SamJBarney:
@Pokechu22 The only instances where channels are use are with builtin `MC|` channels, which this channel doesn't affect, and Forge clients with mods, which they use to send custom packets. AFAIK, only the mods that create the channel communicate on that channel.

Also, what would these valid situations be where you would have more than one listener per channel?

SamJBarney:
@Pokechu22 
As for calling the general hook as well, I believe that would be inefficient. I originally came up with this while I was trying to figure out how to make plugins work for forge mods. If I was to emulate a heavily modded server (60+ mods), each plugin would have its hook called anytime a plugin packet would come in, and only 1 of them would need the data from the packet. That's 59+ hooks being called that don't need to, and that is extremely inefficient.

That is when I came up with this. It allows only 1 call into lua instead of multiple. If other plugins need information from other plugins, or to set some data, they can use the plugin's provided interface.

Pokechu22:
With Bukkit plugins you can have multiple plugins listening on the same channel.  Also, don't forget about liteloader mods (and also base edits), which can also use plugin channels - WorldEditCUI is done via liteloader if I recall correctly.

I'm having trouble coming up with a _specific_ situation where multiple listeners is important, but I can see it being useful for a logging plugin and a handler plugin.

For performance reasons, yes, it makes sense to not broadcast it to all plugins, but isn't that still confusing behavior (since the hook is only called some of the time)?

tigerw:
What would the advantage be for callbacks vs testing for a channel when the hook fires?

SamJBarney:
Bukket plugins are all server-side. There are no client-side bukket plugins. And if there are bukket plugins that mimic forge mods, they really should be built in a single package instead of multiple.

This was really built to replace that hook, for the most part. Calling the hook only happens if there are no handlers for it, so it acts as a backup.

As for logging, I'm ambivalent on it.

SamJBarney:
@tigerw One of my previous comments touches on the subject.

tigerw:
More granular hook listener registrations, essentially?

Seems like something which could be generalised to the plugin manager for more future proof API design. But that's a lot of work.

SamJBarney:
I'm not sure it could what other hooks would benefit from this, as most hooks need to allow multiple handlers for an event.
I would love to rework this so that the data coming in is parsed according to packet specifications, allowing channel data to be pre-parsed before passing it to the handler. But that is a story for another day

SamJBarney:
Can someone verify that I have added the ChannelManagement.h file to `src/Bindings/CMakeLists.txt` as required? I believe I've done it correctly.

SamJBarney:
@bearbin I think I put the file that CircleCI is requesting in correctly. Could you verify that for me please?

bearbin:
I think the new header file should be in BINDING_DEPENDENCIES, not HDRS according to the script. I think that is the correct place so it's not an error.

madmaxoft:
(replying blind, can't see code on my phone)
A header that is used for generating bindings through ToLua++ should be in both. HDRS is the list of files included into IDE projects ,and BINDING_DEPENDENCIES is used to generate proper dependencies for the generated files.
-- 

lib/tolua++
```diff
@@ -1 +1 @@
-Subproject commit 5d69f56b06fdfe1e67cd2572083abb863ec3b708
+Subproject commit 44cd937e7687dadec80068116a2e1d683b139d06
```
tigerw: ?

src/Bindings/ManualBindings.cpp
```diff
@@ -3718,6 +3720,50 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
...
+	// Check params:
+	cLuaState L(tolua_S);
+	if (!L.CheckParamUserType(1, "cChannelManager"))
+	{
+		return 0;
+	}
+	cChannelManager * self = reinterpret_cast<cChannelManager *>(tolua_tousertype(tolua_S, 1, nullptr));
+	if (self == nullptr)
+	{
+		return cManualBindings::lua_do_error(tolua_S, "invalid 'self' in function 'cCompositeChat:UnderlineUrls'");
```
tigerw: ?

src/Bindings/ManualBindings.cpp
```diff
@@ -3718,6 +3720,50 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
...
+	cLuaState::cCallbackPtr Callback;
+	L.GetStackValue(3, Callback);
+	auto ChannelCallback = std::make_shared<cChannelCallback>(*Plugin, Callback);
+
+	self->RegisterChannel(Channel, ChannelCallback);
+
+	// Cut away everything from the stack except for the cChannelManager reference; return that
+	lua_settop(L, 1);
+	return 1;
+}
```
tigerw: Is tolua really incapable of generating this automatically?

src/Bindings/ManualBindings.cpp
```diff
@@ -3718,6 +3720,50 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
...
+
+	AString Channel;
+	L.GetStackValue(2, Channel);
+	cLuaState::cCallbackPtr Callback;
+	L.GetStackValue(3, Callback);
+	auto ChannelCallback = std::make_shared<cChannelCallback>(*Plugin, Callback);
+
+	self->RegisterChannel(Channel, ChannelCallback);
+
+	// Cut away everything from the stack except for the cChannelManager reference; return that
```
tigerw: Don't you want to return what RegisterChannel returns?

src/ChannelCallback.h
```diff
@@ -0,0 +1,12 @@
+#include "Bindings/PluginLua.h"
+
+class cChannelCallback
```
tigerw: Man, callbacks? In C++11, llamas!

uh

lambdas are the new rad!

----

Is this possible with lambdas, so you don't introduce an extra class + two files? Though again, I remember `cLuaState::cCallback` being xoft's creation.

If so, xoft, can't lua directly pass functions to C++?

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "World.h"
+
+
+class cChannelBroadcastCallback : public cPlayerListCallback
+{
+	AString m_Channel;
+	AString m_Data;
+public:
+	cChannelBroadcastCallback(const AString & a_Channel, const AString & a_Data)
+	: m_Channel(a_Channel), m_Data(a_Data){}
```
tigerw: My goodness, style my dear sir, style!

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "ChannelCallback.h"
+#include "ClientHandle.h"
+#include "Entities/Player.h"
+#include "OSSupport/CriticalSection.h"
+#include "World.h"
+
+
+class cChannelBroadcastCallback : public cPlayerListCallback
+{
+	AString m_Channel;
```
tigerw: Could this be moved under it's own private tag, underneath the public declarations? For consistency :)

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+{
+	auto channels = BreakApartChannels(a_Channels);
+	a_Client.UnregisterChannels(channels);
+}
+
+
+
+
+
+bool cChannelManager::ClientHasChannel(cClientHandle & a_Client, const AString & a_Channel)
```
tigerw: Is this function *really* necessary? :P

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+
+
+
+bool cChannelManager::RegisterChannel(const AString & a_Channel, cChannelCallbackPtr a_Callback)
+{
+	cCSLock Lock(m_CSCallbacks);
+	if (m_ChannelCallbacks.find(a_Channel) == m_ChannelCallbacks.end())
+	{
+		m_ChannelCallbacks.insert(std::make_pair(a_Channel, a_Callback));
```
tigerw: I think (?) insert will tell you if the value was already present, so you can do a one-liner.

Also, if you use emplace, I again think (?) you can drop the make_pair.

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+	return false;
+}
+
+
+
+
+
+void cChannelManager::AddClientToChannel(cClientHandle & a_Client, const AString & a_Channel)
+{
+	a_Client.RegisterChannel(a_Channel);
```
tigerw: Note: ChannelManager appears multithread-capable whereas ClientHandle is not.

Not much you can do though if plugins have a ClientHandle and they're misusing it.

Is this function and its counterpart necessary? They don't really have any handling for server-to-client plugin messaging.

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+	// Otherwise try to use it.
+	if (has_channel)
+	{
+		cCSLock Lock(m_CSCallbacks);
+
+		auto callback = m_ChannelCallbacks.find(a_Channel);
+
+		// If a callback has been registered for this channel, call it.
+		// Otherwise, call the hook
+		if (has_channel && (callback != m_ChannelCallbacks.end()))
```
tigerw: Is has_channel not always true here?

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+		{
+			callback->second->Call(a_Client, a_Data);
+		}
+		else
+		{
+			cPluginManager::Get()->CallHookPluginMessage(a_Client, a_Channel, a_Data);
```
tigerw: Oh no, backwards compatibility. ;P

I doubt anyone would notice if you just... removed the hook... or at least deprecate it.

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+	}
+	else
+	{
+		// Ignore if client sent something but didn't register the channel first
+		LOGD("Player %s sent a plugin message on channel \"%s\", but didn't REGISTER it first", a_Client.GetUsername().c_str(), a_Channel.c_str());
+		a_Client.SendPluginMessage("UNREGISTER", a_Channel);
+	}
+
+}
+void cChannelManager::HandlePluginUnloading(const cPluginLua * a_Plugin)
```
tigerw: xoft's not gonna be happy about the spacing here, I think.

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+{
+	cCSLock Lock(m_CSCallbacks);
+
+	for (auto it = m_ChannelCallbacks.begin(); it != m_ChannelCallbacks.end(); ++it)
```
tigerw: I suppose a ranged loop wouldn't work?

src/ChannelManager.h
```diff
@@ -0,0 +1,70 @@
...
+{
+// tolua_end
+public:
+	typedef SharedPtr<cChannelCallback> cChannelCallbackPtr;
+	typedef std::unordered_map<AString, cChannelCallbackPtr> CallbackMap;
+	cChannelManager()
+	{
+
+	}
+	~cChannelManager(){}
```
tigerw: Are these necessary?

src/Server.h
```diff
@@ -143,6 +144,9 @@ class cServer
 	from the settings. */
 	bool ShouldAllowMultiWorldTabCompletion(void) const { return m_ShouldAllowMultiWorldTabCompletion; }
 
+	/** Returns the channel manager. */
+	cChannelManager * GetChannelManager();  // tolua_export
```
tigerw: Why no reference?

src/Bindings/ManualBindings.cpp
madmaxoft: Unfortunately, tolua doesn't do callbacks.

src/ChannelManager.cpp
madmaxoft: Also, avoid names without an underscore. They are hard to read:
`you_don_t really_know_the_type_and_name`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -437,6 +437,22 @@ return
...
+
+			Functions =
+			{
+				AddClientToChannel = { Params = "{{cClientHandle}}, Channel Name", Notes = "Adds and registers a client on the specified channel." },
+				RemoveClientFromChannel = { Params = "{{cClientHandle}}, Channel Name", Notes = "Removes and unregisters a client on the specified channel." },
+				RegisterChannel = { Params = "Channel Name, Callback", Return = "bool", Notes = "Registers a handler for a specific channel. Returns true if it was able to be registered." },
+				RemoveChannel = { Params = "Channel Name", Return = "bool", Notes = "Removes the handler for the specified channel. Returns true if the handler was removed." },
+				BroadcastChannelMessage = { Params = "Channel Name, Message, {{cWorld|World}}", Notes = "Sends the message on the specified channel. Limits the message to the specified world, if provided." }
+			}
+		}, -- cChannelManager
```
madmaxoft: I find the API documentation rather lacking. The function descriptions don't tell much about what they are used for. I'd appreciate an explanation (why do I need to add a client to a channel?) and, best of all, a piece of an example code.

src/Bindings/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ SET (SRCS
 
 SET (HDRS
 	Bindings.h
+	../ChannelManager.h
```
madmaxoft: Okay, this should not be here, the file belongs to the project in CMakeLists.txt up one level. However, it needs to be listed in `BINDING_DEPENDENCIES`

madmaxoft:
So, you've written a big chunk of API code, and there's no test code for it? How do you know the bindings work?

You should add at least a test in the Debuggers plugin, so that:
1. you test your bindings
2. we know how to use the API
3. you have a working example for the API docs.

src/ChannelManager.cpp
SamJBarney: When you say 'they', what are you referring to? The protocol? The client?

src/ChannelManager.cpp
SamJBarney: I don't see why not. :)

src/ChannelManager.cpp
SamJBarney: Well, the majority of this code was written to be a halfway point. Right now it just merely wraps code in cClientHandle. However, someone was working on moving cClientHandles to being managed by the server, and I wanted to move all channel related code and data into cChannelManager. I don't know if that move has been accomplished or not, so I wanted to leave it at this halfway point until I know for certain.

src/ChannelManager.cpp
SamJBarney: Thanks for the advice. :)

src/ChannelManager.cpp
SamJBarney: Have the formatting guides changed again? What are the basic standards now?

src/ChannelManager.cpp
SamJBarney: Yeah, I need to fix it

src/Bindings/CMakeLists.txt
SamJBarney: Thank you!

lib/tolua++
SamJBarney: Crap. I missed that. Gotta remove this.

src/ChannelManager.cpp
SamJBarney: I probably could get away with one.

src/ChannelManager.h
SamJBarney: Not really. Artifacts I need to get rid of

src/ChannelManager.cpp
madmaxoft: Actually, I think the hook should be called each and every time a plugin message is received, regardless of whether it has a callback or not.

src/ChannelManager.cpp
madmaxoft: The rules haven't changed, we've always used CamelCaseNames.

src/Server.h
```diff
@@ -241,6 +245,8 @@ class cServer
 	Initialized in InitServer(), used in Start(). */
 	AStringVector m_Ports;
 
+	cChannelManager m_ChannelManager;
+
```
madmaxoft: I think it might be better to make it a `UniquePtr<cChannelManager>`, that way you don't need to `#include` the class declaration file here (include it in the cpp instead) and thus it reduces compile-time dependencies.

src/ChannelCallback.h
madmaxoft: You cannot make this a lambda, because it has multiple functions (`Call` and `BelongsTo`); a lambda could wrap the callback, but wouldn't be capable of saying whether it belongs to a certain plugin or not.

However, I do believe that this class could be made internal to the `cChannelManager` class, and moved in its source and header files.

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+#include "ChannelManager.h"
+#include "ChannelCallback.h"
+#include "ClientHandle.h"
+#include "Entities/Player.h"
+#include "OSSupport/CriticalSection.h"
+#include "World.h"
+
+
+class cChannelBroadcastCallback : public cPlayerListCallback
```
madmaxoft: How about moving this entire class into the `cChannelManager` class? `cChannelManager::cBroadcastCallback`.

src/ChannelManager.cpp
madmaxoft: Not sure, what with changing the underlying container while iterating over it.
I suppose it could be done with `std::remove_if`, but that tends to be complicated. I'd leave this as is.

src/ChannelManager.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+void cChannelManager::HandlePluginUnloading(const cPluginLua * a_Plugin)
+{
+	cCSLock Lock(m_CSCallbacks);
+
+	for (auto it = m_ChannelCallbacks.begin(); it != m_ChannelCallbacks.end(); ++it)
+	{
+		if (it->second->BelongsTo(a_Plugin))
+		{
+			m_ChannelCallbacks.erase(it);
```
madmaxoft: When you erase while iterating, you need to reassign to `it` and not increment it:
```cpp
if (condition)
{
   itr = container.erase(itr);
}
else
{
  ++itr;
}
```

src/ChannelManager.cpp
madmaxoft: Even better, since this is used only in a single function inside the cpp, move the declaration to that function.

src/ChannelManager.cpp
tigerw: I remember some sources suggesting that unordered_map would not invalidate any iterators bar the deleted one when erasing.

src/ChannelManager.cpp
tigerw: Okay.

> However, someone was working on moving cClientHandles to being managed by the server
> 
Sounds like #3115 :D

src/ChannelManager.cpp
madmaxoft: Which is exactly the problem - the `erase` invalidates `it` and then the next for loop iteration tries to increment that invalid iterator.

tigerw: Makes sense.

I hope I didn't leave some crashworthy code in the server; I remember looking this up.

src/ChannelManager.cpp
SamJBarney: @madmaxoft may I ask why you believe this?

src/ChannelManager.cpp
SamJBarney: Yes, I think it is 👍

src/ChannelManager.cpp
LogicParrot: >crashworthy

Server/Plugins/APIDump/APIDesc.lua
SamJBarney: I'm getting there :)

src/ChannelManager.cpp
madmaxoft: All of our hooks are called for data before the data is default-processed, and provides a means for the plugin to completely block the action that would normally be performed. So one could consider the hooks as the means of injecting deep inspection code.

With the changes in this PR, it could be said that the hook is pretty much meaningless for regular plugins, and could only be useful for specialties - which have a high probability of wanting to intercept all plugin messages, not only those unregistered ones. For example a plugin that would collect statistics about all incoming plugin messages would be impossible without a global hook.

SamJBarney:
I have some local testing, but you are right. I'll have some tests written soon

src/ChannelManager.cpp
tigerw: Wasn't the point of this not having to, for each plugin message, call into Lua for every plugin? If the hook is going to always be called, again, what's the benefit of adding callbacks?

madmaxoft: No, the point was to simplify plugins' workflow with plugin messages.

src/Bindings/ManualBindings.cpp
tigerw: So what do you get when say, you pass a function to a method, a la

```lua
RegisterChannel(
    Channel,
    function()
        -- Potatoes
    end
)

src/ChannelManager.cpp
tigerw: > @tigerw One of my previous comments touches on the subject. Specifically it is potentially one lookup and call instead of always iterating over and calling every hook callback.

?

Pokechu22:
Sorry for the response delay.

My point was that the Bukkit API imposed no restrictions on channel registration and unregistration.  And the bukkit version can be considered canonical since dinnerbone wrote both it and the vanilla implementation.

An example of one of those weird channels is the [channel (singular) used by bungeecord]( http://www.spigotmc.org/wiki/bukkit-bungee-plugin-messaging-channel/).  It's something that _could_ be extracted to an API, but for many uses is trivial enough not to be.  Except it sends responses via the same channel.  Having all plugins listen on the global hook for it would be fragile and would break if only 1 plugin registered it the other way.  And unless there's an existing API for it, plugins won't try to use one since in most cases it would be excessive.

The key point of that bungeecord anecdote is _not_ "We need an API for bungeecord".  It's that there are odd channels (plural) that don't fit your assumptions.  Trying to make an API for those assumptions will only make for a frustrating, potentially impossible, API when an unusual channel is _inevitably_ needed.

SamJBarney:
See, that's the funny thing, the link you posted talks about something that I have mulled over and over. When a plugin message comes in the string that represents the channel and subchannel is formatted like this: `<Channel>|<Subchannel>`. You can see an example of this formatting in the `cProtocol180::HandleVanillaPluginMessage` method.

From what I can tell, when you are referring to reusing the same channel, you are referring to subchannels and such.
In Cuberite, a channel and subchannel are treated as a single channel. It is semantically the same, but much easier to implement.

Pokechu22:
> When a plugin message comes in the string that represents the channel and subchannel is formatted like this: `<Channel>|<Subchannel>`. You can see an example of this formatting in the `cProtocol180::HandleVanillaPluginMessage` method.

> From what I can tell, when you are referring to reusing the same channel, you are referring to subchannels and such.
In Cuberite, a channel and subchannel are treated as a single channel. It is semantically the same, but much easier to implement.

I'm referring to something different.  What you seem to think I'm saying is that (for instance) `MC` is a main channel  and `MC|AdvCmd`, `MC|Brand`, `MC|DebugPath`… as subchannels.  That isn't what I'm arguing, and I agree that those should not be treated specially – each is a separate channel.  (And I agree – the `|` is just semantics.  Not all messages use it, though many do.)

The thing is, `BungeeCord` _does_ have subchannels.  But, unlike the semantics-only `|` delineated subchannels, `BungeeCord`'s subchannels are in the payload.  That's what makes it more complex.  (For instance, if I remember how `DataOutputStream` structures itself, a `ConnectOther` packet would be on the `BungeeCord` channel with the payload being `\x07ConnectOther\x03Rob\x0dCreativePlots`.). See how that's more complex?

SamJBarney:
I see that, but I don't see the need for writing code for the occasional exception when plugin authors can take care of that on the lua side of things.

Pokechu22:
I'm  not saying that you need to handle subchannels.  I'm saying that for subchannels to be something plugins can handle themselves, multiple plugins need to be able to listen on the same channel.

SamJBarney:
Update: Still working on the example. Having to figure out Forge's extra packets, when they need to be sent, and what they do. 

Documentation updates will be up later this week.

Issue #910, opened 2014-04-18
All entities should be constructed of components


SamJBarney:
I know this is an idea that I have talked about on the forum, but I think I should move it here just temporarily so that it can be looked at without getting cluttered down.

Another reason is that before I would get started on this, I would ask that ALL changes being made to any type of entity be put on hold until the rewrite is completed to avoid merge conflicts. That is what happened last time that I started on it, and I do not want it to happen again.
## Proposal

ALL entities should be constructed from discrete components which define the entity as a whole. These components should be separated into the following categories:
- Movement - Defines how the entity moves, if at all
- Attack - Defines how the entity attacks targets
- Physical - Handles collision detection
- AI - Handles how the entity reacts to being interacted with, along with how the AI acts in general.
- Environmental - Handles how the entity reacts to the environment, such as items being destroyed when they hit a cactus.
- System - Handles packet creation for letting the client know that the entity was spawned. Ex: Mob spawning is different than exp spawning

Along with that, each type of component provides a Lua interface to allow for custom components to be defined within Lua plugins.

Each vanilla mob will have default components defined in the settings file for MCServer that can be adjusted.

This change should allow for a greater flexibility for MCServer without compromising speed, and allow us to get a good, solid AI in place that is extensible.

I really would like to get this done soon. Please let me know if my conditions at the top of this are acceptable.

SphinxC0re:
Any updates on this one?

SafwatHalaby:
Is this still relevant with the new AI?

SphinxC0re:
It might be

worktycho:
I think its something we should do, but not as a complete rewrite. Just refactors here and there, when working on the methods for other reasons.

SafwatHalaby:
Yes, a complete rewrite feels like overkill to me.

PR #3295, opened 2016-08-05
Refactor mob code into modular "behaviors"

SafwatHalaby:
This work in progress is an attempt to make the mob code much more readable and modular. It should allow rapid development of the missing mobs, among other things.

After trying out several different approaches, I believe I've found the best way to refactor the mob code, but I want your opinion before I refactor everything.

This PR is a fully working "sample" of my plan, I only refactored `cPassiveMonster` and some of `cMonster` for now. The only regression is that idle agressive mobs do not wander around. 

I will now explain the idea behind "behaviors" and what problems this solves.
### Problems in the current mob code
- Non modular and complex to follow, trivial changes often break things, the mobs' decision logic is spread across classes and  isn't easy to deduce.
- "Updrift" of code: When two mob types which aren't related need a shared feature, the shared code is put in their first shared ancestor, making classes up the hierarcy tree fat and introducing the shared methods and variables to derived classes that never use them. In other cases, the code is simply left out of some derived classes. For instance, this is why we don't have wolf breeding.
- The "state system" is outdated and practically overriden. All mobs have 3 "official" states (Chasing, Idle, Escaping), but in reality some mobs never need some of these states, and some mobs have many other states (following player, making love, sitting, etc). The actual state lies in various boolean flags, timers, and entity pointers that trump the "official" state system, and these aren't grouped together systematically.
### behaviors solution (In Concept)

The code is divided into classes representing atomic units of behavior. Each monster is a mashup of various behaviors.

Examples of behaviors:
 Agression toward players
 Attraction to certain held items
 Fearing daylight
 Fearing cats
 Protection of players
 Wandering around
 Cowardness (running away when hit)
 Following a player
 Protecting a player

Splitting the code into such behaviourally related classes obviously makes readability and maintainability much easier by grouping behaviors together in self contained modules. It also provides very powerful "plug-and-play" system; I can easily make a chicken defend an owner or make cows agressive simply by adding or removing the relevant behaviors. The behaviors are mobtype agnostic.
### behaviors (Code-specific)

In order to use a behavior, the monster must:
1. Have a member variable of the desired behavior, cbehaviorX
2. Make sure to call the proper behavior methods. They typically involve "forwarding"
   methods to the behavior, e.g. calling "cbehaviorX::Tick from the monster's Tick".
   The methods that should be forwarded vary by behavior and are documented in a behavior's header file.
3. Some behaviors expect their host monster to have specific methods. In such rare casess, the monster
   must also override method cMonster::someMethod. For instance, `cBehaviorBreeder` expects the host
   to implement a method which dictates a child's attributes based on two parents.  *(Edit: Removed interfaces for simplicity). *

**Active ticks:**
In a monster tick, the monster sequentially executes its behaviors' `ActiveTick()` methods for every behavior that has one, stopping at the first `ActiveTick()` which returns true. Returning true means the behavior has decided to control the monster. This means that each tick, exactly one behavior controls the monster.
This also means that the ordering of the "ActiveTick()" calls determines
priorities. We don't want our "wander" behavior to control us if we're a zombie at sunlight. We'd rather be controlled by the anti-burning behavior and seek shade. Similarly, chasing players is higher priority than wandering but lower priority than seeking shade. 

**Passive ticks:**
In a monster tick, the monster should call `Tick()` for every behavior that has one. (After the ActiveTick)
This is a "passive tick " which should not control the mob's actions. It is only used for bookkeeping, e.g. updating a behavior's timers and cooldowns. **Edit:** Or for actions that need to be done regardless of the current state (e.g. chicken laying eggs).

Below are two priority examples:

Example 1: For a zombie, we could execute ActiveTick
in the following order:
 cbehaviorBurnsAtDaylight.ActiveTick()
 cbehaviorAgressive::ActiveTick(),
 cbehaviorWanderer::ActiveTick()

This makes the mob wander around if there's neither daylight burns nor players, and attack players
only if there's no daylight burns.

Example 2: Wolf
 cbehaviorProtector::ActiveTick()     // priority 1 is owner protection. Always returns false if ownerless.
 cbehaviorItemFollower::ActiveTick()  // following food.
 cbehaviorTamed()::ActiveTick()       // Following owner. This always returns true unless we are ownerless.
 cbehaviorWanderer()::ActiveTick()    // Wander around only if ownerless.

It isn't mandatory for a behavior to have a Tick nor an ActiveTick.
### In short

behaviors rock.
Opinions please! I don't want to go further without a green light.

### DIFF
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -12,10 +12,11 @@
 
 
 
-cAggressiveMonster::cAggressiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height) :
-	super(a_ConfigName, a_MobType, a_SoundHurt, a_SoundDeath, a_Width, a_Height)
+cAggressiveMonster::cAggressiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height, int a_AggressionLightLevel) :
+	super(a_ConfigName, a_MobType, a_SoundHurt, a_SoundDeath, a_Width, a_Height), m_BehaviorAggressive(this, a_AggressionLightLevel)
 {
 	m_EMPersonality = AGGRESSIVE;
+	ASSERT(GetBehaviorChaser() != nullptr);
 }
 
 
@@ -27,14 +28,17 @@ void cAggressiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	super::Tick(a_Dt, a_Chunk);
 
+	cBehaviorChaser * BehaviorChaser = GetBehaviorChaser();
+	cBehaviorWanderer * BehaviorWanderer = GetBehaviorWanderer();
+
 	for (;;)
 	{
-		GetBehaviorAggressive()->ActiveTick();
-		if (GetBehaviorChaser()->ActiveTick())
+		m_BehaviorAggressive.ActiveTick();
+		if (BehaviorChaser->ActiveTick())
 		{
 			break;
 		}
-		if (GetBehaviorWanderer()->ActiveTick())
+		if ((BehaviorWanderer != nullptr) && BehaviorWanderer->ActiveTick(a_Dt, a_Chunk))
 		{
 			break;
 		}
@@ -43,5 +47,5 @@ void cAggressiveMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		break;
 	}
 
-
+	BehaviorChaser->Tick();
 }
--- a/src/Mobs/AggressiveMonster.h
+++ b/src/Mobs/AggressiveMonster.h
@@ -1,10 +1,10 @@
 #pragma once
 
 #include "Monster.h"
+#include "Behaviors/BehaviorAggressive.h"
 
 
-
-
+typedef std::string AString;
 
 class cAggressiveMonster :
 	public cMonster
@@ -13,11 +13,10 @@ class cAggressiveMonster :
 
 public:
 
-	cAggressiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height);
+	cAggressiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height, int a_AggressionLightLevel);
 
 	virtual void Tick          (std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
+
+private:
+	cBehaviorAggressive m_BehaviorAggressive;
 } ;
-
-
-
-
--- a/src/Mobs/Bat.cpp
+++ b/src/Mobs/Bat.cpp
@@ -6,8 +6,7 @@
 #include "../Chunk.h"
 
 
-cBat::cBat(void) :
-	super("Bat", mtBat,  "mob.bat.hurt", "mob.bat.death", 0.5, 0.9)
+cBat::cBat(void) : super("Bat", mtBat,  "mob.bat.hurt", "mob.bat.death", 0.5, 0.9)
 {
 	SetGravity(-2.0f);
 	SetAirDrag(0.05f);
--- a/src/Mobs/Bat.h
+++ b/src/Mobs/Bat.h
@@ -7,10 +7,8 @@
 
 
 
-class cBat :
-	public cPassiveMonster
+class cBat : public cMonster
 {
-	typedef cPassiveMonster super;
 
 public:
 	cBat(void);
--- a/src/Mobs/Behaviors/BehaviorBreeder.cpp
+++ b/src/Mobs/Behaviors/BehaviorBreeder.cpp
@@ -18,12 +18,13 @@ iBehaviorBreeder::~iBehaviorBreeder()
 
 
 
-cBehaviorBreeder::cBehaviorBreeder(iBehaviorBreeder * a_ParentInterface) :
-	m_ParentInterface(a_ParentInterface),
+cBehaviorBreeder::cBehaviorBreeder(cMonster * a_Parent, cItems & a_BreedingItems) :
+	m_Parent(a_Parent),
 	m_LovePartner(nullptr),
 	m_LoveTimer(0),
 	m_LoveCooldown(0),
-	m_MatingTimer(0)
+	m_MatingTimer(0),
+	m_BreedingItems(a_BreedingItems)
 {
 	m_Parent = dynamic_cast<cMonster *>(m_ParentInterface);
 	ASSERT(m_Parent != nullptr);
@@ -67,16 +68,13 @@ bool cBehaviorBreeder::ActiveTick()
 			m_Parent->GetWorld()->DoWithEntityByID(BabyID, Callback);
 			if (Callback.Baby != nullptr)
 			{
-				auto BabyInterface = ToInterface(Callback.Baby);
-				BabyInterface->InheritFromParents(m_Parent, m_LovePartner);
+				Callback.Baby->InheritFromParents(m_Parent, m_LovePartner);
 			}
 
 			cFastRandom Random;
 			World->SpawnExperienceOrb(Pos.x, Pos.y, Pos.z, 1 + Random.NextInt(6));
 
-			auto PartnerInterface =  ToInterface(m_LovePartner);
-			cBehaviorBreeder & PartnerBreedingBehavior = PartnerInterface->GetBehaviorBreeder();
-			PartnerBreedingBehavior.ResetLoveMode();
+			m_LovePartner->GetBehaviorBreeder()->ResetLoveMode();
 			ResetLoveMode();
 		}
 		return true;
@@ -111,8 +109,8 @@ bool cBehaviorBreeder::ActiveTick()
 					return false;
 				}
 
-				auto PartnerBreedingBehavior = ToBehavior(PotentialPartner);
-				auto MyBreedingBehavior = ToBehavior(m_Me);
+				auto PartnerBreedingBehavior = PotentialPartner->GetBehaviorBreeder();
+				auto MyBreedingBehavior = m_Me->GetBehaviorBreeder();
 
 				// If the potential partner is not in love
 				// Or they already have a mate, do not breed with them
@@ -165,10 +163,9 @@ void cBehaviorBreeder::Tick()
 
 void cBehaviorBreeder::Destroyed()
 {
-	UNUSED(m_ParentInterface);
 	if (m_LovePartner != nullptr)
 	{
-		ToBehavior(m_LovePartner)->ResetLoveMode();
+		m_LovePartner->GetBehaviorBreeder()->ResetLoveMode();
 	}
 }
 
@@ -181,10 +178,8 @@ void cBehaviorBreeder::OnRightClicked(cPlayer & a_Player)
 	// If a player holding breeding items right-clicked me, go into love mode
 	if ((m_LoveCooldown == 0) && !IsInLove() && !m_Parent->IsBaby())
 	{
-		cItems BreedingItems;
-		m_ParentInterface->GetBreedingItems(BreedingItems);
 		short HeldItem = a_Player.GetEquippedItem().m_ItemType;
-		if (BreedingItems.ContainsType(HeldItem))
+		if (m_BreedingItems.ContainsType(HeldItem))
 		{
 			if (!a_Player.IsGameModeCreative())
 			{
@@ -235,24 +230,4 @@ bool cBehaviorBreeder::IsInLove() const
 bool cBehaviorBreeder::IsInLoveCooldown() const
 {
 	return (m_LoveCooldown > 0);
-}
-
-
-
-
-
-iBehaviorBreeder * cBehaviorBreeder::ToInterface(cMonster * a_Monster)
-{
-	auto ptr = dynamic_cast<iBehaviorBreeder *>(a_Monster);
-	ASSERT(ptr != nullptr);
-	return ptr;
-}
-
-
-
-
-
-cBehaviorBreeder * cBehaviorBreeder::ToBehavior(cMonster * a_Monster)
-{
-	return &ToInterface(a_Monster)->GetBehaviorBreeder();
 }
--- a/src/Mobs/Behaviors/BehaviorBreeder.h
+++ b/src/Mobs/Behaviors/BehaviorBreeder.h
@@ -3,7 +3,6 @@
 // Grants breeding capabilities to the mob
 
 class cBehaviorBreeder;
-class iBehaviorBreeder;
 
 class cWorld;
 class cMonster;
@@ -18,7 +17,7 @@ class cBehaviorBreeder
 {
 
 public:
-	cBehaviorBreeder(iBehaviorBreeder * a_ParentInterface);
+	cBehaviorBreeder(cMonster * a_Parent, cItems & a_BreedingItems);
 
 	// Functions our host Monster should invoke:
 	void Tick();
@@ -42,11 +41,7 @@ public:
 	bool IsInLoveCooldown() const;
 
 private:
-	static iBehaviorBreeder * ToInterface(cMonster * a_Monster);
-	static cBehaviorBreeder * ToBehavior(cMonster * a_Monster);
-
 	/** Our parent */
-	iBehaviorBreeder * m_ParentInterface;
 	cMonster * m_Parent;
 
 	/** The monster's breeding partner. */
@@ -60,4 +55,6 @@ private:
 
 	/** The monster is engaged in mating, once this reaches zero, a baby will be born. Decrements by 1 per tick till reaching zero, then a baby is made and ResetLoveMode() is called. */
 	int m_MatingTimer;
+
+	cItems & m_BreedingItems;
 };
--- a/src/Mobs/Behaviors/BehaviorChaser.h
+++ b/src/Mobs/Behaviors/BehaviorChaser.h
@@ -19,6 +19,7 @@ public:
 	// Functions our host Monster should invoke:
 	bool ActiveTick();
 	void Destroyed();
+	void Tick();
 
 	// Our host monster will call these once it loads its config file
 	void SetAttackRate(float a_AttackRate);
--- a/src/Mobs/PassiveMonster.cpp
+++ b/src/Mobs/PassiveMonster.cpp
@@ -9,11 +9,20 @@
 
 
 
-cPassiveMonster::cPassiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height) :
+cPassiveMonster::cPassiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height, cItems & a_BreedingItems, cItems & a_FollowedItems) :
 	super(a_ConfigName, a_MobType, a_SoundHurt, a_SoundDeath, a_Width, a_Height),
-	m_BehaviorBreeder(this), m_BehaviorItemFollower(this), m_BehaviorCoward(this)
+	m_BehaviorBreeder(this), m_BehaviorItemFollower(this, a_FollowedItems), m_BehaviorCoward(this)
 {
 	m_EMPersonality = PASSIVE;
+}
+
+
+
+
+
+cPassiveMonster::~cPassiveMonster()
+{
+
 }
 
 
--- a/src/Mobs/PassiveMonster.h
+++ b/src/Mobs/PassiveMonster.h
@@ -7,14 +7,14 @@
 #include "Behaviors/BehaviorCoward.h"
 
 
-
+typedef std::string AString;
 class cPassiveMonster : public cMonster
 {
 	typedef cMonster super;
 
 public:
-	cPassiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height);
-
+	cPassiveMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, double a_Width, double a_Height, cItems & a_BreedingItems, cItems & a_FollowedItems);
+	virtual ~cPassiveMonster();
 	virtual void Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
 	virtual void OnRightClicked(cPlayer & a_Player) override;
 
@@ -22,17 +22,6 @@ public:
 	virtual bool DoTakeDamage(TakeDamageInfo & a_TDI) override;
 
 	virtual void Destroyed(void) override;
-
-	// Implementing the iBehaviorBreeder interface
-	/** Returns the items that make the animal breed - this is usually the same as the ones that make the animal follow, but not necessarily. */
-	virtual void GetBreedingItems(cItems & a_Items) const override { GetFollowedItems(a_Items); }
-
-	virtual void InheritFromParents(cMonster * a_Parent1, cMonster * a_Parent2) override { }
-
-	// Implementing the iBehaviorItemFollower interface
-
-	/** Returns the items that the animal of this class follows when a player holds it in hand. */
-	virtual void GetFollowedItems(cItems & a_Items) const override { }
 
 private:
 	cBehaviorBreeder m_BehaviorBreeder;


LogicParrot:
Naming terminology change proposals are welcome. Particularly:
- `ActiveTick`
- `TraitBreeder`

sphinxc0re:
Sounds like  a job for Rust :)
https://doc.rust-lang.org/book/traits.html

LogicParrot:
These aren't like rust traits. I just picked that name (other proposals welcome). They're a bit like mixins, but are mob specific.

madmaxoft:
I don't think the `traits` should be hard-coded for each monster type, rather, they seem like `behaviors` that could all simply implement a base interface and through that interface be stored in each mob as an array (which specifies their priority). We could then drop the specific mob classes altogether and have a single class, cMob, that would be fully customizable to represent any mob in the game. Trading creepers! :)

This has been discussed before on the forum, but I can't find the link right now. If I remember correctly, @worktycho was quite involved.

LogicParrot:
>I don't think the traits should be hard-coded for each monster type, rather, they seem like behaviors that could all simply implement a base interface and through that interface be stored in each mob as an array (which specifies their priority). We could then drop the specific mob classes altogether and have a single class, cMob, that would be fully customizable to represent any mob in the game. Trading creepers! :)

I was considering implementing a special array-based class called "CustomMob" which has an array of traits, unlike the hardcoded classes, and expose that via LUA for the plugin makers' pleasures.

As for a pure array solution. Let's see... Some initial thoughts:
- Could it harm performance or memory usage?
- Is it worth the benefits? What benefits does it provide? What are the drawbacks?
- Note that we'd still have to "hard code" how the arrays are filled for each builtin mobtype.
- The overall code may be shorter.

LogicParrot:
I'll adopt the `behavior` term and change the code accordingly.

LogicParrot:
CI is complaining about some undocumented API symbols in a class that I didn't modify.
```
cFloater.CanPickup
cFloater.GetAttachedMobID
cFloater.GetOwnerID
```

LogicParrot:
Can someone shed some light on the CI error? The 3 symbols are in fact documented.

LogicParrot:
How should I deal with behaviors that currently only 1 mobtype uses?
For instance, egg drops are only chicken specific. Options:
 - Keep the behavior in cChicken
 - Create a `behaviorItemDropper` 

The latter allows more customization in the future for modders, (a gold-laying cow?)

LogicParrot:
The array method is superior for customization because Lua plugins can modify the default mobs' arrays and mobs' behavior can be modified after they spawn. That sounds very attractive. I'll take a step back and think about this. I'd love to see the aforementioned forum post and perhaps this discussion should be moved to the forum. 

The work done till now is not wasted, I believe I can easily modify the current behaviors to fit the new idea.

tigerw:
Work was done on this in the ComponentRewrite branch, what do you think of it?

src/Entities/Entity.cpp
```diff
@@ -222,8 +222,15 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
 	}
 
 	cChunk * ParentChunk = GetParentChunk();
-	m_World->QueueTask([this, ParentChunk](cWorld & a_World)
+
+	// We remove the entity from memory 5 ticks later, giving everyone plenty of time
```
tigerw: This sounds off. Are you talking about threads? Why is it not possible for "someone" to actually need > 5 seconds to read IsTicking?

tigerw:
Relevant issue #910.

src/Entities/Entity.cpp
LogicParrot: I wanted to ask you about this.

Previously, we had code like `If (m_Target->IsDestroyed()) m_Target = nullptr`, but that was troublesome because sometimes m_Target is already freed from memory and basically calling this is UB and there were occasional crashes.

Back then, my solution was the complex "cPawn::StopEveryoneFromTargetingMe" and several other functions that nullify m_Target when needed. But looking back it is seemingly overkill, if we just postpone freeing the entity for 2 or more ticks (5 was just arbitrary), that everyone gets a chance to tick and check `IsDestroyed` (now replaced with IsTicking) and nullify the pointers and the problem is solved, is it not?

Unless of course there was another reason for my complex solution, but I do not remember any.

LogicParrot:
On a first glance, #910 seems to aim to split things by domain (movement, physics, etc), not by behavior, so these seem like different approaches.

LogicParrot:
@madmaxoft A single class approach seems not viable; some things are specific to exactly one mob type, like sheep fur color, and stuffing those in behaviors seems unclean, I think we still need cSheep for sheep-only things to keep it tidy.

However, we can completely flatten the tree; No need for cPassiveMonster, cAgressiveMonster, etc. These are getting hollowed by my changes anyways. So, we could have a class for each mobtype, housing the mob specifics, and all those classes inherit cMob, which has the behavior array logic.

madmaxoft:
The "visuals" could be classes specific for the mob (visual) type. A sheep would have a `cVisualSheep` that would store the wool color, the wolf would have a `cVisualWolf` storing the collar color, a horse would have a `cVisualHorse` storing the kind of the horse etc. The visual classes would only provide information about their type (what mob they map to in-game, such as sheep or horse), all the rest will be directly accessed by code like `reinterpret_cast<cVisualSheep *>(m_Visual)`.

Pokechu22:
> For instance, egg drops are only chicken specific. Options:

> * Keep the behavior in cChicken
> * Create a behaviorItemDropper

I haven't actually looked at the implementation yet, so this may not be an issue, but from my understanding, wouldn't this not work?  Your other examples are all pathfinding related, but here it's something you'd always run.  Chickens lay eggs regardless as to whether they're wandering, following seeds, panicking, or anything else; if the behavior is only called if prior ones don't run, then it seems like it wouldn't work.  

I think it would be good to have item dropping as a behavior, though, assuming it would work.

LogicParrot:
@Pokechu22 Hence the passive and active ticks. Passive ticks run regardless of the current "state", the egg laying behavior would only implement passive ticks. Sorry if the naming is bad.

LogicParrot:
@madmaxoft It's not just about visuals. Consider breeding: There's a breeding function which is sheep-specific -it determines the newborn's color based on parents. - and a breeding function which is horse-specific. How would that work in the case you described?

In my case, which is already implemented in the PR, it would work like this:  `cBehaviorBreeder` dynamic casts the newborn `cMob` into an `iBehaviorBreeder`, and calls `iBehaviorBreeder::InheritFromParents(parent1, parent2)`. cSheep and cHorse both implement `iBehaviorBreeder` (and both inherit cMob), so the proper function is called.

LogicParrot:
So, they contain "body specific" things, which are mostly visuals, but also methods, and some of those methods require polymorphism.

madmaxoft:
Oh well, do the behaviors for now, we'll see how the rest will go afterwards.

Make a separate behavior even for those very specific actions, such as dropping an egg (`cItems` in general, through a constructor param)

src/Entities/Entity.cpp
```diff
@@ -222,8 +222,15 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
...
+	// to read IsTicking(formerly IsDestroyed), finding out it's false, and nulling out their pointers.
+	// LogicParrot TODO Clear out the previous much more complicated mechanism
+	// (See: cPawn::StopEveryoneFromTargetingMe() and its friends)
+	// Take world travel into account.
+	m_World->ScheduleTask(40, [this, ParentChunk](cWorld & a_World)
```
NiLSPACE: Where does 40 ticks come from?

LogicParrot: Yes. magic, but 2 should be enough. Here's the deal (Also see my response to @tigerw )

`cMonster::m_Target` is an entity pointer, and it used to cause trouble when mobs are freed from memory or teleported to new worlds because other entities still pointed at them. The solution was some relatively complex set of functions that nullified m_Target when needed, (cPawn::stopEveryoneFromTargetingMe, etc.). In this PR, there are new entity pointers like `m_Attacker`  because the code is split into modular behaviors so the old technique doesn't cover them - and seems needlesly complex anyways -. The new technique is much simpler: Do not free entities from memory immediately when they are destroyed, just wait a bit. During the wait, everyone holding a pointer to that entity would call `If (!myPointer.IsTicking()) myPointer = nullptr`. 2 ticks should be enough, assuming everyone makes this check once per tick. (1 tick isn't enough because if A is freed prior to B and B makes the check, then B just used a dangling pointer).

LogicParrot:
>Oh well, do the behaviors for now, we'll see how the rest will go afterwards.

Full steam ahead! 
I plan to refactor existing mobs first, then give wolf breeding and endermen a shot. The ease - or lack thereof - of doing that should be a rough indicator of how good this system is.

SamJBarney:
@LogicParrot Good luck with this! I started working on something similar in the ComponentRewrite branch, so if you need any help, feel free to ask :)

LogicParrot:
Thanks :)

LogicParrot:
I decided to remove those "behavior interfaces" like `iBehaviorBreeder`, because in most cases I can get things done without them. There are only a handful of methods that actually need this in all behaviors combined, so I made those few methods virtual members of cMonster.

LogicParrot:
Today I realized that an attack-capable mob is best implemented with two behaviors: `Striking` and `chasing`. Examples:

**Zombie:**
Chase behavior: Simple target approach (cBehaviorChaserSimple)
Attack behavior: Simple melee attack (cBehaviorStrikerSimple)

**Skeleton:**
Chase behavior: Simple target approach (cBehaviorChaserSimple)
Attack behavior: Arrow shooting (cBehaviorStrikerArcher)

**Enderman:**
Chase behavior: Jumpy target approach (cBehaviorChaserTeleporting)
Attack behavior: Simple melee attack (cBehaviorStrikerSimple)

**Blaze:**
Chase behavior: None
Attack behavior: Fireball shooting (cBehaviorStrikerFireball)

*Perhaps `cBehaviorStrikerFireball` and `cBehaviorStrikerArcher` can actually be the same behavior, `cBehaviorStrikerShooter`, which determines what to shoot based on initialization parameters. I'll decide as I code.*

Pokechu22:
> Perhaps `cBehaviorStrikerFireball` and `cBehaviorStrikerArcher` can actually be the same behavior,  `cBehaviorStrikerShooter`, which determines what to shoot based on initialization parameters. I'll decide as I code.

Same thing could probably be done for witch potions (but that also would need custom logic since they throw multiple types of potions).

src/Mobs/Behaviors/BehaviorCoward.cpp
```diff
@@ -17,7 +17,7 @@ cTraitCoward::cTraitCoward(cMonster * a_Parent) :
 
 
 
-bool cTraitCoward::ActiveTick()
+bool cBehaviorCoward::ActiveTick()
 {
 	if (m_Attacker == nullptr)
 	{
```
firnsan: @LogicParrot 
If a chicken is attack,   it's  `MoveToPosition` will be call, and  `pathfinder`  will be trigger  to move to the destination.  
(if i am wrong, please notice me that)
But  how does `pathfinder`  know the chicken need to run , or fly (jump) or not, and how fast the chicken runs to the destination?

firnsan:
I encounter a core dump in mobs branch:

```
[ced05142baadc2bc|23:54:35] from commit id: approx: 2df8f8cca61bf2583b2a474eaa7a9ccb7bc90eac built at: approx: 2016-08-06 11:12:41 +0300
......
......
[b1ebcc61d0a5bc83|23:49:01] Spawning cSpider #598 at {251, 43, 57}
[b1ebcc61d0a5bc83|23:49:02] Destroying entity #237 (cPickup) from chunk (25, 0)
[b1ebcc61d0a5bc83|23:49:02] Destroying entity #238 (cPickup) from chunk (25, 0)
[b1ebcc61d0a5bc83|23:49:03] Spawning cSpider #599 at {274, 33, -158}
[b1ebcc61d0a5bc83|23:49:03] Spawning cSlime #600 at {400, 23, 15}
[b1ebcc61d0a5bc83|23:49:03] Spawning cSlime #601 at {399, 25, 78}
[b1ebcc61d0a5bc83|23:49:05] Spawning cSpider #605 at {240, 43, 85}
[b1ebcc61d0a5bc83|23:49:05] Spawning cSpider #606 at {253, 43, 69}
[b1ebcc61d0a5bc83|23:49:05] Spawning cSpider #607 at {244, 43, 96}
[b1ebcc61d0a5bc83|23:49:07] Destroying entity #607 (cSpider) from chunk (4, -18)
[b1ebcc61d0a5bc83|23:49:07] Assertion failed: a_Entity->GetParentChunk() == this, file /data1/minecraft_src/LogicParrot/cuberite/src/Chunk.cpp, line 1958
./Cuberite_debug(_Z15PrintStackTracev+0x23)[0xb4d2f3]
./Cuberite_debug(_ZN6cChunk12RemoveEntityEP7cEntity+0x45)[0xa49071]
./Cuberite_debug[0xe050e6]
./Cuberite_debug[0xe0b09c]
./Cuberite_debug(_ZNKSt8functionIFvR6cWorldEEclES1_+0x49)[0xb2494d]
./Cuberite_debug(_ZN6cWorld15TickQueuedTasksEv+0x205)[0xb16dc1]
./Cuberite_debug(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x4da)[0xb16430]
./Cuberite_debug(_ZN6cWorld11cTickThread7ExecuteEv+0xad)[0xb0c3e3]
./Cuberite_debug(_ZN9cIsThread9DoExecuteEv+0x2f)[0xb4266b]
./Cuberite_debug(_ZNKSt7_Mem_fnIM9cIsThreadFvvEEclIIEvEEvPS0_DpOT_+0x65)[0xb43c73]
./Cuberite_debug(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EE9_M_invokeIJLm0EEEEvSt12_Index_tupleIJXspT_EEE+0x43)[0xb43bc3]
./Cuberite_debug(_ZNSt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS1_EEclEv+0x1d)[0xb43ab9]
./Cuberite_debug(_ZNSt6thread5_ImplISt12_Bind_simpleIFSt7_Mem_fnIM9cIsThreadFvvEEPS3_EEE6_M_runEv+0x1c)[0xb43a36]
/usr/lib64/libstdc++.so.6(+0xd4ed6)[0x7f76a0218ed6]
/lib64/libpthread.so.0(+0x7aa1)[0x7f76a0484aa1]
/lib64/libc.so.6(clone+0x6d)[0x7f769f9feaad]
Cuberite_debug: /data1/minecraft_src/LogicParrot/cuberite/src/Chunk.cpp:1958: void cChunk::RemoveEntity(cEntity*): Assertion `0' failed.
Aborted (core dumped)
```

Seadragon91:
@firnsan I don't think that the branch is ready for testing.

LogicParrot:
The branch was good for testing at first, but the latest commit is somewhat unstable. So yes, please do not test yet.

firnsan:
ok, i will going  on to test the `master`.

PureTryOut:
So... Is anyone still working on this? It seems promising.

LogicParrot:
I am still working on it, slowly. This isn't abandoned. I am just too busy at the moment.

src/Entities/Entity.cpp
LogicParrot: I think this hack could be eliminated once we merge #3489

src/Entities/Entity.cpp
LogicParrot: I think this hack could be eliminated once we merge #3489

PureTryOut:
Good to know, thanks for your response. As long as it isn't abandoned ;)

LogicParrot:
Need to wait for https://github.com/cuberite/cuberite/pull/3489

LogicParrot:
@sphin

LogicParrot:
@sphinxc0re ?

LogicParrot:
My PRs should eventually get merged. Especially this one.

sphinxc0re:
I closed every PR that is older than december 2016. If you are working on it you can always reopen the PR when you are ready. At this point none of your PRs are mergable

LogicParrot:
I understand the need for closing inactive PRs, and I think useful ones should be [tagged](https://github.com/cuberite/cuberite/pulls?q=is%3Apr+is%3Aclosed+label%3A%22status%2Fextract+useful+bits%22) in hopes someone will one day extract the good code out of them.

But *active* PRs should remain open regardless of age. Closing hides the fact that someone is working on a certain feature. If you close "refactor mob code" while active, someone could get the impression that mob code is not being worked on and they may submit a PR themselves. Also, I can reopen but others can't, and closing is a way of implicitly telling them "we don't want your code".

sphinxc0re:
That is exactly what I'm trying to tell developers. If you won't hesitate working on your code, we won't bother keeping your PR open. That is very strict and keeps the PR flow steady. This PR is absolutely not mergable and that is why I closed it. Also because the age but mostly because it is not mergable and hangs around useless.

sphinxc0re:
So the message would be: If your code is not _ready_, we don't want it and that is correct.

LogicParrot:
>If you won't hesitate working on your code, we won't bother keeping your PR open... If your code is not ready, we don't want it and that is correct.

This is absolutely crazy. This is at least the second time you lash out at code contributors and this is unacceptable to me.  No one owes you anything. This is a volunteer FOSS effort and we do it in our free time. It isn't some corporate deadline oriented company. If someone's code is not ready yet, you patiently wait for it to be ready and you thank him for his work. 

You close active PRs for some arbitrary notion of "flow". You want to shuffle the tags I use for issue lookups notion of "efficiency". You remove my "check if 1.9 only" tag. You're dealing with totally superficial things for some vague notions of "efficiency", "strictness", "steady flow", when in reality the only thing we really need is for people to write code, documentation, and perform testing. You *want* my code, and that is the thing that actually drives projects forward.

madmaxoft:
I have to side with @sphinxc0re on this one. This PR has been "in the making" for ages and it doesn't seem it would be ready any time soon. The last few months don't seem like you (@LogicParrot) were working on it at all. So basically you are blocking anyone else from actually developing the feature if they ever do have time. Yes, we do want your code, but we want it *only* if it is ready, it is clean and it works. A PR is saying "here, have this code"; your PR is "hey, I'm working on this, don't anyone else dare, and if I ever get to finish it, you can merge it"

madmaxoft:
And this discussion should be moved to the forum: https://forum.cuberite.org/thread-2908-post-27798.html

yangm97:
I think that Julian might be under the impression that a messy workflow and lack of focus might be slowing things down. The lack of defined goals and doing many things at once are proven to slow progress, psychologically. So I think he’s right.

Meanwhile, we do need more manpower to actually do things, so I also think you’re right.

A wise man once has said: [Commit Often, Perfect Later](https://sethrobertson.github.io/GitBestPractices/). Maybe merging pull requests like there’s no tomorrow  and then doing like monthly stable releases would speed things up? Maybe, maybe not, I don’t know. What I do know is there are projects similar to cuberite (non-profit and FOSS) with a similar number of contributors (sometimes even less) managing to move faster, so yeah, changing the workflow and imposing rules here and there might actually speed things up.

LogicParrot:
You agreed to close PRs that haven't had commits in a month, but I am simply waiting for tiger's entity changes to avoid conflicts, so this is a special case where lack of commits is justified (and is marked with status/blocked). But anyways the specific PR policy is not what's bothering me. I would like to contact you privately but I lost my forum access and password reset is failing. How can I resolve this?

LogicParrot:
>And this discussion should be moved to the forum: https://forum.cuberite.org/thread-2908-post-27798.html

I admit I missed that post. Should have discussed things there before reopening. Apologize for that.

LogicParrot:
Login worked. It turns out to be a Javascript issue The site used to work without JS but now there's a new  reCAPTCHA. I hope you don't mind I delete the posts here to undo the noise I caused. I'll move my relevant bits to the forum.

LogicParrot:
Opening just to let others know I am working on this and to avoid duplicates. The branch itself will never be merged. Too many conflicts with @tigerw 's smart pointers. I'm rewriting it from scratch. The concepts remain the same. New Pull request soon.

LogicParrot:
Actually, the conflicts are with other things and not the smart pointer update. (I expected the update to modify m_Target and friends but it did not). 

Anyways, I am working on a new PR. I'd advise against making any major mob changes for now to avoid conflicts. I am working on it on a daily bases, so this won't stagnate for too long.

LogicParrot:
If anyone is interested in seeing the new code early please see this branch: https://github.com/LogicParrot/cuberite/tree/mobs2

Please note the code is rapidly evolving and is incomplete. Also, some parts are there for temporary testing - For instance th `for(;;)` in AgressiveMonster/PassiveMonster will be replaced by a vector of `cBehaviors` that will be iterated.

LogicParrot:
Cows now work fully using the behaviors system. Branch still unstable for other mobs.

madmaxoft:
At first look, your Behaviors are not exactly what I had in mind when we were discussing it. I'll have a better look over the weekend if it's just my misunderstanding or you're using something different.

LogicParrot:
@madmaxoft OK. Is the hard-coding as member functions the problem? If so, note that this may or may not be temporary. (It has pros and cons). I can easily move them into a vector of unique-pointers whenever we want to, making each mobs' behaviors extensible at runtime.

madmaxoft:
The original thought was that a behavior is a "slot" inside the mob where you could plug many different instances, each implementing the same interface. Each mob, regardless of its type, would have the same amount and same types of slots, and only the runtime assignment of behavior classes into these slots would differentiate between the various mob types.

Basically (high level, without thinking about it too much about details), what we want is a class that decides what to do when the mob for example:
 - sees another mob
 - sees a player
 - starts burning
 - is rightclicked by the player
 - is attacked by another entity

So for example aggressive mobs would have a `cBehaviorSeePlayerAttack` instance in their "sees a player" slot, passive mobs and neutral mobs would have a `cBehaviorSeePlayerIgnore` in that slot, breedable mobs would have a `cBehaviorSeePlayerCheckBreedableItem` in that slot.

That's why I'm very confused as to why your behaviors need an `AttachToMonster` method - in my understanding of behaviors, the behavior classes don't need to keep track of the mob, they only are called from within the tick function (or others) like `this->m_BehaviorSeePlayer(*this, ...);` (assuming the `operator ()` was what implements the behavior).

LogicParrot:
There is a fundamental misunderstanding. The behaviors were meant to be mixin-like classes that you plug arbitrarily into monsters. You can plug 0 to a monster, making it a non reacting dummy, or 20. 

A mob is an "empty shell" that receives events e.g. rightClick, destroy, take damage, and then dispatches them to interested behaviors. The behaviors act upon the events. Different behaviors register to different events.

For instance, the cow currently has two behaviors that are interested in the rightclick event: Breeding and itemReplacing(milking). When we right click a cow, that event is dispatched to both. If we're holding wheat, the breeding Behavior reacts, and if we're holding a bucket, the itemReplace behavior reacts, replacing the bucket with a milk  bucket.

This makes behaviors modular, we can plug the itemReplace behavior to a lava slime (and pass its constructor a lava bucket and not a milk bucket), and now we can extract lava buckets from lava slimes by right clicking them.

LogicParrot:
I did some undiscussed things in Tick though: `passiveTicks` are now seperated into `preTick` and `postTick`. And `ActiveTicks` are now simply `Tick`.

madmaxoft:
Okay, that sounds reasonable. As long as it's a `vector` of behaviors :)
Do the behaviors register for the events they are interested in, or are the events sent to every behavior?

LogicParrot:
Oh! there's also variants of behaviors which closely resembles what you describe. There are multiple ways to chase an opponent for instance, so we have an abstract `behaviorChaser`, we have derived classes like `BehaviorChaserSneeker` and `BehaviorChaserTeleporter`. We then attach `BehaviorChaserSneeker` to a creeper and `BehaviorChaserTeleporter` to an enderman. So perhaps the misunderstanding is not so fundemental. It's just that not everyone has a chaser slot. Some monsters have chasing behaviors attached and some don't.

yangm97:
So it’s even better than before? 😮

Will you also export these to the Lua API? So that you can add to or override any mob’s behaviour (making lobbies/shoppings come to my mind).

LogicParrot:
>Do the behaviors register for the events they are interested in, or are the events sent to every behavior?

Behaviors only register for the events they are registered in, each behavior specifies its desired events in its `attachToMonster` method. The monster has multiple vectors, one for each event type. Behaviors are pushed to the vectors of the events they want. The Tick event vector is special: The behaviors are asked, in order, if `IsControlDesired`, and the first one that says yes is the only one that gets the tick event. So the order of the behaviors in that vector dictate the mob's priorities. e.g. a cow's cowardness behavior is first in the Tick vector and wandering is last, so hitting it always triggers running away no matter what it was doing before, and the wander gets the tick event only if no other behavior desires the tick event.

LogicParrot:
>So it’s even better than before? 😮

Well, the variants were there before. Some behaviors have variants and some don't. But it does incorporate some lessons learned from before, mainly in tick handling! For instance, when a behavior takes over, it sometimes needs "setup" and  "tear down" stages that were not possible before. Example: When `BehaviorCoward` takes control of the cow, it sets the speed to x3 in the setup stage (`ControlStarting`), ticks for a while, then sets it back to normal in the  tear down stage (`ControlEnding`).

As for a Lua API, The behaviors are currently hard-coded, but I am doing them with future user extensibility in mind at every step. We have vectors in which behavior can be swapped in and out, hopefully and I'm trying to keep it generic, for instance, rather than `behaviorMilking` we have `behaviorItemReplacer` that can be used for other user defined things too (like the lava 4 comments example above).

So, the code should become technically ripe for a future API export. But I don't want to make promises I cannot keep! I'm aiming the code in that direction but we'll see what time permits and where it eventually goes. The first pull request *will not* include API exports. It's a bad idea to export that early because we need testing, and experience with the system to see where it can be tweaked and improved. It'd be a waste of time to export a highly volatile API.

I guess I outlined most of my undiscussed changes in these comments, but I'll write them in an orderly manner once the new pull request is ready.

LogicParrot:
@madmaxoft @tigerw What's your opinion of the following commit? It is much safer than the previous raw entity pointers like `m_Target` and far simpler than  `cEntity::StopEveryoneFromTargetingMe` and its accompanying functions: https://github.com/LogicParrot/cuberite/commit/00febf68ad8bc36546c02b0c0b926189fb3955a1

Progress update/self-reminder: All passives working. Agressives don't crash and just wander around, squids work. Wolves todo. Ocelots unstested.

peterbell10:
IMO `cMobPointer` looks like a more dangerous version of `weak_ptr`.

LogicParrot:
One reason I didn't use weak_ptrs is that i'm reluctant to change `cChunk:m_Entities` from unique_ptrs to shared ptrs at the moment, because it may significantly fatten an already fat PR. But that's something I'm seriously considering in a followup PR. `cMobPointer` is still safer than the pure raw that was done till now.

 Also note that `weak_ptrs` on their own don't get the job done: moving a world does not invalidate a weak_ptr. We'd still need a check of some sort. (Moving a world does invalidate cMobPointer by returning false at `isTicking`). @worktycho previously suggested copying the entity and destroying the older one rather than moving it to resolve this.

peterbell10:
You could always have a class wrap a `weak_ptr` so that locking it compares the world automatically.

LogicParrot:
Agreed. selfnote: Make sure current world travels change m_World *before* transferring ownership to the new thread. selfnote2: Would a weak ptr /  cMobPointer invalidate properly on rapid back and forth teleports? Is a counter needed?

madmaxoft:
The `cMobPointer` you created is multithread-unsafe. Better use `shared_ptr` and `weak_ptr`, I have always assumed this will be needed with AI / pathfinding / behaviors.

LogicParrot:
The `cMobPointer` wrapper is needed even when using an underlying `weak_ptr`; We must not use a pointer which is pointing to an entity in other worlds, so we must check for the world whenever we convert the `weak_ptr` to a temporary `shared_ptr`. This is better done with a wrapper rather than manual world comparisons all over the code.

I was planning to use `cMobPointer`, and then potentially replacing its underlying raw pointer with a `weak_ptr` in a seperate PR later. This PR will really grow if everything is done in one hit.

Also, just like its predecessor (m_Target), `cMobPointer` is never expected to point cross threads, so thread safety is not strictly needed.

LogicParrot:
Progress update: Zombies kind are working now (still some problems with sun avoidance), and I think I've homed in on the final Attack mechanism. The rest of the aggressive mobs should be more straightforward. Wolves will probably be the biggest challenge to overhaul.

LogicParrot:
branch moved: https://github.com/cuberite/cuberite/tree/mobs2

PR #3301, opened 2016-08-07
Removed "." and ".." from cFile:GetFolderContents().

### LINKED ISSUE #3299: Using cFile:GetFolderContents in combination with cFile:Delete can be dangerous
### Expected behavior
The function `cFile:GetFolderContents` should return the content of a folder, but without the current folder and the upper folder..

### Actual behavior
The function `cFile:GetFolderContents` returns the content of a folder.
Problem here is that it also returns the current folder and the upper folder. In combination with `cFile:Delete`, it deletes the current folder and the upper folder
Example Plugin APIDump, has this problem [main_APIDump.lua#L1288L1289](https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/main_APIDump.lua#L1288L1289)
```
API/Static/.
API/Static/..
API/Static/.gitignore
API/Static/decoda_debug_settings.png
API/Static/decoda_logo.png
API/Static/decoda_workspace.png
API/Static/zbs_logo.png
API/Static/zbs_workspace.png
```
The plugin APIDump loops over the content of the folder API/Static and deletes all files/folders in the current folder and in the upper folder.

1) Rename/Delete a already existing API folder in the root directory
2) Enable the plugin APIDump
3) Start the server
4) Run the command api

The folder API is not existent and in the sever console comes this message:
```23:13:01] Cannot output HTML API: API/index.html: No such file or directory```

### Scenario:
A developer writes a plugin where the player files are in a folder named `CTF/playefiles`. He is testing his reset function, that loops over the folder content and uses `cFile:Delete`. Only to find then out that his whole plugin is deleted.

### Solution
The function `cFile:GetFolderContents` shouldn't return the upper and current folder.

madmaxoft:
Fixes #3299 .

### DIFF
--- a/src/OSSupport/File.cpp
+++ b/src/OSSupport/File.cpp
@@ -524,6 +524,13 @@ AStringVector cFile::GetFolderContents(const AString & a_Folder)
 		{
 			do
 			{
+				if (
+					(strcmp(FindFileData.cFileName, ".") == 0) ||
+					(strcmp(FindFileData.cFileName, "..") == 0)
+				)
+				{
+					continue;
+				}
 				AllFiles.push_back(FindFileData.cFileName);
 			} while (FindNextFileA(hFind, &FindFileData));
 			FindClose(hFind);
@@ -546,6 +553,13 @@ AStringVector cFile::GetFolderContents(const AString & a_Folder)
 			struct dirent *dirp;
 			while ((dirp = readdir(dp)) != nullptr)
 			{
+				if (
+					(strcmp(dirp->d_name, ".") == 0) ||
+					(strcmp(dirp->d_name, "..") == 0)
+				)
+				{
+					continue;
+				}
 				AllFiles.push_back(dirp->d_name);
 			}
 			closedir(dp);


madmaxoft:
Haven't actually tested this on Linux, could someone do that please?

Seadragon91:
Tested it under Linux and it works. Now the plugin APIDump also works correctly.

Seadragon91:
Rebased the pull request.

NiLSPACE:
This breaks allot of my plugins I'm afraid :|

Seadragon91:
You are using the `.``, `..` in a plugin?

NiLSPACE:
No, but I sometimes remove the first two items in the array before doing stuff.

Seadragon91:
@NiLSPACE Under Linux, the current folder is not at top, only the upper folder. A for loop would be necessary.

LogicParrot:
So, there might have been a Linux bug in those plugins anyways.

Seadragon91:
Ag, my Plugin SkyBlock is now also broken...

sphinxc0re:
We should write an announcement on twitter and the newsletter

PR #3306, opened 2016-08-09
Fixed NetherBrick recipes

jammet:
Again. This used to work, according to a player.


mathias-github:
Thanks! I forgot to fix this when I changed the item name.

PR #3307, opened 2016-08-10
Change Plugins.m_Version from int to double

RubenEu:
I think that is interesting to change that type variable for plugins that have minor versions.

For example: [Cuberite:Essentials](https://github.com/cuberite/Essentials/blob/master/Info.lua) version actually is 0.2, but with int the version is narrowed to 0.

### DIFF
--- a/src/Bindings/Plugin.h
+++ b/src/Bindings/Plugin.h
@@ -114,8 +114,8 @@ public:
 	const AString & GetName(void) const  { return m_Name; }
 	void SetName(const AString & a_Name) { m_Name = a_Name; }
 
-	int GetVersion(void) const     { return m_Version; }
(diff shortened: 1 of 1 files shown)


bearbin:
Is there actally any importance in having the version be a number? Having it be a string would be most sensible.

RubenEu:
I think that double is enough for compare and have minor versions specificated.

Seadragon91:
This could break plugins, if they have no int values in version.

madmaxoft:
Actually, this is a very bad idea. The version is meant to be programmatically compared and is NOT intended at all to be visible to the server admins. More or less the same as Android's API levels.

One counterexample to break your thinking: how about version `0.13`, is it really older than version `0.6`? Your code would make it so.

PR #3310, opened 2016-08-11
Add UrlClient API

madmaxoft:
Adds a new API, cUrlClient, for simplifying HTTP(S) requests to other servers.
 - [x] C++ API class `cUrlClient` wrapping the HTTP parser
 - [x] TLS support for the API
 - [x] Export the new class to Lua

### DIFF
--- a/Server/Plugins/APIDump/Classes/Network.lua
+++ b/Server/Plugins/APIDump/Classes/Network.lua
@@ -1,7 +1,7 @@
 
 -- Network.lua
 
--- Defines the documentation for the cNetwork-related classes
+-- Defines the documentation for the cNetwork-related classes and cUrlClient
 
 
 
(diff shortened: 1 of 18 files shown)


sphinxc0re:
So, is this the long awaited HTTPClient? I don't really get it

madmaxoft:
This is the raw C++ version of the thing - it implements the actual callback-based client. Have a look at the new `tests/HTTP/UrlClientTest.cpp` file that shows the usage.

The remaining work to do is the Lua API export; I want to do that in an intelligent way, so that the requests are simple to use yet customizable, if needed:
```lua
-- Basic request:
cUrlClient:Request("GET", "https://google.com",
  function (a_Body)
    -- Process the entire response body received in a_Body
  end
)


-- "Streaming" request with custom headers and options:
-- Define the callbacks to use:
local callbacks =
{
  OnBodyData = function(a_BodyPart)
    -- Another part of the body was received, handle it
  end,
  OnBodyFinished = function()
    -- No more incoming data
  end,
}
-- Serialize a table representing a form into a multipart/form-data format:
local reqBody = cUrlClient:SerializePostFormMultiPart(
  {
    ["Request"] = "CheckPluginVersion",
    ["PluginIdent"] = "cuberite.org/WorldEdit",
  }
)
-- Make the actual request:
cUrlClient:Request("POST", "https://pluginrepo.cuberite.org/versioncheck", callbacks,
  {  -- Custom headers:
    ["Repo-Protocol-Version"] = "1",
    ["Accept-Language"] = "*/*",
    ["Content-Type"] = "multipart/form-data"
  },
  reqBody,
  {  -- Options:
    ["MaxRedirects"] = 10,
    ["OwnCert"] = "SomeClientCertData",
    ["OwnPrivKey"] = "PrivateKeyForOwnCert",
    ["OwnPrivKeyPassword"] = "PasswordForOwnPrivKey",
  }
)
```

sphinxc0re:
So it is indeed an HTTPClient. Why don't you call it like that?

madmaxoft:
Because it can do `http` as well as `https` now, and it's extensible so it could do more protocols in the future (`ftp` comes to mind, it could wrap around local files as well to unify the interface). Just like Python's `urllib`

sphinxc0re:
Okay, you have a point there 👍

madmaxoft:
Oh god, I just love errors in templates. Check it out:
https://travis-ci.org/cuberite/cuberite/jobs/153664892#L1674
It really takes up over 1600 lines to report that error :-O

madmaxoft:
Yay, finally the linux build passes, and now the only thing missing is the new API documentation.

NiLSPACE:
@madmaxoft Is there a chance that we can get something like [this](https://forum.cuberite.org/thread-2062.html)?

madmaxoft:
Almost. You can store anything in the table that you put the callbacks in, but those callbacks don't get that table as their hidden first param (`self`), so you cannot use the colon calling convention for them.

NiLSPACE:
Then how do you access the table from the callback? That was the whole idea, that you can access the object so every network request has it's own member variable.

It's not really required, but it would make things allot easier.

madmaxoft:
It would be possible with some trickery (enclose the table itself into the function closure) but it would be ugly.

I agree that having the callbacks table as the first param is a neat idea. I'll try if it is doable in C++ easily.

NiLSPACE:
Great :)

madmaxoft:
I now consider this complete and ready for review.

The API documentation can be previewed in HTML form at http://cuberite.xoft.cz/UrlClient/cUrlClient.html

madmaxoft:
Note that `cUrlClient:SerializeHTMLFormMultiPart()` is not part of the API yet.

NiLSPACE:
That looks great :) I'm wondering though, does calling the function with the table take the metatable in consideration? Otherwise `__index` won't work and you can't use it as a class.

sphinxc0re:
@madmaxoft Is there a way to mark it as 'not yet implemented' ?

madmaxoft:
@sphinxc0re There is no such function in the API yet, so it is "not yet implemented" by definition. I just wrote that because I used that function in the example in my comment above.

NiLSPACE:
Other than the link in the docs that is invalid and [my question](https://github.com/cuberite/cuberite/pull/3310#issuecomment-241642907) I'd really like to get this merged quickly :)

madmaxoft:
I have fixed the link. As for your question, it should work, but haven't actually tested it. The code uses `lua_getfield()` to get the function out of the table, and that should trigger the `index` metamethod: http://pgl.yoyo.org/luai/i/lua_getfield

NiLSPACE:
Great, then LGTM :D

NiLSPACE:
I was porting WorldEdit's version checker over, but Cuberite crashed while parsing the headers [here](https://github.com/cuberite/cuberite/blob/master/src/HTTP/EnvelopeParser.cpp#L121)

This was most likely because there was a header with no value.
`0x04b9d2f5 "X-Geo-Block-List:\r\nX-GitHu..`

The url where it crashed was: https://raw.githubusercontent.com/cuberite/WorldEdit/master/Info.lua

madmaxoft:
Good catch, I'll fix it once I get home.

madmaxoft:
Fixed in #3342.

PR #3311, opened 2016-08-11
Lua tracked ref

madmaxoft:
This adds the same treatment as `cLuaState::cCallback` (tracking, reset when plugin unloads) to a separate `cTrackedRef` class that can hold a reference to any Lua object. This is in turn used for the Lua callback tables in `cTableRef`.

### DIFF
--- a/src/Bindings/LuaNameLookup.cpp
+++ b/src/Bindings/LuaNameLookup.cpp
@@ -10,9 +10,8 @@
 
 
 
-cLuaNameLookup::cLuaNameLookup(const AString & a_Query, cPluginLua & a_Plugin, int a_CallbacksTableStackPos):
-	m_Plugin(a_Plugin),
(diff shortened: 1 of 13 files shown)


madmaxoft:
Basically this PR makes more stuff use the same approach as in `cLuaState::cCallback`.

NiLSPACE:
Ooh, that's great. I can't wait :P

madmaxoft:
Hm, this is not good. I'm pretty sure I had a more advanced version of this code, but somehow seem to have lost it :(
I'll have to work on this a bit more.

madmaxoft:
Went through some 10 possible locations where I could have written that code, haven't found it in any of them, so I had to rewrite it from scratch. And I think it's even better now than what I remember writing it the first time.

Incidentally, git's `stash` feature is really, *really* underdeveloped. There's basically no way to inspect the stash contents.

PR #3316, opened 2016-08-14
Introducing World:SetSpawn() API and Lua binding

ElNounch:


madmaxoft:
Looks good to me. I'd appreciate one more person reviewing this, though.

src/World.cpp
```diff
@@ -324,6 +324,33 @@ void cWorld::SetNextBlockTick(int a_BlockX, int a_BlockY, int a_BlockZ)
...
+{
+		cIniFile IniFile;
+
+		IniFile.ReadFile(m_IniFileName);
+		IniFile.SetValueF("SpawnPosition", "X", a_X);
+		IniFile.SetValueF("SpawnPosition", "Y", a_Y);
+		IniFile.SetValueF("SpawnPosition", "Z", a_Z);
+		if (IniFile.WriteFile(m_IniFileName))
+		{
+			m_SpawnX = a_X;
```
bearbin: Should these be initialised even if the file cannot be written?

bearbin:
Aside from my comment, LGTM.

src/World.cpp
madmaxoft: I don't have an opinion in either way. Once you get a file write error, having a correct spawn coord is the least of your problems; perhaps a log message would be good in this case, though.

NiLSPACE: I think that could only cause confusion after a restart. *what!? the spawn is back to where it was! It was at the new place a minute ago.* Though of course they'd have to miss the warning log in the console for that as well.

ElNounch: That was the idea, if it ultimately failed (couldn't store information), don't look at all as if it worked.

madmaxoft:
Ay, don't merge master, rebase onto it. Next time.

PR #3327, opened 2016-08-17
Fix cWorld:QueueTask() Lua's binding

ElNounch:
Manual binding expect three parameters : self, a number, a function
And then read self and a function...
Removing the extra check for the broken second argument


madmaxoft:
Thanks, I must have missed that.

PR #3328, opened 2016-08-18
Lua binding for cIniFile.noID

### LINKED ISSUE #3324: No Lua binding cIniFile.noID
Inherited noID enum of cIniFile class isn't exported to Lua's bindings.

ElNounch:
Fix #3324

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -3821,6 +3821,10 @@ void cManualBindings::Bind(lua_State * tolua_S)
 			tolua_function(tolua_S, "GetOutputBlockPos", tolua_cHopperEntity_GetOutputBlockPos);
 		tolua_endmodule(tolua_S);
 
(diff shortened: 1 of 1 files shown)


sphinxc0re:
@cuberite/core-team someone seems to be quite busy here 😉

madmaxoft:
Does that really need manual binding, can't it be exported automatically?

madmaxoft:
Now we have a choice - this PR, or #3329? :)

ElNounch:
@madmaxoft : both are ugly :worried:, proper fix of the dependency would be neat. :grimacing:
Don't you think so ? :grinning:

madmaxoft:
ToLua++ could handle the dependency, but we don't want to export the base class.

I'm trying a slightly better version of my original approach, without the preprocessor guards.

ElNounch:
Superseded by https://github.com/cuberite/cuberite/pull/3329

PR #3330, opened 2016-08-18
Delay EntityChangedWorld players' callback until Entity fully linked …

ElNounch:
…to world

Otherwise, some API call just don't seems to happen
gitignore tweak for test executables

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1827,8 +1827,7 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
 		);
 		ParentChunk->RemoveEntity(this);
-		a_World->AddPlayer(this);
-		cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*this, a_OldWorld);
+		a_World->AddPlayer(this, &a_OldWorld);  // New world will appropriate and announce client at his next tick
 	});
 	return true;
 }
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -2990,10 +2990,10 @@ void cWorld::CollectPickupsByPlayer(cPlayer & a_Player)
 
 
 
-void cWorld::AddPlayer(cPlayer * a_Player)
+void cWorld::AddPlayer(cPlayer * a_Player, cWorld * a_OldWorld)
 {
 	cCSLock Lock(m_CSPlayersToAdd);
-	m_PlayersToAdd.push_back(a_Player);
+	m_PlayersToAdd.emplace_back(a_Player, a_OldWorld);
 }
 
 
@@ -3010,7 +3010,10 @@ void cWorld::RemovePlayer(cPlayer * a_Player, bool a_RemoveFromChunk)
 	}
 	{
 		cCSLock Lock(m_CSPlayersToAdd);
-		m_PlayersToAdd.remove(a_Player);
+		m_PlayersToAdd.remove_if([&](const std::pair< cPlayer *, cWorld * > & value) -> bool
+		{
+			return (value.first == a_Player);
+		});
 	}
 	{
 		cCSLock Lock(m_CSPlayers);
@@ -3895,7 +3898,7 @@ void cWorld::AddQueuedPlayers(void)
 	ASSERT(m_TickThread.IsCurrentThread());
 
 	// Grab the list of players to add, it has to be locked to access it:
-	cPlayerList PlayersToAdd;
+	cAwaitingPlayerList PlayersToAdd;
 	{
 		cCSLock Lock(m_CSPlayersToAdd);
 		std::swap(PlayersToAdd, m_PlayersToAdd);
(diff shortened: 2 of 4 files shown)


src/World.cpp
```diff
@@ -3933,16 +3938,26 @@ void cWorld::AddQueuedPlayers(void)
 	}  // Lock(m_CSClients)
 
 	// Stream chunks to all eligible clients:
-	for (cPlayerList::iterator itr = PlayersToAdd.begin(), end = PlayersToAdd.end(); itr != end; ++itr)
+	for (auto & AwaitingPlayer : PlayersToAdd)
 	{
-		cClientHandle * Client = (*itr)->GetClientHandle();
+		auto & Player(AwaitingPlayer.first);
```
madmaxoft: I very much prefer the assignment syntax, rather than construction. At a glance, this is difficult to distinguish from a function call.

src/World.cpp
```diff
@@ -3933,16 +3938,26 @@ void cWorld::AddQueuedPlayers(void)
...
+			Client->SendWholeInventory(*Player->GetWindow());
 		}
 	}  // for itr - PlayersToAdd[]
+
+	// Call EntityChangedWorld callback on all eligible clients
+	for (auto & AwaitingPlayer : PlayersToAdd)
+	{
+		if (AwaitingPlayer.second)
+		{
+			cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*((cEntity *) AwaitingPlayer.first), *AwaitingPlayer.second);
```
madmaxoft: Avoid C-style cast.

src/World.cpp
```diff
@@ -3933,16 +3938,26 @@ void cWorld::AddQueuedPlayers(void)
...
 			Client->SendHealth();
-			Client->SendWholeInventory(*(*itr)->GetWindow());
+			Client->SendWholeInventory(*Player->GetWindow());
 		}
 	}  // for itr - PlayersToAdd[]
+
+	// Call EntityChangedWorld callback on all eligible clients
+	for (auto & AwaitingPlayer : PlayersToAdd)
+	{
+		if (AwaitingPlayer.second)
```
madmaxoft: Use explicit comparison to `nullptr` on pointers.

src/Entities/Player.cpp
```diff
@@ -1827,8 +1827,7 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
 		);
 		ParentChunk->RemoveEntity(this);
-		a_World->AddPlayer(this);
-		cRoot::Get()->GetPluginManager()->CallHookEntityChangedWorld(*this, a_OldWorld);
+		a_World->AddPlayer(this, & a_OldWorld);
```
madmaxoft: Remove the space after `&`

src/World.h
```diff
@@ -263,7 +264,7 @@ class cWorld :
 	/** Adds the player to the world.
 	Uses a queue to store the player object until the Tick thread processes the addition event.
 	Also adds the player as an entity in the chunkmap, and the player's ClientHandle, if any, for ticking. */
-	void AddPlayer(cPlayer * a_Player);
+	void AddPlayer(cPlayer * a_Player, cWorld * a_OldWorld = nullptr);
```
madmaxoft: Add an explanation for the `a_OldWorld` param to the Doxy-comment.

ElNounch:
Thanks for the comments, should fit you better now. :wink:
It even pass the CI checks now ! :stuck_out_tongue_winking_eye:

src/World.cpp
```diff
@@ -3010,7 +3010,10 @@ void cWorld::RemovePlayer(cPlayer * a_Player, bool a_RemoveFromChunk)
 	}
 	{
 		cCSLock Lock(m_CSPlayersToAdd);
-		m_PlayersToAdd.remove(a_Player);
+		m_PlayersToAdd.remove_if([&](const std::pair< cPlayer *, cWorld * > & value) -> bool
+		{
+			return (value.first == a_Player);
```
LogicParrot: Why?

LogicParrot: My bad. Ignore this.

LogicParrot:
Looks good to me

tigerw:
I expect the 'proper' way to fix this is to make AddPlayer synchronous. Alas, #3115 is not ready yet.

LogicParrot:
How is that better? The sender world would have to wait and right now it doesn't, and the recipient would still have to behave the same, finish the tick and all, then accept the player. Is it simpler in terms of code?

tigerw:
Definitely. For one you replace all this PlayersToAdd nonsense with QueueTask.

LogicParrot:
Oh, Queuetask is convenient, but that's not the same as "synchronos", is it not?

tigerw:
Well, with other changes as well to make cWorld only accessed from one thread.

Why did you quote "synchronous"? :P

LogicParrot:
Because I don't fully understand what's synchronous here. One world leaves a "message", the other world eventually gets to it by executing the queued tasks.

But I'm all for Queuetasks rather than the queues, if that's what you mean.

tigerw:
I was referring to making AddPlayer add a player immediately instead of delaying with a queue.

madmaxoft:
@tigerw you can't really do that, because that would tie the two worlds' tick threads, which would make them lag, especially if this was extended to regular entities moving between worlds (nether portal). This operation needs to be asynchronous.

tigerw:
Ah, but introducing QueueTask. Could you take a moment to review our Lord and Saviour, #3115?

PR #3331, opened 2016-08-19
cLuaState: Added template to push multiple values in a single call.

madmaxoft:
Simplifies writing bindings.


sphinxc0re:
I noticed that every LuaState instance (L, S) is named differently. I would be best to introduce a convention for that

madmaxoft:
I'm not sure if it's worth the effort.

PR #3333, opened 2016-08-19
WebAdmin: Report opened ports.

madmaxoft:
Simply outputs the webadmin ports to log.

### DIFF
--- a/src/HTTP/HTTPServer.cpp
+++ b/src/HTTP/HTTPServer.cpp
@@ -110,12 +110,12 @@ bool cHTTPServer::Initialize(void)
 	// Notify the admin about the HTTPS / HTTP status
 	if (m_Cert.get() == nullptr)
 	{
-		LOGWARNING("WebServer: The server is running in unsecured HTTP mode.");
+		LOGWARNING("WebServer: The server will run in unsecured HTTP mode.");
 		LOGINFO("Put a valid HTTPS certificate in file 'webadmin/httpscert.crt' and its corresponding private key to 'webadmin/httpskey.pem' (without any password) to enable HTTPS support");
 	}
 	else
 	{
-		LOGINFO("WebServer: The server is running in secure HTTPS mode.");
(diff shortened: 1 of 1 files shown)


madmaxoft:
Such a nice issue number and wasted for such a trivial matter :P

tigerw:
There's always #5318008

sphinxc0re:
or just

#01189998819991197253

tigerw:
To be honest, judging by the number of bugs in Cuberite, that's a perfectly plausible number to reach.

#itcrowd

LogicParrot:
That is impossible to reach regardless of the amount of bugs, because there's a leftmost zero.

*(Unless the merger of #3115 causes some spacetime rapture, making everything possible)*

tigerw:
Curses

Cl1608Ho:
If we skip that leftmost zero and say we create 1000 issues/pr per second (and say we are not getting rate limited by github), we would still need 37 billions of years to accomplish that. 

Although ruby might handle these things automatically (github is a rails application) this number would take up more than 64 bits to store so isn't reachable in most programming languages except for when you are working with strings. 

Also if we assume each issue takes one byte of storage (which it doesn't, even the issue number takes up more) we would still need (github needs) 275 Petabytes of storage just for that (275000 Terabyte). Let's assume each issue needs one kilobyte and you can shift each unit one up. 

And lastly, such a big issue number would probably lead to problems in the website because something like pagination isn't made for billions of pages (just because the numbers would get too long and too wide)

tigerw:
okay okay i concede defeat already

tigerw:
still, a bunch of large numbers don't mean something's impossible, per se

tigerw:
maybe just... improbable

tigerw:
though with the number of bugs we have, maybe it is, I don't know.

bearbin:
#03333

tigerw:
Very self-referential there.

PR #3334, opened 2016-08-20
Add ccache support to build phase

ElNounch:


ElNounch:
More tweaks needed...

LogicParrot:
Worth noting that ccache is working for me regardless of this. I made it work by using the PATH method.

bearbin:
Yep, it's working fine for me also on the buildserver.

madmaxoft:
I have no knowledge at all about this, but I've always assumed that `ccache` works transparently, without any need for support from the projects. If it works for other people, including the buildserver, then it needs no support in the code, but rather at least some hints on how to make it work on the documentation level. Perhaps someone who made it work could write a short how-to in the forum?

LogicParrot:
Yes, ccache is generally supposed to be transparent and a Cuberite-specific howto shouldn't be needed. I am not sure why cmake configs would have ccache-specific flags.

LogicParrot:
Judging from #3337, it seems this was intended to enable caching at the travis CI.

ElNounch:
Once locally setup on your system, it's transparent, yes.
But on CI platform, you fully revisit your setup at each build, plus clang takes a little bit more attention.
Sorry not to have been clearer.

PR #3337, opened 2016-08-21
Travis CI: updated to Trusty's image and ccache

ElNounch:
Add ccache support to build phase.
Work-around Cmake specifying -lpthread before linked files.

What https://github.com/cuberite/cuberite/pull/3334 was supposed to be...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,29 @@
 language: cpp
-sudo: false
+
+sudo: required
+dist: trusty
 
 compiler:
 - clang
 - gcc
 
+cache:
+  ccache
+
 addons:
   apt:
-    sources:
-    - ubuntu-toolchain-r-test
     packages:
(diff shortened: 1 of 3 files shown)


bearbin:
I don't understand the advantage of ccache for travis, since a new virtual machine is built for every build. ccache only speeds up subsequent builds, the first one is actually slower.

ElNounch:
The ccache's directory is moving from builds to builds, pristine. :yum:
(More [info](https://docs.travis-ci.com/user/caching/#ccache-cache) )

madmaxoft:
What are the changes in SetFlags.cmake for? I'd be more comfortable if someone with more Linux background than me checked this PR.

madmaxoft:
Unless `ccache` can be enabled without requiring sudo, I'm in favor of NOT making this change at all.

PR #3342, opened 2016-08-23
HTTP: Fixed empty headers parsing.

madmaxoft:
Empty HTTP headers would cause a crash before this fix.


NiLSPACE:
Merge?

madmaxoft:
I was just typing the same :)

PR #3344, opened 2016-08-24
Using Backward-cpp for stacktraces under Linux

ElNounch:
With [Backward-cpp](https://github.com/bombela/backward-cpp) :
Turning

```
[c6fc1e936562b741|22:29:09] Assertion failed: !"Unhandled block face!", file /home/elnounch/cuberite/src/Blocks/BlockTrapdoor.h, line 82
/home/cuberite/Server/Cuberite_debug(_Z15PrintStackTracev+0x32)[0xb532b2]
/home/cuberite/Server/Cuberite_debug(_ZN21cBlockTrapdoorHandler19BlockFaceToMetaDataE10eBlockFace+0x67)[0xca9a65]
/home/cuberite/Server/Cuberite_debug(_ZN21cBlockTrapdoorHandler25GetPlacementBlockTypeMetaER15cChunkInterfaceP7cPlayeriii10eBlockFaceiiiRhS5_+0x35)[0xca99d7]
/home/cuberite/Server/Cuberite_debug(_ZN12cItemHandler25GetPlacementBlockTypeMetaEP6cWorldP7cPlayeriii10eBlockFaceiiiRhS5_+0x139)[0xc80289]
/home/cuberite/Server/Cuberite_debug(_ZN12cItemHandler16GetBlocksToPlaceER6cWorldR7cPlayerRK5cItemiii10eBlockFaceiiiRSt6vectorI9sSetBlockSaIS9_EE+0x85)[0xc7f959]
/home/cuberite/Server/Cuberite_debug(_ZN12cItemHandler13OnPlayerPlaceER6cWorldR7cPlayerRK5cItemiii10eBlockFaceiii+0x38f)[0xc7f4a1]
/home/cuberite/Server/Cuberite_debug(_ZN13cClientHandle16HandleRightClickEiii10eBlockFaceiiiRK5cItem+0xb92)[0xa6a2f6]
/home/cuberite/Server/Cuberite_debug(_ZN12cProtocol19022HandlePacketBlockPlaceER11cByteBuffer+0x14e)[0xce9afc]
/home/cuberite/Server/Cuberite_debug(_ZN12cProtocol19012HandlePacketER11cByteBufferj+0x3fd)[0xce87ff]
/home/cuberite/Server/Cuberite_debug(_ZN12cProtocol19015AddReceivedDataEPKcm+0x983)[0xce7d91]
/home/cuberite/Server/Cuberite_debug(_ZN12cProtocol19012DataReceivedEPKcm+0x101)[0xcde135]
[...]

PR #3347, opened 2016-08-24
Demangling C++ function names in stack traces

ElNounch:
Using abi::__cxa_demangle

### DIFF
--- a/src/OSSupport/StackTrace.cpp
+++ b/src/OSSupport/StackTrace.cpp
@@ -9,7 +9,9 @@
 	#include "../StackWalker.h"
 #else
 	#ifdef __GLIBC__
+		#include <cxxabi.h>
 		#include <execinfo.h>
+		#include <regex>
 	#endif
 	#include <unistd.h>
 #endif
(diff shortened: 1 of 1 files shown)


src/OSSupport/StackTrace.cpp
```diff
@@ -34,11 +36,35 @@ void PrintStackTrace(void)
...
+
 			auto numItems = backtrace(stackTrace, ARRAYCOUNT(stackTrace));
-			backtrace_symbols_fd(stackTrace, numItems, STDERR_FILENO);
+			char** buffer = backtrace_symbols(stackTrace, numItems);
+
+			for (auto frameId = (numItems - 1); frameId > 0; frameId--)
+			{
+				bool handled = false;
+				std::cmatch m;
+				if (std::regex_match (buffer[frameId], m, lineSplitter))
```
madmaxoft: Do we need to match things with a regex, can't we just send everything for demangling?

I just fear that once a crash is hit, nothing can be considered stable enough, especially memory allocation.

ElNounch: Shouldn't use backtrace or Stackwalker then, but a wrapper shell script :
```shell
#!/bin/sh
ulimit -c unlimited
$@
if [ -e core ]; then
    gdb $1 ./core -ex backtrace -ex quit
    echo Compressing and renaming 'core' file...
    gzip --best core;
    mv core.gz core-$(stat -c '%Y' core.gz).gz;
fi
```

madmaxoft: Okay, still, there's my first question: Why regex at all, why not send everything in?

ElNounch:
Need to isolate the mangled name part, yes.

src/OSSupport/StackTrace.cpp
ElNounch: Need to isolate the mangled name part, yes.

sphinxc0re:
This PR and #3344 look promising. Any updates? @ElNounch

Issue #210, opened 2013-10-06
CryptoPP ASM is never turned on.


bearbin:
Did the makefile changes do this? When building with GCC, the CRYPTOPP-DISABLE-ASM flag is still there.

PR #3348, opened 2016-08-25
Added support for the Minecraft 1.10 protocol(#210)

### LINKED ISSUE #3229: Minecraft 1.10 Protocol Support
Now that Minecraft 1.10 is released (already), we should add protocol support for it.

[Bountysource](https://www.bountysource.com/issues/35062502-minecraft-1-10-protocol-support?utm_source=35062502&utm_medium=shield&utm_campaign=ISSUE_BADGE)

bibo38:
This PR implements the Minecraft 1.10 protocol.
Furthermore it adds an Metadata enum to replace the hard coded indexes.

Fixes #3229

### DIFF
--- a/src/Protocol/Protocol110x.cpp
+++ b/src/Protocol/Protocol110x.cpp
@@ -35,6 +35,15 @@ Implements the 1.10.x protocol classes:
 
 
 
+// The disabled error is intended, since the Metadata have overlapping indexes
+// based on the type of the Entity.
+//
+// IMPORTANT: The enum is used to automate the sequential counting of the
+// Metadata indexes. Adding a new enum value causes the following values to
+// increase their index. Therefore the ordering of the enum values is VERY important!
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wduplicate-enum"
+
 namespace Metadata
 {
 	enum Metadata_Index
@@ -263,6 +272,8 @@ namespace Metadata
 	};
 }
 
+#pragma clang diagnostic pop  // Restore ignored clang errors
+
 
 
 
@@ -385,14 +396,10 @@ void cProtocol1100::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_E
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
 
 			// The following expression makes Minecarts shake more with less health or higher damage taken
-			// It gets half the maximum health, and takes it away from the current health minus the half health:
-			/*
-			Health: 5 | 3 - (5 - 3) = 1 (shake power)
-			Health: 3 | 3 - (3 - 3) = 3
-			Health: 1 | 3 - (1 - 3) = 5
-			*/
 			auto & Minecart = reinterpret_cast<const cMinecart &>(a_Entity);
-			a_Pkt.WriteVarInt32((((a_Entity.GetMaxHealth() / 2) - (a_Entity.GetHealth() - (a_Entity.GetMaxHealth() / 2))) * Minecart.LastDamage()) * 4);
+			auto maxHealth = a_Entity.GetMaxHealth();
+			auto curHealth = a_Entity.GetHealth();
+			a_Pkt.WriteVarInt32(static_cast<UInt32>((maxHealth - curHealth) * Minecart.LastDamage() * 4));
 
 			a_Pkt.WriteBEUInt8(SHAKING_DIRECTION);  // (doesn't seem to effect anything)
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
@@ -412,11 +419,11 @@ void cProtocol1100::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_E
 					a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
 					int Content = MinecartContent.m_ItemType;
 					Content |= MinecartContent.m_ItemDamage << 8;
-					a_Pkt.WriteVarInt32(Content);
+					a_Pkt.WriteVarInt32(static_cast<UInt32>(Content));
 
 					a_Pkt.WriteBEUInt8(BLOCK_Y);
 					a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
-					a_Pkt.WriteVarInt32(RideableMinecart.GetBlockHeight());
+					a_Pkt.WriteVarInt32(static_cast<UInt32>(RideableMinecart.GetBlockHeight()));
 
 					a_Pkt.WriteBEUInt8(SHOW_BLOCK);
 					a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
@@ -562,7 +569,7 @@ void cProtocol1100::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob
 			auto & Creeper = reinterpret_cast<const cCreeper &>(a_Mob);
 			a_Pkt.WriteBEUInt8(CREEPER_STATE);  // (idle or "blowing")
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
-			a_Pkt.WriteVarInt32(Creeper.IsBlowing() ? 1 : -1);
+			a_Pkt.WriteVarInt32(Creeper.IsBlowing() ? 1 : static_cast<UInt32>(-1));
 
 			a_Pkt.WriteBEUInt8(CREEPER_POWERED);
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
@@ -580,7 +587,7 @@ void cProtocol1100::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob
 			a_Pkt.WriteBEUInt8(CARRIED_BLOCK);
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_BLOCKID);
 			UInt32 Carried = 0;
-			Carried |= Enderman.GetCarriedBlock() << 4;
+			Carried |= static_cast<UInt32>(Enderman.GetCarriedBlock() << 4);
 			Carried |= Enderman.GetCarriedMeta();
 			a_Pkt.WriteVarInt32(Carried);
 
@@ -633,18 +640,18 @@ void cProtocol1100::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mob
 
 			a_Pkt.WriteBEUInt8(HORSE_TYPE);
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
-			a_Pkt.WriteVarInt32(Horse.GetHorseType());
+			a_Pkt.WriteVarInt32(static_cast<UInt32>(Horse.GetHorseType()));
 
 			a_Pkt.WriteBEUInt8(HORSE_VARIANT);  // Color / style
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
 			int Appearance = 0;
 			Appearance = Horse.GetHorseColor();
 			Appearance |= Horse.GetHorseStyle() << 8;
-			a_Pkt.WriteVarInt32(Appearance);
+			a_Pkt.WriteVarInt32(static_cast<UInt32>(Appearance));
 
 			a_Pkt.WriteBEUInt8(ARMOR);
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_VARINT);
-			a_Pkt.WriteVarInt32(Horse.GetHorseArmour());
+			a_Pkt.WriteVarInt32(static_cast<UInt32>(Horse.GetHorseArmour()));
 
 			a_Pkt.WriteBEUInt8(BABY);
 			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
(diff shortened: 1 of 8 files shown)


Pokechu22:
+1 for the use of the enum, although Travis (more specifically clang) seems to be complaining about it:

> ```
> /home/travis/build/cuberite/cuberite/src/Protocol/Protocol110x.cpp:59:3: error: 
>       element AREAEFFECTCLOUD_COLOR has been implicitly assigned 7 which another
>       element has been assigned [-Werror,-Wduplicate-enum]
> ```

along with some signed conversion warnings being treated as errors (which also apply to Protocol19x).  You should be able to disable those errors by adding

```cmake
	set_source_files_properties(Protocol19x.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=old-style-cast -Wno-error=sign-conversion -Wno-error=conversion -Wno-error=switch-enum -Wno-error=switch -Wno-error=duplicate-enum")
```

to the `if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")` part of CMakeLists.

madmaxoft:
Do NOT disable warnings! Fix them!

src/Protocol/Protocol110x.cpp
```diff
@@ -0,0 +1,857 @@
...
+		// MinecartCommandBlock
+		COMMAND = _MINECART_NEXT,
+		LAST_OUTPUT,
+
+		// MinecartFurnace
+		MINECARTFURNACE_POWERED = _MINECART_NEXT,
+
+		// TNTPrimed
+		FUSE_TIME = _ENTITY_NEXT,
+	};
```
madmaxoft: I think the enums should have a prefix specifying the entity they relate to, and they should all be explicitly assigned to numbers - without the `_NEXT`.

src/Protocol/Protocol110x.cpp
```diff
@@ -0,0 +1,857 @@
...
+void cProtocol1100::SendSoundEffect(const AString & a_SoundName, double a_X, double a_Y, double a_Z, float a_Volume, float a_Pitch)
+{
+	ASSERT(m_State == 3);  // In game mode?
+
+	cPacketizer Pkt(*this, 0x19);  // Named sound effect packet
+	Pkt.WriteString(a_SoundName);
+	Pkt.WriteVarInt32(0);  // Master sound category (may want to be changed to a parameter later)
+	Pkt.WriteBEInt32(static_cast<Int32>(a_X * 8.0));
+	Pkt.WriteBEInt32(static_cast<Int32>(a_Y * 8.0));
+	Pkt.WriteBEInt32(static_cast<Int32>(a_Z * 8.0));
```
madmaxoft: Use `FloorC` instead of `static_cast`

src/Protocol/Protocol110x.cpp
```diff
@@ -0,0 +1,857 @@
...
+
+			// The following expression makes Minecarts shake more with less health or higher damage taken
+			// It gets half the maximum health, and takes it away from the current health minus the half health:
+			/*
+			Health: 5 | 3 - (5 - 3) = 1 (shake power)
+			Health: 3 | 3 - (3 - 3) = 3
+			Health: 1 | 3 - (1 - 3) = 5
+			*/
+			auto & Minecart = reinterpret_cast<const cMinecart &>(a_Entity);
+			a_Pkt.WriteVarInt32((((a_Entity.GetMaxHealth() / 2) - (a_Entity.GetHealth() - (a_Entity.GetMaxHealth() / 2))) * Minecart.LastDamage()) * 4);
```
madmaxoft: I have fixed this code in the 1.9 protocol in the FixWarnings branch, you might want to do the same.

src/Protocol/Protocol110x.cpp
bibo38: If you wish a prefix, than I add one. I just wanted to keep the names as short as possible.

But I disagree with the explicit number assignment. Look at the [Metadata changes](http://wiki.vg/index.php?title=Pre-release_protocol&oldid=7972#Entity_Metadata)! `Entity` got a new Metadata(No Gravity) and every descendant of `Entity` increased their index values to be sequential again.
With explicit numbering you need to change **all** Indexes(and make no mistake :smirk: )! Using this enum construct, you only need to add the new value! Also you can see the (Minecraft) class hierachy. E.g. MinecartFurnace -> Minecart -> Entity.

bibo38:
> Do NOT disable warnings! Fix them!

In the case of the signed conversion warnings, I agree!

But I would classify the duplicate enum warnings as intended. Fixing them would require not to use enums, which would look much more cluttered:

    const static UInt8 AIR = 1;
    const static UInt8 CUSTOM_NAME = AIR + 1;
    const static UInt8 CUSTOM_NAME_VISIBLE = CUSTOM_NAME + 1;
    [...]

Furthermore you need to change two lines instead of one line for the same effect as the enum(variables with overlapping values in a namespace).

The "Error" of Clang should prevent this scenario:

    enum Weather {
    SUNNY, RAINY = 0
    }
    [...]
    int curWeather = SUNNY;
    if (curWeather == RAINY)
        // Always true. Find the bug!

But in this case the enum is only required to have literals instead of hardcoded numbers. So I would justify this as an false error.

src/Protocol/Protocol110x.cpp
bibo38: Sorry, but I don't find the FixWarnings branch. I have looked at [all branches](https://github.com/cuberite/cuberite/branches/all) , but no FixWarnings branch appeared...

Seadragon91: It has already been merged into master.

madmaxoft:
As for the duplicate enums, when they are explicitly assigned their value (which I requested as well), the clang error should disappear.

src/Protocol/Protocol110x.cpp
madmaxoft: Yes, but that change created a completely new protocol (and hence, for us, a completely new enum), so there's no point in having them auto-assigned if they are copied and edited anyway. Having the compiler assign numbers to enums is a sure way to introduce bugs later on when someone edits the enum without realizing this.

mathiascode:
Whenever a player enters your visibility range, Minecraft crashes. Here's an example:
```
---- Minecraft Crash Report ----
// Would you like a cupcake?

Time: 8/25/16 5:57 PM
Description: Ticking entity

java.lang.ClassCastException: java.lang.Float cannot be cast to java.lang.Byte
	at sf.cx(SourceFile:2000)
	at sf.cz(SourceFile:2008)
	at sf.m(SourceFile:1595)
	at zs.m(SourceFile:272)
	at bno.m(SourceFile:55)
	at aid.a(SourceFile:1527)
	at aid.g(SourceFile:1505)
	at aid.k(SourceFile:1388)
	at bcx.t(SourceFile:1652)
	at bcx.av(SourceFile:968)
	at bcx.a(SourceFile:396)
	at net.minecraft.client.main.Main.main(SourceFile:124)


A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Thread: Client thread
Stacktrace:
	at sf.cx(SourceFile:2000)
	at sf.cz(SourceFile:2008)
	at sf.m(SourceFile:1595)
	at zs.m(SourceFile:272)
	at bno.m(SourceFile:55)
	at aid.a(SourceFile:1527)
	at aid.g(SourceFile:1505)

-- Entity being ticked --
Details:
	Entity Type: null (bno)
	Entity ID: 543
	Entity Name: AshleyPlayzGames
	Entity's Exact location: 960.50, 6.00, 639.50
	Entity's Block location: World: (960,6,639), Chunk: (at 0,0,15 in 60,39; contains blocks 960,0,624 to 975,255,639), Region: (1,1; contains chunks 32,32 to 63,63, blocks 512,0,512 to 1023,255,1023)
	Entity's Momentum: 0.00, 0.00, 0.00
	Entity's Passengers: []
	Entity's Vehicle: ~~ERROR~~ NullPointerException: null
Stacktrace:
	at aid.k(SourceFile:1388)

-- Affected level --
Details:
	Level name: MpServer
	All players: 2 total; [bnn['MissingTexture'/539, l='MpServer', x=960.49, y=6.00, z=634.42], bno['AshleyPlayzGames'/543, l='MpServer', x=960.50, y=6.00, z=639.50]]
	Chunk stats: MultiplayerChunkCache: 25, 25
	Level seed: 0
	Level generator: ID 00 - default, ver 1. Features enabled: false
	Level generator options: 
	Level spawn location: World: (32,64,-5), Chunk: (at 0,4,11 in 2,-1; contains blocks 32,0,-16 to 47,255,-1), Region: (0,-1; contains chunks 0,-32 to 31,-1, blocks 0,0,-512 to 511,255,-1)
	Level time: 14448 game time, 15448 day time
	Level dimension: 0
	Level storage version: 0x00000 - Unknown?
	Level weather: Rain time: 0 (now: true), thunder time: 0 (now: false)
	Level game mode: Game mode: creative (ID 1). Hardcore: false. Cheats: false
	Forced entities: 3 total; [bnn['MissingTexture'/539, l='MpServer', x=960.49, y=6.00, z=634.42], bno['AshleyPlayzGames'/543, l='MpServer', x=960.50, y=6.00, z=639.50], bno['AshleyPlayzGames'/543, l='MpServer', x=960.50, y=6.00, z=639.50]]
	Retry entities: 0 total; []
	Server brand: Cuberite
	Server type: Non-integrated multiplayer server
Stacktrace:
	at bln.a(SourceFile:364)
	at bcx.b(SourceFile:2535)
	at bcx.a(SourceFile:405)
	at net.minecraft.client.main.Main.main(SourceFile:124)
```

src/Protocol/Protocol110x.cpp
Pokechu22: Except that the enum means that people don't have to change 200 different constants, which is what took 1.10 protocol so long to be written in the first place.  Auto-assigning them means that it's easier to fix later, and _they're auto-assigned in a similar way in vanilla minecraft anyways_.

Pokechu22:
> Do NOT disable warnings! Fix them!

My rationale for suggesting disabling them was that A) the first set of warnings are already disabled for the other protocols (so copying the rest of the code for the protocol basically includes copying those warnings), and B) for `-Wno-error=duplicate-enum` necessary for the current pattern used by the code (which IMO is a better pattern than hardcoding the IDs).  While it would be a good idea to fix A) (which can be done via the protocol changes from #3346 / d2e8643), B) isn't really possible to fix and that should remain disabled.

> Whenever a player enters your visibility range, Minecraft crashes. Here's an example:

[Hopper](http://hopper.minecraft.net/crashes/search/) suggests that that is a new crash.  Congratulations!  You've broken the game in a brand new way.

More seriously, assuming that's from 1.10.2, I can use MCP to say that the topmost method of that stacktrace `net.minecraft.entity.EntityLivingBase.isHandActive(SourceFile:2000)` and it's happening while checking the HAND_STATES metadata field (6), where it's expecting a byte and getting a float.  That's odd since you're not writing that field, and that field is defaulted with a Byte value of 0.  I'm checking what's happening internally right now.

bibo38:
@Pokechu22 The problem of this bug is (again), that the `SendPlayerSpawn` method creates it's metadata without using `SendEntityMetadata`. Therefore it uses the old 1.9 hardcoded values and breaks the client.
This problem has driven me crazy, when the client received the wrong metadata for Pickups, but my enum had the correct values :rage: . #codeduplication is bad

@mathiascode Thanks for testing. As noted above the Client crash is caused by some code duplication, which I've missed.
The sound problem is caused by a datatype change, which also changed the range of valid values.

src/Protocol/Protocol110x.cpp
bibo38: I agree to @Pokechu22 ! I need to copy the whole enum for the new protocol, but I only need to add the few new Metadata entries. This is the point of auto-assignment!

> Having the compiler assign numbers to enums is a sure way to introduce bugs later on when someone edits the enum without realizing this.

There are two scenarios, when somebody edits this enum:

- My enum is incorrect: Then the person fixing this knows, how the enum works, because it created the issue.

- A new Protocol is released: Then the implementing person can miss this. But there are multiple people reviewing the new protocol. Also people like @mathiascode test the new changes. The probability of a bug is rare.

And an occurring bug is easy to fix. Better having a bugged new protocol than a heavily delayed new protocol(Just look at the issue request date :smirk: ).

bibo38:
@mathiascode Can you check, if the bugs are fixed in the new branch?

mathiascode:
Yes, the bugs are fixed now. Thanks!

bibo38:
@worktycho @tigerw @sphinxc0re @LogicParrot @madmaxoft I will implement the enum prefixes soon. Is there anything else blocking this commit from merging into master?

LogicParrot:
>Edit: Now this should be mergeable!

The branch has to be rebased again because `master` was updated. I thought someone enabled doing this through the web at some point?

bibo38:
@LogicParrot I've just rebased the PR.

LogicParrot:
> Is there anything else blocking this commit from merging into master?

@madmaxoft 's green light, I assume.

src/Protocol/Protocol110x.cpp
LogicParrot: Ping @bibo38

bibo38: ty. Just fixed it.

mathiascode:
I've been using this for about a week now, and I haven't noticed any new bugs.

bearbin:
@madmaxoft merge then?

LogicParrot:
I did a squash-merge on accident. That made history a bit unclean. Sorry.

LogicParrot:
Please ignore my previous comment. Nothing is unclean. I got confused.

src/Protocol/Protocol110x.cpp
```diff
@@ -0,0 +1,879 @@
...
+
+
+// The disabled error is intended, since the Metadata have overlapping indexes
+// based on the type of the Entity.
+//
+// IMPORTANT: The enum is used to automate the sequential counting of the
+// Metadata indexes. Adding a new enum value causes the following values to
+// increase their index. Therefore the ordering of the enum values is VERY important!
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wduplicate-enum"
```
madmaxoft: These two pragmas trigger a warning in MSVC, they should have been in an `#ifdef` block

PR #3350, opened 2016-08-25
Changing world send a Respawn packet

ElNounch:
Correct visual glitches (seeing previous world from new coordinates point of view until refresh)
Handling Respawning in current dimension

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1800,6 +1800,11 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	// Send the respawn packet:
 	if (a_ShouldSendRespawn && (m_ClientHandle != nullptr))
 	{
+		// Respawning in current dimension is illegal if wasn't dead, let's tour a bit...
+		if (a_World->GetDimension() == GetWorld()->GetDimension())
+		{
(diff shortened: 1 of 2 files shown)


src/Entities/Player.cpp
```diff
@@ -1800,6 +1800,11 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	// Send the respawn packet:
 	if (a_ShouldSendRespawn && (m_ClientHandle != nullptr))
 	{
+		// Respawning in current dimension is illegal if wasn't dead, let's tour a bit...
```
LogicParrot: I would add a more descriptive comment so that people in the future would understand this. Something like:
```
When teleporting to a world of the same dimension as the current world, we must first send the respawn packet of some other dimension, otherwise the client will have visual issues. The real respawn packet is sent immediately afterwards below.
```

LogicParrot:
Are there any cases where we call `MoveToWorld` with `false`?

LogicParrot:
Do not merge yet. This does not resolve visual glitches when portal traveling. I'll post a followup PR based on this.

ElNounch:
Aren't you suffering from lack of ChunkStay instead ?

LogicParrot:
@ElNounch Nope. cEntity's `scanPortal` had hardwired dimension types. See #3351.

LogicParrot:
For instance, #3352 is not resolved by this but is resolved by #3351.

PR #3351, opened 2016-08-27
Proper respawn packets on dimension travel

### LINKED ISSUE #2405: Walls of random blocks in loading chunks

### LINKED ISSUE #2547: Teleportation - Server sends wrong chunks

### LINKED ISSUE #3352: Nether world is corrupt clientside when traveling from nether to nether
Client version:   Only 1.9.x is affected
Server OS:        All operating systems are effected
Commit id:        d52afadad491474f63e8184cb5a08c0ca21ab48c

### Expected behavior
Proper Nether world when teleporting from Nether to Nether.

### Actual behavior
Nether has random blocks and chunks everywhere.

Setup up two nether worlds A and B, and make sure that A's portal leads to B, by editing `A/world.ini` and writing: `OverworldName=B`. Now build a nether portal in A and step in.

LogicParrot:
Closes #2547
Closes #2405
Closes #3352
Replaces #3350
Replaces #3114
Completes the "[Stable world travel](https://github.com/cuberite/cuberite/milestone/20?closed=1)" milestone :)

Minecraft clients do not like respawning to the same dimension, unless the player dies. This PR takes this into account when teleporting between dimensions. If one is teleporting from a world of dimension A to a world of dimension A, a dummy respawn packet of some dimension B is sent first, and the real respawn packet of dimension A is sent afterwards.

Also, the cEntity teleportation code used to falsely assume that a linked world's dimension always has correspondance to the portal type and the current world. For instance, it assumed that when stepping into a Nether portal from the Overworld, one always ends up in a Nether. This caused the wrong respawn packets to be sent. The PR fixes this as well.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -62,6 +62,7 @@ int cClientHandle::s_ClientCount = 0;
 // cClientHandle:
 
 cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
+	m_LastSentDimension(dimNotSet),
 	m_CurrentViewDistance(a_ViewDistance),
 	m_RequestedViewDistance(a_ViewDistance),
 	m_IPString(a_IPString),
@@ -368,6 +369,7 @@ void cClientHandle::Authenticate(const AString & a_Name, const AString & a_UUID,
 
 	// Return a server login packet
 	m_Protocol->SendLogin(*m_Player, *World);
+	m_LastSentDimension = World->GetDimension();
 
 	// Send Weather if raining:
 	if ((World->GetWeather() == 1) || (World->GetWeather() == 2))
@@ -2704,7 +2706,21 @@ void cClientHandle::SendResetTitle()
 
 void cClientHandle::SendRespawn(eDimension a_Dimension, bool a_ShouldIgnoreDimensionChecks)
 {
-	m_Protocol->SendRespawn(a_Dimension, a_ShouldIgnoreDimensionChecks);
+	// If a_ShouldIgnoreDimensionChecks is true, we must be traveling to the same dimension
+	ASSERT((!a_ShouldIgnoreDimensionChecks) || (a_Dimension == m_LastSentDimension));
+
+	if ((!a_ShouldIgnoreDimensionChecks) && (a_Dimension == m_LastSentDimension))
+	{
+		// The client goes crazy if we send a respawn packet with the dimension of the current world
+		// So we send a temporary one first.
+		// This is not needed when the player dies, hence the a_ShouldIgnoreDimensionChecks flag.
+		// a_ShouldIgnoreDimensionChecks is true only at cPlayer::respawn, which is called after
+		// the player dies.
+		eDimension TemporaryDimension = (a_Dimension == dimOverworld) ? dimNether : dimOverworld;
+		m_Protocol->SendRespawn(TemporaryDimension);
+	}
+	m_Protocol->SendRespawn(a_Dimension);
+	m_LastSentDimension = a_Dimension;
 }
 
 
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1433,19 +1433,22 @@ bool cEntity::DetectPortal()
 				}
 				m_PortalCooldownData.m_TicksDelayed = 0;
 
+				// Nether portal in the nether
 				if (GetWorld()->GetDimension() == dimNether)
 				{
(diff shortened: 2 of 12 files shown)


src/ClientHandle.cpp
```diff
@@ -2704,7 +2704,43 @@ void cClientHandle::SendResetTitle()
...
+
+	eDimension CurrentDimension = m_Player->GetWorld()->GetDimension();
+	if ((!a_ShouldIgnoreDimensionChecks) && (a_Dimension == CurrentDimension))
+	{
+		// The client goes crazy if we send a respawn packet with the dimension of the current world
+		// So we send a temporary one first
+		// This is not needed when the player dies, hence the a_ShouldIgnoreDimensionChecks flag.
+		eDimension TemporaryPacket = (a_Dimension == dimOverworld) ? dimNether : dimOverworld;
+		// LogRespawn(TemporaryPacket);
+		m_Protocol->SendRespawn(TemporaryPacket);
```
LogicParrot: Temporary Dimension

src/Entities/Entity.cpp
```diff
@@ -1454,23 +1457,33 @@ bool cEntity::DetectPortal()
...
+					eDimension DestionationDim = DestinationWorld->GetDimension();
 
 					m_PortalCooldownData.m_ShouldPreventTeleportation = true;
 
 					if (IsPlayer())
 					{
-						reinterpret_cast<cPlayer *>(this)->AwardAchievement(achEnterPortal);
-						reinterpret_cast<cPlayer *>(this)->GetClientHandle()->SendRespawn(dimNether);
+
+
```
LogicParrot: Spaces

src/ClientHandle.cpp
```diff
@@ -2704,7 +2704,20 @@ void cClientHandle::SendResetTitle()
 
 void cClientHandle::SendRespawn(eDimension a_Dimension, bool a_ShouldIgnoreDimensionChecks)
 {
-	m_Protocol->SendRespawn(a_Dimension, a_ShouldIgnoreDimensionChecks);
+	eDimension CurrentDimension = m_Player->GetWorld()->GetDimension();
```
LogicParrot: Should I use m_LastSentDimension instead? Or should I simply remove `m_LastSentDimension`?

LogicParrot:
This has been tested ingame with all possible teleport permutations (nether>nether, nether>overworld, etc.), I teleported both via portals and the Essentials `/portal` command. All packets were sent as expected, (Only end portals weren't tested).

src/ClientHandle.cpp
```diff
@@ -2704,7 +2706,23 @@ void cClientHandle::SendResetTitle()
...
+		// The client goes crazy if we send a respawn packet with the dimension of the current world
+		// So we send a temporary one first.
+		// This is not needed when the player dies, hence the a_ShouldIgnoreDimensionChecks flag.
+		// a_ShouldIgnoreDimensionChecks is true only at cPlayer::respawn, which is called after
+		// the player dies.
+		eDimension TemporaryDimension = (a_Dimension == dimOverworld) ? dimNether : dimOverworld;
+		// LOG("################## Respawn: %s", DimensionToString(TemporaryDimension).c_str());
+		m_Protocol->SendRespawn(TemporaryDimension);
+	}
+	// LOG("################## Respawn: %s", DimensionToString(a_Dimension).c_str());
```
madmaxoft: Remove the logging?

madmaxoft:
I'm considering whether the check for the same dimension should be implemented in `cClientHandle`, as you've done it, or in `cProtocol` descendants. After all, it is a thing of the protocol; if the client ever fixes the "bug" with travelling to the same dimension, we will no longer need the check for that version of the protocol. On the other hand, it means copying the code over to each new `cProtocol`.

LogicParrot:
As of now, all protocol versions have this bug. We can trivially move this to `cProtocol` and duplicate it if ever needed.

src/ClientHandle.cpp
LogicParrot: Done.

PR #3356, opened 2016-08-28
Breaking ice in survival now gives a spreading water's source

### LINKED ISSUE #3355: Broken ice blocks reappear after logging out and in again
Client version:   1.9
Server OS:        Raspbian  Jessie (Raspberri Pi 3 B)
Commit id:         d52afadad491474f63e8184cb5a08c0ca21ab48c

Ice blocks reappear after being broken and replaced with other materials. I have already tried to issue `save-all` before logging out, and restarting the server.

As the only player in a server, place an ice block somewhere, break it and place a stone block in the same spot. 

Log out and back in again. 

The ice block appears instead of the stone block. This also works with other blocks, including lava.

ElNounch:
Also fix #3355 (breaking ice blocks not really destroying it)

### DIFF
--- a/src/Blocks/BlockIce.h
+++ b/src/Blocks/BlockIce.h
@@ -38,8 +38,7 @@ public:
 				return;
 			}
 
-			a_ChunkInterface.FastSetBlock(a_BlockX, a_BlockY, a_BlockZ, E_BLOCK_WATER, 0);
-			// This is called later than the real destroying of this ice block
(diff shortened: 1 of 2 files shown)


Cl1608Ho:
Does this also cover the SilkTouch enchantment?

ElNounch:
"If it ain't broke, don't fix it." :tm:

Looks like Silk touch works for ice, any matter ?

PR #3151, opened 2016-04-19
Smart chunk unloading

LogicParrot:
Work in progress. This is a smarter replacement for #3142

- The server has  a `MaxChunkRAM` setting, configurable in `settings.ini` (default: 30 MiB)
- The server has a `MinimizeRam` setting, configurable in `settings.ini` (default: 1)

**Terminology:**
Dirtiness:
- Clean chunk: No changes were made, no need to save to disk before unloading.
- Dirty chunk: A chunk which underwent changes. Must save to disk before unloading.

Usage:

- Unused chunk: A chunk with no players (nor chunkStay, etc.) nearby, the server may unload this.
- Used chunk: A chunk which is actively used by the game. The server must not unload this.

Other:

- Saving a chunk: Turns dirty chunks into clean chunks by saving them to disk. After a save, the chunk is no longer dirty and the server may unload it.
- Unloading a chunk: Removing a chunk from RAM and freeing that RAM. We can only unload chunks that are both unused and clean.

The server has two saving/unloading strategies.

**Strategy 1: `MinimizeRam` is 0:**
 The server is RAM-greedy, and uses all the RAM specified in `MaxChunkRAM` as a chunk cache in order to improve performance. Chunks are saved/unloaded as late as possible. Disk I/O is significantly reduced. Performance is improved. RAM usage is typically static and will not go below or above  `MaxChunkRAM`.
 - Dirty chunks are saved in 5 minute intervals to avoid data loss in case of a crash.
 - If current RAM usage exceeds `MaxChunkRAM`, a minimum amount of clean unused chunks is unloaded in order to preserve the `MaxChunkRAM` limit. The chunks are chosen randomly.
 - If there are no clean unused chunks left to unload and `MaxChunkRAM` still cannot be maintained, a save cycle is forced, allowing more chunks to be unloaded.
 - Other than the above three scenarios, chunks are never unloaded or saved.
 - The amount of RAM used by chunks is normally static and equals `MaxChunkRAM`
 - If the activity is too high and the `MaxChunkRAM` limit cannot be maintained even after all measures were taken, the "agressive" strategy is used. This happens when the RAM required by the **used chunks** exceeds `MaxChunkRAM`.

**Strategy 2: `MinimizeRam` is 1:** (Might remove this strategy)
 The server is RAM-economic, attempts to never reach `MaxChunkRAM`. Clean unused Chunks are unloaded as soon as possible.
 - Dirty chunks are saved in 5 minute intervals to avoid data loss in case of a crash.
[...]

PR #3359, opened 2016-08-29
Configurable dirty unused chunk cap to avoid RAM overuse

LogicParrot:
A simpler alternative to https://github.com/cuberite/cuberite/pull/3151 which resolves Raspberrypi chokes when flying too fast.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -2704,11 +2704,28 @@ void cChunkMap::SaveAllChunks(void)
 
 
 
-int cChunkMap::GetNumChunks(void)
+size_t cChunkMap::GetNumChunks(void)
 {
 	cCSLock Lock(m_CSChunks);
-	return static_cast<int>(m_Chunks.size());  // TODO: change return value to unsigned type
+	return m_Chunks.size();
+}
 
+
+
+
+
+size_t cChunkMap::GetNumUnusedDirtyChunks(void)
+{
+	cCSLock Lock(m_CSChunks);
+	size_t res = 0;
+	for (const auto & Chunk : m_Chunks)
+	{
+		if (Chunk.second->IsValid() && Chunk.second->CanUnloadAfterSaving())
+		{
+			res += 1;
+		}
+	}
+	return res;
 }
 
 
--- a/src/ChunkMap.h
+++ b/src/ChunkMap.h
@@ -388,7 +388,10 @@ public:
 
 	cWorld * GetWorld(void) { return m_World; }
 
-	int GetNumChunks(void);
+	size_t GetNumChunks(void);
+
+	/** Returns the number of unused dirty chunks. Those are chunks that we can save and then unload */
+	size_t GetNumUnusedDirtyChunks(void);
 
 	void ChunkValidated(void);  // Called by chunks that have become valid
 
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -453,7 +453,7 @@ void cWorld::Start(void)
 	// The presence of a configuration value overrides everything
 	// If no configuration value is found, GetDimension() is written to file and the variable is written to again to ensure that cosmic rays haven't sneakily changed its value
 	m_Dimension = StringToDimension(IniFile.GetValueSet("General", "Dimension", DimensionToString(GetDimension())));
-	int UnusedDirtyChunksCap = IniFile.GetValueSetI("General", "UnusedChunkCap", 500);
+	int UnusedDirtyChunksCap = IniFile.GetValueSetI("General", "UnusedChunkCap", 1000);
 	if (UnusedDirtyChunksCap < 0)
 	{
 		UnusedDirtyChunksCap *= -1;
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -149,7 +149,7 @@ cWorld::cWorld(const AString & a_WorldName, eDimension a_Dimension, const AStrin
 	m_WorldAge(0),
 	m_TimeOfDay(0),
 	m_LastTimeUpdate(0),
-	m_LastUnload(0),
+	m_LastChunkCheck(0),
 	m_LastSave(0),
 	m_SkyDarkness(0),
 	m_GameMode(gmNotSet),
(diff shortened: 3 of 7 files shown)


LogicParrot:
TODO remove log() calls and add book documentation. @tigerw , what says you?

src/ChunkMap.h
```diff
@@ -390,6 +390,9 @@ class cChunkMap
 
 	int GetNumChunks(void);
 
+	/** Returns the number of unused dirty chunks. Those are chunks that we can save and then unload */
+	int GetNumUnusedDiryChunks(void);
```
tigerw: Diry me, this spelling is most peculiar.

src/World.cpp
```diff
@@ -1059,12 +1060,23 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
...
 	if (m_WorldAge - m_LastUnload > std::chrono::seconds(10))  // Unload every 10 seconds
 	{
+		LOG("######## %d / %d", GetNumUnusedDiryChunks(), m_UnusedDirtyChunksCap);
 		UnloadUnusedChunks();
+		if (m_WorldAge - m_LastSave > std::chrono::seconds(30))
+		{
+			if (GetNumUnusedDiryChunks() > m_UnusedDirtyChunksCap)
+			{
+				LOG("######## CAP SAVE");
+				SaveAllChunks();
```
tigerw: Does this also unload chunks? If not, you could move it before the UnloadUnusedChunks call to do everything together.

src/World.cpp
```diff
@@ -1059,12 +1060,23 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
...
+		LOG("######## PERIODIC SAVE");
 		SaveAllChunks();
 	}
 
+
 	if (m_WorldAge - m_LastUnload > std::chrono::seconds(10))  // Unload every 10 seconds
 	{
+		LOG("######## %d / %d", GetNumUnusedDiryChunks(), m_UnusedDirtyChunksCap);
 		UnloadUnusedChunks();
+		if (m_WorldAge - m_LastSave > std::chrono::seconds(30))
```
tigerw: Necessary? I would assume that you would want to enforce the cap as quickly as possible.

src/ChunkMap.cpp
```diff
@@ -2715,6 +2715,23 @@ int cChunkMap::GetNumChunks(void)
 
 
 
+int cChunkMap::GetNumUnusedDiryChunks(void)
+{
+	int res = 0;
+	for (const auto & Chunk : m_Chunks)
```
madmaxoft: You need to lock `m_CSChunks` in order to iterate over `m_Chunks`.

src/World.cpp
```diff
@@ -1059,12 +1060,23 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
 
 	if (m_WorldAge - m_LastSave > std::chrono::minutes(5))  // Save each 5 minutes
 	{
+		LOG("######## PERIODIC SAVE");
```
madmaxoft: Don't log this

src/World.cpp
```diff
@@ -1059,12 +1060,23 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
...
 		SaveAllChunks();
 	}
 
+
 	if (m_WorldAge - m_LastUnload > std::chrono::seconds(10))  // Unload every 10 seconds
 	{
+		LOG("######## %d / %d", GetNumUnusedDiryChunks(), m_UnusedDirtyChunksCap);
```
madmaxoft: Remove the logging

src/World.cpp
madmaxoft: We don't want to check on each tick, though. I think a time limit like this is reasonable, although I'd go for just a few (10) seconds.

src/World.cpp
```diff
@@ -1059,12 +1060,23 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
...
 		UnloadUnusedChunks();
+		if (m_WorldAge - m_LastSave > std::chrono::seconds(30))
+		{
+			if (GetNumUnusedDiryChunks() > m_UnusedDirtyChunksCap)
+			{
+				LOG("######## CAP SAVE");
```
madmaxoft: Remove the logging.

src/World.cpp
madmaxoft: Doesn't matter, the saving happens asynchronously on another thread, so the chunks still wouldn't get unloaded.

src/World.h
```diff
@@ -685,6 +685,9 @@ class cWorld :
 	/** Returns the number of chunks loaded	 */
 	int GetNumChunks() const;  // tolua_export
 
+	/** Returns the number of unused dirty chunks. That's the number of chunks that would be unloaded shortly after calling "SaveAllChunks" */
+	int GetNumUnusedDiryChunks(void);
```
madmaxoft: How about exporting this to Lua, too?

src/World.cpp
tigerw: It is enclosed in such a block though.

src/World.cpp
tigerw: daayum.

src/World.cpp
madmaxoft: Right, didn't see that. Then it is indeed needless IMO.

src/World.h
```diff
@@ -851,6 +854,8 @@ class cWorld :
 	} ;
 
 
+	int m_UnusedDirtyChunksCap;
```
madmaxoft: Add Doxy-comment.

src/World.cpp
LogicParrot: The outer 10 sec block does not guarantee 10 sec spacing between all saves.

If a periodic save happens right before the 10 second block is entered, another save would immediately follow.

I guess this is a bit confusing because I hijacked the `m_LastUnload` variable for saving.

src/World.h
tigerw: size_t? Or unsigned?

LogicParrot: I exported. Do I need to manually modify `APIDesc`? Or have we got the automatic doxy import idea implemented?

LogicParrot:
I exported. Do I need to manually modify APIDesc? Or have we got the automatic doxy import idea implemented?

madmaxoft:
You need to add the API documentation

The automatic doxy-to-docs process is non-incremental - it cannot add to already existing documentation, so it is a semi-manual process. One needs to run the `src/Bindings/DiffAPIDesc.lua` script and then manually copy the reported new symbols to APIDesc.lua or the relevant file.

src/World.cpp
```diff
@@ -1057,16 +1064,20 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
...
+		// Save each 5 minutes
+		if (m_WorldAge - m_LastSave > std::chrono::minutes(5))
+		{
+			SaveAllChunks();
+		}
+		// Save if we have too many dirty unused chunks
+		else if (GetNumUnusedDiryChunks() > m_UnusedDirtyChunksCap)
+		{
+			SaveAllChunks();
+		}
```
tigerw: *Really* minor point, but don't the comments normally go inside the blocks and not between them? :)

madmaxoft: I'd prefer the comments inside the block, too; especially when it's visually breaking the `else if` away.

LogicParrot:
All done.

tigerw:
I'm confused about how and where you're getting both dirty *and* unused headwear for chunks. Do they come from the factory already grimy? Or are they clearance stock? Or what?

Besides, considering what we put them through, shouldn't our chunks get something more, stylish? I'm sure Bountysource can provide for something nicer.

LogicParrot:
Lol! Someone *must* post this on [/r/ProgrammerHumor/](https://www.reddit.com/r/ProgrammerHumor/) !

tigerw:
U wot?! U think i'm havin a laugh m8?

LogicParrot:
Should I multiply the cap by the number of online players?

tigerw:
Well. If you're not going to take the welfare of your workers seriously, then there's not much more that I can say.

----

In all seriousness, no complaints :)

PR #3361, opened 2016-08-29
cChunk::SetAlwaysTicked implies cChunk::stay

LogicParrot:
It makes more sense if always ticking chunks are always loaded too.


LogicParrot:
Does anyone disagree with this change?

tigerw:
Apparently not :P

src/Chunk.cpp
```diff
@@ -1715,10 +1715,12 @@ void cChunk::SetAlwaysTicked(bool a_AlwaysTicked)
 	if (a_AlwaysTicked)
 	{
 		m_AlwaysTicked += 1;
+		Stay(a_AlwaysTicked);
```
tigerw: So is m_AlwaysTicked now equivalent to m_StayCount? Or is there something preventing chunks from being ticked? I'm presuming the latter.

LogicParrot: `m_AlwaysTicked` is always smaller or equal to `m_StayCount`. A chunk may have chunkStay without being always ticked.

LogicParrot: In other words, the call to `Stay` may either be made independently, or implicitly through a call to `SetAlwaysTicked`. It's simply weird to want an always ticking but not always loaded chunk.

LogicParrot: In other words, the call to `Stay` may either be made independently, or implicitly through a call to `SetAlwaysTicked`. It's simply weird to want an always ticking but not always loaded chunk. The other way around is still possible.

madmaxoft:
Weird thing that this has broken the APIDump documentation and the plugin didn't complain in CI.
Fixed in #3400

madmaxoft:
Oh, it failed to even load the APIDump plugin. But the CI won't mark that as an error. That needs to be fixed, too.

PR #3362, opened 2016-08-31
Add Lua API export for titles.

mohe2015:
Fixes https://github.com/cuberite/cuberite/issues/1806
The /title command should be implemented in the Core plugin.


mohe2015:
Fixes https://github.com/cuberite/cuberite/issues/1806

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -596,7 +596,14 @@ end
...
 				Kick = { Params = "Reason", Return = "", Notes = "Kicks the user with the specified reason" },
+        SendHideTitle = { Params = "", Return = "", Notes = "Sends the client to hide the title." },
 				SendPluginMessage = { Params = "Channel, Message", Return = "", Notes = "Sends the plugin message on the specified channel." },
+        SendResetTitle = { Params = "", Return = "", Notes = "Sends the client to reset the title." },
+        SendSetSubTitle = { Params = "SubTitle", Return = "", Notes = "Sends the client the subtitle." },
+        SendSetRawSubTitle = { Params = "RawSubTitle", Return = "", Notes = "Sends the client the raw subtitle." },
+        SendSetTitle = { Params = "Title", Return = "", Notes = "Sends the client the title." },
+        SendSetRawTitle = { Params = "RawTitle", Return = "", Notes = "Sends the client the raw title." },
+        SendTitleTimes = { Params = "FadeInTicks, DisplayTicks, FadeOutTicks", Return = "", Notes = "Sends the client the title times and shows the title." },
```
madmaxoft: Use tabs to indent

madmaxoft: The wording is unfortunate.
`Send X to Y` is better.
Still, if I remember correctly, the title functions are a little bit unintuitive in that they don't cause the title to be displayed immediately, perhaps that should be documented as well and a better phrasing should be used to indicate that.

mohe2015:
Do you have an idea for a better phrasing? I'm not so good in English.

madmaxoft:
I'll try in a moment, feel free to discuss before changing the code - my English is not so good either.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -596,7 +596,14 @@ end
 				HasPluginChannel = { Params = "ChannelName", Return = "bool", Notes = "Returns true if the client has registered to receive messages on the specified plugin channel." },
 				IsUUIDOnline = { Params = "UUID", Return = "bool", IsStatic = true, Notes = "Returns true if the UUID is generated by online auth, false if it is an offline-generated UUID. We use Version-3 UUIDs for offline UUIDs, online UUIDs are Version-4, thus we can tell them apart. Accepts both 32-char and 36-char UUIDs (with and without dashes). If the string given is not a valid UUID, returns false."},
 				Kick = { Params = "Reason", Return = "", Notes = "Kicks the user with the specified reason" },
+				SendHideTitle = { Params = "", Return = "", Notes = "Sends to hide the title." },
```
madmaxoft: `Hides the title.` should be enough. I assume this does actually hide the title on the client. Does it hide the subtitle as well?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -596,7 +596,14 @@ end
...
 				SendPluginMessage = { Params = "Channel, Message", Return = "", Notes = "Sends the plugin message on the specified channel." },
+				SendResetTitle = { Params = "", Return = "", Notes = "Sends to reset the title." },
```
madmaxoft: What exactly is the difference between `Hide` and `Reset`?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -596,7 +596,14 @@ end
...
+				SendSetSubTitle = { Params = "SubTitle", Return = "", Notes = "Send the subtitle but does not show it. Use SendTitleTimes to show the title." },
```
madmaxoft: `Sends the Subtitle to the client. Doesn't display make the client it yet, use SendTitleTimes() to show both the Title and the Subtitle.`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -596,7 +596,14 @@ end
...
+				SendSetRawSubTitle = { Params = "RawSubTitle", Return = "", Notes = "Sends the raw subtitle but does not show it. Use SendTitleTimes to show the title." },
```
madmaxoft: What is the difference between regular `Subtitle` and `RawSubtitle`? The documentation should explain it.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -596,7 +596,14 @@ end
...
+				SendSetTitle = { Params = "Title", Return = "", Notes = "Sends the title but does not show it. Use SendTitleTimes to show the title." },
+				SendSetRawTitle = { Params = "RawTitle", Return = "", Notes = "Sends the raw title but does not show it. Use SendTitleTimes to show the title." },
+				SendTitleTimes = { Params = "FadeInTicks, DisplayTicks, FadeOutTicks", Return = "", Notes = "Sends the title times and shows the title." },
```
madmaxoft: `Sends the request to display the Title and Subtitle, previously set with SendSetTitle, SendSetRawTitle, SendSetSubTitle and SendSetRawSubTitle, to the client.`

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: Ah, I see, regular takes a `cCompositeChat`, while `Raw` takes a string (JSON-formatted, I expect)

mohe2015:
SendResetTitle only hides the title and SendHideTitle hides both.

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: Use `{{cCompositeChat|Title}}` to indicate the param type.

mohe2015:
But the following code does not show the title like expected:
```
Player:GetClientHandle():SendHideTitle()
Player:GetClientHandle():SendSetRawTitle("title")
Player:GetClientHandle():SendSetRawSubTitle("subtitle")
Player:GetClientHandle():SendTitleTimes(5, 80, 5)
```

mohe2015:
http://wiki.vg/Protocol#Title

mohe2015:
I think the client has some issues if you send the hide title and title times instantly after that because the title and subtitle does not show then.

mohe2015:
If you use SendHideTitle you have to use SendTitleTimes before SendSet*. Otherwise the title and the subtitle does not show. I will update the documentation.

mohe2015:
Can somebody try if SendHideTitle works, please?
I have written a test plugin for that: [TestPlugin.zip](https://github.com/cuberite/cuberite/files/449660/TestPlugin.zip)


What happens and what should happen:
1. /show shows title and subtitle
2. /hide hides the title and subtitle
3. /reshow does nothing but should show the title again (**Am I right?**)

mathiascode:
Sorry for the delay. I'll test it once I come home later today.

mohe2015:
Ok @mathiascode

mathiascode:
This is what happens when I execute /show. I'm using Minecraft 1.9.4.
screenshot from 2016-09-11 02-19-58

mohe2015:
That's probably not the fault of the lua export but the fault of the title protocol implementation. I'll look into it later.

mohe2015:
I've investigated the changes and found out that all versions after ~1.8 do not support raw titles. I think we should remove them because 1.8 also supports normal json titles and the following should work on all versions:
`Player:GetClientHandle():SendSetTitle(cCompositeChat("test"))
Player:GetClientHandle():SendSetSubTitle(cCompositeChat("test"))
Player:GetClientHandle():SendTitleTimes(5, 80, 5)
`

What do you mean?

mohe2015:
I've found out, why that happens. Until 1.9 you could just send the raw text for the title. But since 1.9 you have to send a correctly formatted json string. So everything works like expected.

Can this get merged?

mohe2015:
Moved to #3408 because I had to change the branch. I'm sorry.

Issue #3364, opened 2016-09-03
Lighting thread startup crash


LogicParrot:
We had a startup crash in Travis.
https://travis-ci.org/cuberite/cuberite/jobs/157274141
```
[07:59:02] Regenerating webadmin.ini, all settings will be reset
[07:59:02] WebServer: The server will run in unsecured HTTP mode.
[07:59:02] Put a valid HTTPS certificate in file 'webadmin/httpscert.crt' and its corresponding private key to 'webadmin/httpskey.pem' (without any password) to enable HTTPS support
[07:59:02] There are no ranks, migrating old-style INI files to new DB ranks...
[07:59:02] (This might take a while)
[07:59:02] Rank migration failed, creating default ranks...
[07:59:02] Default ranks created.
[07:59:02] Loaded 465 crafting recipes
[07:59:02] Loaded 22 furnace recipes and 56 fuels
[07:59:02] Loaded 88 brewing recipes
[07:59:02] -- Loading Plugins --
[07:59:02] Initialised Core v.15
[07:59:02] Initialized TransAPI v.1
[07:59:02] Initialized ChatLog v.3
[07:59:02] -- Loaded 3 Plugins --
[07:59:02] Chosen a new random seed for world: 1181341034
[07:59:03] Preparing spawn (world): 19.25% (77/400; 74.90 chunks / sec)
[07:59:04] Preparing spawn (world): 51.50% (206/400; 126.72 chunks / sec)
[07:59:05] Preparing spawn (world): 85.00% (340/400; 132.41 chunks / sec)
[07:59:05] Generated spawnpoint position at {0.50, 62.00, -7.50}
[07:59:05]   D:    | Cuberite has encountered an error and needs to close
[07:59:05] Details | SIGSEGV: Segmentation fault
[07:59:05] Cuberite Travis gcc RELEASE build id: 9661.3
[07:59:05] from commit id: 2b5663d4b4e8d364c8d61e9f08ee828ed9ad2535 built at: Sat Sep  3 07:53:22 UTC 2016
./Cuberite(_Z15PrintStackTracev+0x14)[0x6299a4]
./Cuberite[0x616a73]
/lib/x86_64-linux-gnu/libc.so.6(+0x360b0)[0x7faad77b50b0]
/lib/x86_64-linux-gnu/libpthread.so.0(pthread_mutex_lock+0x4)[0x7faad836be84]
[...]

PR #3365, opened 2016-09-03
Added gdb to travis

Seadragon91:
Retry of #2585.

Added gdb and script commands.
Also code that crashes the server after initialized, will be removed after success.

In #3364 we have `core dumped`, it looks like that core files are now generated automatically. Therefore I retry this.


Seadragon91:
Okay bad try, sorry.

PR #3366, opened 2016-09-03
TNT explosion rate limit

LogicParrot:
Main change: TNT explosion rate limit
Secondary change to cWorld:
cWorld now has a tick variable. I did that to avoid adding yet another m_LastX, and because I want the explosion rate to be tick-dependent and not time-dependent. That way, if the world lags, the explosion rate per tick does not increase.

Secondary change to the explosion  loop:
I made the explosion loop scan inside-out. This prevents chain explosions from having a single preferred general direction on low explosion rates.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1711,97 +1711,107 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 			return;
 		}
 
-		for (int x = -ExplosionSizeInt; x < ExplosionSizeInt; x++)
+		for (int radius = 1; radius <= ExplosionSizeInt; ++radius)
 		{
-			for (int y = -ExplosionSizeInt; y < ExplosionSizeInt; y++)
+			for (int x = -radius; x <= radius; x++)
 			{
-				if ((by + y >= cChunkDef::Height) || (by + y < 0))
+				for (int y = -radius; y <= radius; y++)
 				{
-					// Outside of the world
-					continue;
-				}
-				for (int z = -ExplosionSizeInt; z < ExplosionSizeInt; z++)
-				{
-					if ((x * x + y * y + z * z) > ExplosionSizeSq)
+					if ((by + y >= cChunkDef::Height) || (by + y < 0))
 					{
-						// Too far away
+						// Outside of the world
 						continue;
 					}
-
-					BLOCKTYPE Block = area.GetBlockType(bx + x, by + y, bz + z);
-					switch (Block)
+					for (int z = -radius; z <= radius; z++)
 					{
-						case E_BLOCK_TNT:
+						if ((x * x + y * y + z * z) > ExplosionSizeSq)
 						{
-							// Activate the TNT, with a random fuse between 10 to 30 game ticks
-							int FuseTime = 10 + m_World->GetTickRandomNumber(20);
-							m_World->SpawnPrimedTNT(a_BlockX + x + 0.5, a_BlockY + y + 0.5, a_BlockZ + z + 0.5, FuseTime);
-							area.SetBlockTypeMeta(bx + x, by + y, bz + z, E_BLOCK_AIR, 0);
-							a_BlocksAffected.push_back(Vector3i(bx + x, by + y, bz + z));
-							break;
+							// Too far away
+							continue;
+						}
+						if ((abs(x) != radius) && (abs(y) != radius) && (abs(z) != radius))
+						{
+							continue;
 						}
 
-						case E_BLOCK_OBSIDIAN:
-						case E_BLOCK_BEACON:
-						case E_BLOCK_BEDROCK:
-						case E_BLOCK_BARRIER:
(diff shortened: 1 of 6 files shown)


src/ChunkMap.cpp
```diff
@@ -1711,97 +1711,107 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 			return;
 		}
 
-		for (int x = -ExplosionSizeInt; x < ExplosionSizeInt; x++)
+		for (int radius = 1; radius <= ExplosionSizeInt; ++radius)
```
LogicParrot: Better variable names. Explain that it's an inside-out sphere loop.

src/ChunkMap.cpp
```diff
@@ -1711,97 +1711,107 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
...
+						BLOCKTYPE Block = area.GetBlockType(bx + x, by + y, bz + z);
+						switch (Block)
 						{
-							// Turn into simulated water:
-							area.SetBlockType(bx + x, by + y, bz + z, E_BLOCK_WATER);
-							break;
-						}
+							case E_BLOCK_TNT:
+							{
+								if (m_World->GetExplosionLimiter().RequestExplosion())
```
NiLSPACE: Would it perhaps be better to schedule the explosion? If I read this correctly there is a chance that some TNT stays after blowing up a huge chunk of TNT simply because there is too much of it.

LogicParrot: That's a good point. It would also rid us of the looping changes.

src/ChunkMap.cpp
```diff
@@ -1711,97 +1711,110 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
...
-					// Outside of the world
-					continue;
-				}
-				for (int z = -ExplosionSizeInt; z < ExplosionSizeInt; z++)
-				{
-					if ((x * x + y * y + z * z) > ExplosionSizeSq)
+					if ((by + y >= cChunkDef::Height) || (by + y < 0))
 					{
-						// Too far away
+						// Outside of the world
```
LogicParrot: Outside the sphere

src/ChunkMap.cpp
LogicParrot: Chain explosions still occur, but when too many TNTs are there, some TNTS are left behind when the explosion chain is over. But one could say that inhibiting oversized explosion chains is a good thing for a server which specified a low rate. Less lag and less block destruction.

So, what do we want, from a user-perspective, the current PR or the scheduling proposal?

NiLSPACE: I'm not sure. @cuberite/core-team what do you think is best?

I thought of something else where this could possibly cause issues. If somewhere on the server allot of TNT goes off, and I try to light only a few (lets say 5) TNT somewhere else unaware of the huge explosion further ahead, is there a chance that mine won't go off?

LogicParrot: >is there a chance that mine won't go off?

The limit does not cover manually ignited TNTs, so your first TNT would go off, but the remaining 4 TNTs have a chance of not exploding in a chain reaction.

Per chunk limit, rather than per-world, would solve this.

NiLSPACE: I guess, but with explosions TNT entities tend to get thrown around allot.

LogicParrot: Also, maybe it's wiser to have a per world limit. The main goal is preventing server overload. If different people blow different TNTs different places at the same time, they can cause still overload the server, unless the cap is shared and is per world.

NiLSPACE: Yeah, I agree.

tigerw:
I think we should investigate the cause which necessitates this - what is it about TNT that causes, allegedly, so much lag? After all, TNT is nothing special but entities, explosion packets, and block accesses, none of which seem like, even in large numbers, that they should be slow. Or perhaps they *are* slow?

In that case, surely there must be some potential for optimisation?

LogicParrot:
Optimizations are good, but I think a cap is needed anyways. TNTs cause a rapidly growing chain reaction, so a sufficient amount of them will cause a positive feedback loop strong enough to bring the server to its knees no matter how optimized we get. A player can always add more and more TNTs.

tigerw:
Makes sense.


Then in that case, I would recommend a per world limit with exceeded explosions pushed onto the next tick.




Also, is using an extra class considered good practice? I guess it helps modularise things somewhat.









From: NiLSPACE
Sent: ‎Sunday‎, ‎11‎ ‎September‎ ‎2016 ‎10‎:‎46
To: cuberite/cuberite
Cc: Tiger Wang, Team mention





Optimizations are good, but I think a cap is needed anyways. TNTs cause a rapidly growing chain reaction, so a sufficient amount of them will cause a positive feedback loop strong enough to bring the server to its knees no matter how optimized we get. A player can always add more and more TNTs.

PR #3376, opened 2016-09-12
Lua API description reformat

### LINKED ISSUE #3375: Reformat Lua API Desc to explicitly include param types
Currently the Lua API description (in the APIDump built-in plugin) has no machine-readable indication of the functions' parameters' and return values's types. We should provide an explicit declaration of the types.

Before:
```lua
FnName = { { Params = "Command, Handler, Permission, [HelpString]", Return = "bool", Notes = ...}, ... }
```

After:
```lua
FnName =
{
  {
    Params =
    {
      { Type = "string", Name = "Command" },
      { Type = "function", Name = "Handler" },
      { Type = "string", Name = "Permission" },
      { Type = "string", Name = "HelpString", Optional = true},
    },
    ...
  },
}
```

This is rather lengthy, but is worth it - from this format we could generate better docs and even extract the API info for automated plugin testing).

The reformatting would ideally be done by a script that could guess the common parameter types - it is already being done in the https://github.com/madmaxoft/ManualApiDump plugin.

madmaxoft:
Reformatted the Lua API description to contain type information for parameters and return values.

Also found out that the `cFloater` class was documented in two places, with the second documentation overwriting the first one; fixed and added a detection that marks CI build as failed if this happens again.

Fixes #3375

### DIFF
--- a/CIbuild.sh
+++ b/CIbuild.sh
@@ -31,4 +31,9 @@ EOF
 		cat ./NewlyUndocumented.lua
 		exit 1
 	fi
+	if [ -f ./DuplicateDocs.txt ]; then
+		echo "ERROR: API documentation has duplicate symbol warnings:"
+		cat ./DuplicateDocs.txt
+		exit 1
+	fi
 fi
--- a/Server/.gitignore
+++ b/Server/.gitignore
@@ -32,6 +32,7 @@ motd.txt
 *.xml
 mcserver_api.lua
 cuberite_api.lua
+DuplicateDocs.txt
 
 # Ignore the webadmin certs / privkey, so that no-one commits theirs by accident:
 webadmin/httpscert.crt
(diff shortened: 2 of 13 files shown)


NiLSPACE:
Lgtm. This should clear up quite allot of uncertainty with the types :)

NiLSPACE:
The APIDocs just got updated, and I noticed all the tables for the arguments. I believe it would look nicer if we'd make those tables extend in width to the end of the cell. Adding
```css
td table {
	width: 100%;
}
```
does the trick for me.

I noticed it also extends a few other tables (for example the tables for a hook's parameters), but it still looks fine.

madmaxoft:
Feel free to modify the CSS, I just went for the functional part.

sphinxc0re:
Also the return value shouldn't concist of a table. Looks a bit weird:
bildschirmfoto 2016-09-13 um 15 08 12

madmaxoft:
I think that looks good enough, and some functions do have extended documentation on the return values, specifying both their names and types: http://apidocs.cuberite.org/cWorld.html#GetBlockInfo_1

PR #3380, opened 2016-09-24
Move entity metadata writing to the entities themselves (for 1.9+)

Pokechu22:
**This is provided as an alternative way of implementing, but not necessarily the best one**.  [Here's a table of what I think the pros and cons of this method, the current method, and several similar ones](https://github.com/cuberite/cuberite/pull/3380#issuecomment-250043754).  If you have suggestions or ideas about which is best architecturally, post them!  (Even if it's "the current method is better").

---

This moves the logic to write [entity metadata](http://wiki.vg/Entities) from the protocol classes to entities themselves.  This allows for much cleaner code, and avoids copying all of the metadata writing code just to increment a number for all entities.

The core concept behind it is to move metadata serialization to a virtual method in `cEntity`, and to have each subclass write its own metadata after the that of its superclass.  It's possible to determine the protocol version being written for, meaning that fields that are added in newer versions are only be written in newer versions (and IDs are shifted accordingly).  For 1.10 there's only one such field that cuberite uses, which is the "No Gravity" field on the base entity class (but since it's on the base entity class, the addition of that field affects all entities).

Doing it this way only makes sense because metadata is assigned incrementally.  In 1.8, metadata wasn't assigned incrementally, but it is in 1.9 and above.  Since it's assigned incrementally, if a field is added to a class low on the hierarchy, all of the fields after it have their ID incremented as well, creating lots of work if IDs are hardcoded.  Dynamically creating the IDs based off of what was written already is a much simpler system in terms of maintenance.

This new system was designed with [1.11's new metadata fields](http://wiki.vg/index.php?title=Pre-release_protocol&oldid=8118#Entity_Metadata) in mind and it should be easy to implement those when the time comes.  1.8 is not affected by this change because 1.8 uses somewhat random metadata assignment rather than incremental assignment.


Pokechu22:
I know why the build's failing - `Globals.h` is getting included twice somehow.  However, I'm not sure _why_ (or where) that's happening.

madmaxoft:
Before you go any further, wait!
We have specifically removed the entity serialization code from `cEntity` and its descendants, so that the entire protocol is encapsulated in the `cProtocol` descendant. That way multiple protocols can have different serializations and it's much easier to write a new protocol class alongside an older one. I believe this is much cleaner concept than having a `cEntity` object that knows how to serialize itself in two (or more) different protocol versions.

madmaxoft:
I think a better way would be to rewrite the `cEntity` class to provide a generic storage for attributes (a key-type-value storage, where key is the property's Index, type is an enum specifying the datatype (int, bool, string, uuid...) and the value is a variant storing the actual data) which gets populated by each subclass' constructor; each subclass then provides getters and setters for its named properties (`cHorse::IsTamed()`) that actually use the attribute storage.
Serialization can then be written as an operation on that attribute storage, rather than a `cEntity` type, and it will be a simple for-loop over all the attributes stored in the entity.

madmaxoft:
Oh and stop using "BlockID". That doesn't make sense - an ID is a unique identification, which, for a block, is its <world + coords>. There's always only up to 1 block of a single ID.
BlockType is what you want - it describes what type the block is (and there could be multiple instances of that block type in a world).

Pokechu22:
> That way multiple protocols can have different serializations and it's much easier to write a new protocol class alongside an older one.

I designed it `WriteMetadata` takes an abstract `cMetadataWriter`, but protocols declare their own (eg `cMetadataWriter190`) which handles formatting it.  I feel like doing it that way does meet that requirement (the only reason why there is no `cMetadataWriter1100` is because 1.9 and 1.10 have identical metadata formats as far as data types).

> I think a better way would be to rewrite the `cEntity` class to provide a generic storage for attributes (a key-type-value storage, where key is the property's Index, type is an enum specifying the datatype (int, bool, string, uuid...) and the value is a variant storing the actual data) which gets populated by each subclass' constructor; each subclass then provides getters and setters for its named properties (`cHorse::IsTamed()`) that actually use the attribute storage.
> 
> Serialization can then be written as an operation on that attribute storage, rather than a `cEntity` type, and it will be a simple for-loop over all the attributes stored in the entity.

I agree that that method would be better, I'm just not familiar enough with C++ to do it that way (I wanted to write it that way but didn't feel comfortable with it).

The one thing that I'm not sure of is how you could handle multiple versions' different metadata with that.  The property indexes vary by version now, and I don't think the incremental index thing I did here would work with a generic storage system.  There might be a way to do it that I'm missing, though.

> Oh and stop using "BlockID". That doesn't make sense - an ID is a unique identification, which, for a block, is its . There's always only up to 1 block of a single ID.
 BlockType is what you want - it describes what type the block is (and there could be multiple instances of that block type in a world).

Sorry, BlockID is just the notation I'm used to (both [Minecraft Wiki](http://minecraft.gamepedia.com/Data_values#Block_IDs) and [wiki.vg](http://wiki.vg/Entities#Entity_Metadata_Format "OptBlockID in the table") use it.  I'll switch to BlockType here though since that's what the rest of the project is using.

madmaxoft:
You're right, the generic attribute storage won't be able to work with multiple protocols. But your changes, in order to be able to do that, bring protocol dependency into the `cEntity` class, and I believe that is not the right way to go.

How about making a `cMetadataWriter` an internal class of the protocol, and the protocol can call the functions that serialize the metadata. So the `WriteEntityMetadata()` function creates a `cMetadataWriter` instance and uses that to serialize the metadata, instead of writing them directly to the `cPacketizer`.

If you decide to go this way, perhaps start with a fresh new branch rather than this experimental one; in the end we could compare all three approaches and select the one we like the most. So far I have to admit I like the original code much more than your proposed changes.

Pokechu22:
> But your changes, in order to be able to do that, bring protocol dependency into the `cEntity` class, and I believe that is not the right way to go.

I've avoided that - the only protocol dependency is upon protocol version number enum (which was moved to Globals.h), not the actual `cProtocol` classes.

> How about making a `cMetadataWriter` an internal class of the protocol, and the protocol can call the functions that serialize the metadata. So the `WriteEntityMetadata()` function creates a `cMetadataWriter` instance and uses that to serialize the metadata, instead of writing them directly to the `cPacketizer`.

I feel like the code I already have is almost exactly the same as what you want, unless I'm missing something.  There _is_ [a `cMetadataWriter`](https://github.com/cuberite/cuberite/pull/3380/files#diff-6f1a071e652c6f599291ec6ba0c1cac9), which is a base class that is protocol-unspecific, and also [a `cMetadataWriter190`](https://github.com/cuberite/cuberite/pull/3380/files#diff-8513dfc10d7dc6ef552751696458cb39R340) internal to `cProtocol190` which handles actually writing to the `cPacketizer`.

Here's some demos of how it's used - I'm not sure how this could be made cleaner.

`cBoat`:

```c++
void cBoat::WriteMetadata(cMetadataWriter & a_Writer) const
{
	super::WriteMetadata(a_Writer);
	a_Writer.WriteInt(GetLastDamage());  // Time since last hit
	a_Writer.WriteInt(GetForwardDirection());  // Forward direction
	a_Writer.WriteFloat(GetDamageTaken());  // Damage taken
	a_Writer.WriteInt(GetType());  // Type
	a_Writer.WriteBool(IsRightPaddleUsed());  // Right paddle
	a_Writer.WriteBool(IsLeftPaddleUsed());  // Left paddle
}
```

`cSheep`:

```c++
void cSheep::WriteMetadata(cMetadataWriter & a_Writer) const
{
	super::WriteMetadata(a_Writer);
	Int8 SheepMetadata = 0;
	SheepMetadata = static_cast<Int8>(GetFurColor());
	if (IsSheared())
	{
		SheepMetadata |= 0x10;
	}
	a_Writer.WriteByte(SheepMetadata);  // Sheep flags (wool & sheered)
}
```

src/Entities/Entity.cpp
```diff
@@ -2285,7 +2285,8 @@ void cEntity::WriteMetadata(cMetadataWriter & a_Writer) const
 	a_Writer.SkipMeta();  // Is silent
 	if (a_Writer.m_ProtocolVersion >= PROTO_VERSION_1_10_0)
 	{
-		a_Writer.WriteBool(GetGravity() == 0);  // No gravity
+		const double EPS = 0.0000001;
```
sphinxc0re: Please document this constant. You probably want to declare it somewhere else as well.

Pokechu22: The constant is also defined in `SetYawFromSpeed` and `SetPitchFromSpeed`, which is why I copied it again.

I think it might actually be better to not use an epsilon from 0 and instead check if it's changed from the default gravity (since NO_GRAVITY means the clientside gravity code isn't applied).  I think a `DEFAULT_GRAVITY` constant would work well for that.

Working on this...

Pokechu22: Done

madmaxoft:
My point is that NO networking code should be in `cEntity` and its descendants. So rather than in `cEntity::WriteMetadata()`, the code should be in `cProtocol` descendants. Possible solution:
```cpp
class cProtocol
{
  // Current entrypoint for writing any entity metadata; relays work to specialized worker functions below
  virtual void WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_Entity) = 0;

  // Each method calls the SerializeMetadata() method for the class parent, then writes its own metadata
  virtual void SerializeMetadata(cMetadataWriter & a_MetadataWriter, const cEntity & a_Entity);
  virtual void SerializeMetadata(cMetadataWriter & a_MetadataWriter, const cMonster & a_Monster);
  virtual void SerializeMetadata(cMetadataWriter & a_MetadataWriter, const cCreeper & a_Creeper);
}

...

void cProtocol::SerializeMetadata(cMetadataWriter & a_MetadataWriter, const cCreeper & a_Creeper)
{
  // Call the parent class' serialization:
  SerializeMetadata(a_MetadataWriter, static_cast<const cMonster &>(a_Creeper);

  // Serialize creeper's own data:
  a_MetadataWriter.WriteBool(a_Creeper.IsCharged);
}
```
Not sure if C++ allows this, might need tweaking (each descendant's method name might need to be unique).

As I said, since this approach is quite different from this PR, it might be better to try it in a separate PR.

Pokechu22:
I feel like that would be less organized, but I'll try it, for comparison purposes.

src/Entities/Entity.cpp
```diff
@@ -21,6 +21,8 @@
 
 UInt32 cEntity::m_EntityCount = 0;
 cCriticalSection cEntity::m_CSCount;
+const float cEntity::DEFAULT_GRAVITY = -9.81f;  // Meters per second per second
+const double cEntity::EPS = 0.0000001;
```
sphinxc0re: I'm no sure about these either. First of all why isn't the gravity defined for the cWorld class? And secondly, could you possibly change EPS to the full name EPSYLON? Characters are cheap 😄

Pokechu22:
I'm going to revert the gravity changes since it turns out that most entities have a gravity of -32.0 or -16.0, not -9.81; it's probably better to leave control of the no gravity tag for something else since right now it's causing problems (players aren't falling, for instance).  I did change the EPS constant name, though (so that it matches its only other use).

Pokechu22:
@madmaxoft: OK, I've got things up for comparison (I'm not going to submit PRs for each though since that would be confusing): [entity-metadata-1](https://github.com/cuberite/cuberite/compare/master...Pokechu22:entity-metadata-1?expand=1), [entity-metadata-2](https://github.com/cuberite/cuberite/compare/master...Pokechu22:entity-metadata-2?expand=1), and [entity-metadata-3](https://github.com/cuberite/cuberite/compare/master...Pokechu22:entity-metadata-3?expand=1).  Note that entity-metadata-1 is the same as this PR; also, the histories on the other branches will be cleaned up later.

* Entity-metadata-1 has `cEntity::WriteMetadata`, where each entity writes their own metadata from within their class.
* Entity-metadata-2 has `cProtocol::WriteMetadata`, where there are a bunch of methods in `cProtocol` to write metadata and things are cast to write the next level; however choosing the right staring method requires a large block of `IsA` checks.
* Entity-metadata-3 is basically the same as entity-metadata-1 in that there are `cEntity::WriteMetadata` methods, but rather than having those methods be implemented in the cpp file for each entity, they're all implemented in a single `Protocol/EntityMetadata.cpp` file, which I think is a good mix between allowing for polymorphism within `WriteMetadata` (the only thing that needs to be called is `cEntity::WriteMetadata` and the right one is chosen) and having the metadata be within the protocol code (although I'm not sure if it goes against C++ style).

And here's a table of comparisons that I think are relevant between each one:

<table>
<tr><th></th><th>Master</th><th>1</th><th>2</th><th>3</th></tr>
<tr><th>Automatic metadata ID assignment</th><td>:x:</td><td>✔️</td><td>✔️</td><td>✔️</td></tr>
<tr><th>Metadata writing located in protocol</th><td>✔️ (Protocol19x.cpp)</td><td>:x: (Entity.cpp and subclasses)</td><td>✔️ (Protocol19x.cpp)</td><td>✔️ (EntityMetadata.cpp)</td></tr>
<tr><th>Polymorphic metadata method selection</th><td>:x:</td><td>✔️</td><td>:x:</td><td>✔️</td></tr>
<tr><th>Methods implemented in declaring file</th><td>✔️</td><td>✔️</td><td>✔️</td><td>:x:</td></tr>
</table>

madmaxoft:
I still like method 2 the most, and instead of the `IsA` block, you could use two nested `switch`es - one on `cEntity::GetEntityType()` and one inside the `etMonster` case, on `cMonster::GetMobType()`. Also don't include all the entity and mob types in Protocol.h, instead forward-declare them, and include them only in Protocol.cpp:
```cpp
// fwd:
class cEntity;
class cMonster;
...

void DoSomething(cEntity & a_Entity);  // compiles OK
```

I don't understand the row "Polymorphic metadata method selection" in your table, what exactly do you mean by that? Because in my opinion method 2 does use the polymorphism, although, I have to admit, in a non-standard way.

Pokechu22:
> instead of the  IsA  block, you could use two nested  switch es - one on `cEntity::GetEntityType()` and one inside the `etMonster` case, on `cMonster::GetMobType()`

I can do that, although that basically makes it the same code as there originally was but split across more methods (which isn't necessarily a bad thing)

> Also don't include all the entity and mob types in Protocol.h, instead forward-declare them, and include them only in Protocol.cpp:
> 
> ```c++
> // fwd:
> class cEntity;
> class cMonster;
> ```

I didn't know that was possible; I'll switch to it (though I'm curious about the exact benefit - does it help speed up compile times?).

> I don't understand the row "Polymorphic metadata method selection" in your table, what exactly do you mean by that? Because in my opinion method 2 does use the polymorphism, although, I have to admit, in a non-standard way.

I think that if I know the terminology correctly, both methods 1 and 2 use inheritance, but only method 1 uses polymorphism (though I may be drawing a nonexistent distinction).  What I'm referring to is the ability to call just a single `WriteMetadata` method and the correct one is chosen by the compiler/at runtime instead of having to manually switch/cast and call the right method.

Pokechu22:
I suppose that getting some additional opinions would be helpful: @bibo38 @sphinxc0re

madmaxoft:
I'd really love to hear @worktycho 's opinion on this, he's knowledgable on this "architectural" stuff.

src/Mobs/Rabbit.cpp
```diff
@@ -45,3 +45,13 @@ void cRabbit::GetDrops(cItems & a_Drops, cEntity * a_Killer)
...
 }
 
+
+
+
+
+void cRabbit::WriteMetadata(cMetadataWriter & a_Writer) const
+{
+	super::WriteMetadata(a_Writer);
+	a_Writer.WriteInt(static_cast<Int32>(GetRabbitType()));  // Rabit type
```
bibo38: Rabit is missing a 'b'

src/Mobs/Sheep.cpp
```diff
@@ -207,3 +207,19 @@ NIBBLETYPE cSheep::GenerateNaturalRandomColor(void)
...
+void cSheep::WriteMetadata(cMetadataWriter & a_Writer) const
+{
+	super::WriteMetadata(a_Writer);
+	Int8 SheepMetadata = 0;
+	SheepMetadata = static_cast<Int8>(GetFurColor());
+	if (IsSheared())
+	{
+		SheepMetadata |= 0x10;
+	}
+	a_Writer.WriteByte(SheepMetadata);  // Sheep flags (wool & sheered)
```
bibo38: sheared instead of sheered

src/Mobs/ZombiePigman.cpp
```diff
@@ -51,3 +51,17 @@ void cZombiePigman::KilledBy(TakeDamageInfo & a_TDI)
 
 
 
+
+void cZombiePigman::WriteMetadata(cMetadataWriter & a_Writer) const
+{
+	super::WriteMetadata(a_Writer);
+	a_Writer.WriteBool(IsBaby());  // Is baby
+	// Note: Zombie pigmen inherit zombie metadata
```
bibo38: pigm**a**n

src/Mobs/Wolf.cpp
```diff
@@ -402,3 +402,37 @@ void cWolf::InStateIdle(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 
 
+
+
+
+void cWolf::WriteMetadata(cMetadataWriter & a_Writer) const
+{
+	super::WriteMetadata(a_Writer);
+
+	// Ageable fields (since wolf inherits cPassiveAgressiveMonster instead of cPassiveMonster, this isn't handled directly)
```
bibo38: \- Agressive
\+ Ag**g**ressive

#diffstyle :yum:

src/Mobs/MagmaCube.cpp
```diff
@@ -38,3 +38,14 @@ AString cMagmaCube::GetSizeName(int a_Size)
...
+
+
+
+
+
+void cMagmaCube::WriteMetadata(cMetadataWriter & a_Writer) const
+{
+	super::WriteMetadata(a_Writer);
+	// Note: This is slime metadata
+	a_Writer.WriteInt(GetSize());  // Size
```
bibo38: If the Slime metadata would change, it would be changed in the `cSlime` class, but who would know, that the `cMagmaCube` also needs to be changed :confused:?

src/Mobs/ZombiePigman.cpp
Pokechu22: I'm pretty sure that this one should be plural (you'd say "Pigs inherit animal metadata" and not "Pig inherits animal metadata"), so pigmen makes sense.  (Although you would say `cZombiePigman` inherits `cZombie` metadata, singular, I feel like referring to the entities themselves seems like it should be plural).  If it still seems like this should be singular I can change it.

src/Mobs/Wolf.cpp
```diff
@@ -402,3 +402,37 @@ void cWolf::InStateIdle(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+	a_Writer.WriteBool(IsBaby());
```
bibo38: This is a more obscure hierarchy problem. When you change `cPassiveMonster`, you would think, that everything is ok, since there are x classes inheriting and calling the changed method.

**How should you know about the Wolfes?**

src/Mobs/ZombiePigman.cpp
```diff
@@ -51,3 +51,17 @@ void cZombiePigman::KilledBy(TakeDamageInfo & a_TDI)
...
+	a_Writer.SkipMeta();  // Zombie type / profession
+	a_Writer.SkipMeta();  // Is converting
+	a_Writer.SkipMeta();  // Are hands up
```
bibo38: Same as `cWolf`.

src/Mobs/Bat.cpp
```diff
@@ -14,3 +14,17 @@ cBat::cBat(void) :
...
+
+void cBat::WriteMetadata(cMetadataWriter & a_Writer) const
+{
+	// super isn't used here because that refers to cPassiveMonster, which contains the is baby flag.
+	// Since bat doesn't extend Ageable but does extend insentient, we need to skip the ageable metadata.
+	// This is an odd special case that is caused by the differences between vanilla's entity hierarchy and
+	// Cuberite's entity hierarchy.
+	cMonster::WriteMetadata(a_Writer);
+	a_Writer.WriteByte(IsHanging() ? 1 : 0);  // Flags
+}
```
bibo38: Same as `cWolf`

bibo38:
Now to the problem where the Metadata code should go:

Single Responsibility Principle

Because the `cEntity` classes are used to store and modify the entity properties, they should not be used to handle the Protocol. Like `cWorld` doesn't create NBT-Files and compresses World Chunks for the Minecraft Protocol. Furthermore you are able to remove a protocol just by removing a file and editing a second file, instead of editing about 30 Entity classes.

So I totally agree to Madmaxoft. I would also use the switch statement, because it doesn't require about 30 string comparisions(which the `IsA` method uses). I would recommend the following style:

```cpp
switch(entity->GetType())
{
    case WOLF: WriteMetadata<cWolf>(); break;
    case PIG:      WriteMetadata<cPig>(); break;
    default: Assert(!"What is THAT?");
}
```

The templates are not mandatory, but having each code seperate helps to keep the switch statement more readable(**you see (intentional) missing breaks**).

Templates would help to generalize the superclass calling. This helps to prevent (intentional) missing super calls.

src/Mobs/Wolf.cpp
bibo38: Yep, didn't see that :+1:

src/Mobs/ZombiePigman.cpp
bibo38: Yep, you're right. I've just compared it with the class name.

bibo38:
>  I didn't know that was possible; I'll switch to it (though I'm curious about the exact benefit - does it help speed up compile times?).

Yep, It minimizes the amount of indirect includes. When a header file changes, every file including this header file needs to be recompiled. But most C++ files don't use all Includes, which are in the header files and are therefore needlessly recompiled.

Example: You change `Entity.h`. Without forward includes the ProtocolRecognizer is recompiled again, although if it doesn't use any method from `cEntity`. This wasted 2 seconds of compile time.

Pokechu22:
Is it actually possible to fix the issue with `cSlime` / `cMagmaCube` without changing cuberite's class structure?  Even with madmaxoft's method, I think one would need two methods, one for `cSlime` and one for `cMagmaCube`, since they don't share methods (even though they should).  So even if a method similar to madmaxoft's is used, we'd still need to change the class structure.

> I would also use the switch statement, because it doesn't require about 30 string comparisions (which the `IsA` method uses).

I totally agree that `IsA` is the wrong choice (the only reason I used it was that it was easy to automatically generate).  I still don't like the switch, though, because it actually requires a nested switch (or several!) - one for cEntity::GetEntityType(), one for cMonster::GetMobType(), etcetra.  And if we were to fix the class structure, I don't know if that strategy can hold (you'd have to disambiguate slime and magma cube somehow, and given that the type is put in the `cMonster` constructor, I'm not sure if that would be possible for subclasses).

I guess what I'm trying to say is that I don't like having to switch up the hierarchy - if it were a flat switch it might be better.

> > What I'm referring to is the ability to call just a single WriteMetadata method and the correct one is chosen by the compiler/at runtime instead of having to manually switch/cast and call the right method.

> Templates could generalize the superclass calling. This also helps to prevent (intentional) missing super calls. But sadly there is no way to extend C++ classes by design. 

I know next to nothing about how C++'s templates work (though I'd be happy to learn; it's just something I've never worked with).

Would it be possible (and, would it be a good idea) to put something into the CLASS_PROTODEF macro for `cEntity` that automatically defines a virtual method that calls the right protocol method?  Maybe, something like (this is a guess about the syntax)

```
virtual void CallWriteMetadata(cProtocol190 & a_Protocol, cMetadataWriter & a_Writer) const override \
	{ \
		a_Protocol.WriteMetadata<#classname>(*this, a_Writer); \
	}
```

If so we would still be putting protocol logic in entities, but it would only be to identify the right method to call.  Maybe a template (or an interface?) could be used to remove the `cProtocol190` part and make it more abstract, I'm not sure exactly...

Pokechu22:
I did some more research, and according to [this](http://stackoverflow.com/a/7629011/3991344) seems that what I want is called multimethods/multiple dispatch.  [Wikipedia says](https://en.wikipedia.org/wiki/Multiple_dispatch#C.2B.2B) that C++ doesn't support it directly, but there are ways to get similar behavior.  I'll see if something like that can be done here.

----

I just realized another perk of madmaxoft's method - `cProtocol1100` could override only `cProtocol190`'s entity method to add the no gravity field, which would let us remove the dependency upon protocol versions within metadata writing (the appropriate protocol instance would know what version it is).

bibo38:
> Would it be possible (and, would it be a good idea) to put something into the CLASS_PROTODEF macro for cEntity that automatically defines a virtual method that calls the right protocol method? Maybe, something like (this is a guess about the syntax)

It would be totally possible. But it is just a different version of your current solution. Maybe even worse, since the Entities need to include cProtocol for the class definition and cProtocol needs including the entities, which creates a cyclic dependency.
But the problem persists, that the Entities have a relation to the Protocol. It's like adding a "SendEntity" method to cEntity or implementing the cClientHandle methods into the cPlayer class.

> I did some more research, and according to this seems that what I want is called multimethods/multiple dispatch

Which is basically Reflection in Java. C++ supports this also a bit with dynamic_cast, but the problem is, that C++ tends to be an efficient language and dynamic lookups are slow and people tend to use them for every Problem.
But I don't see the difference between the madmaxorf's switch and the dynamic_casts.



> I just realized another perk of madmaxoft's method - cProtocol1100 could override only cProtocol190's entity method to add the no gravity field, which would let us remove the dependency upon protocol versions within metadata writing (the appropriate protocol instance would know what version it is).

I've used inheritance also for the cProtocol1100. Can you give a concrete example?

Pokechu22:
> > I just realized another perk of madmaxoft's method - cProtocol1100 could override only cProtocol190's entity method to add the no gravity field, which would let us remove the dependency upon protocol versions within metadata writing (the appropriate protocol instance would know what version it is).

> I've used inheritance also for the cProtocol1100. Can you give a concrete example?

What I meant was that you could inherit/override only the methods for entities that changed, so you'd only need to override `cProtocol1100::WriteMetadata<cPawn>` and you wouldn't need to copy the entire entity metadata method.  It seems like you can't override template methods, though, so doing that would require unique names (I think) or casts.

> Furthermore you are able to remove a protocol just by removing a file and editing a second file, instead of editing about 30 Entity classes.

Which is what effectively would also happen if `cProtocol190` was removed since then all of the metadata methods that weren't overwritten would need to be copied to `cProtocol1100`.

In fact, I'd say that you'd only need to edit entity classes where the removed protocol caused a change.  So if you remove support for `cProtocol190`, the only change that would be necessary is to edit `cEntity::WriteMetadata` by changing

```c++
if (a_Writer.m_ProtocolVersion >= PROTO_VERSION_1_10_0)
{
    a_Writer.SkipMeta();  // No gravity
}
```

to

```c++
a_Writer.SkipMeta();  // No gravity
```

which doesn't seem like that big of a change.  True, identifying the fact that the `if` statement is now unnecessary would be hard, but that could probably be fixed by using `if (a_Writer.m_ProtocolVersion <= PROTO_VERSION_1_9_4)` - then when `PROTO_VERSION_1_9_4` is removed, that statement will generate an error which should make it clear that it can be removed.

>  But I don't see the difference between the madmaxorf's switch and the `dynamic_casts`.

It works better with the entity metadata hierarchy.  The switch _would_ be fine if there was a single `eEntityType` that had _all_ entities in it (IE, base entities like items and boats; animals such as sheep; monsters such as creepers).  That's not possible with the current layout - you need a switch upon the different `eEntityType`s, and then if it's `etMonster`, you need to switch on `eMonsterType`.

Quick side note: `eMonsterType` seems to be indexed by the network ID of the entity while `cEntityType` isn't.  The distinction seems to be that [Spawn Object](http://wiki.vg/Protocol#Spawn_Object) packet is used for `eEntityType` (in most cases) while [Spawn Mob](http://wiki.vg/Protocol#Spawn_Mob) is used for `eMonsterType`.  I suppose the fact that those entities don't have entity IDs [listed on the mobs section of the Entities article](http://wiki.vg/index.php?title=Entities&oldid=8200#Mobs).  But they actually _do_ have entity IDs (well, most of them); [see Burger](http://pokechu22.github.io/Burger/1.10.html#entities).  There's only a few exceptions: The fishing rod bobber, lightning bolts, and players.

OK, back to `eMonsterType`: Due to the way that `cMonster`'s constructor works (it takes the entity's name, `eMonsterType`, sounds, and size), if we wanted to have a real hierarchy, then each subclass _also_ needs to provide a constructor like that so that subclasses can provide the params.  Right now subclasses only have a `void` constructor.

The switch **is perfectly fine with the current code**.  But that's not the point - as you mentioned, `cSlime` and `cMagmaCube` share metadata.  There's several other similar relationships (eg with `cZombie` / `cZombiePigman`).  So we'd need to change the `cMonster` constructor to support subclasses if we wanted to reuse `cSlime` code in `cMagmaCube`.  (Probably, a protected constructor and a public constructor would work well).  If we do that, then the switch would still be fine.

The other option (which is what I've been complaining about) is to have, say, a `eSlimeType`, and then when you see `mtSlime` you would need to switch _again_.  That's ugly, and definitely shouldn't be done.

> But the problem persists, that the Entities have a relation to the Protocol. It's like adding a "SendEntity" method to cEntity or implementing the cClientHandle methods into the cPlayer class.

I'm going to make a different argument: It's like moving entity NBT serialization to `cEntity`.  Which, in my mind, seems to be fine.  That's not the way cuberite does it, but it still feels natural to me (and it's the way vanilla does it).

You say: "Because the `cEntity` classes are used to store and modify the entity properties, they should not be used to handle the Protocol.".  My response is that writing entity properties, both to disk and to the network, is equally so a behavior of something that would store entity properties.  Or, at least, it isn't illogical to put it there, just different from the way that cuberite's currently doing it.

Considering metadata again: having a generic metadata storage object was previously mentioned.  While I like that approach, as I've mentioned before it doesn't work with multiple protocol versions.  But if we could ignore that (this is all hypothetical - it doesn't seem possible to ignore it), then what's the practical difference between having a `cEntity::WriteMetadata` and that storage object?  In both cases, there's a dependency upon the protocol-specific IDs (or have protocol-specific keys of some sort). (Outside of the scope of IDs, though, a storage object would allow detecting changes to metadata and only send the changed values, which would be nice).

&nbsp;

I'm going to try to make a different argument, one that isn't _immediately_ relevant but will be important for the (distant) future: mods.  Obviously, this won't be happening any time soon, but it's something that needs to be considered.

Assuming that we allow plugins to define new entities, they're going to have to be able to send metadata for those entities.  To give an example of how an existing forge mod does this, take a look at [Butterflies from Forestry's lepidopterology section](https://github.com/ForestryMC/ForestryMC/blob/mc-1.10/src/main/java/forestry/lepidopterology/entities/EntityButterfly.java).  (Why forestry?  It was the first 1.10 mod that I found.)  You can see that it creates several `DataParameter` objects, registers them with the `dataManager` in `entityInit`, and saves/loads NBT within `writeEntityToNBT` and `readEntityFromNBT`.  So a plugin would need to be able to define both what metadata the mod entity has, and how it saves itself.  

I'm not going to try and propose a mechanism for defining new entity classes at runtime via lua - that's out of my league.  But what I do think is that being able to just write methods within the class for new entities is the best way for modding.  You could have a metadata writer registry and a NBT writer registry, but that's just going to get messy.  And mods can also extend vanilla classes (for instance, custom minecart types) - it's not feasible to just make mod entities support overriding; you need to allow doing it with regular entities too.

There's the issue of multiple mod versions.  It's best not to think about that too much right now, but it still must be considered - probably, there won't be support for multiple versions of the same mod at the same time, but if a mod wants to do that, then it'd need to compare mod versoins.

bibo38:
Just got an idea, which could solve the problem and would make @madmaxoft happy.

Create an empty class hierarchy, which matches the Minecraft class hierarchy. These classes all have a virtual `WriteEntityMetadata(int & counter, cEntity * entity, EntityWriter * writer)` method. Now you can implement these methods like the current ones.

Now you just need the switch and then call the appropriate WriteEntityMetadata method.

> That's not possible with the current layout - you need a switch upon the different eEntityTypes, and then if it's etMonster, you need to switch on eMonsterType.

I don't understand the switch problem. For me this code looks good:

    switch(entity.getEntityType()) {
        case etTNT: TNTEnitity().WriteMetadata(0, entity, writer); break;
        case etMonster: switch(getMonsterType()) {
            case mtBat: BatEntity().WriteMetadata(0, entity, writer); break;
        }
        break;
    }



What do you think about this approach?

Pokechu22:
My first thought on seeing that: I like the syntax.

I'm not sure how creating additional classes would solve the inheritance issue, though.  You couldn't have the writer for slimes and the writer for magma cubes be the same, because `cSlime::GetSize` and `cMagmaCube::GetSize` aren't the same method (unless templates let you work around that).  Basically, to get that to work you'd still want Cuberite's entity class hierarchy to match Minecraft's.  That doesn't mean having the additional classes would be bad, but it doesn't get us out of having to change the hierarchy.

Another thought: how would the individual classes handle different protocol versions?  Using the version number would probably be the only way (which isn't _really_ an issue, but it's still not ideal).  I already have code that depends on the version number, so that's probably fine.

I'm curious why you have `int & counter` as a parameter in your sample.  The way that I've done it, the `EntityWriter` equivalent keeps track of the count, which seems "cleaner" to me - you'd only call `writer.WriteInt(value)` and not `writer.WriteInt(counter++, value)`.  Is there any specific reason you did it that way, or was it just part of the way you did the sample?

I'll try and create a mock-up of this (probably based off of what I did with method 3).

> I don't understand the switch problem. For me this code looks good

Honestly I'm not sure why the switch bothers me so much.  It doesn't really seem like a problem when looked at objectively.  Something about it being nested just irks me.  I'll not worry about that for now, though.

-----

Also, I just noticed this:

> It's like adding a "SendEntity" method to cEntity.

There's `cEntity::SpawnOn(cClientHandle & a_Client)` (and `cMonster::SpawnOn(cClientHandle & a_Client)` and several other overrides).  It selects (effectively) which packet to use by calling `SendSpawnMob` or `SendSpawnObject` or whichever.  That is at a higher level than what I'm doing here, but it's the same general idea.

-----

What I'm now thinking about is 1.11.  1.11 adds a bunch of new entities, but that's easy to handle.  The part of 1.11 that is going to be difficult is the way they split a bunch of entities - where there was just `minecraft:zombie` and a `Type` metadata field, there is now `minecraft:zombie`, `minecraft:husk`, and `minecraft:zombie_villager` (and technichally also `minecraft:zombie_pigman`).  This is somewhat hard to deal with - each of those has a different ID used with spawn mob (while previously, they were the same ID).  And each one had didfferent metadata fields.  The same problem applies to horses: `minecraft:horse`, `minecraft:zombie_horse`, `minecraft:skeleton_horse`, `minecraft:donkey`, `minecraft:mule`, and `minecraft:llama`.  All of which share some metadata, but not a lot of it - `minecraft:horse` has fields that `minecraft:skeleton_horse` doesn't (armor and variant).  Achieving compatibility with multiple protocols that way would be a pain.  We probably should figure out how that will work - sending the right entity IDs would be difficult, and choosing the right metadata would be important.

madmaxoft:
I'd really prefer not to have the separate class hierarchy either. One thing that should be kept in mind while designing this: There will be an entity (or mob only?) overhaul in the future. We're going to unify all mobs (or entities) into a single class which will use components for different things - one of the components will be the "game visuals" - whether the entity is displayed as a zombie or a spider. Other components may be:
- Pathfinding abilities (imagine zombies climbing walls like spiders, ghasts could swim in water etc.)
- AI
- Damage resistance and immunity
- ...
So the separate class hierarchy will more or less mirror the Game visuals component.

Is the approach I proposed really that bad that you don't want to use it no matter what?

Pokechu22:
I completely forgot about #3295.  That does change the needed approach.

> I'd really prefer not to have the separate class hierarchy either.
> So the separate class hierarchy will more or less mirror the Game visuals component.

This is a very interesting point, and changes some of the arguments I've made before.

I think that it actually _does_ make sense to have the separate class hierarchy, still, but it would be _for_ the game visuals component.  These classes would _store_ the metadata fields and send them; the regular entity classes would manipulate it.  Or something like that; I'll have to experiment but it does seem like it would make sense to write it in a way that is compatible with #3295's goals.

> Is the approach I proposed really that bad that you don't want to use it no matter what?

I feel like your approach is better than the way it's currently done.  I'm just not sure if it's the best approach, especially given other changes.

madmaxoft:
I still think the entire network protocol should be encapsulated in the `cProtocol` descendant classes and should not go anywhere else. As simple as that.

Pokechu22:
I disagree with that.  In my opinion, it makes more sense to have metadata writing within the entities themselves (or at least, the same amount of sense, which combines with my existing biases to generate a preference).  

It's been said that putting entity metadata writing in the entities is a violation of single responsibility - entities are responsible for storing values AND serializing them.  I think that argument applies equally to putting it in the protocol classes though: in that case, the protocol classes are handling packets AND metadata (and several other things).

I think the reason why I dislike the `switch` that's previously been mentioned is that it requires adding a new `case` for any new entities (or entity renderers), which means that it's **impossible to add new entities (or entity renderers) via plugins at runtime**.  Thus using a `switch` for that makes it impossible to support mods that add new entities via plugins that are loaded at runtime.  (And, from the way that I understand templates, you also couldn't use templates to do this at runtime either).

Now, I might have made an incorrect assumption: I've been assuming that you could extend C++ classes from within lua.  That might not actually be the case.  If there isn't any way of doing it, the main benefit I thought that putting the metadata writing within `cEntity` would grant isn't an actual benefit.  So we probably shouldn't actually do it that way.

Or maybe I don't understand lua and both/neither of these methods would work.

I still think splitting the metadata storage and the entity logic would fit well with #3295 - that I can work on.

Cl1608Ho:
If I understand it correctly, Metadata should rather not go into cProtocol, because someone pointed out it that we should keep things a bit seperated and not tie everything to cProtocol, to make fake players possible :P

PR #3381, opened 2016-09-24
Fixed cCompositeChat's LuaAPI bindings.

madmaxoft:
The tolua-generated constructor would return an extra string value.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -3698,11 +3698,13 @@ end
 				},
 				Clear =
 				{
+					Returns = "self",
 					Notes = "Removes all parts from this object",
 				},
 				constructor =
 				{
 					{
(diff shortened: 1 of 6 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -3703,6 +3703,7 @@ end
 				constructor =
 				{
 					{
+						Returns = { {Type = "cCompositeChat"} },
```
NiLSPACE: Shouldn't all constructors return the class they are in?

madmaxoft: Yes. But I don't want to force it in the simulator, because this way I can find the problematic exports

madmaxoft:
In the end I've checked and corrected a lot of other `cCompositeChat`'s API. I found out some functions were exported both manually and automatically, which could have caused trouble.

PR #3391, opened 2016-10-03
APIDump: Check param and return types.

madmaxoft:
This adds a check to all PRs that the Lua API documentation doesn't use param / return types that are not declared in the API.


madmaxoft:
If there are no objections, I'd like to merge this as soon as possible.

madmaxoft:
Thanks guys :)

PR #3395, opened 2016-10-08
Initialize m_bIsInBed in cPlayer constructor.

sgtbigman:
This fixes an issue where the return value from the IsInBed method of
the cPlayer class in the lua API would not return a boolean value.


madmaxoft:
Good catch, thanks.

PR #3396, opened 2016-10-09
Teleportation now uses QueueTask

### LINKED ISSUE #3389: TeleportToCoords not working after player switches world
Client version:   1.10.2
Server OS:        Linux

### Expected behavior
Players should teleport to the correct coordinates specified in TeleportToCoords after switching worlds.

### Actual behavior
The player doesn't teleport to the coordinates, and stays at the spawn point. When the player doesn't have to switch worlds, it teleports to the correct coordinates.

Try commands like /tp, /warp in Core and Essentials.

LogicParrot:
Closes #3389

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1447,13 +1447,16 @@ void cPlayer::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
 	//  ask plugins to allow teleport to the new position.
 	if (!cRoot::Get()->GetPluginManager()->CallHookEntityTeleport(*this, m_LastPosition, Vector3d(a_PosX, a_PosY, a_PosZ)))
 	{
-		SetPosition(a_PosX, a_PosY, a_PosZ);
-		FreezeInternal(GetPosition(), false);
-		m_LastGroundHeight = a_PosY;
-		m_bIsTeleporting = true;
-
-		m_World->BroadcastTeleportEntity(*this, GetClientHandle());
-		m_ClientHandle->SendPlayerMoveLook();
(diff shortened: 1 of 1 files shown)


src/Entities/Player.cpp
```diff
@@ -1447,13 +1447,16 @@ void cPlayer::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
...
 	if (!cRoot::Get()->GetPluginManager()->CallHookEntityTeleport(*this, m_LastPosition, Vector3d(a_PosX, a_PosY, a_PosZ)))
 	{
-		SetPosition(a_PosX, a_PosY, a_PosZ);
-		FreezeInternal(GetPosition(), false);
-		m_LastGroundHeight = a_PosY;
-		m_bIsTeleporting = true;
-
-		m_World->BroadcastTeleportEntity(*this, GetClientHandle());
-		m_ClientHandle->SendPlayerMoveLook();
+		m_World->QueueTask([this, a_PosX, a_PosY, a_PosZ](cWorld & a_World)
```
madmaxoft: No, this is not good. If the player disconnects between the enqueue and the execution of the task, the server will crash. You cannot capture `this`.

LogicParrot: Good point. This also means `DoMoveToWorld` may be currently unsafe.
I also found out `ScheduleMoveToWorld` is buggy when called by plugins.
I am working on a proper fix all 3 methods.

Can we merge this first? warps are currently broken and they are critical for many servers. The fix is worth the minor risk of a crash. I will be fixing this fully very soon.

madmaxoft: Just capture the Player's UniqueID and use `DoWithEntityByID`

LogicParrot: Ok.

PR #3399, opened 2016-10-10
cPlayer::DoMoveToWorld now thread safe

### LINKED ISSUE #3389: TeleportToCoords not working after player switches world

### LINKED ISSUE #3398: World travel API has some bugs
Commit id:        1e8cae641592cc4ae7b6f1b997c514f9d2be422b

We have problems with the world travel API.

##### cPlayer::MoveToWorld
MoveToWorld is exposed directly to plugins. This is not thread safe. I've mitigated this by wrapping some of `DoMoveToWorld` with `QueueTask`, but this is not enough. Also, it introduces a bug; it is dangerous for `QueueTask` to capture a `cPlayer`. Although `DoMoveToWorld` is marked obsolote in the API docs, it has no real alternative. See below. 

##### cPlayer::ScheduleMoveToWorld
This is supposed to be the safer alternative, but it's broken. It assumes it is always called by nether/end portal travel, therefore it may send the wrong respawn packets due to a wrong `a_shouldSendRespawn` flag.

I'll fix `ScheduleMoveToWorld`.

I think we should either remove `MoveToWorld ` from the API, or make it a wrapper for `ScheduleMoveToWorld`.

Related: #3179 #3396

LogicParrot:
@tigerw @madmaxoft I would appreciate feedback. 
- is this safe?
- Does this close #3398 (not to be confused with #3389) by making `DoMoveToWorld` safe and `ScheduleMoveToWorld` unnecessary?

selfnote: figure out if this is really needed

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1775,7 +1775,8 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	ASSERT(a_World != nullptr);
 	ASSERT(IsTicking());
 
-	if (GetWorld() == a_World)
+	cWorld * OldWorld = GetWorld();
+	if (OldWorld == a_World)
 	{
 		// Don't move to same world
 		return false;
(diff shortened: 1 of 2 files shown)


src/Entities/Player.cpp
```diff
@@ -1775,78 +1775,102 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
...
+		cWorld *OldWorld = GetWorld();
+		OldWorld->QueueTask([this, ParentChunk, a_World](cWorld & a_OldWorld)
+		{
+			LOGD("Warping player \"%s\" from world \"%s\" to \"%s\". Source chunk: (%d, %d) ",
+				this->GetName().c_str(),
+				a_OldWorld.GetName().c_str(), a_World->GetName().c_str(),
+				ParentChunk->GetPosX(), ParentChunk->GetPosZ()
+			);
+			ParentChunk->RemoveEntity(this);
+			a_World->AddPlayer(this, &a_OldWorld);  // New world will appropriate and announce client at his next tick
```
LogicParrot: Fix the wording in the comment. (old wording was good)

LogicParrot:
Note that `DoMoveToWorld` can currently be called from plugins (outside a tick thread)

LogicParrot:
I think the `DoWithPlayer` approach is flawed because it does not guarantee execution of the code inside the tick thread, unlike the queueTask approach. I believe the proper approach is a `DoWithPlayer` encapsulated in a `Queuetask`.

madmaxoft:
I think this is okay, just some minor style issues:
- member variables should have a `m_` prefix. The `a_` prefix is used for function arguments
- Initialize member variables in constructor, rather than explicitly assigning

Rebase onto master, too?

madmaxoft:
No, still not right. Still capturing `this` into a queued task.

I think it should be done this way:
1. `DoMoveToWorld()` schedules a task onto the current world's tick thread
2. Current world's task asks the plugins for confirmation, clears chunks and removes the player entity. Then it queues a task onto the new world's tick thread
3. The new world's task adds the player.

Note that these calls should not communicate using the `cPlayer` object, they need an indirect reference. This will prove extremely problematic in step 3.

LogicParrot:
Do plugins always work within the tick thread?

madmaxoft:
No, they can have a callback that gets executed in almost any thread.

LogicParrot:
I am asking because I am questioning the need for #3113, it is possible that the code prior to it was just fine.

>No, they can have a callback that gets executed in almost any thread.

Is it the plugin developer's responsibility to be careful about e.g. not executing `SetPosition` in the wrong thread? Or should that method have internal safeguards?

madmaxoft:
There's no clear indication in the API, and no clear consensus. We do have a "Webserver vs world thread" article in the API docs, but that's very one-sided and I don't think many plugin developers actually bother to even read it.

From another point of view, we can't have everything asynchronous, writing a plugin for such a system would be a nightmare (`SetPosition(X, Y, OnPositionSetCallback)` - ugh!). There needs to be a better way. For many of the API functions, threading simply doesn't matter (much) - if the position is read by another thread wrong for one tick is not a critical issue.

LogicParrot:
And yet, if `DoMoveToWorld` is being called outside a tick thread, then a `cPlayer` was definitely misused by the plugin dev and somehow made it past a tick thread. Perhaps by being stored raw in a plugin's variable. right?

madmaxoft:
Not necessarily, any thread can use `cWorld:DoWithPlayerByUUID(uuid, callbackFn)` and the `callbackFn` is executed in that thread, while having a valid `cPlayer` object.

LogicParrot:
I see. So, in a sense, we are playing around with undefined behavior, because  `cPlayer` has many methods that are meant for the tick thread.

src/Entities/Player.cpp
```diff
@@ -1788,62 +1789,85 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 		return false;
```
LogicParrot: Is this safe, regardless of which thread calls `DoMoveToWorld`?

madmaxoft:
On the other hand, having a valid `cPlayer` instance means that the list of players is locked through its CS by the current thread, so it should be actually safe to use anything

LogicParrot:
I have applied some changes. As for #3389, I'd say it was the Essentials' fault. The plugin performed a world travel to the spawnpoint followed immediately by a teleport to the proper destination. That happened to work, however the API docs say

```
the[world travel] move is asynchronous - the entity is moved into a queue and will be moved from that queue into the destination world at some (unpredictable) time in the future. 
```
The unpredictability changed a bit and the plugin broke. I have submitted a fix here: https://github.com/cuberite/Essentials/pull/64

LogicParrot:
>On the other hand, having a valid cPlayer instance means that the list of players is locked through its CS by the current thread, so it should be actually safe to use anything

But is locking the player list enough? I believe an entity in a tick thread may call `cPlayer::something` without locking the player list.

LogicParrot:
Oh wait, the world tick locks that list, doesn't it?

LogicParrot:
>On the other hand, having a valid cPlayer instance means that the list of players is locked through its CS by the current thread, so it should be actually safe to use anything

For this to be true, the tick thread must lock the player list. I am not sure this is happening.

madmaxoft:
It does. See `cWorld:TickClients()`.

LogicParrot:
But we also have mob ticking. A mob may hold a player as its `m_Target` or as some other pointer, and then mutate the player via a mob tick, which happens in `cWorld::TickMobs`. That method does not hold the lock.

LogicParrot:
Of course one could argue mobs should always access players via `DoWithPlayerByUUID` and never directly, but I really doubt this is the current reality.

PR #3404, opened 2016-10-13
Fixed APIDump error

mathiascode:
A curly bracket in APIDesc.lua broke the APIDump plugin and automatic builds.


mathiascode:
Sorry, didn't see that #3400 contains a fix.

PR #3407, opened 2016-10-16
APIDump: Added missing syntax to cUrlClient

NiLSPACE:


madmaxoft:
This shouldn't be needed at all, the `IsOptional` member already indicates that the params are optional.

NiLSPACE:
The first syntax accepts an url and a table, while the other accepts an url and a function.

madmaxoft:
Oh, right, I'm blind. In it goes, then.

PR #3409, opened 2016-10-17
Add Block hardness.

mohe2015:
The block hardness can be used for breaking block checks later.


src/BlockInfo.h
```diff
@@ -70,6 +70,9 @@ class cBlockInfo
 	/** Sound when placing this block */
 	AString m_PlaceSound;
 
+	/** Defines how long it takes to break the block */
```
madmaxoft: What are the units?

src/BlockInfo.cpp
```diff
@@ -802,6 +802,211 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
...
+	a_Info[E_BLOCK_BIRCH_DOOR          ].m_Hardness = 3;
+	a_Info[E_BLOCK_JUNGLE_DOOR         ].m_Hardness = 3;
+	a_Info[E_BLOCK_ACACIA_DOOR         ].m_Hardness = 3;
+	a_Info[E_BLOCK_DARK_OAK_DOOR       ].m_Hardness = 3;
+	// Synonym or ID compatibility
+	a_Info[E_BLOCK_YELLOW_FLOWER] = a_Info[E_BLOCK_DANDELION];
+	a_Info[E_BLOCK_RED_ROSE] = a_Info[E_BLOCK_FLOWER];
+	a_Info[E_BLOCK_WOODEN_DOOR] = a_Info[E_BLOCK_OAK_DOOR];
+	a_Info[E_BLOCK_FENCE_GATE] = a_Info[E_BLOCK_OAK_FENCE_GATE];
+	a_Info[E_BLOCK_WOODEN_STAIRS] = a_Info[E_BLOCK_OAK_WOOD_STAIRS];
```
madmaxoft: Style: Align the block so that the `=`s are at the same column.

src/BlockInfo.h
mohe2015: That needs further investigation. They are the numbers from net.minecraft.block.Block.blockHardness.

mohe2015:
I forgot to document the API changes...

src/BlockInfo.h
mohe2015: done (api docs).

mohe2015:
Done. I'll submit an anti fastbreak module which uses the block hardness soon. Should anticheat modules be implemented directly or using something like #ifdef ANTICHEAT ?

Pokechu22:
Make sure that it doesn't trigger in creative.  Also, don't forget about (/reimplement) [MC-5694](https://bugs.mojang.com/browse/MC-5694) - if the block break is canceled due to it happening too fast, re-send the block!  (Otherwise the client will think it's missing and the server won't - since cuberite doesn't have block collision you won't have the annoying "can't move into that air block" effect, but it's still annoying to have random floating blocks that you thought you removed).

IMO an `#ifdef ANTICHEAT` is not needed for _this_ behavior as it's a fundamental property of the game that generally doesn't need to be turned off.  It would make sense for things that are more fancy though such as trying to catch people flying/speeding.

Pokechu22:
It might make sense to have a configuration setting for the tolerance, though - my brief impression from looking at `net.minecraft.server.management.PlayerInteractionManager.blockRemoving(BlockPos)` is that it only needs to be 70% dug for the server to completely accept it (but the units there are very strange so it might not be 70%).  

The accuracy of it partially would depend on the TPS - if it needs to be 100% broken for it to actually break serverside, then any lag would cause the break to fail.

So a configuration option for tolerances would probably be ideal, with 70% being the default and allowing ranges of 0% (disabled) to 100% (probably with a warning).

mohe2015:
Thanks @Pokechu22.

Closed. See  #3411 which actually uses the block hardness.

PR #3410, opened 2016-10-19
Added SendMessageRaw for sending json string.

Seadragon91:
This allows a developer to send a json string to the client. I added the optional value `eChatType` that is `ctChatBox` as default. Should I add a info about `cJson` to the description?

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11698,6 +11698,22 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Sends the specified message to the player.",
 				},
+				SendMessageRaw =
+				{
+					Params =
+					{
+						{
+							Name = "Json",
+							Type = "string",
+						},
+						{
+							Name = "eChatType",
+							Type = "number",
+							IsOptional = true,
+						}
+					},
+					Notes = "Sends the specified json string to the player. The optional value eChatType (default ctChatBox) can be ctChatBox, ctSystem or ctAboveActionBar. You can use {{cJson}} to build a json string.",
+				},
 				SendMessageFailure =
 				{
 					Params =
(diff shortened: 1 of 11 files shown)


Seadragon91:
Corrected style fixes and removed unused boolean.

src/ClientHandle.cpp
```diff
@@ -2271,6 +2271,24 @@ void cClientHandle::SendChat(const cCompositeChat & a_Message)
 
 
 
+void cClientHandle::SendChatRaw(const AString & a_MessageRaw, eChatType a_Type)
+{
+	cWorld * World = GetPlayer()->GetWorld();
```
NiLSPACE: What do you need the world object for?

Seadragon91: Ops :)

Seadragon91: Removed

Seadragon91:
I added a info about `cJson`.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11698,6 +11698,22 @@ a_Player:OpenWindow(Window);
...
+							Name = "Json",
+							Type = "string",
+						},
+						{
+							Name = "eChatType",
+							Type = "number",
+							IsOptional = true,
+						}
+					},
+					Notes = "Sends the specified json string to the player. The optional value eChatType (default ctChatBox) can be ctChatBox, ctSystem or ctAboveActionBar. You can use {{cJson|cJson}} to build a json string.",
```
madmaxoft: You don't need the second `|cJson` in the link. The syntax is `{{url|linktext}}` or `{{url}}` (if the link text is the same as url; url can be a class name. So you can use `{{cJson}}` or `{{cJson|the built in JSON serializer}}` etc.

Seadragon91:
Okay for merge?

madmaxoft:
It'd be great for the API function to take either a raw string, or a Lua table which it would automagically serialize into json. But that's something for a separate PR.

PR #3411, opened 2016-10-19
Anticheat fastbreak

mohe2015:


src/Entities/Player.cpp
```diff
@@ -2655,12 +2655,12 @@ float cPlayer::GetLiquidHeightPercent(NIBBLETYPE a_Meta)
 
 bool cPlayer::IsInsideOfBlock(BLOCKTYPE a_Block)
 {
-	BLOCKTYPE Block = m_World->GetBlock(GetPosX(), m_Stance, GetPosZ());
+	BLOCKTYPE Block = m_World->GetBlock((int)GetPosX(), (int)m_Stance, (int)GetPosZ());
```
Seadragon91: Please use `static_cast<int>(value)`, instead of `(int)`.

mohe2015: I'm not so familiar with C++. What's the difference?

mohe2015: I found it out.

Seadragon91: `(int)` is used in c `and static_cast<int>()` is the replacement for it in c++. Also the compiler would warn of use of `old-style cast`, if there is no flag to ignore it.

src/Entities/Player.cpp
```diff
@@ -2636,3 +2636,101 @@ void cPlayer::FreezeInternal(const Vector3d & a_Location, bool a_ManuallyFrozen)
...
+	return (float)(a_Meta + 1) / 9.0f;
+}
+
+
+
+
+
+bool cPlayer::IsInsideOfBlock(BLOCKTYPE a_Block)
+{
+	BLOCKTYPE Block = m_World->GetBlock(static_cast<int>(GetPosX()), static_cast<int>(m_Stance), static_cast<int>(GetPosZ()));
```
madmaxoft: Use `FloorC` instead of the `static_cast`s, to make it explicit in what direction the coords are truncated.

src/Entities/Player.cpp
```diff
@@ -2636,3 +2636,101 @@ void cPlayer::FreezeInternal(const Vector3d & a_Location, bool a_ManuallyFrozen)
...
+	if (Block != a_Block)
+	{
+		return false;
+	}
+	NIBBLETYPE Meta = GetWorld()->GetBlockMeta(static_cast<int>(GetPosX()), static_cast<int>(m_Stance), static_cast<int>(GetPosZ()));
+	float f = GetLiquidHeightPercent(Meta) - 0.11111111f;
+	float f1 = static_cast<float>(m_Stance + 1) - f;
+	bool flag = m_Stance < f1;
```
madmaxoft: Parentheses around comparison

src/Entities/Player.cpp
```diff
@@ -2636,3 +2636,101 @@ void cPlayer::FreezeInternal(const Vector3d & a_Location, bool a_ManuallyFrozen)
...
+			break;
+		case 1:
+			f *= 0.09f;
+			break;
+		case 2:
+			f *= 0.0027f;
+			break;
+		default:
+			f *= 8.1e-4f;
+		}
```
madmaxoft: Bad `switch` formatting. Use an extra indent level for the `case` statements, and use braces for each case:
```cpp
switch(...)
{
  case 0:
  {
    ...
    break;
  }
  case 1:
  ...
}
```

madmaxoft: In this simple case, you may condense each case into a single line:
```cpp
switch (intensity)
{
	case 0: f *= 0.3f; break;
	case 1: f *= 0.09f; break;
	case 2: f *= 0.0027f; break;
	default: f *= 0.00081f; break;
}
```

src/Entities/Player.cpp
```diff
@@ -2636,3 +2636,101 @@ void cPlayer::FreezeInternal(const Vector3d & a_Location, bool a_ManuallyFrozen)
...
+
+
+
+float cPlayer::GetPlayerRelativeBlockHardness(BLOCKTYPE a_Block)
+{
+	float blockHardness = cBlockInfo::GetHardness(a_Block);
+	float digSpeed = GetDigSpeed(a_Block);
+	float canHarvestBlockDivisor = GetEquippedItem().GetHandler()->CanHarvestBlock(a_Block) ? 30.0f : 100.0f;
+	// LOGD("blockHardness: %f, digSpeed: %f, canHarvestBlockDivisor: %f\n", blockHardness, digSpeed, canHarvestBlockDivisor);
+	return blockHardness < 0 ? 0 : digSpeed / blockHardness / canHarvestBlockDivisor;
```
madmaxoft: Parentheses around comparison, parenthesize the multi-division as well to indicate the intended order.

src/Entities/Player.h
```diff
@@ -528,6 +528,9 @@ class cPlayer :
 	The player removes its m_ClientHandle ownership so that the ClientHandle gets deleted. */
 	void RemoveClientHandle(void);
 
+	/** Returns the relative block hardness for the block a_Block */
```
madmaxoft: Relative to what? What stats does this take into account? (block type, player surroundings, equipped tool...)

src/Entities/Player.cpp
```diff
@@ -2636,3 +2636,101 @@ void cPlayer::FreezeInternal(const Vector3d & a_Location, bool a_ManuallyFrozen)
...
+			break;
+		case 2:
+			f *= 0.0027f;
+			break;
+		default:
+			f *= 8.1e-4f;
+		}
+	}
+
+	if ((IsInsideOfBlock(E_BLOCK_WATER) || IsInsideOfBlock(E_BLOCK_STATIONARY_WATER)) && !(GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::eEnchantment::enchAquaAffinity) > 0))
```
madmaxoft: Wrap this `if` statement onto multiple lines based on the "tree" of operations.
How about lava, doesn't that impair digging speed as well?

src/Entities/Player.h
```diff
@@ -709,4 +712,14 @@ class cPlayer :
 	/** Pins the player to a_Location until Unfreeze() is called.
 	If ManuallyFrozen is false, the player will unfreeze when the chunk is loaded. */
 	void FreezeInternal(const Vector3d & a_Location, bool a_ManuallyFrozen);
+
+	/** Returns how high the liquid is in percent. Used by IsInsideOfBlock */
+	float GetLiquidHeightPercent(NIBBLETYPE a_Meta);
+
+	/** Checks if the player is inside of the block (use only for liquid!) */
+	bool IsInsideOfBlock(BLOCKTYPE a_Block);
```
madmaxoft: How about renaming to `IsInsideLiquidBlock()` that checks against all liquid types at once?

src/Entities/Player.h
```diff
@@ -709,4 +712,14 @@ class cPlayer :
...
+
+	/** Returns the dig speed using the current tool on the block a_Block */
```
madmaxoft: What units?

src/Items/ItemHandler.h
```diff
@@ -154,6 +154,10 @@ class cItemHandler
 	Defaults to false unless overridden. */
 	virtual bool CanHarvestBlock(BLOCKTYPE a_BlockType);
 
+	/** Returns the strength to break a specific block.
+	Defaults to 1 unless overriden. */
+	virtual float GetStrVsBlock(BLOCKTYPE a_Block);
```
madmaxoft: Use full wording for function name. How about `GetBlockBreakingStrength`?

src/Items/ItemPickaxe.h
```diff
@@ -107,6 +107,28 @@ class cItemPickaxeHandler :
...
 		return false;
 	}
+
+
+	virtual float GetStrVsBlock(BLOCKTYPE a_Block)
+	{
+		if (!IsBlockMaterial_IRON(a_Block) && !IsBlockMaterial_ANVIL(a_Block) && !IsBlockMaterial_ROCK(a_Block))
+		{
+			return super::GetStrVsBlock(a_Block);
+		} else
```
madmaxoft: `else` on separate line

src/Items/ItemPickaxe.h
```diff
@@ -107,6 +107,28 @@ class cItemPickaxeHandler :
...
+		{
+			switch (m_ItemType)
+			{
+				case E_ITEM_WOODEN_PICKAXE: return 2.0f;
+				case E_ITEM_STONE_PICKAXE:  return 4.0f;
+				case E_ITEM_IRON_PICKAXE:   return 6.0f;
+				case E_ITEM_GOLD_PICKAXE:   return 12.0f;
+				case E_ITEM_DIAMOND_PICKAXE: return 8.0f;
```
madmaxoft: Align the `return`s

src/Items/ItemPickaxe.h
```diff
@@ -107,6 +107,28 @@ class cItemPickaxeHandler :
...
+			}
+		}
+		ASSERT(!"Something is wrong here... ask mohe2015");
```
madmaxoft: Don't ask, explain. mohe2015 might not be available in two years' time.

src/Items/ItemShears.h
```diff
@@ -90,6 +90,14 @@ class cItemShearsHandler :
 			}
 		}
 	}
+
+
+
+	virtual float GetStrVsBlock(BLOCKTYPE a_Block)
+	{
+		return !IsBlocks_WEB(a_Block) && !IsBlockMaterial_LEAVES(a_Block) ? (IsBlocks_WOOL(a_Block) ? 5.0f/* WOOL */ : super::GetStrVsBlock(a_Block) /* none of them */) : 15.0f /* WEB and LEAVES */;
```
madmaxoft: Way too complicated, use a proper `if` statement and break this into smaller parts.

src/Items/ItemShovel.h
```diff
@@ -59,4 +59,24 @@ class cItemShovelHandler : public cItemHandler
 		return false;
 	}
 
+	virtual float GetStrVsBlock(BLOCKTYPE a_Block)
+	{
+		if ((a_Block == E_BLOCK_CLAY) || (a_Block == E_BLOCK_DIRT) || (a_Block == E_BLOCK_FARMLAND) || (a_Block == E_BLOCK_GRASS) || (a_Block == E_BLOCK_GRAVEL) || (a_Block == E_BLOCK_MYCELIUM) || (a_Block == E_BLOCK_SAND) || (a_Block == E_BLOCK_SNOW) || (a_Block == E_BLOCK_SNOW_BLOCK) || (a_Block == E_BLOCK_SOULSAND)/* a_Block == 208 (GRASS_PATH, not implemented yet) */)
```
madmaxoft: Use a `switch` instead of this long `if`

src/Items/ItemShovel.h
```diff
@@ -59,4 +59,24 @@ class cItemShovelHandler : public cItemHandler
...
+		{
+			switch (m_ItemType)
+			{
+				case E_ITEM_WOODEN_SHOVEL: return 2.0f;
+				case E_ITEM_STONE_SHOVEL:  return 4.0f;
+				case E_ITEM_IRON_SHOVEL:   return 6.0f;
+				case E_ITEM_GOLD_SHOVEL:   return 12.0f;
+				case E_ITEM_DIAMOND_SHOVEL: return 8.0f;
```
madmaxoft: Align

src/Items/ItemShovel.h
```diff
@@ -59,4 +59,24 @@ class cItemShovelHandler : public cItemHandler
...
+			}
+		} else
```
madmaxoft: separate line

src/Items/ItemSword.h
```diff
@@ -56,6 +56,20 @@ class cItemSwordHandler :
...
+
+
+	virtual float GetStrVsBlock(BLOCKTYPE a_Block)
+	{
+		if (a_Block == E_BLOCK_COBWEB)
+		{
+			return 15.0f;
+		} else
+		{
+			return !IsBlockMaterial_PLANTS(a_Block) && !IsBlockMaterial_VINE(a_Block) && !IsBlockMaterial_CORAL(a_Block) && !IsBlockMaterial_LEAVES(a_Block) && !IsBlockMaterial_GOURD(a_Block) ? 1.0f : 1.5f;
```
madmaxoft: Break up into smaller parts

mohe2015:
```
			if (IsBlockMaterial_PLANTS(a_Block)
			 || IsBlockMaterial_VINE(a_Block)
			 || IsBlockMaterial_CORAL(a_Block)
			 || IsBlockMaterial_LEAVES(a_Block)
			 || IsBlockMaterial_GOURD(a_Block))
			{
				return 1.5f;
			} else
			{
				return 1.0f;
			}
Would that be better? Or do you have any other idea.
```

madmaxoft: Put the `||` at the end of the line, and use an extra indent level and it'll be alright:
```cpp
	if (
		IsBlockMaterial_PLANTS(a_Block) ||
		IsBlockMaterial_VINE(a_Block) ||
		IsBlockMaterial_CORAL(a_Block) ||
		IsBlockMaterial_LEAVES(a_Block) ||
		IsBlockMaterial_GOURD(a_Block) ||
	)
	{
		return 1.5f;
	}
```
Don't forget that `else` always goes to a separate line.

mohe2015: Thank you very much for your big review. I'll fix everything tomorrow because I'm to tired. (Germany 22:26 - to late for programming for me).

src/BlockInfo.cpp
```diff
@@ -802,6 +802,211 @@ void cBlockInfo::Initialize(cBlockInfoArray & a_Info)
...
+	a_Info[E_BLOCK_JUNGLE_FENCE        ].m_Hardness = 2.0f;
+	a_Info[E_BLOCK_DARK_OAK_FENCE      ].m_Hardness = 2.0f;
+	a_Info[E_BLOCK_ACACIA_FENCE        ].m_Hardness = 2.0f;
+	a_Info[E_BLOCK_SPRUCE_DOOR         ].m_Hardness = 3.0f;
+	a_Info[E_BLOCK_BIRCH_DOOR          ].m_Hardness = 3.0f;
+	a_Info[E_BLOCK_JUNGLE_DOOR         ].m_Hardness = 3.0f;
+	a_Info[E_BLOCK_ACACIA_DOOR         ].m_Hardness = 3.0f;
+	a_Info[E_BLOCK_DARK_OAK_DOOR       ].m_Hardness = 3.0f;
+	// Synonym or ID compatibility
+	a_Info[E_BLOCK_YELLOW_FLOWER       ].m_Hardness = a_Info[E_BLOCK_DANDELION      ].m_Hardness;
```
Pokechu22: This shouldn't be necessary, since the value of the `E_BLOCK_YELLOW_FLOWER` constant is the same as `E_BLOCK_DANDELION` - this is effectively `a_Info[37].m_Hardness = a_Info[37].m_Hardness`.  (Same for the other synonyms)

mohe2015:
I hope I haven't forgotten anything.

madmaxoft:
Search for all `} else` and fix those :P

mohe2015:
Somebody else made that mistake in  src/Blocks/BlockPiston.cpp  xD

src/BlockInfo.h
```diff
@@ -70,7 +70,7 @@ class cBlockInfo
 	/** Sound when placing this block */
 	AString m_PlaceSound;
 
-	/** Block's hardness. If a player hits a block a value gets incremented every tick by the block's hardness * a multiplier (which tool are you using?, enchantings?, effects?, ...) until it is greater or equal to one. Then the block is broken. */
+	/** Block's hardness. The bigger the harder the block. */
```
sphinxc0re: This doesn't make any sense since every block in Minecraft is the same size

src/ClientHandle.h
```diff
@@ -63,6 +63,9 @@ class cClientHandle  // tolua_export
 	static const int MAX_VIEW_DISTANCE = 32;
 	static const int MIN_VIEW_DISTANCE = 1;
 
+	/** The percentage how much a block has to be broken.
+	Should be a value between 0.7 (70% broken) and 1 (100% broken) depending on lag. 
+	Can be set in settings.ini [AntiCheat] FastBreakPercentage=(from 0 to 100) */
 	static float FASTBREAK_PERCENTAGE;
```
sphinxc0re: Shouldn't this be initialized with a default value?

src/Defines.h
```diff
@@ -594,9 +594,9 @@ inline bool IsBlockMaterial_PLANTS(BLOCKTYPE a_BlockType)
 		case E_BLOCK_CARROTS:
 		case E_BLOCK_POTATOES:
 		// Blocks which don't exist in Cuberite yet.
-//  case 199:
-//  case 200:
-//  case 207:
+		case 199:
+		case 200:
+		case 207:
```
sphinxc0re: What are these magic numbers? Are there any constants with the same value?

src/Defines.h
```diff
@@ -655,9 +655,9 @@ inline bool IsBlockMaterial_IRON(BLOCKTYPE a_BlockType)
 		case E_BLOCK_HOPPER:
 		case E_BLOCK_IRON_TRAPDOOR:
 		// Blocks which don't exist in Cuberite yet.
-//  case 210:
-//  case 211:
-//  case 255:
+		case 210:
+		case 211:
+		case 255:
```
sphinxc0re: Again, what are these magic numbers?

src/Defines.h
```diff
@@ -787,15 +799,15 @@ inline bool IsBlockMaterial_ROCK(BLOCKTYPE a_BlockType)
...
-//  case 216:
+		case 201:
+		case 202:
+		case 203:
+		case 204:
+		case 205:
+		case 206:
+		case 213:
+		case 215:
+		case 216:
```
sphinxc0re: Once again, magic numbers 😉

src/BlockInfo.h
Seadragon91: I think he the greater the value the longer you need to break it.

mohe2015: Yes, I'll change it

src/Defines.h
mohe2015: There aren't because these blocks don't exist in cuberite yet. Should I add them in BlockID.h (I think there) or should I add comments what they are?

madmaxoft: Yes, add them to the header file please.

src/ClientHandle.h
mohe2015: It gets initialized in ClientHandle.cpp because it is not possible to initialize it in the header file.

mohe2015: But I'm not so familiar with C++ (rather Java) so you can tell me how to to it right it it is wrong.

mohe2015: Also a default value is not required because it gets initialized in Root.cpp

mohe2015:
There is a problem: There is a gap between minecraft:black_shulker_box (234) and minecraft:structure_block (255). See http://minecraft.gamepedia.com/Data_values#Block_IDs. I have changed it in the source. Please check if my changes are ok.

mohe2015:
How can I fix that error (found undocumented API symbols) ?

NiLSPACE:
It means you've added new things to the Lua API, but didn't add it to the API documentation. You'll have to them here: https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/APIDesc.lua#L14474

mohe2015:
I know but I just added new block constants and they have a regex declared:
`eBlockType =
				{
					Include = "^E_BLOCK_.*",
					TextBefore = [[
						These constants are used for block types. They correspond directly with MineCraft's data values
						for blocks.
					]],
},`

But it doesn't work.

NiLSPACE:
It wants you to document the individual constants.

Pokechu22:
That's done by setting the minimum break percentage to 0% - if it's 0%, then there essentially is no check.

mohe2015:
But we could add something to disable it entirely (for performance)

NiLSPACE:
Execute this tiny script as a plugin to generate some documentation for all blockid's:
```lua
local file = io.open("output.lua", "w");
file:write([[Constants =
{
]])
for key,value in pairs(_G) do
	if (key:match("^E_BLOCK") or key:match("^E_ITEM")) then
		local name = key:match("^E_.-_(.*)"):gsub("_", " "):lower()
		file:write("\t", key, " =\n\t{\n\t\tNotes = \"The ", key:match("E_ITEM") and "itemid" or "blockid", " for a ", name, "\"\n\t},\n")
	end
end
file:close()
```

madmaxoft:
Don't use `id` for a `type`.

mohe2015:
Why can't I find the other block types in APIDesc.lua? I can just find the regex.

madmaxoft:
None of the block types have ever been documented, and that has been before the CI checks. So it has become the baseline against which all PRs are compared - if they introduce new API symbols, they have to be documented, but symbols that haven't been documented before are an exempt from this rule.

What @NiLSPACE  is proposing is to automatically create documentation for all of them at once.

src/Items/ItemShears.h
```diff
@@ -90,6 +90,14 @@ class cItemShearsHandler :
 			}
 		}
 	}
+
+
+
+	virtual float GetBlockBreakingStrength(BLOCKTYPE a_Block)
+	{
+		return !IsBlocksWeb(a_Block) && !IsBlockMaterialLeaves(a_Block) ? (IsBlocksWool(a_Block) ? 5.0f/* WOOL */ : super::GetBlockBreakingStrength(a_Block) /* none of them */) : 15.0f /* WEB and LEAVES */;
```
madmaxoft: This is still too complicated, use proper `if` statements instead of the ternary operator.

src/Defines.h
```diff
@@ -509,6 +512,317 @@ inline bool IsBlockFence(BLOCKTYPE a_BlockType)
 
 
 
+
+// net.minecraft.block.material.Material.WOOD
```
madmaxoft: Please remove all references to Minecraft code in all such comments to these functions.

src/Items/ItemShovel.h
```diff
@@ -59,4 +59,36 @@ class cItemShovelHandler : public cItemHandler
...
+			case E_BLOCK_FARMLAND:
+			case E_BLOCK_GRASS:
+			case E_BLOCK_GRAVEL:
+			case E_BLOCK_MYCELIUM:
+			case E_BLOCK_SAND:
+			case E_BLOCK_SNOW:
+			case E_BLOCK_SNOW_BLOCK:
+			case E_BLOCK_SOULSAND:
+			// Not implemented in cuberite yet.
+			case 208:
```
madmaxoft: Still a magic number here

mohe2015:
@NiLSPACE Thank you for the code!

@madmaxoft Thank you for your explanation.

The complete code I was using:


```
function Initialize(Plugin)
	Plugin:SetName("TestPlugin")
	Plugin:SetVersion(1)


local file = io.open("output.lua", "w");
file:write([[Constants =
{
]])
for key,value in pairs(_G) do
    if (key:match("^E_BLOCK") or key:match("^E_ITEM")) then
        local name = key:match("^E_.-_(.*)"):gsub("_", " "):lower()
        file:write("\t", key, " =\n\t{\n\t\tNotes = \"The ", key:match("E_ITEM") and "itemtype" or "blocktype", " for ", name, "\"\n\t},\n")
    end
end
file:close()


	LOG("Initialized " .. Plugin:GetName() .. " v." .. Plugin:GetVersion())
	return true
end

```

Please tell me if I should change the phrases.

src/Items/ItemShears.h
sphinxc0re: ☹️ I don't like this either. It looks really complicated and ugly...

madmaxoft: It's actually very difficult to understand since it's a nested ternary operator and you'd need to know the exact operator precedence in C++ (which could be different from Java) to decipher which branch is selected when. That's why a clear `if` statement is a much preferred solution.

madmaxoft:
The phrases are okay, but the output is not alpha-sorted like the rest of the docs.
I'll have a look at your test plugin code and help you make it alpha-sort the stuff later on in the day, gotta go now.

mohe2015:
https://www.lua.org/pil/19.3.html should help. Look at the code at the end.

NiLSPACE:
I didn't test this, but I think this is what you want:
```lua
local blocks, idx = {}, 1
for key,value in pairs(_G) do
    if (key:match("^E_BLOCK") or key:match("^E_ITEM")) then
		blocks[idx] = key
		idx = idx + 1
    end
end
table.sort(blocks)

local file = io.open("output.lua", "w");
file:write([[Constants =
{
]])
for _,key in pairs(blocks) do
	local name = key:match("^E_.-_(.*)"):gsub("_", " "):lower()
	file:write("\t", key, " =\n\t{\n\t\tNotes = \"The ", key:match("E_ITEM") and "itemtype" or "blocktype", " for ", name, "\"\n\t},\n")
end
file:close()
```

madmaxoft:
Oh I do know how to sort arrays alright. And @NiLSPACE your code was still wrong, using `pairs` instead of `ipairs`; I fixed it :)

NiLSPACE:
But it would still have worked with 'pairs' instead of 'ipairs'

madmaxoft:
No, because `pairs` doesn't have a defined order of iteration, it can return the pairs in any order.

mohe2015:
You forgot an "end" after line 5. But it worked.

NiLSPACE:
<offtopic>
@madmaxoft Why did you split the items and blocks in different tables? Even without it the table should've been ordered properly.
</offtopic>

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -14489,6 +14507,1804 @@ end
 			},
 			Constants =
 			{
+Constants =
+{
```
madmaxoft: Now there's a duplicate `Constants = {`, making the file syntactically bad.

mohe2015: Fixed it. But why did the CI check not fail?

madmaxoft: Failing to load API descriptions from any of the files is not marked as a hard error in the CI, mainly because the test is "default to success" rather than "default to failure". I'll probably invert the logic of the test soon.

madmaxoft: Check rewritten in #3415

mathiascode:
Can you add the structure void block as well (ID 217)?

src/ClientHandle.cpp
```diff
@@ -1243,13 +1246,32 @@ void cClientHandle::HandleBlockDigFinished(int a_BlockX, int a_BlockY, int a_Blo
...
+	{
+		// Fix for very fast tools.
+		m_BreakProgress += m_Player->GetPlayerRelativeBlockHardness(a_OldBlock);
+		if (m_BreakProgress < FASTBREAK_PERCENTAGE)
+		{
+			LOGD("Break progress of player %s was less than expected: %f < %f\n", m_Player->GetName().c_str(), m_BreakProgress * 100, FASTBREAK_PERCENTAGE * 100);
+			// AntiFastBreak doesn't agree with the breaking. Bail out. Send the block back to the client, so that it knows:
+			m_Player->GetWorld()->SendBlockTo(a_BlockX, a_BlockY - 1, a_BlockZ, m_Player);  // Bug with doors and things like that.
+			m_Player->GetWorld()->SendBlockTo(a_BlockX, a_BlockY, a_BlockZ, m_Player);
+			m_Player->GetWorld()->SendBlockTo(a_BlockX, a_BlockY + 1, a_BlockZ, m_Player);  // Bug with doors and things like that.
```
Pokechu22: Actually, I have the opposite comment.  I'm 99% sure that there are *no* blocks that change neighboring blocks clientside.  Doors only do it serverside.  Doubleplants only do it serverside (and thus experience [graphical glitches](https://bugs.mojang.com/browse/MC-53439) in some cases).  Beds only do it serverside, in all cases (both exploding in the nether and breaking).  Vines only do it serverside.  Etcetera.  So you really only should need to repair the modified block and not the ones above/below/near it.

mohe2015: @Pokechu22 Thanks. You're **almost** right. Strangely doors are the only blocks which change neighboring blocks clientside. If you break the top of a door it works like expected but if you break the bottom of a door the top does not reappear. **_So we have to send the modified block and (if it is a door) the block above._**

Pokechu22: Never mind.  I took a closer look and `onBlockHarvested` _is_ called clientside as well as serverside.  So you do need to resend those for:

* Pistons, when the head is broken (the base is also broken)
* Double plants, maybe... maybe not
* Doors (top/bottom)
* Beds (both halves)

The client also deletes NBT for flower pots and skulls.  So if the block is resent, you need to resend that too.

I'd recommend explicitly handling those blocks rather than doing a general case.

Pokechu22: The thing is that you'd need to resend tile entities too, which gets complicated.  Not super complicated, but it does make things difficult (and could be a performance issue).

mohe2015: That's also an issue if plugins want to cancel block breaking so we should create a function in cPlayer or somewhere which resends the required things around a broken block. Do you know what the vanilla server does if the block Is broken too fast?

Pokechu22: > That's also an issue if plugins want to cancel block breaking so we should create a function in cPlayer or somewhere which resends the required things around a broken block. 

It looks like `cClientHandle::HandleBlockDigFinished` does this; it currently resends the block that was broken and the player's position if it was canceled (look for `CallHookPlayerBreakingBlock`).

Craftbukkit does a similar thing, except it always sends a set block to air packet and then will send an additional set block packet resseting it _only_ if the event is canceled.  It also resends the block's tile entity packet if it has one.  No handling of doors that I can see.  I'll check the bug tracker to see if there's stuff about doors being poorly handled in a bit.

> That's also an issue if plugins want to cancel block breaking so we should create a function in cPlayer or somewhere which resends the required things around a broken block.



> Do you know what the vanilla server does if the block Is broken too fast?

I mentioned this on #3409.  As observable through [MC-5694](https://bugs.mojang.com/browse/MC-5694), the block isn't always resent.  (It might be resent in some cases, and I actually think it is due to "popping back" behavior I've seen, but it isn't always).  In any case the vanilla server probably isn't a good model for this behavior.

sphinxc0re:
Will plugins be able to manipulate the hardness of a Block after this PR?

mohe2015:
No, but why should they? @sphinxc0re

sphinxc0re:
Okay, so it would be possible to create a plugin that makes blocks more or less breakable at e.g. random?

madmaxoft:
Actually with these changes, I believe the plugins could already modify the blocks' hardness - the variable `cBlockInfo::m_Hardness` is exported to Lua read/write.

mohe2015:
Oh, I didn't know that works. @madmaxoft. That's pretty cool.

mohe2015:
@madmaxoft @sphinxc0re There would be a problem if you change the hardness directly: The client would ignore it.

mohe2015:
So you have to give him an effect or something like that anyways.

sphinxc0re:
Is there a way to tell the client the current hardness of a block? Or is that behavior hardcoded into the client? @Pokechu22

mohe2015:
@sphinxc0re It's hardcoded. You would have to give him an effect or an enchantment.

Pokechu22:
Giving an effect or enchantment wouldn't work anyways because of latency - you'd have to give the effect after the player started mining, but by that time the client will have already made progress in mining with the other speed.  It probably would be better to just not allow plugins to change the hardness.

mohe2015:
@Pokechu22 You're right but if you want the player to mine blocks e.g. twice as fast/slow you can just give him an effect.

Pokechu22:
Yea, changing the relative break speeds for all blocks is possible (either with an attribute on the item or with an effect).  But what I meant is that it isn't possible to change the hardness for a single block.

mohe2015:
Why did the CI check fail?

Pokechu22:
At the very bottom of the release build:

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

Which seems to have caused it to treat it as a build error and thus fail everything.  It seems like it froze during or slightly after world generation.

mohe2015:
Yes but why did it freeze? I don't think there is something wrong with the code because the other builds succeed.

sphinxc0re:
Try rebasing and force-pushing

mohe2015:
That does also work. (close and open pull request)

madmaxoft:
Woohoo, finally merged :)

sphinxc0re:
Nice work! 😁

NiLSPACE:
Apparently there is a problem where people get messages while not cheating: https://forum.cuberite.org/thread-2660-post-27185.html#pid27185

Was this configurable?

yangm97:
I thought my machine was possessed or something. I did not bother opening an issue because simply reducing the FastBreak percentage fixed the issue. 
But I think there could be some investigation around it. Like, why the default value work for some but not for others?

Pokechu22:
Leaves break practically instantly, so it might be that the fact that multiple leaves can be broken on the same tick causes issues with the PR.  Not sure exactly.

yangm97:
In my experience, leaves, dirt and other things that break almost instantly didn't have much trouble, but tree logs almost always would trigger FastBreak. Don't know if it's a pattern or coincidence but usually the bottom most log block breaks fine, the others don't.

madmaxoft:
Perhaps the fastbreak threshold should be scaled by the player's ping time, somehow.

mohe2015:
@madmaxoft Basically that's a good idea but what about ping spoofing? I think a better idea would be to lower FastBreakPercentage in settings.ini because I tested it on a local server and 97 is a very high value.

madmaxoft:
How does ping spoofing work? Can it alter the ping time both ways, or just increase or just decrease it? We could try to mitigate ping spoofing by using a random ID for the ping, instead of an incrementing counter, and if the spoofing can only decrease the ping time (my assumption), it doesn't pose a threat to the scale-by-ping solution, because the only thing it could do is make the fastbreak anticheat stricter.

mohe2015:
http://wiki.vg/Protocol#Keep_Alive_.28clientbound.29
That means you can increase ping time artificially. The Notchian server already uses random ids so we should also do that. If we keep using an incrementing counter you can probably also decrease ping time.

Cl1608Ho:
Just saying, responding faster to packets than you receive them is very very hard :P I would guess you can only increase the ping time by waiting before responding to packets....

mohe2015:
Faster responses aren't the problem. Just artificially slower ones because they can be used to trick the anticheat if we usw ping time.

PR #3416, opened 2016-10-27
LuaJson: Report serialization errors instead of crashing.

### LINKED ISSUE #3414: Crash in cJson:Serialize / Server restarts in a endless loop
Commit id:        (502b5ec)

Yep there are two possibilities 😄 

### Actual behavior
Server crashes or restarts in a endless loop.

Add this function to a plugin
```
function func(...)
	cJson:Serialize(arg)
end
```
Call it it from the function `Initialize`: `func(1)`

1) Server is not running.
Add the code and start the server, endless loop.

2) Server is running.
Add the code and run reload, server will crash.

madmaxoft:
Fixes #3414

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -9476,7 +9476,7 @@ end
 							Type = "string",
 						},
 					},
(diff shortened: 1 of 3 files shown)


Seadragon91:
Nice adding this `?w=1`, will hide white space and tab changes :-)
https://github.com/cuberite/cuberite/pull/3416/files?w=1

PR #3422, opened 2016-11-06
Updated sounds and effect IDs

mathiascode:
The sound effect names and effect IDs have been updated for 1.9 and 1.10. I removed the placement sounds because they are sent client-side.


Pokechu22:
A few things:

1. The noteblock usage of the block action packet was more recently documented ([here](http://wiki.vg/Block_Actions#Note_Block)).  You don't need to (and shouldn't) actually send a play sound packet in addition to the block action; if you do it'll create two of the same sound (effectively making it twice as loud as it should be, IIRC)
2. The same thing applies with [pistons](http://wiki.vg/Block_Actions#Piston), although the piston logic is more complicated and I'd recommend still leaving the play sound packet there for now.
3. `cPawn` contains logic to play the hard fall Effect, but it was commented out in 1.9 because of the wrong effect ID.  Since the effect ID should be correct now, that probably should be uncommented.

mathiascode:
Maybe I've missed something, but was the effect ID for the falling effect removed in 1.9?

Pokechu22:
> Maybe I've missed something, but was the effect ID for the falling effect removed in 1.9?

Uuuuuh...  I thought it had its ID changed.  Now I've looked and I looked and I can't find it in 1.8 either.  It seems like it only existed in 1.7.  Now I'm extra confused.

Well, it seems like that particle is a separate issue now.

mathiascode:
It doesn't seem like the block packet currently sends any sounds. I guess there would need to be additional changes in the server code for that?

madmaxoft:
Am I right to assume that this will effectively mute the 1.9 and earlier clients? I. e. that the sound names are protocol-specific?

mathiascode:
1.8 clients will lose sounds, but 1.9 and 1.10 will have sounds. The sound names will probably not change in a long time.

src/World.cpp
```diff
@@ -3449,7 +3449,7 @@ bool cWorld::SetTrapdoorOpen(int a_BlockX, int a_BlockY, int a_BlockZ, bool a_Op
 	if (a_Open != IsOpen)
 	{
 		SetBlockMeta(a_BlockX, a_BlockY, a_BlockZ, Meta ^ 0x4);
-		BroadcastSoundParticleEffect(EffectID::SFX_RANDOM_DOOR_OPEN_CLOSE, a_BlockX, a_BlockY, a_BlockZ, 0);
+		BroadcastSoundParticleEffect(EffectID::SFX_RANDOM_WOODEN_TRAPDOOR_OPEN, a_BlockX, a_BlockY, a_BlockZ, 0);
```
mathiascode: It should, and sounds for iron trapdoors and iron doors should be different too, but I don't know enough C++ to make such changes. If someone wants to change it, feel free to improve my [sounds](https://github.com/mathiascode/cuberite/tree/sounds) branch.

madmaxoft:
I'll update this to current master, then squash-merge; we can address the minor concerns later on separately.

madmaxoft:
I attempted to fix the merge conflicts using the GitHub web editor, I wonder how it went :)

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -16759,28 +16759,28 @@ end
...
 				{
-					Notes = "A wither head",
+					Notes = "A flag in the metadata of heads that indicates that the head is a wither head.",
 				},
 				E_META_HEAD_ZOMBIE =
 				{
-					Notes = "A zombie head",
-				},
+					Notes = "A flag in the metadata of heads that indicates that the head is a zombie head.",
+        },
```
madmaxoft: Indent failure

mathiascode:
Ready for merge.

PR #3428, opened 2016-11-09
LuaState: Fixed stack balance when calling functions

### LINKED ISSUE #3425: Error in Lua callback causes leaks in Lua stack
When Cuberite calls into Lua, and the callback raises an error, Cuberite logs the error and then leaks values on the Lua stack. The leaks can accumulate up to a total crash.

Repro:
Add the following code to a console command handler in a plugin:
```lua
local test
return test[1]  -- error here - attempting to index a nil value
```

Execute the console command that has the code in the handler, Cuberite leaks values on the Lua stack and crashes later on (possibly in another Lua callback).

When a stack balance check is added to `cLuaState::Call()`, it fails.

madmaxoft:
Fixes #3425


madmaxoft:
This is now ready for review and squash-merge.

Issue #1586, opened 2014-11-05
Implement a simplex noise


madmaxoft:
The 2D and 3D variants should be implemented as replacements for the original Perlin noise.

madmaxoft:
We now have a Simplex noise implementation, however, it is up to 10 times slower than our current noise. Unless someone can optimize it really *really* well, there's no way we could use that in a generator. See the `Tools/NoiseSpeedTest` project for the implementation and comparison to the current noise implementation, in Cuberite's most often used scenario.

PR #3432, opened 2016-11-14
Added NoiseSpeedTest project.

madmaxoft:
This adds a speed test between our current noise implementation and a Simplex Noise implementation.

Ref.: #1586

Currently the results are speaking strongly in favor of keeping current code, the current noise generator is about 11 times faster for our usage scenario than the Simplex noise implementation:
```
Classic noise took 394 milliseconds
Classic noise took 393 milliseconds
SimplexNoise<float> took 5358 milliseconds
SimplexNoise<float> took 5355 milliseconds
SimplexNoise<double> took 5823 milliseconds
SimplexNoise<double> took 5820 milliseconds
```

### DIFF
--- a/Tools/NoiseSpeedTest/NoiseSpeedTest.cpp
+++ b/Tools/NoiseSpeedTest/NoiseSpeedTest.cpp
@@ -1,0 +1,149 @@
+// NoiseSpeedTest.cpp
+
+// Implements the main app entrypoint
+
+/*
+This program compares the performance of the highly-optimized noise implementation in Cuberite, and the Simplex noise.
+Since the Simplex noise is not yet implemented in Cuberite, an own implementation is provided.
+Also, the performance difference between using a float and double as datatype is measured, by using a templatized Simplex noise.
+
+The testing is done on a usage of the generator that is typical for the Cuberite's terrain generator: generate a 3D array of numbers with
+not much variance in the coords. The exact sizes and coord ranges were adapted from the cNoise3DComposable generator.
+*/
+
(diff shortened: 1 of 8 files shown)


madmaxoft:
Results from a Raspberry Pi (model B (old), 10 times fewer iterations):
```
Classic noise took 975 milliseconds
Classic noise took 974 milliseconds
SimplexNoise<float> took 7256 milliseconds
SimplexNoise<float> took 7257 milliseconds
SimplexNoise<double> took 7788 milliseconds
SimplexNoise<double> took 7595 milliseconds
```
"Only" 7 times faster, but still impressive.

Tools/NoiseSpeedTest/NoiseSpeedTest.cpp
```diff
@@ -0,0 +1,149 @@
...
+	auto timeEnd = std::chrono::high_resolution_clock::now();
+	auto msec = std::chrono::duration_cast<std::chrono::milliseconds>(timeEnd - timeStart);
+	printf("Classic noise took %d milliseconds, returned total %f\n", static_cast<int>(msec.count()), total);
+}
+
+
+
+
+
+static void measureSimplexNoiseFloat(int a_NumIterations)
```
NiLSPACE: I might be too used to C#, but can't `measureSimplexNoiseFloat` and `measureSimplexNoiseDouble` be mixed together using a template function?

madmaxoft: Done in #3435

PR #3438, opened 2016-11-18
Removed ClientHandle.h dependencies from common headers.

madmaxoft:
This results in faster re-compile times when changing the `src/ClientHandle.h` file.

Removed the dependencies from `src/Entities/Player.h` and `src/World.h`


madmaxoft:
Unfortunately this results in `src/Entities/Player.cpp` being a bit chatty with the redirected functions (`SendMessage` et al), but I believe this is a better way.

PR #3440, opened 2016-11-20
cWolf: Fixed targetting a nullptr.

### LINKED ISSUE #3437: Attacking a Wolf Causes the Server to Crash
When running Minecraft 1.10, attacking a wolf (tamed or otherwise) will crash the server. 

Happens when:
     - Melee attacking a wolf.
     - Shooting a wolf with a bow.

Does not happen when:
     - Mobs damage a wolf.
     - Wolves take fall damage.
     - Wolves take damage from environmental hazards like cacti, TNT, lava, or fire (even if the player places   
       them).
     - Players ignite a wolf with flint and steel.

So in short the server only crashes when the player directly does damage to a wolf.

madmaxoft:
Fixes #3437


Seadragon91:
Tested this. Rebased and will merge after success.

PR #3442, opened 2016-11-22
Added API export for LuaCheck

Seadragon91:
This is a step required for [CuberitePluginChecker/#26](https://github.com/cuberite/CuberitePluginChecker/issues/26)

The plugin APIDump will create the file `Server/.luacheckrc`. Here a link how it currently looks [.luacheckrc](https://gist.github.com/Seadragon91/5560cefb67d8cf412934833bea646655)

 The current code is not good and needs to be rewritten. The table `globals` should contain:
* ~Class names that have constructors e.g. `cItem`~
* ~Class names that have static functions / constants~
* Add all class names, much simpler and there are not many
* All global constants / functions from http://api-docs.cuberite.org/Globals.html

Would need a few pointers how I can do that better.

### DIFF
--- a/Server/Plugins/APIDump/main_APIDump.lua
+++ b/Server/Plugins/APIDump/main_APIDump.lua
@@ -1635,6 +1635,75 @@ end
 
 
 
+local function DumpLuaCheck(a_API)
+	LOG("Creating file .luacheckrc...")
+	local file = io.open(".luacheckrc", "w")
+	
+	file:write([[
+-- This file is the config file for the tool named Luacheck
+-- Documentation: http://luacheck.readthedocs.io/en/stable/index.html
+
+-- Ignore unused function and loop arguments
+unused_args = false
+
+-- Allow self defined globals
+allow_defined = true
+
+-- Ignore this functions
+ignore =
+{
+	"Initialize",  -- Plugin
+	"OnDisable",  -- Plugin
+	"RegisterPluginInfoCommands",  -- InfoReg.lua
+	"RegisterPluginInfoConsoleCommands",  -- InfoReg.lua
+	"g_PluginInfo",  -- Info.lua
+}
+
+-- Ignore files / directories
+exclude_files =
+{
+	"tests/" -- CuberitePluginChecker
+}
+
+-- All globals from cuberite (classes, enums, functions)
+globals =
+{
+]])
+	
+	-- Export all global symbols
+	for _, cls in ipairs(a_API) do
+		if cls.Name == "Globals" then
+			-- Global functions
+			for _, func in ipairs(cls.Functions) do
+				file:write("\t\"", func.Name, "\",\n")
+			end
+			
+			-- Global constants
+			for _, const in ipairs(cls.Constants) do
+				file:write("\t\"", const.Name, "\",\n")
+			end
+			
+			-- Global constants from all groups
+			for _, group in pairs(cls.ConstantGroups) do
+				for _, const in pairs(group.Constants) do
+					file:write("\t\"", const.Name, "\",\n")
+				end
+			end
+		else
+			file:write("\t\"", cls.Name, "\",\n")
+		end
+	end
+	
+	file:write("}\n")
+	file:close()
+	
+	LOG("Config file .luacheckrc created...")
+end
+
+
+
+
+
 --- Returns true if a_Descendant is declared to be a (possibly indirect) descendant of a_Base
 local function IsDeclaredDescendant(a_DescendantName, a_BaseName, a_API)
 	-- Check params:
@@ -1803,6 +1872,9 @@ local function DumpApi()
 
 	-- Dump all available API objects in format used by ZeroBraneStudio API descriptions:
 	DumpAPIZBS(API)
+	
+	-- Export the API in a format used by LuaCheck
+	DumpLuaCheck(API)
 
 	LOG("APIDump finished");
 	return true
@@ -2062,3 +2134,4 @@ function Initialize(Plugin)
 
 	return true
 end
+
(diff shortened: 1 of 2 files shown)


Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -617,6 +620,14 @@ local function ReadDescriptions(a_API, a_Desc)
 					local NumInGroup = 0;
 					for _, incl in ipairs(group.Include or {}) do
 						for _, cons in ipairs(cls.Constants) do
+							if cons.Name:match(incl) then
+								if not cons.Name:match("^HOOK_") and not cons.Name:match("^cr") and not cons.Name:match("^ps")
+								and not cons.Name:match("^pk") and not cons.Name:match("^ba") and not cons.Name:match("^ms")
+								and not cons.Name:match("^bs") then
+									table.insert(GlobalsLuacheck, cons.Name)
```
madmaxoft: I don't get it, you exclude a lot of constants and then dump the rest without any namespace, is that how Luacheck needs to have them? I'd have expected Luacheck to either ignore classes' constants altogether, or use proper namespacing for them.

Seadragon91:
Thanks for the hint. I reworked it completely and it looks now much better. I add now all class names, much simpler.

Seadragon91:
Had to correct a few more problems: file write, missing quotation marks and had to make the function local. Now it looks good for me.

Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -1635,6 +1635,73 @@ end
...
+{
+	"tests/" -- CuberitePluginChecker
+}
+
+-- All globals from cuberite (classes, enums, functions)
+globals =
+{
+]]
+
+	file:write(Content)
```
madmaxoft: You could write the text to the file directly, without assigning it to `Contents`.

Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -1635,6 +1635,73 @@ end
...
+unused_args = false
+
+-- Allow self defined globals
+allow_defined = true
+
+-- Ignore this functions
+ignore =
+{
+	"Initialize",  -- Plugin
+	"OnDisable",  -- Plugin
```
madmaxoft: Will the plugins need to specify all their hook and command handlers in their copies of this file?

madmaxoft:
That is much better, thanks.

Server/Plugins/APIDump/main_APIDump.lua
Seadragon91: Done.

Seadragon91:
Now to CuberitePluginChecker. If a plugin doesn't have the file `.luacheckrc`, I would download it from the build server. Could a zip file like `ManualAPI.zip` be added to the build server.

Server/Plugins/APIDump/main_APIDump.lua
Seadragon91: I think because that the hook / command handler is used in the register function, luacheck doesn't report any problems with them.

Seadragon91:
I will edit it adding a link to the documentation and more.

Seadragon91:
Added link and a missing empty line.

Server/Plugins/APIDump/main_APIDump.lua
Seadragon91: A plugin that uses Info.lua has no register function, but the handlers are inside of the Info.lua and luacheck doesn#t show any warnings.

madmaxoft:
Both the comments and the commit message are too centered about what the code is doing, and not WHY it's doing it.
I think a simple "Added API export for LuaCheck" would be a much better commit message, and the comment (line 1876) would be better as `Export the API in a format used by LuaCheck`

Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -1635,6 +1635,75 @@ end
...
+{
+	"tests/" -- CuberitePluginChecker
+}
+
+-- All globals from cuberite (classes, enums, functions)
+globals =
+{
+]])
+	
+	-- Add all class names
```
madmaxoft: Stale comment - the code adds not only class names, but also global constants and functions. How about `Export all global symbols` ?

Seadragon91:
Have done the changes.

madmaxoft:
OK, I have no more comments :)

Seadragon91:
@madmaxoft Should the file `.luacheckrc` be added to the build server?
I can add a check to CuberitePluginChecker that will download the file and add it to the plugin directory, if the plugin doesn't have it. Or should the plugin developer add the file if he upgrade the `circle.yml`.

madmaxoft:
I'm not sure yet. Is there anything in the file that a plugin may want to customize? If you use that file for the most prominent plugins, do they get any (false-positive) warnings? If there is any customization to be done by the plugin, I'd probably include the file as an example directly in the CuberitePluginChecker repo's `tests` folder as an example that the plugin authors should modify to their needs (and update, if needed). If it is completely static in regard to different plugins, then it makes sense to build it on the buildserver regularly and offer it for a download. I don't think CuberitePluginChecker should do that, though - it should be done by each plugin's CI script.

Seadragon91:
Some plugin needs to make customization for example the plugin gallery has a callback named `WorldEditCallback` that is called by WorldEdit and luacheck warns on it. Another one would be shadowing up value.
I think the best is to add the file directly to the plugin. If you want I can add the file `.luachekcrc` to the plugn gallery and modify CuberitePluginChecker to install and run the plugin gallery.

Will merge this now and we can go on with discussion on https://github.com/cuberite/CuberitePluginChecker/issues/26.

Issue #2409, opened 2015-07-30
(Desert) Villages are somethimes partially submerged.


LO1ZB:
(Desert) Villages are somethimes partially submerged, when they cross a river.
![minecraft 1 8 7 30 07 2015 16_30_52](https://cloud.githubusercontent.com/assets/7173513/8986078/baea3b18-36da-11e5-910c-6135a3c02a12.png)
Seed: 1503938543 (Default world.ini)
X: 2345 Z: 1580

NiLSPACE:
This is for all types of villages.

SphinxC0re:
But they shouldn't

NiLSPACE:
Of course, I was just pointing out that it happens with all types of villages. :)

Haxi52:
![2015-09-26_11 54 25](https://cloud.githubusercontent.com/assets/10181722/10118756/c3a57950-6445-11e5-8082-4055deae6844.png)

Some buildings appear suspended above water as well.

madmaxoft:
I'll try to fix this and provide a generic solution for the structure generator

PR #3459, opened 2016-11-24
Prefabs: Implemented support for ExpandFloorStrategy.

madmaxoft:
This fixes village houses "floating" above water, part of #2409


madmaxoft:
Ah, stupid me, I didn't check the new metadata for all the villages and it bugged - all the villages are now set to `None`, rather than `RepeatBottomTillSolid`. This will need a fix :(

NiLSPACE:
With the latest pre-compiled binary I get errors that the `ExpandFloorStrategy` is unknown for the NetherFort prefab.

PR #3470, opened 2016-12-04
Fix meta mirror

### LINKED ISSUE #3367: Prefabs: Blockmeta's are sometimes incorrect when rotating
Client version:   1.9
Commit id:       5625598afac8db3b35d790b8d1228997b228f228

Seed: 784561775
Position: 900 65 -94

madmaxoft:
Fixes #3367

### DIFF
--- a/Server/Plugins/Debuggers/Debuggers.lua
+++ b/Server/Plugins/Debuggers/Debuggers.lua
@@ -987,6 +987,24 @@ end;
 
 
 
+function HandleGenRailsCmd(a_Split, a_Player)
+	local MAX_RAIL_META = 9
+	local pos = a_Player:GetPosition()
+	local ba = cBlockArea:new()
+	ba:Create(2 * MAX_RAIL_META + 3, 4, 3, cBlockArea.baTypes + cBlockArea.baMetas)
+	ba:FillRelCuboid(0, 2 * MAX_RAIL_META + 2, 0, 0, 0, 2, cBlockArea.baTypes, E_BLOCK_STONE)
+	ba:FillRelCuboid(0, 2 * MAX_RAIL_META + 2, 1, 3, 0, 2, cBlockArea.baTypes, E_BLOCK_AIR)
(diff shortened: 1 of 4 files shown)


madmaxoft:
There were two problems with the mirroring code:
 - Rails meta mirroring was wrong, it assumed specialized rails and so didn't work for regular rails
 - BlockArea mirroring didn't mirror the middle slice if the size along the flip axis was an odd number

Both are fixed by this PR, and the Debuggers plugin was extended to provide a testing area for rails upon executing the `/genrails` in-game command.

NiLSPACE:
Wouldn't the same problem occur [here](https://github.com/cuberite/cuberite/blob/ad10303f897bc3380da4dc20af780e3e25aac23b/src/BlockArea.cpp#L1327), [here](https://github.com/cuberite/cuberite/blob/ad10303f897bc3380da4dc20af780e3e25aac23b/src/BlockArea.cpp#L1341), [here](https://github.com/cuberite/cuberite/blob/ad10303f897bc3380da4dc20af780e3e25aac23b/src/BlockArea.cpp#L1362), [here](https://github.com/cuberite/cuberite/blob/ad10303f897bc3380da4dc20af780e3e25aac23b/src/BlockArea.cpp#L1376), [here](https://github.com/cuberite/cuberite/blob/ad10303f897bc3380da4dc20af780e3e25aac23b/src/BlockArea.cpp#L1403) and [here](https://github.com/cuberite/cuberite/blob/ad10303f897bc3380da4dc20af780e3e25aac23b/src/BlockArea.cpp#L1417)?

madmaxoft:
No, and that's the cause of the bug in the first place :)
The point is that if the size is an odd number, the middle "row" of blocks doesn't need to be swapped, but it needs to be meta-rotated. Because the functions started as "no-meta" and then were copied and extended into "with-meta", I missed the point that it needs to meta-mirror the middle "row".

As they say, There are only two hard things in Computer Science: naming things, cache invalidation and *off-by-one errors*. :)

PR #3471, opened 2016-12-06
Fix comparator segfaults

### LINKED ISSUE #3417: Segmentation fault (Place a flower pot near a comparator)
Client version:   1.10
Server OS:        Linux
Commit id:        fdfafed03d2b93ac53b944fcafa625ab7c3e91cc
### Expected behavior

Don't crash
### Actual behavior

Crash:

```
[61cc848f2f3e1667|20:37:32] HandleRightClick: {-44, 95, -6}, face 1, HeldItem: -1:0:0 * 0; dist: 2.18
[61cc848f2f3e1667|20:37:33] HandleRightClick: {-44, 95, -6}, face 1, HeldItem: -1:0:0 * 0; dist: 2.18
[61cc848f2f3e1667|20:37:33] HandleRightClick: {-44, 95, -6}, face 1, HeldItem: -1:0:0 * 0; dist: 2.18
[61cc848f2f3e1667|20:37:34] HandleRightClick: {-44, 94, -5}, face 1, HeldItem: flowerpot:0 * 1; dist: 2.80
[61cc848f2f3e1667|20:37:35] GetSlot: Invalid slot number, 0 out of -524121288 slots

Thread 8 "Cuberite_debug" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffed7f6700 (LWP 21811)]
0x0000000000e9cda1 in cRedstoneComparatorHandler::GetPowerLevel(Vector3<int> const&, unsigned char, unsigned char)::cContainerCallback::Item(cBlockEntity*) (
    this=0x7fffed7f5a50, a_BlockEntity=0x7fffe0c28b00)
    at /home/moritz/Dokumente/cuberite-antihack/cuberite/src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h:62
62                      Fullness += static_cast<float>(Contents.GetSlot(Slot).m_ItemCount) / Contents.GetSlot(Slot).GetMaxStackSize();
(gdb)
```

Place a flower pot near a comparator? But I don't know.

**The debugger is currently running! You can send me commands if you need further information.**

marvinkopf:
The handler would get called for any BlockEntity, but not every BlockEntity is a BlockEntityWithItems. Downcasting with static_cast is UB on fail.

fixes #3417

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
@@ -54,7 +54,14 @@ public:
 
 			virtual bool Item(cBlockEntity * a_BlockEntity) override
 			{
-				auto & Contents = static_cast<cBlockEntityWithItems *>(a_BlockEntity)->GetContents();
+				// Skip BlockEntities that don't have slots
+				auto BlockEntityWithItems = dynamic_cast<cBlockEntityWithItems *>(a_BlockEntity);
+				if (BlockEntityWithItems == nullptr)
+				{
+					return false;
+				}
+
+				auto & Contents = BlockEntityWithItems->GetContents();
 				float Fullness = 0;  // Is a floating-point type to allow later calculation to produce a non-truncated value
 
 				for (int Slot = 0; Slot != Contents.GetNumSlots(); ++Slot)


src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
```diff
@@ -54,7 +54,13 @@ class cRedstoneComparatorHandler : public cRedstoneHandler
 
 			virtual bool Item(cBlockEntity * a_BlockEntity) override
 			{
-				auto & Contents = static_cast<cBlockEntityWithItems *>(a_BlockEntity)->GetContents();
+				auto BlockEntityWithItems = dynamic_cast<cBlockEntityWithItems *>(a_BlockEntity);
+				if (!BlockEntityWithItems)
```
sphinxc0re: Looks good to me but could you explain the falsy/truthyness of `BlockEntityWithItems` to me?

madmaxoft:
I'm not too happy about a `dynamic_cast`, but what can I do :P

src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
madmaxoft: We use explicit comparison to `nullptr` for pointers. `!var` is only ever used for bool variables. Please change this.

src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
```diff
@@ -54,7 +54,13 @@ class cRedstoneComparatorHandler : public cRedstoneHandler
 
 			virtual bool Item(cBlockEntity * a_BlockEntity) override
 			{
-				auto & Contents = static_cast<cBlockEntityWithItems *>(a_BlockEntity)->GetContents();
+				auto BlockEntityWithItems = dynamic_cast<cBlockEntityWithItems *>(a_BlockEntity);
```
madmaxoft: Add a comment above this block of code, explaining why it's here. Something as simple as `// Skip BlockEntitites that have no contents` will work.

src/Simulator/IncrementalRedstoneSimulator/RedstoneComparatorHandler.h
```diff
@@ -54,7 +54,14 @@ class cRedstoneComparatorHandler : public cRedstoneHandler
 
 			virtual bool Item(cBlockEntity * a_BlockEntity) override
 			{
-				auto & Contents = static_cast<cBlockEntityWithItems *>(a_BlockEntity)->GetContents();
+				// Skip BlockEntities that don't have slots
+				auto BlockEntityWithItems = dynamic_cast<cBlockEntityWithItems *>(a_BlockEntity);
+				if (BlockEntityWithItems != nullptr)
```
sphinxc0re: You are returning `false` if this is **not** a `nullptr` ?? Is this right? Before, you were checking for falsyness.

marvinkopf: Ah, obvious mistake. Should be right now.

madmaxoft:
Now it's a great fix, thank you :)

tigerw:
I remember you saying a while back that you didn't want RTTI
Ref: https://github.com/cuberite/cuberite/pull/152

But presumably there isn't too much overhead with it? Or is it better adding a object-type enum to block entities like normal entities do? Or changing everything to dynamic cast and removing said enum from entities? Inconsistency is of course undesirable.

madmaxoft:
I believe an enum is faster than dynamic-casting, but have no idea about how much faster. Without a benchmark, I won't force either way.

Yes, RTTI should be avoided for simple tasks such as deciding the kind of an entity. This is a bit more involved, the question the code is asking is "is this object descended from `cBlockEntityWithItems`?", which would be a bit more involved if we wanted to use an enum - there would need to be a `switch` and a list of BlockEntity types that are indeed descended; each time a new descendant was added to minecraft, we'd need to revisit this switch, so from maintenance point of view, this solution is better.

PR #3472, opened 2016-12-08
Powered rails can kick-start minecarts

### LINKED ISSUE #3017: Powered rails can't kickstart a Minecart properly
You push the button and the cart should normally start - with or without a person riding it.
For some reason, it doesn't, for me. The cart instead seems to be pulled towards the stone block. A few times while testing I was partially clipped inside - and could go right through it if I used the walk buttons while sitting in the cart. It went in the wrong direction.

marvinkopf:
fixes #3017

### DIFF
--- a/src/Entities/Minecart.cpp
+++ b/src/Entities/Minecart.cpp
@@ -448,18 +448,14 @@ void cMinecart::HandlePoweredRailPhysics(NIBBLETYPE a_RailMeta)
 			// If rail is powered check for nearby blocks that could kick-start the minecart
 			else if ((a_RailMeta & 0x8) == 0x8)
 			{
-				BLOCKTYPE BlockSouth = m_World->GetBlock(POSX_TOINT, POSY_TOINT, static_cast<int>(ceil(GetPosZ())));
-				BLOCKTYPE BlockNorth = m_World->GetBlock(POSX_TOINT, POSY_TOINT, POSZ_TOINT - 1);
-				// Only kick-start the minecart if a block is on onee side, but not both
-				if ((!IsBlockRail(BlockNorth) && cBlockInfo::IsSolid(BlockNorth))
-					&&
-					!(!IsBlockRail(BlockSouth) && cBlockInfo::IsSolid(BlockSouth)))
+				bool IsBlockZP = IsSolidBlockAtOffset(0, 0, 1);
+				bool IsBlockZM = IsSolidBlockAtOffset(0, 0, -1);
+				// Only kick-start the minecart if a block is on one side, but not both
+				if (IsBlockZM && !IsBlockZP)
 				{
 					AddSpeedZ(AccelDecelSpeed);
 				}
-				else if (!(!IsBlockRail(BlockNorth) && cBlockInfo::IsSolid(BlockNorth))
-					&&
-					(!IsBlockRail(BlockSouth) && cBlockInfo::IsSolid(BlockSouth)))
+				else if (!IsBlockZM && IsBlockZP)
 				{
 					AddSpeedZ(AccelDecelNegSpeed);
 				}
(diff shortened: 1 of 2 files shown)


src/Entities/Minecart.cpp
```diff
@@ -698,6 +728,20 @@ void cMinecart::SnapToRail(NIBBLETYPE a_RailMeta)
 
 
 
+bool cMinecart::TestBlockCollision(int a_XOffset, int a_YOffset, int a_ZOffset)
+{
+	BLOCKTYPE Block = m_World->GetBlock(POSX_TOINT + a_XOffset, POSY_TOINT + a_YOffset, POSZ_TOINT + a_ZOffset);
```
sphinxc0re: Where are these `POS#_TOINT` variables coming from?

src/Entities/Minecart.cpp
```diff
@@ -777,13 +817,10 @@ bool cMinecart::TestBlockCollision(NIBBLETYPE a_RailMeta)
 		}
 		case E_META_RAIL_CURVED_ZM_XM:
 		{
-			BLOCKTYPE BlockXM = m_World->GetBlock(POSX_TOINT - 1, POSY_TOINT, POSZ_TOINT);
-			BLOCKTYPE BlockZM = m_World->GetBlock(POSX_TOINT, POSY_TOINT, POSZ_TOINT - 1);
+			bool BlockXM = TestBlockCollision(-1, 0, 0);
+			bool BlockZM = TestBlockCollision(0, 0, -1);
```
sphinxc0re: Wouldn't it make sense to rename these since the type has changed?

src/Entities/Minecart.cpp
```diff
@@ -794,13 +831,10 @@ bool cMinecart::TestBlockCollision(NIBBLETYPE a_RailMeta)
 		}
 		case E_META_RAIL_CURVED_ZM_XP:
 		{
-			BLOCKTYPE BlockXP = m_World->GetBlock(POSX_TOINT + 1, POSY_TOINT, POSZ_TOINT);
-			BLOCKTYPE BlockZM = m_World->GetBlock(POSX_TOINT, POSY_TOINT, POSZ_TOINT - 1);
+			bool BlockXP = TestBlockCollision(1, 0, 0);
+			bool BlockZM = TestBlockCollision(0, 0, -1);
```
sphinxc0re: Wouldn't it make sense to rename these since the type has changed?

src/Entities/Minecart.cpp
```diff
@@ -811,13 +845,10 @@ bool cMinecart::TestBlockCollision(NIBBLETYPE a_RailMeta)
 		}
 		case E_META_RAIL_CURVED_ZP_XM:
 		{
-			BLOCKTYPE BlockXM = m_World->GetBlock(POSX_TOINT - 1, POSY_TOINT, POSZ_TOINT);
-			BLOCKTYPE BlockZP = m_World->GetBlock(POSX_TOINT, POSY_TOINT, POSZ_TOINT + 1);
+			bool BlockXM = TestBlockCollision(-1, 0, 0);
+			bool BlockZP = TestBlockCollision(0, 0, +1);
```
sphinxc0re: Wouldn't it make sense to rename these since the type has changed?

src/Entities/Minecart.cpp
```diff
@@ -828,13 +859,10 @@ bool cMinecart::TestBlockCollision(NIBBLETYPE a_RailMeta)
 		}
 		case E_META_RAIL_CURVED_ZP_XP:
 		{
-			BLOCKTYPE BlockXP = m_World->GetBlock(POSX_TOINT + 1, POSY_TOINT, POSZ_TOINT);
-			BLOCKTYPE BlockZP = m_World->GetBlock(POSX_TOINT, POSY_TOINT, POSZ_TOINT + 1);
+			bool BlockXP = TestBlockCollision(1, 0, 0);
+			bool BlockZP = TestBlockCollision(0, 0, 1);
```
sphinxc0re: Wouldn't it make sense to rename these since the type has changed?

src/Entities/Minecart.cpp
```diff
@@ -698,6 +728,20 @@ void cMinecart::SnapToRail(NIBBLETYPE a_RailMeta)
 
 
 
+bool cMinecart::TestBlockCollision(int a_XOffset, int a_YOffset, int a_ZOffset)
```
sphinxc0re: Have you thought about inlining this method?

marvinkopf: For the performance? Usually compilers already do a good job deciding if a function should be inlined.

src/Entities/Minecart.cpp
marvinkopf: The macros are defined in the entity class.

src/Entities/Minecart.cpp
marvinkopf: Sure.

src/Entities/Minecart.cpp
sphinxc0re: Yeah, it should be alright 👍

src/Entities/Minecart.h
```diff
@@ -78,6 +78,8 @@ class cMinecart :
 	void SnapToRail(NIBBLETYPE a_RailMeta);
 	/** Tests if a solid block is in front of a cart, and stops the cart (and returns true) if so; returns false if no obstruction */
 	bool TestBlockCollision(NIBBLETYPE a_RailMeta);
+	/** Tests if a solid block is at a specific offset of the minecart position */
+	bool IsBlockAtOffset(int a_XOffset, int a_YOffset, int a_ZOffset);
```
madmaxoft: Please rename to `IsSolidBlockAtOffset` to make it clear what the function does.

madmaxoft:
I thought I found a bug in the code, until I had a second look, and only then realized that it's not possible to trigger the behavior I thought was wrong. But it lead me to another realization - your code uses `north`, `south` etc., while we prefer to use `ZM`, `ZP` etc. Could you still change that please? I believe it's far easier to refer to the axis directly, than having to visualise which way "north" is in Minecraft axis system.

sphinxc0re:
@madmaxoft why is it called `ZM` again and not `ZN`? That would make much more sense because it would resemble `z-negative`

madmaxoft:
`Z Plus` and `Z Minus`

marvinkopf:
It's alright, I've changed it.

PR #3476, opened 2016-12-10
Export GetPosition and GetSpeed from cEntity as a copy instead of a pointer to lua.

### LINKED ISSUE #2690: cPlayer:GetPosition() returns a pointer to the Vector3d object
I wanted to select 2 positions like WorldEdit. For that I used the function GetPosition() from cPlayer. The positions then I stored in a table to use it later.
Found then out that the function GetPosition returns a pointer to the Vector3d instead of a new Vector3d object :)

I use this workaround from the lua side.
`local position = Vector3d(Player:GetPosition())`
This would then create a new Vector3d object and not point to the current player position.

This should be changed, storing a player position as pointer can crash the server if accessed later and the player is not online :)

Seadragon91:
Closes #2690.
Looks like this are the only two methods that should return a copy.

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -3799,6 +3799,50 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
 
 
 
+static int tolua_cEntity_GetPosition(lua_State * tolua_S)
+{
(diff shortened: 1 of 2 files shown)


Seadragon91:
Looks like circleci "deadlocked".

PR #3478, opened 2016-12-15
Fixed minecart destruction using deallocated memory.

madmaxoft:
The scheduled task was accessing memory that could have already been freed. As a result, the chunk loader was asked to load chunk at coords [0x80000000, 0x80000000].


madmaxoft:
The CI builds fail due to non-working ApiDocs subdomain. Merging despite the CI fail.

Issue #3479, opened 2016-12-15
Support the 1.11 network protocol


madmaxoft:
Add support for the 1.11 network protocol.

Protocol docs: http://wiki.vg/index.php?title=Pre-release_protocol&direction=prev&oldid=8291

This issue is only about the protocol, not about the other features of that release.

  - [x] Protocol plumbing (new class, version number, client recognition)
  - [x] Fix packet IDs for normal gameplay (clients don't D/C)
  - [ ]  Skeletons are invisible
  - [ ]  Wither skeletons, villager zombies, skeleton horses and zombie horses don't spawn at the moment
  - [ ]  Spawn eggs are blank due to changes in vanilla entity names
  - [ ]  Mob spawners don't work due to changes in vanilla entity names
  - [ ]  Chests and enderchests are invisible

Pokechu22:
The hard part will be... entities.  Even if you don't support Llamas, all horse variants got split into their own entities (with their own metadata).  Same with skeletons and zombies.

Also, [Protocol History](http://wiki.vg/Protocol_History#1.11) will help you out, and [Entities](http://wiki.vg/Entities) contains most of the 1.11 changes in a cleaner format than pre-release protocol.

madmaxoft:
Okay, I've done the basics, now someone better than me could do the rest - the entities etc. :)

mathiascode:
Minecraft 1.11.2 has been released with protocol version 316.

madmaxoft:
![oh god why](http://cdn.pophangover.com/wp-content/uploads/2012/07/disgusted-oh-god-why-text.png)

Pokechu22:
They the protocol version number because they added a new item (iron nuggets) and a few new metadata fields.  But there aren't any other changes that could affect the project - obviously, we'll need to make the entity changes at some point, but for now they don't matter.

mathiascode:
These are the issues I've found when testing 1.11:
- Shooting into the ground many times with a bow disconnects the client
- Snowballs and enderpearls go through players and blocks
- Chests and enderchests are invisible (known issue)
- Every mob disconnects the client (known issue)
- Spawn eggs turn into blank spawn eggs

mathiascode:
It seems like any kind of pickup kicks the client.

mathiascode:
What needs to be fixed:
- Snowballs and enderpearls go through players and blocks
- Skeletons are invisible
[...]

PR #3480, opened 2016-12-15
Support for the 1.11 protocol

madmaxoft:
Basic support for the 1.11 protocol. Doesn't support mobs yet (any mob will D/C clients immediately)

Also changed the protocol classes' names for better legibility, and slightly refactored to allow simple overrides for handling incoming packets.

Ref.: #3479

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -2909,7 +2909,7 @@ void cChunk::BroadcastBlockEntity(int a_BlockX, int a_BlockY, int a_BlockZ, cons
 
 
 
-void cChunk::BroadcastCollectEntity(const cEntity & a_Entity, const cPlayer & a_Player, const cClientHandle * a_Exclude)
+void cChunk::BroadcastCollectEntity(const cEntity & a_Entity, const cPlayer & a_Player, int a_Count, const cClientHandle * a_Exclude)
 {
 	for (auto itr = m_LoadedByClient.begin(); itr != m_LoadedByClient.end(); ++itr)
 	{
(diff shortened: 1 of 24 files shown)


madmaxoft:
@NiLSPACE , @sphinxc0re , @Pokechu22 what do you think of the new class names, are they alright?

Pokechu22:
I like the idea of naming them that way (especially for the file names), but I feel like it might be clearer to name the first versions `Protocol_1_9_0` instead of `Protocol_1_9`.  It's not to significant but I think it makes it more obvious that it's intended for the first version.  (Of course, mojang doesn't name them that way, so it can go either way)

sphinxc0re:
I totally agree with @Pokechu22 here. The classnames are much better to read and better to distinguish

NiLSPACE:
Yeah, the names are much better.

madmaxoft:
I'm not too enthusiastic about `1_9_0` having the `_0` suffix, there are iterations that have only a single protocol version, so it does make sense to have the first version without any suffix and additional versions with a suffix.

NiLSPACE:
Having no suffix could also indicate that the class is used for all 1.9 protocols, so I'm inclined to keep it with the suffix.

madmaxoft:
The problem with that is that you don't know whether there will be a new version, until they make the next "major" update, at which point you'll be renaming the class just for the sake of the suffix.

madmaxoft:
Okay, two against one, I'll rename the classes to include the suffix, and then I'm done with this PR :)

NiLSPACE:
Great :)

NiLSPACE:
Now the websites needs to be updated.

sphinxc0re:
@mathiascode to the rescue. Also update the banner

madmaxoft:
I wouldn't advertise 1.11 support just yet. It's very preliminary, I have a feeling that even when a second player joins the game, both players' clients will crash. I'd wait until #3479 is closed.

mathiascode:
Once entities work fine again, I will update the website.

NiLSPACE:
It seems the OnPlayerRightClick hook is called twice in 1.11. Is that a bug in our code or did the protocol actually change to do that?

Issue #3474, opened 2016-12-09
Segfault on stop


Seadragon91:
Client version:  1.10.2
Commit id:        build on e8fb85be885f6c32fdeb861b12fb8402de66ebe5
Changed the cert check to be optional and added g flag for debug informations. I couldn't use the debug build, because I always got an assertion error that the player is nullptr.

### Actual behavior
Server crashes on stop

### Steps to reproduce the behavior
I tested the mojang authentication, installed a web server and added a php script that returns the profile of the user. For the UUID I generated a md5 string from the username. Spammed then the server with pycraft.

I stopped the server, while the server tried to authenticate users.
```
Thread 3 (Thread 0x7f6725088700 (LWP 11302)):
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00000000009a4028 in epoll_dispatch ()
#2  0x00000000009990b1 in event_base_loop ()
#3  0x00000000005f5aee in cNetworkSingleton::RunEventLoop(cNetworkSingleton*) ()
#4  0x00007f67256f7cdf in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f67259c6464 in start_thread (arg=0x7f6725088700) at pthread_create.c:333
#6  0x00007f67251719df in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:105

Thread 2 (Thread 0x7f67264ffb80 (LWP 11301)):
#0  0x00007f67259c76bd in pthread_join (threadid=140080871696128, thread_return=0x0) at pthread_join.c:90
#1  0x00007f67256f7f67 in std::thread::join() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00000000005f4fee in cIsThread::Wait() ()
#3  0x00000000005b7bf3 in cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface, std::default_delete<cSettingsRepositoryInterface> >) ()
#4  0x00000000005e9047 in UniversalMain(std::unique_ptr<cSettingsRepositoryInterface, std::default_delete<cSettingsRepositoryInterface> >) ()
#5  0x00000000005e6c87 in main ()

Thread 1 (Thread 0x7f671e98e700 (LWP 11304)):
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:58
[...]

Issue #2917, opened 2016-01-30
Cacti do not harm mobs when they are on top of them.


BitOfAByte:
If a mob gets on top of a cactus block it is no longer harmed by it.

Bond-009:
This should be fixed.

Issue #902, opened 2014-04-18
Mobs don't spawn with items


pikachu784:
Mobs aren't seen with items when spawned.  Mob functions seem the same, however, the mobs are not SEEN holding or wearing items.

PR #3482, opened 2016-12-16
Fixes for issues #3437, #902, and #2917

Frownigami1:
Issue #2917 fix: Mobs now take damage when standing on top of a cactus block.

Issue #902 fix: Zombie pigmen now spawn holding a golden sword.

Issue #3474 fix: Attacking a wolf will no longer crash the server.

### DIFF
--- a/src/Cuberite.vcxproj.user
+++ b/src/Cuberite.vcxproj.user
@@ -24,4 +24,8 @@
     <LocalDebuggerWorkingDirectory>../Server</LocalDebuggerWorkingDirectory>
     <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <LocalDebuggerWorkingDirectory>../Server</LocalDebuggerWorkingDirectory>
+    <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
+  </PropertyGroup>
 </Project>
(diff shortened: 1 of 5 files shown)


madmaxoft:
Your editor seems misconfigured to modify the indent in the wrong way. Please undo the (unrelated) changes to the indents.

As for git workflow, normally commit the changes as another commit, then rebase-with-squash, and finally force-push the changes.

The commit message is not ideal either - it makes me look up the issues to actually know what is going on. Better would be to briefly mention the problems fixed. For that reason, it's also not good that you combined three fixes into a single commit; they should each have had a separate commit.

madmaxoft:
Are you even sure 3474 is the right issue? It seems unrelated to the text you provided and there doesn't seem to be any change in the code related to that.

src/Cuberite.vcxproj.user
```diff
@@ -24,4 +24,8 @@
     <LocalDebuggerWorkingDirectory>../Server</LocalDebuggerWorkingDirectory>
     <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <LocalDebuggerWorkingDirectory>../Server</LocalDebuggerWorkingDirectory>
+    <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
+  </PropertyGroup>
 </Project>
```
madmaxoft: Actually this file is rather needed - it sets up the initial project settings so that running in the debugger results in running the server in the correct folder.

Though it might be better to generate this file using cmake, next to the real project file (not assuming in-tree build), rather than store it in the git repo and assuming an in-tree build.

mohe2015:
If somebody tries to read the changes the link ignores the indent changes: https://github.com/cuberite/cuberite/pull/3482/files?w=1

sphinxc0re:
@Frownigami1 any updates on this?

sphinxc0re:
@Frownigami1 if you are still working on this, please rebase onto current master, squash your commits into one and above all of that fix the upcoming conflicts

sphinxc0re:
Closing because of no further communication

LogicParrot:
Selfnote: Review and extract the useful bits.

PR #3484, opened 2016-12-17
Added SHA checksums for android Jenkins build; Few improvements

Cl1608Ho:
Enables Jenkins to create SHA checksums for the files, because it uses said compile.sh, not a seperate script

### DIFF
--- a/android/compile.sh
+++ b/android/compile.sh
@@ -70,10 +70,10 @@ case "$1" in
 	;;
 
 	all)
-		echo "Packing server.zip"
+		echo "Packing server.zip ..."
 		mkdir -p Server
-		cd ../Server
-		zip -r ../android/Server/server.zip *
+		cd $BASEDIR/../Server
+		zip -r $BASEDIR/../android/Server/server.zip *
 
 		for arch in armeabi armeabi-v7a arm64-v8a mips mips64 x86 x86_64; do
 			echo "Doing ... $arch ..." && \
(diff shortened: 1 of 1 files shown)


android/compile.sh
```diff
@@ -70,10 +70,10 @@ case "$1" in
 	;;
 
 	all)
-		echo "Packing server.zip"
+		echo "Packing server.zip ..."
 		mkdir -p Server
-		cd ../Server
-		zip -r ../android/Server/server.zip *
+		cd $BASEDIR/../Server
+		zip -r $BASEDIR/../android/Server/server.zip *
```
bearbin: The `../android` looks like it is useless to me, can you confirm its purpose?

bearbin:
Also, other builds on the buildserver use tar for the creation of archive files. Is there a particular reason why this script uses zip or would it be possible to change to tar?

android/compile.sh
Cl1608Ho: Sorry, my mistake.... Dunno why :P

Cl1608Ho:
This is android. And because androud apps are written in Java, we can use Java's built in zip functions. No need for libraries, no need to hassle with that :) besides, what are the real advantages of tar?

bearbin:
CircleCI froze, merging anyway.

tigerw:
> real advantages of tar?

It makes your hard disk platters sticky. So obviously, there are advantages for SSD manufacturers because you are forced to buy new drives that are solid-state.

tigerw:
It's also great for sealing leaky roofs.

PR #3488, opened 2016-12-19
Minor changes

mathiascode:
I've updated a few outdated links, and corrected some mistakes in the API Documentation's canonical tags.


README.md
```diff
@@ -5,7 +5,7 @@ Cuberite is a Minecraft-compatible multiplayer game server that is written in C+
 
 Cuberite can run on Windows, *nix and Android operating systems. This includes Android phones and tablets as well as Raspberry Pis.
 
-We currently support Release 1.8 and 1.9 Minecraft protocol versions.
+We currently support Release 1.8 - 1.10 Minecraft protocol versions.
```
Cl1608Ho: Uhm also 1.11 protocol, right?

Server/README.txt
```diff
@@ -1,16 +1,16 @@
----------------------------------------------------------------------------------
-   Welcome to your new Cuberite server, compatible with Minecraft 1.8 and 1.9!
----------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+   Welcome to your new Cuberite server, compatible with Minecraft 1.8 - 1.10!
```
Cl1608Ho: Here too, right? (1.11)

sphinxc0re:
We don't fully support the 1.11 protocol yet and clients might crash. So we shouldn't endorse using it

madmaxoft:
I've been using the 1.11 for some time now with the Gallery server, and although it seems stable, even for multiple players, it doesn't support entities yet and has some bugs (chests are invisible etc.), so let's not advertise it just yet.

COMPILING.md
```diff
@@ -96,7 +96,7 @@ cmake -DCMAKE_BUILD_TYPE=RELEASE ..
 make -j`nproc`
 ```
 
-This will build Cuberite in release mode, which is better for almost all cases. For more `cmake` options, or for building in debug mode, see the section below.
+This will build Cuberite in release mode, which is better for almost all cases. For more `cmake` options, or for building in debug mode, see [CMake on Unix-based platforms](#cmake-on-unix-based-platforms).
```
bearbin: Perhaps this would be better as:

> which is better in almost all cases

or 

> which is better for almost all uses

bearbin:
Wow, that's a lot of sites that have enabled HTTPS just recently - impressive progress.

PR #3489, opened 2016-12-19
Changed entity ownership model to use smart pointers

tigerw:
Split off from #3115 on sage advice of many others (and the Anti Huge Commits Committee).

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -136,13 +136,12 @@ cChunk::~cChunk()
 	// Remove and destroy all entities that are not players:
 	cEntityList Entities;
 	std::swap(Entities, m_Entities);  // Need another list because cEntity destructors check if they've been removed from chunk
-	for (auto Entity : Entities)
+	for (auto & Entity : Entities)
 	{
 		if (!Entity->IsPlayer())
 		{
 			// Scheduling a normal destruction is neither possible (Since this chunk will be gone till the schedule occurs) nor necessary.
 			Entity->DestroyNoScheduling(false);  // No point in broadcasting in an unloading chunk. Chunks unload when no one is nearby.
-			delete Entity;
 		}
 	}
 
@@ -300,9 +299,9 @@ void cChunk::GetAllData(cChunkDataCallback & a_Callback)
 
 	a_Callback.ChunkData(m_ChunkData);
 
-	for (auto Entity : m_Entities)
+	for (const auto & Entity : m_Entities)
 	{
-		a_Callback.Entity(Entity);
+		a_Callback.Entity(Entity.get());
 	}
 
 	for (auto & KeyPair : m_BlockEntities)
@@ -531,7 +530,7 @@ void cChunk::CollectMobCensus(cMobCensus & toFill)
 	}
 
 	Vector3d currentPosition;
-	for (auto entity : m_Entities)
+	for (auto & entity : m_Entities)
 	{
 		// LOGD("Counting entity #%i (%s)", (*itr)->GetUniqueID(), (*itr)->GetClass());
 		if (entity->IsMob())
@@ -634,7 +633,7 @@ void cChunk::SpawnMobs(cMobSpawner & a_MobSpawner)
 			continue;
 		}
 
-		cEntity * newMob = a_MobSpawner.TryToSpawnHere(this, TryX, TryY, TryZ, Biome, MaxNbOfSuccess);
+		auto newMob = a_MobSpawner.TryToSpawnHere(this, TryX, TryY, TryZ, Biome, MaxNbOfSuccess);
 		if (newMob == nullptr)
 		{
 			continue;
@@ -658,7 +657,7 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 	// If we are not valid, tick players and bailout
 	if (!IsValid())
 	{
-		for (auto Entity : m_Entities)
+		for (const auto & Entity : m_Entities)
 		{
 			if (Entity->IsPlayer())
 			{
@@ -683,7 +682,7 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 		m_IsDirty = KeyPair.second->Tick(a_Dt, *this) | m_IsDirty;
 	}
 
-	for (cEntityList::iterator itr = m_Entities.begin(); itr != m_Entities.end();)
+	for (auto itr = m_Entities.begin(); itr != m_Entities.end();)
 	{
 		// Do not tick mobs that are detached from the world. They're either scheduled for teleportation or for removal.
 		if (!(*itr)->IsTicking())
@@ -708,20 +707,22 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
 			continue;
 		}
 
-		if ((((*itr)->GetChunkX() != m_PosX) ||
-			((*itr)->GetChunkZ() != m_PosZ))
+		if (
+			((*itr)->GetChunkX() != m_PosX) ||
+			((*itr)->GetChunkZ() != m_PosZ)
 		)
 		{
-			// This block is very similar to RemoveEntity, except it uses an iterator to avoid scanning the whole m_Entities
-			// The entity moved out of the chunk, move it to the neighbor
-
-			(*itr)->SetParentChunk(nullptr);
-			MoveEntityToNewChunk(*itr);
 			// Mark as dirty if it was a server-generated entity:
 			if (!(*itr)->IsPlayer())
 			{
 				MarkDirty();
 			}
+
+			// This block is very similar to RemoveEntity, except it uses an iterator to avoid scanning the whole m_Entities
+			// The entity moved out of the chunk, move it to the neighbor
+			(*itr)->SetParentChunk(nullptr);
+			MoveEntityToNewChunk(std::move(*itr));
+
 			itr = m_Entities.erase(itr);
 		}
 		else
@@ -750,7 +751,7 @@ void cChunk::TickBlock(int a_RelX, int a_RelY, int a_RelZ)
 
 
 
-void cChunk::MoveEntityToNewChunk(cEntity * a_Entity)
+void cChunk::MoveEntityToNewChunk(OwnedEntity a_Entity)
 {
 	cChunk * Neighbor = GetNeighborChunk(a_Entity->GetChunkX() * cChunkDef::Width, a_Entity->GetChunkZ() * cChunkDef::Width);
 	if (Neighbor == nullptr)
@@ -764,28 +765,29 @@ void cChunk::MoveEntityToNewChunk(cEntity * a_Entity)
 	}
 
 	ASSERT(Neighbor != this);  // Moving into the same chunk? wtf?
-	Neighbor->AddEntity(a_Entity);
+	auto & Entity = *a_Entity;
+	Neighbor->AddEntity(std::move(a_Entity));
 
 	class cMover :
 		public cClientDiffCallback
 	{
 		virtual void Removed(cClientHandle * a_Client) override
 		{
-			a_Client->SendDestroyEntity(*m_Entity);
+			a_Client->SendDestroyEntity(m_Entity);
 		}
 
 		virtual void Added(cClientHandle * a_Client) override
 		{
-			m_Entity->SpawnOn(*a_Client);
+			m_Entity.SpawnOn(*a_Client);
 		}
 
-		cEntity * m_Entity;
+		cEntity & m_Entity;
 
 	public:
-		cMover(cEntity * a_CallbackEntity) :
+		cMover(cEntity & a_CallbackEntity) :
 			m_Entity(a_CallbackEntity)
 		{}
-	} Mover(a_Entity);
+	} Mover(Entity);
 
 	m_ChunkMap->CompareChunkClients(this, Neighbor, Mover);
 }
@@ -1866,15 +1868,15 @@ void cChunk::CollectPickupsByPlayer(cPlayer & a_Player)
 	double PosY = a_Player.GetPosY();
 	double PosZ = a_Player.GetPosZ();
 
-	for (cEntityList::iterator itr = m_Entities.begin(); itr != m_Entities.end(); ++itr)
+	for (auto & Entity : m_Entities)
 	{
-		if ((!(*itr)->IsPickup()) && (!(*itr)->IsProjectile()))
+		if ((!Entity->IsPickup()) && (!Entity->IsProjectile()))
 		{
 			continue;  // Only pickups and projectiles can be picked up
 		}
-		float DiffX = static_cast<float>((*itr)->GetPosX() - PosX);
-		float DiffY = static_cast<float>((*itr)->GetPosY() - PosY);
-		float DiffZ = static_cast<float>((*itr)->GetPosZ() - PosZ);
+		float DiffX = static_cast<float>(Entity->GetPosX() - PosX);
+		float DiffY = static_cast<float>(Entity->GetPosY() - PosY);
+		float DiffZ = static_cast<float>(Entity->GetPosZ() - PosZ);
 		float SqrDist = DiffX * DiffX + DiffY * DiffY + DiffZ * DiffZ;
 		if (SqrDist < 1.5f * 1.5f)  // 1.5 block
 		{
@@ -1884,13 +1886,13 @@ void cChunk::CollectPickupsByPlayer(cPlayer & a_Player)
 			);
 			*/
 			MarkDirty();
-			if ((*itr)->IsPickup())
+			if (Entity->IsPickup())
 			{
-				(reinterpret_cast<cPickup *>(*itr))->CollectedBy(a_Player);
+				reinterpret_cast<cPickup *>(Entity.get())->CollectedBy(a_Player);
 			}
 			else
 			{
-				(reinterpret_cast<cProjectileEntity *>(*itr))->CollectedBy(a_Player);
+				reinterpret_cast<cProjectileEntity *>(Entity.get())->CollectedBy(a_Player);
 			}
 		}
 		else if (SqrDist < 5 * 5)
@@ -1986,7 +1988,7 @@ void cChunk::RemoveClient(cClientHandle * a_Client)
 
 	if (!a_Client->IsDestroyed())
 	{
-		for (auto Entity : m_Entities)
+		for (auto & Entity : m_Entities)
 		{
 			/*
 			// DEBUG:
@@ -2024,34 +2026,59 @@ bool cChunk::HasAnyClients(void) const
 
 
 
-void cChunk::AddEntity(cEntity * a_Entity)
+void cChunk::AddEntity(OwnedEntity a_Entity)
 {
 	if (!a_Entity->IsPlayer())
 	{
 		MarkDirty();
 	}
 
-	ASSERT(std::find(m_Entities.begin(), m_Entities.end(), a_Entity) == m_Entities.end());  // Not there already
+	auto EntityPtr = a_Entity.get();
 
-	m_Entities.push_back(a_Entity);
-	ASSERT(a_Entity->GetParentChunk() == nullptr);
-	a_Entity->SetParentChunk(this);
+	ASSERT(std::find(m_Entities.begin(), m_Entities.end(), a_Entity) == m_Entities.end());  // Not there already
+	m_Entities.emplace_back(std::move(a_Entity));
+
+	ASSERT(EntityPtr->GetParentChunk() == nullptr);
+	EntityPtr->SetParentChunk(this);
 }
 
 
 
 
 
-void cChunk::RemoveEntity(cEntity * a_Entity)
+OwnedEntity cChunk::RemoveEntity(cEntity & a_Entity)
 {
-	ASSERT(a_Entity->GetParentChunk() == this);
-	a_Entity->SetParentChunk(nullptr);
-	m_Entities.remove(a_Entity);
+	ASSERT(a_Entity.GetParentChunk() == this);
+	ASSERT(!a_Entity.IsTicking());
+	a_Entity.SetParentChunk(nullptr);
+
 	// Mark as dirty if it was a server-generated entity:
-	if (!a_Entity->IsPlayer())
+	if (!a_Entity.IsPlayer())
 	{
 		MarkDirty();
 	}
+
+	OwnedEntity Removed;
+	m_Entities.erase(
+		std::remove_if(
+			m_Entities.begin(),
+			m_Entities.end(),
+			[&a_Entity, &Removed](decltype(m_Entities)::value_type & a_Value)
+			{
+				if (a_Value.get() == &a_Entity)
+				{
+					ASSERT(!Removed);
+					Removed = std::move(a_Value);
+					return true;
+				}
+
+				return false;
+			}
+		),
+		m_Entities.end()
+	);
+
+	return Removed;
 }
 
 
@@ -2060,13 +2087,13 @@ void cChunk::RemoveEntity(cEntity * a_Entity)
 
 bool cChunk::HasEntity(UInt32 a_EntityID)
 {
-	for (cEntityList::const_iterator itr = m_Entities.begin(), end = m_Entities.end(); itr != end; ++itr)
+	for (const auto & Entity : m_Entities)
 	{
-		if ((*itr)->GetUniqueID() == a_EntityID)
+		if (Entity->GetUniqueID() == a_EntityID)
 		{
 			return true;
 		}
-	}  // for itr - m_Entities[]
+	}
 	return false;
 }
 
@@ -2077,14 +2104,14 @@ bool cChunk::HasEntity(UInt32 a_EntityID)
 bool cChunk::ForEachEntity(cEntityCallback & a_Callback)
 {
 	// The entity list is locked by the parent chunkmap's CS
-	for (cEntityList::iterator itr = m_Entities.begin(), itr2 = itr; itr != m_Entities.end(); itr = itr2)
+	for (auto itr = m_Entities.begin(), itr2 = itr; itr != m_Entities.end(); itr = itr2)
 	{
 		++itr2;
 		if (!(*itr)->IsTicking())
 		{
 			continue;
 		}
-		if (a_Callback.Item(*itr))
+		if (a_Callback.Item(itr->get()))
 		{
 			return false;
 		}
@@ -2099,7 +2126,7 @@ bool cChunk::ForEachEntity(cEntityCallback & a_Callback)
 bool cChunk::ForEachEntityInBox(const cBoundingBox & a_Box, cEntityCallback & a_Callback)
 {
 	// The entity list is locked by the parent chunkmap's CS
-	for (cEntityList::iterator itr = m_Entities.begin(), itr2 = itr; itr != m_Entities.end(); itr = itr2)
+	for (auto itr = m_Entities.begin(), itr2 = itr; itr != m_Entities.end(); itr = itr2)
 	{
 		++itr2;
 		if (!(*itr)->IsTicking())
@@ -2112,7 +2139,7 @@ bool cChunk::ForEachEntityInBox(const cBoundingBox & a_Box, cEntityCallback & a_
 			// The entity is not in the specified box
 			continue;
 		}
-		if (a_Callback.Item(*itr))
+		if (a_Callback.Item(itr->get()))
 		{
 			return false;
 		}
@@ -2136,11 +2163,11 @@ bool cChunk::DoWithEntityByID(UInt32 a_EntityID, cEntityCallback & a_Callback, b
 bool cChunk::DoWithEntityByID(UInt32 a_EntityID, cLambdaEntityCallback a_Callback, bool & a_CallbackResult)
 {
 	// The entity list is locked by the parent chunkmap's CS
-	for (cEntityList::iterator itr = m_Entities.begin(), end = m_Entities.end(); itr != end; ++itr)
+	for (const auto & Entity : m_Entities)
 	{
-		if (((*itr)->GetUniqueID() == a_EntityID) && ((*itr)->IsTicking()))
+		if ((Entity->GetUniqueID() == a_EntityID) && (Entity->IsTicking()))
 		{
-			a_CallbackResult = a_Callback(*itr);
+			a_CallbackResult = a_Callback(Entity.get());
 			return true;
 		}
 	}  // for itr - m_Entitites[]
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -28,7 +28,8 @@ class cEntity;
 class cClientHandle;
 class cBlockEntity;
 
-typedef std::list<cEntity *>          cEntityList;
+typedef std::unique_ptr<cEntity> OwnedEntity;
+typedef std::vector<OwnedEntity> cEntityList;
 typedef std::map<int, cBlockEntity *> cBlockEntities;
 
 
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1497,38 +1497,38 @@ void cChunkMap::RemoveClientFromChunks(cClientHandle * a_Client)
 
 
 
-void cChunkMap::AddEntity(cEntity * a_Entity)
+void cChunkMap::AddEntity(OwnedEntity a_Entity)
 {
 	cCSLock Lock(m_CSChunks);
 	cChunkPtr Chunk = GetChunk(a_Entity->GetChunkX(), a_Entity->GetChunkZ());
 	if (Chunk == nullptr)  // This will assert inside GetChunk in Debug builds
 	{
 		LOGWARNING("Entity at %p (%s, ID %d) spawning in a non-existent chunk, the entity is lost.",
-			static_cast<void *>(a_Entity), a_Entity->GetClass(), a_Entity->GetUniqueID()
+			static_cast<void *>(a_Entity.get()), a_Entity->GetClass(), a_Entity->GetUniqueID()
 		);
 		return;
 	}
-	Chunk->AddEntity(a_Entity);
+	Chunk->AddEntity(std::move(a_Entity));
 }
 
 
 
 
 
-void cChunkMap::AddEntityIfNotPresent(cEntity * a_Entity)
+void cChunkMap::AddEntityIfNotPresent(OwnedEntity a_Entity)
 {
 	cCSLock Lock(m_CSChunks);
 	cChunkPtr Chunk = GetChunk(a_Entity->GetChunkX(), a_Entity->GetChunkZ());
 	if (Chunk == nullptr)  // This will assert inside GetChunk in Debug builds
 	{
 		LOGWARNING("Entity at %p (%s, ID %d) spawning in a non-existent chunk, the entity is lost.",
-			static_cast<void *>(a_Entity), a_Entity->GetClass(), a_Entity->GetUniqueID()
+			static_cast<void *>(a_Entity.get()), a_Entity->GetClass(), a_Entity->GetUniqueID()
 		);
 		return;
 	}
 	if (!Chunk->HasEntity(a_Entity->GetUniqueID()))
 	{
-		Chunk->AddEntity(a_Entity);
+		Chunk->AddEntity(std::move(a_Entity));
 	}
 }
 
@@ -1553,17 +1553,18 @@ bool cChunkMap::HasEntity(UInt32 a_UniqueID)
 
 
 
-void cChunkMap::RemoveEntity(cEntity * a_Entity)
+OwnedEntity cChunkMap::RemoveEntity(cEntity & a_Entity)
 {
 	cCSLock Lock(m_CSChunks);
-	cChunkPtr Chunk = a_Entity->GetParentChunk();
+	cChunkPtr Chunk = a_Entity.GetParentChunk();
 
-	// Even if a chunk is not valid, it may still contain entities such as players; make sure to remove them (#1190)
 	if (Chunk == nullptr)
 	{
-		return;
+		return nullptr;
 	}
-	Chunk->RemoveEntity(a_Entity);
+
+	// Remove the entity no matter whether the chunk itself is valid or not (#1190)
+	return Chunk->RemoveEntity(a_Entity);
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -133,7 +133,7 @@ cClientHandle::~cClientHandle()
 			}
 			m_Player->DestroyNoScheduling(true);
 		}
-		delete m_Player;
+		m_PlayerPtr.reset();
 		m_Player = nullptr;
 	}
 
@@ -157,8 +157,12 @@ void cClientHandle::Destroy(void)
 		cCSLock Lock(m_CSOutgoingData);
 		m_Link.reset();
 	}
+
+	// Temporary (#3115-will-fix): variable to keep track of whether the client authenticated and had the opportunity to have ownership transferred to the world
+	bool WasAddedToWorld = false;
 	{
 		cCSLock Lock(m_CSState);
+		WasAddedToWorld = (m_State >= csAuthenticated);
 		if (m_State >= csDestroying)
 		{
 			// Already called
@@ -186,7 +190,23 @@ void cClientHandle::Destroy(void)
 		{
 			player->StopEveryoneFromTargetingMe();
 			player->SetIsTicking(false);
-			world->RemovePlayer(player, true);
+
+			if (WasAddedToWorld)
+			{
+				// If ownership was transferred, our own smart pointer should be unset
+				ASSERT(!m_PlayerPtr);
+
+				m_PlayerPtr = world->RemovePlayer(*player, true);
+
+				// And RemovePlayer should have returned a valid smart pointer
+				ASSERT(m_PlayerPtr);
+			}
+			else
+			{
+				// If ownership was not transferred, our own smart pointer should be valid and RemovePlayer's should not
+				ASSERT(m_PlayerPtr);
+				ASSERT(!world->IsPlayerReferencedInWorldOrChunk(*player));
+			}
 		}
 		player->RemoveClientHandle();
 	}
@@ -359,7 +379,8 @@ void cClientHandle::Authenticate(const AString & a_Name, const AString & a_UUID,
 		m_Protocol->SendLoginSuccess();
 
 		// Spawn player (only serversided, so data is loaded)
-		m_Player = new cPlayer(m_Self, GetUsername());
+		m_PlayerPtr = cpp14::make_unique<cPlayer>(m_Self, GetUsername());
+		m_Player = m_PlayerPtr.get();
 		/*
 		LOGD("Created a new cPlayer object at %p for client %s @ %s (%p)",
 			static_cast<void *>(m_Player),
@@ -2203,7 +2224,7 @@ void cClientHandle::ServerTick(float a_Dt)
 
 			// Add the player to the world (start ticking from there):
 			m_State = csDownloadingWorld;
-			m_Player->Initialize(*(m_Player->GetWorld()));
+			m_Player->Initialize(std::move(m_PlayerPtr), *(m_Player->GetWorld()));
 			return;
 		}
 	}  // lock(m_CSState)
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -135,9 +135,9 @@ const char * cEntity::GetParentClass(void) const
 
 
 
-bool cEntity::Initialize(cWorld & a_World)
+bool cEntity::Initialize(OwnedEntity a_Self, cWorld & a_EntityWorld)
 {
-	if (cPluginManager::Get()->CallHookSpawningEntity(a_World, *this))
+	if (cPluginManager::Get()->CallHookSpawningEntity(a_EntityWorld, *this))
 	{
 		return false;
 	}
@@ -151,13 +151,13 @@ bool cEntity::Initialize(cWorld & a_World)
 
 	ASSERT(m_World == nullptr);
 	ASSERT(GetParentChunk() == nullptr);
-	a_World.AddEntity(this);
+	a_EntityWorld.AddEntity(std::move(a_Self));
 	ASSERT(m_World != nullptr);
 
-	cPluginManager::Get()->CallHookSpawnedEntity(a_World, *this);
+	cPluginManager::Get()->CallHookSpawnedEntity(a_EntityWorld, *this);
 
 	// Spawn the entity on the clients:
-	a_World.BroadcastSpawnEntity(*this);
+	a_EntityWorld.BroadcastSpawnEntity(*this);
 
 	return true;
 }
@@ -230,8 +230,10 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
 			this->GetUniqueID(), this->GetClass(),
 			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
 		);
-		ParentChunk->RemoveEntity(this);
-		delete this;
+
+		// Make sure that RemoveEntity returned a valid smart pointer
+		// Also, not storing the returned pointer means automatic destruction
+		VERIFY(ParentChunk->RemoveEntity(*this));
 	});
 	Destroyed();
 }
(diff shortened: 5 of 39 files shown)


madmaxoft:
I'd like to review this more carefully, please give me more time :)

src/ClientHandle.cpp
```diff
@@ -184,7 +188,23 @@ void cClientHandle::Destroy(void)
...
+
+				// And RemovePlayer should have returned a valid smart pointer
+				ASSERT(m_PlayerPtr);
+			}
+			else
+			{
+				// If ownership was not transferred, our own smart pointer should be valid and RemovePlayer's should not
+				ASSERT(m_PlayerPtr);
+				ASSERT(!world->RemovePlayer(*player, true));
+			}
```
madmaxoft: This is ugly and I really really hope you have a better solution stashed for not-too-late-later.

src/Entities/Entity.cpp
```diff
@@ -228,8 +228,10 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
 			this->GetUniqueID(), this->GetClass(),
 			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
 		);
-		ParentChunk->RemoveEntity(this);
-		delete this;
+
+		// Make sure that RemoveEntity returned a valid smart pointer
+		// Also, not storing the returned pointer means automatic destruction
+		ASSERT(ParentChunk->RemoveEntity(*this));
```
madmaxoft: Wrong, this gets compiled out in Release mode. Use `VERIFY` instead.

src/Entities/Entity.cpp
```diff
@@ -133,9 +133,9 @@ const char * cEntity::GetParentClass(void) const
 
 
 
-bool cEntity::Initialize(cWorld & a_World)
+bool cEntity::Initialize(std::unique_ptr<cEntity> a_Entity, cWorld & a_EntityWorld)
```
madmaxoft: Rename `a_Entity` to `a_Self` to make it more obvious what the param should be. Also, ugly!

src/Entities/Entity.h
```diff
@@ -153,7 +153,7 @@ class cEntity
 
 	/** Spawns the entity in the world; returns true if spawned, false if not (plugin disallowed).
 	Adds the entity to the world. */
-	virtual bool Initialize(cWorld & a_World);
+	virtual bool Initialize(std::unique_ptr<cEntity>a_Entity, cWorld & a_EntityWorld);
```
madmaxoft: `a_Self`

src/Entities/Player.cpp
```diff
@@ -1956,7 +1956,9 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 		GetWorld()->BroadcastDestroyEntity(*this);
 
 		// Remove player from world
-		GetWorld()->RemovePlayer(this, false);
+		// Make sure that RemovePlayer didn't return a valid smart pointer, due to the second parameter being false
+		// We remain valid and not destructed after this call
+		ASSERT(!GetWorld()->RemovePlayer(*this, false));
```
madmaxoft: `VERIFY`

src/Entities/Player.cpp
```diff
@@ -1998,8 +2000,7 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 			a_OldWorld.GetName().c_str(), a_World->GetName().c_str(),
 			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
 		);
-		ParentChunk->RemoveEntity(this);
-		a_World->AddPlayer(this, &a_OldWorld);  // New world will take over and announce client at its next tick
+		a_World->AddPlayer(std::unique_ptr<cPlayer>(static_cast<cPlayer *>(ParentChunk->RemoveEntity(*this).release())), &a_OldWorld);  // New world will take over and announce client at its next tick
```
madmaxoft: Ugh, this is ugly. Why can't this be done directly on the ptr that RemoveEntity returns?

src/Items/ItemBow.h
```diff
@@ -69,17 +69,12 @@ class cItemBowHandler :
 		}
 
 		// Create the arrow entity:
-		cArrowEntity * Arrow = new cArrowEntity(*a_Player, Force * 2);
-		if (Arrow == nullptr)
+		auto ArrowPtr = cpp14::make_unique<cArrowEntity>(*a_Player, Force * 2);
+		auto Arrow = ArrowPtr.get();
+		if (!Arrow->Initialize(std::move(ArrowPtr), *a_Player->GetWorld()))
```
madmaxoft: Man, this code has some danger potential. I was just typing "`ArrowPtr->Initialize(...)` should work", but then I realized, what would get called first, the `std::move` or the indirection?

We need to come up with a better interface for initialization (after this PR is merged).

src/Items/ItemFishingRod.h
```diff
@@ -244,9 +244,9 @@ class cItemFishingRodHandler :
 		}
 		else
 		{
-			cFloater * Floater = new cFloater(a_Player->GetPosX(), a_Player->GetStance(), a_Player->GetPosZ(), a_Player->GetLookVector() * 15, a_Player->GetUniqueID(), static_cast<int>(100 + static_cast<unsigned int>(a_World->GetTickRandomNumber(800)) - (a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLure) * 100)));
-			Floater->Initialize(*a_World);
+			auto Floater = cpp14::make_unique<cFloater>(a_Player->GetPosX(), a_Player->GetStance(), a_Player->GetPosZ(), a_Player->GetLookVector() * 15, a_Player->GetUniqueID(), static_cast<int>(100 + static_cast<unsigned int>(a_World->GetTickRandomNumber(800)) - (a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLure) * 100)));
 			a_Player->SetIsFishing(true, Floater->GetUniqueID());
+			Floater->Initialize(std::move(Floater), *a_World);
```
madmaxoft: This order reversal is dangerous, what if `SetIsFishing` was rewritten to actually do something with the floater's ID, searching the world for it for example? I think the initialization should come first.

src/Items/ItemPainting.h
```diff
@@ -70,8 +70,8 @@ class cItemPaintingHandler :
 				{ "BurningSkull" }
 			};
 
-			cPainting * Painting = new cPainting(gPaintingTitlesList[a_World->GetTickRandomNumber(ARRAYCOUNT(gPaintingTitlesList) - 1)].Title, a_BlockFace, a_BlockX, a_BlockY, a_BlockZ);
-			Painting->Initialize(*a_World);
+			auto Painting = cpp14::make_unique<cPainting>(gPaintingTitlesList[a_World->GetTickRandomNumber(ARRAYCOUNT(gPaintingTitlesList) - 1)].Title, a_BlockFace, a_BlockX, a_BlockY, a_BlockZ);
+			Painting->Initialize(std::move(Painting), *a_World);
```
madmaxoft: Danger! What happens first, the `std::move` or the `->` ? If the move happens first, then the dereference happens on an empty pointer, BOOM.

src/MobSpawner.cpp
```diff
@@ -373,14 +372,15 @@ cMonster * cMobSpawner::TryToSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY,
...
 	{
-		cMonster * newMob = cMonster::NewMonsterFromType(m_MobType);
+		auto newMob = cMonster::NewMonsterFromType(m_MobType);
 		if (newMob)
 		{
-			m_Spawned.insert(newMob);
+			m_Spawned.insert(std::move(newMob));
 		}
-		toReturn = newMob;
+		return newMob;
```
madmaxoft: Wait, what? You add the new mob to the internal list and then return the (now empty) variable? Double-check this. At the very least, the doxy-comment for this function needs an update.

src/Mobs/Blaze.cpp
```diff
@@ -39,17 +39,11 @@ bool cBlaze::Attack(std::chrono::milliseconds a_Dt)
 		// Setting this higher gives us more wiggle room for attackrate
 		Vector3d Speed = GetLookVector() * 20;
 		Speed.y = Speed.y + 1;
-		cFireChargeEntity * FireCharge = new cFireChargeEntity(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
-		if (FireCharge == nullptr)
+		auto FireCharge = cpp14::make_unique<cFireChargeEntity>(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
+		if (!FireCharge->Initialize(std::move(FireCharge), *m_World))
```
madmaxoft: Danger, unknown order of evaluation

src/Mobs/Ghast.cpp
```diff
@@ -39,17 +39,11 @@ bool cGhast::Attack(std::chrono::milliseconds a_Dt)
 		// Setting this higher gives us more wiggle room for attackrate
 		Vector3d Speed = GetLookVector() * 20;
 		Speed.y = Speed.y + 1;
-		cGhastFireballEntity * GhastBall = new cGhastFireballEntity(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
-		if (GhastBall == nullptr)
+		auto GhastBall = cpp14::make_unique<cGhastFireballEntity>(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
+		if (!GhastBall->Initialize(std::move(GhastBall), *m_World))
```
madmaxoft: Danger, unknown order of evaluation

src/Mobs/Skeleton.cpp
```diff
@@ -57,17 +57,11 @@ bool cSkeleton::Attack(std::chrono::milliseconds a_Dt)
 		Vector3d Inaccuracy = Vector3d(Random.NextFloat(0.5) - 0.25, Random.NextFloat(0.5) - 0.25, Random.NextFloat(0.5) - 0.25);
 		Vector3d Speed = (GetTarget()->GetPosition() + Inaccuracy - GetPosition()) * 5;
 		Speed.y = Speed.y - 1 + Random.NextInt(3);
-		cArrowEntity * Arrow = new cArrowEntity(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
-		if (Arrow == nullptr)
+		auto Arrow = cpp14::make_unique<cArrowEntity>(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
+		if (!Arrow->Initialize(std::move(Arrow), *m_World))
```
madmaxoft: Danger, unknown order of evaluation

src/Simulator/SandSimulator.cpp
```diff
@@ -61,8 +61,8 @@ void cSandSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
 				Pos.x, Pos.y, Pos.z, ItemTypeToString(BlockType).c_str(), ItemTypeToString(BlockBelow).c_str()
 			);
 			*/
-			cFallingBlock * FallingBlock = new cFallingBlock(Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
-			FallingBlock->Initialize(m_World);
+			auto FallingBlock = cpp14::make_unique<cFallingBlock>(Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
+			FallingBlock->Initialize(std::move(FallingBlock), m_World);
```
madmaxoft: Danger, unknown order of evaluation

src/World.h
```diff
@@ -792,7 +792,7 @@ class cWorld :
 	/** Spawns a mob of the specified type. Returns the mob's UniqueID if recognized and spawned, cEntity::INVALID_ID otherwise */
 	virtual UInt32 SpawnMob(double a_PosX, double a_PosY, double a_PosZ, eMonsterType a_MonsterType, bool a_Baby = false) override;  // tolua_export
 
-	UInt32 SpawnMobFinalize(cMonster * a_Monster);
+	UInt32 SpawnMobFinalize(std::unique_ptr<cMonster> a_Monster);
```
madmaxoft: Add documentation for this function - what does it do with the ownership, does it remember the monster somewhere or does it destroy it?

src/Chunk.cpp
```diff
@@ -1950,34 +1955,58 @@ bool cChunk::HasAnyClients(void) const
...
 
-void cChunk::AddEntity(cEntity * a_Entity)
+void cChunk::AddEntity(std::unique_ptr<cEntity> a_Entity)
 {
 	if (!a_Entity->IsPlayer())
 	{
 		MarkDirty();
 	}
 
+	auto EntityPtr = a_Entity.get();
```
tigerw: But lines 1970 and 1971 are after 1968?

src/ClientHandle.cpp
tigerw: > This is ugly 

a large part of clienthandle, world, and player. :P

src/Entities/Entity.cpp
tigerw: oops

src/Entities/Player.cpp
tigerw: Because someone broke inheritance by making players specially need AddPlayer rather than AddEntity, and I couldn't get downcasting working with unique pointers.

src/Items/ItemBow.h
tigerw: > but then I realized, what would get called first, the std::move or the indirection?

\*something about worktycho and standards\*

> We need to come up with a better interface for initialization (after this PR is merged).

Indeed. I would suggest that Initialize does not try to call AddEntity; whomever created the entity should. Also, merge Initialize into the constructor.

src/Items/ItemPainting.h
tigerw: > BOOM

nu-uh. The server actually goes picking strawberries.

src/MobSpawner.cpp
tigerw: oops

src/Items/ItemBow.h
tigerw: > but then I realized, what would get called first, the std::move or the indirection?

OK. I'm going to try to do a Tycho and quote from the standard here.

-----------
From **N4140**:

§ 5.2.2.8
> [Note: The evaluations of the postﬁx expression and of the arguments are all unsequenced relative to one another. All side eﬀects of argument evaluations are sequenced before the function is entered (see 1.9). —end note]

where by postfix expression they mean:

§ 5.2.2.1
> A function call is a postﬁx expression followed by parentheses containing a possibly empty, comma-separated list of initializer-clauses which constitute the arguments to the function. The postﬁx expression shall have function type or pointer to function type.

I'm interpreting that as meaning `->` and `move` do not have a specified order. Initialize really shouldn't be taking owning pointers to itself.

madmaxoft: I think that's only half of the relevant part. The other half is determining the function to call in the first place. Because the function is virtual, it needs to be decided in runtime. The compiler has to do these things (not necessarily in this order):
 - Dereference the `Arrow` pointer and retrieve its virtual function table, from there determine the address of the function to call
 - Evaluate and push all parameters onto the stack (including the hidden `this` parameter)
 - Jump to the function's address (simplified-speaking)

So the problems this code has are:
 1. Does dereferencing `Arrow` happen before evaluating the `std::move`?
 2. Does pushing the `this` happen before the `std::move`?

If I remember correctly, at least (2) is explicitly undefined, so the compiler is free to choose.

For now, fix the code by adding a temporary variable holding the naked pointer; we'll fix it later by refactoring the `Initialize` function away.

madmaxoft:
Heh, and you wanted to merge this as a small part of an even larger PR :P

tigerw:
now you're just rubbing it in

src/World.h
tigerw: It seems to be a wrapper around cEntity::Initialize which hopefully will be refactorzd-awehy, eventually.

madmaxoft: Still, it needs a proper documentation. Explain it like I'm five years old :)

LogicParrot:
>Anti Huge Commits Committee

tigerw:
well, it was founded by you :P

tigerw:
xoft, review again?

LogicParrot:
Entity pointers used to break in funny subtle ways. So I advise thorough in-game testing before merging, including, but not limited to, wolves, projectiles, and world travel.

tigerw:
mhm. Talk about brittle code.

madmaxoft:
I'll review this later, hopefully tomorrow. It's already too late and I'm too tired for work this serious :P

src/ChunkDef.h
```diff
@@ -31,8 +31,9 @@ class cEntity;
 class cClientHandle;
 class cBlockEntity;
 
-typedef std::list<cEntity *>        cEntityList;
-typedef std::list<cBlockEntity *>   cBlockEntityList;
+typedef std::unique_ptr<cEntity> StoredEntity;
```
madmaxoft: I know it's nitpicky, but how about `OwnedEntity` instead? It makes more sense in my opinion :)

src/ChunkMap.cpp
```diff
@@ -1595,17 +1595,18 @@ bool cChunkMap::HasEntity(UInt32 a_UniqueID)
...
 
-	// Even if a chunk is not valid, it may still contain entities such as players; make sure to remove them (#1190)
 	if (Chunk == nullptr)
 	{
-		return;
+		return nullptr;
 	}
-	Chunk->RemoveEntity(a_Entity);
+
+	// Remove the entity as long as the chunk exists, regardless of validity (#1190)
```
madmaxoft: Regardless of *whose* validity? I think my previous was still better:
`Remove the entity no matter whether the chunk itself is valid or not`

src/Entities/ProjectileEntity.cpp
```diff
@@ -270,15 +270,15 @@ cProjectileEntity * cProjectileEntity::Create(eKind a_Kind, cEntity * a_Creator,
...
-		case pkWitherSkull:   return new cWitherSkullEntity     (a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkArrow:         return cpp14::make_unique<cArrowEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkEgg:           return cpp14::make_unique<cThrownEggEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkEnderPearl:    return cpp14::make_unique<cThrownEnderPearlEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkSnowball:      return cpp14::make_unique<cThrownSnowballEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkGhastFireball: return cpp14::make_unique<cGhastFireballEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkFireCharge:    return cpp14::make_unique<cFireChargeEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkExpBottle:     return cpp14::make_unique<cExpBottleEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
+		case pkSplashPotion:  return cpp14::make_unique<cSplashPotionEntity>(a_Creator, a_X, a_Y, a_Z, Speed, *a_Item);
+		case pkWitherSkull:   return cpp14::make_unique<cWitherSkullEntity>(a_Creator, a_X, a_Y, a_Z, Speed);
```
madmaxoft: I'd still like to have the vertical alignment for the function params, since they are all the same. Could you keep the spaces between `>` and `(a_Creator...`?

src/Items/ItemItemFrame.h
```diff
@@ -38,11 +38,9 @@ class cItemItemFrameHandler :
 
 		if (Block == E_BLOCK_AIR)
 		{
-			cItemFrame * ItemFrame = new cItemFrame(a_BlockFace, a_BlockX, a_BlockY, a_BlockZ);
-			if (!ItemFrame->Initialize(*a_World))
+			auto ItemFrame = cpp14::make_unique<cItemFrame>(a_BlockFace, a_BlockX, a_BlockY, a_BlockZ);
+			if (!ItemFrame->Initialize(std::move(ItemFrame), *a_World))
```
madmaxoft: Danger, unknown evaluation order.

src/Mobs/Ghast.cpp
```diff
@@ -39,17 +39,15 @@ bool cGhast::Attack(std::chrono::milliseconds a_Dt)
...
-		cGhastFireballEntity * GhastBall = new cGhastFireballEntity(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
-		if (GhastBall == nullptr)
-		{
-			return false;
-		}
-		if (!GhastBall->Initialize(*m_World))
+
+		auto GhastBall = cpp14::make_unique<cGhastFireballEntity>(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
+		auto GhastBallPtr = GhastBall.get();
+
```
madmaxoft: Get rid of this empty line, the entire block is the ghastball init.

src/Mobs/Skeleton.cpp
```diff
@@ -57,19 +57,16 @@ bool cSkeleton::Attack(std::chrono::milliseconds a_Dt)
...
-		cArrowEntity * Arrow = new cArrowEntity(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
-		if (Arrow == nullptr)
-		{
-			return false;
-		}
-		if (!Arrow->Initialize(*m_World))
+
+		auto Arrow = cpp14::make_unique<cArrowEntity>(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
+		auto ArrowPtr = Arrow.get();
+
```
madmaxoft: Get rid of the empty line

src/Mobs/Blaze.cpp
```diff
@@ -39,19 +39,18 @@ bool cBlaze::Attack(std::chrono::milliseconds a_Dt)
...
-		cFireChargeEntity * FireCharge = new cFireChargeEntity(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
-		if (FireCharge == nullptr)
-		{
-			return false;
-		}
-		if (!FireCharge->Initialize(*m_World))
+
+		auto FireCharge = cpp14::make_unique<cFireChargeEntity>(this, GetPosX(), GetPosY() + 1, GetPosZ(), Speed);
+		auto FireChargePtr = FireCharge.get();
+
```
madmaxoft: Get rid of this empty line.

src/Items/ItemPainting.h
```diff
@@ -70,8 +70,10 @@ class cItemPaintingHandler :
 				{ "BurningSkull" }
 			};
 
-			cPainting * Painting = new cPainting(gPaintingTitlesList[a_World->GetTickRandomNumber(ARRAYCOUNT(gPaintingTitlesList) - 1)].Title, a_BlockFace, a_BlockX, a_BlockY, a_BlockZ);
-			Painting->Initialize(*a_World);
+			auto Painting = cpp14::make_unique<cPainting>(gPaintingTitlesList[a_World->GetTickRandomNumber(ARRAYCOUNT(gPaintingTitlesList) - 1)].Title, a_BlockFace, a_BlockX, a_BlockY, a_BlockZ);
+			auto PaintingPtr = Painting.get();
+
```
madmaxoft: Remove this empty line.

src/Items/ItemFishingRod.h
```diff
@@ -244,9 +244,11 @@ class cItemFishingRodHandler :
 		}
 		else
 		{
-			cFloater * Floater = new cFloater(a_Player->GetPosX(), a_Player->GetStance(), a_Player->GetPosZ(), a_Player->GetLookVector() * 15, a_Player->GetUniqueID(), static_cast<int>(100 + static_cast<unsigned int>(a_World->GetTickRandomNumber(800)) - (a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLure) * 100)));
-			Floater->Initialize(*a_World);
-			a_Player->SetIsFishing(true, Floater->GetUniqueID());
+			auto Floater = cpp14::make_unique<cFloater>(a_Player->GetPosX(), a_Player->GetStance(), a_Player->GetPosZ(), a_Player->GetLookVector() * 15, a_Player->GetUniqueID(), static_cast<int>(100 + static_cast<unsigned int>(a_World->GetTickRandomNumber(800)) - (a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLure) * 100)));
+			auto FloaterPtr = Floater.get();
+
```
madmaxoft: Remove this empty line.

src/Simulator/SandSimulator.cpp
```diff
@@ -61,8 +61,12 @@ void cSandSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
 				Pos.x, Pos.y, Pos.z, ItemTypeToString(BlockType).c_str(), ItemTypeToString(BlockBelow).c_str()
 			);
 			*/
-			cFallingBlock * FallingBlock = new cFallingBlock(Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
-			FallingBlock->Initialize(m_World);
+
+			auto FallingBlock = cpp14::make_unique<cFallingBlock>(Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
+			auto FallingBlockPtr = FallingBlock.get();
+
```
madmaxoft: Remove this empty line.

src/World.cpp
```diff
@@ -2211,11 +2214,11 @@ void cWorld::SpawnItemPickups(const cItems & a_Pickups, double a_BlockX, double
 			continue;
 		}
 
-		cPickup * Pickup = new cPickup(
+		auto Pickup = cpp14::make_unique<cPickup>(
 			a_BlockX, a_BlockY, a_BlockZ,
 			*itr, IsPlayerCreated, static_cast<float>(a_SpeedX), static_cast<float>(a_SpeedY), static_cast<float>(a_SpeedZ)
 		);
-		Pickup->Initialize(*this);
+		Pickup->Initialize(std::move(Pickup), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2190,11 +2193,11 @@ void cWorld::SpawnItemPickups(const cItems & a_Pickups, double a_BlockX, double
 		float SpeedY = static_cast<float>(a_FlyAwaySpeed * GetTickRandomNumber(50));
 		float SpeedZ = static_cast<float>(a_FlyAwaySpeed * (GetTickRandomNumber(10) - 5));
 
-		cPickup * Pickup = new cPickup(
+		auto Pickup = cpp14::make_unique<cPickup>(
 			a_BlockX, a_BlockY, a_BlockZ,
 			*itr, IsPlayerCreated, SpeedX, SpeedY, SpeedZ
 		);
-		Pickup->Initialize(*this);
+		Pickup->Initialize(std::move(Pickup), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2225,9 +2228,10 @@ void cWorld::SpawnItemPickups(const cItems & a_Pickups, double a_BlockX, double
 
 UInt32 cWorld::SpawnFallingBlock(int a_X, int a_Y, int a_Z, BLOCKTYPE BlockType, NIBBLETYPE BlockMeta)
 {
-	cFallingBlock * FallingBlock = new cFallingBlock(Vector3i(a_X, a_Y, a_Z), BlockType, BlockMeta);
-	FallingBlock->Initialize(*this);
-	return FallingBlock->GetUniqueID();
+	auto FallingBlock = cpp14::make_unique<cFallingBlock>(Vector3i(a_X, a_Y, a_Z), BlockType, BlockMeta);
+	auto ID = FallingBlock->GetUniqueID();
+	FallingBlock->Initialize(std::move(FallingBlock), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2242,9 +2246,10 @@ UInt32 cWorld::SpawnExperienceOrb(double a_X, double a_Y, double a_Z, int a_Rewa
 		return cEntity::INVALID_ID;
 	}
 
-	cExpOrb * ExpOrb = new cExpOrb(a_X, a_Y, a_Z, a_Reward);
-	ExpOrb->Initialize(*this);
-	return ExpOrb->GetUniqueID();
+	auto ExpOrb = cpp14::make_unique<cExpOrb>(a_X, a_Y, a_Z, a_Reward);
+	auto ID = ExpOrb->GetUniqueID();
+	ExpOrb->Initialize(std::move(ExpOrb), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2253,21 +2258,23 @@ UInt32 cWorld::SpawnExperienceOrb(double a_X, double a_Y, double a_Z, int a_Rewa
...
 		default:
 		{
 			return cEntity::INVALID_ID;
 		}
 	}  // switch (a_MinecartType)
-	Minecart->Initialize(*this);
-	return Minecart->GetUniqueID();
+
+	auto ID = Minecart->GetUniqueID();
+	Minecart->Initialize(std::move(Minecart), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2276,32 +2283,29 @@ UInt32 cWorld::SpawnMinecart(double a_X, double a_Y, double a_Z, int a_MinecartT
 
 UInt32 cWorld::SpawnBoat(double a_X, double a_Y, double a_Z)
 {
-	cBoat * Boat = new cBoat(a_X, a_Y, a_Z);
-	if (Boat == nullptr)
+	auto Boat = cpp14::make_unique<cBoat>(a_X, a_Y, a_Z);
+	auto ID = Boat->GetUniqueID();
+	if (!Boat->Initialize(std::move(Boat), *this))
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2276,32 +2283,29 @@ UInt32 cWorld::SpawnMinecart(double a_X, double a_Y, double a_Z, int a_MinecartT
...
+	auto TNT = cpp14::make_unique<cTNTEntity>(a_X, a_Y, a_Z, a_FuseTicks);
+	auto ID = TNT->GetUniqueID();
 	TNT->SetSpeed(
-		a_InitialVelocityCoeff * (GetTickRandomNumber(2) - 1), /** -1, 0, 1 */
+		a_InitialVelocityCoeff * (GetTickRandomNumber(2) - 1), /* -1, 0, 1 */
 		a_InitialVelocityCoeff * 2,
 		a_InitialVelocityCoeff * (GetTickRandomNumber(2) - 1)
 	);
-	return TNT->GetUniqueID();
+	TNT->Initialize(std::move(TNT), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -2888,10 +2892,9 @@ void cWorld::SetChunkData(cSetChunkData & a_SetChunkData)
 
 	// Initialize the entities (outside the m_ChunkMap's CS, to fix FS #347):
 	cEntityList Entities;
-	std::swap(a_SetChunkData.GetEntities(), Entities);
-	for (cEntityList::iterator itr = Entities.begin(), end = Entities.end(); itr != end; ++itr)
+	for (auto & Entity : a_SetChunkData.GetEntities())
 	{
-		(*itr)->Initialize(*this);
+		Entity->Initialize(std::move(Entity), *this);
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -3737,22 +3742,20 @@ UInt32 cWorld::SpawnMobFinalize(cMonster * a_Monster)
...
-		delete a_Monster;
-		a_Monster = nullptr;
 		return cEntity::INVALID_ID;
 	}
 
+	auto & Monster = *a_Monster;
+
 	// Initialize the monster into the current world.
-	if (!a_Monster->Initialize(*this))
+	if (!a_Monster->Initialize(std::move(a_Monster), *this))
```
madmaxoft: Danger, unknown order of evaluation.

src/World.cpp
```diff
@@ -3761,18 +3764,15 @@ UInt32 cWorld::SpawnMobFinalize(cMonster * a_Monster)
...
-	cProjectileEntity * Projectile = cProjectileEntity::Create(a_Kind, a_Creator, a_PosX, a_PosY, a_PosZ, a_Item, a_Speed);
-	if (Projectile == nullptr)
-	{
-		return cEntity::INVALID_ID;
-	}
-	if (!Projectile->Initialize(*this))
+	auto Projectile = cProjectileEntity::Create(a_Kind, a_Creator, a_PosX, a_PosY, a_PosZ, a_Item, a_Speed);
+	auto ID = Projectile->GetUniqueID();
+
+	if ((Projectile == nullptr) || !Projectile->Initialize(std::move(Projectile), *this))
```
madmaxoft: Danger, unknown order of evaluation.
Remove the empty line above.
No need to check for `nullptr` after actually dereferencing the ptr on the above line. Rather, keep the original separate checks for nullptr and then for initialization.

src/World.cpp
```diff
@@ -3958,29 +3958,35 @@ void cWorld::AddQueuedPlayers(void)
...
 			auto & Player = AwaitingPlayer.first;
-			ASSERT(std::find(m_Players.begin(), m_Players.end(), Player) == m_Players.end());  // Is it already in the list? HOW?
+			ASSERT(std::find(m_Players.begin(), m_Players.end(), Player.get()) == m_Players.end());  // Is it already in the list? HOW?
 			LOGD("Adding player %s to world \"%s\".", Player->GetName().c_str(), m_WorldName.c_str());
 
-			m_Players.push_back(Player);
+			m_Players.push_back(Player.get());
 			Player->SetWorld(this);
 
+			auto PlayerPtr = Player.get();
```
madmaxoft: Move this line 1 down, below the comment.

madmaxoft:
Still more dangerous UBs, and some stylistics.

yangm97:
What kind of in-game test does this require?

sphinxc0re:
@tigerw any updates?

sphinxc0re:
@tigerw What's the matter with this PR?

LogicParrot:
>What kind of in-game test does this require?

Attempting to crash the game by playing around with mobs and possibly world travel. Arrows and tamed wolves are good crash candidates. Does it crash when an already dead skeleton's arrow hit someone? Does it crash when wolves kill their target? What happens if I do `/portal` when a mob is attacking me? etc.

tigerw:
It is, as they say, "in the pipeline" [😊]

Seadragon91:
I just tested @LogicParrot suggestions, no crash.

madmaxoft:
I'd still like my requested changes to be addressed.

src/Items/ItemBow.h
worktycho: The simplest thing would be to make the Initialize function static. Its only overriden by two entities, Wither and Player,and I'm sure we could move the ownership transfer out of the virtual method.

Seadragon91:
Whats now the status @tigerw? Any change to get this finished?

Seadragon91:
Should we just close this?

tigerw:
Absolutely not; memory safety is very important.

yangm97:
Besides, there was a PR that depended on this I think.

src/Entities/ProjectileEntity.cpp
tigerw: but code is dataaaa and styles should be handled by the editooorrr ... \*pout*

src/Items/ItemBow.h
tigerw: Could someone remind me what ill-effects would happen again? (It's been so long...) I cannot identify a link between the `Arrow` pointer variable, the std::move of `ArrowPtr`, and `this` because I am under the impression that everything is initialised and at no point is anything destroyed.

src/World.cpp
tigerw: > No need to check for nullptr after actually dereferencing the ptr

whoops. compiler would've had a field day with that...

src/Items/ItemFishingRod.h
```diff
@@ -252,14 +252,13 @@ class cItemFishingRodHandler :
...
+			auto FloaterPtr = cpp14::make_unique<cFloater>(a_Player->GetPosX(), a_Player->GetStance(), a_Player->GetPosZ(), a_Player->GetLookVector() * 15, a_Player->GetUniqueID(), (Random.RandInt(100, 900) - static_cast<int>(a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLure) * 100)));
+			auto Floater = FloaterPtr.get();
+			if (!Floater->Initialize(std::move(FloaterPtr), *a_World))
 			{
-				delete Floater;
-				Floater = nullptr;
 				return false;
 			}
-			a_Player->SetIsFishing(true, Floater->GetUniqueID());
+			a_Player->SetIsFishing(true, FloaterPtr->GetUniqueID());
```
Seadragon91: This change should be reverted as `FloaterPtr` has been moved above.

tigerw:
@madmaxoft @LogicParrot will you kindly take a look again?

Seadragon91:
I see that you have there inconsistency with the suffix `Ptr`, sometimes it's a unique_ptr or it's a raw pointer.
Example: `ArrowPtr` is a unique_ptr, but `FloaterPtr` is a raw pointer.

tigerw:
You are completely correct.

src/Items/ItemFishingRod.h
Seadragon91: Now FloaterPtr is a raw pointer.

tigerw:
Nao?

Seadragon91:
Server crashes on joining:
`cEntity::IsPlayer (this=0x0) at /home/lukas/cpp/cuberite/src/./UI/../Entities/Entity.h:168`

tigerw:
..wow. I'm an idiot.

tigerw:
Is this now ok?

tigerw:
Query?

Seadragon91:
Can we merge @madmaxoft?

src/Chunk.cpp
```diff
@@ -708,20 +707,24 @@ void cChunk::Tick(std::chrono::milliseconds a_Dt)
...
-			MoveEntityToNewChunk(*itr);
+
 			// Mark as dirty if it was a server-generated entity:
 			if (!(*itr)->IsPlayer())
 			{
 				MarkDirty();
 			}
+
+			MoveEntityToNewChunk(std::move(*itr));
+
```
madmaxoft: Can the `SetParentChunk` call from a few lines above be moved to here, as well as the comment belonging to it? So that the logical code block doesn't get broken.

src/ClientHandle.cpp
```diff
@@ -184,7 +188,23 @@ void cClientHandle::Destroy(void)
...
+				m_PlayerPtr = world->RemovePlayer(*player, true);
+
+				// And RemovePlayer should have returned a valid smart pointer
+				ASSERT(m_PlayerPtr);
+			}
+			else
+			{
+				// If ownership was not transferred, our own smart pointer should be valid and RemovePlayer's should not
+				ASSERT(m_PlayerPtr);
+				ASSERT(!world->RemovePlayer(*player, true));
```
madmaxoft: This should probably use `VERIFY` instead of `ASSERT` (so that the `RemovePlayer` call stays even in Release mode).

src/Entities/Player.cpp
madmaxoft: At least break this up into several lines, use a temp variable, and move the comment above this block.

src/Entities/ProjectileEntity.cpp
madmaxoft: Let me give you this pearl of programming wisdom:

**Code is written once, but read twenty times**

So do all of us a favor and make it so that it reads more easily - you'll be saving all of us, including yourself, ton of time later on :)

src/SetChunkData.cpp
```diff
@@ -6,6 +6,7 @@
 #include "Globals.h"
 #include "SetChunkData.h"
 #include "BlockEntities/BlockEntity.h"
+#include "Entities/Entity.h"
```
madmaxoft: This include seems not needed, or is it?

src/World.cpp
```diff
@@ -1005,9 +1005,12 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
...
 			Entity->SetWorld(this);
-			m_ChunkMap->AddEntity(Entity);
-			ASSERT(!Entity->IsTicking());
-			Entity->SetIsTicking(true);
+
+			auto EntityPtr = Entity.get();
+			m_ChunkMap->AddEntity(std::move(Entity));
+
+			ASSERT(!EntityPtr->IsTicking());
+			EntityPtr->SetIsTicking(true);
```
madmaxoft: Why the empty lines? Just keep it together.

src/World.h
```diff
@@ -267,13 +267,13 @@ class cWorld :
...
 	If a_OldWorld is provided, a corresponding ENTITY_CHANGED_WORLD event is triggerred after the addition. */
-	void AddPlayer(cPlayer * a_Player, cWorld * a_OldWorld = nullptr);
+	void AddPlayer(std::unique_ptr<cPlayer> a_Player, cWorld * a_OldWorld = nullptr);
 
 	/** Removes the player from the world.
 	Removes the player from the addition queue, too, if appropriate.
 	If the player has a ClientHandle, the ClientHandle is removed from all chunks in the world and will not be ticked by this world anymore.
 	@param a_RemoveFromChunk determines if the entity should be removed from its chunk as well. Should be false when ticking from cChunk. */
-	void RemovePlayer(cPlayer * a_Player, bool a_RemoveFromChunk);
+	std::unique_ptr<cPlayer> RemovePlayer(cPlayer & a_Player, bool a_RemoveFromChunk);
```
madmaxoft: The return value should be documented.

src/Entities/ProjectileEntity.cpp
tigerw: Wise words.

src/ClientHandle.cpp
tigerw: I think the intention was that here, the player shouldn't have been added to the world at all (WasAddedToWorld is false, disconnection before Authenticate called) and so the debug-only call is just to verify that, at least for the appropriate chunk, it did not store any owning handles.

The threading relation between Authenticate and Destroy should be OK as you added a mutex guarding actions depending on the value of `m_State`.

src/World.cpp
tigerw: Looks nicer?

src/ClientHandle.cpp
madmaxoft: Still, if it ever happens in Release mode, something should be done - at the very least, log a message; I'd also kick the client, something, anything :)

src/World.cpp
madmaxoft: Not really. Empty lines are good for separating logical pieces of code, but if every other line is empty, it's like they aren't there anymore.

This whole thing has one purpose - add the entity into the chunkmap, with all bells and whistles relevant to it. So it makes sense for it to be one logical block of code.

src/ClientHandle.cpp
tigerw: Isn't the purpose of an assert to catch faulty assumptions in debug though, so release is faster?

src/World.h
tigerw: An aside; did we decide to use Javadoc-style comments (the `@param` tag)? I believe XML comments are fully supported by Doxygen.

src/SetChunkData.cpp
tigerw: cSetChunkData stores a vector of StoredEntity's so compilation fails without it.

src/ClientHandle.cpp
madmaxoft: Okay, I guess I'm being overly paranoic again.

madmaxoft: This just feels wrong. So you're calling `RemovePlayer` only to check that the player is not actually present, you don't expect any removal. That's pretty misleading.

src/World.h
madmaxoft: We didn't decide anything, some people use them and I don't care enough to either enforce or discourage them.

src/ClientHandle.cpp
tigerw: I could add a debug-only function to ensure there are no stray references left in the world or chunk, or just remove the check.

The intention of decoupling the lifetime of the player and client handle would refactor the WasAddedToWorld hack away, so I do not expect this to stay for ever. It is, after all, just a shim to get smart pointers working.

madmaxoft: Observe your first reply: "I think the intention was...." - even you're not sure anymore, so the code really smells :)
Yes please, add a debug-only function for this check. It won't be the first nor the last. (*ehm* *ehm* `cCriticalSection::IsLockedByCurrentThread` *ehm* *ehm* )

src/ChunkMap.cpp
```diff
@@ -1497,38 +1497,38 @@ void cChunkMap::RemoveClientFromChunks(cClientHandle * a_Client)
...
 	cChunkPtr Chunk = GetChunk(a_Entity->GetChunkX(), a_Entity->GetChunkZ());
 	if (Chunk == nullptr)  // This will assert inside GetChunk in Debug builds
 	{
 		LOGWARNING("Entity at %p (%s, ID %d) spawning in a non-existent chunk, the entity is lost.",
-			static_cast<void *>(a_Entity), a_Entity->GetClass(), a_Entity->GetUniqueID()
+			static_cast<void *>(a_Entity.get()), a_Entity->GetClass(), a_Entity->GetUniqueID()
 		);
 		return;
 	}
 	if (!Chunk->HasEntity(a_Entity->GetUniqueID()))
```
peterbell10: How could the chunk already have the entity if it's owned by the current scope?

madmaxoft:
I hate this PR. I've already read the code for like ten times, after each change, got a major headache out of it and still I have to read it and review it again. This is what hell feels like - doing the same thing over and over again, but a little bit different, but the same :P

yangm97:
#prayforxoft

tigerw:
I've no doubt you've seen worse :)

src/ChunkMap.cpp
tigerw: The original change was in af4a21e with modification to current form in 366ecf9, back in 2014.

The comment for the check was
> // Add to chunkmap, if not already there (Spawn vs MoveToWorld):

I can't see why this is needed since it seems when moving to another world, even the same one, the entity is always removed first from a chunkmap. (@madmaxoft?)

It seems, especially with the new changes, the check isn't needed, unless something goes very wrong and the smart pointer no longer holds unique ownership?

Why not leave it, just in case though; don't want to give xoft an even bigger mess to go through ;)

peterbell10: > Why not leave it

If the check ever fails it's basically a guaranteed use after delete though.  IMO the function should be removed and it's use in `cWorld` should be replaced with an `ASSERT` (or `VERIFY`?) and a call to `AddEntity`.

madmaxoft:
Well, maybe the redstone sim was worse, but for that I didn't care about what it did. This one, I need to care.

src/ChunkMap.cpp
tigerw: Would you be ok if I put this change in a new PR?

tigerw:
\*we have reports of a casualty; I repeat, we have reports of a casualty*

----------

So that means merge?

src/ChunkMap.cpp
peterbell10: I think it is literally only used once in the whole of cuberite so it shouldn't be a big change but I suppose there's no real harm in it staying short term.

tigerw:
The suspense is unbearable.

madmaxoft:
Alright, let's merge and deal with the resulting fire afterwards.

tigerw:
I have the hoses and fire appliance ready.

LogicParrot:
Time for serious ingame testing.

madmaxoft:
@tigerw please learn to use squash-merging. You're making mess in the git history.

LogicParrot:
@tiger are you planning any followups? We should synchronize to avoid conflicts. I may modify the m_Target system.

LogicParrot:
Surprisingly, things aren't going up in flames and so far I could not find any bugs related to this changeset (world travel not tested yet).

LogicParrot:
We're going to have many behaviors that sometimes need entity pointers. (Similar to cMonster::m_Target). But I see that cChunk::m_entities is a vector of unique_ptrs.  Any advice on how to make this safe cleanly?

LogicParrot:
Possible approaches:

#1: https://github.com/cuberite/cuberite/pull/3295#r73788185

#2: Storing a UUID, and attempting to convert it to a raw pointer in the tick, never storing the raw pointer. If the entity is no longer present, the conversion fails.

#3: Shared ptr.

PR #3491, opened 2016-12-23
Refactored to put URL Encoding / Decoding in a single place.

madmaxoft:
The canon in the Lua API is `cUrlParser:UrlDecode` and `cUrlParser:UrlEncode`. In the C++ code the functions are `URLDecode` and `URLEncode`, both in StringUtils.

The decoding now supports `%u0xxxx` characters and provides proper error checking.

Since this change is rather deep, I'd like someone to verify that nothing major has broken for regular gameplay (I can't test that right now). The refactoring has touched reading strings from the MC protocol.


madmaxoft:
I've tested this with 1.10 client and all seems working.

PR #3495, opened 2016-12-28
Protocol: Add default cases to switch statements

stone3311:
Fixed compiler warnings about values not being handled in some switch statements.


sphinxc0re:
I'm not sure about this. @cuberite/core-team

madmaxoft:
Those warnings were there for a reason - the code is incomplete and does indeed need finishing. Removing the warnings will hide this issue.

PR #3497, opened 2016-12-28
Fixed #3386

### LINKED ISSUE #3386: Wither effect deaths crash server
Client version:   1.10.2
Server OS:        Linux
### Expected behavior

Player dies normally from wither effect, without causing a server crash.
### Actual behavior

When a player dies while having the wither effect, the server crashes.

Apply the wither effect with
`Player:AddEntityEffect(20, 9999, 1, 1)`
in a plugin, and wait for your player to die.

stone3311:
Fixed issue #3386.

`Effect->OnTick(*this);` caused the iterator to become invalid. This issue is fixed by saving the effect and calling `OnTick` after the loop.

### DIFF
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -48,6 +48,8 @@ void cPawn::Destroyed()
 
 void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
+	std::vector<cEntityEffect*> EffectsToTick;
+
 	// Iterate through this entity's applied effects
 	for (tEffectMap::iterator iter = m_EntityEffects.begin(); iter != m_EntityEffects.end();)
 	{
@@ -55,7 +57,8 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		cEntityEffect::eType EffectType = iter->first;
 		cEntityEffect * Effect = iter->second;
 
-		Effect->OnTick(*this);
+		// Call OnTick later to make sure the iterator won't be invalid
+		EffectsToTick.push_back(Effect);
 
 		// Iterates (must be called before any possible erasure)
 		++iter;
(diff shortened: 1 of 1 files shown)


sphinxc0re:
Does this only fix the one issue? Are there more issues leading back to the same problem?

src/Entities/Pawn.cpp
```diff
@@ -48,14 +48,17 @@ void cPawn::Destroyed()
 
 void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
+	std::vector<cEntityEffect*> EffectsToTick;
```
madmaxoft: Style: add space before the `*`

src/Entities/Pawn.cpp
```diff
@@ -69,6 +72,12 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		// TODO: Check for discrepancies between client and server effect values
 	}
 
+
+	for (cEntityEffect * Effect : EffectsToTick)
```
madmaxoft: Use `auto` or even `const auto`, if possible.

madmaxoft:
This is definitely a needed change and there might be more issues caused by that error, not necessarily reported.

sphinxc0re:
Why is CircleCI dumb again? :grin:

madmaxoft:
I've just checked it and found out they had added an option not to build PRs from forked repos and made it the default, I reset it, so now the builds should work again. We'll see when @stone3311 pushes the changes I requested to his branch, it should trigger the rebuild.

stone3311:
The crash had only happened when dying due to the wither effect. Getting killed or issuing /kill in the console while having the wither effect did not crash the server.

sphinxc0re:
you should rebase your branch onto the current master @stone3311

stone3311:
Ok, I rebased it onto the current master @sphinxc0re

madmaxoft:
Nah, there was no need to rebase, this is a change to be squash-merged anyway. And you didn't rebase onto master, you merged master (which we don't like doing, it creates loops in the history). Thanks for trying, anyway. Merging now.

stone3311:
Sorry about that.

PR #3498, opened 2016-12-29
Track skin part and main hand preferences (#2149)

### LINKED ISSUE #2149: Mask doesn't show up on skins
My skin has a mask (I think that's what it's called, the extra layer on top of the head) that shows up on every server except my test MCServer
mcserver skin bug

Pokechu22:
Track the skin part preference and main hand preference within Client Settings.

This implements the main hand preference.  It doesn't do anything else with it (it doesn't implement dual-wielding), but it does let you specify that you want the left hand to be the main hand visually (1.9+).

Fixes #2149.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11351,6 +11351,16 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Returns the player's current maximum speed, relative to the game default speed. Takes into account the sprinting / flying status.",
 				},
+				GetMainHand =
+				{
+					Returns =
+					{
+						{
+							Type = "eMainHand",
+						},
+					},
+					Notes = "Returns the player's main hand.",
+				},
 				GetName =
 				{
 					Returns =
@@ -11410,6 +11420,16 @@ a_Player:OpenWindow(Window);
 						},
 					},
 					Notes = "Returns the player's maximum sprinting speed, relative to the game default speed. Defaults to 1.3, but plugins may modify it for faster or slower sprinting.",
+				},
+				GetSkinParts =
+				{
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+					},
+					Notes = "Returns the player's current set of skin part flags.  This is a bitwise OR of various {{eSkinPart}} constants.  Note that HasSkinPart may be easier to use in most situations.",
 				},
 				GetStance =
 				{
@@ -11534,6 +11554,23 @@ a_Player:OpenWindow(Window);
 						},
 					},
 					Notes = "Returns true if the player has the specified permission",
+				},
+				HasSkinPart =
+				{
+					Params =
+					{
+						{
+							Name = "Part",
+							Type = "eSkinPart",
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns true if the player has the specified skin part enabled",
 				},
 				Heal =
 				{
@@ -12063,6 +12100,17 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Sets the 'IsFishing' flag for the player. The floater entity ID is expected for the true variant, it can be omitted when IsFishing is false. FIXME: Undefined behavior when multiple fishing rods are used simultanously",
 				},
+				SetMainHand =
+				{
+					Params =
+					{
+						{
+							Name = "Hand",
+							Type = "eMainHand",
+						},
+					},
+					Notes = "Sets the main hand of the player.",
+				},
 				SetName =
 				{
 					Params =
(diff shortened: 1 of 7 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11535,6 +11555,23 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Returns true if the player has the specified permission",
 				},
+				HasSkinPart =
+				{
+					Params =
+					{
+						{
+							Name = "Part",
+							Type = "string",
```
madmaxoft: Wrong type

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11411,6 +11421,16 @@ a_Player:OpenWindow(Window);
...
 				},
+				GetSkinParts =
+				{
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+					},
+					Notes = "Returns the player's current set of skin part flags.",
```
madmaxoft: A bit more info about the return value would be good - how is the set represented, what does the number mean, a link to the skin part flags enum...

src/Defines.h
```diff
@@ -203,6 +203,41 @@ enum eMobHeadRotation
 
 
 
+enum eHand
+{
+	HAND_MAIN = 0,
+	HAND_OFF = 1
+} ;
```
madmaxoft: This enum doesn't seem to be used anywhere at all.

src/Defines.h
```diff
@@ -203,6 +203,41 @@ enum eMobHeadRotation
...
+
+
+
+
+
+enum eMainHand
+{
+	MAIN_HAND_LEFT = 0,
+	MAIN_HAND_RIGHT = 1
```
madmaxoft: We prefer Delphi-style names for enums in new code - `mhLeft`, `mhRight`

src/Defines.h
```diff
@@ -203,6 +203,41 @@ enum eMobHeadRotation
...
+
+enum eSkinPart
+{
+	SKIN_PART_CAPE = 0x01,
+	SKIN_PART_JACKET = 0x02,
+	SKIN_PART_LEFT_SLEEVE = 0x04,
+	SKIN_PART_RIGHT_SLEEVE = 0x08,
+	SKIN_PART_LEFT_PANTS = 0x10,
+	SKIN_PART_RIGHT_PANTS = 0x20,
+	SKIN_PART_HAT = 0x40
```
madmaxoft: Again, Delphi-style enum names - `spCape` etc.

src/Entities/Player.h
```diff
@@ -499,6 +499,13 @@ class cPlayer :
 	The blocks in range (a_BlockX - a_Range, a_BlockX + a_Range) are sent (NY-metric). */
 	void SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_Range = 1);
 
+	bool HasSkinPart(eSkinPart a_Part) const { return (m_SkinParts & a_Part) != 0; }
+	UInt8 GetSkinParts(void) const { return m_SkinParts; }
+	void SetSkinParts(UInt8 a_Parts);
```
madmaxoft: Cannot use `UInt8` for bindings (missing range-checks), use a regular `int` or `unsigned` instead.

src/Entities/Player.h
```diff
@@ -682,6 +689,12 @@ class cPlayer :
 
 	AString m_CustomName;
 
+	/** Displayed skin part bit mask */
+	UInt8 m_SkinParts;
```
madmaxoft: Size-specific integers should only be used in protocol stuff or when it's critical to memory use; for generic stuff use generic types, such as `unsigned`

src/Defines.h
Pokechu22: True.  I created it for future use, though, because if there was only a `eMainHand` then people might think to use _that_ for main/off hand purposes in the future (which'd cause confusion).  However, probably there should be a note explaining that.

src/Entities/Player.h
Pokechu22: Is it possible/logical to use the enum here?  Or should C++ enum types not be used for flag variables?  (IE, would `eSkinPart m_SkinParts` or `eSkinParts m_SkinParts` be valid/logical?  Or should I switch to `unsigned`?)

Pokechu22:
Rebased and attempted to address all requests except for the `eHand` one.

src/Entities/Player.h
madmaxoft: No, `enum` means "one of", you want multiple choice, so it has to be a regular numeric type. Another possibility would be `std::set<eSkinPart>`, but that would be an overkill and it would be difficult to convert between the protocol value and this value. I believe an `unsigned` is the best approach here.

Pokechu22: I used `int` - should I change it to `unsigned`?

madmaxoft: Nah, it's okay.

PR #3501, opened 2016-12-29
Fix crash after calling OnTick on removed effect

stone3311:
This fixes a crash introduced in #3497 where the `OnTick` function of a removed effect would be called, resulting in a segfault.


madmaxoft:
Good catch, I haven't thought of that while reading 3497.

Issue #3506, opened 2016-12-30
FastBreak anticheat bugs


mathiascode:
The FastBreak anticheat currently has a few issues. Blocks such as ice, rails and pressure plates cannot be dug with a pickaxe, instead the "FastBreak?" message appears in chat. It seems like other Cuberite users have had issues with FastBreak as well (https://forum.cuberite.org/thread-2660.html), so I suggest that we disable the anticheat temporarily until the issue is fixed.

mohe2015:
Yes, we should do that. If I find time I'll try to fix the issue.

PR #3509, opened 2017-01-01
Disabled anticheat temporarily

mathiascode:
This PR disables the fastbreak anticheat temporarily until #3506 is solved.


mohe2015:
I think that would only disable it if the entry isn't already in the config file.

mohe2015:
cClientHandle::FASTBREAK_PERCENTAGE = 0
should work.

madmaxoft:
I'm in favor of hard-setting it to zero, too, ignoring the INI setting.

LogicParrot:
Missing `;`

PR #3511, opened 2017-01-01
Temporarily Disabled Anticheat

LogicParrot:
#3509 alternative


madmaxoft:
You should add a note in a comment explaining why the INI file read is disabled.

PR #3512, opened 2017-01-01
StringUtils: Dropped an unneeded copy in lower-/upper-casing.

madmaxoft:


madmaxoft:
@tigerw Any chance of getting that review?

Issue #796, opened 2014-03-11
Add ability to specify a custom shutdown message


Nikeron:
Add a "shutdown message" settings...

Gamerboy59:
Is it supported now?

SphinxC0re:
I don't think so. But I believe this to be kinda easy to implement. On the other hand, this is very low on our todo list

Gamerboy59:
Ya, it's no core feature but good for some extra comfort and also a good feature for the webadmin.

Schwertspize:
well, i have looked in the code, i can nowhere find the "Disconnected" reason or any kick packet. should we kick players before shutting the server down? It appears that we dont or cannot send a reason if we just close the connection

Gamerboy59:
Yes, players should be kicked before shutdown. I think that's the way to do it.

Schwertspize:
Good. Then it should be rather easy to implement 😛 I just thought it was the case that we should handle the connection a bit different. Anyway, I am going to try that.

sweetgiorni:
I'm gonna take a whack at this if no one objects.

Seadragon91:
I think this can be closed?

PR #3518, opened 2017-01-03
Custom Disconnect Message

sweetgiorni:
Adds a m_ShutdownMessage option to settings.ini. When the stop command
is issued, players are kicked with said message before the server shuts
down.

Although, it seems like the server fails to send the packet if you don't wait at all. My solution isn't permanent, but I couldn't figure out a better way to do it. Any suggestions?  :confused:

 #796

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -332,6 +332,22 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 
 void cRoot::StopServer()
 {
+	// Kick all players from the server with custom disconnect message
+	class cPlayerCallback : public cPlayerListCallback
+	{
+		AString m_ShutdownMessage;
+		virtual bool Item(cPlayer * a_Player)
+		{
+			a_Player->GetClientHandlePtr()->Kick(m_ShutdownMessage);
+			return false;
+		}
+	public:
+		cPlayerCallback(AString a_ShutdownMessage) : m_ShutdownMessage(a_ShutdownMessage) {}
+	}PlayerCallback((m_Server->GetShutdownMessage()));
+
+	cRoot::Get()->ForEachPlayer(PlayerCallback);
+	// What's a better way to do this?
+	std::this_thread::sleep_for(std::chrono::seconds(1));
 	m_TerminateEventRaised = true;
 	m_StopEvent.Set();
 	m_InputThreadRunFlag.clear();
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -343,11 +343,14 @@ void cRoot::StopServer()
 		}
 	public:
 		cPlayerCallback(AString a_ShutdownMessage) : m_ShutdownMessage(a_ShutdownMessage) {}
-	} PlayerCallback((m_Server->GetShutdownMessage()));
+	};
 
-	cRoot::Get()->ForEachPlayer(PlayerCallback);
-	// What's a better way to do this?
-	std::this_thread::sleep_for(std::chrono::seconds(1));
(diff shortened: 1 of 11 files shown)


src/Root.cpp
```diff
@@ -332,6 +332,22 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
...
+	{
+		AString m_ShutdownMessage;
+		virtual bool Item(cPlayer * a_Player)
+		{
+			a_Player->GetClientHandlePtr()->Kick(m_ShutdownMessage);
+			return false;
+		}
+	public:
+		cPlayerCallback(AString a_ShutdownMessage) : m_ShutdownMessage(a_ShutdownMessage) {}
+	}PlayerCallback((m_Server->GetShutdownMessage()));
```
madmaxoft: Style: Add a space after `}`

madmaxoft:
You added a new API function, `cServer:GetShutdownMessage()`, so you need to document it (that's why the CI is failing). Have a look at Server/Plugins/APIDump/APIDesc.lua` file, that's where the documentation resides, I think the format is pretty straightforward.

madmaxoft:
I don't think we can come up with a way to wait for the D/C packet that is bulletproof, so just waiting is okay. However, the wait should be avoided if there were no players.

src/Root.cpp
```diff
@@ -332,6 +332,24 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
...
+	{
+		AString m_ShutdownMessage;
+		virtual bool Item(cPlayer * a_Player)
+		{
+			a_Player->GetClientHandlePtr()->Kick(m_ShutdownMessage);
+			return false;
+		}
+	public:
+		cPlayerCallback(AString a_ShutdownMessage) : m_ShutdownMessage(a_ShutdownMessage) {}
+	};
```
sphinxc0re: Style: Pls insert an empty line here

src/Root.cpp
```diff
@@ -332,6 +332,24 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
...
+	if (m_Server->GetNumPlayers())
```
sphinxc0re: You shouldn't hope for falsyness here

madmaxoft: I think the flow should be inverse - set a flag in `cPlayerCallback` that a D/C message has been sent to a player, execute it `ForEachPlayer` and then check the value:
```cpp
cRoot::Get()->ForEachPlayer(callback);
if (callback.m_HasSentDisconnect)
{
  std::this_thread::sleep_for(...);
}
```

As it is now, the code is prone to race conditions - a player can connect or disconnect between the `GetNumPlayers()` call and the `ForEachPlayer()` call. Not that it would matter much here, but let's not get in the habit of having problematic code.

sweetgiorni:
Alright, I fixed that stuff I think. Let me know if there's anything else.

madmaxoft:
@sphinxc0re This should have been a squash-merge! Now there are duplicate commits in the history, and loops, and whatnot.

sphinxc0re:
Oh damn! I missed that! Has been a long night. I shouldn't hit the merge button after 11:00 PM anymore 😅

mohe2015:
I don't know but isn't it possible to undo the merge and squash-merge then?

madmaxoft:
No, once it's in the master, it stays there.

Of course it is possible in git to overwrite a branch, but that shouldn't be done on a branch that is publicly shared.

PR #3520, opened 2017-01-04
Changed fish launching mechanism

sweetgiorni:
Before, fishing loot wouldn't fly towards the player fast enough; it would appear next to the floater and then sink. I added some magic numbers and now the fish fly at a much more appropriate velocity. 

There was also the problem that depending on how long the player waited before reeling in the loot, the floater could be deep underwater or closer to the surface. This resulted in the loot having inconsistent velocities due to the extra drag the water adds. I resolved this by saving the position of the floater just before it sinks underwater and spawning the loot there instead.
 :cloud: :fish:  :cloud:


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6585,6 +6585,15 @@ cFile:DeleteFile("/usr/bin/virus.exe");
...
 					Notes = "Returns the EntityID of the player who owns the floater.",
 				},
+				GetBitePos =
+				{
+					Returns =
+					{
+							Name = "BitePosition",
+							Type = "Vector3d",
+					},
+					Notes = "Returns the position of the floater just before a fish bites.",
```
NiLSPACE: Perhaps also provide info what it returns if a fish hasn't bitten yet?

src/Items/ItemFishingRod.h
```diff
@@ -236,9 +239,11 @@ class cItemFishingRodHandler :
 				{
 					return true;
 				}
-				Vector3d FloaterPos = FloaterInfo.GetPos();
+				Vector3d FloaterPos = FloaterInfo.GetBitePos();
+				const float FISH_SPEED_MULT_XZ = 2.7f;
+				const float FISH_SPEED_MULT_Y = 2.0f;
```
NiLSPACE: Do these numbers work correctly even while the player is closer or further away from the floater?

Server/Plugins/APIDump/APIDesc.lua
sweetgiorni: Fixed

src/Items/ItemFishingRod.h
sweetgiorni: I tweaked it again a little bit but yes, for the most part it scales well for all _reasonable_ heights and distances. Although, the minecraft wiki says that moving 32 blocks away should break the line, but that doesn't happen. That's probably the next thing I'll work on.

sphinxc0re:
@sweetgiorni any updates on this one?

PR #3522, opened 2017-01-04
Fixed BroadcastSoundParticleEffect API function

### LINKED ISSUE #3505: BroadcastSoundParticleEffect doesn't work
### Expected behavior
Plugin developers can use BroadcastSoundParticleEffect in their plugins without issues.

### Actual behavior
BroadcastSoundParticleEffect does not work at the moment. The API documentation says that you should specify an effect ID, but it returns this error in the console:
```
error in function 'BroadcastSoundParticleEffect'.
     argument #2 is 'number'; 'const EffectID' expected.
```
This problem was introduced in 83870f9fc0e19da646f408a2d6636adfe4b5c002.

Use the super pickaxe in WorldEdit. Activate it by executing the command //, and break blocks using a pickaxe.

sweetgiorni:
Changed EffectID and SmokeDirection into normal enums, then added EffectID.h to AllToLua.pkg.
The function can now be used from lua like so:
~~~
world:BroadcastSoundParticleEffect(SFX_MOB_ZOMBIE_METAL,Player:GetPosX(),Player:GetPosY(),Player:GetPosZ(), 0)
~~~
or
~~~
world:BroadcastSoundParticleEffect(1011,Player:GetPosX(),Player:GetPosY(),Player:GetPosZ(), 0)
~~~

Fixes #3505

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -16811,6 +16811,19 @@ end
 						specifically otherwise in the world.ini file.
 					]],
 				},
+				EffectID = 
+				{
+					Include = 
+					{
+						"^SFX_.*",
+						"^PARTICLE_*",
+					},
+					TextBefore = [[
+						These constants are used for effect ids. They are used in the {{cWorld}}:BroadcastSoundParticleEffect()
+						function.
+						
+					]],
+				},
 				eExplosionSource =
 				{
 					Include = "^es.*",
@@ -16996,6 +17009,21 @@ end
 				MetaValues =
 				{
 					Include = "^E_META_.*",
+				},
+				EffectID = 
+				{
+					Include = 
+					{
+						"^SOUTH.*",
+						"^NORTH.*",
+						"EAST",
+						"CENTRE",
+						"WEST",
+					},
+					TextBefore = [[
+						These constants are smoke directions ids. They are used in the {{cWorld}}:BroadcastSoundParticleEffect()
+						function.
+					]],
 				},
 			},
 		},
--- a/src/EffectID.h
+++ b/src/EffectID.h
@@ -1,6 +1,8 @@
 #pragma once
 
-enum class EffectID : Int32
+// tolua_begin
+
+enum EffectID
 {
 	SFX_RANDOM_CLICK_1 = 1000,
 	SFX_RANDOM_CLICK_2 = 1001,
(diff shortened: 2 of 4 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -16997,6 +17010,21 @@ end
 				{
 					Include = "^E_META_.*",
 				},
+				EffectID =
```
madmaxoft: Wrong name

src/EffectID.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
-enum class EffectID : Int32
+// tolua_begin
+
+enum EffectID
```
madmaxoft: This is not a good change. The `: Int32` specifier was there on purpose, so that the enum cannot be implicitly-converted from other types. Since ToLua doesn't support this syntax, you should use a slightly different approach:
```cpp
#ifndef TOLUA_EXPORT_PARSER
  enum class EffectID: Int32
#else
  enum EffectID  // tolua_export
#endif
// tolua_begin
{
  ...
};
```

madmaxoft:
Also, you need to explicitly document each enum value in the Lua API docs (in addition to the grouping that you've already done).

src/EffectID.h
sweetgiorni: That doesn't seem to work. I'm pretty green with tolua but is this relevant?
http://www8.cs.umu.se/kurser/TDBD12/VT04/lab/lua/tolua++.html
>New on tolua++: All other preprocessor directives are ignored.

Server/Plugins/APIDump/APIDesc.lua
sweetgiorni: Erm... the other constant group names seem to just be the name of whatever the enum is. Can I have a hint?

src/EffectID.h
madmaxoft: How exactly doesn't it work? What is your code?

Basically ToLua processes only lines that are either marked with `// tolua_export`, or between `// tolua_begin` and `// tolua_end`; those lines, when filtered, need to comprise a valid C++ source. Naturally, the tolua_begin and tolua_end lines need to be properly balanced and only 1 level deep.

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: Exactly, and the name of the enum is `SmokeDirection`, not `EffectID`. Not to mention that having a duplicate entry in the description table silently hides the previous same-name entries.

sphinxc0re:
@sweetgiorni are you still working on this?

sphinxc0re:
Closing because of absence of communication

PR #3526, opened 2017-01-06
Added smelting recipes and fuels

Bond-009:
http://minecraft.gamepedia.com/Smelting


madmaxoft:
Looks good, but why did you commit the update to the Core plugin, that is not needed, right? Do you know enough git to remove that commit? If not, it's alright, it's not a showstopper, only it'd be neater :)

Issue #3492, opened 2016-12-24
C++ assert failure when Lua callback asserts


madmaxoft:
When executing a callback in a Lua plugin hits a Lua `assert`, the server then C++-asserts inside `cLuaState`.

To replicate:
Execute the following plugin code:
```lua
someWorld:ChunkStay({0, 0}, function() assert(false) end, function() assert(false) end)
```
with a Cuberite Debug binary. The server asserts that there are not 2 values on the stack while removing the error handler.

Investigate what the stack contents are and how many values must be popped off in this case, and whether it's the same as hitting a real error handler (possibly handled with ZBS's debugger callback).

madmaxoft:
Fixed by #3755

PR #3529, opened 2017-01-08
LuaState: Fixed race condition in ref tracking.

madmaxoft:
Fixes a problem mostly seen when Lua callbacks are called from multiple threads (seen with new GalExport communicating with MCSchematicToPng via JSON messages)

The problem was that the references used "canon Lua state" value, which they queried from the Lua state without locking its mutex first. Now the value is stored in a global map with a single mutex protecting it, so it no longer touches the Lua state before locking its mutex.

I assume this is a partial cause of #3492 , but not a complete fix yet.


sphinxc0re:
Why did you merge the master? Did you use the GH update tool?

madmaxoft:
Yes, as long as a PR gets merged via "squash-merge", the merge-master commits are okay, and in fact are a good way to make sure the PR still works with current master.

PR #3534, opened 2017-01-14
Changed places where dead bushes generate

Bond-009:
I'm not sure about `E_META_DIRT_*`

### DIFF
--- a/Server/items.ini
+++ b/Server/items.ini
@@ -540,6 +540,7 @@ activatorrail=157
 dropper=158
 stained_hardened_clay=159
 stainedhardenedclay=159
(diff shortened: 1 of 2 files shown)


madmaxoft:
Unfortunately this won't work at all, the finisher only expects and checks blocktypes, it ignores blockmetas. You'd need to update the code to check the metas as well

NiLSPACE:
Perhaps the `DeadBushes` finisher should be moved to `cFinishGenClumpTopBlock` instead?

Bond-009:
I have no idea how to change it so it accepts blockmetas.

PR #3537, opened 2017-01-15
Lua stress test

madmaxoft:
Added a `cLuaState` stress test for multithreading + callbacks + reloading, found and fixed one issue in the ref tracking that could have caused crashes.


madmaxoft:
Note: this PR should be merged, rather than squash-merged

tests/LuaThreadStress/LuaThreadStress.cpp
```diff
@@ -0,0 +1,154 @@
...
+	while (!a_ShouldTerminate->load())
+	{
+		// Pick a random operation on the Lua state and peform it:
+		switch (rnd() % 4)
+		{
+			case 0:
+			{
+				// Get a new callback:
+				callbackSeed = callbackSeed + 1;
+				callback = getCallback(*a_LuaState, callbackSeed);
```
NiLSPACE: Shouldn't this have a break?

madmaxoft: It doesn't strictly need to, but indeed *something* is missing - either a `break`, or a `// fallthrough` comment

PR #3544, opened 2017-01-17
DeadlockDetect more info

madmaxoft:
The deadlock detector now lists more information about the deadlock, including the lock-state of some selected CSs.

This has been inspired by recent email exchange with hebe-87 who seems to be running into deadlocks too often.

### DIFF
--- a/Server/Plugins/Debuggers/Debuggers.lua
+++ b/Server/Plugins/Debuggers/Debuggers.lua
@@ -2420,6 +2420,34 @@ end
 
 
 
+function HandleConsoleDeadlock(a_Split)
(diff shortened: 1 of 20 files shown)


madmaxoft:
I'd like to get this merged soon. Does anyone object?

I know the code is not very clean - accessing member variables in a thread-unsafe way, but it's either that, or we lose (quite a lot of) performance altogether, and the fragile code is only called when the DeadlockDetect is about to abort the server anyway, so I think it doesn't matter much. And it still provides more info than the current nothingness.

NiLSPACE:
No objection here.

PR #3548, opened 2017-01-18
Webadmin: Added a missing CS lock.

madmaxoft:
This fixes an assert when accessing the login-page and the plugin-failure pages in the webadmin.

Fixes #3547

PR #3551, opened 2017-01-18
WebAdmin: Renamed two neighboring locks for better clarity / style.

madmaxoft:
Fixed a naming issue introduced in #3548

### DIFF
--- a/src/WebAdmin.cpp
+++ b/src/WebAdmin.cpp
@@ -340,8 +340,8 @@ void cWebAdmin::HandleWebadminRequest(cHTTPServerConnection & a_Connection, cHTT
 	// Try to get the template from the Lua template script
 	if (ShouldWrapInTemplate)
 	{
-		cCSLock Lock(m_CS);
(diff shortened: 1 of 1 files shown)


src/WebAdmin.cpp
```diff
@@ -340,8 +340,8 @@ void cWebAdmin::HandleWebadminRequest(cHTTPServerConnection & a_Connection, cHTT
 	// Try to get the template from the Lua template script
 	if (ShouldWrapInTemplate)
 	{
-		cCSLock Lock(m_CS);
-		cLuaState::cLock lock(m_TemplateScript);
+		cCSLock lockSelf(m_CS);
+		cLuaState::cLock lockTemplate(m_TemplateScript);
```
Seadragon91: l -> L?

madmaxoft: If you say so...

madmaxoft: I'm more inclined for javaCase, the thing I'll wage war against is silly_unreadable_t snake_case.

PR #3557, opened 2017-01-19
Improved easyinstall.sh to support macOS and FreeBSD

Bond-009:


easyinstall.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
madmaxoft: Does this really require Bash? I believe the old version was universal, and if possible, we'd like to keep that

madmaxoft:
@sphinxc0re could you verify that this new script still runs on Linux? Without requiring advanced shell such as bash?

madmaxoft:
(I have update the branch, which means this now has to be squash-merged, rather than regular-merged.)

falconxunit:
dd 7sw>490wisisisizjdu3 we as s7x9u2d88s i3 WA s7

⁣Envoyé par TypeApp ​

Le 13 févr. 2017 2:00 PM, à 2:00 PM, Mattes D <notifications@github.com> a écrit:

falconxunit:
zzek from dkc9d I the mse more e I

⁣Envoyé par TypeApp ​

Le 13 févr. 2017 3:10 PM, à 3:10 PM, Julian Laubstein <notifications@github.com> a écrit:

PR #3560, opened 2017-01-21
Changed big flowers so they can only be placed on dirt

### LINKED ISSUE #3559: Big flowers can be placed on anything
Client version:   1.10.2
Server OS:        Linux
Commit id:        97980d11038e57faa905a27977545fb77ff2c5bd

### Expected behavior
Big flowers should only be able to be placed on dirt.

### Actual behavior
Big flowers can be placed on any block.

Place a big flower (e.g. tall grass or a peony) on anything.

Altenius:
Fixes #3559

### DIFF
--- a/src/Blocks/BlockBigFlower.h
+++ b/src/Blocks/BlockBigFlower.h
@@ -75,12 +75,22 @@ public:
 		}
 	}
 
+	bool IsMetaTopPart(NIBBLETYPE a_Meta)
+	{
+		return (a_Meta & 0x08) != 0;
+	}
+
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, int a_RelX, int a_RelY, int a_RelZ, const cChunk & a_Chunk) override
 	{
+		if (a_RelY <= 0)
+		{
+			return false;
+		}
 		BLOCKTYPE BlockType;
 		NIBBLETYPE BlockMeta;
 		a_Chunk.GetBlockTypeMeta(a_RelX, a_RelY - 1, a_RelZ, BlockType, BlockMeta);
-		return ((a_RelY > 0) && (IsBlockTypeOfDirt(BlockType) || ((BlockType == E_BLOCK_BIG_FLOWER) && !(BlockMeta & 0x08))) && (a_RelY < cChunkDef::Height - 1) && ((a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_AIR) || (a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_BIG_FLOWER)));
+
+		return IsBlockTypeOfDirt(BlockType) || ((BlockType == E_BLOCK_BIG_FLOWER) && !IsMetaTopPart(BlockMeta));
 	}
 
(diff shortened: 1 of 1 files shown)


Altenius:
Looks like Travis CI failed because apt-get install couldn't find a package.

src/Blocks/BlockBigFlower.h
```diff
@@ -77,7 +77,10 @@ class cBlockBigFlowerHandler :
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, int a_RelX, int a_RelY, int a_RelZ, const cChunk & a_Chunk) override
 	{
-		return ((a_RelY > 0) && (a_Chunk.GetBlock(a_RelX, a_RelY - 1, a_RelZ) != E_BLOCK_AIR) && (a_RelY < cChunkDef::Height - 1) && ((a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_AIR) || (a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_BIG_FLOWER)));
+		BLOCKTYPE BlockType;
+		NIBBLETYPE BlockMeta;
+		a_Chunk.GetBlockTypeMeta(a_RelX, a_RelY - 1, a_RelZ, BlockType, BlockMeta);
```
madmaxoft: This will fail if `a_RelY` < 1

Overall, the condition below is really complicated, refactor it using standalone `if` statements to make it more readable.

Altenius: What do you mean by standalone if statements? Why not break it into multiple lines?

Altenius:
I fixed the `a_RelY == 0` code and fixed another problem where tall flowers can only be placed if there's air above them, which isn't the case in vanilla.

src/Blocks/BlockBigFlower.h
```diff
@@ -77,7 +77,15 @@ class cBlockBigFlowerHandler :
...
-		return ((a_RelY > 0) && (a_Chunk.GetBlock(a_RelX, a_RelY - 1, a_RelZ) != E_BLOCK_AIR) && (a_RelY < cChunkDef::Height - 1) && ((a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_AIR) || (a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ) == E_BLOCK_BIG_FLOWER)));
+		if (a_RelY <= 0)
+		{
+			return false;
+		}
+		BLOCKTYPE BlockType;
+		NIBBLETYPE BlockMeta;
+		a_Chunk.GetBlockTypeMeta(a_RelX, a_RelY - 1, a_RelZ, BlockType, BlockMeta);
+
+		return IsBlockTypeOfDirt(BlockType) || ((BlockType == E_BLOCK_BIG_FLOWER) && !(BlockMeta & 0x08));
```
madmaxoft: Style: Replace `!(BlockMeta & 0x08)` with an explicit comparison - we only allow bool values in conditions, the result of that expression is NOT a bool value.

Perhaps even better, replace it with a call to new function `IsMetaTopPart` - adds proper meaning and is much more maintainable.

madmaxoft:
Wait, does this mean that a tall flower can drill through any block above it? So it would be possible to break above the Nether in vanilla?

QuImUfu:
No, in vanilla you can't place duble plants, if there is something else than air in the upper block.

Altenius:
What I meant was flowers can be placed if there's air above the top part.

Issue #2858, opened 2016-01-11
Mobs: monster spawn issues


jammet:
As of about a week ago, I have finally started to enable monster mobs on my server, but due to these issues, I'm only a select few:

* Currently, if monsters are enabled, they spawn only on the surface, never in caves.

* The type of monsters seems completely random, and rare ones show up just as often as regular ones. Endermen aren't supposed to be all around - on here I get them just as often as regular zombies. Because there are still more creepers than what is typical in vanilla, I disabled those as well.

* Spiders attack you even during the day.

* Monsters don't seem to have attack animations. You just take damage.

* Monsters don't make some of their typical noises.

I'm not sure if listing all these as one issue is an okay thing to do.

LogicParrot:
It's best to not create mega-issues like this one. The spawning part is a duplicate of #2159.

jammet:
My apologies, I wasn't sure what to do -- split it up or maybe not do it at all. ^^' Thanks!

PR #3563, opened 2017-01-24
made survival the default gamemode

### LINKED ISSUE #2600: Make 'Survival' the default gamemode
We can keep 'Creative' as gamemode for debugging

Alcamech:
Closes #2600

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -391,7 +391,7 @@ void cWorld::Start(cDeadlockDetect & a_DeadlockDetect)
 	m_SpawnX = 0;
 	m_SpawnY = cChunkDef::Height;
 	m_SpawnZ = 0;
-	m_GameMode = eGameMode_Creative;
+	m_GameMode = eGameMode_Survival;
 
 	cIniFile IniFile;
 	if (!IniFile.ReadFile(m_IniFileName))


sphinxc0re:
Thank you soooooo much for this change ❤️

LogicParrot:
I don't think this is a good idea. Survival is less complete than creative. At the very least, mobs and mob spawning should be fixed first.

Alcamech:
Under the issue you commented that mobs no longer spawned too close to players. I thought that was the issue preventing this from being done. Mind filling me in on what else is wrong with the mob spawning ?

sphinxc0re:
The default gamemode should be survival so you can start playing directly after starting the server. Having sane defaults is always a good idea. Having creative mode in debug builds by default would be a good idea though. It's easier to test that way.

LogicParrot:
> Mind filling me in on what else is wrong with the mob spawning ?

Mobs no longer spawn too close to players, but tuning is needed for the number of mobs that spawn, and the mixture of types that spawn, and most mobtypes need to be implemented. Also, #2858 claims mobs do not spawn in caves, though I haven't tested that yet.

Also, pushing is broken, so it is currently disabled. (Zombie hits or arrows do not push the player). There are probably several more issues that I missed.

> I thought that was the issue preventing this from being done.

Spawning near players was a major issue and there was no way I'd agree to having survival as the default prior to fixing that. The remaining issues are less severe and I'll leave the choice to the other devs.

sphinxc0re:
The change is good anyway. So I'd say merge. People who want to play on the server want to play on survival regardless of feature completeness as stated here #2600

sphinxc0re:
@madmaxoft any thoughts? @Alcamech please rebase and squash the commits onto the current master

madmaxoft:
Don't care; no need to squash / rebase.

PR #3564, opened 2017-01-27
ApiDump: Added merge code to .luacheckrc

Seadragon91:
The merge code will be added into the file `.luacheckrc`, that is created by APIDump.

### DIFF
--- a/Server/Plugins/APIDump/main_APIDump.lua
+++ b/Server/Plugins/APIDump/main_APIDump.lua
@@ -1696,7 +1696,53 @@ globals =
 		end
 	end
 
-	file:write("}\n")
+	file:write("}\n\n")
+
+	-- Add merge code
+	file:write([[
+-- ## Main ##
+
+function WriteTable(a_File, a_TableName)
+	a_File:write(a_TableName, " = \n{\n")
+	for _, Entry in ipairs(_G[a_TableName]) do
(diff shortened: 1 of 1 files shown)


Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -1696,7 +1696,53 @@ globals =
...
+function WriteTable(a_File, a_TableName)
+	a_File:write(a_TableName, " = \n{\n")
+	for _, Entry in ipairs(_G[a_TableName]) do
+		a_File:write("\t\"", Entry, "\",\n")
+	end
+	a_File:write("}\n\n")
+end
+
+-- Load plugins's luacheck
+local FilePluginLuacheck = assert(loadfile(".plugin_luacheck"))
```
madmaxoft: Oh COME ON! Really, assert? What if a plugin doesn't need any config?

Seadragon91: Ops, sorry. Will correct that,

sphinxc0re: And why aren't you loading the normal `.luacheck` file?

Seadragon91: I load a plugin specific one, the merge code is for merging both options / tables into one file.

Not sure what you really mean.

PR #3565, opened 2017-01-27
APIDump: Corrected luacheck file to load a plugin's luacheck if exists

Seadragon91:
@madmaxoft Are there any other changes that I should include?

### DIFF
--- a/Server/Plugins/APIDump/main_APIDump.lua
+++ b/Server/Plugins/APIDump/main_APIDump.lua
@@ -1702,45 +1702,26 @@ globals =
 	file:write([[
 -- ## Main ##
 
-function WriteTable(a_File, a_TableName)
-	a_File:write(a_TableName, " = \n{\n")
-	for _, Entry in ipairs(_G[a_TableName]) do
-		a_File:write("\t\"", Entry, "\",\n")
-	end
-	a_File:write("}\n\n")
-end
-
 -- Load plugins's luacheck
-local FilePluginLuacheck = assert(loadfile(".plugin_luacheck"))
-local PluginLuacheck = {}
-setfenv(FilePluginLuacheck, PluginLuacheck)
-FilePluginLuacheck()
+local FilePluginLuacheck = loadfile(".luacheckrc_plugin")
 
-for Option, Value in pairs(PluginLuacheck) do
-	if (type(Value) == "table") and not(_G[Option] == nil) then
-		-- Merge tables together
-		for _ , Entry in ipairs(Value) do
-			table.insert(_G[Option], Entry)
+if FilePluginLuacheck ~= nil then
+	local PluginLuacheck = {}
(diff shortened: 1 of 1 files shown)


Seadragon91:
Do not merge yet. The current problems are:
1) The file luacheck_cuberite cant not be used directly
2) Unecessary recreate of a new file, if `.plugin_luacheck` not exists

Will look to improve that.

Seadragon91:
Okay my new approach:
File `.plugin_luacheck` exists:
- Run with `lua luacheck_cuberite merge` -> merges both files and creates file `.luacheckrc`

File `.plugin_luacheck` doesn't exists:
- Move file `luacheck_cuberite` to `.luacheckrc`

And at last run `luacheck . --codes`.

To avoid a syntax problem, the file `luacheck_cuberite` needs to bail out before the runable code is called.
This can be done by adding this check:
```
-- Check if the first arg is merge, if not bail out
if (arg[1] ~= "merge") then
	return
end
```

This solves both problems.

Seadragon91:
@madmaxoft If this idea is good for you, then I can add the additional code to avoid the syntax problem.

madmaxoft:
I don't see why exactly the code as it is now cannot be used directly by LuaCheck? I believe it should work - LuaCheck loads this file as config, and this file loads the plugin-specific file, if it exists, and merges it into the Cuberite-specific config., with the plugin's config taking precedence in conflicts. That seems like a workable solution.

madmaxoft:
Oh, right, didn't see the bottom (unchanged) part. I'd remove that and make this file directly usable as `.luacheckrc`

Seadragon91:
Okay that's really nice. I didn't thought about that.
One problem is that the tables and options are not in `_G`:
- lua 5.1 in `getfenv(1)`
- lua 5.2 in `_ENV`

Is support for lua 5.1 enough?

madmaxoft:
Since Cuberite itself uses Lua 5.1, I think supporting only that version is okay. And I suppose it could be made multi-version later on - just like CuberitePluginChecker supports multiple Lua versions.

madmaxoft:
I've just tried the idea myself, removed the file-writing from the bottom of the generated file and used it directly as `.luacheckrc` and it seems to work for me.

Seadragon91:
Yes it works really good :).

Seadragon91:
Forgot to rebase...

madmaxoft:
Now one more change I'd like you to consider: How about naming the file `.luacheckrc_plugin`, so that it's closer to what people might expect for LuaCheck.

madmaxoft:
No need to rebase, GH now supports rebase-merge :)

Seadragon91:
That's really good the github rebase and merge :)

madmaxoft:
Looks good on my side, merge if no objections from anyone else.

PR #3569, opened 2017-02-12
PieceGenerator: Added rotation-aware vertical connectors.

madmaxoft:
These allow vertical connections that connect through even-sized logical connections (2 blocks wide tree trunk etc.), as well as requiring specific rotations for vertical connections.

Also in this changeset is the ability to load textual connector descriptions in addition to numeric, so that a `.cubeset` file can specify `Direction = "x+"` instead of `Direction = 5`


src/Generating/PieceGenerator.cpp
```diff
@@ -181,6 +181,247 @@ cPiece::cConnector::cConnector(const Vector3i & a_Pos, int a_Type, eBlockFace a_
...
+}
+
+
+
+
+
+bool cPiece::cConnector::StringToDirection(const AString & a_Value, eDirection & a_Out)
+{
+	// First try converting as a number:
+	auto dirInt = atoi(a_Value.c_str());
```
NiLSPACE: Didn't we have a special function to convert a string to a number?

madmaxoft: Right. Will fix.

NiLSPACE: https://github.com/cuberite/cuberite/blob/b3b723b4531a3781ab847a2652cf192303e78941/src/StringUtils.h#L153

src/StringUtils.h
```diff
@@ -223,7 +223,7 @@ bool StringToInteger(const AString & a_str, T & a_Num)
 /** Returns an integer from a key-value string map.
```
NiLSPACE: The comment isn't valid anymore.

NiLSPACE:
Looks alright to me. I only looked briefly at the logic though.

src/StringUtils.h
madmaxoft: I think it still is. It means `an integer` as any integral type, not as `int` specifically.
Would changing it to `a number` or `an integral number` be better?

NiLSPACE: `integral number` sounds good.

madmaxoft: I think I've got it:
```cpp
/** Returns a number (of type T) from a key-value string map.
Returns a_Default if the key is not present or the value is not a number representable as T. */
```

NiLSPACE: Is the output always floored, or can it also have a decimal value?

madmaxoft: Only digits and optional `-` and `+` are accepted (see `StringToInteger` above that it uses). If the input contains anything else, the parsing fails and `a_Default` is returned. That's why it has `Integer` in the name - it only accepts integers.

PR #3575, opened 2017-02-21
Add 1.11.1/1.11.2 protocol

mathiascode:
No new features, but 1.11.1 and 1.11.2 clients can join.

### DIFF
--- a/src/Protocol/Protocol_1_11.cpp
+++ b/src/Protocol/Protocol_1_11.cpp
@@ -36,7 +36,7 @@ Implements the 1.11 protocol classes:
 
 
 
-cProtocol_1_11_0::cProtocol_1_11_0(cClientHandle * a_Client, const AString &a_ServerAddress, UInt16 a_ServerPort, UInt32 a_State) :
+cProtocol_1_11_0::cProtocol_1_11_0(cClientHandle * a_Client, const AString & a_ServerAddress, UInt16 a_ServerPort, UInt32 a_State) :
 	super(a_Client, a_ServerAddress, a_ServerPort, a_State)
 {
 }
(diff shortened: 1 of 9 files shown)


src/Protocol/Protocol_1_11.cpp
```diff
@@ -187,3 +187,51 @@ void cProtocol_1_11_0::HandlePacketStatusRequest(cByteBuffer & a_ByteBuffer)
 
 
 
+cProtocol_1_11_1::cProtocol_1_11_1(cClientHandle * a_Client, const AString &a_ServerAddress, UInt16 a_ServerPort, UInt32 a_State) :
```
madmaxoft: Style: spaces around `&`

madmaxoft:
Looks alright, after the style issue is fixed, feel free to merge.

src/Protocol/Protocol_1_11.cpp
falconxunit: ok good show me

madmaxoft:
@falconxunit what are you on about? It's the second time you're babbling nonsense in our tracker.

mathiascode:
Done. Some packet IDs and code was from the 1.8 protocol, which caused client disconnection when mobs spawned or the player tried to collect pickups. I've corrected these.

Every mob works properly now, except for the skeleton, which is invisible. I guess we can fix this later.

madmaxoft:
Does this mean we have a more-or-less functional 1.11 support? Good job!

Pokechu22:
> Every mob works properly now, except for the skeleton, which is invisible. I guess we can fix this later.

Do wither skeletons work?  Also, do zombie villagers work?  Finally, do the horse variants work?  Since the IDs of these entities were split in 1.11, I'm not sure if they'd be functional.

Still, good work!

madmaxoft:
Let's not get carried away in this PR. It is complete now, anything else is a good candidate for another PR. Making further changes would have made it need a new review, much more in-depth.

mathiascode:
Wither skeletons, zombie villagers, zombie horses and skeleton horses don't spawn at the moment. Further works needs to be done to make these functional again.

Pokechu22:
Yep, I expected that (and it's perfectly reasonable).  Just wanted to make sure that _all_ of them were broken, not just one of them (which would be very strange).

PR #3576, opened 2017-02-21
BlockCrops: Fixed comments and reformatted control flow.

madmaxoft:
Found a leftover bad comment, fixed it, together with a minor control flow reformat.

### DIFF
--- a/src/Blocks/BlockCrops.h
+++ b/src/Blocks/BlockCrops.h
@@ -14,61 +14,23 @@ class cBlockCropsHandler :
 	public cBlockPlant
 {
 	typedef cBlockPlant Super;
+
 public:
-	cBlockCropsHandler(BLOCKTYPE a_BlockType)
-		: Super(a_BlockType, true)
+
+	cBlockCropsHandler(BLOCKTYPE a_BlockType):
+		Super(a_BlockType, true)
 	{
 	}
+
+
 
 	virtual void ConvertToPickups(cItems & a_Pickups, NIBBLETYPE a_Meta) override
 	{
 		cFastRandom rand;
 
-		// Beetroots have three stages before fully grown
-		if (a_Meta >= RipeMeta)
+		// If not fully grown, drop the "seed" of whatever is growing:
+		if (a_Meta < RipeMeta)
 		{
-			// Is fully grown, drop the entire produce:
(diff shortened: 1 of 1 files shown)


src/Blocks/BlockCrops.h
```diff
@@ -81,9 +43,51 @@ class cBlockCropsHandler :
 					break;
 				}
 			}
+			return;
 		}
+
+		// Fully grown, drop the produce:
```
sphinxc0re: s/produce/product/

madmaxoft: I think `produce` is the correct word, it does mean "fruit, vegetables etc. that have grown and are collected"; `product` is too generic. Of course I could be wrong, any English natives care to chime in? @bearbin @tigerw ?

Pokechu22: `produce` is correct, but it sounds weird because it's an uncountable noun (like `sheep`).  `product`  could work, but it's more general ([Wiktionary](https://en.wiktionary.org/wiki/produce#Noun) lists it as a hypernym).

Perhaps `crop's produce` would be better?

madmaxoft: `sheep` is a bad example for an uncountable noun, since it is actually really weird that it is uncountable in English. `flour` would have been a better one, I guess.

If I compare `drop the produce` with `drop the flour`, I don't feel much difference (linguistically), you can drop anything, countable or not. And making it `drop the crop's produce` doesn't seem to improve on it, rather, it's like duplicating the word.

So now we have:
 - `drop the produce`
 - `drop the product`
 - `drop the crop's produce`

Which one will it be? Or yet another one?

Now this entire PR will get stuck on a trivial language matter... :)

sphinxc0re: `drop the crop's produce` seems good to me

madmaxoft: Oh well, you talked me into it :)

sphinxc0re:
So squash-merge does rebasing as well?

madmaxoft:
Yes. Squash-merge converts all changes in the PR into a single commit on top of current master, optionally changing the commit message. So it's a squash-rebase-amend thingy.

PR #3577, opened 2017-02-21
Corrected particles

mathiascode:
I wrote the wrong IDs in my previous PR, and included uppercase letters, causing the particles to become unknown to the server.

### DIFF
--- a/src/Protocol/Protocol_1_9.cpp
+++ b/src/Protocol/Protocol_1_9.cpp
@@ -1724,9 +1724,8 @@ bool cProtocol_1_9_0::CompressPacket(const AString & a_Packet, AString & a_Compr
 
 int cProtocol_1_9_0::GetParticleID(const AString & a_ParticleName)
 {
-	static bool IsInitialized = false;
 	static std::map<AString, int> ParticleMap;
-	if (!IsInitialized)
+	if (ParticleMap.empty())
 	{
 		// Initialize the ParticleMap:
 		ParticleMap["explode"]          = 0;
@@ -1750,7 +1749,7 @@ int cProtocol_1_9_0::GetParticleID(const AString & a_ParticleName)
 		ParticleMap["dripwater"]        = 18;
 		ParticleMap["driplava"]         = 19;
 		ParticleMap["angryvillager"]    = 20;
-		ParticleMap["happyVillager"]    = 21;
+		ParticleMap["happyvillager"]    = 21;
 		ParticleMap["townaura"]         = 22;
 		ParticleMap["note"]             = 23;
 		ParticleMap["portal"]           = 24;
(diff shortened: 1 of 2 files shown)


madmaxoft:
Is it just me, or is the `IsInitialized` for the particle map never set to true, causing initialization over and over again? Not that it's this PR's fault, just an observation.

madmaxoft:
If I were to write the code, I'd use `if (ParticleMap.empty())` and get rid of the `IsInitialized` altogether.

sphinxc0re:
@madmaxoft Should we open an issue for that or can this be included in this PR?

madmaxoft:
I think this PR would be a good place to fix it - it is a Particle fix, after all :)

@mathiascode can you replace the code at the top of the function (L1708 - L1710 in 1.8, similarly in all other protocols) with this:
```cpp
static std::map<AString, int> ParticleMap;
if (ParticleMap.empty())
```

mathiascode:
Done.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -1724,9 +1724,8 @@ bool cProtocol_1_9_0::CompressPacket(const AString & a_Packet, AString & a_Compr
 
 int cProtocol_1_9_0::GetParticleID(const AString & a_ParticleName)
 {
-	static bool IsInitialized = false;
 	static std::map<AString, int> ParticleMap;
-	if (!IsInitialized)
```
sphinxc0re: Is this `IsInitialized` variable set anywhere else? Did you remove all its occurrences?

mathiascode: It also exists in WebAdmin.cpp, line 441-443

sphinxc0re: That is kinda confusing. I'm sorry @madmaxoft could you explain to me how this variable is used?

madmaxoft: A `static` variable is somewhere between a local variable (it is only accessible in the block where it is declared) and a global variable (it has one value, which stays between multiple entrances into that block). Basically, imagine this as being declared as a global variable `bool cProtocol_1_9_0::GetParticleID::Scope1::IsInitialized`. So it is only accessible in this `GetParticleID()` function, but it keeps its value from one function call to another.

Here's some code to illustrate:
```cpp
void printCounter(void)
{
  static int counter = 0;
  printf("%d\n", counter);
  counter += 1;
}

void printSequence(void)
{
  printCounter();  // Prints "0", "counter" is now 1
  printCounter();  // Prints "1", "counter" is now 2
  printCounter();  // Prints "2", "counter" is now 3
  counter = 4;  // compile error, "counter" not accessible from here
}
```

The primary purpose was supposed to be to make the `ParticleMap` initialization happen only once. The same effect can be obtained by asking the map if it's empty.

madmaxoft: So this variable has nothing to do with the one in WebAdmin :)

PR #3578, opened 2017-02-22
Add support for alternate config file locations

GarethNelson:
Simple little patch, adds ability to select a different config file from the command line.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -51,6 +51,7 @@ WebFreak001
 worktycho
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+Gareth Nelson
 
 Please add yourself to this list if you contribute to the official Cuberite project
 and license your work under the Apache License 2.0
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,6 +10,7 @@ derouinw
 Diusrex
 Duralex
 FakeTruth (founder)
+Gareth Nelson
 HaoTNN
 Howaner
 jan64
@@ -51,7 +52,6 @@ WebFreak001
 worktycho
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
-Gareth Nelson
 
 Please add yourself to this list if you contribute to the official Cuberite project
 and license your work under the Apache License 2.0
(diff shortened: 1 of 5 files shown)


madmaxoft:
This will not work yet, there's another mention of `"settings.ini"` in `cPluginManager::ReloadPluginsNow()`

src/Root.cpp
```diff
@@ -141,8 +141,14 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 
 	LOG("Reading server config...");
 
+	std::string conf_file = "settings.ini";
+	if (a_OverridesRepo->HasValue("Server","ConfigFile"))
+	{
+	conf_file = a_OverridesRepo->GetValue("Server","ConfigFile");
```
NiLSPACE: The indentation is wrong here.

GarethNelson: Can you elaborate on the indentation?
Also, @madmaxoft i'm working on PluginManager

GarethNelson:
@NiLSPACE Please check src/Root.cpp again - I believe the indentation is now correct style

src/Root.cpp
NiLSPACE: You should add an extra tab for `conf_file = a_OverridesRepo..`

With the brace you're opening a new block, so by adding a new tab you're making the piece of code much more readable. This is especially more noticable if there is more code inside a block,

NiLSPACE:
Yes, that's it :) Could you change over the `std::string` to `AString`?

GarethNelson:
Working on it, next commit

GarethNelson:
@NiLSPACE think this will be good once the CI tests pass?

src/Root.h
```diff
@@ -58,6 +58,8 @@ class cRoot
 
 	void Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo);
 
+	AString m_SettingsFile;
```
madmaxoft: We usually put variables first, then functions. So move this up.

Also, it's not a "file", it's a "filename".

madmaxoft: And add a doxy-comment (`/** ... */`) explaining what this is, exactly.

GarethNelson: See commit 25fe4a5 below

GarethNelson:
@madmaxoft please review changes (and note for some reason travis is not passing despite showing as passed over on the travis site)

CONTRIBUTORS
```diff
@@ -51,6 +51,7 @@ WebFreak001
 worktycho
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+Gareth Nelson
```
madmaxoft: Add yourself to the alpha-sorted position in the list.

madmaxoft:
One more change and it should be good.

GarethNelson:
@madmaxoft Done

madmaxoft:
Thanks

PR #3581, opened 2017-02-23
Cleaned up Server directory

mathiascode:
I removed some files from the homepage builds, since they added unnecessary bloat.


sphinxc0re:
Why did you remove the favicon?

mathiascode:
It overwrites custom favicons when updating the server, and the Windows builds never had one anyway. The User's Manual contains information about the favicon.png file.

sphinxc0re:
@madmaxoft @mathiascode I think the idea isn't bad but this seems to be the wrong place for the change. What @mathiascode wants is to eliminate unused or confusing files from the distributed server directory. This is not wrong in fact I thought about making this change a few times. But  this should be done in the artifact creation section of the build script. Am I wrong @bearbin ?

mathiascode:
Different approach: everything that should be in the Server folder is there, the build server script removes unnecessary files.

https://github.com/cuberite/build-scripts/pull/1

sphinxc0re:
Should we close this then?

mathiascode:
I still did changes in this PR, including files that are bundled with the Windows builds.

mathiascode:
Anyone against moving third party licenses and debug scripts to the Server folder instead of Install?

madmaxoft:
I don't get it. First you want to clean up the Server folder, and then you put in files that are only needed when packaging the release into the folder. I think things are perfectly fine as they are now and don't need any of your changes. The Server folder contains stuff to run the server when debugging from the source, and the Install folder contains anything extra needed when packaging the executable.

PR #3584, opened 2017-02-25
Fixes for issues #902 and #2917

Bond-009:
Code from closed PR #3482 (not my code)

Issue #2917 fix: Mobs now take damage when standing on top of a cactus block.

Issue #902 fix: Zombie pigmen now spawn holding a golden sword.

### DIFF
--- a/src/Mobs/ZombiePigman.cpp
+++ b/src/Mobs/ZombiePigman.cpp
@@ -35,11 +35,17 @@ void cZombiePigman::GetDrops(cItems & a_Drops, cEntity * a_Killer)
 }
 
 
+
+
+
 void cZombiePigman::SpawnOn(cClientHandle & a_ClientHandle)
 {
 		super::SpawnOn(a_ClientHandle);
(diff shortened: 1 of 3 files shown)


src/Mobs/ZombiePigman.cpp
```diff
@@ -34,7 +35,11 @@ void cZombiePigman::GetDrops(cItems & a_Drops, cEntity * a_Killer)
 }
 
 
-
+void cZombiePigman::SpawnOn(cClientHandle & a_ClientHandle)
```
madmaxoft: Style: Please keep the 5 empty lines between top-level functions. It helps readability.

madmaxoft:
Just a style issue, other than that, good code.

(note to self): When merging, please specify the exact issues fixed by this, rather than their numbers.

sphinxc0re:
@madmaxoft this might be a good thing in general. Explaining what the merge of the changeset solves might result in a cleaner changelog. That way, we could easily see, what happend since the last X commits

madmaxoft:
@sphinxc0re We do that most of the time, this was one of the few exceptions that needed correcting.

@Bond-009 Thanks.

PR #3585, opened 2017-02-25
Renamed OSX to macOS and use curl to get EasyInstall script

Bond-009:
Use curl to get the EasyInstall script because the script itself uses curl.

I didn't change the cmd to get the compile.sh script because it didn't work in my testing.

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -4,7 +4,7 @@ COMPILING
 To compile Cuberite from source, you need the following set of software:
 
  * CMake
- * Platform-specific make tool \(windows would be MSVC, linux/macos GNU make, etc.\)
+ * Platform-specific make tool \(Windows would be MSVC, Linux/macOS GNU make, etc.\)
  * C compiler
  * modern C++ compiler and linker
 
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -4,7 +4,7 @@ COMPILING
 To compile Cuberite from source, you need the following set of software:
 
  * CMake
- * Platform-specific make tool \(windows would be MSVC, linux/osx GNU make, etc.\)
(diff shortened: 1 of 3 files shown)


COMPILING.md
```diff
@@ -4,7 +4,7 @@ COMPILING
 To compile Cuberite from source, you need the following set of software:
 
  * CMake
- * Platform-specific make tool \(windows would be MSVC, linux/osx GNU make, etc.\)
+ * Platform-specific make tool \(windows would be MSVC, linux/macos GNU make, etc.\)
```
sphinxc0re: It's `macOS` and should be so on every page mentioning this.

README.md
```diff
@@ -22,7 +22,7 @@ There are several ways to obtain Cuberite.
 ##### The EasyInstall script
 This script will download the correct binary from the project site.
 
-    sh -c "$(wget -O - https://download.cuberite.org)"
+    curl -sSfL https://download.cuberite.org | sh
```
sphinxc0re: What is the difference between those two?

Bond-009: Use curl because the script itself uses it and not wget

COMPILING.md
Bond-009: That is what I did

Pokechu22: I think @sphinxc0re is referring to the capitalization: `macOS` instead of `macos`.

Bond-009: Oh ok, but Linux and Windows where also lower case

README.md
sphinxc0re: :+1:

bearbin:
Looks good, merging.

PR #3589, opened 2017-02-26
Dead bushes drop sticks

Bond-009:
Dead bushes drop sticks or a dead bush when using shears


src/Blocks/BlockHandler.cpp
```diff
@@ -473,12 +473,14 @@ void cBlockHandler::DropBlock(cChunkInterface & a_ChunkInterface, cWorldInterfac
...
 				case E_BLOCK_CROPS:
 				case E_BLOCK_DARK_OAK_DOOR:
+				case E_BLOCK_DEAD_BUSH:
 				case E_BLOCK_DOUBLE_RED_SANDSTONE_SLAB:
 				case E_BLOCK_DOUBLE_STONE_SLAB:
 				case E_BLOCK_DOUBLE_WOODEN_SLAB:
 				case E_BLOCK_FIRE:
 				case E_BLOCK_FARMLAND:
 				case E_BLOCK_FLOWER_POT:
+				case E_BLOCK_GRASS:
```
madmaxoft: I believe this is incorrect, this is the grass block (dirt block with grass texture on the top side). Did you mean `E_BLOCK_TALL_GRASS`?

Bond-009: You are right

madmaxoft:
Now these changes are controversial and I'm not sure I want to allow them. By changing the blocktype names, you're breaking existing plugins that use these names. For this reason, we've had a policy of not removing names, and rather creating aliases from the old name to the new one.

I think such a change shouldn't be done in this PR, but rather in a separate one, if at all - so that the deadbushes change can be merged and the rest can be discussed.

Bond-009:
Yeah I know, I was stupid add committed to the wrong branch

madmaxoft:
You could force-push the previous commit to be the top of the branch, or you could cherry-pick the original commit into a new branch and start a new PR. Or I could cherry-pick and squash your commits directly onto master and close this PR. Your choice.

Bond-009:
FFS I suck at git

PR #3591, opened 2017-02-26
Deadbushes drop sticks

Bond-009:


src/Blocks/BlockDeadBush.h
```diff
@@ -42,6 +42,21 @@ class cBlockDeadBushHandler :
 		UNUSED(a_Meta);
 		return 0;
 	}
+
+	virtual void OnDestroyedByPlayer(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cPlayer * a_Player, int a_BlockX, int a_BlockY, int a_BlockZ) override
```
Pokechu22: Is this the best method to override?  [`BlockTallGrass` uses `DropBlock`](https://github.com/cuberite/cuberite/blob/5effe2cc7916e067e6f6579a73c1101a4e58dd63/src/Blocks/BlockTallGrass.h#L36).  That would also allow a plugin callback:

```a_BlockPluginInterface.CallHookBlockToPickups(a_Digger, a_BlockX, a_BlockY, a_BlockZ, m_BlockType, Meta, Drops);```

madmaxoft: We have too many similar methods to override, and then it leads to situations like this. It seems like the `DropBlock` method would be preferred for this indeed.

Bond-009: I will change it when I have time.

Bond-009:
Do I need to change anything else?

madmaxoft:
Looks good. Sorry for the late reply.

PR #3601, opened 2017-02-28
Add 1.11 entity metadata

### LINKED ISSUE #3597: 1.11 - invisible skeletons and dark arrows
Client version:   1.11 and above, Vanilla, OS - Windows
Server OS:        Linux
Only shadows are visible
P.S.
Oh, sorry - I've just read "Add 1.11.1/1.11.2 protocol" you already know this xD

Pokechu22:
This change adjusts the IDs to match the skeleton, zombie, and horse
changes made in 1.11.  **However, it doesn't do so completely**; it doesn't
handle the split entity IDs.  This means that zombie villagers, wither
skeletons, and horse variants all use their default versions (zombies,
skeletons, and regular horses).  But it should fix #3597.

Additionally, this corrects a mistake in 1.10's metadata writer, where
the ageable baby ID was used instead of the zombie baby ID.  Those
constants have the same value, so it wasn't a programatic issue, but it
was a semantic problem.

### DIFF
--- a/src/Protocol/CMakeLists.txt
+++ b/src/Protocol/CMakeLists.txt
@@ -28,9 +28,10 @@ SET (HDRS
 )
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-	set_source_files_properties(Protocol_1_9.cpp  PROPERTIES COMPILE_FLAGS "-Wno-error=switch-enum -Wno-error=switch")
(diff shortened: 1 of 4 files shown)


Pokechu22:
I should note that I haven't tested this ingame yet (don't have access to a client).  But it _should_ work.

madmaxoft:
CircleCI is acting up, the build succeeded, but wasn't properly reported in GH.

PR #3616, opened 2017-03-04
Added static flag and removed static info from text.

Seadragon91:


Seadragon91:
Do not merge yet. Found one more.

Seadragon91:
The class `cBoundingBox` has a constructor and a few function that has incorrect params. Will correct that in own pull request.

Seadragon91:
This one can be merged, if it's okay.

PR #3619, opened 2017-03-11
Removed development plugins from settings file

mathiascode:
We should probably remove (some of) the development plugins from the downloads on the homepage, to prevent cases where new users enable them on production servers. This could probably be done by modifying the build scripts for Linux builds, not sure when it comes to Windows.

- APIDump
- Debuggers
- HookNotify
- NetworkTest
- TestLuaRocks


madmaxoft:
I think the plugins themselves should stay - they are there for newcoming plugin authors, to provide some initial examples. No need for them in the INI files, true.

Issue #3593, opened 2017-02-27
Server runs in a loop when the port is already bound


HelenaKitty:
Server OS:        Linux
Commit id:        1c31cb7eecc92d2fb218da8948e3201b618a46d2

### Expected behavior
The server should have shut down safely when it couldn't bind to the port.

### Actual behavior
Instead the server ran in a loop

### Steps to reproduce the behavior

Hold the port so nothing can bind to it, then start the Cuberite server and it should loop.

madmaxoft:
This should be pretty easy to fix - the server already shows an error message if it cannot bind to any port, so add a call to terminate the server next to the log message. Marking as Easy.

HelenaKitty:
Agreed!

PR #3620, opened 2017-03-12
Fixing the issue for #3593 due to restart loop

willi123yao:
I am quite new to developing this and I am not sure whether this will cause any other complications.
Checks whether the port binding was successful in Server.cpp


willi123yao:
Fixing some issue... pardon me

madmaxoft:
Already fixed in #3621

PR #3621, opened 2017-03-12
Fixed looping when unable to bind port (#3593)

johnsoch:


Seadragon91:
The close button of the console doesn't work.

madmaxoft:
@Seadragon91 what exactly do you mean? "Doesn't work" as in "is disabled, cannot press", or "is enabled, pressing it does nothing" or "is enabled, pressing it crashes"

In my opinion disabling that button was a mistake.

Seadragon91:
It's disabled. Adding this [code](https://github.com/cuberite/cuberite/blob/eb7d684aad150fefae720c040d37fa5913284c90/src/Root.cpp#L233), enables it again.

madmaxoft:
I don't see much relevance to this PR, though - if the port binding fails, the server stops anyway, so is there should be no need to re-enable it manually anyway. Or does it stay disabled even when the server terminates (if the console window stays open)? If so, then adding those three lines to this PR's `else` branch would make sense.

Seadragon91:
Yes the button gets disabled on startup and stays disabled if it can not bind to any port.

johnsoch:
Did the close button stop working after fixing the restart loop, or is this something that was always that way. It seems like enabling it should happen at the end of the `cRoot::Start` method, when `m_TerminateEventRaised` is true. A better option would be not having to disable it in the first place, but I don't have a Windows computer available right now to look at it further.

Seadragon91:
It was always the way. The code to enable the close button will be called if the server startups successfully, but if the server can not bind to any port the code will be never called.
Adding the 3 [lines](https://github.com/cuberite/cuberite/blob/eb7d684aad150fefae720c040d37fa5913284c90/src/Root.cpp#L233), fixes the problem.

madmaxoft:
@johnsoch The original idea was to disable the button only when the server was starting, and re-enable it once the server was fully up and could respond to the close event without losing data. Your addition skips the re-enabling, and the button seems stuck on disabled even after Cuberite terminates. Just copy and paste those three lines into your `else` branch so that we can get this PR merged, and we can discuss whether the button-disabling is a good thing elsewhere.

johnsoch:
I added the code for enabling the close button to the previously added `else` branch.

PR #3624, opened 2017-03-16
API: Corrected param types in cBoundingBox

Seadragon91:
Corrected the param types in APIDesc, hope that they are correct,


madmaxoft:
Good catch. These had been auto-generated by a mildly-intelligent script that decided that anything having a `Min` or `Max` should be a number.

PR #3628, opened 2017-03-17
Warning fix, remove dead code

editor700:
In GCC there is about 100 warnings due to unhandled enums in switch statements. To fix it, I added default statements in all switch.
Also there is unreachable code due to if condition always return true/false.


madmaxoft:
That is actually not helpful - it hides the errors, rather than make them stand out for a proper fix. We've already had such attempts and they were closed, too. Sorry.

madmaxoft:
As for the unreachable code, it is reachable in MSVC, so it's a portability thing.

PR #3629, opened 2017-03-17
Boat spawned, remove it from player's hand

Seadragon91:
This fixes a duplicate bug. Boat successfully placed, but not removed from player's hand.

### DIFF
--- a/src/Items/ItemBoat.h
+++ b/src/Items/ItemBoat.h
@@ -98,6 +98,12 @@ public:
 		cBoat * Boat = new cBoat(x + 0.5, y + 0.5, z + 0.5);
 		Boat->Initialize(*a_World);
 
+		// Remove boat from players hand
+		if (!a_Player->IsGameModeCreative())
+		{
+			a_Player->GetInventory().RemoveOneEquippedItem();
+		}
+
 		return true;
 	}
 } ;


src/Items/ItemBoat.h
```diff
@@ -98,6 +98,12 @@ class cItemBoatHandler :
 		cBoat * Boat = new cBoat(x + 0.5, y + 0.5, z + 0.5);
 		Boat->Initialize(*a_World);
 
+		// Remove boat from players hand
+		if (!a_Player->IsGameModeCreative())
```
madmaxoft: What if the player is in Spectator mode? Should they be allowed to place boats, and what to do then?

Pokechu22: You can't place blocks in spectator mode; you have no hotbar.  There [already seems to be a check](https://github.com/cuberite/cuberite/blob/01d9beee9ee60aa1ead52f7796c762a8372b9ca4/src/ClientHandle.cpp#L1482) that prevents using items in spectator mode.  Even if that check isn't complete, the check shouldn't be put into boats specifically; it should be in a more general location (as it already seems to be).

madmaxoft: Okay then, just asking (I have no idea how the Spectator mode works, haven't played Minecraft for ages, actually :)

Seadragon91: Maybe you mean adventure mode? Boat can be placed and destroyed.

PR #3630, opened 2017-03-18
Replace double grass and large fern with placed block

### LINKED ISSUE #3626: Visual bug on placing blocks
Client version:   1.8
Server OS:        Linux
Commit id:        9fa8477b73ce878c631ba875ec10746b4143ba75

### Expected behavior
The block should replace it and not create another one.

### Actual behavior
Only a single block should be placed, however there are 2 of them.

Take a dead bush or a double-tall grass and place a block on it. 

#### Notes:
However, it is actually only one block as when you break the additional block, the block over it will also break. If you break the original block it will be a single-hit-dig and there will be no drops.

####Screenshot:
_Before placing:_

_After placing:_

Seadragon91:
Fixes #3626.
If a block is placed on a big flower, the upper or lower part should be replaced by air. I am not sure, where I have to do that.

### DIFF
--- a/src/Blocks/BlockBigFlower.h
+++ b/src/Blocks/BlockBigFlower.h
@@ -18,6 +18,11 @@ public:
 	{
 	}
 
+	virtual bool DoesIgnoreBuildCollision(cPlayer * a_Player, NIBBLETYPE a_Meta) override
+	{
+		return (((a_Meta & E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) != 0) || (a_Meta & E_META_BIG_FLOWER_LARGE_FERN) != 0);
+	}
+
 	virtual void DropBlock(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cBlockPluginInterface & a_BlockPluginInterface, cEntity * a_Digger, int a_BlockX, int a_BlockY, int a_BlockZ, bool a_CanDrop) override
 	{
 		NIBBLETYPE Meta = a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ);
@@ -54,7 +59,13 @@ public:
 		NIBBLETYPE FlowerMeta = Meta & 0x7;
 		if (!a_Player->IsGameModeCreative())
 		{
-			if (((FlowerMeta == 2) || (FlowerMeta == 3)) && (a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS))
+			if (
+				(a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS) &&
+				(
+					(FlowerMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) ||
+					(FlowerMeta == E_META_BIG_FLOWER_LARGE_FERN)
+				)
+			)
 			{
 				MTRand r1;
 				if (r1.randInt(10) == 5)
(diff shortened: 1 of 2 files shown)


Seadragon91:
Now the upper or lower block is replaced by air. The block is still visible on the client, before it's replaced by air.

src/Blocks/BlockBigFlower.h
```diff
@@ -18,6 +18,11 @@ class cBlockBigFlowerHandler :
 	{
 	}
 
+	virtual bool DoesIgnoreBuildCollision(cPlayer * a_Player, NIBBLETYPE a_Meta) override
+	{
+		return (a_Meta & E_META_BIG_FLOWER_DOUBLE_TALL_GRASS || a_Meta & E_META_BIG_FLOWER_LARGE_FERN);
```
madmaxoft: I'd prefer the explicit longer condition:
`((a_Meta & E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) != 0) || ...`
After all, the whole codebase has the style that we don't use integers as booleans, we always use an explicit bool result.

src/Blocks/BlockBigFlower.h
```diff
@@ -54,7 +59,7 @@ class cBlockBigFlowerHandler :
 		NIBBLETYPE FlowerMeta = Meta & 0x7;
 		if (!a_Player->IsGameModeCreative())
 		{
-			if (((FlowerMeta == 2) || (FlowerMeta == 3)) && (a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS))
+			if (((FlowerMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) || (FlowerMeta == E_META_BIG_FLOWER_LARGE_FERN)) && (a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS))
```
madmaxoft: Split this into multiple lines

src/Blocks/BlockBigFlower.h
```diff
@@ -73,6 +78,11 @@ class cBlockBigFlowerHandler :
 				a_Player->UseEquippedItem();
 			}
 		}
+
+		if (((FlowerMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) || (FlowerMeta == E_META_BIG_FLOWER_LARGE_FERN)) && (a_Player->GetEquippedItem().m_ItemType != E_ITEM_SHEARS))
```
madmaxoft: Split into multiple lines

Seadragon91:
Okay for merge`?

src/Blocks/BlockBigFlower.h
```diff
@@ -54,7 +59,11 @@ class cBlockBigFlowerHandler :
 		NIBBLETYPE FlowerMeta = Meta & 0x7;
 		if (!a_Player->IsGameModeCreative())
 		{
-			if (((FlowerMeta == 2) || (FlowerMeta == 3)) && (a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS))
+			if (
+				((FlowerMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) ||
+				(FlowerMeta == E_META_BIG_FLOWER_LARGE_FERN)) &&
+				(a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS)
```
madmaxoft: This is poorly split, it hides the different levels of the condition operators. Consider instead:
```cpp
if (
	(a_Player->GetEquippedItem().m_ItemType == E_ITEM_SHEARS) &&
	(
			(FlowerMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) ||
			(FlowerMeta == E_META_BIG_FLOWER_LARGE_FERN)
	)
)
```

src/Blocks/BlockBigFlower.h
```diff
@@ -73,6 +82,15 @@ class cBlockBigFlowerHandler :
 				a_Player->UseEquippedItem();
 			}
 		}
+
+		if (
+			((FlowerMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS) ||
+			(FlowerMeta == E_META_BIG_FLOWER_LARGE_FERN)) &&
+			(a_Player->GetEquippedItem().m_ItemType != E_ITEM_SHEARS)
```
madmaxoft: Again, poor split.

Seadragon91:
Did the changes you requested, now okay?

PR #3631, opened 2017-03-18
Catch crash if any and print backtrace from every thread

Seadragon91:
This adds the feature to catch crashes in travis:
* Adds gdb to travis
* Starts cuberite under gdb and on exit of Cuberite gdb runs a few commands and then exits with cuberite's exit code.

Example of crash:
[All runs](https://travis-ci.org/Seadragon91/cuberite/builds/212259342)
[clang debug](https://travis-ci.org/Seadragon91/cuberite/jobs/212259346)

Example of no crash:
[All runs](https://travis-ci.org/Seadragon91/cuberite/)
[clang debug](https://travis-ci.org/Seadragon91/cuberite/jobs/212263506)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.8
+    - gdb
 
 script: ./CIbuild.sh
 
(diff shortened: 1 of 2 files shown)


Seadragon91:
Can I merge it?

madmaxoft:
Looks reasonable, but then, I'm not a Linux guru.

PR #3637, opened 2017-03-20
Added blocks that can be washed away by water

Bond-009:
http://minecraft.gamepedia.com/Breaking#Best_tools

Breaking double tall grass with water gives wrong drops (it should drop nothing)

### DIFF
--- a/src/Simulator/FluidSimulator.cpp
+++ b/src/Simulator/FluidSimulator.cpp
@@ -31,14 +31,22 @@ bool cFluidSimulator::CanWashAway(BLOCKTYPE a_BlockType)
 {
 	switch (a_BlockType)
 	{
+		case E_BLOCK_ACTIVATOR_RAIL:
+		case E_BLOCK_ACTIVE_COMPARATOR:
 		case E_BLOCK_BEETROOTS:
+		case E_BLOCK_BIG_FLOWER:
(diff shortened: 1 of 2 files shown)


Seadragon91:
In the link you posted, I see that all rails can be washed away.

Bond-009:
Fixed

Seadragon91:
What means `washed away`? Block is destroyed, when water is placed on it or it can be transported by water?

src/Simulator/FluidSimulator.cpp
```diff
@@ -31,26 +31,36 @@ bool cFluidSimulator::CanWashAway(BLOCKTYPE a_BlockType)
...
+	case E_BLOCK_RED_MUSHROOM:
+	case E_BLOCK_RED_ROSE:
+	case E_BLOCK_SNOW:
+	case E_BLOCK_SUGARCANE:
+	case E_BLOCK_TALL_GRASS:
+	case E_BLOCK_TORCH:
+	case E_BLOCK_TRIPWIRE_HOOK:
+	case E_BLOCK_TRIPWIRE:
+	case E_BLOCK_YELLOW_FLOWER:
+	{
```
Seadragon91: Sorry, please correct the format.

Bond-009: I already did

PR #3642, opened 2017-03-22
Piece generator updates

madmaxoft:
The piece generator has been slightly refactored (to allow more variants in the future, such as the ConnectAll generator for ocean monuments and mansions).

When merging, please rebase-merge (or regular-merge if possible), do not squash-merge.


madmaxoft:
Do not merge yet, it seems the "optimization" doesn't actually work (again).

madmaxoft:
This is now ready for review.

Note that I have refrained from doing any optimization, as every change had a very volatile performance impact. Even just rewriting the `cPieceGeneratorBFSTree`'s code to use C++11 range-based for-loops resulted in a 10% performance drop!

src/Generating/PieceGeneratorBFSTree.cpp
```diff
@@ -0,0 +1,345 @@
...
+	{
+		// No available connections, bail out
+		return false;
+	}
+	ASSERT(WeightTotal > 0);
+
+	// Choose a random connection from the list, based on the weights:
+	int rnd = (m_Noise.IntNoise3DInt(a_Connector.m_Pos.x, a_Connector.m_Pos.y, a_Connector.m_Pos.z) / 7) % WeightTotal;
+	size_t ChosenIndex = 0;
+	for (cConnections::const_iterator itr = Connections.begin(), end = Connections.end(); itr != end; ++itr, ++ChosenIndex)
```
worktycho: Why use this code over `std::discrete_distribution`?

src/Generating/PieceGeneratorBFSTree.cpp
```diff
@@ -0,0 +1,345 @@
...
+	rnd = rnd >> 16;
+
+	// Choose a random supported rotation:
+	int Rotations[4] = {0};
+	int NumRotations = 1;
+	for (size_t i = 1; i < ARRAYCOUNT(Rotations); i++)
+	{
+		if (StartingPiece->CanRotateCCW(static_cast<int>(i)))
+		{
+			Rotations[NumRotations] = static_cast<int>(i);
```
madmaxoft: No need to optimize this, it's called only once per the whole structure.

src/Generating/PieceGeneratorBFSTree.cpp
madmaxoft: I had no idea something like that existed. And anyway it doesn't seem to be constructible in iterations (pushing the individual weights one at a time) and even if it were, it would have resulted in an extra allocation for the weights buffer.

src/Generating/PieceGeneratorBFSTree.cpp
```diff
@@ -0,0 +1,345 @@
...
+					Hitbox.p1.x, Hitbox.p1.y, Hitbox.p1.z,
+					Hitbox.p2.x, Hitbox.p2.y, Hitbox.p2.z,
+					Hitbox.DifX() + 1, Hitbox.DifY() + 1, Hitbox.DifZ() + 1
+				);
+				DebugConnectorPool(ConnectorPool, NumProcessed + 1);
+				//*/
+			}
+		}
+		NumProcessed++;
+		if (NumProcessed > 1000)
```
madmaxoft: It is not expected to hit this condition even once, the condition is there only to avoid some extreme cases.

madmaxoft:
@worktycho is alive, YAY! :)

worktycho:
I'd like to play around with this at some point but looks like a good base implentation. Should be able to become active on the project again in a couple of months.

PR #3644, opened 2017-03-22
Choose # of threads based on system info (compile.sh)

### LINKED ISSUE #3014: compile.sh todo

Bond-009:
Fixes #3014

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -56,7 +56,7 @@ errorArguments ()
 	echo "options:"
 	echo "  -m  The compilation mode. Either \"Release\" or \"Debug\". Defaults to \"$DEFAULT_BUILDTYPE\""
 	echo '  -t  The number of threads to use for compiling'
-	echo '      If unspecified, a "smart guess" is attempted (Currently simply picks 2)'
+	echo '      If unspecified, a "smart guess" is attempted'
 	echo '  -b  The branch to compile. (Currently unused and pinned to MASTER)'
 	echo '  -n  Prevent interactive mode'
 	echo '  -d  Dry run. Print the chosen settings and exit'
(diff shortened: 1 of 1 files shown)


LogicParrot:
I'd like to note that my system chokes on 4 threads, even though it is quad core. Perhaps check should be smarter than that and take RAM into account. Or maybe just cap it at 2 or 3.

Bond-009:
`Maybe just cap it at 2 or 3`
The speed difference is too big to just cap it such a low value, I use 8 threads and it's waaaay faster than the old default 2 threads.

LogicParrot:
It's better to be on the safe side and not choke people's systems. If they know their PC can handle it, they can override the default.

Or we could go the RAM checking road. But it might be overkill.

Bond-009:
How much RAM do you have, because I run make on my VPS with 4 threads with only 2GB of RAM in total and it works fine.

LogicParrot:
I use 4GBs of RAM. 

I am still not completely sure, but this might be a *Linux* issue. Some sources seem to indicate that Linux is not optimized for *desktops*, and when under heavy paging or IO, GUIs freeze. This is not noticable on servers.

Sources:

https://lwn.net/Articles/682582/

https://superuser.com/questions/1115983/prevent-system-freeze-unresponsiveness-due-to-swapping-run-away-memory-usage

bearbin:
This is particularly relevant for early RasPis (~256 MiB of RAM). Even later ones would choke with 4 threads, and 2 is the max. I suggest changing the default number of threads to 1, but still allowing 'smart' choosing behind an option, e.g:

    --parallel   # runs with `CPUTHREADS` threads
    --parallel 2 # runs with 2 threads

LogicParrot:
Whatever we do, the default shouldn't kill raspberry Pis, and the current default does. This PR needs fixing.

LogicParrot:
I tried to run compilation on a secondary harddrive. Things are much smoother now, even with `-j4`. I speculate the compiler queues up too many writes, so paging requests and write requests of other programs have to sit in a very long queue. This is just a guess based on the above links.

LogicParrot:
@bearbin I am working on it. Do you think Raspberry Pis should default to 1 or 2?

bearbin:
I think the default number of threads should be 1, for maximum compatibility. The newer pies with 1GB of ram can handle 2, but not the older ones.

bearbin:
Replaced by #3960

PR #3645, opened 2017-03-23
If player is dead, set wolf's target nullptr.

Seadragon91:
If the player dies and respawns, the wolf/wolves still have him as target and will attack him again. In vanilla minercaft that's no the case.


Seadragon91:
A few problems have appeared with that, not sure what's going on there.

Issue #3627, opened 2017-03-17
Golden apples eating bug


willi123yao:
Client version:   1.8
Server OS:        Linux
Commit id:        9fa8477b73ce878c631ba875ec10746b4143ba75

### Expected behavior
In creative mode, it should be impossible to eat food at all.

### Actual behavior
When right-clicking golden apples in creative, it will get consumed after a few seconds

### Steps to reproduce the behavior
Go into creative mode, get a golden apple(enchanted or non-enchanted) and right-click. After 3 secs it will be consumed.

Lightn1ng:
Can't reproduce here.

PR #3656, opened 2017-03-30
Don't feed golden apple in creative mode

Lightn1ng:
Fixes issue #3627

Took me 3 pull requests to eventually realize that it doesn't work with trailing whitespace

Ryan El Kochta <ryane512@gmail.com>

### DIFF
--- a/src/Items/ItemGoldenApple.h
+++ b/src/Items/ItemGoldenApple.h
@@ -20,6 +20,11 @@ public:
 
 	virtual bool EatItem(cPlayer * a_Player, cItem * a_Item) override
 	{
+		// If player in creative mode, don't feed them:
+		if(a_Player->IsGameModeCreative())
+		{
+			return false;
+		}
 		// Feed the player:
 		FoodInfo Info = GetFoodInfo();
 		a_Player->Feed(Info.FoodLevel, Info.Saturation);
--- a/src/Items/ItemGoldenApple.h
+++ b/src/Items/ItemGoldenApple.h
@@ -21,7 +21,7 @@ public:
 	virtual bool EatItem(cPlayer * a_Player, cItem * a_Item) override
 	{
 		// If player in creative mode, don't feed them:
-		if(a_Player->IsGameModeCreative())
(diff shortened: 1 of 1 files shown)


Lightn1ng:
well, crap. I have had 4 tries to get the code styling just right. 4th times a charm?

worktycho:
Locking at this its making me wonder why on earth the Golden Apple Eating code calls cPlayer::Feed directly rather than delegating to the parent which would fix this problem. Also the current behavior for cuberite is that it lets you eat, but doesn't take the item. Is there a scenario where that could behave differently?

Lightn1ng:
@worktycho Interesting because the other foods that give effects are spider eyes and rotten flesh, and they're handled normally in ItemFood.h

worktycho:
I think its because unlike spider eyes and rotten flesh the golden apple gives at least two status effects. And the current GetEatEffect API can only handle one effect.

Lightn1ng:
@worktycho I'll look into maybe making another api for multiple status effects. Not necessarily aware how to make 1 function to either apply multiple or one.

lkolbly:
What's the status of this? Asking because, in fixing pufferfish, I got rid of the entire GetEatEffect API in PR #3875 and replaced it with how golden apples do it - i.e., directly calling AddEatEffect from EatItem. So that PR and this PR have some things in common.

Lightn1ng:
@lkolbly still not been merged. These people don't seem to like PRs.

_Sent from my Samsung SM-G900V using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

madmaxoft:
Sorry for the delay.

I wouldn't mind merging this, but @worktycho had some concerns, so it's up to him to decide if they have been "addressed enough", or we'll ignore them for now, or something else.

worktycho:
It fixes the problem, but only by fixing the symptom, not that actual root cause which #3875 is addressing, so I prefer that approach.

madmaxoft:
I assume this PR is no longer needed?

worktycho:
Should not be.

PR #3667, opened 2017-04-07
APIDoc: Corrected a few links

Seadragon91:
Corrected class names, added missing `Globals` to enums and adjusted APIDump, it has added `.html` to `https` links.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11451,7 +11451,7 @@ a_Player:OpenWindow(Window);
 							Type = "number",
 						},
 					},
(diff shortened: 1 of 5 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -12175,7 +12175,7 @@ a_Player:OpenWindow(Window);
 							Type = "number",
 						},
 					},
-					Notes = "Sets the skin part flags of the player.  The value should be a bitwise OR of several {{eSkinPart}} constants.",
+					Notes = "Sets the skin part flags of the player.  The value should be a bitwise OR of several {Globals|eSkinPart} constants.",
```
madmaxoft: This looks wrong. I think the best would be `{Globals#eSkinPart|eSkinPart}`, both here and above for the getter. That is, unless the APIDump code already strips the front part.

Seadragon91: Corrected all parts. Should have looked at the web pages :)

PR #3668, opened 2017-04-08
Abort enumeration if error occurs in callback and added note into APIDoc

### LINKED ISSUE #3665: Enumeration still runs if a error occurs in a callback
Server OS:        Windows
Commit id:        4d930b877cbfda32c5d6b7518d889bd6bdc24dea

### Expected behavior
If the code inside of a callback function has an error. The enumeration should be aborted.

### Actual behavior
If the code inside of a callback function has an error. The enumeration is not aborted.
Will end when every entity in the world have been processed. Log file will be spammed with lots of same lines.

Add this code to a command handler and run it
```
cRoot:Get():GetDefaultWorld():ForEachEntity(
	function(a_Entity)
		a_Entity:Infinity()
		return true
	end)
```

Seadragon91:
Fixes #3665.

Return true in callbacks and added note into APIDoc.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -10074,7 +10074,7 @@ a_Player:OpenWindow(Window);
 							Type = "boolean",
 						},
 					},
-					Notes = "If a map with the specified ID exists, calls the CallbackFunction for that map. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cMap|Map}})</pre> Returns true if the map was found and the callback called, false if map not found.",
+					Notes = "If a map with the specified ID exists, calls the CallbackFunction for that map. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cMap|Map}})</pre> Returns true if the map was found and the callback called, false if map not found. Note: If an error occurs in the callback, this function returns true.",
 				},
 				GetNumMaps =
 				{
@@ -12433,7 +12433,7 @@ a_Player:OpenWindow(Window);
 							Type = "boolean",
 						},
 					},
-					Notes = "If there is the player with the uuid, calls the CallbackFunction with the {{cPlayer}} parameter representing the player. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cPlayer|Player}})</pre> The function returns false if the player was not found, or whatever bool value the callback returned if the player was found.",
+					Notes = "If there is the player with the uuid, calls the CallbackFunction with the {{cPlayer}} parameter representing the player. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cPlayer|Player}})</pre> The function returns false if the player was not found, or whatever bool value the callback returned if the player was found. Note: If an error occurs in the callback, this function returns true.",
 				},
 				FindAndDoWithPlayer =
 				{
@@ -12454,7 +12454,7 @@ a_Player:OpenWindow(Window);
 							Type = "boolean",
 						},
 					},
-					Notes = "Calls the given callback function for the player with the name best matching the name string provided.<br>This function is case-insensitive and will match partial names.<br>Returns false if player not found or there is ambiguity, true otherwise. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cPlayer|Player}})</pre>",
+					Notes = "Calls the given callback function for the player with the name best matching the name string provided.<br>This function is case-insensitive and will match partial names.<br>Returns false if player not found or there is ambiguity, true otherwise. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cPlayer|Player}})</pre>Note: If an error occurs in the callback, this function returns true.",
 				},
 				ForEachPlayer =
 				{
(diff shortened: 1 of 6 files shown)


madmaxoft:
I'm not so sure about aborting the calling function. So far we've always done "minimize the impact of an error". What's next, you want to unload a plugin if anything throws an error?

Seadragon91:
Well had thought that it would be logical that an exception is pushed to the next upper function. It's for me a UB, why still running code after callback function finished with an error.

I can remove the call to lua_do_error and update the APIDocs, that true will be returned in case of an failure in the callback function.

madmaxoft:
Such a behavior would be quite inconsistent with the rest of the server. Failing callbacks don't break other such functions (DoWith, ...). Even worse, how about asynchronous callbacks that can but needn't complete immediately (`cNetwork` et al.)? And as I said, where do you stop such an error - will you be like the C++ runtime library that terminates the whole program on an unhandled exception? Right now the server draws the line quite consistently - it terminates the topmost call from C++ to Lua code, but keeps the rest of the plugin going. Which looks pretty Defined to me :)

Seadragon91:
Okay will rework like you wanted it :)

Seadragon91:
Reworked it and added a own commit for changes in code and for APIDoc.

src/Bindings/ManualBindings.cpp
```diff
@@ -1237,15 +1237,20 @@ static int tolua_cPluginManager_ForEachCommand(lua_State * tolua_S)
...
+			// An error occurred
+			return true;
 		}
 		cLuaState & m_LuaState;
 		cLuaState::cRef & m_FnRef;
 	} Callback(L, FnRef);
 
 	// Execute and push the returned value:
 	L.Push(cPluginManager::Get()->ForEachCommand(Callback));
+
```
Seadragon91: Reverting this change.

src/Bindings/ManualBindings.h
```diff
@@ -427,14 +446,12 @@ class cManualBindings
 			virtual bool Item(Ty2 * a_Item) override
 			{
 				bool res = false;
-				if (!m_LuaState.Call(m_FnRef, a_Item, cLuaState::Return, res))
+				if (m_LuaState.Call(m_FnRef, a_Item, cLuaState::Return, res))
 				{
-					LOGWARNING("Failed to call Lua callback");
-					m_LuaState.LogStackTrace();
```
Seadragon91: Is it okay to remove that or should I revert this?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -10074,7 +10074,7 @@ a_Player:OpenWindow(Window);
 							Type = "boolean",
 						},
 					},
-					Notes = "If a map with the specified ID exists, calls the CallbackFunction for that map. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cMap|Map}})</pre> Returns true if the map was found and the callback called, false if map not found.",
+					Notes = "If a map with the specified ID exists, calls the CallbackFunction for that map. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cMap|Map}})</pre> Returns true if the map was found and the callback called, false if map not found. Note: If an error occurs in the callback, the callback returns true.",
```
madmaxoft: the callback returns true -> this function returns true.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -12433,7 +12433,7 @@ a_Player:OpenWindow(Window);
 							Type = "boolean",
 						},
 					},
-					Notes = "If there is the player with the uuid, calls the CallbackFunction with the {{cPlayer}} parameter representing the player. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cPlayer|Player}})</pre> The function returns false if the player was not found, or whatever bool value the callback returned if the player was found.",
+					Notes = "If there is the player with the uuid, calls the CallbackFunction with the {{cPlayer}} parameter representing the player. The CallbackFunction has the following signature: <pre class=\"prettyprint lang-lua\">function Callback({{cPlayer|Player}})</pre> The function returns false if the player was not found, or whatever bool value the callback returned if the player was found. Note: If an error occurs in the callback, the callback returns true.",
```
madmaxoft: the callback -> this function (and pretty much everywhere else in the APIDocs)

src/Bindings/ManualBindings.cpp
madmaxoft: Remove the empty line altogether, the `Push` and `return` belong together.

src/Bindings/ManualBindings.h
madmaxoft: I believe the failure in the callback is already reported to the console, including the stack trace, so there should be no need to report the failure again.

madmaxoft:
I don't think this needs two separate commits, I'd do it in a single commit if I was writing this.

Server/Plugins/APIDump/APIDesc.lua
Seadragon91: Do you mean the callback function or the API function? The API function returns false, if the callback returns true.
https://api.cuberite.org/cWorld.html#ForEachEntity_1

madmaxoft: Ah, right. But that's the point, really, with your wording it isn't clear what happens.

Seadragon91:
Not all DoWith-functions returns false.
This returns true on error:
```
DoWithEntityByID
DoWithPlayer
DoWithPlayerByUUID
FindAndDoWithPlayer
```

Is it okay?

Seadragon91:
@madmaxoft Should I change the DoWith-functions that doesn't return false?

madmaxoft:
Yes, change those to return false on failure, too. The general idea is to return true only on success; a raised error in the handler should not be considered a success.

Seadragon91:
I just saw that sentence in all the DoWith-Furnace / Chest and all other functions:
`it returns the bool value that the callback has returned`
Looking at [Chunk.cpp#L2244](https://github.com/cuberite/cuberite/blob/85f1169/src/Chunk.cpp#L2244) this is wrong.
Swapping the two return values and then it would be correct, but then the function will return true, too, in case of an error.

tigerw:
I don't think syntax errors in callback functions should be communicated as a return value. What is the output of broken code? It's a meaningless question. At the very least we should just leave the actual return value in these cases undefined (in the API docs), aborting the enumeration. Better would be to throw a Lua error, but that may need some changes to PushCallPop to do properly, without printing multiple duplicate error messages.

tigerw:
How about we fix some of #4266, apply the enumeration abort changes (but without codifying the return value in the docs).

Hopefully in some migration to sol2 in the distant future, a call into Lua that fails throws an exception that propagates out the C++ DoWith/ForEach callback and gets neatly handled somewhere in the sol2 runtime.

PR #3675, opened 2017-04-19
New icon colors

mathiascode:
Updated the icon colors to the new color scheme.


Cl1608Ho:
Would he nice if you asked us before deciding on such a color scheme! 

The gradients are nice, but the black was a lot better imo!

madmaxoft:
The discussion took place in the website repo: cuberite/cuberite.github.io#77
I personally like the new icons more, they are more distinguished.

madmaxoft:
I think the Server/webadmin/files/favicon.ico file should be updated too. But I'm not sure if it should be the exact same as the main site's favicon, because that would make it indistinguishable again.

NiLSPACE:
Perhaps the gradient should be reversed?

PR #3686, opened 2017-05-01
Don't return const ref params and removed functions from ManualBindings.cpp

Seadragon91:


madmaxoft:
You seem to have forgotten to commit the changes in the ToLua++ submodule.

Seadragon91:
Yes just did that, should have them pushed at first :)

madmaxoft:
Seems like we have a new LuaBindings expert :)

madmaxoft:
This doesn't seem to work for me, the `cWebAdmin:GetHTMLEscapedString()` still returns its param as the second return value.

madmaxoft:
Ah, I see, that's because on my system (Win, MSVC, no `xxd`) the build process doesn't use the system-provided Lua, and therefore compiles a ToLua++ binary rather than using it as Lua scripts.

PR #3687, opened 2017-05-01
APIDoc: Removed non-existent functions and added missing return types

Seadragon91:


Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -3167,7 +3150,14 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
...
 				},
-				Notes = "Spawns a {{cTNTEntity|primed TNT entity}} at the specified coords, with the given fuse ticks. The entity gets a random speed multiplied by the InitialVelocityCoeff, 1 being the default value.",
+				Returns =
+				{
+					{
+						Name = "EntityID",
+						Type = "number",
+					},
+				},
+				Notes = "Spawns a {{cTNTEntity|primed TNT entity}} at the specified coords, with the given fuse ticks. The entity gets a random speed multiplied by the InitialVelocityCoeff, 1 being the default value. Returns the EntityID of the spawned primed TNT entity , or -1 on failure.",
```
Seadragon91: Just see that the there is no check if the `Initialize` returns false.
[World.cpp#L2287](https://github.com/cuberite/cuberite/blob/74e6f41/src/World.cpp#L2287)

Will check if there are more entities / mobs that are missing a check there.

Seadragon91: Will do that in a own pull request.

Seadragon91: Ops wrong number if spawn failed, it should be 0.

Seadragon91:
Reverted the change to `SpawnPrimedTNT`, will be added in a own pull request.

Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -2850,7 +2827,13 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
...
 					},
 				},
-				Notes = "Opens or closes a trapdoor at the specific coordinates.",
+				Returns =
+				{
+					{
+						Type = "boolean",
+					},
+				},
+				Notes = "Opens or closes a trapdoor at the specific coordinates. Returns true on success, false if there is no trap door or it's already opened or closed.",
```
madmaxoft: `or it's already in the requested state` might be better

Seadragon91: Changed it.

PR #3688, opened 2017-05-02
Added missing checks for Initialize function and updated APIDoc

Seadragon91:
This adds missing checks for the `Initialize`-function to all functions that wants to spawn a entity. Also added `cEntity#INVALID_ID` to all the functions in the APIDoc, if the spawning failed.

### DIFF
--- a/Server/Plugins/APIDump/Classes/World.lua
+++ b/Server/Plugins/APIDump/Classes/World.lua
@@ -2910,7 +2910,7 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 						Type = "number",
 					},
 				},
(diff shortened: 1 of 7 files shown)


worktycho:
The ownership semantics of Initialize are confusing and force any code interacting with it to use raw pointers. I think we need to make Initialize a static method that takes a unique_ptr to the object, rather than making is an object method that takes ownership of this.

Other than that, the early exits are important.

madmaxoft:
I would have preferred a solution without explicit `delete` calls, but I guess this is good enough. The `Initialize` code is a hot candidate for rewrite anyway, as @worktycho says.

Seadragon91:
Forgot a check:
[World.cpp#L2881](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L2881)

worktycho:
This isn't really the place for it, but just looking into refactoring Initialize, cPlayer::Initialize seems to have very different semantics from the other entity initialization functions. AddPlayer vs AddEntity is a problem.

PR #3691, opened 2017-05-04
Revised handling of time increments to ripen for cBlockCropsHandler blocks

dbhicks:
I'm fairly new to Cuberite. After installing and running the server, I initiated a survival map, but noticed immediately that many of my crops weren't producing the expected yield. I think the culprit was that the literal being passed into the template class was too high or the test value in cWorld::GrowRipePlant was too low. In either case, the crops in question would never have their meta value incremented to the point where ConvertToPickups would consider it ripe enough to bear fruit.

I also tried to cut down on some of the repeated literals with some additional constants and a function.

This builds on my system and, after some simple testing, seems to solve the problem.

### DIFF
--- a/src/Blocks/BlockCrops.h
+++ b/src/Blocks/BlockCrops.h
@@ -29,7 +29,7 @@ public:
 		cFastRandom rand;
 
 		// If not fully grown, drop the "seed" of whatever is growing:
-		if (a_Meta < RipeMeta)
+		if (a_Meta < (RipeMeta - 1))
 		{
 			switch (m_BlockType)
 			{
@@ -97,7 +97,7 @@ public:
 
 		// If there is still room to grow and the plant can grow, then grow.
 		// Otherwise if the plant needs to die, then dig it up
-		if ((Meta < RipeMeta) && (Action == paGrowth))
+		if ((Meta < (RipeMeta - 1)) && (Action == paGrowth))
 		{
 			a_Chunk.FastSetBlock(a_RelX, a_RelY, a_RelZ, m_BlockType, ++Meta);
 		}


madmaxoft:
I'm with @worktycho here, the template parameter should have been kept, and only the changes to `ConvertToPickups` should be in this PR.

madmaxoft:
Oh my, I think we may have mislead you. I've revised the crops code and it seems to me that the problem is that the code assumes `RipeMeta` is the value of the metadata for the ripe crops block, but the constructors (in BlockHandler.cpp) actually fill it with a value one larger (4 instead of 3, 8 instead of 7). So it seems to me that the most proper fix would be to actually change the values in BlockHandler.cpp, what do you think?

dbhicks:
I would probably leave it the way it is now for readability/clarity, since the constructor takes a 1-indexed number and I think people in the future will be less likely to make similar errors to the one that was made here if it remains to be 1-indexed.

madmaxoft:
I see a naming problem. The value is called `Meta`, but it currently is not a Meta value, it's rather a NumMetaStates value. This makes a nice WTF.

dbhicks:
Coming into this project blind, I can confirm it was a puzzling choice.

madmaxoft:
Exactly the reason why I think it should be changed :)

dbhicks:
Perhaps I'm confused. Are you suggesting that this PR should be expanded to involve itself with revising variable naming conventions? That, to me, seems unrelated to the intent of my PR, which is to make crops yield fruit when ripe. It may also be a task better suited for someone with greater project-wide perspective than I have. I'd be happy to open an issue to discuss renaming the variable. I think it would be best to take care of that in another branch.

worktycho:
You're right, I'm going to merge this now pending CI success on the updated branch. Refactors tend to have a tendency to grow so lets get this merged to fix the bug for now.

madmaxoft:
Nah, I just wanted you to take the other approach - instead of passing in `<4>` and `<8>` when the objects are constructed, pass in `<3>` and `<7>`, so that the `RipeMeta` value is a real meta.

I guess I'll change it now, since the PR is already merged.

PR #3693, opened 2017-05-05
Prevent infinite loop when logging fails

josh-parris:
If logging fails because of file permissions or whatever, the program goes into a hard loop. Fixed.

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -119,6 +119,7 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	auto fileLogListenerRet = MakeFileListener();
 	if (!fileLogListenerRet.first)
 	{
+		this->m_TerminateEventRaised = true;
 		LOGERROR("Failed to open log file, aborting");
 		return;
 	}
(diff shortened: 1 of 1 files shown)


Seadragon91:
Confirmed. changed permission of my log folder and cuberite runs in a loop.

Seadragon91:
The only problem I see if someone opens cuberite NOT over a terminal, it will directly close without any warning.

josh-parris:
Well, yes, it will close without warning. But its only means of communicating problems - the log files *and* the console aren't available, what mechanism can it use to complain?  Chewing up all the CPU isn't really "communicating" is it.

josh-parris:
It could replace the web server page with an error log.

madmaxoft:
Anyway, it's an improvement over the current status, so I'd say merge, just there's a slight style issue.

src/Root.cpp
```diff
@@ -119,6 +119,7 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	auto fileLogListenerRet = MakeFileListener();
 	if (!fileLogListenerRet.first)
 	{
+		this->m_TerminateEventRaised = true;
```
madmaxoft: Please get rid of the `this->` qualification, it is not needed.

PR #3694, opened 2017-05-05
Corrected brewingstands and added support for fuel

### LINKED ISSUE #3692: Brewing stand doesn't work
Client version:   1.11
Server OS:        Windows

### Actual behavior
Unusable, get out-of-sync error. I think the cause is the new slot (fuel slot)

Will look into fixing that.

Seadragon91:
Do not merge yet.

Done:
* Added new fuel slot
* Added support for fuel: NBT reading, writing and logic
* Updated slot handling: Disallow placing items into wrong slots
* Completely reworked the slot handling for the brewing stand

Fixes #3692

### DIFF
--- a/Server/Plugins/APIDump/Classes/BlockEntities.lua
+++ b/Server/Plugins/APIDump/Classes/BlockEntities.lua
@@ -441,6 +441,26 @@ return
 				},
 				Notes = "Returns the item in the right bottle slot",
 			},
+			GetFuelSlot =
+			{
+				Returns =
+				{
+					{
+						Type = "cItem",
+					},
+				},
+				Notes = "Returns the item in the top left fuel slot",
+			},
+			GetRemainingFuel =
+			{
+				Returns =
+				{
+					{
+						Type = "number",
+					},
+				},
+				Notes = "Returns the remaining fuel that is left. It's the amount of brewing operations that can be done (20 for one blaze powder).",
+			},
 			GetTimeBrewed =
 			{
 				Returns =
@@ -495,6 +515,17 @@ return
 				},
 				Notes = "Sets the item in the right bottle slot",
 			},
+			SetFuelSlot =
+			{
+				Params =
+				{
+					{
+						Name = "FuelSlot",
+						Type = "cItem",
+					},
+				},
+				Notes = "Sets the item in the top left fuel slot",
+			},
 		},
 		Constants =
 		{
@@ -513,6 +544,10 @@ return
 			bsRightBottle =
 			{
 				Notes = "Index of the right bottle slot",
+			},
+			bsFuel =
+			{
+				Notes = "Index of the top left fuel slot",
 			},
 			ContentsHeight =
 			{
--- a/src/BlockEntities/BrewingstandEntity.cpp
+++ b/src/BlockEntities/BrewingstandEntity.cpp
@@ -24,7 +24,8 @@ cBrewingstandEntity::cBrewingstandEntity(int a_BlockX, int a_BlockY, int a_Block
 	m_BlockMeta(a_BlockMeta),
 	m_IsDestroyed(false),
 	m_IsBrewing(false),
-	m_TimeBrewed(0)
+	m_TimeBrewed(0),
+	m_RemainingFuel(0)
 {
 	m_Contents.AddListener(*this);
 	for (int i = 0; i < 3; i++)
@@ -76,6 +77,7 @@ bool cBrewingstandEntity::UsedBy(cPlayer * a_Player)
 	{
 		BroadcastProgress(0, 0);
 	}
+	BroadcastProgress(1, m_RemainingFuel);
 	return true;
 }
 
@@ -95,12 +97,11 @@ bool cBrewingstandEntity::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	// The necessary brewing time, has been reached
 	if (m_TimeBrewed >= m_NeedBrewingTime)
 	{
-		const cBrewingRecipes::cRecipe * Recipe = nullptr;
 		BroadcastProgress(0, 0);
 		m_IsBrewing = false;
 		m_TimeBrewed = 0;
 
-		// Return if the hook has been canceled
+		// Return if the hook has been cancelled
 		if (cPluginManager::Get()->CallHookBrewingCompleting(*m_World, *this))
 		{
 			return false;
@@ -111,7 +112,27 @@ bool cBrewingstandEntity::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		Ingredient.m_ItemCount -= 1;
 		m_Contents.SetSlot(bsIngredient, Ingredient);
 
+		// Fuel slot
+		m_RemainingFuel--;
+		if (m_RemainingFuel <= 0)
+		{
+			if (!m_Contents.GetSlot(bsFuel).IsEmpty())
+			{
+				cItem Fuel = m_Contents.GetSlot(bsFuel);
+				Fuel.m_ItemCount -= 1;
+				m_Contents.SetSlot(bsFuel, Fuel);
+				m_RemainingFuel = 20;
+				BroadcastProgress(1, m_RemainingFuel);
+			}
+		}
+		else
+		{
+			BroadcastProgress(1, m_RemainingFuel);
+		}
+
+
 		// Loop over all bottle slots and update available bottles
+		const cBrewingRecipes::cRecipe * Recipe = nullptr;
 		for (int i = 0; i < 3; i++)
 		{
 			if (m_Contents.GetSlot(i).IsEmpty() || (m_CurrentBrewingRecipes[i] == nullptr))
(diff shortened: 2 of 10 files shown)


src/BlockEntities/BrewingstandEntity.h
```diff
@@ -59,6 +60,8 @@ class cBrewingstandEntity :
 	/** Returns the time that the current items has been brewing, in ticks */
 	short GetTimeBrewed(void) { return m_TimeBrewed; }
 
+	short GetFuel(void) { return m_Fuel; }
```
Seadragon91: Missing export to lua and doxy comment.

Seadragon91:
Completely reworked the slot handling for the brewing stand. Didn't got any problems doing testing.

Seadragon91:
There are still a few problems with it. For example:
1) Bottle slot contains a bottle
2) Place clicks on the bottle slot with an item that is invalid

This will result in a out-of-sync. Should this cases also be handled?

Seadragon91:
Handled the case above and now the slot handling should work like in vanilla ... should :)

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -441,6 +441,16 @@ return
...
 			},
+			GetFuel =
+			{
+				Returns =
+				{
+					{
+						Type = "number",
+					},
+				},
+				Notes = "Returns the fuel that is left.",
```
madmaxoft: `Returns the amount of fuel left`.

src/BlockEntities/BrewingstandEntity.cpp
```diff
@@ -111,7 +112,27 @@ bool cBrewingstandEntity::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		Ingredient.m_ItemCount -= 1;
 		m_Contents.SetSlot(bsIngredient, Ingredient);
 
+		// Fuel slot
+		m_Fuel--;
+		if (m_Fuel == 0)
```
madmaxoft: Shouldn't a `<= 0` check be here, or a `> 0` check for the line above?

src/BlockEntities/BrewingstandEntity.cpp
```diff
@@ -111,7 +112,27 @@ bool cBrewingstandEntity::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+		{
+			if (!m_Contents.GetSlot(bsFuel).IsEmpty())
+			{
+				cItem Fuel = m_Contents.GetSlot(bsFuel);
+				Fuel.m_ItemCount -= 1;
+				m_Contents.SetSlot(bsFuel, Fuel);
+				m_Fuel = 20;
```
madmaxoft: Does each fuel type give the same amount? (I don't know the new game mechanics)

src/BlockEntities/BrewingstandEntity.cpp
```diff
@@ -258,19 +298,10 @@ void cBrewingstandEntity::UpdateProgressBars(bool a_ForceUpdate)
...
-
-
-
-
-
 void cBrewingstandEntity::ContinueBrewing(void)
 {
 	// Continue brewing if number is greater than 0
-	if (m_TimeBrewed > 0)
+	if (m_TimeBrewed > 0 && m_Fuel > 0)
```
madmaxoft: Parentheses around the comparisons.

src/BlockEntities/BrewingstandEntity.h
```diff
@@ -119,6 +126,9 @@ class cBrewingstandEntity :
 	/** Amount of ticks that the current item has been brewed */
 	short m_TimeBrewed;
 
+	/** The fuel for the brewing stand */
+	short m_Fuel;
```
madmaxoft: What's `the fuel`, is it the item being consumed, or some entity? How about renaming to `FuelTicksLeft` (and all the related functions renamed accordingly)?

src/BrewingRecipes.h
```diff
@@ -40,6 +40,9 @@ class cBrewingRecipes
 
 	/** Returns true if the item is a bottle / potion, false if not. */
 	bool IsBottle(const cItem & a_Bottle) const;
+
+	/** Returns true if the item is the fuel, false if not. */
+	bool IsFuel(const cItem & a_Bottle) const;
```
madmaxoft: `a_Bottle`?

src/UI/SlotArea.cpp
```diff
@@ -1975,100 +1975,98 @@ void cSlotAreaBrewingstand::Clicked(cPlayer & a_Player, int a_SlotNum, eClickAct
...
-				case caDblClick:
+				super::ShiftClicked(a_Player, a_SlotNum, Slot);
+				break;
+			}
+			default:
+			{
+				if (!BR->IsFuel(DraggingItem))
 				{
-					DblClicked(a_Player, a_SlotNum);
+					// Deny placing a invalid item into the ingredient slot
```
madmaxoft: Isn't this the `fuel` slot?

src/UI/SlotArea.cpp
```diff
@@ -2078,9 +2076,13 @@ void cSlotAreaBrewingstand::Clicked(cPlayer & a_Player, int a_SlotNum, eClickAct
 
 
 
-void cSlotAreaBrewingstand::HandleBrewedItem(cPlayer & a_Player)
+void cSlotAreaBrewingstand::HandleBrewedItem(cPlayer & a_Player, const cItem & a_ClickedItem)
 {
-	a_Player.AwardAchievement(achBrewPotion);
+	// Check if the clicked item has damage value 0 (water bottle)
```
madmaxoft: -> `Award an achievement if the item is not a water bottle (is a real brewed potion)`

Server/Plugins/APIDump/Classes/BlockEntities.lua
Pokechu22: How about `Returns the remaining amount of fuel`?

madmaxoft: Okay, and the amount is measured how?

src/BlockEntities/BrewingstandEntity.cpp
Pokechu22: There's only one type of fuel: blaze powder (which always gives the same amount).  Nothing else can be put into the slot.

Seadragon91: There is only one fuel item `blaze powder` :)

src/BlockEntities/BrewingstandEntity.h
Pokechu22: Fuel in brewing stands doesn't work by ticks; it's decremented each time brewing starts (but once it starts, the brewing of that batch of potions will continue without decrementing it further).

madmaxoft: Okay, but we still need a more descriptive variable name, so that it's clear that it's not the actual item, but some measure of the fuel left. `m_FuelLeft` would be okayish with me, if we can't think of anything better.

madmaxoft: (and with @Pokechu22 's description in the doxy-comment and somewhere in the APIDoc)

Seadragon91: The fuel means how many brewing operations can be done.
I would say new name `FuelRemaining` and the doxy comment, to:
`Amount of brewing operations that can be done with the remaining fuel.`

madmaxoft: Sounds good. Don't forget to rename the related functions, too.

src/BlockEntities/BrewingstandEntity.cpp
Seadragon91: @Pokechu22 "can be" - we could allow it :), but want vanilla behavior.

src/BlockEntities/BrewingstandEntity.h
Seadragon91: No auto update in github :(. Only saw your comments after refresh...

src/BlockEntities/BrewingstandEntity.cpp
Seadragon91: Changed to `<= 0`.

src/BlockEntities/BrewingstandEntity.cpp
Seadragon91: Added them, can the style check script not detect this?

src/BrewingRecipes.h
Seadragon91: Changed both to `a_Item`, as there are different items passed and then checked there,

src/UI/SlotArea.cpp
Seadragon91: Changed.

Seadragon91:
Updated and also added missing functions `SetSlotFuel` and `GetSlotFuel`.

Seadragon91:
Should have done a local build, updated old function names to new ones.

src/BlockEntities/BrewingstandEntity.cpp
Seadragon91: Looked into the style check and see why it's not checked.

PR #3700, opened 2017-05-09
Exported cFallingBlock and cExpOrb

Seadragon91:
Maybe I could also export `cBoat`, there is the question which functions it makes sense to export.
Also added a note to `INVALID_ID` in `cEntity`.


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -10922,6 +10926,89 @@ a_Player:OpenWindow(Window);
...
 		},
+		cExpOrb =
+		{
+			Desc = [[
+				This class represents a experience orb. This entity can be spawned by using cWorld:SpawnExperienceOrb.
+				It's also spawned when a monster is killed or a player is mining coal.
+			]],
+			Functions =
+			{
+				GetAge =
```
NiLSPACE: What are the Age methods used for? I find it strange that the age of an entity can be changed. Also, any idea why only the experience orb has an age value?

NiLSPACE:
My question doesn't really have anything to do with this PR, I was just curious.

Server/Plugins/APIDump/APIDesc.lua
Seadragon91: The pickup entity has also a age method and they have both a life time of 5 mins.
Should I add this to the description? Would make sense.

Seadragon91:
Your question IS valid ;). Will add this info to both entities.

Server/Plugins/APIDump/APIDesc.lua
NiLSPACE: I guess it would. Perhaps later on it's possible to put it in a separate class which `cPickup` and `cExpOrb` inherit from.

Seadragon91:
Added lifetime to description.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6004,6 +6004,10 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 			},
 			Constants =
 			{
+				INVALID_ID =
+				{
+					Notes = "Special entity id, that is used for functions that returns a zero. For example if the spawning with cWorld:SpawnMob failed.",
```
madmaxoft: Do not mention a zero. Instead, use `failure`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -10922,12 +10926,95 @@ a_Player:OpenWindow(Window);
 			},
 			Inherits = "cEntity",
 		},
+		cExpOrb =
+		{
+			Desc = [[
+				This class represents a experience orb. This entity can be spawned by using cWorld:SpawnExperienceOrb.
```
madmaxoft: `aN experience orb`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -10922,12 +10926,95 @@ a_Player:OpenWindow(Window);
...
+				It's also spawned when a monster is killed or a player is mining coal. The default lifetime of a experience orb is 5 mins.
```
madmaxoft: `an`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -10922,12 +10926,95 @@ a_Player:OpenWindow(Window);
...
+					},
+					Notes = "Sets the experience amount.",
+				},
+			},
+			Inherits = "cEntity",
+		},
+		cFallingBlock =
+		{
+			Desc = [[
+				This class represents a falling block. This entity can be spawned by using cWorld:SpawnFallingBlock.
```
madmaxoft: Perhaps linkify the function name?

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: `Special value of an entity ID, indicating a failure. Used primarily in functions that create entities when the entity cannot be created.`

Seadragon91:
For `cBoat`, I will export for now the function `GetType()`, it returns the wood type of the boat. If another functions should be exported, then they could be done later.

I also see that `cMinecart` is not exported, this one looks more difficult.

Seadragon91:
Did the changes.

Will do `cBoat` in a own pull request, because the type is not handled this also means to modify the protocol and the spawn function should have a new param for the type (optional).

PR #3702, opened 2017-05-09
Fixed Lua bindings for const objects.

### LINKED ISSUE #3701: Lua const correctness
When using the `cLuaState::Push(const TYPE * a_Value)` functions, we're breaking stuff and preparing for a crash:
1. We're breaking const-correctness, since the compiler assumes the object will not be modified, but Lua can modify that object at will
2. Usually this is used in API bindings for return values:
```cpp
int someApiImplementation(lua_State *)
{
  Vector3d someValue;
  LuaState.Push(someValue);  // Want to return a Vector3d object
  return 1;
}
```
This gives Lua a "dangling pointer" (`someValue` object doesn't exist after exiting from the API function implementation, but Lua can still access it via the original pointer).

madmaxoft:
Removes the bindings that break const-correctness, and added an explicit note of the copy-making nature of the calls.

Fixes #3701

### DIFF
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -864,44 +864,11 @@ void cLuaState::Push(const AStringVector & a_Vector)
 
 
 
-void cLuaState::Push(const cCraftingGrid * a_Grid)
-{
-	ASSERT(IsValid());
-
-	tolua_pushusertype(m_LuaState, reinterpret_cast<void *>(const_cast<cCraftingGrid *>(a_Grid)), "cCraftingGrid");
-}
-
-
-
-
-
-void cLuaState::Push(const cCraftingRecipe * a_Recipe)
-{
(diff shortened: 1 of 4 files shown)


madmaxoft:
Do not merge yet, this has a problem with `bool` conversion:
```cpp
const HTTPRequest * req = ...;
Push(req);  // calls Push(bool)
```

madmaxoft:
I hate C++'s (and C's) pointer-to-bool conversions.

madmaxoft:
Should be alright now.

Seadragon91:
Did you had an crash, or how did you detect that problem?

madmaxoft:
I was working on some new manual bindings that were supposed to return a `Vector3d` object, and the API seemingly returned garbage. That's when I realized that the const Push functions are a bad problem.

PR #3704, opened 2017-05-09
Tracer replacement

madmaxoft:
This gets rid of the `cTracer` class that is poorly documented and weirdly implemented, with proper implementations in `cLineBlockTracer`. Also added are new utility functions in `cLineBlockTracer` that greatly simplify things previously done with `cTracer` - LineOfSight tracing and FirstSolidHit tracing.

Also exported the new utility functions to Lua API and provided examples in the Debuggers plugin.

Off-topic: @Seadragon91 This is the new `Vector3d`-returning API :)

### DIFF
--- a/Server/Plugins/APIDump/Classes/Geometry.lua
+++ b/Server/Plugins/APIDump/Classes/Geometry.lua
@@ -977,52 +977,103 @@ return
 	{
 		Desc = [[
 This class provides an easy-to-use interface for tracing lines through individual
-blocks in the world. It will call the provided callbacks according to what events it encounters along the
-way.</p>
-<p>
-For the Lua API, there's only one static function exported that takes all the parameters necessary to do
-the tracing. The Callbacks parameter is a table containing all the functions that will be called upon the
-various events. See below for further information.
+blocks in the world. It can either be used to call the provided callbacks according
+to what events it encounters along the way, or there are shortcut functions used for
+the most popular tracing reasons - line of sight and solid hits.
 		]],
 		Functions =
 		{
+			FirstSolidHitTrace =
+			{
+				{
+					IsStatic = true,
+					Params =
+					{
+						{ Name = "World",  Type = "cWorld" },
+						{ Name = "StartX", Type = "number" },
+						{ Name = "StartY", Type = "number" },
+						{ Name = "StartZ", Type = "number" },
+						{ Name = "EndX",   Type = "number" },
+						{ Name = "EndY",   Type = "number" },
+						{ Name = "EndZ",   Type = "number" },
+					},
+					Returns =
+					{
+						{ Name = "HasHitSolid",    Type = "boolean" },
+						{ Name = "HitCoords",      Type = "Vector3d" },
+						{ Name = "HitBlockCoords", Type = "Vector3i" },
+						{ Name = "HitBlockFace",   Type = "eBlockFace" },
+					},
+					Notes = "If the specified line hits a solid block, return true and the coordinates / face of the first such solid block hit. Returns false if there's no solid block on that line.",
+				},
+				{
+					IsStatic = true,
+					Params =
+					{
+						{ Name = "World", Type = "cWorld" },
+						{ Name = "Start", Type = "Vector3d" },
+						{ Name = "End",   Type = "Vector3d" },
+					},
+					Returns =
+					{
+						{ Name = "HasHitSolid",    Type = "boolean" },
+						{ Name = "HitCoords",      Type = "Vector3d" },
+						{ Name = "HitBlockCoords", Type = "Vector3i" },
+						{ Name = "HitBlockFace",   Type = "eBlockFace" },
+					},
+					Notes = "If the specified line hits a solid block, return true and the coordinates / face of the first such solid block hit. Returns false if there's no solid block on that line.",
+				},
+			},  -- FirstSolidHitTrace
+			LineOfSightTrace =
+			{
+				{
+					IsStatic = true,
+					Params =
+					{
+						{ Name = "World",  Type = "cWorld" },
+						{ Name = "StartX", Type = "number" },
+						{ Name = "StartY", Type = "number" },
+						{ Name = "StartZ", Type = "number" },
+						{ Name = "EndX",   Type = "number" },
+						{ Name = "EndY",   Type = "number" },
+						{ Name = "EndZ",   Type = "number" },
+						{ Name = "Sight",  Type = "number" },
+					},
+					Returns =
+					{
+						{ Name = "CanSee", Type = "boolean" },
+					},
+					Notes = "Returns true if the two points specified are within line of sight of each other. The Sight parameter specifies which blocks are considered transparent for the trace, it is a combination of {{cLineBlockTracer#eLineOfSight|losXXX}} values added together."
+				},
+				{
+					IsStatic = true,
+					Params =
+					{
+						{ Name = "World", Type = "cWorld" },
+						{ Name = "Start", Type = "Vector3d" },
+						{ Name = "End",   Type = "Vector3d" },
+						{ Name = "Sight", Type = "number" },
+					},
+					Returns =
+					{
+						{ Name = "CanSee", Type = "boolean" },
+					},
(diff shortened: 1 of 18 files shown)


Server/Plugins/APIDump/Classes/Geometry.lua
```diff
@@ -977,52 +977,103 @@ return
...
+						{ Name = "EndZ",   Type = "number" },
 					},
+					Returns =
 					{
-						Name = "Callbacks",
-						Type = "table",
+						{ Name = "HasHitSolid",    Type = "bool" },
+						{ Name = "HitCoords",      Type = "Vector3d" },
+						{ Name = "HitBlockCoords", Type = "Vector3i" },
+						{ Name = "HitBlockFace",   Type = "eBlockFace" },
```
madmaxoft: I'm not too happy with the names for these three return values (params in C++ code). If anyone can suggest a better name, I'm open to suggestions.
`HitCoords` are the exact coords of the hit
`HitBlockCoords` are the coords of the block that has been hit
`HitBlockFace` is the face of the block that has been hit.

madmaxoft:
In-game testing is needed for entity physics handling - do pickups and mobs handle collisions with solid blocks properly (or at least "as good as before this change"?)

src/Bindings/ManualBindings.cpp
```diff
@@ -2962,6 +3175,47 @@ static int tolua_cScoreboard_GetTeamNames(lua_State * L)
 
 
 
+/* method: Trace of class  cTracer */
+static int tolua_cTracer_Trace(lua_State * a_LuaState)
```
NiLSPACE: Didn't we have a specific file for deprecated methods?

madmaxoft: You're right, I'll move it.

Server/Plugins/APIDump/Classes/Geometry.lua
NiLSPACE: `HitCoords` -> `IntersectionPosition` or `IntersectionCoords` ? The place where two lines meet is called an intersection, so it makes sense.

The rest seem fine by me.

madmaxoft: It's not two lines, it's a line against a cube.

NiLSPACE: I guess you could argue that a 3d cube is made out of 2d squares which is made out of 1d lines.

madmaxoft:
@NiLSPACE are you using `cTracer` anywhere in WorldEdit? Need help migrating?

Server/Plugins/APIDump/Classes/Geometry.lua
madmaxoft: The best I got so far is `ExactHitCoords`. Intersection just doesn't have the right feel to it at all.

NiLSPACE:
WorldEdit should be using `cLineBlockTracer` everywhere. The only plugin using `cTracer` that I can remember is [this very old plugin](https://forum.cuberite.org/thread-771.html?highlight=JumpTo) which is totally broken by now.

mathiascode:
Some of my plugins use cTracer, as well as Essentials in the cuberite organization, but I'll migrate them to cLineBlockTracer soon.

NiLSPACE:
When searching the entire organization only Essentials comes up: https://github.com/search?q=org%3Acuberite+cTracer&type=Code

madmaxoft:
@mathiascode Probably not worth migrating until this PR is merged, so that you can use the new utility functions in `cLineBlockTracer` that are an almost direct replacement.

src/LineBlockTracer.h
```diff
@@ -33,6 +33,14 @@ class cLineBlockTracer :
 	typedef cBlockTracer super;
 
 public:
+	enum eLineOfSight
```
worktycho: This seems like it should be a bitset of flags rather than a set of three. What if a plugin wants to see through only water or only lava?

src/LineBlockTracer.cpp
```diff
@@ -49,6 +50,160 @@ bool cLineBlockTracer::Trace(cWorld & a_World, cBlockTracer::cCallbacks & a_Call
...
+
+bool cLineBlockTracer::IntToLineOfSight(int a_Value, eLineOfSight & a_LineOfSight)
+{
+	switch (a_Value)
+	{
+		case losAir:          a_LineOfSight = losAir; return true;
+		case losAirWater:     a_LineOfSight = losAirWater; return true;
+		case losAirWaterLava: a_LineOfSight = losAirWaterLava; return true;
+	}
+	return false;
```
worktycho: This should be a default case.

src/LineBlockTracer.cpp
```diff
@@ -49,6 +50,160 @@ bool cLineBlockTracer::Trace(cWorld & a_World, cBlockTracer::cCallbacks & a_Call
...
+
+		case losAirWater:
+		{
+			static class LineOfSightCallbacksAirWater:
+				public cLineBlockTracer::cCallbacks
+			{
+			public:
+				LineOfSightCallbacksAirWater() {}
+
+				virtual bool OnNextBlock(int a_BlockX, int a_BlockY, int a_BlockZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, eBlockFace a_EntryFace) override
```
worktycho: What sort of code does this method generate? it seems a lot of repetition if its not a significant perf gain.

madmaxoft: These callbacks are called in a rather tight loop, so I went for performance, rather than short code.

src/LineBlockTracer.h
madmaxoft: These three are the most common LineOfSight options. If a plugin wants a different one, it will need to use the full `cLineBlockTracer:Trace()` method with custom callbacks.

Even like this it is an improvement over the old `cTracer` that had only one option for LineOfSight tracing, equivalent to `losAirWater`.

src/LineBlockTracer.cpp
worktycho: It feels like the virtual function call is going to be a bigger part of that. I'm now wondering if there is any way we can get rid of that, would it be worth templating the tracer based on what blocks we are interested/not interested in?

Also the callbacks are massively ugly. I'd love to suggest changing the callbacks to use closures but I'm not sure thats possible without introducing another allocation.

madmaxoft: Templating might be the next step, if we find out this piece of code needs optimizing.

The callbacks are implemented as a class because they can have multiple entrypoints (`OnNextBlock`, `OnOutOfWorld`, `OnNoMoreHits` etc.), I don't think that can be reasonably achieved with closures.

madmaxoft:
@worktycho Alright, I'll change the callbacks around a bit.

madmaxoft:
@worktycho Changed the code, now there's just one `callbacks` class. Also the enum is now used as a bitfield.

madmaxoft:
I've tested this in-game and it seems that the physics works more or less unchanged.

spekdrum:
Hi, it seems that since this commit 2c3c1f mobs are static. Monsters  are not moving towards players, etc. Maybe I'm wrong. The fact is that todays master has this problem but 7c4576a (the first I tried) works. I've been looking the commits between and the only that affects monsters "sight" seems to be this one.

madmaxoft:
Could you git-bisect the exact commit that did it?

I tested basic physics handling with pickups in water (other than water direction behaving a bit funny, it worked well) and a few zombies and creepers, all seemed normal.

Seadragon91:
Bisected it. Yes it's this pull request.

mathiascode:
Pickups also started moving/floating.

Seadragon91:
@mathiascode All confirmed, arg...

spekdrum:
Confirmed, checkin out 7c4576 mobs are moving again. And checking out 2c3c1f they become still.

Sorry for the ignorance but Is it a normal behavior that mobs (monsters) gets destroyed when player disconnects, and on reconnect they are gone? It happens on 7c4576.

mathiascode:
Can we revert this change until it's fixed?

madmaxoft:
Don't, the fix shouldn't be too difficult, and it's been a week, so a day or two more won't harm that much.

PR #3707, opened 2017-05-12
Fix some compiler warnings (Linux)

josh-parris:
I noticed that a build produces many compiler warnings.  I fixed some.  I'm not sure which commits are acceptable.


madmaxoft:
Unfortunately most of these "fixes" are actually anti-fixes - either they just hide the original issue, or even make the code wrong.

src/Bindings/LuaJson.cpp
```diff
@@ -177,11 +177,11 @@ static Json::Value JsonSerializeTable(cLuaState & a_LuaState)
 {
 	Json::Value res;
 	lua_pushnil(a_LuaState);
+	int idx = 0;
```
madmaxoft: What warning was the reason for this move? I'm not getting any warnings here.

josh-parris: idx may be used uninitialized in this function [-Wmaybe-uninitialized]

josh-parris:
ManualBindings_Network.cpp - the code is inconsistent at the very least.  In one place ints are used, another UInt16 (which can't be range checked). I thought the type was correct, as the error thrown out will specify what's wrong, rather than silently eating the number and failing it because of range checking.

The changes to "Change loop to ensure at least one pass and fix compiler warning" in two functions render the assertion at the top of those functions defunct, as the loop definitely executes at least one pass.

The IsJungleVariant() change is trivial and silences the warning.

The StructGen.cpp HEIGHTTYPE fix is inconsequential as you're not going to generate a chuck that doesn't have bottom. The exhaustive listing of enumerations is necessary for the warning to sound when there's an actual problem.

I can see with some of the full lists of enums there may be enums that are missing because of missing functionality - if that's the case, drop those enums.

Which changes make the code wrong?

Seadragon91:
Can confirm the cNetwork:CreateUDPEndpoint port check under windows. The function `GetStackValues` "sees" that the passed number doesn't fit into `UInt16`.

Test code: `cNetwork:CreateUDPEndpoint(100000, {})` won't reach the port check.

madmaxoft:
@Seadragon91 That's an ambiguous way to say that.

The `GetStackValue()` function for all integer types checks the ranges of the numbers received from Lua and reports failure if the number doesn't fit.

PR #3708, opened 2017-05-13
Use FastWriter instead of StyledWriter

mathiascode:
Also send the latest protocol version number like other servers do, instead of an invalid one.
Fixes http://minecraft-mp.com/ support for Cuberite.

### DIFF
--- a/src/Protocol/ProtocolRecognizer.h
+++ b/src/Protocol/ProtocolRecognizer.h
@@ -20,6 +20,7 @@
 // Adjust these if a new protocol is added or an old one is removed:
 #define MCS_CLIENT_VERSIONS "1.8.x, 1.9.x, 1.10.x, 1.11.x"
 #define MCS_PROTOCOL_VERSIONS "47, 107, 108, 109, 110, 210, 315, 316"
+#define MCS_LATEST_PROTOCOL_VERSION 316
 
 
 
--- a/src/Protocol/ProtocolRecognizer.cpp
+++ b/src/Protocol/ProtocolRecognizer.cpp
@@ -1157,7 +1157,7 @@ void cProtocolRecognizer::SendPingStatusResponse(void)
 	// Version:
 	Json::Value Version;
 	Version["name"] = "Cuberite " MCS_CLIENT_VERSIONS;
-	Version["protocol"] = 0;  // Force client to think this is an invalid version (no other good default)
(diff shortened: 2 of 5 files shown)


madmaxoft:
Travis is acting up: `abuse detected: known offender (request looked fishy)`
Ref.: https://travis-ci.org/cuberite/cuberite/requests

Can you amend the commit and force-push, to try building once more?

src/Protocol/ProtocolRecognizer.h
```diff
@@ -20,6 +20,7 @@
 // Adjust these if a new protocol is added or an old one is removed:
 #define MCS_CLIENT_VERSIONS "1.8.x, 1.9.x, 1.10.x, 1.11.x"
 #define MCS_PROTOCOL_VERSIONS "47, 107, 108, 109, 110, 210, 315, 316"
+#define MCS_LATEST_PROTOCOL_VERSION "316"
```
Pokechu22: I'm not 100% sure with how the JSON writer works, but won't this being a string cause the protocol version to be sent as a string, instead of a number?

Shouldn't this be just

```c++
#define MCS_LATEST_PROTOCOL_VERSION 316
```
?

Minecraft's own JSON parser will silently correct it, but I'm pretty sure that that other 3rd-party programs won't be as happy about it.

mathiascode: You're correct. Thanks for pointing it out.

Seadragon91:
What's going on which travis...?

madmaxoft:
It seems @mathiascode is a `known offender`, so his PRs won't get built =D

yangm97:
Hide your kids and hide your wife from @mathiascode !

Seadragon91:
Well looks like the only way to correct this problem, is to contact travis support per e-mail (found in a issue).

madmaxoft:
Okay, @mathiascode write to them, Cc me (anything at xoft dot cz) if you think you need my support.

mathiascode:
Looks like I got busted. D:

But seriously, I wonder what I've done to trigger their system. I'll have to send an e-mail then.

madmaxoft:
Did they say why you were tagged as "known offender"?

mathiascode:
"Our system automatically flagged your account for possible abuse, but after further review, we were able to clear that up."

PR #3709, opened 2017-05-15
VarInt metadata written correctly for boats

### LINKED ISSUE #3703: cBoat: Poblem with type meta data
Client version:   1.11.2
Server OS:        Windows

Added type meta for the boat, but it doesn't work. I miss something. I added logging in Protocol_1_11 and see that the type value 4 (used it for testing) is always send, but the boat I see on client side is still from type 0.

Created a commit where I added the type for the boat and exported it:
https://github.com/Seadragon91/cuberite/commit/ee75ea67721ac369527de7b30f4074da5c62cb74

http://wiki.vg/Entities#Boat

peterbell10:
Fixes  #3703 

I have tested with the 1.11.2 client and boat types now display correctly.

### DIFF
--- a/src/Protocol/Protocol_1_10.cpp
+++ b/src/Protocol/Protocol_1_10.cpp
@@ -511,11 +511,11 @@ void cProtocol_1_10_0::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & 
 
 			a_Pkt.WriteBEInt8(BOAT_LAST_HIT_TIME);
 			a_Pkt.WriteBEInt8(METADATA_TYPE_VARINT);
-			a_Pkt.WriteBEInt32(Boat.GetLastDamage());
+			a_Pkt.WriteVarInt32(Boat.GetLastDamage());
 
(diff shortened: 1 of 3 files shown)


Seadragon91:
Ah that was the problem, thanks. Now I can finish my export boat branch :)

madmaxoft:
Are you sure about the changes in the 1.10 and 1.9 protocols? One protocol using varints is no guarantee of an earlier protocol using them too; you should check each protocol separately.

peterbell10:
All versions previously used METADATA_TYPE_VARINT so at the very least it now agrees with the other data in the packet. I'll give it a test though.

Seadragon91:
Works correctly in 1.9 / 1.10 and 1.11.

madmaxoft:
Thanks.

peterbell10:
Ok, boat types confirmed in 1.9 and 1.10.
Although I can't be sure of

Seadragon91:
@peterbell10 
Yes the problem is spawning the boat with right-click doesn't set the type.  As the different boats all habe a own item id. I tested your change in my branch and there it works correctly.

PR #3710, opened 2017-05-15
Exported boat and added type.

Seadragon91:
Also added the wood types in the item handler of the boat.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11009,6 +11009,149 @@ a_Player:OpenWindow(Window);
 			},
 			Inherits = "cEntity",
 		},
+		cBoat =
+		{
+			Desc = [[
+				This class represents a boat. This entity can be spawned by using {{cWorld#SpawnBoat_1|cWorld:SpawnBoat}}.
+			]],
+			Functions =
+			{
+				GetMaterial =
+				{
+					Returns =
+					{
+						{
+							Name = "Material",
+							Type = "cBoat#eMaterial",
+						},
+					},
+					Notes = "Returns the material of the boat.",
+				},
+				MaterialToString =
+				{
+					IsStatic = true,
+					Params =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "Material",
+							Type = "string",
+						},
+					},
+					Notes = "Returns the boat material as a string.",
+				},
+				ItemToMaterial =
+				{
+					IsStatic = true,
+					Params =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "Material",
+							Type = "cBoat#eMaterial",
+						},
+					},
+					Notes = "Returns the eMaterial that should be used for a boat created from the specified item. Returns bmOak if not a boat item.",
+				},
+				MaterialToItem =
+				{
+					IsStatic = true,
+					Params =
+					{
+						{
+							Name = "Material",
+							Type = "cBoat#eMaterial",
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+					},
+					Notes = "Returns the boat item of the boat material",
+				},
+				StringToMaterial =
+				{
+					IsStatic = true,
+					Params =
+					{
+						{
+							Name = "Material",
+							Type = "string",
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "Material",
+							Type = "cBoat#eMaterial",
+						},
+					},
+					Notes = "Returns the boat material for the passed string. Returns oak if not valid.",
+				},
+				SetMaterial =
+				{
+					Params =
+					{
+						{
+							Name = "Material",
+							Type = "cBoat#eMaterial",
+						},
+					},
+					Notes = "Set the material of the boat.",
+				},
+			},
+			Constants =
+			{
+				bmOak =
+				{
+					Notes = "",
+				},
+				bmSpruce =
+				{
+					Notes = "",
+				},
+				bmBirch =
+				{
+					Notes = "",
+				},
+				bmJungle =
+				{
+					Notes = "",
+				},
+				bmAcacia =
+				{
+					Notes = "",
+				},
+				bmDarkOak =
+				{
+					Notes = "",
+				},
+			},
+			ConstantGroups =
+			{
+				eMaterial =
+				{
+					Include = "bm.*",
+					TextBefore = "These constans are the different wood materials of the boat.",
+				},
+			},
+			Inherits = "cEntity",
+		},
 		cPickup =
 		{
 			Desc = [[
--- a/Server/Plugins/APIDump/Classes/World.lua
+++ b/Server/Plugins/APIDump/Classes/World.lua
@@ -2902,6 +2902,10 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 						Name = "Z",
 						Type = "number",
 					},
+					{
+						Name = "Material",
+						Type = "cBoat#eMaterial",
+					},
 				},
 				Returns =
 				{
@@ -2910,7 +2914,7 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 						Type = "number",
 					},
 				},
-				Notes = "Spawns a boat at the specific coordinates. Returns the EntityID of the new boat, or {{cEntity#INVALID_ID|cEntity#INVALID_ID}} if no boat was created.",
+				Notes = "Spawns a {{cBoat|boat}} at the specific coordinates. Returns the EntityID of the new boat, or {{cEntity#INVALID_ID|cEntity#INVALID_ID}} if no boat was created.",
 			},
 			SpawnExperienceOrb =
 			{
--- a/src/Entities/Boat.cpp
+++ b/src/Entities/Boat.cpp
@@ -13,10 +13,10 @@
 
 
 
-cBoat::cBoat(double a_X, double a_Y, double a_Z) :
+cBoat::cBoat(double a_X, double a_Y, double a_Z, eMaterial a_Material) :
 	super(etBoat, a_X, a_Y, a_Z, 0.98, 0.7),
 	m_LastDamage(0), m_ForwardDirection(0),
-	m_DamageTaken(0.0f), m_Type(0),
+	m_DamageTaken(0.0f), m_Material(a_Material),
 	m_RightPaddleUsed(false), m_LeftPaddleUsed(false)
 {
 	SetMass(20.0f);
@@ -55,7 +55,7 @@ bool cBoat::DoTakeDamage(TakeDamageInfo & TDI)
 			if (TDI.Attacker->IsPlayer())
 			{
 				cItems Pickups;
-				Pickups.Add(cItem(E_ITEM_BOAT));
+				Pickups.Add(MaterialToItem(m_Material));
 				m_World->SpawnItemPickups(Pickups, GetPosX(), GetPosY(), GetPosZ(), 0, 0, 0, true);
 			}
 		}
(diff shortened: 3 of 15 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11009,6 +11009,28 @@ a_Player:OpenWindow(Window);
 			},
 			Inherits = "cEntity",
 		},
+		cBoat =
+		{
+			Desc = [[
+				This class represents a boat. This entity can be spawned by using {{cWorld#SpawnBoat_1|cWorld:SpawnBoat}}.
+				There are five wood types: 0 = oak, 1 = spruce, 2 = birch, 3 = jungle, 4 = acacia, 5 = dark oak.
```
madmaxoft: How about adding an `eBoatType` enum with these values, to avoid magic numbers

Seadragon91: Added it :)

Seadragon91:
Okay APICheck failed. How do I have to add the constants? I looked into cEntity and added `Constants` and `ConstantGroups`, but the generated `cBoat` page still only shows the inherited `INVALID_ID`.

madmaxoft:
You added the enum to the global namespace. Move it into the `cBoat` class declaration instead.

Seadragon91:
Moved `eBoatType` to the class.

Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -2910,7 +2915,7 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 						Type = "number",
 					},
 				},
-				Notes = "Spawns a boat at the specific coordinates. Returns the EntityID of the new boat, or {{cEntity#INVALID_ID|cEntity#INVALID_ID}} if no boat was created.",
+				Notes = "Spawns a {{cBoat|boat}} at the specific coordinates, the boat type is optional (default 0). Returns the EntityID of the new boat, or {{cEntity#INVALID_ID|cEntity#INVALID_ID}} if no boat was created.",
```
madmaxoft: `0` -> `btOak` (linkified?)

src/Entities/Boat.h
```diff
@@ -13,14 +13,26 @@
...
-
+// tolua_begin
 
 class cBoat :
 	public cEntity
 {
 	typedef cEntity super;
 
 public:
+	enum eBoatType
```
madmaxoft: I think the `Boat` part of the name is superficial. Compare: `cBoat::eBoatType` vs `cBoat::eType`. (but keep the `bt` prefix on values)

src/Entities/Boat.h
```diff
@@ -30,14 +42,15 @@ class cBoat :
 	virtual void Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
 	virtual void HandleSpeedFromAttachee(float a_Forward, float a_Sideways) override;
 
-	cBoat(double a_X, double a_Y, double a_Z);
+	/** The type is the wood type of the boat. There are five boat types (0 = oak, 1 = spruce, 2 = birch, 3 = jungle, 4 = acacia, 5 = dark oak) */
```
madmaxoft: This entire comment is not needed anymore.

src/Entities/Boat.h
```diff
@@ -30,14 +42,15 @@ class cBoat :
...
+	cBoat(double a_X, double a_Y, double a_Z, eBoatType a_BoatType = btOak);
 
 	int GetLastDamage(void) const { return m_LastDamage; }
 	int GetForwardDirection(void) const { return m_ForwardDirection; }
 
 	float GetDamageTaken(void) const { return m_DamageTaken; }
 
-	int GetType(void) const { return m_Type; }
+	int GetType(void) const { return m_Type; }  // tolua_export
```
madmaxoft: Should return `e(Boat)Type`.

src/Items/ItemBoat.h
```diff
@@ -94,8 +94,43 @@ class cItemBoatHandler :
...
+			case E_ITEM_DARK_OAK_BOAT:
+			{
+				BoatType = cBoat::btDarkOak;
+				break;
+			}
+			default:
+			{
+				break;
+			}
+		}
```
madmaxoft: This switch might be useful as a separate static method in the `cBoat` class, possibly exported to Lua.

Now it occurred to me, do we have boat spawning by a dispenser implemented? That should make use of this, too.

src/World.h
```diff
@@ -451,9 +452,9 @@ class cWorld :
 	Returns the UniqueID of the spawned minecart, or cEntity::INVALID_ID on failure. */
 	UInt32 SpawnMinecart(double a_X, double a_Y, double a_Z, int a_MinecartType, const cItem & a_Content = cItem(), int a_BlockHeight = 1);
 
-	/** Spawns a boat at the given coordinates.
+	/** Spawns a boat at the given coordinates. The type is the wood type of the boat.
 	Returns the UniqueID of the spawned boat, or cEntity::INVALID_ID on failure. */
-	UInt32 SpawnBoat(double a_X, double a_Y, double a_Z);
+	UInt32 SpawnBoat(double a_X, double a_Y, double a_Z, cBoat::eBoatType a_BoatType = cBoat::btOak);
```
madmaxoft: I think the param should not have a default, to force each usage to strictly take care of the type.

src/World.h
```diff
@@ -451,9 +452,9 @@ class cWorld :
 	Returns the UniqueID of the spawned minecart, or cEntity::INVALID_ID on failure. */
 	UInt32 SpawnMinecart(double a_X, double a_Y, double a_Z, int a_MinecartType, const cItem & a_Content = cItem(), int a_BlockHeight = 1);
 
-	/** Spawns a boat at the given coordinates.
+	/** Spawns a boat at the given coordinates. The type is the wood type of the boat.
```
madmaxoft: This comment addition is not needed anymore, it's obvious now with the `e(Boat)Type` usage.

Server/Plugins/APIDump/Classes/World.lua
madmaxoft: Perhaps even remove the default, as noted with the C++ implementation.

src/Entities/Boat.h
Pokechu22: ... not to mention that the comment is inaccurate: there are 6 types; it's the maximum type value that is 5.

Seadragon91: Ah yes, hehe. This zero...

Server/Plugins/APIDump/Classes/World.lua
Seadragon91: Removed it.

src/Entities/Boat.h
Seadragon91: Done, yes it's better.

src/Entities/Boat.h
Seadragon91: Fixed.

src/Items/ItemBoat.h
Seadragon91: Moved it.

src/World.h
Seadragon91: Removed.

src/World.h
Seadragon91: Okay, changed it.

Seadragon91:
Added also loading and saving of the type, had forgot that part. Tested and it works.

src/Items/ItemBoat.h
```diff
@@ -95,7 +95,7 @@ class cItemBoatHandler :
 		}
 
 		// Spawn block at water level
-		cBoat * Boat = new cBoat(x + 0.5, y + 0.5, z + 0.5);
+		cBoat * Boat = new cBoat(x + 0.5, y + 0.5, z + 0.5, cBoat::GetTypeFromItem(a_Player->GetEquippedItem().m_ItemType));
```
Seadragon91: Whut? How can this work, I pass here a short instead of a cItem...

src/Entities/Boat.h
```diff
@@ -13,14 +13,26 @@
...
-
+// tolua_begin
 
 class cBoat :
 	public cEntity
 {
 	typedef cEntity super;
 
 public:
+	enum eType
```
madmaxoft: Hm, it just occurred to me that this might be better off being called `Material` rather than `Type` - it's much more descriptive. What do you think? (besides the lot of work this makes for you :)

src/Entities/Boat.h
```diff
@@ -46,14 +63,15 @@ class cBoat :
 
 	void UpdatePaddles(bool rightPaddleUsed, bool leftPaddleUsed);
 
+	static eType GetTypeFromItem(const cItem & a_Item);
```
madmaxoft: Doxy-comment should say what happens on unrecognized `a_Item`.

src/Entities/Boat.cpp
```diff
@@ -169,3 +169,47 @@ void cBoat::UpdatePaddles(bool a_RightPaddleUsed, bool a_LeftPaddleUsed)
...
+		{
+			return cBoat::btAcacia;
+		}
+		case E_ITEM_DARK_OAK_BOAT:
+		{
+			return cBoat::btDarkOak;
+		}
+		default:
+		{
+			LOGD("%s: Item type not handled %d.", __FUNCTION__, a_Item.m_ItemType);
```
madmaxoft: Should be a `LOGWARNING`

src/Entities/Boat.h
madmaxoft: Also, export to Lua.

src/Entities/Boat.cpp
```diff
@@ -169,3 +169,47 @@ void cBoat::UpdatePaddles(bool a_RightPaddleUsed, bool a_LeftPaddleUsed)
...
+			return cBoat::btJungle;
+		}
+		case E_ITEM_ACACIA_BOAT:
+		{
+			return cBoat::btAcacia;
+		}
+		case E_ITEM_DARK_OAK_BOAT:
+		{
+			return cBoat::btDarkOak;
+		}
```
madmaxoft: This is the kind of `switch` that would make sense to use the in-line form:
```cpp
case E_ITEM_BOAT:        return cBoat::btOak;
case E_ITEM_SPRUCE_BOAT: return cBoat::btSpruce;
...
```

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -419,8 +419,44 @@ void cNBTChunkSerializer::AddBasicEntity(cEntity * a_Entity, const AString & a_C
...
+		{
+			Type = "acacia";
+			break;
+		}
+		case cBoat::btDarkOak:
+		{
+			Type = "dark_oak";
+			break;
+		}
+	}
```
madmaxoft: Again, could be an in-line `switch`.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1656,11 +1656,43 @@ void cWSSAnvil::LoadEntityFromNBT(cEntityList & a_Entities, const cParsedNBT & a
...
 	if (!LoadEntityBaseFromNBT(*Boat.get(), a_NBT, a_TagIdx))
 	{
 		return;
 	}
+
+	int TypeIdx = a_NBT.FindChildByName(a_TagIdx, "Type");
+	cBoat::eType Type = cBoat::btOak;
+	if (TypeIdx > 0)
+	{
+		if (a_NBT.GetString(TypeIdx) == "oak")
```
madmaxoft: Get the string into a temporary variable, then compare (to avoid searching for it / creating the `AString` multiple times).

src/WorldStorage/NBTChunkSerializer.cpp
madmaxoft: And it could be a static function in `cBoat`, exported to Lua.

src/WorldStorage/WSSAnvil.cpp
madmaxoft: And again, makes sense to put this into a separate static function in `cBoat` exported to Lua.

src/Entities/Boat.h
Seadragon91: The NBT has `Type` as the key, but yes `Material` makes more sense, what now?

madmaxoft: Save as `Type` to NBT, but use `Material` everywhere else? I personally prefer code (and API) that is descriptive.

src/Entities/Boat.h
Seadragon91: Exported.

src/WorldStorage/WSSAnvil.cpp
Seadragon91: Moved and exported it.

src/WorldStorage/NBTChunkSerializer.cpp
Seadragon91: Moved and exported it.

src/Entities/Boat.cpp
Seadragon91: Changed.

src/Entities/Boat.cpp
Seadragon91: Changed.

src/Entities/Boat.h
```diff
@@ -30,30 +42,46 @@ class cBoat :
...
+	// tolua_begin
+
+	/** Returns the eMaterial of the boat */
+	eMaterial GetMaterial(void) const { return m_Material; }
+
+	/** Sets the eMaterial of the boat */
+	void SetMaterial(cBoat::eMaterial a_Material) { m_Material = a_Material; }
+
+	/** Returns the eMaterial for the passed item. Returns bmOak on failure */
+	static eMaterial GetMaterialFromItem(const cItem & a_Item);
```
madmaxoft: Most of our code uses `Type1ToType2` function names for such static conversions (`ItemTypeToString`, `StringToInteger` etc.), I think it makes more sense, the `Get` part is not needed.

madmaxoft: The doxycomment should specify what kind of item it expects - do you put in `E_ITEM_PLANKS` and it returns the material from that? :)

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11009,6 +11009,127 @@ a_Player:OpenWindow(Window);
...
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "Material",
+							Type = "string",
+						},
+					},
+					Notes = "Returns the boat material as a string.",
```
madmaxoft: `IsStatic = true,` is missing

Seadragon91: Ops...

src/Entities/Boat.h
Seadragon91: `/** Returns the eMaterial for the passed item boat. Returns bmOak if it's no boat item */`

madmaxoft: /** Returns the eMaterial that should be used for a boat created from the specified item. Returns bmOak if not a boat item. */

Seadragon91:
The visual studio compiler is to stupid :(
`'cBoat::MaterialToString': not all control paths return a value`

If I now add a default case, clang comes and says: `will never be executed` - unreachable code.

madmaxoft:
`#ifndef __clang` - already used several times in the code.

Seadragon91:
Changed the static API functions, added a new one `MaterialToItem` and drop the correct boat item if destroyed.

src/Entities/Boat.cpp
```diff
@@ -169,3 +169,105 @@ void cBoat::UpdatePaddles(bool a_RightPaddleUsed, bool a_LeftPaddleUsed)
...
+	{
+		case bmOak:     return "oak";
+		case bmSpruce:  return "spruce";
+		case bmBirch:   return "birch";
+		case bmJungle:  return "jungle";
+		case bmAcacia:  return "acacia";
+		case bmDarkOak: return "dark_oak";
+	}
+	#ifndef __clang__
+		return "oak";
```
madmaxoft: The common practice is to add an `ASSERT` here - we want to catch this if it ever happens in MSVC.

src/Entities/Boat.h
```diff
@@ -30,30 +42,49 @@ class cBoat :
...
+	static eMaterial ItemToMaterial(const cItem & a_Item);
+
+	/** Returns the boat item of the boat material */
+	static cItem MaterialToItem(eMaterial a_Material);
+
+	/** Returns the eMaterial as string */
+	static AString MaterialToString(const eMaterial a_Material);
+
+	/** Returns the boat material for the passed string. Returns oak if not valid */
+	static eMaterial StringToMaterial(const AString  a_Material);
```
madmaxoft: The parameter should be a const reference.

PR #3712, opened 2017-05-17
Clang 5.0 fixes

Seadragon91:
- Added override keyword
- Removed inherited member variables

Also:
* Changed the variable `m_Speed` to `m_MaxSpeed` in Horse.cpp/h as the variable is inherited

### DIFF
--- a/src/Bindings/LuaChunkStay.h
+++ b/src/Bindings/LuaChunkStay.h
@@ -32,7 +32,7 @@ class cLuaChunkStay
 public:
 	cLuaChunkStay();
 
-	~cLuaChunkStay() { }
+	virtual ~cLuaChunkStay() override { }
 
 	/** Adds chunks in the specified Lua table.
 	Can be called only once.
--- a/src/Generating/Caves.h
+++ b/src/Generating/Caves.h
@@ -69,7 +69,6 @@ class cStructGenWormNestCaves :
 public:
 	cStructGenWormNestCaves(int a_Seed, int a_Size = 64, int a_Grid = 96, int a_MaxOffset = 128) :
 		super(a_Seed, a_Grid, a_Grid, a_MaxOffset, a_MaxOffset, a_Size, a_Size, 100),
-		m_Noise(a_Seed),
 		m_Size(a_Size),
 		m_MaxOffset(a_MaxOffset),
 		m_Grid(a_Grid)
@@ -79,7 +78,6 @@ public:
 protected:
 	class cCaveSystem;  // fwd: Caves.cpp
 
-	cNoise       m_Noise;
 	int          m_Size;  // relative size of the cave systems' caves. Average number of blocks of each initial tunnel
 	int          m_MaxOffset;  // maximum offset of the cave nest origin from the grid cell the nest belongs to
 	int          m_Grid;  // average spacing of the nests
--- a/src/Generating/MineShafts.cpp
+++ b/src/Generating/MineShafts.cpp
@@ -1293,7 +1293,6 @@ cStructGenMineShafts::cStructGenMineShafts(
 	int a_ChanceCorridor, int a_ChanceCrossing, int a_ChanceStaircase
 ) :
 	super(a_Seed, a_GridSize, a_GridSize, a_MaxOffset, a_MaxOffset, a_MaxSystemSize, a_MaxSystemSize, 100),
-	m_Noise(a_Seed),
 	m_GridSize(a_GridSize),
 	m_MaxSystemSize(a_MaxSystemSize),
 	m_ProbLevelCorridor(std::max(0, a_ChanceCorridor)),
--- a/src/Generating/Ravines.cpp
+++ b/src/Generating/Ravines.cpp
@@ -82,7 +82,6 @@ protected:
 
 cStructGenRavines::cStructGenRavines(int a_Seed, int a_Size) :
 	super(a_Seed, a_Size, a_Size, a_Size, a_Size, a_Size * 2, a_Size * 2, 100),
-	m_Noise(a_Seed),
 	m_Size(a_Size)
 {
 }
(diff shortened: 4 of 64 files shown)


Seadragon91:
Let's first see if clang 3.4 can compile it. Can not install it on my system anymore...

Seadragon91:
Okay all checks green.

src/Bindings/LuaChunkStay.h
```diff
@@ -32,7 +32,7 @@ class cLuaChunkStay
 public:
 	cLuaChunkStay();
 
-	~cLuaChunkStay() { }
+	~cLuaChunkStay() override { }
```
madmaxoft: I think an `override` should always be a `virtual` (it is implied, but it makes sense to make it explicit)

src/Generating/MineShafts.cpp
```diff
@@ -1293,13 +1293,13 @@ cStructGenMineShafts::cStructGenMineShafts(
...
 ) :
 	super(a_Seed, a_GridSize, a_GridSize, a_MaxOffset, a_MaxOffset, a_MaxSystemSize, a_MaxSystemSize, 100),
-	m_Noise(a_Seed),
 	m_GridSize(a_GridSize),
 	m_MaxSystemSize(a_MaxSystemSize),
 	m_ProbLevelCorridor(std::max(0, a_ChanceCorridor)),
 	m_ProbLevelCrossing(std::max(0, a_ChanceCorridor + a_ChanceCrossing)),
 	m_ProbLevelStaircase(std::max(0, a_ChanceCorridor + a_ChanceCrossing + a_ChanceStaircase))
 {
+	m_Noise = a_Seed;
```
madmaxoft: No need for this, the superclass already does that.

src/Generating/Caves.h
```diff
@@ -69,17 +69,16 @@ class cStructGenWormNestCaves :
 public:
 	cStructGenWormNestCaves(int a_Seed, int a_Size = 64, int a_Grid = 96, int a_MaxOffset = 128) :
 		super(a_Seed, a_Grid, a_Grid, a_MaxOffset, a_MaxOffset, a_Size, a_Size, 100),
-		m_Noise(a_Seed),
 		m_Size(a_Size),
 		m_MaxOffset(a_MaxOffset),
 		m_Grid(a_Grid)
 	{
+		m_Noise = a_Seed;
```
madmaxoft: No need for this, the superclass already does that.

src/Generating/Ravines.cpp
```diff
@@ -82,9 +82,9 @@ class cStructGenRavines::cRavine :
 
 cStructGenRavines::cStructGenRavines(int a_Seed, int a_Size) :
 	super(a_Seed, a_Size, a_Size, a_Size, a_Size, a_Size * 2, a_Size * 2, 100),
-	m_Noise(a_Seed),
 	m_Size(a_Size)
 {
+	m_Noise = a_Seed;
```
madmaxoft: No need for this, the superclass already does that.

src/Generating/RoughRavines.cpp
```diff
@@ -247,6 +246,8 @@ cRoughRavines::cRoughRavines(
 	m_MaxCeilingHeightCenter(a_MaxCeilingHeightCenter),
 	m_MinCeilingHeightCenter(a_MinCeilingHeightCenter)
 {
+	m_Seed = a_Seed;
```
madmaxoft: No need for this, the superclass already does that.

src/Generating/VillageGen.cpp
```diff
@@ -336,14 +336,15 @@ cVillageGen::cVillageGen(
...
 	super(a_Seed, a_GridSize, a_GridSize, a_MaxOffset, a_MaxOffset, a_MaxSize, a_MaxSize, 100),
-	m_Noise(a_Seed + 1000),
 	m_MaxDepth(a_MaxDepth),
 	m_MaxSize(a_MaxSize),
 	m_MinDensity(a_MinDensity),
 	m_MaxDensity(a_MaxDensity),
 	m_BiomeGen(a_BiomeGen),
 	m_HeightGen(a_HeightGen)
 {
+	m_Noise = a_Seed + 1000;
```
madmaxoft: This is a problem. This changes the generated structure, in the old code there are two instances of `cNoise`, each with a separate seed. In the new code, only one instance with this seed is used.

Instead, you should rename this class's `m_Noise` to something else and keep it as a second `cNoise` instance.

src/Mobs/Horse.h
```diff
@@ -44,7 +44,7 @@ class cHorse :
 
 	bool m_bHasChest, m_bIsEating, m_bIsRearing, m_bIsMouthOpen, m_bIsTame, m_bIsSaddled;
 	int m_Type, m_Color, m_Style, m_Armour, m_TimesToTame, m_TameAttemptTimes, m_RearTickCount;
-	float m_Speed;
+	float m_SpeedHorse;
```
madmaxoft: Bad name. Should be something along the lines of `MaxSpeed`

src/Noise/Noise.h
```diff
@@ -21,7 +21,6 @@ class cNoise
 {
 public:
 	cNoise(int a_Seed);
-	cNoise(const cNoise & a_Noise);
```
madmaxoft: Why remove the copy-constructor? I think it should stay

Seadragon91:
Added virtual keyword and reverted the changes for the cNoise constructor.

src/Generating/VillageGen.cpp
```diff
@@ -399,7 +399,7 @@ cGridStructGen::cStructurePtr cVillageGen::CreateStructure(int a_GridX, int a_Gr
 	{
 		return cStructurePtr();
 	}
-	auto rnd = m_Noise.IntNoise2DInt(a_OriginX, a_OriginZ) / 11;
+	auto rnd = m_RandNoise.IntNoise2DInt(a_OriginX, a_OriginZ) / 11;
```
madmaxoft: Btw, this is exactly what the clang warning is for :)

Issue #3384, opened 2016-09-30
Block entitiy type mismatches slow down server


mathiascode:
Client version:   1.10.2
Server OS:        Linux
Commit id:        b268db4caae1b5612ae0f3adabf164efc06de368
### Expected behavior

Cuberite should run normally and not freeze/slow down.
### Actual behavior

When too many block entities are placed with WorldEdit, world preparation on startup will come to a halt.

Example:

```
[00:22:01] Preparing spawn (world): 30.50% (122/400; 53.34 chunks / sec)
[00:22:05] Preparing spawn (world): 30.75% (123/400; 0.26 chunks / sec)
[00:22:06] Block entity type mismatch: exp "Chest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "TrappedChest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "Chest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "TrappedChest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "Chest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "TrappedChest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "Chest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "TrappedChest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "Chest", got "MobSpawner".
[00:22:06] Block entity type mismatch: exp "TrappedChest", got "MobSpawner".
[00:22:13] Preparing spawn (world): 37.50% (150/400; 3.46 chunks / sec)
[00:22:13] Block entity type mismatch: exp "Chest", got "MobSpawner".
[00:22:13] Block entity type mismatch: exp "TrappedChest", got "MobSpawner".
```
### Steps to reproduce the behavior
1. Add WorldEdit to a server
2. Create different block entity spheres close to each other, e.g. //sphere chest 100, //sphere commandblock 100, //sphere hopper 100
3. Restart the server, and it should complain about block entity type mismatches. If you try to place more and more, the server will eventually refuse to start.

madmaxoft:
I don't think the slowness is caused by the mismatch error, but rather just by having to load that many block entities. If you perform the same but don't use a trapped chest, does it still slow down the load? What is the memory footprint of the server - does it run out of memory?

madmaxoft:
[...]

Issue #1901, opened 2015-04-26
Change cChunk::m_BlockEntities into a std::map


madmaxoft:
Currently the block entities are stored in a list - that is highly inefficient for when there are more block entities in a chunk. A single chunk full of mobspawners takes a whopping 10 minutes to load on my laptop in a debug build; most of the time is spent checking whether a block entity exists at the specified coords (inside `cChunk::CreateBlockEntities()`)

worktycho:
Probably better to use unordered_map. It has O(1) lookup and better temporal locality of memory accesses, but would require more allocations during the initial fill. Do you have any idea what the average number of block entities per chunk is? Then we could reserve that many elements ahead of time and avoid most of the allocations.

madmaxoft:
I'd say the average is on the order of tens of entities, hundreds at most. The malicious chunk had over 30k block entities in it.

worktycho:
So if we initially reserve 10 entities that malicious chunk would involve 12 reallocations with unordered_map, so not many.

Seadragon91:
Closed by #3717.

PR #3717, opened 2017-05-19
Store cChunk::m_BlockEntities in a map

peterbell10:
Currently, performance degrades drastically when there are many block entities in a chunk [#1901, #3384].  Switching from a `std::list` to a `std::map` or `std::unordered_map` has greatly improved performance in my tests.

I also cleaned up the `ForEachBlockEntity` and `DoWithBlockEntityAt` functions as there was far too much copy-pasted code.

## Testing method
The server was compiled with MSVC in release mode and run without the debugger.
A new world was generated and used `//sphere hopper 70` from WorldEdit.
After closing the server I backed up the world/region folder.

5 times for each container the region folder was restored from backup and the server started.
The times taken to startup the server was recorded for each.

## Results
image

| Container | Mean startup time (s) |
|---|:---:|
| `std::list` | 59.6(9) |
| `std::map` | 3.6(2) |
| `std::unordered_map` | 2.84(9) |

Clearly there is a huge gain from using an associative container.
There was also a slight improvement from the unordered variety but I'm not sure how well it would translate into normal usage so I've left it as a map.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -2193,26 +2193,31 @@ bool cChunk::ForEachFurnace(cFurnaceCallback & a_Callback)
 
 
 
-bool cChunk::DoWithBlockEntityAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBlockEntityCallback & a_Callback)
+template <class tyEntity, BLOCKTYPE... tBlocktype>
+bool cChunk::GenericDoWithBlockEntityAt(int a_BlockX, int a_BlockY, int a_BlockZ, cItemCallback<tyEntity>& a_Callback)
 {
 	// The blockentity list is locked by the parent chunkmap's CS
-	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(), itr2 = itr; itr != m_BlockEntities.end(); itr = itr2)
+	cBlockEntity * Block = GetBlockEntity(a_BlockX, a_BlockY, a_BlockZ);
+	if (Block == nullptr)
 	{
-		++itr2;
-		if (((*itr)->GetPosX() != a_BlockX) || ((*itr)->GetPosY() != a_BlockY) || ((*itr)->GetPosZ() != a_BlockZ))
-		{
-			continue;
-		}
+		return false;  // No block entity here
+	}
+	if ((sizeof...(tBlocktype) != 0) &&  // Let empty list mean all block entities
+		(!IsOneOf<tBlocktype...>(Block->GetBlockType()))
+	)
+	{
+		return false;  // Not any of the given tBlocktypes
+	}
+	return !a_Callback.Item(static_cast<tyEntity *>(Block));
+}
 
-		if (a_Callback.Item(*itr))
-		{
-			return false;
-		}
-		return true;
-	}  // for itr - m_BlockEntitites[]
 
-	// Not found:
-	return false;
+
+
+
+bool cChunk::DoWithBlockEntityAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBlockEntityCallback & a_Callback)
+{
+	return GenericDoWithBlockEntityAt<cBlockEntity>(a_BlockX, a_BlockY, a_BlockZ, a_Callback);
 }
 
 
@@ -2220,30 +2225,9 @@ bool cChunk::DoWithBlockEntityAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBloc
 
 bool cChunk::DoWithBeaconAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBeaconCallback & a_Callback)
 {
-	// The blockentity list is locked by the parent chunkmap's CS
-	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(), itr2 = itr; itr != m_BlockEntities.end(); itr = itr2)
-	{
-		++itr2;
-		if (((*itr)->GetPosX() != a_BlockX) || ((*itr)->GetPosY() != a_BlockY) || ((*itr)->GetPosZ() != a_BlockZ))
-		{
-			continue;
-		}
-		if ((*itr)->GetBlockType() != E_BLOCK_BEACON)
-		{
-			// There is a block entity here, but of different type. No other block entity can be here, so we can safely bail out
-			return false;
-		}
-
-		// The correct block entity is here
-		if (a_Callback.Item(reinterpret_cast<cBeaconEntity *>(*itr)))
-		{
-			return false;
-		}
-		return true;
-	}  // for itr - m_BlockEntitites[]
-
-	// Not found:
-	return false;
+	return GenericDoWithBlockEntityAt<cBeaconEntity,
+		E_BLOCK_BEACON
+	>(a_BlockX, a_BlockY, a_BlockZ, a_Callback);
 }
 
 
@@ -2252,30 +2236,9 @@ bool cChunk::DoWithBeaconAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBeaconCal
 
 bool cChunk::DoWithBrewingstandAt(int a_BlockX, int a_BlockY, int a_BlockZ, cBrewingstandCallback & a_Callback)
 {
-	// The blockentity list is locked by the parent chunkmap's CS
-	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(), itr2 = itr; itr != m_BlockEntities.end(); itr = itr2)
-	{
-		++itr2;
-		if (((*itr)->GetPosX() != a_BlockX) || ((*itr)->GetPosY() != a_BlockY) || ((*itr)->GetPosZ() != a_BlockZ))
-		{
-			continue;
-		}
-		if ((*itr)->GetBlockType() != E_BLOCK_BREWING_STAND)
-		{
-			// There is a block entity here, but of different type. No other block entity can be here, so we can safely bail out
-			return false;
-		}
-
-		// The correct block entity is here
-		if (a_Callback.Item(reinterpret_cast<cBrewingstandEntity *>(*itr)))
-		{
-			return false;
-		}
-		return true;
-	}  // for itr - m_BlockEntitites[]
-
-	// Not found:
-	return false;
+	return GenericDoWithBlockEntityAt<cBrewingstandEntity,
+		E_BLOCK_BREWING_STAND
+	>(a_BlockX, a_BlockY, a_BlockZ, a_Callback);
 }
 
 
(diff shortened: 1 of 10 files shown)


madmaxoft:
I like most of what I see here very much, just some minor issues; I'll point them out tomorrow, I can't see too well anymore now :)

src/Chunk.cpp
```diff
@@ -335,27 +336,32 @@ void cChunk::SetAllData(cSetChunkData & a_SetChunkData)
...
-		delete *itr;
+		delete KeyPair.second;
 	}
 	m_BlockEntities.clear();
-	std::swap(a_SetChunkData.GetBlockEntities(), m_BlockEntities);
+	for (cBlockEntity * Block : a_SetChunkData.GetBlockEntities())
+	{
+		AddBlockEntityClean(Block);
+	}
+	a_SetChunkData.GetBlockEntities().clear();
```
madmaxoft: I believe the `a_SetChunkData`'s block entities should use the `std::map`, too. In fact, just change the `cBlockEntityList` typedef to `std::map`, then you can keep the `swap` here. If it's not too much trouble, rename the typedef to `cBlockEntities`.

src/Chunk.cpp
```diff
@@ -457,18 +463,7 @@ void cChunk::WriteBlockArea(cBlockArea & a_Area, int a_MinBlockX, int a_MinBlock
...
-			((*itr)->GetPosX() == a_BlockX) &&
-			((*itr)->GetPosY() == a_BlockY) &&
-			((*itr)->GetPosZ() == a_BlockZ)
-		)
-		{
-			return true;
-		}
-	}  // for itr - m_BlockEntities[]
-	return false;
+	return GetBlockEntity(a_BlockX, a_BlockY, a_BlockZ) != nullptr;
```
madmaxoft: Style: We use parentheses around comparison expressions, both in `if` statements and in `bool` assignments / returns.

src/Chunk.cpp
```diff
@@ -1837,25 +1835,23 @@ void cChunk::CollectPickupsByPlayer(cPlayer & a_Player)
...
-			m_World->BroadcastBlockEntity(a_PosX, a_PosY, a_PosZ);
-			return true;
-		}
-	}  // for itr - m_BlockEntities[]
-	return false;
+		return false;  // Not a block entity
+	}
+	if ((Entity->GetBlockType() != E_BLOCK_WALLSIGN) &&
+		(Entity->GetBlockType() != E_BLOCK_SIGN_POST)
+	)
```
madmaxoft: Style:
```cpp
if (
  (a != b) &&
  (c != d)
)
{
  ...
```

src/Chunk.cpp
```diff
@@ -2076,36 +2074,73 @@ bool cChunk::DoWithEntityByID(UInt32 a_EntityID, cLambdaEntityCallback a_Callbac
...
+template <BLOCKTYPE, BLOCKTYPE...> bool IsOneOfImpl(BLOCKTYPE, std::false_type);
+template <class = void> bool IsOneOfImpl(BLOCKTYPE, std::true_type);
+
+template <BLOCKTYPE ... Types>
+bool IsOneOf(BLOCKTYPE x)
+{
+	return IsOneOfImpl<Types ...>(x,
+		std::integral_constant<bool, sizeof...(Types) == 0>()
+	);
+}
```
madmaxoft: Move the function to the top of the file (we "section" files based on globals / class A / class B / ...). Or even better, move it to src/BlockID.h (outside the `tolua_begin` .. `tolua_end` block).

Add a really good doxycomment about what this does, and all the edge cases.

src/Chunk.cpp
```diff
@@ -2076,36 +2074,73 @@ bool cChunk::DoWithEntityByID(UInt32 a_EntityID, cLambdaEntityCallback a_Callbac
...
 	// The blockentity list is locked by the parent chunkmap's CS
-	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(), itr2 = itr; itr != m_BlockEntities.end(); itr = itr2)
+	for (auto & KeyPair : m_BlockEntities)
 	{
-		++itr2;
-		if (a_Callback.Item(*itr))
+		cBlockEntity * Block = KeyPair.second;
+		if ((sizeof...(tBlocktype) == 0) ||  // Let empty list mean all block entities
+			(IsOneOf<tBlocktype...>(Block->GetBlockType()))
+		)
```
madmaxoft: Style: multi-line `if` (same as already in a comment above)

src/Chunk.cpp
```diff
@@ -2202,85 +2194,51 @@ bool cChunk::ForEachDropSpenser(cDropSpenserCallback & a_Callback)
...
-		++itr2;
-		if (((*itr)->GetPosX() != a_BlockX) || ((*itr)->GetPosY() != a_BlockY) || ((*itr)->GetPosZ() != a_BlockZ))
-		{
-			continue;
-		}
+		return false;  // No block entity here
+	}
+	if ((sizeof...(tBlocktype) != 0) &&  // Let empty list mean all block entities
+		(!IsOneOf<tBlocktype...>(Block->GetBlockType()))
+	)
```
madmaxoft: Style: Multi-line `if`

src/Chunk.cpp
```diff
@@ -2618,31 +2361,24 @@ bool cChunk::DoWithFlowerPotAt(int a_BlockX, int a_BlockY, int a_BlockZ, cFlower
...
-
-		// There is a block entity here, but of different type. No other block entity can be here, so we can safely bail out
-		return false;
-	}  // for itr - m_BlockEntitites[]
+		return false;  // Not a block entity
+	}
+	if ((Entity->GetBlockType() != E_BLOCK_WALLSIGN) &&
+		(Entity->GetBlockType() != E_BLOCK_SIGN_POST)
+		)
+	{
```
madmaxoft: Style: Multi-line `if`

src/Chunk.h
```diff
@@ -276,6 +276,12 @@ class cChunk :
 	bool DoWithEntityByID(UInt32 a_EntityID, cEntityCallback & a_Callback, bool & a_CallbackResult);  // Lua-accessible
 	bool DoWithEntityByID(UInt32 a_EntityID, cLambdaEntityCallback a_Callback, bool & a_CallbackResult);  // Lambda version
 
+	/** Calls the callback for each tyEntity; returns true if all block entities processed, false if the callback aborted by returning true
+	tBlocktype is a list of the blocktypes convertible to tyEntity which are to be called. If no block type is given the callback is called for every block entity
```
madmaxoft: `tBlockTypes are blocktypes convertible to tyEntity...`
(`list` could be mistaken for a `std::list`)

src/Chunk.cpp
```diff
@@ -1694,19 +1700,11 @@ cBlockEntity * cChunk::GetBlockEntity(int a_BlockX, int a_BlockY, int a_BlockZ)
...
-		{
-			return *itr;
-		}
-	}  // for itr - m_BlockEntities[]
+	int RelX = a_BlockX - m_PosX * cChunkDef::Width;
+	int RelZ = a_BlockZ - m_PosZ * cChunkDef::Width;
 
-	return nullptr;
+	auto itr = m_BlockEntities.find(MakeIndex(RelX, a_BlockY, RelZ));
+	return itr == m_BlockEntities.end() ? nullptr : itr->second;
```
madmaxoft: Style: parentheses around comparison.
(I don't want to have to look up whether this is evaluated as `(a == b) ? c : d` or `a == (b ? c : d)` )

madmaxoft:
Thanks for implementing this, I've been meaning to do it but never got around to it, and it really helps the performance. I'd personally go with `unordered_map` because there's no place in the code that needs the blockentities sorted in any way, but either is good.

src/Chunk.cpp
```diff
@@ -125,22 +126,22 @@ cChunk::~cChunk()
 
 	// LOGINFO("### delete cChunk() (%i, %i) from %p, thread 0x%x ###", m_PosX, m_PosZ, this, GetCurrentThreadId());
 
-	for (cBlockEntityList::iterator itr = m_BlockEntities.begin(); itr != m_BlockEntities.end(); ++itr)
+	for (auto & KeyPair : m_BlockEntities)
 	{
-		delete *itr;
+		delete KeyPair.second;
```
worktycho: Does this mean that m_BlockEntites owns the BlockEntity? In which case we could we move this data structure to unique pointers?

src/Chunk.cpp
```diff
@@ -2076,36 +2074,73 @@ bool cChunk::DoWithEntityByID(UInt32 a_EntityID, cLambdaEntityCallback a_Callbac
 
 
 
-bool cChunk::ForEachBlockEntity(cBlockEntityCallback & a_Callback)
+// Helper function - there's probably a better place for this
+
+template <BLOCKTYPE, BLOCKTYPE...> bool IsOneOfImpl(BLOCKTYPE, std::false_type);
+template <class = void> bool IsOneOfImpl(BLOCKTYPE, std::true_type);
```
peterbell10: It needs to be callable with empty template arguments like `IsOneOfImpl<>` so it has to be a template with defaulted arguments.  The choice of `class = void` over an empty `BLOCKTYPE ...` just ensures that if `IsOneOf` dispatches to the wrong overload it is always a compile error.

It's definitely ugly but if SFINAE doesn't work properly in msvc then I don't see a better way.

src/Chunk.cpp
madmaxoft: I'd postpone the ownership rewrite into a separate PR, this one is about as much complicated as I'm comfortable with reviewing.

src/Chunk.cpp
madmaxoft: Why exactly do we need the empty arguments? I have done a similar varparam thing with `cLuaState::Push()` and didn't need an empty one.

peterbell10: `IsOneOf<Types ...>` always calls `IsOneOfImpl<Types ...>` which for the empty parameter pack is `IsOneOfImpl<>`. 

If it were simply `bool IsOneOfImpl(BLOCKTYPE, std::true_type)` then it wouldn't be considered in overload resolution because it isn't a template.

madmaxoft: This seems much simpler and works for me (in MSVC, at least):
```cpp
template <BLOCKTYPE Type1, BLOCKTYPE Type2, BLOCKTYPE ... TailTypes>
bool IsOneOf(BLOCKTYPE a_BlockType)
{
	return (a_BlockType == Type1) || IsOneOf<Type2, TailTypes...>(a_BlockType);
}

template<BLOCKTYPE Type>
bool IsOneOf(BLOCKTYPE a_BlockType)
{
	return (a_BlockType == Type);
}
```

peterbell10: I agree that implementation is simpler but IMO having the empty case is useful. Partly because the code currently using it won't compile without it but also because then it behaves identically to this C++17 fold expression: `((a_BlockType == Type) || ...)`

madmaxoft: If you need the empty case, just adding a third function taking no params should be enough for that.

peterbell10: I've realised it actually works fine as is without any tag dispatch.  It still handles empty case and only needs two overloads but still needs the default template argument.

peterbell10:
Ok, the branch is back up to date and I think I've addressed your comments @madmaxoft.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -673,7 +673,10 @@ void cWSSAnvil::LoadBlockEntitiesFromNBT(cBlockEntityList & a_BlockEntities, con
 		}
 
 		// Add the BlockEntity to the loaded data:
-		a_BlockEntities.push_back(be.release());
+		auto Idx = cChunkDef::MakeIndex(be->GetRelX(), be->GetPosY(), be->GetRelZ());
+		a_BlockEntities.insert({ Idx, be.get() });
+		// Release after inserting in case it throws.
+		be.release();
```
madmaxoft: Thanks for fixing this subtle error :)

madmaxoft:
Whoever merges this, don't forget to squash-merge.

PR #3720, opened 2017-05-20
Fixed tracer usage in Entity physics handling.

madmaxoft:
Fixes entity movement problems introduced in #3704.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1122,12 +1122,14 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		Vector3d HitCoords;
 		Vector3i HitBlockCoords;
 		eBlockFace HitBlockFace;
-		if (cLineBlockTracer::FirstSolidHitTrace(*GetWorld(), NextPos, NextPos + NextSpeed, HitCoords, HitBlockCoords, HitBlockFace))
+		Vector3d wantNextPos = NextPos + NextSpeed * DtSec.count();
+		auto isHit = cLineBlockTracer::FirstSolidHitTrace(*GetWorld(), NextPos, wantNextPos, HitCoords, HitBlockCoords, HitBlockFace);
+		if (isHit)
 		{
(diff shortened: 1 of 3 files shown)


madmaxoft:
The `else` branch was in the wrong nesting level, and the IsOnGround check was at a wrong block face (bottom instead of top). Entities move properly now (though still not the same as in Vanilla).

Seadragon91:
Sorry, but the mobs are really strange, when the encounter a block to climb up. They are then placed on the upper block and looking to the opposite.

Seadragon91:
Tested pickups in water, they are moving but in the opposite direction, for all directions :)

Seadragon91:
Inverting the values in `switch (WaterDir)` in file `Entity.cpp` fixes the moving direction.

madmaxoft:
The water direction code is all wrong and will need replacing anyway.

mathiascode:
I'll do some testing for this PR.

madmaxoft:
I think it's definitely better than current master, so even if it isn't 100% perfect, it should be merged and the rest dealt with later.

mathiascode:
What I noticed so far:
- The "Verification failed: bb.CalcLineIntersection(m_Start, m_End, LineCoeff, Face), file cuberite/src/LineBlockTracer.cpp, line 125" crash still occurs randomly, but frequently
- Mobs have a hard time walking up blocks seem to teleport and glitch into the ground when doing it or standing close to an edge
- Pickups and primed TNT teleport to the ground a few blocks above it instead of falling normally

madmaxoft:
I've got everything almost fixed, only the mob pathfinding still has some problems with jumps. Will publish a fixed version soon.

madmaxoft:
The code as it is now works fine for me. Mobs walk normally and there are no asserts. I'd like more people to test this out, though.

Seadragon91:
Will test it.

Seadragon91:
Okay mobs jumps and walks now fine. Still got the crash at the verify line after 15 mins.

mathiascode:
Everything seems normal again.

PR #3721, opened 2017-05-20
Don't destroy monster when last target type is a player

spekdrum:
In current Cuberite version if you are pursued by monsters you just have to disconnect and connect again to get rid of them. If no other player is in your chunk monsters will get destroyed. These changes avoid that.
I considered the situation of several players moving from one place to another at night, making monsters to target them. In this case those mobs will never get destroyed until someone kills them or the Sun burns them or whatever. I think this is not really a problem in terms of gameplay, but I ignore if is there any problem in terms of performance.

### DIFF
--- a/src/Mobs/Monster.h
+++ b/src/Mobs/Monster.h
@@ -177,7 +177,7 @@ public:
 	static cMonster * NewMonsterFromType(eMonsterType a_MobType);
 
 	/** Returns if this mob last target was a player to avoid destruction on player quit */
-	const bool IsLastTargetWasAPlayer() const { return m_LastTargetWasAPlayer; }
+	const bool WasLastTargetAPlayer() const { return m_WasLastTargetAPlayer; }
 
 protected:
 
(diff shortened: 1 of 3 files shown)


src/Mobs/Monster.h
```diff
@@ -251,6 +254,8 @@ class cMonster :
 	int m_Age;
 	int m_AgingTimer;
 
+	bool m_LastTargetWasAPlayer;
```
madmaxoft: Please rename to `m_WasLastTargetAPlayer` and the getter to `WasLastTargetAPlayer()`

spekdrum: Done!

madmaxoft:
You should initialize the `m_WasLastTargetAPlayer` in the constructor, to avoid using an unassigned value.

LogicParrot:
It's worth noting creative players can bypass this.

- Mob chases you in survival (m_WasLastTargetAPlayer is now true)
- Switch to creative (m_WasLastTargetAPlayer is still true)
- Wait until the chunk unloads. Depending on server load and the UnusedChunkCap setting, the time varies, but it never takes more than 5 minutes and it can take seconds.
- The chunk unloads, now the mob is destroyed and saved to NBT along with the chunk.
- Log in again. The mob loads, and now m_WasLastTargetAPlayer is **false**
- When in survival mode, m_WasLastTargetAPlayer is immediately set back to true.
- When in creative mode, it does not, so the next log out destroys the mob.

This has little significance in practical survival gameplay, but it may taint some tests for #3745.

spekdrum:
If the purpose is just testing, in order to find a solution for this invisible thing, maybe you can just modify the source code and force set m_WasLastTargetAPlayer flag to true in the constructor.

Issue #3696, opened 2017-05-05
Entitites turn invisible


mathiascode:
Client version:   1.11.2
Server OS:        Linux
Commit id:        05c3e8bed8d1966f5973380b7fc0c033989cb7ba

### Expected behavior
Entities are always visible when they should be.

### Actual behavior
Chests, mobs and pickups randomly turn invisible.

### Steps to reproduce the behavior
Play survival and notice invisible skeletons and zombies attacking you.

mathiascode:
Probably related to https://github.com/cuberite/cuberite/issues/3609 and https://github.com/cuberite/cuberite/issues/3582

Pokechu22:
Anything in the client log when it's happening?

mathiascode:
Nothing appears in the client log.

spekdrum:
Hi, same problem here, using cuberite last week version. Same behavior with clients from 1.10 to 1.11.2. I compiled from sources on a debian in a Cubox-i 4pro

spekdrum:
Now that chests are not longer invisible this issue is about mobs. I suspect this is related with chunk scope. I mean, when a player is within a chunk everything in It is "ticking". If no player is in a chunk everything stops. I think that maybe mobs needs to broadcast themselves to players when they become active, so that clients (players) know their existence.
I would like to do some testing towards this theory. Anyone knows how to show some kind of in-game visual aids to show chunk borders?

Seadragon91:
The mc client has this :).
Run F3 + q and it will show a list of all commands and F3 + g shows the chunk borders :)

spekdrum:
Oh thats it (facepalm lol)

LogicParrot:
This is being reported multiple times (and also in 1.8.9). Apparently it's not a rare issue. Marking as urgent since it can totally break survival.

Buom01:
This continue to appear to me when they are a lot of mob (more than 60 in my test).
Reproduce: Place around 100 well placed and fixed mob.
Go away (get it far from our rendering view).
Go back.
You should experiment that some of them are missing, but still prevent block placement.

Tested with minecart (and armor stand which are not ready)

DJPlaya:
IDK if its related, but i have a similar Issue.
Randomly Mobs or Players turn invisable, latest Stable build, just the preinstalled Plugins

KeepBotting:
This is indeed still an issue using latest builds and 1.12.2 clients.

[...]

Issue #3403, opened 2016-10-12
Chests are invisible after Re-login


Seadragon91:
Client version:   1.10
Server OS:        Windows/Linux
Commit id:        (4e33569110d9a4e71b1d2fa4ea674225c8d36b88)
### Expected behavior

Chests / trapped chest / ender chest should be visble after re-login.
### Actual behavior

Chests are invisible.
### Steps to reproduce the behavior

1) Login with 1.11.2 client place chest
2) Re-login
3) Chest is not visible anymore

This affects 1.11.2 clients in 1.9 it works fine. Maybe a missing packet update or data value problem.

NiLSPACE:
Are they back when you lief gin with a different MC version?

Seadragon91:
I can see the blocks in 1.9, but not in 1.10 after re-login. Because of this I think a packet has been changed in 1.10.

Pokechu22:
Are they quite literally _invisible_?  And if so, what does "invisible" mean?  Can you still see their hitbox if you look at them, and the model just isn't rendering?  (I'd check this myself but I don't have time to right now).

I don't think any packet would should be necessary to make chests visible; they should show up by default (except for in 1.10.1 and only 1.10.1, since 1.10.1 had some extra block entity requirements not present in 1.10.0 and 1.10.2)

Seadragon91:
They are not rendered after re-login, but I can open them. If I open it or destroy a block near them, then they visible again.

Pokechu22:
Hm.  I know invisible chests occur with [triple chests](https://bugs.mojang.com/browse/MC-34365), but that shouldn't ever happen with single chests.  There's also [MC-65874](https://bugs.mojang.com/browse/MC-65874), which involves chests not rendering even with single chests, but that was supposed to have been fixed in 14w32c (even though people are still claiming to experience it, but that seems to be a memory-related issue, and if other chests are displaying that's probably not it).

Seadragon91:
I read that the current chunk update contains a field named `Block entities`, does the current protcol adds the blockentities?
http://wiki.vg/Protocol#Chunk_Data

Pokechu22:
It adds them using the update block entities packet IIRC, instead of in the field there.

Seadragon91:
I still have this problem maybe it's my PC, Can someone confirm it?

mathiascode:
This is still a problem, and it's not only chests that are invisible. #3696

spekdrum:
can't this issue be closed?

[...]

PR #3722, opened 2017-05-21
Temporal fix for invisible chests (#3479, #3403, #3696)

spekdrum:
I did this fix that solves invisible chests. I know its a bit tricky but it works. Despite this could not be merged maybe gives a bit of light to main coders about what's happening. It seems a protocol problem to me.

### DIFF
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -36,11 +36,8 @@ cChestEntity::~cChestEntity()
 
 void cChestEntity::SendTo(cClientHandle & a_Client)
 {
-	// The chest entity doesn't need anything sent to the client when it's created / gets in the viewdistance
-	// All the actual handling is in the cWindow UI code that gets called when the chest is rclked
-
-	// Temporary fix for invisible chests (#3479, #3403)
-	a_Client.SendUpdateBlockEntity(*this);
+	// Send a dummy "number of players with chest open" packet to make the chest visible:
+	a_Client.SendBlockAction(m_PosX, m_PosY, m_PosZ, 1, 0, m_BlockType);
 }
 
 
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -5,6 +5,8 @@
 #include "../Item.h"
 #include "../Entities/Player.h"
 #include "../UI/ChestWindow.h"
+#include "../ClientHandle.h"
+#include "../Protocol/ProtocolRecognizer.h"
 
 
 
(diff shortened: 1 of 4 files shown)


src/BlockEntities/ChestEntity.cpp
```diff
@@ -38,7 +40,11 @@ void cChestEntity::SendTo(cClientHandle & a_Client)
 	// The chest entity doesn't need anything sent to the client when it's created / gets in the viewdistance
 	// All the actual handling is in the cWindow UI code that gets called when the chest is rclked
 
-	UNUSED(a_Client);
+	// Temporary fix for invisible chests (#3479, #3403)
+	if (a_Client.GetProtocolVersion() >= cProtocolRecognizer::PROTO_VERSION_1_9_0) 
+	{
+		a_Client.SendUpdateBlockEntity(*this);
+	}
```
madmaxoft: How about getting rid of the condition altogether and sending the update every time? Should not cause much trouble, and removes the dependency on the protocol handling.

madmaxoft: According to the protocol wiki http://wiki.vg/Block_Actions#Chest , there is a chest block action, does sending that instead fix the problem? I'd prefer it as a cleaner solution.

spekdrum: Ok, I'm trying it

spekdrum: Actually I was looking here http://wiki.vg/Protocol in "Update Block Entity" section. There's no action related with chests there. 
According to the link you posted I have to check if chest is open to send "1" followed by the number of players looking at. Or 0 if it's closed. But in this context action field "1" means "Set data of a mob spawner".  Anyway now is sending "0", that means chest closed .

Seadragon91: With `SendUpdateBlockEntity` in chest and ender chest they are then both visible.

madmaxoft: I meant that it would make more sense to try this:
```cpp
// Send a dummy "number of players with chest open" packet to make the chest visible:
a_Client.SendBlockAction(m_PosX, m_PosY, m_PosZ, 1, 0, m_BlockType);
```
I did try and the chests are no longer invisible, it works perfectly and doesn't use any undocumented packet.

madmaxoft: And it even works ( = doesn't break) for 1.8 clients.

spekdrum: Ok!

Seadragon91:
Add this fix in EnderChestEntity.cpp, too. Thanks

madmaxoft:
squash-merge if CI builds okay

PR #3731, opened 2017-05-27
Minor cBrewingRecipes cleanup

peterbell10:
`cBrewingRecipes` currently uses the odd `m_pState` variable through which all data is accessed.  Unless I'm missing something, this is simply a way to get around `const`ness.  Instead I use `mutable` and avoid the extra indirection.

Also save some further indirections by storing `cRecipe` items within the struct itself.

### DIFF
--- a/src/BrewingRecipes.cpp
+++ b/src/BrewingRecipes.cpp
@@ -168,7 +168,10 @@ const cBrewingRecipes::cRecipe * cBrewingRecipes::GetRecipeFrom(const cItem & a_
 
 		auto FoundRecipe = std::find_if(m_Recipes.cbegin(), m_Recipes.cend(), [&](const std::unique_ptr<cRecipe>& a_Recipe)
 		{
-			return ((a_Recipe->Input.m_ItemDamage == SplashItemDamage) && (a_Recipe->Ingredient.IsEqual(a_Ingredient)));
+			return (
+				(a_Recipe->Input.m_ItemDamage == SplashItemDamage) &&
+				(a_Recipe->Ingredient.IsEqual(a_Ingredient))
+			);
 		});
 
 		if (FoundRecipe == m_Recipes.cend())
--- a/src/BrewingRecipes.cpp
+++ b/src/BrewingRecipes.cpp
@@ -2,7 +2,6 @@
 #include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
 
 #include "BrewingRecipes.h"
-#include "Item.h"
 
 #include <fstream>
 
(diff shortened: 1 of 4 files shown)


madmaxoft:
Actually the `m_pState` is used mainly for breaking header-file dependencies (the pimpl idiom). Not too useful in this case.

The const_cast will need revisiting in the future, because our Lua bindings don't support constness, so it allows modification of the object. On the other hand, making a copy of the output item and pushing that instead will break the API, because the C++ object is destroyed before Lua even has a chance of touching it; accessing it later could result in crashes. So far your change of the manual binding is the safer one of these two unwanted side-effects.

src/BrewingRecipes.cpp
```diff
@@ -187,64 +147,45 @@ const cBrewingRecipes::cRecipe * cBrewingRecipes::GetRecipeFrom(const cItem & a_
...
-		for (auto & Recipe : m_pState->Recipes)
+		auto FoundRecipe = std::find_if(m_Recipes.cbegin(), m_Recipes.cend(), [&](const std::unique_ptr<cRecipe>& a_Recipe)
 		{
-			if ((Recipe->Input->m_ItemDamage == SplashItemDamage) && (Recipe->Ingredient->IsEqual(a_Ingredient)))
-			{
-				FoundRecipe = &Recipe;
-				break;
-			}
-		}
+			return ((a_Recipe->Input.m_ItemDamage == SplashItemDamage) && (a_Recipe->Ingredient.IsEqual(a_Ingredient)));
```
madmaxoft: Style: line too long, split this into a multi-line condition:
```cpp
return (
  (a == b) &&
  (c == d)
);
```

src/BrewingRecipes.h
```diff
@@ -48,11 +48,14 @@ class cBrewingRecipes
...
 	Logs a warning to the console on input error. */
-	void AddRecipeFromLine(const AString & a_Line, unsigned int a_LineNum);
+	void AddRecipeFromLine(AString a_Line, unsigned int a_LineNum);
 
 	/** Parses an item string, returns true if successful. */
 	bool ParseItem(const AString & a_String, cItem & a_Item);
 
-	struct sBrewingRecipeState;
-	std::unique_ptr<sBrewingRecipeState> m_pState;
+	using cRecipes = std::vector<std::unique_ptr<cBrewingRecipes::cRecipe>>;
```
madmaxoft: Move this line to the top of the `private` section. We lay out things in classes: types first, then data members, and functions last.

peterbell10:
I believe the style issues are fixed now.

I agree that `const_cast` is generally bad but at least the recipes are now `const` on the C++ side.

PR #3734, opened 2017-05-27
Debuggers: Added the "/come" command to test mob pathfinding.

madmaxoft:
Issuing the command instructs the mob you're looking at to navigate to your coords.


madmaxoft:
Note that this code doesn't do much on the current master, you need to either revert #3704 or apply #3720 for the mobs to actually move.

PR #3738, opened 2017-05-28
Check for nil in cPlayer:OpenWindow and close windows on reload

### LINKED ISSUE #3727: API: Passing nil to cPlayer:OpenWindow crashes the server
Client version:   1.11
Server OS:        Windows

### Actual behavior
Passing `nil` to the function `OpenWindow` from `cPlayer` crashes the server.

Should we allow passing nil? Closing a open window, if any.

Seadragon91:
Fixes #3727 

Also fixed a crash, if the player has a open window that has been created by a plugin. Fixed by closing window from all players on command reload.

### DIFF
--- a/src/Entities/Player.h
+++ b/src/Entities/Player.h
@@ -229,7 +229,7 @@ public:
 	// tolua_begin
 
 	/** Opens the specified window; closes the current one first using CloseWindow() */
-	void OpenWindow(cWindow * a_Window);
+	void OpenWindow(cWindow & a_Window);
 
 	/** Closes the current window, resets current window to m_InventoryWindow. A plugin may refuse the closing if a_CanRefuse is true */
 	void CloseWindow(bool a_CanRefuse = true);
--- a/src/Bindings/LuaWindow.cpp
+++ b/src/Bindings/LuaWindow.cpp
@@ -4,8 +4,10 @@
 
 #include "Globals.h"
 #include "LuaWindow.h"
+#include "../Entities/Player.h"
 #include "../UI/SlotArea.h"
 #include "PluginLua.h"
+#include "Root.h"
 #include "lua/src/lauxlib.h"  // Needed for LUA_REFNIL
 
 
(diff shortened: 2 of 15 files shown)


madmaxoft:
I think a better solution would be:
 - Check in `OpenWindow` C++ code, if the passed object is a `nullptr` (what to do then? close current?)
 - Modify the `OpenWindow` C++ signature to use a reference rather than a pointer, tolua generates proper bindings then, including the `nil` check.

As for closing all the windows, that should not be necessary, the window object should be freed by Lua's GC and that should close the windows client-side. Definitely this is the wrong kind of fix - fixing the result, rather than the cause.

Seadragon91:
For the lua window I read in [LuaWindow](https://github.com/cuberite/cuberite/blob/master/src/Bindings/LuaWindow.cpp) that it will keep a ref to itself to avoid being removed by GC. This ref will be available as long there is a player that has the window open.

madmaxoft:
It will keep a **Lua** reference. That means that Lua will still GC it when the `cLuaState` is closed. Perhaps the only problem is that the `cPlayer` objects aren't notified when that happens (I don't know, needs testing). So the closing would be better suited into the `cLuaWindow` destructor.

Seadragon91:
Reworked it:
- Added overloaded method of OpenWindow accepting a ref to the window
- Closing cLuaWindow in destructor

Boths tested and works fine. Also tested if the cLuaWindow is still valid, if a ref is kept to open the same later again.

Seadragon91:
Added missing `false` (a_CanRefuse) to `CloseWindow`

Seadragon91:
Okay for merge?

src/Entities/Player.h
```diff
@@ -226,11 +226,14 @@ class cPlayer :
...
 
-	// tolua_begin
-
 	/** Opens the specified window; closes the current one first using CloseWindow() */
 	void OpenWindow(cWindow * a_Window);
 
+	// tolua_begin
+
+	/** Calls OpenWindow with the pointer of it. This is exported by tolua and it adds the nil check that we want */
+	void OpenWindow(cWindow & a_Window) { OpenWindow(&a_Window); }
```
madmaxoft: Why not remove the pointer-based version and just keep this one?

Seadragon91: Because of additional 10 line changes, change pointer to ref in all calls, I can change it, if you want :)

Seadragon91: Fixed all calls now, just search and replace :)

src/Entities/Player.h
```diff
@@ -228,8 +228,8 @@ class cPlayer :
 
 	// tolua_begin
 
-	/** Opens the specified window; closes the current one first using CloseWindow() */
-	void OpenWindow(cWindow * a_Window);
+	/** Opens the specified window; closes the current one first using CloseWindow(). This is exported by tolua and it adds the nil check we want */
```
madmaxoft: I don't think the extra comment is necessary. It's something you've just fixed and no-one will be interested in it anymore. Export to Lua is visible because of the `tolua` comment just above.

Seadragon91: Removed it.

PR #3739, opened 2017-05-28
Avoid pickups to sink into blocks and disappear

### LINKED ISSUE #3729: Pickups move erratically, hide inside floor
Client version:   1.11.2
Server OS:        Debian
Commit id:        84bdba

### Expected behavior
Pickups falling on the floor should just rest on it...

...but frequently happens that they tend to rest on the block below, trespassing the one where they should lay on. Apparently the pickup becomes invisible but if you look carefully when a block becomes a pickup, it just falls and "sinks" in the top block until it falls to the underlying block. 
If you destroy the block where the  drop is hidden you will see that both (the new one and the old one) drops gets together and sink again.

At the very bottom of a chunk if you try to replicate this behavior, you will see drops dancing, like if they couldn't find a stable floor to rest on. They sink down a bit but they suddenly go up, over and over again.

I didn't have a look at the code yet, but it looks like the typical collision detection problem usually found in games in development.

(edited by madmaxoft to fix naming conventions / typos)

spekdrum:
This fixes #3729.

The problem was that sometimes pickups never rest (getting speed 0). This is needed in order to send the appropriate message to client in cEntity::BroadcastMovementUpdate. In addition, in this last function, independently from sending speed 0 to client, if a difference from last position was detected this difference was sent to client again (resulting in the weird behaviour on pickups).
I had to use the m_bHasSentNoSpeed flag to control this and allow pickups inside solid blocks to get out.

I also added an experimental code to show the animation of pickups getting together. It does not always work, but it doesn't hurt and its cool when it works  :P


src/Entities/Entity.cpp
```diff
@@ -1134,6 +1135,14 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 				m_bOnGround = true;
 				NextPos.y = HitBlockCoords.y + 1;
+
+				// If this is a pickup and it hits the ground then stop it
+				if (GetEntityType() == eEntityType::etPickup)
+				{
+					NextSpeed.x = 0.0;
+					NextSpeed.y = 0.0;
+					NextSpeed.z = 0.0;
+				}
```
madmaxoft: I really don't like having specific code in the generic function. Wouldn't it be possible to move this into `cPickup`'s override of the `Tick()` method?

spekdrum: Sure, you are right, mutch better this way. I fix it tonight.

spekdrum: I've just removed this code, I did some tests and it's no longer needed.

Seadragon91:
@spekdrum 
There is no need to merge the master into it. We can rebase it on master or squash it for multiple commits.

Seadragon91:
Tested and yes now the pickups doesn't sink anymore in the ground, much better now thanks :)

src/Entities/Pickup.cpp
```diff
@@ -159,7 +165,7 @@ void cPickup::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 			}
 
 			// Try to combine the pickup with adjacent same-item pickups:
-			if ((m_Item.m_ItemCount < m_Item.GetMaxStackSize()))  // Don't combine if already full
+			if ((m_Item.m_ItemCount < m_Item.GetMaxStackSize()) && this->IsOnGround())  // Don't combine if already full or not on ground
```
madmaxoft: Please remove the `this->` part, it's not needed.

spekdrum: Oops, done

Seadragon91: @madmaxoft Just asking what's the problem with `this`?

madmaxoft: There's no problem, it's just unnecessary, and slightly distracting ("why is it there and no-where else? What did the dev want to say by that?")

spekdrum: I agree

PR #3741, opened 2017-05-29
Double chest fixes

peterbell10:
This fixes 3 issues I noticed with double chests:

- Normal and trapped chests next to each other would open a double chest window
- Slot changes in the secondary chest were not being broadcast
- Placing a chest in +x of another did not update the original chest's metadata

### DIFF
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -127,21 +127,28 @@ void cChestEntity::ScanNeighbours()
 	{
 	public:
 		cChestEntity * m_Neighbour;
+		BLOCKTYPE      m_ChestType;
 
-		cFindNeighbour() :
-			m_Neighbour(nullptr)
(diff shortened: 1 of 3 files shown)


src/BlockEntities/ChestEntity.cpp
```diff
@@ -127,21 +127,28 @@ void cChestEntity::ScanNeighbours()
...
+			m_Neighbour(nullptr),
+			m_ChestType(a_ChestType)
 		{
 		}
 
 		virtual bool Item(cChestEntity * a_Chest) override
 		{
+			if (a_Chest->GetBlockType() != m_ChestType)
+			{
+				// Don't neighbour with the wrong type of chest
```
Seadragon91: Please change this, maybe: `Neighbor block is not from same chest type`

Seadragon91:
Tested and works fine, improve the comment and it can be merged,

peterbell10:
Alright, I've changed the comment.

Issue #2975, opened 2016-02-05
Wearing armors does not reduce damage when being attacked by mobs


dogtopus:
Also the armor durability does not reduce after the attack

PR #3743, opened 2017-05-31
Added armor durability reduction when player is attacked. (#2301)

### LINKED ISSUE #2301: Armor does not reduce durability
Hitting another player with a weapon won't get his armor durability reduced.

Also tested empty-handed, health only reduces and not armor.

lkolbly:
What it says on the tin. Reduces the durability of each piece of armor by max(floor(RawDamage/4),1), as per http://minecraft.gamepedia.com/Armor

Also adds the max damage for each type of armor.

Ironically, I came across this while building a BedWars plugin, for which I don't actually want this functionality. :-P But it seemed like a good thing to fix anyway, and I'll just deal with it in my plugin.

Also I'll be leaving the country in a couple of days, so I won't be very responsive, but I'll be back at the end of June. Sorry for the drive-by PR, but I had the changes and didn't want to forget that I did them by the time I come back.

Fixes #2301, which is a part of #2975


src/Item.cpp
```diff
@@ -68,6 +68,27 @@ short cItem::GetMaxDamage(void) const
...
+		case E_ITEM_IRON_LEGGINGS:   return 226;
+		case E_ITEM_IRON_BOOTS:      return 196;
+		case E_ITEM_CHAIN_HELMET:    return 166;
+		case E_ITEM_CHAIN_CHESTPLATE:return 241;
+		case E_ITEM_CHAIN_LEGGINGS:  return 226;
+		case E_ITEM_CHAIN_BOOTS:     return 196;
+		case E_ITEM_DIAMOND_HELMET:  return 364;
+		case E_ITEM_DIAMOND_CHESTPLATE: return 529;
+		case E_ITEM_DIAMOND_LEGGINGS:return 496;
+		case E_ITEM_DIAMOND_BOOTS:   return 430;
```
madmaxoft: Please keep these alpha-sorted (and sort into already existing).

madmaxoft:
Other than the sorting issue, this looks good. Thanks.

lkolbly:
@madmaxoft Okay, I sorted them in and then rebased it.

Issue #2667, opened 2015-11-20
Experience orbs orbit player forever


LogicParrot:
On some rare occasions, when touching an experience orb, it is not picked. Instead, it orbits the player until they logout.

Schwertspize:
it happens in Vanilla too, some times (I only tried single player)

SafwatHalaby:
Thanks. So this might be a client side issue.

tigerw:
The server doesn't exactly replicate the behaviour of experience orbs, and the server certainly doesn't actually move them with as much grace as they should, so I would say most of the time the orbiting is due to the orbs not actually reaching the player serverside.

madmaxoft:
The server doesn't move the orbs towards nearby players, the client assumes it does, and moves the orb, until it's close enough that the client just assumes it cal collect the orb, while it's still too far for the server. This does indeed sometime happen in vanilla as well, though not that often.

So this issue boils down to implementing the movement of the orbs towards the nearest player.

tigerw:
I'm fairly certain STR or someone implemented that though.

NiLSPACE:
Yeah, I was the one who implemented experience orbs. I was never able to recreate the physics from vanilla though. Because of that sometimes the experience orb starts moving for the client, but server doesn't think it does resulting in this bug.

NiLSPACE:
A quick fix is to make an experience orb broadcast it's position once in a while. For example every 100 ticks or so. Of course we still have to implement the proper physics code eventually.

PR #3744, opened 2017-06-03
Fix experience orbs so they can be collected

### LINKED ISSUE #3740: Experience orbs cannot be collected
Client version:   1.11.2
Server OS:        Windows
Commit id:        bc6e2100c1cb5c695856e2a86afd202dec715967

### Expected behavior
Experience orbs should be able to be collected. This was working on recent versions.

### Actual behavior
They fly arround the player with no effect.

Kill any monster in survival mode and go to get the experience orb.

spekdrum:
Fixes #3740 (maybe #2667 too), adds a performance improvement and makes orbs animation smoother.


Seadragon91:
Like the changes and yes #2667 can be closed, too. Now the orbs doesn't flying a long time around the player.

PR #3745, opened 2017-06-03
Mitigates invisible entities issue #3696

### LINKED ISSUE #3696: Entitites turn invisible

spekdrum:
Changes:
- Spawn mobs in a chunk to player when player enters the chunk 
- Spawn mob to player when it sees the player
- Corrected a protocol send bytes order for mob spawn yaw, pitch and head-yaw (according to http://wiki.vg/Protocol#Spawn_Mob)

This mostly fixes #3696. It's not perfect because sometimes, when moving, you can see some mob popping out, but it's better than the alternative, that is you couldn't see them. 

At least with this my sheeps never disappear :P


spekdrum:
To test this I suggest doing the following:
- Go to survival mode
- Set time to 14000
- Wait until some mobs go for you
- Switch to creative
- Fly up a bit and put some distance from the mobs until you don't see them (not very far, just let them disappear and stop)
- Go again to mobs position.

Without these changes mobs are invisible ~80% times. With them they are visible almost always. There is a chance that they are still hidden. I'm still investigating. In survival mode there is no problem, they will pop up as soon as they see you.

Animals work in other way, actually they do not see you (if I'm not wrong). With this PR only when you walk in their chunk they will pop up (if they were hidden). I wonder if spawn mobs to player in a radius of chunks (instead of just one) could be the solution.  I also wonder if this would be any problem in terms of spamming too much packets to client.

Seadragon91:
Thanks for the explanation, was still breaking my head how to TEST this.

tigerw:
I expect that the client is taking it upon themselves to unload the mobiles/entities once it thinks they are far away (or some other condition)? I don't recall the server being *that* fancy.

There must be some way to generalise this taking into account view distance, chunks sent to the player and such.

Pokechu22:
> I expect that the client is taking it upon themselves to unload the mobiles/entities once it thinks they are far away (or some other condition)? I don't recall the server being that fancy.

The client doesn't do that.  It does stop drawing them when you're too far away (with different distances for the types of entities), but it resumes drawing them if you get closer, and doesn't "forget" about the entities.  The server is responsible for unloading the entities when the player is too far.

An exception is entities entering (or being spawned in?) unloaded chunks, in which case Weird Stuff&trade; happens.

madmaxoft:
So I suppose that this ultimately makes the client overflow with duplicate entities? Not the right solution. We should find out why the entities are invisible in the first place.

spekdrum:
I agree, this is just a temporal fix. But apart from not being the appropriate solution, we don't have a better one right now. This patch works and makes Cuberite more playable on survival games. It's really annoying without it.
I think we could add this but with a big FIXME before every code block and a linked issue explaining the problem. I can take care of this.

madmaxoft:
Alright, although I'm really not happy about this, I guess it's better than nothing. Just a note - you probably already saw how long some of those TODOs and FIXMEs stay in the code. It's sometimes better to avoid the bad fix and keep the non-working code so that it forces someone to finally look at it thoroughly and come up with a proper fix, than half-fix the issue and have it lurk in the shadowy depths.

tigerw:
What lies in the shadowy depths

> shadowy depths

Pokechu22:
You could check if entities actually do build up via the `E:` reading in F3.  It might be that they are building up, or might be that they don't; would help identify other bugs.

spekdrum:
@madmaxoft I agree with your point. "_Nothing lasts longer than a provisional solution_"
Another option is to leave this pull request here as a workaround / start point for the definitive fix.

LogicParrot:
>    Fly up a bit and put some distance from the mobs until you don't see them (not very far, just let them disappear and stop)
>  Go again to mobs position.
> Without these changes mobs are invisible ~80% times. 

This definitely did not happen before. I have a huge backlog to catch up with but I'll try to find out what's going on. 

Maybe this is a case of healthy mob despawns? I remember implementing despawns for hostiles only.

LogicParrot:
>I expect that the client is taking it upon themselves to unload the mobiles/entities once it thinks they are far away (or some other condition)? I don't recall the server being that fancy.

I did implement hostile mob despawns past a certain distance.

LogicParrot:
is there a chance this PR produces false images of mobs already despawned on the server? It's an absolute guess and I'm not up to date with the code but it's something worth testing!

LogicParrot:
I am almost certain hostile mobs despawn whenever a chunk unloads.

LogicParrot:
Another point worth investigating: In the past, there was a bug where the player would get hit by seemingly invisible mobs. It turned out the mobs were actually visible, but were locked in attack mode and attack remotely. Killing those mobs would stop the attacks.

spekdrum:
> This definitely did not happen before. I have a huge backlog to catch up with but I'll try to find out what's going on.
@LogicParrot Mobs that ever targeted a player don't get destroyed since this commit 2359611c57edc70f68b1ba05bb87dbf9c3bbc11d

spekdrum:
> is there a chance this PR produces false images of mobs already despawned on the server? It's an absolute guess and I'm not up to date with the code but it's something worth testing!

I don't think so, because spawns are always done from an existing entity (mob).

spekdrum:
> I am almost certain hostile mobs despawn whenever a chunk unloads.

I also suspect that, and was the reason why I did this approach for this issue.
But this is more complex thant that. In my tests, many times when I just login I couldn't see my sheeps. With this fix sheeps are still invisible until you get into the chunk were they belong.

LogicParrot:
This problem did not exist before. We need to pinpoint the problem commit, perhaps through git bisect. Is there a chance https://github.com/cuberite/cuberite/commit/2359611c57edc70f68b1ba05bb87dbf9c3bbc11d keeps the mobs but sends false despawn packets?

LogicParrot:
>keeps the mobs but sends false despawn packets?

Seems unlikely with such a simple change.

mathiascode:
I think this issue might've started happening after commit c89b20695df7d2a7afc85806d147e0b9ab1a23d1 which introduced working 1.11 mobs. Using Minecraft 1.11.2 with MultiMC launcher, I get this error multiple times in the launcher log:

```
[15:52:49] [Client thread/ERROR]: Couldn't render entity
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Boolean
	at zm.di(SourceFile:266) ~[1.11.2.jar:?]
	at bml.a(SourceFile:48) ~[1.11.2.jar:?]
	at bvl.a(SourceFile:141) [1.11.2.jar:?]
	at bvq.a(SourceFile:41) [1.11.2.jar:?]
	at bvq.a(SourceFile:16) [1.11.2.jar:?]
	at bur.a(SourceFile:359) [1.11.2.jar:?]
	at bur.a(SourceFile:346) [1.11.2.jar:?]
	at bqm.a(SourceFile:604) [1.11.2.jar:?]
	at bqe.a(SourceFile:1282) [1.11.2.jar:?]
	at bqe.b(SourceFile:1187) [1.11.2.jar:?]
	at bqe.a(SourceFile:1032) [1.11.2.jar:?]
	at bes.av(SourceFile:993) [1.11.2.jar:?]
	at bes.a(SourceFile:398) [1.11.2.jar:?]
	at net.minecraft.client.main.Main.main(SourceFile:124) [1.11.2.jar:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_121-solus]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_121-solus]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_121-solus]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_121-solus]
	at org.multimc.onesix.OneSixLauncher.launchWithMainClass(OneSixLauncher.java:236) [NewLaunch.jar:?]
	at org.multimc.onesix.OneSixLauncher.launch(OneSixLauncher.java:297) [NewLaunch.jar:?]
	at org.multimc.EntryPoint.listen(EntryPoint.java:162) [NewLaunch.jar:?]
	at org.multimc.EntryPoint.main(EntryPoint.java:53) [NewLaunch.jar:?]
```

mathiascode:
I'm having trouble reproducing this issue in the latest version of Cuberite. Does it still happen to you?

Pokechu22:
... Great, more broken entity metadata?  That's skeleton meta, specifically the field that was "is swinging arms" (which actually does some more stuff (drawing bows), but it looks like I haven't updated the wiki yet).

Thing is, I don't see code that's writing actual meta for skeletons (though I think variants might still be broken).

mathiascode:
Oh no, I'm on the wrong track. That error message was fixed by you in a later commit in March, and is not related to the new issue at all.

Back to square one...

LogicParrot:
I wonder if this is reproducible in 1.8. That's the last protocol I tested, and there are were no protocol issues with entities.

LogicParrot:
This = mob invisibility

LogicParrot:
>Fly up a bit and put some distance from the mobs until you don't see them (not very far, just let them disappear and stop)

How did you fly away when testing this? If you flew away by switching to creative, then the mobs de-targeted you, allowing them to despawn.

LogicParrot:
I am not able to reproduce this (barring the exception mentioned in #3721 where the mob is actually destroyed forever and is not actually invisible). Are there any factors not mentioned in your reproduction steps @spekdrum ?

spekdrum:
I will try to reproduce it again and tell you, there has been many changes since this. I missed the part that would read something like "Make a walled square to prevent mobs to leave your test area when you go far" when I wrote those steps. Maybe this helps.

LogicParrot:
>"Make a walled square to prevent mobs to leave your test area when you go far"

The mobs aren't escaping. I did wall them (and in the process I discovered https://github.com/cuberite/WorldEdit/issues/131 by chance)

CyrusTheVirusG:
mob invisibility may be due to protocol error

CyrusTheVirusG:
http://wiki.vg/Protocol#Spawn_Mob

We're sending an extra UUID

bearbin:
I'm closing this as it has been more than 6 months since it was opened, and there has been no activity recently.

PR #3751, opened 2017-06-08
Fixed skylight going down through transparent blocks.

### LINKED ISSUE #3746: light bug
Client version:   1.11.2/1.9
Server OS:        Linux

When I start the server the blocks are dark even if it's the day, and when I place a block the light comes back around it.
It's maybe because of the walls (they go to the 26 th layer).
a screen :
https://image.noelshack.com/fichiers/2017/22/1496580946-2017-06-04-14-42-54.png

I hope I have been clear, and someone can help me.

madmaxoft:
Fixes #3746.

Tested by commenting out the loader's light validation, thus enforcing a relight on all loaded chunks, then using a setup similar to #3746 - walls of stone and a ceiling of glass blocks.
Before this fix:
before

After this fix:
after

Also checked a single-block-wide vertical tunnel filled with torches, also got correct lighting by this fix.

### DIFF
--- a/src/LightingThread.cpp
+++ b/src/LightingThread.cpp
@@ -368,7 +368,18 @@ void cLightingThread::PrepareSkyLight(void)
 		for (int x = 1; x < cChunkDef::Width * 3 - 1; x++)
 		{
 			int idx = BaseZ + x;
-			int Current   = m_HeightMap[idx] + 1;
+			// Find the lowest block in this column that receives full sunlight (go through transparent blocks):
+			int Current = m_HeightMap[idx];
+			ASSERT(Current < cChunkDef::Height);
+			while (
+				(Current >= 0) &&
+				cBlockInfo::IsTransparent(m_BlockTypes[idx + Current * BlocksPerYLayer])
+			)
+			{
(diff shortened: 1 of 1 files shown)


madmaxoft:
Wait before merging, this has a possible problem at the top of the world.

Seadragon91:
Tested this in a maze. I still have sometimes dark walls after I placed a torch, but this is corrected after a few seconds. Maybe the lighting has to do many changes?

Pictures:


After a few second it's corrected, too:

Pokechu22:
The client takes some time to rerender lighting changes across chunk borders - might be that (check in F3 to see if the value actually is unchanged, if possible)

Seadragon91:
Tested it with F3 + g - shows the chunk borders as a grahical grid. The dark areas are starting at chunk borders.

madmaxoft:
Cuberite currently doesn't update the lighting after placing blocks, that's done client-side. Nothing to do with this PR.

madmaxoft:
Note that this PR is only about direct sunlight, there's no point in testing light-emissive blocks at all, there hasn't been any change there.

PR #3754, opened 2017-06-08
FastRandom rewrite

peterbell10:
This is mainly motivated by the fact that, despite the name, `cFastRandom` is actually _slower_ than `MTRand`.  Below is the rate the two underlying random engines could produce random outputs using `operator ()` without a distribution (on my machine at least).

|                      | GCC (GiB/s) | MSVC (GiB/s) |
|---------------|-------------|--------------|
| `minstd_rand` | 1.08        | 0.28         |
| `mt19937`     | 1.57        | 1.21         |

Of course missing from this is that `mt19937` is more costly to construct.  So this switches from frequently constructed, function local `cFastRandom`s to long lived, thread local `MTRand` instances.

Additionally, differences in the interfaces of these seems to have led to a fair few off by one errors.  So, I've unified and broadened the interfaces of the two random classes into a generic engine wrapper to make them consistent and hopefully a bit simpler to use right.

### DIFF
--- a/src/Blocks/BlockPlant.h
+++ b/src/Blocks/BlockPlant.h
@@ -77,7 +77,7 @@ protected:
 	{
 		// Plant can grow if it has the required amount of light, and it passes a random chance based on surrounding blocks
 		PlantAction Action = HasEnoughLight(a_Chunk, a_RelX, a_RelY, a_RelZ);
-		if ((Action == paGrowth) && !GetRandomProvider().RandBool(1.0/GetGrowthChance(a_Chunk, a_RelX, a_RelY, a_RelZ)))
+		if ((Action == paGrowth) && !GetRandomProvider().RandBool(1.0 / GetGrowthChance(a_Chunk, a_RelX, a_RelY, a_RelZ)))
 		{
 			Action = paStay;
 		}
--- a/src/Blocks/BlockPlant.h
+++ b/src/Blocks/BlockPlant.h
@@ -75,10 +75,9 @@ protected:
 	*/
 	virtual PlantAction CanGrow(cChunk & a_Chunk, int a_RelX, int a_RelY, int a_RelZ)
 	{
-		cFastRandom rand;
 		// Plant can grow if it has the required amount of light, and it passes a random chance based on surrounding blocks
 		PlantAction Action = HasEnoughLight(a_Chunk, a_RelX, a_RelY, a_RelZ);
-		if ((Action == paGrowth) && (rand.NextInt(GetGrowthChance(a_Chunk, a_RelX, a_RelY, a_RelZ)) != 0))
+		if ((Action == paGrowth) && !GetRandomProvider().RandBool(1.0/GetGrowthChance(a_Chunk, a_RelX, a_RelY, a_RelZ)))
 		{
 			Action = paStay;
 		}
--- a/src/FastRandom.cpp
+++ b/src/FastRandom.cpp
@@ -1,7 +1,7 @@
 
 // FastRandom.cpp
 
-// Implements the cFastRandom class representing a fast random number generator
+#include <mutex>
 
 #include "Globals.h"
 #include "FastRandom.h"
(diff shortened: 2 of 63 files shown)


src/FastRandom.cpp
```diff
@@ -1,7 +1,7 @@
 
 // FastRandom.cpp
 
-// Implements the cFastRandom class representing a fast random number generator
+#include <mutex>
```
madmaxoft: This needs to go below `"Globals.h"`

src/FastRandom.cpp
```diff
@@ -13,109 +13,49 @@
 #elif defined (_MSC_VER)
 	#define ATTRIBUTE_TLS static __declspec(thread)
 #else
-	#error "Unknown thread local storage qualifier"
+	#define ATTRIBUTE_TLS thread_local
```
madmaxoft: This is wrong - you can't be sure that this is true for all other compilers. You should specify the attributes for known compilers and leave the `#else` branch intact.

src/FastRandom.cpp
```diff
@@ -13,109 +13,49 @@
...
+static std::list<std::unique_ptr<MTRand>> DeleteList;
 
 
 
 
 
-
-int cFastRandom::GenerateRandomInteger(int a_Begin, int a_End)
+/** Some compilers don't support thread_local for non-POD types, this is purely a work around for that restriction.
+There should be minimal overhead for the non-initializing case and all thread's instances are deleted properly. */
```
madmaxoft: This comment is not documentation, so it should not use the doxy-comment syntax. Also it should be inside the function, where the pointers are actually declared.

src/FastRandom.h
```diff
@@ -23,58 +16,188 @@ salts, the values they get will be different.
...
+	{
+	}
 
-	cFastRandom(void);
 
-	/** Returns a random int in the range [0 .. a_Range - 1]; a_Range must be less than 1M */
-	int NextInt(int a_Range);
+	/** Initialize with a SeedSequence. */
+	template <class SeedSeq>
+	cRandomWrapper(SeedSeq& a_SeedSeq) :
```
madmaxoft: Spaces around `&`, no space in front of `:`.

src/FastRandom.h
```diff
@@ -23,58 +16,188 @@ salts, the values they get will be different.
...
+	{
+		std::uniform_real_distribution<RealType> dist;
+		return dist(m_Engine);
+	}
+
+
+
+
+
+	/** Return a random bool with the given probabitlity of being true. */
```
madmaxoft: Fix typo

src/World.cpp
```diff
@@ -812,7 +810,7 @@ eWeather cWorld::ChooseNewWeather()
 		case eWeather_Rain:
 		{
 			// 1 / 8 chance of turning into a thunderstorm
-			return ((m_TickRand.randInt() % 256) < 32) ? eWeather_ThunderStorm : eWeather_Sunny;
+			return GetRandomProvider().RandBool(1.0 / 8) ? eWeather_ThunderStorm : eWeather_Sunny;
```
madmaxoft: I think I'd prefer a direct number, rather than this "mixed division".

madmaxoft:
I like this idea. Fix the small problems and it's good to go.

src/FastRandom.cpp
peterbell10: It seemed reasonable to have standard C++ as the default but I'll revert it if you think that's wrong.

madmaxoft: Oh, it's already standard. Then we should probably remove the `#ifdef` altogether.

peterbell10: Seems like msvc on appveyor is failing with `thread_local`.  It has been standard since C++11 but that compiler is fairly old.

peterbell10:
@madmaxoft I think that's most of the little issues but I'm not sure what the stance is on disabling warnings.  Any thoughts?

src/Blocks/BlockPlant.h
```diff
@@ -75,10 +75,9 @@ class cBlockPlant : public cBlockHandler
 	*/
 	virtual PlantAction CanGrow(cChunk & a_Chunk, int a_RelX, int a_RelY, int a_RelZ)
 	{
-		cFastRandom rand;
 		// Plant can grow if it has the required amount of light, and it passes a random chance based on surrounding blocks
 		PlantAction Action = HasEnoughLight(a_Chunk, a_RelX, a_RelY, a_RelZ);
-		if ((Action == paGrowth) && (rand.NextInt(GetGrowthChance(a_Chunk, a_RelX, a_RelY, a_RelZ)) != 0))
+		if ((Action == paGrowth) && !GetRandomProvider().RandBool(1.0/GetGrowthChance(a_Chunk, a_RelX, a_RelY, a_RelZ)))
```
madmaxoft: Spaces around the `/`

src/FastRandom.cpp
```diff
@@ -1,121 +1,66 @@
...
 #include "Globals.h"
 #include "FastRandom.h"
 
+#include <mutex>
 #include <random>
 
+#ifdef __clang__
+	#pragma clang diagnostic ignored "-Wglobal-constructors"
+#endif
+
```
madmaxoft: Usually we try to fix whatever the warnings are hinting at, unless the fix is too unwieldy. Better than disabling the warning on a whole file where it could actually hide something important.

In this case, wouldn't it work if you refactored the two globals into `static` variables inside the only function that uses them?

src/MobSpawner.cpp
```diff
@@ -202,7 +198,7 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
 					(BlockBelow == E_BLOCK_GRASS) || (BlockBelow == E_BLOCK_LEAVES) || (BlockBelow == E_BLOCK_NEW_LEAVES)
 				) &&
 				(a_RelY >= 62) &&
-				(Random.NextInt(3) != 0)
+				(Random.RandBool(2.0/3))
```
madmaxoft: Spaces around '/'.

src/Mobs/Monster.cpp
```diff
@@ -1125,11 +1124,10 @@ void cMonster::AddRandomUncommonDropItem(cItems & a_Drops, float a_Chance, short
 
 void cMonster::AddRandomRareDropItem(cItems & a_Drops, cItems & a_Items, unsigned int a_LootingLevel)
 {
-	MTRand r1;
-	unsigned int Count = static_cast<unsigned int>(static_cast<unsigned long>(r1.randInt()) % 200);
-	if (Count < (5 + a_LootingLevel))
+	auto & r1 = GetRandomProvider();
+	if (r1.RandBool((5 + a_LootingLevel)/200.0))
```
madmaxoft: Spaces around '/'.

src/Mobs/Sheep.cpp
```diff
@@ -121,7 +121,7 @@ void cSheep::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	}
 	else
 	{
-		if (m_World->GetTickRandomNumber(600) == 1)
+		if (GetRandomProvider().RandBool(1.0/600.0))
```
madmaxoft: Spaces

src/Mobs/Villager.cpp
```diff
@@ -32,7 +32,7 @@ bool cVillager::DoTakeDamage(TakeDamageInfo & a_TDI)
 
 	if ((a_TDI.Attacker != nullptr) && a_TDI.Attacker->IsPlayer())
 	{
-		if (m_World->GetTickRandomNumber(5) == 3)
+		if (GetRandomProvider().RandBool(1.0/6.0))
```
madmaxoft: Spaces

src/FastRandom.cpp
peterbell10: Yes, It just means that every call to `GetRandomProvider` has to first check if they are initialised.
Probably not an issue.

peterbell10:
All the spaces are in and I think I've found all the ones you didn't point out as well.

src/FastRandom.cpp
madmaxoft: Okay, let's keep it this way, then.

madmaxoft:
Thanks. Looks good to me, let's see if anyone else disagrees or squash-merges :)

Bond-009:
This PR also doesn't compile on macOS with clang

```
cuberite/src/MobSpawner.cpp:115:41: error: implicit conversion changes signedness:
      'unsigned long' to 'typename iterator_traits<__tree_const_iterator<eMonsterType, __tree_node<eMonsterType, void *> *,
      long> >::difference_type' (aka 'long') [-Werror,-Wsign-conversion]
                std::advance(itr, GetRandomProvider().RandInt<size_t>(allowedMobsSize - 1));
```

madmaxoft:
We really need to include macOS clang in CI tests :)

mathiascode:
FreeBSD builds are failing with the same error. https://cuberite.org/status/

PR #3758, opened 2017-06-10
Corrected check for level of subcommand and fixed multiple levels not working

### LINKED ISSUE #3759: Subcommands that contains subcommands
I am not sure if something like that make sense.

Example:
Command and sub commands:
`/cmd` - no handler
`/cmd s1` - has a handler
`/cmd s2` - has a handler

Would it make sense if a sub command itself can have another sub command that has a handler?:
`/cmd s1 x` - has a handler

I tested this and the handler for `x` is not reachable, the handler from `s1` will be called instead.

Seadragon91:
Fixes cuberite/WorldEdit#110
Fixes #3759

Now the server will print for command `/brush a`:
`The /brush command doesn't support verb a`
instead of the error that something was gone wrong.

### DIFF
--- a/Server/Plugins/InfoReg.lua
+++ b/Server/Plugins/InfoReg.lua
@@ -59,7 +59,7 @@ local function MultiCommandHandler(a_Split, a_Player, a_CmdString, a_CmdInfo, a_
 	-- A verb was specified, look it up in the subcommands table:
 	local Subcommand = a_CmdInfo.Subcommands[Verb]
 	if (Subcommand == nil) then
-		if (a_Level > 1) then
+		if (a_Level + 1 > 1) then
 			-- This is a true subcommand, display the message and make MCS think the command was handled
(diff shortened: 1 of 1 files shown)


Server/Plugins/InfoReg.lua
```diff
@@ -59,7 +59,7 @@ local function MultiCommandHandler(a_Split, a_Player, a_CmdString, a_CmdInfo, a_
 	-- A verb was specified, look it up in the subcommands table:
 	local Subcommand = a_CmdInfo.Subcommands[Verb]
 	if (Subcommand == nil) then
-		if (a_Level > 1) then
+		if (a_Level >= 1) then
```
Seadragon91: I think `a_Level + 1` makes more sense, is better to understand. Also above is used `a_Level + 1`.

Seadragon91:
Fixed multiple level of sub commands not working.

PR #3763, opened 2017-06-13
Double tall grass drops seeds when broken by water

### LINKED ISSUE #3673: Double tall grass doesn't drop the correct item when broken by water
Client version:   1.11.2
Server OS:         macOS
Commit id:         Unknown

### Expected behavior
It should drop seeds.

### Actual behavior
It drops double plants with meta 4-7. 6-7 are non-existing metas
screen shot 2017-04-18 at 18 13 07

Break double tall grass with water.

dbhicks:
This PR resolves #3673.

Some debugging showed that double tall flower blocks, when broken with water, would trigger DropBlock() for the block at ground/water level and then trigger DropBlock() **_again_** for the block one block above ground/water level. When the top half of a tall flower is broken, it attempts to take the tall flower meta value from the block below it, but if that block was first broken (as it is in the case of flooding) this meta value is meaningless. This was resulting in out of range meta values that didn't make sense with regard to tall flower blocks.

In addition to making tall grass blocks drop seeds, this fix also ensures that when the top half of a tall flower triggers DropBlock(), the block below must also be a tall flower block before calling super::DropBlock().


Seadragon91:
Dig a hole 2 block deep, place plant in it and place water in the near. The plant will be broken and a item is dropped twice.

dbhicks:
This appears to have been the exact same problem in reverse. Adding a second check fixes it for me.

Seadragon91:
Thanks works now fine.

src/Blocks/BlockBigFlower.h
```diff
@@ -28,24 +28,31 @@ class cBlockBigFlowerHandler :
 		NIBBLETYPE Meta = a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ);
 		if (Meta & 0x8)
 		{
-			super::DropBlock(a_ChunkInterface, a_WorldInterface, a_BlockPluginInterface, a_Digger, a_BlockX, a_BlockY - 1, a_BlockZ, a_CanDrop);
+			if (a_ChunkInterface.GetBlock(a_BlockX, a_BlockY - 1, a_BlockZ) == E_BLOCK_BIG_FLOWER)
```
madmaxoft: You need to check if `a_BlockY - 1` is inside the world height (`cChunkDef::IsValidHeight()`)

src/Blocks/BlockBigFlower.h
```diff
@@ -28,24 +28,31 @@ class cBlockBigFlowerHandler :
...
+			{
+				super::DropBlock(a_ChunkInterface, a_WorldInterface, a_BlockPluginInterface, a_Digger, a_BlockX, a_BlockY - 1, a_BlockZ, a_CanDrop);
+			}
 		}
 		else
 		{
-			super::DropBlock(a_ChunkInterface, a_WorldInterface, a_BlockPluginInterface, a_Digger, a_BlockX, a_BlockY, a_BlockZ, a_CanDrop);
+			if (a_ChunkInterface.GetBlock(a_BlockX, a_BlockY + 1, a_BlockZ) == E_BLOCK_BIG_FLOWER)
```
madmaxoft: Same - check the Y coord

madmaxoft:
Wouldn't it make more sense to actually check in `DropBlock()` and destroy the entire big flower at once, rather than keeping the "orphaned" second block to be destroyed later?

dbhicks:
I think you're absolutely right that simply removing the other block directly is a better approach, but I was a little leery of changing too much from the original function. Here's an alternate version that calls super::DropBlock on the affected flower half, replacing the other half with air.

src/Blocks/BlockBigFlower.h
```diff
@@ -26,26 +26,35 @@ class cBlockBigFlowerHandler :
...
 		{
-			super::DropBlock(a_ChunkInterface, a_WorldInterface, a_BlockPluginInterface, a_Digger, a_BlockX, a_BlockY - 1, a_BlockZ, a_CanDrop);
+			--AlternateY;
 		}
 		else
 		{
+			++AlternateY;
+		}
+
+		if (cChunkDef::IsValidHeight(AlternateY) && a_ChunkInterface.GetBlock(a_BlockX, AlternateY, a_BlockZ) == E_BLOCK_BIG_FLOWER)
```
madmaxoft: This could use a comment - `// Also destroy the other block of the big flower` for example

madmaxoft: Ideally, put the comment before the `if` above and remove this empty line, it's all one logical block.

madmaxoft:
That looks so much better. Just make sure you've tested it whether it works :)

dbhicks:
Comment added; tested by flooding tall grass and other big flowers, both at ground level and from above. Both work as expected: flowers drop one flower each, tall grass drops seeds.

Seadragon91:
I tested it and it works :)

madmaxoft:
Great, thanks :)

PR #3766, opened 2017-06-14
Replace evdns with getaddrinfo and getnameinfo

### LINKED ISSUE #3574: Remove EvDNS usage in favor of synchronous getaddrinfo() etc.
Since the EvDNS is unstable and prone to errors (#2418 and others), we should stop using it and instead use the OS-provided `getaddrinfo` and `getnameinfo` API calls.

This will need changes in the source - the OS-provided APIs are synchronous, while we expect asynchronous-with-callback. The easiest way to convert would be to have a separate background thread (multiple threads?) for the queries, with an input queue. This will need to be adapted for the `cNetwork:HostnameToIP` and `cNetwork:IPToHostname` LuaAPIs, as well as our internal `cTCPLinkImpl::Connect()` implementation.

peterbell10:
Closes #3574

### DIFF
--- a/src/OSSupport/GetAddressInfoError.h
+++ b/src/OSSupport/GetAddressInfoError.h
@@ -27,3 +27,4 @@ inline AString ErrorString(int a_ErrorCode)
 		return gai_strerror(a_ErrorCode);
 	#endif
 }
+
--- a/src/OSSupport/GetAddressInfoError.h
+++ b/src/OSSupport/GetAddressInfoError.h
@@ -1,0 +1,29 @@
+#pragma once
+
+
+
+/** Returns the readable form of a getaddressinfo type error code */
+inline AString ErrorString(int a_ErrorCode)
+{
+	// Note gai_strerror is not threadsafe on windows
+	#ifdef _WIN32
+		AString ErrorStr(GAI_STRERROR_BUFFER_SIZE, '\0');
+
+		int MsgLen = FormatMessageA(
+			FORMAT_MESSAGE_FROM_SYSTEM |
+			FORMAT_MESSAGE_IGNORE_INSERTS |
+			FORMAT_MESSAGE_MAX_WIDTH_MASK,
+			nullptr,
+			a_ErrorCode,
+			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+			&ErrorStr[0],
+			GAI_STRERROR_BUFFER_SIZE,
+			nullptr
+		);
+
+		ErrorStr.resize(MsgLen);
+		return ErrorStr;
+	#else
+		return gai_strerror(a_ErrorCode);
+	#endif
+}
--- a/src/OSSupport/GetAddressInfoError.h
+++ b/src/OSSupport/GetAddressInfoError.h
@@ -7,7 +7,7 @@ inline AString ErrorString(int a_ErrorCode)
 {
 	// Note gai_strerror is not threadsafe on windows
 	#ifdef _WIN32
-		AString ErrorStr(GAI_STRERROR_BUFFER_SIZE, '\0');
+		char ErrorStr[GAI_STRERROR_BUFFER_SIZE + 1];
 
 		int MsgLen = FormatMessageA(
 			FORMAT_MESSAGE_FROM_SYSTEM |
@@ -16,13 +16,12 @@ inline AString ErrorString(int a_ErrorCode)
 			nullptr,
 			a_ErrorCode,
 			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
-			&ErrorStr[0],
+			ErrorStr,
 			GAI_STRERROR_BUFFER_SIZE,
 			nullptr
 		);
 
-		ErrorStr.resize(MsgLen);
-		return ErrorStr;
+		return AString(ErrorStr, MsgLen);
 	#else
 		return gai_strerror(a_ErrorCode);
 	#endif
--- a/src/OSSupport/GetAddressInfoError.h
+++ b/src/OSSupport/GetAddressInfoError.h
@@ -17,7 +17,7 @@ inline AString ErrorString(int a_ErrorCode)
 			a_ErrorCode,
 			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
 			ErrorStr,
-			GAI_STRERROR_BUFFER_SIZE,
+			sizeof(ErrorStr) - 1,
 			nullptr
 		);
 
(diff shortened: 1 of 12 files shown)


src/OSSupport/TCPLinkImpl.cpp
```diff
@@ -70,41 +70,75 @@ cTCPLinkImplPtr cTCPLinkImpl::Connect(const AString & a_Host, UInt16 a_Port, cTC
...
+		}
+
+		virtual void OnNameResolved(const AString & a_Name, const AString & a_IP) override
+		{
+		}
+	};
+
+	// Schedule the host query
+	cNetwork::HostnameToIP(a_Host, std::make_shared<cHostnameCallback>(res, a_Port));
+	return res;
```
madmaxoft: Will this work with `a_Host` specifying an IP address, too? Because the old code did, you could specify either a hostname or an IP address. We want to keep that functionality.

tests/Network/CMakeLists.txt
```diff
@@ -12,10 +12,13 @@ set (Network_SRCS
 	${CMAKE_SOURCE_DIR}/src/OSSupport/Event.cpp
 	${CMAKE_SOURCE_DIR}/src/OSSupport/HostnameLookup.cpp
 	${CMAKE_SOURCE_DIR}/src/OSSupport/IPLookup.cpp
+	${CMAKE_SOURCE_DIR}/src/OSSupport/ISThread.cpp
```
madmaxoft: Wrong capitalization, will fail on *nix. `IsThread`

tests/Network/CMakeLists.txt
```diff
@@ -29,10 +32,13 @@ set (Network_HDRS
 	${CMAKE_SOURCE_DIR}/src/OSSupport/Event.h
 	${CMAKE_SOURCE_DIR}/src/OSSupport/HostnameLookup.h
 	${CMAKE_SOURCE_DIR}/src/OSSupport/IPLookup.h
+	${CMAKE_SOURCE_DIR}/src/OSSupport/ISThread.h
```
madmaxoft: Wrong capitalization.

madmaxoft:
Good rewrite, except a few details.

I'd really appreciate if you dropped a note to the related issues first, saying that you want to work on this and that. Sometimes there are a few thoughts that I'd like to share about what implementation I had in mind. In this case, there could've been multiple threads processing the queue, and the threads could have been universal, which would be the basis for the generic "task scheduler" framework. But whatever, for now let's keep it this way, just fix the errors :)

src/OSSupport/HostnameLookup.cpp
```diff
@@ -5,62 +5,78 @@
...
 
 
 
 
 
+static const char * ErrorString(int a_ErrorCode)
+{
+	// Avoid the windows API wide character version
+#ifdef _WIN32
+	return gai_strerrorA(a_ErrorCode);
```
madmaxoft: From MSDN:
 > Note that the gai_strerror function is not thread safe, and therefore, use of traditional Windows Sockets functions such as the WSAGetLastError function is recommended.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms738514(v=vs.85).aspx

Since there's already the `#ifdef _WIN32`, we could as well use the WSA function.

Also note that we indent even preprocessor code.

src/OSSupport/TCPLinkImpl.cpp
peterbell10: Running on NetworkTest it seems to work with IPs as well.  So I think it's good.

madmaxoft: Okay, I checked both *nix and Windows docs for the `getaddrinfo` function and they both allow IP addresses, too. So that's good.

peterbell10:
I did intentionally try to keep the `cNetworkLookup` class extremely minimal with the view to using a more generic executor in future.  It just seemed like the path of least resistance for now.

src/OSSupport/GetAddressInfoError.h
```diff
@@ -0,0 +1,29 @@
...
+		AString ErrorStr(GAI_STRERROR_BUFFER_SIZE, '\0');
+
+		int MsgLen = FormatMessageA(
+			FORMAT_MESSAGE_FROM_SYSTEM |
+			FORMAT_MESSAGE_IGNORE_INSERTS |
+			FORMAT_MESSAGE_MAX_WIDTH_MASK,
+			nullptr,
+			a_ErrorCode,
+			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+			&ErrorStr[0],
```
madmaxoft: This is a hack that is not guaranteed to be safe. Use a char buffer of (moderately large) fixed size instead.

I know we use this trick on a few places, but we try to phase it out as much as possible.

peterbell10: Is it really a hack? Looking at [cppreference](http://en.cppreference.com/w/cpp/string/basic_string) is says (emphasis mine):

>The elements of a basic_string are stored contiguously, that is, for a basic_string s, &*(s.begin() + n) == &*s.begin() + n for any n in [0, s.size()), or, equivalently, **a pointer to s[0] can be passed to functions that expect a pointer to the first element of a CharT[] array**. 

I'll change it if you wish but it seems like a good idea to save the extra copying.

madmaxoft: As far as I can tell, that doesn't guarantee that the pointer will stay valid after the statement (the assignment, or in the case of cppreference, the function call) is executed.
In my opinion it would be legal for a `std::string` implementation to return a temporary object that has a copy of the contained string including the NUL terminator, has `const char *` conversion defined on it, so that it is possible to read it, and in its destructor frees the copy of the string - it would still pass as a valid `std::string` but will not work with your (our) code. None of the implementations I know of currently do that, but it is possible for some obscure STL implementation to try it (for example to implement true copy-on-write strings, what do I know).

peterbell10: Ok, I've removed it.

src/OSSupport/GetAddressInfoError.h
```diff
@@ -0,0 +1,29 @@
...
+
+		int MsgLen = FormatMessageA(
+			FORMAT_MESSAGE_FROM_SYSTEM |
+			FORMAT_MESSAGE_IGNORE_INSERTS |
+			FORMAT_MESSAGE_MAX_WIDTH_MASK,
+			nullptr,
+			a_ErrorCode,
+			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+			ErrorStr,
+			GAI_STRERROR_BUFFER_SIZE,
```
madmaxoft: Sorry to complain about this again :)
Please use `sizeof(ErrorStr)` instead - if someone decides to modify the buffer size, the code would become unsafe.

madmaxoft:
Thanks.

Cl1608Ho:
Sorry, just saw this right now. 

Does this work on Android? If so, will it finally fix the authentication problems (due to dns) on Android or will it produce even more :P?

madmaxoft:
No idea, but I suppose it should work better now - we're using OS-provided functions rather than implementing "our own" (read: LibEvent's) DNS resolver code.

Someone will have to try it out :)

tigerw:
Regression #4257.

PR #3772, opened 2017-06-15
BlockEntities: Support cloning self.

madmaxoft:
Preliminary work for adding BlockEntities support into `cBlockArea`. This allows the code to create an exact copy of each BlockEntity subtype, via the `cBlockEntity::Clone()` function.

Opened a separate PR in order to check compilation on multiple targets and for discussion.

### DIFF
--- a/src/BlockEntities/BeaconEntity.cpp
+++ b/src/BlockEntities/BeaconEntity.cpp
@@ -11,13 +11,14 @@
 
 
 
-cBeaconEntity::cBeaconEntity(int a_BlockX, int a_BlockY, int a_BlockZ, cWorld * a_World):
-	super(E_BLOCK_BEACON, a_BlockX, a_BlockY, a_BlockZ, 1, 1, a_World),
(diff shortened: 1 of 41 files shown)


madmaxoft:
I'd like to continue building on top of this, does anyone oppose a merge?

Seadragon91:
Merge it.

Bond-009:
macOS clang in the CI tests would be nice :)

```
cuberite/src/./BlockEntities/DropSpenserEntity.h:46:15: error: 'CopyFrom'
      overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
        virtual void CopyFrom(const cBlockEntity & a_Src);
                     ^
cuberite/src/Simulator/IncrementalRedstoneSimulator/../../BlockEntities/BlockEntityWithItems.h:47:15: note: 
      overridden virtual function is here
        virtual void CopyFrom(const cBlockEntity & a_Src) override;
```

madmaxoft:
Fixed in master.

Perhaps we could get at least CI builds with a newer clang on Linux, then it could possibly detect these as well.

PR #3773, opened 2017-06-15
Entity: Replaced a mutexed counter with a std::atomic.

madmaxoft:
Fixes a warning from clang.

Also prettified one ugly one-line macro.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -19,15 +19,21 @@
 
 
 
-UInt32 cEntity::m_EntityCount = 0;
-cCriticalSection cEntity::m_CSCount;
+static UInt32 GetNextUniqueID(void)
+{
+	static std::atomic<UInt32> counter(1);
+	return counter.fetch_add(1);
+}
 
 
 
 
+
+////////////////////////////////////////////////////////////////////////////////
+// cEntity:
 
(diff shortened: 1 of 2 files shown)


src/Entities/Entity.cpp
```diff
@@ -19,15 +19,21 @@
...
 
 
 
 
+////////////////////////////////////////////////////////////////////////////////
+// cEntity:
+
 cEntity::cEntity(eEntityType a_EntityType, double a_X, double a_Y, double a_Z, double a_Width, double a_Height):
-	m_UniqueID(INVALID_ID),  // Proper ID will be assigned later in the constructor code
+	m_UniqueID(GetNextUniqueID()),
```
peterbell10: Out of curiosity, why not just `++m_EntityCount`?

NiLSPACE: It's not thread-safe. That's why there was a lock previously.

peterbell10: To be clear, I meant making `m_EntityCount` atomic and using the atomic pre-increment.

madmaxoft: Because the `m_EntityCount` would still be a de-facto global object requiring a global constructor.

peterbell10: Ah, makes sense. Thank you.

PR #3775, opened 2017-06-16
Remove reference to EAI_NODATA

peterbell10:
The [BSD build](https://cuberite.org/status/) is now failing with ```error: use of undeclared identifier 'EAI_NODATA'```
According to [this](https://sourceware.org/glibc/wiki/NameResolver#Return_value), `EAI_NODATA` is non-standard and isn't defined on FreeBSD.  Moreover, on windows it's just an alias for `EAI_NONAME` so I've changed it to that.

I'm not sure if this is the right solution though.  Having an `#ifndef` and aliasing it like on windows might be a better option.


madmaxoft:
This is good enough for us.

PR #3782, opened 2017-06-17
Fix two block flower drops

### LINKED ISSUE #3778: Double tall grass and large fern top parts drop sunflower
Client version:   1.8.9 and 1.12
Server OS:        macOS
Commit id:        0f21bda12cdb06fb2bcf35ec345fc1dab5e07c9a

### Expected behavior
When broken normally, all variants will drop 0 or 1 seeds.
http://minecraft.gamepedia.com/Grass

### Actual behavior
It drops a sunflower.

Break the top part of double tall grass or large fern.

peterbell10:
Fixes #3778

### DIFF
--- a/src/Blocks/BlockBigFlower.h
+++ b/src/Blocks/BlockBigFlower.h
@@ -58,7 +58,11 @@ public:
 
 		if (a_BlockMeta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS)
 		{
-			a_Pickups.push_back(cItem(E_ITEM_SEEDS));
+			// Double tall grass has 1 in 24 chance to drop seeds
+			if (GetRandomProvider().RandBool(1.0 / 24.0))
+			{
+				a_Pickups.push_back(cItem(E_ITEM_SEEDS));
+			}
 		}
 		else if (a_BlockMeta != E_META_BIG_FLOWER_LARGE_FERN)
 		{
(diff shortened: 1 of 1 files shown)


Seadragon91:
I tested breaking of upper and lower part with and without water and it looks good.

peterbell10:
Right, sorry.  Can confirm, I tested with the different flowers, breaking top and bottom by hand and with water.

madmaxoft:
So what's with this and #3763 ? It seems these two PRs are fighting over this code.

Bond-009:
No, #3763 fixed an issue where double plants dropped the wrong item when broken by water. (#3673)
But it created a new issue where double plants drop a sunflower when you break the top part. (#3778)

Seadragon91:
The current master work correct for me.

peterbell10:
Most of the double plants (Peony, Rose and Lilac) have the lower bits of their meta set on both blocks.  The problem is that double tallgrass and large ferns don't seem to have this meta for the top block.

So, while the top flower's meta works some of the time, the code shouldn't rely on it otherwise the grasses will drop sunflower which has lower bits all zero.

Alternatively the top block meta probably *should* have all the information, but I think then you have to dig into generators and bonemeal related stuff.

Bond-009:
Just 2 very small things, there is a 50% chance that seeds drop not 100% and large fern should drop seeds too.

dbhicks:
Interesting. When I was testing #3763, I was using big flower blocks that I placed in creative since that's easier than trying to hunt down flowers that were generated. Big flowers placed in creative seem to consistently have the correct meta values, where the flower type metas are consistent in the lower bits. I couldn't find any generated large ferns, but tall grass and sunflowers both seem to have meta 11. Breaking them doesn't yield sunflowers for me, though. They don't seem to drop anything at all for me.

peterbell10:
Closing because #3786 is probably a better solution.

PR #3785, opened 2017-06-19
Lighting Optimisations

peterbell10:
I noticed #822 and thought I'd take a look at the lighting code.  This doesn't look like much but I think the results speak for themselves.

I'm not entirely sure the best way to profile the lighting code but what I did was delete all the region folders from my worlds (leaving the seed the same) and run with a profiler during spawn prepare.  I repeated this 4 times with each branch and averaged the ratio of samples in the lighting thread to the generating thread.

image

This was done with MSVC on x86 so it would  be interesting to see how this compares on other platforms as well or if there's a better way to profile this.  I also tried compiling with gcc and ran cachegrind  on it and the results looked positive but I'm not that familiar with cachegrind so I'm not too confident about those results.

### DIFF
--- a/src/LightingThread.cpp
+++ b/src/LightingThread.cpp
@@ -252,7 +252,7 @@ void cLightingThread::LightChunk(cLightingChunkStay & a_Item)
 
 	ReadChunks(a_Item.m_ChunkX, a_Item.m_ChunkZ);
 
-	PrepareBlockLight();
+	PrepareBlockLight3();
 	CalcLight(m_BlockLight);
 
 	PrepareSkyLight();
@@ -361,6 +361,12 @@ void cLightingThread::PrepareSkyLight(void)
 	memset(m_IsSeed1, 0, sizeof(m_IsSeed1));
 	m_NumSeeds = 0;
 
+	// Fill the top of the chunk with all-light:
+	if (m_MaxHeight < cChunkDef::Height - 1)
+	{
+		std::fill(m_SkyLight + (m_MaxHeight + 1) * BlocksPerYLayer, m_SkyLight + ARRAYCOUNT(m_SkyLight), 15);
+	}
+
 	// Walk every column that has all XZ neighbors
 	for (int z = 1; z < cChunkDef::Width * 3 - 1; z++)
 	{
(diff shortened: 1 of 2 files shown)


src/LightingThread.cpp
```diff
@@ -494,6 +500,35 @@ void cLightingThread::PrepareBlockLight2(void)
 
 
 
+void cLightingThread::PrepareBlockLight3()
```
NiLSPACE: What's with the 3 in the name? Third attempt?

peterbell10: There's was already a `PrepareBlockLight2` but I mainly left it so you can easily distinguish them in profiles.  Should probably just collapse them in the end though.

NiLSPACE: Ah, I see.

madmaxoft:
So the gist of this optimization is that the blocklight preparation doesn't scan the entire chunk height, but instead uses the maximum height from the heightmap? And you get such a huge speedup from that?

peterbell10:
The main issue is the traversal in y direction which is basically a cache miss every time.  This code is much more cache and prefetch friendly.

For prepare skylight for example, the vast majority of any chunk is just a huge block of air which eventually gets set to full light using only y traversals.  Whereas this turns most of it into linear traversal.  That turns out to make a huge difference.

madmaxoft:
Makes sense. And I have no idea why the blocklight even uses the Y-first traversal. Makes sense.

I wonder what kind of difference this makes on the ARM. Is there any way you could check?

Also, any chance of splitting the `cBlockInfo`-related changes off to a separate PR? It doesn't seem essential for this and causes CI errors.

peterbell10:
It might just be weird quirk of MSVC but that change to `cBlockInfo` actually seemed to speed up the lighting code.  I'm not entirely sure why: maybe inlining?
I'll split it off anyway.

madmaxoft:
I guess that's because the compiler can generate better "magic statics" code that we can. Unfortunately it doesn't work properly (thread-safe) in MSVC until 2017, afaik.

peterbell10:
OK, it took an absolute age to compile but I tried it out on my raspberry pi.  Couldn't get a profiler working but here's the wall clock startup time after deleting region folders.
image

madmaxoft:
I still think more can be done for #822, but still, this is great progress already.

PR #3791, opened 2017-06-20
Fix shears (and vines)

Bond-009:
* Shears now lose durability when breaking any block
* Leaves don't drop 2 leaves anymore when broken by shears
* Removed the chance to drop saplings or apples when leaves are broken by shears
* Vines can't be attached to ender chests, pistons and redstone repeaters

### DIFF
--- a/src/Items/ItemShears.h
+++ b/src/Items/ItemShears.h
@@ -35,15 +35,10 @@ public:
 		if ((Block == E_BLOCK_LEAVES) || (Block == E_BLOCK_NEW_LEAVES))
 		{
 			cItems Drops;
-			cBlockHandler * Handler = cBlockInfo::GetHandler(Block);
-
-			Handler->ConvertToPickups(Drops, BlockMeta);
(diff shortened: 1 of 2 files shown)


src/Items/ItemShears.h
```diff
@@ -68,7 +63,7 @@ class cItemShearsHandler :
 
 	virtual short GetDurabilityLossByAction(eDurabilityLostAction a_Action) override
 	{
-		return 0;
+		return 1;
```
madmaxoft: Are shears used by each and every action in vanilla? I thought there were exceptions.

Pokechu22: Yes (at least, for each block break).

Perhaps you're thinking of way back when, when shears could be enchanted with silk touch and efficiency and were usable as an infinite durability shovel.  That's no longer possible.

madmaxoft: Okay, I knew there was something, but yes, let's fix that :)

PR #3794, opened 2017-06-21
Fixed memory leak in cClientHandle.

madmaxoft:
This fixes a memory leak in `cClientHandle` - the "ping" connections didn't reset their `m_Self`, causing a circular dependent `shared_ptr` and memory leaks in the end.

Additionally, this changes `cClientHandle::m_Protocol` to `unique_ptr`.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -95,7 +95,7 @@ cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
 	m_LastPlacedSign(0, -1, 0),
 	m_ProtocolVersion(0)
 {
-	m_Protocol = new cProtocolRecognizer(this);
+	m_Protocol = cpp14::make_unique<cProtocolRecognizer>(this);
 
 	s_ClientCount++;  // Not protected by CS because clients are always constructed from the same thread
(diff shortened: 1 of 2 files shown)


madmaxoft:
This may be the longest-standing memory leak we have had, possibly the one causing the Gallery server's memory usage growth. :D

src/ClientHandle.cpp
```diff
@@ -171,12 +170,13 @@ void cClientHandle::Destroy(void)
 	}
 
 	LOGD("%s: destroying client %p, \"%s\" @ %s", __FUNCTION__, static_cast<void *>(this), m_Username.c_str(), m_IPString.c_str());
+	auto player = m_Player;
+	m_Self.reset();
```
madmaxoft: This line is the actual fix, the rest is just cosmetics.

tonibm19:
Damn.
I remember complaining about growing memory use maybe 2 or 3 years ago xD.
Finally fixed.
But does it affect both windows&linux? Because I think the one I had affected linux only.

madmaxoft:
It affects both. Most likely this leak is not that old, it must have sprouted when the login protocol was changed, I'd guess no more than a year ago.

PR #3795, opened 2017-06-21
cBlockArea supports block entities.

### LINKED ISSUE #2461: Support block entities in cBlockArea instances
The `cBlockArea` class needs to read and write (and thus store, too) information about block entities contained within its data.

Implementing this may require slight changes into `cBlockEntity`-derived classes as well in order to support containment in a `cBlockArea`, without a `cWorld` instance assigned to them.

There should be a special `baBlockEntities` flag indicating whether block entities should be read from / written to world, similar to how `baTypes` and such are handled.

madmaxoft:
Fixes #2461.

Adds support for reading and writing block entities in `cBlockArea`. Additionally, it moves the parameter checks from `cBlockArea`'s C++ implementation to the Lua API bindings; the C++ code is now expected to always pass in valid values (`ASSERT`).

Tested using WorldEdit's `//copy` and `//paste` commands (need update, too).


NiLSPACE:
You weren't lying about changing an awful lot of code. :P

src/Bindings/LuaState.cpp
```diff
@@ -1762,6 +1762,30 @@ bool cLuaState::CheckParamEnd(int a_Param)
...
+		return true;
+	}
+
+	// Not the correct parameter
+	lua_Debug entry;
+	VERIFY(lua_getstack(m_LuaState, 0,   &entry));
+	VERIFY(lua_getinfo (m_LuaState, "n", &entry));
+	AString ErrMsg = Printf(
+		"Error in function '%s'. The 'self' parameter is not of the expected type, \"instance of %s\". Make sure you're using the correct calling convention (obj:fn() instead of obj.fn()).",
+		(entry.name != nullptr) ? entry.name : "?", a_SelfClassName
```
NiLSPACE: Perhaps something like `<unknown>` might be better than a questionmark?

NiLSPACE:
I assume this doesn't yet copy the contents of chests, but only creates the actual blockentity.

src/Bindings/LuaState.cpp
madmaxoft: You're right, I'll change it.

madmaxoft:
It fully copies all block entities (thanks to #3772).

you underestimate my power

NiLSPACE:
Right, for some reason I didn't think about that PR.

src/Chunk.cpp
```diff
@@ -420,16 +420,16 @@ void cChunk::WriteBlockArea(cBlockArea & a_Area, int a_MinBlockX, int a_MinBlock
 	int BlockEndX   = std::min(a_MinBlockX + a_Area.GetSizeX(), (m_PosX + 1) * cChunkDef::Width);
 	int BlockStartZ = std::max(a_MinBlockZ, m_PosZ * cChunkDef::Width);
 	int BlockEndZ   = std::min(a_MinBlockZ + a_Area.GetSizeZ(), (m_PosZ + 1) * cChunkDef::Width);
-	int SizeX = BlockEndX - BlockStartX;
+	int SizeX = BlockEndX - BlockStartX;  // Size of the union
 	int SizeZ = BlockEndZ - BlockStartZ;
```
NiLSPACE: Isn't there a `cBlockArea::GetSize()`?

src/Chunk.cpp
```diff
@@ -420,16 +420,16 @@ void cChunk::WriteBlockArea(cBlockArea & a_Area, int a_MinBlockX, int a_MinBlock
...
-	int OffX = BlockStartX - m_PosX * cChunkDef::Width;
+	int SizeY = std::min(a_Area.GetSizeY(), cChunkDef::Height - a_MinBlockY);
+	int OffX = BlockStartX - m_PosX * cChunkDef::Width;  // Offset within the chunk where the union starts
 	int OffZ = BlockStartZ - m_PosZ * cChunkDef::Width;
```
NiLSPACE: Wasn't there a method in `cChunkDef` for this?

Seadragon91:
The APIDoc for cBlockArea is in rewrite?

madmaxoft:
What do you mean? I just moved it into a separate file because it is rather huge already.
As for the various parameter combinations, I decided not to document them just yet.

src/Chunk.cpp
madmaxoft: There is, but it doesn't do what this does - this is the size of the geometric union of the `cBlockArea` and the current chunk. Could be the whole area, could be just a part.

src/Chunk.cpp
madmaxoft: It is rather unwieldy for this case (has output params rather than multi-returns).

Seadragon91:
Then I think you forgot the commit the new file. I don't see it :)

madmaxoft:
You're right, I forgot about it when moving stuff between the computers. Added now, sorry.

madmaxoft:
Yay, finally accepted by CI.

madmaxoft:
Note that this PR doesn't yet support loading and saving block entities in `.schematic` data. First #3802 will need to be implemented.

Issue #2469, opened 2015-09-08
Cannot place wood pressure plate on fence


jammet:
I've seen people make tables with that kind of combination, and for some reason, I cannot do it on Cuberite right now.

Bond-009:
This was fixed in #3796

Seadragon91:
Thanks.

PR #3796, opened 2017-06-21
Pressure plates can be placed on fences and hoppers

Bond-009:
This fixes issue #2469

### DIFF
--- a/src/Blocks/BlockPressurePlate.h
+++ b/src/Blocks/BlockPressurePlate.h
@@ -23,12 +23,31 @@ public:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, int a_RelX, int a_RelY, int a_RelZ, const cChunk & a_Chunk) override
 	{
-		if (a_RelY <= 0)
+		if (a_RelY - 1 <= 0)
 		{
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


src/Blocks/BlockPressurePlate.h
```diff
@@ -28,7 +28,26 @@ class cBlockPressurePlateHandler :
 			return false;
 		}
 
-		return (cBlockInfo::FullyOccupiesVoxel(a_Chunk.GetBlock(a_RelX, a_RelY - 1, a_RelZ)));
+		// TODO: check if the block is upside-down slab or upside-down stairs
+		BLOCKTYPE Block = a_Chunk.GetBlock(a_RelX, a_RelY - 1, a_RelZ);
```
madmaxoft: Thies needs a check (or a change to the check above) - if `a_RelY` is zero, this will crash.

Bond-009: Do you mean `a_RelY - 1`?

madmaxoft: `a_RelY - 1` must be at least zero for this to work (Y coord ranges from 0 to 255). So when attempting to place a pressure plate at the bottom-most layer of the world, your code will read a block at Y coord `-1`; that is definitely not good and can crash the server.

Bond-009: It should be fixed now.

Issue #2729, opened 2015-12-11
Animals don't follow you using leads


rekuen:
I' have noticed that animals don't cross an open fence gate when using food to make them follow you, they just turn around when are in front of a fence gate.

Also they don't follow me when I use a lead on them, I only test this under Survival using the latest build.

SafwatHalaby:
Regarding gates, there's already an issue here: https://github.com/cuberite/cuberite/issues/2062
I'll rename this ticket and only keep the lead issue.

rekuen:
OK, sorry for the duplicate.

sagr4019:
is anybody still working on this?

LogicParrot:
No, I am working on a big AI rewrite though, and if someone works in parallel on this, it would likely cause collisions.

Issue #3186, opened 2016-05-14
Right clicking a tamed wolf will not make it sit


ahyattdev:
### Expected behavior

The wolf will sit
### Actual behavior

The wolf does not sit
### Steps to reproduce the behavior

Tame a wolf and try getting it to sit

mathias-github:
Until this bug is fixed, you can make a wolf sit by holding an item with a right-click feature (e.g. a  bow, shield)

PR #3798, opened 2017-06-21
Leads, leashed mobs and leash knots on fences

### LINKED ISSUE #2915: Leads do not work properly
Part of this is already covered in issue #2729 but leads do not work at all.

When using a lead on horses you occasionally get on the horse.

When using a lead on an animal and then clicking on a fence gate the lead does not attach the animal to the gate.

As #2729 say animals do not follow you when using a lead.

Related to #2729 leads do not break after long distances.

spekdrum:
- Adds leads handling, leashed mobs and leash knots on fences.
- Drops appropriate pickup on unleash.
- Save and load knots from map and re-leash mobs that were leashed before server close.
- Knots sound.
- Support from 1.8 to 1.12 protocol versions.

It's been a lot of work and testing but that was very enjoyable :P

Fixes #2915 and #2729

**Tests done:**
- Create some leads ( /i lead 24 ). They are stack-able now, as in vanilla.
- Go for passive mobs (pigs, cows, horses...) and right click them with a lead on main hand. They will be leashed to you. 
- Try on others than passive mobs (wont work).
- Take more than one mob at a time. You are now Santa Claus.
- Look them following you.
- Try to go far from the mob (> 10 blocks). Leads should break with no drop.
- Unleash a leashed mob right clicking on it, then used lead will become a pickup.
- Kill the mob while it's leashed to you, leash link (the drawn rope between you and the mob) will go out and there will be no drop.
- Catch a mob with a lead and then right click on a fence, a leash knot will show up. The leash sound should be heard. The mob will unleash from your player and get leashed to the knot.
- Try to leash more mobs to the same fence block. No more knots will be created, the existing will be reused. No sound will be reproduced.
- Try to leash more mobs to the knot (it's different to the server right clicking on a knot than right click on the fence sustaining the knot). It will work.
- Try to break a knot, all mobs leashed will be released.
- Try to break a fence where there is a knot. As in vanilla, the knot will break passed about a second, and mobs will be released.
- Try to leash a mob that's currently leashed to a fence leash knot. No effect and no rope line drawing. That last has been tricky to fix, because client anticipates to server and draws the line of the leash between player and the mob without regard from the current state of the mob.
- Try to put several leash knots on adjacent fence blocks, logout, close server, open server again, login. Every mob will be leashed to it's knot again.

**Not implemented:**
- Leashed mobs in vanilla can go around up to 5 blocks distance from the knot. Know mobs just go to the target pos like in other situations.
- Surely I forgot some LUA and doc stuff


spekdrum:
Agree with upgrading to cMonster, this will also make cleaner code.
I will think on a solution for the chunks problem, first I will try to reproduce it.

The delay for the knot is as in vanilla. I  thought they did it this way to point out to the player he ruined a knot breaking the fence or whatever. But maybe vanilla behaves this way for some limitation that cuberite dont suffer :p

Im on holidays now, will put hands on this in a week or so. Happy coding ;)

spekdrum:
I'm working on the possibility to leash any monster and I found a problem. With any aggressive monster (tried creeper and skeleton) the client sends the right click event twice, when you just right click once on it. The result is that the monster gets leashed but immediately unleashed.
Should I have to use a variable to store last leash action tick count in order to avoid this? So ugly...

Seadragon91:
Ah that problem. Had also been reported in this issue with sitting a wolf: https://github.com/cuberite/cuberite/issues/3186

spekdrum:
Looking at the call stack, having put a breakpoint in cClientHandle::HandleUseEntity it seems that is the client that sends the packet twice...

spekdrum:
I found an interesting point with the double right click problem.
This is the code that gets executed twice:
```
void cProtocol_1_9_0::HandlePacketUseEntity(cByteBuffer & a_ByteBuffer)
{
	HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, EntityID);
	HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Type);

	switch (Type)
	{
		case 0:
		{
			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand)
			m_Client->HandleUseEntity(EntityID, false);
			break;
		}
```

The "Hand" variable first time is 0 and second time is 1. According to the protocol doc (http://wiki.vg/Protocol#Use_Entity) that means main hand (0) or off-hand (1). Maybe we should check which hand holds the item we want to use with the entity, but we need to change the code a bit in order to let this parameter arrive to every place. I suggest to initially ignore calls for the off-hand and implement off-hand actions later. It's indeed a very easy fix. This way we can get rid of those double action related issues till a complete solution is done.

spekdrum:
@madmaxoft  As suggested I moved the logic to cMonster with a flag.
I've been doing some tests regarding the chunk borders potencial problem. I reproduced the problem, no crash happened, just mobs became unleashed and knot was still there (with no mob leashed). To fix that, I removed the line that sets m_LeashToPos to nullptr. This way the mob keeps trying till the knot is found. I introduced a stepped Tick there to avoid unnecessary calcs.

The delay for the knot destroy upon fence wipe is still there, because vanilla does this way. I like to remark the player the knot is gone. If you insist I have no problem to remove it, but please consider my argument :)

spekdrum:
I added an overload for cBoundingBox in order to get horizontal and vertical radius

Seadragon91:
Should this work with monsters like creeper now?

spekdrum:
Only if you override m_CanBeLeashed flag. By default only passive monsters have this flag on.
You can see it in cMonster constructor.

spekdrum:
I fixed a possible leash lose when a mob is looking for it's knot and server gets closed. Also stopped mobs when they are leashed to a knot that has not been found yet.

spekdrum:
Added speeding up mobs getting far from player when leashed to a player as Vanilla does.

src/BoundingBox.h
```diff
@@ -27,6 +27,7 @@ class cBoundingBox
 	cBoundingBox(double a_MinX, double a_MaxX, double a_MinY, double a_MaxY, double a_MinZ, double a_MaxZ);
 	cBoundingBox(const Vector3d & a_Min, const Vector3d & a_Max);
 	cBoundingBox(const Vector3d & a_Pos, double a_Radius, double a_Height);
+	cBoundingBox(const Vector3d & a_Pos, double a_Radius, double a_Height, double a_VerticalOffset);
```
madmaxoft: This deserves a doxy-comment.

src/ClientHandle.cpp
```diff
@@ -1704,6 +1704,7 @@ void cClientHandle::HandleUpdateSign(
 void cClientHandle::HandleUseEntity(UInt32 a_TargetEntityID, bool a_IsLeftClick)
 {
 	// TODO: Let plugins interfere via a hook
+	LOGD("Use entity %d", a_TargetEntityID);
```
madmaxoft: This should probably be deleted.

src/Entities/LeashKnot.cpp
```diff
@@ -0,0 +1,154 @@
...
+		if (m_TicksToSelfDestroy <= 0)
+		{
+			Destroy();
+			m_World->BroadcastSoundEffect("entity.leashknot.break", GetPosX(), GetPosY(), GetPosZ(), 1, 1);
+		}
+	}
+	else
+	{
+		// Detect if fence is destroyed. Get block at knot position and check if its solid
+		BLOCKTYPE BlockID = m_World->GetBlock(static_cast<int>(GetPosX()), static_cast<int>(GetPosY()), static_cast<int>(GetPosZ()));
```
madmaxoft: Use `FloorC` instead of the casts.

src/Mobs/Horse.cpp
```diff
@@ -131,24 +131,28 @@ void cHorse::OnRightClicked(cPlayer & a_Player)
 	}
 	else if (a_Player.GetEquippedItem().IsEmpty())
 	{
-		if (m_Attachee != nullptr)
+		// Check if leashed / unleashed to player before try to ride
+		if (!m_IsLeadActionJustDone)
```
madmaxoft: `Lead` or `Leash`?

src/Mobs/Monster.cpp
```diff
@@ -582,6 +685,27 @@ void cMonster::OnRightClicked(cPlayer & a_Player)
 			a_Player.GetInventory().RemoveOneEquippedItem();
 		}
 	}
+
+	// Using leads
+	m_IsLeadActionJustDone = false;
```
madmaxoft: Leash

src/Mobs/Monster.h
```diff
@@ -74,6 +76,30 @@ class cMonster :
...
 	virtual void EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk);
 
+	/** Returns whether the mob can be leashed. */
+	bool CanBeLeashed() const { return m_CanBeLeashed; }
+
+	/** Sets whether the mob ban be leashed, for extensibility in plugins */
+	void SetCanBeLeashed(bool a_CanBeLeashed) { m_CanBeLeashed = a_CanBeLeashed; }
+
+	/** Returns whether the monster is leashed to an entity. */
+	virtual bool IsLeashed() const { return (m_LeashedTo != nullptr); }
```
madmaxoft: Why `virtual`?

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1234,6 +1234,30 @@ void cProtocol_1_12::SendScoreUpdate(const AString & a_Objective, const AString
...
+}
+
+
+
+
+
+void cProtocol_1_12::SendUnleashEntity(const cEntity & a_Entity)
+{
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, 0x3C);  // Set Attach Entity packet
```
madmaxoft: Prefer lowercase hex digits

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1234,6 +1234,30 @@ void cProtocol_1_12::SendScoreUpdate(const AString & a_Objective, const AString
 
 
 
+void cProtocol_1_12::SendLeashEntity(const cEntity & a_Entity, const cEntity & a_EntityLeashedTo)
+{
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, 0x3C);  // Set Attach Entity packet
```
madmaxoft: Prefer lowercase hex digits

src/Protocol/Protocol_1_8.h
```diff
@@ -90,6 +90,7 @@ class cProtocol_1_8_0 :
 	virtual void SendHideTitle                  (void) override;
 	virtual void SendInventorySlot              (char a_WindowID, short a_SlotNum, const cItem & a_Item) override;
 	virtual void SendKeepAlive                  (UInt32 a_PingID) override;
+	virtual void SendLeashEntity               (const cEntity & a_Entity, const cEntity & a_EntityLeashedTo) override;
```
madmaxoft: Wrong alignment, one more space

src/Protocol/Protocol_1_9.cpp
```diff
@@ -643,6 +643,30 @@ void cProtocol_1_9_0::SendKeepAlive(UInt32 a_PingID)
 
 
 
+void cProtocol_1_9_0::SendLeashEntity(const cEntity & a_Entity, const cEntity & a_EntityLeashedTo)
+{
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, 0x3A);  // Set Attach Entity packet
```
madmaxoft: lowercase hex

src/Protocol/Protocol_1_9.cpp
```diff
@@ -643,6 +643,30 @@ void cProtocol_1_9_0::SendKeepAlive(UInt32 a_PingID)
...
+}
+
+
+
+
+
+void cProtocol_1_9_0::SendUnleashEntity(const cEntity & a_Entity)
+{
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, 0x3A);  // Set Attach Entity packet
```
madmaxoft: lowercase hex

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -555,6 +556,33 @@ void cNBTChunkSerializer::AddMonsterEntity(cMonster * a_Monster)
...
+		{
+			m_Writer.AddByte("Leashed", 1);
+
+			const Vector3d * LeashedToPos = nullptr;
+
+			if (a_Monster->GetLeashToPos() != nullptr)
+			{
+				LeashedToPos = a_Monster->GetLeashToPos();
+			}
+			else if (a_Monster->GetLeashedTo()->IsLeashKnot())
```
madmaxoft: Could both LeashedToPos and LeashedTo be nullptr?

madmaxoft:
This is definitely on the right way, just a few cosmetics now :)

src/WorldStorage/NBTChunkSerializer.cpp
spekdrum: No, because the outer _if_ ensures that one of them is not null. 
a_Monster->IsLeashed() it's the same as a_Monster->GetLeashedTo() != nullptr

spekdrum:
Should I mark dirty a chunk when a player places a leash knot on a fence?

Seadragon91:
A new entity is created, not? If yes then it should be marked dirty

spekdrum:
Ok, do not merge please, I'm going to mark them all dirty :P

spekdrum:
I checked the code and there is no further change needed. MarkDirty() is already called later, once the entity is initialized and added to vector m_EntitiesToAdd from World. Some happens when destroyedd. Everything happens in cWorld::Tick.

**So, for me this is done!**

src/Blocks/BlockFence.h
```diff
@@ -0,0 +1,142 @@
...
+public:
+	cBlockFenceHandler(BLOCKTYPE a_BlockType) :
+		cMetaRotator<cBlockHandler, 0x03, 0x02, 0x03, 0x00, 0x01, true>(a_BlockType)
+	{
+	}
+
+	virtual bool OnUse(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cPlayer * a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ) override
+	{
+		// this vector allows to store the reference to an existing knot, if found inside the callback function
+		std::vector<cLeashKnot *> LeashKnotList;
```
madmaxoft: Why a `vector`? A simple pointer should suffice for this.

src/Entities/LeashKnot.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+
+
+
+
+void cLeashKnot::OnRightClicked(cPlayer & a_Player)
+{
+	super::OnRightClicked(a_Player);
+
+	// Check leashed nearby mobs to tight them to this knot
```
madmaxoft: `tight` -> `tie`

src/Entities/LeashKnot.cpp
```diff
@@ -0,0 +1,154 @@
...
+				return false;
+			}
+
+			// All conditions met, unleash from player and leash to fence
+			PotentialLeashed->GetLeashedTo()->RemoveLeashedMob(PotentialLeashed, false, false);
+			m_Knot->AddLeashedMob(PotentialLeashed);
+			return false;
+		}
+	} Callback(this, &a_Player);
+	a_Player.GetWorld()->ForEachEntityInBox(cBoundingBox(GetPosition(), 8, 8, -4), Callback);
```
madmaxoft: This big callback code seems repeated in `cBlockFenceHandler::OnUse`. Could it be moved into a method of its own (`TieHandLeashedMobs` (?)) and shared?

src/Entities/LeashKnot.cpp
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+	else
+	{
+		// Detect if fence is destroyed. Get block at knot position and check if its solid
+		BLOCKTYPE BlockID = m_World->GetBlock(FloorC(GetPosX()), FloorC(GetPosY()), FloorC(GetPosZ()));
+		if (!cBlockInfo::IsSolid(BlockID))
+		{
+			m_ShouldSelfDestroy = true;
+		}
```
madmaxoft: Does this need to happen here? Couldn't the fence being broken signalize that the knot should be destroyed?

src/Mobs/Monster.cpp
```diff
@@ -367,6 +420,56 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			}
+		} Callback(this);
+		m_World->ForEachEntityInBox(cBoundingBox(m_LeashToPos, 0.5, 1), Callback);
+	}
+	else if (IsLeashed())  // Mob is leashed to an entity
+	{
+		// TODO: leashed mobs in vanilla can move around up to 5 blocks distance from leash origin
+		MoveToPosition(m_LeashedTo->GetPosition());
+
+		// If distance to target > 10 break lead
```
madmaxoft: `lead` -> `leash`

src/Mobs/Monster.cpp
```diff
@@ -582,6 +685,27 @@ void cMonster::OnRightClicked(cPlayer & a_Player)
 			a_Player.GetInventory().RemoveOneEquippedItem();
 		}
 	}
+
+	// Using leads
```
madmaxoft: leash

src/Entities/Entity.cpp
```diff
@@ -2247,3 +2264,38 @@ void cEntity::SetPosition(const Vector3d & a_Position)
...
+	if (broadcast)
+	{
+		m_World->BroadcastLeashEntity(*a_Monster, *this);
+	}
+}
+
+
+
+
+void cEntity::RemoveLeashedMob(cMonster * a_Monster, bool a_DropPickup, bool broadcast)
```
madmaxoft: Style: `a_ShouldBroadcast`

src/Entities/Entity.cpp
```diff
@@ -2247,3 +2264,38 @@ void cEntity::SetPosition(const Vector3d & a_Position)
 
 
 
+
+void cEntity::AddLeashedMob(cMonster * a_Monster, bool broadcast)
```
madmaxoft: Style: `a_ShouldBroadcast`

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -3122,6 +3143,58 @@ bool cWSSAnvil::LoadMonsterBaseFromNBT(cMonster & a_Monster, const cParsedNBT &
 		a_Monster.SetCustomNameAlwaysVisible(CustomNameVisible);
 	}
 
+	// Leashed to a knot
+	bool Leashed;
+	int LeashedIdx = a_NBT.FindChildByName(a_TagIdx, "Leashed");
+	if (LeashedIdx >= 0)
```
madmaxoft: Use an early bailout (negate the condition, `return`) to avoid unnecessary indent levels.

madmaxoft: Perhaps even move this into a separate function, since `LoadMonsterBase` is already pretty long by itself.

madmaxoft:
There are still quite a few instances of lead / leash confusion, and a few calls to `LOGD` that should probably be removed.

spekdrum:
Ready! :+1:

src/Blocks/BlockFence.h
```diff
@@ -0,0 +1,113 @@
...
+{
+public:
+	cBlockFenceHandler(BLOCKTYPE a_BlockType) :
+		cMetaRotator<cBlockHandler, 0x03, 0x02, 0x03, 0x00, 0x01, true>(a_BlockType)
+	{
+	}
+
+	virtual bool OnUse(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cPlayer * a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ) override
+	{
+		cLeashKnot * LeashKnot = nullptr;
```
madmaxoft: This still feels clunky. How about:
```cpp
class cLeashKnot
{
  ...
  /** Returns the leash knot entity representing the knot at the specified position.
  Returns nullptr if there's no knot. */
  static cLeashKnot * FindKnotAtPos(cWorld & a_World, Vector3i a_BlockPos);
};
```

Then here:
```cpp
auto leashKnot = cLeashKnot::FindKnotAtPos(a_Player->GetWorld(), {a_BlockX, a_BlockY, a_BlockZ});
auto knotAlreadyExists = (leashKnot != nullptr);
```

src/Mobs/Monster.cpp
```diff
@@ -367,6 +420,53 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			virtual bool Item(cEntity * a_Entity) override
+			{
+				if (a_Entity->IsLeashKnot())
+				{
+					a_Entity->AddLeashedMob(m_Monster);
+					m_Monster->SetLeashToPos(nullptr);
+				}
+				return false;
+			}
+		} Callback(this);
```
madmaxoft: This could reuse the new `cLeashKnot::FindKnotAt` function

src/Mobs/Monster.h
```diff
@@ -74,6 +76,30 @@ class cMonster :
...
+
+	/** Returns the entity to where is leashed this mob, returns nullptr if it's not leashed */
+	cEntity * GetLeashedTo() const { return m_LeashedTo; }
+
+	/** Sets entity position to where is leashed this mob */
+	void SetLeashToPos(Vector3d * pos) { m_LeashToPos = pos; }
+
+	/** Gets entity position to where mob should be leashed */
+	Vector3d * GetLeashToPos() const { return m_LeashToPos; }
+
```
madmaxoft: It would make sense exporting some of these to Lua API. Use the `// tolua_begin` and `// tolua_end` markers to auto-export them.

src/Mobs/Monster.h
```diff
@@ -274,8 +300,23 @@ class cMonster :
 	/** Adds weapon that is equipped with the chance saved in m_DropChance[...] (this will be greter than 1 if picked up or 0.085 + (0.01 per LootingLevel) if born with) to the drop */
 	void AddRandomWeaponDropItem(cItems & a_Drops, unsigned int a_LootingLevel);
 
+	/** Entity leashed to */
+	cEntity * m_LeashedTo;
+
+	/** Entity pos leashed to, for leash knots */
+	Vector3d * m_LeashToPos;
```
madmaxoft: This deserves a better comment - why is it a pointer and not a regular member? And anyway, why not a `std::unique_ptr`?

src/Mobs/Monster.h
spekdrum: Ok. All of them maybe?

src/Mobs/Monster.h
spekdrum: Oh, sorry, that's smelling of memory leak. 
I will change it to a regular member. Is there any benefit of using unique_ptr instead of a regular member here? Thanks

madmaxoft: I thought this member was set to `nullptr` for some special cases. if not, then by all means change to a regular object.

src/Mobs/Monster.h
madmaxoft: Depends. Do all of them do something that could be used by plugins and is safe from mis-use by plugins? I'd imagine the `SetLeashedTo` doesn't do exactly what one would expect - leash the monster to an entity completely (including the protocol broadcast etc.; rather that seems like an internal method.

src/Entities/LeashKnot.h
```diff
@@ -0,0 +1,45 @@
...
+public:
+
+	// tolua_end
+
+	CLASS_PROTODEF(cLeashKnot)
+
+	cLeashKnot(eBlockFace a_BlockFace, double a_X, double a_Y, double a_Z);
+
+	/** Looks for mobs leashed to a player and ties them to this knot */
+	void TieNearbyMobs(cPlayer & a_Player, bool a_ShouldBroadCast);
```
madmaxoft: The name is somewhat misleading, it doesn't tie all nearby mobs. `TiePlayersMobs` might be better.

src/Mobs/Monster.h
spekdrum: You are right, nullptr assignment is needed in certain cases.

spekdrum:
I'll fix It tomorrow:

cMonster.CanBeLeashed (function)
cMonster.GetLeashedTo (function)
cMonster.IsLeashed (function)
cMonster.SetCanBeLeashed (function)
cMonster.SetLeashedTo (function)
cMonster.SetUnleashed (function)

spekdrum:
All fixed!

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -4274,7 +4284,11 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 				},
 				etItemFrame =
 				{
-					Notes = "",
+					Notes = "The entity is a item frame",
```
madmaxoft: `an`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8533,6 +8557,17 @@ a_Player:OpenWindow(Window);
...
+				GetLeashedTo =
+				{
+					Returns =
+					{
+						{
+							Name = "LeashedTo",
+							Type = "cEntity",
+						},
+					},
+					Notes = "Returns the entity to where this mob is leashed, returns nullptr if it's not leashed",
```
madmaxoft: In Lua, `nil` is used instead of `nullptr`.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8693,6 +8749,28 @@ a_Player:OpenWindow(Window);
...
+				SetLeashedTo =
+				{
+					Params =
+					{
+						{
+							Name = "Entity",
+							Type = "cEntity",
+						},
+					},
+					Notes = "Leash the monster to a entity.",
```
madmaxoft: Is there any restriction on the entity, or can it be any entity at all? Would be worth mentioning.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8693,6 +8749,28 @@ a_Player:OpenWindow(Window);
...
+				},
+				SetUnleashed =
```
madmaxoft: The function name feels "Unenglish", a simple `Unleash` makes more sense.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8693,6 +8749,28 @@ a_Player:OpenWindow(Window);
...
+				{
+					Params =
+					{
+						{
+							Name = "ShouldDropPickup",
```
madmaxoft: `ShouldDropLeashPickup`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8693,6 +8749,28 @@ a_Player:OpenWindow(Window);
...
+							Type = "boolean",
+						},
+					},
+					Notes = "Unleash the monster from an entity.",
```
madmaxoft: `Unleash the monster` would be enough.

src/BoundingBox.h
```diff
@@ -27,6 +27,12 @@ class cBoundingBox
 	cBoundingBox(double a_MinX, double a_MaxX, double a_MinY, double a_MaxY, double a_MinZ, double a_MaxZ);
 	cBoundingBox(const Vector3d & a_Min, const Vector3d & a_Max);
 	cBoundingBox(const Vector3d & a_Pos, double a_Radius, double a_Height);
+
+	/** Constructor that allows to define a bounding box given a center point (a_Pos), an horizontal radius (a_Radius),
```
madmaxoft: `an` -> `a`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8693,6 +8749,28 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Sets the custom name visiblity of this monster. If it's false, you only see the name when you sight the mob. If it's true, you always see the custom name.",
 				},
+				SetLeashedTo =
```
madmaxoft: Why not call the function simply `LeashTo`?

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: The function implementation doesn't seem to implement the protocol-sending etc.; if a plugin calls this, will it be enough to completely unleash the monster?

madmaxoft:
Sorry, not just yet :)

Server/Plugins/APIDump/APIDesc.lua
spekdrum: When referring to the item, should we call it lead? In game it's name is 'lead'.

madmaxoft: Personally I'd call it `leash`, because `lead` can mean other things, too (are there mods that add other ores, such as lead, into the game?)
After all, we have `planks` and `log` and no `wood`, because it caused confusion, and everyone was happy with it :)

Server/Plugins/APIDump/APIDesc.lua
spekdrum: Sorry but I don't know Lua capabilities over Cuberite very well. So I don't have the knowledge to expose this functions in the right way. Maybe only cEntity.AddLeashedMob / cEntity.RemoveLeashedMob should be called from Lua plugins because there is all the work needed, including the protocol thing. In fact Monster.LeashTo and Monster.Unleash are called from there, so they are kind of internal functions.
So, can I add lua_begin / lua_end in cEntity in order to export those functions (cEntity.AddLeashedMob / cEntity.RemoveLeashedMob)? They are currently not being exported.

Server/Plugins/APIDump/APIDesc.lua
spekdrum: It's ok :)

Server/Plugins/APIDump/APIDesc.lua
spekdrum: cEntity.HasAnyMobLeashed would be another one to include in Lua

madmaxoft: Then perhaps add a new function `cMonster::LeashTo(cEntity * a_Entity)` that just calls the `AddLeashedMob`, and export that to Lua instead. Similarly with unleashing.

madmaxoft: Or maybe the other way around - make `AddLeashedMob` only process the internal list of leashed entities, and make `cMonster::LeashTo` do all the hard work, including the protocol sending.

spekdrum: At first I did what you are saying on your last proposal (put the hard work inside mob class) but that gave me some memory lock problems when closing the server.
So maybe creating those 2 new functions in cMonster will work. Sorry, but why not just export cEntity.AddLeashedMob / cEntity.RemoveLeashedMob / cEntity.HasAnyMobLeashed and hide cMonster's LeashTo / Unleash? They are fully functional for the purpose.

madmaxoft: The function names are misleading. The name says "I'm adding something to some internal list", while in fact it does all the heavy work of the network comm. That's probably my biggest gripe about them.

spekdrum: I see. No problem, I can rename cEntity.AddLeashedMob / cEntity.RemoveLeashedMob to "Leash" and "Unleash" respectively. What do you think?

madmaxoft: Those functions should be "on the monster" which is being leashed / unleashed, not "on the entity" to which the monster is being leashed.

I think the Lua API would be best if it could be used like this:
```lua
-- monsterX are cMonster instances from somewhere
monster1:LeashTo(player1)
monster2:LeashToPos(Vector3i(blockX, blockY, blockZ))  -- creates a new leash knot
monster3:Unleash()
```

spekdrum:
I've finally been able to move the leash / unleash main logic from cEntity to cMonster. So now LeashTo / Unleash functions can be exposed to Lua.

spekdrum:
This is done for me. Is there any fix pending?

src/Mobs/Monster.h
```diff
@@ -74,6 +76,32 @@ class cMonster :
...
+	void SetCanBeLeashed(bool a_CanBeLeashed) { m_CanBeLeashed = a_CanBeLeashed; }
+
+	/** Returns whether the monster is leashed to an entity. */
+	bool IsLeashed() const { return (m_LeashedTo != nullptr); }
+
+	/** Leash the monster to an entity. */
+	void LeashTo(cEntity * a_Entity, bool a_ShouldBroadcast = true);
+
+	/** Unleash the monster. */
+	void Unleash(bool a_ShouldDropLeashPickup, bool a_ShouldBroadcast = true, bool a_ShouldRemoveFromEntity = true);
```
madmaxoft: The `a_ShouldRemoveFromEntity` parameter should be documented more in depth - what does it do when set to true and false?

src/Mobs/Monster.h
```diff
@@ -74,6 +76,32 @@ class cMonster :
 	virtual void CheckEventSeePlayer(cChunk & a_Chunk);
 	virtual void EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk);
 
+	// tolua_begin
```
madmaxoft: Add an empty line after this, to make it stand out more. Same with the finalizing `// tolua_end`

src/Mobs/Monster.h
```diff
@@ -274,8 +302,23 @@ class cMonster :
...
+
+	/** Entity pos where this mob was leashed to. Used when deserializing the chunk in order to make the mob find the leash knot. */
+	std::unique_ptr<Vector3d> m_LeashToPos;
+
+	/** Mob has ben leashed or unleashed in current player action. Avoids double actions on horses. */
+	bool m_IsLeashActionJustDone;
+
+	/** Determines whether a monster can be leashed */
+	bool m_CanBeLeashed;
+
```
madmaxoft: These should be moved above the function declarations. We sort each class section: types first, then variables, and finally functions.

src/Mobs/Monster.h
madmaxoft: The function has three params, so the Lua binding accepts three params, too. Does it make sense for all three of them to be accessible to plugins? And their documentation is missing.

If you don't want to write manual bindings for this, just add an overload that has less params and calls this internally, and make it the only exported API function.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1951,6 +1955,23 @@ void cWSSAnvil::LoadItemFrameFromNBT(cEntityList & a_Entities, const cParsedNBT
 
 
 
+void cWSSAnvil::LoadLeashKnotFromNBT(cEntityList & a_Entities, const cParsedNBT & a_NBT, int a_TagIdx)
+{
+	std::unique_ptr<cLeashKnot> LeashKnot = cpp14::make_unique<cLeashKnot>(BLOCK_FACE_NONE, 0.0, 0.0, 0.0);
```

madmaxoft:
Still a few more changes, but we're definitely getting there. It's mostly cosmetics now, except for the API function.

spekdrum:
Ok, doing those changes. Thanks.

**Please, I need a bit of help here:**
I merged the recent changes made in BlockFence.h. But now this function is not called when a player breaks a fence:
`virtual void OnDestroyedByPlayer(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cPlayer * a_Player, int a_BlockX, int a_BlockY, int a_BlockZ)`

This is function called instead:
`virtual void OnDestroyed(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, int a_BlockX, int a_BlockY, int a_BlockZ)`

I don't know if this is the right behaviour, but I have to handle this situation anyway, because a fence can be broken by something (no player) and a potencial knot on it must be set to destroy. 
I need to get a reference to cWorld. When OnDestroyedByPlayer was called I could get &cWorld from a_Player, but I don't have the player reference inside OnDestroyed function. cChunkInterface  has a ChunkMap, that has m_World, but has no getter. Can anyone help me with this? Thanks!

Seadragon91:
What's about adding a function that returns the world in `cChunkInterface`?
`cWorld * GetWorld(void) { return m_ChunkMap->GetWorld(); }`

spekdrum:
@Seadragon91 Perfect. Will do that.

src/Mobs/Monster.h
spekdrum: I better do the overload. My intention is that Unleash has only one parameters for plugins. I thought that the optionals ( = true ) were hiding the other two paremeters for plugins.

madmaxoft:
**NO, DON'T!**

The `cChunkInterface` and `cWorldInterface` are meant to isolate things, so that not everything depends on a master god object (`cWorld`). Instead, if you need a function from `cWorld`, look for it in one of those, and if it's not there, add it as an interface (`virtual fn(...) = 0` in `cXYZInterface`, then `virtual fn(...) override` in `cWorld`)

spekdrum:
> NO, DON'T!
xDDD 
ok

spekdrum:
Done

src/Entities/Entity.cpp
```diff
@@ -218,6 +227,13 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
 	ASSERT(GetParentChunk() != nullptr);
 	SetIsTicking(false);
 
+	// Unleash leashed mobs
+	for (auto LeashedMob : m_LeashedMobs)
+	{
+		LeashedMob->Unleash(true, true, false);
+	}
+	m_LeashedMobs.clear();
```
madmaxoft: Can you try if this works? Removing the need for `Unleash`'s  last parameter.
```cpp
while (!m_LeashedMobs.empty())
{
  m_LeashedMobs.front()->Unleash(true, true);
}
```

src/Entities/LeashKnot.cpp
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+
+
+
+
+cLeashKnot * cLeashKnot::FindKnotAtPos(cWorldInterface & a_WorldInterface, Vector3i a_BlockPos)
+{
+	cLeashKnot * LeashKnot = nullptr;
```
madmaxoft: Remove this variable, make `LookForKnot::m_LeashKnot` a simple pointer and do `return CallbackFindKnot.m_LeashKnot` in the end.

src/Mobs/Monster.cpp
```diff
@@ -367,6 +421,39 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
+void cMonster::CalcLeashActions()
+{
+	// This mob just spotted in the world and [m_LeashToPos not null] shows that should be leashed to a leash knot at m_LeashToPos.
+	// This keeps trying until knot is found
```
madmaxoft: Add a note that the leash knot may be in a different chunk that needn't be loaded, to make it clear why this can happen.

src/Mobs/Monster.cpp
```diff
@@ -582,6 +669,26 @@ void cMonster::OnRightClicked(cPlayer & a_Player)
...
+	if (IsLeashed() && (GetLeashedTo() == &a_Player))  // a player can only unleash a mob leashed to him
+	{
+		Unleash(!a_Player.IsGameModeCreative());
+	}
+	else if (IsLeashed())
+	{
+		// Mob is already leashed but client anticipates the server action and draws a leash link, so we need to send current leash to cancel it
+		m_World->BroadcastLeashEntity(*this, *this->GetLeashedTo());
+	}
+	else if (CanBeLeashed() && (EquippedItem.m_ItemType == E_ITEM_LEAD))
```
madmaxoft: How about adding `E_ITEM_LEASH` equivalent to the lead, but better named, and using that everywhere in code instead?

src/Mobs/Monster.cpp
```diff
@@ -1294,3 +1401,60 @@ cMonster::eFamily cMonster::GetMobFamily(void) const
...
 	return FamilyFromType(m_MobType);
 }
+
+
+
+
+
+void cMonster::LeashTo(cEntity * a_Entity, bool a_ShouldBroadcast)
+{
+	m_LeashedTo = a_Entity;
```
madmaxoft: You should probably add `ASSERT(m_LeashedTo == nullptr)`, so that when broken code in the future tries to call this while already leashed, the error is detected early and precisely.

src/Entities/Entity.cpp
spekdrum: Like, much better

spekdrum:
That's kinda poetic man, you are leashing me to this xD

src/Mobs/Monster.cpp
madmaxoft: Oh wait, is this the directly exported Lua API? Plugins are bound to call this while already leashed, because we plugin authors are stupid and too lazy to read docs. So make the function behave well in such a case - ideally re-leash the mob to the new entity, or if not practically possible, don't change the leash-ness.

madmaxoft: Similar with `Unleash` being called while not leashed.

spekdrum: You mean check if already leashed and if so then do nothing, (and viceversa) right?

src/BlockID.h
```diff
@@ -459,6 +459,7 @@ enum ENUM_ITEM_ID : short
 	E_ITEM_GOLD_HORSE_ARMOR = 418,
 	E_ITEM_DIAMOND_HORSE_ARMOR = 419,
 	E_ITEM_LEAD = 420,
+	E_ITEM_LEASH = E_ITEM_LEAD,
```
Bond-009: image

spekdrum:
Done. 
For LeashTo / Unleash I've just add code to check if already / not leashed and do nothing if so.

src/BlockID.h
madmaxoft: 'cause I said so. `Leash` is a much better name than `lead`, because it's unambiguous. Lead can be also a metal, or a verb.

Bond-009: Then why not only use `Leash`?

Pokechu22: Because the vanilla item is named `minecraft:lead`?

spekdrum: Cause he/she said so :smile: 
Not very important think to argue about, others did it before :stuck_out_tongue: 
https://english.stackexchange.com/questions/273617/difference-between-lead-and-leash

madmaxoft:
Yayyy, approved :)

spekdrum:
crying-cat

madmaxoft:
Leash cat

spekdrum:
LOL a leashed cat!

madmaxoft:
I'm afraid the recent merge of int-to-vector in cBoundingBox has produced some merge conflicts. Can you fix those please?

spekdrum:
Sure

spekdrum:
I hope I did it well. I also had to change a bit of code affected by the entity ownership changes.

madmaxoft:
🎉

PR #3801, opened 2017-06-22
Load entities from old and new names

### LINKED ISSUE #3799: Add support for new entity names
Forum-post: https://forum.cuberite.org/thread-2980-post-28333.html#pid28333

Cuberite only reads entities that don't have the prefix `minecraft:`.
I want to add the support for them. Idea of method format:
`bool CheckEntityType(const char * a_IDTag, size_t a_IDTagLength, const AString & a_Type1, const AString & a_Type2 = "");`

`a_Type1` and `a_Type2` are the entity types to check, e.g. `boat` or `minecraft:boat`. The second one is optional. Is this okay so?

Seadragon91:
Fixes #3799
This adds support for reading the new entity names.

Source used for the entity names that are used for saving:
http://minecraft.gamepedia.com/Data_values/Entity_IDs

### DIFF
--- a/src/WorldStorage/WSSAnvil.cpp
+++ b/src/WorldStorage/WSSAnvil.cpp
@@ -1474,228 +1474,143 @@ cBlockEntity * cWSSAnvil::LoadSignFromNBT(const cParsedNBT & a_NBT, int a_TagIdx
 
 void cWSSAnvil::LoadEntityFromNBT(cEntityList & a_Entities, const cParsedNBT & a_NBT, int a_EntityTagIdx, const char * a_IDTag, size_t a_IDTagLength)
 {
-	if (strncmp(a_IDTag, "Boat", a_IDTagLength) == 0)
+	typedef void (cWSSAnvil::*EntityLoaderFunc)(cEntityList &, const cParsedNBT &, int a_EntityTagIdx);
+	typedef std::map<AString, EntityLoaderFunc> EntityLoaderMap;
+	static const EntityLoaderMap EntityTypeToFunction
 	{
-		LoadBoatFromNBT(a_Entities, a_NBT, a_EntityTagIdx);
-	}
-	else if (strncmp(a_IDTag, "EnderCrystal", a_IDTagLength) == 0)
+		{ "Boat",                          &cWSSAnvil::LoadBoatFromNBT },
+		{ "minecraft:boat",                &cWSSAnvil::LoadBoatFromNBT },
+		{ "EnderCrystal",                  &cWSSAnvil::LoadEnderCrystalFromNBT },
+		{ "minecraft:ender_crystal",       &cWSSAnvil::LoadEnderCrystalFromNBT },
+		{ "FallingBlock",                  &cWSSAnvil::LoadFallingBlockFromNBT },
+		{ "minecraft:falling_block",       &cWSSAnvil::LoadFallingBlockFromNBT },
+		{ "Minecart",                      &cWSSAnvil::LoadOldMinecartFromNBT },
+		{ "MinecartChest",                 &cWSSAnvil::LoadMinecartCFromNBT },
+		{ "minecraft:chest_minecart",      &cWSSAnvil::LoadMinecartCFromNBT },
+		{ "MinecartFurnace",               &cWSSAnvil::LoadMinecartFFromNBT },
+		{ "minecraft:furnace_minecart",    &cWSSAnvil::LoadMinecartFFromNBT },
+		{ "MinecartTNT",                   &cWSSAnvil::LoadMinecartTFromNBT },
+		{ "minecraft:tnt_minecart",        &cWSSAnvil::LoadMinecartTFromNBT },
+		{ "MinecartHopper",                &cWSSAnvil::LoadMinecartHFromNBT },
+		{ "minecraft:hopper_minecart",     &cWSSAnvil::LoadMinecartHFromNBT },
+		{ "MinecartRideable",              &cWSSAnvil::LoadMinecartRFromNBT },
+		{ "minecraft:minecart",            &cWSSAnvil::LoadMinecartRFromNBT },
+		{ "Item",                          &cWSSAnvil::LoadPickupFromNBT },
+		{ "minecraft:item",                &cWSSAnvil::LoadPickupFromNBT },
+		{ "Painting",                      &cWSSAnvil::LoadPaintingFromNBT },
+		{ "minecraft:painting",            &cWSSAnvil::LoadPaintingFromNBT },
+		{ "PrimedTnt",                     &cWSSAnvil::LoadTNTFromNBT },
+		{ "minecraft:tnt",                 &cWSSAnvil::LoadTNTFromNBT },
+		{ "XPOrb",                         &cWSSAnvil::LoadExpOrbFromNBT },
+		{ "minecraft:xp_orb",              &cWSSAnvil::LoadExpOrbFromNBT },
+		{ "ItemFrame",                     &cWSSAnvil::LoadItemFrameFromNBT },
+		{ "minecraft:item_frame",          &cWSSAnvil::LoadItemFrameFromNBT },
+		{ "Arrow",                         &cWSSAnvil::LoadArrowFromNBT },
+		{ "minecraft:arrow",               &cWSSAnvil::LoadArrowFromNBT },
+		{ "SplashPotion",                  &cWSSAnvil::LoadSplashPotionFromNBT },
+		{ "minecraft:potion",              &cWSSAnvil::LoadSplashPotionFromNBT },
+		{ "Snowball",                      &cWSSAnvil::LoadSnowballFromNBT },
+		{ "minecraft:snowball",            &cWSSAnvil::LoadSnowballFromNBT },
+		{ "Egg",                           &cWSSAnvil::LoadEggFromNBT },
+		{ "minecraft:egg",                 &cWSSAnvil::LoadEggFromNBT },
+		{ "Fireball",                      &cWSSAnvil::LoadFireballFromNBT },
+		{ "minecraft:fireball",            &cWSSAnvil::LoadFireballFromNBT },
+		{ "SmallFireball",                 &cWSSAnvil::LoadFireChargeFromNBT },
+		{ "minecraft:small_fireball",      &cWSSAnvil::LoadFireChargeFromNBT },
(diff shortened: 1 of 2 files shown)


src/WorldStorage/WSSAnvil.cpp
```diff
@@ -3167,7 +3183,10 @@ bool cWSSAnvil::LoadEntityBaseFromNBT(cEntity & a_Entity, const cParsedNBT & a_N
 bool cWSSAnvil::LoadMonsterBaseFromNBT(cMonster & a_Monster, const cParsedNBT & a_NBT, int a_TagIdx)
 {
 	float DropChance[5];
-	if (!LoadFloatsListFromNBT(DropChance, 5, a_NBT, a_NBT.FindChildByName(a_TagIdx, "DropChances")))
+	if (
+		!LoadFloatsListFromNBT(DropChance, 5, a_NBT, a_NBT.FindChildByName(a_TagIdx, "DropChances")) &&
+		!LoadFloatsListFromNBT(DropChance, 2, a_NBT, a_NBT.FindChildByName(a_TagIdx, "HandDropChances"))
```
Seadragon91: The new nbt has no DropChances` in it. There is `HandDropChances` (2 entries) and `ArmorDropChances` (4 entries).
Not sure how to fix that with the hands. Should I only read one for the hand?

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1472,17 +1472,33 @@ cBlockEntity * cWSSAnvil::LoadSignFromNBT(const cParsedNBT & a_NBT, int a_TagIdx
 
 
 
+bool cWSSAnvil::CheckEntityType(const char * a_IDTag, size_t a_IDTagLength, const AStringVector & a_ExpectedTypes)
+{
+	for (auto & et : a_ExpectedTypes)
```
madmaxoft: Use `const auto &`

src/WorldStorage/WSSAnvil.cpp
madmaxoft: I think it would be best if you separated the two codepaths completely:
```cpp
if (LoadFloatsListFromNBT(..., "DropChances")
{
  // old code
}
if (LoadFloatsListFromNBT(..., "HandDropChances")
{
  // Handle hands somehow - perhaps a sum?
}
if (LoadFloatsListFromNBT(..., "ArmorDropChances")
{
  // Set armor drop chances
}
```
(Note that no path returns false now, we've made all the tags optional)

src/WorldStorage/WSSAnvil.h
```diff
@@ -150,11 +150,14 @@ class cWSSAnvil :
 	This function extracts the text from either version. */
 	AString DecodeSignLine(const AString & a_Line);
 
-	/** Returns true iff the "id" child tag inside the specified tag equals (case-sensitive) any of the specified expected types.
+	/** Returns true if the "id" child tag inside the specified tag equals (case-sensitive) any of the specified expected types.
```
madmaxoft: Actually `iff` is the correct terminus technicus, being short for "if and only if". But whatever

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1472,17 +1472,33 @@ cBlockEntity * cWSSAnvil::LoadSignFromNBT(const cParsedNBT & a_NBT, int a_TagIdx
...
+}
+
+
+
+
+
 void cWSSAnvil::LoadEntityFromNBT(cEntityList & a_Entities, const cParsedNBT & a_NBT, int a_EntityTagIdx, const char * a_IDTag, size_t a_IDTagLength)
 {
-	if (strncmp(a_IDTag, "Boat", a_IDTagLength) == 0)
+	if (CheckEntityType(a_IDTag, a_IDTagLength, { "Boat", "minecraft:boat" }))
```
madmaxoft: These will probably have a poor performance, because you're creating the AStringVector each time when loading an entity. Consider pulling these out to a separate static const variable:
```cpp
  static const AStringVector entityTypesBoat({"Boat", "minecraft:boat"});
  // ...
  if (CheckEntityType(a_IDTag, a_IDTagLength, entityTypesBoat))
```

madmaxoft:
Now that I thought about this more, perhaps an even better approach than using a `const AStringVector &` with `CheckEntityType` would be to use something like this:
```cpp
typedef (cWSSAnvil::*EntityLoaderFunc)(cEntityList & a_Entities, const cParsedNBT & a_NBT, int a_EntityTagIdx);
typedef std::map<AString, EntityLoaderFunc> EntityLoaderMap;

static const EntityLoaderMap =
{ "Boat", &cWSSAnvil::LoadBoatFromNBT},
{ "minecraft:boat", &cWSSAnvil::LoadBoatFromNBT"},
...
};
```
Might need some syntax tweaking, but it would be faster (binary lookup instead of sequential) and look more elegant.

src/WorldStorage/WSSAnvil.cpp
Seadragon91: Yes that was just a hotfix, that I could test it. Will seperate it.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1490,11 +1490,26 @@ bool cWSSAnvil::CheckEntityType(const char * a_IDTag, size_t a_IDTagLength, cons
...
+		{ "Boat", &cWSSAnvil::LoadBoatFromNBT },
+		{ "minecraft:boat", &cWSSAnvil::LoadBoatFromNBT }
+
+	};
+
+	auto it = EntityTypeToFunction.find(a_IDTag);
+	if (it != EntityTypeToFunction.end())
+	{
+		(this->*it->second)(a_Entities, a_NBT, a_EntityTagIdx);
+	}
```
Seadragon91: Created the map and added boat. I do something wrong.
If I load the flat_world from vanilla, the boat is spawned. Restarted the server and the boat is not spawned anymore. The boat was saved a `Boat` in nbt, confirmed that.
Placed then a sheep and a boat in a hole, the sheep is spawned after restart.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1472,17 +1472,48 @@ cBlockEntity * cWSSAnvil::LoadSignFromNBT(const cParsedNBT & a_NBT, int a_TagIdx
...
 {
-	if (strncmp(a_IDTag, "Boat", a_IDTagLength) == 0)
+	static const EntityLoaderMap EntityTypeToFunction
+	{
+		{ "Boat", &cWSSAnvil::LoadBoatFromNBT },
+		{ "minecraft:boat", &cWSSAnvil::LoadBoatFromNBT }
+
+	};
+
+	auto it = EntityTypeToFunction.find(a_IDTag);
```
madmaxoft: `a_IDTag` is not NUL-terminated, so you'll need an explicit size when converting to `AString`. Use `...find(AString(a_IDTag, a_IDTagLength));`

Seadragon91:
Reworked it. Added in main post the link where I got the entitiy names.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1474,19 +1474,119 @@ cBlockEntity * cWSSAnvil::LoadSignFromNBT(const cParsedNBT & a_NBT, int a_TagIdx
...
+		{ "Witch", &cWSSAnvil::LoadWitchFromNBT },
+		{ "minecraft:witch", &cWSSAnvil::LoadWitchFromNBT },
+		{ "WitherBoss", &cWSSAnvil::LoadWitherFromNBT },
+		{ "minecraft:wither", &cWSSAnvil::LoadWitherFromNBT },
+		{ "Wolf", &cWSSAnvil::LoadWolfFromNBT },
+		{ "minecraft:wolf", &cWSSAnvil::LoadWolfFromNBT },
+		{ "Zombie", &cWSSAnvil::LoadZombieFromNBT },
+		{ "minecraft:zombie", &cWSSAnvil::LoadZombieFromNBT },
+		{ "PigZombie", &cWSSAnvil::LoadPigZombieFromNBT },
+		{ "minecraft:zombie_pigman", &cWSSAnvil::LoadPigZombieFromNBT },
```
madmaxoft: Align the function names vertically (use spaces)

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1474,19 +1474,119 @@ cBlockEntity * cWSSAnvil::LoadSignFromNBT(const cParsedNBT & a_NBT, int a_TagIdx
...
+	};
+
+	auto it = EntityTypeToFunction.find(AString(a_IDTag, a_IDTagLength));
+	if (it != EntityTypeToFunction.end())
+	{
+		(this->*it->second)(a_Entities, a_NBT, a_EntityTagIdx);
+	}
+
+	if (strncmp(a_IDTag, "Minecart", a_IDTagLength) == 0)
```
madmaxoft: You could add a new `LoadTypedMinecart` function that would handle these, and have everything in the map.

src/WorldStorage/WSSAnvil.h
```diff
@@ -92,6 +92,8 @@ class cWSSAnvil :
 		bool OpenFile(bool a_IsForReading);
 	} ;
 	typedef std::list<cMCAFile *> cMCAFiles;
+	typedef void (cWSSAnvil::*EntityLoaderFunc)(cEntityList &, const cParsedNBT &, int a_EntityTagIdx);
+	typedef std::map<AString, EntityLoaderFunc> EntityLoaderMap;
```
madmaxoft: These could be declared in the cpp file - they are not part of the loader's public API anyway.

madmaxoft:
That looks much better, doesn't it? :)

src/WorldStorage/WSSAnvil.cpp
Seadragon91: Done.

src/WorldStorage/WSSAnvil.cpp
Seadragon91: Used `LoadOldMinecartFromNBT` as name, as very old :)

Seadragon91:
Yes looks much better :)

src/WorldStorage/WSSAnvil.cpp
madmaxoft: Then you should probably remove this block as well :)

Seadragon91: Ops :)

PR #3808, opened 2017-06-24
Corrected APIDoc for cBlockArea and exported two missing functions

### LINKED ISSUE #3807: WorldEdit: attempt to call method 'SetRelBlockTypeMeta' (a nil value)
Client version:   1.11.2
Server OS:        Windows

### Expected behavior
WorldEdit works correctly.

### Actual behavior
WorldEdit fails because it uses `cBlockArea` methods that are missing bindings:
```
[592713a6|19:54:36] LUA: Plugins/WorldEdit/Classes/ShapeGenerator.lua:300: attempt to call method 'SetRelBlockTypeMeta' (a nil value)
[592713a6|19:54:36] Stack trace:
[592713a6|19:54:36]   Plugins/WorldEdit/Classes/ShapeGenerator.lua(300): SetBlock
[592713a6|19:54:36]   Plugins/WorldEdit/Classes/ShapeGenerator.lua(317): MakeSphere
[592713a6|19:54:36]   Plugins/WorldEdit/functions.lua(499): (no name)
[592713a6|19:54:36]   [C](-1): ChunkStay
[592713a6|19:54:36]   Plugins/WorldEdit/functions.lua(490): CreateSphereInCuboid
[592713a6|19:54:36]   Plugins/WorldEdit/Commands/Generation.lua(263): (no name)
[592713a6|19:54:36] Stack trace end
[592713a6|19:54:36] LUA: BreakIntoDebugger() not found / not a function
[592713a6|19:54:36] Error in <attached> calling function <callback>()
```

Checkout #3806 (master crashes before it gets to this point)
Use any of WorldEdit's `//{shape}` type commands e.g. `//sphere air 10`

Seadragon91:
Fixes #3807 

Done:
* Corrected param types and syntax for return types
* Added static flag to function
* Exported missing functions to lua

### DIFF
--- a/src/Bindings/ManualBindings_BlockArea.cpp
+++ b/src/Bindings/ManualBindings_BlockArea.cpp
@@ -835,7 +835,7 @@ static int GetRelBlock(lua_State * a_LuaState)
 
 	// Get the block info:
 	L.Push((self->*Fn)(coords.x, coords.y, coords.z));
-	return 0;
+	return 1;
 }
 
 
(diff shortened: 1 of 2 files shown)


Seadragon91:
Exported the two functions.

src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -951,6 +951,114 @@ static int SetRelBlock(lua_State * a_LuaState)
...
+
+	// Read the common params:
+	cBlockArea * self;
+	if (!L.GetStackValues(1, self))
+	{
+		return cManualBindings::ApiParamError(a_LuaState, "Cannot read the 'self' param.");
+	}
+
+	// Check the datatype's presence:
+	if ((self->GetDataTypes() & cBlockArea::baTypes) == 0)
```
madmaxoft: Use `HasBlockTypes` and `HasBlockMetas`, check both in the same condition.

src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -951,6 +951,114 @@ static int SetRelBlock(lua_State * a_LuaState)
...
+	{
+		return cManualBindings::ApiParamError(a_LuaState, "The area doesn't contain the datatype (%d).", cBlockArea::baMetas);
+	}
+
+	// Read the overloaded params:
+	Vector3i coords;
+	auto idx = readVector3iOverloadParams(L, 2, coords, "coords");
+	if (!self->IsValidCoords(coords))
+	{
+		return cManualBindings::ApiParamError(a_LuaState, "The coords ({%d, %d, %d}) are out of range ({%d, %d, %d}).",
```
madmaxoft: The error message should contain both lower and upper bounds for the range (`self->GetOrigin`, `self->GetOrigin() + self->GetSize()` )

src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -951,6 +951,114 @@ static int SetRelBlock(lua_State * a_LuaState)
...
+			coords.x, coords.y, coords.z,
+			self->GetSizeX(), self->GetSizeY(), self->GetSizeZ()
+		);
+	}
+
+	BLOCKTYPE Block;
+	L.GetStackValues(idx, Block);
+
+	NIBBLETYPE Meta;
+	L.GetStackValues(idx + 1, Meta);
```
madmaxoft: Read both `block` and `meta` at the same time, using `GetStackValues`; prefer lowerCamelCase.

src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -986,6 +1094,8 @@ void cManualBindings::BindBlockArea(lua_State * a_LuaState)
 			tolua_function(a_LuaState, "SetBlockSkyLight",        SetBlock<NIBBLETYPE, cBlockArea::baSkyLight, &cBlockArea::SetRelBlockSkyLight>);
 			tolua_function(a_LuaState, "SetRelBlockType",         SetRelBlock<BLOCKTYPE,  cBlockArea::baTypes,    &cBlockArea::SetRelBlockType>);
 			tolua_function(a_LuaState, "SetRelBlockMeta",         SetRelBlock<NIBBLETYPE, cBlockArea::baMetas,    &cBlockArea::SetRelBlockMeta>);
+			tolua_function(a_LuaState, "SetBlockTypeMeta",        tolua_cBlockArea_SetBlockTypeMeta);
```
madmaxoft: Alpha-sort these properly.

src/Bindings/ManualBindings_BlockArea.cpp
Seadragon91: Done

src/Bindings/ManualBindings_BlockArea.cpp
Seadragon91: I had copied the code from `SetRelBlock` there exists the same problem.

Seadragon91:
I had copied the code from `SetRelBlock` there exists the same problem with the size.

madmaxoft:
You need to differentiate between `Rel` and non-`Rel` functions. `Rel` functions operate on coords relative to the block area, the lower bound is always {0, 0, 0}, the upper bound is the area size. Non-`Rel` function operate on absolute coords (world coords from when the block area was read), and their lower bound is `m_Origin` and upper bound is `m_Origin + m_Size`. The code for `SetRelBlock` doesn't print the lower bound because it's always the same, all zeroes. `SetBlock` prints both bounds.

Seadragon91:
Here a 1 instead of a 0 should be returned:
[ManualBindings_BlockArea.cpp#L838](https://github.com/cuberite/cuberite/blob/master/src/Bindings/ManualBindings_BlockArea.cpp#L838)

Seadragon91:
With the functions `IsValidCoords` and `IsValidRelCoords` a out-of-bound read and write shouldn't occur anymore?

madmaxoft:
Yes, those check the coords for validity.

Seadragon91:
Are the changes now, correct? :)

src/BlockArea.cpp
```diff
@@ -507,7 +507,7 @@ bool cBlockArea::Read(cForEachChunkProvider & a_ForEachChunkProvider, int a_MinB
 	ASSERT(cChunkDef::IsValidHeight(a_MinBlockY));
 	ASSERT(cChunkDef::IsValidHeight(a_MaxBlockY));
 	ASSERT(a_MinBlockX <= a_MaxBlockX);
-	ASSERT(a_MinBlockY <= a_MaxBlockX);
+	ASSERT(a_MinBlockY <= a_MaxBlockY);
```
Seadragon91: Ah ops

src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -951,6 +951,105 @@ static int SetRelBlock(lua_State * a_LuaState)
...
+	cBlockArea * self;
+	if (!L.GetStackValues(1, self))
+	{
+		return cManualBindings::ApiParamError(a_LuaState, "Cannot read the 'self' param.");
+	}
+
+	// Check if block types and metas are present:
+	if (!self->HasBlockTypes() || !self->HasBlockMetas())
+	{
+		return cManualBindings::ApiParamError(a_LuaState, "The area doesn't contain the datatypes (%d) and (%d).", cBlockArea::baTypes, cBlockArea::baMetas);
```
madmaxoft: Why not just say that the area doesn't contain baTypes and baMetas? That's more understandable for plugin authors.

PR #3810, opened 2017-06-24
Fix cBlockArea assertion and mis-indexing

### LINKED ISSUE #3809: cChunkDef::MakeIndex(): coords out of range
Client version:  1.12
Server OS:        Windows

### Actual behavior
Tested the command from issue #3807 in a flat world, that I created in vanilla.
Run the command `//sphere air 10` at coords `-19 10 10` and this appears in console:
```
[23:53:36] cChunkDef::MakeIndex(): coords out of range: {17, 4, 1}; returning fake index 0
[23:53:36] cChunkDef::MakeIndex(): coords out of range: {19, 4, 1}; returning fake index 0
[23:53:36] cChunkDef::MakeIndex(): coords out of range: {21, 4, 1}; returning fake index 0
```

peterbell10:
Haven't tested but this should fix #3809 as it's the only use of `cChunkDef::MakeIndex` that's related to `cBlockArea`.  Also fixes an odd assertion I ran into.


Seadragon91:
Tested this and the error is gone, thanks :)

madmaxoft:
Thanks, and sorry for introducing the error.

PR #3811, opened 2017-06-25
Added check if datatype is present in block area

Seadragon91:
Fixes crash if a call is done with an datatype that is not present in block area.


src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -590,6 +590,12 @@ static int tolua_cBlockArea_RelLine(lua_State * a_LuaState)
 	{
 		return cManualBindings::ApiParamError(a_LuaState, "Invalid baDataTypes combination (%d).", dataTypes);
 	}
+	if ((self->GetDataTypes() & dataTypes) != dataTypes)
+	{
+		return cManualBindings::ApiParamError(a_LuaState, "Requesting datatypes not present in the cBlockArea. Got only 0x%02x, requested 0x%02x",
```
madmaxoft: -> `Requested`

Issue #3805, opened 2017-06-24
Use cManualBindings::ApiParamError for manual bindings


madmaxoft:
The `cManualBindings::ApiParamError` function is created for reporting errors in the API - be it invalid parameters or unreadable values or anything else. It provides diagnostic output. It should be used everywhere where currently Cuberite uses a `LOGWARN`, `LogStackTrace`, `return` sequence in the manual bindings.

Issue #3800, opened 2017-06-22
Use cLuaState::CheckParamSelf for all manual bindings


madmaxoft:
The `cLuaState::CheckParamSelf` function has been added to specifically check the first parameter of all class member functions, it has special reporting for errors ("use correct calling convention"). It should be used for all manual bindings. The `cBlockArea` bindings are an example of usage.

madmaxoft:
Additionally, a new `cLuaState::CheckParamStaticSelf` should be created for static functions (such as `cNetwork:Connect`)

madmaxoft:
This will not be needed anymore after the `ReadParams` update that seems to be coming along fine so far.
( https://forum.cuberite.org/thread-2727-post-28462.html#pid28462 )

pwnorbitals:
@madmaxoft bump, should this issue still exist ?

madmaxoft:
The `ReadParams` update never made it into the final code, so yes, this is still an issue that would be nice to address.

PR #3812, opened 2017-06-26
Stricter API checks in cNetwork

madmaxoft:
This is a part of #3805 and #3800, but the changes are relevant to code that is currently hot, so I want to merge them early.

Calling the `cNetwork` API functions with invalid parameters (such as port out of range) will now raise a Lua error, rather than returning with a console warning only.

### DIFF
--- a/src/Bindings/ManualBindings_Network.cpp
+++ b/src/Bindings/ManualBindings_Network.cpp
@@ -30,7 +30,7 @@ static int tolua_cNetwork_Connect(lua_State * L)
 
 	cLuaState S(L);
 	if (
-		!S.CheckParamUserTable(1, "cNetwork") ||
+		!S.CheckParamStaticSelf("cNetwork") ||
 		!S.CheckParamString(2) ||
 		!S.CheckParamNumber(3) ||
 		!S.CheckParamTable(4) ||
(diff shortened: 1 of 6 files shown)


src/Bindings/ManualBindings_Network.cpp
```diff
@@ -99,20 +92,13 @@ static int tolua_cNetwork_CreateUDPEndpoint(lua_State * L)
...
-		LOGWARNING("cNetwork:CreateUDPEndpoint() cannot read its parameters, failing the request.");
-		S.LogStackTrace();
-		S.LogStackValues("Values on the stack");
-		S.Push(false);
-		return 1;
+		return S.ApiParamError("Cannot read parameters");
 	}
 
 	// Check validity:
 	if ((port < 0) || (port > 65535))
```
Seadragon91: The variable port has the type `UInt16`. A number not in range of 0 - 65535 will be catched by `GetStackValues` above.
Example lua code:
`local Server = cNetwork:CreateUDPEndpoint(80000, {})`

madmaxoft: Right, but the error message is misleading. I'll change the `port `variable to `int`, at least for now, and keep the check

PR #3818, opened 2017-06-28
Update tolua++ and export EffectID

### LINKED ISSUE #3505: BroadcastSoundParticleEffect doesn't work

peterbell10:
This uses the new typed and scoped enum support in tolua++ to remove the "hackery" for the enum of  block ids.  Also exports `EffectID` and `SmokeDirection` which fixes #3505 .


Seadragon91:
Thanks for exporting will test it.

@madmaxoft Here circleci hangs, can you take a look? Thanks :)

Seadragon91:
Tested with world edit and works fine.

Seadragon91:
Ah one problem I see this enums are bound to the enum class name. Is it okay that way @madmaxoft ?

madmaxoft:
Looks good. No idea how to re-run the CircleCI build, other than @peterbell10 pushing another commit to this PR's branch

src/Bindings/BindingsProcessor.lua
```diff
@@ -977,7 +978,38 @@ local function installEnumHook()
 				enumValues[numEnumValues] = txt
 			end
 		end
-		local res = oldEnumerate(a_Name, "{" .. table.concat(enumValues, ",") .. "}", a_VarName)
+		local res = oldEnumerate(a_Name, "{" .. table.concat(enumValues, ",") .. "}", a_VarName, a_Type)
+		res.DoxyComments = doxyComments
+		return res
+	end
+	--Hook for scoped enums
```
madmaxoft: Add an empty line above here.

madmaxoft:
@Seadragon91 what exactly do you mean by that?

Seadragon91:
The monster type can be directly accessed by doing `mtSheep` for this ones you need to do `EffectID.SFX_RANDOM_DISPENSER_DISPENSE`

madmaxoft:
Yes, I believe that's better

peterbell10:
I've pushed with the added empty line but it looks like circleci still isn't working:
`Your build was not run - reason code (:github-failure)`

madmaxoft:
This time the build was visible on CircleCI, so I restarted it and it seems to work.

Seadragon91:
I also okay with that, but I think a example should be added. As all current global enums can be accessed directly and this ones cold confuse the dev :)

Seadragon91:
Oh sorry then, my mistake...

PR #3820, opened 2017-06-28
Fix horse taming

### LINKED ISSUE #2924: Horses do not react to being tamed properly.
Before a horse is teamed it can still be controlled by the player. When it is being controlled it can move much faster than it should be able to, as well as looking towards one point at all times, not the direction you are moving. It will buck off correctly.

After a few tries the horse will no longer buck you off, I think this is where the horse should be tamed but the particles do not appear. The horse moves in the same way as it did before.

After the horse reaches the "tamed" point and you dismount any attempts to mount the horse will result in the horse not being mounted and the camera being pointed in the direction that the horse is facing.

See this video to explain further.

https://youtu.be/3ERYeLDj5VI?t=7s

Bond-009:
Fixes #2924

### DIFF
--- a/src/Mobs/Horse.cpp
+++ b/src/Mobs/Horse.cpp
@@ -129,7 +129,7 @@ void cHorse::OnRightClicked(cPlayer & a_Player)
 			a_Player.AttachTo(this);
 		}
 	}
-	else
+	else if (a_Player.GetEquippedItem().IsEmpty())
 	{
 		if (m_Attachee != nullptr)
 		{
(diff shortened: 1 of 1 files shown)


src/Mobs/Horse.cpp
```diff
@@ -104,23 +105,28 @@ void cHorse::OnRightClicked(cPlayer & a_Player)
...
+				a_Player.AttachTo(this);
 			}
-			m_bIsSaddled = true;
-			m_World->BroadcastEntityMetadata(*this);
 		}
-		else if (!a_Player.GetEquippedItem().IsEmpty())
+		else
 		{
-			// The horse doesn't like being hit, make it rear:
-			m_bIsRearing = true;
```
spekdrum: The logic handling horse being hit has been removed. Is that on purpose?

Bond-009: When you right click a untamed horse without a saddle you should mount it.

spekdrum: That's true. How about overriding DoTakeDamage?

Bond-009: I found the proper place and added the sound :)

PR #3821, opened 2017-06-28
Fixes double right click on entities till off-hand actions gets implemented

### LINKED ISSUE #3186: Right clicking a tamed wolf will not make it sit

spekdrum:
Fixes #3186
When you right click on a aggressive monster (does not happens on passive mobs) this code gets executed twice:

```
void cProtocol_1_9_0::HandlePacketUseEntity(cByteBuffer & a_ByteBuffer)
{
	HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, EntityID);
	HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Type);

	switch (Type)
	{
		case 0:
		{
			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand)
			m_Client->HandleUseEntity(EntityID, false);
			break;
		}
```

The "Hand" variable first time is 0 and second time is 1. According to the protocol doc (http://wiki.vg/Protocol#Use_Entity) that means main hand (0) or off-hand (1). The reason is that client sends an action with main hand and an action with the off-hand. 

This results on double actions. Affects mob leashing (in development) and right clicking a tamed wolf issue (#3186), maybe others.

Maybe we should check which hand holds the item we want to use with the entity, but we need to change the code a bit in order to let this parameter arrive to every place. 

This fix ignores calls for the off-hand. This way we can get rid of those double action related issues till a complete solution is done.

### DIFF
--- a/src/Protocol/Protocol_1_9.cpp
+++ b/src/Protocol/Protocol_1_9.cpp
@@ -2666,7 +2666,7 @@ void cProtocol_1_9_0::HandlePacketUseEntity(cByteBuffer & a_ByteBuffer)
 		case 0:
 		{
 			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand);
-			if (Hand == 0)  // TODO: implement handle off-hand (1) actions
+			if (Hand == hMain)  // TODO: implement handling of off-hand actions; ignore them for now to avoid processing actions twice
 			{
 				m_Client->HandleUseEntity(EntityID, false);
 			}
--- a/src/Protocol/Protocol_1_9.cpp
+++ b/src/Protocol/Protocol_1_9.cpp
@@ -2665,8 +2665,11 @@ void cProtocol_1_9_0::HandlePacketUseEntity(cByteBuffer & a_ByteBuffer)
 	{
 		case 0:
 		{
-			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand)
-			m_Client->HandleUseEntity(EntityID, false);
+			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand);
+			if (Hand == 0)  // TODO: implement handle off-hand (1) actions
+			{
+				m_Client->HandleUseEntity(EntityID, false);
+			}
 			break;
 		}
 		case 1:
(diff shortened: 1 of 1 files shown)


src/Protocol/Protocol_1_9.cpp
```diff
@@ -2665,8 +2665,11 @@ void cProtocol_1_9_0::HandlePacketUseEntity(cByteBuffer & a_ByteBuffer)
 	{
 		case 0:
 		{
-			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand)
-			m_Client->HandleUseEntity(EntityID, false);
+			HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, Hand);
+			if (Hand == 0)  // TODO: implement handle off-hand (1) actions
```
Pokechu22: I'm not sure if this comment makes it completely clear why we're ignoring it.  Maybe something like

```c++
// TODO: implement handling of off-hand actions; ignore them for now to avoid processing actions twice
```

would be better.

Pokechu22: We actually do have an `eHand` enum (not to be confused with `eMainHand`); consider using `hMain` and `hOff` instead of the numbers.

madmaxoft: I don't think that's such a good idea. The protocol should not depend on internal Cuberite representations. If we ever decide to change the enum for any reason, we'll suddenly and quietly break this protocol.

Pokechu22: The enum matches the protocol (I added it for that purpose and use `eMainHand` the same way in client settings).  Protocol-wise, the game is serializing an enum by its index (that's what "VarInt Enum" on the wiki indicates), not converting an internal representation to a number, so I think it's fine to also use an enum in cuberite.

madmaxoft: It currently matches *this* protocol. It may not match a future protocol if Mojang decide to mess with it. I believe it would be better to use a named value that is specific to this protocol - a `static const` value declared either at the top of this function, or the top of this file, so that it's local and constant.

PR #3823, opened 2017-06-30
Added bed entity

### LINKED ISSUE #3787: Beds are invisible after re-login
Client version:   1.12
Server OS:        macOS
Commit id:        bb020bbe78434a221f6d404d7f83e5bd804ed541

### Expected behavior
Beds should be visble after re-login.

### Actual behavior
Beds are invisible after re-login.

Place a bed, re-login -> the bed is invisible.

Seadragon91:
Fixes #3787

Done:
* Export cBedEntity to lua
* Set color of bed through item damage value
* Added bed to APIDoc
* Add saving / loading of color
* Add recipe(s) for crafting colored bed
* Destroy the bed entity of the other part
* Overloaded method `ConvertToPickups` with params coords and digger to have a way to access the block entity
* Fix sending color of bed to player

~I am not happy with the current block entity name. The bed normally takes two blocks and there is now a `cBedEntity` for every bed block. Whats about using `cBedBlockEntity` or `cBedPartEntity` as name?~

~Have also a problem with destroying the bed entity of the other part of the bed. Destroying the block entity in `cBlockBedHandler::OnDestroyed` results in a stack overflow. Any good ideas?~

### DIFF
--- a/Server/crafting.txt
+++ b/Server/crafting.txt
@@ -383,7 +383,6 @@ AcaciaFenceGate          = Stick, 1:1, 1:2, 3:1, 3:2 | AcaciaPlanks, 2:1, 2:2
 Anvil                    = IronBlock, 1:1, 2:1, 3:1 | IronIngot, 2:2, 1:3, 2:3, 3:3
 ArmorStand               = Stick, 1:1, 1:3, 2:1, 2:2, 3:1, 3:3 | StoneSlab, 2:3
 Beacon                   = Glass, 1:1, 1:2, 2:1, 3:1, 3:2 | Obsidian, 1:3, 2:3, 3:3 | NetherStar, 2:2
-Bed                      = Planks^-1, 1:2, 2:2, 3:2 | Wool^-1, 1:1, 2:1, 3:1
 BirchFence, 3            = BirchPlanks, 1:1, 1:2, 3:1, 3:2 | Stick, 2:1, 2:2
 BirchFenceGate           = Stick, 1:1, 1:2, 3:1, 3:2 | BirchPlanks, 2:1, 2:2
 Bookandquill             = Book, * | feather, * | inksac, *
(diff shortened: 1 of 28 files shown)


Seadragon91:
Fixed destroying the other bed part. I wish I could drop the bed in `ConvertToPickups`, but the problem is to get the color of the bed.

Seadragon91:
Fixed the stack overflow problem by first using `FastSetBlock` to replace the bed with air then `SetBlock` to destroy the bed entity.

Seadragon91:
Overloaded method `ConvertToPickups` to have the params coords and the digger to access the block entity. With that I can get the color. The bed is really a special case :)

Seadragon91:
I added now a scheduled task to delay broadcasting the bed entity. My assumption is that the block entity reaches the client to quickly and it maybe has not the bed, atm.

Server/crafting.txt
```diff
@@ -931,3 +930,24 @@ FireworkStar     = FireworkStar, * | Dye ^-1, * | Dye ^-1, * | Dye ^-1, * | Dye
...
+Bed^6                    = Planks^-1, 1:2, 2:2, 3:2 | Wool^6, 1:1, 2:1, 3:1
+Bed^7                    = Planks^-1, 1:2, 2:2, 3:2 | Wool^7, 1:1, 2:1, 3:1
+Bed^8                    = Planks^-1, 1:2, 2:2, 3:2 | Wool^8, 1:1, 2:1, 3:1
+Bed^9                    = Planks^-1, 1:2, 2:2, 3:2 | Wool^9, 1:1, 2:1, 3:1
+Bed^10                   = Planks^-1, 1:2, 2:2, 3:2 | Wool^10, 1:1, 2:1, 3:1
+Bed^11                   = Planks^-1, 1:2, 2:2, 3:2 | Wool^11, 1:1, 2:1, 3:1
+Bed^12                   = Planks^-1, 1:2, 2:2, 3:2 | Wool^12, 1:1, 2:1, 3:1
+Bed^13                   = Planks^-1, 1:2, 2:2, 3:2 | Wool^13, 1:1, 2:1, 3:1
+Bed^14                   = Planks^-1, 1:2, 2:2, 3:2 | Wool^14, 1:1, 2:1, 3:1
+Bed^15                   = Planks^-1, 1:2, 2:2, 3:2 | Wool^15, 1:1, 2:1, 3:1
```
madmaxoft: We should really think of a way to support multiple recipe files for multiple protocols - at least so that the admin can choose which version to use.

madmaxoft: Or maybe think of a different format that would support versioning.

src/Blocks/BlockBed.cpp
```diff
@@ -145,4 +159,72 @@ bool cBlockBedHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterface
...
+void cBlockBedHandler::ConvertToPickups(cEntity * a_Digger, cItems & a_Pickups, NIBBLETYPE a_BlockMeta, int a_BlockX, int a_BlockY, int a_BlockZ)
+{
+	class cBedColor :
+		public cBedCallback
+	{
+	public:
+		short m_Color;
+
+		cBedColor()
+		{
```
madmaxoft: I think you should initialize `m_Color` here with a default value, just in case the callback fails for some reason.

madmaxoft:
I think the `cBedEntity` name is good enough.

src/BlockEntities/BedEntity.h
```diff
@@ -0,0 +1,46 @@
...
+	// tolua_begin
+
+	/** Returns the color of the bed */
+	short GetColor(void) const { return m_Color; }
+
+	/** Set the color of the bed. */
+	void SetColor(short a_Color) { m_Color = a_Color; }
+
+	/** Returns true if this bed block has the meta for head set */
+	bool IsHead(void) { return ((m_BlockMeta & 0x08) == 0x08); }
```
Seadragon91: I remove the bool and directly check now if the head meta is set. One problem I found is If I change the meta with `cWorld:SetBlockMeta` the meta in the block entity is not updated. The methods would need to update the block meta  of the block entity.

src/Blocks/BlockBed.cpp
Seadragon91: Set default color to red.

src/BlockEntities/BedEntity.h
madmaxoft: The name is somewhat confusing. How about `IsHeadPiece` or `IsPillowBlock` or something like that?

madmaxoft: Is that needed for regular vanilla behavior, or just for plugins that change things around? I think propagating meta changes to the block entity would be better in a separate PR.

Seadragon91: For plugins and I see that the meta is also used in cloning or furnace if it's lit. Yes better in a new pull request.

Seadragon91: Changed to `IsPillowBlock`

Seadragon91: Should I set in in `cChunk::SetMeta` or should we change the block entities to always get the block meta from the world?

madmaxoft:
Do I read this right that the bed entity is created for both bed blocks? What happens, if each one specifies a different color? :)

Seadragon91:
Then you have a bed with different colors :).

I just see that I have a problem with `SetColor`. Need to broadcast then the bed entity, but it's not good as it's also called if a bed is loaded from chunk.

madmaxoft:
Perhaps the loader should not use that function, then.

Seadragon91:
Well I need to set the color :) Whats about adding it to the constructor optional?

Seadragon91:
@madmaxoft Can I merge :)

madmaxoft:
One question: Does a 1.9 client survive this change?

Seadragon91:
The item drop is a problem for all clients below 1.12...

madmaxoft:
I suppose that's acceptable.

PR #3824, opened 2017-06-30
Players in creative and spectator aren't affected by cacti

### LINKED ISSUE #2916: Cannot move well when touching a cactus.
When you run into a cactus you are slowed by more than you should be. Along with being slowed you cannot jump. This also affects people in creative mode even though the movement speed slowing should only affect those in survival or hardcore.

Bond-009:
Fixes #2916

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1357,6 +1357,15 @@ void cEntity::TickInVoid(cChunk & a_Chunk)
 
 void cEntity::DetectCacti(void)
 {
+	if (
+		IsPlayer() &&
+		((reinterpret_cast<cPlayer *>(this))->IsGameModeCreative() || (reinterpret_cast<cPlayer *>(this))->IsGameModeSpectator())
+	)
+	{
+		// Ignore Cacti if the entity is player in creative or spectator mode
+		return;
+	}
+
 	int X = POSX_TOINT, Y = POSY_TOINT, Z = POSZ_TOINT;
 	double w = m_Width / 2;
 	if (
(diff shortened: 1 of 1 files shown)


src/Entities/Entity.cpp
```diff
@@ -1357,6 +1357,14 @@ void cEntity::TickInVoid(cChunk & a_Chunk)
 
 void cEntity::DetectCacti(void)
 {
+	if (
+		IsPlayer() && 
+		((reinterpret_cast<cPlayer *>(this))->IsGameModeCreative() || (reinterpret_cast<cPlayer *>(this))->IsGameModeSpectator()))
```
Seadragon91: Please change to:
```
if (
	IsPlayer() &&
	((reinterpret_cast<cPlayer *>(this))->IsGameModeCreative() || (reinterpret_cast<cPlayer *>(this))->IsGameModeSpectator())
)
```
The closing round  bracket on a new line.

Bond-009:
What's wrong with CircleCI?

Seadragon91:
No idea...

Seadragon91:
@Bond-009 Please fix the trailing white space. Hope now that circle builds, changed a setting.

Seadragon91:
Okay still doesn't work :(

Bond-009:
Maybe I should move the check to this line.
https://github.com/Bond-009/cuberite/blob/857ca0fa4ae5eab94547a41193b32585c093e7b8/src/Entities/Entity.cpp#L914

Seadragon91:
Thanks for moving the check :)

PR #3826, opened 2017-06-30
BigFlower fixes

### LINKED ISSUE #3606: Right clicking rose bushes while holding a rose bush
Client version:   1.9
Server OS:        Linux
Commit id:        83fa6cb851536839ebd85b1e91fda928f47eb674

### Expected behavior

Right clicking rose bushes while holding rose bush should not drop any plants.

### Actual behavior

However, what actually happens is a rose bush and sunflower ends up being dropped

Right click a rose bush while holding a rose bush

### LINKED ISSUE #3825: Top of rose bush and peony are not visible on client
Client version:   confirmed with 1.9.0 and 1.11.2
Server OS:        Windows
Commit id:        3352706

### Expected behavior
Both blocks are visible even after relog.

### Actual behavior
The bottom block is visible but the top is not.  If you place a block where the top should be it is there on the client but disappears on relog.
Note this happens for rose bushes and peony but none of the other big flowers.

Place a rose bush or peony and relog.

peterbell10:
This fixes #3606 and a couple other issues I noticed while testing it.

* Lilacs ignored build collision on server but not client causing desyncs.
* Big flowers could be placed on any solid block but `CanBeAt` only allows them on dirt types.
* Placing big flowers under another block would overwrite the block.

Edit: Now also fixes #3825

### DIFF
--- a/src/Blocks/BlockHandler.h
+++ b/src/Blocks/BlockHandler.h
@@ -111,15 +111,7 @@ public:
 	/** Checks if the player can build "inside" this block.
 	For example blocks placed "on" snow will be placed at the same position. So: Snow ignores Build collision
 	*/
-	virtual bool DoesIgnoreBuildCollision(void);
-
-	/** Similar to DoesIgnoreBuildCollision(void), but is used for cases where block's meta or
-	player's item-in-hand is needed to determine collision (thin snow) */
-	virtual bool DoesIgnoreBuildCollision(cPlayer *, NIBBLETYPE a_Meta)
-	{
-		UNUSED(a_Meta);
-		return DoesIgnoreBuildCollision();
-	}
+	virtual bool DoesIgnoreBuildCollision(cChunkInterface & ChunkInterface, Vector3i a_RelPos, cPlayer & a_Player, NIBBLETYPE a_Meta);
 
 	/** Returns if this block drops if it gets destroyed by an unsuitable situation.
 	Default: true */
--- a/src/Blocks/BlockHandler.h
+++ b/src/Blocks/BlockHandler.h
@@ -110,8 +110,11 @@ public:
 
 	/** Checks if the player can build "inside" this block.
 	For example blocks placed "on" snow will be placed at the same position. So: Snow ignores Build collision
+	@param a_Pos Position of the block
+	@param a_Player Player trying to build on the block
+	@param a_Meta Meta value of the block currently at a_Pos
 	*/
-	virtual bool DoesIgnoreBuildCollision(cChunkInterface & ChunkInterface, Vector3i a_RelPos, cPlayer & a_Player, NIBBLETYPE a_Meta);
+	virtual bool DoesIgnoreBuildCollision(cChunkInterface & ChunkInterface, Vector3i a_Pos, cPlayer & a_Player, NIBBLETYPE a_Meta);
 
 	/** Returns if this block drops if it gets destroyed by an unsuitable situation.
 	Default: true */
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -15493,6 +15493,10 @@ end
 				{
 					Notes = "The player has swapped their held item with the item in their offhand slot (1.9)",
 				},
+				E_META_BIG_FLOWER_TOP =
+				{
+					Notes = "The metadata of a big flower block that indicates it is the top block.",
+				},
 				E_META_CONCRETE_BLACK =
 				{
 					Notes = "A flag in the metadata of concete that indicates that the concrete is black.",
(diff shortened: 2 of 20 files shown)


Seadragon91:
Circleci is really strange. I see after looking at the pull request, that it runs the test if the contributor is at least a member of cuberite...

Seadragon91:
If I replay a large fern or double grass with peony only the bottom part is placed. It's not a client glitch.

peterbell10:
There is no code path that only places one of the two blocks, are you sure it isn't #3825 ?  Does the same thing happen on master?

Seadragon91:
Ah didn't saw the issue

Seadragon91:
@madmaxoft Is the change with `DoesIgnoreBuildCollision` that has now params okay for you?

src/Blocks/BlockHandler.h
```diff
@@ -111,15 +111,7 @@ class cBlockHandler
...
-	virtual bool DoesIgnoreBuildCollision(void);
-
-	/** Similar to DoesIgnoreBuildCollision(void), but is used for cases where block's meta or
-	player's item-in-hand is needed to determine collision (thin snow) */
-	virtual bool DoesIgnoreBuildCollision(cPlayer *, NIBBLETYPE a_Meta)
-	{
-		UNUSED(a_Meta);
-		return DoesIgnoreBuildCollision();
-	}
+	virtual bool DoesIgnoreBuildCollision(cChunkInterface & ChunkInterface, Vector3i a_RelPos, cPlayer & a_Player, NIBBLETYPE a_Meta);
```
madmaxoft: The parameters are not too intuitive, the doxycomment should explain what they are.

madmaxoft:
Having one function is definitely better, and the params make sense, so I'm for that change. Just document stuff :)

src/Blocks/BlockHandler.h
Seadragon91: Is the vector changed? if not it could be changed to a const ref.

src/Blocks/BlockHandler.h
```diff
@@ -110,8 +110,11 @@ class cBlockHandler
 
 	/** Checks if the player can build "inside" this block.
 	For example blocks placed "on" snow will be placed at the same position. So: Snow ignores Build collision
+	@param a_Pos Position of the block
+	@param a_Player Player trying to build on the block
+	@param a_Meta Meta value of the block
```
madmaxoft: The block currently at that place, or the block being placed?

src/Blocks/BlockHandler.h
peterbell10: The vector isn't changed but it's just 3 ints. Not sure it's worth saving a copy but forcing out of registers anyway.

madmaxoft: `const Vector3i` in the very least; not sure whether a ref would make noticeable difference and which way :P

peterbell10: Look at [this](https://godbolt.org/g/rCa68n), GCC packs the vector into two registers vs forcing it onto the stack.

madmaxoft: Good for x64, but I'm not sure about ARM, don't know the architecture enough to decipher the disassembly.

peterbell10: Looks like the 32-bit arm version does keep the arguments on the stack but arm64 does the same as x64. Though in this case it's mostly called with a temporary in which case the reference is strictly more work regardless of platform.

src/Blocks/BlockHandler.h
peterbell10: Was it just the meta that was unclear or was it all of them?

madmaxoft: Just the meta.
Just curious, why not make the position parameter `const`? It could give the compiler more room for optimizations.

peterbell10: I may be wrong but I was under the impression that top level `const` is ignored for function parameters.

madmaxoft: It is ignored when calling the function, but in the function definition, it makes the object `const`, so the compiler knows (and enforces) that it cannot be changed.

peterbell10: Just to be clear, do you only want `const` in the definitions or do you want the declaration to have `const` as well?

madmaxoft:
```cpp
virtual bool DoesIgnoreBuildCollision(cChunkInterface & ChunkInterface, const Vector3i a_Pos, cPlayer & a_Player, NIBBLETYPE a_Meta);
```
The definition and declaration have to match, of course.

madmaxoft: Okay, that's news to me. Reading through some discussions on the topic, I think this is pointless anyway, so let's just keep it as it is.

PR #3828, opened 2017-07-01
You can't use beds when there are hostile mobs nearby

### LINKED ISSUE #705: You can use beds when monsters are nearby
You can sleep in a bed although there are monsters nearby. 
Normally, when a monster is in range of 8 horizontally and 5 vertically you can't use a bed and you get the message "You may not rest now, there are monsters nearby."

Bond-009:
Fixes #705

### DIFF
--- a/src/Blocks/BlockBed.cpp
+++ b/src/Blocks/BlockBed.cpp
@@ -4,6 +4,9 @@
 #include "BroadcastInterface.h"
 #include "Entities/../World.h"
 #include "Entities/Player.h"
+#include "../BoundingBox.h"
+#include "../Mobs/Monster.h"
+#include "../Entities/Entity.h"
 
 
 
@@ -32,6 +35,22 @@ void cBlockBedHandler::OnDestroyed(cChunkInterface & a_ChunkInterface, cWorldInt
 		}
 	}
 }
+
+
+
+
+
+class cFindMobs :
+	public cEntityCallback
+{
+	virtual bool Item(cEntity * a_Entity) override
+	{
+		return (
+			(a_Entity->GetEntityType() == cEntity::etMonster) &&
+			(static_cast<cMonster*>(a_Entity)->GetMobFamily() == cMonster::mfHostile)
+		);
+	}
+};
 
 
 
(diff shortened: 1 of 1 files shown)


Seadragon91:
I think the detection for the height needs to be adjusted. In cBoundingBox-API I read that the height starts from minY to maxY + height. A monster placed nearby a bed with a height diff of 3 blocks deep is not detected.

Seadragon91:
Currently it's minY to maxY + 5, should be minY - 5 to maxY + 5 or?

Bond-009:
So the code in cBoundingBox needs to be changed?

spekdrum:
I introduced an overloaded constructor for cBoundingBox in PR #3798 that allows to do that:
```
cBoundingBox::cBoundingBox(const Vector3d & a_Pos, double a_Radius, double a_Height, double a_VerticalOffset) :
 	m_Min(a_Pos.x - a_Radius, a_Pos.y + a_VerticalOffset, a_Pos.z - a_Radius),
+	m_Max(a_Pos.x + a_Radius, a_Pos.y + a_VerticalOffset + a_Height, a_Pos.z + a_Radius)
 {
 }
```

Maybe you can copy it, but it's not revised yet.

Seadragon91:
@Bond-009 cBoundingBox is correct and the API also. You could change it to this:
`a_Player->GetPosition()`
to
`a_Player->GetPosition() - Vector3i(0, 5, 0)`
This will subtract 5 blocks from the start pos and then it's okay.

Bond-009:
Ok, I wasn't sure if that was intended :)

src/Blocks/BlockBed.cpp
```diff
@@ -88,14 +107,23 @@ bool cBlockBedHandler::OnUse(cChunkInterface & a_ChunkInterface, cWorldInterface
...
+		if ((Meta & 0x4) == 0x4)
+		{
+			a_Player->SendMessageFailure("This bed is occupied");
+		}
+		else
 		{
-			NIBBLETYPE Meta = a_ChunkInterface.GetBlockMeta(a_BlockX, a_BlockY, a_BlockZ);
-			if ((Meta & 0x4) == 0x4)
+			cFindMobs FindMobs;
+			if (!a_Player->GetWorld()->ForEachEntityInBox(cBoundingBox(a_Player->GetPosition() - Vector3i(0, 5, 0), 8, 10), FindMobs))
```
Seadragon91: Why changed `8, 5` to `8, 10`? Now it its startY - 5 to startY + 10

Bond-009: Yes, that should be a 5 blocks above and under the player, no?

Seadragon91: Ah ops, sorry :) Really confused me hehe

Seadragon91:
For closing a issue via a [keyword](https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues) it has to be followed by the number.
Change it to `Fixes #705`

Bond-009:
Cool, I didn't know that :+1:

Issue #3822, opened 2017-06-28
Optimize Tick logic reducing the frequency in certain cases


spekdrum:
Client version:   any
Server OS:        any
Commit id:       3f7bf88

There is a lot of Tick logic executing 20 times per second. In most cases it would be not necessary and could be reduced to twice per second or so. For instance, all the mob's logic involved in path finding.
This will contribute Cuberite to be even more efficient.

peterbell10:
I believe the path finder incrementally calculates the complete path over several ticks.  If it were called less frequently it would just have more work to do in a single tick and could make the server less responsive.

As for the general approach it's an interesting idea but I can see some potential issues:

* If updates are too infrequent it could give the impression the server is lagging.

* If done naively, ticks which are highly divisible would have spikes in workload.

* If the decision whether to tick is made just before the actual tick logic then it might not even be much faster as the instructions and data still have to be loaded into cache.

The first issue I suppose just requires discipline in choosing frequency but the others would probably need some kind of tick scheduling that could spread the work over the ticks and selectively call the code rather than having tick methods that bail out when it isn't the right time.

NiLSPACE:
We could check if the chunks in which a path is calculated have changed and only recalculate when necessary.

spekdrum:
Maybe everything inside Tick logic involving callbacks for entity processing, finding closest player, states change, etc. could be the candidate code to be less frequently executed.

PR #3829, opened 2017-07-02
Added basic ocelot behavior

### LINKED ISSUE #1379: Chests blocked from openning
Chests are blocked from openning when hoppers are above them. You also can't stack chests like you can on the vanila server.

### LINKED ISSUE #2367: Support ocelot type
We could support ocelot type:
- [ ] wild
- [ ] black
- [ ] red
- [ ] siamese

Bond-009:
Fixes #1379

~~At the moment the ocelot isn't correctly saved to the disk, when you restart the server it's the basic type and untamed. I have no idea where/how to fix this.~~

EDIT: now also fixes #2367

### DIFF
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -243,7 +243,7 @@ bool cChestEntity::IsBlocked()
 		(GetPosY() >= cChunkDef::Height - 1) ||
 		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
 		(
-			(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) && 
+			(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) &&
 			!GetWorld()->ForEachEntityInBox(cBoundingBox(Vector3d(GetPosX(), GetPosY() + 1, GetPosZ()), 1, 1), FindSittingCat)
 		)
 	);
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -6,6 +6,8 @@
 #include "../Entities/Player.h"
 #include "../UI/ChestWindow.h"
 #include "../ClientHandle.h"
+#include "../Mobs/Ocelot.h"
+#include "../BoundingBox.h"
 
 
 
@@ -217,11 +219,33 @@ void cChestEntity::DestroyWindow()
 
 
 
+class cFindSittingCat :
+	public cEntityCallback
+{
+	virtual bool Item(cEntity * a_Entity) override
+	{
+		return (
+			(a_Entity->GetEntityType() == cEntity::etMonster) &&
+			(static_cast<cMonster*>(a_Entity)->GetMobType() == eMonsterType::mtOcelot) &&
+			(static_cast<cOcelot*>(a_Entity)->IsSitting())
+		);
+	}
+};
+
+
+
+
+
 bool cChestEntity::IsBlocked()
 {
-	// TODO: cats are an obstruction
+	cFindSittingCat FindSittingCat;
 	return (
 		(GetPosY() >= cChunkDef::Height - 1) ||
-		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()))
+		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
+		((GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) && !GetWorld()->ForEachEntityInBox(cBoundingBox(Vector3d(GetPosX(), GetPosY() + 1, GetPosZ()), 1, 1), FindSittingCat))
 	);
 }
+
+
+
+
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -226,8 +226,8 @@ class cFindSittingCat :
 	{
 		return (
 			(a_Entity->GetEntityType() == cEntity::etMonster) &&
-			(static_cast<cMonster*>(a_Entity)->GetMobType() == eMonsterType::mtOcelot) &&
-			(static_cast<cOcelot*>(a_Entity)->IsSitting())
+			(static_cast<cMonster *>(a_Entity)->GetMobType() == eMonsterType::mtOcelot) &&
+			(static_cast<cOcelot *>(a_Entity)->IsSitting())
 		);
 	}
 };
@@ -242,7 +242,10 @@ bool cChestEntity::IsBlocked()
 	return (
 		(GetPosY() >= cChunkDef::Height - 1) ||
 		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
-		((GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) && !GetWorld()->ForEachEntityInBox(cBoundingBox(Vector3d(GetPosX(), GetPosY() + 1, GetPosZ()), 1, 1), FindSittingCat))
+		(
+			(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) && 
+			!GetWorld()->ForEachEntityInBox(cBoundingBox(Vector3d(GetPosX(), GetPosY() + 1, GetPosZ()), 1, 1), FindSittingCat)
+		)
 	);
 }
 
--- a/src/Mobs/Ocelot.cpp
+++ b/src/Mobs/Ocelot.cpp
@@ -16,7 +16,7 @@ cOcelot::cOcelot(void) :
 	m_IsSitting(false),
 	m_IsTame(false),
 	m_IsBegging(false),
-	m_Type(0),
+	m_CatType(ctWildOcelot),
 	m_OwnerName("")
 {
 }
@@ -151,7 +151,9 @@ void cOcelot::OnRightClicked(cPlayer & a_Player)
 					a_Player.GetInventory().RemoveOneEquippedItem();
 				}
 
-				if (GetRandomProvider().RandBool(1.0 / 3.0))
+				auto & Random = GetRandomProvider();
+
+				if (Random.RandBool(1.0 / 3.0))
 				{
 					// Taming succeeded
 					SetIsBegging(false);
@@ -159,7 +161,7 @@ void cOcelot::OnRightClicked(cPlayer & a_Player)
 					SetMaxHealth(20);
 					SetIsTame(true);
 					SetOwner(a_Player.GetName(), a_Player.GetUUID());
-					SetOcelotType(GetRandomProvider().RandInt<int>(1, 3));
+					SetCatType(static_cast<eCatType>(Random.RandInt<int>(1, 3)));
 					m_World->BroadcastEntityStatus(*this, esWolfTamed);
 					m_World->GetBroadcaster().BroadcastParticleEffect("heart", static_cast<Vector3f>(GetPosition()), Vector3f{}, 0, 5);
 				}
--- a/src/Mobs/Ocelot.cpp
+++ b/src/Mobs/Ocelot.cpp
@@ -129,7 +129,7 @@ void cOcelot::OnRightClicked(cPlayer & a_Player)
 	if (!IsTame())
 	{
 		if (
-			IsBegging() && 
+			IsBegging() &&
 			((a_Player.GetPosition() - GetPosition()).Length() <= 3)
 		)
 		{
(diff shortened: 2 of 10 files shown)


Seadragon91:
Saving entities is done in [NBTChunkSerializer](https://github.com/cuberite/cuberite/blob/7c7f85117bcc5365381cf92c03d97a688195f9a4/src/WorldStorage/NBTChunkSerializer.cpp) look at function `AddMonsterEntity`.
For loading a entity look here [WSSAnvil.cpp](https://github.com/cuberite/cuberite/blob/7c7f85117bcc5365381cf92c03d97a688195f9a4/src/WorldStorage/WSSAnvil.cpp#L1475)

Bond-009:
Just saw the edit, I did some things wrong.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -3039,6 +3055,64 @@ void cWSSAnvil::LoadWolfOwner(cWolf & a_Wolf, const cParsedNBT & a_NBT, int a_Ta
...
+		// There is no owner, bail out:
+		return;
+	}
+
+	// Convert name to UUID, if needed:
+	if (OwnerUUID.empty())
+	{
+		// This wolf has only playername stored (pre-1.7.6), look up the UUID
+		// The lookup is blocking, but we're running in a separate thread, so it's ok
+		OwnerUUID = cRoot::Get()->GetMojangAPI().GetUUIDFromPlayerName(OwnerName);
```
Seadragon91: I don't think this is a good idea. Imagine loading a vanilla world with cuberite that has a dozens of ocelots with owners :)

Bond-009: It shouldn't even be needed if everything is saved correctly.

src/BlockEntities/ChestEntity.cpp
```diff
@@ -217,11 +219,33 @@ void cChestEntity::DestroyWindow()
...
+
 bool cChestEntity::IsBlocked()
 {
-	// TODO: cats are an obstruction
+	cFindSittingCat FindSittingCat;
 	return (
 		(GetPosY() >= cChunkDef::Height - 1) ||
-		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()))
+		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
+		((GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) && !GetWorld()->ForEachEntityInBox(cBoundingBox(Vector3d(GetPosX(), GetPosY() + 1, GetPosZ()), 1, 1), FindSittingCat))
```
madmaxoft: This line is rather long, split it in two, and use an additional level of indent:
```cpp
return (
  cond1 ||
  cond2 ||
  (
    cond3 &&
    cond4
  )
);
```

src/BlockEntities/ChestEntity.cpp
```diff
@@ -217,11 +219,33 @@ void cChestEntity::DestroyWindow()
...
 
+class cFindSittingCat :
+	public cEntityCallback
+{
+	virtual bool Item(cEntity * a_Entity) override
+	{
+		return (
+			(a_Entity->GetEntityType() == cEntity::etMonster) &&
+			(static_cast<cMonster*>(a_Entity)->GetMobType() == eMonsterType::mtOcelot) &&
+			(static_cast<cOcelot*>(a_Entity)->IsSitting())
```
madmaxoft: Spaces in front of `*`

src/Mobs/Ocelot.cpp
```diff
@@ -0,0 +1,194 @@
...
+void cOcelot::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
+{
+	super::Tick(a_Dt, a_Chunk);
+	if (!IsTicking())
+	{
+		// The base class tick destroyed us
+		return;
+	}
+
+	if (!IsTame() && !IsBaby())
```
madmaxoft: This should not be done every tick, the player-finding and path-finding are costly operations.

src/Mobs/Ocelot.cpp
```diff
@@ -0,0 +1,194 @@
...
+	{
+		cPlayer * a_Closest_Player = m_World->FindClosestPlayer(GetPosition(), 10);
```
madmaxoft: This is not a good behavior. If there are two players near to each other, the cat should choose the one with the breeding items, although the player needn't be the closest.

madmaxoft: Also the line of sight should be checked, otherwise cats would move towards underground players :)

src/Mobs/Ocelot.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+
+
+void cOcelot::SpawnOn(cClientHandle & a_ClientHandle)
+{
+	super::SpawnOn(a_ClientHandle);
+	if (GetRandomProvider().RandBool(1.0 / 7.0))
+	{
+		m_World->SpawnMob(GetPosX(), GetPosY(), GetPosZ(), m_MobType, true);
+		m_World->SpawnMob(GetPosX(), GetPosY(), GetPosZ(), m_MobType, true);
```
madmaxoft: Why the same line twice?

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -657,7 +657,18 @@ void cNBTChunkSerializer::AddMonsterEntity(cMonster * a_Monster)
 			}
 			case mtOcelot:
 			{
-				m_Writer.AddInt("Age", reinterpret_cast<const cOcelot *>(a_Monster)->GetAge());
+				const cOcelot *Ocelot = reinterpret_cast<const cOcelot *>(a_Monster);
```
madmaxoft: Use `auto`

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -3039,6 +3055,49 @@ void cWSSAnvil::LoadWolfOwner(cWolf & a_Wolf, const cParsedNBT & a_NBT, int a_Ta
 
 
 
+void cWSSAnvil::LoadOcelotOwner(cOcelot & a_Ocelot, const cParsedNBT & a_NBT, int a_TagIdx)
```
madmaxoft: Instead of duplicating the code, you should change the wolf function to something like 
```cpp
/** Loads the owner name and UUID from the entity at the specified NBT tag.
Returns a pair of {name, uuid}. If the entity is not owned, both are empty strings. */
std::pair<AString, AString> LoadEntityOwner(const cParsedNBT & a_NBT, int a_TagIdx);
```

src/Mobs/Ocelot.cpp
Bond-009: There is a 1/7 chance that 2 baby ocelots spawn with the ocelot.

src/Mobs/Ocelot.cpp
Bond-009: Where/when should it be done?

src/Mobs/Ocelot.cpp
madmaxoft: But that should definitely not be handled in the code that is supposed to send the ocelot spawning information to the client. It should be in the code that has spawned the ocelot in the first place.

Note that your code could lead to an infinite loop - the `cWorld::SpawnMob` actually calls `cEntity::SpawnOn` function

src/Mobs/Ocelot.cpp
madmaxoft: I think once per 23 ticks (prime - undivisible) is good enough.

src/Mobs/Ocelot.cpp
Bond-009: Normally the ocelot should run away from people (I haven't implement that) so if 2 players are standing close to each other it shouldn't really matter because it will run away from the player without the breeding items. (I think)

src/Mobs/Ocelot.cpp
Bond-009: I don't know anywhere else in the code that it would run on every ocelot spawn.
And the infinite loop can be avoided by checking if the current ocelot isn't baby.

src/WorldStorage/WSSAnvil.cpp
Bond-009: Is there a cleaner/better way to do this?
https://github.com/cuberite/cuberite/pull/3829/files#diff-a4136529bdf04bfe56434fa0b68698a0R3027

src/Mobs/Ocelot.cpp
madmaxoft: Well, there aren't too many places in the code where mobs are spawned - either in the world generator (`cFinishGenPassiveMobs`), the mob spawner block entity (`cMobSpawnerEntity`) or the spawn eggs item handler (`cItemSpawnEggHandler`). Those would be much better places.

src/WorldStorage/WSSAnvil.cpp
madmaxoft: Somewhat:
```cpp
return std::pair<AString, AString>();  // Should return a pair of empty strings
return std::make_pair(OwnerName, OwnerUuid);  // Returns a filled pair, no need to specify types
```

src/Mobs/Ocelot.cpp
madmaxoft: Okay, I guess we can keep it like that for now, until proper behaviors are implemented.

src/Mobs/Ocelot.cpp
Bond-009: Yeah, but it should happen in all 3 of these cases and then you have duplicated code.

madmaxoft: I prefer code that is (slightly) duplicated over code that does something completely out of place unexpected.
I'm not even sure if all three of those instances should spawn the extras. I doubt spawning an ocelot from a mob spawner would spawn the kittens in Vanilla.

src/Mobs/Ocelot.h
```diff
@@ -13,17 +14,47 @@ class cOcelot :
...
+	}
+	void SetOcelotType (int a_Type)         { m_Type = a_Type; }
+
+protected:
+
+	bool      m_IsSitting;
+	bool      m_IsTame;
+	bool      m_IsBegging;
+	int       m_Type;
+	int       m_CheckPlayerTickCount;
```
madmaxoft: Use `cMonster::m_Age` instead. (`if (m_Age % 23 == 0)`)

src/Mobs/Ocelot.h
```diff
@@ -13,17 +14,47 @@ class cOcelot :
...
 	}
 
-	CLASS_PROTODEF(cOcelot)
+	// Get functions
+	bool    IsSitting     (void) const override { return m_IsSitting; }
+	bool    IsTame        (void) const override { return m_IsTame; }
+	bool    IsBegging     (void) const { return m_IsBegging; }
+	AString GetOwnerName  (void) const { return m_OwnerName; }
+	AString GetOwnerUUID  (void) const { return m_OwnerUUID; }
+	int     GetOcelotType (void) const {return m_Type; }
```
madmaxoft: Vertical alignment failure

src/Mobs/Ocelot.h
```diff
@@ -13,17 +14,47 @@ class cOcelot :
...
+		m_OwnerUUID = a_NewOwnerUUID;
+	}
+	void SetOcelotType (int a_Type)         { m_Type = a_Type; }
+
+protected:
+
+	bool      m_IsSitting;
+	bool      m_IsTame;
+	bool      m_IsBegging;
+	int       m_Type;
```
madmaxoft: Use an enum instead of an int.

madmaxoft: And provide a better descriptive name. How about `FurColor` or `FurStyle`?

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2468,6 +2468,27 @@ void cWSSAnvil::LoadOcelotFromNBT(cEntityList & a_Entities, const cParsedNBT & a
 		return;
 	}
 
+	std::pair<AString, AString> OwnerInfo = LoadEntityOwner(a_NBT, a_TagIdx);
```

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2848,7 +2869,12 @@ void cWSSAnvil::LoadWolfFromNBT(cEntityList & a_Entities, const cParsedNBT & a_N
 		return;
 	}
 
-	LoadWolfOwner(*Monster.get(), a_NBT, a_TagIdx);
+	std::pair<AString, AString> OwnerInfo = LoadEntityOwner(a_NBT, a_TagIdx);
```
madmaxoft: `auto`

src/Mobs/Ocelot.h
Bond-009: Isn't `m_Age` always 1 if the ocelot is an adult?

madmaxoft: Now that I look in the code, I can't really find any place that would set the age to anything other than `1`. So okay, keep a separate variable for now, but add a short doxycomment to it explaining its purpose.

Seadragon91: @madmaxoft Maybe you mean world age?
`m_World->GetWorldAge()`

madmaxoft: No, I specifically meant the `cMonster::m_Age`, which I expected to be the age of the mob, in ticks. We will still need such a counter in the future, or alternatively the world tick on which the mob was created; this will be used to implement the delayed processing in `Tick` functions, as described in #3822

src/Mobs/Ocelot.h
Bond-009: I changed it to `CatType` to be consisted with the Entity Data.

PR #3831, opened 2017-07-02
Changed Lua plugins to only execute files ending in .lua

lkolbly:
Old code executed all files containing ".lua" in the filename. This restricts it to just files ending in ".lua".

This is useful for e.g. emacs, which will put "foo.lua~" autosave files in the plugin directory (if you're editing a plugin with a "foo.lua" file). Now it won't treat those tilde files as being executable code.

I wasn't sure where to put the constant, I didn't want an unnamed magic constant.


src/Bindings/PluginLua.cpp
```diff
@@ -26,6 +26,7 @@ extern "C"
 #undef TOLUA_TEMPLATE_BIND
 #include "tolua++/include/tolua++.h"
 
+const std::string LUA_FILE_SUFFIX = ".lua";
```
madmaxoft: Move it into the loader function and use `static const AString` as the datatype.

src/Bindings/PluginLua.cpp
```diff
@@ -113,7 +114,7 @@ bool cPluginLua::Load(void)
 	bool HasInfoLua = false;
 	for (AStringVector::const_iterator itr = Files.begin(), end = Files.end(); itr != end; ++itr)
 	{
-		if (itr->rfind(".lua") != AString::npos)
+		if (itr->rfind(LUA_FILE_SUFFIX) == itr->size() - LUA_FILE_SUFFIX.size())
```
madmaxoft: Why not just compare the extension (from the last dot) to the match string? Then you won't need the superfluous static const variable at all.

lkolbly: Good point, implemented with find_last_of and substr.

src/Bindings/PluginLua.cpp
```diff
@@ -113,7 +113,8 @@ bool cPluginLua::Load(void)
 	bool HasInfoLua = false;
 	for (AStringVector::const_iterator itr = Files.begin(), end = Files.end(); itr != end; ++itr)
 	{
-		if (itr->rfind(".lua") != AString::npos)
+		size_t FormatStart = itr->find_last_of('.');
```
madmaxoft: `Format`? I believe `Extension` would be more descriptive.

lkolbly: Yes, it would (and also avoids the confusing verb/noun confusion). Changed.

PR #3832, opened 2017-07-02
cBlockInfo static initialisation

peterbell10:
Changes `cBlockInfo::Get` to use normal static initialisation and fixes the formatting in BlockInfo.cpp


src/BlockInfo.cpp
```diff
@@ -6,934 +6,933 @@
...
+	Info[E_BLOCK_BIG_FLOWER                   ].m_IsSolid = false;
+	Info[E_BLOCK_BROWN_MUSHROOM               ].m_IsSolid = false;
+	Info[E_BLOCK_CARROTS                      ].m_IsSolid = false;
+	Info[E_BLOCK_CHORUS_PLANT                 ].m_IsSolid = false;
+	Info[E_BLOCK_CHORUS_FLOWER                ].m_IsSolid = false;
+	Info[E_BLOCK_COBWEB                       ].m_IsSolid = false;
+	Info[E_BLOCK_CROPS                        ].m_IsSolid = false;
+	Info[E_BLOCK_DANDELION                    ].m_IsSolid = false;
+	Info[E_BLOCK_DEAD_BUSH                    ].m_IsSolid = false;
+	Info[E_BLOCK_DETECTOR_RAIL                ].m_Is
```
worktycho: I think msvc needs a std::move here.

Seadragon91:
I can compile that with VS 2017. On appveyor VS 2012 is used, maybe it's to old.

peterbell10:
Looks like I've got it working now but I have to say, using a more up to date compiler would have made things a lot simpler.

madmaxoft:
AppVeyor is using VS2013, that is the lowest MSVC version that we currently support.

PR #3833, opened 2017-07-06
Lua plugin cColor

lkolbly:
Manually exports cColor Set and Get functions to the plugin API.

Permits this:
```
function setColor(Item, Red, Green, Blue)
   Item.m_ItemColor:Set(Red, Green, Blue)
   r,g,b = Item.m_ItemColor:Get()
   LOG("Set color to "..r.." "..g.." "..b)
end
```

I stuck with Item.m_ItemColor instead of Item:Color() because m_Enchantments is the same way.


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1730,6 +1730,56 @@ end
...
+						{
+							Name = "Green",
+							Type = "number"
+						},
+						{
+							Name = "Blue",
+							Type = "number"
+						},
+					},
+					Returns =
```
Seadragon91: There is no need to add the table `Returns` if nothing is returned.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1730,6 +1730,56 @@ end
...
+					{
+					},
+					Notes = "Sets the color's red, green, and blue components. Values range from 0 to 255."
+				},
+				Get =
+				{
+					Params =
```
Seadragon91: Same here.

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,57 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
 
+static int tolua_cColor_Set(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
+	if (
+		!L.CheckParamNumber(2) ||
+		!L.CheckParamNumber(3) ||
+		!L.CheckParamNumber(4)
```
Seadragon91: Can be replaced with `CheckParamNumber(2, 4)` the functions accepts a range to check.

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,57 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
 
 
 
+static int tolua_cColor_Set(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
```
Seadragon91: Replace `tolua_S:CheckParamSelf("cColor")` this check if the object is a instance of the class and will also print a warning if not.

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,57 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	)
+	{
+		// TODO: Print a warning?
+		return 0;
+	}
+	if (self == nullptr)
+	{
+		LOGWARNING("%s: invalid self (%p)", __FUNCTION__, static_cast<void *>(self));
+		return 0;
+	}
```
madmaxoft: This check will not be needed once you use `L.CheckParamSelf` above.

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,57 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+
+	self->SetRed(static_cast<unsigned char>(lua_tointeger(tolua_S, 2)));
+	self->SetGreen(static_cast<unsigned char>(lua_tointeger(tolua_S, 3)));
+	self->SetBlue(static_cast<unsigned char>(lua_tointeger(tolua_S, 4)));
```
madmaxoft: Consider using this instead:
```cpp
unsigned char r, g, b;
if (!L.GetStackValues(2, r, g, b))
{
  L.ApiParamError("Cannot read the RGB parameters");  // Probably out of datatype range
}
self->Set(r, g, b);  // Add a method to cColor to set all components at once
```

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,57 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
+	if (self == nullptr)
+	{
+		LOGWARNING("%s: invalid self (%p)", __FUNCTION__, static_cast<void *>(self));
+		return 0;
+	}
+
+	L.Push(self->GetRed());
+	L.Push(self->GetGreen());
+	L.Push(self->GetBlue());
```
madmaxoft: `L.Push` can take multiple values:
```cpp
L.Push(self->GetRed(), self->GetGreen(), self->GetBlue());
```

src/Bindings/ManualBindings.cpp
```diff
@@ -3727,6 +3780,11 @@ void cManualBindings::Bind(lua_State * tolua_S)
 			tolua_function(tolua_S, "SendPluginMessage", tolua_cClientHandle_SendPluginMessage);
 		tolua_endmodule(tolua_S);
 
+		tolua_beginmodule(tolua_S, "cColor");
+			tolua_function(tolua_S, "Set", tolua_cColor_Set);
+			tolua_function(tolua_S, "Get", tolua_cColor_Get);
```
madmaxoft: Perhaps add the per-component setter functions, too?

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,57 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	cLuaState L(tolua_S);
+
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
+	if (
+		!L.CheckParamNumber(2) ||
+		!L.CheckParamNumber(3) ||
+		!L.CheckParamNumber(4)
+	)
+	{
+		// TODO: Print a warning?
```
madmaxoft: No warning needed, the `L.CheckXYZ` function already print a detailed warning message.

src/Bindings/ManualBindings.cpp
madmaxoft: You can even combine this with the check immediately below into a single `if`

src/Bindings/ManualBindings.cpp
madmaxoft: Or are those already auto-exported?

Server/Plugins/APIDump/APIDesc.lua
lkolbly: Done.

Server/Plugins/APIDump/APIDesc.lua
lkolbly: Removed.

src/Bindings/ManualBindings.cpp
lkolbly: Ooh, much cleaner.

src/Bindings/ManualBindings.cpp
lkolbly: Added check before the cast.

src/Bindings/ManualBindings.cpp
lkolbly: Done.

src/Bindings/ManualBindings.cpp
lkolbly: cColor already had a SetColor method, so didn't need to add it.

src/Bindings/ManualBindings.cpp
lkolbly: cColor isn't already exported - tolua has problems parsing it.

lkolbly: I didn't do per-component getter/setter functions because it'd be a lot of duplicated code and I couldn't think of a use case (maybe a plugin that had R/G/B sliders? But even if someone found a usecase, they could write a short workaround function).

```
function SetRed(Item, R)
   local _, G, B = Item.m_ItemColor:Get()
   Item.m_ItemColor:Set(R, G, B)
end
```

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,49 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+
+
+
+
+
+static int tolua_cColor_Get(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
```
Seadragon91: Please use `CheckParamSelf` here, too. Also remove the nullptr check then

lkolbly: Derp. Done.

Seadragon91: Thanks :)

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,48 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
 
 
 
+static int tolua_cColor_Set(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+
+	if (!L.CheckParamNumber(2, 4) || !L.CheckParamSelf("cColor"))
```
madmaxoft: For consistency, we check params in the order they come in.

src/Bindings/ManualBindings.cpp
```diff
@@ -3684,6 +3684,48 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	{
+		return 0;
+	}
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
```
madmaxoft: Should be possible to use `L.GetStackValues` on `self`, too (combine with below).

src/Bindings/ManualBindings.cpp
```diff
@@ -3693,10 +3735,12 @@ void cManualBindings::Bind(lua_State * tolua_S)
 		tolua_usertype(tolua_S, "cLineBlockTracer");
 		tolua_usertype(tolua_S, "cStringCompression");
 		tolua_usertype(tolua_S, "cUrlParser");
+		tolua_usertype(tolua_S, "cColor");
```
madmaxoft: Please alpha-sort

src/Bindings/ManualBindings.cpp
```diff
@@ -3693,10 +3735,12 @@ void cManualBindings::Bind(lua_State * tolua_S)
...
 		tolua_cclass(tolua_S, "cCryptoHash",        "cCryptoHash",        "", nullptr);
 		tolua_cclass(tolua_S, "cLineBlockTracer",   "cLineBlockTracer",   "", nullptr);
 		tolua_cclass(tolua_S, "cStringCompression", "cStringCompression", "", nullptr);
 		tolua_cclass(tolua_S, "cUrlParser",         "cUrlParser",         "", nullptr);
+		tolua_cclass(tolua_S, "cColor",             "cColor",             "", nullptr);
```
madmaxoft: alpha-sort

src/Bindings/ManualBindings.cpp
madmaxoft: Okay, I guess it's not much of an issue.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1730,6 +1730,50 @@ end
...
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+						{
+							Type = "number",
+						},
+						{
+							Type = "number",
```
madmaxoft: Add the `Name` to these, too.

src/Bindings/ManualBindings.cpp
lkolbly: I was having a bit of trouble with that:

```
cColor * self;
unsigned char r, g, b;
if (!L.GetStackValues(1, self, r, g, b)) {
   ...
}
```

Throws a giant error message of the form:

```
/home/lane/minecraft/cuberite/src/Bindings/LuaState.h:663:7: error: no type named ‘type’ in ‘struct std::enable_if<false, void>’
/home/lane/minecraft/cuberite/src/Bindings/LuaState.h: In instantiation of ‘bool cLuaState::GetStackValues(int, Arg1&&, Args&& ...) [with Arg1 = cColor*&; Args = {unsigned char&, unsigned char&, unsigned char&}]’:
/home/lane/minecraft/cuberite/src/Bindings/ManualBindings.cpp:3698:40:   required from here
/home/lane/minecraft/cuberite/src/Bindings/LuaState.h:685:7: note: candidate: template<class T> bool cLuaState::GetStackValue(int, cLuaState::cOptionalParam<T>&&)
  bool GetStackValue(int a_StackPos, cOptionalParam<T> && a_ReturnedVal)
       ^~~~~~~~~~~~~
/home/lane/minecraft/cuberite/src/Bindings/LuaState.h:685:7: note:   template argument deduction/substitution failed:
In file included from /home/lane/minecraft/cuberite/src/Bindings/ManualBindings.h:12:0,
                 from /home/lane/minecraft/cuberite/src/Bindings/ManualBindings.cpp:4:
/home/lane/minecraft/cuberite/src/Bindings/LuaState.h:757:7: note:   mismatched types ‘cLuaState::cOptionalParam<T>’ and ‘cColor*’
   if (!GetStackValue(a_StartStackPos, std::forward<Arg1>(a_Arg1)))
        
In file included from /home/lane/minecraft/cuberite/src/Bindings/LuaState.h:733:0,
                 from /home/lane/minecraft/cuberite/src/Bindings/ManualBindings.h:12,
                 from /home/lane/minecraft/cuberite/src/Bindings/ManualBindings.cpp:4:
/home/lane/minecraft/cuberite/src/Bindings/LuaState_Declaration.inc:94:6: note: candidate: bool cLuaState::GetStackValue(int, cArrowEntity*&)
 bool GetStackValue(int a_StackPos, PtrcArrowEntity & a_ReturnedVal);
      ^~~~~~~~~~~~~
/home/lane/minecraft/cuberite/src/Bindings/LuaState_Declaration.inc:94:6: note:   no known conversion for argument 2 from ‘cColor*’ to ‘cArrowEntity*&’
/home/lane/minecraft/cuberite/src/Bindings/LuaState_Declaration.inc:95:6: note: candidate: bool cLuaState::GetStackValue(int, const cArrowEntity*&)
 bool GetStackValue(int a_StackPos, ConstPtrcArrowEntity & a_ReturnedVal);
      ^~~~~~~~~~~~~
/home/lane/minecraft/cuberite/src/Bindings/LuaState_Declaration.inc:95:6: note:   no known conversion for argument 2 from ‘cColor*’ to ‘const cArrowEntity*&’
/home/lane/minecraft/cuberite/src/Bindings/LuaState_Declaration.inc:96:6: note: candidate: bool cLuaState::GetStackValue(int, cBeaconEntity*&)
 bool GetStackValue(int a_StackPos, PtrcBeaconEntity & a_ReturnedVal);
      ^~~~~~~~~~~~~
```
Which continues for several hundred different types. I suspect this has to do with the fact that cColor isn't exported by tolua? (since it seems to be matching it against every class that is exported by tolua)

Other than this, I've alpha sorted everything.

madmaxoft: Your suspicion is correct. So why not add the `cColor` to ToLua? It would then gain all the partial getters and setters, you wouldn't need to export `SetColor` manually, and `GetColor` could use `GetStackValues` for a `cColor *` value.

lkolbly: I had tried that earlier without success, but this time I managed to make it work by removing the type information from the enum (it's already being statically cast in several places anyway). Regardless, just pushed that.

lkolbly: And we don't have to touch ManualBindings now that it's exported through tolua.

madmaxoft: It should work even with the typed enum now, there has been a commit recently that enabled support for it in ToLua. (Don't forget to `git submodule update` in your local repo)

lkolbly: Doesn't seem to work for me:
```
***curr code for error is  unsigned int
        {
                COLOR_MIN = 0,
                COLOR_MAX = 255,
                COLOR_LIMIT = 256,
                COLOR_NONE = 0xFFFFFFFF,
        };
...
stack traceback:
        [string "../../lib/tolua++/src/bin/lua/basic.lua"]:94: in function 'tolua_error'
        ../../lib/tolua++/src/bin/lua/all.lua:30: in main chunk
        [C]: in function 'dofile'
        BindingsProcessor.lua:55: in function 'invokeToLua'
        BindingsProcessor.lua:1026: in main chunk
        [C]: in ?

** tolua: parse error
stack traceback:
        [C]: in function 'error'
        [string "../../lib/tolua++/src/bin/lua/container.lua"]:787: in function 'doparse'
        [string "../../lib/tolua++/src/bin/lua/container.lua"]:799: in function 'parse'
        [string "../../lib/tolua++/src/bin/lua/class.lua"]:199: in function 'Class'
        [string "../../lib/tolua++/src/bin/lua/container.lua"]:720: in function 'doparse'
        [string "../../lib/tolua++/src/bin/lua/container.lua"]:799: in function 'parse'
        [string "../../lib/tolua++/src/bin/lua/package.lua"]:309: in function 'Package'
        [string "../../lib/tolua++/src/bin/lua/doit.lua"]:55: in function <[string "../../lib/tolua++/src/bin/lua/doit.lua"]:27>
        [C]: in function 'xpcall'
        ../../lib/tolua++/src/bin/lua/all.lua:26: in main chunk
        [C]: in function 'dofile'
        BindingsProcessor.lua:55: in function 'invokeToLua'
        BindingsProcessor.lua:1026: in main chunk
        [C]: in ?.

Code being processed:
unsigned int
```
According to lib/tolua++/include/tolua++.h, it's on version 1.0.92 (I did a submodule update).

madmaxoft: It would still be nice to have the `Get` method that returns all three components; ToLua cannot export such a function, so ManualBindings for it should be done.

Alright, keep the enum untyped for now. @peterbell10 would know more about those typed enums.

src/Bindings/ManualBindings.cpp
```diff
@@ -3297,6 +3297,24 @@ static int tolua_cChunkDesc_GetBlockTypeMeta(lua_State * a_LuaState)
...
 
+static int tolua_cColor_GetColor(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+
+	if (!L.CheckParamSelf("cColor"))
+	{
+		return 0;
+	}
+	cColor * self = reinterpret_cast<cColor *>(tolua_tousertype(tolua_S, 1, nullptr));
```
madmaxoft: Now you should be able to use `L.GetStackValue(1, self);`

src/Bindings/ManualBindings.cpp
lkolbly: Fair 'nuff, just added it back (renamed to GetColor, to match SetColor).

peterbell10: It's not failing because it's typed, it's failing because it's anonymous.  Just give it a name like `eColorConstants` and add `// tolua_export` to the closing curly for the class and it should work.

src/Bindings/ManualBindings.cpp
lkolbly: Right, that was the point of this. :P

Just changed it, it compiles (with gcc, testing with clang now).

src/Bindings/ManualBindings.cpp
lkolbly: Yup, that compiled (named it eColorLimits). Just now pushed.

lkolbly:
Finally added CheckLuaStyle as a pre-commit hook to my repo, which I probably should have done a long time ago.

Anyway, now it passes automated CI.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1730,6 +1730,178 @@ end
 				},
 			},
 		},
+		cColor =
+		{
+			Desc = [[
+				Encapsulates a RGB color, e.g. for armor.
```
madmaxoft: I think the range of the RGB components should be mentioned somewhere - is it 0 - 255 each, or 0.0 - 1.0 each?

src/Color.h
```diff
@@ -51,8 +51,8 @@ class cColor
 
 	/** Resets the color */
 	void Clear() { m_Color = static_cast<unsigned int>(COLOR_NONE); }
-	// tolua_end
```
madmaxoft: This makes the `m_Color` member exported, probably not what we want. Keep this and remove the one after the class declaration.

Note: `// tolua_begin` and `// tolua_end` mark blocks that are parsed by ToLua, `// tolua_export` marks a single line that is parsed. ToLua basically first removes everything outside of these, and then tries to parse it as a regular C++ code.

Server/Plugins/APIDump/APIDesc.lua
lkolbly: 0-255 each.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1730,6 +1730,178 @@ end
...
+			{
+				Clear =
+				{
+					Notes = "Resets the color to uninitialized."
+				},
+				constructor =
+				{
+					{
+						Returns = { {Type="cColor"} },
+						Notes = "Creates an uninitialized cColor. Each component must be between 0 and 255, inclusive.",
```
madmaxoft: `uninitialized`? Since it accepts the RGB components, I'd say it's perfectly initialized.

PR #3834, opened 2017-07-06
If entity is a player, send relmove packets.

### LINKED ISSUE #3790: Can't see other player running/jumping/something like this
Client version:   1.1.2
Server OS:        Linux Ubuntu 16.04

I can't see other player. They stay only on the spawning place. When I beat them, than I can see they running or crafting. After a reconnect the same problem comes up.

lkolbly:
I don't know whether this is the correct fix, but this appears to fix #3790 without reverting #3739.

My understanding is that, for players, m_Speed is not set as the player moves around, thus causing BroadcastMovementUpdate to not send the relmove packets for players.

This PR forces BroadcastMovementUpdate to send relmove packets if the entity is a player.

Another possibility would be to just set m_Speed correctly on the player entity, but the client doesn't seem to send us that info, so I don't know whether that's right (we'd have to estimate it, and then when we set it we wouldn't want to send it back to the client and overwrite their speed).

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1920,7 +1920,7 @@ void cEntity::BroadcastMovementUpdate(const cClientHandle * a_Exclude)
 		}
 
 		// Only send movement if speed is not 0 and 'no speed' was sent to client
-		if (!m_bHasSentNoSpeed)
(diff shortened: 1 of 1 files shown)


Seadragon91:
Tested with two clients. With this fix I see now the player moving and jumping. Thanks for the fix.

PR #3838, opened 2017-07-09
Use unique_ptr for array ownership

peterbell10:
Replaces all the cases I could find of raw pointers owning arrays with `unique_ptr`.


peterbell10:
Oh well, I guess clang has doomed this.

PR #3840, opened 2017-07-09
Added sound effect when horse rejects a player trying to tame it

spekdrum:
Added sound effect when horse rejects a player trying to tame it. 
It's the same sfx when horse rearing but with higher pitch.


spekdrum:
Oops I miss origin repo

PR #3843, opened 2017-07-10
Changing pickup lifetime and combining semantics, making these adjustable in the lua api.

lkolbly:
* Made pickup lifetimes configurable from the lua API.
* Made whether or not a pickup is allowed to be combined with other pickups configurable from the API.
* When a pickup combines with another one, the age is reset to zero ([like in vanilla](http://minecraft.gamepedia.com/Item_(entity))).
* Added cWorld:CreatePickup so that plugins can create a pickup directly, and retrieve the Entity ID for it (for future handling by the plugin).

Useful for e.g. an item generator, so that there is a cap on how many items are uncollected in the generator at one time:
```
local PickupId = World:CreatePickup(generator.x + 0.5, generator.y + 2.0, generator.z + 0.5, cItem(drop.type, 1), 0, 0, 0)
World:DoWithEntityByID(PickupId, function (Pickup)
			      Pickup:SetLifetime(200) -- 10 seconds
			      Pickup:SetCanCombine(false)
			      LOG("Pickup will last for "..Pickup:GetLifetime().." ticks")
end)
```
Or, more concisely:
```
local PickupId = World:CreatePickup(generator.x + 0.5, generator.y + 2.0, generator.z + 0.5, cItem(drop.type, 1), 0, 0, 0, 200, false)
```
(the cap is derived from putting this in a loop, so you run this e.g. every 2 seconds to get a cap of 5.)

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -9260,6 +9260,16 @@ a_Player:OpenWindow(Window);
 			]],
 			Functions =
 			{
+				CanCombine =
+				{
+					Returns =
+					{
+						{
+							Type = "boolean"
+						}
+					},
+					Notes = "Returns whether this pickup is allowed to combine with other similar pickups.",
+				},
 				CollectedBy =
 				{
 					Params =
@@ -9298,6 +9308,16 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Returns the item represented by this pickup",
 				},
+				GetLifetime =
+				{
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+					},
+					Notes = "Returns the total length of this pickup's lifespan, in ticks.",
+				},
 				IsCollected =
 				{
 					Returns =
@@ -9328,6 +9348,28 @@ a_Player:OpenWindow(Window);
 						},
 					},
 					Notes = "Sets the pickup's age, in ticks.",
+				},
+				SetCanCombine =
+				{
+					Params =
+					{
+						{
+							Name = "CanCombine",
+							Type = "boolean",
+						},
+					},
+					Notes = "Sets whether this pickup is allowed to combine with other similar pickups.",
+				},
+				SetLifetime =
+				{
+					Params =
+					{
+						{
+							Name = "SetLifetime",
+							Type = "number",
+						},
+					},
+					Notes = "Sets the total lifespan of this pickup before it despawns, in ticks.",
 				},
 			},
 			Inherits = "cEntity",
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -9365,11 +9365,11 @@ a_Player:OpenWindow(Window);
 					Params =
 					{
 						{
-							Name = "SetLifetime",
+							Name = "LifeTimeInTicks",
 							Type = "number",
 						},
 					},
-					Notes = "Sets the total lifespan of this pickup before it despawns, in ticks.",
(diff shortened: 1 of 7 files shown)


madmaxoft:
I'm not too happy with the `SpawnItemPickups` / `CreatePickup` confusion this creates. The methods should be named similarly, after all they basically differ only in one parameter. It could even be a single overloaded method.

lkolbly:
Maybe "SpawnItemPickup" (singular)? I think making it an overloaded method could be problematic, since they have different semantics (SpawnItemPickups spawns enough pickups to put the specified items on the map, CreatePickup directly creates a single pickup with the given item, and therefore can return the EntityID).

madmaxoft:
Okay, singular non-overloaded sounds good.

src/Entities/Pickup.cpp
```diff
@@ -92,12 +96,14 @@ class cPickupCombiningCallback :
 
 
 
-cPickup::cPickup(double a_PosX, double a_PosY, double a_PosZ, const cItem & a_Item, bool IsPlayerCreated, float a_SpeedX /* = 0.f */, float a_SpeedY /* = 0.f */, float a_SpeedZ /* = 0.f */)
+cPickup::cPickup(double a_PosX, double a_PosY, double a_PosZ, const cItem & a_Item, bool IsPlayerCreated, float a_SpeedX /* = 0.f */, float a_SpeedY /* = 0.f */, float a_SpeedZ /* = 0.f */, int a_LifetimeTicks /* = 6000 = 5 minutes */, bool a_CanCombine /* = true */)
```
madmaxoft: I think having the defaults here, although commented, is not practical. They serve no purpose here and can change.

lkolbly: Okay. Should I get rid of the SpeedX default comments as well?

madmaxoft: Yes please

lkolbly: Done.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9329,6 +9349,28 @@ a_Player:OpenWindow(Window);
...
+						},
+					},
+					Notes = "Sets whether this pickup is allowed to combine with other similar pickups.",
+				},
+				SetLifetime =
+				{
+					Params =
+					{
+						{
+							Name = "SetLifetime",
```
madmaxoft: `LifeTimeInTicks` would be a better name

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9329,6 +9349,28 @@ a_Player:OpenWindow(Window);
...
+							Type = "number",
+						},
+					},
+					Notes = "Sets the total lifespan of this pickup before it despawns, in ticks.",
```
madmaxoft: Does it reset the timer? (so that the  lifespan specified is measured from now?)

lkolbly: This function does not. It might be more intuitive if it did, though?

madmaxoft: Not sure. Is there a way to query the current age?

lkolbly: Yeah, GetAge() returns the age in ticks.

lkolbly: So if you really want to, you can set "how much longer the pickup will live", but it requires math.

madmaxoft: Okay, math is okay, as long as there is a way. Just add a note to this function's docs with this info.

lkolbly: Or instead of math you can do SetAge(0).

Documented.

PR #3846, opened 2017-07-12
FastRandom: Remove discrepancy between arg and return value type.

madmaxoft:
MSVC would generate a warning due to this. And after all, if you want a random number in certain range, you better make sure all of that range fits the return type.

### DIFF
--- a/src/FastRandom.h
+++ b/src/FastRandom.h
@@ -97,8 +97,8 @@ public:
 
 
 	/** Return a random IntType in the range [0, a_Max]. */
-	template <class IntType = int, class ArgType>
-	IntType RandInt(ArgType a_Max)
+	template <class IntType = int>
+	IntType RandInt(IntType a_Max)
 	{
 		ASSERT((a_Max >= 0) && (a_Max <= std::numeric_limits<IntType>::max()));
 		Detail::cUniform<IntType> dist(IntType(0), static_cast<IntType>(a_Max));
--- a/src/FastRandom.h
+++ b/src/FastRandom.h
@@ -77,8 +77,8 @@ public:
 
 
 	/** Return a random IntType in the range [a_Min, a_Max]. */
-	template <class IntType = int, class ArgType>
-	IntType RandInt(ArgType a_Min, ArgType a_Max)
+	template <class IntType = int>
+	IntType RandInt(IntType a_Min, IntType a_Max)
 	{
 		ASSERT(
 			(a_Max >= a_Min) &&
@@ -122,8 +122,8 @@ public:
 
 
 	/** Return a random RealType in the range [a_Min, a_Max). */
-	template <class RealType = float, class ArgType>
-	RealType RandReal(ArgType a_Min, ArgType a_Max)
+	template <class RealType = float>
+	RealType RandReal(RealType a_Min, RealType a_Max)
 	{
 		std::uniform_real_distribution<RealType> dist(a_Min, a_Max);
 		return dist(m_Engine);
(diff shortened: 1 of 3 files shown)


madmaxoft:
@peterbell10 does this make sense to you? :)

peterbell10:
It makes sense without it. I had it that way originally to force `IntType` as an explicit template parameter but I suppose it's not really necessary.
It just means that for something like:
```
char a, b;
auto c = RandInt(a, b);
```
`c` is now a `char` instead of an `int`.

madmaxoft:
Hm, but clang doesn't seem to like it :(

madmaxoft:
After all, why make it more complicated than it needs to be? :)

Ready for review.

src/FastRandom.h
```diff
@@ -97,8 +97,8 @@ class cRandomWrapper
 
 
 	/** Return a random IntType in the range [0, a_Max]. */
-	template <class IntType = int, class ArgType>
-	IntType RandInt(ArgType a_Max)
+	template <class IntType = int>
+	IntType RandInt(IntType a_Max)
```
peterbell10: Shouldn't the `RandInt(a_Min, a_Max)` overload be changed as well as the `RandReal` versions or they will have different behaviour.

src/MobSpawner.cpp
```diff
@@ -107,14 +107,12 @@ eMonsterType cMobSpawner::ChooseMobType(EMCSBiome a_Biome)
...
 	size_t allowedMobsSize = allowedMobs.size();
 	if (allowedMobsSize > 0)
 	{
-		std::set<eMonsterType>::iterator itr = allowedMobs.begin();
-
-		using DiffType = decltype(itr)::difference_type;
-		std::advance(itr, GetRandomProvider().RandInt<DiffType>(allowedMobsSize - 1));
-
+		auto itr = allowedMobs.begin();
+		std::advance(itr, GetRandomProvider().RandInt(allowedMobsSize - 1));
```
peterbell10: I think this has to be `difference_type` , otherwise it's an [error on macOS](https://builds.cuberite.org/job/Cuberite%20OSX%20x64%20Master/213/console).

madmaxoft: Right, and using `difference_type` in MSVC generates an Intellisense parse error and compile-time warning. Stupid `std::advance`. Might as well rewrite it using a simple loop with none of the problems.

madmaxoft:
This is now more of a MobSpawner rewrite than a FastRandom rewrite, but still, it should fix all the issues.

madmaxoft:
Anyone opposed to a merge?

tigerw:
Not so much as an objection, but could you use `std::is_integral` and `std::is_floaty_punt` to enforce types? Do you even need to?

(Also, why use `class RealType` when you can use `typename RealType` :P)

tigerw:
At which point you could perhaps either use static_assert, or even fancier, enable_if and consolidate the Rand*Type* distinctions into one template'd ball of magic.

tigerw:
Also in e.g. https://github.com/cuberite/cuberite/blob/88aa6b97fd2ec18d8c984ada17c1d9205d1cd696/src/FastRandom.h#L140 I cannot quite figure out if that is a C-style cast or a function call. Do we permit the former?

madmaxoft:
I'm trying to make the minimal fix, not rewrite everything from scratch. Yes, `typename` would have been better in my opinion. And the C-style-cast-like call could most probably be omitted completely. But... not the point of this PR. Make your own. With blackjack. And hookers :D

tigerw:
In any case I am not opposed. However, since spekdrum reacted with a thumbs up to your "is anyone opposed" question, can we surmise that he, in fact, is opposed?

spekdrum:
Oops sorry I'm not opposed. I ment to thumbs up the"merge" part :)

PR #3850, opened 2017-07-13
Check for intersection between placed blocks and entities.

### LINKED ISSUE #3196: Blocks can be placed inside yourself
Client version:   1.9.4
Commit id:        f73de02684ce2c5f9fb2e8e36cfe65dc4ac91dea
### Expected behavior

When you right click below you holding a solid block, nothing will happen
### Actual behavior

When you right click below you holding a solid block, a block will be placed, causing you to be pushed out of the way

Hold a solid block, face down, and right click

lkolbly:
Fixes #3196.

Creates a DoesIntersectBlock class inside cPlayer::PlaceBlocks, which detects if an entity intersects any of the blocks being placed. There's a small amount of leeway given (epsilon), so that bridging still works (where the player is placing blocks exactly at the end of their feet).

Computes the total AABB of the blocks being placed, and iterates through every entity in that range.

### DIFF
--- a/src/Entities/Pickup.h
+++ b/src/Entities/Pickup.h
@@ -36,6 +36,8 @@ public:
 
 	virtual void Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
 
+	bool DoesPreventBlockPlacement(void) const override { return false; }
+
 	/** Returns whether this pickup is allowed to combine with other similar pickups */
 	bool CanCombine(void) const { return m_bCanCombine; }  // tolua_export
 
--- a/src/Entities/Pickup.h
+++ b/src/Entities/Pickup.h
@@ -36,7 +36,7 @@ public:
 
 	virtual void Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
 
-	bool DoesPreventBlockPlacement(void) const override { return false; }
+	virtual bool DoesPreventBlockPlacement(void) const override { return false; }
 
 	/** Returns whether this pickup is allowed to combine with other similar pickups */
 	bool CanCombine(void) const { return m_bCanCombine; }  // tolua_export
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -2683,14 +2683,12 @@ bool cPlayer::DoesPlacingBlocksIntersectEntity(const sSetBlockVector & a_Blocks)
 	{
 	public:
 		const std::vector<cBoundingBox> & m_BoundingBoxes;
-		cWorld * m_World;
 
 		// The distance inside the block the entity can still be.
 		const double EPSILON = 0.0005;
 
-		DoesIntersectBlock(const std::vector<cBoundingBox> & a_BoundingBoxes, cWorld * a_World) :
-			m_BoundingBoxes(a_BoundingBoxes),
-			m_World(a_World)
+		DoesIntersectBlock(const std::vector<cBoundingBox> & a_BoundingBoxes) :
+			m_BoundingBoxes(a_BoundingBoxes)
 		{
 		}
 
@@ -2713,7 +2711,7 @@ bool cPlayer::DoesPlacingBlocksIntersectEntity(const sSetBlockVector & a_Blocks)
 			}
 			return false;
 		}
-	} Callback(PlacementBoxes, World);
+	} Callback(PlacementBoxes);
 
 	// See if any entities in that bounding box collide with anyone
 	if (!World->ForEachEntityInBox(PlacingBounds, Callback))
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -2676,6 +2676,8 @@ bool cPlayer::DoesPlacingBlocksIntersectEntity(const sSetBlockVector & a_Blocks)
 		}
 	}
 
+	cWorld * World = GetWorld();
+
 	// Check to see if any entity intersects any block being placed
 	class DoesIntersectBlock : public cEntityCallback
 	{
@@ -2707,10 +2709,10 @@ bool cPlayer::DoesPlacingBlocksIntersectEntity(const sSetBlockVector & a_Blocks)
 			}
 			return false;
 		}
-	} Callback(PlacementBoxes, m_World);
+	} Callback(PlacementBoxes, World);
 
 	// See if any entities in that bounding box collide with anyone
-	if (!m_World->ForEachEntityInBox(PlacingBounds, Callback))
+	if (!World->ForEachEntityInBox(PlacingBounds, Callback))
 	{
 		return true;
 	}
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -2669,13 +2669,13 @@ bool cPlayer::PlaceBlocks(const sSetBlockVector & a_Blocks)
 				int y = blk.GetY();
 				int z = blk.GetZ();
 				cBoundingBox BlockBox = BlockHandler->GetPlacementCollisionBox(
-				        m_World->GetBlock(x - 1, y, z),
+					m_World->GetBlock(x - 1, y, z),
 					m_World->GetBlock(x + 1, y, z),
 					m_World->GetBlock(x, y - 1, z),
 					m_World->GetBlock(x, y + 1, z),
 					m_World->GetBlock(x, y, z - 1),
 					m_World->GetBlock(x, y, z + 1)
-			        );
+				);
 				BlockBox.Move(x, y, z);
 
 				// Put in a little bit of wiggle room
@@ -2705,13 +2705,15 @@ bool cPlayer::PlaceBlocks(const sSetBlockVector & a_Blocks)
 			m_World->GetBlock(x, y + 1, z),
 			m_World->GetBlock(x, y, z - 1),
 			m_World->GetBlock(x, y, z + 1)
-	        );
+		);
 		BlockBox.Move(x, y, z);
 
 		if (HasInitializedBounds)
 		{
 			PlacingBounds = PlacingBounds.Union(BlockBox);
-		} else {
+		}
+		else
+		{
 			PlacingBounds = BlockBox;
 			HasInitializedBounds = true;
 		}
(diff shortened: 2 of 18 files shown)


src/Entities/Player.cpp
```diff
@@ -2643,6 +2643,64 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+	// Compute the bounding box of the blocks to be placed
+	double MaxDouble = std::numeric_limits<double>::max();
+	double MinDouble = std::numeric_limits<double>::min();
+	Vector3d BlocksMin(MaxDouble, MaxDouble, MaxDouble);
+	Vector3d BlocksMax(MinDouble, MinDouble, MinDouble);
+	for (auto blk: a_Blocks)
+	{
+		BlocksMin = BlocksMin.Min(blk.GetPos());
+		BlocksMax = BlocksMax.Max(blk.GetPos() + Vector3d(1, 1, 1));
+	}
```
madmaxoft: This will most likely fail for some block that don't occupy the full voxel, or occupy space out of the current voxel (fence?).

src/Vector3.h
```diff
@@ -184,6 +184,26 @@ class Vector3
...
+
+	/** Returns a new vector with each coordinate set to the max */
+	inline Vector3<T> Max(const Vector3<T> & a_Other) const
+	{
+		return Vector3<T>(
+			std::max(x, a_Other.x),
+			std::max(y, a_Other.y),
+			std::max(z, a_Other.z)
+		);
+	}
```
madmaxoft: These functions seem like an unnecessary kludge. I think a better solution would be to add an `Engulf` function to `cBoundingBox`, similar to `cCuboid::Engulf`, and use that instead.

src/Entities/Player.cpp
lkolbly: Hadn't thought of that. Do blocks have a GetBoundingBox() function?

madmaxoft: Not that I'm aware of. If you do decide to add one, you should name it a `CollisionBox`, there could be other bounding boxes, too.

madmaxoft: Some blocks' collision boxes may depend on the surroundings. Not sure what the getter function should receive as params.

lkolbly: I can't think of any blocks that depend on their surroundings? I think all blocks collision boxes can be determined from the type and the block meta (I'm a little fuzzy on how all that works, but fences store the direction they go in as part of the block).

And the collision box may not even be a cube, in the case of stairs.

madmaxoft: Fences don't store their direction anywhere. And door blocks need both blocks to determine the exact door position.

I believe stairs are represented by a half-step collision box in Vanilla; although I'm not sure about this.

lkolbly: Stairs are half full-height and half half-height (a cube with an edge cut out).

Maybe the parameter could be the 6 cardinal neighbors (+/- X,Y,Z)? I don't think any block cares about blocks further away than that (except tripwire hooks, now that I research it).

At least with tripwire hooks, we could place it in a folded state, and when it immediately extends it just pushes the player. I'll have to load up a vanilla world later and see how that works.

For doors, both blocks are being placed at the same time, so asking "what's the bounding box if we place this block here" won't work, assuming we test one block at a time.

Which hints at another way to solve the problem: What if instead of "collision box" we had a different "placement box", which is fixed for a block type (or, block type and direction, to handle doors), and doesn't require neighbor information.

In vanilla, you can't place stairs if you're inside the full bounding cube (place a slab, sneak to the edge, place stairs on the ground next to the slab).

madmaxoft: Sounds great :)

lkolbly: Okay, I'll go try that out.

src/Vector3.h
lkolbly: cBoundingBox already has a Union() method, it turns out (it had a bug, but this PR fixes that as well).

lkolbly:
Just pushed, now it implements a GetPlacementCollisionBox in cBlockHandler which can be overidden. This method is passed the block types of the six cardinal neighbors of the blocks being passed.

There's a bit of code duplication in Player.cpp, with regards to getting the placement block. It may make sense to factor that into a function.

lkolbly:
...and just committed the new BlockFence.h file.

src/Entities/Player.cpp
```diff
@@ -2643,6 +2643,93 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+	public:
+		const sSetBlockVector & m_Blocks;
+		cWorld * m_World;
+
+		// The distance inside the block the entity can still be.
+		const double EPSILON = 0.05;
+
+		DoesIntersectBlock(const sSetBlockVector & a_Blocks, cWorld * a_World) :
+			m_Blocks(a_Blocks)
+			, m_World(a_World)
```
madmaxoft: Style: The comma should be at the end of the previous line

src/Entities/Player.cpp
```diff
@@ -2643,6 +2643,93 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+	if (!m_World->ForEachEntityInBox(PlacingBounds, Callback))
+	{
+		// Abort - re-send all the current blocks in the a_Blocks' coords to the client:
+		for (auto blk2: a_Blocks)
+		{
+			m_World->SendBlockTo(blk2.GetX(), blk2.GetY(), blk2.GetZ(), this);
+		}
+		return false;
+	}
+
```
madmaxoft: This whole block is already rather large, it would be better to move it into a separate function.

src/Entities/Player.cpp
```diff
@@ -2643,6 +2643,93 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+		int y = blk.GetY();
+		int z = blk.GetZ();
+		cBoundingBox BlockBox = BlockHandler->GetPlacementCollisionBox(
+			m_World->GetBlock(x - 1, y, z),
+			m_World->GetBlock(x + 1, y, z),
+			m_World->GetBlock(x, y - 1, z),
+			m_World->GetBlock(x, y + 1, z),
+			m_World->GetBlock(x, y, z - 1),
+			m_World->GetBlock(x, y, z + 1)
+		);
```
madmaxoft: This looks the same as the code in the entity callback. I think it'd be better to remember the bounding boxes here and pass them to the entity callback as a `std::vector<cBoundingBox>`.

src/Entities/Player.cpp
lkolbly: I moved the whole block into cPlayer::DoesPlacingBlocksIntersectEntity. I wanted to also move the 10-line snippet of getting the placement box for a hypothetical block into a function, but couldn't find a good place for it (putting it with the cBlockInfo pulls in a cWorld dependency that I don't think belongs, putting it into cWorld seems like it doesn't really fit either). The whole function comes out to 75 lines, which is big but manageable.

lkolbly:
Just now removed the min/max functions from vector, as well.

src/Entities/Player.cpp
```diff
@@ -2641,8 +2641,101 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+	{
+		cBlockHandler * BlockHandler = cBlockInfo::GetHandler(blk.m_BlockType);
+		int x = blk.GetX();
+		int y = blk.GetY();
+		int z = blk.GetZ();
+		cBoundingBox BlockBox = BlockHandler->GetPlacementCollisionBox(
+			m_World->GetBlock(x - 1, y, z),
+			m_World->GetBlock(x + 1, y, z),
+			m_World->GetBlock(x, y - 1, z),
+			m_World->GetBlock(x, y + 1, z),
```
madmaxoft: The two lines above will fail for blocks at the bottom or top of the world (Y == 0, Y == cChunkDef::Height - 1)

src/Entities/Player.cpp
```diff
@@ -2641,8 +2641,101 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+		{
+			cBoundingBox EntBox(a_Entity->GetPosition(), a_Entity->GetWidth() / 2, a_Entity->GetHeight());
+			for (auto BlockBox: m_BoundingBoxes)
+			{
+
+				// Put in a little bit of wiggle room
+				BlockBox.Expand(-EPSILON, -EPSILON, -EPSILON);
+				if (EntBox.DoesIntersect(BlockBox))
+				{
+					return true;
```
madmaxoft: Does this mean that pickups prevent the placement as well? I suppose there are entities that shouldn't prevent placement.

src/Entities/Player.cpp
lkolbly: Added inline checks, assuming that any block that is placed at Y=0 or Y=height-1 treats the void as air (at the moment, no blocks actually check above or below).

src/Entities/Player.cpp
lkolbly: Hm, good point, added a cEntity::DoesPreventBlockPlacement method.

src/Entities/Pickup.h
```diff
@@ -36,6 +36,8 @@ class cPickup :
 
 	virtual void Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk) override;
 
+	bool DoesPreventBlockPlacement(void) const override { return false; }
```
madmaxoft: Mark it `virtual`, too.

lkolbly: Done.

src/Entities/Player.cpp
```diff
@@ -2641,8 +2641,105 @@ void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_R
...
+	}
+
+	cWorld * World = GetWorld();
+
+	// Check to see if any entity intersects any block being placed
+	class DoesIntersectBlock : public cEntityCallback
+	{
+	public:
+		const std::vector<cBoundingBox> & m_BoundingBoxes;
+		cWorld * m_World;
```
madmaxoft: `m_World` doesn't seem to be used in the callback anymore, or am I blind?

lkolbly: Nope, not blind, or if so then using a decent screen reader. :-)

PR #3851, opened 2017-07-13
Break the cactus block when it grows next to a block.

Bond-009:
Stops this from happening:
screen shot 2017-07-13 at 16 26 30

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1144,24 +1144,32 @@ int cChunk::GrowCactus(int a_RelX, int a_RelY, int a_RelZ, int a_NumBlocks)
 		if (UnboundedRelGetBlockType(a_RelX, Top + i, a_RelZ, BlockType) && (BlockType == E_BLOCK_AIR))
 		{
 			UnboundedRelFastSetBlock(a_RelX, Top + i, a_RelZ, E_BLOCK_CACTUS, 0);
-			if (
-				(
-					UnboundedRelGetBlockType(a_RelX + 1, Top + i, a_RelZ, BlockType) &&
-					(cBlockInfo::IsSolid(BlockType) || (BlockType == E_BLOCK_LAVA) || (BlockType == E_BLOCK_STATIONARY_LAVA))
-				) ||
-				(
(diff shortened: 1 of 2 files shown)


src/Chunk.cpp
```diff
@@ -1142,11 +1142,18 @@ int cChunk::GrowCactus(int a_RelX, int a_RelY, int a_RelZ, int a_NumBlocks)
...
-		if (UnboundedRelGetBlock(a_RelX, Top + i, a_RelZ, BlockType, BlockMeta) && (BlockType == E_BLOCK_AIR))
+		if (UnboundedRelGetBlockType(a_RelX, Top + i, a_RelZ, BlockType) && (BlockType == E_BLOCK_AIR))
 		{
-			// TODO: Check the surrounding blocks, if they aren't air, break the cactus block into pickups (and continue breaking blocks above in the next loop iterations)
 			UnboundedRelFastSetBlock(a_RelX, Top + i, a_RelZ, E_BLOCK_CACTUS, 0);
+			if (
+				(UnboundedRelGetBlockType(a_RelX + 1, Top + i, a_RelZ, BlockType) && (BlockType == E_BLOCK_AIR)) ||
+				(UnboundedRelGetBlockType(a_RelX - 1, Top + i, a_RelZ, BlockType) && (BlockType == E_BLOCK_AIR)) ||
+				(UnboundedRelGetBlockType(a_RelX, Top + i, a_RelZ + 1, BlockType) && (BlockType == E_BLOCK_AIR)) ||
+				(UnboundedRelGetBlockType(a_RelX, Top + i, a_RelZ - 1, BlockType) && (BlockType == E_BLOCK_AIR))
```
madmaxoft: The wiki says "solid material", so things such as torches should be acceptable, too.

Bond-009: 👍 Fixed

madmaxoft: Oh, I thought this should first be researched, the wiki is way too vague in its wording, I wasn't sure.

Bond-009: I checked it in Vanilla first, non-solids like torches and redstone comparators can be placed next to cacti, but solid blocks like fences etc cant

Issue #2015, opened 2015-05-14
Mobs not standing correctly on slabs and lilypads


falkseidl:
Mobs standing in the height of a full block, if they are walking on blocks that have a smaller height.
It happens, if they walk for their self on this blocks, but also if the player spawns a mob at that place.

Screenshots:
![2015-05-14_10 59 10](https://cloud.githubusercontent.com/assets/5732964/7628857/457f8eb2-fa28-11e4-87e0-225a0ee63f2a.png)

![2015-05-14_10 58 36](https://cloud.githubusercontent.com/assets/5732964/7628843/3029b272-fa28-11e4-996c-a41028a638ab.png)

SafwatHalaby:
The pathfinder has partial blocks awareness yet, it's in the [to-do list](http://forum.mc-server.org/showthread.php?tid=1895) and will be fixed :)

Thanks for pointing out the lilypad issue.

madmaxoft:
I don't think the pathfinder should be partial-block-aware. It should only provide paths through entire blocks.

This is an issue of the physics engine, though, that should really be aware of correct block hitboxes.

SafwatHalaby:
Good point, but in the meanwhile this can be delegated to `monster.cpp`:
"If next waypoint is in the air and there's a slab beneath it, lower it half a block"
"If next waypoint is in the air and there's a lilypad beneath it, lower it 1 block"

falkseidl:
Happens also on trapdoors or snow. So all special blocks like this.

SafwatHalaby:
yes, all partial blocks are like this.

SafwatHalaby:
I tried fixing it, but as @madmaxoft  said, looks like it requires changes to the physics engine. The physics doesn't care about half slabs at all. I completely neutralized the pathfinder and the mobs were still flying when spawned over slabs.

worktycho:
So basicly the physics engine needs to have full bounding box support.

SafwatHalaby:
Doesn't have to be full, can be "cheaty full", give special treatment to exotic blocks.

SafwatHalaby:
Fixing lilypads and snow sounds easy - I'll try to make the physics engine simply ignore flat blocks.

worktycho:
That was an accident. Are there any blocks that aren't either empty or half blocks?

worktycho:
I forgot, steps. What other shapes are there.

SafwatHalaby:
[...]

PR #3852, opened 2017-07-13
Sitting cats block enderchests from opening

Bond-009:
Also fixed an issue where normal chests couldn't be opened at the top of the world but enderchests could.

### DIFF
--- a/src/Mobs/Ocelot.h
+++ b/src/Mobs/Ocelot.h
@@ -2,7 +2,6 @@
 #pragma once
 
 #include "PassiveMonster.h"
-#include "../Entities/Entity.h"
 
 
 
--- a/src/Bindings/LuaFunctions.h
+++ b/src/Bindings/LuaFunctions.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include "Logger.h"
 #include <time.h>
 // tolua_begin
 
--- a/src/BlockInServerPluginInterface.h
+++ b/src/BlockInServerPluginInterface.h
@@ -11,7 +11,6 @@
 
 #include "Blocks/BlockPluginInterface.h"
 #include "World.h"
-#include "Root.h"
 #include "Bindings/PluginManager.h"
 
 
--- a/src/Blocks/BlockAnvil.h
+++ b/src/Blocks/BlockAnvil.h
@@ -2,7 +2,6 @@
 #pragma once
 
 #include "BlockHandler.h"
-#include "../World.h"
 #include "../Entities/Player.h"
 #include "../UI/AnvilWindow.h"
 
--- a/src/Blocks/BlockBed.h
+++ b/src/Blocks/BlockBed.h
@@ -4,10 +4,8 @@
 #pragma once
 
 #include "BlockEntity.h"
-#include "BlockHandler.h"
 #include "MetaRotator.h"
 #include "ChunkInterface.h"
-#include "../World.h"
 #include "../Entities/Entity.h"
 
 
--- a/src/Blocks/BlockBrewingStand.h
+++ b/src/Blocks/BlockBrewingStand.h
@@ -2,7 +2,6 @@
 #pragma once
 
 #include "BlockEntity.h"
-#include "BlockHandler.h"
 #include "MetaRotator.h"
 
 
--- a/src/Blocks/BlockFlowerPot.h
+++ b/src/Blocks/BlockFlowerPot.h
@@ -1,7 +1,6 @@
 
 #pragma once
 
-#include "BlockHandler.h"
 #include "BlockEntity.h"
 
 
--- a/src/Blocks/BlockFurnace.h
+++ b/src/Blocks/BlockFurnace.h
@@ -2,7 +2,6 @@
 #pragma once
 
 #include "BlockEntity.h"
-#include "../World.h"
 #include "../Blocks/BlockPiston.h"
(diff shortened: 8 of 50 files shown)


src/BlockEntities/EnderChestEntity.cpp
```diff
@@ -46,10 +48,36 @@ void cEnderChestEntity::SendTo(cClientHandle & a_Client)
...
+{
+	virtual bool Item(cEntity * a_Entity) override
+	{
+		return (
+			(a_Entity->GetEntityType() == cEntity::etMonster) &&
+			(static_cast<cMonster *>(a_Entity)->GetMobType() == eMonsterType::mtOcelot) &&
+			(static_cast<cOcelot *>(a_Entity)->IsSitting())
+		);
+	}
+};
```
madmaxoft: Rather than copying the code into two places, you should refactor it into a separate static method in the `cOcelot` class, something like `static bool cOcelot::IsCatSittingOnTopOfBlock(a_World, a_BlockX, a_BlockY, a_BlockZ)`

madmaxoft:
I'm not entirely sure whether the proposed change is a good direction to take - extending the `cOcelot` class, when, ultimately, we want to get rid of all the specific entity classes ("component refactor"). Thoughts, @cuberite/server people?

tigerw:
Is there any progress update on the component refactor?

madmaxoft:
Nothing specific, just bits and pieces around when adding new code.

src/Mobs/Ocelot.h
```diff
@@ -55,6 +55,8 @@ class cOcelot :
 	}
 	void SetCatType (eCatType a_CatType)    { m_CatType = a_CatType; }
 
+	static bool IsCatSittingOnBlock(cWorld * a_World, const Vector3d & a_BlockPosition);
```
madmaxoft: Add a doxycomment for this function.
Pass Vector3d by-value.

src/BlockEntities/EnderChestEntity.cpp
```diff
@@ -48,8 +49,16 @@ void cEnderChestEntity::SendTo(cClientHandle & a_Client)
...
 {
-	// TODO: cats are an obstruction
-	if ((GetPosY() < cChunkDef::Height - 1) && !cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())))
+	if (
+		(GetPosY() < cChunkDef::Height - 1) &&
+		(
+			!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
+			(
+				(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) &&
+				!cOcelot::IsCatSittingOnBlock(GetWorld(), Vector3d(GetPos()))
```
madmaxoft: Does this mean that a cat sitting on a chest that has a torch above it still doesn't block the chest from opening? Is that vanilla behavior?

src/BlockEntities/ChestEntity.cpp
```diff
@@ -219,32 +218,16 @@ void cChestEntity::DestroyWindow()
...
-		!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
+		(GetPosY() < cChunkDef::Height - 1) &&
 		(
-			(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) &&
-			!GetWorld()->ForEachEntityInBox(cBoundingBox(Vector3d(GetPosX(), GetPosY() + 1, GetPosZ()), 1, 1), FindSittingCat)
+			!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
+			(
+				(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ()) == E_BLOCK_AIR) &&
+				!cOcelot::IsCatSittingOnBlock(GetWorld(), Vector3d(GetPos()))
+			)
```
madmaxoft: I think it might be easier to understand if you broke this return statement into several `if` statements.
```cpp
// Check if there is a solid block above the chest:
if (GetPosY() < cChunkDef::height - 1)
{
  // <check block solidity>
}

// Check if there's a cat sitting on top:
if (cOcelot::IsCatSittingOnBlock(...))
{
  return true;
}

// Not blocked
return false;
```
(Note that I'm assuming that a cat sitting on top of a chest is not affected by the block type above the chest, and the check still works if the chest is at the top of the world, with the cat sitting on it being above the world)

Bond-009:
screen shot 2017-08-03 at 10 07 06

This should block the chest, but it doesn't. I hope that this is because of issue #2015 and not my code.

src/Mobs/Ocelot.h
```diff
@@ -55,6 +55,9 @@ class cOcelot :
 	}
 	void SetCatType (eCatType a_CatType)    { m_CatType = a_CatType; }
 
+	/** Returns true if the cat is sitting above the given position */
```
madmaxoft: `the cat is` -> `there's a cat`

Bond-009:
Why do I suck so hard with git?...

Issue #3676, opened 2017-04-23
[BUG] Appears the threshold was to small.


djlw78:
Client version:   1.11.2
Server OS:        Windows
Commit id:       7960d142ca3e28e0e4550faf131398899dca90dc

### Expected behavior
To handle UNDO/REDO w/ world edit quickly

### Actual behavior

[13:44:13] --- Started Log ---
[13:44:13] Cuberite  build id: 491
[13:44:13] from commit id: 7960d142ca3e28e0e4550faf131398899dca90dc built at: 491
[13:44:13] Creating new server instance...
[13:44:13] Reading server config...
[13:44:13] Starting server...
[13:44:13] Compatible clients: 1.8.x, 1.9.x, 1.10.x, 1.11.x
[13:44:13] Compatible protocol versions 47, 107, 108, 109, 110, 210, 315, 316
[13:44:13] WebServer: The server will run in unsecured HTTP mode.
[13:44:14] Put a valid HTTPS certificate in file 'webadmin/httpscert.crt' and its corresponding private key to 'webadmin/httpskey.pem' (without any password) to enable HTTPS support
[13:44:14] Loaded 726 crafting recipes
[13:44:14] Loaded 52 furnace recipes and 66 fuels
[13:44:14] Loaded 88 brewing recipes
[13:44:14] -- Loading Plugins --
[13:44:14] Initialised Core v.15
[13:44:14] Initialized ChatLog v.3
[13:44:14] [WorldEdit] Enabling WorldEdit v0.1.11
[13:44:14] -- Loaded 3 Plugins --
[13:44:15] Your WorldEdit plugin is up-to-date
[13:44:16] Did not find an acceptable spawnpoint. Generated a random spawnpoint position at {0.00, 168.00, 0.00}
[13:44:17] WebAdmin is running on port(s) 8080
[13:44:17] Startup complete, took 3421ms!
[13:47:08] User ***** authenticated with UUID cd4dad9bc5ef4747b33a904ed3500766
[13:47:09] Player ***** has joined the game
[13:47:42] Player ***** is executing command "/difficulty" in world "world" at world age 23751.
[13:47:44] Player ***** tried to execute forbidden command: "/difficulty"
difficulty
[13:47:49] Executing console command: "difficulty"
[13:47:49] Unknown command, type 'help' for all commands.
reload
[13:51:36] Executing console command: "reload"
[13:51:36] -- Loading Plugins --
[13:51:36] Disabled Core!
[...]

PR #3854, opened 2017-07-14
Simulators: Added area-based wakeup.

madmaxoft:
Provides better performance especially for handling larger `cBlockArea` writes (such as in #3676) and for explosions.

Also doxy-commented the simulator classes.


madmaxoft:
As measured on a Win32 release build, this improves simulator wakeup performance five to ten times.

NiLSPACE:
Should we wait merging this until we figure out why the redstone simulator takes so long to wake up in large areas?

madmaxoft:
That's a completely unrelated issue, this can be merged without any regard to it.

Issue #3588, opened 2017-02-26
Rails don't always reorient properly when corners formed.


Sheyin:
Client version:   1.10 (vanilla)
Server OS:        Windows 64
Commit id:       7c17f77c84711607bc0f2db5f2991887bf5c9b79 built at: 464

### Expected behavior
When rails are placed, and a corner is formed, the adjacent rail should rotate to continue the turn.

### Actual behavior
When the corner rail is placed, it (correctly) turns into a corner piece, but the next rail of the turn will not rotate to continue the turn.

### Steps to reproduce the behavior
Place two rails diagonal to one another as if forming a turn in the rails like below.  They should be facing the same direction.
![rails1](https://cloud.githubusercontent.com/assets/8756070/23337573/7200e664-fbb8-11e6-9e62-2f6578c2c7eb.jpg)

Place one rail adjacent to both pieces.  A corner piece will form and look like the image below.
![rails2](https://cloud.githubusercontent.com/assets/8756070/23337576/7972613e-fbb8-11e6-8cdd-f4c1ac3acf61.jpg)

HelenaKitty:
I was about to write a bug report about the same problem, but it seems I no longer need to!

sphinxc0re:
@Sheyin Thanks for the report! I wonder why noone ever encountered this bug...

PascalMinder:
The is also the problem if two power rails should climb one block, only one is correct. The other stays horizontal.

Issue #2914, opened 2016-01-30
Rails can transform types in some cases.


BitOfAByte:
I'm not sure exactly what causes this but a powered or activator rail can transform to a normal rail when a rail at one block lower is broken.

![Basic Issue 1](https://cloud.githubusercontent.com/assets/6497430/12693221/92467298-c6d2-11e5-92dd-45f6f1373c44.png)
![Basic Issue 2](https://cloud.githubusercontent.com/assets/6497430/12693220/9243b580-c6d2-11e5-98d2-89e947e1d5e1.png)

It is not related to being near the corner, it also works on an edge.

![Edge 1](https://cloud.githubusercontent.com/assets/6497430/12693243/1ce6260a-c6d3-11e5-8799-2619ea388403.png)
![Edge 2](https://cloud.githubusercontent.com/assets/6497430/12693244/1ce6987e-c6d3-11e5-82a1-d1ef299ce30b.png)

It is related to having the tracks run to not connect with it.

![Tracks not connecting 1](https://cloud.githubusercontent.com/assets/6497430/12693252/5f01e7fe-c6d3-11e5-97fa-4e7694f3e754.png)
![Tracks not connecting 2](https://cloud.githubusercontent.com/assets/6497430/12693253/5f050b78-c6d3-11e5-8440-e7948b8b724a.png)

It doesn't work if there is another rail connected to it.

![Not working with complete connection 1](https://cloud.githubusercontent.com/assets/6497430/12693301/950e6600-c6d4-11e5-84eb-db36b6eb23a5.png)
![Not working with complete connection 2](https://cloud.githubusercontent.com/assets/6497430/12693302/9511e08c-c6d4-11e5-9599-c66d3445fb34.png)

More than one rail can be changed at once.

![All at once 1](https://cloud.githubusercontent.com/assets/6497430/12693269/cdc7eb16-c6d3-11e5-839d-531e0bfe976b.png)
![All at once 2](https://cloud.githubusercontent.com/assets/6497430/12693270/cdc86852-c6d3-11e5-9269-8ca78350e6ac.png)

It works with powered rails to rails (activator rails and detector rails work in all cases too).

![Powered to rail 1](https://cloud.githubusercontent.com/assets/6497430/12693272/cdcd1a1e-c6d3-11e5-92f0-8cb1e37dd5cf.png)
[...]

PR #3855, opened 2017-07-14
Rail Fixes

### LINKED ISSUE #3848: Rails don't reorient properly when straight line formed
Client version:   1.12
Server OS:        Windows 8.1 x64 
Commit id:        d838ef7

### Expected behavior
When rails are placed, the adjacent rails should rotate to continue them. (Similar to #3588)

### Actual behavior
When the rail is placed, it correctly turn, but the adjacent rails will not rotate.

Place two rails in vertical position like below. 
bug_1_initial
bug_1_behavior

lkolbly:
Made cBlockHandler OnPlacedByPlayer call OnPlaced, made BlockRailHandler use NeighborChanged instead of OnNeighborChanged.

Fixes #3848, #3588, and #2914.


Seadragon91:
Tested and works good.

PR #3856, opened 2017-07-15
Allocate redstone component handlers upfront

peterbell10:
Replaces use of `cIncrementalRedstoneSimulator::CreateComponentHandler` with something more like `cBlockInfo::GetBlockHandler` which for large area fills I found to be about a 7x speedup.

All the handlers were already stateless apart from keeping track of their world so I just had to move `cWorld &` to the parameters.

Related issue: #3676

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
@@ -29,25 +29,48 @@
 
 
 
+const cRedstoneHandler * cIncrementalRedstoneSimulator::GetComponentHandler(BLOCKTYPE a_BlockType)
+{
+	struct sComponents:
+		public std::array<std::unique_ptr<cRedstoneHandler>, 256>
+	{
+		sComponents()
+		{
+			for (size_t i = 0; i != 256; ++i)
+			{
+				(*this)[i] = cIncrementalRedstoneSimulator::CreateComponent(static_cast<BLOCKTYPE>(i));
+			}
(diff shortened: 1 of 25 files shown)


src/Simulator/IncrementalRedstoneSimulator/IncrementalRedstoneSimulator.cpp
```diff
@@ -29,25 +29,48 @@
 
 
 
+const cRedstoneHandler * cIncrementalRedstoneSimulator::GetComponentHandler(BLOCKTYPE a_BlockType)
+{
+	struct sComponents:
+		public std::array<std::unique_ptr<cRedstoneHandler>, 256>
```
worktycho: How about making this an array of cRedstoneHandlers rather than an array of pointers to redstone handlers since they are just vtable pointers. Then we don't have to worry about any allocations.

worktycho: No, I'm being an idiot, the current code uses null to signal block that can't interact with redstone at all. Ideally we'd move all the actual solid block handling code out of needing a dedicated handler because they are so sommon and have a different quick scan strategy, but that would be out of scope for this PR.

tigerw:
I was hoping to get static thread locals back in ;p

xoft, are you still using VS 2013?

madmaxoft:
Why would you want static thread locals when it's much more elegant this way? :P

Yes, I'm still using VS2013, and I'm using VS2017. We decided on C++11, which is what VS2013 uses, so that's what we stick with for now.

PR #2314, opened 2015-07-04
Anvil Support and Bug Fixes

parlock:
I have implemented the support for Anvils to use enchantments.  And fixed several bugs in the existing Anvil code (UI not updating the output item, Repair name renaming not working as vanilla MC)

PR #3857, opened 2017-07-16
Added anvil enchantment handling.

### LINKED ISSUE #1608: Anvil and enchantments not working

lkolbly:
Fixes #1608.

Adds enchantment handling to anvils. Along the way, adds a variety of functions to cEnchantments and cItem to keep track of mutually-exclusive enchantments, enchantment level caps, and which items can have which enchantments.

Also, this PR hardcodes a lot of things about enchantments (mutual exclusions, valid items, level caps) in the C++ code, which makes me feel kinda bad inside. I don't know what it would take to put it in a config file, and enchantment level stuff from cEnchantments::AddItemEnchantmentWeights should also go to a config file. It may warrant some thought, I'd be curious what y'all think (it may be enough work to warrant another PR down the road, that's what I had planned on - this PR adding just anvil enchanting for now).

Due credit to @Rakete1111 and @parlock of #2183 and #2314, respectively, I used those PRs for guidance.


lkolbly:
It's failing Travis because I didn't document some Lua symbols, I'll do that tomorrow. Is there a way I can run that test on my local machine, or does it only test it on Travis?

Pokechu22:
Load the APIDump plugin and then run `apicheck`:

```
load APIDump
apicheck
```

lkolbly:
Thanks!

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6553,6 +6570,24 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
 					},
 					Notes = "Adds the specified amount to the item count. Returns self (useful for chaining).",
 				},
+				AddEnchantmentsFromItem =
+				{
+					Params =
+					{
+						{
+							Name = "Sacrifice",
```
madmaxoft: I would avoid calling it the `Sacrifice`, since that would imply that this function destroys the item somehow. How about `Additive`, or `ToAdd`?

src/Enchantments.cpp
```diff
@@ -167,6 +194,141 @@ bool cEnchantments::IsEmpty(void) const
...
+	case enchEfficiency:           return 5;
+	case enchSilkTouch:            return 1;
+	case enchUnbreaking:           return 3;
+	case enchFortune:              return 3;
+	case enchPower:                return 5;
+	case enchPunch:                return 2;
+	case enchFlame:                return 1;
+	case enchInfinity:             return 1;
+	case enchLuckOfTheSea:         return 3;
+	case enchLure:                 return 3;
```
madmaxoft: Style: the `case` statements should have one extra indent (indent is dictated by curly-braces).

madmaxoft: This goes for all the `switch` statements in this PR.

src/Enchantments.cpp
```diff
@@ -167,6 +194,141 @@ bool cEnchantments::IsEmpty(void) const
...
+	return 0;
+}
+
+
+
+
+
+bool cEnchantments::IsCompatibleWith(int a_EnchantmentID) const
+{
+	for (auto mutex: m_IncompatibleEnchantments)
```
madmaxoft: Calling it `mutex` is misleading us to think about thread sync. How about `excl`?

src/Enchantments.h
```diff
@@ -103,6 +104,9 @@ class cEnchantments
 	/** Returns true if there are no enchantments */
 	bool IsEmpty(void) const;
 
+	/** Returns true if all of our enchantments are compatible with the other enchantment */
```
madmaxoft: ... `with the specified enchantment`

src/Enchantments.h
```diff
@@ -111,6 +115,12 @@ class cEnchantments
 
 	// tolua_end
 
+	/** Get the XP cost multiplier for the enchantment (for anvils) */
+	static int GetMultiplier(int a_EnchantmentID, bool WithBook);
```
madmaxoft: `GetXPCostMultiplier`?

src/Enchantments.h
```diff
@@ -149,7 +159,19 @@ class cEnchantments
...
-} ;  // tolua_export
+
+public:
+	/** Make this class iterable */
+	cMap::iterator begin() { return m_Enchantments.begin(); }
+	cMap::iterator end()   { return m_Enchantments.end(); }
+
+private:
+	/** A vector of all of the mutually-exclusive sets of enchantments */
+	std::vector<std::set<int> > m_IncompatibleEnchantments;
```
madmaxoft: Why exactly does each `cEnchantments` object have a copy of this?

Personally, I'd make a static function that would return the data, and use the C++ singleton approximation (`static const` object inside the function) for a single initialization point.

src/Item.cpp
```diff
@@ -419,6 +419,199 @@ bool cItem::EnchantByXPLevels(int a_NumXPLevels)
...
+
+
+
+
+
+bool cItem::CanHaveEnchantment(int a_EnchantmentID)
+{
+	if (m_ItemType == E_ITEM_ENCHANTED_BOOK)
+	{
+		// Enchanted books can take anything
```
madmaxoft: Even mutually exclusive enchantments?

src/Item.cpp
```diff
@@ -419,6 +419,199 @@ bool cItem::EnchantByXPLevels(int a_NumXPLevels)
...
+		return true;
+	}
+
+	// The organization here is based on the summary at:
+	// http://minecraft.gamepedia.com/Enchanting
+	// as of July 2017 (Minecraft 1.12).
+
+	// Hand tool enchantments
+	std::set<int> SwordEnchantments =
```
madmaxoft: `static const`?

src/UI/SlotArea.cpp
```diff
@@ -1114,6 +1116,7 @@ void cSlotAreaAnvil::UpdateResult(cPlayer & a_Player)
 				Output.Empty();
 				SetSlot(2, a_Player, Output);
 				m_ParentWindow.SetProperty(0, 0, a_Player);
+				m_ParentWindow.SendSlot(a_Player, this, 2);
```
madmaxoft: The `SetSlot` function should send the slot to the player, doesn't it do so?

src/Item.h
```diff
@@ -201,6 +201,15 @@ class cItem
 	Returns true if the item was enchanted, false if not (not enchantable / too many enchantments already). */
 	bool EnchantByXPLevels(int a_NumXPLevels);  // tolua_export
 
+	/** Adds this specific enchantment to this item, returning the cost */
+	int AddEnchantment(int a_EnchantmentID, unsigned int a_Level, bool a_WithBook);
```
madmaxoft: Why not export this to Lua as well?

src/UI/SlotArea.cpp
lkolbly: I was having some trouble with the output slot updating correctly in the client when I commented the .SendSlot calls out, I'll look more into it and try to figure out why SetSlot isn't sending the slot, or why my client isn't picking up on it (to reproduce: just stick armor and a Protection enchanted book in an anvil, observe the output not appear. I'm using a 1.12 client).

If you want I can open a separate issue for this.

src/Item.cpp
lkolbly: Good point, a book can have either Depth Strider or Feather Falling, but not both. But, that's true for boots as well. This function doesn't do the mutual-exclusivity check. (so it's just a lookup table to tell whether the enchantment is good for the ItemType)

I wrote a docstring in the header. to make that less ambiguous.

src/Enchantments.h
lkolbly: That works much better.

Server/Plugins/APIDump/APIDesc.lua
lkolbly: Went with Additive, since ToAdd to me implies that it's merging the items themselves (whereas it's just using the enchantments as an additive).

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6553,6 +6570,49 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
...
+							Type = "boolean",
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+					},
+					Notes = "Adds the given enchantment at the given level to this item, following anvil enchantment combining rules. Returns the XP level cost of the addition.",
```
madmaxoft: The WithBook parameter should be documented - what it's used for.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6553,6 +6570,49 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
...
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "LevelCost",
+							Type = "number",
+						},
+					},
+					Notes = "Adds the enchantments from the sacrificial item to this item, returning the cost as if this were an anvil.",
```
madmaxoft: sacrificial -> additive, or specified.

src/Enchantments.h
```diff
@@ -103,6 +104,9 @@ class cEnchantments
 	/** Returns true if there are no enchantments */
 	bool IsEmpty(void) const;
 
+	/** Returns true if all of our enchantments are compatible with the specified enchantment */
+	bool IsCompatibleWith(int a_EnchantmentID) const;
```
madmaxoft: For thought: how about `CanAddEnchantment`, at least for me that is a more descriptive name.

madmaxoft: Also, what happens if specifying an enchantment already present? Should it return true or false? Doc.

src/Item.h
```diff
@@ -201,6 +201,16 @@ class cItem
 	Returns true if the item was enchanted, false if not (not enchantable / too many enchantments already). */
 	bool EnchantByXPLevels(int a_NumXPLevels);  // tolua_export
 
+	/** Adds this specific enchantment to this item, returning the cost */
+	int AddEnchantment(int a_EnchantmentID, unsigned int a_Level, bool a_WithBook);  // tolua_export
+
+	/** Adds the enchantments on a_Other to this item, returning the
+	XP cost of the transfer. */
+	int AddEnchantmentsFromItem(cItem & a_Other);  // tolua_export
```
madmaxoft: Does this modify `a_Other`? If not, why not make it `const`? If yes, document.

src/UI/SlotArea.cpp
madmaxoft: Output not appearing is what I'd expect from the code above - "clear the Output item, then set it into the output slot". Or am I misunderstanding something?

lkolbly: The output slot doesn't appear to be updating on the client. e.g. if you don't have the SendSlot on line 1204 (sending the output of a valid enchanting), then the client won't show the item even though SetSlot was called.

src/Item.h
lkolbly: It does not, but internally it requires cItem's begin() and end() to return const_iterators, which I just implemented.

src/Enchantments.h
lkolbly: CanAddEnchantment is a better name, and it also removes the ambiguity of if the enchantment is present (because you can add an enchantment to an item already with that enchantment).

Server/Plugins/APIDump/APIDesc.lua
lkolbly: Changed.

Server/Plugins/APIDump/APIDesc.lua
lkolbly: Documented.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6553,6 +6570,49 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
...
+						{
+							Name = "Enchantment ID",
+							Type = "number",
+						},
+						{
+							Name = "Level",
+							Type = "number",
+						},
+						{
+							Name = "WithBook",
```
madmaxoft: I'm still struggling with this. Would it make more sense to call it `FromBook`?

src/Enchantments.h
```diff
@@ -111,6 +115,12 @@ class cEnchantments
 
 	// tolua_end
 
+	/** Get the XP cost multiplier for the enchantment (for anvils) */
+	static int GetXPCostMultiplier(int a_EnchantmentID, bool WithBook);
```
madmaxoft: The `WithBook` (`FromBook`) should be documented here, too.

src/Item.h
```diff
@@ -201,6 +201,16 @@ class cItem
 	Returns true if the item was enchanted, false if not (not enchantable / too many enchantments already). */
 	bool EnchantByXPLevels(int a_NumXPLevels);  // tolua_export
 
+	/** Adds this specific enchantment to this item, returning the cost */
+	int AddEnchantment(int a_EnchantmentID, unsigned int a_Level, bool a_WithBook);  // tolua_export
```
madmaxoft: Document the book param.

src/UI/SlotArea.cpp
madmaxoft: Still this feels wrong. If you replace all the `SendSlot` calls with `SetSlot`, does it work?

Server/Plugins/APIDump/APIDesc.lua
lkolbly: FromBook works.

src/UI/SlotArea.cpp
lkolbly: No, it turns out that cSlotAreaTemporary::SetSlot doesn't call SendSlot (the workbench, the other cSlotAreaTemporary subclass, also called SendSlot whenever it called SetSlot).

I added a call to SendSlot at the end of SetSlot.

madmaxoft: That's great news - a proper fix :)

Issue #3653, opened 2017-03-30
Make ping packet compatible with server status tools


mathiascode:
### Expected behavior
Cuberite servers show up as online on all server list/server status websites.

### Actual behavior
Very few server lists think Cuberite is running, most report a Cuberite server as offline.

Current response from a Cuberite server:
```
{
   "description" : {
      "text" : "Cuberite - in C++!"
   },
[...]

PR #3859, opened 2017-07-18
cProtocolRecognizer: Support incomplete packets

peterbell10:
Looking into #3653 with dinnerbone's status query tool; `DataReceived` seems to always get first just the length and then the rest of the packet.  With this it works about as well as the other tools from the issue.

### DIFF
--- a/src/Protocol/ProtocolRecognizer.cpp
+++ b/src/Protocol/ProtocolRecognizer.cpp
@@ -83,46 +83,58 @@ void cProtocolRecognizer::DataReceived(const char * a_Data, size_t a_Size)
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


tigerw:
Merge?

madmaxoft:
wait

tigerw:
wait

madmaxoft:
I need to go through this thoroughly once again, the logic around `m_InPingForUnrecognizedVersion` seems weird, but I don't have the time. Hopefully tomorrow.

tigerw:
Well, TryRecogniseProtocol has a side effect which changes m_InPingForUnrecognizedVersion which is why it's checked twice, I suppose.

PR #3861, opened 2017-07-20
Remove smart pointer and other macros from Globals.h

peterbell10:
Gets rid of mixed usage of the `std::{smart}_ptr` and `{Smart}Ptr` as well as a few other macros which weren't really used anywhere.

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -38,10 +38,6 @@
 
 	#define OBSOLETE __declspec(deprecated)
 
-	// No alignment needed in MSVC
-	#define ALIGN_8
-	#define ALIGN_16
-
 	#define FORMATSTRING(formatIndex, va_argsIndex)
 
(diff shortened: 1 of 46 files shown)


madmaxoft:
Aaaargh, the underscore plague is here again to bite us

src/Globals.h
```diff
@@ -90,12 +86,7 @@
 
 	#define OBSOLETE __attribute__((deprecated))
 
-	#define ALIGN_8 __attribute__((aligned(8)))
-	#define ALIGN_16 __attribute__((aligned(16)))
-
 	// Some portability macros :)
-	#define stricmp strcasecmp
```
madmaxoft: You removed all the portability macros, so remove the comment, too.

peterbell10: Done

PR #3862, opened 2017-07-20
Tentative fix for player-limit race condition

### LINKED ISSUE #2257: Race condition in max player count
If more than one player attempts to join the server, in a short enough space of time, when the server is nearly at its max slot count, it can cause the server to allow more players on than the maximum.

tigerw:
* Attempts to fix #2257

Derived from d233e9843148313c71fbaba96ccff660e47b07b1

Tycho commented about DOS attacks if the kick logic was done during handshake, so I moved it to after authentication. I hope that this does not merely shift the problem from an artificially unresponsive server (limited by kicks during handshake due to too many handshakes) to an actual unresponsive server (limited by all those evil clients getting to the authenticate stage and loading down the auth. thread). But in the second case at least genuine clients can connect if the server can handle it :)

I also hope that the threading logic is sound.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -179,6 +179,8 @@ void cClientHandle::Destroy(void)
 
 	if (player != nullptr)
 	{
+		cRoot::Get()->GetServer()->PlayerDestroyed();
+
 		auto world = player->GetWorld();
 		if (world != nullptr)
 		{
@@ -322,6 +324,8 @@ void cClientHandle::Kick(const AString & a_Reason)
 
 void cClientHandle::Authenticate(const AString & a_Name, const AString & a_UUID, const Json::Value & a_Properties)
 {
+	cRoot::Get()->GetServer()->PlayerCreated();
+
 	cWorld * World;
 	{
 		cCSLock lock(m_CSState);
@@ -662,6 +666,9 @@ void cClientHandle::HandleNPCTrade(int a_SlotNum)
 
 void cClientHandle::HandlePing(void)
 {
+	/* TODO: unused function, handles Legacy Server List Ping
+	http://wiki.vg/Protocol#Legacy_Server_List_Ping suggests that servers SHOULD handle this packet */
+
 	// Somebody tries to retrieve information about the server
 	AString Reply;
 	const cServer & Server = *cRoot::Get()->GetServer();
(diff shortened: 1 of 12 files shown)


madmaxoft:
[Does not ~compute~ compile](https://www.youtube.com/watch?v=ZBAijg5Betw)

tigerw:
ey xoft, I wanted to change the type of NumPlayers and MaxPlayers to size_t considering they aren't expected to go negative. However, this means that the OnServerPing hook needs a signedness change. Is this OK?

madmaxoft:
That's okay.

src/ClientHandle.cpp
```diff
@@ -1860,17 +1867,14 @@ bool cClientHandle::HandleHandshake(const AString & a_Username)
 {
 	if (a_Username.length() > 16)
 	{
-		Kick("Your username is too long(>16 characters)");
+		Kick("Your username is too long (>16 characters)");
 		return false;
 	}
 
-	if (!cRoot::Get()->GetPluginManager()->CallHookHandshake(*this, a_Username))
```
worktycho: The removal of this `!`seems like it would cause a subtle change in behavior.

src/Server.cpp
```diff
@@ -656,6 +626,12 @@ void cServer::KickUser(int a_ClientID, const AString & a_Reason)
 
 void cServer::AuthenticateUser(int a_ClientID, const AString & a_Name, const AString & a_UUID, const Json::Value & a_Properties)
 {
+	if (GetNumPlayers() >= GetMaxPlayers())
```
worktycho: This still leaves potential for a race condition if multiple threads can execute AuthenticateUser. Is this function tied to a single thread? If not you can fix it by moving this condition to after taking the m_CSClients lock, so that only one thread can be between checking the condition and incrementing, but you should put a comment in here documenting that if you do so.

tigerw: Yup. I seem to have made a fatal incorrect assumption about concurrency (again :P)

src/ClientHandle.cpp
tigerw: The [API document on the hook](https://api.cuberite.org/OnHandshake.html) says that plugins should return true to deny entry. Is this correct?

worktycho: In which case the old code is buggy and needs to be fixed.

madmaxoft:
Still does not compile.

src/Server.cpp
peterbell10: Another option would be to increment `m_PlayerCount` first and then check if they're allowed in.  That way accessing `m_PlayerCount` is still wait free.

tigerw: Indeed, I may have tried something similar in d233e9843148313c71fbaba96ccff660e47b07b1. However, would this method have the potential to cause two clients vying for one remaining slot to both get kicked, with no one being let in?

tigerw:
Is this acceptable?

src/Server.cpp
peterbell10: As long the comparison is done on the result of the `fetch_add` then It shouldn't be possible.

tigerw: I was thinking if two threads incremented the count simultaneously and then proceeded to check the condition if would fail for both, but in fact, surely this function only ever is called by one thread per server run, since either authentication is on (and the single auth. thread calls it) or authentication is off, in which the server thread calls it.

Then if this is right, synchronisation is not needed...?

tigerw:
Even if the check is only done by one thread, I will leave it as-is (guarded by mutex) since in any case it seems like a good idea, less prone to break in the future as a side-effect of say, dynamically updating the ShouldAuthenticate variable when a config. file is changed, if that gets implemented.

PR #3864, opened 2017-07-21
Implement protocol level chunk sparsing

### LINKED ISSUE #2258: Add support for chunk sending bitmask
This would be a significant performance boost to initial chunk streaming.

peterbell10:
Closes #2258

`cChunkDataSerializer` now operates on a `cChunkData` instead of plain arrays so it can use the "Primary Bit Mask" to avoid serialising anything for empty sections.  I haven't profiled it but it should be an improvement.

Tested with 1.8.0, 1.9.0 and 1.9.4 (the three different implementations).

References for [current](http://wiki.vg/Protocol#Chunk_Data) and [old](http://wiki.vg/index.php?title=SMP_Map_Format&oldid=7164) chunk formats.


src/ChunkData.h
```diff
@@ -96,23 +99,14 @@ class cChunkData
...
 
-	struct sChunkSection
-	{
-		BLOCKTYPE  m_BlockTypes   [SectionHeight * 16 * 16]    ;
-		NIBBLETYPE m_BlockMetas   [SectionHeight * 16 * 16 / 2];
-		NIBBLETYPE m_BlockLight   [SectionHeight * 16 * 16 / 2];
-		NIBBLETYPE m_BlockSkyLight[SectionHeight * 16 * 16 / 2];
-	};
+	/** Returns the number of sections present (i.e. non-air). */
+	UInt32 Sections() const;
```
madmaxoft: Use a more descriptive name, such as `NumSections`

src/Protocol/ChunkDataSerializer.cpp
```diff
@@ -77,32 +95,46 @@ void cChunkDataSerializer::Serialize47(AString & a_Data, int a_ChunkX, int a_Chu
...
-		Packet.WriteBEUInt8(static_cast<unsigned char>(BlockType >> 4));
-	}
+		for (size_t BlockIdx = 0; BlockIdx != cChunkData::SectionBlockCount; ++BlockIdx)
+		{
+			BLOCKTYPE BlockType = a_Section.m_BlockTypes[BlockIdx] & 0xFF;
+			NIBBLETYPE BlockMeta = a_Section.m_BlockMetas[BlockIdx / 2] >> ((BlockIdx & 1) * 4) & 0x0f;
+			Packet.WriteBEUInt8(static_cast<unsigned char>(BlockType << 4) | BlockMeta);
+			Packet.WriteBEUInt8(static_cast<unsigned char>(BlockType >> 4));
+		}
+	});
```
madmaxoft: Style: add one more level of indent for the lambda function's body.

src/ChunkData.h
peterbell10: `NumSections` is already used for the total number of sections per chunk.  How about `NumPresentSections`?

src/ChunkData.cpp
```diff
@@ -46,103 +39,90 @@ cChunkData::cChunkData(cAllocationPool<cChunkData::sChunkSection> & a_Pool) :
...
-				Free(m_Sections[i]);
-				m_Sections[i] = other.m_Sections[i];
-				other.m_Sections[i] = nullptr;
-			}
-		}
 		return *this;
 	}
-#endif
+
+	if (&m_Pool != &a_Other.m_Pool)
```
worktycho: This shouldn't happen, and arguably the whole memory pool thing is badly thought out and should be redone, but I don't want to get into that argument again because we end up in the caching debate again.

src/ChunkData.cpp
```diff
@@ -325,18 +305,54 @@ NIBBLETYPE cChunkData::GetSkyLight(int a_RelX, int a_RelY, int a_RelZ) const
...
 		}
 	}
-	return copy;
+}
+
+
+
+
+
+cChunkData cChunkData::Copy(void) const
```
worktycho: This should be the using the code from the copy constructor so that copies are explict, as they are really expensive.

src/ChunkDataCallback.h
```diff
@@ -101,3 +101,37 @@ class cChunkDataSeparateCollector :
...
+	};
+
+	cChunkDataCopyCollector():
+		m_Pool(cpp14::make_unique<MemCallbacks>()),
+		m_Data(m_Pool)
+	{
+	}
+
+
+	cListAllocationPool<cChunkData::sChunkSection, cChunkData::NumSections> m_Pool;  // Keep 1 chunk worth of reserve
```
worktycho: I don't think this is a good idea, as we are only using this pool for 1 chunk data so there no point having a pool. Again, in hindsight I think the pool adds no value, and we need to put more thought into how we handle large buffer.

peterbell10: `cListAllocationPool` isn't thread safe so this _has_ to have it's own pool.  I could maybe add a new type of pool that uses a static buffer though.

src/ChunkData.cpp
peterbell10: Perhaps it shouldn't but handling it properly seems better to me than creating subtly wrong code.

worktycho: The original implementation basically assumed one pool per world and chunk data objects would stay on the world thread so it was a sensible assumption to make. Given the changes you are now making I see that this is now worth having.

src/ChunkDataCallback.h
worktycho: Oh right, of course I forgot that, not sure what the sensible solution here is, because the allocation pool implementation is so stupid.

src/ChunkData.cpp
```diff
@@ -46,103 +39,90 @@ cChunkData::cChunkData(cAllocationPool<cChunkData::sChunkSection> & a_Pool) :
...
 
 
 
 
-#if __cplusplus < 201103L
-	// auto_ptr style interface for memory management
-	cChunkData::cChunkData(const cChunkData & a_Other) :
-		m_IsOwner(true),
-		m_Pool(a_Other.m_Pool)
+cChunkData::cChunkData(const cChunkData & a_Other):
```
peterbell10: Can you explain what's wrong with having copy construct and assign?
I've tried removing them but I just end up with two functions `Copy` and `Assign` that in essence do the exact same thing.

worktycho: Its that copies are very expensive, we are copying a potentially 200K object. And the copy function should alone be all thats needed because you then move assign the result of that.

peterbell10: That doesn't work I'm afraid.  First copy creates a copy with the same pool, then if it gets moved into an object with a different pool it does _another_ copy.  In fact maybe the copy function should be removed altogether to avoid that happening.

worktycho: Or we just get rid of the pool. I'm not sure it actually adds any benefit, and it causes issues like this.

peterbell10: Should I just get rid of all trace of `cAllocationPool` then?

madmaxoft: I'm not so sure, the idea still seems good to me - we have a special pool for allocating chunk data, and that pool will (later on) have methods to deal with out-of-memory stuff.

peterbell10: Well if the pools are staying then I think the choice is between keeping the copy operator or to have the same thing but call it `Assign`.  Either way I'm thinking the `Copy` function needs to be removed.

worktycho: @madmaxoft But we never are, because predicting/handling out of memory on a consumer OS is basically impossible. The only way I am aware of to handle it in a sane way is a configurable application limit on pool size.

madmaxoft: A configurable limit is something we'd like to have, too.

worktycho: @madmaxoft The problem is how do you handle the default and min. A max could be implemented with a counter, its only if we cache things that a pool is useful, and you need a min memory for that, not something that easy to set a default for.

peterbell10:
What's happening with this?  I'm not really sure what changes are needed.

worktycho:
At this point I'd approve the changes with the caveat that I think the whole memory pool idea needs rework.

peterbell10:
@madmaxoft could you take another look at this?  I've been sitting on similar functionality for anvil but it's blocked waiting for this.

PR #3858, opened 2017-07-16
Updated armor cover calculation.

lkolbly:
Fixes #2975.

Updates how armor cover is calculated. Now has a function which calculates the armor cover and a function which calculates the armor cover from enchantments. Armor durability reduction is just from the amount blocked by the armor (not enchantments).

Has been tested with hits from other players (bare handed and with swords) and from zombies (bare handed), with the target player both wearing armor and not wearing armor.

PR #3866, opened 2017-07-22
Fixed check to see if block entity is in merge source.

lkolbly:
If you create a nether world using seed "1760462746", then (while generating the nether fortress, I believe) there will be a mob spawner (block type 52) which has no corresponding block entity in a_Src, causing a segfault here.

(to reproduce: Create the above world with a debugger, keep recreating the world until it crashes, 3-5 times should do it, I find it only segfaults sometimes)

This may hint at a bigger issue (mob spawners not having block entities in the source material), but right here the check should not copy the block entity if it does not exist.


tigerw:
Fix for CI run test crash at #3858.

PR #3869, opened 2017-07-23
Implement Forge protocol handshake support

satoshinm:
Adds initial support for the Forge handshake protocol, as documented on http://wiki.vg/Minecraft_Forge_Handshake

Includes a Lua API to allow plugins to register as server-side Forge mods, which are included in the server list ping and ModList packet in the handshake exchanged with Forge clients.

No support for adding new game content in this pull request, but this lays the groundwork, and even without it currently allows server plugins to be written to hook the login handshake to require or deny specific client mods (or Forge itself). Example/test plugin: https://github.com/satoshinm/SampleForgeMod.lua/blob/master/main.lua


Pokechu22:
This is neat!

I don't have time to fully look over it (and I'm not exactly too deeply involved in cuberite anymore).  However, was the wiki.vg article up-to-date?  Or was there anything on it that is no longer accurate?  If there's anything newer to be added, it'd be great if you could update it for me (or point it out and I'll take a look).

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,304 @@
...
+			auto &ServerMods = m_Client->GetForgeMods();
+
+			size_t ModCount = ServerMods.GetNumMods();
+
+			Buf.WriteBEInt8(Discriminator_ModList);
+			Buf.WriteVarInt32(static_cast<UInt32>(ModCount));
+			for (size_t i = 0; i < ModCount; ++i)
+			{
+				Buf.WriteVarUTF8String(ServerMods.GetModNameAt(i));
+				Buf.WriteVarUTF8String(ServerMods.GetModVersionAt(i));
```
lkolbly: I don't know the Forge protocol at all, but does the order that the mods are sent actually matter? If not, then in cForgeMods you could do away with the parallel arrays m_ModNames and m_ModVersions and just have the m_Mods (and then iterate through that with the map::iterator).

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,304 @@
...
+	// Dimension TODO
+	Message.push_back('\0');
+	Message.push_back('\0');
+	Message.push_back('\0');
+	Message.push_back('\0');
+
+	m_Client->SendPluginMessage("FML|HS", Message);
+}
+
+AStringMap cForgeHandshake::ParseModList(const char * a_Data, size_t a_Size)
```
lkolbly: What if this returned a cForgeMods? Or, could this be a method of the cForgeMods class? (just thinking, then it's more clear what this function is actually returning)

src/Server.h
```diff
@@ -214,6 +229,18 @@ class cServer
 	int m_MaxPlayers;
 	bool m_bIsHardcore;
 
+	// TODO: std::map<int, cForgeMods &> data structure?
```
lkolbly: Yeah, you should be able to do std::map<int, cForgeMods>, and then index it using the protocol number (then you wouldn't have to have a separate variable for every version, you could just create a cForgeMods instance whenever someone registers a new protocol).

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11689,6 +11797,40 @@ end
...
+						{
+							Name = "ModVersion",
+							Type = "string",
+						},
+						{
+							Name = "ProtocolVersionNumber",
+							Type = "number",
+						},
+					},
+					Notes = "Add a Forge mod name/version to the server ping list for all protocol versions.",
```
lkolbly: For all protocol versions, or just the specified one?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11689,6 +11797,40 @@ end
...
+						{
+							Name = "ModName",
+							Type = "string",
+						},
+						{
+							Name = "ModVersion",
+							Type = "string",
+						},
+					},
+					Notes = "Add a Forge mod name/version to the server ping list.",
```
madmaxoft: What happens if two plugins register the same `ModName`, possibly with different versions?

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: Do we even want per-protocolversion mods? After all, the server is usually meant to be played by a single client version, the multi-version support is there mainly for transitions and for historical purposes.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1466,6 +1476,16 @@ end
...
 				},
+				IsModded =
+				{
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns true if the client is modded with Forge.",
```
madmaxoft: Should this be Forge-specific, or (future-proofing) would this mean any client-side mod? Perhaps consider renaming to indicate Forge-specific.

src/Bindings/PluginManager.h
```diff
@@ -142,6 +142,7 @@ class cPluginManager
 		HOOK_WEATHER_CHANGING,
 		HOOK_WORLD_STARTED,
 		HOOK_WORLD_TICK,
+		HOOK_LOGIN_FORGE,
```
madmaxoft: Alpha-sort

src/ClientHandle.cpp
```diff
@@ -64,6 +64,8 @@ float cClientHandle::FASTBREAK_PERCENTAGE;
 // cClientHandle:
 
 cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
+	m_ForgeHandshake(this),
+	m_ForgeMods(0),
```
madmaxoft: Never EVER mix a pointer type with an integer type. Initialize with `nullptr`, or even better, use a smart pointer.

src/Protocol/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ SET (SRCS
 	Protocol_1_11.cpp
 	Protocol_1_12.cpp
 	ProtocolRecognizer.cpp
+	ForgeHandshake.cpp
+	ForgeMods.cpp
```
madmaxoft: Alpha-sort

src/Protocol/CMakeLists.txt
```diff
@@ -27,6 +29,8 @@ SET (HDRS
 	Protocol_1_11.h
 	Protocol_1_12.h
 	ProtocolRecognizer.h
+	ForgeHandshake.h
+	ForgeMods.h
```
madmaxoft: Alpha-sort

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+#include "Globals.h"
+#include "ForgeHandshake.h"
+#include "json/json.h"
+#include "../Server.h"
+#include "../Bindings/PluginManager.h"
+#include "../ClientHandle.h"
+#include "../Root.h"
+
+cForgeHandshake::cForgeHandshake(cClientHandle *client) : m_IsForgeClient(false), m_Errored(false), m_Client(client)
```
madmaxoft: `a_Client`; put member initializations onto separate lines.

src/Protocol/ForgeHandshake.cpp
madmaxoft: I vote for returning a `cForgeMods` instance.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+AStringMap cForgeHandshake::ParseModList(const char * a_Data, size_t a_Size)
+{
+	cByteBuffer Buf(a_Size);
+	Buf.Write(a_Data, a_Size);
+
+	Int8 Discriminator;
+	Buf.ReadBEInt8(Discriminator);
+
+	ASSERT(Discriminator == 2);
```
madmaxoft: Do not assert. Check extensively, consider all client code broken-by-default. :)

src/Protocol/ForgeHandshake.cpp
madmaxoft: Style: Separate top-level functions from each other by 5 empty lines.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,304 @@
...
+					break;
+				}
+			}
+			break;
+		}
+
+		default:
+			LOG("Unexpected Forge packet %d received", Discriminator);
+			SetError();
+			return;
```
madmaxoft: Style: Braces for this code block.

src/Protocol/ForgeHandshake.h
```diff
@@ -0,0 +1,65 @@
...
+#include <stddef.h>
+
+// fwd:
+namespace Json
+{
+	class Value;
+}
+class cClientHandle;
+
+class cForgeHandshake
```
madmaxoft: Style: 5 empty lines before the class definition.

src/Protocol/ForgeHandshake.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+	cForgeHandshake(cClientHandle * client);
+
+	void AugmentServerListPing(Json::Value & ResponseValue);
+	void BeginForgeHandshake(const AString & a_Name, const AString & a_UUID, const Json::Value & a_Properties);
+
+	void SendServerHello();
+	void DataReceived(cClientHandle * a_Client, const char * a_Data, size_t a_Size);
+
+	AStringMap ParseModList(const char * a_Data, size_t a_Size);
```
madmaxoft: All public symbols should have doxy-comments about what they do, what parameter constaints they expect etc. For private symbols this is optional, but highly recommended as well.

src/Protocol/ForgeHandshake.h
```diff
@@ -0,0 +1,65 @@
...
+		ClientPhase_WAITINGSERVERCOMPLETE = 3,
+		ClientPhase_PENDINGCOMPLETE = 4,
+		ClientPhase_COMPLETE = 5,
+	};
+
+	enum
+	{
+		ServerPhase_WAITINGCACK = 2,
+		ServerPhase_COMPLETE = 3,
+	};
```
madmaxoft: These enums should probably be moved into the cpp file, they are not part of the public interface, changing them will cause big unneeded recompilation.

src/Protocol/ForgeHandshake.cpp
madmaxoft: You could even remove `cForgeMods` completely and just use a plain `AStringMap` in its place.

src/Protocol/Protocol.h
```diff
@@ -62,6 +63,10 @@ class cProtocol
 
 	/** Called when client sends some data */
 	virtual void DataReceived(const char * a_Data, size_t a_Size) = 0;
+	void ForgeDataReceived(const char * a_Data, size_t a_Size)
```
madmaxoft: Add an empty line and doxy-comment.

madmaxoft: Will this need to be `virtual`?

src/Protocol/ForgeHandshake.cpp
satoshinm: Removed `m_ModVersions`, but if I use an AStringMap instead of cForgeMods class, what would be the best way to bridge this to the Lua API? Can I bridge arrays, or maps to Lua with tolua++, is there an example I could follow of how to do this? If so I agree cForgeMods could be removed entirely (I only added it since the Lua interface seemed to only support simple types, as far as I could find).

Server/Plugins/APIDump/APIDesc.lua
satoshinm: I think so, my reasoning is the Forge mod version is inherently game-version specific (some mods have taken to putting the game version into their version itself, for example ironchests is version "1.12-7.0.31.818"). It remains to be seen but the way I envision it, there would be one Cuberite plugin, supporting multiple client mod versions, since there may be only minor or even no differences between them in a given mod. And the mod version has to be correct (different for each game version) for the client to show the server as compatible.

Maybe there's another way to accomplish this. Instead of a per-protocol mod concept (where the mod won't even show up on other protocol versions), maybe the register mod function could take the mod name, and then a map of protocol version to mod version? On the other hand, per-protocol mods may allow for smoother upgrade transitions: if a mod isn't supported for or applicable to a version, it won't even appear in that version's ModList.

src/Protocol/ForgeHandshake.cpp
lkolbly: cItems (in Item.h:223) is a class built solely to wrap around a vector of items. There they do it also so that they can raise appropriate warnings for invalid parameters.

So I don't know if tolua can bridge fancy types, but whoever wrote cItems didn't think it could (and I think it'd require the std::map class declaration being marked as tolua_export, or manual bindings).

src/Server.h
satoshinm: Removed the separate member variables in favor of a map, but not clear on how I can get the list of supported protocols? I see there is an unnamed enum in cProtocolRecognizer, and GetVersionTextFromInt() switch/cases on each version, and there are string representations `MCS_CLIENT_VERSIONS` and `MCS_PROTOCOL_VERSIONS`, but couldn't find a method to get an array/list/vector of the supported protocols. For now I added an array I iterate over to initialize (must be done at initialization time, so that RegisterForgeMod() adds to all supported protocols by default) in cServer, but should this array be moved to cProtocolRecognizer?

Server/Plugins/APIDump/APIDesc.lua
satoshinm: Forge-specific, or technically, if the "Forge Mod Loader" is installed on the client which is a part of Forge. Won't pick up LiteLoader or other loaders, or other non-loader mods. I'll change IsModded to "IsForgeClient", a clearer description.

src/Protocol/Protocol.h
satoshinm: I think it shouldn't need to be, the way I currently implemented it is the Forge protocol is implemented only within cForgeHandshake, not each cProtocol_X class (no need to override) -- it hasn't differed significantly between game protocol versions, any changes should be minor enough to handle within the cForge classes themselves, imho.

satoshinm:
@Pokechu22 Yes, thanks for the wiki.vg reference! I have not found any errors with it so far, Forge's handshake protocol hasn't appeared to have changed much since then (1.7-1.8). The main changes up to 1.12 seem to be related to the registry ([1.11.x](https://github.com/MinecraftForge/MinecraftForge/tree/1.11.x/src/main/java/net/minecraftforge/fml/common/registry) vs [1.12.x](https://github.com/MinecraftForge/MinecraftForge/tree/1.12.x/src/main/java/net/minecraftforge/fml/common/registry)), older versions have many more registry classes, but the wiki documentation, and this pull request, hasn't gotten that far yet to document/implement the registry stuff. Maybe later :)

satoshinm:
Note: CircleCI https://circleci.com/gh/cuberite/cuberite/3700 passed but TravisCI https://travis-ci.org/cuberite/cuberite/jobs/257166762 is failing:

```
cPluginManager.HOOK_LOGIN_FORGE (constant)
```

to add  to Plugins/APIDump/Classes/Plugins.lua...

Server/Plugins/APIDump/APIDesc.lua
satoshinm: They'll overwrite, with the new version. What would be the preferred error handling strategy in this case, check if a mod of the name is already registered, if so change the Lua API to return a boolean false?

Related, but perhaps out of scope: would it be possible/worthwhile to have a plugin instance somehow associated with its registered mods, so they can be automatically deregistered when the plugin unloads? If this association was present, then attempting to re-register a duplicate mod could also generate a nice log message saying which plugin already claimed it. Not clear how to implement though.

src/ClientHandle.h
```diff
@@ -377,13 +385,18 @@ class cClientHandle  // tolua_export
 
 	bool IsPlayerChunkSent();
 
+	cForgeHandshake m_ForgeHandshake;
+
+	cForgeMods * m_ForgeMods;
```
peterbell10: If this is owning it should be a `std::unique_ptr`

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+void cForgeHandshake::BeginForgeHandshake(const AString & a_Name, const AString & a_UUID, const Json::Value & a_Properties)
+{
+	ASSERT(m_IsForgeClient);
+
+	m_Name = &a_Name;
+	m_UUID = &a_UUID;
+	m_Properties = &a_Properties;
+
+	AStringVector Channels = { "FML|HS", "FML", "FML|MP", "FML", "FORGE" };
```
peterbell10: If this never grows then it should just be a `std::array`.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,352 @@
...
+	Buf.Write(a_Data, a_Size);
+
+	UInt32 NumMods;
+	Buf.ReadVarInt32(NumMods);
+
+	for (size_t i = 0; i < NumMods; ++i)
+	{
+		AString Name, Version;
+		Buf.ReadVarUTF8String(Name);
+		Buf.ReadVarUTF8String(Version);
```
peterbell10: You should probably test that the reads don't fail.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+		Mods.insert(std::pair<AString, AString>(Name, Version));
```
peterbell10: `insert` here actually takes a `std::pair<const AString, AString>` but it's probably cleanest to just use uniform initialisation:
```
Mods.insert({ Name, Version });
```

src/Protocol/ForgeMods.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+
+
+cForgeMods::cForgeMods(AStringMap a):m_Mods(a)
+{
+	for (const auto & item: a)
+	{
+		m_ModNames.push_back(item.first);
+	}
+}
```
peterbell10: You're copying `a` twice here. You should either take it by const reference or you should move into `m_Mods` after filling `m_ModNames`.

src/Protocol/ForgeMods.h
```diff
@@ -0,0 +1,46 @@
...
+	// tolua_end
+
+	cForgeMods();
+	cForgeMods(AStringMap a);
+
+	static const cForgeMods & Unmodded(void);
+
+	void Add(AString & a_Name, AString & a_Version);
+
+	void Remove(AString & a_Name);
```
peterbell10: Everything taking strings here should be const references.

src/Protocol/ForgeMods.cpp
satoshinm:
```c++
	for (const auto & item: a)
	{
		m_ModNames.push_back(item.first);
	}
```

@peterbell10 I think I understand the problem, but having some trouble fixing it. How exactly could I change it to take a const reference? Is the const reference not `const auto &` in the loop variable, or do I need it elsewhere? Or change `AStringVector m_ModNames` to be a vector of references instead somehow? Newbie at C++...

lkolbly: I think the issue (correct me if I'm wrong) is actually in
```
cForgeMods::cForgeMods(AStringMap a):m_Mods(a)
```
When the cForgeMods constructor is called, like so:
```
cForgeMods ModList(MyMapOfModThings);
```
The variable MyMapOfModThings is copied into the stack space of the constructor method, where it is copied again to the m_Mods member variable (via the map constructor).

If you do:
```
cForgeMods::cForgeMods(const AStringMap & a):m_Mods(a)
```
Then you skip the first copy, and it copies directly from MyMapOfModThings into the member variable.

(actually, I think just passing by reference does that as well, but you don't modify a, so might as well pass it as const)

Server/Plugins/APIDump/APIDesc.lua
lkolbly: What if the Cuberite plugin was required to support all protocol versions supported by the server? I don't play with a lot of Forge mods, but the few I've seen all require the client to have the mod in order to join the server, the mod isn't optional (Imagine if the mod adds a block type, every client connecting would have to have the plugin, otherwise they wouldn't be able to see the block).

So then if the plugin only worked with some protocol versions, the server administrator could configure the server to only allow supported protocol versions (this is a thing that can happen, right? I thought I saw a PR about this somewhere).

Then we don't have to keep track of which forge mods support which protocols. There's just a single list of forge mods, which have to support all the protocols that the server supports.

I don't know if that's too limiting in terms of Forge mods, though. I don't know of any examples that benefit from per-protocolversion mods, but then I haven't seen many examples.

Pokechu22: There are some forge mods that only need to be on the client (or the server), though for the most part forge isn't used for those kinds of changes (as it is overkill).

But in this case, the issue isn't so much supporting/not supporting different protocol versions, as getting the appropriate mod version number of different protocol versions.

madmaxoft: Okay, now it makes sense to me to have a protocol-specific list of mods. Let's keep it this way for now.

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: I think the best approach, from the plugin author's perspective, would be to raise a Lua error. In order to do that, though, you'll need to bind the function manually, so that it can do the check itself and raise the error. See the `cLuaState::ApiParamError` for code that raises a Lua error.

With such behavior, a plugin couldn't do much damage if it didn't check for errors, and it could still provide some kind of fallback by using `pcall`:
```lua
local isSuccess = pcall(cServer:Get().RegisterForgeMod, "name", "version")
if not(isSuccess) ...
```

satoshinm: Changed to manual binding for RegisterForgeMod\*, tested with:

```lua
local isSuccess, err = pcall(function () cRoot:Get():GetServer():RegisterForgeMod("foo", "1.2.3") end)
```

in https://github.com/satoshinm/SampleForgeMod.lua/commit/5cd2ea580ad411df96125bd366b4ae476c0c6a05

src/Protocol/ForgeMods.cpp
satoshinm: Changed to `cForgeMods::cForgeMods(const AStringMap & a):m_Mods(a)`

satoshinm:
@madmaxoft incorporated all the feedback (think I got all of it), anything else?

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,365 @@
...
+			LOG("ParseModList failed to read mod version at i=%zu", i);
+			break;
+		}
+
+		Mods.insert({Name, Version});
+
+		LOGD("ParseModList name=%s, version=%s", Name.c_str(), Version.c_str());
+	}
+
+	return Mods;
```
madmaxoft: I don't think the extreme logging is needed, it would only spam the server console when broken clients (or new Forge versions with different protocol) connect. If you want to keep it, make it `LOGD` (in Debug builds only).

Also, the code is broken into too little pieces needlessly by the empty lines. The entire `for`-loop should be a single block. Basically you should be able to add a comment after each empty line that would describe (in high level) what the next block is doing. Description such as `Read all mod names and versions` is okay, description such as `Push the parsed name and version into the map` is not okay, because the code makes it obvious.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+
+
+AStringMap cForgeHandshake::ParseModList(const char * a_Data, size_t a_Size)
+{
+	AStringMap Mods;
+
+	if (a_Size < 4)
+	{
+		LOG("ParseModList invalid packet, missing length (size=%zu)", a_Size);
```
madmaxoft: This goes for all `%zu` usages in this PR

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,365 @@
...
+					Ack.push_back(ServerPhase_COMPLETE);
+					m_Client->SendPluginMessage("FML|HS", Ack);
+
+					// Now finish logging in
+					m_Client->FinishAuthenticate(*m_Name, *m_UUID, *m_Properties);
+					break;
+				}
+			}
+			break;
+		}
```
madmaxoft: This `switch` makes the function rather large and contains lots of rather unrelated stuff. Especially with the nested `switch`.

I'd move each `case`'s contents into a separate function, then compress the `switch` to single-line `case`s:
```cpp
switch (a_Data[0])
{
  case Discriminator::ClientHello: return HandleClientHello(a_Data, a_Size);
  ...
}
```

src/Protocol/ForgeHandshake.h
```diff
@@ -0,0 +1,58 @@
...
+	/** True if the Forge handshake is in an errored state. */
+	bool m_Errored;
+
+	/** The client handle undergoing this Forge handshake. */
+	cClientHandle * m_Client;
+
+	/** Values saved from BeginForgeHandshake() for continuing the normal handshake after Forge completes. */
+	const AString * m_Name;
+	const AString * m_UUID;
+	const Json::Value * m_Properties;
```
madmaxoft: Store these by-value, rather than by-pointer, to make copies instead of relying on the pointers to still be valid when used at a later time.

src/Protocol/ForgeMods.h
```diff
@@ -0,0 +1,46 @@
...
+	void Remove(const AString & a_Name);
+
+	AStringMap GetMods() { return m_Mods; }
+
+	bool HasMods() { return m_Mods.size() != 0; }
+
+private:
+	AStringMap m_Mods;
+	AStringVector m_ModNames;
+} ;  // tolua_export
```
madmaxoft: Is there any specific reason why the whole `cForgeMods` class can't be replaced with simple `AStringMap`? Is there any case where the order of the mods is important?

There's a special binding for `AStringMap` that allows you to give and receive it from Lua functions just like any other type:
```cpp
cLuaState L(...);
AStringMap map;
L.CallFunction("DoSomething", map);  // Call a function that has an AStringMap param

AStringMap res;
L.CallFunction("ReturnSomething", cLuaState::Ret, res);  // Call a function that returns an AStringMap
```

src/Protocol/Protocol.h
```diff
@@ -63,6 +64,12 @@ class cProtocol
 	/** Called when client sends some data */
 	virtual void DataReceived(const char * a_Data, size_t a_Size) = 0;
 
+	/** Called when client sends some data over the FML|HS plugin channel */
+	void ForgeDataReceived(const char * a_Data, size_t a_Size)
+	{
+		m_Client->m_ForgeHandshake.DataReceived(m_Client, a_Data, a_Size);
+	}
```
madmaxoft: Since this is only ever called from within the `cClientHandle`, why does it have to go through `cProtocol`? It brings in many dependencies into this header file needlessly.

src/Server.cpp
```diff
@@ -220,6 +220,25 @@ bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_Shoul
...
+		cProtocolRecognizer::PROTO_VERSION_1_10_0,
+		cProtocolRecognizer::PROTO_VERSION_1_11_0,
+		cProtocolRecognizer::PROTO_VERSION_1_11_1,
+		cProtocolRecognizer::PROTO_VERSION_1_12,
+	};
+	for (size_t i = 0; i < sizeof(Versions) / sizeof(Versions[0]); ++i)
+	{
+		cForgeMods Mods;
+		m_ForgeModsByVersion.insert({Versions[i], Mods});
+	}
```
madmaxoft: No need to pre-initialize this, just add a check to the registration function so that it creates the map entry if it doesn't exist yet.

src/Server.h
```diff
@@ -71,6 +72,20 @@ class cServer
...
+	/** Add a Forge mod to the server ping list for one protocol version. */
+	bool RegisterForgeModForProtocol(AString & a_ModName, AString & a_ModVersion, UInt32 a_ProtocolVersionNumber);
+	// tolua_begin
+
+	/** Remove a Forge mod from the server ping list. */
+	void UnregisterForgeMod(AString & a_ModName);
+
+	/** Remove a Forge mod to the server ping list for one protocol version. */
+	void UnregisterForgeModForProtocol(AString & a_ModName, UInt32 a_ProtocolVersionNumber);
+
```
madmaxoft: Now you removed the following functions from the API. `// tolua_begin` missing here.

src/Server.h
```diff
@@ -71,6 +72,20 @@ class cServer
 	int  GetNumPlayers(void) const;
 	void SetMaxPlayers(int a_MaxPlayers) { m_MaxPlayers = a_MaxPlayers; }
 
+	// tolua_end
```
madmaxoft: Add an empty line after this so that it stands out more.

src/Server.h
```diff
@@ -148,6 +163,8 @@ class cServer
 	from the settings. */
 	bool ShouldAllowMultiWorldTabCompletion(void) const { return m_ShouldAllowMultiWorldTabCompletion; }
 
+	cForgeMods & GetRegisteredForgeMods(const UInt32 a_Protocol);
```
madmaxoft: All public functions should have a doxy-comment.

src/Server.cpp
madmaxoft: (this is assuming that the all-version functions are dropped)

src/Bindings/ManualBindings.cpp
```diff
@@ -3134,13 +3135,76 @@ static int tolua_cRoot_GetFurnaceRecipe(lua_State * tolua_S)
 
 
 
+static int tolua_cServer_RegisterForgeModForProtocol(lua_State * a_LuaState)
+{
+	cLuaState L(a_LuaState);
+	if (
+		!L.CheckParamUserType(1, "cServer") ||
```
madmaxoft: Use `CheckParamSelf` instead (in all instances in this PR) - #3800

Server/Plugins/APIDump/APIDesc.lua
satoshinm: Sounds reasonable, will remove all-protocol and rename the per-protocol to RegisterForgeMod/UnregisterForgeMod

Server/Plugins/APIDump/APIDesc.lua
satoshinm: Fixed API description in https://github.com/satoshinm/cuberite/pull/2/commits/d7e5e5858d391a0ca4cc2f554c1a4ce8c34124fa

src/Bindings/Plugin.h
```diff
@@ -69,6 +69,7 @@ class cPlugin
 	virtual bool OnKilled                   (cEntity & a_Victim, TakeDamageInfo & a_TDI, AString & a_DeathMessage) = 0;
 	virtual bool OnKilling                  (cEntity & a_Victim, cEntity * a_Killer, TakeDamageInfo & a_TDI) = 0;
 	virtual bool OnLogin                    (cClientHandle & a_Client, UInt32 a_ProtocolVersion, const AString & a_Username) = 0;
+	virtual bool OnLoginForge               (cClientHandle & a_Client) = 0;
```
satoshinm: The mod list can be retrieved from `Client:GetForgeMods()` but I'll see about including it as a parameter

src/ClientHandle.h
```diff
@@ -377,13 +385,18 @@ class cClientHandle  // tolua_export
 
 	bool IsPlayerChunkSent();
 
+	cForgeHandshake m_ForgeHandshake;
+
+	std::unique_ptr<cForgeMods> m_ForgeMods;
+
```
satoshinm: Moved to private, added public accessors for cProtocols. Is there a preference for the ordering of data members? Noticed it seems to be a more logical grouping than alphabetical, and the order has to be matched by the constructor initializer list or compiler error. For now I put m_Forge\* at the top

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,365 @@
...
+/** Discriminator byte values prefixing the FML|HS packets to determine their type. */
+enum
+{
+	Discriminator_ServerHello = 0,
+	Discriminator_ClientHello = 1,
+	Discriminator_ModList = 2,
+	Discriminator_RegistryData = 3,
+	Discriminator_HandshakeReset = -2,
+	Discriminator_HandshakeAck = -1,
+};
```
satoshinm: Made this change, seems good. But for the client/server phases, I was surprised this fails to compile with a duplicate enum definition "COMPLETE":

```c++
/** Client handshake state phases. */
enum ClientPhase: unsigned char
{
	WAITINGSERVERDATA = 2,
	WAITINGSERVERCOMPLETE = 3,
	PENDINGCOMPLETE = 4,
	COMPLETE = 5,
};

/** Server handshake state phases. */
enum ServerPhase: unsigned char
{
	WAITINGCACK = 2,
	COMPLETE = 3,
};
```

satoshinm: Hmm, changing Discriminator to a named enum works in clang but fails with gcc https://travis-ci.org/cuberite/cuberite/jobs/258806610:

/home/travis/build/cuberite/cuberite/src/Protocol/ForgeHandshake.cpp:343:8: error: ‘Discriminator’ is not a class, namespace, or enumeration
   case Discriminator::ClientHello:

lkolbly: What if you used an enum class? Those have better type safety, also (you can't mix the enum with normal numbers). Also values are scoped within the class (fair warning: I haven't actually used them in a project, I've only just now tried them out in an editor).

https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum

satoshinm: `enum class` looks like it should work, separate namespaces though I'll have to also cast the values to use as integers, trying it now

satoshinm: enum class still fails on gcc :( although clang is fine, this causes the Travis-CI build to fail: https://travis-ci.org/cuberite/cuberite/jobs/258935845

/home/travis/build/cuberite/cuberite/src/Protocol/ForgeHandshake.cpp: In member function ‘void cForgeHandshake::DataReceived(cClientHandle*, const char*, size_t)’:
/home/travis/build/cuberite/cuberite/src/Protocol/ForgeHandshake.cpp:343:26: error: ‘Discriminator’ is not a class, namespace, or enumeration
   case static_cast<Int8>(Discriminator::ClientHello):
                          ^
/home/travis/build/cuberite/cuberite/src/Protocol/ForgeHandshake.cpp:347:26: error: ‘Discriminator’ is not a class, namespace, or enumeration
   case static_cast<Int8>(Discriminator::ModList):
                          ^
/home/travis/build/cuberite/cuberite/src/Protocol/ForgeHandshake.cpp:351:26: error: ‘Discriminator’ is not a class, namespace, or enumeration
   case static_cast<Int8>(Discriminator::HandshakeAck):
                          ^
I'll have to revert it unless we can find a compatible solution...

src/Protocol/ForgeHandshake.h
satoshinm: Struggling how to do this, if I change to:

```c++
	const AString m_Name;
	const AString m_UUID;
	const Json::Value m_Properties;
```

then assign in BeginForgeHandshake:

```c++
	m_Name = a_Name;
	m_UUID = a_UUID;
	m_Properties = a_Properties;
```

"No viable overloaded '='"

lkolbly: You can't assign to a const variable (though C++ often surprises me), maybe try making the variables non-constant?

src/Protocol/ForgeMods.h
satoshinm: Cool, I didn't know AStringMap had a binding, I only made cForgeMods because I couldn't find out how to bridge AStringMap to Lua (mod ordering is not important). AStringMap would be simpler and preferable. Started removing cForgeMods, but still having trouble making the Lua API function return AStringMap:

If I have after // tolua_begin this declaration:

```c++
	const AStringMap & GetForgeMods(void) const;
```

then it is bridged as "userdata", I think we want it to be a Lua "table", so plugins can do something like:

```lua
function OnLoginForge(Client)
    local mods = Client:GetForgeMods()

    for name, version in pairs(mods) do
        LOG("SampleForgeMod mod " .. name .. " version " .. version)
    end
```

except this fails: Warning: [e74216953bd757eb|15:30:02] LUA: Plugins/SampleForgeMod/main.lua:60: bad argument #1 to 'pairs' (table expected, got userdata)

CallFunction shouldn't be used in the manual bindings should it?

I now see there is a Lua API function manually bound, returning a table: tolua_cFile_GetFolderContents - but on the C++ side returns an AStringVector. Trying similar code but with AStringMap:

```c++
	cLuaState S(L);
...
	S.Push(Client->GetForgeMods());
```

GetUUIDsFromPlayerNames, GetPermissions, GetRestrictions, GetTeamNames, all return an AStringVector (not AStringMap) as a Lua table. Is there no example of a C++ function returning AStringMap bridged to a Lua table?

src/Protocol/ForgeHandshake.h
satoshinm: Thanks! Silly mistake, that works, will change

src/Protocol/ForgeMods.h
Seadragon91: Not sure if it helps, but the `return 0` at end is wrong, the number should be the amount of pushed values.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -122,17 +122,42 @@ cProtocol_1_9_0::cProtocol_1_9_0(cClientHandle * a_Client, const AString & a_Ser
...
 	{
-		LOGD("Player at %s connected via BungeeCord", Params[1].c_str());
 		m_ServerAddress = Params[0];
-		m_Client->SetIPString(Params[1]);
-		m_Client->SetUUID(cMojangAPI::MakeUUIDShort(Params[2]));
-		m_Client->SetProperties(Params[3]);
+
+		if (Params[1] == "FML")
+		{
+			LOG("Forge client connected!");
```
Seadragon91: `LOGD`?

src/Bindings/ManualBindings.cpp
```diff
@@ -3134,6 +3135,37 @@ static int tolua_cRoot_GetFurnaceRecipe(lua_State * tolua_S)
...
+	UInt32 Protocol;
+	L.GetStackValues(1, Server, Name, Version, Protocol);
+
+	if (!Server->RegisterForgeMod(Name, Version, Protocol))
+	{
+		tolua_error(L, "duplicate Forge mod name registration", nullptr);
+		return 0;
+	}
+
+	return 1;
```
Seadragon91: Nothing returned `return 0?`

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+void cForgeHandshake::BeginForgeHandshake(const AString & a_Name, const AString & a_UUID, const Json::Value & a_Properties)
+{
+	ASSERT(m_IsForgeClient);
+
+	m_Name = a_Name;
+	m_UUID = a_UUID;
+	m_Properties = a_Properties;
+
+	std::array<AString, 5> Channels{{ "FML|HS", "FML", "FML|MP", "FML", "FORGE" }};
```
Seadragon91: If this array is not modified, it could be a `static const`

src/Protocol/ForgeMods.h
satoshinm: Thanks, that was it! Always the simplest things. Tested in https://github.com/satoshinm/SampleForgeMod.lua/commit/bd069a53d5faeb95a060a9c226bf79a7f767324a works now returning a Lua table from the AStringMap, so I completely removed the cForgeMods class

src/Bindings/Plugin.h
satoshinm: Attempting to add `AStringMap & a_Mods` parameter but it is received by the Lua plugin as a boolean? (expected table)

Seadragon91: I have a branch that adds a hook for window click and there I had passed a const cItem and got the same problem with boolean. Looked at the code again and have no idea how I fixed that and the changes are small.

Pushing a AStringMap from ManualBindings works fine and a AStringVector in hook works fine, too. I have no idea whats going on there...

madmaxoft: You need to pass `const AStringMap &`. Unfortunately C++'s param matching inherits the ugly custom from C that almost anything can be converted to a `bool`

src/Protocol/ForgeHandshake.cpp
madmaxoft: Then perhaps a slightly different approach
```cpp
namespace Discriminator
{
  static const Int8 ClientHello = 1;
  static const Int8 ModList = 2;
  ...
}

...

switch (a_Data[0])
{
  case Discriminator::ClientHello: ...
}
```

src/Protocol/ForgeMods.h
madmaxoft: Use `S.CheckParamSelf` for the first param, then `S.GetStackValue(1, Client);` instead of using the reinterpret cast (and no need to check Client for nullptr then, CheckParamSelf already includes that check)

src/Bindings/Plugin.h
Seadragon91: Got it working by removing the &, changing the call:
`return CallSimpleHooks(..., &a_Map, &a_List);`
to
`return CallSimpleHooks(..., a_Map, a_List);`

It doesn't matter if the params are const or not.

satoshinm: Made both changes, works well

src/Protocol/ForgeHandshake.cpp
satoshinm: I like it, will try this. It also has the benefit of detecting unused variables (unlike in enum/enum class).

satoshinm:
Changed to remove cForgeMods class and made other requested changes, ready for review

satoshinm:
Another change: made the `m_ForgeMods` member variable of cClientHandle an `AStringMap` value instead of `std::unique_ptr<AStringMap>`, because GetForgeMods() has to return a value - if there were no mods, it had to create an empty placeholder AStringMap value. By removing the pointer, `m_ForgeMods` can be returned directly, no special casing. This does mean that non-modded clients will have the same `m_ForgeMods` value as a modded client with no mods (Forge itself installed, but no Forge mods), but this difference is already detectable using `m_ForgeHandshake.m_IsForgeClient` (`cClientHandle::IsForgeClient()`).

Seadragon91:
Changes to plugin Core are in here that reverts a few commits. Run `git submodule update` to update the plugin locally and another push will remove it.

satoshinm:
Fixed the Core submodule version change in https://github.com/cuberite/cuberite/pull/3869/commits/a51b6066806687a26085dd3b8563ba71e348ac15, now matches https://github.com/cuberite/cuberite/commit/ec08bf6255759de3389041c3cbb361b970d3ca7b (no longer part of "files changed")

src/Bindings/ManualBindings.cpp
```diff
@@ -2343,6 +2344,34 @@ static int tolua_cClientHandle_SendPluginMessage(lua_State * L)
 
 
 
+static int tolua_cClientHandle_GetForgeMods(lua_State * L)
+{
+	cLuaState S(L);
+	if (
+		!S.CheckParamUserType(1, "cClientHandle") ||
```
Seadragon91: Use `CheckParamSelf` it checks if the class is a instance and is not nil. The nullptr check below can then be removed.

src/Bindings/ManualBindings.cpp
```diff
@@ -2343,6 +2344,34 @@ static int tolua_cClientHandle_SendPluginMessage(lua_State * L)
...
+		!S.CheckParamEnd(2)
+		)
+	{
+		return 0;
+	}
+	cClientHandle * Client = reinterpret_cast<cClientHandle *>(tolua_tousertype(L, 1, nullptr));
```
madmaxoft: Use `S.GetStackValue(1, Client)` instead.

src/Bindings/ManualBindings.cpp
```diff
@@ -3743,6 +3803,7 @@ void cManualBindings::Bind(lua_State * tolua_S)
 			tolua_constant(tolua_S, "MAX_VIEW_DISTANCE", cClientHandle::MAX_VIEW_DISTANCE);
 			tolua_constant(tolua_S, "MIN_VIEW_DISTANCE", cClientHandle::MIN_VIEW_DISTANCE);
 			tolua_function(tolua_S, "SendPluginMessage", tolua_cClientHandle_SendPluginMessage);
+			tolua_function(tolua_S, "GetForgeMods", tolua_cClientHandle_GetForgeMods);
```
madmaxoft: Alpha-sort the (two) functions. Perhaps add an empty line to separate the functions from the constants.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,366 @@
...
+		LOGD("ParseModList failed to read mod count");
+		return Mods;
+	}
+
+	for (size_t i = 0; i < NumMods; ++i)
+	{
+		AString Name, Version;
+		if (!Buf.ReadVarUTF8String(Name))
+		{
+			LOGD("ParseModList failed to read mod name at i=" SIZE_T_FMT, i);
```
madmaxoft: Spaces around `=` (both here and below)

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,366 @@
...
+		LOGD("Received ClientHello with FML protocol version %d", FmlProtocolVersion);
+		if (FmlProtocolVersion != 2)
+		{
+			LOG("Unsupported FML client protocol version received in ClientHello: %d", FmlProtocolVersion);
+			SetError();
+		}
+	}
+	else
+	{
+		LOG("Unexpectedly short ClientHello received");
```
madmaxoft: or long?

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,366 @@
...
+		return;
+	}
+
+	// Send server ModList
+
+	// TODO: max size?
+	cByteBuffer Buf(1024);
+
+	// Send server-side Forge mods registered by plugins
+	auto &ServerMods = m_Client->GetForgeMods();
```
madmaxoft: Space after `&`

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,366 @@
...
+			AString Ack;
+			Ack.push_back(Discriminator::HandshakeAck);
+			Ack.push_back(ServerPhase::COMPLETE);
+			m_Client->SendPluginMessage("FML|HS", Ack);
+
+			// Now finish logging in
+			m_Client->FinishAuthenticate(m_Name, m_UUID, m_Properties);
+			break;
+		}
+	}
```
madmaxoft: Add a `default` case that errors out (protocol error)

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,366 @@
...
+		return;
+	}
+
+	int Discriminator = a_Data[0];
+
+	switch (Discriminator)
+	{
+		case Discriminator::ClientHello:
+			HandleClientHello(a_Client, a_Data, a_Size);
+			break;
```
madmaxoft: Add braces for each `case` code block

madmaxoft: Or, use the single-line format:
```cpp
case Discriminator::ClientHello: HandleClientHello(a_Client, a_Data, a_Size); break;
case Discriminator::ModList:     HandleModList    (a_Client, a_Data, a_Size); break;
...
```

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+		case Discriminator::HandshakeAck:
+			HandleHandshakeAck(a_Client, a_Data, a_Size);
+			break;
+
+		default:
+		{
+			LOG("Unexpected Forge packet %d received", Discriminator);
+			SetError();
```
madmaxoft: This sequence(LOG, SetError) seems to be everywhere. Why not make `SetError` take an error string param that it logs? And change it to LOGD

src/Server.cpp
```diff
@@ -237,6 +237,48 @@ bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_Shoul
 
 
 
+bool cServer::RegisterForgeMod(AString & a_ModName, AString & a_ModVersion, UInt32 a_ProtocolVersionNumber)
+{
+	AStringMap & Mods = GetRegisteredForgeMods(a_ProtocolVersionNumber);
```
madmaxoft: Use more `auto` (almost everywhere)

src/Protocol/ForgeHandshake.cpp
satoshinm: This revealed the ClientPhase::COMPLETE phase was received but not handled; added default and moved FinishAuthenticate in case ClientPhase::COMPLETE

src/Protocol/ForgeHandshake.cpp
satoshinm: Will need to see about how to wrap the variadic arguments of LOGD(...) in SetError

satoshinm: Changed to variadic macro, `LOGD(__VA_ARGS__)` then sets `m_Errored = true`

satoshinm:
👍

src/Protocol/ForgeHandshake.cpp
madmaxoft: Oh, I thought you'd just make it a `void SetError(const AString &)` and use `Printf` for the formatting, if needed.

This way is not really good, because the logging function is called twice, so it emits two messages, with two timestamps.

satoshinm: Will change to `Printf`, didn't know of that function but agreed its much better

src/Protocol/ForgeHandshake.h
```diff
@@ -44,6 +44,8 @@ class cForgeHandshake
 	void HandleModList(cClientHandle * a_Client, const char * a_Data, size_t a_Size);
 	void HandleHandshakeAck(cClientHandle * a_Client, const char * a_Data, size_t a_Size);
 
+	void SetError(const AString & message);
```
madmaxoft: You should probably doxy-comment what this function is intended to do - does it drop the client connection, does it raise an exception, does it phone Mojang to complain... :)

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,371 @@
...
+	}
+}
+
+
+
+
+
+void cForgeHandshake::DataReceived(cClientHandle * a_Client, const char * a_Data, size_t a_Size)
+{
+	LOGD("Received Forge data: " SIZE_T_FMT " bytes: %s", a_Size, a_Data);
```
madmaxoft: Is it wise, dumping potentially binary data to the log? Also, this could crash Cuberite, since there's no guarantee a_Data is NUL-terminated.
If you want extreme logging, use `CreateHexDump` from StringUtils.h. But I think the logging is already needlessly quite chatty, even for a debug version.

src/Protocol/ForgeHandshake.h
```diff
@@ -0,0 +1,63 @@
...
+#pragma once
+
+#include <stddef.h>
+#include "json/json.h"
+
+// fwd:
+namespace Json
+{
+	class Value;
+}
```
madmaxoft: No need for this, you're already including json.h

src/Protocol/ForgeHandshake.h
```diff
@@ -0,0 +1,63 @@
...
+	void BeginForgeHandshake(const AString & a_Name, const AString & a_UUID, const Json::Value & a_Properties);
+
+	/** Send the ServerHello packet in the Forge handshake. */
+	void SendServerHello();
+
+	/** Process received data from the client advancing the Forge handshake. */
+	void DataReceived(cClientHandle * a_Client, const char * a_Data, size_t a_Size);
+
+	/** True if the client advertised itself as a Forge client. */
+	bool m_IsForgeClient;
```
madmaxoft: Move the data members above the functions. Same in the `private` area.

src/Protocol/Protocol.h
```diff
@@ -15,6 +15,7 @@
 #include "../Map.h"
 #include "../ByteBuffer.h"
 #include "../EffectID.h"
+#include "../ClientHandle.h"
```
madmaxoft: Is this include still needed? There's no other change in Protocol.h, so I'd assume it's not.

src/Server.h
```diff
@@ -144,6 +152,9 @@ class cServer
 	from the settings. */
 	bool ShouldAllowMultiWorldTabCompletion(void) const { return m_ShouldAllowMultiWorldTabCompletion; }
 
+	/** Get the Forge mods registered for a given protocol. */
+	AStringMap & GetRegisteredForgeMods(const UInt32 a_Protocol);
```
madmaxoft: Make it return a `const AStringMap &`, so that the caller cannot modify the mod list after this call.

src/Server.h
```diff
@@ -70,6 +70,14 @@ class cServer
 	size_t GetNumPlayers(void) const { return m_PlayerCount; }
 	void SetMaxPlayers(size_t a_MaxPlayers) { m_MaxPlayers = a_MaxPlayers; }
 
+	// tolua_end
+	/** Add a Forge mod to the server ping list. */
+	bool RegisterForgeMod(AString & a_ModName, AString & a_ModVersion, UInt32 a_ProtocolVersionNumber);
+	// tolua_begin
```
madmaxoft: Separate the `tolua` comments with empty lines both above and below the comment, to make it stand out more.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,371 @@
...
+	cByteBuffer Buf(a_Size);
+	Buf.Write(a_Data, a_Size);
+	UInt32 NumMods;
+	if (!Buf.ReadVarInt32(NumMods))
+	{
+		SetError("ParseModList failed to read mod count");
+		return Mods;
+	}
+
+	for (size_t i = 0; i < NumMods; ++i)
```
satoshinm: If I change i to UInt32, clang suggests:

> Values of type 'UInt32' should not be used as format arguments; add an explicit cast to 'unsigned int' instead
> Replace 'SIZE_T_FMT, ' with '%uSIZE_T_FMT, (unsigned int)'

Is this right? I changed to %d with no warning.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,371 @@
...
+	if (cRoot::Get()->GetPluginManager()->CallHookLoginForge(*a_Client, ClientMods))
+	{
+		SetError("Modded client refused by plugin");
+		return;
+	}
+
+	// Send server ModList
+
+	// TODO: max size?
+	cByteBuffer Buf(1024);
```
satoshinm: Good idea scaling by mod count. 1024 per mod may be too high though, this is only for the short mod name and version number string. I'll go with 256

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,371 @@
...
+			UInt32 NumDummies = 0;
+
+			Buf.WriteBool(HasMore);
+			Buf.WriteVarUTF8String(RegistryName);
+			Buf.WriteVarInt32(NumIDs);
+			Buf.WriteVarInt32(NumSubstitutions);
+			Buf.WriteVarInt32(NumDummies);
+
+			AString RegistryData;
+			Buf.ReadAll(RegistryData);
```
satoshinm: HasMode signifies this is the last RegistryData packet. RegistryData packets always have this same structure (bool HasMore followed by the registry name and IDs, which here are empty, will have to be populated once registries are supported but out of scope of this PR for now).

src/Server.h
satoshinm: It's needed to complete the type for accessing m_Client in cProtocol_1_10_0::HandlePacketStatusRequest:

```c++
	m_Client->ForgeAugmentServerListPing(ResponseValue); 
```

but, I could move it to Protocol_1_10.cpp closer to where it is needed

src/Protocol/Protocol.h
satoshinm: It's needed to complete the type for accessing m_Client in cProtocol_1_10_0::HandlePacketStatusRequest:

```c++
	m_Client->ForgeAugmentServerListPing(ResponseValue); 
```

but, I could move it to Protocol_1_10.cpp closer to where it is needed

src/Server.h
satoshinm: But they do need to modify it, RegisterForgeMod() and UnregisterForgeMod() use this function GetRegisteredForgeMods() to get the map per version to modify

lkolbly: If I read it right, RegisterForgeMod and UnregisterForgeMod can directly access m_ForgeModsByVersion, so they can modify that directly and GetRegisteredForgeMods can still return a const, preventing other callers from modifying it.

src/Protocol/ForgeHandshake.cpp
madmaxoft: `%u` should be the right format

src/Server.h
satoshinm: They could but then the body of GetRegisteredForgeMods() would need to be reimplemented. Suppose I could make it private then have a public accessor which is const.

satoshinm: I tried changing to a public const accessor with a private non-const function to get the mods for modification, but my 2¢ it seems less elegant: https://github.com/satoshinm/cuberite/pull/2/commits/aa596db7fa34abb66852d4f3fe0e5b845113f2fc so I reverted it, @madmaxoft @lkolbly let me know what you think

madmaxoft: No, that was the best solution.

madmaxoft:
Note that this PR has a slight "conflict" with #3908 - the new protocol will need the server list ping added to it, too.

src/Server.h
satoshinm: Alright, I re-added it

satoshinm:
So many versions.. not sure how to best handle changes from unmerged branches, but if https://github.com/cuberite/cuberite/pull/3908 (1.12.1) is merged first, then this PR should be updated to include this change: https://github.com/satoshinm/CuberiteForge/commit/5c95c2adb1d267defcc0416daf0dc8b3e8eb1330. I merged the two branches together and tested with this plugin change: https://github.com/satoshinm/SampleForgeMod.lua/commit/e6a1ea8ed949d78047da29c437a2b545711ad054 and it showed up correctly on 1.12.1:

screen shot 2017-08-09 at 8 48 22 pm

satoshinm:
Any other changes needed?

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,367 @@
...
+	static const Int8 WAITINGSERVERDATA = 2;
+	static const Int8 WAITINGSERVERCOMPLETE = 3;
+	static const Int8 PENDINGCOMPLETE = 4;
+	static const Int8 COMPLETE = 5;
+}
+
+/** Server handshake state phases. */
+namespace ServerPhase
+{
+	static const Int8 WAITINGCACK = 2;
```
lkolbly: `WaitingClientAck` if you want to match the Discriminator camelcasing.

src/Protocol/ForgeHandshake.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+void cForgeHandshake::HandleModList(cClientHandle * a_Client, const char * a_Data, size_t a_Size)
+{
+	LOGD("Received ModList");
+
+	auto ClientMods = ParseModList(a_Data + 1, a_Size - 1);
+	AString ClientModsString;
+	for (auto & item: ClientMods)
+	{
+		ClientModsString.append(item.first);
```
lkolbly: Would
```
ClientModsString.append(Printf("%s@%s, ", item.first, item.second))
```
work? It might be slightly slower, but it'd be one line instead of four.

madmaxoft: We even have an `AppendPrintf` function :)

madmaxoft:
Okay, I agree with merge. @lkolbly 's notes would be good to address, but I don't think they're absolutely critical. :)

madmaxoft:
I've merged the 1.12.1 support, so now this can be updated.

src/Protocol/ForgeHandshake.cpp
lkolbly: Ooh, nifty. That's planning ahead.

madmaxoft: Actually, it's just an implementation detail of how all the `Printf`-like functions are implemented - they all go through this one :)

madmaxoft:
The builds are still failing :(

satoshinm:
Merge errors, will investigate

src/Protocol/ForgeHandshake.cpp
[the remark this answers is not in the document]
satoshinm: `WAITINGCACK` matches what Forge calls it, not essential for interoperability but I think it's worthwhile to match their conventions to reduce confusion

src/Protocol/ForgeHandshake.cpp
satoshinm: Slight change to pass C strings, cleaner and works well:

```c++
AppendPrintf(ClientModsString, "%s@%s, ", item.first.c_str(), item.second.c_str());
```

satoshinm:
Fixed, I think this is ready finally...

src/Server.cpp
```diff
@@ -235,6 +235,56 @@ bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_Shoul
...
+	auto & Mods = RegisteredForgeMods(a_ProtocolVersionNumber);
+
+	return Mods.insert({a_ModName, a_ModVersion}).second;
+}
+
+
+
+
+
+void cServer::UnregisterForgeMod(AString &a_ModName, UInt32 a_ProtocolVersionNumber)
```
Seadragon91: Please change to `const`, tolua returns params that AString references. Marking it as const avoid it.

madmaxoft: Wow, I've missed that one (and the style issue). Good thing I'm not the only reviewer :)

Seadragon91: Sorry I cheated, APIFuzzing...

src/Server.h
```diff
@@ -202,6 +215,9 @@ class cServer
 	size_t m_MaxPlayers;
 	bool m_bIsHardcore;
 
+	/** Map of protocol version to Forge mods for that version. */
+	std::map<UInt32, AStringMap> m_ForgeModsByVersion;
```
madmaxoft: Since you're already touching this file due to @Seadragon91 's review, might be worth detailing what this is:
```cpp
/** Map of protocol version to Forge mods (map of ModName -> ModVersionString) for that version. */
```
Perhaps add it to the public getter, too.

src/Server.cpp
madmaxoft: Anything that discovers problems is good, even better if it's automatic :)
Might be reasonable to add APIFuzzing to the CI builds.

Seadragon91: @madmaxoft Created a forum post

satoshinm: Added `const` to the AString references in both RegisterForgeMod and UnregisterForgeMod

PR #3871, opened 2017-07-25
Add cUUID class

### LINKED ISSUE #3870: Represent UUIDs numerically with a dedicated class
Somewhat related to  #2264.  Maybe @worktycho has even tried this before.

Currently UUIDs are stored and passed around as strings which I think has a few issues:
 * There is little to no validation of the UUIDs.
 * Mixed usage of short and dashed UUIDs means something like `DoWithPlayerByUUID` might fail even if the UUID is correct.
 * Slight waste of memory.  With SSO strings, `sizeof(AString)` is likely more than the 16 bytes needed to store a UUID and that's before the actual string is allocated.

Having a strongly typed UUID class which parses and stores the UUID in numeric form could solve all of these issues.  The current lua API could even be kept by manually binding the lua side too and from strings.

peterbell10:
Closes #3870


src/BlockEntities/MobHeadEntity.h
```diff
@@ -65,6 +60,14 @@ class cMobHeadEntity :
 
 	// tolua_end
 
+	/** Sets the player components for the mob heads with player type. */
+	void SetOwner(const cUUID & a_OwnerUUID, const AString & a_OwnerName,
+		const AString & a_OwnerTexture, const AString & a_OwnerTextureSignature
+	);  // Exported in ManualBindings.cpp
```
madmaxoft: When breaking into multiple lines, please break after the first `(`, too.
(So that all param groups start at the same indent level)

src/Server.cpp
```diff
@@ -18,6 +18,7 @@
 #include "Protocol/ProtocolRecognizer.h"
 #include "CommandOutput.h"
 #include "FastRandom.h"
+#include "UUID.h"
```
madmaxoft: I think the include might be unnecessary - the object is only passed through by reference.

madmaxoft:
How about we only mention the `cUUID` variants in the Lua API documentation, and then in the `cUUID`'s documentation, we mention that anywhere a `cUUID` is expected, Cuberite also allows a string representing the UUID.

It will probably need some special handling in [CuberitePluginChecker](https://github.com/cuberite/CuberitePluginChecker) because that tool parses the API docs programmatically, but to me it seems more plugin-author-friendly.

peterbell10:
I'm slightly hesitant about that because it means all future functions taking UUIDs will have to support strings as well.  Not saying that's necessarily bad but it's something to be careful about.

src/Server.cpp
peterbell10: You're right. I was think `GetUniqueID` would be returning a UUID.

peterbell10:
Is [this](https://github.com/cuberite/CuberitePluginChecker/blob/6461d3889688b16b1279eadcc633fade63598f82/ApiLoader.lua#L162-L185) the only place where I would need to add changes?  I'm not really familiar with the plugin checker.

madmaxoft:
I think it will need more changes than that - it actually provides dummy implementation for all API functions, so that implementation will need to know it can operate on a string, too.

But don't worry about that here, we can fix this later on when we have a plugin that actually uses the stuff.

madmaxoft:
Huh, for some reason my review on this was still pending. Sorry for the delay, I didn't notice.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -12343,6 +12342,133 @@ end
 				},
 			},
 		},
+		cUUID =
+		{
+			Desc = [[
+				Class representing a Universally Unique Identifier.
+				Functions taking a cUUID will also accept a string convertible to a UUID by FromString.
```
madmaxoft: This might need some emphasizing. How about:
```Note that all Cuberite's API functions that take a cUUID parameter will also accept a string in its place, as long as that string can be converted to a cUUID (using the {{#FromString_1|cUUID:FromString}} function).```
(I didn't check the FromString link for validity, please make sure it works :)

peterbell10:
I've added your suggestion to the docs (tested the link and it worked) as well as rebased onto master.

bearbin:
@madmaxoft Merge?

Issue #3853, opened 2017-07-14
Replace int-based X, Y, Z params with Vector3 / cCuboid / cBoundingBox


madmaxoft:
Currently the code sometimes uses three `int` coords for the X, Y, Z params, sometimes it uses a `Vector3`. Even worse, when specifying a cuboid, the code sometimes uses the `(MinX, MaxX, MinY, MaxY, MinZ, MaxZ)` ordering and sometimes the `(MinX, MinY, MinZ, MaxX, MaxY, MaxZ)` ordering.

We should replace most - if not all - of these with `Vector3` and `cCuboid` / `cBoundingBox`. If there are overloads for both, remove the `int`-based versions (but keep them in Lua API for compatibility reasons, possibly with a warning).

This should be done incrementally in small steps, not in one big PR, otherwise it risks never being done due to too many merge conflicts.

madmaxoft:
Note: There's a poll in the forum for this: https://forum.cuberite.org/thread-2987.html

peterbell10:
Posting this here because I want it to be visible after #3874 is merged but it's very much in response to that discussion.
I've written a little benchmark to help decide whether a `Vector3i` should be passed by value or by reference.  I can't say I'm 100% confident of these results because benchmarking something as small as a function call is bound to be hard but this is my best attempt anyway.
You can fiddle with the benchmark [here](http://quick-bench.com/wHRLcjKNcS8aOMKSZLOtllU2LRw).  I've tried to benchmark both methods in two ways: the breadth and depth benchmarks.  The idea being that the depth variant uses tail recursion and the vectors are not forced onto the stack unlike the breadth variants.

Here's one of the results from O2 (NB: breadth and depth aren't comparable):
![o2](https://user-images.githubusercontent.com/13238737/28757253-72f94f6a-7576-11e7-9ba1-fbb2c24272c2.png)

If my method is sound then when the values are on the stack, passing by const ref is very marginally faster but when they can be kept in registers, passing by value is more noticeably faster.

worktycho:
Also, if the functions are inlined as they will be in a loop, it quite frankly doesn't make the blindest bit of difference. With the following use functions there is 5-7 prolog instructions as the actual addition is inlined. If we aren't in a loop then code loading with give more than enough latency to hide the loads.
```
ValVector3 SelfAdd(ValVector3 a_Vec, int Depth)
{
    for (int i = 0; i < 256*256; i++) {
        a_Vec = a_Vec + a_Vec;
[...]

PR #3874, opened 2017-07-26
Started changing int parameters to vector parameters. (in cCuboid and simulators)

lkolbly:
So I was thinking about #3853, so I spent a couple minutes pulling on that thread.

Is this a step in the desired direction? If not, that's fine. Also, I don't know if someone else is currently working on a huge fix-it-all PR, which is also fine. Except that that sounds very painful.

Also, this changes the cCuboid assignment operator, because I think it's slightly faster this way? (no need to call the constructor an extra time for the parameter)

I haven't benchmarked it, so I don't know if this is faster. I think it might be, at least for x86, because we aren't constantly pushing ints. (are there specific procedures used for benchmarking?)


src/ChunkMap.cpp
```diff
@@ -740,17 +740,17 @@ bool cChunkMap::DoWithChunkAt(Vector3i a_BlockPos, std::function<bool(cChunk &)>
 
 
 
-void cChunkMap::WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ)
+void cChunkMap::WakeUpSimulators(const Vector3i & a_Block)
 {
 	cCSLock Lock(m_CSChunks);
 	int ChunkX, ChunkZ;
-	cChunkDef::BlockToChunk(a_BlockX, a_BlockZ, ChunkX, ChunkZ);
+	cChunkDef::BlockToChunk(a_Block.x, a_Block.z, ChunkX, ChunkZ);
```
madmaxoft: Perhaps a `Vector3i`-based overload of this would be useful.

src/World.h
```diff
@@ -493,10 +493,10 @@ class cWorld :
 	double GetSpawnZ(void) const { return m_SpawnZ; }
 
 	/** Wakes up the simulators for the specified block */
-	virtual void WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ) override;
+	virtual void WakeUpSimulators(const Vector3i & a_Block) override;
 
 	/** Wakes up the simulators for the specified area of blocks */
-	void WakeUpSimulatorsInArea(int a_MinBlockX, int a_MaxBlockX, int a_MinBlockY, int a_MaxBlockY, int a_MinBlockZ, int a_MaxBlockZ);
+	void WakeUpSimulatorsInArea(const Vector3i & a_Min, const Vector3i & a_Max);
```
madmaxoft: Note that this function is exported to Lua API, so the old signature still needs to be available.
I'd say it'd be best to keep the old signature and introduce a new one using a `cCuboid`.

src/Cuboid.h
```diff
@@ -18,11 +18,10 @@ class cCuboid
 	cCuboid(const cCuboid & a_Cuboid) : p1(a_Cuboid.p1), p2(a_Cuboid.p2) {}
 	cCuboid(const Vector3i & a_p1, const Vector3i & a_p2) : p1(a_p1), p2(a_p2) {}
 	cCuboid(int a_X1, int a_Y1, int a_Z1) : p1(a_X1, a_Y1, a_Z1), p2(a_X1, a_Y1, a_Z1) {}
-	cCuboid(int a_X1, int a_Y1, int a_Z1, int a_X2, int a_Y2, int a_Z2) : p1(a_X1, a_Y1, a_Z1), p2(a_X2, a_Y2, a_Z2) {}
```
madmaxoft: This function is exported to Lua API, so you either need to keep it, or provide a manual binding (possibly emitting a warning, such as in src/Bindings/DeprecatedBindings.cpp)

madmaxoft:
Basically this is exactly the approach I had in mind - work in small increments to gradually change everything within multiple PRs.

You nedd to take care of the Lua API, though. We don't want to break the plugins, so any function that is exported to Lua needs to stay. That doesn't mean it has to stay in the C++ API, it is possible (and we're already doing a lot of it) to provide a manual binding for a function that wraps parameters or calls something else, and logs a deprecation warning to the console - see the src/Bindings/DeprecatedBindings.cpp file. The `cTracer:Trace` binding (`tolua_cTracer_Trace`) is a good example.

madmaxoft:
Perhaps before making these PRs, it would be nice to have a way to automatically detect that auto-exported Lua API symbols have stayed the same. If you look into the APIDump plugin (Server/Plugins/APIDump), the `apicheck` command handler will give you an idea of how to implement such a check.

src/ChunkMap.cpp
lkolbly: I wasn't sure whether to use Vector3i's for things where there's just X/Z coordinates: Is it worth it to use a 3-valued object to only represent two values? Or should I add a Vector2 object?

madmaxoft: Right, sorry, I missed the fact that there's no Y coord.
Still, perhaps consider this signature, as an overload (not replacement) for ease of use?
```cpp
cChunkDef::BlockToChunk(
  const Vector3i & a_BlockCoords,  // The Y coord is ignored
  int & a_OutChunkX,
  int & a_OutChunkZ
);
```

lkolbly:
r.e. Lua deprecation - it's probably easier to just add a LOGWARNING to the C++ function. Then people will still be discouraged from using it within C++, and we don't have to write up a manual binding function in DeprecatedBindings.

src/ChunkMap.cpp
lkolbly: Or what about:
```
cChunkDef::BlockToChunk(
  const Vector3i & a_BlockCoords,  // The Y coord is ignored
  Vector3i & a_OutChunk  // The Y coord is ignored
);
```

madmaxoft:
The problem with that is that it wouldn't show the Lua callstack, which is definitely a nice-to-have with larger plugins.

Basically you can copy the code for DeprecatedBindings from Bindings.cpp (auto-generated) before getting rid of the C++ function autoexport. Although not as pretty as the manually-written version, it would still work.

src/ChunkMap.cpp
madmaxoft: Then it would make sense to make the chunk coords vector the return value.

satoshinm:
Is there performance overhead of changing from int parameters to vector objects?

lkolbly:
I'm a little unclear. If there is a difference, I think it's small - the difference between pushing three ints on a stack vs. pushing a single address on the stack and dereferencing it. The dereference would like hit a fast cache, since the parameter would presumably be recently-accessed. Similarly, pushing the three ints can add up, increasing memory usage for long call chains. (again, though, I think these are both negligible effects, relative to all the other shenanigans happening)

The real reason I want to do this is because I wasted an hour trying to track down a bug that I thought was in my code but was actually in cBoundingBox::Union (see https://github.com/cuberite/cuberite/pull/3850/files#diff-a4a7cf2917583d7d3eed72a53731ace8R131). Switching to vectors everywhere will be cleaner and hopefully avoid this sort of issue going forward.

I don't know exactly why the votes went the way they did, but I have to imagine that's a reason: https://forum.cuberite.org/thread-2987.html
(there's a reference on the forum thread about performance, too, suggesting that Vector3i's are faster on x86)

lkolbly:
@madmaxoft It doesn't look as simple as just copy/pasting the autogenerated binding for the old (int args) method, because there are two totally different overloads, and only one of them is deprecated (so code has to be added to check which parameters are being called, and both would have to be manually exported in DeprecatedBindings).

You're right though that a stacktrace would be nice with larger plugins. Maybe add a method "PrintStackTrace" to cPluginManager, and then call it from the (auto-exported) deprecated function? I don't know if that would get tricky, because at that point you don't know even if the function was called from a plugin, or which plugin made the call.

src/ChunkMap.cpp
lkolbly: Thinking about it more, I feel like using Vector3 in place of a Vector2 is much worse in terms of code readability (consider e.g. cChunkMap::MarkChunkDirty, or any cChunkMap - is there a "Y coord is ignored" comment for every function?). Unless there's an objection I'll add a Vector2 class and use that where appropriate.

Vector2 shouldn't be complicated, since Vector3 is fairly short and Vector2 will have slightly fewer methods.

lkolbly:
At the very least, if the deprecation notice were just a print statement in the old function, then plugin developers should be able to grep for the function name.

src/ChunkMap.cpp
madmaxoft: Perhaps not a `Vector2`, but `ChunkCoords` (doesn't it exist already?) instead, because the coords used are `x` and `z`, instead of the traditional `x` and `y`.

src/Bindings/DiffAPIDesc.lua
```diff
@@ -129,14 +129,17 @@ end
...
-	local numParams
+	local numParams = 0
 	local numOptionalParams = 0
 	if (not(a_FunctionDesc.Params) or (a_FunctionDesc.Params == "")) then
 		numParams = 0
 	else
-		_, numParams = string.gsub(a_FunctionDesc.Params, ",", "")
-		numParams = numParams + 1
-		_, numOptionalParams = string.gsub(a_FunctionDesc.Params, "%b[]", "")
+		for a,b in pairs(a_FunctionDesc.Params) do
```
madmaxoft: The usual Lua convention is to use `k` and `v` for `pairs`-based iteration ("key" and "value"). Also, missing a space after the comma.

src/Blocks/BlockButton.h
```diff
@@ -34,7 +34,7 @@ class cBlockButtonHandler :
 		Meta |= 0x08;
 
 		a_ChunkInterface.SetBlockMeta(a_BlockX, a_BlockY, a_BlockZ, Meta, false);
-		a_WorldInterface.WakeUpSimulators(a_BlockX, a_BlockY, a_BlockZ);
+		a_WorldInterface.WakeUpSimulators(Vector3i(a_BlockX, a_BlockY, a_BlockZ));
```
madmaxoft: I wonder if the compiler would accept the "universal initializer" format here (and in similar places):
```cpp
a_WorldInterface.WakeUpSimulators({a_BlockX, a_BlockY, a_BlockZ});
```

src/ChunkDef.h
```diff
@@ -130,16 +150,26 @@ class cChunkDef
 	/** Converts absolute block coords to chunk coords: */
 	inline static void BlockToChunk(int a_X, int a_Z, int & a_ChunkX, int & a_ChunkZ)
 	{
-		a_ChunkX = a_X / Width;
-		if ((a_X < 0) && (a_X % Width != 0))
+		// This version is deprecated in favor of the vector version
+		// If you're developing new code, use the other version.
+		cChunkCoords ChunkCoords = BlockToChunk(Vector3i(a_X, 0, a_Z));
```
madmaxoft: Use `auto`

src/ChunkDef.h
```diff
@@ -130,16 +150,26 @@ class cChunkDef
...
+		a_ChunkX = ChunkCoords.m_ChunkX;
+		a_ChunkZ = ChunkCoords.m_ChunkZ;
+	}
+
+	/** The Y coordinate is ignored */
```
madmaxoft: `The Y coordinate of a_Pos is ignored.`

src/ChunkMap.cpp
```diff
@@ -740,17 +740,16 @@ bool cChunkMap::DoWithChunkAt(Vector3i a_BlockPos, std::function<bool(cChunk &)>
...
 
-void cChunkMap::WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ)
+void cChunkMap::WakeUpSimulators(const Vector3i & a_Block)
 {
 	cCSLock Lock(m_CSChunks);
-	int ChunkX, ChunkZ;
-	cChunkDef::BlockToChunk(a_BlockX, a_BlockZ, ChunkX, ChunkZ);
-	cChunkPtr Chunk = GetChunkNoGen(ChunkX, ChunkZ);
+	cChunkCoords ChunkCoord = cChunkDef::BlockToChunk(a_Block);
+	cChunkPtr Chunk = GetChunkNoGen(ChunkCoord);
```
madmaxoft: Get rid of `ChunkCoord`, pass the function return value directly.

src/ChunkMap.h
```diff
@@ -109,10 +109,17 @@ class cChunkMap
...
 
 	/** Calls the callback for the chunk at the block position specified, with ChunkMapCS locked; returns false if the chunk doesn't exist, otherwise returns the same value as the callback */
-	bool DoWithChunkAt(Vector3i a_BlockPos, std::function<bool(cChunk &)> a_Callback);
+	bool DoWithChunkAt(const Vector3i & a_BlockPos, std::function<bool(cChunk &)> a_Callback);
 
 	/** Wakes up simulators for the specified block */
-	void WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ);
+	void WakeUpSimulators(const Vector3i & a_Block);
+
+	// DEPRECATED
```
madmaxoft: When saying "deprecated", always specify what should be used instead (it may be obvious now, but it might get lost through time).

src/World.h
```diff
@@ -29,6 +29,7 @@
 #include "Blocks/WorldInterface.h"
 #include "Blocks/BroadcastInterface.h"
 #include "EffectID.h"
+#include "Cuboid.h"
```
madmaxoft: Don't include unless absolutely needed. Use forward declaration instead.

worktycho:
@satoshinm @lkolbly It makes next to no difference/no difference from a performance perspective to use objects in c++ depending on calling convention. Because you talk about references I assume you are referring to the MSVC x64 ABI. The vector3 object should be small enough to be passed by value on that ABI so it should just treat it as 3 ints. Some ABIs might result in the data being put on the stack rather than kept in registers, but it should stay in the L1 cache, so thats a ~4-10 cycle penalty at most, not worth worrying about.

madmaxoft:
@worktycho I assume there's a difference between these two:
```cpp
void FnByInts(int a_X, int a_Y, int a_Z)
{
  for (int i = 0; i < 1000000; ++i)
  {
    x += a_X;  // Uses an int directly, probably already in a register
  }
}

void FnByObj(const Vector3i & a_Coords)
{
  for (int i = 0; i < 1000000; ++i)
  {
    x += a_Coords.x;  // Uses an int through indirection, a_Coords is basically a ptr
  }
}
```
It would, of course, only make a difference in very tight loops.

worktycho:
@madmaxoft Which is why for a vector, you should not be using const reference, gcc 7.1 generates identical optimized code for these 2 functions:
```
int FnByInts(int a_X, int a_Y, int a_Z)
{
  int x = 0;  
  for (int i = 0; i < 1000000; ++i)
  {
    x += a_X;  // Uses an int directly, probably already in a register
  }
  return x;
}

int FnByObj(Vector3i a_Coords)
{
  int x = 0;
  for (int i = 0; i < 1000000; ++i)
  {
    x += a_Coords.x;  // Uses an int through indirection, a_Coords is basically a ptr
  }
  return x;
}
```

src/Blocks/WorldInterface.h
```diff
@@ -73,6 +73,6 @@ class cWorldInterface
 	virtual int GetHeight(int a_BlockX, int a_BlockZ) = 0;
 
 	/** Wakes up the simulators for the specified block */
-	virtual void WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ) = 0;
+	virtual void WakeUpSimulators(const Vector3i & a_Block) = 0;
```
worktycho: Use pass by value `Vector3i a_Block`

src/ChunkDef.h
```diff
@@ -130,16 +150,26 @@ class cChunkDef
...
-		if ((a_X < 0) && (a_X % Width != 0))
+		// This version is deprecated in favor of the vector version
+		// If you're developing new code, use the other version.
+		auto ChunkCoords = BlockToChunk({a_X, 0, a_Z});
+		a_ChunkX = ChunkCoords.m_ChunkX;
+		a_ChunkZ = ChunkCoords.m_ChunkZ;
+	}
+
+	/** The Y coordinate of a_Pos is ignored */
+	inline static cChunkCoords BlockToChunk(const Vector3i & a_Pos)
```
worktycho: Use pass by value `Vector3i a_Pos`

src/ChunkMap.cpp
```diff
@@ -115,19 +115,19 @@ cChunkPtr cChunkMap::GetChunk(int a_ChunkX, int a_ChunkZ)
 
 
 
-cChunkPtr cChunkMap::GetChunkNoGen(int a_ChunkX, int a_ChunkZ)
+cChunkPtr cChunkMap::GetChunkNoGen(const cChunkCoords & a_Chunk)
```
worktycho: Use pass by value `Vector3i a_Chunk`

src/ChunkMap.cpp
```diff
@@ -715,7 +715,7 @@ bool cChunkMap::DoWithChunk(int a_ChunkX, int a_ChunkZ, cChunkCallback & a_Callb
 }
 
 
-bool cChunkMap::DoWithChunkAt(Vector3i a_BlockPos, std::function<bool(cChunk &)> a_Callback)
+bool cChunkMap::DoWithChunkAt(const Vector3i & a_BlockPos, std::function<bool(cChunk &)> a_Callback)
```
worktycho: Keep the pass by value.

src/ChunkMap.cpp
```diff
@@ -740,17 +740,15 @@ bool cChunkMap::DoWithChunkAt(Vector3i a_BlockPos, std::function<bool(cChunk &)>
 
 
 
-void cChunkMap::WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ)
+void cChunkMap::WakeUpSimulators(const Vector3i & a_Block)
```
worktycho: Use pass by value `Vector3i a_Block`

src/ChunkMap.h
```diff
@@ -109,10 +109,17 @@ class cChunkMap
 	bool DoWithChunk(int a_ChunkX, int a_ChunkZ, cChunkCallback & a_Callback);
 
 	/** Calls the callback for the chunk at the block position specified, with ChunkMapCS locked; returns false if the chunk doesn't exist, otherwise returns the same value as the callback */
-	bool DoWithChunkAt(Vector3i a_BlockPos, std::function<bool(cChunk &)> a_Callback);
+	bool DoWithChunkAt(const Vector3i & a_BlockPos, std::function<bool(cChunk &)> a_Callback);
```
worktycho: Use pass by value `Vector3i a_Block`

src/ChunkMap.h
```diff
@@ -109,10 +109,17 @@ class cChunkMap
...
 
 	/** Wakes up simulators for the specified block */
-	void WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ);
+	void WakeUpSimulators(const Vector3i & a_Block);
```
worktycho: Use pass by value `Vector3i a_Block`

src/Cuboid.cpp
```diff
@@ -24,10 +24,10 @@ static bool DoIntervalsIntersect(int a_Min1, int a_Max1, int a_Min2, int a_Max2)
 ////////////////////////////////////////////////////////////////////////////////
 // cCuboid:
 
-cCuboid & cCuboid::operator =(cCuboid a_Other)
+cCuboid & cCuboid::operator =(const cCuboid & a_Other)
```
worktycho: Is this a worthwhile change?

src/Simulator/Simulator.cpp
```diff
@@ -19,20 +19,20 @@
 
 
 
-void cSimulator::WakeUp(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk)
+void cSimulator::WakeUp(const Vector3i & a_Block, cChunk * a_Chunk)
```
worktycho: Use pass by value `Vector3i a_Block`

src/Simulator/Simulator.h
```diff
@@ -39,7 +39,7 @@ class cSimulator
 	}
 
 	/** Called when a block changes */
-	void WakeUp(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk);
+	void WakeUp(const Vector3i & a_Block, cChunk * a_Chunk);
```
worktycho: Use pass by value `Vector3i a_Block`

src/Simulator/Simulator.h
```diff
@@ -54,9 +54,15 @@ class cSimulator
 protected:
 	friend class cChunk;  // Calls AddBlock() in its WakeUpSimulators() function, to speed things up
 
-	/** Called to simulate a new block */
+	/** Called to simulate a new block (DEPRECATED in favor of vector-parametered version) */
 	virtual void AddBlock(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk) = 0;
 
+	/** Called to simulate a new block */
+	virtual void AddBlock(const Vector3i & a_Block, cChunk * a_Chunk)
```
worktycho: Use pass by value `Vector3i a_Block`

src/Simulator/SimulatorManager.cpp
```diff
@@ -58,11 +58,11 @@ void cSimulatorManager::SimulateChunk(std::chrono::milliseconds a_Dt, int a_Chun
 
 
 
-void cSimulatorManager::WakeUp(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk)
+void cSimulatorManager::WakeUp(const Vector3i & a_Block, cChunk * a_Chunk)
```
worktycho: Use pass by value `Vector3i a_Block`

src/Simulator/SimulatorManager.h
```diff
@@ -36,7 +36,7 @@ class cSimulatorManager
 	void SimulateChunk(std::chrono::milliseconds a_DT, int a_ChunkX, int a_ChunkZ, cChunk * a_Chunk);
 
 	/* Called when a single block changes, wakes all simulators up for the block and its face-neighbors. */
-	void WakeUp(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk);
+	void WakeUp(const Vector3i & a_Block, cChunk * a_Chunk);
```
worktycho: Use pass by value `Vector3i a_Block`

src/World.cpp
```diff
@@ -1294,9 +1294,9 @@ void cWorld::UpdateSkyDarkness(void)
 
 
 
-void cWorld::WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ)
+void cWorld::WakeUpSimulators(const Vector3i & a_Block)
```
worktycho: Use pass by value `Vector3i a_Block`

src/World.h
```diff
@@ -493,9 +493,19 @@ class cWorld :
 	double GetSpawnZ(void) const { return m_SpawnZ; }
 
 	/** Wakes up the simulators for the specified block */
-	virtual void WakeUpSimulators(int a_BlockX, int a_BlockY, int a_BlockZ) override;
+	virtual void WakeUpSimulators(const Vector3i & a_Block) override;
```
worktycho: Use pass by value `Vector3i a_Block`

src/Cuboid.cpp
lkolbly: Performance wise, I think (without having measured it) it's the same order of magnitude as passing Vectors as reference vs value.

Readability wise, I think it's much more obvious what's going on with the equal signs than with the swaps.

peterbell10: I can't see why it's even defined in the first place, it's nothing the compiler wouldn't generate anyway.

worktycho: Maybe just delete it then.

lkolbly: There needs to be an explicit copy assignment operator because there's a user-declared copy constructor (and clang will complain), but I didn't see why there had to be a user-declared copy constructor so I deleted that.

CONTRIBUTING.md
```diff
@@ -21,6 +21,12 @@ Here are the conventions:
   - `ThisIsAProperFunction()` `This_is_bad()` `this_is_bad()` `GoodVariableName` `badVariableName`.
  * All member variables start with `m_`, all function parameters start with `a_`, all class names start with `c`.
   - `class cMonster { int m_Health; int DecreaseHealth(int a_Amount); }`
+ * Function parameters that are coordinates should be passed using an appropriate storage container, and not as three separate arguments.
+  - e.g. for a block position, Vector3i. For an entity position, Vector3d. For a chunk coordinate, cChunkCoords.
+  - For a 3-dimensional box of blocks, use cCuboid. For an axis-aligned bounding box, use cBoundingBox.
+ * Parameters smaller than 4 elements (e.g. Vector3, cChunkCoords) should be passed by reference. All other parameters should be passed by const reference, where applicable.
```
madmaxoft: Small things should be passed by value, not by ref.

src/ChunkMap.h
```diff
@@ -486,7 +493,12 @@ class cChunkMap
 	cChunkPtr GetChunk(int a_ChunkX, int a_ChunkZ);
 
 	/** Constructs a chunk and queues the chunk for loading if not valid, returning it; doesn't generate */
-	cChunkPtr GetChunkNoGen(int a_ChunkX, int a_ChunkZ);
+	// Deprecated in favor of the vector version
+	cChunkPtr GetChunkNoGen(int a_ChunkX, int a_ChunkZ)
+	{
+		return GetChunkNoGen(cChunkCoords(a_ChunkX, a_ChunkZ));
+	}
+	cChunkPtr GetChunkNoGen(cChunkCoords a_Chunk);
```
madmaxoft: I think the `cChunkCoords` version should go directly under the doxy comment, and the deprecated version can go without an explicit doxy-comment.

src/ChunkDef.h
```diff
@@ -112,7 +132,7 @@ class cChunkDef
 	/** Converts relative block coordinates into absolute coordinates with a known chunk location */
 	inline static Vector3i RelativeToAbsolute(const Vector3i & a_RelBlockPosition, int a_ChunkX, int a_ChunkZ)
 	{
-		return Vector3i(a_RelBlockPosition.x + a_ChunkX * Width, a_RelBlockPosition.y, a_RelBlockPosition.z + a_ChunkZ * Width);
+		return {a_RelBlockPosition.x + a_ChunkX * Width, a_RelBlockPosition.y, a_RelBlockPosition.z + a_ChunkZ * Width};
```
madmaxoft: If replacing `const Vector3i &` in the param list with `Vector3i` works without further changes to the code, consider doing so now. If it requires other changes, let's leave it into another PR.

lkolbly: I went ahead and changed it in the major files this PR already touches without any hitches (CI's still running, though), I didn't want to expand the scope of the PR to touch every instance in the codebase.

madmaxoft: Good thinking. Looks good :)

madmaxoft: I just hope it didn't break the Lua bindings somehow...

lkolbly: That occurred to me, apicheck isn't complaining though. In both scenarios tolua will grab a pointer to the object, but for the pass-by-value it will dereference the pointer as soon as it gets it (so there's an extra copy there, could be important for tight loops, but plugins have so much overhead anyway). (and in pass-by-reference it's a const pointer, so there's no chance of accidentally changing the data in lua)

src/Simulator/Simulator.h
```diff
@@ -55,6 +55,12 @@ class cSimulator
 	friend class cChunk;  // Calls AddBlock() in its WakeUpSimulators() function, to speed things up
 
 	/** Called to simulate a new block */
+	virtual void AddBlock(Vector3i a_Block, cChunk * a_Chunk)
+	{
+		AddBlock(a_Block.x, a_Block.y, a_Block.z, a_Chunk);
+	}
+
+	/** Called to simulate a new block (DEPRECATED in favor of vector-parametered version) */
 	virtual void AddBlock(int a_BlockX, int a_BlockY, int a_BlockZ, cChunk * a_Chunk) = 0;
```
madmaxoft: So can we get rid of this overload now? It seems everything is ready for the change.

lkolbly: I had to add a few files in Simulators/ to PR, but it's pushed.

madmaxoft:
@worktycho I think the changes you requested are now satisfied, do you approve the PR now?

lkolbly:
Fixed merge conflicts.

madmaxoft:
Updated to see current build status, planning to squash-merge afterwards.

PR #3876, opened 2017-07-29
cParsedNBT: Improved error reporting

### LINKED ISSUE #2756: cParsedNBT malformed file handing.
How should we handle a malformed NBT file? At the moment we have an assert for an invalid tag that is easily reached by a malformed file, but some invalid files only set the invalid flag.

Should we log the fact the file is malformed? If so at which log level?

Should we assert or abort?

peterbell10:
Closes  #2756

This replaces the boolean `IsValid` error reporting with an `eNBTParseError` enum value internally and a `std::error_code` in the interface.

### DIFF
--- a/src/WorldStorage/FastNBT.h
+++ b/src/WorldStorage/FastNBT.h
@@ -19,6 +19,7 @@ It directly outputs a string containing the serialized NBT data.
 
 #pragma once
 
+#include <system_error>
 #include "../Endianness.h"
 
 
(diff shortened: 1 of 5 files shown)


src/WorldStorage/FastNBT.h
```diff
@@ -107,6 +108,72 @@ struct cFastNBTTag
...
+
+
+
+
+class cNBTParseErrorCategory final:
+	public std::error_category
+{
+	cNBTParseErrorCategory() = default;
+public:
+	/** Catagory name */
```
madmaxoft: Typo

src/WorldStorage/FastNBT.h
```diff
@@ -107,6 +108,72 @@ struct cFastNBTTag
...
+		static cNBTParseErrorCategory Category;
+		return Category;
+	}
+};
+
+
+
+
+
+// The following is required to make an error_code constructable from an eNBTParseError
```
madmaxoft: Constructible

PR #3877, opened 2017-07-30
Implemented books

Seadragon91:
Done:
* Class named `cBookContent` exported to lua - Is the name okay? Anyone has a better one?
* Documentation in APIDoc
* NBT: Loading / saving
* Check for channels: `MC|BEdit` and `MC|BSign`
* Player data: Loading / saving
* Hooks: `HOOK_PLAYER_EDITING_BOOK` and `HOOK_PLAYER_EDITED_BOOK`
*  Sending of the changed item
* NBT content of books is now read, if creative inventory is opened

The book can be edited and opened in clients 1.9 and above.


src/Protocol/Protocol_1_9.cpp
```diff
@@ -2870,6 +2881,55 @@ void cProtocol_1_9_0::HandleVanillaPluginMessage(cByteBuffer & a_ByteBuffer, con
...
+			// Plugin denied the editing of the book
+			return;
+		}
+
+		// Book has been edited, inform plugins
+		cRoot::Get()->GetPluginManager()->CallHookPlayerEditedBook(*Player, BookItem.m_BookContent, IsSigned);
+
+		cInventory & inv = Player->GetInventory();
+		inv.SetHotbarSlot(inv.GetEquippedSlotNum(), BookItem);
+		SendWholeInventory(*Player->GetWindow());  // TODO: Use SendSlot
```
Seadragon91: I see that there is a method named `SendSlot` in `cWindow`, but it requires a slot area.

madmaxoft: I think this piece of code should be factored out into the `cClientHandle` or `cPlayer` class.

Then, the `cWindow` class could probably implement a new method, `SendSlot`, that would take a Player and SlotNum parameters, find the correct cSlotArea and call its SendSlot.

Seadragon91: Found this one... `Player.GetInventory().SendEquippedSlot();`

Should the handling of the books moved into cClientHandle / cPlayer?

src/Bindings/ManualBindings.cpp
```diff
@@ -3702,6 +3702,62 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	for (int i = 1; i <= NumPages; i++)
+	{
+		lua_rawgeti(L, 2, i);
+		AString Page;
+		L.GetStackValue(-1, Page);
+		if (!Page.empty())
+		{
+			Pages.push_back(Page);
+		}
+		lua_pop(L, 1);
```
Seadragon91: Is it possible to use`cStackTable::ForEachArrayElement` here?

madmaxoft: Yes, it should be possible, look at the [LuaChunkStay example](https://github.com/cuberite/cuberite/blob/master/src/Bindings/LuaChunkStay.cpp#L28-L42). Just push each value into a vector.

Seadragon91: I tried to use `StackTable`
```
cLuaState::cStackTablePtr Table;
L.GetStackValue(2, Table);
Table->ForEachArrayElement([=](cLuaState & a_LuaState, int a_Index) -> bool
	{
		LOG("Type = %s", a_LuaState.GetTypeText(a_Index));
		return false;
	}
);
```
This prints userdata, table and string. I expected three strings.

madmaxoft: Do you have the code on a branch so that I can give it a try?

Seadragon91: Looking at LuaChunkStay again, I would have to use `lua_rawgeti` then it should work. Another problem I just see is that I cannot add the string value to a vector that is outside of the function, all variables outside of the function are marked const.

Can I keep the current code?

madmaxoft: The `a_Index` parameter to the callback specifies the array index that is being processed. Get the value by using `a_LuaState.GetStackValue(-1, txt)`

Seadragon91:
Fixed book content lost on open creative inventory. Nothing left on my todo.

Server/Plugins/APIDump/Hooks/OnPlayerEditingBook.lua
```diff
@@ -0,0 +1,27 @@
...
+		DefaultFnName = "OnPlayerEditingBook",  -- also used as pagename
+		Desc = [[
+			This hook is called whenever a {{cPlayer|player}} is editing a book.
+			See also the {{OnPlayerEditingBook|HOOK_PLAYER_EDITED_BOOK}} hook for a similar hook, is called when a
+			player has edited a book.
+		]],
+		Params =
+		{
+			{ Name = "Player", Type = "{{cPlayer}}", Notes = "The player that is editing the book" },
+			{ Name = "BookContent", Type = "{{cBookContent}}", Notes = "The class that contains the current info of the book" },
```
madmaxoft: Is this before or after the change? Can we get both, somehow? Ideally so that a plugin can modify the new content.

src/Bindings/ManualBindings.cpp
```diff
@@ -3702,6 +3702,61 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	cLuaState::cStackTablePtr Pages;
+	L.GetStackValues(1, BookContent, Pages);
+	Pages->ForEachArrayElement([=](cLuaState & a_LuaState, int a_Index) -> bool
+		{
+			AString Page;
+			a_LuaState.GetStackValue(-1, Page);
+			BookContent->AddPage(Page);
+			return false;
+		}
+	);
```
madmaxoft: This seems to add pages, instead of setting them. So calling this multiple times will not replace, but instead append. Intended?

src/Bindings/Plugin.h
```diff
@@ -74,6 +74,8 @@ class cPlugin
 	virtual bool OnPlayerBrokenBlock        (cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) = 0;
 	virtual bool OnPlayerDestroyed          (cPlayer & a_Player) = 0;
 	virtual bool OnPlayerEating             (cPlayer & a_Player) = 0;
+	virtual bool OnPlayerEditedBook         (cPlayer & a_Player, cBookContent & a_BookContent, bool a_IsSigned) = 0;
```
madmaxoft: `const` content?

src/Bindings/PluginLua.cpp
```diff
@@ -1053,6 +1071,8 @@ const char * cPluginLua::GetHookFnName(int a_HookType)
 		case cPluginManager::HOOK_PLAYER_BREAKING_BLOCK:        return "OnPlayerBreakingBlock";
 		case cPluginManager::HOOK_PLAYER_BROKEN_BLOCK:          return "OnPlayerBrokenBlock";
 		case cPluginManager::HOOK_PLAYER_EATING:                return "OnPlayerEating";
+		case cPluginManager::HOOK_PLAYER_EDITED_BOOK:           return "OnPlayerEditedBook";
+		case cPluginManager::HOOK_PLAYER_EDITING_BOOK:          return "OnPlayerSigningdBook";
```
madmaxoft: Name mismatch.

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -105,7 +105,7 @@ void cNBTChunkSerializer::AddItem(const cItem & a_Item, int a_Slot, const AStrin
 	// Write the tag compound (for enchantment, firework, custom name and repair cost):
 	if (
 		(!a_Item.m_Enchantments.IsEmpty()) ||
-		((a_Item.m_ItemType == E_ITEM_FIREWORK_ROCKET) || (a_Item.m_ItemType == E_ITEM_FIREWORK_STAR)) ||
+		((a_Item.m_ItemType == E_ITEM_FIREWORK_ROCKET) || (a_Item.m_ItemType == E_ITEM_FIREWORK_STAR) || (a_Item.m_ItemType == E_ITEM_WRITTEN_BOOK)) || (a_Item.m_ItemType == E_ITEM_BOOK_AND_QUILL) ||
```
madmaxoft: Split this into two lines

Server/Plugins/APIDump/Hooks/OnPlayerEditingBook.lua
Seadragon91: The content of the book can be modified.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6932,14 +6932,19 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
 			},
 			Variables =
 			{
+				m_BookContent =
+				{
+					Type = "{{cBookContent|cBookContent}}",
```
Seadragon91: This `Type = "cBookContent"` doesn't work for member variables. have to use `{{cBookContent}}`. Will look into APIDump what there is wrong.

src/Bindings/ManualBindings.cpp
Seadragon91: Changed, thanks.

src/Bindings/Plugin.h
Seadragon91: Changed.

src/Bindings/PluginLua.cpp
Seadragon91: Corrected

src/Bindings/Plugin.h
```diff
@@ -74,6 +74,8 @@ class cPlugin
 	virtual bool OnPlayerBrokenBlock        (cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) = 0;
 	virtual bool OnPlayerDestroyed          (cPlayer & a_Player) = 0;
 	virtual bool OnPlayerEating             (cPlayer & a_Player) = 0;
+	virtual bool OnPlayerEditedBook         (cPlayer & a_Player, const cBookContent & a_BookContent, bool a_IsSigned) = 0;
+	virtual bool OnPlayerEditingBook        (cPlayer & a_Player, const cBookContent & a_BookContent, bool a_IsSigned) = 0;
```
madmaxoft: Not exactly what I had in mind - now it expects the plugin cannot modify the book content.
I was looking for something like:
```cpp
virtual bool OnPlayerEditingBook        (cPlayer & a_Player, const cBookContent & a_OriginalContent, cBookContent & a_NewContentEditable, bool a_IsSigned) = 0;
```

Seadragon91: I reverted the const change for testing. With the current code I can access the content of the book and modify it in the hook.
Is there a specific problem why a new variable should be used for return?

madmaxoft: I'd like the plugin to be able to see the original contents of the book (before the player even started editing it). That's all.

At least I assume that a player can edit a book multiple times, can't they?

Seadragon91: That works with the current code. I tested it and I can get the current content of the book in in the hooks and can modify it. It work's fine. Do you want test it? A example plugin?

madmaxoft: How do you know where to get the contents - from which item?

Seadragon91: The content of the books that the player types in? That is [send](https://github.com/cuberite/cuberite/blob/books/src/Protocol/Protocol_1_9.cpp#L2902) by the client. If a plugin dev wants to read / modify the content of the book he can then use later `<the item>.m_BookContent`

madmaxoft: I assume the book workflow is like this (for example):
User has a book that contains `abc`. They edit it, they type in `def`. As they click "save" or whatever there is in the UI, the client sends the packet to the server requesting the book content change.

Now I, as a plugin author, want to get both the `abc` (the original text) and the `def` (the new text) as a parameter to the "editing" hook; and I want to be able to change the `def` to `ghi` for some reason. The next plugin that gets the hook then receives `abc` as the original text and `ghi` as the new text.

Seadragon91: It's possible to get the current book content over the equipped item. Is that enough for you? If yes I will add a note to both hooks.

madmaxoft: So it's not possible to edit a book not currently equipped?

Anyway, just why not provide the old contents to the hook nevertheless? It's minimal amount of work and it will simplify the lives of plugin devs.

Seadragon91: Added a new param and changed the names. I just saw another problem. If a player signs a book in vanilla the text is saved into a json string.
This is now the same here, if the book is signed. The pages in `OriginalContent` are strings, in `NewContent` they are json strings.

src/Bindings/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ set(BINDING_DEPENDENCIES
 	../Vector3.h
 	../WebAdmin.h
 	../World.h
+	../BookContent.h
```
madmaxoft: Alpha-sort

src/CMakeLists.txt
```diff
@@ -151,6 +152,7 @@ SET (HDRS
 	VoronoiMap.h
 	WebAdmin.h
 	World.h
+	BookContent.h
```
madmaxoft: Alpha-sort both.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -3308,12 +3379,26 @@ void cProtocol_1_9_0::WriteItem(cPacketizer & a_Pkt, const cItem & a_Item)
...
+		!a_Item.m_ItemColor.IsValid() &&
+		(ItemType != E_ITEM_POTION) &&
+		(ItemType != E_ITEM_SPAWN_EGG) &&
+		(ItemType != E_ITEM_WRITTEN_BOOK) && (ItemType != E_ITEM_BOOK_AND_QUILL))
 	{
 		a_Pkt.WriteBEInt8(0);
 		return;
 	}
 
+	if ((ItemType == E_ITEM_BOOK_AND_QUILL) && (a_Item.m_BookContent.GetPages().size() == 0))
```
madmaxoft: Use `.empty` instead of `.size()` to check emptiness.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -7010,6 +7015,113 @@ local Item5 = cItem(E_ITEM_DIAMOND_CHESTPLATE, 1, 0, "thorns=1;unbreaking=3");
 				},
 			},
 		},
+		cBookContent =
+		{
+			Desc = [[
+This class contains the information for a signed or writeable book: The author, title and the pages. A page of the writeable book is a simple string. For a signed book it can be a json string. For the json string use {{cCompositeChat}} to create a page and then the function {{cCompositeChat#CreateJsonString_1|CreateJsonString}} to get the json string.
```
madmaxoft: How about adding an overload that accepts a `cCompositeChat` and performs the conversion internally?

madmaxoft: Or even using `cCompositeChat` exclusively for all setters ? (But getters should still return a string, in case there's some extra json that `cCompositeChat` wouldn't handle)

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -7010,6 +7015,113 @@ local Item5 = cItem(E_ITEM_DIAMOND_CHESTPLATE, 1, 0, "thorns=1;unbreaking=3");
...
+				SetPages =
+				{
+					Params =
+					{
+						{
+							Name = "Pages",
+							Type = "table",
+						},
+					},
+					Notes = "Set the pages of the book",
```
madmaxoft: `Sets all the pages of the book`

How about adding a `SetPage(PageNum, Text)`, and appropriate `GetPage(PageNum)`

Server/Plugins/APIDump/APIDesc.lua
Seadragon91: I will add an overload, because a json string will only work for signed books.

Server/Plugins/APIDump/APIDesc.lua
Seadragon91: Zero-based?

madmaxoft: Not sure. Lua traditionally uses 1-based, but I think we already have a few things exported 0-based.

madmaxoft: Make it 1-based. It means manual bindings, but it's more consistent for the plugin devs. The C++ side should be 0-based, of course.

Server/Plugins/APIDump/APIDesc.lua
Seadragon91: Or should I accept for both book types a json string. Then internally if the book is not signed get the string from key text and add this to the page?

madmaxoft: Ideally accept and store json for both and then in the protocol handler convert whatever is not supported into a supported form. So that if MC decides to support json in the future, we don't have trouble upgrading stuff.

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -104,7 +104,8 @@ void cNBTChunkSerializer::AddItem(const cItem & a_Item, int a_Slot, const AStrin
 	// Write the tag compound (for enchantment, firework, custom name and repair cost):
 	if (
 		(!a_Item.m_Enchantments.IsEmpty()) ||
-		((a_Item.m_ItemType == E_ITEM_FIREWORK_ROCKET) || (a_Item.m_ItemType == E_ITEM_FIREWORK_STAR)) ||
+		((a_Item.m_ItemType == E_ITEM_FIREWORK_ROCKET) || (a_Item.m_ItemType == E_ITEM_FIREWORK_STAR) ||
```
lkolbly: Since this condition is a string of ORs, I think the double opening parentheses can be removed. (grouping the FIREWORK_ROCKET, FIREWORK_STAR, and WRITTEN_BOOK conditionals)

madmaxoft: At the very least the double-opening-parenthesis is mismatched, closing in between the two conditions for books. I'd vote for removing it as well (but keep the firework stuff on one line and the book stuff on another line.

src/Bindings/Plugin.h
Seadragon91: Storing the pages as json is also not good. As a json is made of multiple parts, e.g. color, formatting, extra parts. Getting the text from it is no good. Whats about using cCompositeChat for the pages? The text can be extracted with `ExtractText` and `CreateJsonString` can return a json string.

madmaxoft: Sounds good.

Seadragon91:
Using `cCompositeChat` internal for the pages is also not good. It's still now strings, but the methods accepts a `cCompositeChat` and if the book is signed it stores as json string or if not as simple string.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -7014,6 +7019,174 @@ local Item5 = cItem(E_ITEM_DIAMOND_CHESTPLATE, 1, 0, "thorns=1;unbreaking=3");
...
+							{
+								Name = "Index",
+								Type = "number",
+							},
+							{
+								Name = "Page",
+								Type = "string",
+							},
+						},
+						Notes = "Set's the page at the given index. Note: one-based",
```
madmaxoft: `Sets`

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -7014,6 +7019,174 @@ local Item5 = cItem(E_ITEM_DIAMOND_CHESTPLATE, 1, 0, "thorns=1;unbreaking=3");
...
+							{
+								Name = "Index",
+								Type = "number",
+							},
+							{
+								Name = "Page",
+								Type = "cCompositeChat",
+							},
+						},
+						Notes = "Set's the page at the given index. For signed book saves it as json string, otherwise as simple string. Note: one-based",
```
madmaxoft: `Sets`

Server/Plugins/APIDump/Hooks/OnPlayerEditedBook.lua
```diff
@@ -0,0 +1,27 @@
+return
+{
+	HOOK_PLAYER_EDITED_BOOK =
+	{
+		CalledWhen = "A player has edited a book.",
+		DefaultFnName = "OnPlayerEditedBook",  -- also used as pagename
+		Desc = [[
+			This hook is called whenever a {{cPlayer|player}} has edited a book.
```
madmaxoft: `whenever` -> `after`

Server/Plugins/APIDump/Hooks/OnPlayerEditingBook.lua
```diff
@@ -0,0 +1,28 @@
+return
+{
+	HOOK_PLAYER_EDITING_BOOK =
+	{
+		CalledWhen = "A player is editing a book.",
+		DefaultFnName = "OnPlayerEditingBook",  -- also used as pagename
+		Desc = [[
+			This hook is called whenever a {{cPlayer|player}} is editing a book.
+			See also the {{OnPlayerEditingBook|HOOK_PLAYER_EDITED_BOOK}} hook for a similar hook, is called when a
+			player has edited a book.
```
madmaxoft: Add the note that the plugin may override the player edits by changing the NewContent parameter

Server/Plugins/APIDump/Hooks/OnPlayerEditedBook.lua
```diff
@@ -0,0 +1,27 @@
...
+			See also the {{OnPlayerEditingBook|HOOK_PLAYER_EDITING_BOOK}} hook for a similar hook, is called when a
+			player is editing a book.
+		]],
+		Params =
+		{
+			{ Name = "Player", Type = "cPlayer", Notes = "The player that edited the book" },
+			{ Name = "NewContent", Type = "cBookContent", Notes = "Contains the new content of the book" },
```
madmaxoft: Should have the `OldContent` parameter, too.

src/Bindings/ManualBindings.cpp
```diff
@@ -3967,6 +3967,227 @@ static int tolua_cEntity_GetSpeed(lua_State * tolua_S)
...
+	{
+		cCompositeChat * CompositeChat = nullptr;
+		L.GetStackValue(2, CompositeChat);
+		if (BookContent->IsSigned())
+		{
+			BookContent->AddPage(CompositeChat->CreateJsonString());
+		}
+		else
+		{
+			BookContent->AddPage(CompositeChat->ExtractText());
```
madmaxoft: This will remove all formatting that could have been applied to the `cCompositeChat`. We might want to keep it, using the old-style formatting codes, or is that not possible?

src/Bindings/ManualBindings.cpp
```diff
@@ -3995,6 +4216,14 @@ void cManualBindings::Bind(lua_State * tolua_S)
 		tolua_function(tolua_S, "Base64Decode",          tolua_Base64Decode);
 		tolua_function(tolua_S, "md5",                   tolua_md5_obsolete);  // OBSOLETE, use cCryptoHash.md5() instead
 
+		tolua_beginmodule(tolua_S, "cBookContent");
+			tolua_function(tolua_S, "AddPage", tolua_cBookContent_AddPage);
+			tolua_function(tolua_S, "GetPage", tolua_cBookContent_GetPage);
+			tolua_function(tolua_S, "GetPages", tolua_cBookContent_GetPages);
+			tolua_function(tolua_S, "SetPage", tolua_cBookContent_SetPage);
+			tolua_function(tolua_S, "SetPages", tolua_cBookContent_SetPages);
```
madmaxoft: Align the parameters vertically

src/Bindings/PluginManager.cpp
```diff
@@ -961,14 +999,14 @@ bool cPluginManager::CallHookPlayerJoined(cPlayer & a_Player)
 
 
 
-bool cPluginManager::CallHookPlayerLeftClick(cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, char a_Status)
+bool cPluginManager::CallHookPlayerOpeningWindow(cPlayer & a_Player, cWindow & a_Window)
```
madmaxoft: Why the move? The functions were kinda alpha-sorted before, now it doesn't make sense

src/BookContent.h
```diff
@@ -0,0 +1,84 @@
...
+	/** Removes all pages */
+	void ClearPages(void)
+	{
+		m_Pages.clear();
+	}
+
+	/** Returns a ref to the vector. Used in ManualBindings and for saving the book */
+	const std::vector<AString> & GetPages(void) const { return m_Pages; }
+
+	/** Read the book content from nbt. The boolean a_SaveAsJson is optional. If the book is signed, the text should be in a json string */
```
madmaxoft: Doesn't explain what the SaveAsJson does nor when exactly it should be used

src/BookContent.h
```diff
@@ -0,0 +1,84 @@
...
+	static void ParseFromNBT(int TagTag, cBookContent & a_BookContent, const cParsedNBT & a_NBT, bool a_SaveAsJson = false);
```
madmaxoft: Prefer to return-by-value, rather than pass-by-out-param

src/Bindings/PluginManager.cpp
Seadragon91: I moved the two functions, because I had old hook names. Not sure why the change looks that strange ...

madmaxoft: Just swap the functions back into the order they were in.

src/Bindings/ManualBindings.cpp
Seadragon91: Ah well this is a problem. What do you mean with old style?

madmaxoft: Old-style formatting uses `§` in the text to switch colors and styles. The new formatting uses JSON for the different parts.

Issue #3467, opened 2016-11-27
Item lore modified in creative inventory


madmaxoft:
When using lore containing an empty line (two consecutive line-ends), it gets modified when the item is moved in creative inventory.

To observe, use the Debuggers' plugin's `/setlore` and `getlore` commands, and a 1.10.2 client.

Steps to reproduce:
Switch to creative.
Add an item to player hotbar.
Set lore on the item that includes two back-apostrophes: ```a``b```
Verify that the lore on the `cItem` is correct in Cuberite (`/getlore`)
Open (creative) inventory and move the item from one slot to another.
Check the lore on the moved item. The double-back-apostrophes are no longer there: ```a`b` ```

This is caused by the client ignoring multiple line-ends, handling them as a single one, and our inventory code trusting the client to move the item, so instead of keeping track of the item in hand while being moved, the CreativeInventoryAction packet's Item is being set into the slot (`cSlotAreaInventoryBase::Clicked()`)

Pokechu22:
> This is caused by the client ignoring multiple line-ends, handling them as a single one, and our inventory code trusting the client to move the item, so instead of keeping track of the item in hand while being moved, the CreativeInventoryAction packet's Item is being set into the slot ( cSlotAreaInventoryBase::Clicked() )

The creative inventory does not move items.  It deletes them and recreates them.  Trying to validate the movement of items in creative is a Bad Idea&trade; as you're probably going to run into situations where there wasn't movement (eg, a player trashes one enchanted book and then grabs another book out of the creative inventory - to the server that's indistinguishable from them picking up an enchanted book and placing down one with different meta).

That said: The client supports multiple line ends.

![2016-11-27_15 55 46](https://cloud.githubusercontent.com/assets/8334194/20652997/0a9f665c-b4ba-11e6-8dd4-d19dcb310128.png)

(created by running `/give @p minecraft:feather 1 0 {display:{Lore:["Test","","","foo"]}}`).  Remember, lore isn't one string with `\n` in it - it's a list of strings.  It might be that cuberite is storing it as 1 long string, and then parsing it as iterating through all the lines and adding `\n` to the end of each - that could cause this behavior (I don't know how the client treats multiple `\n`s in text exactly, but I know it does do some kind of trimming similar to Java's [`String.split`](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#split\(java.lang.String\)).)

peterbell10:
Fixed by #3882

PR #3882, opened 2017-08-01
Represent cItem::m_Lore as an AStringVector

### LINKED ISSUE #3475: Represent cItem::m_Lore as an AStringVector
Currently `cItem::m_Lore` is represented as a single string, which makes it impossible for Cuberite to make available the entire character set - one character needs to be set apart as the string separator when serializing / parsing the lore.

We should instead represent the lore as a full `AStringVector`, so that the full character set can be used for the lore. This change should be done as soon as possible, while there are not too many plugins actually using the `m_Lore` member.

peterbell10:
Closes #3475.

I've gone for the second option suggested by @madmaxoft, keeping the old `m_Lore` name reserved as a deprecated binding.  The new version is called `m_LoreTable` though I'd happily accept a better name.

This does break setting lore from the lua constructor though because I can't see a way to override constructor bindings generated with tolua++.

### DIFF
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -347,6 +347,58 @@ static int tolua_set_cBlockInfo_m_PlaceSound(lua_State * tolua_S)
 
 
 
+static int tolua_get_cItem_m_Lore(lua_State * tolua_S)
+{
+	// Maintain legacy m_Lore variable as Lore table split by ` (grave-accent)
+	cLuaState L(tolua_S);
+	if (!L.CheckParamSelf("cItem"))
+	{
+		return 0;
+	}
+
+	cItem * Self = nullptr;
+	L.GetStackValue(1, Self);
+
+	AString LoreString = StringJoin(Self->m_LoreTable, "`");
+
+	L.Push(LoreString);
+
+	LOGWARNING("Warning in read from variable 'cItem.m_Lore': m_Lore is deprecated. Please use m_LoreTable");
+	L.LogStackTrace(0);
+	return 1;
+}
+
+
+
+
+
+static int tolua_set_cItem_m_Lore(lua_State * tolua_S)
+{
+	// Maintain legacy m_Lore variable as Lore table split by ` (grave-accent)
+	cLuaState L(tolua_S);
+	if (
+		!L.CheckParamSelf("cItem") ||
+		!L.CheckParamString(2)
+		)
+	{
+		return 0;
+	}
+
+	cItem * Self = nullptr;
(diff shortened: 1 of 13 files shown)


Seadragon91:
What's about checking the passed param to `m_Lore`? With `L.GetTypeText` you can get the type of the passed param. If it's a string print a warning and push the string into the vector.

peterbell10:
Do you mean for the constructor?

Seadragon91:
I mean the lua variable `m_Lore`

peterbell10:
As @madmaxoft says in the issue:
> The problem with that is when a plugin attempts to read the value, we can't know if it expects a string or a table.

I'm not sure there's a good answer to that.

Seadragon91:
Ah yes sorry :(

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -292,6 +292,58 @@ static int tolua_AllToLua_StringToMobType00(lua_State* tolua_S)
...
+	}
+
+	cItem * Self = nullptr;
+	L.GetStackValue(1, Self);
+
+	AString LoreString = StringJoin(Self->m_LoreTable, "`");
+
+	L.Push(LoreString);
+
+	LOGWARNING("Warning in read from variable 'cItem.m_Lore': m_Lore is deprecated. Please use m_LoreTable");
```
madmaxoft: Too wordy. `cItem.m_Lore is deprecated, use cItem.m_LoreTable instead` is enough.

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -292,6 +292,58 @@ static int tolua_AllToLua_StringToMobType00(lua_State* tolua_S)
...
+		return 0;
+	}
+
+	cItem * Self = nullptr;
+	AString LoreString;
+	L.GetStackValues(1, Self, LoreString);
+
+	Self->m_LoreTable = StringSplit(LoreString, "`");
+
+	LOGWARNING("Warning in write to variable 'cItem.m_Lore': m_Lore is deprecated. Please use m_LoreTable");
```
madmaxoft: Same re-wording.

madmaxoft:
Nice going :)

peterbell10:
Looks like this fixes #3467 as well.  Though while testing that it revealed a problem with the get bindings not being const.  It works now though.

madmaxoft:
I'm afraid the latest PR merges have produced some merge conflicts (sorry), can you fix those please?

peterbell10:
Sure, it's conflicting with my own PR so I would've been dealing with it either way :)

PR #3883, opened 2017-08-02
Shortened DoIntervalsIntersect to two comparisons, made inline.

lkolbly:
r.e. the conversation in #3853, this shortens what's apparently a commonly called function.

The arguments are in fact sorted (min < max), because there are asserts in cCuboid::DoesIntersect.

### DIFF
--- a/src/Cuboid.cpp
+++ b/src/Cuboid.cpp
@@ -81,23 +81,6 @@ int cCuboid::GetVolume(void) const
 
 
 
-bool cCuboid::DoesIntersect(const cCuboid & a_Other) const
-{
-	ASSERT(IsSorted());
-	ASSERT(a_Other.IsSorted());
-
-	// In order for cuboids to intersect, each of their coord intervals need to intersect
-	return (
-		DoIntervalsIntersect(p1.x, p2.x, a_Other.p1.x, a_Other.p2.x) &&
-		DoIntervalsIntersect(p1.y, p2.y, a_Other.p1.y, a_Other.p2.y) &&
-		DoIntervalsIntersect(p1.z, p2.z, a_Other.p1.z, a_Other.p2.z)
-	);
-}
-
-
-
-
-
 bool cCuboid::IsCompletelyInside(const cCuboid & a_Outer) const
 {
 	ASSERT(IsSorted());
--- a/src/Cuboid.cpp
+++ b/src/Cuboid.cpp
@@ -13,7 +13,7 @@ static inline bool DoIntervalsIntersect(int a_Min1, int a_Max1, int a_Min2, int 
 	ASSERT(a_Min1 <= a_Max1);
 	ASSERT(a_Min2 <= a_Max2);
 	return (
-		(a_Min1 <= a_Max1) &&
+		(a_Min1 <= a_Max2) &&
 		(a_Max1 >= a_Min2)
 	);
 }
--- a/src/Cuboid.cpp
+++ b/src/Cuboid.cpp
@@ -8,12 +8,13 @@
 
 
 /** Returns true if the two specified intervals have a non-empty union */
-static bool DoIntervalsIntersect(int a_Min1, int a_Max1, int a_Min2, int a_Max2)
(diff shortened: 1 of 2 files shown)


src/Cuboid.cpp
```diff
@@ -8,12 +8,13 @@
...
-	return (
-		((a_Min1 >= a_Min2) && (a_Min1 <= a_Max2)) ||  // Start of first  interval is within the second interval
-		((a_Max1 >= a_Min2) && (a_Max1 <= a_Max2)) ||  // End   of first  interval is within the second interval
-		((a_Min2 >= a_Min1) && (a_Min2 <= a_Max1))     // Start of second interval is within the first interval
+	ASSERT(a_Min1 <= a_Max1);
+	ASSERT(a_Min2 <= a_Max2);
+	return !(
+		(a_Min1 > a_Max1) ||
+		(a_Max1 < a_Min2)
 	);
```
peterbell10: This is really nitpicky but I feel like it's more natural without the negation :
```return ((a_Min1 <= a_Max2) && (a_Max1 >= a_Min2));```

Your version reads to me like "the intervals intersect if they don't not intersect".

madmaxoft: I think I'd prefer that as well.

lkolbly: I find it easier to convince myself it's correct if they don't not intersect (since I can reason about each comparison separately), but I'm also pretty confidant that DeMorgan was correct, so I'm ambivalent.

Changed.

Seadragon91: Is that correct? Two times Max1?

lkolbly: No. Which means that if it's wrong, building a few worlds with cuberite won't crash horribly (because it was wrong, and it didn't crash).

lkolbly: It also means that I should probably stop making PRs late at night.

worktycho:
The inline is pointless, since that only relates to ODR, not actual inlining. By inlining I meant moving the whole function, combined with DoesIntersect into Cuboid.h.

lkolbly:
Would it be worthwhile to duplicate the code for all three axes? (or, would a #define be reasonable?)

peterbell10:
Just put `DoIntervalsIntersect` into the header with `DoesIntersect`. Definitely don't turn it into a macro though.

lkolbly:
Okay, I moved the definition of DoIntervalsIntersect into the class.

madmaxoft:
@worktycho I'm curious, how does moving the function definition into the header affect things? I thought most compilers nowadays performed link-time code generation anyway, so it shouldn't matter where the function definition is.

D4N:
@madmaxoft gcc does not perform link time optimization unless you add the the -flto flag. When gcc uses link time optimization, it saves the GIMPLE representation of each object file and during linking the optimizer is started once again using the saved GIMPLE representation. It seems to work quite well, but it can screw up timings due to additional inlining and can crash a debugger (it does with gdb for bigger projects). So your mileage may vary.
However, when you put a function in the header file, it can be inlined during the ordinary optimization step, without link time optimization.

madmaxoft:
And people tell us MSVC is the inferior tool, tsss.

Alright, let's do this then. Move the entire `DoesIntersect` to the header, too.

lkolbly:
Does MSVC (or clang, for that matter) do link time optimization? I'm a hard-core Linux user, I don't have much XP with it (because it's an inferior tool).

Anyway, moved it into the header.

D4N:
Clang can do link time optimization, too (https://llvm.org/docs/LinkTimeOptimization.html). Clang uses basically the same method as GCC, only it saves the llvm bytecode instead of gimple. I think msvc can also employ lto, but I am not sure, never used it (long time, semi-hardcore Linux user, too).

madmaxoft:
MSVC has had LTCG for a very long time, apparently since MSVC.NET (2003?)
Ref.: https://msdn.microsoft.com/en-us/library/bb985904.aspx

PR #3897, opened 2017-08-04
cBlockInfo: Remove/deprecate place sound

peterbell10:
The `cBlockInfo::m_PlaceSound` strings are not used internally and are never actually populated.  This just removes it completely.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -202,7 +202,7 @@ return
 							Type = "string",
 						},
 					},
-					Notes = "Returns the name of the sound that is played when placing the block of this type.",
+					Notes = "(<b>DEPRECATED</b>) Not used by cuberite internally and always returns an empty string.",
 				},
 				GetSpreadLightFalloff =
 				{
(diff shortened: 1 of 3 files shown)


peterbell10:
Looks like apicheck doesn't like this.  Is there a way to have intentionally undocumented symbols or would it be best to leave it documented but note that it's deprecated?

lkolbly:
I don't know how to add intentionally undocumented symbols (apicheck downloads the list from a URL), but my strategy so far has been to keep the documentation but mark it deprecated:
https://github.com/cuberite/cuberite/pull/3874/files#diff-c1e36f42041d0718cb487ee0c2526153R3382

That way if a plugin dev sees an instance of the old deprecated function in their code, they can still look up what the parameters meant so they can upgrade it.

madmaxoft:
The list at the URL is updated regularly (every four hours, I believe) from the current sources, by building Cuberite and running the `api` command from the APIDump plugin. There is intentionally no way to add undocumented symbols, because the goal is to have all of them documented in the end. The proper way of doing this is what you proposed - documenting the stuff as OBSOLETE (and adding what should be used instead, whenever appropriate).

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -202,7 +202,7 @@ return
 							Type = "string",
 						},
 					},
-					Notes = "Returns the name of the sound that is played when placing the block of this type.",
+					Notes = "(<b>DEPRECATED</b>)",
```
madmaxoft: If there's no way for the plugins to change what sounds are played, then say so here.

peterbell10: Is it okay now or should it be more explicit?

madmaxoft: Good enough :)

PR #3901, opened 2017-08-05
Add cLuaWindow OnClicked Callback

lkolbly:
Created a callback for cLuaWindow that's called whenever the player clicks the window. The callback is allowed to stop further processing of the click (for example, to block shift-clicking items out of a window).

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -8048,7 +8048,7 @@ This class is used by plugins wishing to display a custom window to the player, 
 							Type = "function",
 						},
 					},
-					Notes = "Sets the function that the window will call when it is about to process a click from a player. See below for the signature of the callback function.",
+					Notes = "Sets the function that the window will call when it is about to process a click from a player. See {{#additionalinfo_1|below}} for the signature of the callback function.",
 				},
 				SetOnClosing =
 				{
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -8039,6 +8039,17 @@ This class is used by plugins wishing to display a custom window to the player, 
 					},
 					Notes = "Returns the cItemGrid object representing the internal storage in this window",
 				},
+				SetOnClicked =
+				{
+					Params =
+					{
+						{
+							Name = "OnClickedCallback",
+							Type = "function",
+						},
+					},
+					Notes = "Sets the function that the window will call when it is about to process a click from a player. See below for the signature of the callback function.",
+				},
 				SetOnClosing =
 				{
 					Params =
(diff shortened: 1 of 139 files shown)


Seadragon91:
The code looks good and small :)
My question is, if it should be possible to detect a click in all different windows? E.g.: anvil, brewing stand, furnace.

lkolbly:
It only works for cLuaWindow, my use case being preventing shift-clicking in a shop opened by a plugin. Doing it by callback seems cleanest for this use case, but since anvils/brewing stands/furnaces/chests don't have associated cLuaWindows it'd be hard to do them as a per-window callback.

So then it might be best to consolidate everything as a hook-based system, or make everything accessible as a cLuaWindow (the latter might be harder).

Or it might make sense to have a hook-based system for built-in windows, and a callback-based system for plugin-created windows.

Seadragon91:
Looking at the code again I think I see see there two problems. Will need to test it myself to be sure.

> Or it might make sense to have a hook-based system for built-in windows, and a callback-based system for plugin-created windows.

This one could be a good idea.

> Doing it by callback seems cleanest for this use case, but since anvils/brewing stands/furnaces/chests don't have associated cLuaWindows it'd be hard to do them as a per-window callback.

In my branch I have added hooks in `cWindow::Clicked` that works, but as I mentioned before there are a few problems with my approach. It also more complex.

Seadragon91:
As for the shop, what's about dragging the item out of the window? :)

Seadragon91:
Did this changes:
https://github.com/Seadragon91/cuberite/commit/ea01d1fea22c92a02feda170751c7c1c947ccd8b

I think it makes sense to add the clicked item to the function params

Also replaced `SendWholeWindow...` with `BroadcastWholeWindow();`. Is required if multiple players are working on the same window.

And at last added send the dragged item to avoid a ghost item on the mouse cursor. Added
`a_Player.GetClientHandle()->SendInventorySlot(-1, -1, a_Player.GetDraggingItem());`

lkolbly:
Agreed, passing the clicked item to the callback makes sense, although some clicks may not be on an item (such as clicks outside the window), but that should be fine.

Another idea is, what if we added a HOOK_PLAYER_OPENS_WINDOW, which notifies plugins whenever the player opens any cWindow, and then moved SetOnClick to cWindow instead of cLuaWindow? Then at least there's a unified way to intercept window clicks.

lkolbly:
Yeah, for the shop I prevented just dragging the items out of the window using a horribly complex reenterant SetOnSlotChanged callback. I'm hoping to simplify that too :-)

lkolbly:
I pulled in ea01d1fea22c92a02feda170751c7c1c947ccd8b and added a HOOK_PLAYER_OPENING_WINDOW hook. I don't fully understand the intricacies of Lua states yet, so I don't want to expand the callback system to built-in windows for fear of introducing subtle bugs stemming from Lua's GC.

madmaxoft:
The original intent was that the `cWindow` objects exported to Lua will be used for handling all the regular windows, using hooks (HOOK_WINDOW_CLICK), and `cLuaWindow` is specifically for windows created by the plugins. So adding the window click hook is the next logical step after this PR.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -8039,6 +8039,17 @@ This class is used by plugins wishing to display a custom window to the player,
...
+				SetOnClicked =
+				{
+					Params =
+					{
+						{
+							Name = "OnClickedCallback",
+							Type = "function",
+						},
+					},
+					Notes = "Sets the function that the window will call when it is about to process a click from a player. See below for the signature of the callback function.",
```
lkolbly: I couldn't find any prior examples for how to make a link to a section, but this worked (tested with APIDump and a browser).

lkolbly:
Okay. @Seadragon91 has a branch that pretty much implements HOOK_WINDOW_CLICK, after this I'll be sure to avoid that part of the code to avoid conflicts :-)

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: That's exactly the right way :)

Issue #3458, opened 2016-11-23
Allow plugins to change what appears above a players head


PureTryOut:
It's quite hard to describe it, so [an image](http://static.planetminecraft.com/files/resource_media/screenshot/1312/Scoreboards_5116889.jpg) might do it better. Look at the player "TheBoo99"; beneath his nickname he has a text. I would like to have the ability to add text like that, and possibly even change the name displayed there entirely (without actually changing the players name). I could for example give the players name a color.

Pokechu22:
OK, talking about the internals:

In that image, it's from the scoreboard.  Both coloration of the name and the text are based off of the scoreboard.

Coloration is based off of the team the player is in; look at [the teams packet](http://wiki.vg/Protocol#Teams) for more info.  In vanilla, you can change this by first creating a new team, and then putting the player on that team:

```
/scoreboard teams add gold_team Gold Team
/scoreboard teams option gold_team color gold
/scoreboard teams join gold_team @p
```

The text below the name is based off of scoreboard objectives; look at the [score objective](http://wiki.vg/Protocol#Scoreboard_Objective) and [display scoreboard](http://wiki.vg/Protocol#Display_Scoreboard) packets.

```
/scoreboard objectives add test dummy Test Objective
/scoreboard objectives setdisplay belowName test
/scoreboard players set @p test 10
```

For this, you can only set an objective name and value.  All players will display the same objective; this cannot be avoided.  Also it'll always display in the format of `<Value> <Objective>`; this also cannot be changed.  I'm unsure as to whether `<Value>` is forced to be a valid integer even though it's sent as a string, but for some reason I think it is.

I _think_ cuberite may actually already have a scoreboard API you can use (though I'm not familiar with it).

----

Now, technically name changes beyond just changing the color can be specified, but not through a scoreboard packet.  You can do it with the [tab list item](http://wiki.vg/Protocol#Player_List_Item) and [spawn player](http://wiki.vg/Protocol#Spawn_Player) packets, but I am (99%) sure that you can't change a player's name beyond what you can do with teams once they've spawned without relogging them.

Issue #2373, opened 2015-07-19
function for the api


endes123321:
Hello, I recently discovered cuberite and I start making plugins, but I have been limited somewhat by the api and I would like to propose these functions and classes for the api:

* a function to send a scoreboard just a player and not all players
* a function for riding a entity
* a function for send titles
* a class to interact directly with the protocol
* a class to interact with an entity ai
* a class to add NBT tags to block and entities

worktycho:
what do you mean by interact with the protocol? MCServer is written to support multiple protocol versions, so low level protocol changes can get complicated.

NiLSPACE:
A interface like [this](https://github.com/cuberite/cuberite/pull/1826#issuecomment-84707564) could be used for this. It would allow developers to send other packets then the broadcastXYZ packets, or send the same type of packets, but with information that can't be send when using the broadcast functions.

worktycho:
I don't think that exposing protocol stuff is bad, just that we need to be careful.

krisk557:
The scoreboard stuff would be really useful.
Currently it's the only thing that prevents me and my team to completely drop craftbukkit and migrate to cuberite. :(

SamJBarney:
The scoreboard per player was something I was going to ask be implemented :P
I personally think it is a very good idea

LO1ZB:
@SamJBarney that feature is used by almost every minigame server I know.

endes123321:
@krisk557 Scoreboard per player and editing NBT tags are very important in my opinion.
@worktycho  can interact with the protocol would be an advantage for Cubelite

SphinxC0re:
I updated your comment so we now have a task list

PR #3910, opened 2017-08-07
Personalized scoreboards

### LINKED ISSUE #3238: Ability to create different scoreboards for each player
Currently you can only have one scoreboard per world.
It would be really useful if we were able to create different scoreboards for each player in the server.
(It is a common approach for example on bukkit/spigot minigame servers.)

lkolbly:
This is a first stab at per-player scoreboards. This is useful in minigames to provide a scoreboard showing various stats, where each player sees a different scoreboard.

The personal scoreboard is accessed through the cClientHandle of the client using GetScoreBoard().

This scoreboard co-exists awkwardly with the world scoreboards. Both are technically active, but confusion can ensue if the server (or the plugins on it) use a combination of the world scoreboards and personal scoreboards, especially if there are naming collisions. The intended usage is either a plugin sticks with the old way of world-wide scoreboards, or entirely uses personalized scoreboards, ignoring the global scoreboards.

When a player logs in (HOOK_LOGIN), plugins can use the SetShouldUseGlobalScoreboard method to denote which they would like to use with the player, and this determines which scoreboard is initially sent to the player. That defaults to true, and so this PR's API changes are 100% backward compatible.

(on a related note, I think there's a bug where when a player changes worlds they don't get sent the updated scoreboard of the world they're changing to, but I haven't found any opened issues on the subject so I'll have to do some more testing)

Also, the personalized scoreboards only support dummy objectives for now.

Unlike the world scoreboard, the personal scoreboard is not saved to disk. I was thinking, if that's a desired feature, it could be saved alongside the player's .json file in players/ (maybe players/UUID_scoreboard.dat).

As far as code goes, this PR creates a pure-virtual cScoreboardAttachee class, which is the interface for an object that a scoreboard can be attached to (cWorld and cClientHandle implement this interface). cScoreboard keeps a pointer to a cScoreboardAttachee object, and calls methods on that instead of directly on a cWorld.

Fixes #3238 and parts of #2373 and #3458.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -1409,6 +1409,16 @@ end
 					},
 					Notes = "Returns the view distance that the player request, not the used view distance.",
 				},
+				GetScoreBoard =
+				{
+					Returns =
+					{
+						{
+							Type = "cScoreboard",
+						},
+					},
+					Notes = "Returns the personalized scoreboard for this client. NOTE: This scoreboard should never be used when the world scoreboard for the world the player is in is also in use (never use both at the same time).",
+				},
 				GetUniqueID =
 				{
 					Returns =
@@ -1694,6 +1704,17 @@ end
 						},
 					},
 					Notes = "Sets the locale that Cuberite keeps on record. Initially the locale is initialized in protocol handshake, this function allows plugins to override the stored value (but only server-side and only until the user disconnects).",
+				},
+				SetShouldUseGlobalScoreBoard =
+				{
+					Params =
+					{
+						{
+							Name = "ShouldUseGlobalScoreBoard",
+							Type = "boolean",
+						},
+					},
+					Notes = "This should be in a HOOK_LOGIN call. If true (default), then when the player joins they will be sent the world scoreboard. If false, then the player will be initialized with the personal scoreboard.",
 				},
 				SetUsername =
 				{
@@ -11435,7 +11456,7 @@ end
 		cScoreboard =
 		{
 			Desc = [[
-				This class manages the objectives and teams of a single world.
+				This class manages objectives and teams of a scoreboard. Every world has a scoreboard, as well as every cClientHandle. Only one of these should be used - a server should either use the per-world scoreboard or the personalized per-client scoreboard, never both at the same time. For the personalized per-client scoreboard, the only functional objective type is the dummy objective.
 			]],
 			Functions =
 			{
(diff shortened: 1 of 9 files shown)


madmaxoft:
To be honest, I don't think the old scoreboard system has ever been used, because it was never properly finished nor even documented. So in my opinion you could actually replace the old system completely.

One thing to note though - do we want to bind the scoreboards to the client handle (i.e. the connection handler)? At first thought I'd think it'd be more natural to bind it to the Player objects, so that in the future non-human (AI) players might get a scoreboard, too.

lkolbly:
Replacing the old system will make this a much simpler change. World-wide scoreboards could still be implemented as a plugin shim, if someone needed that.

Scoreboards (especially personalized ones) are really a way of sending arbitrary boxes of text to a client. An AI player would most likely not be able to understand them. An AI player could still appear in every client's scoreboard, since objectives contain a map of arbitrary strings to integers (so the arbitrary string could be any player's name, or even something like "Blue ✔" to denote that blue team is still around).

One could imagine an AI designed to read a specific scoreboard from some minigame plugin, but at that point the integration would be tight enough that the minigame plugin might as well directly control the AI (or provide some other event structure for the bots).

peterbell10:
> Scoreboards (especially personalized ones) are really a way of sending arbitrary boxes of text to a client.

Not all scoreboard objectives are displayed to the client and moreover objectives relate to entities in general, not specifically to players (maybe not in Cuberite but in minecraft they do).  I would say that scoreboards are primarily a tool to allow command blocks to associate arbitrary data with an entity.  Which may or may not be displayed to a client.

Would it make more sense to keep objective tracking as global but to have per-client control of which objectives are displayed?  This separate API could even specifically support arbitrary text boxes that are kept separate from the real scoreboard objectives.

madmaxoft:
I don't understand the Minecraft scoreboards enough to have any opinion on this, sorry.

lkolbly:
@peterbell10 I like the idea of an API specifically for arbitrary text boxes, perhaps called cHud (or cDisplay? Or cSidebarDisplay? I'm not really sure).

cHud would be maintained per-client handle, and would provide an API like this:
```
void SetSidebarTextLine(int LineNo, string Content);
void SetTabListText(string OtherPlayer, string Content);
void SetPlayerFloatyText(string OtherPlayer, string Content); // The text that floats above a player's head
```
To set arbitrary text.

Also, it'd be able to link to objectives:
```
void SetDisplayObjective(string ObjectiveName, string KeyName, bool shouldDisplay, WhereToDisplay)
```
So that the objective (and maybe specific keys on it) would auto-update.

So cHud would exclusively handle what's displayed on the client, and cScoreboard would exclusively handle internal objectives (and cScoreboard::SetDisplay would be removed/deprecated).

peterbell10:
I don't know enough about the protocol to say how feasible it is but I like the general idea.

Issue #3426, opened 2016-11-07
[Feature] Architecture organization


yangm97:
Ideally, binaries, configuration and user data should not be mixed together. Currently, cuberite, and most of the minecraft servers, use a single folder for everything, without much organization (ie. there's no "worlds" folder, if a server runs many worlds the root directory becomes a big mess; plugins and plugins data are stored together; etc.).
Since this is (thankfully) not a java application but instead a native, first class citizen application, it should be easy to implement it in a more organized way. 

This could be useful for:
- Docker users: currently I have mapped 22 volumes on my container... not ideal nor easy, while I could possibly be mapping 3 or so directories: worlds, settings, plugins; 
- Hosting companies and users looking forward space saving: a single cuberite binary on $PATH could host a thousand servers – this could also allow cuberite to be installed from a package manager; each plugin folder could also be symlinked between servers for even more space optimization;
- Building a plugin package manager should be easy, since plugin "binaries" would not not be mixed with user data;
- Upgrading multiple servers would be easier;
- SELinux compliance (?);
- More use cases probably exist.

I see three ways of doing this: 
1. A major overhaul of cuberite structure, making it more unix-like, storing it's binary at /usr/bin, etc;
2. Keep the one-folder-to-rule-them-all, most people are used to this simplicity after all, but organize it such as bin, plugins, settings, worlds, data, etc;
3. Both: storage method would be defined by compilation flag or binary flag.

There should also be a migration guide/script, for those wanting to switch between these two models.

sphinxc0re:
Thoughts so far? I think this is a beautiful idea @cuberite/core-team

Pokechu22:
I agree with most of this, but...

> Since this is (thankfully) not a java application but instead a native, first class citizen application, it should be easy to implement it in a more organized way.

Language has nothing to do with it.  The notchian server implementation (and bukkit/spigot) support a `--universe` option to change the location of the data; it just defaults to the current directory.  True, things are still really messy within that location, but you can't say that it's java's fault and it mostly matches the way that the client does it.  (Also, on the notchian implementation, but not bukkit/spigot, everything world-related is contained in the `world` folder (all dimensions included); this isn't the case with bukkit due to multiworld support).

Also, you do need to be careful about vanilla support - the actual contents of the worlds directory can't be changed since it needs to be (at least mostly) anvil-compatible.  But you could put them in a subfolder without issue, methinks.

yangm97:
[...]

PR #3912, opened 2017-08-08
Made world data paths adjustable, and added API to temporarily disable saving chunks to disk.

lkolbly:
This PR makes the data path that world data is saved in adjustable at world creation time (via constructor parameter and settings.ini parameter), which will be useful for restructuring the world files (as in #3426). The Anvil format, player stats, and maps are all modified to accommodate this change.

It adds the optional [WorldPaths] section to the settings.ini file, which specifies the directories that various worlds are stored in, for example:
```
[WorldPaths]
world=Worlds/overworld
world_nether=Worlds/nether
world_the_end=Worlds/end
```

This PR also adds the SetSavingEnabled method to cWorld and cRoot, which permits disabling saving chunk files to disk (note that it doesn't prevent writing the ini file to disk). This allows implementing the /save-off and /save-on commands. It will also allow easily resetting maps, I think will be useful for minigames (resetting the map then just becomes an issue of reloading the world).

m_SavingEnabled is atomic, so that cRoot can access it from its thread safely for console commands.

Tweaked the HOOK_CHUNK_UNLOADING documentation to say that the ability for the plugin to actually prevent the chunk from unloading is a feature not a bug.


madmaxoft:
I'm not too happy about having the world config split into two separate sections. But I can't think of a better way.

lkolbly:
Ditto. On further thought, maybe another idea would be:
```
[Server]
...
DefaultWorld=world

[Worlds]
world=Worlds/world
world_nether=Worlds/world_nether
```

Where [Worlds] just specifies the paths, but then the [Server] section gets muddied with the DefaultWorld tag.

lkolbly:
The [WorldPaths] is optional, so it should only appear in the rare instances where the paths are not the default. (the default for <name> would be Worlds/<name> for the architecture change)

madmaxoft:
Migration to that format would be non-trivial.

If only we already had Lua-based config files, we could simply do:
```lua
Worlds =
{
  {
    Name = "world",
    Path = "Worlds/world",
    SomeOtherParam = 3.1415,
  },
  -- ...
}
```

lkolbly:
If I make the config files lua-based (in a future PR), can I make the crafting/brewing/etc files lua-based too?

madmaxoft:
Yes, but please discuss this first in the forum, we have had some discussions there already and there are some ideas that you might want to consider before diving in :)

src/Root.h
```diff
@@ -137,6 +137,9 @@ class cRoot
 	/** Saves all chunks in all worlds */
 	void SaveAllChunks(void);  // tolua_export
 
+	/** Sets whether saving chunks is enabled */
```
madmaxoft: `for all worlds`

src/World.h
```diff
@@ -111,6 +112,12 @@ class cWorld :
 
 	// tolua_begin
 
+	/** Get whether saving chunks is enabled */
+	bool GetSavingEnabled(void) const { return m_SavingEnabled; }
```
madmaxoft: Rename to `IsSavingEnabled` and `m_IsSavingEnabled`

src/World.h
```diff
@@ -871,6 +881,9 @@ class cWorld :
 
 	AString m_WorldName;
 
+	/** The path to the root directory for the world files */
```
madmaxoft: `Doesn't contain the trailing path separator.`

Server/Plugins/APIDump/Hooks/OnChunkUnloading.lua
```diff
@@ -8,8 +8,7 @@ return
 			Cuberite calls this function when a chunk is about to be unloaded from the memory. A plugin may
 			force Cuberite to keep the chunk in memory by returning true.</p>
 			<p>
-			FIXME: The return value should be used only for event propagation stopping, not for the actual
-			decision whether to unload.
+			CAUTION: Preventing the server from unloading chunks can cause the server to use too much RAM, which will adversly affect both performance and stability (i.e. your computer will get slow and crash). Return true sparingly.
```
lkolbly: Today I learned.

lkolbly:
Alright, made some fixes and pulled the latest master.

I'll start working on Lua config files once all the file architecture dust settles. :-)

lkolbly:
Alright, _now_ it passes CI.

bearbin:
Resolved the conflict, just waiting for CI to pass then merge.

PR #3916, opened 2017-08-10
Don't assert when placing beds in older versions

mathiascode:


madmaxoft:
Does this actually work with the older clients? Shouldn't the protocol code just bail out without sending anything?

mathiascode:
When I placed different variants of beds in older clients, the beds were always red, but had the correct color in 1.12.

Seadragon91:
I guess we can merge as the assert will only occur in debug mode.

mathiascode:
What do you want?

madmaxoft:
I think it's okay as-is

Issue #3913, opened 2017-08-10
Sending multiple respawn packets crashes server


mathiascode:
Client version:   1.12.1
Server OS:        Linux
Commit id:        028ed77ca9579eb38a02007e98d75e0dbe9509d2

### Expected behavior
Server doesn't crash if multiple respawn packets are sent while being located in another world than the default.

### Actual behavior
Server crashes. Does not happen if located in the default world. Related to #3649. 

From the looks of it, when the first respawn packet is sent, the server is [removing](https://github.com/cuberite/cuberite/blob/759618b0356a357d06a89f4348437a8de52d2078/src/World.cpp#L3076) the player and [warping](https://github.com/cuberite/cuberite/blob/79eb4c3aed59edce58d2529bf9c6334f7fe6fbad/src/Entities/Player.cpp#L2043) the player to the default world. Before the player has been [added](https://github.com/cuberite/cuberite/blob/759618b0356a357d06a89f4348437a8de52d2078/src/World.cpp#L4010) to the default world, the second respawn packet starts [removing](https://github.com/cuberite/cuberite/blob/759618b0356a357d06a89f4348437a8de52d2078/src/World.cpp#L3076) the player from the default world, causing a crash.

If you delay the second respawn packet with 1 tick, it is ignored, which should happen.

Release:
```
[14:13:26]   D:    | Cuberite has encountered an error and needs to close
[14:13:26] Details | SIGSEGV: Segmentation fault
[14:13:26] Cuberite local build build id: Unknown
[14:13:26] from commit id: approx: 028ed77ca9579eb38a02007e98d75e0dbe9509d2 built at: approx: 2017-08-04 00:01:22 +0200
./Cuberite(_Z15PrintStackTracev+0x15)[0x55596c3c7a15]            
./Cuberite(+0x29ea04)[0x55596c3b5a04]                     
/lib/x86_64-linux-gnu/libc.so.6(+0x33030)[0x7f6f43b61030]
./Cuberite(_ZN6cChunk12RemoveEntityEP7cEntity+0x2a)[0x55596c2f7e5a]
./Cuberite(+0x5001e5)[0x55596c6171e5]
./Cuberite(_ZN6cWorld15TickQueuedTasksEv+0x1e2)[0x55596c3a9cf2]
./Cuberite(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x35d)[0x55596c3aa11d]
[...]

PR #3921, opened 2017-08-11
Removed incorrect assert in SendRespawn

lkolbly:
If a player respawns to a different dimension (sleep in the overworld, die in the nether), they call SendRespawn with a_ShouldIgnoreDimensionChecks=true, which leads to an assertion error because the dimensions are in fact different.

My 1.12 client was fine with respawning across dimensions, both from a plugin-triggered respawn and a death-triggered respawn.

While not directly causing any bugs I know of, this hides the true cause of #3913 (since the debugger crashes on the assert error).


lkolbly:
Edit: Changed the name from the grandiose "Fixed respawning across dimensions" to the more humble "Removed incorrect assert in SendRespawn."

PR #3930, opened 2017-08-18
Award player an achievement when creating a beacon

Bond-009:
https://minecraft.gamepedia.com/Beacon#Achievements

### DIFF
--- a/src/BlockEntities/BeaconEntity.cpp
+++ b/src/BlockEntities/BeaconEntity.cpp
@@ -187,16 +187,42 @@ void cBeaconEntity::UpdateBeacon(void)
 		m_IsActive = (m_BeaconLevel > 0);
 	}
 
-	if (m_BeaconLevel != OldBeaconLevel)
+	if ((m_BeaconLevel != OldBeaconLevel) && (m_BeaconLevel == 4))
 	{
 		// Send window update:
 		if (GetWindow() != nullptr)
 		{
 			GetWindow()->SetProperty(0, m_BeaconLevel);
 		}
-	}
 
-	// TODO: Add achievement
+		class cPlayerCallback :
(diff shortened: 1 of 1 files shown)


bearbin:
Clang generates an error in the build - see the travis failure.

lkolbly:
I haven't tried it to be sure, but shouldn't it only award the achievement if m_BeaconLevel == 4? (or whatever the highest is)

https://minecraft.gamepedia.com/Achievements

williamhatcher:
> Be within a 20×20×14 cuboid centered on a beacon block when it realizes it has become powered

I think you're right!

lkolbly:
Huh, interesting, so there's just the one achievement, "The Beaconator", which is for creating a full pyramid, but the advancements split it into two - "Bring Home the Beacon" which only requires turning on a beacon and "Beaconator" which is the full beacon.

Bond-009:
`The player can achieve the Beaconator achievement by placing a beacon on a level-4 pyramid of specific mineral blocks.`
You are right.
Ref: https://minecraft.gamepedia.com/Beacon#History

src/BlockEntities/BeaconEntity.cpp
```diff
@@ -187,16 +187,42 @@ void cBeaconEntity::UpdateBeacon(void)
...
+				{
+					a_Player->AwardAchievement(eStatistic::achFullBeacon);
+				}
+				return false;
+			}
+
+		private:
+			Vector3d m_Position;
+		} PlayerCallback(Vector3d(m_PosX, m_PosY, m_PosZ));
+		GetWorld()->ForEachPlayer(PlayerCallback);
```
madmaxoft: We really need to invest the time to turn these into lambdas...

PR #3932, opened 2017-08-19
Give arthropods slowness IV when hit with bane of arthropods

Bond-009:


src/Entities/Entity.cpp
```diff
@@ -457,10 +457,15 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
 					case mtSilverfish:
 					{
 						a_TDI.RawDamage += static_cast<int>(ceil(2.5 * BaneOfArthropodsLevel));
-						// TODO: Add slowness effect
+						// The duration of the effect is a random value between 1 and 1.5 seconds at level I,
+						// increasing the max duration by 0.5 seconds each level, up to 3.5s with Bane of Arthropods V.
+						// Ref: https://minecraft.gamepedia.com/Enchanting#Bane_of_Arthropods
+						double MaxDuration = std::min(1 + BaneOfArthropodsLevel * 0.5, 3.5);
```
lkolbly: 1 + 5 * 0.5 = 3.5 anyway, so do we really want the min() statement? Bane of Arthropods should only appear naturally up to level 5, I believe, so it won't affect normal gameplay, but a plugin author may want to give a sword a super high Bane of Arthropods level if they want longer slowness for whatever reason.

src/Entities/Entity.cpp
```diff
@@ -457,10 +457,15 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
...
+						double Duration = GetRandomProvider().RandReal<double>(1, MaxDuration * 20);
```
lkolbly: `RandReal<double>(1, MaxDuration) * 20` (minimum amount of time being 1 second rather than 1 tick)?

src/Entities/Entity.cpp
```diff
@@ -457,10 +457,15 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
...
+						Monster->AddEntityEffect(cEntityEffect::effSlowness, static_cast<int>(ceil(Duration)), 4);
```
peterbell10: If I'm not mistaken this is a duration between 1 _tick_ and `MaxDuration` seconds so that should be something like `RandReal<double>(1, MaxDuration) * 20`.
Although I think it would be better to generate an `int` instead of using `ceil(double)` which I think biases against the minimum value.

src/Entities/Entity.cpp
Bond-009: Oops, I misplaced the `* 20`, my bad.

src/Entities/Entity.cpp
Bond-009: IDK, I was just trying to replicate the Vanilla behavior.

lkolbly: I think you end up with Vanilla behavior either way.

Bond-009: Can't you enchant a tool in Vanilla with a higher level than possible with "normal gameplay"?

lkolbly: In Vanilla BoA is capped at level 5, if you try to anvil a book with Level 5 BoA with a sword with Level 5 BoA the result will still be level 5.

src/Entities/Entity.cpp
peterbell10: Perhaps the word bias wasn't strong enough. This is what I'm talking about:
boa
Out of 1 million random numbers generated with this method, the minimum value was generated _0 times_.  There's no reason to use double and it comes with huge correctness issues.
> Say you have a problem, so you use floating-point.
Now you have 2.000001 problems.

Bond-009: Ok, would this be a better solution?
```c++
int Duration = GetRandomProvider().RandInt(1, static_cast<int>(1 + BaneOfArthropodsLevel * 0.5));
Monster->AddEntityEffect(cEntityEffect::effSlowness, Duration * 20, 4);
```

lkolbly: `static_cast<int>(1 + BaneOfArthropodsLevel * 0.5)` will return 1, 2, or 3, and the random will then only choose an exact number of seconds (that will never give, say, 1.5 seconds).

If you multiply by 20 inside the RandInt it should work fine, though:
```
int Duration = 20 + GetRandomProvider().RandInt(0, static_cast<int>(10 * BaneOfArthropodsLevel));
Monster->AddEntityEffect(cEntityEffect::effSlowness, Duration, 4);
```
(pulled the 20 out in front, so there weren't a bunch of 20s everywhere)

src/Entities/Entity.cpp
```diff
@@ -457,10 +457,14 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
 					case mtSilverfish:
 					{
 						a_TDI.RawDamage += static_cast<int>(ceil(2.5 * BaneOfArthropodsLevel));
-						// TODO: Add slowness effect
+						// The duration of the effect is a random value between 1 and 1.5 seconds at level I,
+						// increasing the max duration by 0.5 seconds each level
+						// Ref: https://minecraft.gamepedia.com/Enchanting#Bane_of_Arthropods
+						int Duration = 20 + GetRandomProvider().RandInt(0, BaneOfArthropodsLevel * 10);  // Duration in ticks
+						Monster->AddEntityEffect(cEntityEffect::effSlowness, Duration, 4);
```
peterbell10: You can actually just say `RandInt(Max)` and 0 is implicitly the minimum but that's just superficial.

PR #3936, opened 2017-08-21
Sample .editorconfig

williamhatcher:
Feel free to add onto this as y'all see fit.
This is only a starter set if you will.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,8 +2,7 @@
 root = true
 
 [*]
-indent_style = true
-end_of_line = lf
+indent_style = tab
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+
+root = true
+
+[*]
+indent_style = true
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+spaces_around_operators = true
+


.editorconfig
```diff
@@ -0,0 +1,11 @@
+
+root = true
+
+[*]
+indent_style = true
+end_of_line = lf
```
madmaxoft: (and indeed I get them right currently, because my git is configured properly)

williamhatcher: Can we force pre-commit hooks?

williamhatcher: Okay. Have we discussed making the same thing for all plugins/lua code included in the build?

bearbin: @madmaxoft Can you override `.editorconfig` to use your own preferred settings on your editor? If you can do that then we can set `end_of_line=lf`. The editorconfig should only be a recommendation, with the final control up to the user

williamhatcher: The docs say to leave values blank in these cases

bearbin: @williamhatcher can you provide a link to the docs?

williamhatcher:
Let's move the discussion out of that commit.
Here's the link to their wiki
https://github.com/editorconfig/editorconfig/wiki
@bearbin

madmaxoft:
Note to MSVC users: Using MSVC17 together with this file causes MSVC to force-add UTF8 BOM to all edited files. You need to update to "Update 15.3" to fix the issue.

williamhatcher:
Sorry, I don't use MSVC. Maybe put that note in the contributing file?

madmaxoft:
Nah, most people do update their stuff. And if they don't and search for "UTF8 BOM" in here, they'll find the comment.

PR #3937, opened 2017-08-21
Changed some int parameters to vector parameters

Bond-009:


src/Entities/ExpBottleEntity.h
```diff
@@ -30,15 +30,17 @@ class cExpBottleEntity :
 
 	cExpBottleEntity(cEntity * a_Creator, double a_X, double a_Y, double a_Z, const Vector3d & a_Speed);
 
+	cExpBottleEntity(cEntity * a_Creator, Vector3d a_Pos, const Vector3d & a_Speed);
```
peterbell10: `a_Speed` could be by value as well.

src/World.h
```diff
@@ -454,7 +454,7 @@ class cWorld :
 
 	/** Spawns a boat at the given coordinates.
 	Returns the UniqueID of the spawned boat, or cEntity::INVALID_ID on failure. */
-	UInt32 SpawnBoat(double a_X, double a_Y, double a_Z, cBoat::eMaterial a_Material);
+	UInt32 SpawnBoat(Vector3d a_Pos, cBoat::eMaterial a_Material);
```
peterbell10: This is exported to lua, you'll either need to keep both as overloads or add a manual binding that accepts both forms.

src/World.h
```diff
@@ -463,7 +463,7 @@ class cWorld :
 	/** Spawns a new primed TNT entity at the specified block coords and specified fuse duration.
 	Initial velocity is given based on the relative coefficient provided.
 	Returns the UniqueID of the created entity, or cEntity::INVALID_ID on failure. */
-	UInt32 SpawnPrimedTNT(double a_X, double a_Y, double a_Z, int a_FuseTimeInSec = 80, double a_InitialVelocityCoeff = 1);
+	UInt32 SpawnPrimedTNT(Vector3d a_Pos, int a_FuseTimeInSec = 80, double a_InitialVelocityCoeff = 1);
```
peterbell10: Here too.

madmaxoft:
Oh, yes, and I skipped the redstone simulator altogether again. That is @tigerw 's domain :)

Issue #898, opened 2014-04-17
Mobs can't suffocate.


pikachu784:
Pretty self explanatory, Mobs just won't suffocate from being inside of a block or drown from water.

tigerw:
Close to fixing the monster's heading going through blocks issue presented in the video. #4776

PR #3942, opened 2017-08-21
Gives all entities the default airlevel on creation

### LINKED ISSUE #2665: Possibly not a bug so much as a workaround, Guardians and Squids burn to death on land.
Guardians in vanilla take no damage out of whiter, while squids take damage, but not while on fire. Just pointing that out.

The chicken bug that is now an every mob bug would also go here, but it's been reported already elsewhere.

Bond-009:
* Guardians don't take damage on land
* Squids suffocate on land

Fixed an issue commented on #898 "Mobs don't appear to have an air level though, that's an issue."
Fixes #2665


LogicParrot:
From a brief test, squids now suffocate in water too. Apparently a problem with the `IsSubmerged()` check always returning false. Maybe the pigs are sometimes/always spawning partially inside a block.

Take all of this with a grain of salt. Tested really quick.

bearbin:
@LogicParrot I diagnosed the issue, I will submit a PR soon.

PR #3943, opened 2017-08-22
Added check in cEntity::TickBurning for whether the entity is plannig to change worlds.

### LINKED ISSUE #3939: Player doesn't stop burning when dying
Client version:   1.12.1
Server OS:        Linux
Commit id:        ad3192d6962536e21233710febca9f128b8e0b80

### Expected behavior
Player stops burning after dying.

### Actual behavior
Player still burns. Started happening in a recent build of Cuberite.

Die in lava, and respawn.

lkolbly:
If a player respawns, ScheduleMoveToWorld is called, but the player is not actually moved until a later tick. However, the TickBurning method is called before the player is actually moved, but after the Respawn method has made the player stop burning. Thus, if the player is in contact with fire or lava, it will start burning immediately before actually moving.

Fixes #3939.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1173,6 +1173,12 @@ void cEntity::ApplyFriction(Vector3d & a_Speed, double a_SlowdownMultiplier, flo
 
 void cEntity::TickBurning(cChunk & a_Chunk)
 {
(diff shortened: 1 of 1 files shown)


LogicParrot:
-Apologies. Just testing the review system-

PR #3948, opened 2017-08-22
Wider use of lambdas

peterbell10:
* Replace `cItemCallback` with ~~`cFunctionRef`~~ `std::function` for lambda compatibility
  * Replace some use of "ad hoc functor classes" with lambdas.

I also took the opportunity to change the signature of the DoWith and ForEach callbacks to take references.


src/FunctionRef.h
```diff
@@ -0,0 +1,62 @@
...
+#undef new
+
+// Declared only so it can be partially specialized
+template <class Signature>
+class cFunctionRef;
+
+/** A light-weight, type-erased reference to a function object.
+Non-owning, be mindful of lifetimes. */
+template <class Ret, class... Args>
+class cFunctionRef<Ret(Args...)>
```
madmaxoft: Why exactly do we need a new object for this? Shouldn't using `std::function` suffice? What is the use-case that forces us to use this?

peterbell10: While a `std::function` would work, it is comparatively heavy-weight. It has to copy the object which might require heap allocations (depending on the implementations small functor optimisation limit).  This on the other hand has (or at least should have) no overhead compared to the inheritance method.

Not as much of a motivation but this will also support non-copyable functors.

peterbell10:
Would it be easier if I split this into multiple PRs?  One adding support for lamdas and separate ones for the use of lambdas.

src/FunctionRef.h
madmaxoft: Can you elaborate with an example? I still don't get it :(

madmaxoft:
I don't think splitting is necessary, it'll just take me some time to absorb the new framework's inner workings; after that it's gonna be fast (I hope).

peterbell10:
I'm just slightly worried about how likely conflicts are as this is such a fundamental functionality.

src/FunctionRef.h
peterbell10: Sorry but which part don't you get?  Do you mean why `std::function` might allocate memory?

madmaxoft:
You won't avoid conflicts by splitting the new-file away from this PR :)

src/FunctionRef.h
madmaxoft: I'd like an example for which you felt `std::function` was not good enough (or unusable) and which the new framework fixes.

peterbell10: It's mainly for performance reasons.  Similar to `std::string` vs `std::string_view`.
I can leave it out and use `std::function` if you'd prefer.

madmaxoft: It seems like a premature optimization to me. Let's go with `std::function` for now and if it proves to be a bottleneck in the future, we can revisit this framework.

peterbell10: Categorising 50 lines of code as a "framework" seems a bit extreme.

madmaxoft: Yeah, but it's dense code, quite difficult to understand what it actually does :) Which is exactly the definition of framework (or library - ever looked into STL? :)

peterbell10: I could add more documentation in `cFunctionRef` if that helps.  The principle is very simple and it isn't doing much really.

It's not much different from the c-style stuff used in the network code.

Seadragon91:
We really need tests... . None of the player commands from Core works: e.g. `/gm` or `/give`. They uses the API function `cWorld:FindAndDoWithPlayer`

peterbell10:
I think I just found the need for `cFunctionRef`...
`cRoot::FindAndDoWithPlayer` uses a mutable callback which the `std::function` takes a copy of so the state of the original callback is never updated.  With `cFunctionRef` the lack of copy would keep this working.

PR #3953, opened 2017-08-23
Scoreboard implementation

lkolbly:
DO NOT MERGE, yet.

Still lacking:
- Protocol support for 1.8, 1.9, and 1.12.1.
- Several objective types (it may make sense to implement these in a different PR, just to keep this PR size down. I'm happy to implement them in this PR though)
-- deathCount, playerKillCount, totalKillCount, and health are all implemented within the scoreboard, but aren't updated by the other sections of the code.

I assume we don't want to merge without protocol support for the protocols, and I don't want to duplicate the code for those protocols until there have been a few eyeballs on the 1.12 protocol support, in case someone wants that code to be different, I don't have to do the changes in a bunch of places.

This implements per-world scoreboards, including handling when players move between worlds. Much of the actual business logic is handled by the client, so all the server has to do is tell the clients what the teams are.

Criteria are now a combination of an eCriteria, which specifies the class, and an integer sub criteria whose meaning changes depending on the criteria class. For example, for EntityKill and EntityKilledBy the sub criteria is the entity type number.


bearbin:
@lkolbly How is this implementation coming along - any chance of it being merge-ready soon?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9132,12 +9132,22 @@ a_Player:OpenWindow(Window);
...
 				},
+				GetPlayers =
+				{
+					Returns =
+					{
+						{
+							Type = "table",
+						},
+					},
+					Notes = "Returns all of the players this objective is tracking",
```
madmaxoft: Players, as in `cPlayer` instances, or playernames, or UUIDs? Array-table, or a map?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9095,11 +9095,11 @@ a_Player:OpenWindow(Window);
 					Params =
 					{
 						{
-							Name = "string",
+							Name = "player",
```
madmaxoft: Name or UUID or `cPlayer`? Use a descriptive param name: `playerName`, `playerUuid`

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1358,6 +1358,79 @@ void cProtocol_1_12::HandlePacketAdvancementTab(cByteBuffer & a_ByteBuffer)
...
+	{
+		Pkt.WriteBEUInt8(3);
+	}
+	else
+	{
+		Pkt.WriteBEUInt8(4);
+	}
+
+	Pkt.WriteVarInt32(static_cast<UInt32>(a_Delta.size()));
+	for (auto name : a_Delta)
```
madmaxoft: `const auto & name`

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1358,6 +1358,79 @@ void cProtocol_1_12::HandlePacketAdvancementTab(cByteBuffer & a_ByteBuffer)
...
+		Pkt.WriteBEUInt8(FriendlyFlags);
+
+		Pkt.WriteString("always");  // Name tag visibility
+		Pkt.WriteString("always");  // Collision rule
+		Pkt.WriteBEInt8(static_cast<Int8>(a_Team.GetColor()));  // Color
+	}
+
+	if (a_Mode == 0)
+	{
+		std::set<AString> TeamPlayers = a_Team.GetMembers();
```
madmaxoft: Use `auto`

src/Protocol/Protocol.h
```diff
@@ -131,6 +131,8 @@ class cProtocol
 	virtual void SendSpawnVehicle               (const cEntity & a_Vehicle, char a_VehicleType, char a_VehicleSubType) = 0;
 	virtual void SendStatistics                 (const cStatManager & a_Manager) = 0;
 	virtual void SendTabCompletionResults       (const AStringVector & a_Results) = 0;
+	virtual void SendTeam                       (const cTeam & a_Team, Byte a_Mode) = 0;
```
madmaxoft: Use the `eProtocolAction` enum here as well, only convert in the last moment before sending.

src/Scoreboard.cpp
```diff
@@ -105,6 +166,20 @@ void cObjective::Reset(void)
 
 
 
+AStringVector cObjective::GetPlayers(void) const
+{
+	AStringVector Players;
+	for (auto player : m_Scores)
```
madmaxoft: `const auto & player`

src/Scoreboard.cpp
```diff
@@ -288,15 +444,17 @@ cScoreboard::cScoreboard(cWorld * a_World) : m_World(a_World)
 
 cObjective * cScoreboard::RegisterObjective(const AString & a_Name, const AString & a_DisplayName, cObjective::eType a_Type)
 {
+	if (m_Objectives.count(a_Name) > 0)
+	{
+		return nullptr;
```
madmaxoft: Shouldn't it return the existing objective instead?

madmaxoft:
The scoreboard sources are very, very unintuitive. Even after reading them, I have no idea how the scoreboard system works or how I could use it in a plugin.

You should add doxy-comments for all the public functions in all the scoreboard.h classes. Consider even renaming those functions to make it easier to understand what the function does (all the "Player" functions - do they work on names, uuids or cPlayer instances?), I don't think any plugin is using these, so breaking the API in this case is acceptable and very much desired.

Overall, make more use of `auto` and when iterating over stuff, `const auto &`.

lkolbly:
@bearbin Functionally speaking, it's basically fully functional except for the missing protocols, though the code still needs much polish. For the reasons @madmaxoft mentions, though, I don't feel like it's particularly great code (like, I understand how it all fits together now, but it's taken a lot of learning to get here).

@madmaxoft I'll add some documentation and some examples. Some of the Player functions (the ones in cTeam) are really Entity functions, and some (the ones in cObjective) are really key/value functions. I'll see if I can't change it around.

src/Scoreboard.cpp
lkolbly: Maybe, I don't think so though, for two reasons:
1) If someone creates a new objective, they probably expect it to not have any scores set (so if they create it and add 1 to someone's score, they would be surprised), and
2) If the objective is already created, it's likely that it's being used somewhere else to track something else, so there could be a conflict between the two uses of the scoreboard (imagine if two plugins both used a scoreboard with the same name, for whatever reason).

lkolbly:
@bearbin What I said earlier is slightly misleading - there are a lot of non-crucial features missing (objective criterias, team collision rules and team visibility rules), which don't necessarily need to be implemented this PR, but the core scoreboard features are functional (sending objectives, teams, swapping out scoreboards when players change worlds).

src/Scoreboard.h
```diff
@@ -52,11 +58,32 @@ class cObjective
 		otStatEntityKilledBy
```
madmaxoft: Since the enum has been renamed, the prefix should be changed, too (the only argument against this would be when it is exported to the API, but we agreed that the API can change safely)

src/Scoreboard.h
```diff
@@ -52,11 +58,32 @@ class cObjective
...
 	};
 
-	// tolua_end
+	struct eType
```
madmaxoft: Since this is a struct now, it shouldn't have the `e` prefix.

src/Scoreboard.h
```diff
@@ -52,11 +58,32 @@ class cObjective
...
+		{
+			return (m_Criteria == a_Other.m_Criteria) && (m_SubCriteria == a_Other.m_SubCriteria);
+		}
+
+		eCriteria m_Criteria;
+		int m_SubCriteria;
+	};
 
 	static eType StringToType(const AString & a_Name);
+	static AString TypeToString(eType a_Type);
```
madmaxoft: Don't export these two to Lua API directly, but rather manually convert them to the two enum values:
```lua
cObjective:StringToType("...") -- returns number, number
cObjective:TypeToString(otDummy, 0)
```

madmaxoft:
After starting the review, I realized that this is probably using the wrong approach to the scoreboard system and teams.

We need to decide on the representation first. Do we store the individual entity scores inside the scoreboard itself, or inside the entity representation? Do we store player teams only in the scoreboard, or as links in the player objects? How do we even represent the individual player scores for each objective? What is the API to query it / update it?

lkolbly:
> Do we store the individual entity scores inside the scoreboard itself, or inside the entity representation?

Entity scores can't be stored inside a cEntity, because there can exist scores for "entities" that don't exist (if I read the question correctly). Also, the score still persists even if the entity disappears, for scores that are linked to an entity.

That said, there is still the issue of how to handle the non-dummy scores. By my reckoning, three are three broad objective criteria: Those that can represent anything (dummy), those that represent something concrete that changes on "notable events" (e.g. deathCount, playerKillCount, etc.) and those that represent fluid statistics that change often (e.g. food, air)

So there is the question of how to retrieve those stats. Right now, the system is that the player calls a scoreboard method to update all deathCount scores for the given player (there could be multiple objectives with that).

> Do we store player teams only in the scoreboard, or as links in the player objects?

If a player leaves a world and comes back, the scoreboard for that world needs to know what team they were on when they left (so a cPlayer instance would have to store a cTeam pointer for every world on the server).

src/Scoreboard.h
lkolbly: `RegisterObjective`'s API-facing signature would have to change to take either two ints or a string (or, plugins would have to manually construct an eType (now called Criteria in my local branch)). Its current use is:

```
Scoreboard:RegisterObjective("list", "List Scores", cObjective:StringToType("dummy"))
```

Same for `cScoreboard::ForEachObjectiveWith`.

madmaxoft: Yes, and it would actually still work - Lua allows functions to return multiple values and when such a function is called as the last "parameter" of another function, all of the return values are passed:
```lua
local function giveValues()
  return "hello", 7, "worlds"
end

-- These two print the same thing:
print("hello", 7, "worlds")
print(giveValues())
````

Comment (lkolbly) [reply src/Scoreboard.h]: Huh, that's pretty useful (as a language feature). Today I learned.

That said, I'm not entirely sure we want to do that here. If in the future we wanted to give RegisterObjective another (optional) parameter, existing plugins would break (since it's no longer the last two parameters).

It may make sense to export cObjective::Criteria (née cObjective::eType) to the lua API (maybe even as cObjectiveCriteria). Then also instead of the static methods for converting to/from strings, those could be a Criteria method and constructor, respectively:

```
local Objective = Scoreboard:RegisterObjective("list", "List Scores", cObjectiveCriteria("dummy"))
LOG("Has criteria ", Objective:GetCriteria():ToString())
```

madmaxoft: I'm hesitant to introduce a new API type just for wrapping two numbers representing some type. Would it be reasonable to pack it into a single number (`enum` even better)? Since there is a list of all possible values, it would make sense to represent it as a single enum, possibly with helper functions to break it apart into components?

lkolbly: Vanilla supports 1,791 different criterion for scoreboards (22 criterion classes + all the subcriteria), explicitly enumerating them would be tricky. But, it could be done as a bitfield (16 bits + 16 bits would work, and fit it into an int). Then lua sees it as a single number. Helper functions could be exposed so lua can query the class, convert to strings, and so forth.

src/Scoreboard.h
lkolbly: Changed to "cr..."

lkolbly:
Oh, forgot to mention in the commit message: added code to increment any crStatEntityKill objectives.

toxamin:
is this ever going to get merged?

lkolbly:
I'm about to be OOO for a week, but when I get back I could try to fix all the merge conflicts that have come up. (unless someone else wants to take over before then)

toxamin:
any updates?

lkolbly:
...no, unfortunately, sorry. If you want, feel free to take over this PR.

I can add it to my backlog if you want the feature but don't have C++ chops, but it's been a while since I looked at the codebase, so it would take me a bit of time to re-familiarize myself with it.

PR #3959, opened 2017-08-25
Changed BroadcastSoundEffect, SendSoundEffect, and CastThunderbolt parameters to vectors

lkolbly:
As it says on the tin.

I tried to avoid conflicts with #3937, but if not I'll update this branch after that one's committed.


src/UI/ChestWindow.h
```diff
@@ -35,7 +35,7 @@ class cChestWindow :
 
 protected:
 	cWorld * m_World;
-	int m_BlockX, m_BlockY, m_BlockZ;  // Position of the chest, for the window-close packet
+	Vector3i m_Block;  // Position of the chest, for the window-close packet
```
madmaxoft: Rename this to avoid confusion about the contents (is it the blocktype, or the blockhandler?) `m_BlockPos` would be nice.

src/UI/EnderChestWindow.h
```diff
@@ -30,7 +30,7 @@ class cEnderChestWindow :
 
 protected:
 	cWorld * m_World;
-	int m_BlockX, m_BlockY, m_BlockZ;  // Position of the enderchest, for the window-close packet
+	Vector3i m_Block;  // Position of the enderchest, for the window-close packet
```
madmaxoft: Again `m_BlockPos`

src/BlockEntities/DropSpenserEntity.cpp
```diff
@@ -75,7 +75,7 @@ void cDropSpenserEntity::DropSpense(cChunk & a_Chunk)
 	if (SlotsCnt == 0)
 	{
 		// Nothing in the dropspenser, play the click sound
-		m_World->BroadcastSoundEffect("block.dispenser.fail", static_cast<double>(m_PosX), static_cast<double>(m_PosY), static_cast<double>(m_PosZ), 1.0f, 1.2f);
+		m_World->BroadcastSoundEffect("block.dispenser.fail", {static_cast<double>(m_PosX), static_cast<double>(m_PosY), static_cast<double>(m_PosZ)}, 1.0f, 1.2f);
```
peterbell10: Would `Vector3d(m_PosX, m_PosY, m_PosZ)`  work instead?

madmaxoft: I think it should, and it's much prettier, yes :)

madmaxoft: Anyway, it's temporary, because `m_Pos` should be rewritten into a vector eventually, and conversions between the vector types are defined.

Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -51,6 +51,34 @@ return
 				Params =
 				{
 					{
+						Name = "Block",
```
madmaxoft: Rename to `BlockPos` to avoid ambiguity

src/Chunk.cpp
```diff
@@ -2712,15 +2712,15 @@ void cChunk::BroadcastUnleashEntity(const cEntity & a_Entity)
 
 
 
-void cChunk::BroadcastBlockAction(int a_BlockX, int a_BlockY, int a_BlockZ, char a_Byte1, char a_Byte2, BLOCKTYPE a_BlockType, const cClientHandle * a_Exclude)
+void cChunk::BroadcastBlockAction(Vector3i a_Block, char a_Byte1, char a_Byte2, BLOCKTYPE a_BlockType, const cClientHandle * a_Exclude)
```
madmaxoft: Again, `a_BlockPos`

src/Chunk.cpp
```diff
@@ -3050,15 +3050,15 @@ void cChunk::BroadcastSpawnEntity(cEntity & a_Entity, const cClientHandle * a_Ex
 
 
 
-void cChunk::BroadcastThunderbolt(int a_BlockX, int a_BlockY, int a_BlockZ, const cClientHandle * a_Exclude)
+void cChunk::BroadcastThunderbolt(Vector3i a_Block, const cClientHandle * a_Exclude)
```
madmaxoft: `a_BlockPos`.
Is thunderbolt block-aligned? Shouldn't it be a `Vector3d` instead? In such a case, `a_Position` would be more adequate.

lkolbly: In Vanilla Minecraft, it is not block aligned, but in Cuberite it appears to be block aligned all the way to the protocol.

PR #3960, opened 2017-08-25
compile.sh update. Fixed -d and -n, intelligent thread choice

LogicParrot:
Probably slightly over-engineered. We could just let people choose threads and default to 1 if no choice is made. But it works well, and is pretty smart.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -403,7 +403,6 @@ if [ $STATE_INTERACTIVE -eq 1 ]; then
 		printf %s "Otherwise choose less. Leave empty to choose $DEFAULT_THREADS: "
 	fi
 
-	# DEF
 	read CHOICE_THREADS
 fi
 
--- a/compile.sh
+++ b/compile.sh
@@ -13,7 +13,6 @@ set -e
 # Constants:
 DEFAULT_BUILDTYPE="Release" # Other options: "Debug"
 DEFAULT_BRANCH="master"     # Other options: None currently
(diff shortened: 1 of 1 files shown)


compile.sh
```diff
@@ -56,10 +55,12 @@ errorArguments ()
 	echo "options:"
 	echo "  -m  The compilation mode. Either \"Release\" or \"Debug\". Defaults to \"$DEFAULT_BUILDTYPE\""
 	echo '  -t  The number of threads to use for compiling'
-	echo '      If unspecified, a "smart guess" is attempted'
+	echo "      If unspecified, at most $MAX_DEFAULT_THREADS threads are used. The special value CORES attempts to set the number of"
+	echo '      threads to the number of computer cores.'
 	echo '  -b  The branch to compile. (Currently unused and pinned to MASTER)'
-	echo '  -n  Prevent interactive mode'
-	echo '  -d  Dry run. Print the chosen settings and exit'
+	echo '  -n yes: Prevent interactive mode. Unnecessary in combination with other arguments.'
```
LogicParrot: The reason for the yes is an apparent limitation in getopts

LogicParrot:
I'll replace this with something much simpler.

LogicParrot:
Simplified.
- Print the user's core count in interactive mode, allowing them to choose that if they wish.
- Textual remark for raspberry Pi and for the choice in general.
- Always default to 1 if unspecified by the user.

PR #3961, opened 2017-08-25
Fix mobs not burning in daylight when on snow

### LINKED ISSUE #3945: Mobs not burning in daylight
Client version:   1.8.9
Server OS:        Linux
Commit id:        approx: 9936d07d9b9f24721576ca01d8c846145894e43e

### Expected behavior
The mobs should burn.

Look for mobs at daytime.

bearbin:
or other non-transparent partial blocks.

Fixes #3945

Why was 1 subtracted from the mob's head height? Removing it and changing the loop condition should fix the issue I believe. I don't think we need to consider the block inside the mob's head for burning purposes.


madmaxoft:
Won't this break other things? Such as half-slab roofs that should cover the mobs when they're underneath it, but the new code makes them burn?

LogicParrot:
Consider a two block zombie. Spanning y=5,6

`mobY + height - 1 = 5 + 2 - 1 = 6` which is exactly where the head is unless I made some oversight?

bearbin:
@madmaxoft 

Half slabs are transparent so this code change would have no effect on that, but the core problem is solvable. It would be possible to fix that, but then either short mobs (<1 block high) would be handled incorrectly or a list of partial, non-transparent blocks that still allow burning when inside the mob's head would have to be kept.

@LogicParrot The problem there is that mobs are <2 blocks tall. A 1.8 block tall zombie would be:

floor(5+1.8) = 6
6-1 = 5

LogicParrot:
Good point. The proper generic solution would be to  `ceil(5+1.8) - 1`

bearbin:
`ceil(x) = 1 === floor(x)`

LogicParrot:
Counterexample:

````
ceil(5) - 1 = 4
Floor(5) = 5
```

src/Mobs/Monster.cpp
```diff
@@ -1363,17 +1363,17 @@ bool cMonster::WouldBurnAt(Vector3d a_Location, cChunk & a_Chunk)
...
 			return true;
 		}
-		// Start with the highest block and scan down to the mob's head.
+		// Start with the highest block and scan down to just above the mob's head.
 		// If a non transparent is found, return false (do not burn). Otherwise return true.
 		// Note that this loop is not a performance concern as transparent blocks are rare and the loop almost always bailes out
 		// instantly.(An exception is e.g. standing under a long column of glass).
 		int CurrentBlock = Chunk->GetHeight(Rel.x, Rel.z);
-		while (CurrentBlock >= MobHeight)
+		while (CurrentBlock > MobHeight)
```
LogicParrot: Why not scan mob's head area?  e.g. a 1.5 high mob under a half slab (when the physics supports that sometime in the future)

LogicParrot: ...Or some future half non transparent block

LogicParrot:
To elaborate on my counterexample, your code fails for a perfectly 2 block high mob. If it's standing on `y=5`, then:

Expected head pos: 6
Your code yields: `floor(5 + 2) = 7`
ceiling yields: `ceil(5 + 2) - 1 = 6`

madmaxoft:
Not really - if a mob is standing on a partial block (snow, trapdoor, halfslab) their head may be offset in a way that will make the code skip one block, which could be just the crucial one.

bearbin:
So we would have to keep a list of all non-transparent partial blocks that need to be considered as transparent for the purpose of mob burning, @madmaxoft ?

madmaxoft:
No, avoid blocktype lists as much as we can.

The more I look at this code, the more I think it is actually a no-change: the `MobHeight` is one higher, and the `while` loop checks one less block. Does this even help the original issue? Have you tested?

bearbin:
This is not a non-change - the MobHeight is one higher, and the while loop checks one block higher - the checking has been adjusted up in both places rather than up in one and down in the other.

LogicParrot:
>Not really - if a mob is standing on a partial block (snow, trapdoor, halfslab) their head may be offset in a way that will make the code skip one block, which could be just the crucial one.

Good point but I think half slabs need lots of code changes anyways. I'd say ignore that problem for now and just replace `floor` with `ceil`.

>This is not a non-change - the MobHeight is one higher, and the while loop checks one block higher - the checking has been adjusted up in both places rather than up in one and down in the other.

So, for a futuristic, exactly 2 block tall mob, the scan ends 2 blocks above its head.

LogicParrot:
>Good point but I think half slabs need lots of code changes anyways. I'd say ignore that problem for now and just replace floor with ceil.

To clarify, mobs currently just float on half slabs as if they're full blocks (unless someone made substantial changes while I was gone).

madmaxoft:
And that's the reason why I shouldn't review PRs when tired. Things slip past way too easily.

LogicParrot:
@bearbin is there any reason why you changed `while (CurrentBlock >= MobHeight)`?  I think `ceil` is literally the only line that needs to change.

bearbin:
The reason why I changed `while (CurrentBlock >= MobHeight)` was so that the server only checks the blocks above the mob's head, (This is so a < 1 block tall burnable mob in snow would burn).

If a blocktype list is not possible, the only choices are either:

 * Don't consider mobs with heads inside a block to be shaded.
 * Consider snow under a short mob as shading it.

I chose the first as I thought it less likely to happen.

LogicParrot:
>(This is so a < 1 block tall burnable mob in snow would burn).

Fair enough. Looks good to me.

PR #3962, opened 2017-08-25
Leashes work in 1.12.1

mathiascode:
The packet ID changed in Minecraft 1.12.1.

### DIFF
--- a/src/Protocol/Protocol_1_12.cpp
+++ b/src/Protocol/Protocol_1_12.cpp
@@ -1125,6 +1125,30 @@ void cProtocol_1_12::SendEntityMetadata(const cEntity & a_Entity)
 
 
 
+void cProtocol_1_12::SendLeashEntity(const cEntity & a_Entity, const cEntity & a_EntityLeashedTo)
+{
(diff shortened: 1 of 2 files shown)


src/Protocol/Protocol_1_12.cpp
```diff
@@ -1125,6 +1125,30 @@ void cProtocol_1_12::SendEntityMetadata(const cEntity & a_Entity)
...
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, 0x3c);  // Set Attach Entity packet
+	Pkt.WriteBEUInt32(a_Entity.GetUniqueID());
+	Pkt.WriteBEInt32(-1);  // Unleash a_Entity
+}
+
+
+
+
+
```
madmaxoft: Why the code move?

Seadragon91: Sorted because of the packet id?

mathiascode: Yes

PR #3964, opened 2017-08-26
Update mbedtls to 2.5.1

### LINKED ISSUE #3070: Cuberite SSL implementation Important Vulnerabilities (webadmin)
As I was implementing Let's Encrypt certificates I run a test to and to my surprise I've seen that cuberite implementation of PolarSSL has several important vulnerabilities:

SSLv3 protocol should be disabled, it's not.
RC4 cipher should be disabled, it's not.

So it's vulnerable to PODDLE

The cipher preference can also be change I recommend following the Mozilla Foundation guidelines, and  PolarSSL  supports it, force the preference of strong ciphers first, and remove those like RC4 and others known to be weak.

https://tls.mbed.org/api/ssl_8h.html (search for mbedtls_ssl_conf_ciphersuites )

https://wiki.mozilla.org/Security/Server_Side_TLS

DH params are also 1024 by default, it should be at least 2048 to avoid Logjam Attack:
https://tls.mbed.org/kb/cryptography/providing-diffie-hellman-or-dhm-parameters

peterbell10:
Fixes #3070

The biggest change other than naming is that what was `ssl_context` has been split into `mbedtls_ssl_context` and `mbedtls_ssl_config`.  The idea being that the config is constant during operation and can be shared between multiple ssl contexts which necessitates a new `cSslConfig` class and some api changes as config changes can't happen after an ssl context is setup.

Trying it out, I'm able to connect to a server with authentication enabled.


madmaxoft:
@tigerw again merges without waiting for reviews >:(

peterbell10:
Is it okay that this now depends on the ARMmbed/mbedtls repo?  Could someone with write access to cuberite/polarssl put the submodule commit there?

madmaxoft:
That's exactly what I was against. I prefer to keep our libs in our repos, for complete control over them.

Another gripe is with renaming the repos, it will cause some pain when updating local cloned copies.

I've updated the repo permissions, you should now be able to push whatever is needed to the polarssl repo.

peterbell10:
I know the pain of renaming as I was switching back and forth a lot while working on it but I thought keeping the name polarssl would be potentially confusing.

Thanks for the permissions, I'll go push it now.

LogicParrot:
My local compilation broke for some reason, (clearing the build folder completely and recompiling resolved it).

LogicParrot:
The error had to do with a `config.pl` script not being found.

PR #3966, opened 2017-08-27
Repair the removed constructor of cCuboid `cCuboid(otherCuboid)`

### LINKED ISSUE #3965: The LUA exposition of cCuboid(otherCuboid) were removed
Client version:   1.12.1
Server OS:        Linux (x86)
Commit id:        6bc503151746ea05842009983c7de932fa80cd03

Hi, I open this issuse because the constructor https://api.cuberite.org/cCuboid.html#()%20%3Ci%3E(constructor)%3C%2Fi%3E_5 isn't working anymore.

Note: I have resolving the problem, I send a pull request, just wait and update your cuberite version when it will be done

### Expected behavior

```lua
local SCuboid = cCuboid(otherCuboid);
```

Should produce a copy

### Actual behavior

From worldedit //stack

```lua
[00:17:39] LUA: error in function 'new'.
     argument #2 is 'cCuboid'; '[no object]' expected.

[00:17:39] Stack trace:
[00:17:39]   [C](-1): (no name)
[00:17:39]   [C](-1): cCuboid
[00:17:39]   Plugins/WorldEdit/Commands/Region.lua(448): (no name)
[00:17:39] Stack trace end
[00:17:39] Error in <attached> calling function <callback>()
```

Use WorldEdit, make a selection, try //copy, or //stack, (...)

Buom01:
Hi,
See: https://github.com/cuberite/cuberite/issues/3965
And also:
- https://github.com/cuberite/WorldEdit/issues/131
- https://github.com/cuberite/WorldEdit/pull/132

I just put one line removed https://github.com/cuberite/cuberite/commit/7bdbfad1bbabb84e650261ad31d2d9b47f8b12a9#diff-340f510ebeacdfbbbd90d90ff8090930L13
Probably removed by accident

Fixes #3965

### DIFF
--- a/src/Cuboid.cpp
+++ b/src/Cuboid.cpp
@@ -10,8 +10,12 @@
 ////////////////////////////////////////////////////////////////////////////////
 // cCuboid:
 
-
-
+cCuboid & cCuboid::operator =(cCuboid a_Other)
+{
+	std::swap(p1, a_Other.p1);
+	std::swap(p2, a_Other.p2);
+	return *this;
+}
 
 
 void cCuboid::Assign(int a_X1, int a_Y1, int a_Z1, int a_X2, int a_Y2, int a_Z2)
--- a/src/Cuboid.h
+++ b/src/Cuboid.h
@@ -17,6 +17,12 @@ public:
 	cCuboid(const Vector3i & a_p1, const Vector3i & a_p2) : p1(a_p1), p2(a_p2) {}
 	cCuboid(int a_X1, int a_Y1, int a_Z1) : p1(a_X1, a_Y1, a_Z1), p2(a_X1, a_Y1, a_Z1) {}
 
+	// tolua_end
+
+	cCuboid & operator =(cCuboid a_Other);
+
+	// tolua_begin
+
 	// DEPRECATED, use cCuboid(Vector3i, Vector3i) instead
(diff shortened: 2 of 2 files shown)


lkolbly:
I think it was I who accidentally removed that in #3874. :P 

Anyway, to make clang happy you'll need to re-add the copy assignment operator also: https://github.com/cuberite/cuberite/pull/3874/files#diff-2ebbdacd5770190cb86d8a1ccea60dcbL23

peterbell10:
It's probably better to just have the declaration in an `#if 0` and add a comment that its for tolua only.

mohe2015:
#ifdef LUAONLY should also work because LUAONLY is not defined and everybody would know what it means

madmaxoft:
@peterbell10 , @mohe2015 Hiding things with an always-false ifdef is not an option - ToLua++ still needs the C++ implementation which will be called by the generated bindings code.

peterbell10:
The C++ implementation _does_ exist.  It's generated by the compiler and as such it has the proper `noexcept` specifications and (if the same were true for `Vector3i`) `cCuboid` can remain trivially copyable both of these can make significant differences in performance when used with the standard library.

madmaxoft:
Okay, I forgot about the copy-constructor defaulting rules, I didn't think it would be created by the compiler. Then by all means do an `#ifdef` solution. Using a non-existing macro to make it more human-readable is preferred: `#ifdef TOLUA_PARSE` for example.

PR #3968, opened 2017-08-28
WebAdmin makeover

yangm97:
It seems like people on phones and other mobile devices are having issues using the webadmin, so I shamelessly copied stuff from cuberite.org in order to make the webadmin interface responsive. 

Since the webadmin has no templating engine, I had to hardcode some stuff… but hey, it works better than the old one.

So far I have done only the login page, hope to get this finished by tomorrow.

PS: This is my first (cuberite) PR, so be gentle, @madmaxoft 😁


yangm97:
screen shot 2017-08-28 at 06 10 18

williamhatcher:
Maybe try using this?

https://github.com/hughperkins/Jinja2CppLight

I don't know C++ but this is very light weight, has no dependences, and uses the popular Jinja templating

madmaxoft:
No, we don't need a templating engine, we already have Lua if we need something scripted.

williamhatcher:
It was just a suggestion. Maybe we can just do includes for the header and
footer?

yangm97:
@williamhatcher I tried looking for a Jekyll implementation in lua, so that the webadmin and the site could literally share code, but I didn’t find one.

I’m not too concerned about this right now – See my opinions regarding the web interface [here](https://forum.cuberite.org/thread-2632.html).

yangm97:
screen shot 2017-08-30 at 06 09 14

Updated the admin page. Not mobile friendly yet, judge by the desktop view. Thoughts?

Cl1608Ho:
A thin vertical line between the bar on the left and the content could be helpful.

I.e. A right border of the right color and the margin set correctly for the left box

madmaxoft:
I'll be brutally honest. I like the current one much more. The new style feels too disorganized, everything thrown into the same area without any sense of division.

Note that I'm quite old-fashioned and don't like many of the current "modern" UIs / webpages for similar reasons.

bearbin:
I like the colours and the consistency between this and the main site, but usability should always come above looks. The site as designed above is a downgrade in this regard from the old site. A few things I immediately notice on the screenshot:

 - Centered text makes it harder to read the vertical list.
 - Insufficient distinction between links and headings (aside from buttons or giant links, there is no excuse for the lack of an underline). A different colour is important too.
 - No separation between navigation and content.
 - Insufficient contrast for action button.
 - Irrelevant bottom bar - none of that information pertains to the webadmin.

Also, when you come to do the mobile layout, don't fall into the trap of hiding content just because the user is on mobile - the feature is there for a reason, don't hide it.

yangm97:
First of all, yes it’s very crude. Wanted to push something barely function as early as possible to gather feedback. There was no instance of two columns on the original website, so I just did the needed CSS (for displaying two columns) without adding any style because again I wanted to get feedback. So don’t panic 😆

> And don't use the fork icon for logout. Octicons have an icon for logout as well

That was a placeholder. Since I didn’t know where these icons came from, I just left this to the ~finishing touches~ (aka ask Mathias)

> A thin vertical line between the bar on the left and the content could be helpful

> without any sense of division

> No separation between navigation and content.

I know. I will try the vertical line. But what else would you recommend?

> there is no excuse for the lack of an underline

I disagree with this one. Not even the current webadmin uses underlines for links. Anyway, besides changing colours, I was thinking about adding bullet points or something similar to the icon the current webadmin puts to the left of links.

> Insufficient contrast for action button

That button was meant to be green, but the html side is missing the needed class. I will see if I can apply the right style to every button.

> Irrelevant bottom bar - none of that information pertains to the webadmin

screen shot 2017-08-30 at 23 47 57

🙃

> Also, when you come to do the mobile layout, don’t fall into the trap of hiding content just because the user is on mobile - the feature is there for a reason, don't hide it.

Besides adding a hamburger menu, there’s not much else left to do, nor hide. I have moved the stats from the top to the bottom bar, but I don’t think I like the fact it’s sticky. I may move the stats to the top on mobile but make it so it’s not sticky anymore.

Cl1608Ho:
I don't know where those icons come from exactly, but I guessed github's octicons: https://octicons.github.com. Which is wrong anyway cause it the lots of other icons used. Anyway

I would prefer any nonsticky version over any sticky one because often you don't have much space on mobile and adding a sticky bar at the bottom takes away quite a lot: not only the text itself, but also the margins and paddings that take up space (which is needed for the eye candy).

Better stick to either bottom or top but don't make it sticky please

And besides, as the webadmin is made for extensibility stuff like buttons should rather be styled globally imo instead of with fancy classes 😛

yangm97:
> And besides, as the webadmin is made for extensibility stuff like buttons should rather be styled globally imo instead of with fancy classes 😛

Turns out it was not a class issue – the current webadmin (plugins) use `<input>` instead of `<button>`.

mathiascode:
Nice to see some progress on this. I started working on a Webadmin design matching the main website a few weeks ago as well, but I managed to delete everything. 😐

mathiascode:
I made a rough sidebar design.
screenshot from 2017-09-01 02 14 05

yangm97:
Fancy! So we’re going for a full width layout, without a `max-width` container?

For the link colours, I like both. Maybe keep the current color for sidebar and use the other color for links on the content? (or the reverse).

You may have noticed I replaced the old reset with `normalize.css`. It was causing to some odd behaviour, like lack of (default) padding on lists and lack of border on `<hr>` to name a few.

yangm97:
> hope to get this finished by tomorrow

> commented 4 days ago

Cl1608Ho:
@yangm97 i actually don't know what would be better, cause it doesn't feel good if everything is stuck to the far left side :P depending on the size of the content I guess

yangm97:
@Cl1608Ho yeah, I got that feeling too. Maybe we could reverse the background color between the two and/or use a “card” design. See this print from kubernetes web ui I got from google images:

ui-dashboard

stefaang:
Nice work!
The kubernetes design with white cards is better indeed.
Mattias, I'd leave out the arrows and maybe even the bullets.

mathiascode:
screenshot from 2017-09-01 15 13 53
or
screenshot from 2017-09-01 15 11 56

madmaxoft:
I can't tell the difference between the two

mathiascode:
One is more like a white card, which was suggested above, while the other basically swaps the background color of the sidebar and main background.

bearbin:
Unless there are going to be multiple 'cards' the second one is much more appropriate. A single card is just extra padding and visual confusion.

mathiascode:
Proper cards will need some restructuring of the HTML code. I'll experiment a little to see if it's worth it or not.

madmaxoft:
Since the plugins are unlikely to make use of the cards, I'd say let's use the cardless design.

stefaang:
Regarding time / weather: I would help if they are split into 2 sections. And some icons would make it more obvious and attractive too.

yangm97:
@stefaang good suggestion, but I think this should be discussed on the [cuberite/core](https://github.com/cuberite/core) repository. This PR, on this repository, can only touch the outer “shell”.

mathiascode:
I don't have write access to the cuberite repository, but I have uploaded my modified code here: https://github.com/mathiascode/cuberite/tree/webadmin

The shell is pretty much done, except for updated reset and global styles for common HTML elements. The main page content controlled by the Core plugin is far from done, and we'll have to come up with some mobile-friendly table styling.

yangm97:
> we’ll have to come up with some mobile-friendly table styling

There very few instances where tables are actually required though.

madmaxoft:
@mathiascode You can still create a PR from your fork of the repo

yangm97:
I’m having some strange conflicts trying to merge mathias’ code (it’s not based on my fork).

mathiascode:
I removed some unneeded CSS leftovers. Sorry if it caused issues.

yangm97:
No problem. I can amend your changes on my fork and credit you on the commit message, sounds good?

yangm97:
Nvm, @Cl1608Ho told me on IRC how to credit you properly. 

I think the only remaining features to add to this PR is hamburger menu for phones, default heading style  and responsive tables. 

I will take a look at these this week.

mathiascode:
I integrated the sidebar into the bottom mobile menu in my branch.

shofian05121992:
shofian05121992@gmail.com

ePirat:
@madmaxoft 

> No, we don't need a templating engine, we already have Lua if we need something scripted.

Using a template engine has a lot of benefits though, a very big one being that it makes the code a lot easier to maintain. Another one is that it provides (usually) a lot of safety features, like ensuring proper escaping of variables, preventing XSS injections.

Hardcoding the "template" and mixing it with application logic is imho not a good solution.

mathiascode:
I've lost the source code for my changes, and will no longer work on this.

PR #3969, opened 2017-08-28
cCuboid: restore default copy construct and assign

peterbell10:
This is what I was talking about in #3966.  `#if 0` blocks as a hint to tolua++ but letting the compiler do the actual implementation.  The same thing is done in [`cItem`](https://github.com/cuberite/cuberite/blob/a5869b3c0909e658ece6a9c27037292ad138b918/src/Item.h#L81-L98).

Even if you don't care about `is_nothrow_copy_constructible` or `is_trivially_copy_constructable` hopefully we can agree that achieving the same goal with less code is a good thing.  Or at the very least that the compiler is less likely to write bugs.


madmaxoft:
@tigerw If you can't squash-merge, please don't merge at all.

tigerw:
The commits looked logically separate...

madmaxoft:
Then at least rebase-merge

PR #3975, opened 2017-08-29
cUUID: Fix FromHexDigit for digits >= 'a'

### LINKED ISSUE #3974: Commands broken if authentication mode is off
Client version:   1.12.1
Server OS:        Windows

### Expected behavior
Commands should work in game if authentication mode is off.

### Actual behavior
Get always trying to execute forbidden command. This has been broken by [Add cUUID class](https://github.com/cuberite/cuberite/pull/3871)

Get clean server:
1) Switch Authenticate to 0 in settings.ini
2) Start server
3) Login with client
4) Run `op <player name>` from console -> Should give the player admin rights
5) Try to run command `/rank` in game -> Will print `trying to execute forbidden command` in console

Running `/rank <player name>` from console shows that player has admin rights.

peterbell10:
Fixes #3974.

I should really write some tests for `cUUID`.

### DIFF
--- a/src/UUID.cpp
+++ b/src/UUID.cpp
@@ -68,11 +68,11 @@ static Byte FromHexDigit(char a_Hex)
 	}
 	if (('a' <= a_Hex) && (a_Hex <= 'f'))
 	{
-		return static_cast<Byte>(a_Hex - 'a');
+		return static_cast<Byte>(10 + (a_Hex - 'a'));
 	}
 	if (('A' <= a_Hex) && (a_Hex <= 'F'))
 	{
-		return static_cast<Byte>(a_Hex - 'A');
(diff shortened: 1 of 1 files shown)


src/UUID.cpp
```diff
@@ -68,11 +68,11 @@ static Byte FromHexDigit(char a_Hex)
 	}
 	if (('a' <= a_Hex) && (a_Hex <= 'f'))
 	{
-		return static_cast<Byte>(a_Hex - 'a');
+		return static_cast<Byte>(a_Hex - ('a' - 10));
```
madmaxoft: I think `10 + a_Hex - 'a'` would have been more understandable.

peterbell10: Done

Seadragon91:
Thanks for fine.

PR #3908, opened 2017-08-07
Add support for 1.12.1

Bond-009:
Fixes #3904

PR #3977, opened 2017-08-29
GetPacketID for protocol packet IDs

lkolbly:
I made a forum post (https://forum.cuberite.org/thread-3003-post-28517.html#pid28517) to float this idea, and then I got bored and implemented it anyway. Basically, this PR implements a method GetPacketID, which is basically a giant switch statement to look up the packet ID.

Protocols which have packets that are just packet ID changes can override the GetPacketID method, and then they don't have to copy/paste code. This PR removes about a thousand lines of duplicated protocol code as well.

The intent is to eliminate #3908 style code duplication and #3962 style patches when Mojang changes packet IDs.

GitHub is having trouble showing the diff in a meaningful way, the idea is that every send packet has the hardcoded packet ID replaced with a call to GetPacketID, and methods which override a parent method while only changing the packet ID are deleted entirely with the new packet ID recorded in PacketID.cpp.

All of the GetPacketID methods are in a single file to help the compiler optimize the super::GetPacketID calls.

I did some basic testing for each of 1.9, 1.9.4, 1.12, and 1.12.1 protocol clients:
- Logged in
- Switch to survival, break a block, pick up the drop, place it back.
- Run around and sprint and jump a bit.
- Punch a pig (or cow, or chicken) to death.
- Switch to night, use a bed.
- Switch to creative, run around and sprint and jump a bit.


src/Protocol/PacketID.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+
+
+
+
+UInt32 cProtocol_1_9_0::GetPacketId(eOutgoingPackets a_Packet)
+{
+	switch (a_Packet)
+	{
+	case sendAttachEntity:          return 0x40;
```
madmaxoft: Style: all the `case`s should have one more indent (the curly brace controls the indent)

src/Protocol/Protocol.h
```diff
@@ -56,9 +56,84 @@ class cProtocol
...
+		sendWindowItems,
+		sendWindowClose,
+		sendWindowOpen,
+		sendWindowProperty
+	};
+
 	/** Called when client sends some data */
 	virtual void DataReceived(const char * a_Data, size_t a_Size) = 0;
 
+	/** Called to figure out a packet ID */
```
madmaxoft: This could use a bit more description.
`/** Called to figure out the actual packet ID from the protocol-agnostic packet type. */

lkolbly: Fixed.

src/Protocol/PacketID.cpp
lkolbly: Fixed.

src/Protocol/Protocol.h
madmaxoft: Not really, you swapped the `protocol-agnostic packet type` into `packet agnostic protocol type` :)

lkolbly: And here I was, thinking all P-words were interchangeable :-P

bearbin:
This seems to have generated a conflict (I have no idea how, since it was fine before). Could this be merged manually?

lkolbly:
Did a bunch of merges, haven't tested (presumably fine, though).

Seadragon91:
@bearbin Did you checked that you had `Squash and merge` and not `Rebase and merge` selected?

bearbin:
@Seadragon91 That's not the issue, see the screenshot below.

selection_002

I tried merging using the CLI and git indicated a collision, and the history seemed to be completely messed up. 

I will try again now.

Seadragon91:
Check the select field, there is `Rebase and merge`, but this should be done as `Squash and merge`. Click on the arrow select it and the warning should be gone.

bearbin:
Thanks for the help, I thought the box was grayed out like the button.

Seadragon91:
Next time please also change the commit content. There are now all commit messages listed

src/Protocol/Protocol.h
```diff
@@ -56,9 +56,84 @@ class cProtocol
 
 	virtual ~cProtocol() {}
 
+	/** A list of all outgoing packets */
+	enum eOutgoingPackets
```
tigerw: Would you consider using scoped enums? Or too verbose?

src/Protocol/Protocol.h
```diff
@@ -56,9 +56,84 @@ class cProtocol
...
+	{
+		sendAttachEntity = 0,
```
tigerw: I think the convention around the rest of the code for enum names is the first letters of the enum name + the field name, so for example opSendAttachEntity.

src/Protocol/ProtocolRecognizer.h
```diff
@@ -52,6 +52,9 @@ class cProtocolRecognizer :
 	/** Translates protocol version number into protocol version text: 49 -> "1.4.4" */
 	static AString GetVersionTextFromInt(int a_ProtocolVersion);
 
+	/** GetPacketId is implemented in each protocol version class */
+	virtual UInt32 GetPacketId(eOutgoingPackets a_Packet) override { return 0; }
```
tigerw: It doesn't look like this should ever be reached? Perhaps if so, it is better to fail an assertion or something instead of returning zero.

src/Protocol/Protocol.h
lkolbly: I think that'd be pretty verbose, already some lines are going past 80 columns.

tigerw: Don't let yourself be shackled by the width of some IBM punchcard! Believe in the power of the horizontal scrollbar and set yourself free! (or widescreen monitor if your prefer)

lkolbly: ...but then what should I do with all these boxes of punch cards? And here I thought I was so clever buying them in bulk.

tigerw: hm.
..
..
...I have it! Tape them together, and get a [music box](https://www.youtube.com/watch?v=L4bfyzQsbeM)!

It'll be like programming, but wait, it's musical programming, but *wait*, on punch cards which were designed for actual programming, but *wait*! you \*are* a programmer! Just how deep can we go with the levels of indirection‽

PR #3978, opened 2017-08-29
Prevent Squids from suffocating in water

bearbin:
This fixes squids from suffocating in water, as reported by @LogicParrot in #3942. Also fixes code duplication `cMonserter:WouldBurnAt()`.

Code is tested and squids do not suffocate in water, player does suffocate normally, however I have some questions about the code that should probably be resolved before merging:

 1. What is the purpose of the `+0.1` on [line 1657](https://github.com/cuberite/cuberite/blob/master/src/Entities/Entity.cpp#L1657) of Entity.cpp - should I get rid of it?
 2. Should squids use `IsSwimming()` or `IsSubmerged()`?  Currently they suffocate if their head is out of the water, but their height is low so their head should never be out of the water.
 2. Should I create methods of the cEntity class to get the head height, or to check if the entity is out of the world (above or below)? Those operations seem to be duplicated many times in the code, and centralising them could help avoid logic bugs.

**Please do not squash-merge, the two commits here are logically separate, rebase-merge instead.**

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1681,7 +1681,8 @@ void cEntity::SetSwimState(cChunk & a_Chunk)
 	m_IsSwimming = IsBlockWater(BlockIn);
 
 	// Check if the player is submerged:
-	VERIFY(a_Chunk.UnboundedRelGetBlockType(RelX, RelY + 1, RelZ, BlockIn));
+	int HeadHeight = CeilC(GetPosY() + GetHeight()) - 1;
+	VERIFY(a_Chunk.UnboundedRelGetBlockType(RelX, HeadHeight, RelZ, BlockIn));
 	m_IsSubmerged = IsBlockWater(BlockIn);
 }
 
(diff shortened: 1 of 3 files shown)


LogicParrot:
1. @tigerw ping. You introduced this in https://github.com/cuberite/cuberite/commit/fd7fc7e59efebd6b28012c8e8433f53ac853c555.

2. Can anyone think of any case where the choice matters? 

3. I fully agree, but I think we should centralize those after I finish the mob PR to minimize conflicts. Lots of code has been moved around.

lkolbly:
2. Maybe it matters when going down a waterfall? I haven't tested it, though.

lkolbly:
Imagine if the squid is at the very edge of the waterfall, so their head is in air but they're still swimming.

bearbin:
@tigerw Could you explain the magic number referenced in my question?

tigerw:
I do believe it was an attempt to resolve floaty punt errors when a player stands such that GetPosY() is some exact integral value.

tigerw:
That is,

Floaty punt

a floaty punt.

tigerw:
I presume squids will no longer drown there?

bearbin:
Was there a specific case you were working to avoid with the floaty punt addition, or was it just out of an abundance of caution?

In any case, I would say 0.1 is a bit big really (is there that much imprecision‽), and it needs to be a named constant not just a magic number. Let me know your thoughts.

tigerw:
I suppose the specific case was just that it wasn't consistently setting the right states.

Perhaps this is no longer the case, I don't really mind what you do. :P

Issue #3649, opened 2017-03-26
MoveToWorld crashes when called twice at the same tick


mathiascode:
Client version:   1.11.2
Server OS:        Linux
Commit id:        9fa8477b73ce878c631ba875ec10746b4143ba75

### Expected behavior
MoveToWorld should not crash the server.

### Actual behavior
If MoveToWorld is repeated many times, it crashes the server.

```
[14:43:44]   D:    | Cuberite has encountered an error and needs to close
[14:43:44] Details | SIGSEGV: Segmentation fault
[14:43:44] Cuberite local build build id: Unknown
[14:43:44] from commit id: approx: 9fa8477b73ce878c631ba875ec10746b4143ba75 built at: approx: 2017-03-10 01:45:11 -0600
./Cuberite(_Z15PrintStackTracev+0x1c)[0xa120ac]
./Cuberite[0x9efa89]
/lib/x86_64-linux-gnu/libc.so.6(+0x350e0)[0x7fd7c72400e0]
./Cuberite(_ZNSt4listIP7cEntitySaIS1_EE5beginEv+0x14)[0x8db874]
./Cuberite(_ZNSt4listIP7cEntitySaIS1_EE6removeERKS1_+0x20)[0x8dd130]
./Cuberite(_ZN6cChunk12RemoveEntityEP7cEntity+0x44)[0x8d7ac4]
./Cuberite[0xd87f95]
./Cuberite[14:43:44] Kicking player a for "Server is restarting. Please wait..."
[0xd87b82]
./Cuberite(_ZNKSt8functionIFvR6cWorldEEclES1_+0x58)[0x9dbb58]
./Cuberite(_ZN6cWorld15TickQueuedTasksEv+0x2ac)[0x9cfc7c]                                                                                                     [130/1961]
./Cuberite(_ZN6cWorld4TickENSt6chrono8durationIlSt5ratioILl1ELl1000EEEES4_+0x5c6)[0x9c0fb6]
./Cuberite(_ZN6cWorld11cTickThread7ExecuteEv+0xa5)[0x9c0945]
./Cuberite(_ZN9cIsThread9DoExecuteEv+0x39)[0xa04f59]
./Cuberite(_ZNKSt7_Mem_fnIM9cIsThreadFvvEEclIJEvEEvPS0_DpOT_+0x72)[0xa06642]
[...]

Issue #3976, opened 2017-08-29
Segfault: Command /portal


Seadragon91:
Client version:   1.12.1
Server OS:        Windows/Linux
Commit id:        fa5bf7d5cd2e4abb81c2934d1736251e8692458f

### Expected behavior
Using command `/portal` to switch world should work

### Actual behavior
Server crashes:
```
#0  0x0000000000000000 in ?? ()
#1  0x0000000000834ad3 in cPlayer::FreezeInternal (this=0x7fffb0037920, a_Location=...,
    a_ManuallyFrozen=<error reading variable: access outside bounds of object referenced via synthetic pointer>)
    at /home/lukas/cpp/cuberite/src/Entities/Player.cpp:2900
#2  cPlayer::DoMoveToWorld(cWorld*, bool, Vector3<double>)::$_0::operator()(cWorld&) const (this=0x7fffe94bb580,
    a_OldWorld=...) at /home/lukas/cpp/cuberite/src/Entities/Player.cpp:2018
#3  std::_Function_handler<void (cWorld&), cPlayer::DoMoveToWorld(cWorld*, bool, Vector3<double>)::$_0>::_M_invoke(std::_Any_data const&, cWorld&) (__functor=..., __args=...)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:1731
#4  0x00000000005f77f4 in std::function<void (cWorld&)>::operator()(cWorld&) const (__args=...,
    this=<optimized out>)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/functional:2127
#5  cWorld::TickQueuedTasks (this=<optimized out>) at /home/lukas/cpp/cuberite/src/World.cpp:1213
#6  0x00000000005ecbc1 in cWorld::Tick (this=<optimized out>, a_Dt=..., a_LastTickDurationMSec=...)
    at /home/lukas/cpp/cuberite/src/World.cpp:1023
#7  0x00000000005ec82f in cWorld::cTickThread::Execute (this=<optimized out>)
    at /home/lukas/cpp/cuberite/src/World.cpp:103
#8  0x00007ffff7608e6f in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x00007ffff7122494 in start_thread (arg=0x7fffecff5700) at pthread_create.c:333
[...]

PR #3922, opened 2017-08-12
Changed MoveToWorld to ScheduleMoveToWorld in cPlayer::Respawn

lkolbly:
If Respawn is called multiple times in a single tick, the MoveToWorld mechanics can become confused. ScheduleMoveToWorld has the same functionality (by way of calling MoveToWorld), while being idempotent (within a single tick).

Fixes #3913.

Issue #3651, opened 2017-03-26
MoveToWorld and ScheduleMoveToWorld status


mathiascode:
What should we do with MoveToWorld and ScheduleMoveToWorld? Both of them have their own issues:
MoveToWorld:
- Unstable, crashes may occur https://github.com/cuberite/cuberite/issues/3649
- Sends respawn packet

ScheduleMoveToWorld:
- No crashes
- Doesn't send respawn packets
- World dimensions don't change when changing worlds, resulting in weird looking chunks

Both:
- Player not always sent to world spawn: https://github.com/cuberite/cuberite/issues/3420

ScheduleMoveToWorld was supposed to be removed in https://github.com/cuberite/cuberite/pull/3401, but many months have passed. What is the status of that?

Seadragon91:
Also `MoveToWorld` is asynchronous. Doing a teleport after calling it, is not guaranteed to work as expected.

LogicParrot:
@mathiascode , I'll try to investigate these issues and perhaps revive #3401 when I finish the mob PR.

PR #3979, opened 2017-08-29
Added ShouldSendRespawn parameter to ScheduleMoveToWorld

lkolbly:
As part of a fix for #3976, this PR adds a ShouldSendRespawn parameter to ScheduleMoveToWorld.

See the Core PR https://github.com/cuberite/Core/pull/203

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -3946,8 +3946,13 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 							Type = "boolean",
 							IsOptional = true,
 						},
+						{
+							Name = "ShouldSendRespawn",
+							Type = "boolean",
+							IsOptional  = true,
+						},
 					},
-					Notes = "Schedules a MoveToWorld call to occur on the next Tick of the entity. If ShouldSetPortalCooldown is false (default), doesn't set any portal cooldown, if it is true, the default portal cooldown is applied to the entity.",
+					Notes = "Schedules a MoveToWorld call to occur on the next Tick of the entity. If ShouldSetPortalCooldown is false (default), doesn't set any portal cooldown, if it is true, the default portal cooldown is applied to the entity. If ShouldSendRespawn is false (default), no respawn packet is sent, if it is true then a respawn packet is sent to the client.",
 				},
 				SetGravity =
 				{
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1353,12 +1353,13 @@ void cEntity::DetectCacti(void)
 
 
 
-void cEntity::ScheduleMoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown)
+void cEntity::ScheduleMoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown, bool a_ShouldSendRespawn)
 {
 	m_NewWorld = a_World;
 	m_NewWorldPosition = a_NewPosition;
 	m_IsWorldChangeScheduled = true;
 	m_WorldChangeSetPortalCooldown = a_SetPortalCooldown;
+	m_WorldChangeSendRespawn = a_ShouldSendRespawn;
 }
 
 
@@ -1378,7 +1379,7 @@ bool cEntity::DetectPortal()
 			m_PortalCooldownData.m_ShouldPreventTeleportation = true;
 		}
 
-		MoveToWorld(m_NewWorld, false, m_NewWorldPosition);
+		MoveToWorld(m_NewWorld, m_WorldChangeSendRespawn, m_NewWorldPosition);
 		return true;
 	}
 
--- a/src/Entities/Entity.h
+++ b/src/Entities/Entity.h
@@ -428,7 +428,7 @@ public:
 	virtual void TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ);
 
 	/** Schedules a MoveToWorld call to occur on the next Tick of the entity */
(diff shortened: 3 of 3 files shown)


LogicParrot:
I haven't reviewed this and this is a general remark:

 I think someone should take a step back and look at the world travel mechanism as a whole. We're patching up specific issues and ignoring the bigger picture. The API has redundancies, inconsistencies, and the code is probably messy too. There was a point where world travel worked well. Here are references to get started:

https://github.com/cuberite/cuberite/milestone/20

https://github.com/cuberite/cuberite/pull/3401

I'll try dealing with them after I finish the mob PR (if they're unfixed till then).

LogicParrot:
Also see my comment at https://github.com/cuberite/Core/pull/203

lkolbly:
https://xkcd.com/1428/

@LogicParrot I will mull it over a bit, but I think changing to ScheduleMoveToWorld will fix #3649 and #3651 (though the latter isn't really an issue, but this PR fixes the respawn packet issues).

As far as I know, the problem with MoveToWorld directly is that each call puts another function on the queue to run in the world tick thread - when the next world tick comes, they all run. When each one runs, they [remove the player](https://github.com/cuberite/cuberite/blob/master/src/Entities/Player.cpp#L2014) from the current world, and add it to the new world.

Which is fine, except that in cWorld, new player and to-be-removed players are also [kept in a queue](https://github.com/cuberite/cuberite/blob/master/src/World.cpp#L3067), which means that by the time the MoveToWorld lambda is called a second time (in the tick), the player hasn't yet been added to the new world, so when it tries to remove the player it crashes.

From my understanding, ScheduleMoveToWorld by its nature restricts there to being only a single outstanding MoveToWorld lambda (per player) at any given time. There are surely many ways to fix it - the lambda could be rewritten in such a way that it works if called multiple times in a single tick, MoveToWorld could refuse to enqueue multiple lambdas, ScheduleMoveToWorld could be used as the front end for MoveToWorld.

Take a look at #3922, which was a similar solution to a similar problem.

Anyway, that's my two cents.

LogicParrot:
Thanks for the details!

Is there any merit in having *both* methods of world travel?

It's been a while since I dealt with world travel, and I have no strong preference for `MoveToWorld` specifically. (It seemed better at the time to me). If the other way around is better, that's fine too. The out of date documentation originally obsoleted `MoveToWorld` for a no longer valid reason, but you're saying there are also other reasons which I wasn't aware of. 

My main point is that ideally, the code should eventually be cleaned up. We surely have redundancies. We stacked patch upon patch without looking at the whole thing. Among other cleanups, perhaps one of the methods can be removed unless there's a need for both (keeping the API call exposed for backward compatibility, which only wraps the unremoved one). 

This PR looks good for the specific problems at hand.

mathiascode:
My concern is plugins. Most, if not all plugins that make use of world travel in some way are using MoveToWorld instead of ScheduleMoveToWorld at the moment.

LogicParrot:
Perhaps the API can remain intact. One method would just internally call the other.

lkolbly:
> Is there any merit in having both methods of world travel?

I don't think so, no. Exposing both would probably only serve to confuse plugin authors.

> ideally, the code should eventually be cleaned up.

Yeah, it'd be nice to step back and take a clean look at it. Looking at it now, it seems to me that coordinating these asynchronous removal/adding queues causes a lot of pain (and having to freeze the player, etc.).

> Among other cleanups, perhaps one of the methods can be removed unless there's a need for both

If we decide to put all our chips on ScheduleMoveToWorld, we'd still have to keep at least one overload of MoveToWorld (or a differently-named but similar function) internally.

> Most, if not all plugins that make use of world travel in some way are using MoveToWorld instead of ScheduleMoveToWorld at the moment.

> Perhaps the API can remain intact. One method would just internally call the other.

Backwards compatibility is good. I think, semantically, MoveToWorld is actually almost identical to ScheduleMoveToWorld, as long as you only call it at most once per tick (except that MoveToWorld calls the changing world hook immediately, ScheduleMoveToWorld would wait).

LogicParrot:
Yes, the code is very painful. (See the redundancy discussion at #3115).

Freezing is essential to prevent chunk falling when stepping/flying into unloaded chunk at the same world or traveling to a new world with unloaded chunks. Both scenarios are common on Raspberry Pis.

>Which is fine, except that in cWorld, new player and to-be-removed players are also kept in a queue, which means that by the time the MoveToWorld lambda is called a second time (in the tick), the player hasn't yet been added to the new world, so when it tries to remove the player it crashes.

What bothers me is that @mathiascode 's description at #3649 doesn't seem to indicate stuffing two travels at one tick. There could be other issues. (And travel was stable before).

lkolbly:
> Both scenarios are common on Raspberry Pis.

I have a few Pis laying around, I'll have to play around with it.

> What bothers me is that @mathiascode 's description at #3649 doesn't seem to indicate stuffing two travels at one tick

>for i=1,10 do
>	Player:MoveToWorld("world_nether")
>end

LogicParrot:
Oh. That'd obviously crash, but I was referring to:

>I haven't been able to find a consistent way to reproduce this when using MoveToWorld normally. Sometimes I've walked into a portal with normal walking speed, and the server has crashed.

But now that I think about it, portal travel uses `ScheduleMoveToWorld`.

LogicParrot:
>I have a few Pis laying around, I'll have to play around with it.

Try flying to an unloaded chunk with Essentials' `/flyspeed 10` command. That should trigger the freeze mechanism even on an average PC.

peterbell10:
>#3649 doesn't seem to indicate stuffing two travels at one tick.
I believe the first move sets `m_ParentChunk` to `nullptr` then the second calls `RemoveEntity` on it and it segfaults.

Issue #3159, opened 2016-04-22
Kick packet, can be ignored from client / Server doesn't close connection, correctly


Seadragon91:
### Expected behavior

Connection to client should be fully closed, if the player has been kicked / banned from server.
### Actual behavior

Connection is not closed and player is still in the world.
### Steps to reproduce the behavior

Used this [minecraft client python](https://github.com/ammaraskar/pyCraft) to test packets.

109C:
That's odd, before when I tried this the server ended the connection after kicking me.

Seadragon91:
Did you use your minecraft client?

109C:
I used the vanilla client with a proxy that just dropped all the kick / disconnect packets.

Seadragon91:
Strange. I tested this with the python client a few times. Every time the client still was connected.

tigerw:
I think fixed in #3115.

Seadragon91:
I see that there was a call to [Destroy](https://github.com/cuberite/cuberite/commit/291370e367f9c51f894ff132860ac2b55727181f) a while ago, but it has been removed.

How could this problem be fixed?

tigerw:
Fixed in a pull request that is yet to be merged because I've been too busy :(

Seadragon91:
No way to block a unwanted player.

Seadragon91:
If I connect with pycraft to the server and run `restart`. The server restarts and then aborts with the message that the server port is already in use. If I close pycraft before restarting it works fine.

PR #3980, opened 2017-08-30
OnLogin now kicks players properly

### LINKED ISSUE #3868: HOOK_LOGIN causes client to hang and timeout, instead of getting kicked
Client version:   1.12
Server OS:        macOS
Commit id:        6e8ec2fe340f8fc63f9394a4701aeace6e7e2701

### Expected behavior

From https://api.cuberite.org/OnLogin.html: "If the function returns true, no other plugins are called for this event and the client is kicked."

### Actual behavior

Instead of getting kicked, the client hangs on "Logging in..." for some time, then fails with "Timed out". 

screen shot 2017-07-22 at 4 23 59 pm

Tested with this plugin:

```lua
function Initialize(Plugin)
    cPluginManager.AddHook(cPluginManager.HOOK_LOGIN, OnLogin)
    return true
end

function OnLogin(Client, ProtocolVersion, UserName)
    LOG("Received HOOK_LOGIN")
    -- Client:Kick("Kicking client")
    return true
end 
```

Returning true from HOOK_LOGIN denies the login, this part works, but there seems to be a missing kick packet or something along those lines, because the client just hangs instead of being properly kicked.

If I comment out the Client::Kick call (wherein I tried to specify a custom kick message, is this possible? **edit**: kicking in HOOK_PLAYER_JOIN works as expected), the server does log that it sent the kick message:

```
Log: [827c124d61017f6c|16:27:10] Client "::ffff:127.0.0.1" connected!
Log: [827c124d61017f6c|16:27:10] New ClientHandle created at 0x10de11388
Log: [29da2fe1e6426e34|16:27:10] Received HOOK_LOGIN
Log: [29da2fe1e6426e34|16:27:10] Sending a DC: "Kicking client"
```

but it doesn't appear on the client. After the client times out, there is an assertion failure in the cClientHandle destructor:

Error: [827c124d61017f6c|16:27:40] Assertion failed: m_State == csDestroyed, file cuberite/src/ClientHandle.cpp, line 115
Assertion failed: (0), function ~cClientHandle, file cuberite/src/ClientHandle.cpp,

bearbin:
Fixes #3868 

The code is tested and works.

Related to this - maybe my understanding of the code is off, but I couldn't see any way that kicking is enforced after the disconnect packet is sent. Am I missing something?

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -722,9 +722,10 @@ bool cClientHandle::HandleLogin(const AString & a_Username)
 		// Let the plugins know about this event, they may refuse the player:
 		if (cRoot::Get()->GetPluginManager()->CallHookLogin(*this, m_ProtocolVersion, a_Username))
 		{
-			Destroy();
+			SendDisconnect("Login Rejected!");
 			return false;
 		}
+
 		m_State = csAuthenticating;
 	}  // lock(m_CSState)
 


src/ClientHandle.cpp
```diff
@@ -722,9 +722,10 @@ bool cClientHandle::HandleLogin(const AString & a_Username)
 		// Let the plugins know about this event, they may refuse the player:
 		if (cRoot::Get()->GetPluginManager()->CallHookLogin(*this, m_ProtocolVersion, a_Username))
 		{
-			Destroy();
+			SendDisconnect("Login Rejected!");
```
LogicParrot: I keep forgetting who handles what in clientHandle code. Does not calling destroy cause leaks?

peterbell10:
For the related point: #3159

bearbin:
@peterbell10, thanks for the link. I'll put that on my todo list.

src/ClientHandle.cpp
bearbin: I believe that the Disconnect packet should be sent first. Once the player disconnects, the server automatically calls the `Destroy()` method. Other places where the player is kicked also use the `SendDisconnect` method.

LogicParrot: I see. This is not PR-specific but is the client's cooperation required? Can a malicious client ignore `SendDisconnect`, preventing player destruction?

bearbin: Yes, a malicious client could ignore the packet. I intend to fix the issue, as discussed below with @peterbell10.

PR #3981, opened 2017-08-30
Added cLuaStateParams::Read

madmaxoft:
Added `cLuaStateParams::Read` function for writing high-level bindings.

DO NOT MERGE YET, this doesn't work in gcc / clang and I need help from someone knowledgable with "template magic" to help me fix it.

### DIFF
--- a/src/Bindings/LuaStateParams.h
+++ b/src/Bindings/LuaStateParams.h
@@ -9,6 +9,7 @@
 
 
 #include "LuaState.h"
+#include "tolua++/include/tolua++.h"
 
 
 
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -3,9 +3,9 @@
 
 #include "ManualBindings.h"
 #undef TOLUA_TEMPLATE_BIND
+#include "LuaStateParams.h"
 #include <sstream>
 #include <iomanip>
-#include "tolua++/include/tolua++.h"
 #include "polarssl/md5.h"
 #include "polarssl/sha1.h"
 #include "PluginLua.h"
@@ -3452,6 +3452,56 @@ static int tolua_cServer_RegisterForgeMod(lua_State * a_LuaState)
 
 
 
+static int tolua_cServer_TestApiSpeedOld(lua_State * a_LuaState)
+{
+	cLuaState L(a_LuaState);
+	if (
+		!L.CheckParamSelf("cServer") ||
+		!L.CheckParamString(2, 3) ||
+		!L.CheckParamNumber(4) ||
+		!L.CheckParamEnd(5)
+	)
+	{
+		return 0;
+	}
+
+	cServer * Server;
+	AString Name, Version;
+	UInt32 Protocol;
+	L.GetStackValues(1, Server, Name, Version, Protocol);
+
+	return 0;
+}
+
+
+
+
+
+static int tolua_cServer_TestApiSpeedNew(lua_State * a_LuaState)
+{
+	cLuaState L(a_LuaState);
+	cServer * Server;
+	AString Name, Version;
+	UInt32 Protocol;
+	auto self = cLuaStateParams::self(Server);
+	auto staticSelf = cLuaStateParams::staticSelf<cServer>();
+	switch (cLuaStateParams::Read(L,
+		std::tie(self,       Name, Version, Protocol),
+		std::tie(staticSelf, Name, Version, Protocol)
+	))
+	{
+		case 0:
+		{
+			// Do nothing
+		}
+	}
+	return 0;
+}
+
+
+
+
+
 static int tolua_cScoreboard_GetTeamNames(lua_State * L)
 {
 	cLuaState S(L);
@@ -4220,7 +4270,9 @@ void cManualBindings::Bind(lua_State * tolua_S)
 		tolua_endmodule(tolua_S);
 
 		tolua_beginmodule(tolua_S, "cServer");
-			tolua_function(tolua_S, "RegisterForgeMod",            tolua_cServer_RegisterForgeMod);
+			tolua_function(tolua_S, "RegisterForgeMod", tolua_cServer_RegisterForgeMod);
+			tolua_function(tolua_S, "TestApiSpeedOld",  tolua_cServer_TestApiSpeedOld);
+			tolua_function(tolua_S, "TestApiSpeedNew",  tolua_cServer_TestApiSpeedNew);
 		tolua_endmodule(tolua_S);
 
 		tolua_beginmodule(tolua_S, "cStringCompression");


src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+	/** Utility struct that provides the functionality to cLuaStateParams::GetTypeDescription.
+	Supports decorators such as cLuaStateParams::cSelf<T>.
+	Must be a struct in order to support decorators, and must not be inside a class (gcc / clang complain about that).
+	The general declaration must not be used by the code, all types need a specialization returning the correct type. */
+	template <typename T>
+	struct TypeDescription
+	{
+		static const char * desc()
+		{
+			static_assert(false, "Type is missing a TypeDescription<> template");
```
peterbell10: I think clang and gcc don't like this.  Maybe leave `desc` out altogether or even leave the struct undefined.

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+		}
+
+		// No more overloads, report failure
+		return -1;
+	}
+
+
+
+	/** Helper struct to implement iterating over std::tuple elements */
+	template <int N> struct SizeT {};
```
peterbell10: `std::integral_constant<size_t, N>`?

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+		return GetStackValue(a_LuaState, N, std::get<N - 1>(a_Overload));
+	}
+
+
+	/** Terminator for the above compile-time-recursive function. */
+	template <typename... T>
+	static bool ReadSingleOverloadRecurse(cLuaState & a_LuaState, std::tuple<T...> & a_Overload, SizeT<1>)
+	{
+		return GetStackValue(a_LuaState, 1, std::get<0>(a_Overload));
+	}
```
peterbell10: Again this could be an empty base case with just `return true`.

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+
+
+	/** Reads one value from the Lua stack.
+	Returns true on success, false on failure.
+	The hard work is delegated into cLuaState that already has this function, but we need to specialize it for decorators (cNonNil, cSelf etc.). */
+	template <typename T>
+	static bool GetStackValue(cLuaState & a_LuaState, int a_StackPos, T && a_ReturnedVal)
+	{
+		return a_LuaState.GetStackValue(a_StackPos, std::forward<T>(a_ReturnedVal));
+	}
```
peterbell10: I don't think forwarding make sense here; rvalues shouldn't be out parameters.

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+
+
+	/** Reads one value from the Lua stack.
+	Returns true on success, false on failure.
+	Specialization: Retrieves a value that should represent a non-nil pointer. Used primarily for ReadParams(... nonNil(...)) */
+	template <typename T>
+	static bool GetStackValue(cLuaState & a_LuaState, int a_StackPos, cNonNil<T> & a_ReturnedVal)
```
peterbell10: Shouldn't a `cNonNil` be taken by value as it is itself a reference?

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+	{
+		auto res = GetStackValue(a_LuaState, a_StackPos, a_ReturnedVal.m_Dest);
+		if (res)
+		{
+			if (a_ReturnedVal.m_Dest == nullptr)
+			{
+				return false;
+			}
+		}
+		return res;
```
peterbell10: Are the nested if statements not just equivalent to `res && (a_ReturnedVal.m_Dest != nullptr))`?

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+	static AStringVector BuildMatcherErrorMessages(
+		cLuaState & a_LuaState,
+		const std::tuple<T1...> & a_Overload1,
+		T2 &&... a_OtherOverloads
+	)
+	{
+		auto res = BuildMatcherErrorMessages(a_LuaState, std::forward<T2>(a_OtherOverloads)...);
+		auto signature = BuildSingleOverloadDescription<T1...>();
+		auto msg = GetMatcherErrorMessage(a_LuaState, a_Overload1);
+		res.emplace_back(std::move(Printf("(%s): %s", signature.c_str(), msg.c_str())));
```
peterbell10: Don't need to `std::move` here.

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+		return GetTypeDescription<T1>() + ", " + BuildSingleOverloadDescription<T2, T3...>();
+	}
+
+
+	/** Terminator for the above function. */
+	template <typename T>
+	static AString BuildSingleOverloadDescription()
+	{
+		return GetTypeDescription<T>();
+	}
```
peterbell10: Again, an empty case is cleaner imo.

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+	|  |  |     +- Detail::TypeDescription<T>::desc
+	|  |  +- GetMatcherErrorMessage
+	|  |     +- GetParamMatchError <recursive>
+	|  |        +- CheckValueType
+	|  |           +- cLuaState::GetStackValue
+	|  +- cLuaState::ApiParamError
+	+- return [never reached]
+	*/
+	template <typename... OverloadTuples>
+	static int Read(cLuaState & a_LuaState, OverloadTuples &&... a_Overloads)
```
peterbell10: Again, I don't think rvalues make sense here.

src/Bindings/LuaStateParams.h
madmaxoft: Yes, I'll leave it out. I just wanted some place where the compiler would error out and there would be a direct comment about what the programmer should do if they ever encounter it. Suppose it could be put directly into the `cLuaStateParams::GetTypeDesc` function.

src/Bindings/LuaStateParams.h
madmaxoft: Just changing the type to `size_t` seemed to help in my tests

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+			return false;
+		}
+		if (lua_isnone(a_LuaState, sizeof...(T)))
+		{
+			// Too few params
+			return false;
+		}
+
+		// Read the tuple, compile-time-recursively:
+		return ReadSingleOverloadRecurse(a_LuaState, std::forward<std::tuple<T...>>(a_Overload), SizeT<sizeof...(T)>());
```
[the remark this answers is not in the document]
madmaxoft: I'm still a bit hazy about the details why `forward`ing is even needed, if you can recommend any good reading on it, please do.

src/Bindings/LuaStateParams.h
madmaxoft: It's just to cover all cases. Not sure if MSVC actually doesn't try to match a reference and won't accept a value.

src/Bindings/LuaStateParams.h
madmaxoft: Yes they are, but this form seems more explicit about what it requires. If you find it too ugly, I could change it.

src/Bindings/LuaStateParams.h
madmaxoft: I don't think I can do that, because the recursion adds a `, ` separator in between the levels, so the last level still needs to return something valid.

src/Bindings/LuaStateParams.h
```diff
@@ -0,0 +1,571 @@
...
+
+
+
+	/** Returns the Lua type representing the specified C++ type.
+	Supports decorators such as cSelf, references and pointers. */
+	template <typename T>
+	static AString GetTypeDescription(void)
+	{
+		return Detail::TypeDescription<typename std::remove_pointer<typename std::remove_reference<T>::type>::type>::desc();
+	}
```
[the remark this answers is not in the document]
madmaxoft: Since such a type is not mapped to a Lua type, the best outcome would be a `TypeError<> specialization not found`. Didn't test.

madmaxoft:
I have no idea how a visitor-based interface would look like. My goal is to make writing API bindings as easy as possible; from what I've seen some visitor-related code in Java, it always tends to be overly explicit.

src/Bindings/LuaStateParams.h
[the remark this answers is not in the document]
peterbell10: I learned it from the book "Effective Modern C++" by Scott Meyers.  Not sure about online references.

peterbell10:
I was thinking of somthing like:
```
CallOverloaded(
    cOverload<int, bool, char>(
        [](int a_Int, bool a_Bool, char a_Char)
        {
            // Use this overload
        }
    ),
    cOverload<int, bool, char, Vector3i>(
        [](int a_Int, bool a_Bool, char a_Char, Vector3i a_Vec)
        {
            // Use this overload
        }
    )
);
```

But it's just a sketch.  I haven't thought the details.

madmaxoft:
That has the unfortunate detail of duplicating the parameter types twice - in the template parameter and then in the function parameters. On the other hand, the API parameters don't need to be declared beforehand, which is nice.

peterbell10:
Yes I can't think of a way to avoid that unfortunately. It doesn't seem that different from having to use `std::tie` though and it pushes switching between implementations into the "library" code rather than the "user" code.

src/Bindings/LuaStateParams.h
peterbell10: Ah you're right, I missed that.

src/Bindings/LuaStateParams.h
peterbell10: Maybe it's just a quirk of how I read code but if `res` were called something like `canGetValue` then the one liner reads almost like English:  "can get value and returned val does not equal nullptr". It's probably just personal taste but I find it easier to read.

tigerw:
Do you mean the "error: no matching function for call to 'ReadInternal'" part?

Maybe you could try placing the std::ties into their own variables in tolua_cServer_TestApiSpeedNew, since it seems to want references to them.

And perhaps for "no matching function for call to 'GetParamMatchError'" the SizeT struct definition could be changed in LuaStateParams.h to use size_t N instead of int N since it complains that the "template argument does not have the same type as the its corresponding template parameter 'int' vs 'size_t'"

tigerw:
Or even sprinkle some consts in ReadInternal and children where it wants tuples so it knows you won't change them.

peterbell10:
In the light of [this SO post](https://stackoverflow.com/questions/11893141/) I think it should be possible to omit the object wrapper entirely and just have:
```cpp
CallOverloaded(
    [](int, bool, char){},
    [](int, bool, char, Vector3i){}
);
```

madmaxoft:
One thing that I can''t find a solution for is that we read all the parameters from the Lua stack either using a template recursion or by wrapping them into a `std::tuple` and there's no way to unpack the `std::tuple` into the lambda call afterwards.

peterbell10:
You should be able to do something similar to [`std::apply`](http://en.cppreference.com/w/cpp/utility/apply). Though I think it should be possible without actually using a tuple.  Something along the lines of:

```cpp
template <typename... Ts> struct TypeList{};

template<typename Lambda, typename Types, size_t... Args>
int CallSingleOverloadImpl(Lambda a_Lambda, Types, std::index_sequence<Is...>)
{
  return a_Lambda(GetValue<Is, Types>()...)
}

template<typename Lambda, typename... Args>
int CallSingleOverload(Lambda a_Lambda, TypeList<Args...>)
{
  return CallSingleOverloadImpl(
    a_Lambda,
    TypeList<Args...>{},
    std::make_index_sequence<sizeof...(Args)>()
  );
}
```

Where `GetValue` would index into the type list and return a value of that type.

peterbell10:
It's not pretty but I've got (what appears to be) a working proof of concept.
https://github.com/cuberite/cuberite/compare/LuaStateParamsRead...peterbell10:LuaStateParamsRead

madmaxoft:
That is a bit above my current level of knowledge, but I think it could be the right solution. I'll play with it some more over the weekend. Thanks.

PR #3982, opened 2017-08-30
Return to cuberite/polarssl for mbedtls submodule

peterbell10:


tigerw:
sumbodule

heh, new GitHub feature

tigerw:
Why is this needed? Wouldn't this just revert the update?

peterbell10:
I've already pushed the relevant commit to [cuberite/polarssl](https://github.com/cuberite/polarssl/tree/mbedtls_2.5.1) so this doesn't revert anything.

tigerw:
I think the intent to have a fork was to change some configuration settings for this specific project. Since it is working fine from the original source now, why have the library point to a source which now again isn't going to be touched for two years or so? Surely code like this is the last thing which should be left unpatched.

bearbin:
I think the idea is that having the code on our own repository allows easier modification if we need to, and prevents the external code disappearing suddenly and breaking the build. Patching is not very different than if we used the external repo, as the submodule has to be manually updated in any case.

PR #3985, opened 2017-08-31
Updated APIDoc:

Seadragon91:
- cPawn: Removed functions inherited from cEntity
- Removed functions that doesn't exists / not exported
- Added missing return types
- cClientHandle:SetLocale: Added const to passed AString ref
- Replaced param `cEntityEffect` with `cEntityEffect#eType`
- Removed API functions that accepts a instance of `cEntityEffect`: The class only has static functions
- Removed API functions that have been marked to be removed

### DIFF
--- a/src/Entities/Entity.h
+++ b/src/Entities/Entity.h
@@ -217,7 +217,6 @@ public:
 	int GetChunkZ(void) const { return FloorC(m_Position.z / cChunkDef::Width); }
 
 	void SetHeadYaw (double a_HeadYaw);
-	void SetHeight  (double a_Height);
 	void SetMass    (double a_Mass);
 	void SetPosX    (double a_PosX) { SetPosition({a_PosX, m_Position.y, m_Position.z}); }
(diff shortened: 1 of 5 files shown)


Seadragon91:
Found more to change. Please not merge.

Seadragon91:
Is ready for review.

src/Entities/Entity.h
```diff
@@ -300,6 +297,10 @@ class cEntity
 
 	// tolua_end
 
+	void SetHeight(double a_Height);
+
+	void SetWidth(double a_Width);
```
madmaxoft: Why exactly are these not exported anymore?

Seadragon91: In the APIDoc is written `Notes = "FIXME: Remove this from API",`

Seadragon91: It is okay to remove or should I revert?

madmaxoft: Merged as-is, we'll re-export these if they become needed.

PR #3988, opened 2017-09-01
Lighting now generally consistent with vanilla

### LINKED ISSUE #3849: You see too deep into water
Client version:   1.12
Server OS:        macOS
Commit id:        approx: 6a1984112146a14c96f43e9cdf758f5104d6d2a3

### Expected behavior
This is a screenshot is taken in Vanilla:
screen shot 2017-07-13 at 15 34 27

### Actual behavior
This is a screenshot is taken in Cuberite:
screen shot 2017-07-13 at 15 35 55

Look into water

## Note
It looks the same as in Vanilla with night vision:
screen shot 2017-07-13 at 15 34 42

bearbin:
Skylight is now dispersed by some blocks, instead of passing
through unimpeded.

Some blocks which were not marked as transparent are now marked as
such.

Water and other such blocks now attenuate light with the correct
intensity.

Generally changes were based on documentation in the Minecraft
Wiki: https://minecraft.gamepedia.com/Opacity#On_block_light;
however during play-testing on vanilla, lava was found not to
attenuate sky or block-light so the attenuation was removed.

This fixes #3849

This has been tested in game, but it could do with a review from somebody more knowledgeable about the lighting code.


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -294,6 +294,24 @@ return
...
+							Type = "number",
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns whether skylight may pass through a transparent block unimpeded.",
```
madmaxoft: "What transparent block"? -> `a block of the specified type` ? Or is it specific for transparent blocks only?

src/BlockInfo.h
```diff
@@ -28,9 +28,12 @@ class cBlockInfo
 	/** How much light do the blocks consume? */
 	NIBBLETYPE m_SpreadLightFalloff;
 
-	/** Is a block completely transparent? (light doesn't get decreased(?)) */
+	/** Is a block completely transparent to blocklight? (light doesn't get decreased(?)) */
 	bool m_Transparent;
 
+	/** Does a block disperse sky light? (only relevant for transparent blocks) */
+	bool m_SkylightDispersant;
```
madmaxoft: Rename to `m_IsSkylightDispersant` or `m_DoesDisperseSkylight` to indicate the bool-ness

bearbin: Should `m_Transparent` not also be `m_IsTransparent` then?

madmaxoft: Would have been better, but it's been exported to the API for too long so changing it might be dangerous :(

bearbin: The change is incoming.

src/BlockInfo.h
```diff
@@ -28,7 +28,7 @@ class cBlockInfo
 	/** How much light do the blocks consume? */
 	NIBBLETYPE m_SpreadLightFalloff;
 
-	/** Is a block completely transparent? (light doesn't get decreased(?)) */
+	/** Is a block completely transparent to blocklight? (light doesn't get decreased(?)) */
```
peterbell10: Aren't slabs and stairs considered transparent but stop blocklight spreading?

Pokechu22: "light doesn't get decreased" is misleading anyways - block light always decreases by 1; that comment would only make sense in regards to skylight.

bearbin: @peterbell10 Yes, slabs and stairs are considered transparent yet stop all light spreading. In fact this PR adds that functionality. 

In truth, the "transparent" property isn't really anything to do with lighting, just that most "transparent" blocks let BlockLight (and often SkyLight) pass through. I will change the comment.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -294,6 +294,24 @@ return
...
+							Type = "number",
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns whether skylight may pass through a block of the specified type unimpeded.",
```
madmaxoft: Isn't this the other way around? If this returns `true`, it means that the skylight will NOT pass through this block.

bearbin: Updated.

PR #3991, opened 2017-09-02
Use CancelIoEx to notify input thread

tigerw:
- Removed potentially fragile calls to send a virtual keypress to the console.

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -187,7 +187,7 @@ template class SizeChecker<UInt8,  1>;
 #ifdef _WIN32
 
 	#define WIN32_LEAN_AND_MEAN
-	#define _WIN32_WINNT _WIN32_WINNT_WS03  // We want to target Windows XP with Service Pack 2 & Windows Server 2003 with Service Pack 1 and higher
+	#define _WIN32_WINNT _WIN32_WINNT_VISTA  // We want to target Windows XP with Service Pack 2 & Windows Server 2003 with Service Pack 1 and higher
 
 	// Windows SDK defines min and max macros, messing up with our std::min and std::max usage
 	#define NOMINMAX
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -243,7 +243,7 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 		for (;;)
 		{
 			m_StopEvent.Wait();
-
+			m_InputThreadRunFlag.clear();
 			break;
 		}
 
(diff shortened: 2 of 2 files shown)


src/Globals.h
```diff
@@ -187,7 +187,7 @@ template class SizeChecker<UInt8,  1>;
 #ifdef _WIN32
 
 	#define WIN32_LEAN_AND_MEAN
-	#define _WIN32_WINNT _WIN32_WINNT_WS03  // We want to target Windows XP with Service Pack 2 & Windows Server 2003 with Service Pack 1 and higher
```
bearbin: This comment seems outdated.

src/Root.cpp
```diff
@@ -252,7 +252,7 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 
 		LOG("Shutting down server...");
 		m_Server->Shutdown();
-	}  // if (m_Server->Start()
```
bearbin: Why was this comment removed?

tigerw: It didn't seem particularly useful to me :P

tigerw:
Also, does anyone know why pthread_kill returns error 3, yet still appears to work and stop the thread? It's giving a false error every time the server stops on Linux.

src/Root.cpp
bearbin: I think it's probably better left - it just gives an indication of what the long indented block was for.

bearbin:
Error 3 suggests that the thread being killed is already dead. I have no idea why it would happen though.

madmaxoft:
Ugh, again my review got stuck and wasn't posted.

Do we really want to lose XP compatibility over something that works anyway? I don't think this PR is necessary and it won't be needed in the long run (once Cuberite is split into the library part and the (G)UI part).

tigerw:
Is anyone running the server on XP? It seems unlikely.

When is this split projected to happen?

bearbin:
We shouldn't reduce compatibility just for the sake of reducing compatibility. You can run Cuberite on an Android device, even though you would be mad to do so - why not XP?

Also interested to know about this proposed split - can you link me some more information?

madmaxoft:
Here's one forum thread about the split, though I seem to recall it being discussed in other places, too, but can't find it.
https://forum.cuberite.org/thread-2740.html

Quite a lot of hardware is limited to XP (or equivalent), so let's just not cut it away.

[OT] Man, it would be awesome to actually see Cuberite running on an ATM :D

tigerw:
But sending virtual keypresses to a console window to close it is hardly a permanent solution.

Why are we so hung up about Windows XP?

madmaxoft:
Yes, it's not a permanent solution, it's there just until the library / UI split.

Incidentally, did you test your change with the `restart` console command and webadmin button? Does it work, and does it let the console to receive further commands after the restart?

peterbell10:
Closing as this has been inactive for over 6 months.

tigerw:
Well, whatever. @madmaxoft I do believe restart worked correctly, but I may have to try again to confirm.

PR #3993, opened 2017-09-02
Replace ItemCallbacks with lambdas

### LINKED ISSUE #3990: Core plugin commands not working
Client version:   1.12.1
Commit id:        700bbdabf53ab3b47151271526570f4a862b87f2

Multiple basic commands do not work. `gamemode` example:

```
#Ingame
> /gm c
< [INFO] Player not found

#Console
>gm creative <playername>
> Executing console command: "gm c <redacted playername>"
< Player not found
```

This is game breaking for some people. Marked important.

peterbell10:
I've tested this fixes #3990.

### DIFF
--- a/src/FunctionRef.h
+++ b/src/FunctionRef.h
@@ -1,0 +1,62 @@
+
+#pragma once
+
+// Guard against macro new
+#pragma push_macro("new")
+#undef new
+
+// Declared only so it can be partially specialized
+template <class Signature>
+class cFunctionRef;
+
+/** A light-weight, type-erased reference to a function object.
+Non-owning, be mindful of lifetimes. */
+template <class Ret, class... Args>
+class cFunctionRef<Ret(Args...)>
+{
+public:
+	/** Construct from a function object. */
+	template <class FunctionObject>
+	cFunctionRef(FunctionObject && a_FunctionObject)
+	{
+		static_assert(  // Ensure function object is callable with signature Ret(Args...)
+			std::is_convertible<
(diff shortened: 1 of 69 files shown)


LogicParrot:
Please note the lambda changes were reverted in master. You need to re-add everything in this pull request.

peterbell10:
Well I guess this will have to include all of #3948 now.

src/FunctionRef.h
```diff
@@ -0,0 +1,54 @@
+﻿
+#pragma once
+
+// Declared only so it can be partially specialized
+template <class Signature>
+class cFunctionRef;
+
+/** A light-weight, type-erased reference to a function object.
+Non-owning, be mindful of lifetimes. */
```
madmaxoft: Should be worth noting that this is almost the same as std::function, but the difference is ...

madmaxoft:
I assume the changes are identical to the original PR that was merged and then reverted, and then only changed into using `cFunctionRef`. The cFunctionRef still needs a bit more documentation, it would be great to know why it's needed, why std::function won't work for us (even add a reference to the GH issue in the doxy-comment), so that anyone reading the code knows immediately what it does, without diving into its details.

peterbell10:
Yes the changes are identical, in fact this is literally #3948 with my last commit `git reset` away.

madmaxoft:
And I assume you tested against the 3990

peterbell10:
Yup, assuming the comment didn't break anything anyway.

PR #3995, opened 2017-09-03
Rename cEntity:m_IsSubmerged to m_WaterState

### LINKED ISSUE #3987: Rename cEntity::m_IsSubmerged to m_IsHeadInWater
The member variable should be renamed to match its intended purpose. The getter should be renamed, too, but needs to keep the old Lua binding as well as the new, with the old one emitting a log message with a lua stack trace (see src/Bindings/ManualBindings.cpp)

bearbin:
And related changes, including manual bindings changes. Fixes #3987 

This has been tested in game and appears to work.

See #3996 for an alternative take on this issue.


bearbin:
Closing, as overall preference was for the other solution.

PR #3996, opened 2017-09-03
Rename cEntity:m_IsSubmerged to m_IsUnderWater

### LINKED ISSUE #3987: Rename cEntity::m_IsSubmerged to m_IsHeadInWater

bearbin:
And related changes, including manual bindings changes. Fixes #3987 

This has been tested in game and appears to work.

Please don't merge this, I will rebase and merge this manually once it's approved.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -3944,16 +3944,6 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 					},
 					Notes = "Returns true if the entity represents a {{cTNTEntity|TNT entity}}",
 				},
-				IsUnderWater =
-				{
-					Returns =
-					{
-						{
-							Type = "boolean",
-						},
-					},
-					Notes = "Returns true if the top of the entity is in water",
-				},
 				Killed =
 				{
 					Params =
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -3724,6 +3724,16 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 					},
 					Notes = "Returns true if the entity is invisible",
 				},
+				IsInWater =
+				{
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns true if the bottom of the entity is in water",
+				},
 				IsItemFrame =
 				{
 					Returns =
@@ -3872,7 +3882,7 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 							Type = "boolean",
 						},
 					},
-					Notes = "Returns true if the mob or player is submerged in water (head is in a water block). Note, this function is only updated with mobs or players.",
+					Notes = "Returns true if the entity is submerged in water (head is in a water block). Note, this function is only updated with mobs or players. <b>Currently deprecated in favour of IsUnderWater()</b>",
 				},
 				IsSwimming =
 				{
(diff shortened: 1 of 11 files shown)


bearbin:
I have updated this PR with new code, and would like a review.

Particularly, does it matter to modify the state of the entity during the checks before finally setting the variables to their final value at the end, as I am doing here? If these functions are only called during the tick thread, I can't see any issue but if it's better I can change them.

Also, Entities/Pickup.cpp line 159 checks for fire/lava both within the block of the pickup and below in order to delete it. Could this be replaced with a simple check of IsInLava() || IsInFire()?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -3886,6 +3926,16 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 					},
 					Notes = "Returns true if the entity represents a {{cTNTEntity|TNT entity}}",
 				},
+				IsUnderWater =
```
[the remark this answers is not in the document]
bearbin: Nope, that's a remnant of the old code, I'll remove it.

src/Entities/Entity.cpp
```diff
@@ -1647,37 +1602,70 @@ bool cEntity::MoveToWorld(const AString & a_WorldName, bool a_ShouldSendRespawn)
...
+		{
+			for (int y = MinY; y <= MaxY; y++)
+			{
+				BLOCKTYPE Block;
+				if (!a_Chunk.UnboundedRelGetBlockType(x, y, z, Block))
+				{
+					LOGD("SetSwimState failure: RelX = %d, RelY = %d, RelZ = %d, Pos = %.02f, %.02f}",
+						x, y, z, GetPosX(), GetPosZ()
+					);
+					return;
```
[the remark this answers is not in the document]
bearbin: Yes, that should be fine, I'll still log the error anyway.

src/Bindings/ManualBindings.cpp
```diff
@@ -3956,6 +3956,56 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
...
+	}
+
+
+	// API function deprecated:
+	LOGWARNING("cEntity:IsSubmerged() is deprecated. Use cEntity:IsHeadInWater() instead.");
+	cLuaState::LogStackTrace(tolua_S);
+
+	L.Push(Mtolua_new((bool)(self->IsHeadInWater())));
+	return 1;
+}
```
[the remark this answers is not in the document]
bearbin: Yes, will do. Thanks for linking an example.

bearbin:
> This would be better I think. That way it works properly when the pickup is between two blocks.

> I'm also wondering if IsOnFire would be enough.

I think IsOnFire would be enough, but that's uncovered another issue: SetSwimState is only called for mobs and players, so this probably breaks catching entities on fire. I'll have to call SetSwimState for all entities.

src/Bindings/ManualBindings.cpp
```diff
@@ -3958,22 +3958,22 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
...
 {
+	// Check the params:
 	cLuaState L(tolua_S);
-
-	// Get the params:
-	cEntity * self = reinterpret_cast<cEntity *>(tolua_tousertype(tolua_S, 1, nullptr));
-	if (self == nullptr)
+	if (!L.CheckParamSelf("cEntity"))
 	{
-		LOGWARNING("%s: invalid self (%p)", __FUNCTION__, static_cast<void *>(self));
```
bearbin: Is removing this warning OK or do I need to replace it with a new one not depending on self?

peterbell10: `CheckParamSelf` give a useful error message on its own.

bearbin: Great. Everything else is fixed per your comments then. The build failed on AppVeyor due to a merge conflict, and travis didn't get triggered for some reason but it builds on my machine.

peterbell10: Can you fix the merge conflict and hopefully travis will work next time.

bearbin: Done.

bearbin:
The builds of the latest commit passed, what CI are you referring to?

peterbell10:
I don't mean this PR specifically, I mean the issues with travis builds in general.  Until #4140 gets merged, I can't merge anything.

PR #3999, opened 2017-09-03
Shutdown connection when disconnect packet sent

### LINKED ISSUE #3159: Kick packet, can be ignored from client / Server doesn't close connection, correctly

bearbin:
Fixes #3159

Tested with a normal client, and it works fine, but no advanced testing undergone.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2521,6 +2521,9 @@ void cClientHandle::SendDisconnect(const AString & a_Reason)
 		LOGD("Sending a DC: \"%s\"", StripColorCodes(a_Reason).c_str());
 		m_Protocol->SendDisconnect(a_Reason);
 		m_HasSentDC = true;
+		// csKicked means m_Link will be shut down on the next tick. The
+		// disconnect packet data is sent in the tick thread so the connection
+		// is closed there after the data is sent.
 		m_State = csKicked;
 	}
 }
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2087,6 +2087,12 @@ void cClientHandle::Tick(float a_Dt)
 		return;
 	}
 
+	// If player has been kicked, terminate the connection:
+	if (m_State == csKicked)
+	{
+		m_Link->Shutdown();
+	}
+
 	// If destruction is queued, destroy now:
 	if (m_State == csQueuedForDestruction)
 	{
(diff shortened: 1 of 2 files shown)


Seadragon91:
Tested with pycraft and the client is now kicked. Thanks for the fix.

src/ClientHandle.cpp
```diff
@@ -2515,6 +2521,7 @@ void cClientHandle::SendDisconnect(const AString & a_Reason)
 		LOGD("Sending a DC: \"%s\"", StripColorCodes(a_Reason).c_str());
 		m_Protocol->SendDisconnect(a_Reason);
 		m_HasSentDC = true;
+		m_State = csKicked;
```
madmaxoft: Why not just shutdown the link here? Postponing until the (next) tick seems needless.

bearbin: The server crashes with a segfault if you do that, so I decided it wasn't the best idea.

madmaxoft: That's weird, and it means a race condition is probably present even in your code (although probably hard-to-replicate).
Can you investigate into that segfault a bit?

bearbin: I will try to, if I can replicate it.

bearbin: I found the race condition that was causing the segfault - once the client closed the connection after receiving the DC packet, `Destroy()` was called on the client, which would attempt to `Shutdown()` a `nullptr` `m_Link`.

However, if `m_Link` is shut down immediately after `m_Protocol->SendDisconnect()` is called, the kick message doesn't get through to the player - the connection must end before the data in transit gets through.

madmaxoft: Ah, right, now I remember. The packet sending is done from the client tick thread, so the DC packet is queued here, and then at least one tick is needed for it to be actually sent over the link. So yeah, this code is perfect, then.

Perhaps add a note of this next to this line, so that we know in the future.

bearbin:
The comment is added.

madmaxoft:
Sorry, it's not good enough. It doesn't explain that the packet sending is delayed and the data is actually sent in the Tick thread.

bearbin:
I have changed the comment again for you.

PR #4001, opened 2017-09-03
Added assert to cProtocolRecognizer::GetPacketId.

lkolbly:
As @tigerw mentioned in #3977, calling cProtocolRecognizer::GetPacketId may return an unexpected result if you expect it to return the packet ID.

This PR adds an assert. An alternative solution might be to return m_Protocol->GetPacketId, but I believe this is better because protocol IDs should stay hidden inside the protocol structure (and so nobody should be calling this and expecting results).

### DIFF
--- a/src/Protocol/ProtocolRecognizer.h
+++ b/src/Protocol/ProtocolRecognizer.h
@@ -52,13 +52,6 @@ public:
 	/** Translates protocol version number into protocol version text: 49 -> "1.4.4" */
 	static AString GetVersionTextFromInt(int a_ProtocolVersion);
 
-	/** GetPacketId is implemented in each protocol version class */
-	virtual UInt32 GetPacketId(eOutgoingPackets a_Packet) override
-	{
-		ASSERT(!"cProtocolRecognizer::GetPacketId should never be called! Something is horribly wrong! (this method being called implies that someone other than a Protocol-derived class is calling GetPacketId)");
-		return 0;
-	}
-
 	/** Called when client sends some data: */
 	virtual void DataReceived(const char * a_Data, size_t a_Size) override;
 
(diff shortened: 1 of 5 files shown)


src/Protocol/ProtocolRecognizer.h
```diff
@@ -53,7 +53,11 @@ class cProtocolRecognizer :
 	static AString GetVersionTextFromInt(int a_ProtocolVersion);
 
 	/** GetPacketId is implemented in each protocol version class */
-	virtual UInt32 GetPacketId(eOutgoingPackets a_Packet) override { return 0; }
+	virtual UInt32 GetPacketId(eOutgoingPackets a_Packet) override
+	{
+		ASSERT(!"cProtocolRecognizer::GetPacketId should never be called! Something is horribly wrong! (it implies that someone other than a Protocol-derived class is calling GetPacketId)");
```
bearbin: I think `it implies` is wrong here - it should be `this implies`.

lkolbly: Maybe `this method being called implies`?

tigerw:
I wonder if this can be enforced at compile time.

lkolbly:
StackOverflow says it violates the substitution principle, unless we refactor the inheritance structure substantially: https://stackoverflow.com/questions/21920595/is-there-a-way-to-prevent-a-base-class-method-from-being-called-on-an-instance-o

peterbell10:
I may be missing something but aren't you looking for `protected` access?

lkolbly:
...yes, that does exactly what we want. Accessible from cProtocol and nowhere else (I'll leave the assert in, though, in case cProtocolRecognizer tries to call it).

lkolbly:
Also added an assertion to the 1.8 protocol's GetPacketId, since that one's also unimplemented.

src/Protocol/Protocol_1_8.h
```diff
@@ -176,6 +173,9 @@ class cProtocol_1_8_0 :
 	/** Adds the received (unencrypted) data to m_ReceivedData, parses complete packets */
 	void AddReceivedData(const char * a_Data, size_t a_Size);
 
+	/** Nobody inherits 1.8, so it doesn't use this method */
+	virtual UInt32 GetPacketId(eOutgoingPackets a_Packet) override { ASSERT(!"GetPacketId for cProtocol_1_8_0 is not implemented."); return 0; }
```
madmaxoft: This deserves a properly indented function body.

lkolbly: Indented.

Issue #3994, opened 2017-09-02
cEntity:Destroy: Verify failed


Seadragon91:
Server OS:        Windows

### Actual behavior
Server crashes:
```
 	Cuberite.exe!abort() Line 77
 	Cuberite.exe!NonCtrlHandler(int a_Signal) Line 107
 	Cuberite.exe!raise(int signum) Line 516
 	Cuberite.exe!abort() Line 71
 	Cuberite.exe!terminate() Line 58
 	[External Code]		Annotated Frame
 	Cuberite.exe!_execute_onexit_table::__l2::<lambda>() Line 206
 	Cuberite.exe!__crt_seh_guarded_call<int>::operator()<void <lambda>(void),int <lambda>(void) & __ptr64,void <lambda>(void) >(__acrt_lock_and_call::__l2::void <lambda>(void) && setup, _execute_onexit_table::__l2::int <lambda>(void) & action, __acrt_lock_and_call::__l2::void <lambda>(void) && cleanup) Line 204
 	[Inline Frame] Cuberite.exe!__acrt_lock_and_call(const __acrt_lock_id) Line 908
 	Cuberite.exe!_execute_onexit_table(_onexit_table_t * table) Line 231
 	Cuberite.exe!common_exit::__l2::<lambda>() Line 215
 	Cuberite.exe!__crt_seh_guarded_call<void>::operator()<void <lambda>(void),void <lambda>(void) & __ptr64,void <lambda>(void) >(__acrt_lock_and_call::__l2::void <lambda>(void) && setup, common_exit::__l2::void <lambda>(void) & action, __acrt_lock_and_call::__l2::void <lambda>(void) && cleanup) Line 224
 	[Inline Frame] Cuberite.exe!__acrt_lock_and_call(const __acrt_lock_id) Line 908
 	Cuberite.exe!common_exit(const int return_code, const _crt_exit_cleanup_mode cleanup_mode, const _crt_exit_return_mode return_mode) Line 263
>	[Inline Frame] Cuberite.exe!cEntity::Destroy::__l2::<lambda>(cWorld &) Line 251
 	[Inline Frame] Cuberite.exe!std::_Invoker_functor::_Call(cEntity::Destroy::__l2::void <lambda>(cWorld &) &) Line 1790
 	[Inline Frame] Cuberite.exe!std::invoke(cEntity::Destroy::__l2::void <lambda>(cWorld &) &) Line 1790
[...]

Issue #3971, opened 2017-08-29
cNBTChunkSerializer::Entity: a_Entity was nullptr


Seadragon91:
Server OS:        Windows
Commit id:        a5869b3c0909e658ece6a9c27037292ad138b918

### Actual behavior
Server crashes. Will look if I can get more info.

```
>	Cuberite.exe!cNBTChunkSerializer::Entity(cEntity * a_Entity) Line 990	C++
 	Cuberite.exe!cChunk::GetAllData(cChunkDataCallback & a_Callback) Line 302	C++
 	Cuberite.exe!cChunkMap::GetChunkData(int a_ChunkX, int a_ChunkZ, cChunkDataCallback & a_Callback) Line 904	C++
 	[Inline Frame] Cuberite.exe!cWorld::GetChunkData(int) Line 2990	C++
 	Cuberite.exe!cWSSAnvil::SaveChunkToNBT(const cChunkCoords & a_Chunk, cFastNBTWriter & a_Writer) Line 499	C++
 	Cuberite.exe!cWSSAnvil::SaveChunkToData(const cChunkCoords & a_Chunk, std::basic_string<char,std::char_traits<char>,std::allocator<char> > & a_Data) Line 344	C++
 	Cuberite.exe!cWSSAnvil::SaveChunk(const cChunkCoords & a_Chunk) Line 160	C++
 	Cuberite.exe!cWorldStorage::SaveOneChunk() Line 269	C++
 	Cuberite.exe!cWorldStorage::Execute() Line 220	C++
 	[Inline Frame] Cuberite.exe!std::_Invoker_pmf_pointer::_Call(void(cIsThread::*)()) Line 1534	C++
 	[Inline Frame] Cuberite.exe!std::invoke(void(cIsThread::*)() &&) Line 1534	C++
 	[Inline Frame] Cuberite.exe!std::_LaunchPad<std::unique_ptr<std::tuple<void (__cdecl cIsThread::*)(void),cIsThread *>,std::default_delete<std::tuple<void (__cdecl cIsThread::*)(void),cIsThread *> > > >::_Execute(std::tuple<void (__cdecl cIsThread::*)(void),cIsThread *> &) Line 240	C++
[...]

PR #4003, opened 2017-09-03
Add cWorld::RemoveEntity and use in cEntity

peterbell10:
Credit to @lkolbly for spotting this in #3971.
I have no idea if it fixes the issue but `RemoveEntity` is definitely not thread safe.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -238,17 +238,17 @@ void cEntity::Destroy(bool a_ShouldBroadcast)
 		m_World->BroadcastDestroyEntity(*this);
 	}
 
-	cChunk * ParentChunk = GetParentChunk();
-	m_World->QueueTask([this, ParentChunk](cWorld & a_World)
+	auto ParentChunkCoords = cChunkDef::BlockToChunk(GetPosition());
+	m_World->QueueTask([this, ParentChunkCoords](cWorld & a_World)
 	{
 		LOGD("Destroying entity #%i (%s) from chunk (%d, %d)",
 			this->GetUniqueID(), this->GetClass(),
-			ParentChunk->GetPosX(), ParentChunk->GetPosZ()
+			ParentChunkCoords.m_ChunkX, ParentChunkCoords.m_ChunkZ
 		);
 
 		// Make sure that RemoveEntity returned a valid smart pointer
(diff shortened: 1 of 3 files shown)


Seadragon91:
@peterbell10 Is that a fix for the failed Verify, too?

peterbell10:
It fixes some data races with entities but I still get failed verifies.

Seadragon91:
If you are interested in looking into data races. You could look into [clang threadsanitizer](http://releases.llvm.org/3.9.1/tools/clang/docs/ThreadSanitizer.html)

peterbell10:
I can't get APIFuzzing checkapi to fail the verify any more so I'm fairly convinced this fixes #3994.

Seadragon91:
Tested with the code provided in the #issue 3994 and no crash. Thanks for the fix :)

peterbell10:
I'm not sure but I think `ASSERT(IsTicking())` can be removed from `cEntity::Destroy` now unless it is checking more than just being added to the world.

src/World.cpp
```diff
@@ -3709,6 +3709,35 @@ bool cWorld::HasEntity(UInt32 a_UniqueID)
 
 
 
+OwnedEntity cWorld::RemoveEntity(cEntity & a_Entity)
+{
+	// Check if the entity is in the chunkmap:
+	auto Entity = m_ChunkMap->RemoveEntity(a_Entity);
+
+	// If not, check if the entity is in the queue to be added to the world:
+	if (Entity == nullptr)
```
bearbin: To save on indentation here, would checking for `!= nullptr` and returning in that case not be better?

src/Entities/Entity.cpp
```diff
@@ -1600,16 +1600,15 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 
 	// Queue add to new world and removal from the old one
 	cWorld * OldWorld = GetWorld();
-	cChunk * ParentChunk = GetParentChunk();
 	SetWorld(a_World);  // Chunks may be streamed before cWorld::AddPlayer() sets the world to the new value
```
bearbin: This comment seems wrong, could it be updated?

src/World.cpp
peterbell10: Changed

PR #4004, opened 2017-09-05
Stabilise MoveToWorld

### LINKED ISSUE #3649: MoveToWorld crashes when called twice at the same tick

### LINKED ISSUE #3976: Segfault: Command /portal

peterbell10:
Fixes #3649 and fixes #3976.  I've run the scripts from the issues with much bigger loops to confirm they're fixed.

This uses the method I mentioned in #3649 but with the struct renamed `sWorldChangeInfo`. `DoMoveToWorld` now performs the move immediately and should also work correctly when the entity is still waiting to be added to the world because it uses `cWorld::RemoveEntity` and `cWorld::RemovePlayer` which check their respective adding queues.


worktycho:
This approach is fundamentally different from the previous approach which used the queue task function to execute a delayed spawn in a tick on the other worlds threads, it will need careful reviewing to ensure that it does not cause new potential deadlocks in plugins.

peterbell10:
I'm not sure what you mean @worktycho, `DoMoveToWorld` is always called from within the queued task on the tick thread.  In fact this PR has `MoveToWorld` doing less work outside of that task than before.

worktycho:
@peterbell10 As I understand it DoMoveToWorld references both worlds in a single thread, where the previous strategy was to have the removal executed on the old worlds tick thread, and the spawn on the new worlds tick thread.

src/Entities/Entity.cpp
```diff
@@ -1559,68 +1536,81 @@ bool cEntity::DetectPortal()
...
-	if (this->IsMob())
-	{
-		cMonster * Monster = static_cast<cMonster*>(this);
-		Monster->SetTarget(nullptr);
-		Monster->StopEveryoneFromTargetingMe();
-	}
 
-	// Queue add to new world and removal from the old one
-	cWorld * OldWorld = GetWorld();
-	SetWorld(a_World);  // Chunks may be streamed before cWorld::AddPlayer() sets the world to the new value
```
lkolbly: Does SetWorld need to still be called for cEntity? (or, is it and I just missed it)

(at any rate, old plugins might expect the entity's world to be the new world immediately after calling MoveToWorld)

peterbell10:
If you mean the call to `BroadcastSpawnEntity` then `cPlayer::DoMoveToWorld` was already calling it from the old worlds tick thread.  The call in `cEntity` seemed to be missing altogether, not coordinated with the new world.

src/Entities/Entity.cpp
peterbell10: It is set in `cWorld::AddEntity`.  Setting immediately from `MoveToWorld` could be up for discussion though.

lkolbly: This is ostensibly the PR that first introduced that particular call to SetWorld:
https://github.com/cuberite/cuberite/pull/1061

This PR later added the comment when it was moved into the queue system:
https://github.com/cuberite/cuberite/pull/2986

peterbell10: Note that at the time of #1061 `cWorld::AddEntity` didn't actually call SetWorld immediately.

src/Entities/Entity.cpp
```diff
@@ -1559,68 +1536,81 @@ bool cEntity::DetectPortal()
...
+	if (cRoot::Get()->GetPluginManager()->CallHookEntityChangingWorld(*this, *a_World))
+	{
+		// A Plugin isn't allowing the entity to change world
+		return false;
+	}
 
+	if (!IsWorldChangeScheduled())
+	{
+		// Schedule a new world change.
+		m_WorldChangeInfo = cpp14::make_unique<sWorldChangeInfo>();
```
madmaxoft: This is thread-unsafe. One thread may set the `m_WorldChangeInfo` to a valid value while another thread is between the `if` above and this line.

src/Entities/Entity.cpp
```diff
@@ -1559,68 +1536,81 @@ bool cEntity::DetectPortal()
...
+		GetWorld()->QueueTask(
+			[this](cWorld &)
+			{
+				DoMoveToWorld(*m_WorldChangeInfo);
+				m_WorldChangeInfo.reset();
```
madmaxoft: Again unsafe, another thread may schedule a move while DoMoveToWorld is executing, and that change will get lost.

peterbell10:
I think the thread safety issues with `m_WorldChangeInfo` could be fixed with an atomic version of `unique_ptr`.  However, I think all the world change stuff should be called under the chunk map cs anyway (which would mean the task is still wrong).

madmaxoft:
`MoveToWorld` can be called by the plugins from virtually any thread (including webadmin), so thread safety there is still a concern.

peterbell10:
Wouldn't a plugin have to go through a DoWith type function first though?

madmaxoft:
That's true. In that case, if you always expect a CS to be locked in the method, why not add an ASSERT for it? It would make debugging much easier :)

peterbell10:
Hrm, it seems that a plugin can get a `cPlayer` without the CS from command handlers.  This potentially breaks a lot of assumptions in the entity code.

peterbell10:
Is there a reason why `CallHookEntityChangedWorld` was never called in [`cPlayer::DoMoveToWorld`](https://github.com/cuberite/cuberite/blob/b12f4ef7d58cfe4d815feb2db1f88f223c7f2a61/src/Entities/Player.cpp#L1982-L2063)?

src/Entities/Entity.cpp
lkolbly: It may not be an issue anymore, then.

peterbell10:
I find it a bit odd that players are different from every other entity but as long as it's not missing I guess it's okay then.

src/Entities/Entity.cpp
peterbell10: Is the new atomic version okay?

lkolbly: I think conceivably https://github.com/cuberite/cuberite/pull/4004/files/#diff-f312969b3897e838a1085ffb60b7c4dfR1612 could be called between https://github.com/cuberite/cuberite/pull/4004/files/#diff-f312969b3897e838a1085ffb60b7c4dfR1604 and the if statement, which would cause a WorldChangeInfo to be scheduled twice in a row. I don't think that'd actually cause any problems, though, since most of the failsafes are in DoMoveToWorld.

peterbell10: Both exchanges are atomic and the comparison is done on the exchange result and not from an additional load of `m_WorldChangeInfo` so I don't think that's an issue.

src/OSSupport/AtomicUniquePtr.h
```diff
@@ -0,0 +1,81 @@
...
+		{
+			// Only release ownership from the caller if the exchange occurred
+			a_Desired.release();
+		}
+		return DidExchange;
+	}
+
+	T * exchange(std::unique_ptr<T> a_Ptr, std::memory_order a_Order = std::memory_order_seq_cst) NOEXCEPT
+	{
+		return m_Ptr.exchange(a_Ptr.release(), a_Order);
```
madmaxoft: Is this safe? If the exchange doesn't happen, `a_Ptr` stays released, I think. I'm not too educated in the memory ordering and these operations, though.

peterbell10: >If the exchange doesn't happen
Not sure what you mean. it's an unconditional exchange so it should always happen.

madmaxoft: Right, sorry, I was thinking about compare_and_exchange

bearbin:
@peterbell10, @madmaxoft, is this PR going to be ready to merge soon?

src/Entities/Entity.cpp
```diff
@@ -1559,68 +1536,87 @@ bool cEntity::DetectPortal()
...
+		// Moving to same world, don't need to remove from world
+		return;
 	}
 
 	// Stop ticking, in preperation for detaching from this world.
 	SetIsTicking(false);
 
-	// Tell others we are gone
-	GetWorld()->BroadcastDestroyEntity(*this);
+	a_WorldChangeInfo.m_NewWorld->BroadcastSpawnEntity(*this);
```
peterbell10: This line makes me uncomfortable.  It's locking the new world while the lock for the current is already held and thus a potential deadlock.

I think the cleanest way to fix this is to move the Broadcast call into `cWorld::Tick` when `m_EntitiesToAdd` is processed.  Anybody have thoughts on this?

mathiascode:
Tested this with #4484 and #4487, haven't noticed any issues so far. Also haven't seen any issues with invisible players in any form.

src/Entities/Entity.cpp
```diff
@@ -158,19 +157,28 @@ bool cEntity::Initialize(OwnedEntity a_Self, cWorld & a_EntityWorld)
...
 	cPluginManager::Get()->CallHookSpawnedEntity(a_EntityWorld, *this);
 
+	return true;
+}
+
+
+
+
+
+void cEntity::OnAddToWorld(cWorld & a_World)
```
mathiascode: Update m_LastSentPosition here, so entities spawn at the right location client-side.

PR #4007, opened 2017-09-07
Fix crash for players out of the world. Fixes #4006

### LINKED ISSUE #4006: Crash when flying above ~256 blocks
Client version:   1.12.1
Server OS:        Linux
Commit id:        approx: b12f4ef7d58cfe4d815feb2db1f88f223c7f2a61

### Actual behavior
The server crashed.

Keep flying up.

#### Core dump + exec + log files:
[cuberite_crash.tar.gz](https://github.com/cuberite/cuberite/files/1285283/cuberite_crash.tar.gz)

bearbin:
This PR is deliberately kept short to avoid merge conflicts with my other PRs.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1662,7 +1662,9 @@ bool cEntity::MoveToWorld(const AString & a_WorldName, bool a_ShouldSendRespawn)
 void cEntity::SetSwimState(cChunk & a_Chunk)
 {
 	int RelY = FloorC(GetPosY() + 0.1);
-	if ((RelY < 0) || (RelY >= cChunkDef::Height - 1))
+	int HeadRelY = CeilC(GetPosY() + GetHeight()) - 1;
+	ASSERT(RelY <= HeadRelY);
+	if ((RelY < 0) || (HeadRelY >= cChunkDef::Height))
 	{
 		m_IsSwimming = false;
 		m_IsSubmerged = false;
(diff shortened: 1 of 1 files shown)


src/Entities/Entity.cpp
```diff
@@ -1662,7 +1662,9 @@ bool cEntity::MoveToWorld(const AString & a_WorldName, bool a_ShouldSendRespawn)
 void cEntity::SetSwimState(cChunk & a_Chunk)
 {
 	int RelY = FloorC(GetPosY() + 0.1);
-	if ((RelY < 0) || (RelY >= cChunkDef::Height - 1))
+	int HeadRelY = CeilC(GetPosY() + GetHeight()) - 1;
+	ASSERT(RelY <= HeadRelY);
+	if ((RelY < 0) || (HeadRelY >= cChunkDef::Height - 1))
```
peterbell10: What should happen if an entity's feet are in water but its head is out of the world (and vice versa)?

bearbin: It doesn't matter?

lkolbly: If I read the code correctly (haven't tested it), IsSwimming would be false if the head is out of the world but the feet are in water, which means that if an entity were burning then walking in water at the world height would not put them out.

I think the flipped bug is also true, but it's hard to see how that situation would occur (since you'd just fall out of the water).

bearbin: That would be true, however it really isn't an important bug; fixing a crash is more important than walking slightly slower at the top of the world.

bearbin: After this is merged, that bug can be fixed in due course.

lkolbly: Fair.

bearbin:
Merging this, as it's a minor change and important.

PR #4009, opened 2017-09-08
Updated compile.sh script

Bond-009:
Changes:
* Gets number of threads on FreeBSD
* Fixes an issue where the script would return an error when you chose no compile mode
* Improves consistency
* 'Fixes' a wording issue: cores != threads
* yum -> dnf

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This script will download the correct binary from the project site.
 
 ### Compiling
 
- - You can compile automatically for Linux / *nix with the `compile.sh` script. The script is described below.
+ - You can compile automatically for Linux, macOS and FreeBSD with the `compile.sh` script. The script is described below.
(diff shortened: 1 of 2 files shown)


LogicParrot:
>Gets number of threads on FreeBSD

Are you sure yours is the right wording? Aren't we **getting the number of cores**, and then **choosing the number of compilation threads**? What does "getting the number of threads" mean?

LogicParrot:
Nevermind that, I understand now. I would change `THREAD_COUNT` to `CPU_THREAD_COUNT` to avoid confusion between CPU threads and compilation threads.

LogicParrot:
The YUM issue was not fixed.

Bond-009:
@LogicParrot What do you mean?

PR #4013, opened 2017-09-09
Fix switch warnings

peterbell10:
* Fix a variety of `-Wswitch` and `-Wswitch-enum` warnings
  * Remove unneeded `-Wno-error` flags

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,4 +1,4 @@
-﻿project (Cuberite)
+project (Cuberite)
 
 
 include_directories (SYSTEM "${CMAKE_CURRENT_SOURCE_DIR}/../lib/")
@@ -164,13 +164,9 @@ include_directories (SYSTEM "${CMAKE_CURRENT_SOURCE_DIR}/../lib/TCLAP/include")
 configure_file("BuildInfo.h.cmake" "${CMAKE_CURRENT_SOURCE_DIR}/BuildInfo.h")
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-	set_source_files_properties(BiomeDef.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=switch-enum")
 	set_source_files_properties(BlockID.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors")
-	set_source_files_properties(BoundingBox.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors")
 	set_source_files_properties(ByteBuffer.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors")
 	set_source_files_properties(ClientHandle.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors ")
-	set_source_files_properties(CompositeChat.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors")
-	set_source_files_properties(MobSpawner.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=switch-enum")
 	set_source_files_properties(Statistics.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors")
 endif()
 
--- a/src/Generating/StructGen.cpp
+++ b/src/Generating/StructGen.cpp
@@ -198,65 +198,69 @@ int cStructGenTrees::GetNumTrees(
 	const cChunkDef::BiomeMap & a_Biomes
 )
 {
-	int NumTrees = 0;
-	for (int x = 0; x < cChunkDef::Width; x++) for (int z = 0; z < cChunkDef::Width; z++)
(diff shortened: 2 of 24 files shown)


src/Generating/StructGen.cpp
```diff
@@ -198,65 +198,69 @@ int cStructGenTrees::GetNumTrees(
...
+			case biColdTaigaM:           return 30;
+			case biMegaSpruceTaiga:      return 30;
+			case biMegaSpruceTaigaHills: return 30;
+			case biExtremeHillsPlusM:    return 4;
+			case biSavannaM:             return 8;
+			case biSavannaPlateauM:      return 12;
+			case biMesaBryce:            return 4;
+			case biMesaPlateauFM:        return 12;
+			case biMesaPlateauM:         return 12;
+			default:                     return 0;
```
bearbin: Is it right to have a default here? Ideally we should just list them all, and then the compiler will warn us when a new biome is added, so we can specify the tree level.

At least the default should assert or something.

src/Protocol/Protocol_1_10.cpp
```diff
@@ -1003,5 +1003,7 @@ void cProtocol_1_10_0::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_
 			a_Pkt.WriteBool(ZombiePigman.IsBaby());
 			break;
 		}  // case mtZombiePigman
+
+		default: break;
```
bearbin: Again, is default appropriate here?

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -732,22 +732,7 @@ void cNBTChunkSerializer::AddMonsterEntity(cMonster * a_Monster)
...
-			case mtGiant:
-			case mtGuardian:
-			case mtIronGolem:
-			case mtMooshroom:
-			case mtSilverfish:
-			case mtSnowGolem:
-			case mtSpider:
-			case mtSquid:
-			case mtWitch:
+			default:
```
bearbin: Is default what we want here?

src/Generating/StructGen.cpp
peterbell10: Well there are 80 biomes and only ~50 cases here so the defaulting behaviour must be intended.

src/Protocol/Protocol_1_10.cpp
peterbell10: I've made everything explicit in the 1.12 protocol but I'm not sure if it makes sense for anything other than the latest protocol.

src/Generating/StructGen.cpp
bearbin: I think probably we should handle the rest of these, unless any new biomes are expected not to have trees,

src/Protocol/Protocol_1_10.cpp
bearbin: I suppoe this makes sense, although I'd like another dev to weigh in here.

lkolbly: If you explicitly listed parrots (added in 1.12), to this (1.10) switch, it's unclear what it could do other than "break". Also every time a mob is added then every protocol needs to be updated whether the protocol supports it or not. The problem being if in 1.13 somebody implements parrots but forgets to add them to the 1.12 protocol, but I think that's less of a problem (and if it does happen, the bug report will have a quick and easy fix).

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1652,6 +1655,7 @@ void cWSSAnvil::LoadOldMinecartFromNBT(cEntityList & a_Entities, const cParsedNB
 		case 2: LoadMinecartFFromNBT(a_Entities, a_NBT, a_TagIdx); break;  // Minecart with furnace
 		case 3: LoadMinecartTFromNBT(a_Entities, a_NBT, a_TagIdx); break;  // Minecart with TNT
 		case 4: LoadMinecartHFromNBT(a_Entities, a_NBT, a_TagIdx); break;  // Minecart with Hopper
+		default: LOGWARNING("cWSSAnvil::LoadOldMinecartFromNBT: Unhandled minecart type"); break;
```
lkolbly: It could be useful for triaging to print out the minecart type.

src/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-﻿project (Cuberite)
+project (Cuberite)
```
madmaxoft: What's the change here? I don't see anything.

src/Protocol/Protocol_1_10.cpp
madmaxoft: I think leaving out the mobs that the protocol doesn't know is about the best thing we could do.

src/CMakeLists.txt
peterbell10: I think this is removing a [UTF8 BOM](https://github.com/cuberite/cuberite/pull/3936#issuecomment-325988099) that got added at some point.

Issue #916, opened 2014-04-21
Blocks burning under the rain


axisd:
![2014-04-21_11 51 53](https://cloud.githubusercontent.com/assets/2901788/2753211/a63a5dfa-c92c-11e3-892c-ff754ec72291.png)

bearbin:
Fixed by #4017

PR #4017, opened 2017-09-10
Improve fire performance during rain

bearbin:
Entities will now be extinguished if they are under non-solid blocks, and fire will now be extinguished by rain.

Fixes issue #916

Please don't squash merge this - the first commit is not related to the second. I will do a manual rebase-merge when the time comes (to squash the relevant commits).


Bond-009:
Rain can go through some non-solid blocks:
MCWiki: "Raindrops fall through ladders, vines, carpets, redstone repeaters, snow layers, mob heads, flower pots and cobwebs. They are stopped by signs, banners, doors, fence gates, trapdoors, pressure plates, glass panes and iron bars."
Ref: https://minecraft.gamepedia.com/Rain#Rain_particles

Bond-009:
Now you can replace this https://github.com/cuberite/cuberite/blob/master/src/Mobs/Enderman.cpp#L210 with your new `cWorld::IsWeatherWetAtBlock` method.

bearbin:
> Rain can go through some non-solid blocks:
MCWiki: "Raindrops fall through ladders, vines, carpets, redstone repeaters, snow layers, mob heads, flower pots and cobwebs. They are stopped by signs, banners, doors, fence gates, trapdoors, pressure plates, glass panes and iron bars."
Ref: https://minecraft.gamepedia.com/Rain#Rain_particles

😞 not another list of blocktypes to keep. Why can't minecraft just be consistent?

> Now you can replace this https://github.com/cuberite/cuberite/blob/master/src/Mobs/Enderman.cpp#L210 with your new cWorld::IsWeatherWetAtBlock method.

Will do.

Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -2286,6 +2286,21 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
...
+				Params =
+				{
+					{
+						Name = "Pos",
+						Type = "Vector3i",
+					},
+				},
+				Returns =
+				{
+					Type = "boolean",
```
Seadragon91: Missing additional brackets.

bearbin: Fixed.

src/Chunk.h
```diff
@@ -95,7 +95,7 @@ class cChunk :
 	/** Marks all clients attached to this chunk as wanting this chunk. Also sets presence to cpQueued. */
 	void MarkRegenerating(void);
 
-	/** Returns true iff the chunk has changed since it was last saved. */
+	/** Returns true if the chunk has changed since it was last saved. */
```
madmaxoft: This is not a typo, it is a thing normally used in maths and logic. It means `if and only if`.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -274,9 +274,9 @@ return
 							Type = "boolean",
 						},
 					},
-					Notes = "Returns whether the specified block type can hold snow atop.",
+					Notes = "Returns whether the specified block type lets rain pass through.",
```
madmaxoft: So an `IsRainBlocker` is true if it lets rain pass through, and false if it **blocks** the rain? Where's the logic in that?

src/Simulator/FireSimulator.cpp
```diff
@@ -25,15 +25,14 @@
 
 #define MAX_CHANCE_REPLACE_FUEL 100000
 #define MAX_CHANCE_FLAMMABILITY 100000
+#define CHANCE_BASE_RAIN_EXTINGUISH 0.2
+#define CHANCE_AGE_M_RAIN_EXTINGUISH 0.03
```
madmaxoft: These should get some comments

src/Simulator/FireSimulator.cpp
```diff
@@ -121,29 +142,28 @@ void cFireSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
 		}
 
 		// Burn out the fire one step by increasing the meta:
+		if (!BurnsForever && (y > 0))
+		{
+			a_Chunk->SetMeta(x, y, z, BlockMeta + 1);
```
madmaxoft: What if the BlockMeta is already `0x0f`? This will roll over.

src/World.h
```diff
@@ -817,6 +817,8 @@ class cWorld :
 		return (IsWeatherWet() && !IsBiomeNoDownfall(GetBiomeAt(a_BlockX, a_BlockZ)));
 	}
 
+	virtual bool IsWeatherWetAtBlock(Vector3i a_Pos);
```
madmaxoft: Add doxycomment. Document whether the Y coord must be valid, adapt code to that.

src/Simulator/FireSimulator.cpp
bearbin: There isn't anything else changing the BlockMeta, so that wouldn't be an issue. However I've moved this code block back to where it was before anyway.

bearbin:
I applied your changes @madmaxoft

src/Chunk.h
```diff
@@ -95,7 +95,7 @@ class cChunk :
 	/** Marks all clients attached to this chunk as wanting this chunk. Also sets presence to cpQueued. */
 	void MarkRegenerating(void);
 
-	/** Returns true if the chunk has changed since it was last saved. */
+	/** Returns true iff the chunk has changed since it was last saved. */
```
sphinxc0re: Why not `ifff` ? 😁

bearbin:
What should I do about the global constructor warning for the Vector3i changes?

peterbell10:
Making `Vector3`'s constructors `constexpr` would fix it. Do we have `constexpr` support though?

bearbin:
Should I just disable the warning, if we don't have constexpr support?

src/Entities/Entity.cpp
```diff
@@ -1187,12 +1187,9 @@ void cEntity::TickBurning(cChunk & a_Chunk)
 	}
 
 	// Fire is extinguished by rain
-	if (GetWorld()->IsWeatherWetAt(POSX_TOINT, POSZ_TOINT))
+	if (GetWorld()->IsWeatherWetAtBlock({POSX_TOINT, POSY_TOINT, POSZ_TOINT}))
```
madmaxoft: If the entity falls out of the world or is above the world, this will pass invalid Y coord to the function. Although the above-world doesn't really matter, it still breaks the `IsWeatherWetAtBlock`'s contract about valid coords. The below-world will trigger an assert.

src/Mobs/Enderman.cpp
```diff
@@ -195,37 +195,10 @@ void cEnderman::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	}
 
 	// Take damage when touching water, drowning damage seems to be most appropriate
-	if (CheckRain() || IsSwimming())
+	if (GetWorld()->IsWeatherWetAtBlock({POSX_TOINT, POSY_TOINT, POSZ_TOINT}) || IsSwimming())
```
madmaxoft: Again, missing a check for valid Y coord.

src/World.cpp
```diff
@@ -311,6 +311,30 @@ void cWorld::ChangeWeather(void)
 
 
 
+bool cWorld::IsWeatherWetAtBlock(Vector3i a_Pos)
+{
+	ASSERT(a_Pos.y >= 0);
```
madmaxoft: This assert doesn't match the doxy-comment. Use `cChunkDef::IsValidHeight()`

src/World.h
```diff
@@ -817,6 +817,12 @@ class cWorld :
 		return (IsWeatherWet() && !IsBiomeNoDownfall(GetBiomeAt(a_BlockX, a_BlockZ)));
 	}
 
+	/** Returns true if the specified location has any precipitation (rain or
+	storm), taking into account biomes and any blocks above the specified
```
madmaxoft: Does this include snow, too? Is that what we want? (check wiki)

bearbin:
I have made the changes that you requested, but I would like some more guidance about what to do with the compiler error.

madmaxoft:
I think the cleanest for this PR is to disable the warning.

Another PR could then be made to introduce constexpr support: Could we mark the `Vector3i` constructors as `CONSTEXPR` and `#define` it to nothing in VS2013, and to `constexpr` in all other compilers? Or do we decide to drop VS2013 support altogether?

src/Blocks/WorldInterface.h
```diff
@@ -65,7 +65,7 @@ class cWorldInterface
 
 	virtual void SetTimeOfDay(int a_TimeOfDay) = 0;
 
-	/** Returns true if it is raining, stormy or snowing at the specified location. This takes into account biomes. */
+	/** Returns true if it is raining or storming at the specified location. This takes into account biomes. */
```
madmaxoft: This makes an inconsistent API. `IsWeatherWet` considers snow wet, `IsWeatherWetAt` doesn't.
I think a new set of API functions for the rain-only condition should be made instead. There's little difference between rain and storm (still rain), so it'd make sense to use `IsWeatherRaining___` style for both of these, and keep the `__Wet__` to include the snow.

bearbin: I had looked through the code that uses IsWeatherWetAt, and all of those places only want to know about rain or storm. There isn't anywhere in the code that would want to know if there was any precipitation (including snow) so unless you think such a function would be useful for plugins, I don't see enough reason to change it.

madmaxoft: I actually created the function exactly for the purpose of unifying rain and snow into a single term "wet". If those parts of the code are not interested in snow, they are buggy. I still think the function should be kept with its meaning of "rain or snow".

bearbin: I think my issue is that 'wet' weather in minecraft terms means storm or rain but not snow, as snow does not do many of the things that rain or storm do (put out fire, fill cauldrons). 

What could I call a new function that just takes into account 'minecraft wet' weather, when IsWeatherWet and IsWeatherRain are both taken?

madmaxoft: Oh, I might have made a mistake. `Wet` is really supposed to be only rain / rain in thunderstorm, not snow. For the rain + snow version, there's `Downfall`. Sorry for the confusion.

So yes, use `Wet` and change it to exclude snow (in all instances), document it, too.

bearbin:
I have now updated the documentation.

madmaxoft:
The `cWorld::IsWeatherWet` seems to have inconsistent Lua and C++ documentation.

bearbin:
I'll take a look soon, thanks for letting me know.

src/BlockInfo.h
```diff
@@ -79,6 +82,10 @@ class cBlockInfo
 	inline static bool IsTransparent              (BLOCKTYPE a_Type) { return Get(a_Type).m_Transparent;         }
 	inline static bool IsOneHitDig                (BLOCKTYPE a_Type) { return Get(a_Type).m_OneHitDig;           }
 	inline static bool IsPistonBreakable          (BLOCKTYPE a_Type) { return Get(a_Type).m_PistonBreakable;     }
+	inline static bool IsRainBlocker              (BLOCKTYPE a_Type)
+	{
+		return ((Get(a_Type).m_IsRainBlocker) || Get(a_Type).m_IsSolid);
```
lkolbly: One thought I had was to iterate over all the block infos (in cBlockInfoArray()) and manually set IsRainBlocker for every block that has IsSolid. That way in the future there could be solid blocks that don't block rain (and, more immediately, this method could just return `Get(a_Type).m_IsRainBlocker`).

Regardless, `Get(a_Type).m_IsRainBlocker` doesn't need parentheses around it.

bibo38: I would go a step further and question the whole `cBlockInfo` class. Why not just a set of simple functions?

But for this PR I agree with @lkolbly : Make `IsRainBlocker` and `m_IsRainBlocker` consistent or make `Get` and all `cBlockInfo` variables private, so nobody can accidentially write `Get(mySpecialBlock).m_IsRainBlocker`.

bearbin:
I've addressed the concerns above, and this is now ready to review again.

src/Simulator/FireSimulator.cpp
```diff
@@ -97,20 +99,42 @@ void cFireSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
...
 
+		auto BurnsForever = DoesBurnForever(a_Chunk->GetBlock(x, (y - 1), z));
+		auto BlockMeta = a_Chunk->GetMeta(x, y, z);
+
+		auto Raining = false;
+		for (size_t i = 0; i < ARRAYCOUNT(gCrossCoords); i++)
+		{
+			if (m_World.IsWeatherWetAtBlock(AbsPos + gCrossCoords[i]))
+			{
+				Raining = true;
```
peterbell10: Isn't this just `std::any_of`?

src/BlockInfo.cpp
```diff
@@ -549,53 +556,101 @@ cBlockInfo::cBlockInfoArray::cBlockInfoArray()
 	Info[E_BLOCK_WOOL                         ].m_IsSnowable = true;
 
 	// Nonsolid blocks:
+	Info[E_BLOCK_ACTIVATOR_RAIL               ].m_IsRainBlocker = true;
 	Info[E_BLOCK_ACTIVATOR_RAIL               ].m_IsSolid = false;
+	Info[E_BLOCK_AIR                          ].m_IsRainBlocker = true;
 	Info[E_BLOCK_AIR                          ].m_IsSolid = false;
```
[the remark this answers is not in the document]
bearbin: These are grouped by block type as all solid blocks also block rain particles - if a new solid block is added the corresponding rain blocker part should be added too.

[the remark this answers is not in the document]
peterbell10: Why not just do it as a loop then?  Then there's no doubt that `m_IsSolid` implies `m_IsRainBlocker`.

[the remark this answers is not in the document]
bearbin: That's a much better idea, I'll change it to a loop.

src/Simulator/CMakeLists.txt
```diff
@@ -29,6 +29,10 @@ SET (HDRS
 	VaporizeFluidSimulator.h
 )
 
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+	set_source_files_properties(FireSimulator.cpp PROPERTIES COMPILE_FLAGS "-Wno-error=global-constructors")
+endif()
```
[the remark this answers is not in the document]
bearbin: I disabled the warnings this way because it was what I thought most of the other code used to disable unwanted warnings. In any case, I don't have any experience using pragmas to disable warnings, could you link another place in the code where they are used so I can see what I need to do?

src/World.cpp
```diff
@@ -542,6 +542,30 @@ void cWorld::ChangeWeather(void)
 
 
 
+bool cWorld::IsWeatherWetAtBlock(Vector3i a_Pos)
+{
+	ASSERT(cChunkDef::IsValidHeight(a_Pos.y));
```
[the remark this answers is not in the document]
bearbin: Logically, this function makes no sense being called when outside of the world, as there are no blocks at which the weather could be wet. However, it would be possible to make this return the result of IsWeatherWetAt() if the Y coord is outside of the world. I don't know if that's really sensible or not, do you have an opinion?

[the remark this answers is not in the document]
peterbell10: Maybe the name maybe changed from `Block` to `Pos` or is that too vague?

As for the behaviour, in my opinion:
 * anything above the first rain blocker should be wet, including above the world.
 * anything below the first rain blocker should not be wet, including below the world.

One thing I'm not sure about is: if there are no rain blockers in a column, should positions under the world be considered wet?

src/Simulator/CMakeLists.txt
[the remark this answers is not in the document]
peterbell10: Some of the code does it this way, some with pragmas.  If you look at #4013 though, you see lots of warnings disabled this way become stale.
Here's an example of doing it with pragmas:
https://github.com/cuberite/cuberite/blob/aff140365de993602724f7324e78b6283bf71155/src/Bindings/LuaState.cpp#L2031-L2040

src/Simulator/FireSimulator.cpp
```diff
@@ -97,20 +99,39 @@ void cFireSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
...
 			// The block is no longer eligible (not a fire block anymore; a player probably placed a block over the fire)
 			FLOG("FS: Removing block {%d, %d, %d}",
-				itr->x + a_ChunkX * cChunkDef::Width, itr->y, itr->z + a_ChunkZ * cChunkDef::Width
+				AbsPos.x, AbsPos.y, AbsPos.z
 			);
 			itr = Data.erase(itr);
 			continue;
 		}
 
+		auto BurnsForever = ((y < 0) || DoesBurnForever(a_Chunk->GetBlock(x, (y - 1), z)));
```
peterbell10: This isn't quite right:
 * The check should be `y <= 0` because `y - 1` should be a valid block coord.
 * If the fire is at `y <= 0` then it _shouldn't_ burn forever because there is no block below which burns forever.

src/World.cpp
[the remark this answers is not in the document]
bearbin: Both `Pos` and `Block` seem a bit vague. Maybe `IsWeatherWetAtXYZ`?

Your behaviour seems good, and I think the most sensible option is to consider positions under the world wet if there are no rain blockers above.

src/Simulator/FireSimulator.cpp
peterbell10: Looking at my comment again it's a bit confusing so here's the code I had in mind:
```cpp
auto BurnsForever = (
    (y > 0) &&  // Must have a block below it
    DoesBurnForever(a_Chunk->GetBlock(x, (y - 1), z))  // And that block does burn forever
);
```

bearbin: If you look again at the later commits, I realised that and fixed it :P

bearbin:
The last test failure seems to be a random failure, and the compilation actually worked.

peterbell10:
Looks like "LuaThreadStress-test" hung.  Obviously unrelated so I've just restarted the build.
There might be a deadlock in the test though.

src/World.cpp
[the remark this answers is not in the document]
peterbell10: `IsWeatherWetAtXYZ` sounds all right.  Maybe a name that tells the reader it considers rain blocking would be better but I can't think of anything now.

I tried it out in game and endermen won't teleport if it starts raining while they're below the world which suggests they're not considered wet.  This also matches up with the particle effects.  Are there any other interactions with rain I could check this with?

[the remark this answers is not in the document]
bearbin: The only other places IsWeatherWetAtBlock is used currently is for burning entities. 

If the behaviour there is consistent I'll change the behaviour below the world.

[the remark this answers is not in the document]
peterbell10: Alright, I've had a chance to check and entities do continue to burn below the world.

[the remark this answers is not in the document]
bearbin: OK, great. I'll change the code, and the name.

bearbin:
I would think this is ready for merge now, possibly pending another review if anybody else wants to do one.

Issue #2925, opened 2016-01-30
Mobs die too fast in lava


BitOfAByte:
When a mob is spawned in lava and cannot escape it dies too quickly. A creeper should take 5 instances of damage from lava and then die, but it actually dies in 4. A zombie should take 6 instances of damage but it too dies in 4.

BitOfAByte:
Seems to be related to #2197.

LogicParrot:
I closed #2197 because this is more general.

PR #4020, opened 2017-09-11
Adjust lava damage values

bearbin:
Fixes issue #2925

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1293,7 +1293,7 @@ void cEntity::TickBurning(cChunk & a_Chunk)
 		m_TicksSinceLastFireDamage++;
 		if (m_TicksSinceLastFireDamage >= FIRE_TICKS_PER_DAMAGE)
 		{
-			if (!IsFireproof())
+			if (!IsFireproof() && !HasLava)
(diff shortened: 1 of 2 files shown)


src/Entities/Entity.cpp
```diff
@@ -1293,7 +1293,7 @@ void cEntity::TickBurning(cChunk & a_Chunk)
 		m_TicksSinceLastFireDamage++;
 		if (m_TicksSinceLastFireDamage >= FIRE_TICKS_PER_DAMAGE)
 		{
-			if (!IsFireproof())
+			if (!(IsFireproof() || HasLava))
```
madmaxoft: I think the DeMorgan-ed version would be a bit easier to understand:
```cpp
if (!IsFireproof() && !HasLava)  // Not fireproof, and hasn't taken lava-damage already (#2197)
```

bearbin: Fixed.

bearbin:
The travis bulild actually passed, it just errored when downloading the packages.

madmaxoft:
Then the easiest thing to do is to restart the job, you should have the permissions for that. I did it for you now.

Issue #2920, opened 2016-01-30
Attempting to place a snow tile on a fence uses up blocks


BitOfAByte:
When you attempt to place a snow tile on top of a fence it should just prevent the tile from being placed, but instead it briefly shows the tile where it would be placed and then takes the block from your inventory.

PR #4022, opened 2017-09-12
Snow and rail placement fixes

### LINKED ISSUE #3507: Placing a rail above another rail is possible but shouldn't
Client version:   1.10.2
Server OS:        Linux (Ubuntu 16.04 LTS)
Commit id:        c8b1d8a37d6073786df34836ed1e2c88652ac629

### Expected behavior
You can't place a rail above another rail.

### Actual behavior
If you try to place a rail above another rail the rail above appears very short and drops directly. It shouldn't be possible to even place it.

Try to place a rail above another rail.

peterbell10:
Fixes #3507 and #2920.
Makes me wonder if allowing placement from `GetPlacementBlockTypeMeta` and checking `CanBeAt` ought to be the same logic for all blocks.  Are there any counterexamples?

### DIFF
--- a/src/Blocks/BlockSnow.h
+++ b/src/Blocks/BlockSnow.h
@@ -11,6 +11,11 @@ class cBlockSnowHandler :
 	public cBlockHandler
 {
 public:
+	enum
+	{
+		FullBlockMeta = 7  // Meta value of a full-height snow block
+	};
+
 	cBlockSnowHandler(BLOCKTYPE a_BlockType)
 		: cBlockHandler(a_BlockType)
 	{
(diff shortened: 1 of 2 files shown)


src/Blocks/BlockSnow.h
```diff
@@ -93,6 +98,16 @@ class cBlockSnowHandler :
...
 		return a_Position.y < (cBlockInfo::GetBlockHeight(a_BlockType) * (a_BlockMeta & 7));
 	}
+
+private:
+
+	/** Returns true if snow can be placed on top of a block with the given type and meta. */
+	static bool CanBeOn(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta)
+	{
+		// If block below is snowable, or it is a thin slow block and has a meta of 7 (full thin snow block), say yay
+		return (cBlockInfo::IsSnowable(a_BlockType) || ((a_BlockType == E_BLOCK_SNOW) && (a_BlockMeta == 7)));
```
NiLSPACE: I know this is just copied from above, but preferably we don't want magic numbers. Any chance the meta's could have their own static value or is that overkill?

peterbell10: The value is used in a few different places so I've added a constant `FullBlockMeta` for it.

NiLSPACE:
Perhaps it should be added to https://github.com/cuberite/cuberite/blob/master/src/BlockID.h#L526 ?
Something like `E_META_SNOW_LAYERS_1` or something? There are most likely better names.

peterbell10:
It's just `E_META_SNOW_LAYERS_N = N - 1`.  A function would make more sense but either way I think it's overkill.

src/Blocks/BlockSnow.h
```diff
@@ -91,8 +98,18 @@ class cBlockSnowHandler :
...
+		return a_Position.y < (cBlockInfo::GetBlockHeight(a_BlockType) * (a_BlockMeta & 0x07));
 	}
+
+private:
+
+	/** Returns true if snow can be placed on top of a block with the given type and meta. */
+	static bool CanBeOn(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta)
+	{
+		// If block below is snowable, or it is a thin slow block and is a full thin snow block, say yay
+		return (cBlockInfo::IsSnowable(a_BlockType) || ((a_BlockType == E_BLOCK_SNOW) && (a_BlockMeta == FullBlockMeta)));
```
madmaxoft: Wrap this into a multi-line condition, use indent to indicate the structure.

src/Blocks/BlockSnow.h
```diff
@@ -11,6 +11,8 @@ class cBlockSnowHandler :
 	public cBlockHandler
 {
 public:
+	enum { FullBlockMeta = 7 };
```
madmaxoft: Wrap this into multiple lines, add a comment to the number `// Meta value of a full-height snow block`

madmaxoft:
I don't think we need the layers' metas named.

peterbell10:
I've fixed the indentation.

Issue #1134, opened 2014-06-28
Condense packet sending functions


tigerw:
In my likely flawed opinion, I think that having a one line function in four files for each packet send operation contributes to bloat and bad readability, especially since xoft insists on five lines betwixt declarations.

We should condense everything into a Broadcast/SendPacket function that takes an enumeration of which specific packet. Then, in one function, decide based on that value what relevant cProtocol function to call.

PR #4027, opened 2017-09-14
Add cPluginManager::GenericCallHook

peterbell10:
Replace repetitive CallHook functions with a generic implementation.

Somewhat related to #1134

### DIFF
--- a/src/Bindings/PluginManager.cpp
+++ b/src/Bindings/PluginManager.cpp
@@ -12,13 +12,6 @@
 #include "../IniFile.h"
 #include "../Entities/Player.h"
 
-#define FIND_HOOK(a_HookName) HookMap::iterator Plugins = m_Hooks.find(a_HookName);
-#define VERIFY_HOOK \
-	if (Plugins == m_Hooks.end()) \
-	{ \
-		return false; \
-	}
-
 
 
 
@@ -209,13 +202,15 @@ void cPluginManager::Tick(float a_Dt)
 		ReloadPluginsNow();
 	}
 
-	FIND_HOOK(HOOK_TICK);
-	if (Plugins != m_Hooks.end())
+	auto Plugins = m_Hooks.find(HOOK_TICK);
+	if (Plugins == m_Hooks.end())
 	{
-		for (auto * Plugin : Plugins->second)
-		{
-			Plugin->Tick(a_Dt);
-		}
+		return;
+	}
+
+	for (auto * Plugin : Plugins->second)
+	{
+		Plugin->Tick(a_Dt);
 	}
 }
 
@@ -223,20 +218,16 @@ void cPluginManager::Tick(float a_Dt)
 
 
 
-template <cPluginManager::PluginHook HookName, class ... PmfArgs, class ... RealArgs>
-bool cPluginManager::GenericCallHook(bool (cPlugin::*a_HookFunction)(PmfArgs...), RealArgs & ... a_Args)
+template <class HookFunction>
+bool cPluginManager::GenericCallHook(PluginHook a_HookName, HookFunction a_HookFunction)
 {
-	FIND_HOOK(HookName);
-	VERIFY_HOOK;
-
-	for (auto * Plugin : Plugins->second)
+	auto Plugins = m_Hooks.find(a_HookName);
+	if (Plugins == m_Hooks.end())
 	{
-		if ((Plugin->*a_HookFunction)(a_Args...))
-		{
-			return true;
-		}
+		return false;
 	}
-	return false;
+
+	return std::any_of(Plugins->second.begin(), Plugins->second.end(), a_HookFunction);
 }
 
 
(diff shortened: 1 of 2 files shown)


bearbin:
That's a big saving on lines, great cleanup!

You should set up a [pre-commit hook](https://github.com/cuberite/cuberite/blob/master/CONTRIBUTING.md#pre-commit-hook) so that you don't get CircleCI failures from bad style.

src/Bindings/PluginManager.cpp
```diff
@@ -223,14 +223,15 @@ void cPluginManager::Tick(float a_Dt)
 
 
 
-bool cPluginManager::CallHookBlockSpread(cWorld & a_World, int a_BlockX, int a_BlockY, int a_BlockZ, eSpreadSource a_Source)
+template <cPluginManager::PluginHook HookName, class ... PmfArgs, class ... RealArgs>
```
lkolbly: Do PmfArgs and RealArgs have to be separate? This works for me on cpp.sh:
```
// Example program
#include <iostream>
#include <string>

using namespace std;

template <class ... RealArgs>
bool foo(bool (*fn)(RealArgs...), RealArgs && ... a_Args) {
    return fn(forward<RealArgs>(a_Args)...); // also, std::forward with universal references
}

bool bar(int x) {
    return x > 5;
}

bool bar2(int x, int y) {
    return x > y;
}

int main()
{
    cout << foo(bar, 5) << endl;
    cout << foo(bar2, 5, 4) << endl;
}
```

(for better or worse, I've been learning a lot of C++ lately, now I'm stuck thinking about things like parameter packs and universal references)

peterbell10: It doesn't work where the deduced type is different from the function signature. [For example](http://cpp.sh/4ndb) changing the above to:
``` cpp
bool bar(int& x) {
    return x > 5;
}
```

Failes to compile with this error:
```
 In function 'int main()':
29:23: error: no matching function for call to 'foo(bool (&)(int&), int)'
29:23: note: candidate is:
15:6: note: template<class ... RealArgs> bool foo(bool (*)(RealArgs ...), RealArgs&& ...)
15:6: note: template argument deduction/substitution failed:
29:23: note: inconsistent parameter pack deduction with 'int&' and 'int'
```

lkolbly: That fails because it's passing bar() a rvalue (I think? Or is it an xvalue?), but bar is expecting a lvalue reference.

If you replace bar's literal parameter with a variable, it compiles fine:
http://cpp.sh/27kbz

peterbell10: Compilation fails before it even tries to call bar, the issue is the type deduction.  Perhaps it was a bad example though.

The other way around still fails as well though:
http://cpp.sh/7qkan

peterbell10:
I've removed the pmf and variadics from `GenericCallHook` in favour of a simpler lambda interface.  Not quite the same savings on lines but hopefully more readable this way.

src/Bindings/PluginManager.h
```diff
@@ -444,6 +444,12 @@ class cPluginManager
 
 	/** Returns the folders that are specified in the settings ini to load plugins from. */
 	AStringVector GetFoldersToLoad(cSettingsRepositoryInterface & a_Settings);
+
+	/** Calls a_HookFunction on each plugin registered to the hook HookName.
+	Returns false if the action is to continue or true if the plugin wants to abort.
+	Accessible only from within PluginManager.cpp */
+	template <class HookFunction>
+	bool GenericCallHook(PluginHook a_HookName, HookFunction a_HookFunction);
```
bibo38: could be a `const` function 😼 .

src/Bindings/PluginManager.cpp
```diff
@@ -209,33 +202,45 @@ void cPluginManager::Tick(float a_Dt)
...
-		for (PluginList::iterator itr = Plugins->second.begin(); itr != Plugins->second.end(); ++itr)
-		{
-			(*itr)->Tick(a_Dt);
-		}
+		return;
+	}
+
+	for (auto * Plugin : Plugins->second)
+	{
+		Plugin->Tick(a_Dt);
```
bibo38: Why doesn't this use the `GenericCallHook`?

src/Bindings/PluginManager.h
peterbell10: It could be but the question is should it be?
It can arbitrarily modify the state of plugins and so isn't really `const`. It does depend on whether you think the plugins count as part of the manager though.

src/Bindings/PluginManager.cpp
peterbell10: Because it can't. Tick doesn't return a bool like the other hooks.

src/Bindings/PluginManager.cpp
```diff
@@ -209,33 +202,45 @@ void cPluginManager::Tick(float a_Dt)
...
 	}
 }
 
 
 
 
 
-bool cPluginManager::CallHookBlockSpread(cWorld & a_World, int a_BlockX, int a_BlockY, int a_BlockZ, eSpreadSource a_Source)
+template <class HookFunction>
```
madmaxoft: I think I'd prefer `typename` here; although they are both the same, I use `class` when the template param is supposed to be an actual class, and `typename` when it's supposed to be something else, type or function.

madmaxoft:
Thanks.
Squash-merge if builds succeeds.

PR #4030, opened 2017-09-15
cRoot: Make PollPeriod representation 32 bit

peterbell10:
Should fix the build errors @Bond-009 mentions [here](https://github.com/cuberite/cuberite/commit/0968a4d360d5a28f9d474eb1bf878ed5155b1310).

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -1,4 +1,4 @@
-﻿
+
 #include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
 
 #include "Root.h"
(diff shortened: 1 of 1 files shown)


peterbell10:
Yup, it's removing the BOM.

madmaxoft:
Hmm, jumping through hoops just to make a compiler happy on all platforms. Yup, that's C++ :)
This is getting rather chatty. Honestly, I'd rather just the simple code with a comment and be done with it:
```cpp
Timeout.tv_usec = 100 * 1000;  // 100 msec
```
Any clarity gained by using `std::chrono::milliseconds` is lost through the horrid type names and `using`s.

peterbell10:
Sadly I think you're right.

PR #4033, opened 2017-09-18
Removed UTF-8 BOM

Seadragon91:
Used this command
`find . -type f -exec sed -i -e '1s/^\xEF\xBB\xBF//' {} \;`
to remove all UTF-8 BOM from all files.

### DIFF
--- a/src/Cuberite.vcxproj.user
+++ b/src/Cuberite.vcxproj.user
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <LocalDebuggerWorkingDirectory>../Server</LocalDebuggerWorkingDirectory>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-﻿# This is the top-level CMakeLists.txt file for the Cuberite project
+# This is the top-level CMakeLists.txt file for the Cuberite project
 #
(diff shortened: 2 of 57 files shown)


src/Cuberite.vcxproj.user
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
madmaxoft: Not sure if VS will accept it this way. Did you check?
(The file is only used to provide defaults for when the project is first ever opened)

Seadragon91: Changed settings and the BOM is not readded.

madmaxoft: As I said, this file is only ever used for the defaults. Your settings are then saved to another file.

Seadragon91: I changed the working directory for x64 release / debug builds and VS changed this file. This are the changes:
```
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
    <LocalDebuggerWorkingDirectory>..\Server</LocalDebuggerWorkingDirectory>
    <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
    <LocalDebuggerWorkingDirectory>..\Server</LocalDebuggerWorkingDirectory>
    <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
  </PropertyGroup>
```

madmaxoft: That is somewhat unexpected, I suppose VS behavior with these files changed. Feel free to add it.

madmaxoft: Although I actually wanted to make CMake write this file, since the build can be started out-of-source, which would not use this file.

PR #4036, opened 2017-09-19
Fixed webadmin certificate reading.

### LINKED ISSUE #4035: mbedTLS fails to parse HTTPS cert and key
After upgrading from PolarSSL to mbedTLS, the HTTPS cert and key fail to parse.

https://forum.cuberite.org/thread-3017-post-28589.html#pid28589 contains more info and test data.

madmaxoft:
Fixes #4035

### DIFF
--- a/src/mbedTLS++/CryptoKey.cpp
+++ b/src/mbedTLS++/CryptoKey.cpp
@@ -120,16 +120,19 @@ int cCryptoKey::ParsePublic(const void * a_Data, size_t a_NumBytes)
 int cCryptoKey::ParsePrivate(const void * a_Data, size_t a_NumBytes, const AString & a_Password)
 {
 	ASSERT(!IsValid());  // Cannot parse a second key
+	// mbedTLS requires that PEM-encoded data is passed including the terminating NUL byte,
+	// and DER-encoded data is decoded properly even with an extra trailing NUL byte, so we simply add one to everything:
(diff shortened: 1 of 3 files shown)


madmaxoft:
I've tested the change with both a PEM- and a DER-encoded cert, both work.

peterbell10:
Should the string not be a null terminated already? Could you not just use `CertFile.size() + 1`?

madmaxoft:
The file contents is stored as NUL-terminated, but if it's in DER format, we don't want to include the NUL byte (stupid mbedTLS behavior). So it's either we pass a different size, or push a NUL byte. I chose the pushing because the code looks cleaner to me.

I mentally distinguish between `data()` / `size()` calls for binary data within an `AString`, and `c_str()` / `length()` calls when it's a printable string. While `c_str()` guarantees NUL-termination, `data()` doesn't have to.

peterbell10:
How about callers always include the null and handle different sizes within `cX509Cert::Parse`?

I get you're point but the distinction between `data()` and `c_str()` isn't present in c++11.

madmaxoft:
I'll try if DER parsing fails when an extra NUL byte is added to it. If not, it would make the most sense to simply always include it.

madmaxoft:
Yes, DER parsing works even with the added NUL byte, so I decided to simply force-add it everywhere.

madmaxoft:
Fixed a grammar error (Czenglish).

PR #4043, opened 2017-09-22
APIDump: Check for IsOptional and updated APIDoc

Seadragon91:
Please do not merge yet
  - [x] Added check in APIDump for `IsOptional` and adds optional to param type
  - [x] Updated a few functions that have optional params
  - [ ] Update more functions that have already optional params
  - [ ] Make params optional, e.g. `SetBlock` and `FastSetBlock` that have `BlockMeta`,  make default 0

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -6929,6 +6929,7 @@ These ItemGrids are available in the API and can be manipulated by the plugins, 
 						{
 							Name = "FromBook",
 							Type = "boolean",
+							IsOptional = true,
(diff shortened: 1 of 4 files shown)


madmaxoft:
(edited the description to use GFM checkboxes)

madmaxoft:
Now that I think about it, having an `IsOptional` is not enough, we should also include the default value somehow. And since a default implies optionality, the flag could be removed in favor of a `Default` flag. Does that make sense?

Seadragon91:
Yes using a default value is better. Should I use `nil` for `cClientHandle`, `cPlayer`?

PR #4044, opened 2017-09-22
cWorld: Move most work from Start into the constructor

### LINKED ISSUE #4042: If plugin manipulates world on startup server crashes
Commit id:        e883aa828c304600fcd707b50df9515011311db8

### Expected behavior
Server should not crash.

### Actual behavior
Server crashes.

Load this plugin code:
```
function Initialize(a_Plugin)
	a_Plugin:SetName("Crash")
	cRoot:Get():GetDefaultWorld():SetBlock(0, 100, 0, 1, 0)
	return true
end
```

Ref, forum post:
https://forum.cuberite.org/thread-2937.html

peterbell10:
Fixes #4042. Tested with the script from the issue.

`Start` now does nothing other than starting the world's threads.  This means everything is initialised properly before the plugins are loaded.  Also, as the threads are still not started in the constructor, you keep the guarantee that chunks can't be generated before hooks are registered.

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -120,7 +120,11 @@ void cWorld::cTickThread::Execute(void)
 ////////////////////////////////////////////////////////////////////////////////
 // cWorld:
 
-cWorld::cWorld(const AString & a_WorldName, const AString & a_DataPath, eDimension a_Dimension, const AString & a_LinkedOverworldName) :
+cWorld::cWorld(
+	const AString & a_WorldName, const AString & a_DataPath,
+	cDeadlockDetect & a_DeadlockDetect, const AStringVector & a_WorldNames,
+	eDimension a_Dimension, const AString & a_LinkedOverworldName
+):
 	m_WorldName(a_WorldName),
 	m_DataPath(a_DataPath),
 	m_LinkedOverworldName(a_LinkedOverworldName),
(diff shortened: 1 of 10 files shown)


src/World.cpp
```diff
@@ -146,7 +150,7 @@ cWorld::cWorld(const AString & a_WorldName, const AString & a_DataPath, eDimensi
 	m_LastChunkCheck(0),
 	m_LastSave(0),
 	m_SkyDarkness(0),
-	m_GameMode(gmNotSet),
+	m_GameMode(gmSurvival),
```
madmaxoft: This is an unrelated change, should not be in this PR

peterbell10: It was set in start before, now it is set in the constructor.  This is entirely related.

madmaxoft: Okay, didn't see that.

madmaxoft:
Does it make sense to keep this separate in two commit, or should we squash-merge?

bearbin:
My thoughts: would it compile and run in the intermediate stage, and do the commits solve logically separate issues? In this case, I think it it could go either way - I would lean towards keeping them separate as they wouldn't be any more confusing that way and more history is always good.

madmaxoft:
I think it doesn't make sense to have this in two separate commits, but I could be wrong.

peterbell10:
The first commit does compile and run.  It fixes the issue that `m_ChunkMap` is `nullptr` but doesn't fix the issue entirely because `SetBlock` will try to interact with the storage thread before it's initialised.
They are very closely related though so having a single commit makes sense.

Issue #3863, opened 2017-07-21
Some blocks of multi-block structures stay after explosions


Bond-009:
Client version:   1.12
Server OS:        Linux
Commit id:        approx: 53d5afb83ca5425ad60a9867ec3b0cca0e794ac8

### Expected behavior
The top part destroys together with the lower part.

### Actual behavior
![2017-07-21_12 23 46](https://user-images.githubusercontent.com/21289123/28459936-e64d58ee-6e0f-11e7-953c-17cecbc1fa02.png)
![2017-07-21_13 01 06](https://user-images.githubusercontent.com/21289123/28461774-93cc3d20-6e18-11e7-9c64-5d956f6729c4.png)
![2017-07-21_13 23 47](https://user-images.githubusercontent.com/21289123/28461778-98bbcc38-6e18-11e7-9a2c-efd45574c962.png)

### Steps to reproduce the behavior
Try to explode only the lower part of the block.

If you replicate this there is a 1/4 chance that the tnt will move in the right direction and only explode the lower part.
![2017-07-21_13 04 53](https://user-images.githubusercontent.com/21289123/28461782-9c2e587c-6e18-11e7-842d-82658694fc39.png)

madmaxoft:
I wonder if a similar thing is possible in vanilla. Difficult to test, though.
I suppose a similar thing would happen with other multi-block structures, such as doors.

Bond-009:
I tested some more and it works with beds, doors, snow and ferns (which should be the same as grass)

Seadragon91:
I think for bed the problem is that the cBlockArea doesn't remove block entities.

tigerw:
#4418 probably related too.

tigerw:
Explosions use Block Areas, and Block Areas use FastSetBlock, so unsurprising that nothing updates.

PR #4052, opened 2017-09-27
Fix explosion interaction with block entities.

### LINKED ISSUE #4051: Creeper blowed up and chest and content of chest completely lost
Client version:   1.12.2

### Expected behavior

Expect that some contents of the chests are exist.

### Actual behavior

Creeper blowed up - the content of large chest is completely lost.

```
[12:21:32] WorldLoader(world): Block entity mismatch: block type air (0), type "Chest", at {199, 96, -212}; the entity will be lost.
[12:21:32] WorldLoader(world): Block entity mismatch: block type air (0), type "Chest", at {200, 96, -212}; the entity will be lost.
```

1. Fill your chest with rare items :D
1. Wait next to your chest until it's night
1. Let a creeper blow up your chest
1. Check what items are remained

peterbell10:
Fixes  #4051.

`OffX` and `OffZ` are chunk relative coords but `cBlockEntity::GetPos` is is world coords so `affectedArea.IsInside` didn't work properly.

Edit:  There's a problem with this.  If a chest is in the explosion area but isn't blown up, its items will still drop.
Edit2: I've fixed it by moving the call to destroy into `DoExplosionAt`.

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -452,14 +452,15 @@ void cChunk::WriteBlockArea(cBlockArea & a_Area, int a_MinBlockX, int a_MinBlock
 	}  // for y
 
 	// Erase all affected block entities:
-	cCuboid affectedArea(
-		{OffX, a_MinBlockY, OffZ},
-		{OffX + SizeX - 1, a_MinBlockY + SizeY - 1, OffZ + SizeZ - 1}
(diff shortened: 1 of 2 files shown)


bearbin:
Can you check if this fixes #3863 as well?

peterbell10:
I checked and it doesn't.

PR #4060, opened 2017-10-14
Updated the README

Bond-009:
* Add a curl command for the compile.sh script
* Improve status shields
* Comply with the commonmark spec


README.md
```diff
@@ -29,23 +30,28 @@ This script will download the correct binary from the project site.
...
 
 This script downloads the source code and compiles it. The script is smart enough to notify you of missing dependencies and instructing you on how to install them. The script doesn't work for Windows.
 
+Using curl:
+
+    sh -c "$(curl -sSfL -o - https://compile.cuberite.org)"
+
+Or using wget:
+
     sh -c "$(wget -O - https://compile.cuberite.org)"
```
peterbell10: Is there any benefit from giving two options?  Especially seeing as the easyinstall.sh instructions only give curl.

Bond-009: On macOS wget isn’t installed by default, the opposite is true for some BSD’s and Linux distros.
The easyinstall script itself uses curl, so a wget command would be useless.

peterbell10: Fair enough.

PR #4061, opened 2017-10-14
Fixed FreeBSD build

### LINKED ISSUE #4028: FreeBSD build fails
Unreachable code warning: https://builds.cuberite.org/job/Cuberite-FreeBSD-x64-Master/738/console

Bond-009:
Fixes #4028 (I hope)
I don't know if this is the best solution, but it's better than failing to build on FreeBSD.


madmaxoft:
I think the assert should be inside the `#ifdef`, too, otherwise I'd expect clang to fail debug builds.

Bond-009:
I don't think it will, this is an example where it's done the same way (with the assert outside the `#ifndef`)
https://github.com/cuberite/cuberite/blob/master/src/Entities/ArrowEntity.cpp#L69

madmaxoft:
Okay then, let's stay consistent.

Issue #1401, opened 2014-09-08
C++11 standardisation


tigerw:
Standardise our implementations.
- [x] cTimer should be replaced with `std::chrono`
- [x] Mersenne Twister random should be replaced with `std::mersenne_twister_engine`
- [x] cFastRandom should be replaced with `std::minstd_rand`, with the file being renamed to Random.h
- [x] cMersenneTwister should be merged into Random.h and made to use `std::mt19337`
- [x] Our threading and locking implementation should be replaced with `std::thread`
- [ ] Various `atoi()` calls should be replaced with @worktycho's `StringToInteger<>`, which itself should be made to use `std::stoi` combined with an error handler for overflows, etc.
- [ ] Potentially replace `printf()` calls with `std::cout`
- [x] Replace "ad hoc Functor classes" with lambdas
- [ ] Use std::array wherever possible
- [x] Replace `NULL` with `nullptr` _en masse_
- [x] Replace `&Data[0]` accesses with `.data()` for vectors
- [ ] Replace our cCSLock with `std::lock_guard`

Anything more?

yangm97:
>  Potentially replace `printf()` calls with `std::cout`

Can I do this? The “potentially” there got me uncertain.

madmaxoft:
There are not many places where the server uses `printf` itself. On the other hand, there's a lot of places that use printf-like syntax - the logging, and various text formatting. Problem with replacing that is that the logging goes not only to stdout, but also to a file, and text formatting needs to go to a string, so it's not a direct replacement.

Also, I think the C++ streams syntax is hideous, it's really bad for translations, if nothing else, because it breaks the format string into multiple smaller ones and forces the order of the parameters.

What I think would be an ideal solution would be to have a logger that would support type safety with a syntax like this:
```cpp
int val1 = 42;
AString val2 = "42";
AString blob = cFile::ReadWholeFile("image.png");
LOG() << "Some format string with placeholders: the second value is %2, the first one is %1" << val1 << val2;
LOGD() << "Hex dump: %1" << LogHexDump(blob);
```
If you can work out a system that supports this syntax, and is smart so that for example the hex dump is created only when it's really needed, then by all means go ahead. Possibly even make it a separate logging library that other programs beside Cuberite can include (don't forget about the self-tests, at the very least).

madmaxoft:
Note: https://github.com/fmtlib/fmt looks promising at first quick glance

madmaxoft:
[...]

PR #4065, opened 2017-10-25
Add the fmt library

peterbell10:
Related to discussions in #1401 and #3542
At the moment this is mainly replacing `AppendVPrintf` with `fmt::sprintf` which has some nice benefits:
 * `fmt::ArgList` gives a type safe alternative to varargs.
 * Keeps `printf` syntax so existing calls still work.
 * Fully portable: no need for `SIZE_T_FMT` and friends any more.

I'm in favour of using this to gradually adopt fmt's native python style formatting but it's not required.


src/BrewingRecipes.cpp
```diff
@@ -55,7 +55,7 @@ void cBrewingRecipes::ReloadRecipes(void)
 		AddRecipeFromLine(ParsingLine, LineNum);
 	}  // while (getline(ParsingLine))
 
-	LOG("Loaded " SIZE_T_FMT " brewing recipes", m_Recipes.size());
+	LOG("Loaded %zu brewing recipes", m_Recipes.size());
```
lkolbly: %zu, not {}?

peterbell10: `LOG` still uses the `printf` style formatting to save having to change every single log call.  It could still be changed to `FLOG` though, in which case `{}` would be correct.

lkolbly: Okay, fair. As long as it doesn't break MSVC (I don't have MSVC, so I can't test it).

peterbell10: It uses `fmt::sprintf` which doesn't rely on the standard library so it's fully portable.

src/Generating/PieceGeneratorBFSTree.cpp
```diff
@@ -294,14 +294,14 @@ void cPieceGeneratorBFSTree::PlacePieces(int a_BlockX, int a_BlockZ, int a_MaxDe
...
-	printf("  Connector pool: " SIZE_T_FMT " items\n", a_ConnectorPool.size() - a_NumProcessed);
+	fmt::print("  Connector pool: {} items\n", a_ConnectorPool.size() - a_NumProcessed);
 	size_t idx = 0;
 
 	typedef cPieceGeneratorBFSTree::cFreeConnectors::difference_type difType;
 
 	for (auto itr = a_ConnectorPool.cbegin() + static_cast<difType>(a_NumProcessed), end = a_ConnectorPool.cend(); itr != end; ++itr, ++idx)
 	{
-		printf("    " SIZE_T_FMT ": {%d, %d, %d}, type %d, direction %s, depth %d\n",
+		fmt::print("    {}: {{{}, {}, {}}}, type {}, direction {}, depth {}\n",
```
madmaxoft: This is exactly the reason why I don't like the Python-like placeholders. At least use them as `{0}` etc.

madmaxoft:
I approve this PR except the `{}` formatting, if you change it at least to `{0}`, consider it approved :)

I have a feeling we might have a problem where we send some custom format strings to the library, and those don't expect the curly braces to be processed. Perhaps adding a top-level exception handler for the `FormatError` exception and logging as much info from it as possible before terminating would be smart?

peterbell10:
I might be misunderstanding your second point but I don't think there's any problem.  Everything that takes printf style formatting won't treat curly braces as special characters.  Either way, a top-level exception handler is probably a good idea.

Also for fmt's licence, do I just put a copy in Install/ThirdPartyLicenses/ or is there more to it?

madmaxoft:
To illustrate my second point, consider the following (hypothetical) code:
```
Printf("Some string with {braces} and %d values in it", 1);
```
I assume that such code will break after switching to fmt. Or am I missing something?

madmaxoft:
As for the license, yes, the file in ThirdPartyLicenses is all we've done for all libraries so far. It gets copied into the executable bundle upon build, so it should be ok.

peterbell10:
Everything in `fmt/format.h` such as `fmt::format` will only format braces. To get printf support you also include `fmt/printf.h` which provides separate functions such as `fmt::sprintf` (which is what `Printf` uses) which only format '%'s.

There's no mixing of formatting styles at any point so the example you gave would work fine.

src/Generating/PieceGeneratorBFSTree.cpp
peterbell10: Okay, all the placeholders are numbered now.

bearbin:
@peterbell10 Is this good to merge?

peterbell10:
Maybe it's best to wait until we fix the CI issues so we can be sure the fmt headers don't break the build.

peterbell10:
I realised the `TEST_GLOBALS` versions didn't have the `FLOG` functions. Now that's fixed and it's compiling cleanly with clang 5.0 I'd say this is good to merge.

peterbell10:
> consider it approved :)

I'd forgotten about this.  Sorry for leaving this open so long.

Issue #4024, opened 2017-09-13
Crash when importing singleplayer world


bearbin:
```
[f6d8e6ed55ba10|11:06:51] Preparing spawn (createsting): 4.94% (4/81; 3.67 chunks / sec)
[af3f4edf624155c4|11:06:51] Spawning cSquid #1 at {54.50, 61.00, -48.50}
[af3f4edf624155c4|11:06:51] Spawning cSquid #2 at {54.50, 61.00, -48.50}
[af3f4edf624155c4|11:06:51] Spawning cSquid #3 at {60.50, 61.00, -63.50}
[f6d8e6ed55ba10|11:06:52] Preparing spawn (createsting): 16.05% (13/81; 8.98 chunks / sec)
[4a8701785a947b47|11:06:52] Assertion failed: m_Tags[static_cast<size_t>(a_Tag)].m_Type == TAG_Short, file /home/bearbin/github/cuberite/cuberite/src/WorldStorage/FastNBT.h, line 262
./Cuberite_debug(_Z15PrintStackTracev+0x1f)[0x5633ea379581]
./Cuberite_debug(_ZNK10cParsedNBT8GetShortEi+0x58)[0x5633ea61ac0e]
./Cuberite_debug(_ZN9cWSSAnvil21LoadEntityBaseFromNBTER7cEntityRK10cParsedNBTi+0x1ef)[0x5633ea637109]
./Cuberite_debug(_ZN9cWSSAnvil16LoadSheepFromNBTERSt6vectorISt10unique_ptrI7cEntitySt14default_deleteIS2_EESaIS5_EERK10cParsedNBTi+0x8f)[0x5633ea635a3f]
./Cuberite_debug(_ZN9cWSSAnvil17LoadEntityFromNBTERSt6vectorISt10unique_ptrI7cEntitySt14default_deleteIS2_EESaIS5_EERK10cParsedNBTiPKcm+0x1c24)[0x5633ea632208]
./Cuberite_debug(_ZN9cWSSAnvil19LoadEntitiesFromNBTERSt6vectorISt10unique_ptrI7cEntitySt14default_deleteIS2_EESaIS5_EERK10cParsedNBTi+0xea)[0x5633ea62ae90]
[...]

PR #4076, opened 2017-11-22
Added Github-related rules to Contributing.md

madmaxoft:
These should make it easier for us to review code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,10 +80,10 @@ Github-related conventions
 --------------------------
 The following rules are not strictly enforced, but if you follow them, you will make it easier for us to review your changes.
  - Use good short description for your PRs. `Store Health as float` is good, `Fix #4024` is not good.
-	- If a PR fixes an issue, mention it in the PR description, but **not** in the commits themselves. Typically, we add the line `Fixes #007` to the bottom of the PR description, this makes Github link the PR to the issue, auto-close the issue when the PR is merged, and also is included in the merge message.
-	- Focus on one thing only. Each PR should avoid making unrelated changes - those makes the history more difficult to understand later on, and they make the review a headache.
+ - If a PR fixes an issue, mention it in the PR description, but **not** in the commits themselves. Typically, we add the line `Fixes #007` to the bottom of the PR description, this makes Github link the PR to the issue, auto-close the issue when the PR is merged, and also is included in the merge message.
+ - Focus on one thing only. Each PR should avoid making unrelated changes - those makes the history more difficult to understand later on, and they make the review a headache.
  - Feel free to rebase, amend or otherwise manipulate the commits in the PR (of course, as long as you're the only one working on the PR). We mostly squash-merge PRs, so your internal commit structure will not be important in the end.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -76,6 +76,15 @@ src/CheckBasicStyle.lua 1>&2 -g
 
 Note that the check script is not smart enough to catch everything, so not having any warnings does not necessarily imply that you followed the conventions fully. The other humans working on this will perform more checks before merging.
 
+Github-related conventions
+--------------------------
+The following rules are not strictly enforced, but if you follow them, you will make it easier for us to review your changes.
+ - Use good short description for your PRs. `Store Health as float` is good, `Fix #4024` is not good.
```
bearbin: I agree with having descriptive commit titles, but is there any reason to discourage adding issue references to commit descriptions?

Cl1608Ho: For me it seems that naming a commit "fixes #something" rather than "store health as float" only makes sense when viewed on github, and not anywhere else. When you just take the git repo, you should have a description about what the commit is doing not just a reference to any bug tracker. Imo.

bearbin: I absolutely agree that all commits should have descriptive titles, but if there is a relevant issue, I think it really ought to be mentioned. Just to be clear, here is my idea of an 'ideal' commit message:

    Store entity health as float rather than short
    
    Originally changed in vanilla 1.x. 
    Fixes #4024

Pokechu22: I concur, the title should describe it on its own, but the description should include "fixes #id" -- among other things, the "fixes" magic word actually has [special behavior](https://help.github.com/articles/closing-issues-using-keywords/) (the issue is automatically closed when merged) that is useful to preserve.

Cl1608Ho: I agree, but most people don't read commit descriptions (so you don't really notice it but github does? (or does github? Idk)) and even less people use commit descriptions....

madmaxoft:
I've changed the wording slightly to avoid doubt.

My intention is to have short descriptive titles for PRs; their descriptions (the first comment in the PR) should by all means be as long and as detailed as needed.

Note the distinction between PR and commit titles / descriptions. In my opinion it is not good to mention issues directly from commit messages, because Github auto-links the commit into the issue, and then when someone rebases, squash-merges or otherwise deletes the commit, it gets messy - links to commits are not valid anomore, or even worse, the auto-link creates multiple mentions in the issue.
On the other hand, PRs should link to the issue in their description, this makes the github auto-close machinery work, and it creates exactly one mention, as naturally expected.

peterbell10:
I agree; linking issues from commits is too spammy, especially when rebasing.

It could still be useful to link from the squashed commit though.  That way you can go straight from git history to the issue.

madmaxoft:
Perhaps an example should be added to the docs to better illustrate this?

peterbell10:
I thought the default was a list of all the merged commits like in 0dd172b80ffc949cf87119ad2003589d7eb82c46.

Either way I don't think it would be that relevant to someone who isn't doing the merge themself so I'm not sure it should be in this file.

Issue #4079, opened 2017-11-26
Certain block entities cause memory leak


mathiascode:
Client version:   1.12.2
Server OS:        Linux
Commit id:        (Found at the top of the server's console output)

### Expected behavior
Memory leak doesn't occur when placing chests, hoppers, droppers and trapped chests (with WorldEdit?).

### Actual behavior
Memory is never freed.

### Steps to reproduce the behavior
1. Install WorldEdit.
2. Type `//sphere chest 20` in chat
3. RAM usage has increased a lot

peterbell10:
On the linked PR there is still a large increase in RAM usage and I'm pretty confident it isn't a leak because RAM usage goes back to zero after issuing the stop command.

I think it just genuinely takes a lot of RAM.  Some back of the envelope calculations:
 * A sphere of radius 20 is ~35,000 blocks
 * Each chest has has a 3x9 inventory or 27 `cItem` instances.
 * On my machine a `cItem` is 192 bytes.
 * Therefore each sphere has ~173MB of `cItem`s.

Then I think the sphere will exist twice, once in the actual world and once in WorldEdit's undo stack.  Clearly using hundreds of megabytes for empty chests is hugely wasteful though.  It might be worth looking into lazily allocating the chest's storage or reducing the size of an empty `cItem`.

peterbell10:
I'm going to look into making `cItemGrid` allocate lazily.  That way chests won't take any memory until items are put into them.

PR #4080, opened 2017-11-27
Manage block entity lifetime with unique_ptr

peterbell10:
Looking at the block area code with #4079 in mind I found several suspicious looking pieces of code like this:
https://github.com/cuberite/cuberite/blob/0dd172b80ffc949cf87119ad2003589d7eb82c46/src/Chunk.cpp#L488-L494
where block entities are being removed but not deleted.
Moving to smart pointers seemed easier in the long run than going through all of the block entity logic with a fine tooth comb.


peterbell10:
Is there any interest in this?

tigerw:
yes there is

src/ChunkDef.cpp
```diff
@@ -0,0 +1,14 @@
...
+#include "ChunkDef.h"
+#include "BlockEntities/BlockEntity.h"
+
+
+
+
+
+void sBlockEntityDeleter::operator () (cBlockEntity * a_BlockEntity) const NOEXCEPT
+{
+	delete a_BlockEntity;
```
tigerw: A custom deleter?

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -985,14 +983,14 @@ cBlockEntity * cWSSAnvil::LoadBeaconFromNBT(const cParsedNBT & a_NBT, int a_TagI
 		LoadItemGridFromNBT(Beacon->GetContents(), a_NBT, Items);
 	}
 
-	return Beacon.release();
+	return OwnedBlockEntity{ Beacon.release() };
```
tigerw: Can't we move it out?

src/BlockEntities/BlockEntity.cpp
```diff
@@ -75,36 +75,36 @@ bool cBlockEntity::IsBlockEntityBlockType(BLOCKTYPE a_BlockType)
...
-		case E_BLOCK_HOPPER:        return new cHopperEntity      (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_JUKEBOX:       return new cJukeboxEntity     (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_LIT_FURNACE:   return new cFurnaceEntity     (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_MOB_SPAWNER:   return new cMobSpawnerEntity  (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_NOTE_BLOCK:    return new cNoteEntity        (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_SIGN_POST:     return new cSignEntity        (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_TRAPPED_CHEST: return new cChestEntity       (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
-		case E_BLOCK_WALLSIGN:      return new cSignEntity        (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World);
+		case E_BLOCK_BEACON:        return OwnedBlockEntity{new cBeaconEntity      (a_BlockType, a_BlockMeta, a_BlockX, a_BlockY, a_BlockZ, a_World)};
```
tigerw: Consider cpp14::make_unique?

tests/LuaThreadStress/Stubs.cpp
```diff
@@ -284,9 +284,9 @@ bool cBlockHandler::IsInsideBlock(Vector3d a_Position, const BLOCKTYPE a_BlockTy
 
 
 
-cBlockEntity * cBlockEntity::CreateByBlockType(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, int a_BlockX, int a_BlockY, int a_BlockZ, cWorld * a_World)
+OwnedBlockEntity cBlockEntity::CreateByBlockType(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, int a_BlockX, int a_BlockY, int a_BlockZ, cWorld * a_World)
 {
-	return nullptr;
```
tigerw: I would argue nullptr conveys intention a little better

src/BlockArea.cpp
```diff
@@ -673,36 +673,22 @@ void cBlockArea::Crop(int a_AddMinX, int a_SubMaxX, int a_AddMinY, int a_SubMaxY
...
+			auto Pos = be->GetPos();
+			if (CropBox.IsInside(Pos))
 			{
 				// The block entity is within the new coords, recalculate its coords to match the new area:
-				posX -= a_AddMinX;
-				posY -= a_AddMinY;
-				posZ -= a_AddMinZ;
-				be->SetPos(posX, posY, posZ);
-				m_BlockEntities->insert({MakeIndex(posX, posY, posZ), std::move(be)});
+				Pos -= AddMin;
```
tigerw: operator overloads are so satisfying :D

src/BlockArea.cpp
```diff
@@ -673,36 +673,22 @@ void cBlockArea::Crop(int a_AddMinX, int a_SubMaxX, int a_AddMinY, int a_SubMaxY
...
 	{
-		auto maxX = m_Size.x - a_SubMaxX;
-		auto maxY = m_Size.y - a_SubMaxY;
-		auto maxZ = m_Size.z - a_SubMaxZ;
+		const Vector3i AddMin{ a_AddMinX, a_AddMinY, a_AddMinZ };
+		const cCuboid CropBox{ AddMin, m_Size - Vector3i{a_SubMaxX, a_SubMaxY, a_SubMaxZ} };
 
 		// Move and crop block Entities:
 		cBlockEntities oldBE;
 		std::swap(oldBE, *m_BlockEntities);
```
tigerw: I admit I don't know how BlockAreas manage block entities so I can't make any suggestions. Perhaps @madmaxoft can shed some light; why are block entities being deleted here? I thought they were World-owned?

src/ChunkDef.h
```diff
@@ -31,7 +31,13 @@ class cChunkCoords;
 
 typedef std::unique_ptr<cEntity> OwnedEntity;
 typedef std::vector<OwnedEntity> cEntityList;
-typedef std::map<int, cBlockEntity *> cBlockEntities;
+
+struct sBlockEntityDeleter
+{
+	void operator () (cBlockEntity * a_BlockEntity) const NOEXCEPT;
+};
```
madmaxoft: Why exactly do we need this custom deleter, that does exactly the same as the regular deleter?

I assume it's the same issue I've come across with `unique_ptr`: That classes having them as members then need their destructor defined in a place where the pointed-to object is defined as well. This is usually better solved by providing an empty destructor in the class's cpp file, rather than having a custom deleter.

src/ChunkDef.h
```diff
@@ -31,7 +31,13 @@ class cChunkCoords;
...
+typedef std::unique_ptr<cBlockEntity, sBlockEntityDeleter> OwnedBlockEntity;
+typedef std::map<int, OwnedBlockEntity> cBlockEntities;
```
tigerw: I wonder if the ordered-ness of a map matters here?

src/ChunkDef.h
tigerw: I thought you might be able to shed some light, the original sBlockEntityDeleter did appear in BlockArea code XD (line 2733)

src/BlockArea.cpp
madmaxoft: BEs in a `cBlockArea` are owned by the `cBlockArea`. They are clones of the ones in the `cWorld` from which they were read (if at all), and when writing to a `cWorld`, they are again cloned.

src/BlockEntities/BlockEntity.cpp
madmaxoft: Can't with the custom deleter. Another point to get rid of it, if possible.

src/ChunkDef.h
peterbell10: `delete be;` requires a full definition of `cBlockEntity` whereas `sBlockEntityDeleter{}(be)` only requires `cBlockEntity` be forward declared.

src/BlockArea.cpp
tigerw: Thanks. Is it a possibility to use the world ones directly and not clones? At a minimum there shouldn't be threading issues.

src/ChunkDef.h
madmaxoft: Yes, but on the other hand it keeps us from using `make_unique` and requires the conversions between `unique_ptr` and `OwnedBlockEntity`.

src/BlockArea.cpp
madmaxoft: No, because you don't expect that changing a BE in a BlockArea to change a BE in the World. There's no COW.

src/ChunkDef.h
madmaxoft: @peterbell10 If you used a regular default-deleter `unique_ptr`, how much would it affect the rest of the code? How many places (order of 1, 10, 100?) would need changing?

tigerw:
it's... beautiful...

Issue #4031, opened 2017-09-16
Discontinue support for VS2013


peterbell10:
### Proposal
Cuberite should discontinue support for Visual Studio 2013 and require at least Visual Studio 2015.

### Motivation
In #3856 @madmaxoft says:
> We decided on C++11, which is what VS2013 uses, so that's what we stick with for now.

But VS2013 lacks many [C++11 features][MSDN]:

* `noexcept`
* `constexpr`
* magic statics: i.e. thread safe `static` initialisation
* ref-qualifiers: i.e. the ability to overload member functions by value category.
e.g. `Copy()` members could move when called on an r-value.
* attributes: e.g. `NORETURN` -> `[[noreturn]]`
* `thread_local`:  See [FastRandom.cpp][1] for workarounds needed today
* `= default` for move construction and move-assignment
* user defined literals: e.g.  `std::chrono::milliseconds{100}` could become `100_ms`
* unicode string literals: e.g. from [ChatColor.cpp][2]
    ```cpp
    const char * cChatColor::Color = "\xc2\xa7"; // or in other words: "§" in UTF-8
    // Becomes
    const char * cChatColor::Color = u8"§";
    ```
* `char{n}_t`
* `inline namespace`
* inheriting constructors
* `alignof` and `alignas`
* extended `sizeof`
* `std::quick_exit`

And that's not mentioning the compiler bugs that make some supported C++11 features hard to use.

[xoft]:https://github.com/cuberite/cuberite/pull/3856#issuecomment-316445731
[MSDN]:https://msdn.microsoft.com/en-us/library/hh567368.aspx
[1]:https://github.com/cuberite/cuberite/blob/ad33a2d866b0b76dd64a965163b51fd858f33c16/src/FastRandom.cpp#L20-L40
[2]:https://github.com/cuberite/cuberite/blob/ad33a2d866b0b76dd64a965163b51fd858f33c16/src/ChatColor.cpp#L5

peterbell10:
An important note is that most modern C++ libraries require VS2015.
I was looking into implementing NBT tags in `cItem` (for #3682) and the cleanest way I can think of is to use a variant.  However, with the exception of boost, all the libraries I can find require VS2015.

madmaxoft:
[...]

PR #4085, opened 2017-12-04
Fixed Clang 5.0 compile errors

bibo38:
While trying to compile the Cuberite code with Clang 5.0 I've got mostly errors of `NULL` being used instead of `nullptr`.
This commit fixed all errors.

### DIFF
--- a/src/Bindings/CMakeLists.txt
+++ b/src/Bindings/CMakeLists.txt
@@ -168,7 +168,7 @@ set_source_files_properties(${BINDING_OUTPUTS} PROPERTIES GENERATED TRUE)
 set_source_files_properties(${CMAKE_SOURCE_DIR}/src/Bindings/Bindings.cpp PROPERTIES COMPILE_FLAGS -Wno-error)
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-	set_source_files_properties(Bindings.cpp PROPERTIES COMPILE_FLAGS ${COMPILE_FLAGS} "-Wno-old-style-cast -Wno-missing-prototypes")
+	set_source_files_properties(Bindings.cpp PROPERTIES COMPILE_FLAGS "-Wno-old-style-cast -Wno-missing-prototypes -Wno-zero-as-null-pointer-constant")
 endif()
 
 if(NOT MSVC)
--- a/src/Bindings/CMakeLists.txt
+++ b/src/Bindings/CMakeLists.txt
@@ -167,8 +167,11 @@ set_source_files_properties(${BINDING_OUTPUTS} PROPERTIES GENERATED TRUE)
 
 set_source_files_properties(${CMAKE_SOURCE_DIR}/src/Bindings/Bindings.cpp PROPERTIES COMPILE_FLAGS -Wno-error)
 
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-	set_source_files_properties(Bindings.cpp PROPERTIES COMPILE_FLAGS "-Wno-old-style-cast -Wno-missing-prototypes -Wno-zero-as-null-pointer-constant")
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+	if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5) # Workaround for VERSION_GREATER_EQUAL, which is only supported on CMake 3.7+
+		set(ADDITIONAL_FLAGS "-Wno-zero-as-null-pointer-constant")
+	endif()
+	set_source_files_properties(Bindings.cpp PROPERTIES COMPILE_FLAGS "-Wno-old-style-cast -Wno-missing-prototypes ${ADDITIONAL_FLAGS}")
 endif()
 
 if(NOT MSVC)
--- a/src/Protocol/MojangAPI.cpp
+++ b/src/Protocol/MojangAPI.cpp
@@ -5,8 +5,6 @@
 
 #include "Globals.h"
 #include "MojangAPI.h"
-#include "SQLiteCpp/Database.h"
-#include "SQLiteCpp/Statement.h"
 #include "../IniFile.h"
 #include "json/json.h"
 #include "mbedTLS++/BlockingSslClientSocket.h"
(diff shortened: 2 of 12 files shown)


src/main.cpp
```diff
@@ -72,6 +72,12 @@ bool cRoot::m_RunAsService = false;
 
 
 #ifndef _DEBUG
+// Because SQLiteCpp uses NULL instead of nullptr, we need to disable the Clang warning for std::signal here
```
peterbell10: I assume there's no reason for SQLiteCpp to be here.

peterbell10:
I don't know why it isn't linking with github properly but this breaks older clang versions:
https://travis-ci.org/cuberite/cuberite/builds/311506351

src/main.cpp
bibo38: yep, just a copy-paste failure 👍

src/Bindings/CMakeLists.txt
```diff
@@ -168,7 +168,7 @@ set_source_files_properties(${BINDING_OUTPUTS} PROPERTIES GENERATED TRUE)
 set_source_files_properties(${CMAKE_SOURCE_DIR}/src/Bindings/Bindings.cpp PROPERTIES COMPILE_FLAGS -Wno-error)
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-	set_source_files_properties(Bindings.cpp PROPERTIES COMPILE_FLAGS ${COMPILE_FLAGS} "-Wno-old-style-cast -Wno-missing-prototypes")
+	set_source_files_properties(Bindings.cpp PROPERTIES COMPILE_FLAGS "-Wno-old-style-cast -Wno-missing-prototypes -Wno-zero-as-null-pointer-constant")
```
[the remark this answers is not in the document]
bibo38: Because it is irritating. You may think, that it resolves to `-Wno-error`. But actually it resolves to nothing :trollface: , because there is no variable named `COMPILE_FLAGS`. It is only a property, which you can store in a variable via [get_target_property](https://cmake.org/cmake/help/latest/command/get_target_property.html).

bibo38:
@peterbell10 Oh, nice. I thought, that it only builds the branch after the PR has been reviewed (and maybe automerges it afterwards).
Now the branch should also build on older Clang versions.

src/Protocol/MojangAPI.cpp
```diff
@@ -15,6 +13,19 @@
 #include "../OSSupport/IsThread.h"
 #include "../Root.h"
 
+// Because SQLiteCpp uses NULL instead of nullptr, we need to disable the Clang warning here
+#if __clang_major__ >= 5
+	#pragma clang diagnostic push
+	#pragma clang diagnostic ignored "-Wzero-as-null-pointer-constant"
```
peterbell10: This works for now but I think it would be better if the libraries were included as system headers.

bibo38: It would be even better, if we "fix" the SQLiteCpp project (You can detect, if `nullptr` is supported by CMake and inject a header redirecting `nullptr` to `NULL`).

And I'm not the biggest fan of the `isystem` switch. See the explanation on [Reddit](https://www.reddit.com/r/cpp/comments/5arvg9/watch_out_qt_headers_suppress_warnings_with/), why this switch could harm.

peterbell10: Thanks for the link, very enlightening.

bibo38:
@peterbell10 Submitted a PR to SQLiteCpp.

peterbell10:
I just noticed, according to their [README][SQLCppReadme], SQLiteCpp doesn't officially support VS2013. So without #4031 it might be non-trivial to update to the latest version.  

[SQLCppReadme]:https://github.com/SRombauts/SQLiteCpp/blob/master/README.md

bibo38:
@peterbell10 Done ;). But the `#pragma clang` in `main.cpp` is caused by the Standard-C-Library macro, so it need to stay.

bibo38:
> I just noticed, according to their README, SQLiteCpp doesn't officially support VS2013

Just because the tests are being executed with VS2015, this doesn't mean, that VS2013 isn't supported. Just look at the compiler section:

>And the following IDEs/Compilers
>
>   - GCC 4.8.4, 4.9.3, 5.3.0 and 6.1.1 (C++03, C++11, C++14, C++1z)
>   - Clang 3.5 and 3.8
>   - Xcode 8
>   - Visual Studio Community 2015
>   - Eclipse CDT under Linux

So it tests against C++03 (with GCC 4.8.4). And VS2013 fully supports C++03. Furthermore the CMake buildfiles doesn't even require a C++11 compatible compiler. The C++03 compability was the hardest challenge for my PR (and to leave the CMake version as low as possible).

peterbell10:
>it tests against C++03 (with GCC 4.8.4). And VS2013 fully supports C++03

Being standard compliant isn't necessary or sufficient to compiling under MSVC :).  
Now looking at the actual code I see it's probably fine though.

bibo38:
> Being standard compliant isn't necessary or sufficient to compiling under MSVC :).

Any examples? I've only got a compiler bug when compiling some C++11 code. But this was fixed in a VS update.

bibo38:
@peterbell10 The official SQLiteCpp repository now uses `nullptr`. Could you update the SQLiteCpp repository of Cuberite, so that we don't need a seperate branch anymore?

peterbell10:
>Any examples?

Off the top of my head, MSVC doesn't have full two-phase name lookup support and that's C++98.
Though It wasn't meant to be specific to MSVC: I don't think there's any compiler that is 100% conformant as that would require no bugs at all.

>The official SQLiteCpp repository now uses nullptr.

I've rebased the changes from `RemoveGoogleTest` onto the latest version in the `RemoveGoogleTest2` branch: Cuberite/SQLiteCpp@df5e81dbe48d7a9f7e2661acc8dda918d2b6d087

bearbin:
It seems that this hasn't fixed all of the issues, clang5 still fails on the jsoncpp library.

peterbell10:
Also the `__clang_major__` checks don't work on the [AppleClang build][1] which has different version numbers.

[1]:https://travis-ci.org/cuberite/cuberite/jobs/319680211

PR #4086, opened 2017-12-05
cWindow: Fix slot area priority when double click stacking.

### LINKED ISSUE #4084: Double click to stack items doesn't prioritize items in crafting table/grid
Client version:   1.12.2
Server OS:        Windows
Commit id:        0dd172b80ffc949cf87119ad2003589d7eb82c46

### Expected behavior
With stack-able items in crafting table/grid, double click same items in inventory, stacks the item from crafting table/grid and then inventory.
E.g., I have 4 cobble stone in crafting table/grid, a stack of 60 and a stack of 64 cobble stones in inventory. After double click the stack of 60 cobble stones, The stack becomes 64, which takes 4 from crafting table/grid.

### Actual behavior
I have 4 cobble stone in crafting table/grid, a stack of 60 and a stack of 64 cobble stones in inventory. After double click the stack of 60 cobble stones, The stack becomes 64, which takes 4 from the other stack. Unfortunately the client would assume the items in crafting table/grid stacked away, they temporarily disappears, and drops on ground after closing the window.

As described above.

peterbell10:
Fixes #4084

This gives the same behaviour as my 1.12.2 client. Not sure why it was so complicated before.


madmaxoft:
If the behavior changed, I suppose the doxy-comment in the header file should be changed as well.

madmaxoft:
I assume older clients prioritized the slot area that was dblclicked, and this behavior has changed recently to prioritize the crafting grid. Still not sure about all the variants, can you test for the same behavior in the negative test cases? Such as dblclicking on a main inventory item while the same item is in the crafting grid and the hotbar, or dblclicking the hotbar, etc. Somewhere where one would expect the stack to be taken not from the crafting grid, but from the dblclicked location instead.

peterbell10:
It's been a while since I tested it but I did check with several different UIs including normal inventory, work bench and chests.  I looked at the behaviour for all different slot areas as well and all of them seemed to just go in order.

PR #4094, opened 2017-12-18
Smelting Gives Experience

### LINKED ISSUE #3517: No xp from smelting stuff
Client version:   1.8.9
Server OS:         Linux
Commit id:        c8b1d8a37d6073786df34836ed1e2c88652ac629

### Expected behavior
Get experience by manually unloading the furnace.

### Actual behavior
You don't get experience by manually unloading the furnace.

Smelt something

sweetgiorni:
Smelting now gives experience as described by the rules on https://minecraft.gamepedia.com/Smelting.
Fixes  #3517


sweetgiorni:
Yeah, that makes sense. That'd be more organized and the experience values would be easy to customize.
How about this - I can change the line parsing in FurnaceRecipe.cpp to accept either two or three arguments. The optional third argument would be the exp value. I don't know if it's the best syntax, but the easiest way to implement it would be to accept a recipe in the form of 
`Beef = Steak = .35`
Add the corresponding variable to cRecipe and it will be well integrated.

Thoughts?

peterbell10:
Sounds good but I'd prefer something other that `=` as the separator as it's not really the same as the current meaning.  How about `$`?  e.g. `Beef = Steak $ 0.35`

sweetgiorni:
Much cleaner now. I couldn't find an existing StringToFloat function so I just added one...
For some reason, when GitHub previews furnace.txt it shows the exp values all staggered and it's ugly and I hate it. It doesn't do that in any of my text editors.

Cl1608Ho:
It seems Travis is failing because you didn't document that api method you created :)

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -169,6 +169,25 @@ bool cFurnaceEntity::ContinueCooking(void)
...
 }
 
+int cFurnaceEntity::GetReward(void)
+{
+	float RawReward = m_CurrentRecipe->Reward;
+	float Multiplier = GetOutputSlot().m_ItemCount;
+	if ((RawReward != 0.0) && (Multiplier != 0.0))
+	{
+		RawReward *= Multiplier;
+		int Reward = int(RawReward);
```
madmaxoft: Don't use C-style casts. Use our functions `FloorC` or `CeilC` instead, then you can even use `auto` for the type.

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -169,6 +169,25 @@ bool cFurnaceEntity::ContinueCooking(void)
 	return m_IsCooking;
 }
 
+int cFurnaceEntity::GetReward(void)
+{
+	float RawReward = m_CurrentRecipe->Reward;
+	float Multiplier = GetOutputSlot().m_ItemCount;
+	if ((RawReward != 0.0) && (Multiplier != 0.0))
```
madmaxoft: No need for this condition - if either is zero, the result will still be zero

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -169,6 +169,25 @@ bool cFurnaceEntity::ContinueCooking(void)
...
+	{
+		RawReward *= Multiplier;
+		int Reward = int(RawReward);
+		float Remainder = RawReward - Reward;
+		// Remainder is used as the percentage chance of getting an extra exp point
+		if ((Remainder != 0.0) && (GetRandomProvider().RandBool(Remainder)))
```
madmaxoft: No need for the zero check in the condition.

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -169,6 +169,25 @@ bool cFurnaceEntity::ContinueCooking(void)
 	return m_IsCooking;
 }
 
+int cFurnaceEntity::GetReward(void)
+{
+	float RawReward = m_CurrentRecipe->Reward;
```
madmaxoft: Doesn't `m_CurrentRecipe` get reset if the furnace finishes cooking completely (input slot gets empty)?
At the very least this function should check if `m_CurrentRecipe` is valid, and short-circuit return zero when not.

madmaxoft:
I'm not sure if the furnace.txt format is good. What if I changed these lines:
```
ChainmailBoots      = IronNugget  $ 0.1
ChainmailChestplate = IronNugget  $ 1.0
```
(Note the different experience values)
What happens then? Does the furnace remember which recipe was used for smelting the result? What if I feed it a few chainmail boots and then a few chestplates, and only then collect the nuggets?

I think the entire reward system should be reworked, so that each finished smelting operation updates an internal reward counter, which gets remembered persistently (does Vanilla have something like this in the block entity NBT data?) and when collecting the results, the remembered reward counter should be used to dispense experience and then be reset.

src/BlockEntities/FurnaceEntity.h
```diff
@@ -84,6 +84,9 @@ class cFurnaceEntity :
 	/** Returns true if there's time left before the current fuel is depleted */
 	bool HasFuelTimeLeft(void) const { return (GetFuelBurnTimeLeft() > 0); }
 
+	/** Calculates and returns the experience reward for the current recipe */
+	int GetReward(void);
```
madmaxoft: This adds the function into the plugin interface (it is in the `tolua` block). You will need to add the function documentation to the API description files in ~/Server/Plugins/APIDump/Classes/*.lua. See the ~/Server/Plugins/APIDump/APIDesc.lua file for details on the format that is understood by the system.

sweetgiorni:
It took a few months, but there you go. I may have messed something up with all the commits...

src/StringUtils.h
```diff
@@ -146,8 +146,8 @@ extern AStringVector MergeStringVectors(const AStringVector & a_Strings1, const
 /** Concatenates the specified strings into a single string, separated by the specified separator. */
 extern AString StringsConcat(const AStringVector & a_Strings, char a_Separator);
 
-
-
+/** Converts a string into a double. Returns false if the conversion fails. */
+extern bool StringToFloat(const AString & a_String, float & a_Num);
```
bearbin: The missing blank lines here should really be restored.

src/UI/SlotArea.cpp
```diff
@@ -1932,6 +1940,16 @@ void cSlotAreaFurnace::OnSlotChanged(cItemGrid * a_ItemGrid, int a_SlotNum)
 
 void cSlotAreaFurnace::HandleSmeltItem(const cItem & a_Result, cPlayer & a_Player)
 {
+	if (& a_Player == m_Furnace->GetLastSmelter())
+	{
+		int Reward = m_Furnace->GetReward();
+		LOG("Reward %d\n", Reward);
+		if (Reward)
```
bearbin: I think this could better be put as `if (Reward > 0)`

src/UI/SlotArea.cpp
```diff
@@ -1932,6 +1940,16 @@ void cSlotAreaFurnace::OnSlotChanged(cItemGrid * a_ItemGrid, int a_SlotNum)
 
 void cSlotAreaFurnace::HandleSmeltItem(const cItem & a_Result, cPlayer & a_Player)
 {
+	if (& a_Player == m_Furnace->GetLastSmelter())
+	{
+		int Reward = m_Furnace->GetReward();
+		LOG("Reward %d\n", Reward);
```
bearbin: This should be a LOGD, if it's needed at all, we don't need to log an event every time a player takes an item out of a furnace.

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -173,6 +175,23 @@ bool cFurnaceEntity::ContinueCooking(void)
 
 
 
+int cFurnaceEntity::GetReward(void)
```
bearbin: The name of this function would be better as `GetAndResetReward`, it's not obvious that the reward counter is reset after it is called.

src/BlockEntities/FurnaceEntity.cpp
```diff
@@ -173,6 +175,23 @@ bool cFurnaceEntity::ContinueCooking(void)
...
+{
+	int Reward = FloorC(m_RewardCounter);
+	float Remainder = m_RewardCounter - static_cast<float>(Reward);
+	// Remainder is used as the percentage chance of getting an extra xp point
```
bearbin: `percentage` here is incorrect

src/UI/SlotArea.cpp
```diff
@@ -1932,6 +1940,16 @@ void cSlotAreaFurnace::OnSlotChanged(cItemGrid * a_ItemGrid, int a_SlotNum)
 
 void cSlotAreaFurnace::HandleSmeltItem(const cItem & a_Result, cPlayer & a_Player)
 {
+	if (& a_Player == m_Furnace->GetLastSmelter())
```
bearbin: You should remove the space between `&` and `a_Player`.

sweetgiorni:
Fixed all that. 
For some reason I was under the impression that locking XP to the player who earned it is a vanilla feature, but further testing proved that is isn't (Although I think it should be!) I removed that, but I did leave in the GetLastSmelter api since it's already there.

src/StringUtils.h
```diff
@@ -146,6 +146,9 @@ extern AStringVector MergeStringVectors(const AStringVector & a_Strings1, const
 /** Concatenates the specified strings into a single string, separated by the specified separator. */
 extern AString StringsConcat(const AStringVector & a_Strings, char a_Separator);
 
+/** Converts a string into a double. Returns false if the conversion fails. */
```
bearbin: 'double'???

Server/furnace.txt
```diff
@@ -42,74 +43,73 @@
...
+IronLeggings        = IronNugget                         $ 0.1
+IronPickaxe         = IronNugget                         $ 0.1
+IronShovel          = IronNugget                         $ 0.1
+IronSword           = IronNugget                         $ 0.1
+LapisOre            = LapisLazuli                        $ 0.2
+LightBlueTerracotta = LightBlueGlazedTerracotta          $ 0.1
+LightGrayTerracotta = LightGrayGlazedTerracotta          $ 0.1
+LimeTerracotta      = LimeGlazedTerracotta               $ 0.1
+Log                 = CharCoal                           $ 0.1
+Log2                = CharCoal                           $ 0.1
```
peterbell10: The wiki says charcoal is worth 0.15

PR #4096, opened 2017-12-19
Concrete mixing

earboxer:
Features

Concrete powder turns into concrete when
* It is next to water when it receives a block update
* It falls onto a water block (even with Physics SandInstantFall=1)

Concerns
* <strike>When placed in deeper water (placed against a block), it will fall down a block before solidifying. Expected behavior is for concrete powder to turn into concrete because it is mixed with water (placing inside a water block). I think this should be checked before it becomes a falling entity, but I'm not sure how. (Or perhaps we ought to make a separate simulator for concrete?) (Maybe have something trigger on block movement/placement)</strike>
* <strike>Water spread updates doesn't call NeighborChanged, so I've kept some of the code in the fluid simulator. Are these 'NeighborChanged' supposed to be our block update structure? If so, then the fluid simulators should be creating these block updates whenever they spread. (by my tests, this appears to not be the case).</strike> (**Using Check() function, thanks @peterbell10** )

Conventions
* <strike>SoakNeighbors() is called by cFloodyFluidSimulator::SimulateBlock(), which takes in the coordinates as integers. According to the Code Conventions, we should pass these as `Vector3i`. Should I re-write SimulateBlock() to use `Vector3i`,<strike> or just fix the function that I added?</strike>(**EDIT: fixed SoakNeighbors for code conventions, We can always fix the others in a different Pull Request.**)</strike>
* Does the code in `FallingBlock.cpp` <strike>and `SandSimulator.cpp`</strike>(**EDIT:added documentation**) seem self-documenting enough, or should something be added to either of these?


src/Simulator/FloodyFluidSimulator.cpp
```diff
@@ -427,5 +429,39 @@ bool cFloodyFluidSimulator::HardenBlock(cChunk * a_Chunk, int a_RelX, int a_RelY
...
+	NIBBLETYPE BlockMeta;
+	// The blocks adjacent and below this water can be soaked
+	Vector3i Coords[] =
+	{
+		a_Rel + Vector3i( 1,  0,  0),
+		a_Rel + Vector3i(-1,  0,  0),
+		a_Rel + Vector3i( 0,  0,  1),
+		a_Rel + Vector3i( 0,  0, -1),
+		a_Rel + Vector3i( 0, -1,  0),
+	};
```
madmaxoft: Prefer a static array of offsets, then add them to `a_Rel` in the for-loop.

src/Simulator/FloodyFluidSimulator.h
```diff
@@ -44,6 +44,9 @@ class cFloodyFluidSimulator :
 	Returns whether the block was changed or not. */
 	bool HardenBlock(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta);
 
+	/** Checks if neighboring blocks are soakable, if so, converts them to soaked version. */
```
madmaxoft: Explain in the doxy-comment what the `a_BlockType` and `a_Meta` params are - they aren't immediately obvious.

madmaxoft:
Good first PR, thank you :)

earboxer:
Addressed review. (Squashed my commits so that the two mechanics (falling into water, water flowing next to it) are by themselves).

earboxer:
I'm going to modify this to be done by the block handler rather than fluid simulator (will mean faster conversion, and behaviorally mimicking java edition) (also, more readable code).

peterbell10:
>Are these 'NeighborChanged' supposed to be our block update structure?

I think you're looking for `Check`:
https://github.com/cuberite/cuberite/blob/6309c6a97fdbabfde978358f5f9a0f61ab74f91f/src/Blocks/BlockHandler.h#L138-L141

earboxer:
Thank you! I'll try that. I was noticing some performance issues earlier, so hopefully that fixes it. (checking 5 blocks every times water does anything isn't the best for performance).

src/Blocks/BlockConcretePowder.h
```diff
@@ -0,0 +1,98 @@
...
+		{
+			Vector3i( 1, 0,  0),
+			Vector3i(-1, 0,  0),
+			Vector3i( 0, 0,  1),
+			Vector3i( 0, 0, -1),
+			Vector3i( 0, 1,  0),
+		};
+		for (size_t i = 0; i < ARRAYCOUNT(Coords); i++)
+		{
+			if (IsBlockWater(a_Chunk.GetBlock({a_Rel.x + Coords[i].x, a_Rel.y + Coords[i].y, a_Rel.z + Coords[i].z})))
```
peterbell10: This has to use `a_Chunk.UnboundedRelGetBlockType` because the neighbours might be in a different chunk.

src/Blocks/BlockConcretePowder.h
```diff
@@ -0,0 +1,98 @@
...
+
+		static const Vector3i Coords[] =
+		{
+			Vector3i( 1, 0,  0),
+			Vector3i(-1, 0,  0),
+			Vector3i( 0, 0,  1),
+			Vector3i( 0, 0, -1),
+			Vector3i( 0, 1,  0),
+		};
+		for (size_t i = 0; i < ARRAYCOUNT(Coords); i++)
```
peterbell10: This loop can be replaced with `std::any_of`.

earboxer: Has this been used before in Cuberite?

earboxer:
I made the suggested changes except using `std::any_of`. I'll admit I'm not familiar with it, but I'm mostly concerned it will make the code less readable. (And with a simple search, I haven't found any usage of it in Cuberite at the present time).

src/Blocks/BlockConcretePowder.h
peterbell10: Admittedly, my search only finds one:
https://github.com/cuberite/cuberite/blob/6309c6a97fdbabfde978358f5f9a0f61ab74f91f/src/Simulator/FireSimulator.cpp#L124-L129
and it was only added yesterday at my request.

However, using an algorithms gives the loop a name.  That makes it much better at expressing the real intent: "a concrete block should soak if _any of_ its neighbours are water".  It's the exact same reason why `std::min(a, b)` is preferred over `(a < b) ? a : b`.

bearbin:
Have you pushed the code using `std::any_of`? It still looks to be the old version to me.

src/Blocks/BlockConcretePowder.h
earboxer: Okay. Implemented.

src/Simulator/FloodyFluidSimulator.cpp
earboxer: Done

src/Simulator/FloodyFluidSimulator.h
earboxer: And done

PR #4101, opened 2017-12-22
Fix Travis build

peterbell10:
* Fixes the remaining clang 5.0 warnings
 * Fixes AppleClang builds
 * Adds a clang 3.5 build to travis in addition to the new default of 5.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,72 @@
 language: cpp
 sudo: false
 
-os:
-- linux
-- osx
-
-compiler:
-- clang
-- gcc
+# Use Linux by default
+os: linux
 
 matrix:
-  exclude:
-  - os: osx
-    compiler: gcc
-
-addons:
-  apt:
-    sources:
-    - ubuntu-toolchain-r-test
-    packages:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -1,40 +1,74 @@
...
+    # Default clang
+    - compiler: clang
+      addons: &gdb
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gdb
+      env: *Release
+    - os: linux
```
bearbin: Why is os explicitly specified here, when it is not specified elsewhere?

peterbell10: Just an oversight from when I added the default.  Fixed now.

peterbell10:
I think fixing jsoncpp would be best but cuberite doesn't have its own fork.  If you could fork it @bearbin then I would fix the warnings properly.

bearbin:
I forked jsoncpp, so if you could submit a PR that would be great.

peterbell10:
I just pushed the changes to a new branch as cuberite wasn't on master anyway.

PR #4104, opened 2017-12-23
Fix style of Tools

bearbin:


bearbin:
I'll fix the items you pointed out and have a go at getting CBS to ignore stuff in comments.

bearbin:
I've had a look at CheckBasicStyle, and it seems like getting it to exclude stuff in comments would be a lot more work than it's worth, tbh. I think it will be staying how it is for now, by me at least.

peterbell10:
That's a shame, maybe it would be better to look into clang-format or even a custom clang tool.  Obviously beyond the scope of this PR though.

bearbin:
Merged manually.

PR #4105, opened 2017-12-23
Removed cBlockInfo.m_IsSnowable

Bond-009:
According to MCwiki snow will generate on all non-transparent blocks with the exception of packed ice.
Ref: https://minecraft.gamepedia.com/Snowfall#Effects
and https://minecraft.gamepedia.com/Snow_(layer)#Snowfall


src/BlockInfo.h
```diff
@@ -40,9 +40,6 @@ class cBlockInfo
 	/** Does a block disperse sky light? (only relevant for transparent blocks) */
 	bool m_IsSkylightDispersant;
 
-	/** Can this block hold snow atop? */
-	bool m_IsSnowable;
```
peterbell10: This breaks the lua api, you should manually bind accessors for `m_IsSnowable` like when `m_PlaceSound` was removed in #3897.

bearbin:
Snow will generate on leaves, so this logic can't be right. Perhaps snow will only generate on blocks that disperse or block light? (Including ice and leaves AFAIK).

peterbell10:
According to https://minecraft.gamepedia.com/Leaves, Leaves are only considered partially transparent so I guess they aren't counted.  I'm not sure how that fits in with `cBlockInfo` though.

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -347,6 +347,41 @@ static int tolua_set_cBlockInfo_m_PlaceSound(lua_State * tolua_S)
...
+
+static int tolua_set_cBlockInfo_m_IsSnowable(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+	if (!L.CheckParamSelf("cBlockInfo"))
+	{
+		return 0;
+	}
+
+	LOGWARNING("cBlockInfo.m_IsSnowable is deprecated");
```
bearbin: Can you mention what it is deprecated in favour of here?

src/BlockInfo.cpp
```diff
@@ -313,6 +313,7 @@ cBlockInfo::cBlockInfoArray::cBlockInfoArray()
 	Info[E_BLOCK_SANDSTONE_STAIRS             ].m_Transparent = true;
 	Info[E_BLOCK_SAPLING                      ].m_Transparent = true;
 	Info[E_BLOCK_SIGN_POST                    ].m_Transparent = true;
+	Info[E_BLOCK_SNOW                         ].m_Transparent = true;
```
bearbin: I'm confused as to why you have added this - a full block of snow is not transparent.

bearbin:
> According to https://minecraft.gamepedia.com/Leaves, Leaves are only considered partially transparent so I guess they aren't counted. I'm not sure how that fits in with cBlockInfo though.

Upon further investigation, the only other partially transparent block is cobweb, but that allows snow through. It seems this is a more complex matter :|

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -347,6 +347,41 @@ static int tolua_set_cBlockInfo_m_PlaceSound(lua_State * tolua_S)
...
 
+static int tolua_get_cBlockInfo_m_IsSnowable(lua_State * tolua_S)
+{
+	cLuaState L(tolua_S);
+	if (!L.CheckParamSelf("const cBlockInfo"))
+	{
+		return 0;
+	}
+
+	L.Push("");
```
peterbell10: This should at the very least be returning a boolean. Even better would be to have this return `cBlockInfo::IsSnowable()`.  Possibly by storing the block id in the `cBlockInfo` instance without exporting it to lua.

Bond-009:
Until somebody finds a way to have ice, leaves and cobwebs be partially transparent in cBlockInfo, I could just add ice and leaves as exceptions in the IsSnowable method. Is this acceptable as a temporary solution?

bearbin:
That's probably acceptable as a temporary solution, but you need to fix the other issues that we pointed out before this would be ready to merge.

bearbin:
@Bond-009, just a reminder that there's still a comment outstanding on this PR - are you going to come back to it?

src/Bindings/DeprecatedBindings.cpp
Bond-009: So would returning false (or true) be good enough, because with my limited knowledge of C++ I can't find a decent way to store the block id in the cBlockInfo instance.

peterbell10: Do you mind if I just make the change myself?

Bond-009: I don't mind at all. 👍

peterbell10:
Sorry for that mess... I'll try to fix my local branch before pushing again.

Issue #4068, opened 2017-11-11
Webadmin footer covering content in Safari


ePirat:
Server OS:        Linux
Commit id:        c54bf40

### Expected behavior
The footer of the webadmin should be after the content, or at least, not covering it, without the ability to scroll.

### Actual behavior
Scrolling a long page to the bottom, the footer covers the content.
Scrolling down further is impossible, so the save button can't be clicked.

<img width="1122" alt="screenshot" src="https://user-images.githubusercontent.com/323210/32693399-d4f578dc-c72a-11e7-8e23-43ccbe6c8c17.png">

### Steps to reproduce the behavior
Open a longer page, like `/Core/Plugins` in Safari 11.0.2 and scroll to the bottom.

ePirat:
A simple workaround is to set `padding-bottom: 65px;` for `.row2`.

bearbin:
@ePirat, can you test with my change in #4107?

willi123yao:
Tested your change, problem persists in Safari browser. However, ePirat's solution seems to be working.

PR #4107, opened 2017-12-24
Remove duplicated margin statements

bearbin:
Fixes issue #4068, hopefully.


peterbell10:
I don't know about safari but it's still looks wrong on microsoft edge.

bearbin:
Thanks for checking @peterbell10. I'm not a CSS wizard and as I don't have a broken browser to test it on I'm not really sure what else could be the problem.

PR #4113, opened 2017-12-26
.editorconfig fixes

bearbin:
Added specification for the type of line endings preferred and removed `spaces_around_operators` as it does not appear in the specification.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,7 +3,6 @@ root = true
 
 [*]
 charset = utf-8
-end_of_line = lf
 indent_style = tab
 insert_final_newline = true
 trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -2,9 +2,8 @@
 root = true
 
 [*]
-indent_style = tab
 charset = utf-8
-trim_trailing_whitespace = true
+end_of_line = lf
```
peterbell10: @madmaxoft's comment from #3936:
>This shouldn't be handled by editorconfig, but rather by git. I want CRLFs on my Windows machine and LFs on my Linux machine.

bearbin: I had forgotten about that conversation. Let me fix the commit.

PR #4114, opened 2017-12-26
Water bottles are drinkable potions

### LINKED ISSUE #4111: Water bottle should be drinkable
Client version:   1.12.2
Server OS:        Windows
Commit id:        f0b05e78f70fa883bfdc74568bd7bf914025306c

### Expected behavior
Holding a water bottle, right click, and drinks the water bottle, returns a glass bottle.

### Actual behavior
Throw it away and it breaks.

As described before.

bearbin:
Fixes #4111 

There seem to be some other issues with potions as well that will need some more intensive fixes, but this will fix the immediate issue in #4111.

### DIFF
--- a/src/Entities/EntityEffect.cpp
+++ b/src/Entities/EntityEffect.cpp
@@ -125,10 +125,10 @@ int cEntityEffect::GetPotionEffectDuration(short a_ItemDamage)
 
 bool cEntityEffect::IsPotionDrinkable(short a_ItemDamage)
 {
-	// Drinkable potion if 13th lowest bit is set
-	// Water bottles are also drinkable potions
-	// Ref.: https://minecraft.gamepedia.com/Java_Edition_data_values#Potions
-	return (a_ItemDamage == 0) || ((a_ItemDamage & 0x2000) != 0);
+	// Potions are drinkable if they are not splash potions.
+	// i.e. potions are drinkable if the 14th lowest bit is not set
+	// Ref.: https://minecraft.gamepedia.com/Java_Edition_data_values#.22Splash_potion.22_bit
+	return ((a_ItemDamage & 0x4000) == 0);
 }
 
 
--- a/src/Entities/EntityEffect.cpp
+++ b/src/Entities/EntityEffect.cpp
@@ -126,8 +126,9 @@ int cEntityEffect::GetPotionEffectDuration(short a_ItemDamage)
 bool cEntityEffect::IsPotionDrinkable(short a_ItemDamage)
 {
 	// Drinkable potion if 13th lowest bit is set
+	// Water bottles are also drinkable potions
 	// Ref.: https://minecraft.gamepedia.com/Potions#Data_value_table
-	return ((a_ItemDamage & 0x2000) != 0);
+	return (a_ItemDamage == 0) || ((a_ItemDamage & 0x2000) != 0);
 }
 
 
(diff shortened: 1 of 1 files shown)


Bond-009:
I don't think that this is a good solution as there also are water splash potions.

P.S. the ref to the mcwiki in the comment is outdated and should be replaced by this one: https://minecraft.gamepedia.com/Java_Edition_data_values#Potions

peterbell10:
Testing the "splash_bit" from the wiki page seems to work correctly:
```cpp
return ((a_ItemDamage & 0x4000) == 0);
```

bearbin:
> I don't think that this is a good solution as there also are water splash potions.

This solution works as water splash potions have a data value != 0. However @peterbell10's suggestion looks to be better.

> P.S. the ref to the mcwiki in the comment is outdated and should be replaced by this one: https://minecraft.gamepedia.com/Java_Edition_data_values#Potions

Done

> Testing the "splash_bit" from the wiki page seems to work correctly:

This looks to be a good solution, but it's not exactly clear that a function `IsPotionDrinkable` should check "Is potion not a spash potion". Are there any potions that are neither drinkable nor splash potions?

peterbell10:
I don't think there are any non-drinkable, non-splash potions.  Still, It might be better to put splash bit testing into it's own function `IsPotionSplash`.  Especially as the original issue caused splash potion creation so obviously `IsPotionDrinkable` is being used in this sense.

The code seems to use not drinkable to imply splash in other places as well:

https://github.com/cuberite/cuberite/blob/4dcff34fecde217efd27bae310029d951d5f5252/src/Entities/EntityEffect.cpp#L112

bearbin:
> I don't think there are any non-drinkable, non-splash potions.

OK, well if that's the case then I don't see any point in making a change to the name of the function. It may not be ideal, but since MC does not currently make the distinction there's no point in us making that distinction. I think changing the comment text as I have done is sufficient.

Bond-009:
There are lingering potions, but I don't know if they are counted as a splash potion.
Ref: https://minecraft.gamepedia.com/Lingering_Potion

bearbin:
I think the lingering potions are just splash potions, at least with regard to the splash potion data bit.

peterbell10:
>I don't see any point in making a change to the name of the function
I was actually meaning a separate function but I suppose it doesn't really matter.

src/Entities/EntityEffect.cpp
```diff
@@ -125,9 +125,10 @@ int cEntityEffect::GetPotionEffectDuration(short a_ItemDamage)
 
 bool cEntityEffect::IsPotionDrinkable(short a_ItemDamage)
 {
-	// Drinkable potion if 13th lowest bit is set
-	// Ref.: https://minecraft.gamepedia.com/Potions#Data_value_table
-	return ((a_ItemDamage & 0x2000) != 0);
+	// Potions are drinkable if they are not splash potions.
+	// i.e. potions are drinkable if the 14th lowest bit is not set
+	// Ref.: https://minecraft.gamepedia.com/Java_Edition_data_values#.22Splash_potion.22_bit
+	return (a_ItemDamage & 0x4000) == 0;
```
peterbell10: Isn't the convention to _always_ have brackets around comparisons? Like how it was before.

bearbin: Yes...

Pokechu22:
Do we really want to keep the old potion ID system?  It's a bit of a mess IMO (though if 1.8 is still supported, something like this is needed).

bearbin:
@Pokechu22 It works, and there are higher priority enhancements to make, but if you or anybody else wanted to give replacing it a go that would be great :)

9caihezi:
Is lingering potion implemented? Is so we need a new issue for it. Currently lingering will become splash potion, and cannot be made by dragon breath.

bearbin:
I don't believe that lingering potions are implemented, as they were only added to vanilla in 1.9. In any case, unless there is another issue about implementing them it would be a good idea to create one.

PR #4116, opened 2017-12-26
Protocol: Fix potion metadata parsing

### LINKED ISSUE #4115: Creating Thick or Awkward Potions from Creative inventory gives wrong result
Client version:   1.12.2
Server OS:        Linux
Commit id:        96b72ab88ec2610a41707b9705808d3c1c46046a

### Expected behavior

Potion of correct type placed in inventory.

### Actual behavior

The wrong type of potion is produced. For awkward potions it is slowness and for thick it is poison. 

As title.

### Other information

Awk potion data value: Hex 10 / Dec 16
Slowness potion data value: Hex 0a / Dec 10

Thick potion data value: Hex 20 / Dec 32
Poison potion data value: Hex 14 / Dec 20

It seems almost like there is some conversion from heximal going on here. None of the other potions seem to be affected.

peterbell10:
Fixes #4115

### DIFF
--- a/src/Protocol/Protocol_1_9.cpp
+++ b/src/Protocol/Protocol_1_9.cpp
@@ -3128,11 +3128,11 @@ void cProtocol_1_9_0::ParseItemMetadata(cItem & a_Item, const AString & a_Metada
 					}
 					else if (PotionEffect.find("thick") != AString::npos)
 					{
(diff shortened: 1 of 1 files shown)


bearbin:
One of the builds failed again on the mac slave. It was a different test this time though.

PR #4120, opened 2017-12-27
Damage fishing rod after use

### LINKED ISSUE #4118: Fishing rod does not lose durability
Client version:   1.12.2
Server OS:        Windows
Commit id:        f0b05e78f70fa883bfdc74568bd7bf914025306c

### Expected behavior
Use fishing damages the rod.

### Actual behavior
The rod does not lose durability.

As described.

Bond-009:
Fixes #4118

* Damage fishing rod after use
* Give xp to the player for catching something
* Fixed junk drops

### DIFF
--- a/src/Items/ItemFishingRod.h
+++ b/src/Items/ItemFishingRod.h
@@ -89,6 +89,7 @@ public:
 						return true;
 					}
 				);
+				a_Player->UseEquippedItem(5);
 			}
 			else if (FloaterInfo.CanPickup())
 			{
@@ -96,22 +97,27 @@ public:
 				int ItemCategory = Random.RandInt(99);
 				if (ItemCategory <= 4)  // Treasures 5%
 				{
-					int Treasure = Random.RandInt(5);
-					switch (Treasure)
+					switch (Random.RandInt(5))  // Each piece of treasure has an equal chance of 1 / 6
 					{
 						case 0:
 						{
-							Drops.Add(cItem(E_ITEM_BOW));  // TODO: Enchantments
+							cItem Bow(E_ITEM_BOW, 1, Random.RandInt<short>(50));
+							Bow.EnchantByXPLevels(Random.RandInt(22, 30));
+							Drops.Add(Bow);
 							break;
 						}
 						case 1:
 						{
-							Drops.Add(cItem(E_ITEM_BOOK));  // TODO: Enchanted book
+							cItem Book(E_ITEM_ENCHANTED_BOOK);
+							Book.EnchantByXPLevels(30);
+							Drops.Add(Book);
 							break;
 						}
 						case 2:
 						{
-							Drops.Add(cItem(E_ITEM_FISHING_ROD, 1, Random.RandInt<short>(50)));  // Fishing rod with durability. TODO: Enchantments on it
+							cItem Rod(E_ITEM_FISHING_ROD, 1, Random.RandInt<short>(50));
+							Rod.EnchantByXPLevels(Random.RandInt(22, 30));
+							Drops.Add(Rod);
 							break;
 						}
 						case 3:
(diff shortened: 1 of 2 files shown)


bearbin:
Looks good. Are you also going to implement the LOTS enchantment?

src/Items/ItemFishingRod.h
```diff
@@ -104,22 +104,27 @@ class cItemFishingRodHandler :
...
 				if (ItemCategory <= 4)  // Treasures 5%
 				{
-					int Treasure = Random.RandInt(5);
-					switch (Treasure)
+					switch (Random.RandInt(5))  // Each piece of treasure has an equal chance of 1 / 6
 					{
 						case 0:
 						{
-							Drops.Add(cItem(E_ITEM_BOW));  // TODO: Enchantments
+							cItem Bow = cItem(E_ITEM_BOW, 1, Random.RandInt<short>(50));
```
peterbell10: Why are you constructing the items like this?

src/Items/ItemFishingRod.h
```diff
@@ -143,55 +148,57 @@ class cItemFishingRodHandler :
 				}
 				else if (ItemCategory <= 14)  // Junk 10%
 				{
-					int Junk = Random.RandInt(70);
-					if (Junk <= 1)
+					int Junk = Random.RandInt(82);
+					if (Junk < 10)  // 10 / 83 chance of spawning a bowl
```
peterbell10: This seems ideal for a [`std::discrete_distribution`](http://en.cppreference.com/w/cpp/numeric/random/discrete_distribution).

src/Items/ItemFishingRod.h
Bond-009: Is there another/better way to do this?

peterbell10: What's wrong with the normal way?
```cpp
cItem Bow(E_ITEM_BOW, 1, Random.RandInt<short>(50));
```

src/Items/ItemFishingRod.h
```diff
@@ -100,26 +100,37 @@ class cItemFishingRodHandler :
 			}
 			else if (FloaterInfo.CanPickup())
 			{
+				uint LotSLevel = a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLuckOfTheSea);
```
bearbin: This breaks the build on windows. It would be better to use a UInt32

Bond-009: Done

src/Items/ItemFishingRod.h
```diff
@@ -100,26 +100,37 @@ class cItemFishingRodHandler :
 			}
 			else if (FloaterInfo.CanPickup())
 			{
+				UInt32 LotSLevel = a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLuckOfTheSea);
+
+				// Chances for getting an item from the category for each level of Luck of the Sea (0 - 3)
+				const int TreasureChances[] = {50, 71, 92, 113};  // 5% | 7.1% | 9.2% | 11.3%
+				const int JunkChances[] = {100, 81, 61, 42};  // 10% | 8.1% | 6.1% | 4.2%
```
peterbell10: Potential buffer overrun if a player cheats in a fishing rod with LotS X.

Bond-009: Does the chance for treasure increase when X > 3?

peterbell10: I can't find anything on the wiki.  I'm sure it would be okay to just treat anything >3 the same as level 3 though.

src/Items/ItemFishingRod.h
```diff
@@ -97,29 +97,45 @@ class cItemFishingRodHandler :
 						return true;
 					}
 				);
+				a_Player->UseEquippedItem(5);
 			}
 			else if (FloaterInfo.CanPickup())
 			{
+				UInt32 LotSLevel = a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLuckOfTheSea);
```
bearbin: A call to std::min would be more appropriate here.

src/Items/ItemFishingRod.h
```diff
@@ -89,29 +89,41 @@ class cItemFishingRodHandler :
 						return true;
 					}
 				);
+				a_Player->UseEquippedItem(5);
 			}
 			else if (FloaterInfo.CanPickup())
 			{
+				UInt32 LotSLevel = std::min(a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLuckOfTheSea), 3);
```
peterbell10: you need to use the unsigned literal `3u` because the lhs of `std::min` is unsigned.

PR #4125, opened 2018-01-01
Fix cUrlClient leak

### LINKED ISSUE #4040: Memory leak in Networking
Client version:   N/A
Server OS:        Windows
Commit id:       e0d1f79

### Actual behavior
Running in debug, Visual Studio consistently reports memory leaks when WorldEdit is loaded:
```
Detected memory leaks!
Dumping objects ->
[...]
D:\git\cuberite\src\OSSupport\TCPLinkImpl.cpp(68) : {392593} client block at 0x05FA0A30, subtype 0, 116
[...]
D:\git\cuberite\src\HTTP\UrlClient.cpp(43) : {392544} client block at 0x05FA8508, subtype 0, 348 bytes long.
 Data: <  ?   ?     GET > 98 89 3F 01 AC 89 3F 01 F0 9D C8 05 47 45 54 00 
```

Start a server with WorldEdit loaded.

peterbell10:
Fixes #4040 

* The TCP connection is now shutdown after `OnBodyFinished`
* Any open connections are closed when `cNetworkSingleton::Terminate()` is called.
* Removed ownership cycles in `cUrlClientRequest`

I also added a check to the test to ensure there are no leaks.

### DIFF
--- a/src/HTTP/UrlClient.cpp
+++ b/src/HTTP/UrlClient.cpp
@@ -54,12 +54,11 @@ public:
 		m_Callbacks->OnError(a_ErrorMessage);
 
 		// Terminate the request's TCP link:
-		auto link = m_Link;
+		auto link = m_Link.lock();
 		if (link != nullptr)
 		{
 			link->Close();
 		}
-		m_Self.reset();
 	}
 
 
@@ -76,7 +75,7 @@ public:
 		{
 			return nullptr;
 		}
-		cX509CertPtr cert;
+		cX509CertPtr cert = std::make_shared<cX509Cert>();
 		if (!cert->Parse(itr->second.data(), itr->second.size()))
 		{
 			LOGD("OwnCert failed to parse");
@@ -92,7 +91,7 @@ public:
 		{
 			return nullptr;
 		}
-		cCryptoKeyPtr key;
+		cCryptoKeyPtr key = std::make_shared<cCryptoKey>();
 		auto passItr = m_Options.find("OwnPrivKeyPassword");
 		auto pass = (passItr == m_Options.end()) ? AString() : passItr->second;
 		if (!key->ParsePrivate(itr->second.data(), itr->second.size(), pass))
@@ -126,15 +125,15 @@ protected:
 	/** Extra options to be used for the request. */
 	AStringMap m_Options;
 
-	/** SharedPtr to self, so that this object can keep itself alive for as long as it needs,
+	/** weak_ptr to self, so that this object can keep itself alive as needed by calling lock(),
 	and pass self as callbacks to cNetwork functions. */
-	std::shared_ptr<cUrlClientRequest> m_Self;
+	std::weak_ptr<cUrlClientRequest> m_Self;
 
 	/** The handler that "talks" the protocol specified in m_UrlScheme, handles all the sending and receiving. */
 	std::shared_ptr<cSchemeHandler> m_SchemeHandler;
 
 	/** The link handling the request. */
-	cTCPLinkPtr m_Link;
+	std::weak_ptr<cTCPLink> m_Link;
 
 	/** The number of redirect attempts that will still be followed.
 	If the response specifies a redirect and this is nonzero, the redirect is followed.
@@ -171,7 +170,6 @@ protected:
 	virtual void OnError(int a_ErrorCode, const AString & a_ErrorMsg) override
 	{
 		m_Callbacks->OnError(Printf("Network error %d (%s)", a_ErrorCode, a_ErrorMsg.c_str()));
-		m_Self.reset();
 	}
 
 
@@ -495,15 +493,17 @@ void cUrlClientRequest::RedirectTo(const AString & a_RedirectUrl)
 		return;
 	}
 
+	// Keep ourself alive while the link drops ownership
+	auto Self = m_Self.lock();
+
 	// Do the actual redirect:
-	m_Link->Close();
+	m_Link.lock()->Close();
 	m_Url = a_RedirectUrl;
 	m_NumRemainingRedirects = m_NumRemainingRedirects - 1;
-	auto res = DoRequest(m_Self);
+	auto res = DoRequest(Self);
 	if (!res.first)
 	{
 		m_Callbacks->OnError(Printf("Redirection failed: %s", res.second.c_str()));
-		return;
 	}
 }
 
@@ -562,9 +562,6 @@ void cUrlClientRequest::OnRemoteClosed()
 	{
 		handler->OnRemoteClosed();
 	}
-
-	// Let ourselves be deleted
-	m_Self.reset();
 }
 
 
@@ -592,7 +589,8 @@ std::pair<bool, AString> cUrlClientRequest::DoRequest(std::shared_ptr<cUrlClient
 		return std::make_pair(false, Printf("Unknown Url scheme: %s", m_UrlScheme.c_str()));
 	}
 
-	if (!cNetwork::Connect(m_UrlHost, m_UrlPort, m_Self, m_Self))
+	// Connect and transfer ownership to the link
+	if (!cNetwork::Connect(m_UrlHost, m_UrlPort, a_Self, a_Self))
 	{
 		return std::make_pair(false, "Network connection failed");
 	}
--- a/src/HTTP/UrlClient.cpp
+++ b/src/HTTP/UrlClient.cpp
@@ -330,8 +330,8 @@ public:
 	// cHTTPResponseParser::cCallbacks overrides:
 	virtual void OnError(const AString & a_ErrorDescription) override
 	{
-		m_ParentRequest.CallErrorCallback(a_ErrorDescription);
 		m_Link = nullptr;
+		m_ParentRequest.CallErrorCallback(a_ErrorDescription);
 	}
 
 
--- a/src/HTTP/UrlClient.cpp
+++ b/src/HTTP/UrlClient.cpp
@@ -430,6 +430,8 @@ public:
 		else
 		{
 			m_ParentRequest.GetCallbacks().OnBodyFinished();
+			// Finished recieving data, shutdown the link
+			m_Link->Shutdown();
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


peterbell10:
>I would suggest you merge it

I can't merge without an approving review.  If you're not comfortable reviewing this then it'll just have to wait.

Probably worth waiting for xoft's input anyway.

src/HTTP/UrlClient.cpp
```diff
@@ -54,12 +54,11 @@ class cUrlClientRequest:
 		m_Callbacks->OnError(a_ErrorMessage);
 
 		// Terminate the request's TCP link:
-		auto link = m_Link;
+		auto link = m_Link.lock();
 		if (link != nullptr)
```
bibo38: Just learned  about `std:.weak_ptr`.
Now I've also learned the nice trick `if(auto link = m_Link.lock())` from SO 😄.
Maybe you also could use this to limit the scope of the `link` variable.

src/HTTP/UrlClient.cpp
```diff
@@ -76,7 +75,7 @@ class cUrlClientRequest:
 		{
 			return nullptr;
 		}
-		cX509CertPtr cert;
+		cX509CertPtr cert = std::make_shared<cX509Cert>();
```
bibo38: Why? C++ has default constructors...

src/HTTP/UrlClient.cpp
```diff
@@ -493,15 +493,17 @@ void cUrlClientRequest::RedirectTo(const AString & a_RedirectUrl)
 		return;
 	}
 
+	// Keep ourself alive while the link drops ownership
+	auto Self = m_Self.lock();
+
 	// Do the actual redirect:
-	m_Link->Close();
+	m_Link.lock()->Close();
```
bibo38: `lock()` could return an emtpy `shared_ptr`. And then everything will go 💥 .
So it should be:

    if(auto link = m_Link.lock())
        link->Close();

src/OSSupport/NetworkSingleton.cpp
```diff
@@ -102,11 +103,25 @@ void cNetworkSingleton::Terminate(void)
...
+	// Close all open connections:
 	{
 		cCSLock Lock(m_CS);
-		m_Connections.clear();
-		m_Servers.clear();
+		// Must take copies because Close will modify lists
+		auto Conns = m_Connections;
+		for (auto & Conn : Conns)
+		{
+			Conn->Close();
```
bibo38: Sorry, but WHY do the connections modify some lists? And WHY only, when calling `Close()` and not via the destructor?
I liked the old code of just clearing the lists and let the destructors close the connections.
So why we just go the old Java way and close everything, we someday opened? We have destructors 🎉 .

peterbell10: > WHY do the connections modify some lists?

When the connection is closed by a remote it has to remove itself from the network singleton's list otherwise it would leak.

> And WHY only, when calling Close() and not via the destructor?

Because if there is a `shared_ptr` to it in the list, the destructor will never be called.

>I liked the old code of just clearing the lists and let the destructors close the connections.

Each connection has a `shared_ptr` to itself, if the singleton shutsdown before every connection is closed then it's guaranteed to leak.

src/HTTP/UrlClient.cpp
peterbell10: Indeed it does. `std::shared_ptr`'s default constructor doesn't allocate anything while `std::make_shared` does.

bibo38: Oh, men... Just thought, that empty parameters to `make_shared` would create an empty shard_ptr...

src/OSSupport/NetworkSingleton.cpp
bibo38: > When the connection is closed by a remote it has to remove itself from the network singleton's list otherwise it would leak.

When the `NetworkSingleton` wants to have every active connection, then it also should remove them (observe a connection failure and remove it...). But this is just a minor issue for me.

 > Because if there is a shared_ptr to it in the list, the destructor will never be called.

This was meant for the next comment: E.g. Why don't we just clear the list and the destructors will be called, if the connection is not used elsewhere...

> Each connection has a shared_ptr to itself

Ok, thats a bad design... The right way would be giving a `weak_ptr`/`shared_ptr` on the heap to the callbacks.
Or at least clear the self pointer after every callback, so the object just dies of after the last callback has beed executed (only works, if there can only be one active callback for one Connection).

peterbell10: >  the NetworkSingleton [...] should remove them

The only way this could possibly work would be if the singleton held `weak_ptr`s and regularly polled them to see if they'd expired.  This is a complete waste of time since the connection knows when it's closed.

>Why don't we just clear the list and the destructors will be called

Because the link's destructor _won't_ be called.

>Ok, thats a bad design... The right way would be giving a weak_ptr/shared_ptr on the heap to the callbacks.

The connection has to stay alive until the remote has closed it.  The best solution would be to have libevent hold the `shared_ptr` to the connection but that's never going to happen because it's in C.

bibo38: > The only way this could possibly work would be if the singleton held weak_ptrs and regularly polled them to see if they'd expired.

What about the observer pattern? E.g. the NetworkSingleton registers an callback/observer at the connection object. And if the connection is closed, it informs everyone observing this event.

> > Why don't we just clear the list and the destructors will be called

> Because the link's destructor won't be called.

Yep, I know, because of that bad design self pointer... This was meant as "Why cant we use a good design, where we just wait till nobody references the shared_ptr anymore and it will autoclose the connection automatically"

> The connection has to stay alive until the remote has closed it.

So I can't close the connection myself 😱 😆?
But seriously, all active connections are currently already held open by `cNetworkSingleton::m_Connections`. So, why do we need `m_Self`?

>  The best solution would be to have libevent hold the shared_ptr to the connection but that's never going to happen because it's in C.

`static_cast<void*>(new std::shared_ptr(m_MyWeakPtr.lock()))` (assuming, that the `weak_ptr` has a vaild reference).
Now I have a C-compatible `void*` pointer for my callbacks.

And as I've just seen, that the [current usage of m_Self](https://github.com/cuberite/cuberite/blob/cd88a117352970120df224a9e87a38b3884b1c3a/src/OSSupport/TCPLinkImpl.cpp#L365) is **undefined behaviour**:

    cTCPLinkImplPtr Self = static_cast<cTCPLinkImpl *>(a_Self)->m_Self;
    if (Self == nullptr)
    {
        // The link has already been freed
        return;
    }

So, assuming that the connection has been closed with `Close()`:
- `m_Self` doesn't hold a pointer
- `NetworkSingleton` doesn't hold a pointer
- So the `shared_ptr` deletes this class instance...
- Now libevent may call the callback (maybe from a request made right before `Close()` has been called)
- **`a_Self` now points to already freed memory!**
- So we get some `m_Self` from the already freed memory (maybe this is already another connection)...
- And we use the garbage `Self` variable to check, if the object has already beed deleted 💥 . `rand()` would be more efficient 🗡 !

peterbell10: >What about the observer pattern?

There's no practical difference from the current way it's done and an observer pattern.

>So I can't close the connection myself 😱 😆?

Not cleanly. The connection has to finish sending data and perform a termination handshake first:
https://en.wikipedia.org/wiki/Transmission_Control_Protocol#Connection_termination

>Now I have a C-compatible void* pointer for my callbacks.

Yes, and the responsibility is still on `cTCPLink` to delete that `shared_ptr` so you haven't gained anything.

bibo38: > Not cleanly. The connection has to finish sending data and perform a termination handshake first

Doesn't handle libevent/Windows/Linux these implementation specific things? Otherwise I don't see, where the current implementation waits, when I call `Close()`. If I'm right, `Close()` destroys `m_Self` (and the other from `cNetworkSingleton`) and therefore instantly deletes the current instance.

> Yes, and the responsibility is still on cTCPLink to delete that shared_ptr so you haven't gained anything.

The responsibilty lies at the event handler! It deletes the `shared_ptr` on the heap after executing it's functionality. So after the last event has been delivered, the last `shared_ptr` is destroyed and the object is being deleted (which then closes the communication). But in the current implementation `static_cast<void*>(new std::weak_ptr(m_MyWeakPtr))` is enough (because the callback doesn't need to do anything, if the object instance is already dead).

Furthermore I now have no undefined behaviour, which I described in my last comment 🎉 ! So I've gained a **lot** 👊 .

peterbell10: > I don't see, where the current implementation waits

`Close()` doesn't but `Shutdown()` does.

>The responsibilty lies at the event handler!

Which is `cTCPLinkImpl::EventCallback`.

>static_cast<void*>(new std::weak_ptr(m_MyWeakPtr))

And when do you call delete on the `weak_ptr`?  You've just shifted the problem.

bibo38: > Close() doesn't but Shutdown() does.

So why do you dirty close all active connections? I would only expect calling `Close()` only in cases like crashes, but not on a proper shutdown.

> Which is cTCPLinkImpl::EventCallback.

Which is static and therefore is nothing like a member variable/function.

> And when do you call delete on the weak_ptr? You've just shifted the problem.

Now I see the problem 😄. I thought, that for every callback there would be a executing call like `PollEventsAsync(X, Callback, new weak_ptr)` and I could just delete it after executing the callback.

Then we need to use a static container to store the instances. E.g. `std::unordered_map<int, Connection*> instances`. The `int` is an unique connection identifier and is passed as the parameter to the callback. Now the destructor of `Connection` removes it's own pointer and signalizes the callback, that this connection doesn't exist anymore.

But maybe it is enough to just remove `m_Self` and in the destructor set the callbacks to `NULL`, so that libevent cannot call any callback anymore. Then just add a mutex/semaphore/... to `Communication` to let parallel running event handlers finish before the destructor finishes.

madmaxoft: To add a bit to defend the current style, it is very much dictated by LibEvent and subtle bugs we encountered: one thread asking LibEvent to stop delivering events for a connection, while LibEvent is currently executing a callback in another thread -> trouble any way you look at it. Even worse, sometimes LibEvent calls two consecutive calls into the callbacks before they are un-registered.

I like the new rewrite.

Issue #3695, opened 2017-05-05
build an End Portal?


G-XVI:
Client version:   1.8 - 1.11.2
Server OS:        Linux

I cant make an End Portal.
Cant place an Eye of Ender of the End Portal Frame Block.

Bond-009:
See #3445

earboxer:
#4126 does this with the exception that  it doesn't create the block entity (tile entity). The portals become invisible when you relog.

Issue #3445, opened 2016-11-22
Eye of Ender issues


yangm97:
Eye of Ender does... something? Looks like it's trying to open a schematic according to the console, when you right click with an Eye of Ender.
Eye of Ender doesn't do intended functions: can't be thrown, can't be inserted into end portal frame.

Taken from #1109

Bjornskjald:
Still happening, any fix?

earboxer:
https://github.com/cuberite/cuberite/pull/4126 does the right clicking bit to insert into end portal frame.

PR #4126, opened 2018-01-02
Create Ender Portal with Eyes of Ender

earboxer:
The algorithm was designed so that an invalid portal will fail, no matter which block had the eye inserted in last.

All portals must have facing the center. (Stand where you want the portal to appear, place portal frames around you.

fixes half of #3445 
works on #3695

Note: Need to create block entity (tile entity) so that portals don't become invisible when you relog.


src/Items/ItemEyeOfEnder.h
```diff
@@ -0,0 +1,59 @@
+
+#pragma once
+
+#include "ItemHandler.h"
+#include "ItemThrowable.h"
```
bearbin: Since this class extends cItemThrowableHandler, updating the comment in ItemThrowable.h to include ender pearls would be a good idea

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+
+	virtual bool GetPlacementBlockTypeMeta(
+		cChunkInterface & a_ChunkInterface, cPlayer & a_Player,
+		int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
+		int a_CursorX, int a_CursorY, int a_CursorZ,
+		BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta
+	) override
+	{
+		a_BlockType = m_BlockType;
+		a_BlockMeta = cBlockRedstoneRepeaterHandler::RepeaterRotationToMetaData(a_Player.GetYaw());
```
bearbin: This seems an odd choice of function, I think this should either be separated out of cBlockRedstoneRepeaterHandler or copied into this handler, as there's nothing logically connecting redstone repeaters and ender portal frame blocks.

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+		return true;
+	}
+
+
+
+
+
+	virtual void OnPlaced(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, int a_BlockX, int a_BlockY, int a_BlockZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) override
+	{
+		/*
```
bearbin: This comment seems misplaced, should it not be in FindAndSetPortal instead?

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+	{
+		E_VALID_PORTAL_FRAME = 0,
+		E_INVALID_PORTAL_FRAME = 1,
+		E_NOT_PORTAL_FRAME = 2,
+	};
+
+
+
+
+	/** Returns 0 if portal cannot be made, 1 if portal was made. */
```
bearbin: This is a boolean function, not an int.

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+	bool FindAndSetPortal(Vector3i a_FirstFrame, NIBBLETYPE a_Direction, cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface)
+	{
+		const int MIN_PORTAL_WIDTH = 3;
+		const int MAX_PORTAL_WIDTH = 4;
+
+		
+		Vector3i Left[4] =
```
bearbin: Constructing these every time seems silly to me, since these coord adjustments don't change with the portal in question. Should these be globalised, as in the [fire simulator](https://github.com/cuberite/cuberite/blob/6309c6a97fdbabfde978358f5f9a0f61ab74f91f/src/Simulator/FireSimulator.cpp#L42)

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+						{
+							// Don't build a portal that is too long.
+							return 0;
+						}
+					}
+					Former = Former + Left[(a_Direction + Turns) % 4];
+					Proper++;
+					continue;
+				}
+				else if (EdgeValid == E_INVALID_PORTAL_FRAME)
```
bearbin: These two elses could be combined into a single early-bail-out here.

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+
+
+
+
+	/** Check a specific block to see if it is a valid portal frame. */
+	int IsValidFrame(Vector3i a_Pos, NIBBLETYPE a_ShouldFace, cChunkInterface & a_ChunkInterface)
+	{
+		BLOCKTYPE BlockType;
+		NIBBLETYPE BlockMeta;
+		if (! a_ChunkInterface.GetBlockTypeMeta(a_Pos, BlockType, BlockMeta))
```
bearbin: Style: spaces before the ! are not good.

src/Items/ItemEyeOfEnder.h
```diff
@@ -0,0 +1,59 @@
...
+	{
+		BLOCKTYPE FacingBlock;
+		NIBBLETYPE FacingMeta;
+		a_World->GetBlockTypeMeta(a_BlockX, a_BlockY, a_BlockZ, FacingBlock, FacingMeta);
+		switch (FacingBlock)
+		{
+			case E_BLOCK_END_PORTAL_FRAME:
+			{
+				// Fill the portal frame. 0x4 is the bit for holding the eye of ender.
+				if ((FacingMeta & 4) != 4)
```
bearbin: Normally the hex repreentation is used, i.e `0x4`

src/Items/ItemEyeOfEnder.h
```diff
@@ -0,0 +1,59 @@
...
+				{
+					a_World->SetBlock(a_BlockX, a_BlockY, a_BlockZ, E_BLOCK_END_PORTAL_FRAME, FacingMeta | 4);
+					if (!a_Player->IsGameModeCreative())
```
bearbin: Is spectator mode relevant here?

src/Blocks/BlockEndPortalFrame.h
```diff
@@ -0,0 +1,239 @@
...
+		}
+		return 1;
+	}
+
+
+
+
+
+	/** Check a specific block to see if it is a valid portal frame. */
+	int IsValidFrame(Vector3i a_Pos, NIBBLETYPE a_ShouldFace, cChunkInterface & a_ChunkInterface)
```
bearbin: This should return the enum type should it not?

src/Items/ItemEyeOfEnder.h
earboxer: In spectator mode, you can't use items.

The item should be used up in Survival and Adventure modes.

src/Blocks/BlockEndPortalFrame.h
earboxer: There is no space before this `!`.

Since it's an operator, I have a space after it.

bearbin: Sorry, a slip of the fingers. I meant after. Even though `!` is an operator, it is an exception to the general rule and has no spaces before or after.

peterbell10:
For reference here's an attempt that I didn't quite get round to finishing:
https://github.com/cuberite/cuberite/compare/master...peterbell10:EyeOfEnder

src/Blocks/BlockEndPortalFrame.h
earboxer: I was initially copying the way the nether portal had done it. (`BlockFire.h`) But it makes sense to move it.

src/Blocks/BlockEndPortalFrame.h
earboxer: Globally, they wouldn't be too useful, since the orientation of the portal blocks is not the same as the orientation of all other blocks. But I can make them static const.

src/Items/ItemEyeOfEnder.h
earboxer: Got it.

earboxer:
Sure. I'll work on making it more readable.

earboxer:
Travis error
```
Globals.E_META_END_PORTAL_EYE (constant)
Globals.E_META_END_PORTAL_FRAME_XM (constant)
Globals.E_META_END_PORTAL_FRAME_XM_EYE (constant)
Globals.E_META_END_PORTAL_FRAME_XP (constant)
Globals.E_META_END_PORTAL_FRAME_XP_EYE (constant)
Globals.E_META_END_PORTAL_FRAME_ZM (constant)
Globals.E_META_END_PORTAL_FRAME_ZM_EYE (constant)
Globals.E_META_END_PORTAL_FRAME_ZP (constant)
Globals.E_META_END_PORTAL_FRAME_ZP_EYE (constant)
Globals.E_META_END_PORTAL_NO_EYE (constant)
```

Is there somewhere else these need to be added?

peterbell10:
You need to add documentation to [Server/Plugins/APIDump/APIDesc.lua](https://github.com/cuberite/cuberite/blob/177273006e121acdcedbe74e3ee4988439cd10e3/Server/Plugins/APIDump/APIDesc.lua#L16276).

earboxer:
I've unmarked this as "WIP". Probably not going to fix the rest of the problems with it (since I haven't touched it in over a year. It half-fixes two issues, and includes TODOs in the code for fixing the rest of the problems. It would be better to have this than to not to.

peterbell10:
I agree with @bearbin's review in that your portal finding algorithm seems a bit confusing. Obviously I'm biased but my version seems to solve the problem just as well while being (IMO) more straight forward and easier to read. 

What are your thoughts: do you think your version is better in some way? How would you feel about merging our versions together?

earboxer:
My version allows for the minimum and maximum dimensions of the portal to be changed. (allows 3x3, 3x4, 4x3, and 4x4 portals). Since we haven't designed strongholds yet, a designer could have the dimensions based on an even number or an odd number and still be able to center the portal in the room. (remembering how java edition 1.7.2 allowed nether portals to be defined as a range of rectangular sizes)

My version also allows portal blocks to be placed within a portal, which is consistent with 1.12 behavior.

peterbell10:
>My version also allows a portal to form even if there are portal frames inside of it, which is consistent with 1.12 behavior.

Fair enough, I didn't realise the behaviour had changed in 1.12. I suppose that justifies a little extra complexity but I still thnk your version could be simpler. I'll have a think about it.

peterbell10:
I had a go at improving the readability, please take a look at the [commit](https://github.com/cuberite/cuberite/commit/7860463721e77fb0b3e24ce3bbd9a4ff26f9ffba) and tell me what you think.

I haven't changed the logic of the algorithm at all, except that the width of the 0th edge is now counted slightly differently. Other than that, I just tried to give things better names and make the control flow a bit more straightforward.

earboxer:
Looks good (assuming it still works fine).

One thing I'd note is that it seems a little weird to have

`IsValidFrameAtPos(cChunkInterface & a_ChunkInterface, Vector3i a_BlockPos, NIBBLETYPE a_ShouldFace)`

*and*

`bool IsPortalFrame(BLOCKTYPE BlockType)`

Is this calling getBlock/getBlockTypeMeta more than the previous code? Does it matter?

peterbell10:
>Is this calling getBlock/getBlockTypeMeta more than the previous code?

This version calls each of `GetBlock` and `GetBlockTypeMeta` exactly once per loop iteration.
The previous version calls `GetBlockTypeMeta` once for corners but twice for edges.

For corners this is slightly more work. However, there are more edges than corners and `GetBlock` should be slighly cheaper than `GetBlockTypeMeta`. So the more common case may be slightly cheaper.

>Does it matter?

The loop is only doing a small number of iterations anyway, so I think it's fast enough either way.

@earboxer would it be alright if I commit the code to this PR?

earboxer:
Sure, that's fine with me!

PR #4128, opened 2018-01-02
cLuaState: Update Push for entities.

### LINKED ISSUE #4127: cItemFrame class not implemented
Client version:   1.12
Server OS:        Windows 7
Commit id:        612

I'm trying to access cItemFrame:GetItem and cItemFrame:GetItemRotation but the methods doesn't seem to be implemented. Example:

``` lua
function Initialize(Plugin)
    Plugin:SetName("Testing")
    Plugin:SetVersion(1)
    LOG("Initialised " .. Plugin:GetName() .. " v." .. Plugin:GetVersion())
    cPluginManager:AddHook(cPluginManager.HOOK_PLAYER_RIGHT_CLICKING_ENTITY, OnPlayerRightClickingEntity)
    return true
end

function OnPlayerRightClickingEntity(Player, Entity)
    if Entity:IsItemFrame() then
        -- attempt to call method 'GetItemRotation' (a nil value)
        Player:SendMessage("GetItemRotation: " .. Entity:GetItemRotation())
    elseif Entity:IsMob() then
        -- working perfectly
        Player:SendMessage("Mob age: " .. Entity:GetAge())
    end
    
    return false
end

```

peterbell10:
Fixes #4127

Some classes were exported but were only pushed as a `cEntity` meaning exported functions were inaccessible.  This brings `cLuaState::Push(cEntity *)` up to date with all exported entities.

### DIFF
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -960,54 +960,42 @@ void cLuaState::Push(cEntity * a_Entity)
 	}
 	else
 	{
-		switch (a_Entity->GetEntityType())
-		{
-			case cEntity::etMonster:
+		const char * ClassName = [&]
 			{
-				// Don't push specific mob types, as those are not exported in the API:
-				tolua_pushusertype(m_LuaState, a_Entity, "cMonster");
-				break;
-			}
-			case cEntity::etPlayer:
-			{
-				tolua_pushusertype(m_LuaState, a_Entity, "cPlayer");
-				break;
-			}
-			case cEntity::etPickup:
-			{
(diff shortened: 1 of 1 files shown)


peterbell10:
@bearbin The lambda means I can write
```cpp
case cEntity::etBoat:         return "cBoat";
```
instead of
```cpp
case cEntity::etBoat:         ClassName = "cBoat"; break;
```
Which imo is cleaner.  I can change it if you don't like it.

bearbin:
I don't have a strong feeling either way, it just felt a bit more awkward to me. Keep it how it is.

PR #4130, opened 2018-01-03
Golden Apple Fixes

### LINKED ISSUE #4087: Golden apple does not take effect
Client version:   1.12.2
Server OS:        Windows/Linux/macOS/BSD/...
Commit id:        all

### Expected behavior
Eat a golden apple, got corresponding effect

### Actual behavior
Eating (enchanted) golden apple has no effect.

As above.

The golden apple handler is special, but it's changed in commit 10d42a2452e19ca98506df4d85c0e8f37b3e8981.

A bonus bug: the item is consumed before the saturation check. Which means, not consuming the food when saturated relies on client side check and the check in HandleRightClick. If there's a buggy implementation of golden apple, the food can be consumed in vain.

IMHO, consuming the item should be in cPlayer::FinishEating, otherwise there's no clean way to have the item consumed with recursive call of EatItem function.

bearbin:
Fixes #4087

### DIFF
--- a/Server/crafting.txt
+++ b/Server/crafting.txt
@@ -50,7 +50,7 @@ DarkOakPlanks, 4 = DarkOakLog, *
 EnderChest       = EyeOfEnder,  2:2 | Obsidian, 1:1, 1:2, 1:3, 2:1, 2:3, 3:1, 3:2, 3:3
 Furnace          = Cobblestone, 1:1, 1:2, 1:3, 2:1, 2:3, 3:1, 3:2, 3:3
 JunglePlanks, 4  = JungleLog, *
-OakPlanks, 4     = OakLog, * 
+OakPlanks, 4     = OakLog, *
 SprucePlanks, 4  = SpruceLog, *
 Stick, 4         = Planks^-1, 2:2, 2:3
 Torch, 4         = Stick, 1:2 | Coal^-1, 1:1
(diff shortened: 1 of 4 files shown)


bearbin:
This isn't really related to this PR as such, but I was wondering why golden apples have their own item handler while other food doesn't (even if they give status effects e.g. rotten flesh or give items back e.g. soup) - is there something I'm missing differentiating them?

src/Entities/Player.h
```diff
@@ -331,6 +332,16 @@ class cPlayer :
 	void SetFoodTickTimer             (int a_FoodTickTimer);
 	void SetFoodExhaustionLevel       (double a_FoodExhaustionLevel);
 
+	// tolua_end
+
+	/** Adds to FoodLevel and FoodSaturationLevel, returns true if any food has
+	been consumed, false if player "full". Convenience method that simply calls
+	Feed(a_Food, a_Saturation) with the values from FoodInfo.
+	*/
+	bool Feed(cItemHandler::FoodInfo a_FoodInfo);
```
peterbell10: I don't really see the point of adding this to `cPlayer`.  Especially if it means adding more header include dependencies.

peterbell10:
>I was wondering why golden apples have their own item handler while other food doesn't

There used to be a function `GetEatEffect` which allowed only one effect to be added without creating an item handler.  That made Golden apples a special case but it was removed in 10d42a2452e19ca98506df4d85c0e8f37b3e8981.

Other items probably should have their own item handlers as well instead of having big switches in `cItemFoodHandler` though.

src/Entities/Player.h
bearbin: I can remove this if you want, I just felt that `Feed()`ing a player should be possible with a single component representation of food.

bearbin:
Thanks for the pointer, I'll make a note to separate the other food items out in due course.

src/Entities/Player.h
peterbell10: I suppose another way to avoid having to include ItemHandler.h is to have `FoodInfo` in a namespace rather than nested in `cItemHandler`.  Then it can just be forward declared.

bearbin: I'll just remove the function instead, it's not really too important.

PR #4131, opened 2018-01-04
CMake: Fix test builds on MSVC

peterbell10:
[Linux builds on Jenkins](https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/928/console) are failing with this cmake error:
```
CMake Error at CMakeLists.txt:95 (add_executable):
  Target "ProtoProxy" links to target "fmt::fmt" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```
because ProtoProxy is being built as the master project and it is missing the call to `add_subdirectory` for `fmt`.  This adds the missing calls and solves the resulting duplication caused when the tools are built as a subproject.

Questions:
* Is it worth adding a `cmake` folder or should I just put the scripts at top level?  If it stays then it might be worth moving `CheckLua.cmake` and `SetFlags.cmake` into the folder as well.
* After finding the solution in `cmake/NoDuplicateSubdirectories.cmake` I realised it's probably solving the same issue as `lib/mbedtls.cmake` but in a more general way.  Which solution should be preferred and should the other be rewritten in the same way?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,14 @@
 version: 1.0.{build}
 configuration: Debug
 clone_depth: 50
+
 before_build:
 - echo %TIME%
 - git submodule update --init
 - echo %TIME%
-- cmake -G "Visual Studio 12" .
+- cmake -G "Visual Studio 12" -DSELF_TEST=1 -DBUILD_TOOLS=1 .
 - echo %TIME%
+
 build:
   project: Cuberite.sln
   parallel: true
--- a/tests/OSSupport/CMakeLists.txt
+++ b/tests/OSSupport/CMakeLists.txt
@@ -23,6 +23,7 @@ if (WIN32)
 		${OSSupport_SRCS}
 		${OSSupport_HDRS}
 	)
+	target_link_libraries(OSSupport PUBLIC fmt::fmt)
 endif()
 
 # Define individual tests:


bibo38:
> Is it worth adding a cmake folder or should I just put the scripts at top level? If it stays then it might be worth moving CheckLua.cmake and SetFlags.cmake into the folder as well.

I would definitely move every *.cmake to a common folder. So `lib/mbedtls.cmake` should also go into the folder.

 > because ProtoProxy is being built as the master project

My fix to this Problem: Change the build script!
Explanation: The idea of choosing every directory as a CMake source is nice. But the idea behind this is, that I can simply copy folder X to my project and use the library/executable in my CMake project. To achieve this properly, we need to layer the folders. E.g.:

    ProtoProxy > Cuberite-X > fmt
                      > Cuberite-Y

Now, I would be able to build Cuberite-Y, Cuberite-X, fmt and ProtoProxy independently.
The main problem with this approach is the duplication of common dependencies like fmt, Test-Frameworks, ... 

When you now use `add_subdirectory(../../i/really/know/my/parents)`, you assume things about your parent folders. It's like using `..` in Includes: `#include "../../../usr/include/dbus-1.0/dbus/dbus.h"`. Now this file works fine on my system. But moving the file could cause problems.

The second problem are the duplicated subdirectories. While your CMakeScript can fix the double inclusion, you need to do workarounds for the configuration of that subproject. This is why `mbedtls.cmake` exists! It sets the appropriate configuration options and then includes the subdirectory. Without this script, the configuration options would need to be duplicated...

My Solution is simple: Only use one folder as central folder. And if you only want to build ProtoProxy, then generate the Makefiles, go into `Tools/ProtoProxy` and hit `make`. Now CMake will only compile ProtoProxy. And you don't have folder duplication problems or *special* files like `mbedtls.cmake`.

peterbell10:
Do you have any thoughts about @bibo38's idea of changing the scripts used by Jenkins instead?

bibo38:
@bearbin By the way: Is the build script in the Cuberite repo or only in the Jenkins installation?

My idea is this build script:

    mkdir build && cd build
    cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_TOOLS=ON ..
    cd Tools/ProtoProxy
    make

Just tested it on the current master and strangely it also finds fmt 😕 . Because the tools are included before the libraries in the global `CMakeLists.txt`, it shouldn't find them.

bearbin:
For some builds the script is only in jenkins, e.g. https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/

For some, the script is stored in the build-scripts repo: https://builds.cuberite.org/job/cuberite/job/master/job/linux-x64/

I think changing the build scripts would probably be the more sensible idea.

bibo38:
> I think changing the build scripts would probably be the more sensible idea.

Just to be sure: Do you mean the Jenkins build scripts? Or the CMake-Scripts?

bearbin:
Jenkins scripts was what I meant.

peterbell10:
[Looks like](https://cuberite.org/status/) @bearbin got it working 👍

peterbell10:
https://builds.cuberite.org/job/Cuberite-FreeBSD-x64-Master/799/console

Looks like the build script needs to be updated for the FreeBSD build as well.

PR #4137, opened 2018-01-07
Generate cactuses with height according to MaxCactusHeight in world.ini

### LINKED ISSUE #4135: Generated cactuses are only one block tall in desert biomes
Client version:   All supported (1.8->1.12.2)
Server OS:        Raspbian 9 Stretch (GNU/Linux)
Commit id:        757231cc6e777b8f4717d1467ef7efa01c7fde15

### Expected behavior
Cactuses structures are 1, 2 or 3 block tall

(image taken in a local Minecraft world)

### Actual behavior
Cactuses structures are only 1 block tall 

(image taken on Cuberite instance)

Simply getting into a desert biome with cactuses. Seed of the world with the bug is 300388357.

### Notes
I am more or less at ease with C++, and would love to solve this bug by myself. The sources being quite large, could I have some clues about where to find the files involved ? Thank you very much !

Cocosushi6:
When generating foliage, create cactuses whose height vary in the interval [1; MaxCactusHeight] (with MaxCactusHeight being declared in world.ini) 

Fixes #4135


src/Generating/FinishGen.cpp
```diff
@@ -14,6 +14,7 @@
 #include "../Simulator/FireSimulator.h"
 #include "../World.h"
 #include "../IniFile.h"
+#include "../BlockEntities/BlockEntity.h"
```
bearbin: Why has this include been added?

src/Generating/FinishGen.cpp
```diff
@@ -801,6 +802,11 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
 					)
 					{
 						a_ChunkDesc.SetBlockType(x, ++Top, z, E_BLOCK_CACTUS);
+
+						for (int i = 0; i < (x % m_MaxCactusHeight); i++)
+						{
+							a_ChunkDesc.SetBlockType(x, ++Top, z, E_BLOCK_CACTUS);
```
bearbin: This code is a bit confusing, could you use `Top + i + 1` instead?

src/Generating/FinishGen.cpp
Cocosushi6: I think it's a leftover from some experimentations (was trying to get the MaxCactusHeight variable from ini files). I'll remove it, as it's useless indeed :)

src/Generating/FinishGen.cpp
Cocosushi6: Yup, i'll replace `++Top`. Apparently, it's also the case for sugar canes - I'll include a fix for it too

src/Generating/FinishGen.h
```diff
@@ -260,11 +260,13 @@ class cFinishGenSprinkleFoliage :
 	public cFinishGen
 {
 public:
-	cFinishGenSprinkleFoliage(int a_Seed) : m_Noise(a_Seed), m_Seed(a_Seed) {}
+	cFinishGenSprinkleFoliage(int a_Seed, int a_MaxCactusHeight, int a_MaxSugarcaneHeight) : m_Noise(a_Seed), m_Seed(a_Seed), m_MaxCactusHeight(a_MaxCactusHeight), m_MaxSugarcaneHeight(a_MaxSugarcaneHeight) {}
```
peterbell10: This line is getting rather long, you should break it up:
```cpp
cFinishGenSprinkleFoliage(int a_Seed, int a_MaxCactusHeight, int a_MaxSugarcaneHeight):
    m_Noise(a_Seed),
    m_Seed(a_Seed),
    m_MaxCactusHeight(a_MaxCactusHeight),
    m_MaxSugarcaneHeight(a_MaxSugarcaneHeight)
{
}
```

Cocosushi6:
Ok, I'll do it now. Thanks for accepting the pull request !
The values in World.ini were already present before, they're in the default example [here](https://book.cuberite.org/#3), as well as in the (deprecated, I think) wiki [here](https://github.com/cuberite/cuberite/wiki/Configuring-World.ini#plants-section)

src/Generating/FinishGen.cpp
```diff
@@ -801,6 +805,11 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
 					)
 					{
 						a_ChunkDesc.SetBlockType(x, ++Top, z, E_BLOCK_CACTUS);
+
+						for (int i = 0; i < (x % m_MaxCactusHeight); i++)
```
peterbell10: `(x % m_MaxCactusHeight)` isn't the ideal way to do this because the cacti will line up in sawtooth patterns.  You should use `m_Noise` to make it look more random.

src/Generating/FinishGen.cpp
```diff
@@ -801,6 +805,11 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
...
+						{
+							a_ChunkDesc.SetBlockType(x, Top + i + 1, z, E_BLOCK_CACTUS);
```
peterbell10: You do need to modify the Top variable because it's used to set the chunk's heightmap.  This will leave the heightmap wrong.  Same with the sugarcane.

src/Generating/FinishGen.cpp
```diff
@@ -708,8 +708,12 @@ bool cFinishGenSprinkleFoliage::TryAddSugarcane(cChunkDesc & a_ChunkDesc, int a_
 		return false;
 	}
 
-	// All conditions met, place a sugarcane here:
-	a_ChunkDesc.SetBlockType(a_RelX, a_RelY + 1, a_RelZ, E_BLOCK_SUGARCANE);
+	// All conditions met, place a sugarcane here
+	for (int i = 0; i < (a_RelX % m_MaxSugarcaneHeight); i++)
+	{
+		a_ChunkDesc.SetBlockType(a_RelX, a_RelY + i + 2, a_RelZ, E_BLOCK_SUGARCANE);
```
peterbell10: I don't think `+ 2` is right here, won't the sugarcanes hover above the ground?

Cocosushi6:
Hello and sorry for the huge delay !
I'll implement this as soon as possible, although I have one slight little problem with the use of m_Noise : how can I use it to generate a more random-like height ? It has multiple methods and I don't really get which one is appropriate for this use. 
Thank you very much for your help :)

src/Generating/FinishGen.cpp
Cocosushi6: Yes indeed, I'll change this

peterbell10:
I believe something as simple as this should do the job:
```cpp
int Height = 1 + (m_Noise.IntNoise2DInt(x, z) % m_MaxCactusHeight);
```

Cocosushi6:
Ok, I think I have addressed the latest issues. I also created a new method for generating cactuses, and modified the one to generate sugar canes for them to create a full height cacti/sugarcane instead of only one block. They still return true if they are able to place at least one block, and else return false (as the previous worked). 

Thank you very much for your help :)

src/Generating/FinishGen.cpp
```diff
@@ -665,51 +665,106 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
-			break;
+			return (i == 0) ? false : true;
 		}
-		default:
+
+		switch (a_ChunkDesc.GetBlockType(a_RelX, a_RelY, a_RelZ))
 		{
-			return false;
+			case E_BLOCK_SAND :
+			case E_BLOCK_CACTUS :
```
bibo38: You forgot red sand 😏 .

src/Generating/FinishGen.cpp
```diff
@@ -665,51 +665,106 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
+				return (i == 0) ? false : true;
+			}
 		}
+
+		int y = a_RelY + 1;
+		if ((
+			!(a_ChunkDesc.GetBlockType(a_RelX + 1, y, a_RelZ)     == E_BLOCK_AIR) ||
+			!(a_ChunkDesc.GetBlockType(a_RelX - 1, y, a_RelZ)     == E_BLOCK_AIR) ||
+			!(a_ChunkDesc.GetBlockType(a_RelX,     y, a_RelZ + 1) == E_BLOCK_AIR) ||
+			!(a_ChunkDesc.GetBlockType(a_RelX,     y, a_RelZ - 1) == E_BLOCK_AIR))
```
bibo38: A cactus can also be placed near some non solid blocks:


Based on my wiki these are:

- Sugarcane
- Lever
- Button
- Flower
- Sapling
- Armor Stand
- Torch
- Carpets
- Rails
- Snow
- Repeater
- Comparator

Maybe only sugarcane could create a problem, but I would suggest creating a function to check for the nearby block, something like `BlockInfo::CanBePlacedAgainstCactus`. Maybe the code handling the cactus destructon on block place already has such a function ❓

src/Generating/FinishGen.cpp
```diff
@@ -665,51 +665,106 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
 
 
 
-
-
 ////////////////////////////////////////////////////////////////////////////////
 // cFinishGenSprinkleFoliage:
 
-bool cFinishGenSprinkleFoliage::TryAddSugarcane(cChunkDesc & a_ChunkDesc, int a_RelX, int a_RelY, int a_RelZ)
+bool cFinishGenSprinkleFoliage::TryAddCactus(cChunkDesc &a_ChunkDesc, int a_RelX, HEIGHTTYPE & a_RelY, int a_RelZ)
```
peterbell10: Style: There should be a space between `&` and `a_ChuckDesc`

src/Generating/FinishGen.cpp
peterbell10: I would prefer `(a != b)` to `!(a == b)`.

src/Generating/FinishGen.cpp
```diff
@@ -665,51 +665,106 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
+			}
+			default:
+			{
+				return (i == 0) ? false : true;
+			}
+		}
+
+		// Water is required next to the block below the sugarcane (if the block below isn't sugarcane already)
+		if (
+			!IsWater(a_ChunkDesc.GetBlockType(a_RelX - 1, a_RelY, a_RelZ)) &&
```
peterbell10: Redundant check: after first iteration it is guaranteed to be a sugarcane.

src/Generating/FinishGen.h
```diff
@@ -260,14 +260,24 @@ class cFinishGenSprinkleFoliage :
...
+	int    m_MaxCactusHeight;
+	int    m_MaxSugarcaneHeight;
+
+	/** Tries to place sugarcane at the coords specified, returns true if successful, updates the top variable (hence the & a_RefY) */
+	bool TryAddSugarcane(cChunkDesc & a_ChunkDesc, int a_RelX, HEIGHTTYPE & a_RelY, int a_RelZ);
 
-	/** Tries to place sugarcane at the coords specified, returns true if successful */
-	bool TryAddSugarcane(cChunkDesc & a_ChunkDesc, int a_RelX, int a_RelY, int a_RelZ);
+	/** Tries to place cactus at the coords specified, returns true if successful, updates the top variable (hence the & a_RefY) */
+	bool TryAddCactus(cChunkDesc & a_ChunkDesc, int a_RelX, HEIGHTTYPE & a_RelY, int a_RelZ);
```
peterbell10: Could you just return the updated height rather than via an in-out parameter.  I think it's more readable that way.

src/Generating/FinishGen.cpp
```diff
@@ -775,36 +830,26 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
...
-						(x > 0) && (x < cChunkDef::Width - 1) &&
-						(z > 0) && (z < cChunkDef::Width - 1) &&
-						(val1 + val2 > 0.5f) &&
-						(a_ChunkDesc.GetBlockType(x + 1, y, z)     == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x - 1, y, z)     == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x,     y, z + 1) == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x,     y, z - 1) == E_BLOCK_AIR) &&
-						IsDesertVariant(a_ChunkDesc.GetBiome(x, z))
-					)
+					if (val1 + val2 > 0.5f)
```
peterbell10: This change will make sugar cane rarer in non-desert biomes.  You should keep the `IsDesertVariant` in the if.

src/Generating/FinishGen.cpp
```diff
@@ -665,51 +665,106 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
-	switch (a_ChunkDesc.GetBlockType(a_RelX, a_RelY, a_RelZ))
+	for (int i = 0; i < 1 + (m_Noise.IntNoise2DInt(a_RelX, a_RelZ) % m_MaxCactusHeight); i++)
 	{
-		case E_BLOCK_DIRT:
-		case E_BLOCK_GRASS:
-		case E_BLOCK_SAND:
+		if (
+			(a_RelX < 1) || (a_RelX >= cChunkDef::Width - 1) ||
+			(a_RelY < 1) || (a_RelY >= cChunkDef::Height - 2) ||
+			(a_RelZ < 1) || (a_RelZ >= cChunkDef::Width - 1)
```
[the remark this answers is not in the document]
peterbell10: `(a_RelY >= cChunkDef::Height - 2)` is because the cactus is placed _above_ `a_RelY`.  I agree `a_RelY < 1` seems off though.

src/Generating/FinishGen.cpp
peterbell10: For that to make a difference, the switch in `GenFinish` would need updating as well.

src/Generating/FinishGen.cpp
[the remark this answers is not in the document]
peterbell10: I'm not sure there's enough meaning there to make a meaningful name.

src/Generating/FinishGen.h
bibo38: > Could you just return the updated height rather than via an in-out parameter. I think it's more readable that way.

I wouldn't agree, because you could forget to update the height (the function name doesn't imply to return a height. So calls like `TryAddCactus(chunk, relX, relY, relZ)` don't look wrong). 

But maybe some sort of transparent BlockSetter interface, which abstracts the height change away would maybe be the superior way:

```cpp
void TryAddSugarcane(Vector pos, BlockSetter & setter)
{
    setter.setBlockAt(1, 2, 3, E_STONE);
    setter.setBlockAt(pos, E_SUGARCANE);
}
```

peterbell10: >I wouldn't agree, because you could forget to update the height

You could equally well forget to update the height returned by the out param.  You still need to call `a_ChunkDesc.SetHeight` either way.

In actual fact, why not just update the heightmap in the `TryPlace` function.

bibo38: > In actual fact, why not just update the heightmap in the TryPlace function.

I don't see, where `TryPlace` is being used. I would have used the `SetBlockType` function.

peterbell10: I meant `TryPlaceCactus` and `TryPlaceSugarcane`.

bibo38: Yeah, that would also be better. Is there any reason, not to put the heightmap thing into `SetBlockType`, etc. functions?

peterbell10: I would assume it's to avoid doing unnecessary work.  It would be interesting to see a profile/benchmark though.

Cocosushi6: Indeed, I think putting it in the `TryPlace` functions is a better solution - I had kept the boolean return value in case making checks to see if a sugarcane or a cacti are placeable was ever needed.

src/Generating/FinishGen.cpp
Cocosushi6: Isn't red sand a meta type of the `E_BLOCK_SAND` (l. 767 in `BlockID.h`) ? Using `E_BLOCK_SAND` does not handles both ?

peterbell10: [You're right](https://minecraft.gamepedia.com/Sand)

src/Generating/FinishGen.cpp
```diff
@@ -665,51 +665,106 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
+		)
 		{
-			break;
+			return (i == 0) ? false : true;
```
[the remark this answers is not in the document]
Cocosushi6: Well, I didn't remember having been drunk the evening I coded this... will remove ! Although I think putting a comment would be more explicit and avoid creating a variable for readability ?

src/Generating/FinishGen.cpp
bibo38: Oh men. Must have been the day, I've learned about the meta removal in 1.13.

src/Generating/FinishGen.cpp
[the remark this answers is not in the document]
bibo38: I prefer code, because comments get more easily out of sync.
In your case, the variable even reduces the repetition of `i != 0` (and maybe the repetition of the comment explaining this).

But this is just a personal preference. You can also use a comment.

src/Generating/FinishGen.cpp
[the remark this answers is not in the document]
bibo38: But consider the Chunk height being `16`.
Now this code exits the function, when `a_RelY` is 14 (= 16 - 2).
But the cacus would be placed (if the condition wouldn't prevent this) at `++a_RelY` (= 15), which is a valid location.

Cocosushi6:
Hello, and sorry again for the delay :( 
I think I have addressed the latest issues, including generating close to non-solid blocks (in case new ones appear, I used the `cBlockInfo::IsSolid` property). I noticed a bug though, which was present without my changes (tested with latest version from the website, build #972) : sugarcanes are very rare in non-desert biomes, even after leaving the `IsDesertVariant` in the if. I suspect it comes from the `case E_BLOCK_GRASS` in `cFinishGenSprinkleFoliage::GenFinish`. Should I file a new bug ? It is more in relation with noise-generation, with which I am not familiar at all (learning them is still an option though). 
Thanks again for your support :)

peterbell10:
>Should I file a new bug ?

Absolutely, if you think it's seriously rare then it should be noted.  I certainly wouldn't expect you to fix it in this PR.

src/Generating/FinishGen.cpp
```diff
@@ -776,36 +837,26 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
...
-						++Top;
-					}
 					else if ((val1 > 0.5) && (val2 < -0.5))
 					{
 						a_ChunkDesc.SetBlockTypeMeta(x, ++Top, z, E_BLOCK_PUMPKIN, static_cast<int>(val3 * 8) % 4);
 					}
+					else
+					{
+						TryAddSugarcane(a_ChunkDesc, x, Top, z);
+					}
```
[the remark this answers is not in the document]
Cocosushi6: I had changed it because updating the `Top` variable was now useless (as it was done inside the `TryAddSugarcane` function), but indeed, this could be one of the reasons of its rarity. I'll change it.

src/Generating/FinishGen.cpp
```diff
@@ -666,29 +666,72 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
+
 	// We'll be doing comparison to neighbors, so require the coords to be 1 block away from the chunk edges:
 	if (
-		(a_RelX < 1) || (a_RelX >= cChunkDef::Width  - 1) ||
-		(a_RelY < 1) || (a_RelY >= cChunkDef::Height - 2) ||
-		(a_RelZ < 1) || (a_RelZ >= cChunkDef::Width  - 1)
-	)
+			(a_RelX < 1) || (a_RelX >= cChunkDef::Width - 1) ||
+			(a_RelZ < 1) || (a_RelZ >= cChunkDef::Width - 1) ||
+			(a_RelY + CactusHeight >= cChunkDef::Height - 1)
```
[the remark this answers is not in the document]
bibo38: Also the conditions seem to be indented twice:
indent

(The dark green indention shouldn't be there)

src/Generating/FinishGen.cpp
```diff
@@ -776,36 +837,26 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
...
-						(a_ChunkDesc.GetBlockType(x + 1, y, z)     == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x - 1, y, z)     == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x,     y, z + 1) == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x,     y, z - 1) == E_BLOCK_AIR) &&
-						IsDesertVariant(a_ChunkDesc.GetBiome(x, z))
-					)
+					if (val1 + val2 > 0.5f && IsDesertVariant(a_ChunkDesc.GetBiome(x, z)))
 					{
-						a_ChunkDesc.SetBlockType(x, ++Top, z, E_BLOCK_CACTUS);
+						TryAddCactus(a_ChunkDesc, x, Top, z);
```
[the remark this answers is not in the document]
bibo38: A seperate `if` condition is better, because you need the same for the sugarcane. E.g. if sugarcane can't be placed, try cacti.
The best construct I currently can think of:
```cpp
if(val1 + val2 > 0.5f)
{
  if(!TryAddCactus(a_ChunkDesc, x, Top, z))
  {
    TryAddSugarcane(a_ChunkDesc, x, Top, z);
  }
}
else
{
  if(!TryAddSugarcane(a_ChunkDesc, x, Top, z))
  {
    TryAddCactus(a_ChunkDesc, x, Top, z);
  }
}
```
Note: The code above implies, that `IsDesertVariant` is checked inside the `TryAddCactus` function.

src/Generating/ComposableGenerator.cpp
```diff
@@ -525,7 +525,9 @@ void cComposableGenerator::InitFinishGens(cIniFile & a_IniFile)
 		}
 		else if (NoCaseCompare(finisher, "SprinkleFoliage") == 0)
 		{
-			m_FinishGens.push_back(cFinishGenPtr(new cFinishGenSprinkleFoliage(Seed)));
+			int MaxCactusHeight 	= a_IniFile.GetValueI("Plants", "MaxCactusHeight", 3);
+			int MaxSugarcaneHeight 	= a_IniFile.GetValueI("Plants", "MaxSugarcaneHeight", 3);
+			m_FinishGens.push_back(cFinishGenPtr(new cFinishGenSprinkleFoliage(Seed, MaxCactusHeight, MaxSugarcaneHeight)));
```
bibo38: I would suggest to use the [`std::make_shared`](http://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared) function. [Stack Overflow](https://stackoverflow.com/questions/20895648/difference-in-make-shared-and-normal-shared-ptr-in-c/20895705#20895705) has a good explanation of the benefits.

src/Generating/FinishGen.cpp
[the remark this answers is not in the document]
peterbell10: > if sugarcane can't be placed, try cacti

It didn't behave like that before so it shouldn't be changed to do that.  This PR should only change the height of sugarcanes and cacti but not where they generate.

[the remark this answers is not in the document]
bibo38: Yep, you're right... I overlooked this behaviour in the old code.
But please don't mix the `TryAddX` functions into the `val1 + val2` condition, when fixing this:
```cpp
if(val1 + val2 <= 0.5f || !TryAddCactus(a_ChunkDesc, x, Top, z))
{
  TryAddSugarcane(a_ChunkDesc, x, Top, z);
}
```

Cocosushi6:
Alright, I have addressed the latest issues. Thanks !

src/Generating/FinishGen.cpp
```diff
@@ -666,29 +666,72 @@ void cFinishGenVines::GenFinish(cChunkDesc & a_ChunkDesc)
...
+		{
+			return cactusExists;
+		}
+
+		// All conditions are met, we can place a cactus here
+		a_ChunkDesc.SetBlockType(a_RelX, ++a_RelY, a_RelZ, E_BLOCK_CACTUS);
+	}
+
+	// Update chunk height after adding cactus
+	a_ChunkDesc.SetHeight(a_RelX, a_RelZ, a_RelY);
```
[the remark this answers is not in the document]
peterbell10: To be clear, the reason I first suggested having `SetHeight` in these functions was to avoid needing an 'in-out' param. It would have also need some reworking of the logic in `GenFinish` to not use the `Top` variable.  As it is currently, this call is redundant and in fact for `TryAddCactus`, it doesn't even update the heightmap when it bails out early.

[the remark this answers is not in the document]
Cocosushi6: Yup, I intended to remove them in the new commit, but forgot. I'll remove them.

Cocosushi6:
Alright, thanks !

src/Generating/FinishGen.cpp
```diff
@@ -789,29 +846,22 @@ void cFinishGenSprinkleFoliage::GenFinish(cChunkDesc & a_ChunkDesc)
...
-						(x > 0) && (x < cChunkDef::Width - 1) &&
-						(z > 0) && (z < cChunkDef::Width - 1) &&
-						(val1 + val2 > 0.5f) &&
-						(a_ChunkDesc.GetBlockType(x + 1, y, z)     == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x - 1, y, z)     == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x,     y, z + 1) == E_BLOCK_AIR) &&
-						(a_ChunkDesc.GetBlockType(x,     y, z - 1) == E_BLOCK_AIR) &&
-						IsDesertVariant(a_ChunkDesc.GetBiome(x, z))
-					)
+					if (val1 + val2 > 0.5f && IsDesertVariant(a_ChunkDesc.GetBiome(x, z)))
```
bibo38: Because all other checks (like if a cacti is near a solid block) moved to `TryAddCactus`, I would also move this desert check to `TryAddCactus`.

bibo38:
By the way: If you haven't allowed contributors to change your branch, you'll need to rebase your changes ontop of master (otherwise they cannot be merged by people like @peterbell10 ).

Cocosushi6:
Yup, will do. Thanks !

PR #4140, opened 2018-01-09
Travis: Remove gdb

peterbell10:
Sudo-less workers no longer have permission to disable ASLR.
This is seen [here](https://travis-ci.org/cuberite/cuberite/builds/326872644) where each failed build has this error message from gdb:

>warning: Error disabling address space randomization: Operation not permitted

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,27 +18,18 @@ matrix:
         - TRAVIS_CUBERITE_BUILD_TYPE=DEBUG CUBERITE_PATH=./Cuberite_debug
     # Default clang
     - compiler: clang
-      addons: &gdb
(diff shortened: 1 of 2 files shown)


madmaxoft:
Is there any way to grab a stacktrace if the executable segfaults on Travis? That's the main reason for using gdb in the first place

peterbell10:
Would installing the signal handlers that call `PrintStackTrace` in debug builds be enough?

gdb does seem to be completely broken by TravisCI's latest update though. At this point gdb doesn't seem to be doing anything other than breaking the builds.  The only other option seems to be switching to `sudo: required` and putting up with the slower builds.

PR #4147, opened 2018-01-15
New movement system for leashed entities

### LINKED ISSUE #4146: entities are not flying with lead
Client version:   1.12.2
Server OS:        Windows 7
Commit id:       a72891fbb091440711cbbf95db7a433d641fcc6c

### Expected behavior
Mobs like cow or chicken should move and fly with player

### Actual behavior
Mobs are moving, but not flying

1. Spawn a mob and lead it
2. try to fly

and sorry for my bad English)

peterbell10:
Fixes #4146

Instead of pathfinding towards the leashed to entity, monsters are now accelerated towards the leashed to entity as if they were attached by a spring.

Here's a gif of it in action: https://gfycat.com/DaringUniformAtlasmoth

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -22,10 +22,6 @@
 
 
 
-// Ticks to wait to do leash calculations
-#define LEASH_ACTIONS_TICK_STEP 10
-
-
 
 /** Map for eType <-> string
 Needs to be alpha-sorted by the strings, because binary search is used in StringToMobType()
@@ -400,10 +396,7 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	}  // switch (m_EMState)
 
 	// Leash calculations
-	if ((m_TicksAlive % LEASH_ACTIONS_TICK_STEP) == 0)
-	{
-		CalcLeashActions();
-	}
+	CalcLeashActions(a_Dt);
 
 	BroadcastMovementUpdate();
 
@@ -422,7 +415,7 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
-void cMonster::CalcLeashActions()
+void cMonster::CalcLeashActions(std::chrono::milliseconds a_Dt)
 {
 	// This mob just spotted in the world and [m_LeashToPos not null] shows that should be leashed to a leash knot at m_LeashToPos.
 	// This keeps trying until knot is found. Leash knot may be in a different chunk that needn't or can't be loaded yet.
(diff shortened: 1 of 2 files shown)


src/Mobs/Monster.cpp
```diff
@@ -435,19 +428,41 @@ void cMonster::CalcLeashActions()
...
 		}
 	}
-	else if (IsLeashed())  // Mob is already leashed to an entity: follow it.
+
+	if (!IsLeashed())
+	{
+		return;
+	}
+
+	static const double LeashNaturalLength = 5.0;   // Length at equilibrium
```
bearbin: The comment here suggests that when entities are closer than the NaturalLength, they are pushed away. Perhaps "Length where the leash begins to pull" would be better? That doesn't sound great either so idk.

src/Mobs/Monster.cpp
```diff
@@ -435,19 +428,41 @@ void cMonster::CalcLeashActions()
...
-		// TODO: leashed mobs in vanilla can move around up to 5 blocks distance from leash origin
-		MoveToPosition(m_LeashedTo->GetPosition());
+		LOGD("Leash broken (distance)");
+		Unleash(false);
+	}
+	else if (Distance > LeashNaturalLength)
+	{
+		// Accelerate monster towards the leashed to entity:
+		auto Extension = Distance - LeashNaturalLength;
+		auto Direction = Displacement / Distance;
```
bearbin: Is this not just `Displacement.Normalize()`?

src/Mobs/Monster.cpp
```diff
@@ -435,19 +428,41 @@ void cMonster::CalcLeashActions()
...
+		// Stop mobs from floating up when on the ground
+		if (IsOnGround() && (Acceleration.y < std::abs(GetGravity())))
 		{
-			LOGD("Leash broken (distance)");
-			Unleash(false);
+			Acceleration.y = 0.0;
 		}
+
+		// Apply the acceleration
+		using namespace std::chrono;
```
bearbin: Does this really add enough to justify the extra line?

src/Mobs/Monster.cpp
```diff
@@ -435,19 +428,41 @@ void cMonster::CalcLeashActions()
...
-	else if (IsLeashed())  // Mob is already leashed to an entity: follow it.
+
+	if (!IsLeashed())
+	{
+		return;
+	}
+
+	static const double LeashNaturalLength = 5.0;   // Length at equilibrium
+	static const double LeashMaximumLength = 10.0;  // Length where the leash breaks
+	static const double LeashSpringConstant = 20.0;  // How elastic the leash is
```
bearbin: Similarly here, a higher value means less elastic, which is not the intuiitive meaning of the comment.

src/Mobs/Monster.cpp
peterbell10: Yes, equilibrium isn't quite right here.  I'll change it.

src/Mobs/Monster.cpp
peterbell10: I'm struggling to come think of language that isn't technical. It makes sense to me because the spring constant is proportional to the [elastic modulus](https://en.wikipedia.org/wiki/Young%27s_modulus).

Would "stiff" make more sense to you?

src/Mobs/Monster.cpp
peterbell10: Yes direction is the normalised displacement, I was just factoring out the length calculation as it's needed elsewhere.  I can change it to `NormalizeCopy` if you prefer.

src/Mobs/Monster.cpp
peterbell10: I find it more readable this way.  Cuberite's style conventions don't tend to be economic with line counts anyway.

peterbell10:
I took another look at the behaviour and realised it was a bit strange when the mob doesn't try to pathfind towards the player and just gets dragged behind. The reason I removed it in the first place was I didn't like how the mobs would stand on top of the player.  So instead of removing the pathing, I've fixed it with the `CloseFollowDistance`.

src/Mobs/Monster.cpp
```diff
@@ -435,19 +412,54 @@ void cMonster::CalcLeashActions()
...
+		using namespace std::chrono;
+		AddSpeed(Acceleration * duration_cast<duration<double>>(a_Dt).count());
+	}
+
+	// Passively follow the leashed to entity:
+	if (Distance > CloseFollowDistance)
+	{
+		const Vector3d TargetBlock((LeashedToPos - Direction * CloseFollowDistance).Floor());
+		// Move to centre of target block face
+		MoveToPosition(TargetBlock + Vector3d{ 0.5, 0.0, 0.5 });
```
bearbin: I'm not too familiar with the entity movement calculations, but is this movement cancelled when the leashed entity gets within the CloseFollowDistance? I feel like this could end up with mobs sitting on top of the player again.

peterbell10: You're right that the movement isn't cancelled because `MoveToPosition` isn't called.  However the mobs aren't pathing towards the block the player is standing on in the first place. Before rounding, the target block is the nearest point on a sphere of radius `CloseFollowDistance` centred on the the leashed to entity.

If you're wondering why the if is necessary then: it actually stops leashed mobs from walking away when the player walks within the `CloseFollowDistance`

bearbin: Thanks for explaining, I understand it now and the logic looks good.

bearbin:
The OSX builders on travis appear to be broken - but the build has passed otherwise for this PR.

peterbell10:
https://www.traviscistatus.com/ claims the build workers were broken but should be fixed and working through a backlog.

I would merge this but I don't have permission to merge without CI passing.  If you want to merge @bearbin, go ahead.

PR #4149, opened 2018-01-16
cClientHandle: Always call HOOK_PLAYER_RIGHT_CLICK

peterbell10:
Fixes the changes in hook calling from #4089.
This fixes WorldEdit's wand and compass for me.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1431,7 +1431,10 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 	cPluginManager * PlgMgr = cRoot::Get()->GetPluginManager();
 
 	bool Success = false;
-	if (IsWithinReach && !m_Player->IsFrozen())
+	if (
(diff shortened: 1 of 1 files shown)


bearbin:
I was waiting for you to reply to @9caihezi's comment in #4089:

> The assumption I made is that at most one of CallHookPlayerRightClick, CallHookPlayerUsingBlock,
CallHookPlayerUsingItem and CallHookPlayerEating is being called depends on what item you hold.
Does it make sense to use UsingItem callback instead of right click?

peterbell10:
The API docs make it clear that the hooks aren't be exclusive.  From [`HOOK_PLAYER_RIGHT_CLICKED`](https://api.cuberite.org/OnPlayerRightClick.html):

>This hook is called when Cuberite receives a right-click packet from the client. It is called before any processing whatsoever is performed on the packet, [...]
Plugins may refuse the default processing for the packet, causing Cuberite to behave as if the packet has never arrived.

So this hook should be called for every right click packet.

Issue #4132, opened 2018-01-04
Fishing rods should take different amounts of damage depending on use


bearbin:
Reference: https://minecraft.gamepedia.com/Item_durability

This depends on both #4123 and #4120 being merged.

PR #4151, opened 2018-01-17
Damage fishing rod when catching the hook on a block, then reeling it in

Bond-009:
#4132

### DIFF
--- a/src/Items/ItemFishingRod.h
+++ b/src/Items/ItemFishingRod.h
@@ -234,6 +234,14 @@ public:
 				a_Player->UseEquippedItem(1);
 				cRoot::Get()->GetPluginManager()->CallHookPlayerFished(*a_Player, Drops);
 			}
+			else
+			{
+				BLOCKTYPE Block = a_World->GetBlock(FloaterInfo.GetPos() - Vector3d(0, 0.1, 0));
+				if ((Block != E_BLOCK_AIR) && !IsBlockWater(Block))
+				{
+					a_Player->UseEquippedItem(2);
+				}
+			}
 		}
 		else
 		{


Bond-009:
I did, and it worked in all scenarios that I tested it in.

src/Items/ItemFishingRod.h
```diff
@@ -234,6 +234,10 @@ class cItemFishingRodHandler :
 				a_Player->UseEquippedItem(1);
 				cRoot::Get()->GetPluginManager()->CallHookPlayerFished(*a_Player, Drops);
 			}
+			else if (!IsBlockWater(a_World->GetBlock(FloaterInfo.GetPos())))
```
peterbell10: This doesn't seem right.  If the floater is in the air that's still not water but it shouldn't be taking damage.  I think `IsSolid` would be better, or whatever the floater's physics uses for collisions would probably be best.

Bond-009: Thanks, I didn't think about that.

peterbell10:
I tried it in game and it never damages the fishing rod.  I'm guessing @bearbin's right and the floater does indeed sit outside the block.

bearbin:
So the code should instead be checking for an air block, and the code for floater placement double-checked to confirm.

Bond-009:
Did you test it before before the changed to address your comment or after (or both)?

peterbell10:
@bearbin I don't think there's any way around checking a (small) volume like in #4136.

@Bond-009 I tested both: at first it was always damaged, now it is never damaged.

Bond-009:
@peterbell10 is this a good solution?

bearbin:
I don't believe this will handle hooks caught on vertical walls - will it?

PR #4160, opened 2018-01-20
Add new flow direction calculating algorithm

bionext03:


peterbell10:
Please don't open multiple pull requests for the same piece of code, it hides the discussion and review comments from #4156.  Really you should add your changes as new commits and it can be squashed into one when it gets merged.  However, if you do want to completely changing the commit history then you just need to push your changes with this:
```
git push --force-with-lease
```
The new history will then be shown on the same Github PR.

bionext03:
Ok. Sorry just a week I use git.

src/Simulator/FluidSimulator.h
```diff
@@ -46,7 +46,7 @@ class cFluidSimulator :
 	virtual bool IsAllowedBlock(BLOCKTYPE a_BlockType) override;
 
 	/** Gets the flowing direction. If a_Over is true also the block over the current block affects the direction (standard) */
-	virtual Direction GetFlowingDirection(int a_X, int a_Y, int a_Z, bool a_Over = true);
+	virtual Vector3f GetFlowingDirection(int a_X, int a_Y, int a_Z);
```
peterbell10: You need to update the doxy-comment as well.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,94 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
+	vDirection.z = LevelPoint[1] - LevelPoint[3];
 
-	if (a_X - X < 0)
-	{
-		return X_PLUS;
-	}
+	float Length = static_cast<float>(sqrt(vDirection.x * vDirection.x + vDirection.z * vDirection.z));
 
-	if (a_Z - Z > 0)
+	if (Length != 0.0f)
```
peterbell10:
```cpp
if (Direction.HasNonZeroLength())
{

File: src/Simulator/FluidSimulator.cpp [hunk 3]
```diff
@@ -128,100 +128,94 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
-	Points.push_back(new Vector3i(a_X + 1, a_Y, a_Z));
-	Points.push_back(new Vector3i(a_X, a_Y, a_Z + 1));
-	Points.push_back(new Vector3i(a_X, a_Y, a_Z - 1));
+	Points[0] = { a_X + 1, a_Y, a_Z };
+	Points[1] = { a_X, a_Y, a_Z + 1 };
+	Points[2] = { a_X - 1, a_Y, a_Z };
+	Points[3] = { a_X, a_Y, a_Z - 1 };
 
-	for (auto itr = Points.cbegin(), end = Points.cend(); itr != end; ++itr)
+	for (int i = 0; i < 4; i++)
```
peterbell10: You should use `ARRAYCOUNT(Points)` instead of manually specifying the length.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,94 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
 
 
 
-// TODO Not working very well yet :s
-Direction cFluidSimulator::GetFlowingDirection(int a_X, int a_Y, int a_Z, bool a_Over)
+Vector3f cFluidSimulator::GetFlowingDirection(int a_X, int a_Y, int a_Z)
 {
+	Vector3f vDirection;
+	Vector3i Points[4];
+	int LevelPoint[4];
```
peterbell10: Instead of declaring your variables at the top of the function, you should introduce them as they are needed and initialise them in the same statement.

src/Simulator/FluidSimulator.cpp
bionext03: This code is faster (only two operands, not three).

peterbell10: The root isn't calculated twice because you don't need the exact length for `HasZeroLength`.
As for the number of operands, you'd maybe save at best one or two CPU cycles. I don't think it's worth it.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,96 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
 
 
 
-// TODO Not working very well yet :s
-Direction cFluidSimulator::GetFlowingDirection(int a_X, int a_Y, int a_Z, bool a_Over)
+Vector3f cFluidSimulator::GetFlowingDirection(int a_X, int a_Y, int a_Z)
 {
+	Vector3f vDirection;
```
peterbell10: This variable isn't actually used until line 208. The other cases would be clearer as `return {};` for the zero vector.
Also remember to drop the prefix and just call this `Direction`.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,96 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
+		CentralPoint -= 8;
 	}
 
-	std::vector< Vector3i * > Points;
-
-	Points.reserve(4);  // Already allocate 4 places :D
+	int LevelPoint[4];
+	Vector3i Points[4];  // blocks around the checking pos
+	Points[0] = { a_X + 1, a_Y, a_Z };
+	Points[1] = { a_X, a_Y, a_Z + 1 };
```
peterbell10: I'd prefer this to be a single statement:

peterbell10: I'd prefer this to be a single statement:
```cpp
Vector3i Points[]
{
    { a_X + 1, a_Y, a_Z },
    // ...
};
```

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,95 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
 
-	if (a_Z - Z < 0)
+	if (Direction.HasNonZeroLength())
 	{
-		return Z_PLUS;
+		Direction.Normalize();
 	}
 
-	return NONE;
+	return {};
```
peterbell10: This should be returning `Direction`.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,95 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
-				Z = Pos->z;
+				BlockID = m_World.GetBlock(Points[i].x, Points[i].y - 1, Points[i].z);
+
+				if (BlockID == m_FluidBlock)
+				{
+					LevelPoint[i] = CentralPoint + 1;
+				}
+				else if (BlockID == m_StationaryFluidBlock)
+				{
+					LevelPoint[i] = CentralPoint;
```
peterbell10: This doesn't seem right. Don't stationary fluid blocks behaved exactly the same as the non-stationary versions?

bionext03: This eliminates the erratic flow in stationary water, try to run the game without this code and throw the object into a standing pond in which the bottom has a stepped depression.

peterbell10: I'm not sure we're on the same page here.  A stationary water block is any water that isn't going to be updated by the fluid sim.  Stationary water can be flowing in any direction.

Also, could you expand on what you mean by "stepped depression"?

bionext03: But it's work good.

peterbell10: Well, the vast majority of water is stationary so maybe the non-stationary fluid case is wrong.

peterbell10: The following loop seems to work perfectly for me:
```cpp
for (size_t i = 0; i < ARRAYCOUNT(Points); i++)
{
	BlockID = m_World.GetBlock(Points[i].x, Points[i].y, Points[i].z);

	if (IsAllowedBlock(BlockID))
	{
		LevelPoint[i] = 0x07 & m_World.GetBlockMeta(Points[i].x, Points[i].y, Points[i].z);
	}
	else
	{
		LevelPoint[i] = CentralPoint;
	}
}
```

bionext03: Ok, I will try this.

bionext03: This is the wrong way.

bionext03: Sorry, it works. But we still have to test.

bionext03: Not so good. Water simulator Vanilla. 

They must move.

bionext03: All the same, it works correctly.

bionext03:
I think that's all. The next thing to do is correct the handling of collisions.

src/Simulator/FluidSimulator.cpp
peterbell10: I can't seem to reproduce that, for me the diamonds are flowing as expected.

bionext03: It was my mistake, they also behave on the original server.

peterbell10: Oh, OK.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,64 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
-	NIBBLETYPE LowestPoint = m_World.GetBlockMeta(a_X, a_Y, a_Z);  // Current Block Meta so only lower points will be counted
-	int X = 0, Z = 0;  // Lowest Pos will be stored here
+	NIBBLETYPE CentralPoint = m_World.GetBlockMeta(a_X, a_Y, a_Z);
 
-	if (IsAllowedBlock(m_World.GetBlock(a_X, a_Y + 1, a_Z)) && a_Over)  // check for upper block to flow because this also affects the flowing direction
+	if (CentralPoint > 7)
 	{
-		return GetFlowingDirection(a_X, a_Y + 1, a_Z, false);
+		CentralPoint -= 8;
 	}
```
bearbin: I think this would be better represented as:

    NIBBLETYPE CentralPoint = (m_World.GetBlockMeta(a_X, a_Y, a_Z) & 0x7);

bionext03:
Ok,thank's.

src/Simulator/FluidSimulator.cpp
bionext03: I agree, changed.

bionext03:
Ok.

bearbin:
Hopefully the mac builds on travis will have finished by then :P

src/Entities/Entity.cpp
```diff
@@ -1032,40 +1032,18 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	}
 
 	// Get water direction
-	Direction WaterDir = m_World->GetWaterSimulator()->GetFlowingDirection(BlockX, BlockY, BlockZ);
+	Vector3f WaterDir = m_World->GetWaterSimulator()->GetFlowingDirection(BlockX, BlockY, BlockZ);
 
 	m_WaterSpeed *= 0.9;  // Reduce speed each tick
 
-	switch (WaterDir)
+	if (WaterDir.x != 0.0f)
```
madmaxoft: No need for this condition - if the `x` is zero, then adding it will have no effect anyway.
Just `Normalize` the `WaterDir` and then add it to `m_WaterSpeed` directly.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,58 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
-Direction cFluidSimulator::GetFlowingDirection(int a_X, int a_Y, int a_Z, bool a_Over)
+Vector3f cFluidSimulator::GetFlowingDirection(int a_X, int a_Y, int a_Z)
 {
 	if (!cChunkDef::IsValidHeight(a_Y))
 	{
-		return NONE;
+		return {};
 	}
+
 	BLOCKTYPE BlockID = m_World.GetBlock(a_X, a_Y, a_Z);
```
madmaxoft: Change this to `auto BlockType`

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,58 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
 		{
-			NIBBLETYPE Meta = m_World.GetBlockMeta(Pos->x, Pos->y, Pos->z);
-
-			if (Meta > LowestPoint)
-			{
-				LowestPoint = Meta;
-				X = Pos->x;
-				Z = Pos->z;
-			}
+			LevelPoint[i] = 0x07 & m_World.GetBlockMeta(Points[i].x, Points[i].y, Points[i].z);
```
madmaxoft: This calculation doesn't work properly with falling water (bit 0x08 set). That needs a special case.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,58 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
-	Points.reserve(4);  // Already allocate 4 places :D
-
-	// add blocks around the checking pos
-	Points.push_back(new Vector3i(a_X - 1, a_Y, a_Z));
-	Points.push_back(new Vector3i(a_X + 1, a_Y, a_Z));
-	Points.push_back(new Vector3i(a_X, a_Y, a_Z + 1));
-	Points.push_back(new Vector3i(a_X, a_Y, a_Z - 1));
-
-	for (auto itr = Points.cbegin(), end = Points.cend(); itr != end; ++itr)
+	for (unsigned long i = 0; i < ARRAYCOUNT(LevelPoint); i++)
```
madmaxoft: Use `size_t` instead.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,58 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
+	if (!IsAllowedBlock(BlockID))  // No Fluid -> No Flowing direction :D
 	{
-		return Y_MINUS;
+		return {};
 	}
-	*/
 
-	NIBBLETYPE LowestPoint = m_World.GetBlockMeta(a_X, a_Y, a_Z);  // Current Block Meta so only lower points will be counted
-	int X = 0, Z = 0;  // Lowest Pos will be stored here
+	NIBBLETYPE CentralPoint = m_World.GetBlockMeta(a_X, a_Y, a_Z) & 0x07;
```
madmaxoft: This calculation doesn't work properly with falling water (bit 0x08 set). That needs a special case, adding an `y` component to the direction.

src/Simulator/FluidSimulator.cpp
```diff
@@ -128,100 +128,58 @@ bool cFluidSimulator::IsHigherMeta(NIBBLETYPE a_Meta1, NIBBLETYPE a_Meta2)
...
-		return X_PLUS;
-	}
+	Direction.x = LevelPoint[0] - LevelPoint[2];
+	Direction.z = LevelPoint[1] - LevelPoint[3];
 
-	if (a_Z - Z > 0)
+	if (Direction.HasNonZeroLength())
 	{
-		return Z_MINUS;
+		Direction.Normalize();
```
madmaxoft: I think it's better not to normalize here, and let the clients normalize, if they want (so that those that don't get a perf bonus). Either way, this is worth mentioning in the doxy-comments of the function.

src/Entities/Entity.cpp
peterbell10: I think the condition is needed.  This way the water speed will carry momentum through sections of still water but if the condition weren't there it would be immediately set to 0.

madmaxoft: Not really. I propose changing the whole block of lines 1037 - 1047 (of the new code) with a single line:
```cpp
m_WaterSpeed = m_WaterSpeed * 0.9 + WaterDir.NormalizeCopy() * 0.4;
```
This is equivalent to the whole code block. Basically if `WaterDir.x == 0`, then adding it to `m_WaterSpeed` will have no effect anyway, so the `if` can go away, and the whole calculation can be `Vector3`-ized.

madmaxoft: At least use a proper float-comparison - you cannot compare a `float` value for equality, because you'll almost never reach it due to rounding errors. Use `if (std::abs(WaterDir.x) < 0.0005)`

bionext03:
You can modify the code yourself. This development is no longer interesting to me. Thank you for attention.

peterbell10:
Unfortunately, it seems that adding the downward velocity from falling water breaks the proper flowing behaviour so I've removed it for now. Some further testing also reveals that the water direction shouldn't be normalised at all, diagonally flowing water simply moves faster than straight flowing water does.

PR #4161, opened 2018-01-21
Deal with covered switches consistently

peterbell10:
Fixes a number of `<function>: not all control paths return a value`  warnings on MSVC.

Currently almost every switch that gets added seems to end up with several revisions to fix warnings on one or more of the compilers. So I've also introduced the `COVERED_SWITCH` global macro so these can be handled more consistently in future.

### DIFF
--- a/src/BlockArea.cpp
+++ b/src/BlockArea.cpp
@@ -2458,120 +2458,124 @@ void cBlockArea::MergeByStrategy(const cBlockArea & a_Src, int a_RelX, int a_Rel
 	int DstOffZ = std::max(0,  a_RelZ);  // Offset in Dst where to start writing
 	int SizeZ   = std::min(a_Src.GetSizeZ() - SrcOffZ, GetSizeZ() - DstOffZ);  // How many blocks to copy
 
-	switch (a_Strategy)
+	[&]
 	{
-		case cBlockArea::msOverwrite:
+		switch (a_Strategy)
 		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorOverwrite<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
-				m_Size.x, m_Size.y, m_Size.z
-			);
-			break;
-		}  // case msOverwrite
+			case cBlockArea::msOverwrite:
+			{
+				InternalMergeBlocks<MetasValid, MergeCombinatorOverwrite<MetasValid> >(
+					GetBlockTypes(), a_Src.GetBlockTypes(),
+					DstMetas, SrcMetas,
+					SizeX, SizeY, SizeZ,
+					SrcOffX, SrcOffY, SrcOffZ,
+					DstOffX, DstOffY, DstOffZ,
+					a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
+					m_Size.x, m_Size.y, m_Size.z
+				);
+				return;
+			}  // case msOverwrite
 
-		case cBlockArea::msFillAir:
-		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorFillAir<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
-				m_Size.x, m_Size.y, m_Size.z
-			);
-			break;
-		}  // case msFillAir
+			case cBlockArea::msFillAir:
+			{
+				InternalMergeBlocks<MetasValid, MergeCombinatorFillAir<MetasValid> >(
+					GetBlockTypes(), a_Src.GetBlockTypes(),
+					DstMetas, SrcMetas,
+					SizeX, SizeY, SizeZ,
+					SrcOffX, SrcOffY, SrcOffZ,
+					DstOffX, DstOffY, DstOffZ,
+					a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
+					m_Size.x, m_Size.y, m_Size.z
+				);
+				return;
+			}  // case msFillAir
 
-		case cBlockArea::msImprint:
-		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorImprint<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
-				m_Size.x, m_Size.y, m_Size.z
-			);
-			break;
-		}  // case msImprint
+			case cBlockArea::msImprint:
+			{
+				InternalMergeBlocks<MetasValid, MergeCombinatorImprint<MetasValid> >(
+					GetBlockTypes(), a_Src.GetBlockTypes(),
+					DstMetas, SrcMetas,
+					SizeX, SizeY, SizeZ,
+					SrcOffX, SrcOffY, SrcOffZ,
+					DstOffX, DstOffY, DstOffZ,
+					a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
+					m_Size.x, m_Size.y, m_Size.z
+				);
+				return;
+			}  // case msImprint
 
-		case cBlockArea::msLake:
-		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorLake<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
-				m_Size.x, m_Size.y, m_Size.z
-			);
-			break;
-		}  // case msLake
+			case cBlockArea::msLake:
+			{
+				InternalMergeBlocks<MetasValid, MergeCombinatorLake<MetasValid> >(
+					GetBlockTypes(), a_Src.GetBlockTypes(),
+					DstMetas, SrcMetas,
+					SizeX, SizeY, SizeZ,
+					SrcOffX, SrcOffY, SrcOffZ,
+					DstOffX, DstOffY, DstOffZ,
+					a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
+					m_Size.x, m_Size.y, m_Size.z
+				);
+				return;
+			}  // case msLake
 
-		case cBlockArea::msSpongePrint:
-		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorSpongePrint<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
-				m_Size.x, m_Size.y, m_Size.z
-			);
-			break;
-		}  // case msSpongePrint
+			case cBlockArea::msSpongePrint:
+			{
+				InternalMergeBlocks<MetasValid, MergeCombinatorSpongePrint<MetasValid> >(
+					GetBlockTypes(), a_Src.GetBlockTypes(),
+					DstMetas, SrcMetas,
+					SizeX, SizeY, SizeZ,
+					SrcOffX, SrcOffY, SrcOffZ,
+					DstOffX, DstOffY, DstOffZ,
+					a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
+					m_Size.x, m_Size.y, m_Size.z
+				);
+				return;
+			}  // case msSpongePrint
 
-		case cBlockArea::msDifference:
-		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorDifference<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
-				m_Size.x, m_Size.y, m_Size.z
-			);
-			break;
-		}  // case msDifference
+			case cBlockArea::msDifference:
+			{
+				InternalMergeBlocks<MetasValid, MergeCombinatorDifference<MetasValid> >(
+					GetBlockTypes(), a_Src.GetBlockTypes(),
+					DstMetas, SrcMetas,
+					SizeX, SizeY, SizeZ,
+					SrcOffX, SrcOffY, SrcOffZ,
+					DstOffX, DstOffY, DstOffZ,
+					a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
+					m_Size.x, m_Size.y, m_Size.z
+				);
+				return;
+			}  // case msDifference
 
-		case cBlockArea::msSimpleCompare:
-		{
-			InternalMergeBlocks<MetasValid, MergeCombinatorSimpleCompare<MetasValid> >(
-				GetBlockTypes(), a_Src.GetBlockTypes(),
-				DstMetas, SrcMetas,
-				SizeX, SizeY, SizeZ,
-				SrcOffX, SrcOffY, SrcOffZ,
-				DstOffX, DstOffY, DstOffZ,
-				a_Src.GetSizeX(), a_Src.GetSizeY(), a_Src.GetSizeZ(),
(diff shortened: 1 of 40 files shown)


src/WorldStorage/FastNBT.cpp
```diff
@@ -30,85 +30,78 @@ static const int MAX_LIST_ITEMS = 10000;
...
+
 
 
 ////////////////////////////////////////////////////////////////////////////////
 // cNBTParseErrorCategory:
 
 AString cNBTParseErrorCategory::message(int a_Condition) const
 {
-	switch (static_cast<eNBTParseError>(a_Condition))
+	switch (a_Condition)
```
bibo38: Why? Your macro should be the fix for the clang warning (as you've done it in the other 90%). Not some dirty `AS_INT` trick (without even explaining, why it is necessary).
Then I also don't need the `AS_INT` macro.

tests/HTTP/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ set (HTTP_SRCS
 	${CMAKE_SOURCE_DIR}/src/HTTP/TransferEncodingParser.cpp
 	${CMAKE_SOURCE_DIR}/src/HTTP/UrlClient.cpp
 	${CMAKE_SOURCE_DIR}/src/HTTP/UrlParser.cpp
+	${CMAKE_SOURCE_DIR}/src/OSSupport/StackTrace.cpp
```
bibo38: I don't get, why this is necessary for this PR.

src/BlockArea.cpp
```diff
@@ -2572,14 +2572,7 @@ void cBlockArea::MergeByStrategy(const cBlockArea & a_Src, int a_RelX, int a_Rel
 			break;
 		}  // case msMask
 
-		#ifndef __clang__  // Clang complains about a default case in a switch with all cases covered
-		default:
-		{
-			LOGWARNING("Unknown block area merge strategy: %d", a_Strategy);
-			ASSERT(!"Unknown block area merge strategy");
-			return;
-		}
```
bibo38: I'm not that familar, what the `UNREACHABLE` macro does.
But I would like to specify the warning for each `COVERED_SWITCH`. Then users running a release build can also report a warning like: "Unknown block area merge strategy: 77", which helps more than no warning (changed behaviour) or the generic warning "Covered switch should never hit default".

Maybe a `COVERED_DEFAULT` macro would be more flexible. So that I can write something like this:

    COVERED_DEFAULT
    {
        LOGWARNING("We broke X")
        ASSERT(!"Should not happen")
        return KeepItWorkingValue;
    }

src/WorldStorage/FastNBT.cpp
peterbell10: Because the switch isn't covered.  The input is an int and therefore could take any value.

I rewrote it to use `int` in the switch to avoid the default being optimised away as comments elsewhere suggest.  `AS_INT` is simply less visual noise than `static_cast<int>()`.

tests/HTTP/CMakeLists.txt
peterbell10: Because `UNREACHABLE` uses `PrintStackTrace()` and so needs to link to it.

src/BlockArea.cpp
bearbin: If you look down at the bottom of these changes, UNREACHABLE is defined, and it does log an informative message.

src/WorldStorage/FastNBT.cpp
```diff
@@ -30,85 +30,78 @@ static const int MAX_LIST_ITEMS = 10000;
 	#define PROPAGATE_ERROR(X) do { auto Err = (X); if (Err != eNBTParseError::npSuccess) return Err; } while (false)
 #endif
 
+#define AS_INT(x) static_cast<int>(x)
+
```
bearbin: Is this really necessary? To me this just adds confusion to what the code does later.

src/BlockArea.cpp
bibo38: Oh, yeah. Then this is fine.

peterbell10: >Maybe a `COVERED_DEFAULT` macro would be more flexible. So that I can write something like this:

A big part of this PR is trying to minimise the difference in behaviour between different platforms. As it stands many places are just a warning message on some platforms but undefined behaviour on others.  That seems very bad to me.

bearbin:
This seems sensible. I'm slightly confused by the change regarding the eNBTParseError switch, but otherwise the idea is good.

src/WorldStorage/FastNBT.cpp
peterbell10: I just though that it was too visually noisy with the full `static_cast` for every label:

```cpp
case static_cast<int>(eNBTParseError::npNoTopLevelCompound):
```

src/Globals.h
```diff
@@ -352,6 +352,15 @@ template class SizeChecker<UInt8,  1>;
 	#define assert_test(x) ( !!(x) || (assert(!#x), exit(1), 0))
 #endif
 
+#define UNREACHABLE(x) do { FLOGERROR("Hit unreachable code: {0}, file {1}, line {2}", #x, __FILE__, __LINE__); PrintStackTrace(); std::terminate(); } while (false)
```
bibo38: The current implementation uses `ASSERT` to only kill debug executables, but lets the production executable alive. So why you just call `std::terminate`, which also kills the production executable?

peterbell10: Under clang builds it's undefined behaviour so I don't want anyone relying on it working in release mode.

src/BlockArea.cpp
bibo38: I've just thought, that you could use the `#pragma` things on the macro to suppress Clang from giving the warnings, when using the macro... But sadly, it doesn't work.

src/Globals.h
bibo38: Sorry, but I don't understand the logic. It was always UB in clang. But on GCC/MSVC the last code had a backup solution to keep the application running.

bibo38: Sorry, but I don't understand the logic. It was always UB in clang. But on GCC/MSVC the last code had a backup solution to keep the application running.

bibo38:
> As it stands many places are just a warning message on some platforms but undefined behaviour on others. That seems very bad to me.

Sorry, but I can't agree with this intention for this PR. It changes the behaviour from backup solution(other) + undefined behaviour(Clang) to crash(other) + undefined behaviour(Clang)
This change doesn't make the behaviour better (or unify it). Because users expect a release not to crash, but this PR just removes the backup code, which previously would maybe give the user a chance to save his world.

I would suggest the following solutions:
- Just use the macro to minimize the `#ifdef` usage (but keep the backup-solution!)
- Throw all covered defaults away (and maybe disable the warnings from GCC/MSVC). Now our Clang builds make sure, that we have a covered switch. And if this should fail, we have a compiler consistent UB.
- Globally disable the `-Wcovered-switch-default` warning for Clang, so we **never have UB** and always have a backup solution.

src/WorldStorage/FastNBT.cpp
bibo38: It would be better to change `eNBTParseError` from `enum class` to a simple `enum` with a namespace around it. Then put a comment above, which explains, that `std::error_category` requires this to be an integer and therefore we don't need strict type-checking.
Now you can remove all the `static_cast` from the `eNBTParseError` enum.
As a bonus, you could add a typedef: `using eNBTParseError_t = int`

peterbell10:
>Throw all covered defaults away (and maybe disable the warnings from GCC/MSVC).

__That's what this PR is__.
`COVERED_SWITCH` is a way to disable the warning and at the same time inject useful behaviour, a crash message.  If you like, you can think of it as UB that happens to result in an error message and a call to `std::terminate`.

src/WorldStorage/FastNBT.cpp
peterbell10: I'm not keen on the idea of reducing the type safety globally just to save some typing within a single function.

>`using eNBTParseError_t = int`

This wouldn't allow assigning to a `std::error_code`: I can't overload `std::is_error_code_enum` for `int`...

bibo38:
> COVERED_SWITCH is a way to disable the warning and at the same time inject useful behaviour, a crash message.

But the consistency is missing! Think of the following (bad) code:

```cpp
enum class Color
{
        RED
};

int main()
{
        int i = 77;
        switch(static_cast<Color>(i))
        {
                case Color::RED: return 1;
                COVERED_SWITCH;
        }
}
```
Based on your `COVERED_SWITCH` implementation, this program runs fine with Clang (with an return value of 0, on my machine), but will produce a crash on MSVC/GCC. And now I get a bug report, which 10% can't reproduce, because they use Clang.

So either just remove the `COVERED_SWITCH` statement from the above code and suppress the MSVC/GCC warning. Now we have UB everywhere. But when we believe, that the Clang static analyzer will check the coverage, we are good to go.


But my preferred alternative is to **always have defined behaviour**! So why not do this:
```cpp
enum class Color
{
        RED
};

#define UNREACHABLE_DEFAULT(value) default: WARN("We have a problem at line XYZ"); return value
int main()
{
        int i = 77;
        switch(static_cast<Color>(i))
        {
                case Color::RED: return 1;
                UNREACHABLE_DEFAULT(0);
        }
}
```
Now suppress the Clang warning, and the program will always return 0 (on any compiler) and warn me about the missing statement. Maybe let it crash in debug mode, but try to not crash it in release! Repairing a broken world is much more frustrating, then having a client disconnect, because of giving the client a `dtUnknown_7` damage type.

src/WorldStorage/FastNBT.cpp
bibo38: > error_condition( int val, const error_category& cat );
> template< class ErrorConditionEnum > error_condition( ErrorConditionEnum e );

Sorry, I now see, that `std::error_condition` is the problem. Allowing things like `std::error_condition(77, std::iostream_category())` just kills the type safety.

The `value()` function just shows, that this was designed for C-Enums:
```cpp
if( errorCode.category() == boost::asio::error::misc_category )
    {
        switch (errorCode.value())
        {
        case boost::asio::error::eof:
            debugMsg << ". Server has disconnected.";
            break;
        case boost::asio::error::connection_refused:
            debugMsg << ". Connection Refused";
            break;
        default:
            debugMsg << ". Unknown Error.";
            break;
        }
    }
```
And then somebody came ahead and just glued the Static typed enums at this class with `is_error_condition_enum` and `template< class ErrorConditionEnum > error_condition( ErrorConditionEnum e );`. And this lead us to subtile errors like the above.

I'm ok with your way. But I would suggesst using a template instead of a macro (so I can do `as_int(eNBTParseError::npNeedBytes)`, but not `as_int(cEntityEffect::effWither)`). This also removes the `#undef`.

peterbell10:
>this program runs fine with Clang (with an return value of 0, on my machine), but will produce a crash on MSVC/GCC

It's simple, the behaviour of a covered switch that reaches the default is unspecified.  As such you shouldn't do it. Maybe a comment as such in globals.h would be useful.

However, if it really is UB then you can't say _anything_ about clang's output.  Nasal demons and all that.

bibo38:
> It's simple, the behaviour of a covered switch that reaches the default is unspecified.

You mean, that it is currently unspecified for Clang builds? Because the [C++ reference](https://en.cppreference.com/w/cpp/language/switch) says: `If condition evaluates to the value that doesn't match any of the case: labels, and the default: label is present, control is transferred to the statement labeled with the default: label.`

>  As such you shouldn't do it.

Yep, I shouldn't make bugs. But I'm only a human. The question is: Do you want to catch these bugs yourself or rely on Clang/Reviews to catch the bugs.
But the current is not acceptable: "If you run MSVC/GCC, I catch the bug myself and print an error. But if you use Clang, then I don't want to catching the bug."

> However, if it really is UB then you can't say anything about clang's output. Nasal demons and all that.

It [is UB](https://en.cppreference.com/w/cpp/language/return): `Flowing off the end of a value-returning function (except main) without a return statement is undefined behavior. `

But as I said, **on my machine** it returned 0. The main problem is, that UB could mean an execution without any error.

peterbell10:
>It is UB: Flowing off the end of a value-returning function (except main)

Except main being key there.  Falling off the end of main isn't undefined, it will always return 0.  I was referring to this:
>Values of integer, floating-point, and other enumeration types can be converted by static_cast or explicit cast, to any enumeration type. If the underlying type is not fixed and the source type is integer type or enumeration, the result is `unspecified (until C++17)` `undefined behavior (since C++17)` if the value, converted to the enumeration's underlying type, is out of range

Now the key part of your example:
```cpp
static_cast<Color>(77)
```
In C++17 (which we need to support because MSVC doesn't have strict standard settings) the result is  undefined behaviour.  Beyond here there is no point trying to reason about it any further.

peterbell10:
Okay, I've tried a different approach with `eNBTParseError`.  I put the error category into the .cpp file so users have to rely on `make_error_code` which guarantees it's a valid enum.  In which case the switch is covered.

bibo38:
> Okay, I've tried a different approach with eNBTParseError. I put the error category into the .cpp file so users have to rely on make_error_code which guarantees it's a valid enum. In which case the switch is covered.

Looks good, but be aware, that it isn't bulletproof:
```cpp
std::error_code(77, make_error_code(eNBTParseError::npSuccess).category())
```
(but who would do that 😼 )

>  Falling off the end of main isn't undefined

Wow, that was a bad example... And I didn't realize the missing UB when writing the argument 🤣 .

> In C++17 (which we need to support because MSVC doesn't have strict standard settings) the result is undefined behaviour.

Ok, I understand. But why not just disable the MSVC/GCC warnings instead of `COVERED_SWITCH`? Because these warnings are simply false positives on scoped enums.
And furthermore GCC can also detect an uncovered switch (`-Wswitch`). But when using the `COVERED_SWITCH` macro incorrectly, it will also prevent this warning from showing up (except from Clang 😒 ).

peterbell10:
>why not just disable the MSVC/GCC warnings

Because they're very general warnings like "control reaches the end of non-void function".  The only way to disable them locally is with a huge mess of `#ifdef` and `#pragma's for every switch.

>it will also prevent this warning from showing up (except from Clang 😒 ).

Well everything is going to be compiled by clang as part of CI anyway so I don't see the issue.

bibo38:
> Because they're very general warnings like "control reaches the end of non-void function".

But at least in my case, I almost only encounter them in switch-situations. And in most of the time, it is already covered... Now I need an extra macro to "fix" this wrong warnining...

> Well everything is going to be compiled by clang as part of CI anyway so I don't see the issue.

Yeah, so if somebody misses a return (with the disabeled warning), it will show up in the CI (without any wrong warning).
On the other hand, if I developed with Clang I might need to add an superfluous `COVERED_SWITCH` to my switch-case to "fix" the wrong GCC warning. For me this is just unnecessary.

>  The only way to disable them locally is with a huge mess of #ifdef and #pragma's for every switch.

I mean to disable it globally!
Due to the fact, that Clang already has a lot more warnings than GCC, I don't see the problem disabeling a "general" warning. For me it is a similar reason, like the reason `-Wunreachable-code` [was removed from GCC](https://gcc.gnu.org/ml/gcc-help/2011-05/msg00360.html): You cannot rely on the warning!

P.S.: Why did you remove the const in your last commit?

peterbell10:
>I don't see the problem disabeling a "general" warning.

Clang and GCC both compile for the same OS and standard library so I expect there to be very little code that's compiled on one but not the other (mostly covered switch defaults as it happens).  On the other hand windows does have much more conditionally compiled code, mainly in OSSupport, Root.cpp and main.cpp.  We can't rely on Clang to report error's in windows specific code, so we shouldn't turn MSVC's warning level down.  None of the switches are conditionally compiled so they're safe though.

>Why did you remove the const in your last commit?

It was (somehow) causing this error on clang 3.5:
default initialization of an object of const type [...] requires a user-provided default constructor.

Maybe a bug with explicitly defaulted default-constructors?

bibo38:
> It was (somehow) causing this error on clang 3.5:
default initialization of an object of const type [...] requires a user-provided default constructor.

This is actually a requirement from the [C++ standard](http://en.cppreference.com/w/cpp/language/default_initialization). Under notes we have:

> If T is a const-qualified type, it must be a class type with a user-provided default constructor. 

Explanation: https://stackoverflow.com/questions/33809441/default-initialization-of-const-qualified-type-with-no-user-provided-constructor#33809698

A workaround mentioned on http://open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#253.
TLDR: Use `= {}` to initialise the const object.

> windows does have much more conditionally compiled code

Ok, your point. But I've come up with an alternative:

```cpp
int convert(MyEnum x)
{
    switch(x)
    {
        case A: return 1;
        case B: return 2;
    }

    UNREACHABLE;
}
```
Now just make the `UNREACHABLE` macro to do nothing on Clang (or is there any case, where we would need the macro in Clang?). This solves both problems:

- No warnings are disabled
- GCC can emit a warning, if the switch is not covered (`-Wswitch`).
- (We save the `COVERED_SWITCH` macro)

P.S.: Maybe it might be good to just define empty macros for Clang like `do {} while(false)`. This would allow Clang to detect a missing semicolon after the macro.

peterbell10:
> This is actually a requirement from the C++ standard.

I guess I'm not up on my standardese, an explicitly defaulted member counts as "user-declared" but evidently not "user-provided".

>I've come up with an alternative

Doesn't work for non-returning switches but it remains to be seen how common those are.  I'll give it a go and see how it works out.

peterbell10:
Okay, it requires a bit of reworking to get all of the covered switches returning from every case.  Particularly the mineshaft generator change is a bit odd.

On the plus side though, clang seems to be happy with the  `UNREACHABLE` outside the switch.

src/Generating/MineShafts.cpp
```diff
@@ -767,27 +767,34 @@ void cMineShaftCorridor::PlaceChest(cChunkDesc & a_ChunkDesc)
...
-			break;
-		}
+			switch (m_Direction)
+			{
+				case dirXM:
+				case dirXP:
+				{
+					return
+					{
+						m_BoundingBox.p1.x + m_ChestPosition - BlockX,
```
bibo38: You could also modify the variables directly from the lambda, like it was in the old code (just capture them by reference). This could look a bit better than the tuples.

src/Entities/Player.cpp
```diff
@@ -1060,29 +1060,31 @@ void cPlayer::KilledBy(TakeDamageInfo & a_TDI)
...
-			case dtAdmin: DamageText = "was administrator'd"; break;
-			case dtExplosion: DamageText = "blew up"; break;
-			case dtAttack: DamageText = "was attacked by thin air"; break;
-			COVERED_SWITCH;
-		}
+		const AString DamageText = [&]
+			{
+				switch (a_TDI.DamageType)
+				{
+					case dtRangedAttack:    return "was shot";
```
bibo38: This looks like a place to extract a function like `GetDeathMessage(DamageType)`. Just makes this function a bit shorter.

src/BlockArea.cpp
```diff
@@ -2571,8 +2571,6 @@ void cBlockArea::MergeByStrategy(const cBlockArea & a_Src, int a_RelX, int a_Rel
 			);
 			break;
 		}  // case msMask
-
-		COVERED_SWITCH;
 	}  // switch (a_Strategy)
```
bibo38: You've deleted a `COVERED_SWITCH` here without a `UNREACHABLE`. Is this intended?

PR #4167, opened 2018-01-22
Update m_IsSolid assignments in BlockInfo

### LINKED ISSUE #4166: Stairs are solid.
Is it a bug or so conceived? I think it's because the stairs are solid.

bearbin:
Fixes #4166


bearbin:
This should not be merged at present as I am going to fix the issues surrounding the placement of ladders in this PR too.

bearbin:
This isn't a workaround for bounding boxes, it's just an attempt to make the non-solid blocks match up with vanilla. I used the [wiki](https://minecraft.gamepedia.com/Solid_block), although I don't believe the list is fully correct (full snow blocks are non-solid??).

peterbell10:
Wow, these categories really make no sense to me.  The wiki says:

>A block can be considered solid if it has a collision box that players, mobs, or other entities cannot move through

Yet a ladder, which does have a collision box that stops entities isn't considered solid...

peterbell10:
>This should not be merged at present as I am going to fix the issues surrounding the placement of ladders in this PR too.

Is it safe to say this no longer applies @bearbin?

PR #4168, opened 2018-01-22
Improving the collision physics algorithm.

bionext03:


bionext03:
Since all entity are represented as cylinders, we need the class cCylinderBlockTracer. I decided to develop an algorithm based on the theorem on separating axes. Close this request or leave it open and show the results here?

peterbell10:
It's from the discussion here so I'd say keep it in this PR.

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 
-			// Avoid movement in the direction of the blockface that has been hit:
-			switch (HitBlockFace)
+		for (unsigned long i = 0; i < ARRAYCOUNT(SizeLimit); i++)
+		{
+			switch (i)
 			{
-				case BLOCK_FACE_XM:
-				case BLOCK_FACE_XP:
+				case 0:
```
bibo38: This just looks wrong. It is like putting a sequential program into a loop:

```cpp
for(int i = 0; i < 3; i++)
{
    switch(i)
    {
        case 0: FirstInstruction; break;
        case 1: SecondInstruction; break;
        case 2: ThirdInstruction; break;
    }
}
```
instead of
```cpp
FirstInstruction;
SecondInstruction;
ThirdInstruction;
```
And if somebody changes the loop counter, we may have a hard day...

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+					{
+						SizeLimit[i] *= NextSpeed.NormalizeCopy();
+						if (NextSpeed.y < 0.0)
+						{
+							SizeLimit[i].y *= 0;
+						}
+					}
+				case 1:
```
bibo38: Is this missing `break` intended?

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			switch (i)
 			{
-				case BLOCK_FACE_XM:
-				case BLOCK_FACE_XP:
+				case 0:
+					{
+						SizeLimit[i] *= NextSpeed.NormalizeCopy();
+						if (NextSpeed.y < 0.0)
+						{
+							SizeLimit[i].y *= 0;
```
bibo38: Why multiply with `0`? Why not just set it to `0`?

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+				case 2:
 				{
-					NextSpeed.y = 0;
+					if (NextSpeed.y == 0.0)
+					{
+						continue;
+					}
+					else if (NextSpeed.y < 0.0)
+					{
+						SizeLimit[i].y = 0;
```
bibo38: Soo, in the X and Z case, you multiply `SizeLimit` with `-1`. But if Y is negative, `SizeLimit` becomes `0` 😕 ?

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+					case 0:
+					{
+						if ((HitBlockFace == BLOCK_FACE_XM) || (HitBlockFace == BLOCK_FACE_XP))
+						{
+							NextSpeed.x = 0;
+							NextPos.x = HitCoords.x - SizeLimit[i].x;
+							break;
+						}
+					}
+					case 1:
```
bibo38: Is this also the special case of a fallthrough, if we didn't hit a X-Blockface?

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
-			{
-				// We hit the ground, adjust the position to the top of the block:
-				m_bOnGround = true;
-				NextPos.y = HitBlockCoords.y + 1;
-			}
+			{ m_Width, m_Height, m_Width },
+			{ m_Width, 0, 0 },
+			{ 0, m_Height, 0 },
+			{ 0, 0, m_Width },
+		};
```
bibo38: This just reminds me of a matrix... Maybe create a `cMatrix` class, if you need to do some matrix operations?

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 
-			// Avoid movement in the direction of the blockface that has been hit:
-			switch (HitBlockFace)
+		for (unsigned long i = 0; i < ARRAYCOUNT(SizeLimit); i++)
```
bibo38: Since C++11, we now have `std::array`, which removes the need for the `ARRAYCOUNT` macro.

src/Entities/Entity.cpp
```diff
@@ -1080,57 +1080,126 @@ void cEntity::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
-	if (NextSpeed.SqrLength() > 0.0f)
-	{
-		Vector3d HitCoords;
-		Vector3i HitBlockCoords;
-		eBlockFace HitBlockFace;
-		Vector3d wantNextPos = NextPos + NextSpeed * DtSec.count();
-		auto isHit = cLineBlockTracer::FirstSolidHitTrace(*GetWorld(), NextPos, wantNextPos, HitCoords, HitBlockCoords, HitBlockFace);
-		if (isHit)
+	if (NextSpeed.HasNonZeroLength())
+	{
```
bibo38: I would prefer to create a seperate function for this 200 lines. Otherwise it is hard to see, where the if-statement ends.

PR #4169, opened 2018-01-22
Fix android build

### LINKED ISSUE #4165: Android builds down
Although https://cuberite.org/status reports android builds are working, individual compilation fails. The script however does not exit on a compile failure, so Jenkins continues building and doesn't report an error.

Is this related to #4163?

peterbell10:
Should fix #4165


peterbell10:
Question: Is it possible to have Jenkins build a pull request on request before merging?  It would be nice to sort these kinds of issues before it gets to master.

bearbin:
It is possible to set up jenkins as a CI on GitHub, in fact that's what we did before travis and appveyor. I can't remember why we switched to them in the first place, there must have been a good reason. I can set up a simple builder quite easily if you wanted, I would just need a suitable build script.

PR #4171, opened 2018-01-23
Ocelots don't take fall damage

### LINKED ISSUE #4157: ocelot's getting damage when falling
Client version:   1.12.2
Server OS:        Windows 7
Commit id:        https://github.com/cuberite/cuberite/commit/ec9e0eecf616bb190e4d0be76f1c120cac9eb60b

### Expected behavior
ocelot doesn't get damage when it falls

### Actual behavior
ocelot dies or gets some damage

1. place some blocks
2. spawn ocelot
3. make him fall

sorry for my bad English)

Bond-009:
Fixes #4157

### DIFF
--- a/src/Mobs/Ocelot.cpp
+++ b/src/Mobs/Ocelot.cpp
@@ -218,3 +218,13 @@ bool cOcelot::IsCatSittingOnBlock(cWorld * a_World, Vector3d a_BlockPosition)
 
 
 
+
+bool cOcelot::DoTakeDamage(TakeDamageInfo & a_TDI)
+{
+	if (a_TDI.DamageType == dtFalling)
+	{
+		return false;
+	}
+
+	return super::DoTakeDamage(a_TDI);
+}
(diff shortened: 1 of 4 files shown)


src/Mobs/Ocelot.cpp
```diff
@@ -218,3 +218,8 @@ bool cOcelot::IsCatSittingOnBlock(cWorld * a_World, Vector3d a_BlockPosition)
 
 
 
+
+void cOcelot::HandleFalling(void)
+{
+	// empty - Ocelots don't take fall damage
```
peterbell10: This function did a lot more than just do damage.  A better way to do damage immunity this is to override `DoTakeDamage` and not call `super::DoTakeDamage` for `dtFalling`.

bearbin: A similar method was used in #4136, so you may want to look there for any guidance.

Bond-009: Should I change the fall damage logic for chickens too? (It's done the same way as this PR)

peterbell10: I think that would be best.

Bond-009: Done

PR #4173, opened 2018-01-25
Stop cFunctionRef constructor from disabling default copy constructor.

peterbell10:
This fixes an issue I was made aware through [this reddit comment](https://www.reddit.com/r/cpp/comments/7svbj7/is_stdfunction_really_the_best_we_can_do_lukas/dt86azb/). There's a flaw in my original `cFunctionRef` implementation: when constructing a `cFunctionRef` from another `cFunctionRef`,  instead of just copying the pointers it actually wraps it in another layer of indirection.  For example a `ForEach` function in `cChunk` that was called via `cWorld` will have to follow three function pointers for each call of the callback instead of just one.


bibo38:
Just being curious: Is `std::decay` only for removing the cv-qualifiers + references or is it also needed for the `std::add_pointer<F>::type` functionality?

And would it be equivalent to make template specialisations, instead of the `enable_if`?
E.g.:

```cpp
template<> cFunctionRef(const cFunctionRef &) = default;
template<> cFunctionRef(cFunctionRef &&) = default; // With the exception of MSVC ;D
```

peterbell10:
> Is `std::decay` only for removing the cv-qualifiers + references

Yes, the other situations wouldn't make any sense for a `cFunctionRef`.  It isn't a function to convert to a function pointer and an array of them wouldn't be a callable.

>And would it be equivalent to make template specialisations, instead of the enable_if?

You can't specialise a function template but you could overload it, in which case you would need to provide all possible combinations of cv-ref, most of which can't be defaulted:
```cpp
cFunctionRef(cFunctionRef &);
cFunctionRef(const cFunctionRef &);
cFunctionRef(volatile cFunctionRef &);
cFunctionRef(const volatile cFunctionRef &);
cFunctionRef(cFunctionRef &&);
cFunctionRef(const cFunctionRef &&);
cFunctionRef(volatile cFunctionRef &&);
cFunctionRef(const volatile cFunctionRef &&);
```
If you don't provide all of them then the forwarding reference would be a better match.

bibo38:
> You can't specialise a function template but you could overload it

Interesting... I always thought, that you would need full specialisation to prevent ambigue overloads... 📚 

> most of which can't be defaulted

Just curious, if these could be handled by the template constructor. Because `cFunctionRef x(const_cast<const cFunctionRef &&>(std::move(otherFunction)))` could create some funny results.

peterbell10:
To be clear, you can still partially specialise, but not fully specialise.
>I always thought, that you would need full specialisation to prevent ambigue overloads

If you really want to know why this isn't the case I would point you to [cppreference](http://en.cppreference.com/w/cpp/language/overload_resolution#Best_viable_function).
The gist of which is that non-template overloads are preferred where otherwise there is ambiguity.

>Just curious, if these could be handled by the template constructor.

Of course. If you pass a `const cFunctionRef &&` the `enable_if` would disable the forwarding constructor and so the overload set is:
```
cFunctionRef(const cFunctionRef&);
cFunctionRef(cFunction &&);
```

The `const` can't be removed and so the copy-constructor is chosen.

tigerw:
merge?

madmaxoft:
To be honest, I have no idea what this does. So don't expect a review from me.

peterbell10:
If you're not convinced, as an appeal to authority, here's two standard library maintainers describing the problem and presenting this solution:
https://youtu.be/wYd2V4nPn0E?t=11m54s

SFINAE isn't pretty but there isn't really any other way to fix it.

madmaxoft:
I do know what sfinae is and can use it in simple stuff, but I don't know what `cFunctionRef` exactly does for us, so that's why I don't understand the code.

peterbell10:
`cFunctionRef` is used whenever you call one of the `cWorld::DoWith` functions, e.g. 
```cpp
m_World->DoWithChunkAt(BlockPos, [](cChunk & a_Chunk) -> bool
    {
        // ...
    }
);
```

It's looks like you're calling it with a lambda but that would require `DoWithChunkAt` to be a template. The function is really being called with `cFunctionRef<bool(cChunk&)>` which is constructed from the lambda via a templated constructor.  This gives you a generic pointer to the lambda which works great.

The problem comes in when `cWorld` forwards that down to the chunk map
```cpp
return m_ChunkMap->DoWithChunkAt(a_BlockPos, a_Callback);
```
The chunk map version has the exact same signature so you would want it to just call the copy constructor for `cFunctionRef`.  However, what actually happens is the `cFunctionRef` constructor sees it's being passed an object that's callable with the right signature and  the templated constructor happens to be a better match than the copy constructor.  So in effect you have to follow a two layers of indirection to get to the original lambda, instead of just copying the pointer itself.  

It still works as it is but each layer of `cFunctionRef` that the lambda is wrapped in incurs an extra indirect function call and copying of arguments passed to the function.

madmaxoft:
Great explanation, thanks. Now, why isn't this in the `cFunctionRef`'s doxy-comments? This has explained the matter so much better than `light-weight, type-erased reference to a function object` :)

peterbell10:
That's probably a good idea, should I add it in this PR?

madmaxoft:
I leave that up to you. Although not exactly the focus of this PR, it is highly relevant, so I see no problem doing it here.

Doxyfile
```diff
@@ -194,7 +194,7 @@ SEPARATE_MEMBER_PAGES  = NO
 # The TAB_SIZE tag can be used to set the number of spaces in a tab.
 # Doxygen uses this value to replace tabs by spaces in code fragments.
 
-TAB_SIZE               = 2
+TAB_SIZE               = 4
```
peterbell10: I had to change this to get doxygen to format the code blocks properly.
https://www.stack.nl/~dimitri/doxygen/manual/markdown.html#mddox_code_blocks

PR #4178, opened 2018-02-02
Make offline UUIDs consistent with vanilla.

### LINKED ISSUE #4177: Using same UUID (compat to spigot/bukkit etc)
Client version:   1.12.2
Server OS:        Windows
Commit id:        95108c134d410040efadfa6bb00c8ce3486fb7fa built at: 629

### Why?
It can help to import existent worlds with entities owned by a player.

### Expected behavior
If the UUID is 5 in spigot/bukkit/paperclip(/vanlilla?)
Than it is the same in cuberite

### Actual behavior
Another UUID is generated -> another player

Start cuberite and another MC server (spigot/bukkit/paperclip)

### Alternative solution
Map a current Cuberite UUID to any other.

peterbell10:
Fixes #4177

Bear in mind that this is a breaking change to existing Cuberite worlds and permissions settings.


bearbin:
This feels like a good idea to me, but it is a breaking change and I can't think of a really compelling reason to merge.

peterbell10:
I have to say I don't think this should be merged as-is either.  However, having player files work properly across different servers does seem like a big win.

I wonder if it's possible to perform some migration.  I know for example player data files have contain the username so in theory that can be preserved.

src/Entities/Player.cpp
```diff
@@ -2063,8 +2082,26 @@ bool cPlayer::LoadFromDisk(cWorldPtr & a_World)
...
-	// Load from the offline UUID file, if allowed:
+	// Check for old offline UUID filename, if it exists migrate to new filename
 	cUUID OfflineUUID = cClientHandle::GenerateOfflineUUID(GetName());
+	auto OldFilename = GetUUIDFileName(GetOldStyleOfflineUUID(GetName()));
+	LOG(OldFilename.c_str());
+	auto NewFilename = GetUUIDFileName(m_UUID);
+	// Only move if there isn't already a new file
+	if (!cFile::IsFile(NewFilename) && cFile::IsFile(OldFilename))
+	{
+		cFile::CreateFolder(NewFilename.substr(0, 10));  // Ensure folder exists to move to
```
madmaxoft: I think we've come to the point where abstracting the folder name away into a separate function is necessary, instead of the magic substrings.

Please add a function `static AString GetUUIDFolderName(const cUUID & a_Uuid)` that returns the folder (relative to current folder) in which the player file for the specified UUID is located, then use `cFile::CreateFolderRecursive()` on it here and in `cPlayer::SaveToDisk()`.

src/Entities/Player.cpp
```diff
@@ -36,6 +36,25 @@
 // 1000 = once per second
 #define PLAYER_LIST_TIME_MS std::chrono::milliseconds(1000)
 
+namespace
```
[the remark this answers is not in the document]
peterbell10: While it is functionally equivalent to declaring the function `static`, I like the added structure of having all internal declarations in a single block.

src/Entities/Player.cpp
peterbell10: Done

src/Entities/Player.cpp
```diff
@@ -36,6 +36,41 @@
...
+This can be used both for online and offline UUIDs. */
+AString GetUUIDFolderName(const cUUID & a_Uuid)
+{
+	AString UUID = a_Uuid.ToShortString();
+
+	AString res("players/");
+	res.append(UUID, 0, 2);
+	res.push_back('/');
+	return res;
+}
```
madmaxoft: I think this should be a public API of the `cPlayer` class, even accessible from Lua, and it should already contain the `FILE_IO_PREFIX` in the returned path.

madmaxoft: I meant the plugins should have means to locate the file and folder where the player data is stored. Not the right PR for this, though.

madmaxoft:
Merge if CIs succeed.

Note that this will break existing servers using offline mode, players will lose their items. Do we care?

Issue #3682, opened 2017-04-27
Chest NBT Tag glitch


djlw78:
Client version:   1.11.2
Server OS:        Windows
Commit id:       7960d142ca3e28e0e4550faf131398899dca90dc

### Expected behavior
When pressing Ctrl+MiddleClick on a check it copies the check with NBT data of that chest's contents. Then placing it down you have a full copied chest.

### Actual behavior
When copying the appropriate size of chest then placing it down it comes out blank (no slots full) when opened.

### Steps to reproduce the behavior
Find a chest. Hover over then chest. Ctrl+MiddleClick. Find a new place. Place the chest down. Discover that it is empty. Be sad. (tested in the same chunk, not tested with trapped)

mrbesen:
Samething with dispensers, dropper, hopper, trapped chest and furnace.

PR #4179, opened 2018-02-03
Add an NBT tree structure

peterbell10:
This is the first step in fixing #3682 and adding support for storing arbitrary NBT in `cItem`.


bibo38:
Just for the naming I would prefer `NBTTree`, because I thought first, that you would add NBT data to a tree 🤣 .

peterbell10:
The name is meant to complement `cFastNBT` and `cParsedNBT`.

src/WorldStorage/TreeNBT.cpp
```diff
@@ -0,0 +1,205 @@
+
+#include "Globals.h"
+
+#include "WorldStorage/TreeNBT.h"
```
madmaxoft: This should not have the path - it's in the same folder.

src/WorldStorage/TreeNBT.h
```diff
@@ -0,0 +1,780 @@
+
+#pragma once
+
+#include "NBTDef.h"
```
madmaxoft: It would be beneficial to add a comment here explaining how to use the new classes, what is the designed API workflow etc. Since the file declares quite a few classes, one would get lost trying to figure out how to actually use it.

12xx12:
@peterbell10 are you still working on this? May I help you?

I'd like to use the NBT tags with my banner pull request

peterbell10:
As far as I remember this works as-is. It just never got reviewed.

bearbin:
Time for a review then, I'll try rebasing and have a go this evening.

12xx12:
So I rebased on master.

There were two words which required attention in MVSC.
The tests are not done with the new test "engine"

I fixed this but i'm not quite shure if I should commit.

In all I think the code should be merged

src/WorldStorage/TreeNBT.h
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	size_type size() const { return m_Tags.size(); }
+	bool empty() const { return m_Tags.empty(); }
+
+	void reserve(size_type a_ReserveSize) { m_Tags.reserve(a_ReserveSize); }
+
+	void clear() { m_Tags.clear(); }
+
+	void swap(cList & a_Other) NOEXCEPT
```
12xx12, line 437:
```suggestion
	void swap(cList & a_Other) noexcept
```

MSVC doesn't like this. Didn't check other

src/WorldStorage/TreeNBT.h
```diff
@@ -0,0 +1,484 @@
...
+	{
+		m_Tags.swap(a_Other.m_Tags);
+	}
+
+	friend void swap(cList & a_Lhs, cList & a_Rhs) NOEXCEPT
```
12xx12, line 442:
```suggestion
	friend void swap(cList & a_Lhs, cList & a_Rhs) noexcept
```

same here

tests/WorldStorage/TreeNBTTest.cpp
```diff
@@ -0,0 +1,228 @@
+
+#include "Globals.h"
+#include "WorldStorage/TreeNBT.h"
+
```
12xx12, line 4:
```suggestion

#include "../TestHelpers.h"
```

The new test "engine" wasn't implemented there
The rest of the file would also require some attention

12xx12 (changes requested):

12xx12:
Another thing: we should enable the Packet writer and the NBT write to write those tags into themselves for easy transimission and saving

tigerw:
Bearbin and I took a look at this and we came to the conclusion that this was a reimplementation of std::variant.

Something like this may suffice.

```cpp
#include <cassert>
#include <iostream>
#include <map>
#include <variant>
#include <vector>

class NBT
{
public:

	using List = std::vector<NBT>;
	using Compound = std::map<std::string, NBT>;

	using NubbyTree = std::variant<
		int,
		bool,
		List,
		Compound
	>;

	NBT(NubbyTree canopy)
	{
		top = canopy;
	}

	auto & Expose() const
	{
		return top;
	}

	void Poosh(NBT nubby)
	{
		assert(std::holds_alternative<List>(top));

		auto & lIsT = std::get<List>(top);
		assert(
			lIsT.empty() ||
			lIsT.back().top.index() == nubby.top.index()
		);

		lIsT.emplace_back(std::move(nubby));
	}

private:  // its privates

	NubbyTree top;
};

void MakeNubby()
{
	NBT Root{ NBT::List() };
	Root.Poosh({ 1 });
	Root.Poosh({ 2 });

	for (const auto & Item : std::get<NBT::List>(Root.Expose()))
	{
		std::cout << std::get<int>(Item.Expose()) << '\n';
	}
}

int main()
{
    MakeNubby();
}
```

12xx12:
Yeah I just noticed that also that you use a variant and make a nice interface

12xx12:
So what do we do now? I'd like to have the NBT for items added for other projects (Banner, Potions,...)

PureTryOut:
Has there been any progress on this recently?

12xx12:
Check the discussion in #5015

PR #4180, opened 2018-02-03
Add Zombie Villager

Bond-009:


src/Mobs/Monster.cpp
```diff
@@ -1176,33 +1179,44 @@ std::unique_ptr<cMonster> cMonster::NewMonsterFromType(eMonsterType a_MobType)
 
 			return cpp14::make_unique<cHorse>(HorseType, HorseColor, HorseStyle, HorseTameTimes);
 		}
+		case mtZombieVillager:
+		{
+			int Profession = Random.RandInt(6);
```
bibo38: #MagicNumber. Use or introduce some constant, which holds the number of `eVillagerType` members.

src/Mobs/Monster.cpp
```diff
@@ -1176,33 +1179,44 @@ std::unique_ptr<cMonster> cMonster::NewMonsterFromType(eMonsterType a_MobType)
...
+			if (Profession == 6)
+			{
+				// Give farmers a better chance of spawning
+				Profession = 0;
```
bibo38: Maybe a `std::piecewise_constant_distribution` (http://en.cppreference.com/w/cpp/numeric/random/piecewise_constant_distribution) is better suited.

src/Mobs/Monster.cpp
```diff
@@ -1176,33 +1179,44 @@ std::unique_ptr<cMonster> cMonster::NewMonsterFromType(eMonsterType a_MobType)
...
+			}
+
+			return cpp14::make_unique<cZombieVillager>(static_cast<cVillager::eVillagerType>(Profession));
```
bibo38: I would prefer a seperate function, which handles the conversation from integer to enum. This would allow adding asserts to check, if the integer is out of range (which would be UB with a `static_cast`).

src/Mobs/Villager.cpp
```diff
@@ -107,6 +107,26 @@ void cVillager::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 
+
+void cVillager::KilledBy(TakeDamageInfo & a_TDI)
+{
+	super::KilledBy(a_TDI);
+
+	// TODO: 0% chance on Easy, 50% chance on Normal and 100% chance on Hard
+	if (GetRandomProvider().RandBool(0.5) && (a_TDI.Attacker != nullptr) && (a_TDI.Attacker->IsMob()))
+	{
+		eMonsterType MonsterType = (reinterpret_cast<cMonster *>(a_TDI.Attacker)->GetMobType());
```
bibo38: Isn't a `static_cast` possible?

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -722,16 +723,21 @@ void cNBTChunkSerializer::AddMonsterEntity(cMonster * a_Monster)
 			case mtZombie:
 			{
 				const cZombie *Zombie = reinterpret_cast<const cZombie *>(a_Monster);
-				m_Writer.AddByte("IsVillager",   Zombie->IsVillagerZombie() ? 1 : 0);
-				m_Writer.AddByte("IsConverting", Zombie->IsConverting() ? 1 : 0);
-				m_Writer.AddInt ("Age",          Zombie->GetAge());
+				m_Writer.AddInt("Age",            Zombie->GetAge());
```
bibo38: Since this is now a single AddInt call, you can remove the spaces.

src/Mobs/Monster.cpp
peterbell10: Or better yet, a function that just returns a random profession to factor out from the `mtVillager` case above.

src/Mobs/Monster.cpp
peterbell10: You mean [`std::discrete_distribution`](http://en.cppreference.com/w/cpp/numeric/random/discrete_distribution).

bibo38: Yep, didn't see this nice distribution ;D.

src/cmake-build-debug/include/Globals.h
```diff
@@ -0,0 +1,2 @@
+/* This file allows Globals.h to be included with an absolute path */
+#include "/Users/Lode/dev/Github/cuberite/src/Globals.h"
```
bibo38: Get this file out of the PR. I would suggest rebasing (or just commit the removal and squash all commits, when the PR works).

To prevent this problem in the future, move the `cmake-build-debug` folder outside of the Git repository or into a directory named `build` (as the directory build is already in the gitignore file, so any subdirectory is ignored :smirk: )

Server/monsters.ini
```diff
@@ -212,3 +212,10 @@ AttackRate=1.0
 IsFireproof=1
 MaxHealth=20
 SightDistance=25.0
+
+[ZombieVillager]
+AttackDamage=4.0
+AttackRange=2.0
+AttackRate=1
```
bearbin: This should now have a decimal point to be consistent with other mobs.

src/Mobs/Villager.cpp
```diff
@@ -107,6 +107,26 @@ void cVillager::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
+
+void cVillager::KilledBy(TakeDamageInfo & a_TDI)
+{
+	super::KilledBy(a_TDI);
+
+	// TODO: 0% chance on Easy, 50% chance on Normal and 100% chance on Hard
```
bearbin: This is another place where difficulty needs to be moved back into the server from Core.

src/Mobs/Villager.cpp
```diff
@@ -203,7 +223,17 @@ bool cVillager::IsBlockFarmable(BLOCKTYPE a_BlockType)
...
-	return false;
 }
 
+
+
+
+
+cVillager::eVillagerType cVillager::GetRandomProfession()
+{
+	int Profession = GetRandomProvider().RandInt(cVillager::eVillagerType::vtMax - 1);
```
bearbin: This no longer encodes the increased farmer probability, is that intended?

src/Mobs/Zombie.cpp
```diff
@@ -9,10 +9,19 @@
 
 
 
+<<<<<<< HEAD
```
bearbin: Catastrophic merge failure detected!

src/Protocol/Protocol_1_11.cpp
```diff
@@ -1168,6 +1168,23 @@ void cProtocol_1_11_0::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_
...
+		case mtZombieVillager:
+		{
+			auto & ZombieVillager = reinterpret_cast<const cZombieVillager &>(a_Mob);
+			a_Pkt.WriteBEUInt8(ZOMBIE_IS_BABY);
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
+			a_Pkt.WriteBool(ZombieVillager.IsBaby());
+
+			a_Pkt.WriteBEUInt8(ZOMBIE_VILLAGER_CONVERTING);
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
+			a_Pkt.WriteBool(ZombieVillager.ConversionTime());
```
bearbin: This isn't a bool?

src/Protocol/Protocol_1_12.cpp
```diff
@@ -934,6 +931,23 @@ void cProtocol_1_12::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mo
...
+		case mtZombieVillager:
+		{
+			auto & ZombieVillager = reinterpret_cast<const cZombieVillager &>(a_Mob);
+			a_Pkt.WriteBEUInt8(ZOMBIE_IS_BABY);
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
+			a_Pkt.WriteBool(ZombieVillager.IsBaby());
+
+			a_Pkt.WriteBEUInt8(ZOMBIE_VILLAGER_CONVERTING);
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
+			a_Pkt.WriteBool(ZombieVillager.ConversionTime());
```
bearbin: Similarly, not a bool.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2943,15 +2945,40 @@ void cWSSAnvil::LoadWolfFromNBT(cEntityList & a_Entities, const cParsedNBT & a_N
...
 
-	std::unique_ptr<cZombie> Monster = cpp14::make_unique<cZombie>(IsVillagerZombie);
+	int AgeableIdx  = a_NBT.FindChildByName(a_TagIdx, "Age");
+	if (AgeableIdx > 0)
+	{
+		int Age;
+		switch (a_NBT.GetType(AgeableIdx))
+		{
+			case TAG_Byte: Age = static_cast<int>(a_NBT.GetByte(AgeableIdx)); break;
+			case TAG_Int:  Age = a_NBT.GetInt (AgeableIdx); break;
```
bearbin: There's an extraneous space before the `(`

src/Mobs/Villager.cpp
Bond-009: Yes

bibo38:
Since @bearbin fixed the TravisCI build on master, it would be good to rebase the branch (and maybe squash the Address comments), so that this PR also builds fine.

bearbin:
If this gets a positive review I will merge manually.

bearbin:
Merged manually in 4b3043f62

PR #4185, opened 2018-02-10
cLightingThread: Faster chunk reading

peterbell10:
Doesn't look like much but I'm measuring `cLightingThread::ReadChunks` to be ~10x faster.

### DIFF
--- a/src/LightingThread.cpp
+++ b/src/LightingThread.cpp
@@ -20,7 +20,6 @@ class cReader :
 	virtual void ChunkData(const cChunkData & a_ChunkBuffer) override
 	{
 		BLOCKTYPE * OutputRows = m_BlockTypes;
-		int InputIdx = 0;
 		int OutputIdx = m_ReadingChunkX + m_ReadingChunkZ * cChunkDef::Width * 3;
 		for (size_t i = 0; i != cChunkData::NumSections; ++i)
 		{
--- a/src/LightingThread.cpp
+++ b/src/LightingThread.cpp
@@ -22,19 +22,30 @@ class cReader :
 		BLOCKTYPE * OutputRows = m_BlockTypes;
 		int InputIdx = 0;
 		int OutputIdx = m_ReadingChunkX + m_ReadingChunkZ * cChunkDef::Width * 3;
-		int MaxHeight = std::min(+cChunkDef::Height, m_MaxHeight + 16);  // Need 16 blocks above the highest
-		for (int y = 0; y < MaxHeight; y++)
+		for (size_t i = 0; i != cChunkData::NumSections; ++i)
 		{
-			for (int z = 0; z < cChunkDef::Width; z++)
+			auto * Section = a_ChunkBuffer.GetSection(i);
+			if (Section == nullptr)
 			{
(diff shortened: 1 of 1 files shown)


src/LightingThread.cpp
```diff
@@ -22,19 +22,30 @@ class cReader :
...
-				a_ChunkBuffer.CopyBlockTypes(OutputRows + OutputIdx * 16, static_cast<size_t>(InputIdx * 16), 16);
-				InputIdx++;
-				OutputIdx += 3;
-			}  // for z
-			// Skip into the next y-level in the 3x3 chunk blob; each level has cChunkDef::Width * 9 rows
-			// We've already walked cChunkDef::Width * 3 in the "for z" cycle, that makes cChunkDef::Width * 6 rows left to skip
-			OutputIdx += cChunkDef::Width * 6;
-		}  // for y
+				// Skip to the next section
+				OutputIdx += 9 * cChunkData::SectionHeight * cChunkDef::Width;
```
madmaxoft: This seems wrong, I'd say one ` * cChunkDef::Width` is missing.

peterbell10: `OutputIdx` is multiplied by `cChunkDef::Width` whenever it's used so this is correct.

src/LightingThread.cpp
```diff
@@ -78,6 +88,7 @@ class cReader :
 		m_BlockTypes(a_BlockTypes),
 		m_HeightMap(a_HeightMap)
 	{
+		std::fill_n(m_BlockTypes, cChunkDef::NumBlocks * 9, E_BLOCK_AIR);
```
madmaxoft: Do I read this right that it first fills the entire buffer and then overwrites parts with the actual chunk data? Wouldn't is be even better to only fill the skipped parts of the buffer, on line 30?

peterbell10: I tried it out and it's 2-3x _slower_ if you fill selectively.

madmaxoft: That's weird, I'd assume it's caused by the cache being primed by this. Might be worth trying the perf out on an ARM chip to see if it has the same characteristics - the caches there, if present, work in different ways.

peterbell10: I don't think it's that weird really. Filling the whole thing is a single call to `memset` whereas filling a section at a time is ~100 different strided loops that are all inteleaved.

If I can get cross-compilation working later I'll try it out on ARM but I don't think it's likely to be that different.

madmaxoft: Oh, right, the chunk data is interleaved with other chunks' data, didn't think of that. Now it makes more sense, true. Still I'd like to see a comparison on an ARM, since those have completely different RAM access characteristics, so many optimizations like this are actually anti-optimizations on ARMs.

peterbell10: I got it cross-compiled for raspberry pi 3 and doing it this way is still ~2x faster.

madmaxoft: Great :)

madmaxoft:
Squash-merge if build succeeds

PR #4190, opened 2018-02-27
Add HOOK_MONSTER_MOVED

jonathathan:


NiLSPACE:
It still has to be documented for api.cuberite.org

Is it also possible to add the previous position as an argument? Other than that it looks fine.

jonathathan:
Thanks for the comments @NiLSPACE, I've added two commits.

Is this the correct way to add hook documentation?

NiLSPACE:
I'm not quite sure. @madmaxoft, do you know why it's failing?

madmaxoft:
You've documented the hook itself, but you also added a new constant to the API, `HOOK_MONSTER_MOVED`, which needs documenting too:
https://github.com/cuberite/cuberite/blob/999faa9d90453435ab1c61935a448fef01ff29e7/Server/Plugins/APIDump/Classes/Plugins.lua#L803

Server/Plugins/APIDump/Hooks/OnMonsterMoved.lua
```diff
@@ -0,0 +1,27 @@
+return
+{
+	HOOK_MONSTER_MOVED =
+	{
+		CalledWhen = "After a monster has moved to a new position.",
+		DefaultFnName = "OnMonsterMoved",  -- also used as pagename
+		Desc = [[
+            This hook is called after a {{cMonster|monster}} has moved, and the
+            entity move packet has been broadcast to clients.</p>
```
bibo38: Where is the opening `<p>` ?

src/Bindings/PluginManager.cpp
```diff
@@ -808,6 +808,25 @@ bool cPluginManager::CallHookLoginForge(cClientHandle & a_Client, AStringMap & a
...
 
 
+bool cPluginManager::CallHookMonsterMoved(cMonster & a_Monster, const Vector3d & a_OldPosition, const Vector3d & a_NewPosition)
+{
+	FIND_HOOK(HOOK_MONSTER_MOVED);
+	VERIFY_HOOK;
+
+	for (PluginList::iterator itr = Plugins->second.begin(); itr != Plugins->second.end(); ++itr)
+	{
+		if ((*itr)->OnMonsterMoved(a_Monster, a_OldPosition, a_NewPosition))
```
bibo38: I would suggest `for(auto * plugin : Plugins->second)`.
Or maybe merge https://github.com/cuberite/cuberite/pull/4027 first.

Server/Plugins/APIDump/Hooks/OnMonsterMoved.lua
madmaxoft: The APIDump plugin inserts the initial and terminal `<p>`; if you want more paragraphs, only the inner ones are needed. This code is correct.

bibo38: But now we have two terminal `</p>` :confused: ? Just look at the [cCraftingGrid-Doc](https://api.cuberite.org/cCraftingGrid.html):

    <p>Also, an object of this type is used in <a
    href="cCraftingRecipe.html">cCraftingRecipe</a>'s ConsumeIngredients() function for
    specifying the exact number of ingredients to consume in that recipe; plugins may use this to
    apply the crafting recipe.</p></p>

If I understand it correctly, I would need to use it like this:

    Some Text</p><p>New Paragraph

But not at the end of the of the text!

madmaxoft: Right, the terminating `</p>` shouldn't be here.

src/Bindings/PluginManager.cpp
madmaxoft: Merged, @jonathathan please update this branch (`git fetch upstream master:master`, `git rebase master`, fix conflicts)

Issue #3670, opened 2017-04-14
a few hundred wolves too many


nebulousdog:
Client version:   `1.11.2`
Server OS:         `Ubuntu 16.04.2 LTS (GNU/Linux 4.4.0-72-generic x86_64)`
Commit id:        `6c420ce5570ca4ac2aa5cabf80c3fa743d11ecf2`

### Expected behavior
"normal" amount of wolves

### Actual behavior
many hundreds of wolves

### Steps to reproduce the behavior
I noticed a "few" extra wolves around than usual, but when we logged in today, there were many many hundreds. We've starting killing them with lava, but there are still enough to be lagging the server.

<img width="1440" alt="doges" src="https://cloud.githubusercontent.com/assets/2218331/25030969/0a9585f8-207e-11e7-85e9-7a10503b8a12.png">

This picture is from the country side, but where there are holes where they can get trapped, its horrifying.

/cc @gavindinubilo

djlw78:
Congratulations. We are on the same page finding bugs when just doing normal stuff.

PR #4197, opened 2018-03-19
Wolves only spawn in taiga biomes

Bond-009:
Ref: https://minecraft.gamepedia.com/Wolf#Spawning

Maybe helps #3670

### DIFF
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -1565,8 +1565,7 @@ eMonsterType cFinishGenPassiveMobs::GetRandomMob(cChunkDesc & a_ChunkDesc)
 			break;
 		}
 
-		// Add wolves in forest and spruce forests
-		case biForest:
+		// Add wolves in taiga biomes
 		case biTaiga:
 		case biMegaTaiga:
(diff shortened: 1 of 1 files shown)


madmaxoft:
I could have sworn that I found wolves in regular forests in some older Java client, I guess there were too many wolves, so they reduced their spawn.

leeadama:
It's only in the "Bedrock" edition, what I understand used to be called "Pocket" where the wolves only spawn in Taiga.  In the full edition (Java) they spawn in several biomes.

Are you targeting the "Bedrock" edition ?  

Java Edition & Legacy Console Edition
Wolves spawn naturally in forests, taiga, mega taiga and cold taiga, as well as the "hills" and "mutated" variants of taiga, mega taiga and cold taiga, in packs of up to 4, but sometimes they may spawn in six. 

Bedrock Edition
Wolves only spawn naturally in taiga biomes and their variants. 16.6% will be babies.
Naturally-spawned wolves are untamed, and will become hostile if attacked by the player.

Bond-009:
Cuberite targets the Java Edition.
I just misread the wiki, it’s fixed in #4

PR #4198, opened 2018-03-25
Removed duplicate code

### LINKED ISSUE #3670: a few hundred wolves too many

Bond-009:
This should also fix #3670

### DIFF
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -14,6 +14,7 @@
 #include "../Simulator/FireSimulator.h"
 #include "../World.h"
 #include "../IniFile.h"
+#include "../MobSpawner.h"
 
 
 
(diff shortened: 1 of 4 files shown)


src/MobSpawner.cpp
```diff
@@ -340,6 +296,139 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
...
+		{
+			ListOfSpawnables.insert(mtHorse);
+			break;
+		}
+
+		// Add wolves in forest biomes
+		case biForest:
+		{
+			ListOfSpawnables.insert(mtWolf);
+		}
```
madmaxoft: A `break` seems to be missing here; if this is intentional, add a comment: `// fall-through`

src/MobSpawner.cpp
```diff
@@ -45,75 +45,31 @@ bool cMobSpawner::CheckPackCenter(BLOCKTYPE a_BlockType)
...
-		addIfAllowed(mtBat, allowedMobs);
-		addIfAllowed(mtSpider, allowedMobs);
-		addIfAllowed(mtZombie, allowedMobs);
-		addIfAllowed(mtSkeleton, allowedMobs);
-		addIfAllowed(mtCreeper, allowedMobs);
-		addIfAllowed(mtSquid, allowedMobs);
-		addIfAllowed(mtGuardian, allowedMobs);
-
-		if ((a_Biome != biDesert) && (a_Biome != biBeach) && (a_Biome != biOcean))
+		std::set<eMonsterType>::iterator itr = m_AllowedTypes.find(MobType);
```
madmaxoft: Use `auto`

src/MobSpawner.cpp
```diff
@@ -340,6 +290,109 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
...
+		if (
+			(a_Biome != biJungle) &&
+			(a_Biome != biJungleHills) &&
+			(a_Biome != biJungleEdge) &&
+			(a_Biome != biJungleM) &&
+			(a_Biome != biJungleEdgeM))
+		{
+			ListOfSpawnables.insert(mtOcelot);
+		}
+		if (
```
bearbin: These conditions for biomes look to be all inverted.

src/Generating/FinishGen.cpp
```diff
@@ -1505,100 +1506,26 @@ bool cFinishGenPassiveMobs::TrySpawnAnimals(cChunkDesc & a_ChunkDesc, int a_RelX
...
 {
 
-	std::set<eMonsterType> ListOfSpawnables;
+	std::vector<eMonsterType> ListOfSpawnables;
 	int chunkX = a_ChunkDesc.GetChunkX();
 	int chunkZ = a_ChunkDesc.GetChunkZ();
 	int x = (m_Noise.IntNoise2DInt(chunkX, chunkZ + 10) / 7) % cChunkDef::Width;
 	int z = (m_Noise.IntNoise2DInt(chunkX + chunkZ, chunkZ) / 7) % cChunkDef::Width;
 
-	// Check biomes first to get a list of animals
```
bearbin: Is re-constructing this vector every time not a waste of resources? I think this could be better done just once.

src/Generating/FinishGen.cpp
```diff
@@ -1505,100 +1506,26 @@ bool cFinishGenPassiveMobs::TrySpawnAnimals(cChunkDesc & a_ChunkDesc, int a_RelX
...
-	auto MobIter = ListOfSpawnables.begin();
-	using diff_type =
-		std::iterator_traits<decltype(MobIter)>::difference_type;
-	diff_type RandMob = static_cast<diff_type>
-		(static_cast<size_t>(m_Noise.IntNoise2DInt(chunkX - chunkZ + 2, chunkX + 5) / 7)
-		% ListOfSpawnables.size());
-	std::advance(MobIter, RandMob);
-
-	return *MobIter;
+	return ListOfSpawnables[GetRandomProvider().RandInt(ListOfSpawnables.size() - 1)];
```
peterbell10: I think this should still be using noise instead of rng as it's part of the generator.

src/World.cpp
```diff
@@ -1199,7 +1199,7 @@ void cWorld::TickMobs(std::chrono::milliseconds a_Dt)
 				else
 				{
 					auto & Wolf = static_cast<cWolf &>(Monster);
-					if (Wolf.IsAngry())
+					if (!Wolf.IsAngry())
 					{
 						Monster.Destroy(true);
```
peterbell10: Will this destroy tamed wolves as well?

src/MobSpawner.cpp
```diff
@@ -340,6 +290,109 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
...
+	std::set<eMonsterType> ListOfSpawnables;
+	// Check biomes first to get a list of animals
+	if (a_Biome == biNether)
+	{
+		ListOfSpawnables.insert(mtGhast);
+		ListOfSpawnables.insert(mtMagmaCube);
+		ListOfSpawnables.insert(mtZombiePigman);
+		ListOfSpawnables.insert(mtEnderman);
+	}
+	else if (a_Biome == biEnd)
```
peterbell10: Is there a reason you've moved away from `switch` to `if`-`else` chains?

Bond-009: Looks like I misread your last review :/

src/World.cpp
Bond-009: It shouldn't, wolves only get angry when they aren't tamed https://github.com/cuberite/cuberite/blob/master/src/Mobs/Wolf.cpp#L58

peterbell10: But this is destroying wolves that are _not_ angry.

src/MobSpawner.cpp
```diff
@@ -45,75 +45,26 @@ bool cMobSpawner::CheckPackCenter(BLOCKTYPE a_BlockType)
...
-	size_t allowedMobsSize = allowedMobs.size();
-	if (allowedMobsSize > 0)
+	if (AllowedMobs.empty())
 	{
-		return allowedMobs[GetRandomProvider().RandInt(allowedMobsSize - 1)];
+		return mtInvalidType;
 	}
-	return mtInvalidType;
+
+	return AllowedMobs[GetRandomProvider().RandInt(AllowedMobs.size() - 1)]
```
bearbin: You're missing a semicolon here, which is causing the builds to fail.

bearbin:
You're missing another semicolon - https://travis-ci.org/cuberite/cuberite/jobs/361752039#L1327

Can you try compiling on your local machine to avoid these silly omissions?

Bond-009:
@bearbin Sorry I couldn't compile on the machine I made those commits from, next time I'll wait till I'm behind my Linux machine :/

src/MobSpawner.cpp
```diff
@@ -340,6 +296,139 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
...
+	}
+
+	if (
+		(a_Biome != biDesertHills) &&
+		(a_Biome != biDesert) &&
+		(a_Biome != biDesertM) &&
+		(a_Biome != biBeach) &&
+		(a_Biome != biOcean) &&
+		(a_Biome != biDeepOcean))
+	{
```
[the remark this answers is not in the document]
Bond-009: IMO this is easier to understand than a switch statement that looks like this:
```c++
switch (biome)
    case biBeach:
    case biOcean:
    case ...:
         break;
    default:
    {
    // logic
    }
```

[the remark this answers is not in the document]
peterbell10: Okay, fair enough.

Issue #3516, opened 2017-01-02
Server thinks player is still in-game when disconnected for hours


jammet:
Client version:   1.8.9
Server OS:       Raspbian
Commit id:      c8b1d8a37d6073786df34836ed1e2c88652ac629

### Expected behavior
Client should have been properly disconnected and the player removed from the server.

### Actual behavior
Haven't been watching, but I am standing next to what appears to be a connected player, also listed in the player list, when this happened, according to the server log below.

Also, for some reason, the "player" was almost buried in spiders attacking it, but probably "invincible". Some of these spiders must have spawned in the well lit, tiny house he was standing in, most were crowding the outside walls.

```
[18:24:27] User Sledgie65 authenticated with UUID 43f2f65b9f2d4c69a8f70b7a247e7e91                                     <W>2017-01-01 19:46:55.779 1 => <4:Black_Lion(15)> Moved Black_Lion:4(15) to Closed Team 3[19:2]
[18:24:27] Player Sledgie65 has joined the game                                                                        <W>2017-01-01 19:49:12.778 1 => <15:Jammet(1)> Moved Jammet:15(1) to Closed Team 2[17:2]
[18:30:12] Kicking player Sledgie65 for "Too many blocks were destroyed per unit time - hacked client?"                <W>2017-01-01 19:49:15.809 1 => <5:Kirauni(29)> Moved Kirauni:5(29) to Closed Team 2[17:2]
[18:30:12] Kicking player Sledgie65 for "Too many blocks were destroyed per unit time - hacked client?"                <W>2017-01-01 19:49:32.901 1 => <5:Kirauni(29)> Kirauni not allowed to Move in Closed Team 3
[18:30:12] Kicking player Sledgie65 for "Too many blocks were destroyed per unit time - hacked client?"                <W>2017-01-01 19:49:56.819 1 => <4:Black_Lion(15)> Moved Black_Lion:4(15) to Closed Team 2[17:2]
[18:30:12] Kicking player Sledgie65 for "Too many blocks were destroyed per unit time - hacked client?"                <W>2017-01-01 22:03:42.275 1 => <5:Kirauni(29)> Connection closed: The remote host closed the connection [1]
[18:30:12] Kicking player Sledgie65 for "Too many blocks were destroyed per unit time - hacked client?"                <W>2017-01-01 22:09:56.228 1 => <3:(-1)> New connection: 84.191.173.195:49733
[...]

PR #4199, opened 2018-03-26
Properly destroy player

mathiascode:
A player would be partially disconnected from the server on timeout; the disconnect message would be sent to the client, but the player would remain in the world. This PR fixes the issue.


peterbell10:
This might fix #3516.

PR #4201, opened 2018-03-27
Broadcast playerlist removal in every world

### LINKED ISSUE #3138: Tab list sometimes doesn't update

mathiascode:
Fixes #3138


mathiascode:
To be honest, I don't know. I've been trying to look into the invisible player issue, but I can't find any consistent ways of reproducing the issue. We will have to see after merging I guess.

PR #4204, opened 2018-04-02
Update submodule's url in compile

earboxer:
People who had checked out the code earlier will find the build script not working.

(jsoncpp and libevent commit hash not found).

git submodule sync restores the urls so they point to the proper repositories, and can find the correct commits (cuberite/jsoncpp, instead of the previous one).

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -470,6 +470,7 @@ fi
 
 # Git: Submodules.
 echo " --- Updating submodules..."
+git submodule sync
 git submodule update --init
 
 


Cl1608Ho:
Looks good to me

madmaxoft:
Thanks for the `git sybmodule sync` trick, I had no idea this thing existed; I used to delete and re-clone / re-configure / re-compile all my repos. Because I usually keep like 4 copies on each of my computers, it was a pain :P

Issue #3062, opened 2016-03-01
Java interface


AlfieC:
Hey,

I'm mid-way through writing a Java interface for Cuberite, allowing for (almost) seamless integration of Spigot plugins. If I submit a PR, will it get accepted?

- alfie

madmaxoft:
Really? I somehow doubt this integration can be done. If the code is clean enough, I don't see a reason not to accept it; perhaps show us an example of your work in progress?

AlfieC:
It's not exactly the cleanest - primarily java based, uses java's native functionality to recreate all of the Bukkit API I can find suitable substitutes for.

madmaxoft:
How is it done, architecturally? Is it some form of a library linked to Cuberite, plus some glue code, or is it rather some form of a wrapper around the entire Cuberite server?

What I'm asking this for: If it's the latter, it might make more sense to actually create a new repo with the java bridge and use the Cuberite source as a submodule in that repo. Of course, we'd still need to push any changes required to the Cuberite source into this repo.

AlfieC:
Because the events would obviously need to be modifiable and preventable, it's a wrapper around Cuberite in its entirety.

bibo38:
This is a great idea, because everybody would be able to exchange their Bukkit-compatible server with Cuberite :smile: !
But it is currently **impossible**, because of a licensing issue.

The Bukkit-API is licensed under the [GPL v3](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/LICENCE.txt), while Cuberite is licensed under the [Apache License v2](https://github.com/cuberite/cuberite/blob/master/LICENSE).
In order to make this possible, Cuberite would need to be licensed under **GPL v3**, which is hard work.

bearbin:
@bibo38 I think you misunderestimate the problem: it would be more work to use the Bukkit API **code**, than to just write our own code that adheres to the Bukkit API. And simply being compatible does not mean that any copyright infringement has occurred.

bibo38:
@bearbin I don't think, that using the Bukkit API code would be more work(fewer work, because the Interfaces and Enums are availiable). The Bukkit API defines the basic Interfaces and Classes, which are guranteed to be availiable to every Bukkit Plugin.

The implementation is seperated and named CraftBukkit. But we need to provide at least everything, which is defined in the Bukkit API. This generates nearly the same source as Bukkit(maybe no Javadocs).

[...]

PR #4210, opened 2018-04-05
Cleaned up the CMake files and started seperating dependencies

bibo38:
I've updated the `CMakeLists.txt` to create a basis for seperating dependencies. This is mostly based on the `target_link_libraries` and the `target_include_directories` ability, to set includes/libraries only for specific targets. By issuing a `target_link_library` command for an executable, the include and library files/directories will now be automatically added to the executable.

Pros:
  - Fewer path related dependencies. The main path related dependencies are currently files in the `src` directory, as these are not compiled into a library, but into the Cuberite executable. E.g. the `Globals.h` include path must be manually added.
  - Fixed a little Clang error in `PluginLua.cpp`, which was otherwise ignored, because the lua includes were previously marked as system includes.
  - Removed the MSVC specific all in one project. Now it also has seperated projects per library.

Cons:
  - Currently the precompiled headers are broken (at least on MSVC), because each library now has the ability to set their custom precompiled header (so we can get rid of the `MSVC stupidness requires us...` include comments :smirk: ). But this would increase the changes dramatically. Maybe [Cotire](https://github.com/sakra/cotire) is an even better solution.
  - CMake version is now bumped to 3.5: I haven't found any reason to support older versions of CMake (but this changes should also work with CMake 3.2 (and maybe even lower)). And CMake also starts to support C++11 specific functionality in the newer versions (like [target_compile_features](https://cmake.org/cmake/help/v3.1/command/target_compile_features.html)).

P.S.: Did I mention the reduction of 200 lines :grin:


madmaxoft:
> Pros: Removed the MSVC specific all in one project. Now it also has seperated projects per library.

This is actually a con, the all-in-one project was there for a purpose.

madmaxoft:
Mostly, to settle dev differences. Windows devs want a single project, Linux devs want a lib-broken.

In my (Windows-centric) opinion, breaking into libs doesn't make sense, because they aren't libraries as such, there's no interface and no-one else could use them, it's just splitting the project into some (more or less artificial) groups of related files.

madmaxoft:
Somewhat related: [The F5 manifesto](http://www.khalidabuhakmeh.com/the-f5-manifesto-for-net-developers) - the link says .NET, but it's a valid point for classic C++ as well.

With libraries, MSVC doesn't work too F5-ish.

peterbell10:
How about making them [`OBJECT` libraries](https://cmake.org/cmake/help/v3.6/manual/cmake-buildsystem.7.html#id20) when building with MSVC?

It might make the structure closer to the linux build but still being a single build target.  Presumably that would also mean a single project in visual studio that can be F5'd.

madmaxoft:
How about making the Linux build use single project as well?

peterbell10:
The only downside I can think of is that tests and tools might want to link to something like OSSupport that is more free-standing.

madmaxoft:
They link to specific CPP files anyway, in order to avoid dragging in too many dependencies.
(But I agree, specifically pulling OSSupport out as a separate library would be useful for the future)

bibo38:
> The only downside I can think of is that tests and tools might want to link to something like OSSupport that is more free-standing.

I totally agree to make OSSupport and mbedTLS++ totally free standing (maybe even move them to lib?).

> How about making them OBJECT libraries when building with MSVC?

I want to minimize the differences in the `CMakeLists.txt` between Windows and Linux. Either we have a single huge project or a bunch of libraries.

> In my (Windows-centric) opinion, breaking into libs doesn't make sense, because they aren't libraries as such, there's no interface and no-one else could use them, it's just splitting the project into some (more or less artificial) groups of related files.

I agree. **But my main goal is to have some kind of library, which is used to build Tests and Tools.** The current situation of having the Tests and Tools using specific Cpp-files makes the addition/removal of dependencies harder, because you need to do it in the main Cuberite application and for each Test/Tool.

The simplest case would mean to have a Library libCuberite, which contains the whole source and an executable Cuberite, which only has the `main`-Method. But because I don't really like this simple separation (because of the tiny executable project), I've made libraries for each project with the aim, that these logical blocks may be independent parts in the future (e.g. if you want to have some Tool to modify World files, you just pick the WorldStorage library...).

And to realize things like #3062, I need Cuberite in a library form to attach Java bindings to it.

> They link to specific CPP files anyway, in order to avoid dragging in too many dependencies.

Sorry, but I don't get the problem of having the Tools/Tests to link to the whole Cuberite code.
I currently only see the benefits of having the Tools/Tests link against a (huge) libCuberite:
- Maybe no recompilation (haven't tested this yet, but should be because of the precompiled header compiler arguments...)
- No dependency problems. If I add fmt to libCuberite, I don't need to add it to the Test/Tools. Furthermore I don't have any linker problems (e.g. I only need `TCPLinkImpl.cpp`, but this also needs symbols from `IPLookup.cpp`).

madmaxoft:
Most tests cannot include the full `libCuberite`, because they usually test only a part of the code and they replace dependencies with dummy implementations (see for example https://github.com/cuberite/cuberite/blob/master/tests/Generating/Stubs.cpp ) - the full implementation would likely crash, because Cuberite's framework wouldn't be running within the test.

bibo38:
Ah, good to know...
Are the tools also affected by this stubbing? Or would they run with the full suite?

madmaxoft:
I *think* most tools are too unrelated to Cuberite code, they mostly don't share much more than just OSSupport.

PR #4211, opened 2018-04-05
Update m_LastSentPosition in TeleportToCoords and DoMoveToWorld.

### LINKED ISSUE #4049: New logged in players are not visible by other players
Client version:   1.12.2

### Expected behavior

The players can see each other.

### Actual behavior

New logged in players are not visible by other players.

1. Player 1 logs in
1. Player 2 logs in
1. Player 2 moves in front of Player 1

* Player 1 can't see Player 2
* Player 2 can see Player 1

peterbell10:
Fixes #4049 or at least the version described by @mathiascode.

The issue is that when a player is teleported, it can be spawned on clients before a movement update has been processed.  Since 5d64451f74af1839177f9b93d76cd393b7a58ad1 uses `m_LastSentPosition`, this means the spawn packets contain the position from _before_ being teleported.  Presumably the client then ignores packets from players teleporting from outside of the clients loaded chunks. 

This is fixed by updating `m_LastSentPosition` immediately when a player teleports.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1541,7 +1541,7 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	auto OldChunkCoords = cChunkDef::BlockToChunk(GetPosition());
 
 	// Set position to the new position
-	SetPosition(a_NewPosition);
+	ResetPosition(a_NewPosition);
 
 	// Stop all mobs from targeting this entity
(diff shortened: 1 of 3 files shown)


leeadama:
Can we get this applied soon ?  This is an large issue for many server owners.

madmaxoft:
Looks good to me, and we have confirmation it works: https://forum.cuberite.org/thread-3170-post-28991.html#pid28991

madmaxoft:
Squash-merge if CIs succeed

PR #4214, opened 2018-04-11
Use clang-tidy to check more code conventions

### LINKED ISSUE #4164: Use clang-tidy to check more code conventions
Clang-Tidy has the ability to check for things like the `m_` prefix on class member variables (as stated in the contribution guidlines).
Example usage: https://sarcasm.github.io/notes/dev/clang-tidy.html
All possible checks: https://clang.llvm.org/extra/clang-tidy/checks/list.html

There also exists a CMake integration (CMake >= 3.6).

So we should add this at least to the PR-Builds, so that @madmaxoft (or sb. else) doesn't need to point out missing prefixes 😸.

Bond-009:
Fixes #4164

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -4,5 +4,11 @@ CheckOptions:
   value:           m_
 #- key:             readability-identifier-naming.PrivateMemberCase
 #  value:           CamelCase
+#- key:             readability-identifier-naming.FunctionCase
+#  value:           CamelCase
+#- key:             readability-identifier-naming.Enum
+#  value:           CamelCase
+#- key:             readability-identifier-naming.EnumPrefix
+#  value:           e
 WarningsAsErrors: '*'
 #HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,14 +1,22 @@
 Checks: '-*,readability-identifier-naming'
 CheckOptions:
 - key:             readability-identifier-naming.PrivateMemberPrefix
-  value:           m_
+  value:           'm_'
+- key:             readability-identifier-naming.ClassConstantCase
+  value:           aNy_CasE
+- key:             readability-identifier-naming.ClassConstantPrefix
+  value:           ''
+- key:             readability-identifier-naming.ConstantMemberCase
+  value:           aNy_CasE
+- key:             readability-identifier-naming.ConstantMemberPrefix
+  value:           ''
 #- key:             readability-identifier-naming.PrivateMemberCase
 #  value:           CamelCase
 #- key:             readability-identifier-naming.FunctionCase
 #  value:           CamelCase
-#- key:             readability-identifier-naming.Enum
+#- key:             readability-identifier-naming.EnumCase
 #  value:           CamelCase
 #- key:             readability-identifier-naming.EnumPrefix
-#  value:           e
+#  value:           'e'
 WarningsAsErrors: '*'
 #HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,3 +2,4 @@ Checks: '-*,readability-identifier-naming'
 CheckOptions:
 - key:             readability-identifier-naming.PrivateMemberPrefix
   value:           m_
+WarningsAsErrors: '*'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -4,6 +4,7 @@ CheckOptions:
   value:           'm_'
 - key:             readability-identifier-naming.ClassConstantCase
   value:           aNy_CasE
+# an empty *Prefix needs a *Case to work
 - key:             readability-identifier-naming.ClassConstantPrefix
   value:           ''
 #- key:             readability-identifier-naming.PrivateMemberCase
@@ -13,4 +14,4 @@ CheckOptions:
 #- key:             readability-identifier-naming.EnumCase
 #  value:           camelBack
 WarningsAsErrors: '*'
-#HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
+HeaderFilterRegex: '/cuberite/src/\.?[^\.]'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,5 +2,7 @@ Checks: '-*,readability-identifier-naming'
 CheckOptions:
 - key:             readability-identifier-naming.PrivateMemberPrefix
   value:           m_
+#- key:             readability-identifier-naming.PrivateMemberCase
+#  value:           CamelCase
 WarningsAsErrors: '*'
 #HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,3 +3,4 @@ CheckOptions:
 - key:             readability-identifier-naming.PrivateMemberPrefix
   value:           m_
 WarningsAsErrors: '*'
+#HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,17 +6,11 @@ CheckOptions:
   value:           aNy_CasE
 - key:             readability-identifier-naming.ClassConstantPrefix
   value:           ''
-- key:             readability-identifier-naming.ConstantMemberCase
-  value:           aNy_CasE
-- key:             readability-identifier-naming.ConstantMemberPrefix
-  value:           ''
 #- key:             readability-identifier-naming.PrivateMemberCase
 #  value:           CamelCase
 #- key:             readability-identifier-naming.FunctionCase
 #  value:           CamelCase
 #- key:             readability-identifier-naming.EnumCase
-#  value:           CamelCase
-#- key:             readability-identifier-naming.EnumPrefix
-#  value:           'e'
+#  value:           camelBack
 WarningsAsErrors: '*'
 #HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,4 @@
+Checks: '-*,readability-identifier-naming'
+CheckOptions:
+- key:             readability-identifier-naming.PrivateMemberPrefix
+  value:           m_
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ logs
 players
 world
 world_nether
+world_the_end
 
 # cmake stuff
 CMakeFiles/
@@ -111,6 +112,10 @@ Server/cuberite_api.lua
 Server/official_undocumented.lua
 Server/NewlyUndocumented.lua
 Server/.luacheckrc
+compile_commands.json
 
 # compile.sh folder
 build-cuberite
+
+# clang-tidy
+tidy-fixes.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -117,5 +117,4 @@ compile_commands.json
 build-cuberite
 
 # clang-tidy
-tidy-fixes.yaml
 tidy-build
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,6 @@ logs
 players
 world
 world_nether
-world_the_end
 
 # cmake stuff
 CMakeFiles/
--- a/.gitignore
+++ b/.gitignore
@@ -118,3 +118,4 @@ build-cuberite
 
 # clang-tidy
 tidy-build
+run-clang-tidy.py
--- a/.gitignore
+++ b/.gitignore
@@ -119,3 +119,4 @@ build-cuberite
 
 # clang-tidy
 tidy-fixes.yaml
+tidy-build
--- a/circle.yml
+++ b/circle.yml
@@ -14,7 +14,7 @@ test:
  override:
   - cd src && find . -name \*.cpp -or -name \*.h > AllFiles.lst && lua CheckBasicStyle.lua
   - cd src/Bindings && lua CheckBindingsDependencies.lua
-  - curl -O "https://raw.githubusercontent.com/llvm-mirror/clang-tools-extra/45fb9b20ed7da2f6b95d83e5ef45bb536f49d8ca/clang-tidy/tool/run-clang-tidy.py"
-  - chmod +x run-clang-tidy.py
+  - curl -o "$HOME/bin/run-clang-tidy.py" "https://raw.githubusercontent.com/llvm-mirror/clang-tools-extra/45fb9b20ed7da2f6b95d83e5ef45bb536f49d8ca/clang-tidy/tool/run-clang-tidy.py"
+  - chmod +x "$HOME/bin/run-clang-tidy.py"
   - git submodule update --init
   - ./clang-tidy.sh
--- a/circle.yml
+++ b/circle.yml
@@ -4,9 +4,17 @@
 
 dependencies:
  pre:
-  - sudo apt-get install lua5.1
+  - sudo apt-get update
+  - sudo apt-get install lua5.1 cmake clang-3.9 clang-tidy-3.9
+  - sudo ln -s /usr/bin/clang-tidy-3.9 /usr/bin/clang-tidy
+  - sudo ln -s /usr/bin/clang-apply-replacements-3.9 /usr/bin/clang-apply-replacements
+  - pip install PyYAML
 
 test:
  override:
   - cd src && find . -name \*.cpp -or -name \*.h > AllFiles.lst && lua CheckBasicStyle.lua
   - cd src/Bindings && lua CheckBindingsDependencies.lua
+  - curl -O "https://raw.githubusercontent.com/llvm-mirror/clang-tools-extra/master/clang-tidy/tool/run-clang-tidy.py"
+  - chmod +x run-clang-tidy.py
+  - git submodule update --init
+  - ./clang-tidy.sh
--- a/circle.yml
+++ b/circle.yml
@@ -14,7 +14,7 @@ test:
  override:
   - cd src && find . -name \*.cpp -or -name \*.h > AllFiles.lst && lua CheckBasicStyle.lua
   - cd src/Bindings && lua CheckBindingsDependencies.lua
-  - curl -O "https://raw.githubusercontent.com/llvm-mirror/clang-tools-extra/master/clang-tidy/tool/run-clang-tidy.py"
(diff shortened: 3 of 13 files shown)


bibo38:
You're header filter currently fails (at least in CircleCI).
My output is:
```
Error while processing [...]/src/main.cpp.
[...]/src/../lib/SQLiteCpp/include/SQLiteCpp/Transaction.h:72:17: warning: invalid case style for private member 'mDatabase' [readability-identifier-naming]
    Database&   mDatabase;  //!< Reference to the SQLite Database Connection
                ^~~~~~~~~~
                m_mDatabase
[...]/src/../lib/SQLiteCpp/include/SQLiteCpp/Transaction.h:73:17: warning: invalid case style for private member 'mbCommited' [readability-identifier-naming]
    bool        mbCommited; //!< True when commit has been called
                ^~~~~~~~~~~
                m_mbCommited
[...]/src/./HTTP/../IniFile.h:50:23: warning: invalid case style for private member 'keys' [readability-identifier-naming]
        std::vector<key>     keys;
                             ^~~~~
                             m_keys
[...]/src/./HTTP/../IniFile.h:51:23: warning: invalid case style for private member 'names' [readability-identifier-naming]
        std::vector<AString> names;
                             ^~~~~~
                             m_names
[...]/src/./HTTP/../IniFile.h:52:23: warning: invalid case style for private member 'comments' [readability-identifier-naming]
        std::vector<AString> comments;
                             ^~~~~~~~~
                             m_comments
```
(I've invoked clang-tidy with `clang-tidy -header-filter=src/ ../src/main.cpp`)
Therefore CircleCI should fail the check.

Furthermore the exclusion of some files would be good to avoid errors:
```
Error while processing /home/ubuntu/cuberite/src/Bindings/LuaJson.cpp.
/home/ubuntu/cuberite/src/Bindings/./LuaState.h:44:10: error: 'LuaState_Typedefs.inc' file not found [clang-diagnostic-error]
         ^
```

Or maybe we could cache the generated files between the builds (only the generated header/source files...). Then we wouln't need those ugly excludes...

circle.yml
```diff
@@ -4,9 +4,17 @@
...
+  - sudo apt-get install lua5.1 cmake clang-3.9 clang-tidy-3.9
+  - sudo ln -s /usr/bin/clang-tidy-3.9 /usr/bin/clang-tidy
+  - sudo ln -s /usr/bin/clang-apply-replacements-3.9 /usr/bin/clang-apply-replacements
+  - pip install PyYAML
 
 test:
  override:
   - cd src && find . -name \*.cpp -or -name \*.h > AllFiles.lst && lua CheckBasicStyle.lua
   - cd src/Bindings && lua CheckBindingsDependencies.lua
+  - curl -O "https://raw.githubusercontent.com/llvm-mirror/clang-tools-extra/master/clang-tidy/tool/run-clang-tidy.py"
```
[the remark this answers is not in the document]
Bond-009: I think that in newer/some versions of the clang-tidy package this script in installed by default, including it in the main repo seems off.
Also possible license issues?

clang-tidy.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+FILE="tidy-fixes.yaml"
+REGEX=".+/cuberite/src/\w+/.+"
```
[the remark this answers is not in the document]
Bond-009: Yeah, I did this with the intent to exclude weird paths like '*/cuberite/src/../lib/*', forgot about the files directly in the 'src' dir, I will change the regex to include those and accept other dir names.

clang-tidy.sh
```diff
@@ -0,0 +1,25 @@
...
+	run-clang-tidy.py $ARGS
+fi
+
+if [ $? -ne 0 ]; then
+	echo "run-clang-tidy.py failed to run"
+	exit 1
+fi
+
+if [ -s $FILE ]; then
+	echo "Found violations, fixes can be found in $FILE"
```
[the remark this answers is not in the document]
peterbell10: I think treating warnings as errors should make `run-clang-tidy.py` fail the build on any check violation.

Try adding this to `.clang-tidy`
```yaml
WarningsAsErrors: '*'
```

clang-tidy.sh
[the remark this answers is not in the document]
peterbell10: I think the issue is include paths that follow parent directories like `src/../lib/`.  Maybe it would work if `\w+/` was replaced with `[^\.]*`.

circle.yml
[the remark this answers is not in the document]
peterbell10: I thought that even clang-3.9 included the python script. Either way, you need a newer version to return the correct error code. I am okay with downloading it here as long as you use a permalink instead of a branch link.

i.e. the link should have the full commit hash in it:
`https://raw.githubusercontent.com/llvm-mirror/clang-tools-extra/aa3548e35fb8c3700fcd2c6a22e9200e58237773/clang-tidy/tool/run-clang-tidy.py`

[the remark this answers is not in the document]
bibo38: >  Also possible license issues?

Not really a problem ([License file](https://github.com/llvm-mirror/clang-tools-extra/blob/master/LICENSE.TXT)). You only need to copy the license into cuberite (so that no one assumes, that the script is under the Apache License). See [Stack Overflow](https://softwareengineering.stackexchange.com/questions/234511/what-is-the-best-practice-for-arranging-third-party-library-licenses-paperwork) for an example.
(Or you just add `clang-tools-extra` as a submodule :rofl: )

Edit: We already have a folder with third party licenses: https://github.com/cuberite/cuberite/tree/master/Install/ThirdPartyLicenses

> I think that in newer/some versions of the clang-tidy package this script in installed by default

> I thought that even clang-3.9 included the python script.

Yeah, I've checked. The ArchLinux package ships the script. But it is placed at [`/usr/share/clang/run-clang-tidy.sh`](https://www.archlinux.org/packages/extra/x86_64/clang/files/) which is not part of the default `PATH`. Ubuntu on the other hand changes the filename to [`run-clang-tidy-6.0.py`](https://packages.ubuntu.com/bionic/amd64/clang-tidy-6.0/filelist). Therefore I would suggest to always use a fixed local one. This prevents nasty debugging problems, when somebody does have an incompatible `run-clang-tidy.py` script in their path.

> including it in the main repo seems off.

The alternative would be to guess the `run-clang-tidy.sh` path. Or to download the script inside `clang-tidy.sh` (which I dislike, because you create more hidden online dependencies. Currently it is enough to clone the repo and submodules to work offline. Furthermore this allows anyone, who can modify the response to execute airbitrary code... **So if you choose this option, check at least the hash of the downloaded file**).

clang-tidy.sh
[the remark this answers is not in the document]
bibo38: Just read in [the docs](https://clang.llvm.org/extra/clang-tidy/):

> This option overrides the 'HeaderFilter' option in .clang-tidy file, if any.

I think, that putting the regex to `.clang-tidy` is the superior choice.

clang-tidy.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+set -e
+
+FIXES_FILE="tidy-fixes.yaml"
```
bibo38: Is this file even necessary now?

.gitignore
```diff
@@ -111,6 +111,11 @@ Server/cuberite_api.lua
 Server/official_undocumented.lua
 Server/NewlyUndocumented.lua
 Server/.luacheckrc
+compile_commands.json
 
 # compile.sh folder
 build-cuberite
+
+# clang-tidy
+tidy-fixes.yaml
```
bibo38: Should be generated inside `tidy-build` by the script.

clang-tidy.sh
Bond-009: It isn't needed for the CI test, I left it in for local tests, auto-fixing, ...

bibo38: But then give the user the ability to enable auto fixing. E.g. pass all given args to `run-clang-tidy`. So I can run `./clang-tidy.sh -fix` to fix my errors...

.gitignore
```diff
@@ -111,6 +111,10 @@ Server/cuberite_api.lua
 Server/official_undocumented.lua
 Server/NewlyUndocumented.lua
 Server/.luacheckrc
+compile_commands.json
 
 # compile.sh folder
 build-cuberite
+
+# clang-tidy
+tidy-build
```
bibo38: If you don't plan to put `run-clang-tidy.py` into this repo, then add it to the list of ignored files.

.clang-tidy
```diff
@@ -0,0 +1,5 @@
+Checks: '-*,readability-identifier-naming'
+CheckOptions:
+- key:             readability-identifier-naming.PrivateMemberPrefix
+  value:           m_
+WarningsAsErrors: '*'
```
bibo38: As stated before, I would suggest to put the header filter into this file.
This just helps, if I want to run clang-tidy manually (maybe just on a single file...)

.gitignore
Bond-009: I'll add it, although not really necessary since you would just add the file to your PATH (after #4225 `run-clang-tidy.py` should be in PATH)

src/IniFile.h
```diff
@@ -42,14 +42,14 @@ class cIniFile : public cSettingsRepositoryInterface
 
 	struct key
 	{
-		std::vector<AString> names;
-		std::vector<AString> values;
-		std::vector<AString> comments;
+		std::vector<AString> m_names;
+		std::vector<AString> m_values;
+		std::vector<AString> m_comments;
```
peterbell10: If you're changing these anyway then shouldn't it be `m_Names`, `m_Values`, etc.

.clang-tidy
Bond-009: `Error parsing /home/bond/Documents/GitHub/cuberite/.clang-tidy: Invalid argument
YAML:6:15: error: unknown key 'HeaderFilter'
HeaderFilter: '.+/cuberite/src/[^\.].+'`
clang-tidy doesn't like it, or I don't know the YAML syntax good enough.

.gitignore
bibo38: If you intent, that everyone should have the file in the path, you should remove the `../run-clang-tidy.sh` check and just add the cuberite path to the `PATH` variable in CircleCI 1.

.clang-tidy
Bond-009: `HeaderFilterRegex: '.+/cuberite/src/[^\.].+'`
Now it run clang-tidy with this as header filter: `^/home/bond/Documents/GitHub/cuberite/tidy-build/.*`

src/IniFile.h
bibo38: Should be the `readability-identifier-naming.PrivateMemberCase` check. With the value of `CamelCase`

Bond-009: @bibo38 I'm currently avoiding adding new checks since every new check adds 300+ new changes that I need to make. There are a lot of checks that need to be added when the basics work.

src/Tracer.cpp
```diff
@@ -194,54 +194,54 @@ bool cTracer::Trace(const Vector3f & a_Start, const Vector3f & a_Direction, int
...
 		else
 		{
-			tMax.z += tDelta.z;
-			pos.z += step.z;
+			m_tMax.z += m_tDelta.z;
+			m_pos.z += m_step.z;
 		}
 
-		if (step.x > 0.0f)
+		if (m_step.x > 0.0f)
```
bibo38: The extra check was only meant to catch them all...

src/NetherPortalScanner.h
```diff
@@ -39,8 +39,8 @@ class cNetherPortalScanner : public cChunkStay
 	static const int SearchSolidBaseWidth = 3;
 
 	/** Where to place the player out from the face and across the face */
-	const double OutOffset = 2;
-	const double AcrossOffset = 0.5;
+	const double m_OutOffset = 2;
+	const double m_AcrossOffset = 0.5;
```
peterbell10: I think these are supposed to be `static` constants and so shouldn't have the `m_` prefix.

.clang-tidy
Bond-009: Ok, thanks. Yes, `IniFile.h` etc are checked, I don't know why, but they are.

src/NetherPortalScanner.h
```diff
@@ -39,8 +39,8 @@ class cNetherPortalScanner : public cChunkStay
 	static const int SearchSolidBaseWidth = 3;
 
 	/** Where to place the player out from the face and across the face */
-	const double m_OutOffset = 2;
-	const double m_AcrossOffset = 0.5;
+	const double OutOffset = 2;  // NOLINT
+	const double AcrossOffset = 0.5;  // NOLINT
```
peterbell10: Just use the `static` keyword, it shouldn't trigger clang-tidy.

Bond-009: Yeah, but than it fails to build for me :/

peterbell10: I forgot that you need to add a setting for 
```
readability-identifier-naming.ConstantMember
```
and that will override the `PrivateMember` setting when it's marked `const`.

bibo38: > and that will override the `PrivateMember` setting when it's marked `const`

But should this be allowed?:
```
class A
{
    const int x;
}
```

If not, then we should use `readability-identifier-naming.ClassConstantPrefix` to only match `static` class constants.

And note, that you need also need to set the case. Otherwise Clang-Tidy will ignore an empty prefix:
```
- key:             readability-identifier-naming.ClassConstantPrefix
  value:           ''
- key:             readability-identifier-naming.ClassConstantCase
  value:           'aNy_CasE'
```

src/NetherPortalScanner.h
```diff
@@ -39,8 +39,8 @@ class cNetherPortalScanner : public cChunkStay
 	static const int SearchSolidBaseWidth = 3;
 
 	/** Where to place the player out from the face and across the face */
-	const double OutOffset = 2;  // NOLINT
-	const double AcrossOffset = 0.5;  // NOLINT
+	const double OutOffset = 2;
+	const double AcrossOffset = 0.5;
```
bibo38: Why is this not `static`? The last build just failed, because `readability-identifier-naming.ClassConstantPrefix` only works on `static` constant fields.

.clang-tidy
```diff
@@ -1,14 +1,22 @@
 Checks: '-*,readability-identifier-naming'
 CheckOptions:
 - key:             readability-identifier-naming.PrivateMemberPrefix
-  value:           m_
+  value:           'm_'
+- key:             readability-identifier-naming.ClassConstantCase
+  value:           aNy_CasE
+- key:             readability-identifier-naming.ClassConstantPrefix
```
bibo38: `readability-identifier-naming.ClassConstantPrefix` is a specialisation of `readability-identifier-naming.ConstantMemberPrefix`. So you only need to change one of these keys (`readability-identifier-naming.ClassConstantPrefix` to only set the prefix for `static` members).

src/NetherPortalScanner.h
Bond-009: `/home/bond/Documents/GitHub/cuberite/src/Entities/../NetherPortalScanner.h:42:22: error: in-class initializer for static data member of type 'const double' requires 'constexpr' specifier [-Wstatic-float-init]
        static const double OutOffset = 2;`
`constexpr` is uses nowhere else in the code so I just left it and chenged the check.

bibo38: This only means, that you should put the definition outside the class:
```cpp
class X
{
    static const int z;
}
const int X::z = 42; // If you have a Cpp file, put this line into it.

Comment (Bond-009) [reply src/NetherPortalScanner.h]: @bibo38 How does that work when the member is private?

Comment (bibo38) [reply src/NetherPortalScanner.h]: > How does that work when the member is private?

What do you mean? It's like private member functions, which you define outside of the class.

Btw:
> If a static data member **of integral or enumeration type** is declared const (and not volatile), it can be initialized with an initializer in which every expression is a constant expression, right inside the class definition

(http://en.cppreference.com/w/cpp/language/static)
:rofl:

Comment (Bond-009) [reply src/NetherPortalScanner.h]: Yeah, it wasn't working because I made a typo :/, thanks.

File: .clang-tidy [hunk 14]
```diff
@@ -0,0 +1,16 @@
...
+  value:           ''
+#- key:             readability-identifier-naming.PrivateMemberCase
+#  value:           CamelCase
+#- key:             readability-identifier-naming.FunctionCase
+#  value:           CamelCase
+#- key:             readability-identifier-naming.EnumCase
+#  value:           camelBack
+WarningsAsErrors: '*'
+#HeaderFilterRegex: '.+/cuberite/src/[^\.].+'
```
bibo38: Just uncomment, even if it isn't necessary for the script.
I would also suggest to change it to `/cuberite/src/\.?[^\.]`:
- Allows to have cuberite in `/cuberite` (e.g. root path)
- Allows `/cuberite/src/./Foo.cpp` and `/cuberite/src/x.cpp`

src/Tracer.cpp
```diff
@@ -346,7 +346,7 @@ int cTracer::intersect3D_SegmentPlane(const Vector3f & a_Origin, const Vector3f
 
 
 
-int cTracer::GetHitNormal(const Vector3f & start, const Vector3f & end, const Vector3i & a_BlockPos)
+int cTracer::GetHitNormal(const Vector3f & a_start, const Vector3f & a_end, const Vector3i & a_BlockPos)
```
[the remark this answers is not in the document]
Bond-009: Why did I do this to myself? :confused:

.clang-tidy
```diff
@@ -0,0 +1,16 @@
+Checks: '-*,readability-identifier-naming'
+CheckOptions:
+- key:             readability-identifier-naming.PrivateMemberPrefix
+  value:           'm_'
+- key:             readability-identifier-naming.ClassConstantCase
+  value:           aNy_CasE
+- key:             readability-identifier-naming.ClassConstantPrefix
+  value:           ''
```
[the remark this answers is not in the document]
peterbell10: An empty prefix is also the default, the key point should be ignoring the `PrivateMember` behaviour.

clang-tidy.sh
```diff
@@ -0,0 +1,20 @@
...
+REGEX=".+/cuberite/src/[^\.].+"
+# TODO: Add -quiet when Circle CI is updated to 2.0 with a newer version of Ubuntu
+ARGS="-header-filter $REGEX -export-fixes $FIXES_FILE "$@" $REGEX"
+
+mkdir -p tidy-build
+cd tidy-build
+cmake --target Cuberite -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
+run-clang-tidy.py $ARGS || TIDY_FAILED=1
+
+if run-clang-tidy.py $ARGS; then
```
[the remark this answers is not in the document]
Bond-009: Good catch :+1:

PR #4217, opened 2018-04-26
Fix fishing timer

peterbell10:
Fixes ["Fishing Speed Too Slow"](https://forum.cuberite.org/thread-3175-post-29000.html#pid29000).

Interestingly, the constants @NiLSPACE points out are actually correct:
```cpp
(Random.RandInt(100, 900) - static_cast<int>(a_Player->GetEquippedItem().m_Enchantments.GetLevel(cEnchantments::enchLure) * 100))
```
100 to 900 ticks is the correct timing of 5-45 seconds. However, the timer is only updated when the floater is in the water and the server side position was actually bobbing in and out of the water. This meant the timer took ~2-3x longer than it should.

With this change the floater position is always in the water and so the timer works as expected.

### DIFF
--- a/src/Entities/Floater.cpp
+++ b/src/Entities/Floater.cpp
@@ -147,6 +147,11 @@ void cFloater::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 				}
 			}
 		}
+	}
+
+	// Check water at the top of floater otherwise it floats into the air above the water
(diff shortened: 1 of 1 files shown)


leeadama:
I made the change in your commit, rebuilt it and tested on a clear day with the sun directly over head using a regular pole no enchantments (same as last time)

This time I got :

1. 21 Seconds
2. 6 Seconds
3. 19 Seconds
4. 37 Seconds
5. 36 Seconds
6. 12 Seconds

Looks good! All of them are within the range expected (5 - 45 sec).  And it makes fishing a much more interesting proposition.

Thank You!

Issue #4215, opened 2018-04-16
replace all NULL with nullptr in src/Bindings/Bindings.cpp


QianNangong:
`NULL` and `0` as null pointer is deprecated. When compiling in `Release` scheme on macOS, `-Werror,-Wzero-as-null-pointer-constant` is enabled, so it will fail.
`nullptr` is accepted by all c++ compiler with c++ 11 support, so it should not break anything.

bibo38:
The main problem is that `Bindings.cpp` is generated with changing the code via [BindingsProcessor.lua](https://github.com/cuberite/cuberite/blob/master/src/Bindings/BindingsProcessor.lua) and then applying ToLua++. So you would need to change ToLua++ to fix this problem.

For non OSX-Systems I've added a quite ugly workaround in the [CMakeLists.txt](https://github.com/cuberite/cuberite/blob/master/src/Bindings/CMakeLists.txt#L172). Basically it adds the correct flags to Clang, if it detects the correct version. Without version checking, it would fail on older Clangs, because these don't know the warning...
But sadly the Clang Versions are different on OSX. Therefore they are currently ignored.

madmaxoft:
I think changing ToLua++ is the way to go, it should not be too difficult to do :)

CNAmira:
@bibo38 According to CMP0025 in CMake, Apple Clang is identified as `AppleClang` instead of `Clang`.

bibo38:
@CNAmira Thanks ;D. Didn't know that! (But I'm wondering, why AppleClang doesn't need these waning suppression)

PR #4218, opened 2018-04-27
Update Bindings to generate nullptr instead of NULL.

madmaxoft:
Ref.: #4215


bibo38:
Don't miss to remove my Clang hack from [Bindings/CMakeLists.txt](https://github.com/cuberite/cuberite/blob/master/src/Bindings/CMakeLists.txt#L172) 😏.

Issue #915, opened 2014-04-21
Add enchantment effects.


bennasar99:
Enchantments that need to be implemented:
- [x] Aqua Affinity
- [x] Bane Of Arthropods
- [x] Blast Protection
- [x] Efficiency
- [x] Feather Falling
- [x] Fire Aspect
- [x] Fire Protection
- [x] Flame
- [ ] Fortune
- [x] Infinity
- [x] Knockback
- [x] Looting
- [x] Luck Of The Sea
- [x] Lure
- [x] Power
- [x] Projectile Protection
- [x] Protection
- [x] Punch
- [x] Respiration
- [x] Sharpness
- [x] SilkTouch
- [x] Smite
- [x] Thorns
- [x] Unbreaking
- [ ] Depth Strider
- [ ] Frost Walker
- [ ] Curse of Binding
- [ ] Sweeping Edge
- [ ] Mending
- [ ] Curse of Vanishing

Bond-009:
Can somebody make this into a list and check the ones that are already done?

peterbell10:
I've turned it into a list.  Do you know which ones are done though?

Bond-009:
I searched trough the code and these are already implemented: Luck Of The Sea, Lure, Infinity, Flame, Punch, Power, Unbreaking, Silk Touch, Efficiency, Looting, Fire Aspect, Knockback, Bane Of Arthropods, Smite, Sharpness, Thorns, Aqua Affinity, Respiration, Projectile Protection, Blast Protection, Feather Falling, Fire Protection and Protection.

bearbin:
I belive that unbreaking is only implemented for tools and weapons, not armour, so I have removed that from the list.

PR #4220, opened 2018-04-27
Add unbreaking for armor

Bond-009:
Ref: https://minecraft.gamepedia.com/Enchanting#Unbreaking

#915

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -941,10 +941,29 @@ void cPlayer::ApplyArmorDamage(int DamageBlocked)
 	{
 		ArmorDamage = 1;
 	}
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 0, ArmorDamage);
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 1, ArmorDamage);
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 2, ArmorDamage);
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 3, ArmorDamage);
+
+	for (int i = 0; i < 4; i++)
+	{
+		// (60 + (40 / (UnbreakingLevel + 1)))% chance the armor takes damage
+		// Ref: https://minecraft.gamepedia.com/Enchanting#Unbreaking
+		unsigned int UnbreakingLevel = m_Inventory.GetArmorSlot(i).m_Enchantments.GetLevel(cEnchantments::enchUnbreaking);
+		double chance = (0.6 + (0.4 / (UnbreakingLevel + 1)));
+
+		// When durability is reduced by multiple points
+		// Unbreaking is applied for each point of reduction.
+		for (int dp = 1; dp <= ArmorDamage; dp++)
+		{
+			if (!GetRandomProvider().RandBool(chance))
+			{
+				continue;
+			}
+
+			if (m_Inventory.DamageItem(cInventory::invArmorOffset + i, 1))
+			{
+				m_World->BroadcastSoundEffect("entity.item.break", GetPosition(), 0.5f, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
+			}
+		}
+	}
 }
 
 
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -936,11 +936,7 @@ void cPlayer::SetFlying(bool a_IsFlying)
 
 void cPlayer::ApplyArmorDamage(int a_DamageBlocked)
 {
-	short ArmorDamage = static_cast<short>(a_DamageBlocked / 4);
-	if (ArmorDamage == 0)
-	{
-		ArmorDamage = 1;
-	}
+	short ArmorDamage = static_cast<short>(std::max(a_DamageBlocked / 4, 1));
 
 	for (int i = 0; i < 4; i++)
 	{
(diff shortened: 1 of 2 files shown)


src/Entities/Player.cpp
```diff
@@ -941,10 +941,29 @@ void cPlayer::ApplyArmorDamage(int DamageBlocked)
...
+		// (60 + (40 / (UnbreakingLevel + 1)))% chance the armor takes damage
+		// Ref: https://minecraft.gamepedia.com/Enchanting#Unbreaking
+		unsigned int UnbreakingLevel = m_Inventory.GetArmorSlot(i).m_Enchantments.GetLevel(cEnchantments::enchUnbreaking);
+		double chance = (0.6 + (0.4 / (UnbreakingLevel + 1)));
+
+		// When durability is reduced by multiple points
+		// Unbreaking is applied for each point of reduction.
+		for (int dp = 1; dp <= ArmorDamage; dp++)
+		{
+			if (!GetRandomProvider().RandBool(chance))
```
peterbell10: Another way to do this would be with [`std::binomial_distribution`](http://en.cppreference.com/w/cpp/numeric/random/binomial_distribution).

```cpp
std::binomial_distribution<short> Dist(ArmorDamage, chance);
auto ReducedDamage = Dist(GetRandomPovider().Engine());
m_Inventory.DamageItem(cInventory::invArmorOffset + i, ReducedDamage)
```

src/Entities/Player.cpp
```diff
@@ -941,10 +941,32 @@ void cPlayer::ApplyArmorDamage(int DamageBlocked)
...
 	}
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 0, ArmorDamage);
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 1, ArmorDamage);
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 2, ArmorDamage);
-	m_Inventory.DamageItem(cInventory::invArmorOffset + 3, ArmorDamage);
+
+	for (int i = 0; i < 4; i++)
+	{
+		// (60 + (40 / (UnbreakingLevel + 1)))% chance the armor takes damage
+		// Ref: https://minecraft.gamepedia.com/Enchanting#Unbreaking
```
bibo38: This comment should be over `double chance = [...]`.

src/Entities/Player.cpp
```diff
@@ -941,10 +941,32 @@ void cPlayer::ApplyArmorDamage(int DamageBlocked)
...
+		{
+			continue;
+		}
+		unsigned int UnbreakingLevel = ArmorPiece.m_Enchantments.GetLevel(cEnchantments::enchUnbreaking);
+		double chance = (0.6 + (0.4 / (UnbreakingLevel + 1)));
+
+		// When durability is reduced by multiple points
+		// Unbreaking is applied for each point of reduction.
+		std::binomial_distribution<short> Dist(ArmorDamage, chance);
+		auto ReducedDamage = Dist(GetRandomProvider().Engine());
```
bibo38: (I'll personally prefer to use the concrete types for simple values. E.g. `short` instead of `auto`.)

src/Entities/Player.cpp
```diff
@@ -941,10 +941,32 @@ void cPlayer::ApplyArmorDamage(int DamageBlocked)
...
+		if (ReducedDamage == 0)
+		{
+			continue;
+		}
+		if (m_Inventory.DamageItem(cInventory::invArmorOffset + i, 1))
+		{
+			m_World->BroadcastSoundEffect("entity.item.break", GetPosition(), 0.5f, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
```
bibo38: These calculations should definitely go into a seperate function. And the `0.75` + `GetUniqueID()` calculation should be explained (at least by using a function name like `CalculateSoundVolume`).

Bond-009:
@bibo38 It's only for armor because Unbreaking was already implemented for tools and it calculates the chance in another way. https://github.com/cuberite/cuberite/issues/915#issuecomment-355769905

src/Entities/Player.cpp
Bond-009: I just copied this from here: https://github.com/cuberite/cuberite/blob/master/src/Entities/Player.cpp#L2349
for consistency.

src/Entities/Player.cpp
```diff
@@ -934,17 +934,18 @@ void cPlayer::SetFlying(bool a_IsFlying)
 
 
 
-void cPlayer::ApplyArmorDamage(int DamageBlocked)
+void cPlayer::ApplyArmorDamage(int a_DamageBlocked)
 {
-	short ArmorDamage = static_cast<short>(DamageBlocked / 4);
+	short ArmorDamage = static_cast<short>(a_DamageBlocked / 4);
```
bibo38: > Any hit from a damage source that can be blocked by armor will remove one point of durability from each piece of armor worn for every 4 () of incoming damage (rounded down, **but never below 1**).

(https://minecraft.gamepedia.com/Armor#Damage_types)

Bond-009: Yes, that is what these lines do. Am I missing something?

bibo38: You're missing `std::min`. Otherwise damages of `3` or lower would become `0`.

Bond-009: The next lines change the value to 1 if it's 0, but `std::min` is indeed better, I will change it.

bibo38: Oh, just overlooked the following if.

Issue #4221, opened 2018-04-28
Change to Circle CI 2.0


Bond-009:
Circle CI 1.0 will stop working August 31.
```
We’re sunsetting CircleCI 1.0: August 31, 2018 is the final day for 1.0 builds.
Refer to CircleCI 2.0 Hello World document for users and developers to get started with CircleCI 2.0. 
```
This issue is blocking #4214 since the current Ubuntu image being used doesn't have clang-tidy in its package repo.
As a temporary fix someone with acces to the Circle CI Project settings could change the Ubuntu image to 14.04
Ref: https://circleci.com/docs/1.0/build-image-trusty/

peterbell10:
I think I've set Circle to 14.04, can you try pushing to #4214?

PR #4225, opened 2018-05-03
WIP: Move to CircleCI-2.0

peterbell10:


peterbell10:
I got it working eventually: https://circleci.com/workflow-run/ceb726c9-6a76-4ad4-8bed-41da4da81f29.
Should probably wait for #4214 to be merged before moving over to it though.

bibo38:
>  Should probably wait for #4214 to be merged before moving over to it though.

Isn't it the other way round? #4221 mentions, that it blocks #4214 and that Ubuntu 14.04 is only a temporary fix. Furthermore you've mentioned, that the currently used `clang-tidy` version lacks support for returning the correct error code (see https://github.com/cuberite/cuberite/pull/4214#discussion_r184867079). So I think, that this PR should be merged first, so that #4214 doesn't need to get it working with very old `clang-tidy` versions.

peterbell10:
I was under the impression that #4214 had already gotten the CircleCI specific stuff working and it was mainly a matter of improving the `.clang-tidy` file and shell script.  As such, I didn't think #4214 would become any simpler by changing.

I guess my interpretation doesn't really matter though; @Bond-009 would know best.

Bond-009:
Yeah, all CircleCI specific stuff has already been dealt with and I'm planning to only change the script and config file. But, I'm not stopping you from merging this first :)

peterbell10:
Since CircleCI 2.0 is now reporting each job separately, the required status checks will need to be updated.

https://circleci.com/docs/2.0/workflows-waiting-status/

.circleci/images/cuberite-circleci/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ubuntu:trusty
+
+# Add LLVM 6.0 repository
+RUN apt-get update && \
+	apt-get install -y software-properties-common wget && \
+	wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add - && \
+	apt-add-repository "deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-6.0 main" && \
+	apt-add-repository ppa:ubuntu-toolchain-r/test
```
bibo38: For what do you need `ubuntu-toolchain-r/test`?

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+    working_directory: ~/cuberite
+    docker:
+    - image: peterbell10/cuberite-circleci:0.0.2
+    steps:
+    - checkout
+    - run: git submodule update --init
+    - save_cache:
+        key: "cuberite-{{ .Environment.CIRCLE_SHA1 }}"
+        paths:
+        - ~/cuberite
```
bibo38: Just recently read on [Travis CI](https://docs.travis-ci.com/user/caching/):

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source:

Did you check, if caching leads to any speed improvements?

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
+version: 2
+jobs:
+  checkout_cuberite:
+    working_directory: ~/cuberite
+    docker:
+    - image: peterbell10/cuberite-circleci:0.0.2
```
bibo38: Can we auto-deploy this image to the docker hub? Or at least to a place, where multiple users could update the image?
I'm mainly thinking about ways to update the image via pull requests...

peterbell10: Seems like [it's possible](https://docs.docker.com/docker-hub/builds/). In fact [cuberite/docker-ci](https://github.com/cuberite/docker-ci) might even be doing something similar already.

.circleci/images/cuberite-circleci/Dockerfile
bibo38: Do you have a source? I just found the deb command at https://apt.llvm.org/

Btw.: Why don't use Ubuntu Bionic (18.04LTS)? It has `clang-tidy` already in the [main repositories](https://packages.ubuntu.com/bionic/clang-tidy-6.0).

.circleci/config.yml
bibo38: And how about the packaging and uploading? If it's below ~10s, I'm fine. Otherwise checking out both would be faster.

.circleci/images/cuberite-circleci/Dockerfile
peterbell10: > Do you have a source?

It's on the page you linked
```
# Also add the following for the appropriate libstdc++
deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu trusty main
```

> Why don't use Ubuntu Bionic

I'll try it out

bibo38: Ah, there it is hidden :bulb: ... Looks just soo similar

.circleci/config.yml
peterbell10: It's ~8s.

.circleci/images/cuberite-circleci/Dockerfile
```diff
@@ -1,18 +1,11 @@
...
 RUN apt-get update && \
 	apt-get install -y git lua5.1 cmake clang-6.0 clang-tidy-6.0 python-pip && \
 	pip install PyYAML 
 
-RUN update-alternatives --install /usr/bin/clang             clang             /usr/bin/clang-6.0          600 && \
-	update-alternatives --install /usr/bin/clang++           clang++           /usr/bin/clang++-6.0        600 && \
-	update-alternatives --install /usr/bin/clang-tidy        clang-tidy        /usr/bin/clang-tidy-6.0     600 && \
-	update-alternatives --install /usr/bin/run-clang-tidy.py run-clang-tidy.py /usr/bin/run-clang-tidy-6.0 600
+RUN update-alternatives --install /usr/bin/clang             clang             /usr/bin/clang-6.0             600 && \
+	update-alternatives --install /usr/bin/clang++           clang++           /usr/bin/clang++-6.0           600 && \
```
bibo38: Maybe use spaces to keep the correct indentation level despite editor settings.
Or use something like this:
```
RUN \
    update-alternatives ... \
    update-alternatives ...

File: .circleci/config.yml [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+        paths:
+        - ~/cuberite
+
+  check_basic_style:
+    working_directory: ~/cuberite/src
+    docker:
+    - image: peterbell10/cuberite-circleci:0.0.3
+    steps:
+    - restore_cache: { keys: [ "cuberite-{{ .Environment.CIRCLE_SHA1 }}" ] }
+    - run: find . -name \*.cpp -or -name \*.h > AllFiles.lst
```
bibo38: It is already created by [src/CMakeLists.txt](https://github.com/cuberite/cuberite/blob/master/src/CMakeLists.txt#L285).
But maybe it would be a good additional test to check, that the `find` output is identical to the one from CMake (in case it missed a file). I could create an Issue, if this new test doesn't fit into this PR.

.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
...
+    - run: lua CheckBasicStyle.lua
+    - run: cd Bindings && lua CheckBindingsDependencies.lua
+
+  clang-tidy:
+    working_directory: ~/cuberite
+    docker:
+    - image: peterbell10/cuberite-circleci:0.0.3
```
bibo38: You could use YAML-References to reduce the version duplication. See [.travis.yml](https://github.com/cuberite/cuberite/blob/master/.travis.yml#L21) for an example.

.circleci/images/cuberite-circleci/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:bionic
+
+# Install build dependencies
+RUN apt-get update && \
+	apt-get install -y git lua5.1 cmake clang-6.0 clang-tidy-6.0 python-pip && \
+	pip install PyYAML
```
bibo38: Already in the Ubuntu repository as [python-yaml](https://packages.ubuntu.com/bionic/python-yaml).

.circleci/config.yml
peterbell10: `cmake` doesn't run on this job.

.circleci/config.yml
```diff
@@ -0,0 +1,41 @@
+version: 2
+jobs:
+  checkout_cuberite:
+    working_directory: ~/cuberite
+    docker: &cube_docker
+    - image: peterbell10/cuberite-circleci:0.0.3
```
madmaxoft: I'm not too familiar with Docker, does this mean that it uses something outside the canon repo? Can it be moved into the main repo?

peterbell10: Yes, It's on dockerhub and it has to be for Circle to be able to download it.  There is a cuberite dockerhub account though and I think [cuberite/docker-ci](https://github.com/cuberite/docker-ci) might be automatically build by dockerhub.  I'm not sure though, maybe @bearbin knows?

peterbell10: Okay, I've had a look at [docker hub's docs](https://docs.docker.com/docker-hub/builds/#understand-the-build-process) and it seems that the dockerfile has to be in its own repo for the automated builds to work.  After that's set up, whoever is in control of the [cuberite dockerhub account](https://hub.docker.com/u/cuberite/) would need to setup an automated build.

peterbell10: Alternatively, we could setup a new branch for [cuberite/docker-build-image](https://hub.docker.com/r/cuberite/docker-build-image/) that includes all of the clang stuff.

peterbell10: @madmaxoft could you create a new circle-ci branch on the cuberite/docker-build-image github?

peterbell10: It's actually literally called [cuberite/docker-build-image](https://github.com/cuberite/docker-build-image).  That's the one that's already setup to build automatically.

madmaxoft: I've added you there as well.

Mindavi:
After August the 31st CircleCI 1.0 will cease to exist.

PR #4226, opened 2018-05-03
Replaced C style arrays with std::array

bibo38:
Like said in the description, I've gone through the source and replaced the C-style arrays with instances of `std::array`. I've catched most of them by removing the `ARRAYCOUNT` macro. But there may be arrays, which I've missed.

This should solve one checkpoint on #1401


madmaxoft:
I'd say it's caused by the error two lines down from that - it uses an undefined type.

bibo38:
Ok, that could explain it. Is this some sort of VS2013 bug, because the type is the class, in which this static variable of `std::array` is declared?

bibo38:
Just found a whole bunch of other C-style arrays via `grep -E -R '[a-zA-Z_:]+\s+[a-zA-Z]+\s*\[' src/ | grep -v // | grep -v '/\*\*'`.
After fixing these, I'll remove the WIP prefix.

tigerw:
wowee that's a big change, but in theory nothing should break if there weren't any typos, right>

pwnorbitals:
This looks like a good refactoring but would need a lot of work to review properly.
@bibo38, could you please rebase and split into multiple PRs ? for example, one per folder

PR #4227, opened 2018-05-07
Now checking Variable prefixes with Clang-Tidy

bibo38:
(and I always thought, that reviews would catch 98% of these style violations :fox_face: )

Furthermore I've added the `g_`-Prefix to the coding conventions, because I've seen it being used a lot in the cuberite source.

The only thing I'm unsure is, if constant expressions also should get the prefix. This causes things like this (`Discriminator` is just a namespace):
```cpp
Buf.WriteBEInt8(Discriminator::g_ServerHello);
```
(https://github.com/bibo38/Cuberite/blob/7e7a844b46109b75a6a0c54ac4f9a55de2be03a3/src/Protocol/ForgeHandshake.cpp#L129)

P.S.: @Bond-009 `./clang-tidy.sh -fix` prevents headaches :smile:

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,13 @@
 Checks: '-*,readability-identifier-naming'
 CheckOptions:
-- key:             readability-identifier-naming.PrivateMemberPrefix
+- key:             readability-identifier-naming.MemberPrefix
+  value:           'm_'
+- key:             readability-identifier-naming.ParameterPrefix
+  value:           'a_'
+- key:             readability-identifier-naming.GlobalVariablePrefix
+  value:           'g_'
+# Disables the 'g_' prefix for static member variables
+- key:             readability-identifier-naming.ClassMemberPrefix
   value:           'm_'
 - key:             readability-identifier-naming.ClassConstantCase
   value:           aNy_CasE
--- a/src/BlockID.cpp
+++ b/src/BlockID.cpp
@@ -190,7 +190,7 @@ protected:
 
 
 bool cBlockIDMap::m_bHasRunInit = false;
-static cBlockIDMap gsBlockIDMap;
+static cBlockIDMap g_gsBlockIDMap;
 
 
 
@@ -224,11 +224,11 @@ int BlockStringToType(const AString & a_BlockTypeString)
 		return res;
 	}
 
-	if (!gsBlockIDMap.m_bHasRunInit)
+	if (!g_gsBlockIDMap.m_bHasRunInit)
 	{
-		gsBlockIDMap.init();
+		g_gsBlockIDMap.init();
 	}
-	return gsBlockIDMap.Resolve(TrimString(a_BlockTypeString));
+	return g_gsBlockIDMap.Resolve(TrimString(a_BlockTypeString));
 }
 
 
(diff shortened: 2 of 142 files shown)


src/BlockID.cpp
```diff
@@ -190,7 +190,7 @@ class cBlockIDMap
 
 
 bool cBlockIDMap::m_bHasRunInit = false;
-static cBlockIDMap gsBlockIDMap;
+static cBlockIDMap g_gsBlockIDMap;
```
peterbell10: This is why you can't just blindly pass `-fix` and not check the output.

madmaxoft:
Most of the violations are either in prehistoric code, or in Lua bindings, where it is somewhat customary to call the lua state parameter `L` (or `tolua_s`, as generated by ToLua++)

Note that your fixes still aren't perfect - the part after the prefix should be capitalized: `a_Tolua_s` (or even better, `a_LuaState`), `a_ArgP`, `a_Args`, `a_Status` ...

src/Blocks/BlockFence.h
```diff
@@ -17,8 +17,8 @@ class cBlockFenceHandler :
 	// These are the min and max coordinates (X and Z) for a straight fence.
 	// 0.4 and 0.6 are really just guesses, but they seem pretty good.
 	// (0.4 to 0.6 is a fence that's 0.2 wide, down the center of the block)
-	const double MIN_COORD = 0.4;
-	const double MAX_COORD = 0.6;
+	const double m_MIN_COORD = 0.4;
+	const double m_MAX_COORD = 0.6;
```
peterbell10: These should be static constants.

.clang-tidy
```diff
@@ -1,6 +1,13 @@
 Checks: '-*,readability-identifier-naming'
 CheckOptions:
-- key:             readability-identifier-naming.PrivateMemberPrefix
+- key:             readability-identifier-naming.MemberPrefix
+  value:           'm_'
```
peterbell10: This includes `struct` as well but shouldn't.

.clang-tidy
```diff
@@ -1,6 +1,13 @@
...
+- key:             readability-identifier-naming.ParameterPrefix
```
peterbell10: Add a case as well, they should be `CamelCase`.

src/BlockID.cpp
bibo38: Yeah, I thought, that I've run `grep g_g`...

bibo38: You've got lucky ;D. Just 2/142 files contained it...

bibo38:
> Note that your fixes still aren't perfect - the part after the prefix should be capitalized: a_Tolua_s (or even better, a_LuaState), a_ArgP, a_Args, a_Status ...

Yep, I know. But I wanted to keep the PR "minimalistic", so that it can be merged faster (vs a gigantic one, which implements all coding conventions at once, but takes 3 months to get merged).

Or do you want the casing also to be corrected in this PR?

.clang-tidy
bibo38: > All member variables start with m_, all function parameters start with a_, all class names start with c.

Do you agree @madmaxoft ? Then I would change the sentence to

> All class member variables start with m_, all function parameters start with a_, all class names start with c.

peterbell10:
It's not going to increase the line count if it's only changing the same variable names.

bibo38:
> It's not going to increase the line count if it's only changing the same variable names.

```bash
$ grep -R -E 'a_[a-z]' src/ | wc -l
25628
```

Or do you have a quick approach to filter these lines based on a given diff? Otherwise I would already need to review them manually when adding the case check. On the other hand, I should be able to catch most wrong things in this PR with regexes.

peterbell10:
I'm not sure I'd trust that regex... e.g. `lua_tostring` would match.

bibo38:
```bash
$ grep -R -E '[^a-zA-Z_]a_[a-z]' --exclude=Bindings.cpp src/ | wc -l
269
```
Should be possible in this PR :smile_cat: 

P.S.: Should I ignore the generated binding files for `clang-tidy` or fix the naming issues in tolua++?

peterbell10:
I think ignoring Bindings.cpp would be best, for now anyway.

12xx12:
Did the original idea work? should we revive this?

PR #4229, opened 2018-05-10
Changes to client's right clicking on a block

nii42:
Placing blocks next to a block that also counts as usable was not possible, e.g. for fences (usable with mob leashes). 

With first checking if the block is usable, the method ignored following actions like placing blocks next to it or using the item in the player's hand in case using the block didn't work.
The ClientHandle should now be able to find exactly one executable action (if possible).

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1423,8 +1423,6 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 	cWorld * World = m_Player->GetWorld();
 	cPluginManager * PlgMgr = cRoot::Get()->GetPluginManager();
 
-	bool Success = false;
-
 	bool SuccessBlockUsable = false;
 	bool SuccessPlaceable = false;
 	bool SuccessItemUsable = false;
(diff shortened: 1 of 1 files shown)


src/ClientHandle.cpp
```diff
@@ -1424,6 +1424,11 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 	cPluginManager * PlgMgr = cRoot::Get()->GetPluginManager();
 
 	bool Success = false;
```
NiLSPACE: I guess this variable can be removed entirely now right?

nii42: Actually true. I somehow thought that this variable is used as a return value.

peterbell10:
I'm not sure I like this, I think this will probably change other right clicking interactions in unexpected ways. For example I've found that I can now place blocks against buttons, but only if they're already pressed.

It seems to me like the real issue here is that `cBlockFence::IsUsable` doesn't have enough information to give the right answer.

nii42:
Adjusting the check in `cBlockFence::IsUsable` was my first idea aswell, but since this is a `cBlockHandler` method, I didn't want to change it and for every other derived BlockHandler aswell. Wanna get a bit more of a general overview on how the system works first (just started a few days ago).

Thanks for the hint with the button behavior, gonna look into that. Things like that are of course not supposed to happen.

peterbell10:
You could always add a new overload that defaults to calling the current overload
```cpp
bool cBlockHandler::IsUsable(cPlayer & a_Player)
{
    return IsUsable();
}
```

That way you would only need to change `cBlockFence` and the `cBlockHandler` base class.

PR #4235, opened 2018-06-04
Support cygwin compilation

cengizIO:
Hello! 

Although I know that only MSVC is supported on Windows, I still wanted
to try building with cygwin and fixed the issues along the way.

1) `M_PI` is not in <math.h> in cygwin. (Any suggestions for a better
place for our define?)
2) `strerror_r` is supported ONLY with _GNU_SOURCE flag.
3) gcc 7.1 in cygwin throws linker error with `stack-protector-all`.
I didn't want to modify any library variables so I've disabled stack
protector only for SQLiteCpp if we're building with cygwin.

After those, cygwin successfuly builds and everything seems to be fine
compilation-wise.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,7 +280,6 @@ endif()
 
 if (CYGWIN)
 	add_definitions(-D_GNU_SOURCE)
-	set(M_PI 3.14159265358979323846264338327950288) # This should be moved elsewhere
 	target_compile_options(SQLiteCpp PRIVATE -fno-stack-protector)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,6 +278,12 @@ if (WIN32)
 	add_subdirectory(lib/luaproxy/)
 endif()
 
+if (CYGWIN)
+	add_definitions(-D_GNU_SOURCE)
+	set(M_PI 3.14159265358979323846264338327950288) # This should be moved elsewhere
+	target_compile_options(SQLiteCpp PRIVATE -fno-stack-protector)
+endif()
+
 # We use EXCLUDE_FROM_ALL so that only the explicit dependencies are used
 # (mbedTLS also has test and example programs in their CMakeLists.txt, we don't want those)
 include(lib/mbedtls.cmake EXCLUDE_FROM_ALL)
--- a/src/OSSupport/Errors.cpp
+++ b/src/OSSupport/Errors.cpp
@@ -23,7 +23,7 @@ AString GetOSErrorString( int a_ErrNo)
 	// According to https://linux.die.net/man/3/strerror_r there are two versions of strerror_r():
 
 	// GNU version of strerror_r()
-	#if defined(__GLIBC__) && defined( _GNU_SOURCE) && !defined(ANDROID) \
+	#if (defined(__GLIBC__) && defined( _GNU_SOURCE) && !defined(ANDROID)) \
 		|| defined(__CYGWIN__)
 
 	char * res = strerror_r( errno, buffer, ARRAYCOUNT(buffer));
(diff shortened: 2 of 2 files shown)


src/OSSupport/Errors.cpp
```diff
@@ -22,7 +22,9 @@ AString GetOSErrorString( int a_ErrNo)
 
 	// According to https://linux.die.net/man/3/strerror_r there are two versions of strerror_r():
 
-	#if defined(__GLIBC__) && defined( _GNU_SOURCE) && !defined(ANDROID)  // GNU version of strerror_r()
+	// GNU version of strerror_r()
+	#if defined(__GLIBC__) && defined( _GNU_SOURCE) && !defined(ANDROID) \
+		|| defined(__CYGWIN__)
```
madmaxoft: These should be properly parenthesised to indicate the correct evaluation order.

CMakeLists.txt
```diff
@@ -278,6 +278,12 @@ if (WIN32)
 	add_subdirectory(lib/luaproxy/)
 endif()
 
+if (CYGWIN)
+	add_definitions(-D_GNU_SOURCE)
+	set(M_PI 3.14159265358979323846264338327950288) # This should be moved elsewhere
```
madmaxoft: According to this, the proper way is to define `_USE_MATH_DEFINES`:
https://stackoverflow.com/questions/1727881/how-to-use-the-pi-constant-in-c#1727896

cengizIO:
@madmaxoft please re-check

madmaxoft:
So does `M_PI` actually work without any further flags?

madmaxoft:
We could actually try Cygwin in Appveyor, it is preinstalled there, along with MinGW:
https://www.appveyor.com/docs/build-environment/#mingw-msys-cygwin

cengizIO:
@madmaxoft for its use in [cuberite/src/Matrix4.h](https://github.com/cuberite/cuberite/blob/master/src/Matrix4.h#L6), it's fine. However I'm not sure if the inclusion order of that header is significant for others or not. (Maybe tools?)

cengizIO:
@madmaxoft I'm confused with the `appveyor.yml` layout for multiple builds. should we create two separate files?

madmaxoft:
I'd say this is fine as it is, no need to change something that is working now.

As for Appveyor, I have no idea how that works, so some research is needed. Creating a new issue was the best course of action.

Let's get this PR merged as soon as it gets another positive review.

CMakeLists.txt
```diff
@@ -278,6 +278,11 @@ if (WIN32)
 	add_subdirectory(lib/luaproxy/)
 endif()
 
+if (CYGWIN)
+	add_definitions(-D_GNU_SOURCE)
+	target_compile_options(SQLiteCpp PRIVATE -fno-stack-protector)
```
peterbell10: This probably deserves a comment explaining why it's needed.

src/OSSupport/Errors.cpp
```diff
@@ -22,7 +22,9 @@ AString GetOSErrorString( int a_ErrNo)
 
 	// According to https://linux.die.net/man/3/strerror_r there are two versions of strerror_r():
 
-	#if defined(__GLIBC__) && defined( _GNU_SOURCE) && !defined(ANDROID)  // GNU version of strerror_r()
+	// GNU version of strerror_r()
+	#if (defined(__GLIBC__) && defined( _GNU_SOURCE) && !defined(ANDROID)) \
```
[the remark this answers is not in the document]
madmaxoft: Interesting, I thought it was used. If not, then the best course of action is of course to remove it altogether.

pwnorbitals:
Is this still considered ?

cengizIO:
@pwnorbitals I no longer have access to a Windows machine. This can be closed and ignored.

PR #4240, opened 2018-06-15
Flush stdout after resetting console colour.

### LINKED ISSUE #4238: In the linux, there is a little problem that changing text color
Client version:   
Server OS:        Windows/Linux
Commit id:        
Hello, it is not very serious problem. If i run cuberite in linux, my command text color is changed to previous message's color

### Expected behavior
image

### Actual behavior
image

peterbell10:
Fixes #4238

### DIFF
--- a/src/LoggerListeners.cpp
+++ b/src/LoggerListeners.cpp
@@ -145,6 +145,7 @@
 		{
 			// Whatever the console default is
 			printf("\x1b[0m");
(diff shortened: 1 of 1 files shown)


madmaxoft:
This does a double-flush for each logged line, which wouldn't be good if we logged many things often. Fortunately we don't, so this is satisfactory.

PR #4241, opened 2018-06-16
 support swap item between main hand and offhand by press key "F"

changyongGuo:
Hi,

I found cuberite don't support to press key "F" to swap items on main hand and offhand, so I implemented this feature.

Best regards
Changyong

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1188,7 +1188,14 @@ void cClientHandle::HandleLeftClick(int a_BlockX, int a_BlockY, int a_BlockZ, eB
 
 		case DIG_STATUS_SWAP_ITEM_IN_HAND:
 		{
-			// TODO: Not yet implemented
+
+			cItem EquippedItem = m_Player->GetEquippedItem();
+			cItem OffhandItem = m_Player->GetEquippedShield();
+
+			cInventory & Intentory = m_Player->GetInventory();
+			Intentory.SetShieldSlot(EquippedItem);
+			Intentory.SetHotbarSlot(Intentory.GetEquippedSlotNum(), OffhandItem);
+
 			return;
 		}
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -63,3 +63,4 @@ williamhatcher (Helped with API documentation updates and bug fixes)
 worktycho
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+changyongGuo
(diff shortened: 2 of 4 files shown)


changyongGuo:
Could someone help me to solve this error message : 

**ERROR: Newly undocumented API symbols found:**

I am confused, I had add API document in APIDesc.lua, what else file I should modify?

peterbell10:
You added a function to `cEntity` but your documentation is under `cInventory`.

changyongGuo:
@peterbell10  thanks, I updated API document.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -3317,6 +3317,16 @@ local Hash = cCryptoHash.sha1HexString("DataToHash")
 					},
 					Notes = "Returns the weapon that the entity has equipped. Returns an empty cItem if no weapon equipped or not applicable.",
 				},
+        GetEquippedShield =
```
NiLSPACE: You used spaces here instead of tabs which caused the indentation to be malformed here.

NiLSPACE:
The code seems good, but I'm not sure if the name is accurate. It's not only for shields right? Perhaps `GetOffHandEquipedItem`. Any native speakers who know a better name?

Server/Plugins/APIDump/APIDesc.lua
changyongGuo: sorry about this, I will adjust my IDE formater

changyongGuo:
code updated, please check again

src/ClientHandle.cpp
```diff
@@ -1188,7 +1188,14 @@ void cClientHandle::HandleLeftClick(int a_BlockX, int a_BlockY, int a_BlockZ, eB
 
 		case DIG_STATUS_SWAP_ITEM_IN_HAND:
 		{
-			// TODO: Not yet implemented
+
+			cItem EquippedItem = m_Player->GetEquippedItem();
+			cItem OffhandItem = m_Player->GetOffHandEquipedItem();
+
+			cInventory & Intentory = m_Player->GetInventory();
+			Intentory.SetShieldSlot(EquippedItem);
```
NiLSPACE: This still contains 'Shield'. Perhaps `SetOffHandSlot`?

NiLSPACE:
Looks great :) Just that one comment and it can be merged :)

src/ClientHandle.cpp
Bond-009: @NiLSPACE  That function is already exported to the lua API and we don't want to break compatibility.
You can change it if you want (without breaking compatibility) but I think that it is out of scope for this PR.

NiLSPACE: Oh, I see. Perhaps we can rename `SetShieldSlot` and create a manualbinding to the Lua API for backwards compatibility, but that's something for later on :)

Issue #4232, opened 2018-05-22
Some village houses generate with some walls facing the wrong way


Rijenkii:
Client version:   1.12.2
Server OS:        Linux
Commit id:        b9fdaf8a942ea2f79e48e3f735ba8916f3364e70

### Expected behavior
Houses should generate normally

### Actual behavior
This
![2018-05-22_15 45 57](https://user-images.githubusercontent.com/5338332/40351898-b351acfc-5dd7-11e8-99d9-0da3ac071413.png)
![2018-05-22_15 46 04](https://user-images.githubusercontent.com/5338332/40351899-b37cb7d0-5dd7-11e8-949f-554f17fa78ef.png)
![2018-05-22_15 46 08](https://user-images.githubusercontent.com/5338332/40351900-b3a68344-5dd7-11e8-9d7a-f333953642f4.png)

### Steps to reproduce the behavior
1. Generate a world with seed 512375971
2. Go to the village at the coordinates x=100, y=-140
3. Observe

madmaxoft:
Wow, that is weird. I've noticed something similar a few times, but never this large, it was just one or two bad blocks, I've suspected that some block's rotation was broken, but this seems like a completely different cause that would also explain what I was seeing all those years ago.
Thanks for the seed and coords, that's exactly what we need to fix that.

changyongGuo:
Hi,

After check with source code, I found there is a bug in blockArea roration, I will submit a patch to fix this.

PR #4242, opened 2018-06-23
fix block area rotation bug

changyongGuo:
fix mirror method bug in class cBlockArea.
this bug is caused by wrong block type switch during mirror block group.
this commit will fix issue #4232

before fix:
before

after fix:
after


madmaxoft:
Looks good, but cannot be merged until the dependent PR is merged and this one fixed.

changyongGuo:
sorry I don't have much experience about GIT, I will close this pull request and submit a new one that start from master branch

PR #4243, opened 2018-06-24
fix block area rotation bug

changyongGuo:
re-submit pull request because of wrong code base in last pull request #4242.

fix mirror method bug in class cBlockArea.
this bug is caused by wrong block type switch during mirror block group.
this commit will fix issue #4232

before fix:
41808993-b3c9f48c-7718-11e8-9fa7-3945c7a6d857

after fix:
41808995-b94b194a-7718-11e8-9f21-5b9f39221c63


changyongGuo:
Hi,

I have noticed one of Travis CI build failed because of time out, I don't think this is caused by this commit, do I need to trigger another Travis build?

madmaxoft:
Thanks for the fix

PR #4244, opened 2018-06-24
ignore build collision for block type is flower

changyongGuo:
fix bug that block can be put onto flower.

capture


changyongGuo:
oh...sorry, I thought it will have same behavoir as tall grass. I should check more information before submit.

I will close this pull request.

PR #4246, opened 2018-06-29
Adds optimized Vanilla-Explosion-Algorithm

lixfel:
Hi,
this is my first Pull Request :).
If I made something wrong, please tell me.

This is an optimized Version of the Vanilla explosion-algorithm. (But not perfect, an recursive algorithm without traces would be way faster)
My sources and a short description, what I've done, is in the comment in front of cChunkMap::DoExplosionAt(...)

If you want to know, how I generated my "pre-computed-values/angels" (Angle111, Angle112, Angle123), ask me, I will attach these generator here.

Additionally I changed the addition to the vector while spawning a tnt (+0.5 in every direction) from the calling functions to cWorld::SpawnPrimedTNT(), because this was missing, when a tnt was charged via a redstone component.


madmaxoft:
Sorry, but we cannot use this code - you've adapted Bukkit code, which is basically just decompiled original Minecraft code. This would endanger us legal-wise, our code needs to be written new from scratch.

https://forum.cuberite.org/thread-1584.html

lixfel:
I used majorly the information from Minecraft Wiki, only for one detail (the handling of air because of the theoratical optimised BlastResistance-value of 0.09) used I the bukkit source code. Sad.

Issue #4189, opened 2018-02-19
Quick slot selection resets after disconnect


denmark111:
Client version:   1.12.2
Server OS:        Windows 10
Commit id:        d3c1c626f569e5aa58085425924cca45927b6199

### Expected behavior
Server should remember user's quick slot selection

### Actual behavior
Quick slot selection resets after disconnect.   

### Steps to reproduce the behavior
1. Connect to the server, select any quick slot other then the first one.  
2. Disconnect from the server.  
3. Reconnect to the server and check first quick slot selected.

denmark111:
I thought about adding integer variable to cPlayer for saving last selected hotslot index   
but since cPlayer already has cInventory that includes hotbar data,   
I'll look for way to save it in JSON format using `void cInventory::SaveToJson(Json::Value & a_Value)`   
Feel free to correct me if I'm heading to the wrong direction

madmaxoft:
I don't think the `cInventory` class is the right place for saving the hotbar selection. At the very least, it loads and saves the JSON **array** of items, so there's no clean way of saving the additional info. I'd save the slot selection in `cPlayer:SaveToDisk()` instead.

denmark111:
Thanks for your advice `@madmaxoft`  
Sounds good. I'll just make one more space in `Json::Value root` array and save it there.  
So I'm working on this issue but it seems that I don't have any functions to get what slot user selected?  
I thought that `GetFirstUsedSlot(void)` in ItemGrid.cpp was, but that just returns last filled item in hotbar.   
If there is one, it would be great to know the name of it.  

BTW sorry if I sounded weird cause English is not my primary language

denmark111:
Hmm...  
by using `GetEquippedSlotNum(void)` and `SetEquippedSlotNum(int a_SlotNum)` functions,  
I was able to save and load the index what user has selected.  
I also verified that the value is correctly saved by printing out message to debug console.  
But the problems is that the game doesn't want to reflect the index loaded from save data.  
What could be the reason?

PR #4249, opened 2018-07-07
recover hotbar selected slot after reconnect

changyongGuo:
1, implement protocol message SendHeldItemChange
2, add save / load inventory equipped item slot in JSON
3, send held item slot message after player connect to server

this commit fix issue #4189


changyongGuo:
function argument code style updated, please check again

Issue #4139, opened 2018-01-09
TNT's damage players in creative mode


TC1061:
Client version:   Latest
Server OS:        Windows 7
Commit id:        2529a898351c4b7c6372570ba53a7e25678980a9

### Expected behavior
Bombs don't damage player in creative mode

### Actual behavior
Bombs damage player in creative mode

### Steps to reproduce the behavior
1. Switch to creative mode (/gm 1)
2. Place a TNT, then stand close with it.
3. Explode the TNT.

mathiascode:
I can't reproduce this issue.

TC1061:
I have all default plugins and some my, but they do not affect this. After producing this behavior I wrote a plugin-fixer for this issue and then bombs doesn't damage players in creative mode at all.

bearbin:
Can you send a gif/video of this? I can't really understand what you mean by damage. Are players knocked back, or is the hurt sound effect played, ...? Creative players have no hearts on the client so the normal sense of being "damaged" doesn't make sense.

Bond-009:
I can't reproduce this issue either.

peterbell10:
I'm not sure if it's what @TC1061 is referring to but TNT does incorrectly apply knockback to players in creative mode.

TC1061:
When tnt is exploded, in creative mode hurt sound effect is played and player can even die.

PR #4251, opened 2018-07-08
rewrite explosion knock back implement

changyongGuo:
1, implement method to calculate entity explosion exposure rate
2, rewrite explosion impact based on wiki, add armor blast protection for impact
3, player will not be effected by explosion when flying in creative mode

this commit could solve player can be knock back bug in issue #4139

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1814,10 +1814,21 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 				a_Entity.TakeDamage(dtExplosion, nullptr, static_cast<int>((1 / DistanceFromExplosion.Length()) * 6 * ExplosionSizeInt), 0);
 			}
 
-			// Apply force to entities around the explosion - code modified from World.cpp DoExplosionAt()
-			DistanceFromExplosion.Normalize();
-			DistanceFromExplosion *= ExplosionSizeInt * ExplosionSizeInt;
-			a_Entity.AddSpeed(DistanceFromExplosion);
+			double Length = DistanceFromExplosion.Length();
+			if (Length <= ExplosionSizeInt)  // Entity is impacted by explosion
+			{
+				float EntityExposure = a_Entity.GetExplosionExposureRate(ExplosionPos, ExplosionSizeInt);
+
+				// Exposure reduced by armor
+				EntityExposure = EntityExposure * (1.0 - a_Entity.GetEnchantmentBlastKnockbackReduce());
+
+				double Impact = (1 - (Length / ExplosionSizeInt / 2)) * EntityExposure;
+
+				DistanceFromExplosion.Normalize();
+				DistanceFromExplosion *= Impact;
+
+				a_Entity.AddSpeed(DistanceFromExplosion);
+			}
 
 			return false;
 		}
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -690,6 +690,33 @@ int cEntity::GetEnchantmentCoverAgainst(const cEntity * a_Attacker, eDamageType 
 
 
 
+
+float cEntity::GetEnchantmentBlastKnockbackReduce()
+{
+	uint8_t MaxLevel = 0;
+
+	const cItem ArmorItems[] = { GetEquippedHelmet(), GetEquippedChestplate(), GetEquippedLeggings(), GetEquippedBoots() };
+
+	for (auto & Item : ArmorItems)
+	{
+		uint8_t Level = Item.m_Enchantments.GetLevel(cEnchantments::enchBlastProtection);
+		if (Level > MaxLevel)
+		{
+			// Get max blast protection
+			MaxLevel = Level;
+		}
+	}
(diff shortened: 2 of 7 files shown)


src/Entities/Player.cpp
```diff
@@ -2999,3 +2999,43 @@ float cPlayer::GetPlayerRelativeBlockHardness(BLOCKTYPE a_Block)
...
+		else
+		{
+			// Flying player don't impact from explosion
+			return 0;
+		}
+	}
+	else
+	{
+		return super::GetExplosionExposureRate(a_ExplosionPosition, a_ExlosionPower);
+	}
```
peterbell10: Isn't this a little over complicated, why not just have one if?
```cpp
if (
    IsGameModeSpectator() ||
    (IsGameModeCreative() && !IsOnGround())
)
{
    return 0;  // No impact from explosion
}

return super::GetExplosionExposureRate(a_ExplosionPosition, a_ExlosionPower);
```

src/Entities/Entity.h
```diff
@@ -333,6 +333,10 @@ class cEntity
 	/** Returns the hitpoints that the currently equipped armor's enchantments would cover */
 	virtual int GetEnchantmentCoverAgainst(const cEntity * a_Attacker, eDamageType a_DamageType, int a_Damage);
 
+	/** Returns explosion knock back reduce percent from blast protection level
+	@return : knock back reduce percent */
+	virtual float GetEnchantmentBlastKnockbackReduce();
```
peterbell10: 'Reduction' would be more grammatically correct here.

src/Entities/Entity.cpp
```diff
@@ -690,6 +690,33 @@ int cEntity::GetEnchantmentCoverAgainst(const cEntity * a_Attacker, eDamageType
...
+
+float cEntity::GetEnchantmentBlastKnockbackReduce()
+{
+	uint32_t MaxLevel = 0;
+
+	const cItem ArmorItems[] = { GetEquippedHelmet(), GetEquippedChestplate(), GetEquippedLeggings(), GetEquippedBoots() };
+
+	for (auto & Item : ArmorItems)
+	{
+		uint32_t Level = Item.m_Enchantments.GetLevel(cEnchantments::enchBlastProtection);
```
peterbell10: Style: You can either use `unsigned` or `UInt32` but the `_t` versions shouldn't be used.

src/Entities/Entity.h
```diff
@@ -545,6 +549,12 @@ class cEntity
 	/** Returs whether the entity has any mob leashed to */
 	bool HasAnyMobLeashed() const { return m_LeashedMobs.size() > 0; }
 
+	/** a lightweight calculation approach to get explosion exposure rate
+	@param a_ExplosionPosition : explosion position
+	@param a_ExlosionPower : explosion power
+	@return : exposure rate */
```
peterbell10: You shouldn't put colon's in the doxy-comments, it will look wrong in the generated docs. i.e. something like
### Returns
        : exposure rate

src/ChunkMap.cpp
```diff
@@ -1814,10 +1814,21 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
...
-			a_Entity.AddSpeed(DistanceFromExplosion);
+			double Length = DistanceFromExplosion.Length();
+			if (Length <= ExplosionSizeInt)  // Entity is impacted by explosion
+			{
+				float EntityExposure = a_Entity.GetExplosionExposureRate(ExplosionPos, ExplosionSizeInt);
+
+				// Exposure reduced by armor
+				EntityExposure = static_cast<float>(EntityExposure * (1.0 - a_Entity.GetEnchantmentBlastKnockbackReduce()));
+
+				double Impact = (1 - (Length / ExplosionSizeInt / 2)) * EntityExposure;
```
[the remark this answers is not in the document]
changyongGuo: actually, about this impact calculation, what is describled in wiki is kind confuse.
https://minecraft.gamepedia.com/Explosion

```
For every entity within a 2×[power] block sphere of the explosion center, the impact is (1-[distance from explosion/power/2])×[exposure] (see section below on exposure)
```

I am not sure about how impact should calculate, 
* 1 - (distance / (power / 2))
* 1 - ((distance / power) / 2)

I prefer first one, because it is more make sense and impact strength will be more smooth at the edge of explosion effect box.

Am I correct?

src/Entities/Entity.h
```diff
@@ -333,6 +333,10 @@ class cEntity
 	/** Returns the hitpoints that the currently equipped armor's enchantments would cover */
 	virtual int GetEnchantmentCoverAgainst(const cEntity * a_Attacker, eDamageType a_DamageType, int a_Damage);
 
+	/** Returns explosion knock back reduction percent from blast protection level
+	@return : knock back reduce percent */
```
peterbell10: You left one last colon

changyongGuo: updated

madmaxoft:
CI failed

changyongGuo:
submited new commit to fix compile warning

PR #4252, opened 2018-07-09
cBlockArea: Write all present data types by default

peterbell10:
Fixes cuberite/WorldEdit#130
This is essentially @madmaxoft's solution:
>Basically, the universal fix is to write something like:
>```lua
>area:Write(..., area:GetDataTypes())
>```
>But that results in boilerplate code that no-one ever touches :(

But this way the boilerplate is gone and instead the behaviour is part of the API.

### DIFF
--- a/Server/Plugins/APIDump/Classes/BlockArea.lua
+++ b/Server/Plugins/APIDump/Classes/BlockArea.lua
@@ -1898,7 +1898,7 @@ return
 							Type = "boolean",
 						},
 					},
-					Notes = "Writes the area into World at the specified coords, returns true if successful. baTypes and baMetas are written.",
+					Notes = "Writes the area into World at the specified coords, returns true if successful. All present data types are written.",
 				},
 				{
 					Params =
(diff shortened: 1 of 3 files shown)


madmaxoft:
I don't think this solves the issue, because it doesn't change the Lua bindings at all - they still provide the (bad) defaults:
https://github.com/cuberite/cuberite/blob/06eb5cc70da76f6a95afa9c9a3d692ca39e39a76/src/Bindings/ManualBindings_BlockArea.cpp#L703

Doesn't C++ allow the default parameter to be a function call?
```cpp
bool Write(cForEachChunkProvider & a_ForEachChunkProvider, int a_MinBlockX, int a_MinBlockY, int a_MinBlockZ, int a_DataTypes = GetDataTypes());
```
on the C++ side, then equivalent changes on the Lua bindings side.

peterbell10:
> it doesn't change the Lua bindings at all

Thanks for spotting that, I hadn't realised it was bound manually.  On a side note, isn't this usually indicated in the header, e.g.
https://github.com/cuberite/cuberite/blob/06eb5cc70da76f6a95afa9c9a3d692ca39e39a76/src/World.h#L159



>Doesn't C++ allow the default parameter to be a function call?

I think the problem is that it's a member function, I get errors if I try
```shell
../src/Blocks/../BlockArea.h:123:143: error: cannot call member function ‘int cBlockArea::GetDataTypes() const’ without object   
```

madmaxoft:
Usually it is marked, but not everywhere, and since most if not all `cBlockArea` functions are exported, it hasn't been done here. Consistency, I know...

Issue #4250, opened 2018-07-07
I can't place rails on hoppers


tomStegmaier:
Client version:   12.2.2
Server OS:        Linux
Commit id:        06eb5cc70da76f6a95afa9c9a3d692ca39e39a76

### Expected behavior
the rail stays at the hopper

### Actual behavior
it disappears after 0.2 seconds and is again in my inventory

### Steps to reproduce the behavior
put a hopper at the map and try to put a rail on top of it

changyongGuo:
Hi,

This issue is caused by error block handle in BlockRail.h, I will create a pull request for this.

Best regards
Changyong

PR #4253, opened 2018-07-12
Made Hopper a m_FullyOccupiesVoxel block pursuant to #4250

csunday95:
Hello all, first time contributing to this project. I noticed issue #4250 and thought it might be a good place to start digging through the source code and possibly fix the issue. User @changyongGuo commented that they would create a pull request, but I hadn't seen it added yet. As for the proposed fix, I checked the vanilla minecraft behavior, and you can in fact place surface mountable items like switches on hoppers in addition to rail. I think further checking against vanilla behavior might be necessary for more obscure interactions of items with the m_FullyOccupiesVoxel field, such as for breaking ice blocks(?), but as far as I can tell placing a rail on a hopper isn't really a special case.

### DIFF
--- a/src/Blocks/BlockButton.h
+++ b/src/Blocks/BlockButton.h
@@ -71,6 +71,15 @@ public:
 		BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta
 	) override
 	{
+		BLOCKTYPE Block;
+		NIBBLETYPE Meta;	
+		AddFaceDirection(a_BlockX, a_BlockY, a_BlockZ, a_BlockFace, true);  // Set to clicked block
+		a_ChunkInterface.GetBlockTypeMeta({a_BlockX, a_BlockY, a_BlockZ}, Block, Meta);
+		
+		if(!CanBePlacedOn(Block, Meta, a_BlockFace))
+		{
+			return false;
+		}
 		a_BlockType = m_BlockType;
 		a_BlockMeta = BlockFaceToMetaData(a_BlockFace);
 		return true;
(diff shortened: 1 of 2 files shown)


csunday95:
Here's a list of all the interactions with m_FullyOccupiesVoxel that I'll need to check against vanilla

 - [ ] Can buttons be placed on hopppers on any side?
 - [ ] Does farmland reverts to dirt when a hopper is placed adjacent?
 - [ ] Does ice turn into water when broken?
 - [ ] Can levers be placed on any side?
 - [x] Can rail be placed atop a hopper?
 - [ ] Can redstone blocks be placed atop a hopper?
 - [ ] Can redstone repeaters be placed atop a hopper?
 - [ ] Can torches/ redstone torches be placed on any side except bottom?
 - [ ] Can tripwire hooks be placed?
 - [ ] Do hoppers block lilypads from being placed under them?

src/Blocks/BlockButton.h
```diff
@@ -71,6 +71,15 @@ class cBlockButtonHandler :
 		BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta
 	) override
 	{
+		BLOCKTYPE Block;
```
csunday95: buttons require same style of compatibility checks within GetPlacementBlockTypeMeta as for torches; see [here](https://github.com/csunday95/cuberite/blob/2ccaa5eb9f41e166d18d24736b9242839f48bc12/src/Blocks/BlockTorch.h#L27)

src/Blocks/BlockButton.h
```diff
@@ -113,15 +122,36 @@ class cBlockButtonHandler :
 		}
 	}
 
+	static bool CanBePlacedOn(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, eBlockFace a_BlockFace)
```
csunday95: Since button place-ability behavior is specific to the target block, we need a function akin to what's seen in [BlockTorch.h](https://github.com/csunday95/cuberite/blob/2ccaa5eb9f41e166d18d24736b9242839f48bc12/src/Blocks/BlockTorch.h#L88)

src/Blocks/BlockButton.h
```diff
@@ -113,15 +122,36 @@ class cBlockButtonHandler :
...
+	{
+		switch(a_BlockType)
+		{
+			case E_BLOCK_HOPPER:
```
csunday95: For example, vanilla behavior allows buttons to be placed only on top of hoppers, not on other sides

changyongGuo:
Hi csunday95,

welcome!
Sorry it takes me lot of time to prepare pull request for block placement issue. Currently there are lots of overlap between what you prepare to submit and what's in my pull request #4256 .

I hope we could discuss a clear workscope about how to  fix block placement issue, so we don't waste time on duplicate things..

Best regards
Changyong

csunday95:
Hello changyongGuo,

I was thinking a little about a more extensible solution to the issue, and the problem really comes down to specific blocks having different allowed item placements. Right now torches have a pretty well developed system for determining if it may be placed or not through the CanBePlacedOn() method, so generalizing that system by adding a function to the cBlockHandler base class seems appropriate to me. Each block header can then override it with special behavior for instance with the current BlochTorch.h implementation. A check for 'placeability' also needs to take place within GetPlacementBlockTypeMeta(), or at least before placement is tentatively allowed, otherwise we get the non-vanilla behavior: a button may be placed on the side of a hopper, but on the next tick it is removed and converted to an item because CanBeAt() returns false. 

As for workscope, I could take a look at the branch you're working on and see if there's improvements I could do in parallel. Otherwise it'd probably be easiest for me to wait for your pull request and then work on top of that, since there is likely to be a large amount of overlap. I can keep experimenting with a few different ideas on my fork to see what works well/ cleanly.

PR #4258, opened 2018-07-20
cIsThread: Reset m_ShouldTerminate after the thread has stopped

### LINKED ISSUE #4257: restart command prevents hostname lookups
Following #3766 DNS lookups are offloaded to a home-grown queue which is owned by the network singleton. This causes state preservation across restarts (cIsThread::m_ShouldTerminate remains true) and the authenticator thread to block indefinitely.

Best is to probably explicitly reset the variable in cIsThread::Start (and remove all unnecessary sets in other threads, e.g. cAuthenticator::Start), or alternatively, override the Start function.

peterbell10:
This allows threads to be restarted after stopping.

Fixes #4257

### DIFF
--- a/src/OSSupport/IsThread.cpp
+++ b/src/OSSupport/IsThread.cpp
@@ -60,8 +60,7 @@ cIsThread::cIsThread(const AString & a_ThreadName) :
 
 cIsThread::~cIsThread()
 {
-	m_ShouldTerminate = true;
-	Wait();
+	Stop();
 }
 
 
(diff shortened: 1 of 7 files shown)


src/OSSupport/IsThread.cpp
```diff
@@ -110,14 +109,9 @@ bool cIsThread::Start(void)
 
 void cIsThread::Stop(void)
 {
-	if (!m_Thread.joinable())
-	{
-		// The thread hasn't been started or has already been joined
-		return;
-	}
```
peterbell10: I've also removed this because it's redundant, `Wait` already checks `m_Thread.joinable()`.

Issue #4216, opened 2018-04-17
XP Orb doesn't do what it appears to do.


leeadama:
Client version:   1.12.2
Server OS:        Android
Commit id:       98d807187bbaec4cd65dbe7c1ecac0435efb6df0

### Expected behavior
XP orb should fly to Player and be picked up.

### Actual behavior
XP orb appears to fly to Player but isn't picked up.  It just kind of fly's around the player.

In reality it is actually sitting on the ground near the spot where the monster died.  It isn't picked up until I walk forward over that spot.  You can hear it "Ping" even though you can't see it.

After you walk forward and pick it up the visible orb disappears.

Occasionally I can't find where the Orb actually is and the "Flying" orb will stay with me for 5 minutes or so until something times out and it goes away.

### Steps to reproduce the behavior
Kill something from a slight distance, through a doorway, etc.  Observe the behavior of the XP Orb.  Walk toward where the monster was.  Hear the XP Orb be picked up only after you move over the right spot.

You can see it here : https://youtu.be/O-sVAOsDNHw

tigerw:
Unfortunately the fancy physics of XP orbs is completely unimplemented serverside, I think what you see is a client prediction of what should be happening. We'll get on to it, eventually...

PR #4259, opened 2018-07-22
Experience orb

changyongGuo:
1, implement closest player detection feature in cWorld::DoWithClosestPlayer
2, replace original cWorld::FindClosestPlayer
3, implement split experience reward into small orbs feature
4, modify speed calculation in cExpOrb::Tick to simulate orbs 'fly toward' player effect

this pull request solve issue #4216


src/Blocks/WorldInterface.h
```diff
@@ -48,9 +48,8 @@ class cWorldInterface
 	Returns the mob's UniqueID if recognized and spawned, or cEntity::INVALID_ID on failure. */
 	virtual UInt32 SpawnMob(double a_PosX, double a_PosY, double a_PosZ, eMonsterType a_MonsterType, bool a_Baby) = 0;
 
-	/** Spawns an experience orb at the given location with the given reward.
-	Returns the UniqueID of the spawned experience orb, or cEntity::INVALID_ID on failure. */
-	virtual UInt32 SpawnExperienceOrb(double a_X, double a_Y, double a_Z, int a_Reward) = 0;
+	/** Spawns an experience orb at the given location with the given reward, by default split into multiple experience orbs if possible */
+	virtual void SpawnExperienceOrb(double a_X, double a_Y, double a_Z, int a_Reward, bool a_Split = true) = 0;
```
NiLSPACE: Any chance you could make it return a list of entity id's? That would allow plugins to influence newly spawned experience orbs.

changyongGuo:
code updated, add new method for splite experience orb, and remove compile warning for clang

src/World.cpp
```diff
@@ -3289,40 +3339,70 @@ bool cWorld::DoWithPlayerByUUID(const cUUID & a_PlayerUUID, cPlayerListCallback
...
+	});
+
+	if (Player == nullptr)
+	{
+		return false;  // Player not found
+	}
+	else
+	{
+		a_Callback(*Player);
+		return true;
```
peterbell10: I'm sorry but please put the actual player searching code back to the way it was.  The only change needed to turn this from `FindClosestPlayer` to `DoWithClosestPlayer` was calling the callback at the end of the function.

I'm not a fan of the other changes:
 * `ForEachEntityInBox` has to do more work than looping over players because there are more entities than players
 * the actual player selection code has become more complicated for no apparent reason
 * and most importantly, you've missed the entire point of a `DoWith` function.  It's supposed to call the callback under a lock which isn't the case here because the lock has move into `ForEachEntityInBox`.

src/Entities/ExpOrb.h
```diff
@@ -44,9 +44,14 @@ class cExpOrb :
 
 	// tolua_end
 
+	/** Split reward into small values */
+	static void Splite(int a_Reward, std::vector<int> & a_SplitReward);
```
peterbell10: Spelling

src/World.h
```diff
@@ -464,6 +464,10 @@ class cWorld :
 	Returns the UniqueID of the spawned experience orb, or cEntity::INVALID_ID on failure. */
 	virtual UInt32 SpawnExperienceOrb(double a_X, double a_Y, double a_Z, int a_Reward) override;
 
+	/** Spawns an experience orb at the given location with the given reward, by default split into multiple experience orbs if possible,
+	Returns spawned experience orbs UniqueID */
+	virtual std::vector<UInt32> SpawnSplitExperienceOrbs(double a_X, double a_Y, double a_Z, int a_Reward, bool a_Split = true);
```
peterbell10: Remove the `bool`, this function should always split because it's part of the name.

src/World.cpp
```diff
@@ -2273,19 +2273,69 @@ UInt32 cWorld::SpawnFallingBlock(int a_X, int a_Y, int a_Z, BLOCKTYPE BlockType,
...
+	}
+
+	SpeedLimit *= 100;
+	auto & Random = GetRandomProvider();
+
+	for (auto Element : Orbs)
+	{
+		float SpeedX = Random.RandInt(-SpeedLimit, SpeedLimit) / 100.0f;
+		float SpeedY = Random.RandInt(50) / 100.0f;
+		float SpeedZ = Random.RandInt(-SpeedLimit, SpeedLimit) / 100.0f;
```
peterbell10: Is there a reason you're using `RandInt` and converting to float instead of just `RandReal`?

src/Entities/ExpOrb.h
```diff
@@ -18,8 +18,8 @@ class cExpOrb :
 
 	CLASS_PROTODEF(cExpOrb)
 
-	cExpOrb(double a_X, double a_Y, double a_Z, int a_Reward);
-	cExpOrb(const Vector3d & a_Pos, int a_Reward);
+	cExpOrb(double a_X, double a_Y, double a_Z, int a_Reward, float a_SpeedX = 0.f, float a_SpeedY = 0.f, float a_SpeedZ = 0.f);
+	cExpOrb(const Vector3d & a_Pos, int a_Reward, float a_SpeedX = 0.f, float a_SpeedY = 0.f, float a_SpeedZ = 0.f);
```
peterbell10: Is this change really adding value? Why not just set the speed after construction, that's how it's done for TNT already:
https://github.com/cuberite/cuberite/blob/98d807187bbaec4cd65dbe7c1ecac0435efb6df0/src/World.cpp#L2337-L2353

src/World.cpp
changyongGuo: alright, I agree what you said, I will change based on orignal method.

src/World.cpp
changyongGuo: sorry I searched and didn't find this method exist

src/Entities/ExpOrb.h
changyongGuo: experience orbs need to created with speed if multiple orbs are generated.

There are similar operation in method **cWorld::SpawnItemPickups**

changyongGuo: I agree I can set speed after create entity, but I think these are just two different methods to implement same feature, is it neccessary to create new commit to modify it ?

peterbell10: Well, the interface would need to be changed either way because it isn't using `Vector3d` (see #3853).

changyongGuo: code updated for set speed after orb created

changyongGuo:
need help on generated unused function in bindings.cpp
```
[ 75%] Building CXX object Tools/ProtoProxy/CMakeFiles/ProtoProxy.dir/__/__/src/StringUtils.cpp.o
/Users/travis/build/cuberite/cuberite/src/Bindings/Bindings.cpp:218:12: error: unused function 'tolua_collect_std__vector_UInt32_' [-Werror,-Wunused-function]
static int tolua_collect_std__vector_UInt32_ (lua_State* tolua_S)
```

how could I remove this warning?

changyongGuo:
disabled lua bind for method SpawnSplitExperienceOrbs, sorry I don't know how to manually bind method to lua interface

src/Entities/ExpOrb.h
```diff
@@ -44,9 +44,14 @@ class cExpOrb :
 
 	// tolua_end
 
+	/** Split reward into small values */
+	static void Split(int a_Reward, std::vector<int> & a_SplitReward);
```
madmaxoft: I think a better signature would be to return the `std::vector` directly, rather than using it as an out-param:
```cpp
static std::vector<int> & Split(int a_Reward);
```

src/Entities/ExpOrb.h
```diff
@@ -44,9 +44,14 @@ class cExpOrb :
...
+
 protected:
 	int m_Reward;
 
 	/** The number of ticks that the entity has existed / timer between collect and destroy; in msec */
 	std::chrono::milliseconds m_Timer;
+
+	bool MoveToPlayer;
```
madmaxoft: Member variables should have the `m_` prefix.
Additionally, the name should indicate `bool`-ness (`MoveToPlayer` is a command, `ShouldMoveToPlayer` or `IsMovingToPlayer` is a property)

src/Entities/ExpOrb.cpp
```diff
@@ -47,33 +50,74 @@ void cExpOrb::SpawnOn(cClientHandle & a_Client)
...
 		{
-			Vector3f a_PlayerPos(a_ClosestPlayer->GetPosition());
-			a_PlayerPos.y++;
-			Vector3f a_Distance(a_PlayerPos - GetPosition());
-			double Distance(a_Distance.Length());
-			if (Distance < 0.5f)
-			{
-				LOGD("Player %s picked up an ExpOrb. His reward is %i", a_ClosestPlayer->GetName().c_str(), m_Reward);
-				a_ClosestPlayer->DeltaExperience(m_Reward);
+			LOGD("Player %s picked up an ExpOrb. His reward is %i", a_Player.GetName().c_str(), m_Reward);
```
madmaxoft: This should be removed in the final code.

src/Entities/ExpOrb.cpp
```diff
@@ -47,33 +50,74 @@ void cExpOrb::SpawnOn(cClientHandle & a_Client)
...
+	// Find closest player within 6.5 meter (slightly increase detect range to have same effect in client)
+	m_World->DoWithClosestPlayer(GetPosition(), 6.5, [&](cPlayer & a_Player) -> bool
 	{
-		cPlayer * a_ClosestPlayer(m_World->FindClosestPlayer(Vector3f(GetPosition()), 5, false));
-		if ((a_ClosestPlayer != nullptr) && (!a_ClosestPlayer->IsGameModeSpectator()))
+		FindPlayer = true;
+		Vector3f a_PlayerPos(a_Player.GetPosition());
+		a_PlayerPos.y += 0.8f;
+		Vector3f a_Distance(a_PlayerPos - GetPosition());
+		double Distance(a_Distance.Length());
```
madmaxoft: We prefer the `auto var = initializer` style.

src/World.h
```diff
@@ -279,8 +279,8 @@ class cWorld :
 	/** Finds a player from a partial or complete player name and calls the callback - case-insensitive */
 	bool FindAndDoWithPlayer(const AString & a_PlayerNameHint, cPlayerListCallback a_Callback);  // >> EXPORTED IN MANUALBINDINGS <<
 
-	// TODO: This interface is dangerous - rewrite to DoWithClosestPlayer(pos, sight, action)
-	cPlayer * FindClosestPlayer(Vector3d a_Pos, float a_SightLimit, bool a_CheckLineOfSight = true);
+	/** Calls the callback for closest player for given position, return true if player was found and callback called, false if player not found. */
+	bool DoWithClosestPlayer(Vector3d a_Pos, float a_SightLimit, cPlayerListCallback a_Callback, bool a_CheckLineOfSight = true, bool a_IgnoreSpectator = true);
```
madmaxoft: I think `DoWithNearestPlayer` is a better name.

madmaxoft: Also, if the player is found, this function should return whatever value the callback returned (all our callback-based `DoWith` functions work this way)

madmaxoft: `a_SightLimit` should probably be renamed to `a_RangeLimit`, because it wouldn't make sense if `a_CheckLineOfSight` was `false`.

src/Entities/ExpOrb.cpp
```diff
@@ -47,33 +50,74 @@ void cExpOrb::SpawnOn(cClientHandle & a_Client)
 void cExpOrb::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	DetectCacti();
+	m_TicksAlive++;
 
-	// Check player proximity no more than twice per second
-	if ((m_TicksAlive % 10) == 0)
+	bool FindPlayer = false;
```
madmaxoft: Not needed, check the return value from `DoWithNearestPlayer`

src/Entities/ExpOrb.cpp
```diff
@@ -47,33 +50,74 @@ void cExpOrb::SpawnOn(cClientHandle & a_Client)
...
+		if (MoveToPlayer)  // Already flying to player
+		{
+			Vector3d SpeedDelta(a_Distance);
+			SpeedDelta.Normalize();
+			SpeedDelta *= 3;
+
+			Vector3d CurrentSpeed = GetSpeed();
+
+			CurrentSpeed += SpeedDelta;
+
```
madmaxoft: Too many empty lines with no effect. Only use empty lines for blocks of code for which it makes sense to add a comment:
```cpp
PreviousCodeBlock;

// What the next block does:
CodeBlock;
CodeBlockContinued;
MoreCode;
...
```

src/Entities/ExpOrb.cpp
```diff
@@ -47,33 +50,74 @@ void cExpOrb::SpawnOn(cClientHandle & a_Client)
...
-				m_World->BroadcastSoundEffect("entity.experience_orb.pickup", GetPosition(), 0.5f, (0.75f + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
+			m_World->BroadcastSoundEffect("entity.experience_orb.pickup", GetPosition(), 0.5f, (0.75f + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
 
-				Destroy(true);
-				return;
+			Destroy(true);
+			return true;
+		}
+
+		/** Experience orb will "float" or glide toward the player up to a distance of 6 blocks,
```
madmaxoft: Don't use doxy-comments in place of regular comments

src/Entities/ExpOrb.cpp
```diff
@@ -47,33 +50,74 @@ void cExpOrb::SpawnOn(cClientHandle & a_Client)
...
+		speeding up as they get nearer to the player, Speed range 6 - 12 m per second, accelerate 60 m per second^2 */
+		if (MoveToPlayer)  // Already flying to player
```
madmaxoft: Why do we even need this `if`? The results shouldn't be too different if applied on the first tick as well when the player is detected.

madmaxoft:
Here's the Lua binding code for the function, put it in src/Bindings/ManualBindings_World.cpp (it still needs registration in `cManualBindings::BindWorld` at the bottom of that file, you can do that)

```cpp
static int tolua_cWorld_DoWithNearestPlayer(lua_State * tolua_S)
{
	// Check params:
	cLuaState L(tolua_S);
	if (
		!L.CheckParamSelf("cWorld") ||
		!L.CheckParamNumber(2, 5) ||
		!L.CheckParamFunction(6) ||
		// Params 7 and 8 are optional bools, no check for those
		!L.CheckParamEnd(9)
	)
	{
		return 0;
	}

	// Get parameters:
	cWorld * Self;
	double PosX, PosY, PosZ, RangeLimit;
	cLuaState::cRef FnRef;
	bool CheckLineOfSight = true, IgnoreSpectators = true;  // Defaults for the optional params
	L.GetStackValues(1, Self, PosX, PosY, PosZ, RangeLimit, FnRef, CheckLineOfSight, IgnoreSpectators);

	if (!FnRef.IsValid())
	{
		return L.ApiParamError("Expected a valid callback function for parameter #5");
	}

	// Call the function:
	bool res = Self->DoWithNearestPlayer(Vector3d(PosX, PosY, PosZ), RangeLimit, [&](cPlayer & a_Player)
		{
			bool ret = false;
			L.Call(FnRef, &a_Player, cLuaState::Return, ret);
			return ret;
		}
	);

	// Push the result as the return value:
	L.Push(res);
	return 1;
}
```

src/Entities/ExpOrb.h
madmaxoft: This member may be unneeded after all, see later comments.

madmaxoft:
The `cWorld` class has a separate ManualBindings file, because there's a lot of bound functions. You should move the newly added bindings to that file as well: src/Bindings/ManualBindings_World.cpp

src/World.cpp
```diff
@@ -3309,7 +3360,15 @@ cPlayer * cWorld::FindClosestPlayer(Vector3d a_Pos, float a_SightLimit, bool a_C
 		ClosestDistance = Distance;
 		ClosestPlayer = *itr;
 	}
-	return ClosestPlayer;
+
+	if (ClosestPlayer)
+	{
+		return a_Callback(*ClosestPlayer);
```
peterbell10: Should this return `!a_Callback(ClosestPlayer)` instead?  There are existing `DoWith` functions that use both conventions:

https://github.com/cuberite/cuberite/blob/c94d7184ebaf7e8540f717c70c1e03ae62e5a7bd/src/Chunk.cpp#L2279
https://github.com/cuberite/cuberite/blob/c94d7184ebaf7e8540f717c70c1e03ae62e5a7bd/src/World.cpp#L2807
... and a third for good measure...
https://github.com/cuberite/cuberite/blob/c94d7184ebaf7e8540f717c70c1e03ae62e5a7bd/src/World.cpp#L2753-L2754

changyongGuo: In my understanding,
callback in *ForEach..* method return false means continue, return true to abort.
But in *DoWith..* method, there is no such rule, so what the return value means depend on each method. 
some *DoWith..* don't care about callback return value, return true means callback had being called.

In method *DoWithNearestPlayer*, method return what callback returns, I think it is make sense.

peterbell10: I have no doubt that in isolation all of these were justified but when taken together it makes the interface to `cWorld` inconsistent.  As far as I can tell, there are a lot more `DoWith` functions that try to behave like a single element `ForEach` function. So, for consistency, it might be better to go that way. 

I think it's at least worth discussing.

changyongGuo:
ah...sorry I need more support about lua API check.

1. how could I run API document check locally? I tried to compile with `SELF_TEST=ON` and run `ctest -C Debug`, but there is no lua API check.

2. why CI report error after method have been manually ported in `src/Bindings/ManualBindings_World.cpp`:
```
cWorld.DoWithNearestPlayer (function)
```

src/World.cpp
changyongGuo: Hi @peterbell10 

I agree we should refactor `ForEach` and `DoWith` method in `cWorld` (maybe in another pull request), there are lots of interface return style and comment inconsistent.

After read source code , here is my proposal:
1. for every `ForEach` method , 
    * callback return **true** to abort current for each loop, return **false** to continue loop.
    * 'ForEach' method return **true** if all element is processed, return **false** if abort by callback.

2. for every `DoWith` method,
    * `DoWith` return **false** if callback is not executed for some reason.
    * if callback executed, `DoWith` return whatever callback returns, so that it is easy to get callback return value from `DoWith` return value.

peterbell10: The interface can't really be changed because the functions are exported to the Lua API. Which is why new interfaces should try to be consistent with the old.

peterbell10:
The lua API checking is actually done by a plugin rather than a unit test.  You could maybe run the same [script as travis uses][1] but I wouldn't recommend it as it is only intented for use in travis.  To just run the API checks you need to start cuberite and run
```
load APIDump
apicheck
```

As for the error, it means you need to add documentation for the new function in the Lua API.  This is again done by the (APIDump plugin)[2]; normally you would add your documentation to [APIDesc.lua][3] but in the case of `cWorld` it has [it's own file][4].

[1]: https://github.com/cuberite/cuberite/blob/master/CIbuild.sh
[2]: https://github.com/cuberite/cuberite/tree/master/Server/Plugins/APIDump
[3]: https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/APIDesc.lua
[4]: https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/Classes/World.lua

src/World.cpp
madmaxoft: What @changyongGuo describes is the general idea I had when designing those functions from the start, if they behave different anywhere, it should be considered a bug and fixed, even if it means breaking plugins in the process (I don't think there are that many plugins that use those APIs, yet, so it should be fine). But such a change is for another PR, this PR should add the function so that it adheres to this API.

I'll check all the `DoWith` functions and make a list of those that need changing, as a separate issue.

src/World.h
```diff
@@ -469,7 +469,7 @@ class cWorld :
 
 	/** Spawns an experience orb at the given location with the given reward, split into multiple experience orbs if possible,
 	Returns spawned experience orbs UniqueID */
-	virtual std::vector<UInt32> SpawnSplitExperienceOrbs(double a_X, double a_Y, double a_Z, int a_Reward) override;  // TODO: Exported in ManualBindings.cpp
+	virtual std::vector<UInt32> SpawnSplitExperienceOrbs(double a_X, double a_Y, double a_Z, int a_Reward) override;
```
peterbell10: Please leave the comment as `// Exported in ManualBindings_World.cpp`

src/Bindings/ManualBindings_World.cpp
```diff
@@ -621,6 +665,38 @@ static int tolua_cWorld_ScheduleTask(lua_State * tolua_S)
...
+	{
+		return 0;
+	}
+
+	cWorld * self = nullptr;
+	double X = 0;
+	double Y = 0;
+	double Z = 0;
+	int Reward = 0;
+	L.GetStackValues(1, self, X, Y, Z, Reward);
```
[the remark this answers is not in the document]
changyongGuo: Is it ok to port API to lua with different argument list ? will this lead to some confusion?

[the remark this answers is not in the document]
changyongGuo: I am really confused here, I don't know how should I write method argument description for DoWithNearestPlayer, according to `tolua_cWorld_DoWithNearestPlayer`,  it get parameter x,y,z from input 2 - 5, should I descript parameter as number rather than vector3d?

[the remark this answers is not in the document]
peterbell10: 1. It's okay to have different signatures in the Lua API but if you want to change the functions to use `Vector3d` on the C++ side then that would be even better.

2. The Lua docs should always match the bindings so if the bindings take a vector then the docs should say it takes a vector.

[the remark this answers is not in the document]
peterbell10: I'm realising now that your second comment is probably referring to the `CheckParam` part, in which case the way to do it would be with `CheckParamUserType`:

```cpp

// Check params:
cLuaState L(tolua_S);
if (
	!L.CheckParamSelf("cWorld") ||
	!L.CheckParamUserType(2, "Vector3<double>") ||
    // Other params ...
)
{
	return 0;
}
```

changyongGuo:
code updated, please review again

src/Bindings/ManualBindings_World.cpp
```diff
@@ -830,6 +875,39 @@ static int tolua_cWorld_ScheduleTask(lua_State * tolua_S)
...
```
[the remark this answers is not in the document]
changyongGuo: I don't think we are on same page.
this is another function `SpawnExperienceOrb`, do I need to change this one to vector3d ??
Do I need to change this function parameter list?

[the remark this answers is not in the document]
changyongGuo: I mis-understand your request, I ported DoWithNearestPlayer with vector3d interface

changyongGuo:
updated code for lua binding.

src/Bindings/ManualBindings_World.cpp
```diff
@@ -463,6 +463,52 @@ static int tolua_cWorld_DoWithPlayerByUUID(lua_State * tolua_S)
...
+	cWorld * Self;
+	Vector3d * Position;
+	double RangeLimit;
+	cLuaState::cRef FnRef;
+	bool CheckLineOfSight = true, IgnoreSpectators = true;  // Defaults for the optional params
+	L.GetStackValues(1, Self, Position, RangeLimit, FnRef, CheckLineOfSight, IgnoreSpectators);
+
+	if (!FnRef.IsValid())
+	{
+		return L.ApiParamError("Expected a valid callback function for parameter #5");
```
peterbell10: This should now say parameter number 3 (it's offset by 1 from the `CheckParam` above because the "self" parameter is passed implicitly).

src/Bindings/ManualBindings_World.cpp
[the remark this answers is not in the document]
peterbell10: Sorry for the confusion, `SpawnExperienceOrb` is already exported to the API so I don't expect you to change it.  It's also good that `DoWithNearestPlayer` is exported with vectors too.

We're very slowly trying to change functions with three separate numbers as a coordinate to using `Vector3`s and all new lua exports should be using the new style.  It's harder to deal with changes to the lua API because the deprecated interfaces have to be maintained via manual bindings and plug-in authors would have to change their code.

src/World.h
```diff
@@ -465,6 +465,14 @@ class cWorld  // tolua_export
 	Returns the UniqueID of the spawned experience orb, or cEntity::INVALID_ID on failure. */
 	virtual UInt32 SpawnExperienceOrb(double a_X, double a_Y, double a_Z, int a_Reward) override;
 
+	// tolua_end
+
+	/** Spawns an experience orb at the given location with the given reward, split into multiple experience orbs if possible,
+	Returns spawned experience orbs UniqueID */
```
madmaxoft: Again adapt the Lua API docs wording.

src/MobSpawner.cpp
```diff
@@ -86,8 +86,14 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, int a_RelX, int a_RelY, int a_R
...
-	cPlayer * a_Closest_Player = a_Chunk->GetWorld()->FindClosestPlayer(a_Chunk->PositionToWorldPosition(a_RelX, a_RelY, a_RelZ), 24);
-	if (a_Closest_Player != nullptr)  // Too close to a player, bail out
+	bool PlayerExist = false;
+	a_Chunk->GetWorld()->DoWithNearestPlayer(a_Chunk->PositionToWorldPosition(a_RelX, a_RelY, a_RelZ), 24, [&](cPlayer & a_Player) -> bool
+	{
+		PlayerExist = true;
+		return true;
+	});
+
+	if (PlayerExist)  // Too close to a player, bail out
```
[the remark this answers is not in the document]
changyongGuo: is it too complex to put lambda expression into if condition?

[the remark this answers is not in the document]
madmaxoft: I think it's alright, personally I'd write it like this:
```cpp
// If too close to any player, don't spawn anything:
auto worldPos = a_Chunk->PositionToWorldPosition(a_RelX, a_RelY, a_RelZ);
static const double rangeLimit = 24;
if (a_Chunk->GetWorld()->DoWithNearestPlayer(worldPos, rangeLimit,
	[](cPlayer & a_Player)
	{
		return true;
	}
)
{
	return false;
}
```

madmaxoft:
The CI build stalled, I've restarted the job.

PR #4260, opened 2018-07-22
Optimise chunk set

### LINKED ISSUE #1244: Optimize chunk setting after loading and generating
Commit c0b62ef139a65ca648135fb6999e6623438fdd71 has introduced a queue for setting the chunk data; unfortunately this means that the data will get copied around a lot - first into the `cSetChunkData` object and then into the chunk itself. Some places could optimize some of the copies away - such as the generator, when it compresses the blockmetas into yet another separate array which is then copied into `cSetChunkData`; the Anvil loader could compose the chunk blocktypes into the object's blocktypes array directly, etc.

peterbell10:
Closes #1244

Initially I was just going to add the `cChunkData` to `cSetChunkData` but profiling revealed that the copying wasn't even the biggest slowdown.  Much more time was being spent in `cChunk::CreateBlockEntities` and `cChunk::WakeUpSimulators` than was in `memcpy` so I've made those significantly faster as well.

Optimisations performed:
 * `cSetChunkData` now stores blocks in a `cChunkData` object
 * `cChunkData` objects can now perform moves even if they are using different pools
 * `cChunk::CreateBlockEntities` now iterates in the correct order and only over present chunk sections
 * Similarly for `cChunk::WakeUpSimulators`
 * `cSetChunkData::CalculateHeightMap` now shortcuts to the highest present chunk section before checking blocks directly

Performance results from timing Cuberite start-up with the world already generated:

| Current Master | This PR      |
|----------------|--------------|
| 1282 ± 76 ms   | 1023 ± 45 ms |

image

### DIFF
--- a/src/SetChunkData.cpp
+++ b/src/SetChunkData.cpp
@@ -8,6 +8,17 @@
 #include "BlockEntities/BlockEntity.h"
 #include "Entities/Entity.h"
 
+namespace
+{
+struct sMemCallbacks:
+	cAllocationPool<cChunkData::sChunkSection>::cStarvationCallbacks
+{
+	virtual void OnStartUsingReserve() override {}
+	virtual void OnEndUsingReserve() override {}
+	virtual void OnOutOfReserve() override {}
+};
+}  // namespace (anonymous)
+
 
 
 
@@ -15,6 +26,8 @@
 cSetChunkData::cSetChunkData(int a_ChunkX, int a_ChunkZ, bool a_ShouldMarkDirty) :
 	m_ChunkX(a_ChunkX),
 	m_ChunkZ(a_ChunkZ),
+	m_Pool(cpp14::make_unique<sMemCallbacks>(), cChunkData::NumSections),
+	m_ChunkData(m_Pool),
 	m_IsLightValid(false),
 	m_IsHeightMapValid(false),
 	m_AreBiomesValid(false),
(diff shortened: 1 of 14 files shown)


src/SetChunkData.cpp
```diff
@@ -95,14 +94,25 @@ cSetChunkData::cSetChunkData(
...
+	size_t MaxSection = 0;
+	for (size_t i = cChunkData::NumSections - 1; i != 0; --i)
+	{
+		if (m_ChunkData.GetSection(i) != nullptr)
+		{
+			MaxSection = i;
+			break;
+		}
+	}
+	const int MaxHeight = static_cast<int>(MaxSection + 1) * cChunkData::SectionHeight - 1;
```
madmaxoft: Couldn't you just use `int` for `MaxSection` and do without the cast?

src/SetChunkData.h
```diff
@@ -51,20 +49,11 @@ class cSetChunkData
...
-	const cChunkDef::BlockNibbles & GetBlockLight(void) const { return m_BlockLight; }
-
-	/** Returns the internal storage of the block types, read-only. */
-	const cChunkDef::BlockNibbles & GetSkyLight(void) const { return m_SkyLight; }
+	/** Returns the internal storage of block types, metas and lighting. */
+	cChunkData & GetChunkData(void) { return m_ChunkData; }
 
 	/** Returns the internal storage for heightmap, read-only. */
-	const cChunkDef::HeightMap & GetHeightMap(void) const { return m_HeightMap; }
+	cChunkDef::HeightMap & GetHeightMap(void) { return m_HeightMap; }
```
madmaxoft: The comment no longer matches the function (non-const)

src/SetChunkData.cpp
peterbell10: `cChunkData::GetSection` takes a `size_t` so AFAIK there has to be a cast somewhere.

src/SetChunkData.h
peterbell10: Sorry, stale change.  I've fixed it now.

src/SetChunkData.cpp
madmaxoft: Fair enough.

PR #4262, opened 2018-07-23
At long last... Piston animations!

### LINKED ISSUE #57: Pistons not animating correctly
~~The block action packet (0x36) is either not sent or not always sent to the client. 95% of the time, a piston extends or retracts without animation.~~

It seems QueueSetBlock doesn't tick the correct number of ticks.

### LINKED ISSUE #3198: Piston pushing has no animation
Client version:   1.9.4
Commit id:        f73de02684ce2c5f9fb2e8e36cfe65dc4ac91dea
### Expected behavior

When a redstone signal is applied to a piston, an extending animation will be shown
### Actual behavior

When a redstone signal is applied to a piston, the block it's pushing teleports

Put a block infront of a piston and apply a redstone signal

tigerw:
Five years in the making, and after making xoft *very* mad; here it is!

* Fixes #3198
* Fixes #57 (again lol)

### DIFF
--- a/src/Blocks/BlockPiston.cpp
+++ b/src/Blocks/BlockPiston.cpp
@@ -214,36 +214,58 @@ bool cBlockPistonHandler::CanPushBlock(
 
 void cBlockPistonHandler::ExtendPiston(Vector3i a_BlockPos, cWorld & a_World)
 {
-	BLOCKTYPE pistonBlock;
-	NIBBLETYPE pistonMeta;
-	a_World.GetBlockTypeMeta(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, pistonBlock, pistonMeta);
-
-	if (IsExtended(pistonMeta))
 	{
-		// Already extended, bail out
-		return;
+		// Broadcast block action first. Will do nothing if piston cannot in fact push
+
+		BLOCKTYPE pistonBlock;
+		NIBBLETYPE pistonMeta;
+		a_World.GetBlockTypeMeta(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, pistonBlock, pistonMeta);
+		a_World.BroadcastBlockAction(a_BlockPos, PistonExtendAction, pistonMeta, pistonBlock);
 	}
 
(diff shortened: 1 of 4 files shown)


src/Blocks/BlockPiston.cpp
```diff
@@ -214,37 +214,58 @@ bool cBlockPistonHandler::CanPushBlock(
...
-	if (IsExtended(pistonMeta))
 	{
-		// Already extended, bail out
-		return;
+		// Broadcast block action first. Will do nothing if piston cannot in fact push
+
+		BLOCKTYPE pistonBlock;
+		NIBBLETYPE pistonMeta;
+		a_World.GetBlockTypeMeta(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, pistonBlock, pistonMeta);
+		a_World.BroadcastBlockAction(a_BlockPos, 0, pistonMeta, pistonBlock);
```
NiLSPACE: Should this magic number be replaced with a constant?

src/Blocks/BlockPiston.cpp
```diff
@@ -253,52 +274,69 @@ void cBlockPistonHandler::ExtendPiston(Vector3i a_BlockPos, cWorld & a_World)
...
-	NIBBLETYPE pistonMeta;
-	a_World.GetBlockTypeMeta(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, pistonBlock, pistonMeta);
-
-	if (!IsExtended(pistonMeta))
 	{
-		// Already retracted, bail out
-		return;
+		BLOCKTYPE pistonBlock;
+		NIBBLETYPE pistonMeta;
+		a_World.GetBlockTypeMeta(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, pistonBlock, pistonMeta);
```
NiLSPACE: Here as well?

src/Blocks/BlockPiston.cpp
tigerw: They've been here since the code was first added, and you would be so heartless as to remove them now? 👁‍🗨 👁‍🗨

PR #3135, opened 2016-04-09
1.9 / 1.9.2 / 1.9.3 / 1.9.4 protocol support

Pokechu22:
This updates to the 1.9 protocol, and also supports 1.9.1 and 1.9.2 (which have separate protocol IDs).  This update includes the new chunk format.  It does _not_ add any new blocks or items.  It also makes a few other changes:

* Fixed a bug with CheckBasicStyle.lua that caused it to not properly strip line endings, causing issues when used in Cygwin
* Add a `/blk` command for debugging (which was used to figure out what was supposed to be what when trying to diagnose highly broken chunks)
* Split entity mounting/dismounting into separate functions, because 1.9's set passengers packet works with the ridden entity rather than the rider and thus wouldn't be portable with the old method (where the ridden entity would be null when dismounting).

There's a few things that I know are not quite finished yet:

* Spawn eggs don't work since they have different IDs now
* The offhand slot is broken and doesn't work at all.  I haven't tried to write code to deal with it since that's a more complex part of the project that I don't quite understand
* The "[Effect](http://wiki.vg/Protocol#Effect)" packet (the one that does particles and sounds) uses new IDs in 1.9; I haven't tried to update the IDs used in the project since they are in a single enum which doesn't have a good way of converting.
* UUIDs were added to a bunch of packets, but since there isn't a full UUID generated, I use a very non-unique UUID generated from the entity's ID.  The client doesn't seem to care, but this could cause issues.
* I suspect that the new slot in the brewing inventory will cause things to not work properly in the brewing stand (if the brewing stand is implemented), but have not tested it.
* Boats don't seem to work
* No palette is created for chunks, meaning that all blocks are sent with the full 13 bits.  This is more of a performance issue than anything, and since the whole chunk section is sent even if the upper parts are empty, may dramatically reduce performance in comparison to what could be possible.

The main technique I used was following the [Protocol History](http://wiki.vg/Protocol_History) page, rather than trying to rewrite the entire protocol from scratch.  Of course, I wrote the protocol history page a bit ago by using [Burger](https://github.com/mcdevs/Burger)... which I had updated even before that.  So I used a lot of information that wasn't there a month or so ago.  By using protocol history, I could work version-by-version and test along the way, rather than making all the changes and hoping it works.

For the chunk format info, wiki.vg's article is wrong.  I'm going to rewrite sometime later, but the method I used here is based off of what is used by the game itself.

@LogicParrot edit:
Below are some testing binaries.
[...]

Issue #4124, opened 2017-12-30
UrlClient-test failures on TravisCI apple builds


peterbell10:
Client version:   N/A
Server OS:        macOS
Commit id:       Anything after 5b6b04375270f3f3ba90780ecf692d56af157e49 (When apple was added to travis)

### Expected behavior
UrlClient-test always passes.

### Actual behavior
UrlClient-test will sometimes fail outright, other times it deadlocks and the travis worker times out.
[The first case](https://travis-ci.org/cuberite/cuberite/builds/303659238) was a timeout and happened very soon after apple was added to travis.
[This one](https://travis-ci.org/cuberite/cuberite/jobs/321859533) actually failed with a segfault though timing out seems more common.

### Steps to reproduce the behavior
Unknown.

peterbell10:
Looks like this is still happening:
https://travis-ci.org/cuberite/cuberite/jobs/333420151

tigerw:
https://travis-ci.org/cuberite/cuberite/jobs/407388570

Evidently Mac OS's fault.

madmaxoft:
Still there, and getting more frequent.
(link unavailable, job was restarted several times)

tigerw:
Seems like that fixed it.

PR #4263, opened 2018-07-23
Reimplemented fall particles

tigerw:
Regression when 1.9.x support introduced in #3135, originally implemented in #461.

### DIFF
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -428,9 +428,14 @@ void cPawn::HandleFalling(void)
 			TakeDamage(dtFalling, nullptr, Damage, Damage, 0);
 
 			// Fall particles
-			// TODO: Re-enable this when effects in 1.9 aren't broken (right now this uses the wrong effect ID in 1.9 and the right one in 1.8)
-			// int ParticleSize = static_cast<int>((std::min(15, Damage) - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f);
-			// GetWorld()->BroadcastSoundParticleEffect(EffectID::PARTICLE_FALL_PARTICLES, POSX_TOINT, POSY_TOINT - 1, POSZ_TOINT, ParticleSize);
+			GetWorld()->BroadcastParticleEffect(
+				"blockdust",
+				GetPosition(),
+				{ 0, 0, 0 },
+				(Damage - 1.f) * ((0.3f - 0.1f) / (15.f - 1.f)) + 0.1f,  // Map damage (1 - 15) to particle speed (0.1 - 0.3)
+				static_cast<int>((Damage - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f),  // Map damage (1 - 15) to particle quantity (20 - 50)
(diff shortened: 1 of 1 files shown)


src/Entities/Pawn.cpp
```diff
@@ -428,9 +428,14 @@ void cPawn::HandleFalling(void)
...
 			// Fall particles
-			// TODO: Re-enable this when effects in 1.9 aren't broken (right now this uses the wrong effect ID in 1.9 and the right one in 1.8)
-			// int ParticleSize = static_cast<int>((std::min(15, Damage) - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f);
-			// GetWorld()->BroadcastSoundParticleEffect(EffectID::PARTICLE_FALL_PARTICLES, POSX_TOINT, POSY_TOINT - 1, POSZ_TOINT, ParticleSize);
+			GetWorld()->BroadcastParticleEffect(
+				"blockdust",
+				GetPosition(),
+				{ 0, 0, 0 },
+				(Damage - 1.f) * ((0.3f - 0.1f) / (15.f - 1.f)) + 0.1f,  // Map damage (1 - 15) to particle speed (0.1 - 0.3)
+				static_cast<int>((Damage - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f),  // Map damage (1 - 15) to particle quantity (20 - 50)
```
NiLSPACE: Where did we get the math from?

src/Entities/Pawn.cpp
```diff
@@ -428,9 +428,14 @@ void cPawn::HandleFalling(void)
 			TakeDamage(dtFalling, nullptr, Damage, Damage, 0);
 
 			// Fall particles
-			// TODO: Re-enable this when effects in 1.9 aren't broken (right now this uses the wrong effect ID in 1.9 and the right one in 1.8)
-			// int ParticleSize = static_cast<int>((std::min(15, Damage) - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f);
-			// GetWorld()->BroadcastSoundParticleEffect(EffectID::PARTICLE_FALL_PARTICLES, POSX_TOINT, POSY_TOINT - 1, POSZ_TOINT, ParticleSize);
+			GetWorld()->BroadcastParticleEffect(
```
peterbell10: This function already exists by the way:

https://github.com/cuberite/cuberite/blob/31a11a6df4922b590a50a5ff3d3c00d42a45599d/src/Broadcaster.h#L11-L13

You have to get to it through a `cBroadcaster`
```cpp
GetWorld()->GetBroadcaster().BroadcastParticleEffect([...]);
```

src/Entities/Pawn.cpp
tigerw: Technically https://github.com/cuberite/cuberite/commit/de2cce9b79ab60427daa86dcd1da99db7d7ea161#diff-ab7bd0f99b0b93eb1655e0775b322933 but I just wrote code to map one range of numbers to another. Both ranges were guesses but looks about right.

Also, m a t h e m a t i c
(use maths, it's superior ;))

tigerw:
The build failed for networking reasons. I wonder why.

peterbell10:
Looks like #4124, only happens on apple for some reason.

PR #4264, opened 2018-07-24
Broadcast refactor

peterbell10:
I've had this code sitting around for a while and #4263 has just reminded me of it.

Currently a broadcast gets manually forwarded through many layers of code,
```
cWorld -> cChunkMap -> cChunk -> cClientHandle -> cProtocol
```
Whereas this uses `DoWithChunk` functions to remove broadcast functions from `cChunkMap` and `cChunk`.
This was inspired by @worktycho's [`cBroadcaster`](https://github.com/cuberite/cuberite/commit/1386fdb899100b9d7ba97f63ac40678b5c3afe0b) but uses templates and lambdas instead of code generation.

Unlike `cBroadcaster`, I've kept the broadcasting functions in `cWorld` so the main API isn't being changed.  However, I recognised that `cBroadcaster` was also useful to avoid include dependencies on `World.h`. So, I've improved `cBroadcastInterface` to include _all_ broadcasts and added `OpaqueWorld.h` to allow `cWorld` to be treated as an opaque type and still be able to convert it into any of it's interfaces.

Ref #1134 (Maybe closes? It still doesn't address duplication in `cClientHandle` though)


tigerw:
dat d e c o u p l i n g


you are officially Tycho's new favourite developer

tigerw:
+826 -1524

...it's... *beautiful!* Net reduction in code = automatic approval

tigerw:
Merge?

tigerw:
If nobody responds in 30 minutes I am legally allowed to merge

madmaxoft:
I'm responding just so that you don't rush-merge it in :) I want to first see this beauty for myself.

tigerw:
drat foiled again

src/World.h
```diff
@@ -160,57 +159,79 @@ class cWorld :
...
+	{
+		LOG("BroadcastSoundEffect with double position arguments is deprecated, use vector-parametered version instead.");
+		BroadcastSoundEffect(a_SoundName, { a_X, a_Y, a_Z }, a_Volume, a_Pitch, a_Exclude);
+	}
+
+	void BroadcastSoundParticleEffect(const EffectID a_EffectID, int a_SrcX, int a_SrcY, int a_SrcZ, int a_Data, const cClientHandle * a_Exclude = nullptr)
+	{
+		BroadcastSoundParticleEffect(a_EffectID, { a_SrcX, a_SrcY, a_SrcZ }, a_Data, a_Exclude);
+	}
+	// tolua_end
```
madmaxoft: These would be better implemented in Bindings/ManualBindings.cpp, so that they can show the Lua stack trace (consider trying to fix an error message like this from a large plugin, such as WorldEdit).

madmaxoft:
Requesting mostly cosmetic changes, the code is, indeed, beautiful :D

src/World.h
peterbell10: This is how it was being done already, the only change here is that I've moved the deprecated overloads into the header.

madmaxoft:
I still don't understand the relevance of `Loading` in those functions

src/World.h
madmaxoft: Oh well, I guess a separate PR, then.

peterbell10:
Do my additions make the comments any clearer? I'm not sure how to describe it in a way that isn't just: "Clients in the `cChunk::m_LoadedByClient` list".

madmaxoft:
How about naming the first two functions `ForClientsWithChunk`, and the comment along the lines of
`/** Calls the function object a_Func for every client who has the specified chunk. */`
Similarly, `ForClientsWithEntity`
"Having a {chunk|entity}" feels much more understandable than "loading a {chunk|entity}" in this context.

peterbell10:
Okay, I took your names and I also renamed `ForClientsLoadingPos` to `ForClientsWithChunkAtPos` to be consistent with the `DoWithChunkAtPos` function naming.

madmaxoft:
This looks good. I thought you could use the same name for the two functions, but this works well, too.

PR #4265, opened 2018-07-25
cWorld: Manually bind deprecated broadcast functions

peterbell10:
As per @madmaxoft's comment on https://github.com/cuberite/cuberite/pull/4264#discussion_r204769193


Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -389,6 +421,30 @@ return
...
+			{
+				Params =
+				{
+					{
+						Name = "EffectID",
+						Type = "number",
+					},
+					{
+						Name = "SourcePos",
+						Type = "Vector"
```
madmaxoft: This is the one causing failed CI, needs to be `Vector3f` instead.

peterbell10: Thanks, fixed now.

src/Bindings/ManualBindings_World.cpp
```diff
@@ -13,48 +13,258 @@
...
+		!CheckParamVectorOr3Numbers(L, "Vector3<double>", 3, VolumeIndex) ||
+		!L.CheckParamNumber(VolumeIndex, VolumeIndex + 1)
+	)
+	{
+		return 0;
+	}
+
+	if (VolumeIndex != 4)  // Not the vector overload
+	{
+
```
madmaxoft: Needless extra line

PR #4268, opened 2018-07-28
WIP: A framework for modular mob AI

peterbell10:
I'm thinking about the possibility of continuing work on the [`mobs2`][mobs2] branch. To that end, I've been reading up on modular game AI and think [behaviour trees][wiki] might be an interesting alternative to @SafwatHalaby's design.

## Motivation
The current design of `mobs2` gives control to whichever behaviour in a list is the first to requests it.  This allows prioritisation (e.g. Attacking is more important than wandering) but if more complex decision making is required, it has to be embedded into the behaviour itself. For example, spiders are only aggressive at night and so `cBehaviorAggressive` is forced to take a callback function that is used by `cSpider` to tell the behaviour whether or not to be aggressive. If left unchecked, this could lead to a mess of special cases embedded into each behavior.  Further, it only works if you know where these customisation points are required in the first place, if we want lua to be able to modify the AI in interesting ways then this likely won't be flexible enough. 

## Possible solution - Behaviour Trees
Behaviour trees offer a more general way of defining the AI's decision making and acting process.  Leaf nodes in the tree represent some action that the AI can perform; other nodes provide control flow for deciding which actions to perform.  A node can return one of three status codes, success, failure or running.  Success and failure allow nodes to dictate the control flow, and running indicates that a task will take multiple ticks to complete. For example, to implement a spider, you might have an action `cIsNightTime` that returns success at night and have control flow that will only transfer to the aggressive behaviour sub-tree if that returns success.  A `cMoveToPosition` action takes multiple ticks to complete and so returns running until it either reaches its destination or fails. 

For a more in depth introduction to behaviour trees, I can recommend [this article][gama].

Using this type of system will allow modularity in the actions that can be performed but also allow handling of special cases like spiders without changing the modular components themselves.  Further, because the trees follow a generic interface, there's nothing to stop you from building these trees dynamically through some factory function. This could allow lua plugins to define their own AI from a lua table or json file without any need to expose the internal workings of `cMonster`.

## Toy example
I've made this toy AI for a cow to help demonstrate some of the ideas.  In the `cCow` constructor, I build the tree indicated by the diagram below.  Each tick, the cow checks if its health is below 50% and if it is, it ticks the "panicking" sub-tree where it picks a random position and moves to it.  If its health is above 50%, it will instead "wander" by moving randomly and waiting for a random period of time before moving again.  If at any point you attack the cow and its health drops below 50%; the `cHealthRange` check will succeed, the wandering subtree will be cancelled and the cow starts panicking.

While "panicking" and "wandering" would fit into the `mobs2` design, there would be no way to define this health range condition to control when we move into the panicked state.  Also, the fact that I was able to create both panicking and wandering without defining any new classes should give some glimpse into the flexibility of this approach.

####  I would appreciate any thoughts/questions/criticisms

cow

[mobs2]: https://github.com/cuberite/cuberite/compare/mobs2
[wiki]: https://en.wikipedia.org/wiki/Behavior_tree_(artificial_intelligence,_robotics_and_control)
[gama]: https://www.gamasutra.com/blogs/ChrisSimpson/20140717/221339/Behavior_trees_for_AI_How_they_work.php


tigerw:
What are the drawbacks with just writing normal code? It's like you're trying to define the AI using XML or something, which is laudable but perhaps unnecessary?

peterbell10:
First I should say my intention is that the "actions" would be much more complex than the examples given here.  So, a large part of the AI would just be normal code.

The main benefit of this is the separation of responsibility.  The actions shouldn't concern itself with decision making for the AI, that is left to the behaviour tree structure.  This means that the decision making isn't sprawled out through several layers of inheritance and is instead contained in just one place.
Similarly, the person creating the tree doesn't need to manually handle state changes and remember to cancel the current task etc.  This is assured by the implementation of the tree.

madmaxoft:
This whole thing is just screaming to be actually implemented in Lua :)
We have all the plumbing there for the plugins, it should be pretty straightforward to implement the BTs in Lua, with the added benefit of using a Lua table for the "blackboard" storage. Behaviors could then be simply reloaded while the server is running, while debugging them.

peterbell10:
Would you actually be able to implement the behaviours from lua though?

tigerw:
> AI
> implemented in Lua

final_5c92ab2c9c9240001446cc9e_665017

12xx12:
So I had a look at this again, updated to master and fixed compile.

I realy like the Ideas you added.

**Lua**
For the sake of consistency we can think about using Lua. I would suggest a layout similar to the `APIDesc.lua` file

12xx12:
I did some brainstorming

We would need some Behaviours for:

- [ ] follow entity (identified by unique ID maybe)
Like cats and dogs
- [ ] attack entity (identified by unique ID maybe)
just one designated entity
- [ ] keep distance from entity (identified by unique ID maybe)
e.g. crepper and phantoms keeping distance from cats
- [ ] drop items (maybe XP)
eggs from chickens. Maybe even use this on death to drop loot
- [ ] breeding functionality
- [ ] denial behaviour when trying to tame a donkey/horse/mule
- [ ] teleport to location
enderman, wolfes, cats
- [ ] fire projectiles
blazes, whitches, skeleton, shulker, llama

- [ ] select nearest entity of multiple types (result is the ID of the entity maybe) 
for attacking. And some monsters attack other monsters
- [ ] check surrounding players for item(s) (variable range and slot)
pinkins for enderman, following players with food in their hand
- [ ] find block with certain conditions (light,...)
foxes sleeping, agressive monster in daylight
- [ ] check if another entity is attacked
wolfes, llamas with wandering traders
- [ ] check time
spiders, foxes, ...
- [ ] check weather

This is just a raw sketch. Feel free to make remarks. I'd like to jupstart this again

peterbell10:
> Lua
> For the sake of consistency we can think about using Lua. I would suggest a layout similar to the APIDesc.lua file

I guess cause you mention `APIDesc.lua` you mean describing the static trees in lua, rather than actually running lua code when evaluating the tree? If you mean the first one, then I agree it makes sense but I don't think needs to be a top priority. If you mean the second, then I would be seriously concerned about performance if lots of lua code was being run for every entity on every tick.

> drop items (maybe XP)
> eggs from chickens. Maybe even use this on death to drop loot

Makes perfect sense for chickens. Don't think it makes sense sense for death drops though. That's not really part of the entities decision making (i.e intelligence), and I think loot tables would make them user configurable anyway (?).

> check if another entity is attacked
> wolfes, llamas with wandering traders

An interesting point is we don't want to change from event-based (AI reacting to an attack) to polling based (AI has to check every tick to see if the entity was attacked). Ideally I think the AI would be able to subscribe to events on another entity and when the event is triggered, it would be put on the AI's blackboard. Something like that anyway.

Overall, I suggest starting small like I was trying yo do with the cow example. We shouldn't need to rewrite every mobs for this to be useful. And I think being over-ambitious was part of why the earlier attempts at AI rewrites didn't materialise.

12xx12:
> I guess cause you mention `APIDesc.lua` you mean describing the static trees in lua, rather than actually running lua code when evaluating the tree? 

Yeah this is what i meant - enything else would defeat the purpose of this project

> Overall, I suggest starting small like I was trying yo do with the cow example. We shouldn't need to rewrite every mobs for this to be useful. And I think being over-ambitious was part of why the earlier attempts at AI rewrites didn't materialise.

So we start with the cow and then we everything bit by bit and if we need new behaviour we just add it

peterbell10:
> So we start with the cow and then we everything bit by bit and if we need new behaviour we just add it

The cow was just a simple example, but yes. I think changing over one mob at a time and adding behaviours as they're needed will be easier than trying to design everything upfront. Recognising good building-block behaviours like your list and keeping it in mind is still a great idea though.

12xx12:
@peterbell10 

Do you mind me doing some work on this?

PR #4272, opened 2018-07-29
Ocelots no longer multiply exponentially

### LINKED ISSUE #4271: Server slowing to a crawl
Client version:   1.12.2
Server OS:        Linux
Commit id:        90d552a702a435f644c29f75fc2a85c0d71ee510

The server is spitting SetSwimState failures all over the place and using 100% of a single core.
Video: https://youtu.be/fw4GpIy_xhg

peterbell10:
Along with a call to `destroyentities`, this fixes #4271

I'm guessing the intention of this code was to modify the normal spawning of ocelots. However, `cEntity::SpawnOn` is actually called to send the entity to an individual client.  That means this code was run for every single player, every time they were sent a chunk with ocelots in it.  Thus, the ocelots population would grow exponentially as players log in and move around.

### DIFF
--- a/src/Mobs/Ocelot.cpp
+++ b/src/Mobs/Ocelot.cpp
@@ -11,6 +11,15 @@
 
 
 
+// TODO: Ocelots should have a chance of spawning with two kittens
+/*
+	if (!IsBaby() && GetRandomProvider().RandBool(1.0 / 7.0))
+	{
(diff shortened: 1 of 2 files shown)


Bond-009:
Wow, nice catch. Had I looked at the comment for `cEntity::SpawnOn` when I wrote that code a year ago it wouldn't have bitten me in ass now.

PR #4274, opened 2018-07-31
Initial support for 1.13 clients

tigerw:
+ Can look at *shape* of world
+ Can look at *shape* of inventory

That's about it, unless you count that all players now have one singular statistic:

### DIFF
--- a/src/Protocol/CMakeLists.txt
+++ b/src/Protocol/CMakeLists.txt
@@ -14,6 +14,7 @@ SET (SRCS
 	Protocol_1_10.cpp
 	Protocol_1_11.cpp
 	Protocol_1_12.cpp
+	Protocol_1_13.cpp
 	ProtocolRecognizer.cpp
 )
 
@@ -29,6 +30,7 @@ SET (HDRS
 	Protocol_1_10.h
 	Protocol_1_11.h
 	Protocol_1_12.h
+	Protocol_1_13.h
 	ProtocolRecognizer.h
 )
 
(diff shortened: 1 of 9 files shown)


Pokechu22:
Just as advice for this process: cuberite currently doesn't implement chunk section palettes, but instead uses metadata.  Metadata simply isn't a functioning idea anymore as of 1.13; there isn't enough context.  Implementing palettes properly will work for multiple versions and the serialization process could figure out the right ID from the version's global palette based off of a section palette.  Data generators will also be helpful for this; both of those things are documented on wiki.vg.

tigerw:
?

This just changes the id and packet formats slightly so clients can connect. Doesn't stop any internal representation changes.

madmaxoft:
Wait, what? I thought you had to completely change the format in which the chunk data was sent, if you didn't do that, how does the client even know what the blocks are?

tigerw:
??

The ids sent to the client aren't correct due to having no palette support, which explains the emphasis on the *shape* of the world. So the block types are rendered incorrectly, and in the interest of incremental, reviewable PRs (and basic 1.13 support), the server backend hasn't changed. The code treats the E_BLOCK_ constants' values as values in the global palette.

What do you mean by "completely change the format in which the chunk data was sent"? cChunkData was not modified.

ForTheReallys:
Is there any update on this?

madmaxoft:
Not on this, I'm working on a proper support, but it's gonna take more time (it did take Mojang over a year, after all)

ForTheReallys:
I just started reading up on the minecraft protocols so I'm not too familiar with them yet, but is there anything I can do to help with this?

madmaxoft:
#4372 would be helpful, and it just means massaging external data sources until you get out the data in a format we want :)

ForTheReallys:
Dang I didn't see this til it was too late. Let me know if there's anything else I can help with. In the meantime, I'll just see if there are any issues I can help resolve

tigerw:
ehm hello again @madmaxoft, how do I use your fancy new palettes? Last I checked 1.13 clients connect with this, they just need the right values sent to them.

madmaxoft:
Ahh, the Tiger, my old young nemesis :)
Look at the source for PalettedBlockArea, it has extensive comments. Also, there's a test of the class that shows how it expects to be used.

madmaxoft:
I think I'll use parts of this to work on the 1.13 protocol proper. I've just tested it, the client managed to connect, but a few seconds later crashed due to some bad packet. And I finally see what you mean by the "shape" of the terrain :)

tigerw:
:)

definitely seeing the wood on the trees there

madmaxoft:
Closing as obsoleted by #4390.

PR #4280, opened 2018-08-03
cloc: Updated list of ignored files.

madmaxoft:
I had this sitting in the repo for some time, finally got around to committing.


cloc-exclude.txt
```diff
@@ -1,13 +1,10 @@
...
-src/LeakFinder.cpp
-src/LeakFinder.h
-src/StackWalker.cpp
-src/StackWalker.h
-VC2008
-VC2013
+src/Bindings/docs
+src/Bindings/Bindings.cpp
+src/Bindings/Bindings.h
+doxy
```
peterbell10: Should dev-docs be here as well

madmaxoft: Yes, good point. Added.

PR #4283, opened 2018-08-11
Fix: Create players folder recursively

tympaniplayer:
Problem: On a new server the players folder was not created on windows.

Root Cause:
`GetUUIDFolderName` was returning a folder structure for players with `/` while CreateFolderRecursively was checking for `\\` for win32. 

I struggled with the right approach. If I change `GetPathSeparator` it could have other implications. If I change `GetUUIDFolderName` it could also have other implications.

Let me know if you would like me to take a different approach on this fix. 

I know it's just a small fix, but it can save some windows server admins from having to add the players folder as a workaround.

### DIFF
--- a/src/OSSupport/File.cpp
+++ b/src/OSSupport/File.cpp
@@ -474,7 +474,7 @@ bool cFile::CreateFolderRecursive(const AString & a_FolderPath)
 
 	// Go through each path element and create the folder:
 	auto len = a_FolderPath.length();
-	auto PathSep = GetPathSeparator()[0];
+	auto PathSep = '/';
 	for (decltype(len) i = 0; i < len; i++)
 	{
 		if (a_FolderPath[i] == PathSep)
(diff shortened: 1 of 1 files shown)


tympaniplayer:
Thank you for the feedback and thank you for the welcome! My background is in C# so I expect that my code style may be a bit off (not formatting). Is the way I concat the strings the best way to do it in C++? Also there was one more line of code at 2144 in Players.cpp that uses `players/` but it mentions old file names. It looks like it's another place I should use the correct separator but wanted to get your opinion if only the file name itself is "oldstyle" or if the whole path is (doesn't look like it though)

tympaniplayer:
Ok I understand, sorry about my misunderstanding. I can revert the changes and just fix up `CreateFolderRecursive` instead?

peterbell10:
Yes, Sorry. I should have just said that in the first place.

tympaniplayer:
No worries! Thanks for the help. I'll get to work.

tympaniplayer:
Not sure if this is the best approach but it works 🙂

PR #4285, opened 2018-08-13
solve new player login crash issue

changyongGuo:
currently if new player login, server will crash during send player ability to client.
this pull request solve this issue.


src/ClientHandle.cpp
```diff
@@ -2725,7 +2727,10 @@ void cClientHandle::SendEntityAnimation(const cEntity & a_Entity, char a_Animati
 
 void cClientHandle::SendPlayerAbilities()
 {
-	m_Protocol->SendPlayerAbilities();
+	if (m_Player != nullptr)
+	{
+		m_Protocol->SendPlayerAbilities();
```
peterbell10: There is only one place in the entire codebase where `m_Player = nullptr` is possible and it was only introduced in #4248.  I think you should be fixing that directly:
https://github.com/cuberite/cuberite/blob/bdc8e0b83c5e3600a372397e1ee7a5c02c93117a/src/Entities/Player.cpp#L151-L152

peterbell10:
@changyongGuo are you not working on this any more?

changyongGuo:
oh..i think i misunderstand what you mean, i thought i need to create pull request from original game mode branch.

i will prepare new pull request later, sorry i dont have much time on this issue

peterbell10:
You can just make your changes in this pull request. After your previous pull request was merged, you shouldn't really continue to work on that branch because it might not merge properly.

Feel free to reopen this pull request if you'd like and it's completely fine if you need more time.

Issue #4112, opened 2017-12-25
Cannot place more than one lapis in enchanting table


9caihezi:
Client version:   1.12.2
Server OS:        Windows
Commit id:        f0b05e78f70fa883bfdc74568bd7bf914025306c

### Expected behavior
As described.

### Actual behavior
As described.

### Steps to reproduce the behavior
As described.

tympaniplayer:
I'm taking a look at this one now. Both slots in the window are treated the same right now which is the root cause of the problem.

PR #4286, opened 2018-08-13
Handle the lapis slot separately

tympaniplayer:
This is my attempt at a fix for issue #4112. The root cause of the issue was that the lapis slot was treated exactly the same as the enchanting slot, so it on the server side it would only ever slot one item. 

My fix is to check if its the second slot in the window, then check if it's lapis (it would slot whatever). If it is lapis I call the base click handler.

### DIFF
--- a/src/UI/SlotArea.cpp
+++ b/src/UI/SlotArea.cpp
@@ -1485,7 +1485,7 @@ void cSlotAreaEnchanting::Clicked(cPlayer & a_Player, int a_SlotNum, eClickActio
 	if (a_SlotNum == 1)
 	{
 		// Lapis slot can have a full stack handle it normally, also check for empty hand
-		if ((DraggingItem.m_ItemType <= 0) || ((DraggingItem.m_ItemType == E_ITEM_DYE) && (DraggingItem.m_ItemDamage == E_META_DYE_BLUE)))
+		if ((DraggingItem.IsEmpty()) || ((DraggingItem.m_ItemType == E_ITEM_DYE) && (DraggingItem.m_ItemDamage == E_META_DYE_BLUE)))
 		{
 			return cSlotArea::Clicked(a_Player, a_SlotNum, a_ClickAction, a_ClickedItem);
 		}
--- a/src/UI/SlotArea.cpp
+++ b/src/UI/SlotArea.cpp
@@ -1535,7 +1535,7 @@ void cSlotAreaEnchanting::Clicked(cPlayer & a_Player, int a_SlotNum, eClickActio
 
 void cSlotAreaEnchanting::DistributeStack(cItem & a_ItemStack, cPlayer & a_Player, bool a_Apply, bool a_KeepEmptySlots, bool a_BackFill)
 {
-	if ((a_ItemStack.m_ItemType == E_ITEM_DYE) || (a_ItemStack.m_ItemDamage == E_META_DYE_BLUE))
+	if ((a_ItemStack.m_ItemType == E_ITEM_DYE) && (a_ItemStack.m_ItemDamage == E_META_DYE_BLUE))
 	{
 		// It's lapis, put it in the lapis spot.
 		const cItem * Slot = GetSlot(1, a_Player);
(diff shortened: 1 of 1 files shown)


src/UI/SlotArea.cpp
```diff
@@ -1482,7 +1482,17 @@ void cSlotAreaEnchanting::Clicked(cPlayer & a_Player, int a_SlotNum, eClickActio
 		bAsync = true;
 	}
 	cItem & DraggingItem = a_Player.GetDraggingItem();
-
+	if (a_SlotNum == 1)
+	{
+		// Lapis slot can have a full stack handle it normally
+		if (DraggingItem.m_ItemType != E_BLOCK_LAPIS_ORE)
```
peterbell10: This is lapis ore rather than harvested lapis.  For the item it should be `m_ItemType == E_ITEM_DYE` and `m_ItemDamage == E_META_DYE_BLUE`.

tympaniplayer:
Looks like one of the Travis Clang builds hung. Is there a way to force it to rebuild?

src/UI/SlotArea.cpp
tympaniplayer: Oops yep! Good catch.

peterbell10:
I'm getting desyncs when I put non-lapis items into the lapis slot.  Maybe you need to broadcast the window when that happens?

tympaniplayer:
The broadcast doesn't help. I'll look into what I can do.

tympaniplayer:
Oh I have my logic all screwed up

tympaniplayer:
Desyncs should no longer happen, though I'm not sure if I should keep the window broadcast in there or not.

peterbell10:
The broadcast is fine as I think it only runs when there's a desync.

You will also need to change `DistributeStack` as well to handle shift clicking into window.

tympaniplayer:
Cheers. I'll make sure `DistrubuteStack` is working. I'll check the other click methods as well just in case.

src/UI/SlotArea.cpp
```diff
@@ -1484,17 +1484,17 @@ void cSlotAreaEnchanting::Clicked(cPlayer & a_Player, int a_SlotNum, eClickActio
 	cItem & DraggingItem = a_Player.GetDraggingItem();
 	if (a_SlotNum == 1)
 	{
-		// Lapis slot can have a full stack handle it normally
-		if ((DraggingItem.m_ItemType != E_ITEM_DYE) || (DraggingItem.m_ItemDamage != E_META_DYE_BLUE))
+		// Lapis slot can have a full stack handle it normally, also check for empty hand
+		if ((DraggingItem.m_ItemType <= 0) || (DraggingItem.m_ItemType == E_ITEM_DYE) && (DraggingItem.m_ItemDamage == E_META_DYE_BLUE))
```
peterbell10: I would prefer `DraggingItem.IsEmpty()` for clarity's sake.

src/UI/SlotArea.cpp
```diff
@@ -1535,6 +1535,28 @@ void cSlotAreaEnchanting::Clicked(cPlayer & a_Player, int a_SlotNum, eClickActio
 
 void cSlotAreaEnchanting::DistributeStack(cItem & a_ItemStack, cPlayer & a_Player, bool a_Apply, bool a_KeepEmptySlots, bool a_BackFill)
 {
+	if ((a_ItemStack.m_ItemType == E_ITEM_DYE) || (a_ItemStack.m_ItemDamage == E_META_DYE_BLUE))
```
peterbell10: This should be `&&` not `||`

src/UI/SlotArea.cpp
```diff
@@ -1482,7 +1482,21 @@ void cSlotAreaEnchanting::Clicked(cPlayer & a_Player, int a_SlotNum, eClickActio
 		bAsync = true;
 	}
 	cItem & DraggingItem = a_Player.GetDraggingItem();
+	if (a_SlotNum == 1)
+	{
+		// Lapis slot can have a full stack handle it normally, also check for empty hand
+		if ((DraggingItem.m_ItemType <= 0) || ((DraggingItem.m_ItemType == E_ITEM_DYE) && (DraggingItem.m_ItemDamage == E_META_DYE_BLUE)))
```
peterbell10: Since my other comment seems to be hidden, I'll duplicate it here:

I would prefer `DraggingItem.IsEmpty()` for clarity's sake.

tympaniplayer:
Hi @peterbell10, I changed it to use `DraggingItem.IsEmpty()` instead. Are there any other changes you would like me to make?

Issue #4294, opened 2018-09-03
Make it possible to disable/enable Burn in daylight for Monsters


yukiisbored:
I think it would be nice for an Apocalyptic-themed server to be able to disable/enable Burn in daylight for Monsters instead of being forced to have Vanilla gameplay.

The closest thing we can get is using [`SetIsFireproof`][3] which is possible from the Lua API but this results in the AI still using the "I'm burning in daylight" AI and visually the entity is still on fire. It also means the entity is immune to fire attacks including ones that are done by the players.

In Bukkit, it is possible, thanks to [`EntityCombustEvent`][1], to "deny" the combustion event, may be we can also implement that but I think making [`SetBurnsInDaylight`][2] accessible from the Lua API and adding an option for it on `monsters.ini` is enough.

[1]: https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityCombustEvent.html
[2]: https://github.com/cuberite/cuberite/blob/master/src/Mobs/Monster.h#L140
[3]:https://api.cuberite.org/cMonster.html#SetIsFireproof

PR #4295, opened 2018-09-03
Add BurnsInDaylight to Lua API and Monsters.ini

yukiisbored:
This commit also adds BurnsInDaylight to check if the Monster burns in daylight
or not.

Closes https://github.com/cuberite/cuberite/issues/4294

### DIFF
--- a/src/Mobs/Monster.h
+++ b/src/Mobs/Monster.h
@@ -137,7 +137,7 @@ public:
 	void ResetAttackCooldown();
 
 	void SetBurnsInDaylight(bool a_BurnsInDaylight) { m_BurnsInDaylight = a_BurnsInDaylight; }  // tolua_export
-	const bool BurnsInDaylight() { return m_BurnsInDaylight; }  // tolua_export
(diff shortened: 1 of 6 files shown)


src/Mobs/Monster.h
```diff
@@ -136,8 +136,8 @@ class cMonster :
 	void SetCanPickUpLoot(bool a_CanPickUpLoot) { m_CanPickUpLoot = a_CanPickUpLoot; }
 	void ResetAttackCooldown();
 
-	/** Sets whether the mob burns in daylight. Only evaluated at next burn-decision tick */
-	void SetBurnsInDaylight(bool a_BurnsInDaylight) { m_BurnsInDaylight = a_BurnsInDaylight; }
+	void SetBurnsInDaylight(bool a_BurnsInDaylight) { m_BurnsInDaylight = a_BurnsInDaylight; }  // tolua_export
+	bool BurnsInDaylight() { return m_BurnsInDaylight; }  // tolua_export
```
madmaxoft: This function should probably be `const`.

PR #4300, opened 2018-10-01
Don't reset speed when taking damage

### LINKED ISSUE #4062: Velocity dead-stop
Client version:   1.12.2
Server OS:        Windows 10 64-bit

Jump as usual no matter velocity and land orientation
### Expected behavior

Lands and cancels jump
### Actual behavior

Jump from high area, hold jump button until you hit the ground.

Video of the occurrence: https://youtu.be/OV-1N5fz5qM

Bond-009:
Fixes  #4062

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1000,10 +1000,6 @@ void cPlayer::ApplyArmorDamage(int a_DamageBlocked)
 
 bool cPlayer::DoTakeDamage(TakeDamageInfo & a_TDI)
 {
-	SetSpeed(0, 0, 0);
-	// Prevents knocking the player in the wrong direction due to
-	// the speed vector problems, see #2865
-	// In the future, the speed vector should be fixed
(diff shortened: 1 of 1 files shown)


src/Entities/Player.cpp
```diff
@@ -1000,10 +1000,6 @@ void cPlayer::ApplyArmorDamage(int a_DamageBlocked)
 
 bool cPlayer::DoTakeDamage(TakeDamageInfo & a_TDI)
 {
-	SetSpeed(0, 0, 0);
-	// Prevents knocking the player in the wrong direction due to
-	// the speed vector problems, see #2865
-	// In the future, the speed vector should be fixed
```
peterbell10: The issue described in this comment is now back.

Bond-009: Is resetting the speed when knockback != 0 an acceptable solution?

peterbell10: Honestly, I'm not sure.  You'll just have to test it out.

PR #4310, opened 2019-01-22
1.13: BlockTypeRegistry

madmaxoft:
This is the first step towards 1.13 compatibility, implementing the registry of all BlockTypes.

(ref: https://forum.cuberite.org/thread-3215-post-29239.html#pid29239 )

The first commit only shows the API and documentation, after discussion I'll implement the actual block type registrations using the existing handler code.

The main points I took into consideration while writing this code:
 - later on plugins will be allowed to register block types (perhaps even the vanilla blocks will be in a plugin)
 - extensible "hint" system, so that blocks may hint at some useful property (`IsSnowable`, `PipeVolume`), either fixed, or based on the BlockState (`LightValue` for redstone lamps)
 - reuse existing `cBlockHandler` code (for now)

~~The registry is not currently plugged anywhere in the executable, it is expected that it will be accessible through `cRoot` once it is finalized.~~ The Registry is now available through `cRoot` and initialized, but not used yet.


src/BlockTypeRegistry.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// BlockTypeRegistry:
+
+void BlockTypeRegistry::registerBlockType(
+	const AString & aPluginName,
+	const AString & aBlockTypeName,
+	std::shared_ptr<cBlockHandler> aHandler,
+	const std::map<AString, AString> & aHints,
+	const std::map<AString, BlockInfo::HintCallback> & aHintCallbacks
```
NiLSPACE: Perhaps we should make the registering of hints a different method. That way a plugin can extend existing functionality without having to completely override a previous block registration.

src/BlockTypeRegistry.h
```diff
@@ -0,0 +1,158 @@
...
+
+/** Complete information about a single block type.
+The BlockTypeRegistry uses this structure to store the registered information. */
+class BlockInfo
+{
+public:
+
+	/** Callback is used to query block hints dynamically, based on the current BlockState.
+	Useful for example for redstone lamps that can be turned on or off. */
+	using HintCallback = std::function<AString(const AString & aTypeName, const BlockState & aBlockState)>;
```
NiLSPACE: Perhaps we should also provide the world and vector of the requested block?

src/BlockTypeRegistry.h
```diff
@@ -0,0 +1,158 @@
...
+		std::shared_ptr<cBlockHandler> aHandler,
+		const std::map<AString, AString> & aHints = std::map<AString, AString>(),
+		const std::map<AString, HintCallback> & aHintCallbacks = std::map<AString, HintCallback>()
+	);
+
+
+	/** Retrieves the value associated with the specified hint for this specific BlockTypeName and BlockState.
+	Queries callbacks first, then hints if a callback doesn't exist.
+	Returns an empty string if hint not found at all. */
+	AString hintValue(
```
NiLSPACE: It could be useful to have `hintValueI`, `hintValueD`, `hintValueF` methods like the `cIniFile` class.

src/BlockTypeRegistry.h
madmaxoft: Not really - that would mean that the hint would need to be queried for each block in the world separately, while right now it means that the hint can be queried once per BlockArea's palette.

src/BlockTypeRegistry.h
madmaxoft: Could be useful, although there are many more usages (`LightValue` hint will most likely be converted to a `UInt8`, ...)

src/BlockTypeRegistry.cpp
madmaxoft: Excellent idea, I'll add that.

src/BlockTypeRegistry.h
madmaxoft: I won't add it now, but it could be added later on.

src/BlockTypeRegistry.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+// BlockInfo:
+
+BlockInfo::BlockInfo(
+	const AString & aPluginName,
+	const AString & aBlockTypeName,
+	std::shared_ptr<cBlockHandler> aHandler,
+	const std::map<AString, AString> & aHints,
+	const std::map<AString, BlockInfo::HintCallback> & aHintCallbacks
```
madmaxoft: I'm conflicted whether to use `map` or `unordered_map` here. Considering the number of elements we expect this to have, I'd say it's an unnecessary detail right now.

src/BlockTypeRegistry.cpp
```diff
@@ -56,6 +56,24 @@ AString BlockInfo::hintValue(
...
+void BlockInfo::setHint(const AString & aHintKey, const AString & aHintValue)
+{
+	mHints[aHintKey] = aHintValue;
+}
+
+
+
+
+
+void BlockInfo::removeHint(const AString & aHintKey)
```
NiLSPACE: Should we allow a plugin to remove the hint of another plugin though?

madmaxoft:
I think this is now as ready as can be, without disturbing anything else. Can I get a review?

src/BlockTypeRegistry.cpp
```diff
@@ -0,0 +1,229 @@
...
+	// Nothing found, return empty string:
+	return AString();
+}
+
+
+
+
+
+void BlockInfo::setHint(const AString & aHintKey, const AString & aHintValue)
+{
```
NiLSPACE: What if the hint was already set from a callback? Currently the value would be silently ignored when `hintValue` would be called as the callback variant is checked first.

madmaxoft: The callbacks are considered higher priority, so they "overwrite" anything set statically.

Yeah, I know, it's not 100% flexible, but it's good enough for our purposes and easy enough to both implement and understand.

madmaxoft: I've added mentions of this behavior in the doxy-comments.

NiLSPACE: Should we also print a message when `setHint` is called for something that is already done using a callback? Otherwise someone might be confused why his hint isn't being used. Or is that overkill?

madmaxoft: That would require us to call the callback to know whether it provides the value or not. Definitely overkill.

madmaxoft: Took a second look, a warning is a cheap way to improve later coding, so I added that.

Seadragon91:
Compiling fails under vs 2019, with multiple errors of `functional is not a member of std`. Adding `#include <functional>` to Globals.h fixed that error.

madmaxoft:
Thanks for the report, fixed directly in master

PR #4315, opened 2019-02-19
Limit the size of cListAllocationPool's free list

### LINKED ISSUE #4314: Memory not freeing
Client version:   1.12.2
Server OS:        Linux
Commit id:        6c3dd41717664c75bf77392dd4a4df3b35678fa5

### Expected behavior

Memory should be freed when chunks unload.

### Actual behavior

Memory only increasing and unloading chunks doesn't free any memory.

Download Linux version of Cuberite.
Fly over the world.
Watch memory increasing and not freeing when chunks unload.

peterbell10:
Fixes #4314 

Since #4260, most chunk sections are actually allocated in the WorldStorage and ChunkGenerator threads which don't have access to the world's allocation pool.  However, when the chunks are freed they are kept in the world's free lists.  This asymmetry means the free list will essentially grow unbounded and the memory isn't actually reused. The exception being the few occasions when chunk sections are allocated from the world tick thread directly.

This just limits the max size of the free list. It might be better if the Generator and Storage threads could use the same allocation pool but that would mean making `cListAllocationPool` thread safe.

At this point it might make more sense to just remove `cAllocationPool` for good.  It seems very unlikely that cuberite could handle 'out of memory' gracefully.

### DIFF
--- a/src/AllocationPool.h
+++ b/src/AllocationPool.h
@@ -11,18 +11,18 @@ class cAllocationPool
 public:
 	class cStarvationCallbacks
 	{
-		public:
-			virtual ~cStarvationCallbacks() {}
+	public:
+		virtual ~cStarvationCallbacks() {}
 
-			/** Is called when the reserve buffer starts to be used */
-			virtual void OnStartUsingReserve() = 0;
+		/** Is called when the reserve buffer starts to be used */
+		virtual void OnStartUsingReserve() = 0;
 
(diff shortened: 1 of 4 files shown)


madmaxoft:
With the 1.13 changes we already will have to phase the allocation pool out, since the chunks will no longer be of the same size. So yes, in the long run I agree with the removal, but for now let's just fix this.

PR #4321, opened 2019-03-28
Respect return value of cLuaWindow's OnClicked handler

DaMatrix:
Current behavior treats the return value as `true` regardless, meaning any cLuaWindow with an OnClicked handler set will have an unusable inventory.


DaMatrix:
wait i think i messed this up, there's a lot more changes here than i wanted...

PR #4324, opened 2019-04-15
Allow chests to be opened with non-solid blocks on top of them

DaMatrix:
This allows chests to be opened when the block above them isn't a full block, preserving vanilla behavior (e.g. placing stairs on top of a chest doesn't block the chest)

### DIFF
--- a/src/BlockEntities/ChestEntity.cpp
+++ b/src/BlockEntities/ChestEntity.cpp
@@ -208,7 +208,7 @@ bool cChestEntity::IsBlocked()
 	return (
 		(GetPosY() < cChunkDef::Height - 1) &&
 		(
-			!cBlockInfo::IsTransparent(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
+			cBlockInfo::FullyOccupiesVoxel(GetWorld()->GetBlock(GetPosX(), GetPosY() + 1, GetPosZ())) ||
 			!cOcelot::IsCatSittingOnBlock(GetWorld(), Vector3d(GetPos()))
 		)
 	);


peterbell10:
Stairs already seem to work on the current master.  Can you give an example of what this would fix?

DaMatrix:
For example this chest:



would be openable in vanilla, but is unopenable in the current version of Cuberite from what I saw while playing around with it earlier, and works fine with this. I didn't mean to mention stairs, as they're already transparent, this is mainly about slabs.

peterbell10:
Okay, I've had a look at the cases where `IsTransparent == FullyOccupiesVoxels` and it seems to give a mix of right and wrong behaviours.

| Block ID | Blocks Chest | Matches vanilla | Block Name                |
|----------|--------------|-----------------|---------------------------|
| 43       | ☑️            | ☑️               | Double stone slab         |
| 44       | ❎            | ☑️               | Stone slab                |
| 46       | ❎            | ☑️               | TNT                       |
| 52       | ☑️            | ☑️               | Mob Spawner               |
| 74       | ☑️            | ❎               | Glowing Redstone Ore      |
| 79       | ☑️            | ❎               | Ice                       |
| 80       | ❎            | ❎               | Snow block                |
| 88       | ❎            | ❎               | Soul sand                 |
| 125      | ☑️            | ☑️               | Double wooden slab        |
| 152      | ☑️            | ❎               | Block of redstone         |
| 165      | ❎            | ❎               | Slime block               |
| 166      | ☑️            | ☑️               | Barrier                   |
| 169      | ❎            | ☑️               | Sea lantern               |
| 178      | ❎            | ☑️               | Inverted daylight sensor  |
| 181      | ☑️            | ☑️               | Double red sandstone slab |
| 198      | ❎            | ☑️               | End rod                   |
| 204      | ☑️            | ☑️               | Purpur double slab        |
| 209      | ❎            | ☑️               | End gateway               |
| 232      | ❎            | ☑️               | Green shulker box         |
| 249      | ❎            | ❎               | Red glazed terracotta

peterbell10:
I also seem to have found a bug since green shulker boxes and red glazed terracotta shouldn't be behaving differently from the other colours.

Bond-009:
According to the wiki chest can't be opened when there is an opaque block above.
Ref: https://minecraft.gamepedia.com/Chest#Container
And single slabs should be partially transparent.
Ref: https://minecraft.gamepedia.com/Slab
The problem is that we don't have partially transparent blocks in Cuberite, it's either one or the other, nothing in between.

peterbell10:
I wonder if it's even possible to categorise the partially transparent blocks?  It does effect a lot of different mechanics: https://minecraft.gamepedia.com/Opacity

PR #4325, opened 2019-04-15
Make Nether portals spawn the player inside of them

DaMatrix:
Currently the player is spawned immediately in front of them. 

Simply changing `cNetherPortalScanner::OutOffset` to `0.5` wasn't enough, as the player would always be spawned on top of the portal, however checking for non-solid blocks instead of air fixes this.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1546,9 +1546,6 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	// Take note of old chunk coords
 	auto OldChunkCoords = cChunkDef::BlockToChunk(GetPosition());
 
-	// Prevent the entity from teleporting back immediately, in case the destination is a portal
-	m_PortalCooldownData.m_ShouldPreventTeleportation = true;
-
 	// Set position to the new position
 	ResetPosition(a_NewPosition);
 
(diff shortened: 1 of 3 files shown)


src/Entities/Entity.cpp
```diff
@@ -1546,6 +1546,9 @@ bool cEntity::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 	// Take note of old chunk coords
 	auto OldChunkCoords = cChunkDef::BlockToChunk(GetPosition());
 
+	// Prevent the entity from teleporting back immediately, in case the destination is a portal
+	m_PortalCooldownData.m_ShouldPreventTeleportation = true;
```
peterbell10: This should not be needed, it should already be set when the entity is detected inside of the portal.

src/Entities/Player.cpp
```diff
@@ -2072,6 +2072,9 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 		// Broadcast the player into the new world.
 		a_World->BroadcastSpawnEntity(*this);
 
+		// Prevent the player from teleporting back immediately, in case the destination is a portal
+		m_PortalCooldownData.m_ShouldPreventTeleportation = true;
```
peterbell10: Same here

src/Entities/Entity.cpp
DaMatrix: Ah, now that you mention it I'd added those while trying to figure out why the player was spawning on top of the portals and then didn't realize it could be safely removed now (which I just tested and it works fine).

DaMatrix:
I seem to have accidenally tracked some stuff from my local branch in here... one moment

PR #4328, opened 2019-04-16
Fix player skin sending through BungeeCord

DaMatrix:
This fixes sending of player skins through BungeeCord by actually parsing the JSON instead of setting the player properties as a string.

### DIFF
--- a/src/Protocol/Protocol_1_9.cpp
+++ b/src/Protocol/Protocol_1_9.cpp
@@ -154,7 +154,16 @@ cProtocol_1_9_0::cProtocol_1_9_0(cClientHandle * a_Client, const AString & a_Ser
 				UUID.FromString(Params[2]);
 				m_Client->SetUUID(UUID);
 
-				m_Client->SetProperties(Params[3]);
+                Json::Value root;
+                Json::Reader reader;
+                if (!reader.parse(Params[3], root)) {
+                    LOGERROR("Unable to parse player properties: '%s'", Params[3]);
+                } else {
+                    m_Client->SetProperties(root);
+                    LOGD("ipstring: %s", Params[1]);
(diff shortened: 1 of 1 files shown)


src/Protocol/Protocol_1_9.cpp
```diff
@@ -973,11 +978,14 @@ void cProtocol_1_9_0::SendPlayerListAddPlayer(const cPlayer & a_Player)
 	Pkt.WriteVarInt32(1);
 	Pkt.WriteUUID(a_Player.GetUUID());
 	Pkt.WriteString(a_Player.GetPlayerListName());
+	LOGD("Sending player tab list for \"%s\"...", a_Player.GetPlayerListName());
```
peterbell10: This functions is called once for every player in the server, that's why I think these `LOGD`s would be too spammy.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -978,6 +983,8 @@ void cProtocol_1_9_0::SendPlayerListAddPlayer(const cPlayer & a_Player)
 	Pkt.WriteVarInt32(Properties.size());
 	for (auto & Node : Properties)
 	{
+		LOGD("name: %s", Node.get("name", "").asString());
```
peterbell10:
```suggestion
```

src/Protocol/Protocol_1_9.cpp
```diff
@@ -978,6 +983,8 @@ void cProtocol_1_9_0::SendPlayerListAddPlayer(const cPlayer & a_Player)
...
+		LOGD("value: %s", Node.get("value", "").asString());
```
peterbell10:
```suggestion
```

PR #4329, opened 2019-04-17
Enchancing mob API

Buom01:
This change add :SetStatic(Boolean) :IsStatic() to all enity.
That's useful for controlling them totally with a plugin and making them unannimed (
Moreover I expose GetAttached() AttachTo(Entity) Detach() IsAttachedTo(Entity) to Lua API

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -13,6 +13,7 @@ beeduck
 bibo38
 birkett (Anthony Birkett)
 Bond_009
+Buom_01
 changyongGuo
 Cocosushi6
 derouinw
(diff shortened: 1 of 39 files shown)


NiLSPACE:
Hi!

For a long time we planned to make the mobs modular which would make it really easy to let plugins take control of mobs (See for example #4268 and https://forum.cuberite.org/thread-820.html?highlight=mob+modular). The result would be that we won't have to put ifs in every single function to make sure it's static. Though, as you might see, we've been planning it since 2013 :/

I believe the reason why all the `Attached` functions aren't exported is because it's not safe. If Lua tries to manipulate an attached entity it might not exist any more. That's why we use callbacks to manipulate entities, as they lock the world to make sure nothing can change the current state of the entity. It would probably be safe to return the entity id though.

Buom01:
I admit that as you said, the code of my pull request is bad quality and it was not very thoughtful.

Concerning the `Attached` functions, I begin to understand how works Cuberite.

I will not give up so soon

PR #4330, opened 2019-04-19
Armor stand + broadcast player's left hand

Buom01:
Info: Since I has a one month interruption, I will not continue to enhance too much and correct the pull before the issue https://github.com/cuberite/cuberite/issues/3696 is solved. Sorry but no utility to got a mob if it bug or not showing up.
I made a flag using armor stand properties, but if only half of it is showing randomly.....

---

Hi. Finally, I have developed armor stand because almost every minecraft server use it and Cuberite still haven't it. However, thanks to variables about armor stand that was already defined in protocols and helped me a lot.

What I have tested both on 1.12 and 1.8 ?
- Putting armor stand. Direction facing to player (rounded to 45 degree aka Pi/4 radian).
- Appearance of the wore armor
- Destroying the armor and dropped items

What I have tested on 1.12 ?
- Sizes (maker, normal, small)
- Rotation of the head, but all parts should be working

What should be ready ?
- Using it as marker (aka hologram) [Plugin provided](https://github.com/Buom01/StaticHolograms)
- Using it as normal
- Should works from imported map

Are there problems ?
- ~~The OnRightClick event never fire when used with armor part, but works with tag. If you know why let me know !~~ Solved !
- We can put multiple armor stand at the same place
- Custom name must be set to always visible to be visible (else it never show, even when looking to it). Is it the behaviour of minecraft vanilla ?
- ~~Moving armor stand in server side doesn't reflect immediately to the client correctly~~
- (ArmorStand inherits from entity but it almost don't use its property like IsOnFire. It can't event send it to client cause it have custom Flags integration)

Whats doesn't works ?
- ~~Players can't add armor on it (bug cited in problems)~~

Differences with vanilla ?
- ~~Armor stands are not saved like vanilla. I'm needing infos to the original minecraft format for proper integration.~~
- ~~Hitting armor stand make it red like hurted animals when it is empty, and don't shake when release part of armor~~
- ~~It don't release what was shot but it release armor in a defined order~~
- Dropped items probably don't drop at the same place as vanilla
- ~~Only no gravity mode. On vanilla player placed armor stand have gravity, but it was send by the server that it has gravity at the opposite of the current implantation wich always tell the client that armor stand haven't gravity.~~ Gravity works with random values, update if you know rights one.
- Shooting armor stand with bow and arrow in creative still give drops

Things that probably need review or better coding ?
- ~~Is it OK to call BroadcastMovementUpdate at every Tick ?~~
- ~~probably cWSSAnvil::LoadArmorStandFromNBT~~
- ~~cArmorStand::OnRightClicked is a big block of code too~~
- ~~cArmorStand::KilledBy seem to be good enough~~

What's news ?
- OnClickedAt implented next to OnRightClicked
- Little support for left hand, just send the dedicated packet for mob

That's it.
Tell me your suggestions,
If I can also help on other part of Cuberite (acording to my skills cause I'm more a web dev), you can always request me


src/Entities/ArmorStand.cpp
```diff
@@ -0,0 +1,461 @@
...
+		SetInvulnerableTicks(0);
+		GetWorld()->BroadcastEntityMetadata(*this);
+	}
+}
+
+
+
+
+
+void cArmorStand::OnRightClicked(cPlayer & a_Player)  // Only works for tags, else the callback is not called, IDK Why ...
```
peterbell10: I have figured out why your right click handler isn't being called. The armour stand has its own variant of the use entity packet which isn't handled by the protocol yet.
https://github.com/cuberite/cuberite/blob/8b4d37771f6948f57f9e2d6359c574c2b411cc05/src/Protocol/Protocol_1_9.cpp#L2787-L2793

src/Entities/ArmorStand.cpp
```diff
@@ -0,0 +1,461 @@
...
+		return;
+	}
+
+	cItems Pickup;
+	if (!m_LeftHand.IsEmpty())
+	{
+		Pickup.push_back(m_LeftHand);
+		SetOffHandEquipedItem(cItem());
+	}
+	else if (!m_RightHand.IsEmpty())
```
peterbell10: Why are these conditions all `else`? Surely the armour stand should drop all of its armour pieces, not just the first one.

src/Entities/ArmorStand.cpp
```diff
@@ -0,0 +1,461 @@
...
+#include "../ClientHandle.h"
+#include "../Chunk.h"
+#include "../World.h"
+
+
+
+
+
+cArmorStand::cArmorStand(Vector3d a_Pos, double a_Yaw, short a_Size):
+	cEntity(etArmorStand, a_Pos.x, a_Pos.y, a_Pos.z, 0.25*a_Size, 0.9875*a_Size),  // a_Size = 0 : Marker ;  a_Size = 1 : Small ;  a_Size = 0 : Normal
```
peterbell10: What exactly is marker and why does it is controlled by `m_Size`? Would a small marker not be a valid use case?

Buom01: According to:
https://wiki.vg/Entity_metadata#Objects
https://minecraft.gamepedia.com/Armor_Stand

Marker state of an armour stand highly depend of size. But I think that I will prefere two arguments bool a_Small and bool a_Marker. That's smaller in term and ram and it will be more clear to understand

src/Entities/ArmorStand.cpp
Buom01: Yes, I don't know with which gameplay I hit the part of the armor stand to get it back individually, but you are right (tested in 1.13.2); All fall suddenly

src/Entities/ArmorStand.cpp
peterbell10: Thanks for the wiki link but I'm not sure if "size" there is only referring to the collision volume. Based off the fact that they're both booleans, it sounds like it should be possible to have normal and small markers though.

src/Entities/ArmorStand.cpp
```diff
@@ -0,0 +1,461 @@
...
+		return;
+	}
+
+	if (a_Player.GetEquippedItem().IsEmpty())
+	{
+		return;
+	}
+	short ItemType = a_Player.GetEquippedItem().m_ItemType;
+
+	if ((ItemType == E_ITEM_NAME_TAG) && !a_Player.GetEquippedItem().m_CustomName.empty())  // Add a name to the armor stand
```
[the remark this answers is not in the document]
Buom01: I will take care of it

src/Entities/ArmorStand.cpp
Buom01: Thanks a lot ! I had solved it locally, you will see it nextly

src/Entities/ArmorStand.cpp
Buom01: Then I will separate m_Size to m_Marker and m_Small.
But isn't enough to have setHeight and setWidth available separatly ?
Because the wiki define size according to the type of armor stand is it.
At the other hand I agree that is not a clear definition of what size should be; it's not defined bidirectionally

Extract:
> Marker: 1 or 0 (true/false) - if true, ArmorStand's size will be set to 0, have a tiny hitbox and disable interactions with it. May not exist.

> Name | bounding box x and z | bounding box y
> ArmorStand | normal: 0.5 marker: 0.0 small: 0.25 | normal: 1.975 marker: 0.0 small: 0.9875

peterbell10: >or should I simply replace m_Size by m_Marker and m_Small ?

I think so, although I would call them `m_IsMarker` and `m_IsSmall`.

Buom01:
I don't understand why I don't have similar error that Travis's Build
How to reproduce locally ?

src/Entities/ArmorStand.cpp
Buom01: Done

peterbell10:
>How to reproduce locally ?

Some of the build errors will be specific to clang which generally has more warnings than gcc (We use 5.0).

For the documentation errors, you need to run the APIDump plugin which is described in our [lua exporting docs](https://github.com/cuberite/cuberite/blob/master/dev-docs/ExportingAPI.html). TLDR, from within cuberite run these commands
```shell
load APIDump
checkapi
```

src/ClientHandle.h
```diff
@@ -369,6 +369,7 @@ class cClientHandle  // tolua_export
 	);
 	void HandleUnmount          (void);
 	void HandleUseEntity        (UInt32 a_TargetEntityID, bool a_IsLeftClick);
+	void HandleUseEntityAt      (UInt32 a_TargetEntityID, Vector3f a_TargetPos, bool a_IsLeftClick);
```
peterbell10: `a_IsLeftClick` is a really confusing name since you can right click with the item in you left hand. I think `eHand` would be much better than a `bool` here.

peterbell10:
Also, for some reason I can't add boots to the armour stand before adding another item first.

src/Entities/ArmorStand.cpp
```diff
@@ -0,0 +1,528 @@
...
+	m_IsMarker(false),
+	m_HasArms(false),
+	m_HasBasePlate(true),
+	m_HeadRotation(0.0, 0.0, 0.0),
+	m_BodyRotation(0.0, 0.0, 0.0),
+	m_LeftArmRotation(-10.0, 0.0, -10.0),
+	m_RightArmRotation(-15.0, 0.0, 10.0),
+	m_LeftLegRotation(-1.0, 0.0, -1.0),
+	m_RightLegRotation(1.0, 0.0, 1.0),
+	m_LeftHand(cItem()),
```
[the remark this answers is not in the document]
peterbell10: Just leave out the initialiser completely and it will be a default constructed item.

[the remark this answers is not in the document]
Buom01: Done.
However I notice that code could be updated like that, like ItemFrame

src/Protocol/Protocol_1_10.cpp
```diff
@@ -558,6 +558,83 @@ void cProtocol_1_10_0::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity &
...
+			}
+			if (ArmorStand.IsMarker())
+			{
+				Flags |= 0x10;
+			}
+			a_Pkt.WriteBEUInt8(ARMOR_STAND_STATUS);  // Index
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BYTE);  // Type
+			a_Pkt.WriteBEInt8(Flags);
+
+			if (!ArmorStand.IsInvisible())  // It will probably don't change anything if it's invisible because the CustomName will probably don't be different. So save a lot of packets.
```
peterbell10:
```suggestion
			if (!ArmorStand.IsInvisible())  // No point sending orientation if it's invisible
```

src/Protocol/Protocol_1_11.cpp
```diff
@@ -796,6 +796,83 @@ void cProtocol_1_11_0::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity &
...
+			}
+			if (ArmorStand.IsMarker())
+			{
+				Flags |= 0x10;
+			}
+			a_Pkt.WriteBEUInt8(ARMOR_STAND_STATUS);  // Index
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BYTE);  // Type
+			a_Pkt.WriteBEInt8(Flags);
+
+			if (!ArmorStand.IsInvisible())  // It will probably don't change anything if it's invisible because the CustomName will probably don't be different. So save a lot of packets.
```
peterbell10:
```suggestion
			if (!ArmorStand.IsInvisible())  // No point sending orientation if it's invisible
```

src/Protocol/Protocol_1_12.cpp
```diff
@@ -611,6 +611,83 @@ void cProtocol_1_12::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_
...
+			}
+			if (ArmorStand.IsMarker())
+			{
+				Flags |= 0x10;
+			}
+			a_Pkt.WriteBEUInt8(ARMOR_STAND_STATUS);  // Index
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BYTE);  // Type
+			a_Pkt.WriteBEInt8(Flags);
+
+			if (!ArmorStand.IsInvisible())  // It will probably don't change anything if it's invisible because the CustomName will probably don't be different. So save a lot of packets.
```
peterbell10:
```suggestion
			if (!ArmorStand.IsInvisible())  // No point sending orientation if it's invisible
```

src/Protocol/Protocol_1_9.cpp
```diff
@@ -455,8 +455,13 @@ void cProtocol_1_9_0::SendEntityEquipment(const cEntity & a_Entity, short a_Slot
...
 	cPacketizer Pkt(*this, GetPacketId(sendEntityEquipment));  // Entity Equipment packet
 	Pkt.WriteVarInt32(a_Entity.GetUniqueID());
-	// Needs to be adjusted due to the insertion of offhand at slot 1
-	if (a_SlotNum > 0)
+	// See https://wiki.vg/Protocol#Entity_Equipment
+	// TODO: Enable player left hand. Currently only user by armor stand.
+	if (a_SlotNum == 5)  // Left hand
+	{
+		a_SlotNum = 1;
+	}
```
peterbell10: If you aren't using this anymore then it should be removed.

src/World.h
```diff
@@ -461,6 +462,10 @@ class cWorld  // tolua_export
 	Returns the UniqueID of the spawned boat, or cEntity::INVALID_ID on failure. */
 	UInt32 SpawnBoat(Vector3d a_Pos, cBoat::eMaterial a_Material);
 
+	/** Spawns an armor stand at the given coordinates.
+	Returns the UniqueID of the spawned armor stand, or cEntity::INVALID_ID on failure. */
+	UInt32 SpawnArmorStand(Vector3d a_Pos, double a_Yaw);
```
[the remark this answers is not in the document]
peterbell10: You still haven't documented this. I would expect something like
```cpp
\param a_Yaw Facing direction in degrees from south, going clockwise
```

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -887,6 +888,82 @@ void cNBTChunkSerializer::AddItemFrameEntity(cItemFrame * a_ItemFrame)
...
+			m_Writer.AddDouble("", a_ArmorStand->GetLeftLegRotation().x);
+			m_Writer.AddDouble("", a_ArmorStand->GetLeftLegRotation().y);
+			m_Writer.AddDouble("", a_ArmorStand->GetLeftLegRotation().z);
+		m_Writer.EndList();
+		m_Writer.BeginList("RightLegRotation", TAG_Double);
+			m_Writer.AddDouble("", a_ArmorStand->GetRightLegRotation().x);
+			m_Writer.AddDouble("", a_ArmorStand->GetRightLegRotation().y);
+			m_Writer.AddDouble("", a_ArmorStand->GetRightLegRotation().z);
+		m_Writer.EndList();
+		m_Writer.BeginList("Items", TAG_Compound);
```
peterbell10: The wiki says this is split into "HandItems" and "ArmorItems"
https://minecraft.gamepedia.com/Armor_Stand#Data_values

src/World.h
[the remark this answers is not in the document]
Buom01: Confirmed and updated

src/Entities/ArmorStand.h
```diff
@@ -0,0 +1,140 @@
...
+
+	/** Is the custom name of this armor stand always visible? If not, you only see the name when you sight the mob. */
+	bool IsCustomNameAlwaysVisible(void) const { return m_CustomNameAlwaysVisible; }
+
+	/** Sets the custom name visiblity of this armor stand.
+	If it's false, you only see the name when you sight the mob. If it's true, you always see the custom name. */
+	void SetCustomNameAlwaysVisible(bool a_CustomNameAlwaysVisible);
+
+	/** Returns true if the armor stand has gravity. */
+	bool HasGravity(void) const { return (m_Gravity < 0.0f) || (m_Gravity > 0.0f); }
```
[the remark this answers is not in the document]
Buom01: Should I define `FLT_EPSILON` in `Globals.h` or prefer to define it at the beginning of the file ?

src/Protocol/Protocol_1_9.cpp
Buom01: As said in the comment, armor stand use it
However it's not used by player's class

```
void cArmorStand::SetOffHandEquipedItem(const cItem & a_LeftHand)
{
	m_LeftHand = a_LeftHand;
	if (m_World != nullptr)
	{
		m_World->BroadcastEntityEquipment(*this, 5, a_LeftHand);  // See above (In SpawnOn) about the slot ID
	}
}
```
Moreover, for compatibility issues, left hand was added at the end and not at the ID 1 like wants the protocol since 1.9
This ways, previous versions keep equipment working

src/ClientHandle.cpp
```diff
@@ -1742,6 +1742,32 @@ void cClientHandle::HandleUseEntity(UInt32 a_TargetEntityID, bool a_IsLeftClick)
 
 
 
+void cClientHandle::HandleUseEntityAt(UInt32 a_TargetEntityID, Vector3f a_TargetPos, eHand a_Hand)
+{
+	// TODO: Let plugins interfere via a hook
+	// TODO: Take care of TargetPos position
```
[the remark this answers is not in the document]
peterbell10: Why do you keep marking comments as resolved when you haven't addressed them?

src/Entities/ArmorStand.h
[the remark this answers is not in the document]
peterbell10: `FLT_EPSILON` is defined in [`<climits>`](https://en.cppreference.com/w/cpp/types/climits) which I think is probably already included from `"Globals.h"`

src/Protocol/Protocol_1_9.cpp
peterbell10: Okay, so why have you disabled it in `cPlayer` then?

src/WorldStorage/NBTChunkSerializer.cpp
Buom01: I will take care of it

src/Entities/ArmorStand.h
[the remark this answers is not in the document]
Buom01: Then my code seem OK
`bool HasGravity(void) const { return (std::abs(m_Gravity) > std::numeric_limits<float>::epsilon(); }`

I implanted it like it's in some other cuberite's file

src/ClientHandle.cpp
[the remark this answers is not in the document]
Buom01: Sorry I just seen your question now.
I mark them resolved when I have made modifications and committed them, not necessarily when I had pushed them. It permit me to see what's not done or got stuck.

Should I keep all of them open ?

src/WorldStorage/NBTChunkSerializer.cpp
Buom01: Should be solved, I then updated a part of the code to make it vanilla compliant

src/Protocol/Protocol_1_9.cpp
Buom01: I just add the line `// a_Client.SendEntityEquipment(*this, 5, m_Inventory.GetOffHandEquipedItem());  // Use number 5 for compatibility issues - Not activated since left hand is not now managed !`

And like said, I haven't enabled it because `GetOffHandEquipedItem` doesn't exist
I research again, and I now found that's because it's called `GetShieldSlot`

Then I will update the code to make the shield visible by everyone, but I should test it before commit

Buom01:
I think that it's (almost?) finished

src/ClientHandle.cpp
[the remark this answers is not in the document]
peterbell10: I would prefer that you wait until the code had been pushed. That way I can see that it has actually been resolved.

src/Protocol/Protocol_1_9.cpp
peterbell10: Does this work with a 1.8 client? I think you might need to add 
```cpp
if (a_SlotNum == 5)
{
	return;
}
```

_Before_ the `cPacketizer` is constructed to avoid sending items to an invalid slot.

Buom01: I confirm that it's doesn't make a bad behavior on 1.8 except that the player can't see other's left hand's content.
I added your part of code as suggested to avoid a noise.

src/Protocol/Protocol_1_10.cpp
```diff
@@ -558,6 +558,83 @@ void cProtocol_1_10_0::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity &
...
+			}
+			if (ArmorStand.IsMarker())
+			{
+				Flags |= 0x10;
+			}
+			a_Pkt.WriteBEUInt8(ARMOR_STAND_STATUS);  // Index
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BYTE);  // Type
+			a_Pkt.WriteBEInt8(Flags);
+
+			if (!ArmorStand.IsMarker())  // No point sending orientation if it's a marker. However invisible armor stand keep armor visible.
```
peterbell10: Why would markers not need to broadcast their orientation? You can still see them and pose them.

Buom01: Like shown in https://youtu.be/Vgc_tZ2GHvg?t=67
Invisible armor stand are not so invisible (confusing tag name)
At the opposite, armor stand sets as marker are really and totally invisible. The only part still shown is the custom name if sets as always visible.

Then sending armor stand's body rotation is still useful if the armor stand has an armor.
However an armor stand sets as marker will not be shown at all, then theses packets are not useful in this case.

Else if I am wrong ?

peterbell10: When I try out your code, I can see the markers though? Either way, something is wrong here.

madmaxoft: This discussion shows that a proper comment in the code is needed to explain this.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9613,6 +9627,454 @@ a_Player:OpenWindow(Window);
...
+				},
+				HasCustomName =
+				{
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns true if the player has a custom name.",
```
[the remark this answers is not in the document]
madmaxoft: This still needs fixing?

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9613,6 +9627,454 @@ a_Player:OpenWindow(Window);
...
+				},
+				GetCustomName =
+				{
+					Returns =
+					{
+						{
+							Type = "string",
+						},
+					},
+					Notes = "Returns the custom name of this armor stand. If the player hasn't a custom name, it will return an empty string.",
```
madmaxoft: player -> armor stand again?

Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -3119,6 +3119,30 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
...
+						},
+					},
+					Returns =
+					{
+						{
+							Name = "EntityID",
+							Type = "number",
+						},
+					},
+					Desc = "Spawns an armor stand at the given coordinates and with the given orientation in degrees from south, going clockwises. Returns the UniqueID of the spawned armor stand, or {{cEntity#INVALID_ID|cEntity#INVALID_ID}} on failure.",
```
madmaxoft: Please don't use `south`, we prefer to name the directions according to the axes: `X+`, `Z-` etc. The Mojang definition of "south" has changed at least once in Minecraft history.

Server/Plugins/APIDump/Hooks/OnPlayerClickingAtEntity.lua
```diff
@@ -0,0 +1,23 @@
...
+		DefaultFnName = "OnPlayerClickingAtEntity",  -- also used as pagename
+		Desc = [[
+			This hook is called when the {{cPlayer|player}} right-clicks an {{cEntity|entity}}. Plugins may
+			override the default behavior or even cancel the default processing.
+		]],
+		Params =
+		{
+			{ Name = "Player", Type = "{{cPlayer}}", Notes = "The player who has right-clicked the entity" },
+			{ Name = "Entity", Type = "{{cEntity}} descendant", Notes = "The entity that has been right-clicked" },
+			{ Name = "TargetPosition", Type = "{{Vector3f}}", Notes = "The location where the player hit" },
```
madmaxoft: `hit` -> `clicked`

src/Bindings/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ set(BINDING_DEPENDENCIES
 	../Entities/Pawn.h
 	../Entities/Player.h
 	../Entities/Painting.h
+	../Entities/ArmorStand.h
```
madmaxoft: Please alpha-sort (and possibly fix the ArrowEntity out of order, too)

src/Bindings/Plugin.h
```diff
@@ -86,6 +86,7 @@ class cPlugin
 	virtual bool OnPlayerPlacingBlock       (cPlayer & a_Player, const sSetBlock & a_BlockChange) = 0;
 	virtual bool OnPlayerRightClick         (cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ) = 0;
 	virtual bool OnPlayerRightClickingEntity(cPlayer & a_Player, cEntity & a_Entity) = 0;
+	virtual bool OnPlayerClickingAtEntity   (cPlayer & a_Player, cEntity & a_Entity, const Vector3f & a_TargetPos) = 0;
```
madmaxoft: Please alpha-sort.
Also, is this a right-click or left-click? Might want to rename.

src/Items/ItemHandler.cpp
```diff
@@ -153,6 +154,7 @@ cItemHandler * cItemHandler::CreateItemHandler(int a_ItemType)
 		case E_ITEM_HEAD:                return new cItemMobHeadHandler(a_ItemType);
 		case E_ITEM_SNOWBALL:            return new cItemSnowballHandler();
 		case E_ITEM_SPAWN_EGG:           return new cItemSpawnEggHandler(a_ItemType);
+		case E_ITEM_ARMOR_STAND:         return new cItemAmorStandHandler(a_ItemType);
```
madmaxoft: Alpha-sort

src/World.h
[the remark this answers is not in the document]
madmaxoft: Replace "south" with axis direction.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2011,6 +2014,212 @@ void cWSSAnvil::LoadPaintingFromNBT(cEntityList & a_Entities, const cParsedNBT &
...
+	std::unique_ptr<cArmorStand> ArmorStand = cpp14::make_unique<cArmorStand>(Vector3d(), 0.0);
+	if (!LoadEntityBaseFromNBT(*ArmorStand.get(), a_NBT, a_TagIdx))
+	{
+		return;
+	}
+
+
+	/**
+		Basic properties
+	*/
```
madmaxoft: Please don't use doxy-comments in regular code. Use a normal single-line `// Comment`

src/Bindings/Plugin.h
Buom01: No, I re-verified my code. It could be whatever according to https://wiki.vg/Protocol#Use_Entity
By default I assume it's a right click for the 1.8 protocol
And it parse which hand is it in 1.9 protocol

src/Entities/ArmorStand.h
```diff
@@ -0,0 +1,140 @@
...
+
+	/** Sets the custom name visiblity of this armor stand.
+	If it's false, you only see the name when you sight the mob. If it's true, you always see the custom name. */
+	void SetCustomNameAlwaysVisible(bool a_CustomNameAlwaysVisible);
+
+	/** Returns true if the armor stand has gravity. */
+	bool HasGravity(void) const { return (std::abs(m_Gravity) > std::numeric_limits<float>::epsilon()); }
+
+	/** Returns true if the armor stand is a normal sized one. */
+	bool IsNormal(void) const { return !m_IsSmall; }
```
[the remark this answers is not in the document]
Buom01: Done, but I kept variable like m_IsSmall as is. Just changed methods' name.

src/Protocol/Protocol_1_10.cpp
Buom01: I updated it, wait the push to see if it's OK

Buom01:
I think that it's almost OK. Eventually things that goes wrong about doc.

mathiascode:
I will test this in the beginning of next week.

src/Bindings/Plugin.h
```diff
@@ -73,6 +73,7 @@ class cPlugin
 	virtual bool OnPlayerAnimation          (cPlayer & a_Player, int a_Animation) = 0;
 	virtual bool OnPlayerBreakingBlock      (cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) = 0;
 	virtual bool OnPlayerBrokenBlock        (cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) = 0;
+	virtual bool OnPlayerClickingAtEntity   (cPlayer & a_Player, cEntity & a_Entity, const Vector3f & a_TargetPos) = 0;
```
madmaxoft: `Vector3` values should be passed by-value

src/Bindings/PluginManager.h
```diff
@@ -119,6 +119,7 @@ class cPluginManager
 		HOOK_PLAYER_PLACING_BLOCK,
 		HOOK_PLAYER_RIGHT_CLICK,
 		HOOK_PLAYER_RIGHT_CLICKING_ENTITY,
+		HOOK_PLAYER_CLICKING_AT_ENTITY,
```
madmaxoft: We've got "right-clicking" and "clicking", how about making this "left-clicking" so that the naming convention is the same?

src/ClientHandle.cpp
```diff
@@ -1755,6 +1755,28 @@ void cClientHandle::HandleUseEntity(UInt32 a_TargetEntityID, bool a_IsLeftClick)
...
 
+void cClientHandle::HandleUseEntityAt(UInt32 a_TargetEntityID, Vector3f a_TargetPos, eHand a_Hand)
+{
+	// If it is a right click, call the entity's OnRightClicked() handler:
+	cWorld * World = m_Player->GetWorld();
+	World->DoWithEntityByID(a_TargetEntityID, [=](cEntity & a_Entity)
+		{
+			if (
+				!cPluginManager::Get()->CallHookPlayerClickingAtEntity(*m_Player, a_Entity, a_TargetPos) &&
+				!m_Player->IsGameModeSpectator()  // Spectators cannot interact with every entity
```
madmaxoft: This means that the hook is executed for spectators, but then the `OnClickedAt` is not called. This behavior is not documented anywhere in the hook docs, is it even wanted?

src/Entities/Entity.h
```diff
@@ -562,6 +580,9 @@ class cEntity
 	/** Called when the specified player right-clicks this entity */
 	virtual void OnRightClicked(cPlayer & a_Player) {}
 
+	/** Called when the specified player interact at specific coords of this entity. Depend of entity and mostly used by ArmorStand */
+	virtual void OnClickedAt(cPlayer & a_Player, Vector3f a_TargetPos, eHand a_Hand) {}
```
madmaxoft: Is this right-click, left-click or any-click?

src/ClientHandle.cpp
Buom01: Not necessary wanted. For me it's totally normal that a plugin must have the priority on all things, even if spectator theoretically can't interact with entity.
Since the behavior is the same for other entity, it's implicitly the same.
Do I answer your question ?

This condition mean: If no plugins interfere, and the player isn't a spectator, then interact

src/Entities/Entity.h
Buom01: Undefined / Both
https://wiki.vg/index.php?title=Protocol&oldid=14204#Use_Entity
https://github.com/Buom01/cuberite/blob/armor-stand/src/Protocol/Protocol_1_9.cpp#L2853
As you see in the code, that's happen when the player use an entity with coords. Nothing more.
The doc says both, but in practice I don't remember.

src/Bindings/PluginManager.h
Buom01: See "Is this right-click, left-click or any-click?"

Buom01:
I think that some code should be write for news protocols relative to armor stand. I haven't read changes were made for now.

KrystilizeNevaDies:
Multiple Armor Stands in one place is a vanilla feature btw

Buom01:
Code is still online. Feel free to update the code yourself. Closing for now.

tigerw:
Thanks, will (eventually) do, re-opening so this doesn't get buried/forgotten about

bearbin:
Closing in favour of #5307, will reanimate that PR instead!

PR #4332, opened 2019-05-05
Ignore whitespace only lines in brewing and furnace recipes

peterbell10:
Currently, whitespace only lines spam the console with warnings. 
```
[11:41:01] brewing.txt: line 12: A line with '+' was expected
[11:41:01] Offending line: ""
```
As well as allowing manual whitespace, this handles the case where windows line endings are used with a linux build of cuberite. In which case you get lines with a single carriage return `"\r"`.


src/BrewingRecipes.cpp
```diff
@@ -45,8 +45,9 @@ void cBrewingRecipes::ReloadRecipes(void)
 			ParsingLine.erase(FirstCommentSymbol);
 		}
 
-		if (ParsingLine.empty())
+		if (std::all_of(ParsingLine.cbegin(), ParsingLine.cend(), isspace))
```
madmaxoft: It seems it could be useful to extract this test into a separate function, `IsOnlyWhitespace`, somewhere in StringUtils. How about that?

PR #4341, opened 2019-06-03
Fix directory traversal bug

wbrbr:
I found a neat little directory traversal bug in the webadmin code.
What the server does is that if the requested URL isn't "/", "/webadmin" or "/~webadmin" it will just try to fetch the corresponding file and send it back. The URL is sanitized before that: since we only want to serve files that are in the `webadmin/files/` directory, we search for `../` in the URL and remove it.

The problem here is that this sanitization doesn't work: if an attacker sends `....//file.foo` the sanitizer will transform it into `../file.foo` and they will have successfully bypassed the filter. For example if I go to `localhost:8080/....//....//webadmin.ini` I get the *webadmin.ini* file containing all the user passwords in plaintext.

I see three ways of fixing this:

* transform the relative path to an absolute path and check that it begins with the full path to *webadmin/files*
* apply the sanitization filter in a loop while there is something to sanitize
* just remove `..` instead of `../`. I think this fixes the bug but it might be less robust than the other two solutions (the first being probably the most robust but also the most complicated since we need to support Windows and Unix paths)

### DIFF
--- a/src/WebAdmin.cpp
+++ b/src/WebAdmin.cpp
@@ -375,13 +375,12 @@ void cWebAdmin::HandleFileRequest(cHTTPServerConnection & a_Connection, cHTTPInc
 		}
 	}
 
-	// Remove all ".." strings:
-	// ReplaceString(FileURL, "..", "");
-
 	// Read the file contents and guess its mime-type, based on the extension:
 	AString Content = "<h2>404 Not Found</h2>";
 	AString ContentType = "text/html";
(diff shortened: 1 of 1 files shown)


peterbell10:
Nice catch!

I'm not a fan of just replacing `..`, since it could potentially be part of an innocent file name. The other solutions all sound good. Another alternative would be to just fail as soon as any instances of `../` are found.

wbrbr:
Here you go

Issue #4340, opened 2019-05-31
Creeper targets dead player


RomainL972:
Client version:   1.11.1
Server OS:        Linux
Commit id:        24a8456f79aa35cfe8c3859c880e1bffeae088c6

### Expected behavior
When a player dies, the mobs don't target it

### Actual behavior
A creeper next to where a player died explodes

### Steps to reproduce the behavior
You need two players to see it
-Player 1 dies (doesn't respawn yet)
-Player 2 goes into creative mode and spawns a Creeper where Player 1 died
-The Creeper will explode for no reason
-The blocks flying around during an explosion is a nice feature by the way

madmaxoft:
I assume it's not only creepers, all hostile mobs probably have the same behavior?

RomainL972:
I didn't test with any other mob but I guess it works too.

Sent from ProtonMail mobile

-------- Original Message --------

Meynex:
Proof of concept fix:
file:  /src/Entities/Player.cpp
Line: 1296 - 1299
fix:
bool cPlayer::CanMobsTarget(void) const
{
	return IsGameModeSurvival() || IsGameModeAdventure() || m_health > 0;
}

PR #4344, opened 2019-06-15
Fix mobs targeting dead players

joeftiger:
In reference to #4340.


src/Entities/Player.cpp
```diff
@@ -1295,7 +1295,7 @@ bool cPlayer::IsGameModeSpectator(void) const
 
 bool cPlayer::CanMobsTarget(void) const
 {
-	return IsGameModeSurvival() || IsGameModeAdventure();
+	return (IsGameModeSurvival() || IsGameModeAdventure()) && m_Health > 0;
```
madmaxoft: Please use parentheses around the comparison for clarity.

madmaxoft:
```suggestion
	return (IsGameModeSurvival() || IsGameModeAdventure()) && (m_Health > 0);
```

PR #4345, opened 2019-06-16
Change LightingThread's world verification to an assert

ashleydavies:
Oops - there is a mistake in the rationale behind this PR. Sorry for the bother!

### DIFF
--- a/src/LightingThread.cpp
+++ b/src/LightingThread.cpp
@@ -336,7 +336,7 @@ void cLightingThread::ReadChunks(int a_ChunkX, int a_ChunkZ)
 		for (int x = 0; x < 3; x++)
 		{
 			Reader.m_ReadingChunkX = x;
-			VERIFY(m_World.GetChunkData(a_ChunkX + x - 1, a_ChunkZ + z - 1, Reader));
+			ASSERT(m_World.GetChunkData(a_ChunkX + x - 1, a_ChunkZ + z - 1, Reader));
 		}  // for z
 	}  // for x
 


madmaxoft:
I suppose you already found out why this is a `VERIFY` rather than `ASSERT` - the code inside the macro is actually needed for the execution, no matter what build it is in. Still, thanks for trying out Cuberite for a academic purposes, and hopefully you'll like it enough to come back once you finish your studies, for regular development :)

ashleydavies:
Yup, my bad; I didn't investigate the code enough & didn't notice the callback passed along - pretty obvious in hindsight!

It seems like a great codebase, though, so I'd love to contribute in future!

madmaxoft:
If you don't mind, I'd like to include your original text here, for reference:

Hi there! I have been working on extending a performance analysis tool as part of my degree's dissertation, and tested it out on Cuberite (incidentally, I found Cuberite while randomly browsing C++ projects, but I've long loved Minecraft and this seems like a great project!).

One of the things the tool does is monitor low-level synchronisation mechanisms, and identify cases of especially-high contention on mutexes. I noticed when running it that there is a significant amount of criticality around a specific stack trace, specifically an update within the Redstone Simulator:

[cuberite_sync](https://user-images.githubusercontent.com/934797/59559249-f5cd9e00-8ffa-11e9-8a9a-3f3775ddc61a.png)

This is the most "critical" stack trace it identified, and what stood out was 77% of the time this thread wanted to acquire the lock, it was held by the lighting thread.

The offending code in the lighting thread is within `ReadChunks` - specifically this line:

`VERIFY(m_World.GetChunkData(a_ChunkX + x - 1, a_ChunkZ + z - 1, Reader));`

This causes the lighting thread to acquire the global chunk lock for verification to take place, which causes contention and wrestling with other threads over it.

To verify this was an issue, tried to quantify what sort of difference it makes. Without much experience of your codebase, (and also being a little tight on time) I've only quantified the difference this makes on the lighting thread. To do so, I added a couple of lines of code within LightingThread.cpp within LightChunk (not pictured: a line at the start to measure begin):

[Screenshot 2019-06-16 at 05 56 38](https://user-images.githubusercontent.com/934797/59559282-83a98900-8ffb-11e9-8fb8-e992591f6077.png)

Essentially, every 100 calls, it averages the time it took for last 100 iterations and then prints it.

I then booted up my server with and without the VERIFY line commented out, and found that:

- On server initialisation, >200 lighting updates are twice as fast (going from an average of around 8ms to around 4ms each - an improvement of about 0.8 seconds on the lighting thread at boot time)
- While I stand still in a random chunk on an unmodified map, I notice:
  - Much less variance in lighting update speeds - always in the realm of 6ms, rather than wildly varying between 6 and 11ms
  - Much lower lighting update speeds

This is my brief log of during-game (i.e. ignoring the two that trigger at initialisation) updates:

After change:

```
[bccd487d845f2d96|05:49:20] The average lighting duration was 6182168ns
[bccd487d845f2d96|05:49:21] The average lighting duration was 6656868ns
[bccd487d845f2d96|05:49:22] The average lighting duration was 6645023ns
[bccd487d845f2d96|05:49:22] The average lighting duration was 6166565ns
[bccd487d845f2d96|05:49:23] The average lighting duration was 6933211ns
```

Before change (the timestamps are after because I undid the change to record these):

```
[1eb0f49ad7c8680|05:50:24] The average lighting duration was 5942752ns
[1eb0f49ad7c8680|05:50:25] The average lighting duration was 8124353ns
[1eb0f49ad7c8680|05:50:26] The average lighting duration was 10832519ns
[1eb0f49ad7c8680|05:50:27] The average lighting duration was 9336435ns
```

In addition to this, I haven't measured the difference this makes to any other threads - which may be quite substantial: the wild variations, even after server initialisation, would imply to me that there is a lot of contention going on for this lock (in addition to the traces from my tool).

I debated making this an issue, instead, in case you have a strong desire to keep this as a VERIFY rather than `ASSERT`, but thought a PR was more appropriate. I wonder if it may even be better to have an intermediate macro which is compiled out for builds where the user opts into more performance improvements at the cost of less logging information when things go wrong?

Hope this is helpful, and apologies in advance incase I've misunderstood anything!

PR #4346, opened 2019-06-27
Fix building with clang 8.0

Bond-009:


madmaxoft:
If there's a specific warning for the use of extra semicolons, I propose we instead turn that warning off - this diff clearly shows we don't want it.

Bond-009:
@peterbell10 That's just a nasty hack IMO.
@madmaxoft There are also plenty of places where we don't have semicolons after the macro.

madmaxoft:
The "nasty hack" is a pretty standard way of enforcing semicolons after macros. And I for one would very much appreciate if we did require the semicolons, so that the macros don't stand out in the code as something different from everything else.

Bond-009:
Alright, I changed it.

PR #4347, opened 2019-06-27
 Lock hopper when powered by redstone

### LINKED ISSUE #3451: Stopping a hopper with redstone doesn't work
Trying to use redstone on a hopper (Which would normally make it stop passing items) does not work.

Taken from #1109

Bond-009:
Fixes #3451


tigerw:
Changes seem reasonable.

but also it's just occurred to me how much boilerplate we need to write. so much indirection!

tigerw:
Travis is complaining about an undocumented API symbol, you could probably just remove the export for now.

mathiascode:
Any update on this?

Bond-009:
I’m still confused by the lua bindings

mathiascode:
Shouldn't it just be a case of documenting DoWithHopperAt in Server/Plugins/APIDump/Classes/World.lua, like the other DoWith...At methods? Or am I missing something?

PR #4353, opened 2019-07-17
Added missing return type and corrected one

Seadragon91:
Run APICheck and found this two.


madmaxoft:
I've updated this to master so that the CI runs completely (it seems to have stalled for some reason)

PR #4354, opened 2019-07-21
Check for nil in cWorld:SpawnSplitExperienceOrbs binding

### LINKED ISSUE #4352: API: Crash on call to cWorld:SpawnSplitExperienceOrbs
Server OS:        Windows
Commit id:        (9dc1343bda0592f2d27bb216eb21fdec97f73945)

### Expected behavior
Server should not crash

### Actual behavior
Server crashes

1) Add this code to a plugin inside a command handler
2) Call the command
`cRoot:Get():GetDefaultWorld():SpawnSplitExperienceOrbs(nil, 1)`

masp:
Adds explicit nil check for 'position' parameter to prevent crashes if invalid.

Fixes #4352

### DIFF
--- a/src/Bindings/ManualBindings_World.cpp
+++ b/src/Bindings/ManualBindings_World.cpp
@@ -898,7 +898,8 @@ static int tolua_cWorld_SpawnSplitExperienceOrbs(lua_State* tolua_S)
 		tolua_error(tolua_S, "Invalid 'self' in function 'SpawnSplitExperienceOrbs'", nullptr);
 		return 0;
 	}
-	if (Position == nullptr) {
+	if (Position == nullptr)
+	{
 		tolua_error(tolua_S, "Error in function 'SpawnSplitExperienceOrbs' arg #2. Value must not be nil.", nullptr);
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


src/Bindings/ManualBindings_World.cpp
```diff
@@ -890,14 +890,19 @@ static int tolua_cWorld_SpawnSplitExperienceOrbs(lua_State* tolua_S)
...
 	int Reward;
 	L.GetStackValues(1, self, Position, Reward);
 	if (self == nullptr)
 	{
 		tolua_error(tolua_S, "Invalid 'self' in function 'SpawnSplitExperienceOrbs'", nullptr);
 		return 0;
 	}
+	if (Position == nullptr)
+	{
+		tolua_error(tolua_S, "Error in function 'SpawnSplitExperienceOrbs' arg #2. Value must not be nil.", nullptr);
```
madmaxoft: The message might be misleading, `Position == nullptr` due to other causes as well - bad parameter type. A better message would be `Value must be a valid Vector3`, or even better `Expected Vector3, got XYZ instead`.

masp: Oh, I thought that the `CheckParamUserType(2, "Vector3<double>")` at the top of the function would check for that kind of type mismatch, and that only `nil` was slipping through because it's a valid `Vector3<double>`.

madmaxoft: You're right, I've missed that. Sorry.

PR #4358, opened 2019-08-07
change finalDamage from int to float

Aplaus228:


madmaxoft:
You don't need to close a pull request if you want to change more stuff. Simply push a new commit, it will be automatically added to the pull request.

Issue #4360, opened 2019-08-08
Fix "operator =" warning from tolua


madmaxoft:
When building, the tolua generator outputs the following warning:
```
CUSTOMBUILD : ** tolua warning : No support for operator =, ignoring.
```

Fix the cause of the warning (trying to export `operator =` to Lua).
Ideally, add an output to the tolua warning so that it's easier to identify the offending C++ code.

Seadragon91:
This can be closed.

madmaxoft:
Not sure. The one warning itself has been fixed, but if it ever comes back, the cryptic message is there again with no help whatsoever about where the warning comes from.

Seadragon91:
I took a look at this. The code of all files is put into a single variable, therefore without bigger changes it would not be possible to get the file and the line number that causes the warning.

It would be possible to print the code that and make it more obvious. Would that be enough?

madmaxoft:
The parser should be able to tell the class name, that would be absolutely helpful.

Seadragon91:
Are this [changes](https://github.com/Seadragon91/toluapp/commit/f913c29af2184b4abf57a6c5da68eb2099d0cf94#diff-a8897e1df87afae4b458d15080beb405) okay?

Example output:
```
** tolua warning: No support for operator =
Code that contains the unsupported operator:
cBoundingBox & operator=(const cBoundingBox & a_Other)

```

madmaxoft:
Almost, the output is exactly what I'm after, but it should still go through the `warning` function, rather than regular `print`. Please make a PR with such a change, you're just a tiny step away from it :)

Seadragon91:
Switched to using [warning](https://github.com/Seadragon91/toluapp/commit/e0ee6b64e9238fa3e3dea92a06779cdc503fc4b3). I used a local variable to build the output together. The function warning would print out two newlines for every line.

madmaxoft:
Still need to upgrade the tolua subrepo in here...

tigerw:
tolua is upgraded

PR #4361, opened 2019-08-08
cBoundingBox: Fix tolua warning

peterbell10:
Fixes the root cause of #4360. A better warning would be nice too though.


madmaxoft:
Weird that it's been there for 2 years :)

PR #4362, opened 2019-08-08
Removed coord-based API from cCuboid.

madmaxoft:
The API is removed from the C++ side, and manually obsoleted from the Lua side.

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -3992,8 +3992,7 @@ static int tolua_cCuboid_Assign(lua_State * tolua_S)
 		self->Assign(*pt1, *pt2);
 		return 0;
 	}
-	tolua_error(tolua_S, "invalid parameter to 'cCuboid:Assign', expected a (cCuboid) or a (Vector3i, Vector3i).", nullptr);
-	return 0;
+	return L.ApiParamError("Invalid parameter, expected either a cCuboid or two Vector3i-s.");
 }
 
 
@@ -4041,8 +4040,7 @@ static int tolua_cCuboid_IsInside(lua_State * tolua_S)
 			return 1;
 		}
 	}
-	tolua_error(tolua_S, "invalid parameter to 'cCuboid:IsInside', expected a (Vector3i) or (Vector3d).", nullptr);
-	return 0;
+	return L.ApiParamError("Invalid parameter #2, expected a Vector3i or a Vector3d.");
 }
 
 
@@ -4074,8 +4072,7 @@ static int tolua_cCuboid_Move(lua_State * tolua_S)
 	Vector3i * offset = nullptr;
 	if (!L.GetStackValue(2, offset) || (offset == nullptr))
 	{
-		tolua_error(tolua_S, "invalid parameter to 'cCuboid:Move', expected a Vector3i.", nullptr);
-		return 0;
+		return L.ApiParamError("Invalid parameter #2, expected a Vector3i.");
 	}
 	self->Move(*offset);
 	return 0;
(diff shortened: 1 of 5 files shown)


src/Cuboid.h
```diff
@@ -20,15 +20,14 @@ class cCuboid
...
-	// DEPRECATED, use cCuboid(Vector3i, Vector3i) instead
-	cCuboid(int a_X1, int a_Y1, int a_Z1, int a_X2, int a_Y2, int a_Z2) : p1(a_X1, a_Y1, a_Z1), p2(a_X2, a_Y2, a_Z2)
-	{
-		LOGWARNING("cCuboid(int, int, int, int, int, int) constructor is deprecated, use cCuboid(Vector3i, Vector3i) constructor instead.");
-	}
+	// tolua_end
+	// Exported in ManualBindings.cpp to support the old deprecated coord-based overload.
 
-	void Assign(int a_X1, int a_Y1, int a_Z1, int a_X2, int a_Y2, int a_Z2);
+	void Assign(const Vector3i & a_Point1, const Vector3i & a_Point2);
```
peterbell10:
```suggestion
	void Assign(Vector3i a_Point1, Vector3i a_Point2);
```

src/Cuboid.h
```diff
@@ -82,12 +75,19 @@ class cCuboid
...
 	Assumes both cuboids are sorted. */
 	bool IsCompletelyInside(const cCuboid & a_Outer) const;
 
-	/** Moves the cuboid by the specified offsets in each direction */
-	void Move(int a_OfsX, int a_OfsY, int a_OfsZ);
+	// tolua_end
+
+	/** Moves the cuboid by the specified offset.
+	Exported in ManualBindings to support the old deprecated coord-based overload. */
+	void Move(const Vector3i & a_Offset);
```
peterbell10:
```suggestion
	void Move(Vector3i a_Offset);
```

src/Bindings/ManualBindings.cpp
```diff
@@ -3954,6 +3954,137 @@ static int tolua_cCompositeChat_UnderlineUrls(lua_State * tolua_S)
...
+		LOGWARNING("cCuboid:Move(x, y, z) is deprecated, use cCuboid:Move(Vector3i) instead.");
+		L.LogStackTrace();
+		self->Move({x, y, z});
+		return 0;
+	}
+
+	Vector3i * offset = nullptr;
+	if (!L.GetStackValue(2, offset) || (offset == nullptr))
+	{
+		tolua_error(tolua_S, "invalid parameter to 'cCuboid:Move', expected a Vector3i.", nullptr);
```
peterbell10:
```suggestion
		return L.ApiParamError("invalid parameter, expected a Vector3i.");
```
Same for the other functions.

src/Cuboid.h
madmaxoft: Why is it better to make a copy?

peterbell10: See https://github.com/cuberite/cuberite/issues/3853#issuecomment-318932036

madmaxoft: Alright, I'll change that. I assume that works only for small `Vector3i`s, not for the larger `Vector3f`, `Vector3d` or `cCuboid` instances, right?

peterbell10: See [CONTRIBUTING.md] for the guidelines we agreed upon at the time.
> - Parameters smaller than 4 elements (e.g. Vector3, cChunkCoords) should be passed by value. All other parameters should be passed by const reference, where applicable.
>   - `Foo(Vector3d a_Param1, const cCuboid & a_Param2)`
>   - See the discussion in issue #3853


[CONTRIBUTING.md]: https://github.com/cuberite/cuberite/blob/master/CONTRIBUTING.md#L28-L30

madmaxoft: Thanks for the pointers. I got a bit rusty by not working on Cuberite for a long time :)

PR #4365, opened 2019-08-09
Improved CMake generator

madmaxoft:
Little improvements to CMake-generated project files:
  - Set MSVC startup project to Cuberite
  - Generate BuildInfo.h out of source

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,7 +71,6 @@ Makefile
 *.d
 *.so
 tests/*/*-exe
-BuildInfo.h
 CMakeCache.txt
 CMakeFiles
 Makefile
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -24,7 +24,7 @@
 #include "../BlockEntities/MobHeadEntity.h"
 #include "../BlockEntities/FlowerPotEntity.h"
 #include "../BoundingBox.h"
-#include "../BuildInfo.h"
+#include "BuildInfo.h"
 #include "../ClientHandle.h"
 #include "../CommandOutput.h"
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 
 
 
-cmake_minimum_required (VERSION 2.8.7)
+cmake_minimum_required (VERSION 3.6)
```
madmaxoft: This is needed for the `VS_STARTUP_PROJECT` property

madmaxoft: Wonder if it'll fail silently on older CMake versions (CircleCI has 3.5.1)

peterbell10: According to the docs, it should fail noisily:
>If the current version of CMake is lower than that required it will stop processing the project and report an error. 

Are you sure that Circle is using 3.5.1? We use a custom docker image on circle that might be more up to date than the default.

madmaxoft: I meant using an older cmake with the new folder property.

Yes, Circle is using 3.5.1, it did fail noisily, that's why I had to change it back.

peterbell10: We could update circle if that works out better.

madmaxoft: Doesn't really matter, it does seem to work this way, too. In a way, this provided me with a test I didn't have the guts to perform, now we know it works with older cmake, too.

Thinking about it, since we declare compatibility with cmake 2.8.7, we should probably test with that version (or update this line to the lowest version we test with).

peterbell10: There are definitely other files with higher requirements. They should probably be unified though.

PR #4366, opened 2019-08-10
Added null check for position

Seadragon91:
This check if the passed position is not null. The current error, when position is null is something about a invalid callback and this could be confusing.


src/Bindings/ManualBindings_World.cpp
```diff
@@ -487,6 +487,12 @@ static int tolua_cWorld_DoWithNearestPlayer(lua_State * tolua_S)
 	bool CheckLineOfSight = true, IgnoreSpectators = true;  // Defaults for the optional params
 	L.GetStackValues(1, Self, Position, RangeLimit, FnRef, CheckLineOfSight, IgnoreSpectators);
 
+	if (Position == nullptr)
+	{
+		tolua_error(tolua_S, "Error in function 'DoWithNearestPlayer' arg #2. Value must not be nil.", nullptr);
```
madmaxoft: Why not use `L.ApiParamError`?

Seadragon91: Ah sorry, I copied it. There are a few places that still uses tolua_error.

PR #4374, opened 2019-08-23
Added BlockState implementation for 1.13 support.

madmaxoft:
This adds a `BlockState` class that represents the individual block's state (like pre-1.13 block meta). It supports fast equality checks through a checksum.

The checksum algorithm can be later changed, based on the collisions we see once we get real data in; for now the checksum is a simple shift-and-xor

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,5 @@
 version: 1.0.{build}
+image: Visual Studio 2017
 configuration: Debug
 clone_depth: 50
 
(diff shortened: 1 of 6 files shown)


madmaxoft:
Uff, I had no idea things would be this hairy on such a simple concept :P

madmaxoft:
Do not merge yet (if it even succeeds), I'll update this to be 2 separate commits, the Appveyour upgrade and the BlockState code.

madmaxoft:
Ready for review

PR #4376, opened 2019-08-26
Improved testing framework.

madmaxoft:
Removed `testassert` and `CheckAsserts` in favor of `TEST_EQUAL`, `TEST_ASSERTS` et al in tests/TestHelpers.h.

Added more testing options.

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -296,7 +296,7 @@ template class SizeChecker<UInt8,  1>;
 /** Faster than (int)floorf((float)x / (float)div) */
 #define FAST_FLOOR_DIV(x, div) (((x) - (((x) < 0) ? ((div) - 1) : 0)) / (div))
 
-// Own version of assert() that writes failed assertions to the log for review
+// Own version of ASSERT() that plays nicely with the testing framework
 #ifdef TEST_GLOBALS
 
 	class cAssertFailure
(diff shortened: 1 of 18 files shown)


madmaxoft:
Not sure if we can make any library work with our custom `ASSERT`s. And it's already written :)

peterbell10:
Would it not just be a case of
```cpp
#include <catch.hpp>
#define ASSERT REQUIRE
```

madmaxoft:
Not really, because we want to test that the code we call raises an assert (see the `TEST_ASSERTS` macro). But that code resides in the main source tree in src/ and is not part of the test, so it must not include the test library.

peterbell10:
I think the biggest problem with this test framework is that when multiple test-cases exist it isn't always clear which one has failed. This was a problem before as well though, so I'm fine with merging.

madmaxoft:
Both the assertion and test exceptions now store the exact file name, line number and function name of the failure, so it should be pretty apparent where the problem is.

peterbell10:
Unless I'm misunderstanding, when an assertion fails you can tell which assertion failed but not which test case was running when the assertion trigered.

madmaxoft:
Hm, that's true. Perhaps next refactor :)

PR #4377, opened 2019-08-26
Added a basic PalettedBlockArea implementation

madmaxoft:
This is the minimal PalettedBlockArea implementation, on which the 1.13+ support can be built.

This PR includes #4376 since it's dependent on the new testing features; I will rebase this when the dependency is merged.

### DIFF
--- a/src/BlockTypePalette.cpp
+++ b/src/BlockTypePalette.cpp
@@ -1,0 +1,80 @@
+#include "Globals.h"
+#include "BlockTypePalette.h"
+
+
+
+
+BlockTypePalette::BlockTypePalette()
+{
+	// Nothing needed yet
+}
+
+
+
+
+
+UInt32 BlockTypePalette::index(const AString & aBlockTypeName, const BlockState & aBlockState)
+{
+	auto idx = maybeIndex(aBlockTypeName, aBlockState);
+	if (idx.second)
+	{
+		return idx.first;
+	}
+
+	// Not found, append:
+	mPalette.push_back(std::make_pair(aBlockTypeName, aBlockState));
+	return static_cast<UInt32>(mPalette.size() - 1);
+}
+
+
+
+
+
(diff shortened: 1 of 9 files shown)


madmaxoft:
Wow, the PBA test code is so large that Github doesn't show it by default :) Finally we can say we have some proper tests.

tigerw:
xoft your function names are in violation of your own style guide XD

madmaxoft:
Which ones?
Note: in https://forum.cuberite.org/thread-3215.html I've declared I'll change the naming convention to Java-like. When the core 1.13 is finished, I'll try writing a Lua script to change all the rest of the sources.

tigerw:
oh no even xoft's been taken over by javaStyle

:(

madmaxoft:
In the first place xoft hates underscores :P

src/BlockTypePalette.cpp
```diff
@@ -0,0 +1,80 @@
...
+BlockTypePalette::BlockTypePalette()
+{
+	// Nothing needed yet
+}
+
+
+
+
+
+UInt32 BlockTypePalette::index(const AString & aBlockTypeName, const BlockState & aBlockState)
```
NiLSPACE: I'm not really sure if `index` is a right name if it can also append the blockstate when it's not there yet. I'm not really fond of the alternatives I came up with though:

- indexOrAdd
- findIndexOrAdd
- indexOfOrAdd
- safeIndexOf

madmaxoft: I think this is the most common operation, you want the index of a blockspec and you don't care if it's there or not. Basically it's the same as `std::map::operator[]`.

I think it's more understandable to have the pair `index` and `maybeIndex`, than `indexOrAdd` and `indexOrFail`. Still, I'm open to suggestions, too.

NiLSPACE: Lets keep it like this for now.

PR #4382, opened 2019-09-02
Separate the Generator from World and Plugin interfaces

madmaxoft:
The generator now doesn't depend on any `cChunkSink` or `cPluginInterface`, its sole responsibility is to generate a single chunk on request.
Moved the generator defaults into the `cComposableGenerator`, rather than `cWorld`.
Added a basic test that checks the default Overworld and Nether generator.

### DIFF
--- a/src/ChunkGeneratorThread.cpp
+++ b/src/ChunkGeneratorThread.cpp
@@ -1,0 +1,264 @@
+#include "Globals.h"
+#include "ChunkGeneratorThread.h"
+#include "Generating/ChunkGenerator.h"
+#include "Generating/ChunkDesc.h"
+
+
+
+
+
+/** If the generation queue size exceeds this number, a warning will be output */
+const size_t QUEUE_WARNING_LIMIT = 1000;
+
+/** If the generation queue size exceeds this number, chunks with no clients will be skipped */
+const size_t QUEUE_SKIP_LIMIT = 500;
+
+
+
+
+
+cChunkGeneratorThread::cChunkGeneratorThread(void) :
+	Super("cChunkGeneratorThread"),
+	m_Generator(nullptr),
+	m_PluginInterface(nullptr),
+	m_ChunkSink(nullptr)
+{
+}
+
+
+
+
+
+cChunkGeneratorThread::~cChunkGeneratorThread()
+{
+	Stop();
+}
+
+
+
+
+
+bool cChunkGeneratorThread::Initialize(cPluginInterface & a_PluginInterface, cChunkSink & a_ChunkSink, cIniFile & a_IniFile)
+{
+	m_PluginInterface = &a_PluginInterface;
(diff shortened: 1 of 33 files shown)


src/ChunkGeneratorThread.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+
+
+
+
+void cChunkGeneratorThread::Execute(void)
+{
+	// To be able to display performance information, the generator counts the chunks generated.
+	// When the queue gets empty, the count is reset, so that waiting for the queue is not counted into the total time.
+	int NumChunksGenerated = 0;  // Number of chunks generated since the queue was last empty
```
NiLSPACE: Shouldn't variables start lower-cased?

src/ChunkGeneratorThread.h
```diff
@@ -0,0 +1,162 @@
...
+
+		QueueItem(cChunkCoords a_Coords, bool a_ForceRegeneration, cChunkCoordCallback * a_Callback):
+			m_Coords(a_Coords),
+			m_ForceRegeneration(a_ForceRegeneration),
+			m_Callback(a_Callback)
+		{
+		}
+	};
+
+	using Queue = std::list<QueueItem>;
```
NiLSPACE: Wouldn't it make more sense to use a `std::queue` instead?

peterbell10: Or the queue from `OSSupport/Queue.h` which should be thread-safe already.

src/ChunkGeneratorThread.cpp
madmaxoft: This is basically only code that has been moved, so I didn't change it.
Subconsciously I don't use the new conventions on old code just yet, I'll change everything at once when changing the API.

src/ChunkGeneratorThread.h
madmaxoft: Again, code that has been just moved around.
`std::queue` is just a wrapper over `std::deque` by default.

OSSupport/Queue.h might be interesting, didn't remember we had one. I'll have a look at it.

madmaxoft: I has a look at the `cQueue`, but in the end, `cChunkGeneratorThread` does a bit more than just read and write to a queue, it calculates the throughput, and I'm too lazy to adapt that to use the `cQueue`, it isn't a straightforward change.

madmaxoft:
Well if that isn't weird - Debug build tests succeed, Release build tests fail because the generated chunk doesn't seem to have any Nether-specific blocks. WAT?

madmaxoft:
I'll hotfix the test for now (it doesn't consider nether quarts ore nor lava as a nether block), but we need to find out why there's differring data generated on debug / release builds, and whether it happens only on Linux, or on Windows, too.

PR #4383, opened 2019-09-04
Updated LibEvent to release 2.1.11-stable.

madmaxoft:
This finally restores my ability to compile on Windows and Linux from the same source folder (on a network drive).

LibEvent broke this long ago by writing a config file into the source folder, rather than build folder. Now it's finally fixed.


madmaxoft:
Note: After merging this, it is best to completely remove your local build folder and re-run cmake from a fresh start. LibEvent has changed too much of their cmake code and the upgrade is not clean, it fails to build rather often.

PR #4385, opened 2019-09-07
Fix generator repeatability

### LINKED ISSUE #4384: Chunk generator creates different data on different platforms / build-types
While implementing #4382 it by chance became obvious that our default Nether generator generates different data in Debug and Release builds on Linux. We need to find out why and fix it, the assumption has always been that the generator is platform- and build-type-independent.

As the first step, a new test needs to be made to check the generators, how many of them are affected. Then, fix all those that fail the test.

madmaxoft:
This hot-fixes the Nether generator producing differrent data on Linux Release builds, and adds a check for the default Overworld and Nether generators whether they produce the same output on all platforms.

Fixes #4384.

### DIFF
--- a/tests/Generating/BasicGeneratorTest.cpp
+++ b/tests/Generating/BasicGeneratorTest.cpp
@@ -3,6 +3,7 @@
 #include "Generating/ChunkDesc.h"
 #include "../TestHelpers.h"
 #include "IniFile.h"
+#include "mbedTLS++/Sha1Checksum.h"
 
 
 
@@ -27,6 +28,21 @@ static void verifyChunkDescHeightmap(const cChunkDesc & a_ChunkDesc)
 			}  // for y
 		}  // for z
 	}  // for x
+}
+
+
+
+
+
+static AString chunkSHA1(const cChunkDesc & a_ChunkDesc)
+{
+	cSha1Checksum cs;
+	cs.Update(a_ChunkDesc.GetBlockTypes(), cChunkDef::Width * cChunkDef::Width * cChunkDef::Height);
+	cSha1Checksum::Checksum digest;
+	cs.Finalize(digest);
+	AString res;
+	cSha1Checksum::DigestToJava(digest, res);
+	return res;
 }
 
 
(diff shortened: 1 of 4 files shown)


madmaxoft:
Huh, I first added the tests and then the fix, but rebased to put the two commits in reverse order. Now Github shows them in the original order, while they are in the right order in the code. That's becoming weird.

tests/Generating/BasicGeneratorTest.cpp
```diff
@@ -217,7 +218,94 @@ static void testGenerateNether()
...
+
+
+/** Checks that the specified generator generates chunk that match the specified checksums. */
+static void checkChunkChecksums(
+	cChunkGenerator & aGenerator,
+	const std::vector<CoordsWithChecksum> & aCoordsWithChecksum,
+	const AString & aDimension
+)
+{
+	LOG("Testing the repeatability of the %s generator", aDimension.c_str());
```
peterbell10:
```suggestion
	LOG("Testing the repeatability of the %s generator", aDimension);
```

tests/Generating/BasicGeneratorTest.cpp
```diff
@@ -217,7 +218,94 @@ static void testGenerateNether()
...
+	for (const auto & coords: aCoordsWithChecksum)
+	{
+		cChunkDesc chd(coords.mCoords);
+		aGenerator.Generate(coords.mCoords.m_ChunkX, coords.mCoords.m_ChunkZ, chd);
+		auto checksum = chunkSHA1(chd);
+		TEST_EQUAL_MSG(checksum, coords.mChecksum,
+			Printf("%s chunk %s SHA1: expected %s, got %s", aDimension.c_str(), coords.mCoords.ToString().c_str(), coords.mChecksum.c_str(), checksum.c_str())
```
peterbell10:
```suggestion
			Printf("%s chunk %s SHA1: expected %s, got %s", aDimension, coords.mCoords.ToString(), coords.mChecksum, checksum)
```

tests/Generating/BasicGeneratorTest.cpp
madmaxoft: I don't think a `std::string` is allowed to be passed through `...` varargs, I seem to remember this being an undefined behavior.

peterbell10: It doesn't use c-style varargs any more
https://github.com/cuberite/cuberite/blob/d8d1e6e1d2f668cb7b34eae7484e59e2310e847a/src/LoggerSimple.h#L21

madmaxoft: Okay, template magic... :P

PR #4387, opened 2019-09-08
Generators: Use cChunkCoords

madmaxoft:
The generators now use `cChunkCoords` instead of two `int`s.

### DIFF
--- a/src/Generating/BioGen.h
+++ b/src/Generating/BioGen.h
@@ -33,7 +33,7 @@ protected:
 	EMCSBiome m_Biome;
 
 	// cBiomeGen overrides:
-	virtual void GenBiomes(int a_ChunkX, int a_ChunkZ, cChunkDef::BiomeMap & a_BiomeMap) override;
+	virtual void GenBiomes(cChunkCoords a_ChunkCoords, cChunkDef::BiomeMap & a_BiomeMap) override;
 	virtual void InitializeBiomeGen(cIniFile & a_IniFile) override;
 } ;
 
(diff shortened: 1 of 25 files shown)


src/Generating/BioGen.h
```diff
@@ -48,31 +48,37 @@ class cBioGenCache :
 	typedef cBiomeGen super;
 
 public:
+
 	cBioGenCache(cBiomeGenPtr a_BioGenToCache, size_t a_CacheSize);
-	virtual ~cBioGenCache() override;
+
```
peterbell10: Perhaps keep the destructor as `override` and `= default` to ensure the base class has a `virtual` destructor?
```suggestion
	virtual ~cBioGenCache() override = default;
```

src/Generating/HeiGen.h
```diff
@@ -29,10 +29,9 @@ class cHeiGenCache :
 {
 public:
 	cHeiGenCache(cTerrainHeightGenPtr a_HeiGenToCache, size_t a_CacheSize);
-	virtual ~cHeiGenCache() override;
```
peterbell10: Same here
```cpp
virtual ~cHeiGenCache() override = default;
```

src/Generating/BioGen.h
madmaxoft: Does gcc 4.8 support that? (It's still used for RasPi cross-builds on the build server)

peterbell10: According to [cppreference](https://en.cppreference.com/w/cpp/compiler_support) it's supported defaulted and deleted functions since gcc 4.4

madmaxoft: Changed.

PR #4388, opened 2019-09-09
Protocol: Use logical outgoing packet types

madmaxoft:
This extends the "logical outgoing packet type" to be used everywhere for consistency, and for better comm-logging (the log now shows textual packet types for outgoing packets).


madmaxoft:
This is a cursed PR, I already restarted the CI job twice and it has stalled again :(

src/Protocol/PacketID.cpp
```diff
@@ -1,184 +0,0 @@
-#include "Globals.h"
-
-#include "Protocol_1_9.h"
-#include "Protocol_1_12.h"
-
-
-
-
-
-UInt32 cProtocol_1_9_0::GetPacketId(eOutgoingPackets a_Packet)
```
peterbell10: Why was this file removed? It was specifically made a seperate file so that `Super::GetPacketId` could be inlined.

madmaxoft: I don't think it would be inlined, even when separated into a file - the function is virtual, after all.
Also, this intent was not documented in the file, and third, micro-optimization without actual benchmarking, against maintainability.

peterbell10: >I don't think it would be inlined, even when separated into a file - the function is virtual, after all.

`Super::GetPacketId` doesn't use virtual dispatch, otherwise it would just call itself recursively.

> [...] against maintainability.

Personally I think having one file to go to find packet IDs is useful in itself.

madmaxoft: Not really, usually you only implement a new protocol and wiki.vg has a list of packets that changed their numbers, so that's what you want to edit. Having one function defined in another file than the rest feels counter-intuitive.

PR #4389, opened 2019-09-10
Fix race condition in UrlClientTest

peterbell10:
Tentative fix for #4124

Consider there are two threads, one is in `cEvent::Set`
https://github.com/cuberite/cuberite/blob/3bc0f07fea72f4334f4e68227a065fba10e83d4d/src/OSSupport/Event.cpp#L53-L60

and the other is in `cEvent::Wait`
https://github.com/cuberite/cuberite/blob/3bc0f07fea72f4334f4e68227a065fba10e83d4d/src/OSSupport/Event.cpp#L28

If there is a spurious wakeup in the waiting thread after the setting thread has already set `m_ShouldContinue` then it is possible for the waiting thread to wake before `notify_one()` is called. 

In the `cUrlClient` tests, the `cEvent` object is owned by the waiting thread and is destroyed immediately after it wakes. This can lead to the setting thread calling `notify_one()` on a condition variable _after_ it's already destroyed.

I recently run into this issue on another project and the problem manifests mostly on macOS, resulting in a mixture of deadlocks and occasional segfaults. Same as #4124.

### DIFF
--- a/tests/HTTP/UrlClientTest.cpp
+++ b/tests/HTTP/UrlClientTest.cpp
@@ -18,8 +18,8 @@ class cCallbacks:
 	public cUrlClient::cCallbacks
 {
 public:
-	cCallbacks(cEvent & a_Event):
-		m_Event(a_Event)
+	cCallbacks(std::shared_ptr<cEvent> a_Event):
+		m_Event(std::move(a_Event))
 	{
 		++g_ActiveCallbacks;
 		LOGD("Created a cCallbacks instance at %p", reinterpret_cast<void *>(this));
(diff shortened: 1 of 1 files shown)


madmaxoft:
Is this limited to the test, or is there a similar construct in Cuberite itself?

peterbell10:
Here the problem is the test itself, I can take a look over the other uses of `cEvent` for the same issue.

madmaxoft:
If you have the time and the nerves for it, sure, please do. We can still merge this PR already, it does fix a problem that hurts us :)

PR #4390, opened 2019-09-10
Initial 1.13 protocol support

madmaxoft:
The client is able to hold the connection, the terrain renders (but in wrong blocks), the items in inventory show invalid items.

This is before the protocol block or item type maps, and before the flattening.

Based on @tigerw 's code in #4274, cleaned up and modernized a bit.


src/Protocol/ChunkDataSerializer.cpp
```diff
@@ -440,3 +441,142 @@ void cChunkDataSerializer::Serialize110(AString & a_Data, int a_ChunkX, int a_Ch
...
+				Packet.WriteBuf(a_Section.m_BlockSkyLight, sizeof(a_Section.m_BlockSkyLight));
+			}
+		}
+	);
+
+	// Write the biome data
+	UInt32 Munged[BiomeDataSize];
+	for (size_t i = 0; i != BiomeDataSize; i++)
+	{
+		Munged[i] = static_cast<UInt32>(m_BiomeData[i]) << 24;
```
madmaxoft: This is wrong for BE architectures

madmaxoft:
`cWorld` change reverted, the protocol works well without it.

PR #4392, opened 2019-09-11
Add pragma to ignore fallthrough warnings within Unicode Inc. code

E14:
While building locally (using GCC) I got a ton of warnings from fallthrough switches within the Unicode code. 
It's better with clang, which only reports once, but GCC throws warnings on every include it seems.

This change should fix this for GCC.
The PR is mostly to verify this does not interfere on the other build systems, merge it if you like.

### DIFF
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -414,8 +414,10 @@ AString UnicodeCharToUtf8(unsigned a_UnicodeChar)
 
 
 
+#ifdef __GNUC__
 #pragma GCC diagnostic push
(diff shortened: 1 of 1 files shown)


madmaxoft:
Now MSVC complains about an unknown pragma:
https://ci.appveyor.com/project/madmaxoft/mcserver/builds/27345093#L728

You should add an `#ifdef` for gcc / clang.

E14:
> Now MSVC complains about an unknown pragma:
Why did I expect that...

Done, but if that doesn't do it, I'll probably just switch to clang or add a hook that applies the patch and forget about this, this compiler-fighting is a bit weird

PR #4400, opened 2019-09-24
Fixed MSVC warnings

madmaxoft:
MSVC Win64 build complained a lot about converting from `int` to `float`, and from `size_t` to `int`.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -288,7 +288,7 @@ void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_R
 	}
 	ApplyArmorDamage(ArmorCover);
 
-	cEntity::TakeDamage(a_DamageType, a_Attacker, a_RawDamage, FinalDamage, a_KnockbackAmount);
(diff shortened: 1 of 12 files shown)


tonibm19:
hi admin

you hope you doing well

thanks for contacting me

i really was a wonderfull day on your office

i hope we can talk together again

vanessa

madmaxoft:
Travis has finished building, but seems to have lost the notification, so it still looks as if pending in the status here.

src/Entities/Entity.cpp
```diff
@@ -518,7 +518,7 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
 
 			if (Random.RandBool(Chance / 100.0))
 			{
-				a_TDI.Attacker->TakeDamage(dtAttack, this, 0, Random.RandInt(1, 4), 0);
+				a_TDI.Attacker->TakeDamage(dtAttack, this, 0, Random.RandReal(1.0f, 4.0f), 0);
```
peterbell10: Isn't raw damage usually done in increments of 1 hp? I think just casting the result of `RandInt` makes more sense.

madmaxoft: I can't find any reference to this, and since the UI always rounds to the nearest half-heart, I don't think this matters much.

PR #4403, opened 2019-09-28
Refactored more of Entities and BlockEntities to use Vector3.

madmaxoft:
The BlockEntities use `Vector3i` for position storage.
Both BlockEntities and Entities use `Vector3` in constructor.

Notably, added new utility functions to `Vector3`:
`added(ofs)` returns a copy of the vector moved by the specified offset (3 axes)
`addedX(ofs)` returns a copy of the vector moved by ofs in the X axis
`addedXZ(ofsX, ofsZ)` returns a copy of the vector moved by the specified offsets in X and Z axes.
etc.
These are used when iterating over the neighbors of existing `Vector3` instances.

Despite the looks of it, this *is* a step towards 1.13 :)

PR #4405, opened 2019-09-29
Cactus can now grow

Seadragon91:
No wonder I would never see a cactus grow on my skyblock island :)

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1145,7 +1145,7 @@ int cChunk::GrowCactus(int a_RelX, int a_RelY, int a_RelZ, int a_NumBlocks)
 			for (auto & Coord : Coords)
 			{
 				if (
-					UnboundedRelGetBlockType(a_RelX + Coord.x, Top + 1, a_RelZ + Coord.z, BlockType) &&
+					UnboundedRelGetBlockType(a_RelX + Coord.x, Top + i, a_RelZ + Coord.z, BlockType) &&
 					(
 						cBlockInfo::IsSolid(BlockType) ||
 						(BlockType == E_BLOCK_LAVA) ||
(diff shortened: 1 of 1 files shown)


Seadragon91:
Ah yes you are right. I also added a change for dropping the cactus, without it you can not create a cactus farm.

madmaxoft:
I hope you tested the changes, I just read the code :)

src/Chunk.cpp
```diff
@@ -1145,20 +1145,33 @@ int cChunk::GrowCactus(int a_RelX, int a_RelY, int a_RelZ, int a_NumBlocks)
...
+					UnboundedRelGetBlockType(a_RelX + Coord.x, Top + i, a_RelZ + Coord.z, BlockType) &&
 					(
 						cBlockInfo::IsSolid(BlockType) ||
 						(BlockType == E_BLOCK_LAVA) ||
 						(BlockType == E_BLOCK_STATIONARY_LAVA)
 					)
 				)
 				{
+					// Remove cactus
+					GetWorld()->DigBlock(a_RelX + GetPosX() * cChunkDef::Width, Top + i, a_RelZ + GetPosZ() * cChunkDef::Width);
```
madmaxoft: Doesn't `DigBlock` already create the pickups?

src/Chunk.cpp
```diff
@@ -1145,20 +1145,33 @@ int cChunk::GrowCactus(int a_RelX, int a_RelY, int a_RelZ, int a_NumBlocks)
...
+					// Drop cactus
+					cBlockHandler* Handler = BlockHandler(E_BLOCK_CACTUS);
+					cChunkInterface ChunkInterface(GetWorld()->GetChunkMap());
+					cBlockInServerPluginInterface PluginInterface(*GetWorld());
+					Handler->DropBlock(
+						ChunkInterface,
+						*GetWorld(),
+						PluginInterface,
+						nullptr,
+						a_RelX + GetPosX() * cChunkDef::Width  + (Coord.x * -1),
```
madmaxoft: ??? `* -1` ???
Why not just `- Coord.x` ?

Seadragon91: I want to drop the block on the other side. On the left side is the stone on the right side is air, therefore *-1.

madmaxoft: That's worth a comment.

Anyway, I just wanted to point out that adding a number multiplied by -1 is the same as subtracting the number.

src/Chunk.cpp
Seadragon91: It doesn't. From the [API](https://api.cuberite.org/cWorld.html#DigBlock_1): `Replaces the specified block with air, without dropping the usual pickups for the block.`

Seadragon91:
The cactus farm I tested works. Now looking why the hopper is not working correctly.

madmaxoft:
If you want to have a look at the hoppers, better do so after #4403 is merged, it changes some stuff in hoppers.

PR #4413, opened 2019-10-04
Improved large jungle and acacia tree generation 

NiLSPACE:

### DIFF
--- a/src/Generating/Trees.cpp
+++ b/src/Generating/Trees.cpp
@@ -30,30 +30,34 @@ static const sCoords Corners[] =
 } ;
 
 // Array with possible directions for a branch to go to.
-static const Vector3d AvailableBranchDirections[] =
+static const std::array<Vector3d, 32> & availableBranchDirections()
 {
-	{ -1, 0, 0 }, { 0, 0, -1  },
-	{ -1, 0, 1 }, { -1, 0, -1 },
-	{ 1, 0, 1  }, { 1, 0, -1  },
-	{ 1, 0, 0  }, { 0, 0, 1   },
+	static const std::array<Vector3d, 32> res =
+	{
+		{
+			{ -1, 0, 0 }, { 0, 0, -1  },
+			{ -1, 0, 1 }, { -1, 0, -1 },
+			{ 1, 0, 1  }, { 1, 0, -1  },
+			{ 1, 0, 0  }, { 0, 0, 1   },
 
-	{ -0.5, 0, 0   }, { 0, 0, -0.5    },
-	{ -0.5, 0, 0.5 }, { -0.5, 0, -0.5 },
-	{ 0.5, 0, 0.5  }, { 0.5, 0, -0.5  },
-	{ 0.5, 0, 0    }, { 0, 0, 0.5     },
+			{ -0.5, 0, 0   }, { 0, 0, -0.5    },
+			{ -0.5, 0, 0.5 }, { -0.5, 0, -0.5 },
+			{ 0.5, 0, 0.5  }, { 0.5, 0, -0.5  },
+			{ 0.5, 0, 0    }, { 0, 0, 0.5     },
 
-	{ -1, 0.5, 0 }, { 0, 0.5, -1  },
-	{ -1, 0.5, 1 }, { -1, 0.5, -1 },
-	{ 1, 0.5, 1  }, { 1, 0.5, -1  },
-	{ 1, 0.5, 0  }, { 0, 0.5, 1   },
+			{ -1, 0.5, 0 }, { 0, 0.5, -1  },
+			{ -1, 0.5, 1 }, { -1, 0.5, -1 },
+			{ 1, 0.5, 1  }, { 1, 0.5, -1  },
+			{ 1, 0.5, 0  }, { 0, 0.5, 1   },
 
-	{ -0.5, 0.5, 0   },  { 0, 0.5, -0.5    },
-	{ -0.5, 0.5, 0.5 },  { -0.5, 0.5, -0.5 },
-	{ 0.5, 0.5, 0.5  },  { 0.5, 0.5, -0.5  },
-	{ 0.5, 0.5, 0    },  { 0, 0.5, 0.5     },
-
-};
-
+			{ -0.5, 0.5, 0   },  { 0, 0.5, -0.5    },
+			{ -0.5, 0.5, 0.5 },  { -0.5, 0.5, -0.5 },
+			{ 0.5, 0.5, 0.5  },  { 0.5, 0.5, -0.5  },
+			{ 0.5, 0.5, 0    },  { 0, 0.5, 0.5     },
+		}
+	};
+	return res;
+}
 
 // BigO = a big ring of blocks, used for generating horz slices of treetops, the number indicates the radius
 
@@ -446,8 +450,8 @@ void GetLargeAppleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a
 	for (int i = 4; i < Height; i++)
 	{
 		// Get a direction for the trunk to go to.
-		Vector3d BranchStartDirection = AvailableBranchDirections[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY + i, a_BlockZ)) % ARRAYCOUNT(AvailableBranchDirections)];
-		Vector3d BranchDirection = AvailableBranchDirections[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY / i, a_BlockZ)) % ARRAYCOUNT(AvailableBranchDirections)] / 3;
+		Vector3d BranchStartDirection = availableBranchDirections()[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY + i, a_BlockZ)) % availableBranchDirections().size()];
+		Vector3d BranchDirection = availableBranchDirections()[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY / i, a_BlockZ)) % availableBranchDirections().size()] / 3;
 
 		int BranchLength = 2 + a_Noise.IntNoise3DInt(a_BlockX * a_Seq, a_BlockY * a_Seq, a_BlockZ * a_Seq) % 3;
 		GetTreeBranch(E_BLOCK_LOG, E_META_LOG_APPLE, a_BlockX, a_BlockY + i, a_BlockZ, BranchLength, BranchStartDirection, BranchDirection, a_BlockY + Height, a_LogBlocks);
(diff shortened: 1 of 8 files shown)


madmaxoft:
To fix the CI build, you'll need to wrap the `static const` globals into a function that returns them:
```cpp
static const std::array<Vector3d, ...> & availableBranchDirections()
{
  static const std::array<Vector3d, ...> res =
  {
    { -1, 0, 0 }, { 0, 0, -1  },
    ...
  };
  return res;
}
```
That should work around that warning / error.

src/Generating/Trees.cpp
```diff
@@ -367,40 +446,15 @@ void GetLargeAppleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a
...
-	};
-
 	// Create branches
 	for (int i = 4; i < Height; i++)
 	{
 		// Get a direction for the trunk to go to.
-		Vector3d BranchStartDirection = AvailableDirections[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY + i, a_BlockZ)) % ARRAYCOUNT(AvailableDirections)];
-		Vector3d BranchDirection = AvailableDirections[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY / i, a_BlockZ)) % ARRAYCOUNT(AvailableDirections)] / 3;
+		Vector3d BranchStartDirection = availableBranchDirections()[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY + i, a_BlockZ)) % availableBranchDirections().size()];
+		Vector3d BranchDirection = availableBranchDirections()[static_cast<size_t>(a_Noise.IntNoise3DInt(a_BlockX, a_BlockY / i, a_BlockZ)) % availableBranchDirections().size()] / 3;
```
madmaxoft: This brought me to an idea, why not just use a function that returns a single vector, based on the coords and the sequence?
`Vector3d pickBranchDirection(Vector3i a_BlockPos, int a_Seq)`
I think it'd make the code much more readable.

NiLSPACE: Yeah, that does look better.

src/Generating/Trees.cpp
```diff
@@ -29,6 +29,37 @@ static const sCoords Corners[] =
...
+			{ 1, 0.5, 0  }, { 0, 0.5, 1   },
+
+			{ -0.5, 0.5, 0   },  { 0, 0.5, -0.5    },
+			{ -0.5, 0.5, 0.5 },  { -0.5, 0.5, -0.5 },
+			{ 0.5, 0.5, 0.5  },  { 0.5, 0.5, -0.5  },
+			{ 0.5, 0.5, 0    },  { 0, 0.5, 0.5     },
+		}
+	};
+
+	return directions[(static_cast<size_t>(a_BlockPos.SqrLength()) + a_Seq) % directions.size()];
```
madmaxoft: I don't think `SqrLength` is a good replacement, you should still use `a_Noise.IntNoise3DInt()`

madmaxoft:
Tip: You can set up a Linux VM on your dev computer, network-share the Cuberite sources from your development folder to it, and compile-check it even before committing the code. At least that's what I do, except I have a separate Linux machine :)

That way you're less likely to add "fix linux compilation" commits later on, not to mention it forces you to think about platform independence more.

NiLSPACE:
I just found out that that the Linux subsystem for Windows also works.

Seadragon91:
I also use wsl, but with debian, you can get debian for wsl from the microsoft store :)

madmaxoft:
Could you rebase against current master, so that the hopper fix is not included in this PR?

src/Generating/Trees.cpp
```diff
@@ -29,6 +29,38 @@ static const sCoords Corners[] =
 	{1, 1},
 } ;
 
+// Array with possible directions for a branch to go to.
```
madmaxoft: No longer an array; change to doxy-comment and document the individual parameters and return value.

src/Generating/Trees.cpp
```diff
@@ -29,6 +29,38 @@ static const sCoords Corners[] =
...
+static const Vector3d & pickBranchDirection(cNoise a_Noise, Vector3i a_BlockPos, int a_Seq)
```
madmaxoft: Not a big deal, but I think the signature should be:
```cpp
static const Vector3d pickBranchDirection(const cNoise & a_Noise, Vector3i a_BlockPos, int a_Seq)
```

src/Generating/Trees.cpp
```diff
@@ -947,29 +988,35 @@ void GetJungleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a_Noi
...
-	// Vines are around the BigO4, but not in the corners; need proper meta for direction
-	static const sMetaCoords Vines[] =
+	static const sMetaCoords VinesTrunk[] =
 	{
-		{-2, -5, 1}, {-1, -5, 1}, {0, -5, 1}, {1, -5, 1}, {2, -5, 1},  // North face
-		{-2,  5, 4}, {-1,  5, 4}, {0,  5, 4}, {1,  5, 4}, {2,  5, 4},  // South face
-		{5,  -2, 2}, {5,  -1, 2}, {5,  0, 2}, {5,  1, 2}, {5,  2, 2},  // East face
-		{-5, -2, 8}, {-5, -1, 8}, {-5, 0, 8}, {-5, 1, 8}, {-5, 2, 8},  // West face
-		// TODO: vines around the trunk, proper metas and height
-	} ;
```
madmaxoft: Why remove the vines around the leaves?

src/Generating/Trees.h
```diff
@@ -61,8 +61,8 @@ void GetSmallAppleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a
 /** Generates an image of a large (branching) apple tree */
 void GetLargeAppleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a_Noise, int a_Seq, sSetBlockVector & a_LogBlocks, sSetBlockVector & a_OtherBlocks);
 
-/** Generates a branch for a large apple tree */
-void GetLargeAppleTreeBranch(int a_BlockX, int a_BlockY, int a_BlockZ, int a_BranchLength, Vector3d a_StartDirection, Vector3d a_Direction, int a_TreeHeight, cNoise & a_Noise, sSetBlockVector & a_LogBlocks);
+/** Generates a branch. Returns the position where the branch ends. */
+Vector3d GetTreeBranch(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, int a_BlockX, int a_BlockY, int a_BlockZ, int a_BranchLength, Vector3d a_StartDirection, Vector3d a_Direction, int a_TreeHeight, sSetBlockVector & a_LogBlocks);
```
madmaxoft: This is getting a ridiculous amount of parameters and many of them make no sense at the first glance. This should be clarified in the doxy-comment.
Also, "generate" should be explained as "fills a_LogBlocks with ???", and "branch" means just the log, or the leaves, too?

src/Generating/Trees.cpp
NiLSPACE: Vanilla doesn't have them either. Also, I didn't really like the look of all the long vines hanging from the leaves.

src/Generating/Trees.cpp
NiLSPACE: I'll move it into `cTree` while I'm at it.

src/Generating/Trees.h
NiLSPACE: > Also, "generate" should be explained as "fills a_LogBlocks with ???"

In that case shouldn't all the other tree image methods be changes as well?

Also, how about I replace all the `a_BlockXYZ` with a `Vector3i`? That will reduce the amount of parameters and will be more in line with #3853.

src/Generating/Trees.cpp
madmaxoft: Weird, I believed it did have them, dunno why. Alright then,

src/Generating/Trees.h
madmaxoft: Any change in documentation that makes it clearer is good.

`Vector3i` change sounds good as well.

src/Generating/Trees.cpp
```diff
@@ -159,13 +240,13 @@ void GetTreeImageByBiome(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a_No
 		case biColdBeach:
 		{
 			// Apple or birch trees:
-			if (a_Noise.IntNoise3DInt(a_BlockX, a_BlockY + 16 * a_Seq, a_BlockZ + 16 * a_Seq) < 0x5fffffff)
+			if (a_Noise.IntNoise3DInt(a_BlockPos.addedY(16 * a_Seq).addedZ(16 * a_Seq)) < 0x5fffffff)
```
madmaxoft: This looks weird. Personally, I'd pull the `a_Seq`-based offset into a local variable `seqOffset` and use a regular vector addition in all the places. But since this (`addedX().addedY().addedZ()`) is used everywhere else here, I don't mind so much.

src/Generating/Trees.cpp
```diff
@@ -423,29 +478,26 @@ void GetLargeAppleTreeImage(int a_BlockX, int a_BlockY, int a_BlockZ, cNoise & a
...
 		CurrentPos += Direction;
-		if (CurrentPos.y >= a_TreeHeight)
-		{
-			return;
-		}
-		Direction -= a_Direction;
+		Direction += a_Direction;
 		Direction.Clamp(-1.0, 1.0);
-		a_LogBlocks.push_back(sSetBlock(FloorC(CurrentPos.x), FloorC(CurrentPos.y), FloorC(CurrentPos.z), E_BLOCK_LOG, GetLogMetaFromDirection(E_META_LOG_APPLE, Direction)));
+		a_LogBlocks.push_back(sSetBlock(FloorC(CurrentPos.x), FloorC(CurrentPos.y), FloorC(CurrentPos.z), a_BlockType, GetLogMetaFromDirection(a_BlockMeta, Direction)));
```
madmaxoft: Use `CurrentPos.Floor()` together with the shiny new `Vector3i`-based constructor :)

NiLSPACE:
Any idea why the build works in Visual Studio and WSL/Clang but not here?

madmaxoft:
> 
> 
> Any idea why the build works in Visual Studio and WSL/Clang but not here?

The build is failing for tools, not for Cuberite itself. You're most likely not compiling tools at all (`cmake -DBUILD_TOOLS=1`).

The problem is that the tools' Globals.h doesn't define `TOLUA_TEMPLATE_BIND` like Cuberite's Globals.h does. Just copy it over and it should work.

madmaxoft:
Still more `Clamp`-ing needed in NoiseSpeedTest

NiLSPACE:
I'm currently in the middle of moving, so I don't have allot of time to work on this for the time being.

madmaxoft:
I'll try to fix this PR, compile-wise.

NiLSPACE:
Thanks! It looks good to me!

PR #4417, opened 2019-10-10
Refactored block-to-pickup conversion.

### LINKED ISSUE #4406: Centralize block-to-pickup conversion
Provide a function that implements a block in a world being dug up and converted into pickups. This should be then used for all digging operations - by the player, by the physics simulators, etc.

The function needs to take into consideration the block itself, any block entity attached to it (so that breaking a chest spills its contents), the tool used for the digging (digging up leaves with a shear produces different pickups than with an empty hand; enchantments)

### LINKED ISSUE #4418: Piston's handle can be destroyed seperately from the piston itself
Client version:   1.12.2
Server OS:        Windows
Commit id:       83013d19ca77f8ff736ed7b9e85d2c7b3250476a

> English is not my native language, please excuse typing errors.

When I tried to remove a redstone machine using TNTs, I found that an activated piston was without its handle.

### Expected behavior

When the handle of a piston is destroyed, the piston will also be destroyed.

1. Put a piston and activate it;
1. Put a TNT and surround it with stones to reduce explosion power;
1. Activate the TNT to destroy the piston's handle.

madmaxoft:
Created a function `cWorld::DropBlockAsPickups` that drops the block as pickups; every code path that drops blocks as pickups uses that function, rather than calling the BlockHandlers and plugins (or forgetting to call the BlockHandlers / plugins) itself.
Added a function to query the pickups that would be created by digging the block (`cWorld::PickupsFromBlock`)
Refactored the BlockHandlers to support this.
Overall cleaned the BlockHandlers up a bit.
Changed a few functions to use `Vector3i` instead of XYZ ints.

Closes #4406.
Fixes #4418.


src/Blocks/BlockBed.cpp
```diff
@@ -155,14 +153,12 @@ void cBlockBedHandler::OnPlacedByPlayer(cChunkInterface & a_ChunkInterface, cWor
...
 {
-	short Color = E_META_WOOL_RED;
-	a_WorldInterface.DoWithBedAt(a_BlockX, a_BlockY, a_BlockZ, [&](cBedEntity & a_Bed)
-		{
-			Color = a_Bed.GetColor();
-			return true;
-		}
-	);
-	a_Pickups.Add(cItem(E_ITEM_BED, 1, Color));
+	short color = E_META_WOOL_RED;
```
NiLSPACE: I've seen `short` used instead of `NIBBLETYPE` a few times in the Cuberite codebase. Is there any reason to use `short` instead of `NIBBLETYPE`?

src/ClientHandle.cpp
```diff
@@ -1352,22 +1351,29 @@ void cClientHandle::HandleBlockDigFinished(int a_BlockX, int a_BlockY, int a_Blo
...
+	{
+		World->DigBlock(absPos);
+	}
+
+	// Damage the tool:
+	auto dlAction = (cBlockInfo::IsOneHitDig(a_OldBlock) ? cItemHandler::dlaBreakBlockInstant : cItemHandler::dlaBreakBlock);
+	m_Player->UseEquippedItem(dlAction);
 
+	World->BroadcastSoundParticleEffect(EffectID::PARTICLE_SMOKE, absPos, a_OldBlock, this);
+	blockHandler->OnPlayerBrokeBlock(ChunkInterface, *World, *m_Player, absPos, a_OldBlock, a_OldMeta);
```
madmaxoft: This is probably the most important change - instead of letting the ItemHandler do the actual block breaking, the new centralized `cWorld::DropBlockAsPickup` is called.

src/World.h
```diff
@@ -554,7 +575,37 @@ class cWorld  // tolua_export
...
+	Replaces the specified block with air, and calls the apropriate block handlers (OnBreaking(), OnBroken()).
+	Wakes up the simulators.
+	Doesn't produce pickups, use DropBlockAsPickups() for that instead.
+	Returns true on success, false if the chunk is not loaded. */
+	bool DigBlock(int a_X, int a_Y, int a_Z)
+	{
+		return DigBlock({a_X, a_Y, a_Z});
+	}
+
+	/** Digs the specified block, and spawns the appropriate pickups for it.
```
NiLSPACE: What happens with blocktypes without an appropriate pickup such as air?

madmaxoft: Good catch with air, it needs a special handler not to drop anything. Although this doesn't get called for air because the client won't break air blocks, this function needs to be safe for it, too.

Anyway, this function spawns anything that the block handler's `ConvertToPickups` returns, for some blocks it's already empty, so no pickups are spawned.

madmaxoft: Fixed by adding a new mixin for blocks with no drops.

src/Blocks/BlockPlant.h
```diff
@@ -1,32 +1,41 @@
...
 
 
-#pragma once
 
-#include "BlockHandler.h"
 
-class cBlockPlant : public cBlockHandler
+/** Base class for plants that use light values to decide whether to grow or not. */
+template <bool NeedsLightToGrow>
+class cBlockPlant:
```
peterbell10: Is there any reason for this to be a template rather than a derived class that overrides `CanGrow`?

src/Chunk.h
```diff
@@ -152,15 +152,21 @@ class cChunk :
 
 	cWorld * GetWorld(void) const { return m_World; }
 
-	void SetBlock(int a_RelX, int a_RelY, int a_RelZ, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, bool a_SendToClients = true);
+	void SetBlock(Vector3i a_RelBlockPos, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, bool a_SendToClients);
```
peterbell10: Why is the default being removed?

src/Item.h
```diff
@@ -234,6 +235,14 @@ class cItems  // tolua_export
 	: public std::vector<cItem>
 {  // tolua_export
 public:
+
+	cItems(const cItems &) = default;
+	cItems(cItems &&) = default;
+	cItems & operator = (const cItems &) = default;
```
peterbell10: Move assign?

src/Items/ItemBucket.h
```diff
@@ -164,13 +166,7 @@ class cItemBucketHandler :
...
 				return false;
 			}
-
-			cBlockHandler * Handler = BlockHandler(CurrentBlockType);
-			if (Handler->DoesDropOnUnsuitable())
-			{
-				cChunkInterface ChunkInterface(a_World->GetChunkMap());
-				Handler->DropBlock(ChunkInterface, *a_World, a_PluginInterface, a_Player, BlockPos.x, BlockPos.y, BlockPos.z);
-			}
+			a_World->DropBlockAsPickups(BlockPos, a_Player, nullptr);
```
peterbell10: Why do you ignore `DoesDropOnUnsuitable` now?

src/Blocks/BlockBigFlower.h
```diff
@@ -155,31 +101,36 @@ class cBlockBigFlowerHandler :
...
 		else
 		{
 			// Was lower part
-			Vector3i UpperPart = BlockPos + Vector3i(0, 1, 0);
-			if (a_ChunkInterface.GetBlock(UpperPart) == m_BlockType)
+			auto upperPartPos = a_BlockPos + Vector3i(0, 1, 0);
+			if (a_ChunkInterface.GetBlock(upperPartPos) == a_OldBlockType)
 			{
-				a_ChunkInterface.FastSetBlock(UpperPart, E_BLOCK_AIR, 0);
+				a_ChunkInterface.DropBlockAsPickups(upperPartPos);
```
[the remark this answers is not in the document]
madmaxoft: Plugins may override the behavior through hooks.

src/Blocks/BlockBrewingStand.h
```diff
@@ -2,28 +2,27 @@
...
-	}
 
-	virtual void ConvertToPickups(cItems & a_Pickups, NIBBLETYPE a_BlockMeta) override
+	cBlockBrewingStandHandler(BLOCKTYPE a_BlockType):
+		super(a_BlockType)
 	{
-		a_Pickups.push_back(cItem(E_ITEM_BREWING_STAND, 1, 0));
 	}
 
-	virtual bool IsUseable() override
```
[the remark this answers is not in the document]
madmaxoft: Because it's implemented in the base class, `cBlockEntityHandler`

src/Blocks/BlockPlant.h
madmaxoft: Only me not wanting to refactor more than needed.

This should probably be refactored into a base class that reads the light and sends it to a virtual `checkLight` function that would be overridden in two descendants, `cBlockPlantWantsLight` and `cBlockPlantHatesLight` that override just the `checkLight` function.

src/Chunk.h
madmaxoft: I was trying to see how many call sites actually used the parameter set to `false`, and it slipped through. I guess it can stay now, it makes the callers think "do I need to send this change to the clients or not?"

src/Items/ItemBucket.h
madmaxoft: Makes more sense - the block needs to go in any way, so drop it as nature-produced pickup - which is exactly what `DropBlockAsPickup` with nullptr Digger and Tool promise.

src/Blocks/BlockBigFlower.h
```diff
@@ -42,106 +42,52 @@ class cBlockBigFlowerHandler :
...
+			return {};  // No drops from the top part
 		}
 
-		NIBBLETYPE Meta = a_BlockMeta & 0x7;
-		if (Meta == E_META_BIG_FLOWER_DOUBLE_TALL_GRASS)
+		// With shears, drop self (tall grass and fern only?):
+		if ((a_Tool != nullptr) && (a_Tool->m_ItemType == E_ITEM_SHEARS))
+		{
+			// Bit 0x08 specifies whether this is a top part or bottom; cut it off from the pickup:
+			return cItem(m_BlockType, 1, a_BlockMeta & 0x07);
```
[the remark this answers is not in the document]
madmaxoft: I've re-checked the wiki and updated the comments, it should behave exactly as wiki says.
https://minecraft.gamepedia.com/Flower
https://minecraft.gamepedia.com/Grass

madmaxoft:
@peterbell10 I've hopefully addressed all your comments.

src/Blocks/BlockPlant.h
peterbell10: If you derive `cBlockPlantWantsLight` from `cBlockPlantHatesLight` (or vice versa) then there should be no duplicated code.

src/Chunk.h
peterbell10: I feel like it's very rare to ever pass `false`. Not sure I could think of an example.

madmaxoft: Exactly, I'd say the param could be removed. But don't want to go there in this PR.

src/Blocks/BlockPlant.h
madmaxoft: Deriving one from another feels wrong, basic OOP says you don't derive a dog from a cat, you derive both from an animal. Still, outside the scope for this PR, I think. I just needed to get rid of the bool param to the constructor, so that I can use it as a mixin.

src/Chunk.h
peterbell10: If the end goal is to remove the parameter, why add it back in to all of the call sights?

madmaxoft: Alright, so I went ahead and checked, there's not a single place in Cuberite that would use `false`; however, the function is exported to Lua through `cWorld:SetBlock`, and I can't vouch for the plugins. However, since it makes no sense to not notify clients, I'm removing the parameter altogether.

peterbell10: Should we keep the parameter in the lua version of `cWorld:SetBlock` and just ignore the value?

peterbell10: And mark as OBSOLETE of course.

madmaxoft: No, let it burn.

madmaxoft:
I have tested the basics:
- breaking dirt or grass block produces dirt pickup
- breaking stone produces cobblestone
- breaking a grass block under a tallgrass makes the tallgrass go away too, and drops seeds with some chance
- breaking half a door breaks the entire door and produces a single door pickup
- breaking a sugarcane block breaks all the sugarcane blocks above it, producing pickups for each of them.
I expect problems with this to be small and solvable in small PRs upon their discovery.

madmaxoft:
@NiLSPACE Any objections to merging?

mathiascode:
@madmaxoft DropBlockAsPickups and PickupsFromBlock show up in my self-generated API docs, but are missing on https://api.cuberite.org for some reason.

madmaxoft:
Thanks for bringing this up, it was caused by the build script using an out-of-source build and recently we changed the server executable destination. So the script was executing the old binary.

PR #4419, opened 2019-10-13
Grow in handlers

### LINKED ISSUE #4407: Move growing crops from Chunk to BlockHandlers
Currently there are functions in `cChunk` / `cChunkMap` / `cWorld` for growing most kinds of produce - wheat, melons, sugarcane ...
These need to be moved away into the respective `cBlockHandler` descendants.

#4406 needs to be done before this, because a lot of crops break self on some conditions.

1.13 conversion needs this, so that block types are not hard-coded into the server core.

madmaxoft:
Moved block growing from `cWorld` / `cChunk` into the specific `cBlockHandler` descendants and `cItemHandlerDye` for bonemeal-related actions.

~Note that this PR includes the changes from #4417, I will rebase this once that PR is merged.~ (rebased)

Also note that this PR removes the possibility to configure per-block bonemeal-ability. It is expected to return after 1.13 changes.

Fixes #4407.


src/Blocks/BlockCactus.h
```diff
@@ -77,31 +66,101 @@ class cBlockCactusHandler :
...
+			if (!a_Chunk.UnboundedRelGetBlockType(pos, blockType) || (blockType != E_BLOCK_AIR))
+			{
+				// Cannot grow there
+				return i;
+			}
+
+			a_Chunk.UnboundedRelFastSetBlock(pos, E_BLOCK_CACTUS, 0);
+
+			// Check surroundings. Cacti may ONLY be surrounded by non-solid blocks; if they aren't, drop as pickup and bail out the growing
+			static const Vector3i neighborOffsets[] =
```
NiLSPACE: The cactus popping off seems perfect for a `OnUpdate`. Does the current code still work when I place a block next to the cactus?

madmaxoft: Should still work, thanks to `CanBeAt` (can't try right now). But it's true that the checking code should be reused, rather than having a copy here. Will fix.

madmaxoft: Can't fix easily, `CanBeAt` takes too different a set of parameters; changing `CanBeAt` to accept a single `cChunk` and `Vector3i` makes too many unrelated changes in other handlers, I'd prefer not to go that deep within this PR - especially when the block handlers will most likely be dropped entirely for the 1.13 change and then re-written from scratch.

Issue #4378, opened 2019-08-26
Faster Appveyor builds


madmaxoft:
Currently the Appveyor builds take between 15 and 30 minutes (increased substantially since the upgrade in fcc836f2a5ca39a68aa60c7fb0fa637044a5346b). Most of the time is spent in the cmake configuring phase, when cmake tries to detect various headers, macros etc., mainly for libEvent.

It might be possible to improve build speed either by providing the relevant entries on the command line, or by pre-populating the cache.

peterbell10:
Wait, does cuberite now require VS2017? Any chance of bumping the standard as well?

madmaxoft:
Cuberite itself doesn't require VS2017, but the tests currently do. There was a weird error regarding `initializer_list` in the test (and I suppose it would have occurred later on in the main code as well):
https://ci.appveyor.com/project/madmaxoft/mcserver/builds/26917859#L777

I already bumped the standard to C++17 in #4377 , at least in the PBA tests (so that I can use structured binding for returning multiple values from a function.

madmaxoft:
So we can't bump the standard right now due to the cross-building tools for raspberry pi on the build server. I'm removing the C++17 features from the PR.

madmaxoft:
We could probably move back to MSVC2013, thus run on the faster Appveyor infrastructure.

peterbell10:
If you're already switching things around then it might be the time to move to VS 2015. See #4031

madmaxoft:
Could be. I think the Appveoyr worker VMs do have VS2015 together with VS2013, so it could be the same (faster) VM than the VS2017 one. I probably won't be switching things around much, I just tried to use some C++17 features and failed :)

madmaxoft:
The builds could be made much faster if we used a build cache. The LibEvent project uses it successfully, so it could probably be adapted from there: https://github.com/libevent/libevent/blob/master/appveyor.yml

PR #4422, opened 2019-10-21
CI: Add arm build to travis

peterbell10:
Travis is adding support for arm based CPUs. 

https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support


madmaxoft:
Is it wise to add on-ARM compilation? It almost doubles the build time. It might be better to cross-compile for ARM on an x64 and then only test the actual executable on an ARM. But that might be difficult, if not impossible, to set up.

madmaxoft:
Alright, let's do this, but we should also try to use some cache for the builds so that they all speed up (similar to https://github.com/cuberite/cuberite/issues/4378#issuecomment-529096402)

peterbell10:
Okay, so everything except macos is now using `ccache`. However, even after a lot of fiddling about, the `arm64` builds aren't actually any faster... It looks like travis's caching mechanism is actually broken on their `arm64` images at the moment:
```
Installing md5deep
Reading package lists...
Building dependency tree...
Reading state information...
md5deep is already the newest version (4.4-2).
0 upgraded, 0 newly installed, 0 to remove and 22 not upgraded.
/home/travis/.casher/bin/casher: line 230: md5deep: command not found
```

peterbell10:
On the plus side, everything else is much quicker now so the overall CI running time is still actually faster (as long as nobody changes `World.h`).

PR #4424, opened 2019-11-04
Disable block placement in adventure mode

mathiascode:
Players shouldn't be able to place blocks in adventure mode.


mathiascode:
Any opinions on this PR?

PR #4431, opened 2019-12-02
ProtocolBlockTypePalette merged to BlockTypePalette

### LINKED ISSUE #4381: Create an upgrade table for blocks
We'll need a table that will provide the data for upgrading blocks from 1.12 to 1.13. Basically, a map of {BlockTypeNumber, BlockMeta} to {BlockTypeName, BlockState}, such as this:
```
0	0	minecraft:air
1	0	minecraft:stone
1	1	minecraft:granite
...
54	2	minecraft:chest	facing	north	type	single
54	3	minecraft:chest	facing	south	type	single
...
```
This will be used by the world loader to upgrade older worlds. For now, we just need it as a string, preferably as a github gist. The format is `<BlockTypeNumber><tab><BlockMeta><tab><BlockTypeName><tab><BlockState1><BlockState1Value>...`
For blocks that need more information in 1.13+ than the old data value provides (such as a chest doesn't know in older formats whether it's a single or double chest, or a wall doesn't know which sides it is attached to), choose a reasonable default, like shown in the example.

Note: This will be A LOT of boring manual work! Multiple people can work on this, if you claim a range of the block type numbers. I don't suppose there's any datasource we could use to extract the data, rather than typing it manually, but maybe someone has already done something like this (maybe Minutor?) Investigate first.

madmaxoft:
The two objects were almost identical in functionality, so they were merged into a single one.

Fixes #4381.


madmaxoft:
This is now ready for review.
The main points are:
- `ProtocolBlockTypePalette` functionality has been moved to `BlockTypePalette` and the `ProtocolBlockTypePalette` type removed completely
- `BlockTypePalette` can load the palette from strings / files, in three formats (described in the header):
    - blocks.json, as produced by vanilla server (so that new protocols' palettes can be added simply by users, rather than having to wait for us)
    - Regular TSV, a somewhat stripped version of the same data, produced by $/Tools/BlockTypePaletteGenerator; this file type is better suited for storage in the repository, rather than the raw blocks.json (550 KiB instead of 1.4 MiB; and avoiding possible copyright issues from Mojang)
    - Upgrade TSV, a special version of the regular TSV that uses BlockType + BlockMeta representation for the numeric IDs and is meant for upgrading pre-1.13 worlds (generated from Minutor data)

Note that the class is not yet used by the main codebase, so the only client code for it is the tests. THe first usage will come in a subsequent PR.

Tools/BlockTypePaletteGenerator/UpgradeGenerator.lua
```diff
@@ -0,0 +1,223 @@
...
+	return registry
+end
+
+
+
+
+
+--- Returns the prefix that is common for all block type names in the registry
+-- aRegistry is the parsed registry, as returned from parseRegistry()
+local function findCommonPrefix(aRegistryTable)
```
NiLSPACE: I see some of these functions are in `UpgradeGenerator.lua` and `Generator.lua`. Is it an idea to put those in a different file so they can be shared?

madmaxoft: Not worth the trouble for just a handful of functions. The scripts will most likely never be touched again, only run against newer data files from newer minecraft versions.

PR #4433, opened 2019-12-05
WebAdmin improvements

mathiascode:
- Cleaned up unused, duplicated and messy CSS
- Simplified HTML
- Fixed scaling on smaller screens
- Made UI elements larger for better touch screen experience (Android)
- Changed the color scheme to match Cuberite's green color
- Fixes an issue where the logout link wouldn't work if you were located on a subpage

- (Updated Cuberite Windows icon to match Android's shape)

Fixes #4068

image
image
image

PR #4436, opened 2019-12-11
Increased padding on bottom of webadmin page

ajusa:
This is to resolve issue #4068 

I was unable to reproduce the issue locally, as I don't have access to any machines running MacOS. However, given that two people responded with saying that the fix worked for them, and it has no noticeable issues with the latest version of Chrome, I created this pull request.

Before:
Screenshot from 2019-12-10 19-27-00

After:
Screenshot from 2019-12-10 19-35-08

As a side note, I realize that #4433 will probably fix this issue, but it targets an overall redesign of the webadmin page, and this pull request aims to fix this specific issue for current users. Let me know if there are any issues!

### DIFF
--- a/Server/webadmin/files/style.css
+++ b/Server/webadmin/files/style.css
@@ -51,6 +51,7 @@ h1 {
     vertical-align: middle;
     margin-top: 25px;
 	margin-bottom: 25px;
+    padding-bottom: 65px;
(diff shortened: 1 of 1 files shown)


madmaxoft:
Thanks, and welcome to the team :)

PR #4440, opened 2019-12-25
Output the binaries into a per-configuration Server subfolder.

madmaxoft:
Make (sym)links to the original Server subfolder's items from the per-configuration Server subfolder.

This way we can compile x86 and x64 targets from the same set of sources, each has a different binary output folder. Yet they all share the same base plugins, webadmin etc.

Previously:
```
$
+-- Server (binaries from all builds go here)
  +-- Plugins ...
```

This PR:
```
$
+-- Server (unused)
+-- BuildDir1
  +-- Server (binaries for BuildDir1 output here)
    +-- Plugins ... (symlink to $/Server/Plugins ...)
+-- BuildDir2
  +-- Server (binaries for BuildDir2 output here)
    +-- Plugins ... (symlink to $/Server/Plugins ...)
```


madmaxoft:
(waiting for CI artifact collection to show the problems with the PR)

madmaxoft:
@mathiascode I'd appreciate your feedback here; you touched the Appveyor builds last :) This PR might change the paths to the artifacts needed for the main webpage downloads.

mathiascode:
I'd prefer it if we didn't include paths in artifact names. While the current way of pushing artifacts doesn't support changing the file name without copying the files to the root folder, using the appveyor command should work.

Something like this under after_build (or a separate script) should do, but I can't test it right now.
```
- appveyor PushArtifact '%BUILD_DIR%\Server\Cuberite.zip' -FileName Cuberite.zip
- appveyor PushArtifact '%BUILD_DIR%\Server\PDBs.zip' -FileName PDBs.zip
- appveyor PushArtifact '%BUILD_DIR%\Server\AutoAPI.zip' -FileName AutoAPI.zip
- appveyor PushArtifact '%BUILD_DIR%\Server\ManualAPI.zip' -FileName ManualAPI.zip
- appveyor PushArtifact '%BUILD_DIR%\Server\.luacheckrc' -FileName .luacheckrc
```

madmaxoft:
Thanks, trying that out now.
I've squashed the intermediate commits, they're not needed anymore; just left the artifact pushing as a separate commit for now.

madmaxoft:
Done. Big rebuilds now :)

mathiascode:
Linux builds are crying a bit right now.

madmaxoft:
What do you mean?

mathiascode:
https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/1066/console

madmaxoft:
Hm, in-source builds are really bothersome. I'll have to come up with a fix for that.

madmaxoft:
#4442.

mathiascode:
AppVeyor seems to restore outdated server files now. I noticed that the webadmin changes aren't included in new builds.
image

madmaxoft:
Yeah, the cache there is not helping at all, and causes all kinds of problems. I'll try setting the cache only to the CMake-related files.

mathiascode:
I checked the contents of the Cuberite zip file, and it seems like the following files are missing compared to earlier:
- furnace.txt
- README.txt
- Cuberite_high_detail_debug.cmd
- Cuberite_medium_detail_debug.cmd
- lua5.1.dll

mathiascode:
Also a bit interesting that install_windows_service.cmd and delete_windows_service.cmd are included in Linux builds, but not the Windows ones.

PR #4442, opened 2019-12-28
Fix in-source builds

madmaxoft:
When configuring, don't create symlinks if link == orig.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -344,31 +344,41 @@ SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME} PROPERTIES
 
 
 # Create a symbolic link from ${orig} to ${link}
+# If the orig and link point to the same thing, does nothing (-> in-source build)
 # If ${link} already exists, does nothing.
 function(make_symlink orig link)
 	# Get OS dependent path to use in `execute_process`
-	message("Creating symlink, orig = ${orig}; link = ${link}")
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -344,30 +344,33 @@ SET_TARGET_PROPERTIES(${CMAKE_PROJECT_NAME} PROPERTIES
...
 	file(TO_NATIVE_PATH "${orig}" orig)
 	file(TO_NATIVE_PATH "${link}" link)
 
-	if (NOT EXISTS ${link})
-		if (CMAKE_HOST_UNIX)
-			set(command ln -s ${orig} ${link})
-		else()
-			if (IS_DIRECTORY ${orig})
-				set(command cmd.exe /c mklink /j ${link} ${orig})
+	if (NOT "${orig}" STREQUAL "${link}")
```
peterbell10: This makes the nesting quite deep. Will an early return work instead?
```cmake
if (EXISTS "${link}")
	return()
endif()
```

madmaxoft: That's better, I couldn't find if cmake had returns. I'll rework this.

madmaxoft:
The Appveyor times are insane, over an hour for a CI build :(

madmaxoft:
The builds still fail when configured into a folder requiring escaping (such as containing a space). But I'm already in my pajamas.

PR #4444, opened 2019-12-29
Appveyor: Cache only the CMakeCache file.

madmaxoft:
Caching the entire build folder would overwrite the symlinked Plugins, webadmin etc.


madmaxoft:
This seems to already work, the first build was taking ages (~29 minutes per single configuration), but it failed due to some internal Appveyor error, I've restarted it now and the cache has already been re-used.

mathiascode:
AppVeyor, stop being so drunk...
image

PR #4452, opened 2020-01-03
Strip debug symbols from Android builds

mathiascode:
The Android binaries are currently massive (100MB each), compared to 10MB in 2018.

If someone has a better solution/knows what's going on, let me know.

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -22,7 +22,7 @@ include_directories(SYSTEM
 )
 
 # Disable some compiler warnings (the lazy way out)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-double-promotion -Wno-sign-conversion")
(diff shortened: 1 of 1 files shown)


mathiascode:
Seems like Android builds were never stripped, but simply smaller with debug symbols included before.

android/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ include_directories(SYSTEM
 )
 
 # Disable some compiler warnings (the lazy way out)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-double-promotion -Wno-sign-conversion")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-double-promotion -Wno-sign-conversion -Wno-unused-command-line-argument -s")
```
madmaxoft: `-s` strips relocations as well, is that intended?

mathiascode: Doesn't seem to make a difference as far as I can see.

madmaxoft: If it works, then no problem.

Issue #4453, opened 2020-01-03
BlockTypePalette loading is too slow in MSVC debug


madmaxoft:
It takes over a second to load a single protocol's palette in MSVC debug build, which makes starting Cuberite with multiple protocols a pain.

madmaxoft:
There's nothing that can be done about this, the slowness is caused by memory allocation inside `std::map`, we can't work around it in any simple way.

madmaxoft:
I was wrong, the parser had some room left for improvements, but only by about 30 %. The majority of the time is still spent in the debug CRT / `std::map`.

PR #4450, opened 2020-01-02
Use UpgradeBlockTypePalette

madmaxoft:
Implemented temporary block type mapping for 1.13+ clients, using the UpgradeBlockTypePalette and per-protocol BlockTypePalettes.

This allows 1.13+ clients to connect and see most of the terrain in proper block types, as long as the UpgradeBlockTypePalette is correct (it's very bad at this time):
image

PR #4455, opened 2020-01-04
Faster palette load

madmaxoft:
This improves the loading speed of the `BlockTypePalette` TSV loader by about 30%.

MSVC debug builds take a lot of time to load a single protocol's palette (2.5 seconds on my machine), most of which is caused by `std::map` memory allocations from the debug pool. This change improves the load time to 1.7 seconds, still too much, but much better.
MSVC Release builds, as well as any Linux build, load the same palette really fast (0.03 seconds), so this is purely for better debugging experience.

~Note that this PR is built on top of #4450, I'll rebase it once that gets merged.~ done

Fixes (partially) #4453.

### DIFF
--- a/src/BlockTypePalette.cpp
+++ b/src/BlockTypePalette.cpp
@@ -129,11 +129,11 @@ void BlockTypePalette::loadFromString(const AString & aString)
 	// Detect format by checking the header line (none -> JSON):
 	if (aString.substr(0, hdrTsvRegular.length()) == hdrTsvRegular)
 	{
-		return loadFromTsv(aString, false);
+		return loadFromTsv(aString);
 	}
 	else if (aString.substr(0, hdrTsvUpgrade.length()) == hdrTsvUpgrade)
 	{
-		return loadFromTsv(aString, true);
+		return loadFromTsv(aString);
 	}
 	return loadFromJsonString(aString);
 }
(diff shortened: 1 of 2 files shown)


madmaxoft:
Ping. This is now ready for review.

src/BlockTypePalette.cpp
```diff
@@ -194,92 +194,214 @@ void BlockTypePalette::loadFromJsonString(const AString & aJsonPalette)
...
+				case HeaderValue:
+				{
+					throw LoadFailedException(Printf("Header value contains TAB (line %d)", line));
+				}
+
+				case Index:
+				{
+					try
+					{
+						id = static_cast<UInt32>(std::stoul(aTsvPalette.substr(lastIdx, i - lastIdx)));
```
peterbell10: Possible truncation on linux.

src/BlockTypePalette.cpp
```diff
@@ -194,92 +194,214 @@ void BlockTypePalette::loadFromJsonString(const AString & aJsonPalette)
...
+					state = isUpgrade ? Meta : BlockTypeName;
+					break;
+				}
+
+				case Meta:
+				{
+					id = id * 16;
+					try
+					{
+						id = id + static_cast<UInt32>(Clamp(std::stoi(aTsvPalette.substr(lastIdx, i - lastIdx)), 0, 15));
```
peterbell10: Why not error instead of `Clamp`? Or at least emit a warning.

madmaxoft:
I'm not sure if per-whitespace scanning would make the code cleaner. I'll probably try to write it (as a separate PR) and then we can decide which one is better.

madmaxoft:
Closing, the alternative at #4463 was chosen instead.

PR #4469, opened 2020-02-15
Send entities upon load

### LINKED ISSUE #3696: Entitites turn invisible

Buom01:
These changes where tested with https://github.com/cuberite/cuberite/commit/e290cb4e9b608fbff114205500540ebd324f54a0 and with my armor stand patch https://github.com/Buom01/cuberite

The difference on my case is instant, but I fail to reproduce it with hand spawned mobs.
Now all of my own plugin's mobs are shown instead of ~20% (seem random).
Edit : I will also try @mathiascode world to be sure about the change even if it should be also OK for his world case.

The mystery is: Why some of these mobs were shown while it should be totally hidden since it where probably (or logically) never sent to the client ?
I think that after my pull request some mob could be sent twice.

Additional fix:
- CamelCase on `Client` in the for loop
- Name argument in `cChunkSender::Entity`

Fix #3696


mathiascode:
Does not seem to work in my world case. Boats are still invisible until you touch the spot they're supposed to be in.

src/ChunkSender.h
```diff
@@ -120,7 +120,7 @@ class cChunkSender:
 	// NOTE that m_BlockData[] is inherited from the cChunkDataCollector
 	unsigned char m_BiomeMap[cChunkDef::Width * cChunkDef::Width];
 	std::vector<Vector3i> m_BlockEntities;  // Coords of the block entities to send
-	// TODO: sEntityIDs    m_Entities;       // Entity-IDs of the entities to send
+	std::vector<UInt32> m_Entities;  // Entity-IDs of the entities to send
```
madmaxoft: This is named wrong, it's not entities, it's their IDs.

mathiascode:
https://github.com/cuberite/cuberite/blob/master/src/Chunk.cpp#L1729 is the other place in the code where entities are sent to clients, and has the issues that cause entities to be "invisible" (entity spawn is never broadcast to the client).

mathiascode:
@Buom01 Could you test your changes with https://github.com/cuberite/cuberite/blob/master/src/Chunk.cpp#L1729-L1740 removed?

Titaniumtown:
It seems like (according to my testing) that this pr fixes #3696

MeMuXin:
When the entity is actually loaded on the server works excellent, fails sometimes when spawning mob/player

Titaniumtown:
oh, I only tested with another player

Titaniumtown:
@MeMuXin I guess this would only be a partial fix then.

Titaniumtown:
Is this going to be pushed?

tigerw:
hold on

Titaniumtown:
ok

tigerw:
Merged manually.

Titaniumtown:
wdym @tigerw?

tigerw:
Changes merged in #4532 if I'm not mistaken

PR #4473, opened 2020-02-24
Create build script for Jenkins (and use it)

bearbin:
This build script is currently used for Jenkins builds for linux x86 and x64.

Is there a way this could be combined with the CIBuild.sh script - I'm not sure exactly whta that script is used for at the moment so I didn't want to break anything but if it's possible reducing the duplication would be a very good idea.

Once this is merged, the https://github.com/cuberite/build-scripts repository should be free to be deleted permanently.


mathiascode:
CIBuild.sh seems to be used by Travis alone. I don't think combining the scripts is a good idea, they don't have much in common.

jenkinsbuild.sh
```diff
@@ -0,0 +1,83 @@
...
+
+export CCACHE_CPP2=true
+CACHE_ARGS="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
+
+# Build
+CXX=$CXXCOMP CC=$CCOMP cmake . -DNO_NATIVE_OPTIMIZATION=1 -DBUILD_TOOLS=1 -DCMAKE_BUILD_TYPE=${COMPILEMODE^^} -DFORCE_32=${FORCE32^^}
+make -j 2
+
+# Prepare artifacts directory
+mkdir artifacts
```
mathiascode: This broke existing download links (and scripts that depend on them).
Old: https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/lastSuccessfulBuild/artifact/Cuberite.tar.gz
New: https://builds.cuberite.org/job/Cuberite%20Linux%20x64%20Master/lastSuccessfulBuild/artifact/artifacts/Cuberite.tar.gz

bearbin: Unfortunately it did - the Jenkins UI hid the change to the URL and I didn't think to check it manually :( 

I have reverted this change and I will create a proper canonical link for future use that just redirects to the current download location; but for the forseeable future we have to stick with the current location.

PR #4476, opened 2020-02-28
Prevent container item duplication

mathiascode:
Regression in #4417

I'm not sure what the original intention was, but this PR prevents container items (in chests, hoppers etc) from duplicating when breaking the container block.

### DIFF
--- a/src/BlockEntities/BlockEntityWithItems.cpp
+++ b/src/BlockEntities/BlockEntityWithItems.cpp
@@ -26,20 +26,6 @@ cBlockEntityWithItems::cBlockEntityWithItems(
 
 
 
-void cBlockEntityWithItems::Destroy(void)
-{
-	// Drop the contents as pickups:
-	ASSERT(m_World != nullptr);
-	cItems Pickups;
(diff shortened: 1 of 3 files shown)


madmaxoft:
I'd say this is the proper place to drop the contents as pickups; where elsewhere do they get dropped?

mathiascode:
> I'd say this is the proper place to drop the contents as pickups; where elsewhere do they get dropped?

https://github.com/cuberite/cuberite/blob/365cbc6e1cea96741e26c9ce912b003f8fd2c62c/src/BlockEntities/BlockEntityWithItems.cpp#L29

madmaxoft:
And that's the wrong place - we do need block entities with no world (for example in the `cBlockArea` instances) and destroying those should spawn no pickups.

mathiascode:
Alright, I've removed the old code for creating pickups.

mathiascode:
Anything else?

Bond-009:
RIP 1.8 behavior :stuck_out_tongue_closed_eyes:

PR #4477, opened 2020-03-01
Explosions move primed TNT and blocks

mathiascode:
https://github.com/cuberite/cuberite/pull/4251 improved explosion knockback for players, but broke knockback for primed TNT and falling blocks in the process. Currently, primed TNT and falling blocks are "glued" to their spawn positions. This PR brings back behavior prior to https://github.com/cuberite/cuberite/pull/4251 for these cases.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1315,7 +1315,7 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 				EntityExposure = EntityExposure * (1.0f - a_Entity.GetEnchantmentBlastKnockbackReduction());
 
 				double Impact = (1 - ((Length / ExplosionSizeInt) / 2)) * EntityExposure;
-				
+
 				if (a_Entity.IsTNT() || a_Entity.IsFallingBlock())
 				{
 					// Larger impact for primed TNT and falling blocks, to make them move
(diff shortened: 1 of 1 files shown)


src/ChunkMap.cpp
```diff
@@ -1316,6 +1316,12 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 
 				double Impact = (1 - ((Length / ExplosionSizeInt) / 2)) * EntityExposure;
 
+				if (a_Entity.IsTNT() || a_Entity.IsFallingBlock())
```
peterbell10: Shouldn't this be done by overriding `cFallingBlock::GetExplosionExposureRate` and `cTNTEntity::GetExplosionExposureRate`?

mathiascode: Doesn't work well with the impact calculation in this case.

peterbell10: In what way?

bearbin: @mathiascode bump?

mathiascode: With a larger exposure rate, tnt and falling blocks either don't fly far enough, or fly way too far away. I don't have time to look deeper into how the algorithm works right now, feel free to mess around with these changes if you want.

mathiascode:
Closing temporarily, this is not a proper fix. There seem to be issues with the calculations.

PR #4480, opened 2020-03-01
Fix invisible player issue when changing worlds

### LINKED ISSUE #4049: New logged in players are not visible by other players

### LINKED ISSUE #4082: PlayerList bug makes players invisible.
Client version:   1.12.2
Server OS:        Windows/Linux/macOS/BSD/...
Commit id:        0dd172b (Found at the top of the server's console output)

### Expected behavior
Connected players should see new players.
All players should appear in player list and be visible

### Actual behavior
Only new players see connected players.
One player appears in player list.

Compile and run. Connect two clients.

Working on it, will submit PR when I got it figured out.

mathiascode:
Don't broadcast a player into the new world too early, otherwise the player's position will be wrong client side for others in the world (sometimes the player can be hundreds of blocks away from their real position, "invisible").

Also fixes an issue where non-player entities sent through a portal would not spawn client-side for players in the portal world (invisible mob).

Partially fixes() #4049
Should fix() #4082
Potentially fixes() #3696, but we will have to see

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -2069,9 +2069,6 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 			}
 		}
 
-		// Broadcast the player into the new world.
-		a_World->BroadcastSpawnEntity(*this);
-
 		// Queue add to new world and removal from the old one
 
 		// Chunks may be streamed before cWorld::AddPlayer() sets the world to the new value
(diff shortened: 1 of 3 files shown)


src/Entities/Player.cpp
```diff
@@ -2086,6 +2083,9 @@ bool cPlayer::DoMoveToWorld(cWorld * a_World, bool a_ShouldSendRespawn, Vector3d
 		// New world will take over and announce client at its next tick
 		auto PlayerPtr = static_cast<cPlayer *>(ParentChunk->RemoveEntity(*this).release());
 		a_World->AddPlayer(std::unique_ptr<cPlayer>(PlayerPtr), &a_OldWorld);
+
+		// Broadcast the player into the new world.
+		a_World->BroadcastSpawnEntity(*this);
```
madmaxoft: (Just a quick question) 
Shouldn't it be actually moved to where `a_World->AddPlayer()` does the actual work of adding the player? Possibly in the `a_World`'s own thread, instead of the old world's tick thread.

mathiascode: Seems like a much cleaner solution, and would ensure that entity spawns are always broadcasted.

mathiascode:
Obsolete once #4004 is merged.

PR #4481, opened 2020-03-01
Play correct sound when entities are hit by arrows

mathiascode:
Gets rid of the ding sound when being hit by arrows, which doesn't play in vanilla.

### DIFF
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -125,7 +125,7 @@ void cArrowEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 	}
 
 	// Broadcast successful hit sound
-	GetWorld()->BroadcastSoundEffect("entity.arrow.hit_player", GetPosition(), 0.5, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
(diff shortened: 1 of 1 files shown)


src/Entities/ArrowEntity.cpp
```diff
@@ -125,7 +125,7 @@ void cArrowEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 	}
 
 	// Broadcast successful hit sound
-	GetWorld()->BroadcastSoundEffect("entity.arrow.hit_player", GetPosition(), 0.5, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
+	GetWorld()->BroadcastSoundEffect("entity.arrow.hit", GetPosition(), 0.5, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
```
peterbell10: Maybe it should be `successful_hit`

Ref: http://www.theredengineer.com/1.9-playsound-list.html

mathiascode: successful_hit is the previous name for hit_player

peterbell10: Oh, I misunderstood. This sound is played at the arrow's position, not to the player who shot the bow.

PR #4487, opened 2020-03-03
Update LastSentPosition when entity has no speed

peterbell10:


src/Entities/Entity.cpp
```diff
@@ -1878,72 +1878,69 @@ void cEntity::TeleportToCoords(double a_PosX, double a_PosY, double a_PosZ)
...
+	else
+	{
+		// Speed is zero, send this to clients once only as well as an absolute position
+		if (!m_bHasSentNoSpeed)
 		{
-			// Movin'
 			m_World->BroadcastEntityVelocity(*this, a_Exclude);
-			m_bHasSentNoSpeed = false;
+			m_World->BroadcastTeleportEntity(*this, a_Exclude);
+			m_LastSentPosition = GetPosition();
```
peterbell10: This is the main change, previously `m_LastSentPosition` becomes invalid whenever an entity stops moving for even a single tick. That could cause big desyncs if the movement is very start-stop.

peterbell10:
Oh, and the diff only makes sense if ignoring whitespace:
https://github.com/cuberite/cuberite/pull/4487/files?utf8=%E2%9C%93&diff=unified&w=1

mathiascode:
Seems like there were quite a lot of issues that stemmed from BroadcastMovementUpdate. I'll compile and test.

mathiascode:
Boat positions update properly in 1.9+ now, but players are still "invisible" (in the wrong location client-side) sometimes. Can be reproduced by having two players standing still next to each other, and disconnecting/reconnecting a few times.

peterbell10:
Maybe try this + #4480?

mathiascode:
Doesn't seem to help. You still have to fly around for a bit before the other player appears.

madmaxoft:
Is squash & merge ok?

peterbell10:
Yeah, that's fine I think.

PR #4488, opened 2020-03-04
Don't send entity velocity for boats

### LINKED ISSUE #3449: Boats will stutter when over only 1 block of water, drive-able when falling (glide), and move fast on land
Taken from #1109

mathiascode:
Fixes an issue where steering would stop the boat.

Fixes #3449

### DIFF
--- a/src/Entities/Boat.cpp
+++ b/src/Entities/Boat.cpp
@@ -15,6 +15,7 @@
 
 cBoat::cBoat(Vector3d a_Pos, eMaterial a_Material) :
 	super(etBoat, a_Pos, 0.98, 0.7),
+	m_LastSentPosition(a_Pos),
 	m_LastDamage(0), m_ForwardDirection(0),
 	m_DamageTaken(0.0f), m_Material(a_Material),
 	m_RightPaddleUsed(false), m_LeftPaddleUsed(false)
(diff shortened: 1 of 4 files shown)


src/Entities/Boat.h
```diff
@@ -78,6 +79,7 @@ class cBoat :
 
 	void UpdatePaddles(bool rightPaddleUsed, bool leftPaddleUsed);
 private:
+	Vector3d m_LastSentPosition;
```
peterbell10: You should make this `protected` in `cEntity`, otherwise all the logic in `cEntity` will be using the wrong coords.

src/Entities/Boat.cpp
```diff
@@ -39,6 +40,35 @@ void cBoat::SpawnOn(cClientHandle & a_ClientHandle)
 
 
 
+void cBoat::BroadcastMovementUpdate(const cClientHandle * a_Exclude)
```
peterbell10: I just noticed there's a [vehicle move packet](https://wiki.vg/Protocol#Vehicle_Move_.28clientbound.29). Should we be using that instead?

Also, could we exclude the rider from the broadcasts? I assume the rider's client knows the boat's position anyway.

mathiascode: The clientbound vehicle move packet is implemented. Ideally we would implement the serverbound one too, but I had some issues getting it to play nicely with the client.

I tried to exclude the rider, but once you leave the boat, the client sends the boat back to the spawn position until you rejoin.

peterbell10:
Does entity look control the orientation of the boat on other clients? Can an unoccupied boat rotate for a client?

src/Entities/Boat.cpp
peterbell10: That's a shame, oh well.

mathiascode: For future reference: serverbound vehicle move packet is implemented, clientbound is not. I mixed them up.

mathiascode:
> Does entity look control the orientation of the boat on other clients? Can an unoccupied boat rotate for a client?

No, only steering does (as it works in vanilla, afaik).

peterbell10:
I thought you could bump into a boat and it rotates away from you (at least in vanilla).

mathiascode:
Bumping into the boat moves it, but does not rotate it.

mathiascode:
Everything should be resolved now.

PR #4489, opened 2020-03-04
CI: Install ccache on macos

peterbell10:


peterbell10:
Looks like macos is still broken:
https://travis-ci.community/t/macos-build-fails-because-of-homebrew-bundle-unknown-command/7296

PR #4493, opened 2020-03-08
Introduce recipe book functionality

TooAngel:
The recipe book helps especially new players. Missing it gives the
impression that cuberite is not as advanced as it is.

The handling of the recipe book uses the following functions:

- Unlock Recipes
(https://wiki.vg/index.php?title=Protocol&oldid=14204#Unlock_Recipes) to
make recipes available and show the notification for new recipes.
Initialization is done on player login for known ones, the update is done
when new items are discovered.
- Craft Recipe Request
(https://wiki.vg/index.php?title=Protocol&oldid=14204#Craft_Recipe_Request)
when the user selects a recipe from the recipe book to fill the slots.

Known recipes are initialized on player login via `Unlock Recipes` with
`Action` 0.
As soon as a new recipe is discovered this is added via `Unlock Recipes`
with `Action` 1.

To be able to know and recognize new recipes the player class is
extended with `KnownItems` and `KnownRecipes`. As soon as a player
touches an item this is compared to the list of `KnownItems`, if the
item is unknown the recipes are checked for this item and the other
ingredients are checked with the list of `KnownItems`. If a full match
is discovered the recipe is unlocked with the client and stored in the
`KnownRecipes`.

To unlock recipes the recipe ID is sent to the client. A mapping file
(for protocol 1.12.2) translated the minecraft recipe names to ids. The
crafting.txt is extended with and minecraft recipe names is possible.

Limitations:
Only a single recipe is added to the crafting area. Multiple clicks or
shift click does not increase the number of builds.


madmaxoft:
Thanks for the effort. Unfortunately, it will need quite a bit more thinking and work.
For starters, I assume recipe IDs are protocol-specific, so we cannot just have them directly in the crafting.txt, but there needs to be a per-protocol translation. Perhaps vanilla has some recipe names (`minecraft.recipe.workbench.oakwood` ?) Have a look at the data generators: https://wiki.vg/Data_Generators - if there is any generic identification of recipes. Then we'd need per-protocol map for the recipe name -> protocol ID.

You also seem to have committed several files that should not be in source control - the binaries in Tools subfolder.

My personal experience with recipe unlocking in vanilla minecraft has been terrible so far - recipes wouldn't unlock even when I had all the actual ingredients, and on the other hand some recipes unlocked despite me not having ever obtained an ingredient. So I find this feature completely useless.

CONTRIBUTORS
```diff
@@ -64,3 +64,4 @@ williamhatcher (Helped with API documentation updates and bug fixes)
 worktycho
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+TooAngel
```
madmaxoft: Please alpha-sort

src/ClientHandle.cpp
```diff
@@ -3135,6 +3135,43 @@ void cClientHandle::SendUseBed(const cEntity & a_Entity, int a_BlockX, int a_Blo
...
+	for (cCraftingRecipes::cRecipeSlots::const_iterator itr = Recipe->m_Ingredients.begin(); itr != Recipe->m_Ingredients.end(); ++itr)
+	{
+		// TODO this needs to depend on the size of the grid (crafting table 3x3, just crafting 2x2)
+		LOGINFO("Ingredient %d %d %d %d", itr->x, itr->y, itr->m_Item.m_ItemType, 1 + itr->x + 2 * itr->y);
+		int RemovedItems = m_Player->GetInventory().RemoveItem(itr->m_Item);
+		this->SendWholeInventory(*m_Player->GetWindow());
+		LOGINFO("Needed %d removed %d", itr->m_Item.m_ItemCount, RemovedItems);
+		m_Player->GetWindow()->SetSlot(*m_Player, 1 + itr->x + 3 * itr->y, itr->m_Item);
+	}
+}
```
madmaxoft: This function is completely mis-named. I have no idea what it does, but I'm pretty sure it doesn't send anything, and it seems to remove from player inventory instead.

src/CraftingRecipes.cpp
```diff
@@ -287,6 +287,60 @@ cCraftingRecipes::~cCraftingRecipes()
 
 
 
+bool cCraftingRecipes::doWeKnowAllIngredients(cRecipe * Recipe, int ItemID, std::map<int, bool> KnownItems)
```
madmaxoft: Refrain from using "we" in function names.
`areAllIngredientsKnown` is a better name.
Style: arguments should have the `a_` (or `a`) prefix.
`KnownItems` should be a const reference, otherwise calling this function creates a needless copy.

src/CraftingRecipes.cpp
```diff
@@ -287,6 +287,60 @@ cCraftingRecipes::~cCraftingRecipes()
...
+{
+	bool containsNewItem = false;
+	for (cRecipeSlots::const_iterator itr = Recipe->m_Ingredients.begin(); itr != Recipe->m_Ingredients.end(); ++itr)
```
madmaxoft: Prefer to use range-based `for` loops.

src/CraftingRecipes.h
```diff
@@ -170,8 +174,7 @@ class cCraftingRecipes
 
 	/** Searches for anything dye related for leather, calculates the appropriate color value, and sets the resulting value. */
 	void HandleDyedLeather(const cItem * a_CraftingGrid, cCraftingRecipes::cRecipe * a_Recipe, int a_GridStride, int a_GridWidth, int a_GridHeight);
-} ;
-
-
-
 
+private:
+	bool doWeKnowAllIngredients(cRecipe * Recipe, int ItemID, std::map<int, bool> KnownItems);
```
madmaxoft: Add doxy-comments

src/Entities/Player.cpp
```diff
@@ -193,13 +195,67 @@ bool cPlayer::Initialize(OwnedEntity a_Self, cWorld & a_World)
 
 	cPluginManager::Get()->CallHookSpawnedEntity(*GetWorld(), *this);
 
+	LOGINFO("Items size %d", this->KnownItems.size());
+	LOGINFO("Recipes size %d", this->KnownRecipes.size());
+	if (this->KnownRecipes.size() == 0)
```
madmaxoft: Don't compare size if all you're interested in is emptiness, use `empty()` instead.

src/Entities/Player.h
```diff
@@ -578,6 +578,9 @@ class cPlayer :
 	Otherwise it returns the dig speed (float GetDigSpeed(BLOCKTYPE a_Block)) divided by the block hardness (cBlockInfo::GetHardness(BLOCKTYPE a_Block)) divided by 30 if the player can harvest the block and divided by 100 if he can't. */
 	float GetPlayerRelativeBlockHardness(BLOCKTYPE a_Block);
 
+	void AddKnownItem(int ItemID);
+	std::map<int, bool> KnownItems;
```
madmaxoft: Use `std::set` instead.
Don't mix function definitions and member variables together. The general in-class order is: types, variables, functions.

TooAngel:
> Thanks for the effort. Unfortunately, it will need quite a bit more thinking and work.
> For starters, I assume recipe IDs are protocol-specific, so we cannot just have them directly in the crafting.txt, but there needs to be a per-protocol translation. Perhaps vanilla has some recipe names (`minecraft.recipe.workbench.oakwood` ?) Have a look at the data generators: https://wiki.vg/Data_Generators - if there is any generic identification of recipes. Then we'd need per-protocol map for the recipe name -> protocol ID.
> 
> You also seem to have committed several files that should not be in source control - the binaries in Tools subfolder.
> 
> My personal experience with recipe unlocking in vanilla minecraft has been terrible so far - recipes wouldn't unlock even when I had all the actual ingredients, and on the other hand some recipes unlocked despite me not having ever obtained an ingredient. So I find this feature completely useless.

Thx, yeah I also realized that this needs more time, experience and development, but I'm on it. With the PR I just want to make sure that I'm not totally steering in the wrong direction and waste time.

I also came to the conclusion, that the ID to recipe mapping needs to be version / protocoll based. As far as I undestood the IDs are somehow generated based on the ordering of reading the files. That means it can change in each version. Will have a look into the provided link.

I'm pretty new to Minecraft (< a month) and the recipe book was very helpful to me, so the absence in Cuberite very surprising. On the vanilla java server I had good experience with it, so it should be possible - challenge accepted.

madmaxoft:
You might want to actually try compiling on your computer, so that you avoid the needless errors on the CI servers.
Right now the failure is related to the bindings generator - it's a piece of Lua code that scans the C++ files and generates the Bindings.cpp file that implements the Lua bindings. Unfortunately its error messages are cryptic, I'll try helping you on this one in a moment.

src/Inventory.h
```diff
@@ -35,7 +35,7 @@ class cInventory :
 	// tolua_begin
 {
 public:
-
+	int currentRecipe = 155;
```
madmaxoft: This line confuses the Bindings generator. I understand from the code it's only temporary and only used in a single function, so I suggest you move it as a `static` variable to that function.

TooAngel: :+1: Exactly

madmaxoft:
There's no need for the "fixup" in the commit messages, we'll squash your PR into a single commit before merging, anyway.

TooAngel:
> There's no need for the "fixup" in the commit messages, we'll squash your PR into a single commit before merging, anyway.

I'm using the git fixup functionality `git commit --fixup` so that it is clear that the commits will be squashed. Before changing the PR from draft I would rebase the changes to have a nice looking PR.

madmaxoft:
As for the recipe ID maps for each protocol, we already have a `Server/Protocol` subfolder that contains protocol-specific stuff (currently only for 1.13+, but it's a good place), so I suggest you add the per-protocol map files there and add a class, `RecipeMapper`, that reads each protocol's mapping on server start and then provides a function, `UInt32 recipeNameToProtocolID(const AString & aRecipeName)` that looks up the recipes and returns a predefined value if recipe not found.

madmaxoft:
Since you're the only one working on this PR, you can safely just force-push the amended commits.

TooAngel:
> Since you're the only one working on this PR, you can safely just force-push the amended commits.

Yeah, but sometimes I recognize that I messed something up even after committing and this way I'm able to go back while developing. Once done `git rebase --autosquash` will do the work anyway for me.

Is it somehow annoying right now that there are a couple of commits in this PR?

madmaxoft:
Not really, just my brain acting up upon seeing the multiple fixups in a single commit message.

TooAngel:
Question regarding the recipes - They should be identical to the Minecraft ones? I think I saw some wildcards in the Cuberite recipes which are not in the Minecraft ones. Can I adapt the recipes?

FYI: It seems that the first 10 IDs are reserved, the rest is in reverse alphabetical order (only took me 300 recipes to find that out ;-)). Still, I have to check each, because I can only get the list from a newer version. I guess some recipes were added, which will lead to a mismatch.

TooAngel:
So, I'm satisfied. Pull Request is now ready for review.

src/RecipeMapper.h
```diff
@@ -0,0 +1,14 @@
...
+class cRecipeMapper
+{
+public:
+	cRecipeMapper(void);
+	~cRecipeMapper();
+	UInt32 recipeNameToRecipeID(const AString & aRecipeName);
+	AString recipeIDToRecipeName(UInt32 recipeId);
+
+private:
+	std::map<UInt32, AString> m_Protocol1_12_2;
```
peterbell10, line 12: Everything outside of the protocol folder should really be protocol version independent. We should use a consistent ID everywhere internally and only convert it at the last second in the protocol handlers.

src/Root.h
```diff
@@ -229,6 +231,7 @@ class cRoot
 	cMonsterConfig * m_MonsterConfig;
 
 	cCraftingRecipes * m_CraftingRecipes;
+	cRecipeMapper *    m_RecipeMapper;
```
peterbell10, line 234: You are leaking this memory. For new code, prefer `std::unique_ptr<cRecipeMapper>` over raw `new` and `delete`.

src/UI/Window.cpp
```diff
@@ -766,3 +766,64 @@ void cWindow::SetProperty(short a_Property, short a_Value, cPlayer & a_Player)
 
 
 
+
+
+void cWindow::ClearCraftingGrid(cPlayer & a_Player)
```
peterbell10, line 771: Since these window changes are specific to crafting, they probably belong in `cSlotAreaCrafting`.

src/RecipeMapper.h
peterbell10: i.e. have a "string -> internal id" mapping and an "internal id -> protocol id". Use internal ids wherever possible.

src/CraftingRecipes.h
```diff
@@ -132,11 +133,18 @@ class cCraftingRecipes
 	{
 		cRecipeSlots m_Ingredients;
 		cItem        m_Result;
+		AString      m_RecipeName;
```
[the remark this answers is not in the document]
TooAngel: Minecraft allows generating certain data, like the recipes (https://wiki.vg/Data_Generators). I use the file names as `m_RecipeName` which uniquely identifies each recipe. This makes it easy to extend for other Minecraft versions.

Of course, we could use IDs instead, if you think an additional layer of abstraction makes sense, I can look into it.

src/RecipeMapper.h
TooAngel: Ok, so I will move `RecipeMapper` into the `Protocol` directory.

peterbell10: I don't particularly care about what folder `cRecipeMapper` is in. I care about what code the protocol-specific IDs are used in. You are using these IDs in code that is meant to be protocol-agnostic.

src/CraftingRecipes.h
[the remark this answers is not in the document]
peterbell10: >This makes it easy to extend for other Minecraft versions.

An internal ID doesn't need to be stable across minecraft versions because we need to translate it to send over the protocol anyway. The index into `m_Recipes` would be a good unique identifier for example.

src/RecipeMapper.h
TooAngel: K, get it, thx

src/CraftingRecipes.h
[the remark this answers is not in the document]
TooAngel: Before starting to implement this, I would like to understand the advantage and get on the same page.

`Server/crafting.txt` (resolving the cuberite recipes to the Minecraft recipes) IMHO we should use the Minecraft names to be able to compare it easily with the official ones and be able to maintain it easily.
`Server/Protocol/x.x.x/base.recipes.txt` (mapping the recipes to the protocol IDs) IMHO we should use the Minecraft names to be able to understand the file easily and don't have to look up every line `somewhere` else to get a clue.
`Server/players/xx/xxx.json` (the list of known recipes per player) IMHO we should use the Minecraft names, which makes it easy to understand the player data. More important if we use custom cuberite IDs and these IDs change for some reason, it will be hard to trace issues in the future.

Does this make sense so far?

So looking only into the server-internal handling:

In `CraftingRecipes.h - cRecipe` the `AString m_RecipeName` could be replaced by an ID (as suggested here), but this also means that there is another mapping of the new ID to the recipe name. If the intent is saving memory, this would be the opposite.

In `RecipeMapper` the mapping to the different protocol IDs could be done with only IDs, `Cuberite specific recipe ID => Minecraft protocol recipe ID`, this could save some memory.

So internally I can understand that this could make sense. IMHO this is a tradeoff between code complexity and minimal increased server start time vs memory consumption.
If this is the current blocker, I will look into it and provide the necessary changes.

tigerw:
Wow, thanks for all your work on this!

src/Entities/Player.cpp
```diff
@@ -193,13 +195,62 @@ bool cPlayer::Initialize(OwnedEntity a_Self, cWorld & a_World)
 
 	cPluginManager::Get()->CallHookSpawnedEntity(*GetWorld(), *this);
 
+	if (m_KnownRecipes.empty())
+	{
+		m_ClientHandle->SendInitRecipes(0);
+	}
+	else
+	{
+		for (auto KnownRecipe : m_KnownRecipes)
```
peterbell10, line 204:
```suggestion
		for (const auto & KnownRecipe : m_KnownRecipes)
```

src/Entities/Player.cpp
```diff
@@ -193,13 +195,62 @@ bool cPlayer::Initialize(OwnedEntity a_Self, cWorld & a_World)
...
+	if (Item.m_ItemType < 0)
+	{
+		return;
+	}
+
+	auto Response = m_KnownItems.insert(Item.CopyOne());
+	if (Response.second)
+	{
+		auto Recipes = cRoot::Get()->GetCraftingRecipes()->findNewRecipesForItem(Item, m_KnownItems);
+		for (auto RecipeName : Recipes)
```
peterbell10, line 229:
```suggestion
		for (const auto & RecipeName : Recipes)
```

src/Entities/Player.cpp
```diff
@@ -193,13 +195,62 @@ bool cPlayer::Initialize(OwnedEntity a_Self, cWorld & a_World)
...
+			AddKnownRecipe(RecipeName);
+		}
+	}
+}
+
+
+
+
+
+void cPlayer::AddKnownRecipe(AString RecipeName)
```
peterbell10, line 240:
```suggestion
void cPlayer::AddKnownRecipe(const AString & RecipeName)
```

src/Item.h
```diff
@@ -169,6 +169,23 @@ class cItem
...
+
+	So items are different when the `ItemType is different` or the `ItemDamage`
+	is different and unequal -1.
+	*/
+	struct itemCompare
+	{
+		bool operator() (const cItem & lhs, const cItem & rhs) const
+		{
+			return (lhs.m_ItemType < rhs.m_ItemType) || ((lhs.m_ItemDamage < rhs.m_ItemDamage) && (lhs.m_ItemDamage != -1));
+		}
```
peterbell10, line 186: This is not good, you should use a lexicographic comparison:

```suggestion
		bool operator() (const cItem & a_Lhs, const cItem & a_Rhs) const
		{
			if (a_Lhs.m_ItemType != a_Rhs.m_ItemType)
			{
				return (a_Lhs.m_ItemType < a_Rhs.m_ItemType);
			}
			if ((a_Lhs.m_ItemDamage == -1) || (a_Rhs.m_ItemDamage == -1))
			{
				return false;  // -1 is a wildcard, damage of -1 alway compares equal
			}
			return (a_Lhs.m_ItemDamage < a_Rhs.m_ItemDamage);
		}
```

src/Item.h
```diff
@@ -169,6 +169,23 @@ class cItem
...
+	/**
+	Compares two items for the same type or category. Type of item is defined
+	via `m_ItemType` and `m_ItemDamage`. Some items (e.g. planks) have the same
+	`m_ItemType` and the wood kind is defined via `m_ItemDamage`. `-1` is used
+	as placeholder for all kinds (e.g. all kind of planks).
+
+	So items are different when the `ItemType is different` or the `ItemDamage`
+	is different and unequal -1.
+	*/
+	struct itemCompare
```
peterbell10, line 181: Naming convention:
```suggestion
	struct sItemCompare
```

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1019,10 +1019,23 @@ UInt32 cProtocol_1_12::GetPacketID(cProtocol::ePacketType a_Packet)
 
 
 
+void cProtocol_1_12::HandleCraftRecipe(cByteBuffer & a_ByteBuffer)
+{
+	HANDLE_READ(a_ByteBuffer, ReadBEUInt8, UInt8, WindowID);
+	HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, RecipeID);
+	HANDLE_READ(a_ByteBuffer, ReadBool,          bool,    MakeAll);
+	m_Client->HandleCraftRecipe(RecipeID);
```
peterbell10, line 1027: No protocol specific IDs should escape `cProtcol` classes. You need to translate protocol id -> cuberite id here.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -679,6 +679,7 @@ UInt32 cProtocol_1_9_0::GetPacketID(cProtocol::ePacketType a_Packet)
 		case pktUpdateScore:           return 0x42;
 		case pktUpdateSign:            return 0x46;
 		case pktUseBed:                return 0x2f;
+		case pktUnlockRecipe:          return 0x31;
```
peterbell10, line 682: This doesn't belong in `cProtocol_1_9_0`

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+cRecipeMapper::~cRecipeMapper()
+{
+}
+
+
+
+
+
+void cRecipeMapper::AddRecipeLine(AString a_ProtocolVersion, int a_LineNum, const AString & a_RecipeLine)
```
peterbell10, line 51:
```suggestion
void cRecipeMapper::AddRecipeLine(const AString & a_ProtocolVersion, int a_LineNum, const AString & a_RecipeLine)
```

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	AString RecipeLine(a_RecipeLine);
```
peterbell10, line 53: Why take an unnecessary copy?

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,103 @@
...
+	AStringVector Sides = StringSplit(RecipeLine, " ");
+	UInt32 Id;
+	if (Sides.size() != 2)
+	{
+		LOGINFO("Recipe incompletely configured %s", a_RecipeLine);
+	}
+	StringToInteger<UInt32>(Sides[0].c_str(), Id);
+	AString Name = Sides[1];
+	m_ProtocolVersionMap[a_ProtocolVersion].insert(std::pair<UInt32, AString>(Id, Name));
```
peterbell10, line 62:
```suggestion
	m_ProtocolVersionMap[a_ProtocolVersion].emplace(Id, Name);
```

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+	AString RecipeLine(a_RecipeLine);
+	AStringVector Sides = StringSplit(RecipeLine, " ");
+	UInt32 Id;
+	if (Sides.size() != 2)
+	{
+		LOGINFO("Recipe incompletely configured %s", a_RecipeLine);
+	}
+	StringToInteger<UInt32>(Sides[0].c_str(), Id);
+	AString Name = Sides[1];
```
peterbell10, line 61:
```suggestion
	const AString & Name = Sides[1];
```

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+
+UInt32 cRecipeMapper::recipeNameToRecipeID(const AString & a_RecipeName, UInt32 a_ProtocolVersion)
+{
+	auto ProtocolMap = m_ProtocolVersionMap.find(cRoot::Get()->GetProtocolVersionTextFromInt(static_cast<int>(a_ProtocolVersion)));
+	if (ProtocolMap == m_ProtocolVersionMap.end())
+	{
+		return 0;
+	}
+	for (auto item: ProtocolMap->second)
```
peterbell10, line 76:
```suggestion
	for (const auto & item: ProtocolMap->second)
```

src/UI/SlotArea.cpp
```diff
@@ -780,6 +781,23 @@ void cSlotAreaCrafting::HandleCraftItem(const cItem & a_Result, cPlayer & a_Play
...
+{
+	for (int pos = 1; pos <= m_GridSize * m_GridSize; pos++)
+	{
+		auto Item = GetSlot(pos, a_Player);
+		if (Item->m_ItemCount > 0)
+		{
+			a_Player.GetInventory().AddItem(*Item);
+			cItem ItemEmpty = Item->CopyOne();
+			ItemEmpty.Empty();
+			SetSlot(pos, a_Player, ItemEmpty);
```
peterbell10, line 794: Is this just clearing the item?
```suggestion
			SetSlot(pos, a_Player, cItem());
```

src/UI/Window.cpp
```diff
@@ -766,3 +766,53 @@ void cWindow::SetProperty(short a_Property, short a_Value, cPlayer & a_Player)
...
 
+
+
+void cWindow::LoadRecipe(cPlayer & a_Player, UInt32 RecipeID)
+{
+	AString RecipeName = cRoot::Get()->GetRecipeMapper()->recipeIDToRecipeName(RecipeID, a_Player.GetClientHandle()->GetProtocolVersion());
+	cCraftingRecipes::cRecipe * Recipe = cRoot::Get()->GetCraftingRecipes()->findRecipeByRecipeName(RecipeName);
+
+	std::list<cItem> usedItems;
+	auto slotAreaCrafting = static_cast<cSlotAreaCrafting *>(m_SlotAreas[0]);
```
peterbell10, line 777: This function definitely doesn't belong in `cWindow`. Probably belongs in `cSlotAreaCrafting`.

src/CraftingRecipes.h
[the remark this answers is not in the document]
peterbell10: >Does this make sense so far?

Yes. Files should use the names, that's fine.

>In `CraftingRecipes.h - cRecipe` the `AString m_RecipeName` could be replaced by an ID (as suggested here), but this also means that there is another mapping of the new ID to the recipe name. If the intent is saving memory, this would be the opposite.

I don't literally mean replace `m_RecipeName` with an ID. I mean to refer to the recipe in code by an ID rather than by name. For example, `findNewRecipesForItem` returns an array of recipe names. In doing so, we have to allocate potentially hundreds of new strings. Instead, we could just return an array of numbers requiring just one allocation.

Also, `findRecipeByRecipeName` for example goes from an O(n) linear search to being an O(1) array index operation if you go for IDs as indexes into the recipe list.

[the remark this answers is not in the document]
TooAngel: Ok, I get it (I hope :-))

I will look into it and let you know once I adapted accordingly.

Btw. thanks for taking the time to look into it and explain it in more detail.

src/Protocol/RecipeMapper.cpp
TooAngel: Made it similar to the `CraftingRecipes.cpp`, maybe didn't took enough care if it is necessary.

src/Protocol/Protocol_1_9.cpp
TooAngel: Well, I have to add it, otherwise the CI server complains about:
```
/Users/travis/build/cuberite/cuberite/src/Protocol/Protocol_1_9.cpp:498:10: error: enumeration value 'pktUnlockRecipe' not handled in switch [-Werror,-Wswitch]
        switch (a_Packet)
```
Or is there any other solution?

TooAngel: Ok, looked a bit more into it, but I couldn't find a better way.

If I understand it correctly I have to define the `pktUnlockRecipe` in the `Protocol.h`. When defining it there at least `clang` complains if it is not covered in the switch statement in the `Protocol_x_x.cpp`. So I need to define it in 1.8 and 1.9, too.

Am I missing something here?

src/CraftingRecipes.cpp
```diff
@@ -283,6 +283,80 @@ cCraftingRecipes::~cCraftingRecipes()
...
+		}
+	}
+	return Recipes;
+}
+
+
+
+
+
+std::map<AString, UInt32> cCraftingRecipes::getRecipeNameMap()
```
peterbell10, line 334: Can we build this map once and reuse it?

src/UI/Window.cpp
```diff
@@ -766,3 +766,10 @@ void cWindow::SetProperty(short a_Property, short a_Value, cPlayer & a_Player)
 
 
 
+
+
+void cWindow::LoadRecipe(cPlayer & a_Player, UInt32 a_RecipeId)
+{
+	auto slotAreaCrafting = static_cast<cSlotAreaCrafting *>(m_SlotAreas[0]);
```
peterbell10, line 773: Not all windows have a `cSlotAreaCrafting`, so this doesn't belong in `cWindow`.

peterbell10:
Also, @TooAngel please don't squash your commits when responding to review comments. It means I have to read the entire 2K line diff from the start again, instead of just reading the new changes.

src/CraftingRecipes.cpp
```diff
@@ -287,6 +287,80 @@ cCraftingRecipes::~cCraftingRecipes()
...
+}
+
+
+
+
+
+std::vector<UInt32> cCraftingRecipes::findNewRecipesForItem(const cItem & a_Item, const std::set<cItem, cItem::sItemCompare> & a_KnownItems)
+{
+	std::vector<UInt32> Recipes;
+	for (UInt32 i = 0; i<m_Recipes.size(); i++)
```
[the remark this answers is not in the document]
TooAngel: It would be nice to have it within the linting, if it is desired.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2117,6 +2137,7 @@ UInt32 cProtocol_1_8_0::GetPacketID(ePacketType a_PacketType)
 		case pktTimeUpdate:            return 0x03;
 		case pktTitle:                 return 0x45;
 		case pktUnloadChunk:           return 0x21;
+		case pktUnlockRecipe:          return 0x31;
```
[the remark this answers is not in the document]
TooAngel: Yes, as mentioned here: https://github.com/cuberite/cuberite/pull/4493#discussion_r422386715

clang complains when it is not defined, not sure if I understood it fully with the protocol pkt definitions. I agree that it is not the best solution to define the `pkt*` when they are not valid for the protocol, but I don't know any better way, yet.

src/CraftingRecipes.cpp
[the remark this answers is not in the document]
peterbell10: It's quite hard to do that with `<` and `>` because of templates; you don't put spaces in `std::vector<int>` for example.

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+
+
+
+UInt32 cRecipeMapper::GetProtocolRecipeId(UInt32 a_RecipeId, UInt32 a_ProtocolVersion)
+{
+	auto ProtocolMap = m_ProtocolVersionMap.find(cRoot::Get()->GetProtocolVersionTextFromInt(static_cast<int>(a_ProtocolVersion)));
+	if (ProtocolMap == m_ProtocolVersionMap.end())
+	{
+		return 0;
```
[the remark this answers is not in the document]
TooAngel: Hm, I guess I need help with this. I included the change to `std::optional<UInt32>` (https://github.com/cuberite/cuberite/pull/4493/commits/e164415a18b349fed9ff4a2615587db65eba7cad#diff-c8b569fcdb964b6253967e43533f707eR11)

but running `make` throws: `cuberite/src/Items/../Protocol/RecipeMapper.h:11:7: error: ‘optional’ in namespace ‘std’ does not name a template type
  std::optional<UInt32> GetProtocolRecipeId(UInt32 a_RecipeId, UInt32 a_ProtocolVersion);`

The branch is rebased on master. I did a fresh clone (to avoid any old compile stuff), `cmake .`, `make`.

Reading it up sounds like C++17 is not enabled,
Cmake says the internet is lying: `-- Using c++ standard c++17`

What am I doing wrong? Sorry, I'm a bit rusty with C++

src/CraftingRecipes.cpp
TooAngel: I guess this is a rhetorical question :-)

src/Protocol/RecipeMapper.cpp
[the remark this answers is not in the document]
TooAngel: Nevermind, ignore me. Travis is happy, so am I.

Seems like I have to fix my local build environment ... somehow.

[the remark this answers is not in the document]
TooAngel: Ok, not that easy: It seems like appveyor is not yet configured properly: `C:\projects\cuberite\src\Protocol/RecipeMapper.h(12,7): error C2039: 'optional': is not a member of 'std' ` not sure if I can do anything here.

TooAngel:
And another round, the mentioned issues are tackled. appveyor does not like `optional` (`C:\projects\cuberite\src\Protocol/RecipeMapper.h(12,7): error C2039: 'optional': is not a member of 'std' `) not sure if I can solve it in my PR, if yes, how?

Reviews are appreciated.

peterbell10:
You need to include the `<optional>` header in `RecipeMapper.h`, or any other headers/sources where you use optional.

TooAngel:
> You need to include the `<optional>` header in `RecipeMapper.h`, or any other headers/sources where you use optional.

Thanks, that solved it. Checks are fixed, have fun reviewing.

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+
+
+void cRecipeMapper::AddRecipeLine(const AString & a_ProtocolVersion, int a_LineNum, const AString & a_RecipeLine, const std::map<AString, UInt32> & a_RecipeNameMap)
+{
+	AStringVector Sides = StringSplit(a_RecipeLine, " ");
+	UInt32 Id;
+	if (Sides.size() != 2)
+	{
+		LOGINFO("Recipe incompletely configured %s", a_RecipeLine);
```
peterbell10, line 72: Should probably stop processing here as ther might not be a `Sides[1]`.
```suggestion
		LOGINFO("Recipe incompletely configured %s", a_RecipeLine);
		return;
```

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,124 @@
...
+	}
+	StringToInteger<UInt32>(Sides[0].c_str(), Id);
```
peterbell10, line 74:
```suggestion
	StringToInteger<UInt32>(Sides[0], Id);
```

src/ClientHandle.cpp
```diff
@@ -3126,6 +3126,35 @@ void cClientHandle::SendUseBed(const cEntity & a_Entity, int a_BlockX, int a_Blo
...
+
+
+
+
+
+void cClientHandle::HandleCraftRecipe(UInt32 a_RecipeId)
+{
+	auto SlotArea = m_Player->GetWindow()->GetSlotArea(0);
+	auto slotAreaCrafting = static_cast<cSlotAreaCrafting *>(SlotArea);
+	slotAreaCrafting->LoadRecipe(*m_Player, a_RecipeId);
```
peterbell10, line 3151: This makes a number of dangerous assumptions. You assume:
- `GetWindow()` doesn't return `nullptr`
- The window has a crafting grid
- That crafting grid is always the zeroth slot area

Use `GetWindowType()` to verify it's the window type you expect. Then instead of relying on slot areas ordering, I suggest putting a `LoadRecipe` method into the concrete window class:

```suggestion
	auto * Window = m_Player->GetWindow();
	if (Window == nullptr)
	{
		return;
	}
	
	auto WindowType = Window->GetWindowType();
	if (Window->GetWindowType() == cWindow::wtInventory)
	{
		static_cast<cInventoryWindow *>(Window)->LoadRecipe(*m_Player, a_RecipeId);
	}
	else if (Window->GetWindowType() == cWindow::wtWorkbench)
	{
		static_cast<cCraftingWindow *>(Window)->LoadRecipe(*m_Player, a_RecipeId);
	}
```

src/CraftingRecipes.cpp
```diff
@@ -283,6 +284,72 @@ cCraftingRecipes::~cCraftingRecipes()
...
+	return m_RecipeNameMap;
+}
+
+
+
+
+
+cCraftingRecipes::cRecipe* cCraftingRecipes::getRecipeById(UInt32 a_RecipeId)
+{
+	return m_Recipes[a_RecipeId - 1];
```
peterbell10, line 346: Since we don't need a special value now, I don't think you need the `+1` `-1` fixups.

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,124 @@
...
+	try
+	{
+		const UInt32 RecipeIndex = a_RecipeNameMap.at(Sides[1]);
+		m_ProtocolVersionMap[a_ProtocolVersion].emplace(Id, RecipeIndex);
+	}
+	catch (const std::out_of_range&)
+	{
+		LOGINFO("Cannot find RecipeId for %s", Sides[1]);
+	}
```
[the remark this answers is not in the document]
TooAngel: What is the background to this. My thought was, that not finding the RecipeId is an exceptional case, so handling it via exception could make sense.

src/CraftingRecipes.cpp
TooAngel: true, sorry, missed that after introducing optional

src/ClientHandle.cpp
TooAngel: Ok, I get your point. I think it is hard to send the package without having the window open (in the minecraft client), but better safe than sorry.

How to handle the implementation in `cInventoryWindow` and `cCraftingWindow`? Duplicate the code? Better idea?

src/Protocol/RecipeMapper.cpp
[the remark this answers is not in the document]
peterbell10: I'm mainly concerned about consistency. `find` is already used throughout the codebase; whereas `at` isn't really used at all in Cuberite.

src/ClientHandle.cpp
peterbell10: True, we don't expect the minecraft client to do this. But generally, we try to be fault tolerant when it comes to protocol inputs. You could have a non-standard client, or have edge cases where a Cuberite plugin breaks the vanilla client's assumptions.

The window's `LoadRecipe` just needs to call `cSlotAreaCrafting::LoadRecipe`.

TooAngel: :+1: 

As you showed in your suggested changes it can be either `cInventoryWindow` or `cCraftingWindow`, both are only derived from `cWindow`. Both would need to have the `LoadRecipe` method. which is a two-liner, so doable.
Is there any better way, derived somehow from each other, ....? I would duplicate the two lines right now unless you have any other fancy idea.

peterbell10: Yes, duplicating a two line function is okay.

src/CraftingRecipes.h
```diff
@@ -143,7 +143,7 @@ class cCraftingRecipes
 	cRecipe * getRecipeById(UInt32 a_RecipeId);
 
 	/** Gets a map of all recipes with name and recipe id */
-	std::map<AString, UInt32> getRecipeNameMap();
+	std::map<AString, UInt32> & getRecipeNameMap();
```
peterbell10, line 146:
```suggestion
	const std::map<AString, UInt32> & getRecipeNameMap();
```

src/UI/Window.cpp
```diff
@@ -766,3 +766,9 @@ void cWindow::SetProperty(short a_Property, short a_Value, cPlayer & a_Player)
 
 
 
+
+
+cSlotArea * cWindow::GetSlotArea(UInt32 a_SlotAreaIndex)
+{
+	return m_SlotAreas[a_SlotAreaIndex];
+}
```
peterbell10, line 774: No need for this any more
```suggestion
```

src/UI/Window.h
```diff
@@ -178,6 +178,12 @@ class cWindow
 	/** Used by cSlotAreas to send individual slots to clients, a_RelativeSlotNum is the slot number relative to a_SlotArea */
 	void SendSlot(cPlayer & a_Player, cSlotArea * a_SlotArea, int a_RelativeSlotNum);
 
+	/** Loads the given Recipe into the crafting grid */
+	void LoadRecipe(cPlayer & a_Player, UInt32 a_RecipeId);
+
+	/** Returns the specific SlotArea */
+	cSlotArea * GetSlotArea(UInt32 a_SlotAreaIndex);
+
```
peterbell10, line 186:
```suggestion
```

src/Entities/Player.cpp
```diff
@@ -2235,6 +2284,22 @@ bool cPlayer::LoadFromFile(const AString & a_FileName, cWorldPtr & a_World)
...
 
+	Json::Value & JSON_KnownItems = root["knownItems"];
+	for (UInt32 i = 0; i < JSON_KnownItems.size(); i++)
+	{
+		cItem Item;
+		Item.FromJson(JSON_KnownItems[i]);
+		m_KnownItems.insert(Item);
+	}
+
+	auto RecipeNameMap = cRoot::Get()->GetCraftingRecipes()->getRecipeNameMap();
```
peterbell10, line 2295:
```suggestion
	const auto & RecipeNameMap = cRoot::Get()->GetCraftingRecipes()->getRecipeNameMap();
```

src/Entities/Player.cpp
```diff
@@ -2235,6 +2284,22 @@ bool cPlayer::LoadFromFile(const AString & a_FileName, cWorldPtr & a_World)
...
+
+	Json::Value & JSON_KnownRecipes = root["knownRecipes"];
+	for (UInt32 i = 0; i < JSON_KnownRecipes.size(); i++)
+	{
+			m_KnownRecipes.insert(RecipeNameMap[JSON_KnownRecipes[i].asString()]);
```
peterbell10, line 2300: This could fail if the recipe list is changed.
```suggestion
			auto RecipeId = RecipeNameMap.find(JSON_KnownRecipes[i].asString());
			if (RecipeId != RecipeNameMap.end())
			{
				m_KnownRecipes.insert(*RecipeId);
			}
```

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,124 @@
...
+	{
+		LOGWARNING("Cannot read file \"Protocol/%s/base.recipes.txt\", no recipe book recipes will be available!", a_ProtocolVersion);
+		return;
+	}
+	f.Close();
+
+	// Split it into lines, then process each line as a single recipe:
+	AStringVector Split = StringSplit(Everything, "\n");
+	m_ProtocolVersionMap[a_ProtocolVersion] = {};
+	auto RecipeNameMap = cRoot::Get()->GetCraftingRecipes()->getRecipeNameMap();
```
peterbell10, line 37:
```suggestion
	const auto & RecipeNameMap = cRoot::Get()->GetCraftingRecipes()->getRecipeNameMap();
```

src/CraftingRecipes.h
TooAngel: You are a bit too fast for me, I'm currently fixing my build environment, so that I can test it properly. I only commited and pushed so that I'm able to check it out cleanly, I'll let you know once I'm satisfied :-)

(While I wouldn't have spotted this (and the following) suggestions, so thanks)

TooAngel:
Satisfied again, new reviews are appreciated.

tigerw:
I think we've kept TooAngel waiting enough, it's been two months... Merging tomorrow

madmaxoft:
Sorry but no, I'm still going through this and I'll want some minor changes.

(I wonder how @tigerw of all people can live with the `javaStyleFunctionNames` here :P )

tigerw:
(...)

src/CraftingRecipes.h
```diff
@@ -170,8 +179,14 @@ class cCraftingRecipes
 
 	/** Searches for anything dye related for leather, calculates the appropriate color value, and sets the resulting value. */
 	void HandleDyedLeather(const cItem * a_CraftingGrid, cCraftingRecipes::cRecipe * a_Recipe, int a_GridStride, int a_GridWidth, int a_GridHeight);
-} ;
-
 
+private:
+	/** Checks if all ingredients of the recipe are within the know items list and if the (new) ItemId is part of the ingredients. This makes sure to only find 'new' recipes */
+	bool isNewCraftableRecipe(const cRecipe * a_Recipe, const cItem & a_Item, const std::set<cItem, cItem::sItemCompare> & a_KnownItems);
```
madmaxoft, line 185:
```suggestion
	/** Checks if all ingredients of the a_Recipe are within the a_KnownItems list and
	if the a_NewItem is part of the ingredients.
	This makes sure to only find 'newly discovered' recipes. */
	bool IsNewCraftableRecipe(
		const cRecipe * a_Recipe,
		const cItem & a_NewItem,
		const std::set<cItem, cItem::sItemCompare> & a_KnownItems
	);
```

src/CraftingRecipes.h
```diff
@@ -10,9 +10,7 @@
 
 #include "Item.h"
 
-
-
-
+#include "Protocol/RecipeMapper.h"
```
madmaxoft, line 13: This include doesn't seem to be needed.

src/CraftingRecipes.h
```diff
@@ -170,8 +179,14 @@ class cCraftingRecipes
...
 	void HandleDyedLeather(const cItem * a_CraftingGrid, cCraftingRecipes::cRecipe * a_Recipe, int a_GridStride, int a_GridWidth, int a_GridHeight);
-} ;
-
 
+private:
+	/** Checks if all ingredients of the recipe are within the know items list and if the (new) ItemId is part of the ingredients. This makes sure to only find 'new' recipes */
+	bool isNewCraftableRecipe(const cRecipe * a_Recipe, const cItem & a_Item, const std::set<cItem, cItem::sItemCompare> & a_KnownItems);
 
+	/** Mapping the minecraft recipe names to the internal cuberite recipe Ids */
+	std::map<AString, UInt32> m_RecipeNameMap;
```
madmaxoft, line 188: Please put field declarations above function declarations (within the `private` section of the class)

src/Entities/Player.cpp
```diff
@@ -194,13 +196,60 @@ bool cPlayer::Initialize(OwnedEntity a_Self, cWorld & a_World)
...
+
+	auto Response = m_KnownItems.insert(a_Item.CopyOne());
+	if (Response.second)
+	{
+		auto Recipes = cRoot::Get()->GetCraftingRecipes()->findNewRecipesForItem(a_Item, m_KnownItems);
+		for (const auto & RecipeId : Recipes)
+		{
+			AddKnownRecipe(RecipeId);
+		}
+	}
```
madmaxoft, line 233:
```suggestion
	if (!Response.second)
	{
		// The item was already known, bail out:
		return;
	}
	
	// Process the recipes that got unlocked by this newly-known item:
	auto Recipes = cRoot::Get()->GetCraftingRecipes()->findNewRecipesForItem(a_Item, m_KnownItems);
	for (const auto & RecipeId : Recipes)
	{
		AddKnownRecipe(RecipeId);
	}
```

src/Entities/Player.cpp
```diff
@@ -194,13 +196,60 @@ bool cPlayer::Initialize(OwnedEntity a_Self, cWorld & a_World)
...
+
+
+
+void cPlayer::AddKnownRecipe(UInt32 a_RecipeId)
+{
+	auto Response = m_KnownRecipes.insert(a_RecipeId);
+	if (Response.second)
+	{
+		m_ClientHandle->SendUnlockRecipe(a_RecipeId);
+	}
```
madmaxoft, line 246:
```suggestion
	if (!Response.second)
	{
		// The recipe was already known, bail out:
		return;
	}
	m_ClientHandle->SendUnlockRecipe(a_RecipeId);
```

src/Entities/Player.h
```diff
@@ -600,6 +600,9 @@ class cPlayer:
 	/** get player explosion exposure rate */
 	virtual float GetExplosionExposureRate(Vector3d a_ExplosionPosition, float a_ExlosionPower) override;
 
+	/** Adds an Item ID to the list of known items */
+	void AddKnownItem(const cItem & a_Item);
```
madmaxoft, line 604:
```suggestion
	/** Adds an Item ID to the list of known items.
	If the item is already known, does nothing. */
	void AddKnownItem(const cItem & a_Item);
```

src/Entities/Player.h
```diff
@@ -795,4 +798,13 @@ class cPlayer:
 	If he is not on ground it also gets divided by 5. */
 	float GetDigSpeed(BLOCKTYPE a_Block);
 
+	/** Add the recipe ID to the known recipes */
+	void AddKnownRecipe(UInt32 RecipeId);
```
madmaxoft, line 802:
```suggestion
	/** Add the recipe ID to the known recipes.
	If the recipe is already known, does nothing. */
	void AddKnownRecipe(UInt32 RecipeId);
```

src/Entities/Player.h
```diff
@@ -795,4 +798,13 @@ class cPlayer:
...
+
+	/** List on known recipes as IDs */
+	std::set<UInt32> m_KnownRecipes;
+
+	/** List of known items as IDs */
+	std::set<cItem, cItem::sItemCompare> m_KnownItems;
+
```
madmaxoft, line 809: Please move these in front of the function declarations in the `protected` section.

src/ItemGrid.cpp
```diff
@@ -440,6 +440,29 @@ int cItemGrid::RemoveItem(const cItem & a_ItemStack)
...
+	if (!m_Slots.IsStorageAllocated())
+	{
+		return nullptr;
+	}
+
+	for (int i = 0; i < m_Slots.size(); i++)
+	{
+		if ((a_RecipeItem.m_ItemType == m_Slots[i].m_ItemType) &&
+			((a_RecipeItem.m_ItemDamage < 0) || (a_RecipeItem.m_ItemDamage == m_Slots[i].m_ItemDamage)))
+		{
```
madmaxoft, line 454:
```suggestion
		if (
			(a_RecipeItem.m_ItemType == m_Slots[i].m_ItemType) &&
			((a_RecipeItem.m_ItemDamage < 0) || (a_RecipeItem.m_ItemDamage == m_Slots[i].m_ItemDamage))
		)
		{
```

madmaxoft: Or even better, use the `sItemCompare` somehow?

TooAngel:
> Thanks @TooAngel this looks great. I took the liberty of pushing a few minor changes, hope you don't mind.

Sure, thanks - Just surprised that you can just push changes to a repository on my account. (Guess because it is a fork, still surprised)

TooAngel:
> I think we've kept TooAngel waiting enough, it's been two months... Merging tomorrow

Well, would like to see it merged, but not because the PR is old, rather because the changes are good and appreciated, so take your time if you need it

peterbell10:
@TooAngel there is a tick box when you open the PR that gives project maintainers permission to push changes to your branch. I believe it only applies to branches you open PRs with.

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1019,10 +1021,27 @@ UInt32 cProtocol_1_12::GetPacketID(cProtocol::ePacketType a_Packet)
 
 
 
+void cProtocol_1_12::HandleCraftRecipe(cByteBuffer & a_ByteBuffer)
+{
+	HANDLE_READ(a_ByteBuffer, ReadBEUInt8, UInt8, WindowID);
+	HANDLE_READ(a_ByteBuffer, ReadVarInt, UInt32, RecipeID);
+	HANDLE_READ(a_ByteBuffer, ReadBool,          bool,    MakeAll);
```
madmaxoft, line 1028:
```suggestion
	HANDLE_READ(a_ByteBuffer, ReadBEUInt8, UInt8,  WindowID);
	HANDLE_READ(a_ByteBuffer, ReadVarInt,  UInt32, RecipeID);
	HANDLE_READ(a_ByteBuffer, ReadBool,    bool,   MakeAll);
```

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1397,3 +1417,54 @@ void cProtocol_1_12_2::SendKeepAlive(UInt32 a_PingID)
...
+	}
+}
+
+
+
+
+
+void cProtocol_1_12_2::SendInitRecipes(UInt32 a_RecipeID)
+{
+	// Translate ID here
```
madmaxoft, line 1447: Comment seems out of place, perhaps a leftover?
```suggestion
```

src/Protocol/Protocol_1_12.cpp
```diff
@@ -1397,3 +1417,54 @@ void cProtocol_1_12_2::SendKeepAlive(UInt32 a_PingID)
...
+	ASSERT(m_State == 3);  // In game mode?
+
+	auto ProtocolRecipeId = cRoot::Get()->GetRecipeMapper()->GetProtocolRecipeId(a_RecipeID, m_Client->GetProtocolVersion());
+	if (ProtocolRecipeId.has_value())
+	{
```
madmaxoft, line 1452: Prefer the negative condition and bailing out of the function, to avoid deep nesting.
```cpp
	if (!ProtocolRecipeId.has_value())
	{
		return;
	}
	// rest of function body
```

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1587,6 +1587,26 @@ void cProtocol_1_8_0::SendUseBed(const cEntity & a_Entity, int a_BlockX, int a_B
 
 
 
+void cProtocol_1_8_0::SendUnlockRecipe(UInt32 a_RecipeID)
+{
+	// Not implemented
```
madmaxoft, line 1592: Is it "not implemented in Cuberite" or rather "Client doesn't support this feature"?

TooAngel:
> (I wonder how @tigerw of all people can live with the `javaStyleFunctionNames` here :P )

So what is the style guide here, so that I can fix it and comply in the future? `camelCase`, `snake_case`, `CapitalizedCamelCase`, `RAnDoM`, not sure what `javaStyleFunctionNames` are.

src/Protocol/RecipeMapper.cpp
```diff
@@ -0,0 +1,123 @@
+#include "Globals.h"
+#include "RecipeMapper.h"
+#include "../Root.h"
+
+cRecipeMapper::cRecipeMapper(void)
+{
+	const AString Protocols [] = {"1.12.2"};
+	for (const auto & Protocol: Protocols)
```
madmaxoft, line 8: How about listing the `Protocol/` folder's contents and trying each folder? That way any missing protocol can be added by the user.

src/Protocol/RecipeMapper.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "../CraftingRecipes.h"
+#include <optional>
+
+class cRecipeMapper
```
madmaxoft, line 6: Please add a doxy-comment for this class, explaining what it's used for.

tigerw:
CapitalisedCamelCase

(RAnDoM, I'm laughing so hard, that's a stroke of genius)

src/UI/SlotArea.cpp
```diff
@@ -780,6 +781,70 @@ void cSlotAreaCrafting::HandleCraftItem(const cItem & a_Result, cPlayer & a_Play
 
 
 
+void cSlotAreaCrafting::LoadRecipe(cPlayer & a_Player, UInt32 a_RecipeId)
+{
+	if (a_RecipeId == 0)
+	{
+		return;
+	}
+	cCraftingRecipes::cRecipe * Recipe = cRoot::Get()->GetCraftingRecipes()->getRecipeById(a_RecipeId);
```
madmaxoft, line 790:
```suggestion
	auto Recipe = cRoot::Get()->GetCraftingRecipes()->getRecipeById(a_RecipeId);
	if (Recipe == nullptr)
	{
		return;
	}
```

src/UI/SlotArea.h
```diff
@@ -273,6 +273,11 @@ class cSlotAreaCrafting:
 	// Distributing items into this area is completely disabled
 	virtual void DistributeStack(cItem & a_ItemStack, cPlayer & a_Player, bool a_ShouldApply, bool a_KeepEmptySlots, bool a_BackFill) override;
 
+	/** Clear the crafting grid */
+	void ClearCraftingGrid(cPlayer & a_Player);
+
+	/** Loads the given Recipe into the crafting grid */
+	void LoadRecipe(cPlayer & a_Player, UInt32 a_RecipeId);
```
madmaxoft, line 280: How about making these available in the base class, so that the plugins can use them for different windows / areas, too? It would get rid of the nasty `static_cast<cSlotAreaCrafting *>` in `cInventoryWindow`.

madmaxoft:
(I believe `javaStyleFunctionNames` is called `lowerCamelCase`, whereas our `CamelCase` is called `UpperCamelCase`: https://en.wikipedia.org/wiki/Camel_case )

TooAngel:
K, got it, I should use PascalCase and not javaCase (dromedaryCase) ;-)

Btw. my all-time favorite is Camel_Snake_Case ... not.

src/CraftingRecipes.cpp
```diff
@@ -283,6 +284,72 @@ cCraftingRecipes::~cCraftingRecipes()
 
 
 
+bool cCraftingRecipes::isNewCraftableRecipe(const cRecipe * a_Recipe, const cItem & a_Item, const std::set<cItem, cItem::sItemCompare> & a_KnownItems)
```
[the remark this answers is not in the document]
TooAngel: Should I help out with voting for lowerCamelCase and just not change this method until we win? ;-)

src/Entities/Player.h
TooAngel: Just curious, why in the `protected` section?

src/Protocol/Protocol_1_8.cpp
TooAngel: 'Client doesn't support this feature', but I couldn't find a way to only define it in `Protocol_1_12`

src/UI/SlotArea.h
TooAngel: Is this a rhetorical question?

The method is specific for the inventory and crafting slots, it does not make sense in e.g. the hotbar area. I don't like the duplicate code and the casting, but having in for all slot areas feels dirty, too.
I can change it, not my call.

src/Protocol/RecipeMapper.cpp
TooAngel: Puh, yeah, I could look into it. Tbh, I would improve it as soon as we start to add more of the configurations.
Is it desired or fine like this?

src/ItemGrid.cpp
TooAngel: Puh, ok, not sure if it gets more readable. At least the comparison is only at one place.

TooAngel:
I addressed most of the suggestions.

Open points:
- Dynamic loading of protocol-specific recipe configurations based on the existence of the configuration file - Is this feature necessary to get it released
- Extracting `LoadRecipe` to all SlotArea (not only `crafting` and `inventory`), prevents from casting, but introduces functionality which IMHO does not make sense in other slotAreas.

src/UI/SlotArea.h
TooAngel: Well, would be cool, but `ClearCraftingGrid` relies on `m_GridSize` which is only available in `cSlotAreaCrafting`.

TooAngel:
Rebased, loading of recipes adapted to comments - ready to review again.

I commented regarding the slotAreas.

tigerw:
Thanks! Let me take a look.

Titaniumtown:
I'm excited for this PR to be accepted!

tigerw:
Titaniumtown, ever feature-focused :D

Titaniumtown:
:D

tigerw:
Looks good!

Below are some considerations for next steps, I'm fine with merging as-is and iterating in subsequent PRs.

--------

1. The current approach to discovering new recipes isn't fully flexible. Vanilla supports conditional triggers (inside water, has any item, has crafting result etc.) but the current implementation only permits discovery once _all_ ingredients for a crafting recipe are known to the player. I don't think this is Vanilla behaviour.
* We should make a type that represents the various triggers a player might generate. E.g. cItemTrigger(ClickedItem), cInWaterTrigger().
2. cRecipeMapper handles the details of protocol 1.12 by doing the id mapping, however, such details are best done in cProtocol_1_12. This is especially true since 1.13 now uses strings for recipe names which would obsolete cRecipeMapper.
* This class should handle the conversion between a trigger and a recipe. The conversion from a recipe to a protocol identifier should be pushed to the protocol handlers.
3. The 1.12 recipe datafile is short enough to be a switch statement in cProtocol_1_12.
* A protocol file isn't much use to the user since the ids can't be changed. We should replace this datafile with a recipe unlock file encoding the various triggers for each recipe listed in the Minecraft wiki.
4. The changes introduce a 'Cuberite recipe ID', which is just the index of the recipe inside the recipes array. Since this is dependent on the ordering of lines in crafting.txt, it's not really a good ID.
* Consider replacing this with a pointer to a recipe object.

tigerw:
One final thing, alignment by `=` inside crafting.txt appears to be broken for me, due to the `acacia_planks:` additions; the alignment is minor but why does the file now have two identifiers on the LHS? It seems redundant

    acacia_planks: AcaciaPlanks, 4  = AcaciaLog, *
    birch_planks: BirchPlanks, 4   = BirchLog, *
    ^^^^^^^^^^^^^

TooAngel:
> the alignment is minor but why does the file now have two identifiers on the LHS

The original LHS is the result item, the added ones are the minecraft recipe ids. For certain result items e.g. `RedDye` there are multiple recipes `red_dye_from_beetroot` and `red_dye_from_rose_bush`. The newly introduced name allows to be that specific.

If I understand your comment regarding 1.13 correct.  The newly introduced names in the `crafting.txt` are the ones used from 1.13 on.

peterbell10:
>2. `cRecipeMapper` handles the details of protocol 1.12 by doing the id mapping, however, such details are best done in `cProtocol_1_12`. This is especially true since 1.13 now uses strings for recipe names which would obsolete `cRecipeMapper`.

`cProtocol_1_12` is already the only place where this ID mapping is called, so your suggestion is functionally identical.

>The changes introduce a 'Cuberite recipe ID', which is just the index of the recipe inside the recipes array. Since this is dependent on the ordering of lines in crafting.txt, it's not really a good ID.

What makes you say that? It is a unique identifier, that is all an ID is.

tigerw:
Not quite, I'm suggesting that it's not a good idea to be making any assumptions about protocol constants by hardcoding them anywhere except for cProtocolXX. What if Mojang decides to change a name? They've done it before, confer `monster_egg` to `infested_block`. Protocol strings in crafting.txt isn't the right place for them or futureproof.

As an alternative, I think cRecipeMapper should be responsible for doing the _logical_ mapping from a player trigger to a unlocked recipe. The protocols themselves then convert the recipe to a protocol constant.

Well sure it's an ID, but a pointer is also an unique identifier and more suitable here xD

tigerw:
Again, these are for the future, we very much need these changes merged to improve player experience right now.

tigerw:
I think the protocol mapping file and cRecipeMapper's role is fine. After all the recipes are user-configurable; it would be brittle for the mapping to be defined in code.

If there are no further comments I'll push a change to move the underscore_keys out of crafting.txt and merge.

tigerw:
Unforeseen difficulty with that idea, there's a many-to-one mapping from ingredients to result item (duh) so Vanilla disambiguates with e.g. wool_from_string/wool_from_XX. However the result for the recipe is just 'wool' so using this as the key for recipes doesn't work.

We'll fix this when crafting.txt gets replaced with automagically generated recipes from Vanilla, for now merging as-is; please stand by.

Titaniumtown:
I'm happy this is going to be merged!

PR #4496, opened 2020-03-08
Disable ticking if player leaves the server

Seadragon91:
This fixes an assertion error.
I guess something like that happens:
1) Chunk is removed
2) Destructor of player class is called
3) Destructor of entity class is called, that checks if the entity is still being ticked
4) m_IsTicking is true and server crashes
https://github.com/cuberite/cuberite/blob/5a2163d7e6ffa0c5c568be41257fb726d452b3b9/src/Entities/Entity.cpp#L86

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -213,6 +213,7 @@ cPlayer::~cPlayer(void)
 	SaveToDisk();
 
 	m_ClientHandle = nullptr;
+	SetIsTicking(false);
 
 	delete m_InventoryWindow;
 	m_InventoryWindow = nullptr;


src/Entities/Player.cpp
```diff
@@ -213,6 +213,7 @@ cPlayer::~cPlayer(void)
 	SaveToDisk();
 
 	m_ClientHandle = nullptr;
+	SetIsTicking(false);
```
peterbell10: This doesn't look right, you're essentially just disabling the assert here.

I think it should probably be called as soon as the the disconnect packet is received. Or, possibly when the player is actually removed from the world.

Seadragon91: Okay, one of this methods?
* cWorld::RemovePlayer
* cClientHandle::SendDisconnect

peterbell10: Looking in `cClientHandle`, it seems to already do `SetIsTicking(false)`:
https://github.com/cuberite/cuberite/blob/5a2163d7e6ffa0c5c568be41257fb726d452b3b9/src/ClientHandle.cpp#L185
That  would suggest the player entity is being destroyed before `cClientHandle::Destroy` is called.

Seadragon91: Tested both methods, didn't work :(

Seadragon91: Okay closing this and will open an issue, can reproduce it now easily.

PR #4498, opened 2020-03-09
New hook and modified one

MeMuXin:
Hello, i've modified the hook OnPlayerMoving so now plugins can know wheter the player was on a solid block before moving (to know if he has jumped, etc)

And added new hook OnPlayerCrouching to know when a player crouches.

I actually added those stuff for an Elevator plugin lolz.

Tested on  
Debian GNU/Linux 10 (buster)
Kernel: Linux 4.19.0-6-amd64
Architecture: x86-64
Intel Atom N2600 quadcore @ 1.6GHz (netbook)
2GB RAM


src/Entities/Player.cpp
```diff
@@ -2573,8 +2574,8 @@ void cPlayer::UpdateMovementStats(const Vector3d & a_DeltaPos, bool a_PreviousIs
 			// If a jump just started, process food exhaustion:
 			if ((a_DeltaPos.y > 0.0) && a_PreviousIsOnGround)
 			{
-				m_Stats.AddValue(statJumps, 1);
-				AddFoodExhaustion((IsSprinting() ? 0.008 : 0.002) * static_cast<double>(Value));
+					m_Stats.AddValue(statJumps, 1);
```
bearbin: This is an erroneous change.

src/ClientHandle.cpp
```diff
@@ -849,12 +849,11 @@ void cClientHandle::HandlePlayerPos(double a_PosX, double a_PosY, double a_PosZ,
 		return;
 	}
 
-	if (cRoot::Get()->GetPluginManager()->CallHookPlayerMoving(*m_Player, OldPosition, NewPosition))
+	if (cRoot::Get()->GetPluginManager()->CallHookPlayerMoving(*m_Player, OldPosition, NewPosition, PreviousIsOnGround))
 	{
 		SendPlayerMoveLook();
 		return;
 	}
-
```
bearbin: Why was this line removed?

src/Entities/Player.cpp
```diff
@@ -867,14 +866,16 @@ void cPlayer::SetFlyingMaxSpeed(double a_Speed)
 void cPlayer::SetCrouch(bool a_IsCrouched)
 {
 	// Set the crouch status, broadcast to all visible players
-
-	if (a_IsCrouched == m_IsCrouched)
+	if (!cRoot::Get()->GetPluginManager()->CallHookPlayerCrouching(*this, a_IsCrouched))
```
bearbin: If this returns false, should we tell the player that their crouch has been rejected and re-set their pose to normal? This needs testing.

bearbin:
Also, you should add yourself to the CONTRIBUTORS file to indicate that you've licensed the contributed code according to the license.

src/Entities/Player.cpp
MeMuXin: forgot to restore the tabs, first i've added a PlayerIsJumping hook there, later on i've removed, but forgot the tabulation

src/ClientHandle.cpp
MeMuXin: it was a cleanup mistake, when deleting some function i ended up not using, my bad

src/Entities/Player.cpp
MeMuXin: ah, indeed! i'll start working asap

MeMuXin:
Tonight I'll be working on the crouching hook, i've made a modification that may (or probably may not) work, but need to get home before i can say for sure.

madmaxoft:
It seems you're actually editing the code on GitHub. You should really get the code to compile on your own machine in order to test the changes before committing them. You should also set up the pre-commit hook so that you get the style-checking: https://forum.cuberite.org/thread-2076.html

bearbin:
@MeMuXin Are you still working on this PR? I don't want it to get abandoned.

MeMuXin:
I'm working on it, but can't get the hook to nullify the user movement, I'm thinking that it's client sided.... Whenever I get an advancement I'll upload it

bearbin:
If it turns out that crouching is entirely client-sided and you can't uncrouch from the server side, then you'll just have to make the hook non-cancellable and turn it into `HOOK_PLAYER_CROUCHED` instead. But hopefully you can find a solution.

MeMuXin:
Well, actually for my personal uses it works just fine... I can modify the source and documentation to make it non cancelable, and later on come back and see if there is a way to make it cancelable

MeMuXin:
I've modified the source, updated due documentation and passed the local tests... i think thats all :S

Server/Plugins/APIDump/Hooks/OnPlayerCrouched.lua
```diff
@@ -0,0 +1,18 @@
...
+		CalledWhen = "Player crouched in the tick being currently processed.",
+		DefaultFnName = "OnPlayerCrouched",  -- also used as pagename
+		Desc = [[
+			This function is called in each server tick for each {{cPlayer|player}} that has crouched.
+		]],
+		Params =
+		{
+			{ Name = "Player", Type = "{{cPlayer}}", Notes = "The player who has crouched." },
+		},
+		Returns = [[
```
bearbin: The broadcast of positions is not dependent on the return value of the callbacks any more.

MeMuXin: My head is about to collapse, just like the Internet lolz. Mixed up too many files. Sorry. When I get the Travis CI Pass I'll correct it o7

bearbin:
Great, I was just about to merge manually but now you fixed the misplaced file.

mathiascode:
Seems to be reverting a few changes merged into master in the last days. Just make sure they don't get in.

bearbin:
@mathiascode You're right, I'll have to do some manual fixup then after all.

MeMuXin:
Sorry,  I've messed up a few things :( hope I can do better next time

PR #4500, opened 2020-03-12
Move monster speed to monsters.ini

### LINKED ISSUE #2043: Spiders are too slow
In my survival play I've encountered a few minor issues with the AI.

Spiders move much too slowly, and attack slowly as well.
Guardians move on land.
Cows like to stay in the water much more than they should.

anguslmm:
Monster walk and run speeds are now set in `monsters.ini`.

For now, only wolf and spider speeds are set there. Wolf was set to the value that was previously hardcoded, and spider was set simply as a rough estimate of the speed boost they get when chasing players.

If anyone has a good source to use as a reference to what the speeds should be for other monsters, I'm happy to add to this PR setting them all. I was unable to find a reference for that.

Also, I noticed that `EntityEffects` modify the walk speeds directly, so I added run speeds to that logic. If there's any interest, I can open an issue and get started on refactoring EntityEffects to be attached to an entity and computed as necessary (but only when they change). That might simplify effects logic in the future.

Fixes #2043


madmaxoft:
Looks good so far.
Perhaps add a documentation to all the `RelativeXyzSpeed` functions and variables as to what it is relative to - `relative to the Xyz speed configured for the mob in in monsters.ini`. It would be worth documenting *somewhere* that the monsters.ini speeds are in blocks per second, and that the Relative coefficients are applied to it dynamically.

src/MonsterConfig.cpp
```diff
@@ -92,13 +96,15 @@ void cMonsterConfig::AssignAttributes(cMonster * a_Monster, const AString & a_Na
...
-			a_Monster->SetMaxHealth      (static_cast<float>(itr->m_MaxHealth));
-			a_Monster->SetIsFireproof    (itr->m_IsFireproof);
-			a_Monster->SetBurnsInDaylight(itr->m_BurnsInDaylight);
+			a_Monster->SetAttackDamage      (itr->m_AttackDamage);
+			a_Monster->SetAttackRange       (itr->m_AttackRange);
+			a_Monster->SetSightDistance     (itr->m_SightDistance);
+			a_Monster->SetAttackRate        (static_cast<float>(itr->m_AttackRate));
+			a_Monster->SetMaxHealth         (static_cast<float>(itr->m_MaxHealth));
+			a_Monster->SetRelativeWalkSpeed (static_cast<float>(itr->m_RelativeWalkSpeed));
+			a_Monster->SetRelativeRunSpeed  (static_cast<float>(itr->m_RelativeRunSpeed));
```
peterbell10: These functions take `double`, not `float`:
```suggestion
			a_Monster->SetRelativeWalkSpeed (itr->m_RelativeWalkSpeed);
			a_Monster->SetRelativeRunSpeed  (itr->m_RelativeRunSpeed);
```

anguslmm:
> Looks good so far.
> Perhaps add a documentation to all the `RelativeXyzSpeed` functions and variables as to what it is relative to - `relative to the Xyz speed configured for the mob in in monsters.ini`. It would be worth documenting _somewhere_ that the monsters.ini speeds are in blocks per second, and that the Relative coefficients are applied to it dynamically.

@madmaxoft, do you mean in `APIDump.lua`? Or on the `m_RelativeWalkSpeed` and `m_RelativeRunSpeed` variables? Or on the getters and setters for those variables? As it stands right now, speed of a monster is just equal to those values. So coefficients aren't currently applied to them, and if you want to change the speed of a mob outside of `monsters.ini`, you just set it to the new value. Unless I'm missing something?

madmaxoft:
Ideally, I imagine the equation to be `ActualSpeed = IniSpeed * RelativeSpeed * EffectSpeedCoeff`.
The `IniSpeed` is configurable by the server admin on a per-monstertype basis, is in blocks-per-second.
The `RelativeSpeed` is something that the plugins could manipulate on a per-monster basis ("make this particular zombie extremely fast"), defaults to `1`.
The `EffectSpeedCoeff` is calculated from the various slowness and speed effects, `1` by default.

Anyway, whenever there's "relative" in the name, we need to know what it is relative to. So that needs to be documented, ideally everywhere (variables, functions, APIDump).

anguslmm:
Ah, gotcha. That makes the name make more sense. In that case, I'll add a new variable to `Monster` to store the `monsters.ini` speed, and then calculate the actual speed by multiplying by the relative speed.

Also sounds like having two variables for relative speed unnecessarily complicates things. Maybe just use `m_RelativeWalkSpeed` as the multiplier for both?

anguslmm:
@madmaxoft Does this look more like what you were thinking? The name `m_RelativeWalkSpeed` is a little misleading, considering it modifies both walk and run speeds, but I figured it didn't make sense to keep the modifier separate, as it could lead to weird behavior if a plugin dev set one and not the other.

madmaxoft:
I actually liked the possibility to have different walk and run speed multipliers. Anyway, if you want to merge them into a single one, then don't call it `Walk`, that is misleading.

Also rather than `m_DefaultSpeed` speed I think `m_BaseSpeed` is a better name.  My feeling is that "default" is something that is overridden (by any other number; thus the original value is lost), while "base" is something manipulated (such as multiplied; the original value still affects the result).

bearbin:
This PR is getting stale, we should merge or fix some more definite changes that are required.

tigerw:
We could merge and rename Default to Base

bearbin (approved):

PR #4502, opened 2020-03-16
Remove leading slash from command block commands

mathiascode:
Vanilla allows you to include the slash.

### DIFF
--- a/src/BlockEntities/CommandBlockEntity.cpp
+++ b/src/BlockEntities/CommandBlockEntity.cpp
@@ -163,6 +163,11 @@ void cCommandBlockEntity::Execute()
 		return;
 	}
 
+	if (m_Command.empty())
+	{
+		return;
+	}
+
 	class CommandBlockOutCb :
 		public cCommandOutputCallback
 	{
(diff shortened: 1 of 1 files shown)


madmaxoft:
The code says "initial", rather than "trailing".

mathiascode:
Looks like my brain shut down. I'll correct the mistake. :P

src/BlockEntities/CommandBlockEntity.cpp
```diff
@@ -178,6 +178,12 @@ void cCommandBlockEntity::Execute()
 		}
 	} CmdBlockOutCb(this);
 
+	// Remove leading slash if one exists, since console commands don't use them
+	if (m_Command[0] == '/')
```
madmaxoft: Can this codepath be taken even when `m_Command` is empty? Then this would fail.

madmaxoft:
For some reason I can't put this into line comment. Your check won't work, acessing the zero-th element of an empty string will fail.

Instead: `if (!m_Command.empty() && (m_Command[0] == '/'))`
Or even better, add a check to the top of the entire function, if the command is empty, just bail out completely.

mathiascode:
This isn't a good solution. It doesn't handle double slash commands properly.

madmaxoft:
Does vanilla allow (and discard) multiple slashes in front of commands?

Pokechu22:
No, vanilla will only strip at most one slash from a command, and that will cause issues with worldedit commands in command blocks (I don't remember how that's worked around; it might be by prefixing it with a namespace i.e. `worldedit:/set` or `/worldedit:/set`).

src/BlockEntities/CommandBlockEntity.cpp
```diff
@@ -175,21 +180,30 @@ void cCommandBlockEntity::Execute()
 		{
 			// Overwrite field
 			m_CmdBlock->SetLastOutput(cClientHandle::FormatChatPrefix(m_CmdBlock->GetWorld()->ShouldUseChatPrefixes(), "SUCCESS", cChatColor::Green, cChatColor::White) + a_Text);
+			m_CmdBlock->GetWorld()->BroadcastBlockEntity(m_CmdBlock->GetPos());
 		}
 	} CmdBlockOutCb(this);
 
+	AString RealCommand = m_Command;
+
+	// Remove trailing slash if it exists, since console commands don't use them
```
bearbin: The first character can never be a _trailing_ slash, it may only be _leading_.

mathiascode: Oops, looks like I reverted my comment to a previous version. Will fix.

PR #4504, opened 2020-03-17
Improvements to knockback

### LINKED ISSUE #4062: Velocity dead-stop

mathiascode:
- Small tweaks to knockback values to make knockback feel closer to vanilla.
- Added slight height to knockback, like in vanilla.
- Monsters now apply knockback to target.
- Fixes #4062. I haven't encountered the issue the original workaround tried to solve (wrong knockback direction) or any other issues with knockback, when using setspeed instead of addspeed (are there any side effects?)
- Closes #4300

I don't play PVP a lot, so the values and calculations can probably be tweaked further if necessary. Mob knockback still feel too "floaty" and slow compared to players, not sure why.

### DIFF
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -116,7 +116,7 @@ void cArrowEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 	}
 
 	unsigned int PunchLevel = m_CreatorData.m_Enchantments.GetLevel(cEnchantments::enchPunch);
-	double KnockbackAmount = 11 + 10 * PunchLevel;
+	double KnockbackAmount = 10 + 8 * PunchLevel;
 	a_EntityHit.TakeDamage(dtRangedAttack, GetCreatorUniqueID(), Damage, KnockbackAmount);
 
 	if (IsOnFire() && !a_EntityHit.IsInWater())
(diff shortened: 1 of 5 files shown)


src/Entities/ArrowEntity.cpp
```diff
@@ -116,7 +116,7 @@ void cArrowEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 	}
 
 	unsigned int PunchLevel = m_CreatorData.m_Enchantments.GetLevel(cEnchantments::enchPunch);
-	double KnockbackAmount = 11 + 10 * PunchLevel;
+	double KnockbackAmount = 10 + 8 * PunchLevel;
```
TooAngel: I think it could make sense to gives these values names. It's hard to guess what these numbers mean.
`MinimumKnockbackAmount = 10;
PunchLevelMultiplier = 8` would make it IMHO clearer.

peterbell10:
Approved, modulo `int`-`unsigned` issues.

PR #4505, opened 2020-03-17
Blocks only drop pickups when using correct tool

### LINKED ISSUE #4475: Players get item drops without using pickaxe
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: https://github.com/cuberite/cuberite/commit/ddc92ec0345f73f9e834fffc346a5bdb6f3fa3bc

### Expected behavior
Player must use a pickaxe to receive drops from blocks like stone and ores.

### Actual behavior
Item drops are created when a player breaks a block using their hand.

Mine a stone block without holding a pickaxe.

Regression introduced in https://github.com/cuberite/cuberite/pull/4417
Seems like CanHarvestBlock is not called anymore?

maxluchterhand1:
Fixes #4475

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,7 @@ ReleaseProfile/
 *.dir/
 CPackConfig.cmake
 CPackSourceConfig.cmake
+cmake-build-debug
 
 # APIDump-generated status files:
 Server/cuberite_api.lua
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ cloc.xsl
 *.sqlite
 /EveryNight.cmd
 /UploadLuaAPI.cmd
+GPUCache
 AllFiles.lst
 
 # IDE Stuff
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -973,10 +973,18 @@ cItems cChunk::PickupsFromBlock(Vector3i a_RelPos, const cEntity * a_Digger, con
 	NIBBLETYPE blockMeta;
 	GetBlockTypeMeta(a_RelPos, blockType, blockMeta);
 	auto blockHandler = cBlockInfo::GetHandler(blockType);
-	auto toolHandler = a_Tool->GetHandler();
-	auto canHarvestBlock = toolHandler->CanHarvestBlock(blockType);
 	auto blockEntity = GetBlockEntityRel(a_RelPos);
 	cItems pickups (0);
+	bool canHarvestBlock;
+	if (a_Tool)
+	{
+		auto toolHandler = a_Tool->GetHandler();
+		canHarvestBlock = toolHandler->CanHarvestBlock(blockType);
+	}
+	else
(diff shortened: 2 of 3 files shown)


src/World.cpp
```diff
@@ -2201,7 +2201,8 @@ bool cWorld::DigBlock(Vector3i a_BlockPos)
 bool cWorld::DropBlockAsPickups(Vector3i a_BlockPos, const cEntity * a_Digger, const cItem * a_Tool)
 {
 	auto pickups = PickupsFromBlock(a_BlockPos, a_Digger, a_Tool);
-	if (!DigBlock(a_BlockPos))
+	bool CanHarvestBlock {a_Tool->GetHandler()->CanHarvestBlock(GetBlock(a_BlockPos))};
```
peterbell10: Crash if `a_Tool` is `nullptr`.

src/World.cpp
```diff
@@ -2201,7 +2201,8 @@ bool cWorld::DigBlock(Vector3i a_BlockPos)
...
+	if (!CanHarvestBlock || !DigBlock(a_BlockPos))
 	{
 		return false;
```
peterbell10: Docs say this function only returns `false` if the chunk is not loaded.

src/World.cpp
maxluchterhand1: Fixed

src/World.cpp
maxluchterhand1: Check now happens in cChunk::PickupsFromBlock, returning empty list if tool is not valid

src/Chunk.cpp
```diff
@@ -974,7 +974,21 @@ cItems cChunk::PickupsFromBlock(Vector3i a_RelPos, const cEntity * a_Digger, con
...
+	cItems pickups (0);
+	bool canHarvestBlock;
+	if (a_Tool)
+	{
+		auto toolHandler = a_Tool->GetHandler();
+		canHarvestBlock = toolHandler->CanHarvestBlock(blockType);
+	}
+	else
+	{
+		canHarvestBlock = false;
```
peterbell10: Some blocks can be harvested with an empty hand. The empty handler should cover this:
```cpp
auto toolHandler = a_Tool ? a_Tool->GetHandler() : cItemHandler::GetItemHandler(E_ITEM_EMPTY);
```

maxluchterhand1: Accounted for that

.gitignore
```diff
@@ -104,6 +105,7 @@ ReleaseProfile/
 *.dir/
 CPackConfig.cmake
 CPackSourceConfig.cmake
+cmake-build-debug
```
peterbell10: For future reference, you can use `.git/info/exclude` to add to have git ignore files that are only relevant to you.

maxluchterhand1: Alright, thanks for the tip

PR #4506, opened 2020-03-18
Fix armor protection

### LINKED ISSUE #4098: Armor
Client version:   1.12.2
Server OS:        Linux

### Expected behavior
When wearing armor expecting armor gauge

### Actual behavior
When wearing armor there is no armor gauge

When an armor is worn

mathiascode:
Precision was lost when returning GetArmorCoverAgainst's value as int. Since the Core plugin handles world difficulty, it reverted all FinalDamage changes from the armor checks. Armor calculation has been moved to Core. https://github.com/cuberite/Core/pull/218

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -283,7 +283,7 @@ void cEntity::TakeDamage(cEntity & a_Attacker)
 void cEntity::TakeDamage(eDamageType a_DamageType, cEntity * a_Attacker, int a_RawDamage, double a_KnockbackAmount)
 {
 	float FinalDamage = a_RawDamage;
-	double ArmorCover = GetArmorCoverAgainst(a_Attacker, a_DamageType, a_RawDamage);
+	float ArmorCover = GetArmorCoverAgainst(a_Attacker, a_DamageType, a_RawDamage);
 
 	ApplyArmorDamage(static_cast<int>(ArmorCover));
 
@@ -651,7 +651,7 @@ bool cEntity::ArmorCoversAgainst(eDamageType a_DamageType)
 
 
 
-double cEntity::GetEnchantmentCoverAgainst(const cEntity * a_Attacker, eDamageType a_DamageType, int a_Damage)
+float cEntity::GetEnchantmentCoverAgainst(const cEntity * a_Attacker, eDamageType a_DamageType, int a_Damage)
 {
 	int TotalEPF = 0;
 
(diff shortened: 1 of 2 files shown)


src/Entities/Entity.cpp
```diff
@@ -282,17 +282,12 @@ void cEntity::TakeDamage(cEntity & a_Attacker)
...
-	int EnchantmentCover = GetEnchantmentCoverAgainst(a_Attacker, a_DamageType, a_RawDamage);
-	int FinalDamage = a_RawDamage - ArmorCover - EnchantmentCover;
-	if ((FinalDamage == 0) && (a_RawDamage > 0))
-	{
-		// Nobody's invincible
-		FinalDamage = 1;
-	}
-	ApplyArmorDamage(ArmorCover);
+	float FinalDamage = a_RawDamage;
+	double ArmorCover = GetArmorCoverAgainst(a_Attacker, a_DamageType, a_RawDamage);
```
bearbin: Is there any reason for mixing floats and doubles here?

src/Entities/Entity.cpp
```diff
@@ -282,17 +282,12 @@ void cEntity::TakeDamage(cEntity & a_Attacker)
...
+
+	ApplyArmorDamage(static_cast<int>(ArmorCover));
 
-	cEntity::TakeDamage(a_DamageType, a_Attacker, a_RawDamage, static_cast<float>(FinalDamage), a_KnockbackAmount);
+	cEntity::TakeDamage(a_DamageType, a_Attacker, a_RawDamage, FinalDamage, a_KnockbackAmount);
```
bearbin: FinalDamage hasn't actually been changed by this function - where are the enchantment and armour damage reductions now applied if not here?

bearbin:
I'm also rather confused as to how this actually fixes #4098 - as far as I can tell the damage calculation logic is exactly the same and cast to int before damage is applied.

mathiascode:
The code that still exists here only applies damage to the armor items. An integer is still used there, but I didn't look into how it works very extensively. The damage that affects the player's health is handled by Core now, and uses doubles.

src/Entities/Entity.cpp
mathiascode: I suppose I should use float instead of double, since Minecraft uses float for health values.

mathiascode:
Actually, now that I'm testing some mobs with larger attack damage, I think the calculations might be wrong. Will test further.

mathiascode:
False alarm. Calculations are correct, but armor protection didn't apply for zombie attacks due to faulty logic in my Core plugin changes. I will fix it in my Core PR.

mathiascode:
How so? The player is protected, and the armor wears out. What other functionality is missing?

Greaper88:
I've had at least 3 complaints that armor doesn't work properly. Like iron/diamond gear has no more protection than leather if that much. I haven't been given any specific comparisons however.

mathiascode:
Any armor protection damage changes were overwritten by the Core plugin before. It should work fine after this and https://github.com/cuberite/Core/pull/218 is merged.

peterbell10:
#4098 is talking about the UI element above the hotbar that shows the armour cover. Presumably there is some protocol message that needs to be sent to the client.

mathiascode:
Ah, I've completely forgotten about that thing. Should probably be a trivial fix.

PR #4511, opened 2020-03-19
Improve explosion visuals

mathiascode:
- Louder, more vanilla-like explosion sounds.
- Explosion visuals closer to vanilla, but with a slight twist.

Preview: https://streamable.com/6p4sz

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -1408,7 +1408,7 @@ void cWorld::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_Blo
 	cVector3iArray BlocksAffected;
 	m_ChunkMap->DoExplosionAt(a_ExplosionSize, a_BlockX, a_BlockY, a_BlockZ, BlocksAffected);
 
-	float SoundPitchMultiplier = 1.0f + (Random.RandReal(1.0f) - Random.RandReal(1.0f)) * 0.2f;
+	auto SoundPitchMultiplier = 1.0f + (Random.RandReal(1.0f) - Random.RandReal(1.0f)) * 0.2f;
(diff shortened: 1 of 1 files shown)


src/World.cpp
```diff
@@ -1423,6 +1425,17 @@ void cWorld::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_Blo
 		}
 	}
 
+	float ParticleFormula = static_cast<float>(a_ExplosionSize) * 0.33f;
+	float Spread = ParticleFormula * 0.5f;
+	float ParticleCount = std::min((ParticleFormula * 125), 600.0f);
+
+	BroadcastParticleEffect("largesmoke", Vector3f(a_BlockX, a_BlockY - 0.5f, a_BlockZ), Vector3f{}, Spread, ParticleCount);
```
madmaxoft: Pull the `Vector3f(a_BlockX, a_BlockY - 0.5f, a_BlockZ)` out into a temporary variable, use it here and on line 1437. You'll need to static_cast the coords to `float`, or maybe make it a `Vector3d`, it might go through a silent conversion.

src/World.cpp
```diff
@@ -1423,6 +1425,17 @@ void cWorld::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_Blo
 		}
 	}
 
+	float ParticleFormula = static_cast<float>(a_ExplosionSize) * 0.33f;
+	float Spread = ParticleFormula * 0.5f;
+	float ParticleCount = std::min((ParticleFormula * 125), 600.0f);
```
madmaxoft: While you're at it, use `auto` as much as possible.

PR #4512, opened 2020-03-19
Wrong overload of function push() got called when pushing a cEntity*.

### LINKED ISSUE #4499: HOOK_BLOCK_TO_PICKUPS doesn't provide the Digger parameter
Client version: 1.12.2
Server OS: Linux (5.6.0-2-MANJARO)
Cuberite Commit id: 5a2163d7e6ffa0c5c568be41257fb726d452b3b9

### Expected behavior
Item is delivered to inventory when block is broken by player while using this callback:
```
function GiveItems(World, Digger, BlockX, BlockY, BlockZ, BlockType, BlockMeta, Pickups)
	if Digger:IsPlayer() then
		Digger:GetInventory():AddItems(Pickups, true)
		return true
	end
end
```
Worked in older version (commit: 180a43d09721a32ad502f5e583d3b2d797501c00)
### Actual behavior
Item drops down and doesn't come to inventory showing error in console:
```
LUA: Plugins/HNPrison/main.lua:30: attempt to index local 'Digger' (a boolean value)
Stack trace:
 Plugins/HNPrison/main.lua(30): (no name)
Stack trace end
Error in <attached> calling function <callback>()
```

1. Create Cuberite server (5a2163d7e6ffa0c5c568be41257fb726d452b3b9)
2. Add plugin with HOOK_BLOCK_TO_PICKUPS and this callback
```
function GiveItems(World, Digger, BlockX, BlockY, BlockZ, BlockType, BlockMeta, Pickups)
	if Digger:IsPlayer() then
		Digger:GetInventory():AddItems(Pickups, true)
		return true
	end
end
```
3. Break block in Survival mode
4. Boom! Doesn't work.
### Server log
https://gist.github.com/nikkedoy98/3c95eaf607f57b72ec7094915c3b6669

Relevant part of server log:

```
LUA: Plugins/HNPrison/main.lua:30: attempt to index local 'Digger' (a boolean value)
Stack trace:
 Plugins/HNPrison/main.lua(30): (no name)
Stack trace end
Error in <attached> calling function <callback>()
```

maxluchterhand1:
Fixes #4499

When push(cEntity * a_Entity) was called, but a_Entity was _const_ cEntity *,  the overload push(boolean xyz) was called, resulting in errors.

### DIFF
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -951,6 +951,16 @@ void cLuaState::Push(bool a_Value)
 
 void cLuaState::Push(const cEntity * a_Entity)
 {
+	// Once we can make Lua understand constness, this function shall receive a corresponding function body
+	Push(const_cast<cEntity * >(a_Entity));
+}
+
+
+
+
+
+void cLuaState::Push(cEntity * a_Entity)
+{
 	ASSERT(IsValid());
 
 	if (a_Entity == nullptr)
@@ -995,7 +1005,7 @@ void cLuaState::Push(const cEntity * a_Entity)
 				}  // switch (EntityType)
 				UNREACHABLE("Unsupported entity type");
 			}();
-		tolua_pushusertype(m_LuaState, const_cast<void *>(reinterpret_cast<const void *>(a_Entity)), ClassName);
+		tolua_pushusertype(m_LuaState, a_Entity, ClassName);
 	}
 }
 
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -995,7 +995,7 @@ void cLuaState::Push(const cEntity * a_Entity)
 				}  // switch (EntityType)
 				UNREACHABLE("Unsupported entity type");
 			}();
-		tolua_pushusertype(m_LuaState, const_cast<void*>(reinterpret_cast<const void*>(a_Entity)), ClassName);
+		tolua_pushusertype(m_LuaState, const_cast<void *>(reinterpret_cast<const void *>(a_Entity)), ClassName);
 	}
 }
 
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -995,7 +995,7 @@ void cLuaState::Push(const cEntity * a_Entity)
 				}  // switch (EntityType)
 				UNREACHABLE("Unsupported entity type");
 			}();
-		tolua_pushusertype(m_LuaState, (void *) a_Entity, ClassName);
+		tolua_pushusertype(m_LuaState, const_cast<void*>(reinterpret_cast<const void*>(a_Entity)), ClassName);
 	}
 }
 
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -949,7 +949,7 @@ void cLuaState::Push(bool a_Value)
 
 
 
-void cLuaState::Push(cEntity * a_Entity)
+void cLuaState::Push(const cEntity * a_Entity)
 {
 	ASSERT(IsValid());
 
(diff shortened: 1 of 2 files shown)


src/Bindings/LuaState.cpp
```diff
@@ -995,7 +995,7 @@ void cLuaState::Push(cEntity * a_Entity)
 				}  // switch (EntityType)
 				UNREACHABLE("Unsupported entity type");
 			}();
-		tolua_pushusertype(m_LuaState, a_Entity, ClassName);
+		tolua_pushusertype(m_LuaState, const_cast<void*>(reinterpret_cast<const void*>(a_Entity)), ClassName);
```
madmaxoft: Style: Please add spaces between `void` and `*`.

peterbell10:
Instead of using `const_cast`, why not change the `Push` call to pass a non-const `cEntity *` instead? You could even do
```cpp
void Push(const cEntity *) = delete;
```
to ensure it doesn't happen again.

madmaxoft:
That might break some const-ified code, the casts will have to be moved there.
Since Lua has no concept of const objects, it might still be a better idea to do this, and make sure all the code properly casts the `const` away so that the compiler has the right idea about what's going on.

peterbell10:
If a value gets passed to Lua then it definitely shouldn't be `const`. Using `const_cast` here is just misleading.

peterbell10:
I guess it depends on how many calls would need to be fixed.

maxluchterhand1:
> If a value gets passed to Lua then it definitely shouldn't be `const`. Using `const_cast` here is just misleading.


`bool cPluginLua::OnBlockToPickups(
	cWorld & a_World,
	Vector3i a_BlockPos,
	BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta,
	const cBlockEntity * a_BlockEntity,
	const cEntity * a_Digger,
	const cItem * a_Tool,
	cItems & a_Pickups
)
{
	return CallSimpleHooks(
		cPluginManager::HOOK_BLOCK_TO_PICKUPS,
		&a_World,
		a_Digger,
		a_BlockPos.x, a_BlockPos.y, a_BlockPos.z,
		a_BlockType, a_BlockMeta, &a_Pickups
	);
}`

This is where the const comes from. You would either need to change the qualifier here, when it is disassembled by PushCallPop(T && a_Param, Args &&... args), or go with my solution. 

If values passed to lua shouldn't be const, why are the parameters of OnBlockToPickups const in the first place? I really don't know too much about lua.

madmaxoft:
@peterbell10 making the caller use non-`const` version is too much, there's just too much code relying on the `const`ness.

However, I don't like this solution either, the changed method signature looks like it's promising that Lua won't change the object, and that is not true.

How about this solution?
```cpp
void cLuaState::Push(const cEntity * a_Entity)  // New function
{
	// TODO: We don't know how to force const-ness in Lua, for now make it writable:
	Push(const_cast<cEntity *>(a_Entity));
}

void cLuaState::Push(cEntity * a_Entity)  // Keep the old function
```

How I hate the person who was too lazy to write `!= 0` in original C, so they decided that an integer is a boolean, too; this entire problem stems from that historic nuisance.

madmaxoft:
The more I think about it, the more I like having two `Push` functions, one for `const` and one for regular pointer. This fixes the problem and provides a well-defined small surface for the actual solution, if we ever find out how to make Lua obey `const`. In the meantime, it acts as a good-enough warning that this code is not exactly kosher.

madmaxoft:
You really should check your git config for timezone info, the event ordering is messed up.

peterbell10:
@maxluchterhand1 it's also a good idea to tell git your email address so github can associate your commits with your username.
```
git config --global user.email "email@example.com"
```

PR #4513, opened 2020-03-19
Increase speed of splash potion and expbottle

### LINKED ISSUE #1829: Items, tile entities, and bottle o' enchantings fall too slowly
When you drop or throw any of these items, it seems like you're on the moon because they don't fall as fast as they should (in vanilla Minecraft, they fall a lot faster, and in real life, too).

mathiascode:
Fixes #1829

### DIFF
--- a/src/Items/ItemPotion.h
+++ b/src/Items/ItemPotion.h
@@ -40,7 +40,10 @@ public:
 		}
 
 		Vector3d Pos = a_Player->GetThrowStartPos();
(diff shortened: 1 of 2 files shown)


bearbin:
How did you determine the speed values? Just empirically or is there a list of them somewhere.

mathiascode:
Empirically, but I'd love to have a list if other values are based on one.

PR #4514, opened 2020-03-19
When Client sends message longer than 256 bytes, kick him instead of handling message

### LINKED ISSUE #4409: Crash server with large chat message
### Expected behavior
Server should kick the client for trying to send a to large chat message or packet .

### Actual behavior
Server crashes.

Use that [python script](https://github.com/Seadragon91/pyCraft/blob/ddos_commands/run_chat_message.py)  with pycraft, don't forget to switch off `Authentication`.

maxluchterhand1:
fixes #4409

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1536,38 +1536,37 @@ void cClientHandle::HandleChat(const AString & a_Message)
 	{
 		this->Kick(std::string("Please don't exceed the maximum message length of ")
 		+ std::to_string(MAX_CHAT_MSG_LENGTH));
+		return;
+	}
+	// We no longer need to postpone message processing, because the messages already arrive in the Tick thread
+
+	// If a command, perform it:
+	AString Message(a_Message);
+	if (cRoot::Get()->GetServer()->Command(*this, Message))
+	{
+		return;
+	}
+
+	// Not a command, broadcast as a message:
+	cCompositeChat Msg;
+	AString Color = m_Player->GetColor();
+	if (Color.length() == 3)
+	{
+		Color = AString("@") + Color[2];
 	}
 	else
 	{
-		// We no longer need to postpone message processing, because the messages already arrive in the Tick thread
-
-		// If a command, perform it:
-		AString Message(a_Message);
-		if (cRoot::Get()->GetServer()->Command(*this, Message))
-		{
-			return;
-		}
-
-		// Not a command, broadcast as a message:
-		cCompositeChat Msg;
-		AString Color = m_Player->GetColor();
-		if (Color.length() == 3)
-		{
-			Color = AString("@") + Color[2];
-		}
-		else
-		{
-			Color.clear();
-		}
-		Msg.AddTextPart("<");
-		Msg.ParseText(m_Player->GetPrefix());
-		Msg.AddTextPart(m_Player->GetName(), Color);
-		Msg.ParseText(m_Player->GetSuffix());
-		Msg.AddTextPart("> ");
-		Msg.ParseText(Message);
-		Msg.UnderlineUrls();
-		cRoot::Get()->BroadcastChat(Msg);
+		Color.clear();
 	}
+	Msg.AddTextPart("<");
+	Msg.ParseText(m_Player->GetPrefix());
+	Msg.AddTextPart(m_Player->GetName(), Color);
+	Msg.ParseText(m_Player->GetSuffix());
+	Msg.AddTextPart("> ");
+	Msg.ParseText(Message);
+	Msg.UnderlineUrls();
+	cRoot::Get()->BroadcastChat(Msg);
+
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1555,7 +1555,8 @@ void cClientHandle::HandleChat(const AString & a_Message)
 	if ((a_Message.size()) > MAX_CHAT_MSG_LENGTH)
 	{
 		this->Kick(std::string("Please don't exceed the maximum message length of ")
-		+ std::to_string(MAX_CHAT_MSG_LENGTH));
+			+ std::to_string(MAX_CHAT_MSG_LENGTH)
+		);
 		return;
 	}
 	// We no longer need to postpone message processing, because the messages already arrive in the Tick thread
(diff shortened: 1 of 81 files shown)


src/Protocol/Protocol_1_9.cpp
```diff
@@ -2524,7 +2524,15 @@ void cProtocol_1_9_0::HandlePacketBoatSteer(cByteBuffer & a_ByteBuffer)
 void cProtocol_1_9_0::HandlePacketChatMessage(cByteBuffer & a_ByteBuffer)
 {
 	HANDLE_READ(a_ByteBuffer, ReadVarUTF8String, AString, Message);
-	m_Client->HandleChat(Message);
+	if ((Message.size()/8) > MAX_CHAT_MSG_LENGTH)
```
bearbin: Why is message size divided by 8 here? Surely this would result in the max length being 2048, as per your commit message.

maxluchterhand1: I just made it so beacuse Message.size() returns the size of the message in bits. So a size() call on a 256 character message returns 2048. I divide by 8 here so that the value of MAX_CHAT_MSG_LENGTH can correspond to the size in bytes/ amount of characters

bearbin: But Message is just a String, calling size() on it should return bytes rather than bits. Maybe I'm missing something here.

maxluchterhand1: No I was missing something. I was also confused as to why .size() would return bytes. Turns out, it did not. I was using a python script that wrote messages to the server as a bot, and thought I was configuring the number of characters it sends. I was actually configuring the number of times it adds an 8 byte string to the message it is going to send. Nevermind, I fixed that up and increased the MAX_STRING_SIZE to 1024, as 256 was too little.

maxluchterhand1:
Why do we need to limit then characters to 256? And when you ask about how the vanilla client limits characters, do you actually mean the game client or do you mean the vanilla server?

I am kinda confused as to where the 256 comes from. I just picked a number that I thought was appropriate, and then increased the max size to 1024 when it didn't suffice. The game actually handles much bigger messages without any issues. It was only the server crashing due to an unsufficiently sized buffer that was a problem, correct me if I am totally missing out on something here.

bearbin:
The [vanilla server limits messages to 256](https://wiki.vg/Protocol#Chat_Message_.28serverbound.29), I had assumed you were trying to replicate the vanilla behaviour with the first limit of 256. 

When I was referring to 'vanilla' I meant the server, but on the assumption that the server would reject things that break the client. If the client is fine with large messages, then this is a perfect fix.

bearbin:
Also, would you mind fixing your clock? When the commits have the wrong times, it makes the discussion rather hard to follow on the GH interface.

Pokechu22:
Vanilla limits to 256 codepoints, not 256 bytes.  The definition of `String (n)` in the [data types](https://wiki.vg/Protocol#Data_types) section is:

> UTF-8 string prefixed with its size in bytes as a VarInt. Maximum length of `n` characters, which varies by context; up to `n × 4` bytes can be used to encode `n` characters and both of those limits are checked. Maximum `n` value is 32767. The + 3 is due to the max size of a valid length VarInt.

Also, note that the outbound and inbound limits are different: the server can send a chat message with a much higher length (as it uses JSON chat, and that uses `n` = 32767), but the client is limited to plain text with `n` = 256.

maxluchterhand1:
As far as I can tell, my clock settings are correct.
Also, I am completely new to working with GitHub, and I can't really understand this request in your comment: 

"Otherwise, and after the dependency PR has been merged and this rebased on the new master, this looks good."

Could you please give me a more specific explanation as to what still needs to be done for this PR to be valid and get merged?

maxluchterhand1:
test

bearbin:
Your computer may be showing the correct time, but it's generating git commits with the wrong time - the time zone is incorrect. You git commits are being annotated with the time zone as UTC but the time itself is UTC+1. 

The reason this isn't able to be merged currently is that this branch is not self-contained and split off from the one for #4512, so it contains the commits for that branch. Once it's been merged, you need to rebase on the new master and then it'll all be good.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -2524,7 +2524,15 @@ void cProtocol_1_9_0::HandlePacketBoatSteer(cByteBuffer & a_ByteBuffer)
 void cProtocol_1_9_0::HandlePacketChatMessage(cByteBuffer & a_ByteBuffer)
 {
 	HANDLE_READ(a_ByteBuffer, ReadVarUTF8String, AString, Message);
-	m_Client->HandleChat(Message);
+	if ((Message.size()) > MAX_CHAT_MSG_LENGTH)
+	{
+		m_Client->Kick(std::string("Please don't exceed the maximum message length of ")
```
peterbell10: Shouldn't this be in `cClient::HandleChat` so that is protocol version agnostic?

maxluchterhand1: yes, it should. changing this now

src/ClientHandle.cpp
```diff
@@ -1529,34 +1532,42 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
...
-	// Not a command, broadcast as a message:
-	cCompositeChat Msg;
-	AString Color = m_Player->GetColor();
-	if (Color.length() == 3)
+	if ((a_Message.size()) > MAX_CHAT_MSG_LENGTH)
 	{
-		Color = AString("@") + Color[2];
+		this->Kick(std::string("Please don't exceed the maximum message length of ")
+		+ std::to_string(MAX_CHAT_MSG_LENGTH));
 	}
```
peterbell10: Please write this as a guard clause to avoid deep nesting of `if`-`else` blocks:
```cpp
void cClientHandle::HandleChat(const AString & a_Message)
{
    if (a_Message.size() > MAX_CHAT_MSG_LENGTH) {
        Kick(std::string("Please don't exceed the maximum message length of ")
	    	+ std::to_string(MAX_CHAT_MSG_LENGTH));
        return;
    }

    // Rest of code exactly as before
}
```

src/Protocol/Protocol_1_9.cpp
```diff
@@ -56,7 +56,6 @@ Implements the 1.9 protocol classes:
 
 
 
-
```
peterbell10: Left over after moving `MAX_CHAT_MSG_LENGTH`

PR #4517, opened 2020-03-20
Don't call player movement hook if nothing changed

mathiascode:
The client seems to send a position packet every second or so, even if the position didn't change. Does it make sense to call HOOK_PLAYER_MOVING every time?

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -842,6 +842,11 @@ void cClientHandle::HandlePlayerPos(double a_PosX, double a_PosY, double a_PosZ,
 	double OldStance = GetPlayer()->GetStance();
 	auto PreviousIsOnGround = GetPlayer()->IsOnGround();
 
+	#ifdef __clang__
+		#pragma clang diagnostic push
+		#pragma clang diagnostic ignored "-Wfloat-equal"
+	#endif
+
 	if (
 	    (OldPosition == NewPosition) &&
 	    (OldStance == a_Stance) &&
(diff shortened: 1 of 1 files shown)


mathiascode:
Any preferred way of comparing doubles? I'm not that familiar with C++.

bearbin:
There's nothing wrong with your code, you just need to suppress the warning. Since you're just comparing something to its previous value there's no chance for floating point encoding to make things work unintuitively.

src/ClientHandle.cpp
```diff
@@ -839,8 +839,15 @@ void cClientHandle::HandlePlayerPos(double a_PosX, double a_PosY, double a_PosZ,
 
 	Vector3d NewPosition(a_PosX, a_PosY, a_PosZ);
 	Vector3d OldPosition = GetPlayer()->GetPosition();
+	double OldStance = GetPlayer()->GetStance();
 	auto PreviousIsOnGround = GetPlayer()->IsOnGround();
 
+	if ((OldPosition == NewPosition) && (OldStance == a_Stance))
```
peterbell10: Why are you not checking `a_IsOnGround` as well?
```suggestion
	if (
	    (OldPosition == NewPosition) &&
	    (OldStance == a_Stance) &&
	    (PreviousIsOnGround == a_IsOnGround)
	)
```

bearbin:
Style needs fixing then merge by @peterbell10 ?

bearbin:
Hopefully travis recovers soon, it's not been very reliable recently at all.

PR #4518, opened 2020-03-20
Lua API for spectating entities

mathiascode:
Prerequisite for implementing MC's /spectate command.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11132,6 +11132,17 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Sets the player visibility to other players",
 				},
+				SpectateEntity =
+				{
+					Params =
+					{
+						{
+							Name = "Entity",
+							Type = "cEntity",
+						},
+					},
+					Notes = "Allows the player to spectate the target entity. Does not change the player's gamemode to spectator mode.",
+				},
 				TossEquippedItem =
 				{
 					Params =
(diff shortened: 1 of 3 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11132,6 +11132,17 @@ a_Player:OpenWindow(Window);
...
+				SpectateEntity =
+				{
+					Params =
+					{
+						{
+							Name = "Entity",
+							Type = "cEntity",
+						},
+					},
+					Notes = "Allows the player to spectate the target entity. Does not change the player's gamemode to spectator mode.",
```
bearbin: `Allows` -> `Sets`? Also you should mention that self-spectating resets the spectation.

src/Entities/Player.cpp
```diff
@@ -1702,6 +1702,23 @@ void cPlayer::SendRotation(double a_YawDegrees, double a_PitchDegrees)
 
 
 
+void cPlayer::SpectateEntity(cEntity * a_Target)
+{
+	if (static_cast<cPlayer *>(a_Target) == this)
```
peterbell10: Is the cast necessary? If so, I would prefer upcasting `this` (guarunteed to be valid) over downcasting `a_Target` (which might not be a `cPlayer`).

src/Entities/Player.cpp
```diff
@@ -1702,6 +1702,23 @@ void cPlayer::SendRotation(double a_YawDegrees, double a_PitchDegrees)
 
 
 
+void cPlayer::SpectateEntity(cEntity * a_Target)
```
peterbell10: This should either accept `nullptr` or `a_Target` should be a reference. As written, a plugin could crash the server by doing `Player:SpectateEntity(nil)`.

PR #4521, opened 2020-03-21
Add ambient mob sounds

### LINKED ISSUE #3443: Mobs don't make sounds when they move or do periodicals
A zombie groaning occasionally, a sheep baaa'ing.

Taken from #1109

mathiascode:
Nothing too fancy, but does the job. Partially fixes. #3443.

### DIFF
--- a/src/Mobs/Monster.cpp
+++ b/src/Mobs/Monster.cpp
@@ -118,6 +118,9 @@ cMonster::cMonster(const AString & a_ConfigName, eMonsterType a_MobType, const A
 	{
 		GetMonsterConfig(a_ConfigName);
 	}
+
+	// Prevent mobs spawning at the same time from making sounds simultaneously
+	m_AmbientSoundTimer = GetRandomProvider().RandInt(0, 100);
 }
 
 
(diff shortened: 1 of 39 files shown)


src/Mobs/Monster.cpp
```diff
@@ -384,6 +385,24 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 	BroadcastMovementUpdate();
 
+	// Ambient mob sounds
+	if (!m_SoundAmbient.empty())
+	{
+		--m_AmbientSoundTimer;
```
bearbin: You could combine the decrement into the if condition here.

src/Mobs/Monster.cpp
```diff
@@ -384,6 +385,24 @@ void cMonster::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+
+		if (m_AmbientSoundTimer <= 0)
+		{
+			auto & Random = GetRandomProvider();
+			auto ShouldPlaySound = Random.RandBool();
+			if (ShouldPlaySound)
+			{
+				auto SoundPitchMultiplier = 1.0f + (Random.RandReal(1.0f) - Random.RandReal(1.0f)) * 0.2f;
+				m_World->BroadcastSoundEffect(m_SoundAmbient, GetPosition(), 1.0f, SoundPitchMultiplier * 1.0f);
```
bearbin: A multiplication by one shouldn't do anything to SoundPitchMultiplier?

src/Mobs/Squid.cpp
```diff
@@ -9,7 +9,7 @@
 
 
 cSquid::cSquid(void) :
-	super("Squid", mtSquid, "entity.squid.hurt", "entity.squid.death", 0.95, 0.95)
+	super("Squid", mtSquid, "entity.squid.hurt", "entity.squid.death", "", 0.95, 0.95)
```
bearbin: Do squids not have an ambient sound?

src/Mobs/Monster.h
```diff
@@ -287,6 +288,8 @@ class cMonster :
 	bool m_BurnsInDaylight;
 	double m_RelativeWalkSpeed;
 
+	int m_AmbientSoundTimer = GetRandomProvider().RandInt(0, 100);
```
bearbin: This should really be in the constructor.

Greaper88:
"what does the squid say?!"

src/Mobs/Monster.cpp
mathiascode: Ah right, leftover from my explosion changes. Will fix.

src/Mobs/Squid.cpp
mathiascode: Seems like it was introduced in 1.13. I'll add the sound name.

PR #4525, opened 2020-03-21
Pickup spawn improvements

mathiascode:
Pickups are now thrown around when spawned, like in vanilla. Pickups also bail out of the collision detection checks while being created, to prevent them from blasting too far off in e.g. cobblestone generators.


mathiascode:
Don't merge yet, this changes behavior of a_FlyAwaySpeed used in a few places in the code,

mathiascode:
Should be fine now.

PR #4529, opened 2020-03-22
Fix invalid explosion damage

mathiascode:
A 1/0 division would sometimes occur, resulting in players/entities with unlimited health.


bearbin:
Travis is still broken but the code is fine. Merging.

PR #4530, opened 2020-03-22
Fix effect crashes

mathiascode:
Some effects could crash the server due to division by zero.

### DIFF
--- a/src/Entities/EntityEffect.cpp
+++ b/src/Entities/EntityEffect.cpp
@@ -373,7 +373,7 @@ void cEntityEffectRegeneration::OnTick(cPawn & a_Target)
 	}
 
 	// Regen frequency = 50 ticks, divided by potion level (Regen II = 25 ticks)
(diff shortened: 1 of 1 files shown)


src/Entities/EntityEffect.cpp
```diff
@@ -373,7 +373,7 @@ void cEntityEffectRegeneration::OnTick(cPawn & a_Target)
 	}
 
 	// Regen frequency = 50 ticks, divided by potion level (Regen II = 25 ticks)
-	int frequency = FloorC(50.0 / static_cast<double>(m_Intensity + 1));
+	int frequency = FloorC(50.0 / std::min(static_cast<double>(m_Intensity + 1), 50.0));
```
peterbell10: Nit: IMO the division by 0 is more obvious written this way:
```suggestion
	int frequency = std::max(1, FloorC(50.0 / static_cast<double>(m_Intensity + 1));
```

PR #4534, opened 2020-03-23
Furnace and brewing stand drops

mathiascode:
Contents of these block entities were not dropped before. There are still issues that I'm not sure how to solve in a good way:
- E_BLOCK_LIT_FURNACE should drop E_BLOCK_FURNACE
- E_BLOCK_BREWING_STAND should drop E_ITEM_BREWING_STAND

### DIFF
--- a/src/Blocks/BlockBrewingStand.h
+++ b/src/Blocks/BlockBrewingStand.h
@@ -1,7 +1,7 @@
 
 #pragma once
 
-#include "BlockEntity.h"
+#include "../BlockEntities/BrewingstandEntity.h"
 #include "Mixins.h"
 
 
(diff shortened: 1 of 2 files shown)


mathiascode:
Should be fine now.

src/Blocks/BlockBrewingStand.h
```diff
@@ -24,6 +24,21 @@ class cBlockBrewingStandHandler :
 
 
 
+	virtual cItems ConvertToPickups(NIBBLETYPE a_BlockMeta, cBlockEntity * a_BlockEntity, const cEntity * a_Digger, const cItem * a_Tool) override
+	{
+		cItems res(cItem(E_ITEM_BREWING_STAND, 1)); // We have to drop the item form of a brewing stand
+		if (a_BlockEntity != nullptr)
+		{
+			auto be = static_cast<cBrewingstandEntity *>(a_BlockEntity);
```
peterbell10: Should we be worried about block entity mismatch here?

mathiascode: Behavior was taken from cBlockDropSpenserHandler, and it seems to have worked fine there.

PR #4536, opened 2020-03-23
Fix certain item drops

mathiascode:
Meta shouldn't be reset for these blocks, as they have special item forms.

### DIFF
--- a/src/Blocks/BlockComparator.h
+++ b/src/Blocks/BlockComparator.h
@@ -41,33 +41,9 @@ public:
 		return true;
 	}
 
-	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, int a_RelX, int a_RelY, int a_RelZ, const cChunk & a_Chunk, NIBBLETYPE a_BlockMeta) override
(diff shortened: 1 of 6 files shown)


src/Blocks/BlockComparator.h
```diff
@@ -3,16 +3,17 @@
 
 #include "BlockHandler.h"
 #include "BlockRedstoneRepeater.h"
+#include "BlockStairs.h"
```
bearbin: What is this include needed for?

mathiascode: Leftover, will remove.

src/Blocks/BlockRedstoneRepeater.h
```diff
@@ -5,16 +5,17 @@
 #include "Mixins.h"
 #include "ChunkInterface.h"
 #include "BlockSlab.h"
+#include "BlockStairs.h"
```
bearbin: There's another one here.

bearbin:
This looks good once style and above comment is fixed.

PR #4537, opened 2020-03-23
Jukebox improvements

mathiascode:
- Clear meta on drop to prevent invalid block
- Stop playing when broken
- Drop disc when broken

### DIFF
--- a/src/BlockEntities/JukeboxEntity.cpp
+++ b/src/BlockEntities/JukeboxEntity.cpp
@@ -1,6 +1,9 @@
 
-#include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
+// JukeboxEntity.cpp
 
+// Implements the cJukeboxEntity class representing a single jukebox in the world
+
+#include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
 #include "JukeboxEntity.h"
 #include "../World.h"
 #include "../EffectID.h"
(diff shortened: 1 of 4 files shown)


src/BlockEntities/JukeboxEntity.cpp
```diff
@@ -23,6 +26,21 @@ cJukeboxEntity::cJukeboxEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Ve
 
 cJukeboxEntity::~cJukeboxEntity()
 {
+	if (m_World)
+	{
+		// Stop playing music when destroyed by any means
+		m_World->BroadcastSoundParticleEffect(EffectID::SFX_RANDOM_PLAY_MUSIC_DISC, GetPos(), 0);
```
peterbell10: Why is this needed? When is a block entity removed from the world without first calling `Destroy`?

mathiascode: E.g. WorldEdit.

src/BlockEntities/JukeboxEntity.cpp
```diff
@@ -23,6 +26,21 @@ cJukeboxEntity::cJukeboxEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Ve
 
 cJukeboxEntity::~cJukeboxEntity()
 {
+	if (m_World)
```
peterbell10:
```suggestion
	if (m_World && IsPlayingRecord())
```

PR #4538, opened 2020-03-23
Add protocol mob ID remapping

mathiascode:
Initial step for enabling mobs in MC 1.13+. After this we should look into the mob metadata, update it, and enable it for MC 1.13.


src/Protocol/Protocol_1_11.cpp
```diff
@@ -388,7 +388,7 @@ void cProtocol_1_11_0::SendSpawnMob(const cMonster & a_Mob)
 	// TODO: Bad way to write a UUID, and it's not a true UUID, but this is functional for now.
 	Pkt.WriteBEUInt64(0);
 	Pkt.WriteBEUInt64(a_Mob.GetUniqueID());
-	Pkt.WriteVarInt32(static_cast<UInt32>(a_Mob.GetMobType()));
```
bearbin: Legacy sounds like the wrong name for this function - we don't want to have 'Legacy' code in every protocol including the latest one. Maybe `GetProtocolMobID` or something instead?

bearbin:
Clang builds fail due to a missing cast.

mathiascode:
This will break any plugins that use SpawnMob with a hardcoded mob id, but I'm not aware of such plugins.

mathiascode:
Seems like mob IDs are changing in every version after 1.13. More remapping has to be done when we introduce 1.14 and 1.15 support.

src/Protocol/Protocol_1_13.cpp
```diff
@@ -368,6 +368,34 @@ void cProtocol_1_13::SendScoreboardObjective(const AString & a_Name, const AStri
...
+void cProtocol_1_13::SendSpawnMob(const cMonster & a_Mob)
+{
+	ASSERT(m_State == 3);  // In game mode?
+
+	cPacketizer Pkt(*this, pktSpawnMob);
+	Pkt.WriteVarInt32(a_Mob.GetUniqueID());
+	// TODO: Bad way to write a UUID, and it's not a true UUID, but this is functional for now.
+	Pkt.WriteBEUInt64(0);
+	Pkt.WriteBEUInt64(a_Mob.GetUniqueID());
+	Pkt.WriteVarInt32(static_cast<UInt32>(a_Mob.GetMobType()));
```
peterbell10: IMO we should have a `GetProtocolMobID` call here as well and the values of `eMonsterType` shouldn't be linked to any protocol version.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1347,7 +1347,7 @@ void cProtocol_1_8_0::SendSpawnMob(const cMonster & a_Mob)
 
 	cPacketizer Pkt(*this, pktSpawnMob);
 	Pkt.WriteVarInt32(a_Mob.GetUniqueID());
-	Pkt.WriteBEUInt8(static_cast<Byte>(a_Mob.GetMobType()));
+	Pkt.WriteBEUInt8(static_cast<Byte>(GetProtocolMobID(a_Mob.GetMobType())));
```
peterbell10: Instead of casting, `GetProtocolMobID` should return a `Byte`.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1827,6 +1827,49 @@ int cProtocol_1_8_0::GetParticleID(const AString & a_ParticleName)
 
 
 
+UInt32 cProtocol_1_8_0::GetProtocolMobID(int a_MobType)
```
peterbell10:
```suggestion
Byte cProtocol_1_8_0::GetProtocolMobID(eMonsterType a_MobType)
```

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1827,6 +1827,49 @@ int cProtocol_1_8_0::GetParticleID(const AString & a_ParticleName)
...
+		case mtSpider:                return 52;
+		case mtSquid:                 return 94;
+		case mtVillager:              return 120;
+		case mtWitch:                 return 66;
+		case mtWither:                return 64;
+		case mtWolf:                  return 95;
+		case mtZombie:                return 54;
+		case mtZombiePigman:          return 57;
+		default:                      return a_MobType;
+	}
```
peterbell10: Default is not a good idea. We want `-Wenum` errors to tell us if any cases are missing.
```suggestion
	}
	UNREACHABLE("Unsupported mob type");
```

src/Protocol/Protocol_1_13.cpp
mathiascode: I agree. What would be a good numbering scheme for eMonsterType?

bearbin: If we want the least changes to code, and maintain compatibility with old hardcoded plugins, then just keep the values the same. If we want to break compatibility, then just ordering the mobs in the order they were introduced to the game seems to make sense.

mathiascode: Keeping them the same as before is probably the best idea. There's always going to be issues when we eventually get to implementing mobs that were introduced before already implemented ones.

peterbell10: I would say either keep them the same, or don't specify values at all.

mathiascode:
Something similar should probably be done for sound effects as well. The names of some of these have also changed in newer MC versions.

mathiascode:
Should be ready after CI has finished.

src/Protocol/Protocol_1_11.cpp
```diff
@@ -388,7 +388,7 @@ void cProtocol_1_11_0::SendSpawnMob(const cMonster & a_Mob)
...
+	Pkt.WriteVarInt32(static_cast<UInt32>(GetProtocolMobID(a_Mob.GetMobType())));
```
madmaxoft: Please please please don't use `ID` for things that are not IDs. It's a Type, not an ID.
(ID = unique identifier, unique for each instance. Type = kind of an object, possibly shared by multiple objects)

src/Protocol/Protocol_1_13.cpp
```diff
@@ -395,6 +424,50 @@ void cProtocol_1_13::SendUpdateBlockEntity(cBlockEntity & a_BlockEntity)
 
 
 
+UInt32 cProtocol_1_13::GetProtocolMobID(eMonsterType a_MobType)
+{
+	switch (a_MobType)
+	{
+		case mtInvalidType:           return -1;
```
madmaxoft: You'll either need to `static_cast` to `UInt32`, or pick a different value for "bad type". Possibly even make it a `static const` in `cProtocol`.

src/Protocol/Protocol_1_11.cpp
madmaxoft: Also, you should first check if `GetProtocolMobID` returns a valid type, and only then spawn the mob (what happens to the client, if you send a packet with an invalid mob type? What happens if you don't spawn a mob but then send movement and action packets to it?)

src/Protocol/Protocol_1_13.h
```diff
@@ -68,10 +69,14 @@ class cProtocol_1_13 :
...
 	virtual void SendPluginMessage              (const AString & a_Channel, const AString & a_Message) override;
 	virtual void SendScoreboardObjective        (const AString & a_Name, const AString & a_DisplayName, Byte a_Mode) override;
+	virtual void SendSpawnMob                   (const cMonster & a_Mob) override;
 	virtual void SendStatistics                 (const cStatManager & a_Manager) override;
 	virtual void SendTabCompletionResults       (const AStringVector & a_Results) override;
 	virtual void SendUpdateBlockEntity          (cBlockEntity & a_BlockEntity) override;
 
+	/** Converts eMonsterType to protocol-specific mob IDs */
+	static UInt32 GetProtocolMobID(eMonsterType a_MobType);
+
```
madmaxoft: This should not be `static` at all, but rather `virtual` - pure virtual in `cProtocol` and `virtual override` in all descendants.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1827,6 +1828,50 @@ int cProtocol_1_8_0::GetParticleID(const AString & a_ParticleName)
 
 
 
+Byte cProtocol_1_8_0::GetProtocolMobID(eMonsterType a_MobType)
+{
+	switch (a_MobType)
+	{
+		case mtInvalidType:           return 0;
```
madmaxoft: Suddenly you return `0` and not `-1`?

src/Protocol/Protocol_1_13.h
madmaxoft: Make it return `UInt32` in all versions, then `static_cast` in protocols that use shorter datatype.

madmaxoft:
I've written to Appveyor about the increased build failures we're seeing now: https://help.appveyor.com/discussions/problems/26686-builds-randomly-failing-with-unexpected-logger-failure

src/Protocol/Protocol_1_11.cpp
mathiascode: Sigh, tell that to Mojang...

mathiascode: I think we should just remap newer mobs to some other mob in older MC versions. If not, we would get an "invisible" mob in any case, which we don't want more of.

mathiascode:
I added mob types for the 1.11 protocol too, in preparation for https://github.com/cuberite/cuberite/pull/4563.

madmaxoft:
Can all Minecraft versions still display the unused Giant mob? That could be a solution to the "unknown / invalid mob" - better to have a "giant killed me" bug report, than an "invisible mob killed me" :)

mathiascode:
Maybe I'm misunderstanding something, but in what case would we send an invalid mob type? If a mob isn't implemented in a MC version, it should be remapped to a mob that's implemented and looks the most similar to the intended mob. If a mob type isn't taken care of in the switch statements, UNREACHABLE("Unsupported mob type") should make it clear.

src/Protocol/Protocol_1_13.cpp
```diff
@@ -368,6 +369,34 @@ void cProtocol_1_13::SendScoreboardObjective(const AString & a_Name, const AStri
...
+	Pkt.WriteBEInt16(static_cast<Int16>(a_Mob.GetSpeedY() * 400));
+	Pkt.WriteBEInt16(static_cast<Int16>(a_Mob.GetSpeedZ() * 400));
+	WriteEntityMetadata(Pkt, a_Mob);
+	Pkt.WriteBEUInt8(0xff);  // Metadata terminator
+}
+
+
+
+
+
```
peterbell10: If I'm not mistaken, this is exactly the same as the 1.11 version so virtual `GetProtocolMobType` should be the only override needed.
```suggestion
```

src/Protocol/Protocol.h
```diff
@@ -236,6 +237,8 @@ class cProtocol
 	/** Returns the ServerID used for authentication through session.minecraft.net */
 	virtual AString GetAuthServerID(void) = 0;
 
+	/** Converts eMonsterType to protocol-specific mob types */
+	virtual UInt32 GetProtocolMobType(eMonsterType a_MobType) = 0;
```
peterbell10: IMO this shouldn't be public. Consider putting this next to `GetPacketID` in the protected section since they are conceptually similar.

peterbell10:
>Maybe I'm misunderstanding something, but in what case would we send an invalid mob type?

I agree that unsupported mobs should map to a visually similar mob type. However, I think @madmaxoft meant `mtInvalidType` should map to `mtGiant` instead of just 0.

PR #4539, opened 2020-03-23
Initial work on 1.13 entity metadata

mathiascode:
Some initial work on investigating entity metadata in 1.13, to spark some discussion. I'm having issues with boolean values, seems like MC 1.13 doesn't detect them as boolean, and instead gets disconnected from the server or crashes. Any idea what might be happening?

### DIFF
--- a/src/Protocol/Protocol_1_13.cpp
+++ b/src/Protocol/Protocol_1_13.cpp
@@ -22,6 +22,8 @@ Implements the 1.13 protocol classes:
 #include "../Entities/FireworkEntity.h"
 #include "../Entities/SplashPotionEntity.h"
 
+#include "../Mobs/IncludeAllMonsters.h"
+
 #include "../BlockTypePalette.h"
 #include "../ClientHandle.h"
 #include "../Root.h"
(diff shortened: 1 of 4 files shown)


src/Protocol/Protocol_1_13.cpp
```diff
@@ -460,5 +792,635 @@ void cProtocol_1_13::WriteItem(cPacketizer & a_Pkt, const cItem & a_Item)
...
+			break;
+		}  // case mtBat
+
+		case mtChicken:
+		{
+			/* "ata cannot be cast to java.lang.Boolean"? Investigate
+			auto & Chicken = static_cast<const cChicken &>(a_Mob);
+
+			a_Pkt.WriteBEUInt8(AGEABLE_BABY);
+			a_Pkt.WriteBEUInt8(METADATA_TYPE_BOOL);
```
tigerw, line 1036: They added a new metadata type so `METADATA_TYPE_BOOL` needs updating, plus this field should now be a `WriteVarInt32`. Ideally, if metadata largely hasn't changed, we need to figure out a way to not copy paste the entire function XD

https://wiki.vg/index.php?title=Entity_metadata&direction=next&oldid=14048#Entity_Metadata_Format

mathiascode: Thanks! This probably explains some other weird behavior I was experiencing.

mathiascode: I don't remember seeing any large metadata changes when I last worked on this PR. The only real problem was the boolean one, but that can be fixed easily now that we know the cause.

mathiascode:
If anyone wants to work on this, feel free to do so. It's not a current priority for me.

tigerw:
It seems to work, but wiki.vg is making my eyes swim...

Titaniumtown:
I'll to some testing :)

tigerw:
Thanks a ton!

Titaniumtown:
Can confirm. Entities work! One question though, is it a common issue, not being able to use spawn eggs? non of them worked, so i had to do /summon

Titaniumtown:
Also, update aquatic mobs dont spawn (but i know that 1.13 support is SUPER early)
And leaves dont show up (no other blocks though).
I dont think any of those issues are related to this PR, but I just wanted to report in what i see!
:)

tigerw:
Many thanks for testing and reporting! Yeah, 1.13 changed all the block and item identifiers, and we don't have a mapping for the new items yet, so the server doesn't recognise whatever the client sends when you get a spawn egg in creative.

tigerw:
Any further comments?

With regards to the duplication in protocol (metadata, status request) I have some thoughts about how to reduce it, in another PR.

PR #4540, opened 2020-03-23
Send respawn packet by default

mathiascode:
I don't see why the respawn packet is disabled by default. Not sending one breaks the client, and causes confusion for plugin developers.

### DIFF
--- a/src/NetherPortalScanner.cpp
+++ b/src/NetherPortalScanner.cpp
@@ -296,7 +296,7 @@ void cNetherPortalScanner::OnDisabled(void)
 	}
 
 	FLOGD("Placing player at {0}", Position);
-	m_Entity->MoveToWorld(m_World, Position, true);
+	m_Entity->MoveToWorld(m_World, Position, true, false);
 	delete this;
 }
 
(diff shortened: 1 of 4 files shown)


peterbell10:
Lua API docs also need to be updated (or should we just deprecated the argument from Lua?)

src/NetherPortalScanner.cpp
```diff
@@ -296,7 +296,7 @@ void cNetherPortalScanner::OnDisabled(void)
 	}
 
 	FLOGD("Placing player at {0}", Position);
-	m_Entity->MoveToWorld(m_World, Position, true);
+	m_Entity->MoveToWorld(m_World, Position, true, false);
```
peterbell10: There are also the end portal cases to consider.

mathiascode: As far as I can tell, these cases already disable respawn packets explicitly.

mathiascode:
Easier to just leave it in in my opinion.

src/NetherPortalScanner.cpp
peterbell10: Ah you're right, I was forgetting there are multiple signatures.

mathiascode: I wish there weren't so many, but what's done is done.

madmaxoft:
I believe the respawn packet should only be sent when changing dimensions. If you travel from one overworld to another, you must not send a respawn packet.

mathiascode:
Afaik, it's always sent in vanilla (or at least some equivalent to it). If you don't send it, the client jitters around for a bit until the player has teleported, even if the dimension is the same.

PR #4542, opened 2020-03-24
Decrease attack cooldown for monsters

mathiascode:
Monsters attacked way too slowly compared to vanilla.


mathiascode:
Should we specify the attack rates in ticks, if the user desires more rapid attacks?

peterbell10:
It's already a float so can't they just do `AttackRate=0.05` for one tick.

mathiascode:
You're right, I didn't realize since the values in monsters.ini didn't include decimal points.

PR #4549, opened 2020-03-25
Fix off-by-one error and clean error messages in manual bindings for BlockArea

### LINKED ISSUE #4454: cBlockArea::FillRelCuboid: Invalid write of size 1
### Actual behavior
This function writes out-of-bounds.

```lua
local obj = cBlockArea()
obj:Create(10, 10, 10, 47)
obj:FillRelCuboid(1, 1, 1, 98, 1, 1, 1, 1, 1, 1, 1)
```

Running this code under debug will assert here:
https://github.com/cuberite/cuberite/blob/fdc190613a1be12a22413806dca5acbb86ba18f4/src/BlockArea.cpp#L2159

bearbin:
Fixes #4454

If we want to maintain the old behaviour then the last commit should be excluded from this PR, if not (the old behaviour was never used, else it created a crash!) and to make the behaviour of the FillRellCuboid consistent with other functions in the API then it should be left in.


src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -154,16 +154,27 @@ static int tolua_cBlockArea_FillRelCuboid(lua_State * a_LuaState)
...
+	if (!(self->IsValidRelCoords(bounds.p1) && self->IsValidRelCoords(bounds.p2)))
+	{
+		return L.FApiParamError(
+			"The bounds ({0} - {1}) are out of range ({2} - {3})",
+			bounds.p1,
+			bounds.p2,
+			Vector3i(0, 0, 0),
+			(self->GetSize() - Vector3i{1, 1, 1})
+		);
+	}
```
madmaxoft: This changes the function behavior, previously out-of-bounds coords were supposed to work by auto-clamping them, now the operation is downright refused. I'd have preferred the old way - clamp, possibly with a warning message.

bearbin: This does indeed change the function behaviour, but the previous behaviour was crashing the whole server and now it just generates a Lua error. I wanted this function to be consistent with all the others in the BlockArea API, which also generate an error when out-of-range operations are attempted. Is there some reason why clamping is preferable for this function?

madmaxoft: I'm from a "best effort" camp. Like many UI toolkits, if the coords are invalid, but still the operation can be done in a "limited fashion", then do it (and possibly warn about it) rather than refuse altogether.
The other functions don't make sense to even try. But with filling a rect, one can argue that clamping is the ideal "limited fashion".

bearbin:
This PR needs another review and a decision on whether to include the last commit or not.

src/Bindings/ManualBindings_BlockArea.cpp
tigerw: I agree best effort is nice but if as bearbin says everything else is strict then consistency is better. Having to piecemeal support cases where plugins pass bad input and expect it to silently work because we never warned them runs the risk of turning into a maintenance headache (see Windows).

madmaxoft: Okay, overruled.

Issue #4468, opened 2020-02-10
Piston drops cPickup of attached block ( if sticky ) or of itself on retraction


theophriene:
Client version:   1.12.2
Server OS:        Linux
Commit id:        a867388f6400729e2bd881f7f41c69fce130371f

### Steps to reproduce
1. Place a piston and power it. After it extends unpower it.
2. During retraction it will drop piston block, sticky piston block, or the block which was attached to the sticky piston before retraction

![Screenshot from 2020-02-10 06-33-41](https://user-images.githubusercontent.com/60849082/74117056-27106500-4bae-11ea-9f27-8a163b442a28.png)
![Screenshot from 2020-02-10 06-33-32](https://user-images.githubusercontent.com/60849082/74117061-2ed00980-4bae-11ea-9491-3fb9f062e1b8.png)

### Server log
```
[6297524e64d3032c|06:33:23] HandleRightClick: {22, 79, 14}, face 1, Hand: 0, HeldItem: redstonedust:0 * 1; Dist: 3.36
[6297524e64d3032c|06:33:23] Destroying entity #157 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:24] Destroying entity #158 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:24] Destroying entity #159 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:24] Destroying entity #160 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:25] Destroying entity #161 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:25] Destroying entity #162 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:26] Destroying entity #163 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:26] Destroying entity #164 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:27] Destroying entity #165 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:27] Destroying entity #166 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:27] Destroying entity #167 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:28] Destroying entity #168 (cPickup) from chunk (1, 1)
[6297524e64d3032c|06:33:28] Destroying entity #169 (cPickup) from chunk (1, 1)
[...]

PR #4552, opened 2020-03-26
Fix for piston duplication on retract

SirRobo:
Stops piston duplication bug and doesn't break water and lava simulation.

I haven't found anything that is broken after doing this, I can still mine and pick it up in survival.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -640,6 +640,8 @@ void cChunkMap::SetBlock(Vector3i a_BlockPos, BLOCKTYPE a_BlockType, NIBBLETYPE 
 		GetBlockTypeMeta(a_BlockPos, blockType, blockMeta);
 		cChunkInterface ChunkInterface(this);
 		
+		// Hotfix for https://github.com/cuberite/cuberite/issues/4468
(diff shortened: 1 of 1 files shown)


SirRobo:
This is a hotfix for https://github.com/cuberite/cuberite/issues/4468

madmaxoft:
This change goes directly against the 1.13 rewrite.

If you need to disable a specific block's handler, then change that block's handler, don't add exceptions into the core code.

bearbin:
> This is a hotfix for #4468

I meant a comment in the code.

> If you need to disable a specific block's handler, then change that block's handler, don't add exceptions into the core code.

The point is that presently there's no way to modify the OnBroken handler for pistons to only drop blocks when called from SetBlock. Once a proper solution is developed, this can be removed.

SirRobo:
Yeah sorry about that, I got a bit confused. I'll add the link to that issue in the code now.

madmaxoft:
There seems to be a whitespace issue in the newly added code.

bearbin:
Merged manually after whitespace fix.

SirRobo:
I have removed the extra tabs in my new commit

PR #4555, opened 2020-03-26
Fix typo in block break particle coordinate

### LINKED ISSUE #2482: Sheep eat grass from several blocks away
Today I noticed that sheep, when eating, pull the grass out of the dirt 10-20 blocks away from then, seemingly at random.

### LINKED ISSUE #4318: Random explosions of dirt particles above the ground
Client version:   1.12.2
Server OS:        Debian 9
Commit id:        7d512f21913b3ed7dee2a025fdc1dea77a08d83b

### Actual behavior
There are random explosions of dirt particules as if someone was breaking dirt blocks, but in random places above the ground in the overworld. I see it every few seconds.
Edit : it also makes the sound of dirt when it's destroyed

I just compiled and started the server and it does that on its own.

mathiascode:
Fixes #2482
Fixes #4318

### DIFF
--- a/src/Mobs/Sheep.cpp
+++ b/src/Mobs/Sheep.cpp
@@ -117,7 +117,7 @@ void cSheep::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 			{
 				// The sheep ate the grass so we change it to dirt
 				m_World->SetBlock(PosX, PosY, PosZ, E_BLOCK_DIRT, 0);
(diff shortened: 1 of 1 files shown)


madmaxoft:
How does this fix anything, by making the constructor explicit? Or was a different `Vector3` type assumed? Still, how is that different?

mathiascode:
Actually, it should be a Vector3i. Now I'm a bit confused, I reverted my change, and block particles still show up in the correct position...

mathiascode:
When you actually want bugs to appear, they seemingly vanish. How typical.

mathiascode:
Noooo, there's a typo... PosX should be PosZ.

TooAngel:
I like that

madmaxoft:
And that's why we should use `Vector3`s everywhere instead of raw params :)

PR #4557, opened 2020-03-27
Updated plugin list format

mathiascode:
One potential solution for issues where an ini parser removes duplicate keys when the plugin list is loaded.

```
[Plugins]
WorldEdit=1
Core=1
ChatLog=1
ProtectionAreas=0
```

1 enables a plugin, 0 disables it. Commenting out a line can also be used to disable a plugin. I also thought about having a single line listing all enabled plugins with commas, but it would be harder to quickly disable a plugin.

### DIFF
--- a/src/Bindings/PluginManager.cpp
+++ b/src/Bindings/PluginManager.cpp
@@ -1736,7 +1736,8 @@ AStringVector cPluginManager::GetFoldersToLoad(cSettingsRepositoryInterface & a_
 
 	AStringVector res;
 
-	// Get the old format plugin list, and migrate it:
+	// Get the old format plugin list, and migrate it.
+	// Upgrade path added on 27/03/2020
 	auto OldValues = a_Settings.GetValues("Plugins");
 	for (auto NameValue : OldValues)
 	{
@@ -1744,7 +1745,11 @@ AStringVector cPluginManager::GetFoldersToLoad(cSettingsRepositoryInterface & a_
 		if (ValueName.compare("Plugin") == 0)
 		{
 			AString PluginFile = NameValue.second;
-			if (!PluginFile.empty())
+			if (
+				!PluginFile.empty() &&
+				(PluginFile != "0") &&
+				(PluginFile != "1")
+			)
 			{
 				a_Settings.DeleteValue("Plugins", ValueName);
 				a_Settings.SetValue("Plugins", PluginFile, "1");
--- a/src/Bindings/PluginManager.cpp
+++ b/src/Bindings/PluginManager.cpp
@@ -154,8 +154,9 @@ void cPluginManager::ReloadPluginsNow(cSettingsRepositoryInterface & a_Settings)
 void cPluginManager::InsertDefaultPlugins(cSettingsRepositoryInterface & a_Settings)
 {
 	a_Settings.AddKeyName("Plugins");
-	a_Settings.AddValue("Plugins", "Plugin", "Core");
-	a_Settings.AddValue("Plugins", "Plugin", "ChatLog");
+	a_Settings.AddValue("Plugins", "Core", "1");
+	a_Settings.AddValue("Plugins", "ChatLog", "1");
+	a_Settings.AddValue("Plugins", "ProtectionAreas", "0");
 }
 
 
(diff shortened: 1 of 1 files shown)


madmaxoft:
Simple, working, and has an upgrade path. /me likes.

Perhaps a comment should be added to the upgrade path, as to when this was upgraded, so that we can remove it after some time.

src/Bindings/PluginManager.cpp
```diff
@@ -1733,22 +1734,35 @@ AStringVector cPluginManager::GetFoldersToLoad(cSettingsRepositoryInterface & a_
...
 	{
 		AString ValueName = NameValue.first;
 		if (ValueName.compare("Plugin") == 0)
 		{
 			AString PluginFile = NameValue.second;
 			if (!PluginFile.empty())
 			{
-				res.push_back(PluginFile);
+				a_Settings.DeleteValue("Plugins", ValueName);
+				a_Settings.SetValue("Plugins", PluginFile, "1");
```
peterbell10: If I have a plugin called `Plugin` then I might write `Plugin=1`. This would then replace it with `1=1`.

mathiascode: This is why code reviews are great.

src/Bindings/PluginManager.cpp
```diff
@@ -1733,22 +1734,40 @@ AStringVector cPluginManager::GetFoldersToLoad(cSettingsRepositoryInterface & a_
 		InsertDefaultPlugins(a_Settings);
 	}
 
-	// Get the list of plugins to load:
 	AStringVector res;
-	auto Values = a_Settings.GetValues("Plugins");
-	for (auto NameValue : Values)
+
+	// Get the old format plugin list, and migrate it.
+	// Upgrade path added on 27/03/2020
```
madmaxoft: Please use the one and only ISO date format: 2020-03-27

src/Bindings/PluginManager.cpp
```diff
@@ -1733,22 +1734,40 @@ AStringVector cPluginManager::GetFoldersToLoad(cSettingsRepositoryInterface & a_
...
 			AString PluginFile = NameValue.second;
-			if (!PluginFile.empty())
+			if (
+				!PluginFile.empty() &&
+				(PluginFile != "0") &&
+				(PluginFile != "1")
+			)
 			{
-				res.push_back(PluginFile);
+				a_Settings.DeleteValue("Plugins", ValueName);
```
madmaxoft: Is it possible that this deletion can become confused about which of the `Plugin=` lines to remove?

src/Bindings/PluginManager.cpp
mathiascode: The chosen one

src/Bindings/PluginManager.cpp
mathiascode: DeleteValue should remove the first value it finds.

madmaxoft:
OH NO, we completely missed one aspect. The old format specified not only the plugins to load, but also their load order. The new format doesn't specify the load order. Some plugins might have depended on the order.

bearbin:
I'm not sure I understand - with this PR plugins will still be loaded in the order they are in the INI file, won't they?

madmaxoft:
They are? Still that's a non-standard INI file usage, as the key order in the INI file is usually not relevant.
Okay, I'll believe this then and won't mention it again.

tigerw:
> OH NO

this is too funny XD

I am amused at the 9-day delay the epiphany took. Was it one of those "suddenly think of an embarrassing event five years ago whilst lying awake at night" moments?

bearbin:
OTOH, we did have the opportunity to add proper ordering not relying on ordering in the file, if we allow any number > 0 as a value for the plugin key, then the list of plugins to load could be sorted by this with higher-numbered plugins being loaded after lower numbered ones. We could still do that, it would be backwards compatible.

madmaxoft:
I'd say we should do that, @bearbin , great idea. Make it consistent.

@tigerw yes, one of those, just when I was adding my new plugin to the list on my dev server.

mathiascode:
Have fun writing documentation for the average user. :P

PR #4561, opened 2020-03-27
Inventory changes

### LINKED ISSUE #4092: opening chest -> Armor replaced
Client version:   1.12.2
Server OS:        Raspbian Stretch
Commit id:        0dd172b80ffc949cf87119ad2003589d7eb82c46

### Expected behavior
--
### Actual behavior
bildschirmfoto 2017-12-16 um 13 18 33
bildschirmfoto 2017-12-16 um 13 19 06

opening a double chest (no redstone) sometimes causes strange things. 
The armor is "gone".
Instead player holds "line1" from the chest and the item lower left corner as a shield.
(had 7 sand - they disappeared when clicking) 

The Items stay in the "workbench" even if inventory is closed.

When items that replaced the armor are dropped, they return back to what they've been.

bildschirmfoto 2017-12-16 um 13 52 14

dropped them into a fire  :-( so ... 5 lost) 

Seems there are 2 variables mixed up...

Happend twice ... not sure

mathiascode:
Possibly fixes #4092

https://github.com/cuberite/cuberite/pull/1883 changed the window ID counter to start from 0 instead of 1. Wouldn't this potentially conflict with player inventories, which are always ID 0?

### DIFF
--- a/src/UI/Window.cpp
+++ b/src/UI/Window.cpp
@@ -21,7 +21,7 @@
 
 
 
-Byte cWindow::m_WindowIDCounter = 1;
+Byte cWindow::m_WindowIDCounter = 0;
 
 
 
@@ -34,6 +34,8 @@ cWindow::cWindow(WindowType a_WindowType, const AString & a_WindowTitle) :
 	m_IsDestroyed(false),
 	m_Owner(nullptr)
 {
+	ASSERT((m_WindowID > 0) && (m_WindowID < 127));
+
 	if (a_WindowType == wtInventory)
 	{
 		m_WindowID = 0;
--- a/src/UI/Window.cpp
+++ b/src/UI/Window.cpp
@@ -21,23 +21,19 @@
 
 
 
-Byte cWindow::m_WindowIDCounter = 0;
+char cWindow::m_WindowIDCounter = 1;
 
 
 
 
 
 cWindow::cWindow(WindowType a_WindowType, const AString & a_WindowTitle) :
-	m_WindowID(static_cast<char>((++m_WindowIDCounter) % 127)),
+	m_WindowID((++m_WindowIDCounter) % 127),
 	m_WindowType(a_WindowType),
 	m_WindowTitle(a_WindowTitle),
 	m_IsDestroyed(false),
 	m_Owner(nullptr)
 {
-	// The window ID is signed in protocol 1.7, unsigned in protocol 1.8. Keep out of trouble by using only 7 bits:
-	// Ref.: https://forum.cuberite.org/thread-1876.html
-	ASSERT((m_WindowID >= 0) && (m_WindowID < 127));
-
 	if (a_WindowType == wtInventory)
 	{
 		m_WindowID = 0;
--- a/src/UI/Window.cpp
+++ b/src/UI/Window.cpp
@@ -21,7 +21,7 @@
 
 
 
-char cWindow::m_WindowIDCounter = 1;
+Byte cWindow::m_WindowIDCounter = 1;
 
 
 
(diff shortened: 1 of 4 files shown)


src/UI/Window.cpp
```diff
@@ -21,23 +21,19 @@
...
-Byte cWindow::m_WindowIDCounter = 0;
+char cWindow::m_WindowIDCounter = 1;
 
 
 
 
 
 cWindow::cWindow(WindowType a_WindowType, const AString & a_WindowTitle) :
-	m_WindowID(static_cast<char>((++m_WindowIDCounter) % 127)),
+	m_WindowID((++m_WindowIDCounter) % 127),
```
peterbell10: This is a pre-increment so the ID of 0 won't be the first number assigned. Wrapping-around to 0 might be problematic though. Maybe this should be changed to 
```cpp
UInt8 cWindow::GetNextId()
{
    UInt8 Id = ++m_WindowIDCounter;
    return (Id == 0) ? ++m_WindowIDCounter : Id;
}

File: src/UI/Window.cpp [hunk 2]
```diff
@@ -21,23 +21,19 @@
 
 
 
-Byte cWindow::m_WindowIDCounter = 0;
+char cWindow::m_WindowIDCounter = 1;
```
peterbell10: `char` overflow is undefined behaviour. Besides, if it's unsigned in all current protocols then why not make it match?

mathiascode:
Is there a reason why MC doesn't use an unsigned byte for window IDs in the Set Slot-packet? https://wiki.vg/Protocol

peterbell10:
>To set the cursor (the item currently dragged with the mouse), **use -1 as Window ID** and as Slot. 

Yup, that's going to be an issue. So, keep the assert in place and just make it `m_WindowID > 0`.

mathiascode:
Ah, right. What was the point of changing to unsigned in 1.8 to begin with?

src/UI/Window.cpp
madmaxoft: I think `m_WindowID(((++m_WindowIDCounter) % 127) + 1),` should do the trick. The result of the modulo is in range [0, 126], add a 1 to get range [1, 127]. Whether to post- or pre-increment doesn't really matter here, just means the first `1` will be skipped.

madmaxoft:
We should probably split the `cProtocol::SendInventorySlot` function, so that it only sets slots in inventory windows, and add a new `cProtocol::SendDraggedItem` (name subject to discussion) that updates the item currently dragged at the cursor. But not in this PR.

src/UI/Window.cpp
```diff
@@ -28,15 +28,13 @@ Byte cWindow::m_WindowIDCounter = 0;
...
+	m_WindowID(((++m_WindowIDCounter) % 127) + 1),
 	m_WindowType(a_WindowType),
 	m_WindowTitle(a_WindowTitle),
 	m_IsDestroyed(false),
 	m_Owner(nullptr)
 {
-	// The window ID is signed in protocol 1.7, unsigned in protocol 1.8. Keep out of trouble by using only 7 bits:
-	// Ref.: https://forum.cuberite.org/thread-1876.html
-	ASSERT((m_WindowID >= 0) && (m_WindowID < 127));
+	ASSERT((m_WindowID > 0) && (m_WindowID < 127));
```
peterbell10:
```suggestion
	ASSERT((m_WindowID > 0) && (m_WindowID <= 127));
```

mathiascode:
Can someone fix clang compilation on master? I'm out of time for today.

bearbin:
Oops, I have now added casts to UInt8, which fixes the build.

madmaxoft:
Well well well, someone merged before the CI was green :P

bearbin:
I thought travis had passed and appveyor was backlogged as usual, I should have looked a bit more closely!

TooAngel:
You can configure the repository so that merging is only possible if all checks are green.

madmaxoft:
It is configured so here as well, but we're pretty used to "admin overriding" anyway :)

TooAngel:
:-) well, ok, I would have another suggestion, but it sounds like this wouldn't be appreciated ;-)

PR #4563, opened 2020-03-27
Implement wither skeletons

mathiascode:
Depends on https://github.com/cuberite/cuberite/pull/4538, so wither skeletons can be remapped to skeletons in MC versions prior to 1.11. Otherwise, they will appear invisible.


src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -738,6 +738,7 @@ class SerializerCollector:
 			case mtVillager:      EntityClass = "Villager";       break;
 			case mtWitch:         EntityClass = "Witch";          break;
 			case mtWither:        EntityClass = "WitherBoss";     break;
+			case mtWitherSkeleton:        EntityClass = "WitherSkeleton";     break;
```
bearbin: Something's gone wrong with the alignment here.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1562,6 +1562,8 @@ void cWSSAnvil::LoadEntityFromNBT(cEntityList & a_Entities, const cParsedNBT & a
 		{ "minecraft:witch",               &cWSSAnvil::LoadWitchFromNBT },
 		{ "WitherBoss",                    &cWSSAnvil::LoadWitherFromNBT },
 		{ "minecraft:wither",              &cWSSAnvil::LoadWitherFromNBT },
+		{ "WitherSkeleton",                &cWSSAnvil::LoadWitherFromNBT },
```
bearbin: This line doesn't look right.

mathiascode: That explains why random withers spawned. Thanks!

src/WorldStorage/NBTChunkSerializer.cpp
mathiascode: Looks like I forgot to align them entirely.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2662,15 +2664,7 @@ void cWSSAnvil::LoadSilverfishFromNBT(cEntityList & a_Entities, const cParsedNBT
...
 {
-	int TypeIdx = a_NBT.FindChildByName(a_TagIdx, "SkeletonType");
-	if (TypeIdx < 0)
-	{
-		return;
-	}
-
-	bool Type = ((a_NBT.GetByte(TypeIdx) == 1) ? true : false);
-
-	std::unique_ptr<cSkeleton> Monster = cpp14::make_unique<cSkeleton>(Type);
```
madmaxoft: Did the NBT format for skeletons / wither skeletons change? Shouldn't we still support both formats?

src/Mobs/Monster.cpp
```diff
@@ -883,7 +885,7 @@ void cMonster::InStateEscaping(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 void cMonster::ResetAttackCooldown()
 {
-	m_AttackCoolDownTicksLeft = static_cast<int>(20 * m_AttackRate);  // A second has 20 ticks, an attack rate of 1 means 1 hit every second
+	m_AttackCoolDownTicksLeft = static_cast<int>(m_AttackRate);  // A second has 20 ticks, an attack rate of 1 means 1 hit every 3 seconds
```
madmaxoft: This looks like a bad change. Now the comment doesn't make any sense (why does it mention the 20 ticks?) and why is it 3 seconds again? Wasn't this changed recently?

mathiascode: Just a debug change that shouldn't have made it in.

src/WorldStorage/WSSAnvil.cpp
mathiascode: We should probably keep this for legacy support.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2662,15 +2664,34 @@ void cWSSAnvil::LoadSilverfishFromNBT(cEntityList & a_Entities, const cParsedNBT
 
 void cWSSAnvil::LoadSkeletonFromNBT(cEntityList & a_Entities, const cParsedNBT & a_NBT, int a_TagIdx)
 {
+	// Wither skeleton is a separate mob in Minecraft 1.11+, but
+	// we need this to load them from older worlds
```
madmaxoft: This comment doesn't explain sufficiently why there's a witherskeleton in skeleton loading code. Add `where wither skeletons were only a skeletons with a flag`

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2662,15 +2664,34 @@ void cWSSAnvil::LoadSilverfishFromNBT(cEntityList & a_Entities, const cParsedNBT
...
+		if (!LoadMonsterBaseFromNBT(*Monster.get(), a_NBT, a_TagIdx))
+		{
+			return;
+		}
+
+		a_Entities.emplace_back(std::move(Monster));
+		return;
+	}
+
+	std::unique_ptr<cSkeleton> Monster = cpp14::make_unique<cSkeleton>();
```
madmaxoft: Not sure if this works as-is, try it locally first.
```suggestion
	std::unique_ptr<cMonster> Monster;
	if (a_NBT.GetByte(TypeIdx) == 1)
	{
		Monster.reset(new cWitherSkeleton);
	}
	else
	{
		Monster.reset(new cSkeleton);
	}
```

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2865,6 +2886,26 @@ void cWSSAnvil::LoadWitherFromNBT(cEntityList & a_Entities, const cParsedNBT & a
 
 
 
+void cWSSAnvil::LoadWitherSkeletonFromNBT(cEntityList & a_Entities, const cParsedNBT & a_NBT, int a_TagIdx)
+{
+	std::unique_ptr<cWitherSkeleton> Monster = cpp14::make_unique<cWitherSkeleton>();
```

src/WorldStorage/WSSAnvil.cpp
mathiascode: Seems to be working. Thanks!

mathiascode:
Rebased

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2665,33 +2665,23 @@ void cWSSAnvil::LoadSilverfishFromNBT(cEntityList & a_Entities, const cParsedNBT
...
 {
-	// Wither skeleton is a separate mob in Minecraft 1.11+, but
-	// we need this to load them from older worlds
+	// Wither skeleton is a separate mob in Minecraft 1.11+, but we need this to
+	// load them from older worlds where wither skeletons were only a skeleton with a flag
 	int TypeIdx = a_NBT.FindChildByName(a_TagIdx, "SkeletonType");
 	if (TypeIdx < 0)
 	{
 		return;
 	}
```
madmaxoft, line 2674: Is the `SkeletonType` still present in the newer data where witherskeletons are saved as separate entities? If not, this code will drop all skeletons.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -2663,15 +2665,20 @@ void cWSSAnvil::LoadSilverfishFromNBT(cEntityList & a_Entities, const cParsedNBT
...
+	// Wither skeleton is a separate mob in Minecraft 1.11+, but we need this to
+	// load them from older worlds where wither skeletons were only a skeleton with a flag
 	int TypeIdx = a_NBT.FindChildByName(a_TagIdx, "SkeletonType");
-	if (TypeIdx < 0)
+
+	std::unique_ptr<cMonster> Monster;
+	if ((TypeIdx > 0) && (a_NBT.GetByte(TypeIdx) == 1))
 	{
-		return;
+		Monster.reset(new cWitherSkeleton);
```
tigerw, line 2675: I thought make_unique supported upcasting?

tigerw:
Seems reasonable. How do they behave in terms of AI right now?

mathiascode:
The AI is the same as the other walking monsters (good enough). The walking speed is too slow, but I'm waiting for https://github.com/cuberite/cuberite/pull/4500.

src/WorldStorage/WSSAnvil.cpp
tigerw: Why we use `new` + reset instead of make_unique

madmaxoft: I thought `make_unique` could be used only in the initial assignment, but that doesn't make sense. @mathiascode Try `Monster = cpp14::make_unique<cWitherSkeleton>()` here.

tigerw:
Looks great!

PR #4570, opened 2020-03-28
Louder ambient sound for ghasts

mathiascode:
Sound volumes should ideally be specified for every mob, but I don't feel like going through every mob file right now...


mathiascode:
This should be done when mob step sounds are implemented, if you want to keep your sanity.

PR #4572, opened 2020-03-28
Fixed Clang builds

bibo38:
A window id, which was signed was used in an unsigned context. This caused a Clang compilation error.
Since there is one packet (Set Slot, 0x16) that allows negative values, it probably influenced the variable to have a signed value.


bibo38:
@bearbin was faster :rofl:

tigerw:
Thanks though!

PR #4580, opened 2020-03-29
Replace buckets to the selected hotbar slot, rather than the first available.

bearbin:
Replicates vanilla behaviour, as well as being more logical.

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -10745,6 +10745,17 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Places a block while impersonating the player. The {{OnPlayerPlacingBlock|HOOK_PLAYER_PLACING_BLOCK}} hook is called before the placement, and if it succeeds, the block is placed and the {{OnPlayerPlacedBlock|HOOK_PLAYER_PLACED_BLOCK}} hook is called. Returns true iff the block is successfully placed. Assumes that the block is in a currently loaded chunk.",
 				},
+				ReplaceOneEquippedItemTossRest =
+				{
+					Params =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+					},
+					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, tries to place the specified item elsewhere in the inventory. If this is not possible, then any remaining items are tossed. Assumes that the currently equipped slot contains at least one item, behaviour is not defined in the case that it does not.",
+				},
 				Respawn =
 				{
 					Notes = "Restores the health, extinguishes fire, makes visible and sends the Respawn packet.",
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -6637,7 +6637,7 @@ These ItemGrids are available in the API and can be manipulated by the plugins, 
 							Type = "number",
 						},
 					},
-					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, optionally (default true) tries to place the specified item elsewhere in the inventory. Returns the number of items successfully added. If the currently equipped slot is empty, this is ignored.",
+					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, optionally (default true) tries to place the specified item elsewhere in the inventory. Returns the number of items successfully added. If the currently equipped slot is empty, its contents are simply set to the given Item.",
 				},
 				SendEquippedSlot =
 				{
@@ -10754,7 +10754,7 @@ a_Player:OpenWindow(Window);
 							Type = "cItem",
 						},
 					},
-					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, tries to place the specified item elsewhere in the inventory. If this is not possible, then any remaining items are tossed. If the currently equipped slot is empty, this is ignored.",
+					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, tries to place the specified item elsewhere in the inventory. If this is not possible, then any remaining items are tossed. If the currently equipped slot is empty, its contents are simply set to the given Item.",
 				},
 				Respawn =
 				{
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -6223,7 +6223,7 @@ These ItemGrids are available in the API and can be manipulated by the plugins, 
 							Type = "number",
 						},
 					},
-					Notes = "Adds an item to the storage; if AllowNewStacks is true (default), will also create new stacks in empty slots. Returns the number of items added",
+					Notes = "Adds an item to the storage; if AllowNewStacks is true (default), will also create new stacks in empty slots. Fills existing stacks first and fills the hotbar before the main inventory. Returns the number of items added",
 				},
 				AddItems =
 				{
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -6617,6 +6617,28 @@ These ItemGrids are available in the API and can be manipulated by the plugins, 
 					},
 					Notes = "Removes one item from the hotbar's currently selected slot. Returns true on success.",
 				},
+				ReplaceOneEquippedItem =
+				{
+					Params =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+						{
+							Name = "TryOtherSlots",
+							Type = "boolean",
+							IsOptional = true,
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+					},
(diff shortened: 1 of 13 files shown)


peterbell10:
How about adding `cInventory::ReplaceOneEquippedItem(cItem NewItem)` and using it everywhere similar functionality is required. I think there is: bucketing liquids, milking cows, milking mooshrooms for soup and filling water bottles.

bearbin:
That would be a good idea, indeed I forgot about milking and cauldrons in this implementation so I'll implement those.

bearbin:
This isn't quite ready yet, I need to add the APIDump documentation for the new methods, and allow filling water bottles from water sources rather than just cauldrons. The code that has been written is ready for review though.

src/Blocks/BlockCauldron.h
```diff
@@ -27,17 +27,37 @@ class cBlockCauldronHandler :
...
 			{
 				if (Meta < 3)
 				{
 					a_ChunkInterface.SetBlockMeta(a_BlockX, a_BlockY, a_BlockZ, 3);
 					if (!a_Player.IsGameModeCreative())
 					{
-						a_Player.GetInventory().RemoveOneEquippedItem();
-						a_Player.GetInventory().AddItem(cItem(E_ITEM_BUCKET));
+						// Remove water bucket, add empty bucket.
+						a_Player.GetInventory().SetEquippedItem(cItem(E_ITEM_BUCKET));
```
madmaxoft: This should use the `ReplaceOneEquippedItem` function as well, in case plugins in the future decide that water buckets are stackable.

src/Blocks/BlockCauldron.h
```diff
@@ -47,11 +67,33 @@ class cBlockCauldronHandler :
...
+			case E_ITEM_POTION:
+			{
+				// Refill cauldron with water bottles.
+				if ((Meta < 3) && (EquippedItem.m_ItemDamage == 0))
+				{
+					a_ChunkInterface.SetBlockMeta(Vector3i(a_BlockX, a_BlockY, a_BlockZ), ++Meta);
+					if (!a_Player.IsGameModeCreative())
+					{
+						// Remove water bucket, add empty bucket.
+						a_Player.GetInventory().SetEquippedItem(cItem(E_ITEM_GLASS_BOTTLE));
```
madmaxoft: Copypasta comment; also use `ReplaceOneEquippedItem`.

src/Blocks/BlockCauldron.h
```diff
@@ -47,11 +67,33 @@ class cBlockCauldronHandler :
...
+					// Give new potion when the gamemode is not creative:
+					if (!a_Player.IsGameModeCreative())
+					{
+						auto & Inventory = a_Player.GetInventory();
+						auto NewItem = cItem(E_ITEM_POTION);
+						if (Inventory.ReplaceOneEquippedItem(NewItem) == 0)
+						{
+							// The potion didn't fit, toss it as a pickup:
+							a_Player.TossPickup(NewItem);
+						}
```
madmaxoft: This seems like it will be a pretty common construct. Add a function `cPlayer::ReplaceOneEquippedItemTossRest(const cItem &)` that does this (plus calculates the correct amount of leftover items to toss).

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -10712,6 +10745,17 @@ a_Player:OpenWindow(Window);
...
+				ReplaceOneEquippedItemTossRest =
+				{
+					Params =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+					},
+					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, tries to place the specified item elsewhere in the inventory. If this is not possible, then any remaining items are tossed. Assumes that the currently equipped slot contains at least one item, behaviour is not defined in the case that it does not.",
```
madmaxoft: I think the behavior is quite defined - Doesn't remove anything if nothing is there; still tries to stuff Item in there.

madmaxoft: (same with `cInventory:ReplaceOneEquippedItem`)

src/Entities/Player.cpp
```diff
@@ -1942,6 +1942,26 @@ void cPlayer::TossEquippedItem(char a_Amount)
 
 
 
+void cPlayer::ReplaceOneEquippedItemTossRest(const cItem & a_Item)
+{
+	auto Placed = GetInventory().ReplaceOneEquippedItem(a_Item);
```
madmaxoft: In order for the code to be immediately readable without needing to check the type that `ReplaceOneEquippedItem` returns, the variable should be named `NumPlaced` or similar, to indicate that it contains amount, rather than a `cItem`.

src/Entities/Player.cpp
```diff
@@ -1942,6 +1942,26 @@ void cPlayer::TossEquippedItem(char a_Amount)
...
+
+	char ItemsToToss = a_Item.m_ItemCount - static_cast<char>(Placed);
```
madmaxoft: Again, indicate amount, not content, through the name.

Server/Plugins/APIDump/APIDesc.lua
bearbin: I guess I was drawing a difference between the exposed interface (behaviour is not defined) and the implementation, where we have to decide one way or the other. Do you think there is an advantage in specifying the behaviour one way or the other in the public interface?

madmaxoft: I'd say in this case it makes perfect sense to specify the behavior. Honestly, I think Undefined Behavior is a nightmare concept, if we can avoid it, we should.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6617,6 +6617,28 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
...
+							IsOptional = true,
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "number",
+						},
+					},
+					Notes = "Removes one item from the the current equipped item stack, and attempts to add the specified item stack back to the same slot. If it is not possible to place the item in the same slot, optionally (default true) tries to place the specified item elsewhere in the inventory. Returns the number of items successfully added. If the currently equipped slot is empty, this is ignored.",
```
madmaxoft: `this is ignored` Not true, the `Item` is still added. How about `If the currently equipped slot is empty, it is simply set to Item.` ?

bearbin: `this` was intended to be `the fact that the currently equipped slot is empty`, but I can see how this phrasing would lead to confusion. I'll change it.

tigerw:
The big green button beckons...

bearbin:
Well, if you approve it's there for you to click isn't it @tigerw

src/Inventory.cpp
```diff
@@ -219,6 +222,50 @@ bool cInventory::RemoveOneEquippedItem(void)
...
+	RemoveOneEquippedItem();
+
+	auto EquippedItem = GetEquippedItem();
+	if (EquippedItem.IsEmpty())
+	{
+		SetEquippedItem(a_Item);
+		return a_Item.m_ItemCount;
+	}
+
+	// Handle case when equipped item is the same as the replacement item.
```
tigerw, line 237: Hm. When would we remove an item and add the same item back?

src/Items/ItemPotion.h
```diff
@@ -77,8 +77,7 @@ class cItemPotionHandler:
 
 		if (!a_Player->IsGameModeCreative())
 		{
-			a_Player->GetInventory().RemoveOneEquippedItem();
-			a_Player->GetInventory().AddItem(cItem(E_ITEM_GLASS_BOTTLE));
+			a_Player->GetInventory().SetEquippedItem(cItem(E_ITEM_GLASS_BOTTLE));
```
tigerw, line 80: The changes use Replace...TossRest elsewhere, should glass bottles use it too?

bearbin: That was an oversight.

src/Inventory.cpp
bearbin: _We_, as the enlightened programmers, would never do such a thing; plugin developers on the other hand... It's an exported function, why not cover all of the edge cases

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -6617,6 +6617,28 @@ These ItemGrids are available in the API and can be manipulated by the plugins,
...
+				ReplaceOneEquippedItem =
+				{
+					Params =
+					{
+						{
+							Name = "Item",
+							Type = "cItem",
+						},
+						{
+							Name = "TryOtherSlots",
```
tigerw, line 6629: (minor) Perhaps it would be worth holding off exporting this for now and waiting to see if anyone requests it? Only that the code doesn't set the parameter right now.

src/Inventory.cpp
tigerw: fair enough

tigerw:
So am I correct in saying this will also fix a class of bugs in which a potentially important resource (lava buckets etc) will get lost because the inventory is full and the source (empty buckets) was stackable (so didn't vacate the slot), and we forgot to add an explicit TossPickup check?

If so, I'm certain this PR closes some issues regarding buckets.

Server/Plugins/APIDump/APIDesc.lua
madmaxoft: I believe this function is such a basic functionality, that it warrants an immediate export.

madmaxoft:
Yes, it should fix those issues, too. If you can link the issue numbers here, they can be autoclosed, too.

Server/Plugins/APIDump/APIDesc.lua
tigerw: oh I mean the "TryOtherSlots" parameter

madmaxoft: Nah, keep it.

PR #4584, opened 2020-03-30
Prevent player from being destroyed by plugins

bearbin:
Add manual binding, bails out with error message if attempted
entity to destroy is player.

Fix as suggested by @peterbell10

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -4094,9 +4094,14 @@ static int tolua_cEntity_Destroy(lua_State * tolua_S)
 	cEntity * self = nullptr;
 	L.GetStackValue(1, self);
 
+	if (lua_gettop(L) == 2)
+	{
+		LOGWARNING("cEntity:Destroy(bool) is deprecated, use cEntity:Destroy() instead.");
+	}
+
 	if (self->IsPlayer())
 	{
-		return L.ApiParamError("Cannot destroy a player");
+		return L.ApiParamError("Cannot call cEntity:Destroy() on a cPlayer, use cClientHandle:Kick() instead.");
 	}
 
 	self->Destroy();
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -4081,6 +4081,32 @@ static int tolua_cCuboid_Move(lua_State * tolua_S)
 
 
 
+static int tolua_cEntity_Destroy(lua_State * tolua_S)
+{
+	// Check the params:
+	cLuaState L(tolua_S);
+	if (!L.CheckParamSelf("cEntity"))
+	{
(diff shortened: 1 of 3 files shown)


src/Bindings/ManualBindings.cpp
```diff
@@ -4081,6 +4081,32 @@ static int tolua_cCuboid_Move(lua_State * tolua_S)
...
+		return 0;
+	}
+
+	// Get the params:
+	cEntity * self = nullptr;
+	L.GetStackValue(1, self);
+
+	if (self->IsPlayer())
+	{
+		return L.ApiParamError("Cannot destroy a player");
```
madmaxoft: You should add a note that the plugin should use `cPlayer:Kick` instead. Or should we even do that directly, rather than throwing an error?

madmaxoft:
You should also disable the tolua auto-export of the `cEntity::Destroy` function.

src/Bindings/ManualBindings.cpp
peterbell10: The author might want either kicking or respawning. I say let the plugin author decide.

madmaxoft: Calling `Destroy` to respawn doesn't seem likely to me.
A more likely scenario is a plugin that destroys all entities within some range, so it calls Destroy in a cWorld:ForEachEntity; that includes players.

peterbell10: If you kill most entities they are destroyed. If you kill a player it is respawned. In that sense it's analogous.

peterbell10: Either way, I don't think we should try and mimic `Destroy` with any other function.

madmaxoft: Okay, but at least the message to use `Kick` instead should go in. Same as with deprecation, "don't tell them it's the wrong way, tell them what's the right way instead."

src/Entities/Entity.h
```diff
@@ -289,14 +289,16 @@ class cEntity
...
+	// tolua_end
 	/** Destroys the entity, schedules it for memory freeing and broadcasts the DestroyEntity packet */
 	virtual void Destroy();
 
 	OBSOLETE void Destroy(bool a_ShouldBroadcast)
 	{
-		LOGWARNING("cEntity:Destory(bool) is deprecated, use cEntity:Destroy() instead.");
+		LOGWARNING("cEntity:Destroy(bool) is deprecated, use cEntity:Destroy() instead.");
 		Destroy();
 	}
```
peterbell10: I believe this only existed for the benefit of the lua bindings. Should be safe to delete.

bearbin: The warning or the whole function?

peterbell10: whole function

PR #4586, opened 2020-03-30
Send experience on respawn

### LINKED ISSUE #4585: XP levels resets after crossing a Nether Portail
Client version: 1.12.2
Server OS: Ubuntu 16.04 Server 
Cuberite Commit id:  6116f899dee33a6d7bd887f7a57ace17df2ab94f

### Expected behavior
When a player cross the nether portail, all of his stuff (including xp level) should be the same in the overworld and in the nether (this can be false with some plugins and specifics configurations)

### Actual behavior
Going into the Nether and then return to the overworld reset my XP

Apparently going into the Nether and return in the Overworld, don't know if it always happens or sometimes with particular conditions

mathiascode:
Fixes #4585


mathiascode:
Seems like entity effects are cleared when changing worlds, but that seems a bit out of scope for this PR.

PR #4587, opened 2020-03-30
Tweak attack ranges

### LINKED ISSUE #4543: Monsters have too large attack range
The attack range of some monsters, such as spiders and zombies, is too large. This should be tweaked in monsters.ini.

mathiascode:
Fixes #4543

If someone has a list of exact values, I would appreciate it.


madmaxoft:
I can't help but wonder, what are the units of those numbers? Blocks? Relative to per-mob built-in base?

Issue #4575, opened 2020-03-29
Explosions send all players up in the sky


Xenoxis:
Client version: 1.12.2
Server OS: Ubuntu server 16.04 i386
Cuberite Commit id:  fce548139b8c4643cd28f636d5c497d2ca8cd230

### Expected behavior
An explosion shouldn't shoot the player in the sky, and should be only for players around the explosions

### Actual behavior
If a player is around and touch by an explosion, all players (at least in the same world) are sent in the sky with a very huge power, all players mean players that are far from the explosion

### Steps to reproduce the behavior
At least 2 players I believe, TNT, Ghasts or something else that can cause an explosion

mathiascode:
Some issue with GetSpeed() at https://github.com/changyongGuo/cuberite/blob/5bcf079946922bb05ec2ecdb206db296f10704da/src/World.cpp#L1447. Happens more often since Ghasts started flying.

mathiascode:
Also see https://github.com/cuberite/cuberite/issues/4527

Xenoxis:
It happen with creeper explosions also

Xenoxis:
This bug make the game quite unplayable with more than 3 players, everyone is send up in the sky

Issue #4527, opened 2020-03-21
Improve explosion knockback


mathiascode:
Explosion knockback currently isn't enough for non-player entities, such as falling blocks and tnt.

mathiascode:
I almost feel like reverting #4251. I've experienced too many cases with extreme knockback where it shouldn't exist.

PR #4588, opened 2020-03-30
Hotfix for explosion knockback issues

mathiascode:
Hotfix for #4575

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1313,21 +1313,14 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 				a_Entity.TakeDamage(dtExplosion, nullptr, static_cast<int>((1 / std::max(1.0, DistanceFromExplosion.Length())) * 8 * ExplosionSizeInt), 0);
 			}
 
-			double Length = DistanceFromExplosion.Length();
-			if (Length <= ExplosionSizeInt)  // Entity is impacted by explosion
-			{
-				float EntityExposure = a_Entity.GetExplosionExposureRate(ExplosionPos, static_cast<float>(a_ExplosionSize));
+			// Hotfix for #4575
+			// TODO: Enhance knockback to be more vanilla-like. Reapply blast reduction.
+			auto Knockback = (1 / std::max(1.0, DistanceFromExplosion.Length())) * 6 * ExplosionSizeInt;
 
-				// Exposure reduced by armor
-				EntityExposure = EntityExposure * (1.0f - a_Entity.GetEnchantmentBlastKnockbackReduction());
+			DistanceFromExplosion.Normalize();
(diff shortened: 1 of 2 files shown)


src/ChunkMap.cpp
```diff
@@ -1313,21 +1313,14 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
...
-
-				// Exposure reduced by armor
-				EntityExposure = EntityExposure * (1.0f - a_Entity.GetEnchantmentBlastKnockbackReduction());
+			// Hotfix for #4575
+			// TODO: Enhance knockback to be more vanilla-like. Reapply blast reduction.
+			auto Knockback = (1 / std::max(1.0, DistanceFromExplosion.Length())) * 6 * ExplosionSizeInt;
 
-				double Impact = (1 - ((Length / ExplosionSizeInt) / 2)) * EntityExposure;
+			DistanceFromExplosion.Normalize();
+			DistanceFromExplosion = Vector3d(DistanceFromExplosion.x * Knockback, DistanceFromExplosion.y + (Knockback * 0.3), DistanceFromExplosion.z * Knockback);
```
bearbin: Vector3s support vector multiplication, so no need to decompose and recompose.

```suggestion
			DistanceFromExplosion *= Vector3d(1, 0.3, 1) * Knockback;
```

mathiascode: This breaks height knockback (notice the +).

src/ChunkMap.cpp
```diff
@@ -1313,21 +1313,14 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 				a_Entity.TakeDamage(dtExplosion, nullptr, static_cast<int>((1 / std::max(1.0, DistanceFromExplosion.Length())) * 8 * ExplosionSizeInt), 0);
 			}
 
-			double Length = DistanceFromExplosion.Length();
-			if (Length <= ExplosionSizeInt)  // Entity is impacted by explosion
-			{
-				float EntityExposure = a_Entity.GetExplosionExposureRate(ExplosionPos, static_cast<float>(a_ExplosionSize));
-
-				// Exposure reduced by armor
-				EntityExposure = EntityExposure * (1.0f - a_Entity.GetEnchantmentBlastKnockbackReduction());
```
peterbell10: Why remove `EntityExposure`?

mathiascode: It seems to be broken (https://github.com/cuberite/cuberite/issues/4527), and I'm not sure how to fix it/how to incorporate it.

peterbell10: `GetExplosionExposureRate` scales by the fraction of the entities bounding box that overlaps with the explosion, so should only be < 1 if the entity is on the boundary of the explosion.

You've also broken blast protection's effect on knockback.

mathiascode: If you know this code better, then feel free to make contributions or take over this PR. I didn't really want to work on it, but I don't want to leave users exposed to this bug for too long either.

PR #4590, opened 2020-03-30
Fix explosion knockback issues and tweak knockback strength

### LINKED ISSUE #4575: Explosions send all players up in the sky

peterbell10:
Closes #4588, closes #4575

This is #4588 plus a few tweaks. AFAICT, the main fix is always setting the entity's speed, even if they aren't within range of the explosion.

I also compared against vanilla and the knockback was way too strong for players a few blocks from the explosion. So, I've made it so you need to be within .2 blocks of the TNT to get the full knockback strength.

### DIFF
--- a/src/ChunkMap.cpp
+++ b/src/ChunkMap.cpp
@@ -1313,21 +1313,19 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
 				a_Entity.TakeDamage(dtExplosion, nullptr, static_cast<int>((1 / std::max(1.0, DistanceFromExplosion.Length())) * 8 * ExplosionSizeInt), 0);
 			}
 
-			double Length = DistanceFromExplosion.Length();
-			if (Length <= ExplosionSizeInt)  // Entity is impacted by explosion
-			{
-				float EntityExposure = a_Entity.GetExplosionExposureRate(ExplosionPos, static_cast<float>(a_ExplosionSize));
+			float EntityExposure = a_Entity.GetExplosionExposureRate(ExplosionPos, static_cast<float>(a_ExplosionSize));
 
-				// Exposure reduced by armor
-				EntityExposure = EntityExposure * (1.0f - a_Entity.GetEnchantmentBlastKnockbackReduction());
+			// Exposure reduced by armor
+			EntityExposure = EntityExposure * (1.0f - a_Entity.GetEnchantmentBlastKnockbackReduction());
 
-				double Impact = (1 - ((Length / ExplosionSizeInt) / 2)) * EntityExposure;
+			auto Impact = std::pow(std::max(0.2, DistanceFromExplosion.Length()), -1);
+			Impact *= EntityExposure * ExplosionSizeInt / 5.0;
 
-				DistanceFromExplosion.Normalize();
-				DistanceFromExplosion *= Impact;
+			DistanceFromExplosion.Normalize();
+			DistanceFromExplosion *= Vector3d{Impact, 0.0, Impact};
+			DistanceFromExplosion.y += 0.3 * Impact;
 
-				a_Entity.SetSpeed(DistanceFromExplosion);
-			}
+			a_Entity.SetSpeed(DistanceFromExplosion);
 
 			return false;
 		}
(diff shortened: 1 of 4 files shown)


mathiascode:
Non-player knockback is still too weak (tnt cannon), and https://github.com/cuberite/cuberite/issues/4575 still occurs. Spawn a bunch of ghasts, let them shoot you, and you should eventually blast away into the sky.

src/ChunkMap.cpp
```diff
@@ -1313,21 +1313,19 @@ void cChunkMap::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_
...
 
-				DistanceFromExplosion.Normalize();
-				DistanceFromExplosion *= Impact;
+			DistanceFromExplosion.Normalize();
+			DistanceFromExplosion *= Vector3d{Impact, 0.0, Impact};
+			DistanceFromExplosion.y += 0.3 * Impact;
 
-				a_Entity.SetSpeed(DistanceFromExplosion);
-			}
+			a_Entity.SetSpeed(DistanceFromExplosion);
```
bearbin: This just seems wrong - how does setting the speed for all entities even those outside of the explosion radius _reduce_ the interference? This looks like it should set speed for all entities outside of the explosion radius to zero, which is not really a desired outcome.

peterbell10:
>Non-player knockback is still too weak (tnt cannon)

That's #4527, not #4575

>#4575 still occurs. Spawn a bunch of ghasts, let them shoot you, and you should eventually blast away into the sky.

I don't see the connection to #4575. If you're shot at by a ghast then you should be knocked back because you're within range.

src/ChunkMap.cpp
peterbell10: Player speeds aren't really updated properly so sending 0 is preferable to sending junk.

peterbell10: I could make this player only I suppose.

mathiascode:
While it isn't exactly vanilla-like, my PR brought back some knockback for non-player entities, to make TNT cannons possible.

Obviously there should be some knockback from ghast blasts, but the issue is that one of them sometimes sends you hundreds of blocks into the sky. It seems to happen especially when multiple blasts are shot at once.

mathiascode:
Seems to happen quite often when jumping right before the blast hits the ground.

peterbell10:
>While it isn't exactly vanilla-like, my PR brought back some knockback for non-player entities, to make TNT cannons possible.

Okay, I restored your factor of 6.

>Obviously there should be some knockback from ghast blasts, but the issue is that one of them sometimes sends you hundreds of blocks into the sky. It seems to happen especially when multiple blasts are shot at once.

Well completely disabling knockback seems like an over reaction. I've reduced the knockback strength for players so it should be okay.

mathiascode:
General throught/question: why do we have to send player motion in cClientHandle::SendExplosion? SetSpeed in the cChunkMap code should already take care of this for all entities in range, including players, and it means we would get proper knockback for everything (sharpnel, tnt etc).

peterbell10:
>why do we have to send player motion in cClientHandle::SendExplosion? 

Because the client will take the most recent information i.e. what is sent in `SendExplosion`.

>it means we would get proper knockback for everything (sharpnel, tnt etc).

The speed is set for all entities either way.

src/World.cpp
```diff
@@ -1417,17 +1417,20 @@ void cWorld::DoExplosionAt(double a_ExplosionSize, double a_BlockX, double a_Blo
 
 	BroadcastSoundEffect("entity.generic.explode", Vector3d(a_BlockX, a_BlockY, a_BlockZ), 4.0f, SoundPitchMultiplier * 0.7f);
 
+	Vector3d ExplosionPos(a_BlockX, a_BlockY, a_BlockZ);
 	{
 		cCSLock Lock(m_CSPlayers);
-		for (cPlayerList::iterator itr = m_Players.begin(); itr != m_Players.end(); ++itr)
+		for (auto Player : m_Players)
```
tigerw: is this a copy?

peterbell10: It's a pointer

PR #4591, opened 2020-03-31
Pulled the BlockID and BlockInfo headers from Globals.h.

madmaxoft:
We want to isolate the block types to as few places as possible.

Also renamed ID -> Type.


tigerw:
oof, 96 files

madmaxoft:
And still failing, don't review yet, some minor changes are still incoming.

tigerw:
Is there a summary of what's happening with E_BLOCK constants? We are still going to be using integers internally right?

madmaxoft:
Currently, yes. But we want to get rid of them eventually, and use string types, just like 1.13.

tigerw:
Think of the performance! Think of our original vision of low memory consumption!

madmaxoft:
Don't worry, the perf be that bad, and the memory footprint might even be a tiny bit smaller.

Wow, I managed not only to break a test, but even really really break it - now it segfaults >:[

tigerw:
How so?

madmaxoft:
Look at Travis

tigerw:
About the memory footprint being smaller though.
The test failure is surprising certainly.

madmaxoft:
Currently, we use 12 bits per block (8 bits BLOCKTYPE, 4 bits meta).
New system uses palettes and dynamic bits-per-block. Each 16x16x16 area will be stored as a separate `PalettedBlockArea` with its own local palette. It will know how many *different* blocks there are (through the palette entry count) and can therefore use fewer bits per block. For example if there are only 4 block types (air, grass, dirt, stone), it can use 2 bits per block, meaning 6 times less RAM! The assumption is that most of the 16x16x16 areas will have fewer than 12 bits per block (because 12 bits per block means 4096 different block types, which means each and every one of those 16x16x16 blocks is a different type). The increase in RAM usage due to local palettes is expected to be much smaller than the decrease in RAM usage thanks to fewer bits-per-block.

madmaxoft:
Okay, segfaults fixed (that was stupid), but now the generator generates different terrain, and I need to find out why - I didn't change anything in the generator, it must be some dependency on the block types

madmaxoft:
I have updated the chunk checksums in the tests. I carefully checked that the base terrain generated is the same, the differences were cause by the fact that the previous test version didn't have full `cBlockInfo` information, it only used synthesized dummy values, while this version now has full `cBlockInfo` as used by the main server. The different generation is caused by `cBlockInfo::m_CanBeTerraformed`, which was set to false for all blocks in the previous test version.

tigerw:
The tradeoff then is decoding this compacted format into something the rest of this server can use? And we're saying to use minecraft:something strings? Will I have to do a string comparison to check if blocks at different positions are the same?

madmaxoft:
This is now ready for review.

tigerw:
Sorry it's a little difficult to parse the diff; apart from moving definitions out of Global were there any other pertinent changes you wanted to draw attention to?

src/Items/ItemPickaxe.h
```diff
@@ -168,7 +168,7 @@ class cItemPickaxeHandler :
 
 	virtual float GetBlockBreakingStrength(BLOCKTYPE a_Block) override
 	{
-		if (!IsBlockMaterialIron(a_Block) && !IsBlockMaterialAnvil(a_Block) && !IsBlockMaterialRock(a_Block))
```
tigerw: What was wrong here?

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -363,10 +364,10 @@ bool cWSSAnvil::LoadChunkFromNBT(const cChunkCoords & a_Chunk, const cParsedNBT
 	cChunkDef::BlockNibbles BlockLight;
 	cChunkDef::BlockNibbles SkyLight;
 
-	memset(BlockTypes, E_BLOCK_AIR, sizeof(BlockTypes));
-	memset(MetaData,   0,           sizeof(MetaData));
-	memset(SkyLight,   0xff,        sizeof(SkyLight));  // By default, data not present in the NBT means air, which means full skylight
-	memset(BlockLight, 0x00,        sizeof(BlockLight));
+	memset(BlockTypes, 0,    sizeof(BlockTypes));
```
tigerw: You wish to eventually replace this with a palette right? At which point some sort of constant, string or otherwise, for air should be reintroduced.

tests/Generating/BasicGeneratorTest.cpp
```diff
@@ -245,6 +256,11 @@ static void checkChunkChecksums(
 	{
 		cChunkDesc chd(coords.mCoords);
 		aGenerator.Generate(chd);
+		/*
+		cFile f(Printf("Repeatability_%s-%02d-%02d.raw", aDimension, coords.mCoords.m_ChunkX, coords.mCoords.m_ChunkZ), cFile::fmWrite);
+		f.Write(chd.GetBlockTypes(), sizeof(chd.GetBlockTypes()));
+		f.Close();
```
tigerw: Is this intended to remain?

src/Items/ItemPickaxe.h
tigerw: (with IsBlockMaterialAnvil)

tests/Generating/BasicGeneratorTest.cpp
madmaxoft: Yes, it is needed for debugging whenever this test fails, so why not keep it in, although commented-out.

tigerw:
This looks good to me.

madmaxoft:
It seems neither Travis nor Appveyor received the GitHub notification to build, and I have no idea how to re-send them. Blind-merged :)

Issue #4596, opened 2020-04-01
Linux Detect TCP Flood with cuberite


Xenoxis:
Client version: 1.12.2
Server OS: Ubuntu server 16.04 i386 with Linux kernel 4.4.217
Cuberite Commit id: 

Playing with 12 players on the server, and linux `dmesg` tell this after usual check.
Don't know if this is normal

### Server log
```
[375852.589465] TCP: request_sock_TCP: Possible SYN flooding on port 8080. Sending cookies.  Check SNMP counters.
[376429.759274] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[376614.858572] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[377620.224287] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[377818.862788] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[380227.792813] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[380273.463743] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[380570.750592] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[388071.524754] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[435360.075542] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[436116.512951] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[436230.129591] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[436332.054242] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
[455095.518379] TCP: request_sock_TCP: Possible SYN flooding on port 25565. Sending cookies.  Check SNMP counters.
```

Issue #3175, opened 2016-05-01
Cuberite not working with vlan


Disinterpreter:
### Expected behavior

Cuberite must work with VLAN (because forge working)
### Actual behavior

Cuberite don't work with VLAN (i dunno why, ip doesn't work)
eth0      Link encap:Ethernet  HWaddr 08:00:**:**:**:**
          inet6 addr: 2a02:17d0:197:QQQ::QQ/128 Scope:Общий
          inet6 addr: fe80::a00:27ff:fec3:ba1e/64 Scope:Link
eth0.18   Link encap:Ethernet  HWaddr 08:00:QQQ:QQ:QQ:QQ
          inet addr:46.QQ.QQ.QQQ  Bcast:46.QQ.QQ.QQ Mask:255.255.255.224
          inet6 addr: fe80::a00:27ff:fec3:ba1e/64 Scope:Link
### Steps to reproduce the behavior
1. Create VLAN on router
2. Create VLAN on switch
3. Create VLAN on linux-machine (it's internet vlan)
4. Starting server
5. Try to connect

worktycho:
Whats the output of netstat? IPv6 might be causing issues.

Disinterpreter:
[](http://i.imgur.com/5KQvU54.png)

worktycho:
And are you trying to connect on ipv4? That looks like we are only listening on v6 for some reason.

Disinterpreter:
yes,  i want connect to ipv4, but external network ipv4 work only through vlan.

source83:
I suspect I have the same problem, the VM I've got at a hosting company might (or most possibly is) using vlan's.

source83:
"TCP: request_sock_TCP: Possible SYN flooding on port (my web interface port) Dropping request.  Check SNMP counters."
I get this every time I try to connect to the web interface or to the mc server itself.
What port I use, has so far not yielded any difference.

source83:
Ignore my previous messages, I found that it worked with sysctl -w net.ipv4.tcp_syncookies="1"

PR #4597, opened 2020-04-01
Let libevent choose a reasonable backlog value

### LINKED ISSUE #4457: Connections time out in Android 9+
Investigate why connections to the server (Webadmin included) time out on some Android 9 and Android 10 devices.

mathiascode:
References: http://www.wangafu.net/~nickm/libevent-book/Ref8_listener.html
https://github.com/libevent/libevent/blob/620a3fa1d514a0020c17a9a1119400dcdbd97d7a/include/event2/listener.h#L124

For some reason this caused issues on Android, but not on other OSes. I suppose this could possibly close other connection issue bug reports if such exist.

Fixes #4457
Fixes https://github.com/cuberite/android/issues/47
Fixes https://github.com/cuberite/cuberite/issues/4596

### DIFF
--- a/src/OSSupport/ServerHandleImpl.cpp
+++ b/src/OSSupport/ServerHandleImpl.cpp
@@ -205,14 +205,14 @@ bool cServerHandleImpl::Listen(UInt16 a_Port)
 		evutil_closesocket(MainSock);
 		return false;
 	}
-	if (listen(MainSock, 0) != 0)
+	if (listen(MainSock, SOMAXCONN) != 0)
 	{
 		m_ErrorCode = EVUTIL_SOCKET_ERROR();
 		Printf(m_ErrorMsg, "Cannot listen on port %d: %d (%s)", a_Port, m_ErrorCode, evutil_socket_error_to_string(m_ErrorCode));
 		evutil_closesocket(MainSock);
 		return false;
 	}
-	m_ConnListener = evconnlistener_new(cNetworkSingleton::Get().GetEventBase(), Callback, this, LEV_OPT_CLOSE_ON_FREE | LEV_OPT_REUSEABLE, -1, MainSock);
+	m_ConnListener = evconnlistener_new(cNetworkSingleton::Get().GetEventBase(), Callback, this, LEV_OPT_CLOSE_ON_FREE | LEV_OPT_REUSEABLE, 0, MainSock);
 	m_IsListening = true;
 
 	if (!NeedsTwoSockets)
@@ -263,7 +263,7 @@ bool cServerHandleImpl::Listen(UInt16 a_Port)
 		return true;  // Report as success, the primary socket is working
 	}
 
-	if (listen(SecondSock, 0) != 0)
+	if (listen(SecondSock, SOMAXCONN) != 0)
 	{
 		err = EVUTIL_SOCKET_ERROR();
 		LOGD("Cannot listen on secondary socket on port %d: %d (%s)", a_Port, err, evutil_socket_error_to_string(err));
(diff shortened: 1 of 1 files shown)


mathiascode:
Aren't we already calling listen() before evconnlistener_new?
"If backlog is negative, Libevent tries to pick a good value for the backlog; if it is zero, Libevent assumes that you have already called listen() on the socket you are providing it."

peterbell10:
Yes, of course... But that documentation is regarding the backlog parameter to `evconnlistener_new`, not to `listen`.

mathiascode:
Checked https://github.com/cuberite/cuberite/issues/4596. Since our backlog value is proper now, the SYN flood messages aren't appearing anymore.

mathiascode:
Should also fix the later comments in https://github.com/cuberite/cuberite/issues/3175 (and probably the original issue as well).

src/OSSupport/ServerHandleImpl.cpp
```diff
@@ -212,7 +212,7 @@ bool cServerHandleImpl::Listen(UInt16 a_Port)
 		evutil_closesocket(MainSock);
 		return false;
 	}
-	m_ConnListener = evconnlistener_new(cNetworkSingleton::Get().GetEventBase(), Callback, this, LEV_OPT_CLOSE_ON_FREE | LEV_OPT_REUSEABLE, 0, MainSock);
+	m_ConnListener = evconnlistener_new(cNetworkSingleton::Get().GetEventBase(), Callback, this, LEV_OPT_CLOSE_ON_FREE | LEV_OPT_REUSEABLE, -1, MainSock);
```
peterbell10, line 215: This calls `listen` twice.

madmaxoft:
You should change the second parameter, `0`, in the `listen` calls a few lines above each of your current changes. I have no idea why a zero is there, I can't find any documentation saying it's a special value and zero backlog makes little sense. Make it `SOMAXCONN` if all platforms accept that.

mathiascode:
I'm not sure exactly how to address @peterbell10's comment. Do we need the early listen calls? In my testing, if a port was occupied, Cuberite always seemed to bail out before this part of the code was reached.

madmaxoft:
Just change the second param to `listen` on line 208 instead of changing the param to `evconnlistener_new` (put the zero back there); similarly with the second occurrence.

mathiascode:
That was my original change, but peterbell wanted it changed. I'll apply your suggestion, any disagreements have to be worked out.

madmaxoft:
Wait, perhaps @peterbell10 could explain what he meant?

madmaxoft:
Okay, I got confused by your initial comment, probably same as @mathiascode , I thought you wanted the call to `evconnlistener_new` changed instead of the call to `listen`, just like Mathias did.

Anyway, now it seems reasonably; if it still fixes those issues (did you test, @mathiascode ?), I say merge.

peterbell10:
Letting `evconnlistener_new` call `listen` for you is an equally valid solution. But it has to be *instead of* calling `listen` ourself.

mathiascode:
@peterbell10 Thanks for the clarification. I misunderstood your original comment.

Tested on Android and Manjaro, works fine there. Can't test on Windows.

mathiascode:
Appveyor builds are now ready. If @madmaxoft or someone else with a Windows installation can test, I say merge.

madmaxoft:
Confirmed working.

Xenoxis:
This fix reduced a lot the CPU usage on my server (~60% before ->  ~30% now), thanks !

madmaxoft:
That is weird, I think the cause is somewhere else.

Xenoxis:
> 
> 
> That is weird, I think the cause is somewhere else.

I have a very old system for my MC Server, every little optimisations can be felt, I believe this is normal, don't worry

PR #4603, opened 2020-04-02
Remove golden apple after eating

mathiascode:
Golden apples weren't removed from the inventory if the player's health was maxed out.

### DIFF
--- a/src/Items/ItemFood.h
+++ b/src/Items/ItemFood.h
@@ -98,10 +98,7 @@ public:
 
 	virtual bool EatItem(cPlayer * a_Player, cItem * a_Item) override
 	{
-		if (!super::EatItem(a_Player, a_Item))
-		{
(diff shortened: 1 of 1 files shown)


bearbin:
Why can you ignore satiation before removing all other food items, as well as golden apples?

mathiascode:
Hm, satiation probably shouldn't be ignored in cItemFoodHandler. Should we just remove an item in cItemGoldenAppleHandler as well, or is there a better way? Add golden apple handling into cItemFoodHandler?

mathiascode:
Wait, there's the same issue for carrots and potatoes. Will have to look closer.

bearbin:
Add a flag to ItemFoodHandler to ignore satiation, which can be controlled by the subclasses? Or a new EatItemIgnoringSatiation?

Issue #4582, opened 2020-03-30
Server crash when entering the nether


Xenoxis:
Client version: 1.12
Server OS: Ubuntu 16.04
Cuberite Commit id: 6116f899dee33a6d7bd887f7a57ace17df2ab94f

### Expected behavior
Entering in the Nether and spawn at portail in the nether's side

### Actual behavior
Server Crash

### Steps to reproduce the behavior
Going into the nether

### Server log
```
[14:15:25]   D:    | Cuberite has encountered an error and needs to close
[14:15:25] Details | SIGSEGV: Segmentation fault
[14:15:25] Cuberite local build build id: Unknown
[14:15:25] from commit id: approx: 6116f899dee33a6d7bd887f7a57ace17df2ab94f built at: approx: 2020-03-29 21:58:19 +0100
[14:15:25] Kicking player <playername> for "Server shutdown"
./Cuberite(_Z15PrintStackTracev+0x22)[0x82a9c22]
./Cuberite[0x8296e18]
[0xb77b8bb4]
/lib/i386-linux-gnu/libc.so.6(cfree+0x26)[0xb73e8496]
/usr/lib/i386-linux-gnu/libstdc++.so.6(_ZdlPv+0x18)[0xb76428a8]
./Cuberite(_ZN7cEntity11MoveToWorldEP6cWorld7Vector3IdEbb+0xdf)[0x85032ef]
./Cuberite(_ZN20cNetherPortalScanner10OnDisabledEv+0x8c)[0x823a20c]
./Cuberite(_ZN9cChunkMap12DelChunkStayER10cChunkStay+0xac)[0x81e1c0c]
./Cuberite(_ZN10cChunkStay7DisableEv+0x18)[0x81ec168]
./Cuberite(_ZN9cChunkMap12SetChunkDataER13cSetChunkData+0xec)[0x81e1edc]
./Cuberite(_ZN6cWorld12SetChunkDataER13cSetChunkData+0x2c)[0x828366c]
./Cuberite(_ZN6cWorld4TickENSt6chrono8durationIxSt5ratioILx1ELx1000EEEES4_+0xbd)[0x828a9cd]
./Cuberite(_ZN6cWorld11cTickThread7ExecuteEv+0xba)[0x828b11a]
/usr/lib/i386-linux-gnu/libstdc++.so.6(+0xb4a7d)[0xb7671a7d]
/lib/i386-linux-gnu/libpthread.so.0(+0x6295)[0xb7534295]
[...]

PR #4606, opened 2020-04-03
Don't set entity world twice

mathiascode:
https://github.com/cuberite/cuberite/issues/4582#issuecomment-608401321


mathiascode:
Need to investigate further.

bearbin:
Is this still a WIP?

mathiascode:
Seems like all checks passed this time, so should be ready.

tigerw:
Wait, but AddEntity only queues the entity to be added later on in the tick thread, are we sure nothing depends on the "pending" world being set and available immediately?

tigerw:
I guess we are sure then.

tigerw:
Should be ok. The entity isn't (shouldn't) be touched/ticked before added to a world, so that moment when m_World is null or uninitialised when an entity has just been birthed cannot matter. Also DoMoveToWorld sets the destination pre-emptively, so no worries there.

PR #4610, opened 2020-04-03
Remove switch statements from cItemFoodHandler

bearbin:
Fixes issues with golden apples not being edible, and carrots etc not being removed after eating.


bearbin:
There is a bit of duplication in cItemFoodSeedsHandler, otherwise we would have to have multiple inheritance.

tigerw:
Looks good! :D

PR #4611, opened 2020-04-03
Quick fix to make spawn eggs work

mathiascode:
Mob type values changed in https://github.com/cuberite/cuberite/commit/01b8ed5295875262a91b60af878bf2a18c1b7aae, which broke spawn eggs. In the long run, we should really rework how spawn eggs are handled.


madmaxoft:
This is indeed a proper fix.

PR #4615, opened 2020-04-04
Animal spawn fixes

### LINKED ISSUE #2123: Refactor cMobSpawner::CanSpawnHere and cFinishGenPassiveMobs::TrySpawnAnimals
Discussion is in https://github.com/mc-server/MCServer/pull/2107

Unify duplicated logic, and make the methods shorter and more understandable.

### LINKED ISSUE #3705: Mobs spawns without triggering the hook spawning monster
Client version:   1.11.2
Server OS:        Windows
Commit id:         7c4576a

### Expected behavior
If I disable the spawning of mobs by returning true in hook `HOOK_SPAWNING_MONSTER`, no mobs should be spawned.

### Actual behavior
In chunk generator mobs will be added to the `cChunkDesc` and there won't be any call to the hook.
Source where mobs are created and added: [FinishGen.cpp#L1493](https://github.com/cuberite/cuberite/blob/7c4576a/src/Generating/FinishGen.cpp#L1493)

Use the hook above to disable spawning and you will see mobs around.

### LINKED ISSUE #3984: world.ini - turning off natural mob spawning is confusing
There are 3 different controls for animal spawning. The difference is not documented. I'd like to turn off the natural spawning of animals.

```
[Monsters]
...
AnimalsOn=1 (needs to be set to 0)
...

[Generator]
...
Finishers=...etc..., Animals (needs to be set to be removed)
...

[Animals]
...
AnimalSpawnChunkPercentage=10 (needs to be set to 0)
...

```

In general, world.ini is poorly documented.

mathiascode:
Fixes #3984 
Fixes #3705 
Fixes #2123

Animals would usually not spawn outside the chunk gen animal finisher, due to cChunk::SpawnMobs using the wrong coordinates to get the biome type later used in cMobSpawner::GetAllowedMobTypes.

Some animals also had a hard time spawning due to the number of grass blocks covered with foliage.

Since these spawning issues are now solved, I also removed the animal finisher, which mostly caused confusion. It was created with the assumption that animals should not spawn after the initial chunk generation, which is old Minecraft behavior.


src/MobSpawner.cpp
```diff
@@ -152,8 +152,8 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, Vector3i a_RelPos, eMonsterType
 		case mtSheep:
 		{
 			return (
-				(targetBlock == E_BLOCK_AIR) &&
-				(blockAbove == E_BLOCK_AIR) &&
+				(!cBlockInfo::IsTransparent(targetBlock)) &&
```
bearbin, line 155: I think we should also check for solidity here.

mathiascode:
This is not enough to completely fix the issue. Still investigating.

PR #4616, opened 2020-04-04
[WIP]Implement written books

Seadragon91:
- Added new class cBookContent that contains the infos about a book
- Loads and saves the book from player inventory and nbt
- Added hooks: HOOK_PLAYER_EDITING_BOOK and HOOK_PLAYER_EDITED_BOOK

Revived that old branch. Tested and it works in 1.8.9, 1.9.4 and 1.12.2.
Will check the API and look to extend it with `cJson`.

### DIFF
--- a/src/BookContent.h
+++ b/src/BookContent.h
@@ -1,0 +1,84 @@
+
+// BookContent.h
+
+#pragma once
+
+#include "WorldStorage/FastNBT.h"
+
+
+
+
+
+// tolua_begin
+class cBookContent
+{
+public:
+	/** Creates a empty book */
+	cBookContent():
+		m_IsSigned(false)
+	{}
+
+	/** Set the author of the book */
+	void SetAuthor(const AString & a_Author) { m_Author = a_Author; }
+
+	/** Returns the author of the book */
+	const AString & GetAuthor(void) const { return m_Author; }
+
+	/** Set the title of the book */
+	void SetTitle(const AString & a_Title) { m_Title = a_Title; }
+
+	/** Returns the title of the book */
+	const AString & GetTitle(void) const { return m_Title; }
+
+	/** Clears the whole book */
+	void Clear();
+
+	/** Returns true if the book has no author, no title and no pages */
+	bool IsEmpty(void) const;
+
+	// tolua_end
+
+	/** Required in ManualBindings to save the page as simple string or json string */
+	void SetIsSigned(bool a_IsSigned) { m_IsSigned = a_IsSigned; }
+
+	/** Returns true if the book is signed */
+	bool IsSigned(void) const { return m_IsSigned; }
+
+	/** Add a page to the end of the book */
+	void AddPage(const AString & a_Page) { m_Pages.emplace_back(a_Page); }
+
+	/** Returns the page at the index */
+	const AString & GetPage(size_t a_Index) { return m_Pages[a_Index]; }
+
+	/** Changes the page at the index */
+	void SetPage(size_t a_Index, const AString & a_Page) { m_Pages[a_Index] = a_Page; }
+
+	/** Removes all pages */
+	void ClearPages(void)
+	{
+		m_Pages.clear();
+	}
+
+	/** Returns a ref to the vector. Used in ManualBindings and for saving the book */
+	const std::vector<AString> & GetPages(void) const { return m_Pages; }
+
+	/** Read the book content from nbt. The boolean a_SaveAsJson is optional. If the book is signed, the text should be in a json string */
+	static void ParseFromNBT(int TagTag, cBookContent & a_BookContent, const cParsedNBT & a_NBT, bool a_SaveAsJson = false);
+
+	/** Write book content to nbt */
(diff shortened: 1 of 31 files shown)


Seadragon91:
Had thought about adding a static function to deserialize from json string, but I think the API functions are already enough.

src/BookContent.h
```diff
@@ -0,0 +1,84 @@
...
+
+	/** Clears the whole book */
+	void Clear();
+
+	/** Returns true if the book has no author, no title and no pages */
+	bool IsEmpty(void) const;
+
+	// tolua_end
+
+	/** Required in ManualBindings to save the page as simple string or json string */
```
bearbin, line 41: I can't make any sense out of this comment, could you explain a bit more?

src/Protocol/Protocol_1_9.cpp
```diff
@@ -3103,6 +3114,64 @@ void cProtocol_1_9_0::HandleVanillaPluginMessage(cByteBuffer & a_ByteBuffer, con
...
+		}
+		BookItem.m_BookContent.SetIsSigned(IsSigned);
+
+		// The equipped item contains the old book content
+		if (cRoot::Get()->GetPluginManager()->CallHookPlayerEditingBook(Player, Player.GetEquippedItem().m_BookContent, BookItem.m_BookContent, IsSigned))
+		{
+			// Plugin denied the player to edit the book
+			cInventory & inv = Player.GetInventory();
+			inv.SetHotbarSlot(inv.GetEquippedSlotNum(), BookItem);
+			Player.GetInventory().SendEquippedSlot();
```
bearbin, line 3163: The plugin denied editing the book but we set it anyway? This looks wrong.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -3103,6 +3114,64 @@ void cProtocol_1_9_0::HandleVanillaPluginMessage(cByteBuffer & a_ByteBuffer, con
...
+			return;
+		}
+
+		cInventory & inv = Player.GetInventory();
+		inv.SetHotbarSlot(inv.GetEquippedSlotNum(), BookItem);
+		Player.GetInventory().SendEquippedSlot();
```
bearbin, line 3169:
```suggestion
		Player.GetInventory().SetEquippedItem(BookItem);
```

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2969,6 +2980,65 @@ void cProtocol_1_8_0::HandleVanillaPluginMessage(cByteBuffer & a_ByteBuffer, con
...
+			// Plugin denied the player to edit the book
+			cInventory & inv = Player.GetInventory();
+			inv.SetHotbarSlot(inv.GetEquippedSlotNum(), BookItem);
+			Player.GetInventory().SendEquippedSlot();
+			return;
+		}
+
+		cInventory & inv = Player.GetInventory();
+		inv.SetHotbarSlot(inv.GetEquippedSlotNum(), BookItem);
+		Player.GetInventory().SendEquippedSlot();
```
bearbin, line 3036:
```suggestion
		Player.GetInventory().SetEquippedItem(BookItem);
```

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2969,6 +2980,65 @@ void cProtocol_1_8_0::HandleVanillaPluginMessage(cByteBuffer & a_ByteBuffer, con
...
+			IsSigned = false;
+			BookItem = cItem(E_ITEM_BOOK_AND_QUILL);
+			cBookContent::ParseFromNBT(0, BookItem.m_BookContent, NBT);
+		}
+		BookItem.m_BookContent.SetIsSigned(IsSigned);
+
+		// The equipped item contains the old book content
+		if (cRoot::Get()->GetPluginManager()->CallHookPlayerEditingBook(Player, Player.GetEquippedItem().m_BookContent, BookItem.m_BookContent, IsSigned))
+		{
+			// Plugin denied the player to edit the book
```
bearbin, line 3027: Again, plugin denied editing but we do it anyway?

Seadragon91: Fixed, thanks.

src/Protocol/Protocol_1_8.cpp
Seadragon91: Fixed, thanks.

src/BookContent.h
Seadragon91: Changed the comment. In a written book the content is a json string and in book and quill a simple string.

Seadragon91:
I tested against 1.8 and there written book are stored as simple strings and not as json strings.

Seadragon91:
Okay, it' sent as a json string in 1.8, but the text is escaped:
1.9:
```json
{
   "text" : "1\n2\n3\n4\n5"
}
```

1.8:
```json
{
   "text" : "\"1\\n2\\n3\\n4\\n5\""
}
```
Should we change the content, before it's sent? Or ignore it as it's only in 1.8?

bearbin:
When it's received seems to make more sense. Since we do support 1.8 still, we should make all features that can work in that version work.

PR #4620, opened 2020-04-04
Falling blocks can now be spawned at any position.

madmaxoft:
`cFallingBlock` entity can be created at `Vector3d` coords, instead of `Vector3i`.
Added a `/cake` command to Debuggers that throws a cake in a nice slow arc.

### DIFF
--- a/src/Simulator/SandSimulator.cpp
+++ b/src/Simulator/SandSimulator.cpp
@@ -62,7 +62,7 @@ void cSandSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
 			);
 			*/
 
-			auto FallingBlock = cpp14::make_unique<cFallingBlock>(Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
+			auto FallingBlock = cpp14::make_unique<cFallingBlock>(Vector3d(0.5, 0, 0.5) + Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
 			auto FallingBlockPtr = FallingBlock.get();
 			if (!FallingBlockPtr->Initialize(std::move(FallingBlock), m_World))
 			{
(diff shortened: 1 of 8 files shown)


madmaxoft:
That's the problem with Lua API, it actually distinguishes rather hard even between a `Vector3d` and `Vector3f`. It will call the function relevant to the argument type.

True, the docs need updating, will do that.

src/Simulator/SandSimulator.cpp
```diff
@@ -62,7 +62,7 @@ void cSandSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
 			);
 			*/
 
-			auto FallingBlock = cpp14::make_unique<cFallingBlock>(Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
+			auto FallingBlock = cpp14::make_unique<cFallingBlock>(Vector3d(0.5, 0, 0.5) + Pos, BlockType, a_Chunk->GetMeta(itr->x, itr->y, itr->z));
```
tigerw, line 65: Mightn't it be slightly better to let cFallingBlock handle the specifics of adding 0.5 here? Also now World.h sees some duplication when it does the same offset.

madmaxoft: That's true, thanks for the idea. I really shouldn't code so late at night :)

madmaxoft: Wait, no, it won't work, I'd have to duplicate the body of `cWorld:SpawnFallingBlock` in order to do that properly. I think I'll keep it as-is.

tigerw: Template the first parameter of SpawnFallingBlock, you know you want to...

Or at least get this line to call m_World->SpawnFallingBlock, that should work right

bearbin: This should just call into m_World->SpawnFallingBlock

madmaxoft: Tested, replacement works.

madmaxoft: Yes, the original code didn't remove the sand block if creating the entity failed, but I decided against that behavior.

madmaxoft:
Appveyor acting up again, it builds on Windows.

PR #4622, opened 2020-04-05
Initial resource pack support

mathiascode:
The bare minimum for supporting server resource packs. Does not include any Lua APIs.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -425,6 +425,12 @@ void cClientHandle::FinishAuthenticate(const AString & a_Name, const cUUID & a_U
 		cRoot::Get()->BroadcastPlayerListsAddPlayer(*m_Player);
 		cRoot::Get()->SendPlayerLists(m_Player);
 
+		// Send resource pack
+		if (!cRoot::Get()->GetServer()->GetResourcePackUrl().empty())
+		{
+			m_Protocol->SendResourcePack(cRoot::Get()->GetServer()->GetResourcePackUrl());
+		}
+
 		SetState(csAuthenticated);
 	}
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2237,7 +2237,7 @@ void cClientHandle::ServerTick(float a_Dt)
 		cCSLock lock(m_CSState);
 		if (
 			(m_State == csAuthenticated) &&
-			(cRoot::Get()->GetServer()->GetResourcePackUrl().empty() || (m_ResourcePackStatus != -1))
+			(cRoot::Get()->GetServer()->GetResourcePackUrl().empty() || (m_ResourcePackStatus != rpWaiting))
 		)
 		{
 			StreamNextChunk();
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -428,7 +428,7 @@ void cClientHandle::FinishAuthenticate(const AString & a_Name, const cUUID & a_U
 		// Send resource pack
 		if (!cRoot::Get()->GetServer()->GetResourcePackUrl().empty())
 		{
-			m_Protocol->SendResourcePack(cRoot::Get()->GetServer()->GetResourcePackUrl());
+			SendResourcePack(cRoot::Get()->GetServer()->GetResourcePackUrl());
 		}
 
 		SetState(csAuthenticated);
@@ -2234,7 +2234,10 @@ void cClientHandle::ServerTick(float a_Dt)
 
 	{
 		cCSLock lock(m_CSState);
-		if (m_State == csAuthenticated)
+		if (
+			(m_State == csAuthenticated) &&
+			(m_ResourcePackStatus != -1)
+		)
 		{
 			StreamNextChunk();
 
@@ -2244,6 +2247,7 @@ void cClientHandle::ServerTick(float a_Dt)
 			// Add the player to the world (start ticking from there):
 			m_State = csDownloadingWorld;
 			m_Player->Initialize(std::move(m_PlayerPtr), *(m_Player->GetWorld()));
+
 			return;
 		}
 	}  // lock(m_CSState)
@@ -2929,6 +2933,15 @@ void cClientHandle::SendExperience(void)
 void cClientHandle::SendExperienceOrb(const cExpOrb & a_ExpOrb)
 {
 	m_Protocol->SendExperienceOrb(a_ExpOrb);
+}
+
+
+
+
+
+void cClientHandle::SendResourcePack(const AString & a_ResourcePackUrl)
+{
+	m_Protocol->SendResourcePack(a_ResourcePackUrl);
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2247,7 +2247,6 @@ void cClientHandle::ServerTick(float a_Dt)
 			// Add the player to the world (start ticking from there):
 			m_State = csDownloadingWorld;
 			m_Player->Initialize(std::move(m_PlayerPtr), *(m_Player->GetWorld()));
-
 			return;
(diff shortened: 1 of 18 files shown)


src/ClientHandle.h
```diff
@@ -424,6 +428,11 @@ class cClientHandle  // tolua_export
 	AString m_Password;
 	Json::Value m_Properties;
 
+	/** The resource pack status sent from the client
+	Possible values are; 0: successfully loaded, 1: declined, 2: failed download, 3: accepted
+	Reference: https://wiki.vg/Protocol#Resource_Pack_Status */
+	int m_ResourcePackStatus = -1;
+
```
madmaxoft, line 435: Make this an enum

src/Protocol/Protocol_1_9.cpp
```diff
@@ -1233,6 +1233,28 @@ void cProtocol_1_9_0::SendResetTitle(void)
...
+	cPacketizer Pkt(*this, pktResourcePack);
+
+	cSha1Checksum Checksum;
+	Checksum.Update(reinterpret_cast<const Byte *>(a_ResourcePackUrl.c_str()), a_ResourcePackUrl.size());
+	Byte Digest[20];
+	Checksum.Finalize(Digest);
+	AString Sha1Output;
+	for (size_t i = 0; i < ARRAYCOUNT(Digest); i++)
+	{
+		AppendPrintf(Sha1Output, "", Digest[i]);
```
madmaxoft, line 1247: The format string seems erroneous here.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1147,6 +1147,28 @@ void cProtocol_1_8_0::SendResetTitle(void)
...
+	cPacketizer Pkt(*this, pktResourcePack);
+
+	cSha1Checksum Checksum;
+	Checksum.Update(reinterpret_cast<const Byte *>(a_ResourcePackUrl.c_str()), a_ResourcePackUrl.size());
+	Byte Digest[20];
+	Checksum.Finalize(Digest);
+	AString Sha1Output;
+	for (size_t i = 0; i < ARRAYCOUNT(Digest); i++)
+	{
+		AppendPrintf(Sha1Output, "", Digest[i]);
```
madmaxoft, line 1161: The format string seems erroneous here.

src/Protocol/Protocol_1_9.cpp
```diff
@@ -1233,6 +1233,28 @@ void cProtocol_1_9_0::SendResetTitle(void)
...
+	}
```
madmaxoft, line 1248: There's a `cSha1Checksum::DigestToJava` function, if that is not usable here, perhaps add a new one there, `DigestToHex`.

src/ClientHandle.cpp
```diff
@@ -2228,7 +2235,10 @@ void cClientHandle::ServerTick(float a_Dt)
 
 	{
 		cCSLock lock(m_CSState);
-		if (m_State == csAuthenticated)
+		if (
+			(m_State == csAuthenticated) &&
+			(cRoot::Get()->GetServer()->GetResourcePackUrl().empty() || (m_ResourcePackStatus != -1))
+		)
```
madmaxoft, line 2241: So when does the client start streaming chunks if resource pack url is not empty?

mathiascode: I'm starting to realize that this needs a different approach. Not exactly sure how to solve this, but as soon as chunks are streamed, the resource pack prompt closes.

mathiascode:
What do you mean by multiple resource packs? What would the use case be? I don't think that's possible in the protocol.

src/ClientHandle.cpp
mathiascode: Fixed the issue.

src/mbedTLS++/Sha1Checksum.h
```diff
@@ -32,6 +32,9 @@ class cSha1Checksum
 	/** Returns true if the object is accepts more input data, false if Finalize()-d (need to Restart()) */
 	bool DoesAcceptInput(void) const { return m_DoesAcceptInput; }
 
+	/** Converts a SHA1 digest into hex */
+	static void DigestToHex(const Checksum & a_Digest, AString & a_JavaOut);
```
bearbin, line 36: `a_JavaOut`?

mathiascode: Delicious copypasta.

madmaxoft:
I have no idea what resourcepacks are. But the protocol basically allows to send multiple, so it's only a natural question to ask. If it makes no sense, then of course ignore me.

madmaxoft:
And you should really install the CheckBasicStyle commit hook locally :)

src/mbedTLS++/Sha1Checksum.cpp
```diff
@@ -85,6 +85,23 @@ void cSha1Checksum::Finalize(cSha1Checksum::Checksum & a_Output)
 
 
 
+void cSha1Checksum::DigestToHex(const Checksum & a_Digest, AString & a_Out)
+{
+	Checksum Digest;
+	memcpy(Digest, a_Digest, sizeof(Digest));
```
madmaxoft, line 91: Why the copy? It should be possible to read out of `a_Digest` directly

mathiascode:
Can that be automated somehow? I create too many local repos on a regular basis.

madmaxoft:
I just have this in `$/.git/hooks/precommit`:
```
#!/bin/sh

# Redirect output to stderr.
exec 1>&2

exec lua.exe src/CheckBasicStyle.lua -g 
```

I assume you could just copy that file around to the new repositories. And why so many repos anyway?

mathiascode:
Just working on a lot of things at once. I should spend some time on actually tidying up my dev environment though, it has become quite horrendous.

tigerw:
> Just working on a lot of things at once

CPU joke lel

mathiascode:
AMD gang.

madmaxoft:
I think it'd be better to actually try compiling the stuff you're committing, that way you don't go from one thing to another again and again and again because the CI discovers simple errors in the code.

mathiascode:
Well, it compiles under gcc. I should switch to clang.

Bond-009:
> Is it possible to send multiple resource packs in the protocol? Should we allow that?

You can set multiple resource packs in the client, since a resource pack doesn't have to set textures for all blocks you can have it overriding only a couple of textures (e.g. redstone components or glass)

Pokechu22:
While you can choose multiple packs clientside, I believe the server can only specify a single pack and it replaces the pack it's using if a new one is specified.  But I haven't actually tested that.

mathiascode:
That's true, I forgot that the client allows you to combine multiple resource packs. From what I've read, it doesn't seem like you can add multiple resource packs to a server.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1147,6 +1147,25 @@ void cProtocol_1_8_0::SendResetTitle(void)
 
 
 
+void cProtocol_1_8_0::SendResourcePack(const AString & a_ResourcePackUrl)
+{
+	cPacketizer Pkt(*this, pktResourcePack);
+
+	cSha1Checksum Checksum;
+	Checksum.Update(reinterpret_cast<const Byte *>(a_ResourcePackUrl.c_str()), a_ResourcePackUrl.size());
```
Pokechu22, line 1155: I'm pretty sure this is incorrect; it's supposed to be a checksum of the resource pack's contents (which vanilla has you specify in server.properties), not the URL.  I think the client uses it for caching, but I'm not 100% sure.

mathiascode: Seems like you're right. At least it's not an urgent issue, but would be good for caching indeed.

PR #4623, opened 2020-04-05
Only store IDs across ticks

tigerw:
Ironic, we could advise plugins to do this, but couldn't do the same ourselves.

@mathiascode and @peterbell10 do you know why there is a lock + atomics in MoveToWorld? I was under the impression that it should only be called from the tick thread during tick.

Tentative fix for #4582.

### DIFF
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -1407,7 +1407,7 @@ bool cEntity::DetectPortal()
 					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedOverworldName());
 					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
 					LOGD("Jumping %s -> %s", DimensionToString(dimNether).c_str(), DimensionToString(DestionationDim).c_str());
-					new cNetherPortalScanner(*this, TargetWorld, TargetPos, cChunkDef::Height);
+					new cNetherPortalScanner(*this, *TargetWorld, TargetPos, cChunkDef::Height);
 					return true;
 				}
 				// Nether portal in the overworld
@@ -1439,7 +1439,7 @@ bool cEntity::DetectPortal()
 					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedNetherWorldName());
 					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
 					LOGD("Jumping %s -> %s", DimensionToString(dimOverworld).c_str(), DimensionToString(DestionationDim).c_str());
-					new cNetherPortalScanner(*this, TargetWorld, TargetPos, (cChunkDef::Height / 2));
+					new cNetherPortalScanner(*this, *TargetWorld, TargetPos, (cChunkDef::Height / 2));
 					return true;
 				}
 			}
@@ -1487,7 +1487,7 @@ bool cEntity::DetectPortal()
 					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedOverworldName());
 					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
 					LOGD("Jumping %s -> %s", DimensionToString(dimEnd).c_str(), DimensionToString(DestionationDim).c_str());
-					return MoveToWorld(TargetWorld, false);
+					return MoveToWorld(*TargetWorld, false);
 				}
 				// End portal in the overworld
 				else
@@ -1513,7 +1513,7 @@ bool cEntity::DetectPortal()
 					cWorld * TargetWorld = cRoot::Get()->GetWorld(GetWorld()->GetLinkedEndWorldName());
 					ASSERT(TargetWorld != nullptr);  // The linkage checker should have prevented this at startup. See cWorld::start()
 					LOGD("Jumping %s -> %s", DimensionToString(dimOverworld).c_str(), DimensionToString(DestionationDim).c_str());
-					return MoveToWorld(TargetWorld, false);
+					return MoveToWorld(*TargetWorld, false);
 				}
 
 			}
@@ -1577,25 +1577,27 @@ void cEntity::DoMoveToWorld(const sWorldChangeInfo & a_WorldChangeInfo)
 
 
 
-bool cEntity::MoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown, bool a_ShouldSendRespawn)
+bool cEntity::MoveToWorld(cWorld & a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown, bool a_ShouldSendRespawn)
 {
-	ASSERT(a_World != nullptr);
-
 	// Ask the plugins if the entity is allowed to change world
-	if (cRoot::Get()->GetPluginManager()->CallHookEntityChangingWorld(*this, *a_World))
+	if (cRoot::Get()->GetPluginManager()->CallHookEntityChangingWorld(*this, a_World))
 	{
 		// A Plugin isn't allowing the entity to change world
 		return false;
 	}
 
-	if (m_WorldChangeInfo.m_NewWorld != nullptr)
-	{
-		// Avoid scheduling multiple warp tasks
-		return true;
-	}
+	const auto OldWorld = m_WorldChangeInfo.m_NewWorld;
 
 	// Create new world change info
-	m_WorldChangeInfo = { a_World,  a_NewPosition, a_SetPortalCooldown, a_ShouldSendRespawn };
+	// (The last warp command always takes precedence)
+	m_WorldChangeInfo = { &a_World,  a_NewPosition, a_SetPortalCooldown, a_ShouldSendRespawn };
+
+	if (OldWorld != nullptr)
+	{
+		// Avoid scheduling multiple warp tasks
+		// Only move ahead if we came from a "not warping" state
+		return true;
+	}
 
 	// TODO: move to capture when C++14
 	const auto EntityID = GetUniqueID();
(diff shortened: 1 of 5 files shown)


src/NetherPortalScanner.cpp
```diff
@@ -296,8 +297,18 @@ void cNetherPortalScanner::OnDisabled(void)
 		Position.z += OutOffset;
 	}
 
-	FLOGD("Placing player at {0}", Position);
-	m_Entity->MoveToWorld(m_World, Position, true, false);
+	// Lookup our warping entity by ID
+	// Necessary as they may have been destroyed in the meantime (#4582)
+	m_SourceWorld.DoWithEntityByID(
+		m_EntityID,
+		[this, &Position](auto & a_Entity)
```
peterbell10, line 304: That's a c++14 feature
```suggestion
		[this, &Position](cEntity & a_Entity)
```

src/Entities/Entity.cpp
```diff
@@ -1596,19 +1596,42 @@ bool cEntity::MoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPo
...
+		[EntityID = GetUniqueID()](cWorld & a_CurWorld)
+		{
+			a_CurWorld.DoWithEntityByID(
+				EntityID,
+				[](auto & a_Entity)
+				{
+					auto WCI = a_Entity.m_WorldChangeInfo.exchange(nullptr);
+					// TODO: halp
+					// cWorld::cLock Lock(a_CurWorld);
+					a_Entity.DoMoveToWorld(*WCI);
```
peterbell10, line 1628: Queued tasks aren't executed under the chunk map CS which is why `cWorld::cLock` was needed. You should be covered by `DoWithEntityByID` now though.

tigerw:
I've made sWorldChangeInfo a plain member, the rationale being that entites should not be manipulated outside of the tick thread, and MoveToWorld assumes the entity no longer ticks during warp, and so can no longer receive commands to move worlds.

src/Entities/Entity.cpp
```diff
@@ -1589,25 +1590,46 @@ bool cEntity::MoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPo
 		return false;
 	}
 
+	if (m_WorldChangeInfo.m_NewWorld != nullptr)
+	{
+		// Avoid scheduling multiple warp tasks
+		return true;
```
peterbell10, line 1596: The whole point of writing it this way was to take the _latest_ call, not the first call.

src/Entities/Entity.h
```diff
@@ -662,8 +662,8 @@ class cEntity
 
 	cWorld * m_World;
 
-	/** If not nullptr, a world change is scheduled and a task is queued in the current world. */
-	cAtomicUniquePtr<sWorldChangeInfo> m_WorldChangeInfo;
+	/** If field m_NewWorld not nullptr, a world change is scheduled and a task is queued in the current world. */
+	sWorldChangeInfo m_WorldChangeInfo;
```
peterbell10, line 666: >I've made sWorldChangeInfo a plain member, the rationale being that entites should not be manipulated outside of the tick thread

That assumption is false. Read the comment I linked above.

tigerw: If it really enables plugins to call into an entity asynchronous to world tick then MoveToWorld is the least of our problems.

Could you explain where the chunkmap's lock comes into play here?

Seadragon91:
Got this crash during a test (a_WorldChangeInfo.m_NewWorld was nullptr.):
```
Cuberite.exe!cPlayer::DoMoveToWorld(const cEntity::sWorldChangeInfo & a_WorldChangeInfo) Line 2091	C++
[Inline Frame] Cuberite.exe!cEntity::MoveToWorld::__l2::void <lambda>(cWorld &)::__l2::<lambda>(cEntity &) Line 1624	C++
Cuberite.exe!cFunctionRef<bool __cdecl(cEntity &)>::ObjectFunctionCaller<bool <lambda>(cEntity &)>(void * a_Callable, cEntity & <a_Args_0>) Line 71	C++
[Inline Frame] Cuberite.exe!cFunctionRef<bool __cdecl(cEntity &)>::operator()(cEntity &) Line 57	C++
[Inline Frame] Cuberite.exe!cChunk::DoWithEntityByID(unsigned int) Line 1934	C++
[Inline Frame] Cuberite.exe!cChunkMap::DoWithEntityByID(unsigned int) Line 1350	C++
Cuberite.exe!cWorld::DoWithEntityByID(unsigned int a_UniqueID, cFunctionRef<bool __cdecl(cEntity &)> a_Callback) Line 2809	C++
[External Code]	
[Inline Frame] Cuberite.exe!std::_Func_class<void,cWorld &>::operator()(cWorld &) Line 976	C++
Cuberite.exe!cWorld::TickQueuedTasks() Line 1229	C++
Cuberite.exe!cWorld::Tick(std::chrono::duration<__int64,std::ratio<1,1000>> a_Dt, std::chrono::duration<__int64,std::ratio<1,1000>> a_LastTickDurationMSec) Line 1050	C++
Cuberite.exe!cWorld::cTickThread::Execute() Line 124	C++
[External Code]	
[Inline Frame] Cuberite.exe!invoke_thread_procedure(unsigned int(*)(void *)) Line 91	C++
Cuberite.exe!thread_start<unsigned int (__cdecl*)(void * __ptr64)>(void * const parameter) Line 115	C++
[External Code]	
```

Occurred here:
https://github.com/cuberite/cuberite/blob/0d722701861ad9acfa7d36c5156fd184e3ba6bee/src/Entities/Player.cpp#L2091

src/Entities/Entity.cpp
tigerw: Thanks, oops :P

src/Entities/Entity.h
tigerw: notice me peterbell senpai

tigerw:
I guess it's move fast and break things time! 😇

src/Entities/Entity.h
peterbell10: If you made sure the chunk map CS was held during player chat handlers then this would be okay.

tigerw: Player chat is processed in the context of the world tick thread, so this is fine. Did you find any other cases during your work on world travel?

peterbell10: I can't remember any changes touching that code so I doubt it's been fixed.

src/Entities/Entity.cpp
```diff
@@ -1576,47 +1577,70 @@ void cEntity::DoMoveToWorld(const sWorldChangeInfo & a_WorldChangeInfo)
 
 
 
-bool cEntity::MoveToWorld(cWorld * a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown, bool a_ShouldSendRespawn)
+bool cEntity::MoveToWorld(cWorld & a_World, Vector3d a_NewPosition, bool a_SetPortalCooldown, bool a_ShouldSendRespawn)
 {
-	ASSERT(a_World != nullptr);
-
 	// Ask the plugins if the entity is allowed to change world
```
peterbell10, line 1582:
```suggestion
	ASSERT(m_World.GetChunkMap().GetCS().IsLockedByCurrentThread());
	
	// Ask the plugins if the entity is allowed to change world
```
Try running the `/portal` command with this.

src/NetherPortalScanner.h
```diff
@@ -48,11 +48,14 @@ class cNetherPortalScanner : public cChunkStay
...
-	cEntity * m_Entity;
+	/** The ID of the entity that's being moved. */
+	UInt32 m_EntityID;
+
+	/** The world we're moving the entity from, used to query the entity ID. */
+	cWorld & m_SourceWorld;
 
 	/** The world we're moving the entity to. */
-	cWorld * m_World;
+	cWorld & m_World;
```
peterbell10, line 58: This may be academic but I've heard that having references as members can lead to undefined behaviour. As a general rule, I try to avoid them. Using references for the constructor is good though.

A more practical point. Should we worry about `m_SourceWorld` becoming invalidated when the server shuts down?

src/Entities/Entity.cpp
tigerw: It is not locked.

I'm not convinced it has to be though: apart from the thread-safe QueueTask, MoveToWorld only accesses entity variables, and entities can only be accessed from within the tick thread.

src/NetherPortalScanner.h
tigerw: Interesting! I changed them here because indeed the portal scanner is inseparably bound to a world so I thought it would make this dependency clearer.

I'm fairly certain shutting down is a whole can of worms in itself, but given that the scanner is driven by the worlds' tick threads it shouldn't run code when the world ends.

peterbell10: You're referencing the _other_ world. There could be some tick where one world is running but the other is already shutdown.

src/Entities/Entity.cpp
peterbell10: >entities can only be accessed from within the tick thread.

That's simply not true. `cRoot:GetWorld` means any plugin can access the entities, no matter what thread they are on.

tigerw: Though this is incorrect behaviour if the plugin does so, just like if they store an entity reference across ticks and crash the server when it goes stale. I'm certain there are multiple places where we assume world state is only updated during a tick, and anything that violates this needs to be fixed. I suppose what I'm getting at is that there's no point in making MoveToWorld safe if every other cEntity-related function isn't.

src/NetherPortalScanner.h
tigerw: The order should be, all worlds are created, all tick threads start, (server running), (server asked to stop), all tick threads stop, all worlds deleted.

This should preclude any chance of one world still ticking trying to access a dead world.

madmaxoft: And then you get someone asking "But I want to create a new world while the server is running", and once you have that, "But I want to remove a world while the server is running" :P

tigerw: darn users always asking for stuff

Anyway I'm only describing how it is right now.

peterbell10: >all tick threads stop, all worlds deleted

Okay, I verified that's true and there is synchronisation before deleting. So we should be good.

src/Entities/Entity.cpp
peterbell10: >there's no point in making MoveToWorld safe if every other cEntity-related function isn't.

True, the main issue with `MoveToWorld` should be fixed by #4647

tigerw:
I will merge, if everything crashes and burns I'll be held responsible XD

PR #4626, opened 2020-04-07
Add Block Handlers for new blocks

### LINKED ISSUE #4625: Blocks ignore "rotation" etc.
Client version: 1.12.2
Server OS: Linux (Arch)
Cuberite Commit id: 16b3de94c89edae2ba8ff7ad6baa6b322c4571af

### Expected behavior
Placing a observer or terracotta block for example should result in the block having a direction thats facing, depending on the place it was placed from. (Bad explanation, I'm sorry)

### Actual behavior
The block(s) turn all the same way, so they ignore where they are placed from. So the observer faces "down" all the time and the terracotta "south". When placed, they have the correct facing direction for a very short time - might be client side - but when updated by the server, they get turned.

Try to place a observer or terracotta block facing in two different directions.

### Server log
```
-
```

bearbin:
Add missing (simple) blocks from pre-1.13 versions of minecraft.

DONE:

 * Glazed Terracotta
 * Observers
 * Refactor rotating blocks with better solution

FIXES:

Fixes #4625


tigerw:
this is really good

bearbin:
Observers still don't actually _do_ anything, but this is now ready for review.

src/Blocks/Mixins.h
```diff
@@ -164,3 +165,132 @@ class cMetaRotator:
 		return a_Meta;
 	}
 };
+
+
+/** Mixin for blocks whose meta on placement depends on the yaw of the player placing the block. BitMask
+selects the direction bits from the block's meta values. */
+template <class Base, NIBBLETYPE BitMask = 0x7, NIBBLETYPE North = 0x2, NIBBLETYPE East = 0x5, NIBBLETYPE South = 0x3, NIBBLETYPE West = 0x4, bool AssertIfNotMatched = false>
+class cYawRotator:
```
madmaxoft, line 173: Style:
- 5 empty lines between class declarations
- Use 2 digits for hex constants
- The line is quite long, split into separate lines
```suggestion





/** Mixin for blocks whose meta on placement depends on the yaw of the player placing the block.
BitMask selects the direction bits from the block's meta values. */
template <
	class Base,
	NIBBLETYPE BitMask = 0x07,
	NIBBLETYPE North = 0x02,
	NIBBLETYPE East = 0x05,
	NIBBLETYPE South = 0x03,
	NIBBLETYPE West = 0x04,
	bool AssertIfNotMatched = false
>
class cYawRotator:
```

src/Blocks/Mixins.h
```diff
@@ -164,3 +165,132 @@ class cMetaRotator:
...
+
+		a_BlockMeta = (BaseMeta & ~BitMask) | YawToMetaData(a_Player.GetYaw());
+		return true;
+	}
+
+
+
+
+
+	static NIBBLETYPE YawToMetaData(double a_Rotation)
```
madmaxoft, line 205: Add doxy-comments for all non-`override` functions.

src/Blocks/BlockGlazedTerracotta.h
```diff
@@ -0,0 +1,19 @@
...
+{
+	using super = cClearMetaOnDrop<cYawRotator<cBlockHandler, 0x3, 0x1, 0x3, 0x0, 0x2>>;
+
+public:
+
+	cBlockGlazedTerracottaHandler(BLOCKTYPE a_BlockType) : super(a_BlockType)
+	{
+	}
+
+};
```
madmaxoft, line 19: Couldn't this entire thing be replaced with a simple `using cBlockGlazedTerracottaHandler = cClearMetaOnDrop<cYawRotator<cBlockHandler, 0x3, 0x1, 0x3, 0x0, 0x2>>` ? If so, would it make sense to put that into the BlockHandler.cpp file? (this is a real question, not a suggestion as such)

bearbin: I think it can, there's not much point having an empty class really! It just turned out this way because I created this file before starting out the refactoring.

src/Blocks/Mixins.h
bearbin: I have an objection to using 2-digit hex constants for NIBBLETYPEs, it may be consistent but it's also just _wrong_.

madmaxoft: I know that they're only 4-bit, but still they use a full byte in memory; I come from the ancient times of byte-sized hex dumps and have a strong dislike of odd-sized hex constants.

I guess we need a third person to decide this tie.

src/Blocks/BlockChest.h
```diff
@@ -33,21 +33,27 @@ class cBlockChestHandler :
 		BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta
 	) override
 	{
-		a_BlockType = m_BlockType;
-
```
madmaxoft, line 37: Why remove this? This way `a_BlockType` can be undefined upon leaving the function.

src/Blocks/Mixins.h
tigerw: http://random.org

src/Blocks/BlockChest.h
bearbin: Because a_BlockType will be reset by super::GetBlockTypeMeta, and it would be confusing to have it set here as well when the value will be overridden later.

madmaxoft: But there's no call to `super:GetBlockTypeMeta`...

src/Blocks/Mixins.h
madmaxoft: Not helpful, tiger.

src/Blocks/BlockChest.h
bearbin: Look down a bit... If we return false before setting a_BlockType then there should be no issue, right? The placement has failed so the caller has no right to see what block has been placed.

madmaxoft: Ook, I was looking at the wrong source.

src/Blocks/Mixins.h
tigerw: cMetaRotator uses 2 digits almost exclusively (apart from the first argument in BlockBed.h for some reason), so in the interests of consistency I would recommend doing the same.

bearbin: All right, tie broken :P

tigerw:
About observers: they can use the code in RedstoneTorchHandler.h to delay 1 tick, turn on, then delay 1 tick, turn off. GetValidSourcePositions would return empty.

The tricky part might be actually making sure the block update comes from the ಠ_ಠ direction, since Update, where all the action happens, can get called when any side does a SetBlock.

PR #4627, opened 2020-04-08
Wolves and mooshrooms are passive mobs

mathiascode:
Dogs would not let owners sleep due to being classified as hostile.


bearbin:
It looks like a few more mobs need updating as well (I noticed slimes at least) according to the Wiki: https://minecraft.gamepedia.com/Bed#Sleeping

mathiascode:
Seems like those mobs will need special cases (hostile mobs that don't prevent sleeping). Probably better to create a separate PR for all of that.

PR #4632, opened 2020-04-09
Fix Placement of Various Items

### LINKED ISSUE #4629: Chests can be placed in half slabs
Client version: 12.2.2
Server OS: Linux
Cuberite Commit id: 16b3de94c89edae2ba8ff7ad6baa6b322c4571af

### Expected behavior
Chests are not placeable on a block if there is a half slab on the focused side.

### Actual behavior
The half slab dissapears and is replaced with a chest.

Place a half slab on a block and right click the same face of the block with a chest in your hand equipped.

### Server log
```
-
```

### LINKED ISSUE #4630: Doors can't be placed on full block half slabs
Client version: 12.2.2
Server OS: Linux
Cuberite Commit id: 16b3de94c89edae2ba8ff7ad6baa6b322c4571af

### Expected behavior
Doors are placeable on double half slabs.

### Actual behavior
The door can't be placed.

Put 2 half slabs on top of each other so they form a full block and try to place a door on top of that.

### Server log
```
-
```

bearbin:
Prevents items from being placed inside each other, and other things.

TODO:

 * Properly refactor item placement logic to include other bad code paths.
 * Check if plugin hooks are getting called correctly on block replacement destruction inside item placement.

FIXES:

Fixes #4630


src/Items/ItemChest.h
```diff
@@ -28,13 +28,15 @@ class cItemChestHandler:
 		int a_CursorX, int a_CursorY, int a_CursorZ
 	) override
 	{
+		Vector3i absPos(a_BlockX, a_BlockY, a_BlockZ);
```
tigerw, line 31: butWhy

And what was wrong with keeping blockPos down there? Seems like using a_BlockY in the valid height check makes for a shorter line.

bearbin: Preparation for the conversion to Vector3s everywhere! We need vector3s in one place so we should use them everywhere for consistency. And if they're everywhere then the defintion needs to be at the start of the function.

tigerw: Would you consider `BlockPos` instead?

bearbin: This is due to be refactored again, that is indeed a better name.

src/Items/ItemChest.h
```diff
@@ -43,36 +45,37 @@ class cItemChestHandler:
...
+		a_World.GetBlockTypeMeta(absPos.x, absPos.y, absPos.z, clickedBlock, clickedBlockMeta);
+
 		cChunkInterface ChunkInterface(a_World.GetChunkMap());
-		auto blockHandler = BlockHandler(clickedBlock);
-		if (blockHandler->DoesIgnoreBuildCollision(ChunkInterface, blockPos, a_Player, clickedBlockMeta))
+		// Check if the block ignores build collision (water, grass etc.):
+		if (BlockHandler(clickedBlock)->DoesIgnoreBuildCollision(ChunkInterface, absPos, a_Player, clickedBlockMeta))
 		{
-			blockHandler->OnPlayerBreakingBlock(ChunkInterface, a_World, a_Player, blockPos);
+			a_World.DropBlockAsPickups(absPos, &a_Player, nullptr);
```
tigerw, line 54: Do these "replaceable" blocks ever drop anything when replaced?
Also why this change?

bearbin: Think of grass, seeds will be dropped sometimes.

And this was changed to use the new DropBlockAsPickups to enable this functionality, instead of just removing it without dropping anything.

tigerw: I don't recall them popping anything out when say, putting a torch in them, only when left clicking to break.

bearbin: Admittedly I haven't tested. But they _should_. And in any case it should be consistent between chests and any other item.

src/BlockInfo.cpp
```diff
@@ -733,7 +726,6 @@ cBlockInfo::cBlockInfoArray::cBlockInfoArray()
 	Info[E_BLOCK_POWERED_RAIL                 ].m_Transparent = true;
 	Info[E_BLOCK_PUMPKIN_STEM                 ].m_Transparent = true;
 	Info[E_BLOCK_PURPLE_SHULKER_BOX           ].m_Transparent = true;
-	Info[E_BLOCK_PURPUR_DOUBLE_SLAB           ].m_Transparent = true;
```
tigerw, line 736: So we just need the double slab changes from here and we'll fix the door issue right?

bearbin: Should be.

tigerw:
intent to apply changes to PR

bearbin:
Alright, I've somewhat forgotten what changes remained to be done here...

src/Items/ItemChest.h
Janmm14: According to https://minecraft.fandom.com/wiki/Grass#Trivia grass does not drop seed when its being replacing.

bearbin:
Needs approval. @12xx12 ?

12xx12 (approved):

PR #4636, opened 2020-04-10
fixing rotation - rel. #4625

### LINKED ISSUE #4234: Observers always face up
Client version:   1.12.2
Server OS:        Linux
Commit id:        b9fdaf8a942ea2f79e48e3f735ba8916f3364e70

When placed, observers always face up

GefaketHD:
Adapting rotation to read it more easily and match the vanilla ones.
fixes #4234


tigerw:
No objections but I admit I haven't gone through the yaw conditions.

tigerw:
Though hold on, before say you had a rotation of 0, it calculated `Rotation = 0 + 90 + 45`, then hit `if Rotation >= 90 and Rotation <= 180` finally returning `North`.

Now it hits `Rotation >= -45 and Rotation < 45` before `return South`.

If I'm reading this right doesn't this break stuff?

bearbin:
This fixes stuff that was wrong. I worked on this bit of code together with @GefaketHD. The previous version (which I just copied and pasted from where it was before in the code) was not correct, and the new version does correspond with reality.

tigerw:
Ah, so PlayerYawToMetadata returns the mirrored direction?

bearbin:
So the previous one was completely broken when the rotation was between -135 and -180, but otherwise it works the same as the new one.

bearbin:
What this needs adding is an additional parameter for `cYawRotator` and `cPitchYawRotator` `FacesTowardsPlayer`, which if true flips the output of `YawToMetaData`, and then all items that use the rotators need to have their meta values reviewing.

tigerw:
Is it possible to adjust the meta template parameters to handling facing towards/away?

bearbin:
Unfortunately not, because the Metas would have the wrong chirality and then MetaRotateXX would not work properly.

bearbin:
I have now corrected the metas for all blocks using the cYawRotator/cPitchYawRotator mixins.

tigerw:
For the pulls to not diverge
And the features all emerge;

I urge,
go splurge!

*Merge, merge, merge!*

madmaxoft:
And then some guys wonder why we don't like using `North` and `East` etc. It's been changed once in Minecraft history and still the directions are unintuitive. I say get rid of all of them, replace with `XPlus`, `XMinus`, `ZPlus`, `ZMinus`.

GefaketHD:
Tbh, I don't think that XPlus, XMinus etc. is less confusing.

GefaketHD:
North, East, South and West at least have an order and re more intuitively to use, becase they appear in the real life world too.

madmaxoft:
Order? SENW? How? And what do you need directions to be ordered.

If I tell you "what's the coord of the block through this block's east face", you'll need to consult the wiki to actually know what way to manipulate the coords. If I tell you "what's the coord of the XM neighbor block", you know immediately to subtract 1 from the X value, no lookup needed. That's the power of the XM/ XP / ZM / ZP marking. And they do have an order, after all - an alphabetic one :)

madmaxoft:
And leave intuitivity out of this, as long as Minecraft is confused about where east and west is :P

GefaketHD:
Never Eat Sea Weed. Im referring to the compass. No need to go off like that. ^^
But you got a point with the calculations - they are easier with X/Z Plus/Minus.

madmaxoft:
Sorry, I'm a bit flaming hot, there was a terrible incident at ZeroBraneStudio, a self-entitled idiot making a lot of noise, so I'm a bit off.

madmaxoft:
The code changes look alright, then, but I didn't actually test the changes in-game. I hope someone did?

GefaketHD:
Well, I did but I guess that's not what you wanted to know.
Also I tested the door-Issue and I can say: EVERY door faces the wrong way after placed now... :x 
This PR is for Terracotta and Observers and there it works like a charm, but doors seem to be impacted negatively.

tigerw:
I have tested both yaw and pitch rotatable blocks and everything worked, apart from the existing door bug I referenced earlier.

tigerw:
> a self-entitled idiot making a lot of noise

Go on...

ZeroBrane, the Lua IDE?

bearbin:
Link broken to avoid github detecting as a reference: https://github.com/pkulchenko/ZeroBraneStudio [/] issues/1046

tigerw:
That was a live-action trainwreck of a GitHub issue.

Dude really had ZeroBranes.

madmaxoft:
@tigerw Please DON'T feed the troll. We've already posted some support to Paul in the ZBS mailinglist.

PR #4643, opened 2020-04-11
Pressureplateimprovements

### LINKED ISSUE #4637: Pressure plate doesn't make sound
Client version: 12.2.2
Server OS: Linux

### Expected behavior
When u step ona pressure plate or when it is released you hear a sound.

### Actual behavior
It stays quiet.

Step on a pressure plate.

### Server log
```
-
```

GefaketHD:
Trying to make pressure plates staying pressed for at least 1 sec and 0.5 sec after noone is on it anymore. (ref: https://minecraft.gamepedia.com/Pressure_Plate#Redstone_component - Activation-Part)

Edit: fixes #4637


GefaketHD:
[crash.txt](https://github.com/cuberite/cuberite/files/4465303/crash.txt)
Does someone have an idea what exactly causes that crash?

tigerw:
looking

GefaketHD:
I managed to get rid of the crash. But it still is not working as expected. I added comments to tell what I want to do there and what my thoughts were.
Normally, like the code is now, it should be and stay pressed, because the scheduler that would make it releasable is commented out, but that's not the case. It's clicking repeatedly wheile you stand on it...

GefaketHD:
Please try to break it REALLY hard before accepting the changes...I did a lot of testing now - it should not be crashing, but I am not sure if there is not another way to break the system.
I'm thinking of stuck of broken pressure plates etc.

bearbin:
In terms of code structure, the states are sensible but the logic that uses them is a bit complicated and long. I would seperate the use of the states into a seperate function, something like `DoPressurePlateStateTransition` which implements a state machine more explicitly and switches by the states first and then the power level etc inside the block for each state.

GefaketHD:
I will give it a shot.

tigerw:
Infrastructure for delaying redstone actions is available via GetMechanismDelayInfo, I think this can be used here.

GefaketHD:
Oh okay didn't really recognize that, thanks!

GefaketHD:
Why can't I _set_ the delay though?

GefaketHD:
But I'm not really sure, if that does, what I need here...because I need to work with the states I made and the m_MechanismDelays only stores, if it should be on or off. I'm not really delaying a redstone action, but the action to emit the redstone.

What part of the code did you think could be replaced with that MechanismDelay-map?

GefaketHD:
Well the commit naming is bad - it's not really a handler more a function

GefaketHD:
Could someone tell if the failing check is my fault? I did read the log as if someone failed internally in the checking instance...?

madmaxoft:
The Appveyor CI sometimes fails for no specific reason, and this is one of those times. I've restarted the build.

GefaketHD:
Thank you

tigerw:
Something like

```
nowpower
delayinfo

if null delayinfo
	if nowpower = on
		set meta
		add delay of 1 second to maybe power off
		return update

	// nowpower = off
	return update nothing
else
	if delay expired
		if nowpower = on
			set delay of 0.5 seconds to maybe power off
			return update nothing

		// nowpower = off
		delete delayinfo
		set meta
		return update

	// delay not expired
	return update nothing
```

GefaketHD:
Ok I will update that this evening/tonight

tigerw:
The problem with the pseudocode I posted was it doesn't correctly delay 0.5 seconds after the player steps off. To be honest I'm not sure this is better but for your consideration:

```cpp
const auto Power = GetPowerLevel(a_World, a_Position, a_BlockType, a_Meta);
const auto PlateUpdates = GetAdjustedRelatives(a_Position, StaticAppend(GetRelativeLaterals(), cVector3iArray{ OffsetYM() }));

auto Data = static_cast<cIncrementalRedstoneSimulator *>(a_World.GetRedstoneSimulator())->GetChunkData();
auto DelayInfo = Data->GetMechanismDelayInfo(a_Position);

// Resting state?
if (DelayInfo == nullptr)
{
	if (Power == 0)
	{
		// Nothing happened, back to rest
		return {};
	}

	// From rest, a player stepped on us
	// Schedule a minimum 0.5 second delay before even thinking about releasing
	Data->m_MechanismDelays[a_Position] = std::make_pair(10, true);

	// Update power
	Data->SetCachedPowerData(a_Position, PoweringData(E_BLOCK_AIR, Power));

	// Immediately depress plate
	a_World.SetBlockMeta(a_Position, E_META_PRESSURE_PLATE_DEPRESSED);
	return PlateUpdates;
}

// Not a resting state

int DelayTicks;
bool HasExitedMinimumOnDelayPhase;
std::tie(DelayTicks, HasExitedMinimumOnDelayPhase) = *DelayInfo;

// Are we waiting for the initial delay or subsequent release delay?
if (DelayTicks > 0)
{
	// Yes. Are we waiting to release, and found that the player stepped on it again?
	if (!HasExitedMinimumOnDelayPhase && (Power > 0))
	{
		// Reset delay
		*DelayInfo = std::make_pair(0, true);
	}

	return {};
}

// Not waiting for anything. Has the initial delay elapsed?
if (HasExitedMinimumOnDelayPhase)
{
	// Yep, initial delay elapsed. Has the player gotten off?
	if (Power == 0)
	{
		// Yes. Go into subsequent release delay, for a further 0.5 seconds
		*DelayInfo = std::make_pair(10, false);
	}

	// Yes, but player's still on the plate, do nothing
	return {};
}

// Just got out of the subsequent release phase, reset everything and raise the plate
Data->m_MechanismDelays.erase(a_Position);
Data->SetCachedPowerData(a_Position, PoweringData(E_BLOCK_AIR, Power));

a_World.SetBlockMeta(a_Position, E_META_PRESSURE_PLATE_RAISED);
return PlateUpdates;
```

GefaketHD:
> The problem with the pseudocode I posted was it doesn't correctly delay 0.5 seconds after the player steps off. To be honest I'm not sure this is better but for your consideration:
> 
> ```c++
> const auto Power = GetPowerLevel(a_World, a_Position, a_BlockType, a_Meta);
> const auto PlateUpdates = GetAdjustedRelatives(a_Position, StaticAppend(GetRelativeLaterals(), cVector3iArray{ OffsetYM() }));
> 
> auto Data = static_cast<cIncrementalRedstoneSimulator *>(a_World.GetRedstoneSimulator())->GetChunkData();
> auto DelayInfo = Data->GetMechanismDelayInfo(a_Position);
> 
> // Resting state?
> if (DelayInfo == nullptr)
> {
> 	if (Power == 0)
> 	{
> 		// Nothing happened, back to rest
> 		return {};
> 	}
> 
> 	// From rest, a player stepped on us
> 	// Schedule a minimum 0.5 second delay before even thinking about releasing
> 	Data->m_MechanismDelays[a_Position] = std::make_pair(10, true);
> 
> 	// Update power
> 	Data->SetCachedPowerData(a_Position, PoweringData(E_BLOCK_AIR, Power));
> 
> 	// Immediately depress plate
> 	a_World.SetBlockMeta(a_Position, E_META_PRESSURE_PLATE_DEPRESSED);
> 	return PlateUpdates;
> }
> 
> // Not a resting state
> 
> int DelayTicks;
> bool HasExitedMinimumOnDelayPhase;
> std::tie(DelayTicks, HasExitedMinimumOnDelayPhase) = *DelayInfo;
> 
> // Are we waiting for the initial delay or subsequent release delay?
> if (DelayTicks > 0)
> {
> 	// Yes. Are we waiting to release, and found that the player stepped on it again?
> 	if (!HasExitedMinimumOnDelayPhase && (Power > 0))
> 	{
> 		// Reset delay
> 		*DelayInfo = std::make_pair(0, true);
> 	}
> 
> 	return {};
> }
> 
> // Not waiting for anything. Has the initial delay elapsed?
> if (HasExitedMinimumOnDelayPhase)
> {
> 	// Yep, initial delay elapsed. Has the player gotten off?
> 	if (Power == 0)
> 	{
> 		// Yes. Go into subsequent release delay, for a further 0.5 seconds
> 		*DelayInfo = std::make_pair(10, false);
> 	}
> 
> 	// Yes, but player's still on the plate, do nothing
> 	return {};
> }
> 
> // Just got out of the subsequent release phase, reset everything and raise the plate
> Data->m_MechanismDelays.erase(a_Position);
> Data->SetCachedPowerData(a_Position, PoweringData(E_BLOCK_AIR, Power));
> 
> a_World.SetBlockMeta(a_Position, E_META_PRESSURE_PLATE_RAISED);
> return PlateUpdates;
> ```

If you want me to do it like that, I of course will.

GefaketHD:
Just a quick question: does this work with redstone-ticks then...? Because when I tested it, the delays were twice as long as they should have been.

GefaketHD:
And the power should be set seperately - because if it is a weighted pressure plate, the power level could change even if the "depressed/released state" itself doesn't.

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -36,6 +37,11 @@ class cPressurePlateHandler : public cRedstoneHandler
 					FoundPlayer = true;
 				}
 
+				if (a_Entity.IsPickup())
+				{
+					NumberOfEntities += dynamic_cast<cPickup*>(&a_Entity)->GetItem().m_ItemCount;
```
peterbell10, line 42:
```suggestion
					NumberOfEntities += static_cast<cPickup &>(a_Entity).GetItem().m_ItemCount;
```

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -72,16 +78,134 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+				// Nothing happened, back to rest
+				return {};
+			}
+
+			// From rest, a player stepped on us
+			// Schedule a minimum 0.5 second delay before even thinking about releasing
+			ChunkData->m_MechanismDelays[a_Position] = std::make_pair(5, true);
+
+			// manage on-sound
+			AString soundToPlay = "";
```
peterbell10, line 103: Agree with @tigerw. This should be its own `static` function:
```cpp
auto soundToPlay = GetClickOnSound(a_BlockType);
```
Same for the off sound.

tigerw:
Your code was good, don't feel the need to change it :)

And you're right that the proposal didn't update the power of weighted plates + delayed 2x what it should've been.

tigerw:
One suggestion might be to move the switch cases dealing with the sound to play into `cBlockPressurePlateHandler` in Blocks/PressurePlateHandler.h instead.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandler.h
```diff
@@ -14,6 +14,14 @@ class cRedstoneHandler
...
 	DISALLOW_COPY_AND_ASSIGN(cRedstoneHandler);
 
+	enum ENUM_PRESSURE_PLATE_STATE
+	{
+		E_PRESSURE_PLATE_RAISED            = 0,
+		E_PRESSURE_PLATE_INITIALLY_PRESSED = 1,
+		E_PRESSURE_PLATE_HELD_DOWN         = 2,
+		E_PRESSURE_PLATE_WANTS_TO_RELEASE  = 3,
+	} ;
+
```
peterbell10, line 24: This isn't used any more.

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -72,16 +78,134 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+		{
+			// Yes. Are we waiting to release, and found that the player stepped on it again?
+			if (!HasExitedMinimumOnDelayPhase && (Power > 0))
+			{
+				// Reset delay
+				*DelayInfo = std::make_pair(0, true);
+			}
+
+			// Did the power level change and is still above zero?
+			if ((Power != PreviousPower.PowerLevel) && (Power > 0))
```
peterbell10, line 149: Instead of putting `&& (Power > 0)` on every condition, how about returning early?
```cpp
if (Power == 0)
{
    return {};
}
```

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -72,16 +78,134 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+		{
+			// Yep, initial delay elapsed. Has the player gotten off?
+			if (Power == 0)
+			{
+				// Yes. Go into subsequent release delay, for a further 0.5 seconds
+				*DelayInfo = std::make_pair(5, false);
+			}
+
+			// Did the power level change and is still above zero?
+			if ((Power != PreviousPower.PowerLevel) && (Power > 0))
```
peterbell10, line 170: Same here

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
peterbell10, line 42: Make `NumberOfEntities` an `int` to fix the clang warning.

GefaketHD: I don't get a warning there but why not. :)

GefaketHD: Ah you mean the check complained, I just saw it now.

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -72,16 +78,103 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+		if (DelayTicks > 0)
+		{
+			// Nothing changes, if there is nothing on it anymore, because the state is locked.
+			if (Power == 0)
+			{
+				return {};
+			}
+
+			// Yes. Are we waiting to release, and found that the player stepped on it again?
+			if (!HasExitedMinimumOnDelayPhase && (Power > 0))
```
peterbell10, line 129:
```suggestion
			if (!HasExitedMinimumOnDelayPhase)
```

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -72,16 +78,103 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+			{
+				// Reset delay
+				*DelayInfo = std::make_pair(0, true);
+			}
+
+			// Did the power level change and is still above zero?
+			if ((Power != PreviousPower.PowerLevel) && (Power > 0))
```
peterbell10, line 136:
```suggestion
			if (Power != PreviousPower.PowerLevel)
```

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -72,16 +78,103 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+			{
+				// Nothing happened, back to rest
+				return {};
+			}
+
+			// From rest, a player stepped on us
+			// Schedule a minimum 0.5 second delay before even thinking about releasing
+			ChunkData->m_MechanismDelays[a_Position] = std::make_pair(5, true);
+
+			auto soundToPlay = GetClickOffSound(a_BlockType);
```
peterbell10, line 102:
```suggestion
			auto soundToPlay = GetClickOnSound(a_BlockType);
```

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -92,4 +185,53 @@ class cPressurePlateHandler : public cRedstoneHandler
...
+
+private:
+	static AString GetClickOnSound(BLOCKTYPE a_BlockType)
+	{
+		// manage on-sound
+		AString soundToPlay = "";
+		switch (a_BlockType)
+		{
+			case E_BLOCK_STONE_PRESSURE_PLATE:
+				soundToPlay = "block.wood_pressureplate.click_on";
```
peterbell10, line 197: You can drop the `soundToPlay` variable:
```cpp
case E_BLOCK_STONE_PRESSURE_PLATE:
	return "block.wood_pressureplate.click_on";
// etc...
```

GefaketHD: True :D

GefaketHD:
Thank you for your open eyes. :) And sorry for these stupid mistakes

GefaketHD:
This one fixes #4637 btw:)

PR #4644, opened 2020-04-11
More Vector3 in cBlockHandler

madmaxoft:
`cBlockHandler.OnUpdate()` uses `Vector3` params.

Also:
 - Slightly changed how block ticking works, now needs only 1 random and is easier to read.
 - Changed the API function from `cWorld:SetNextBlockTick(number, number, number)` to `cWorld:SetNextBlockToTick(Vector3i)`, with an deprecation warning for the old function. A search through entire GitHub shows no usage other than our Debuggers plugin.

### DIFF
--- a/Server/Plugins/APIDump/Classes/World.lua
+++ b/Server/Plugins/APIDump/Classes/World.lua
@@ -3032,7 +3032,18 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 						Type = "number",
 					},
 				},
-				Notes = "Sets the blockticking to start at the specified block in the next tick.",
+				Notes = "DEPRECATED, use SetNextBlockToTick() instead.",
+			},
+			SetNextBlockToTick =
+			{
+				Params =
+				{
+					{
+						Name = "BlockPos",
+						Type = "Vector3i",
+					},
+				},
+				Notes = "Requests that the specified block be ticked at the start of the next world tick. Only one block per chunk can be queued this way; a second call to the same chunk overwrites the previous call.",
 			},
 			SetSavingEnabled =
 			{
--- a/src/Blocks/BlockCauldron.h
+++ b/src/Blocks/BlockCauldron.h
@@ -19,10 +19,18 @@ public:
 	{
 	}
 
+
+
+
+
 	virtual cItems ConvertToPickups(NIBBLETYPE a_BlockMeta, cBlockEntity * a_BlockEntity, const cEntity * a_Digger, const cItem * a_Tool) override
 	{
 		return cItem(E_ITEM_CAULDRON, 1, 0);
 	}
+
+
+
+
 
 	virtual bool OnUse(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ) override
 	{
@@ -86,28 +94,44 @@ public:
 		return true;
 	}
 
+
+
+
+
 	virtual bool IsUseable() override
 	{
 		return true;
 	}
 
-	virtual void OnUpdate(cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface, cBlockPluginInterface & a_PluginInterface, cChunk & a_Chunk, int a_RelX, int a_RelY, int a_RelZ) override
+
+
+
+
+	virtual void OnUpdate(
+		cChunkInterface & a_ChunkInterface,
+		cWorldInterface & a_WorldInterface,
+		cBlockPluginInterface & a_PluginInterface,
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) override
 	{
-		int BlockX = a_RelX + a_Chunk.GetPosX() * cChunkDef::Width;
-		int BlockZ = a_RelZ + a_Chunk.GetPosZ() * cChunkDef::Width;
-		if (!a_WorldInterface.IsWeatherWetAt(BlockX, BlockZ) || (a_RelY != a_WorldInterface.GetHeight(BlockX, BlockZ)))
+		auto WorldPos = a_Chunk.RelativeToAbsolute(a_RelPos);
+		if (!a_WorldInterface.IsWeatherWetAtXYZ(WorldPos.addedY(1)))
 		{
 			// It's not raining at our current location or we do not have a direct view of the sky
-			// We cannot eat the rain :(
 			return;
 		}
 
-		NIBBLETYPE Meta = a_Chunk.GetMeta(a_RelX, a_RelY, a_RelZ);
+		auto Meta = a_Chunk.GetMeta(a_RelPos);
 		if (Meta < 3)
 		{
-			a_Chunk.SetMeta(a_RelX, a_RelY, a_RelZ, Meta + 1);
+			a_Chunk.SetMeta(a_RelPos, Meta + 1);
 		}
 	}
+
+
+
+
 
 	virtual ColourID GetMapBaseColourID(NIBBLETYPE a_Meta) override
 	{
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -41,26 +41,32 @@ public:
 
 
 
-	virtual void OnUpdate(cChunkInterface & cChunkInterface, cWorldInterface & a_WorldInterface, cBlockPluginInterface & a_PluginInterface, cChunk & a_Chunk, int a_RelX, int a_RelY, int a_RelZ) override
+	virtual void OnUpdate(
+		cChunkInterface & a_ChunkInterface,
+		cWorldInterface & a_WorldInterface,
+		cBlockPluginInterface & a_PluginInterface,
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) override
 	{
-		NIBBLETYPE BlockMeta = a_Chunk.GetMeta(a_RelX, a_RelY, a_RelZ);
+		auto BlockMeta = a_Chunk.GetMeta(a_RelPos);
 
-		if (IsWaterInNear(a_Chunk, a_RelX, a_RelY, a_RelZ))
+		if (IsWaterInNear(a_Chunk, a_RelPos))
 		{
 			// Water was found, set block meta to 7
-			a_Chunk.FastSetBlock(a_RelX, a_RelY, a_RelZ, m_BlockType, 7);
+			a_Chunk.FastSetBlock(a_RelPos, m_BlockType, 7);
 			return;
 		}
 
 		// Water wasn't found, de-hydrate block:
 		if (BlockMeta > 0)
 		{
-			a_Chunk.FastSetBlock(a_RelX, a_RelY, a_RelZ, E_BLOCK_FARMLAND, --BlockMeta);
+			a_Chunk.FastSetBlock(a_RelPos, E_BLOCK_FARMLAND, --BlockMeta);
 			return;
 		}
 
 		// Farmland too dry. If nothing is growing on top, turn back to dirt:
-		BLOCKTYPE UpperBlock = (a_RelY >= cChunkDef::Height - 1) ? static_cast<BLOCKTYPE>(E_BLOCK_AIR) : a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ);
+		auto UpperBlock = cChunkDef::IsValidHeight(a_RelPos.y + 1) ? a_Chunk.GetBlock(a_RelPos.addedY(1)) : E_BLOCK_AIR;
 		switch (UpperBlock)
 		{
 			case E_BLOCK_BEETROOTS:
@@ -75,7 +81,7 @@ public:
 			}
 			default:
 			{
-				a_Chunk.SetBlock({a_RelX, a_RelY, a_RelZ}, E_BLOCK_DIRT, 0);
+				a_Chunk.SetBlock(a_RelPos, E_BLOCK_DIRT, 0);
 				break;
 			}
 		}
@@ -111,11 +117,12 @@ public:
 
 
 
-	bool IsWaterInNear(cChunk & a_Chunk, int a_RelX, int a_RelY, int a_RelZ)
+	/** Returns true if there's either a water source block close enough to hydrate the specified position, or it's raining there. */
+	bool IsWaterInNear(cChunk & a_Chunk, const Vector3i a_RelPos)
 	{
-		if (a_Chunk.GetWorld()->IsWeatherWetAt(a_RelX, a_RelZ))
+		if (a_Chunk.GetWorld()->IsWeatherWetAtXYZ(a_RelPos))
 		{
-			// Rain hydrates farmland, too, except in Desert biomes.
+			// Rain hydrates farmland, too
 			return true;
 		}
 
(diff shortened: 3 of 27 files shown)


madmaxoft:
Really, still no review? I know this is a large one, but it's mostly mechanical - changing the three ints to a Vector3. And the changed block ticking has been tested.

Server/Plugins/APIDump/Classes/World.lua
```diff
@@ -3032,7 +3032,18 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 						Type = "number",
 					},
 				},
-				Notes = "Sets the blockticking to start at the specified block in the next tick.",
+				Notes = "DEPRECATED, use SetNextBlockToTick() instead.",
```
tigerw, line 3035: Just remove it already...

src/Blocks/BlockCauldron.h
```diff
@@ -86,29 +94,45 @@ class cBlockCauldronHandler :
...
+		cWorldInterface & a_WorldInterface,
+		cBlockPluginInterface & a_PluginInterface,
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) override
 	{
-		int BlockX = a_RelX + a_Chunk.GetPosX() * cChunkDef::Width;
-		int BlockZ = a_RelZ + a_Chunk.GetPosZ() * cChunkDef::Width;
-		if (!a_WorldInterface.IsWeatherWetAt(BlockX, BlockZ) || (a_RelY != a_WorldInterface.GetHeight(BlockX, BlockZ)))
+		auto worldPos = a_Chunk.RelativeToAbsolute(a_RelPos);
```
tigerw, line 118: Strongly prefer naming in line with style guide.

src/Blocks/BlockFarmland.h
```diff
@@ -41,26 +41,32 @@ class cBlockFarmlandHandler :
...
 		if (BlockMeta > 0)
 		{
-			a_Chunk.FastSetBlock(a_RelX, a_RelY, a_RelZ, E_BLOCK_FARMLAND, --BlockMeta);
+			a_Chunk.FastSetBlock(a_RelPos, E_BLOCK_FARMLAND, --BlockMeta);
 			return;
 		}
 
 		// Farmland too dry. If nothing is growing on top, turn back to dirt:
-		BLOCKTYPE UpperBlock = (a_RelY >= cChunkDef::Height - 1) ? static_cast<BLOCKTYPE>(E_BLOCK_AIR) : a_Chunk.GetBlock(a_RelX, a_RelY + 1, a_RelZ);
+		auto UpperBlock = cChunkDef::IsValidHeight(a_RelPos.y + 1) ? E_BLOCK_AIR : a_Chunk.GetBlock(a_RelPos.addedY(1));
```
tigerw, line 69: Technically the < 0 is unneeded

src/Blocks/BlockFluid.h
```diff
@@ -10,11 +10,11 @@
 class cBlockFluidHandler :
 	public cBlockHandler
 {
-	typedef cBlockHandler super;
+	using Super = cBlockHandler;
```
tigerw, line 13: Now we're just deliberately introducing inconsistency :P

madmaxoft: Already there, and we want classes to start with a capital, so it's actually fixing an inconsistency.

src/Blocks/BlockGrass.h
```diff
@@ -39,34 +39,38 @@ class cBlockGrassHandler :
...
 			{
-				a_Chunk.FastSetBlock(a_RelX, a_RelY, a_RelZ, E_BLOCK_DIRT, E_META_DIRT_NORMAL);
+				a_Chunk.FastSetBlock(a_RelPos, E_BLOCK_DIRT, E_META_DIRT_NORMAL);
 				return;
 			}
 
-			NIBBLETYPE light = std::max(a_Chunk.GetBlockLight(a_RelX, a_RelY + 1, a_RelZ), a_Chunk.GetTimeAlteredLight(a_Chunk.GetSkyLight(a_RelX, a_RelY + 1, a_RelZ)));
-			// Source block is not bright enough to spread
+			// Make sure that there is enough light at the source block to spread
+			auto light = std::max(a_Chunk.GetBlockLight(a_RelPos.addedY(1)), a_Chunk.GetTimeAlteredLight(a_Chunk.GetSkyLight(a_RelPos.addedY(1))));
```
tigerw, line 69: We add 1 to RelPos four times; worth just modifying a_RelPos?

tigerw: also `Light`

src/Blocks/BlockLeaves.h
```diff
@@ -10,29 +10,85 @@
...
+			{
+				for (int z = a_BlockPos.z - i; z <= a_BlockPos.z + i; z++)
+				{
+					for (int x = a_BlockPos.x - i; x <= a_BlockPos.x + i; x++)
+					{
+						if (a_Area.GetBlockType(x, y, z) != E_BLOCK_SPONGE + i)
+						{
+							continue;
+						}
+						#define PROCESS_NEIGHBOR(x, y, z) \
```
tigerw, line 61: What was the rationale for this change? Surely the pre-processor doesn't respect scope

src/Blocks/BlockLeaves.h
```diff
@@ -104,46 +160,50 @@ class cBlockLeavesHandler :
...
-		NIBBLETYPE Meta = a_Chunk.GetMeta(a_RelX, a_RelY, a_RelZ);
+		auto Meta = a_Chunk.GetMeta(a_RelPos);
 		if ((Meta & 0x04) != 0)
 		{
 			// Player-placed leaves, don't decay
 			return;
 		}
 
-		if ((Meta & 0x8) == 0)
+		if ((Meta & 0x08) == 0)
```
tigerw, line 178: 🤨

src/Chunk.h
```diff
@@ -375,12 +375,12 @@ class cChunk :
 		m_IsSaving = false;
 	}
 
-	/** Sets the blockticking to start at the specified block. Only one blocktick may be set, second call overwrites the first call */
-	inline void SetNextBlockTick(int a_RelX, int a_RelY, int a_RelZ)
+	/** Causes the specified block to be ticked on the next Tick() call.
+	Plugins can use this via the cWorld:SetNextBlockToTick() API.
+	Only one block coord per chunk may be set, a second call overwrites the first call */
+	inline void SetNextBlockToTick(const Vector3i a_RelPos)
```
tigerw, line 381: What use is an API that only allows ticking one block per chunk anyway?

src/Chunk.cpp
```diff
@@ -839,43 +836,27 @@ void cChunk::CheckBlocks()
...
-		m_BlockTickZ = TickZ / 2
-	)
+	// Tick random blocks, but the first one should be m_BlockToTick (so that SetNextBlockToTick() works)
+	auto idx = cChunkDef::MakeIndexNoCheck(m_BlockToTick);
+	for (int i = 0; i < 50; ++i)
 	{
-
-		if (m_BlockTickY > cChunkDef::GetHeight(m_HeightMap, m_BlockTickX, m_BlockTickZ))
+		auto pos = cChunkDef::IndexToCoordinate(static_cast<size_t>(idx));
+		idx = (idx + 1 + m_World->GetTickRandomNumber(cChunkDef::NumBlocks - 2)) % cChunkDef::NumBlocks;
```
tigerw, line 847: This calculation isn't obvious, could you comment?

src/ChunkMap.cpp
```diff
@@ -2100,17 +2100,17 @@ void cChunkMap::Tick(std::chrono::milliseconds a_Dt)
 
 
 
-void cChunkMap::TickBlock(int a_BlockX, int a_BlockY, int a_BlockZ)
+void cChunkMap::TickBlock(const Vector3i a_BlockPos)
 {
+	auto chunkPos = cChunkDef::BlockToChunk(a_BlockPos);
+	auto relPos = cChunkDef::AbsoluteToRelative(a_BlockPos, chunkPos);
```
tigerw, line 2106: The style guide: "am I a joke to you"?

tigerw:
Yes looks good

src/Blocks/BlockLeaves.h
madmaxoft: I like to have the definition as close to the usage as possible, and `#undef` as soon as possible. Macros have no scoping, so they pollute the global namespace and thus are dangerous. Having the macro definition at the top of the file and the function at the bottom was needlessly far away.

src/Chunk.h
madmaxoft: Good enough for a `magic sTick`, in Debuggers :)

src/ChunkMap.cpp
madmaxoft: >  This will break all plugins, so as a side note, I have a proposition. The naming convention of Cuberite classes has been bothering me since forever, and this is the perfect moment to make things right - we're already breaking compatibility, so we can break it thoroughly. How about dropping the "c" prefix to all classes? And while we're at it, how about switching to a style that has long been used and is proven good? Personally, I like the Java-like style with membership prefixes (ClassName, doSomethingFunction(), mMemberVariable, aParameter)

Source: https://forum.cuberite.org/thread-3215.html

src/Chunk.cpp
madmaxoft: Good catch, this was actually needlessly complicated, replaced with a simple random pick.

src/ChunkMap.cpp
tigerw: Yeah but some aren't as keen with Java style :)

madmaxoft: It's definitely more sane than what we have now. And it's long proven. Qt uses it, too, in C++ world, mind you.

tigerw: I see no reason why the first word should be lowercase, as if its somehow less important. Who came up with this style anyway?

src/Blocks/BlockGrass.h
```diff
@@ -79,12 +83,12 @@ class cBlockGrassHandler :
...
 			BLOCKTYPE  DestBlock;
 			NIBBLETYPE DestMeta;
-			if (!cChunkDef::IsValidHeight(a_RelY + OfsY))
+			if (!cChunkDef::IsValidHeight(a_RelPos.y + OfsY))
 			{
 				// Y Coord out of range
 				continue;
 			}
-			Vector3i pos(a_RelX + OfsX, a_RelY + OfsY, a_RelZ + OfsZ);
+			auto pos = a_RelPos + Vector3i(OfsX, OfsY, OfsZ);
```
bearbin, line 91: Move this before the height check to avoid duplication.

src/Blocks/BlockLeaves.h
peterbell10: If you're doing that, why not just make it a lambda?

src/Blocks/BlockVine.h
```diff
@@ -221,25 +221,32 @@ class cBlockVineHandler :
...
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) override
 	{
 		UNUSED(a_ChunkInterface);
 		UNUSED(a_WorldInterface);
 
 		// Vine cannot grow down if at the bottom:
-		if (a_RelY < 1)
+		if (a_RelPos.y < 1)
```
bearbin, line 236: For consistency, check -1 for valid height?

src/Blocks/BlockLeaves.h
madmaxoft: The macro is doing `return` from the entire function, which doesn't translate too well to a lambda. But it's true that it looks better.

madmaxoft:
For some reason I can't reply directly to @bearbin's comment about `Super` being weird.
Unlike `for` or `if`, `Super` is not a language keyword, it's a typedef for a class name, and as such, should use the same convention as other class names.

bearbin:
But `Super` isn't a class, if you try and open `Super.h` you won't find it's definition and super in one class should certainly behave differently from super in another. It's just a deficiency in C++ that we have to use a typedef rather than having super as a keyword.

tigerw:
We really gotta find some sort of consensus about your variable naming scheme. Piecemeal changes just isn't going to cut it, and the style guide is there for a reason.

madmaxoft:
Get me a tool that can change them all at once, and I'll happily run it :P

tigerw:
very funny >:(

madmaxoft:
Rewritten To Make @Tigerw Happy. Now This Looks Awful, But If It's The Way You Want It To Be...

Server/Plugins/APIDump/Classes/World.lua
madmaxoft: It is used in plugins, it gets a Deprecation.

src/Chunk.cpp
```diff
@@ -839,43 +836,27 @@ void cChunk::CheckBlocks()
...
-		cBlockHandler * Handler = BlockHandler(GetBlock(m_BlockTickX, m_BlockTickY, m_BlockTickZ));
+		cBlockHandler * Handler = BlockHandler(GetBlock(Pos));
 		ASSERT(Handler != nullptr);  // Happenned on server restart, FS #243
-		Handler->OnUpdate(ChunkInterface, *this->GetWorld(), PluginInterface, *this, m_BlockTickX, m_BlockTickY, m_BlockTickZ);
-	}  // for i - tickblocks
+		Handler->OnUpdate(ChunkInterface, *this->GetWorld(), PluginInterface, *this, Pos);
+	}  // for i
+
+	// Set a new random coord for the next tick:
+	m_BlockToTick = cChunkDef::IndexToCoordinate(static_cast<size_t>(m_World->GetTickRandomNumber(cChunkDef::NumBlocks - 1)));
```
peterbell10, line 859: `Idx` already holds an unticked index, right?
```suggestion
	m_BlockToTick = cChunkDef::IndexToCoordinate(Idx);
```

src/Blocks/BlockFluid.h
```diff
@@ -99,52 +103,64 @@ class cBlockFluidHandler :
...
 		int z = ((rnd / 16) % 3) - 1;  // -1 .. 1
+		auto Pos = a_RelPos + Vector3i(x, y, z);
 
 		// Check if it's fuel:
 		BLOCKTYPE BlockType;
 		if (
-			((a_RelY + y < 0) || (a_RelY + y >= cChunkDef::Height)) ||
-			!a_Chunk.UnboundedRelGetBlockType(a_RelX + x, a_RelY + y, a_RelZ + z, BlockType) ||
+			!cChunkDef::IsValidHeight(Pos.y) ||
+			!a_Chunk.UnboundedRelGetBlockType(a_RelPos, BlockType) ||
```
peterbell10, line 155:
```suggestion
			!a_Chunk.UnboundedRelGetBlockType(Pos, BlockType) ||
```

src/Blocks/BlockCauldron.h
```diff
@@ -86,29 +94,45 @@ class cBlockCauldronHandler :
...
+		cBlockPluginInterface & a_PluginInterface,
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) override
 	{
-		int BlockX = a_RelX + a_Chunk.GetPosX() * cChunkDef::Width;
-		int BlockZ = a_RelZ + a_Chunk.GetPosZ() * cChunkDef::Width;
-		if (!a_WorldInterface.IsWeatherWetAt(BlockX, BlockZ) || (a_RelY != a_WorldInterface.GetHeight(BlockX, BlockZ)))
+		auto WorldPos = a_Chunk.RelativeToAbsolute(a_RelPos);
+		if (!a_WorldInterface.IsWeatherWetAtXYZ(WorldPos))
```
peterbell10, line 119: I think `IsWeatherWetAtXYZ` classes it as dry if there is a block at the exact coordinate queried.
```suggestion
		if (!a_WorldInterface.IsWeatherWetAtXYZ(WorldPos.addedY(1)))
```

madmaxoft:
Fixed and rebased.

bearbin:
The remaining build failure appears to be just a random network failure, this looks good to go.

PR #4645, opened 2020-04-12
Added a check for byte-sized hex constants.

madmaxoft:
For consistency, all hex constants should have an even number of characters (byte-sized).

### DIFF
--- a/src/BlockType.h
+++ b/src/BlockType.h
@@ -557,7 +557,7 @@ enum ENUM_BLOCK_META : NIBBLETYPE
 	E_META_BIG_FLOWER_LARGE_FERN        = 3,
 	E_META_BIG_FLOWER_ROSE_BUSH         = 4,
 	E_META_BIG_FLOWER_PEONY             = 5,
-	// 0x8 is supposedly a bit flag but all vanilla plants have this value
+	// 0x08 is supposedly a bit flag but all vanilla plants have this value
 	E_META_BIG_FLOWER_TOP               = 10,
 
 	// E_BLOCK_BREWING_STAND metas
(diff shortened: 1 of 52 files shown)


tigerw:
I only voted to use them in the Meta rotation templates for consistency there, I don't really support changing them wholesale everywhere. As it stands this is just forcing you to type an extra zero (which as Pokechu said will always be 0) everywhere, akin to say, needlessly wrapping every function body in an extra set of braces.

madmaxoft:
But single-char hex constants look so ugly...

tigerw:
I feel like I'm entering into a contract negotiation here, but:

I don't have too strong of an opinion of this, I surmise you do. However I insit `TheseNames` are better than `theseNames`, whereas you suggested you don't care that much.

How about a deal?

bearbin:
There is another option -  we could switch from using hex constants to using binary. That way we'll always have 4 digits and it'll be even everywhere. Plus that way we wouldn't have ugly letters in our constants spoiling the aesthetics, only _clean_ 1 and 0s.

Or we could have a #define for nibbles like #define Na 0x0a. That way we don't have to rely on the vagaries of individual compiler implementers and we have our own *Proper* way of doing things. 

But I'm being a bit facetious, I really don't think either of those would be good solutions. 

On topic, this rule enforcing syntax is just too much of a nitpick to be valuable. In general, whether small constants have leading zeros or not makes no difference at all to me. For nibbles though, I do like the single-digit constants as they do remind that there are only 4 usable bits there.

madmaxoft:
Oh well...

tigerw:
Well it's not in the style guide, so no-one can fault you for writing them however you want

PR #4646, opened 2020-04-12
Bindings: Allow coercion between Vector3 subtypes.

madmaxoft:
In manually bound functions, allows one to use any `Vector3<T>` value, as well as a `{x, y, z}` table, in Lua for any `Vector3<T>` parameter.
Has example in Debuggers' `/vector` command.
Unfortunately doesn't work in auto-bindings.

### DIFF
--- a/src/Bindings/BindingsProcessor.lua
+++ b/src/Bindings/BindingsProcessor.lua
@@ -74,7 +74,20 @@ local access =
 -- Map of classname -> true
 local g_HasCustomPushImplementation =
 {
-	cEntity = true
+	cEntity = true,
+}
+
+
+
+
+
+--- Defines classes that have a custom manual GetStackValue() implementation and should not generate the automatic one
+-- Map of classname -> true
+local g_HasCustomGetImplementation =
+{
+	Vector3d = true,
+	Vector3f = true,
+	Vector3i = true,
 }
 
 
@@ -209,13 +222,15 @@ local function OutputLuaStateHelpers(a_Package)
 		f:write("// This file expects to be included form inside the cLuaState class definition\n")
 		f:write("\n\n\n\n\n")
 		for _, item in ipairs(types) do
-			if not(g_HasCustomPushImplementation[item.name]) then
+			if not(g_HasCustomPushImplementation[item.lname]) then
 				f:write("void Push(" .. item.name .. " * a_Value);\n")
 			end
 		end
 		for _, item in ipairs(types) do
-			f:write("bool GetStackValue(int a_StackPos, Ptr" .. item.lname .. " & a_ReturnedVal);\n")
-			f:write("bool GetStackValue(int a_StackPos, ConstPtr" .. item.lname .. " & a_ReturnedVal);\n")
+			if not(g_HasCustomGetImplementation[item.lname]) then
+				f:write("bool GetStackValue(int a_StackPos, Ptr" .. item.lname .. " & a_ReturnedVal);\n")
+				f:write("bool GetStackValue(int a_StackPos, ConstPtr" .. item.lname .. " & a_ReturnedVal);\n")
+			end
 		end
 		f:write("\n\n\n\n\n")
 		f:close()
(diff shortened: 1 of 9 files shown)


src/Bindings/ManualBindings_BlockArea.cpp
```diff
@@ -48,14 +48,12 @@ static int readCuboidOverloadParams(cLuaState & a_LuaState, int a_StartParam, cC
 	else
 	{
 		// Assume the 2-Vector3i version:
-		Vector3i * p1;
-		Vector3i * p2;
-		if (!a_LuaState.GetStackValues(a_StartParam, p1, p2))
+		Vector3i p1;
+		Vector3i p2;
```
bearbin, line 52: These are no longer used?

madmaxoft: No, just temporaries

madmaxoft:
Unfortunately this doesn't want to build with Clang due to the template instantiation deferred to the cpp file; I'm not friends enough with templates to fix this. Perhaps @worktycho or @peterbell10 with their extensive knowledge could step in?

src/Bindings/ManualBindings_BlockArea.cpp
peterbell10: You use `a_Cuboid.p1` directly now, these variables are never used.

peterbell10:
It's only a warning, not a build error. Could just add a `-Wno-error=undefined-func-template` to the `CMakeLists.txt`.

Alternatively, you could have the `Vector3<lua_Number>` version in the cpp file and base the template off of that. With the full template definition in the header file.

madmaxoft:
Declaring the explicit instantiation as `extern` in the header seems to work as well.

madmaxoft:
Ready for review.
Now I only wish the same thing could somehow be done to the auto-bindings...

src/Bindings/BindingsProcessor.lua
```diff
@@ -231,38 +246,40 @@ local function OutputLuaStateHelpers(a_Package)
...
-			f:write("\t\treturn false;\n\t}\n")
-			f:write("\ttolua_Error err;\n")
-			f:write("\tif (tolua_isusertype(m_LuaState, a_StackPos, \"const " .. item.name .. "\", false, &err))\n")
-			f:write("\t{\n")
-			f:write("\t\ta_ReturnedVal = *(static_cast<const " .. item.name .. " **>(lua_touserdata(m_LuaState, a_StackPos)));\n")
-			f:write("\t\treturn true;\n");
-			f:write("\t}\n")
-			f:write("\treturn false;\n")
-			f:write("}\n\n\n\n\n\n")
+			if not(g_HasCustomGetImplementation[item.lname]) then
```
tigerw, line 256: Does Lua support inverting conditions?

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -604,21 +561,6 @@ void DeprecatedBindings::Bind(lua_State * tolua_S)
...
-	tolua_beginmodule(tolua_S, "Vector3f");
-		tolua_function(tolua_S,"abs",   tolua_Vector3_Abs<float>);
-		tolua_function(tolua_S,"clamp", tolua_Vector3_Clamp<float>);
-	tolua_endmodule(tolua_S);
-
-	tolua_beginmodule(tolua_S, "Vector3d");
-		tolua_function(tolua_S,"abs",   tolua_Vector3_Abs<double>);
-		tolua_function(tolua_S,"clamp", tolua_Vector3_Clamp<double>);
-	tolua_endmodule(tolua_S);
-
```
tigerw, line 621: yes, YES!

tigerw:
So am I correct in saying the new specialisations for GetStackValue are what enable both the coercion, and the treatment of Vector3 as not just an opaque pointer in the manual bindings?

src/Bindings/BindingsProcessor.lua
madmaxoft: What do you mean? Invert the condition and `continue`? Lua doesn't have a `continue`, unfortunately.

madmaxoft:
I don't quite understand what you mean, but the changes here allow you to use any kind of `Vector3` as a parameter to any function accepting any `Vector3`, as long as the function is bound manually. The example in Debuggers shows this. Nothing more, nothing less.

src/Bindings/BindingsProcessor.lua
tigerw: aw shucks

tigerw:
Just noting the Vector3 pointer + null check ↝ plain Vector3 changes were great for readability.

src/Bindings/ManualBindings.cpp
```diff
@@ -2975,8 +2975,8 @@ static int tolua_cLineBlockTracer_FirstSolidHitTrace(lua_State * tolua_S)
 		}
 		// Get the params:
 		cWorld * world;
-		Vector3d * start;
-		Vector3d * end;
+		Vector3d start;
+		Vector3d end;
 		if (!L.GetStackValues(idx, world, start, end))
```
peterbell10, line 2980: Above this is `L.IsParamUserType(idx + 1, "Vector3<double>")` and `L.CheckParamUserType(idx + 1, "Vector3<double>", idx + 2)`. I think you need a `CheckParamVector3`.

madmaxoft: What an unbelievable catch

peterbell10: You didn't address this fully. You didn't add a `CheckParamVector3` and if you search for `Vector3<double>` then you can see there is also an issue with `tolua_cLineBlockTracer_LineOfSightTrace`.

PR #4648, opened 2020-04-13
Moreplacementfixes

### LINKED ISSUE #4641: Cactus drops when placed on something else than sand or another cactus
Client version: 12.2.2
Server OS: Linux

### Expected behavior
Cactus would just not be placable. Players hand wouldn't even move.

### Actual behavior
The attempt is made to place the cactus and it just drops as a block.
I think this is _kind of_ similar to the issue with dropping levers, buttons, etc. on glass, lilypads, halfslabs, but it might be not, because the other pieces are not solid blocks.

Place a cactus e.g. on grass.

### Server log
```
-
```

Lazauya:
Lily pads should no longer be place-able on non-water or non-ice blocks; before, they had a "pop off" effect where they would appear for a second before being dropped as an item. Same fix for cactus, except on sand.

Fixes #4641

### DIFF
--- a/src/Defines.h
+++ b/src/Defines.h
@@ -4,6 +4,12 @@
 
 
 
+template <typename T> class Vector3;
+typedef Vector3i Vector3<int>;
+
+
+
+
 
 /** List of slot numbers, used for inventory-painting */
 typedef std::vector<int> cSlotNums;
@@ -424,6 +430,24 @@ extern eDamageType StringToDamageType(const AString & a_DamageString);
 /** Modifies the specified coords so that they point to the block adjacent to the one specified through its specified face.
 If a_Inverse is true, the opposite direction is used instead. */
 void AddFaceDirection(int & a_BlockX, int & a_BlockY, int & a_BlockZ, eBlockFace a_BlockFace, bool a_bInverse = false);
+
+/** a_CurrentPosiiton - the block of whom we would like to get the block next to
+a_BlockFace - the face (direction) in which we would like to get the block
+*/
+inline Vector3i GetBlockNextTo(Vector3i a_CurrentPosition, eBlockFace a_BlockFace)
+{
+	switch (a_BlockFace)
+	{
+		case BLOCK_FACE_XM: return Vector3i(a_CurrentPosition.x - 1, a_CurrentPosition.y, a_CurrentPosition.z);
+		case BLOCK_FACE_XP: return Vector3i(a_CurrentPosition.x + 1, a_CurrentPosition.y, a_CurrentPosition.z); 
+		case BLOCK_FACE_YM: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y - 1, a_CurrentPosition.z); 
+		case BLOCK_FACE_YP: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y + 1, a_CurrentPosition.z); 
+		case BLOCK_FACE_ZM: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y, a_CurrentPosition.z - 1); 
+		case BLOCK_FACE_ZP: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y, a_CurrentPosition.z + 1);
+		default: return a_CurrentPosition;
+	}
+}
+
 
 // tolua_end
 
--- a/src/Defines.h
+++ b/src/Defines.h
@@ -431,22 +431,10 @@ extern eDamageType StringToDamageType(const AString & a_DamageString);
 If a_Inverse is true, the opposite direction is used instead. */
 void AddFaceDirection(int & a_BlockX, int & a_BlockY, int & a_BlockZ, eBlockFace a_BlockFace, bool a_bInverse = false);
 
-/** a_CurrentPosiiton - the block of whom we would like to get the block next to
-a_BlockFace - the face (direction) in which we would like to get the block
-*/
-inline Vector3i GetBlockNextTo(Vector3i a_CurrentPosition, eBlockFace a_BlockFace)
-{
-	switch (a_BlockFace)
-	{
-		case BLOCK_FACE_XM: return Vector3i(a_CurrentPosition.x - 1, a_CurrentPosition.y, a_CurrentPosition.z);
-		case BLOCK_FACE_XP: return Vector3i(a_CurrentPosition.x + 1, a_CurrentPosition.y, a_CurrentPosition.z); 
-		case BLOCK_FACE_YM: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y - 1, a_CurrentPosition.z); 
-		case BLOCK_FACE_YP: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y + 1, a_CurrentPosition.z); 
-		case BLOCK_FACE_ZM: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y, a_CurrentPosition.z - 1); 
-		case BLOCK_FACE_ZP: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y, a_CurrentPosition.z + 1);
-		default: return a_CurrentPosition;
-	}
-}
+/** Same as AddFaceDirection, but contstructs new coordinates and uses vectors
+a_CurrentPosiiton - the block of whom we would like to get the block next to
+a_BlockFace - the face (direction) in which we would like to get the block */
+Vector3i GetBlockNextTo(Vector3i a_CurrentPosition, eBlockFace a_BlockFace);
 
 
 // tolua_end
--- a/src/Defines.h
+++ b/src/Defines.h
@@ -5,7 +5,7 @@
 
 
 template <typename T> class Vector3;
-typedef Vector3i Vector3<int>;
+typedef Vector3<int> Vector3i;
 
 
 
--- a/src/Items/ItemCactus.h
+++ b/src/Items/ItemCactus.h
@@ -8,6 +8,7 @@
 class cItemCactusHandler :
 	public cItemHandler
 {
+	using Super = cItemHandler;
 public:
 	cItemCactusHandler(int a_ItemType) :
 		cItemHandler(a_ItemType)
@@ -22,24 +23,22 @@ public:
 
 
 	virtual bool OnPlayerPlace(
-        cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
-        int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
-        int a_CursorX, int a_CursorY, int a_CursorZ
-    ) 
+		cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
+		int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
+		int a_CursorX, int a_CursorY, int a_CursorZ
+	)
 	{
-		/** We will use the block face and the block coordinates to determine if we can place the cactus
-		There must be sand or cactus under the cactus and no blocks on the sides inorder for it to be placed
-		*/
+		// We will use the block face and the block coordinates to determine if we can place the cactus
+		// There must be sand or cactus under the cactus and no blocks on the sides inorder for it to be placed
 		Vector3i PlacementPos = GetBlockNextTo(Vector3i(a_BlockX, a_BlockY, a_BlockZ), a_BlockFace);
 
 		BLOCKTYPE Surface = a_World.GetBlock(PlacementPos - Vector3i(0, 1, 0));
 		if ((Surface != E_BLOCK_SAND) && (Surface != E_BLOCK_CACTUS))
 		{
-			// Cactus can only be placed on sand and itself
 			return false;
 		}
 
-		const Vector3i Offsets[4] = 
+		const Vector3i Offsets[4] =
 		{
 			{1, 0, 0},
 			{-1, 0, 0},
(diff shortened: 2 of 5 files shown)


src/Defines.h
```diff
@@ -425,6 +431,24 @@ extern eDamageType StringToDamageType(const AString & a_DamageString);
...
+	{
+		case BLOCK_FACE_XM: return Vector3i(a_CurrentPosition.x - 1, a_CurrentPosition.y, a_CurrentPosition.z);
+		case BLOCK_FACE_XP: return Vector3i(a_CurrentPosition.x + 1, a_CurrentPosition.y, a_CurrentPosition.z); 
+		case BLOCK_FACE_YM: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y - 1, a_CurrentPosition.z); 
+		case BLOCK_FACE_YP: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y + 1, a_CurrentPosition.z); 
+		case BLOCK_FACE_ZM: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y, a_CurrentPosition.z - 1); 
+		case BLOCK_FACE_ZP: return Vector3i(a_CurrentPosition.x, a_CurrentPosition.y, a_CurrentPosition.z + 1);
+		default: return a_CurrentPosition;
+	}
+}
```
madmaxoft, line 449: Put the function implementation into the cpp file instead.
The doxy comment should first say what the function does.

src/Items/ItemCactus.h
```diff
@@ -0,0 +1,68 @@
...
+	{
+		return true;
+	}
+
+
+	virtual bool OnPlayerPlace(
+        cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
+        int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
+        int a_CursorX, int a_CursorY, int a_CursorZ
+    )
```
madmaxoft, line 28: Use tabs for indent

src/Items/ItemCactus.h
```diff
@@ -0,0 +1,68 @@
...
+	virtual bool OnPlayerPlace(
+        cWorld & a_World, cPlayer & a_Player, const cItem & a_EquippedItem,
+        int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace,
+        int a_CursorX, int a_CursorY, int a_CursorZ
+    ) 
+	{
+		/** We will use the block face and the block coordinates to determine if we can place the cactus
+		There must be sand or cactus under the cactus and no blocks on the sides inorder for it to be placed
+		*/
+		Vector3i PlacementPos = GetBlockNextTo(Vector3i(a_BlockX, a_BlockY, a_BlockZ), a_BlockFace);
```
madmaxoft, line 33: Prefer to use `auto` as much as possible.

madmaxoft, line 33: Do not use doxy-comments inside functions, use regular comment here instead.

src/Items/ItemCactus.h
```diff
@@ -0,0 +1,68 @@
...
+		for(int i = 0; i < ARRAYCOUNT(Offsets); i++) 
+		{
+			BLOCKTYPE Surrounding = a_World.GetBlock(PlacementPos + Offsets[i]);
+			if(cBlockInfo::IsSolid(Surrounding) || IsBlockLava(Surrounding))
+			{
+				return false;
+			}
+		}
+
+		return cItemHandler::OnPlayerPlace(a_World, a_Player, a_EquippedItem,
```
madmaxoft, line 59: Add `using Super = cItemHandler` to the top of this class (`private`) and use `Super` instead of the specific base. This way if we ever need to change the base (which does happen rather often), we won't need to touch this code.

src/Items/ItemHandler.cpp
```diff
@@ -216,6 +217,11 @@ cItemHandler * cItemHandler::CreateItemHandler(int a_ItemType)
 		}
 
 		case E_ITEM_BEETROOT_SEEDS:
+		case E_BLOCK_CACTUS:
+		{
+			return new cItemCactusHandler(a_ItemType);
```
madmaxoft, line 222: This causes beetroot seeds to have a Cactus handler as well, is that intended?

src/Items/ItemLilypad.h
```diff
@@ -27,23 +26,18 @@ class cItemLilypadHandler :
...
-
 	virtual bool OnItemUse(
-		cWorld * a_World, cPlayer * a_Player, cBlockPluginInterface & a_PluginInterface, const cItem & a_Item,
-		int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace
-	) override
+        cWorld * a_World, cPlayer * a_Player, cBlockPluginInterface & a_PluginInterface, const cItem & a_Item,
+        int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace
+    ) override
 	{
+		LOGD(BlockFaceToString(a_BlockFace).c_str());
```
madmaxoft, line 34: Leftover logging.

src/Items/ItemLilypad.h
```diff
@@ -57,51 +51,146 @@ class cItemLilypadHandler :
...
-						cBlockInfo::FullyOccupiesVoxel(a_CBBlockType)
-						)
-					{
-						// Can't place lilypad on air / in another block!
-						return true;
-					}
+			{	
+				if((((a_CBBlockType == E_BLOCK_STATIONARY_WATER) || (a_CBBlockType == E_BLOCK_WATER)) && (a_CBBlockMeta == 0))
+				|| (a_CBBlockType == E_BLOCK_ICE) || (a_CBBlockType == E_BLOCK_FROSTED_ICE))
+				{
```
madmaxoft, line 57: This is now an unreadable mess. With a condition this long, split it onto multiple lines with *each* of the same-level conditions on a separate lines.

What's wrong with `IsBlockWater`?

src/Items/ItemLilypad.h
```diff
@@ -57,51 +51,146 @@ class cItemLilypadHandler :
...
+			return false;
+		}
+		
+		// Since we hit fluid, try placing the block on top of the fluid we found
+		// Mimick clicking on top of the fluid to do this
+		return OnPlayerPlace(*a_World, *a_Player, a_Item,
+			Callbacks.m_Pos.x, Callbacks.m_Pos.y, Callbacks.m_Pos.z, Callbacks.m_BlockFace,
+			0, 0, 0
+		);
+	}
```
madmaxoft, line 90: The entire `OnItemUse` function is now a big ball of wire with only too-detailed-comments sprinkled in it, with no overview. I assume it works by first tracing the line-of-sight, then if a water block is hit, relaying to `OnPlayerPlace`. If so, split the line-of-sight tracing into a separate function so that the `OnItemUse` is kept short and concise.

src/Items/ItemCactus.h
Lazauya: Is this true? I haven't seen auto used anywhere else

src/Defines.h
```diff
@@ -425,6 +431,24 @@ extern eDamageType StringToDamageType(const AString & a_DamageString);
 If a_Inverse is true, the opposite direction is used instead. */
 void AddFaceDirection(int & a_BlockX, int & a_BlockY, int & a_BlockZ, eBlockFace a_BlockFace, bool a_bInverse = false);
 
+/** a_CurrentPosiiton - the block of whom we would like to get the block next to
+a_BlockFace - the face (direction) in which we would like to get the block
+*/
+inline Vector3i GetBlockNextTo(Vector3i a_CurrentPosition, eBlockFace a_BlockFace)
```
peterbell10, line 437: Is this not the same as the function above: `AddFaceDirection`?

src/Defines.h
```diff
@@ -4,6 +4,12 @@
 
 
 
+template <typename T> class Vector3;
+typedef Vector3<int> Vector3i;
```
peterbell10, line 8: This shouldn't be necessary, `Vector3` is included globally.

src/Items/ItemLilypad.h
```diff
@@ -57,51 +51,146 @@ class cItemLilypadHandler :
...
+		NIBBLETYPE SurfaceMeta = a_World.GetBlockMeta(PlacementPos - Vector3i(0, 1, 0));
+		if(!((((Surface == E_BLOCK_STATIONARY_WATER) || (Surface == E_BLOCK_WATER)) && (SurfaceMeta == 0)) ||  // A water source is below
+			(Surface == E_BLOCK_ICE) || (Surface == E_BLOCK_FROSTED_ICE)))                                     // Or (frosted) ice
 		{
-			a_World->SetBlock(Callbacks.m_Pos.x, Callbacks.m_Pos.y, Callbacks.m_Pos.z, E_BLOCK_LILY_PAD, 0);
-			if (!a_Player->IsGameModeCreative())
+			return false;
+		}
+
+		// Otherwise do the same exact process as cItemHandler::OnPlayerPlace, but ignore build collision
```
peterbell10, line 116: Why not remove the block then call `OnPlayerPlace`?

Lazauya: Remove which block? The lily pad hasn't actually been placed at this point

src/Defines.h
Lazauya: Similar, but more convenient and using vectors. I would go as far to say that AddFaceDirection should be removed entirely in the future (when we completely switch over to Vector3). AddFaceDirection could be argued to be slightly faster, but if you have to do the inverse at some point I think that's moot.

src/Items/ItemHandler.cpp
Lazauya: Nope, thanks

src/Defines.h
peterbell10: Yes, but why call it `GetBlockNextTo` and have less features when it could just be:
```cpp
inline Vector3i AddFaceDirection(Vector3i a_BlockPos, eBlockFace a_BlockFace, bool a_bInverse = false)
{
	AddFaceDirection(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, a_BlockFace, a_bInverse);
	return a_BlockPos;
}
```

src/Items/ItemLilypad.h
Lazauya: I can try IsBlockWater, but the old lily pad code wasn't using it, and I assumed it was for a good reason.

src/Items/ItemLilypad.h
peterbell10: Build collision implies it is overwriting a block already in that place.

Lazauya: Maybe my comment as confusing, sorry, I just meant place it above the water source regardless of the build collision of water. My bad

src/Items/ItemCactus.h
madmaxoft: It's an old codebase, so it's not used everywhere, but we try to use it in all new code.

src/Items/ItemLilypad.h
madmaxoft: B... b... but it's right there, line 61 of the original code.

src/Defines.h
Lazauya: Makes sense, seems like a sensible change

madmaxoft:
The basic style check is still failing, you'll need to fix those as well: https://circleci.com/gh/cuberite/cuberite/8905

src/Items/ItemLilypad.h
peterbell10: Okay, my main concern is that you're copy-pasting fairly complex code.

src/Items/ItemLilypad.h
```diff
@@ -57,51 +51,146 @@ class cItemLilypadHandler :
...
+			return false;
+		}
+	
+		BLOCKTYPE ClickedBlock;
+		NIBBLETYPE ClickedBlockMeta;
+		a_World.GetBlockTypeMeta(absPos.x, absPos.y, absPos.z, ClickedBlock, ClickedBlockMeta);
+
+		cChunkInterface ChunkInterface(a_World.GetChunkMap());
+		
+		absPos = GetBlockNextTo(absPos, a_BlockFace);
```
peterbell10, line 138: This is `PlacementPos` above.

src/Items/ItemLilypad.h
```diff
@@ -57,51 +51,146 @@ class cItemLilypadHandler :
...
+	
+		if (!cChunkDef::IsValidHeight(absPos.y))
+		{
+			// The clicked block is outside the world, ignore this call altogether (#128)
+			return false;
+		}
+	
+		BLOCKTYPE ClickedBlock;
+		NIBBLETYPE ClickedBlockMeta;
+		a_World.GetBlockTypeMeta(absPos.x, absPos.y, absPos.z, ClickedBlock, ClickedBlockMeta);
```
peterbell10, line 134: This isn't used anywhere...

src/Items/ItemLilypad.h
```diff
@@ -57,51 +51,146 @@ class cItemLilypadHandler :
...
+			// Tried to place a block into another?
+			// Happens when you place a block aiming at side of block with a torch on it or stem beside it
+			return false;
+		}
+
+		a_World.DropBlockAsPickups(absPos, &a_Player, nullptr);
+	
+		// Get all the blocks to place:
+		sSetBlockVector blocks;
+		if (!GetBlocksToPlace(a_World, a_Player, a_EquippedItem, absPos.x, absPos.y, absPos.z, a_BlockFace, a_CursorX, a_CursorY, a_CursorZ, blocks))
```
peterbell10, line 162: No point in calling `GetBlocksToPlace`. We know it will be a lily pad.

Just call `a_Player.PlaceBlock(...)` directly.

src/Defines.h
madmaxoft: I've extracted this to a separate PR, because I'd like to use that in my own PR soon.
#4666 .

src/Items/ItemLilypad.h
Lazauya: That seems too hard-coded?

peterbell10: An item handler's entire purpose is to "hard-code" the behaviour of that one specific item. Making it more complicated than necessary just makes it harder to understand the code when coming back to it later.

tigerw:
Should be fixed as part of #5184, many thanks @Lazauya.

PR #4650, opened 2020-04-13
Using Super

madmaxoft:
Changed all `typedef cXYZ super` to `using Super = cXYZ`.
Minor whitespace cleanup.
~Removed `cPluginLua` from Lua API, since the only function it exported was the obsolete `cPluginLua:AddWebTab`; the exported function was kept through `cPlugin` and marked as deprecated properly.~


tigerw:
xoft imposes his will

madmaxoft:
As a self-proclaimed dictator of this project

madmaxoft:
@bearbin continuing the discussion here.
In my opinion `Super` is equivalent to any class - it has functions, member variables etc. A class doesn't need a file with the same name; there's no file `MineShaftDirtRoom` or `Piece::Connector`, and in the same way there's no file for `Super`. It is very close to class-in-class in this regard.

So shall we call it `cSuper`? :P

madmaxoft:
BTW C++ cannot provide a true `super` keyword, because of multiple inheritance. Only an approximation, as available in MSVC's `__super` extension: https://docs.microsoft.com/en-us/cpp/cpp/super?view=vs-2019

bearbin:
I guess my objection is that `Super` means something different in every class, and it's confusing (for new developers) to think there is some single `Super` class or macro that does magic, rather than a simple `using` statement. 

It really isn't that bad though, more important that everything is consistent throughout the codebase.

Server/Plugins/APIDump/Classes/Plugins.lua
```diff
@@ -5,6 +5,21 @@ return
 		Desc = "cPlugin describes a Lua plugin. Each plugin has its own cPlugin object.",
 		Functions =
 		{
+			AddWebTab =
```
peterbell10, line 8: Wrong branch?

madmaxoft: Not really, I removed it at first because tolua wouldn't parse the `using` clause; by the time I found more of those, it was too late to go back. I don't want to change this PR any more, it's huge as it is already.

peterbell10: >I don't want to change this PR any more, it's huge as it is already.

That's exactly why it shouldn't contain unrelated changes. Reviewing the diff would be a nightmare.

madmaxoft:
PR cleaned up, no more Lua API changes..

tigerw:
Well I don't agree in principle but if you insist you'd better merge quickly before conflicts pop up.

madmaxoft:
Is anyone actually reviewing this, or should I just merge? :P

bearbin:
On Thu, 16 Apr 2020 07:07:25 -0700
Mattes D <notifications@github.com> wrote:

> Is anyone actually reviewing this, or should I just merge? :P
> 

Not reviewing ATM, but I can give it a look tonight for any functional
changes that might have slipped in by accident.

peterbell10:
I skimmed through and couldn't see any obvious errors. With such a huge diff I couldn't go through line by line, so can't be entirely sure.

bearbin:
Merging, no conflicts luckily.

src/Entities/ExpOrb.h
```diff
@@ -8,14 +8,15 @@
...
 	public cEntity
 {
-	typedef cEntity super;
-
-public:
 	// tolua_end
 
+	using Super = cEntity;
+
+public:  // tolua_export
```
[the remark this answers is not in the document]
tigerw: Goods and services to the European Union.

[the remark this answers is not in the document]
peterbell10: tolua++ completely ignores anything not in the exported segments (as if it were `ifdef`'d out). So, without this, it would completely ignore the public specifier.

PR #4653, opened 2020-04-15
ServerHandle: Fix closing failed servers.

madmaxoft:
If the cServerHandle failed to listen, closing it would then crash Cuberite.

### DIFF
--- a/src/OSSupport/ServerHandleImpl.cpp
+++ b/src/OSSupport/ServerHandleImpl.cpp
@@ -63,10 +63,15 @@ cServerHandleImpl::~cServerHandleImpl()
 void cServerHandleImpl::Close(void)
 {
 	// Stop the listener sockets:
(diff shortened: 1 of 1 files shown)


src/OSSupport/ServerHandleImpl.cpp
```diff
@@ -63,10 +63,15 @@ cServerHandleImpl::~cServerHandleImpl()
 void cServerHandleImpl::Close(void)
 {
 	// Stop the listener sockets:
-	evconnlistener_disable(m_ConnListener);
+	if (m_ConnListener != nullptr)
+	{
+		evconnlistener_disable(m_ConnListener);
+		m_ConnListener = nullptr;
```
madmaxoft, line 69: This is wrong, it causes the socket to not be closed when the server is finally destroyed.

PR #4658, opened 2020-04-15
Prevent crash when breeding

mathiascode:
No need to broadcast entity metadata here, since the sheep hasn't spawned yet and its world isn't set at this stage.


tigerw:
perchance related to #4606 😅

mathiascode:
@tigerw Yep

madmaxoft:
Does the fur color get to the client later on, then?

bearbin:
@madmaxoft  Entity metadata is written when a mob spawn packet is sent, so yes.

PR #4638, opened 2020-04-10
ಠ_ಠ

tigerw:
OBSERVE

PR #4661, opened 2020-04-16
Implement glowing redstone ore

peterbell10:
While testing #4638 I noticed that clicking on redstone ore wasn't triggering the observer.


tigerw:
Confirm that it works on right click and punch + ಠ_ಠ gets an update.

I think in vanilla jumping around on it also triggers the effect but we can just say our ore is a more inert and stable form.

bearbin:
Our redstone simulator is free of such misguided behaviour as unauthorised BUDs, so too should our redstone ore not generate so many BUs.

Pokechu22:
In vanilla, an entity walking on redstone ore activates it (not just jumping), in addition to left and right clicking.  I assume it's not that hard to detect walking but I'm no longer familiar with the codebase.

bearbin:
Does dropping an item on redstone ore also activate it? Shooting an arrow? If both then it could just be any entity touching it.

peterbell10:
Arrows and items don't trigger it but mobs do.

Pokechu22:
[The wiki says](https://minecraft.gamepedia.com/Redstone_Ore#Light) it happens for mobs other than sneaking players and spiders; to my understanding it specifically happens when the entity/the block would make a step sound.  I'm not sure if step sounds are implemented in cuberite though.

PR #4670, opened 2020-04-17
buttons can now be triggered by arrows.

### LINKED ISSUE #4640: Feature request: Trigger button with arrow
Let a button be pressed when an arrow lands on it until it is removed.

GefaketHD:
Arrows can now trigger buttons.

fixes #4640


src/Blocks/BlockButton.h
```diff
@@ -45,18 +46,10 @@ class cBlockButtonHandler :
...
-					// Block hasn't change in the meantime; set its meta
-					a_World.SetBlockMeta(Pos.x, Pos.y, Pos.z, a_World.GetBlockMeta(Pos) & 0x07, false);
-					a_World.WakeUpSimulators(Pos);
-					a_World.BroadcastSoundEffect("block.stone_button.click_off", SoundPos, 0.5f, 0.5f);
-				}
-			}
-		);
+		auto TickDelay = (m_BlockType == E_BLOCK_STONE_BUTTON) ? 10 : 15;
+
+		auto ChunkData = static_cast<cIncrementalRedstoneSimulator *>(a_Player.GetWorld()->GetRedstoneSimulator())->GetChunkData();
```
peterbell10, line 51: You can't assume this is an instance of `cIncrementalRedstoneSimualtor`. It could be `cRedstoneNoopSimulator` instead, for example.

GefaketHD: True...but it is done many more times in the project.

GefaketHD: What is the 'cRedstoneNoopSimulator' used for?

peterbell10: >True...but it is done many more times in the project.

It's used inside `cIncrementalRedstoneSimualtor` itself, where the fact that it's being called at all lets it know that it's the active simulator. So in there, it's always valid.

>What is the 'cRedstoneNoopSimulator' used for?

It replaces the redstone simulator with "no operation", i.e. it disables redstone entirely. That can help with performance or debugging for example.

GefaketHD: I see...so if redstone is disabled entirely then, I can just dynamic cast it and use the MechanismDelays like I need/wanted to. Because if redstone is disabled, buttons don't need to get triggered at all - right? :)

GefaketHD: I could use the version of handling from before I changed it for the noop-simulator...to visually press the button, right?

Pokechu22:
According to [the wiki](https://minecraft.gamepedia.com/Button#Redstone_power) (and my memory), only wooden buttons are affected by arrows, not stone buttons.

src/Blocks/BlockButton.h
peterbell10: Could you move all delay logic into the redstone handler?

GefaketHD: The way I tried to access the redstone handler from the OnUse-function is not correct like I did it. Can you give me a hint?

peterbell10: Could you just set the button on and let the redstone simulator update do the rest?

@tigerw would know best. But I think when you find the button on and find  `DelayInfo == nullptr` then you should set the initial delay there.

GefaketHD: Okay

GefaketHD: But I will (have to) distinquish between noop and incremental mode - because in noop the button would not come back, if the update would handle the delays

GefaketHD: Which means I could just set the delay initially on button press... @tigerw - what would be best here? I will push and show what I mean with "distinquish"

GefaketHD:
Somehow the stone button is making 2 click sounds when pressed by hand now - couldn't find out, why - its 3:30 here now. I will go sleep and give it a shot again tomorrow.

GefaketHD:
Could $someone test this please and tell me, if it is a problem of my code, that the the stone button clicks twice and the wooden button has a weird *bump* and then a click, when I press them?

tigerw:
Testing.

src/Entities/ArrowEntity.cpp
```diff
@@ -85,6 +85,7 @@ void cArrowEntity::OnHitSolidBlock(Vector3d a_HitPos, eBlockFace a_HitFace)
 
 	// Broadcast arrow hit sound
 	m_World->BroadcastSoundEffect("entity.arrow.hit", BlockHit, 0.5f, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
+	m_World->WakeUpSimulators(m_HitBlockPos);
```
bearbin, line 88: Why is this necessary? Add a comment maybe.

src/Simulator/IncrementalRedstoneSimulator/RedstoneToggleHandler.h
```diff
@@ -77,14 +77,11 @@ class cRedstoneToggleHandler:
...
-		UNUSED(a_World);
-		UNUSED(a_Position);
-
 		switch (a_BlockType)
 		{
 			case E_BLOCK_LEVER: return cBlockLeverHandler::IsLeverOn(a_Meta) ? 15 : 0;
-			case E_BLOCK_STONE_BUTTON:
-			case E_BLOCK_WOODEN_BUTTON: return cBlockButtonHandler::IsButtonOn(a_Meta) ? 15 : 0;
+			case E_BLOCK_STONE_BUTTON: return cBlockButtonHandler::IsButtonOn(a_Meta) ? 15 : 0;
+			case E_BLOCK_WOODEN_BUTTON: return cBlockButtonHandler::IsButtonOn(a_World, a_Position, a_Meta) ? 15 : 0;
```
bearbin, line 84: This is very ugly, this code gives the impression that stone buttons can be on but not give out any redstone signal.

tigerw:
I think the client (1.12) plays a sound all by itself when you press a button. I can comment out everywhere where we send sounds and the "thump" still plays. All we need to do is to play the right sound (we're always doing stone_button right now) only on depress.

I'm sure we can implement this without touching the redstone simulator at all. I propose that arrow entities should be responsible for setting the button meta if they hit a button, and unsetting when they are destroyed or picked up. That way the simulator can be left as-is, only concerning itself about delivering power when the button meta is depressed.

peterbell10:
>I think the client (1.12) plays a sound all by itself when you press a button.

We still might need to send it to nearby players. In which case use `a_Exclude` to exclude the client that pressed the button:
https://github.com/cuberite/cuberite/blob/ee8419701472be9f24c51a41ee7b7ef3cf38f329/src/Blocks/BroadcastInterface.h#L60

src/Blocks/BlockButton.h
```diff
@@ -133,9 +141,77 @@ class cBlockButtonHandler :
...
+		bool FoundArrow = false;
+		a_World.ForEachEntityInBox(cBoundingBox(faceOffset + a_Position, 0.2, 0.2), [&](cEntity & a_Entity)
+			{
+				if (a_Entity.IsArrow())
+				{
+					FoundArrow = true;
+				}
+				return false;
+			}
+		);
```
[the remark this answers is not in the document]
GefaketHD: Did you test it already?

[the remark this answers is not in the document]
bearbin: No, just wrote it according to the documentation. If it doesn't work, what would definitely work would be to `return true;` after `FoundArrow = true;`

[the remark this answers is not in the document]
GefaketHD: Ahh ok, I see. Should work :) Thanks!

src/Entities/ArrowEntity.cpp
GefaketHD: Done.

src/Blocks/BlockButton.h
```diff
@@ -133,9 +141,77 @@ class cBlockButtonHandler :
...
+		{
+			a_World.SetBlockMeta(a_Position, a_World.GetBlockMeta(a_Position) | 0x08, false);
+			a_World.WakeUpSimulators(a_Position);
+			a_World.GetBroadcastManager().BroadcastSoundEffect("block.stone_button.click_on", a_Position, 0.5f, 0.6f);
+			return true;
+		}
+		return false;
+	}
+
+	static bool HasArrowInIt(cWorld & a_World, Vector3i a_Position, NIBBLETYPE a_Meta)
```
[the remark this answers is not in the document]
GefaketHD: Would `HandleArrowInIt` be a better name?

src/Simulator/IncrementalRedstoneSimulator/RedstoneToggleHandler.h
GefaketHD: To be honest, I don't see, why this code would say something like this, but I agree that there might be better ways to handle wodden and stone buttons differently. Just didn't want to give the blocktype to `IsButtonOn` too...

GefaketHD: I will push and you say, if it is better now - it is more like before...but only one function

src/Blocks/BlockButton.h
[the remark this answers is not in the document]
GefaketHD: Please check up on it now. :)

GefaketHD:
> I think the client (1.12) plays a sound all by itself when you press a button. I can comment out everywhere where we send sounds and the "thump" still plays. All we need to do is to play the right sound (we're always doing stone_button right now) only on depress.
> 
> I'm sure we can implement this without touching the redstone simulator at all. I propose that arrow entities should be responsible for setting the button meta if they hit a button, and unsetting when they are destroyed or picked up. That way the simulator can be left as-is, only concerning itself about delivering power when the button meta is depressed.

Sound is fixed.

And I was not able to let the arrow unset the meta if it is picked up, because if there are more arrows in it, the button should remain pressed and the check would have to be executed anyway. That's why I thought it is a better way to let it be checked in the Update-function. But I am too new to the project to decide something like this, so if you want me to change it, I can try.

GefaketHD:
> > I think the client (1.12) plays a sound all by itself when you press a button.
> 
> We still might need to send it to nearby players. In which case use `a_Exclude` to exclude the client that pressed the button:
> 
> https://github.com/cuberite/cuberite/blob/ee8419701472be9f24c51a41ee7b7ef3cf38f329/src/Blocks/BroadcastInterface.h#L60

I did that and tested it with my brother. It's working. :)

src/Blocks/BlockButton.h
```diff
@@ -133,9 +141,77 @@ class cBlockButtonHandler :
...
+	{
+		if (a_Meta & 0x08)
+		{
+			return true;
+		}
+		return false;
+	}
+
+	/** Extracts the ON bit from metadata and returns if true if it is set and checks for arrows in the button */
+	static bool IsButtonOn(cWorld & a_World, Vector3i a_Position, NIBBLETYPE a_Meta)
```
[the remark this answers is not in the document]
GefaketHD: I couldn't think of a better name by the way - Or it would have gotten incredibly long... :-1:
I am not very creative with names.

src/Simulator/IncrementalRedstoneSimulator/RedstoneToggleHandler.h
bearbin: I can't see anything new having been pushed. The issue is the difference between IsButtonOn(meta) and IsButtonOn(meta, position). Why should changing position affect button on-ness. I think passing the BlockType in does make more sense.

GefaketHD: Ah I understand, okay :)
There should some stuff be pushed though :x

src/Blocks/BlockButton.h
```diff
@@ -133,12 +145,79 @@ class cBlockButtonHandler :
...
+		if (a_Meta & 0x08)
+		{
+			return true;
+		}
+		return false;
 	}
-} ;
 
+	/** Returns true if the button is on and handles arrows stuck in a wooden button */
+	static bool IsButtonOn(cWorld & a_World, Vector3i a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_Meta)
```
bearbin, line 158: `CheckAndUpdateButtonState()` ?

madmaxoft:
I think this is an altogether wrong way to go about this.
A better solution would be to actually edit the arrow entity - when it hits a wooden button, then it should activate the button (by setting its meta); nothing else.
The block handler should definitely NOT touch anything in the redstone handlers.

GefaketHD:
> I think this is an altogether wrong way to go about this.
> A better solution would be to actually edit the arrow entity - when it hits a wooden button, then it should activate the button (by setting its meta); nothing else.
> The block handler should definitely NOT touch anything in the redstone handlers.

As I said: It is way harder to let the meta be set *correctly* by the arrow upon hitting a button, because afaik the button is not a solid block, so the arrow doesn't really hit the button block itself but the block the button is attached to. And if you collect or despawn the arrow, the place where the button is at has to be checked for other arrows, so it would just move something we do already in the pressure plate handler (checking for entities nearby) to the destroy-process of the arrow.

madmaxoft:
Upon hitting a block, you get the block pos and the block face. Checking if there's a button attached to that block face is a 2-line operation - `AddFaceDirection`, `GetBlockType`.
It's about the concept. We don't want buttons to know about arrows, we want arrows to know about buttons.
If the thing is already implemented in pressure plates, then even more so it should be unified into "arrow activates these blocks", rather than "these blocks each implement their own arrow detection".

madmaxoft:
So, how I imagine this to work:
- `cArrowEntity` upon hit checks if it hit a button or a pressure plate; if so, it activates them and schedules deactivation for when it despawns
- `cButtonHandler` refactors the activation into a separate function that accepts the time-to-deactivation as param
- The scheduled deactivation task always checks all entities at the pos, if there are any arrows, it just bails out.

Should be simple enough and cover most use-cases. It does not cover when the arrow is destroyed in another way than natural despawn, but that can be fixed subsequentially, for example by the arrow's despawn scheduling the (very same) deactivation task.

tigerw:
However, not having one source (the simulator, since buttons can't have state) be the sole thing responsible for scheduling tasks quickly lead to complicated edge-case handling:

1. You need a way to maintain that there is only ever one task scheduled to depress. Say a user presses the button (and it queues a depress), fires an arrow, immediately collects the arrow (and it queues another task). Now there are two tasks in flight.
2. How to solve this? You could say only one entity is responsible for depress. So now when arrows hit you add code to track which cArrowEntity has a "responsible" flag. When an arrow despawns it must pass this flag to the next arrow, if any.
3. What if the arrow doesn't despawn? Say it falls out the block or a plugin moves it. To avoid holding the flag hostage (and button forever depressed) now we need more logic to reset.

It's definitely a pain.

tigerw:
oh! oh wait, Vanilla appears to just run the scheduled task constantly while the button is held down. e.g. arrows don't consistently delay x ticks when collected off a button, but vary from anything between 0 - x ticks depending on how long away the next task happens to be. Plus, vanilla also doesn't seem to handle the case where you press a button, break it, place a new button, and press it, i.e. the task isn't cancelled when the button is broken which means you can end up with a 1 tick button press when the old depress task fires.

So we can simply use the "is depressed" meta bit as an indicator that there's already a scheduled task in flight, sidestepping everything.

GefaketHD:
@madmaxoft 
I didn't and don't want to let arrows know about buttons. Because on pressure plates, every item laying on them can trigger them - golden ones e.g.
So imo it is more logical to let the pressure plates check if there is anything lying on them, that could trigger them, than letting the items reach out to the plate and add or subtract themselves of a list. It's not only arrows.

The way you explained, how it should work, makes the arrow reach into the redstone simulator more than my version of code. Because I only wake up the simulators and they search for the arrows themselves. What is wrong with that?

@tigerw 
I am sorry that I don't think I got what you were trying to say...In the code version I provided, only the redstone simulator is scheduling tasks, except it is a noop-one, then the visual depressing is scheduled like every normal button press is scheduled in master right now.

And the flag being passed around is not only very likely to not cover *all* edge-cases but at least as hard to debug. And way harder to implement than just using the already existing scheduler in the redstone simulator.

tigerw:
I took the liberty of adding some commits for your consideration, hopefully it demonstrates what I was trying to say.

src/Blocks/BlockButton.h
```diff
@@ -133,12 +123,110 @@ class cBlockButtonHandler :
...
+		a_World.ScheduleTask(
+			TickDelay,
+			[a_Position, a_BlockType](cWorld & a_World)
+			{
+				BLOCKTYPE Type;
+				NIBBLETYPE Meta;
+
+				if (
+					!a_World.GetBlockTypeMeta(a_Position, Type, Meta) ||
+					(Type != a_BlockType) ||
```
bearbin, line 177: So when the button has been mined away this scheduled task will run continually?

tigerw: good point

GefaketHD:
One more time I want to ask: Why is it more logical in your all opinion to save in the arrow if it hit a button?
There are so many cases you can simply not cover with that implementation, because you can't even know them yet and it needs way more CPU power to calculate if there are whatever how many arrows still in the block after one despawns or gets destroyed or collected etc. and I think there will be bugs with that kind of implementation and it just gets more confusing like this.
In my opinion it is way more logical to just have the button check, if it should be on instead of trying to think of anything that could trigger a button or release it...?

I mean at the end of the day it's not my repo and I can't review my own changes, but I feel like you just don't want to let it be done like I did it for whatever reason-the arrows triggering a button is an edge-case, nothing else does this. So why not just let the hit button be checked in the redstone simulator cycle right after the arrow hit it, but implement a whole new and probably buggy feature and let the `Update`-Function of the buttons be unused. ? EVERY other mechanism uses its `Update`-function to determine if it is powered or not - why should the button work differently?

tigerw:
> save in the arrow

(I don't think the arrow saves anything, it only calls into a static method in the button handler on hit.)

> cases you can simply not cover

Which ones were you thinking of, this will be important to know.

-----

At least my thinking for this was:

Perhaps the reason is less technical and more that we're aiming for something easier to maintain, for both old, and new contributors. At least one other person (xoft) thinks the redstone simulator is a total black box/mystery, and I'm sure tightly coupling the button block handler and the simulator by modifying m_MechanismDelays would be too hard to for him to stomach :P

Given that player pushing buttons/arrows pushing buttons is more of an externality, you could reasonably say that it's not of the IncrementalSimulator's concern to react to these events, but instead only handle the redstone power aspect.

Plus, it would be nice for arrows to work even with NoopRedstoneSimulator. 

Sure, queuing tasks indefinitely isn't as good, but you know the saying, programmer time is more expensive than CPU time :)

And once again, I agree: conceptually, it makes total sense for the **button** to be responsible for checks. But, blocks can't store state. So we end up with putting this state in: 1) a perpetual ScheduleTask, 2) the redstone simulator, or 3) the arrow entities. And 1) keeps most of the code in BlockButtonHandler.h, where conceptually it belongs.

madmaxoft:
I still can't see why there can't be a simple solution that actually requires no changes in the redstone simulator and works with all cases without much trouble.

First of all, `cArrowEntity` should notify the block it hits, regardless of what block type it is. Have a new `cBlockHandler` method for it, `OnHitByProjectile`. We can use it for this button and arrow business, and we can later use it for some advanced / fun stuff (bees attacking a player if they shoot their hive, for example).

The button block handler, once it is notified that an arrow hit it, should just turn itself on. With waking up the simulator, there's no actual change in the redstone simulator needed, it will react to the arrow the same as it reacts to actual player pressing the button.

The delayed task in the button block handler needs to be changed. When triggered, it should check if the block at the coords is still a pressed button, and if so, if there are any arrows in it. If not, turn the button on, again waking up the redstone simulator.

When an arrow despawns for whatever reason, it should check whether it's lodged in a button and if so, schedule the button's delayed task (make it a `cButtonHandler::ScheduleUnpressTask`). 

Note that Cuberite has no problem having multiple scheduled tasks for the same block coords, and the task itself will check its preconditions first.

@GefaketHD Yes, I'm absolutely against linking the block handlers to any redstone simulator in any way (other than the standard wakeup call), because it brings in a new dependency between the parts of code. We want as few dependencies as possible, even to the point that individual code parts can be compiled as separate libraries and linked to different executables (think unit tests).

madmaxoft:
@tigerw Or we don't keep the state at all and calculate it when we know it could have changed.

tigerw:
GefaketHD, xoft please consult the proposed changes to see if you have any suggestions.

madmaxoft:
@tigerw That's mostly what I had in mind.

src/Blocks/BlockButton.h
```diff
@@ -133,12 +123,119 @@ class cBlockButtonHandler :
...
+					// Total failure or block changed, bail
+					return;
+				}
+
+				if (IsButtonPressedByArrow(a_World, a_Position, Type, Meta))
+				{
+					// Try again in a little while
+					QueueButtonRelease(a_World, a_Position, a_BlockType);
+					return;
+				}
```
madmaxoft, line 189: Instead of re-queuing, I think the arrows should should re-queue the release when they are destroyed (collected or despawned)

src/Blocks/BlockButton.h
```diff
@@ -133,12 +123,119 @@ class cBlockButtonHandler :
...
+	}
+
+private:
+
+	/** Schedules a recurring event at appropriate intervals to release a button at a given position.
+	The given block type is checked when the task is executed to ensure the position still contains a button. */
+	static void QueueButtonRelease(cWorld & a_Wurld, const Vector3i a_Position, const BLOCKTYPE a_BlockType)
+	{
+		const auto TickDelay = (a_BlockType == E_BLOCK_STONE_BUTTON) ? 20 : 30;
+		a_Wurld.ScheduleTask(
```
madmaxoft, line 168: Please fix the param name.

src/Blocks/BlockButton.h
```diff
@@ -133,12 +123,119 @@ class cBlockButtonHandler :
...
+
+				a_World.SetBlockMeta(a_Position, Meta & 0x07, false);
+				a_World.WakeUpSimulators(a_Position);
+				a_World.BroadcastSoundEffect(SoundToPlayOnRelease, a_Position, 0.5f, 0.5f);
+			}
+		);
+	}
+
+	/** Returns true if an arrow was found in the wooden button */
+	static bool IsButtonPressedByArrow(cWorld & a_World, const Vector3i a_Position, const BLOCKTYPE a_BlockType, const NIBBLETYPE a_Meta)
```
madmaxoft, line 202: `a_Position` -> `a_ButtonPos`
("is it the arrow's pos?")

src/Entities/ArrowEntity.cpp
```diff
@@ -97,6 +98,9 @@ void cArrowEntity::OnHitSolidBlock(Vector3d a_HitPos, eBlockFace a_HitFace)
 	// Broadcast arrow hit sound
 	m_World->BroadcastSoundEffect("entity.arrow.hit", m_HitBlockPos, 0.5f, static_cast<float>(0.75 + (static_cast<float>((GetUniqueID() * 23) % 32)) / 64));
 
+	// To make buttons get pressed if they got hit by the arrow
+	cBlockButtonHandler::OnArrowHit(*m_World, m_HitBlockPos, a_HitFace);
```
madmaxoft, line 102: The comment reads weird.

Comments are there to make the code more readable, not to let out your creative writing spirit. Although we enjoy the joke from time to time, readability is the primary objective.

src/Blocks/BlockButton.h
tigerw: I tried but couldn't resolve this issue:

Right click wooden button (causes a queue), immediately fire arrow into it, collect arrow (as you suggest, causes a queue). Now we have two queues; after the first fires, the second ends up fighting the player if they press the button again.

tigerw: The code currently resolves this by using the depressed state as a "queue in flight" indicator, so nothing adds a new queued task if the button is pressed, and the existing task keeps itself alive until it can release.

madmaxoft: So either we eat up CPU, or we get a bad very-corner-case. Oh well... Can you even simulate that corner case in vanilla? :P

tigerw: I'm half convinced vanilla does this based on how it behaves during testing...

GefaketHD:
> 
> 
> > save in the arrow
> 
> (I don't think the arrow saves anything, it only calls into a static method in the button handler on hit.)

What I was trying to say is: it seems to be logical to let the arrow press the button, but it is so much harder to implement - as you feel right now. And I wouldn't take the effort to program it, due to it is only one small exception that an entity can trigger the button.
> 
> > cases you can simply not cover
> 
> Which ones were you thinking of, this will be important to know.

I don't have every case in mind, that's what I tried to say. You can not think of *every* case that could happen to arrows, buttons, blocks, players, chunks etc. and implement them - there will be bugs with that. I'm sorry, if that sounds like I don't think you are a very good programmer and big brain, but I just do not think this is possible for anyone to do.
> 
> At least my thinking for this was:
> 
> Perhaps the reason is less technical and more that we're aiming for something easier to maintain, for both old, and new contributors. At least one other person (xoft) thinks the redstone simulator is a total black box/mystery, and I'm sure tightly coupling the button block handler and the simulator by modifying m_MechanismDelays would be too hard to for him to stomach :P

TBH, everyone thinks that and noone wants to touch it. You were described to me as the master of the redstone simulator and noone else seems to know, what is happening "in there". But why should everything be a part of software on it's own with no interaction between the handlers? I am watching the changes you are doing to the code and constantly I think "Why did you not just leave it like it is?" There are pressure plates being checked EVERY tick if their current power output is correct. They are "ticked constantly". So you could just add the button there and it would be totally fine - NO interaction between block handler and redstone simulator. Just neat redstone simulator checking up on every button every tick, like it is done with pressure plates, repeaters and torches, and the arrow wouldn't have to notify the simulators, that the button state near the position it hit a block might have changed and you don't have to implement a feature nowhere else being used that will just confuse new contributors. Because noone would check up on a button issue searching for a whole new handling you are implementing rn, but look at it's `Update`-function like in every other redstone mechanism too. Correct me if I'm wrong.
> 
> Given that player pushing buttons/arrows pushing buttons is more of an externality, you could reasonably say that it's not of the IncrementalSimulator's concern to react to these events, but instead only handle the redstone power aspect.

So you want to change pressure plates too? And go ahead with the changes you are doing right now to EVERY item in the game, because *they* are lying on the pressure plate and pressing it down, not the pressure plate presses itself depending on the items lying on it?
> 
> Plus, it would be nice for arrows to work even with NoopRedstoneSimulator.

I agree here though.
> 
> Sure, queuing tasks indefinitely isn't as good, but you know the saying, programmer time is more expensive than CPU time :)

You are just...kind of "paying" a lot of time right now to change a working feature to make it work, but differently, and implementing a new construct, used nowhere else.
> 
> And once again, I agree: conceptually, it makes total sense for the **button** to be responsible for checks. But, blocks can't store state. So we end up with putting this state in: 1) a perpetual ScheduleTask, 2) the redstone simulator, or 3) the arrow entities. And 1) keeps most of the code in BlockButtonHandler.h, where conceptually it belongs.

2) - it is emitting redstone and should be handled there. Because it's the redstone simulators job to check, which block has how much redstone being emitted from it.

tigerw:
Your arguments are reasonable, and clearly your implementation works. I guess this is just one of the times where we have to come to a consensus and stick to it.

Again, my changes are still only a proposal. 

----

Pressure plates are indeed suboptimal right now. Imagine a chunk full of them being ticked constantly... Will definitely need a rethink.

GefaketHD:
So what are we doing now? I mean, @madmaxoft wanted it like you did it too, so I guess it would be correct to implement that...? Tbh this discussion kind of frustrated me regarding this project - like: a lot...I'm sorry for behaving like this...


And yeah, I agree, that there should be a better way to handle that, but I didn't really have a good different idea on that.

tigerw:
Sorry to have caused you frustration; of course it's nothing personal. Does anyone have more thoughts on how to do it?

src/Blocks/BlockButton.h
peterbell10: Why not just let the redstone handler deal with the timing issues? I don't see why the noop simulator needs to have functioning buttons. It's seems natural that redstone devices would behave differently under a different simulator.

madmaxoft:
Cuberite is definitely causing a lot of frustration lately, I had a similar moment with the naming and the byte-sized hex. But hey, it's part of life, I hope it doesn't put you off from further contributing, because clearly you are a capable programmer.

The base principle that I'm trying to follow is that things should push notifications, rather than poll for changes constantly. I guess it's personal, at work we chose the poll architecture a long time ago and we're still paying huge price for that, so I try to avoid it as hell everywhere else. Ideally each entity would notify each block that it has entered or left. Unfortunately not everything is state-less.

Another point is that our code is currently a big ball of wibbly wobbly timey wimey chains of spaghetti, everything depends on everything else. Changing one block handler causes a recompile of almost the entire codebase. And that's plain wrong. We need to split things up, we need to add interfaces, we need to add tests. And tying (one of) the redstone simulators into the block-handlers feels like such a violation of this.
You might not feel too strongly about the alternate simulators, but I do; I've actually written the Floody fluid simulator *before* I wrote the Vanilla one because it just makes so much more sense; the Noop redstone simulator helps when running the server on an under-powered 1st gen RaspberryPi

src/Blocks/BlockButton.h
madmaxoft: Unfortunately that would mean that with the Noop simulator, all buttons will end up forever-pressed eventually, from the players pressing them.
I think this solution is the best, the CPU price is very low.

GefaketHD:
> Sorry to have caused you frustration; of course it's nothing personal. Does anyone have more thoughts on how to do it?

I know it's nothing personal and I know I'm overreacting - don't worry...just wanted to explain, why I behave like I do and why I am a little harsh in my replies. It's nothing personal from my side either.

GefaketHD:
> Cuberite is definitely causing a lot of frustration lately, I had a similar moment with the naming and the byte-sized hex. But hey, it's part of life, I hope it doesn't put you off from further contributing, because clearly you are a capable programmer.
> 

Yeah, I saw that discussion. And no, I will keep working on this project, because I like the idea and everything about it.
> The base principle that I'm trying to follow is that things should push notifications, rather than poll for changes constantly. I guess it's personal, at work we chose the poll architecture a long time ago and we're still paying huge price for that, so I try to avoid it as hell everywhere else. Ideally each entity would notify each block that it has entered or left. Unfortunately not everything is state-less.
> 

I can see, why. And event oriented programming is a very nice concept - and correct me, if I'm wrong, but this is not what cuberite is doing right now. But it is, what your idea of this arrow in a button thing is...so if that should be the general idea, a lot of changes have to be done...right? Or do you want to implement it step by step everywhere - if so, there should be a general thing and best practice for it.
> Another point is that our code is currently a big ball of wibbly wobbly timey wimey chains of spaghetti, everything depends on everything else. Changing one block handler causes a recompile of almost the entire codebase. And that's plain wrong. We need to split things up, we need to add interfaces, we need to add tests. And tying (one of) the redstone simulators into the block-handlers feels like such a violation of this.
> You might not feel too strongly about the alternate simulators, but I do; I've actually written the Floody fluid simulator _before_ I wrote the Vanilla one because it just makes so much more sense; the Noop redstone simulator helps when running the server on an under-powered 1st gen RaspberryPi

madmaxoft:
I'm hoping that Cuberite is very much event-driven. Have you seen the HTTP parsing classes? I haven't seen any such parser elsewhere - a class that gets fed with the data as it comes from the socket, and it spews out sax-like events, when encountered. If you feel something is not as it should be, definitely bring it up, ideally as an issue to be worked on (eventually :)

Alright, back to the topic. Is the code like this acceptable? Can you update this PR to fix the conflicts, so that it can be merged?

src/Blocks/BlockButton.h
```diff
@@ -133,12 +123,119 @@ class cBlockButtonHandler :
...
+			[](cEntity & a_Entity)
+			{
+				return a_Entity.IsArrow();
+			}
+		);
+
+		return FoundArrow;
+	}
+
+	static Vector3d GetButtonOffsetOnBlock(NIBBLETYPE a_Meta)
```
madmaxoft, line 221: This requires a doxycomment on what the meta is (why the YM case returns the *top* ?)

tigerw:
Merging.

PR #4676, opened 2020-04-18
Reduce chance of arrows going black on hit

tigerw:
Before
After

### DIFF
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -74,24 +74,34 @@ bool cArrowEntity::CanPickup(const cPlayer & a_Player) const
 
 void cArrowEntity::OnHitSolidBlock(Vector3d a_HitPos, eBlockFace a_HitFace)
 {
-	Vector3d Hit = a_HitPos;
-	Hit += GetSpeed().NormalizeCopy() / 100000;  // Make arrow sink into block a bit so it lodges (TODO: investigate how to stop them going so far so that they become black clientside)
+	// Save the direction we're going in before Super resets it
+	auto Speed = GetSpeed();
+	Speed.Normalize();
 
-	Super::OnHitSolidBlock(Hit, a_HitFace);
-	Vector3i BlockHit = Hit.Floor();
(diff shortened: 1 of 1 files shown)


src/Entities/ArrowEntity.cpp
```diff
@@ -74,24 +74,23 @@ bool cArrowEntity::CanPickup(const cPlayer & a_Player) const
...
+	Speed.Normalize();
 
-	Super::OnHitSolidBlock(Hit, a_HitFace);
-	Vector3i BlockHit = Hit.Floor();
+	Super::OnHitSolidBlock(a_HitPos, a_HitFace);
 
-	int X = BlockHit.x, Y = BlockHit.y, Z = BlockHit.z;
-	m_HitBlockPos = Vector3i(X, Y, Z);
+	a_HitPos += Speed / 100000;  // Make arrow sink into block a bit so it lodges
+	m_HitBlockPos = a_HitPos.Floor();
```
bearbin, line 84: Is a_HitPos actually used anywhere after this? If so I'm not quite understanding how this works - we add a tiny bit to the hitPos and then remove it again with Floor. How does that achieve anything?

tigerw: From testing the line tracer returns the arrow hit position located on the face of a block; if the arrow hit a block at -5, 95, -5 then a_HitPos would start off as **-4**, 95.5, -4.1, i.e. it collided with the X face. Then we add a bit of speed vector to make a_HitPos -4.0001, 95.5, -4.1 and floor to get exactly -5, 95, -5 which is stored in m_HitBlockPos. Does that sound right?

bearbin: Yes, that makes sense. The arrow is nudged into the block a tiny bit according to its direction of travel, then rounding that gives the coordinates of the block it crashed into.

The comment is terribly confusing though, it needs to explain that all you're doing is getting the position of the block hit, not a position within that block.

tigerw: I have done a copy paste of this discussion

PR #4678, opened 2020-04-18
More arrow patches

tigerw:
* Further reduce instances of appearing black
* Fix packet send and get arrows to more reliably lodge in blocks clientside
* Fix hit detection failing at chunk boundaries
+ Use delegating constructors

~~I don't understand why the client hates Cuberite arrows. It seems to be a case of 1) black but stays in the block, 2) lighted but randomly bounces off, 3) lighted and lodged but you have to send a teleport 3-4 times before it sticks; pick one.~~

We sent a field with the wrong data and the client mildly freaks out.

One more bug: the client takes it upon themselves to despawn the arrow after 60 seconds whether we like it or not. Which leads to ghost arrows.

### DIFF
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -3,28 +3,35 @@
 #include "Player.h"
 #include "ArrowEntity.h"
 #include "../Chunk.h"
+#include "../ClientHandle.h"
+
+
+
+
+
+/** Converts an angle in degrees into a byte representation used by the network protocol */
+static Byte AngleToProto(double a_X)
+{
+	return static_cast<Byte>((a_X * 255) / 360);
+}
 
 
 
 
 
 cArrowEntity::cArrowEntity(cEntity * a_Creator, Vector3d a_Pos, Vector3d a_Speed):
-	Super(pkArrow, a_Creator, a_Pos, 0.5, 0.5),
+	Super(pkArrow, a_Creator, a_Pos, a_Speed, 0.5, 0.5),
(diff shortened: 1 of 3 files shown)


src/Entities/ArrowEntity.cpp
```diff
@@ -3,28 +3,33 @@
...
 #include "ArrowEntity.h"
 #include "../Chunk.h"
+#include "../ClientHandle.h"
+
+
+
+
+
+/** Converts an angle in radians into a byte representation used by the network protocol */
+#define ANGLE_TO_PROTO(X) static_cast<Byte>(X * 255 / 360)
```
madmaxoft, line 13: Yaw and Pitch are stored in degrees in the entity (wrong comment)

madmaxoft: I've created an issue to move this directly into `cProtocol` descendants: #4679.

src/Entities/ArrowEntity.cpp
```diff
@@ -3,28 +3,32 @@
...
 #include "ArrowEntity.h"
 #include "../Chunk.h"
+#include "../ClientHandle.h"
+
+
+
+
+
+/** Converts an angle in degrees into a byte representation used by the network protocol */
+#define ANGLE_TO_PROTO(X) static_cast<Byte>(X * 255 / 360)
```
peterbell10, line 13:
```suggestion
static Byte AngleToProto(double a_X) 
{
	return static_cast<Byte>((a_X * 255) / 360);
}
```

PR #4681, opened 2020-04-19
Added check for item damage

Seadragon91:
- Read ^-1 from furnace recipes
- Updated recipes in furnace.txt where the item damage should be ignored

### DIFF
--- a/Server/furnace.txt
+++ b/Server/furnace.txt
@@ -8,7 +8,10 @@
 #
 # **** Item Definition ****
 # An Item is defined by an Item Type, an amount (and damage)
-# The damage is optional, and if not specified it's assumed to be 0
+# The damage is optional, and if not specified it's assumed to be 0.
+# If a damage of -1 is specified, then the damage won't be checked when matching recipes.
+# Note that when specifying items using names from items.ini, those already contain the damage value in them,
+# so the damage value doesn't need to be specified here.
 #
 # Cactus Green example:
 #    351     :   2    ( ,   1    )
@@ -25,7 +28,7 @@
 #    Cobble  :   0    ,   1    @  200  =    1     :   1    ,   1     $    10
 #   ItemType : Damage , Amount @ ticks = ItemType : Damage , Amount  $ Experience 
 #
-#   If the experience paramater isn't included, experience reward will default to zero.
+#   If the experience parameter isn't included, experience reward will default to zero.
 #
 # **** Fuel ****
 #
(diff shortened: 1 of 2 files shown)


Seadragon91:
Now I can only hope, that nothing breaks :)

Server/furnace.txt
```diff
@@ -8,7 +8,8 @@
 #
 # **** Item Definition ****
 # An Item is defined by an Item Type, an amount (and damage)
-# The damage is optional, and if not specified it's assumed to be 0
+# The damage is optional, and if not specified it uses the ingredient damage to lookup the recipe.
+# If a damage of -1 is specified, then the damage won't be checked.
```
madmaxoft, line 12: This comment about being not specified is very cryptic. How about:
```suggestion
# The damage is optional, and if not specified it's assumed to be 0.
# If a damage of -1 is specified, then the damage won't be checked when matching recipes.
# Note that when specifying items using names from items.ini, those already contain the damage value in them, so the damage value doesn't need to be specified here.
```

Server/furnace.txt
```diff
@@ -43,73 +44,73 @@
...
-YellowTerracotta    = YellowGlazedTerracotta             $ 0.1
+Beef                   = Steak                              $ 0.35
+BlackTerracotta        = BlackGlazedTerracotta              $ 0.1
+BlueTerracotta         = BlueGlazedTerracotta               $ 0.1
+BrownTerracotta        = BrownGlazedTerracotta              $ 0.10
+Cactus                 = CactusGreen                        $ 0.2
+ChainmailBoots^-1      = IronNugget                         $ 0.1
+ChainmailChestplate^-1 = IronNugget                         $ 0.1
+ChainmailHelmet^-1     = IronNugget                         $ 0.1
+ChainmailLeggings^-1   = IronNugget                         $ 0.1
```
madmaxoft, line 55: The spec at the top says `ItemType : Damage`, here you use `ItemType ^ Damage`.

src/FurnaceRecipe.cpp
```diff
@@ -171,7 +172,14 @@ void cFurnaceRecipe::AddRecipeFromLine(const AString & a_Line, unsigned int a_Li
 		return;
 	}
 
-	const AStringVector & InputSplit = StringSplit(Sides[0], "@");
+	const AStringVector & ItemDamageSplit = StringSplit(Sides[0], "^");
+	AString  ItemString = ItemDamageSplit[0];
+	if ((ItemDamageSplit.size() > 1) && (ItemDamageSplit[1] == "-1"))
+	{
+		IgnoreItemDamage = true;
+	}
```
bearbin, line 180: Just reuse the parsing code in ParseItem and check if `Recipe->In->m_ItemDamage < 0` later.

Seadragon91: @bearbin That won't work, the problem is a armor can be damaged and the damage is stored in that value

bearbin: That value isn't the damage of the in-game item, it's the damage of the recipe item so it should work. The damage of the item itself is in `a_Ingredient.m_ItemDamage` AFAIK.

Server/furnace.txt
Seadragon91: Changed it to `:-1`

src/FurnaceRecipe.cpp
```diff
@@ -290,7 +290,11 @@ const cFurnaceRecipe::cRecipe * cFurnaceRecipe::GetRecipeFrom(const cItem & a_In
 			{
 				continue;
 			}
-			else
+			else if ((Recipe.In->m_ItemDamage != -1) && (Recipe.In->m_ItemDamage == a_Ingredient.m_ItemDamage))
+			{
+				BestRecipe = &Recipe;
+			}
+			else if (Recipe.In->m_ItemDamage == -1)
```
madmaxoft, line 297:
```suggestion
			else if ((Recipe.In->m_ItemDamage == -1) || (Recipe.In->m_ItemDamage == a_Ingredient.m_ItemDamage))
```

Seadragon91:
Yes, tested different items: equipment (armor, sword and damaged), fish and normal items like cobblestone, iron ore.

PR #4682, opened 2020-04-19
Correct world access in NetherPortalScanner

tigerw:
hm

### DIFF
--- a/src/NetherPortalScanner.cpp
+++ b/src/NetherPortalScanner.cpp
@@ -297,15 +297,25 @@ void cNetherPortalScanner::OnDisabled(void)
 		Position.z += OutOffset;
 	}
 
+	auto EntityID = m_EntityID;
+	auto & DestinationWorld = m_World;
+	auto DestinationPosition = Position;
+
 	// Lookup our warping entity by ID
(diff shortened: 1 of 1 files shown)


Seadragon91:
Did tests with many many entities. Got no crash :)

src/NetherPortalScanner.cpp
```diff
@@ -297,15 +297,26 @@ void cNetherPortalScanner::OnDisabled(void)
...
-		m_EntityID,
-		[this, &Position](cEntity & a_Entity)
+	// And since this is called from the destination world's thread queue a task on the source world
+	m_SourceWorld.QueueTask(
+		[EntityID, &DestinationWorld, &SourceWorld, DestinationPosition](cWorld & a_World)
 		{
-			FLOGD("Placing player at {0}", Position);
-			a_Entity.MoveToWorld(m_World, Position, true, false);
-			return true;
+			SourceWorld.DoWithEntityByID(
```
peterbell10, line 311: Nit: You don't need to capture `SourceWorld`
```suggestion
			a_World.DoWithEntityByID(
```

tigerw: good point

PR #4683, opened 2020-04-19
Delet SpawnObject params

### LINKED ISSUE #4679: Move angle processing into cProtocol::SendSpawnObject
`cProtocol::SendSpawnObject` currently accepts yaw and pitch parameters as a `Byte`, which causes the protocol-related calculations to leak into the client classes, such as the `cArrowEntity`. The function should accept yaw and pitch in degrees, like stored in the `cEntity`, and do its conversions internally.

tigerw:
* Fix #4679

awkward...

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -3047,9 +3047,9 @@ void cClientHandle::SendSpawnMob(const cMonster & a_Mob)
 
 
 
-void cClientHandle::SendSpawnObject(const cEntity & a_Entity, char a_ObjectType, int a_ObjectData, Byte a_Yaw, Byte a_Pitch)
+void cClientHandle::SendSpawnObject(const cEntity & a_Entity, char a_ObjectType, int a_ObjectData)
 {
-	m_Protocol->SendSpawnObject(a_Entity, a_ObjectType, a_ObjectData, a_Yaw, a_Pitch);
+	m_Protocol->SendSpawnObject(a_Entity, a_ObjectType, a_ObjectData);
 }
 
 
--- a/src/ClientHandle.h
+++ b/src/ClientHandle.h
@@ -211,7 +211,7 @@ public:  // tolua_export
 	void SendSoundParticleEffect        (const EffectID a_EffectID, int a_SrcX, int a_SrcY, int a_SrcZ, int a_Data);
 	void SendSpawnFallingBlock          (const cFallingBlock & a_FallingBlock);
 	void SendSpawnMob                   (const cMonster & a_Mob);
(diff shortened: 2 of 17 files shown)


tigerw:
You thought it was still being used...

BUT NO!
----

It was ***ME***, *a 1.6 leftover*, ***ALL ALONG!***

madmaxoft:
Wait, WHUT? The params were not even used?!?

madmaxoft:
I have an itsy bitsy teenie weenie suggestion: How about unifying all the spawning into `cProtocol::SendSpawnEntity(const cEntity &)` that will figure out which of the bajillion spawning packets to send based on the entity, and take all its data from that entity?

I mean, with `a_ClientHandle.SendSpawnObject(*this, 51, 0);` for `cEnderCrystal`, how long will it be `51` and when will we need a per-protocol translation layer?

The protocol classes should not be modelled after the packets they send / receive, but on the high-level actions they provide, 's'all I'm saying.

Pokechu22:
Object IDs are one of those things that were hardcoded on Mojang's side (and hence are really jank).  They will not change, and there is no point in investing time in improving them specifically.

In 1.14 (specifically 19w05a), Spawn Object was changed to use normal entity IDs (which match those in Spawn Mob), but that can be handled by just using those same IDs.  Entity IDs do shift, but are also exposed by a data generator (if I recall correctly) and on burger.

madmaxoft:
If I remember correctly, some entities already changed their spawn packets; the object data can also change meaning between protocol versions. My point is, the Entity classes should not be the ones taking care of that, the Protocol classes should.

tigerw:
In the current wiki.vg/Protocol I don't see any Spawn Object at all, only Spawn Entity. Interestingly https://wiki.vg/Entity_metadata still references Spawn Object.

Pokechu22:
That confused me as well; someone renamed them to Spawn Entity (and Spawn Living Entity).

tigerw:
How do you know so much about the protocol? You don't happen to work at Mojang, do you? :)

Pokechu22:
I do a lot of work on writing the documentation (mainly research for snapshots, based on [burger](https://github.com/mcdevs/Burger) (which I maintain); that in particular gives me a fair bit of familiarity with when numeric IDs change).  I'm also a volunteer moderator over on Mojang's bug tracker, but I'm not an employee; this has given me some level of familiarity with different bugs and when they were fixed (which helps identify specific changes).

tigerw:
I think before we implement SendSpawnEntity and to avoid duplicating the logic twice, we should get the 1.9 protocol to inherit from 1.8. diff shows ~90% the same, the rest seems like it can be overridden as normal.

PR #4685, opened 2020-04-20
Unify entity spawn packet sending

tigerw:
at the behest of xoft https://github.com/cuberite/cuberite/pull/4683#issuecomment-616217686

### DIFF
--- a/src/Items/ItemItemFrame.h
+++ b/src/Items/ItemItemFrame.h
@@ -41,7 +41,7 @@ public:
 		}
 
 		// Make sure block that will be occupied by the item frame is free now:
-		auto PlacePos = AddFaceDirection(a_ClickedBlockPos, a_ClickedBlockFace);
+		const auto PlacePos = AddFaceDirection(a_ClickedBlockPos, a_ClickedBlockFace);
 		BLOCKTYPE Block = a_World->GetBlock(PlacePos);
 		if (Block != E_BLOCK_AIR)
 		{
@@ -49,7 +49,7 @@ public:
 		}
 
 		// Place the item frame:
-		auto ItemFrame = cpp14::make_unique<cItemFrame>(a_ClickedBlockFace, a_ClickedBlockPos);
+		auto ItemFrame = cpp14::make_unique<cItemFrame>(a_ClickedBlockFace, PlacePos);
 		auto ItemFramePtr = ItemFrame.get();
 		if (!ItemFramePtr->Initialize(std::move(ItemFrame), *a_World))
 		{
--- a/src/Protocol/Protocol.h
+++ b/src/Protocol/Protocol.h
@@ -185,7 +185,6 @@ public:
 	virtual void SendLoginSuccess               (void) = 0;
 	virtual void SendMapData                    (const cMap & a_Map, int a_DataStartX, int a_DataStartY) = 0;
 	virtual void SendPaintingSpawn              (const cPainting & a_Painting) = 0;
-	virtual void SendPickupSpawn                (const cPickup & a_Pickup) = 0;
 	virtual void SendPlayerAbilities            (void) = 0;
 	virtual void SendEntityAnimation            (const cEntity & a_Entity, char a_Animation) = 0;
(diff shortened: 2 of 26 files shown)


madmaxoft:
I'd have preferred if there was no `SendSpawnEntityData`, but instead the `SendSpawnEntity` would have the entity-type forks that are currently in `GetProtocolSendSpawnEntityData`, and instead of `return {some, data}` there'd be a `SendSpawnObjectPacket(some, data)` call, or a call to specialized (overridable) `SendSpawnArrowEntity` etc.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -3918,3 +3850,46 @@ void cProtocol_1_8_0::WriteEntityProperties(cPacketizer & a_Pkt, const cEntity &
...
+				case PType::pkEgg: return 62;
+				case PType::pkGhastFireball: return 63;
+				case PType::pkFireCharge: return 64;
+				case PType::pkEnderPearl: return 65;
+				case PType::pkExpBottle: return 75;
+				case PType::pkSplashPotion: return 73;
+				case PType::pkFirework: return 76;
+				case PType::pkWitherSkull: return 66;
+				default: UNREACHABLE("Spawned an unhandled projectile");
+			}
```
peterbell10, line 3888: The idea is for the compiler to give you warnings if you don't cover all the enum cases. That only happens if there isn't a `default` label.
```suggestion
			}
			UNREACHABLE("Spawned an unhandled projectile");
```

peterbell10: (This happens in a few other places as well)

src/Protocol/Protocol_1_8.h
```diff
@@ -104,10 +103,9 @@ class cProtocol_1_8_0:
 	virtual void SendSetTitle                   (const cCompositeChat & a_Title) override;
 	virtual void SendSetRawTitle                (const AString & a_Title) override;
 	virtual void SendSoundParticleEffect        (const EffectID a_EffectID, int a_SrcX, int a_SrcY, int a_SrcZ, int a_Data) override;
-	virtual void SendSpawnFallingBlock          (const cFallingBlock & a_FallingBlock) override;
+	virtual void SendSpawnEntity                (const cEntity & a_Entity) override;
+	virtual void SendSpawnEntity                (const cEntity & a_Entity, const UInt8 a_ObjectType, const Int32 a_ObjectData);
```
madmaxoft, line 107: This should probably be a `protected` member, since it's expected to be called only from the inside of the protocol class.

src/Items/ItemItemFrame.h
```diff
@@ -34,7 +34,6 @@ class cItemItemFrameHandler :
 
 		AddFaceDirection(a_BlockX, a_BlockY, a_BlockZ, a_BlockFace);  // Make sure block that will be occupied is free
 		BLOCKTYPE Block = a_World->GetBlock(a_BlockX, a_BlockY, a_BlockZ);
-		AddFaceDirection(a_BlockX, a_BlockY, a_BlockZ, a_BlockFace, true);  // We want the clicked block, so go back again
```
madmaxoft, line 37: I see this is "countered" by the protocol code, but what about the data saved to the chunk, is it valid then?

tigerw:
Thanks for your reviews, I'll get on to it soon.

tigerw:
I have addressed your comments, what d'you think?

src/Items/ItemItemFrame.h
tigerw: Wiki says Vanilla changed storage to the block it's in back in 1.8, the last one we support.

tigerw:
Going to merge

src/Protocol/Protocol.h
```diff
@@ -259,9 +256,6 @@ class cProtocol
 	/** Returns the protocol-specific packet ID given the protocol-agnostic packet enum. */
 	virtual UInt32 GetPacketID(ePacketType a_Packet) = 0;
 
-	/** Converts eMonsterType to protocol-specific mob types */
-	virtual UInt32 GetProtocolMobType(eMonsterType a_MobType) = 0;
```
peterbell10, line 263: Why was this moved?

tigerw: It was too high up and forced ProtocolRecognizser to have a dummy `return 0` implementation, now the root of the inheritance chain starts lower, from a concrete protocol 1.8.

tigerw:
Thanks!

PR #4687, opened 2020-04-21
Remove some unnecessary packet sends

tigerw:
In grass and entities. Grass got [pre-emptively](https://github.com/cuberite/cuberite/commit/26ac146f41091dc070d8075f5fc9de25b5a22578#r38650033) fixed by xoft so here's a refactor instead.


madmaxoft:
Basically the changes are good, but the naming is not so great. The function names are not descriptive enough. How about `DieFromCovering` (with the light check moved back to the `OnUpdate`) and `TrySpreadTo`?

madmaxoft:
Also, this should be in two separate PRs.

tigerw:
> Also, this should be in two separate PRs.

Sorry. Do you remember my [SEE DESC] pull requests :p

tigerw:
The light check is wrapped in a block that defines AbovePos and checks its validity. Moving it out seems like it would unravel the first function.

src/Blocks/BlockGrass.h
```diff
@@ -52,72 +52,102 @@ class cBlockGrassHandler :
...
+		// Make sure that there is enough light at the source block to spread
+		const auto Light = std::max(a_Chunk.GetBlockLight(AbovePos), a_Chunk.GetSkyLightAltered(AbovePos));
+		return (Light >= 9);
+	}
+
+
+
+
+
+	/** Attempt to spread grass to a block at the given position. */
```
bearbin, line 108: This should be `TrySpreadTo`, not obvious enough that RelPos is the pos of the dirtblock rather than the grassblock. Dirt does not 'thrive'!

src/Blocks/BlockGrass.h
```diff
@@ -52,72 +52,102 @@ class cBlockGrassHandler :
...
+	Returns whether conditions are so good that the grass can try to spread to neighbours. */
+	static bool TrySurvive(cChunk & a_Chunk, const Vector3i a_RelPos)
+	{
+		const auto AbovePos = a_RelPos.addedY(1);
+		if (!cChunkDef::IsValidHeight(AbovePos.y))
+		{
+			return true;
+		}
+
+		// Grass turns back to dirt when the block Above it is not transparent or water.
```
bearbin, line 88: I see what you're trying to do here, but it just makes it a bit more confusing to Capitalise Above.

tigerw:
Implemented suggestions, any more comments?

Issue #3572, opened 2017-02-15
Crash when manipulating the world generation with a plugin


sphinxc0re:
Server OS:        macOS
Commit id:        dae2adde6563bec4b614230992cae6b1ca965b4f

### Expected behavior
When using my world obfuscation plugin [sphinx0res](https://github.com/sphinxc0re/sphinx0res) it should obfuscate the whole world statically on generation.

### Actual behavior
Instead the server crashes with the following output with the debug build:
```
Log: [790209bf879cf78c|20:40:56] --- Started Log ---
Log: [790209bf879cf78c|20:40:56] Cuberite local build build id: Unknown
Log: [790209bf879cf78c|20:40:56] from commit id: approx: dae2adde6563bec4b614230992cae6b1ca965b4f built at: approx: 2017-02-15 07:05:24 +0200
Log: [790209bf879cf78c|20:40:56] Creating new server instance...
Log: [790209bf879cf78c|20:40:56] Reading server config...
Log: [790209bf879cf78c|20:40:56] Starting server...
Info: [790209bf879cf78c|20:40:56] Compatible clients: 1.8.x, 1.9.x, 1.10.x, 1.11
Info: [790209bf879cf78c|20:40:56] Compatible protocol versions 47, 107, 108, 109, 110, 210, 315
Log: [790209bf879cf78c|20:40:56] Generating protocol encryption keypair...
Log: [790209bf879cf78c|20:40:56] Initialising WebAdmin...
Warning: [790209bf879cf78c|20:40:56] WebServer: The server will run in unsecured HTTP mode.
Info: [790209bf879cf78c|20:40:56] Put a valid HTTPS certificate in file 'webadmin/httpscert.crt' and its corresponding private key to 'webadmin/httpskey.pem' (without any password) to enable HTTPS support
Log: [790209bf879cf78c|20:40:56] Loading settings...
Log: [790209bf879cf78c|20:40:56] Loading crafting recipes from crafting.txt...
Log: [790209bf879cf78c|20:40:56] Loaded 726 crafting recipes
Log: [790209bf879cf78c|20:40:56] Loading furnace recipes...
Log: [790209bf879cf78c|20:40:56] Loaded 52 furnace recipes and 66 fuels
Log: [790209bf879cf78c|20:40:56] Loading brewing recipes...
Log: [790209bf879cf78c|20:40:56] Loaded 88 brewing recipes
Log: [790209bf879cf78c|20:40:56] Loading worlds...
[...]

PR #4689, opened 2020-04-22
Remove blocking GetHeight

### LINKED ISSUE #4688: Deadlock: cChunkMap::GetHeight called from block handler
Client version: <unknown>
Server OS: Linux
Cuberite Commit id: 246acb19f99cf974080e1bdf28acd6a263d6e971

### Actual behavior
Server crashes due to a deadlock. Reported by Memux on discordapp.

### Stack traces
```
Thread 6 (Thread 0x7fffee7f8700 (LWP 8090)):
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:103
#1  0x00007ffff7c867d1 in __GI___pthread_mutex_lock (mutex=0x55555673ff50) at ../nptl/pthread_mutex_lock.c:115
#2  0x0000555555e066b6 in __gthread_mutex_lock (__mutex=0x55555673ff50)
    at /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:748
#3  0x0000555555e06706 in __gthread_recursive_mutex_lock (__mutex=0x55555673ff50)
    at /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:810
#4  0x0000555555e0673a in std::recursive_mutex::lock (this=0x55555673ff50) at /usr/include/c++/8/mutex:107
#5  0x0000555555e0634b in cCriticalSection::Lock (this=0x55555673ff40)
    at /root/cuberite/src/OSSupport/CriticalSection.cpp:23
#6  0x0000555555e06564 in cCSLock::Lock (this=0x7fffee7f7cb0)
    at /root/cuberite/src/OSSupport/CriticalSection.cpp:105
#7  0x0000555555e064b5 in cCSLock::cCSLock (this=0x7fffee7f7cb0, a_CS=...)
    at /root/cuberite/src/OSSupport/CriticalSection.cpp:81
#8  0x0000555555cf0512 in cChunkMap::IsChunkQueued (this=0x55555673ff40, a_ChunkX=0, a_ChunkZ=0)
    at /root/cuberite/src/ChunkMap.cpp:386
#9  0x0000555555dda6fb in cWorld::IsChunkQueued (this=0x7ffff6362010, a_ChunkX=0, a_ChunkZ=0)
    at /root/cuberite/src/World.cpp:2430
#10 0x00005555560ecbe3 in cWorldStorage::LoadChunk (this=0x7ffff63621f0, a_ChunkX=0, a_ChunkZ=0)
    at /root/cuberite/src/WorldStorage/WorldStorage.cpp:288
#11 0x00005555560eca0f in cWorldStorage::LoadOneChunk (this=0x7ffff63621f0)
    at /root/cuberite/src/WorldStorage/WorldStorage.cpp:238
#12 0x00005555560ec983 in cWorldStorage::Execute (this=0x7ffff63621f0)
    at /root/cuberite/src/WorldStorage/WorldStorage.cpp:217
#13 0x0000555555e08cf3 in cIsThread::DoExecute (this=0x7ffff63621f0)
    at /root/cuberite/src/OSSupport/IsThread.cpp:73
#14 0x0000555555e09582 in std::__invoke_impl<void, void (cIsThread::*)(), cIsThread*> (
    __f=@0x55555b8344b0: (void (cIsThread::*)(class cIsThread * const)) 0x555555e08cc4 <cIsThread::DoExecute()>, 
    __t=@0x55555b8344a8: 0x7ffff63621f0) at /usr/include/c++/8/bits/invoke.h:73
#15 0x0000555555e0912f in std::__invoke<void (cIsThread::*)(), cIsThread*> (
    __fn=@0x55555b8344b0: (void (cIsThread::*)(class cIsThread * const)) 0x555555e08cc4 <cIsThread::DoExecute()>, 
    __args#0=@0x55555b8344a8: 0x7ffff63621f0) at /usr/include/c++/8/bits/invoke.h:95
#16 0x0000555555e09893 in std::thread::_Invoker<std::tuple<void (cIsThread::*)(), cIsThread*> >::_M_invoke<0ul, 1ul> (this=0x55555b8344a8) at /usr/include/c++/8/thread:244
#17 0x0000555555e0984e in std::thread::_Invoker<std::tuple<void (cIsThread::*)(), cIsThread*> >::operator() (
    this=0x55555b8344a8) at /usr/include/c++/8/thread:253
#18 0x0000555555e09832 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (cIsThread::*)(), cIsThread*> > >::_M_run (this=0x55555b8344a0) at /usr/include/c++/8/thread:196
#19 0x00007ffff7ef5b2f in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#20 0x00007ffff7c83fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#21 0x00007ffff7bb44cf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 8 (Thread 0x7fffed7f6700 (LWP 8092)):
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:103
#1  0x00007ffff7c867d1 in __GI___pthread_mutex_lock (mutex=0x55555673ff50) at ../nptl/pthread_mutex_lock.c:115
#2  0x0000555555e066b6 in __gthread_mutex_lock (__mutex=0x55555673ff50)
    at /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:748
#3  0x0000555555e06706 in __gthread_recursive_mutex_lock (__mutex=0x55555673ff50)
    at /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:810
#4  0x0000555555e0673a in std::recursive_mutex::lock (this=0x55555673ff50) at /usr/include/c++/8/mutex:107
#5  0x0000555555e0634b in cCriticalSection::Lock (this=0x55555673ff40)
    at /root/cuberite/src/OSSupport/CriticalSection.cpp:23
#6  0x0000555555e06564 in cCSLock::Lock (this=0x7fffed7f5ba0)
    at /root/cuberite/src/OSSupport/CriticalSection.cpp:105
#7  0x0000555555e064b5 in cCSLock::cCSLock (this=0x7fffed7f5ba0, a_CS=...)
    at /root/cuberite/src/OSSupport/CriticalSection.cpp:81
#8  0x0000555555cf0636 in cChunkMap::HasChunkAnyClients (this=0x55555673ff40, a_ChunkX=13, a_ChunkZ=6)
    at /root/cuberite/src/ChunkMap.cpp:408
#9  0x0000555555dda76f in cWorld::HasChunkAnyClients (this=0x7ffff6362010, a_ChunkX=13, a_ChunkZ=6)
    at /root/cuberite/src/World.cpp:2448
#10 0x0000555555cfe4fb in cChunkSender::SendChunk (this=0x7ffff6362930, a_ChunkX=13, a_ChunkZ=6, 
    a_Clients=Python Exception <class 'IndexError'> list index out of range: 
std::unordered_set with 1 element) at /root/cuberite/src/ChunkSender.cpp:225
#11 0x0000555555cfe349 in cChunkSender::Execute (this=0x7ffff6362930) at /root/cuberite/src/ChunkSender.cpp:197
#12 0x0000555555e08cf3 in cIsThread::DoExecute (this=0x7ffff6362930)
    at /root/cuberite/src/OSSupport/IsThread.cpp:73
#13 0x0000555555e09582 in std::__invoke_impl<void, void (cIsThread::*)(), cIsThread*> (
    __f=@0x55555b82f920: (void (cIsThread::*)(class cIsThread * const)) 0x555555e08cc4 <cIsThread::DoExecute()>, 
    __t=@0x55555b82f918: 0x7ffff6362930) at /usr/include/c++/8/bits/invoke.h:73
#14 0x0000555555e0912f in std::__invoke<void (cIsThread::*)(), cIsThread*> (
    __fn=@0x55555b82f920: (void (cIsThread::*)(class cIsThread * const)) 0x555555e08cc4 <cIsThread::DoExecute()>, 
    __args#0=@0x55555b82f918: 0x7ffff6362930) at /usr/include/c++/8/bits/invoke.h:95
#15 0x0000555555e09893 in std::thread::_Invoker<std::tuple<void (cIsThread::*)(), cIsThread*> >::_M_invoke<0ul, 1ul> (this=0x55555b82f918) at /usr/include/c++/8/thread:244
#16 0x0000555555e0984e in std::thread::_Invoker<std::tuple<void (cIsThread::*)(), cIsThread*> >::operator() (
    this=0x55555b82f918) at /usr/include/c++/8/thread:253
#17 0x0000555555e09832 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (cIsThread::*)(), cIsThread*> > >::_M_run (this=0x55555b82f910) at /usr/include/c++/8/thread:196
#18 0x00007ffff7ef5b2f in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#19 0x00007ffff7c83fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#20 0x00007ffff7bb44cf in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Looking at the log files he posted, he did nothing strange and didn't even run any command.

peterbell10:
Fixes #4688

As the issue shows, blocking `GetHeight` is very dangerous because we can't reliably unlock the chunk map mutex. Instead, we should return `optional<int>` and figure out how to respond when the chunk isn't loaded at the call site.

Since we don't have c++17, I'm just imitating the `std::optional` API with `cpp17::optional`.

WIP to figure out what to do with block handlers, entities and lua api which all use `GetHeight` either directly or indirectly.
- Biggest use case is `IsWeatherWetAtXYZ` which doesn't really matter that much. We can probably get away with using a default value there.
- Lua API is tricky because we can't make the descision for the plugin. Either we need to figure out something clever, or just change the API and raise an error.
- Similarly, the spawn height is something that we can't just guess. One thought would be to use `cChunkMap::GenerateChunk` with a callback that sets the spawn. That might leave a gap where the spawn isn't set though which wouldn't be great.


src/Bindings/ManualBindings_World.cpp
```diff
@@ -935,12 +935,11 @@ static int tolua_cWorld_TryGetHeight(lua_State * tolua_S)
 	}
 
 	// Call the implementation:
-	int Height = 0;
-	bool res = self->TryGetHeight(BlockX, BlockZ, Height);
-	L.Push(res);
-	if (res)
+	cpp17::optional<int> Height = self->GetHeight(BlockX, BlockZ);
```
madmaxoft, line 938: Wouldn't `auto` work here?

src/Blocks/WorldInterface.h
```diff
@@ -96,7 +99,7 @@ class cWorldInterface
 	virtual void SetMaxNetherPortalHeight(int a_NewMaxHeight) = 0;
 
 	/** Returns the world height at the specified coords; waits for the chunk to get loaded / generated */
-	virtual int GetHeight(int a_BlockX, int a_BlockZ) = 0;
+	virtual cpp17::optional<int> GetHeight(int a_BlockX, int a_BlockZ) = 0;
```
madmaxoft, line 102: The doxy-comment is wrong now.

src/Bindings/ManualBindings_World.cpp
```diff
@@ -997,6 +1057,8 @@ void cManualBindings::BindWorld(lua_State * tolua_S)
 			tolua_function(tolua_S, "SetSignLines",                 tolua_cWorld_SetSignLines);
 			tolua_function(tolua_S, "SpawnSplitExperienceOrbs",     tolua_cWorld_SpawnSplitExperienceOrbs);
 			tolua_function(tolua_S, "TryGetHeight",                 tolua_cWorld_TryGetHeight);
+			tolua_function(tolua_S, "GetHeight",                    tolua_cWorld_GetHeight);
+			tolua_function(tolua_S, "IsWeatherWetAtXYZ",                    tolua_cWorld_IsWeatherWetAtXYZ);
```
madmaxoft, line 1061: Vertical alignment failed.

tigerw:
I'm not keen on an ad-hoc introduction of optional. It would be better if we prevented things that are ticked under the chunkmap lock from calling back into world. For example, the farmland handler by rights should be able to determine if its wet solely from the cChunk it gets.

Ideally GetHeight would simply return 0 when the chunk isn't loaded, as IsWeatherWetAt should return the global weather state. This is both consistent with what GetBlock does, as well as the physical interpretation that if the chunk is empty it's gonna be exposed to the sky and have no height.

peterbell10:
>It would be better if we prevented things that are ticked under the chunkmap lock from calling back into world. 

I don't think that's relevant to `GetHeight`. If they went through `cChunk::GetNeighbourChunk` for example, they still need to deal with missing chunks. Avoiding reentering into `cWorld` is also made impossible by plugin hooks where `cWorld` may be the only way to write it.

>the farmland handler by rights should be able to determine if its wet solely from the cChunk it gets.

Clearly not, or `cWorld::GetHeight` would always succeed. The problem is that it checks for rain on neighbouring blocks.

>Ideally GetHeight would simply return 0 when the chunk isn't loaded, as IsWeatherWetAt should return the global weather state. This is both consistent with what GetBlock does, as well as the physical interpretation that if the chunk is empty it's gonna be exposed to the sky and have no height.

This comes back to @madmaxoft's point in https://github.com/cuberite/cuberite/pull/4680#discussion_r411130731. Failing silently is not good because we can get incorrect behaviours as a result. e.g. a block might fail a `CanBeAt` check because of the default value and drop when it shouldn't.

As for `GetBlock`, I think it should return optional as well. A missing chunk should be treated specially, not just as air.

tigerw:
In response to not failing silently, GetBlock returns optional etc:

Why? It's practically ingrained into the fabric of Minecraft that chunks might be unloaded. If it's absolutely required that you need the data, use a chunkloader. And what do unloaded chunks contain, if not air?

About incorrect CanBeAt, I think it would suffice to stop ticking chunks with unloaded neighbours, or better yet when getting neighbours directly from a cChunk, simply "failing safe" and doing nothing.

Certainly the farmland handler is strange though, (unless I'm totally blind xD) the call to IsWeatherWetAt doesn't seem to modify a_RelPos at all, and yet it couldn't find its own chunk? wat?

Agreed with the point about going through cWorld in the general case; given that it's not feasible to remove the lock right now GetHeight has to change.

peterbell10:
>It's practically ingrained into the fabric of Minecraft that chunks might be unloaded.

Returning `nullopt` instead of air doesn't mean that unloaded chunks are no longer "ingrained into the fabric of Minecraft"; it just means we can differentiate between "unloaded" vs "loaded and has air at this block".

>If it's absolutely required that you need the data, use a chunkloader. And what do unloaded chunks contain, if not air?

That's not the same at all: a chunk stay gives you the accurate data 100% of the time without failing. `GetBlock` fails silently giving you inaccurate data. `optional` is just encoding that failure into the return value. If treating it as air is fine, then just call `GetBlock().value_or(E_BLOCK_AIR)` and it's the same as before.

>better yet when getting neighbours directly from a cChunk, simply "failing safe" and doing nothing.

That's exactly what this change allows. You can now fail safe when going through `cWorld`.

> Certainly the farmland handler is strange though, (unless I'm totally blind xD) the call to `IsWeatherWetAt` doesn't seem to modify `a_RelPos` at all, and yet it couldn't find its own chunk? wat?

You're right. The original code was using `a_RelPos` as an absolute coordinate for `IsWeatherWetAtXYZ` so it was a bug that it went into different chunks. In general, it's still true that blocks depend on their neighbours though.

tigerw:
I disagree that getting the height of, or a block at an unloaded chunk and receiving 0/air is a failure. The concept is meaningless when for any typical server, GetBlock would 'fail' for most of its inputs.

I argue that, if one needs something more descriptive:
1. Ask if the chunk is valid, and fail if not - it is after all the purpose of chunking
2. Use a call that returns whether a chunk was there
2. Use a chunkstay

To be clear I'm absolutely not against something like TryGetHeight or GetBlockTypeMeta which return bools. My main concern is the introduction of an optional type and the inconsistency this brings with regards to everything that already exists.

peterbell10:
>I disagree that getting the height of, or a block at an unloaded chunk and receiving 0/air is a failure. The concept is meaningless when for any typical server, GetBlock would 'fail' for most of its inputs.

I disagree with the premise. Failure doesn't have to be rare. A compiler for example has an infinite number of input files that would fail to parse. That doesn't make it's any less of a failure. 

Whether you want to call it failure or not, distinguishing a loaded from an unloaded chunk is useful. Having to call a separate `IsValid` function first is just making the API more awkward.

>My main concern is the introduction of an optional type and the inconsistency this brings with regards to everything that already exists.

The `bool TryGet` functions are completely equivalent to returning `optional`. It's entirely historical and now that we have better tools, we would move away from it.

tigerw:
About the last point, surely not. Migrating piecemeal to optional in getheight, when things like getblock will effectively never change doesn't seem right to me. (also I believe xoft strongly dislikes underscores, so value_or is questionable :p). We should strongly prefer to maintain consistency. If they are equivalent, I would really urge that GetHeight be made not to block its caller and load chunks, TryGetHeight left as-is, and we should figure out why the farmland call tried to load any chunks in the first place.

tigerw:
I am hugely concerned about consistency since this is still essentially a small project. A few years ago the cWorldInterface/ChunkInterfaces were introduced, but there was a lack of developer manpower to drive adoption or maintain them. This means that now multiple places have cWorldInterface and a cWorld, or both world and chunk interfaces etc. I feel keeping things simple is our best bet.

peterbell10:
>We should strongly prefer to maintain consistency.

I'm reminded of ["a foolish consistency"](http://slashslash.info/2018/02/a-foolish-consistency/)

> figure out the best way to express something and always express it the same way.
>
>Until you learn a better way. Then embrace the improvement. This inconsistency is called *growth*.

I agree that consistency is good, but only if the pattern you're following is consistently good.

We already have several different ways of encoding "this operation might fail". Many follow the pattern of `TryGetHeight`: start with `Try`, return a `bool` and return the actual value with an out-parameter. However, some actually start with `Unbounded` and others like `GetBlockTypeMeta` don't give any hint they can fail. 

Many callers forget to check the return value. When the function does fail, you end up using an unassigned value. If we step outside of `cWorld`, cuberite also has a common pattern of returning `std::pair<bool, T>` (or `std::pair<T, bool>`). Clearly looking for something like `optional`.

So, there is currently very little consistency anyway. And we should not just aim for consistency, but to consistently use the best option available to us. Not just the one that's most common now.

I propose we use `optional` moving forward because it clearly expresses the intent and embeds it in semantics of the type. It forces the programmer to acknowledge the failure case and makes it visible in code review. You might forget that `GetBlockTypeMeta` can fail, but you can't forget to dereference an `optional`. Otherwise, you would get a compile error.

tigerw:
The consistency of naming, or whether some function returns success in the first place can be discussed on another day, and besides, getblocktypemeta on a chunk cannot fail, I thought?

It's also not like our caller can't call value directly on the optional and end up with a bogus read in Release. Plus what is and isn't caught in code review is a complete unknown quantity, there's no evidence either way to suggest which is more visible.

I think the discussion has boiled down to a philosophical debate about the merits of bool vs optional, since from a technical perspective they both work. The main problem to be fixed was making getheight non-blocking, which you've done.

tigerw:
But wait....
there's a new challenger on the scene.

What about: *returning -1 for a failed getheight?*

This neatly sidesteps you concern about 0 being overloaded!

peterbell10:
>It's also not like our caller can't call value directly on the optional and end up with a bogus read in Release. Plus what is and isn't caught in code review is a complete unknown quantity, there's no evidence either way to suggest which is more visible.

I find this argument completely absurd. If you see `Height.value()` in code review, it's obvious that it's an optional. If you see `m_World->GetBlockTypeMeta`, there is absolutely no hint that it can fail. The proof that this will get past code review, is the fact that there are hundreds of places in cuberite where similar functions are called without checking for  failure. Clearly the status quo isn't working.

>What about: returning -1 for a failed getheight?

That has the exact same problem. It isn't obvious it can fail, so people will forget to check for failure.

tigerw:
Please bear with me, I like a technical debate :)

I was suggesting that if it gets past review, optional does nothing for the "failure" case where an empty chunk is interpreted as containing air. About visibility during review, I argue that just as you know .value implies optional, another hypothetical reviewer knows GetBlockTypeMeta implies a boolean return value, by virtue of knowing the signature of one of the most used class of functions.

And about the status quo not working, I haven't seen evidence of this. The deadlock was caused by GetHeight improperly blocking.

And your final point gets back to the definition of failure, which we couldn't agree on.

hm, what about something like

|            | Unloaded chunk is failure |   |      | Like in C#   |              |
|------------|---------------------------|---|------|--------------|--------------|
| unsigned   | GetHeight                 |   | bool | TryGetHeight | out H        |
| BLOCKTYPE  | GetBlock                  |   | bool | TryGetBlock  | out B        |
| NIBBLETYPE | GetMeta                   |   | bool | TryGetMeta   | out N        |
| pair       | GetBlTyMe                 |   | bool | TryGetBlTyMe | out B, out N |

This way, if you know that your call always will succeed - no overhead, no checks, but if you're wrong, assertion in debug, or fails your plugin. On the other hand, TryXX, as in C#. With bools, or, if you really insist, optional.

tigerw:
Then, ensure in the main server code that entities/blocks on chunk boundaries aren't ticked, so they can never make a failing call.

peterbell10:
@madmaxoft as benevolent dictator, do you have any preference on `bool TryGet` vs `optional<T> Get`?

madmaxoft:
I'm in favor of `Optional<T>`; however, I don't like its underscore-y API (just as @tigerw predicted). I know it's mimicking the API of the C++17's `std::optional`, but still...

So, basically, two questions:
- Why aren't we using C++17 already? If there's no blockers, I'd like to go that way, use the `std::optional` (and auto-ties :) Perhaps build servers? @bearbin ?
- If that's impossible, can we make the symbol names consistent with the rest of the code (`cClass::FunctionName`)? Wrapping a `std::optional` into this API should be easy once we do move to C++17.

madmaxoft:
The only problem I could think of with `GetHeight` getting this overhaul is that it used to be advertised in the Lua API as the "ideal way to make sure a chunk is loaded / generated" (from the times before ChunkStay).

bearbin:
I think build servers are the only reason why we aren't using C++17. It should not be too difficult to change though, all that's required is to manually create a new platform definition and cross-compiler pack for the raspberry pi builds. 

In terms of the actual substance, I strongly dislike introducing `Optional<T>` which is non-standard and used in only one place; but the general principle of optionals and using the standard C++17 form is something I very much support. So my feeling is that if we are in the mood for a larger refactor, we should make everything use the optionals; if not just change to return a boolean and output reference signature.

tigerw:
One other concern (trustme is guaranteed to always be passed not 0):

```cpp
#include <optional>

std::optional<int> get(int x)
{
    if (x == 0)
    {
        return {};
    }

    return 1;
}

int trustme(int x) noexcept
{
   return *get(x); 
}
```

compiles to

```asm
trustme(int):
        mov     eax, 1
        test    edi, edi
        jne     .L5
        xor     eax, eax
        mov     BYTE PTR [rsp-1], 0
        mov     BYTE PTR [rsp-4], 0
        mov     WORD PTR [rsp-3], ax
        mov     eax, DWORD PTR [rsp-8]
.L5:
        ret
```

and calling .value instead

```cpp
#include <optional>

std::optional<int> get(int x)
{
    if (x == 0)
    {
        return {};
    }

    return 1;
}

int trustme(int x) noexcept
{
   return get(x).value(); 
}
```

is even worse

```asm
trustme(int):
        sub     rsp, 8
        test    edi, edi
        je      .L11
        mov     eax, 1
        add     rsp, 8
        ret
trustme(int) [clone .cold]:
.L11:
        mov     edi, 8
        call    __cxa_allocate_exception
        mov     edx, OFFSET FLAT:std::bad_optional_access::~bad_optional_access() [complete object destructor]
        mov     esi, OFFSET FLAT:typeinfo for std::bad_optional_access
        mov     QWORD PTR [rax], OFFSET FLAT:vtable for std::bad_optional_access+16
        mov     rdi, rax
        call    __cxa_throw
```

-----

Compare with

```cpp
#include <cassert>

int get(int x)
{
    if (x == 0)
    {
       assert(!"you goofed");
    }

    return 1;
}

int trustme(int x) noexcept
{
   return get(x); 
}
```

on Release

```asm
trustme(int):
        mov     eax, 1
        ret
```

Are you sure you want to pay this penalty for every Get, even those you either know will absolutely succeed, in Debug and Release? Not to mention littering every call with either an asterisk or .value()

peterbell10:
>Why aren't we using C++17 already? If there's no blockers, I'd like to go that way, use the std::optional (and auto-ties :) Perhaps build servers? @bearbin ?

I would love to move to C++17. I believe, as @bearbin said, the main reason was that the raspberry pi toolchain we use is too old. If that's something we can fix, then absolutely go for that.

>If that's impossible, can we make the symbol names consistent with the rest of the code 

I was going by the precedent of `cpp14::make_unique` that `std` compatibility features should follow the standard naming. Hopefully we will upgrade our toolchains at some point. When that happens, consistent naming should make it easier to convert. I don't have a strong opinion on this though.


@tigerw that's not a fair comparison. You compile an assert in release mode and it compiles to nothing. That's obviously true by the definition of `assert`. A fair comparison would return `bool` from `get` and use an out param. Or, at the very least you would use `VERIFY` instead of `assert`.

tigerw:
xoft: yeah Lua might need to be broken as peterbell points out, unless there's a way to detect the condition and unlock the mutex twice?

peterbell: I was thinking of the hot-path plain int returning GetHeight/GetBlock; of course a similar TryGetX would incur the same penalty as the return value needs to be checked. Wouldn't you agree that it would be good for callers who know the call will always succeed (loaded chunk, chunkstay, player in close, previous call succeeded) to have the choice of using the direct version, instead of checking optional?

peterbell10:
Here is a fairer comparison:
https://gcc.godbolt.org/z/TnRfrJ

The additional cost is 3 extra instructions. Accessing the chunk storage arrays in the first place is quite likely to miss the CPU cache, in which case memory accesses can be 100x more expensive than a normal instruction. So, I genuinely don't think this will be important performance-wise.

If you do _really_ want direct access with minimal overhead (say for example to implement `cBlockArea`) then you would be better off accessing the `cChunk` object directly with `cChunk::GetBlock` etc.

In general, it's also very tricky to know for sure that a chunk will be valid. For example, a player being close means nothing if the server is under high load and the chunk generator is going slowly. Chunk stays also don't quite guarantee the chunk will be present, because of https://github.com/cuberite/cuberite/issues/3572#issuecomment-314722284.

src/Bindings/ManualBindings_World.cpp
```diff
@@ -997,6 +1057,8 @@ void cManualBindings::BindWorld(lua_State * tolua_S)
 			tolua_function(tolua_S, "SetSignLines",                 tolua_cWorld_SetSignLines);
 			tolua_function(tolua_S, "SpawnSplitExperienceOrbs",     tolua_cWorld_SpawnSplitExperienceOrbs);
 			tolua_function(tolua_S, "TryGetHeight",                 tolua_cWorld_TryGetHeight);
+			tolua_function(tolua_S, "GetHeight",                    tolua_cWorld_GetHeight);
+			tolua_function(tolua_S, "IsWeatherWetAtXYZ",            tolua_cWorld_IsWeatherWetAtXYZ);
```
madmaxoft, line 1061: Alpha-sort these into the list.

madmaxoft:
None of the optionals are in such a hot path that any of this would matter greatly. I'd say merge this PR in, then work on getting C++17 work, then remove the `optional` and `make_unique` scaffolding.

src/World.cpp
```diff
@@ -763,10 +767,56 @@ bool cWorld::CanSpawnAt(double a_X, double & a_Y, double a_Z)
...
+	cCanSpawnChunkStay ChunkStay(a_X, a_Z);
+	ChunkStay.Enable(*m_ChunkMap);
+	ChunkStay.Wait();
+
 	static const int ValidSpawnBlocksCount = ARRAYCOUNT(ValidSpawnBlocks);
 
 	// Increase this by two, because we need two more blocks for body and head
-	static const int HighestSpawnPoint = GetHeight(static_cast<int>(a_X), static_cast<int>(a_Z)) + 2;
+	auto Height = GetHeight(static_cast<int>(a_X), static_cast<int>(a_Z));
+	VERIFY(Height.has_value());
```
tigerw, line 818: ASSERT?

src/Bindings/ManualBindings_World.cpp
```diff
@@ -950,6 +949,68 @@ static int tolua_cWorld_TryGetHeight(lua_State * tolua_S)
...
+	}
+
+	// Get params:
+	cWorld * self = nullptr;
+	Vector3i BlockPos;
+	L.GetStackValues(1, self, BlockPos);
+
+	// Call the implementation:
+	auto IsWet = self->IsWeatherWetAtXYZ(BlockPos);
+	L.Push(IsWet.value_or(true));
```
bearbin, line 1006: _not_ raining strikes me as a more useful default, in any case this should be documented in the API docs.

peterbell10: It only fails if it's raining in the chunk but the terrain is not loaded so it can't tell where the rain is falling. On the client that would render as rain falling all the way to void i.e. wet everywhere.

peterbell10: I could also add a `Try` variant if that's desirable.

bearbin: I guess it doesn't really matter that much, the issue _shouldn't_ come up that often. It's just that everywhere else that IsWeatherWetAtXYZ is used, the default on nullopt is false, rather than true.

bearbin:
Closing, replaced by #5101.

std::optional will be used, but not in this incarnation (fallback to world rain state was used instead)

bearbin:
@peterbell10 I fixed the merge conflicts on this PR and stripped out the IsWeatherWet changes, minded to merge after I get my head around the spawnpoint generation change.

Mind giving this a look over to make sure there's nothing silly left?

peterbell10:
Well isn't this a blast from the past. Changes LGTM.

madmaxoft (approved):
Looks good to me, let's break stuff :D

bearbin (approved):

PR #4693, opened 2020-04-25
Corrected furnace recipes for log

Seadragon91:


tigerw:
Just to confirm this means the meta is ignored right? Looks good.

Seadragon91:
If -1 ignores the item damage in the check.

PR #4694, opened 2020-04-25
Fixed spawning of mobs from dispenser

Seadragon91:
The item damage from the spawn egg was used, but `SpawnMob` requires a `eMonsterType`. Use `cItemSpawnEggHandler::ItemDamageToMonsterType` to correct that problem.


tigerw:
Good catch!

Seadragon91:
Whoops, I oversaw the cast. it's unnecessary...

PR #4695, opened 2020-04-27
cWorld * -> cWorld

tigerw:
`map<string, cWorld *>` has mega evolved into `map<string, cWorld>`!

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -53,7 +53,6 @@ cRoot * cRoot::s_Root = nullptr;
 
 
 cRoot::cRoot(void) :
-	m_pDefaultWorld(nullptr),
 	m_Server(nullptr),
 	m_MonsterConfig(nullptr),
 	m_CraftingRecipes(nullptr),
@@ -306,9 +305,6 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	delete m_FurnaceRecipe;   m_FurnaceRecipe = nullptr;
 	delete m_CraftingRecipes; m_CraftingRecipes = nullptr;
 
-	LOG("Unloading worlds...");
-	UnloadWorlds();
-
 	LOGD("Stopping plugin manager...");
 	delete m_PluginManager; m_PluginManager = nullptr;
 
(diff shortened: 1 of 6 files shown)


src/Root.cpp
```diff
@@ -306,9 +305,6 @@ void cRoot::Start(std::unique_ptr<cSettingsRepositoryInterface> a_OverridesRepo)
 	delete m_FurnaceRecipe;   m_FurnaceRecipe = nullptr;
 	delete m_CraftingRecipes; m_CraftingRecipes = nullptr;
 
-	LOG("Unloading worlds...");
-	UnloadWorlds();
```
peterbell10, line 310: Are you sure this isn't needed any more? Yes, the `std::map` will delete the worlds but the ordering could be important.

tigerw: indeed, thanks!

tigerw:
appveyor ded

madmaxoft:
To be honest, I never trusted containers-of-values, never knew if there's the guarantee that the value will stay at one place or if the container is allowed to move it around in memory (thus invalidating pointers).

Additionally, constructs such as `.emplace(std::piecewise_construct, std::forward_as_tuple(` just make my head hurt, why can't it be simple?

All in all, I'd have prefferred a `std::shared_ptr` rewrite, and there are much worse offensors than `cWorld` - take the furnace recipes with their manual deletion :P

tigerw:
Ah, I think it can be a net benefit. The piecewise construct is annoying, we shall have to wait until C++17 to replace it. That said, we can be sure that the worlds never change in cRoot, invalidation begone!

tigerw:
xoft, any idea about application conveyor?

madmaxoft:
No idea about Appveyor, some internal failure; feel free to ask them in their forum :P

madmaxoft:
Appveyor support engaged: https://help.appveyor.com/discussions/problems/27033-pushartifact-fails-with-500-internal-server-error

tigerw:
Thanks!

Meanwhile, is that a go/no go from you to merge?

madmaxoft:
I'm still not convinced, could we compromise on using a `shared_ptr` instead? Other parts could then use `weak_ptr` for referencing the world.

tigerw:
Aw :(

There's nothing functionally wrong with the code as it is right now, so this change was intended only to reduce dynamic allocation, we don't have to do anything to it if you disagree in principle.

Are there any concerns that I can address to convince you more?

madmaxoft:
Okay, I went through the `std::map` documentation and it actually says that the relevant operations keep references valid, so this is valid code.
I still don't like it, in my opinion
```cpp
m_WorldsByName["world"] = std::make_shared<cWorld>("world", "world", a_dd, WorldNames);
```
is much clearer than
```cpp
m_WorldsByName.emplace(std::piecewise_construct, std::forward_as_tuple("world"), std::forward_as_tuple("world", "world", a_dd, WorldNames));
```
As for dynamic allocation, what are you trying to optimize here? The few allocation calls on server startup?

tigerw:
A few allocation calls yes, but mainly reducing the "visual load" of someone reading the code so they don't hit a speedbump when they see that `new` and wonder about memory ownership.

However I agree the emplace call is suboptimal.

Would you consider accepting this PR once we enable C++17 and can then replace the call with `try_emplace("world", "world", dd, names)`? In the meantime I can park the changes in a branch.

madmaxoft:
Sounds good. Perhaps extract the unrelated changes to a separate PR so that they get merged sooner.

tigerw:
Ref: #4771

PR #4697, opened 2020-04-29
NBT: Dynamic list-max-count protection.

### LINKED ISSUE #4696: NBT parsing failed when many furnaces are placed and server is restarted
Client version: 1.12.2
Server OS: Linux

### Actual behavior
Server regenerates chunk, when server is restarted

I attached a zip file with the [world](https://github.com/cuberite/cuberite/files/4547452/world.zip) that causes the error:
```
Loading chunk [-2, 1] for world world from file r.-1.0.mca failed: NBT parsing failed. Offloading old chunk data to file world/region/badchunks/ch.-2.1.2020-04-28-19-22-23.dat and regenerating chunk
```
Use worldedit to place many furnaces. Restart the server and the error appears.

madmaxoft:
Loosened the check agains invalid / malicious NBT list size, from a constant maximum to a maximum of `BytesLeft / 4` items.

Fixes #4696

### DIFF
--- a/src/WorldStorage/FastNBT.cpp
+++ b/src/WorldStorage/FastNBT.cpp
@@ -443,6 +443,7 @@ size_t cParsedNBT::GetMinTagSize(eTagType a_TagType)
 {
 	switch (a_TagType)
 	{
+		case TAG_End:       return 1;
 		case TAG_Byte:      return 1;
 		case TAG_Short:     return 2;
 		case TAG_Int:       return 4;
@@ -455,6 +456,7 @@ size_t cParsedNBT::GetMinTagSize(eTagType a_TagType)
 		case TAG_Compound:  return 1;  // Single TAG_End byte
 		case TAG_IntArray:  return 4;  // 4 bytes for the count
 	}
+	UNREACHABLE("Unsupported nbt tag type");
 }
 
 
--- a/src/WorldStorage/FastNBT.cpp
+++ b/src/WorldStorage/FastNBT.cpp
@@ -10,13 +10,6 @@
 
 
 
-/** If a list being loaded has more than this number of items, it's considered corrupted. */
-static const int MAX_LIST_ITEMS = 10000;
-
-
-
-
-
 // The number of NBT tags that are reserved when an NBT parsing is started.
 // You can override this by using a cmdline define
 #ifndef NBT_RESERVE_SIZE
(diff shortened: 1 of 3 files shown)


src/WorldStorage/FastNBT.cpp
```diff
@@ -257,7 +250,7 @@ eNBTParseError cParsedNBT::ReadList(eTagType a_ChildrenType)
 	NEEDBYTES(4, eNBTParseError::npListMissingLength);
 	int Count = GetBEInt(m_Data + m_Pos);
 	m_Pos += 4;
-	if ((Count < 0) || (Count > MAX_LIST_ITEMS))
+	if ((Count < 0) || (Count > static_cast<int>((m_Length - m_Pos) / 4)))
```
peterbell10, line 253: Why 4? Should this depend on `a_ChildrenType` or is 4 a lower bound?

madmaxoft: No specific reason. I wanted to keep it simple to read. But you're right, it's bound to fail randomly in the future, I'll rewrite it to take the children size into account.

src/WorldStorage/FastNBT.cpp
```diff
@@ -445,6 +439,28 @@ int cParsedNBT::FindTagByPath(int a_Tag, const AString & a_Path) const
...
+		case TAG_Int:       return 4;
+		case TAG_Long:      return 8;
+		case TAG_Float:     return 4;
+		case TAG_Double:    return 8;
+		case TAG_String:    return 2;  // 2 bytes for the string length
+		case TAG_ByteArray: return 4;  // 4 bytes for the count
+		case TAG_List:      return 5;  // 1 byte list type + 4 bytes count
+		case TAG_Compound:  return 1;  // Single TAG_End byte
+		case TAG_IntArray:  return 4;  // 4 bytes for the count
+	}
```
peterbell10, line 457: Add `UNREACHABLE` after the switch to silence clang warnings.

madmaxoft: Actually the warning was worth it, there was one type missing; verifying a new commit now.

tigerw:
Am I correct in saying that the code now essentially does an "ARRAYCOUNT" on the remaining length (?) of the file, with the type being a_ChildrenType, and fails if the Count specified on-disk exceeded what could possibly be held in the space available?

madmaxoft:
Yes, basically.

PR #4698, opened 2020-04-30
Cleanup in CanBeAt functions

ChaosCat:
* Unified cacti surrounding logic instead of copy-pasted code
* Added total light calculation for mushrooms
* Corrected height checks 
* Converted nearby block checks confined to chunk to use the unbounded variant and bail out if the chunk is not loaded as per expected behavior


madmaxoft:
What's going on here, why are @bearbin's commits listed in this PR?

ChaosCat:
Whoops, accidentally PRed the opposite target and source branch. I'll switch it right away

Issue #4677, opened 2020-04-18
BlockHandlers: Verify CanBeAt checks


madmaxoft:
Many of the blocks check in their `CanBeAt` function if they are attached to a fully solid block; some blocks include a check for the top-side of a top-half-slab, some blocks include a check for the bottom side of a bottom-half-slab. Some blocks only check for air. Most of these should probably use the very same logic, they most likely just weren't "found out" as buggy yet.

All the `CanBeAt` functions should be revised and if many of them use the same "solid face" check, this check should be pulled into a separate helper function in `cBlockHandler`.

ChaosCat:
Hello! I discovered the repo yesterday and would like to contribute. I'd like to take this one on

madmaxoft:
Good luck :)

ChaosCat:
Ok so I've finished going over all the implemented CanBeAt-s and this is a summary of things I noticed = there that probably should be changed (A full list in an evernote note):

- abstracting some of the functions (getSurface block instead of rel-pos.y - 1
  - Supporting blacklist/whitelist for more dynamic code design
- Main dispatcher for block checks
- Unify height checks
- Use consts for bitmasking on block meta information
- Some end-of-chunk checks

Thus I think that it may be appropriate to implement the following general "dispatcher" function
which would forward the per-block calls to this one with the appropriate parameters
- Precompiled lists of whitelists/blacklists in const formats (For block validation)
- Mandatory global height validation (flexible range for edge cases)
- ChunkEdges check switch (for stuff that can't be animated)
- General surrounding function with block whitelist/blacklist
- An optional list of bool callbacks for additional really special checks in case (this would usually be empty/null)

I think it's most appropriate to be implemented in cBlockHander as a generic helper function, not sure about where to get the consts and pre-defined white/blacklists. This would cause a very little overhead and would make it much easier to add logic to these kinds of blocks while encapsulating most of the repetitive logic.

madmaxoft:
I admit I have no idea what you mean by "dispatcher". But since each block type already has a virtual `CanBeAt` function, I'd call that a dispatcher, no need to go from a virtual call back to a base class just to be switched on the blocktype again.

This issue was meant more for cleaning things up - many blocks check for solidity below them, but do so using various methods, some of them incomplete. So I'd extract the `IsBlockBelowSolid` function into the base `cBlockHandler` and make the descendants call that.

[...]

PR #4699, opened 2020-04-30
Cleanup in CanBeAt functions

ChaosCat:
* Minor corrections
* Validate the unbounded checks
* Add composite light threshold check for mushroom placement (still behaves a little different from oiginal)
* Use proper height boundary checks
* Unify cacti neighbour check logic
* Type declare unnamed `auto` variables

referencing #4677


src/Blocks/BlockTripwireHook.h
```diff
@@ -83,14 +83,20 @@ class cBlockTripwireHookHandler :
...
+		NIBBLETYPE Meta = a_Chunk.GetMeta(a_RelPos);
+		Vector3i NeighborPos = AddFaceDirection(a_RelPos, MetadataToDirection(Meta), true);
+
 		if (!cChunkDef::IsValidHeight(NeighborPos.y))
 		{
 			return false;
 		}
+
 		BLOCKTYPE NeighborBlockType;
-		a_Chunk.UnboundedRelGetBlockType(a_RelPos, NeighborBlockType);
```
ChaosCat, line 93: On fail this would return false (0) since it's not halting here NeighbotBlockType could be 0 (which is air block type), and this would cause the return to succeed, potentially causing a bug of placing tripwire hooks on air.

src/Blocks/BlockButton.h
```diff
@@ -145,14 +145,19 @@ class cBlockButtonHandler :
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		auto Meta = a_Chunk.GetMeta(a_RelPos);
-		auto SupportRelPos = AddFaceDirection(a_RelPos, BlockMetaDataToBlockFace(Meta), true);
+		NIBBLETYPE Meta = a_Chunk.GetMeta(a_RelPos);
+		Vector3i SupportRelPos = AddFaceDirection(a_RelPos, BlockMetaDataToBlockFace(Meta), true);
```
peterbell10, line 149: Why are you removing `auto`?

src/Blocks/BlockButton.h
```diff
@@ -145,14 +145,19 @@ class cBlockButtonHandler :
...
 		if (!cChunkDef::IsValidHeight(SupportRelPos.y))
 		{
 			return false;
 		}
+
 		BLOCKTYPE SupportBlockType;
-		a_Chunk.UnboundedRelGetBlockType(SupportRelPos, SupportBlockType);
+		if (!a_Chunk.UnboundedRelGetBlockType(SupportRelPos, SupportBlockType))
+		{
+			return false;
```
peterbell10, line 159: We shouldn't fail the check just because the chunk hasn't loaded yet. So, this should return `true` so the block isn't destroyed. Same for all other `UnboundedRelGet` failures.

src/Blocks/BlockButton.h
ChaosCat: Personally getting into the code for the first time this auto confused me for a bit. Why use auto for trivial types? especially when forwarded to other functions as parameters.
It's also inconsistent with the other CanBeAt functions.

src/Blocks/BlockButton.h
ChaosCat: But `bool cChunk::UnboundedRelGetBlockType` would leave out the BlockType on it's undefined value (or 0 depending on compiler behaviour). Wouldn't this cause a bug of buttons being placed on air outside the chunk? or worse, statistically?

src/Blocks/BlockBigFlower.h
```diff
@@ -97,13 +97,15 @@ class cBlockBigFlowerHandler:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		if (a_RelPos.y <= 0)
+		if (!cChunkDef::IsValidHeight(a_RelPos.y))
```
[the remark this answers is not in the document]
ChaosCat: Hmm.. doesn't this implementation capture it correctly?
```
inline static bool IsValidHeight(int a_Height)
{
	return ((a_Height >= 0) && (a_Height < Height));
}
```
0 is still valid in this context, just the upper bound of 0x100 added and made syntactically nicer

src/Blocks/BlockButton.h
peterbell10: I'm not saying don't check the return value. I'm saying to `return true`.

src/Blocks/BlockBigFlower.h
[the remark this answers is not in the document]
peterbell10: No, `y <= 0` is not equivalent to `!(y >= 0)`.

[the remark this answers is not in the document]
peterbell10: And 0 is not valid because `GetBlock(a_RelPos.addedY(-1))` would access a block outside of the chunk at `y = -1`.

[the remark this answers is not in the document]
ChaosCat: Wow... embarrassed it took me 3 checks to see this. A simple a_RelPos.y - 1 would suffice I suppose

src/Blocks/BlockButton.h
ChaosCat: Sorry I'm not really acquainted with the behaviour. wouldn't this still allow the button (and others) to be placed if true is returned? (I'll change this too). The behaviour I imagined was placing a button on an unloaded chunk edge would cause the button to just be dropped as happens when false is returned but I guess my intuition was wrong.

tigerw: literal edge case

I'm not sure how the player will manage to _place_ a button such that the support block is in an unloaded chunk though, given that they have nothing to click on?

src/Blocks/BlockTorch.h
```diff
@@ -192,13 +192,18 @@ class cBlockTorchHandler:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		auto Face = MetaDataToBlockFace(a_Chunk.GetMeta(a_RelPos));
-		auto NeighborRelPos = AddFaceDirection(a_RelPos, Face, true);
+		if (!cChunkDef::IsValidHeight(a_RelPos.y - 1))
+		{
+			return false;
```
tigerw, line 197: I believe that torches can be at Y=0 since they can attach to the side of a block, right?

src/Blocks/BlockButton.h
peterbell10: Imagine a button placed against a solid block in a different chunk. The chunk with the solid block can unload, leaving the button against an unloaded chunk. That button should not drop.

src/Blocks/BlockTorch.h
ChaosCat: Same goes for levers/buttons/tripwirehook I guess. While also a check for attached edge face from meta could be done to validate not being placed on "-1" but I think the overhead for this also edge case is ridiculous.

src/Blocks/BlockButton.h
ChaosCat: Ohhh I didn't know this is called also when chunks are brought to scope and thought it only related to player's action of actually placing a block. I should take a broader look at the code.
On it

peterbell10: `CanBeAt` is called by default from `cBlockHandler::Check` which is called whenever the block or one of its neighbours is updated in any way. That could include things like water flowing which doesn't require player input.

src/Blocks/BlockCactus.h
```diff
@@ -50,12 +35,12 @@ class cBlockCactusHandler :
...
 			if (
 				a_Chunk.UnboundedRelGetBlock(a_RelPos + Coords[i], BlockType, BlockMeta) &&
-				(
+					(
 					cBlockInfo::IsSolid(BlockType) ||
 					(BlockType == E_BLOCK_LAVA) ||
 					(BlockType == E_BLOCK_STATIONARY_LAVA)
+					)
 				)
-			)
```
peterbell10, line 58: You messed up the indentation here.

src/Blocks/BlockCarpet.h
```diff
@@ -50,7 +50,15 @@ class cBlockCarpetHandler:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		return (a_RelPos.y > 0) && (a_Chunk.GetBlock(a_RelPos.addedY(-1)) != E_BLOCK_AIR);
+		if (!cChunkDef::IsValidHeight(a_RelPos.y - 1))
```
peterbell10, line 53: Writing the same thing but with more lines is not a valuable change IMO.

src/Blocks/BlockChest.h
```diff
@@ -84,7 +84,12 @@ class cBlockChestHandler :
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		auto BlockPos = a_Chunk.RelativeToAbsolute(a_RelPos);
+		if (!cChunkDef::IsValidHeight(a_RelPos.y - 1))
+		{
+			return false;
+		}
```
peterbell10, line 90: This doesn't belong here, we aren't reading any block at `a_RelPos.y - 1`.

src/Blocks/BlockCocoaPod.h
```diff
@@ -24,12 +24,21 @@ class cBlockCocoaPodHandler :
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
+		if (!cChunkDef::IsValidHeight(a_RelPos.y - 1))
```
peterbell10, line 27: Same thing, we aren't reading a block at `a_RelPos.y - 1`.

src/Blocks/BlockMushroom.h
```diff
@@ -32,14 +34,22 @@ class cBlockMushroomHandler:
...
 	{
-		if (a_RelPos.y <= 0)
+		if (!cChunkDef::IsValidHeight(a_RelPos.y - 1))
+		{
+			return false;
+		}
+
+		// Cannot be at too much light
+		NIBBLETYPE TotalLight = a_Chunk.GetSkyLightAltered(a_RelPos) + a_Chunk.GetBlockLight(a_RelPos);
+		if (TotalLight > MUSHROOM_MAX_LIGHT_THRESHOLD)
```
peterbell10, line 44: Is this how mushrooms work? Usually `TotalLight` would be `std::max(SkyLightAltered, BlockLight)`.

src/Blocks/BlockCarpet.h
peterbell10: A lot of this PR seems to just be writing the same thing more verbosely.

ChaosCat: > A lot of this PR seems to just be writing the same thing more verbosely.

Well the goal was refactoring the CanBeAts and I do believe that more verbose code here albeit less idiomatic, more friendly to readers. Especially since most of these functions implement much logic, but if there is a preferred idiom coding style I'll go ahead and revert. Nothing was mentioned about that in the guidelines so I had the trigger loose on these changes.

ChaosCat:
On a side note: All of this logic really needs unit tests, especially for all the aforementioned edge cases.
Also a lot of the code is repetitive, maybe an interface that abstracts many of the checks should be considered? It would also be much easier to add future blocks with some common standard checks (Blocks in radius, blockBelow whitelist instead of switch, etc.)

src/Blocks/BlockCocoaPod.h
ChaosCat: Yep, the -1 slipped through with the previous re-amendments

src/Blocks/BlockChest.h
ChaosCat: Fixed

src/Blocks/BlockMushroom.h
ChaosCat: You're right, I used that at first, not sure why I changed it like that (light can be 15 at max anyways). note that denying the placement like that is still not like the vanilla behaviour. It drops when placed instead of just blocking the placement.

tigerw:
Please bear with, I would like to test the changes out.

src/Blocks/BlockCactus.h
```diff
@@ -65,6 +50,32 @@ class cBlockCactusHandler :
...
+			return false;
+		}
+
+		// Check surroundings. Cacti may ONLY be surrounded by non-solid blocks
+		if (!ValidateBlocksSurroundingCactus(a_RelPos, a_Chunk))
+		{
+			return false;
+		}
+
+		return true;
```
peterbell10, line 75:
```suggestion
		return ValidateBlocksSurroundingCactus(a_RelPos, a_Chunk);
```

src/Blocks/BlockChest.h
```diff
@@ -84,7 +84,12 @@ class cBlockChestHandler :
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		auto BlockPos = a_Chunk.RelativeToAbsolute(a_RelPos);
+		if (!cChunkDef::IsValidHeight(a_RelPos.y))
```
peterbell10, line 87: This doesn't make any sense. If it wasn't a valid height then how could there be a block there already?

src/Blocks/BlockCarpet.h
peterbell10: I prefer the terser code here because the check itself is quite simple.  
"The block below me cannot be air".  
You might have to read a few of these functions for that to jump out at you.

src/Blocks/BlockChest.h
ChaosCat: The idea with these extra checks is just to be 100% absolutely sure that there's no chance for exploitation of the block placement mechanics. So assuming a player was able to place a block there in some weird way (say bug in command blocks or other handler). It would immediately be disposed of on the next CanBeAt update. 
Maybe I'm exaggerating here... but the idea was just to regard the handlers as stateless as possible.

I don't know if there's a culling mechanism implemented but I believe it may be a better idea to somehow generalize this height culling, applying it to all blocks instead of manually checking in each handle. Blocks of any type shouldn't ever be beyond these bounds after all.

src/Blocks/BlockBigFlower.h
```diff
@@ -97,13 +97,15 @@ class cBlockBigFlowerHandler:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		if (a_RelPos.y <= 0)
+		if (!cChunkDef::IsValidHeight(a_RelPos.y - 1))
 		{
 			return false;
 		}
+
+		Vector3i BelowBlockPos = a_RelPos.addedY(-1);
```
bearbin, line 105: move this declaration before the call to IsValidHeight and use `BlockBelowPos.y` as the parameter to that call.

src/Blocks/BlockButton.h
bearbin: Auto has advantages even with trivial types, it prevents unintentional casts when the type on the LHS is not the same as the RHS type.

Your supposed confusion will only be present on literally the first read-through of the code, just use auto everywhere it works.

src/Blocks/BlockCarpet.h
bearbin: I don't have any preference for either of these two constructions, but there is an improvement that could be made here - with @ChaosCat's code BlockBelowPos can be defined before the height check, so that DRY is maintained and there's no chance of the two heights getting out of sync in future.

src/Blocks/BlockChest.h
bearbin: Yes, this makes absolutely no sense knowing where these functions are actually used - you'll never get a block out of bounds when iterating through the chunk data.

In any case, this isn't the place for a check, that should be in the base cBlockHandler CanBeAt function, which the derived classes should call into.

src/Blocks/BlockCocoaPod.h
```diff
@@ -24,12 +24,21 @@ class cBlockCocoaPodHandler :
...
 	{
+		if (!cChunkDef::IsValidHeight(a_RelPos.y))
+		{
+			return false;
+		}
+
 		// Check that we're attached to a jungle log block:
 		eBlockFace BlockFace = MetaToBlockFace(a_Chunk.GetMeta(a_RelPos));
-		auto LogPos = AddFaceDirection(a_RelPos, BlockFace, true);
+		Vector3i LogPos = AddFaceDirection(a_RelPos, BlockFace, true);
```
bearbin, line 34: Again, please don't replace auto with a fixed type.

src/Blocks/BlockComparator.h
```diff
@@ -74,19 +74,21 @@ class cBlockComparatorHandler :
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
```
bearbin, line 75: It seems like this implements a very common behaviour -> requiresSolidBlockBelow. This should be extracted into a helper function or BlockHandler mixin.

src/Blocks/BlockTorch.h
```diff
@@ -192,14 +192,20 @@ class cBlockTorchHandler:
 
 	virtual bool CanBeAt(cChunkInterface & a_ChunkInterface, const Vector3i a_RelPos, const cChunk & a_Chunk) override
 	{
-		auto Face = MetaDataToBlockFace(a_Chunk.GetMeta(a_RelPos));
-		auto NeighborRelPos = AddFaceDirection(a_RelPos, Face, true);
+
```
bearbin, line 195: This line has no purpose, remove it.

bearbin:
Closing as unmergeable due to age, I have opened a new PR with the same ideas in #5587.

Issue #4703, opened 2020-05-02
Crash in cClientHandle::HandleUseEntity


Seadragon91:
Client version: 1.12.2
Server OS: Windows
Cuberite Commit id: 0b9b7bc1a8d5cf6f92b802dc376a189ef066e62d

### Steps to reproduce the behavior
Spammed the server with "use entity" packets with pycraft.

Crashed occurred here (m_World" was "nullptr"):
https://github.com/cuberite/cuberite/blob/0b9b7bc1a8d5cf6f92b802dc376a189ef066e62d/src/ClientHandle.cpp#L1758

### Stack trace
```
[Inlineframe] Cuberite.exe!cClientHandle::HandleUseEntity::__l2::<Lambda>(cEntity &) Zeile 1758	C++
Cuberite.exe!cFunctionRef<bool __cdecl(cEntity &)>::ObjectFunctionCaller<bool <Lambda>(cEntity &) >(void * a_Callable, cEntity & <a_Args_0>) Zeile 71	C++
[Inlineframe] Cuberite.exe!cFunctionRef<bool __cdecl(cEntity &)>::operator()(cEntity &) Zeile 57	C++
Cuberite.exe!cWorld::DoWithEntityByID(unsigned int a_UniqueID, cFunctionRef<bool __cdecl(cEntity &)> a_Callback) Zeile 2798	C++
Cuberite.exe!cProtocol_1_9_0::HandlePacketUseEntity(cByteBuffer & a_ByteBuffer) Zeile 985	C++
Cuberite.exe!cProtocol_1_12_1::HandlePacket(cByteBuffer & a_ByteBuffer, unsigned int a_PacketType) Zeile 1321	C++
Cuberite.exe!cProtocol_1_8_0::AddReceivedData(const char * a_Data, unsigned __int64 a_Size) Zeile 1998	C++
Cuberite.exe!cProtocol_1_8_0::DataReceived(const char * a_Data, unsigned __int64 a_Size) Zeile 211	C++
Cuberite.exe!cProtocolRecognizer::DataReceived(const char * a_Data, unsigned __int64 a_Size) Zeile 69	C++
Cuberite.exe!cClientHandle::ProcessProtocolInOut() Zeile 3365	C++
Cuberite.exe!cClientHandle::Tick(float a_Dt) Zeile 2096	C++
Cuberite.exe!cWorld::TickClients(float a_Dt) Zeile 1276	C++
[...]

PR #4706, opened 2020-05-02
Set entity world early

mathiascode:
In hindsight, I'm not happy with https://github.com/cuberite/cuberite/pull/4606. It seems too dangerous, and plugins probably have to possibility to crash the server while the world is unset.

Closes https://github.com/cuberite/cuberite/issues/4703

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -1025,6 +1025,7 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
 	for (auto & Entity : EntitiesToAdd)
 	{
 		auto EntityPtr = Entity.get();
+		ASSERT(EntityPtr->GetWorld() == this);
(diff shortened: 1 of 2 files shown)


tigerw:
looks good

src/World.cpp
```diff
@@ -1024,7 +1024,6 @@ void cWorld::Tick(std::chrono::milliseconds a_Dt, std::chrono::milliseconds a_La
 	}
 	for (auto & Entity : EntitiesToAdd)
 	{
-		Entity->SetWorld(this);
```
peterbell10, line 1027: Can you assert the world is set correctly.

mathiascode:
Did not trigger in my testing.
@Seadragon91 Could you confirm that this fixes your issue?

Seadragon91:
Did a few other tests, too. Had a few deadlocks during testing, but they are already known.

PR #4707, opened 2020-05-02
Modify pickup collection behaviour to correspond to vanilla.

bearbin:
As documented here: https://www.spigotmc.org/threads/item-pickup-radius.337271/#post-3141146

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1643,9 +1643,8 @@ void cChunk::SetAreaBiome(int a_MinRelX, int a_MaxRelX, int a_MinRelZ, int a_Max
 
 void cChunk::CollectPickupsByPlayer(cPlayer & a_Player)
 {
-	double PosX = a_Player.GetPosX();
-	double PosY = a_Player.GetPosY();
-	double PosZ = a_Player.GetPosZ();
+	auto BoundingBox = cBoundingBox(a_Player.GetPosition(), a_Player.GetWidth(), a_Player.GetHeight());
+	BoundingBox.Expand(1, 0.5, 1);
 
 	for (auto & Entity : m_Entities)
 	{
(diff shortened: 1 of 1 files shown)


tigerw:
Thank you for fixing this, it was unbelievably unsatisfying when testing arrows that shooting them into the ceiling made them uncollectable.

src/Chunk.cpp
```diff
@@ -1643,21 +1643,17 @@ void cChunk::SetAreaBiome(int a_MinRelX, int a_MaxRelX, int a_MinRelZ, int a_Max
 
 void cChunk::CollectPickupsByPlayer(cPlayer & a_Player)
 {
-	double PosX = a_Player.GetPosX();
-	double PosY = a_Player.GetPosY();
-	double PosZ = a_Player.GetPosZ();
+	auto BoundingBox = cBoundingBox(a_Player.GetPosition(), a_Player.GetWidth(), a_Player.GetHeight());
```
peterbell10, line 1646: Side note: there should totally be a `cEntity::GetBoundingBox`.

bearbin: I thought of that, but there's only one other place in the whole codebase where it would be used. Still worth it?

peterbell10: I'd say yes. It also looks like it should actually be `GetWidth()/2.0`:
https://github.com/cuberite/cuberite/blob/0b9b7bc1a8d5cf6f92b802dc376a189ef066e62d/src/Entities/Minecart.cpp#L57

peterbell10: Unless that was intentionally twice the entity radius?

bearbin: No, it should just have been the entity radius. I guess with only 0.4 blocks difference it behaves basically the same anyway. I'll make a new PR to fix that and create a new method to get the bounding box.

PR #4708, opened 2020-05-03
Upgrade to C++17 [Submodules]

tigerw:
* Fixed multiple places falling afoul of the one definition rule
* Fixed incorrect format logging
* Fixed all our cmake
* Upgrade submodules to not be 4 years out of date

Improves build times! I got a 37% improvement, your results may vary.

Please merge it quickly... I can't imagine the conflicts on this thing. Plus, I spent so long on it :)

### DIFF
--- a/src/Bindings/ManualBindings.h
+++ b/src/Bindings/ManualBindings.h
@@ -53,9 +53,9 @@ public:
 	static int tolua_do_error(lua_State * L, const char * a_pMsg, tolua_Error * a_pToLuaError);
 	static int vlua_do_error(lua_State * L, const char * a_pFormat, fmt::printf_args a_ArgList);
 	template <typename... Args>
-	static int lua_do_error(lua_State * L, const char * a_Format, const Args & ... args)
+	static int lua_do_error(lua_State * L, const char * a_Format, const Args & ... a_Args)
 	{
-		return vlua_do_error(L, a_Format, fmt::make_printf_args(args...));
+		return vlua_do_error(L, a_Format, fmt::make_printf_args(a_Args...));
 	}
 
 
--- a/src/OSSupport/File.cpp
+++ b/src/OSSupport/File.cpp
@@ -695,10 +695,10 @@ AString cFile::GetExecutableExt(void)
 
 
 
-int cFile::vPrintf(const char * a_Fmt, fmt::printf_args a_ArgList)
+int cFile::vPrintf(const char * a_Format, fmt::printf_args a_ArgList)
 {
 	fmt::memory_buffer Buffer;
-	fmt::vprintf(Buffer, fmt::to_string_view(a_Fmt), a_ArgList);
+	fmt::vprintf(Buffer, fmt::to_string_view(a_Format), a_ArgList);
 	return Write(Buffer.data(), Buffer.size());
 }
 
@@ -710,7 +710,3 @@ void cFile::Flush(void)
 {
 	fflush(m_File);
 }
-
-
-
-
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -52,34 +52,34 @@ static unsigned char HexToDec(char a_HexChar)
 
 
 
-AString & vPrintf(AString & str, const char * format, fmt::printf_args args)
+AString & vPrintf(AString & a_String, const char * a_Format, fmt::printf_args a_ArgList)
 {
-	ASSERT(format != nullptr);
-	fmt::memory_buffer Buffer;
-	fmt::vprintf(Buffer, fmt::to_string_view(format), args);
-	str.assign(Buffer.data(), Buffer.size());
-	return str;
+	ASSERT(a_Format != nullptr);
+	fmt::memory_buffer Buffer;  // Save a string allocation compared to vsprintf
+	fmt::vprintf(Buffer, fmt::to_string_view(a_Format), a_ArgList);
+	a_String.assign(Buffer.data(), Buffer.size());
+	return a_String;
 }
 
 
 
 
 
-AString vPrintf(const char * format, fmt::printf_args args)
+AString vPrintf(const char * a_Format, fmt::printf_args a_ArgList)
 {
-	ASSERT(format != nullptr);
-	return fmt::vsprintf(format, args);
+	ASSERT(a_Format != nullptr);
+	return fmt::vsprintf(a_Format, a_ArgList);
 }
 
 
 
 
 
-AString & vAppendPrintf(AString & a_String, const char * format, fmt::printf_args args)
+AString & vAppendPrintf(AString & a_String, const char * a_Format, fmt::printf_args a_ArgList)
 {
-	ASSERT(format != nullptr);
+	ASSERT(a_Format != nullptr);
 	fmt::memory_buffer Buffer;
-	fmt::vprintf(Buffer, fmt::to_string_view(format), args);
+	fmt::vprintf(Buffer, fmt::to_string_view(a_Format), a_ArgList);
 	a_String.append(Buffer.data(), Buffer.size());
 	return a_String;
 }
--- a/src/Vector3.h
+++ b/src/Vector3.h
@@ -400,26 +400,25 @@ public:
 
 
 
-namespace fmt
-{
 
+/** Allows formatting a Vector<T> using the same format specifiers as for T
+e.g. `fmt::format("{0:0.2f}", Vector3f{0.0231f, 1.2146f, 1.0f}) == "{0.02, 1.21, 1.00}"` */
 template <typename What>
-class formatter<Vector3<What>>:
-	public fmt::formatter<What>
+class fmt::formatter<Vector3<What>> : public fmt::formatter<What>
 {
 	using Super = fmt::formatter<What>;
 
 	template <typename FormatContext, size_t Len>
 	void Write(FormatContext & a_Ctx, const char (& a_Str)[Len])
 	{
-		auto Itr = std::copy_n(&a_Str[0], Len - 1, a_Ctx.out());
+		const auto Itr = std::copy_n(&a_Str[0], Len - 1, a_Ctx.out());
 		a_Ctx.advance_to(Itr);
 	}
 
 	template <typename FormatContext>
 	void Write(FormatContext & a_Ctx, const What & a_Arg)
 	{
-		auto Itr = Super::format(a_Arg, a_Ctx);
+		const auto Itr = Super::format(a_Arg, a_Ctx);
 		a_Ctx.advance_to(Itr);
 	}
 
@@ -439,8 +438,6 @@ public:
 		return a_Ctx.out();
 	}
 };
-
-}
 
 
 
--- a/Tools/ProtoProxy/Connection.cpp
+++ b/Tools/ProtoProxy/Connection.cpp
@@ -10,8 +10,6 @@
 #include "mbedTLS++/CryptoKey.h"
 #include "../../src/Logger.h"
 
-#include "fmt/printf.h"
-
 #ifdef _WIN32
 	#include <direct.h>  // For _mkdir()
 #endif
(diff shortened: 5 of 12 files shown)


CMakeLists.txt
```diff
@@ -13,321 +13,51 @@
...
-	# Check for clang version 3.4 or greater
-	if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND DUMPED_COMPILER_VERSION VERSION_LESS "3.4")
-		message(FATAL_ERROR "You have ${CMAKE_CXX_COMPILER_ID} version ${DUMPED_COMPILER_VERSION}, but at least 3.4 is needed")
-	endif()
-endif()
+cmake_minimum_required (VERSION 3.16)
+project(
+	Cuberite
+	DESCRIPTION "A lightweight, fast and extensible game server for Minecraft"
+	HOMEPAGE_URL "http://cuberite.org"
```
mathiascode, line 20:
```suggestion
	HOMEPAGE_URL "https://cuberite.org"
```

src/Bindings/ManualBindings.h
```diff
@@ -51,8 +51,26 @@ class cManualBindings
 	// Helper functions:
 	static cPluginLua * GetLuaPlugin(lua_State * L);
 	static int tolua_do_error(lua_State * L, const char * a_pMsg, tolua_Error * a_pToLuaError);
-	static int lua_do_error(lua_State * L, const char * a_pFormat, fmt::ArgList a_ArgList);
-	FMT_VARIADIC(static int, lua_do_error, lua_State *, const char *)
+
+	template <typename... Args>
+	static int lua_do_error(lua_State * L, const char * a_pFormat, const Args & ... a_ArgList)
```
peterbell10, line 56: We used `fmt::ArgList` for a reason, this shouldn't be a template. The new equivalent appears to be `fmt::format_args`.

CMakeLists.txt
Titaniumtown: Got to have that security.

src/LoggerSimple.h
```diff
@@ -2,33 +2,59 @@
 // Logging free functions defined in Logger.cpp
 #pragma once
 
-// python style format specified logging
-
-extern void FLOG(const char * a_Format, fmt::ArgList a_ArgList);
-FMT_VARIADIC(void, FLOG, const char *)
+#include "Logger.h"
```
peterbell10, line 5: Literally the entire point of `LoggerSimple.h` is that it __doesn't__ include `Logger.h`.

src/Bindings/ManualBindings.h
peterbell10: Also, only `fmt/core.h` should be included globally.

tigerw: Thanks, but I'm unsure how to use this, it seems the FMT_VARIADIC macro was removed in favour of templates? What was the reason to use ArgList?

src/LoggerSimple.h
tigerw: Depending on how the above discussion goes (meta!) using templates may require it.

src/Bindings/ManualBindings.h
tigerw: And besides, it is known that templates = ***SPEED***

tigerw:
Sorry, I know it's a pain to look at big diffs. It's a frustrating story of things breaking and fixing them as I went along.

src/Bindings/ManualBindings.h
peterbell10: `fmt` is designed to minimised the code templated on the input argument types in order to reduce binary size. In our case it also means we don't need to include all of `fmt` everywhere, just `fmt/core.h`.

```cpp
void Log(const char * a_Format, fmt::format_args);

template <typename Args...>
void Log(const char * a_Format, const Args & ... args)
{
	Log(a_Format, fmt::make_format_args(args...));
}
```

peterbell10:
`fmt` claims to support GCC 4.8, Clang 3.0 and MSVC 19.0 (2015) with C++11 only. So, I suggest updating `fmt` and appveyor in its own PR.

At the bare minimum, only change the file structure of the CMake code _after_ you've made the changes in place so that the commit diff is at least readable.

Many of the dependencies were forked by us to fix upstream bugs. Are you sure those patches are no longer necessary?

Why are you enabling unity builds? That clearly isn't necessary for C++17 and in amongst the other changes I have no idea what you've even changed in order to do that.

tigerw:
e.g. one of the issues I had https://github.com/fmtlib/fmt/issues/1658

I looked at jasoncpp (only behind upstream) sqlitecpp (same) libevent (a few cherry-picked changes by the maintainer wrt compiling, but the PR has the latest anyway).

With the magic of cmake unity builds are `set_target_properties(cuberite UNITY_BUILDS ON)`.

src/Bindings/ManualBindings.h
tigerw: I couldn't get this compiling, it claims that it can't find a formatter for Vector3. Honestly I have no idea why.

I cannot repreoduce binary size claims, for me:

Windoze
master MSVC Release: 7.63MB
this PR MSVC Release: 5.37MB

Leenucks
master Ubuntu Bionic Release: 11MB
this PR Ubuntu Bionic Release: 8.2MB

This seems to suggest that the changes are worth it.

bearbin:
This PR is indeed too big. It should be at least 3 stages:

 - Update dependencies (don't change remote to upstream, merge upstream into our repos - this has the advantage of not breaking all the cloned repos submodules)
 - Update cmake and CI systems
 - Actual C++17 changes.

src/Bindings/ManualBindings.h
peterbell10: Okay, I think this is what we want:
https://gcc.godbolt.org/z/AAzUog

Turns out that the `fmt::format_args` version can't have the same name as the template. Renaming it `vLog` did the trick.

Also note that the `printf`-like functions now have their own args class. `fmt::printf_args` and the corresponding `fmt::make_printf_args` which requires `fmt/printf.h` to be included.

tigerw:
The C++17 changes need to be merged at the same time.

.gitmodules
```diff
@@ -9,23 +9,23 @@
 	url = https://github.com/cuberite/ChatLog.git
 [submodule "lib/mbedtls"]
 	path = lib/mbedtls
-	url = https://github.com/cuberite/polarssl.git
```
bearbin, line 12: Don't change the submodule URLs, it breaks things! Just run a git pull in our vendored version if necessary.

tigerw:
Ok guys, here's the plan of action, because I really need this merged soon:

1. [CMake] Fix https
2. [CMake] Add conditional PCH based on version, reduce minimum version required
3. [Submodules] Update in our cuberite repos
4. [Submodules] Rebase, fix conflicts, add extra changes
5. [CI] Purge Appveyor cache so it builds

After it's merged another PR will come in to complete the SetFlags migration and address peterbell's target_include_directories query.

peterbell10:
I don't see the rush. No one else is touching CMake so it's unlikely to conflict. And we can't merge any C++17 stuff until @bearbin updates the build servers. 
That said, these submodule updates have nothing to do with C++17, right? If you rebase and fix the compile issues then we can merge this ASAP.

We should get these right the first time, not just merge in a broken state and add fixups later.

tigerw:
It's a personal challenge; with Mojang releasing 1.16 I want to move faster. Right now these PRs are blocking some other changes I want to make.

tigerw:
The submodule upgrade has a dependency on the CMake changes, which in turn depend on the CI changes.

peterbell10:
>The submodule upgrade has a dependency on the CMake changes, which in turn depend on the CI changes.

Sorry, but that's not what I meant when I said split it up. Each PR should be independent, including passing CI on its own.

Most of your cmake changes should compile now if you change `CMAKE_CXX_STANDARD` to 11. The submodule updates don't look like they require C++17 (especially the C libraries...), we can merge that soon. The only thing that needs to wait is the CI update and maybe a few final cmake tweaks.

tigerw:
My goodness do you drive a hard bargain

tigerw:
Sorry, it has to be everything at once. A small selection of issues encountered: too old gcc, mbedtls doesn't work, sqlitecpp removed their out-of-source sqlite option, jsoncpp doesn't work, fmt adds an unrecognised clang option, fmt deprecated printf. I'm not spending more time trying to get it to work individually when the end result is everything getting merged together.

peterbell10:
Okay, I'll work on update the submodules then.

src/OSSupport/File.cpp
```diff
@@ -692,11 +692,10 @@ AString cFile::GetExecutableExt(void)
...
 
 
-int cFile::vPrintf(const char * a_Fmt, fmt::printf_args a_ArgList)
+int cFile::vPrintf(const char * a_Format, fmt::printf_args a_ArgList)
 {
-	fmt::memory_buffer Buffer;
-	fmt::printf(Buffer, fmt::to_string_view(a_Fmt), a_ArgList);
-	return Write(Buffer.data(), Buffer.size());
+	AString buf = ::vPrintf(a_Format, a_ArgList);
+	return Write(buf.c_str(), buf.length());
```
peterbell10, line 698: Revert this.

src/StringUtils.cpp
```diff
@@ -52,35 +52,31 @@ static unsigned char HexToDec(char a_HexChar)
...
+AString & vPrintf(AString & a_String, const char * a_Format, fmt::printf_args a_ArgList)
 {
-	ASSERT(format != nullptr);
-	fmt::memory_buffer Buffer;
-	fmt::printf(Buffer, fmt::to_string_view(format), args);
-	str.assign(Buffer.data(), Buffer.size());
-	return str;
+	ASSERT(a_Format != nullptr);
+	a_String = fmt::vsprintf(a_Format, a_ArgList);
+	return a_String;
```
peterbell10, line 59: Revert this

src/StringUtils.cpp
```diff
@@ -52,35 +52,31 @@ static unsigned char HexToDec(char a_HexChar)
...
 
-AString & vAppendPrintf(AString & a_String, const char * format, fmt::printf_args args)
+AString & vAppendPrintf(AString & a_String, const char * a_Format, fmt::printf_args a_ArgList)
 {
-	ASSERT(format != nullptr);
-	fmt::memory_buffer Buffer;
-	fmt::printf(Buffer, fmt::to_string_view(format), args);
-	a_String.append(Buffer.data(), Buffer.size());
+	ASSERT(a_Format != nullptr);
+	a_String += vPrintf(a_Format, a_ArgList);
```
peterbell10, line 79: Revert.

src/Vector3.h
```diff
@@ -395,26 +395,29 @@ class Vector3
 
 
 
+
+// TODO: consolidate
```
peterbell10, line 399: What do you mean?

src/OSSupport/File.cpp
tigerw: Please explain your reasoning.

1. Is there any evidence to suggest that there is any performance improvement, which will offset the deviation from writing obvious, maintainable code?
2. I received a warning that printf into a buffer was deprecated in a newer fmt.

src/Vector3.h
tigerw: The namespace is not required in 17

src/OSSupport/File.cpp
peterbell10: >Is there any evidence to suggest that there is any performance improvement, which will offset the deviation from writing obvious, maintainable code?

It is trivially a performance improvement if you look at the implementation of `fmt::vsprintf`:
https://github.com/fmtlib/fmt/blob/d0623de510c64234dfccbe62dc3f77a1852b1d52/include/fmt/printf.h#L628-L630

It always writes into `fmt::memory_buffer` first and then copies into a new string. So, this code saves 1 memory allocation/deallocation and 1 string copy for every write to the file.

>I received a warning that printf into a buffer was deprecated in a newer fmt.

As in the linked code, it's been renamed to `vprintf`.

tigerw:
reeebased.

PR #4709, opened 2020-05-03
Added check if player is in spectator mode

Seadragon91:
If this is not checked, a player would be able to use a modified client and use this packet to always teleport to a player, with in-game modes not spectator.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1641,6 +1641,12 @@ void cClientHandle::HandleSlotSelected(Int16 a_SlotNum)
 
 void cClientHandle::HandleSpectate(const cUUID & a_PlayerUUID)
 {
+	if (!m_Player->IsGameModeSpectator())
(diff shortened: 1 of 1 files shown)


Seadragon91:
What is appveyor doing?
```
cmake -G "Visual Studio 14 2015 Win64" -DSELF_TEST=1 -DBUILD_TOOLS=1 ..
CMake Error: Error: generator : Visual Studio 14 2015 Win64
Does not match the generator used previously: Visual Studio 16 2019
```
How did the generator changed?

mathiascode:
The generator was bumped in a PR, which messed with the cache.

tigerw:
If this doesn't happen in normal gameplay you could probably m_Client->Kick with a suitably accusatory message ;)

Seadragon91:
Moved it to `cClientHandle::HandleSpectate` and added a kick.

PR #4712, opened 2020-05-03
Refactor minecart collision detection code.

bearbin:
- Use new GetBoundingBox function.
 - Handle descending and ascending rails.
 - Snap to descending rails.

More refactoring is due here, and I have plans to do so, but ideally this would be merged in as I've tested it to be functional and cleaned up as I work on the rest of the rail code.

### DIFF
--- a/src/Entities/Minecart.cpp
+++ b/src/Entities/Minecart.cpp
@@ -158,6 +158,16 @@ void cMinecart::HandlePhysics(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 			// Push cart upwards
 			AddPosY(1);
 		}
+		else
+		{
+			// When a minecart gets to a descending rail, it should go down.
+			chunk->GetBlockTypeMeta(relPos.addedY(-1), InsideType, InsideMeta);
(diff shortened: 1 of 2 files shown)


bearbin:
The behaviour of carts rolling down a hill into a block is non-vanilla, as the new code does not have [this longstanding bug in vanilla](https://bugs.mojang.com/browse/MC-8004). It's practically a feature by now, but since it's open in the bugtracker and dinnerbone hasn't closed it I think we can have our own behaviour :)

tigerw:
Any change to minecarts is a good change, happy to merge.

Testing notes:

The empty UNREACHABLE doesn't compile in Visual C++, maybe add a shouty message string to it.

When I'm facing positive X,
I click on height 71 and carts move not (<70 is fine, 70 is glitchy)
Stuck cart
I click on 73 and carts are super glitchy
Noclip cart

bearbin:
I don't know how I ended up with an empty unreachable, it should have had a message...

That bug is very odd, I shall have to try and reproduce it. Did it happen for any of the other cardinal directions?

PR #4713, opened 2020-05-04
Add tree generation for ExtemeHills and other bioms

### LINKED ISSUE #244: Add the Extreme Hills+ Biome
Contains spruce trees

mBornand:
1. Set the correct trees for extreme hills and variants may Fix #244
2. Allow Trees to be generated for values under 1 per Chunk ( Extreme Hills, plains, ...)

The tree ratio in the extreme Hills may be no accurate, also the number of trees in plains is very heuristic.


src/Generating/StructGen.cpp
```diff
@@ -50,12 +50,21 @@ void cStructGenTrees::GenFinish(cChunkDesc & a_ChunkDesc)
...
+			int NumTrees = GetNumTrees(BaseX, BaseZ, Dest->GetBiomeMap(), RandomNumTree);
 
 			sSetBlockVector OutsideLogs, OutsideOther;
 			for (int i = 0; i < NumTrees; i++)
 			{
-				GenerateSingleTree(BaseX, BaseZ, i, *Dest, OutsideLogs, OutsideOther);
+				GenerateSingleTree(BaseX, BaseZ, i, *Dest, OutsideLogs, OutsideOther, [](cNoise &, Vector3i) { return true; });
+			}
+			// generate a tree probability RandomNumTree / DividerForOnePerC
+			if (true)
```
bearbin, line 62: Is this actually supposed to have an effect - have another look at this code.

mBornand: Sorry I committed to fast

mBornand:
Now it is based on float the values for bioms with a tiny amount of trees less than 1 / Chunk may still need some corrections

src/Generating/StructGen.cpp
```diff
@@ -204,87 +221,88 @@ int cStructGenTrees::GetNumTrees(
...
-			case biMegaTaigaHills:       return 25;
-			case biExtremeHillsPlus:     return 3;
-			case biSavanna:              return 8;
-			case biSavannaPlateau:       return 12;
-			case biMesa:                 return 2;
-			case biMesaPlateauF:         return 8;
-			case biMesaPlateau:          return 8;
+			case biOcean:                return 2.0;
+			case biPlains:               return 0.03125;
+			case biDesert:               return .0;
```
madmaxoft, line 226: Please don't use the horrible "nothing point something" format. And anyway, what's wrong with a plain zero?

src/Generating/StructGen.h
```diff
@@ -16,7 +16,7 @@
 #include "ComposableGenerator.h"
 #include "../Noise/Noise.h"
 
-
+#include <functional>
```
madmaxoft, line 19: This doesn't seem to be used by the header, move it to cpp file, if it is needed at all.

src/Generating/StructGen.h
```diff
@@ -59,7 +60,10 @@ class cStructGenTrees :
 		sSetBlockVector & a_Overflow
 	);
 
-	int GetNumTrees(
+	/** Get the number of Tree to generate for chunk (a_ChunkX; a_ChunkZ)
+	if it returns zero, a_randomNum is set to a range of 0 to 1023 indicating the probability that a tree should spawn: a_RandomNum / cStructGenTrees::DividerForOnePerC
+	else a_RandomNum is 0 */
+	double GetNumTrees(
```
madmaxoft, line 66: This doxy-comment makes absolutely no sense. Please clarify it.

src/Generating/StructGen.h
mBornand: it is no more used i removed it

src/Generating/StructGen.cpp
mBornand: 0 doesn't work because of auto
.0 and 0.0 works, I have change it.

src/Generating/StructGen.h
mBornand: Sorry I forgot to change it in an preview commit

src/Generating/StructGen.cpp
```diff
@@ -204,87 +221,88 @@ int cStructGenTrees::GetNumTrees(
...
+			case biMesaPlateauM:         return 12.0;
 			// Non-biomes
 			case biInvalidBiome:
 			case biNumBiomes:
 			case biVariant:
 			case biNumVariantBiomes:
 			{
 				ASSERT(!"Invalid biome in cStructGenTrees::GetNumTrees");
-				return 0;
+				return .0;
```
bearbin, line 293: Missing a leading 0 here.

src/Generating/StructGen.cpp
```diff
@@ -204,87 +221,88 @@ int cStructGenTrees::GetNumTrees(
...
 			}
 		}
 		UNREACHABLE("Unsupported biome");
 	};
 
-	int NumTrees = 0;
+	double NumTrees = .0;
```
bearbin, line 299: and here

src/Generating/Trees.cpp
```diff
@@ -301,26 +307,34 @@ void GetTreeImageByBiome(Vector3i a_BlockPos, cNoise & a_Noise, int a_Seq, EMCSB
...
-		case biDesertM:
-		case biExtremeHillsM:
-		case biFlowerForest:
-		case biTaigaM:
-		case biIcePlainsSpikes:
-		case biJungleM:
-		case biJungleEdgeM:
-		case biColdTaigaM:
+		{
+			// TODO: implement trees
```
bearbin, line 313: Please could you specify what type of trees are expected in this comment? Just to act as a reminder when they are eventually added.

src/Generating/StructGen.cpp
mBornand: changed

src/Generating/Trees.cpp
mBornand: I have changed it.

tigerw:
Merging tomorrow.

src/Generating/StructGen.cpp
```diff
@@ -50,14 +50,35 @@ void cStructGenTrees::GenFinish(cChunkDesc & a_ChunkDesc)
...
+			if (NumTrees < 1)
 			{
-				GenerateSingleTree(BaseX, BaseZ, i, *Dest, OutsideLogs, OutsideOther);
+				Vector3i Pos;
+				Pos.x = (m_Noise.IntNoise3DInt(BaseX + BaseZ, BaseZ, 0) / 19) % cChunkDef::Width;
+				Pos.z = (m_Noise.IntNoise3DInt(BaseX - BaseZ, 0, BaseZ) / 19) % cChunkDef::Width;
+
+				Pos.y = Dest->GetHeight(Pos.x, Pos.z);
+
+				// If this test is not done with exact position for some reason the trees are cut at Chunk border
```
madmaxoft, line 64: I'm concerned about this comment, `for some reason` means something is misunderstood, and that's always bad.

madmaxoft: @mBornand Can you explain what you mean by `exact position`? Do you mean the equality test (`<=` instead of `<`) or that you use block coords (`BaseX * cChunkDef::Width + Pos.x` instead of, for example, just `Pos.x`)? Or something else?

mBornand: It is just important that block position is used and not the coordinates of the chunk because trees can also generate across chunks.
I edited the comment.

madmaxoft: Ok, so it's about the absolute coords instead of relative. That is pretty much evident, so the comment is rather useless, I'd remove it altogether. As well as the empty lines between the z and xz assignments of Pos, in both branches.

mBornand: If the tree is on chunk border it is generated for both chunk, so if the coordinates change the tree is cut.

madmaxoft: Exactly, so the comment is not really needed, it is self-evident that full absolute coords must be used for the test.

mBornand: It is removed

PR #4714, opened 2020-05-04
Remove cRankManager migration code

mathiascode:
Implemented 6 years ago in https://github.com/cuberite/cuberite/pull/1290
Pretty sure nobody has these old ini files anymore.


src/RankManager.cpp
```diff
@@ -2079,28 +1685,6 @@ bool cRankManager::IsRestrictionInGroup(const AString & a_Restriction, const ASt
...
 
 
-void cRankManager::NotifyNameUUID(const AString & a_PlayerName, const cUUID & a_UUID)
-{
-	ASSERT(m_IsInitialized);
-	cCSLock Lock(m_CS);
-
-	try
-	{
-		SQLite::Statement stmt(m_DB, "UPDATE PlayerRank SET PlayerName = ? WHERE PlayerUUID = ?");
```
peterbell10, line 2089: Are you sure this table isn't used by the webadmin or plugins?

mathiascode: The table is still used to store player ranks as usual. This function was necessary to fill in incomplete data when UUIDs were introduced. Since you've been forced to specify a player UUID and name when setting ranks for a couple of years now, it shouldn't be necessary anymore.

peterbell10: I don't understand. This function updates the player's name using an existing UUID as the key. I don't see how it could be related to migration since the ini data doesn't have a UUID to compare against.

mathiascode: It's weird, all the comments seem to indicate that it was supposed to do the opposite.
E.g. "The a_MojangAPI param is used when migrating from old ini files, to look up player UUIDs."
The original PR lists "Make cMojangAPI update the cached PlayerNames in the ranks".

mathiascode: I'll dig deeper into the original commits, to make sure I didn't misunderstand something.

peterbell10: It was added in https://github.com/cuberite/cuberite/commit/0c04bf962ed025789c1979c6d4fb122735b1a46b which has commit name "cMojangAPI updates cRankManager's playernames."

I think the intention is that player names associated with an account can be changed online so this keeps it up to date.

mathiascode: Ignore my previous message, this isn't part of the migration code. It was added after these comments.

mathiascode: > It was added in [0c04bf9](https://github.com/cuberite/cuberite/commit/0c04bf962ed025789c1979c6d4fb122735b1a46b) which has commit name "cMojangAPI updates cRankManager's playernames."
> 
> I think the intention is that player names associated with an account can be changed online so this keeps the the rank DB up to date.

Seems to be right. I'll keep the code in.

src/RankManager.cpp
```diff
@@ -421,21 +55,9 @@ void cRankManager::Initialize(cMojangAPI & a_MojangAPI)
...
-		cRankManagerIniMigrator Migrator(*this, a_MojangAPI);
-		if (Migrator.Migrate())
-		{
-			LOGINFO("Ranks migrated.");
-			// The default rank has been set by the migrator
-			return;
-		}
-
-		// Migration failed. Add some defaults
-		LOGINFO("Rank migration failed, creating default ranks...");
```
madmaxoft, line 438: This log message should have stayed, reworded to `Creating default ranks...`

PR #4716, opened 2020-05-04
Upgrade to C++17 [CI]

tigerw:
* Build on something with a semblance of modernity

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,78 +1,61 @@
 language: cpp
-sudo: false
+cache: ccache
 
 # Use Linux by default
 os: linux
+dist: bionic
+
+# CMake version out of date: update
+# TODO: add ARM64, PPC, SPARC builds when we find CMake for them
+addons:
+  snaps:
+  - name: cmake
+    confinement: classic
+    channel: latest
 
 matrix:
   include:
-    # gcc on arm64
-    - compiler: gcc
-      arch: arm64
-      addons:
-        apt:
-          packages:
-            - ccache
-      env: &Debug
-        - TRAVIS_CUBERITE_BUILD_TYPE=DEBUG
-    # AppleClang
+    # AppleClang Release
     # OSX workers are slower to start up.  Having these first in the build matrix makes travis faster overall.
     - os: osx
-      compiler: clang
+      osx_image: xcode11.3
       before_install:
         - HOMEBREW_NO_AUTO_UPDATE=1 brew install ccache
       env: &Release
         - TRAVIS_CUBERITE_BUILD_TYPE=RELEASE
+
+    # AppleClang Debug
     - os: osx
-      compiler: clang
-      env: *Debug
+      osx_image: xcode11.3
       before_install:
         - HOMEBREW_NO_AUTO_UPDATE=1 brew install ccache
-    # Default clang
+      env: &Debug
+        - TRAVIS_CUBERITE_BUILD_TYPE=DEBUG
+
+    # Default Clang Release
     - compiler: clang
+      before_install: &use-cmake
+        - export PATH=/snap/bin/:${PATH}
       env: *Release
+
+    # Default Clang Debug
     - compiler: clang
+      before_install: *use-cmake
       env: *Debug
-    # clang 3.5
-    - compiler: clang
-      dist: trusty
-      addons: &clang35
-        apt:
-          packages:
-            - clang++-3.5
-            - clang-3.5
-      before_install:
-        - CC=clang-3.5;CXX=clang++-3.5
-      env: *Release
-    - compiler: clang
-      dist: trusty
-      addons: *clang35
-      before_install:
-        - CC=clang-3.5;CXX=clang++-3.5
-      env: *Debug
-    # gcc 4.8
+
+    # Default GCC Release
     - compiler: gcc
-      addons: &gcc48
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-            - g++-4.8
-            - gcc-4.8
-      before_install:
-        - CC=gcc-4.8;CXX=g++-4.8
+      before_install: *use-cmake
       env: *Release
+
+    # Default GCC Debug
     - compiler: gcc
-      addons: *gcc48
-      before_install:
-        - CC=gcc-4.8;CXX=g++-4.8
+      before_install: *use-cmake
       env: *Debug
 
 before_script:
   - export PATH=$(echo "$PATH" | sed -e 's/:\/usr\/lib\/ccache//')
 script: ./travisbuild.sh
-cache: ccache
-
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,23 +32,22 @@ matrix:
       env: &Debug
         - TRAVIS_CUBERITE_BUILD_TYPE=DEBUG
 
-    # Default Clang Release
+    # Clang 7.0 - Release
     - compiler: clang
       before_install: &use-cmake
         - export PATH=/snap/bin/:${PATH}
       env: *Release
 
-    # Default Clang Debug
+    # Clang 7.0 - Debug
     - compiler: clang
       before_install: *use-cmake
       env: *Debug
 
-    # Default GCC Release
+    # GCC 7.4 - Release, CMake 3.12
     - compiler: gcc
-      before_install: *use-cmake
       env: *Release
 
-    # Default GCC Debug
+    # GCC 7.4 - Debug
     - compiler: gcc
       before_install: *use-cmake
       env: *Debug
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -6,16 +6,16 @@ To compile Cuberite from source, you need the following set of software:
  - CMake
  - Platform-specific make tool (Windows would be MSVC, Linux/macOS GNU make, etc.)
  - C compiler
- - Modern C++ compiler and linker
+ - Modern C++17 capable compiler and linker
 
 To contribute code, you also need a Git client.
 
 Windows
 -------
 
-We use Microsoft Visual Studio for Windows compilation. It is possible to use other toolchains, but we don't test against them and they aren't supported. Visual Studio 2013 Express for Desktop is being actively used for development.
+We use Microsoft Visual Studio for Windows compilation. It is possible to use other toolchains, but we don't test against them and they aren't supported. Visual Studio 2017 Community and above are being actively used for development.
 
-You can find download links for VS2013 Express [here][1].
+You can find download links for VS2019 Community [here][1].
 
 Next, you need to download and install [CMake][2]. You should download a full installation package, so that the installer will set everything up for you (especially the paths).
 
@@ -56,6 +56,8 @@ macOS
 
  - Install CMake from its [website][9] or homebrew: `brew install cmake`.
 
+Cuberite requires Xcode 11.3 or newer.
+
 ### Getting the Source
 
 ```
@@ -77,7 +79,13 @@ Install git, make, cmake and clang (or gcc), using your platform's package manag
 ```
 sudo apt-get install git make cmake clang
 ```
-Ensure that you have modern C++ compiler and linker (Clang 3.5+, GCC 4.8+, or VS 2013+).
+Ensure that you have modern C++ compiler and linker (Clang 7.0+, GCC 7.4+, or VS 2017+).
+Cuberite also requires CMake 3.13 or newer.  You might find that your distribution
+defaults are too out of date, in which case you may need to add a new `apt` source,
+or download directly from the projects' websites:
+
+ - [Clang][Clang-download]
+ - [CMake][2]
 
 ### Getting the Source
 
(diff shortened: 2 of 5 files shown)


tigerw:
The C++17 changes need to be merged at the same time.

peterbell10:
Why is the travis change more complicated than just update the compiler version?

tigerw:
'cos it's beautiful, that's why

(ok ok I'll minimise it)

peterbell10:
The line endings are messed up.

tigerw:
argh

well on the plus side it's easier to read!

bearbin:
Make it pass appveyor, then approved.

.travis.yml
```diff
@@ -1,84 +1,69 @@
...
+
+    # AppleClang Debug
+    - os: osx
+      osx_image: xcode11.3
+      before_install:
+        - HOMEBREW_NO_AUTO_UPDATE=1 brew install ccache
+      env: &Debug
+        - TRAVIS_CUBERITE_BUILD_TYPE=DEBUG
+
+    # Default Clang Release
```
peterbell10, line 35: We need to pick new minimum compiler versions and test those.

.travis.yml
```diff
@@ -1,84 +1,69 @@
...
+
+# Use Linux by default
+os: linux
+dist: bionic
+
+# CMake version out of date: update
+# TODO: add ARM64, PPC, SPARC builds when we find CMake for them
+addons:
+  snaps:
+  - name: cmake
```
peterbell10, line 12: Does `apt` not have a version new enough? If not then that suggests you're requiring too high a version.

.travis.yml
```diff
@@ -1,84 +1,69 @@
...
+    # Default Clang Release
+    - compiler: clang
+      cache: false # Fails on precompiled headers otherwise
+      before_install: &use-cmake
+        - export PATH=/snap/bin/:${PATH}
+      env: *Release
+
+    # Default Clang Debug
+    - compiler: clang
+      cache: false # Fails on precompiled headers otherwise
```
peterbell10, line 44: No it doesn't. We use clang with ccache right now.

.travis.yml
peterbell10: g++-8 and clang-7 seems to be the minimum, looking at this table:
https://en.cppreference.com/w/cpp/compiler_support

peterbell10: The travis workers seem to be using g++ 7.4 and clang 7.0 which seems alright. We should make sure those versions are pinned in CI and update the README.

tigerw: Travis never upgrades anything, but I can upgrade the readme.

.travis.yml
tigerw: lol, apt never has a version new enough, it's become a running joke in every C/C++ .travis.yml ever to add ubuntu-toolchain-ppa-test

besides I need it for pch/unity builds

.travis.yml
tigerw: Oh, the [*pain*](https://travis-ci.org/github/cuberite/cuberite/jobs/682415107#L1004)

PCH failing on Clang

All the C++17 PRs need to be merged together, so PCH is going to be enabled on all platforms.

peterbell10: Try it again with `travisbuild.sh`.  There were some clang-specific issues that required `CCACHE_CPP2` to be defined.

peterbell10: Also, ccache is a much bigger win than pch so I would disable pch instead of ccache.

.travis.yml
peterbell10: If it's not on apt it will be hard for people to contribute. I'm on a pretty new install and my distro only has cmake 3.13.

.travis.yml
tigerw: OK. I can add a Travis-specific flag if the shell script doesn't work, but in order to test I'll need everything merged, the see-makes are in another PR xD

.travis.yml
tigerw: This is a conundrum, on one hand it's a backwards compatible one-time setup, but on the other it's still a hurdle. Windows is fine, Ubuntu with snaps is fine, shall we add instructions to download from cmake.org for others?

tigerw: It might be as simple as `wget https://github.com/Kitware/CMake/releases/download/v3.17.2/cmake-3.17.2-Linux-x86_64.sh && sh cmake-3.17.2-Linux-x86_64.sh`.

bearbin: Linux distributions condition their users rather strongly to _not_ execute random shell scripts off the internet to install software (in contrast to windows' free-for-all);we really shouldn't require our users use an unstable version of their distribution to compile cuberite. Limiting ourselves to Debian stable/ubuntu LTS makes sense (as those are probably the most-installed 'stable' distros)

tigerw: I might be able to add a condition enabling PCH depending on an up-to-date CMake version.

...a free-for-all, with up to 35% faster build speeds ;)))

mathiascode:
This would be a good opportunity to update the COMPILING.md file, more specifically these sections: https://github.com/cuberite/cuberite/blob/master/COMPILING.md#windows
https://github.com/cuberite/cuberite/blob/master/COMPILING.md#linux-freebsd-etc

tigerw:
Merging tomorrow morning.

appveyor.yml
```diff
@@ -35,10 +35,12 @@ for:
       - job_name: Windows-x64-debug
 
   before_build:
+  - if exist %BUILD_DIR% rmdir /S /Q %BUILD_DIR%
```
madmaxoft, line 38: Why are you removing the builddir? Doesn't that just get rid of all the meticulously added caching?

madmaxoft:
You will NOT merge until there's an approving review.

appveyor.yml
tigerw: To satisfy the request for building on AppVeyor, it will be removed before merge.

tigerw:
Approval is subject to passing AppVeyor, see a couple of comments up.

tigerw:
By the way xoft what compiler version are you using?

tigerw:
All done. Any further objections?

tigerw:
> This would be a good opportunity to update the COMPILING.md file

Indeed, will do so after the bulk of C++17 is merged.

.travis.yml
peterbell10: I'll add that you can get unity builds and PCH with [`cotire`](https://github.com/sakra/cotire/) which works on old cmakes. I played around with it for cuberite a couple years ago but in the end it didn't seem worth it. The speedup wasn't that great, I don't know how you measured 35% but that wasn't my experience testing on travis.

Unity builds are also really bad for CI because they make broken code valid. If you're missing an include but happen to be unified with a file that does have that include, then it can compile anyway. It also brakes valid code because anonymous namespaces and static functions suddenly need to have unique names.

peterbell10:
This is blocked on the build server updates.

.travis.yml
peterbell10: Actually, looking at the `cotire` docs I remember the PCH was recompiled for each library separately. `cotire` + flattening the build structure might give the 35% you're seeing.

bearbin:
https://builds.cuberite.org has GCC 8 including for the RasPi builders since last night. Is there any other buildserver that needs to be updated?

peterbell10:
Oh, great! Just waiting for the cmake version discussion then.

tigerw:
Thanks! I'm not aware of any other servers.

Final call for objections!

tigerw:
Oh cmake version minimum is locked at the one in Debian Buster, with more features enabled when higher.

peterbell10:
If we don't require the latest CMake then we should have at least one CI run without the snap and you should restore the arm build on travis.

tigerw:
I do believe all the exotic architectures have cmake 3.0.

peterbell10:
The arm64 CI currently has cmake 3.5 on ubuntu xenial. I presume it's just whatever comes with the distro. It definitely isn't 3.0 since we require 3.2 atm.

tigerw:
Oops, that should've been 3.10. Nonetheless Debian has 3.13.

tigerw:
Just to confirm, we don't require the latest CMake but do require the one that comes in the Debian repository. Unfortunately's Ubuntu's is older, so the snap is required.

tigerw:
I will re-add exotic builds later.

bearbin:
Ubuntu has recently had a new ESR release, so travis should hopefully update their build images soon. I don't know what their timetable is for that though, if it's soon enough then we might be able to postpone this PR until it's done and not have to change anything back.

tigerw:
Focal Fossa! Nice name.

I don't mind changing stuff later, I just want this merged xD

peterbell10:
But the workers travis workers have cmake 3.12.4, not 3.10. That's plenty modern isn't it?

tigerw:
For one reason or another the builds running on LXC infrastructure are old, please see:

[System/360](https://travis-ci.org/github/cuberite/cuberite/jobs/681664575#L537)
[PowerPC](https://travis-ci.org/github/cuberite/cuberite/jobs/681664572#L532)
[Leg64](https://travis-ci.org/github/cuberite/cuberite/jobs/681664572#L532)

peterbell10:
Ah okay, weird.

COMPILING.md
```diff
@@ -79,7 +79,7 @@ Install git, make, cmake and clang (or gcc), using your platform's package manag
 ```
 sudo apt-get install git make cmake clang
 ```
-Ensure that you have modern C++ compiler and linker (Clang 7.0+, GCC 7.4+, or VS 2019+).
+Ensure that you have modern C++ compiler and linker (Clang 7.0+, GCC 7.4+, or VS 2015+).
 Cuberite also requires CMake 1.12 or newer.  You might find that your distribution
```
bearbin, line 83: Uh, 1.12? That sounds a bit out of date

mathiascode: That backward compatibility. :P

COMPILING.md
```diff
@@ -77,7 +79,13 @@ Install git, make, cmake and clang (or gcc), using your platform's package manag
 ```
 sudo apt-get install git make cmake clang
 ```
-Ensure that you have modern C++ compiler and linker (Clang 3.5+, GCC 4.8+, or VS 2013+).
+Ensure that you have modern C++ compiler and linker (Clang 7.0+, GCC 7.4+, or VS 2015+).
```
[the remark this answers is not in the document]
tigerw: omg, it's too late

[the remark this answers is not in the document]
tigerw: It should be 2017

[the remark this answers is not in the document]
peterbell10: If 2017 isn't in CI then we don't support it.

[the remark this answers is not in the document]
tigerw: i'm developing on 2017, I'd be the first to complain if it didn't work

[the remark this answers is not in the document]
peterbell10: Then why use 2019 in CI?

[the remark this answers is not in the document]
tigerw: 2019 has a couple of codegen improvements (e.g. reduced binary sizes for code with exceptions) and since the Windows builds are being served by the AppVeyor artefacts, we certainly wish to give our users the bestest code ;)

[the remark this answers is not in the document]
peterbell10: That's a good point, but it's a bit unfortunate. Could we use 2017 for the debug build?

tigerw:
HEAVENS ABOVE IT BUILDS

Merge?

mathiascode:
https://github.com/cuberite/cuberite/blob/45e55873a98bb357a91b79379533057e7b20ce05/GETTING-STARTED.md#useful-resources: VS2013 -> Visual Studio

https://github.com/cuberite/cuberite/blob/45e55873a98bb357a91b79379533057e7b20ce05/CONTRIBUTING.md#code-conventions: mentions C++11 and MSVC 2013

mathiascode:
Other than that, this PR looks okay to me.

COMPILING.md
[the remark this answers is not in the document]
tigerw: I don't know how to do it on AppVeyor. In any case we're targeting a language version, not a compiler. I tell cmake that 17 is required, if it passes I'm happy. If anything we need to build on the latest release, especially given that newer ones have more warnings that will fail the build.

tigerw:
Thanks, updated. Everything still passes, good to go peterbell?

appveyor.yml
```diff
@@ -35,11 +46,12 @@ for:
       - job_name: Windows-x64-debug
 
   before_build:
+  - if exist %BUILD_DIR% rd /q /s %BUILD_DIR% 2> NUL
```
tigerw, line 49: I don't agree with this change, AppVeyor is very slow without the cache.

Could you explain your reasoning for wanting to build on 2017?

peterbell10: >I don't agree with this change, AppVeyor is very slow without the cache.

Huh... I put it there for the same reason you did, the first build breaks when you change image.

>Could you explain your reasoning for wanting to build on 2017?

Even if VS 2017 supports all of C++17, supporting the compiler means supporting their non-standard quirks and bugs as well. VS 2013 was particularly bad for this and I hope they've improved since. But in general, we can't know for sure that a change will compile if it isn't in CI.

PR #4718, opened 2020-05-04
Update fmt to 6.2.0

peterbell10:
Isolates fmt changes from #4708 with some additions

- Fixed the `Vector3` formatter
- Use `format_args` like versions everywhere
- Use `fmt::memory_buffer` to avoid some string copying

Works with C++11, might need Appveyor update though.


src/OSSupport/File.h
```diff
@@ -163,8 +163,12 @@ class cFile
 	/** Returns the list of all items in the specified folder (files, folders, nix pipes, whatever's there). */
 	static AStringVector GetFolderContents(const AString & a_Folder);  // Exported in ManualBindings.cpp
 
-	int Printf(const char * a_Fmt, fmt::ArgList);
-	FMT_VARIADIC(int, Printf, const char *)
+	int vPrintf(const char * a_Fmt, fmt::printf_args);
+	template <typename... Args>
+	int Printf(const char * a_Fmt, const Args & ... a_Args)
+	{
+		return vPrintf(a_Fmt, fmt::make_printf_args(a_Args...));
```
tigerw, line 170: mhm, you might be surprised, this actually increased my binary size by 0.1

tigerw:
oi, no competition

tigerw:
But in all seriousness the other PR is already split up, I propose merging your fancy additions here into that.

peterbell10:
>oi, no competition

Just thought it made sense I take over for fmt since it's me that added it in the first place. This PR should also support the current compiler set so is really independent of the C++17 stuff.

Tools/GrownBiomeGenVisualiser/Globals.h
```diff
@@ -191,7 +191,23 @@ typedef unsigned char Byte;
...
+	#pragma clang diagnostic ignored "-Wc++2a-compat"
+#endif
+
 #include "fmt/format.h"
+#include "fmt/printf.h"
+
+#ifdef __clang__
+	#pragma clang diagnostic pop
+#endif
+
```
madmaxoft, line 210: Wouldn't it be easier to put all this into one separate include file? `#include "fmt.h"`

src/CommandOutput.h
```diff
@@ -16,8 +16,13 @@ class cCommandOutputCallback
 	virtual ~cCommandOutputCallback() {}  // Force a virtual destructor in subclasses
 
 	/** Syntax sugar function, calls Out() with Printf()-ed parameters; appends a newline" */
-	void Out(const char * a_Fmt, fmt::ArgList);
-	FMT_VARIADIC(void, Out, const char *)
+	void vOut(const char * a_Fmt, fmt::printf_args);
```
madmaxoft, line 19: The doxy-comment now doesn't fit

src/OSSupport/File.h
peterbell10: I tried replacing the main `Printf` override in `StringUtils.h` with a templated version and my results were:
clang - 0.1 MB increase
gcc- 0.01 MB increase

Not earth shattering but still worse than before. What compiler are you testing on?

tigerw: Oh not in StringUtils, just in this one, File.h, Visual C++.

Anyway consistency dictates this is fine.

tigerw:
yeah I'll rebase 4708 off this

PR #4719, opened 2020-05-05
Require semi-colon at end of function-like macros

peterbell10:
clang >= 8  has a "-Wextra-semi-stmt" warning for redundant semi-colons.
Main build was fixed in #4346, this finishes the job for `tests` and `Tools`.


tigerw:
the do while false construction is ludicrous

peterbell10:
The preferred option is to not use macros at all :)

Barring that, this is what we're stuck with.

tigerw:
woo! I got a smiley out of peterbell! You always sound so stoical in your replies :D

tigerw:
might as well merge

PR #4722, opened 2020-05-07
Remove old Android leftovers

mathiascode:
These are leftovers from the old Android app that once resided in the main repo.
Support for Squirrel was removed in 2012, see: https://forum.cuberite.org/thread-610.html


tigerw:
The system can handle writing to relative paths now? (or it always could?)

mathiascode:
It has worked fine since our current Android app was introduced at least. :P

PR #4717, opened 2020-05-04
Upgrade to C++17 [CMake]

tigerw:
+ Enable CMake-native pre-compiled headers
+ Enable unity builds

PR #4727, opened 2020-05-07
Update submodules

peterbell10:
Closes #4708

This updates jsoncpp, mbedtls, TCLAP and SQLiteCpp to their latest stable release. A few additional changes were needed:

- jsoncpp deprecated `Reader`, `FastWriter` and `StyledWriter` which I've replaced with some helper functions in `JsonUtils.cpp`
- SQLiteCpp changed how it builds with external sqlite libraries, now expecting them to be installed. The simplest path was to remove sqlite from cuberite's submodule and just use SQLiteCpp's internal version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,11 +266,8 @@ add_subdirectory(lib/fmt)
 # Add proper includes for LibEvent's event-config.h header:
 include_directories(SYSTEM ${LIBEVENT_INCLUDE_DIRS})
 
-# Prettify jsoncpp_lib_static name in VS solution explorer
-set_property(TARGET jsoncpp_lib_static PROPERTY PROJECT_LABEL "jsoncpp")
-# jsoncpp uses these for ccache support, clashing with our method
-set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "")
-set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "")
+# Prettify jsoncpp_lib name in VS solution explorer
+set_property(TARGET jsoncpp_lib PROPERTY PROJECT_LABEL "jsoncpp")
 
 if (WIN32)
 	add_subdirectory(lib/luaproxy/)
@@ -303,7 +300,7 @@ if (MSVC)
 		event_extra_static
 		expat
 		fmt
-		jsoncpp_lib_static
+		jsoncpp_lib
 		lua
 		luaexpat
 		mbedcrypto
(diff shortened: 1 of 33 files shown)


tigerw:
I know the repos in cuberite/ exist but I'm still not super keen on changing them to fix include directory inclusion, not least because on every subsequent update we have to keep applying the patch (and fixing the conflicts...). It seems like (referencing the other PR) that at least some of those target_include_directories are needed.

What happened with the `RULE_LAUNCH_COMPILE` property?

peterbell10:
I will submit a PR to SQLiteCpp fixing their CMake for next time.

>What happened with the RULE_LAUNCH_COMPILE property?

`jsoncpp` were using it to emulate `CXX_COMPILER_LAUNCHER` which is the preferred method to enable cmake. Using both resulted in the compiler being called as:
```
ccache ccache g++
```

tigerw:
suuuper weird, definitely warrants further investigation; I was only having ccache issues with clang on newest jsoncpp, maybe they fixed it in master.

In any case, merge away!

peterbell10:
Yes, it's fixed in jsoncpp 1.9. They overhauled their entire CMake in 1.9 but it requires too high a CMake version for current CI.

tigerw:
Please avoid duplicating work in the future, it doesn't send the right message.

peterbell10:
@tigerw if you had just split up #4708 like you were asked, then I wouldn't have had to take over. To be frank, #4717 is completely unreadable. I'm sure the changes are quite small yet the diff is 1.5K lines. How on earth am I supposed to review that?

tigerw:
The majority of the changes are procedural, but since it is a refactor of the build scripts, it is necessarily atomic and given that every folder has a CMakeLists, unfortunately large. Please ask if there are any queries as to its contents.

peterbell10:
>The majority of the changes are procedural, but since it is a refactor of the build scripts, it is necessarily atomic 

That's just nonsense. There is no requirement that the CMake be refactored into different files in the same commit. That's where most of the diff comes from. The actual meaningful changes are obscured in thousands of lines of irrelevance.

>Please ask if there are any queries as to its contents.

I should be able to read the diff and decide for myself, or only ask for minor clarification. What's the point of code review if I have to take your word the changes are correct?

tigerw:
The meaningful changes are the CMake refactor.

I am requesting that you ask about the rationale behind a change if there is uncertainty of its correctness.

peterbell10:
>The meaningful changes are the CMake refactor.

No they aren't. The meaningful changes are the changes that actually effect the build system. These include enabling IPO, unity builds, PCH, library structure, C++ version, etc. All of these actually change how the code is built.

Splitting the root CMake file into 8 new files doesn't actually change the build process at all.

>I am requesting that you ask about the rationale behind a change if there is uncertainty of its correctness.

I can't see what the changes actually are... Without a readable diff, it's far too much effort. For example, I would have had no idea that the sqlite changes even existed if it weren't for doing it myself. The changed value of `SQLITECPP_INTERNAL_SQLITE` is completely hidden in the diff; it just looks like copy-paste.

tigerw:
In that case, I will temporarily stash the `include()` refactoring.

peterbell10:
That would be much appreciated. Thank you.

PR #4728, opened 2020-05-08
Fix SetBlockMeta call in observer handler

peterbell10:
Noticed an MSVC warning about `NIBBLETYPE` being cast to `bool`. The last argument is `bool a_ShouldMarkDirty` so I don't think passing `a_Meta` makes sense.


tigerw:
Thanks! I wonder what happened here...

PR #4730, opened 2020-05-09
Fix clang-tidy errors

peterbell10:
Some style guide violations are causing the clang-tidy run to fail. I guess that CI entry hasn't been working properly for quite a while.


tigerw:
Does this explain the failures on docker clang ≥8?

Also xoft's gonna blow a fuse when he sees those underscores

peterbell10:
This check was always there, even before the docker changes. Looking at some old CI runs, it was just quietly ignoring the failures before.

peterbell10:
To clarify, clang-tidy was reporting the errors before but `run-clang-tidy.py` wasn't returning the error for some reason.

PR #4731, opened 2020-05-09
Enable C++17 in build

peterbell10:


tigerw:
We will need to come back some day to change those ints to unsigned ones

PR #4738, opened 2020-05-13
Enable some more clang-tidy linter checks

peterbell10:
This enables a small selection of extra `clang-tidy` checks. See documentation for meanings:
https://clang.llvm.org/extra/clang-tidy/checks/list.html

For each commit I enabled a check and ran `clang-tidy.sh` with `-fix` to apply the auto fix-its. I had to manually reviewed all the changes because the fix-its aren't perfect. In some cases they even introduced use-after-move bugs.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2078,12 +2078,10 @@ bool cClientHandle::CheckBlockInteractionsRate(void)
 	ASSERT(m_Player != nullptr);
 	ASSERT(m_Player->GetWorld() != nullptr);
 
-	if ((cRoot::Get()->GetServer()->ShouldLimitPlayerBlockChanges()) && (m_NumBlockChangeInteractionsThisTick > MAX_BLOCK_CHANGE_INTERACTIONS))
-	{
-		return false;
-	}
-
-	return true;
+	return !(
+		(cRoot::Get()->GetServer()->ShouldLimitPlayerBlockChanges()) &&
+		(m_NumBlockChangeInteractionsThisTick > MAX_BLOCK_CHANGE_INTERACTIONS)
+	);
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2078,10 +2078,12 @@ bool cClientHandle::CheckBlockInteractionsRate(void)
 	ASSERT(m_Player != nullptr);
 	ASSERT(m_Player->GetWorld() != nullptr);
 
-	return !(
-		(cRoot::Get()->GetServer()->ShouldLimitPlayerBlockChanges()) &&
-		(m_NumBlockChangeInteractionsThisTick > MAX_BLOCK_CHANGE_INTERACTIONS)
-	);
+	if (!cRoot::Get()->GetServer()->ShouldLimitPlayerBlockChanges())
+	{
+		return true;
+	}
+
+	return (m_NumBlockChangeInteractionsThisTick <= MAX_BLOCK_CHANGE_INTERACTIONS);
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -213,7 +213,10 @@ void cClientHandle::GenerateOfflineUUID(void)
 
 
 
-AString cClientHandle::FormatChatPrefix(bool ShouldAppendChatPrefixes, AString a_ChatPrefixS, AString m_Color1, AString m_Color2)
+AString cClientHandle::FormatChatPrefix(
+	bool ShouldAppendChatPrefixes, const AString & a_ChatPrefixS,
+	const AString & m_Color1, const AString & m_Color2
+)
 {
 	if (ShouldAppendChatPrefixes)
 	{
@@ -3310,7 +3313,7 @@ void cClientHandle::SocketClosed(void)
 void cClientHandle::SetSelf(cClientHandlePtr a_Self)
 {
 	ASSERT(m_Self == nullptr);
-	m_Self = a_Self;
+	m_Self = std::move(a_Self);
 }
 
 
--- a/src/Defines.cpp
+++ b/src/Defines.cpp
@@ -160,17 +160,10 @@ AString BlockFaceToString(eBlockFace a_BlockFace)
 
 bool IsValidBlock(int a_BlockType)
 {
-	if (
-		(
-		(a_BlockType > -1) &&
-		(a_BlockType <= E_BLOCK_MAX_TYPE_ID)
-		) ||
+	return(
+		((a_BlockType > -1) && (a_BlockType <= E_BLOCK_MAX_TYPE_ID)) ||
 		(a_BlockType == 255)  // the blocks 253-254 don't exist yet -> https://minecraft.gamepedia.com/Data_values#Block_IDs
-	)
-	{
-		return true;
-	}
-	return false;
+	);
 }
 
 
--- a/src/Defines.cpp
+++ b/src/Defines.cpp
@@ -160,7 +160,7 @@ AString BlockFaceToString(eBlockFace a_BlockFace)
 
 bool IsValidBlock(int a_BlockType)
 {
-	return(
+	return (
 		((a_BlockType > -1) && (a_BlockType <= E_BLOCK_MAX_TYPE_ID)) ||
 		(a_BlockType == 255)  // the blocks 253-254 don't exist yet -> https://minecraft.gamepedia.com/Data_values#Block_IDs
 	);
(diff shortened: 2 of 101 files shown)


src/Map.cpp
```diff
@@ -151,14 +151,12 @@ bool cMap::UpdatePixel(unsigned int a_X, unsigned int a_Z)
 				ChunkHeight /= 4;
 				while (((--Height) != -1) && IsBlockWater(a_Chunk.GetBlock(RelX, Height, RelZ)))
 				{
-					continue;
```
tigerw, line 154: It's better and more explicit without, I think an empty brace block just looks weird.

src/Mobs/EnderDragon.cpp
```diff
@@ -19,7 +19,6 @@ cEnderDragon::cEnderDragon(void) :
 
 void cEnderDragon::GetDrops(cItems & a_Drops, cEntity * a_Killer)
 {
-	return;
```
tigerw, line 22: maybe the same here

tigerw:
Skimmed it but looks great. What sort of move-after-use did clang introduce?

peterbell10:
Some stuff like:

```cpp
m_Var = std::move(a_Var);
ASSERT(a_Var != nullptr);
```

@tigerw you also seem to have messed up the clang-tidy CI. You seem to have pushed an old version of the docker to dockerhub.

tigerw:
There were some issues with newer clang flagging violations, maybe it's been fixed in the meantime.

peterbell10:
Which means we should fix the violations, not change the CI without telling anyone.

peterbell10:
I think we might actually need to upgrade to clang-tidy-9.0. It's getting some false-negatives. e.g. It wants to reject this:
```cpp
if (!CopyTableFrom(a_SrcLuaState, a_StackIdx, a_NumAllowedNestingLevels - 1))
{
	LOGWARNING("%s: Failed to copy table in pos %d.", __FUNCTION__, a_StackIdx);
	return false;
}
return true;

```

tigerw:
Sure, might as well go the newest clang that works... On newer apt clang installs you only need update-alternatives for cc and c++, not even clang-tidy etc. from what I recall, so that'll simplify the script.

peterbell10:
The update-alternatives was strictly necessary for `run-clang-tidy` when I updated it last week. I think it's better to play it safe.

src/Mobs/EnderDragon.cpp
peterbell10: Made it a comment.

tigerw:
Also is this new-fangled .clang_format file now in competition with CheckBasicStyle?

src/Mobs/PathFinder.cpp
```diff
@@ -278,9 +278,6 @@ bool cPathFinder::PathIsTooOld() const
 	{
 		acceptableDeviation = 1;
 	}
-	if ((m_FinalDestination - m_DeviationOrigin).SqrLength() > acceptableDeviation * acceptableDeviation)
-	{
-		return true;
-	}
-	return false;
+	const auto DeviationSqr = (m_FinalDestination - m_DeviationOrigin).SqrLength();
+	return (DeviationSqr > acceptableDeviation * acceptableDeviation);
```
tigerw, line 282: Just a minor thing, I think style calls for bracketing the multiplicands

tigerw:
Changes look eminently reasonable, ready to merge?

peterbell10:
We can't use clang-format for format checking because Cuberite's code style is sufficiently weird that it doesn't support it (as far as I can tell). However, `clang-tidy` read the `.clang-format` file when generating the fix-it code, which is why it's useful.

Also, this would be ready to merge if I could push circle-ci the docker image. My internet is being weird at the moment and `docker push` keeps timing out.

tigerw:
>  Cuberite's code style is sufficiently weird

Triggered

tigerw:
Is the image updated? I've access to a fast connection if needed.

peterbell10:
It's still failing for me but only docker is not working. I wonder if it's my docker that's misconfigured or something.

If you could try building and pushing the image, that  would be great.

tigerw:
The scripts on docker-build-image, circle-ci branch right?

tigerw:
Why are the update-alternatives priorities so high :P

tigerw:
Shouldn't we be updating alternatives for cc (default gcc) and c++ (default none), instead of clang and clang++? The clang-tidy cmake is currently detecting the C compiler as GCC but C++ as Clang.

tigerw:
```

420 warnings generated.
clang-tidy-9 -header-filter=/cuberite/src/\.?[^\.] -export-fixes /tmp/tmpzqtlr3il/tmpn5fas9e1.yaml -p=/root/cuberite/tidy-build -quiet /root/cuberite/src/Generating/ChunkDesc.cpp
420 warnings generated.
clang-tidy-9 -header-filter=/cuberite/src/\.?[^\.] -export-fixes /tmp/tmpzqtlr3il/tmp2vn5r60r.yaml -p=/root/cuberite/tidy-build -quiet /root/cuberite/src/Generating/ChunkGenerator.cpp
420 warnings generated.
clang-tidy-9 -header-filter=/cuberite/src/\.?[^\.] -export-fixes /tmp/tmpzqtlr3il/tmpivatztg4.yaml -p=/root/cuberite/tidy-build -quiet /root/cuberite/src/Generating/CompoGen.cpp
420 warnings generated.
clang-tidy-9 -header-filter=/cuberite/src/\.?[^\.] -export-fixes /tmp/tmpzqtlr3il/tmp_dv8dqys.yaml -p=/root/cuberite/tidy-build -quiet /root/cuberite/src/Generating/CompoGenBiomal.cpp
420 warnings generated.
clang-tidy-9 -header-filter=/cuberite/src/\.?[^\.] -export-fixes /tmp/tmpzqtlr3il/tmpmu3zgusg.yaml -p=/root/cuberite/tidy-build -quiet /root/cuberite/src/Generating/BioGen.cpp
420 warnings generated.
```

Elon Musk smoking

tigerw:
Pushed, testing locally with your branch it gave me an extra

```
Diagnostics:
- DiagnosticMessage:
    FileOffset: 4786 [Editor's note: this is line 163]
    FilePath: /root/cuberite/src/Defines.cpp
    Message: redundant boolean literal in conditional return statement
    Replacements: []
  DiagnosticName: readability-simplify-boolean-expr
```

Fixing that made everything pass.

One other thing, I think it might be better to do the (cd src/Bindings && lua Corei3Processor.lua) step after the CMake generate step, since otherwise Lua can't find the docs/ folder and throws a warning.

peterbell10:
>Shouldn't we be updating alternatives for cc (default gcc) and c++ (default none), instead of clang and clang++? The clang-tidy cmake is currently detecting the C compiler as GCC but C++ as Clang.

This shouldn't actually matter since we don't have any c-code in Cuberite itself. It would make sense though, so feel free to add it.

src/Defines.cpp
```diff
@@ -160,17 +160,10 @@ AString BlockFaceToString(eBlockFace a_BlockFace)
 
 bool IsValidBlock(int a_BlockType)
 {
-	if (
-		(
-		(a_BlockType > -1) &&
-		(a_BlockType <= E_BLOCK_MAX_TYPE_ID)
-		) ||
+	return(
```
tigerw, line 163: At the expense of pedantry could I petition you to add a space betwixt the return and bracket?

tigerw:
Ready to merge?

peterbell10:
If this CI run passes, then yes.

src/ClientHandle.cpp
```diff
@@ -2075,12 +2078,10 @@ bool cClientHandle::CheckBlockInteractionsRate(void)
...
-	if ((cRoot::Get()->GetServer()->ShouldLimitPlayerBlockChanges()) && (m_NumBlockChangeInteractionsThisTick > MAX_BLOCK_CHANGE_INTERACTIONS))
-	{
-		return false;
-	}
-
-	return true;
+	return !(
+		(cRoot::Get()->GetServer()->ShouldLimitPlayerBlockChanges()) &&
+		(m_NumBlockChangeInteractionsThisTick > MAX_BLOCK_CHANGE_INTERACTIONS)
+	);
```
madmaxoft, line 2084: Honestly, I find this a lot less readable than the original version.

tigerw: omg it's xoft hello again

madmaxoft: Yea I'm still here, but with less time to read the hundreds of notification emails I get.

PR #4739, opened 2020-05-13
Generate "LuaState_Typedefs.inc" before running clang-tidy

peterbell10:
Should fix the clang-tidy issues


tigerw:
Are the brackets for restoring the current directory?

peterbell10:
Yes, in bash parenthesis runs in a new environment so `(cd x && foo)` runs foo in directory x without changing the current working directory.

Issue #799, opened 2014-03-12
Implement streams-based logging


madmaxoft:
To avoid issues with printf()-based formatting, we should implement a logging framework that provides a stream-based access. The usage would then be something along the lines of:

``` cpp
size_t NumBytes = ...;
LOGD(__FUNCTION __ << "Allocating " << NumBytes << " bytes...");
...
LogWarning() << __FUNCTION__ << "Cannot alloc " << NumBytes << " bytes";
```

peterbell10:
Closing in favour of #3542

PR #4740, opened 2020-05-15
Minimise string copying in logging code

peterbell10:
Currently, the logging process involves:

* `fmt::format` or `Printf` formats log message then copies into an `AString`
* `LogSimple` then calls `Printf` again, formatting and copying into a new string
* Each `cLogger::cListener` then takes another copy of the string

For a typical logger setup, I expect a minimum of 6 copies and 4 memory allocations for each write to the log.

Now all formatting is done into a single `fmt::memory_buffer` and I never allocate a single `AString`. Since we always format into the same buffer, there should be no copying except to resize the storage. `fmt::memory_buffer` also has a 500 character in place buffer, so I expect this version not to allocate memory or copy at all most of the time.

I've also combined the `LoggerSimple.h` `vLOG` type functions into one for all log levels, matching the `cLogger` interface more closely. This also removes some of the boiler-plate in `Logger.cpp`.

### DIFF
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -33,6 +33,7 @@
 #include "../HTTP/UrlParser.h"
 #include "../Item.h"
 #include "../LineBlockTracer.h"
+#include "../Logger.h"
 #include "../Server.h"
 #include "../Root.h"
 #include "../StringCompression.h"
@@ -401,7 +402,7 @@ static int tolua_LOG(lua_State * tolua_S)
 	}
 
 	// If the param is a cCompositeChat, read the log level from it:
-	cLogger::eLogLevel LogLevel = cLogger::llRegular;
+	eLogLevel LogLevel = eLogLevel::Regular;
 	tolua_Error err;
 	if (tolua_isusertype(tolua_S, 1, "cCompositeChat", false, &err))
 	{
@@ -427,7 +428,7 @@ static int tolua_LOGINFO(lua_State * tolua_S)
 		return 0;
 	}
 
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), cLogger::llInfo);
+	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), eLogLevel::Info);
 	return 0;
 }
 
@@ -445,7 +446,7 @@ static int tolua_LOGWARN(lua_State * tolua_S)
 		return 0;
 	}
 
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), cLogger::llWarning);
+	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), eLogLevel::Warning);
 	return 0;
 }
 
@@ -463,7 +464,7 @@ static int tolua_LOGERROR(lua_State * tolua_S)
 		return 0;
 	}
 
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), cLogger::llError);
+	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), eLogLevel::Error);
 	return 0;
 }
 
--- a/src/Bindings/ManualBindings.cpp
+++ b/src/Bindings/ManualBindings.cpp
@@ -33,7 +33,6 @@
 #include "../HTTP/UrlParser.h"
 #include "../Item.h"
 #include "../LineBlockTracer.h"
-#include "../Logger.h"
 #include "../Server.h"
 #include "../Root.h"
 #include "../StringCompression.h"
@@ -368,23 +367,19 @@ static int tolua_StringSplitAndTrim(lua_State * tolua_S)
 /** Retrieves the log message from the first param on the Lua stack.
 Can take either a string or a cCompositeChat.
 */
-static AString GetLogMessage(lua_State * tolua_S)
+static void LogFromLuaStack(lua_State * tolua_S, eLogLevel a_LogLevel)
 {
 	tolua_Error err;
 	if (tolua_isusertype(tolua_S, 1, "cCompositeChat", false, &err))
 	{
-		return static_cast<cCompositeChat *>(tolua_tousertype(tolua_S, 1, nullptr))->ExtractText();
+		auto Msg = static_cast<cCompositeChat *>(tolua_tousertype(tolua_S, 1, nullptr))->ExtractText();
+		Logger::LogSimple(Msg, a_LogLevel);
+		return;
 	}
-	else
-	{
-		size_t len = 0;
-		const char * str = lua_tolstring(tolua_S, 1, &len);
-		if (str != nullptr)
-		{
-			return AString(str, len);
-		}
-	}
-	return "";
+
+	size_t len = 0;
+	const char * str = lua_tolstring(tolua_S, 1, &len);
+	Logger::LogSimple(std::string_view(str, len), a_LogLevel);
 }
 
 
@@ -406,11 +401,13 @@ static int tolua_LOG(lua_State * tolua_S)
 	tolua_Error err;
 	if (tolua_isusertype(tolua_S, 1, "cCompositeChat", false, &err))
 	{
-		LogLevel = cCompositeChat::MessageTypeToLogLevel(static_cast<cCompositeChat *>(tolua_tousertype(tolua_S, 1, nullptr))->GetMessageType());
+		LogLevel = cCompositeChat::MessageTypeToLogLevel(
+			static_cast<cCompositeChat *>(tolua_tousertype(tolua_S, 1, nullptr))->GetMessageType()
+		);
 	}
 
 	// Log the message:
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), LogLevel);
+	LogFromLuaStack(tolua_S, LogLevel);
 	return 0;
 }
 
@@ -428,7 +425,7 @@ static int tolua_LOGINFO(lua_State * tolua_S)
 		return 0;
 	}
 
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), eLogLevel::Info);
+	LogFromLuaStack(tolua_S, eLogLevel::Info);
 	return 0;
 }
 
@@ -446,7 +443,7 @@ static int tolua_LOGWARN(lua_State * tolua_S)
 		return 0;
 	}
 
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), eLogLevel::Warning);
+	LogFromLuaStack(tolua_S, eLogLevel::Warning);
 	return 0;
 }
 
@@ -464,7 +461,7 @@ static int tolua_LOGERROR(lua_State * tolua_S)
 		return 0;
 	}
 
-	cLogger::GetInstance().LogSimple(GetLogMessage(tolua_S), eLogLevel::Error);
+	LogFromLuaStack(tolua_S, eLogLevel::Error);
 	return 0;
 }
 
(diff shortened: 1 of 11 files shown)


src/LoggerListeners.cpp
```diff
@@ -46,40 +46,40 @@
 		}
 
 		#ifdef _DEBUG
-			virtual void Log(AString a_Message, cLogger::eLogLevel a_LogLevel) override
+			virtual void Log(std::string_view a_Message, eLogLevel a_LogLevel) override
 			{
 				Super::Log(a_Message, a_LogLevel);
 				// In a Windows Debug build, output the log to debug console as well:
-				OutputDebugStringA(a_Message.c_str());
+				OutputDebugStringA(AString(a_Message).c_str());
```
peterbell10, line 53: Note we can avoid a further string copy here if we guarantee that `a_Message` has a terminating null character. However, is only used in debug builds anyway so I'm not too bothered.

src/Logger.cpp
```diff
@@ -30,51 +62,57 @@ void cLogger::InitiateMultithreading()
...
+	std::string_view a_Format, eLogLevel a_LogLevel, fmt::printf_args a_ArgList
+)
 {
-	LogSimple(vPrintf(a_Format, a_ArgList), a_LogLevel);
+	fmt::memory_buffer Buffer;
+	WriteLogOpener(Buffer);
+	fmt::printf(Buffer, fmt::to_string_view(a_Format), a_ArgList);
+	fmt::format_to(Buffer, "\n");
+
+	LogLine(std::string_view(Buffer.data(), Buffer.size()), a_LogLevel);
```
peterbell10, line 99: This is the most important change. Writing everything by appending into `Buffer`, then using a `std::string_view` to avoid copies when calling the log listeners.

src/Bindings/ManualBindings.cpp
```diff
@@ -367,23 +367,19 @@ static int tolua_StringSplitAndTrim(lua_State * tolua_S)
...
-		if (str != nullptr)
-		{
-			return AString(str, len);
-		}
-	}
-	return "";
+
+	size_t len = 0;
+	const char * str = lua_tolstring(tolua_S, 1, &len);
+	Logger::LogSimple(std::string_view(str, len), a_LogLevel);
```
tigerw, line 382: Will lua_tolstring not return nullptr then? There was a check before.

src/Globals.h
```diff
@@ -200,46 +200,43 @@ template class SizeChecker<UInt8,  1>;
...
 	}
 
-	#define FLOGERROR   FLOG
-	#define FLOGWARNING FLOG
-	#define FLOGD       FLOG
-	#define FLOGINFO    FLOG
-	#define FLOGWARN    FLOG
+	inline void LogSimple(std::string_view a_Message, eLogLevel)
+	{
+		fmt::print("{}\n", a_Message);
```
tigerw, line 235: std::cout << a_Message << std::endl ?

src/Bindings/ManualBindings.cpp
peterbell10: If it is `nullptr` then `len` will be 0 so we are okay.

src/Globals.h
peterbell10: Why though? This works the same and saves an include.

src/Logger.cpp
```diff
@@ -30,51 +59,57 @@ void cLogger::InitiateMultithreading()
...
 
-void cLogger::vLogPrintf(const char * a_Format, eLogLevel a_LogLevel, fmt::printf_args a_ArgList)
+void cLogger::LogPrintf(
+	std::string_view a_Format, eLogLevel a_LogLevel, fmt::printf_args a_ArgList
+)
 {
-	LogSimple(vPrintf(a_Format, a_ArgList), a_LogLevel);
+	fmt::memory_buffer Buffer;
+	WriteLogOpener(Buffer);
+	fmt::printf(Buffer, fmt::to_string_view(a_Format), a_ArgList);
```
tigerw, line 93: Wasn't printf deprecated and renamed to vprintf?

Why is to_string_view being called on a_Format, itself a string_view?

peterbell10: >Wasn't printf deprecated and renamed to vprintf?

In a future unreleased version we don't use, yes.

And `fmt::string_view` is apparently it's own type.

src/LoggerListeners.cpp
```diff
@@ -15,13 +15,13 @@
...
-		virtual void SetLogColour(cLogger::eLogLevel a_LogLevel) = 0;
+		virtual void SetLogColour(eLogLevel a_LogLevel) = 0;
 		virtual void SetDefaultLogColour() = 0;
 
-		virtual void Log(AString a_Message, cLogger::eLogLevel a_LogLevel) override
+		virtual void Log(std::string_view a_Message, eLogLevel a_LogLevel) override
 		{
 			SetLogColour(a_LogLevel);
-			fputs(a_Message.c_str(), stdout);
+			fwrite(a_Message.data(), 1, a_Message.size(), stdout);
```
tigerw, line 24: oooh systems with 4 byte chars ain't gonna like that 1 ;)

sizeof(string_view::value_type) maybe

tigerw: Or cout << a_Message

peterbell10: `sizeof(char) == 1` by definition, you're thinking of `CHAR_BIT`.

src/Bindings/ManualBindings.cpp
tigerw: I'm not sure, cppreference says

>  The behavior is undefined if [s, s+count) is not a valid range (even though the constructor may not access any of the elements of this range)

better be safe I think

src/Logger.cpp
tigerw: Well does it work with vprintf?

Unfortunate... it's probably more reasonable in that future version which we should use ;)))

src/Globals.h
tigerw: There's an outstanding issue for migrating to streams for output.

src/LoggerListeners.cpp
tigerw: darn. well put sizeof(char) in there anyway to be more obvious what the argument is

src/LoggerListeners.cpp
```diff
@@ -161,33 +161,32 @@ class cVanillaCPPConsoleListener
...
-			case cLogger::llError:
+			case eLogLevel::Error:
 			{
-				LogLevelString = "Error";
+				fputs("Error: ", stdout);
 				break;
 			}
 		}
-		printf("%s: %s", LogLevelString.c_str(), a_Message.c_str());
+		fwrite(a_Message.data(), 1, a_Message.size(), stdout);
```
tigerw, line 189: and here too

tigerw:
yeah looks good. Any binary size reduction?

src/Bindings/ManualBindings.cpp
peterbell10: `[s, s + count)` becomes `[s, s)` when `count = 0`. which is an empty range, therefore nothing has to be valid.

src/Globals.h
peterbell10: Nah, there was an issue for modernising the logging. Which is why `fmt` was added. iostreams are not the way to go IMO.

tigerw: I'm referring to #799

what do you mean they're not the way to go, this is C++...

src/Bindings/ManualBindings.cpp
tigerw: Very well, confirmed by http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0903r0.pdf

src/Globals.h
tigerw: Nonetheless, it seems out of scope for this PR.

peterbell10:
gcc - Increased by 40 KB
clang - Decreased by 12KB

Overall, ¯\\\_(ツ)\_/¯

src/Globals.h
peterbell10: Well iostreams lack i18n support which xoft mentioned in #3542. They're also  notoriously slow since they have their own buffering on top of the system's standard buffers.

tigerw:
> They're also notoriously slow

I got a surprising result; does the below test code look valid?

    > c++ -O3 -flto -Wall main.cpp
    > c++ --version
    c++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

```cpp
#include <iostream>
#include <chrono>

int main()
{
        using namespace std::chrono;

#if 1
        auto t1 = high_resolution_clock::now();

        for (int i = 0; i != 50000; i++)
        {
                printf("%i\n", i);
        }

        fflush(stdout);

        auto t2 = high_resolution_clock::now();

        printf("stdio: %f\n", duration_cast<milliseconds>(t2 - t1).count() / 1000.f);

#else

        std::cout.sync_with_stdio(false);

        auto t1 = high_resolution_clock::now();

        for (int i = 0; i != 50000; i++)
        {
                std::cout << i << '\n';
        }

        std::cout.flush();

        auto t2 = high_resolution_clock::now();

        std::cout << "iostream: " << duration_cast<milliseconds>(t2 - t1).count() / 1000.f << '\n';
#endif
}
```
    stdio: 0.163000
------
```cpp
#include <iostream>
#include <chrono>

int main()
{
        using namespace std::chrono;

#if 0
        auto t1 = high_resolution_clock::now();

        for (int i = 0; i != 50000; i++)
        {
                printf("%i\n", i);
        }

        fflush(stdout);

        auto t2 = high_resolution_clock::now();

        printf("stdio: %f\n", duration_cast<milliseconds>(t2 - t1).count() / 1000.f);

#else

        std::cout.sync_with_stdio(false);

        auto t1 = high_resolution_clock::now();

        for (int i = 0; i != 50000; i++)
        {
                std::cout << i << '\n';
        }

        std::cout.flush();

        auto t2 = high_resolution_clock::now();

        std::cout << "iostream: " << duration_cast<milliseconds>(t2 - t1).count() / 1000.f << '\n';
#endif
}
```
    iostream: 0.042

peterbell10:
```cpp
printf("%i\n", i);
```
The console output is line-buffered by default, meaning that `\n` flushes the buffer automatically. This makes performance worse, but it's usually what you want, otherwise the log line won't be printed to the screen. 

If I remove the '\n', I find `printf` is indeed slightly faster than `cout`. I can also change the buffering mode using `setvbuf`, much like you disabled `sync_with_stdio` for `cout`.

tigerw:
Huh, TIL thanks!

I made a new test with your suggestion of removing the newline. The results are now in millisecs and I removed the sync_with_stdio since it didn't seem to have any effect.

```cpp
#include <iostream>
#include <chrono>

int main()
{
        using namespace std::chrono;

#if 1
        auto t1 = high_resolution_clock::now();

        for (int i = 0; i != 50000; i++)
        {
                printf("%i", i);
        }

        fflush(stdout);

        auto t2 = high_resolution_clock::now();

        printf("\nstdio: %f\n", duration_cast<microseconds>(t2 - t1).count() / 1000.f);

#else

        auto t1 = high_resolution_clock::now();

        for (int i = 0; i != 50000; i++)
        {
                std::cout << i;
        }

        std::cout.flush();

        auto t2 = high_resolution_clock::now();

        std::cout << "\niostream: " << duration_cast<microseconds>(t2 - t1).count() / 1000.f << '\n';
#endif
}
```
Mean of 3 runs:

    stdio: 6.672
    iostream: 5.271

This is somewhat surprising, but in any case
a) it's not reflective of reality (printing 50000 ints in a tight loop?) and
b) at the very least stream performance is very competitive on my implementation.

PR #4741, opened 2020-05-15
Travis: Fix ccache on debug builds

peterbell10:
Debug builds seem to be really slow right now. Wondering if `ccache` isn't working.

### DIFF
--- a/travisbuild.sh
+++ b/travisbuild.sh
@@ -11,6 +11,7 @@ if [ `which ccache` ]; then
 	export CCACHE_CPP2=true
 	CACHE_ARGS="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
 	echo "Using ccache installed at $(which ccache)"
+	ccache --max-size=3G
 	ccache -z # Zero statistics
 fi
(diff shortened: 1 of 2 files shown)


bearbin:
Ccache is working, it's just not getting any cache hits. Cache size appears to be near the set limit, but that shouldn't reduce the hits to near zero really. Maybe the cache is only getting propagated to the next build from the release build somehow?

bearbin:
It's fixed! Reasonably fast builds at last :)

peterbell10:
For some reason the debug builds had a 500 MB cache limit so the cache was being thrashed,

travisbuild.sh
```diff
@@ -11,16 +11,24 @@ if [ `which ccache` ]; then
...
 cmake . -DBUILD_TOOLS=1 -DSELF_TEST=1 ${CACHE_ARGS};
 
 echo "Building..."
-cmake --build . -- -j 2;
-ctest -j 2 -V;
+cmake --build . -j 2
+
+if [ `which ccache` ]; then
+		echo "Built with ccache, outputting cache stats..."
+		ccache -s
```
tigerw, line 25: waaaait what are these, TWO tabs???

PR #4743, opened 2020-05-16
More CMake cleanup

tigerw:


tigerw:
Everything is slightly better now.

.travis.yml
```diff
@@ -26,6 +23,15 @@ jobs:
       env: &Debug
         - TRAVIS_CUBERITE_BUILD_TYPE=Debug
 
+    - name: "Clang 6.0 - Debug"
+      arch: arm64
+      compiler: clang
+      before_install:
+        - wget --output-document=${HOME}/CMake http://anaconda.org/conda-forge/cmake/3.17.0/download/linux-aarch64/cmake-3.17.0-h28c56e5_0.tar.bz2
```
peterbell10, line 30: How about using a `CMAKE_URL` variable and use yaml references to make this `before_install` stage common to all but the osx workers?

.travis.yml
```diff
@@ -1,23 +1,20 @@
 language: cpp
-cache: ccache
+cache:
+  ccache: true
+  directories:
+    - CachedCMakeCache # Avoid rediscovery of types sizes
+    - Testing          # Store ctest cost data
```
peterbell10, line 6: What does this do?

CMake/AddDependencies.cmake
```diff
@@ -0,0 +1,83 @@
+# Set options for SQLiteCpp, disable all their tests and lints:
```
peterbell10, line 1: Is this file a straight copy-paste?

CMake/StampBuild.cmake
```diff
@@ -0,0 +1,60 @@
+# These env variables are used for configuring Travis CI builds.
```
peterbell10, line 1: Is this file copy-pasted?

.travis.yml
```diff
@@ -54,10 +60,16 @@ jobs:
       before_install: *use-cmake
       env: *Debug
 
-before_script:
-  - export PATH=$(echo "$PATH" | sed -e 's/:\/usr\/lib\/ccache//')
```
peterbell10, line 58: Where's this gone?

.travis.yml
```diff
@@ -26,6 +23,15 @@ jobs:
       env: &Debug
         - TRAVIS_CUBERITE_BUILD_TYPE=Debug
 
+    - name: "Clang 6.0 - Debug"
```
peterbell10, line 26: Is that compiler version accurate? If so, it's too old.

tigerw:
Hello again, thanks for the review, hope you had a good weekend!

appveyor.yml
```diff
@@ -5,24 +5,28 @@ image:
...
+  - job_name: Windows x86
     configuration: Release
-    BUILD_DIR: Release-x86
+    BUILD_DIR: Release x86
 
-  - job_name: Windows-x64
+  - job_name: Windows x64
     configuration: Release
-    BUILD_DIR: Release-x64
+    BUILD_DIR: Release x64
```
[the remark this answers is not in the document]
mathiascode: This will break download links for Windows builds, and I'd very much prefer if we didn't introduce more spaces into our URLs.

Server/Install/PackWindowsExecutables.cmd
```diff
@@ -3,19 +3,35 @@
...
 echo load APIDump  1>>cmds.txt
 echo api  1>>cmds.txt
 echo stop  1>>cmds.txt
 Cuberite --port 32767  0<cmds.txt
+
+:: API documentation
 7z a -tzip -y ManualAPI.zip -scsWIN "ManualAPI.lua"
 7z a -tzip -y AutoAPI.zip -scsWIN ".\BindingsDocs\*.lua" -x!_raw.lua
+
+:: Upload artifacts
```
mathiascode, line 32: Are you sure artifact names didn't change? https://github.com/cuberite/cuberite/pull/4440#issuecomment-569365234

appveyor.yml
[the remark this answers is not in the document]
tigerw: BUILD_DIR has a space to test for regressions in proper quoting in the seemake files.
I guess the job name space is unpopular, for another day then ;)

.travis.yml
```diff
@@ -54,10 +60,16 @@ jobs:
       before_install: *use-cmake
       env: *Debug
 
-before_script:
-  - export PATH=$(echo "$PATH" | sed -e 's/:\/usr\/lib\/ccache//')
+# Populate the CMake cache from Travis cache, if any
+# Unfortunately Travis can only cache directories, not single files
+before_script: mv CachedCMakeCache/CMakeCache.txt . || true
```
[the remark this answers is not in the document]
tigerw: It still represents an improvement, think of the cumulative time saved by people waiting on pulls to build! And you also must consider the reduced logspam, and the fact that the sizes of types as stored in the cache don't change for the majority of the builds that are run, so most of the configure process is indirectly contributing to global warming.

.travis.yml
```diff
@@ -1,23 +1,20 @@
...
   include:
     # OSX workers are slower to start up.  Having these first in the build matrix makes travis faster overall.
-    - name: "AppleClang - Release"
-      os: osx
-      osx_image: xcode11.3
-      before_install:
-        - HOMEBREW_NO_AUTO_UPDATE=1 brew install ccache
-      env: &Release
-        - TRAVIS_CUBERITE_BUILD_TYPE=Release
-
```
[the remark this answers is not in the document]
tigerw: It got "replaced" by arm, I thought since Travis is so generous giving us all this free compute time we should avoid running too many. Release Mac is a bit redundant when we have Debug Mac and Debease Linux.

CMake/SetCompilerFlags.cmake
```diff
@@ -0,0 +1,57 @@
+if(MSVC)
```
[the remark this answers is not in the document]
tigerw: ok

.travis.yml
```diff
@@ -26,6 +23,15 @@ jobs:
       env: &Debug
         - TRAVIS_CUBERITE_BUILD_TYPE=Debug
 
+    - name: "Clang 6.0 - Debug"
+      arch: arm64
```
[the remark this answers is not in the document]
tigerw: ok

tests/HTTP/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 include_directories(${CMAKE_SOURCE_DIR}/src/)
 include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/lib/libevent/include)
+include_directories(SYSTEM "${CMAKE_BINARY_DIR}/lib/libevent/include")
```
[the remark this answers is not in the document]
tigerw: me too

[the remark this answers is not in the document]
tigerw: but it can't find event-config.h otherwise since, ahem, not using master means cmake bugs ;)

tests/Network/CMakeLists.txt
```diff
@@ -55,10 +58,8 @@ add_library(Network
 )
 
 target_link_libraries(Network event_core event_extra fmt::fmt mbedtls)
-if (WIN32)
-	target_link_libraries(Network ws2_32.lib)
-else()
-	target_link_libraries(Network event_pthreads_static)
+if(NOT WIN32)
+	target_link_libraries(Network event_pthreads Threads::Threads)
```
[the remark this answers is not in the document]
tigerw: They provide a nice alias

.travis.yml
tigerw: excellent idea

.travis.yml
tigerw: Admittedly small benefit, where ctest can keep track of how long a test took ("the cost") and optimise their run order the next time around. So LuaStressTest (10 seconds) gets run first, instead of the order it was declared (close to last) and holding one thread hostage when everything else is done.

.travis.yml
tigerw: To the annals of history!

That is to say, it doesn't seem to be needed. Caching still works fine.

.travis.yml
tigerw: ageist
It works, doesn't it?

CMake/AddDependencies.cmake
```diff
@@ -0,0 +1,83 @@
...
+	# Check that the libraries are present:
+	if (NOT EXISTS "${PROJECT_SOURCE_DIR}/lib/${DEPENDENCY}/CMakeLists.txt")
+		message(FATAL_ERROR "${DEPENDENCY} is missing in folder lib/${DEPENDENCY}. Have you initialized and updated the submodules / downloaded the extra libraries?")
+	endif()
+
+	# Include all the libraries
+	# We use EXCLUDE_FROM_ALL so that only the explicit dependencies are compiled
+	# (mbedTLS also has test and example programs in their CMakeLists.txt, we don't want those):
+	add_subdirectory("lib/${DEPENDENCY}" EXCLUDE_FROM_ALL)
+endforeach()
```
tigerw, line 40: This was cleaned up, it was a bunch of copypasta before.

CMake/AddDependencies.cmake
```diff
@@ -0,0 +1,83 @@
...
+)
+
+# Link process information library:
+if (WIN32)
+	target_link_libraries(${CMAKE_PROJECT_NAME} PRIVATE Psapi.lib)
+endif()
+
+# Special case handling for libevent pthreads:
+if(NOT WIN32)
+	target_link_libraries(${CMAKE_PROJECT_NAME} PRIVATE event_pthreads)
```
tigerw, line 75: _static got purged

CMake/StampBuild.cmake
tigerw: I deleted setting an environment variable to set CMAKE_BUILD_TYPE in favour of setting the variable directly from the command line.

CMakeLists.txt
```diff
@@ -22,193 +23,29 @@ project(
...
+option(WHOLE_PROGRAM_OPTIMISATION "Enables link time optimisation for Release" ON)
 option(PRECOMPILE_HEADERS "Enable precompiled headers for faster builds" ON)
 option(SELF_TEST "Enables testing code to be built" OFF)
 option(UNITY_BUILDS "Enables source aggregation for faster builds" ON)
 
-# These env variables are used for configuring Travis CI builds.
-if(DEFINED ENV{TRAVIS_CUBERITE_BUILD_TYPE})
-	message("Setting build type to $ENV{TRAVIS_CUBERITE_BUILD_TYPE}")
-	set(CMAKE_BUILD_TYPE $ENV{TRAVIS_CUBERITE_BUILD_TYPE})
-endif()
```
tigerw, line 33: is no more

.travis.yml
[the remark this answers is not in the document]
tigerw: ARM workers are the fastest to start up, but the slowest to finish; the comment talks about the slowest to start up, what should happen here?

.travis.yml
tigerw: hm, because the anaconda archive structure is different from the cmake.org one requiring a different tar flag, and because of the no-unity exception, I ended up with more lines?

CMake/AddDependencies.cmake
tigerw: Change: https://github.com/cuberite/cuberite/pull/4743#discussion_r426781751

Server/Install/PackWindowsExecutables.cmd
tigerw: Does this https://ci.appveyor.com/project/Cuberite/cuberite/builds/32958239/job/w0bqesye5x4jjki0/artifacts look good?

.travis.yml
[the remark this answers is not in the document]
peterbell10: What matters is total time from beginning startup until completely finished. Startup time is the relevant component of time that made osx builds slower compared to normal clang builds.

.travis.yml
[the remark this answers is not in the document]
peterbell10: Correctness beats performance. `CMakeCache.txt` contains a hell of a lot more than just the size of fundamental types.

tests/HTTP/CMakeLists.txt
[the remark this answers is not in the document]
peterbell10: Well it worked before, so you must have changed something to break it. Maybe moving the library `add_subdirectories` after the tests and tools?

tests/Network/CMakeLists.txt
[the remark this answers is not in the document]
peterbell10: Better to explicitly link the `_static` version so we don't accidentally start linking against the shared version.

.travis.yml
peterbell10: Oh well

.travis.yml
peterbell10: The `/usr/lib/ccache` directory contains a bunch of "fake" compilers that are wrapped in `ccache`. We use `CMAKE_<LANG>_COMPILER_LAUNCHER` instead of these.

.travis.yml
peterbell10: Well our `COMPILING.md` file says it's not a supported version. Why not just used gcc which seems to have the same version as the amd64 workers.

tests/HTTP/CMakeLists.txt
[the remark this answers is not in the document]
peterbell10: Actually, I ran into this on master after updating my distro. Guess it's somehow cmake version/policy related.

.travis.yml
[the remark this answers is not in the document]
tigerw: I'm not convinced. The build from scratch ship has already sailed given the use of ccache, and CMake<i>Cache</i>.txt sounds like a prime candidate for cash given that it doesn't change in the vast majority of builds Travis does.

tests/Network/CMakeLists.txt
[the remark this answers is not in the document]
tigerw: Not at all, in fact every other libevent library we link to is against the (static, by configuration) alias, added via their `add_event_library` macro; indeed, it is event_pthreads_static that is for some reason the odd one out.

.travis.yml
tigerw: I mean, ccache seems to be working fine, so less code amirite?

.travis.yml
tigerw: From what I've seen clang tends to give more warnings, and we need them on these exotic arches xD

.travis.yml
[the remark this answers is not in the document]
peterbell10: >The build from scratch ship has already sailed given the use of ccache,

Not true, `ccache` preserves correctness since it only uses the cache when the pre-compiled source, compiler and compile flags are identical. It even detects things like the `__TIME__` macro and prevents caching.

>CMakeCache.txt sounds like a prime candidate for cash given that it doesn't change in the vast majority of builds Travis does.

It can change significantly if you edit the `CMakeLists.txt` files or a submodule updates their cmake. The only way to trust CI in these scenarios would be to clear the cache.

tests/Network/CMakeLists.txt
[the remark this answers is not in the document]
peterbell10: Okay then.

.travis.yml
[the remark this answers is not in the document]
tigerw: I'll defer to your judgement, I think I can see that if say, someone updates a CACHE'd variable set() but doesn't supply FORCE the new value won't be seen by CI.

tigerw:
Seems like travis has a broken ARM.

Server/Install/PackWindowsExecutables.cmd
mathiascode: Looks fine.

SetFlags.cmake
```diff
@@ -233,87 +81,81 @@ macro(enable_profile)
...
+	# Make build use multiple threads under MSVC:
+	add_compile_options(/MP)
+
+	# Make build use Unicode:
+	add_compile_definitions(UNICODE _UNICODE)
+
+	# TODO: level 4, warnings as errors
+else()
+	target_compile_options(
+		${CMAKE_PROJECT_NAME} PRIVATE
```
peterbell10, line 115: Instead of assuming the target name, it would be better if you changed the signature to
```cmake
macro(set_exe_flags target)
```

And likewise for the other macros I think. Better to be explicit about what targets we are effecting.

SetFlags.cmake
```diff
@@ -233,87 +81,81 @@ macro(enable_profile)
...
-#				# flags that are not present in 3.0
-#				set_source_files_properties(${FILENAME} PROPERTIES COMPILE_FLAGS "-Wno-error=covered-switch-default ")
-#				set_source_files_properties(${FILENAME} PROPERTIES COMPILE_FLAGS "-Wno-implicit-fallthrough -Wno-error=extra-semi")
-#				set_source_files_properties(${FILENAME} PROPERTIES COMPILE_FLAGS "-Wno-error=missing-variable-declarations")
-#			endif()
-#		endforeach()
-#	endif()
+	if(CMAKE_CXX_COMPILE_ID STREQUAL "Clang")
+		target_compile_options(
+			${CMAKE_PROJECT_NAME}Cuberite PRIVATE
```
peterbell10, line 138: Typo

tests/HTTP/CMakeLists.txt
peterbell10, line 3:
```suggestion
include_directories(SYSTEM "${LIBEVENT_INCLUDE_DIRS}")
```

tests/Network/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
 include_directories(${CMAKE_SOURCE_DIR}/src/)
 include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/lib/libevent/include)
+include_directories(SYSTEM "${CMAKE_BINARY_DIR}/lib/libevent/include")
```
peterbell10, line 3:
```suggestion
include_directories(SYSTEM "${LIBEVENT_INCLUDE_DIRS}")
```

CMake/AddDependencies.cmake
```diff
@@ -0,0 +1,83 @@
...
+	# We use EXCLUDE_FROM_ALL so that only the explicit dependencies are compiled
+	# (mbedTLS also has test and example programs in their CMakeLists.txt, we don't want those):
+	add_subdirectory("lib/${DEPENDENCY}" EXCLUDE_FROM_ALL)
+endforeach()
+
+# Add required includes:
+target_include_directories(
+	${CMAKE_PROJECT_NAME} SYSTEM PRIVATE
+	"${CMAKE_CURRENT_BINARY_DIR}/lib/libevent/include" # TODO: remove when updating libevent
+	lib/libevent/include
```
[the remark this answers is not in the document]
peterbell10: This was what was used prior to #4717

tigerw:
Merging?

CMake/AddDependencies.cmake
```diff
@@ -0,0 +1,85 @@
+function(build_dependencies TARGET)
```
peterbell10, line 1: I don't think the `add_subdirectories` call makes sense as a function of the target. You couldn't call it for multiple targets for example. Maybe split this into two functions.

SetFlags.cmake
```diff
@@ -233,87 +81,76 @@ macro(enable_profile)
...
-			if ("${CLANG_VERSION}" VERSION_GREATER 3.7)
-				check_cxx_compiler_flag(-Wno-double-promotion HAS_NO_DOUBLE_PROMOTION)
-				if (HAS_NO_DOUBLE_PROMOTION)
-					add_flags_cxx("-Wno-double-promotion")
-				endif()
-			endif()
-			add_flags_cxx("-Wno-error=unused-command-line-argument")
-			add_flags_cxx("-Wno-documentation-unknown-command")
-		endif()
+function(enable_warnings TARGET)
```
peterbell10, line 114: `set_exe_flags` made much more sense. `-fsigned-char` is definitely not a warning.

Tools/AnvilStats/CMakeLists.txt
```diff
@@ -23,8 +19,6 @@ endfunction()
 
 add_subdirectory(../../lib/zlib ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_FILES_DIRECTORY}/lib/zlib)
 
-set_exe_flags()
```
peterbell10, line 26: :-1:

Tools/GrownBiomeGenVisualiser/CMakeLists.txt
```diff
@@ -25,10 +19,6 @@ function(flatten_files arg1)
 endfunction()
 
 
-# Include the libraries:
-
-set_exe_flags()
```
peterbell10, line 30: :-1:

Tools/MCADefrag/CMakeLists.txt
```diff
@@ -28,8 +23,6 @@ endfunction()
 
 add_subdirectory(../../lib/zlib ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_FILES_DIRECTORY}/lib/zlib)
 
-set_exe_flags()
```
peterbell10, line 31: :-1:

Tools/NoiseSpeedTest/CMakeLists.txt
```diff
@@ -2,16 +2,11 @@ cmake_minimum_required(VERSION 3.0.2)
...
 include(../../SetFlags.cmake)
-set_flags()
-set_lib_flags()
-enable_profile()
 
 # Set include paths to the used libraries:
 include_directories(SYSTEM "../../lib")
 include_directories("../../src")
 
-set_exe_flags()
```
peterbell10, line 13: :-1:

Tools/ProtoProxy/CMakeLists.txt
```diff
@@ -23,8 +20,6 @@ endfunction()
 
 add_subdirectory(../../lib/zlib ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_FILES_DIRECTORY}/lib/zlib)
 
-set_exe_flags()
```
peterbell10, line 26: :-1:

appveyor.yml
```diff
@@ -5,24 +5,28 @@ image:
...
+
 # Set up environment variables for build info
 environment:
   CUBERITE_BUILD_SERIES_NAME: AppVeyor
   CUBERITE_BUILD_ID: "%APPVEYOR_BUILD_NUMBER%"
   CUBERITE_BUILD_DATETIME: "%APPVEYOR_REPO_COMMIT_TIMESTAMP%"
 
   matrix:
-  - job_name: Windows-x64-debug
+  - job_name: Windows-x64-Debug
```
peterbell10, line 19: @mathiascode do we distribute the debug build?

tests/HTTP/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 include_directories(${CMAKE_SOURCE_DIR}/src/)
 include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/lib/libevent/include)
+include_directories(SYSTEM "${LIBEVENT_INCLUDE_DIRS}")
```
peterbell10, line 3:
```suggestion
include_directories(SYSTEM "${LIBEVENT_INCLUDE_DIRS}")
```

tests/Network/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
 include_directories(${CMAKE_SOURCE_DIR}/src/)
 include_directories(SYSTEM ${CMAKE_SOURCE_DIR}/lib/libevent/include)
+include_directories(SYSTEM "${LIBEVENT_INCLUDE_DIRS}")
```
peterbell10, line 3:
```suggestion
include_directories(SYSTEM "${LIBEVENT_INCLUDE_DIRS}")
```

.travis.yml
peterbell10: You haven't done anything about this yet.

.travis.yml
peterbell10: Still needs to go back.

SetFlags.cmake
```diff
@@ -20,156 +18,6 @@ macro(add_flags_cxx FLAGS)
...
-macro(get_clang_version)
-	execute_process(
-		COMMAND "${CMAKE_CXX_COMPILER}" "--version"
-		OUTPUT_VARIABLE CLANG_VERSION_OUTPUT)
-	string(REGEX MATCH "version ([0-9]+\\.[0-9]+)" x ${CLANG_VERSION_OUTPUT})
-	set(CLANG_VERSION ${CMAKE_MATCH_1})
-endmacro()
-
-
-macro(set_flags)
```
peterbell10, line 34: You broke the diff by reordering the functions.

SetFlags.cmake
```diff
@@ -233,87 +81,76 @@ macro(enable_profile)
...
-#			endif()
-#		endforeach()
-#	endif()
+		# TODO: actually fix the warnings instead of disabling them
+		# or at least disable on a file-level basis:
+		-Wno-unused-parameter -Wno-missing-noreturn -Wno-padded -Wno-implicit-fallthrough
+		-Wno-double-promotion
+
+		# This is a pretty useless warning, we've already got -Wswitch which is what we need:
+		-Wno-switch-enum
```
peterbell10, line 139: These being `-Wno` was somewhat justified for clang, since we want to be `-Werror` clean. Skipping it for gcc as well makes no sense though.

appveyor.yml
mathiascode: Not officially.

peterbell10: I guess changing this is okay then.

.travis.yml
tigerw: What is your suggestion?

.travis.yml
tigerw: Could you remind me what it does though?

peterbell10: It means cmake will detect the real compiler, not the `ccache` wrapped compiler. So, for example if we want to use a specific compiler version in CI then we aren't relying on there being a `ccache` wrapped version.

.travis.yml
peterbell10: >Why not just used gcc which seems to have the same version as the amd64 workers.

tigerw: In response I suggested that we might need the warnings Clang provides, especially when it highlights some assumption we make that might not hold true on ARM.

Is the compiler version a blocker for you? I think arm server builds are more of a "nice to know it works" instead of "we support this" so building with Clang 6 isn't a problem.

tigerw:
I have `&your_comments` pending your thoughts on Clang on ARM.

tigerw:
Merging tomorrow?

tigerw:
Any final thoughts?

CMakeLists.txt
```diff
@@ -104,114 +49,19 @@ if(WHOLE_PROGRAM_OPTIMISATION)
...
-set(SQLITE_ENABLE_COLUMN_METADATA OFF CACHE BOOL "" FORCE)
-
-# Set options for LibEvent, disable all their tests and benchmarks:
-set(EVENT__DISABLE_OPENSSL   YES CACHE BOOL   "Disable OpenSSL in LibEvent"       FORCE)
-set(EVENT__DISABLE_BENCHMARK YES CACHE BOOL   "Disable LibEvent benchmarks"       FORCE)
-set(EVENT__DISABLE_TESTS     YES CACHE BOOL   "Disable LibEvent tests"            FORCE)
-set(EVENT__DISABLE_REGRESS   YES CACHE BOOL   "Disable LibEvent regression tests" FORCE)
-set(EVENT__DISABLE_SAMPLES   YES CACHE BOOL   "Disable LibEvent samples"          FORCE)
-set(EVENT__LIBRARY_TYPE "STATIC" CACHE STRING "Use static LibEvent libraries"     FORCE)
+# TODO: set_build_stamp()
```
peterbell10, line 55: Something unfinished?

tigerw: I'm planning to, but I decided against making this diff any larger, you'd kill me

peterbell10: If this is missing functionality then can't you just leave the original code?

tigerw: Sorry sorry, StampBuild.cmake is included and adds the build id as before.

But it just executes the file right now. The todo is for making it a function as you suggested for the others. Then we can call set_build_stamp(Cuberite)

peterbell10: I see. Alright then.

tigerw:
Thank you! :D

PR #4745, opened 2020-05-21
libevent: Enable IOCP backend on windows

peterbell10:
I don't have a windows machine to test on, but this should make windows builds use overlapped IO instead of `select()`. @tigerw

### DIFF
--- a/src/OSSupport/NetworkSingleton.cpp
+++ b/src/OSSupport/NetworkSingleton.cpp
@@ -75,12 +75,15 @@ void cNetworkSingleton::Initialise(void)
 	#endif
 
 	// Create the main event_base:
-	m_EventBase = event_base_new();
+	event_config * config = event_config_new();
+	event_config_set_flag(config, EVENT_BASE_FLAG_STARTUP_IOCP);
+	m_EventBase = event_base_new_with_config(config);
 	if (m_EventBase == nullptr)
 	{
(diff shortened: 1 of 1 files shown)


tigerw:
hold on a mo

mo = day or two whilst I fix mah computer

madmaxoft:
Is that even needed? `select()` on Windows is not the retarded BSD version with socket-number-based arrays, but a quite usable "re-understanding" of the API (true sets of sockets) that works pretty well.

peterbell10:
I wouldn't claim to be an expert here but from what I've read, the argument is that each `select` call requires linear time to inspect the `fd_set`s. Whereas,  IOCP just pops the next event from a queue. So IOCP should be able to handle many more concurrent connections with less CPU usage.

Whether this will make a practical difference to us, I don't know. But from our perspective, the `libevent` code provides the same API either way so the usability of the `select()` API doesn't really matter to us.

madmaxoft:
I know, my only point was along the lines of "why change something that works for us", but then, why not :) Go ahead.

tigerw:
Hold on, I'm almost there

tigerw:
Blocked on select

I'm still seeing it blocked on select, maybe libevent support for iocp is incomplete?

peterbell10:
Yes, it still uses `select` to back things like non-buffered sockets (which we don't use). It should have started up more threads for IOCP.

tigerw:
well it can't hurt

PR #4751, opened 2020-05-25
Fire sim: Handle fuel block being destroyed

### LINKED ISSUE #4748: Fire in mid-air
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 9e8598fb1ca359143600d6bb2e8b317126a86bcc

### Expected behavior
Fire should go away when I break netherrack with fire on it.

### Actual behavior
Fire stays, now in mid-air

light fire on netherrack, break netherrack

peterbell10:
Fixes #4748

When the fire sim is woken up on an existing fire, I now check the neighbouring blocks to see if the fire has lost fuel by calling `GetBurnStepTime` and comparing against the stored burn time.

### DIFF
--- a/src/Simulator/FireSimulator.cpp
+++ b/src/Simulator/FireSimulator.cpp
@@ -137,6 +137,16 @@ void cFireSimulator::SimulateChunk(std::chrono::milliseconds a_Dt, int a_ChunkX,
 
 		// FIRE_FLOG("FS: Fire at {0} is stepping", absPos);
 
+		// TODO: Add some randomness into this
+		const auto BurnStep = GetBurnStepTime(a_Chunk, relPos);
+		if (BurnStep == 0)
+		{
+			// Fire has no fuel or ground block, extinguish flame
+			a_Chunk->SetBlock(relPos, E_BLOCK_AIR, 0);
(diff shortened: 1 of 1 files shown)


tigerw:
Am I correct in saying that this fixes the problem where GetBurnStepTime handles fires whose base was destroyed, but this function is only called every 10 seconds for netherrack?

peterbell10:
That's not right. `GetBurnStepTime` always gave the correct result, however you'r right that it was previously only called every 10 seconds for netherrack. The fix here is updaing `cFireSimulator::AddBlock` so that it checks for fires that have lost their fuel source. `AddBlock` is called when neighbour blocks are updated, so the fire will be extinguished in the same tick.

The change in `GetBurnStepTime` is so that it doesn't modify the world when I call it from `AddBlock`. Considering that `SetBlock` itself calls `AddBlock`, I thought it seemed like a bad idea.

tigerw:
Right, but netherrack-fires still burn out when the base is removed, only 10 seconds later when GetBurnStepTime is next called; this PR makes it so that it goes out the next tick instead?

peterbell10:
That's right, it would still burn out eventually.

PR #4755, opened 2020-05-31
Configurable LTO

tigerw:


travisbuild.sh
```diff
@@ -20,12 +20,13 @@ if [ `which ccache` ]; then
 	ccache -z # Zero statistics
 fi
 
-# Work around a Clang + ccache issue with failing
-# builds by disabling precompiled headers
+# Work around a Clang + ccache issue with failing builds by disabling
+# precompiled headers. Turn off LTO for faster build speeds
```
bearbin, line 24: Is disabling LTO required for the build to pass, or does it just improve build speeds - why did the builds earlier fail and now pass?

tigerw: The combination of CMake 3.13, cmake_minimum_required(3.13), and Travis breaks the 3.13 GCC job. Seems like an issue fixed in newer CMake (seems like it stops adding an extra -rdynamic flag). So disabling LTO is like electric cars, faster and greener...

bearbin: In that case I think we should add a check explicitly disabling LTO if CMake is version 3.13, since it's broken somehow and there's no point working out a proper fix.

tigerw: Heavens above.

Testing locally, with exactly CMake 3.13.0.
GrownBiomeGenVisualiser/CMakeLists.txt has cmake_minimum_required 3.0.2.
Configure and compile: works.

Comment out cmake_minimum_required.
Compile: fails.

_Could it be?_

Nuke build folder. Still on CMake 3.13, minimum still commented out.
Configure and build: works.

***A caching issue?***

Maybe Travis ccache stores a flag wrong or something?

peterbell10: The fact that recreating the build folder fixed it only proves it wasn't a `ccache` issue... You need to recreate the `CMakeCache.txt` for some cmake changes to fully take effect.

tigerw: But Travis does recreate CMakeCache, right? You made extra sure of that 😉

PR #4756, opened 2020-06-01
Remove Schematic Chunk height limitation

KrystilizeNevaDies:


madmaxoft:
What would this be for? A schematic with Y > 255 is still pretty useless. Or do you have a use-case?

KrystilizeNevaDies:
A use case is if you had a schematic with a Y that is higher then 255 and then rotated it in world edit before pasting it, there's no tools I can't find that can rotate schematics so this seems like the best option

Issue #4672, opened 2020-04-18
Merge item handlers that only place a block with zero meta.


madmaxoft:
There are multiple `cItemHandler` descendants that only provide overrides for `IsPlaceable()` (returning `true`) and `GetPlacementBlockTypeMeta`, always setting a specific block type and zero meta. These should be merged into a single unified `cSimplePlaceableItemHandler` that is either a template on the block type, or has it as the constructor param.

PR #4766, opened 2020-06-24
Merged trivial item handlers into cSimplePlaceableItemHandler

idomori:
As requested in #4672 , 6 cItemHandlers (cItemBrewingStandHandler, cItemCakeHandler, cItemCauldronHandler, cItemFlowerPotHandler, cItemStringHandler, cItemSugarcaneHandler) are now merged into a cSimplePlaceableItemHandler.

A second parameter that takes in a BLOCKTYPE is added to cSimplePlaceableItemHandler's constructor.

The header files associated with merged handlers are commented out in CMakeList.txt


madmaxoft:
@tigerw We do have a convention of having 2 people review each PR before merging :P

Bond-009:
@madmaxoft You should set the repo up to only allow reviews when there are 2 approving reviews

tigerw:
With the breakneck pace Mojang is moving at I'm concerned that merging a pull request in this project is being treated more that _these changes are "perfect"_ instead of _"these changes provide a net benefit to people using the server"_.

This pull was sitting for two days without any activity—we need to be encouraging newcomers and building our community; that means putting fewer barriers in the path of new contributors and moving faster. We simply can't afford to be excessively stringent on what gets merged because we're not striving for a beautiful codebase—we're striving for a _Vanilla-compatible_ Minecraft server.

PR #4767, opened 2020-06-25
Cactus cannot be placed on a forbidden blocks

### LINKED ISSUE #4641: Cactus drops when placed on something else than sand or another cactus

morsmordere:
A fix for #4641. 
Overrode _GetPlacementBlockTypeMeta_ for CactusBlock, which gets called before a player places the cactus block, so it returns true if the placement is allowed or false if not, in which case the cactus does not drop as a block but stays in the player's hand which is the expected behavior. Tested in Minecraft client 1.12.2.
**Before:**
buggy
**After:**
fixed


src/Blocks/BlockCactus.h
```diff
@@ -23,6 +23,39 @@ class cBlockCactusHandler :
...
+	bool GetPlacementBlockTypeMeta(
+		cChunkInterface & a_ChunkInterface,
+		cPlayer & a_Player,
+		const Vector3i a_PlacedBlockPos,
+		eBlockFace a_ClickedBlockFace,
+		const Vector3i a_CursorPos,
+		BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta
+	) override
+	{
+		if (a_Player.GetWorld()->DoWithChunkAt(a_PlacedBlockPos,
```
madmaxoft, line 37: Style: When using multi-line `if`, split line after `if(`

src/Blocks/BlockCactus.h
```diff
@@ -23,6 +23,39 @@ class cBlockCactusHandler :
...
+			[this, a_PlacedBlockPos, &a_ChunkInterface](cChunk & a_Chunk)
+			{
+				auto RelPos = cChunkDef::AbsoluteToRelative(a_PlacedBlockPos);
+				return CanBeAt(a_ChunkInterface, RelPos, a_Chunk);
+			}
+			))
```
madmaxoft, line 43: Style: This should be have one less indent level (so that it's the same level as the `if` the last parenthesis belongs to.)

madmaxoft:
Since you're also first-time contributing to the project, you should add your name to the CONTRIBUTORS file, too.

src/Blocks/BlockCactus.h
morsmordere: Hopefully my last commits reflected the changes you wanted.

src/Blocks/BlockCactus.h
morsmordere: Hopefully my last commits reflected the changes you wanted.

madmaxoft:
It is going to get merged after a second person has a look at it - we generally want at least two people to review a PR before merging it.

PR #4770, opened 2020-06-30
cProtocolRecognizer goes on a diet

tigerw:
- Removed inheritance from cProtocol


Titaniumtown:
what does this do exactly?

Titaniumtown:
Like what does the commit do? Does it make cuberite faster or something?

tigerw:
It deletes a lot of boilerplate code, but no functional changes :)

Titaniumtown:
ok, got it @tigerw!

src/ClientHandle.cpp
```diff
@@ -3343,9 +3349,24 @@ void cClientHandle::ProcessProtocolInOut(void)
 		cCSLock Lock(m_CSIncomingData);
 		std::swap(IncomingData, m_IncomingData);
 	}
+
 	if (!IncomingData.empty())
 	{
-		m_Protocol->DataReceived(IncomingData.data(), IncomingData.size());
+		// Modifiable string view to support the protocol recogniser
+		// lopping off the initial handshake bytes:
+		std::string_view Data(IncomingData);
```
peterbell10, line 3357: What does this comment mean? `string_view` is by definition non-modifiable, right?

src/ClientHandle.cpp
```diff
@@ -3354,10 +3375,10 @@ void cClientHandle::ProcessProtocolInOut(void)
 		cCSLock Lock(m_CSOutgoingData);
 		std::swap(OutgoingData, m_OutgoingData);
 	}
-	auto link = m_Link;
-	if ((link != nullptr) && !OutgoingData.empty())
+
+	if ((m_Link != nullptr) && !OutgoingData.empty())
```
peterbell10, line 3379: Why the unrelated change?

The original author was probably worried about `m_Link` being reset, so this could be introducing a race condition.

src/ClientHandle.cpp
tigerw: I am passing Data in by reference and assigning to it Data.substr().

src/Protocol/ProtocolRecognizer.cpp
```diff
@@ -1225,22 +222,132 @@ void cProtocolRecognizer::HandlePacketStatusRequest(void)
...
-	if (!m_Buffer.ReadBEInt64(Timestamp))
+	if (!a_Buffer.ReadBEInt64(Timestamp))
 	{
 		return;
 	}
 
-	cPacketizer Pkt(*this, pktPingResponse);
-	Pkt.WriteBEInt64(Timestamp);
+	VERIFY(a_Out.WriteVarInt32(GetPacketID(cProtocol::ePacketType::pktPingResponse)));
+	VERIFY(a_Out.WriteBEInt64(Timestamp));
```
peterbell10: `cPacketizer` also called `SendPacket`. Is that not relevant any more?

src/Protocol/ProtocolRecognizer.cpp
```diff
@@ -1225,22 +222,132 @@ void cProtocolRecognizer::HandlePacketStatusRequest(void)
...
+		{
+			// The protocol has just been recognized, advance data start
+			// to after the handshake and leave the rest to the protocol:
+			a_Data = a_Data.substr(a_SeenData.GetUsedSpace() - a_SeenData.GetReadableSpace());
+
+			// We read more than we can handle, purge the rest:
+#ifdef NDEBUG
+			a_SeenData.SkipRead(a_SeenData.GetReadableSpace());
+#else
+			ASSERT(a_SeenData.SkipRead(a_SeenData.GetReadableSpace()));
```
peterbell10: I don't like the duplication here, how about this instead:
```cpp
[[maybe_unused]] bool Success = a_SeenData.SkipRead(a_SeenData.GetReadableSpace());
ASSERT(Success);
```

src/Protocol/ProtocolRecognizer.cpp
```diff
@@ -1225,22 +222,132 @@ void cProtocolRecognizer::HandlePacketStatusRequest(void)
...
+		}
+	}
+
+	// Flush out all queued uncompressed, unencrypted data:
+	SendPacket(a_Client, OutPacketBuffer);
+
+	// In order to avoid making a whole dummy cProtocol just for old clients,
+	// we depend on cClientHandle calling us to process old pings. We don't
+	// persist state here, so restart parsing from scratch next time and return nothing:
+	a_SeenData.ResetRead();
```
peterbell10: This seems like a weird workaround just to avoid making `cProtocolRecognizer` stateful. But I don't seem why it shouldn't be.

src/ClientHandle.cpp
peterbell10: The comment should say that then.

src/Protocol/ProtocolRecognizer.cpp
tigerw: I call sendpacket after handling all the pings, inside TryRecogniseProtocol

src/Protocol/ProtocolRecognizer.cpp
tigerw: everyday we stray farther from C

src/Protocol/ProtocolRecognizer.cpp
tigerw: BUT NO! SendPacket sends a length + data. I can't go telling it to send two data at once, it's not going to produce a valid packet! soz, i derped

tigerw:
In response to your concern I tried adopting a more functional approach; we now have different handler functions that encodes states "recognising", "has a cProtocol", "old client, handling pings only". How about now?

src/ClientHandle.cpp
```diff
@@ -71,6 +71,46 @@ cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
 	m_CurrentViewDistance(a_ViewDistance),
 	m_RequestedViewDistance(a_ViewDistance),
 	m_IPString(a_IPString),
+	m_ReceivedData(8 KiB),  // We need a larger buffer to support BungeeCord - it sends one huge packet at the start
+	m_ProtocolDataInHandler(
```
peterbell10, line 75: But now you're raising protocol recogniser's logic into `cClientHandle`. Why not just have a class `cProtocolRecognizer`? You could even store it in an `std::optional<cProtocolRecognizer>` and reset it when the concrete protocol is detected. That way we could recover the memory needed for its `m_ReceivedData`.

tigerw: 1. The client handle shouldn't have to carry around the baggage from an ephemeral, one-time operation for the rest of its lifetime, and even an optional still has to take up space.
2. The 'process data in' code can neatly call into whatever is needed to handle it right now, no nullptr/bool IsInUnrecognisedVersion checks leaving you wondering about whether the logic is right.
3. Responding to old versions should be some sort of gimped cProtocol that TryRecognise returns, but absent that it certainly shouldn't be something TryRecognise just happens to do, even though it's already recognised the protocol (as old) "but yeah we just have to keep on calling it in a loop to send data as a side effect", because that is surprising behaviour.

peterbell10: Your implementation already carries state from `cProtocolRecognizer` in cClientHandle forever. `m_ReceivingData` and `m_ProtocolDataInHandler` only exist to serve that purpose.

If your issue is the name, then why not call it `cUnknownProtocolHandler` or something. Either way, pushing more of the logic into `cClientHandle` is just messy IMO.

tigerw: So perhaps what I'm saying is, getting cClientHandle to ...handle the state transition to SendUnrecognisedPing is a natural extension to the already-existing transition from RecogniseProtocol to SendingToProtocol in cClientHandle, because SendUnrecognisedPing is just special case of Sending with less boilerplate. It is not something that should happen in Recognise.

tigerw: oops, comment merge conflict

I intend to make cProtocol use m_ReceivingData soon, which will hopefully enable, once we have protocol stuff happening off-world, some sort of zero-copy where Recogniser and Protocol_XX use and write to the same buffer.

For ProtocolDataInHandler, well, I'd argue that this is a functional-style (?) alternative to

if (IsOldVersion)
 RespondToOld()
else if (m_Protocol != null)
 Recognise()
else
 AddReceived

so it is broader than just a way to carry state for the recogniser, but more a way to track the client handle's protocol state in its entirety.

I'll try to think about how I can put less stuff in clienthandle.

tigerw:
I have once again created interspaced green and red blocks. Now?

madmaxoft:
This feels weird. I like the overall attempt to reduce the `cProtocolRecognizer`, but the lambdas everywhere are awful. Isn't there a way that's more straightforward, without the lambda-adapters everywhere? Especially the lambda-assignment in `cClientHandle`'s constructor is evil.

tigerw:
What would you suggest to initialise the initial handler?

tigerw:
I now use a member function, what do you think?

madmaxoft:
What's wrong with the simple approach? Along the lines of:
```cpp
void cClientHandle::ProcessProtocolInOut()
{
	// ...
	if (!IncomingData.empty())
	{
		if (m_Protocol == nullptr)
		{
			// Protocol not yet detected, store the data and try recognizing:
			m_Protocol = cProtocolRecognizer::TryRecognizeProtocol(IncomingData);
			if (m_Protocol == nullptr)
			{
				// Still not detected, put the IncomingData back into the storage for the next try:
				cCSLock Lock(m_CSIncomingData);
				IncomingData.append(m_IncomingData):
				std::swap(m_IncomingData, IncomingData);
			}
		}
		else
		{
			m_Protocol->DataReceived(IncomingData);
		}
	}
```

No lambdas, no `bind`, all simple code, all in one place that actually handles the incoming data.

Also, since it's already a `Recognizer`, please use `Recognize` instead of `Recognise`. Yes, we do use US English (except where I didn't catch it soon enough).

tigerw:
I want to use a single backing cByteBuffer for all incoming data, for efficiency and also to deal with the case that recognition isn't a read-only operation:

In your example, how should I deal with
1. The recogniser needs to strip off some variable number of bytes from the start of the data (the handshake), since the concrete protocol implementations can't handle it.
2. When the recognizer recognises an old protocol, it likewize needs to remove a variable number of processed bytes (old ping/old handshake), in order to not process the same packet over and over; this was one of the defects of my initial try and one peterbell mentioned.
3. In a similar vein when an old protocol is found, the recogniser needs to switch to a different code path to handle the old handshakes, and needs something to tell it this. In master we're using state to keep track of it, but I am very against carrying state around for a one-time operation for the entire, potentially very long lifetime of cClientHandle.

I wanted to go with the current functional solution since I think it neatly handles these, but am I correct in saying your concern is mainly with the style aspect? Maybe I could do some further thinking to alleviate this?

tigerw:
> Yes, we do use US English

> tigerw
> 📌 London

i am stabbed through the heart :(

madmaxoft:
That should still be possible. The `cProtocolRecogniZer` receives the `cByteBuffer` containing all the data received so far; it tries to recognize the protocol, and if it does, it also "reads" the bytes it's supposed to strip at the front. It then hands the `cByteBuffer` to the new `cProtocol` descendant just before returning it. If it doesn't recognize the protocol, it *keeps* the data in the `cByteBuffer` for the next attempt (there's the state).

Yes, the recogniZer will parse the buffer over and over from the start until it recogniZes the protocol, the assumption is that that happens only a handful of times so it doesn't bother us at all.

madmaxoft:
> > Yes, we do use US English
> 
> > tigerw
> > pushpin London
> 
> i am stabbed through the heart :(

Be glad we're not using some obscure language, such as CzEnglish or HinGlish :)

tigerw:
Indeed, I am doing this already for the "happy path" of finding a protocol we support. But for old versions, barring not consuming the handshake bytes at all (which peterbell disagreed with), we need a way to tell the recogniZer to switch modes.

Also reparzing the same data carries the risk that a malicious client could force the server to do lots of work and net activity without even having to connect.

tigerw:
So, what say you? At the expense of having to initialise the handler we get a one-liner inside ProcessProtocolInOut and functionality neatly confined to the protocol recogniser, and 900 lines of code less. Do I have your go-ahead?

madmaxoft:
"Neatly" is not the word I'd have used here. You need a new datatype for the recognizer's return value, kept in sync in two places. Returning a function is something "neat" code just doesn't do. I still honestly believe that my simple way is much easier to follow (read: maintainable).

tigerw:
> Returning a function is something "neat" code just doesn't do

Haskell developers: visible displeasure

What about my comment on the functional issue of reparsing though? I don't think the simple solution handles that case.

Maybe I could try something else.

madmaxoft:
Reparsing? You mean parsing the same thing over and over, with malicious attackers overwhelming the server? I don't think that should be a problem, there's not that much work done when parsing, and if it becomes a problem, we can fix it later on.

tigerw:
I have made more changes, I think these address the points you raise:
* The client handle no longer concerns itself with any recognition logic
* The recogniser is renamed to UnknownProtocol, and only carries state relevant to its entire lifetime
* No longer return llamas; handlers are tightly confined to one function

madmaxoft:
I'm even less happy with that code. There's no documentation for the `cUnknownProtocol` class, and from its overview it is quite unclear what it does. Why does an unknown protocol have a `cProtocol` pointer inside it? (Rhetorical question) Constructs such as `m_Protocol.operator->()` are just hair-rising.

tigerw:
Sorry, the refactor was barebones for a more readable diff, of course I can add the rest later. Please bear with me a moment.

tigerw:
Hey, thanks for waiting. I've made the changes more complete, main additions are:
* Rename cProtocolRecognizer to cMultiVersionProtocol. Rationale is that the class stores a cProtocol now and dispatches calls to it.
+ Add and adjust class docs
+ Add VersionRecognitionSuccessful to test if the underlying protocol is null

Changes are in d39d8f7

tigerw:
May I count on your support to merge?

src/Protocol/ProtocolRecognizer.h
```diff
@@ -40,130 +37,67 @@ class cProtocolRecognizer:
...
+	{
+		return m_Protocol != nullptr;
+	}
 
-	/** Buffer for the incoming data until we recognize the protocol */
-	cByteBuffer m_Buffer;
+	/** Convenience overload to enable redirecting sends to the underlying implementation. */
+	auto & operator->()
+	{
+		return m_Protocol;
```
peterbell10, line 54: Instead of pretending to be a pointer, why not just have a `GetConcreteProtocol` or something?

src/Protocol/ProtocolRecognizer.h
```diff
@@ -40,130 +37,67 @@ class cProtocolRecognizer:
...
-	void HandlePacketStatusRequest();
-	void HandlePacketStatusPing();
+	/* Ping handler for unrecognised versions. */
+	void HandlePacketStatusPing(cClientHandle & a_Client, cByteBuffer & a_Out);
 
-	/** Tries to recognize protocol based on m_Buffer contents; returns true if recognized */
-	bool TryRecognizeProtocol(void);
+	/** The actual protocol implementation.
+	Created when recognition of the client version succeeds with a version we support. */
+	std::unique_ptr<cProtocol> m_Protocol;
```
peterbell10, line 98: Please stop randomly reordering things in files. It makes the diff so much harder to read.

src/Protocol/ProtocolRecognizer.h
tigerw: We unfortunately have a lot of m_Protocol->SendSaucyXXX inside cClientHandle, so this avoids a mass-replace

src/Protocol/ProtocolRecognizer.h
tigerw: soz

PR #4771, opened 2020-07-02
cWorld * -> cWorld

tigerw:
`map<string, cWorld *>` has mega evolved into `map<string, cWorld>`!

Ref: #4695

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -453,10 +453,9 @@ void cRoot::LoadWorlds(cDeadlockDetect & a_dd, cSettingsRepositoryInterface & a_
 		const AStringVector WorldNames{ "world", "world_nether", "world_the_end" };
 		m_DefaultWorldName = "world";
 
-		// TODO: C++17 try_emplace
-		m_WorldsByName.emplace(std::piecewise_construct, std::forward_as_tuple("world"), std::forward_as_tuple("world", "world", a_dd, WorldNames));
-		m_WorldsByName.emplace(std::piecewise_construct, std::forward_as_tuple("world_nether"), std::forward_as_tuple("world_nether", "world_nether", a_dd, WorldNames, dimNether, "world"));
-		m_WorldsByName.emplace(std::piecewise_construct, std::forward_as_tuple("world_the_end"), std::forward_as_tuple("world_the_end", "world_the_end", a_dd, WorldNames, dimEnd, "world"));
+		m_WorldsByName.try_emplace("world", "world", "world", a_dd, WorldNames);
+		m_WorldsByName.try_emplace("world_nether", "world_nether", "world_nether", a_dd, WorldNames, dimNether, "world");
+		m_WorldsByName.try_emplace("world_the_end", "world_the_end", "world_the_end", a_dd, WorldNames, dimEnd, "world");
 		return;
 	}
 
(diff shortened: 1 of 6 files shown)


tigerw:
> To be honest, I never trusted containers-of-values, never knew if there's the guarantee that the value will stay at one place or if the container is allowed to move it around in memory (thus invalidating pointers).

ahh, I finally understand what you mean... I was thinking, well I never call .erase() on a world, what does he mean by invalidating pointers?

but now I see, you mean what happens with vector when it runs out of capacity, all the elements get moved and their addresses change. Which I agree, would be fatal to the World.

and then you add later on that maps guarantee addresses to their elements stay valid when inserting (how convenient) which resolves this concern.

ahh

src/Root.h
```diff
@@ -212,10 +212,10 @@ class cRoot
 		cCommandOutputCallback * m_Output;
 	} ;
 
-	typedef std::map<AString, cWorld *> WorldMap;
+	typedef std::map<AString, cWorld> WorldMap;
 	typedef std::vector<cCommand> cCommandQueue;
 
-	cWorld * m_pDefaultWorld;
+	AString m_DefaultWorldName;
```
peterbell10, line 218: What's the point of this change?

tigerw: haha, good pointer joke :D

...I've forgotten. Maybe to make unambiguous that it can't be null? I'll try to change it back and see if I hit any blocks.

tigerw: Ok! Reverted with apparently no ill effects.

tigerw:
Merging tomorrow if no further thoughts

src/Root.cpp
```diff
@@ -53,7 +53,6 @@ cRoot * cRoot::s_Root = nullptr;
 
 
 cRoot::cRoot(void) :
-	m_pDefaultWorld(nullptr),
```
peterbell10, line 56: Should probably still be initialising here?

tigerw: There's always a default world, so it's initialised by LoadWorlds

peterbell10: Which isn't run in the constructor, so there is period where it is uninitialised.

tigerw: Sure, but LoadWorlds happens before any plugin, object attempts to call GetDefaultWorld. Even if it didn't, every caller assumes the return value can be used immediately/stored, so it's undefined behaviour anyway if that's violated.

peterbell10: With a `nullptr` you can easily see in the debugger that it's uninitialized. If it's just a random garbage pointer then you don't know if it's uninitialized or if there's memory corruption.

tigerw: Wouldn't you agree though this will never happen, by the above reasoning? So this line of code is unneeded.

madmaxoft: It is a good practice to never leave any pointer uninitialized.

tigerw: I suppose I can add a debug check for it.

tigerw:
Ready to merge?

madmaxoft:
No, put the initialization there unconditionally.
I don't get it what you make such a big fuss of anyway, it's just a single variable, not like it's a perf bottleneck or something.

tigerw:
Unless my reasoning for why is unsound, it's dead code. I don't see the point for initialising it twice. You guys are talking about  hypotheticals but I claim that it will never be uninitialised and (seeing as it's quite a critical codepath so we'd know if it bugged) it never has been uninitialised.

Sure I could just revert the change but this is a learning process for me, so through a discussion I'd like to know and be convinced about why you guys think it's better to do it, and why a debug-only check doesn't satisfy.

tigerw:
People seem busy so reverting unpopular change and merging.

PR #4775, opened 2020-07-12
Remove unnecessary includes

tigerw:
\> Cross platform server

\> A ton of POSIX includes

Crossbreed


peterbell10:
Looks good, can you check in the `Tools` folder as well.

tigerw:
These were done manually, I'm not sure why all the Tools straight up copy Globals.h but I'll need a bit more time to check, hold on

PR #4780, opened 2020-07-20
1.14 connection support

tigerw:
Can connect, but not do anything since incoming packets are ignored.

### DIFF
--- a/src/ByteBuffer.cpp
+++ b/src/ByteBuffer.cpp
@@ -472,7 +472,7 @@ bool cByteBuffer::ReadLEInt(int & a_Value)
 
 
 
-bool cByteBuffer::ReadPosition64(int & a_BlockX, int & a_BlockY, int & a_BlockZ)
+bool cByteBuffer::ReadXYZPosition64(int & a_BlockX, int & a_BlockY, int & a_BlockZ)
 {
 	CHECK_THREAD
 	Int64 Value;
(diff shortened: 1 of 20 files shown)


src/WorldStorage/FastNBT.cpp
```diff
@@ -638,15 +638,15 @@ void cFastNBTWriter::AddByteArray(const AString & a_Name, const char * a_Value,
...
+void cFastNBTWriter::AddIntArray(const AString & a_Name, const Int32 * a_Value, size_t a_NumElements)
 {
 	TagCommon(a_Name, TAG_IntArray);
 	UInt32 len = htonl(static_cast<UInt32>(a_NumElements));
 	size_t cap = m_Result.capacity();
 	size_t size = m_Result.length();
 	if ((cap - size) < (4 + a_NumElements * 4))
 	{
-		m_Result.reserve(size + 4 + (a_NumElements * 4));
+		m_Result.reserve(size + 4 + (a_NumElements * 4));  // TODO: why _size_ + 4 ...?
```
peterbell10, line 649: `size` for what already there, 4 bytes for the array length and `a_NumElements * 4` for the array elements.

tigerw: Thanks. I had the mistaken notion that reserve was relative to the current size instead of the new capacity...

tigerw:
Merging as people are busy.

src/ByteBuffer.cpp
```diff
@@ -488,7 +488,7 @@ bool cByteBuffer::ReadPosition64(int & a_BlockX, int & a_BlockY, int & a_BlockZ)
 
 	// If the highest bit in the number's range is set, convert the number into negative:
 	a_BlockX = ((BlockXRaw & 0x02000000) == 0) ? static_cast<int>(BlockXRaw) : -(0x04000000 - static_cast<int>(BlockXRaw));
-	a_BlockY = ((BlockYRaw & 0x0800) == 0)     ? static_cast<int>(BlockYRaw) : -(0x0800     - static_cast<int>(BlockYRaw));
+	a_BlockY = ((BlockYRaw & 0x0800) == 0)     ? static_cast<int>(BlockYRaw) : -(0x01000    - static_cast<int>(BlockYRaw));
```
madmaxoft, line 491: We really need to have some unit tests for the `cByteBuffer`, it's riddled with magic numbers and changes like this may break something in the old protocols and we wouldn't know.

tigerw: Definitely agreed.

PR #4783, opened 2020-07-26
Moved ChunkSender.cpp/h to Protocol

tigerw:
Opening for @Howaner


bearbin:
What's the reasoning behind this? I can't see any pressing reason this needs to be done right now, although it's such a minor change that we may as well anyway.

tigerw:
y'know, what GLaDOS said about Aperture

peterbell10:
`cChunkSender` also doesn't seem to do anything protocol specific, so I'm not sure why it should be in the protocol folder?

tigerw:
it's a glorified queue to implement a critical piece of protocol functionality

bearbin:
I would just keep it where it is. In any case this conflicts and needs a rebase.

tigerw:
yup, ready.

peterbell10:
You still haven't justified why this should be in the Protocol folder. Everything else in that folder is specific to some protocol version, whether it be a specific minecraft version, or one of the other Mojang APIs. This however is completely protocol agnostic.

tigerw:
Chunk sending is a protocol duty.

peterbell10:
No, it's completely agnostic of the protocol.

tigerw:
Protocol as in, functions that communicate server state to the client to enable it to join and engage in gameplay, not as in JE version 753.

tigerw:
assuming begrudging ok or at least indifference

bearbin:
Maybe not, this PR is pretty pointless and causing grief, closing it.

tigerw:
wait what, I would like this merged please

tigerw:
Reopening to this effect. This is some reorganisation, I don't see any technical objections, and besides discussion has been cordial, no grief at all :)

bearbin:
Reopening is not the correct response here. This PR is bad and I am rejecting it. ChunkSender doesn't belong in  the Protocol directory (which may be better described as _Protocols_).

There may not have been _technical_ objections, because this is a non-technical change, just organisation. This is not something to get worked up over.

tigerw:
I strongly object. We very rarely label PRs as 'bad' and close, then delete the branch entirely. I spent time on this, and also get a say on what is merged or not.

bearbin:
Indeed, you get a say on which PRs are merged or not, but in the case of one's own creation, it is harder to judge its value towards the project. In the case of this PR you had two other developers arguing against a merge. By the rule of majority consensus, the PR should not have gone ahead. And then when you didn't get a reply for **1 day**, you proposed to merge it yourself. That's wrong.

Now I can't speak for peterbell10, but your comment yesterday really set me off, otherwise I wouldn't have closed this PR.

> assuming begrudging ok or at least indifference
 
This reads to me as:

> Nevermind this discussion we've been having; I don't care about your viewpoint; I'm going ahead anyway. And not getting around to Cuberite means you're lazy and can't be bothered to work with me.

Now I'm sure that's not what you meant; but to get into a discussion and clarify that would be a lot of effort, and would far outweigh the value of this PR (which is negative, in any case...). So, although rarely may it be the case, this was the correct time to close an issue and stop discussion.

Now this really is my final comment on this issue.

peterbell10:
I have to say I completely agree with @bearbin. Your pushing to get PRs merged can at times seem quite aggressive. A friendly ping is fine. Assuming we don't care is verging on insulting.

The actual contents of this PR are hard to argue for or against on a technical level since it doesn't make any functional difference to the code. `cChunkSender` bridges `cWorld` to some obviously protocol-related code, but I would say that `cChunkSender` itself is more about general server plumbing than it is about protocols. 

>We very rarely label PRs as 'bad' and close, then delete the branch entirely.

@bearbin and I have both expressed that we don't think this is a valuable change. It's also a completely binary change, so there's no possibility of a compromise. I don't think there's anywhere else to go from here.

tigerw:
I've talked about this with bearbin already, I'm writing this PR off as a bad experience.

About aggressive merging,

a user said unto this project, "you don't stand a chance", "no one will use it". That hits hard! You don't spend seven years of your life on something and not hope that it makes an impact. I recognise we're chasing a moving target, but I'm not giving up.

PR #4785, opened 2020-07-28
Removed BiomeDef and ChunkDef from Globals.h

### LINKED ISSUE #4667: Remove BiomeDef.h and ChunkDef.h from Globals.h
The two headers are by no means anything needed globally, so they should be instead included only in the places they are actually needed, rather than Globals.

Right now, any change in ChunkDef.h causes recompilation of even OSSupport code, for example, which is definitely needless.

RobertPino:
Fixed #4667 -
Removed BiomeDef.h and ChunkDef.h from Globals.h

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -359,5 +359,3 @@ auto ToUnsigned(T a_Val)
 
 // Common headers (part 2, with macros):
 #include "Vector3.h"
-#include "BiomeDef.h"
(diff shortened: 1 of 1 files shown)


madmaxoft:
Well of course it's not *that* easy, you can't just remove them, you need to distribute them to wherever they're needed :)

PR #4786, opened 2020-07-29
imprementing plugin message to send server brand

12xx12:
I think abbout working on the other message types - don't merge right now - this is just for review purposes


Titaniumtown:
I'll test this later! Thanks for addressing my issue!

Titaniumtown:
Just a thought, maybe it could be implemented into the API to be able to change the server "brand".

12xx12:
@Titaniumtown right now the "brand" is just a string hardcoded into the code - it shouldn't be that hard but i don't see the sense in that because brand just describes the kind of te server. Although I am not the person to decide this.

Titaniumtown:
Maybe by default it can be "Cuberite" but it would just be neat to have the functionality.

Titaniumtown:
Btw it's compiling rn. I'll do some testing.

Titaniumtown:
It worked :)

bearbin:
Is this still WIP or ready for review/merge?

12xx12:
WIP. Gotta add a parameter to the transmission which was not there before in die older versions. 

Gonna do that later

12xx12:
@bearbin ready for review (when checks are done)

PR #4789, opened 2020-07-29
Adding Statistics and Achievements for newer Network standarts

12xx12:
I was trying to implement the SendStatistics method for 1.13 and saw the solution for handling statistics and achievents.
If correctly implemented you would just iterate through all avaiable and send them.

I had a look around in the pull requests and fond only the old one from 2014 implementing the 1.8 achievements.

Things I did/do

- [x] move to a string based system
  - [x] there's now a vector of all possible satistics and achievement names. 
  - [x] the map which stores the achievements per player is now String -> int / String -> String (depending on version)
  - [x] create enum to improve readability and easy implementation for future updates
- [x] add lookup tables for every supported version. Achievent GUI names the same for 1.8 - 1.11
  - [x] 1.8
  - [x] 1.8.2
    - [x] Cake Slices Eaten
    - [x] Cauldrons Filled
    - [x] Water Taken from Cauldron
    - [x] Armor Pieces Cleaned
    - [x] Banners Cleaned
    - [x] Interactions with Brewing Stand
    - [x] Interactions with Beacon
    - [x] Droppers Searched
    - [x] Hoppers Searched
    - [x] Dispensers Searched
    - [x] Noteblocks played
    - [x] Noteblocks tuned
    - [x] Plants potted
    - [x] Trapped Chests Triggered
    - [x] Ender Chests Opened
    - [x] Items Enchanted
    - [x] Records Played
    - [x] Interactions with Furnace
    - [x] Interactions with Crafting Table
    - [x] Chests Opened
  - [x] 1.9
    - [x] Added Sneak Time.
    - [x] Added Times Slept in a Bed.
    - [x] Added Distance by Elytra.
  - [x] 1.11.1
    - [x] Removed Junk Fished and Treasure Fished
  - [x] 1.12 removed achievements from lookup
At this point the achievements were officlially named "minecraft.<type>" and transmitted by numeric ID insted of a string ID
  - [x] 1.13
  - [x] 1.13.1
- [x] only send dirty statistics instead if the whole thing
- [x] test
  - [x] 1.8
  - [x] 1.8.2
  - [x] 1.9
  - [x] 1.11.1
  - [x] 1.13
  - [x] 1.13.1
For testing I created a Json file that writes a value to any possbile thing: (and even tests the conversion from the old type)
https://pastebin.com/Yx0UbK8i

- [x] create lua that reads the Registries and creates the new switch/case thing
- [x] checkstyle variable names

My aim would be to make future (1.14+) statistics easier to implement.

Statistics and Achievements are tied in this pull request because they are tied together in 1.8 ->1.11
The New Advancements are gonna be a new pull requestfor Version 1.12 +

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -334,7 +334,7 @@ void cPlayer::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	}
 
 
-	m_Stats.AddValue(statMinutesPlayed, 1);
+	m_Stats.AddValue(Statistic::PlayOneMinute, 1);
 
 	// Handle the player detach, when the player is in spectator mode
 	if (
@@ -742,7 +742,7 @@ void cPlayer::TossItems(const cItems & a_Items)
 		return;
 	}
 
-	m_Stats.AddValue(statItemsDropped, static_cast<StatValue>(a_Items.Size()));
+	m_Stats.AddValue(Statistic::Drop, static_cast<cStatManager::StatValue>(a_Items.Size()));
 
 	const auto Speed = (GetLookVector() + Vector3d(0, 0.2, 0)) * 6;  // A dash of height and a dollop of speed
 	const auto Position = GetEyePosition() - Vector3d(0, 0.2, 0);  // Correct for eye-height weirdness
@@ -1110,7 +1110,7 @@ bool cPlayer::DoTakeDamage(TakeDamageInfo & a_TDI)
 				NotifyNearbyWolves(static_cast<cPawn*>(a_TDI.Attacker), true);
 			}
 		}
-		m_Stats.AddValue(statDamageTaken, FloorC<StatValue>(a_TDI.FinalDamage * 10 + 0.5));
+		m_Stats.AddValue(Statistic::DamageTaken, FloorC<cStatManager::StatValue>(a_TDI.FinalDamage * 10 + 0.5));
 		return true;
 	}
 	return false;
@@ -1168,7 +1168,7 @@ void cPlayer::KilledBy(TakeDamageInfo & a_TDI)
 	{
 		Pickups.Add(cItem(E_ITEM_RED_APPLE));
 	}
-	m_Stats.AddValue(statItemsDropped, static_cast<StatValue>(Pickups.Size()));
+	m_Stats.AddValue(Statistic::Drop, static_cast<cStatManager::StatValue>(Pickups.Size()));
 
 	m_World->SpawnItemPickups(Pickups, GetPosX(), GetPosY(), GetPosZ(), 10);
 	SaveToDisk();  // Save it, yeah the world is a tough place !
@@ -1234,7 +1234,7 @@ void cPlayer::KilledBy(TakeDamageInfo & a_TDI)
 		}
 	}
 
-	m_Stats.AddValue(statDeaths);
+	m_Stats.AddValue(Statistic::Deaths);
 
 	m_World->GetScoreBoard().AddPlayerScore(GetName(), cObjective::otDeathCount, 1);
 }
@@ -1249,7 +1249,7 @@ void cPlayer::Killed(cEntity * a_Victim)
 
 	if (a_Victim->IsPlayer())
 	{
-		m_Stats.AddValue(statPlayerKills);
+		m_Stats.AddValue(Statistic::PlayerKills);
 
 		ScoreBoard.AddPlayerScore(GetName(), cObjective::otPlayerKillCount, 1);
 	}
@@ -1257,10 +1257,10 @@ void cPlayer::Killed(cEntity * a_Victim)
 	{
 		if (static_cast<cMonster *>(a_Victim)->GetMobFamily() == cMonster::mfHostile)
 		{
-			AwardAchievement(achKillMonster);
+			AwardAchievement(Statistic::AchKillEnemy);
 		}
 
-		m_Stats.AddValue(statMobKills);
+		m_Stats.AddValue(Statistic::MobKills);
 	}
 
 	ScoreBoard.AddPlayerScore(GetName(), cObjective::otTotalKillCount, 1);
@@ -1671,43 +1671,32 @@ void cPlayer::SetIP(const AString & a_IP)
 
 
 
-unsigned int cPlayer::AwardAchievement(const eStatistic a_Ach)
+void cPlayer::AwardAchievement(const Statistic a_Ach)
 {
-	eStatistic Prerequisite = cStatInfo::GetPrerequisite(a_Ach);
-
-	// Check if the prerequisites are met
-	if (Prerequisite != statInvalid)
+	// Check if the prerequisites are met:
+	if (!m_Stats.SatisfiesPrerequisite(a_Ach))
 	{
-		if (m_Stats.GetValue(Prerequisite) == 0)
-		{
-			return 0;
-		}
+		return;
 	}
 
-	StatValue Old = m_Stats.GetValue(a_Ach);
-
-	if (Old > 0)
+	// Increment the statistic and check if we already have it:
+	if (m_Stats.AddValue(a_Ach) != 1)
 	{
-		return static_cast<unsigned int>(m_Stats.AddValue(a_Ach));
+		return;
 	}
-	else
+
+	if (m_World->ShouldBroadcastAchievementMessages())
 	{
-		if (m_World->ShouldBroadcastAchievementMessages())
-		{
-			cCompositeChat Msg;
-			Msg.SetMessageType(mtSuccess);
-			Msg.AddShowAchievementPart(GetName(), cStatInfo::GetName(a_Ach));
-			m_World->BroadcastChat(Msg);
-		}
-
-		// Increment the statistic
-		StatValue New = m_Stats.AddValue(a_Ach);
-
-		// Achievement Get!
-		m_ClientHandle->SendStatistics(m_Stats);
-
-		return static_cast<unsigned int>(New);
+		cCompositeChat Msg;
+		Msg.SetMessageType(mtSuccess);
+		// TODO: cCompositeChat should not use protocol-specific strings
+		// Msg.AddShowAchievementPart(GetName(), nameNew);
+		Msg.AddTextPart("Achivement get!");
+		m_World->BroadcastChat(Msg);
 	}
+
+	// Achievement Get!
+	m_ClientHandle->SendStatistics(m_Stats);
 }
 
 
@@ -2334,10 +2323,17 @@ bool cPlayer::LoadFromFile(const AString & a_FileName, cWorldPtr & a_World)
 		m_SpawnWorld = cRoot::Get()->GetDefaultWorld();
 	}
 
-	// Load the player stats.
-	// We use the default world name (like bukkit) because stats are shared between dimensions / worlds.
-	cStatSerializer StatSerializer(cRoot::Get()->GetDefaultWorld()->GetDataPath(), GetName(), GetUUID().ToLongString(), &m_Stats);
-	StatSerializer.Load();
+	try
+	{
+		// Load the player stats.
+		// We use the default world name (like bukkit) because stats are shared between dimensions / worlds.
+		cStatSerializer StatSerializer(m_Stats, cRoot::Get()->GetDefaultWorld()->GetDataPath(), GetUUID().ToLongString());
+		StatSerializer.Load();
+	}
+	catch (...)
+	{
+		LOGWARNING("Failed loading player statistics");
+	}
 
 	FLOGD("Player {0} was read from file \"{1}\", spawning at {2:.2f} in world \"{3}\"",
 		GetName(), a_FileName, GetPosition(), a_World->GetName()
@@ -2476,10 +2472,14 @@ bool cPlayer::SaveToDisk()
 		return false;
 	}
 
-	// Save the player stats.
-	// We use the default world name (like bukkit) because stats are shared between dimensions / worlds.
-	cStatSerializer StatSerializer(cRoot::Get()->GetDefaultWorld()->GetDataPath(), GetName(), GetUUID().ToLongString(), &m_Stats);
-	if (!StatSerializer.Save())
+	try
+	{
+		// Save the player stats.
+		// We use the default world name (like bukkit) because stats are shared between dimensions / worlds.
+		cStatSerializer StatSerializer(m_Stats, cRoot::Get()->GetDefaultWorld()->GetDataPath(), GetUUID().ToLongString());
+		StatSerializer.Save();
+	}
+	catch (...)
 	{
 		LOGWARNING("Could not save stats for player %s", GetName().c_str());
 		return false;
@@ -2659,12 +2659,12 @@ void cPlayer::UpdateMovementStats(const Vector3d & a_DeltaPos, bool a_PreviousIs
 		return;
 	}
 
-	StatValue Value = FloorC<StatValue>(a_DeltaPos.Length() * 100 + 0.5);
+	const auto Value = FloorC<cStatManager::StatValue>(a_DeltaPos.Length() * 100 + 0.5);
 	if (m_AttachedTo == nullptr)
 	{
 		if (IsFlying())
 		{
-			m_Stats.AddValue(statDistFlown, Value);
+			m_Stats.AddValue(Statistic::FlyOneCm, Value);
 			// May be flying and doing any of the following:
 		}
 
@@ -2672,17 +2672,18 @@ void cPlayer::UpdateMovementStats(const Vector3d & a_DeltaPos, bool a_PreviousIs
 		{
 			if (a_DeltaPos.y > 0.0)  // Going up
 			{
-				m_Stats.AddValue(statDistClimbed, FloorC<StatValue>(a_DeltaPos.y * 100 + 0.5));
+				m_Stats.AddValue(Statistic::ClimbOneCm, FloorC<cStatManager::StatValue>(a_DeltaPos.y * 100 + 0.5));
 			}
 		}
 		else if (IsInWater())
 		{
-			m_Stats.AddValue(statDistSwum, Value);
+			// TODO: implement differentiation between diving and swimming
+			m_Stats.AddValue(Statistic::WalkUnderWaterOneCm, Value);
 			AddFoodExhaustion(0.00015 * static_cast<double>(Value));
 		}
 		else if (IsOnGround())
 		{
-			m_Stats.AddValue(statDistWalked, Value);
+			m_Stats.AddValue(Statistic::WalkOneCm, Value);
 			AddFoodExhaustion((IsSprinting() ? 0.001 : 0.0001) * static_cast<double>(Value));
 		}
 		else
@@ -2690,13 +2691,13 @@ void cPlayer::UpdateMovementStats(const Vector3d & a_DeltaPos, bool a_PreviousIs
 			// If a jump just started, process food exhaustion:
 			if ((a_DeltaPos.y > 0.0) && a_PreviousIsOnGround)
 			{
-				m_Stats.AddValue(statJumps, 1);
+				m_Stats.AddValue(Statistic::Jump, 1);
 				AddFoodExhaustion((IsSprinting() ? 0.008 : 0.002) * static_cast<double>(Value));
 			}
 			else if (a_DeltaPos.y < 0.0)
 			{
 				// Increment statistic
-				m_Stats.AddValue(statDistFallen, static_cast<StatValue>(std::abs(a_DeltaPos.y) * 100 + 0.5));
+				m_Stats.AddValue(Statistic::FallOneCm, static_cast<cStatManager::StatValue>(std::abs(a_DeltaPos.y) * 100 + 0.5));
 			}
 			// TODO: good opportunity to detect illegal flight (check for falling tho)
 		}
@@ -2705,15 +2706,15 @@ void cPlayer::UpdateMovementStats(const Vector3d & a_DeltaPos, bool a_PreviousIs
 	{
 		switch (m_AttachedTo->GetEntityType())
 		{
-			case cEntity::etMinecart: m_Stats.AddValue(statDistMinecart, Value); break;
-			case cEntity::etBoat:     m_Stats.AddValue(statDistBoat,     Value); break;
+			case cEntity::etMinecart: m_Stats.AddValue(Statistic::MinecartOneCm, Value); break;
+			case cEntity::etBoat:     m_Stats.AddValue(Statistic::BoatOneCm,     Value); break;
 			case cEntity::etMonster:
 			{
 				cMonster * Monster = static_cast<cMonster *>(m_AttachedTo);
 				switch (Monster->GetMobType())
 				{
-					case mtPig:   m_Stats.AddValue(statDistPig,   Value); break;
-					case mtHorse: m_Stats.AddValue(statDistHorse, Value); break;
+					case mtPig:   m_Stats.AddValue(Statistic::PigOneCm,   Value); break;
+					case mtHorse: m_Stats.AddValue(Statistic::HorseOneCm, Value); break;
 					default: break;
 				}
 				break;
--- a/src/Protocol/Protocol_1_8.cpp
+++ b/src/Protocol/Protocol_1_8.cpp
@@ -19,7 +19,6 @@ Implements the 1.8 protocol classes:
 #include "../EffectID.h"
 #include "../StringCompression.h"
 #include "../CompositeChat.h"
-#include "../Statistics.h"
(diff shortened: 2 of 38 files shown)


tigerw:
It might be easier to deal with the protocol side and the stats tracking side separately.

In 1.14+ the vanilla server provides [data generators](https://wiki.vg/Data_Generators) that gives us a list of all statistics and their protocol IDs. We should probably generate this list and make the existing server code use it, with the protocols handling the translation to IDs.

Then separately, we could start adding more statistics tracking to the main server code.

12xx12:
@tigerw 

That's a nice tool, thank you for your advice - I spent like 30 min figuring out which ID was wich one -.-

So your advice is we use the minecraft names and the .json from the server to convert the names to IDs for sending ( and future compability )

the thing is what are we gona do with the old things and moved entries version < 1.14
for 1.13 I did the manual labour. the other ones may be retrieved from the version history (1.12-1.10) for 1.9 and 1.8 it's already there

So for every protocol:
- we iterate through the list of all statistics
- do a lookup in the json for the version
- if there's no result we don't send else send

the actual counting i would implement later on - should be many one-liners

tigerw:
Yeah, that's what I had in mind. We do support multiple protocols but the actual logic really only follows the very latest Minecraft version. If an old protocol has something that was removed in the latest, we do our best to translate to something similar in the protocol but it's a best-effort attempt.

About the mappings, we already have a couple of cpp files generated from the json, see [Registries/](https://github.com/cuberite/cuberite/tree/master/src/Registries) and [Palettes/](https://github.com/cuberite/cuberite/tree/master/src/Protocol/Palettes). That way we don't have to incur a performance hit looking up a json file each time.

I have a Lua script which I used to make the files in there, it may be of use to you if you want?

12xx12:
@tigerw 

yeah that would be nice - I broke my mind to do the lookup without any big performance losses and having this best as a const map<string, int> would be fastest i think

madmaxoft:
It should be fairly easy to run the basic stylecheck on your own computer. You just need Lua in path and set to execute .lua files, then doubleclicking the src/CheckBasicStyle.lua file should perform the check. It's even possible to make git run the check before each commit.

12xx12:
Sorry for the mess with the commits - made a mistake and mixed some pull requests

madmaxoft:
Yea, you seem to be struggling with git. Need any assistance? Feel free to ask.

You should really push to the PR only when your work is more or less ready, otherwise your commits overload the buildservers.

12xx12:
Yeah - I have use subversion until now and some git for private stuff.

I have a few question: 

if I switch branches do local (not pushed) commits stay there? Might result in less pushes.

The Most of the mess resulted from me cloning not from the main branch but from my other branch...

The other think was that I tried to revert those changes in the wrong order

I know now better 

I'm thinking about ditching the whole branch, copying the changes to a new branch with a few commits to make this less messy. Is this worth it

madmaxoft:
Yes, all local branches in git stay as they are, even if you checkout a different branch or push from a different branch.

12xx12:
The continuous-integration/travis-ci/pr throws:

cStatManager.GetName (function)
cStatManager.GetPrerequisite (function)

do I have to add them manually or run script to add them? I had a look around in the contributing.md but didn't find anything

madmaxoft:
It means you added new functions accessible in the Lua API, by the plugins. You need to add documentation for them, in the APIDump plugin. Go to Server/Plugins/APIDump and find the .lua file that describes the `cStatManager` API (most likely APIDesc.lua), see around in that file and add the documentation for those functions.
Alternatively, if you didn't mean to export them to the API, you need to move them out of the `// tolua_begin` and `// tolua_end` comment pair, which marks the API for exporting.

12xx12:
Thnak you

 @tigerw  I modyfied your script and put it in the folder Tools/BlockTypePaletteGenerator

12xx12:
I do have a few more questions:

1. should global variables have a prefix "g_"?

2. you don't allow underscores in variable names. I am not quite sure how to label the global lookup tables because I try to find obvious names and using the number of the network protocol seems not that obvious. Or may I use the same layout as the protocol names? (With underscores)

src/Entities/Player.cpp
```diff
@@ -1673,35 +1673,36 @@ void cPlayer::SetIP(const AString & a_IP)
...
+	if (Prerequisite != "")
 	{
 		if (m_Stats.GetValue(Prerequisite) == 0)
 		{
 			return 0;
 		}
 	}
 
-	StatValue Old = m_Stats.GetValue(a_Ach);
+	StatValue Old = m_Stats.GetValue(nameNew);
```
12xx12, line 1688: I changed this because the name lookup has already been done and the GetValue with the enum value (a_Ach) simply performs another lookup

src/Entities/Player.cpp
```diff
@@ -1673,35 +1673,36 @@ void cPlayer::SetIP(const AString & a_IP)
...
 
 	if (Old > 0)
 	{
-		return static_cast<unsigned int>(m_Stats.AddValue(a_Ach));
+		return static_cast<unsigned int>(m_Stats.AddValue(nameNew));
```
12xx12, line 1692: I changed this because the name lookup has already been done and the AddValue with the enum value (a_Ach) simply performs another lookup

src/Entities/Player.cpp
```diff
@@ -1673,35 +1673,36 @@ void cPlayer::SetIP(const AString & a_IP)
...
 			cCompositeChat Msg;
 			Msg.SetMessageType(mtSuccess);
-			Msg.AddShowAchievementPart(GetName(), cStatInfo::GetName(a_Ach));
+			Msg.AddShowAchievementPart(GetName(), nameNew);
 			m_World->BroadcastChat(Msg);
 		}
 
 		// Increment the statistic
-		StatValue New = m_Stats.AddValue(a_Ach);
+		StatValue New = m_Stats.AddValue(nameNew);
```
12xx12, line 1705: I changed this because the name lookup has already been done and the AddValue with the enum value (a_Ach) simply performs another lookup

src/Statistics.h
```diff
@@ -133,29 +25,43 @@ class cStatManager
...
 
-	// TODO 10-05-2014 xdot: Use, mine, craft statistics
+	/** returns the count of dirty statistics for the given map aka version */
+	int GetDirtyStatistics(const std::map<AString, int> & map) const;
 
+	/** returns the count of dirty statistics for the given map aka version */
+	int GetDirtyStatistics(const AStringMap & map) const;
 
+private:
+	std::map<AString, int> m_MainStats= {{}};
```
tigerw, line 64: We should not use strings or stringmaps in the internal state. The list of statistics are static data and shouldn't impose a runtime allocation penalty.

src/Statistics.h
```diff
@@ -1,117 +1,9 @@
...
-
-
-
 #pragma once
 
-
-
-
-// tolua_begin
-enum eStatistic
```
tigerw, line 13: Should keep this enum here; it's exported to Lua. Instead, assign each line to its equivalent in the new `enum class Statistic` enum.

src/Statistics.cpp
```diff
@@ -6,199 +6,139 @@
...
 
 
 
 
+int cStatManager::GetDirtyStatistics(const AStringMap & map) const
+{
+	int res = 0;
+	for (auto e = map.begin(); e != map.end(); e++)
+	{
+		if (GetValue(e->first) != 0)
```
tigerw, line 138: Iterating a map isn't the most performant. How about storing another static array of bools in cStatManager, which is set to true each time a statistic is updated?

src/Statistics.cpp
```diff
@@ -6,199 +6,139 @@
...
 
-	m_MainStats[a_Stat] += a_Delta;
 
-	return m_MainStats[a_Stat];
+
+
+
+const AString & cStatManager::GetPrerequisite(const AString & a_Type)
+{
+	return v1_8AchievementPrerequisites.at(a_Type);
```
tigerw, line 109: Protocol-specifics (v1_8) shouldn't bleed into the rest of the server

src/Registries/Statistics.h
```diff
@@ -0,0 +1,833 @@
...
+	{"minecraft.interact_with_furnace",         47},
+	{"minecraft.interact_with_crafting_table",  48},
+	{"minecraft.open_chest",                    49},
+	{"minecraft.sleep_in_bed",                  50},
+	{"minecraft.open_shulker_box",              51}
+};
+
+// V 1.13.1 stuff
+
+const std::map<AString, int> v1_13_1StatLookup =
```
tigerw, line 698: This should be a `switch` statement from `enum class Statistic` to Int32/UInt32 (depending on what the protocol uses). Move the protocol-specific mappings into the namespaces in Protocol/Palettes/Palette_1_1X.

tigerw: The compiler will optimise the switch much better than maps

src/Registries/Statistics.h
```diff
@@ -0,0 +1,833 @@
...
+	achKillWither,        /* The Beginning.       */
+	achFullBeacon,        /* Beaconator           */
+	achBreedCow,          /* Repopulation         */
+	achThrowDiamonds,     /* Diamonds to you!     */
+
+	statCount
+};
+
+/** Enum to describe the different types possible */
+enum statisticType
```
tigerw, line 265: `enum class StatisticType`

src/Registries/Statistics.h
```diff
@@ -0,0 +1,833 @@
...
+{
+	statMined,
```
tigerw, line 267: No need to qualify with `stat`

src/Statistics.cpp
12xx12: this is simply bad naming from me - any version supported 1.8 - 1.11 uses this table because there weren't any changes

src/Protocol/Protocol_1_9.cpp
```diff
@@ -396,6 +396,33 @@ void cProtocol_1_9_0::SendPlayerSpawn(const cPlayer & a_Player)
...
 
+void cProtocol_1_9_0::SendStatistics(const cStatManager & a_Manager)
+{
+	ASSERT(m_State == 3);  // In game mode?
+	int count = a_Manager.GetDirtyStatistics(v1_9StatLookup);
+	// If there are no dirty statistics don't send
+	if (!count)
+	{
+		return;
+	}
```
tigerw, line 407: If GetDirtyStatistics returns the array mapping statistic to dirtiness, then `count` can be something like `std::count(..., ..., true)` and the actual value-writing part a loop over the array, that does a lookup for each dirty statistic value

tigerw:
> I do have a few more questions:
> 
> 
> should global variables have a prefix "g_"?
> 
> 
> you don't allow underscores in variable names. I am not quite sure how to label the global lookup tables because I try to find obvious names and using the number of the network protocol seems not that obvious. Or may I use the same layout as the protocol names? (With underscores)

See the existing naming for stuff in Registries/ and Protocol/Palettes

src/Statistics.cpp
tigerw: Prerequisites sound like they shouldn't be protocol-specific anyway. So this function should return an `enum class Achievement` and get the prerequisite data from the latest 1.16

12xx12: Achievements were remved in 1.12. and replaced with advancements. Up to 1.12 the Achievements were handled like stats in the network

I put this in case mojang added/rermoved something

src/Registries/Statistics.h
12xx12: good to know - conversion in the way

src/Statistics.h
12xx12: does the export care if i change something?

the normal (non class) enum doesn't help me because i can't convert to the Statistics enum

src/Protocol/Protocol_1_9.cpp
12xx12: the thing is not every value that is dirty might be sent by every protocol. We have to check for every protocol if the dirty stats are the ones to be sent

src/Statistics.h
12xx12: Ok - had a look in the binding - it exports the updated version

Note for myself: I changed some names to match the ones used by Minecraft - gotta add the to the lookup to any call using the enum

src/Registries/Statistics.h
12xx12: done

src/Statistics.cpp
12xx12: renamed the map and used switch/case

src/Registries/Statistics.h
12xx12: replaced any map possible with switch/case only one is left. but is only used once if the player joins the game and the file is read.

did some research but there seems no way to use something else because switch/case don't support strings in c++ (which makes sense) and if/else has an O(n) in worst case and the map lookup O(log n)

12xx12: The other way is to replace the string descriptions in the json with just numbers which would be faster but hurt the readability quite bad

src/Registries/Statistics.h
12xx12: done - removed

src/Statistics.cpp
12xx12: i implemented that map but i don't think this brough us any performance improvements - still got to check every value if it's dirty AND if it's send by this version

src/Statistics.h
12xx12: I put the enum back again. added some conversion that you may still use the values from the old enum and added the new ones to the API too

src/Statistics.h
12xx12: The array used is now directly created for each Object and only is map<enum class, int>

src/Protocol/Protocol_1_9.cpp
tigerw: Does the client handle us sending it an empty string/ID of -1?

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1420,18 +1420,26 @@ void cProtocol_1_8_0::SendSpawnMob(const cMonster & a_Mob)
 void cProtocol_1_8_0::SendStatistics(const cStatManager & a_Manager)
 {
 	ASSERT(m_State == 3);  // In game mode?
-
+	std::vector<Statistic> send = {};
```
tigerw, line 1423: We're allocating a vector then throwing it away almost immediately; there's gotta be a better way.

If we went with GetDirtyStatistics returning the dirty array, then the first pass could unset the ones which our protocol don't support, and count the rest. We then write the size, and the second pass would write the data for the dirties remaining.

Also on style: prefer captialised `Send`

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1420,18 +1420,26 @@ void cProtocol_1_8_0::SendSpawnMob(const cMonster & a_Mob)
...
+	for (auto e = a_Manager.GetDirtyStatistics()->begin(); e != a_Manager.GetDirtyStatistics()->end(); e++)
```
tigerw, line 1424: Prefer ranged for loops, for (const auto Item : DirtyStats())

src/Statistics.h
```diff
@@ -1,21 +1,169 @@
...
+
+		Greetings 12xx12 */
+
+enum class Statistic
+{
+	/*  Please do not change order till told it is allowed
+		this is used to map the old API to the new one */
 
+	/* Achievements */
+	achOpenInventory,     /* Taking Inventory     */
```
tigerw, line 40: Couldn't we leave this as the old `enum eStatistics`, and for each old value, do something like

```cpp
enum eStatistics
{
    achOpenInventory = Achievement::OpenInventory,
}
```

tigerw: This enum should also be in Registries/Statistics.h

12xx12: the old achievements ARE statistics and the "new" eStatistics is the old one just moved

gonna move them

src/Statistics.h
```diff
@@ -1,21 +1,169 @@
...
+	achBookcase,          /* Librarian            */
+	achExploreAllBiomes,  /* Adventuring Time     */
+	achSpawnWither,       /* The Beginning?       */
+	achKillWither,        /* The Beginning.       */
+	achFullBeacon,        /* Beaconator           */
+	achBreedCow,          /* Repopulation         */
+	achDiamondsToYou,     /* Diamonds to you!     */
 
+	/* Statistics */
+	leaveGame,
```
tigerw, line 75: Style: `LeaveGame` etc

src/Statistics.h
```diff
@@ -133,30 +261,267 @@ class cStatManager
...
+	/** returns the prerequisite for a achievement */
+	static const Statistic GetPrerequisite(const Statistic a_Type);
 
+	/** returns the map of dirty statistics */
+	const std::map<Statistic, bool> * GetDirtyStatistics() const;
 
+private:
+	static const std::map<AString, Statistic> m_StatNames;
+
+	std::map<Statistic, int> m_MainStats=
```
tigerw, line 303: Could we not use the fixed size array (`int m_MainStats[Statistic::Count]`) that was there before?

tigerw:
Thanks for your work thus far!

src/Statistics.h
12xx12: And tried this but CLion told me: 
"Integral constant expression must have integral or unscoped enumeration type, not 'Statistic'"

src/Statistics.h
12xx12: Long live regular expressions

src/Statistics.h
12xx12: yes, we could do this - should improve the time

12xx12:
You are welcome - I started coding my own server in c++ due to the bad performance of the java based servers

The I hit google - found you and here I am

12xx12:
I hope I get used to the style. until now i just did stuff the google-style

src/Statistics.h
tigerw: hm. perhaps we need to throw the entire keyboard at the problem.

`achOpenInventory = static_cast<std::underlying_type_t<eStatistics>>(Statistic::OpenInventory)`

(though `achOpenInventory = static_cast<int>(Statistic::OpenInventory)` would probably work)

src/Protocol/Protocol_1_8.cpp
12xx12: Your idea is quite good

Style is gonna be changed

src/Statistics.h
12xx12: at the moment the enum and the enum class are ordered the same way so you do simply do static_cast<enum class>(enum) to "convert"

tigerw:
Yeah, about the style, and I quote:

> We can't use clang-format for format checking because Cuberite's code style is sufficiently weird that it doesn't support it
> https://github.com/cuberite/cuberite/pull/4738#issuecomment-628295299

Our style guide isn't a document to be followed, it's a _presence of mind_ xD

src/Statistics.h
tigerw: thanks! also, if the count of stats isn't going to go negative, consider making it `unsigned`

src/Statistics.h
tigerw: I see, I guess that could work then. My only concern is that if we autogenerate the list of new stats and the Lua script sorts them, we might miss any break in the ordering during code review

src/Registries/Statistics.h
tigerw: Agreed.

src/Statistics.cpp
tigerw: Sure, we do the same amount of work in the loop, however maps don't store their data in a contiguous block like an array does, so doing a linear iteration over all the elements in a map has a higher inherent cost

src/Protocol/Protocol_1_9.cpp
12xx12: I just checked - in any case this is the marker not to send.

and in returning the empty string/-1 the statistic isn't queued for sending

src/Statistics.h
12xx12: the only thing i generate at the moment is the lookup statistic -> id to send. new statistics have to be added manually.

i wrote some instructions in the statistics.h how to do this ( and some instructions what not to do.

At some point I manually ordered put some numeric values to the entries then went back to the same order

and for the lookup the order doesn't matter - i sort them by the numeric return value. looks nicer

but I just removed I this from the API because it wasn't neiter documented - see latest comit

12xx12: and if we try to autogenerate all the lookups/enums/... we loose compability for old versions (removed statistics...)

src/Protocol/Protocol_1_8.cpp
12xx12: Oh - I just noticed - since the a_Manager is constant in the transmitting we can't write into the map :/

src/Protocol/Protocol_1_8.cpp
12xx12: this improves readability quite a bit. I Knew that you could do this in java but this also works for c++ - yeah 👍 

done

src/Statistics.h
12xx12: changed the dirty statistics too

src/Statistics.h
12xx12: Done

src/Statistics.h
12xx12: moved enum to registries/statistics.h

src/Statistics.cpp
12xx12: i see what you meen. as we use a C-Style Array now this should not be a concern

src/Statistics.h
12xx12: see later on - now using somethin else

src/Statistics.h
12xx12: see later comments on removing it totaly of the API

12xx12:
The most annoying thing is that some settings file tells my IDE to hard warp at 80 symbols and then the lua checkstyle moans because of wrong indentation...

tigerw:
Thanks!

By the way, I'm wondering about splitting Statistics and Achievements/Advancements from each other (in a future PR). From my understanding they're adding a lot of extra stuff to achievements, so keeping them together with stats might cause us extensibility issues. What do you think?

12xx12:
Attention! There is a difference between Achievements <1.12 and Advancements 1.12+

Achievements is legacy content and removed in 1.11->1.12 and Advancements is the new stuff introduced in 1.12

Achievements are already implemented and not changed in the server side for all supported version.

Advancements are not implemented at all so the PR would be "implement Advancements"

This would include storing, transmitting and detecting the progress

12xx12:
One thing we could do is that we remove move some stuff around so that we can iterate over the statistics for newer versions and for statistics and achievements for old version which would improve the time for newer version by about 32/108

12xx12:
I just noticed - the most of the old Achievements are transported to the new Advancements stuff. I have to check if all the old Achievements are implemented (I don't think so) so we may actually use a new advancement object to ask for the old achievements as a enum class Statistic::: object

tigerw:
Thanks again, imma apply some changes direct to this PR just to avoid pestering you more and hopefully merge.

tigerw:
doing it nao

tigerw:
Done. I changed the static array back to a map because on second thoughts most stats are unused, we don't want to carry out kilobytes of zeroes. Also migrated the serialisation format to 1.16 with namespaced IDs. Any final comments?

12xx12:
Is this loading the old stats json file? I wrote my version in that way that it would read that also the "old" version

tigerw:
Not right now, when the storage format was upgraded to 1.13+ `minecraft:underscore_format` the old `stat.camelCase` got too hard to cleanly upgrade.

I'm torn on this, on one hand is it even worth maintaining a migration mapping given statistics wasn't even working for anyone above 1.8? But on the other hand, maybe there's a use-case for it?

12xx12:
Statistics should now work for everyone. That was the point

The only thing you would loose is the achievements

For the other ones it's more or less a funny thing

tigerw:
Protocol works for everyone definitely.

You would like to see the old json be migrated too?

12xx12:
I personally don't care for achievements and I don't know for what cuberite is used.

This is your (the teams) decision if you want to keep those for the users

tigerw:
I guess I can add an upgrade mapping, would be cool to see those collected-but-not-displayed statistics spring to life.

12xx12:
So for me you can merge - i'm gonna implement the statistics collection then

tigerw:
sorry I've been a bit stuck recently, went down a couple of Cuberite-rabbitholes. working on this right now

12xx12:
No problem - just wanted to tell you - no stress. This is just fun

tigerw:
final kerments?

12xx12:
No, I'm fine

PR #4791, opened 2020-07-29
added HandleCraftItem call to ShiftClickedResult to make sure achievements are awarded

12xx12:
I was doing some changes at the achievements system as I noticed that crafting based achievements are not awarded if you shift click the crafting process

fixed it

### DIFF
--- a/src/UI/SlotArea.cpp
+++ b/src/UI/SlotArea.cpp
@@ -692,6 +692,9 @@ void cSlotAreaCrafting::ShiftClickedResult(cPlayer & a_Player)
 		// Broadcast the window, we sometimes move items to different locations than Vanilla, causing needless desyncs:
 		m_ParentWindow.BroadcastWholeWindow();
 
+		// added this to make sure that achievements are awarded
+		HandleCraftItem(Result, a_Player);
+
 		// If the recipe has changed, bail out:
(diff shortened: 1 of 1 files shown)


12xx12:
Note: I didn't test this - time to sleep for today

12xx12:
Jup, works now

src/UI/SlotArea.cpp
```diff
@@ -692,6 +692,9 @@ void cSlotAreaCrafting::ShiftClickedResult(cPlayer & a_Player)
 		// Broadcast the window, we sometimes move items to different locations than Vanilla, causing needless desyncs:
 		m_ParentWindow.BroadcastWholeWindow();
 
+		// added this to make sure that achievements are awarded
```
madmaxoft, line 695: The comment could use some refactoring. "Added this" has no real meaning in a comment, we don't want to track some situational development, we want to know what the code is doing.
A simple `// Award achievements:` would suffice in this case.

12xx12: @madmaxoft done

madmaxoft: Sorry for being a nit-picker, the comment should start with a capital letter.

PR #4799, opened 2020-08-02
Changed large stack-based allocations to heap-based.

### LINKED ISSUE #3406: Cuberites crashes on generating world on alpine linux
Server OS:        alpine linux
Commit id:        (be916bc7e85843291c5ab8cb4f268bd1daa57b76)
### Actual behavior

Cuberite crashes on generating the world. The stack trace makes no sense.

```
#0  0x00000dac149fe55c in cChunkDef::SetBiome (a_BiomeMap=..., a_X=0, a_Z=14, a_Biome=biTaigaHills)
    at /root/cuberite/src/ChunkDef.h:251
#1  0x00000dac14d5fef6 in cBioGenGrown::GenBiomes (this=0xdac1a3175a0, a_ChunkX=0, a_ChunkZ=-1, a_Biomes=...)
    at /root/cuberite/src/Generating/BioGen.cpp:1025
#2  0x00000dac14d55b1c in cBioGenCache::GenBiomes (this=0xdac1a51b8a0, a_ChunkX=0, a_ChunkZ=-1, a_BiomeMap=...)
    at /root/cuberite/src/Generating/BioGen.cpp:122
#3  0x00000dac14d55e71 in cBioGenMulticache::GenBiomes (this=0xdac1a3185d0, a_ChunkX=0, a_ChunkZ=-1, a_BiomeMap=...)
    at /root/cuberite/src/Generating/BioGen.cpp:171
#4  0x00000dac14d25fe0 in cBiomalNoise3DComposable::CalcBiomeParamArrays (this=0x74819dbe7030, a_ChunkX=0,
    a_ChunkZ=0, a_HeightAmp=..., a_MidPoint=...) at /root/cuberite/src/Generating/Noise3DGenerator.cpp:689
#5  0x00000dac14d258fd in cBiomalNoise3DComposable::GenerateNoiseArrayIfNeeded (this=0x74819dbe7030, a_ChunkX=0,
    a_ChunkZ=0) at /root/cuberite/src/Generating/Noise3DGenerator.cpp:629
#6  0x00000dac14d26c42 in cBiomalNoise3DComposable::GenShape (this=0x74819dbe7030, a_ChunkX=0, a_ChunkZ=0,
    a_Shape=...) at /root/cuberite/src/Generating/Noise3DGenerator.cpp:811
#7  0x00000dac14cd1739 in cComposableGenerator::DoGenerate (this=0xdac1a317520, a_ChunkX=0, a_ChunkZ=0,
    a_ChunkDesc=...) at /root/cuberite/src/Generating/ComposableGenerator.cpp:161
#8  0x00000dac14ccfd50 in cChunkGenerator::DoGenerate (this=0x74819cd3f008, a_ChunkX=0, a_ChunkZ=0)
    at /root/cuberite/src/Generating/ChunkGenerator.cpp:295
#9  0x00000dac14ccfac9 in cChunkGenerator::Execute (this=0x74819cd3f008)
    at /root/cuberite/src/Generating/ChunkGenerator.cpp:275
#10 0x00000dac14b0b8d7 in cIsThread::DoExecute (this=0x74819cd3f008) at /root/cuberite/src/OSSupport/IsThread.cpp:74
#11 0x00000dac14b0d0eb in std::_Mem_fn_base<void (cIsThread::*)(), true>::operator()<, void>(cIsThread*) const (
    this=0xdac1a5b8530, __object=0x74819cd3f008) at /usr/include/c++/5.3.0/functional:600
#12 0x00000dac14b0d059 in std::_Bind_simple<std::_Mem_fn<void (cIsThread::*)()> (cIsThread*)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0xdac1a5b8528) at /usr/include/c++/5.3.0/functional:1531
#13 0x00000dac14b0cf60 in std::_Bind_simple<std::_Mem_fn<void (cIsThread::*)()> (cIsThread*)>::operator()() (
    this=0xdac1a5b8528) at /usr/include/c++/5.3.0/functional:1520
#14 0x00000dac14b0cef0 in std::thread::_Impl<std::_Bind_simple<std::_Mem_fn<void (cIsThread::*)()> (cIsThread*)> >::_M_run() (this=0xdac1a5b8510) at /usr/include/c++/5.3.0/thread:115
#15 0x000074819d787c8a in execute_native_thread_routine () from /usr/lib/libstdc++.so.6
#16 0x000074819da6d4fc in ?? () from /lib/ld-musl-x86_64.so.1
#17 0x0000000000000000 in ?? ()

```

Compiled cuberite with gcc/g++ without any problems. Maybe there are some problems with the use of musl, linking problems or missing flags.

madmaxoft:
Now Cuberite can run with 64 KiB thread stack.

I'm not too happy about the wrappers, but I could not find any other solution. `unique_ptr` just doesn't play nice with `typedef`-ed arrays. Perhaps @peterbell10 could come up with something more elegant?

Fixes #3406.

### DIFF
--- a/src/Generating/ComposableGenerator.cpp
+++ b/src/Generating/ComposableGenerator.cpp
@@ -155,27 +155,33 @@ void cComposableGenerator::GenerateBiomes(cChunkCoords a_ChunkCoords, cChunkDef:
 
 void cComposableGenerator::Generate(cChunkDesc & a_ChunkDesc)
 {
+	// Wrap a cChunkDesc::Shape into a structure so that we can make a unique_ptr out of it:
+	struct ShapeHelper
+	{
+		cChunkDesc::Shape mShape;
+	};
+	auto shape = std::make_unique<ShapeHelper>();
+
 	if (a_ChunkDesc.IsUsingDefaultBiomes())
 	{
 		m_BiomeGen->GenBiomes(a_ChunkDesc.GetChunkCoords(), a_ChunkDesc.GetBiomeMap());
 	}
 
-	cChunkDesc::Shape shape;
 	if (a_ChunkDesc.IsUsingDefaultHeight())
 	{
-		m_ShapeGen->GenShape(a_ChunkDesc.GetChunkCoords(), shape);
-		a_ChunkDesc.SetHeightFromShape(shape);
+		m_ShapeGen->GenShape(a_ChunkDesc.GetChunkCoords(), shape->mShape);
+		a_ChunkDesc.SetHeightFromShape(shape->mShape);
 	}
 	else
 	{
 		// Convert the heightmap in a_ChunkDesc into shape:
-		a_ChunkDesc.GetShapeFromHeight(shape);
+		a_ChunkDesc.GetShapeFromHeight(shape->mShape);
 	}
 
 	bool ShouldUpdateHeightmap = false;
 	if (a_ChunkDesc.IsUsingDefaultComposition())
 	{
-		m_CompositionGen->ComposeTerrain(a_ChunkDesc, shape);
+		m_CompositionGen->ComposeTerrain(a_ChunkDesc, shape->mShape);
 	}
 
 	if (a_ChunkDesc.IsUsingDefaultFinish())
--- a/src/Generating/ComposableGenerator.cpp
+++ b/src/Generating/ComposableGenerator.cpp
@@ -158,7 +158,7 @@ void cComposableGenerator::Generate(cChunkDesc & a_ChunkDesc)
 	// Wrap a cChunkDesc::Shape into a structure so that we can make a unique_ptr out of it:
 	struct ShapeHelper
 	{
-		cChunkDesc::Shape mShape;
+		cChunkDesc::Shape m_Shape;
 	};
 	auto shape = std::make_unique<ShapeHelper>();
 
@@ -169,19 +169,19 @@ void cComposableGenerator::Generate(cChunkDesc & a_ChunkDesc)
 
 	if (a_ChunkDesc.IsUsingDefaultHeight())
 	{
-		m_ShapeGen->GenShape(a_ChunkDesc.GetChunkCoords(), shape->mShape);
-		a_ChunkDesc.SetHeightFromShape(shape->mShape);
+		m_ShapeGen->GenShape(a_ChunkDesc.GetChunkCoords(), shape->m_Shape);
+		a_ChunkDesc.SetHeightFromShape(shape->m_Shape);
 	}
 	else
 	{
 		// Convert the heightmap in a_ChunkDesc into shape:
-		a_ChunkDesc.GetShapeFromHeight(shape->mShape);
+		a_ChunkDesc.GetShapeFromHeight(shape->m_Shape);
 	}
 
 	bool ShouldUpdateHeightmap = false;
 	if (a_ChunkDesc.IsUsingDefaultComposition())
 	{
-		m_CompositionGen->ComposeTerrain(a_ChunkDesc, shape->mShape);
+		m_CompositionGen->ComposeTerrain(a_ChunkDesc, shape->m_Shape);
 	}
 
 	if (a_ChunkDesc.IsUsingDefaultFinish())
--- a/src/Generating/StructGen.cpp
+++ b/src/Generating/StructGen.cpp
@@ -20,7 +20,14 @@ void cStructGenTrees::GenFinish(cChunkDesc & a_ChunkDesc)
 	int ChunkX = a_ChunkDesc.GetChunkX();
 	int ChunkZ = a_ChunkDesc.GetChunkZ();
 
-	cChunkDesc WorkerDesc({ChunkX, ChunkZ});
+	// Wrap the cChunkDesc::Shape into a structure, so that we can make a unique_ptr out of it:
+	struct ShapeHelper
+	{
+		cChunkDesc::Shape mShape;
+	};
+
+	auto WorkerDesc = std::make_unique<cChunkDesc>(a_ChunkDesc.GetChunkCoords());
+	auto WorkerShape = std::make_unique<ShapeHelper>();
 
 	// Generate trees:
 	for (int x = 0; x <= 2; x++)
@@ -34,16 +41,12 @@ void cStructGenTrees::GenFinish(cChunkDesc & a_ChunkDesc)
 
 			if ((x != 1) || (z != 1))
 			{
-				Dest = &WorkerDesc;
-				WorkerDesc.SetChunkCoords({BaseX, BaseZ});
-
-				// TODO: This may cause a lot of wasted calculations, instead of pulling data out of a single (cChunkDesc) cache
-
-				cChunkDesc::Shape workerShape;
-				m_BiomeGen->GenBiomes           ({BaseX, BaseZ}, WorkerDesc.GetBiomeMap());
-				m_ShapeGen->GenShape            ({BaseX, BaseZ}, workerShape);
-				WorkerDesc.SetHeightFromShape   (workerShape);
-				m_CompositionGen->ComposeTerrain(WorkerDesc, workerShape);
+				Dest = WorkerDesc.get();
+				WorkerDesc->SetChunkCoords({BaseX, BaseZ});
+				m_BiomeGen->GenBiomes           ({BaseX, BaseZ}, WorkerDesc->GetBiomeMap());
+				m_ShapeGen->GenShape            ({BaseX, BaseZ}, WorkerShape->mShape);
+				WorkerDesc->SetHeightFromShape   (WorkerShape->mShape);
+				m_CompositionGen->ComposeTerrain(*WorkerDesc, WorkerShape->mShape);
 			}
 			else
 			{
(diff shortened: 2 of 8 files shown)


madmaxoft:
The performance impact of these changes is below measurement error, I tried generating and loading a 40-chunk-radius world and the time it took was the same to the tenth of a second.

tigerw:
This works for me:
```
diff --git a/CMake/Fixups.cmake b/CMake/Fixups.cmake

+target_link_options(Cuberite PRIVATE "-Wl,-z,stack-size=1048576")
```

`Linux Alpine 5.4.43-1-lts #2-Alpine SMP Thu, 28 May 2020 20:13:48 UTC i686 Linux`

```
     [12:15:56] -- Loaded 2 Plugins --
     [12:15:57] Preparing spawn (world): 7.00% (28/400; 27.70 chunks / sec)
     [12:15:58] Preparing spawn (world): 26.50% (106/400; 73.72 chunks / sec)
     [12:15:59] Preparing spawn (world): 48.00% (192/400; 85.40 chunks / sec)
     [12:16:00] Preparing spawn (world): 68.75% (275/400; 81.05 chunks / sec)
     [12:16:01] Preparing spawn (world): 89.00% (356/400; 79.65 chunks / sec)
     [12:16:03] Preparing spawn (world): 97.00% (388/400; 379.65 chunks / sec)
Info [12:16:03] World "world":Generated spawnpoint position at {32.50, 62.00, 32.50}
```

tigerw:
also no one noticed the build fails on alpine ;-;

it doesn't like the cRoot input thread unblocker `select` thingy

madmaxoft:
I think it's healthier to avoid heavy stack usage. So I'd still prefer to merge this or a similar.

peterbell10:
The only way to have `unique_ptr` of array is to leave the extent of the array out of the template. e.g. 
```cpp
std::unique_ptr<NIBBLETYPE[]> Metas(new NIBBLTYPE[256]);
auto Blocks = std::make_unique<BLOCKTYPE[]>(256);
```

That's intended to behave like a pointer to array, not a static sized array though. So you couldn't pass it to a `const Shape &` without casting. One alternative could be to make `Shape` a `std::array`. I think they are mostly inter-compatible and `std::array` can be statically sized in a `unique_ptr`.


>also no one noticed the build fails on alpine ;-;
>it doesn't like the cRoot input thread unblocker select thingy

Probably just missing the include?

madmaxoft:
I guess I prefer the struct-wrapping solution - that way we don't have to repeat the count (and make a mistake in it) every time.

tigerw:
I would be reluctant to penalise all the major platforms offering at least 1MiB to accommodate musl's rather austere choice of 128KiB. Also since all our CI is on >1MiB we've no guarantee some future change will silently break alpine again - indeed this fix already touches two more places outside of just the generator - which is why I think a platform-specific link flag is more robust.

peterbell10:
Allocating 64 KiB of memory is orders of magnitude faster than actually accessing that memory. So, I don't see any point in complicating things by adding compilation flags. For string compression, I could see the chunks size effecting performance. Maybe worth checking the documentation and/or running an isolated benchmark?

madmaxoft:
I think a 40-chunk-radius generation and loading is good enough a benchmark - it completed within 14 seconds, and the changes here didn't cause a noticeable delay (actually the code was ~100 msec faster with them, but that's possibly a measurement error across such a huge task).

I don't think stack-based huge values should be used, I used to use them all the time but then you get a stack overflow somewhere and it's really hard to pinpoint, because the stack is not complete at that point.

src/Generating/ComposableGenerator.cpp
```diff
@@ -155,27 +155,33 @@ void cComposableGenerator::GenerateBiomes(cChunkCoords a_ChunkCoords, cChunkDef:
 
 void cComposableGenerator::Generate(cChunkDesc & a_ChunkDesc)
 {
+	// Wrap a cChunkDesc::Shape into a structure so that we can make a unique_ptr out of it:
+	struct ShapeHelper
```
tigerw, line 159: The struct should have a comment about musl and a link to #3406 for the record

src/Generating/ComposableGenerator.cpp
```diff
@@ -155,27 +155,33 @@ void cComposableGenerator::GenerateBiomes(cChunkCoords a_ChunkCoords, cChunkDef:
...
+	{
+		cChunkDesc::Shape mShape;
```
tigerw, line 161: xoft pls

just call it `Shape`

src/Generating/StructGen.cpp
```diff
@@ -34,16 +41,12 @@ void cStructGenTrees::GenFinish(cChunkDesc & a_ChunkDesc)
...
-				cChunkDesc::Shape workerShape;
-				m_BiomeGen->GenBiomes           ({BaseX, BaseZ}, WorkerDesc.GetBiomeMap());
-				m_ShapeGen->GenShape            ({BaseX, BaseZ}, workerShape);
-				WorkerDesc.SetHeightFromShape   (workerShape);
-				m_CompositionGen->ComposeTerrain(WorkerDesc, workerShape);
+				Dest = WorkerDesc.get();
+				WorkerDesc->SetChunkCoords({BaseX, BaseZ});
+				m_BiomeGen->GenBiomes           ({BaseX, BaseZ}, WorkerDesc->GetBiomeMap());
+				m_ShapeGen->GenShape            ({BaseX, BaseZ}, WorkerShape->mShape);
+				WorkerDesc->SetHeightFromShape   (WorkerShape->mShape);
```
tigerw, line 48: Is the alignment here intentional, it looks like there's an extra space

tigerw:
ok whatevs

Do you see a way of potentially splitting up the chunkdesc into smaller parts and sidestep the issue entirely?

tigerw:
Isn't string compression also used in the protocol? What was performance like for that?

src/Generating/ComposableGenerator.cpp
bearbin: Or not, it's fine...

src/Generating/StructGen.cpp
bearbin: I think it's a typo, but the alignment here is ugly and against conventional style so should probably just be removed entirely.

src/Generating/ComposableGenerator.cpp
madmaxoft: I don't think so, it's about the generic principle of keeping stuff off the stack, not about a specific issue.

src/Generating/ComposableGenerator.cpp
madmaxoft: One of these days Imma write a script that sends all those underscores into oblivion...

madmaxoft:
As for the string compression, I haven't done a benchmark, but I don't expect too much of a difference either, most of the data being compressed or decompressed is really a few kilobytes in size, so in the end we get a difference of, say, ten allocations instead of two. Considering that the actual compression or decompression is a much heavier work on those kilobytes, I wouldn't say this matters much.

peterbell10:
>Considering that the actual compression or decompression is a much heavier work on those kilobytes, I wouldn't say this matters much.

I think you misunderstand. The problem is zlib's `inflate` algorithm might be more efficient when it's decompressing more data at once. In fact, the [zlib website](https://zlib.net/zlib_how.html) suggests this is something to worry about:

>CHUNK is simply the buffer size for feeding data to and pulling data from the zlib routines. **Larger buffer sizes would be more efficient, especially for inflate().** If the memory is available, **buffers sizes on the order of 128K or 256K bytes should be used.**

madmaxoft:
I've gone through the sources to see where we use `inflate`:
- Protocol: incoming packets
- World storage: loaded chunks
- Prefabs: loaded areas
- In-game map items: loading the map contents
- Scoreboard: loading

Most of these are low-throughput operations and they operate on miniscule amounts of data. The only possible effect is in the world loader, but even there the difference is not noticeable.

peterbell10:
How about reusing the output buffer? 
```cpp
bool UsingBuffer = (a_Length <= sizeof(Buffer));
if (UsingBuffer)
{
    strm.next_out = reinterpret_cast<Bytef *>(Buffer);
    strm.avail_out = sizeof(buffer);
}
else
{
    a_Uncompressed.resize(a_Length);
    strm.next_out = reinterpret_cast<Bytef *>(a_Uncompressed.data());
    strm.avail_out = a_Uncompressed.size();
}

for (;;)
{
	// Do first decompression pass, then reset UsingBuffer to true
	if (UsingBuffer)
	{
		a_Uncompressed.append(Buffer, sizeof(Buffer) - strm.avail_out);
    	strm.avail_out = sizeof(buffer);
	}
    else
	{
		a_Uncompressed.resize(a_Uncompressed.size() - strm.avail_out);
        strm.next_out = reinterpret_cast<Bytef *>(Buffer);
		strm.avail_out = sizeof(buffer);
        UsingBuffer = true;
	}
}

```

madmaxoft:
I'm not sure the perf gain is worth the ugly code. Note that there are two places where the buffer contents are appended, so this blob of code would be double in the actual function. And as I said, I don't think the perf's a problem here.

peterbell10:
>I'm not sure the perf gain is worth the ugly code.

Well that's why we should measure.

>Note that there are two places where the buffer contents are appended, so this blob of code would be double in the actual function.

Just put it in a lambda, then call that lambda twice.

tigerw:
I'd like to once again propose my one-liner linker flag which has no overhead

madmaxoft:
You know what? Fine. I give up.

tigerw:
xoft you should be more persistent, fight for your code

I might like to do some throughput measurements (if you haven't already done so) with the proposed changes

PR #4802, opened 2020-08-02
Add correct implementation of crops

### LINKED ISSUE #4801: [BUG] Crops drops aren't like official implementation
Client version: 1.12.2
Server OS: Windows

### Expected behavior
A random value between 0 & 3 (depending on the level of the seed) should be dropped (See https://minecraft.gamepedia.com/Pumpkin_Seeds#Breaking)

### Actual behavior
When destroying crops, only one seed is dropped for melon & pumpkin

- Plant melon seed
- Break melon seed in Survival Mode
- Same with pumpkin

### Server log
```
/
```

### LINKED ISSUE #4803: [BUG] Bone Meal in Dispenser changes crops from age 0 to age 7
Client version: 1.12.2
Server OS: Windows

### Expected behavior
- It should goes from 0 to 2 - 4 and not directly to 7

### Actual behavior
- Crops go from age 0 to age 7

- Put Potato / Carrot / Seeds / Melon / Pumpkin on a farmland block
- Put a Dispenser that looks to the block above the farmland
- Put Bone Meal inside the Dispenser
- Power on the Dispenser

### Server log
```
/
```

### LINKED ISSUE #4805: [BUG] Right-clicking with a bone meal on a full grown BeetRoot in Survival consume 2 bone meals
Client version: 1.12.2

### Expected behavior
No Bone Meal should be consumed

### Actual behavior
2 Bone Meals are consumed

- Plant a Beetroot Seed on a farmland
- Grown it until it's full grown
- Go in survival
- Right-click one time on the Beetroot Seed with at least 2 bone meal in your inventory

### Server log
```
[dcdf76f9faaa5c36|22:38:26] HandleRightClick: {110, 62, 175}, face 1, Cursor {5, 8, 9}, Hand: 0, HeldItem: whitedye:15 * 56; Dist: 2.55
[dcdf76f9faaa5c36|22:38:26] HandleUseItem: Hand: 0; HeldItem: whitedye:15 * 55
[dcdf76f9faaa5c36|22:38:26] HandleRightClick: {110, 62, 175}, face 1, Cursor {5, 8, 9}, Hand: 0, HeldItem: whitedye:15 * 55; Dist: 2.55
```

### Note
Doesn't work with carrots & potatos (Which is normal)

0ddlyoko:
This Pull Request fix some bugs linked to crops.

Some changes are:
- Crops now drops like vanilla minecraft (See [Melon](https://minecraft.gamepedia.com/Melon_Seeds#Breaking), [Pumpkin](https://minecraft.gamepedia.com/Pumpkin_Seeds#Breaking), [Carrots, Potatoes & Wheat](https://minecraft.gamepedia.com/Fortune#Seeds), [Wheat](https://minecraft.fandom.com/wiki/Seeds_(Wheat))
- Fortune works on crops
- Add method `IsFullGrownPlantAt(Vector3i)` in classes **cChunk**, **cChunkMap**, **cWorld**
- Add method `IsFullGrown(cChunk &, Vector3i)` in class **cBlockHandler**
- These methods will check if crop at a given location is full grown
- No longer create a melon / pumpkin when you right-click with a bone meal on a full grown melon / pumpkin seed
- Fixes #4801 
- Fixes #4803
- Fixes #4805 

Futur (tomorrow)
-

### DIFF
--- a/src/BlockEntities/DispenserEntity.cpp
+++ b/src/BlockEntities/DispenserEntity.cpp
@@ -10,6 +10,7 @@
 #include "../Entities/ProjectileEntity.h"
 #include "../Simulator/FluidSimulator.h"
 #include "../Items/ItemSpawnEgg.h"
+#include "../Items/ItemDye.h"
 
 
 
@@ -211,7 +212,8 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
 				DropFromSlot(a_Chunk, a_SlotNum);
 				break;
 			}
-			if (m_World->GrowRipePlant(dispAbsCoord.x, dispAbsCoord.y, dispAbsCoord.z, true))
+			// Simulate a right-click with bonemeal
+			if (cItemDyeHandler::FertilizePlant(*m_World, {dispAbsCoord.x, dispAbsCoord.y, dispAbsCoord.z}))
 			{
 				m_Contents.ChangeSlotCount(a_SlotNum, -1);
 			}
--- a/src/BlockEntities/DispenserEntity.cpp
+++ b/src/BlockEntities/DispenserEntity.cpp
@@ -212,8 +212,9 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
 				DropFromSlot(a_Chunk, a_SlotNum);
 				break;
 			}
-			// Simulate a right-click with bonemeal
-			if (cItemDyeHandler::FertilizePlant(*m_World, {dispAbsCoord.x, dispAbsCoord.y, dispAbsCoord.z}))
+
+			// Simulate a right-click with bonemeal:
+			if (cItemDyeHandler::FertilizePlant(*m_World, dispAbsCoord))
 			{
 				m_Contents.ChangeSlotCount(a_SlotNum, -1);
 			}
--- a/src/BlockEntities/DispenserEntity.cpp
+++ b/src/BlockEntities/DispenserEntity.cpp
@@ -26,24 +26,24 @@ cDispenserEntity::cDispenserEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta
 
 void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
 {
-	Vector3i dispRelCoord(GetRelPos());
-	auto meta = a_Chunk.GetMeta(dispRelCoord);
-	AddDropSpenserDir(dispRelCoord, meta);
-	auto dispChunk = a_Chunk.GetRelNeighborChunkAdjustCoords(dispRelCoord);
-	if (dispChunk == nullptr)
+	Vector3i DispRelCoord(GetRelPos());
+	auto Meta = a_Chunk.GetMeta(DispRelCoord);
+	AddDropSpenserDir(DispRelCoord, Meta);
+	auto DispChunk = a_Chunk.GetRelNeighborChunkAdjustCoords(DispRelCoord);
+	if (DispChunk == nullptr)
 	{
 		// Would dispense into / interact with a non-loaded chunk, ignore the tick
 		return;
 	}
 
-	BLOCKTYPE dispBlock = dispChunk->GetBlock(dispRelCoord);
-	auto dispAbsCoord = dispChunk->RelativeToAbsolute(dispRelCoord);
+	BLOCKTYPE DispBlock = DispChunk->GetBlock(DispRelCoord);
+	auto DispAbsCoord = DispChunk->RelativeToAbsolute(DispRelCoord);
 
 	// Dispense the item:
 	const cItem & SlotItem = m_Contents.GetSlot(a_SlotNum);
-	if (ItemCategory::IsMinecart(SlotItem.m_ItemType) && IsBlockRail(dispBlock))  // only actually place the minecart if there are rails!
+	if (ItemCategory::IsMinecart(SlotItem.m_ItemType) && IsBlockRail(DispBlock))  // only actually place the minecart if there are rails!
 	{
-		if (m_World->SpawnMinecart(dispAbsCoord.x + 0.5, dispAbsCoord.y + 0.5, dispAbsCoord.z + 0.5, SlotItem.m_ItemType) != cEntity::INVALID_ID)
+		if (m_World->SpawnMinecart(DispAbsCoord.x + 0.5, DispAbsCoord.y + 0.5, DispAbsCoord.z + 0.5, SlotItem.m_ItemType) != cEntity::INVALID_ID)
 		{
 			m_Contents.ChangeSlotCount(a_SlotNum, -1);
 		}
@@ -53,15 +53,15 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
 	{
 		case E_ITEM_BUCKET:
 		{
-			LOGD("Dispensing empty bucket in slot %d; dispBlock is \"%s\" (%d).", a_SlotNum, ItemTypeToString(dispBlock).c_str(), dispBlock);
-			switch (dispBlock)
+			LOGD("Dispensing empty bucket in slot %d; DispBlock is \"%s\" (%d).", a_SlotNum, ItemTypeToString(DispBlock).c_str(), DispBlock);
+			switch (DispBlock)
 			{
 				case E_BLOCK_STATIONARY_WATER:
 				case E_BLOCK_WATER:
 				{
 					if (ScoopUpLiquid(a_SlotNum, E_ITEM_WATER_BUCKET))
 					{
-						dispChunk->SetBlock(dispRelCoord, E_BLOCK_AIR, 0);
+						DispChunk->SetBlock(DispRelCoord, E_BLOCK_AIR, 0);
 					}
 					break;
 				}
@@ -70,7 +70,7 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
 				{
 					if (ScoopUpLiquid(a_SlotNum, E_ITEM_LAVA_BUCKET))
 					{
-						dispChunk->SetBlock(dispRelCoord, E_BLOCK_AIR, 0);
+						DispChunk->SetBlock(DispRelCoord, E_BLOCK_AIR, 0);
 					}
 					break;
 				}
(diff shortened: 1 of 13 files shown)


0ddlyoko:
@tigerw 
I think I'm ok with this PR, could you check it and say me if there is any problem with it or any improvement ?

src/Blocks/BlockCrops.h
```diff
@@ -122,8 +149,30 @@ class cBlockCropsHandler:
...
+private:
+
+	/**
+		Generate a random number using a binomial distribution
+		See https://minecraft.gamepedia.com/Fortune#Seeds
+	*/
+	int GenerateAmountSeedDrop(int a_Min, int a_Count, double a_Probability, MTRand &a_Rand)
+	{
+		int SeedCount = a_Min;
+		for (int i(0); i < a_Count; i++)
```
bearbin, line 166:
```suggestion
		for (int i = 0; i < a_Count; i++)
```

This method of initialisation is odd and unusual, for consistency of style please use the normal method.

src/Blocks/BlockCrops.h
```diff
@@ -46,33 +47,49 @@ class cBlockCropsHandler:
...
+		{
+			fortune = a_Tool->m_Enchantments.GetLevel(cEnchantments::eEnchantment::enchFortune);
+		}
+
 		switch (m_BlockType)
 		{
 			case E_BLOCK_BEETROOTS:
 			{
-				char SeedCount = 1 + ((rand.RandInt<char>(2) + rand.RandInt<char>(2)) / 2);  // [1 .. 3] with high preference of 2
+				// TODO Check Fortune Level
```
bearbin, line 61: What do you mean by check fortune level? The values here match the wiki, so the comment should either be removed or you should do any additional checks you need to do.

I don't want to introduce more TODOs where it's avoidable like this.

src/Blocks/BlockStems.h
```diff
@@ -58,25 +83,24 @@ class cBlockStemsHandler:
 	virtual int Grow(cChunk & a_Chunk, Vector3i a_RelPos, int a_NumStages = 1) override
 	{
 		auto oldMeta = a_Chunk.GetMeta(a_RelPos);
-		auto meta = oldMeta + a_NumStages;
-		a_Chunk.SetBlock(a_RelPos, m_BlockType, static_cast<NIBBLETYPE>(std::min(meta, 7)));  // Update the stem
-		if (meta > 7)
-		{
-			if (growProduce(a_Chunk, a_RelPos))
```
bearbin, line 65: Why have you removed the actual growth of the fruit from this function? Where has it gone...

0ddlyoko: With the old function, when you right clicked with a bone meal on a full grown pumpkin / melon seed, a pumpkin / melon block spawn.
Here, it grows the plant to age 7 and doesn't spawn any blocks

bearbin: The problem is, that the `Grow()` function is also called on a normal block update. If it's removed here, then the fruit blocks will not spawn ever. So it needs to be put back here and a different function written for bonemeal growth that doesn't call `growProduce`.

0ddlyoko:
Done @bearbin

src/Blocks/BlockStems.h
```diff
@@ -60,18 +85,28 @@ class cBlockStemsHandler:
 		auto oldMeta = a_Chunk.GetMeta(a_RelPos);
 		auto meta = oldMeta + a_NumStages;
 		a_Chunk.SetBlock(a_RelPos, m_BlockType, static_cast<NIBBLETYPE>(std::min(meta, 7)));  // Update the stem
-		if (meta > 7)
+		if (oldMeta == 7)
```
bearbin, line 88: This change is still wrong. The Grow function was fine as it was.

src/Items/ItemDye.h
```diff
@@ -118,6 +118,11 @@ class cItemDyeHandler :
 			{
 				// Grow by 2 - 5 stages:
 				auto numStages = GetRandomProvider().RandInt(2, 5);
+				// No longer create a melon / pumpkin block when right-clicking on a grown melon / pumpkin seed
```
bearbin, line 121: We should not do nothing here, we only need to cap the numStages in the call to GrowPlantAt to be min(7-currentGrowth, numStages)

bearbin (dismissed):
Idea is good but this code will not actually work correctly.

src/Blocks/BlockStems.h
0ddlyoko: I don't really understand what can I do if I don't have to edit this method, here I've tested this method and it works

bearbin: I'll rework this myself and then merge - but can you explain what your reason was for changing this function in the first place?

0ddlyoko: With the old implementation, when you right-click on a melon / pumpkin seed with level = 6, the level of the plant become 7 (full grown) and a melon / pumpkin block appears.
This is because this method checks when level will become > 7.
With this implementation, if old level is before 7 (so 6, 5, ...), no melon / pumpkin block will spawn.

src/Items/ItemDye.h
0ddlyoko: This is to prevent creating a melon / pumpkin block when someone right-click on a full grown plant with bone meal.
This prevents also dispenser to do it.

src/Blocks/BlockStems.h
0ddlyoko: Another way is to create a special method when a player / dispenser want to grow melon / pumpkin or add a new parameter to this function, but after some tests it seams this solution works

tigerw:
To summarise (please correct any mistakes): our current issue is that bonemeal causes crops both to grow stages, and spawn yummy melon. However, Vanilla doesn't spawn melon when using bonemeal; the crop only advances stages, up to a maximum of 7.

In this case, as you said, I think we should define `Grow` as only making the stem higher. Then, add a new function in `cBlockPlant`, `BearFruit`, that is called when a block tick happens in `OnUpdate`. `cBlockStemsHandler` then overrides `BearFruit` to spawn melon, which now only happens during a random tick.

This also has the advantage of not needing to add IsFullyGrown to cWorld (we're trying to reduce cWorld's size) xD

So something like:

```cpp
// Plant

cBlockPlantHandler::OnUpdate()
{
  if (Grow() == 0)
  {
     // Growth at max, spawn any fruit (pumpkin, melons):
     BearFruit();
  }
}

// Stems

cBlockStemsHandler::Grow(const int Stages)
{
  const auto Meta = std::min(7, OldMeta + Stages);
  SetMeta(Meta);
  return Meta - OldMeta;
}

cBlockStemsHandler::BearFruit()
{
  // Whatever growProduce contained
}
```

D'you think this will work?

bearbin:
Tigerw's suggestion seems reasonable, and also has the advantage of being extensible to other new types of fruit that might be added.

0ddlyoko:
Same for me, I'll implement your solution when I have time

tigerw:
Hey @0ddlyoko are you currently working on this? If you're busy I could apply the changes and merge.

KingCol13:
@tigerw I've got the fortune stuff implemented in a branch ready to pull after I rebase it on top of your style fixes.

0ddlyoko:
Hello @tigerw I'm not working on it right now, you can merge if you want or you can apply changes & merge

tigerw:
Thanks all!

12xx12:
So think I Applied those changes. **But** we still need a function that checks if the plant is fully grown to prevent bonemeal from being used when the plant is already fully grown

tigerw:
Stopping bonemeal being used is just Grow returning 0 right?

12xx12:
Yeah but the code tries throwing the dice if it should grow the plant first and then tries to grow the plant. So if the code decides not to try to grow the plant a bonemeal is still used even if the plant is fully grown

12xx12:
We could move the random condition to the plants themselves but this might impact maintainability because the code is distributed all over the palnts. Now it's nicely all in one place

tigerw:
ItemDye.h right? The equipped item is only consumed if FertilizePlant returns true, but it doesn't if Grow returns <=0 (plant at max growth).

12xx12:
Yes, but if the random check fails the ```FertilizePlant``` function also returns true

tigerw:
ah! For beetroots? We could just say `GrowPlantAt(Random.RandomBool(0.75) ? 1 : 0)`.

tigerw:
No that doesn't work, beetroots can consume bonemeal but not grow.

tigerw:
This could work:

```cpp
if (a_World.GrowPlantAt(a_BlockPos, 1) <= 0)
{
	// Fix #4805
	return false;
}

// Always play effect:
a_World.BroadcastSoundParticleEffect(EffectID::PARTICLE_HAPPY_VILLAGER, a_BlockPos, 0);

// Growth falls back 25% of the time:
if (GetRandomProvider().RandBool(0.25))
{
	a_World.GrowPlantAt(a_BlockPos, -1);
}
```

12xx12:
This could work yeah. But growing and ungrowing the plant seems a bit weird

tigerw:
We can say it's due to crop blight or pests...

12xx12:
:D

I personally don't care.
Do you want to wait for another voice?

tigerw:
Sure, any jections of the ob, @cuberite/core-team?

peterbell10 (dismissed):

src/ChunkMap.cpp
```diff
@@ -1624,6 +1624,23 @@ int cChunkMap::GrowPlantAt(Vector3i a_BlockPos, int a_NumStages)
 
 
 
+bool cChunkMap::IsFullGrownPlantAt(Vector3i a_BlockPos)
```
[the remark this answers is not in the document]
12xx12: We are trying to get rid of this function atm

tigerw (approved):

src/BlockEntities/DispenserEntity.cpp
```diff
@@ -211,7 +211,9 @@ void cDispenserEntity::DropSpenseFromSlot(cChunk & a_Chunk, int a_SlotNum)
 				DropFromSlot(a_Chunk, a_SlotNum);
 				break;
 			}
-			if (m_World->GrowRipePlant(dispAbsCoord.x, dispAbsCoord.y, dispAbsCoord.z, true))
+
+			// Simulate a right-click with bonemeal:
+			if (m_World->GrowRipePlant(dispAbsCoord))
```
12xx12, line 216: Does this actually create the intended behaviour

tigerw: Fixed thanks

12xx12 (approved):

src/World.h
```diff
@@ -855,16 +855,6 @@ class cWorld  // tolua_export
...
-	Grows the plant at the specified block to its ripe stage.
-	a_IsByBonemeal is obsolete, do not use.
-	Returns true if grown, false if not (invalid chunk, non-growable block, already ripe). */
-	bool GrowRipePlant(int a_BlockX, int a_BlockY, int a_BlockZ, bool a_IsByBonemeal = false)
-	{
-		UNUSED(a_IsByBonemeal);
-		return GrowRipePlant({a_BlockX, a_BlockY, a_BlockZ});
-	}
-
 	/** Grows a cactus present at the block specified by the amount of blocks specified, up to the max height specified in the config; returns the amount of blocks the cactus grew inside this call */
```
12xx12, line 858: If we remove this we might want to add this to the deprecated bindings

12xx12 (changes requested):

src/World.h
tigerw: It's been tagged as obsolete in the API docs [here](https://api.cuberite.org/cWorld.html#GrowRipePlant_2) but I agree a warning wouldn't hurt.

tigerw: I ended up putting a warning directly in the header file, since I'd have to deal with the Vector3i variant as well if I exported manually with DeprecatedBindings.cpp. Also thought it's about time to remove the GrowCactus/GrowMelonPumpkin ones which did print warnings.

12xx12 (approved):

PR #4804, opened 2020-08-02
Adding support for 16 steps for rotation in templates for Blocks

12xx12:
I started to implement banners for cuberight and stubled about the problem that the template classes only support 4 steps of rotation.

I'd like to add 16 steps because e.g banners may be placed in 16 different directions.

If there's another way already implemented i'm happy to use that and delete this PR

As the last time feedback if this is neccesary is apreciated before i do the whole thing and trash it

### DIFF
--- a/src/Blocks/Mixins.h
+++ b/src/Blocks/Mixins.h
@@ -464,67 +464,67 @@ public:
 	{
 		if ((a_Rotation >= - 11.25f) && (a_Rotation < 11.25f))
 		{
-			return South
+			return South;
 		}
 		else if ((a_Rotation >= 11.25f) && (a_Rotation < 33.75f))
 		{
-			return SouthSouthWest
+			return SouthSouthWest;
 		}
 		else if ((a_Rotation >= 23.75f) && (a_Rotation < 56.25f))
 		{
-			return SouthWest
+			return SouthWest;
 		}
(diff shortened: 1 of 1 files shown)


12xx12:
Tomorrow I'm gonna share my scetches to determine the mirroring if anyone wants to proveread

tigerw:
yeah hold on a bit, there's going to be a migration to block states at some point in the future so all those NIBBLETYPE Metas will have to be removed.

Is it just banners that support more than 4 directions? You could hardcode it in GetPlacementBlockTypeMeta

12xx12:
I was wondering why the block type was still indicated by a damage value in the crafting and the rotation was still given in hexadecimal and not the new format. Now I know.

I think only heads do.

Is there anyone working on it? If not I got way to much time till the end of October and would do that if I may.

madmaxoft:
I think signposts use 16 rotations as well.

12xx12:
Indeed it does. but it's using a funky way of claculating the direction by actualy calculating the direction instead of doing an if/else

12xx12:
We can't use this for banners or sign post as they already inherit from the BlockEntityManger and the inheritance is not compatible... we can close this I think

bearbin:
This actually looks good, but as said earlier I think this would be better kept for after the blockstate transition, closing for now and we can review after blockstates start to be implemented.

PR #4806, opened 2020-08-02
Adding basic Banner functionality

### LINKED ISSUE #1535: Add Banners
http://minecraft.gamepedia.com/Banner

12xx12:
I'd like to implement the banners. Fixes #1535 

This just adds some basic functionality. The patterns need cNBT to be there. See #5033


12xx12:
This Pull is on hold until there are some NBT-Tags fo items. The basic functionality is here but there isn't any pattern crafting so the fun is missing

loading/saving/sending is implemented for banners and patterns
on placing and breakign patterns aren't transported but the performance sucks whyever

src/BlockEntities/BannerEntity.cpp
```diff
@@ -0,0 +1,125 @@
...
+unsigned char cBannerEntity::GetBaseColor() const
+{
+	return m_BaseColor;
+}
+
+
+
+
+
+const char * cBannerEntity::GetPatternTag(eBannerPattern a_Pattern)
```
12xx12, line 79: this is just there if we need it with the patterns - might be deleted later

12xx12:
This is just the PR for basic Banners - this can be merged in and the patterns can be done in a seperate PR.
I opened another issue for that #5033

tigerw:
I need to first familiarise myself with how banners work

src/WorldStorage/NBTChunkSerializer.cpp
```diff
@@ -378,6 +380,18 @@ class SerializerCollector:
 
 
 
+	void AddBannerEntity(cBannerEntity * a_Entity)
+	{
+		mWriter.BeginCompound("");
+			AddBasicTileEntity(a_Entity,"Banner");
+			mWriter.AddInt("Base", static_cast<int>(a_Entity->GetBaseColor()));
```
tigerw, line 387: The wiki lists a `Patterns` list which includes the base colour I believe

src/Items/ItemBanner.h
```diff
@@ -0,0 +1,221 @@
...
+	virtual bool IsPlaceable(void) override
+	{
+		return true;
+	}
+
+
+
+
+
+	virtual bool GetBlocksToPlace(
```
tigerw, line 40: Will GetPlacementBlockTypeMeta work since the placement is only one block

src/Items/ItemBanner.h
```diff
@@ -0,0 +1,221 @@
...
+			return false;
+		}
+		return true;
+	}
+
+
+
+
+
+	virtual bool OnPlayerPlace(
```
tigerw, line 169: I think it's possible to move the blockentity setting to cBlockBannerHandler::OnPlacedByPlayer; this would help reduce duplication

src/WorldStorage/NBTChunkSerializer.cpp
tigerw: Ah no, this is incorrect. The "Base" in Patterns refers to the Fully Field design.

Nonetheless, am I right in saying the banner's wool colour is encoded in the block, or in older versions the block meta?

12xx12: hmmm... if I recall correctly I checked the entities tags in the notchian implementation with the `/data` command and there was a base tag... and yeah the block or the meta should determine the color

12xx12: Oh and the meta stores the rotation

src/Items/ItemBanner.h
12xx12: No - we have to know the item's metadata to determine the color of the banner which is only known in `OnPlayerPlace`

12xx12:
I'm working on a fix because the block entity doesn't get transmitted after it replaces snow or something like that. It appears after reconnect

tigerw:
Thanks again for all of your work, currently I have a lot of stuff to do for university, do you mind if we get the bossbar PR merged first before this one?

12xx12:
No I don't mind - I myself have my first weekend after the semester. Don't worry!

tigerw (approved):

PR #4812, opened 2020-08-04
Code cleanup in cChunk.PlaceBlock() for checking if BlockEntity should be spawned

12xx12:
changed switch for call in cChunk for call to function cBlockEntity to remove unnecessary confusions while implementing new BlockEntities

### DIFF
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -1323,33 +1323,11 @@ void cChunk::SetBlock(Vector3i a_RelPos, BLOCKTYPE a_BlockType, NIBBLETYPE a_Blo
 	}
 
 	// If the new block is a block entity, create the entity object:
-	switch (a_BlockType)
+	if (cBlockEntity::IsBlockEntityBlockType(a_BlockType))
 	{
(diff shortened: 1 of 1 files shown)


src/Chunk.cpp
```diff
@@ -1323,33 +1323,11 @@ void cChunk::SetBlock(Vector3i a_RelPos, BLOCKTYPE a_BlockType, NIBBLETYPE a_Blo
...
-		case E_BLOCK_BREWING_STAND:
-		{
-			// Fast set block has already marked dirty
-			AddBlockEntityClean(cBlockEntity::CreateByBlockType(a_BlockType, a_BlockMeta, RelativeToAbsolute(a_RelPos), m_World));
-			break;
-		}
-	}  // switch (a_BlockType)
+	if (cBlockEntity::IsBlockEntityBlockType(a_BlockType))
+	{
+		// Fast set block has already marked dirty
```
madmaxoft, line 1328: This comment seems like a useless leftover from something old and long gone, consider removing it, too.

12xx12: done

PR #4813, opened 2020-08-06
Remove BountySource from funding sources

bearbin:
Nobody has used it recently, and it seems to be in a downwards spiral,
with the escrow stealing announcement a few months ago.


tigerw:
In memory of Tip4Commit and Bountysource.

Bounty chocolate

12xx12:
maybe also remove this from books.cuberite.com

PR #4816, opened 2020-08-06
Cleaning up some BlockEntty -> NBT tag code

12xx12:
I noticed that some of the code used for storing a BlockEntity to disk and sending it to the player is duplicated so i started to mere the code in a function of the block entity itself

I started with the beacon as an example if this even accepted the rest may follow if allowed

### DIFF
--- a/src/Protocol/Protocol_1_8.cpp
+++ b/src/Protocol/Protocol_1_8.cpp
@@ -3369,19 +3369,14 @@ void cProtocol_1_8_0::WriteBlockEntity(cPacketizer & a_Pkt, const cBlockEntity &
 		case E_BLOCK_BEACON:
 		{
 			auto & BeaconEntity = static_cast<const cBeaconEntity &>(a_BlockEntity);
-			Writer.AddInt("x",         BeaconEntity.GetPosX());
-			Writer.AddInt("y",         BeaconEntity.GetPosY());
-			Writer.AddInt("z",         BeaconEntity.GetPosZ());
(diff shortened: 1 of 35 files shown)


src/Protocol/Protocol_1_8.cpp
```diff
@@ -3369,19 +3369,14 @@ void cProtocol_1_8_0::WriteBlockEntity(cPacketizer & a_Pkt, const cBlockEntity &
...
 		{
 			auto & BeaconEntity = static_cast<const cBeaconEntity &>(a_BlockEntity);
-			Writer.AddInt("x",         BeaconEntity.GetPosX());
-			Writer.AddInt("y",         BeaconEntity.GetPosY());
-			Writer.AddInt("z",         BeaconEntity.GetPosZ());
-			Writer.AddInt("Primary",   BeaconEntity.GetPrimaryEffect());
-			Writer.AddInt("Secondary", BeaconEntity.GetSecondaryEffect());
-			Writer.AddInt("Levels",    BeaconEntity.GetBeaconLevel());
-			Writer.AddString("id", "Beacon");  // "Tile Entity ID" - MC wiki; vanilla server always seems to send this though
+			BeaconEntity.SerializeBlockEntity(Writer);
```
peterbell10, line 3372: This won't be able to support serialization changes in a future protocol versions.

12xx12: a - the we try to support all versions problem

ok - gonna close this then

12xx12:
Closed for maybe creating compability issues

PR #4817, opened 2020-08-06
Adding Emeralds to generation

12xx12:
While searching the code on learning how the server actually generates the map i noticed that emerald ore is not generated by default - now it does

https://minecraft.gamepedia.com/Emerald_Ore

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -13630,6 +13630,23 @@ end
 					},
 					Notes = "Returns true if the biome is very cold (has snow on ground everywhere, turns top water to ice, has snowfall instead of rain everywhere). Doesn't report mildly cold biomes (where it snows above certain elevation), use IsBiomeCold() for those.",
 				},
+				IsMountainBiome =
+				{
+					Params =
+					{
+						{
+							Name = "Biome",
+							Type = "EMCSBiome",
+						}
+					},
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns if the biome is a mountain type biome. So mutations of the extreme hills biome"
+				},
 				IsValidBlock =
 				{
 					Params =
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -13630,7 +13630,7 @@ end
 					},
 					Notes = "Returns true if the biome is very cold (has snow on ground everywhere, turns top water to ice, has snowfall instead of rain everywhere). Doesn't report mildly cold biomes (where it snows above certain elevation), use IsBiomeCold() for those.",
 				},
-				IsMountainBiome =
+				IsBiomeMountain =
 				{
 					Params =
 					{
--- a/src/BiomeDef.h
+++ b/src/BiomeDef.h
@@ -152,7 +152,7 @@ Doesn't report Very Cold biomes, use IsBiomeVeryCold() for those. */
 extern bool IsBiomeCold(EMCSBiome a_Biome);
 
 /** Returns true if the biome is a mountain type */
-extern bool IsMountainBiome(EMCSBiome a_Biome);
+extern bool IsBiomeMountain(EMCSBiome a_Biome);
 
 /** Returns the height when a biome when a biome starts snowing. */
 extern int GetSnowStartHeight(EMCSBiome a_Biome);
--- a/src/BiomeDef.h
+++ b/src/BiomeDef.h
@@ -151,6 +151,9 @@ extern bool IsBiomeVeryCold(EMCSBiome a_Biome);
 Doesn't report Very Cold biomes, use IsBiomeVeryCold() for those. */
 extern bool IsBiomeCold(EMCSBiome a_Biome);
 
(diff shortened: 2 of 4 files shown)


bearbin:
I don't like this, it feels a little too much like a special case. I think it would be better to add some configuration value to the ore types, what biomes they generate in - and then check the location when actually placing the ore nest.

Pokechu22:
Another thing that'd need special handling (not sure if it's implemented already) is [gold ore, which generates more frequently](https://minecraft.gamepedia.com/Gold_Ore#Natural_generation) in badlands (mesa) biomes.

12xx12:
@Pokechu22 i just checked in with minecraft wiki and saw that too

12xx12:
I just added the check before to save the time of even generating anything

bearbin:
Well, I would wait for another comment by some other developer to see what they think.

12xx12:
this is so a massive edge case that i would have to change the code some more to replicate vanilla behaviour because the ore is generated only in single ores but the code generates mostly two ore veines

12xx12:
So i made all checks pass - maybe someone will comment this.

- bearbin proposed that yould check beforehand - I don't think that is fast as you would do the setup and then dump it if the biome doesn't match. I would do the biome check before doing anything else and return if biome doesn't match

- bearbin also proposed to add a value decribing which biomes should generate - this would cause a massive list of all biomes except a few (for gold) so i would add two lists. one describing the ones allowed the other one the disallowed only one could be populated

src/Generating/FinishGen.cpp
```diff
@@ -1763,7 +1782,16 @@ void cFinishGenOreNests::GenerateOre(
 		nestRnd /= cChunkDef::Width;
 		int BaseY = nestRnd % a_MaxHeight;
 		nestRnd /= a_MaxHeight;
-		int NestSize = a_NestSize + (nestRnd % (a_NestSize / 4));  // The actual nest size may be up to 1 / 4 larger
```
tigerw, line 1766: Would std::min(4, a_NestSize) work

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -13630,6 +13630,23 @@ end
...
+							Type = "EMCSBiome",
+						}
+					},
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+					},
+					Notes = "Returns if the biome is a mountain type biome. So mutations of the extreme hills biome"
```
tigerw, line 13648: For consistency with the others, "Returns true if the biome is mountainous (mutations of the extreme hills biome)."

src/BiomeDef.h
```diff
@@ -151,6 +151,9 @@ extern bool IsBiomeVeryCold(EMCSBiome a_Biome);
 Doesn't report Very Cold biomes, use IsBiomeVeryCold() for those. */
 extern bool IsBiomeCold(EMCSBiome a_Biome);
 
+/** Returns true if the biome is a mountain type */
```
tigerw, line 154: and the final full stop too.

tigerw:
As long as there's aren't too many exceptions, it should be fine. But if Mojang gets itchy fingers (as they are wont to do...) and start adding more we'd need something more generalised. 12xx12 do you see the gold ore generation Pokechu22 mentioned as being easy to add?

src/Generating/FinishGen.cpp
12xx12: calculating  x mod y with 0 < y < 1 results in realy big negative values so i think
```min(a_NestSize, abs(a_NestSize + (nestRnd % (a_NestSize / 4)))```
would do the thing here

Server/Plugins/APIDump/APIDesc.lua
12xx12: done

src/BiomeDef.h
12xx12: added the new description here too

12xx12:
at this point it's only emeralds and the gold is some biomes

the gold should be quite simple. i'm thinking about adding another funtion biomedef.h if it's a mesa biome or not. or i change the if question to an int based one to keep the code from bloating up by doing a 20 part if test

for the gold i would do another check if the block to place is gold, sample the four spots and check if its a mesa biome and then change the values which would resolve in some magic numers in the script which i would explain

src/Generating/FinishGen.cpp
tigerw: I wonder if the issue is that a_NestSize / 4 does an integer division, so if it's less than 4 it rounds to zero, which blows up the modulo with a divide by zero

tigerw:
hm. well this looks ok to me, let's see what others say

12xx12:
the only thing left to talk about is that with the momentarily ore generation code you generate mostly two ores next each other.

src/Generating/FinishGen.cpp
12xx12: so i actually rounds to zero and then crashes with a division by zero

12xx12: you solution works but creates veines with at least one more as the defined size

12xx12:
@tigerw can you add your documentation suggestions? this was closed to fast

PR #4818, opened 2020-08-08
Fix arrows preventing block placement

### LINKED ISSUE #4815: [Bug] You cant place a block when the position is occupied by an arrow
Client version: 1.8.2
Server OS: Windows

### Expected behavior
You can place a block when the position is occupied by a arrow (shot by player or skeleton)
Just tested this in the vanilla single player

### Actual behavior
You can't place the blog

- Shoot arrow
- try to place a block where it landed

ion232:
Changes ArrowEntity to override DoesPreventBlockPlacement to return false instead of the default true.
The implementation of DoesPreventBlockPlacement could be changed if there are instances where an arrow should prevent block placement.

Fixes #4815

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -24,6 +24,7 @@ Gareth Nelson
 GefaketHD
 HaoTNN
 Howaner
+ion232 (Arran Ireland)
 jan64
 jasperarmstrong
 kevinr (Kevin Riggle)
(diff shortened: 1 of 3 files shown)


bearbin:
I think @tigerw was the last person to be working on this area of the code. Arrows seem to be very difficult in that the client does lots of assumptions on its own and breaks in quite silly ways if we don't follow the vanilla method exactly.

I will do some testing of this PR, but can you confirm that arrows correctly pop up out of the newly placed block and don't 'go black', they can still be collected, and that the correct behaviour persists after relogging?

ion232:
@bearbin I've tested the things you've mentioned and my fix appears to work just fine. The behaviour becomes identical to vanilla minecraft.

bearbin:
I just tested again myself and it all looks good, thanks for the contribution!

PR #4819, opened 2020-08-09
Fix enderchest placement rotation

### LINKED ISSUE #4810: [Bug] Ender chest is first displayed the right way and the flipped by 180 degrees
Client version: 1-8-2
Server OS: Windows

### Expected behavior
Anderchest is placed facing the player

### Actual behavior
enderchest is placed facing the player but immediatly flipped by 180 degrees to face away from him

place ander chest

ion232:
The enderchest didn't have the same type of cYawRotator that a normal chest has. I have tested that this works and the fix produces the correct behaviour.

Fixes #4810

### DIFF
--- a/src/Blocks/BlockEnderchest.h
+++ b/src/Blocks/BlockEnderchest.h
@@ -8,9 +8,9 @@
 
 
 class cBlockEnderchestHandler :
-	public cYawRotator<cBlockEntityHandler>
+	public cYawRotator<cBlockEntityHandler, 0x07, 0x03, 0x04, 0x02, 0x05>
 {
-	using Super = cYawRotator<cBlockEntityHandler>;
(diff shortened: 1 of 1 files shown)


src/Blocks/BlockEnderchest.h
```diff
@@ -8,9 +8,9 @@
 
 
 class cBlockEnderchestHandler :
-	public cYawRotator<cBlockEntityHandler>
+	public cYawRotator<cContainerEntityHandler<cBlockEntityHandler>, 0x07, 0x03, 0x04, 0x02, 0x05>
```
peterbell10, line 11: As weird as this may sound, Enderchests aren't "Container Entities" because the ender chest inventory is stored in the player and not the chest. The `cYawRotator` fix on its own should be enough.

tigerw:
merge?

bearbin:
@tigerw Once cContainerEntityHandler is removed.

PR #4826, opened 2020-08-13
removed the calculation in the noteblock block entity

12xx12:
I looked for todo's in the code and this is the first one I know how to fix

I did the calculation in python if anyone is interested where the numbers are from
```python 
for i in range(0,25):
    print(2**((i-12)/12))
```

### DIFF
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -720,7 +720,7 @@ void DeprecatedBindings::Bind(lua_State * tolua_S)
 	tolua_beginmodule(tolua_S, "cNoteEntity");
 		tolua_function(tolua_S, "GetPitch", tolua_AllToLua_cNoteEntity_GetNote00);
 		tolua_function(tolua_S, "SetPitch", tolua_AllToLua_cNoteEntity_SetNote00);
-		tolua_function(tolua_S, "IncrementNote", tolua_AllToLua_cNoteEntity_IncrementNote00);
+		tolua_function(tolua_S, "IncrementPitch", tolua_AllToLua_cNoteEntity_IncrementNote00);
 	tolua_endmodule(tolua_S);
 
 	tolua_endmodule(tolua_S);
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -477,7 +477,7 @@ static int tolua_cNoteEntity_GetPitch(lua_State * tolua_S)
 	if (
 		!LuaState.CheckParamUserType(1, "cNoteEntity") ||
 		!LuaState.CheckParamEnd(2)
-		)
+	)
 	{
 		return 0;
 	}
@@ -493,7 +493,7 @@ static int tolua_cNoteEntity_GetPitch(lua_State * tolua_S)
 		tolua_error(LuaState, "invalid 'self' in function 'GetPitch'", nullptr);
 	}
 	LuaState.Push(Self->GetNote());
-	LOGWARNING("Warning: 'cNoteEntity:GetPitch' function is deprecated. Please use 'cNoteEntity:a' instead.");
+	LOGWARNING("Warning: 'cNoteEntity:GetPitch' function is deprecated. Please use 'cNoteEntity:GetNote' instead.");
 	LuaState.LogStackTrace(0);
 	return 1;
 }
@@ -509,7 +509,7 @@ static int tolua_cNoteEntity_IncrementPitch(lua_State * tolua_S)
 	if (
 		!LuaState.CheckParamUserType(1, "cNoteEntity") ||
 		!LuaState.CheckParamEnd(2)
-		)
+	)
 	{
 		return 0;
 	}
@@ -543,7 +543,7 @@ static int tolua_cNoteEntity_SetPitch(lua_State * tolua_S)
 		!LuaState.CheckParamUserType(1, "cNoteEntity") ||
 		!LuaState.CheckParamNumber(2) ||
 		!LuaState.CheckParamEnd(3)
-		)
+	)
 	{
 		return 0;
 	}
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -10,6 +10,7 @@
 #include "../Entities/Player.h"
 #include "LuaState.h"
 #include "../BlockInfo.h"
+#include "../BlockEntities/NoteEntity.h"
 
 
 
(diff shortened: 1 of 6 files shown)


src/BlockEntities/NoteEntity.cpp
```diff
@@ -245,8 +245,35 @@ void cNoteEntity::MakeSound(void)
...
 	m_World->BroadcastBlockAction(m_Pos, static_cast<Byte>(instrument), static_cast<Byte>(m_Pitch), E_BLOCK_NOTE_BLOCK);
 
-	// TODO: instead of calculating the power function over and over, make a precalculated table - there's only 24 pitches after all
-	float calcPitch = static_cast<float>(pow(2.0f, static_cast<float>(m_Pitch - 12.0f) / 12.0f));
+	float calcPitch;
+	switch (m_Pitch)
+	{
+		case 0: calcPitch = 0.5f;
+		case 1: calcPitch = 0.5297315471796477f;
+		case 2: calcPitch = 0.5612310241546865f;
```
tigerw, line 253: I couldn't bother you to make this into a static function, could I? As this code stands the lack of `break` means execution falls all the way to the bottom and `calcPitch` ends up always with value `2`.

12xx12: You find my lack of breaks disturbing...

I could move this to a static function of course

12xx12: i hope the new commit is in the way you asked for

12xx12:
I think I'm gonna change the m_Pitch to m_Note to prevent confusion

Bond-009:
I’d add the original formula as comment somewhere in CalculatePitch.
Also, CalculatePitch is a misnomer as it doesn’t really calculate anything

src/BlockEntities/NoteEntity.h
```diff
@@ -56,6 +56,10 @@ class cNoteEntity :
 	virtual bool UsedBy(cPlayer * a_Player) override;
 	virtual void SendTo(cClientHandle &) override {}
 
+	/** Calculates the pitch to be send to the players
+	from a pitch value from 0 to 24 as incremented by the IncrementPitch function */
```
madmaxoft, line 60: How about:
```suggestion
	/** Returns the relative pitch (used in the protocol)
	from a note value between 0 and 24 (used in m_Pitch). */
```

src/BlockEntities/NoteEntity.h
```diff
@@ -7,7 +7,7 @@
 
 
 
-enum ENUM_NOTE_INSTRUMENTS
+enum ENUm_Pitch_INSTRUMENTS
```
madmaxoft, line 10: Why rename this in such a weird fashion?

src/BlockEntities/NoteEntity.h
```diff
@@ -56,6 +56,10 @@ class cNoteEntity :
 	virtual bool UsedBy(cPlayer * a_Player) override;
 	virtual void SendTo(cClientHandle &) override {}
 
+	/** Calculates the pitch to be send to the players
+	from a pitch value from 0 to 24 as incremented by the IncrementPitch function */
+	static float GetPitch(char a_Note);
```
madmaxoft, line 61: Perhaps also rename to `PitchFromNote` to be absolutely clear about this?

madmaxoft (changes requested):
I quite liked the renaming of `m_Pitch` to `m_Note`, makes much more sense. Changing the API is okay in this case, you'll just need to provide a compatibility stub in src/Bindings/DeprecatedBindings.cpp (there are a lot of examples there), so that plugins using the old function names will still work, but emit a warning to the console.

madmaxoft:
Let me know if you need help with the deprecated bindings.

src/BlockEntities/NoteEntity.h
12xx12: ```(used in m_Note)``` I think

madmaxoft: Yes, if you rename it again :)

src/BlockEntities/NoteEntity.h
12xx12: Thats a nice name

src/BlockEntities/NoteEntity.h
12xx12: This is a good Question - didn't notice that.. used the rename tool and this was the result - gonna change that

src/BlockEntities/NoteEntity.h
12xx12: Renamed

12xx12:
I thing I got the binding

src/BlockEntities/NoteEntity.h
12xx12: reverted changes

src/BlockEntities/NoteEntity.h
12xx12: changed function documentation

12xx12:
ok - i need help with the deprecated bindings

tigerw:
For what it's worth I'm not seeing any calls to DoWithNoteBlockAt in Lua files when doing a GitHub code search. Probably very few plugins manipulate noteblocks.

src/Bindings/DeprecatedBindings.cpp
```diff
@@ -599,6 +599,13 @@ void DeprecatedBindings::Bind(lua_State * tolua_S)
 		tolua_function(tolua_S, "UpdateSign", tolua_cWorld_SetSignLines);
 	tolua_endmodule(tolua_S);
 
+
+	tolua_beginmodule(tolua_S, "cNoteEntity");
+		tolua_function(tolua_S, "GetPitch", tolua_AllToLua_cNoteEntity_GetNote00);
+		tolua_function(tolua_S, "SetPitch", tolua_AllToLua_cNoteEntity_SetNote00);
+		tolua_function(tolua_S, "IncrementNote", tolua_AllToLua_cNoteEntity_IncrementNote00);
```
madmaxoft, line 606: You can't "redirect" to a handler in automatic bindings - they are marked `static` so they are unavailable from other source files; also we want to output a warning and stacktrace to the console if these old functions are used.

Instead, create new functions for these that remaps the call:
```cpp
static int tolua_cNoteEntity_GetPitch(lua_State * tolua_S)
{
	cLuaState LuaState(tolua_S);

	if (
		!LuaState.CheckParamUserType(1, "cNoteEntity") ||
		!LuaState.CheckParamEnd(2)
	)
	{
		return 0;
	}

	cNoteEntity * Self = nullptr;

	if (!LuaState.GetStackValues(1, Self))
	{
		tolua_error(LuaState, "Failed to read parameters", nullptr);
	}
	if (Self == nullptr)
	{
		tolua_error(LuaState, "invalid 'self' in function 'GetNote'", nullptr);
	}
	LuaState.Push(Self->GetNote());
	LOGWARNING("Warning: 'cNoteEntity:GetPitch' function is deprecated. Please use 'cNoteEntity:GetNote' instead.");
	LuaState.LogStackTrace(0);
	return 1;
}
```

Also, please alpha-sort both the functions in a module, and module by name.

Issue #4428, opened 2019-11-27
Support 1.13 clients


madmaxoft:
This is a general issue for tracking the 1.13 progress.

aviallon:
It seems there is still some work to do. Anyway, incredible code quality. One of the best designed project I've ever seen in my entire life.

leonardus:
Anything I can do to help?

mathiascode:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=84482406)](https://www.bountysource.com/issues/84482406-support-1-13-clients?utm_source=84482406&utm_medium=shield&utm_campaign=ISSUE_BADGE)

madmaxoft:
Note that this has seem some discussion on the forum. There's a general plan there for the support implementation.
https://forum.cuberite.org/thread-3215.html

mathiascode:
At what state are we with all the block stuff? Is something being developed/changed right now?

madmaxoft:
Nothing's being done right now. Sorry.

We have the block type registry, block type transformations between 1.12 and 1.13 (mostly), protocol shim that uses those transformations for a de-facto compatibility, and a very rudimentary `cBlockArea` replacement that uses the 1.13+ blocks (`PalettedBlockArea`).

madmaxoft:
Probably the biggest task now is to pull all block-type-specific behavior from all core server files into the block handlers. No-one outside the $/src/Blocks should even `#include "BlockID.h"`.

madmaxoft:
Similar with items, but those don't hurt us as bad as the block currently do.

mathiascode:
@tigerw What's the current status on 1.13? What's left to do, and are you currently working on something?

tigerw:
There's still much work to do for full support. Protocol-wise thanks to your changes we have the basics (blocks, entities) down.

There were other packet format changes and ID changes (particle effects, paintings) that are yet to be implemented, they require more work on the Registry.

One big thing (but only visual) for blocks is that the client now doesn't help us at all for the rendering of walls, redstone wire, tripwires etc. It expects us to describe exactly which direction they connect via the block state.

I'm currently trying to improve redstone sim performance, hoping to maybe work on protocols more, soon :D

PR #4829, opened 2020-08-19
Adding block entity transmission to 1.13

12xx12:
related: #4428 

I found a missing package type: added PacketSetBeaconEffect
The client sends this to set the beacon effects in the beacon window

At this point this relies on the last implementation of the nbt serialasation implemented in 1.11

This needs more testing

- [x] Set data of a mob spawner
- [x] Set command block text (command and last execution status)
The block spawns but isn't interactable  - further research needed
- [x] Set the level, primary, and secondary powers of a beacon
- [x] Set rotation and skin of mob head
 Mob Heads were renamed during "the flattening" there is a block spawning but neither rotation nor head type works because this is not related to a block entity anymore
- [x] Declare a conduit
the conduit is not implemented (doesn't exist as item or block)
- [x] Set base color and patterns on a banner
dependent on other pull request - is ticked off due to implementation in other pull request
- [x] Set the data for a Structure tile entity
Ok this is fun but not at all implemented - so i tick this of (all that is possible is implemented)
- [x] Set the destination for a end gateway
So the block exists in the enum, but the item/block is not implemented at this point so this is ticked of too
- [x] Set the text on a sign
- [x] Declare a bed
the color of the bed got moved to the block data so it doesn't display the correct color but is useable

Some Types don't work due to missing block data:
- Bed

Some blocks were renamed during "the flattening"
- Mob Heads

Some Blocks are not implemented:
- conduit
- Banner

### DIFF
--- a/src/Protocol/Protocol_1_13.cpp
+++ b/src/Protocol/Protocol_1_13.cpp
@@ -27,6 +27,7 @@ Implements the 1.13 protocol classes:
 #include "../Server.h"
 #include "../World.h"
 #include "../JsonUtils.h"
+#include "../WorldStorage/FastNBT.h"
 
 #include "../Bindings/PluginManager.h"
 
--- a/src/Protocol/Protocol_1_13.cpp
+++ b/src/Protocol/Protocol_1_13.cpp
@@ -173,7 +173,32 @@ void cProtocol_1_13::SendTabCompletionResults(const AStringVector & a_Results)
 
 void cProtocol_1_13::SendUpdateBlockEntity(cBlockEntity & a_BlockEntity)
 {
-	// TODO
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, pktUpdateBlockEntity);
+	cFastNBTWriter Writer;
+
+	Pkt.WriteXYZPosition64(a_BlockEntity.GetPosX(), a_BlockEntity.GetPosY(), a_BlockEntity.GetPosZ());
+
+	Byte Action = 0;
+	switch (a_BlockEntity.GetBlockType())
+	{
+		case E_BLOCK_MOB_SPAWNER:   Action = 1;  break;  // Update mob spawner spinny mob thing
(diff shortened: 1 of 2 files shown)


12xx12:
the client leaves at one of those packets:

https://pastebin.com/YQikUA43

with the decoder exception:
readerIndex(5) + length(4) exceeds writerIndex(6)

12xx12:
i kept the client running for like an hour between the implemented block enteties and it still didn't crash

src/Protocol/Protocol_1_13.cpp
```diff
@@ -173,7 +174,32 @@ void cProtocol_1_13::SendTabCompletionResults(const AStringVector & a_Results)
 
 void cProtocol_1_13::SendUpdateBlockEntity(cBlockEntity & a_BlockEntity)
 {
-	// TODO
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, pktUpdateBlockEntity);
+	cFastNBTWriter Writer;
```
tigerw, line 179: I don't think Writer is being used

tigerw:
Cool! In other news, I feel we need another palette mapping, this time for block entity `Action`...

12xx12:
Until now it was just hardcoded in any Versions Protocoll.cpp

src/Protocol/Protocol_1_13.cpp
12xx12: You are right- this is a leftover from me trying out some things

12xx12: done

src/Protocol/Protocol_1_13.cpp
```diff
@@ -27,6 +27,7 @@ Implements the 1.13 protocol classes:
 #include "../Server.h"
 #include "../World.h"
 #include "../JsonUtils.h"
+#include "../WorldStorage/FastNBT.h"
```
tigerw, line 30: this too

tigerw:
I yes this PR

Issue #1961, opened 2015-05-08
Generator for single-prefab structures ($5)


NiLSPACE:
MCServer should have a generator that can place structures like desert temples or witch huts. All these structures can probably use the same code, but with different prefabs, and different spawning conditions (In what biome can they spawn etc)

Structures:
- [Desert temple](http://minecraft.gamepedia.com/Desert_temple)
- [Jungle temple](http://minecraft.gamepedia.com/Jungle_temple)
- [Witch hut](http://minecraft.gamepedia.com/Generated_structures#Witch_hut)
- [Desert well](http://minecraft.gamepedia.com/Generated_structures#Desert_well)
- [Ocean monument](http://minecraft.gamepedia.com/Ocean_monument)

I believe all of these except the ocean monument are already in the [gallery server](http://forum.mc-server.org/showthread.php?tid=1372&pid=12382#pid12382)

madmaxoft:
The ocean monument will most likely need a separate generator, since its insides vary a bit randomly.

NiLSPACE <notifications@github.com> wrote:

>MCServer should have a generator that can place structures like desert
>temples or witch huts. All these structures can probably use the same
>code, but with different prefabs, and different spawning conditions (In
>what biome can they spawn etc)
>
>Structures:
>* [Desert temple](http://minecraft.gamepedia.com/Desert_temple)
>* [Jungle temple](http://minecraft.gamepedia.com/Jungle_temple)
>* [Witch
>hut](http://minecraft.gamepedia.com/Generated_structures#Witch_hut)
>* [Desert
>well](http://minecraft.gamepedia.com/Generated_structures#Desert_well)
>* [Ocean monument](http://minecraft.gamepedia.com/Ocean_monument)
>
>I believe all of these except the ocean monument are already in the
>[gallery
>server](http://forum.mc-server.org/showthread.php?tid=1372&pid=12382#pid12382)
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/mc-server/MCServer/issues/1961

-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

NiLSPACE:
Now that #2422 is implemented we only need the prefabs to close this right?

madmaxoft:
I guess the `PieceStructures` generator could be manipulated into this, but it would feel like a hack, not to mention that using prefabs that way would break the checks currently being done in the GalExport plugin - namely that each area has at least one connector.

[...]

PR #4830, opened 2020-08-19
Adding Generator For Single Piece Structures

12xx12:
Closing #1961 

This pull request adds the Generator for single piece structures

Forum link: https://forum.cuberite.org/thread-3307-post-29809.html

What has been done:
- [x] Adding cSinglePieceStructureGen
This is similar to cPieceStructureGen and reads any specified Prefab from Prefabs/SinglePiecePrefabs/* and places it
- [x] Added Vanilla structures to world.ini
- [x] Export prefabs for further testing

Small Things to do:
- [x] some whitespace stuff
- [x] remove loot tables -> new PR

Intended use for prefabs is "SinglePieceStructures".
Any Structure that should be generated should have all it's options as starting pieces. (to be checked if works without)


madmaxoft:
So you started churning out code without even talking about it a tiny bit?

I think the generator should be parametrisable in the world.ini file, similar to the `PieceStructures` generator - the `Finishers` could contain for example `SinglePiece: WitchHut` item and it would add a new `cSinglePrefabStructure` generator and load the `WitchHut` prefab into it.

Since this is a completely new piece of code, some discussion over how it's gonna be structured would be nice, join us at the forums for that. PRs are not a good platform, they're too noisy for discussions and they're too late to discuss ideas behind the code.

12xx12:
Yeah - maybe should have done that. Gonna do that the next time defenetly. I wrote some things describing what id like to on the forum

12xx12:
And I opened this PR because I poked around in the code finding out how it actually is done and tried out some things and wanted to save them and show that there is some WIP

madmaxoft:
The build is failing because of linker issues, your new file is not added to the tests.

You can reproduce locally by adding these flags to CMake: `-DBUILD_TOOLS=1 _DSELF_TEST=1` - then the solution / makefile contains all the projects that the CI actually builds.

src/Generating/CMakeLists.txt
```diff
@@ -54,6 +55,7 @@ target_sources(
 	Prefab.h
 	PrefabPiecePool.h
 	PrefabStructure.h
+        SinglePieceStructuresGen.h
```
madmaxoft, line 58: Use tabs for indent

src/Generating/SinglePieceStructuresGen.h
```diff
@@ -0,0 +1,52 @@
...
+#include "Globals.h"
+
+#include "ComposableGenerator.h"
+#include "PrefabPiecePool.h"
+
+/** The Single Prefab Structure Generator:
+	This uses the cGridStructGen to generate the structures on the map
+	This is similar to the Piece Structure Generator but only placing one possible structure
+	The Exported cubeset MUST have all possible structures as start structures
+	else it isn't accessible from the m_Piecepool. */
```
madmaxoft, line 20:
```suggestion
This uses the cGridStructGen to generate the structures on the map
This is similar to the Piece Structure Generator but only placing one possible structure
The Exported cubeset MUST have all possible structures as start structures
else it isn't accessible from the m_Piecepool. */
```

src/Generating/SinglePieceStructuresGen.cpp
```diff
@@ -0,0 +1,249 @@
...
+		return true;
+	}
+
+
+
+
+	// cGridStructGen override
+	virtual cStructurePtr CreateStructure(int a_GridX, int a_GridZ, int a_OriginX, int a_OriginZ) override
+	{
+		LOG("CreateStructure for %s at <%d, %d>", m_Name.c_str(), a_GridX, a_GridZ);
```
madmaxoft, line 113: This looks like a remnant from debugging, probably not wanted in final code.

madmaxoft (dismissed):

src/Generating/CMakeLists.txt
12xx12: Already fixed and changed the settings in my IDE (drives me crazy)

src/Generating/SinglePieceStructuresGen.cpp
12xx12: Yes it is. In what way is the debug output limited? Is there a Compiler variable? I think this would be nice to keep for debugging.

This is also outputted by the other PieceStructureGenerator (code is takes from there)

src/Generating/ComposableGenerator.cpp
```diff
@@ -226,7 +227,8 @@ void cComposableGenerator::InitializeGeneratorDefaults(cIniFile & a_IniFile, eDi
 				"DeadBushes, "
 				"NaturalPatches, "
 				"PreSimulator, "
-				"Animals"
+				"Animals, "
+				"SinglePieceStructures: JungleTemple|WitchHut|DesertTemple|DesertWell"
```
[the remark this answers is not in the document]
12xx12: I think I'm gonna move this next to the villages. That would make most sense I think.

src/Generating/SinglePieceStructuresGen.h
12xx12: done (but not pushed)

src/Generating/ComposableGenerator.cpp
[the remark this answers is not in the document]
12xx12: done (not pushed)

src/Generating/SinglePieceStructuresGen.cpp
madmaxoft: I guess it's a problem in those generators as well.

The logging subsystem is somewhat poor. We have `LOGD`, `LOG`, `LOGWARNING` and `LOGERROR`:
- `LOGD` stays in Debug builds and is removed in Release builds. (this is what this particular line could use, if wanted; but I think this is really a needless piece of information)
- `LOG` is for regular messages, such as status.
- `LOGWARNING` is for warnings, the text is output in a highlighted color in the console
- `LOGERROR` is for big hard errors that can compromise the server. The text output is very highlighted in the console.

There's no other meaning and processing behind these.

src/Generating/CMakeLists.txt
madmaxoft: Normal IDEs should be able to pick up our editorconfig file, so no setup should be necessary. Unless you use something exotic.

src/Generating/SinglePieceStructuresGen.cpp
12xx12: To make it more clear:
this line is taken form the cPieceStructuresGen. The line in the class is NOT excluded when not compiled as debug. So you still get the output informing you about structure placement during generation when using the release version

src/Generating/CMakeLists.txt
12xx12: I use CLion and this mostly happens when I use the rename feature - it even recognizes that there is a clang file included

src/Generating/SinglePieceStructuresGen.cpp
12xx12: So you think we should remove this from ALL generators? I'm gonna do ```LOGD``` at last as this is kind of a cheat

madmaxoft: I think it should be removed altogether, because it makes no sense to report this to the server admin. It could be useful while developing, but not past that stage. It should have been removed from the other generators as well (but don't do that in this PR)

12xx12: Ok - gonna do this for my generator then.

Of course else it's gonna be a big mess when trying to find out what was changed

12xx12:
this should be ready. I have the prefabs ready but not commited as they have to be exported from the actual galery server

madmaxoft:
Well, we do need the prefabs inside this PR, so whatever you need on the Gallery server, let me know. Although it's not strictly required for the prefabs to be exported from the primary server, it is highly recommended, so that there's a single source with a possibility of re-export after a change.

12xx12:
Ok then I'm gonna proofread them again and push

12xx12:
ok- this will take some more time:

- i got to add more sponging to enable the generators to be burried in sand
- i got to fix the crash when using vertical strategy

12xx12:
found out why the server crashes

src/Generating/SinglePieceStructuresGen.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "Globals.h"
+
+#include "ComposableGenerator.h"
+#include "PrefabPiecePool.h"
+
+/** The Single Prefab Structure Generator:
+This uses the cGridStructGen to generate the structures on the map
+This is similar to the Piece Structure Generator but only placing one possible structure
+The Exported cubeset MUST have all possible structures as start structures or the server crashes on generation
```
madmaxoft, line 19: Server crashing on bad (admin) input is a bad idea, can't we fix the code instead?

12xx12: hm...

this would require some changes further down the line. you may ask for the starting piece but if there's no starting piece in the first place you get all pieces. So this would require the load from file function in the piece pool to check if there's a starting piece.

the crash results from the fact that the height generator is a nullpointer in the object responsible for place the piece according to the vertical strategy if there's no starting piece.

so if there is no starting piece we either
a) assign every piece's vertical strategy the generator data
b) refuse to load the cubeset
c) return no pieces and cancel generating

12xx12:
DO NOT MERGE
One thing is to do

12xx12:
this is reaview ready

src/Generating/SinglePieceStructuresGen.h
12xx12: this error shouldn't occure only in the new structures but also in the others if there is no starting structure defined so it would require some changes in the prefal pool class

12xx12: i opened a issue for that. If you think this is related enough i can fix this in this PR

src/Generating/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ target_sources(
 	PiecePool.cpp
 	PieceStructuresGen.cpp
 	Prefab.cpp
+	SinglePieceStructuresGen.cpp
```
madmaxoft, line 23: Please alpha-sort

src/Generating/CMakeLists.txt
```diff
@@ -54,6 +55,7 @@ target_sources(
 	Prefab.h
 	PrefabPiecePool.h
 	PrefabStructure.h
+	SinglePieceStructuresGen.h
```
madmaxoft, line 58: Alpha-sort

src/Generating/SinglePieceStructuresGen.h
madmaxoft: Nah, a separate PR it is. But since you already investigated this, it'd be great if you fixed it too, as long as the memory's still fresh.

tests/Generating/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ set (GENERATING_SRCS
 	${CMAKE_SOURCE_DIR}/src/Generating/PieceGeneratorBFSTree.cpp
 	${CMAKE_SOURCE_DIR}/src/Generating/PiecePool.cpp
 	${CMAKE_SOURCE_DIR}/src/Generating/PieceStructuresGen.cpp
+	${CMAKE_SOURCE_DIR}/src/Generating/SinglePieceStructuresGen.cpp
```
madmaxoft, line 55: Alpha-sort

tests/Generating/CMakeLists.txt
```diff
@@ -119,6 +120,7 @@ set (GENERATING_HDRS
 	${CMAKE_SOURCE_DIR}/src/Generating/PieceGeneratorBFSTree.h
 	${CMAKE_SOURCE_DIR}/src/Generating/PiecePool.h
 	${CMAKE_SOURCE_DIR}/src/Generating/PieceStructuresGen.h
+	${CMAKE_SOURCE_DIR}/src/Generating/SinglePieceStructuresGen.h
```
madmaxoft, line 123: Alpha-sort

madmaxoft (changes requested):
Code's fine, just some minor style issues. Good job!

madmaxoft:
Also, please rebase this on top of current master (and force-push), so that the CI check is up to date.

12xx12:
> 
> 
> Code's fine, just some minor style issues. Good job!

thank you

The CMakeFiles are now alpha-sorted and the branch is up to date

madmaxoft:
We generally prefer rebasing to merging, it makes the history cleaner to view.

madmaxoft (approved):

12xx12:
Oh what's the git-command for that?

madmaxoft:
Usually this works:
```
git fetch upstream master:master
git rebase master
```
(This assumes that your own github repo is named "origin" and you have an "upstream" repo pointing to the official Cuberite repository, use `git remote add upstream https://github.com/cuberite/cuberite` once to add it this way)
The first command downloads the current changes from the official repo's master branch (`master`) and moves you local master branch to point to them (`:master`). The second command rebases your current branch on top of your master branch (which is now the same as the official master).

The rebase itself may produce some conflicts needing manual resolution, especially when your branch is old. Therefore it's a good idea to rebase your branch often, not just at the end of a PR.

You can also do an interactive rebase (`git rebase -i master`) which allows you to manipulate your local commits - merge some of them, reword the commit messages, even reorder the commits. May be useful for those "fixed style" commits, although we will squash most PRs into a single commit anyway. The rule of thumb is, if there's a separate functionality in a commit and the entire codebase compiles, such a commit can stay separate in the final repo; otherwise we squash it.

12xx12:
Thank you for that complete explanation ☺️

bearbin (approved):

PR #4833, opened 2020-08-25
Fix doors and trapdoors closing on server restart

### LINKED ISSUE #4831: Trapdoors do not behave properly
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: a473e8ba52243a1a2a4f3aac840ed884d994fb0b

### Expected behavior
The position of trapdoors should be retained upon server restart.

### Actual behavior
Trapdoors reset to the horizontal orientation upon server restart.

1. Place a trapdoor.
2. Right-click on the trapdoor (putting it in the vertical orientation).
3. Restart the server (the trapdoor will once again be in the horizontal orientation).

peterbell10:
Fixes #4831

The redstone sim should only close doors/trapdoors on the falling edge of a redstone signal, not whenever there is no signal.

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/DoorHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/DoorHandler.h
@@ -40,9 +40,11 @@ namespace DoorHandler
 		Power = std::max(Power, Callback.Power);
 
 		cChunkInterface ChunkInterface(a_Chunk.GetWorld()->GetChunkMap());
+		// Use redstone data rather than block state so players can override redstone control
+		const auto Previous = DataForChunk(a_Chunk).ExchangeUpdateOncePowerData(a_Position, Power);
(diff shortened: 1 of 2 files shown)


tigerw:
Thanks! Do you know if this handles players powering the trapdoor, then closing it manually?

peterbell10:
I managed to simplify the logic a bit further. Turns out we don't really care what state the door is actually in, just the redstone signal.

>Do you know if this handles players powering the trapdoor, then closing it manually?

This new version allows players to close doors opened by redstone but it will reopen on server restart because it goes from unpowered to powered.

tigerw:
Makes sense. I notice that Mojang have added a `powered` state by 1.16, probably to fix this exact issue. As the state gets saved to disk we should be able to eliminate most uses of ExchangePowerData.

Thanks again!

PR #4834, opened 2020-08-26
Fixes windows hack in Compile.sh

Azurethi:
- ``exit`` replaced with ``goto :EOF``
- added ``CLS`` to clear console spam from non-windows commands

Exit causes the entire windows CMD to exit before the user can actually see the message you wanted to print. You could just remove the exit command, but then the last closing brace is interpreted as command causing a spammy "command not recognised" message. ``goto :EOF`` jumps to then end of the file, cleanly returning to the CMD.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -516,9 +516,10 @@ exit 0
 # Called via hack in line 2.
 :windows_detected
 @echo off
+cls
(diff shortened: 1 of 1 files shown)


madmaxoft (approved):

madmaxoft:
Awesome, thanks.

PR #4836, opened 2020-08-26
Changed tree-growing functions in cWorld to use Vector3i coords.

madmaxoft:
The old-style functions are kept in Lua API for backward compatibility, as DeprecatedBindings.

### DIFF
--- a/src/World.h
+++ b/src/World.h
@@ -829,30 +829,25 @@ public:
 	Returns true if the tree is imprinted successfully, false otherwise. */
 	bool GrowTreeImage(const sSetBlockVector & a_Blocks);
 
-	// tolua_begin
-
 	/** Grows a tree at the specified coords.
 	If the specified block is a sapling, the tree is grown from that sapling.
 	Otherwise a tree is grown based on the biome.
-	Returns true if the tree was grown, false if not (invalid chunk, insufficient space). */
-	bool GrowTree(int a_BlockX, int a_BlockY, int a_BlockZ);
+	Returns true if the tree was grown, false if not (invalid chunk, insufficient space).
+	Exported in DeprecatedBindings due to the obsolete int-based overload. */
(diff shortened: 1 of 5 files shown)


src/World.h
```diff
@@ -829,30 +829,25 @@ class cWorld  // tolua_export
...
+	Returns true if the tree was grown, false if not (invalid chunk, insufficient space).
+	Exported in DeprecatedBindings due to the obsolete int-based overload and obsolete additional SaplingMeta param. */
+	bool GrowTreeFromSapling(Vector3i a_BlockPos);
 
 	/** Grows a tree at the specified coords, based on the biome in the place.
-	Returns true if the tree was grown, false if not (invalid chunk, insufficient space). */
-	bool GrowTreeByBiome(int a_BlockX, int a_BlockY, int a_BlockZ);
+	Returns true if the tree was grown, false if not (invalid chunk, insufficient space).
+	Exported in DeprecatedBindings due to the obsolete int-based overload. */
+	bool GrowTreeByBiome(const Vector3i a_BlockPos);
```
tigerw, line 848: Could I trouble you to remove the const in the header, also for GrowTree, other files only tag arguments const in cpp files.

tigerw:
Looks good to me! Was that also a bugfix with the `a_Meta & meta` to `meta & 0x07`?

src/World.h
madmaxoft: Alright.

madmaxoft:
Yes, the meta change is a bugfix. I believe this is what the original author wanted, but failed to implement - the sapling type inside the meta should be the same.

madmaxoft:
This was supposed to be squash-merged :P

tigerw:
Sorry, didn't notice the "Remove const"

But on the other hard, it's never a bad time
```
(•_•)
( •_•)>⌐■-■
(⌐■_■)
````
to see more _flexibility_

PR #4837, opened 2020-08-26
Add more statistic tracking

12xx12:
This PR adds more tracked statistics to cuberite

### DIFF
--- a/src/Mobs/PassiveMonster.h
+++ b/src/Mobs/PassiveMonster.h
@@ -60,6 +60,8 @@ protected:
 	/** The monster's breeding partner. */
 	cPassiveMonster * m_LovePartner;
 
+	cPlayer * m_Feeder;
+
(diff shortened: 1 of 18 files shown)


src/Mobs/PassiveMonster.h
```diff
@@ -60,6 +60,9 @@ class cPassiveMonster:
 	/** The monster's breeding partner. */
 	cPassiveMonster * m_LovePartner;
 
+	/** Remembers the player is was last fed by for statistics tracking */
+	cPlayer * m_Feeder;
```
tigerw, line 64: We shouldn't store a raw pointer to a cPlayer since the player may disconnect, could you store a UUID instead and be like cWorld::DoWithPlayerByUUID?

12xx12: Yes. I just need the stat manager of the player. The way it's obtained doesn't matter.

12xx12: fixed - added also the cow breeding achievement wile at it

Issue #2324, opened 2015-07-09
TnT Chunk generator overloaded -> Deadlock


Sonorpearl:
Hey guys,

to test out the performance I tried to let 1k TnT Explode. On Bukkit/Spigot it aint any Problem, my Server in Java can hold up to 50k TnT.

So I created 1k TnT with WorldEdit and detonated it

     [10:17:07] Chunk generator performance: 0.46 ch / sec (17 ch total)
     [10:17:34] Chunk generator performance: 0.37 ch / sec (24 ch total)
Info [10:17:34] Player Sonorpearl has left the game
     [10:17:35] Chunk generator performance: 1.32 ch / sec (87 ch total)
     [10:17:36] Chunk generator performance: 3.01 ch / sec (205 ch total)
     [10:17:38] Chunk generator performance: 4.66 ch / sec (327 ch total)
     [10:17:39] Chunk generator performance: 6.11 ch / sec (441 ch total)
     [10:17:41] Chunk generator performance: 7.51 ch / sec (557 ch total)
     [10:17:42] Chunk generator performance: 8.95 ch / sec (682 ch total)
     [10:17:44] Chunk generator performance: 10.15 ch / sec (794 ch total)
     [10:17:45] Chunk generator performance: 11.27 ch / sec (904 ch total)
     [10:17:47] Chunk generator performance: 12.40 ch / sec (1020 ch total)
Warn [10:17:47] Chunk generator overloaded, skipping chunk [-27, 18]
Warn [10:17:47] Chunk generator overloaded, skipping chunk [-17, 19]
Warn [10:17:47] Chunk generator overloaded, skipping chunk [9, 18]
Warn [10:17:47] Chunk generator overloaded, skipping chunk [33, 11]
Warn [10:17:47] Chunk generator overloaded, skipping chunk [36, 12]
Warn [10:17:47] Chunk generator overloaded, skipping chunk [37, 12]
Warn [10:17:48] Chunk generator overloaded, skipping chunk [-15, -21]
Warn [10:17:48] Chunk generator overloaded, skipping chunk [-16, -21]
Warn [10:17:48] Chunk generator overloaded, skipping chunk [-18, -24]
Warn [10:17:48] Chunk generator overloaded, skipping chunk [-19, -24]
... (200 Times more)

After about 60 Seconds (Set Interval to that)

Err  [10:23:47] Deadlock detected, aborting the server
Err  [10:23:47]   D:    | MCServer has encountered an error and needs to close
[...]

PR #4845, opened 2020-09-03
Use tracing for explosions

### LINKED ISSUE #1219: Explosion/TnT bugs
- [ ] A double chest stays as a double chest when one of the individual chest blocks that comprise it get destroyed in an explosion and it also keeps its inventory.
- [ ] When placing a block where the second part of the double chest was, all items that where in that half drop on the ground and the chest becomes a single chest.
- [ ] The upper half of a door doesn't get destroyed sometimes and you can open/close it, you cannot destroy it in survival and creative mode.
- [ ] Signs and ladders ignores when the block where there where placed on gets destroyed in an explosion.
- [ ] you can place sings on tnt which gets not destroyed when the tnt is activated
- [ ] explosions can pass. trough obsidian and destroyed blocks get on the top on the obsidian
- [ ] enderchest, anvil and enchanment table gets deroyed from explosions
- [ ] a bed didn't get destroyed fully when one half gets affected by an explosion
- [ ] iron bars sometime ignores explosions
- [ ] tnt can pass trough block whens its activated and moves a few block freely in a random direction

sorry that I'm too lazy to create a new issue for each bug.
I hope that it doesn't include  any duplicate :)

tested with rev bc94ad4b5ae089fe5ee984a827a0575926102bdc (2014_07_18) from my nightbuilds

### LINKED ISSUE #2396: Ghast Blasts
A Ghast Blast destroys cobblestone. It shouldn't.

### LINKED ISSUE #3194: Creeper explosions cause grass blocks to drop themselves rather than dirt blocks

### LINKED ISSUE #3863: Some blocks of multi-block structures stay after explosions

### LINKED ISSUE #4527: Improve explosion knockback

tigerw:
Fixes #1219
Partially addresses #2324
Fixes #2396
Fixes #3194
Fixes #3863
Fixes #4527 (possibly)


src/ClientHandle.cpp
```diff
@@ -2647,7 +2647,27 @@ void cClientHandle::SendExplosion(const Vector3d a_Pos, float a_Radius, const cV
...
+	// Dark smoke particles:
+	SendParticleEffect("largesmoke", a_Position.x, a_Position.y, a_Position.z, 0, 0, 0, Spread, static_cast<int>(ParticleCount));
+
+	Spread = ParticleFormula * 0.35f;
+	ParticleCount = std::min(static_cast<int>(ParticleFormula * 550), 1800);
+
+	// Light smoke particles:
+	SendParticleEffect("explode", a_Position.x, a_Position.y, a_Position.z, 0, 0, 0, Spread, static_cast<int>(ParticleCount));
+
+	// Macromedia Shockwave effect:
```
madmaxoft, line 2669: Please refrain from using puns in the code, they make it difficult to understand for people without the specific background information.

src/Entities/TNTEntity.cpp
```diff
@@ -33,10 +33,13 @@ void cTNTEntity::SpawnOn(cClientHandle & a_ClientHandle)
...
 {
-	m_FuseTicks = 0;
-	Destroy();
 	FLOGD("BOOM at {0}", GetPosition());
-	m_World->DoExplosionAt(4.0, GetPosX(), GetPosY(), GetPosZ(), true, esPrimedTNT, this);
+
+	// Destroy first so the Explodinator doesn't find us (when iterating through entities):
+	Destroy();
+
+	// TODO centre everything else too
```
madmaxoft, line 41: "Everything else"? Like what? The X and Z coords? (already centered)

src/Physics/Explodinator.cpp
```diff
@@ -0,0 +1,405 @@
...
+namespace Explodinator
+{
+	const auto StepUnit = 0.3f;
+	const auto KnockbackFactor = 25U;
+	const auto StepAttenuation = 0.225f;
+	const auto TraceCubeSideLength = 16U;
+	const auto BoundingBoxStepUnit = 0.5f;
+
+	/** Returns how much of a Explosion Lazor's (tm) intensity the given block attenuates. */
+	static float GetAbsorption(const BLOCKTYPE Block)
```
madmaxoft, line 25: Style (everywhere in this file, apparently): Functino arguments should have the `a_` prefix.

src/Physics/Explodinator.cpp
```diff
@@ -0,0 +1,405 @@
...
+			case E_BLOCK_BREWING_STAND:
+			case E_BLOCK_STONE_BUTTON:
+			case E_BLOCK_WOODEN_BUTTON:
+			case E_BLOCK_CAKE:
+			case E_BLOCK_CONCRETE_POWDER:
+			case E_BLOCK_DIRT:
+			case E_BLOCK_FROSTED_ICE:
+			case E_BLOCK_HAY_BALE:
+			case E_BLOCK_ICE: return 0.5f;
+			default: return 0;
```
madmaxoft, line 138: This should be replaced with `cBlockInfo`, so that it's easier to move to the 1.13+ format later on.

src/Physics/Explodinator.cpp
tigerw: It looks like this though so an extra tag is superfluous I think, I don't suppose you'd consider a variation?

Argument highlighting

src/Physics/Explodinator.cpp
```diff
@@ -0,0 +1,405 @@
...
+		const auto SquareRadius = Radius * Radius;
+
+		Chunk.GetWorld()->ForEachEntityInBox({ Position, Radius * 2 }, [&Chunk, Position, Power, Radius, SquareRadius](cEntity & Entity)
+		{
+			// Percentage of rays unobstructed.
+			const auto Exposure = CalculateEntityExposure(Chunk, Entity, Position, SquareRadius);
+			const auto Direction = Entity.GetPosition() - Position;
+			const auto Impact = (1 - (static_cast<float>(Direction.Length()) / Radius)) * Exposure;
+
+			// Don't apply damage to other TNT entities and falling blocks, they should be invincible:
```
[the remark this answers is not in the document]
tigerw: I do still need to apply knockback based on the exposure, would you like me to add a comment for this?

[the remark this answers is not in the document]
bearbin: Oh, I forgot that knockback depends on exposure, nevermind then.

src/Physics/Explodinator.cpp
madmaxoft: Maybe if you consider changing to javaStyleIdentifierNames :P
Not everyone has such a smart IDE; GitHub definitely doesn't when reviewing the PRs.

tigerw:
Parameters renamed as xoft suggests, merge?

src/Physics/Explodinator.cpp
```diff
@@ -0,0 +1,287 @@
...
+		{
+			const auto Below = a_Position.addedY(-1);
+			if ((Below.y >= 0) && cBlockInfo::FullyOccupiesVoxel(a_Chunk.GetBlock(Below)))
+			{
+				// Start a fire:
+				a_Chunk.SetBlock(a_Position, E_BLOCK_FIRE, 0);
+				return;
+			}
+		}
+		else if (const auto Shrapnel = World.GetTNTShrapnelLevel(); (Shrapnel > slNone) && Random.RandBool(0))  // 20% chance of flinging stuff around
```
bearbin, line 144: Nice, C++ finally has scoped declarations in if-expressions!

src/Physics/Explodinator.cpp
```diff
@@ -0,0 +1,287 @@
...
+				Absolute
+			);
+		}
+		else if (a_Fiery && Random.RandBool(1.f / 3.f))  // 33% chance of starting fires if it can start fires
+		{
+			const auto Below = a_Position.addedY(-1);
+			if ((Below.y >= 0) && cBlockInfo::FullyOccupiesVoxel(a_Chunk.GetBlock(Below)))
+			{
+				// Start a fire:
+				a_Chunk.SetBlock(a_Position, E_BLOCK_FIRE, 0);
```
bearbin, line 140: Do we also need to call OnBroken here?

src/Protocol/Protocol_1_8.cpp
```diff
@@ -620,25 +620,19 @@ void cProtocol_1_8_0::SendExperienceOrb(const cExpOrb & a_ExpOrb)
 
 
 
-void cProtocol_1_8_0::SendExplosion(double a_BlockX, double a_BlockY, double a_BlockZ, float a_Radius, const cVector3iArray & a_BlocksAffected, const Vector3d & a_PlayerMotion)
+void cProtocol_1_8_0::SendExplosion(const Vector3f a_Position, const float a_Power)
```
bearbin, line 623: Can you remind me why we don't need BlocksAffected anymore?

bearbin:
I looked through one more time - once the above are resolved this is actually ready to merge.

src/Physics/Explodinator.cpp
```diff
@@ -0,0 +1,287 @@
...
+				auto FallingBlock = std::make_unique<cFallingBlock>(Vector3d(0.5, 0, 0.5) + Absolute, DestroyedBlock, DestroyedMeta);
+				// TODO: correct velocity FallingBlock->SetSpeedY(40);
+				FallingBlock->Initialize(std::move(FallingBlock), World);
+			}
+		}
+
+		// SetBlock wakes up all simulators for the area, so that water and lava flows and sand falls into the blasted holes
+		// It also is responsible for calling cBlockHandler::OnNeighborChanged to pop off blocks that fail CanBeAt
+		// An explicit call to cBlockHandler::OnBroken handles the destruction of multiblock structures
+		// References at (FS #391, GH #):
```
tigerw, line 161: oops, add GH issue reference

src/Protocol/Protocol_1_8.cpp
tigerw: The explosion packet has a field to tell it blocks to set to zero ([ref](https://wiki.vg/Protocol#Explosion)) but since SetBlock already sends this to the client, I wanted to avoid duplication, which should help reduce network traffic. As well avoiding vector allocations in explosion calculation.

tigerw:
Your comments: 0xADD12E55ED

bearbin:
All looks good, but _please_ don't keep force-pushing. It makes it extremely difficult to see what changes you've made and how that matches up with the previous version that the comments were attached to, which github then refuses to show.

The PR will be squished nicely at the end, so no need to maintain a completely clean commit history in the meantime.

bearbin (approved):

tigerw:
oh I rebased onto master, that would be why everything was changed

bearbin:
Yep, better to just merge master into the branch rather than rebase and force push, at least from the perspective of using the GH review and merge system.

Issue #550, opened 2014-01-17
Implement a separate chunk server


madmaxoft:
cChunkGenerator can be reused in a separate project now, so it is possible to create a separate application, a server that will generate chunks requested by the main MCServer process. This chunkserver can run on the same machine, or even over the network. It needs to receive the world configuration in the form of the world.ini file, then it will just process requests for chunks, sending the chunk data back once generated.

As the first step, implement the chunk server process only, without modifying MCServer, and also implement a small testing application that will load an ini file, send it to the chunk server, then send requests for chunks and receive the chunk data. Then we shall evaluate the overall performance and the integration strategies.

madmaxoft:
Looking back to this, I don't believe this should not be done at all.

Plugins are supposed to be able to change the generated chunks, and we're planning API injection points into the generator itself, so that a plugin can provide one of the generator modules. This would completely break standalone chunk servers.

Instead, we should focus the effort into making the chunk generator encapsulated and thread-safe, so that multiple instances can be created for a single world and chunk generation moved to a threadpool.

Issue #4759, opened 2020-06-15
Making World Generation look more like vanilla


Titaniumtown:
So, I was looking into how one could make cuberite world gen more like vanilla. I found a project (https://github.com/Cubitect/cubiomes) that aims to port vanilla like world gen to C. It has various options, including toggling between different versions of minecraft's generation. It even seems to have 1.16 generation figured out. 

Thoughts?

Edit: it seems after further reading, it doesn't generate a world file. But cords where biomes and structures are.

Titaniumtown:
bump, is this an avenue that would be worth looking into?

bearbin:
Please don't 'bump' issues. That looks like a relatively interesting project, but it's only going to get integrated if somebody turns up and actually wants to do it.

Titaniumtown:
ok sorry. I'm looking into it now.

Titaniumtown:
I don't think I have the know how to do this. lol

bearbin:
From my cursory look at it, it seems to only do the biome generation,
and structure placement. The actual 'blocks' don't seem to be something
it's capable of doing. Luckily the cuberite generator is similarly
modular, so to integrate it all that would be necessary would be to
write a biome generator/finisher including it as a library and calling
into it at the relevant places.

I don't have any particular desire to do it myself, but it seems like it
shouldn't be too difficult (although problems will likely pop up...)

A bigger issue would be licensing. cubiomes is GPL licensed which
precludes its inclusion in cuberite with the Apache license. Any
combination of cuberite with cubiomes would end up as GPL as the GPL is
viral. I have no desire to make Cuberite GPL and resultantly impose
additional restrictions on users; neither would there be a desire to
implement some sort of complicated license-avoiding system like whatever
the current successor to CraftBukkit has.

Unfortunately the poor choice of license on their part puts a stop to
things.

Titaniumtown:
Aww that sucks. I'll close the issue then :( It sucks that licensing gets in the way of things.

bearbin:
I have submitted an issue on cubiomes about the license, let's see what
they say: 

https://github.com/Cubitect/cubiomes/issues/34

Titaniumtown:
oooh that would be neat

Titaniumtown:
keep me posted!

Titaniumtown:
Now that the License has been changed. Work can be started on implemented cubiomes in cuberite, right?

madmaxoft:
[...]

PR #4846, opened 2020-09-03
PoC: ChunkHTTPServer, toward a separate chunk server

astralien3000:
Hello,

I was wandering in cuberite issues when I found #550. Then I thought "Hey ! I can do that !".

# Design notes

As what I understood from #550 and the code, the chunk generation code is already quite well separated from the rest. And especially, `cChunkGenerator` is only used by `cChunkGeneratorThread`, so that it almost could be put in a library if `cChunkGenerator` was not using things like Entities/BlockEntities/etc...

To me, the biggest problem if you wanted to put the chunk generator in an other process or even in an other server was the transmission of chunks. There are three solutions : 

 - Use the Minecraft protocol
 - Use the HTTP protocol
 - Use another protocol

Since using another protocol which is not yet used by the server is not a good idea, and I do not know a thing about the Minecraft protocol, HTTP seemed to be good to go.

The other problem was how to encode chunks on the network. I can guess the Minecraft protocol has a way to encode, but since I don't know a thing about it, I did not try too much. Instead, I chose `cereal`, which is a serialization library that I know well and is easy to use. It can serialize data in JSON, XML, Binary and others. I used JSON for debugging and Binary for the real thing.

And here I am with a `ChunkHTTPServer` that (should) mimics the `cChunkGenerator` API over HTTP, and a `ChunkHTTPClient` that (should) replace the `cChunkGeneratorThread`.

# The Proof of Concept

The current proof of concept is able to spawn one `ChunkHTTPServer` for each world. The port is supposed to be configurable in the world's ini file, but the client use hardcoded URLs currently...

Each world also spawn a `ChunkHTTPClient` that replace `cChunkGeneratorThread` (`ChunkHTTPClient` is a copy of `cChunkGeneratorThread`, but with a modified `DoGenerate` function).

Currently the process that serves the chunks is the same as the process that download it, so that there is no performance improvement.

The `ChunkHTTPServer` also does not expose the complete `cChunkGenerator` API, and the `ChunkHTTPClient` still use a "local" chunk generator to do some operations exactly like `cChunkGeneratorThread`.

I did a very basic serialization of chunks, which lacks a lot of things. For example I don't serialize Entities and BlockEntities. Some blocks looked weird when generated (grass becomes dead_bushes...).

# Future

 - implement full cChunkGenerator API
 - completely serialize chunks
 - make ChunkHTTPServer "stadalone"

# What do you think ?

I come with this pull request to know if my ideas are doable / interesting enough to put work on it. Maybe there will be something that will prevent me to complete this system.


tigerw:
This is pretty cool! I would propose using the existing [Chunk Data](https://wiki.vg/Protocol#Chunk_Data) format but I think an out-of-process chunk server could really be a standout feature.

NiLSPACE:
Wouldn't it be possible to create a completely new [Generator](https://github.com/cuberite/cuberite/blob/master/src/Generating/ChunkGenerator.cpp#L36) which would simply send a request to a server? You wouldn't have to worry about `cChunkGeneratorThread` that way.

madmaxoft:
I have to say that I don't believe this should not be done at all.

Plugins are supposed to be able to change the generated chunks, and we're planning API injection points into the generator itself, so that a plugin can provide one of the generator modules. This would completely break standalone chunk servers.

As for your implementation details, bringing in another library just so that you can serialize some data feels like an overkill, the CI compile times are already insane so throwing yet another library into the mix is bad. There's actually a simpler way to serialize, using `cBlockArea`.

astralien3000:
> I would propose using the existing [Chunk Data](https://wiki.vg/Protocol#Chunk_Data) format 

> As for your implementation details, bringing in another library just so that you can serialize some data feels like an overkill, the CI compile times are already insane so throwing yet another library into the mix is bad. There's actually a simpler way to serialize, using cBlockArea.

Ah yes I understand better, I can transform a cChunkDesc into a cBlockArea and then use cSchematicFileSerializer to serialize it. I guess cBlockArea holds all data needed for a complete cChunkDesc ? As I said, I used cereal because I did not know much of the Minecraft protocole (and data serialization) and the code base that implements it.

> Wouldn't it be possible to create a completely new [Generator](https://github.com/cuberite/cuberite/blob/master/src/Generating/ChunkGenerator.cpp#L36) which would simply send a request to a server? You wouldn't have to worry about `cChunkGeneratorThread` that

Hmm, I was afraid that since the generator API was "synchronous"(/blocking), it would not work with an HTTP request handling. But I'm currently using it as synchronous, so it may be better to do it the way you said ! Should be even easier.


> I have to say that I don't believe this should not be done at all.

It should be possible to make the separated chunk server optional, especially if I do what @NiLSPACE said.

> Plugins are supposed to be able to change the generated chunks, and we're planning API injection points into the generator itself, so that a plugin can provide one of the generator modules. This would completely break standalone chunk servers.

Plugins that are able to change the chunk after it is generated should not be affected. Indeed, it can still be executed by the main server just after downloading from the chunk server.

About API injection points into the generator, I did not know that. I guess sharing the plugins between a main server and a chunk server may not be possible because of any global variable in the plugins that would require some sort of communication...
What is the benefit of modifying the generators instead of just being able to change the chunk just after it's generated ? Won't the lua part of the generation be quite slow ? Hence benefiting from being outsourced ^^ ?

Also, depending on what the entry points would be, It may still be possible to execute some part of the generation externally, and some part on the main server ?

DaMatrix:
If you do make it as an ordinary generator, why not just do it as a plugin?

astralien3000:
@DaMatrix, I'm not sure to understand what you said. I don't think plugins can do what I'm trying to do currently. They have some hooks before and after generation, but can't replace it.

astralien3000:
Ok I got the serialization with `cSchematicFileSerializer` working, it's better !
But it looks like generated entities are not serialized, are they ?

madmaxoft:
I guess we never got around to implementing actual entities in `cBlockArea` et al.

astralien3000:
Hello !

Some little news since last time. I have implemented what @NiLSPACE suggested, now if an optional Key (ChunkGeneratorHTTPClient) with some configs (host, port, name) is present is the wolrd.ini file, the ChunkGeneratorHTTPClient will be spawned instead of a "normal" one. Then the ini file is sent to the server to configure the generator there.

Also, the server is able to manage several generators at the same time (one per world, for example), so that there will be no need to spawn one server per world. Yet, I still have to separate the server from the cWorld, but it should be easy now.

One new "problem" I have is that, now I implement the whole Generator API, the GenerateBiomes function is called several times for the same chunk. It seems it's called at a regular time interval. Why ? Is it necessary to call GenerateBiomes so much ? I added a Cache system to ChunkGeneratorHTTPClient to avoid unnecessary HTTP requests.

I also did not deleted the cereal library, I will be working on it.

Cheers

astralien3000:
Ok I have found a way to fully serialize cChunkDesc (with entities and all), but this is DIRTY AF !!!

If I understand everything well, the only class that can serialize entities and blockentities is `NBTChunkSerializer`, and the only class that can unserialize entities and blockentities is `cWSSAnvil`... So I serialized cChunkDesc as if it was saved into a file.

The problem is that `cWSSAnvil` need a `cWorld` to work, so that I needed to create a Dummy World with a Dummy WSSAnvil. Which looks to me overkill since (de)serialization should not need that.

astralien3000:
Hey, I think this PR is ready for reviews.

All features are here : 
 - ChunkGenerator API is completely implemented via HTTP
 - Serialization of chunks is complete (with Entities and BlockEntities), no external lib used.
 - The server is... "kind of" standalone : creating a new exectuable from scratch would have been very complicated, but It is an optional feature of the server (with option --chunk-server). It can be considered as standalone if you start a server with the --chunk-server option, and an other without, that will connect to the first one.

astralien3000:
Finally... Looks like the Windows-x64-Debug did not liked my CLI addition.

astralien3000:
Hello,

I have had no news for 1 month, I just wanted to ask. Should I expect this PR to be reviewed soon ? Or should I expect it to be closed ? Or blocked ?

Cheers

bearbin:
Hi @astralien3000, sorry that this PR has sort of sat around with nothing going on. I don't know if @tigerw or @madmaxoft have any thoughts on this since they commented before?

In any case, I'll try and have a look at this and give some constructive comments over the weekend.

src/ChunkHTTPServer.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "WorldStorage/FastNBT.h"
+#include "StringCompression.h"
+
+#include <regex>
+
+
+
+
+
+static const char DEFAULT_CHUNCKHTTPSERVER_PORTS[] = "5555";
```
bearbin, line 20: There's a typo here: this should be CHUNK as opposed to CHUNCK

src/ChunkHTTPServer.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+
+
+
+
+void ChunkHTTPServer::OnRequestBody(cHTTPServerConnection & a_Connection, cHTTPIncomingRequest & a_Request, const char * a_Data, size_t a_Size)
+{
+	UNUSED(a_Connection);
+
+	LOGD("OnRequestBody (%lld)", (long long)a_Size);
```
bearbin, line 132: Please try not to use builtin integer types aside from int/unsigned for general use and size_t for sizes. Instead use the defined IntXX or UIntXX types.

src/Generating/ChunkGeneratorHTTPClient.cpp
```diff
@@ -0,0 +1,244 @@
...
+		auto Data = std::make_shared<AString>();
+		auto callbacks = std::make_unique<ChunkGeneratorHTTPClientCallbacks>(evtFinished, Data);
+		auto res = cUrlClient::Get(URL, std::move(callbacks));
+
+		if (res.first)
+		{
+			evtFinished->Wait();
+			LOGD("Response size : %lld", (long long)Data->size());
+
+			AString UncrompressedData;
```
bearbin, line 153: Uncrompessed != Uncompressed

src/Generating/ChunkGeneratorHTTPClient.cpp
```diff
@@ -0,0 +1,244 @@
...
+			UncompressStringGZIP(Data->data(), Data->size(), UncrompressedData);
+			cParsedNBT NBT(UncrompressedData.data(), UncrompressedData.size());;
+			NBTChunkSerializer::deserializeBiomeMap(m_GenerateBiomesCache[a_ChunkCoords], NBT);
+		}
+		else
+		{
+			LOGD("Immediate error: %s", res.second.c_str());
```
bearbin, line 160: This should probably give a more informative message than "Immediate error"

src/HTTP/HTTPServer.cpp
```diff
@@ -163,7 +163,7 @@ bool cHTTPServer::Start(cCallbacks & a_Callbacks, const AStringVector & a_Ports)
 		}
 		reportPorts.append(port);
 	}
-	LOGINFO("WebAdmin is running on port(s) %s", reportPorts.c_str());
+	LOGINFO("HTTPServer is running on port(s) %s", reportPorts.c_str());
```
bearbin, line 166: It would be a good idea to identify WebAdmin a bit better in the logs than just HTTPServer - maybe this is another area that needs refactoring.

src/IniFile.cpp
```diff
@@ -186,6 +187,103 @@ bool cIniFile::ReadFile(const AString & a_FileName, bool a_AllowExampleRedirect)
...
+				{
+					AddKeyComment(keyname, line.substr(pLeft + 1));
+				}
+				break;
+			}
+		}  // switch (line[pLeft])
+	}  // while (getline())
+
+	return !(m_Keys.empty() && m_Names.empty() && m_Comments.empty());
+}
```
bearbin, line 281: I'm guessing this is just Copy-Paste with file-IO operations changed to string-IO operations?

If so, this wouldn't pass the DRY sniff-test and would need refactoring before merge (even something as simple as reading the whole file into a string, and passing that through to this would probably be fine).

src/IniFile.h
```diff
@@ -61,6 +61,16 @@ class cIniFile:
...
+
+	/** Reads the contents of the specified string
+	Returns true if successful, false otherwise. */
+	bool ReadString(const AString & a_Data);
+
+	/** Writes data stored in class to the specified string.
+	Returns true on success, false on failure. */
+	bool WriteString(AString & a_Data) const;
+
 	// tolua_begin
```
bearbin, line 74: C++ classes can have _two_ public sections in the same header file? Well it seems so at least, but this would better be put under the extant section just below. 

And these two functions seem like they could be useful exported to the API as well, so no problem with them being inside the `tolua_begin` section.

bearbin (commented):
This looks like a clean first step towards allowing seperate chunk servers.

I don't think this is actually very useful to a general user, but I can imagine some cases where this makes sense, and the maintenance overhead seems quite low to me. If this does cause any issues, it'll probably be for the better in ensuring our generator architecture stays decoupled. I think this concept can be merged in, although the architecture is not quire right yet.

In terms of the architecture, I would prefer that there was a seperate chunk-server instance for each generator, rather than having a single instance for all generators.
Correspondingly, the configuration for the generators should be at the chunk-server side, rather than at the client. Certainly the chunk-serveryness of each world should be configured indepdenently.

I guess my perspective is that the server is not just a work-offload daemon, but rather a self-contained service that gives chunks generated with a specific generator.

This would be beneficial for the use-case of very expensive world generation. In that case, there could be many generators sitting behind a load balancer, and having less coupling is good.

You also don't seem to turn off the regular server functionality when in chunk-server mode. That seems like a very odd choice, the chunk servers should only expose the HTTP port, and not anything else. Dual-use binaries are fine, but the server should be only one of chunk or player at once.

GZIP-compression of chunk data seems like a bit an anti-feature to me, have you tested if it's actually useful or if it just burns CPU for no reason? It could be useful to add a configuration option here if it is actually necessary for low-bandwidth connections.

Can this implementation have multiple clients connected to a single chunk-server? That would be nice if so. Also, what happens in the case of network error or instability - this should be handled gracefully and needs testing.

I've also added a few comments on specific areas or code style - in addition to them, the whole PR needs a search-and-replace done for "Chunck" being replaced with "Chunk".

src/ChunkHTTPServer.cpp
astralien3000: While I agree with you concerning the use of (u)intXX, concerning the printf format : 
"%lld" = long long int decimal
Even if I know that long long always correspond to int64_t.
I just can't make a mistake while writing this way.

Anyway this log is not so useful... Maybe it's better to remove it.

src/HTTP/HTTPServer.cpp
astralien3000: The problem is that this code is in `HTTPServer.cpp `.
Even if this was not confusing because the only HTTPServer was WebAdmin, this is not really clean.

src/IniFile.cpp
astralien3000: yes, the only modification from the "file version" is the beginning and the end of the function, so that the middle can be factorized. TODO

src/IniFile.h
astralien3000: I could not make the tests pass because of an error about documentation which was non-existent. So, in a first step I added the comments you can see, but it did not work. I was unable to understand what the test tool wanted, but I understood it was "because" of the lua API. I finally moved the declarations to not be bothered by this, since adding functions to the lua API is not my concern.

And yes, C++ can have their "sections" declared several times. I use it a lot to separate `private data` from `private functions` from `private constants`, etc... Same for `public members` and `public constructors/destructors` and `public virtual members`...
But these are not real "sections", these are "access modifiers", exactly like in Java for example, except that it is applied to several lines instead of only one in Java.

bearbin: Ah, this is just an organisational pattern that I had not seen before, I should really try to use the technical terms, of course I do know the function of the access modifiers :)

The documentation for lua needs to be added to the Server/Plugins/APIDump/APIDesc.lua file

astralien3000:
> I don't think this is actually very useful to a general user, but I can imagine some cases where this makes sense, and the maintenance overhead seems quite low to me. If this does cause any issues, it'll probably be for the better in ensuring our generator architecture stays decoupled. I think this concept can be merged in, although the architecture is not quire right yet.

To be completely fair, my (wet) dream (did I really said that ?!?) is to have a distributed minecraft server (in a deeper way than BungeeCord for example, not a distributed MULTIverse, a distributed UNIverse) ^^. But I know it is not a main concern of this project (I even think I have read in the forum that it is not a concern at all). I wanted to do this PR because I knew it was going in this direction and I wanted to start somewhere to understand the code. Now that I have passed months to try to read and understand the code, this final aim seems INSANELY hard for someone alone. So that I don't know if I will be able to push this project, especially if I can't find any other step that seems actually doable. But hey, I'm silly and like to do hard things so who knows ? ^^

I can completely understand that this kind of PR can be useless for a majority and even become harmful if it implies maintenance costs. That's why I tried to make it as seamless as possible. That is also why I opened this PR early in the development process and was ready to see it closed or close it myself it it was blocked. But at least I feel happy to read that this PR can be useful even when not considering my distributed aspect.


> In terms of the architecture, I would prefer that there was a seperate chunk-server instance for each generator, rather than having a single instance for all generators.

This can be done in the current state, you can spawn a process with a chunk server for each world, and configure each world ChunkGeneratorHTTPClient with a different host/port.
But a single server can also generate for several worlds. (by adding the "name" endpoint to host and port)
What can do more can do less.


> Correspondingly, the configuration for the generators should be at the chunk-server side, rather than at the client. Certainly the chunk-serveryness of each world on the client should be configured indepdenently.

Chunk-server side configuration seems quite complicated form the current state, the generation configuration seems to be alongside other configurations of the world in the Ini file. So that separating that would induce a big modification for the user (and the developer) and I did not wanted that. Especially if you want backward compatibility.

Letting the client send it's configuration to the server, from my point of view, seems almost cost-less (it's done only once at startup), super-flexible (you can create several endpoints in the same server, at will, and without restarting anything) and was quite easy to implement (I just needed to expose the `Initialize` function alongside `Generate` and `GenerateBiomes`, and use a std::map to map names to generators/configs).

And actually each world can already choose it's generator, so that the Overworld can be externally generated by a ChunkServer while the Ender (maybe less used) can be generated locally as before my PR.


> I guess my perspective is that the server is not just a work-offload daemon, but rather a self-contained service that gives chunks generated with a specific generator.

I'm not really sure what is the difference you make between these two concepts. A generator in itself is an algorithm, which can be configured. But I guess that you think of a "specific generator" as a generating algorithm + it's configuration ? Even with this guess I don't really understand what this implies. Are we only talking about the way the server gets it's configuration ? File or Client request ?


> You also don't seem to turn off the regular server functionality when in chunk-server mode. That seems like a very odd choice, the chunk servers should only expose the HTTP port, and not anything else. Dual-use binaries are fine, but the server should be only one of chunk or player at once.

Good catch, I actually tried two things before making this choice (that I don't liked at first, but it can be justified).

First I tried to create a separate executable. But doing this was not manageable (at least for me alone) because of the CMakeFiles modifications it would implies. Not mentioning the compilation time that may be basically doubled (I heard compilation time was already a problem here ^^).

Then I tried to create a `--chunk-server-only` cli option. But I was unable to disable other services (Minecraft Server, Webadmin, ...) without getting segfaults everywhere. I just don't understand enough the architecture to be able to do this currently.

So I decided to think "cuberite" as a service provider that we can enable or disable. For example in the future, we may introduce options like `--(dis/en)able-webadmin`, `--*able-rcon`, `--*able-minecraft`, and the one that concerns this PR : `--*able-chunk`


> GZIP-compression of chunk data seems like a bit an anti-feature to me, have you tested if it's actually useful or if it just burns CPU for no reason? It could be useful to add a configuration option here if it is actually necessary for low-bandwidth connections.

I basically have no idea of what I'm doing here, I just did as much as possible like other parts of the code. But since it concerned file storage, indeed, this may not be relevant, and even harmful. No idea.


> Can this implementation have multiple clients connected to a single chunk-server? That would be nice if so.

In theory yes, and it's even the normal use of this server because, even if it's the same process, one client that would have it's three worlds externally generated would be considered as 3 different clients. It's also the magic of HTTP (and REST-like APIs) : each request can be considered as the connection of a new client, it just don't care. But I totally rely on the current HTTP server implementation that is used for WebAdmin, and I don't know the details of it's implementation. So I don't know what happens in case of concurrent requests for several worlds (and I did not test, since testing seriously this kind of thing is not straight forward). Maybe at least testing if it does not encounter frequent segfaults could be an "easy" TODO ^^...

> Also, what happens in the case of network error or instability - this should be handled gracefully and needs testing.

For the problem of network instability, it is indeed a big problem that I currently ignored (big TODO here). What I can say is that if the server can't send chunks (because it is not started or because I did not finished it, for example ^^), the client will throw an error (seems like it's mainly about entities spawning in an invalid chunk) and stop.
Even worse, the current ChunkGenerator API does not handle errors in chunk generation explicitly, so that I'm not sure of what I should do. Exceptions ?
I could also implement retries on errors... this is possible I guess.


> I've also added a few comments on specific areas or code style - in addition to them, the whole PR needs a search-and-replace done for "Chunck" being replaced with "Chunk".

Haha sorry for that. I don't know why I make this mistake... Maybe because I confuse with "block" ?

tigerw:
This is definitely valuable: if we host this as an out-of-process, pluggable chunk server, not only would it be possible to serve Cuberite-style generated chunks, but also purely Vanilla generated chunks by somehow hooking/integrating the stock `server.jar` into the chunk server (#4759). I'd expect this to probably go into a separate project in the organisation since it's effectively a standalone executable.

astralien3000:
> This is definitely valuable: if we host this as an out-of-process, pluggable chunk server, not only would it be possible to serve Cuberite-style generated chunks, but also purely Vanilla generated chunks by somehow hooking/integrating the stock `server.jar` into the chunk server (#4759). I'd expect this to probably go into a separate project in the organisation since it's effectively a standalone executable.

Hello, 

ha I finally did not get time nor energy to finish this task... ><

I don't know what you exactly expect to become a separate project, but I was unable to fully separate the chunk generator from the server because there is still a lot of inter-dependencies. Especially with entities/block entities, if I remember well.
But indeed you can use the REST-like API that I created to plug any kind of generator ! That sounds interesting. Of course, I think the API would need some standardization to handle different generators.

PR #4855, opened 2020-09-11
Remove unused Temporary namespace

tigerw:
- Remove unused temporary mapping generation at startup


src/CMakeLists.txt
```diff
@@ -173,7 +165,7 @@ file(WRITE "${CMAKE_BINARY_DIR}/include/Globals.h"
 	"/* This file allows Globals.h to be included with an absolute path */\n#include \"${PROJECT_SOURCE_DIR}/src/Globals.h\"\n")
 
 configure_file("BuildInfo.h.cmake" "${CMAKE_BINARY_DIR}/include/BuildInfo.h")
-target_include_directories(${CMAKE_PROJECT_NAME} PRIVATE "${CMAKE_BINARY_DIR}/include/")
+target_include_directories(${CMAKE_PROJECT_NAME} PRIVATE "${CMAKE_BINARY_DIR}/include/" ".")
```
peterbell10, line 168: ???

peterbell10 (commented):
I haven't been following the Palette stuff, but why Bindings?

tigerw:
I anticipate plugins adding custom blocks dynamically, and it looks like the moved code woudl make a perfect container for plugin blocks with custom properties, so Bindings I guess?

src/CMakeLists.txt
tigerw: ¿¿¿

peterbell10:
If it isn't Lua-related then it doesn't belong in bindings.

tigerw:
Plugin stuff isn't Lua related? xD

Your suggestion?

peterbell10:
Could possibly be used by a plugin != Lua related. By that metric, entities, cRoot and cWorld would all live in the Bindings folder.

tigerw:
Make that exclusively used :)

bearbin:
I'm not sure what the actual reason behind this PR is - just removing temporary stuff doesn't make sense by itself.

I think we need a forum thread going through your roadmap for these PRs and explaining things a bit more as well as clarifying what the ultimate aim is.

tigerw:
OK, I'll write one

tigerw:
[1.13 migration, aims & objectives](https://forum.cuberite.org/thread-3313.html), what do you think?

tigerw:
ready to merge

tigerw:
merging if no further comments

bearbin:
Can you hold off one day? I want to read through your forum thread and properly understand what's going on.

bearbin:
I've read through your post and I now understand what's going on here. Unfortunately I think that this PR is on the wrong track - the BlockState/TypePalette/Registry are __not__ bindings at the moment, and so they shouldn't go in the Bindings folder.

Maybe once the rest of the flattening functionality is implemented, there would be a case for moving these files, either to bindings or to a new directory of their own; but at the moment the right place is right where they are.

tigerw:
please stop closing my pulls... there's more to these changes than just moving files.

bearbin:
Oh all right then... a whole two lines of substantial changes!

They look OK, but you should fully delete the file, rather than simply commenting it out.

tigerw:
Done.

bearbin (approved):

PR #4856, opened 2020-09-11
Prepare for 1.15+

tigerw:


peterbell10:
Is there a reason you include generated code directly in git instead of running the code generator as part of the build system. Doing it this way makes for some atrocious diffs.

tigerw:
The generator needs access to vanilla registries json, I don't know if Mojang would be happy with us bundling those files, and generating them from server.jar at build would mean taking a dependency on Java.

src/BlockState.h
```diff
@@ -0,0 +1,18 @@
...
+{
+	constexpr BlockState(unsigned short ID) :
+		ID(ID)
+	{
+	}
+
+	Block::Type Type() const;
+	bool operator == (Block::Type Type) const;
+	bool operator == (BlockState Block) const;
+	bool operator != (BlockState Block) const;
```
peterbell10, line 15: Prefer non-function operators so you can create the symmetric pairs:
```
bool operator == (A a, B b);
bool operator == (B b, A a);
```

src/Protocol/Palettes/Palette_1_14.h
```diff
@@ -1,12 +1,13 @@
 #pragma once
 
-#include "../../Registries/Items.h"
-#include "../../Registries/Statistics.h"
+#include "BlockState.h"
+#include "Registries/Items.h"
+#include "Registries/Statistics.h"
```
peterbell10, line 5: Include paths should be relative.

peterbell10 (changes requested):

peterbell10:
>The generator needs access to vanilla registries json, I don't know if Mojang would be happy with us bundling those files, and generating them from server.jar at build would mean taking a dependency on Java.

Then I would say either put an intermediate file format into git. Or, at the very least put a giant warning saying this file is auto-generated and shouldn't be edited manually.

src/Protocol/Palettes/Palette_1_14.h
tigerw: could you remind me why? the ../../../ looks unaesthetic.

peterbell10: IDK, I think it was xoft's preference. Either way we shouldn't mix and match, lets stick to how the rest of the code does it.

tigerw: hehe, you are also invoking the principle of consistency I see ;P

src/BlockState.h
```diff
@@ -0,0 +1,18 @@
...
+
+struct BlockState
+{
+	constexpr BlockState(unsigned short ID) :
+		ID(ID)
+	{
+	}
+
+	Block::Type Type() const;
+	bool operator == (Block::Type Type) const;
```
bearbin, line 13: This operator should not exist at all, it's confusing.

src/Protocol/Palettes/Palette_1_14.h
bearbin: Until the issue you opened on this topic is resolved, this should be made relative again.

bearbin (commented):
These changes look good in principle :)

bearbin:
+1 to peterbell10's point about making the auto-generatedness of the files explicit, and providing instructions and code to do the auto-generation; but I agree with this approach of checking in the generated code and not adding an additional code-generation step - we don't expect to be changing these files once they're generated very often.

src/BlockState.h
tigerw: for

if (Chunk.GetBlock(Pos) == Block::Type::Stone)

tigerw:
the files aren't entirely autogenerated, they're being made up manually from multiple autogeneration results. The Lua script also isn't ready yet.

src/BlockState.h
tigerw: Just saw your forum post. I mean the right side is fully typed, why bother with an extra member call? operators are overloaded for convenience after all.

bearbin: It's confusing becuse two blocks may be not equal when compared directly, but equal when compared to a third variable. Equality, even across types, only makes sense when for any set of equal elements, all comparisons within that set return a result of equality!

bearbin: Adding the extra member call makes it not a "magic equality that isn't actually doing an equality comparison", and instead a normal member function call that imposes absolutely no cognitive overhead.

peterbell10:
>they're being made up manually from multiple autogeneration results.

That is horrific. How can anyone but you reproduce these results?

You should split it up into purely auto-generated files and manually edited file. That could be a template file that gets read by the code generator, or it could be a file that `#include`s generated code. Either way, there should be a clear distinction between what is generated and what is created manually.

>I agree with this approach of checking in the generated code and not adding an additional code-generation step - we don't expect to be changing these files once they're generated very often.

My main issue with this is when someone changes the generation script, they also need to remember to manually update the checked in files. The reviewer also has to simply trust that the generated code hasn't been tampered with. Whereas if code-generation is part of the build step it is always kept in sync.

tigerw:
Wait a moment. While I would love to perfect the process and make it less "horrific" there are tens of other things to do with some semblance of 1.16 compatibility sitting at the top, and I'm the only one working on it. So let's get to a place where our users can actually connect with the latest client first.

peterbell10:
I'm concerned is that this is write-only code. Only you know what is even going on here. I have no idea how to even begin reading a 49 thousand line count diff. I presume this is mostly mechanical, so if I could at least look at the scripts that generated it then I could at least see that it's reasonable.

tigerw:
The very ad-hoc scripts are here https://github.com/cuberite/cuberite/tree/1.14/Tools/BlockTypePaletteGenerator in Generator.lua and ItemGenerator.lua. Generator.lua is for blocks.json, ItemGenerator for the various registries in registries.json, with the script patched directly as necessary. Unifying them is ideal and on the roadmap.

PR #4858, opened 2020-09-11
Reverse order of ChunkSender priorities

tigerw:
* Reduces confusion when using overloaded operator< and priority_queue


src/ChunkSender.cpp
```diff
@@ -89,7 +89,7 @@ void cChunkSender::Stop(void)
 
 
 
-void cChunkSender::QueueSendChunkTo(int a_ChunkX, int a_ChunkZ, eChunkPriority a_Priority, cClientHandle * a_Client)
+void cChunkSender::QueueSendChunkTo(int a_ChunkX, int a_ChunkZ, Priority a_Priority, cClientHandle * a_Client)
```
bearbin, line 92: This function looks like it should behave wrongly with the new priority system. As is, it looks like queueing a chunk with lower priority than it already has will be taken forwards. At the very least, this needs a comment to explain what it's doing.

src/ChunkSender.cpp
```diff
@@ -121,7 +121,7 @@ void cChunkSender::QueueSendChunkTo(int a_ChunkX, int a_ChunkZ, eChunkPriority a
 
 
 
-void cChunkSender::QueueSendChunkTo(int a_ChunkX, int a_ChunkZ, eChunkPriority a_Priority, cChunkClientHandles a_Clients)
+void cChunkSender::QueueSendChunkTo(int a_ChunkX, int a_ChunkZ, Priority a_Priority, cChunkClientHandles a_Clients)
```
bearbin, line 124: Same here.

bearbin (changes requested):

tigerw:
Thanks!

bearbin (approved):
I'm not really sure that this PR reflects the createst effort-benefit ratio of all the things that could have be done, but it looks fine to me.

tigerw:
well I'm the one doing it xD

peterbell10 (approved):
Sure, why not.

bearbin:
Well, you go on about wanting to increase the development velocity, I just wanted to point out that this _probably_ isn't a feature the users have all been clamouring over  :slightly_smiling_face:

tigerw:
true, true, but developers are the users of the code, and every dev likes deleting code :D

PR #4862, opened 2020-09-12
BlockHandler: no dynamic allocation

tigerw:


src/BlockInfo.h
```diff
@@ -47,13 +47,14 @@ class cBlockInfo
 
 	// tolua_end
 
-	inline static cBlockHandler * GetHandler      (BLOCKTYPE a_Type) { return Get(a_Type).m_Handler.get();       }
-
 	/** Creates a default BlockInfo structure, initializes all values to their defaults */
 	cBlockInfo();
 
+	/** Gets the blockhandler for the given block type. */
+	static cBlockHandler * GetHandler(BLOCKTYPE a_Type);
```
peterbell10, line 54: Why move this out of line and out of the blockinfo array?

tigerw: The array doesn't store pointers to handlers anymore, and the implementation that calls cBlockHandler::GetBlockHandler is in the cpp where it knows about BlockHandler.h.

peterbell10: >The array doesn't store pointers to handlers anymore

Yes, I can see that. I'm asking why.

tigerw: The handlers are static and cBlockHandler knows where they are and pointers can be retrieved by GetBlockHandler, why have an extra step?

src/Blocks/BlockHandler.cpp
```diff
@@ -1300,7 +1622,7 @@ void Temporary::RegisterAllBlockHandlers(BlockTypeRegistry & aRegistry)
 
 	for (const auto & r: regs)
```
bearbin, line 1623: No commenting out! If this isn't needed, just delete the whole thing (or at least comment out the for loop as well)!

bearbin (changes requested):

src/BlockInfo.h
peterbell10: A switch statement is just another form of lookup table. There is no extra step.

src/Blocks/BlockHandler.cpp
tigerw: Ah, requesting review for #4855

src/BlockInfo.h
tigerw: Sure but you have to generate the table at runtime which then takes up some memory. BlockInfo is going to be made entirely stateless so we won't have a cBlockInfoArray anymore. The current fixed std:;array<256> isn't going to play nicely with 1.13+ or warn us if we forget to handle a block type in a switch statement.

tigerw:
merge?

tigerw:
last call for comments xD

bearbin (approved):
I never liked the big table of blocktypes, this seems like an improvement, aside from the addition of an extra name for each block, which adds potential for typos and copy-paste errors. But I can't see a way around that really, so this is fine.

madmaxoft:
What exactly is wrong with (one-time) dynamic allocation?

tigerw:
One thing is the compiler complains about incomplete switch if we forget to handle a block, but it gets better than that! Now that every handler is enumerated explicitly I'm going to tag everything `constexpr`, to make all init at compile time, no startup cost at all!

Issue #4860, opened 2020-09-11
Bed is placeable on top of one block, which it shouldn't


theophriene:
Client version: 1.10.2
Server OS: linux
Cuberite Commit id: approx: c2f8ceb554982a33bcd4a1e168f6c4e26d0b85dd built at: approx: 2020-09-05 17:13:44 +0200

### Expected behavior
Bed cannot be placed on top of one block

### Actual behavior
The opposite 
![2020-09-12_03 43 11](https://user-images.githubusercontent.com/60849082/92981356-02a0a200-f489-11ea-9d67-96688e8b523f.png)

tigerw:
Minecraft Wiki says

> In Java Edition, beds do not require supporting blocks and can be placed anywhere, provided there is enough room.
> https://minecraft.gamepedia.com/Bed#Placement

Is it outdated?

tigerw:
Tried this on 1.16.3 and beds place anywhere.

tigerw:
There is multiprotocol support but game functionality follows the latest Minecraft behaviour :)

theophriene:
Oops, sorry then. I only compared the behavior against vanilla 1.10.2. It's solved then

PR #4863, opened 2020-09-12
Beds: derestrict clicking block face

tigerw:
Ref: https://github.com/cuberite/cuberite/issues/4860#issuecomment-691545904

### DIFF
--- a/src/Items/ItemBed.h
+++ b/src/Items/ItemBed.h
@@ -22,16 +22,10 @@ public:
 	}
 
 
-
-
-
 	virtual bool IsPlaceable(void) override
 	{
 		return true;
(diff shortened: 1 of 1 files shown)


src/Items/ItemBed.h
```diff
@@ -22,18 +22,12 @@ class cItemBedHandler:
 	}
 
 
-
```
bearbin, line 25: Why change the style here?

bearbin (approved):

src/Items/ItemBed.h
tigerw: It's a header file

Issue #4859, opened 2020-09-11
Some objects aren't destroyed upon destruction of the block they are attached to


ghost:
Client version: 1.10.2
Server OS: linux
Cuberite Commit id:  approx: c2f8ceb554982a33bcd4a1e168f6c4e26d0b85dd built at: approx: 2020-09-05 17:13:44 +0200

### Expected behavior
Attachable objects are dropped once the block they were attached to is destroyed
![2020-09-12_03 36 27](https://user-images.githubusercontent.com/60849082/92981149-3202df00-f488-11ea-82ed-3908913372cc.png)
![2020-09-12_03 36 44](https://user-images.githubusercontent.com/60849082/92981164-3c24dd80-f488-11ea-92fc-222379ed9bad.png)

### Actual behavior
Certain objects still stay after the block underneath (or behind) they were attached to is destroyed
![2020-09-12_03 38 11](https://user-images.githubusercontent.com/60849082/92981183-4c3cbd00-f488-11ea-80da-8604bbd7fe3b.png)
![2020-09-12_03 38 22](https://user-images.githubusercontent.com/60849082/92981169-421abe80-f488-11ea-93a3-4369b09a4679.png)

tigerw:
Thanks!

At least the blocks look like they're just missing some checks.
Painting/frame entities probably need more code in their Tick functions to ensure their block still exists.

tigerw:
Flowerpots can be placed anywhere in newer versions

tigerw:
Paintings/item frames should break after 1 second. (awfully specific of Mojang, why not immediately?)

tigerw:
> why not immediately?
Probably for performance, I guess they didn't want GetBlock calls happening every tick?

Merged into #5299.

PR #4864, opened 2020-09-12
Rail: avoid invalid game states

tigerw:
+ Check CanBeAt
Ref: #4859


bearbin (changes requested):
This PR should be +2 lines changed, not +10-30. Otherwise good.

bearbin (approved):
Well, this time is better...

PR #4867, opened 2020-09-15
[WIP] Implemented immediate Quasiconnectivity in pistons

theophriene:
Wait to see if CI will be successful


tigerw:
ah, perfect timing, I was working on this too.

theophriene:
Just asking, should this consider the case, when the piston, powered from 3 blocks away has an adjacent piston powered directly? On one hand, it is BUD circumvention, as long as the adjacent piston is moved simultaneously, and on the other it's like 8 more blocks to check for adjacency and block type, and orientation, and here goes the performance...

PR #4871, opened 2020-09-15
[WIP] Immediate quasi-connectivity, slightly redone

theophriene:
This is what I ended up with.
It still lacks a couple corner-cases, that would work (even though they're technically BUD) but they require even more recursive back-propagation, which would make piston get unpowered very slow, (slower than adjacent pistons). 
Hope, someone will give a direction where to improve this, if any.

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/PistonHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/PistonHandler.h
@@ -1,8 +1,15 @@
 
 #pragma once
 
+#include "../../Blocks/BlockButton.h"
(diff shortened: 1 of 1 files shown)


tigerw:
I think we need to fix redstone updates during block placement/destruction being too generous.

At the moment every adjacent + linked power position is updated as a catch-all.

If we added a new function to every redstone handler that unconditionally updated all the positions that component needs to, and called this function on placement and destruction (the latter with the old block), it should bring update behaviour in line with Vanilla generally.

tigerw:
?

PR #4873, opened 2020-09-16
[WIP] Fix terracota sticking to slimeblock, add IsBlockMaterialTerracotta(), and code style unification throughout BlockInfo files

theophriene:
* Made terracotta and leaves not pullable with slimeblock
* Added IsBlockMaterialTerracotta function
* Slightly changed code style throughout BlockInfo.h and BlockInfo.h

Awaiting feedback.


src/BlockInfo.h
```diff
@@ -27,23 +27,23 @@ class cBlockInfo
...
-	inline static bool FullyOccupiesVoxel         (BLOCKTYPE a_Type) { return Get(a_Type).m_FullyOccupiesVoxel;  }
-	inline static bool CanBeTerraformed           (BLOCKTYPE a_Type) { return Get(a_Type).m_CanBeTerraformed;    }
-	inline static float GetBlockHeight            (BLOCKTYPE a_Type) { return Get(a_Type).m_BlockHeight;         }
-	inline static float GetHardness               (BLOCKTYPE a_Type) { return Get(a_Type).m_Hardness;            }
+	inline static bool IsSolid                    (const BLOCKTYPE a_Type) { return Get(a_Type).m_IsSolid;             }
+	inline static bool IsUseableBySpectator       (const BLOCKTYPE a_Type) { return Get(a_Type).m_UseableBySpectator;  }
+	inline static bool FullyOccupiesVoxel         (const BLOCKTYPE a_Type) { return Get(a_Type).m_FullyOccupiesVoxel;  }
+	inline static bool CanBeTerraformed           (const BLOCKTYPE a_Type) { return Get(a_Type).m_CanBeTerraformed;    }
+	inline static float GetBlockHeight            (const BLOCKTYPE a_Type) { return Get(a_Type).m_BlockHeight;         }
+	inline static float GetHardness               (const BLOCKTYPE a_Type) { return Get(a_Type).m_Hardness;            }
```
tigerw, line 46: No need to tag const in the header, I'm fairly sure this has no effect anyway

src/Blocks/BlockPiston.cpp
```diff
@@ -176,6 +176,13 @@ bool cBlockPistonHandler::CanPushBlock(
 				// When it's not possible for a direction, then fail
 				return false;
 			}
+
+			// Terracotta and leaves can be pushed, but not pulled - so can't stick
+			const auto StickingBlockType = a_World.GetBlock(a_BlockPos + testDir);
+			if (IsBlockMaterialTerracotta(StickingBlockType) || IsBlockMaterialLeaves(StickingBlockType))
+			{
+				return false;
```
tigerw, line 184: Doesn't this mean slime blocks can no longer push terracotta? So piston -> slime -> terracotta doesn't extend?

theophriene: It is pushed ok. It's not pulled back though. Working on it

theophriene: Done.
Do I revert the const qualifier on all of the functions or just cBlockInfo members?

theophriene: Done.

src/BlockInfo.h
theophriene: Reverted.

theophriene:
Please, forgive me my stupidity

theophriene:
Bugs found

Pokechu22:
`IsBlockMaterialTerracotta` probably isn't the best approach (IMO it'd be better to have something like `DoesBlockStickToSlime`).  Also, note that only glazed terracotta doesn't stick; regular terracotta does stick.  (Though, regular terracotta was previously called hardened clay, and presumably is still called that in cuberite).

theophriene:
Regarding DoesBlockStickToSlime:
I used these two for reference
As infer from here (https://minecraft.gamepedia.com/Slime_Block)
> The types of blocks that can be moved are the same as those that can be pulled by a sticky piston.
And (https://minecraft.gamepedia.com/Piston)
shows that glazed terracotta is the only non-breakable non-falling block, push-able by piston but not pull-able.

And, seems I forgot that leaves are breakable. 
Gonna exclude them from CanBePushed check and inserted into BlockInfo pistonBreakable definition.

Thanks, for finding this.

theophriene:
Should suffice for now. Will open a separate one for adding m_PistonBreakable to blocks missing it

tigerw:
what happened, why close?

Titaniumtown:
> what happened, why close?

Good question.

PR #4874, opened 2020-09-16
Add some const qualifiers to functions

12xx12:
So I'd like to add some const qualifiers to two functions.

I need those two functions in const to make the loot tables to work.

I tried ``´GetEntityEffect()``´ it works,
I used grep for ``´GetEntityEffects()``´... it is not used

The ```GetRandomEnchantmentFromVector()``´ is a dublicate. the non const version is still ther

### DIFF
--- a/src/Enchantments.h
+++ b/src/Enchantments.h
@@ -142,6 +142,9 @@ public:
 	/** Gets random enchantment from Vector and returns it */
 	static cEnchantments GetRandomEnchantmentFromVector(cWeightedEnchantments & a_Enchantments);
 
+	/** Gets random enchantment from Vector and returns it */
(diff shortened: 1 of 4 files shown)


src/Enchantments.h
```diff
@@ -142,6 +142,9 @@ class cEnchantments
 	/** Gets random enchantment from Vector and returns it */
 	static cEnchantments GetRandomEnchantmentFromVector(cWeightedEnchantments & a_Enchantments);
```
tigerw, line 143: Couldn't you tag a_Enchantments as const and change the implementation to use a ranged for?

tigerw: `for (const auto & Enchantment : a_Enchantments)` and suchlike

tigerw (approved):

bearbin (approved):

PR #4875, opened 2020-09-16
Added armor and shulker box cleaning

### LINKED ISSUE #4840: [Bug] Armor cannot be cleaned
Cuberite Commit id: d2de029
I checked the codebase

### Expected behavior
You can clean leather armor to remove any color

### Actual behavior
nothing happens

try to clean it in cauldron

12xx12:
Fixes #4840 

Armor and Shulker boxes can now be cleaned.

Armor:
Resets color

Shulker Box:
Changes the item type to purple shulker box (default color).

This introduces a duplication bug because the block placement action is not canceld when using with a cauldron:
The shulker box in your hand is placed and you recieve one neutral colored box back,
(I have no clue how to fix this)


bearbin:
The so-called duplication bug is not actually one, it is simply a client desync bug. The client expects that the shulker box is placed and renders it as such without waiting for the server to confirm. Relogging will make it go away again.

I'm not sure what actually needs to be done here, whether there is a specific correct packet that should be sent, or if we should just send a block change for the placement block to the client.

bearbin:
Otherwise, with the client desync bug fixed, this code looks good.

bearbin:
This would be a good place to use ProtoProxy, if it is currently working; in order to see what the vanilla server does to prevent the desync.

12xx12:
https://github.com/darcros/node-minecraft-proxy

you meen this?

tigerw:
Let me run vanilla with network logging and see.

My hypothesis for how Vanilla achieves it is that it can't be some packet that's sent when you use the shulker box, since lag may mean this packet takes a while to arrive; so it's either that we're sending the wrong thing, or we're not sending some special property for either the item or the cauldron...

tigerw:
uh, scratch that. Cleaning shulker boxes is a 1.13 thing. I guess you could ignore 1.12 not knowing about it. Or resend the block, but that penalises every future version.

src/Blocks/BlockCauldron.h
```diff
@@ -98,6 +98,49 @@ class cBlockCauldronHandler :
...
+			// Resets any color to default
+			case E_ITEM_LEATHER_BOOTS:
+			case E_ITEM_LEATHER_CAP:
+			case E_ITEM_LEATHER_PANTS:
+			case E_ITEM_LEATHER_TUNIC:
+			{
+				if ((Meta > 0) && ((EquippedItem.m_ItemColor.GetRed() != 255) || (EquippedItem.m_ItemColor.GetBlue() != 255) || (EquippedItem.m_ItemColor.GetGreen() != 255)))
+				{
+					a_ChunkInterface.SetBlockMeta(a_BlockPos, --Meta);
+					LOG("%d", Meta);
```
tigerw, line 112: extra debug log

src/Blocks/BlockCauldron.h
```diff
@@ -98,6 +98,49 @@ class cBlockCauldronHandler :
 						a_Player.ReplaceOneEquippedItemTossRest(cItem(E_ITEM_GLASS_BOTTLE));
 					}
 				}
+				break;
+			}
+			// Resets any color to default
```
tigerw, line 103: By the way for these comments, could you add a colon at the end:

src/Blocks/BlockCauldron.h
```diff
@@ -98,6 +98,49 @@ class cBlockCauldronHandler :
...
+			case E_BLOCK_MAGENTA_SHULKER_BOX:
+			case E_BLOCK_ORANGE_SHULKER_BOX:
+			case E_BLOCK_PINK_SHULKER_BOX:
+			case E_BLOCK_RED_SHULKER_BOX:
+			case E_BLOCK_YELLOW_SHULKER_BOX:
+			{
+				if (Meta > 0)
+				{
+					a_ChunkInterface.SetBlockMeta(a_BlockPos, --Meta);
+					// Give new potion when the gamemode is not creative:
```
tigerw, line 138: Potion?

12xx12:
So there is actually not duplication. the client is only placing the plock on it's own.. So the client only thisplays this..,

Should we send that there was no block placed ? in 1.13 the block doesn't do any funky things

bearbin:
> https://github.com/darcros/node-minecraft-proxy
> 
> you meen this?

No, this: https://github.com/cuberite/cuberite/tree/master/Tools/ProtoProxy

But if vanilla has protocol logging, then that is more than sufficient.

To respond to Tiger, just resend the block, the penalty of _one_ block update really isn't that bad.

12xx12:
Ok, i resend the block now. the client still displays the block and then after a short time the block disappears

bearbin (approved):
Good enough, this looks ready to merge now.

12xx12:
Wait a secodn before merging!

I'm gonna try this on 1.13 just for safety

bearbin:
I wasn't going to merge immediately, I wanted @tigerw to update his thoughts as well since he made a comment before.

12xx12:
Just wanted to mention this for safety. Had a pr merged while doing a small last fix.

This results in no unintended behaviour in 1.13 👍

tigerw:
imma push some teeny style changes, hold on

src/Blocks/BlockCauldron.h
```diff
@@ -98,6 +99,65 @@ class cBlockCauldronHandler :
...
+			case E_BLOCK_PINK_SHULKER_BOX:
+			case E_BLOCK_RED_SHULKER_BOX:
+			case E_BLOCK_YELLOW_SHULKER_BOX:
+			{
+				// TODO: When there is an actual default shulker box add the appropriate changes here! - 19.09.2020 - 12xx12
+				if (Meta > 0)
+				{
+					// This is a workaround for version < 1.13. They client thinks a player placed a shulker and display that to the player
+					// The shulker cleaning was added in 1.13.
+					if (a_Player.GetPosition().y != a_BlockPos.addedY(1).y)  // Player is not standing on top
```
tigerw, line 140: I'm not sure I understand, why does the player's current world position matter?

src/Blocks/BlockCauldron.h
```diff
@@ -98,6 +99,65 @@ class cBlockCauldronHandler :
...
+					{
+						auto NewPos = a_BlockPos;
+						switch (a_BlockFace)
```
tigerw, line 143: Could AddFaceDirection work here?

src/Blocks/BlockCauldron.h
```diff
@@ -98,6 +99,65 @@ class cBlockCauldronHandler :
...
+			case E_BLOCK_LIGHT_GRAY_SHULKER_BOX:
+			case E_BLOCK_LIME_SHULKER_BOX:
+			case E_BLOCK_MAGENTA_SHULKER_BOX:
+			case E_BLOCK_ORANGE_SHULKER_BOX:
+			case E_BLOCK_PINK_SHULKER_BOX:
+			case E_BLOCK_RED_SHULKER_BOX:
+			case E_BLOCK_YELLOW_SHULKER_BOX:
+			{
+				// TODO: When there is an actual default shulker box add the appropriate changes here! - 19.09.2020 - 12xx12
+				if (Meta > 0)
```
tigerw, line 136: It may be neater to invert this, if (Meta == 0) { break; }

tigerw: less indentation and all

src/Blocks/BlockCauldron.h
12xx12: I knew there was a function like that. I tried to find this function...
Whats the class for that function

src/Blocks/BlockCauldron.h
12xx12: I added this to prevent placement if you are on top of the cauldron..

I got to check if it does actually matter...

src/Blocks/BlockCauldron.h
12xx12: Or do ```if (!(Meta == 0)```

src/Blocks/BlockCauldron.h
tigerw: It should be a free function, in Defines.h

12xx12: ah i looked for that in the vector3i class

PR #4877, opened 2020-09-16
Fixed snow golems and added documentation to damage type and added damage type for damage from environment to mobs

### LINKED ISSUE #4876: Building Snowman Crashes Server
Client version: 1.11.1
Server OS: Windows
Cuberite Commit id: a90dedffeba758fa7b7e625b43a3b400da6e5d31

### Expected behavior
Placing two snow blocks with a pumpkin on top spawns snow man

### Actual behavior
server crashes

Try to build a snowman

### Server log
```
[d52308310098c|23:06:22] Assertion failed: a_World.GetBlock(a_PumpkinPos.addedY(-1)) == E_BLOCK_SNOW, file C:\Users\Test\CLionProjects\cuberite\src\Items\ItemPumpkin.h, line 85
```

12xx12:
Fixes #4876 

Upon creating a snowman the server crashed and upon death by environmental damage it crashed again (circular relation in the target)


src/Mobs/SnowGolem.cpp
```diff
@@ -38,7 +38,7 @@ void cSnowGolem::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	}
 	if (IsBiomeNoDownfall(m_World->GetBiomeAt(POSX_TOINT, POSZ_TOINT)))
 	{
-		TakeDamage(*this);
+		TakeDamage(dtAttack, nullptr, GetRawDamageAgainst(*this), GetKnockbackAmountAgainst(*this));
```
madmaxoft, line 41: I don't think `dtAttack` is the right choice, because plugins would expect the Attacker parameter to be valid. Perhaps a new constant `dtEnvironment`?

12xx12: Oh - yeah... that might break some things

Note: when using TakeDamage(*this) the damage type defaults to dtAttack with this as cause (which is not nice.... that lead to the circular relation)

The new constant can be also used for the enderman damage from water

12xx12: Do i need to document this somewhere? the comment says there are some notes about the damage types.

bearbin: The only place you should need to document this is in the enum itself, where there is a comment for each enum value. The API documentation doesn't have an additional note for any of the other damage types, and there isn't a reason dtEnvironment would need it either.

12xx12: So the dtEnvironment isn't added to the API automaticly - do I need to do something. The damage types are exported all at once

bearbin: Yep, you're right. They're all exported at once so once you've added another enum entry, there's nothing more to be done.

bearbin: Actually, I take back that last comment. Although logically you shouldn't need to document the additional entry; because it's a new symbol the build fails; the old dt enumeration values are grandfathered in.

Since you have to document one, you may as well document them all. Add them into the APIDesc.lua file here (making sure to maintain alpha-sorting): https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/APIDesc.lua#L14207

12xx12: Like i did now?

bearbin: Yes, looks good.

bearbin (approved):

PR #4881, opened 2020-09-17
Adding end portal and enchanting table block entities

### LINKED ISSUE #4334: End portal block entity
Client version:   1.12.3
Server OS:        All
Commit id:        24a8456f79aa35cfe8c3859c880e1bffeae088c6

End portal blocks need an associated block entity, otherwise they are invisible to clients.

Related: #4126

12xx12:
This Adds the End Portal and Enchanting table block entity

Fixes #4334 , 

End portals can now be built


12xx12:
Maybe move the end portal check to the eye of ender because you can only have a portal if you place the eye in.

The portal only checks in player placed if it should be activated.

But the eye of ender decides if it's activated.

The wiki says that there are only two known seeds that generate a portal In vanilla. (The Chance is one in a trillion) So this edge case can be ignored I think.

src/Items/ItemEyeOfEnder.h
```diff
@@ -49,7 +49,7 @@ class cItemEyeOfEnderHandler:
 					}
 					// Tries to spawn portal
 					auto ChunkInterface = cChunkInterface(a_World->GetChunkMap());
-					cBlockEndPortalFrameHandler::FindAndSetPortal(a_ClickedBlockPos, FacingMeta & 3, ChunkInterface, *a_World);
```
tigerw: Can you delete the #include "EndPortalFrame.h"?

12xx12: yes

src/BlockEntities/EnchantingTableEntity.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
+
+#include "EnchantingTableEntity.h"
+
+#include "../ClientHandle.h"
+
+
+cEnchantingTableEntity::cEnchantingTableEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World):
+	Super(a_BlockType, a_BlockMeta, a_Pos, a_World),
+	m_CustomName()
```
tigerw, line 13: Could you put this member in the constructor too?

tigerw:
There are a couple debug LOGs in there, I assume you know about them

src/BlockEntities/EnchantingTableEntity.cpp
12xx12: you meen the custom name? 
Yes i could do that

12xx12:
> 
> 
> There are a couple debug LOGs in there, I assume you know about them

Thank you for reminding me

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1532,12 +1533,14 @@ void cProtocol_1_8_0::SendUpdateBlockEntity(cBlockEntity & a_BlockEntity)
...
-		case E_BLOCK_HEAD:          Action = 4; break;  // Update Mobhead entity
-		case E_BLOCK_FLOWER_POT:    Action = 5; break;  // Update flower pot
-		case E_BLOCK_BED:           Action = 11; break;  // Update bed color
+		case E_BLOCK_MOB_SPAWNER:       Action = 1; break;  // Update mob spawner spinny mob thing
+		case E_BLOCK_COMMAND_BLOCK:     Action = 2; break;  // Update command block text
+		case E_BLOCK_BEACON:            Action = 3; break;  // Update beacon entity
+		case E_BLOCK_HEAD:              Action = 4; break;  // Update Mobhead entity
+		case E_BLOCK_FLOWER_POT:        Action = 5; break;  // Update flower pot
+		case E_BLOCK_BED:               Action = 11; break;  // Update bed color
+		case E_BLOCK_ENCHANTMENT_TABLE: Action = 0; break;
```
bearbin, line 1542: What meaning does a zero action have here? Wiki.vg is not very illuminating.

12xx12: it's a bit of abuse or ugly workaround

I needed a way to send the block entities to a player

src/Blocks/BlockEnchantingTable.h
```diff
@@ -0,0 +1,95 @@
...
+		{
+			if (a_Entity.GetBlockType() != E_BLOCK_ENCHANTMENT_TABLE)
+			{
+				return false;
+			}
+			auto EnchantingTable = static_cast<cEnchantingTableEntity &>(a_Entity);
+			if (!EnchantingTable.GetCustomName().empty())
+			{
+				WindowName = EnchantingTable.GetCustomName();
+			}
```
bearbin, line 49: You should use a temporary variable for the custom name here, to avoid violating DRY and prevent any issues with mistyping and later maintenance (although this looks correct).

bearbin:
I would rather keep the existing behaviour with the end portal frame also lighting the portal if the eyes are present; that change strikes me as breaking from vanilla for a completely trivial performance gain.

Otherwise, this looks like a good PR.

src/Protocol/Protocol_1_8.cpp
bearbin: Hm, I'm not a protocol expert, so maybe anyone else can comment here; but if it works then it's fine :+1:

src/Blocks/BlockEnchantingTable.h
12xx12: You mean not defaulting to Enchant but asking fot the name and checking if there is a custom name?

12xx12:
> I would rather keep the existing behaviour with the end portal frame also lighting the portal if the eyes are present; that change strikes me as breaking from vanilla for a completely trivial performance gain.

I did not intent changing anything lighting related. I moved the portal generation the eye of ender because the last eye of ender triggers the generation (see comment above)

I didn't manage to trigger the portal when building it

> Otherwise, this looks like a good PR.

Thank you

src/Blocks/BlockEnchantingTable.h
bearbin: I was just referring to a very minor point, I intended that the above code would become:

```cpp
auto CustomName = EnchantingName.GetCustomName();
if (!CustomName.empty())
{
    WindowName = CustomName;
}
```

That way you avoid calling GetCustomName twice, and potentially causing issues later when one is changed but not the other; or if one is mistyped etc.

bearbin:
> The wiki says that there are only two known seeds that generate a portal In vanilla. (The Chance is one in a trillion) So this edge case can be ignored I think.

So is the behaviour of this code the same as previously, in the case of that one in a trillion edge case?

src/Protocol/Protocol_1_8.cpp
12xx12: Or is there another way?

You may send block entity during the chunk sending BUT this isn't implemented in cuberite.

bearbin: Ah, that makes sense more completely now. You should add a comment here explaining that this is a workaround, with TODO to fix it after adding block entities sending during chunk sending.

12xx12:
> > The wiki says that there are only two known seeds that generate a portal In vanilla. (The Chance is one in a trillion) So this edge case can be ignored I think.
> 
> So is the behaviour of this code the same as previously, in the case of that one in a trillion edge case?

The behaviour is the same EXCEPT in that edge case. The old code checked on frame placement if a portal should spawn. (Which makes no sense at all because the last eye of ender triggers the spawning)

BUT if the end portal is fully generated (that one in a trillion chance) the portal wont't trigger instantly. But I don't know if this is even possible in cuberite

bearbin:
Yes, you are right that the ender eyes being responsible for lighting the portal is much more sensible; if we do want to keep that behaviour, it would make sense to have that as a special case in the generation of the portal frames anyway. So I see that aspect of the PR is also good now.

src/Protocol/Protocol_1_8.cpp
tigerw: Only 1.9.4+'s Chunk Data has any provision to send tile entities. I just tried with 1.8.9-1.9.4 and these versions don't need you to actually send this packet for the enchantment table floating book to appear (haven't tested end portals).

12xx12: I developed with 1.11. and this version required me sending those.

src/Blocks/BlockEnchantingTable.h
12xx12: Oh - now I got it  - added this

src/Protocol/Protocol_1_8.cpp
12xx12: The other block entitires which are not listed there do have a block action making them visible to the client e.g. Ender Chest. BUT the enchanting table and the ender portal frame don't have that.

tigerw: Anyway add the TODO, we can remove the workaround when sending block entities in chunk data is implemented.

src/Blocks/BlockEnchantingTable.h
```diff
@@ -43,7 +43,8 @@ class cBlockEnchantingTableHandler :
 				return false;
 			}
 			auto EnchantingTable = static_cast<cEnchantingTableEntity &>(a_Entity);
-			if (!EnchantingTable.GetCustomName().empty())
+		  	const auto & CustomName = EnchantingTable.GetCustomName();
+			if (!CustomName.empty())
 			{
 				WindowName = EnchantingTable.GetCustomName();
```
bearbin, line 49: Not quite, you are not using the CustomName temporary variable here.

tigerw: you could even auto CustomName = GetCustomName and then std::move(CustomName) xD

src/Blocks/BlockEnchantingTable.h
```diff
@@ -46,11 +46,11 @@ class cBlockEnchantingTableHandler :
 			const auto & CustomName = EnchantingTable.GetCustomName();
 			if (!CustomName.empty())
 			{
-				WindowName = EnchantingTable.GetCustomName();
+				WindowName = CustomName;
 			}
 			return true;
 		});
-		cWindow * Window = new cEnchantingWindow(a_BlockPos, WindowName);
+		cWindow * Window = new cEnchantingWindow(a_BlockPos, std::move(WindowName));
```
bearbin, line 53: @tigerw - is this what you meant? If so please approve :)

bearbin (approved):

src/Protocol/Protocol_1_8.cpp
peterbell10: Weren't these normal blocks that later became block entities? If that's the case you shouldn't send any packet in the versions where they aren't entities. I can't find a reference for when that was though.

tigerw: We can handle that in chunkdataserializer, I don't think Action = 0 has any ill effects as a workaround

peterbell10: Chunk serialisation is irrelevant. You aren't going to resend the entire chunk every time a block entity gets placed.

tigerw:
imma push some teeny style changes, hold on

src/Protocol/Protocol_1_8.cpp
tigerw: We can drop all changes to SendUpdateBlockEntity in all versions when serialisation is fixed. These workarounds are just to declare the entity during initial chunk send so blocks aren't invisible; the client creates tile entities itself during normal block placement.

12xx12: exept for the end portal... If you place a enchanting table you see the book till logout. But the end portal blocks are invisible if they spawn

tigerw: On 1.12 SetBlock an end portal during gameplay shows up fine, no invisibility

12xx12: On relog it doesn't

tigerw: Sure, that's what we're talking about fixing with chunk sending in the future :)

12xx12: > On 1.12 SetBlock an end portal during gameplay shows up fine, no invisibility

This got me a bit confused. I thought you talked about the release implementation

peterbell10: >We can drop all changes to SendUpdateBlockEntity in all versions when serialisation is fixed.

Plugins can also change block entity properties and there's no way the client could possibly know that.

tigerw: Indeed but from what I can see this packet doesn't have any Action corresponding to updating enchanting tables or end portals (1.16.3). By drop changes I meant the special Action = 0 added in this PR needn't stay forever.

src/Blocks/BlockEnchantingTable.h
```diff
@@ -36,61 +32,53 @@ class cBlockEnchantingTableHandler :
 	) override
 	{
 		AString WindowName = "Enchant";
-		a_WorldInterface.DoWithBlockEntityAt(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, [&] (cBlockEntity & a_Entity)
+		a_WorldInterface.DoWithBlockEntityAt(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, [&WindowName](cBlockEntity & a_Entity)
```
tigerw, line 35: Only capture needed

src/Blocks/BlockEnchantingTable.h
```diff
@@ -36,61 +32,53 @@ class cBlockEnchantingTableHandler :
...
 		{
 			if (a_Entity.GetBlockType() != E_BLOCK_ENCHANTMENT_TABLE)
 			{
 				return false;
 			}
-			auto EnchantingTable = static_cast<cEnchantingTableEntity &>(a_Entity);
+
+			const auto & EnchantingTable = static_cast<cEnchantingTableEntity &>(a_Entity);
```
tigerw: Remove accidental copy

src/Items/ItemEnchantingTable.h
```diff
@@ -44,32 +35,29 @@ class cItemEnchantingTableHandler:
 		const Vector3i a_CursorPos
 	) override
 	{
-		// Cannot place a enchanting table at "no face":
-		if (a_ClickedBlockFace == BLOCK_FACE_NONE)
```
tigerw: Super::OnPlayerPlace already handles this and returns false

src/Items/ItemEnchantingTable.h
```diff
@@ -44,32 +35,29 @@ class cItemEnchantingTableHandler:
...
+		if (!Super::OnPlayerPlace(a_World, a_Player, a_EquippedItem, a_ClickedBlockPos, a_ClickedBlockFace, a_CursorPos))
 		{
-			return true;
+			return false;
 		}
-		const auto PlacePos = AddFaceDirection(a_ClickedBlockPos, a_ClickedBlockFace);
 
-		cItem ItemCopy(a_EquippedItem);  // Make a copy in case this is the player's last item and OnPlayerPlace removes it
```
tigerw: This wasn't used

src/BlockEntities/EnchantingTableEntity.cpp
```diff
@@ -1,14 +1,14 @@
 
-// BlockEntityEnchantingTable.cpp
-
 #include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
 
 #include "EnchantingTableEntity.h"
+#include "ClientHandle.h"
```
bearbin, line 5: Uuh Tiger, didn't you get told off for doing this in another PR?

tigerw:
@12xx12 I added the "small fixes" commit, your comments are appreciated

tigerw (approved):

src/BlockEntities/EnchantingTableEntity.cpp
tigerw: let's change it, the ../../ are an exercise in masochism

src/Protocol/Protocol_1_8.cpp
12xx12: resend whole chunk then? :/

src/BlockEntities/EnchantingTableEntity.cpp
bearbin: Open an issue or make a forum thread and get it agreed, don't just sneak it into _somebody else's_ PR!

12xx12:
So you suggested changes are already done?

src/Protocol/Protocol_1_8.cpp
tigerw: I dunno, we can figure that out later.

12xx12: so we agree this works for now and the one rewriting the SehndChunk code adds the block entites?

tigerw:
Yup, just pointing out some relevant changes

src/BlockEntities/EndPortalEntity.cpp
```diff
@@ -1,13 +1,14 @@
...
 #include "EndPortalEntity.h"
-#include "../ClientHandle.h"
+#include "ClientHandle.h"
+
+
+
 
 
-cEndPortalEntity::cEndPortalEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World):
+cEndPortalEntity::cEndPortalEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World) :
```
12xx12, line 11: Is there a reason for this whitespace?

bearbin:
Alright, back to how it was before, @tigerw you're on to approve and merge.

src/BlockEntities/EndPortalEntity.cpp
tigerw: style mainly xD

12xx12:
Go for it

PR #4882, opened 2020-09-17
Fixed leaves and wheat not breakable with piston

theophriene:
Also, can E_ITEM_* be used in Info[*].m_PistonBreakable ?
If it can painting and item frame are to be added too.


tigerw:
Pistons breaking off item frame/painting entities will need special checks, putting E_ITEMs in there won't work by itself.

tigerw:
Looks good

Pokechu22:
`E_BLOCK_BEETROOTS` should probably also be breakable.

bearbin (approved):

PR #4885, opened 2020-09-17
Deleted BiomeDef.h and ChunkDef.h from Globals.h

KingCol13:
As requested by #4667 and attempted in #4785.

Inserted headers where required and compiles.


theophriene:
I found a side problem - GCC is auto-including missing headers, causing no compilation errors when trying to compile  369460011a0a51ffc60c405aafa99f93c33a1e10,

In 'src/StringUtils.h'
```cpp
#include "Globals.h" // For typedefs
```

In "src/BiomeDef.h"
```cpp
// #include <stddef.h>      // Already included in Globals
// #include "StringUtils.h" // Already included in Globals
#include "Globals.h"        // Both of the above + int typedefs
```

In "src/Generating/Trees.h"
```cpp
#include "../BiomeDef.h" // For NIBBLETYPE
#include "../ChunkDef.h" // For sSetBlockVector
```

And it compiles again,
tried with:
``` 
gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
```

KingCol13:
Thanks for the help but this is branch is a mess. I think I'll start again, applying your suggestions and thinking more carefully about what should be included where. I also need to check the build tools compile.

bearbin:
This looks good, is the draft status still necessary or is this ready for review @KingCol13?

src/BlockInfo.h
```diff
@@ -1,6 +1,7 @@
 
 #pragma once
 
+#include "ChunkDef.h"
```
peterbell10, line 4: Block info is included in globals, so this PR won't change anything.

KingCol13: I may not be looking properly but I couldn't see how it was included, could you explain?

peterbell10: Oh sorry, I was looking at one of the other `Globals.h` files. Never mind then.

KingCol13:
I made it ready for review and it compiles, but I'm not so confident in the quality as I mostly mashed compile until it worked and I'm unfamiliar with the codebase on the larger scale. Also I didn't notice any difference in compile time.

bearbin (approved):
Well, it compiles and seems logical. If CI doesn't use unity builds, I see no problem with this.

src/Blocks/ChunkInterface.cpp
```diff
@@ -1,13 +1,12 @@
...
 #include "Globals.h"
 
-#include "ChunkInterface.h"
 #include "BlockHandler.h"
 #include "WorldInterface.h"
 #include "../ChunkMap.h"
 #include "../World.h"
 #include "../BlockInfo.h"
-
+#include "ChunkInterface.h"
```
peterbell10, line 9: If it fails to compile with the include at the top, then there are includes missing from the "ChunkInterface.h" header

src/LineBlockTracer.cpp
```diff
@@ -4,10 +4,10 @@
 // Implements the cLineBlockTracer class representing a cBlockTracer that traces along a straight line between two points
 
 #include "Globals.h"
-#include "LineBlockTracer.h"
 #include "BlockInfo.h"
 #include "World.h"
 #include "Chunk.h"
+#include "LineBlockTracer.h"
```
peterbell10, line 10: Same here

peterbell10 (commented):
I'm a bit uncertain about this because some headers are basically included everywhere even if they're not in "Globals.h". So, it's effectively global anyway.

KingCol13:
Ran fine with:
cmake -DCMAKE_BUILD_TYPE=RELEASE -DBUILD_TOOLS=Yes -DPRECOMPILE_HEADERS=No -DSELF_TEST=ON -DUNITY_BUILDS=OFF -DWHOLE_PROGRAM_OPTIMISATION=OFF .. && cmake --build . && ctest

I agree with @peterbell10 though that things like NIBBLETYPE defined in ChunkDef.h (which then includes BiomeDef.h) seem to be needed almost everywhere.

bearbin:
Maybe this highlights the need to split these header files up a bit more, certainly NIBBLETYPE ought to be defined together with the rest of our types rather than in ChunkDef!

tigerw:
Globals.h is the precompiled header, we need to be careful not to remove commonly used stuff from it and end up slowing down compile times.

src/StringUtils.h
```diff
@@ -8,7 +8,14 @@
...
 #pragma once
 
+#include <cstddef>
 
+#include <string>
+#include <list>
+#include <vector>
+#include <map>
+
+#include "fmt.h"
```
tigerw, line 18: We'd want these headers precompiled for example, @KingCol13 do you know what needed this

peterbell10, line 18: This definitely shouldn't be necessary, all of these headers are in `Globals.h`.

peterbell10 (commented):
Unfortunately, I don't think is going to add any value without a lot more work. "World.h" (which is basically included everywhere) relies on `cChunkDef` so having it in the precompiled header makes sense.

`BiomeDef.h` would be easier to separate out, but is actually very small already; just an enum and a few function declarations. Most likely not worth the effort.

src/StringUtils.h
KingCol13: AString needed <string>, then there's AStringMap AStringList and AStringVector. StringUtils.h is included in Globals.h anyway.

KingCol13:
Take it or leave it, this doesn't really do anything and I'm completely indifferent whether it gets merged or not.

bearbin:
I'm merging this, although there probably isn't much benefit; there shouldn't be any harm either; and it fixes an issue :)

KingCol13:
> there shouldn't be any harm either

It seems it now fails to build, probably one of the recent merges used one of those 2 headers. I'll try to fix this but maybe we should bin it

PR #4889, opened 2020-09-18
Implement redstone quasi-connectivity

tigerw:
On ~~QC~~ QM Einstein wrote:
> I cannot seriously believe in it because the theory cannot be reconciled with the idea that physics should represent a reality in time and space, free from spooky action at a distance

Yeah, then you got this:

Q-what

Outstanding issue: our redstone updates are too generous. See #4871.

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/DropSpenserHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/DropSpenserHandler.h
@@ -65,6 +65,11 @@ namespace DropSpenserHandler
 		UNUSED(a_Chunk);
 		UNUSED(a_BlockType);
 		UNUSED(a_Meta);
+
+		// Consider indirect power:
+		Callback.CheckIndirectPower();
+
+		// Consider normal adjacents:
 		InvokeForAdjustedRelatives(Callback, a_Position, RelativeAdjacents);
 	}
 };
--- a/src/Simulator/IncrementalRedstoneSimulator/ForEachSourceCallback.cpp
+++ b/src/Simulator/IncrementalRedstoneSimulator/ForEachSourceCallback.cpp
@@ -23,25 +23,6 @@ ForEachSourceCallback::ForEachSourceCallback(const cChunk & Chunk, const Vector3
 
 
 
-bool ForEachSourceCallback::ShouldQueryLinkedPosition(const Vector3i Location, const BLOCKTYPE Block)
-{
-	switch (Block)
-	{
-		// Normally we don't ask solid blocks for power because they don't have any (store, dirt, etc.)
-		// However, these are mechanisms that are IsSolid, but still give power. Don't ignore them:
-		case E_BLOCK_BLOCK_OF_REDSTONE:
-		case E_BLOCK_OBSERVER:
-		case E_BLOCK_TRAPPED_CHEST: return false;
-
-		// If a mechanism asks for power from a block, redirect the query to linked positions if:
-		default: return cBlockInfo::IsSolid(Block);
-	}
-}
-
-
-
-
-
 void ForEachSourceCallback::operator()(Vector3i Location)
 {
 	if (!cChunk::IsValidHeight(Location.y))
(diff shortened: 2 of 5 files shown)


Titaniumtown:
I google translated it, and the title is german for "spooky action at a distance". lol

src/Simulator/IncrementalRedstoneSimulator/ForEachSourceCallback.cpp
```diff
@@ -49,6 +49,36 @@ void ForEachSourceCallback::operator()(Vector3i Location)
...
 
+	// Quasiconnectivity handling
+	// If the block is viable and the source offset is elite:
+	if (
+		CanBeIndirectlyPowered(m_CurrentBlock) &&
+		((Location - m_Position) == Vector3i(1, 33, 7))
+	)
+	{
+		const Vector3i OffsetYP(0, 1, 0);
+		const auto Above = m_Position + OffsetYP;
```
tigerw, line 60: .addedY(1)

src/Simulator/IncrementalRedstoneSimulator/ForEachSourceCallback.cpp
```diff
@@ -49,6 +49,36 @@ void ForEachSourceCallback::operator()(Vector3i Location)
 		return;
 	}
 
+	// Quasiconnectivity handling
+	// If the block is viable and the source offset is elite:
```
bearbin, line 53: Can you de-thesaurify this comment for me?

bearbin (commented):
This PR just reinforces that I don't understand redstone...

Unless anyone else has substantitive comments, I say merge.

tigerw:
> This PR just reinforces that I don't understand redstone...
> Unless anyone else has substantitive comments, I say merge.

D:
Do the changes not make sense, or?

bearbin:
The changes make sense, just these are not mechanics I have ever used in my gameplay :P

tigerw:
well me neither 👍

src/Simulator/IncrementalRedstoneSimulator/DropSpenserHandler.h
```diff
@@ -65,6 +65,11 @@ namespace DropSpenserHandler
 		UNUSED(a_Chunk);
 		UNUSED(a_BlockType);
 		UNUSED(a_Meta);
+
+		// Consider indirect power:
+		Callback(a_Position + Vector3i(1, 33, 7));
```
peterbell10, line 70: How about passing a boolean instead of a plate of spaghetti?

tigerw:
nobody cares for Vector3i of 1337 ._.

src/Simulator/IncrementalRedstoneSimulator/ForEachSourceCallback.cpp
tigerw: elite, like e1337 :p

but apparently elitism isn't popular here so I'll stick this in special function to handle QC

src/Simulator/IncrementalRedstoneSimulator/ForEachSourceCallback.cpp
```diff
@@ -78,6 +59,53 @@ void ForEachSourceCallback::operator()(Vector3i Location)
...
+
+
+
+bool ForEachSourceCallback::ShouldQueryLinkedPosition(const BLOCKTYPE Block)
+{
+	switch (Block)
+	{
+		// Normally we don't ask solid blocks for power because they don't have any (store, dirt, etc.)
+		// However, these are mechanisms that are IsSolid, but still give power. Don't ignore them:
+		case E_BLOCK_BLOCK_OF_REDSTONE:
```
theophriene, line 96: There is a problem with those, as the block they power via QC needs an update to actually be powered.
[As shown here](https://minecraft.gamepedia.com/Tutorials/Quasi-connectivity#Update_QC_activation)
And the same goes for attachable redstone components, unless they are attached to the block above the piston, because:
>The following redstone components can activate mechanism components one block away, and update redstone components adjacent to the block they are attached to (including above and below) as well as redstone components adjacent to themselves: 
 >  * Buttons (attaches in any direction)
 > * Detector Rail (attaches only downwards)
 >  * Lever (attaches in any direction)
 >  * Pressure Plates (attaches only downwards)
 >  * Trapped Chest (doesn't actually attach, but updates as if attached to block beneath)
 >  * Tripwire Hook (attaches only sideways)
 >  * Weighted Pressure Plates (attaches only downwards)

Which means they technically can provide power right away if there exists a directly powered redstone component that is adjacent to both the QC-powered component and the source, thus 'transferring' block update to the QC powered component, like this:
1. Without block transferring an update:



2. With an update-transferring block in between the source and the QC site:



This is just what I've run into while tinkering with pistons and redstone piston handler. Hope you'll find anything of this useful.

tigerw: Thanks, if I understand you correctly, you're saying our block updates are too generous right? Yeah that's the case, currently most block changes update 24 (I think) redstone positions around it. That would need a separate fix.

tigerw:
yup, merge?

bearbin (approved):

peterbell10 (approved):

PR #4890, opened 2020-09-18
Trapdoor crash fix

theophriene:
* Fixed trapdoor crashing server due to failed assert
* Added trapdoor rotation upon placement

I'm not familiar enough with api - most likely there's a better way to do things.
Awaiting your suggestions.

~~P.S.: Trpdoors still do move up in one tick after placement / on block update.~~ Not anymore

Also, comments in the code will be removed anyway


theophriene:
Okay, seems to be non-compliant to vanilla. Back to testing

theophriene:
Seems done.
Tested placing on horizontal surface, from the position above/ under the lapis block:

theophriene:
> Seems done.
> Tested placing on horizontal surface, from the position above/ under the lapis block: (all the doors are opened afterwards to see the alignment)
> 
> 

The only issue is that the minecraft is extremely sensitive to position, causing things like this, when player is not aligned properly

bearbin (changes requested):
I don't like calculating a 'safe block face', and just calculating block meta using that face. 

I would replace `BlockFaceToMetaData` with a new function `BlockFaceAndOffsetToMetaData`, replacing the existing BlockFaceToMetaData function.

Also, please don't use C-style casts and instead use C++-style.

Edit: in terms of behaviour, this PR looks good to me. It may not be exactly vanilla behaviour but it's sensical enough.

bearbin:
Also, please add yourself to the CONTRIBUTORS file to accept the licensing terms and be recognised as a Cuberite contributor.

theophriene:
> Also, please add yourself to the CONTRIBUTORS file to accept the licensing terms and be recognised as a Cuberite contributor.

Will do, when code is ready quality-wise

bearbin (approved):

src/Blocks/BlockTrapdoor.h
```diff
@@ -84,31 +84,72 @@ class cBlockTrapdoorHandler :
...
 		a_BlockType = m_BlockType;
-		a_BlockMeta = BlockFaceToMetaData(a_ClickedBlockFace);
+		a_BlockMeta = BlockFaceAndOffsetToMetaData(a_ClickedBlockFace, BlockOffsetRelToPlayer);
 
-		if (a_CursorPos.y > 7)
+		// Trapdoor is placed on top of a block
+		if ((a_ClickedBlockFace == BLOCK_FACE_YP))
+		{
+			// Toggle 'Move up half-block' bit off
+			a_BlockMeta &= a_BlockMeta | 0x8;
```
peterbell10, line 98: This line doesn't do anything (x & (x | y) is just x). Perhaps you meant
```cpp
a_BlockMeta &= ~0x08;
```
which unsets the 8s bit.

peterbell10 (dismissed):

src/Blocks/BlockTrapdoor.h
theophriene: fixed.

tigerw:
On YM/YP scenarios this places the trapdoor with the hinge dependant on the player's position relative to the block, instead vanilla puts the hinge dependant on the player's look vector. Let me try to apply a change

tigerw (approved):

tigerw:
@theophriene pushed some changes for your consideration

theophriene:
Oh, wow! That's what I thought should exist somewhere. I should've checked the base class from the start. Thanks.

bearbin (approved):

PR #4891, opened 2020-09-18
BlockHandler initialisation is a constant expression

tigerw:
If we can't make it all namespaces, this is the next best I guess.

+ Tag handlers constexpr, const as needed
+ Inherit constructors
* Privatise handler functions


tigerw:
```
/root/cuberite/src/Blocks/BlockFence.h:28:15: error: invalid case style for private member 'MAX_COORD' [readability-identifier-naming,-warnings-as-errors]
        const double MAX_COORD = 0.6;
                     ^~~~~~~~~
                     m_MAX_COORD
```

src/Blocks/BlockHandler.h
```diff
@@ -25,9 +25,11 @@ class BlockTypeRegistry;
 class cBlockHandler
 {
 public:
-	cBlockHandler(BLOCKTYPE a_BlockType);
 
-	virtual ~cBlockHandler() {}
```
peterbell10, line 30: This should still have a virtual destructor, even if you don't think you'll need it.

src/Blocks/BlockFence.h
```diff
@@ -17,22 +17,17 @@ class cBlockFenceHandler:
...
 
+	using Super::Super;
+
+private:
+
 	// These are the min and max coordinates (X and Z) for a straight fence.
 	// 0.4 and 0.6 are really just guesses, but they seem pretty good.
 	// (0.4 to 0.6 is a fence that's 0.2 wide, down the center of the block)
 	const double MIN_COORD = 0.4;
 	const double MAX_COORD = 0.6;
```
peterbell10, line 28: Making it `static constexpr` should silence the warnings IIRC.

peterbell10 (changes requested):

src/Blocks/BlockFence.h
tigerw: Thanks!

src/Blocks/BlockHandler.h
tigerw: aww but I looked forward to deleting this. I'm not doing any dynamic allocation, so why do I need one?

peterbell10: AFAIK there's no reason not to include it, and some potentially nasty bugs avoided by including it.

tigerw: Hm, it doesn't let me add a destructor to constexpr

peterbell10: Hrm, I guess it's not possible for `constexpr` then. Oh well.

bearbin (commented):
Soft-approve: the changes look good in principle but I haven't read through every line.

peterbell10 (approved):

PR #4892, opened 2020-09-18
IBM Z

tigerw:


Titaniumtown:
What do you mean by "IBM Z"?

tigerw:
Test Cuberite on IBM S/390 mainframes.

(though travis calls them IBM Z, i'm a little confused)

Titaniumtown:
I wonder if anyone would ever use cuberite on those. lol

bearbin:
I don't see the value in this PR at all - this will extend the build times for no good reason; and lead to enlarged backlogs.

tigerw:
Once we get sqlitecpp updated we can test a quadrumvirate of ARM, PPC, S390, AMD64, that'd be beautiful

bearbin:
In theory that would be nice to ensure we maintain portability, but on the other hand we have zero users on those platforms, and an extra 10 minutes of build for each commit is not nice.

tigerw:
ah ok, well good to know that it builds & runs well on S/390 at this point in time.

Surprisingly s390x is actually the fastest on Travis, nothing else is sub-2 minutes.

Let me unstage the S/390 addition and just leave the other changes.

.travis.yml
```diff
@@ -9,19 +9,23 @@ cache:
...
-# TODO: add IBM builds when we find CMake for them
-#       add PowerPC builds when sqlite3 crashing is fixed
+# Install newer SeeMake on Linux
+addons:
+  snaps:
+    - name: cmake
+      confinement: classic
+      channel: latest
+
+# TODO: add PowerPC builds when sqlite3 crashing is fixed
```
bearbin, line 19: This todo can be removed as well

bearbin (approved):

tigerw:
> (though travis calls them IBM Z, i'm a little confused)

apparently Z stands for zero downtime

PR #4897, opened 2020-09-19
Implement fortune for ores, glowstone and sea lanterns

KingCol13:
Implements fortune for the ores, glowstone and sea lanterns based on wiki rates:
https://minecraft.gamepedia.com/Fortune#Ore

Does not include gravel or anything "organic" (although I'm happy to do this too), as worked on in #4802.


src/Blocks/BlockHandler.cpp
```diff
@@ -656,6 +656,15 @@ bool cBlockHandler::ToolHasSilkTouch(const cItem * a_Tool)
 
 
 
+unsigned int cBlockHandler::ToolFortuneLevel(const cItem * a_Tool)
+{
+	return (a_Tool != nullptr)*a_Tool->m_Enchantments.GetLevel(cEnchantments::enchFortune);
+}
```
bearbin, line 662: Please, no! What's wrong with a normal if statement? We're using C++, not playing golf with C...

src/Blocks/BlockOre.h
```diff
@@ -3,15 +3,75 @@
...
-
 class cBlockOreHandler:
 	public cBlockHandler
 {
 	using Super = cBlockHandler;
 
+private:
+	inline unsigned int fortuneDropMult(unsigned int fortuneLevel)
+	{
+		/** Returns the correct drop multiplier for ores
```
bearbin, line 14: This comment should go before the function header.

src/Blocks/BlockOre.h
```diff
@@ -35,17 +95,19 @@ class cBlockOreHandler:
 			}
 		}
 
-		// TODO: Handle the Fortune enchantment here
+		unsigned int fortuneLevel = ToolFortuneLevel(a_Tool);
+		unsigned int dropMult = fortuneDropMult(fortuneLevel);
```
bearbin, line 99: These two should be `const`, and they can also be `auto`. It's also our style to use UpperCamelCase for local variable names (although I personally prefer camelCase).

bearbin (changes requested):

bearbin (approved):

src/Blocks/BlockOre.h
```diff
@@ -3,15 +3,76 @@
...
+				else if (RandUnif < 0.8)
+				{
+					return 3;
+				}
+				else
+				{
+					return 4;
+				}
+			}
+		}
```
peterbell10, line 68: This switch statement is equivalent to:
```cpp
std::max(1, FloorC(RandUnif * (a_FortuneLevel + 2)));
```

tigerw: the switch returned a 1 multiplier for any level greater than 3; we may still want some limit to avoid high fortune levels crashing the server with a lot of pickups. std::clamp(1, 10, ...) maybe?

PR #4900, opened 2020-09-19
Synchronously save chunks on server shutdown

### LINKED ISSUE #4884: Wolfes sometimes loose color on collar on server restart
Client version: 1.11.1
Server OS: Windows
Cuberite Commit id: 53549a1a4ce320c1e3f449165c290c74d3de54d4

### Expected behavior
Collars can be colored and should keep their color on server restart

### Actual behavior
Color loss on restart

- tame wolf
- color collar
- stop server
- restart

bearbin:
Fixes #4884

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11653,10 +11653,6 @@ a_Player:OpenWindow(Window);
 				{
 					Notes = "Saves all the chunks in all the worlds. Note that the saving is queued on each world's tick thread and this functions returns before the chunks are actually saved.",
 				},
-				SaveAllChunksNow =
-				{
-					Notes = "Saves all the chunks in all the worlds. Note that saving is performed synchronously, and this function will only return when all worlds are saved. For performance reasons, this is NOT RECOMMENDED to be used except where necessary, as this will cause lagspikes.",
-				},
 				SetSavingEnabled =
 				{
 					Params =
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -11653,6 +11653,10 @@ a_Player:OpenWindow(Window);
 				{
 					Notes = "Saves all the chunks in all the worlds. Note that the saving is queued on each world's tick thread and this functions returns before the chunks are actually saved.",
 				},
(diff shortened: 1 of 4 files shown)


peterbell10:
Shouldn't we make sure all queued tasks get executed and not just this special case?

tigerw:
Queued tasks also have a (potentially long) tick delay.

bearbin:
Potentially we could add a `RunAllQueuedTasksNowAndWaitForThemToFinish` function on cWorld, and then use that instead, but we would still need a special case for saving, because we would want the saving to run after all of the other queued tasks (wouldn't we?).

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -11653,6 +11653,10 @@ a_Player:OpenWindow(Window);
 				{
 					Notes = "Saves all the chunks in all the worlds. Note that the saving is queued on each world's tick thread and this functions returns before the chunks are actually saved.",
 				},
+				SaveAllChunksNow =
```
tigerw, line 11656: I would hold off exporting this, it could introduce deadlocks if plugins try to use it in a world thread

bearbin:
I was pushing that change just as you were writing the comment!

peterbell10:
Maybe the queue could be completed during world shutdown? We need to be sure nothing will be queued up after the queue is emptied.

tigerw:
> I was pushing that change just as you were writing the comment!

Synchronously ~~save chunks~~ post review comments on ~~server shutdown~~ commit push.

src/Server.cpp
```diff
@@ -648,7 +648,8 @@ void cServer::Shutdown(void)
 	// Notify the tick thread and wait for it to terminate:
 	m_TickThread.Stop();
 
-	cRoot::Get()->SaveAllChunks();
+	// Save all chunks in all worlds, wait for chunks to be saved:
```
tigerw, line 651: I just looked at cWorld::SaveAllChunks and apparently it only adds it to the world storage queue, it doesn't wait for it to empty. However cRoot waits for the thread to finish to it's still ok, just the comment might be wrong. Could you confirm?

bearbin: Your comment is completely correct, I'll update my comment.

tigerw (approved):

PR #4924, opened 2020-09-21
added dtEnvironment to string conversion

12xx12:
I have done bad things in the past  - longe live compiler warnings

I forgot to add dtEnvironment to the string conversion


peterbell10 (approved):

Seadragon91:
@12xx12 Sorry there is still one more place missing
```
home/lukas/cpp/cuberite/src/Entities/Player.cpp:1186:13: warning: enumeration value 'dtEnvironment' not handled in switch [-Wswitch]
                                switch (a_TDI.DamageType)
```
Calling this lua code will result in a crash: `a_Player:TakeDamage(dtEnvironment, nil, 1000, 0)`

PR #4925, opened 2020-09-22
added dtEnvitonment to player damage text

12xx12:
me bad - me sleepy

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -1203,7 +1203,7 @@ void cPlayer::KilledBy(TakeDamageInfo & a_TDI)
 					case dtAdmin:           return "was administrator'd";
 					case dtExplosion:       return "blew up";
 					case dtAttack:          return "was attacked by thin air";
-					case dtEnvironment:     return "played to much dress up";  // This is not vanilla - added a own pun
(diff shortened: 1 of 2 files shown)


tigerw:
Make sure to get sufficient sleep! :)

src/Entities/Player.cpp
```diff
@@ -1203,6 +1203,7 @@ void cPlayer::KilledBy(TakeDamageInfo & a_TDI)
 					case dtAdmin:           return "was administrator'd";
 					case dtExplosion:       return "blew up";
 					case dtAttack:          return "was attacked by thin air";
+					case dtEnvironment:     return "played to much dress up";  // This is not vanilla - added a own pun
```
tigerw, line 1206: you have to explain the pun, I don't get it...

12xx12: You only get this type of damage if you are an enderman in the rain or a snowman in the desert

12xx12:
Eat sleep code repeat

src/Entities/Player.cpp
bearbin: I prefer "succumbed to the conditions", but your choice would be fine after fixing the typo "to" -> "too".

bearbin (approved):

PR #4926, opened 2020-09-22
Limit fortune level, style fixes

tigerw:
Fixup recent merges.

### DIFF
--- a/src/Bindings/Plugin.h
+++ b/src/Bindings/Plugin.h
@@ -66,7 +66,7 @@ public:
 	virtual bool OnHandshake                (cClientHandle & a_Client, const AString & a_Username) = 0;
 	virtual bool OnHopperPullingItem        (cWorld & a_World, cHopperEntity & a_Hopper, int a_DstSlotNum, cBlockEntityWithItems & a_SrcEntity, int a_SrcSlotNum) = 0;
 	virtual bool OnHopperPushingItem        (cWorld & a_World, cHopperEntity & a_Hopper, int a_SrcSlotNum, cBlockEntityWithItems & a_DstEntity, int a_DstSlotNum) = 0;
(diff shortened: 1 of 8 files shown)


src/Bindings/Plugin.h
```diff
@@ -66,7 +66,7 @@ class cPlugin
 	virtual bool OnHandshake                (cClientHandle & a_Client, const AString & a_Username) = 0;
 	virtual bool OnHopperPullingItem        (cWorld & a_World, cHopperEntity & a_Hopper, int a_DstSlotNum, cBlockEntityWithItems & a_SrcEntity, int a_SrcSlotNum) = 0;
 	virtual bool OnHopperPushingItem        (cWorld & a_World, cHopperEntity & a_Hopper, int a_SrcSlotNum, cBlockEntityWithItems & a_DstEntity, int a_DstSlotNum) = 0;
-	virtual bool OnDropSpense        		(cWorld & a_World, cDropSpenserEntity & a_DropSpenser, int a_SlotNum) = 0;
+	virtual bool OnDropSpense        	    (cWorld & a_World, cDropSpenserEntity & a_DropSpenser, int a_SlotNum) = 0;
```
bearbin, line 69: Indentation here is still wrong.

KingCol13:
It's like getting my homework corrected haha. Would it make sense if I made cItems take unsigned char for ItemCount and all enchantments return unsigned char as well?

bearbin (changes requested):

tigerw:
> It's like getting my homework corrected haha. Would it make sense if I made cItems take unsigned char for ItemCount and all enchantments return unsigned char as well?

I'm just being too pedantic. Moved the limit to the blockhandler function so we don't forget to apply it anywhere. I think it would make sense, might be a lot of work though since cItem is used almost everywhere, but as you wish.

bearbin (approved):

KingCol13:
> too pedantic

be as pedantic as you like, I've never worked on a big project like this before so I'm happy to see how I can improve.

PR #4927, opened 2020-09-22
Fix ice behaviour in world

### LINKED ISSUE #4923: Proper packed ice behaviour

https://minecraft.gamepedia.com/Packed_Ice

### Expected behavior
When breaking packed ice no wather should spawn

### Actual behavior
it does

maybe there are some other things...

12xx12:
This fixes some bahavior around ice blocks in the world.

- They now smelt properly
- The packed ice doesn't spawn anymore water now

fixes 4923 partially atm. got to check some speed bahaviour on packed ice

### DIFF
--- a/src/Blocks/BlockIce.h
+++ b/src/Blocks/BlockIce.h
@@ -51,15 +51,12 @@ private:
 		for (const auto & Offset : Offsets)
 		{
 			auto Position = a_RelPos + Offset;
-			const auto Chunk = a_Chunk.GetRelNeighborChunk(Position.x, Position.z);
+			const auto Chunk = a_Chunk.GetRelNeighborChunkAdjustCoords(Position);
 			if (!Chunk->IsLightValid())
 			{
 				Chunk->GetWorld()->QueueLightChunk(Chunk->GetPosX(), Chunk->GetPosZ());
 				return;
 			}
-
-			Position.x = ((Position.x % cChunk::Width) + cChunk::Width) % cChunk::Width;
-			Position.z = ((Position.z % cChunk::Width) + cChunk::Width) % cChunk::Width;
 
 			if (Chunk->GetBlockLight(Position) > 11)
 			{
--- a/src/Blocks/BlockIce.h
+++ b/src/Blocks/BlockIce.h
@@ -31,6 +31,44 @@ private:
 		}
 	}
 
+	virtual void OnUpdate(
+		cChunkInterface & a_ChunkInterface,
+		cWorldInterface & a_WorldInterface,
+		cBlockPluginInterface & a_PluginInterface,
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) const override
+	{
+		if (a_WorldInterface.GetDimension() == dimNether)
+		{
+			a_Chunk.SetBlock(a_RelPos, E_BLOCK_AIR, 0);
+		}
+		// Artificial light on any of the surrounding block > 11 leads to melting ice
+		bool Smelt = false;
+		Vector3i Pos;
+		cChunk *Chunk;
+
+		Vector3i Coords[7] =
+			{{1, 0, 0}, {-1, 0, 0},
+			{0, 1, 0}, {0, -1, 0},
+			{0, 0, 1}, {0, 0, -1},
+			{0, 0, 0}};
+
+		auto StartAbsPos = a_Chunk.RelativeToAbsolute(a_RelPos);
+
+		for (int i = 0; i < ARRAYCOUNT(Coords); ++i)
+		{
+			Pos = StartAbsPos + Coords[i];
+			Chunk = a_Chunk.GetNeighborChunk(Pos.x, Pos.z);
+			Smelt |= Chunk->GetBlockLight(Chunk->AbsoluteToRelative(Pos)) > 11;
+		}
+
+		if (Smelt)
+		{
+			a_Chunk.SetBlock(a_RelPos, E_BLOCK_STATIONARY_WATER, 0);
+		}
+	}
+
 	virtual void OnBroken(
 		cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface,
 		Vector3i a_BlockPos,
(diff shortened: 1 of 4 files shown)


src/Blocks/BlockIce.h
```diff
@@ -31,6 +31,44 @@ class cBlockIceHandler :
...
+		cWorldInterface & a_WorldInterface,
+		cBlockPluginInterface & a_PluginInterface,
+		cChunk & a_Chunk,
+		const Vector3i a_RelPos
+	) const override
+	{
+		if (a_WorldInterface.GetDimension() == dimNether)
+		{
+			a_Chunk.SetBlock(a_RelPos, E_BLOCK_AIR, 0);
+		}
```
tigerw, line 45: early return here perhaps?

src/Blocks/BlockIce.h
```diff
@@ -31,6 +31,44 @@ class cBlockIceHandler :
...
+		{
+			Pos = StartAbsPos + Coords[i];
+			Chunk = a_Chunk.GetNeighborChunk(Pos.x, Pos.z);
+			Smelt |= Chunk->GetBlockLight(Chunk->AbsoluteToRelative(Pos)) > 11;
+		}
+
+		if (Smelt)
+		{
+			a_Chunk.SetBlock(a_RelPos, E_BLOCK_STATIONARY_WATER, 0);
+		}
```
tigerw, line 69: I would suggest a couple changes: use std::array, keep coordinates relative, use ranged for loop, queue lighting, and early return; something like:

```cpp
std::array<Vector3i, 7> Adjacents = {
  // ...
};

for (const auto Offset : Adjacents)
{
  const auto Position = a_RelPos + Offset;
  const auto Chunk = a_Chunk.GetRelNeighborChunk();

  if (!Chunk->IsLightValid())
  {
    Chunk->GetWorld()->QueueLightChunk();
    return;
  }

  if (Chunk->GetBlockLight(Position) > 11)
  {
    Chunk->SetBlock(Position, WATER);
    return;
  }
}

File: src/Blocks/BlockIce.h [hunk 3]
```diff
@@ -31,6 +31,44 @@ class cBlockIceHandler :
...
+		// Artificial light on any of the surrounding block > 11 leads to melting ice
+		bool Smelt = false;
+		Vector3i Pos;
+		cChunk *Chunk;
+
+		Vector3i Coords[7] =
+			{{1, 0, 0}, {-1, 0, 0},
+			{0, 1, 0}, {0, -1, 0},
+			{0, 0, 1}, {0, 0, -1},
+			{0, 0, 0}};
```
tigerw, line 55: by the way is 0,0,0 needed?

12xx12: It did show some strange behaviour when I didn't add it. The wiki says only surrounding blocks but that didn't melt the blocks like in vanilla

src/Blocks/BlockPackedIce.h
```diff
@@ -0,0 +1,39 @@
...
+private:
+
+	virtual cItems ConvertToPickups(NIBBLETYPE a_BlockMeta, cBlockEntity * a_BlockEntity, const cEntity * a_Digger, const cItem * a_Tool) const override
+	{
+		// Only drop self when using silk-touch:
+		if (ToolHasSilkTouch(a_Tool))
+		{
+			return cItem(m_BlockType);
+		}
+		else
```
tigerw, line 28: Since there's a return already, this else can be removed, I suppose.

src/Blocks/BlockPackedIce.h
```diff
@@ -0,0 +1,39 @@
...
+#pragma once
+
+#include "BlockHandler.h"
+
+
+
+
+
+class cBlockPackedIceHandler :
+		public cBlockHandler
```
tigerw, line 11: `tabtab`! consider `tab` instead

src/Blocks/BlockPackedIce.h
12xx12: But the else is run if the tool was't with Silk touch

src/Blocks/BlockIce.h
tigerw: could be the lighting being invalid perhaps? does sound strange.

12xx12: Maybe. It just melted if the light source block was at the same y coords. Doing some more research tomorrow

src/Blocks/BlockPackedIce.h
12xx12: No i got it

src/Blocks/BlockIce.h
```diff
@@ -25,9 +25,47 @@ class cBlockIceHandler :
...
+		const std::array<Vector3i, 7> Offsets =
+			{Vector3i(1, 0, 0), Vector3i(-1, 0, 0),
+			Vector3i(0, 1, 0), Vector3i(0, -1, 0),
+			Vector3i(0, 0, 1), Vector3i(0, 0, -1)};
+
+		for (const auto & Offset : Offsets)
 		{
-			return {};
+			auto Position = a_RelPos + Offset;
+			const auto Chunk = a_Chunk.GetRelNeighborChunk(Position.x, Position.z);
```
tigerw, line 54: You can GetRelNeighborChunkAdjustCoords to make Position relative to the returned chunk

12xx12: Oh so that's what that method does :D

12xx12:
The smelting across chunk borders works now. Checked at the corner where 4 chunks met

tigerw:
Uno momento por favor

12xx12:
Yup, changes seem appropriate

src/Blocks/BlockIce.h
```diff
@@ -25,9 +25,54 @@ class cBlockIceHandler :
 		{
 			return cItem(m_BlockType);
 		}
```
12xx12, line 27: Now I am confused

tigerw: diff looks weird?

12xx12: I ment to mark then line with ````{{``` why those are needed

tigerw: The std::array {{?

I think the first { is to indicate initialisation of std::array, the second { indicates you want to aggregate-initialise the C-array the std::array contains, and the remaining {x,y,z} initialises the contained Vector3i

12xx12: ok - thank you

tigerw (approved):

Issue #4894, opened 2020-09-19
API: Segfault in destructor of cLuaWindow on shutdown


Seadragon91:
Client version: 1.12.2
Server OS: Windows/Linux
Cuberite Commit id: (272a8029e0b8280f61ad2d3bfd0e8353d4256aec)

### Expected behavior
Server should not crash on shutdown.

### Actual behavior
Server crashes on shutdown.

### Steps to reproduce the behavior
Create a plugin with this code, run console command `testit` and stop the server:
```lua
function Initialize(Plugin)
	Plugin:SetName("TestCode")
	Plugin:SetVersion(1)

	cPluginManager:BindConsoleCommand("testit",
		function(a_Split)
			cLuaWindow(cWindow.wtDropSpenser, 1, 1, 'JustAString')
			print("called")
			return true
		end, "testit")
	
	return true
end
```

### Stack trace
```
#0  __GI___pthread_mutex_lock (mutex=0x10) at ../nptl/pthread_mutex_lock.c:67
#1  0x000000000096cf43 in __gthread_mutex_lock (__mutex=0x10)
    at /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:748
#2  0x000000000096cf15 in __gthread_recursive_mutex_lock (__mutex=0x10)
    at /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:810
#3  0x000000000096d028 in std::recursive_mutex::lock (this=0x10)
    at /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/mutex:107
#4  0x000000000096cbb6 in cCriticalSection::Lock (this=0x0)
    at /home/lukas/cpp/cuberite/src/OSSupport/CriticalSection.cpp:23
#5  0x000000000096cd92 in cCSLock::Lock (this=0x7ffffffecc10)
    at /home/lukas/cpp/cuberite/src/OSSupport/CriticalSection.cpp:105
#6  0x000000000096cd17 in cCSLock::cCSLock (this=0x7ffffffecc10, a_CS=0x0)
    at /home/lukas/cpp/cuberite/src/OSSupport/CriticalSection.cpp:70
#7  0x000000000059c60f in cWorld::cLock::cLock (this=0x7ffffffecc10, a_World=...)
[...]

PR #4928, opened 2020-09-22
Clarify chunk destruction

tigerw:
The aim of this PR is to make sure cChunk's destructor doesn't access objects already destroyed. Destructors should be used for releasing resources under our control; doing extra work especially in delicate scenarios like shutdown is a recipe for crashes.

The crash in #4894 was due to the chunkmap, and thus all chunks, being explicitly (and unexpectedly) destroyed the moment cRoot got a stop signal and called cWorld::Stop. When the plugin manager attempted to unload plugins, cLuaWindow tried to enumerate players and hit an invalid chunkmap.

cWorld::Stop no longer jumps the gun by deleting the chunkmap early. Contrary to the comment, cEntity doesn't do any manipulation during destruction. Of the block entities, only Jukeboxes tried to access the world in its destructor to stop playing music. Thankfully music abruptly stopping on chunk unload is undesired behaviour and the client doesn't accept sound effect packets to unloaded chunks, so the destructor was safely deleted.

Entities need to broadcast to clients a 'destroy' packet when they are despawned, but only if there was a corresponding 'spawn' packet. Nothing needs to be broadcast during server shutdown. This is now all handled by cEntity::OnAddedToWorld/OnRemovedFromWorld.

To this effect cEntity::DestroyNoScheduling has been removed. Firstly, we need to always broadcast despawns since entities in unloaded chunks are known to misbehave: #3696. Additionally, we do not need to call Destroyed since cWorld::RemoveEntity will do it (and if the entity was merely in an unloaded chunk, it has not been Destroyed, only suspended).

The last point should fix two issues, firstly that players no longer broadcast two 'despawn' packets, one from cWorld::RemovePlayer, another from DestroyNoScheduling. Secondly, we no longer forget to reset pointers to destroyed entities when a leashed mob unloads.

### DIFF
--- a/src/BlockEntities/JukeboxEntity.cpp
+++ b/src/BlockEntities/JukeboxEntity.cpp
@@ -24,19 +24,6 @@ cJukeboxEntity::cJukeboxEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Ve
 
 
 
-cJukeboxEntity::~cJukeboxEntity()
-{
-	if (m_World && IsPlayingRecord())
-	{
-		// Stop playing music when destroyed by any means
-		m_World->BroadcastSoundParticleEffect(EffectID::SFX_RANDOM_PLAY_MUSIC_DISC, GetPos(), 0);
-	}
-}
-
-
-
-
-
 void cJukeboxEntity::Destroy(void)
 {
 	ASSERT(m_World != nullptr);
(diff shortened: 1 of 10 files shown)


src/BlockEntities/JukeboxEntity.cpp
```diff
@@ -24,19 +24,6 @@ cJukeboxEntity::cJukeboxEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Ve
 
 
 
-cJukeboxEntity::~cJukeboxEntity()
```
mathiascode, line 27: IIRC, in the past, `Destroy` wasn't called for block entities when doing WE actions, hence the reason this piece of code existed. Seems like this behavior was changed/fixed at some point though, so this code can be deleted.

tigerw: You're right, thanks for raising this.

bearbin (approved):

PR #4930, opened 2020-09-23
Fixing Mob spawner behaviour

### LINKED ISSUE #4550: Unbreak mob spawners
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: b7e88e2c9f96f31e29a85c3d7ef312fb4bd8c549

As far as I can tell, mob spawners currently don't spawn anything due to a couple of failing checks.

12xx12:
Fixes #4550 

There was a mistake in BlockEntity.GetChunkZ which resulted in wrong lookups not the player count is working

- [x] cMobSpawnerEntity::GetNearbyPlayersNum is broken (always returns 0)
- [x] cMobSpawnerEntity::GetNearbyMonsterNum is broken (always returns 0)
- [x] cMobSpawner::CanSpawnHere prevents mob spawner spawning, since it contains code to prevent mobs from spawning within 24 blocks of the player.
- [x] cMobSpawnerEntity::SpawnEntity spawns in the wrong coordinates (relative/absolute coordinate mess)

**Note: Some of this PR is just some housekeeping on variable names**

### DIFF
--- a/src/MobSpawner.cpp
+++ b/src/MobSpawner.cpp
@@ -72,7 +72,7 @@ eMonsterType cMobSpawner::ChooseMobType(EMCSBiome a_Biome)
 
 
 
-bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, Vector3i a_RelPos, eMonsterType a_MobType, EMCSBiome a_Biome)
+bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, Vector3i a_RelPos, eMonsterType a_MobType, EMCSBiome a_Biome, bool a_DisableSolidBelowCheck)
 {
 	if (a_Chunk == nullptr)
 	{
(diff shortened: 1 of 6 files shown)


tigerw:
oof, nice

12xx12:
The MonsterNum should work to. Gonna check this tomorrow

mathiascode:
Nice catch! No wonder I was banging my head on this one...

12xx12:
The can spawn here method seems to be broken anyway. I have monsters spawning at a spanner with me being 24 blocks or closer...

12xx12:
The only thing i am wondering about is that there are some random functions in the code that influence spawn behaviour. Is this realy needed?

src/MobSpawner.cpp
```diff
@@ -242,46 +238,46 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, Vector3i a_RelPos, eMonsterType
...
-				(blockAbove == E_BLOCK_AIR) &&
+				(TargetBlock == E_BLOCK_AIR) &&
+				(BlockAbove == E_BLOCK_AIR) &&
 				(
-					(blockBelow == E_BLOCK_GRASS) || (blockBelow == E_BLOCK_LEAVES) || (blockBelow == E_BLOCK_NEW_LEAVES)
+					(BlockBelow == E_BLOCK_GRASS) || (BlockBelow == E_BLOCK_LEAVES) || (BlockBelow == E_BLOCK_NEW_LEAVES)
 				) &&
 				(a_RelPos.y >= 62) &&
-				(random.RandBool(2.0 / 3.0))
+				(Random.RandBool(2.0 / 3.0))
```
12xx12, line 247: here

12xx12:
This need some longer term testing. I had a deadlock on the server

12xx12:
Seems to work - kept it running for two hours

src/MobSpawner.cpp
mathiascode: The intention is to make certain mobs more rare, although I'm not sure if these values are taken from some documentation or just tweaked until they felt right.

12xx12: I think this doesn't feel right doing this kind of check in that function.

I would perform this operation in the cMobSpawner::TryToSpawnHere function and/or make this customizeable

mathiascode: I agree. I think the whole mob spawning code would benefit from a proper review, as it has been largely unchanged since 2013 and most certainly has differences compared to vanilla nowadays. There are still outdated TODOs [like this](https://github.com/cuberite/cuberite/blob/master/src/Chunk.cpp#L587) in a few places.

Definitely something for a new PR though.

12xx12: So I'm gonna keep it like that for now and open an issue to do a cleanup on mob spawning

src/MobSpawner.cpp
```diff
@@ -467,6 +463,18 @@ std::set<eMonsterType> cMobSpawner::GetAllowedMobTypes(EMCSBiome a_Biome)
...
+	auto AbsPos = a_Chunk->RelativeToAbsolute(a_RelPos);
+	static const double RangeLimit = 24;
+	if (a_Chunk->GetWorld()->DoWithNearestPlayer(AbsPos, RangeLimit, [](cPlayer & a_Player) -> bool
+	{
+		return true;
+	})
+			)
+	{
+		return nullptr;
+	}
```
bearbin, line 476: Is this right? I thought players had to be close to mob spawners for them to work, and not the opposite?

12xx12: This is the code generating monsters independent from the mob spawner blocks.

The section you marked is just moved there from the method that checks if a mob could spawn "there" at that place. Which makes no sense as this code is shared by the natural generation and the mob spawner block. The natural generation checks now if there is a player before checking a mob can spawn at that position

bearbin: Derp, this is fine!

bearbin (approved):

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1343,7 +1343,7 @@ OwnedBlockEntity cWSSAnvil::LoadMobSpawnerFromNBT(const cParsedNBT & a_NBT, int
 	if ((Type >= 0) && (a_NBT.GetType(Type) == TAG_Short))
 	{
 		short MonsterType = a_NBT.GetShort(Type);
-		if ((MonsterType >= 50) && (MonsterType <= 120))
+		if (MonsterType != eMonsterType::mtInvalidType)
 		{
 			MobSpawner->SetEntity(static_cast<eMonsterType>(MonsterType));
```
mathiascode, line 1348: I'm not sure about this piece of code anymore. Since monster types no longer have a fixed value, this will probably break if we add new mobs. https://github.com/cuberite/cuberite/commit/01b8ed5295875262a91b60af878bf2a18c1b7aae#diff-c156838c409845aa17aefa4013707a49

mathiascode: Why do we have this special cause for Cuberite worlds to begin with?

12xx12: A quick note:

When I began with this PR the spawner weren't loaded properly.
I wondered why this checked for 50->120 so I removed it.

We could just dump the numerical saving and use the Minecraft:namespace way. This only should have a minor impact on startup/shutdown and the extra disc space doesn't matter.

12xx12: The thing breaking the whole thing is that if the check ```if ((MonsterType >= 50) && (MonsterType <= 120))``` fails the check for the string isn't even tried. We could also change it that way that we check for string anyway

mathiascode: I'm in favor of dropping the numerical saving and just sticking with the vanilla variant.

12xx12: All in favour raise you hand 🙋🏻‍♂️

12xx12: Side note: we already write both (string and short) into the save file so no problems on compability

tigerw: Make sure to add the relevant dataversion 2566 (1.16.2 I think) eMobType <-> namespaced IDs mapping to WorldStorage\NamespaceSerializer

12xx12: The Id conversion is done by ````cMonster::MobTypeToVanillaName```. Do we move this there?

12xx12: I had a look at the code doing the conversion... it's just iterating through a list of all availeble entities and returning the one matching - realy nice :D

New PR or is it ok to do this together?

12xx12: I am gonna do the back and forth conversion in another PR: There is some more stuff to do:
 - Move the mob enum to registries
 - add the matching functions to NamespaceSerializer
 - ...

12xx12: I did some research...

1.12 and earlier doesn't load the right entity because the name is written in ThisWay
1.13 doesn't even parse chunks...

tigerw: Hopefully once Anvil storage is fully upgraded we can leapfrog directly to full compatibility with 1.16 :)

12xx12: @mathiascode you are right. If we add another mob (and there many left) in alphabetical order we don't keep the same numerical IDs so the number idea is VERY bad as this is dependent of the order in the enum

mathiascode (approved):

PR #4931, opened 2020-09-23
Tile entity pickups

tigerw:
* Unify block entity pickup conversion

### DIFF
--- a/src/BlockEntities/BlockEntity.h
+++ b/src/BlockEntities/BlockEntity.h
@@ -85,7 +85,7 @@ public:
 	OwnedBlockEntity Clone(Vector3i a_Pos);
 
 	/** Returns the contents of this block entity that it would drop if broken.
-	Note that the block itself is not included; that's handled by the block handler. */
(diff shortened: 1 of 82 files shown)


src/BlockEntities/BlockEntity.h
```diff
@@ -83,6 +84,10 @@ class cBlockEntity
 	Uses CopyFrom() to copy the properties. */
 	OwnedBlockEntity Clone(Vector3i a_Pos);
 
+	/** Returns the contents of this block entity that it would drop if broken.
+	Note that the block itself is not included; that's handled by the block handler. */
+	virtual cItems ConvertToPickups() const;
```
bearbin, line 89: This comment doesn't really reflect reality - it should probably be something like

 > Note that the block handler will usually handle the block itself, in addition to any blocks returned here.

bearbin (commented):
LGTM after fixing the comment

tigerw:
Done, thanks. Intention to rebase

bearbin (approved):

PR #4932, opened 2020-09-23
Fortune Drops

KingCol13:
Adds fortune to the things I didn't in #4897 based on the wiki's fortune page.
Some of this was done in #4802, but that still has other useful changes.

I changed a push_back to emplace_back in Item.h, I hope that was ok.

### DIFF
--- a/src/Blocks/BlockCrops.h
+++ b/src/Blocks/BlockCrops.h
@@ -21,6 +21,24 @@ public:
 
 private:
 
+	char FortuneBinomialRandom(char a_Min, char a_Rolls) const
+	{
+		auto & random = GetRandomProvider();
+		char DropNum = a_Min;
+		for (unsigned char i=0; i<a_Rolls; i++)
+		{
+			if (random.RandBool(0.57))
+			{
+				DropNum++;
+			}
+		}
+		return DropNum;
+	}
+
+
+
+
+
 	virtual cItems ConvertToPickups(NIBBLETYPE a_BlockMeta, cBlockEntity * a_BlockEntity, const cEntity * a_Digger, const cItem * a_Tool) const override
 	{
 		auto & rand = GetRandomProvider();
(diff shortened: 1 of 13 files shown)


src/Item.h
```diff
@@ -239,7 +239,7 @@ class cItems  // tolua_export
 
 	void    Add   (short a_ItemType, char a_ItemCount, short a_ItemDamage)
 	{
-		push_back(cItem(a_ItemType, a_ItemCount, a_ItemDamage));
+		emplace_back(cItem(a_ItemType, a_ItemCount, a_ItemDamage));
```
tigerw, line 242: Might as well drop the cItem() since you're constructing it in-place with... emplace.

tigerw:
Looks reasonable at first glance, haven't understood everything yet.

I noticed a lot of cItem(type, 1, 0) as well as some potential rebase issues that reintroduced these forms; since the constructor defaults to 1 item with 0 meta, just cItem(type) would be clearer.

Also there's quite a few constructions of
```
if (X)
  return;
else
  // whatevs
```
where the else branch is superfluous.

src/Blocks/BlockCrops.h
```diff
@@ -21,6 +21,24 @@ class cBlockCropsHandler:
 
 private:
 
+	static char FortuneBinomialRandom(char a_Min, char a_Rolls)
```
bearbin, line 24: Since this function is only used here, I would rather it took 3 argumentsL min, baseRolls, and fortuneLevel; and do the addition internally. 

It would also be better named something like CalculateDropCount - it feels slightly wrong for BlockCrops.h to have the so-named source of truth on fortune calculations.

bearbin (changes requested):

bearbin (approved):

src/Blocks/BlockOre.h
```diff
@@ -33,16 +33,16 @@ class cBlockOreHandler:
 
 		auto & Random = GetRandomProvider();
```
KingCol13, line 34: These changes are all good improvements to the readability, thank you. This line can be moved down a bit now too.

src/Blocks/BlockGravel.h
```diff
@@ -23,10 +23,9 @@ class cBlockGravelHandler :
 			return cItem(E_BLOCK_GRAVEL);
 		}
 
-		// Denominator of probability from wiki, don't let it go below 1
-		double Denominator = std::max(10 - (3 * ToolFortuneLevel(a_Tool)), 1);
-		auto & Random = GetRandomProvider();
-		if (Random.RandBool(1 / Denominator))
+		// Denominator of probability from wiki, don't let it go below 1.
+		const auto Denominator = std::max(10U - 3U * ToolFortuneLevel(a_Tool), 1U);
```
bearbin, line 27: Now here is a place you _don't_ want to use unsigned integers! This now does the opposite of what it should do. Integer by default is a sensible idea, because subtractions exist!

src/Blocks/BlockGravel.h
```diff
@@ -23,10 +23,9 @@ class cBlockGravelHandler :
...
+		if (GetRandomProvider().RandBool(1.0 / Denominator))
```
KingCol13, line 28: Does this cast to a float at the correctly?

bearbin (changes requested):
@tigerw not just style, also introducing bugs!

src/Blocks/BlockGravel.h
tigerw: Yeah it does floating point division

bearbin (approved):

tigerw:
Thanks KingCol!

Issue #2177, opened 2015-05-31
Enchanting Table doesn't write the enchantment that you will have


Rakete1111:
It uses 1.7 behavior, paired with 1.8 behavior.

12xx12:
The author ment that the tootips which should be shown on hovering the enchantment options in the window are not displayed

12xx12:
the function cSlotAreaEnchanting::UpdateResult(cPlayer & a_Player) should send the message

KingCol13:
I will attempt to fix this. Notes from #4934 :

A prerequisite to fixing this is fixing the PRNG for offered enchantments such that they cannot be altered: https://minecraft.gamepedia.com/Enchanting#Affecting_offered_enchantments.

A 4 byte seed should be stored in cPlayer and written on disconnect. This should be used to seed the PRNG state when calculating which enchantments to offer. This seed should only be incremented when an item is enchantment by that player.

Pseudocode for the PRNG can be found at: https://minecraft.gamepedia.com/Enchanting_mechanics. An alternative is:

`auto Noise = cNoise(PlayerSeed);
auto RandomNumer = Noise.IntNoise1D(SomeNumber);`

where SomeNumber should be derived from the values listed on the wiki link above.

Issue #988, opened 2014-05-07
Add hook for enchanting an item


madmaxoft:
When the item is enchanted, there's no way for the plugins to change the outcome. Add a (set of) hook(s) to give plugins full control over the enchanting process. Possibly the hooks should be similar to crafting - one hook to fire before the enchanting, to allow plugins to specify their very own enchantments or disable enchanting, and one hook to fire after the enchantments are chosen by the server, so that the plugins may tweak them. Follow the standard -ING / -ED naming convention.

PR #4934, opened 2020-09-24
Enchanting consumes lapis, removes correct number of levels

KingCol13:
Change to the 1.8+ behaviour of taking lapis and only removing 1-3 levels.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -795,7 +795,7 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 		const auto LapisRequired = a_Enchantment + 1;
 
 		// Only allow enchantment if the player is creative or has sufficient levels and lapis to enchant
-		if (IsCreative || ((m_Player->GetCurrentXp() >= XpRequired) && LapisStack.GetCount() >= LapisRequired))
+		if (IsCreative || ((m_Player->GetCurrentXp() >= XpRequired) && (LapisStack.m_ItemCount >= LapisRequired)))
 		{
 			// If they weren't creative then subtract xp and lapis
 			if (!IsCreative)
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -773,6 +773,7 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 		return;
 	}
 
+	// Bail out if something's wrong with the window
 	if (
 		(m_Player->GetWindow() == nullptr) ||
 		(m_Player->GetWindow()->GetWindowID() != a_WindowID) ||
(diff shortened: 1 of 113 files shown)


12xx12:
maybe consider this #2177 
I did some notes on this
and maybe this #988

KingCol13:
Alright I’ll look into them thanks. maybe for a follow up PR though because changing the items in the slots seems to change the available enchanting options which would the let you keep re-rolling for free.

12xx12:
Yeah I thought about this. Maybe let one of the objects (enchanting table, window, inventory,...) keeping track of that.

Do you know if you put some other item of the same type in the enchanting table you get other enchantments in the given item?

src/Item.h
```diff
@@ -101,6 +101,9 @@ class cItem
 	bool IsCustomNameEmpty(void) const { return (m_CustomName.empty()); }
 	bool IsLoreEmpty(void) const { return (m_LoreTable.empty()); }
 
+	/** Get the number of items in the stack. */
+	char GetCount(void) { return m_ItemCount; }
```
12xx12, line 105: ```m_ItemCount``` is ```public``` for ```cItem```.

So do ```Item.m_ItemCount```
Or add the function to the LuaAPIDoc

KingCol13: How do I do that? I'd guess it probably should be exposed to plugins

12xx12: This is accessible by lua - the member is also bublic there

Any LuaAPI aditions have to be added in Server/Plugins/APIDump
either in APIDesc.lua or in one of the folders mathching the class. And they should be kept in alphabetical order

KingCol13:
Testing in 1.16.3, the available enchantments and the required levels do not change under any condition except when an item is enchanted. The PRNG state could not be stored in the window or enchanting table as even if these are destroyed, the available enchantments are the same. Also the available enchantments are the same for all tools (so if I could currently enchant an axe with fortune 3, then I would also be able to enchant a pickaxe with fortune 3). So the enchanting PRNG state should probably be stored in the player object, and also saved when they disconnect. The wiki actually provides some pseudocode: https://minecraft.gamepedia.com/Enchanting_mechanics.

12xx12:
The wiki talks about a seed for enchantments.

There is a Object called cNoise whicht supports noise base operations.

So maybe we base the random stuff of this and generate each time an item is taken from the enchanting table a seed and use this seed the as base for the cNoise operations.

For example:
```
auto Noise = cNoise(PlayerSeed);
auto RandomNumer = Noise.IntNoise1D(SomeNumber);
```

src/Item.h
KingCol13: I won't make the member private but I think we should encourage the use of getters/setters as a safer way to interact with plugins.

12xx12: Yes You are right. I also wondering about that.

I think the best is to open a thread on the forum and lay down your thoughts there so that everyone can comment on that as items are quite essential to the gameplay/server

src/ClientHandle.cpp
```diff
@@ -788,12 +789,39 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
...
+				LapisStack.AddCount(-LapisRequired);
+				Window->m_SlotArea->SetSlot(1, *m_Player, LapisStack);
+			}
+			// Set the item slot to our new enchanted item
 			Window->m_SlotArea->SetSlot(0, *m_Player, Item);
-			Window->SendSlot(*m_Player, Window->m_SlotArea, 0);
 			Window->BroadcastWholeWindow();
 
+			// Remove enchantment choices
 			Window->SetProperty(0, 0, *m_Player);
```
12xx12, line 825: The properties are used to send the possible enchantment to the player
see here: https://wiki.vg/index.php?title=Protocol&oldid=14204#Window_Property

12xx12:
> The possible enchantments depend on the player's enchantment seed, the item type, and material, and the enchantment level 
- wiki

So the seed is stored int the cPlayer.

We could use SomeNumber to take acount of the other parameters. The Item type is just a number so you could do 
```
auto RandomNumer = Noise.IntNoise1D(ItemToEnchant.m_ItemType);
```

KingCol13:
I agree this should be added, and I'm happy to do it as I've already read everything you've linked to me while I was making this PR. I just think this PR works as a stand-alone incremental improvement and I would like it to be merged separately to these future improvements.

Revealing the offered enchantments looks easy (setting a window property) but it should only come after I've made the necessary changes to the PRNG to avoid exploitation.

12xx12:
I just noticed that my additions would require some further changes in cEnchantments - sorry for bothering you

bearbin (approved):
This looks like a valuable and sensible change.

The only thing I'm a bit unsure about is the getter function, as generally I'm of the opinion that simple getter functions are rather pointless in C++; however it could make sense in the Lua API. Either way, I think this is fine to merge.

bearbin:
@tigerw This is ready to merge - is your review still pending?

tigerw:
beabin thanks

I would advise against adding GetCount for now, the established method is to access m_ItemCount.

KingCol13:
> I would advise against adding GetCount for now, the established method is to access m_ItemCount.

You mean the established member variable har har har ;). I'll remove my calls to it, should I get rid of the method too?

tigerw:
I guess yeah. We need to overhaul cItem to support 1.13 NBT storage, we can discuss how the API should change when that happens.

KingCol13:
> I guess yeah. We need to overhaul cItem to support 1.13 NBT storage, we can discuss how the API should change when that happens.

Seems reasonable, done :)

PR #4937, opened 2020-09-25
Enchanting table shows detail on hover. Enchanting is deterministic.

### LINKED ISSUE #2177: Enchanting Table doesn't write the enchantment that you will have

KingCol13:
This is based on #4934 , diff will be simplified once that is merged.

On QM Einstein wrote:
> God does not play dice with the universe.

While Bell's theorem largely put an end to that opinion, we will follow the lead of Einstein's god by no longer playing dice with enchantments.

An `EnchantmentSeed` is generated when the player first joins the server, this is then used to seed all PRNGs involved in enchanting items. When an item is placed in the enchanting table, the possible enchantment options are generated using the `EnchantmentSeed`, stored in the `cSlotAreaEnchanting` and used to set the hover over info. When the player chooses an option, the corresponding enchanted item is retrieved, and their `EnchantmentSeed` is updated. The `EnchantmentSeed` is saved/loaded on disconnect/reconnect.

This fixes #2177 .

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -772,14 +772,14 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 		return;
 	}
 
-	// Bail out if something's wrong with the window
+	// Bail out if something's wrong with the window:
 	if (
 		(m_Player->GetWindow() == nullptr) ||
 		(m_Player->GetWindow()->GetWindowID() != a_WindowID) ||
 		(m_Player->GetWindow()->GetWindowType() != cWindow::wtEnchantment)
 	)
 	{
-		LOGD("cClientHandle received a non valid enchanting window.");
+		Kick("Enchantment with invalid window - hacked client?");
 		return;
 	}
 
@@ -824,7 +824,7 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 	}
 
 	// Retrieve the enchanted item corresponding to our chosen option (top, middle, bottom)
-	cItem EnchantedItem = Window->m_SlotArea->GetEnchantedOption(a_Enchantment);
+	cItem EnchantedItem = Window->m_SlotArea->SelectEnchantedOption(a_Enchantment);
 
 	// Set the item slot to our new enchanted item:
 	Window->m_SlotArea->SetSlot(0, *m_Player, EnchantedItem);
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -784,61 +784,57 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 	}
 
 	cEnchantingWindow * Window = static_cast<cEnchantingWindow *>(m_Player->GetWindow());
-	short BaseEnchantmentLevel = Window->GetPropertyValue(a_Enchantment);
+	const auto BaseEnchantmentLevel = Window->GetPropertyValue(a_Enchantment);
+
+	// Survival players must be checked they can afford enchantment and have lapis removed
+	if (!m_Player->IsGameModeCreative())
+	{
+		const auto XpRequired = m_Player->XpForLevel(BaseEnchantmentLevel);
+		auto LapisStack = *Window->m_SlotArea->GetSlot(1, *m_Player);  // A copy of the lapis stack.
+		const auto LapisRequired = a_Enchantment + 1;
+
+		// Only allow enchantment if the player has sufficient levels and lapis to enchant:
+		if ((m_Player->GetCurrentXp() >= XpRequired) && (LapisStack.m_ItemCount >= LapisRequired))
+		{
+			/** We need to reduce the player's level by the number of lapis required.
+			However we need to keep the resulting percentage filled the same. */
+
+			const auto TargetLevel = m_Player->GetXpLevel() - LapisRequired;
+			const auto CurrentFillPercent = m_Player->GetXpPercentage();
+
+			// The experience to remove in order to reach the start (0% fill) of the target level.
+			const auto DeltaForLevel = -m_Player->GetCurrentXp() + m_Player->XpForLevel(TargetLevel);
+
+			// The experience to add to get the same fill percent.
+			const auto DeltaForPercent = CurrentFillPercent * (m_Player->XpForLevel(TargetLevel + 1) - m_Player->XpForLevel(TargetLevel));
+
+			// Apply the experience delta:
+			m_Player->DeltaExperience(DeltaForLevel + DeltaForPercent);
+
+			// Now reduce the lapis in our stack and send it back:
+			LapisStack.AddCount(-LapisRequired);
+			Window->m_SlotArea->SetSlot(1, *m_Player, LapisStack);
+		}
+		else
+		{
+			// Not creative and can't afford enchantment, so exit
+			return;
+		}
+	}
+
 	// Retrieve the enchanted item corresponding to our chosen option (top, middle, bottom)
 	cItem EnchantedItem = Window->m_SlotArea->GetEnchantedOption(a_Enchantment);
 
-	const auto SetEnchantAndBroadcast = [this, &EnchantedItem, Window]
-	{
-		// Set the item slot to our new enchanted item:
-		Window->m_SlotArea->SetSlot(0, *m_Player, EnchantedItem);
-		m_Player->PermuteEnchantmentSeed();
+	// Set the item slot to our new enchanted item:
+	Window->m_SlotArea->SetSlot(0, *m_Player, EnchantedItem);
+	m_Player->PermuteEnchantmentSeed();
 
-		// Reset window properties
-		for (short i=0; i<10; i++)
-		{
-			Window->SetProperty(i, 0, *m_Player);
-		}
-		Window->BroadcastWholeWindow();
-	};
-
-	// Creative players can always enchant:
-	if (m_Player->IsGameModeCreative())
+	// Reset window properties
+	for (short i=0; i<10; i++)
 	{
-		SetEnchantAndBroadcast();
-		return;
+		Window->SetProperty(i, 0, *m_Player);
 	}
-
-	const auto XpRequired = m_Player->XpForLevel(BaseEnchantmentLevel);
-	auto LapisStack = *Window->m_SlotArea->GetSlot(1, *m_Player);  // A copy of the lapis stack.
-	const auto LapisRequired = a_Enchantment + 1;
-
-	// Only allow enchantment if the player has sufficient levels and lapis to enchant:
-	if ((m_Player->GetCurrentXp() >= XpRequired) && (LapisStack.m_ItemCount >= LapisRequired))
-	{
-		/*
-		We need to reduce the player's level by the number of lapis required.
-		However we need to keep the resulting percentage filled the same.
-		*/
-
-		const auto TargetLevel = m_Player->GetXpLevel() - LapisRequired;
-		const auto CurrentFillPercent = m_Player->GetXpPercentage();
-
-		// The experience to remove in order to reach the start (0% fill) of the target level.
-		const auto DeltaForLevel = -m_Player->GetCurrentXp() + m_Player->XpForLevel(TargetLevel);
-
-		// The experience to add to get the same fill percent.
-		const auto DeltaForPercent = CurrentFillPercent * (m_Player->XpForLevel(TargetLevel + 1) - m_Player->XpForLevel(TargetLevel));
-
-		// Apply the experience delta:
-		m_Player->DeltaExperience(DeltaForLevel + DeltaForPercent);
-
-		// Now reduce the lapis in our stack and send it back:
-		LapisStack.AddCount(-LapisRequired);
-		Window->m_SlotArea->SetSlot(1, *m_Player, LapisStack);
-
-		SetEnchantAndBroadcast();
-	}
+	Window->BroadcastWholeWindow();
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -792,6 +792,7 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 	{
 		// Set the item slot to our new enchanted item:
 		Window->m_SlotArea->SetSlot(0, *m_Player, EnchantedItem);
+		m_Player->PermuteEnchantmentSeed();
 
 		// Reset window properties
 		for (short i=0; i<10; i++)
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -779,29 +779,26 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 		(m_Player->GetWindow()->GetWindowType() != cWindow::wtEnchantment)
 	)
 	{
+		LOGD("cClientHandle received a non valid enchanting window.");
 		return;
 	}
 
 	cEnchantingWindow * Window = static_cast<cEnchantingWindow *>(m_Player->GetWindow());
-	auto Item = *Window->m_SlotArea->GetSlot(0, *m_Player);  // A copy of the item to be enchanted.
 	short BaseEnchantmentLevel = Window->GetPropertyValue(a_Enchantment);
+	// Retrieve the enchanted item corresponding to our chosen option (top, middle, bottom)
+	cItem EnchantedItem = Window->m_SlotArea->GetEnchantedOption(a_Enchantment);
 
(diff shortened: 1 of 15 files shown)


src/UI/SlotArea.cpp
```diff
@@ -1680,30 +1681,73 @@ void cSlotAreaEnchanting::UpdateResult(cPlayer & a_Player)
...
-		int MiddleSlot = (Base * 2) / 3 + 1;
-		int BottomSlot = std::max(Base, Bookshelves * 2);
+		int OptionLevels[3];
+		OptionLevels[0] = std::max(Base / 3, 1);
+		OptionLevels[1] = (Base * 2) / 3 + 1;
+		OptionLevels[2] = std::max(Base, Bookshelves * 2);
+
+		// Properties set according to: https://wiki.vg/Protocol#Window_Property
+		// Send bits of the seed to the client so it can write a bunch of BS in squiggly letters
+		m_ParentWindow.SetProperty(3, static_cast<short>(EnchantmentSeed & 0xFFFFFFF0), a_Player);
```
KingCol13, line 1700: https://wiki.vg/Protocol#Window_Property says this is the bitmask applied (4 bytes) but it the protocol uses shorts (2 bytes) so I'm a little confused.

src/ClientHandle.cpp
```diff
@@ -773,31 +773,64 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
...
+		// Reset window properties
+		for (short i=0; i<10; i++)
+		{
+			Window->SetProperty(i, 0, *m_Player);
 		}
+
+		// Permute the player's enchantment seed, this should probably be more fancy
+		auto EnchantmentSeed = m_Player->GetEnchantmentSeed();
+		MTRand Random(EnchantmentSeed);
+		m_Player->SetEnchantmentSeed(Random.RandInt<unsigned int>());
```
KingCol13, line 833: Do we want to make this any more fancy, i.e. combine with the the exact item chosen to enchant or something? Might not matter.

src/Enchantments.h
```diff
@@ -139,8 +142,9 @@ class cEnchantments
 	/** Check enchantment conflicts from enchantments from the vector */
 	static void CheckEnchantmentConflictsFromVector(cWeightedEnchantments & a_Enchantments, const cEnchantments & a_FirstEnchantment);
 
-	/** Gets random enchantment from Vector and returns it */
-	static cEnchantments GetRandomEnchantmentFromVector(const cWeightedEnchantments & a_Enchantments);
+	/** Gets random enchantment from Vector and returns it
+	Passing an MTRand PRNG pointer is optional. */
+	static cEnchantments GetRandomEnchantmentFromVector(const cWeightedEnchantments & a_Enchantments, MTRand * Random = nullptr);
```
KingCol13, line 147: Are we happy passing `MTRand` like this?

12xx12: Usually GetRandomProvider() Is used

```auto Random = GetRandomProvider(); ```

This accesible from anywehre

File: src/ClientHandle.cpp (hunk 2 above)
Comment (12xx12) [reply src/ClientHandle.cpp:833]: I personally think we don't need to. Random is random 😅

File: src/UI/SlotArea.cpp (hunk 1 above)
Comment (12xx12) [reply src/UI/SlotArea.cpp:1700]: The whole point of the & operation is just to prevent users from deducting the enchantments.

I would say this is your choice how you mask the seed on sending.

I am also wondering why the byte count does mismatch but hey - bugs became features in the past 😁

File: src/ClientHandle.cpp (hunk 2 above)
Comment (KingCol13) [reply src/ClientHandle.cpp:833]: Yea they shouldn't be able to calculate what enchantments they get next without their enchantment seed anyway.

File: src/UI/SlotArea.cpp (hunk 1 above)
Comment (KingCol13) [reply src/UI/SlotArea.cpp:1700]: Yep and casting down from uint32 to int16 already hides half the seed to stop deduction haha.

File: src/Enchantments.h (hunk 3 above)
Comment (tigerw) [reply src/Enchantments.h:147]: Sure but make it MTRand &

File: src/Entities/Player.cpp [hunk 4]
```diff
@@ -2277,6 +2279,11 @@ bool cPlayer::LoadFromFile(const AString & a_FileName, cWorldPtr & a_World)
 	m_LifetimeTotalXp     = root.get("xpTotal",        0).asInt();
 	m_CurrentXp           = root.get("xpCurrent",      0).asInt();
 	m_IsFlying            = root.get("isflying",       0).asBool();
+	m_EnchantmentSeed     = root.get("enchantmentSeed", 0).asUInt();
```
tigerw, line 2282: Can you pass the default, instead of 0, as the random number?

src/Entities/Player.h
```diff
@@ -263,6 +263,12 @@ class cPlayer:
 
 	// tolua_end
 
+	/** Get the seed for enchanting related PRNGs, don't use this for other purposes. */
+	inline unsigned int GetEnchantmentSeed(void) { return m_EnchantmentSeed; }
+
+	/** Set the seed for enchanting related PRNGs, don't use this for other purposes. */
+	inline void SetEnchantmentSeed(unsigned int a_EnchantmentSeed) { m_EnchantmentSeed = a_EnchantmentSeed; }
```
tigerw, line 270: "OnItemEnchanted" or "PermuteEnchantmentSeed", then the function does the permutation itself, would be better than a setter.

src/Item.h
```diff
@@ -141,8 +141,9 @@ class cItem
 	int GetEnchantability();  // tolua_export
 
 	/** Randomly enchants the item using the specified number of XP levels.
-	Returns true if the item was enchanted, false if not (not enchantable / too many enchantments already). */
-	bool EnchantByXPLevels(int a_NumXPLevels);  // tolua_export
+	Returns true if the item was enchanted, false if not (not enchantable / too many enchantments already).
+	Passing an MTRand PRNG pointer is optional. */
+	bool EnchantByXPLevels(int a_NumXPLevels, MTRand * Random = nullptr);  // tolua_export
```
tigerw, line 146: MTRand ain't exported to Lua, could you make this MTRand & and export manually in ManualBindings.cpp?

src/UI/EnchantingWindow.cpp
```diff
@@ -30,7 +30,7 @@ void cEnchantingWindow::SetProperty(short a_Property, short a_Value, cPlayer & a
 {
 	if ((a_Property < 0) || (static_cast<size_t>(a_Property) >= ARRAYCOUNT(m_PropertyValue)))
 	{
-		ASSERT(!"a_Property is invalid");
+		LOGD("a_Property is invalid");
```
tigerw, line 33: Is this a debugging leftover?

src/UI/EnchantingWindow.h
```diff
@@ -33,10 +33,10 @@ class cEnchantingWindow:
 
 	virtual void DistributeStack(cItem & a_ItemStack, int a_Slot, cPlayer & a_Player, cSlotArea * a_ClickedArea, bool a_ShouldApply) override;
 
-	cSlotArea * m_SlotArea;
+	cSlotAreaEnchanting * m_SlotArea;
 
 protected:
-	short m_PropertyValue[3];
+	short m_PropertyValue[10];
```
tigerw, line 39: std:;array<short, 10>

src/UI/SlotArea.cpp
```diff
@@ -1680,30 +1681,74 @@ void cSlotAreaEnchanting::UpdateResult(cPlayer & a_Player)
...
+		// Initialise the PRNG using the player's enchantment seed
+		auto EnchantmentSeed = a_Player.GetEnchantmentSeed();
+		MTRand Random(EnchantmentSeed);
+
+		// Calculate the levels for the offered enchantment options:
 		int Base = (Random.RandInt(1, 8) + (Bookshelves / 2) + Random.RandInt(0, Bookshelves));
-		int TopSlot = std::max(Base / 3, 1);
-		int MiddleSlot = (Base * 2) / 3 + 1;
-		int BottomSlot = std::max(Base, Bookshelves * 2);
+		int OptionLevels[3];
```
tigerw, line 1693: std::array<unsigned, 3> OptionLevels
{
...
};

src/UI/SlotArea.h
```diff
@@ -397,14 +397,18 @@ class cSlotAreaEnchanting:
 	virtual void OnPlayerAdded  (cPlayer & a_Player) override;
 	virtual void OnPlayerRemoved(cPlayer & a_Player) override;
 
-	/* Get the count of bookshelves who stand in the near of the enchanting table */
+	/* Get the number of bookshelves which are near the enchanting table */
 	int GetBookshelvesCount(cWorld & a_World);
 
+	/* Return the enchanted item matching the chosen option (0, 1, 2) */
+	cItem GetEnchantedOption( unsigned char a_EnchantOption);
```
tigerw, line 404: size_t a_EnchantOption

src/UI/SlotArea.h
```diff
@@ -397,14 +397,18 @@ class cSlotAreaEnchanting:
...
+
 protected:
 	/** Handles a click in the item slot. */
 	void UpdateResult(cPlayer & a_Player);
 
 	Vector3i m_BlockPos;
+	cItems m_EnchantedItemOptions;
```
tigerw, line 411: I guess a std::array<cItem, 3> would suffice

tigerw:
Thanks! Could you rebase this onto master?

src/UI/EnchantingWindow.cpp
KingCol13: It might be by whoever left those broken ASSERTs there, I can remove them? or I can make them into proper asserts

src/Enchantments.h
KingCol13: I used a pointer to indicate it was optional. I'd heard it was convention to pass a pointer if it could be nullptr, and a reference if it had to exist.

src/Entities/Player.cpp
KingCol13: I was trying to avoid an unused rand generation but I don't think it matters so yep :)

src/UI/EnchantingWindow.cpp
tigerw: The ASSERT looks okay, why were they broken?

KingCol13: ah you're right, I think I'll change them to `ASSERT(bool)` instead if `if (!bool) { assert(!message) }` though

bearbin: But then you lose the message in the alert if you just pass in a bool.

KingCol13: alrighty I'll just change it back

src/Enchantments.h
KingCol13: In #4842 madmaxoft said:

> As for by-pointer vs by-reference vs by-value:
Stuff passed by-pointer means that it may be optional (nullptr). When you pass by-reference, you maek it clear to the caller that the parameter must be valid. Filling with loot requires a valid world, so we make it clear by passing by-reference. By-value passing is used for tiny stuff that is cheap to copy (can fit into a few registers) - basically plain numbers and Vector3<> instances were decided to be small enough to always pass by-value.

src/Item.h
KingCol13: Is there a way to only export the function with the default parameter and prevent them from passing the random generator? I don't think they need access to MTRand (and I've no clue how it's done).

tigerw: Let me help with the export

src/Enchantments.h
tigerw: You're passing in a MTRand for every invocation of GetRandomEnchantmentetc right? Or have I misread.

KingCol13: I am for everything related with enchantments so I can make sure the result will always be the same since I'm passing the same PRNG state. However maybe there'd be use cases where you don't care about the PRNG state or reproducibility (maybe for chest loot or plugin calls).

KingCol13: The more I think, the more I realise I've been overcomplicating this. It's not much to expect the caller to call GetRandomSource() for us if needs be.

KingCol13:
Finished rebasing @tigerw, thanks for the reviews :+1: .

Personally I prefer MTRand as an optional argument and one not exposed to lua (as I can't see MTRand being used in a plugin) but if you disagree then do whatever you think is right. Same with the handleEnchant refactor, feel free to revert.

src/UI/SlotArea.h
```diff
@@ -397,14 +397,19 @@ class cSlotAreaEnchanting:
...
+	/* Return the enchanted item matching the chosen option (0, 1, 2) */
+	cItem GetEnchantedOption(size_t a_EnchantOption);
 
 protected:
+
 	/** Handles a click in the item slot. */
 	void UpdateResult(cPlayer & a_Player);
 
 	Vector3i m_BlockPos;
+	std::array<cItem, 3> m_EnchantedItemOptions;
```
tigerw, line 412: What do you think about computing the enchantment in cClientHandle::HandleEnchant insteading of precomputing and storing?

KingCol13: It's possible, but that would mean computing it twice. Once when an item is placed in the slot in order to display possible enchantment options and again when an option is clicked.

src/Entities/Player.cpp
```diff
@@ -1861,6 +1863,17 @@ void cPlayer::SetVisible(bool a_bVisible)
 
 
 
+void cPlayer::PermuteEnchantmentSeed()
+{
+	// Get the next rand uint and use that as the next seed
+	MTRand Random(m_EnchantmentSeed);
+	m_EnchantmentSeed = Random.RandInt<unsigned int>();
```
tigerw, line 1870: Why do we need to make a new MTRand seeded with m_EnchantmentSeed, couldn't we just make a new random seed from GetRandomProvider()?

src/Enchantments.cpp
```diff
@@ -126,6 +126,22 @@ unsigned int cEnchantments::GetLevel(int a_EnchantmentID) const
 
 
 
+int cEnchantments::GetFirstEnchantmentID(void)
+{
+	for (cEnchantments::cMap::const_iterator itr = m_Enchantments.begin(), end = m_Enchantments.end(); itr != end; ++itr)
+	{
+		if (itr->second>0)
```
tigerw, line 133: Is there any case where this is zero?

src/Entities/Player.cpp
KingCol13: We can and probably should used GetRandomProvider. I had misunderstood the enchanting mechanics when I wrote that.

src/Enchantments.cpp
KingCol13: Not as far as I know, I'll change this

tigerw:
Thanks! I'm going to try two last changes, tell me if they sound reasonable

1. Since cEnchantments has a .begin() member, use that instead of GetFirst. I think GetFirst is too specific to be exported and may lead to confusion
2. Try removing the loops resetting window properties, since it seems like the client does it by itself, and we overwrite the properties anyway on the next enchant

KingCol13:
@tigerw yep they both sound very reasonable. Thanks for all the help and sorry about my git messy git struggles haha :)

src/UI/SlotArea.h
KingCol13: We’d have to be careful with the PRNG state too if we did this, or reinitialise from the seed for each option in cSlotArea::update to make the re-calculation match.

tigerw: hmm, good points.

I asked because I was thinking there'd be no point in storing both the enchanted results and the window properties. We may be able to avoid storing the std::array<short, 10> stuff altogether, if we move the "sufficient experience/lapis" calculations into UpdateResult.

The client greys out unavailable enchantments, so HandleEnchant won't get them barring a hacked client. If UpdateResult does the lapis calculations, sets the available enchants into m_EnchantedItemOptions and empties the unavailable ones, then HandleEnchant can just read them

KingCol13: Yep I was just about to say, storing the 3 shorts and the m_EnchantedItemOptions is probably the best. We don't need the other 7 shorts anywhere. I think it's nice to try to be secure against hacked clients where cheap and easy.

KingCol13: Also I haven't done any multiplayer testing. Do you know what happens if two people were to enchant with the same table simultaneously? Do they each get their own window object?

Pokechu22: IIRC yes, like crafting tables and other inventories that don't store items in a block entity (and drop/return the items when the window is closed) give each player their own own window object.  Enchanting tables are just special because they *have* a block entity, but IIRC it's only used for rendering clientside and doesn't have any special behavior serverside.

KingCol13: Just quickly tested and confirmed :)

tigerw (approved):

Issue #4562, opened 2020-03-27
Snow behaviour


Xenoxis:
Client version: 1.12.2
Server OS: Ubuntu 16.04 i386
Cuberite Commit id: 2b32bb0b1928704499aa187ed3bff7433264e742

### Expected behavior
Broking snow on the floor with spade, and then place an a block should work.
Attempt to place a block on snow will automatically broke the snow block and place yours instead

### Actual behavior
Currently, when you place a block after broking snow, the snow block reappears and your block isn't placed.
If you try to place block on snow, your block is placed above the snow block.

### Steps to reproduce the behavior
Found some snow biomes, use tool to broke the snow, then try to place blocks

Xenoxis:
It seems to appears with only the spade (high breaking speed for snow) and when you broke more than one snow block without released left-click

mathiascode:
Something in https://github.com/cuberite/cuberite/pull/4417 broke snow behavior.

KingCol13:
Fixed in #4938 .

Issue #2299, opened 2015-06-28
Pickaxe: Efficiency III is broken


Nanosync:
The state of the game will be out of sync when mining Netherrack. However, this works fine on Nether quartz.

edit: apparently it also glitches sometimes with the reason "too many blocks were destroyed per unit time - hacked client?". This glitch can be fixed by asking the client to restart the entire game.

bearbin:
I think because netherrack is a 0/1-tick block to mine with efficiency 3, but nether quartz takes longer

jammet:
Efficiency also is a problem on max level/diamond shovel, when digging dirt. The blocks disappear in the client, but some of them will return after a reconnect. These things are *too* effective somehow, the server doesn't seem to be able to keep up.

KingCol13:
Confirmed fixed by #4938.

Issue #2121, opened 2015-05-23
Sometimes blocks are only destroyed client side.


SafwatHalaby:
This happened to me twice.
I am currently testing the PathFinder, and sometimes I destroy blocks to see if the mobs will alternate their path. On 2 occasions, the mobs did not do that, instead, they walked "in the air", as if the block wasn't destroyed. On those two occasions, I logged back out then logged back in, and the block that I thought I had removed was there. So the PathFinder was working fine, it's just that the server did not register the block destruction.

I often work with mobs, this could be a trigger. Perhaps blocks have a chance of behaving weirdly when destroyed while a mob is standing on them / touching them?

I can rarely reproduce this, so it could be troublesome to debug.

SafwatHalaby:
Client 1.8.4

jammet:
I know enchantments aren't *really* fully supported yet, but if you enchant a pick with the /enchant command to maximum efficiency and then use that on netherrock... you won't even see the blocks. It completely erases the blocks without spawning them for pickup.
Same with dirt, if you use a shovel with maximum efficiency.

Maybe that's related.

KingCol13:
Possibly fixed by #4938 but more testing required to confirm.

peterbell10:
The issue as described by jammet should certainly be fixed. IMO the original issues is not specific enough to test properly. So, I say close.

PR #4938, opened 2020-09-26
Fix instant mining of blocks not being recognised, tweak anti-cheat

### LINKED ISSUE #4757: No item drops with eff V pickaxe
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 99d96337da2852b3163108334eba84e1ba806cb7

### Expected behavior
Item drops when I break block with eff V diamond pickaxe.

### Actual behavior
Of the blocks I've tested, it seems like sandstone is affected, I should do more testing. The block is destroyed, but no item is produced to be picked up.

Mine block

KingCol13:
This fixes #4757, and some of my notes are found there.

The anti-cheat system was in need of some TLC and it seemed semi-empirical so I updated it with the calculation taken from the wiki: https://minecraft.gamepedia.com/Breaking#Speed.

Takes into account player tool, enchantments, status effects and world position when working out if a block is instantly mined (and hence we would not receive a finished digging packet).


KingCol13:
#2299, #4562 and maybe #2121 are duplicates.

src/Entities/Player.cpp
```diff
@@ -3133,61 +3133,110 @@ bool cPlayer::IsInsideWater()
...
+				MiningSpeed += (EfficiencyModifier * EfficiencyModifier) + 1;
+			}
 		}
 	}
+	else  // If we can't harvest the block then no bonuses
+	{
+		MiningSpeed = 1;
+	}
 
+	// Hase increases speed by 20% per level
```
peterbell10, line 3158:
```suggestion
	// Haste increases speed by 20% per level
```

src/Entities/Player.h
```diff
@@ -590,11 +590,18 @@ class cPlayer:
...
 	void RemoveClientHandle(void);
 
-	/** Returns the relative block hardness for the block a_Block.
-	The bigger it is the faster the player can break the block.
-	Returns zero if the block is instant breakable.
-	Otherwise it returns the dig speed (float GetDigSpeed(BLOCKTYPE a_Block)) divided by the block hardness (cBlockInfo::GetHardness(BLOCKTYPE a_Block)) divided by 30 if the player can harvest the block and divided by 100 if he can't. */
-	float GetPlayerRelativeBlockHardness(BLOCKTYPE a_Block);
+	/** Returns the progress mined per tick for the block a_Block as a fraction
+	(1 would be completely mined)
+	Returns 0 if it can be instantly mined.
```
peterbell10, line 595: Wouldn't returning 1 make more sense?

peterbell10 (approved):
Tested and it worked well. Couple minor comments.

src/Entities/Player.h
KingCol13: It really would, if I had a reason for that inconsistency I've since forgotten, thanks!

src/ClientHandle.cpp
```diff
@@ -1335,10 +1335,10 @@ void cClientHandle::HandleBlockDigFinished(int a_BlockX, int a_BlockY, int a_Blo
...
 	}
 
-	if (!m_Player->IsGameModeCreative() && !cBlockInfo::IsOneHitDig(a_OldBlock))
+	if (!m_Player->IsGameModeCreative() && !m_Player->CanInstantlyMine(a_OldBlock))
 	{
-		// Fix for very fast tools.
-		m_BreakProgress += m_Player->GetPlayerRelativeBlockHardness(a_OldBlock);
+		m_BreakProgress += m_Player->GetMiningProgressPerTick(a_OldBlock);
+		// Check for very fast tools. Maybe instead of FASTBREAK_PERCENTAGE we should check we are within x multiplied by the progress per tick
 		if (m_BreakProgress < FASTBREAK_PERCENTAGE)
```
KingCol13, line 1342: Any opinions on this? I think checking for x ticks missed feels more natural. Then it is coupled to how fast their mining tool is instead of just an arbitrary leeway

peterbell10: That sounds like a good idea. You'll need to update `FASTBREAK_PERCENTAGE` either way:
https://github.com/cuberite/cuberite/blob/c158569af3f8303c1c2472490cfd2f5a1f599d35/src/Root.cpp#L138-L139

KingCol13: Ah it's off completely, I'm not sure I want to turn it on because some of our block hardness values and CanHarvest etc seem to be wrong and this would cause people to be randomly kicked.

KingCol13:
I think this is ready to merge. @bearbin are you happy with this?

bearbin (approved):

PR #4941, opened 2020-09-27
Adding new monster types to enum and saving/loading for easier future implementation

12xx12:
This is a PR that mostly changes nothing. By changing everything

- Added the new monsters to enum
- Added the new monsters to loading and saving
- Added methods in namespace serializer eMonsterType <-> String
- added basic code to send new monsters to client
- added code that transmits striders/husks as skeletons zombies in older versions

- mobs that have been split have a own enum value now (horse types,...)

This PR was initially motivated to add the serialazation and then one thing to the other

I'm gonna check if there are any other uses of the old lookup function and remove that maybe

**Note: After merging this PR the server will write the names with "minecraft:entityname" to disk. this will result in pre 1.13 clients not being able to load those worlds.**


12xx12:
This PR can be used to get rid of 
- random Zombie horses spawning
- Easy adding of striders, ...

12xx12:
Not adding any monsters to the ```cMonster::NewMonsterFromType``` is intentional as a safeguard someone tries to spawn the new enum values

12xx12:
I am trying to make the diff a bit smaller

src/WorldStorage/NamespaceSerializer.cpp
```diff
@@ -131,6 +131,91 @@ namespace NamespaceSerializer
...
+			case mtVindicator:     return "minecraft:vindicator";
+			case mtWanderingTrader:return "minecraft:wandering_trader";
+			case mtWitch:          return "minecraft:witch";
+			case mtWither:         return "minecraft:wither";
+			case mtWitherSkeleton: return "minecraft:wither_skeleton";
+			case mtWolf:           return "minecraft:wolf";
+			case mtZoglin:         return "minecraft:zoglin";
+			case mtZombie:         return "minecraft:zombie";
+			case mtZombieHorse:    return "minecraft:zombie_horse";
+			case mtZombiePigman:   return "minecraft:zombie_pigman";
```
12xx12, line 209: Mojang changed the name of the zombie pigman in 1.16 to zombified piglin. Which name should we choose for saving?

bearbin: The ZPM will never be forgotten! But change happens, we should follow vanilla and move to ZPLs.

12xx12:
Yeah - i think that's that.
I don't think the diff will get smaller.

12xx12:
Waiting for #4994 to be merged to make the map a constexpr

src/Protocol/Protocol_1_10.cpp
```diff
@@ -1001,6 +1015,49 @@ void cProtocol_1_10_0::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_
...
+		case mtGiant:
+		case mtIronGolem:
+		case mtMooshroom:
+		case mtSilverfish:
+		case mtSkeleton:
+		case mtSnowGolem:
+		case mtStray:
+		case mtSpider:
+		case mtSquid:
+		{
```
bearbin, line 1054: Should be a break here?

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1550,71 +1551,116 @@ void cWSSAnvil::LoadEntityFromNBT(cEntityList & a_Entities, const cParsedNBT & a
...
-		{ "minecraft:zombie",              &cWSSAnvil::LoadZombieFromNBT },
-		{ "PigZombie",                     &cWSSAnvil::LoadPigZombieFromNBT },
-		{ "minecraft:zombie_pigman",       &cWSSAnvil::LoadPigZombieFromNBT },
+		{ "PigZombie", &cWSSAnvil::LoadZombiefiedPiglinFromNBT },
 		{ "ZombieVillager",                &cWSSAnvil::LoadZombieVillagerFromNBT },
+
+		// Old names in the new format
+		{ "minecraft:villager_golem",      &cWSSAnvil::LoadIronGolemFromNBT },
+		{ "minecraft:snowman",             &cWSSAnvil::LoadSnowGolemFromNBT },
+		{ "minecraft:zombie_pigman", &cWSSAnvil::LoadZombiefiedPiglinFromNBT },
```
bearbin, line 1593: Indentation is off here.

bearbin (commented):
This looks good, when rebased on the constexpr change this could be merged in.

12xx12:
In the beginning I tried to move the MonsterTypes.h to the Registries directory. But it's someday tied into the Lua Compilation and failed. 

Is it worth the changes to have all those in one place to change the paths in the Codebase?

tigerw:
still looking...

tigerw (approved):

PR #4943, opened 2020-09-27
add SetBlockMeta warning in documentation

### LINKED ISSUE #4058: Door doesn't close when lever is turned off from a plugin
Client version:   1.12.2
Server OS:        Linux

### Expected behavior

Door closes when lever is turned off from a plugin.

### Actual behavior

Door doesn't close when lever is turned off from a plugin.

1. Place a door with next to it a block with a lever on it.
2. Turn the lever on by right-clicking.
3. From a plugin, call `SetBlockMeta` on the world with the location of the lever, and one of the following metas (for example `cRoot:Get():GetWorld("world"):SetBlockMeta(-1, 7, 9, 4, false, true)`):

Meta (+8 for on) | Description
-- | --
0 | Lever on block bottom points east when off
1 | Lever on block side facing east
2 | Lever on block side facing west
3 | Lever on block side facing south
4 | Lever on block side facing north
5 | Lever on block top points south when off
6 | Lever on block top points east when off
7 | Lever on block bottom points south when off

Update: the door closes when I dig the unlit redstone torch next to the lever, but it doesn't open again when I place a redstone torch next to the lever. When I turn the lever on or off as a player, the door just closes or opens fine, even if there's an unlit redstone torch next to it.

Workaround: also `SetBlockMeta` on the door.

My build looks like the following:

Screenshot

pwnorbitals:
Fix #4058. 
Feedback appreciated

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -542,6 +542,7 @@ return
 				only used as a parameter for the {{OnChunkGenerating|OnChunkGenerating}} and
 				{{OnChunkGenerated|OnChunkGenerated}} hooks and cannot be constructed on its own. Plugins can use this
 				class in both those hooks to manipulate generated chunks.
+				Calls to any setter of this class will not trigger simulator updates (lava, water, redstone).
 			]],
 			Functions =
 			{
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -1142,7 +1142,7 @@ return
 							Type = "number",
 						},
 					},
-					Notes = "Sets the block meta at the specified relative coords. Any call to SetBlockMeta will not generate a simulator update (water, lava, redstone), consider using SetBlock instead.",
+					Notes = "Sets the block meta at the specified relative coords.",
 				},
 				SetBlockType =
 				{
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -1119,7 +1119,7 @@ return
 							Type = "EMCSBiome",
 						},
 					},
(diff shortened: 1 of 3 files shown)


bearbin:
I don't think that 'generate a simulator update' is the most understandable wording for less-knowledgeable plugin writers. I think adding examples for the current simulators here would be useful.

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1142,7 +1142,7 @@ return
 							Type = "number",
 						},
 					},
-					Notes = "Sets the block meta at the specified relative coords",
+					Notes = "Sets the block meta at the specified relative coords. Any call to SetBlockMeta will not generate a simulator update (water, lava, redstone), consider using SetBlock instead.",
```
tigerw, line 1145: This is the API for cChunkDesc right? That's just a container for block data during World generation, no setters will generate any simulator updates.

Server/Plugins/APIDump/Classes/BlockArea.lua
```diff
@@ -1621,7 +1621,7 @@ return
 						Type = "number",
 					},
 				},
-				Notes = "Sets the block meta at the specified absolute coords",
+				Notes = "Sets the block meta at the specified absolute coords. Any call to SetBlockMeta will not generate a simulator update (water, lava, redstone), consider using SetBlock instead.",
```
tigerw, line 1624: Similarly for cBlockArea, it doesn't do any simulator updates at all (for speed when setting a large area).

pwnorbitals:
Is it documented somewhere that cBlockArea and cChunkDesc never generate simulator updates ?

tigerw:
Implicitly for cChunkDesc: https://api.cuberite.org/cChunkDesc.html
Not really for cBlockArea: https://api.cuberite.org/cBlockArea.html

tigerw (approved):

PR #4944, opened 2020-09-28
HandleEnchantItem: use a lambda

tigerw:


tigerw:
I'm not super convinced
1. noisy syntax to create the lambda and capture all the variables
2. the 5 lines in question appear someone unrelated to their invocation, forcing you to first realise it's only a definition before you understand what it's trying to do
3. it's a goto with extra steps

peterbell10 (approved):

PR #4946, opened 2020-09-28
Adding Silverfish Spawning Blocks

### LINKED ISSUE #4908: Block spawning silverfishes

### Expected behavior
When breaking those certain blocks a silverfish should be spawned and close blocks should spawn their silverfishes

### Actual behavior
Nothing happens

- give yourself such a block
- place
- break it

12xx12:
added breaking, spawning, animation

Fixes #4908

Todo:
- [x] make attacked silverfishes break nearby blocks and spawn silverfishes


src/Blocks/BlockInfested.h
```diff
@@ -0,0 +1,73 @@
...
+
+
+	virtual void OnPlayerBrokeBlock(
+			cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface,
+			cPlayer & a_Player,
+			Vector3i a_BlockPos,
+			BLOCKTYPE a_OldBlockType, NIBBLETYPE a_OldBlockMeta
+	) const override
+	{
+		SpawnSilverfish(a_ChunkInterface, a_WorldInterface, a_BlockPos);
```
peterbell10, line 52: This should only spawn for creative players.

src/Blocks/BlockInfested.h
```diff
@@ -0,0 +1,73 @@
...
+{
+	using Super = cBlockHandler;
+
+public:
+
+	using Super::Super;
+
+private:
+
+	virtual cItems ConvertToPickups(NIBBLETYPE a_BlockMeta, const cEntity * a_Digger, const cItem * a_Tool) const override
```
peterbell10, line 17: Is it actually possible to harvest an infested block?

peterbell10 (changes requested):

src/Blocks/BlockInfested.h
12xx12: No, not at all.

src/Blocks/BlockInfested.h
12xx12: you are right.

12xx12: Is there function that is call that the player is not broken by a player but by an explosion?

KingCol13: In explodinator.cpp but I think it uses setblock

peterbell10: `OnBroken` I think.

12xx12: i'm not trying to break some blocks i want to know if the block was broken by an explosion :D

12xx12: hm but then we can't prevent spawning them, when the player is in creative

peterbell10: I tested and creative mode doesn't make a difference for explosions, the silverfish were always spawned.

12xx12: So the Blocks break under three conditions:

- being mined by a player in survival
- being destroyed by an explosion
- a nearby silverfish is attacked.


As far as I understood OnBreak is triggered before OnPlayerBreak and if a player breaks the block both get triggered.

So if we do the replacement in OnBreak and OnPlayerBreak we spawn two entities.

If we do the check only in OnBreak we can't prevent them from spawning if the player is in creative mode. If we the check only in OnBreak we can't spawn them as a result from an explosion.

peterbell10: I didn't realise both were called when a player breaks the block. That's not very helpful. 
It might require some reworking of the block handler calls to make this work.

12xx12: Yeah - i checked when the on break function is called... it's a bit of a mess

12xx12:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa what am i doing wrong`?

src/Blocks/BlockInfested.h
```diff
@@ -0,0 +1,81 @@
...
+	{
+		return 11;
+	}
+
+
+
+
+	void SpawnSilverfish(cWorldInterface & a_WorldInterface, Vector3i a_BlockPos) const
+	{
+		// TODO: only display animation if the difficulty allows mob spawns
```
bearbin, line 74: Hm, more like when Cuberite has the concept of difficulty at all!

The comment can stay though, it's fine.

src/Blocks/BlockInfested.h
```diff
@@ -0,0 +1,81 @@
...
+		// Spawn Silverfish
+		a_WorldInterface.SpawnMob(a_BlockPos.x - 0.5f, a_BlockPos.y - 0.5f, a_BlockPos.z - 0.5f, mtSilverfish, false);
+		// Play particle
+		a_WorldInterface.GetBroadcastManager().BroadcastParticleEffect("explode", a_BlockPos, {0.5f, 0.5f, 0.5f}, 0.1f, 50);
```
bearbin, line 78: The 0.5 offset should really be a seperate constant somewhere, as it's shared between spawning the mob, and the particle effect.

bearbin (commented):
Looks good, save for one very minor change.

src/Blocks/BlockInfested.h
mathiascode: There is always the difficulty code in the Core plugin, which should've been ripped out and replaced with a proper API ages ago. :P

bearbin:
You're running into the issue just recently introduced by @tigerw and @peterbell10 -> the BlockInfested class needs to be marked final.

12xx12:
Thank you

bearbin (approved):

tigerw:
allow me to apply some style

src/Mobs/Silverfish.cpp
```diff
@@ -0,0 +1,66 @@
...
+	// If attacker is player or splash potion
+	bool ShouldSpawn = false;
+	if ((a_TDI.DamageType == dtPoison) || (a_TDI.DamageType == dtPotionOfHarming))
+	{
+		ShouldSpawn = true;
+	}
+	else
+	{
+		ShouldSpawn |= a_TDI.Attacker->IsPlayer();
+	}
```
peterbell10, line 39: NIT: This is a bit verbose.
```suggestion
	bool ShouldSpawn = (
		(a_TDI.DamageType == dtPoison) || (a_TDI.DamageType == dtPotionOfHarming) ||
		a_TDI.Attacker->IsPlayer()
	);
```

src/Mobs/Silverfish.cpp
```diff
@@ -0,0 +1,66 @@
...
+	if (!ShouldSpawn)
+	{
+		return SuperResult;
+	}
+	auto Blocks = sSetBlockVector();
+	for (int X = static_cast<int>(m_LastPosition.x - 10); X <= static_cast<int>(m_LastPosition.x + 10); X++)
+	{
+		for (int Y = static_cast<int>(m_LastPosition.y - 5); Y <= static_cast<int>(m_LastPosition.y + 5); Y++)
+		{
+			for (int Z = static_cast<int>(m_LastPosition.z - 10); Z <= static_cast<int>(m_LastPosition.z + 10); Z++)
```
peterbell10, line 50: Prefer `m_Position` over `m_LastPosition` in game logic. (`m_LastPosition` is only used for protocol reasons).

peterbell10 (commented):
Looks good, just a couple minor comments.

12xx12:
@tigerw Do you still want to change something

peterbell10 (approved):

PR #4949, opened 2020-09-28
SQLiteCpp 3.1.1

tigerw:


tigerw:
The architecture I was interested in to see if the SQLite update fixed the PowerPC crash, doesn't work because of
```
$ sudo snap install core
error: cannot perform the following tasks:
- Setup snap "core" (9991) security profiles (cannot setup udev for snap "core": cannot reload udev rules: exit status 2
udev output:
)
- Setup snap "core" (9991) security profiles (cannot reload udev rules: exit status 2
udev output:
)
The command "sudo snap install core" failed and exited with 1 during .
```
how charming.

tigerw:
merge?

peterbell10 (approved):

PR #4950, opened 2020-09-28
It is time to remove the statistics upgrade map

tigerw:


peterbell10:
Surely 1 month is hardly enough of a transition period?

tigerw:
With the breathless flavour of CI we have where only master is really supported, you may be surprised! Though without telemetry no one can really tell. Do you have any suggestions?

tigerw:
ping

12xx12:
My suggestion would be to wait for the next newsletter and send out the message with the info that satitstics data and old achievements with a reasonable date for removal

peterbell10:
Looking again, I see that GitHub rounding down when it said 1 month. It's more like 2 at this point. That's more reasonable but I would prefer, say,  6 months or so. It really doesn't hurt us much to keep it around for a while longer.

bearbin:
Yeah, this mapping doesn't really hurt us ATM. This can go in the deep freeze for another 6 months, then it'll be fine.

tigerw:
Close enough?

koutsie:
+1

koutsie (approved):
ISNI?

12xx12:
I'd rather note this in a "news" article with a date and then do it at that date. This still doesn't hurt anyone

12xx12:
I knew i forgot something in the newsletter

bearbin:
Oh no, another argument for having a newsletter planning issue. Either way this is probably fine to go in now once the conflict is fixed.

tigerw:
😃

tigerw:
I'd like to merge if it passes.

PR #4951, opened 2020-09-28
Adding wolf breading and moving breeding functionality to cMonster

### LINKED ISSUE #4911: Wolf breeding is not possible

### Expected behavior
tamed Wolfes can be fed with any kind of meat to bread and spawn a new wolf

### Actual behavior
No food is accepted when fully healed

- Spawn wolf
- tame it
- try to feed it

12xx12:
I moved the breeding code to cMonster.

I tried adding a class that can be a ancestor for breedables but that failed.
So now any monster can be bread :D

Fixes #4911


12xx12:
Whoops this PR went from 0 to 100 realy fast

12xx12:
I'm not quite shure if we want to keep it like that. It's just an experiment

12xx12:
Breedable was a bad idea... moving the breeding functionality to cMonster

src/Mobs/Wolf.cpp
```diff
@@ -337,6 +372,8 @@ void cWolf::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	{
 		StopMovingToPosition();
 	}
+
+	cMonster::LoveTick();
```
bearbin, line 376: This needs a bit of testing - this way loving takes priority over all other behaviours, is that correct in vanilla? It could be that LoveTick needs to be moved more to the start of the function.

src/Mobs/Wolf.cpp
```diff
@@ -16,7 +17,6 @@ cWolf::cWolf(void) :
 	m_IsTame(false),
 	m_IsBegging(false),
 	m_IsAngry(false),
-	m_OwnerName(""),
```
bearbin, line 19: Why was this removed?

12xx12: my local clang said that this is redundant string initialization

src/Mobs/Wolf.cpp
12xx12: I did some testing...

The behaviour is a bit strage:

The wolfes run into each other and then attack the enemy and then run into each other a create an offspring.

12xx12:
This implementation also derivates from notchian implementation in that way that the wolves chase you if you have their food in your hand. not the respectful distance as ausual but bump into you like every other feedable

src/Mobs/Wolf.cpp
```diff
@@ -231,6 +246,11 @@ void cWolf::OnRightClicked(cPlayer & a_Player)
 				}
 				break;
 			}
+			// multiplication is handheld in cMonster. Just prevents from sitting down.
```
bearbin, line 249: Now a typo replaced with a typo :P

It should be `handled`

src/Mobs/Monster.cpp
```diff
@@ -14,13 +14,15 @@
...
+#include "Items/ItemSpawnEgg.h"
+
 #include "../Chunk.h"
 #include "../FastRandom.h"
 
 #include "PathFinder.h"
 #include "../Entities/LeashKnot.h"
 
-
+const int TPS = 20;
```
bearbin, line 25: This should actually go in one of the global headers, I'm rather surprised we don't have such a constant already, but Monster.cpp isn't really the right place for it.

12xx12: ```Defines.h``` maybe ?

src/Mobs/Monster.cpp
```diff
@@ -1248,6 +1258,195 @@ std::unique_ptr<cMonster> cMonster::NewMonsterFromType(eMonsterType a_MobType)
...
+		}
+	}
+	else
+	{
+		// We have no partner, so we just chase the player if they have our breeding item
+		cItems FollowedItems;
+		GetFollowedItems(FollowedItems);
+		if (FollowedItems.Size() > 0)
+		{
+			m_World->DoWithNearestPlayer(GetPosition(), static_cast<float>(m_SightDistance), [&](cPlayer & a_Player) -> bool
```
NiLSPACE, line 1339: Eventually we probably should create a `DoWithNearestPlayers` which would loop the players in order of the distance between the specified point. Now if the nearest player doesn't have the `FollowedItem` equiped, but the next player does he is ignored.

bearbin (approved):
The build fails, but once you get it to pass, this LGTM.

12xx12:
Some error with a lua state file...

I updated the branch to master let's see what happens

12xx12:
What did I break this time?

bearbin:
This time it's tolua failing. No idea why though

Seadragon91:
Looks like the change from commit `moved tps to Defines.h` causes that problem.

bearbin:
Maybe try not exporting TPS to the lua API. It makes sense to export it, but not if it causes the whole system to break for no reason.

tigerw:
Best to tag TPS as `inline constexpr` so the linker gives it one address only

12xx12:
@tigerw I tried adding your suggestion, But It seems to break the build

tigerw:
goodness me. Wrap it in a
```
#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wmissing-variable-declarations"
#endif
// ...
#ifdef __clang__
#pragma clang diagnostic pop
#endif
```
I can fix arm64 later.

12xx12:
This seems kinda scetchy - but ok

tigerw:
arm64 is on clang6, it needs to be upgraded to clang7 to silence the warning

12xx12:
that makes sense 💡

Issue #4919, opened 2020-09-21
Generate Minecarts with chests instead of plain chests in mineshafts


12xx12:
Cuberite Commit id: 23bca008317e125659d4b97f29cf13da81a7aecd

Issue #4916, opened 2020-09-21
Additional mob spawner options


12xx12:
Client version: 1.12.1
Server OS: Windows
Cuberite Commit id: 23bca008317e125659d4b97f29cf13da81a7aecd

In 1.4.2. the notchian minecraft added the option to customize the mob spawners further to spawn equipped mobs...

https://minecraft.gamepedia.com/Spawner#Custom_spawners

PR #4953, opened 2020-09-30
Adding Cloning Infrastructure for entities

12xx12:
This PR adds a clone infrastructure like the one for block entities.

This is needed for some things:
- Entities in ```cBlockArea```
- Custom Mobs in mob spawners #4916 
- #4919 
- ...

Things that are missing
- [ ] Mobs
- [ ] handle attached entities (that need some tinkering). at the moment only the pointers are copied numbly)
- [ ] testing

### DIFF
--- a/src/Entities/ArrowEntity.cpp
+++ b/src/Entities/ArrowEntity.cpp
@@ -55,6 +55,22 @@ bool cArrowEntity::CanPickup(const cPlayer & a_Player) const
 
 
 
+void cArrowEntity::CopyFrom(const cEntity &a_Src)
+{
+	Super::CopyFrom(a_Src);
+	auto & Arrow = static_cast<const cArrowEntity &>(a_Src);
+	m_PickupState = Arrow.m_PickupState;
+	m_DamageCoeff = Arrow.m_DamageCoeff;
+	m_IsCritical = Arrow.m_IsCritical;
+	m_Timer = Arrow.m_Timer;
+	m_bIsCollected = Arrow.m_bIsCollected;
+	m_HitBlockPos = Arrow.m_HitBlockPos;
+}
+
+
+
+
+
 void cArrowEntity::OnHitSolidBlock(Vector3d a_HitPos, eBlockFace a_HitFace)
 {
 	// Save the direction we're going in before Super resets it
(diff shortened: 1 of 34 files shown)


tigerw:
It's going to be so easy to forget to add a field. Is cloning really necessary?

12xx12:
That's right. We could also use the copy constructor.

My idea was that if the block entities use the same mechanic there is a reason for this.

12xx12:
I think the idea with the block entities have the clone function is that you don't have to know what type of entity you get. It always works reguardless of the kind of entity you get.

12xx12:
I have an idea!

The abstract classes do the clone () call and the non abstract classes use a copy contractor (at best the default) so that less stuff can be forgotten

12xx12:
Or do you want to discuss this in the forum?

tigerw:
I'm just thinking if it would be better (for the two issues you linked) to store the data needed to create the entities, instead of a full-blown copy of the entity itself. I'm hesitant for you to do all this work when the functionality only gets used in say, two places.

tigerw:
(Really the same argument goes for block entities: the fact that you can create a tile entity in a cBlockArea means it's m_World must be a pointer and may be null. You can kinda squint and accept the existing CopyFrom because block entities tend to be really simple, but in my opinion cBlockArea should store the basic stuff needed to recreate a block entity, instead of copying around an actual instance of one.)

12xx12:
I'm just thinking about how you would implement this.

You still would need to go through any enheritance and find out what kind of entity you have. Then you would need some kind of structure to store that data in...

For basic creation you don't need all the data. BUT some entitites recieve some additional information during their lieftime not handed over by the constructor. Like the ones with inventories, Item frames can be rotated. Stuff like that should be respected.

So yeah for basic copying only some data can be copied.
And when we do this i'd suggest doing it right and copy everything.

And the best structure storing the data needed for recreation is the structure copied from.

This will create some overhead for some types. But most of the time it should do the trick.

tigerw:
NBT could work. We already have functions to load/save. In the use cases you pointed out all that's needed is storage, no manipulation or anything.

tigerw:
The one worry is that MobSpawner needs to send its contents to the client, and that involves multiprotocol awkwardness, but saving to NBT via WorldStorage is locked to one version.

12xx12:
Oh - we could recycle the functions from NBTChunkSerializer and WSSAnvil 👍 

actually i wouldn't sent the custom stuff to the client just show the next entity type to the client. In notchian minecraft it's just an array of NBT describing the entities that may spawn,

12xx12:
Ok so we close this and experiment with NBT in those mentioned applications

PR #4955, opened 2020-10-01
Adding more customize options to mob spawners and improving the way to look for surrounding entities

### LINKED ISSUE #3736: Large number of spawners causes deadlocks
Client version:   1.11.2
Server OS:        Linux
Commit id:        226272fc63297b4477ae43f217a83d7f08701b45

### Expected behavior
Placing a large number of mob spawners in a world shouldn't cause the server to lag and deadlock.

### Actual behavior
Every time the chunk with spawners is loaded, the server deadlocks.

With WorldEdit enabled, type
`//sphere mobspawner 25`

12xx12:
Nearly every option is now present. Only the one where you can provide a list of entities to spawn from the spawner.

https://minecraft.gamepedia.com/Spawner#Block_entity

Fixes #3736


tigerw:
Thanks, looks good! This raises some interesting points:

I personally think instead of get/set, directly exposing the property makes for a more "fluent" API (whatever you take that to mean). Much like C# properties. A manual binding can take care of the SpawnRange limitation on the setter. Should we move to properties for future API additions, instead of setters?

I also believe it's nicer for entities'/block entities' constructors to construct a fully-fledged object, instead of having to call SetXXX on it afterwards. That way WSSAnvil can make a complete MobSpawner from serialised data in one go; the tile entity factory method (CreateBlockEntity or something?) can handle providing default values when during block placement.

12xx12:
> I personally think instead of get/set, directly exposing the property makes for a more "fluent" API (whatever you take that to mean). Much like C# properties. A manual binding can take care of the SpawnRange limitation on the setter. Should we move to properties for future API additions, instead of setters?

I personaly don't care as there are only short values to be set. Except the Custom mobs that would require a proper setter.

> I also believe it's nicer for entities'/block entities' constructors to construct a fully-fledged object, instead of having to call SetXXX on it afterwards. That way WSSAnvil can make a complete MobSpawner from serialised data in one go; the tile entity factory method (CreateBlockEntity or something?) can handle providing default values when during block placement.

I actually added the default values to the .h file of the mob spawner so we could just check on saving if those are not the default values and drop those on saving.

yeah that buch of methods create some stackframes - in respect for performance we could do this in one go.

12xx12:
I think this is worth a forum threat to discuss with everyone

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1401,13 +1449,79 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			SetSpawnRange =
+			{
+				Params =
+				{
+					{
+						Name = "SpawnRange",
+						Type = "number",
+					},
+				},
+				Notes = "Sets the diameter of the square the spawner will try to spawn entities in.",
```
peterbell10, line 1472: Radius, not diameter. You should also say it'll be clipped to 20.

src/BlockEntities/MobSpawnerEntity.cpp
```diff
@@ -260,7 +269,7 @@ int cMobSpawnerEntity::GetNearbyMonsterNum(eMonsterType a_EntityType)
 				return;
 			}
 
-			if ((Diff(m_SpawnerPos.x, a_Entity->GetPosX()) <= 8.0) && (Diff(m_SpawnerPos.y, a_Entity->GetPosY()) <= 4.0) && (Diff(m_SpawnerPos.z, a_Entity->GetPosZ()) <= 8.0))
+			if ((Diff(m_SpawnerPos.x, a_Entity->GetPosX()) <= (2 * m_SpawnRange + 1)) && (Diff(m_SpawnerPos.y, a_Entity->GetPosY()) <= 4.0) && (Diff(m_SpawnerPos.z, a_Entity->GetPosZ()) <= (2 * m_SpawnRange + 1)))
```
peterbell10, line 272: The wiki says
>Overrides the maximum number of nearby (within a box of spawnrange*2+1 × spawnrange*2+1 × 8 centered around the spawner block) entities whose IDs match this spawner's entity ID.

The radius of that box should be half the width, so `spawnrange + 0.5`.

peterbell10 (dismissed):
Setters/Getters are fine. These are only accessed in world storage and lua, and tolua needs functions to bind anyway.

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1401,13 +1449,79 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			SetMinSpawnDelay =
+			{
+				Params =
+				{
+					{
+						Name = "MinSpawnDelay",
+						Type = "number",
+					},
+				},
+				Notes = "Sets the minimum amount of ticks the spawner will way before spawning new entities.",
```
bearbin, line 1483:
```suggestion
				Notes = "Sets the minimum amount of ticks the spawner will wait before spawning new entities.",
```

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1401,13 +1449,79 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			SetMaxSpawnDelay =
+			{
+				Params =
+				{
+					{
+						Name = "MaxSpawnDelay",
+						Type = "number",
+					},
+				},
+				Notes = "Sets the maximum amount of ticks the spawner will way before spawning new entities.",
```
bearbin, line 1494:
```suggestion
				Notes = "Sets the maximum amount of ticks the spawner will wait before spawning new entities.",
```

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1401,13 +1449,79 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			SetRequiredPlayerRange =
+			{
+				Params =
+				{
+					{
+						Name = "RequiredPlayerRange",
+						Type = "number",
+					},
+				},
+				Notes = "Sets the maximum amount of blocks the spawner looks for a player during activation check.",
```
bearbin, line 1516: This comment needs rewording,
```suggestion
				Notes = "Sets the maximum Euclidian distance from a player where the spawner can be activated.",
```

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1401,13 +1449,79 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			},
 			SpawnEntity =
 			{
-				Notes = "Spawns the entity. This function automaticly change the spawn delay!",
+				Notes = "Spawns the entity. This function automatically change the spawn delay int the gives limits!",
```
bearbin, line 1520:
```suggestion
				Notes = "Spawns the entity. NOTE: This function resets the delay before the next automatic activation of the spawner.",
```

src/BlockEntities/MobSpawnerEntity.h
```diff
@@ -33,46 +33,68 @@ class cMobSpawnerEntity :
...
 
-	/** Upate the active flag from the mob spawner. This function will called every 5 seconds from the Tick() function. */
+	/** Update the active flag from the mob spawner. This function will called every 5 seconds from the Tick() function. */
 	void UpdateActiveState(void);
 
 	/** Sets the spawn delay to a new random value. */
 	void ResetTimer(void);
 
-	/** Spawns the entity. This function automaticly change the spawn delay! */
+	/** Spawns the entity. This function automatically change the spawn delay! */
```
bearbin, line 45: This comment should correspond with my previous suggestion above.

src/BlockEntities/MobSpawnerEntity.cpp
bearbin: Really, this whole function should be reworked to use cWorld::ForEachEntityInBox

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1375,6 +1391,38 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
 				},
 				Notes = "Returns the spawn delay. This is the tick delay that is needed to spawn new monsters.",
 			},
+			GetMinSpawnDelay =
+			{
+				Returns =
+				{
+					Type = "number",
+				},
+				Notes = "Returns the minimum number of ticks the spawner waits until spawning new entities.",
```
bearbin, line 1400:
```suggestion
				Notes = "Returns the minimum number of ticks the spawner waits until spawning new entities automatically.",
```

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1375,6 +1391,38 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			},
+			GetMaxSpawnDelay =
+			{
+				Returns =
+				{
+					Type = "number",
+				},
+				Notes = "Returns the maximum number of ticks the spawner waits until spawning new entities.",
```
bearbin, line 1408:
```suggestion
				Notes = "Returns the maximum number of ticks the spawner waits until spawning new entities automatically.",
```

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1375,6 +1391,38 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			},
+			GetMaxNearbyEntities =
+			{
+				Returns =
+				{
+					Type = "number",
+				},
+				Notes = "Returns the number of entities of the same type that stop the spawner from spawning more entities.",
```
bearbin, line 1416:
```suggestion
				Notes = "Returns the maximum number of entities of the same type that can be present before the spawner cannot spawn more entities.",
```

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1375,6 +1391,38 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+			},
+			GetRequiredPlayerRange =
+			{
+				Returns =
+				{
+					Type = "number"
+				},
+				Notes = "Returns the maximum activation range for the mob spawner",
```
bearbin, line 1424:
```suggestion
				Notes = "Returns the maximum Euclidian distance from a player where the spawner can be activated.",
```

bearbin (changes requested):

Server/Plugins/APIDump/Classes/BlockEntities.lua
peterbell10: It's not a euclidean distance though, that would describe a sphere not a box.

bearbin: Well that's what the implementation is doing, so if it should be Manhattan distance, then the implementation needs to be changed.

peterbell10: Ah, you're right our implementation doesn't match the wiki. It should be a cube with side-length `2*RequiredPlayerRange`.

src/BlockEntities/MobSpawnerEntity.cpp
12xx12: This might resolve the lag that is introduced when spawning a ton of spanners

Server/Plugins/APIDump/Classes/BlockEntities.lua
12xx12: The spawner actually looks for players in a sphere and spawns entities in a box

src/BlockEntities/MobSpawnerEntity.cpp
```diff
@@ -191,39 +199,22 @@ void cMobSpawnerEntity::SpawnEntity(void)
...
-			}
-
-			if ((m_SpawnerPos - a_Entity->GetPosition()).Length() <= 16)
-			{
-				m_NumPlayers++;
-			}
+			NumPlayers++;
 		}
+		return false;
+	};
```
12xx12, line 215: Havin to return false on success is realy weird

bearbin: It's the C convention, adapted to a boolean return value  - 0 return values mean success, and 0 is false.

12xx12: I was confused because I only used function which only handle one entity and there it didn't matter if you returned true or false. But yeah makes sense

src/BlockEntities/MobSpawnerEntity.h
```diff
@@ -33,46 +33,73 @@ class cMobSpawnerEntity :
...
+	/** Maximum amount of the same entity type in proximity. */
+	short m_MaxNearbyEntities = 6;
+
+	/** Maximum distance to player for activation */
+	short m_RequiredPlayerRange = 16;
+
+	/** The bounding boxes used to look for entities around the spawner.
+	Updated on updating m_RequiredPlayerRange or m_SpawnRange. */
+	cBoundingBox m_PlayerBoundingBox;
+	cBoundingBox m_EntityBoundingBox;
```
peterbell10, line 101: You may be better off just creating the bounding box object on-demand. It's not expensive to construct and saves you from having to put the constructor calculation in multiple places.

src/BlockEntities/MobSpawnerEntity.cpp
```diff
@@ -17,7 +17,9 @@ cMobSpawnerEntity::cMobSpawnerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMe
 	Super(a_BlockType, a_BlockMeta, a_Pos, a_World),
 	m_Entity(mtPig),
 	m_SpawnDelay(100),
-	m_IsActive(false)
+	m_IsActive(false),
+	m_EntityBoundingBox({(float) a_Pos.x, (float) a_Pos.y - 4,                     (float) a_Pos.z}, m_SpawnRange, 8),
+	m_PlayerBoundingBox({(float) a_Pos.x, (float) a_Pos.y - m_RequiredPlayerRange, (float) a_Pos.z}, m_RequiredPlayerRange, m_RequiredPlayerRange * 2)
```
peterbell10, line 22: Don't use c-style casts.

src/BlockEntities/MobSpawnerEntity.cpp
12xx12: It does - yeah 🤯

Server/Plugins/APIDump/Classes/BlockEntities.lua
```diff
@@ -1365,6 +1365,22 @@ World:ForEachChestInChunk(Player:GetChunkX(), Player:GetChunkZ(),
...
+				},
+				Notes = "Returns the number of entities the spawner will try to spawn on each activation.",
+			},
+			GetSpawnRange =
+			{
+				Returns =
+				{
+					Type = "number",
+				},
+				Notes = "Returns the shortest radius of the square the spawner tries to spawn entities in.",
```
bearbin, line 1382: A square cannot have a 'shortest radius'???

Server/Plugins/APIDump/Classes/BlockEntities.lua
bearbin: This hasn't been resolved?

src/BlockEntities/MobSpawnerEntity.cpp
```diff
@@ -136,18 +142,18 @@ void cMobSpawnerEntity::SpawnEntity(void)
...
 					break;
 				}
 
 				auto SpawnRelPos(GetRelPos());
 				SpawnRelPos += Vector3i(
-					static_cast<int>((Random.RandReal<double>() - Random.RandReal<double>()) * 4.0),
+					static_cast<int>((Random.RandReal<double>() - Random.RandReal<double>()) * (float) m_SpawnRange),
 					Random.RandInt(-1, 1),
-					static_cast<int>((Random.RandReal<double>() - Random.RandReal<double>()) * 4.0)
+					static_cast<int>((Random.RandReal<double>() - Random.RandReal<double>()) * (float) m_SpawnRange)
```
bearbin, line 156: These two still have c-style casts, and m_SpawnRange should really be casted to double here.

bearbin (changes requested):

Server/Plugins/APIDump/Classes/BlockEntities.lua
bearbin: Yep, according to the wiki our current behaviour of looking for players in a sphere is correct.

Server/Plugins/APIDump/Classes/BlockEntities.lua
12xx12: It doesn't have a normal radius eiter...

src/BlockEntities/MobSpawnerEntity.cpp
12xx12: whats the c++ way to cast in that case`?

bearbin: ```static_cast<double>(m_SpawnRange)```

12xx12: 🤦 that obvious - my bad

Server/Plugins/APIDump/Classes/BlockEntities.lua
12xx12: "half the lenght of the square"

bearbin: That could work.

bearbin (approved):

peterbell10:
@12xx12 can you rebase this on master. It fails with Werror on clang.

PR #4956, opened 2020-10-01
Move ISSUE_TEMPLATE

12xx12:
I didn't see the ISSUE_TEMPLATE on my mobile so i contacted the github support.

They said that the app looks for a .md file in the folder ISSUE_TEMPLATE so that should do the trick.

I sadly can't test this on my own branch

we may also setup multiple issue templates for differtent causes if we want to. But i don't see a reason why


tigerw:
Ah, there's a GitHub app? Interesting! It appears to look in a different place compared to the web interface.

Looks good to me

12xx12:
Actully the web interface uses the same directory - maybe this is an old version that only used one issue template file

bearbin (approved):
This is a github bug, not our bug...

Anyway, this looks fine.

PR #4957, opened 2020-10-01
Add back issue template for web interface

bearbin:
Well, it may or may not work for the app, but it certainly doesn't work on the web interface.


bearbin:
Ref: #4956

bearbin:
Closing, in favour of revert. The version on master doesn't work on the app either.

PR #4961, opened 2020-10-02
Remove SetProperty(...cPlayer)

tigerw:
Enchantment table, anvil windows are already opened one per-player.


bearbin (approved):
Makes sense, if windows should show different things to different players, they should be seperate windows.

PR #4963, opened 2020-10-03
Fix cmake not adding Werror on clang, and _lots_ of warnings

peterbell10:
It turns out a simple typo in @tigerw's cmake PR (`CMAKE_CXX_COMPILE_ID` instead of `CMAKE_CXX_COMPILE`__`R`__`_ID`) has left us without -Werror enabled on the clang CI for several months.

This would have saved us from #4924 for example.


tigerw:
Let me go through it and see if there are any opportunities for simplification

src/Blocks/BlockCrops.h
```diff
@@ -22,9 +27,9 @@ class cBlockCropsHandler:
 private:
 
 	/** Calculate the number of seeds to drop when the crop is broken. */
-	static char CalculateSeedCount(char a_Min, char a_BaseRolls, char a_FortuneLevel)
+	static char CalculateSeedCount(char a_Min, char a_BaseRolls, unsigned char a_FortuneLevel)
 	{
-		std::binomial_distribution Binomial(a_BaseRolls + a_FortuneLevel, 0.57);
+		std::binomial_distribution<char> Binomial(a_BaseRolls + a_FortuneLevel, 0.57);
```
tigerw, line 32: apparently standard doesn't allow this ;P

src/Blocks/BlockAnvil.h
```diff
@@ -10,6 +10,11 @@
 
 
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnon-virtual-dtor"
+#endif
+
```
tigerw, line 17: I'm going to try tagging all the handlers final, then give the base a =default protected destructor. Hopefully silences it

src/Blocks/BlockIce.h
```diff
@@ -59,7 +59,7 @@ class cBlockIceHandler :
 			}
 		};
 
-		for (const auto Offset : Adjacents)
+		for (const auto & Offset : Adjacents)
```
tigerw, line 62: We're taking Vector3s by value though

peterbell10: This isn't a function call, so the same arguments don't apply. I would be fine with by-value in this situation but it's not worth disabling an otherwise useful warning just to save one character.

tigerw: clang10 even adds a limit to how big the class is before it warns, clearly error here is bogus
codegen seems to be the same either way though, thankfully

peterbell10: Yeah, I'm using clang 10 locally which is why I didn't pick it up in the first commit.

tigerw:
thanks for going through all of these by the way!

We should merge #4961 first for SetProperty, and upgrade ARM64 Travis to Focal Fossa for Clang 7. That should simplify this PR a bit.

peterbell10:
Rebased on #4961. I'd like to get this in quickly, or it's only going to get worse.

tigerw:
OK, tell me if you're done and I can try the blockhandler stuff

peterbell10:
I'm done I think. CI is passing.

src/Blocks/BlockCrops.h
```diff
@@ -29,8 +24,8 @@ class cBlockCropsHandler:
 	/** Calculate the number of seeds to drop when the crop is broken. */
 	static char CalculateSeedCount(char a_Min, char a_BaseRolls, unsigned char a_FortuneLevel)
 	{
-		std::binomial_distribution<> Binomial(a_BaseRolls + a_FortuneLevel, 0.57);
-		return static_cast<char>(a_Min + Binomial(GetRandomProvider().Engine()));
+		std::binomial_distribution Binomial(a_BaseRolls + a_FortuneLevel, 0.57);
+		return a_Min + static_cast<char>(Binomial(GetRandomProvider().Engine()));
```
peterbell10: You broke CI.

src/Blocks/BlockHandler.h
```diff
@@ -241,5 +241,7 @@ class cBlockHandler
 
 protected:
 
-	BLOCKTYPE m_BlockType;
+	~cBlockHandler() = default;
+
+	const BLOCKTYPE m_BlockType;
```
peterbell10, line 246: No const in member variables.

src/Blocks/BlockTorch.h
```diff
@@ -8,15 +8,20 @@
 
 
 
-class cBlockTorchHandler:
-	public cClearMetaOnDrop<cMetaRotator<cBlockHandler, 0x7, 0x4, 0x1, 0x3, 0x2>>
+
+class cBlockStaveHandler :
```
peterbell10, line 12: What? How about `TorchBase` or something.

src/Blocks/BlockTorch.h
```diff
@@ -214,3 +219,41 @@ class cBlockTorchHandler:
...
+public:
+
+	using Super::Super;
+};
+
+
+
+
+
+class cBlockRedstoneTorchHandler final :
```
peterbell10, line 237: Why are you moving this around?

src/Blocks/CMakeLists.txt
```diff
@@ -30,8 +31,8 @@ target_sources(
 	BlockDoor.h
 	BlockDropSpenser.h
 	BlockEnchantingTable.h
-	BlockEnderchest.h
 	BlockEndPortalFrame.h
+	BlockEnderchest.h
```
peterbell10, line 35: What alphabet are you using? Are you sorting uppercase before lower case?

bearbin:
:( clang-tidy failed

I'd also prefer `SimpleBlockHandler` or `DefaultBlockHandler` as opposed to `PlainOldBlockHandler`, as I think that would be easier to understand particularly for ESL people.

Making the handlers final does seem like a bit of a heavy-handed solution, but if it fixes the warning...

tigerw:
Some pre-emptive explanations:

1. I looked up when the clang flags were added and made a compiler version check.
2. Turned off global-constructors, a server doesn't need startup often so it's much preferable to do a one-time init instead of generating magic statics guard code, brought consistent with exit-time-destructors turned off too
3. If the user is fast they can trigger a signal handler before main gets to creating cRoot, so we need the state variables static
4. I was getting an assertion failure in the chunk serializer ASSERT(BitIndex == 0) so worked on it a bit more
5. that's about it

src/Blocks/CMakeLists.txt
tigerw: ask Notepad++...

src/Blocks/BlockTorch.h
tigerw: Consistency with e.g. BlockFluid since it's a super simple inherit of a parent, would you prefer another way?

src/Blocks/BlockHandler.h
tigerw: Why not?

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2272,17 +2275,6 @@ bool cProtocol_1_8_0::HandlePacket(cByteBuffer & a_ByteBuffer, UInt32 a_PacketTy
 			}
 			break;
 		}
-		default:
```
tigerw, line 2275: m_State is currently user-controlled from cMultiVersionProtocol, would need to add some checks

peterbell10:
Right, your having a laugh here tiger. You can't just make sweeping changes to __someone else__'s PR without asking. If you disagree with something, tell me why and we can have a discussion. Don't just commit whatever you want. I've disabled edits from maintainers on this PR.

>Turned off global-constructors, a server doesn't need startup often so it's much preferable to do a one-time init instead of generating magic statics guard code, brought consistent with exit-time-destructors turned off too

Global constructors are discouraged because of the static initialization order fiasco. Exit time destructors have the exact same problem.

>If the user is fast they can trigger a signal handler before main gets to creating cRoot, so we need the state variables static

Wouldn't it make more sense to just register the signal handler after `cRoot` is initialized so we know it's in a valid state?

src/Protocol/Protocol_1_8.cpp
peterbell10: You shouldn't use enum types to represent values that might be invalid enum values.

src/Blocks/BlockHandler.h
peterbell10: See this explanation of `std::launder`: https://stackoverflow.com/a/39382728 The important point is:

>Because `n` is a `const` variable, the compiler is free to assume that `u.x.n` shall always be 1.

And we can break that compiler assumption by use of placement-new in the same memory location. You might not do this directly, but something like a container might do so internally and your code starts to break in horrible ways.

Why introduce these foot-gun. Especially when, in this case, the class is only used as `constexpr` and so can't mutate anyway.

bearbin (approved):

tigerw:
For collaboration! And I did say I would make changes :)

We're initialising string-maps, which don't access any other statics. There's a clear performance argument here.

What about during restart, when cRoot gets destroyed, and then a signal comes through? cRoot::Get() still won't be valid.

What did you not like about the other changes, why were they reverted? e.g. Redstone, CMake flags, m_StayCount, m_ViewDistance, cSlotArea... Were they not correct?

peterbell10:
>For collaboration! And I did say I would make changes :)

You said you would make one specific change to block handlers.

>Redstone

You reverted my changes in the `cPressurePlateHandler` which avoided all casts and instead replaced it with more casts. You also reverted my changes to fix statement macro warnings, even though we use the `do{ } while(false)` pattern throughout the code.

>CMake flags

If you want to lower the warning level, let's have a discussion first.  And I don't see what's so bad about `try_add_flag`. It made for cleaner code IMO and you just deleted it.

>m_StayCount

You snuck this change in without a PR in 167247328a663cba69d752cf1042028becb6266c and was causing problems. So I reverted it. Signed types are easier to work with, and I don't see what we gain from making it unsigned. It's pretty unlikely we'd get to 2 billion stays and wrap-around behaviour is definitely not desirable here.

>cSlotArea

You just rewrote my change for no discernable reason. Why should the `OptionLevels` array be `unsigned` when it's used as a `short`?

tigerw:
Let me try to reply to these points. So about signedness of StayCount, we need to use the type system to encode what we intend the values to represent. The worry isn't about wrap around or negative values, it's about a logic error causing a mismatch in keeping track of the "reference count", so wrapping isn't an argument for or against. Just as you would agree one doesn't represent the count of elements in, say a list, with a signed type, so shouldn't AlwaysTicked/StayCount.

Likewise for enchantment levels, they're something which cannot be negative, the client doesn't handle negative enchantments well, and indeed EnchantByXPLevels crashes if it's provided a negative level, hence only an unsigned (short) makes sense. The cast to a signed short is more due to the protocol datatype unfortunately bleeding into the internal server via window property storage.

For pressure plates too, we're again counting the size of entities around us. The cast is unfortunately a historical artefact of cItem::m_ItemCount being again signed, but casting isn't something to avoid because it again lets us maintain the semantics of NumberOfEntities being a count, that doesn't make sense being negative.

Everyone agrees the do {} while (false) thing is weird and confusing, we really should avoid introducing more of these constructions.

You mentioned you'd like this merged quickly so the above can be addressed in a subsequent PR. About cRoot startup, if I'm not wrong using cRoot::Get() in the signal handlers will cause a crash if CTRL-C/Break is pressed during pre-startup and during restarts. Could you please confirm?

tigerw:
The warnings are valuable, since it's been approved, anyone against merge first improve later?

bearbin:
Unfortunately this PR broke the android build: https://builds.cuberite.org/job/android/124/console

Somewhat highlighting an omission in our CI system, if android code isn't built at all...

Issue #4964, opened 2020-10-03
When breaking a block: OnBreak and OnPlayerBreak are triggered both


12xx12:
Cuberite Commit id: 23bca00

Ref https://github.com/cuberite/cuberite/pull/4946#discussion_r498993593

PR #4967, opened 2020-10-04
fixing OnBreak and OnPlayerBreak are triggered both

12xx12:
This is a kind of workaround to prevent double calls.

I also updated the comments on the functions

Fixes: #4964


src/ClientHandle.cpp
```diff
@@ -1407,7 +1407,8 @@ void cClientHandle::HandleBlockDigFinished(int a_BlockX, int a_BlockY, int a_Blo
 	}
 	else
 	{
-		World->DigBlock(absPos);
+		// Using the chunk maps DigBlock function to prevent calling BreakBlock and OnPlayerBrokeBlock simultaneously
+		World->GetChunkMap()->DigBlock(absPos);
```
12xx12, line 1411: This is actually the main change. Just skipping the call to World.DigBlock() which calls the OnBreak and then the Chunkmap.DigBlock()

tigerw:
Hm. What do you think about consolidating OnBroken and OnPlayerBrokeBlock? Much like ConvertToPickups, pass a const cEntity * a_Digger to OnBroken.

12xx12:
Hmm.. This need some thinking. Do we have any blocks that require different behaviour when broken by player or explosion?

tigerw:
Yep, and there aren't that many uses of OnBroken either so it may be do-able

src/Blocks/BlockBigFlower.h
```diff
@@ -141,6 +141,20 @@ class cBlockBigFlowerHandler:
...
 
+	virtual void OnPlayerBrokeBlock(
+			cChunkInterface & a_ChunkInterface, cWorldInterface & a_WorldInterface,
+			cPlayer & a_Player,
+			Vector3i a_BlockPos,
+			BLOCKTYPE a_OldBlockType, NIBBLETYPE a_OldBlockMeta
+	) const override
+	{
+		OnBroken(a_ChunkInterface, a_WorldInterface, a_BlockPos, a_OldBlockType, a_OldBlockMeta);
+	}
```
peterbell10, line 152: Instead of making the two one function, you could also move this default into `cBlockHandler`. That way `OnPlayerBroke` always takes precedence over `OnBroken` but you only need to define `OnBroken` to cover both cases.

12xx12: That works too - now we have to decide what is the better way

tigerw: One consideration would be future-proofing for when we need to distinguish between more than just players/other; TNT entities, endermen, enderdragons, creepers can all break blocks too.

12xx12: Zomies also e.g. doors

src/ClientHandle.cpp
12xx12: migt wanna revert this ...

tigerw (approved):

PR #4968, opened 2020-10-04
Fixed current end generator

NiLSPACE:
Fixed the current end generator.

There was also a problem where the `m_MinChunkXYZ` variables were reset to 0. As the end generation needs an update anyway since the end now continues to generate after the initial island I removed those variables instead of finding the problem.

### DIFF
--- a/src/Generating/EndGen.cpp
+++ b/src/Generating/EndGen.cpp
@@ -40,10 +40,6 @@ cEndGen::cEndGen(int a_Seed) :
 	m_FrequencyX(80),
 	m_FrequencyY(80),
 	m_FrequencyZ(80),
-	m_MinChunkX(0),
-	m_MaxChunkX(0),
-	m_MinChunkZ(0),
-	m_MaxChunkZ(0),
 	m_LastChunkCoords(0x7fffffff, 0x7fffffff)  // Use dummy coords that won't ever be used by real chunks
 {
 	m_Perlin.AddOctave(1, 1);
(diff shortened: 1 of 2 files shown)


tigerw (approved):
sure looks better than before!

bearbin (approved):

NiLSPACE:
It looks like it was supposed to look ;) I suspect something went wrong when the shape generators were introduced.

NiLSPACE:
I suddenly had a realisation why the `m_MinMax` variables reset to 0. They were never really filled. `InitializeCompoGen` was called which initializes the variables, but the place where they are truly needed is the shape generator, and `cEndGen` doesn't override `InitializeShapeGen`.

The end generation might not need a 3d perlin noise though. a 2d noise might be enough to create floating islands similarly to my [FloatingIslands plugin](https://github.com/NiLSPACE/FloatingIsland/blob/master/generator.lua#L19-L64) (please ignore the horrible code quality). The maximum height could be limited using the formula `x^3 - 3*x` (where x is distance from spawn) to make sure there is a void from the main island compared to the outer islands.

A 3d noise will probably generate nicer looking terrain though. Another option would be to keep the current code, somehow make it generate terrain indefinately, not only in the main area, and limit the air threshold perhaps using the same formula as before.

tigerw:
Would 2D noise create a sort of conical shape and the base of the island?

NiLSPACE:
Yes, it would. I made a quick PoC to see how it would look like:




A quick view using minutor reveals the entire map:
Image


The biggest problem is that the edges end pretty abrupt:

NiLSPACE:
I personally prefer the shapes of the original generator though.

12xx12:
I like the 2d Noise Version better. I thinks it's closer to the notchian implementation

tigerw:
Yours is better in my opinion but they both look good

NiLSPACE:
For reference, here is a small part of the vanilla end:

image

tigerw:
the central part looks like a fried egg that's broken up at the edges xD

NiLSPACE:
Also a reminder that now that we've found the problem with the original generator that we can generate shapes like these just by changing a few values in the `world.ini`

image

tigerw:
nice effect! make it get sparser, instead of spikier :)

Titaniumtown:
Woah! I'm glad people finally added end generation!

PR #4969, opened 2020-10-05
Warnings improvements

tigerw:
* Turn off global-constructors warning. These are needed to implement cRoot signal handler functionality
* Add Clang flags based on version lookup instead of a compile test. The CMake config process is single threaded and slow enough already
* Reduced GetStackValue verbosity
+ Clarify EnchantmentLevel, StayCount, AlwaysTicked, ViewDistance signedness
+ Give SettingsRepositoryInterface a move constructor to simplify main.cpp code
- Remove do {} while (false) construction in redstone handler

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -111,14 +111,6 @@ function(set_global_flags)
 	endif()
 endfunction()
 
-function (try_add_flag TARGET FLAG)
-	include(CheckCXXCompilerFlag)
-	check_cxx_compiler_flag("${FLAG}" "HAS_FLAG_${FLAG}")
-	if ("${HAS_FLAG_${FLAG}}")
-		target_compile_options(${TARGET} PRIVATE "${FLAG}")
-	endif()
-endfunction()
-
 function(set_exe_flags TARGET)
 	if (MSVC)
 		# TODO: MSVC level 4, warnings as errors
@@ -158,15 +150,27 @@ function(set_exe_flags TARGET)
 			-Wno-switch-enum
 
 			# Weverything with Clang exceptions:
-			-Weverything -Wno-error=disabled-macro-expansion -Wno-weak-vtables
-			-Wno-exit-time-destructors -Wno-string-conversion -Wno-c++98-compat-pedantic
-			-Wno-documentation -Wno-documentation-unknown-command -Wno-reserved-id-macro
-			-Wno-error=unused-command-line-argument
+			-Weverything -Wno-exit-time-destructors -Wno-error=disabled-macro-expansion
+			-Wno-weak-vtables -Wno-string-conversion -Wno-c++98-compat-pedantic -Wno-documentation
+			-Wno-documentation-unknown-command -Wno-reserved-id-macro -Wno-error=unused-command-line-argument
 		)
 
-		# We aren't using C++11
-		try_add_flag(${TARGET} -Wno-return-std-move-in-c++11)
-		# int to float conversions happen a lot, not worth fixing all warnings
-		try_add_flag(${TARGET} -Wno-implicit-int-float-conversion)
+		if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7)
+			target_compile_options(
+				${TARGET} PRIVATE
+
+				# We aren't using C++11:
+				-Wno-return-std-move-in-c++11
+			)
+		endif()
+
+		if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
+			target_compile_options(
+				${TARGET} PRIVATE
+
+				# int to float conversions happen a lot, not worth fixing all warnings:
+				-Wno-implicit-int-float-conversion
+			)
+		endif()
 	endif()
 endfunction()
--- a/src/Chunk.cpp
+++ b/src/Chunk.cpp
@@ -479,8 +479,15 @@ bool cChunk::HasBlockEntityAt(Vector3i a_BlockPos)
 
 void cChunk::Stay(bool a_Stay)
 {
-	m_StayCount += (a_Stay ? 1 : -1);
-	ASSERT(m_StayCount >= 0);
+	if (a_Stay)
+	{
+		m_StayCount++;
+	}
+	else
+	{
+		ASSERT(m_StayCount != 0);
+		m_StayCount--;
+	}
 }
 
 
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -67,7 +67,7 @@ float cClientHandle::FASTBREAK_PERCENTAGE;
 ////////////////////////////////////////////////////////////////////////////////
 // cClientHandle:
 
-cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
+cClientHandle::cClientHandle(const AString & a_IPString, unsigned a_ViewDistance) :
 	m_LastSentDimension(dimNotSet),
 	m_ForgeHandshake(this),
 	m_CurrentViewDistance(a_ViewDistance),
@@ -473,7 +473,7 @@ bool cClientHandle::StreamNextChunk(void)
 	cCSLock Lock(m_CSChunkLists);
 
 	// High priority: Load the chunks that are in the view-direction of the player (with a radius of 3)
-	for (int Range = 0; Range < m_CurrentViewDistance; Range++)
+	for (unsigned Range = 0; Range < m_CurrentViewDistance; Range++)
 	{
 		Vector3d Vector = Position + LookVector * cChunkDef::Width * Range;
 
@@ -513,8 +513,10 @@ bool cClientHandle::StreamNextChunk(void)
 	}
 
 	// Low priority: Add all chunks that are in range. (From the center out to the edge)
-	for (int d = 0; d <= m_CurrentViewDistance; ++d)  // cycle through (square) distance, from nearest to furthest
+	for (unsigned Range = 0; Range <= m_CurrentViewDistance; ++Range)  // cycle through (square) distance, from nearest to furthest
 	{
+		const int d = static_cast<int>(Range);
+
 		// For each distance add chunks in a hollow square centered around current position:
 		cChunkCoordsList CurcleChunks;
 		for (int i = -d; i <= d; ++i)
@@ -569,8 +571,8 @@ void cClientHandle::UnloadOutOfRangeChunks(void)
 		cCSLock Lock(m_CSChunkLists);
 		for (auto itr = m_LoadedChunks.begin(); itr != m_LoadedChunks.end();)
 		{
-			int DiffX = Diff((*itr).m_ChunkX, ChunkPosX);
-			int DiffZ = Diff((*itr).m_ChunkZ, ChunkPosZ);
+			const auto DiffX = Diff((*itr).m_ChunkX, ChunkPosX);
+			const auto DiffZ = Diff((*itr).m_ChunkZ, ChunkPosZ);
 			if ((DiffX > m_CurrentViewDistance) || (DiffZ > m_CurrentViewDistance))
 			{
 				ChunksToRemove.push_back(*itr);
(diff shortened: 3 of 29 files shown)


src/SettingsRepositoryInterface.h
```diff
@@ -11,6 +11,7 @@ class cSettingsRepositoryInterface
 	};
 
 	cSettingsRepositoryInterface() = default;
+	cSettingsRepositoryInterface(cSettingsRepositoryInterface &&) = default;
```
peterbell10, line 14: Base classes with virtual functions generally shouldn't be movable or copyable. That's a recipe for object slicing.

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
```diff
@@ -24,7 +24,8 @@ namespace PressurePlateHandler
 
 			if (Entity.IsPickup())
 			{
-				NumberOfEntities += static_cast<cPickup &>(Entity).GetItem().m_ItemCount;
+				const auto & Pickup = static_cast<cPickup &>(Entity);
+				NumberOfEntities += static_cast<size_t>(Pickup.GetItem().m_ItemCount);
```
peterbell10, line 28: I see no improvement here.

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandler.cpp
```diff
@@ -30,59 +30,56 @@
...
+		case E_BLOCK_TRAPPED_CHEST:         return TrappedChestHandler::Callback;       \
+		case E_BLOCK_TRIPWIRE_HOOK:         return TripwireHookHandler::Callback;       \
+		default:                                                                        \
+		{                                                                               \
+			if (cBlockDoorHandler::IsDoorBlockType(BlockType))                          \
+			{                                                                           \
+				return DoorHandler::Callback;                                           \
+			}                                                                           \
+		}                                                                               \
+	}
```
peterbell10, line 82: You say that "Everyone agrees the `do {} while (false)` thing is weird and confusing" but clearly not. If I thought it was bad, I wouldn't have done it.

It makes macros behave syntactically more like functions which can only be a good thing.

src/UI/SlotArea.cpp
```diff
@@ -1684,18 +1684,18 @@ void cSlotAreaEnchanting::UpdateResult(cPlayer & a_Player)
...
-	const std::array<short, 3> OptionLevels
+	const auto Base = (Random.RandInt(1U, 8U) + (Bookshelves / 2) + Random.RandInt(0U, Bookshelves));
+	const std::array<unsigned, 3> OptionLevels
 	{
-		static_cast<short>(std::max(Base / 3, 1)),
-		static_cast<short>((Base * 2) / 3 + 1),
-		static_cast<short>(std::max(Base, Bookshelves * 2))
+		std::max(Base / 3, 1U),
+		(Base * 2) / 3 + 1,
+		std::max(Base, Bookshelves * 2)
```
peterbell10, line 1698: Why change any of this?

src/WorldStorage/NamespaceSerializer.cpp
```diff
@@ -131,129 +131,125 @@ namespace NamespaceSerializer
...
+		{ "cuberite:achievement.spawnWither",        Statistic::AchSpawnWither },
+		{ "cuberite:achievement.killWither",         Statistic::AchKillWither },
+		{ "cuberite:achievement.fullBeacon",         Statistic::AchFullBeacon },
+		{ "cuberite:achievement.breedCow",           Statistic::AchBreedCow },
+		{ "cuberite:achievement.diamondsToYou",      Statistic::AchDiamondsToYou}
+	};
 
 	Statistic ToCustomStatistic(const std::string_view ID)
 	{
-		static const std::unordered_map<std::string_view, Statistic> CustomStatistics
```
peterbell10, line 140: Checking the guard variable is going to be negligible compared to the actual container lookup time.

src/main.cpp
```diff
@@ -86,7 +86,7 @@ static void NonCtrlHandler(int a_Signal)
 		case SIGTERM:
 		{
 			// Server is shutting down, wait for it...
-			cRoot::Get()->Stop();
```
peterbell10, line 89: The signal handlers are set on line 297. Simply move that to after `cRoot` has been constructed and unset it before destroying `cRoot`.

Having a weird half singleton, half static variable object doesn't make sense to me.

peterbell10 (changes requested):

SetFlags.cmake
```diff
@@ -159,14 +151,27 @@ function(set_exe_flags TARGET)
 
 			# Weverything with Clang exceptions:
 			-Weverything -Wno-error=disabled-macro-expansion -Wno-weak-vtables
-			-Wno-exit-time-destructors -Wno-string-conversion -Wno-c++98-compat-pedantic
-			-Wno-documentation -Wno-documentation-unknown-command -Wno-reserved-id-macro
-			-Wno-error=unused-command-line-argument
+			-Wno-global-constructors -Wno-exit-time-destructors -Wno-string-conversion
+			-Wno-c++98-compat-pedantic -Wno-documentation -Wno-documentation-unknown-command
+			-Wno-reserved-id-macro -Wno-error=unused-command-line-argument
```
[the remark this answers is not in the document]
tigerw: I'm loathe to pollute code with compiler flags and the warning-turned-error turns off language functionality that is clearly beneficial.

src/Chunk.cpp
```diff
@@ -479,8 +479,15 @@ bool cChunk::HasBlockEntityAt(Vector3i a_BlockPos)
...
-	ASSERT(m_StayCount >= 0);
+	if (a_Stay)
+	{
+		m_StayCount++;
+	}
+	else
+	{
+		ASSERT(m_StayCount != 0);
+		m_StayCount--;
+	}
```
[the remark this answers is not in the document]
tigerw: I could replace it with
```
ASSERT(a_Stay || (StayCount > 0));
StayCount = a_Stay ? (StayCount + 1) : (StayCount - 1);
```
but this construction was to match the body of SetAlwaysTicked.

ints won't save us from a malfunctioning plugin as a general rule, and not in this specific case either since CanUnload checks StayCount is exactly zero.

"If it wasn't for the STL" :P
What do you do at your work by the way? Sounds cool

src/IniFile.h
```diff
@@ -77,7 +77,6 @@ class cIniFile:
 	cIniFile(void);
 
 	// tolua_end
-	virtual ~cIniFile() override = default;
```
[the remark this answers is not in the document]
tigerw: how so?

src/Protocol/ChunkDataSerializer.cpp
```diff
@@ -503,26 +503,27 @@ inline void cChunkDataSerializer::WriteSectionDataSeamless(const cChunkData::sCh
...
+		if (Remaining >= 0)
 		{
 			// There were some bits remaining: we've filled the buffer. Flush it:
 			m_Packet.WriteBEUInt64(Buffer);
 
 			// And write the remaining bits, setting the new BitIndex:
-			Buffer = Value >> (64 - BitIndex);
-			BitIndex = a_BitsPerEntry - (64 - BitIndex);
+			Buffer = static_cast<UInt64>(Value >> (a_BitsPerEntry - Remaining));
+			BitIndex = static_cast<unsigned char>(Remaining);
```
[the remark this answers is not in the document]
tigerw: That reminds me, I forgot to ask you why you changed it first xD

src/Protocol/ProtocolRecognizer.cpp
```diff
@@ -270,21 +268,16 @@ std::unique_ptr<cProtocol> cMultiVersionProtocol::TryRecognizeLengthedProtocol(c
...
-					throw TriedToJoinWithUnsupportedProtocolException(
-						fmt::format("Invalid next game state: {}", NextStateValue)
-					);
-				}
-			}
-		}();
+			case 1:  return cProtocol::State::Status;
+			case 2:  return cProtocol::State::Login;
+			case 3:  return cProtocol::State::Game;
+			default: throw TriedToJoinWithUnsupportedProtocolException("Your client isn't supported.\nTry connecting with Minecraft " MCS_CLIENT_VERSIONS);
```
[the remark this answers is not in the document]
tigerw: Because it's user facing! Need to keep up appearances, y'know.

I can put the newlines back

src/Protocol/Protocol_1_8.cpp
```diff
@@ -1726,8 +1726,7 @@ bool cProtocol_1_8_0::CompressPacket(const AString & a_Packet, AString & a_Compr
 		----------------------------------------------
 		*/
 		const UInt32 DataSize = 0;
-		const auto PacketSize = static_cast<UInt32>(
-			cByteBuffer::GetVarIntSize(DataSize) + UncompressedSize);
+		const auto PacketSize = static_cast<UInt32>(cByteBuffer::GetVarIntSize(DataSize) + UncompressedSize);
```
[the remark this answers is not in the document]
tigerw: I can put the ); on a newline if you want, this was just style.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2968,8 +2966,8 @@ void cProtocol_1_8_0::HandleVanillaPluginMessage(cByteBuffer & a_ByteBuffer, con
 	}
 	else if (a_Channel == "MC|Beacon")
 	{
-		HANDLE_READ(a_ByteBuffer, ReadBEInt32, Int32, Effect1);
-		HANDLE_READ(a_ByteBuffer, ReadBEInt32, Int32, Effect2);
+		HANDLE_READ(a_ByteBuffer, ReadBEUInt32, UInt32, Effect1);
+		HANDLE_READ(a_ByteBuffer, ReadBEUInt32, UInt32, Effect2);
```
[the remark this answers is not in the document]
tigerw: The original plugin message docs weren't very specific, but ReadBEInt32 is ReadBEUInt32 but copying the UInt32 into an Int32. Certainly beacon effects are nonnegative.

src/Simulator/IncrementalRedstoneSimulator/PressurePlateHandler.h
tigerw: what do you mean?

src/Simulator/IncrementalRedstoneSimulator/RedstoneHandler.cpp
tigerw: For the uninitiated, do while false reads like dead code which should be removed, and they would be right and everything would compile and work fine. Until they hit a clang-tidy error, of all things, complaining about a semicolon, before realising that it is there not to do anything, but to satisfy a style checker. If we claim to be stewards of maintainable and _obvious_ code then we can't be introducing these C-isms.

On the other hand it's true this is being used already, so consistency dictates we'd do the same here. :/

src/UI/SlotArea.cpp
tigerw: You were the one who changed it first xD

less casts, better statement of intent?

src/WorldStorage/NamespaceSerializer.cpp
tigerw: well it's still not free and doesn't get us anything

src/main.cpp
tigerw: I don't want to mutate the signals stuff after startup since it feels like artificially creating a problem for ourselves and then having to write more code to solve it. If you'd prefer I can try enclosing the statics into their own support class away from cRoot?

src/Root.cpp
```diff
@@ -45,8 +46,9 @@
 
 
 
-extern bool g_RunAsService;
-cRoot * cRoot::s_Root = nullptr;
+decltype(cRoot::s_Root)      cRoot::s_Root;
+decltype(cRoot::s_NextState) cRoot::s_NextState;
+decltype(cRoot::s_StopEvent) cRoot::s_StopEvent;
```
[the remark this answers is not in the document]
tigerw: They are part of the class since they control the state of the global cRoot, and are hence related to it, no? Much like s_Root already is a member.

src/ClientHandle.cpp
```diff
@@ -808,8 +810,8 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 			// The experience to add to get the same fill percent.
 			const auto DeltaForPercent = CurrentFillPercent * (m_Player->XpForLevel(TargetLevel + 1) - m_Player->XpForLevel(TargetLevel));
 
-			// Apply the experience delta:
-			m_Player->DeltaExperience(FloorC(DeltaForLevel + DeltaForPercent));
+			// Apply the experience delta, rounded for greater accuracy:
+			m_Player->DeltaExperience(static_cast<int>(std::lround(DeltaForLevel + DeltaForPercent)));
```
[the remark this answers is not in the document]
tigerw: True, this was mainly so we don't always floor the negative experience deltas and give players less experience than expected.

SetFlags.cmake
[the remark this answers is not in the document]
peterbell10: The warning exists because global constructors are a potential "gotcha". Having to explicitly disable the warning just signifies that the author has thought about the issues and deemed it okay. It also increases visibility of the issue in code-review.

tigerw:
Mergey?

Issue #4888, opened 2020-09-18
Missing features in old versions


12xx12:
### Version 1.0
The End:
- [ ] Overworld fortress
- [ ] Generation of the End
- [ ] Being able to throw Eye of Ender
- [ ] Leaving the end
- [ ] Monster spawn stone bricks

Mobs:
- [ ] Ender Dragon
- [ ] Spawning Villager in Villages (No trading until 1.3)

### Version 1.1
- [ ] Blacksmith buildings in villages now have a chest that contains loot.
- [ ] Non-operator users are auto-kicked in multiplayer servers if they spam in chat too much.
       Operators are not notified of the kick; they must assume that the user was auto-kicked.

### Version 1.2.1
- [ ] Iron Golem spawning in Villages
- [ ] Zombie Sieges
- [ ] Villager Breeding
- [ ] Wolf Breeding

### Version 1.3.1
- [ ] Wood Blocks (the ones with the texture on all sides)
- [ ] Villager trading
- [ ] Writing books
- [ ] Starting Chest
- [ ] Large Biomes Generator (?)

### Version 1.4.2
- [ ] Pumpkin Pie
- [ ] Wither Boss bar
- [ ] Whitch
- [ ] Mob Spawner specialized NBT tags to spawn equiped mobs
- [ ] NBT tags as potion descriptors
- [ ] Carrots and potatoes can be found in farms in villages

### Version 1.5
- [ ] Daylight sensor
- [ ] New Rails
- [ ] New Minecarts
- [ ] Abandoned mineshaft: Now spawn items in minecarts with chests, instead of chests
- [ ] Renamed item containers will now display that name when viewing the container's contents
- [ ] Renamed Command Blocks now use their name instead of @ in the chat

### Version 1.6
- [ ] Horse Riding
- [ ] Attribute system
- [ ] Added chest generation in nether fortresses

### Version 1.7
- [ ] colored glass / glass panels
- [ ] proper packed ice
- [ ] Block with mobs in them (e.g. silver fish)
- [ ] Minecart with command block
- [ ] Witch huts now generate with a witch inside.

This list does not claim be to be complete. I went through the patch notes on the wiki and tried out some stuff. It's just a start.

I'll try to link the matching issues

AndreaSmetankova:
Hi, I would like to help but I'm new to this and find this repositar only recently. Can you advise me where to start?

12xx12:
Hi,

[...]

PR #4970, opened 2020-10-06
TNT Changes

KingCol13:
Mum said it was my turn to play with the TNT code.

- Make TNT explosions always drop pickups (1.14 behaviour)
- Always drop items that wiki says should always drop
- Get rid of destination and just keep direction and intensity (seems to be wiki behaviour)
- Give each ray in explosion different random intensity

### DIFF
--- a/src/Physics/Explodinator.cpp
+++ b/src/Physics/Explodinator.cpp
@@ -201,19 +201,13 @@ namespace Explodinator
 	}
 
 	/** Traces the path taken by one Explosion Lazor (tm) with given direction and intensity, that will destroy blocks until it is exhausted. */
-	static void DestructionTrace(cChunk * a_Chunk, Vector3f a_Origin, const Vector3f a_Destination, const unsigned a_Power, const bool a_Fiery, float a_Intensity, const cEntity * const a_ExplodingEntity)
+	static void DestructionTrace(cChunk * a_Chunk, Vector3f a_Origin, const Vector3f a_Step, const unsigned a_Power, const bool a_Fiery, float a_Intensity, const cEntity * const a_ExplodingEntity)
 	{
 		// The current position the ray is at.
 		auto Checkpoint = a_Origin;
 
-		// The total displacement the ray must travel.
-		const auto TraceDisplacement = (a_Destination - a_Origin);
-
-		// The displacement that they ray in one iteration step should travel.
-		const auto Step = TraceDisplacement.NormalizeCopy() * StepUnit;
-
-		// Loop until we've reached the prescribed destination:
-		while (true)//(TraceDisplacement > (Checkpoint - a_Origin))
+		// Loop until intensity runs out:
+		while (true)
 		{
 			auto Position = Checkpoint.Floor();
 			if (!cChunkDef::IsValidHeight(Position.y))
@@ -242,7 +236,7 @@ namespace Explodinator
 			a_Chunk = Neighbour;
 
 			// Increment the simulation, weaken the ray:
-			Checkpoint += Step;
+			Checkpoint += a_Step;
 			a_Intensity -= StepAttenuation;
 		}
 	}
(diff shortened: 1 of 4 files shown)


bearbin (commented):
Looks OK, but I'll let @tigerw give this a proper review.

It would be nice to have a proper way to configure the explodinator wrt blocks being thrown around, and the probabilities etc - but I'm not sure if this PR is the right place for that.

tigerw:
Haven't forgotten this

tigerw:
@KingCol13 the tracing strategy was changed from YP/YM as a pair & 4 sides as a quadruple to YP/YM, ZP/ZM, XP/XM as pairs, if I'm reading this right. Did you see benefits to using the latter?

KingCol13:
@tigerw yep I changed it, I couldn't understand the YP/YM and 4 side system despite the ASCII diagrams and I didn't see any performance impact in changing. To me, each pair of sides having it's own loop seems a little easier to understand, but that's fairly subjective.

I also changed the implementation to use a 2x2 cube centred on the origin as in this diagram: https://minecraft.gamepedia.com/File:ExplosionRay.png, but again this change doesn't change the outcome.

src/Physics/Explodinator.cpp
```diff
@@ -214,59 +232,48 @@ namespace Explodinator
...
-		const int HalfSide = TraceCubeSideLength / 2;
 
 		// The following loops implement the tracing algorithm described in http://minecraft.gamepedia.com/Explosion
 
 		// Trace rays from the explosion centre to all points in a square of area TraceCubeSideLength * TraceCubeSideLength
-		// in the YM and YP directions:
-		for (int OffsetX = -HalfSide; OffsetX < HalfSide; OffsetX++)
+		// Careful to avoid duplicates along edges
+		// Top and bottom sides:
+		for (float OffsetX = -1; OffsetX < 1; OffsetX+=GridPointSpacing)
```
tigerw, line 251: As far as I can tell vanilla behaviour is using `int`s here so we should keep it as before

Titaniumtown:
About that 1.14 tnt item dropping behavior, isn't Cuberite now going to try to focus on older features first? https://github.com/cuberite/cuberite/issues/4888#issuecomment-704315535

mathiascode:
In my opinion, forcing (volunteer) contributors to do specific things is a good way to get rid of them. Of course you should listen to users as much as possible, but I feel most open source projects just don't have the liberty of being too picky.

tigerw:
I have returned! With measurements of the performances.

Off the cuff test, running DamageBlocks 10000 times with a fixed world:

1. PR verbatim: 140s
2. master: 124s
3. This PR, but with integer offsets: 119s

If I had to explain the numbers I'd say using floating points in the loop counters is slower. Separating the tracing loops vs having front/back/left/right done all at once looks like margin-of-error stuff, maybe the former helps compiler optimisation a little, I don't know.

I agree the tracing as it stands in master is unclear, even with the diagrams. @KingCol13 any objections to me changing the offset to int and merging?

tigerw:
About the features, as I understand there are two points of consideration; one is the version a given MC feature was introduced, another is version of the feature's behaviour we implement.

For example, explosions (super early feature), and whether explosions always drop pickups (1.14 apparently).

The linked discussion was about focusing on implementing features in ascending order of introduction to avoid a piecemeal experience of supported vs unsupported features. On the other hand, from what I can tell our policy of nevertheless implementing the latest version's behaviour, when a feature is added, is still the same.

KingCol13:
@tigerw Thanks for doing the proper tests! I’d just assumed the loop increments would be negligible compared to the raytracing but the numbers don’t lie and this is definitely a sizeable speedup.

No objections at all, go ahead with the ints and merge. Cheers for the help.

tigerw (approved):

PR #4972, opened 2020-10-07
Beacon now checks for players in surrounding square

12xx12:
Any beacon checks now for any player in a BoundingBox around it and applies it's effects.

This way this is less performant with few players but way more performant with many players.

### DIFF
--- a/src/BlockEntities/BeaconEntity.cpp
+++ b/src/BlockEntities/BeaconEntity.cpp
@@ -225,7 +225,7 @@ void cBeaconEntity::GiveEffects(void)
 		return;
 	}
 
-	int Radius = m_BeaconLevel * 10 + 10;
+	double Radius = static_cast<double>(m_BeaconLevel) * 10 + 10;
 	short EffectLevel = 0;
 	if ((m_BeaconLevel >= 4) && (m_PrimaryEffect == m_SecondaryEffect))
 	{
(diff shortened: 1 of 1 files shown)


src/BlockEntities/BeaconEntity.cpp
```diff
@@ -234,28 +234,22 @@ void cBeaconEntity::GiveEffects(void)
 
 	bool HasSecondaryEffect = (m_BeaconLevel >= 4) && (m_PrimaryEffect != m_SecondaryEffect) && (m_SecondaryEffect > 0);
 
-	Vector3d BeaconPosition(m_Pos);
-	GetWorld()->ForEachPlayer([=](cPlayer & a_Player)
+	auto Area = cBoundingBox(m_Pos, Radius, cChunkDef::Height - 1, static_cast<double>(-m_Pos.y));
```
bearbin, line 237: According to the wiki, this should actually be:

```suggestion
	auto Area = cBoundingBox(m_Pos, Radius, Radius + static_cast<double(cChunkDef::Height), -Radius);
```

That way the AABB extends above the beacon by 256 blocks, and below it by Radius.

tonibm19:
hi admin

you hope you doing well

thanks for contacting me

i really was a wonderfull day on your office

i hope we can talk together again

vanessa

bearbin (approved):

PR #4973, opened 2020-10-07
Simple end gen

NiLSPACE:
Implements a new end generator:

Image2

Vanilla also has various tiny islands, but I believe it's better to create a finisher for that. I still have some ideas on how to use a 3d noise which in my opinion creates nicer looking islands, but I believe this will suffice for now.


PureTryOut:
I'm just looking at the picture, but does this not generate a portal (non-activated at first) that's (eventually) supposed to bring you back to the overworld yet? I guess this is literally just the generation of the islands?

12xx12:
You are correct. I'm currently working on the portal and towerr generation

Titaniumtown:
Woah! Good work! I'm glad there is more focus on properly implementing the end. Would love to see the game being able to be completed on Cuberite!

12xx12:
A quick note:

In 1.9 biomes were added in the end. One for the main island and some for the surrounding islands. 

The biome for the central island determines the size of the void around the central island

NiLSPACE:
The End has changed more than I thought. I guess i'll add those biomes the next time I'm working on The End. I plan to do that when working on the 3d noise variant. The biomes themselves shouldn't be too hard to do. I can use the distorted voronoi as a basis.

NiLSPACE:
@cuberite/core-team

12xx12:
I added the biomes to the todo list in the issue for the end.

I think this would be the next step before adding the other end stuff (villages, plant thingies,...)

bearbin:
LGTM, any reason this can't go in @cuberite/core-team ?

mathiascode (approved):

PR #4975, opened 2020-10-08
Obsidian pillars, end fountain, Ender dragon spawning (The one with the cursed build)

12xx12:
This PR adds the obsidian pillars and the end fountain to the end generator.

cage around one of the pillars

Top view

- The pillars are configurable
  - Radius of circle
  - The height, radius and if the tower should have a cage
- The end fountain uses a ```.scematic``` file to be generated from


NiLSPACE:
I think it's better to have the end portal as a prefab instead of hardcoding the blocks in the sourcecode. I quickly made an endportal and exported it on the Gallery server:

```lua

-- EndPortal.cubeset

-- Defines the prefabs in the group EndPortal

-- NOTE: This file has been generated automatically by GalExport!
-- Any manual changes will be overwritten by the next automatic export!




Cubeset =
{
	Metadata =
	{
		CubesetFormatVersion = 1,
		ExportDate = "2020-10-08 15:17:04",
	},

	Pieces =
	{
		{
			OriginData =
			{
				ExportName   = "EndPortal",
				Name         = "Plains 405",
				GalleryName  = "Plains",
				GalleryIndex = "405",
				ID           = "1714",
				CreatorName  = "NiLSPACE",
			},
			Size =
			{
				x = 7,
				y = 5,
				z = 7,
			},
			Hitbox =
			{
				MinX = 0,
				MinY = 0,
				MinZ = 0,
				MaxX = 6,
				MaxY = 4,
				MaxZ = 6,
			},
			Connectors =
			{
			},
			Metadata =
			{
				["AddWeightIfSame"] = "0",
				["AllowedRotations"] = "7",
				["DefaultWeight"] = "100",
				["DepthWeight"] = "",
				["ExpandFloorStrategy"] = "None",
				["IsStarting"] = "0",
				["MergeStrategy"] = "msSpongePrint",
				["MoveToGround"] = "0",
			},
			BlockDefinitions =
			{
				".:  0: 0",  -- air
				"a:  7: 0",  -- bedrock
				"b: 50: 4",  -- torch
				"c: 50: 2",  -- torch
				"d: 50: 1",  -- torch
				"e: 50: 3",  -- torch
				"m: 19: 0",  -- sponge
			},
			BlockData =
			{
				-- Level 0
				"mmmmmmm",  --  0
				"mmaaamm",  --  1
				"maaaaam",  --  2
				"maaaaam",  --  3
				"maaaaam",  --  4
				"mmaaamm",  --  5
				"mmmmmmm",  --  6

				-- Level 1
				"mmaaamm",  --  0
				"ma...am",  --  1
				"a.....a",  --  2
				"a..a..a",  --  3
				"a.....a",  --  4
				"ma...am",  --  5
				"mmaaamm",  --  6

				-- Level 2
				".......",  --  0
				".......",  --  1
				".......",  --  2
				"...a...",  --  3
				".......",  --  4
				".......",  --  5
				".......",  --  6

				-- Level 3
				".......",  --  0
				".......",  --  1
				".......",  --  2
				"...a...",  --  3
				".......",  --  4
				".......",  --  5
				".......",  --  6

				-- Level 4
				".......",  --  0
				".......",  --  1
				"...b...",  --  2
				"..cad..",  --  3
				"...e...",  --  4
				".......",  --  5
				".......",  --  6

			},
		},  -- EndPortal
	},  -- Pieces
}





```

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,252 @@
...
+	// Choose random height
+	int Index = m_Noise.IntNoise3DInt(a_AbsPos) % m_TowerHeights.size();
+	do
+	{
+		Index = (Index + 1) % m_TowerHeights.size();
+		Properties = m_TowerHeights[Index];
+	} while (Properties.m_Height == 0);
+
+	// Make sure the tower does not cross chunk boarders
+	Pos.x = std::max(Pos.x, Properties.m_Radius - 1);
```
NiLSPACE, line 192: We have the function `Clamp` for this.

src/Generating/EnderDragonFightStructuresGen.h
```diff
@@ -0,0 +1,53 @@
...
+		bool m_HasCage;
+
+		sTowerProperties(int a_Height, int a_Radius, bool a_HasCage) :
+			m_Height(a_Height),
+			m_Radius(a_Radius),
+			m_HasCage(a_HasCage)
+		{
+		}
+	};
+	std::array<sTowerProperties, 10> m_TowerHeights =
```
NiLSPACE, line 32: Isn't it easier to determine these properties on the spot using an intnoise?
```cpp
int height = 70 + m_Noise.IntNoise3dInt(TowerPos) % 30;
int radius = 3 + m_Noise.IntNoise3dInt(TowerPos) % 3;
bool hasCage = (m_Noise.IntNoise3dInt(TowerPos) % 10) < 2;
```

12xx12: Yes it would be but the notchian implementation does have exakt value for that:
https://minecraft.gamepedia.com/End_spike

You clould do this on the fly if you like as there is a system behind the tower proportions

NiLSPACE: I see. How about we make it configurable? Something like:
```ini
ObsydianPillars=76|3|false, 79|3|true, 82|3|true, ...
```

The code to generate the positions in a circle would have to be modified to handle different amounts of pillars though.

12xx12: That can be easily done. The code simply chooses an angle an starts placing towers evenly spaced in a circle. You can change the angle between towers to accommodate for more towers

12xx12: Do you think the radius of the circle should be modified or adjusted by the cound of towers

10 Towers -> 43 radius
20 Towers -> 86 radius
...

NiLSPACE: You could make that configurable as well :)
```ini
ObsydianPillarsRadius=43
```

Titaniumtown:
Even more progress to make the game able to be completed on Cuberite! Amazing job guys!

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,252 @@
...
+	Pos.x = std::min(Pos.x, cChunkDef::Width - Properties.m_Radius);
+	Pos.z = std::max(Pos.z, Properties.m_Radius - 1);
+	Pos.z = std::min(Pos.z, cChunkDef::Width - Properties.m_Radius);
+
+	// Place obsidian pillar
+	for (int Y = 0; Y <= Properties.m_Height - 2; Y++)
```
NiLSPACE, line 198: Why not go top to bottom and break when reaching something other than air?

12xx12:
About the Prefab:

I have no clue how to use these except with the piece pool.
But for this use I don't think carrying around a whole object that powerfull is neccecary. This is only spawned when the end is initially generated

If there is another way feel free to push into this PR

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -5,22 +5,34 @@
...
+			LOGWARNING("Got unmatching parameters on generating obsidian pillars: %s, Please use \"Height|Radius|HasCage\", ...", TowerProperty);
+			continue;
+		}
+		int Height = std::stoi(TowerPropertyVector[0]);
+		int Radius = std::stoi(TowerPropertyVector[1]);
+		bool HasCage = false;
+		if (NoCaseCompare(TowerPropertyVector[2], "true") == 0)
+		{
+			HasCage = true;
+		}
```
NiLSPACE, line 27: Perhaps there should be a warning when an unexpected value is used here?

NiLSPACE:
I believe you still have to add the generator to the generating test's cmake file. `tests/Generating/CMakeLists.txt`

NiLSPACE:
I believe you missed the header file in the cmakelist.

For the prefab, as it's just a single structure, we could just put the schematic in the Prefabs folder and use `cSchematicFileSerializer::LoadFromSchematicFile(area, filePath)` to load the prefab into a cBlockArea.

12xx12:
I just copied the error from the MVSC and didn't bother to change the datatypes...

Now one of the builds refused to finish...

12xx12:
The ```cSchematicFileSerializer::LoadFromSchematicFile(area, filePath)``` doesn't accept cubesets. Only ```.scematic``` files with NBT description of the world

NiLSPACE:
Yes, I already exported it: [EndPortal_EndPortal.zip](https://github.com/cuberite/cuberite/files/5353774/EndPortal_EndPortal.zip)

The cubeset format is mostly focussed on multiple structures or pieces. The end portal however is a single structure which only occurs once. For this reason I think having a schematic instead of a cubeset is fine. I'm not sure how others feel about this (@cuberite/core-team) as we don't have allot of binary files in the repo.

12xx12:
Ok with the scematic file everything works fine.
Except that the torches are placed one block to high. I moved them already on the galery server

NiLSPACE:
Ah, I've re-exported it: [EndPortal_EndPortal.zip](https://github.com/cuberite/cuberite/files/5353939/EndPortal_EndPortal.zip)

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,282 @@
...
+		(a_ChunkDesc.GetChunkZ() > 6) ||
+		(a_ChunkDesc.GetChunkZ() < -6))
+	{
+		return;
+	}
+	// Places the exit portal
+	auto Coords = a_ChunkDesc.GetChunkCoords();
+	if (Coords == cChunkCoords({0, 0}))
+	{
+		a_ChunkDesc.WriteBlockArea(m_Fountain, -3, 62, -3, cBlockArea::msSpongePrint);
```
NiLSPACE, line 85: Would these coordinates work when I use a portal with a different size?

NiLSPACE:
@bearbin do you know why travis fails? It seems like there is some kind of timeout.

src/Generating/EnderDragonFightStructuresGen.cpp
12xx12: No

12xx12: This could be easily fixed. Whyever you would change the prefab. When the dragon dies it will spawn the portal. This spawning would know the size of the portal

12xx12: We could make this configurable. We just need to the the space where the exit portal should spawn

NiLSPACE: If the endportal prefab is configurable it makes sense that the generator knows how to handle prefabs with different sizes right?

12xx12: Yeah of course - but we need to note down the size somewhere for the ender dragon spawn the exit portal

or we search the area araound (0, 62, 0) and spawn in the area around it till we hit bedrock. Although this will break the server if someone ever breaks the bedrock of the fountain in creative mode.

Another thing i noticed that the end portal regenerated if someone travels to the end. That should be done by us also.

that still might break the exit portal

NiLSPACE: The obsidian floor at spawn also regenerates everytime something travels to the end. Isn't the portal always at exactly 0,y,0 ?

NiLSPACE: The regenerating part is harder for us at the moment as the generator isn't build to regenerate specific structures in existing chunks.

12xx12: > The obsidian floor at spawn also regenerates everytime something travels to the end.
Oh yeah... that thing...

> Isn't the portal always at exactly 0,y,0 ?
It is

bearbin:
Build failure looks like the build slave just hung, which happens sometimes. I've restarted it and we can see if it works properly this time.

tigerw:
> I'm not sure how others feel about this (cuberite/core-team) as we don't have allot of binary files in the repo.

No objections

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,282 @@
...
+		return;
+	}
+	PlaceTower(a_ChunkDesc, It->second);
+}
+
+
+
+
+
+void cEnderDragonFightStructuresGen::PlaceFountainSouthEast(cChunkDesc & a_ChunkDesc) const
```
NiLSPACE, line 115: These methods can now be removed as we're using a cBlockArea

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,190 @@
...
+	// Generate Positions in a circle
+	for (int i = 0; i < static_cast<int>(m_TowerProperties.size()); i++)
+	{
+		auto TowerPos = Vector3i(static_cast<int>(FloorC(a_Radius * cos(Angle))), 0, static_cast<int>(FloorC(a_Radius * sin(Angle))));
+		auto ChunkX = static_cast<int>(FloorC(TowerPos.x / cChunkDef::Width));
+		if (TowerPos.x < 0)
+		{
+			ChunkX--;
+		}
+		auto ChunkZ = static_cast<int>(FloorC(TowerPos.z / cChunkDef::Width));
```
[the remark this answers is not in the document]
peterbell10: There are lots in `GenFinish` as well.

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,190 @@
...
+		if (TowerPos.z < 0)
+		{
+			ChunkZ--;
+		}
+
+		m_TowerPos[cChunkCoords(ChunkX, ChunkZ)] = TowerPos;
```
peterbell10, line 63: Are towers chunk-aligned in the notchian implementation?

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,190 @@
...
+		for (int X = -2; X <= 2; ++X)
+		{
+			for (int Z = -2; Z <= 2; ++Z)
+			{
+				a_ChunkDesc.SetBlockType(Pos.x + X, Properties.m_Height + 2, Pos.z + Z, E_BLOCK_IRON_BARS);
+			}
+		}
+	}
+	// Set the height to zero that this tower has been generated
+	m_TowerProperties[Index].m_Height = 0;
```
peterbell10, line 189: Any reason not to remove the tower from the list entirely?
```suggestion
	std::swap(m_TowerProperties[Index], m_TowerProperties.back());
	m_TowerProperties.resize(m_TowerProperties.size() - 1);
```

src/Generating/EnderDragonFightStructuresGen.h
```diff
@@ -0,0 +1,37 @@
...
+		int m_Height;
+		int m_Radius;
+		bool m_HasCage;
+
+		sTowerProperties(int a_Height, int a_Radius, bool a_HasCage) :
+			m_Height(a_Height),
+			m_Radius(a_Radius),
+			m_HasCage(a_HasCage)
+		{
+		}
```
peterbell10, line 30: You don't need a constructor, just use aggregate initialization:
```cpp
sTowerProperties{Height, Radius, HasCage};
```

tests/Generating/Stubs.cpp
```diff
@@ -404,3 +406,704 @@ cItem::cItem(
...
+
+cPawn::~cPawn(void)
+{
+}
+
+
+
+
+
+void __cdecl cPawn::Destroyed(void)
```
peterbell10, line 828:
```suggestion
void cPawn::Destroyed(void)
```

src/Generating/EnderDragonFightStructuresGen.cpp
12xx12: Yeah - there might be some tinkering to do if we want to place them 100% right

src/Generating/EnderDragonFightStructuresGen.cpp
12xx12: No they aren't. I could do some thinking on making it work. I have an idea

12xx12:
As the error on the clang travis ARM build still fails I think this might be related the way of choosing the tower height.

I had a problem during development, that the choosing of the tower to place because the noise didn't generate all numbers in the range. I think i fixed that but it still might be there.

@bearbin does the buildserver kill the process after some time?
I'll try to ad a log that prints the chosen tower and commit again to check if this causes the error

bearbin:
The buildserver does kill the process if it hangs for 10 or so minutes - and it looks like this code is actually hanging when compiled with clang 6 on arm64 for some reason, not just a random thing since it's repeatable.

I'm surprised that Cuberite is hanging without the deadlock detector kicking in though... maybe this needs more investigation.

12xx12:
it seems the generator is not covered by the deadlock detector.

I'm gonna add the debug output and push to see whats the cause

NiLSPACE:
The deadlock detector only detects deadlocks in world threads. Perhaps the deadlock is somewhere else?

12xx12:
hm... in my generator there seems to be no deadlock. Everything works smoothly on setup and generation

12xx12:
The Log always stops during the generation in the nether. I don't think i changed somethin there

NiLSPACE:
Could it be a resource that both the netherfortress finisher and EnderDragonFightStructuresGen are accessing?

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,193 @@
...
+		(a_ChunkDesc.GetChunkZ() < -6))
+	{
+		return;
+	}
+	// Places the exit portal
+	auto Coords = a_ChunkDesc.GetChunkCoords();
+	if (Coords == cChunkCoords({0, 0}))
+	{
+		auto EnderDragon = std::make_unique<cEnderDragon>();
+		EnderDragon->SetPosition({0.0, 80.0, 0.0});
```
NiLSPACE, line 90: Perhaps `Clamp(a_ChunkDesc.GetHeight(0, 0) + 20, 0, cChunkDef::Height)` for y might be better. Cuberite has a really configurable world generator, so it's possible someone generates really extreme terrain which could cause the dragon to spawn inside endstone.

12xx12:
The only thing that is accessed from the end generator is the fountain prefab. And this in a seperate folder with another file ending.

12xx12:
The funky thing is that the issue only occurs on the ARM build.

In one build i tried if the initialsation passes. That works. It doesn't seem to be the file access.

NiLSPACE:
Hmm.. It's also really strange that it stops in the middle of a log. As if the main thread is blocked while flushing.

12xx12:
Ok I checked the builds. The error is there before I added the prefab import. So I think the error is further down the line

NiLSPACE:
The first commit that has this where clang did compile is 4ae80f2b4194f26261a4b0495d18950d02013001

12xx12:
Everythign else before that is just a Noise Generator from a seed and the fountain being placed manually and access to the provided chunkinterface

NiLSPACE:
Perhaps the problem lies in the initialization of the end structures but the log wasn't flushed to std::out yet. We might think the nether is still generating, but it's possible that has already finished.

Could you set the default finishers for the end back to empty? If that commit succeeds it's probably in the initialization part.

NiLSPACE:
It seems to be stuck again. Does anyone have a Raspberry pi or something to try this out?

12xx12:
i'm confused...

If we remove the default finisher that means my code doesn't run at all... So this issue isn't cause by this PR if I understand

NiLSPACE:
It certainly looks like it, but for some reason only this PR is affected by it.

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,193 @@
...
+				{
+					a_ChunkDesc.SetBlockType(NewPos.x, NewPos.y, NewPos.z, E_BLOCK_OBSIDIAN);
+				}
+			}
+		}
+	}
+	// Spawn the bedrock
+	a_ChunkDesc.SetBlockType(Pos.x, Properties.m_Height - 1, Pos.z, E_BLOCK_BEDROCK);
+	// Spawn the Ender Crystal - the original passed position is not used because the tower might have been moved due to crossing chunk borders
+	auto EnderCrystal = std::make_unique<cEnderCrystal>(Vector3d(Pos.x + cChunkDef::Width * a_ChunkDesc.GetChunkX() + 0.5, Properties.m_Height, Pos.z + cChunkDef::Width * a_ChunkDesc.GetChunkZ() + 0.5));
```
peterbell10, line 157:
```suggestion
	auto EnderCrystal = std::make_unique<cEnderCrystal>(a_AbsPos + Vector3d(0.5, Properties.m_Height, 0.5));
```

TheUserCreated:
This does in fact build perfectly fine on my RasPi 4 using either clang 7 or GCC 8.3

bearbin:
After some hard work trying to get this reproduced on real hardware, no success was found. The build succeeded on a 64bit raspi. 

I think the next thing to do is to say this PR is contaminated, and close it. Create another with identical code (but new commits rebased off master) and see what happens.

peterbell10:
I re-run the build with the travis caches cleared and it still failed. That and the fact this is happening consistently suggests it really is the PR that's causing it. @12xx12 what happens if you remove the m_FinishGens.push_back` line entirely?

12xx12:
Gonna do that ASAP.

Might it be related to the way I access the ini file during generation?

12xx12:
I just  noticed that it takes "ages" compared to the other files to build the EnderDragoinFightStructureGen.cpp during testing

12xx12:
Ok the execution still fails if the creation of the generator object is not done at all...

I'm gonna remove everything in ComposebleGeneratorIadded

12xx12:
I'm confused..

I removed the new generator from the cMakelist.txt file and the executionm still crashes

12xx12:
Please continue conversation on PR #4993

I you have anything about the failing I think this PR is the right place.

@cuberite/core-team Do we keep this open to further inspect the issue or do we assing this to the gost in the machine?

bearbin:
Closing in favour of #4993, let's keep this safely quarantined.

PR #4976, opened 2020-10-08
Anvil fixes

### LINKED ISSUE #4974: Anvil Durability Bugs
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: a9031b6bae742b333b1b390192fa590f2ecb07ea

### Expected behavior
Can't combine a full durability tool with an unenchanted tool
Tool appears to take damage after being repaired.

### Actual behavior
Can combine full durability tools with unenchanted tools.
Tool doesn't appear to take damage after being repaired.

Combine two max durability tools and mine blocks with it.

KingCol13:
Fixes #4974 

- No fixing tools to negative damages

- No result if the output will be the same as the target

### DIFF
--- a/src/UI/SlotArea.cpp
+++ b/src/UI/SlotArea.cpp
@@ -1138,6 +1138,7 @@ void cSlotAreaAnvil::UpdateResult(cPlayer & a_Player)
 		Output.Empty();
 		SetSlot(2, a_Player, Output);
 		m_ParentWindow.SetProperty(0, 0);
+		m_MaximumCost = 0;
 		return;
 	}
 
@@ -1161,6 +1162,7 @@ void cSlotAreaAnvil::UpdateResult(cPlayer & a_Player)
 				Output.Empty();
 				SetSlot(2, a_Player, Output);
 				m_ParentWindow.SetProperty(0, 0);
+				m_MaximumCost = 0;
 				return;
 			}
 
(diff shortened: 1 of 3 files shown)


bearbin (approved):

src/UI/SlotArea.cpp
```diff
@@ -1248,22 +1252,31 @@ void cSlotAreaAnvil::UpdateResult(cPlayer & a_Player)
...
+		Output.m_RepairCost = RepairCost;
+	}
+
+	// If after everything, output will be the same then no point enchanting
+	if (Target.IsEqual(Output))
+	{
+		Output.Empty();
+		SetSlot(2, a_Player, Output);
+		m_ParentWindow.SetProperty(0, 0);
+		return;
```
peterbell10, line 1276: If you look on line 1039, I think this would still cost the player XP. Instead of special casing the `SetProperty`, you need to set `m_MaximumCost` correctly.
```suggestion
		m_MaximumCost = 0;
```

peterbell10 (dismissed):

src/UI/SlotArea.cpp
peterbell10: Well actually, I guess if the output slot is empty you can't actually take it anyway. Either way, keeping `m_MaximumCost` up to date makes sense to me.

PR #4977, opened 2020-10-09
Include sys/select.h on non-glibc Linux platforms

PureTryOut:
Otherwise fd_set will be unknown for example on Musl libc systems

### DIFF
--- a/src/Root.cpp
+++ b/src/Root.cpp
@@ -13,6 +13,10 @@
 	#include <signal.h>
 	#if defined(__linux__)
 		#include <fstream>
+
+		#if !defined(__GLIBC__)
+			#include <sys/select.h>
+		#endif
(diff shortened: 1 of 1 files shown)


bearbin (approved):
Looks good: travis is failing for unrelated reasons - CMake's download servers are failing for the download of our version of CMake.

Since you've started working on this, it would be nice if you could resurrect #4799 - there's no reason it couldn't be merged in as is, xoft just got a bit burned out in the review.

PureTryOut:
Uh, I'll be honest, I know next to nothing of C/C++ :sweat_smile: 

I can only fix simple issues like this (missing includes, common glibc/Musl incompabilities with solutions found by quick internet searches), I wouldn't even know where to start with #4799 sadly.

PR #4980, opened 2020-10-09
Add default value nullptr for parameter a_Digger and added digger to cWorld:DigBlock in lua api

Seadragon91:
Pull request #4967 broke api function cWorld:DigBlock, by making the digger not optional.
Lua code like:
No digger passed: `cRoot:Get():GetDefaultWorld():DigBlock(1, 1, 1)` fails, but should work
Adding nil for digger: `cRoot:Get():GetDefaultWorld():DigBlock(1, 1, 1, nil)` works
Fixed by adding nullptr to a_Digger.

Also added overloaded function for Vector3i in API Doc.
Copied the sentence `The optional Digger parameter specifies the entity who dug the block, usually a player` from function `DropBlockAsPickups`.

### DIFF
--- a/Server/Plugins/APIDump/Classes/World.lua
+++ b/Server/Plugins/APIDump/Classes/World.lua
@@ -541,28 +541,56 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 			},
 			DigBlock =
 			{
-				Params =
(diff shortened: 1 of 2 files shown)


tigerw (approved):
Thanks!

PR #4982, opened 2020-10-09
Readding enderman spawning

12xx12:
Yeah the end stuff is getting further and the enderman are missing...

The TODO said that there should be  a bug that needs fixing. I checked with the issues. there aren't andy bugs with enderman in them

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -880,9 +880,9 @@ void cWorld::InitializeAndLoadMobSpawningValues(cIniFile & a_IniFile)
 	AString DefaultMonsters;
 	switch (m_Dimension)
 	{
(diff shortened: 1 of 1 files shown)


KingCol13:
I just messed with them and they didn't seem to do any damage to me while they were attacking, maybe that's what it meant?

12xx12:
They don't even take blocks if i remember correctly... had a look in the code

bearbin:
I think we need proper _working_ enerman AI before this can be merged in. If you want to have a go at implementing that, have at it, otherwise this should be closed.

12xx12:
Closing this PR. Might feeling like reopening later.

PR #4984, opened 2020-10-10
Remove debug output on PieceStructuresGen.cpp

12xx12:
In the single piece structures PR we discussed that this should be removed

This PR is mostly changing variable names to the new NamingSceme

### DIFF
--- a/src/Generating/PieceStructuresGen.cpp
+++ b/src/Generating/PieceStructuresGen.cpp
@@ -8,7 +8,6 @@
 #include "PrefabStructure.h"
 #include "PieceGeneratorBFSTree.h"
 #include "../IniFile.h"
-#include "../Stopwatch.h"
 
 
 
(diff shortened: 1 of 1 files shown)


src/Generating/PieceStructuresGen.cpp
```diff
@@ -63,11 +63,10 @@ class cPieceStructuresGen::cGen:
 	// cGridStructGen overrides:
 	virtual cStructurePtr CreateStructure(int a_GridX, int a_GridZ, int a_OriginX, int a_OriginZ) override
 	{
-		cStopwatch sw(Printf("CreateStructure for %s at <%d, %d>", m_Name.c_str(), a_GridX, a_GridZ));
```
12xx12, line 66: Removing this is the main change

bearbin (changes requested):
Please also remove the include for Stopwatch.h, if it is no longer required.

bearbin (approved):

PR #4992, opened 2020-10-11
Update comments and code for Load/SaveOneChunk so they match

bearbin:


peterbell10 (approved):

bearbin:
I had a look in the git blame, this was leftover code from when there were callbacks for chunk loading/unloading.

PR #4993, opened 2020-10-11
Obsidian pillars, end fountain, not Ender dragon spawning

12xx12:
This PR adds the obsidian pillars, the end fountain and the initial ender dragon to the end generator.

cage around one of the pillars

Top view

- The pillars are configurable
  - Radius of circle
  - The height, radius and if the tower should have a cage
- The end fountain uses a ```.scematic``` file to be generated from

Todo:
- [x] Generate the towers across chunk borders


NiLSPACE:
It's alive!

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,191 @@
...
+	// If not in the chunks to write
+	if ((a_ChunkDesc.GetChunkX() > 6) ||
+		(a_ChunkDesc.GetChunkX() < -6) ||
+		(a_ChunkDesc.GetChunkZ() > 6) ||
+		(a_ChunkDesc.GetChunkZ() < -6))
+	{
+		return;
+	}
+	// Places the exit portal
+	auto Coords = a_ChunkDesc.GetChunkCoords();
```
bearbin, line 83: Looks like things could be simplfiied here, if the Coords line was just moved a little bit earlier.

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+	auto Pos = cChunk::AbsoluteToRelative(a_AbsPos);
+	sTowerProperties Properties = {0, 0, false};
+	// Choose random height
+	unsigned long Index = static_cast<unsigned long>(m_Noise.IntNoise3DInt(a_AbsPos)) % m_TowerProperties.size();
+	do
+	{
+		Index = (Index + 1) % m_TowerProperties.size();
+		Properties = m_TowerProperties[Index];
+	} while (Properties.m_Height == 0);
```
bearbin, line 129: What have we got here? An infinite loop when m_Height is 0 for all towers.

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+	// Make sure the tower does not cross chunk boarders
+	Pos.x = Clamp(Pos.x, Properties.m_Radius - 1, cChunkDef::Width - Properties.m_Radius);
+	Pos.z = Clamp(Pos.z, Properties.m_Radius - 1, cChunkDef::Width - Properties.m_Radius);
```
bearbin, line 133: No validation on m_Radius?

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,191 @@
...
+			{
+				Vector3i NewPos = {Pos.x + X, Y, Pos.z + Z};
+				// The 3 was achieved by trial and error till the shape matched the notchian implementation
+				if ((NewPos - Vector3i(Pos.x, Y, Pos.z)).SqrLength() < Properties.m_Radius * Properties.m_Radius - 3)
+				{
+					a_ChunkDesc.SetBlockType(NewPos.x, NewPos.y, NewPos.z, E_BLOCK_OBSIDIAN);
+				}
+			}
+		}
+	}
```
bearbin, line 150: The if-condition does not use `Y` at all here, it would be much more efficient if the Y-loop was the inner loop here.

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,191 @@
...
+			a_ChunkDesc.SetBlockType(Pos.x + 2, Properties.m_Height + 1, Pos.z + I, E_BLOCK_IRON_BARS);
+
+			a_ChunkDesc.SetBlockType(Pos.x + I, Properties.m_Height - 1, Pos.z - 2, E_BLOCK_IRON_BARS);
+			a_ChunkDesc.SetBlockType(Pos.x + I, Properties.m_Height, Pos.z - 2, E_BLOCK_IRON_BARS);
+			a_ChunkDesc.SetBlockType(Pos.x + I, Properties.m_Height + 1, Pos.z - 2, E_BLOCK_IRON_BARS);
+
+			a_ChunkDesc.SetBlockType(Pos.x + I, Properties.m_Height - 1, Pos.z + 2, E_BLOCK_IRON_BARS);
+			a_ChunkDesc.SetBlockType(Pos.x + I, Properties.m_Height, Pos.z + 2, E_BLOCK_IRON_BARS);
+			a_ChunkDesc.SetBlockType(Pos.x + I, Properties.m_Height + 1, Pos.z + 2, E_BLOCK_IRON_BARS);
+		}
```
bearbin, line 179: This is rather unintuitive code and also places multiple bars - it should be refactored.

bearbin (changes requested):

12xx12:
Yeah i'm gonna redo the tower placement.

I'm not quit shure if I should draw them into a cBlockArea during initialization or if I should do this while execution

12xx12:
I tested it with radius 30 the whole tower won't generate - further research is need why it doesn't

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,272 @@
...
+	for (size_t I = 0; I < TowerProperties.size(); I++)
+	{
+		auto TowerPos = Vector3i(FloorC(a_Radius * cos(Angle)), 0, FloorC(a_Radius * sin(Angle)));
+		TowerProperties[I].m_Pos = TowerPos;
+
+		// Check all crossed chunks
+		const int & Width = cChunkDef::Width;
+		for (int X = -TowerProperties[I].m_Radius - Width; X <= TowerProperties[I].m_Radius + Width; X+=std::min(TowerProperties[I].m_Radius, Width))
+		{
+			for (int Z = -TowerProperties[I].m_Radius - Width; Z <= TowerProperties[I].m_Radius + Width; Z+=std::min(TowerProperties[I].m_Radius, Width))
```
12xx12, line 112: For some reason the ```cChunkDef::Width``` cant be accessed in the ```for``` statements...

12xx12: Hm - okl... it's only in the constructor...

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -57,7 +58,7 @@ const std::array<Vector3i, 26> cEnderDragonFightStructuresGen::m_CageAir =
 
 
 
-cEnderDragonFightStructuresGen::cEnderDragonFightStructuresGen(int a_Seed, const AString &a_TowerProperties, int a_Radius) :
+cEnderDragonFightStructuresGen::cEnderDragonFightStructuresGen(int a_Seed, const AString &a_TowerProperties, int a_Radius, int a_ChunkWidth) :
```
NiLSPACE, line 61: Why add this as a parameter when you can just access it from cChunkDef?

12xx12: It failed on linking with MVSC with the error that the symbol didn't exist. See my earlier comment

src/Generating/EnderDragonFightStructuresGen.cpp
12xx12: @NiLSPACE this one

src/Generating/EnderDragonFightStructuresGen.cpp
NiLSPACE: Whoops, gotta learn how to read I guess ;)

Would moving everything to some kind of `Initialize` method fix the problem?

12xx12: Gonna try that later 👍🏻

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,272 @@
...
+		{
+			Vector3i NewPos = {Pos.x + X, 1, Pos.z + Z};
+			if (cChunkDef::IsValidRelPos(NewPos))
+			{
+				// The 3 was achieved by trial and error till the shape matched the notchian implementation
+				if ((NewPos - Vector3i(Pos.x, 1, Pos.z)).SqrLength() < a_Properties.m_Radius * a_Properties.m_Radius - 3)
+				{
+					for (int Y = 0; Y <= a_Properties.m_Height - 2; Y++)
+					{
+						a_ChunkDesc.SetBlockType(NewPos.x, Y, NewPos.z, E_BLOCK_OBSIDIAN);
```
12xx12, line 223: Maybe we should check for malformated towers to prevent a player from generating tower that surpass the height limit

src/Generating/EnderDragonFightStructuresGen.h
```diff
@@ -0,0 +1,35 @@
+
+#pragma once
+
+/*
+https://minecraft.gamepedia.com/End_spike
```
12xx12, line 5: Add some links to the other things spawned with this PR

src/Generating/EnderDragonFightStructuresGen.cpp
12xx12: Now it's getting weird... Still doesn't compiler...

NiLSPACE:
I'm guessing it's because you're still initializing from the constructor. How about calling Init after you've constructed the object?

12xx12:
yeah but WHY doesn't this work :o (accessing the cChunkDef::Width)

NiLSPACE:
I'm not sure. I tried googling it, but only found issues where the compiler couldn't find the constructor itself. Perhaps someone more knowledgable with C++ knows.

NiLSPACE:
@tigerw do you know why Clang fails to compile?

12xx12:
This PR hates me :D

12xx12:
@cuberite/core-team

So this is kind of strange;
Any function that is called from the ````cComposebleGenecrator``` can't acces ```cChunkDef::Width``` for whatever reason.
This also includes methods called by methods called by... you know it

tigerw:
wat

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,282 @@
...
+	std::vector<sTowerProperties> TowerProperties;
+	for (const auto & TowerProperty : TowerPropertiesVector)
+	{
+		const auto TowerPropertyVector = StringSplitAndTrim(TowerProperty, "|");
+		if (TowerPropertyVector.size() != 3)
+		{
+			LOGWARNING("Got unknown parameters on generating obsidian pillars: %s, Please use \"Height|Radius|HasCage\"; ...", TowerProperty);
+			continue;
+		}
+		int Height = std::min(std::stoi(TowerPropertyVector[0]), cChunkDef::Height - 2);  // The highest block placed is two blocks above the given height (the cage above some towers)
```
tigerw, line 89: Try saying const auto MaxHeight = cChunkDef::Height - 2, then using MaxHeight. Seems like cChunkDef::Height is being treated as a true constant literal, no symbol emitted to the linker (yay!) but std::min takes arguments by reference (and literals have no address).

tigerw: Weird though, considering std::min asks for _const_ references, and clearly you can pass normal integral literals to it without complaint...

tigerw: Huh, apparently it's just the language.

```cpp
struct S {
    static const int x = 0; // static data member
    // a definition outside of class is required if it is odr-used
};
const int& f(const int& r);
 
int n = b ? (1, S::x) // S::x is not odr-used here
          : f(S::x);  // S::x is odr-used here: a definition is required

              ^^^^ 😮
```

https://en.cppreference.com/w/cpp/language/definition#ODR-use

12xx12:
Yeah the linker seems to work - thanks @tigerw

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+
+void cEnderDragonFightStructuresGen::Init(const AString & a_TowerProperties, int a_Radius)
+{
+	const auto ChunkWidth = cChunkDef::Width;
+	// Loads the fountain schematic
+	if (!cSchematicFileSerializer::LoadFromSchematicFile(
+		m_Fountain, AString("Prefabs") + cFile::GetPathSeparator() + "SinglePieceStructures" + cFile::GetPathSeparator() + "EndFountain.schematic"))
+	{
+		LOGWARNING("EnderDragonFightStructuresGen is missing it's end fountain prefab, please update your cuberite server files! There will be no end fountain!");
```
mathiascode, line 77:
```suggestion
		LOGWARNING("EnderDragonFightStructuresGen is missing its end fountain prefab, please update your Cuberite server files! There will be no end fountain!");
```

src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -0,0 +1,283 @@
...
+		{
+			HasCage = true;
+		}
+		else if (NoCaseCompare(TowerPropertyVector[2], "false") == 0)
+		{
+			HasCage = false;
+		}
+		else
+		{
+			LOGWARNING("Got unknown value for boolean if the tower: %s should have a cage! %s. Tower wont be generated!", TowerProperty, TowerPropertyVector[2]);
```
mathiascode, line 103:
```suggestion
			LOGWARNING("Got unknown value for boolean of the tower: %s should have a cage! %s. Tower will not be generated!", TowerProperty, TowerPropertyVector[2]);
```

mathiascode (commented):
Minor nitpick

mathiascode (approved):
Looks fine once changes are made to account for https://github.com/cuberite/cuberite/commit/be2bf999c280d6892b38272d3f783b3462f0b745.

12xx12:
this PR -.-

tigerw:
I know the feeling. Sleep on it, maybe?

PR #4998, opened 2020-10-13
Fix Android builds

tigerw:
Construct paths relative to the Cuberite sources with PROJECT_SOURCE_DIR, instead of wherever the first CMakeLists.txt file happened to be with CMAKE_SOURCE_DIR.

In Android's case, the latter was in a folder called android/ but that's not the root of the source tree, so any file path built off that root was wrong. This caused file-specific warnings exclusions to fail to apply.


peterbell10 (approved):
Personally I'd prefer `Cuberite_SOURCE_DIR` but this is a clear improvement either way, so whatever.

PR #5000, opened 2020-10-15
Adding tests for Namespace serializer

12xx12:
This adds a test for the namespace serializer

This simply uses the back and forth conversion on the namespace serializer and checks if the result is the same. Just to catch some typos.

Gonna add tests for #4941 when merged


tigerw:
The test makes sense but since the mapping is generated, wouldn't this just be testing that Mojang's registries are consistent?

12xx12:
You did autogenerate the mapping from the registries?

Yeah.. I did my additions by hand in another PR

tigerw:
They were generated, and as peterbell said we aim (ideally) to have a script generate the registries and palettes without any intervention. I'd say the Lua file could eventually be tested, but the output doesn't need tests.

(Time will tell how well this comment ages...)

PR #5001, opened 2020-10-16
cBlockInfo: further cleanup

tigerw:
+ Use switch statements instead of lookup in dynamically initialised arrays
- Remove some deprecated bindings

### DIFF
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -292,32 +292,6 @@ tolua_lerror:
 
 
 
-static int tolua_cBlockInfo_Get(lua_State * tolua_S)
-{
-	cLuaState L(tolua_S);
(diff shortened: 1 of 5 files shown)


12xx12:
How do you feel about adding every block type to the switch statements and adding a UNREACHABLE at the end? So that everybody adding new enum values is forced to add them there.

Or will this BlockInfo statements diappear when we did the transisiton to 1.13 Block handling?

tigerw:
Certainly switch will enable strict enforcement, but I don't want to do it now since it'll be a lot of work going through all the properties, for all the blocks to make sure they're correct... It'd be great if we found a source that described all the properties, like a datafile or registry, but I haven't seen any in the 1.13 generators.

tigerw:
merge?

bearbin (approved):

PR #5002, opened 2020-10-16
Adding Item Tags and struct for easy variant access

12xx12:
This PR adds item tags to cuberite.

https://minecraft.gamepedia.com/Tag#Items

It's a way in vanilla to describe multible blocks and check or obtain them.

It uses the same mechanic the StringToItem tool uses.
Item tags are defined in the items.ini - the server will stop and ask the user to download the new items.ini if the server is updated to this PR

This is nedded for cLootTables. I just try to make the diff smaller

This also adds some new aliases to the items.ini

I thought about adding a enum but this doesn't match the idea of "don't hardcode this stuff".


tigerw:
I would suggest autogen directly into C++ for this too, like the rest of the registry. Perf arguments aside, a items.ini configuration isn't that flexible either since it's a serverwide setting. Additionally I'd recommend using the Item registry for new code, this should also simplify generation since the names will be the same. I would gladly help if you need any.

12xx12:
**Generate into C++**
I did this already in the Loottable PR check there for an idea.

There I tried to to less allocation an did it a bit experimental.

**Use the new item registry**
Does this already work? I need some cItems objects to be returned

12xx12:
closed in favour of #4842

PR #5004, opened 2020-10-17
Adding some string utils split from loot tables

12xx12:
This PR adds two string utils functions I added for the loot tables.

### DIFF
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -1128,14 +1128,15 @@ void Capitalize(AString & a_String)
 
 
 
-AString NamespaceConverter(AString a_String)
+AString NamespaceConverter(AString & a_String)
 {
-	ReplaceString(a_String, "minecraft:", "");
-	ReplaceString(a_String, "_", " ");
-	Capitalize(a_String);
-	ReplaceString(a_String, " ", "");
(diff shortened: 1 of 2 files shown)


tigerw:
I'm not yet sure how NamespaceConverter would be used, but does NamespaceSerializer::SplitNamespacedID in WorldStorage\NamespaceSerializer.h not satisfy?

12xx12:
It also removes "_" and so on. But yeah. I think I can make it work wir that too

12xx12:
I just wanted to be able to support loot tables in minecraft:naming_convention but also in CuberiteNamingConvention

tigerw:
I'd say keep the namespaced IDs in their original form, it may confuse people if we create our own scheme.

@12xx12 would you like to close #5002 #5004, I don't mind looking at the large #4842.

12xx12:
Yeah we dropped the c++ parsing anyway

I'm gonna close those as soon as I did the changes to the main PR.

This week I don't have much time to finish it so this will wait until next week. Or maybe not (Corona stuff)

tigerw:
Thanks!

12xx12:
Closed in favour of #4842

PR #5010, opened 2020-10-28
Updating Ender crystal

### LINKED ISSUE #4990: End crystal places bedrock and fire 1 block too high
Client version: 1.12.2
Server OS: Raspbian GNU/Linux 10
Cuberite Commit id:
- b3f5b9dbc3fc3f7f31a8f7e75b8d7673f5e84378 built at: approx: 2020-10-10 21:05:21 +0200

### Expected behavior
On breaking the end crystal, it should leave behind the block it was placed on with fire on top, provided that block is not destroyed in the explosion.

### Actual behavior
The crystal creates a bedrock block & fire which are one block higher than expected. Currently it just does this with two SetBlock calls, which will give players a method of placing bedrock wherever they like once end crystals are placeable.

Break an end crystal
image
image
(fire went out before screenshot, was on top)

12xx12:
This PR adds the new data for the ender crystal entity.
Fixes #4990

### DIFF
--- a/src/Entities/EnderCrystal.cpp
+++ b/src/Entities/EnderCrystal.cpp
@@ -36,7 +36,7 @@ void cEnderCrystal::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 	// No further processing (physics e.t.c.) is needed
 	if (m_World->GetDimension() == dimEnd)
 	{
-		m_World->SetBlock({POSX_TOINT, POSY_TOINT + 1, POSZ_TOINT}, E_BLOCK_FIRE, 0);
+		m_World->SetBlock(POS_TOINT.addedY(1), E_BLOCK_FIRE, 0);
 	}
 }
 
@@ -52,7 +52,7 @@ void cEnderCrystal::KilledBy(TakeDamageInfo & a_TDI)
 
 	Destroy();
 
-	m_World->SetBlock(POSX_TOINT, POSY_TOINT,     POSZ_TOINT, E_BLOCK_FIRE,    0);
+	m_World->SetBlock(POSX_TOINT, POSY_TOINT, POSZ_TOINT, E_BLOCK_FIRE, 0);
 }
 
 
--- a/src/Entities/EnderCrystal.cpp
+++ b/src/Entities/EnderCrystal.cpp
@@ -10,8 +10,9 @@
 
 
 
-cEnderCrystal::cEnderCrystal(Vector3d a_Pos):
-	Super(etEnderCrystal, a_Pos, 1.0, 1.0)
+cEnderCrystal::cEnderCrystal(Vector3d a_Pos, bool a_ShowBottom):
+	Super(etEnderCrystal, a_Pos, 1.0, 1.0),
+	m_ShowBottom(a_ShowBottom)
 {
 	SetMaxHealth(5);
 }
@@ -33,6 +34,10 @@ void cEnderCrystal::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	UNUSED(a_Dt);
 	// No further processing (physics e.t.c.) is needed
+	if (m_World->GetDimension() == dimEnd)
+	{
+		m_World->SetBlock({POSX_TOINT, POSY_TOINT + 1, POSZ_TOINT}, E_BLOCK_FIRE, 0);
+	}
 }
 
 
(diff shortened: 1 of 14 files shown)


src/Entities/EnderCrystal.cpp
```diff
@@ -33,6 +34,10 @@ void cEnderCrystal::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 {
 	UNUSED(a_Dt);
 	// No further processing (physics e.t.c.) is needed
+	if (m_World->GetDimension() == dimEnd)
+	{
+		m_World->SetBlock({POSX_TOINT, POSY_TOINT + 1, POSZ_TOINT}, E_BLOCK_FIRE, 0);
```
tigerw, line 39: Should we check the block already there == fire? Not sure if it's possible for pistons to push stuff into POSY_TOINT + 1 or to spawn a crystal where +1 is occupied? Otherwise it would be possible for crystals to act like an arbitrary block deleter.

Also POSITION_TOINT.addedY(1) should do it.

src/Entities/EnderCrystal.cpp
```diff
@@ -47,8 +52,7 @@ void cEnderCrystal::KilledBy(TakeDamageInfo & a_TDI)
 
 	Destroy();
 
-	m_World->SetBlock(POSX_TOINT, POSY_TOINT,     POSZ_TOINT, E_BLOCK_BEDROCK, 0);
-	m_World->SetBlock(POSX_TOINT, POSY_TOINT + 1, POSZ_TOINT, E_BLOCK_FIRE,    0);
+	m_World->SetBlock(POSX_TOINT, POSY_TOINT,     POSZ_TOINT, E_BLOCK_FIRE,    0);
```
tigerw, line 55: no need for the alignment spaces anymore :p

src/Entities/EnderCrystal.h
```diff
@@ -18,10 +18,27 @@ class cEnderCrystal :
...
 	CLASS_PROTODEF(cEnderCrystal)
 
-	cEnderCrystal(Vector3d a_Pos);
+	cEnderCrystal(Vector3d a_Pos, bool a_ShowBottom);
+
+	// Getters and Setters
+	bool ShowsBottom() const { return m_ShowBottom; }
+	void SetShowBottom(bool a_ShowBottom) { m_ShowBottom = a_ShowBottom; }
+
+	const Vector3i & GetBeamTarget() const { return m_BeamTarget; }
```
tigerw, line 27: returning Vector3i by value should be okay here

src/Entities/EnderCrystal.cpp
12xx12: The wiki states that any block is replaces with fire

https://minecraft.gamepedia.com/End_Crystal?mobileaction=toggle_view_desktop#Properties

tigerw: interesting, thanks.

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -1718,11 +1718,42 @@ void cWSSAnvil::LoadBoatFromNBT(cEntityList & a_Entities, const cParsedNBT & a_N
...
+			if (CoordinateLine > 0)
+			{
+				BeamTarget.z = a_NBT.GetInt(CoordinateLine);
+			}
+		}
+	}
+	CurrentLine = a_NBT.FindChildByName(a_TagIdx, "ShowBottom");
+	if (CurrentLine > 0)
+	{
+		EnderCrystal->SetShowBottom(a_NBT.GetByte(CurrentLine) == 0);
```
tigerw, line 1754: Should this be == 1?

src/Entities/EnderCrystal.h
```diff
@@ -18,10 +18,27 @@ class cEnderCrystal :
...
 
-	cEnderCrystal(Vector3d a_Pos);
+	cEnderCrystal(Vector3d a_Pos, bool a_ShowBottom);
+
+	// Getters and Setters
+	bool ShowsBottom() const { return m_ShowBottom; }
+	void SetShowBottom(bool a_ShowBottom) { m_ShowBottom = a_ShowBottom; }
+
+	const Vector3i GetBeamTarget() const { return m_BeamTarget; }
+	void SetBeamTarget(const Vector3i a_BeamTarget) { m_BeamTarget = a_BeamTarget; }
```
tigerw, line 28: no need for const Vector3i here

12xx12:
I maybe should reveal the 'cEnderCrystal' to the lua API while at it

tigerw (approved):

PR #5011, opened 2020-10-29
Added new flowers on bonemeal use

### LINKED ISSUE #1074: Implement Flowers and biome-dependent grass bonemealing
http://minecraft.gamepedia.com/Flowers

12xx12:
fixes #1074

### DIFF
--- a/src/Items/ItemDye.h
+++ b/src/Items/ItemDye.h
@@ -5,7 +5,6 @@
 #include "../World.h"
 #include "../Entities/Player.h"
 #include "../Blocks/BlockCocoaPod.h"
-#include <optional>
 
 
 
@@ -101,7 +100,7 @@ public:
 	Returns true if the plant was fertilized successfully, false if not / not a plant.
 	Note that successful fertilization doesn't mean successful growth - for blocks that have only a chance to grow,
 	fertilization success is reported even in the case when the chance fails (bonemeal still needs to be consumed). */
-	bool FertilizePlant(cWorld & a_World, Vector3i a_BlockPos)
+	static bool FertilizePlant(cWorld & a_World, Vector3i a_BlockPos)
 	{
 		BLOCKTYPE blockType;
 		NIBBLETYPE blockMeta;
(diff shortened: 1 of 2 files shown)


src/Items/ItemDye.h
```diff
@@ -216,51 +216,139 @@ class cItemDyeHandler :
...
-		- 20 % big grass (2-block tall grass)
-		- 60 % tall grass (1-block tall grass)
-		- 20 % flowers (biome dependent variants)
+		- 0 up to 8  big grass  (2-block tall grass)
+		- 8 up tp 24 tall grass (1-block tall grass)
+		- 0 up to 8  flowers    (biome dependent variants)
 	The new plants are spawned within 7 taxicab distance of a_BlockPos, on a grass block.
 	Broadcasts a particle for each new spawned plant. */
-	void growPlantsAround(cWorld & a_World, Vector3i a_BlockPos)
+	static void GrowPlantsAround(cWorld & a_World, Vector3i a_BlockPos)
```
tigerw, line 224: I'd recommend a construction like:

```cpp
static std::optional<Vector3i> GetRandomPlantPosition(const Vector3i BlockPosition, const bool IsTallGrass)
{
  const int OffsetY = ...;
  if (!ValidHeight)
  {
    return {};
  }

  const Vector3i Offset(X, Y, Z);
  if (
    (GetBlock(BlockPosition + Offset) != GRASS) ||
    (....)
  )
  {
    return {};
  }

  return Offset;
}
```

looks neater than a lambda I think.

tigerw:
seen, thanks. I am so behind on reviews this week...

tigerw:
It seems like these changes spawn much fewer plants than before. This might be due to the code running fewer random position checks; before we were doing 40 guaranteed but now it's 3 random numbers added together with an average of 24 or something. I think we should rework this so it does a random walk (looking for grass blocks) from the clicked block, each time it needs to spawn a plant. This will be more like vanilla behaviour of spawning from the centre out, and up the chances of finding a suitable grass to spawn plants on, compared to picking random coordinates which tend to be in the air.

12xx12:
So we don't Grant all 40 plants always but we grant all the plants the random generator decided to grant in this bonemeal use?

tigerw:
At least we try to. Unfortunately the wiki isn't very clear on this, we might have to invent something that looks reasonable.

12xx12:
I'm gonna do some thinking tomorrow and come back

tigerw:
I made an attempt, could you check if it seems sane?

src/Items/ItemDye.h
```diff
@@ -216,51 +216,185 @@ class cItemDyeHandler :
...
+			)
+		)
+		{
+			if (
+				!cChunkDef::IsValidHeight(Position.y) ||
+				(a_World.GetBlock(Position) != E_BLOCK_GRASS)  // Are we looking at grass?
+			)
+			{
+				// Not grass or invalid height, restart random walk and bail:
+				Position = a_Position;
```
12xx12, line 374: Maybe also reset the the Tries variable so that we completely restart the tun

12xx12: Although that might result in infinite loops when there's actually no space for a plant

tigerw (approved):

src/Items/ItemDye.h
tigerw: Yes, Tries acts a sort of a "fuel limit" for the loop as well as a walk distance limit

PR #5014, opened 2020-10-30
some const and inheritance correctness

12xx12:
yeah - me IDE flashed this at me and I fixed it

i thought about enforcing more const correcness in more classes - i don't know if it's neccesary

### DIFF
--- a/src/Blocks/WorldInterface.h
+++ b/src/Blocks/WorldInterface.h
@@ -61,7 +61,7 @@ public:
 	virtual void SendBlockTo(int a_BlockX, int a_BlockY, int a_BlockZ, cPlayer & a_Player) = 0;
 
 	/** Sends the block on those coords to the player */
-	inline void SendBlockTo(const Vector3i a_BlockPos, cPlayer & a_Player)
+	virtual inline void SendBlockTo(const Vector3i a_BlockPos, cPlayer & a_Player)
 	{
 		SendBlockTo(a_BlockPos.x, a_BlockPos.y, a_BlockPos.z, a_Player);
 	}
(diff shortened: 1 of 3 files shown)


src/Blocks/WorldInterface.h
```diff
@@ -61,7 +61,7 @@ class cWorldInterface
 	virtual void SendBlockTo(int a_BlockX, int a_BlockY, int a_BlockZ, cPlayer & a_Player) = 0;
 
 	/** Sends the block on those coords to the player */
-	inline void SendBlockTo(const Vector3i a_BlockPos, cPlayer & a_Player)
+	virtual inline void SendBlockTo(const Vector3i a_BlockPos, cPlayer & a_Player)
```
peterbell10, line 64: Nah, only one overload needs to be virtual.

peterbell10 (changes requested):

src/Blocks/WorldInterface.h
12xx12: Waht do you meen? Whe I remove the virtual expression we're back to square one and can't explicitly overide the function

tigerw: In fact I think you can delete the body of this altogether and make it pure virtual

peterbell10: `SendBlockTo(int, int, int)` can and should do the exact same thing as `SendBlockTo(Vector3i)`, so there's no reason for this to be virtual.

12xx12: But the method in cWorld shadows this method in cWorldInterface. This was the reason for this PR

tigerw:
peterbell’s point perhaps is that we should delete the overridden method in cWorld and keep the implementation in cWorldInterface, not marked as virtual, acting purely as a forwarder?

Sent from my Windows 10 Mobile ??

tigerw (approved):

PR #5013, opened 2020-10-30
Fixing ender crystal

12xx12:
fixed explosion
fixed ticking
fixed network
reavealed everything to lua

PR #5017, opened 2020-10-31
Ender Crytal Fix and report proper cmake file for luabindingscheck fail

12xx12:
#5013 is stopping on nether generation again

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -9553,7 +9553,7 @@ a_Player:OpenWindow(Window);
 						{
 							Type = "boolean",
 						},
-						Notes = "Returns true if the EnderCrystal displays its beam.",
+						Notes = "Returns true if the ender crystal displays its beam.",
 					},
 				},
 				GetBeamTarget =
@@ -9575,7 +9575,7 @@ a_Player:OpenWindow(Window);
 							Type = "Vector3i",
 						},
 					},
-					Notes = "Sets the target of the beam of this Ender Crystal.",
+					Notes = "Sets the target of the beam of this ender crystal.",
 				},
 				SetDisplayBeam =
 				{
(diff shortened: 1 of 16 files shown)


Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9539,6 +9539,78 @@ a_Player:OpenWindow(Window);
...
+					Returns =
+					{
+						{
+							Type = "boolean",
+						},
+						Notes = "Returns true if the bedrock base is displayed.",
+					},
+				},
+			},
+		},
```
Seadragon91, line 9613: The class cEnderCrystal inherits from cEntity in c++-class. This required to have the entity functions in the api website: `Inherits = "cEntity",`

12xx12: i'm confused by the error message generated by the checkstyle check. Any other class added in the ```src/Bindings/AllToLua.pkg``` are not in the ```CMakeList.txt``` file

tigerw: The error message is outdated, try adding to CMake\GenerateBindings.cmake (by the way could you please update the message 😅)

tigerw:
Hey 12xx12, I've prepared two small changes that I want to add;

1. Change order of members in cEnderCrystal (put all the bools after) for a small sizeof reduction (360B down to 352B)
2. Make cWorld::SpawnEnderCrystal only take a position and ShowBase, as you had before, just for easier usage for plugins. The reasoning being that the beam target is likely to constantly change, so plugins should favour the setter instead.

Does this sound ok to you?

12xx12:
Hi @tigerw 

1. yeah that sounds really good
2. I'm not quite sure. I'd rather add a new constructor for the API because on loading from disk the values are set and the setter methods invoke a network action which requires the world member not to be a nullptr. I did add a check if the object is a nullptr but you proposed to have a constructor which set's all values and not use the setter methods

12xx12:
Oh - do we have to update the plug-in with a or too? If yes I would do this also in this PR

tigerw:
For 2. I added a 2nd constructor for normal spawning with a simplified param list. Which plugin do you mean?

12xx12:
The Core Plugin with the /summon command

tigerw:
Good point, it would be nice to get the submodule updated all in one PR, but it's up to you.

tigerw (approved):

12xx12:
I'm gonna do a submodule update on another PR

PR #5023, opened 2020-11-03
Players are no longer kicked when clicking on bedrock

### LINKED ISSUE #5022: Disabling checking breaking a bedrock is possible?
Disabling checking breaking a bedrock is possible?
Or write the simple cfg rule(s)? 
https://github.com/cuberite/cuberite/blob/3381c0f6d6671a485283c889b036c0a431e2a2b9/src/ClientHandle.cpp#L1359

tigerw:
* Fixes #5022

### DIFF
--- a/src/BlockInfo.cpp
+++ b/src/BlockInfo.cpp
@@ -838,8 +838,13 @@ bool cBlockInfo::FullyOccupiesVoxel(const BLOCKTYPE Block)
 
 bool cBlockInfo::IsOneHitDig(const BLOCKTYPE Block)
 {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wfloat-equal"
+
 	// One hit break blocks:
 	return GetHardness(Block) == 0;
+
+#pragma clang diagnostic pop
 }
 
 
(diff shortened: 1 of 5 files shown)


src/BlockInfo.cpp
```diff
@@ -838,43 +838,13 @@ bool cBlockInfo::FullyOccupiesVoxel(const BLOCKTYPE Block)
...
-		case E_BLOCK_SAPLING:
-		case E_BLOCK_SLIME_BLOCK:
-		case E_BLOCK_TNT:
-		case E_BLOCK_TALL_GRASS:
-		case E_BLOCK_TORCH:
-		case E_BLOCK_TRIPWIRE_HOOK:
-		case E_BLOCK_TRIPWIRE: return true;
-		default: return false;
-	}
+	return GetHardness(Block) == 0;
```
12xx12, line 845: This can be simplified
```suggestion
	return !GetHardness(Block);
```

see http://cpp.sh/4w4sm3

12xx12: although I didn't check if this also triggers clang. I'm thinking about a simple idea to get rid of the clang pacifier

tigerw: It triggers another warning (conversion float to bool).

I think the warning is bogus here; we're able to compare to exactly floating zero because GetHardness returns exactly zero.

12xx12: Yeah - you are right you actually set it to 0. Maybe leave a comment?

tigerw: OK, fixed.

bennasar99:
Welcome to Google!

Hi

thanks for join us!

Reg Google

12xx12 (approved):

PR #5025, opened 2020-11-05
Adding Boss bar

### LINKED ISSUE #4915: Add boss health bar

The ender dragon and the Wither have a health bar popping up on the players hud. this is also modifyable in the comman /bossbar

https://minecraft.gamepedia.com/Commands/bossbar

I didnt find any documentation on the bar itself sadly

12xx12:
Adds boss bar

Fixes #4915

Todo:

- [x] let cRoot know of boss bars and let it handle the lookup by id

### DIFF
--- a/src/Broadcaster.cpp
+++ b/src/Broadcaster.cpp
@@ -171,6 +171,19 @@ void cWorld::BroadcastBlockEntity(Vector3i a_BlockPos, const cClientHandle * a_E
 
 
 
+void cWorld::BroadcastBossBarUpdateHealth(UInt32 a_UniqueID, float a_FractionFilled)
+{
+	ForClientsInWorld(*this, nullptr, [&](cClientHandle & a_Client)
+		{
+			a_Client.SendBossBarUpdateHealth(a_UniqueID, a_FractionFilled);
+		}
+	);
+}
+
+
+
+
+
 void cWorld::BroadcastChat(const AString & a_Message, const cClientHandle * a_Exclude, eMessageType a_ChatPrefix)
 {
 	ForClientsInWorld(*this, a_Exclude, [&](cClientHandle & a_Client)
@@ -444,6 +457,19 @@ void cWorld::BroadcastPlayerListRemovePlayer(const cPlayer & a_Player, const cCl
 
 
 
+void cWorld::BroadcastPlayerListUpdateDisplayName(const cPlayer & a_Player, const AString & a_CustomName, const cClientHandle * a_Exclude)
+{
+	ForClientsInWorld(*this, a_Exclude, [&](cClientHandle & a_Client)
+		{
+			a_Client.SendPlayerListUpdateDisplayName(a_Player, a_CustomName);
+		}
+	);
+}
+
+
+
+
+
 void cWorld::BroadcastPlayerListUpdateGameMode(const cPlayer & a_Player, const cClientHandle * a_Exclude)
 {
 	ForClientsInWorld(*this, a_Exclude, [&](cClientHandle & a_Client)
@@ -462,19 +488,6 @@ void cWorld::BroadcastPlayerListUpdatePing(const cPlayer & a_Player, const cClie
 	ForClientsInWorld(*this, a_Exclude, [&](cClientHandle & a_Client)
 		{
 			a_Client.SendPlayerListUpdatePing(a_Player);
-		}
-	);
-}
-
-
-
-
-
-void cWorld::BroadcastPlayerListUpdateDisplayName(const cPlayer & a_Player, const AString & a_CustomName, const cClientHandle * a_Exclude)
-{
-	ForClientsInWorld(*this, a_Exclude, [&](cClientHandle & a_Client)
-		{
-			a_Client.SendPlayerListUpdateDisplayName(a_Player, a_CustomName);
 		}
 	);
 }
--- a/src/Broadcaster.cpp
+++ b/src/Broadcaster.cpp
@@ -171,9 +171,9 @@ void cWorld::BroadcastBlockEntity(Vector3i a_BlockPos, const cClientHandle * a_E
 
 
 
-void cWorld::BroadcastBossBarUpdateHealth(UInt32 a_UniqueID, float a_FractionFilled)
+void cWorld::BroadcastBossBarUpdateHealth(const cEntity & a_Entity, UInt32 a_UniqueID, float a_FractionFilled)
 {
-	ForClientsInWorld(*this, nullptr, [&](cClientHandle & a_Client)
+	ForClientsWithEntity(a_Entity, *this, nullptr, [&](cClientHandle & a_Client)
 		{
 			a_Client.SendBossBarUpdateHealth(a_UniqueID, a_FractionFilled);
 		}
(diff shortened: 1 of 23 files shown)


12xx12:
There is a bug ATM if the entity dies the destructor of cCompositeChat causes the server to crash with a memory access violation

12xx12:
What is actually done when travis checks the API?

tigerw:
I thought boss bars were per-world instead of per-server. If this is so cWorld should track them

src/Protocol/Protocol_1_9.cpp
```diff
@@ -104,6 +104,78 @@ void cProtocol_1_9_0::SendAttachEntity(const cEntity & a_Entity, const cEntity &
...
+void cProtocol_1_9_0::SendBossBar(const cBossBar & a_BossBar, cBossBar::eOperation a_Operation)
+{
+	ASSERT(m_State == 3);  // In game mode?
+	cPacketizer Pkt(*this, pktBossBar);
+	// TODO: Bad way to write a UUID, and it's not a true UUID, but this is functional for now.
+	Pkt.WriteBEUInt64(0);
+	Pkt.WriteBEUInt64(a_BossBar.GetUniqueID());
+	Pkt.WriteVarInt32(static_cast<unsigned int>(a_Operation));
+	LOG(a_BossBar.GetTitle().CreateJsonString());
+	LOG(a_BossBar.GetJsonTitle());
```
tigerw, line 116: 🪵

tigerw:
How about a BossBarManager class stored in cWorld (like cMapManager) which maintains the owning `unordered_map` of cMonster * -> BossBar? It gets a Tick() every tick which it uses to update the bars.

Bosses inform the manager when the bar needs to be created inside their OnAddToWorld().
Likewise inform the manager to remove the bar inside their OnRemoveFromWorld().

This way we decouple cMonster from having to know about the boss bar.

src/Protocol/Protocol_1_9.cpp
tigerw: this emoji is apparently too new, it should be of a wooden log :p

12xx12: I see it :o 

Yea I had some fun with composite chat crashing the server

tigerw:
Another idea. Given bossbars are a UI element that is stateless, we don't need any BossBarManager or BossBar classes. Instead, just a `BroadcastBossBar(variant<Add, Remove, UpdateHealth, ...>)` function that is effectively a wrapper over the packet, and we export cClientHandle::SendBossBar to Lua. That gives maximum flexibility to plugins and the internal server, and will also reduce relatively expensive Lua <-> C++ binding transitions, plus is probably easier to implement.

tigerw:
Firstly I think it'd be good to move the cCompositeChat changes to a separate pull

12xx12:
I thought a little:

I think we should have both solutions. A `BroadcastBossBar` command useable by Lua (where you keep track of the ID and so on yourself). I am not quite shure too this for monsterns is the right thing. we would need to store if there is a boss bar, the settings, the id... for every monster reguardless if there is one. 

So maybe have a unique pointer to a struct containing every needed information and have a check in the main update loop and this checks if there is a boss bar and broadcasts it.

The reason the 'cBossBar' has so many member functions is that the packet supports all kinds of different changes which are marked differently. So the cBossBar is the wrapper for the packet.

tigerw:
I have something, will you mind if I push it here, or would you prefer I open another PR?

12xx12:
Hi ahead - push 👍🏻

tigerw:
before: a49d45473385553671daa52bdd8ee9bc877ea854

src/Broadcaster.cpp
```diff
@@ -171,6 +171,19 @@ void cWorld::BroadcastBlockEntity(Vector3i a_BlockPos, const cClientHandle * a_E
 
 
 
+void cWorld::BroadcastBossBarUpdateHealth(UInt32 a_UniqueID, float a_FractionFilled)
+{
+	ForClientsInWorld(*this, nullptr, [&](cClientHandle & a_Client)
```
12xx12, line 176: I don't think we want to send the bar to everybody

src/Protocol/Protocol_1_8.cpp
```diff
@@ -356,9 +383,22 @@ void cProtocol_1_8_0::SendDestroyEntity(const cEntity & a_Entity)
...
+
+	if (!a_Entity.IsMob())
+	{
+		return;
+	}
+
+	const auto & Mob = static_cast<const cMonster &>(a_Entity);
+	if ((Mob.GetMobType() == mtEnderDragon) || (Mob.GetMobType() == mtWither))
+	{
+		SendBossBarRemove(Mob.GetUniqueID());
```
12xx12, line 400: Any other BossBar has to be added manually so the Plugin is respnsible for removal 👍

12xx12 (commented):
This works too. 

Why did you change some PlayerList Code?

tigerw:
To alpha sort, I couldn't help myself...

src/Broadcaster.cpp
12xx12: Just with the loaded chunks or smth - like before

src/Mobs/EnderDragon.cpp
```diff
@@ -2,6 +2,8 @@
 #include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
 
 #include "EnderDragon.h"
+#include "ClientHandle.h"
+#include "CompositeChat.h"
```
12xx12, line 6: ../

It's annoying :(

src/Mobs/Wither.cpp
```diff
@@ -5,6 +5,8 @@
 
 #include "../World.h"
 #include "../Entities/Player.h"
+#include "ClientHandle.h"
+#include "CompositeChat.h"
```
12xx12, line 10: ../

12xx12:
One more thing:

What happens when I want to change the color of a boss bar during existance?

tigerw:
True, I ran out of time implementing it so just did the stuff the core server cares about. Would you like to see all the relevant protocol functions exported in this PR?

12xx12:
If we do it now we can do it properly 🤷🏻‍♂️

12xx12:
At some point this was there. I'll push that code

src/Broadcaster.cpp
12xx12: Oh... Hm... This is more complicated. We don't habe a entity to attach to so no position to determine if it should be sent

12xx12: We could do a optional parameter with coordinates that is hidden from the API so that we can send it just for some chunks.

It seems like the vanilla implementation displays the non entity bound bar to everyone

tigerw: If it's only being called by cWither & cEnderDragon adding a cEntity & parameter should work.

12xx12: this also applies for all other new Broadcast Methods. So there needs to be a solution for all methods

tigerw: We only added one (server internal) broadcast

tigerw:
how's this?

12xx12:
👍

PR #5050, opened 2020-11-17
Note block Left Click Fix

ProjectBM:
* Player can now left click on note block and play back the sound with an animation.

Fixes #4839

PR #5052, opened 2020-11-17
Note Block Left Click Fix V2

### LINKED ISSUE #4839: No noteblock playback on left click
Client version: any
Server OS: any
Cuberite Commit id: d2de029c024309b0ac2429373cc8ef2e56f880e0

### Expected behavior
When left clicking a note block in survival you play back the sound without changing the pitch

### Actual behavior
no sound playback or a animation

- place noteblock
- go into survival
- left click on it

ProjectBM:
Redo of PR #5050 since Travis CI build timed out.

* Player can now left click on note block and play back the sound with an animation.

Fixes #4839


12xx12:
@bearbin this is strange - we have a build breaking before executing the server not while generating the nether

tigerw:
I'm inclined to blame Travis

12xx12:
There error is reproduceable. I just restarted the ARM build

ProjectBM:
Strange it still fails 🤔

tigerw:
Quite irregular. @ProjectBM to aid debugging, you could try adding an `echo "Starting server"` just before the `./Cuberite` in travisbuild.sh, and in src\main.cpp, replace `return UniversalMain(argc, argv, false);` at the end with `std::cerr << "test" << std::endl; return 0;` as a sanity check.

ProjectBM:
Did as you said and ran travisbuild.sh but it seems to fail. Strangely enough it failed while building. I usually compile with `cmake . -DCMAKE_BUILD_TYPE=DEBUG && make` so it might be due to the difference in build type.

I ran it with the changes you mentioned and without it and travisbuild.sh still seem to fail. I also tried it on my master branch and the script still failed.

I opened another PR and that branch passes Clang 6.0 - Debug so I can't tell if this is an issue on my end or if Travis is just breaking.

12xx12:
Travis seems to break on this build. I had some PRs failing in the ARM build too. We had someone with a ARM processor recrreate the steps... that didn't fail - so best add those debug outputs as tigerw suggested

ProjectBM:
For clarification, should I make these changes and make a commit so Travis CI reruns with those changes or should I be running it locally?

12xx12:
make these changes and commit - we need to sanity check the travis machnine

12xx12:
maybe we should not not return the universal main to compile everything

ProjectBM:
Seems that the change caused a compile error. Should I change `static int UniversalMain` to `static inline int UniversalMain` and try Travis again?
```
In file included from /home/travis/build/cuberite/cuberite/CMakeFiles/Cuberite.dir/Unity/unity_8_cxx.cxx:4:
/home/travis/build/cuberite/cuberite/src/main.cpp:211:12: error: 'static' function 'UniversalMain' declared in header file should be declared 'static inline' [-Werror,-Wunneeded-internal-declaration]
static int UniversalMain(int argc, char * argv[], bool RunningAsService)

12xx12:
change the return statement back the way it was.

`/root/cuberite/src/main.cpp:211:12: error: function 'UniversalMain' is not needed and will not be emitted [clang-diagnostic-unneeded-internal-declaration]
static int UniversalMain(int argc, char * argv[], bool RunningAsService)`

12xx12:
hm... you removed the "test" to std err.. it seems like the executeable isn't launching

ProjectBM:
Uhhh woops, let me add back in

12xx12:
Hm... so there is a process running but not output 🤔

12xx12:
Is there anything else logged with the stdio or is everything else logged with the log macros?

ProjectBM:
I think everything is logged by macros since the only outputs using stdio are exceptions/failures.

I could be wrong though since I'm not too familiar with the infrastructure for outputs

12xx12:
@cuberite/core-team

tigerw:
Master also had a ARM failure: https://travis-ci.org/github/cuberite/cuberite/builds/745295445
But now it's ok again: https://travis-ci.org/github/cuberite/cuberite/builds/745929845

May be an intermittent issue, we'll see. Nonetheless at the time of writing the build's been going 1hr+ with no sign of x64 workers starting, so I wouldn't discount CI just yet.

12xx12:
#5056  worked fine

tigerw:
@ProjectBM would you like to revert the testing changes? We can give CI another shot.

tigerw:
#5057 passes

Issue #4940, opened 2020-09-26
No Clickthrough When Placing Water With Head In Trapdoor


KingCol13:
Client version: 1.12.2
Server OS: Linux Raspbian
Cuberite Commit id: e82644eebd397f29220abf1b1fd172dcf8f85bb4

### Expected behavior
When placing water while head is in a trapdoor, the trapdoor should be clicked-through and the water placed where the player is looking.

### Actual behavior
Trapdoor is deleted and replaced with water.

### Steps to reproduce the behavior
Attempt to place water like this:
![CuberiteTrapdoorWaterClickthrough](https://user-images.githubusercontent.com/48412633/94351777-42ed4c00-0055-11eb-99fe-927f37d6b8ef.png)

KingCol13:
Seems to be same as #2923

PR #5053, opened 2020-11-18
[Draft] fix the water above the player when on transparent block

theophriene:
In an attempt to fix: #4940

Any criticism is endorsed and welcomed. )

Edit: On a side note, maybe there should be a global const for maximum clickable distance, instead of these 5s all over the place?


theophriene:
And, it seems the bug is still there, except one block lower (( 
Diggin deeper...

PR #5056, opened 2020-11-22
Fix lilypad displacing block above

### LINKED ISSUE #4254: Lilypads displace blocks above them
Client version:   1.12.2
Server OS:       Linux (CentOS7)
Commit id:        06eb5cc70da76f6a95afa9c9a3d692ca39e39a76

### Expected behavior
Lilypad cannot be placed if a block occupies the space above the water upon which it would reside

### Actual behavior
Lilypad is allowed to be placed upon the water and the block above is converted to an item

(1) Place a water source block in the ground and break a single block next to it
(2) place a block over the water source block
(3) attempt to place a lilypad from the empty space next to the source block that was broken in step 1 (required to get an angle on the source block)

### LINKED ISSUE #4255: Lilypads reset water physics in water non-source blocks
Client version:   1.12.2
Server OS:        Linux (CentOS7)
Commit id:       06eb5cc70da76f6a95afa9c9a3d692ca39e39a76

### Expected behavior
Lilypads cannot be placed at all upon non-source flowing water; right clicking while a lilypad is equipped on flowing water is completely ignored

### Actual behavior
Lilypad momentarily appears on block under flowing water, but is immediately displaced by water attempting to flow onto that block and becomes an item

(1) Place a water source block such that blocks of flowing, non-source water are created
(2) attempt to place a lilypad upon a block of non-source flowing water such that the viewing angle also intersects with a block underneath the water

ProjectBM:
`cLineBlockTracer` seems to trace diagonally. This means at a certain angle, the player can interact with a water source block even if there is a block above it covering the player's vision. This displaces the block above.

Added a check that only allows air blocks to be replaced when using `cLineBlockTracer`.

This displacement also happens when a player right clicks the side of a block to try to place a lilypad.

Added a similar check that only allows air blocks and water blocks to be replaced. I included water blocks too because you can technically right click on the side of a block and replace a water block in vanilla. 

While fixing the above mentioned issues, I also had to add a check that only allows lilypads to be placed when there is a water block below it (just happened to fix a related issue).

Fixes #4254 
Fixes #4255 (Happened to fix this while fixing 4254)

### DIFF
--- a/src/Items/ItemLilypad.h
+++ b/src/Items/ItemLilypad.h
@@ -93,14 +93,6 @@ public:
 						return false;
 					}
 					a_CBBlockPos = AddFaceDirection(a_CBBlockPos, BLOCK_FACE_YP);  // Always place pad at top of water block
-					if (
-						!IsBlockWater(a_CBBlockType) &&
-						cBlockInfo::FullyOccupiesVoxel(a_CBBlockType)
-						)
-					{
-						// Can't place lilypad on air / in another block!
-						return true;
-					}
 					m_HasHitFluid = true;
 					m_Pos = a_CBBlockPos;
 					return true;
--- a/src/Items/ItemLilypad.h
+++ b/src/Items/ItemLilypad.h
@@ -48,6 +48,24 @@ public:
 		{
 			// Clicked on a face of a submerged block; vanilla allows placement, so should we
 			auto PlacePos = AddFaceDirection(a_ClickedBlockPos, a_ClickedBlockFace);
+
+			// Lilypad should not replace non air and non water blocks
+			auto BlockToReplace = a_World->GetBlock(PlacePos);
+			if ((BlockToReplace != E_BLOCK_AIR) &&
+				(BlockToReplace != E_BLOCK_WATER) &&
+				(BlockToReplace != E_BLOCK_STATIONARY_WATER))
+			{
+				return false;
+			}
+
+			// Lilypad should be placed only if there is a water block below
+			auto BlockBelow = a_World->GetBlock(PlacePos - Vector3i(0, 1, 0));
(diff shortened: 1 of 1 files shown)


12xx12 (approved):

src/Items/ItemLilypad.h
```diff
@@ -100,6 +118,13 @@ class cItemLilypadHandler:
```
tigerw, line 118: Ah, this is kinda embarrassing on my part. The original commit for lilypads was 283a66bcae5f26d2b5038ba0959314a687c4d8b6#diff-c25a12152914ebf4cd50f2ac9cd08ce8631772948705300e963bc97d96c31b30. I then proceeded to introduce a regression in f763242e5c5d8e6469685c4be95d3dc264ed6a67#diff-c25a12152914ebf4cd50f2ac9cd08ce8631772948705300e963bc97d96c31b30 by deleting the GetBlock with the new coordinates. You can see the remains of that check on line 96 above. Could you move the check from below into the callback?

ProjectBM: To do this I would need to re-introduce the `cWorld * m_World` variable into the callback. Right now it doesn't seem like the callback has a way to run `GetBlock`.

Should I just re-add it?

tigerw: True, in that case we should delete the redundant check in the callback altogether.

ProjectBM: Gotcha, I just tested with the check in the callback and it seems to have some problems anyways
- If you're completely submerged in water then you can still place a lilypad due to `!IsBlockWater(Block)`
- If there's a half slab above the water (doesn't occupy a full voxel) then the half slab gets replaced due to `cBlockInfo::FullyOccupiesVoxel(Block)`

PR #5060, opened 2020-11-28
Fix Block Entity Placement in Generation

### LINKED ISSUE #5035: Desert temples spawn with invisible and unusable chests
Client version: 1.12.2

Server OS: Linux
Cuberite Commit id: a0afd132330ab538848d5ac263c021c1247fd688

### Expected behavior
Chests to appear and be usable

### Actual behavior
Screenshot_20201116_134733

They can't be opened either so I can't verify if the type of loot is correct.

Start a new world with seed 1492584284 and go to the desert temple located at X -48 and Z 71.

12xx12:
`cBlockArea` in `cChunkDesc` now handles block entities
some minor changes, typos...
Block entities validate and correct their position when put into the world.

Fixes #5035
testes with jungle temples


12xx12:
The underlying issue is that somewhere in the process of placing a prefab structure the block entities lose their position

PureTryOut:
Awesome! Note that the description of the PR should contain "Fixes #5035" (you placed the # at the wrong position) for Github to mark the issue as related and to close it automatically once this is merged.

12xx12:
Thank you for the reminder - yeah I was a bit frustrated when opening this PR 😄

tigerw:
Thanks!

src/BlockArea.h
```diff
@@ -362,7 +362,7 @@ class cBlockArea
 	bool HasBlockMetas    (void) const { return (m_BlockMetas    != nullptr); }
 	bool HasBlockLights   (void) const { return (m_BlockLight    != nullptr); }
 	bool HasBlockSkyLights(void) const { return (m_BlockSkyLight != nullptr); }
-	bool HasBlockEntities (void) const { return (m_BlockEntities != nullptr); }
+	bool HasBlockEntities (void) const { return m_BlockEntities.operator bool(); }
```
tigerw, line 365: :O what's the reason here?

src/Chunk.cpp
```diff
@@ -310,20 +310,26 @@ void cChunk::SetAllData(cSetChunkData & a_SetChunkData)
...
-		}  // for KeyPair - m_BlockEntities
-	#endif  // _DEBUG
+			ASSERT(cChunkDef::IndexToCoordinate(KeyPair.first) == KeyPair.second->GetRelPos());
+		#endif  // _DEBUG
 
-	// Set all block entities' World variable:
-	for (auto & KeyPair : m_BlockEntities)
-	{
+		// Reset Pointer
+		KeyPair.second->SetWorld(nullptr);
```
tigerw, line 326: Why set nullptr world?

src/Generating/ChunkDesc.h
```diff
@@ -220,7 +220,8 @@ class cChunkDesc
 	inline BlockNibbleBytes &        GetBlockMetasUncompressed(void) { return *(reinterpret_cast<BlockNibbleBytes *>(m_BlockArea.GetBlockMetas())); }
 	inline cChunkDef::HeightMap &    GetHeightMap             (void) { return m_HeightMap; }
 	inline cEntityList &             GetEntities              (void) { return m_Entities; }
-	inline cBlockEntities &          GetBlockEntities         (void) { return m_BlockEntities; }
+	inline const cBlockEntities &    GetBlockEntities         (void) const { return m_BlockArea.GetBlockEntities(); }
```
tigerw, line 223: What is the const version for?

src/Generating/ChunkDesc.cpp
```diff
@@ -587,21 +587,21 @@ cBlockEntity * cChunkDesc::GetBlockEntity(int a_RelX, int a_RelY, int a_RelZ)
...
-	auto be = cBlockEntity::CreateByBlockType(GetBlockType(a_RelX, a_RelY, a_RelZ), GetBlockMeta(a_RelX, a_RelY, a_RelZ), {AbsX, a_RelY, AbsZ});
-	if (be == nullptr)
+	auto BlockEntity = cBlockEntity::CreateByBlockType(GetBlockType(a_RelX, a_RelY, a_RelZ), GetBlockMeta(a_RelX, a_RelY, a_RelZ), {AbsX, a_RelY, AbsZ});
+	if (BlockEntity == nullptr)
 	{
 		// No block entity for this block type
 		return nullptr;
 	}
-	auto res = m_BlockEntities.emplace(Idx, std::move(be));
+	auto res = m_BlockArea.GetBlockEntities().emplace(Idx, std::move(BlockEntity));
```
tigerw, line 604: Does BlockArea not keep track of its internal blockentities list correctly?

src/Chunk.cpp
12xx12: For whatever reason in debug mode the nether is mega weird on startup and some block entities have already a set world. This is just for safety

src/Generating/ChunkDesc.h
12xx12: The comment said there should be read only access

src/Generating/ChunkDesc.cpp
12xx12: No. When using the prefab the positions are not set correctly.

The relative position is correct but the absolute is wrong. This only happens when using a prefab. This code is only a bad fix - I didn't find the underlying issue.

The problem is that a block entity is not informed by the world where it's actual position is. The broadcasting, loading, saving is done with the position stored in the block entity.

src/BlockArea.h
12xx12: You mean for the weird access to the book operator?

Normally you can use it simply. I don't know why the compiler denied it here

tigerw:
As you say there seems to be an underlying issue, I think this warrants further investigation (e.g. with ChunkDesc seemingly doing the work of BlockArea).

12xx12:
Yeah, I think that's a bug inside `cBlockArea` when merging into the world the position of the block entity is not put in the right chunk

bearbin (approved):

PureTryOut:
Thinga are happening again! 🎉
It's such a shame that this project moves so slowly, it has so much potential...

Issue #5007, opened 2020-10-24
Nether portal being unlit on other side


OwenJF:
Client version: 1.12.2
Server OS: Windows
Cuberite Commit id: 2231

Nether portal should be lit on other side.

Nether portal is not lit on other side

Make nether portal, light, go through, other side is not lit.

OwenJF:
Checking if has already been reported, may delete.

tigerw:
Thanks!

Looks like Nether Portal generation relied incorrectly on SetBlock executing block handlers:

https://github.com/cuberite/cuberite/blob/64442b05f8aa39918f5cb75503c7f57b0bd62827/src/NetherPortalScanner.cpp#L266-L267

Changing it to `m_World.PlaceBlock` should do the trick. Would you like to try? :)

cflep:
I'm gonna try (it's my second contribution ever)

PR #5062, opened 2020-11-29
Lit the nether portal when switching dimensions

cflep:
The nether portal is now lit after going to another dimension

Fixed Issue  #5007

### DIFF
--- a/src/NetherPortalScanner.cpp
+++ b/src/NetherPortalScanner.cpp
@@ -264,7 +264,7 @@ void cNetherPortalScanner::BuildNetherPortal(Vector3i a_Location, Direction a_Di
 	}
 
 	// Fill the frame (place a fire in the bottom)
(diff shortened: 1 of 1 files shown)


cflep:
I don't know what the issue ist with travis-ci, I just compiled it on my Raspberry Pi 4 (newest Raspberry Pi OS) and it worked without any issues. Same on my 64bit Ubuntu 20.04

PureTryOut:
You might want to consider naming PR's differently in the future. Normally you would name a PR to describe what it actually accomplishes or fixes. In this case I would call it something like "Lit the nether portal when switching dimensions" (or something along those lines). That way you can see what a PR does without actually opening it, and it makes it easier to search for it or filter it.

tigerw:
The ARM builds on CI have been flaky recently, we're not entirely sure why...

PR #5078, opened 2020-12-13
Only enter chase mode if Agressive Mob has los to player.

leeadama:
This fixes what imho is the most problematic issue with the game at the moment.  The fact that currently, Mobs have superman vision.  They see you through walls, doors, hiding in a tree, underground.  If there is a mob within it's detection range then it knows exactly where you are and it swarms you regardless.

This causes problems mainly for me because you can't sleep if an Aggressive Mob is around, and this issue causes them to gather around your house.  (Also, if you dare go outside they all jump on you at once.)

I simply added a line of sight check before chase mode is entered into.  I used pre-existing code I modified from elsewhere in the source.

I am not an experienced C++ developer, but I have done what I can and I've been testing it for a while on my Android Phone Server with no issues.  The game is MUCH more enjoyable to play with this one simple change IMHO.

Thanks for your consideration.
(the1robert on forum).

More details here : https://forum.cuberite.org/thread-3331.html

### DIFF
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -42,9 +42,19 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
 	{
 		return;
 	}
+	// Check to see if Mob can see the player.
+	Vector3d MyHeadPosition1 = GetPosition() + Vector3d(0, GetHeight(), 0);
+	Vector3d TargetPosition1 = a_Player->GetPosition() + Vector3d(0, a_Player->GetHeight(), 0);
+	if (cLineBlockTracer::LineOfSightTrace(*GetWorld(), MyHeadPosition1, TargetPosition1, cLineBlockTracer::losAirWaterLava))
+	{
+		Super::EventSeePlayer(a_Player, a_Chunk);
+		m_EMState = CHASING;
+	}
+	else
+	{
+		return;
+	}
 
-	Super::EventSeePlayer(a_Player, a_Chunk);
-	m_EMState = CHASING;
 }
 
 


tigerw:
Thanks! It looks like the code in Monster.cpp is passing a flag to explicitly ignore line of sight for some reason

https://github.com/cuberite/cuberite/blob/e35519ec8ab1b958408ab2a90b88dffc1bcc7fb2/src/Mobs/Monster.cpp#L767-L771

leeadama:
> 
> 
> Thanks! It looks like the code in Monster.cpp is passing a flag to explicitly ignore line of sight for some reason
> 
> https://github.com/cuberite/cuberite/blob/e35519ec8ab1b958408ab2a90b88dffc1bcc7fb2/src/Mobs/Monster.cpp#L767-L771

I'm sure it's my issue  but I don't understand.  I see the player reference and chunk reference being passed here, but that is all.   Also, my edit only applied to Aggressive Monsters, this file seems to affect more than that.  Chickens and Cows and things are all mentioned in this file (Monster.cpp).

What I am saying is that I don't understand the relationship between this bit of code and my fix.  

I would greatly appreciate it if you (or someone else) could explain it to me?

Thanks!

~
Robert

tigerw:
there's a `, false` hiding on line 771, I believe that turns off LOS tracing. Why it's there escapes me...

leeadama:
> there's a `, false` hiding on line 771, I believe that turns off LOS tracing. Why it's there escapes me...

I've done some preliminary testing and it appears at first glance that changing it to true does enable los checks.
I will do more testing over the next few days and update.

leeadama:
> there's a `, false` hiding on line 771, I believe that turns off LOS tracing. Why it's there escapes me...

I've done some preliminary testing and it appears at first glance that changing it to true does enable los checks.
I will do more testing over the next few days and update.

----------

Further testing confirms this.  It seems to work exactly the same as my "fix" did.  Closing this PR and I will make a new one with the correct way of enabling this.

leeadama:
Further testing confirms this.  It seems to work exactly the same as my "fix" did.  Closing this PR and I will make a new one with the correct way of enabling this.

PR #5079, opened 2020-12-13
Ignore CanFly flag sent by the client

mathiascode:
In Minecraft 1.16, the Player Abilities packet was modified to only include a flag specifying if the player is flying or not. This broke flying when connected to a Cuberite server through ViaVersion on a 1.16 client, as CanFly defaults to false in the code.

Since flying works just fine on a vanilla 1.12.2 server with ViaVersion, and only the IsFlying flag remains in 1.16, I think it's safe to assume that the vanilla server has been ignoring the CanFly flag for a while.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -835,12 +835,11 @@ void cClientHandle::HandleEnchantItem(UInt8 a_WindowID, UInt8 a_Enchantment)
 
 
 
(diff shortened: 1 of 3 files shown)


Pokechu22:
Per the [1.15.2 documentation](https://wiki.vg/index.php?title=Protocol&oldid=16067#Player_Abilities_.28serverbound.29):

> The vanilla client sends this packet when the player starts/stops flying with the Flags parameter changed accordingly. **All other parameters are ignored** by the vanilla server.

If the client was able to enable its own flight (or god mode or creative) at will, that'd be a bit exploitable :P

PR #5096, opened 2020-12-31
Fix AppVeyor IA32 builds

tigerw:
> The default target platform name (architecture) is that of the host and is provided in the CMAKE_VS_PLATFORM_NAME_DEFAULT variable.

AppVeyor uses AMD64. Oops.


bearbin:
Now the cache is cleared, should the cmakecache del itself be deleted?

bearbin (approved):

PR #5101, opened 2021-01-05
Convert most calls to blocking GetHeight/GetBiomeAt to direct chunk accesses

### LINKED ISSUE #5094: Internal Fault
Hey Guys, 
After successfully setting up my Minecraft 1.12.2 server on my raspberry pi 4 2gb i run into some problems.
The biggest Problem is, that the server keeps crashing in between every 7minutes to 4 hours.
The output says this:
cuberite has encountered an error and needs to close| sigabrt: server self-terminated due to an internal fault | | cuberite local build (id: unknown) | from commit approx: abc96bb4ce0bfb5627e74f4f2e66f19eaaa12736
We tried loading less Chunks, but the Situation didn't get any better.
I would be really happy, if someone could help me, because we are really enjoying playing minecraft on Cuberite.
Thanks

tigerw:
* Hopefully fixes #5094

### DIFF
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -115,12 +115,13 @@ private:
 	/** Returns true if there's either a water source block close enough to hydrate the specified position, or it's raining there. */
 	static bool IsWaterInNear(const cChunk & a_Chunk, const Vector3i a_RelPos)
 	{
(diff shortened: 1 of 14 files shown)


bearbin (approved):
Not sure about the overloading of IsWeatherWetAt, but otherwise fine.

PR #5111, opened 2021-01-17
Implementation of totem of undying behaviour

### LINKED ISSUE #5047: Implement totem of undying behaviour
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: a0afd132330ab538848d5ac263c021c1247fd688

### Expected behavior
When a player receives normally fatal damage while holding a totem of undying in their off-hand (the shield slot), they instead receive the damage and then instantly get half a heart restored, saving them from dying. This also works for any mob that holds it in their off-hand.

### Actual behavior
The player dies and the totem is unused

Put a totem of undying in the off-hand and die

dImrich:
Fixes #5047 

Tested on 1.12.2, works exactly like on notchian server as is explained here:
https://minecraft.gamepedia.com/Totem_of_Undying

### DIFF
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -106,52 +106,23 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 
 
 
-bool cPawn::DeductTotem(cPlayer * a_Player)
-{
-	auto & inv = a_Player->GetInventory();
-	if (inv.GetEquippedItem().m_ItemType == E_ITEM_TOTEM_OF_UNDYING)
-	{
-		inv.SetEquippedItem(cItem());
-		return true;
-	}
-	if (inv.GetShieldSlot().m_ItemType == E_ITEM_TOTEM_OF_UNDYING)
-	{
-		inv.SetShieldSlot(cItem());
-		return true;
-	}
-	return false;
-}
 
 
 void cPawn::KilledBy(TakeDamageInfo & a_TDI)
 {
 	ClearEntityEffects();
 
-	if (IsPlayer())
+	// Is death eligible for totem reanimation?
+	if (DeductTotem(a_TDI.DamageType))
 	{
-		auto Player = static_cast<cPlayer *>(this);
+		m_World->BroadcastEntityStatus(*this, esTotemOfUndying);
 
-		//  If the player is holding a totem of undying in their off-hand or
-		//  main-hand slot and receives otherwise fatal damage, the totem saves the player from death.
-		// Is death eligible for totem reanimation?
-		if (a_TDI.DamageType != dtAdmin && a_TDI.DamageType != dtInVoid && DeductTotem(Player))
-		{
+		AddEntityEffect(cEntityEffect::effAbsorption, 100, 1);
+		AddEntityEffect(cEntityEffect::effRegeneration, 900, 1);
+		AddEntityEffect(cEntityEffect::effFireResistance, 800, 0);
 
-			m_World->BroadcastEntityStatus(*this, esTotemOfUndying);
-
-			AddEntityEffect(cEntityEffect::effAbsorption, 100, 1);
-			AddEntityEffect(cEntityEffect::effRegeneration, 900, 1);
-			AddEntityEffect(cEntityEffect::effFireResistance, 800, 0);
-
-			m_Health = 1;
-
-			return;
-		}
-	}
-	else
-	{
-		// todo implement when mobs will be able to pick up items based on
-		// CanPickUpLoot attribute
+		m_Health = 1;
+		return;
 	}
 
 	Super::KilledBy(a_TDI);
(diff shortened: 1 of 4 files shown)


src/Entities/Pawn.cpp
```diff
@@ -111,6 +111,37 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+		cInventory & inv = Player->GetInventory();
+
+		//  If the player is holding a totem of undying in their off-hand or
+		//  main-hand slot and receives otherwise fatal damage, the totem saves the player from death.
+		if ((inv.GetShieldSlot().m_ItemType == E_ITEM_TOTEM_OF_UNDYING ||
+			 inv.GetEquippedItem().m_ItemType == E_ITEM_TOTEM_OF_UNDYING) &&
+				a_TDI.DamageType != dtAdmin && a_TDI.DamageType != dtInVoid
+			)
+		{
+			inv.SetShieldSlot(cItem());
```
tigerw, line 127: Won't this clear the shield slot, even if the totem was in the hotbar?

src/Entities/Pawn.cpp
```diff
@@ -111,6 +111,37 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+			m_World->BroadcastEntityStatus(*this, esTotemOfUndying);
+
+			AddEntityEffect(cEntityEffect::effAbsorption, 100, 1);
+			AddEntityEffect(cEntityEffect::effRegeneration, 900, 1);
+			AddEntityEffect(cEntityEffect::effFireResistance, 800, 0);
```
tigerw, line 132: Should all effects be cleared as well?

src/Entities/Pawn.cpp
dImrich: Sleep deprivation... I fixed it in a new commit.

src/Entities/Pawn.cpp
dImrich: They are already cleared on first line of cPawn::KilledBy.

src/Entities/Pawn.cpp
```diff
@@ -111,6 +111,44 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
+				a_TDI.DamageType != dtAdmin && a_TDI.DamageType != dtInVoid
+			)
+		{
+			if (inv.GetShieldSlot().m_ItemType == E_ITEM_TOTEM_OF_UNDYING) {
+				inv.SetShieldSlot(cItem());
+			}
+			else
+			{
+				inv.SetEquippedItem(cItem());
+			}
```
bearbin, line 133: It's a bit ugly checking the equipped item in the previous if-statement, then immediately checking it again, how about extracting this into a seperate function?

```suggestion
		// Is death eligible for totem reanimation?
		if (a_TDI.DamageType != dtAdmin && a_TDI.DamageType != dtInVoid && DeductTotem(Player))
		{
```

where DeductTotem is implemented similarly:
```cpp
bool DeductTotem(cPlayer * a_Player)
{
	auto & inv = a_Player->GetInventory();
	if (inv.GetShieldSlot().m_ItemType == E_ITEM_TOTEM_OF_UNDYING)
	{
		inv.SetShieldSlot(cItem());
		return true;
	}
	if (inv.GetEquippedItem().m_ItemType == E_ITEM_TOTEM_OF_UNDYING)
	{
		inv.SetEquippedItem(cItem());
		return true;
	}
	return false;
}
```

bearbin: Actually, there's another problem here too. The equipped totem should be removed before the shield slot totem - I've fixed my suggestion above but yours also has that issue.

dImrich: I made the change you suggested, it works but travis ci builds failed with 2 errors. Second error is absurd and first weird.
`/home/travis/build/cuberite/cuberite/src/Entities/Pawn.cpp:109:6: error: no previous prototype for function 'DeductTotem' [-Werror,-Wmissing-prototypes]
bool DeductTotem(cPlayer * a_Player)
     ^
/home/travis/build/cuberite/cuberite/src/Entities/Pawn.cpp:133:16: error: unused variable 'inv' [-Werror,-Wunused-variable]
                cInventory & inv = Player->GetInventory();
                             ^
2 errors generated.`

bearbin: The first error can be resolved by adding a declaration for DeductToken to Pawn.h or (preferably) making the function static. 

The second is simple, you just need to remove the definition of inv on line 133, as it is no longer used (now it's part of the function you moved the rest into).

bearbin:
One other thing, please would you add yourself to the CONTRIBUTORS file, so you are recognised as a contributor and also accept the licensing terms.

bearbin (approved):

src/Entities/Pawn.cpp
```diff
@@ -106,11 +106,54 @@ void cPawn::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
...
 
+bool cPawn::DeductTotem(cPlayer * a_Player)
+{
+	auto & inv = a_Player->GetInventory();
+	if (inv.GetEquippedItem().m_ItemType == E_ITEM_TOTEM_OF_UNDYING)
+	{
+		inv.SetEquippedItem(cItem());
+		return true;
+	}
+	if (inv.GetShieldSlot().m_ItemType == E_ITEM_TOTEM_OF_UNDYING)
```
PureTryOut, line 117: Wait, this means that if you hold a totem of undying in both your main hand _and_ the offhand, both will be used right?
I just tested it in vanilla and this is not correct behaviour, only one should be used (the one in the main hand).

So my suggestion would be making this an else if.

tigerw:
hm, circleci has taken their name too literally and gotten stuck; style checking isn't working

bearbin (approved):

PR #5113, opened 2021-01-19
New Pawn_Fell hook + micro optimizations in cPawn

KrystilizeNevaDies:
Implemented HOOK_PAWN_FELL for when a cPawn falls enough to get damaged.
Also removed a bit of unnecessary maths found in cPawn for optimization

### DIFF
--- a/Server/Plugins/APIDump/Hooks/OnPawnFell.lua
+++ b/Server/Plugins/APIDump/Hooks/OnPawnFell.lua
@@ -1,0 +1,25 @@
+return
+{
+	HOOK_PAWN_FELL =
+	{
+		CalledWhen = "A pawn has fallen onto a block",
+		DefaultFnName = "OnPawnFell",  -- also used as pagename
+		Desc = [[
+			This hook is called whenever a cPawn has fallen onto a block.
+		]],
+		Params =
+		{
+			{ Name = "Player", Type = "{{cPlayer}}", Notes = "The player from whom the packet was received" },
+			{ Name = "Animation", Type = "number", Notes = "The kind of animation" },
+		},
+		Returns = [[
+			If the function returns false or no value, the next plugin's callback is called. Afterwards, the
+			server broadcasts the animation packet to all nearby clients. If the function returns true, no other
+			callback is called for this event and the packet is not broadcasted.
+		]],
+	},  -- HOOK_PAWN_FELL
+}
+
+
+
+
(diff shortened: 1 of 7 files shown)


src/Entities/Pawn.cpp
```diff
@@ -404,15 +404,19 @@ void cPawn::HandleFalling(void)
...
+
 			// Fall particles
 			GetWorld()->BroadcastParticleEffect(
 				"blockdust",
 				GetPosition(),
 				{ 0, 0, 0 },
-				(Damage - 1.f) * ((0.3f - 0.1f) / (15.f - 1.f)) + 0.1f,  // Map damage (1 - 15) to particle speed (0.1 - 0.3)
-				static_cast<int>((Damage - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f),  // Map damage (1 - 15) to particle quantity (20 - 50)
+				(Damage - 1.f) * 0.01428571429 + 0.1f,  // Map damage (1 - 15) to particle speed (0.1 - 0.3) Full formula: ((Damage - 1.f) * ((0.3f - 0.1f) / (15.f - 1.f)) + 0.1f)
+				static_cast<int>((Damage - 1.f) * 2.142857143 + 20.f),  // Map damage (1 - 15) to particle quantity (20 - 50) Full formula: static_cast<int>((Damage - 1.f) * ((50.f - 20.f) / (15.f - 1.f)) + 20.f)
```
bearbin, line 419: This will be optimised out by the compiler, it's clearer how it was before.

Server/Plugins/APIDump/Hooks/OnPawnFell.lua
```diff
@@ -0,0 +1,25 @@
...
+	{
+		CalledWhen = "A pawn has fallen onto a block",
+		DefaultFnName = "OnPawnFell",  -- also used as pagename
+		Desc = [[
+			This hook is called whenever a cPawn has fallen onto a block.
+		]],
+		Params =
+		{
+			{ Name = "Player", Type = "{{cPlayer}}", Notes = "The player from whom the packet was received" },
+			{ Name = "Animation", Type = "number", Notes = "The kind of animation" },
```
bearbin, line 13: This doesn't match up with how the hook is actually called, with a `Damage` rather than an `Animation`.

tigerw:
Does the existing HOOK_TAKE_DAMAGE (damage type dtFalling) not suffice for your needs?

PR #5114, opened 2021-01-22
Fix debug macro situation

tigerw:
Use the standard NDEBUG.


src/Resources/Cuberite.rc
```diff
@@ -16,21 +16,23 @@ BEGIN
 	2 "A lightweight, fast and extensible game server for Minecraft"
 END
 
-#define VERSION 0,0,13,37
-#define VERSION_STRING "0.0.13.37"
+#define VERSION 1,3,3,7
+#define VERSION_STRING "1.3.3.7"
```
bearbin, line 20: Does this change have any affect? I'm guessing not, but it is a major version change so potentially this could mess up people's VS?

tigerw: It's just what shows up in the properties dialog in Explorer, and I don't think anyone/anything else is using the exe metadata.

Dialog

bearbin (approved):

PR #5119, opened 2021-01-31
CMake changes

tigerw:


tigerw:
- [ ] Upgrade ARM64 to Focal Fossa for Clang 7
- [ ] Fix the inline constexpr disabled warnings that were due to a Clang 6 bug

bearbin (approved):

12xx12:
`clang: error: the clang compiler does not support '-mcpu=native'`

PR #5121, opened 2021-02-04
Adds playerlist header and footer broadcasting

dImrich:
Notchian server does not implement this so I made it as simple as possible.
Maybe it would be even better to have it exported only in ClientHandle and let plugins handle the rest.


dImrich:
I tested it and it works on 1.8, 1.9, 1.9.4, 1.10.2, 1.11.2, 1.12.2

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -1584,6 +1584,26 @@ end
...
+								Type = "cCompositeChat",
+								Name = "a_Footer",
+							},
+						},
+						Returns =
+						{
+						},
+						Desc = "Sends the header and footer of the player list to the client."
+					},
+				},
```
tigerw, line 1606: Looks like an extra set of outer braces, and an extra Returns. Could you also delete the a_ from a_Header/a_Footer, and the same for the other API additions.

tigerw:
Thanks, looks great. We can finally have in-game server branding xD

tigerw:
Thanks!

dImrich:
It's my pleasure, thanks for a quick review and merge.

PR #5122, opened 2021-02-04
cPieceModifier interface and cPieceModifierRandomizeBlocks class

### LINKED ISSUE #4824: Villages should generate with all types of crops in different growth stages
Client version: 1.8.9
Server OS: Windows
Cuberite Commit id: (Found at the top of the server's console output)

### Expected behavior
Villages should generate with all types of crops in different growth stages

### Actual behavior
Villages are only generated with fully grown wheat plants

dImrich:
###  **cPieceModifierRandomizeBlocks** class
 **Params:**
  - One or multiple blocks to replace ( after first pipe ) _- (Required)_
    - Block type _- (Required)_
  - One or multiple blocks to randomize ( after second pipe ) _- (Required)_
    - Block type _- (Required)_
    - Weight _- (Required)_
    - Meta params _- (Optional)_
  - Meta params _- (Optional)_
    -  Maximum Meta _- (Required)_
    -  Minimum Meta _- (Required)_
    - Maximum in noise range  _- (Optional)_
    - Minimum in noise range  _- (Optional)_

Village plant bed prefab examples:
`["Modifiers"] = "{RandomizeBlocks|wheatblock|wheatblock,30[7];potatoes,30[2];carrots,80[5,7,5,9]}",`
`["Modifiers"] = "{RandomizeBlocks|wheatblock|wheatblock,50;potatoes,90;carrots,30[3,7,3,13]}",`
`["Modifiers"] = "{RandomizeBlocks|wheatblock,potatoes|wheatblock,20[1,7,1,12];potatoes,40[3,7];carrots,30[3,7]}",`

Fixes #4824

### DIFF
--- a/src/Generating/PieceModifier.cpp
+++ b/src/Generating/PieceModifier.cpp
@@ -1,0 +1,402 @@
+
+// PieceModifier.cpp
+
+// Implements the various classes descending from cPiece::cPieceModifier
+
+#include "Globals.h"
+#include "PieceModifier.h"
+#include "../Noise/Noise.h"
+
+
+
+
+
+// Constant that is added to seed
+static const int SEED_OFFSET = 135 * 13;
+
+
+
+
+
+// Emit a warning if the first param is true
+#define CONDWARNING(ShouldLog, Fmt, ...) \
+	do { \
+		if (ShouldLog) \
+		{ \
+			LOGWARNING(Fmt, __VA_ARGS__); \
+		} \
+	} while (false)
+
+
+
+
+
+////////////////////////////////////////////////////////////////////////////////
+/** A modifier which is pseudo-randomly replacing blocks to other types and metas. */
+class cPieceModifierRandomizeBlocks:
+	public cPiece::cPieceModifier
+{
+public:
+	cPieceModifierRandomizeBlocks(void) :
+		m_Seed()
+	{
+	}
+
+	virtual bool InitializeFromString(const AString & a_Params, bool a_LogWarnings) override
+	{
+		m_AllWeights = 0;
+		AString Params = a_Params;
+
+
+		/** BlocksToReplace parsing */
+		auto idxPipe = Params.find('|');
+		if (idxPipe != AString::npos)
+		{
+			AString blocksToReplaceStr = Params.substr(0, idxPipe);
+			auto blocksToReplace = StringSplitAndTrim(blocksToReplaceStr, ",");
+			for (size_t i = 0; i < blocksToReplace.size(); i++)
+			{
+				BLOCKTYPE blockType = static_cast<BLOCKTYPE>(BlockStringToType(blocksToReplace[i]));
+				if ((blockType == E_BLOCK_AIR) && !NoCaseCompare(blocksToReplace[i], "Air"))
+				{
+					CONDWARNING(a_LogWarnings, "Cannot parse block type from string \"%s\"!", blocksToReplace[i].c_str());
+					return false;
+				}
+				m_BlocksToReplace[blockType] = 1;
+			}
+
+			Params = Params.substr(idxPipe + 1, Params.length() - 1);
+		}
+		if (m_BlocksToReplace.size() == 0)
+		{
+			CONDWARNING(a_LogWarnings, "You must specify at least one block to replace%s!", "");
+			return false;
+		}
+
+
+		/** Meta params parsing */
+		auto idxSqBracketStart = Params.find('[');
+		auto idxSqBracketStartLast = Params.find_last_of('[');
+
+		bool isMultiMeta = false;
+		if ((idxSqBracketStart != idxSqBracketStartLast) && (idxSqBracketStartLast != Params.length() - 1))
+		{
+			// Meta per block type
+			isMultiMeta = true;
+		}
+		else
+		{
+			auto idxSqBracketEnd = Params.find(']');
+			if ((idxSqBracketEnd == Params.length() - 1) && (idxSqBracketStart != AString::npos))
+			{
+				AString metaParamsStr = Params.substr(idxSqBracketStart + 1, Params.length() - idxSqBracketStart - 2);
+				std::array<int, 4> metaParamsInt;
+				if (!ParseMeta(metaParamsStr, metaParamsInt, a_LogWarnings))
+				{
+					return false;
+				}
+
+				m_MinMeta = metaParamsInt[0];
+				m_MaxMeta = metaParamsInt[1];
+				m_MinNoiseMeta = metaParamsInt[2];
+				m_MaxNoiseMeta = metaParamsInt[3];
+
+				Params = Params.substr(0, idxSqBracketStart);
+			}
+		}
+
+
+		// BlocksToRandomize parsing
+		auto BlocksToRandomize = StringSplitAndTrim(Params, ";");
+		for (size_t i = 0; i < BlocksToRandomize.size(); i++)
+		{
+			AString block = BlocksToRandomize[i];
+
+			cRandomizedBlock Block{};
+
+			if (isMultiMeta)
+			{
+				auto sqBrStart = block.find('[');
+				if (sqBrStart != AString::npos)
+				{
+					auto sqBrEnd = block.find(']');
+					if (sqBrEnd != block.size() - 1)
+					{
+						CONDWARNING(a_LogWarnings, "If present, block meta params must be at the end of block to randomize definition \"%s\"!", block.c_str());
+						return false;
+
+					}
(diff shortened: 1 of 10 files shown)


12xx12:
Thank you for your PR!

That's a nice solution to randomize generation of anything placed by the generator. (I think)

Maybe add something to the randomize strategy setup string that provides some options (blocks, integers,...) something like
`["RandomizeStrategy"] = "Crops|Village|Beetroots;Carrots;Potatoes;Air;"`

This might be used to get the options for certain generators. (Like Stronghold like you said)
And add some customization for everyone.

Can you add some screenshots to show the results of your changes?

12xx12:
Another thing: The files used for generation come from our gallery server where you can build the structures ingame. Any changes done to the files should be done on the server also to provide consistency.

The old host of the gallery server has taken ther server down at this point. we should think about a replacement

@cuberite/core-team

NiLSPACE:
I don't think @madmaxoft took the server down as it's still making backups. I believe his DNS failed or something. Perhaps he's still willing to host the server.

This change would probably also require changes to the GalExport plugin to support this additional setting. I do agree with @12xx12 . I'd prefer the randomized blocks to be configurable instead of hardcoded inside the cuberite source code.

Other than that this PR seems good :)

dImrich:
> Maybe add something to the randomize strategy setup string that provides some options (blocks, integers,...) something like
> `["RandomizeStrategy"] = "Crops|Village|Beetroots,0.5;Carrots,0.2;Potatoes,0.2;Air,0.1;"`

Yeah I thought about making it more configurable exactly like in your example but it is a little bit ugly solution storing multiple blocks type/meta/weight in string so I decided to go Notchian way hiding it in the code and using only identifiers for the same strategies. 

> Can you add some screenshots to show the results of your changes?

Sure.
image
image


> Another thing: The files used for generation come from our gallery server where you can build the structures ingame. Any changes done to the files should be done on the server also to provide consistency.

Yes, I know but there isn't another way currently. I thought world and sqlite files from that server are lost so I didn't mind editing and adding few lines.

> The old host of the gallery server has taken ther server down at this point. we should think about a replacement
> 
> @cuberite/core-team

Are world and databases public or someone active has them?

NiLSPACE:
The backups can be found here: http://cuberite.xoft.cz/GalleryBackup/

dImrich:
@NiLSPACE Thanks for backups! Yeah I do agree too but I was too lazy 😃 
GalExport seems to only warn about using unknown metadata but it should be definitely be added to it. I will create PR in GalExport repo after this merges.

madmaxoft:
Huh, the Gallery server should still be running, the only problem is that I don't have a public IPv4 address anymore, so I'm doing things through an SSH reverse tunnel through a public VPS host. I guess I never checked the actual Minecraft port being available; the webserver got borked a week ago when I upgraded my last Intel machine to a new AMD one. I'll fix the portmap in a moment, any changes to the prefabs should be done on the Gallery server so that it's the only one source of truth.

madmaxoft:
The Gallery server is up again, the portmap was missing and the firewall on the VPS prevented connections.

dImrich:
I rewroted it. Here is some examples:

`["RandomizeStrategy"] = "Crops|beetroots,5;carrots,15;potatoes,16;air,1"`

`["RandomizeStrategy"] = "Crops|beetroots,5;carrots,15;potatoes,16;air,5[2, 6]"`

`["RandomizeStrategy"] = "Crops|beetroots,5;carrots,15;potatoes,16;air,1[2, 6, 0, 9]"`

`["RandomizeStrategy"] = "Crops|cobblestone|beetroots,5;carrots,15;potatoes,16;air,1"`

Integers in squared brackets at the end of string are: MinGrowthStage, MaxGrowthStage, m_MaxNoiseGrowthStage, m_MinNoiseGrowthStage.
Cobblestone in last example is block which is being replaced instead of wheat.

After this it should be renamed to something like RandomizeStrategyBasic as it can randomize a lot more than crops now.

src/Generating/RandomizeStrategy.cpp
```diff
@@ -0,0 +1,264 @@
...
+			AString blockToReplaceStr = Params.substr(0, idxPipe);
+			StringToItem(blockToReplaceStr, blockToReplace);
+			if (blockToReplace.IsEmpty())
+			{
+				m_BlockToReplace = blockToReplace.m_ItemType;
+			}
+			else
+			{
+				// Failed to parse block to replace:
+				CONDWARNING(a_LogWarnings, "Cannot parse block to replace from string \"%s\"!", blockToReplaceStr.c_str());
```
NiLSPACE, line 63: I believe a return false is missing here.

src/Generating/RandomizeStrategy.cpp
```diff
@@ -0,0 +1,264 @@
...
+					m_MaxGrowthStage = GrowthValue;
+					break;
+				case 2:
+					m_MaxNoiseGrowthStage = GrowthValue;
+					break;
+				case 3:
+					m_MinNoiseGrowthStage = GrowthValue;
+					break;
+				default:
+					CONDWARNING(a_LogWarnings,"Unsupported crops growth param \"%d\"!", GrowthValue);
```
NiLSPACE, line 107: Perhaps this should return false as well?

NiLSPACE:
I think it would be better if the strategy doesn't default to crops. How about making it mandatory to specify which blocks to replace?

src/Generating/RandomizeStrategy.cpp
dImrich: Block to replace is optional param like I showed in examples, default is wheat in this case of crops. Maybe after I rename it to Basic not Crops it should be required param.

src/Generating/RandomizeStrategy.cpp
dImrich: Yeah perhaps, I intentionally didn't return there 😀

src/Generating/RandomizeStrategy.cpp
NiLSPACE: But in this case the warning is shown when a block is actually specified but it failed to parse. Currently if the parsing fails `m_BlockToReplace` isn't filled but it still goes ahead right?

NiLSPACE: So if I'd use this string it wouldn't fill `m_BlockToReplace` but it would still go ahead with only a warning:

`["RandomizeStrategy"] = "Crops|someblockthatdoesntexist|beetroots,5;carrots,15;potatoes,16;air,1"`

dImrich:
> I think it would be better if the strategy doesn't default to crops. How about making it mandatory to specify which blocks to replace?

Yes, like I wrote above, after last commit I should rename it and remove all remains of crops. Also block/s to replace should be required, without any default.
I'll do it this evening.

src/Generating/RandomizeStrategy.cpp
dImrich: Yes you are right, it was wrong design letting invalid block's be in metadata without warning.

src/Generating/PieceModifier.cpp
```diff
@@ -0,0 +1,410 @@
...
+				BLOCKTYPE blockType = static_cast<BLOCKTYPE>(BlockStringToType(blocksToReplace[i]));
+				if ((blockType == E_BLOCK_AIR) && !NoCaseCompare(blocksToReplace[i], "Air"))
+				{
+					CONDWARNING(a_LogWarnings, "Cannot parse block type from string \"%s\"!", blocksToReplace[i].c_str());
+					return false;
+				}
+				m_BlocksToReplace[blockType] = 1;
+			}
+
+			if (m_BlocksToReplace.size() == 0)
```
NiLSPACE, line 67: Should this if condition be outside the `if (idxPipe != AString::npos)` scope? Otherwise it would still be possible to continue while having no pipeline right?

src/Generating/PieceModifier.cpp
```diff
@@ -0,0 +1,410 @@
...
+				if (sqBrStart != AString::npos)
+				{
+					auto sqBrEnd = block.find(']');
+					if (sqBrEnd != block.size() - 1)
+					{
+						CONDWARNING(a_LogWarnings, "If present, block meta params must be at the end of block to randomize definition \"%s\"!", block.c_str());
+						return false;
+
+					}
+					AString metaParamsStr = block.substr(sqBrStart + 1, block.size()-sqBrStart-2);
```
NiLSPACE, line 129: Could you add spaces between the minus operators? I'm not sure why the style checker doesn't see it.

NiLSPACE (approved):

PR #5129, opened 2021-02-14
fixed generator for mega taiga

### LINKED ISSUE #243: Add Mega Taiga Biome
This is a very rare biome
- [x] Contains 2x2 Spruce trees
- [x] Has a dirt block variant known as Podzol.
Currently obtained only by the silk touch enchantment.
- [x] Contains dirt blocks that do not allow grass to grow.
They have a data value of 1.
- [x] Moss Stone "boulders" are scattered around in this biome.
- [x] Dead Bushes
- [x] Mushrooms
- [x] Ferns

12xx12:
This is a very rare biome
- [x] Contains 2x2 Spruce trees
- [x] Has a dirt block variant known as Podzol.
Currently obtained only by the silk touch enchantment.
- [x] Contains dirt blocks that do not allow grass to grow.
They have a data value of 1.
- [x] Moss Stone "boulders" are scattered around in this biome.
- [x] Dead Bushes, Mushrooms, and Ferns spawn in this biome.

Most of the stuff has already been done. There are some small details missing

fixes #243 

Looks like this right now:

### DIFF
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -2235,7 +2235,7 @@ void cFinishGenForestRocks::GenFinish(cChunkDesc & a_ChunkDesc)
 		0,
 		m_Noise.IntNoise2DInt(a_ChunkDesc.GetChunkX(), a_ChunkDesc.GetChunkZ()) % cChunkDef::Width
 		);
-	Pos.y = a_ChunkDesc.GetHeight(Pos.x, Pos.z);
+	Pos.y = a_ChunkDesc.GetHeight(Pos.x, Pos.z) % cChunkDef::Height;
 
 	if (!cChunkDef::IsValidRelPos(Pos))
 	{
@@ -2257,25 +2257,25 @@ void cFinishGenForestRocks::GenFinish(cChunkDesc & a_ChunkDesc)
 		Radius = 3;
 	}
 
-	auto YSize = (m_Noise.IntNoise2DInt(a_ChunkDesc.GetChunkX(), a_ChunkDesc.GetChunkZ()) % 2) + 1;
-
 	Pos.x = Clamp(Pos.x, 0 + Radius, cChunkDef::Width - Radius - 1);
 	Pos.z = Clamp(Pos.z, 0 + Radius, cChunkDef::Width - Radius - 1);
 
-	auto StartBlock = a_ChunkDesc.GetBlockType(Pos.x, Pos.y + YSize, Pos.z);
+	auto StartBlock = a_ChunkDesc.GetBlockType(Pos.x, Pos.y + Radius, Pos.z);
 	while (!((StartBlock == E_BLOCK_DIRT) || (StartBlock == E_BLOCK_GRASS)))
 	{
 		Pos.y -= 1;
-		if (!cChunkDef::IsValidRelPos(Pos.addedY(-YSize)))
+		if (!cChunkDef::IsValidRelPos(Pos.addedY(-Radius)))
 		{
+			LOGWARNING("Test");
 			return;
 		}
-		StartBlock = a_ChunkDesc.GetBlockType(Pos.x, Pos.y + YSize, Pos.z);
+		LOGERROR("Height: %d", Pos.y);
+		StartBlock = a_ChunkDesc.GetBlockType(Pos.x, Pos.y + Radius, Pos.z);
 	}
 
 	for (int x = -Radius; x <= Radius; x++)
 	{
-		for (int y = -YSize; y <= YSize; y++)
+		for (int y = -Radius; y <= Radius; y++)
 		{
 			for (int z = -Radius; z <= Radius; z++)
 			{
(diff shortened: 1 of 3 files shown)


12xx12:
Updated screenshot with mossy cobble stone

Titaniumtown:
Anything preventing this being merged?

12xx12:
Me not being able to merge my own PR 😀

src/Generating/FinishGen.cpp
```diff
@@ -2185,3 +2213,82 @@ void cFinishGenOrePockets::imprintSphere(
...
+
+
+
+void cFinishGenForestRocks::GenFinish(cChunkDesc & a_ChunkDesc)
+{
+	auto Biome = a_ChunkDesc.GetBiome(cChunkDef::Width / 2, cChunkDef::Width / 2);
+	if ((Biome != biMegaTaiga) && (Biome != biMegaTaigaHills))
+	{
+		return;
+	}
```
tigerw, line 2230: Should this come after we've picked an X, Z or is this intended?

src/Generating/FinishGen.cpp
```diff
@@ -2185,3 +2213,82 @@ void cFinishGenOrePockets::imprintSphere(
...
+
+	// Choose random position in chunk and place boulder around it
+	auto Pos = Vector3i(
+		m_Noise.IntNoise2DInt(a_ChunkDesc.GetChunkX(), a_ChunkDesc.GetChunkZ()) % cChunkDef::Width,
+		0,
+		m_Noise.IntNoise2DInt(a_ChunkDesc.GetChunkX(), a_ChunkDesc.GetChunkZ()) % cChunkDef::Width
+		);
+	Pos.y = a_ChunkDesc.GetHeight(Pos.x, Pos.z) % cChunkDef::Height;
+
+	if (!cChunkDef::IsValidRelPos(Pos))
```
tigerw, line 2240: Why wouldn't Pos be valid here?

src/Generating/FinishGen.cpp
```diff
@@ -2185,3 +2213,82 @@ void cFinishGenOrePockets::imprintSphere(
...
+	{
+		Radius = 2;
+	}
+	else if (RadiusChance > ThreeLimit)
+	{
+		Radius = 3;
+	}
+
+	Pos.x = Clamp(Pos.x, 0 + Radius, cChunkDef::Width - Radius - 1);
+	Pos.z = Clamp(Pos.z, 0 + Radius, cChunkDef::Width - Radius - 1);
```
tigerw, line 2261: `// Maintain distance from edge of chunk to avoid getting clipped:`

Also an extra `0 +`, or is that intended?

src/Generating/FinishGen.cpp
```diff
@@ -2185,3 +2213,82 @@ void cFinishGenOrePockets::imprintSphere(
...
+	auto StartBlock = a_ChunkDesc.GetBlockType(Pos.x, Pos.y + Radius, Pos.z);
+	while (!((StartBlock == E_BLOCK_DIRT) || (StartBlock == E_BLOCK_GRASS)))
+	{
+		Pos.y -= 1;
+		if (!cChunkDef::IsValidRelPos(Pos.addedY(-Radius)))
+		{
+			return;
+		}
+		StartBlock = a_ChunkDesc.GetBlockType(Pos.x, Pos.y + Radius, Pos.z);
+	}
```
tigerw, line 2272: GetBlockType at the start might be called with out of bounds Y if Pos.y is high enough.

If I understand correctly, this always sinks the tip of the boulder flush with the ground. What do you think about placing the boulder's centre at the first valid StartBlock, then randomly moving down by [0, -Radius], and finally clamping Pos.y?,

src/Generating/FinishGen.cpp
```diff
@@ -2185,3 +2213,82 @@ void cFinishGenOrePockets::imprintSphere(
...
+		for (int y = -Radius; y <= Radius; y++)
+		{
+			for (int z = -Radius; z <= Radius; z++)
+			{
+				if (!cChunkDef::IsValidRelPos({ Pos.x + x, Pos.y + y, Pos.z + z }))
+				{
+					continue;
+				}
+
+				if (Vector3d(x, y, z).Length() - 0.25 > Radius)
```
tigerw, line 2285: Vector3f and SqrLength > (Radius^2 + 0.5 * Radius) would be a small optimisation, sqrt is apparently expensive

src/Generating/FinishGen.cpp
12xx12: actually we could use the position x,z to check if that block is valid - you are right

src/Generating/FinishGen.cpp
12xx12: Good Question

src/Generating/FinishGen.cpp
12xx12: This is just left over from experimenting

src/Generating/FinishGen.cpp
12xx12: calculating sqrt sux on any  cpu - yeah

src/Generating/FinishGen.cpp
12xx12: The thing is that the position might be on top of a tree (the same reason why we don't get grass generation beneath trees) so we would need to move this down to the ground first and then randomize

12xx12:
Updated generation

12xx12:
I like the new stones better

PR #5133, opened 2021-02-20
Mark UNREACHABLE with intrinsics

tigerw:


src/Globals.h
```diff
@@ -35,11 +35,15 @@
...
 
+	#define UNREACHABLE_INTRINSIC __assume(false)
+
 #elif defined(__GNUC__)
 
 	// TODO: Can GCC explicitly mark classes as abstract (no instances can be created)?
 	#define abstract
 
+	#define UNREACHABLE_INTRINSIC __builtin_unreachable()
+
```
12xx12, line 46: This will check the paths at compile time - right?

12xx12: Ok - it's more complicated. But It tells the compiler/optimiser that there is no more code that can be executed

12xx12 (approved):

tigerw:
Thanks!

PR #5141, opened 2021-03-04
Modular MobAi with Parser

12xx12:
On #4248 @peterbell10  started a mob AI.

I gave it a spin and added a LUA parser. This is just a raw scetch. On Windows the TPS is like 10 atm in debug moment.

Just an example


bearbin (commented):
This looks good. Ideally this wouldn't just languish as a PR - could this be merged in behind a feature flag until we do some more development on it?

12xx12:
This can be merged if we want. I am not quite sure about performance.

And the only thing holding this back is me not being sure about how tiny the building blocks for the breeding behaviour should be

src/Entities/Entity.cpp
```diff
@@ -1663,10 +1663,10 @@ void cEntity::SetSwimState(cChunk & a_Chunk)
 			{
 				BLOCKTYPE Block;
 				if (!a_Chunk.UnboundedRelGetBlockType(x, y, z, Block))
-				{
+				{ /*
 					LOGD("SetSwimState failure: RelX = %d, RelY = %d, RelZ = %d, Pos = %.02f, %.02f}",
 						x, y, z, GetPosX(), GetPosZ()
-					);
+					); */
 					continue;
```
12xx12, line 1670: Undo

src/Mobs/AI/BehaviorTree.cpp
```diff
@@ -0,0 +1,577 @@
...
+		case eStatus::bsCancelled: return "Cancelled";
+	}
+	UNREACHABLE("Invalid BehaviorTree::eStatus");
+}
+
+
+
+
+
+std::unique_ptr<cTask> CreateBehaviourTree(eMonsterType a_MobType, const AString & a_CustomFileName)
```
12xx12, line 49: Maybe have a separate constructor - one from string (for custom path) and one from mob type (for default)

src/Mobs/AI/BehaviorTree.cpp
```diff
@@ -0,0 +1,577 @@
...
+	auto Lua = cLuaState(LuaSubsystemName);
+	Lua.Create();
+	cLuaState::cLock lock(Lua);
+	Lua.LoadFile(FilePath);
+
+	auto BehavioursLua = Lua.WalkToNamedGlobal("BehaviorTree");
+
+	if (!BehavioursLua.IsValid() || !lua_istable(Lua, 1))
+	{
+		LOGWARNING("The BehaviorTree file %s doesn't contain any behaviours", FilePath.c_str());
```
12xx12, line 80: Maybe better announce it doesn't contain valid ones

src/Mobs/AI/BehaviorTree.cpp
```diff
@@ -0,0 +1,577 @@
...
+	{
+		return std::make_unique<cSequence>(a_LuaState);
+	}
+	else if (NoCaseCompare(a_BehaviourName, "Succeeder") == 0)
+	{
+		return std::make_unique<cSucceeder>(a_LuaState);
+	}
+	else
+	{
+		LOGERROR("Unimplemented behaviour parsed!: %s", a_BehaviourName);
```
12xx12, line 172: Maybe exit in debug mode?

src/Mobs/AI/BehaviorTree.cpp
```diff
@@ -0,0 +1,577 @@
...
+}
+
+
+
+
+
+////////////////////////////////////////////////////////////////////////////////
+// cGuardedBehavior:
+
+cGuardedTask::cGuardedTask(cLuaState & a_LuaState)
```
12xx12, line 209: I am not sure if we should keep all the available options on the same file (see Finishgen.cpp)

src/Mobs/Cow.cpp
```diff
@@ -9,8 +9,41 @@
...
+			std::make_unique<cGuardedTask>(
+				std::make_unique<cHealthRange>(0.0f, 0.5f * m_MaxHealth),
+				std::move(Panic)
+			),
+			std::make_unique<cGuardedTask>(
+				nullptr,
+				std::move(Wander)
+			)
+		)
+	);
```
12xx12, line 44: Remove duplicate

12xx12 (commented):
Some notes I got

12xx12:
So this can be merged in as is. This can be worked on bit by bit.

12xx12:
Check if tasks may be split into multiple files so we don't create a new generator finisher file

12xx12:
Working at this again. Adding some documentation and maybe tests.

Adding also a way to trigger walking towards a player when a proper item is held.

PR #5146, opened 2021-03-06
Mitigate MSVC crash

tigerw:
Reference: http://ci.appveyor.com/project/Cuberite/cuberite/builds/38087390/job/p857ibg3x87naw36/messages

### DIFF
--- a/src/ChunkData.h
+++ b/src/ChunkData.h
@@ -130,33 +130,24 @@ public:
 
 
 
-namespace ChunkDef
-{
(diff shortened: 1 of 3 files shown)


12xx12:
I saw that bug too and tried to fix it by forcing a certain callback not a class as a template - but that didn't seem to work.

12xx12 (approved):

12xx12:
By the way the MSVC doesn't work deterministic - sometimes it works - sometimes it doesn't

Issue #1318, opened 2014-08-13
Start using emplace_back() instead of push_back(new x()) where possible


worktycho:
This is reminder for after we move to c++11

tigerw:
Is this in the context of shared/unique ptrs or constructing objects generally? The "new" makes me unsure, in any case a search for `push_back\(.*\(` would help.

12xx12:
do we want to close this after #5149

PR #5149, opened 2021-03-06
Some emplace_back replacements

12xx12:
ref #1318

So I did some reading.
If I understood correctly emplace_back is worth when a new object is created in the call. So I went and replaced those calls.

There are some cases where the object is also stored somewhere else. Those I kept at `push_back`. 
There are some cases where an object is created to be pushed and returned. about those I was unsure and didn't replace.

### DIFF
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -45,7 +45,7 @@ void cCompositeChat::Clear(void)
 
 void cCompositeChat::AddTextPart(const AString & a_Message, const AString & a_Style)
 {
-	m_Parts.push_back(TextPart{{ a_Message, a_Style, {} } });
+	m_Parts.emplace_back(TextPart{{ a_Message, a_Style, {} } });
 }
 
 
@@ -54,7 +54,7 @@ void cCompositeChat::AddTextPart(const AString & a_Message, const AString & a_St
 
 void cCompositeChat::AddClientTranslatedPart(const AString & a_TranslationID, const AStringVector & a_Parameters, const AString & a_Style)
 {
-	m_Parts.push_back(ClientTranslatedPart{{ a_TranslationID, a_Style, {} }, a_Parameters });
+	m_Parts.emplace_back(ClientTranslatedPart{{ a_TranslationID, a_Style, {} }, a_Parameters });
 }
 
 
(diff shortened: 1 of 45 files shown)


src/WorldStorage/WorldStorage.cpp
```diff
@@ -167,8 +167,8 @@ void cWorldStorage::QueueSaveChunk(int a_ChunkX, int a_ChunkZ)
 void cWorldStorage::InitSchemas(int a_StorageCompressionFactor)
 {
 	// The first schema added is considered the default
-	m_Schemas.push_back(new cWSSAnvil    (m_World, a_StorageCompressionFactor));
-	m_Schemas.push_back(new cWSSForgetful(m_World));
+	m_Schemas.emplace_back(new cWSSAnvil    (m_World, a_StorageCompressionFactor));
```
tigerw, line 170: I think all the pointers should remain push_back, ref http://stackoverflow.com/questions/15783342/should-i-use-c11-emplace-back-with-pointers-containers

src/CompositeChat.cpp
```diff
@@ -45,7 +45,7 @@ void cCompositeChat::Clear(void)
 
 void cCompositeChat::AddTextPart(const AString & a_Message, const AString & a_Style)
 {
-	m_Parts.push_back(TextPart{{ a_Message, a_Style, {} } });
+	m_Parts.emplace_back(TextPart{{ a_Message, a_Style, {} } });
```
tigerw, line 48: I think these should remain push_back. Unless you want to write a constructor for all the parts, then you could use `emplace_back(std::in_place_type<TextPart>, ...)`. The constructor is needed because ref http://stackoverflow.com/questions/8782895/why-doesnt-emplace-back-use-uniform-initialization

tigerw:
Thanks for going through all of these!

12xx12:
No worries

There are some push_back(std::move()) calls left over where I wasn't sure if they should be replaced

src/WorldStorage/WorldStorage.cpp
12xx12: I thought the lists all contained smart pointers. But yeah - if we push back normal pointers you're right

12xx12: There are actually some smart pointer lists, arrays,...

e.g. the BioGen caches

tigerw (approved):

tigerw:
Fixes #1318

Issue #4868, opened 2020-09-15
Blocks placement non-conformaties to vanilla 1.10.2


theophriene:
Client version: 1.10.2
Server OS: Linux
Cuberite Commit id:  d9d71e1 built at: approx: 2020-09-13 14:44:20 +0100

### Can be placed on top of themselves (or block if specified), but shouldn't
1. #### Drop upon adjacent block update - (though can be interacted with without them dropping)
* repeaters, comparators 
* buttons (any)
* pressure plates (any)
* levers
* Saplings, and flowers (before mushrooms in palette) - on top of solid blocks (eg. iron block)
2. #### Keep staying after block update
* mushrooms (any) 

### Can't be be put, but should
1. #### Is not created at all (or immediately destroyed without pickup)
* Snow on top of snow (can only be generated)
* Anything in place of deep snow -> is placed one block higher, as if it's solid
* Rails (any) on top-half slab
* Armor stands (if implemented)
2. #### Is created, until adjacent BUD
* Pressure plates (any) on top-half slabs
* Buttons (any) on top-half slabs

### These realign themselves the one possible way, regardless of the direction being placed
* Levers - rotate
* ender rods - rotate
* trapdoors (any) - move to upper half of the block
* daylight sensors - move to lower half of the block

Those are the ones, I've discovered so far. Though it's 1.10.2, so something may have changed since then, IDK.

tigerw:
Ref #4237

Thanks for detailed notes!

tigerw:
Ref #2919

PR #5151, opened 2021-03-10
Block State PR

12xx12:
This is just to show the progress and to get some additional input

TODO:
https://github.com/12xx12/cuberite/projects/2

Preview Server: 12xx12.de

These things are changed
- Block States are now used internally
- Many Funtions that took 3 integers and vector for coordinates now only take Vectors


tigerw:
yes, this looks about right. What do you think we should do about the API, pass the BLOCKTYPE/NIBBLETYPE through LegacyPalette?

12xx12:
On the forum there is a thread for that. @NiLSPACE made the proposal to map globals to the block state values.

There are still some problems. Especially for the new blocks. (There is no old ENUM constants)

I think there will be some API breaking leftover - we will see this after their server compiles again.

At this point I'm at Unity file 27 and counting down. (0 is the last one)

I'll be not coding for the weekend so knock yourself out if you want

12xx12:
Another thing: 

I'm gonna do some gameplay testing 

BUT either I need help or we can maybe YOLO this and push too Master and wait for the error reports

PureTryOut:
I can help testing but I would need to know what to test exactly. Should I test specific blocks, should they have changed behaviour or the same as before, etc.

12xx12:
We need to test... everything

Placing, interacting, breaking, loading saving

12xx12:
I feel like it's 30-40% done

12xx12:
Oh and the reason the is so much new code is that we can't do mathematic operations for rotations anymore but have nested switch statements (eg in the block handlers)

12xx12:
Status update: The PR compiles until Unity file 22 with MVSC

12xx12:
I'm at Unity File 17. 14597 lines of compiler errors to go

tigerw:
Thanks for starting such a big undertaking! Having worked with the blockstates stuff for a bit, what do you think about its ease of use, ergonomics? Do you have any suggestions on how to improve usability? For example, might it be worth manually combining some of the separate material blocks into one, to reduce `switch`ing on stuff (e.g. OakFence/StoneFence/AcaciaFence -> `Fence(Material=xx)`)?

12xx12:
I tried to wrap any uneasy to use (like similar blocks like stairs, slabs,...) into functions from the block handlers - this seemed the most convenient way for me

12xx12:
I'm so lookig forward to when this finaly compiles :D

12xx12:
Does anyone have a Idea how to keep track of the testing process. Like a online form service where we enter the status of the blocks?

12xx12:
@tigerw
The waterlogged block states seem to be missing. I assume you autogenerated this from the server .jar file. Can we add this? and the default constructors of the blocks are not constexpr. So for example:

```
	namespace AcaciaPressurePlate
	{
		constexpr BlockState AcaciaPressurePlate(const bool Powered)
		{
			if (Powered) return 3881;
			else return 3882;
		}
		BlockState AcaciaPressurePlate();
		bool Powered(BlockState Block);
	}
```
the function `BlockState AcaciaPressurePlate();` is not marked constexpr which makes some constexpr switch statements impossible (if that even makes sense)

12xx12:
Or are you thinking about adding some stuff in the Type() function in the BlockSate class?

Hm... this might be nice

tigerw:
I was thinking some manual registry changes, so something like

```cpp
			case BlockType::BlackBed:     return BlackBed::Facing(a_Block);
			case BlockType::BlueBed:      return BlueBed::Facing(a_Block);
			case BlockType::BrownBed:     return BrownBed::Facing(a_Block);
			case BlockType::CyanBed:      return CyanBed::Facing(a_Block);
			case BlockType::GrayBed:      return GrayBed::Facing(a_Block);
			case BlockType::GreenBed:     return GreenBed::Facing(a_Block);
			case BlockType::LightBlueBed: return LightBlueBed::Facing(a_Block);
			case BlockType::LightGrayBed: return LightGrayBed::Facing(a_Block);
			case BlockType::LimeBed:      return LimeBed::Facing(a_Block);
			case BlockType::MagentaBed:   return MagentaBed::Facing(a_Block);
			case BlockType::OrangeBed:    return OrangeBed::Facing(a_Block);
			case BlockType::PinkBed:      return PinkBed::Facing(a_Block);
			case BlockType::PurpleBed:    return PurpleBed::Facing(a_Block);
			case BlockType::RedBed:       return RedBed::Facing(a_Block);
			case BlockType::WhiteBed:     return WhiteBed::Facing(a_Block);
			case BlockType::YellowBed:    return YellowBed::Facing(a_Block);
```

becomes

```cpp
			case BlockType::Bed:     return Bed::Facing(a_Block);
```

and Bed gets an extra `Bed::Color()` property.

-----

Where do you need to constexpr-construct a default block state?

12xx12:
> I was thinking some manual registry changes, so something like
> 
> ```c++
> 			case BlockType::BlackBed:     return BlackBed::Facing(a_Block);
> 			case BlockType::BlueBed:      return BlueBed::Facing(a_Block);
> 			case BlockType::BrownBed:     return BrownBed::Facing(a_Block);
> 			case BlockType::CyanBed:      return CyanBed::Facing(a_Block);
> 			case BlockType::GrayBed:      return GrayBed::Facing(a_Block);
> 			case BlockType::GreenBed:     return GreenBed::Facing(a_Block);
> 			case BlockType::LightBlueBed: return LightBlueBed::Facing(a_Block);
> 			case BlockType::LightGrayBed: return LightGrayBed::Facing(a_Block);
> 			case BlockType::LimeBed:      return LimeBed::Facing(a_Block);
> 			case BlockType::MagentaBed:   return MagentaBed::Facing(a_Block);
> 			case BlockType::OrangeBed:    return OrangeBed::Facing(a_Block);
> 			case BlockType::PinkBed:      return PinkBed::Facing(a_Block);
> 			case BlockType::PurpleBed:    return PurpleBed::Facing(a_Block);
> 			case BlockType::RedBed:       return RedBed::Facing(a_Block);
> 			case BlockType::WhiteBed:     return WhiteBed::Facing(a_Block);
> 			case BlockType::YellowBed:    return YellowBed::Facing(a_Block);
> ```
> 
> becomes
> 
> ```c++
> 			case BlockType::Bed:     return Bed::Facing(a_Block);
> ```

Yeah - this stuff is move into the blockhandler classes.
So in cBlockBedHandler there is

```c++
    static inline bool IsBlockBed(BlockState a_Block)
    static inline eBlockFace GetBlockFace(BlockState a_Block)
    static inline bool IsHeadPart(BlockState a_Block)
    static inline bool IsBedOccupied(BlockState a_Block)
```

12xx12:
> Where do you need to constexpr-construct a default block state?

I actually forgot that ;D 🤦

12xx12:
Compiles until unity file 9 with MVSC
Only 2786 lines of erros with gcc to go

12xx12:
https://forms.office.com/r/5GjVji8RNv

This should do the trick when collecting test data.
Am I missing something?

tigerw:
Seems good, though "what does work" might be redundant? Before publishing we want to fix #4868/#4237 and rebase on top to avoid false positives.

12xx12:
I'm not quite sure if this makes sense because then we're doing work twice sometimes because there are some operations with metas that can't be done with the block stated (adding for rotation,...)

12xx12:
Ok - it compiles now fully :D

12xx12:
Can anyone familiar with the manual exports help me fix the manually exported functions of cBlockArea @cuberite/core-team

12xx12:
We have many things to do

12xx12:
Yeah - I got 1.13 not throwing an error on connection... It was the weather packet -.- 
There seems to be no weather packet for 1.13 - https://wiki.vg/index.php?title=Protocol&oldid=14889

12xx12:
Oh and @cuberite/core-team:

We have kind a big problem: The metadata for some blocks is only written in the "legal" bytes. So for example the air block can have metadata written or in memory BUT actually has no metadata at all. This takes quite a bit of performance as we have to determine the actual correct metadata for the block or even load the false data -> see image above. We can do this with if/else or a even bigger switch statement to create block but I am unsure if this actually is great for performance.

I trie to check where the data is created to prevent old memory content to be sent to the player but the data is created inside the zlib library :(

tigerw:
I don't understand what you mean, could you rephrase?

12xx12:
This is just an example from loading a world from disk...

```
[a89cef58|09:37:28] 0, 255, 11111111
[a89cef58|09:37:28] 0, 127, 01111111
[a89cef58|09:37:28] 0, 63, 00111111
[a89cef58|09:37:28] 0, 31, 00011111
[a89cef58|09:37:28] 0, 15, 00001111
[a89cef58|09:37:28] 0, 7, 00000111
[a89cef58|09:37:28] 0, 3, 00000011
[a89cef58|09:37:28] 0, 1, 00000001
```

The first number is the block (in this case Air). The second number is the decimal representation of the block meta and the third value is the binary representation of the meta value.

A this point i wrote it tha way it bitshifts every try it shifts the meta on digit to the right but that breaks loading. (see image). If I just try to get the block with a meta value of 0 this works but the world won't be fully recreated because there is some information lost. I'm not quite sure how to proceede at this point

12xx12:
Or a dirt block

```
[aceb5029|09:43:13] 3, 17, 00010001
[aceb5029|09:43:13] 3, 8, 00001000
[aceb5029|09:43:13] 3, 4, 00000100
```

tigerw:
I see now. Is it just air (and a couple of other blocks) or can every block type exhibit the same issue? It seems you've found a latent bug.

12xx12:
Any Block - if you look at the image you see the sandstone stairs being misplaced and the dead bushes being there (instead of grass).

Tried both with a save from nothian client and our save

tigerw:
I meant the meta value being mismatched with the block type bug. Would something like this work?

```cpp
PaletteUpgrade::FromBlock(BLOCKTYPE Type, NIBBLETYPE Meta)
{
 switch (Type << 4 | Meta)
  case 0 << 4 | 0: return Air::Air();
  // ...

 // Now these are the metas that are mismatched, just use the block type, and return the default block state:
 switch (Type)
  case 0: return Air::Air();
  case 1: return Stone::Stone();
}
```

So essentially ignore the meta, but _only_ when it's wrong.

12xx12:
No I tried this already. I just called the function again with a meta value of 0.
This is my default case. And if the meta was 0 i return Air.

For example the logs diffirenciate their type by meta. so we need to case on the actual values that can be there but also the digits the can't be there and don't matter and contain junk

tigerw:
But if a blocktype, say log, needs a corresponding meta, that meta must be present and correct. Otherwise how is the server right now, sending the correct data to the client? The meta can't be junk because `master` is working properly. So the converter should work fine.

If a blocktype, say air or dirt, doesn't need meta, the expected case is that meta is zero. But say we have bugs where we write nonzero meta, or leave the value as garbage. Well okay, that's fine, because the blocktype is what matters in the conversion, so just clear the meta to zero.

12xx12:
I f*ck up:

Your old version:
switch ((Block << 4) | Meta)

My broken version:
switch ((Block << 8) | Meta)
This one made the errors.

My working version now
switch ((Block << 4) | Meta & 0x04)

tigerw:
That makes sense, great! :)

src/WorldStorage/WSSAnvil.cpp
```diff
@@ -398,9 +401,21 @@ bool cWSSAnvil::LoadChunkFromNBT(const cChunkCoords & a_Chunk, const cParsedNBT
...
+		if ((BlockData != nullptr) && (SkyLightData != nullptr) && (BlockLightData != nullptr))
 		{
-			Data.BlockData.SetSection(*reinterpret_cast<const ChunkBlockData::SectionType *>(BlockData), *reinterpret_cast<const ChunkBlockData::SectionMetaType *>(MetaData), static_cast<size_t>(Y));
+			BlockState Blocks[ChunkBlockData::SectionBlockCount];
+			for (size_t I = 0; I < ChunkBlockData::SectionBlockCount; I++)
+			{
+				unsigned char Meta = 0;
+				if (MetaData != nullptr)
+				{
+					Meta = static_cast<unsigned char>(MetaData[I]);
```
tigerw, line 412: Don't forget to say `cChunkDef::ExpandNibble(MetaData, i)`

The meta values in the old format at packed two entries into one byte.

12xx12: There we go

12xx12:
For some reason the performance is ver Very VERY bad - like 1FPS on 3070 :(

tigerw:
how many chunk updates in F3 debug? Does it get better after a while?

12xx12:
Like 20 maybe...
The problem ist that I wrote into the code that detects if a leaf should decay and realy many items were spawned...

12xx12:
Just wanted to do a status update: 

Still working on it: I just noticed that it makes more sense to also add the world loading and conversion code because there are some blocks which need to be converted in context and also the startup performance is poor

12xx12:
I added a preview server: 12xx12.de

12xx12:
So... This is passing all checks. Gameplay seems ok.

Still... There are some weird bugs and the server crashing on shutdown and the core plugin being broken and me not being sure if I manually added all functions with BLOCKTYPE, META transitions

12xx12:
Woooohooooo - the checks are passing

12xx12:
So this works. There are a few kinks to work out. And when stopping the server, it crashes.

MightyFilipns:
Superseded by #5599. Which is now merged into experimental. Closing.

PR #5153, opened 2021-03-11
Fixed some death messages showing up even when they are disabled.

NiLSPACE:
The world setting was only checked when a player died of natural causes

### DIFF
--- a/src/Broadcaster.cpp
+++ b/src/Broadcaster.cpp
@@ -186,6 +186,11 @@ void cWorld::BroadcastBossBarUpdateHealth(const cEntity & a_Entity, UInt32 a_Uni
 
 void cWorld::BroadcastChat(const AString & a_Message, const cClientHandle * a_Exclude, eMessageType a_ChatPrefix)
 {
+	if ((a_ChatPrefix == mtDeath) && !ShouldBroadcastDeathMessages())
+	{
+		return;
+	}
+
 	ForClientsInWorld(*this, a_Exclude, [&](cClientHandle & a_Client)
 		{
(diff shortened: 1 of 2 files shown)


src/World.h
```diff
@@ -188,7 +188,13 @@ class cWorld  // tolua_export
 	virtual void BroadcastChatSuccess(const AString & a_Message, const cClientHandle * a_Exclude = nullptr) override { BroadcastChat(a_Message, a_Exclude, mtSuccess); }
 	virtual void BroadcastChatWarning(const AString & a_Message, const cClientHandle * a_Exclude = nullptr) override { BroadcastChat(a_Message, a_Exclude, mtWarning); }
 	virtual void BroadcastChatFatal  (const AString & a_Message, const cClientHandle * a_Exclude = nullptr) override { BroadcastChat(a_Message, a_Exclude, mtFailure); }
-	virtual void BroadcastChatDeath  (const AString & a_Message, const cClientHandle * a_Exclude = nullptr) override { BroadcastChat(a_Message, a_Exclude, mtDeath); }
+	virtual void BroadcastChatDeath  (const AString & a_Message, const cClientHandle * a_Exclude = nullptr) override
+	{
+		if (ShouldBroadcastDeathMessages())
+		{
+			BroadcastChat(a_Message, a_Exclude, mtDeath);
```
12xx12, line 195: Can't you do this check in `cWorld::BrodcastChat`?
This will make the diff smaller

NiLSPACE: Yeah, good idea!

12xx12 (approved):

PR #5160, opened 2021-03-19
Change TimeOfDay to WorldDate

### LINKED ISSUE #4737: Day count in f3 not implemented
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: e053f72db824a16c1cb59107315001473058e2bc

### Expected behavior
There should be a number in the f3 menu stating how many in game days have past.

### Actual behavior
The Day count is always 0.

Open f3 on server running cuberite

### LINKED ISSUE #5159: World age isn't being sent to the client(?)
Client version: Any.
Server OS: Linux (Armbian)
Cuberite Commit id: 01a4e696b3d2c973cdd1fb4345d747bd10e93ad9

### Expected behavior
The client gets the worlds age.

### Actual behavior
It doesn't

image

Join a cuberite server. 
Press F3.
Look at Day counter.

tigerw:
* Do not wrap at 20 minutes, continue incrementing
* Fixes #4737
* Fixes #5159

### DIFF
--- a/Server/Plugins/APIDump/APIDesc.lua
+++ b/Server/Plugins/APIDump/APIDesc.lua
@@ -1870,7 +1870,7 @@ end
 							Type = "number",
 						},
 						{
-							Name = "TimeOfDay",
+							Name = "WorldDate",
 							Type = "number",
 						},
 						{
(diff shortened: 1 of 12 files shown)


tigerw:
cWorld::SetTimeOfDay is now poorly named and using beds will still reset the day count to zero, but for natural progression of time, the client should correctly displays days.

12xx12:
The time is now displayed properly but the /time command resets the timer back to the first day.
But it think this is a issue of the core plugin

12xx12:
maybe have a function that takes the values up to 24000 and always skips forward and one that takes absolute values

tigerw:
I tried to make SetTimeOfDay preserve the day count, that should fix Core.

12xx12:
I realy like how the time transition animation now always goes forward 👍

12xx12 (approved):

PR #5167, opened 2021-03-27
Fix Windows XP to 7 compatibility

tigerw:
* Partially reverts 01a4e696b
* Unify thread names
- Remove use of GetThreadId API

Protocol encryption uses AES128 in CFB8 mode as a stream cipher, which in theory would mean it can decrypt one byte at a time. This is important for the network, because any number of bytes can arrive at once, and a packet can start and end anywhere within them. However, the platform's CryptoAPI on <Windows 7 (8/8.1 untested) only accepts decryption of buffers that are multiples of the key size (128 bit / 8 = 16 bytes), which broke the [test server](https://forum.cuberite.org/thread-3261.html) that is running on Windows 7.

CryptoAPI disassembly

The multiples check. `div eax,ebx`.

Backtrace

EAX = 0x4A = ciphertext length
EBX = 0x10 = IV length

Fix by adding a PLATFORM_CRYPTOGRAPHY macro that is 1 when targeting Windows 8>. It would be good to provide "legacy Windows (XP-7)" and "modern Windows (8+)" official builds because ideally we still want to eventually replace mbedtls with CryptoAPI on Windows.

### DIFF
--- a/Tools/MCADefrag/MCADefrag.cpp
+++ b/Tools/MCADefrag/MCADefrag.cpp
@@ -126,7 +126,7 @@ AString cMCADefrag::GetNextFileName(void)
 // cMCADefrag::cThread:
 
 cMCADefrag::cThread::cThread(cMCADefrag & a_Parent) :
-	super("MCADefrag thread"),
+	super("MCA Defragmentor"),
 	m_Parent(a_Parent),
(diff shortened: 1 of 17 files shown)


tigerw:
urgent because demo server is broken

Issue #4266, opened 2018-07-26
Unify the DoWith function behavior


madmaxoft:
All the `DoWith` functions should adhere to the same behavior - they should return false if the object is not found, and they should return the value from the callback when the object was found.

Ref.: https://github.com/cuberite/cuberite/pull/4259#discussion_r205378652

madmaxoft:
Functions (checked if already adhering):
- [x] `cPluginManager::DoWithPlugin`
- [x] `cBlockArea::DoWithBlockEntityAt`
- [x] `cBlockArea::DoWithBlockEntityRelAt`
- [] `cChunk::DoWithEntityByID` - uses separate out param for the returned value
- [] `cChunk::DoWithBlockEntityAt` - inverted
- [] `cChunk::DoWithBeaconAt` - inverted
- [] `cChunk::DoWithBedAt` - inverted
- [] `cChunk::DoWithBrewingstandAt` - inverted
- [] `cChunk::DoWithChestAt` - inverted
- [] `cChunk::DoWithDispenserAt` - inverted
- [] `cChunk::DoWithDropperAt` - inverted
- [] `cChunk::DoWithDropSpenserAt` - inverted
- [] `cChunk::DoWithFurnaceAt` - inverted
- [] `cChunk::DoWithNoteBlockAt` - inverted
- [] `cChunk::DoWithCommandBlockAt` - inverted
- [] `cChunk::DoWithMobHeadAt` - inverted
- [] `cChunk::DoWithFlowerPotAt` - inverted

madmaxoft:
What a mess

madmaxoft:
The funny thing is, functions that do invert the callback return value do so while breaking the docs, the `cWorld::DoWith` function have this in their API docs:

*The function returns false if there is no chest, or if there is, it returns the bool value that the callback has returned.*

peterbell10:
And the doxy-comments agree with the code but not the Lua API docs:
https://github.com/cuberite/cuberite/blob/b8ab03bc6b41dbe8262f87def3ff3513924315d4/src/World.h#L592

madmaxoft:
Another possible way to tackle this would be to separate the two values and report both:
```cpp
std::pair<bool, bool> DoWithXYZ(...);
```
The first value tells whether the object was found, the second value is what the callback returned (or false if object not found?)
The Lua API could reflect this by having two return values.

peterbell10:
[...]

PR #5168, opened 2021-03-28
Unify DoWithBlockEntity

tigerw:
+ DoWith calls now broadcast the block entity and mark the chunk dirty
+ Add block entity change queue to synchronise BE updates with block updates
* Fixed a few incorrect assertions about BE type
- Remove manual overloads


tigerw:
Partially addresses #4266.

PR #5173, opened 2021-03-30
Fix finisher generating invalid pumpkin

tigerw:
Result was black spot in the ground because the meta exceeded 3 and the client didn't render anything.

Been here since the beginning? http://github.com/cuberite/cuberite/blob/fbabf9ee8c7f0940e6f7d01e0362849bf4f6396b/source/FinishGen.cpp#L159


tigerw:
I know @12xx12 would have fixed this anyway, just for posterity

12xx12 (approved):
YOLO - another merge :3

12xx12:
I think there will be some worlds generated befor I finish the PR

tigerw:
Thank you!

Issue #5164, opened 2021-03-24
Splash potion of harming has range issue


rehwinkel:
Client version: 1.9.4
Server OS: Arch Linux
Cuberite Commit id: 90369deb5a870674ab02a26a93a71f06eafe9293

### Expected behavior
If entities are far enough away, they should not receive any damage.

### Actual behavior
Entities take damage 10 or more blocks away from the potion impact. They seem to take 0 damage, however the damage effect is still shown and make damage sounds.

### Steps to reproduce the behavior
Throw potion of harming anywhere and look at entities nearby.

### Server log
```
[79ef0849918ae52f|12:05:58] HandleUseItem: Hand: 0; HeldItem: 373:16396:16396 * 1
[79ef0849918ae52f|12:05:58] Destroying entity #403 (cSplashPotionEntity) from chunk (-11, -6)
```

12xx12:
Fixed #5176

PR #5176, opened 2021-04-05
Change to Cuboid calculation for splash distance

tympaniplayer:
This is a fix for #5164 

The wiki mentions that the splash distance for a splash potion is 5 actually 8.25 x 8.25 x 4.25. I updated that instead of using the   20 spheroid distance. Basically everything < 20 was being affected by the splash 6 potion, but with the reduction harming had no effect except visual

### DIFF
--- a/src/Entities/SplashPotionEntity.cpp
+++ b/src/Entities/SplashPotionEntity.cpp
@@ -65,8 +65,11 @@ void cSplashPotionEntity::Splash(Vector3d a_HitPos)
 				return false;
 			}
 
-			double SplashDistance = (a_Entity.GetPosition() - a_HitPos).Length();
-			if (SplashDistance >= 20)
+			Vector3d SplashDistanceCube = (a_Entity.GetPosition() - a_HitPos);
+			double SplashDistance = SplashDistanceCube.Length();
+
+			// Wiki says 8.25 x 8.25 x 4.25 Cuboid.
+			if ((std::max(SplashDistanceCube.x, SplashDistanceCube.z)) > 8.25 || (SplashDistanceCube.y > 4.25))
 			{
 				// Too far away
 				return false;
--- a/src/Entities/SplashPotionEntity.cpp
+++ b/src/Entities/SplashPotionEntity.cpp
@@ -70,7 +70,6 @@ void cSplashPotionEntity::Splash(Vector3d a_HitPos)
 			}
 
 			// y = -0.25x + 1, where x is the distance from the player. Approximation for potion splash.
-			// TODO: better equation
 			double SplashDistance = (a_Entity.GetPosition() - a_HitPos).Length();
 			double Reduction = -0.25 * SplashDistance + 1.0;
 			Reduction = std::max(Reduction, 0.0);
--- a/src/Entities/SplashPotionEntity.cpp
+++ b/src/Entities/SplashPotionEntity.cpp
@@ -57,7 +57,11 @@ void cSplashPotionEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 
 void cSplashPotionEntity::Splash(Vector3d a_HitPos)
 {
-	m_World->ForEachEntity([=](cEntity & a_Entity)
+	double XZCalculation = 8.25/2;
+	double YCalculation = 4.25/2;
+	cBoundingBox SplashDistanceBox = cBoundingBox(a_HitPos.x - XZCalculation, a_HitPos.x + XZCalculation, a_HitPos.y - YCalculation, a_HitPos.y + YCalculation, a_HitPos.z - XZCalculation, a_HitPos.z + XZCalculation);
+
+	m_World->ForEachEntityInBox(SplashDistanceBox, [=](cEntity & a_Entity)
 		{
 			if (!a_Entity.IsPawn())
 			{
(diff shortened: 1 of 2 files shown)


tigerw:
We have a cWorld::ForEachEntityInBox; if we supply a bounding box centred on the splash potion entity, of the size the Wiki gives, it'd be more efficient than going through all entities in the world.

tympaniplayer:
Alright! I’ll take a look at that. Thanks for the advice

tympaniplayer:
I went ahead and re-implemented this using the cWorld::ForEachEntityInBox method that you suggested.

src/Entities/SplashPotionEntity.cpp
```diff
@@ -57,26 +57,21 @@ void cSplashPotionEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 
 void cSplashPotionEntity::Splash(Vector3d a_HitPos)
 {
-	m_World->ForEachEntity([=](cEntity & a_Entity)
+	double XZCalculation = 8.25/2;
+	double YCalculation = 4.25/2;
+	cBoundingBox SplashDistanceBox = cBoundingBox(a_HitPos.x - XZCalculation, a_HitPos.x + XZCalculation, a_HitPos.y - YCalculation, a_HitPos.y + YCalculation, a_HitPos.z - XZCalculation, a_HitPos.z + XZCalculation);
```
bearbin, line 62: The `Radius` constructor for cBoundingBox would be slightly clearer here.

src/Entities/SplashPotionEntity.cpp
```diff
@@ -57,26 +57,21 @@ void cSplashPotionEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
 
 void cSplashPotionEntity::Splash(Vector3d a_HitPos)
 {
-	m_World->ForEachEntity([=](cEntity & a_Entity)
+	double XZCalculation = 8.25/2;
+	double YCalculation = 4.25/2;
```
bearbin, line 61: Radius, and Height would be more intuitive, especially when using the simpler constructor below.

src/Entities/SplashPotionEntity.cpp
```diff
@@ -57,26 +57,21 @@ void cSplashPotionEntity::OnHitEntity(cEntity & a_EntityHit, Vector3d a_HitPos)
...
-			{
-				// Too far away
-				return false;
-			}
-
 			// y = -0.25x + 1, where x is the distance from the player. Approximation for potion splash.
 			// TODO: better equation
+			double SplashDistance = (a_Entity.GetPosition() - a_HitPos).Length();
 			double Reduction = -0.25 * SplashDistance + 1.0;
 			Reduction = std::max(Reduction, 0.0);
```
bearbin, line 76: Hmm, this isn't your code, but calling AddEntityEffect with a zero reduction means the existing effect is removed and the new effect has 0 duration. This seems wrong although I can't be totally sure of the vanilla behaviour - could you confirm?

tympaniplayer: Sure let me see if I can confirm. I assume that's why the original author put that todo in there.

src/Entities/SplashPotionEntity.cpp
tympaniplayer: Actually now that I think about it, the wiki mentions is a cuboid calculation and not a sphere. Unless I'm mistaken, wouldn't a Radius object (and the cBoundingBox constructor) be incorrect?

bearbin: Maybe it's not more inuitive then! It's just a different name for the same thing -> Radius == X/Z Distance from the centre, Height == Y Distance above the given point

https://github.com/cuberite/cuberite/blob/89fb16fc2717f24be516d2425e415b16120d07b4/src/BoundingBox.cpp#L32

Either way, the way you've done it already is fine, no need to make a change. I was just pointing out the alternative function available.

src/Entities/SplashPotionEntity.cpp
tympaniplayer: `For instant effects (i.e. Healing or Harming), the potency of the effect reduces linearly from 100% on a direct hit to 0% at 4 blocks' distance. For other effects, the potency is unchanged, but the duration decreases linearly on the same scale (rounded to the nearest 1⁄20 second), with no effect being applied if the duration would be 1 second or less. ` It does seem like a zero reduction is correct. Seems like the effect is not applied if it's 1 second or less.

bearbin: The issue is not that the effect won't be applied, but that hitting a player with a zero-effect splash potion will remove the existing effect - I think this may need a different function to CheckGreaterAndAddEntityEffect(), unless _that_ behaviour is vanilla as well.

src/Entities/SplashPotionEntity.cpp
tympaniplayer: Ohhh Yes I see. Well intuitive may have nothing to do with it :) I'm a business programmer by day, so game programming is a bit of a challenge for me (in a good way!)

src/Entities/SplashPotionEntity.cpp
tympaniplayer: Ahh I see what you are saying. And I see that it is _not_ the way it is in vanilla. So I think that I need to implement this as suggested. Should it be fixed as part of this, or logged as a separate issue?

tympaniplayer: Ok sorry for the spam. But when testing this behavior out, I don't seem to be getting the effect removed. So although it does look like via the code that is what should happen, but i'm not seeing it. Let me keep testing and playing around

bearbin: I think the right path here is to remove the TODO in this PR, as you've verified the calculation is correct, then open a new issue to fix the other problem, which may be more pervasive (we should check that other places where entity effects are added do not inadvertently remove existing effects).

bearbin:
One last change, please could you add yourself to the CONTRIBUTORS file to recognise your contribution and acceptance of the license terms.

tympaniplayer:
Woops yes! I was in there at one point, but it's been a while so I was probably cleared out. Thanks!

bearbin (approved):
Odd, anyway, this looks good now. Thanks!

Issue #5043, opened 2020-11-16
Chorus fruit and ender pearls should teleport the player off of rideable mobs


PureTryOut:
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: a0afd132330ab538848d5ac263c021c1247fd688

### Expected behavior
Throwing an ender pearl or eating chorus fruit should teleport a player off a rideable mob like a horse or a pig.

### Actual behavior
This doesn't happen, the player stays mounted.

### Steps to reproduce the behavior
Throw an ender pearl while riding a horse.

PR #5179, opened 2021-04-07
Teleport off of mount when throwing enderpearl

tympaniplayer:
Here is a fix for #5043 but only the enderpearl. This doesn't do anything with the chorusfruit problem. My day job is managed code, so there may be some weird c++ smells in here. Please let me know if you see anything like that.

### DIFF
--- a/src/Entities/Player.h
+++ b/src/Entities/Player.h
@@ -604,7 +604,7 @@ public:
 	// cEntity overrides:
 	virtual void AttachTo(cEntity * a_AttachTo) override;
 	virtual void Detach(void) override;
-	virtual void Detach(bool isTeleporting);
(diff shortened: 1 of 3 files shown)


src/Entities/Player.h
```diff
@@ -604,6 +604,7 @@ class cPlayer:
 	// cEntity overrides:
 	virtual void AttachTo(cEntity * a_AttachTo) override;
 	virtual void Detach(void) override;
+	virtual void Detach(bool isTeleporting);
```
bearbin, line 607: This should be `a_IsTeleporting`

bearbin (commented):
Aside from one name, looks good.

bearbin (approved):

PR #5182, opened 2021-04-10
2021-04 Cumulative Quality Rollup for Cuberite

tigerw:
* Fix sending incorrect date values on world change
* Fall particles: check for valid Y
* Fix block break particles
* Streamline player abilities handling
* Update entity sizes
* Fix incorrect name in deserialiser for zombie pigmen
* Remove unused GetClassStatic in cWorld
* Unify multiprotocol entity animations
* Improve bed handling robustness
* Fix thrown snowball/egg hit animations
* Add magical critical effect
* Add animations for shield/item block & break
* More cProtocol cleanup
* cPlayer: move some constants out of header
* Thrown potions/enderpearls: cleanup
* Hoppers: use 'locked' bit in meta
* Implement random ticks more faithfully

### DIFF
--- a/src/Protocol/Protocol_1_9.cpp
+++ b/src/Protocol/Protocol_1_9.cpp
@@ -367,19 +367,6 @@ void cProtocol_1_9_0::SendEntityPosition(const cEntity & a_Entity)
 
 
 
-void cProtocol_1_9_0::SendEntityStatus(const cEntity & a_Entity, char a_Status)
-{
-	ASSERT(m_State == 3);  // In game mode?
-
-	cPacketizer Pkt(*this, pktEntityStatus);
-	Pkt.WriteBEUInt32(a_Entity.GetUniqueID());
-	Pkt.WriteBEInt8(a_Status);
-}
-
-
-
-
-
 void cProtocol_1_9_0::SendExperienceOrb(const cExpOrb & a_ExpOrb)
 {
 	ASSERT(m_State == 3);  // In game mode?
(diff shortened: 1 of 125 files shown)


12xx12:
On it

src/WorldStorage/NamespaceSerializer.cpp
```diff
@@ -501,7 +501,7 @@ eMonsterType NamespaceSerializer::ToMonsterType(const std::string_view a_ID)
 		{ "zoglin",           mtZoglin },
 		{ "zombie",           mtZombie },
 		{ "zombie_horse",     mtZombieHorse },
-		{ "zombie_pigman",    mtZombiePigman },
+		{ "zombified_piglin", mtZombiePigman },
```
12xx12, line 504: this was intended to provide support for old save files. the other version is further up/down

src/Protocol/Protocol_1_9.cpp
```diff
@@ -51,9 +51,11 @@ Implements the 1.9 protocol classes:
...
 
 
-/** Value for main hand in Hand parameter for Protocol 1.9. */
-static const UInt32 MAIN_HAND = 0;
-static const UInt32 OFF_HAND = 1;
+// Value for main hand in Hand parameter for Protocol 1.9.
+#define MAIN_HAND 0
+
+// Value for left hand in MainHand parameter for Protocol 1.9.
+#define LEFT_HAND 0
```
12xx12, line 58: Is this intended to be the same value?

src/WorldStorage/NamespaceSerializer.cpp
tigerw: I see a "PigZombie" mapping. Were we saving as "zombie_pigman" before?

src/Protocol/Protocol_1_9.cpp
tigerw: Yeah, the first is for main/offhand attack, second is for the left handed player/right handed player setting.

src/WorldStorage/NamespaceSerializer.cpp
12xx12: Just in case someone tried to load a map from a version that uses the name

12xx12: So I'd suggest keeping both

tigerw: Done

12xx12 (approved):

12xx12:
I only saw the zombie pigman change in the last update - is that right?

tigerw:
Many thanks, yes, only changed NamespaceSerializer.cpp.

PR #5194, opened 2021-04-17
Added y bounds checking for block place operations.

### LINKED ISSUE #5191: API: cPlayer:PlaceBlock negative y pos crashes the server
Server OS: Linux
Cuberite Commit id: 5ae924ec74bab199284c451552040bcc6f39bb3a

### Actual behavior
Server crashes.

Pass a negative number for y pos:
`player:PlaceBlock(100, -100, 100, 1, 1)`

Kiryu144:
Fixes #5191

### DIFF
--- a/src/Entities/Player.cpp
+++ b/src/Entities/Player.cpp
@@ -2333,15 +2333,6 @@ void cPlayer::LoadRank(void)
 
 
 
-bool cPlayer::PlaceBlock(const Vector3i a_Position, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta)
-{
-	return PlaceBlocks({ { a_Position, a_BlockType, a_BlockMeta } });
-}
-
-
-
-
-
 void cPlayer::SendBlocksAround(int a_BlockX, int a_BlockY, int a_BlockZ, int a_Range)
 {
 	// Collect the coords of all the blocks to send:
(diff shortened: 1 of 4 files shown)


src/Entities/Player.cpp
```diff
@@ -789,8 +788,7 @@ void cPlayer::SetBedPos(const Vector3i & a_Pos)
 void cPlayer::SetBedPos(const Vector3i & a_Pos, cWorld * a_World)
 {
 	m_LastBedPos = a_Pos;
-	ASSERT(a_World != nullptr);
-	m_SpawnWorldName = a_World->GetName();
+	m_SpawnWorldName = a_World ? a_World->GetName() : m_World->GetName();
```
Seadragon91, line 791: I don't think that's a good idea. Better would be to change `cWorld * a_World` to `cWorld & a_World`. On lua side if a nil is then passed, it will abort the plugin and shows a stack trace in console instead of a crash of the server.

Kiryu144: Oh well, I actually didn't want this commit to be in this pr. But thanks, I'll have a look into your suggestion!

Seadragon91: Np. You can create a own pr for that.

12xx12:
Hi and welcome to the project. For copyright reasons you need to add yourself to the CONTRIBUTORS file. (And be mentioned in the newsletter when it is released)

bearbin:
@Kiryu144 Just giving you a ping, there are still a few things that need fixing before this can be merged in.

bearbin (approved):

12xx12:
I actually didn't check if this fixes all possible crashes

bearbin:
If it fixes some, still worth merging in :)

tigerw:
these checks should be moved to ManualBindings_World.cpp, the C++ side already does checking at the points of call so cChunkMap omits checks for performance.

12xx12:
The functions in the manual exports of `cWorld` have checks in place.
The functions in question breaking the server are not manually exported. So this seems the best place to check

tigerw:
I've added the manual export.

PR #5199, opened 2021-04-20
Fixing grass not generating below trees

### LINKED ISSUE #5135: Grass is not generated below trees
Client version: x.x.x
Cuberite Commit id: (Found at the top of the server's console output)

### Expected behavior
Grass below trees

### Actual behavior
No grass

I think this is related that the height map includes trees but the generator checks for the block at the top of the height map which is in case of a tree not grass so doesn't plant grass

12xx12:
Fixes #5135

### DIFF
--- a/src/Generating/FinishGen.cpp
+++ b/src/Generating/FinishGen.cpp
@@ -542,6 +542,21 @@ void cFinishGenTallGrass::GenFinish(cChunkDesc & a_ChunkDesc)
 				continue;
 			}
 
+			// Walk below trees:
+			auto BlockBelow = a_ChunkDesc.GetBlockType(x, y - 1, z);
+
+			while (
+				(BlockBelow == E_BLOCK_LEAVES) ||
+				(BlockBelow == E_BLOCK_NEW_LEAVES) ||
+				(BlockBelow == E_BLOCK_LOG) ||
+				(BlockBelow == E_BLOCK_NEW_LOG) ||
+				(BlockBelow == E_BLOCK_AIR)
+			)
+			{
+				y--;
+				BlockBelow = a_ChunkDesc.GetBlockType(x, y - 1, z);
+			}
+
 			// Check if long grass can be placed:
 			if (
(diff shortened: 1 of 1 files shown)


src/Generating/FinishGen.cpp
```diff
@@ -542,6 +542,21 @@ void cFinishGenTallGrass::GenFinish(cChunkDesc & a_ChunkDesc)
...
+			while (
+				(BlockBelow == E_BLOCK_LEAVES) ||
+				(BlockBelow == E_BLOCK_NEW_LEAVES) ||
+				(BlockBelow == E_BLOCK_LOG) ||
+				(BlockBelow == E_BLOCK_NEW_LOG) ||
+				(BlockBelow == E_BLOCK_AIR)
+			)
+			{
+				y--;
+				BlockBelow = a_ChunkDesc.GetBlockType(x, y - 1, z);
```
bearbin, line 557: This needs a bounds check, at least.

bearbin (changes requested):

src/Generating/FinishGen.cpp
12xx12: yeah - you're right

bearbin (approved):

tigerw:
can we re-order the finishers so the grass generator runs before trees are planted?

12xx12:
that would work too

bearbin:
I can imagine there would be weird edge cases doing it the other way too, any particular reason not to do it this way @tigerw?

12xx12:
Negative:
I am not sure what happens if the generator tries to place a tree where already grass is

Positive:
Less performance impact

tigerw:
Performance mainly, ideally we'd reduce Y-major searching because the stride is 256 elements.

bearbin:
I suggest we merge this unless somebody wants to do a performance test then.

tigerw:
OK, I can investigate reordering later then.

PR #5200, opened 2021-04-20
Disables the weather packet for 1.13

12xx12:
I didn't find any documentation of any explicit weather packet. When there is downfall in the world and a 1.13 client wants to join i will get disconnected because the client can't interpret the packet.


bearbin (approved):
Fine. Let's get this merged in.

tigerw:
SendWeather is a logical packet that maps to ChangeGameState, seems like we forgot to update the packet ID for pktWeather to 0x20 in 1.13.

PR #5201, opened 2021-04-20
Make Windows go tictictick, not tick. tick. tick.

### LINKED ISSUE #5140: Low TPS on Windows because chrono sleeps too long
Client version: 1.12.2
Server OS: Windows 10 Pro 19042.844 x64

Good day.

After installing a clean server for Windows x64, I found that the TPS does not rise above 16. I tried to build the server from the source code (release version), but I got exactly the same result. Reducing the view distance or disabling monsters and animals did not change the situation.

Screen

The server ran on two computers:
```
Ryzen 5 2600 6C12T 3.8GHz
32GB RAM
NVME SSD

Intel core i5 6500 4C
8GB RAM
SSD
```
Unfortunately, there is no time to check the work on Linux yet.

The following message appears in the server logs when a player spawns, but I'm not sure if it can somehow relate to the problem.
```
[23:05:20] cChunkMap::AddEntity: Entity at 0x2a970282010 (cPlayer, ID 55) spawning in a non-existent chunk.
```

tigerw:
* Fixes #5140

Considered alternatives to increasing system interrupt tick rate:
* timeSetEvent: high accuracy and precision but only when telling it to call a function periodically from its own internal thread. Won't scale to multiple worlds.
* Use waitable timers with CREATE_WAITABLE_TIMER_HIGH_RESOLUTION flag: only available on fairly new version of Windows.
* Busywaiting: would work but we probably don't need _that_ high accuracy.
* PID-esque, dynamic waiting algorithm, ensuring _on average_ we have 20TPS: kinda works, but necessarily meant some ticks waiting less than 50ms (>20TPS) to make the mean time 20. But we have an obligation to plugins that TPS <= 20 in HOOK_WORLD_TICK.

Tradeoff for using `timeBeginPeriod`: changes system-wide tick rate in older versions of Windows, unknown impact on power consumption and performance (more scheduler invocations & overhead?).

Mitigations: apparently really popular function, used in lots of software, so follow the crowd amirite?


bearbin:
The other refactoring caused the build to fail :|

12xx12:
You mean make windows go brrrr 🤦

bearbin:
Well, the windows change looks good. Some of the other refactoring looks fine, some I can't make head nor tail of - if it works properly on all platforms then I guess it's fine to merge in.

tigerw:
I cut pasted a bunch of windows-specific stuff from main.cpp into OSSupport/

PR #5204, opened 2021-04-22
Add HOOK_ENTITY_MOVING

12xx12:
This PR takes over #4190 - I can't push to the original PR.
I think i got the documentation


bearbin (approved):

tigerw:
This is a notification only hook, wouldn't a cancellable one be more useful? Possibly even one that generalises to all entities, along the lines of HOOK_ENTITY_MOVING

tigerw:
It also seems like this will be called every tick, whether the monster moves or not.

bearbin:
Yes, it is effectively a onMonsterTick hook, but since we broadcast the movement packet every tick anyway... it does match up :P

tigerw:
Understood, however BroadcastMovementUpdate does check for zero speed and avoid sending if so.

12xx12:
So new changes are 
- [ ] do this for any entity.
Is it possible to cast to monster (or other) in the lua API or is this covered by the nature of lua not being typed language?
- [ ] verif that the hook is actually only triggered if we move a distance > 0

12xx12:
The hook is now cancelable and denies any motion for the entity which tries to move. 
There are some movements that shouldn't be cancelable so there is a extra parameter for the `SetPosition`- function that skips the hook (used in loading and spawning e.g)

bearbin (approved):

12xx12:
WAIT - i forgot something

bearbin:
I'm not merging, Tiger still has a pending review. You can relax.

tigerw:
I think this hook will have too much of an impact on tick performance. Plugins using it have no choice but to subscribe to every entity on the server (there may be many) with processing done in Lua. Since SetPosition is almost free right now it's called everywhere.

pwnorbitals:
@tigerw do you suggest we completely drop&close this PR or would it be worth to pursue modifications to keep the hook and reduce the impact on tick perf ?

madmaxoft:
I'm with @tigerw here, I don't think this is the right way to handle this. Pretty much every entity is moving most of the time, so the perf impact would be large, with very little benefit.

What is the original usecase for needing this functionality? Can we implement it in another way? Do we really need a hook to be called for every entity on every tick?

PR #5215, opened 2021-04-29
Chest, weather, crash, and miscellaneous fixes

tigerw:


src/BlockEntities/ChestEntity.h
```diff
@@ -62,11 +69,9 @@ class cChestEntity :
 	/** Opens a new chest window where this is the primary chest and any neighbour is the secondary. */
 	void OpenNewWindow();
```
bearbin: This comment could be clarified - specifying that this is only _allowed_ in that condition, and that chest order will not somehow be switched around.

bearbin (approved):
Looks good :)

12xx12 (approved):

tigerw:
Thanks all!

tigerw:
Since water is a polar molecule, the streets are the south pole and the clouds are the north pole, hence the water on the street pulls the water out of the clouds by the principle of magnetism, so it rains! /r/TILilearned

PR #5230, opened 2021-06-05
Migrate off CircleCI to Jenkins

bearbin:
First step in moving CI off Travis and centralising it in one place.

Ideally all checks would be on builds.cuberite.org, but that will require improving the situation around build slaves - I need to spin up some more capacity for linux and we don't have any windows slaves at the moment so AppVeyor will have to stay for now.

For now this is an improvement because we control the infrastructure, so there's no chance of getting cut off, and this prevents contributors being required to log in to CircleCI to see their test failures.


bearbin:
And yes, I will fix the logo on the check dashboard so it's not my icon but the cuberite one instead, but I need to make a new GH account for that as my existing bearbot got eaten.

12xx12:
> Does it:
> 
>     * run docker?

It does now

>     * have a reasonable internet connection (i.e. 50/50 or better)?

Don't actually know - it's rented in a proffessional hosting service

>     * have at least 2 GB and 2 cores spare for the build operations?

2 GB Ram but only 1 core

> If yes and it isn't going to go down in a month, it would be helpful :)

I think the voucher lasts for 9 or 10 months and then i start to pay.

bearbin:
Once CircleCI is gone, can we retire the [docker-build-image](https://github.com/cuberite/docker-build-image), or is it used for anything else?

bearbin:
And @12xx12 that sounds good - I contacted you on Discord so that can get moved forwards.

Janmm14:
Did you consider github workflows?

bearbin:
GitHub actions is an option, but as we already have the buildserver and can't move off it for distribution builds, consolidating the continuous integration builds on the same system makes sense. That way we will have one, rather than two sets of maintenance to deal with.

12xx12:
For the appveyor replacement we would need a vm running windows, right?

bearbin:
Indeed, that's right. But for security considerations that would ideally be dedicated just to building cuberite, since we're distributing the binaries that would come off of that.

12xx12:
The cheapest I found were around 6€ (including tax) or 7$(here i'm not sure) per month

bearbin:
I will be able to fund that if we decide it's necessary to move off AppVeyor. But that can come later, let's get the first stage done first.

12xx12:
No worries just wanted to put a number out there for the future

12xx12:
I took the liberty to update the regex searching for files to inspect only files from the root src directory (not the lib dir)

bearbin:
LGTM?

12xx12:
Seems good. What happens to the old checks if we merge? In this PR they Report as error

12xx12:
Oh and do we need to update the README.md file with the Jenkins status?

bearbin:
I have updated the README file, and the old checks have to be manually removed after this is merged - they fail on this branch because the circleCI config files have been deleted but the webhook not yet disabled.

12xx12 (approved):

PR #5233, opened 2021-06-06
Fixed crash when opening inventory in older version

12xx12:


bearbin (approved):

bearbin (changes requested):
Actually, this switch statement looks totally messed up, I don't think this is the fix at all and the rather the IsStatisticPresents need to be switched around so they match up with their Cases again.

12xx12:
The switch statement just checks if you actually can achieve the achievements. I can't quite see what the error might be

bearbin:
Oh, yes I understand what it does now. Yes this is the right fix :facepalm:

bearbin (approved):

PR #5244, opened 2021-06-20
Added NamespaceSerializer::Prettify method to get player-friendly names of mobs.

TheHyper45:
Additionally cEntity::BroadcastDeathMessage method now displays an appropriate death message when a tamed pet is killed by another mob (handles custom names).

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -164,8 +164,8 @@ function(set_exe_flags TARGET)
 			-Wno-switch-enum
 
 			# Weverything with Clang exceptions:
-			-Weverything -Wno-exit-time-destructors -Wno-error=disabled-macro-expansion
-			-Wno-weak-vtables -Wno-string-conversion -Wno-c++98-compat-pedantic -Wno-documentation
+			-Weverything -Wno-exit-time-destructors -Wno-error=disabled-macro-expansion -Wno-weak-vtables
+			-Wno-string-conversion -Wno-c++98-compat-pedantic -Wno-c++2a-compat-pedantic -Wno-documentation
 			-Wno-documentation-unknown-command -Wno-reserved-id-macro -Wno-error=unused-command-line-argument
 		)
 


src/WorldStorage/NamespaceSerializer.cpp
```diff
@@ -546,3 +546,87 @@ eMonsterType NamespaceSerializer::ToMonsterType(const std::string_view a_ID)
 
 	return MonsterTypes.at(a_ID);
 }
+
+
+
+
+
+std::string_view NamespaceSerializer::Prettify(const eMonsterType a_ID, const bool a_IsTamed)
```
12xx12, line 554: Actually, this meant more universally. The runtime is great, but I meant you provide `some_string` and get `Some String`

12xx12: And yes, this will print ocelot for tamed cats. But this should be fixable by implementing proper cat<->ocelot implementation

TheHyper45: So you meant that this function should take a string and prettify that input string and not that it constructs a string from a mob ID?

12xx12: Yes 👍

12xx12: That was not that clear from me - sorry

DarkoGNU:
I've tested this in-game (just to be sure). Death messages work correctly 😁

src/Entities/Entity.cpp
```diff
@@ -2372,12 +2364,22 @@ void cEntity::BroadcastDeathMessage(TakeDamageInfo & a_TDI)
 		{
 			GetWorld()->BroadcastChatDeath(DeathMessage);
 		}
-	}
-	else
+	}  // This will trigger if a player / tamed pet has been killed by another mob / tamed pet.
+	else if (a_TDI.Attacker->IsMob())
```
bearbin, line 2368: Should we handle non-mob death attributions - is such a thing possible?

src/WorldStorage/NamespaceSerializer.h
```diff
@@ -23,4 +23,7 @@ namespace NamespaceSerializer
 	eMonsterType ToMonsterType(std::string_view a_ID);
 
 	std::pair<Namespace, std::string_view> SplitNamespacedID(std::string_view ID);
+
+	// Examples: Input: "wolf" -> Output: "Wolf", Input: "iron_golem" -> Output: "Iron Golem"
+	AString Prettify(AString a_Name, const bool a_IsTamed = false);
```
bearbin, line 28: This name needs to be more descriptive. Maybe `PrettifyEntityName`?

12xx12: This can be useable for any type of enum represented in the file.

this was at least my thought

bearbin: with an IsTamed parameter? No thanks! Rename to PrettifyEntityName, with IsTamed, and move the core logic to a new Prettify function (also exported, without IsTamed) would be a much better solution.

tigerw:
wait, if we're rewriting this we should use the translation keys: https://pokechu22.github.io/Burger/1.12.2.html#language:death

src/WorldStorage/NamespaceSerializer.h
12xx12: Jup 👍

src/Entities/Entity.cpp
12xx12: Can you given an example?

bearbin: I was just trying to make the point that the previous code handled the case where !Attacker->IsMob(), and this code doesn't. Should we?

12xx12: I'm not sure if there is anything left. Dame from no one is handled, damage from player is handled and damage from mobs. I don't know of any case left

12xx12:
How would one do such a thing?

tigerw:
If we send this JSON string to the client

```json
[{"translate": "death.attack.explosion.player", "with": [{"translate": "entity.Creeper.name"}, "12xx12"]}]
```

It outputs "Creeper was blown up by 12xx12".

cCompositeChat may need some modification but hopefully we can use it as:

```cpp
cCompositeChat DeathMessage;

switch (DeathType)
case dtExplosion:
  DeathMessage.AddComponent(
    TranslationComponent(
      TranslationMessage::DeathAttackExplosionPlayer, {
        TranslationComponent(TranslationMessage::EntityCreeperName),
        StringComponent("12xx12")
      }
    )
  )
```

No hardcoded english-language strings needed in the server, client handles it all.

12xx12:
Does this translation key change at any point? This seems kinda sus to me using some hardcoded stuff for all versions. 

See particle effects (I think)

12xx12:
Hm... This is too much for this PR (I think) - but good point

12xx12:
I added the changes to the PR and merged master

bearbin (approved):

bearbin:
@TheHyper45 Why did you delete this pr?

DarkoGNU:
@bearbin Well, I asked my old friend to help me with my farmland trampling implementation, but @TheHyper45 is a little inexperienced with git aaaand... he made some mistakes

I think he'll be happy to restore the deleted code. I gave him some suggestions about how to do it, but he needed to do some other stuff at that moment. @bearbin I think you'll even more experienced with git than me, so you can suggest us the best way to restore that code

bearbin:
That's good, I was just waiting for another review before merging, I got a bit worried that TheHyper45 might have thought we abandoned this and closed it because of that... 

The issue here is that all the work's been done in the master branch rather than a feature branch - it's possible to reset your own master branch but that's also a bit of a pain. 

Easiest way to get the code back is to do a `git checkout -b prettify && git reset --hard 239928297412a19f994aa8a3537a66ab7f9e29d6` which will create a new prettify branch and restore the old code. Then you just need to push the branch and make a new PR. Unfortunately github doesn't support switching the source branch of a PR so you will have to create a new one :(

DarkoGNU:
> The issue here is that all the work's been done in the master branch rather than a feature branch

I think I've made the same mistake. What about [Pull Request Retargeting](https://github.blog/changelog/2020-05-19-pull-request-retargeting/)?

Will any changes to my master affect the commit history in my pull request (after it's been merged) though? Because if not, that's not really an issue

bearbin:
Unfortunately pull request regtargeting is only relevant to the target branch of the PR rather than the source, which is the issue here. 

After a PR is merged, it doesn't matter if you reset the branch it was sourced from - using master is just a pain when you're doing multiple PRs at once, or want to use master for its original purpose as a main branch which you maintain in your fork to follow upstream.

PR #5246, opened 2021-06-21
Performance improvements when spawning many entities

12xx12:
I put cuberite in the VS performance profiler and spawned a lot of animals and took a look what makes the server slow down.
In the end, it's cause by sampling the world around the entity.

In the end, the bottleneck is that we have to find the chunk to ask for the blocks.

There is a double call to `GetBlock` and `GetMeta` which can be removed so that there is one less search each call. -> 53ce8b7

And I replaced the `std::map` in `cChunkMap` with a `unordered_map`. The map has a time of `O(log(n))` in any case. And the `unordered_map` has an armotised runtime of `O(1)` in good case (good hashing). And `O(n)` in worst case (bad hashing). I took a function for hashing that is a Bijection to decrease the collision to a little amount (or 0) so we get a time close to `O(1)`

The downside of this is that `unordered_map` uses more memory!

https://www.geeksforgeeks.org/map-vs-unordered_map-c/

Subjectively, this improved the performance. I can do some actual measurements.

I also thought about copying the part around the pawn into a `cBlockArea`. But I'll have to check the performance of that.

What are your thoughts?

### DIFF
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -97,6 +97,22 @@ public:
 
 
 
+/** Used to hash ChunkCoors in cChunkMap */
+class cChunkChoordHasher
+{
+  public:
+	// Use sum of lengths of first and last names
+	// as hash function.
+	size_t operator()(const cChunkCoords & a_Coords) const
+	{
+		// https://stackoverflow.com/questions/22826326/good-hashcode-function-for-2d-coordinates
+		size_t Temp = (a_Coords.m_ChunkZ + ((a_Coords.m_ChunkX + 1) / 2));
+		return a_Coords.m_ChunkX + (Temp * Temp);
+	}
+};
+
+
+
 
 /** Constants used throughout the code, useful typedefs and utility functions */
 class cChunkDef
--- a/src/ChunkDef.h
+++ b/src/ChunkDef.h
@@ -97,23 +97,6 @@ public:
 
 
 
-/** Used to hash ChunkCoors in cChunkMap */
-class cChunkChoordHasher
-{
-  public:
-	// Use sum of lengths of first and last names
-	// as hash function.
-	size_t operator()(const cChunkCoords & a_Coords) const
-	{
-		// https://stackoverflow.com/questions/22826326/good-hashcode-function-for-2d-coordinates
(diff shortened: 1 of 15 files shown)


bearbin:
I will need to look more into the change of map -> unordered map, but how about just getting the chunk once and using UnboundedRelGetBlock, which should be much faster than getting a chunk 10 times? That way the chunkmap doesn't need to be queried.

In fact, there's not any need to use the chunkmap at all, since Tick has a Chunk, and that could easily be passed in to HandleFalling.

12xx12:
That's a way better solution. And if the block isn't in the same chunk we ignore it.

Still we may think about the transition. We don't need any order and theoretically this should be faster.

mathiascode:
There seems to be a comment on line 301 in ChunkMap.h, added by @tigerw in https://github.com/cuberite/cuberite/pull/2565:
`Uses a map (as opposed to unordered_map) because sorted maps are apparently faster.`

bearbin:
> And if the block isn't in the same chunk we ignore it.

UnboundedRel functions automatically chase the chunk pointers through neighbours, so they'll get the block quickly, even if it's in a different chunk!

12xx12:
I did a test with 2k cows. On Release it's smooth as butter. On debug not :(

12xx12:
Research will be done

12xx12:
I will also check all tick methods for direct calls to World.GetBlock() and so on and replace them with the GetRelUnbound methods

12xx12:
The changes actually don't have many effects on the release and Debug build.
The Many Mobs -> LAG correlation only happens badly on the Debug build.

Is the Debug build that bad when looking at performance?

bearbin:
This is still a draft right, until the other chunkmap-elimination changes have been made, and a better analysis of the performance impacts carried out?

tigerw:
Also don't forget the chunk accesors take chunk-relative coordinates instead of absolute world coordinates. I'd love to do a comparison to see exactly how much overhead our world-to-chunk lookup takes versus following chunk neighbour pointers.

12xx12:
I think the most overhead was generated by the Debug-Features. At some point I'll take a look at the build with debug symbols only

Issue #4531, opened 2020-03-22
Chunks are not sent to clients at times


mathiascode:
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 00ae9604e11e2f0859fb219029c903662fe27bfd

### Expected behavior
Chunks are always visible for clients.

### Actual behavior
Chunks aren't always sent to clients, rendering block textures invisible.

### Steps to reproduce the behavior
Happens to random chunks when traveling in a world sometimes. A consistent way to reproduce is to teleport to a location, e.g. /tp 55555 64 33333, and immediately teleport somewhere else, e.g. /spawn.

mathiascode:
If you teleport quick enough, chunks sometimes don't exist, and the player is stuck in a frozen state until relogging.

mathiascode:
Player freezing is done intentionally here:
https://github.com/cuberite/cuberite/blob/00ae9604e11e2f0859fb219029c903662fe27bfd/src/Entities/Player.cpp#L414-L417

tigerw:
Copied from https://github.com/cuberite/cuberite/issues/5245#issuecomment-867948533

> Having fully loaded the chunks in your area, when teleporting to a faraway location and immediately back, the server doesn't send chunks anymore. This is because after chunks are fully loaded, an internal value is set to your current chunk to stop further chunk streaming. Teleporting away and back unloads every chunk clientside, but because that internal value is unchanged, chunk streaming remains stopped.

To fix this m_LastSentChunk is reset to a bogus value whenever the player's current chunk doesn't match.

Issue #5245, opened 2021-06-20
When you teleport to others in creative who are far away, they seem invisible until you rejoin.


inphasebrick274:
Client version: 1.12.2
Server OS: /Linux/    - done on a raspberry pi 4b
Cuberite Commit id: 81e299f00ce3df587166ccc94f5997555cce5210

### Expected behavior

When you teleport to people in creative, you can see them.

### Actual behavior

When you teleport to people in creative, they appear invisible.

### Steps to reproduce the behavior

1. Have at least 2 people on the server with authentication on
2. Go far away from the other player, and then proceed to teleport to them using /tp
3. You will notice that you cannot see the other player, despite them being there and being able to place blocks

### Server log
This issue is relatively reproduceable, and does not have any abnormalities in the server log.
```

Relevant part of server log
```

tigerw:
When you yourself are in creative mode, or the other person in creative mode? The behaviour doesn't manifest when both are in survival?

inphasebrick274:
When I was testing it with a friend specifically, we were both in creative mode, but I am not sure about survival, as I have not yet tested that, but according to someone else on the forum who seems to be having the same issue "I am on the latest build and when I teleport to others in the creative gamemode, I appear invisible to other players (or that is they have told me on numerous occasions). Sometimes this behavior can be reproduced, whereas other times they can see me.  Changing gamemodes doesn't seem to have any effect on the matter."

inphasebrick274:
I'll attempt to reproduce the issue in survival mode or something else

tigerw:
I've found two problems.

1. Having fully loaded the chunks in your area, when teleporting to a faraway location and immediately back, the server doesn't send chunks anymore. This is because after chunks are fully loaded, an internal value is set to your current chunk to stop further chunk streaming. Teleporting away and back unloads every chunk clientside, but because that internal value is unchanged, chunk streaming remains stopped.
2. This issue's cause. We're sending old position values when a teleporting player appears for another player. The old position value is where the teleporting player came from, and may be in an unloaded chunk. Hence invisibility, because the client doesn't like entities being spawned in unloaded chunks, even if we later on update their position.

PR #5254, opened 2021-06-28
Chunk streaming improvements

tigerw:


12xx12:
Is this the fix for invisible mobs?

tigerw:
For teleporting to a faraway land and back again breaking chunk sending. Hopefully fixes #4531. I'd also like to work on the other chunk related issues #5245 and #3696.

tigerw:
Copied from https://github.com/cuberite/cuberite/issues/5245#issuecomment-867948533

> Having fully loaded the chunks in your area, when teleporting to a faraway location and immediately back, the server doesn't send chunks anymore. This is because after chunks are fully loaded, an internal value is set to your current chunk to stop further chunk streaming. Teleporting away and back unloads every chunk clientside, but because that internal value is unchanged, chunk streaming remains stopped.

To fix this m_LastSentChunk is reset to a bogus value whenever the player's current chunk doesn't match.

12xx12:
The protoproxy tool is broken on compile - I really don't know if this good working with the broken API

12xx12:
I will take a look at this

src/ClientHandle.cpp
```diff
@@ -388,25 +392,30 @@ void cClientHandle::FinishAuthenticate(const AString & a_Name, const cUUID & a_U
...
 	if ((m_LastStreamedChunkX == ChunkPosX) && (m_LastStreamedChunkZ == ChunkPosZ))
 	{
 		// All chunks are already loaded. Abort loading.
-		return true;
+		return;
+	}
+	else
+	{
+		m_LastStreamedChunkX = 0x7fffffff;
+		m_LastStreamedChunkZ = 0x7fffffff;
```
12xx12, line 410: Why those magic numbers?

src/mbedTLS++/AesCfb128Encryptor.cpp
```diff
@@ -43,8 +43,8 @@ void cAesCfb128Encryptor::Init(const Byte a_Key[16], const Byte a_IV[16])
 
 
 
-void cAesCfb128Encryptor::ProcessData(std::byte * const a_EncryptedOut, const std::byte * const a_PlainIn, size_t a_Length)
+void cAesCfb128Encryptor::ProcessData(std::byte * const a_PlainIn, const size_t a_Length)
 {
 	ASSERT(IsValid());  // Must Init() first
-	mbedtls_aes_crypt_cfb8(&m_Aes, MBEDTLS_AES_ENCRYPT, a_Length, m_IV, reinterpret_cast<const unsigned char *>(a_PlainIn), reinterpret_cast<unsigned char *>(a_EncryptedOut));
+	mbedtls_aes_crypt_cfb8(&m_Aes, MBEDTLS_AES_ENCRYPT, a_Length, m_IV, reinterpret_cast<const unsigned char *>(a_PlainIn), reinterpret_cast<unsigned char *>(a_PlainIn));
```
12xx12, line 49: Now we do inplace encryption?

12xx12 (commented):
So to summarize:
- we now encrypt sent data in place
- you rephrased some code that triggers the chunk sending
right?
- we now can process the data sent by a protocol before sending (e.g encrypt) instead of hiding this somewhere (I like this)

src/mbedTLS++/AesCfb128Encryptor.cpp
tigerw: Yes, on the protocol out buffer.

tigerw:
Yes. For the third, if you mean additional processing by plugins, in theory but it's binary data so not too useful.

12xx12:
Or disabling decryption for testing purposes - the hardcore way

src/ClientHandle.cpp
12xx12: I see the sense of those values, but an explanation would be nice. or a `#define INSANEVALUE 0x7fffffff`

12xx12: Now I see the big change - you invalidate the last sent chunk if we're not in it 💡

12xx12 (approved):

12xx12:
Note that some compiling is still broken

tigerw:
Thanks, let me change the constant.

12xx12:
Looks good to me

12xx12:
Is there a reason you don't squash your PRs?

I think it's nicer to have a commit for each PR and have a number attached to it to check the thoughts by the developer.

In the end, it's just convenience. So it's not that important

Janmm14:
you could use a merge commit with the PR number

12xx12:
I usually just push the big green button and GitHub does everything else :D

tigerw:
Yeah you're right, this could've been squashed

12xx12:
There is always a next PR to merge 😄

PR #5258, opened 2021-07-08
Item frame maps

### LINKED ISSUE #2596: Can not see maps in itemframes
hi,
if you have maps in item frames, you cant see it, after a login

bug

pick it up and have a look at one of this maps,
only after that you can see it in the frames too!

maybe the server did not send it to clients?

hope my 1st ever issue is correct :)

KingCol13:
Contains a few static_casts to fix compiling with `-werror`.

Fixes #2596.

While I was working on that I noticed the item frames sometimes didn't save the rotation or item held across server restarts, marking the chunk dirty seemed to fix this.


12xx12 (commented):
Thanks 👍🏻

Looks good - I'm gonna check gameplay and the approve

And yeah - less compile warnings

12xx12 (approved):

PR #5270, opened 2021-07-20
Setting command-block's command requires a permission.

madmaxoft:
Fixes a possible security issue reported through the support email.


12xx12 (approved):

bearbin (approved):
This looks good - do we have docs forthe commandblock.set permission?

madmaxoft:
There's no docs, I don't know where to put it. Any ideas?
Anyway, it will mostly just work(tm), since the only people who are supposed to be able to edit commandblocks are ops, who have all permissions anyway.

bearbin:
Yeah, probably file that one on the later list. LGTM :)

12xx12:
later(tm)

PR #5276, opened 2021-07-26
Paintings tweaks

Rorkh:
Now they are creates sounds they should at spawn and destroy and they can't be critical hitted (they skip default on damage logic cause amount of damage should not be checked for them as critical hits and damage dealt to them shouldn't be registered in statistics (according to Vanilla server behavior)).
Little checklist:

- [x] Sounds when placed and broke
- [x] No critical damage
- [ ] Check if painting fits the surface

### DIFF
--- a/src/Entities/Painting.cpp
+++ b/src/Entities/Painting.cpp
@@ -24,6 +24,8 @@ void cPainting::SpawnOn(cClientHandle & a_Client)
 {
 	Super::SpawnOn(a_Client);
 	a_Client.SendPaintingSpawn(*this);
+	
+	m_World->BroadcastSoundEffect("entity.painting.place", GetPosition(), 1, 1);
 }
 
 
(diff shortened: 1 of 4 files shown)


src/Entities/Painting.cpp
```diff
@@ -41,3 +43,11 @@ void cPainting::GetDrops(cItems & a_Items, cEntity * a_Killer)
 
 
 
+
+void cPainting::KilledBy(TakeDamageInfo & a_TDI)
+{
+	Super::KilledBy(a_TDI);
+	Destroy();
+
+	m_World->BroadcastSoundEffect("entity.painting.break", GetPosition(), 1, 1);
```
bearbin, line 52: Destroy should be moved to the end of this function, as it frees the memory used by the entity and so calling GetPosition is a use-after-free.

bearbin (changes requested):

bearbin (commented):
This looks good. It's not clear from your checklist whether you want to implement the additional behaviour in this PR or in a later one. Either is fine, but if you do want to continue working on this, it would be best to mark this PR as a draft.

Another point - you need to add yourself to the CONTRIBUTORS file in this PR before it can be merged, so that you show acceptance of the license, and so that your contribution to the project is properly recognised.

Rorkh:
Yeah, I would like better do this feature like a other PR later especially since I'm stuck a little in it.

bearbin (approved):
LGTM, thanks for the contribution!

PR #5277, opened 2021-07-26
Haybale damage reduction implementation

Rorkh:
I have used data from official Minecraft wiki (https://minecraft.fandom.com/wiki/Hay_Bale) but still it takes 9 block to give 1 heart damage in Vanilla while in my implementation all 13. 
Is it critical and should I take it personally? It seems like a problem of general fall damage calculation, not a haybale modifier ones

### DIFF
--- a/src/Entities/Pawn.cpp
+++ b/src/Entities/Pawn.cpp
@@ -423,19 +423,16 @@ void cPawn::HandleFalling(void)
 				Damage = static_cast<int>(static_cast<float>(Damage) * 0.33);
 			}
 
-			const auto Below = POS_TOINT.addedY(-1);
-			const auto BlockBelow = GetWorld()->GetBlock(Below);
-
-			if (BlockBelow == E_BLOCK_HAY_BALE)
-			{
-				Damage = std::clamp(static_cast<int>(static_cast<float>(Damage) * 0.2), 1, 20);
-			}
-
-			TakeDamage(dtFalling, nullptr, Damage, static_cast<float>(Damage), 0);
-
 			// Fall particles:
-			if (Below.y >= 0)
+			if (const auto Below = POS_TOINT.addedY(-1); Below.y >= 0)
 			{
+				const auto BlockBelow = GetWorld()->GetBlock(Below);
+
+				if (BlockBelow == E_BLOCK_HAY_BALE)
+				{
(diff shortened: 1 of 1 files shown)


src/Entities/Pawn.cpp
```diff
@@ -423,18 +423,26 @@ void cPawn::HandleFalling(void)
...
 			}
 
+			const auto Below = POS_TOINT.addedY(-1);
+			const auto BlockBelow = GetWorld()->GetBlock(Below);
+
+			if (BlockBelow == E_BLOCK_HAY_BALE)
+			{
+				Damage = std::clamp(static_cast<int>(static_cast<float>(Damage) * 0.2), 1, 20);
+			}
+
```
bearbin, line 433: This should be moved inside the `if (Below.y >= 0)` block, to guard the GetBlock.

bearbin (changes requested):
Looks good, excepting one point.

bearbin:
I'm not sure what the ideal thing to do with the fall damage calculation would be - this at least adds the feature, even if the constants are wrong. What is the behaviour at larger heights - does the damage correspond there or does it also differ from vanilla? Best to rule out off-by-one errors before getting into anything more involved.

Rorkh:
> 
> 
> I'm not sure what the ideal thing to do with the fall damage calculation would be - this at least adds the feature, even if the constants are wrong. What is the behaviour at larger heights - does the damage correspond there or does it also differ from vanilla? Best to rule out off-by-one errors before getting into anything more involved.

Made some more tests, in order to spend two hearts it is takes to fall 20 blocks on Vanilla, while 23 in my impl. Considering that it's working with floating values, it is quite reasonable and around expections

src/Entities/Pawn.cpp
Rorkh: > 
> 
> This should be moved inside the `if (Below.y >= 0)` block, to guard the GetBlock.

Should I move TakeDamage in it or move it after if? By practice seems like there is no damage to player applied if there is no particles, but I believe that there was a reason to do as it is now.

bearbin: I would move the TakeDamage to after the if, so the update to Damage by the hay bale check is applied. I don't _think_ that it would do any harm to put it inside the if, but keeping the behaviour the same as it is now (by moving it after) just makes sure that we don't introduce any new bugs.

bearbin (approved):

PR #5281, opened 2021-08-19
Slime spawinig in swamps

### LINKED ISSUE #2233: Slime spawn in light
Slime blocks are spawning in swamp biome during the day.

According to gamepedia: Slimes spawn in the Overworld in specific chunks below layer 40. They can also spawn in swamp biomes between layers 50-70 in light levels of 7 or less. (http://minecraft.gamepedia.com/Slime)

related to #1828 

This is on the first day the world was generated
slime001

JK2Kgit:
Fixes #2233 

Also implements other limitation from wiki (https://minecraft.fandom.com/wiki/Slime)

- height
- moon phases
- 50% chance

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,6 @@ Server/CONTRIBUTORS
 Server/LICENSE
 Server/Licenses
 Server/itemblacklist
-Server/BACKERS
-Server/lua
 Tools/MCADefrag/MCADefrag
 Tools/ProtoProxy/ProtoProxy
 Testing/
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ Server/CONTRIBUTORS
 Server/LICENSE
 Server/Licenses
 Server/itemblacklist
+Server/BACKERS
+Server/lua
 Tools/MCADefrag/MCADefrag
 Tools/ProtoProxy/ProtoProxy
 Testing/
--- a/src/MobSpawner.cpp
+++ b/src/MobSpawner.cpp
@@ -208,8 +208,10 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, Vector3i a_RelPos, eMonsterType
 		case mtMagmaCube:
 		case mtSlime:
 		{
+			const int AMOUNT_MOON_PHASES = 8;
 			auto maxLight = Random.RandInt(0, 7);
-			auto phaseNumber = int(std::floor(a_Chunk->GetWorld()->GetWorldAge().count() / 24000)) % 8;
+			auto moonPhaseNumber = static_cast<int>(std::floor(a_Chunk->GetWorld()->GetWorldAge().count() / 24000)) % AMOUNT_MOON_PHASES;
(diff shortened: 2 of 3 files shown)


bearbin (approved):
LGTM, thanks for your contribution :)

src/MobSpawner.cpp
```diff
@@ -208,12 +208,27 @@ bool cMobSpawner::CanSpawnHere(cChunk * a_Chunk, Vector3i a_RelPos, eMonsterType
...
+					(
+						(a_Biome == biSwampland) &&
+						(a_RelPos.y >= 50) &&
+						(a_RelPos.y <= 70) &&
+						(SkyLight <= maxLight) &&
+						(BlockLight <= maxLight) &&
+						((std::abs(phaseNumber - 4) / 4) > Random.RandReal()) &&
+						(Random.RandBool(0.5))
+					)
+				)
```
NiLSPACE, line 231: If I understood `If the fraction of the moon that is bright is greater than a random number (from 0 to 1)`  correctly I would expect something like this:
```cpp
const float AMOUNT_MOON_PHASES = 8f;
auto moonPhase = static_cast<float>(std::floor(a_Chunk->GetWorld()->GetWorldAge().count() / 24000)) % AMOUNT_MOON_PHASES;
auto threshold = std::abs(static_cast<float>(moonPhase) - (AMOUNT_MOON_PHASES / 2)) / (AMOUNT_MOON_PHASES / 2);

if (GetRandomProvider().RandBool(threshold))
{
	// .. 
}
```

Also, I'm not really sure why `maxLight` is a random value. It seems a slime can spawn at any lightlevel under 7, just not every tick.

bearbin: Is the behaviour in the PR not equivalent to the code you have suggested? It only differs in using `threshold > Random.RandReal()` rather than `Random.RandBool(threshold)`

maxLight being a random value is what's specified in the wiki so unless that's wrong no need to differ. It makes sense that slimes should spawn more often at lower light levels in any case.

NiLSPACE: Ah yes, it's documented in the wiki. Just ignore this comment.

.gitignore
```diff
@@ -9,6 +9,8 @@ Server/CONTRIBUTORS
 Server/LICENSE
 Server/Licenses
 Server/itemblacklist
+Server/BACKERS
+Server/lua
```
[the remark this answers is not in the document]
JK2Kgit: I forgot to don't update .gitignore but
There already is BACKERS in root directory and that one is genrated on compilation or running server I'm not sure
And lua is executable and wasn't included in repo so i just thought that it shouldn't be there
Should I remove these changes or merge with them?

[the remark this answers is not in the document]
JK2Kgit: I will just undo these changes don't nesesary in this case

JK2Kgit:
Ok i think it is done.

NiLSPACE (approved):

PR #5291, opened 2021-09-14
BungeeGuard style proxy security and OnlyAllowBungee in config

### LINKED ISSUE #2432: Add AllowOnlyBungeeCord option in server settings
I think it would be easy and nice to add this option. If true, only players from bungee cord can connect to server :+1: .It would be very great and useful!

LittleBigBug:
This PR adds support for [BungeeGuard](https://www.spigotmc.org/resources/bungeeguard.79601/) style protection (and therefore supports [Velocity](https://Velocitypowered.com)) if the server cannot be hidden behind an SSH tunnel or a LAN. It also adds an option to only allow proxy connections.

I needed this for myself but I thought it would be a useful feature to have normally.
Fixes #2432

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -222,6 +222,28 @@ bool cClientHandle::IsUUIDOnline(const cUUID & a_UUID)
 
 
 
+void cClientHandle::ProxyInit(const AString & a_IPString, const cUUID & a_UUID)
+{
+	this->SetIPString(a_IPString);
+	this->SetUUID(a_UUID);
+
+	this->m_ProxyConnection = true;
+}
+
+
+
+
+
+void cClientHandle::ProxyInit(const AString & a_IPString, const cUUID & a_UUID, const Json::Value & a_Properties)
+{
+	this->SetProperties(a_Properties);
+	this->ProxyInit(a_IPString, a_UUID);
+}
+
+
+
+
+
 void cClientHandle::ProcessProtocolOut()
 {
 	decltype(m_OutgoingData) OutgoingData;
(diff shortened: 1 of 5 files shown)


tigerw:
The problem is that BungeeCord needs auth to be off due to the connection proxy, but this opens up unauthenticated direct connections if the IP gets leaked? Is the `bungeecord-token` field a de facto standard?

Janmm14:
> The problem is that BungeeCord needs auth to be off due to the connection proxy, but this opens up unauthenticated direct connections if the IP gets leaked? Is the `bungeecord-token` field a de facto standard?

It has been adapted in Velocity and Waterfall (Bungee fork) and afaik BungeeGuard is the only plugin with such a functionality.

src/ClientHandle.cpp
```diff
@@ -264,6 +286,58 @@ void cClientHandle::Kick(const AString & a_Reason)
...
+	{
+		for (auto & Node : GetProperties())
+		{
+			if (Node.get("name", "").asString() == "bungeeguard-token")
+			{
+				AString SentToken = Node.get("value", "").asString();
+
+				if (ForwardSecret.compare(SentToken) == 0)
+				{
+					AuthenticatedSecret = true;
```
bearbin, line 317: It looks to me like you could just do a short-circuit `return true` here, and not have to have a AuthenticatedSecret var at all.

src/ClientHandle.cpp
```diff
@@ -264,6 +286,58 @@ void cClientHandle::Kick(const AString & a_Reason)
...
+	{
+		Kick("You can only connect to this server using a Proxy.");
+
+		return false;
+	}
+
+	cServer * Server = cRoot::Get()->GetServer();
+
+	// Proxy Forward Secret Check (BungeeGuard)
+	const AString & ForwardSecret = Server->GetProxyForwardSecret();
```
bearbin, line 301: Is this a standardised name for this secret? It would make much more sense to call this a "ProxySharedSecret" rather than a forward secret - which makes me think the secret is somehow temporal and does not bring to mind forwarding at all.

bearbin (approved):
Looks good with a bit of refactoring.

src/ClientHandle.cpp
```diff
@@ -281,6 +351,11 @@ void cClientHandle::Authenticate(const AString & a_Name, const cUUID & a_UUID, c
 
 		ASSERT(m_Player == nullptr);
 
+		if (!BungeeAuthenticate())
+		{
+			return;
```
tigerw, line 356: Don't forget to Kick.

LittleBigBug: BungeeAuthenticate already does that. Should it be here?

tigerw:
Is there a reason to have OnlyBungeeCord = 1 but no shared secret? If not `!shared secret.empty()` could be equivalent to OnlyBungeeCord.

LittleBigBug:
Maybe if it's a LAN only server. Do you want me to change that?

src/ClientHandle.cpp
tigerw: good point, maybe we can move it out for clarity at the expense of detailed kick messages, but later

tigerw:
not now, maybe later. Thanks!

PR #5298, opened 2021-09-15
Ghast fireball reflection

### LINKED ISSUE #5256: The ghast fireball can't be deflected according to comment
```c++
// TODO: Deflecting the fireballs by arrow- or sword- hits
```

LittleBigBug:
This PR implements Ghast fireball reflections which Fixes #5256. It implements the "Return to Sender" Achievement.

This PR makes some changes to cEntity::TakeDamage which I wasn't sure what the best way was to go about, I needed eExplosionSource available in cEntity::TakeDamageInfo and also to pass the fireball as the attacker. (removing the Pawn check in Entity.cpp#L294)

It doesn't yet work with arrows. I don't know the best approach for that. I attempted an implementation in cGhastFireballEntity::OnHitEntity with nothing working.


tigerw:
Do you forsee any problems with dropping eExplosionSource and relying only on the entity parameter?

LittleBigBug:
I'd still imagine eExplosionSource would be useful to have in TakeDamageInfo for stuff like the Lua API, too. But I don't believe so. It just seems less expensive/safer than checking the cast first.

LittleBigBug:
I remember thinking about how cEntity::TakeDamage could take a class (DamageInfo) argument instead of having many arguments to create the DamageInfo, and fill in the blanks.

If it works for you with just the entity parameter I have no qualms either way.

PR #5300, opened 2021-09-16
Player respawn fix

### LINKED ISSUE #3611: Players spawn in the wrong location after death
Client version:   1.9
Server OS:        Linux
Commit id:        83fa6cb851536839ebd85b1e91fda928f47eb674

### Expected behavior

Players should spawn according to /spawn, or some kind of death spawnpoint (if implemented)

### Actual behavior

Instead, players spawn in the wrong place, also in a place I'd like to keep private.

Kill yourself

JK2Kgit:
Planing to fix #3611 and other issues realted to respawn

- [x] Check bed destruction upon respawn
- [X] Skip bed check when respawnpoint set via `/spawnpoint`
- [X] Lua api update
- [X] Update /spawnpoint command in Core plugin cuberite/Core#244 (megre together)

I have a question i need to aditional information to implent all functinalities
- If it is a deafult world spawn or was it changed
- Is is set up trougth bed or trought comand

Possible soultions
1. Varaible (propably char) to indicate respawn position state with values `0 for unseet spawnpoint`, `1 for force spawnpoint(command)`, `2 for bed` and futuristic `3 for nether anchor`
2. Two boolean variables `m_CheckBedUponRespawn` and `m_IsDeafultSpawnpoint / m_IsChanged`

Changes in files are draft and aren't relevant. This pull request is to mainly ask for which solution is better and in future to ask for help in changes with lua api


JK2Kgit:
Ok i did it i think as it shoud be but i expect that it still require few small changes so i will be glad for opinion tested in game works (i thnk build are failing due to unupdated Core plugin if not then i'm waitin for feadback)

JK2Kgit:
Ok the error are becouse of missing documentation and i have absolutly no idea how to add documentaion so i would be realy gratfull for explaining me how to do it or pointing me to according page/file/anything.

KingCol13:
Hiya, the documentation is for the functions exported to the lua API (between the `// tolua` blocks). The documentation is stored in `Server/Plugins/APIDump/APIDesc.lua`, you can follow the examples there and apply for your new/edited functions.

tigerw:
We should agree on what changes the API needs before you go and write the docs :p

I agree that `SetBedPos` is currently mis-named, what is does is more like `SetRespawnPos`, with no bed-specific checks at all, and Core depends on this behaviour for respawn point override.

However, Vanilla doesn't have the concept of a non-bed respawn point.

I propose that we implement the bed block existence checks into SetBedPos/Respawn and modify the API docs to state this. The stock server will no longer unconditionally teleport the player to `m_BedPos`.

Then, we can separately consider if Core should keep on implementing its non-Vanilla respawn point override `/setspawn`. If so, it should do this by setting the position inside `HOOK_PLAYER_RESPAWNED`.

JK2Kgit:
`/spawnpoint` is vanilla command. Additionally in 1.16 there is nether anchor allowing to set respawn point in nether so vanilla does have a concept of not bed respawn point.

tigerw:
Yes it is, I was wrong about that. In that case an `m_IsRespawnPointForced` bool to control whether the bed/respawn anchor checks happen should be sufficient. SetBedPos automatically sets that to false, a new SetRespawnPosition sets it to true. We can rename `m_BedPos` to `m_RespawnPosition`. The bed checks themselves can go in cPlayer::Respawn. If they fail, it will set `m_RespawnPosition` to world spawn and `m_IsRespawnPointForced` to true. Will this work?

tigerw:
Your solution 2 essentially. I don't see why we need a separate case for unset spawn/default spawn though.

JK2Kgit:
> Yes it is, I was wrong about that. In that case an `m_IsRespawnPointForced` bool to control whether the bed/respawn anchor checks happen should be sufficient. SetBedPos automatically sets that to false, a new SetRespawnPosition sets it to true. We can rename `m_BedPos` to `m_RespawnPosition`. The bed checks themselves can go in cPlayer::Respawn. If they fail, it will set `m_RespawnPosition` to world spawn and `m_IsRespawnPointForced` to true. Will this work?

Yes I think so I will implement and test this one I will have a bit of Time thank for opinion. The only problem I can see is with anchor it have different places for spawning player but that can be with check of block type weather its a bed or anchor and should work fine.

tigerw:
looks good, thanks! just the API docs now, you may want to change `GetRespawnPos` back to `GetBedPos` for now otherwise you'd have to deprecate the old name, it's one less thing to have to do

JK2Kgit:
OK if check are green and there are no grammar or spelling error i believe it is done

JK2Kgit:
Thanks for help with my comments and naming. And also may i ask why you moved message from chat to over toolbar, I believe that vanilla behaviour is chat. I don't mean it is wrong or something i'm just curious.

tigerw:
The vanilla way to send that message is via the [Change Game State](https://wiki.vg/Protocol#Change_Game_State) packet, which is totally weird and I don't think is right. The next best thing is sending as a "game information" chat message, i.e. above action bar, which is also consistent with all other bed interaction messages. So it's different from vanilla, but I propose waiting to see if anyone complains.

tigerw (approved):

tigerw:
Thanks for your work!

JK2Kgit:
> The vanilla way to send that message is via the [Change Game State](https://wiki.vg/Protocol#Change_Game_State) packet, which is totally weird and I don't think is right. The next best thing is sending as a "game information" chat message, i.e. above action bar, which is also consistent with all other bed interaction messages. So it's different from vanilla, but I propose waiting to see if anyone complains.

Thanks for explanation and I agree it is a bit weird.

PR #5301, opened 2021-09-17
Prevent placing of hangables on illegal blocks/Break when support block broken

### LINKED ISSUE #5299: Some Hangables can be placed on blocks they aren't supposed to and won't break when blocks they are on are destroyed.
Client version: 1.12
Server OS: Linux x64
Cuberite Commit id: 11ca2cf4c01426176e0b9f5111aa594afd10983a

Some hangable items like Paintings and Item frames can be placed on items they shouldn't be able to, like ladders and minecart rails.

image
image
image
image

-----

Client version: 1.10.2
Server OS: linux
Cuberite Commit id:  approx: c2f8ceb554982a33bcd4a1e168f6c4e26d0b85dd built at: approx: 2020-09-05 17:13:44 +0200

### Expected behavior
Attachable objects are dropped once the block they were attached to is destroyed

### Actual behavior
Certain objects still stay after the block underneath (or behind) they were attached to is destroyed

LittleBigBug:
This PR fixes #5299 
It also implements the weird second-or-so before the item frame/painting actually breaks. I just use a std::chrono clock to track that.

- [x] Destroy Paintings/Item Frames when there's no or 'illegal' block behind them
- [x] Don't allow hangables to be placed on illegal blocks 
- [x] Vanilla Behaviour

### DIFF
--- a/src/Entities/HangingEntity.cpp
+++ b/src/Entities/HangingEntity.cpp
@@ -2,6 +2,7 @@
 #include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
 
 #include "HangingEntity.h"
+#include "BlockInfo.h"
 #include "Player.h"
 #include "Chunk.h"
 #include "../ClientHandle.h"
@@ -22,6 +23,26 @@ cHangingEntity::cHangingEntity(eEntityType a_EntityType, eBlockFace a_Facing, Ve
 
 
 
+bool cHangingEntity::IsValidSupportBlock(const BLOCKTYPE a_BlockType)
+{
+	return cBlockInfo::IsSolid(a_BlockType) && (a_BlockType != E_BLOCK_REDSTONE_REPEATER_OFF) && (a_BlockType != E_BLOCK_REDSTONE_REPEATER_ON);
+}
+
+
+
+
+
+void cHangingEntity::KilledBy(TakeDamageInfo & a_TDI)
+{
+	Super::KilledBy(a_TDI);
+
+	Destroy();
+}
+
+
+
+
+
 void cHangingEntity::SpawnOn(cClientHandle & a_ClientHandle)
 {
 	SetYaw(GetProtocolFacing() * 90);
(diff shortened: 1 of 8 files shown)


src/Entities/HangingEntity.cpp
```diff
@@ -29,3 +30,48 @@ void cHangingEntity::SpawnOn(cClientHandle & a_ClientHandle)
...
+
+	if (OppositeFacing >= 4)
+	{
+		OppositeFacing -= 4;
+	}
+
+	const auto & Pos = GetPosition();
+	const auto World = a_Chunk.GetWorld();
+	const auto SupportBlockPos = AddFaceDirection(Pos, ProtocolFaceToBlockFace(OppositeFacing));
+	BLOCKTYPE Block = World->GetBlock(SupportBlockPos);
```
bearbin, line 46: The block should be got from the chunk, and the world only fetched later (if required).

bearbin (changes requested):
I'm concerned with the performance of this implementation. Querying blocks for every hangable every tick is rather intensive. Is this actually vanilla behaviour? I would have thought that the delay before paintings are removed is actually because the support is not checked every tick, but rather randomly on say 1/10 of ticks. Please confirm with what vanilla actually does - is it a timer or random when the hangable drops?

src/Entities/HangingEntity.cpp
LittleBigBug: What if the block the entity is hanging on isn't in the same chunk? Should I check that first instead?

LittleBigBug:
I thought the same thing and was going to force-push later. I got distracted with the armor stand branch! I'll do that.

tigerw:
I think the reason for Vanilla's delay when the support is broken is exactly to avoid checking each tick.

LittleBigBug:
Yes, I know. Maybe I should have set this as a draft.

src/Entities/HangingEntity.cpp
bearbin: Ah, if you use the RelGetBlock function it will pointer-chase across chunks so there's no need to do bounds checking or worry about that. It should be much faster than looking up the block in the chunk map of the world.

LittleBigBug: Oh! Great to know, thank you. I'll do that in the next patch

bearbin (approved):

bearbin:
Oops, one more fix needed for Clang compilation.

tigerw:
Thanks @LittleBigBug and @bearbin!

PR #5302, opened 2021-09-18
Mobs patches

Rorkh:
- [x] Implement Sniper Duel achievment
- [x] Only nether-native mobs can see through lava
- [x] Fix blaze shooting with ghast sound

### DIFF
--- a/src/Mobs/Skeleton.cpp
+++ b/src/Mobs/Skeleton.cpp
@@ -68,7 +68,3 @@ void cSkeleton::SpawnOn(cClientHandle & a_ClientHandle)
 	Super::SpawnOn(a_ClientHandle);
 	a_ClientHandle.SendEntityEquipment(*this, 0, cItem(E_ITEM_BOW));
 }
-
-
-
-
--- a/src/Entities/Entity.cpp
+++ b/src/Entities/Entity.cpp
@@ -396,7 +396,7 @@ bool cEntity::DoTakeDamage(TakeDamageInfo & a_TDI)
 
 		if (a_TDI.Attacker != nullptr)
 		{
-			a_TDI.Attacker->Killed(this);
+			a_TDI.Attacker->Killed(*this, a_TDI.DamageType);
 		}
 
 		return true;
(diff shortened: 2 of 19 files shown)


Rorkh:
Oh fuck. I'm dumb, sorry

Rorkh:
I think that so it must be alright. I wanted to separate it in two pulls but i'm sometimes dumb with git

src/Mobs/Skeleton.cpp
```diff
@@ -63,6 +63,28 @@ bool cSkeleton::Attack(std::chrono::milliseconds a_Dt)
...
+		cPlayer * Player = static_cast<cPlayer *>(a_TDI.Attacker);
+
+		int DistX = FloorC(Player->GetPosX()) - FloorC(GetPosX());
+		int DistZ = FloorC(Player->GetPosZ()) - FloorC(GetPosZ());
+
+		if (DistX * DistX + DistZ * DistZ >= 2500)
+		{
+			Player->AwardAchievement(CustomStatistic::AchSnipeSkeleton);
+		}
+	}
```
tigerw, line 81: It could be better to move this to `cPlayer::Killed` where some achievement stuff is already at.
Is flooring the individual components then subtracting, instead of subtracting then flooring, intentional?

tigerw: I looked again and whilst the signature is `void cPlayer::Killed(cEntity * a_Victim)` you don't have access to the DamageType. You could modify the signature to pass it, that way Skeleton.cpp doesn't have to `#include Player.cpp`, but this may be too much trouble. Your choice.

Rorkh: I'd like to remain all on the same places where it is. I think, specific achievements rewards, not general like kill any enemy should be placed where it logically should to avoid additional checks and make it more intuitive.

bearbin (changes requested):
Looks good, subject to fixing the clang-tidy fail on the build. Once that's done this is ready to merge.

tigerw (approved):

tigerw:
Praise the Sun.

PR #5303, opened 2021-09-19
Spectation: add dedicated pathway for spectator mode

### LINKED ISSUE #5296: Exiting first-person view through a mobs eyes in spectator mode (pressing shift) crashes the server
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 11ca2cf4c01426176e0b9f5111aa594afd10983a

### Expected behavior
Exiting first-person view through a mob in spectator mode should just return to normal view and not crash the server

### Actual behavior
Server crashes with SIGSEGV

1. Go in spectator mode
2. Left-click on a mob to enter first-person view of it
3. Press shift to exit it again

### Server log
```
fish: Job 1, './Cuberite' terminated by signal SIGSEGV (Address boundary error)
```

Yup, that's literally all the log there is for this...

tigerw:
+ Sync player rotation with spectated entity.
+ Add dedicated infrastructure to cPlayer for handling spectation, instead of misusing entity riding.
* Avoid infinite recursion when exiting spectation, fixes #5296

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1675,7 +1675,7 @@ void cClientHandle::HandleUseEntity(UInt32 a_TargetEntityID, bool a_IsLeftClick)
 	{
 		m_Player->GetWorld()->DoWithEntityByID(a_TargetEntityID, [=](cEntity & a_Entity)
 		{
-			m_Player->AttachTo(&a_Entity);
+			m_Player->SpectateEntity(&a_Entity);
 			return true;
 		});
 		return;
--- a/src/Entities/Boat.cpp
+++ b/src/Entities/Boat.cpp
@@ -17,17 +17,18 @@
 class cBoatCollisionCallback
 {
 public:
-	cBoatCollisionCallback(cBoat * a_Boat, cEntity * a_Attachee) :
+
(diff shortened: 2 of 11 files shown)


tigerw:
I did a cursory check of all the call sites and I didn’t see any null being passed in. But mainly, if it were null, the line above would have crashed already.

(Cristian moment)

Sent from my Windows 10 Mobile ??

bearbin:
It's a raw pointer that's being dereferenced. We need a check, there's nothing preventing this function being called from somewhere else with null.

tigerw:
`a_AttachTo->m_Attachee = this;` will crash if `a_AttachTo == nullptr` right?

bearbin:
Yes, dereferencing a nullptr is not a good idea. So the original code was wrong, because the null-check was too late. But that doesn't mean it can be removed entirely :P

tigerw:
Passing null is incorrect anyway, there's a `Detach` for that. Additionally assuming the check is removed by the compiler we've no bug reports about that function crashing. So I predict all uses are correct. I can check again tho.

bearbin:
Yes, I think all _current_ uses are correct, but from a quick glance at that function a future developer could pass in nullptr without realising what could happen. We should avoid footguns. Add an assert, which will allay any fears about runtime cost.

bearbin (approved):

PR #4501, opened 2020-03-16
Implement formatting codes

mathiascode:
Implements formatting codes similar to Bukkit (&-symbol).
Fixes #2695
Fixes https://github.com/cuberite/Core/issues/84

PR #5304, opened 2021-09-21
Player chat message formatting removed

### LINKED ISSUE #5287: Having an @ in your message acts weird while sending messages
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 4b8952e438da8e57d6442a0b7451dac8060f154b

### Expected behavior
When not using commands, `@` and any characters after it should appear as normal in messages.

### Actual behavior
Having an `@` in the message causes weird behaviour. E.g. sending the message `@t` causes an empty message to be send, but sending just the `@` does send that character. Sending something with `@d` makes the rest of the text appear pink, so I'm guessing these are color codes somehow? It's quite unexpected at least.

As for use-case, I was sharing a [Matrix](https://matrix.org) address with someone and those are in the format of `@username:server.domain`. Sharing e-mail addresses is also quite hard. Of course most people won't do this, but it should still be possible.

The full message (with `@`) character is shown correctly in the console however.

1. Send any message with an @ and some character after it immediately.

Rorkh:
As it doesn't processes at Vanilla server and produces unexpected and uncontrollable behavior at all.
Maybe it should be regulable in settings as extra-feature instead of remove?

fix #5287

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1475,7 +1475,12 @@ void cClientHandle::HandleChat(const AString & a_Message)
 	Msg.AddTextPart(m_Player->GetName(), Color);
 	Msg.ParseText(m_Player->GetSuffix());
 	Msg.AddTextPart("> ");
-	Msg.AddTextPart(Message);
+	if (m_Player->HasPermission("chat.format"))
+	{
+		Msg.ParseText(Message);
+	} else {
+		Msg.AddTextPart(Message);
+	}
 	Msg.UnderlineUrls();
 	cRoot::Get()->BroadcastChat(Msg);
 }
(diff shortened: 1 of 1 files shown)


LittleBigBug:
If it should be in cuberite at all, it should just be implemented as an optional or permission-based feature using formatting codes like #4501

tigerw (approved):

mathiascode:
This will break the default MOTD. https://github.com/cuberite/Core/blob/master/core_motd.lua

Rorkh:
I'd like to make it permission-based. Will do it now.

> 
> 
> If it should be in cuberite at all, it should just be implemented as an optional or permission-based feature using formatting codes like #4501

Ok, I agree. I'd like to make it permission-based rather than optional. It will give more control over this thing.

Issue #2320, opened 2015-07-07
Make generators multi-threaded


SphinxC0re:
When a huge world is explored by players, the server is acting slow

jammet:
This would help a lot on the Raspberry Pi 2. (4 cores)

NiLSPACE:
According to @madmaxoft the generator is already async: https://github.com/cuberite/cuberite/issues/2324#issuecomment-119897046

Howaner:
I can confirm this.
The World TPS goes really down at chunk loading.

tigerw:
Maybe the cause is then something in the tick thread waiting for the generator to complete a chunk.

SphinxC0re:
@NiLSPACE I think @madmaxoft meant that the Chunk generation is async to the world interaction

SamJBarney:
@LO1ZB is working on a multi-threaded generator right now.

LO1ZB:
I said that I'm trying to make it multithreaded, I dono have much experience with multithreading (yet).

madmaxoft:
The only big decision is whether to make one instance support multiple threads, or have a separate instance for each thread; also whether to share the caches or not.

LO1ZB <notifications@github.com> wrote:

>I said that I'm trying to make it multithreaded, I dono have much
>experience with multithreading (yet).
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/cuberite/cuberite/issues/2320#issuecomment-120473828

-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

NiLSPACE:
Why would we share the caches? That shouldn't help allot since the seeds are most likely different right?

worktycho:
I have one question. Does anyone have any evidence that the world generation computations are the bottleneck? If so can you post it. Otherwise I think we need to double check what is causing the problem fist, because it could easily be a queue or something else non-intuitive.

LO1ZB:
I would give you a full profile of the generation if I knew where the profile confuguration went....

LO1ZB:
I'm profiling and let MCServer generate 251001 Chunks.

LO1ZB:
Exportet profile data (xml and cvs)
https://mega.nz/#!khx1TIAa!GuW5d8m5c-IE9V4mWtCl6o1If-6b5lraSxzCDRmv4TY

worktycho:
[...]

Issue #4835, opened 2020-08-26
Windows: Conflicting runtime libraries


madmaxoft:
As of commit d2de029c024309b0ac2429373cc8ef2e56f880e0, the build process on Windows seems half-broken - some of the libraries use the `Multithreaded (/MT)` runtime library and the main executable with other libraries use `Multithreaded DLL (/MD)`. The linker outputs an error message and links the Multithreaded DLL runtime, which is not what we want in the end. The Debug builds have a similar issue (`/MTd` vs `/MDd` runtime settings).

Tested with VS2017 and CMake 3.14.

This is most likely caused by the recent CMake reorganization.

tigerw:
😦

Could you try clearing the build folder, just in case?

madmaxoft:
Even after starting a build in a fresh new folder, this still happens.

madmaxoft:
Ah, the `CMAKE_MSVC_RUNTIME_LIBRARY` is not supported on CMake 3.14, it's new in CMake 3.15. So either we drop support for CMake 3.14 on Windows, or provide a workaround.

tigerw:
Are you willing to upgrade? :)

madmaxoft:
Already upgraded, but still, we should then raise the CMake minimum version required.

PR #5308, opened 2021-09-25
Implemented multi-threaded chunk generation #2320

Project579:
This Pull Request tries to solve issue #2320 , the solution is not perfect as there are still bottlenecks, mostly caused by the locks in World, ChunkMap, LightingThread and the ones around the caches and some generators; even with these bottlenecks the performance difference is still very noticeable especially on systems with 6 or more cores, performance on low end systems with 4 or less cores may have decreased and should be verified, some performance tuning can be achieved by modifying the constants in ThreadPool.h.

Side notes:
Even with the changes in this Pull Request chunk **loading** on the client side still looks slow (especially on Windows where the timers are low resolution and thread wake up is slow), meaning that there must still be some bottlenecks, my guesses are chunk loading and saving, networking and lighting code, also I noticed that it _may_ be beneficial to have a dedicated thread for chunk saving (instead of the current implementation where it's shared with loading) but the biggest show stopper to any changes for added parallelism seems to be the way the  ChunkMap lock is implemented, there is no easy way to make a lot of the read only operations concurrent (thus allowing more then one thread at a time to use the chunk map and each chunk).

Changes:
* Added oneTBB library;
* Added ThreadPool class, used only to initialize TBB with
  the required parameters;
* Changed ChunkGeneratorThread to spawn a tbb task for each chunk,
  made the chunk generation request queue concurrent;
* Added "const" and "static" to many functions and parameters that where missing it
  to make it easier to spot already thread safe code;
* Added some mutexes around caches used in the chunk generation process;
* Made m_NextIdx in SpawnPrepare  atomic;
* Fixed undefined behaviour in cWorld::TickQueuedEntityAdditions;
* Changed ChunkSender to use concurrent data structures.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -65,3 +65,6 @@
 [submodule "lib/libdeflate"]
 	path = lib/libdeflate
 	url = https://github.com/cuberite/libdeflate
+[submodule "lib/oneTBB"]
+	path = lib/oneTBB
+	url = https://github.com/oneapi-src/oneTBB.git
--- a/CMake/AddDependencies.cmake
+++ b/CMake/AddDependencies.cmake
@@ -23,6 +23,16 @@ function(build_dependencies)
 	set(JSONCPP_WITH_CMAKE_PACKAGE OFF CACHE BOOL "Generate and install cmake package files")
 	set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build jsoncpp_lib as a shared library.")
 	set(BUILD_OBJECT_LIBS OFF CACHE BOOL "Build jsoncpp_lib as a object library.")
+	
+	set(TBB_TEST OFF CACHE BOOL "TBB: Enable testing")
+	set(TBB_EXAMPLES OFF CACHE BOOL "TBB: Enable examples")
+	set(TBB_STRICT ON CACHE BOOL "TBB: Treat compiler warnings as errors")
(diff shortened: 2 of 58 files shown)


Project579:
Looks like issue #4835 decided to bite me as TBB explicitly checks for CMake 3.15 when CMAKE_MSVC_RUNTIME_LIBRARY is set.

tigerw:
mhm thanks. I'm interested in knowing what lock contentions you saw; I assume chunk generation is only (?) the caches, lighting a chunk sending the world lock, but anything else?

tigerw:
We may end up rolling our own thread pool, it doesn't look like Thread Building Blocks has very good static lib support.

Project579:
I decided to use TBB because it has a lot of concurrent data structures which are very difficult to implement, regarding static lib support the problems start to arise only if you use TBB in other static libraries or DLLs as long as its only linked to the main executable it's fine, but using your own thread pool will still be beneficial if you want to track it via the DeadLockDetector.

Regarding Lock contention I made an Excel Sheet, the analysis was done in Windows 10 x64 via Intel VTune, I simply set the world pre-generation to 128 chunks with an empty world folder and let it finish, should be easy to reproduce: 
[LockContentionResults.xlsx](https://github.com/cuberite/cuberite/files/7231044/LockContentionResults.xlsx)

madmaxoft:
Is there any way you could benchmark this to see the actual speed difference between 1 thread generating, and, say, 4 threads generating?

PR #5311, opened 2021-10-01
Code refactor, const ref, lower scope and minor optimizations

GermanAizek:
The code is cleaned under the C++ standard, more correct use of methods and techniques


src/Generating/EnderDragonFightStructuresGen.cpp
```diff
@@ -170,10 +170,11 @@ void cEnderDragonFightStructuresGen::GenFinish(cChunkDesc &a_ChunkDesc)
 		EnderDragon->SetPosition({0.0, static_cast<double>(a_ChunkDesc.GetHeight(0, 0) + 20), 0.0});  // Spawns the dragon 20 blocks above the terrain at (0, 0)
 		a_ChunkDesc.GetEntities().emplace_back(std::move(EnderDragon));
 		*/  // Todo: 25.10.20 - Add the ender dragon spawning when the dragon behaves properly - 12xx12
+		int size = static_cast<int>(FloorC(-m_Fountain.GetSizeX() / 2));
 		a_ChunkDesc.WriteBlockArea(m_Fountain,
-			static_cast<int>(FloorC(-m_Fountain.GetSizeX() / 2)),
+			size,
 			62,
-			static_cast<int>(FloorC(-m_Fountain.GetSizeX() / 2)),
```
tigerw, line 176: @12xx12 do you know if this was meant to be GetSizeZ?

tigerw:
Could you remove the const-ref changes? We found that pass-by-value for small objects is faster.

src/Generating/EnderDragonFightStructuresGen.cpp
12xx12: Feels like it. I can't remember if that was an intention of mine.

I think this is a small mistake although but it did slip because the fountain is a square.

GermanAizek:
> 
> 
> Could you remove the const-ref changes? We found that pass-by-value for small objects is faster.

Ok

bearbin:
I have to say that these changes appear to be completely pointless, the optimisations that you have performed manually will be performed automatically by the compiler, the constref changes are counterproductive, and the stylistic changes do not help with readability. I am closing this PR as it's fundamentally a waste of time for all involved. Your talents would be much better spent adding features or fixing bugs rather than churning away at vapours.

PR #5313, opened 2021-10-02
World: change spawnpoint type to int

tigerw:
As Vanilla does.

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -734,7 +734,7 @@ void cWorld::GenerateRandomSpawn(int a_MaxSpawnRadius)
 	}
 
 	// A search grid (searches clockwise around the origin)
-	static const int HalfChunk = static_cast<int>(cChunkDef::Width / 2.0f);
+	static const int HalfChunk = cChunkDef::Width / 2;
 	static const Vector3i ChunkOffset[] =
 	{
 		Vector3i(0, 0, HalfChunk),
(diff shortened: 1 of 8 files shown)


src/World.cpp
```diff
@@ -270,9 +270,9 @@ cWorld::cWorld(
 	if (m_IsSpawnExplicitlySet)
 	{
 		LOGD("Spawnpoint explicitly set!");
-		m_SpawnX = IniFile.GetValueF("SpawnPosition", "X", m_SpawnX);
-		m_SpawnY = IniFile.GetValueF("SpawnPosition", "Y", m_SpawnY);
-		m_SpawnZ = IniFile.GetValueF("SpawnPosition", "Z", m_SpawnZ);
+		m_SpawnX = IniFile.GetValueI("SpawnPosition", "X", m_SpawnX);
+		m_SpawnY = IniFile.GetValueI("SpawnPosition", "Y", m_SpawnY);
+		m_SpawnZ = IniFile.GetValueI("SpawnPosition", "Z", m_SpawnZ);
```
bearbin, line 275: How compatible is this with existing decimal spawn points - will the decimal just be ignored or does this fail?

tigerw: It uses `atoi` to convert, and quite luckily its behaviour of "ignore everything after the number" works in our favour!

bearbin (approved):

bearbin:
(well, once the clang fail is fixed)

tigerw:
Thank you!

PR #5317, opened 2021-10-23
Update MobSpawner.cpp

ThomasDel45:


tigerw:
Hi, could you explain the changes? I don't think zombies are required to spawn on grass, and SkyLight is an unsigned variable so can't be smaller than zero.

bearbin:
These changes do not make sense. Closing.

PR #5318, opened 2021-11-01
gh-action for auto releases

thecoder-001:
Makes getting latest builds from GitHub saner compared to build artifacts.


thecoder-001:
Created pr to the wrong project. Sorry for the trouble.

PR #5320, opened 2021-11-09
Miscellaneous fixes

tigerw:
Some things found when preparing other changes.

+ Set m_IsFlying, invulnerability flag for spectators
* Call ProcessProtocolIn before ticking so all changes are visible at once
* Fix uninitialised chunk unload time variable access
* Rename SendRespawn parameter to reflect what it's used for
* Invert sense of growth probability
- Remove redundant initialisations in cPlayer that are already handled by LoadFromDisk

### DIFF
--- a/src/Generating/BioGen.cpp
+++ b/src/Generating/BioGen.cpp
@@ -5,7 +5,6 @@
 
 #include "Globals.h"
 #include "BioGen.h"
-#include <iostream>
 #include "IntGen.h"
 #include "ProtIntGen.h"
 #include "../IniFile.h"
@@ -1197,6 +1196,8 @@ std::unique_ptr<cBiomeGen> cBiomeGen::CreateBiomeGen(cIniFile & a_IniFile, int a
 
 // Change to 1 to enable the perf test:
 #if 0
+
+#include <iostream>
 
 class cBioGenPerfTest
 {
--- a/src/Physics/Explodinator.cpp
+++ b/src/Physics/Explodinator.cpp
@@ -241,7 +241,6 @@ namespace Explodinator
 	Currently missing conduits from 1.13 */
 	static bool BlockAlwaysDrops(const BLOCKTYPE a_Block)
 	{
-		// If it's a Shulker box
 		if (IsBlockShulkerBox(a_Block))
 		{
 			return true;
--- a/src/WorldStorage/StatisticsSerializer.cpp
+++ b/src/WorldStorage/StatisticsSerializer.cpp
@@ -55,26 +55,15 @@ static void LoadCustomStatFromJSON(StatisticsManager & Manager, const Json::Valu
 	for (auto it = a_In.begin(); it != a_In.end(); ++it)
 	{
 		const auto & Key = it.key().asString();
-		const auto StatInfo = NamespaceSerializer::SplitNamespacedID(Key);
-		if (StatInfo.first == NamespaceSerializer::Namespace::Unknown)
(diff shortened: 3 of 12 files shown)


src/WorldStorage/StatisticsSerializer.cpp
```diff
@@ -55,26 +55,15 @@ static void LoadCustomStatFromJSON(StatisticsManager & Manager, const Json::Valu
...
-		}
-		catch (const std::out_of_range &)
-		{
-			FLOGWARNING("Invalid statistic type \"{}\"", StatName);
-		}
-		catch (const Json::LogicError &)
-		{
-			FLOGWARNING("Invalid statistic value for type \"{}\"", StatName);
-		}
+		Manager.Custom[NamespaceSerializer::ToCustomStatistic(Name)] = it->asUInt();
```
12xx12, line 66: why is the error handling removed?

src/Physics/Explodinator.cpp
```diff
@@ -241,7 +241,6 @@ namespace Explodinator
 	Currently missing conduits from 1.13 */
 	static bool BlockAlwaysDrops(const BLOCKTYPE a_Block)
 	{
-		// If it's a Shulker box
```
12xx12, line 244: The floor is made out of floor

src/Generating/BioGen.cpp
```diff
@@ -5,7 +5,6 @@
 
 #include "Globals.h"
 #include "BioGen.h"
-#include <iostream>
```
12xx12, line 8: Hm?

bearbin:
Why has std::move been replaced with std::swap? I don't see why a swap should be required in those cases, semantically.

src/WorldStorage/StatisticsSerializer.cpp
tigerw: if there's a parse error it should stop the player from joining instead of discarding that statistic value and losing it forever

src/Physics/Explodinator.cpp
tigerw: the floor is lava and you can't changemymind

src/Generating/BioGen.cpp
tigerw: I moved in down into the #define block

src/WorldStorage/StatisticsSerializer.cpp
12xx12: Oh - the try/catch is just moved to kick the player?

tigerw:
True it's the same but I want to replace them all with a WorkQueue<T> at some point to reuse the same vector memory. The server predominantly uses std::swap in all candidate locations and I don't want to miss anything when the search-replace comes along. I can change it back if you want?

src/Physics/Explodinator.cpp
KingCol13: Oh how devs will struggle without my insightful commentary

src/WorldStorage/StatisticsSerializer.cpp
tigerw: Yes, cClientHandle handles the exception and kicks the player if any error occurred when creating the cPlayer.

12xx12:
i'm fine with  this.

@bearbin has the last call on merging

bearbin:
Yep, it looks odd but I don't think it's really a problem.

bearbin (approved):

Issue #5322, opened 2021-11-14
Server can crash upon receiving certain packets.


AirOne01:
Client version: 1.12 ([Wurst](https://github.com/Wurst-Imperium/Wurst7) v6.35.1)
Server OS: Windows 11 Home (v.21H2 build 22000.258)
Cuberite Commit id: d49ce751ba6f8f2eb171633cea30a1119935f8c7

### Expected behavior
Server ignores incoming packets

### Actual behavior
Server crash on certain packets being sent

### Steps to reproduce the behavior
The best way to reproduce this is by using the latest version of the Wurst cheat client.
1. Join your Cuberite server
2. Open the Wurst menu and search for the `LogSpammer` cheat.
![image](https://user-images.githubusercontent.com/21955960/141684815-1223c349-4beb-4387-a96f-5da9d2f8895f.png)
3. Activate it.
4. The server will crash about a second after you enabled the cheat.
![image](https://user-images.githubusercontent.com/21955960/141684837-14ff9b69-6d86-487a-a166-f28f6966280e.png)
If you restart the server and join with the cheat still enabled, it will crash again.

### Server log
```
Info [15:15:42] User AirOne01 authenticated with UUID 4c36470614584778bf6f9254e6e32346
     [15:15:42] A player connected through a proxy without requiring a forwarding secret. If open to the internet, this is very insecure!
Info [15:15:42] Player AirOne01 has joined the game
Warn [15:15:42] cChunkMap::AddEntity: Entity at 0x20cfe7da390 (cPlayer, ID 140) spawning in a non-existent chunk.
     [15:15:43] Unhandled vanilla plugin channel: "MC|PickItem".
Err  [15:15:43] Verification failed: a_ByteBuffer.ReadSome(Message, a_ByteBuffer.GetReadableSpace() - 1), file C:/projects/cuberite/src/Protocol/Protocol_1_8.cpp, line 2883
Err  [15:15:43] Failure report: 

  :(   | Cuberite has encountered an error and needs to close
       | SIGABRT: Server self-terminated due to an internal fault
       |
       | Cuberite AppVeyor (id: 2399)
       | from commit d49ce751ba6f8f2eb171633cea30a1119935f8c7
```

### Additional notes
For this test I only used native plugins (`Core` and `ChatLog`)
I have absolutely no idea if this is too sensitive to be posted here, but since I don't see a lot of projects using Cuberite for production, I'm assuming it's fine.

tigerw:
[...]

Issue #1692, opened 2014-12-29
Disconnected when setting up a command block


carc1n0gen:
When I try to create a simple command block I get disconnected and the command in the block doesn't save.

carc1n0gen:
I am very new to github and pretty much only used it for a few little things of my own.  This was the first project on github I felt like contributed to so I don't know how it all works so I am sorry if it is wrong to ask.  Is this issue being looked in to?  I have confirmed the problem to happens in both the precompiled linux 64 bit build and by compiling myself, and precompiled Windows 64 bit build.

madmaxoft:
I confirm the bad behavior, when a client edits the commandblock's text and hits the Done button, the server kicks them (and commits suicide in Debug mode due to badly parsed PluginMessage packet).

madmaxoft:
Fixed by ignoring the garbage data at the end of the packets.

Pokechu22:
FYI, for the command block it isn't garbage data but instead an extra Boolean for whether output from the command block should be stored (which is why it initially happened).  Don't know about what it is in 1.7.10, though.

tigerw:
Thanks, added the missed read.

PR #5325, opened 2021-11-15
Protocol: modernise plugin message handling

tigerw:
- Remove handling for garbage data in command block sets
+ Add correct boolean value read (#1692)
* Make better use of namespaced-id and string_view

Follow up to #5322.

### DIFF
--- a/src/Protocol/Protocol_1_13.cpp
+++ b/src/Protocol/Protocol_1_13.cpp
@@ -30,6 +30,7 @@ Implements the 1.13 protocol classes:
 #include "../World.h"
 #include "../JsonUtils.h"
 #include "../WorldStorage/FastNBT.h"
+#include "WorldStorage/NamespaceSerializer.h"
 
 #include "../Bindings/PluginManager.h"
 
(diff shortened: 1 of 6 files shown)


src/Protocol/Protocol_1_13.cpp
```diff
@@ -660,21 +661,25 @@ void cProtocol_1_13::HandlePacketPluginMessage(cByteBuffer & a_ByteBuffer)
...
-		m_Client->SetClientBrand(Brand);
+		if (Value == "brand")
+		{
+			HANDLE_READ(a_ByteBuffer, ReadVarUTF8String, AString, Brand);
 
-		// Send back our brand, including the length:
-		m_Client->SendPluginMessage("minecraft:brand", "\x08""Cuberite");
-		return;
+			m_Client->SetClientBrand(Brand);
+			m_Client->SendPluginMessage("brand", "\x08""Cuberite");  // Send back our brand, including the length.
```
bearbin, line 674: What's the advantage of putting the comment after the code? It just makes it harder to read on narrow screens.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2554,28 +2554,88 @@ void cProtocol_1_8_0::HandlePacketPlayerPosLook(cByteBuffer & a_ByteBuffer)
...
+
 	HANDLE_READ(a_ByteBuffer, ReadVarUTF8String, AString, Channel);
 
+	const std::string_view ChannelView = Channel;
+
 	// If the plugin channel is recognized vanilla, handle it directly:
-	if (Channel.substr(0, 3) == "MC|")
+	if (ChannelView.substr(0, 3) == "MC|")
 	{
-		HandleVanillaPluginMessage(a_ByteBuffer, Channel);
```
bearbin, line 2562: The control flow is clearer if this is a seperate function, rather than having a single function be 80 lines long. It certainly doesn't need to be exposed though, it would be fine to make it a private/static function.

bearbin (approved):
Generally looks OK.

src/Protocol/Protocol_1_8.cpp
tigerw: Yep, I added it back and made the virtual function useful, all the 1.8 - 1.12 messages should be read now.

src/Protocol/Protocol_1_13.cpp
tigerw: Nicely aligned `m_Client->S` lines... How narrow is your screen anyway xD

PR #5343, opened 2021-11-25
Fixed Compiler Warnings

12xx12:
This PR fixes some compiler warnings. This is just for me to check if I introduce errors in my block state PR.


src/BlockEntities/HopperEntity.cpp
```diff
@@ -254,7 +254,7 @@ bool cHopperEntity::MovePickupsIn(cChunk & a_Chunk)
 				{
 					m_bFoundPickupsAbove = true;
 
-					int PreviousCount = m_Contents.GetSlot(i).m_ItemCount;
+					char PreviousCount = m_Contents.GetSlot(i).m_ItemCount;
```
bearbin, line 257:
```suggestion
					auto PreviousCount = m_Contents.GetSlot(i).m_ItemCount;
```

src/Generating/BioGen.cpp
```diff
@@ -796,7 +796,7 @@ EMCSBiome cBioGenTwoLevel::SelectBiome(int a_BiomeGroup, size_t a_BiomeIdx, int
 		EMCSBiome OuterBiome;
 	} ;
 
-	static BiomeLevels bgOcean[] =
+	static BiomeLevels bgOceanBlocks[] =
```
bearbin, line 799: Was this a warning??

src/Generating/FinishGen.cpp
```diff
@@ -1840,7 +1840,7 @@ cFinishGenOres::OreInfos cFinishGenOres::OreInfosFromString(const AString & a_Or
 			LOGWARNING("Cannot parse ore information from string, invalid number in OreInfo \"%s\".", ore.c_str());
 			continue;
 		}
-		res.emplace_back(oreType, oreMeta, maxHeight, numNests, nestSize);
+		res.emplace_back(static_cast<BLOCKTYPE>(oreType), oreMeta, maxHeight, numNests, nestSize);
```
bearbin, line 1843: Probably better to change BlockStringToType here so that it returns BlockType, rather than casting the result...

src/Generating/BioGen.cpp
12xx12: Yes- there is another variable in the global scope with the same name

src/Generating/FinishGen.cpp
12xx12: Actually, The function may return -1 if the block is unknown and the data type blew everything up. I already tried this

12xx12: `BLOCKTYPE` is `unsigned char` so no way we get a -1 in there

bearbin: Ugh, C-style brittleness... If only strong type systems had caught on sooner. Nevermind, let's just mark it down as another place to use optional types once we get around to that conversion effort.

bearbin (approved):
Looks good. I'm slightly concerned by the wholesale change of item stack sizes to char from int, let's hope that won't cause any issues; also there are a few more casts than would be ideal, but since they're just making what was implict now explicit that should be fine.

12xx12:
I agree with your concern about the changes to the stack sizes. But we only allow a maximum of 64 per stack. So I don't see a scenario where we actually have a number larger then 265

src/Entities/Pickup.cpp
```diff
@@ -43,7 +43,7 @@ class cPickupCombiningCallback
 		cItem & Item = OtherPickup.GetItem();
 		if ((Distance < 1.2) && Item.IsEqual(m_Pickup->GetItem()) && OtherPickup.CanCombine())
 		{
-			short CombineCount = Item.m_ItemCount;
+			char CombineCount = Item.m_ItemCount;
 			if ((CombineCount + m_Pickup->GetItem().m_ItemCount) > Item.GetMaxStackSize())
```
tigerw, line 47: could we have CombineCount = 64, m_ItemCount = 64, and their sum = 128 overflowing signed `char`?

12xx12: 8 Bit = 256
7 Bit = 128

so yeah 64+64 will result in 0. I'm gonna think of something else.

tigerw: we could change the storage of cItem::m_ItemCount to unsigned, might be a bit a work but we were planning on it anyway

bearbin: Unsigned brings it's own pitfalls. We need the compilers to warn on all the arithmetic footguns, then some more sensible semantics might come about.

12xx12: I think I'm going to take this change out of this PR.

And then do the unsigned char change in another and fix this there.

When using unsigned char this should suffice. 8 bits may contain a value of 255 which is enough for this, isn't it?

tigerw: The additions will work, and an item count will ever start off negative. But, we need to verify in places when a subtraction `minuend - subtrahend` happens that `minuend >= subtrahend` otherwise with unsigned the result will wrap instead of going negative.

bearbin:
@tigerw Please confirm that this is all good for you now.

tigerw:
```
   s
   t
l👁👃👁kin'
   l
   l

**Comment by bearbin** [comment]
Looking presumed complete @tigerw ?

**Comment by tigerw** [comment]
i'm trying to figure out if the hooks can be made to take eBlockFace instead of char xD

**Comment by tigerw** [comment]
sorry this took so long, and thanks for going through all of these warnings!

File: src/Protocol/Protocol_1_8.cpp [hunk 5]
```diff
@@ -504,7 +513,7 @@ void cProtocol_1_8_0::SendEntityAnimation(const cEntity & a_Entity, const Entity
 		return;
 	}
 
-	if (const auto AnimationID = GetProtocolEntityAnimation(a_Animation); AnimationID != static_cast<unsigned char>(-1))
+	if (const auto AnimationID = GetProtocolEntityAnimation(a_Animation); AnimationID != unsigned char(-1))
```
bearbin, line 516: `unsigned char(-1)`

Really, C++, why would you allow such ugly syntax...

tigerw: mmm keyword spaces

it does look 🤔 but i saw the PR used `type(literal)` syntax elsewhere so I made it consistent. do you want to change?

bearbin: No change necessary :)

tigerw: I WAS WRONG
c++ doesn't allow this notation to have spaces
i'll change it back

src/Entities/Pickup.cpp
tigerw: I just realised, the code as written is correct, since the operands of the addition are both `short` to get promoted to `int`, similar with the comparison.

PR #5344, opened 2021-11-26
transition to non-pointer item handler

12xx12:
This PR replaces the pointer based item handlers with the same method as the block handlers.

I need to proofread the initialization. This will be done tomorrow

### DIFF
--- a/src/Blocks/BlockMobSpawner.h
+++ b/src/Blocks/BlockMobSpawner.h
@@ -71,8 +71,8 @@ private:
 		}
 
 		const auto Player = static_cast<const cPlayer *>(a_Digger);
-		auto Handler = Player->GetEquippedItem().GetHandler();
-		if (!Player->IsGameModeSurvival() || !Handler->CanHarvestBlock(E_BLOCK_MOB_SPAWNER))
+		auto & Handler = Player->GetEquippedItem().GetHandler();
+		if (!Player->IsGameModeSurvival() || !Handler.CanHarvestBlock(E_BLOCK_MOB_SPAWNER))
 		{
 			return;
 		}
--- a/src/Inventory.cpp
+++ b/src/Inventory.cpp
@@ -260,8 +260,7 @@ int cInventory::ReplaceOneEquippedItem(const cItem & a_Item, bool a_TryOtherSlot
 	cItem ItemsToAdd = a_Item;
 	if (EquippedItem.IsEqual(ItemsToAdd))
 	{
-		cItemHandler Handler(ItemsToAdd.m_ItemType);
-		auto AmountToAdd = std::min(static_cast<char>(Handler.GetMaxStackSize() - EquippedItem.m_ItemCount), ItemsToAdd.m_ItemCount);
+		auto AmountToAdd = std::min(static_cast<char>(cItemHandler::For(ItemsToAdd.m_ItemType).GetMaxStackSize() - EquippedItem.m_ItemCount), ItemsToAdd.m_ItemCount);
 
 		EquippedItem.m_ItemCount += AmountToAdd;
 		SetEquippedItem(EquippedItem);
--- a/src/Inventory.cpp
+++ b/src/Inventory.cpp
@@ -78,7 +78,7 @@ int cInventory::HowManyCanFit(const cItem & a_ItemStack, int a_BeginSlotNum, int
 	}
 
 	char NumLeft = a_ItemStack.m_ItemCount;
-	int MaxStack = ItemHandler(a_ItemStack.m_ItemType)->GetMaxStackSize();
+	int MaxStack = cItemHandler::For(a_ItemStack.m_ItemType).GetMaxStackSize();
 	for (int i = a_BeginSlotNum; i <= a_EndSlotNum; i++)
 	{
 		const cItem & Slot = GetSlot(i);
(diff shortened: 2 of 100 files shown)


src/Blocks/BlockMobSpawner.h
```diff
@@ -71,8 +71,8 @@ class cBlockMobSpawnerHandler final :
 		}
 
 		const auto Player = static_cast<const cPlayer *>(a_Digger);
-		auto Handler = Player->GetEquippedItem().GetHandler();
-		if (!Player->IsGameModeSurvival() || !Handler->CanHarvestBlock(E_BLOCK_MOB_SPAWNER))
+		auto & Handler = Player->GetEquippedItem().GetHandler();
```
bearbin, line 74: You use `Player->GetEquippedItem().GetHandler()` here, but `cItemHandler::For(Player->GetEquippedItem().m_ItemType)` below, and `cItemHandler::For(m_Player->GetEquippedItem())` in the next file down. It would look nicer if there was some consistency in how we're doing this - why's it even possible to do it these three ways in the first place!

src/Inventory.cpp
```diff
@@ -78,7 +78,7 @@ int cInventory::HowManyCanFit(const cItem & a_ItemStack, int a_BeginSlotNum, int
 	}
 
 	char NumLeft = a_ItemStack.m_ItemCount;
-	int MaxStack = ItemHandler(a_ItemStack.m_ItemType)->GetMaxStackSize();
+	int MaxStack = cItemHandler::For(a_ItemStack.m_ItemType).GetMaxStackSize();
```
bearbin, line 81: Similarly here, GetMaxStackSize exists directly on cItem, we should either be using these helpers, or getting rid of them.

src/Items/ItemBigFlower.h
```diff
@@ -8,14 +8,14 @@
...
+class cItemBigFlowerHandler final:
 	public cItemHandler
 {
 	using Super = cItemHandler;
 
 public:
 
-	cItemBigFlowerHandler():
+	constexpr cItemBigFlowerHandler(int a_ItemType):
 		Super(E_BLOCK_BIG_FLOWER)
```
bearbin, line 19: should this be a_ItemType here, or the arg removed?

src/Items/ItemDefaultItem.h
```diff
@@ -0,0 +1,12 @@
+//
+// Created by marc_ on 26.11.2021.
+//
+
```
bearbin, line 4: Hopefully Git keeps this information without the assistance of comments!

12xx12: My IDE being "helpful"

src/Blocks/BlockMobSpawner.h
tigerw: I'd say we can use convenience accessors on cItem when we have an instance, while For is for when we only have a bare item type.

src/Items/ItemBigFlower.h
tigerw: You could say `using Super::Super` to pull in cItemHandler's constructor.

tigerw:
Thanks for doing this!

src/Items/ItemBigFlower.h
bearbin: I'm no C++ expert to know a better way, but moving the item ID to the construction doesn't strike me as being that big of an issue. The only problem here is the argument that does nothing, just pass it through.

src/Blocks/BlockMobSpawner.h
bearbin: @tigerw That seems like a reasonable distinction to make. In that case should we remove the overload for `For` that takes a cItem? But we can certainly convert the uses in this PR to be consistent.

@12xx12 that method does seem workable, since constructing a cItem from an item id is simple, but that's still a bit more overhead than just calling `For` directly. Still an option though.

12xx12: I can't think of a situation where you would need a Item handler whiteout having a cItem. That's why I would hide the function from everything else

src/Items/ItemBigFlower.h
12xx12: Yeah That's no problem. But I'd rather have a constructor without an argument.

src/Blocks/BlockMobSpawner.h
bearbin: It does seem like all the instances you found in this PR have cItems, and it's not exported to the API. Go ahead with the removal then!

src/Inventory.cpp
12xx12: I am just using the helpers

src/Items/ItemHandler.cpp
```diff
@@ -724,24 +1315,7 @@ char cItemHandler::GetMaxStackSize(void)
 
 
 
-bool cItemHandler::IsTool()
```
12xx12, line 727: By the way. This is a unused function.

12xx12:
> Thanks for doing this!

You're welcome.

I usually do live audio and lighting tech stuff. But a gig got cancelled due to Corona. So I had a day of spare time

bearbin (approved):
Looks good, thanks!

Issue #4671, opened 2020-04-18
Extend HOOK_PLAYER_FISHING and HOOK_PLAYER_FISHED with Experience amount


madmaxoft:
The `HOOK_PLAYER_FISHING` and `HOOK_PLAYER_FISHED` hooks should get a new parameter, `ExperienceAmount`, that specifies the amount of experience the player gets for the fishing operation' `HOOK_PLAYER_FISHING` should also be able to change the amount.

npresley0506:
Hey, I'm a student doing a final project for my class that requires me to work on an open-source project. Mind if I try grabbing this one and taking a crack at it?

bearbin:
@npresley0506 Of course, we always encourage new contributors. This issue isn't asssigned to anyone so it's free for you to work on. As a new contributor you should take a look at the [getting started](https://github.com/cuberite/cuberite/blob/master/GETTING-STARTED.md) and [contributing](https://github.com/cuberite/cuberite/blob/master/CONTRIBUTING.md) files for some guidance and help with the project. If you have any other questions I recommend joining the Discord group or the forum as there are several contributors there that could offer help.

feyokorenhof:
I don't know if this is still being worked on but I'd be happy to take a look.

bearbin:
Unless there's a PR for an issue, or it's actively under discussion/assignment, it's probably not being worked on. Feel free to have a go :)

bearbin:
Fixed by #5345

PR #5345, opened 2021-11-28
Added ExperienceAmount variable to HOOK_PLAYER_FISHING and HOOK_PLAYER_FISHED

feyokorenhof:
Basically only added ExperienceAmount (int) to all the places where the hooks get called. I have a feeling I missed something though; Where do we give the experience to the player ?

I'm new to C++ and didn't see 'const int' anywhere else but I still used it in HOOK_PLAYER_FISHED for good measures. Maybe someone can clear that up for me. My guess is const is only viable for variables that take up a lot of memory.

#4671


feyokorenhof:
Hmm I don't like that it adds my previous commits. Is this something only I can see and if not how can I prevent this x)

bearbin:
> Hmm I don't like that it adds my previous commits. Is this something only I can see and if not how can I prevent this x)

Nope, everyone can see them :)

Looks like you worked on your own master branch, rather than a feature branch, previously. You need to do a hard reset of your own master branch, and then rebase your this feature branch on top of the clean master branch, e.g:

```
$ git stash
$ git checkout master
$ git reset --hard upstream/master
$ git checkout fishinghooksexperience
$ git rebase master
```

That may not be your exact problem, and I advise you to do some more learning about Git as it does require some knowledge and experience to use properly, but it's powerful once you get the concepts behind it.

feyokorenhof:
Yeah that's what I thought. Forgot to make a new branch when I commited my last PR. I'll rebase and make a new pr thanks!

src/Bindings/Plugin.h
```diff
@@ -76,8 +76,8 @@ class cPlugin
 	virtual bool OnPlayerBrokenBlock        (cPlayer & a_Player, int a_BlockX, int a_BlockY, int a_BlockZ, char a_BlockFace, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta) = 0;
 	virtual bool OnPlayerDestroyed          (cPlayer & a_Player) = 0;
 	virtual bool OnPlayerEating             (cPlayer & a_Player) = 0;
-	virtual bool OnPlayerFished             (cPlayer & a_Player, const cItems & a_Reward) = 0;
-	virtual bool OnPlayerFishing            (cPlayer & a_Player, cItems & a_Reward) = 0;
+	virtual bool OnPlayerFished             (cPlayer & a_Player, const cItems & a_Reward, const int ExperienceAmount) = 0;
+	virtual bool OnPlayerFishing            (cPlayer & a_Player, cItems & a_Reward, int ExperienceAmount) = 0;
```
bearbin, line 80: This int actually needs to be an int reference, as the point of this hook is to allow the plugin to change the ExperienceAmount (so it shouldn't be a const(ant) parameter). It needs to be a ref (&) so that changes made inside the hook apply to the parameter of the call, rather than the internal copy.

bearbin (changes requested):

src/Bindings/Plugin.h
feyokorenhof: Ah I see that makes sense. But const can still be used for the OnPlayerFished, right?

bearbin:
You don't need to make a new PR if you just force push after the rebase, the existing PR will be updated. GitHub PRs are branch-based rather than commit or patch based.

src/Bindings/Plugin.h
bearbin: Yep, that's the ideal type to use there. you can't change the experience given after it's already happened! A regular const int (not a reference) like you put is right.

src/Bindings/PluginManager.cpp
```diff
@@ -788,11 +788,11 @@ bool cPluginManager::CallHookPlayerFished(cPlayer & a_Player, const cItems & a_R
 
 
 
-bool cPluginManager::CallHookPlayerFishing(cPlayer & a_Player, cItems a_Reward)
+bool cPluginManager::CallHookPlayerFishing(cPlayer & a_Player, cItems a_Reward, int ExperienceAmount)
```
feyokorenhof, line 791: I'm a bit confused of why a_Reward here is not a (const) reference. Is this because in FISHED the reward is already settled so we use const reference for performance. And in FISHING we use a copy instead becuase the rewards will change while fishing so sending back and forth references isn't ideal? I might be completely wrong about this but I'm curious :p. 
And what about the ExperienceAmount in this case.

bearbin: You've found a bug :P This should be a non-const reference as well.

feyokorenhof: Hahaha okay didn't think about that.

feyokorenhof:
I followed your steps to rebase and hard reset so that should be fixed now. Now I know what stash is used for so thanks x)

feyokorenhof:
f (cRoot::Get()->GetPluginManager()->CallHookPlayerFishing(a_Player, Drops))
Looks like I forgot a few files as clang-tidy is failing

bearbin (commented):
I'm not sure what happened with the rebase, but it doesn't look like the force push did anything... 

Anyway, you just need to update the documentation for the hook in APIDump/Classes/Plugins.lua then this will be ready to merge.

bearbin:
Well, after fixing the compile of course as well :)

feyokorenhof:
Hmm weird. Should I just make a new branch then or is it fine like this?

Also: I've overlooked ItemFishingRod.h where the hook gets called:
`cRoot::Get()->GetPluginManager()->CallHookPlayerFished(a_Player, Drops);` (missing experienceAmount - clang-tidy)

An experience orb (1-6xp) is actually dropped there:
`a_World.SpawnExperienceOrb(a_Player.GetPosition(), Random.RandInt(1, 6));`.

I could store that xp in a variable and pass it to CallHookPlayerFished but wouldn't that be doing the same thing twice?

bearbin (approved):
I made the required changes.

PR #5346, opened 2021-12-10
Mitigate against CVE-2021-44228

bearbin:
Cuberite is not directly affected, but could relay attack payloads to users. This adds some sanitisation to chat messages (which are logged insecurely in Vanilla).


12xx12:
I think this is sufficient to catch malicious messages.

Maybe add a link in the comment. Just In case for future questions.

Pokechu22:
I would recommend rejecting messages containing just `${` - mojang's [new log configuration for 1.7-1.11](https://launcher.mojang.com/v1/objects/14fedd49897b3f6160975bdf3a53e1e7eab71d29/client-1.7.xml) stops logging messages if they match the regex `.*\$\{[^}]*\}.*` (i.e. messages that contain `${` followed by `}`).

bearbin:
@Pokechu22 If that's filtered out there's no possibility of exploitation at all, and if that causes issues it can be rolled back and replaced with something more considered.

PR #5353, opened 2021-12-27
Adding missing block infos

Persson-dev:
I just wanted to make beetroots transparent and non solid (for the farmers to be able to harvest them). Then, i realized that some blocks were missing so i added them.
The observer block is really weird because in the wiki they say it's transparent, but mobs can spawn on it and redstone can be placed on it too.

### DIFF
--- a/src/BlockInfo.cpp
+++ b/src/BlockInfo.cpp
@@ -895,6 +895,7 @@ bool cBlockInfo::IsSolid(const BLOCKTYPE Block)
 		case E_BLOCK_ACTIVATOR_RAIL:
 		case E_BLOCK_ACTIVE_COMPARATOR:
 		case E_BLOCK_AIR:
+		case E_BLOCK_BEETROOTS:
 		case E_BLOCK_BIG_FLOWER:
 		case E_BLOCK_BROWN_MUSHROOM:
 		case E_BLOCK_CARPET:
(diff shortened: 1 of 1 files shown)


src/BlockInfo.cpp
```diff
@@ -1054,6 +1061,7 @@ bool cBlockInfo::IsTransparent(const BLOCKTYPE Block)
 		case E_BLOCK_OAK_DOOR:
 		case E_BLOCK_OAK_FENCE_GATE:
 		case E_BLOCK_OAK_WOOD_STAIRS:
+		case E_BLOCK_OBSERVER:
```
bearbin, line 1064: Does changing this actually affect redstone placement as you mentioned? If so I think we should leave this change off. Otherwise these all look good.

Persson-dev: It does not "cut" redstone wires (see below). So it behaves a little like a transparent block too.
observer

Persson-dev: Also, it does not conduct redstone.
observer

bearbin: Is this Cuberite or vanilla? All we care about is that they behave the same.

Persson-dev: It's vanilla.

bearbin: Good, seems like it does behave like a transparent block then.

bearbin (approved):

Issue #3239, opened 2016-06-23
New 1.9 command blocks not supported.


EnderFlop:
Client version:  1.9
Commit id:        ec998e821fd905966ed14ed47a29520ec1d1d782 built at: 314

### Expected behavior
A 1.9 custom map would work regularly, even with the new command blocks

### Actual behavior
When I load a custom 1.9 map with the new kind of command blocks on it, the console spams Block entity mismatch errors for every command block and then proceeds to crash. Here is the error for one of the new Chain Command Blocks.

WorldLoader(world): Block entity mismatch: block type chaincommandblock (211), type "Control", at {6, 10, -3}; the entity will be lost.

### Steps to reproduce the behavior
I used the map Ballistic Wars by Cloud Wolf
[Ballistic_Wars_v2.zip](https://github.com/cuberite/cuberite/files/329254/Ballistic_Wars_v2.zip)

Please add new command block support!

EnderFlop:
Bump
Also, comparators don't work either. Here is an image of the error
![cuberite error](https://cloud.githubusercontent.com/assets/18680718/16597293/982d631c-42a4-11e6-86f9-9a4b91b31b3a.png)

madmaxoft:
Crash is bad, that's right; but still command blocks aren't supported properly even in the older protocol versions, we still haven't decided on the full API to be used for them (they currently use console commands API, which is not really what is wanted)

EnderFlop:
Do you know when/if a decision will be made? What is the difference between console commands API and other possible APIs?

madmaxoft:
Here's the API discussion, as you can see, it was started almost a year ago:
https://forum.cuberite.org/thread-2195.html

mathiascode:
Doesn't seem to crash anymore at least.

12xx12:
NOTE: The command blocks should use their custom name if the name was changes

PR #5358, opened 2021-12-30
Add player permission level

Persson-dev:
I added permission level to players. It's used to indicate to the client which operations it can perform (opening command blocks, switching gamemodes, executing commands).
It should be set with the Core plugin (op command) to be fully functional (for now, it does nothing).
I tested to give me level 4 permission level and i could open command blocks and use the keyboard shortcut F3+N to switch between creative and spectator mode.
I used 1.12.2 version to test this but the specifications have not changed since 1.8 so i just had to modify 1.8 protocol.
https://minecraft.fandom.com/wiki/Permission_level


tigerw:
I think the `int` should be derived from the existing permission system instead of being its own independent thing.

Persson-dev:
Do you mean assigning the int to a rank or something?

bearbin:
I agree, taking @tigerw's comment to mean - the permission level sent in the protocol should be derived from a HasPermission call, and the numerical value should not be stored or exposed on the player.

Persson-dev:
The goal is actually to expose it to the player. If it's not exposed to the player, the client does not even try to open the gui of a command or change the gamemode using F3+N. It prints "Unable to change gamemode, no permission" instead even if you are op. Or maybe i just misunderstood you.

tigerw:
Thanks, you just uncovered part of the mystery of #3239

There's a [RankManager](https://github.com/cuberite/cuberite/blob/master/src/RankManager.h) handling Bukkit-style fine-grained permissions, see [WorldEdit](https://github.com/cuberite/WorldEdit#biome) for example. Instead of bolting on a separate vanilla-style operator level permission system, we should find a way to translate our string permissions to integer levels.

For example, Vanilla level 1 is "bypass spawn protection". We could map Core's current bypass spawn protection permission [core.spawnprotect.bypass](https://github.com/cuberite/Core/blob/68519b3e0e7fc129f2fae21c7a766c83167c3cc7/core_worlds.lua#L114) to level 1.

To do the mapping, maybe something like

```cpp
EntityAnimation GetPermissionLevel(Player)

if !Player.HasPermission("core.spawnprotect.bypass")
  return EntityAnimation::PlayerSetsOperatorLevelZero

if !Player.HasPermission("comandblock.set") || !Player.HasPermission("core.clear") || ...
    return EntityAnimation::PlayerSetsOperatorLevelOne

// ...
return EntityAnimation::PlayerSetsOperatorLevelTwo
```

Persson-dev:
Is it better now ? Is it enough to send the permission level to the client when SetPlayerRank is called or are they more functions to cover ?

bearbin:
Permissions could also change when they're modified for a rank, rather than just when a player is assigned to one. I'm also not sure if we should set the permission level based on having all the permissions for a level or just having at least one for a level. Since it's a client side thing, we can still deny the action afterwards - so I think we should set the level based on the highest permission-level one of their rank's permissions requires.

Persson-dev:
> I think we should set the level based on the highest permission-level one of their rank's permissions requires.


The permission is now based on the highest permission-level one of the player's rank. 
Also, the permission is now updated whenever a rank or group has changed.
It should be good now.

bearbin (approved):

Persson-dev:
I don't want to hurry you, but is this PR still good or do you want me to make more changes ?

tigerw:
Sorry about my delay. The plugin docs says plugins are expected to manually invoke LoadRank to update ranks after any changes, and indeed Core does this. It also has the nice effect of batching any effects of a sequence of rank/permission operations. With the current code we could end up with some performance issues if every rank manager operation sends packets to every player in the server. I'm still thinking through if `UpdatePlayersPermissionLevel` is needed and if we will be okay solely relying on the requirement to call LoadRank, and having the network packet sent in that function.

Persson-dev:
I moved the update function to LoadRank and it seems to work fine. However, LoadRank is called once when the player is not yet initialized so i have to skip the UpdatePermissionLevel call to not crash Cuberite with segmentation fault.
I currently check if m_MaxHealth is initialized but maybe there is a cleaner way to do it.

tigerw:
Thank you!

Issue #5354, opened 2021-12-28
Server Crashes on Placing Item in Anvil with 1.13.2


12xx12:
This is tested on the master branch. 

If you place an item in an anvil, the server crashes because it tries to read the wrong number of by with  PacketType 28.

bearbin:
Fixed in #5362

PR #5362, opened 2021-12-31
Fixing anvil crash issue

Persson-dev:
Prevents the server from crashing when an item is put in an anvil with 1.13.2 version (#5354). Should i fully implement it or leave it like it is ?


bearbin (approved):
I think this is good enough, it fixes the crash, so I'll merge. But if you want to implement the feature afterwards, go ahead!

PR #5363, opened 2021-12-31
Fixed diagonal redstone powering

### LINKED ISSUE #2219: Redstone wire should not travel down slabs
Redstone wire should not travel down slabs

### LINKED ISSUE #5335: Redstone power elevator with slabs don't work
Client version: 1.12.2
Server OS: Windows 11
Cuberite Commit id: 95bc2eac222a12c049b1aef799ac280fa9123af4

### Expected behavior
Power is conducted up to the top in the following build (left image) using slabs or transparent blocks such as glowstone.
image
This is caused by a bug that has been kept by Mojang as a feature (like BUD). I couldn't find the original bug on the tracking site, but I did find duplicates leading to similar behaviour, like [this one](https://bugs.mojang.com/browse/MC-50507).

### Actual behavior
Power is not conducted (right image).

Reproduce the setup shown in the images with any kind of slabs.

havel06:
Redstone wire accepts power from lower diagonal through transparent blocks. Fixes #5335. Fixes #2219.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -27,6 +27,7 @@ feyokorenhof
 Gareth Nelson
 GefaketHD
 HaoTNN
+havel06 (Michal Havlíček)
 Howaner
 ion232 (Arran Ireland)
 jan64
--- a/src/Simulator/IncrementalRedstoneSimulator/RedstoneWireHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/RedstoneWireHandler.h
@@ -306,31 +306,29 @@ namespace RedstoneWireHandler
 					return;
 				}
 
-				if (Front == FrontState::Up)
-				{
-					Callback(Relative + OffsetYP);
-				}
-
 				BLOCKTYPE LateralBlock;
 				a_Chunk.UnboundedRelGetBlockType(Relative, LateralBlock);
 
-				// Don't check YM side diagonal when blocked by a non-transparent block
-				if (!cBlockInfo::IsTransparent(LateralBlock))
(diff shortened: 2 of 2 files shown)


tigerw:
This can be simplified I think, insert this conjunction below line 145
```cpp
!cBlockInfo::IsTransparent(LateralBlock)  // Which is sitting on an opaque block (redstone ladders)
```

Testing a bit we're doing exactly the opposite of Vanilla, power flows down a redstone ladder instead of up. This is because wires try looking for connecting wires above them first.

This extra conjunction stops that if the wire above is sitting on a slab/glowstone/transparent block so wires in the ladder configuration now only connect to (hence receive power from) wires below. Thus we reverse the direction of power flow.

tigerw:
Also fixes #2219 from 6 years ago.

tigerw:
Hm although the suggestion doesn't work if the wire snakes around, hold on

havel06:
I'll look into it further.

havel06:
Did a slight rewrite, redstone power now only climbs upwards on transarent blocks.

tigerw:
Try changing line 149 to
```cpp
SetDirectionState(Offset, Block, cBlockInfo::IsTransparent(LateralBlock) ? TemporaryDirection::Side : TemporaryDirection::Up);
```
this should also ensure the wire displays correctly when 1.13 comes along.

havel06:
Got it.

tigerw:
Thanks! Now the code neatly handles wires with state `Side` and only takes power from below and laterally, hence the ladders work, so I think the changes in `ForValidSourcePositions` aren't needed anymore.

havel06:
Indeed!

tigerw (approved):

bearbin (approved):

PR #5369, opened 2022-01-01
Custom 25565 port page

### LINKED ISSUE #5357: Sending a HTTP request to the minecraft port (25565) actually returns something readeable with Firefox. Let's do something funny with that

Rorkh:
Fixes #5357

I'm not really well at C++ and cuberite codebase but tried to realise custom 25565 port page capability that reads it from page.html file in root cuberite server folder.
Feel free to make suggestions.


Server/page.html
```diff
@@ -0,0 +1 @@
+<script>window.location.replace("https://www.youtube.com/watch?v=dQw4w9WgXcQ");</script>
```
bearbin, line 1: This should validate, even if it's just a rickroll.

src/Protocol/ProtocolRecognizer.cpp
```diff
@@ -76,6 +76,39 @@ AString cMultiVersionProtocol::GetVersionTextFromInt(cProtocol::Version a_Protoc
 
 
 
+bool cMultiVersionProtocol::CheckHTTPRequest(cClientHandle & a_Client, ContiguousByteBuffer & a_Data)
```
bearbin, line 79: The name here is off, this should be TryHandleHTTPRequest (better yet, separate out the checking and the handling into two functions.

Server/page.html
Rorkh: what do you mean?

bearbin: Well, give it a `<html>` tag at least, a doctype would be nice, etc.

src/Protocol/ProtocolRecognizer.cpp
```diff
@@ -90,6 +132,12 @@ void cMultiVersionProtocol::HandleIncomingDataInRecognitionStage(cClientHandle &
 		return;
 	}
 
+	bool isHTTP = TryHandleHTTPRequest(a_Client, a_Data);
+	if (isHTTP)
+	{
+		return;
+	};
```
bearbin, line 139:
```suggestion
	if (TryHandleHTTPRequest(a_Client, a_Data))
	{
		return;
	};
```

tigerw:
Shouldn't we send a 302 See Other?

```cpp
const std::string_view Response("HTTP/1.0 303 See Other\r\nLocation: http://www.youtube.com/watch?v=dQw4w9WgXcQ\r\n\r\n");
a_Client.SendData({ reinterpret_cast<const std::byte *>(Response.data()), Response.size() });
a_Client.Destroy();

**Comment by bearbin** [comment]
That would be a lot simpler, but there is an advantage to this method: the user can replace page.html with their own to customise.

**Comment by tigerw** [comment]
The redirect can be a configuration in settings.ini, I'm sure server admins can host an HTTP server or already have a website. Serving files seems out of the remit of this easter egg I think.

**Comment by Rorkh** [comment]
Yeah. Agree with tigerw. But if there no redirect url should we show something instead of
`U�S{"text":"Your client isn\'t supported.
Try connecting with Minecraft 1.8.x-1.12.x"}`
Blank screen?

**Comment by tigerw** [comment]
Thanks!

If the admin manually cleared the redirect url let's assume their intention was "don't do anything special". Less work to implement!

File: src/Protocol/ProtocolRecognizer.cpp [hunk 4]
```diff
@@ -238,6 +243,37 @@ void cMultiVersionProtocol::SendDisconnect(cClientHandle & a_Client, const AStri
...
+	ContiguousByteBuffer Buffer;
+	m_Buffer.ReadSome(Buffer, 10U);
+	m_Buffer.ResetRead();
+
+	// The request line, hastily decoded with the hope that it's encoded in US-ASCII.
+	const std::string_view Value(reinterpret_cast<const char *>(Buffer.data()), Buffer.size());
+
+	if (Value == u8"GET / HTTP")
+	{
+		const auto Response = fmt::format(u8"HTTP/1.0 303 See Other\r\nLocation: {}\r\n\r\n", cRoot::Get()->GetServer()->GetCustomRedirectUrl());
```
Bond-009, line 264: Causes errors on clang 13:
```c++
/home/bond/dev/cuberite/src/Protocol/ProtocolRecognizer.cpp:262:15: error: type of UTF-8 string literal will change from array of const char to array of const char8_t in C++20 [-Werror,-Wc++20-compat]
        if (Value == u8"GET / HTTP")
                     ^
/home/bond/dev/cuberite/src/Protocol/ProtocolRecognizer.cpp:262:15: note: remove 'u8' prefix to avoid a change of behavior; Clang encodes unprefixed narrow string literals as UTF-8
        if (Value == u8"GET / HTTP")
                     ^~
/home/bond/dev/cuberite/src/Protocol/ProtocolRecognizer.cpp:264:37: error: type of UTF-8 string literal will change from array of const char to array of const char8_t in C++20 [-Werror,-Wc++20-compat]
                const auto Response = fmt::format(u8"HTTP/1.0 303 See Other\r\nLocation: {}\r\n\r\n", cRoot::Get()->GetServer()->GetCustomRedirectUrl());
                                                  ^
/home/bond/dev/cuberite/src/Protocol/ProtocolRecognizer.cpp:264:37: note: remove 'u8' prefix to avoid a change of behavior; Clang encodes unprefixed narrow string literals as UTF-8
                const auto Response = fmt::format(u8"HTTP/1.0 303 See Other\r\nLocation: {}\r\n\r\n", cRoot::Get()->GetServer()->GetCustomRedirectUrl());
```

tigerw: Disabled the warning for now since we need the header string to be encoded in ASCII and not a system-native encoding.

PR #5371, opened 2022-01-01
Handlers: update item and block handlers

### LINKED ISSUE #5367: Chests disappear when killed or rejoining server
Client version: 1.12
Server OS: Windows
Cuberite Commit id: 1edc6c8601f9d00751bf7c9f7fcc69d1e18bb46a

### Expected behavior
Chests should stay where they are and not disappear whenever someone gets killed or leaves the server

### Actual behavior
Chests disappear whenever you get killed or leave and rejoin the server 

1. Place a chest and put blocks in the chest
2. Get killed or leave and rejoin the server
3. Chest should not be there anymore

4. Place a newly crafted chest on the same location as the chest that disappeared and it wont use the chest to place another chest (So chest reappears but you still have your newly crafted chest in hand)
5. Open the reappeared chest and you will find the items you put in it in step 1.

tigerw:
+ Resend blocks when server rejects placement because in a block

* Fix chest handler not invoked, fixes #5367
* Fix comparator handler not invoked
* Update some naming

### DIFF
--- a/src/Blocks/BlockEnderChest.h
+++ b/src/Blocks/BlockEnderChest.h
@@ -6,7 +6,7 @@
 
 
 
-class cBlockEnderchestHandler final :
+class cBlockEnderChestHandler final :
 	public cYawRotator<cBlockEntityHandler, 0x07, 0x03, 0x04, 0x02, 0x05>
 {
 	using Super = cYawRotator<cBlockEntityHandler, 0x07, 0x03, 0x04, 0x02, 0x05>;
@@ -38,7 +38,3 @@ private:
 	}
 
 };
-
-
-
-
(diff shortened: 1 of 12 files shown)


bearbin:
Is there any reason some of the Item handlers had Block in their name - I can't make sense of why that would be, hopefully it wasn't indicating something important! But this looks fine.

tigerw:
If I understood 12xx12's intent correctly 'Block' denotes an item handler for a technical block, so I tried to update them all based on that.

(With the BlockStates PR those 'Block' handlers will all be deleted when cItemHandler is made to work with ItemType instead of re-using ENUM_BLOCK_TYPE.)

12xx12:
You are right. For example the doors have an item and a block ID at the moment.

In the old implementation the door item would get the door handler. And the door block would get the default block handler

12xx12 (approved):

PR #5372, opened 2022-01-01
ClientHandle: improve right-click robustness

tigerw:
+ Add checks for result of GetBlockTypeMeta
+ Kick if the client sent an invalid block face or coordinate
* Update outdated comments

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1437,38 +1437,44 @@ void cClientHandle::FinishDigAnimation()
 
 void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ, bool a_UsedMainHand)
 {
-	// This function handles three actions:
-	// (1) Place a block;
-	// (2) "Use" a block: Interactive with the block, like opening a chest/crafting table/furnace;
-	// (3) Use the held item targeting a block. E.g. farming.
-	//
-	// Sneaking player will not use the block if hand is not empty.
-	// Frozen player can do nothing.
-	// In Game Mode Spectator, player cannot use item or place block, but can interactive with some block depending on cBlockInfo::IsUseableBySpectator(BlockType)
-	//
-	// If the action failed, we need to send an update of the placed block or inventory to the client.
-	//
-	// Actions rejected by plugin will not lead to other attempts.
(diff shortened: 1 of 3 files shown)


src/ClientHandle.cpp
```diff
@@ -1437,38 +1437,44 @@ void cClientHandle::FinishDigAnimation()
...
+
+	If the action failed, we need to send an update of the placed block or inventory to the client.
+
+	Actions rejected by plugin will not lead to other attempts.
+	E.g., when opening a chest with a dirt in hand, if the plugin rejects opening the chest, the dirt will not be placed. */
+
+	if ((a_BlockFace == BLOCK_FACE_NONE) || !cChunkDef::IsValidHeight(a_BlockY))
+	{
+		throw std::runtime_error("Invalid click - hacked client?");
+	}
```
bearbin, line 1457: This is not a runtime error. And WHY would you use an exception here??? LOGD + return is more than fine.

bearbin (changes requested):

src/ClientHandle.cpp
tigerw: bearbin doesn't like exceptions!

There should be a kick because a connected player sent an unambiguously invalid packet, so log a hacked client message to notify the admin.

An exception instead of Kick() because anywhere an incoming packet is being processed and we realise it's malicious we need to stop immediately. An exception goes up to ProcessProtocolIn() which responds by kicking and a disconnect. But a Kick() + return runs the risk of the server continuing to process data in the receive buffer.

The type to throw can be changed, maybe `InvalidDataException`.

bearbin: Hmm. I can see what you're getting at here, but I can't characterise this as malicious data. Junk data, yes, but we should handle that. No excuse for just throwing our hands up and kicking the player. If we just drop the packet, nothing bad happens. And give a LOGD to alert the admins or the devs, fine.

12xx12: I think just logging the error and ignoring the packet is da wae.

tigerw: Kicking the player in this case, not shutting down xD

OK if we want to ignore instead of throw an accusation at them and kick, where there's no apparent security issue with ignoring, let me open another PR to unify all the other places where this happens.

LOGD instead of LOGINFO to avoid an easy way to logspam, though we lose the notification in Release.

bearbin (approved):

Issue #5323, opened 2021-11-14
Fall damage miscalculation


AirOne01:
Client version: 1.12.2
Server OS: Windows 11 Home (v.21H2 build 22000.258)
Cuberite Commit id: d49ce751ba6f8f2eb171633cea30a1119935f8c7 

### Expected behavior
- No fall damage is taken when jumping on cobwebs
- Entities that are spawned in mid-air take damage from their fall

### Actual behavior
- Player sometimes receive fall damage when touching the floor in cobwebs
- Sometimes upon landing on cobwebs no damage will be taken, which is expected, but *further fall damage will be ignored* somehow. (tested teleportation and "natural" fall damage - by climbing a wall and fallign from it). It can be cancelled by getting attacked another entity.
- Entities that are spawned in mid-air don't damage from their fall

### Steps to reproduce the behavior
Two tests:
- Player fall on cobwebs
  - Either build a tower and climb back on it every time or teleport to a greater y. Same behavior expected.
- Entity is spawned in mid-air and falls to ground

### Server log
Nothing special indicated by the server log

PR #5374, opened 2022-01-02
Fix compilation with Clang 13.0.0

Bond-009:
Clang 13 changes:
* `-Wreserved-identifier` and `-Wunused-but-set-variable` got added
* the `-Wreturn-std-move-in-c++11` option was removed

Builds still fail when native optimizations are enabled

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -158,7 +158,7 @@ function(set_exe_flags TARGET)
 			# TODO: actually fix the warnings instead of disabling them
 			# or at least disable on a file-level basis:
 			-Wno-missing-noreturn -Wno-padded -Wno-implicit-fallthrough
-			-Wno-double-promotion -Wno-reserved-identifier
+			-Wno-double-promotion
 
 			# This is a pretty useless warning, we've already got -Wswitch which is what we need:
 			-Wno-switch-enum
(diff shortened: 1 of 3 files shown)


src/Entities/Pawn.cpp
```diff
@@ -377,7 +376,6 @@ void cPawn::HandleFalling(void)
 			if (j == 0)
 			{
 				IsFootInWater |= IsBlockWater(BlockType);
-				IsFootInLiquid |= IsFootInWater || IsBlockLava(BlockType) || (BlockType == E_BLOCK_COBWEB);  // okay so cobweb is not _technically_ a liquid...
```
bearbin, line 380: I think we should be doing something with this, rather than just deleting the LOC...

Looks like it's meant to be used in the code below, but somebody forgot.

SetFlags.cmake
```diff
@@ -158,7 +158,7 @@ function(set_exe_flags TARGET)
 			# TODO: actually fix the warnings instead of disabling them
 			# or at least disable on a file-level basis:
 			-Wno-missing-noreturn -Wno-padded -Wno-implicit-fallthrough
-			-Wno-double-promotion
+			-Wno-double-promotion -Wno-reserved-identifier
```
bearbin, line 161: This breaks previous clangs :facepalm:

bearbin (changes requested):

src/Entities/Pawn.cpp
Bond-009: It can always get added back, it's not a complicated piece of code.

tigerw: > Player sometimes receive damage upon falling in cobwebs

possibly causing part of #5323

bearbin (approved):

bearbin:
Sorry for the delayed review, my email client decided to disconnect from the world and I didn't notice until now :|

PR #5375, opened 2022-01-02
Suspicious packets: give 'em the silent treatment

tigerw:
Ref: #5372


bearbin (approved):

bearbin:
I think in due course we could add a OnInvalidPacket hook for plugins to use if they want to replicate the old behaviour, but this is good for now.

PR #5379, opened 2022-01-06
Add working villages

Persson-dev:
I started to implement villages. Since it's a big thing, you can already tell me if i should change some things so i don't waste too much time doing it again. It can then allow iron golems to spawn and villager to breed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -62,6 +62,7 @@ target_sources(
 	StringCompression.cpp
 	StringUtils.cpp
 	UUID.cpp
+	Village.cpp
 	VoronoiMap.cpp
 	WebAdmin.cpp
 	World.cpp
@@ -143,6 +144,7 @@ target_sources(
 	StringUtils.h
 	UUID.h
 	Vector3.h
+	Village.h
 	VoronoiMap.h
 	WebAdmin.h
 	World.h
--- a/src/Village.cpp
+++ b/src/Village.cpp
@@ -1,0 +1,137 @@
+
+#include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
+
+#include "Village.h"
+
+#include "World.h"
+
+
+
+
+
+////////////////////////////////////////////////////////////////////////////////
+// cVillageHouse:
+
+
+
+
+
+bool cVillageHouse::IsValidHouse()
+{
+	// TODO: check the inside and the outside
+	return true;
+}
+
+
+
+
+
+bool cVillageHouse::IsDeserted()
+{
+	return m_Village->GetVillageTime() - m_LastVillagerTime >= HOUSE_VALID_TIME;
+}
+
+
+
+
+
+void cVillageHouse::UpdateHouse()
+{
+	m_LastVillagerTime = m_Village->GetVillageTime();
+}
+
+
+
+
+
+////////////////////////////////////////////////////////////////////////////////
+// cVillage:
+
+
+
+
+
+void cVillage::Tick(std::chrono::milliseconds a_Dt)
+{
+	m_Tick += a_Dt.count();
+
+	for (std::size_t I = 0; I < m_Houses.size(); I++)
+	{
+		cVillageHouse & House = m_Houses[I];
+
+		if (House.IsDeserted())
+		{
+			RemoveHouse(I);
+		}
+	}
+}
+
+
+
+
+
+void cVillage::UpdateBoundaries()
+{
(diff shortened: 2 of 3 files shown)


tigerw:
Villages went through a rework in 1.14 and seems to have been simplified, i.e. much less state storage. The village.dat format got replaced with [Point of Interest](https://minecraft.fandom.com/wiki/Java_Edition_level_format#poi_format) and it seems village mechanics are much more driven by villager/iron golem entities instead of a village object.

So if we upgrade cVillage into something more generic, PointOfInterest maybe, it'll handle not only village stuff but all the other points of interest like portals and beehives.

Persson-dev:
I was trying to implement the old village mechanics (pre 1.14) to have a coherence with the targeted version of Cuberite. 
But if you want the new villages mechanics, I can add them too and try something with Point of Interest (which is new for me).

tigerw:
Somewhere in the doddering old 1.12 Cuberite there's a fresh young 1.18 compatible Cuberite struggling to emerge. Let's help it on it's way!

ZefenDEV:
> Somewhere in the doddering old 1.12 Cuberite there's a fresh young 1.18 compatible Cuberite struggling to emerge. Let's help it on it's way!

yes, it would not hurt to make support for 1.18, but I think it will be difficult

tigerw:
Found this: [poi](https://gitlab.bixilon.de/bixilon/pixlyzer-data/-/raw/master/version/1.18.1/points_of_interest.json) corresponding to [blocks](https://gitlab.bixilon.de/bixilon/pixlyzer-data/-/raw/master/version/1.18.1/blocks.json)

Guess poies are simply a list of blocks, supplied by the generator, updated with SetBlocks, and saved to disk.

I can't find end portals in there, maybe they aren't stored, but are looked up based on the current seed?

Persson-dev:
Implementing poies was simplier than i thought. I just added the poi object to see if it's fine and then i will implement them in the world.
I don't know what to do for villager professions because for now, with the available blocks (pre 1.14), we can only get Cleric and Leatherworkers.
Should i keep professions random (as it is) or make them depend on poies ?

Persson-dev:
Now, the poies are created when a poi block is placed (or removed when it has been replaced). The only thing remaining is the storage. And then, i will be able to start working on villages if everything is good.

Persson-dev:
Poies are stored in the poi folder instead of the region one.
I don't know where to start. Can someone give me advices ?

12xx12 (changes requested):
Thank you for your work. Sorry for the long wait.

Maybe the title of the PR is misleading :D 

In total I Thing this is ready to merge beside some comments I made.

Persson-dev:
It has been a very long time. I almost forgot this PR. Right now, i don't have time to continue working on it because of my studies. I will probably work on this PR during computer classes if I'm too bored xD

Persson-dev:
While we were learning how to use char arrays, I applied your suggestions, it should be good. For now, this PR is kind of useless because the poies are not stored in disk. I didn't do that because i didn't know which file to edit to make this work.
Where should I start to save/load the poies on disk?

12xx12:
The wiki is quite straightforward 

https://minecraft.fandom.com/wiki/Java_Edition_level_format#poi_format

Persson-dev:
I meant in Cuberite because poies are not stored in region files but in other files.

12xx12:
Any writing to disk is done in src/WorldStorage. The actual call is done in a async way. I suggest looking in cWorld or check where the code from src/WorldStorage is called.

I think the POIs should be stored at the same point in time the world is stored

Persson-dev:
6 months after, I'm back.
I found this easier to store poies within the region files instead of doing the "Vanilla" way. So they are stored in chunk sections. It looks something like this :
image
The code is far from being perfect but it works for now.
Hope it helps.

12xx12:
Thank your for the effort.

The goal is to be like vanilla Minecraft is. So loading and saving POIs should be done in the same way Mojang does

Persson-dev:
I tried to replicate the storage of Vanilla Minecraft as much as possible. The official one looks something like this :
image

The thing is that pois are not stored in the region directory but in another one called poi (the files have the same name and it's stored a bit like regions) and I have no idea on how to do that in Cuberite. I found this easier to just put everything in the region files. So if you know how to do that, you can reuse some of my code

12xx12:
I see. I did some poking around in the storage code, and it's not very friendly for this kind of modification. Next PR here I come

Issue #5365, opened 2021-12-31
Sticky pistons should drop their block when powered by a short pulse.


havel06:
Client version: 1.12
Server OS: Debian Linux

### Expected behavior

When a sticky piston gets powered by a 2 game tick (1 redstone tick) pulse, it should push away the block directly in front of it, without retracting it back.

### Actual behavior

The block gets retracted immediately.

### Steps to reproduce the behavior

Place a sticky piston facing a pushable block. Power the piston with a 2 game tick pulse.

InfRandomness:
I am taking a shot at fixing the issue

tigerw:
Thanks for taking a look! The relevant code is in BlockPiston.h/cpp and Simulator/IncrementalRedstoneSimulator/PistonHandler.h

AirOne01:
I reported something "similar" in #4605, if someone is interested.

PR #5382, opened 2022-01-12
Fix sticky piston pulling item when retracting if cycle was less than 2 ticks

InfRandomness:
Fixes https://github.com/cuberite/cuberite/issues/5365


InfRandomness:
Fixes https://github.com/cuberite/cuberite/issues/5365

InfRandomness:
There are still a few issues with this : 
After a few cycles of the piston extending and pushing the block, retracting and not pulling the block, then extending back, the block seems to completely disappear until logging out and back in
It also seems to happen so fast that it appears as laggy from a client

tigerw:
Could you revert the formatting changes? The diff has a lot of noise.

InfRandomness:
Did this work ?

PR #5383, opened 2022-01-12
Adds a test to determine piston stickyness

InfRandomness:
This commit adds a condition to prevent sticky pistons from pulling items back if they have been extended for <= 2 ticks


src/Blocks/BlockPiston.cpp
```diff
@@ -150,6 +151,12 @@ void cBlockPistonHandler::RetractPiston(Vector3i a_BlockPos, cWorld & a_World)
 				return;
 			}
 
+	if (ExtensionWorldTickTime - World.GetWorldTickAge() <= 2_tick)
+            {
+                // Piston has been activated for less or equal to 2 ticks, bail out
+                return;
+            }
```
bearbin, line 158: ExtensionWorldTickTime is not persisted, so when chunks are un- and re-loaded, this will always return when a newly-loaded extended piston is retracted. Is that the correct behaviour?

bearbin (commented):
Looks reasonable, just want to make sure the semantics here are desired.

Also, please add your name to the CONTRIBUTORS file so your contribution is recognised and to indicate your agreement to the LICENSE terms.

KingCol13:
Since there is only one handler and everything is static wouldn't this mean that if:

Piston 1 extends at tick 0
Piston 2 extends at tick 500
Piston 1 retracts at tick 501

Piston 1 will not retract its connected block as piston 2 has just set `ExtensionWorldTickTime`?

I think we need to keep a separate `ExtensionWorldTickTime` for each piston somehow.

bearbin:
Oh yes, this is a BlockHandler not a BlockEntity! This won't work at all with more than one piston in the world, it needs something to be stored per-piston.

bearbin:
Possibly in the redstone simulator? Not sure though, this needs a bit of thought.

Pokechu22:
Pistons normally become a block entity while moving ("block 36"/the moving piston block thing).  That block entity knows what block is being pushed and when the push should finish (if I recall correctly; the wiki should document it in more detail).

bearbin:
We don't have a block entity for pistons at the moment (at least I think so). Creating one and using it to store this value makes sense.

src/Blocks/BlockPiston.cpp
InfRandomness: Yeah, this might be an issue if the player let's a 2 tick clock running and goes to unload and load the chunk

InfRandomness:
I have created a PistonEntity class to represent both pistons and sticky pistons, but now I'm unsure where to go from there

src/BlockEntities/PistonEntity.h
```diff
@@ -0,0 +1,32 @@
...
+	virtual void SendTo(cClientHandle & a_Client) override;
+	virtual bool UsedBy(cPlayer * a_Player) override;
+
+	void SetExtensionWorldTicktime(const cTickTimeLong & a_ExtensionWorldTickTime);
+
+	cTickTimeLong GetExtensionWorldTickTime(void) const { return m_ExtensionWorldTickTime; }
+private:
+
+	cTickTimeLong m_ExtensionWorldTickTime;
+
```
InfRandomness, line 31: Maybe we will need to have the WorldTickTime in here ?

src/BlockEntities/PistonEntity.cpp
```diff
@@ -0,0 +1,52 @@
+// PistonEntity.cpp
+
+// Implements the cPistonEntity class representing a single piston in the world
+
+#include "Globals.h"  // NOTE: MSVC stupidness requires this to be the same across all modules
+#include "PistonEntity.h"
+
+cPistonEntity::cPistonEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World):
```
InfRandomness, line 8: For some reason, the constructor seems as if it never gets called, even though https://github.com/cuberite/cuberite/blob/1590d5bd1243f6406387bf8c1d319593ffe83983/src/BlockEntities/BlockEntity.cpp#L108-L109 has been added

tigerw:
I have a feeling adding moving piston tile entities is going to be a lot of work (and introduce dupe bugs), we should figure out how vanilla pistons respond to signals and extension, retraction timings before starting any changes.

Questions like how long does it take to extend, what happens when it loses or gains signal, a part of the piston is destroyed or a block is placed in the way during extension, retraction, what happens if the signal is pulsed in the same tick. Ideally there would be some way to replicate vanilla while only relying on the piston blocks as state. If we start adding too much complexity there's the risk of us repeating the countless vanilla piston dupe bugs over the years.

InfRandomness:
> I have a feeling adding moving piston tile entities is going to be a lot of work (and introduce dupe bugs), we should figure out how vanilla pistons respond to signals and extension, retraction timings before starting any changes.
> 
> Questions like how long does it take to extend, what happens when it loses or gains signal, a part of the piston is destroyed or a block is placed in the way during extension, retraction, what happens if the signal is pulsed in the same tick. Ideally there would be some way to replicate vanilla while only relying on the piston blocks as state. If we start adding too much complexity there's the risk of us repeating the countless vanilla piston dupe bugs over the years.

But why would we need to add moving piston tile entities in the first place?
Piston entities should be enough, right ? We only need this to define a behavior proper to each instance of piston on a world

tigerw:
I think Vanilla doesn't have piston entities, only the moving piston entities.

InfRandomness:
> I think Vanilla doesn't have piston entities, only the moving piston entities.

So we MUST have it too ? I think that as long as we can keep away from (possible uselessly) added complexity, it's better, no?

tigerw:
nah I don't like moving piston entities, hopefully we can figure something out which doesn't need any block entities at all. First though, we should have a spec of how pistons should behave on a tick-by-tick level, but I can't find this information on the wiki.

InfRandomness:
> nah I don't like moving piston entities, hopefully we can figure something out which doesn't need any block entities at all. First though, we should have a spec of how pistons should behave on a tick-by-tick level, but I can't find this information on the wiki.

Do you mind elaborating ? I'm not quite sure I understood everything ._.

tigerw:
At the moment Cuberite pistons work like this.

From a resting state,

Tick 0:
* If received redstone power
  * Check if is retracted
  * If yes, broadcast an extension block action & queue an extension task 1 tick later
* If lost power
  * Check if is extended
  * If yes, broadcast a retraction block action & queue a retraction task 1 tick later
* If power unchanged, do nothing

Tick 1:
* If received redstone power
  * Check if is retracted
  * If yes, broadcast an extension block action & queue an extension task 1 tick later
* If lost power
  * Check if is extended
  * If yes, broadcast a retraction block action & queue a retraction task 1 tick later
* If power unchanged, do nothing
* If an extension task fired
  * Check if can extend
  * If so, set the entire state atomically, including piston base, arm, and pushed blocks
* If a retraction task fired
  * Check if can retract
  * If so, set the entire state atomically, including piston base, arm, and pulled block

and tick 2, 3 repeats.

From this we can see that how a piston behaves depends on the state of the world (`GetBlock` on the piston position and redstone power), and scheduled tasks which call into the piston handler.

We also see why pistons don't drop blocks as vanilla does, because the state of the pulled block is updated atomically, the block is either pulled or not; there are no intermediate states visible to the outside.

Pokechu22:
> a part of the piston is destroyed or a block is placed in the way during extension

I believe the answer here is that the moving piston block entity is considered indestructible. At least, it's impossible to place a block in the way during extension because the destination block is also converted into a moving piston block (with its block entity) during the moving process, and you can't place something into that. Breaking I'm less sure of; you can't mine it, but I'm not sure what happens if it's destroyed via an explosion.

tigerw:
If we want to replicated vanilla behaviour, it seems we have to both
* extend the piston extension/retraction time from 1 tick to 3, and
* potentially split the atomic update of the blocks into separate updates of the piston base, arm, and pushed/pulled block

But, to do this robustly, I think we need to know exactly what vanilla does each tick a piston is active and how it responds to certain events in these newly introduced intermediate states, and formulate some sort of system to imitate it. As an example, why does vanilla take 3 ticks to extend a piston? How is it scheduling the events? Is it 1 tick delay + 2 ticks actual animation? Or something else?

The system we eventually come up with will ideally not need moving piston block entities, because that's extra work to save and load to disk, and unclear is if those entities need to be sent to the client, and if doing that would conflict with the piston animation the client plays (which also using moving piston entities). But if that's the only way, oh well. However a piston entity wouldn't work because we'd diverge from what vanilla does - the save file could have our own custom block entity in it - and ideally Cuberite tries to remain compatible.

tigerw:
I don't have the answers to this, maybe Pokechu our resident expert will, but these are only relevant if you want to overhaul pistons to make them 100% vanilla identical.

-----
If `ExtensionWorldTickTime` is indeed a quick fix for this issue and you only need it to be instanced per-piston, you could consider adding a map in `cIncrementalRedstoneSimulatorChunkData` which will store these times per block coordinate.

tigerw:
(as bearbin has already suggested at the start)

InfRandomness:
> However a piston entity wouldn't work because we'd diverge from what vanilla does - the save file could have our own custom block entity in it - and ideally Cuberite tries to remain compatible.
Right, this is where it causes problems at, this means that the world wouldn't be able to be used by regular minecraft-server

Pokechu22:
> and unclear is if those entities need to be sent to the client

I don't think you need to send anything special to the client other than the [piston block action](https://wiki.vg/Block_Actions#Piston), but unfortunately that's basically all I know on the subject; I don't have any further details on hand (and things may have changed since that rather sparse documentation was written, too).

InfRandomness:
> I don't have the answers to this, maybe Pokechu our resident expert will, but these are only relevant if you want to overhaul pistons to make them 100% vanilla identical.

I mean, if not doing so could cause some major issues, then I see no reason why I would want to implement a temporary fix, I might as well try to overhaul the piston system (although I think I'm far away from having the technical knowledge to do such a thing)

Pokechu22:
Actually, I remembered something else.  I think the only relevant property of the moving piston/block 36 block entity for servers is that when the progress field hits its maximum, the block changes to the stored block.  Everything else is only relevant clientside.  Here's [an old video](https://www.youtube.com/watch?v=GbYDvQPT2N8) showcasing an unusual use for block 36 (this predates command blocks); I think there was a commentated video that gave more context but I can't find it (the video credits [Minecraft Forum user DiEvAl](https://www.minecraftforum.net/members/DiEvAl/posts) though, so maybe it's somewhere in their posts).

Pokechu22:
I managed to dig up further context: [this video](https://web.archive.org/web/20121214131211/http://www.youtube.com/watch?v=D1IyzfEF040&gl=US&hl=en) (which is no longer public, but is archived [here](https://web.archive.org/web/20121214131235oe_/http://r1---sn-a5m7ln7s.c.youtube.com/videoplayback?upn=qlIjaJxNuco&sparams=cp%2Cgcr%2Cid%2Cip%2Cipbits%2Citag%2Cratebypass%2Csource%2Cupn%2Cexpire&fexp=926400%2C910019%2C914060%2C916625%2C922401%2C920704%2C912806%2C925703%2C928001%2C922403%2C922405%2C929901%2C913605%2C929104%2C913546%2C913556%2C920201%2C913302%2C919009%2C914903%2C911116%2C910221%2C901451%2C902556&ms=au&expire=1355515662&itag=22&ipbits=8&gcr=us&sver=3&ratebypass=yes&mt=1355490673&ip=207.241.237.184&mv=m&source=youtube&key=yt1&cp=U0hUSlVMU19LTENONF9NRldHOjJHVHZ0TFgwOWcw&id=0f5232cdf105d38d&newshard=yes&signature=87FB658D84E9B4A9B7FCB149E1B18D0C6BCA63A8.60C14D1A5B49E157D2468AB787FC6860C12DF33E) (invisible on the main page due to dependence on YouTube's flash player, I think, but also archive.org's download speed is rather slow and you're best off saving that video manually (~10 minutes), waiting for the download to finish, and then viewing it locally)).  This video is from late 2012 (9 years old, yikes!), but should be somewhat vaguely informative.  Note that in that case, `progress` is set to a negative value; from that I gather that normally `progress` starts at 0.0 and ends at 1.0.

InfRandomness:
So uh, I'm not sure where I would start of with something like this.. Trying to put together a list of things that have to be done for that overhaul is gonna be a challenge in itself

InfRandomness:
I have decided to revert everything, as a plan to avoid using an entity altogether

src/Blocks/BlockPiston.cpp
InfRandomness: I'm not sure if this is still relevant with the changes I just pushed

src/Simulator/IncrementalRedstoneSimulator/RedstoneSimulatorChunkData.h
```diff
@@ -93,6 +93,7 @@ class cIncrementalRedstoneSimulatorChunkData final : public cRedstoneSimulatorCh
 	/** Structure storing position of mechanism + it's delay ticks (countdown) & if to power on. */
 	std::unordered_map<Vector3i, std::pair<int, bool>, VectorHasher<int>> m_MechanismDelays;
 
+	std::unordered_map<Vector3i, cTickTimeLong> PistonExtensionWorldTickTime;
```
KingCol13, line 96:
```suggestion
	std::unordered_map<Vector3i, cTickTimeLong, VectorHasher<int>> PistonExtensionWorldTickTime;
```
Need the hasher to fix compile.

src/Blocks/BlockPiston.cpp
```diff
@@ -150,6 +150,12 @@ void cBlockPistonHandler::RetractPiston(Vector3i a_BlockPos, cWorld & a_World)
 				return;
 			}
 
+	if (!should_pull_block)
+	{
+		// Piston has been activated for less or equal to 2 ticks, bail out
+		// return;
```
KingCol13, line 156: Should this be uncommented?

KingCol13: also this block should have an extra indent

PR #5384, opened 2022-01-13
Fix WriteXZYPosition64 and add tests

Bond-009:
Added tests for both
Noticed and fixed a bug in WriteXZYPosition64
[godbolt](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAA2AHZydwAZAiZsADk/ACNsUikecgAHdCViFyYvX38g0Iys5yFI6Li2ROSeVIdsJxyRIhZSIjy/AJD7bEdSpmbWonLYhKSU%2Bxa2joLupSnhqNGq8dqASnt0H1JUTi4iAE807BwaAGos4GjMM/ohYFv7s4BJYUDpMwBmGy0AQUPjqcLgQriczlEiGcodCXsJPhYvj9fmY/gCTthzj4mJdro8THiHgpXkR3oiUf8jujMdiQbiITDoUS4QjvuTyfF0N4zgAlbAsTAADQAmgAtVSZbJCd4QYnvM4ANQkPmw5HBwjOlkCZxYAH0rHdUABrAWq%2Bma7V6g2GoWm9Xm3X69BGkVrNnBJEogCcAHpvWcvEwAG5JSFEJDYM5y0gNbAEYM3eLEJRqkhnT5nDBkTBKEBsrSeplEeFnR1GgXclgAdw1nwAImcIIrfBGvrhW2mABxrDUWLVadRaT40YcjxFQ31nAAq6DSZ17Z0TRCUeYLxOLpetFerX3rjaVLc%2BbcPc8C3fN/a0o%2B%2BDLOE4AsgRMJhGGceBYF0mV4X11aRVua7uTbKj2fYDkOI7Dq6143refpWOgRAkGwJ4fkuebIvmE7POcYYRkgILhvMqFqmcuFnEwYykGAXDJqQxjABGBDJioRCqhgQYhqR4bkZRKboOR2DACwzjBrmfxeg6VoCgBDYQBu5ZViBZwXhYWhqepZ51juymaQAYhcLTOKgOriPMXweBCrZyVJW7diAZwALQQBe0jqepjkGcJBDGaZRDmZZh7WU6xq2VBHr5pJwVCjJUAbkK/7ngOHZqZptbaVo3Ywl8%2BnzF5PksGZnwWcIVlxbZZz2U5F48O50IOZ5RkmQVflFQFuBBUa8VVmsYUrpFzoxR1hp/opiVaKpbkZQB6V6Q13lNYVxUte1G4jZWdmOc5A6uZNHm5Y1vn%2BSVgWraFZLifm0ZENsTCkaQyrnciwS1lwGz0NwACs/ABFwOjkOg3AeLYtgXFsOwthYnx8OQRDaK9GyGiAH0WAAdO8nrBJ8gSfB2HaBIEeNY4Y3DSPwbBI1oaMdp8nzBG%2Bnodq5H3SKEP1/QDXD8DmWgw3DGxwLAKAYDg%2BDEGQlDUHQjCsBw3DQ4IwhiBInAyHIwjKGomi/XoIAWIY9FoMD1g9H0ORuEwnjeJ0uthBbIyVNUUgfekEr9DMAR68UkpMPblFOybDT9IM0xWwUev1I0QjB4sFR%2Bzwzu5SH%2BQe5MQy%2BysNQfRsShg7s3BokCOJgnc%2BIlw8sofKyqKUoXtJgvSN5riy4UUoCGJnFiRc3GXBJnIWpJV63VIdzSoI3A3DLfs36EclyvL8sKYquzk0oVwq%2B62pC9qWsFJopkp/XWpvB8786rrie6K4TgGwZtFxEZRjGcZgouyYQvx6aZqQ2bkjCE66QQUgRFKwRiUPhGgoZwzIVTGREgs5UwsXvsCK4xEohIIwKQaMTgzhe36H1U%2BxoZL7XmodVqx12pAQPEeXAnYwrjj9NOWc85X74LikQwyJDmpHWWhAYh%2BVFoVysnwhaLUPD90rhQ/cmkgZFRPJpahZxkZ0NgmcB8T4XxvlQsuC6npD4inYXlER3ChEcP4aIwRgVhGkLERYyRzZpHmVoTWBRONlETngohdAyFmGfj%2BDCK6N07oPUHmYZ6r1iZcC%2BuQNm/AOZA2sCDHO2xdg9nhPwWG2serkHDPycYEA3okzJiAaQnoUZQ3xhYDswQtDx09NID6nxolw3%2BtwLmIAeYZJ0FkxGH0eAoxqR2CwwQQiBAsK5VSnpnbvS4J8b6zSObpL5uQAWCB4AQCFl4tIDAkgSwgBgNgWzGDJGALUPWUsiBJBzBAeIzTEysFIAcOW/B9kcGEAAeSYPQR52tyA4DYPRSQPzCBYJEtgHMPzsDqAaD4C5TzKDCF6M0%2BgBB4h0QeV4HAzSiCkAIOTXg4SaBGGAEoeUBBsCVjeccH68t5BK0kKrBWigVAaGafoVIRgTCGwScbZF8QcyQA2DOfo4KHJvNmQHSOAQIDuHdgYCISwHarBdiUHIsrUi4JyOnR2dReiByaAsNVEqg4LC1UqxO7RQ4BDqCahVcds65xVjDaMew%2BAFMiXMn5HN1B4wcnKYAqBUCvj6e%2BCA8SbDGzOKLEgpBUmpH9Js7ZMbLBQzWIszJGwck4GSPkiJpNyDkw%2BjzGJLTOb2HabzTJyzEDrLQAmo5uz9mHPGCct8AgGAXKAdQG5Py7mtG%2BdDF5zAiAfK%2Bc0v5AK9h/WBXq4M4K/qQuhbC/F8KLnTL%2BrytFBwMWTvSTivFrqBBEpJWSilVK4WMrpSrWQjKNYsp%2BfoPWHLTBGzsLy/lOahU5BFWKrmurJXm0tsnOVdtbUZwMBqoQhqIM%2B1A9qo1%2BqhiGojsatOsGzUGstQYc1pqaj2uSZwPW2LsAuvCdMqJxavU%2Br9QGoNKMQ1htsKqKNZBUl63jQcxNrHU0Vq6QjEAUMUYfWGVoaQb4eCfC0MEKp7xc1FMLU0z1rSy0dL5nx6QHYUYMx4MESGkNggfQxp6LQUzuDioo0pzp8MIkWA9ezCzqnyC3yyK4aQQA%3D)

### DIFF
--- a/src/ByteBuffer.cpp
+++ b/src/ByteBuffer.cpp
@@ -788,7 +788,7 @@ bool cByteBuffer::WriteXZYPosition64(Int32 a_BlockX, Int32 a_BlockY, Int32 a_Blo
 	CheckValid();
 	return WriteBEInt64(
 		(static_cast<Int64>(a_BlockX & 0x3FFFFFF) << 38) |
-		(static_cast<Int64>(a_BlockZ & 0x3FFFFFF) << 26) |
+		(static_cast<Int64>(a_BlockZ & 0x3FFFFFF) << 12) |
 		(static_cast<Int64>(a_BlockY & 0xFFF))
 	);
 }
--- a/tests/ByteBuffer/ByteBufferTest.cpp
+++ b/tests/ByteBuffer/ByteBufferTest.cpp
@@ -70,8 +70,40 @@ static void TestWrap(void)
 
 
 
+static void TestXYZPositionRoundtrip(void)
+{
+	cByteBuffer buf(50);
+	buf.WriteXYZPosition64(-33554432, -2048, -33554432); // Testing the minimun values
+	int x, y, z;
+	TEST_TRUE(buf.ReadXYZPosition64(x, y, z));
+	TEST_EQUAL(x, -33554432);
+	TEST_EQUAL(y, -2048);
+	TEST_EQUAL(z, -33554432);
+}
+
+
+
+
+
+static void TestXZYPositionRoundtrip(void)
+{
+	cByteBuffer buf(50);
+	buf.WriteXZYPosition64(-33554432, -2048, -33554432); // Testing the minimun values
+	int x, y, z;
+	TEST_TRUE(buf.ReadXZYPosition64(x, y, z));
+	TEST_EQUAL(x, -33554432);
+	TEST_EQUAL(y, -2048);
+	TEST_EQUAL(z, -33554432);
+}
+
+
+
+
+
 IMPLEMENT_TEST_MAIN("ByteBuffer",
 	TestRead();
 	TestWrite();
 	TestWrap();
+	TestXYZPositionRoundtrip();
+	TestXZYPositionRoundtrip();
 )


src/ByteBuffer.cpp
```diff
@@ -772,7 +764,7 @@ bool cByteBuffer::WriteXYZPosition64(Int32 a_BlockX, Int32 a_BlockY, Int32 a_Blo
 	CHECK_THREAD
 	CheckValid();
 	return WriteBEInt64(
-		(static_cast<Int64>(a_BlockX & 0x3FFFFFF) << 38) |
+		(static_cast<Int64>(a_BlockX) << 38) |
```
bearbin, line 767: This is undefined behaviour, the old one was correct.

src/ByteBuffer.cpp
```diff
@@ -518,15 +514,11 @@ bool cByteBuffer::ReadXZYPosition64(int & a_BlockX, int & a_BlockY, int & a_Bloc
...
-	UInt32 BlockYRaw = (Value & 0x0fff);            // Bottom 12 bits
-
-	// If the highest bit in the number's range is set, convert the number into negative:
-	a_BlockX = ((BlockXRaw & 0x02000000) == 0) ? static_cast<int>(BlockXRaw) : (static_cast<int>(BlockXRaw) - 0x04000000);
-	a_BlockY = ((BlockYRaw & 0x0800)     == 0) ? static_cast<int>(BlockYRaw) : (static_cast<int>(BlockYRaw) - 0x01000);
-	a_BlockZ = ((BlockZRaw & 0x02000000) == 0) ? static_cast<int>(BlockZRaw) : (static_cast<int>(BlockZRaw) - 0x04000000);
+	// Convert the 64 received bits into 3 coords
+    // First we shift them to the top to set the sign bit in the correct position
+	a_BlockX = static_cast<int>(Value >> 38);  // Top 26 bits
+	a_BlockY = static_cast<int>(static_cast<Int64>(static_cast<UInt64>(Value) << 52) >> 52);  // Middle 12 bits
```
bearbin, line 520: Comment's wrong here.

bearbin (changes requested):
The undefined behaviour is incorrect, and needs to be changed. Otherwise this change does appear to reduce the number of instructions here.

Even with the UB fixed though, I'm not sure this PR is worth it: there's no point in instruction-counting unless this bit of code is particularly hot, and we're giving up strictly-defined semantics for implementation-defined semantics (which will probably be fine everywhere Cuberite runs, but I don't think it's really worth it).

Bond-009:
@bearbin In that case I'll revert it and just add the tests and fix `WriteXZYPosition64`, sounds good?
(I just copied this over from my rust project where I don't have to worry about C++ UB and friends ;) )

bearbin:
Yep, the bugfix looks fine, good catch on that.

tigerw:
Read is used in block place and block dig, using arithmetic shifts is 1.1 times faster according to [quick-bench](https://quick-bench.com/q/sqo_QlPjgs0KjFRzJTRsNrgYSVw).

bearbin:
> The current code in ReadXYZ also shifts a signed value right so no change on relying on implementation defined here, i'd say fewer instructions are always better. Agreed we still need the mask in Write, cpp20 gave up on sign-magnitude and ones-complement but we're not there yet.

Not the case. We did a signed right shift, but then threw away the carry bit. No matter what the implementation chooses, we would get the right answer with the old code. The new one was not guaranteed.

Thanks for confirming that the performance change is negligible. Merging.

bearbin (approved):

PR #5385, opened 2022-01-14
ByteBuffer: fully defined Position handling

tigerw:
Cast input ints into unsigned to make them 2s complement guaran☕d.

### DIFF
--- a/src/ByteBuffer.cpp
+++ b/src/ByteBuffer.cpp
@@ -487,8 +487,8 @@ bool cByteBuffer::ReadLEInt(int & a_Value)
 bool cByteBuffer::ReadXYZPosition64(int & a_BlockX, int & a_BlockY, int & a_BlockZ)
 {
 	CHECK_THREAD
-	Int64 Value;
-	if (!ReadBEInt64(Value))
+	UInt64 Value;
+	if (!ReadBEUInt64(Value))
 	{
 		return false;
 	}
@@ -512,8 +512,8 @@ bool cByteBuffer::ReadXYZPosition64(int & a_BlockX, int & a_BlockY, int & a_Bloc
 bool cByteBuffer::ReadXZYPosition64(int & a_BlockX, int & a_BlockY, int & a_BlockZ)
 {
 	CHECK_THREAD
-	Int64 Value;
-	if (!ReadBEInt64(Value))
+	UInt64 Value;
+	if (!ReadBEUInt64(Value))
 	{
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


12xx12 (approved):

src/ByteBuffer.cpp
```diff
@@ -771,10 +771,10 @@ bool cByteBuffer::WriteXYZPosition64(Int32 a_BlockX, Int32 a_BlockY, Int32 a_Blo
...
 	CHECK_THREAD
 	CheckValid();
-	return WriteBEInt64(
-		(static_cast<Int64>(a_BlockX & 0x3FFFFFF) << 38) |
-		(static_cast<Int64>(a_BlockY & 0xFFF) << 26) |
-		(static_cast<Int64>(a_BlockZ & 0x3FFFFFF))
+	return WriteBEUInt64(
+		((static_cast<UInt64>(a_BlockX) & 0x3FFFFFF) << 38) |
+		((static_cast<UInt64>(a_BlockY) & 0xFFF) << 26) |
+		(static_cast<UInt64>(a_BlockZ) & 0x3FFFFFF)
```
bearbin, line 777: This just switches an unambiguously valid and defined left-shift to a confusing cast of a a signed integer to unsigned - although the result will be the right value it reles on befuddling semantics of casting negative integers to unsigneds.

bearbin (commented):
The changes to ReadXXXPosition64 are fine: although the previous code was strictly defined already, the new changes do make it easier to understand.

The WriteBEUInt64 changes are just confusing: casting negative ints to unsigneds requires actual understanding of the spec and the previous code was strictly defined already! I'd rather drop those changes.

src/ByteBuffer.cpp
tigerw: What if Int32 a_BlockX was a sign-magnitude representation? The protocol needs 2s complement! :o

[from here](https://timsong-cpp.github.io/cppwp/n4659/conv.integral)
> If the destination type is unsigned, the resulting value is the least unsigned integer congruent to the source integer (modulo 2n where n is the number of bits used to represent the unsigned type). [ Note: In a two's complement representation, this conversion is conceptual and there is no change in the bit pattern (if there is no truncation).  — end note ]

So it ensures a_BlockX is in 2s complement representation.

nb this is an exercise in amusement, not bugfixing

bearbin: Luckily C++20 mandates 2s complement, we don't have to worry about sign-magnitude anymore :)

tigerw: aye. nonetheless since we're interpreting a value and writing it out into serialised form as a sequence of bytes a u64 is appropriate for that.

bearbin:
Oh whatever, this PR is pointless, let's get it merged so we don't waste any more time.

PR #5388, opened 2022-01-22
Improving 1.14 support

Persson-dev:
I improved 1.14 protocol support and it seems to work.
I just have a strange issue : sometimes on the client, an entire chunk disappear with no reason and reappear when I place a block in that chunk.
This PR is not complete but it allows 1.14 clients to join.


bearbin (approved):
Looks good, thanks for the changes!

bearbin:
Just need to fix the build, and does @tigerw or @12xx12 want to review?

12xx12:
I like this PR.

Maybe we also finally should finish the 1.13 protocol support also. 

Review coming up soon

Persson-dev:
It should be all good now. The client does not calculate light anymore so it should be sent through the new update light packet, otherwise it's playable.

src/Protocol/Protocol_1_14.cpp
```diff
@@ -104,8 +206,84 @@ void cProtocol_1_14::SendPaintingSpawn(const cPainting & a_Painting)
...
+	Pkt.WriteBEInt32(GetProtocolParticleID(a_ParticleName));
+	Pkt.WriteBool(false);
+	Pkt.WriteBEFloat(a_SrcX);
+	Pkt.WriteBEFloat(a_SrcY);
+	Pkt.WriteBEFloat(a_SrcZ);
+	Pkt.WriteBEFloat(a_OffsetX);
+	Pkt.WriteBEFloat(a_OffsetY);
+	Pkt.WriteBEFloat(a_OffsetZ);
+	Pkt.WriteBEFloat(a_ParticleData);
+	Pkt.WriteBEInt32(a_ParticleAmount);
```
12xx12, line 223: This seems to be exactly the same code as the method it overrides

src/Protocol/Protocol_1_14.cpp
```diff
@@ -104,8 +206,84 @@ void cProtocol_1_14::SendPaintingSpawn(const cPainting & a_Painting)
...
 {
+	ASSERT(m_State == 3);  // In game mode?
+
+	// Note: Particles from block break missing
+
+	cPacketizer Pkt(*this, pktSoundParticleEffect);
+	Pkt.WriteBEInt32(static_cast<int>(a_EffectID));
+	Pkt.WriteXYZPosition64(a_SrcX, a_SrcY, a_SrcZ);
+	Pkt.WriteBEInt32(a_Data);
+	Pkt.WriteBool(false);
```
12xx12, line 286: This is exactly the same code as some code it overrides. Was the method of writing the position changed, maybe? (XYZ ->  XZY)

12xx12 (changes requested):
Hi and Sorry the review took so long. I have a few minor comments. I'm happy this is moving forward

src/Protocol/Protocol_1_14.cpp
Persson-dev: It's indeed, the same code but the call to GetProtocolParticleID in the 1.14 protocol returns the particle in 1.14 protocol. While in the 1.13 protocol class, it's calling GetProtocolParticleID of the 1.13 protocol. Maybe I misunderstood that ?

src/Protocol/Protocol_1_14.cpp
Persson-dev: Same thing apply here for the call of GetProtocolParticleID.

src/Protocol/Protocol_1_14.cpp
```diff
@@ -21,11 +50,31 @@ Implements the 1.14 protocol classes:
...
 
 
+#define HANDLE_READ(ByteBuf, Proc, Type, Var) \
+	Type Var; \
+	do { \
+		if (!ByteBuf.Proc(Var))\
+		{\
+			return;\
+		} \
+	} while (false)
```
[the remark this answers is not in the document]
Persson-dev: I just did the same thing as the others protocol source files but i can move it to the header Protocol.h if you want.

src/Protocol/Protocol_1_14.cpp
12xx12: Hm. I'm not 100% percent sure about inheritance in this case. Anyway you already have a positive review on your PR so that should be fine

src/Protocol/Protocol_1_14.cpp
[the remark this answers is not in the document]
12xx12: Nono, this wasn't meant especcially for you. If you want to, you can open another PR and fix this.

Persson-dev:
I think the bug that I mentioned with the chunks disappearing is caused by the heightmap but i'm not sure.
This needs more testing.
I tried to make Cuberite send the heightmap to the client but for some reason, the client just crashed. Maybe i did something wrong ?
Anyway, the rest seems to work.

12xx12:
Maybe create another issue describing your findings

12xx12:
I tried this again and when connecting a packet with too much data is sent :( did a complete new build and new world and ...

Sadly I can't tell which packet it is

src/Protocol/Protocol_1_14.cpp
bearbin: Yep, as the GetProtocolParticleID is static rather than virtual the implementation of this function needs to be copied. Ideally the function would become virtual if it needs to be used in this way.

PR #5389, opened 2022-01-22
Added newline checking functionality to IniFile.cpp In the Header Comments Section

JoeParimucha:


CONTRIBUTORS
```diff
@@ -92,3 +92,4 @@ worktycho
 Xenoxis
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+JoeParimucha
```
bearbin, line 95: This should be added at the correct alphabetical position, rather than at the end.

bearbin (changes requested):
I'm sorry but this code doesn't make any sense. You seem to be splitting comments on a newline when writing them out - which means that reading in and writing out the same file will lead to corruption. 

The behaviour that is wanted is escaping newlines on output so a `\n` is output instead, and unescaping `\n` to a real newline on input.

I'm also a bit skeptical of this approach overall - the INI file parser is a very simple manual parser at the moment, which is fine. Adding loads more functionality to it, though, will inevitably lead to bugs and inconsistencies - it would probably be better to get a better library in instead or a proper generated parser.

bearbin:
I'm going to close this as it needs rework, and I haven't heard anything back. Feel free to reopen if you pick this back up again.

PR #5391, opened 2022-02-07
Spectator fixes

tigerw:
+ Keep a list of spectators so that pointer clean-up can happen when the spectated is killed.
* Fix invalid game state when riding/spectating and then entering or leaving spectator mode.
* Prevent spectators from using items.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1498,10 +1498,11 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 
 		const auto & ItemHandler = HeldItem.GetHandler();
 		const auto & BlockHandler = cBlockHandler::For(BlockType);
+		const bool BlockUsable = BlockHandler.IsUseable() && (m_Player->IsGameModeSpectator() ? cBlockInfo::IsUseableBySpectator(BlockType) : !(m_Player->IsCrouched() && !HeldItem.IsEmpty()));
 		const bool Placeable = ItemHandler.IsPlaceable() && !m_Player->IsGameModeAdventure() && !m_Player->IsGameModeSpectator();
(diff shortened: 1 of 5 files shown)


12xx12 (approved):

src/ClientHandle.cpp
```diff
@@ -1498,10 +1498,11 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 
 		const auto & ItemHandler = HeldItem.GetHandler();
 		const auto & BlockHandler = cBlockHandler::For(BlockType);
+		const bool BlockUsable = BlockHandler.IsUseable() && (m_Player->IsGameModeSpectator() ? cBlockInfo::IsUseableBySpectator(BlockType) : !(m_Player->IsCrouched() && !HeldItem.IsEmpty()));
 		const bool Placeable = ItemHandler.IsPlaceable() && !m_Player->IsGameModeAdventure() && !m_Player->IsGameModeSpectator();
-		const bool BlockUsable = BlockHandler.IsUseable() && (!m_Player->IsGameModeSpectator() || cBlockInfo::IsUseableBySpectator(BlockType));
+		const bool Useable = !m_Player->IsGameModeSpectator();
```
bearbin, line 1503: Naming - this is not usability - merely spectator status. If you're going to extract this to a variable, it should be reused in the BlockUsable calculation and Placeable calcuations above also.

tigerw: I added another condition to the if-ladder line 1541, "usability", where before it would unconditionally use the item. From what I know at the moment all items are usable apart from when you're a spectator. But if I'm wrong and the condition is actually more complex or it changes in the future following the bool Placeable/BlockUsable patterns makes sense. I can separately extract IsGameModeSpectator to a variable if you want.

bearbin: Then this should be ItemUsable, not just Useable. Extracting to a seperate variable is :shrug: but it would be nice.

bearbin (approved):

Issue #5333, opened 2021-11-20
Double tall grass drops seeds when broken in creative mode


AirOne01:
Client version: 1.12.2
Server OS: Windows 11
Cuberite Commit id: 95bc2eac222a12c049b1aef799ac280fa9123af4

### Expected behavior
When breaking double tall grass in creative, nothing is dropped

### Actual behavior
Seed sometimes drop when breaking double tall grass

### Steps to reproduce the behavior
Break double tall grass.

bearbin:
I believe this is actually vanilla.

AirOne01:
I just realised my issue can be missunderstood, sorry about that. I was actually refering to *double* tall grass.

tigerw:
Easy because cBlockBigFlowerHandler literally calls DropBlockAsPickups in its OnBroken handler. This seems deliberate, so we mainly need to figure out if there's justification to do that and change to SetBlock(AIR) if not.

QDaily117:
Are you guys still wanting this change to happen? I would be more than happy to make this change for you guys

tigerw:
Yes, thanks! I agree it's a bug.

RodargDev:
@bearbin this has been resolved and can be closed

PR #5392, opened 2022-02-07
Double tall grass drops seeds when broken in creative mode fix

RodargDev:
Issue: #5333 

Removed if statement checking if the top part of the block was broken this would prevent seed dropping if the top was broken, this is why the dropBlockAsPickups was added but that isn't very consistent with other models. So I replaced the dropBlockAsPickups to SetBlock(Air) and now no seeds drop in creative and in survival seeds drop for both the top part of the grass and bottom part. 

Ive been testing it and i think the problem has been resolved

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -66,6 +66,7 @@ Persson-dev
 pokechu22
 ProjectBM
 pwnOrbitals
+Rodarg
 Rorkh
 rs2k
 SamJBarney
@@ -92,4 +93,3 @@ worktycho
 Xenoxis
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
-Rodarg
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -92,3 +92,4 @@ worktycho
 Xenoxis
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+Rodarg
--- a/src/Blocks/BlockBigFlower.h
+++ b/src/Blocks/BlockBigFlower.h
@@ -125,9 +125,12 @@ private:
 			auto lowerPartPos = a_BlockPos - Vector3i(0, 1, 0);
 			if (a_ChunkInterface.GetBlock(lowerPartPos) == a_OldBlockType)
 			{
-				if (a_Digger->IsPlayer() && (static_cast<const cPlayer *>(a_Digger))->IsGameModeCreative()) {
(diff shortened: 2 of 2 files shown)


src/Blocks/BlockBigFlower.h
```diff
@@ -49,10 +49,6 @@ class cBlockBigFlowerHandler final :
 
 	virtual cItems ConvertToPickups(const NIBBLETYPE a_BlockMeta, const cItem * const a_Tool) const override
 	{
-		if (IsMetaTopPart(a_BlockMeta))
```
tigerw, line 52: The top half of big flowers don't store any flower type metadata, so I suspect with these changes punching the top half no longer gives the right drops.

src/Blocks/BlockBigFlower.h
```diff
@@ -125,7 +121,7 @@ class cBlockBigFlowerHandler final :
 			auto lowerPartPos = a_BlockPos - Vector3i(0, 1, 0);
 			if (a_ChunkInterface.GetBlock(lowerPartPos) == a_OldBlockType)
 			{
-				a_ChunkInterface.DropBlockAsPickups(lowerPartPos);
```
tigerw, line 128: You could keep dropping the bottom as a pickup but that causes the creative dropping issue. So _that's_ why it happens...

I guess you'll need to check if `a_Digger->IsPlayer()` and if `static_cast<cPlayer *>(a_Digger)->IsGameModeCreative()` then `SetBlock` else `DropBlockAsPickups`.

RodargDev:
@tigerw while testing this i found a bug with the spawn protection and tall grass. When breaking the top part of a tall grass block in spawn protection it turns into a floating flower.  After relogging it will be tall grass again. I will probaply create another issue for that because i cant seem to figure out whats wrong there

tigerw:
Spawn protection block re-send is buggy:
https://github.com/cuberite/cuberite/blob/5fe1fe899bfe55b1c27fb03a7d7e76b0b67b9cc7/src/ClientHandle.cpp#L1368-L1369

Clearly won't re-send the bottom block if you break the top half. All instances of SendBlockTo in that file need to be replaced with SendBlocksAround(radius: 2) :p

Would you like to add yourself to CONTRIBUTORS (in the root folder)?

RodargDev:
@tigerw added my name to the list !

CONTRIBUTORS
```diff
@@ -92,3 +92,4 @@ worktycho
 Xenoxis
 xoft (Mattes Dolak/madmaxoft on GH)
 Yeeeeezus (Donated AlchemistVillage prefabs)
+Rodarg
```
bearbin, line 95: The list is alphabetical :)

RodargDev: Lol such a retard hahah just thought i would put it on the end of the list

bearbin (approved):

PR #5396, opened 2022-03-22
Below bedrock break crash

### LINKED ISSUE #5395: Server crashes when mining bedrock with hack client
Client version: **Every client/ Server side bug**
Server OS: **Linux, Debian 11 bullseye Openvz virt**
Cuberite Commit id: (from commit **de5b89dbea938db5a7ea58f59a9b1a91d0df4208** built at: Thu Feb 17 19:02:07 UTC 2022t)

https://user-images.githubusercontent.com/101947415/159139880-bb78e427-9115-4f79-9f20-43fec9d15ffc.mp4

KingCol13:
Fixes #5395.

### DIFF
--- a/src/Blocks/BlockPiston.cpp
+++ b/src/Blocks/BlockPiston.cpp
@@ -215,6 +215,12 @@ bool cBlockPistonHandler::CanPushBlock(
 	Vector3iSet & a_BlocksPushed, const Vector3i & a_PushDir
 )
 {
+	if (!cChunkDef::IsValidHeight(a_BlockPos.y))
+	{
+		// Can't push a void block.
+		return false;
+	}
+
 	const static std::array<Vector3i, 6> pushingDirs =
 	{
 		{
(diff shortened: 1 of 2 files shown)


src/Blocks/BlockPiston.cpp
```diff
@@ -228,11 +234,6 @@ bool cBlockPistonHandler::CanPushBlock(
 	NIBBLETYPE currMeta;
 	a_World.GetBlockTypeMeta(a_BlockPos, currBlock, currMeta);
 
-	if (!cChunkDef::IsValidHeight(a_BlockPos.y))
-	{
-		return !a_RequirePushable;
```
KingCol13, line 233: I didn't really understand `a_RequirePushable`, to me `CanPushBlock` should just return whether the block can be pushed independent on whether we need it to be pushed or not. Maybe @tigerw can shed some light on this.

tigerw: `RequirePushable` should be named` IsPushing`! When retracting, `IsPushing == false` which changes the behaviour of `CanPush`, making it ignore a piston-breakable block instead of killing it. In this case I agree returning false is fine, this will stop extension and leave retraction unaffected.

tigerw: this file could do with some improvements to clarity, the logic is dense...

tigerw (approved):

tigerw:
thanks and apologies!

Issue #5220, opened 2021-05-08
API: Calling this function with a high y position crashes the server


Seadragon91:
Server OS: Linux
Cuberite Commit id: ce8d8388d630fa936500e6843496afe113ad8bc5

### Actual behavior
Server crashes.

### Steps to reproduce the behavior
Call this code from a plugin. The chunk has to be loaded, join the server and teleport to `0 300 0`.
`cRoot:Get():GetDefaultWorld():GrowTreeFromSapling(Vector3d(0, 1000, 0))`

### Stack trace
```
#0  0x000000000042a933 in ChunkDataStore<unsigned char, 2048ul, (unsigned char)0>::Get(Vector3<int>) const ()
#1  0x000000000042d22f in cChunkMap::GetBlockMeta(Vector3<int>) const ()
#2  0x00000000004ada5d in cWorld::GrowTreeFromSapling(Vector3<int>) ()
#3  0x00000000005349c2 in tolua_cWorld_GrowTreeFromSapling(lua_State*) ()
#4  0x0000000000796a5e in luaD_precall(lua_State*, lua_TValue*, int) ()
#5  0x00000000007a7b73 in luaV_execute(lua_State*, int) ()
#6  0x0000000000796ec9 in luaD_call(lua_State*, lua_TValue*, int) ()
#7  0x00000000007971d6 in luaD_pcall(lua_State*, void (*)(lua_State*, void*), void*, long, long) ()
#8  0x0000000000792e7e in lua_pcall ()
#9  0x000000000053c64e in cLuaState::CallFunction(int) ()
[...]

PR #5397, opened 2022-03-24
Fix cChunkMap issues below with coords below y=0

### LINKED ISSUE #5395: Server crashes when mining bedrock with hack client

hle0:
Should fix #5395 and some other issues. See discussion in #5220. I noticed that lighting the void on fire (using a flint and steel on the side of bedrock at y=0) crashed the server before this PR, but not after. This should fix most other similar issues.


bearbin:
Sorry for the delay in reviewing, I was a bit busy - I'll get round to it tomorrow.

src/ChunkMap.cpp
```diff
@@ -585,6 +595,11 @@ void cChunkMap::SetBlock(Vector3i a_BlockPos, BLOCKTYPE a_BlockType, NIBBLETYPE
 
 bool cChunkMap::GetBlockTypeMeta(Vector3i a_BlockPos, BLOCKTYPE & a_BlockType, NIBBLETYPE & a_BlockMeta) const
 {
+	if (!cChunkDef::IsValidHeight(a_BlockPos.y))
+	{
+		return false;
```
bearbin, line 600: I have a slight concern here that we might end up with use of uninitialised memory as a consequence of this return - idiomatic use of GBTM does not initialise the arguments, and with an early return they are not initialised here either. 

I think we should zero them here for safety.

bearbin (commented):
This looks good generally. I think it might be slightly better though if we created new functions that do this checking, cWorld.SafeBlockTypeMeta or something - otherwise the behaviour of chunks' GBTM and worlds' GTBM functions will be different which could lead to confusion.

src/ChunkMap.cpp
hle0: I could be wrong, but I think there are other cases where early return does not initialize the arguments; for example, if `!Chunk->IsValid()`. I don't entirely understand when a Chunk would be invalid, but that's another case where the arguments will be left uninitialized. Also, most (all?) C++ compilers should zero the values anyway afaik, since they're enums, although I'm not totally sure.

bearbin: You're right, we need to add an initialisation to the invalid chunk case as well. Most C++ compilers do zero-initialise variables, but that is non-standard undefined behaviour which we should not rely on.

madmaxoft: I think the modus operandi "If the return value is `false`, then the out params are uninitialized" is valid. It should just probably be mentioned in the doxy-comment. All the callsites should be doing `if (!GetBlockTypeMeta(...)) { return ...; }` so the uninitialized values don't matter.

madmaxoft: Perhaps, is there some marker that would tell the compilers that ignoring the return value of this function is a bad thing?

bearbin (approved):
Reviewing this so we can close out some stale PRs. 

I still think that this is a slightly flawed idea; we also need to do the same for the SetBlock functions. 

But better to fix the immediate issue and let us refactor later.

tests/CompositeChat/CompositeChatTest.cpp
```diff
@@ -109,6 +109,7 @@ static void TestParser5(void)
 
 
 
+
```
bearbin, line 112: This was a style violation caught by my pre-commit hook during the merge; easier just to fix it.

PR #5398, opened 2022-03-26
fix various compiler warnings

hle0:
Fixed some compiler warnings I was getting with gcc 11.2.0. There were also some bugs that these compiler warnings caught, and I've fixed those.

### DIFF
--- a/src/Bindings/LuaServerHandle.cpp
+++ b/src/Bindings/LuaServerHandle.cpp
@@ -143,7 +143,7 @@ cTCPLink::cCallbacksPtr cLuaServerHandle::OnIncomingConnection(const AString & a
 	cCSLock Lock(m_CSConnections);
 	m_Connections.push_back(res);
 
-	return std::move(res);
+	return res;
 }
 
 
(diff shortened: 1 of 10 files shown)


src/Bindings/LuaServerHandle.cpp
```diff
@@ -143,7 +143,7 @@ cTCPLink::cCallbacksPtr cLuaServerHandle::OnIncomingConnection(const AString & a
 	cCSLock Lock(m_CSConnections);
 	m_Connections.push_back(res);
 
-	return std::move(res);
```
bearbin, line 146: What's the warning generated here? This should be valid I would have thought.

DarkoGNU: `warning: redundant move in return statement [-Wredundant-move]`

[Return value optimization](https://sigcpp.github.io/2020/06/08/return-value-optimization)

bearbin: Well that warning is a bit useless, it's not exactly like this does any harm, and the criteria for when variables don't get moved  isn't a super simple mnemonic... But this change is fine.

bearbin (approved):

PR #5401, opened 2022-04-03
Implement farmland trampling

### LINKED ISSUE #1179: Farmland doesn't revert to grass when landing on it from a jump
When a player or a mob lands on farmland after falling from sufficient height ("jump"), farmland should revert back to dirt and anything growing on top should be lost..

### LINKED ISSUE #5069: Farmland stays when jumping
Client version: 1.11.2
Server OS: Linux
Cuberite Commit id: Cuberite local build (id: Unknown)

### Expected behavior
When jumping on farmland, the block should be reset to dirt.

### Actual behavior
Nothing happens

Jump on farmland

### LINKED ISSUE #5402: You can fall through farmland when it naturally turns into dirt
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: de5b89dbea938db5a7ea58f59a9b1a91d0df4208

### Expected behavior

When farmland turns into dirt (because it's dry and there's no water sources nearby), you should stay on top of the block.

### Actual behavior

When farmland turns into dirt, you fall through it.

1. Create some farmland. Make sure there's no water source nearby and it doesn't rain.
2. Stand on it.
3. Wait patiently.
4. You've fallen through it.

### Additional notes

Vanilla Minecraft also had a similar bug - see [MC-104259](https://bugs.mojang.com/browse/MC-104259). I think I'll fix this bug when working on #5401, as I don't want farmland trampling to be buggy.

### Screenshots

First
Second
(cute!)

Third
Fourth

DarkoGNU:
Just a draft right now. I've encountered a tiny issue...

So, I'm implementing farmland trampling. I've determined experimentally chances of trampling (described in #5069).

I've decided to create a function cPawn::HandleFarmlandTrampling() that will be responsible for handling trampling. It will be called in cPawn::HandleFalling(). Its body is empty at the moment (I need to implement it). Let me know if there's something wrong with this approach - I'll be happy to implement trampling in another way.

The issue I'm facing - cPawn::m_LastGroundHeight isn't accurate, it misbehaves with thick snow. I'll try to explain the issue with some screenshots including debug messages. These piles of snow are 0.75 blocks high, so you can't just walk onto them.
Overview
Falling from the left pile of snow (without jumping onto it):
Left
Falling from the right pile of snow (with jumping onto it):
Right
So, m_LastGroundHeight depends on whether I jump onto snow or walk into it.

I can try to fix m_LastGroundHeight in this pull request or fix it in the future (another PR) and just implement trampling right now. Trampling should generally work OK without the fix, as the issue doesn't affect slabs - and it's probably pretty rare to have multilayered snow near farms. Trampling will be harder to thoroughly test without the fix (look at #5069) though.

I'd also like to make it possible to turn off trampling in world.ini.

Fixes #5069
Fixes #1179
(I think they're duplicates)
Fixes #5402

### DIFF
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -161,7 +161,7 @@ private:
 		auto upperBlock = a_ChunkInterface.GetBlock(a_BlockPos.addedY(1));
 		if (cBlockInfo::FullyOccupiesVoxel(upperBlock))
 		{
-			// Until the fix above is done, this line *should* also suffice:
+			// Until the fix above is done, this line should also suffice:
 			// a_ChunkInterface.SetBlock(a_BlockPos, E_BLOCK_DIRT, 0);
 			a_ChunkInterface.DoWithChunkAt(a_BlockPos, [&](cChunk & Chunk)
 			{
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -25,19 +25,10 @@ public:
 
 	using Super::Super;
 
-	void TurnToDirt(cChunk & a_Chunk, Vector3i a_RelPos) const
-	{
-		TurnToDirt(a_Chunk, a_RelPos, a_Chunk.RelativeToAbsolute(a_RelPos));
-	}
-
-
-
-
-
 	/** Turns farmland into dirt.
 	Will first check for any colliding entities and teleport them to a higher position.
 	*/
-	void TurnToDirt(cChunk & a_Chunk, Vector3i a_RelPos, Vector3i a_AbsPos) const
+	static void TurnToDirt(cChunk & a_Chunk, Vector3i a_RelPos, Vector3i a_AbsPos)
 	{
 		static const auto FarmlandHeight = cBlockInfo::GetBlockHeight(E_BLOCK_FARMLAND);
 		static const auto FullHeightDelta = 1 - FarmlandHeight;
@@ -108,7 +99,8 @@ private:
 			}
 			default:
 			{
-				TurnToDirt(a_Chunk, a_RelPos);
+				auto AbsPos = a_Chunk.RelativeToAbsolute(a_RelPos);
+				TurnToDirt(a_Chunk, a_RelPos, AbsPos);
 				break;
 			}
 		}
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -25,6 +25,11 @@ public:
 
 	using Super::Super;
 
+	void TurnToDirt(cChunkInterface & a_ChunkInterface, Vector3i a_BlockPos) const
+	{
+		a_ChunkInterface.SetBlock(a_BlockPos, E_BLOCK_DIRT, 0);
+	}
+
 private:
 
 	virtual cItems ConvertToPickups(const NIBBLETYPE a_BlockMeta, const cItem * const a_Tool) const override
@@ -76,7 +81,7 @@ private:
 			}
 			default:
 			{
-				a_Chunk.SetBlock(a_RelPos, E_BLOCK_DIRT, 0);
+				TurnToDirt(a_ChunkInterface, a_RelPos);
 				break;
 			}
 		}
@@ -104,7 +109,7 @@ private:
 		auto upperBlock = a_ChunkInterface.GetBlock(a_BlockPos.addedY(1));
 		if (cBlockInfo::FullyOccupiesVoxel(upperBlock))
 		{
-			a_ChunkInterface.SetBlock(a_BlockPos, E_BLOCK_DIRT, 0);
+			TurnToDirt(a_ChunkInterface, a_BlockPos);
 		}
 	}
 
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -34,7 +34,7 @@ public:
 
 
 
-	/** Turns farmland into dirt. 
+	/** Turns farmland into dirt.
 	Will first check for any colliding entities and teleport them to a higher position.
 	*/
 	void TurnToDirt(cChunk & a_Chunk, Vector3i a_RelPos, Vector3i a_AbsPos) const
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -12,6 +12,7 @@
 #include "BlockHandler.h"
 #include "ChunkInterface.h"
 #include "../BlockArea.h"
+#include "../Chunk.h"
 
 
 
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "BlockHandler.h"
+#include "ChunkInterface.h"
 #include "../BlockArea.h"
 
 
@@ -28,12 +29,12 @@ public:
 	/** Turns farmland into dirt.
 	Will first check for any colliding entities and teleport them to a higher position.
 	*/
-	static void TurnToDirt(cChunk & a_Chunk, Vector3i a_RelPos, Vector3i a_AbsPos)
+	static void TurnToDirt(cWorld & a_World, Vector3i a_AbsPos)
 	{
 		static const auto FarmlandHeight = cBlockInfo::GetBlockHeight(E_BLOCK_FARMLAND);
 		static const auto FullHeightDelta = 1 - FarmlandHeight;
 
-		a_Chunk.ForEachEntityInBox(
+		a_World.ForEachEntityInBox(
 			cBoundingBox(Vector3d(0.5, FarmlandHeight, 0.5) + a_AbsPos, 0.5, FullHeightDelta),
 			[&](cEntity & Entity)
 			{
@@ -41,7 +42,7 @@ public:
 				return false;
 			});
 
-		a_Chunk.SetBlock(a_RelPos, E_BLOCK_DIRT, 0);
+		a_World.SetBlock(a_AbsPos, E_BLOCK_DIRT, 0);
 	}
 
 
@@ -99,8 +100,9 @@ private:
 			}
 			default:
 			{
+				auto World = a_Chunk.GetWorld();
 				auto AbsPos = a_Chunk.RelativeToAbsolute(a_RelPos);
-				TurnToDirt(a_Chunk, a_RelPos, AbsPos);
+				TurnToDirt(*World, AbsPos);
 				break;
 			}
 		}
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -28,6 +28,9 @@ public:
 
 	using Super::Super;
 
+	/** Turns farmland into dirt.
+	Will first check for any colliding entities and teleport them to a higher position.
+	*/
 	static void TurnToDirt(cChunk & a_Chunk, Vector3i a_AbsPos)
 	{
 		auto RelPos = cChunkDef::AbsoluteToRelative(a_AbsPos);
@@ -61,8 +64,8 @@ public:
 				// The sketchy solution. It will almost always completely stop the player:
 				// Entity.TeleportToCoords(Entity.GetPosX(), Entity.GetPosY() + FullHeightDelta, Entity.GetPosZ());
 
-				// Another sketchy solution. It will rarely completely stop the player.
-				// It often causes weird pitch / yaw changes
+				// Another sketchy solution. It sometimes stops the player completely.
+				// It sometimes causes weird stutter
 				Entity.AddPosY(FullHeightDelta);
 				if (Entity.IsPlayer())
 				{
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -57,19 +57,16 @@ public:
 				{
 					return false;
 				}
-
-				// The solution which will make the player fall through the ground:
-				// Entity.SetPosition(Entity.GetPosition().addedY(FullHeightDelta));
-
-				// The sketchy solution. It will almost always completely stop the player:
-				// Entity.TeleportToCoords(Entity.GetPosX(), Entity.GetPosY() + FullHeightDelta, Entity.GetPosZ());
-
-				// Another sketchy solution. It sometimes stops the player completely.
-				// It sometimes causes weird stutter
 				Entity.AddPosY(FullHeightDelta);
+
+				// Players need a packet that will update their position
 				if (Entity.IsPlayer())
 				{
 					auto Player = static_cast<cPlayer *>(&Entity);
+					// This works, but it's much worse than Vanilla.
+					// This can be easily improved by implementing relative
+					// "Player Position And Look" packets! See
+					// https://wiki.vg/Protocol#Player_Position_And_Look_.28clientbound.29
 					Player->GetClientHandle()->SendPlayerMoveLook();
 				}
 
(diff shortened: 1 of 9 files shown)


DarkoGNU:
Also, looks like I might have to adjust my values a tiny bit, as reporting falling height lower by 0.125 (e.g. 0.625 for 6-layer snow, which "should" be 0.75) is specific to snow - for example, slabs report 0.500, not 0.375

DarkoGNU:
I tested the ravager exception from Minecraft Wiki. It turns out that ravagers don't trample farmlands when just stepping on them - they only destroy the plants, not the farmland. This makes my work a little easier, as I don't have to think about the code being adaptable to ravagers in the future - this should be an issue related only to ravagers, not trampling in general
Ravager

The exception about small mobs is true - they never trample (in Java Edition, at least). Also, even big animals, like horses or ravagers, trample only one farmland when they hit the ground, even though they appear to be much bigger - so it would be logical if they trampled multiple farmlands (but Minecraft isn't logical). This one particularly will make the code a lot easier than if I had to calculate e.g. how many farmlands a big mob should trample

Time for coding

Pokechu22:
> Also, even big animals, like horses or ravagers, trample only one farmland when they hit the ground, even though they appear to be much bigger

See https://bugs.mojang.com/browse/MC-1133 for further context.

DarkoGNU:
> > Also, even big animals, like horses or ravagers, trample only one farmland when they hit the ground, even though they appear to be much bigger
> 
> See https://bugs.mojang.com/browse/MC-1133 for further context.

Oh, so this behavior may be unintended? I think I should just go ahead and implement trampling just like it works in vanilla. Fixing LastGroundHeight, determining how many blocks below a mob should affect when hitting the ground... this is becoming a little much for someone who never before worked on a big project (me). Well, I always looked at the source code and thought to myself "I hardly understand anything, there's no way I'm contributing to this", but after trying with Cuberite, it seems like it isn't really that scary :joy:

DarkoGNU:
Well, I adjusted the thresholds, fixed the style (I hope), and made HandleFarmlandTrampling non-virtual - I think that it won't really have to be overridden, even when ravagers get implemented. I still need to write the actual code :satisfied:

Pokechu22:
> Oh, so this behavior may be unintended? I think I should just go ahead and implement trampling just like it works in vanilla anyway. 

I think it's more that vanilla went with the simplest approach to detecting if a block's being interacted with, and there are some downsides to that approach.  It's still fine to go with it, though.

The other thing to be careful about is making sure that players and other entities can't get stuck or fall through farmland when it becomes dirt (see [MC-104259](https://bugs.mojang.com/browse/MC-104259)); this matters because dirt has a larger hitbox than farmland (but cuberite might not implement that hitbox difference).  This includes one entity already standing on the farmland, and then another entity trampling it.

DarkoGNU:
> The other thing to be careful about is making sure that players and other entities can't get stuck or fall through farmland when it becomes dirt (see [MC-104259](https://bugs.mojang.com/browse/MC-104259)); this matters because dirt has a larger hitbox than farmland (but cuberite might not implement that hitbox difference). This includes one entity already standing on the farmland, and then another entity trampling it.

Cuberite implements that difference, I've just quickly tested this with my debug code for checking the falling height. So, thanks for pointing this out, I could introduce a bug otherwise

DarkoGNU:
And... well, I've discovered a bug in Cuberite! So, Cuberite implements reverting farmlands to dirt naturally (when no water source is nearby), but it looks like [MC-104259](https://bugs.mojang.com/browse/MC-104259) isn't prevented in any way, so I end up falling through dirt...

And I was so happy that I'll probably find code preventing [MC-104259](https://bugs.mojang.com/browse/MC-104259) in the code which implements turning dry farmlands to dirt. Well, more coding for me. And I probably should open an issue for this (unless there's one already)

DarkoGNU:
Update: I'm making progress... slowly. The fix for #5402 seems to work fine for players and mobs, but dropped items still fall through the farmland.

This can be fixed by increasing the teleportation height for them, but then it looks odd, especially that even teleporting them higher by 1/32 of a block (when compared to other entities) still doesn't fix the bug for them, so they might need something like 1/16 higher for this approach

DarkoGNU:
Ok, if someone more experienced is reading... I'm a little bit stuck. My issues:

- How to call cBlockFarmlandHandler::TurnToDirt from cPawn? Include BlockFarmland.h? Or is there a better solution? Also, how to obtain the chunk? GetParentChunk() returns a constant pointer, I need something non-constant
- How to call TurnToDirt from cBlockFarmlandHandler::OnNeighborChanged? Specifically - how to obtain the chunk... Basically, I'm having the same issue as above
- Issue with dropped items (cPickup) still falling through when farmland turns into dirt

Any help and advice welcome 😄

DarkoGNU:
Oh, I've just realized what I've done. Calling HandleFarmlandTrampling inside `if ((Damage > 0) && !FallDamageAbsorbed)` 🥲 

And that's why it was buggy

src/Blocks/BlockFarmland.h
```diff
@@ -76,7 +101,9 @@ class cBlockFarmlandHandler final :
 			}
 			default:
 			{
-				a_Chunk.SetBlock(a_RelPos, E_BLOCK_DIRT, 0);
+				auto World = a_Chunk.GetWorld();
+				auto AbsPos = a_Chunk.RelativeToAbsolute(a_RelPos);
```
bearbin, line 105: Since our bounding box is a single block large, there's no need to get the world, just pass in the chunk here and call ForEachEntityInBox on the chunk instead.

src/Blocks/BlockFarmland.h
```diff
@@ -101,9 +128,11 @@ class cBlockFarmlandHandler final :
 		}
 
 		// Check whether we should revert to dirt:
+		// TODO: fix for signs and slabs (possibly more blocks) - they should destroy farmland
 		auto upperBlock = a_ChunkInterface.GetBlock(a_BlockPos.addedY(1));
 		if (cBlockInfo::FullyOccupiesVoxel(upperBlock))
 		{
+			// TODO: somehow use TurnToDirt
```
bearbin, line 135: DoWithChunkAt on the interface?

src/Entities/Entity.h
```diff
@@ -42,6 +42,7 @@
 		} \
 	} while (false)
 
+#define VOLUME GetWidth() * GetHeight()
```
bearbin, line 45: We don't normally use macros like this anymore, the ones above are legacy. If this is used in loads of places it could be added as a private member function, but I don't think this needs abstraction at all honestly, just copy the calculation whenever it's used.

src/Entities/Pawn.h
```diff
@@ -33,6 +33,16 @@ class cPawn:
 	virtual void HandleFalling(void);
 	virtual void OnRemoveFromWorld(cWorld & a_World) override;
 
+	/** Handles farmland trampling when hitting the ground.
+	Algorithm:
+	fall height <= 0.6875 blocks: no trampling
+	fall height > 0.6875 and <= 1.0625: 25% chance of trampling
```
bearbin, line 39: Your implementation of this gives a different probability - 0.25 * 0.66 as the second else also applies when it gets through the first.

bearbin (changes requested):
Looking good generally, a few changes needed but you're on the right track :)

Always one step back when you go two forwards, as you've found out with the falling-through-the-ground bug. But it's fixed so good work.

bearbin:
The undocumented symbol error is to do with our plugin API documentation, which isn't colocated with the code, but instead in the `Server/Plugins/APIDump/Classes/World.lua` file (other classes are documented in their own files, or in the mega APIDesc.lua file where they're not split out).

DarkoGNU:
Thanks for your help

src/Blocks/BlockFarmland.h
DarkoGNU: Done

DarkoGNU: Note: right now, it's only for testing that it works correctly. It looks like the current implementation of `OnNeighborChanged` won't destroy farmland after placing any non-solid block (signs, anything else?), so there's no risk of any entity getting stuck or falling though farmland anyway

src/Blocks/BlockFarmland.h
DarkoGNU: Done

src/Blocks/BlockFarmland.h
```diff
@@ -25,6 +27,29 @@ class cBlockFarmlandHandler final :
...
+	static void TurnToDirt(cWorld & a_World, Vector3i a_AbsPos)
+	{
+		static const auto FarmlandHeight = cBlockInfo::GetBlockHeight(E_BLOCK_FARMLAND);
+		static const auto FullHeightDelta = 1 - FarmlandHeight;
+
+		a_World.ForEachEntityInBox(
+			cBoundingBox(Vector3d(0.5, FarmlandHeight, 0.5) + a_AbsPos, 0.5, FullHeightDelta),
+			[&](cEntity & Entity)
+			{
+				Entity.TeleportToCoords(Entity.GetPosX(), Entity.GetPosY() + FullHeightDelta, Entity.GetPosZ());
```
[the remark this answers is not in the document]
DarkoGNU: This is a little more complicated. I've left some alternative solutions in my recent commit - but none of the solutions I've tried is good, either the player falls through farmland or there's some visible stutter (and possibly they're even completely stopped) when they're moved up

src/Entities/Pawn.cpp
```diff
@@ -463,6 +466,14 @@ void cPawn::HandleFalling(void)
 
 		m_bTouchGround = true;
 		m_LastGroundHeight = GetPosY();
+
+		// Farmland trampling. Mobs smaller than 0.512 cubic blocks won't trample (Java Edition's behavior)
+		if (GetWorld()->IsFarmlandTramplingEnabled() &&
+			(BlockAtFoot == E_BLOCK_FARMLAND) &&
+			(GetWidth() * GetHeight() >= 0.512))
```
bearbin, line 473: Just realised - this doesn't calculate the volume at all! It's the area (is that what was intended)?

src/Entities/Pawn.h
bearbin: Yep much better now.

src/Entities/Pawn.cpp
DarkoGNU: Well... no. Another thing I'll need to fix

DarkoGNU: Fixed. I'm also wondering - shouldn't a new member variable for volume (m_Volume), or even specifically for trampling (m_bTramplesFarmland) be introduced? Or would it be just an useless micro-optimization?

I guess micro-optimization, because any difference would probably only be visible with huge farms, with mobs roaming freely over them... but it never hurts to wonder & ask

src/Entities/Pawn.cpp
```diff
@@ -468,9 +468,10 @@ void cPawn::HandleFalling(void)
 		m_LastGroundHeight = GetPosY();
 
 		// Farmland trampling. Mobs smaller than 0.512 cubic blocks won't trample (Java Edition's behavior)
+		// We only have width and height, so we have to calculate Width^2
 		if (GetWorld()->IsFarmlandTramplingEnabled() &&
 			(BlockAtFoot == E_BLOCK_FARMLAND) &&
-			(GetWidth() * GetHeight() >= 0.512))
+			(std::pow(GetWidth(), 2) * GetHeight() >= 0.512))
```
bearbin, line 474: pow is slower than just repeated multiplication, it's only necessary when you need to do variable-exponent powers or noninteger exponents.

src/Entities/Pawn.cpp
bearbin: I would suggest this would be a pointless micro-optimisation to store a derived value of height and width (and might even slow things down if memory usage is expensive)

src/Entities/Pawn.cpp
DarkoGNU: Well, that's fine. The performance would probably end up the same after compiler's optimizations though. Fixed

DarkoGNU:
Note: this still needs to be debugged (e.g. I still don't get any trampling when dropping a zombie on farms). I'm a little busy this weekend, so it may take some time

src/Entities/Pawn.cpp
bearbin: Yep, probably not too much different in reality. :+1:

bearbin:
Sounds good. Is this PR now ready, with the mobs issue seperate and ready to be fixed in another PR if somebody wants?

DarkoGNU:
Almost. Just let me investigate the mob issue a bit more, then I'll go through my own code and make sure it's all good

DarkoGNU:
Well, it looks like my PR will make #2015 worse for farmlands. Precisely - right now, mobs receive fall damage when falling on farmlands (this will be described in my comment on #2015). With my changes, mobs don't receive fall damage when falling on farmlands.

Why?

Because I added farmland's height to `cBlockInfo::GetBlockHeight`.

Yes, I've tested that. By making sure that `cBlockInfo::GetBlockHeight` returns the correct value, something has become more bugged 💩 

Crazy...

DarkoGNU:
Also, I think that I can make a workaround that will make farmland trampling work fine for mobs.

To make the workaround possible, I'd need to revert the farmland addition into `cBlockInfo::GetBlockHeight`.

Why? If I remember correctly, when mobs receive damage correctly (before adding the farmland to `GetBlockHeight`), the thing that still prevent trampling from working (with my implementation) is the part which determines whether a mob stands on a farmland.

I believe I'd just need to use the "block at foot" for players and the "block below" for mobs. I -believe-, because I'm in my bed right now and I'm not THAT determined to turn on my laptop and test that right now...

Is that even a good idea, though? Should a workaround like that be implemented?

DarkoGNU:
OK, the state of my pull request after the latest commit:

### Trampling

Working fine, but only for players

### Falling through the farm / getting stuck in it

Fixed - for players, mobs and dropped items

### Smoothness

Worse than Vanilla, but it can be improved - implementing relative `SendPlayerMoveLook` should help a lot!

About the smoothness - there's a quick and dirty implementation of relative `SendPlayerMoveLook` in my branch [RelativePlayerMovePacket](https://github.com/DarkoGNU/cuberite/tree/RelativePlayerMovePacket).

In my opinion, it works much better for farmland trampling than absolute packets, but it's not problem-free. There's some weird pitch/yaw bug, it looks even more serious than with absolute packets.

Video with absolute packets - [click here](https://www.youtube.com/watch?v=KOy2NLv4OpU). Look at the hoe, if it displays some weird movement, the player definitely feels a stutter, even if you can't see it on the video.

Video with relative packets - [click here](https://www.youtube.com/watch?v=csPhpsUtjNI). Every time the pitch/yaw bug occurs, it's clearly visible, but otherwise, the player's movement is butter smooth

Pokechu22:
Ah, I forgot about that pitch/yaw issue... That's a client bug; see [MC-104259](https://bugs.mojang.com/browse/MC-120545) (which still hasn't been fixed).  When a teleport packet occurs, the client re-wraps its angles to be in the -180-+180 range, but otherwise this doesn't happen.  So when the teleport occurs, it notices that the angle jumped by 360 degrees, and animates that.  Note that the angle displayed in F3 _is_ wrapped on display, obscuring the issue.

The easiest way to reproduce it is to spin around in place a few times, and then trigger the trampling.  You should also be able to reproduce it that way in vanilla.

DarkoGNU:
Oh, you're so helpful! Thanks!

DarkoGNU:
Ready for review, I guess. This can be merged if the stutter with absolute move packets is acceptable. If it's not, I can implement relative move packets in this PR or make a complementary pull request that will accomplish that task

bearbin (approved):
Looks good to me! Relative move would be nice to implement in due course, but a little bit of jank is fine for now.

PR #5403, opened 2022-04-08
Prettify

TheHyper45:
Reopening #5244
Additionally cEntity::BroadcastDeathMessage method now displays an appropriate death message when a tamed pet is killed by another mob (handles custom names).


bearbin (changes requested):
Looks like toupper breaks the windows build - might need an additional include there.

bearbin (approved):

PureTryOut:
Is there any reason you guys don't squash commits? Now the master branch is "infected" with commit messages like [Yet another style fix](https://github.com/cuberite/cuberite/commit/4bad8d859efa4119c0140e2d8f02cef01b9a44d3) which says literally nothing about what happened in that commit.

bearbin:
The real question is why we don't use merge commits. Generally my approach is not to destory history when possible.

PureTryOut:
Personally I don't like merge requests at all, I like my master branch to have a linear and clean timeline :sweat_smile: 
But anything better than this

bearbin:
Merge commits were disabled because xoft didn't like them. I forgot to squash this PR but there is the option to auto-squash - I just don't like it because it squashes the history you actually wanted (that's why we're using git, and not a bunch of zip files, after all!). It's best practice to manually squash before merge but a few extraneous commits in master are fine if you forget really IMO. 

In any case, arguing about whether to merge commit or not doesn't really help us make the software, so let's get back on with that :)

Issue #5404, opened 2022-04-12
Redstone on Y:1 Crashes the server


dxomg:
Client version: Server side issue but 1.12.2
Server OS: Windows but im aware that also work on every other build
Cuberite Commit id: b2e5f4d17bfe6fc8f5cb94d57f011f7705f05e41

### Expected behavior
When i make a redstone clock on Y:1 it would work

### Actual behavior
but when i make a redstone clock on Y:1 it crashes the server

### Steps to reproduce the behavior
Just make a redstone clock on Y:1 just above bedrock

### Server log
Nothing relevant in here

Relevant part of server log:

Err  [12:29:13] Deadlock detected: world world has been stuck at age 2141. Aborting the server.
     [12:29:13] CS at 0x3830768, World world chunkmap: RecursionCount = 1, ThreadIDHash = d5c1b19f855c6526
     [12:29:13] CS at 0x3f82090, World world tasks: RecursionCount = 0, ThreadIDHash = d5c1b19f855c6526
     [12:29:13] CS at 0x53a0768, World world_nether chunkmap: RecursionCount = 0, ThreadIDHash = eabe61e44a3abecc
     [12:29:13] CS at 0x5af2090, World world_nether tasks: RecursionCount = 0, ThreadIDHash = eabe61e44a3abecc
     [12:29:13] CS at 0x6310768, World world_the_end chunkmap: RecursionCount = 0, ThreadIDHash = 81ec6fbfd5439f58
     [12:29:13] CS at 0x6a62090, World world_the_end tasks: RecursionCount = 0, ThreadIDHash = 81ec6fbfd5439f58
     [12:29:13] CS at 0x43419f8, cLuaState plugin APIDump: RecursionCount = 0, ThreadIDHash = 4d25767f9dce13f5
     [12:29:13] CS at 0x4341488, cLuaState plugin ChatLog: RecursionCount = 0, ThreadIDHash = a42193f8f5519dd5
     [12:29:13] CS at 0x4341bc8, cLuaState plugin Core: RecursionCount = 0, ThreadIDHash = 2edbd3b8085443bb
     [12:29:13] CS at 0x4341d98, cLuaState plugin Debuggers: RecursionCount = 0, ThreadIDHash = 4d25767f9dce13f5
     [12:29:13] CS at 0x4341f68, cLuaState plugin DumpInfo: RecursionCount = 0, ThreadIDHash = 4d25767f9dce13f5
[...]

PR #5405, opened 2022-04-13
a fix that work for bedrock and redstone issue

tonitch:
A fix for #5404 that seems to work.

I don't know if there is any side effect.

Basically redstone update was using [RelativeAdjacents](https://github.com/cuberite/cuberite/blob/5ae924ec74bab199284c451552040bcc6f39bb3a/src/Simulator/IncrementalRedstoneSimulator/RedstoneDataHelper.h#L43) witch check for every adjacent block for the redstone dust

Changing it to use [RelativeLaterals](https://github.com/cuberite/cuberite/blob/5ae924ec74bab199284c451552040bcc6f39bb3a/src/Simulator/IncrementalRedstoneSimulator/RedstoneDataHelper.h#L55) doesn't bug anymore but it means that redstone doesn't check under and hover it!

the question is, is it a problem? I don't know the in depth behavior of redstone

### DIFF
--- a/src/Simulator/IncrementalRedstoneSimulator/RedstoneWireHandler.h
+++ b/src/Simulator/IncrementalRedstoneSimulator/RedstoneWireHandler.h
@@ -269,7 +269,7 @@ namespace RedstoneWireHandler
 		a_Chunk.SetMeta(a_Position, Power);
 
 		// Notify all positions, sans YP, to update:
-		for (const auto & Offset : RelativeLaterals) 
+		for (const auto & Offset : RelativeLaterals)
 		{
 			if (Offset == OffsetYP)
 			{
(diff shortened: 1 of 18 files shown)


tigerw:
investigating

tigerw:
some of the helper functions doing the update propagation can propagate to out of bounds Y, question is should the check go in the helper or workqueue processor, needs some thought

src/Simulator/IncrementalRedstoneSimulator/RedstoneWireHandler.h
```diff
@@ -269,7 +269,7 @@ namespace RedstoneWireHandler
 		a_Chunk.SetMeta(a_Position, Power);
 
 		// Notify all positions, sans YP, to update:
-		for (const auto & Offset : RelativeAdjacents)
+		for (const auto & Offset : RelativeLaterals)
```
tigerw, line 272: TODO: use this and add a separate UpdateAdjustedRelative for YM, the loop + if construct is not optimal.

tigerw:
Thanks @tonitch.

PR #5407, opened 2022-04-14
Changing the character from @ to $ for styled chat

tonitch:
Just a proposal for #5287

Can be ignored while the discussion take place I guess

### DIFF
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -107,7 +107,7 @@ void cCompositeChat::ParseText(const AString & a_ParseText)
 	{
 		switch (a_ParseText[i])
 		{
-			case '@':
+			case '$':
 			{
 				// Color code
 				i++;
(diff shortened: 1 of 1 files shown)


bearbin:
I don't think $ is any better than @, and probably worse than &. Either way, it doesn't make that much difference, I'd rather wait for a more comprehensive fix personally.

tonitch:
I'm pretty sure $ is used by bukkit and is kinda "standards" so I though it would make a good replacement ! Usualy in a sentence, $ is always with a space after it (I think) maybe we can check if there is a space after it to parse?

But I can understand that this might not be a good replacement either, maybe & is a even more standard way of doing it u think we have 2 things to check tho,

1 - if there is space or other char than those who would modify the style or
2 - if the & is next to a url 

The style shall not be applied 

We can also make it that the only way to stylish a text is to use the code in between spaces 

This would make a style of the form 

`&5 Hello guys, &a What's up &u today ?`

I don't think this would interfer with any other common or slightly uncommon chat messages

tonitch:
I'm gonna do another pr for asap for the modificaiton in #5287

PR #5409, opened 2022-04-19
Updated protocol functions to Vector3x

### LINKED ISSUE #5408: Check protocol methods for duplicate Vector3x and seperate parameter methods
For example:

https://github.com/cuberite/cuberite/blob/900f95c42b9b6ff333cc60640c17fb197025e280/src/Protocol/Protocol.h#L402
https://github.com/cuberite/cuberite/blob/900f95c42b9b6ff333cc60640c17fb197025e280/src/Protocol/Protocol.h#L403

12xx12:
Also added support in included functions

fixes #5408

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -2683,13 +2683,13 @@ void cClientHandle::SendExplosion(const Vector3f a_Position, const float a_Power
 	auto ParticleCount = std::min(static_cast<int>(ParticleFormula * 125), 600);
 
 	// Dark smoke particles:
-	SendParticleEffect("largesmoke", a_Position, {0.f, 0.f, 0.f,}, Spread, static_cast<int>(ParticleCount));
+	SendParticleEffect("largesmoke", a_Position, {0.f, 0.f, 0.f}, Spread, static_cast<int>(ParticleCount));
 
 	Spread = ParticleFormula * 0.35f;
 	ParticleCount = std::min(static_cast<int>(ParticleFormula * 550), 1800);
 
 	// Light smoke particles:
-	SendParticleEffect("explode", a_Position, {0.f, 0.f, 0.f,}, Spread, static_cast<int>(ParticleCount));
+	SendParticleEffect("explode", a_Position, {0.f, 0.f, 0.f}, Spread, static_cast<int>(ParticleCount));
 
 	// Shockwave effect:
 	m_Protocol->SendExplosion(a_Position, a_Power);
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -61,6 +61,8 @@ int cClientHandle::s_ClientCount = 0;
 
 float cClientHandle::FASTBREAK_PERCENTAGE;
 
+Vector3i cClientHandle::s_IllegalPosition = {0, cChunkDef::Height + 1, 0};
+
 
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -81,16 +83,16 @@ cClientHandle::cClientHandle(const AString & a_IPString, int a_ViewDistance) :
 	m_PingID(1),
 	m_BlockDigAnimStage(-1),
 	m_BlockDigAnimSpeed(0),
-	m_BlockDigAnimPos(),
+	m_BlockDigAnimPos(s_IllegalPosition),
 	m_HasStartedDigging(false),
-	m_LastDigBlockPos(),
+	m_LastDigBlockPos(s_IllegalPosition),
 	m_State(csConnected),
 	m_NumExplosionsThisTick(0),
 	m_NumBlockChangeInteractionsThisTick(0),
 	m_UniqueID(0),
 	m_HasSentPlayerChunk(false),
 	m_Locale("en_GB"),
-	m_LastPlacedSign(0, -1, 0),
+	m_LastPlacedSign(s_IllegalPosition),
 	m_ProtocolVersion(0)
 {
 	s_ClientCount++;  // Not protected by CS because clients are always constructed from the same thread
@@ -1119,16 +1121,13 @@ void cClientHandle::HandleLeftClick(Vector3i a_BlockPos, eBlockFace a_BlockFace,
 		Vector3i BlockPos = a_BlockPos;
 		AddFaceDirection(BlockPos, a_BlockFace);
 
-		if (cChunkDef::IsValidHeight(BlockPos.y) && cBlockInfo::IsClickedThrough(m_Player->GetWorld()->GetBlock(BlockPos)))
+		if (cChunkDef::IsValidHeight(BlockPos) && cBlockInfo::IsClickedThrough(m_Player->GetWorld()->GetBlock(BlockPos)))
 		{
 			a_BlockPos = BlockPos;
 		}
 
-		if (
-			(Diff(m_Player->GetPosX(), static_cast<double>(a_BlockPos.x)) > 6) ||
-			(Diff(m_Player->GetPosY(), static_cast<double>(a_BlockPos.y)) > 6) ||
-			(Diff(m_Player->GetPosZ(), static_cast<double>(a_BlockPos.z)) > 6)
-		)
+		constexpr double MaxBlockDistance = 6.0;
+		if (!cBoundingBox(m_Player->GetPosition(), MaxBlockDistance).IsInside(a_BlockPos))
 		{
 			m_Player->SendBlocksAround(a_BlockPos, 2);
 			return;
@@ -1176,11 +1175,11 @@ void cClientHandle::HandleLeftClick(Vector3i a_BlockPos, eBlockFace a_BlockFace,
 				// When bow is in off-hand / shield slot
 				if (m_Player->GetInventory().GetShieldSlot().m_ItemType == E_ITEM_BOW)
 				{
-					m_Player->GetInventory().GetShieldSlot().GetHandler().OnItemShoot(m_Player, {a_BlockPos}, a_BlockFace);
+					m_Player->GetInventory().GetShieldSlot().GetHandler().OnItemShoot(m_Player, a_BlockPos, a_BlockFace);
 				}
 				else
 				{
-					ItemHandler.OnItemShoot(m_Player, {a_BlockPos}, a_BlockFace);
+					ItemHandler.OnItemShoot(m_Player, a_BlockPos, a_BlockFace);
 				}
 			}
 			return;
@@ -1271,11 +1270,8 @@ void cClientHandle::HandleBlockDigStarted(Vector3i a_BlockPos, eBlockFace a_Bloc
 		return;
 	}
 
-	if (
-		(Diff(m_Player->GetPosX(), static_cast<double>(a_BlockPos.x)) > 6) ||
-		(Diff(m_Player->GetPosY(), static_cast<double>(a_BlockPos.y)) > 6) ||
-		(Diff(m_Player->GetPosZ(), static_cast<double>(a_BlockPos.z)) > 6)
-	)
+	constexpr double MaxBlockDistance = 6.0;
+	if (!cBoundingBox(m_Player->GetPosition(), MaxBlockDistance).IsInside(a_BlockPos))
 	{
 		m_Player->SendBlocksAround(a_BlockPos, 2);
 		return;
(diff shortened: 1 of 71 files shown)


src/ClientHandle.cpp
```diff
@@ -1120,34 +1116,30 @@ void cClientHandle::HandleLeftClick(int a_BlockX, int a_BlockY, int a_BlockZ, eB
...
 
-		int BlockX = a_BlockX;
-		int BlockY = a_BlockY;
-		int BlockZ = a_BlockZ;
-		AddFaceDirection(BlockX, BlockY, BlockZ, a_BlockFace);
+		Vector3i BlockPos = a_BlockPos;
+		AddFaceDirection(BlockPos, a_BlockFace);
 
-		if (cChunkDef::IsValidHeight(BlockY) && cBlockInfo::IsClickedThrough(m_Player->GetWorld()->GetBlock({ BlockX, BlockY, BlockZ })))
+		if (cChunkDef::IsValidHeight(BlockPos.y) && cBlockInfo::IsClickedThrough(m_Player->GetWorld()->GetBlock(BlockPos)))
```
bearbin, line 1122: Now you added the `ChunkDef::IsValidHeight(Vector3)` function, should we standardise on that form rather than the `Vec.y` form? I don't have any particular preference, and probably has negligible performance impact but it's slightly confusing with two overloaded signatures.

src/ClientHandle.cpp
```diff
@@ -1120,34 +1116,30 @@ void cClientHandle::HandleLeftClick(int a_BlockX, int a_BlockY, int a_BlockZ, eB
...
 		}
 
 		if (
-			(Diff(m_Player->GetPosX(), static_cast<double>(a_BlockX)) > 6) ||
-			(Diff(m_Player->GetPosY(), static_cast<double>(a_BlockY)) > 6) ||
-			(Diff(m_Player->GetPosZ(), static_cast<double>(a_BlockZ)) > 6)
+			(Diff(m_Player->GetPosX(), static_cast<double>(a_BlockPos.x)) > 6) ||
+			(Diff(m_Player->GetPosY(), static_cast<double>(a_BlockPos.y)) > 6) ||
+			(Diff(m_Player->GetPosZ(), static_cast<double>(a_BlockPos.z)) > 6)
 		)
```
bearbin, line 1131: Should this use one of our library functions rather than manual comparison to make things clearer (BoundingBox etc.)?

src/ClientHandle.cpp
```diff
@@ -1283,20 +1272,18 @@ void cClientHandle::HandleBlockDigStarted(int a_BlockX, int a_BlockY, int a_Bloc
 	}
 
 	if (
-		(Diff(m_Player->GetPosX(), static_cast<double>(a_BlockX)) > 6) ||
-		(Diff(m_Player->GetPosY(), static_cast<double>(a_BlockY)) > 6) ||
-		(Diff(m_Player->GetPosZ(), static_cast<double>(a_BlockZ)) > 6)
+		(Diff(m_Player->GetPosX(), static_cast<double>(a_BlockPos.x)) > 6) ||
+		(Diff(m_Player->GetPosY(), static_cast<double>(a_BlockPos.y)) > 6) ||
+		(Diff(m_Player->GetPosZ(), static_cast<double>(a_BlockPos.z)) > 6)
```
bearbin, line 1277: Same

src/ClientHandle.cpp
```diff
@@ -1316,35 +1303,31 @@ void cClientHandle::HandleBlockDigStarted(int a_BlockX, int a_BlockY, int a_Bloc
...
-	m_Player->GetWorld()->BroadcastBlockBreakAnimation(static_cast<UInt32>(m_UniqueID), {m_BlockDigAnimX, m_BlockDigAnimY, m_BlockDigAnimZ}, 0, this);
+	m_Player->GetWorld()->BroadcastBlockBreakAnimation(static_cast<UInt32>(m_UniqueID), m_BlockDigAnimPos, 0, this);
 
 	cWorld * World = m_Player->GetWorld();
 	cChunkInterface ChunkInterface(World->GetChunkMap());
-	cBlockHandler::For(DiggingBlock).OnDigging(ChunkInterface, *World, *m_Player, {a_BlockX, a_BlockY, a_BlockZ});
+	cBlockHandler::For(DiggingBlock).OnDigging(ChunkInterface, *World, *m_Player, {a_BlockPos});
 
-	m_Player->GetEquippedItem().GetHandler().OnDiggingBlock(World, m_Player, m_Player->GetEquippedItem(), {a_BlockX, a_BlockY, a_BlockZ}, a_BlockFace);
+	m_Player->GetEquippedItem().GetHandler().OnDiggingBlock(World, m_Player, m_Player->GetEquippedItem(), {a_BlockPos}, a_BlockFace);
```
bearbin, line 1314: Why are these `a_BlockPos`s in curly brackets?

src/ClientHandle.cpp
```diff
@@ -1354,7 +1337,7 @@ void cClientHandle::HandleBlockDigFinished(int a_BlockX, int a_BlockY, int a_Blo
 
 	BLOCKTYPE DugBlock;
 	NIBBLETYPE DugMeta;
-	m_Player->GetWorld()->GetBlockTypeMeta({ a_BlockX, a_BlockY, a_BlockZ }, DugBlock, DugMeta);
+	m_Player->GetWorld()->GetBlockTypeMeta({ a_BlockPos }, DugBlock, DugMeta);
```
bearbin, line 1340: Again, we don't need to be constructing a new vector here.

src/ClientHandle.cpp
```diff
@@ -1453,7 +1434,7 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 	Actions rejected by plugin will not lead to other attempts.
 	E.g., when opening a chest with a dirt in hand, if the plugin rejects opening the chest, the dirt will not be placed. */
 
-	if ((a_BlockFace == BLOCK_FACE_NONE) || !cChunkDef::IsValidHeight(a_BlockY))
+	if ((a_BlockFace == BLOCK_FACE_NONE) || !cChunkDef::IsValidHeight(a_BlockPos))
```
bearbin, line 1437: Another position - standardise?

src/ClientHandle.cpp
```diff
@@ -1462,7 +1443,7 @@ void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, e
 	// TODO: We are still consuming the items in main hand. Remove this override when the off-hand consumption is handled correctly.
 	a_UsedMainHand = true;
 
-	const Vector3i ClickedPosition(a_BlockX, a_BlockY, a_BlockZ);
+	const Vector3i ClickedPosition(a_BlockPos);
```
bearbin, line 1446: I don't think this vector serves any purpose now that the arguments are a vector3.

src/ClientHandle.cpp
```diff
@@ -1425,19 +1408,17 @@ void cClientHandle::FinishDigAnimation()
...
-	m_BlockDigAnimZ = -1;
+	m_LastDigBlockPos = {-1, -1, -1};
 }
 
 
 
 
 
-void cClientHandle::HandleRightClick(int a_BlockX, int a_BlockY, int a_BlockZ, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ, bool a_UsedMainHand)
+void cClientHandle::HandleRightClick(Vector3i a_BlockPos, eBlockFace a_BlockFace, int a_CursorX, int a_CursorY, int a_CursorZ, bool a_UsedMainHand)
```
bearbin, line 1421: Should CursorXXX not also be a vector?

src/ClientHandle.cpp
```diff
@@ -1782,15 +1763,15 @@ void cClientHandle::HandleWindowClick(UInt8 a_WindowID, Int16 a_SlotNum, eClickA
...
 
 void cClientHandle::HandleUpdateSign(
-	int a_BlockX, int a_BlockY, int a_BlockZ,
+	Vector3i a_BlockPos,
 	const AString & a_Line1, const AString & a_Line2,
 	const AString & a_Line3, const AString & a_Line4
 )
 {
-	if (m_LastPlacedSign.Equals(Vector3i(a_BlockX, a_BlockY, a_BlockZ)))
+	if (m_LastPlacedSign.Equals(Vector3i(a_BlockPos)))
```
bearbin, line 1771: No need for the reconstruction here.

src/ClientHandle.cpp
```diff
@@ -1884,7 +1865,7 @@ void cClientHandle::HandleUseItem(bool a_UsedMainHand)
 
 	LOGD("HandleUseItem: Hand: %d; HeldItem: %s", a_UsedMainHand, ItemToFullString(HeldItem).c_str());
 
-	if (PlgMgr->CallHookPlayerRightClick(*m_Player, -1, 255, -1, BLOCK_FACE_NONE, 0, 0, 0))
+	if (PlgMgr->CallHookPlayerRightClick(*m_Player, {-1, 255, -1}, BLOCK_FACE_NONE, {0, 0, 0}))
```
bearbin, line 1868: Can we put the DefaultBlockPos and DefaultCursorPos into a variable so they're not so much magic numbers.

src/ClientHandle.cpp
```diff
@@ -2708,13 +2687,13 @@ void cClientHandle::SendExplosion(const Vector3f a_Position, const float a_Power
...
 	// Dark smoke particles:
-	SendParticleEffect("largesmoke", a_Position.x, a_Position.y, a_Position.z, 0, 0, 0, Spread, static_cast<int>(ParticleCount));
+	SendParticleEffect("largesmoke", a_Position, Vector3f(), Spread, static_cast<int>(ParticleCount));
 
 	Spread = ParticleFormula * 0.35f;
 	ParticleCount = std::min(static_cast<int>(ParticleFormula * 550), 1800);
 
 	// Light smoke particles:
-	SendParticleEffect("explode", a_Position.x, a_Position.y, a_Position.z, 0, 0, 0, Spread, static_cast<int>(ParticleCount));
+	SendParticleEffect("explode", a_Position, Vector3f(), Spread, static_cast<int>(ParticleCount));
```
bearbin, line 2696: I slightly prefer the `{0., 0., 0.}` list-initialisation form here for clarity.

src/World.h
```diff
@@ -311,10 +311,19 @@ class cWorld  // tolua_export
 	void ChunkLoadFailed(int a_ChunkX, int a_ChunkZ);
 
 	/** Sets the sign text, asking plugins for permission first. a_Player is the player who this change belongs to, may be nullptr. Returns true if sign text changed. */
-	bool SetSignLines(int a_BlockX, int a_BlockY, int a_BlockZ, const AString & a_Line1, const AString & a_Line2, const AString & a_Line3, const AString & a_Line4, cPlayer * a_Player = nullptr);  // Exported in ManualBindings.cpp
+	bool SetSignLines(Vector3i a_BlockPos, const AString & a_Line1, const AString & a_Line2, const AString & a_Line3, const AString & a_Line4, cPlayer * a_Player = nullptr);
+	bool SetSignLines(int a_BlockX, int a_BlockY, int a_BlockZ, const AString & a_Line1, const AString & a_Line2, const AString & a_Line3, const AString & a_Line4, cPlayer * a_Player = nullptr)  // Exported in ManualBindings.cpp
+	{
+		return SetSignLines({a_BlockX, a_BlockY, a_BlockZ}, a_Line1, a_Line2, a_Line3, a_Line4, a_Player);
+	}
```
bearbin, line 318: This is already manually bound so the legacy signature can be removed.

bearbin (changes requested):

src/ClientHandle.cpp
12xx12: Replace With xD

src/ClientHandle.cpp
12xx12: Found some more methods

src/Mobs/PathFinder.cpp
```diff
@@ -185,74 +185,78 @@ bool cPathFinder::EnsureProperPoint(Vector3d & a_Vector, cChunk & a_Chunk)
...
+				break;
 			}
 		}
-		breakBothLoops:
 
 		// Go down to the lowest air block.
 		if (InTheAir)
 		{
-			while (a_Vector.y > 0)
+			while (cChunkDef::IsValidHeight(a_Vector.addedY(-1)))
```
bearbin, line 239: This is an infinite loop.

bearbin (changes requested):
I think the change to IsValidHeight is in general a good idea, looking at it now it's been implemented. The only problem is when you have to construct a new vector just to call the function. But there are only a few of those so maybe they can just be ignored.

Only remaining thing is the broken loop logic.

src/Mobs/PathFinder.cpp
12xx12: Whoops - thank you. That fell through when refactoring that weird code above

12xx12:
And yeah. This is a bit weird. Some of those calls could be removed by changing the whole function. There are four or five calls left over

bearbin (approved):

PR #5411, opened 2022-04-21
Protocol sources fixes

Persson-dev:
As mentioned in #5388, i moved the HANDLE_READ and HANDLE_PACKET_READ macros in Packetizer.h.
I also fixed some compiler warnings in the 1.14 protocol source file.

### DIFF
--- a/src/Protocol/Protocol_1_14.cpp
+++ b/src/Protocol/Protocol_1_14.cpp
@@ -719,7 +719,6 @@ UInt8 cProtocol_1_14::GetProtocolEntityType(const cEntity & a_Entity) const
 				case PType::pkSplashPotion: return 81;
 				case PType::pkFirework: return 26;
 				case PType::pkWitherSkull: return 92;
-				default: break;
 			}
 			break;
 		}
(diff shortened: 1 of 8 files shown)


12xx12 (approved):

src/Protocol/Protocol_1_14.cpp
```diff
@@ -732,7 +719,9 @@ UInt8 cProtocol_1_14::GetProtocolEntityType(const cEntity & a_Entity) const
 				case PType::pkSplashPotion: return 81;
 				case PType::pkFirework: return 26;
 				case PType::pkWitherSkull: return 92;
+				default: break;
```
bearbin, line 722: Not sure about this change, this switch should be an all-cases-handled one so we get a warning when updating the PTypes. (The break on the line below is fine though).

12xx12: Oh yeah - I remember. That's why there are that many useless case statements in the monster section.

src/Protocol/Protocol_1_14.cpp
```diff
@@ -733,6 +720,7 @@ UInt8 cProtocol_1_14::GetProtocolEntityType(const cEntity & a_Entity) const
 				case PType::pkFirework: return 26;
 				case PType::pkWitherSkull: return 92;
 			}
+			break;
```
12xx12, line 723: Shouldn't the ```break;``` statement be inside the ```{ }```

12xx12: Yeah - My bad. The GitHub web interface made it seem like the break was somewhere else.

PR #5412, opened 2022-04-21
Add support for 1.14.x protocols

Persson-dev:
I tested joining with 1.14.x clients and it seems to work.

### DIFF
--- a/src/Protocol/Protocol.h
+++ b/src/Protocol/Protocol.h
@@ -346,7 +346,11 @@ public:
 		v1_13   = 393,
 		v1_13_1 = 401,
 		v1_13_2 = 404,  // TODO: this constant should be in WebServer
-		v1_14   = 477
+		v1_14   = 477,
+		v1_14_1 = 480,
+		v1_14_2 = 485,
+		v1_14_3 = 490,
+		v1_14_4 = 498
 	};
 
 	enum State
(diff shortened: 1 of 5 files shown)


src/Protocol/Protocol.h
```diff
@@ -346,7 +346,11 @@ class cProtocol
 		v1_13   = 393,
 		v1_13_1 = 401,
 		v1_13_2 = 404,  // TODO: this constant should be in WebServer
```
12xx12, line 348: Maybe move this comment somewhere usefull (The top maybe)

12xx12 (commented):
I'll have a look later and compile this. Since I am experiencing problems, you aren't

12xx12 (commented):
Thanks!

Aren't there any packets changed (usually there is one or so)

Persson-dev:
It might be the case but the wiki (https://minecraft.fandom.com/fr/wiki/1.14.x) says that these are minor releases and the wiki protocol (https://wiki.vg) is not very clear.

bearbin (approved):
If we discover any packet differences, then we can fix it with this boilerplate. Looks fine.

12xx12:
I just wanted to say that it also works fine.

src/Protocol/Protocol.h
```diff
@@ -330,6 +330,7 @@ class cProtocol
 		Pose
 	};
 
+	// TODO: these constants should be in WebServer
```
tigerw, line 333: hah this was an HTTP 404 joke, it doesn't make sense now xD

PR #5413, opened 2022-04-23
Implement relative SendPlayerMoveLook. Use it in TurnToDirt

DarkoGNU:
A quick improvement for #5401 - it makes farmland trampling much smoother.

I tried to avoid duplicating code, I hope I've done this the right way - else, I'll make any requested fixes.

I've tested this on 1.12.2. Farmland trampling and teleporting works.

On 1.8.9, teleporting works, but... farmland trampling doesn't. That's caused by [MC-85162](https://bugs.mojang.com/browse/MC-85162) - #5401 relies on farmland being only 15 pixels high, but that was implemented in Minecraft 1.10.1. Indeed, after switching my client to 1.10.1, trampling works fine.

Another thing I'll need to fix, unless not fully supporting older clients is OK

### DIFF
--- a/src/Blocks/BlockFarmland.h
+++ b/src/Blocks/BlockFarmland.h
@@ -44,7 +44,7 @@ public:
 	/** Turns farmland into dirt.
 	Will first check for any colliding entities and teleport them to a higher position.
 	*/
-	static void TurnToDirt(cChunk & a_Chunk, Vector3i a_AbsPos, Vector3i a_RelPos)
+	static void TurnToDirt(cChunk & a_Chunk, const Vector3i a_AbsPos, const Vector3i a_RelPos)
 	{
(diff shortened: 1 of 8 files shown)


DarkoGNU:
I'm not sure how to approach fixing `TurnToDirt` and fixing trampling for older clients (and probably mobs too) in another pull request, as the required code changes may somewhat interfere with what's done here. I'll commit this here for now

DarkoGNU:
I think I'm ready, reviewers are welcome

bearbin (approved):
This looks good, thanks!

Not fully supporting older clients is fine, client support is on a best effort basis with no guarantee of every feature working on any particular version. Of course ideally if there's a workaround then we can apply it, but it's not essential!

PR #5414, opened 2022-04-23
Fix trampling for older clients and mobs

DarkoGNU:
This PR kind of relies on merging #5413. Without merging this, older clients and mobs will be moved up a little bit to "prevent" falling through the farmland, since they'll somehow still intersect with the farmland.

Yeah, this is the workaround I contemplated in my comments on #5401. Because it turns that not only mobs couldn't trample, but older clients (affected by [MC-85162](https://bugs.mojang.com/browse/MC-85162)) too, I decided to implement that workaround.

I've decided to refactor `cPawn::HandleFalling` a tiny bit, so `cPawn::HandleFarmlandTrampling` doesn't have to compute things and perform checks which have to be done in `cPawn::HandleFalling` anyway. It makes `HandleFarmlandTrampling` shorter and simpler. If my refactoring isn't a good idea, I can revert it and just perform these checks & compute stuff in `HandleFarmlandTrampling`.

I've tested this PR. Now farmlands are trampled by mobs, older clients and newer clients! Particle effects work just like before. Also, trampling a farmland or any jumping on a position like Y:1 doesn't crash the server. Also important to check that, I guess

### DIFF
--- a/src/BlockInfo.cpp
+++ b/src/BlockInfo.cpp
@@ -1164,7 +1164,7 @@ float cBlockInfo::GetBlockHeight(const BLOCKTYPE Block)
 		case E_BLOCK_DARK_OAK_FENCE:      return 1.5;
 		case E_BLOCK_DARK_OAK_FENCE_GATE: return 1.5;
 		case E_BLOCK_ENCHANTMENT_TABLE:   return 0.75;    // 12 pixels
-		case E_BLOCK_FARMLAND:            return 0.9375;  // 15 pixels
+		// case E_BLOCK_FARMLAND:         return 0.9375;  // prevents trampling for mobs (#2015) and older clients (MC-85162)
 		case E_BLOCK_FENCE:               return 1.5;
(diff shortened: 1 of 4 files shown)


DarkoGNU:
By the way, it would be nice if the commit hook checked only staged changes. I'm fixing my style and forgetting to commit all the time.

But I understand if that would be hard to implement

DarkoGNU:
I think I'm ready, reviewers are welcome.

Nice find with that head (look at my latest commit)

DarkoGNU:
After some investigation, I think that these two blocks (heads and pots) are the only two which we really need to check in order to determine whether we should proceed with trampling.

That's still not a pretty solution. We probably should determine the block that the player is **really** standing on. But that won't work for farmlands in case of:

- mobs - until we fix the physics engine, I guess
- older clients ([MC-85162](https://bugs.mojang.com/browse/MC-85162)) - I'm not even sure whether this can be fixed on the server side

12xx12 (approved):

PR #5415, opened 2022-04-24
Make snow golems throw snowballs at hostile mobs. Add monster infighting

### LINKED ISSUE #3453: Snow golems don't attack mobs like they should
Taken from #1109

DarkoGNU:
Just a draft right now. I don't want to be the farmland guy, so I've decided to try something different.

Fixes #3453

### DIFF
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -36,9 +36,9 @@ void cAggressiveMonster::InStateChasing(std::chrono::milliseconds a_Dt, cChunk &
 
 
 
-void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
+void cAggressiveMonster::EventSeePlayer(cPawn * a_Pawn, cChunk & a_Chunk)
 {
-	Super::EventSeePlayer(a_Player, a_Chunk);
+	Super::EventSeePlayer(a_Pawn, a_Chunk);
 	m_EMState = CHASING;
 }
 
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -45,7 +45,7 @@ bool cAggressiveMonster::DoTakeDamage(TakeDamageInfo & a_TDI)
 
 	if (GetTarget() != nullptr)
 	{
-		// The suitability of the target should already be checked every tick
+		// The suitability of the target should already be checked
 		m_EMState = CHASING;
 	}
 
@@ -142,7 +142,7 @@ void cAggressiveMonster::CheckEventSeeMob(cChunk & a_Chunk)
 	// Enumerate all monsters within sight:
 	m_World->ForEachEntityInBox(
 		cBoundingBox(MyHeadPosition, m_SightDistance * 2),
-		[&](cEntity & a_Entity)
+		[this, &TargetMonster, &ClosestDistance, MyHeadPosition, Tracer](cEntity & a_Entity)
 	{
 		if (!a_Entity.IsMob())
 		{
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -36,6 +36,26 @@ void cAggressiveMonster::InStateChasing(std::chrono::milliseconds a_Dt, cChunk &
 
 
 
+bool cAggressiveMonster::DoTakeDamage(TakeDamageInfo & a_TDI)
+{
+	if (!Super::DoTakeDamage(a_TDI))
+	{
+		return false;
+	}
+
+	if (GetTarget() != nullptr)
+	{
+		// The suitability of the target should already be checked every tick
+		m_EMState = CHASING;
+	}
+
+	return true;
+}
+
+
+
+
+
 void cAggressiveMonster::EventSeePlayer(cPawn * a_Pawn, cChunk & a_Chunk)
 {
 	Super::EventSeePlayer(a_Pawn, a_Chunk);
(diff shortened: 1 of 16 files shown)


DarkoGNU:
OK, throwing snowballs done. I've just copied the code from skeletons and adjusted it, like tigerw suggested. I don't know why, but I've got some failed assertions about unsaved projectile entities when testing this. It happened all the time, then... it stopped and I can no longer reproduce this

DarkoGNU:
What a shame that snow golem's behavior (attack hostile mobs) isn't implemented for the iron golem. Less code to copy-paste for me 😢 

I guess that's the next thing I'll do, after fixing snow golems

DarkoGNU:
Notes about the [Minecraft Wiki](https://minecraft.fandom.com/wiki/Snow_Golem) page:

- it looks like snow golems never attacks ghasts
- they'll attack iron golems only by accident (then, the iron golem will kill the poor snowman)
- they'll attack zombie pigmen, which are now called _zombified piglins_

DarkoGNU:
OK, this is working fairly well right now, I just need to do some debugging...
I get failed `ASSERT(IsTicking())` in `cPawn::NoLongerTargetingMe` when I kill the mob targeted by snow golems

DarkoGNU:
Curiously, I'm getting that failed assertion only when at least two snow golems are attacking one monster (and the monster is killed)

DarkoGNU:
Nice, now `ASSERT(Monster->GetTarget() == this)` fails in `cPawn::StopEveryoneFromTargetingMe`

DarkoGNU:
Yay, it seems to work correctly and I'm personally... happy enough with how it's done.

I've made some refactorings to avoid duplicating code and prepare for implementing iron golem's behavior (very similar to snow golems).

Also, `cAggressiveMonster::CheckEventSeeMob`, for example, can be used to make hostile mobs attack iron golems, wolfs - attack animals, etc. - basically any case when a mob is hostile towards another mob.

To sum it up - I think that my changes are a step in the right direction to improving the mob AI.

I'll test my changes before marking this as ready for review. Also, I'll read the diffs once again and make sure there's nothing stupid there (to the best of my ability)

DarkoGNU:
OK, this works fairly well and I'm marking this as ready for review. I've tested this reasonably well - I mean, I've played for several minutes, spawned many snow golems, aggressive monsters, passive and neutral monsters... I haven't seen any new bug.

A note about monster infighting: it kinda worked previously... just the monster's state wasn't being set to `CHASING`, so the monster wouldn't chase its target, which usually meant just wandering off...

The error that CI is throwing is `error: out-of-line definition of 'EventSeePlayer' does not match any declaration in 'cAggressiveMonster'`. I don't know how to fix it - I refactored the code, so that `cAggressiveMonster` now overrides `EventSeeTarget` instead of `EventSeePlayer`.

And I can't just fix `tests/Generating/Stubs.cpp`, because when I make even the tiniest change there, the pre-commit hook finds like 50 violations

src/Mobs/UtilityMonster.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "../World.h"
+
+
+
+
+
+cUtilityMonster::cUtilityMonster(const AString & a_ConfigName, eMonsterType a_MobType, const AString & a_SoundHurt, const AString & a_SoundDeath, const AString & a_SoundAmbient, float a_Width, float a_Height) :
+	Super(a_ConfigName, a_MobType, a_SoundHurt, a_SoundDeath, a_SoundAmbient, a_Width, a_Height)
+{
+	m_EMPersonality = PASSIVE;
```
tonitch, line 15: that is a detail but, isn't it redundant ? ^^

DarkoGNU: Maybe it is, I've basically copied this stuff from `PassiveAggressiveMonster.h'. I'll check

DarkoGNU: I'm not even sure whether making `cUtilityMonster` will be the best course of action. It ensures that `CheckEventSeeMob` or `CheckEventLostMob` is called every tick, but it looks like most hostile mobs will also have to perform these check anyway - they're gonna be hostile towards iron golems.

Unless this is solved in other way, e.g. by making the iron golem set the target of hostile mobs nearby to itself. But that would ignore their range of sight, so... not the best option, I guess

tonitch (changes requested):
I don't think my review have real value there but I like to inspect the code for better understanding and then say it if I notice something! Hope this is not disturbing and just tell me if I should stop ^^

DarkoGNU:
Maybe I should take the refactoring deeper? Get rid of `EventSeePlayer`, `EventSeeMob`, and similar distinctions between players and monsters - completely replace them with stuff like `EventSeeTarget`?

Refactor `CheckEventSeePlayer` and `CheckEventSeeMob` to be one function - `CheckEventSeeTarget`, that would possibly call a specialization for players if a mob is only hostile towards players (I guess it's more efficient), and use a mechanism that works for both mobs and players if a mob is hostile towards at least one type of monsters...

And so on... the code right now is written like monsters are only hostile towards players, but modern Minecraft is different - monsters can be hostile towards iron golems, villagers, animals (wolves), and hostile mobs (snow golems)

src/Mobs/AggressiveMonster.cpp
```diff
@@ -101,3 +121,59 @@ bool cAggressiveMonster::Attack(std::chrono::milliseconds a_Dt)
...
+	const auto Tracer = IsNetherNative() ? cLineBlockTracer::losAirWaterLava : cLineBlockTracer::losAirWater;
+
+	// Enumerate all monsters within sight:
+	m_World->ForEachEntityInBox(
+		cBoundingBox(MyHeadPosition, m_SightDistance * 2),
+		[this, &TargetMonster, &ClosestDistance, MyHeadPosition, Tracer](cEntity & a_Entity)
+	{
+		if (!a_Entity.IsMob())
+		{
+			return false;
```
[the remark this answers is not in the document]
DarkoGNU: This is not a void method, this is a lambda. It needs to return false, if it returns true, it won't iterate over other mobs. That's how `ForEachEntityInBox` works

DarkoGNU:
> I don't think my review have real value

Your reviews are welcome. It's not like I'm a C++ guru. 5 months ago I didn't know what a reference or pointer is, if you want to see, go check out [this](https://github.com/DarkoGNU/DarkoASM/blob/main/DarkoASM/SpecialMap.cpp).

Also, that's my favorite implementation of bidirectional map. It uses two maps, one of them has keys that are literally just natural numbers, starting from zero and increasing by one with every entry. If you don't see what's wrong - a vector would suffice ^^

I've also made that mistake like two weeks ago when helping another student with his assignment. That time, it was less obvious, because the bidirectional maps didn't have natural numbers as keys, but enum entries - which are also natural numbers, starting from zero and increasing by one with each entry... I didn't even bother with fixing that, not worth the effort 💩

DarkoGNU:
Also, I think I'm going to close this pull requests and open two other instead. One - just implementing the attack for snow golems. The second one - refactoring a lot of code for monsters, so that the code is adjusted to the fact that monsters can be hostile not only towards players, but also other monsters

src/Mobs/AggressiveMonster.cpp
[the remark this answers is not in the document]
tonitch: ho, I didn't noticed, I'm not used to lambda in cpp x)

PR #5416, opened 2022-04-28
rework of the color code with & and standard codes

tonitch:
This is a draft for the rework of the chat, as explain in  #5287 (no response so I assume this should be fine)

I want to make it as close as common server as possible so I want to follow this image color code and so using the `&` symbol

### DIFF
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -387,8 +387,14 @@ void cCompositeChat::AddChatPartStyle(Json::Value & a_Value, const AString & a_P
 		{
 			case 'k': a_Value["obfuscated"] = Json::Value(true); break;
 			case 'l': a_Value["bold"] = Json::Value(true); break;
+			case 's':  // Deprecated
+				LOG("Value s in AddChatPartStyle() is deprecated");
 			case 'm': a_Value["strikethrough"] = Json::Value(true); break;
+			case 'u':  // Deprecated
+				LOG("Value u in AddChatPartStyle() is deprecated");
 			case 'n': a_Value["underlined"] = Json::Value(true); break;
+			case 'i':  // Deprecated
+				LOG("Value i in AddChatPartStyle() is deprecated");
 			case 'o': a_Value["italic"] = Json::Value(true); break;
 			case '0': a_Value["color"] = Json::Value("black"); break;
 			case '1': a_Value["color"] = Json::Value("dark_blue"); break;
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -95,7 +95,6 @@ void cCompositeChat::AddShowAchievementPart(const AString & a_PlayerName, const 
 
 
 
-//TODO tonitch
 
 /**
  * Parse the input message to add colors or link then add it to the object.
@@ -118,7 +117,7 @@ void cCompositeChat::ParseText(const AString & a_ParseText)
 		{
 			case '&': //< Color code
 			{
-				if(a_ParseText[i-1] == '\\')
+				if(i != 0 && a_ParseText[i-1] == '\\')
 				{
 					CurrentText.append(a_ParseText, cursor, i-1-cursor).append("&");
 					AddTextPart(CurrentText, CurrentStyle);
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -119,7 +119,7 @@ void cCompositeChat::ParseText(const AString & a_ParseText)
 			{
 				if ((i != 0) && (a_ParseText[i-1] == '\\'))
 				{
-					CurrentText.append(a_ParseText, cursor, i-1-cursor).append("&");
+					CurrentText.append(a_ParseText, cursor, i - cursor - 1).append("&");
 					AddTextPart(CurrentText, CurrentStyle);
 					CurrentText.clear();
 					cursor = ++i;
@@ -128,7 +128,7 @@ void cCompositeChat::ParseText(const AString & a_ParseText)
 
 				if (cursor < i)
 				{
-					CurrentText.append(a_ParseText, cursor, i-cursor);
+					CurrentText.append(a_ParseText, cursor, i - cursor);
 					AddTextPart(CurrentText, CurrentStyle);
 					CurrentText.clear();
 				}
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -391,7 +391,7 @@ void cCompositeChat::AddChatPartStyle(Json::Value & a_Value, const AString & a_P
 				LOGERROR("Value s in AddChatPartStyle() is deprecated");
 			case 'm': a_Value["strikethrough"] = Json::Value(true); break;
 			case 'u':  // Deprecated
-				LOG("Value u in AddChatPartStyle() is deprecated");
+				LOGERROR("Value u in AddChatPartStyle() is deprecated");
 			case 'n': a_Value["underlined"] = Json::Value(true); break;
 			case 'i':  // Deprecated
 				LOG("Value i in AddChatPartStyle() is deprecated");
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -388,7 +388,7 @@ void cCompositeChat::AddChatPartStyle(Json::Value & a_Value, const AString & a_P
 			case 'k': a_Value["obfuscated"] = Json::Value(true); break;
 			case 'l': a_Value["bold"] = Json::Value(true); break;
 			case 's':  // Deprecated
-				LOG("Value s in AddChatPartStyle() is deprecated");
+				LOGERROR("Value s in AddChatPartStyle() is deprecated");
 			case 'm': a_Value["strikethrough"] = Json::Value(true); break;
 			case 'u':  // Deprecated
 				LOG("Value u in AddChatPartStyle() is deprecated");
--- a/src/CompositeChat.cpp
+++ b/src/CompositeChat.cpp
@@ -97,13 +97,13 @@ void cCompositeChat::AddShowAchievementPart(const AString & a_PlayerName, const 
 
 
 /**
- * Parse the input message to add colors or link then add it to the object.
- *
- * It detect every & of the message and the next character for it to colorize.
- * It detect : in the text to detect link structures.
- *
- * @param a_ParseText The input text to parse
- */
+* Parse the input message to add colors or link then add it to the object.
+*
+* It detect every & of the message and the next character for it to colorize.
+* It detect : in the text to detect link structures.
+*
+* @param a_ParseText The input text to parse
+*/
 void cCompositeChat::ParseText(const AString & a_ParseText)
 {
 	size_t len = a_ParseText.length();
@@ -115,9 +115,9 @@ void cCompositeChat::ParseText(const AString & a_ParseText)
 	{
 		switch (a_ParseText[i])
 		{
-			case '&': //< Color code
+			case '&':  //< Color code
 			{
-				if(i != 0 && a_ParseText[i-1] == '\\')
+				if ((i != 0) && (a_ParseText[i-1] == '\\'))
 				{
 					CurrentText.append(a_ParseText, cursor, i-1-cursor).append("&");
 					AddTextPart(CurrentText, CurrentStyle);
@@ -132,7 +132,7 @@ void cCompositeChat::ParseText(const AString & a_ParseText)
 				i++;
 				cursor = i + 1;
 
-				if(a_ParseText[i] == 'r')
+				if (a_ParseText[i] == 'r')
 				{
 					CurrentStyle = "";
 				}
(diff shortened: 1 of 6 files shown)


tonitch:
I think I'm ready for review,
I hope that you will understand my changes. I think it is quite important that we follow standard of minecraft, even community's one, so people are not disturbed by it

I tried to replace every bit that uses style in the code with this new guideline there is only one part I'm not sure...
https://github.com/cuberite/cuberite/blob/3c75377ce679094dfe55ee9aa872dabffd2ba3f0/src/ClientHandle.cpp#L1563-L1568
I'm not sure if ranks color are implemented and so how should I change that... I haven't seens any way of setting the color of a rank....

tonitch:
Is there any update on the review ?

src/CompositeChat.cpp
```diff
@@ -96,50 +96,52 @@ void cCompositeChat::AddShowAchievementPart(const AString & a_PlayerName, const
...
+			case '&':  //< Color code
 			{
-				// Color code
-				i++;
-				if (i >= len)
+				if ((i != 0) && (a_ParseText[i-1] == '\\'))
 				{
-					// Not enough following text
-					break;
+					CurrentText.append(a_ParseText, cursor, i-1-cursor).append("&");
```
madmaxoft, line 122: Style: spaces around `-`

src/CompositeChat.cpp
```diff
@@ -96,50 +96,52 @@ void cCompositeChat::AddShowAchievementPart(const AString & a_PlayerName, const
...
+				if (cursor < i)
 				{
-					// "@@" escape, just put a "@" into the current text and keep parsing as text
-					if (i > first + 1)
-					{
-						CurrentText.append(a_ParseText.c_str() + first, i - first - 1);
-					}
-					first = i + 1;
-					continue;
+					CurrentText.append(a_ParseText, cursor, i-cursor);
```
madmaxoft, line 131: Style: spaces

madmaxoft (changes requested):
The `Server/Plugins/APIDump/APIDesc.lua` also needs to be updated - there's a `cCompositeChat` description there. What about plugin compatibility, did you test any plugins actually using chat colors?

Also, what about `cChatColor`, do we want to keep it? Is it up to date?

tonitch:
> The `Server/Plugins/APIDump/APIDesc.lua` also needs to be updated - there's a `cCompositeChat` description there. 
I did my best, english is a foreign language for me so I might not be clear in my way of explaining it... sorry :)
> What about plugin compatibility, did you test any plugins actually using chat colors?
I didn't, I honestly didn't even tried plugin once since I'm in the cuberite project... So I don't really know which would break except for Core for whom I submitted a patch 
I can maybe do a deprecated backward compatibility for colors as it uses @ and can be detected but for other style effect I don't think it's worth the effort...

> Also, what about `cChatColor`, do we want to keep it? Is it up to date?
I didn't even knew it was a thing, by the look of things this is used by Cuberite/Core... I think it's not a bad thing that both exist as this make an altenative way of coding plugins. either typing the color or going for the standardized minecraft color code
and I also think that if plugin compatibility is a concern, then making cChatColor deprecated at the same time would break every single plugin that uses chat color

12xx12:
I think we should check the plugins released with the cuberite github profile (check the Lua-Repos that are not archived)

12xx12:
Just bumping @tonitch to check if the other plugins were checked.
These changes are good.

tonitch:
> Just bumping @tonitch to check if the other plugins were checked.
> These changes are good.

Sorry I didn't had much time, and will not have for the next two weeks... I will do that as soon as I can!
Thanks for the bump 😊

tonitch:
I'm back! Here is a list of what I checked and what I did!

checked
- [x] Core
- [ ] MagicCarpet
- [ ] Gallery
- [ ] ChatLog
- [ ] Essentials
- [ ] WorldEdit
- [ ] HungerGames
- [ ] APIFuzzing
- [ ] ProtectionAreas
- [ ] Coiny
- [ ] Trady 
- [ ] transapi 
- [ ] transapi-client 
- [ ] SpawnProtect 
- [ ] Handy 
- [ ] ChunkWorx 
- [ ] ChestShop 
- [ ] Commandy

tonitch:
All seems good to me ^^ 
Tell if I need to do anything else ?!

tonitch:
i'm sorry if i'm rude but will this be pulled someday ? ^^

tonitch:
Super! I can wait more, I was just a bit in the dark here ^^
Have fun and best of luck to you!

src/CompositeChat.cpp
```diff
@@ -96,50 +96,52 @@ void cCompositeChat::AddShowAchievementPart(const AString & a_PlayerName, const
 
 
 
-
+/**
+* Parse the input message to add colors or link then add it to the object.
+*
+* It detect every & of the message and the next character for it to colorize.
```
12xx12, line 102:
```suggestion
* It detects every & of the message and the next character for it to colorize.
```

src/CompositeChat.cpp
```diff
@@ -96,50 +96,52 @@ void cCompositeChat::AddShowAchievementPart(const AString & a_PlayerName, const
...
 	{
 		switch (a_ParseText[i])
 		{
-			case '@':
+			case '&':  //< Color code
 			{
-				// Color code
-				i++;
-				if (i >= len)
+				if ((i != 0) && (a_ParseText[i-1] == '\\'))
```
12xx12, line 120: Does the vanilla implementation also use `\` as the escape character?

12xx12:
So in total, the changes look great. I have some minor comments.

src/CompositeChat.cpp
tonitch: I'm pretty sure it doesn't, but in fact it causes trouble when trying to write for instance choco&almond which would take the code &a and write chocolmond with lmond in green... I'm pretty sure there is no way to bypass this in vanilla and I figured it would kill to have this I think ^^ but ofc I can remove it if needed

tests/CompositeChat/CompositeChatTest.cpp
```diff
@@ -38,21 +38,21 @@ static void TestParser1(void)
...
-	TEST_EQUAL(std::get<cCompositeChat::TextPart>(Parts[1]).Style, "@5");
+	TEST_EQUAL(std::get<cCompositeChat::TextPart>(Parts[1]).Style, "35");
 
 	TEST_TRUE(std::holds_alternative<cCompositeChat::UrlPart>(Parts[2]));
-	TEST_EQUAL(std::get<cCompositeChat::UrlPart>(Parts[2]).Style, "@5");
+	TEST_EQUAL(std::get<cCompositeChat::UrlPart>(Parts[2]).Style, "35");
 
 	TEST_TRUE(std::holds_alternative<cCompositeChat::TextPart>(Parts[3]));
-	TEST_EQUAL(std::get<cCompositeChat::TextPart>(Parts[3]).Style, "@5");
+	TEST_EQUAL(std::get<cCompositeChat::TextPart>(Parts[3]).Style, "35");
```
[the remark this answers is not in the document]
tonitch: because it return two colors code. they are all added until you have a &r which means that at this point the color is and 3 then it's 5 and so the style is 35

src/CompositeChat.h
```diff
@@ -185,8 +186,4 @@ class cCompositeChat
 	/** Additional data pertaining to message type, for example, the name of a mtPrivateMsg sender */
 	AString m_AdditionalMessageTypeData;
 
-
-	/** Adds a_AddStyle to a_Style; overwrites the existing style if appropriate.
-	If the style already contains something that a_AddStyle overrides, it is erased first. */
-	void AddStyle(AString & a_Style, const AString & a_AddStyle);
```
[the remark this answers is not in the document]
tonitch: i'm pretty sure I had reasons to that when I did it... but i can't remember... i'll analyse that asap and figure that out

src/CompositeChat.cpp
12xx12: No this is a good idea. Maybe we should add this to the documentation, just in case someone is confused or wants to do this

src/CompositeChat.h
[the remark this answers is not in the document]
12xx12: Maybe you removed the only use case 🤷 Just an idea

tests/CompositeChat/CompositeChatTest.cpp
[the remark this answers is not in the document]
12xx12: Hm weird. But what color does the game display? And earlier if there is a new color code the color was overridden. Is this the same as vanilla?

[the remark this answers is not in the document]
tonitch: it display always the last of the stack. I did it this way because there is not only "color code" but also "formating code" and those can be combined. this is vanilla in this regard. actually I don't know how mojang handle this but I though a "stack" system would be the best ? every &x character add to the stack until one does &r which clear the stack.

src/CompositeChat.h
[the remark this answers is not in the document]
tonitch: might be it. indeed I can't find why this function would be useful and maybe got rid of it... should I implement it back ?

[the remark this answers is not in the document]
12xx12: This isn't is exported to lua so no. This is fine

12xx12 (changes requested):
This is good to go except we need to update the plugin submodules. So the plugin updates need to be merged first. Maybe @tigerw @bearbin  or @NiLSPACE can merge those

mathiascode:
I can't review this PR right now, but I can merge some of the plugin PRs once this PR is merged.

A few questions:
- I assume plugins using the old formatting/color codes keep working without any fatal errors?
- Existing motd messages that use the old color codes will break once the Core PR is merged. Do we care about this?

12xx12:
Nothing will break (in the sense of the application crashing). Using the old color Codes will just result in a weird & character in messages.

Since we provide a proper API to methods in cCompositeChat this might have been dodged.

Since I am writing this on mobile, so I can't check if those methods will stop working (no formatting) because some indicators were changed

tonitch:
All plugins using the API are fine ... The only thing that would break is if plugin writer have implemented color code in the form of "@x"... These character will just be untouched and might be weird but as asked, I checked every plugin in the cuberite's GitHub and I found really little amount of dev using @ char in their code... They almost all use api and so are fine 
Except those i made a PR ofc

12xx12:
I meant if a plugin uses 'AddChatPartStyle' but uses the old keys. Then there would be no formatting without a warning.

12xx12:
Maybe we should add a switch/case case that triggers on the old formatting codes informing the admin that the cold is outdated

tonitch:
> Maybe we should add a switch/case case that triggers on the old formatting codes informing the admin that the cold is outdated

If I remember correctly, I though about that but I'm pretty sure there is conflict... Maybe I can do that only for those that are no more. Should i do anything else for the conflict ?

12xx12:
I don't think so. The new character should be noticed fast :D

src/CompositeChat.cpp
```diff
@@ -404,70 +385,34 @@ void cCompositeChat::AddChatPartStyle(Json::Value & a_Value, const AString & a_P
...
-			case 'u':
-			{
-				// Underlined
-				a_Value["underlined"] = Json::Value(true);
-				break;
-			}
+			case 'k': a_Value["obfuscated"] = Json::Value(true); break;
+			case 'l': a_Value["bold"] = Json::Value(true); break;
+			case 's':  // Deprecated
+				LOG("Value s in AddChatPartStyle() is deprecated");
```
12xx12, line 391:
```suggestion
				LOGERROR("Value s in AddChatPartStyle() is deprecated");
```

Make it red xD

12xx12:
Yeah the deprecation handling is quite inconsistent and since we don't do releases, no dates are set.

src/CompositeChat.cpp
```diff
@@ -404,70 +385,34 @@ void cCompositeChat::AddChatPartStyle(Json::Value & a_Value, const AString & a_P
...
-				a_Value["underlined"] = Json::Value(true);
-				break;
-			}
+			case 'k': a_Value["obfuscated"] = Json::Value(true); break;
+			case 'l': a_Value["bold"] = Json::Value(true); break;
+			case 's':  // Deprecated
+				LOGERROR("Value s in AddChatPartStyle() is deprecated");
+			case 'm': a_Value["strikethrough"] = Json::Value(true); break;
+			case 'u':  // Deprecated
+				LOG("Value u in AddChatPartStyle() is deprecated");
```
12xx12, line 394:
```suggestion
				LOGERROR("Value u in AddChatPartStyle() is deprecated");
```

src/CompositeChat.cpp
```diff
@@ -404,70 +385,34 @@ void cCompositeChat::AddChatPartStyle(Json::Value & a_Value, const AString & a_P
...
+			case 'n': a_Value["underlined"] = Json::Value(true); break;
+			case 'i':  // Deprecated
+				LOG("Value i in AddChatPartStyle() is deprecated");
```
12xx12, line 397:
```suggestion
				LOGERROR("Value i in AddChatPartStyle() is deprecated");
```

12xx12 (changes requested):
Last one. Then I'll approve

12xx12 (approved):
Waiting a day for any vetos

bearbin (approved):
Looks fine generally, I will not veto. I'm not totally confident with your refactor to the chat parsing logic, you seem to have made more changes than strictly necessary here and it's some complicated spaghettti with intrinsically unsafe data structures. 

In an ideal world that whole mess would be refactored with modern C++ and use some safe primitives. But that's a bigger job, this is fine to get finally merged in.

PR #5424, opened 2022-06-01
Remove redundant initialisations

tigerw:
The static buffers here are immediately passed in to be overwritten.


tigerw:
Ref d98e2bdaa3903de29f62791c05472a693b0ea36b

PR #5431, opened 2022-07-10
ClientHandle: Fix block interaction regressions

### LINKED ISSUE #5422: Can't put out fire.
Client version: 1.12.2
Server OS: Windows
Cuberite Commit id: a8ef9d14376fccb4b4b595d52fc9f9c7b5538043

### Expected behavior
When trying to put out the fire, it does it normally.

### Actual behavior
You can't in survival and the block you are looking at also breaks on creative

Try to put out fire in survival and creative

tigerw:
* Fix reach distance being halved.
* Fix fire extinguish not working, fixes #5422.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1117,17 +1117,15 @@ void cClientHandle::HandleLeftClick(Vector3i a_BlockPos, eBlockFace a_BlockFace,
 		/* Check for clickthrough-blocks:
 		When the user breaks a fire block, the client send the wrong block location.
 		We must find the right block with the face direction. */
-
-		Vector3i BlockPos = a_BlockPos;
-		AddFaceDirection(BlockPos, a_BlockFace);
-
-		if (cChunkDef::IsValidHeight(BlockPos) && cBlockInfo::IsClickedThrough(m_Player->GetWorld()->GetBlock(BlockPos)))
+		if (
+			const auto InterferingPosition = AddFaceDirection(a_BlockPos, a_BlockFace);
+			cChunkDef::IsValidHeight(InterferingPosition) && cBlockInfo::IsClickedThrough(m_Player->GetWorld()->GetBlock(InterferingPosition))
+		)
 		{
-			a_BlockPos = BlockPos;
+			a_BlockPos = InterferingPosition;
 		}
 
-		constexpr double MaxBlockDistance = 6.0;
-		if (!cBoundingBox(m_Player->GetPosition(), MaxBlockDistance).IsInside(a_BlockPos))
+		if (!IsWithinReach(a_BlockPos))
 		{
 			m_Player->SendBlocksAround(a_BlockPos, 2);
 			return;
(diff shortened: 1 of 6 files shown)


src/ClientHandle.cpp
```diff
@@ -2101,6 +2078,19 @@ bool cClientHandle::CheckBlockInteractionsRate(void)
...
 
 
+bool cClientHandle::IsWithinReach(const Vector3i a_Position) const
+{
+	// Distance from the block's center to the player's eye height.
+	const double Distance = (Vector3d(0.5, 0.5, 0.5) + a_Position - m_Player->GetEyePosition()).SqrLength();
+
+	// _X 2014 - 11 - 25: I've maxed at 5.26 with a Survival client and 5.78 with a Creative client in my tests.
+	return Distance <= (m_Player->IsGameModeCreative() ? 33.4084 : 27.6676);
+}
```
bearbin, line 2088: I think this ought to be a player function, rather than a clientHandle function

src/ClientHandle.cpp
```diff
@@ -1117,17 +1117,15 @@ void cClientHandle::HandleLeftClick(Vector3i a_BlockPos, eBlockFace a_BlockFace,
 		/* Check for clickthrough-blocks:
 		When the user breaks a fire block, the client send the wrong block location.
 		We must find the right block with the face direction. */
-
-		Vector3i BlockPos = a_BlockPos;
-		AddFaceDirection(BlockPos, a_BlockFace);
```
bearbin, line 1122: Can we kill the original AddFaceDirection that takes a reference now? Would be good if we just had one function with one semantic rather than two that do different things with no warning.

src/ClientHandle.cpp
tigerw: Now that you mention it I've had the thought that all of the client event handlers should be moved to player. Client handle does whatever checks it needs and then calls OnRightClicked/OnBlockDigBegin/OnBowCharging/and so on, leaving the player to process it. I don't think having player effectively expose all of its members via getters so that ClientHandle can act on its behalf is very nice.

src/Defines.h
```diff
@@ -571,11 +571,7 @@ void AddFaceDirection(int & a_BlockX, int & a_BlockY, int & a_BlockZ, eBlockFace
 
 /** Returns the coords of a block that is neighboring the specified position through its specified face.
 If a_IsInverse is true, the opposite direction is used instead. */
-inline Vector3i AddFaceDirection(Vector3i a_Pos, eBlockFace a_BlockFace, bool a_bInverse = false)
-{
-	AddFaceDirection(a_Pos.x, a_Pos.y, a_Pos.z, a_BlockFace, a_bInverse);
-	return a_Pos;
-}
+Vector3i AddFaceDirection(Vector3i a_Pos, eBlockFace a_BlockFace, bool a_bInverse = false);
```
tigerw, line 574: having to write your function prototypes twice in an annoying thing

a_bInverse -> a_Inverse

src/ClientHandle.cpp
tigerw: Dunno about moving it right now, can we leave it?

bearbin: Yeah, it's fine if you want to keep it for now.

bearbin (approved):

PR #5433, opened 2022-07-15
ItemHandler: Fix torch handler not being called

### LINKED ISSUE #5432: Invisible Torches after Reconnecting to Server
Client version: 1.12.2
Server OS: DietPi (debian-based on Raspberry Pi 4)
Cuberite Commit id: unsure

### Expected behavior
Placing a torch during a session, will be visible, and remain visible when reconnecting to the server later. Client is running MX Linux (Debian), MultiMC 5, running instance 1.12.2 without mods/vanilla.

### Actual behavior
See image [here](https://photos.app.goo.gl/w4hZXxTcNPw24X8t8). A torch will remain visible after placing, during gameplay. After re-connecting to server, torch asset does not appear visible; light is cast, and when the block a torch is placed on is mined, the torch is retrievable.

### Server log
```
Relevant part of server log
```

tigerw:
* Fixes #5432.

### DIFF
--- a/src/Items/ItemHandler.cpp
+++ b/src/Items/ItemHandler.cpp
@@ -533,7 +533,7 @@ namespace
 	constexpr cDefaultItemHandler           ItemTerracottaHandler                      (E_BLOCK_TERRACOTTA);
 	constexpr cDefaultItemHandler           ItemTippedArrowHandler                     (E_ITEM_TIPPED_ARROW);
 	constexpr cDefaultItemHandler           ItemTNTHandler                             (E_BLOCK_TNT);
-	constexpr cDefaultItemHandler           ItemTorchBlockHandler                      (E_BLOCK_TORCH);
+	constexpr cItemTorchHandler             ItemTorchHandler                           (E_BLOCK_TORCH);
 	constexpr cDefaultItemHandler           ItemTotemOfUndyingHandler                  (E_ITEM_TOTEM_OF_UNDYING);
 	constexpr cItemChestHandler             ItemTrappedChestHandler                    (E_BLOCK_TRAPPED_CHEST);
 	constexpr cDefaultItemHandler           ItemTripwireHandler                        (E_BLOCK_TRIPWIRE);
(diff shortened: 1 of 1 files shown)


bearbin (approved):

blueyshark:
Not to sound like an absolute ignoramus, but how do I install this fix into my server? Do I have to wait for a repo release?
Thanks for being so speedy on this!

bearbin:
It'll depend on how you installed Cuberite in the first place - you should be able to just download again and use the new executable file. You won't need to wait for a release as we rebuilt the binaries every time a new commit is pushed to master. Not sure how DietPi works but you may be able to do an update of Cuberite or you may need to reinstall the Cuberite package - either should work.

tigerw:
According to [DietPie](https://dietpi.com/docs/software/gaming/#cuberite) the right command is `dietpi-software reinstall 52`.

blueyshark:
@bearbin Hello - one last query, apologies - I've uninstalled/reinstalled cuberite, which removed the cuberite folder from the server's /mnt
I thankfully made a backup of the folder - what of these folders/files would I want to copy back onto the server's mnt without overwriting the updated binaries? Again, thanks for your help :pray:

bearbin:
Again, not sure how DietPi works, but the only important file for this update is going to be the Cuberite executable file itself, the others are all support files that haven't been changed in the last commit - you can overwrite the new copies of them with your backup and it will be fine. You'll also need to copy in your world files - as long as you haven't started up the new server it won't have created a new world and you can just copy in the old one.

blueyshark:
@bearbin hello - I'm sorry to say, I'm still getting the same issue with the torches, after both manually wiping/reinstalling cuberite, and also using the `dietpi-software reinstall 52` that @tigerw mentioned.
Could it be more to do with the client side of things?

bearbin:
Our 64 bit ARM build is down at the moment, are you using a 64 bit pi?

bearbin:
@tigerw The reason the build is broken seems to be that system lua detection is broken and we're trying to execute the compiled lua - any ideas why this would be going on? https://builds.cuberite.org/job/linux-aarch64/36/console

blueyshark:
> Our 64 bit ARM build is down at the moment, are you using a 64 bit pi?

32 bit OS.
Just to be clear - since reinstalling the torches are working as normal. The ones I placed before the fix are still invisible, but that's fine, I will slowly hunt them down and replace. Thanks.

bearbin:
How odd, no idea how that could be happening. Maybe Tiger will have more ideas since he wrote the fix. At least it looks like the build got through to you :)

tigerw:
I think the way nightlybuild.sh is overriding CC and CXX to point to a cross-compiler is interfering with CMake's cross-compile detection. We should make it use a toolchain file like the ARM32 one does.

PR #5435, opened 2022-07-19
Explosions: improve performance

tigerw:
Tested on Release, Windows ia32, 17³ sphere of TNT created with WorldEdit.

Summary: changes reduce time for TPS to recover by **39%** when no pickup entities are spawned, and make the server not freeze compeltely when pickup entities are spawned. The changes achieve this by coalescing pickup spawns during explosions, improving pickup behaviour not to do unnecessary work when idle, and improving block tracer performance.

Below are the tests I ran. Screenshots taken without pickup spawns since the client couldn't handle it otherwise.

Prior (with pickup spawns). Server doesn't recover due to tens of thousands of pickup entities, test stopped at 2 min 25 sec.
Prior (with pickup spawns)

Prior (without pickups). Server recovers in 56 sec.
Prior (without pickups)

Prior blast result.
Prior blast result

After (with pickup spawns). Server recovers to 19.68 TPS after 51 sec, but with high CPU load.
image

After (without pickups). Server recovers in 34 sec.
image

After blast result.
image


src/Physics/Tracers/LineBlockTracer.cpp
```diff
@@ -0,0 +1,282 @@
...
+	{
+		if (a_End.y >= cChunkDef::Height)
+		{
+			a_Callbacks.OnNoMoreHits();
+			return true;
+		}
+
+		a_Start = FixStartAboveWorld(a_Start, a_End);
+		a_Callbacks.OnIntoWorld(a_Start);
+	}
```
bearbin, line 99: This can have it's very own function :)

tigerw:
I have some minor updates for this one, but first I will make #5434 ready, which this branch depends on.

Issue #3349, opened 2016-08-25
Serverside texturepacks?


ytterdal:
Are there any plans for implementing serverside texturepacks?

sphinxc0re:
There are plans, but it is quite difficult to implement since there is no documentation about how the texturepacks are send to the client

mohe2015:
http://wiki.vg/Protocol#Resource_Pack_Send and http://wiki.vg/Protocol#Resource_Pack_Status. That shouldn't be that difficult.

Seadragon91:
@mohe2015 Thanks for the links :)

bibo38:
I would also request the feature of forcing the client to use a texture pack(e.g. kicking the player, when the texture pack is not accepted/loaded).
It doesn't need to be implemented directly into Cuberite, but otherwise it should be possible doing this using Plugins.

madmaxoft:
This should be rather easy to implement.
 - Add a setting to settings.ini for the resource pack URL
 - Upon startup, download the resource pack using `cUrlClient`, and calculate SHA1 checksum (Add a `cServer` function to do that, export to Lua API so that plugins may request an update)
 - Implement the required protocol packets
 - Add a new `cPlayer` member variable and getter (+Lua API) specifying what the client responded in the Resource Pack Status packet (ok, failed, declined; what is "accepted"?)
 - Add a test code in the Debuggers plugin that kicks all players who decline a resource pack (to test the Lua API functionality); this can then be pulled into a separate plugin.

bibo38:
> Add a new cPlayer member variable and getter (+Lua API) specifying what the client responded in the Resource Pack Status packet (ok, failed, declined; what is "accepted"?)

I would suggest to use an Event instead of an variable + Getter. Otherwise a Plugin needs to constantly query the `cPlayer` object to get a status change. And if a plugin needs to save the status, it can save it locally.

P.S.: @madmaxoft Accepted means(like declined) that the user accepted to use the Resource Pack. And after download Minecraft emits the ok Status to inform the server, that the texture pack was downloaded/applied successfully (and otherwise failed).

mohe2015:
I think the SHA1 checksum should be a setting in settings.ini. Otherwise the server has to calculate it every time.

mohe2015:
That's like it is in the vanilla server

madmaxoft:
[...]

PR #5440, opened 2022-09-14
Kick clients when resource pack rejected

plan1231:
This PR attempts to complete what is described in #3349

When a client sends a Resource Pack Status packet indicating that it rejected the server's resource pack, we should kick it.

I'm a complete noob to this, please forgive me if I made any stupid mistakes in this PR.

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1884,6 +1884,17 @@ void cClientHandle::HandleUseItem(bool a_UsedMainHand)
 
 
 
+void cClientHandle::HandleResourcePack(UInt8 a_Status)
+{
+	// Kick player if client declined the resource pack
+	if (a_Status == 1)
+		Kick("You must accept the resource pack");
+}
+
+
+
+
+
 void cClientHandle::HandleRespawn(void)
 {
 	if (m_Player->GetHealth() > 0)
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1887,8 +1887,10 @@ void cClientHandle::HandleUseItem(bool a_UsedMainHand)
 void cClientHandle::HandleResourcePack(UInt8 a_Status)
 {
 	// Kick player if client declined the resource pack
-	if (a_Status == 1)
+	if ((a_Status == 1) && cRoot::Get()->GetServer()->ShouldRequireResourcePack())
+	{
 		Kick("You must accept the resource pack");
+	}
 }
 
 
--- a/src/ClientHandle.h
+++ b/src/ClientHandle.h
@@ -377,6 +377,7 @@ public:  // tolua_export
 
 
 	void HandlePluginMessage    (const AString & a_Channel, ContiguousByteBufferView a_Message);
+	void HandleResourcePack     (UInt8 a_Status);
 	void HandleRespawn          (void);
 	void HandleRightClick       (Vector3i a_BlockPos, eBlockFace a_BlockFace, Vector3i a_Cursor, bool a_UsedMainHand);
 	void HandleSlotSelected     (Int16 a_SlotNum);
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -157,6 +157,7 @@ bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_Shoul
 	m_MaxPlayers = static_cast<size_t>(a_Settings.GetValueSetI("Server", "MaxPlayers", 100));
 	m_bIsHardcore = a_Settings.GetValueSetB("Server", "HardcoreEnabled", false);
 	m_bAllowMultiLogin = a_Settings.GetValueSetB("Server", "AllowMultiLogin", false);
+	m_RequireResourcePack = a_Settings.GetValueSetB("Server", "RequireResourcePack", false);
 	m_ResourcePackUrl = a_Settings.GetValueSet("Server", "ResourcePackUrl", "");
 	m_CustomRedirectUrl = a_Settings.GetValueSet("Server", "CustomRedirectUrl", "https://youtu.be/dQw4w9WgXcQ");
 
--- a/src/Server.h
+++ b/src/Server.h
@@ -94,6 +94,9 @@ public:
 
 	// tolua_end
 
+	/** Returns true if clients must accept resource pack. This is read from the settings. */
+	bool ShouldRequireResourcePack(void) { return m_RequireResourcePack; }
+
 	const AString & GetResourcePackUrl(void) { return m_ResourcePackUrl; }
 
 	std::string_view GetCustomRedirectUrl(void) { return m_CustomRedirectUrl; }
@@ -223,6 +226,7 @@ private:
 	AString m_FaviconData;
 	size_t m_MaxPlayers;
 	bool m_bIsHardcore;
+	bool m_RequireResourcePack;
 	AString m_ResourcePackUrl;
 	AString m_CustomRedirectUrl;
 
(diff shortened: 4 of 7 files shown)


src/ClientHandle.cpp
```diff
@@ -1884,6 +1884,18 @@ void cClientHandle::HandleUseItem(bool a_UsedMainHand)
 
 
 
+
+void cClientHandle::HandleResourcePack(UInt8 a_Status)
+{
+	// Kick player if client declined the resource pack
+	if(a_Status == 1)
+		Kick("You must accept the resource pack");
```
plan1231, line 1892: From the discussion in #3349 , it seems like I need to call a lua hook here. Since I don't know how, I'm leaving this code here for now to signify the gist of what I think the server should do, i.e. kick the client if it rejects the resource pack

src/ClientHandle.h
```diff
@@ -377,6 +377,7 @@ class cClientHandle  // tolua_export
 
 
 	void HandlePluginMessage    (const AString & a_Channel, ContiguousByteBufferView a_Message);
+	void HandleResourcePack     (UInt8 a_Status);
```
plan1231, line 380: I debated between naming this function `HandleResourcePack` and `HandleResourcePackStatus`.
I chose the former because:
1. naming it the latter would mean I would have to add spaces to the rest of the lines to realign the function declarations
2. In later versions of the protocol on the wiki, this packet gets renamed to 'Resource Pack (serverbound)'

Janmm14:
Sometimes a server resource pack might be optional, there should probably be a configuration option whether to kick or not.

src/ClientHandle.cpp
tonitch: I did the part needed for the hook if you want, I made a PR on your own branch at https://github.com/plan1231/cuberite/pull/1

plan1231: Thanks tonitch! Sorry I didn't see this earlier.

src/Server.h
```diff
@@ -223,6 +226,7 @@ class cServer
 	AString m_FaviconData;
 	size_t m_MaxPlayers;
 	bool m_bIsHardcore;
+	bool m_RequireResourcePack;
```
plan1231, line 229: I'm not sure if this field, along with it's associated getter, is still necessary. Initially, I added it so that server owners can indicate that the resource pack is optional and we don't have to kick clients for rejecting it. But tonitch then added a call to a lua hook, so it seems like the same functionality can be done by lua?

tonitch: Actually, I think this can be a good option for non-dev to be able to enable/disable this feature. maybe we can keep both or just in plugin but then it would be better to add an option in core or web to tweak that... I don't know I'm not in the project since long enough to understand the full vision. @12xx12 @madmaxoft  What do you think ?

src/Bindings/PluginManager.cpp
```diff
@@ -1097,6 +1097,18 @@ bool cPluginManager::CallHookProjectileHitEntity(cProjectileEntity & a_Projectil
 
 
 
+bool cPluginManager::CallHookResourcePack(const UInt8 a_Status)
+{
+	return GenericCallHook(HOOK_RESOURCEPACK, [&](cPlugin * a_Plugin){
```
tonitch, line 1102: small syntax error here, I made a pr on your fork for that. That's the reason it doesn't pass jenkins 
https://github.com/plan1231/cuberite/pull/2

src/Server.cpp
```diff
@@ -157,6 +157,7 @@ bool cServer::InitServer(cSettingsRepositoryInterface & a_Settings, bool a_Shoul
 	m_MaxPlayers = static_cast<size_t>(a_Settings.GetValueSetI("Server", "MaxPlayers", 100));
 	m_bIsHardcore = a_Settings.GetValueSetB("Server", "HardcoreEnabled", false);
 	m_bAllowMultiLogin = a_Settings.GetValueSetB("Server", "AllowMultiLogin", false);
+	m_RequireResourcePack = a_Settings.GetValueSetB("Server", "RequireResourcePack", false);
```
tonitch, line 160: depending on https://github.com/cuberite/cuberite/pull/5440/files#r979201537 this line should stay of go

src/ClientHandle.cpp
```diff
@@ -1884,6 +1884,19 @@ void cClientHandle::HandleUseItem(bool a_UsedMainHand)
 
 
 
+// @param a_Status 0: successfully loaded, 1: declined, 2: failed download, 3: accepted
+void cClientHandle::HandleResourcePack(UInt8 a_Status)
+{
+	if (cRoot::Get()->GetPluginManager()->CallHookResourcePack(a_Status))
```
tonitch, line 1890: depending https://github.com/cuberite/cuberite/pull/5440/files#r979201537
this stay like that or is and'd with the config file response

tonitch (changes requested):
there is just the syntax error to sort
then choose what we do about config file but this seems good to me

src/Bindings/PluginManager.cpp
plan1231: I merged that PR but I think it got merged into another branch, so it's not showing up here. Would it be possible to create a PR that merges into `plan1231:resource-pack`?

tonitch: Well, actually there is only one character diff, maybe you can sort it out by yourself ? I don't have access to a PC till tomorrow and it seems like I can't do pr on my phone 😅

plan1231: Oh shoot sorry! I didn't know earlier, because I didn't know how to get the diff between your new commit and this branch, because the PR's target was the master branch. I'll definitely go learn git better so that I don't take up your time with requests like this in the future

tonitch: Don't worry, I can see the confusion, sorry about that!
I tried to redo the pr but unfortunately, and for some weird reason, I can't do a pr on your own branch anymore.
So here is the fix commit I did, you can see the change. https://github.com/cuberite/cuberite/commit/f9d54a0ee707bbbb55b5f9e181a96e3e4a09f369 if you can just implement this commit, or if you can't just modify it by yourself ?
just by modifying this, the compilation check should pass!

tonitch: you still on it tho ?

src/Server.h
12xx12: I am not sure about the Lua hook. Since there is no way to resent the resource pack the only thing you could do is kicking the player and that could be done with a simple config value

tonitch: > I am not sure about the Lua hook. Since there is no way to resent the resource pack the only thing you could do is kicking the player and that could be done with a simple config value

that make sence, I just know some people think about crazy way of using plugin functionality even in java... for instance, admin would not need to accept or someting like that... but in the end... yhea it's pretty specific so might not be needed... in this case we could just reverse to https://github.com/cuberite/cuberite/commit/ba835236af3678f78c6b2fce4844907f5f536321

src/Bindings/PluginManager.cpp
12xx12: Don't worry. No stress at all. We're happy for anyone who wants to contribute! Work at your pace and possibilities

tonitch (approved):
seems good to me then :D

12xx12 (approved):

12xx12:
Waiting a day for any vetos

12xx12:
One more thing:

Due to the license used with this project, you need to add yourself to the CONTRIBUTORS file to show that you agree with the license
@plan1231

PR #5444, opened 2022-09-27
Fix #5442

### LINKED ISSUE #5442: API: Passing -1 to function cMonster:FamilyFromType crashes the server
Server OS: Linux
Cuberite Commit id: b73cd1179fedc4abd5c388ae46f507ff3e36c213

### Expected behavior
Passing -1 to function cMonster:FamilyFromType should not crash the server.

### Actual behavior
Server crashes at:
https://github.com/cuberite/cuberite/blob/b73cd1179fedc4abd5c388ae46f507ff3e36c213/src/Mobs/Monster.cpp#L1173

Call this lua code from a command:
`cMonster:FamilyFromType(-1)`

Found by APIFuzzing.

tonitch:
Should solve #5442 
I'm sorry for the other pr, I just messed it up so I make a new one...

### DIFF
--- a/src/MobCensus.cpp
+++ b/src/MobCensus.cpp
@@ -38,10 +38,8 @@ int cMobCensus::GetCapMultiplier(cMonster::eFamily a_MobFamily)
 		case cMonster::mfPassive: return 11;
 		case cMonster::mfAmbient: return 16;
 		case cMonster::mfWater:   return 5;
-		case cMonster::mfNoSpawn:
-		{
-			break;
-		}
+		case cMonster::mfNoSpawn: break;
+		case cMonster::mfUnknown: break;
 	}
 	UNREACHABLE("Unsupported mob family");
 }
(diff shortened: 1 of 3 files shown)


12xx12 (approved):
This won't crash. Good to merge

tigerw:
Will this work if you pass -2? i.e. I think the problem is Lua calling C++ with arbitrary integer values but C++ not expecting them since it works based off the fixed set of enumeration values. I would suggest writing a manual binding similar to [this](https://github.com/cuberite/cuberite/blob/3b2a501524eb4cd23e53cf5ddb04ea8a615f0460/src/Bindings/ManualBindings_World.cpp#L920) and adding a manual error check like so: https://github.com/cuberite/cuberite/blob/3b2a501524eb4cd23e53cf5ddb04ea8a615f0460/src/Bindings/ManualBindings_World.cpp#L962-L965 to cleanly report to the plugin that it did something wrong.

Something like `if pluginmonstertype < mtInvalidType || pluginmonstertype > zombievillager { error } else { call cMonster::FamilyFromType(pluginmonstertype) }`

PR #5447, opened 2022-10-04
Handle newlines in cIniFile

### LINKED ISSUE #5026: Make cIniFile properly parse newline characters
It's currently impossible to add newline characters (`\n`) inside ini file values. Cuberite will use the characters litterally. Instead it would be nice to make Cuberite handle the newlines like actual newlines. The best place to do this would probably be the [ReadFile](https://github.com/cuberite/cuberite/blob/master/src/IniFile.cpp#L149) and [WriteFile](https://github.com/cuberite/cuberite/blob/master/src/IniFile.cpp#L226) functions. Otherwise all `GetValue` and `GetValueSet` variants would have to be updated.

MacaylaMarvelous81:
This pull request replaces \n with newline a newline when reading an ini file. When writing to an ini file, the newline will be replaced with \n.

Example 1:
```ini
# settings.ini
# ...
[Server]
Description=Line 1\nLine 2
# ...
```
issue5026-1

Example 2:
```lua
--- ...
local iniFile = cIniFile()

iniFile:SetValue("Key", "ValueName", "My Line 1\nMy Line 2")
iniFile:WriteFile("test.ini")
--- ...
```
```ini
# test.ini
[Key]
ValueName=My Line 1\nMy Line 2
```

Fixes #5026


NiLSPACE:
Isn't including regex for this overkill? Wouldn't a `myString.replace("\\n", std::endl)` work as well?

12xx12:
Thank you for your contribution!

Firstly, please add yourself to the CONTRIBUTORS file to acknowledge our license and that your code is used in accordance to that.

Secondly I agree with @NiLSPACE that regex is a bit overkill.

Thirdly: Maybe we should add a sentence [here](https://github.com/cuberite/cuberite/blob/c61dab94bd52ab60da7b14b16402fa4dc20c5e4d/Server/Plugins/APIDump/APIDesc.lua#L5348), that line breaks are now supported and how they work

MacaylaMarvelous81:
Should I use [ReplaceString from StringUtils](https://github.com/cuberite/cuberite/blob/4b8952e438da8e57d6442a0b7451dac8060f154b/src/StringUtils.cpp#L348-L362) instead of regex?

12xx12:
Yes, please

bearbin:
Indeed, regex is a bit overkill, ReplaceString is a good alternative. 

Also, please make sure it's possible to write a literal `\n` in the config! I suggest matching on `\\n` and replacing with `\n` (and the same in reverse). If we had other special characters to escape, replacing `\\` with `\` might make sense but just matching `\\n` makes most sense here to avoid mangling data unnecessarily.

12xx12:
With the constraint mentioned, maybe regex is the way to go. If we want to add more for whatever reason, this is the easiest way to extend.

MacaylaMarvelous81:
Sorry for taking forever to push.. I'm having trouble with pushing to my fork

12xx12 (approved):

src/IniFile.cpp
```diff
@@ -191,6 +193,7 @@ bool cIniFile::WriteFile(const AString & a_FileName) const
 	// Normally you would use ofstream, but the SGI CC compiler has
 	// a few bugs with ofstream. So ... fstream used.
 	fstream f;
+	AString runvalue, writevalue;
```
bearbin, line 196: Not sure what the point of having two separate strings here is - since writevalue is always set to the value of runvalue! Same for rawvalue up above.

Please remove one.

bearbin (changes requested):
This doesn't address my comment made earlier; but I'd be happy to merge without that fix. 

I do want the duplicated variables sorted out, no reason to have them.

12xx12:
I Fixed your request @bearbin

bearbin (approved):

PR #5454, opened 2022-10-30
Create NBT Support with Variant in cItem

12xx12:
This PR tries to kick of further development if additional data is stored into an item.

There is a new member `m_Properties` which is a vector of a variant which stores all possible data.
Access is done with `get<>()` and `set<>()` which both are templated methods which return a optional which contains the requested data.

Any use in the C++ code is adapted to this. Old Lua Binding should be intact by manual binding the old values.

Im am not 100% sure about my Lua implementation, please check that!

### DIFF
--- a/src/Bindings/DeprecatedBindings.cpp
+++ b/src/Bindings/DeprecatedBindings.cpp
@@ -325,7 +325,12 @@ static int tolua_get_cItem_m_Lore(lua_State * tolua_S)
 	const cItem * Self = nullptr;
 	L.GetStackValue(1, Self);
 
-	AString LoreString = StringJoin(Self->m_LoreTable, "`");
+	AString LoreString;
+	auto DisplayProperties = Self->get<cItem::cDisplayProperties>();
+	if (DisplayProperties.has_value())
+	{
+		LoreString = StringJoin(DisplayProperties.value().m_LoreTable, "`");
+	}
 
 	L.Push(LoreString);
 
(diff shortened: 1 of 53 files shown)


tigerw:
Looks good, seems like the right idea! My main initial observation is that std::optional isn't very graceful; one liners are expanding into 4+ lines before our very eyes. I don't think there's much we can do about that.

For the tags we're storing in the variant, I remember vanilla putting a lot of fields in the root level but we could group ours by purpose, mainly to prevent any conflicts. It's also surprising that vanilla calls for the item's damage to also be stored in the variant as a "general tag", this will probably cause a bit a breakage and API headaches.

I can have a more detailed look 🔍 later...

12xx12:
Reguarding `std::optional`

Yeah. This isn't very graceful indeed. I thought about that but returning a Boolean and using a reference as a parameter as the better didn't seem better. And creating the requested value if it didn't exist destroys the purpose of the variant vector.

Reguarding the damage.

Good point. I kept the integer values purposefully out of the variant for easy access and since they consume little memory it shouldn't be that bad.
Since 1.13+ requires us to transmit the item damage in the NBT data to correctly display the damage and the damage is actual damage not some value which might indicate "sub-items" like wood this value might be restricted to only tools/armor/... in future versions. This could be put together with the unbreakable tag I indicated

12xx12:
I need some help with the manual bindings to the firework item. I am not quite sure how to obtain a pointer to the firework item since a pointer is needed

12xx12:
Hello CI are you there?

PR #5456, opened 2022-11-06
MojangAPI: Update certificates

### LINKED ISSUE #4832: Auth with using ely.by alternative to Mojang type Authorization
Client version: 1.12.2
Server OS: Android
Cuberite Commit id: Last at time writing this issue (a473e8b)

How can i use this? If trying to change addressess, error SSL Handshake -0x2700 
https://docs.ely.by/ru/api.html

tigerw:
* Fixes 2115 (again).
* Fixes #4832.

Anyone available to test?


12xx12:
What needs to be tested?

tigerw:
I don't have a working MC account, could you help test online mode log ins and make sure they're successful?

src/Protocol/MojangAPI.cpp
```diff
@@ -43,103 +43,80 @@ const int MAX_PER_QUERY = 100;
...
+		"DQYJKoZIhvcNAQEFBQADggEBAMucN6pIExIK+t1EnE9SsPTfrgT1eXkIoyQY/Esr\n"
+		"hMAtudXH/vTBH1jLuG2cenTnmCmrEbXjcKChzUyImZOMkXDiqw8cvpOp/2PV5Adg\n"
+		"06O/nVsJ8dWO41P0jmP6P6fbtGbfYmbW0W5BjfIttep3Sp+dWOIrWcBAI+0tKIJF\n"
+		"PnlUkiaY4IBIqDfv8NZ5YBberOgOzW6sRBc4L0na4UU+Krk2U886UAb3LujEV0ls\n"
+		"YSEY1QSteDwsOoBrp+uvFRTp2InBuThs4pFsiv9kuXclVzDAGySj4dzp30d8tbQk\n"
+		"CAUw7C29C79Fv1C5qfPrmAESrciIxpg0X40KPMbp1ZWVbd4=\n"
+		"-----END CERTIFICATE-----\n"
+
+		// Amazon Root CA 1 (api.mojang.com)
 		// Downloaded from https://certs.starfieldtech.com/repository/
```
bearbin, line 72: rather `https://www.amazontrust.com/repository/`?

bearbin (approved):
Looks good to me, waiting on @12xx12 to confirm.

12xx12:
It doesn't seem to work for me.

`[c707f29929a46bd7|08:46:56] SecureRequest: SSL reading failed: -0x4c`

12xx12:
Ok. Weird. On My Windows 10 Machine, everything is fine. On my M1 MBP, the login fails with the error mentioned earlier.

bearbin:
I'm going to merge this, better to be working on some platforms than none! We should definitely investigate why it's not working on the M1 though.

tigerw:
0x4C = MBEDTLS_ERR_NET_RECV_FAILED

I reproduced this issue on a Windows:

https://github.com/cuberite/cuberite/blob/8a763d3bedac3eabee9c1ca022be53038ba3fc54/src/mbedTLS%2B%2B/BlockingSslClientSocket.cpp#L278-L299

will return RECV_FAILED when m_IsConnected is false (and more more data), but m_IsConnected is set to false on both clean and unclean disconnections. I think it graceful disconnects should in fact return MBEDTLS_ERR_SSL_CONN_EOF and errored disconnects MBEDTLS_ERR_NET_CONN_RESET.

With that said, given that MojangAPI is trying to implement an ad-hoc HTTP client, I will try to see how easy it is to migrate to our other home-grown HTTP client, cUrlClient.

tigerw:
The TCP class also has similar code, but a different problem in that it always returns WANT_READ. I'm not sure how mbedtls handles not being notified of socket close, hopefully well.

https://github.com/cuberite/cuberite/blob/8a763d3bedac3eabee9c1ca022be53038ba3fc54/src/OSSupport/TCPLinkImpl.cpp#L660-L676

I think re-implementing the whole TCP/TLS/HTTP stack is kinda out of our remit. We could consider using a library in the future; I know Windows has WinHTTP, Mac has NSURLSession, and cURL runs on everything else.

tigerw:
That said cURL is huge, NSURLSession has an Objective-C API, and HTTP libraries in general have opinions about the encryption and networking libraries they use, and make hooking up your own either awkward or impossible, which means integration with our mbedtls and libevent is not trivial.

mathiascode:
> On my M1 MBP, the login fails with the error mentioned earlier.

It would seem like every platform other than Windows is affected by the MBEDTLS_ERR_NET_RECV_FAILED issue (Linux, macOS, Android confirmed).

PR #5460, opened 2022-11-12
adding endermite

### LINKED ISSUE #1408: Add endermites
http://minecraft.gamepedia.com/Endermite

tonitch:
Adding Endermite

currently not working as it's the first mob I'm adding and I don't know what I should register that I didn't yet, but I will figure it out later... 

Trying to clear 1.8 Milestone

TODO:

- [x] Have a 5% chance to spawn when a player-thrown ender pearl lands
    - They spawn at the player's original position when the pearl lands, not at the pearl's landing site. 
- [x] Despawns within 2 minutes, regardless of the player's distance
    - except for when it has been named with a [name tag](https://minecraft.fandom.com/wiki/Name_tag). 
- [x] Attack player within 16 b
- [x] Enderman attack Endermite within 64 b 
- [x] As [arthropods](https://minecraft.fandom.com/wiki/Arthropod), endermites are affected by the [Bane of Arthropods](https://minecraft.fandom.com/wiki/Bane_of_Arthropods) enchantment.

Fixes #1408

### DIFF
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -50,35 +50,50 @@ cMonster * cAggressiveMonster::GetMonsterOfTypeInSight(eMonsterType a_MobType, u
 {
 
 	cMonster * FoundTarget = nullptr;
+	auto MinimumDistance = static_cast<double>(a_SightDistance * a_SightDistance);
+
+	class cCallback : public cBlockTracer::cCallbacks
+	{
+		public:
+		bool OnNextBlock(Vector3i a_BlockPos, BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, eBlockFace a_EntryFace) override
+		{
+			return a_BlockType != E_BLOCK_AIR;
+		}
+	};
+
+	auto Callbacks = cCallback();
+	auto Tracer = cLineBlockTracer(*GetWorld(), Callbacks);
 
 	cEntityCallback Callback = [&](cEntity & a_Entity)
-			{
-				if (!a_Entity.IsMob())
-				{
-					return false;
-				}
+	{
+		if (!a_Entity.IsMob())
+		{
+			return false;
+		}
 
-				cMonster & Monster = dynamic_cast<cMonster &>(a_Entity);
-				if (Monster.GetMobType() != a_MobType)
-				{
-					return false;
-				}
+		auto & Other = dynamic_cast<cMonster &>(a_Entity);
+		if (Other.GetMobType() != a_MobType)
+		{
+			return false;
+		}
 
-				Vector3d MyHeadPosition = GetPosition().addedY(GetHeight());
-				Vector3d TargetPosition = Monster.GetPosition().addedY(Monster.GetHeight());
-				double TargetDistance = (MyHeadPosition - TargetPosition).SqrLength();
+		Vector3d MyHeadPosition = GetPosition().addedY(GetHeight());
+		Vector3d TargetPosition = Other.GetPosition().addedY(Other.GetHeight());
+		double TargetDistance = (MyHeadPosition - TargetPosition).SqrLength();
 
-				if (TargetDistance < (a_SightDistance * a_SightDistance))
-				{
-					FoundTarget = &Monster;
-					return true;
-				}
-				return false;
-			};
+		if (
+			(MinimumDistance > TargetDistance) &&
+			(TargetDistance < (a_SightDistance * a_SightDistance))
+		)
+		{
+			FoundTarget = & Other;
+			return true;
+		}
+		return false;
+	};
 
-	/* cBoundingBox CheckZone(GetPosition().addedXZ(-a_SightDistance, -a_SightDistance), GetPosition().addedXZ(a_SightDistance, a_SightDistance)); */
-	/* m_World->ForEachEntityInBox(CheckZone, Callback); */
-	m_World->ForEachEntity(Callback);
+	cBoundingBox CheckZone(GetPosition().addedXZ(-a_SightDistance, -a_SightDistance), GetPosition().addedXZ(a_SightDistance, a_SightDistance));
+	m_World->ForEachEntityInBox(CheckZone, Callback);
 	return FoundTarget;
 }
 
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -6,6 +6,7 @@
 #include "../World.h"
 #include "../Entities/Player.h"
 #include "../LineBlockTracer.h"
+#include <random>
 
 
 
@@ -40,6 +41,35 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
 {
 	Super::EventSeePlayer(a_Player, a_Chunk);
 	m_EMState = CHASING;
+}
+
+
+
+
+
+bool cAggressiveMonster::CanSeeMobType(eMonsterType a_MobType, cAggressiveMonster * a_Monster, int a_SightDistance)
+{
+	m_World->ForEachEntity([&](cEntity & a_Entity)
+			{
+			cAggressiveMonster* AggMonster;
+			if ((AggMonster = static_cast<cAggressiveMonster*>(&a_Entity)) == nullptr)
+			{
+				return false;
+			}
+			if (AggMonster->GetMobType() != a_MobType)
+			{
+				return false;
+			}
+			Vector3d MyHeadPosition = GetPosition().addedY(GetHeight());
+			Vector3d TargetPosition = AggMonster->GetPosition().addedY(AggMonster->GetHeight());
+			double TargetDistance = (MyHeadPosition - TargetPosition).SqrLength();
+
+			if (TargetDistance < (a_SightDistance * a_SightDistance))
+			{
+				a_Monster = AggMonster;
+				return true;
+			}
+		});
 }
 
 
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -70,6 +70,7 @@ bool cAggressiveMonster::CanSeeMobType(eMonsterType a_MobType, cAggressiveMonste
 				return true;
 			}
 		});
+		return a_Monster != nullptr;
 }
 
 
--- a/src/Mobs/AggressiveMonster.cpp
+++ b/src/Mobs/AggressiveMonster.cpp
@@ -46,60 +46,40 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
 
 
 
-bool cAggressiveMonster::CanSeeMobType(eMonsterType a_MobType, cAggressiveMonster * a_Monster, int a_SightDistance)
+cMonster * cAggressiveMonster::GetMonsterOfTypeInSight(eMonsterType a_MobType, unsigned int a_SightDistance)
 {
-	a_Monster = nullptr;
-	cBoundingBox CheckZone(GetPosition().addedXZ(-a_SightDistance, -a_SightDistance), GetPosition().addedXZ(a_SightDistance, a_SightDistance));
 
-	class cLineBlockTracerCallbacks : public cLineBlockTracer::cCallbacks
-	{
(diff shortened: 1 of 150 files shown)


src/Mobs/Endermite.h
```diff
@@ -0,0 +1,20 @@
...
+class cEndermite:
+	public cAggressiveMonster
+{
+	using Super = cAggressiveMonster;
+
+public:
+
+	cEndermite();
+
+	CLASS_PROTODEF(cSilverfish)
```
12xx12, line 19:
```suggestion
	CLASS_PROTODEF(cEndermite)
```

src/Protocol/Protocol_1_8.cpp
```diff
@@ -3762,7 +3762,6 @@ void cProtocol_1_8_0::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_M
 
 		case mtCat:
 
-		case mtEndermite:
```
12xx12, line 3765: Also check the other protocol classes. The same structure should also be there

12xx12 (commented):
Just a quick comment and a big thank you: **THANK YOU**

tonitch:
@12xx12 I was wondering about my last todo (aka:  Enderman attack Endermite within 64 b)
a Monster can target another Entity with set target. and there is already fonctions to check if the player is near to go aggro on him.
But I can't find one for other mobs.

Should I make a new one or is there another way?
My intuition was to just make a check in tick of enderman but this seems really ineficient....

I think that doing a new method in Pawn would be usefull for enderman that target endermite but also for irongolem that have to target zombie (they don't atm) and even piglin with hoglin, ...

Just need a little nudge here haha

12xx12:
For checking where to place the check for Enderman targeting endermites:

Since there is some code for entities targeting other entities (mainly the player) I would place the check in the same place where those checks are done (see `cAgggresiveMonster`)


I would add a method 
`bool GetClosestEntityOfType(Type, cEntity & a_Entity)` and return true if one was found and put it into a_Entity. Else return false. Maybe add another parameter for range defaulting to 16 since this seems the default mostly.

tonitch:
> Since there is some code for entities targeting other entities (mainly the player) I would place the check in the same place where those checks are done (see `cAgggresiveMonster`)

Yhea seems like the way to go, I tough it wouldn't be good for Irongolem as they are not aggressive monster. but indeed PassiveAggressive monster is sub class of agressive monster so good to go haha ^^ Thanks for that!

12xx12:
For the special case monsters you could overwrite the method where the close monsters are checked and still call the base method of the parent class

tonitch:
pretty much all my todolist is done but I have one slight problem...

The enderman doesn't get any target.

https://github.com/tonitch/cuberite/blob/203bd2cea3afbd49a5d3efca714752bcf23cbef2/src/Mobs/Enderman.cpp#L152-L160

Here the enderman should get the target adresse in the variable target. but as soon as the CanSeeMobType() is terminated. the target ptr is set to 0x00. I checked with gdb, the adress is set before the end of the said function but when it exit, it is not anymore...

Honestly it seems like a trivial bug but I can't seem to figure out how I should do this one... (I will feel dumb afterward I know) can I have a little help here ?

12xx12:
Are you developing on Windows? The MVSC is very forgiving. I switched to the WSL since there are more errors detected.

tonitch:
> Are you developing on Windows? The MVSC is very forgiving. I switched to the WSL since there are more errors detected.

No Arch linux here, I use vim as an ide and it's youcompleteme that give me the error ^^
The actual code is the one that doesn't give me any but that doesn't work... I posted on discord my other attempts but It won't do because cAggressiveMonster can't be assigned with = because of cPawn wich have a operator= = delete ^^

12xx12:
I think there was some weird behaviour if you want to grab a value in a lambda you had to store the lambda in a variable and call it instead of writing it in place

12xx12:
And if you want the loop to continue in the lambda you need to return false (or 0 like in the main method).

12xx12:
I have no idea why the value of a_Monster is always nullptr. I tried assigning the target similar to the target player function (no return value, and process the targeting inside the function. Maybe the same problem occurred there.

src/Mobs/AggressiveMonster.cpp
```diff
@@ -46,6 +47,36 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
 
 
 
+bool cAggressiveMonster::CanSeeMobType(eMonsterType a_MobType, cAggressiveMonster * a_Monster, int a_SightDistance)
+{
+	m_World->ForEachEntity([&](cEntity & a_Entity)
```
12xx12, line 52:
```suggestion
	m_World->ForEachEntityInBox([&](cEntity & a_Entity, ...)
```

Maybe not check every entity. This might lead to lags if multiple locations are loaded. The box should be around the checked entity and have a radius of a_SightDistance

src/Mobs/AggressiveMonster.cpp
```diff
@@ -46,6 +47,36 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
 
 
 
+bool cAggressiveMonster::CanSeeMobType(eMonsterType a_MobType, cAggressiveMonster * a_Monster, int a_SightDistance)
```
12xx12, line 50:
```suggestion
bool cAggressiveMonster::CanSeeMobType(eMonsterType a_MobType, cAggressiveMonster * a_Monster, unsigned int a_SightDistance)
```

Either add a check that processes negative values or only use unsigned

12xx12:
This is my working code (without my suggested changes:

```c++
a_Monster = nullptr;

	auto Callback = [&](cEntity & a_Entity)
	{
		if (a_Entity.GetEntityType() != etMonster)
		{
			return false;
		}
		auto & Monster = static_cast<cMonster &>(a_Entity);
		if (Monster.GetMobType() != a_MobType)
		{
			return false;
		}
		Vector3d MyHeadPosition = GetPosition().addedY(GetHeight());
		Vector3d TargetPosition = Monster.GetPosition().addedY(Monster.GetHeight());
		double TargetDistance = (MyHeadPosition - TargetPosition).SqrLength();

		if (TargetDistance < (a_SightDistance * a_SightDistance))
		{
			a_Monster = &Monster;
			return false;
		}
		return false;
	};
	m_World->ForEachEntity(Callback);
	if (a_Monster != nullptr)
	{
		SetTarget(a_Monster);
	}
	return (a_Monster != nullptr);
```

tonitch:
> And if you want the loop to continue in the lambda you need to return false (or 0 like in the main method).

I though as the enderman have fond a target, he doesn't have to check for more so we can exit the loop. so I returned true

12xx12:
Intuitively, I would choose the closest not the first found

src/Mobs/AggressiveMonster.h
```diff
@@ -30,6 +30,13 @@ class cAggressiveMonster:
 
 	virtual void EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk) override;
 
+	/**
+	* Check if a a_mobtype is around the monster at a SightDistance
+	* return true if it found one and set a_Monster as the Found Monster
+	* return false and leave the a_Monster untouched if
+	*/
+	virtual bool CanSeeMobType(eMonsterType a_mobtype, cAggressiveMonster * a_Monster, int SightDistance=16);
+
```
madmaxoft, line 39:
```suggestion
	/** Returns a mob of the specified type that is visible to this monster within the specified sight distance.
	If there are multiple such monsters, returns a non-specified (random) one of them. 
	Returns nullptr if there is no such monster. */
	virtual cMonster * GetMonsterOfTypeInSight(eMonsterType a_MobType, double SightDistance=16);

```

madmaxoft (dismissed):

src/Mobs/AggressiveMonster.h
madmaxoft: Also, this might be better implemented directly in the `cMonster` class and probably doesn't need to be `virtual` at all.

src/Mobs/Enderman.cpp
```diff
@@ -149,6 +149,16 @@ void cEnderman::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		return;
 	}
 
+	if (m_EMState != CHASING)
+	{
+		cAggressiveMonster* target = nullptr;
+		if (CanSeeMobType(mtEndermite, target, 64))
```
madmaxoft, line 155:
```suggestion
		auto target = GetMonsterOfTypeInSight(mtEndermite, 64);
		if (target != nullptr)
```

12xx12: Why shouldn't you?

madmaxoft: Generally using `auto` is preferred, it makes later refactoring easier (say that the function changes the return type, you don't have to go through every usage of the function)

tonitch:
could use some testing but seems quite finished. (?!)

tonitch:
well, the remaining error of jenkins ci seems like a default Globals.E_META_SPAWN_EGG_ENDERMITE (constant)
should be in https://api.cuberite.org/_undocumented.lua but I have no idea how I should add it in this list...

12xx12:
This should fix the doc error

src/Mobs/AggressiveMonster.cpp
```diff
@@ -46,6 +47,46 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
...
+				{
+					FoundTarget = &Monster;
+					return true;
+				}
+				return false;
+			};
+
+	/* cBoundingBox CheckZone(GetPosition().addedXZ(-a_SightDistance, -a_SightDistance), GetPosition().addedXZ(a_SightDistance, a_SightDistance)); */
+	/* m_World->ForEachEntityInBox(CheckZone, Callback); */
+	m_World->ForEachEntity(Callback);
```
12xx12, line 82:
```suggestion
	cBoundingBox CheckZone(GetPosition().addedXZ(-a_SightDistance, -a_SightDistance), GetPosition().addedXZ(a_SightDistance, a_SightDistance));
	m_World->ForEachEntityInBox(CheckZone, Callback);
```

This might create some massive lag if any entity is checked

src/Mobs/AggressiveMonster.cpp
```diff
@@ -46,6 +47,46 @@ void cAggressiveMonster::EventSeePlayer(cPlayer * a_Player, cChunk & a_Chunk)
...
+				}
+
+				Vector3d MyHeadPosition = GetPosition().addedY(GetHeight());
+				Vector3d TargetPosition = Monster.GetPosition().addedY(Monster.GetHeight());
+				double TargetDistance = (MyHeadPosition - TargetPosition).SqrLength();
+
+				if (TargetDistance < (a_SightDistance * a_SightDistance))
+				{
+					FoundTarget = &Monster;
+					return true;
```
12xx12, line 75: This will result in the the first found match to be the result. It makes more sense to look for the closest. This seems more intuitive.

And technically this doesn't check if the entity actually does **SEE** the other entity.

12xx12 (changes requested):
We are close to merging. I have a few comments

src/Mobs/Enderman.cpp
```diff
@@ -149,6 +149,29 @@ void cEnderman::Tick(std::chrono::milliseconds a_Dt, cChunk & a_Chunk)
 		return;
 	}
 
+	if (m_EMState != CHASING)
+	{
+		cMonster * EndermiteFound = GetMonsterOfTypeInSight(mtEndermite, 64);
```
bearbin, line 154: This is potentially quite an expensive operation in e.g. a mob farm with lots of endermen about. Not sure we can do anything about it though.

bearbin (approved):
Slight performance concern, otherwise LGTM.

src/Mobs/Enderman.cpp
12xx12: Yeah. The only idea that I have, is to cache the result per chunk. But this is kind of expensive and only an edge case.

bearbin: Alright, let' merge and get it done, we can always revisit if there turns out to be a more serious problem.

Issue #5441, opened 2022-09-25
API: Calling function AddFaceDirection with invalid eBlockFace crashes the server


Seadragon91:
Server OS: Linux
Cuberite Commit id: b73cd1179fedc4abd5c388ae46f507ff3e36c213

### Expected behavior
Calling the function AddFaceDirection  with an invalid eBlockFace  should not crash the server.

### Actual behavior
Server crashes at
https://github.com/cuberite/cuberite/blob/b73cd1179fedc4abd5c388ae46f507ff3e36c213/src/Defines.cpp#L407

### Steps to reproduce the behavior
Call this lua code from a command:
`AddFaceDirection(Vector3i(1, 1, 1), -1, false)`

Pull request https://github.com/cuberite/cuberite/pull/5431/ added this call.

Found by APIFuzzing.

schmarti7777771:
Hi, Id like to take a look at this, I'll see if I can figure out what's causing it.

12xx12:
Hi And welcome to the project. This is a nice starting issue. Please check our code style guide :)

schmarti7777771:
sorry it took me so long to get to this, the problem seems to be that the function AddFaceDirection has no actual return value when it is called with eBlockFace = BLOCK_FACE_NONE. The function is meant to return the coordinates of the block neighboring the block in question through the face specified, so what should we return in this scenario?

madmaxoft:
I personally think the best behavior would be:
- C++ code calling `AddFaceDirection` with an invalid BlockFace should trigger an assert in debug mode and just crash in release mode (it's a bad error for C++ code to call this with bad param).
- Lua code calling the API function `AddFaceDirection` should trigger a Lua error, terminating the current Lua script.
This is somewhat tricky to do - you'd need to make a manual binding for the API function, so that the binding takes care of the parameter validation.

PR #5463, opened 2022-12-05
AddFaceDirection undefined behavior fix

schmarti7777771:
This change is in response to https://github.com/cuberite/cuberite/issues/5441. This will cause the function AddFaceDirection to return the original position when called with an invalid eBlockFace, rather than crashing the server. An alternative solution is proposed by madmaxoft in a comment on the original issue, which is unfortunately not within my ability to implement at this time, so that may be worth considering as well.


12xx12 (changes requested):
This doesn't fix the underlying issue. See #5444 for more info

PR #5468, opened 2023-01-08
Allow certain blocks to be placed on top of upside-down stairs/slabs

havel06:
Pressure plates, repeaters, comparators and redstone wire can now be placed on top of upside-down slabs and stairs.

### DIFF
--- a/src/Blocks/BlockComparator.h
+++ b/src/Blocks/BlockComparator.h
@@ -170,10 +170,12 @@ private:
 		else if (cBlockSlabHandler::IsAnySlabType(BelowBlock))
 		{
 			// Check if the slab is turned up side down
-			if ((BelowBlockMeta & 0x08) == 0x08)
-			{
-				return true;
-			}
+			return (BelowBlockMeta & 0x08) == 0x08;
+		}
+		else if (IsBlockStair(BelowBlock))
+		{
+			// Stair must be upside down
+			return ((BelowBlockMeta & 0x04) == 0x04);
 		}
 		return false;
 	}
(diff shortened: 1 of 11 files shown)


src/Blocks/BlockComparator.h
```diff
@@ -170,10 +170,12 @@ class cBlockComparatorHandler final :
 		else if (cBlockSlabHandler::IsAnySlabType(BelowBlock))
 		{
 			// Check if the slab is turned up side down
-			if ((BelowBlockMeta & 0x08) == 0x08)
-			{
-				return true;
-			}
+			return (BelowBlockMeta & 0x08) == 0x08;
+		}
+		else if (IsBlockStair(BelowBlock))
```
bearbin, line 175: I think this function should be part of the cBlockStairsHandler, like the IsSlabType above.

bearbin (commented):
Looks good, just a comment about the location of IsBlockStair.

Also, could you refactor the torch handler to use your new function as well, it uses custom logic at the moment: https://github.com/cuberite/cuberite/blob/a62b2b1be2103d7de2fd66c7304b7473e369be3c/src/Blocks/BlockTorch.h#L23

I would just remove the switch cases for stairs and add a check to your function (and to slabs) in the default case.

src/Blocks/BlockComparator.h
havel06: True, It would be more consistent with the slab function.

havel06:
I have made changes to buttons, rails and levers, so they can now be placed on top of upside down stairs and slabs as well. Also, torches, levers and buttons can now be placed on the "tall" side of a stair block.

bearbin (approved):
Looking good, I'll wait for another review but this is ready to merge I think.

PR #5469, opened 2023-01-12
TNT minecarts exploding when riding over activator rails

havel06:
TNT minecarts now get fused and explode after riding over powered activator rails.


bearbin (approved):
Looks good, apologies for the delay in reviewing.

PR #5475, opened 2023-03-12
Added code to export definitions for a lua-language-server

NiLSPACE:
I still need to write an article to explain how to configure the language server so this PR is still a work in progress, but the definition generation seems to be working now.

The definitions should work with any IDE that can use [lua-language-server](https://github.com/LuaLS/lua-language-server). Here is an example with Visual Studio Code:

https://user-images.githubusercontent.com/1160867/224548605-7651ff31-938e-4c51-bdcc-c0abfaabed29.mp4


NiLSPACE:
I now see why everything in the APIDump plugin was in a single file ;) It's so the functions don't show up in the global environment. I'll move the functions over to the main file later.

madmaxoft:
Seeing the APIDump code makes me want to rewrite it :)

NiLSPACE:
I had the same urge when I looked at some (or most) of the WorldEdit code yesterday ;)

NiLSPACE:
The 'not polluting _G' issue could also be fixed by immediately replacing `_G` with a empty table with `__index` pointing to the old values.

```lua
local newEnv, oldEnv = {}, _G
local setmetatable = setmetatable
for k, v in pairs(_G) do
	newEnv[k] = v;
	oldEnv[k] = nil;
end
_G = setmetatable(oldEnv, {__index = newEnv});
```

Then instead of looping through `_G` you'd loop through `getmetatable(_G).__index`

madmaxoft:
So is it VSCode-only or generic lua-language-server? Just to get the naming right.

NiLSPACE:
I've only tested it on VSCode, but it should work on any lua-language-server. But you're right, I'll change the naming.

Server/Plugins/APIDump/main_APIDump.lua
```diff
@@ -136,7 +136,7 @@ local function CreateAPITables()
 		return res;
 	end
 
-	for i, v in pairs(_G) do
+	for i, v in pairs(getmetatable(_G).__index) do
 		if (
 			(v ~= _G) and           -- don't want the global namespace
```
madmaxoft, line 141: Will this comparison still work? I doubt that.

NiLSPACE: It actually does still work.

The `_preload` code doesn't replace the actual `_G` table. It just takes its values and moves it to a new table. So once it gets to the `_G` key it takes the reference to itself and moves it to the new table.

Put in another way: Usually `_G` references to itself like this:
```
_G._G._G._G...
```

but with `_preload` this changes to
```
_G.__index._G.__index...
  ▲
   because of getmetatable() we start here which means its _G value 
   references back to the global _G environment
```

bearbin (approved):

PR #5476, opened 2023-03-20
Updated LTO Check to new standard.

12xx12:
The due to LTO the Release build didn't link on my machine. The official cmake page suggest it to do it like this:

 https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html


madmaxoft (approved):
Checked on Linux (clang-14) and Windows (VS2022), both work.

12xx12:
I tested with gcc and I just noticed that I have a Mac and will try this with the Xcode compiler

PR #5478, opened 2023-03-21
Now using size_t in item grids and accompanying classes

12xx12:
Well that escalated quickly. 

This PR takes over #4979. And changes cItemGrid completely to use `std::size_t` to index everything. Since -1 was used to indicate illegal slot stuff, the best alternative seems `std::numeric_limits<std::size_t>::max()`

### DIFF
--- a/src/Bindings/LuaState.cpp
+++ b/src/Bindings/LuaState.cpp
@@ -1099,6 +1099,17 @@ void cLuaState::Push(std::chrono::milliseconds a_Value)
 
 
 
+void cLuaState::Push(std::size_t a_Value)
+{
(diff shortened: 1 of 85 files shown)


madmaxoft:
This breaks plugins, as they expect `-1` slot value in several cases.

12xx12:
(╯°□°）╯︵ ┻━┻

12xx12:
So, how do we deal with this?

- Use a static cast where needed and keep using int.
- Manually change the output of the API from the new constant
- other?

Janmm14:
Why you size_t at all for slots when we know slot index will always fit into a single byte

Issue #4901, opened 2020-09-19
#includes relative to src


tigerw:
Remove any ../../../ in #include directives.

bearbin:
I would much prefer relative to `src` than the current backwardsly-relative includes; the only trouble is going through and updating the existing codebase - that's probably not worth it :(

However I would be opposed to non-namespaced includes where the project directory structure is effectively collapsed.

PR #5480, opened 2023-03-26
Auth and MojangAPI now are using UrlClient

### LINKED ISSUE #2115: SecureRequest: Can't connect to sessionserver.mojang.com: SSL handshake failed: -0x2700
Since this week we are facing that issue.

SecureRequest: Can't connect to sessionserver.mojang.com: SSL handshake failed: -0x2700
Kicking player thessy09 for "Failed to authenticate account!"

Before our MCServer running on a Raspberry worked fine.
I installed that MCServer for my kids, so that are able to play in one world together in the Home-LAN 
I did not change anything. 

Does anyone now how to fix that issue.
Thanks a lot in advance for your support 
br. Michael

12xx12:
- Fixed Login Breaking bug
- Auth and MojangAPI now use UrlClient
- fixed bug in UrlClient where one letter was missing in the HTTP Header
- added function to verify Urls from config files and error handling on bad Urls in config for Auth
- Adds test if the newest release can reach the default auth servers
Fixes #2115


src/HTTP/UrlParser.cpp
```diff
@@ -198,3 +199,19 @@ std::pair<bool, AString> cUrlParser::Parse(
 
 
 
+bool cUrlParser::Validate(const AString & a_Url)
+{
+	// https://stackoverflow.com/questions/5620235/cpp-regular-expression-to-validate-url
+	// I didn't use the first solution deliberately because it didn't enforce the http(s) in the beginning
+
+	const std::regex UrlRegex (
```
madmaxoft, line 207: Do we really want to pull in the whole regex machinery just to check a simple URL? Not to mention that it's possible that this check says OK and the `Parse` method still fails, or vice versa. I'd just call `Parse` and decide upon its return value.

src/Protocol/Authenticator.cpp
```diff
@@ -51,6 +50,24 @@ void cAuthenticator::ReadSettings(cSettingsRepositoryInterface & a_Settings)
 	m_Server             = a_Settings.GetValueSet ("Authentication", "Server", DEFAULT_AUTH_SERVER);
 	m_Address            = a_Settings.GetValueSet ("Authentication", "Address", DEFAULT_AUTH_ADDRESS);
 	m_ShouldAuthenticate = a_Settings.GetValueSetB("Authentication", "Authenticate", true);
+
+	if (!cUrlParser::Validate(m_Server))
+	{
+		LOGWARNING("%s %d: Supplied invalid URL for authentication server: \"%s\", using default!", __FUNCTION__, __LINE__, m_Server.c_str());
```
madmaxoft, line 56: The error message should ideally point to the exact setting that is wrong, to help the users.

src/Protocol/Authenticator.cpp
```diff
@@ -51,6 +50,24 @@ void cAuthenticator::ReadSettings(cSettingsRepositoryInterface & a_Settings)
...
+	{
+		LOGWARNING("%s %d: Supplied invalid URL for authentication address: \"%s\", using default!", __FUNCTION__, __LINE__, m_Server.c_str());
+		m_Address = DEFAULT_AUTH_ADDRESS;
+	}
+
+	// prepend https:// if missing
+	if (m_Server.rfind("http") == AString::npos)
+	{
+		m_Server = "https://" + m_Server;
+	}
```
madmaxoft, line 70: This should be done prior to validation, if needed at all. And it will fail if there's an `http` in the domain name, for example.

src/Protocol/Authenticator.cpp
```diff
@@ -1,19 +1,18 @@
...
+#include "Protocol/Authenticator.h"
+#include "Protocol/MojangAPI.h"
+#include "Root.h"
+#include "Server.h"
+#include "ClientHandle.h"
+#include "UUID.h"
+#include "HTTP/UrlParser.h"
+
+#include "IniFile.h"
+#include "JsonUtils.h"
```
madmaxoft, line 13: What's with the changes in the includes? Keep them relative, as they were.

src/Protocol/MojangAPI.cpp
```diff
@@ -883,15 +694,15 @@ void cMojangAPI::Update(void)
 	}
 	if (!PlayerNames.empty())
 	{
-		LOG("cMojangAPI: Updating name-to-uuid cache for %u names", static_cast<unsigned>(PlayerNames.size()));
+		LOG("cMojangAPI: Updating name-to-uuid cache for %u names", PlayerNames.size());
```
madmaxoft, line 697: Is this safe? I believe `%u` expects `uint32_t`, while `PlayerNames.size()` is `size_t`, which is `uint64_t` on 64-bit platforms.

src/Protocol/MojangAPI.h
```diff
@@ -83,6 +79,34 @@ class cMojangAPI
...
+		{
+			m_ResponseBody.append(static_cast<const char *>(a_Data), a_Size);
+		}
+
+		std::shared_ptr<cEvent> m_Event;
+
+		/** The accumulator for the partial body data, so that OnBodyFinished() can send the entire thing at once. */
+		AString & m_ResponseBody;
+	};
+
```
madmaxoft, line 109: This could probably be pulled into the cpp file into an anonymous namespace.

madmaxoft (changes requested):
Note that this completely disables the TLS security, in the sense that it doesn't check the server certificate anymore. Therefore a trivial MITM attack against this would be successful. Do we want to lower the security just because we can't get it working right now?

src/HTTP/UrlParser.cpp
12xx12: Indeed: The `Parse` function will expect us to provide a protocol. This will work without protocol

src/Protocol/Authenticator.cpp
12xx12: see #4901

12xx12:
Did we just check if the string for the certificate matches, or did we validate the certificates?

madmaxoft:
We did check the certs proper - loaded them as root CAs and used them for checking. The whole issue could probably be solved by updating the CA cert list. Although I like the `cUrlClient` solution. Less code duplication.

12xx12:
The cUrlClient seems the best solution. The old implementation was another HTTP client but stripped down. I am trying to find the place where the certs are actually checked and adding the additional code needed to prevent the MITM attack

12xx12:
Anyone can feel free to contribute.

Next week I don't have much time

madmaxoft:
They may have switched to a different root CA, there are many many CAs out there. We only had the specific root CAs they were using at the time.

src/Protocol/Authenticator.cpp
madmaxoft: Oh well, okay for now. But I will get that decision reversed! :D

src/HTTP/UrlParser.cpp
12xx12: My idea was, because we need to supply a bunch of String objects for the `Parse` method. This is just a simple take the string and tell me.

src/Protocol/MojangAPI.h
12xx12: I'd like to have access to this in other classes. Maybe add this to `cUrlClient` as default or basic

src/HTTP/UrlParser.cpp
madmaxoft: It is simple, but it needn't match whatever `Parse` is expecting. My point is, keep this function, but rather than using a regex, call `Parse` from here, discarding its outputs other than the error flag. That way you know that whatever this accepts, `Parse` will accept, too.

src/Protocol/Authenticator.cpp
```diff
@@ -51,6 +50,24 @@ void cAuthenticator::ReadSettings(cSettingsRepositoryInterface & a_Settings)
 	m_Server             = a_Settings.GetValueSet ("Authentication", "Server", DEFAULT_AUTH_SERVER);
 	m_Address            = a_Settings.GetValueSet ("Authentication", "Address", DEFAULT_AUTH_ADDRESS);
 	m_ShouldAuthenticate = a_Settings.GetValueSetB("Authentication", "Authenticate", true);
+
+	// prepend https:// if missing
+	if (m_Server.rfind("http") == AString::npos)
+	{
+		m_Server = "https://" + m_Server;
+	}
```
madmaxoft, line 58: I still think `rfind` is not the right one. We want "If it doesn't start with `http://` or `https://`, then prepend `https://`"

src/Protocol/MojangAPI.h
madmaxoft: That's actually a good idea, move this completely to `cUrlClient` as `BlockingGet()` and `BlockingPost()` etc. methods. Just don't export them to Lua! :D

12xx12:
The only problem is now in the case that the certificate validation fails the action just stops and doesn't return

12xx12:
Can someone else test this? I am not sure if I am blacklisted or I made a mistake. The other side sends a disconnect

NiLSPACE:
> Can someone else test this? I am not sure if I am blacklisted or I made a mistake. The other side sends a disconnect

I just compiled and ran it. I can join without any problems.

12xx12:
Yeah the problem is somewhere in the certificates. When disabling actually checking the certificates this fails :(

src/Protocol/MojangAPI.cpp
```diff
@@ -251,6 +139,37 @@ class cMojangAPI::cUpdateThread:
...
+	void OnBodyData(const void * a_Data, size_t a_Size) override
+	{
+		m_ResponseBody.append(static_cast<const char *>(a_Data), a_Size);
+	}
+
+	std::shared_ptr<cEvent> m_Event;
+
+	/** The accumulator for the partial body data, so that OnBodyFinished() can send the entire thing at once. */
+	AString & m_ResponseBody;
+};
```
12xx12, line 168: Duplicate Code!!

src/mbedTLS++/SslConfig.cpp
```diff
@@ -225,7 +225,6 @@ void cSslConfig::SetCACerts(cX509CertPtr a_CACert)
 
 std::shared_ptr<cSslConfig> cSslConfig::MakeDefaultConfig(bool a_IsClient)
 {
-	// TODO: Default CA chain and SetAuthMode(eSslAuthMode::Required)
```
12xx12, line 228: Todo: Read Todo

src/mbedTLS++/SslConfig.cpp
```diff
@@ -236,7 +235,14 @@ std::shared_ptr<cSslConfig> cSslConfig::MakeDefaultConfig(bool a_IsClient)
 		Ret->SetRng(std::move(CtrDrbg));
 	}
 
-	Ret->SetAuthMode(eSslAuthMode::None);  // We cannot verify because we don't have a CA chain
```
12xx12, line 239: Todo: Read comment

12xx12:
So I der some experiments:

The Certificate Serial Number displayed when enabling SSL Debugging deviates from the one I grabbed from visiting https://api.mojang.com.

The Serial Number Diplayed is: 0D:7B:ED:E9:7D:82:09:96:7A:52:63:1B:8B:DD:18:BD

12xx12:
I just got the certificate displayed in the SSL Dubugger and now it works :shrug:

This isn't a pure root CA just a intermediate certificate to sign other certs.

12xx12:
Hm. What if Mojang sends different certs depending on the client type. Time to spam the endpoint

madmaxoft:
There's no way for the server to tell the client kind before sending the cert. The TLS handshake comes before any client-identifying information (which is good from the security standpoint). The messages exchanged within the TLS handshake are:
1. C->S: ClientHello ("I can do ciphersuites X, Y and Z, compression methods M, N, use this random value, and here's some extensions (e.g. I want to talk to api.mojang.com))
2. S->C: ServerHello ("Here's my certificate", among other things)
3. ...

The only two reasonable explanations I can come up with are:
1. Loadbalancing: each DNS query can return a different IP for the same name, thus redirecting to multiple backend servers
2. IPv4 vs IPv6 going to different servers.

Yeah, TLS is hard. It has to, in order to be secure.

madmaxoft:
Btw, use `openssl s_client` to check the cert reported by the site. With `-debug -msg` params it can even dump the certs in PEM format.

12xx12:
Yeah... I Got It :( 

For some reason Digicert Didn't sign their certificate which witch they signed the Microsoft Certificates with their own Root Certificate. :shrug: (That certificate was included) But used another Certificate. I got that one from the DigiCert Website and now I can login

12xx12:
> Btw, use `openssl s_client` to check the cert reported by the site. With `-debug -msg` params it can even dump the certs in PEM format.

I saw that on my research

12xx12:
> Yeah, TLS is hard. It has to, in order to be secure.

Well actually not that bad. You only have to know once. And the math of public key stuff is nice

src/Protocol/Authenticator.cpp
```diff
@@ -51,6 +50,36 @@ void cAuthenticator::ReadSettings(cSettingsRepositoryInterface & a_Settings)
...
+	if (
+		(std::string_view(m_Server).substr(0, HttpPrefix.size()) != HttpPrefix) &&
+		(std::string_view(m_Server).substr(0, HttpsPrefix.size()) != HttpsPrefix)
+	)
+	{
+		m_Server = "https://" + m_Server;
+	}
+
+	{
+		auto [Success, ErrorMessage] = cUrlParser::Validate(m_Server);
```
madmaxoft, line 67: Naming: bool values should have a `Is`, `Has`, `Should` etc. prefix. So in this case, `IsSuccess`.

src/Protocol/Authenticator.cpp
```diff
@@ -152,38 +181,12 @@ bool cAuthenticator::AuthWithYggdrasil(AString & a_UserName, const AString & a_S
...
-	{
-		return false;
-	}
-
-	// Check the HTTP status line:
-	const AString Prefix("HTTP/1.1 200 OK");
-	AString HexDump;
-	if (Response.compare(0, Prefix.size(), Prefix))
+	// Create and send the HTTP request
+	auto [Success, Response] = cUrlClient::BlockingGet(m_Server + ActualAddress);
```
madmaxoft, line 185: `IsSuccess`

src/Protocol/MojangAPI.cpp
```diff
@@ -650,43 +484,13 @@ void cMojangAPI::QueryNamesToUUIDs(AStringVector & a_NamesToQuery)
...
-		Request += "Content-Type: application/json\r\n";
-		Request += Printf("Content-Length: %u\r\n", static_cast<unsigned>(RequestBody.length()));
-		Request += "\r\n";
-		Request += RequestBody;
-
-		// Get the response from the server:
-		AString Response;
-		if (!SecureRequest(m_NameToUUIDServer, Request, Response))
+		// Create and send the HTTP request
+		auto [Success, Response] = cUrlClient::BlockingPost(m_NameToUUIDAddress, AStringMap(), std::move(RequestBody), AStringMap());
```
madmaxoft, line 488: `IsSuccess`

src/Protocol/MojangAPI.cpp
```diff
@@ -762,45 +566,16 @@ void cMojangAPI::QueryUUIDToProfile(const cUUID & a_UUID)
...
-		LOGINFO("%s failed: bad HTTP status line received", __FUNCTION__);
-		LOGD("Response: \n%s", CreateHexDump(HexDump, Response.data(), Response.size(), 16).c_str());
-		return;
-	}
-
-	// Erase the HTTP headers from the response:
-	size_t idxHeadersEnd = Response.find("\r\n\r\n");
-	if (idxHeadersEnd == AString::npos)
+	// Create and send the HTTP request
+	auto [Success, Response] = cUrlClient::BlockingGet(Address);
```
madmaxoft, line 570: `IsSuccess`

src/Protocol/MojangAPI.cpp
```diff
@@ -762,45 +566,16 @@ void cMojangAPI::QueryUUIDToProfile(const cUUID & a_UUID)
...
-		LOGINFO("%s failed: bad HTTP response header received", __FUNCTION__);
-		LOGD("Response: \n%s", CreateHexDump(HexDump, Response.data(), Response.size(), 16).c_str());
 		return;
 	}
-	Response.erase(0, idxHeadersEnd + 4);
 
 	// Parse the returned string into Json:
 	Json::Value root;
-	AString ParseError;
+	AString ParseError, HexDump;
```
madmaxoft, line 578: Seems like this is not used anymore, debugging leftovers?

src/Protocol/MojangAPI.h
```diff
@@ -11,7 +11,8 @@
 
 #include <time.h>
 
-#include "../UUID.h"
+#include "HTTP/UrlClient.h"
```
madmaxoft, line 14: Is the UrlClient needed by the header? I'd say only cpp.

12xx12:
> 

We could add more Tests for some domains to keep the certs up to date

src/HTTP/UrlClient.cpp
```diff
@@ -677,3 +715,81 @@ std::pair<bool, AString> cUrlClient::Put(
...
+	{
+		EvtFinished->Wait();
+	}
+	else
+	{
+		LOGWARNING("%s: HTTP error: %s", __FUNCTION__, ErrorMessage.c_str());
+		return std::make_pair(false, AString());
+	}
+	return std::make_pair(true, Response);
+}
```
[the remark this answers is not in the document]
12xx12: I would still call `Request` in the end

madmaxoft:
Unless we include the whole root CA list, for example from Mozilla, there's always gonna be false negatives. I think a better way around this is to allow the API to specify whether to check the root CAs at all, and if so, what CAs to use. That's what I was planning to add later on.

src/HTTP/UrlClient.cpp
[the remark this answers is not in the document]
madmaxoft: Yes, but this should make the code smaller, right now there's a lot of repetition.

src/Protocol/MojangAPI.cpp
madmaxoft: Right, I saw no changes below this line, so I thought, why was it here. Didn't notice the variable used to be declared above, which you removed. Sorry about that.

steve-nzr:
Thanks for the MR, it was necessary to fix this!

We observed a few things, I'll try to keep it short. It is focused on **sessionserver.mojang.com** only. I'd like to thank @pastequo that helped & guided me a lot on this!

#### Exploring sessionserver.mojang.com certificate using openssl client

`openssl s_client -showcerts -connect sessionserver.mojang.com:443` shows us this :

- sessionserver.mojang.com certificate was signed by "_GeoTrust Global TLS RSA4096 SHA256 2022 CA1_".
- and "_GeoTrust Global TLS RSA4096 SHA256 2022 CA1_" is in an intermediate certificate and was itself signed by "_DigiCert Global Root CA_".

_DigiCert Global Root CA_ is the root certificate, and also the one you put in GetCACerts() (the first one). It is present on my machine (Mac) as a trusted root CA. Great.

### The code gets another certificate in return during the handshake

So the code should only use this root CA right ? ... right ? Well, no. The code receives another certificate ! A wildcard certificate from microsoft (CN=*.azureedge.net), signed by "_DigiCert Global Root G2_", the second one you put in GetCACerts(). This sounds wrong, and this particular certificate is, for example, not present on my machine.

Running `nslookup sessionserver.mojang.com` returns two entries, and for me it's : _part-0014.t-0009.fdv2-t-msedge.net_. I bet this varies according to your location. Running `openssl s_client -showcerts -connect part-0014.t-0009.fdv2-t-msedge.net:443` now returns the same certificate as the one we see in cuberite ! Interesting.

#### Why are we receiving a Microsoft certificate instead of the sessionserver.mojang.com one ?

This has to do with something called SNI. While openssl specifies the hostname expected from the server, as thus received the right certificate, the "mojang" one, the code received another one, and it's the one from the CNAME DNS lookup, the one from _part-0014.t-0009.fdv2-t-msedge.net_.

Now if I run again my openssl command against the Microsoft server (the one I'm redirected to in Cuberite), but specifying I want to reach the mojang hostname using SNI, its like `openssl s_client -showcerts -servername sessionserver.mojang.com -connect part-0014.t-0009.fdv2-t-msedge.net:443`, it returns the right certificate, the first one I've got using openssl against the sessionsserver domain.

#### The fix

Well, I don't know! We didn't manage to properly set the SNI parameter on mbedtls, this seemed to be the right code `mbedtls_ssl_set_hostname(&m_Ssl, "sessionserver.mojang.com");` but it just makes the TLS verification fail, and maybe it's the right behavior since we receive the wrong certificate.

Here, we want to set the hostname using the SNI TLS extension to _sessionserver.mojang.com_.

----

I'd also say that we shouldn't maintain a list of trusted cert but rather use the one from our machines which are maintaned etc... I was able to make the auth work by only using the G2 cert from your MR (and removing the other 3).

And last but not least, we probably shouldn't trust the G2 certificate since it's the wrong one.

Hope it helps!

12xx12:
Thank you for your extensive research! 

Windows does include the G2 certificate (obviously duuuuh)

I added your solution to my PR

12xx12:
Hm… The Jenkins build is stuck after the tests. But all tests go through

12xx12:
@bearbin
Sorry for pinging you again. The client for the Jenkins build seems to have disconnected during the build

bearbin:
Sorry about that, I've restarted the build, hopefully it'll pass this time.

bearbin:
The issue isn't with the buildserver, it's with the test, the UrlClient test is hanging. Do any of the changes in this PR conflict with the test somehow?

12xx12:
I Updates the test. But on machines (plural is intentional) they go through

12xx12:
Is there a way to see the output to stdout to verify which test doesn't finish?

I just added tests to connect to cuberite and the Mojang auth. servers

12xx12:
I undid the changes to the test. Let's see what happens

bearbin:
I'm not sure how to get the logs out of the test framework, but it's one of those three that's causing the issue, since it now passes...

12xx12:
I'll trial and error which test actually stops working.

I really would like to include this since, if the connection to the auth. server fail for any reason we need to handle this probably mostly certs.

Maybe one of the domains isn't in the DNS server

madmaxoft:
We need to include at least one more root CA, for GitHub, otherwise our own WorldEdit plugin will break:
https://github.com/cuberite/WorldEdit/blob/631b8b9b6d8a966c53f99d2631981197769909d1/Classes/Updater.lua#L27

madmaxoft:
There's a few more hits on the GitHub search, so this can possibly affect even more people.
- https://github.com/Rorkh/gt67r4
- https://github.com/KrystilizeNevaDies/apt-cuberite
- several Cuberite forks with some stuff added to them.

12xx12:
The GitHub website is in the tests for the cUrlClient. So this should work

madmaxoft:
Replaced by #5494 .

PR #5481, opened 2023-03-26
Add Class and Function Identifier Macro

12xx12:
This adds a Macro which simply identifies the function and class name for error messages. Examples taken from the MojangAPI class [Chosen because I worked on that]

Processing is done **during compile time**. Should be cross compatible.

```__METHOD_NAME__``` --> ```cClassName::FunctionName()```

The ```__LINE__``` macro would be nice to have, but I didn't find a simple way to do this at compile time.

### DIFF
--- a/src/Protocol/MojangAPI.cpp
+++ b/src/Protocol/MojangAPI.cpp
@@ -5,17 +5,18 @@
 
 #include "Globals.h"
 #include "MojangAPI.h"
+
+#include "HTTP/UrlClient.h"
+#include "IniFile.h"
+#include "JsonUtils.h"
+#include "json/json.h"
+#include "mbedTLS++/BlockingSslClientSocket.h"
+#include "mbedTLS++/SslConfig.h"
+#include "OSSupport/IsThread.h"
+#include "RankManager.h"
+#include "Root.h"
 #include "SQLiteCpp/Database.h"
 #include "SQLiteCpp/Statement.h"
-#include "../IniFile.h"
-#include "../JsonUtils.h"
-#include "json/json.h"
-#include "../mbedTLS++/BlockingSslClientSocket.h"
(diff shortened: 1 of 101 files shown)

### CODE REVIEW

File: src/Protocol/MojangAPI.cpp [hunk 1]
```diff
@@ -440,13 +440,13 @@ bool cMojangAPI::SecureRequest(const AString & a_ServerName, const AString & a_R
 	Socket.SetExpectedPeerName(a_ServerName);
 	if (!Socket.Connect(a_ServerName, 443))
 	{
-		LOGWARNING("%s: Can't connect to %s: %s", __FUNCTION__, a_ServerName.c_str(), Socket.GetLastErrorText().c_str());
+		LOGWARNING("%s: Can't connect to %s: %s", __METHOD_NAME__, a_ServerName.c_str(), Socket.GetLastErrorText().c_str());
```
madmaxoft, line 443: I think this will not work properly, `LOGWARNING` expects a `char *`, but `__METHOD_NAME__` provides a `std::string_view`

12xx12: With gcc this works: 

`Log: [1f9f141165174f7c|22:16:56] cRoot::Run, Startup complete, took 3036ms!
`

check at a place which is defenetly run

madmaxoft: Works, but is undefined behavior, so...

madmaxoft: I'll check with MSVC, just for the fun.

madmaxoft: Oh, `LOGWARNING` et al. have been converted to `fmt` already, so despite accepting printf-like format strings, they work with `std::string` etc.
Right, my bad. Sorry.

madmaxoft (approved):

PR #5482, opened 2023-03-30
PieceGenerator: Added functionality to end a structure with a custom prefab

NiLSPACE:
Currently the PieceGenerator structures can end rather abruptly. This PR makes it possible to specify a prefab that will be used when the end is reached. The prefab will not be placed if it intersects with existing prefabs though. For villages I hope this would make it possible to create village entrances/gates which I think could look cool. 

I need some more feedback though. I seem to be missing one exit condition as on rare occasions the closure prefab is not placed. I'd check on the Gallery server if it's perhaps because I've got a connector type wrong, but I can't reach it.

### DIFF
--- a/Server/Prefabs/PieceStructures/NetherFort.cubeset
+++ b/Server/Prefabs/PieceStructures/NetherFort.cubeset
@@ -1773,7 +1773,124 @@ Cubeset =
 			},
 		},  -- BlazePlatformOverhang
 
-
+		{
+			OriginData = 
+			{
+				ExportName = "ClosurePiece",
+			},
+			Size =
+			{
+				x = 5,
+				y = 2,
+				z = 1
+			},
+			Hitbox =
+			{
+				MinX = 0,
+				MinY = 0,
+				MinZ = 0,
+				MaxX = 4,
+				MaxY = 2,
+				MaxZ = 0,
+			},
+			Connectors =
+			{
+				{
+					Type = 0,
+					RelX = 2,
+					RelY = 1,
+					RelZ = 0,
+					Direction = 2,
+				}
+			},
+			Metadata =
+			{
+				["AddWeightIfSame"] = "0",
+				["AllowedRotations"] = "7",
+				["DefaultWeight"] = "100",
+				["DepthWeight"] = "",
+				["ExpandFloorStrategy"] = "none",
+				["IsClosurePiece"] = "1",
+				["MergeStrategy"] = "msSpongePrint",
+				["MoveToGround"] = "0",
+				["VerticalStrategy"] = "Range|40|100",
+			},
+			BlockDefinitions =
+			{
+				".:112:0",
+				"a:113:0",
+			},
+			BlockData =
+			{
+				".....",
+				".aaa.",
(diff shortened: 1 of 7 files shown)


madmaxoft:
Wouldn't it be easier not to define a separate set of closure pieces, but rather as regular pieces that can be placed only from certain (tree) depth and have only one connector (thus no more pieces will follow them)? No code change required for that.

NiLSPACE:
> Wouldn't it be easier not to define a separate set of closure pieces, but rather as regular pieces that can be placed only from certain (tree) depth and have only one connector (thus no more pieces will follow them)? No code change required for that.

I did consider that, but as I was going through the cubeset files I noticed that not allot of prefabs only have a single connector. `BridgeCrumble2` for example has 2 connectors at the same coordinates. Though now that I think of it you can just drop the connectors if they are at the same place or have different types.

> Sorry about the Gallery server, we had a power outage and I forgot to restart the server. On the other hand, it's probably unreachable by a vanilla client anyway, since it still suffers from the SSL handshake failure (and can't download new versions anyway since download.cuberite.org seems to have an expired cert)

The webadmin still has a nice interface to inspect the connectors :)

madmaxoft:
The double connectors are there for a reason - each connector N connects to negative-N (so that one-way connections can be made), but if the connection is to be two-way, it needs both connector types. Still, once one of the connector is used, the other is blocked so effectively it still is an end-piece.

NiLSPACE:
Aah ok that makes sense. Looks like we weren't that consistent with it though. BridgeSegment (Nether 16) has only 2 connectors, both type 0 and facing the opposite direction.

madmaxoft:
`0` == `-0`

NiLSPACE:
Looks like we added too many sponges. Occasionally it looked like it didn't finish with a single connector piece, but because of the sponges you just couldn't see it from inside.

NiLSPACE:
If we ever make a Cubeset v2 variant it might be useful to allow the hitbox of a prefab to consist out of multiple cuboids. This currently can be a problem if the shape of a prefab isn't really square. For example the BalconeyTee2 prefab has a hitbox (highlighted with sponges) which is stopping a closure piece from being placed.
2023-03-31_13 21 03 - Copy

NiLSPACE:
I also noticed some pieces have an additional 10 on their Y+ hitbox which sometimes stops a closure piece to be placed.

madmaxoft:
I still think this entire thing is needless, you can achieve the same result using the pieces' weights. This needlessly complicates things, as well as breaks some "promises", such as "max depth".

NiLSPACE:
If the same can be done using the pieces' weights then I guess we can close this. I thought using weights didn't guarantee that a piece would always be placed at the end.

madmaxoft:
Well, there's no guarantee, there's always "best effort" only. But I believe it is the same with your code - it also could end up with the end pieces not placed due to some constraints. So I consider them pretty much equal in what result they achieve.

NiLSPACE:
Alright, then I'll close this PR. If the need ever arises we can always revive it.

PR #5484, opened 2023-04-02
Implement slime chunks.

### LINKED ISSUE #5130: Slimes don't have designated chunks
Client version: x.x.x

havel06:
Fixes #5130.
The slime chunk algorithm isn't compatible with the original Java version (due to copyright reasons), but the chance should be the same.

### DIFF
--- a/Server/Plugins/APIDump/Classes/World.lua
+++ b/Server/Plugins/APIDump/Classes/World.lua
@@ -2315,6 +2315,27 @@ function OnAllChunksAvailable()</pre> All return values from the callbacks are i
 				},
 				Notes = "Returns whether or not saving chunk data is enabled. If disabled, the world will keep dirty chunks in memory forever, and will simply regenerate non-dirty chunks that are unloaded.",
 			},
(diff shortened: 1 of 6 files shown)


NiLSPACE:
Thanks! Any chance you could add a `cWorld:IsSlimeChunk(a_ChunkX, a_ChunkZ)` function and export it to the Lua API?

I'm not really sure what would be better though. Have the `cWorld:IsSlimeChunk` function call `cChunk:IsSlimeChunk` or the other way around.

havel06:
Absolutely! I think that it would be easier to move the code into `cWorld` and let `cChunk::IsSlimeChunk` just be a proxy.

NiLSPACE:
It now fails because the new there is no documentation for the function in the Lua API. This can be added here: https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/Classes/World.lua#L2318

NiLSPACE (approved):

12xx12 (approved):

PR #5486, opened 2023-04-07
byte buffer refactor

12xx12:
- Using smart pointer for allocation
- Adding constants to the decoding for protocol values
- minor copying and pasting stuff

I tried to maintain readability. I thought about using `sizeof()` to define the read/write byte count, but this seems a bit overkill and may lower readability.


12xx12:
Test failing

src/ByteBuffer.cpp
```diff
@@ -416,15 +458,15 @@ bool cByteBuffer::ReadVarInt32(UInt32 & a_Value)
 	CHECK_THREAD
 	CheckValid();
 	UInt32 Value = 0;
-	int Shift = 0;
-	unsigned char b = 0;
+	std::size_t Shift = 0;
+	unsigned char ReadByte = 0;
```
madmaxoft, line 462: I think I prefer the simple `b` here, since English has the unfortunate ambiguity in `Read` (past tense vs imperative). `ReadByte` sounds like `doReadByte` to me, rather than `theByteThatWasRead`

MightyFilipns:
Can someone restart appveyor so it will finish and auto merge this?

MightyFilipns:
Actually, appveyor completed but never reported but the result for some reason
[Link to the build](https://ci.appveyor.com/project/Cuberite/cuberite/builds/48142101)

12xx12:
Done

12xx12:
now its working

madmaxoft (approved):

PR #5490, opened 2023-04-09
Block Handler include processing for block states

12xx12:
- Block types with multiple instances (slabs, fences, ...) have unified methods to find out about their state
- renamed crafting table and wool block handler
- Blocks which can be rotated contain code to rotate the block state

This is a part of the Block State PR. I try to split up my 31k line PR into more sensible parts.

In the end, this just enlarges the executable with stale code xD


12xx12:
newly undocumented symbols

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -14877,6 +14877,40 @@ end
...
+							Type = "eBlockFace",
+						},
+					},
+					Returns =
+					{
+						{
+							Type = "eBlockFace",
+						},
+					},
+					Notes = "Returns the {{Globals#BlockFaces|eBlockFace}} that corresponds to the given {{Globals#BlockFaces|eBlockFace}} after mirroring it around the XY plane (or rotating 180 degrees around it).",
```
madmaxoft, line 14895: So is it mirroring or rotating? In the general sense, those two operations are different, though for Minecraft they may be the same?

12xx12: Indeed, this comment is wrong.

For most blocks, this applies. But if you want to rotate blocks with the ability to be asymmetric you don't have 180 degree rotation but mirroring. For examples vines.

Example viewed from the top:
`_|` should mirror to `|_` but rotation does yield `|^`. Imagine the `^` is a horizontal line at the top.

PR #5493, opened 2023-04-11
WSSAnvil Revamp for multiple versions

12xx12:
# Goal

This PR enables us to support different versions of Minecraft save data. Since the vanilla encodes this is a Data Version variable, we always can query the save file for this. Until now, the loading is not very organized and mostly up to date with 1.12.

# What i did

- added enum for data version
- added parsing
- added (de)serializer for enums
- moved actual deserialization to extra classes
- new structure is similar to cProtocol and descendants

# Todo
- [ ] Check if the implementation matches 1.8; if not, move to other class. See the methods of `cBoat` and `cEndCrystal` in 1.8

# Future
- Add storing  for other version
- Add support for DataVersion in `cNBTChunkSerializer`


12xx12:
I broke something. The .mca files are 4GB :o

12xx12:
I am trying to finish this PR next.

PR #5497, opened 2023-05-11
Removed all Printf-family functions from StringUtils.

madmaxoft:
Replaced them with `fmt::format` calls, including changes to the format strings. Also changed the format strings to use `FMT_STRING`, so that the format is checked compile-time against the arguments. 

Had to also fix code-style violations already present in the code.


madmaxoft:
Going through all the format strings shouldn't be needed, they should be at least somewhat correct (the number and type of params is checked by the `FMT_STRING` macro).

NiLSPACE (approved):

PR #5499, opened 2023-05-19
Added a warning if the webadmin is enabled without any users.

NiLSPACE:


12xx12 (approved):

src/WebAdmin.cpp
```diff
@@ -190,6 +190,13 @@ void cWebAdmin::Reload(void)
 			" This will not take effect until you restart the server."
 		);
 	}
+	else if (!HasUsers())
+	{
+		LOGWARNING(
+			"The webadmin is enabled but has no users configured."
+			" To add new users see webadmin.ini"
```
madmaxoft, line 197:
```suggestion
			" To add new users, edit webadmin.ini"
```

src/WebAdmin.cpp
```diff
@@ -248,6 +255,23 @@ bool cWebAdmin::LoadIniFile(void)
 
 
 
+bool cWebAdmin::HasUsers(void)
```
madmaxoft, line 258:
```suggestion
bool cWebAdmin::HasUsers()
```
`(void)` is C, `()` is C++ :)

madmaxoft (changes requested):

src/WebAdmin.cpp
NiLSPACE: There are a few places where `(void)` is still used. Could be another `effort/easy` issue.

madmaxoft (approved):

src/WebAdmin.cpp
madmaxoft: No, the point of such easy tasks is not to offload simple mechanical work to "interns", but instead to motivate people to join. This proposition is not motivational.

PR #5501, opened 2023-05-19
Only enable werror for debug

### LINKED ISSUE #5500: Compiling release fails on clang 16
There are lots of warnings emitted when using clang-16. We should consider disabling `Werror` for release builds to ensure they compile.

KingCol13:
Fixes #5500

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -152,9 +152,6 @@ function(set_exe_flags TARGET)
 		target_compile_options(
 			${TARGET} PRIVATE
 
-			# Warnings-as-errors only on Clang for now:
-			-Werror
-
 			# TODO: actually fix the warnings instead of disabling them
 			# or at least disable on a file-level basis:
 			-Wno-missing-noreturn -Wno-padded -Wno-implicit-fallthrough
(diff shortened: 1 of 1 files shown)


SetFlags.cmake
```diff
@@ -152,9 +152,6 @@ function(set_exe_flags TARGET)
 		target_compile_options(
 			${TARGET} PRIVATE
 
-			# Warnings-as-errors only on Clang for now:
-			-Werror
-
 			# TODO: actually fix the warnings instead of disabling them
```
12xx12, line 155: *cough*

bearbin:
Not sure I agree with the premise of this change, why would we want to do less checking on release builds as compared to debug?

KingCol13:
This is somewhat subjective, here's my reasoning:

- Compiler authors are free to implement arbitrarily stringent warnings
- We should treat release builds as being built my users
- Compilation by a user of a correct program should not fail just because they have a newer compiler with new warnings

Currently I think the easiest way to ensure the build completes with the current cmake is to pass in `CFLAGS="-Wno-everything" CXXFLAGS="-Wno-everything"`, turning off warnings entirely which isn't ideal.

Would you prefer adding a build flag, enabled by default to disable `Werror`, which we then disable in `compile.sh`?

I think we should make it as easy possible for new users compiling to get a working binary.

bearbin (approved):
Alright, this makes sense for now. Ideally we should be enabling errors in release mode as well during development (the code compiled in release mode is not necessarily the same as that compiled in debug); but this definitely does fix a pain point for users having brittle builds.

Issue #5495, opened 2023-05-10
GZipFile: Remove on-stack allocation


madmaxoft:
Currently `GZipFile::ReadRestOfFile()` allocates the result on stack, resulting in 128 KiB stack usage. VS has a soft-complaint about it.

We should change the `Compression::Result` type to always hold data on the heap. There's no benefit in keeping it stack-based.

MrSach:
Thank you for the guidance on this.
I will investigate the error and also check the header and source files for the `Static` alternative for the `Storage` variant in question

madmaxoft:
Sure, go ahead, a draft pull request is the best way to check if you're heading in the right direction.

MrSach:
More commits have been added earlier this week to the draft pull request.
These propose a more suitable data type for storing compressed/uncompressed data.
Some commentary has been added to the header and source files.

PR #5503, opened 2023-05-24
String Compression - resolve MSVC warnings for stack memory allocation.

MrSach:
Changed the variable `Compresion::Result::Storage` of type `std::variant` to only use `Dynamic` for memory allocation in `./src/StringCompression.h`.

Changed some code that previously defaulted to assign data to the stack in `./src/StringCompression.cpp`.

Initialized `BytesWrittenOut` after MSVC threw another warning:
>Warning C6001 Using uninitialized memory 'BytesWrittenOut.'

in `./src/StringCompression.cpp`.

This PR should resolve issue #5495.

### DIFF
--- a/src/StringCompression.cpp
+++ b/src/StringCompression.cpp
@@ -12,7 +12,6 @@
 
 
 
-
 std::string_view Compression::Result::GetStringView() const
 {
 	const auto View = GetView();
@@ -30,16 +29,7 @@ ContiguousByteBufferView Compression::Result::GetView() const
 	{
 		std::visit([](const auto & Buffer) -> const std::byte *
 		{
-			using Variant = std::decay_t<decltype(Buffer)>;
-
-			if constexpr (std::is_same_v<Variant, Compression::Result::Static>)
-			{
-				return Buffer.data();
-			}
-			else
-			{
-				return Buffer.get();
-			}
+			return Buffer.get();
 		}, Storage), Size
 	};
 }
@@ -74,18 +64,7 @@ Compression::Compressor::~Compressor()
 template <auto Algorithm>
 Compression::Result Compression::Compressor::Compress(const void * const Input, const size_t Size)
 {
-	// First see if the stack buffer has enough space:
-	{
-		Result::Static Buffer;
-		const auto BytesWrittenOut = Algorithm(m_Handle, Input, Size, Buffer.data(), Buffer.size());
-
-		if (BytesWrittenOut != 0)
-		{
-			return { Buffer, BytesWrittenOut };
-		}
-	}
-
-	// No it doesn't. Allocate space on the heap to write the compression result, increasing in powers of 2.
+	// Allocate space on the heap to write the compression result, increasing in powers of 2.
 	// This will either succeed, or except with bad_alloc.
 
 	auto DynamicCapacity = Result::StaticCapacity * 2;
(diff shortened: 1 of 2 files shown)


MrSach:
Pull request has been created and is ready for review.
Thank you for your time and feedback.

src/StringCompression.cpp
```diff
@@ -12,7 +12,6 @@
 
 
 
-
```
madmaxoft, line 15: Please don't erase empty lines between functions (style: there are to be 5 empty lines between function definitions). Make sure you run the `src/CheckBasicStyle.lua` script before committing.

src/StringCompression.cpp
```diff
@@ -187,25 +166,12 @@ Compression::Result Compression::Extractor::ExtractZLib(ContiguousByteBufferView
...
-	}
-
-	// No it doesn't. Allocate space on the heap to write the compression result, increasing in powers of 2.
+	// Allocate space on the heap to write the compression result, increasing in powers of 2.
 
 	auto DynamicCapacity = Result::StaticCapacity * 2;
 	while (true)
 	{
-		size_t BytesWrittenOut;
+		size_t BytesWrittenOut {};
```
madmaxoft, line 174: This is more readable, if MSVC really requires initialization:
```suggestion
		size_t BytesWrittenOut = 0;
```

madmaxoft (changes requested):

src/StringCompression.cpp
MrSach: Ok, I will check the pre-commit hook.
I am not sure why it did not run but I will run the pre-commit hook manually next time before the next commit.
The deleted line will be re-added to allow for 5 lines between function definitions.

src/StringCompression.cpp
MrSach: That is useful to know.
I will use that syntax to initialize instead.

src/StringCompression.h
```diff
@@ -32,8 +33,9 @@ namespace Compression
 		/** Returns a view (of type std::byte) of the internal store. */
 		ContiguousByteBufferView GetView() const;
 
-		/** A store allocated on either the stack or heap. */
-		std::variant<Static, Dynamic> Storage;
+		/** A store allocated on the heap. */
+		/** Note that this remains a variant for now. */
+		std::variant<Dynamic> Storage;
```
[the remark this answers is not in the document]
MrSach: Yes, this previously needed the `std::variant` because of needing to accommodate either a `std::array<std::byte>` or `std::unique_ptr<std::byte[]>` for type safety.

To change `Storage` to a std::vector<std::byte> would need some refactoring in `./src/StringCompression.cpp`, but it would be good to simplify and to boost performance.

Would this be raised as an additional issue on git after this current issue?

src/StringCompression.cpp
madmaxoft: You need to explicitly enable the pre-commit hook on your local machine (and install Lua, if you haven't already). This is a security measure in git, so that foreign repositories can't run arbitrary code on your computer by putting it into a hook.

If you need a Lua interpreter for Windows, my [LunaPaak](https://github.com/madmaxoft/LunaPaak) project could help you - it's a single executable that you download and run once, it will register itself as Lua interpreter, and comes with several libraries built in, so it's pretty good for scripting.

madmaxoft:
You should just commit the changes to your repo under the same branch, this will update this PR with the new code.

MrSach:
Commit `12603f1` appears to have gone through correctly - the others made today can be ignored.
Feedback is welcome and if there are any further changes to make then please let me know.

src/StringCompression.cpp
```diff
@@ -25,22 +29,15 @@ std::string_view Compression::Result::GetStringView() const
...
-			if constexpr (std::is_same_v<Variant, Compression::Result::Static>)
-			{
-				return Buffer.data();
-			}
-			else
-			{
-				return Buffer.get();
-			}
-		}, Storage), Size
+		&(this->Storage.get()->front()),
```
MrSach, line 39: Is this a safe method to use?
I suppose that if `Storage` is not empty then this would work, but if it is then would this cause issues?

src/StringCompression.h
```diff
@@ -32,10 +35,13 @@ namespace Compression
 		/** Returns a view (of type std::byte) of the internal store. */
 		ContiguousByteBufferView GetView() const;
 
-		/** A store allocated on either the stack or heap. */
-		std::variant<Static, Dynamic> Storage;
+		/** A store allocated on the heap.
+		This uses a smart pointer. */
+		std::unique_ptr <std::vector <std::byte>> Storage;
```
MrSach, line 40: `std::unique_ptr <std::vector <std::byte>> Storage;`
As suggested, the data type now uses a `std::vector<std::byte>`.
To allocate on the heap, this has been declared using a smart pointer, specifically `std::unique_ptr`

MrSach:
https://github.com/cuberite/cuberite/blob/12603f14d59d1d8aac26543af1b3540dec717a77/src/StringCompression.cpp#L11C1-L15

Missing line should now have been re-added.

Let me know if there is anything else to refactor/clean up.

Thanks.

PR #5504, opened 2023-05-26
Fixed chunk HeightMap loading

madmaxoft:
Chunks without a HeightMap are normally loaded and their heightmap recalculated.


madmaxoft:
Ouch, the Unity builds have bitten me here, it works with Unity builds, but not in traditional builds.

NiLSPACE (approved):

PR #5507, opened 2023-05-28
Make commands non caps sensible

Heron4gf:
Added HOOK_PRE_EXECUTE_COMMAND hook
Made commands non caps sensible


madmaxoft:
Two points:
- Vanilla Minecraft is case-sensitive in its commands. So we should be, too
- Your PR adds unrelated changes and is "dirty" (imagine every PR had a `// ADDING XYZ FEATURE` comment in all its changes, the code would be a mess)

I vote to close this PR, but want to hear from other community members.

NiLSPACE:
Yeah, I agree. The ToDo list also lists features we don't plan on supporting at all like Java plugins and a Bukkit -> Cuberite plugin translator.

bearbin:
Closing, this almost looks like a mistaken pull request? I agree with @madmaxoft as well, we don't want to deviate from vanilla in this area.

Issue #5496, opened 2023-05-11
Replace hand-escaped JSON with proper JSON serialization


madmaxoft:
In several places within the codebase, we build a JSON string manually using `Printf` or `fmt::format`:
```cpp
SendChatRaw(fmt::format(FMT_STRING("{{\"text\":\"{}\"}}"), EscapeString(a_Message)), a_Type);
```

This should be replaced with code that uses the jsoncpp library already present within the codebase to serialize the object properly, such as:
```cpp
Json::Value RootValue;
RootValue["text"] = a_Message;
SendChatRaw(JsonUtils::WriteFastString(RootValue), a_Type);
```

Possibly the JSON serialization could be lifted to a separate string so that the call becomes:
```cpp
SendChatRaw(SerializeSingleValueJsonObject("text", a_Message), a_Type);
```

Griezn:
I would like to start contributing to this project, and this is an excellent issue to get going. But my question is in which file would I need to create the `SerializeSingleValueJsonObject("text", a_Message)` function?

madmaxoft:
Welcome. Try StringUtils, that's probably the best place for now.

PR #5508, opened 2023-05-30
Replaced hand-escaped JSON

Griezn:
I have replaced all hand-escaped JSON that I (and my IDE) could find. This should resolve issue #5496

### DIFF
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -6,8 +6,7 @@
 #include "Globals.h"
 
 #include "fmt/printf.h"
-#include "json/json.h"
-#include "JsonUtils.h"
+
 
 #ifdef _MSC_VER
 	// Under MSVC, link to WinSock2 (needed by RawBEToUTF8's byteswapping)
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -7,7 +7,6 @@
 
 #include "fmt/printf.h"
 
-
 #ifdef _MSC_VER
 	// Under MSVC, link to WinSock2 (needed by RawBEToUTF8's byteswapping)
 	#pragma comment(lib, "ws2_32.lib")
--- a/src/StringUtils.cpp
+++ b/src/StringUtils.cpp
@@ -6,6 +6,8 @@
 #include "Globals.h"
 
 #include "fmt/printf.h"
+#include "json/json.h"
+#include "JsonUtils.h"
 
 #ifdef _MSC_VER
 	// Under MSVC, link to WinSock2 (needed by RawBEToUTF8's byteswapping)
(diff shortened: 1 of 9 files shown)


src/StringUtils.cpp
```diff
@@ -6,6 +6,8 @@
 #include "Globals.h"
 
 #include "fmt/printf.h"
+#include "json/json.h"
```
madmaxoft, line 9: The compiler complains about this line.

From what I can tell, the StringUtils file is used in other subprojects that don't have the json library added to them. You need to add the jsoncpp dependency to GrownBiomeGenVisualiser.

Details: https://builds.cuberite.org/blue/organizations/jenkins/cuberite/detail/PR-5508/1/pipeline

madmaxoft (changes requested):

madmaxoft:
The JsonCpp library is already included in Cuberite, you should NOT use `find_package` for it. Instead, use `jsoncpp_static` as a dependency for the `GrownBiomeVisualiser` target.

Ref.: https://github.com/cuberite/cuberite/blob/26b7e5ad9edebfce0cbb51a5ba606bb0b334ad2d/CMake/AddDependencies.cmake#L61-L74

Griezn:
Whatever I do the build keeps failing. It keeps saying that it can't find `json/json.h` but I added this line to the CmakeLists.txt:
``include_directories(${PROJECT_SOURCE_DIR}/lib/jsoncpp/include)
``
And I linked the library:
`target_link_libraries(GrownBiomeGenVisualiser fmt::fmt Threads::Threads jsoncpp_static)`

But I keep getting this error:
```
/mnt/c/Users/seppe/CLionProjects/cuberite/src/StringUtils.cpp:9:10: fatal error: json/json.h: No such file or directory
    9 | #include "json/json.h"
      |          ^~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Tools/MCADefrag/CMakeFiles/MCADefrag.dir/build.make:104: Tools/MCADefrag/CMakeFiles/MCADefrag.dir/__/__/src/StringUtils.cpp.o] Error 1
```

madmaxoft:
GrownBiomeGenVisualiser now seems to work, MCADefrag is yet another tool that needs the json library added.

madmaxoft:
Huh, we've had a JsonUtils file all along? That would have been a much better place to put the `SerializeSingleValueJsonObject` function (and probably leave out the `Json` part of the name, since it's in the namespace anyway). Sorry about my first recommendation. Seems like I've lost touch with the codebase.

If you move the function to JsonUtils, you will no longer have to add the jsoncpp dependency to all the tools. That's a big win.

src/StringUtils.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "fmt/printf.h"
 
+
```
madmaxoft, line 10: Remove this change, and we're good :)

madmaxoft (changes requested):

bearbin (changes requested):
Please could you add yourself to the CONTRIBUTORS file to indicate your acceptance of the license, and so your efforts are recognised :)

Otherwise, this looks like a good change!

bearbin (approved):

madmaxoft:
Awesome, thanks.

PR #5514, opened 2023-06-16
Fix empty chunk serializer for protocol 47 (Minecraft 1.8)

sleirsgoevy:
This version of Minecraft does not like chunks that have their section bitmap set to 0 (meaning that the chunk is completely empty/filled with air). When sendings such chunks, at least 1 empty section should be sent.

Although such missing chunks do render as empty, not having them properly loaded causes various bugs, such as the "slow falling" of the player and missing block textures if any blocks get placed there later on.

### DIFF
--- a/src/Protocol/ChunkDataSerializer.cpp
+++ b/src/Protocol/ChunkDataSerializer.cpp
@@ -196,12 +196,16 @@ inline void cChunkDataSerializer::Serialize47(const int a_ChunkX, const int a_Ch
 	m_Packet.WriteBEInt32(a_ChunkX);
 	m_Packet.WriteBEInt32(a_ChunkZ);
 	m_Packet.WriteBool(true);      // "Ground-up continuous", or rather, "biome data present" flag
-	m_Packet.WriteBEUInt16(Bitmask.first);
+
+	// Minecraft 1.8 does not like completely empty packets
+	// Send one completely empty chunk section if this is the case
+	m_Packet.WriteBEUInt16(Bitmask.first ? Bitmask.first : 1);
 
 	// Write the chunk size:
+	// Account for the single empty section if sending an empty chunk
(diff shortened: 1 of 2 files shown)


src/Protocol/ChunkDataSerializer.cpp
```diff
@@ -196,12 +196,16 @@ inline void cChunkDataSerializer::Serialize47(const int a_ChunkX, const int a_Ch
 	m_Packet.WriteBEInt32(a_ChunkX);
 	m_Packet.WriteBEInt32(a_ChunkZ);
 	m_Packet.WriteBool(true);      // "Ground-up continuous", or rather, "biome data present" flag
-	m_Packet.WriteBEUInt16(Bitmask.first);
+
+	// Minecraft 1.8 does not like completely empty packets
+	// Send one completely empty chunk section if this is the case
+	m_Packet.WriteBEUInt16(Bitmask.first ? Bitmask.first : 1);
```
12xx12, line 202: Please explain the magic number. Does this "1" mean the chunk is empty? Or is this just a placeholder.

12xx12 (changes requested):
Hi and welcome to the project. 
Thank you for your contribution!

To enable us to merge the code you wrote please add yourself to the CONTRIBUTORS file. You can do this in the same PR

src/Protocol/ChunkDataSerializer.cpp
sleirsgoevy: The value means that the chunk has the lowest chunk section (16x16x16) non-empty, and the rest is empty. The following code is modified to transmit a single chunk section with all blocks being air.

Bixilon:
empty section bitmask means the chunk should be unloaded

12xx12:
Please add yourself to the CONTRIBUTORS file to acknowledge our license.

After that, we happily can accept your PRs. And maybe future ones.

bearbin (approved):

PR #5515, opened 2023-06-19
Fixed a wrong format string.

madmaxoft:
The `m_ItemCount` is a `char`, so it was default-printed as a character, not a number. This seems to be the only instance of the problem.

Thanks to @dyexlzc for noticing.


madmaxoft:
I'd like to merge this ASAP, any objections?

dyexlzc:
noop

12xx12 (approved):

PR #5521, opened 2023-06-29
Added that ds_store file to gitignore

12xx12:
So this is the first time the file popped up but no the last time

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,6 @@ run-clang-tidy.py
 
 # ctags output
 tags
+
+# mac things
+**/.DS_Store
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -126,3 +126,6 @@ run-clang-tidy.py
 
 # ctags output
 tags
+
+# mac things
+**/.DS_Store
```
PureTryOut, line 131: Just `.DS_Store` is good enough, it'll ignore it in every directory in the project.

Janmm14:
```suggestion
.DS_Store
```

bearbin (approved):

Issue #5542, opened 2023-12-29
Taking damage on TP


seem0nster:
Client version: 1.8.9
Server OS: Linux

### Expected behavior
To do "/home" from 0,75,0 - -200,25,150 Without taking damage while teleporting.

### Actual behavior
Taking damage if the coordinates you are teleporting to are below the Y coord you are at

### Maybe the solution
Setting the velocity to 0 while teleporting.

tonitch:
Confirmed: I tried with a simple `/tp` and the same happen.
I also noticed that when you `/tp ~ ~50 ~` you don't get fall damages.
That is strange I think this is supposed to be handled by https://github.com/cuberite/cuberite/blob/master/src/Entities/Player.cpp#L2530-L2559 which is called by [Player:TeleportToCords()](https://github.com/cuberite/cuberite/blob/20d5cf0bfcea65f2ff1faf86dba2069932cbfb3f/src/Entities/Player.cpp#L1404-L1417)

tonitch:
duped tho : #5286

seem0nster:
hmm, i think it is because it handles fall damage like if you go from Y- 80 to Y- 20, it takes where you fell to, not how long you have spent in the air.

English may be horrible.

NiLSPACE:
Fall damage is currently not handled well. We currently don't look at the y-speed of the player to calculate the damage amount but by looking at the current y-level compared to the previous on-ground y-level: https://github.com/cuberite/cuberite/blob/20d5cf0bfcea65f2ff1faf86dba2069932cbfb3f/src/Entities/Pawn.cpp#L434-L463

Issue #5286, opened 2021-08-26
Teleporting to players on a lower Y level to you applies damage to the player


PureTryOut:
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 4b8952e438da8e57d6442a0b7451dac8060f154b

### Expected behavior
Teleporting to players no matter where they are should not apply damage

### Actual behavior
Teleporting to players who are lower than you are gives you fall damage

### Steps to reproduce the behavior
1. Have 2 accounts on different Y levels, with a distance that while falling should give you fall damage
2. Teleport the higher account to the lower one

KingCol13:
I'm currently looking into this with other fall damage issues #4307 and #5323.

You don't need to acquire a guinea pig to reproduce this, you can simply teleport to a lower Y position.

Another issue I noticed is that teleporting to a higher Y level results in no fall damage after the fall.

KingCol13:
In the Notchian server, if you:

1. Record your position on the ground
2. Teleport high in the air
3. Fall for a while
4. Teleport back down to the ground

you will take fall damage. This implies that the Notchian server might be incrementing a counter of blocks fallen during the fall, which contrasts our approach of recording the last height at which the pawn was on the ground.

tigerw:
I'm sure there was code that attempted to nullify fall damage on teleportation, it seems to have regressed.

I feel like both vanilla and us are recording the Y position when the player moves with speed Y >= 0, and every time the player hits the ground see if that Y position is sufficiently different to current Y to trigger fall damage. Storing the apex of the fall essentially. But I don't think we're correctly determining all the times where Y speed >= 0, plus all the special cases about beds and cobwebs.

PR #5544, opened 2024-02-20
fix: #5542 Fixing damage on Teleport

tonitch:
Reseting the LastGroundHeight on teleport
Using inheritence for teleporting to avoid rewriting code

fix: #5542 (and potentialy other)


tonitch:
should close : https://github.com/cuberite/cuberite/issues/5286

tonitch:
@bearbin there seem to be a broblem with the CI (git not installed or something) and I heard that you are the one in charge ^^ can you take a look ?

bearbin (approved):
Wrong account, this looks fine as a hotfix.

tonitch:
tbh I don't really know how to add the bindings ^^

12xx12:
Add it in the APIDump plugin into the [APIDesc.lua](https://github.com/cuberite/cuberite/blob/master/Server/Plugins/APIDump/APIDesc.lua) file. Search for the other functions from cPawn

Server/Plugins/APIDump/APIDesc.lua
```diff
@@ -9829,6 +9829,25 @@ a_Player:OpenWindow(Window);
 					},
 					Notes = "Removes a currently applied entity effect",
 				},
+				TeleportToCoords =
```
mathiascode, line 9832: I have two questions/concerns:
- How do the API docs handle this override? Do we have other cases where we do this?
- Can this export ever be reverted in the future if necessary, without breaking things?

It's been a while since I touched any of this. If everything is fine, feel free to dismiss this comment.

tonitch: I honestly don't know how this api export works ^^ I though it would be better to use inheritence in this patch but it's indeed not strictly necessary so I can just do a dirty patch instead ^^

PR #5546, opened 2024-03-12
Enable build under clang-16.

### LINKED ISSUE #5462: Build errors using Clang 15
Client version: x.x.x
Server OS: Linux
Cuberite Commit id: a7f287e572866403f1f68984f61b3a650b0e798a

```
/home/bond/dev/cuberite/lib/mbedtls/library/bignum.c:1395:29: error: variable 't' set but not used [-Werror,-Wunused-but-set-variable]
    mbedtls_mpi_uint c = 0, t = 0;
                            ^
1 error generated.
```

madmaxoft:
This enables Cuberite to be built out-of-the-box on FreeBSD with clang-16.

This only masks out the new warnings that were causing the build to fail due to treat-warnings-as-errors, it is NOT fixing the underlying issues!

Fixes #5462

### DIFF
--- a/SetFlags.cmake
+++ b/SetFlags.cmake
@@ -202,5 +202,11 @@ function(set_exe_flags TARGET)
 				-Wno-reserved-identifier
 			)
 		endif()
+		if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 16)
+			target_compile_options(
+				${TARGET} PRIVATE
+				-Wno-unsafe-buffer-usage
+			)
+		endif()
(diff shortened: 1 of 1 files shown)


Bond-009:
This is also an issue with Clang 15, not just 16 and up
https://github.com/cuberite/cuberite/issues/5462

madmaxoft:
Alright, I've included clang-15 into the changes, assuming that only the reported error shows up. I have no way of verifying.

NiLSPACE (approved):

PR #5552, opened 2024-03-21
Moved end generator back to 3d perlin noise

NiLSPACE:
In #4973 I changed the end generator to 2 2d perlin noises. Since I still think the 3d noise looked superior I changed it back, but now the world continues generating after a void. A big advantage to this approach is that the islands don't get cut off

**Main island:**

**Smaller islands out in the void (higher FOW):**

**Minutor overview:**
overview-pow6

The main island could still be improved. It's still possible that the exit portal (coordinates 0, 0) is over a void.

### DIFF
--- a/src/Generating/EndGen.cpp
+++ b/src/Generating/EndGen.cpp
@@ -23,7 +23,7 @@ enum
 	DIM_X = 16  / INTERPOL_X + 1,
 	DIM_Y = 256 / INTERPOL_Y + 1,
 	DIM_Z = 16  / INTERPOL_Z + 1,
-} ;
+};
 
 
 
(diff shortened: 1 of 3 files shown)


src/Generating/EndGen.cpp
```diff
@@ -62,95 +59,118 @@ cEndGen::cEndGen(int a_Seed) :
...
+	NOISE_DATATYPE frequencyX = distanceFromSpawn > m_MainIslandSize * 2 ? m_SmallIslandFrequencyX : m_MainIslandFrequencyX;
+	NOISE_DATATYPE frequencyY = distanceFromSpawn > m_MainIslandSize * 2 ? m_SmallIslandFrequencyY : m_MainIslandFrequencyY;
+	NOISE_DATATYPE frequencyZ = distanceFromSpawn > m_MainIslandSize * 2 ? m_SmallIslandFrequencyZ : m_MainIslandFrequencyZ;
+
+	// Generate the downscaled noise:
+	NOISE_DATATYPE StartX = static_cast<NOISE_DATATYPE>(m_LastChunkCoords.m_ChunkX       * cChunkDef::Width) / frequencyX;
+	NOISE_DATATYPE EndX   = static_cast<NOISE_DATATYPE>((m_LastChunkCoords.m_ChunkX + 1) * cChunkDef::Width) / frequencyX;
+	NOISE_DATATYPE StartZ = static_cast<NOISE_DATATYPE>(m_LastChunkCoords.m_ChunkZ       * cChunkDef::Width) / frequencyZ;
+	NOISE_DATATYPE EndZ   = static_cast<NOISE_DATATYPE>((m_LastChunkCoords.m_ChunkZ + 1) * cChunkDef::Width) / frequencyZ;
+	NOISE_DATATYPE StartY = 0;
```
12xx12, line 114: Aren't we trying to use more of `auto`?

src/Generating/EndGen.cpp
```diff
@@ -62,95 +59,118 @@ cEndGen::cEndGen(int a_Seed) :
...
+	NOISE_DATATYPE EndY   = static_cast<NOISE_DATATYPE>(257) / frequencyY;
```
12xx12, line 115: What does this magic number do? Just one above the chunk limits?

src/Generating/EndGen.cpp
```diff
@@ -62,95 +59,118 @@ cEndGen::cEndGen(int a_Seed) :
...
+			double mult = 3 * ((distanceFromSpawn - m_MainIslandSize) / m_MainIslandSize);
+			double threshold = std::min(pow - mult, minThreshold);
+
+			for (int y = 0; y < m_IslandYOffset; y++)
 			{
-				// The distance from spawn is so big we don't need to calculate the max height anymore.
-				// In fact, if we don't cut it off somewhere there is a chance the maxheight gets too big which
-				// can cause corrupted looking terrain.
-				maxHeightLimit = static_cast<float>(cChunkDef::Height);
+				a_Shape[y + x * 256 + z * 256 * 16] = 0;
```
12xx12, line 162: More magic numbers with available constants. I know that in this case, the code is longer but I think it might be helpful

12xx12 (commented):
The change looks good.

What does the change of noise do? Looking at the old PR the size of the island changes mostly.

I do have some small QOL notes on the code.

Do you mind updating the manual when the PR is merged

NiLSPACE:
The magic numbers are there because I reverted back to the original end generator code which had them, but you're right. I should change those to constants.

src/Generating/EndGen.cpp
NiLSPACE: I think it might be useful to create a separate function in `cChunkDesc` that sets and gets the shape values. Now all generators manually calculate the index.

12xx12: Or in the common grandparent object

src/Generating/EndGen.cpp
NiLSPACE: Is that an ongoing process? I haven't been that involved lately and I don't see the `BiomalNoise3DComposable` generator use `auto` for `NOISE_DATATYPE`

src/Generating/EndGen.cpp
NiLSPACE: `Shape` is a typedef of an array in `cChunkDesc`, so I think it makes sense to put the functions next to that.

12xx12 (approved):
I like the new methods. Maybe other generators can use them at some point

PR #5553, opened 2024-03-31
1.15 support

MightyFilipns:
1.15.x barebones support

C2S packets are mostly unchanged

S2C packets:
Ids of most of them have been increased by 1 or 2
Some packets had their format changed

A new chunk serializer was added - v573.
Incomplete and probably buggy.

- Added basic handling for Book Update C2S Packet
Does nothing at the moment just reads the data from the packet so the server does not crash.

- Implemented some 1.14 and 1.15 items and blocks
They have no functionality currently, they can just be placed if they are blocks.

- In 1.15 vertical biomes were added
The existing 2D biome map is translated into the new format. 

Known bugs

New 1.15 blocks aren't properly serialized when they have to be sent or saved. So after the chunk they are in is unloaded and loaded again they disappear.

Some blocks aren't implemented at all (corals, etc)

In 1.15.1 and 1.15.2 lava has the water texture when placed. 

Some spawn eggs don't work or spawn the wrong mob.
Mobs in 1.15 are incomplete.

~~Biomes aren't implemented so everywhere is the ocean biome.~~

When placing block chunks glitch and all blocks in that chunk disappear.
They show up again if a few blocks are broken in the same chunk

When using redstone to open or close doors the upper block of the door does not update. (looks like it is present in older versions as well)

Using a fishing rod crashes the server

player movement isn't correct


Janmm14:
bunch of changes are only: tabs -> spaces
i suggest undoing such changes

MightyFilipns:
> bunch of changes are only: tabs -> spaces
> i suggest undoing such changes

Should be fixed now

PR #5556, opened 2024-04-25
Fix filename template for saved maps

mjagdis:
The conversion from printf formats missed one. All maps save to the same file :-(


bearbin (approved):

bearbin:
Thanks, great catch!

PR #5557, opened 2024-04-25
Save changed maps every 5 minutes

mjagdis:
Previously they were only ever saved on a clean exit of the server. That is somewhat... fragile... so save them periodically with the same interval that we save player data.

### DIFF
--- a/src/Map.cpp
+++ b/src/Map.cpp
@@ -23,6 +23,7 @@ cMap::cMap(unsigned int a_ID, cWorld * a_World):
 	m_Scale(3),
 	m_CenterX(0),
 	m_CenterZ(0),
+	m_Dirty(false),
 	m_World(a_World),
 	m_Name(fmt::format(FMT_STRING("map_{}"), m_ID))
 {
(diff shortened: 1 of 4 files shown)


src/Map.cpp
```diff
@@ -40,6 +41,7 @@ cMap::cMap(unsigned int a_ID, int a_CenterX, int a_CenterZ, cWorld * a_World, un
 	m_Scale(a_Scale),
 	m_CenterX(a_CenterX),
 	m_CenterZ(a_CenterZ),
+	m_Dirty(true),
```
bearbin, line 44: Is there a reason why this is true whereas the other initialisation on line 26 sets it to false? Doesn't make sense that adding a center position and scale would make the empty map dirty but not having them would be a clean map.

bearbin (commented):
Otherwise this looks good, sorry for the slow review.

src/Map.cpp
mjagdis: m_Dirty means the map needs saving. Centre position and scale are things that are saved as part of the map files. Ergo, if you change them you need to save the map.

bearbin: What you meant to say was that this constructor is only called when creating a new map (which needs to be saved) while the other constructor is only used when loading a map (which does not need to be saved).

bearbin, line 44:
```suggestion
	m_Dirty(true), // This constructor is for creating a brand new map in game, it will always need saving.
```

src/Map.cpp
```diff
@@ -23,6 +23,7 @@ cMap::cMap(unsigned int a_ID, cWorld * a_World):
 	m_Scale(3),
 	m_CenterX(0),
 	m_CenterZ(0),
+	m_Dirty(false),
```
bearbin, line 26:
```suggestion
	m_Dirty(false), // This constructor is for an empty map object which will be filled by the caller with the correct values - it does not need saving.
```

bearbin (approved):

PR #5558, opened 2024-04-29
Set TCP_NODELAY on connections

### LINKED ISSUE #5065: Investigate TCP_NODELAY
Nagle's algorithm helps prevent the case where packets contain mostly header overhead and tiny fractions of payload, because the application does a network send immediately when any amount of data are generated. However:

ClientHandle does its own packet coalescing (per tick); we should be okay to request an immediate send without further buffering, and I understand that this is also desirable in general for game networking.

Other uses of TCP sockets (RCON, HTTP) can probably leave this option unchanged.

mjagdis:
Fixes #5065 

Minecraft/cuberite does its own buffering at the app layer and there's little point holding back a trailing partial packet.

### DIFF
--- a/src/Globals.h
+++ b/src/Globals.h
@@ -98,6 +98,7 @@
 
 	#include <arpa/inet.h>
 	#include <netinet/in.h>
+	#include <netinet/tcp.h>
 	#include <sys/socket.h>
 	#include <unistd.h>
 #endif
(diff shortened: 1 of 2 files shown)


bearbin (approved):
LGTM, @12xx12 any thoughts?

mathiascode (approved):

PR #5565, opened 2024-06-19
Preserve banner names across place and pick up

### LINKED ISSUE #5564: Banners lose their names after being placed and picked up
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: a66a67c

### Expected behavior

Banner name should be retained.

### Actual behavior

Banner name is gone.

Name a banner on an anvil. Place it down. Pick it up. Bang! And the name is gone!

mjagdis:
As with named enchanting tables, copy banner names back and forth between items and block entities. NBT format verified against actual Java Edition chunk saves.

Fixes #5564

### DIFF
--- a/src/BlockEntities/BannerEntity.h
+++ b/src/BlockEntities/BannerEntity.h
@@ -25,15 +25,19 @@ class cBannerEntity :
 
 public:
 
-	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World);
-	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World, unsigned char a_BaseColor);
+	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World, unsigned char a_BaseColor = 15, AString a_CustomName = "");
 
-	unsigned char GetBaseColor() const;
-	void SetBaseColor(unsigned char a_Color);
+	unsigned char GetBaseColor() const { return m_BaseColor; }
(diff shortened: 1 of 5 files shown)


src/BlockEntities/BannerEntity.h
```diff
@@ -25,16 +25,20 @@ class cBannerEntity :
 
 public:
 
-	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World);
-	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World, unsigned char a_BaseColor);
+	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World, unsigned char a_BaseColor = 15, AString a_CustomName = "");
```
bearbin, line 28: Previously the default BaseColor was 1, now it's 15?

mjagdis: It's a bit pedantic, I guess :-). Banner colours are inverted 0-15 so 15 is white (0 is black and 1 is just somewhat random). That said crafting a banner always sets damage (i.e. colour) on the crafted item, loading by WSSAnvil always sets the colour to whatever was saved and placing a banner always does a SetBaseColor using the source item's damage. Unless there's a way to magic a banner into existence from Lua (I haven't checked) those two defaulted args have no real purpose and could be vacuumed up?

bearbin, line 28:
```suggestion
	cBannerEntity(BLOCKTYPE a_BlockType, NIBBLETYPE a_BlockMeta, Vector3i a_Pos, cWorld * a_World, unsigned char a_BaseColor = 1, AString a_CustomName = "");
```

They could probably be cleaned up, something for another day though. Let's leave it how it was at the start.

bearbin (approved):

PR #5567, opened 2024-06-23
Add NBT (de)serialize for 1.13+ protocols

### LINKED ISSUE #5566: Creating a map dups map #0
Client version: 1.8.9, 1.9.4, 1.12.2 / 1.13.2
Server OS: Linux
Cuberite Commit id: latest master

### Expected behavior

Right clicking an empty map creates a new map and allocates a new map number.

### Actual behavior

The new map is created with the expected next map number but the count of "Map #0" items also increases by one.

Give yourself some empty maps (works in creative or survival). Right click. Now you have a "Map #0". Right click again and you have "Map #1" and two "Map #0"s. This continues until you have a whole stack of "Map #0"s then the duping stops - it doesn't creates a new stack.

In 1.13.2 the behaviour is... different... Sometimes it adds a map to a new inventory slot, sometimes it increases the count on an existing slot (the last one added?). However ALL maps appear as "Map #0". And yes, I know SendMap isn't implemented for the 1.13 protocol and I'm not expecting the maps to have any actual content. But surely the items should be somewhat sensible? I'd be fine if it just didn't work but this is just weird!

mjagdis:
Fixes #5566 

Protocols for 1.13 and later clients need to send/receive metadata as blocks of NBT.

Map creation when an empty map is used needs to refresh potentially affected inventory slots because the client makes some bad predictions about how they might change.


Janmm14:
Are you positive its a specific issue jsut with map # 0 only?

mjagdis:
It's not specific to map 0 as such. What happens is that when you right click an empty map the client goes, "ah, a map is being added, empty maps - 1, add a new map, uh, oh, maybe... 0?" and adds a "Map #0" to the inventory. That happens (visibly) before HandlePacketUseItem even gets called on the server. The server then does the right thing, allocates a new map number, creates the item, adds it to the inventory and sends the client a SetSlot to tell it what happened. But the client's predicted map 0 went "somewhere" - maybe stacked with another "Map #0", maybe into an empty slot. So we need to tell the client what SHOULD be in any slot we know is either empty or has one or more "Map #0"s. One of them is wrong on the client but we don't know which one :-(

The 1.13 problem, ah, "enhances" that because if the client doesn't get the map number in item NBT data it thinks _every_ map is #0 - even the ones correctly added by the server. Even more entertaining is that if you don't parse NBT on item data received by HandlePacketCreativeAction then the as soon as you open the inventory in creative mode the _server_ becomes persuaded that all your maps are #0 too! Passing NBT backwards and forwards probably also fixes, or at least heps with, other issues with damage, repair costs, custom name and other stuff.

Interesting times, huh?

PR #5570, opened 2024-07-01
Uninitialised value fix

solvictor:
Prevent some conditional jumps on uninitialised values. Found with valgrind in the terrain generator and Client Handler.

### DIFF
--- a/src/Noise/InterpolNoise.h
+++ b/src/Noise/InterpolNoise.h
@@ -348,17 +348,13 @@ public:
 		ASSERT(a_StartY < a_EndY);
 
 		// Calculate the integral and fractional parts of each coord:
-		int FloorX[MAX_SIZE];
-		int FloorY[MAX_SIZE];
+		int FloorX[MAX_SIZE] = {};
+		int FloorY[MAX_SIZE] = {};
 		NOISE_DATATYPE FracX[MAX_SIZE];
 		NOISE_DATATYPE FracY[MAX_SIZE];
 		int SameX[MAX_SIZE];
 		int SameY[MAX_SIZE];
 		int NumSameX, NumSameY;
-
-		// Prevent uninitialized value when Cell.move()
-		memset(FloorX, 0, sizeof(int) * MAX_SIZE);
-		memset(FloorY, 0, sizeof(int) * MAX_SIZE);
 
 		CalcFloorFrac(a_SizeX, a_StartX, a_EndX, FloorX, FracX, SameX, NumSameX);
 		CalcFloorFrac(a_SizeY, a_StartY, a_EndY, FloorY, FracY, SameY, NumSameY);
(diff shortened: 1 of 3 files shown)


mjagdis:
Is that a real problem or just something that valgrind needs some help with? The Floor array are outputs from CalcFloorFrac as are how much of them are used. Any uninitialized part should be irrelevant.

If you really need them fully initialized it might be nicer to say, "int FloorX[MAX_SIZE] = { };" instead of using memset.

solvictor:
Can someone help me figure out why Windows-x64-Debug is not passing ? Everything looks good.

bearbin:
Not sure why it crashed, maybe the build agent got GCd or something. Reran and it worked :)

bearbin (changes requested):
Looks like a reasonable change. Please could you add your name in the correct alpha-sorted position to the CONTRIBUTORS file to recognise your contribution under the correct license?

bearbin (commented):
This looks sensible, I was just going to merge, but thinking again about what @mjagdis said, if we're actually getting uninitialised reads from these vars there's probably something wrong with the algorithm or its usage somewhere and we ought to look into that.

Do you have a reproducible test case that will trigger these uninitialised reads? I'd like to take a look and step through the code and debug what's actually happening - OFC feel free to do this yourself as well!

bearbin:
I was referring to the noise change above, client handle change looks fine. If you could get a nice minimal example (maybe just make up a cpp file that uses the noise gen to trigger the read) that would be ideal!

solvictor:
To reproduce the bug you can start a server with an empty world, the bug occurs when creating new chunks. To debug it you must compile in Debug mode and let the arrays uninitialized and launch the server with `valgrind --track-origins=yes ./Cuberite`. I don't know how the algorithm is supposed to work or what part of it is accessing theese values that are not supposed to be accessed.

iamnoksio:
it's a perfect temp fix, push it

mjagdis:
> it's a perfect temp fix, push it

The m_ProxyConnection bit, yes - it's a proper fix.

The Floor bit, no. There is no such thing as a "temp fix" and applying this would hide the only thing that suggests anything might be wrong here. So either someone needs to explain why there are accesses to Floor beyond a_Size and why those accesses should return 0 rather than some other value or someone needs to figure out what actually needs fixing.

I'd split it in two, a patch and an issue.

mjagdis:
Forget the issue. See #5576

src/Noise/InterpolNoise.h
```diff
@@ -348,13 +348,14 @@ class cInterpolNoise
...
-		int FloorY[MAX_SIZE];
+		int FloorX[MAX_SIZE] = {};
+		int FloorY[MAX_SIZE] = {};
 		NOISE_DATATYPE FracX[MAX_SIZE];
 		NOISE_DATATYPE FracY[MAX_SIZE];
 		int SameX[MAX_SIZE];
 		int SameY[MAX_SIZE];
 		int NumSameX, NumSameY;
+
 		CalcFloorFrac(a_SizeX, a_StartX, a_EndX, FloorX, FracX, SameX, NumSameX);
```
madmaxoft, line 359: These are not good changes. They hurt performance by needlessly filling the arrays with zeroes and then immediately overwriting the values with the proper calculated numbers.
Also the `CalcFloorFrac` call is tightly coupled to the declarations, the additional empty line breaks this logical tie.

As @mjagdis already pointed out, this needs a deeper fix, which he even provided, so please remove this from the PR.

madmaxoft (changes requested):

src/Noise/InterpolNoise.h
madmaxoft, line 1: Changes to this file should be dropped from the PR altogether. Then it's fine to go.

madmaxoft (changes requested):

madmaxoft (approved):
Awesome, thanks

solvictor:
@bearbin Can you review please ?

madmaxoft:
I took the liberty of merging it, the change itself is tiny so it's okay, and Bearbin's request was met.

iamnoksio:
FINALLY DAMN, MERGING AFTER MONTHS IS CRAZY

PR #5572, opened 2024-07-02
Convert double to ints with floor rather than truncating

### LINKED ISSUE #5571: Item frames won't always stay where they are put
Client version: all
Server OS: all
Cuberite Commit id: current master

### Expected behavior

Item frames stay where they are put if the block they are on is valid.

### Actual behavior

Item frames only stay put sometimes. Other times they drop after a few seconds.

Just spam item frames around and watch some of them rain down.

mjagdis:
Fixes #5571 

The implicit conversion from double or float to integers simply truncates or rounds towards zero. That's not what we want and leads to off-by-one errors when we have a negative value. One side effect of that is that if an item frame is placed on a block that has one or more negative components in its position vector then the on-tick check to see if the item frame is still attached to a valid block looks at the wrong block and draws an erroneous conclusion (i.e. the item frame drops).

What we want is for Vector3d to be converted to Vector3i by flooring each component rather than truncating.

### DIFF
--- a/src/Vector3.h
+++ b/src/Vector3.h
@@ -30,11 +30,18 @@ public:
 
 	// tolua_end
 	// Conversion constructors where U is not the same as T leaving the copy-constructor implicitly generated
-	template <typename U, typename = typename std::enable_if<!std::is_same<U, T>::value>::type>
-	constexpr Vector3(const Vector3<U> & a_Rhs):
+	template <typename U, std::enable_if_t<(!std::is_same<U, T>::value) && ((!std::is_integral<T>::value) || (std::is_integral<U>::value)), bool> = true>
+	constexpr Vector3<T>(const Vector3<U> & a_Rhs):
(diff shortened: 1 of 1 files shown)


bearbin:
This is a fairly major change, since the behaviour of all casts from Vector3[d,f] to Vector3i will be modified. Not saying this is wrong, it's probably actually an improved semantic that'll clear up some bugs but we need to be quite careful that we're not actually introducing 10 bugs in the process of fixing one.

bearbin:
As an interim fix, might it be better to do an explicit floor in the on-block checking routine rather than relying on the implic integer conversion? That way we only affect the known-buggy code and not the whole codebase.

mjagdis:
If you ifdef the second flooring converter out the compiler will tell you all the places that do hidden 3d-to-3i conversions. I couldn't see anything that gave me reason to make everything explicit.

bearbin (approved):

PR #5573, opened 2024-07-02
The second value in Rotation is pitch not roll

mjagdis:
The second value we read for "Rotation" in WSSAnvil entity NBT is pitch not roll (it's correct in the NBT serializer). Getting it wrong messes up the hit boxes of top and bottom placed item frames. Which the server doesn't support. Yet... Maybe there are other things it matters for though?


bearbin (approved):
Good catch!

PR #5576, opened 2024-07-24
Fix access to uninitialized space

mjagdis:
Sooo...

CalcFloorFrac sets up the Floor arrays using a_Size element and returning Same and NumSame. We then loop over the x, y, z dimensions from 0 to NumSame doing a Cell.Generate followed by a Cell.Move. The Cell.Generate does the real work, the Cell.Move sets up for the next iteration. But for the last iteration of each loop there is no next and the Floor[To] passed to Cell.Move is off the end of the space used in Floor. For the last iterations it looks like we should be wrapping back to the beginning of the Floor arrays. It doesn't seem to make an observable difference (it's all noise :-) ). It just avoids the dodgy access.

### DIFF
--- a/src/Noise/InterpolNoise.h
+++ b/src/Noise/InterpolNoise.h
@@ -441,19 +441,19 @@ public:
 				{
 					int ToX = FromX + SameX[x];
 					Cell.Generate(FromX, ToX, FromY, ToY, FromZ, ToZ);
-					if (++x < NumSameX)
+					if (++x < NumSameX)  // Call Move() every time except for the last loop iteration
 					{
 						Cell.Move(FloorX[ToX], CurFloorY, CurFloorZ);
 						FromX = ToX;
 					}
 				}
-				if (++y < NumSameY)
+				if (++y < NumSameY)  // Call Move() every time except for the last loop iteration
 				{
 					Cell.Move(FloorX[0], FloorY[ToY], CurFloorZ);
 					FromY = ToY;
 				}
 			}  // for y
-			if (++z < NumSameZ)
+			if (++z < NumSameZ)  // Call Move() every time except for the last loop iteration
 			{
 				Cell.Move(FloorX[0], FloorY[0], FloorZ[ToZ]);
 				FromZ = ToZ;
--- a/src/Noise/InterpolNoise.h
+++ b/src/Noise/InterpolNoise.h
@@ -427,28 +427,37 @@ public:
 
 		// Calculate query values using Cell:
 		int FromZ = 0;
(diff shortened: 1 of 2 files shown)


madmaxoft (changes requested):
Good catch. However, when re-reading the code, it seems we can skip the `Move` calls at the end of each `for` loop completely.
Also, don't forget there's a `Generate2D` function as well using the same approach, needing the same fix.

madmaxoft:
The way I see it, at the end of an inner loop there is a `Move` operation (line 444), followed by another `Move` just below that inner loop (line 447). So when the loop terminates, the inner `Move` should not be needed.

mjagdis:
It's not obvious to me that the second Move cancels out the effect of the first. The arguments are different and Move uses all of them to generate the next m_WorkRnds so skipping a move would seem to change the sequence. Mind you, so does injecting an arbitrary value. And I haven't looked in detail at Move. And I don't even know if it matters? It's all, literally, noise, isn't it?

madmaxoft:
`Move` just sets up the internal variables so that the call to `Generate` has everything set up correctly; calling it should reset the entire internal state based only on the params, so calling it twice is the same as calling it only the second time.
Basically, it calculates the noise values at the integral coords, with `Generate` then interpolating between those at the non-integral coords.

mjagdis:
Well, Move seems to use state saved by the prior call as well as the params. But I can't see that it matters so...

madmaxoft:
It only uses that state as a cache, not to recalculate the same noise values over and over again if they are already present.

src/Noise/InterpolNoise.h
```diff
@@ -427,28 +427,37 @@ class cInterpolNoise
...
 				int CurFloorY = FloorY[FromY];
 				int FromX = 0;
-				for (int x = 0; x < NumSameX; x++)
+				for (int x = 0; x < NumSameX;)
 				{
 					int ToX = FromX + SameX[x];
 					Cell.Generate(FromX, ToX, FromY, ToY, FromZ, ToZ);
-					Cell.Move(FloorX[ToX], CurFloorY, CurFloorZ);
-					FromX = ToX;
+					if (++x < NumSameX)
```
madmaxoft, line 444:
```suggestion
					if (++x < NumSameX)
					if (++x < NumSameX)  // Call Move() everytime except for the last loop iteration
```
Etc. in respective locations

madmaxoft (changes requested):
The construct is somewhat unexpected, so it warrants some comments around it.

madmaxoft (approved):

madmaxoft:
Sorry for the long wait.

PR #5581, opened 2024-08-30
fix custom mob names

### LINKED ISSUE #5577: Custom names crash 1.13+ clients
Client version: 1.13.2
Server OS: Linux
Cuberite Commit id: HEAD

### Expected behavior

Names float happily above mobs.

### Actual behavior

Client crashes with an invalid JSON error.

Add a name to a tag on an anvil, hold the tag, right click a mob and... boom!

mrbesen:
fixes #5577 

is there some method to convert a string to a chat component and that to a json? I guess that would bee the correct way of doing this.
Maybe we should create a `cPacketizer::WriteChatComponent()` ?

### DIFF
--- a/src/Protocol/Protocol_1_13.cpp
+++ b/src/Protocol/Protocol_1_13.cpp
@@ -969,7 +969,7 @@ void cProtocol_1_13::WriteMobMetadata(cPacketizer & a_Pkt, const cMonster & a_Mo
 		// TODO: As of 1.9 _all_ entities can have custom names; should this be moved up?
 		WriteEntityMetadata(a_Pkt, EntityMetadata::EntityCustomName, EntityMetadataType::OptChat);
 		a_Pkt.WriteBool(true);
-		a_Pkt.WriteString(a_Mob.GetCustomName());
+		a_Pkt.WriteString("{\"text\":\"" + a_Mob.GetCustomName() + "\"}");
 
 		WriteEntityMetadata(a_Pkt, EntityMetadata::EntityCustomNameVisible, EntityMetadataType::Boolean);
 		a_Pkt.WriteBool(a_Mob.IsCustomNameAlwaysVisible());


mjagdis:
It would be simpler just to do:
```
a_Pkt.WriteString(JsonUtils::SerializeSingleValueJsonObject("text", a_Mob.GetCustomName()));
```
That needs to be done for all mob, item and player custom names as well and I believe it can be done for all protocols Cuberite supports. The plain text support in pre-1.13 clients comes from way back. All the client protocols Cuberite supports should handle both plain text and JSON. It's just that 1.13 clients **require** JSON.

HOWEVER...

On that basis and because nothing in the core cares it may be even better to say that all the m_CustomNames are simply JSON strings and have SetCustomName check if it gets parseable JSON (test with JsonUtils::ParseString) and wrap in "{\"text\":\"...\"}" if it doesn't. That allows the full possibilities of [text formats](https://wiki.vg/Text_formatting#Text_components) with minimal pain.

N.B. GetCustomName is also exposed to Lua. I don't know if anyone would care if it returns a JSON string rather than plain text. I doubt anything uses GetCustomName but... YMMV...

mrbesen:
> It would be simpler just to do:
> 
> ```
> a_Pkt.WriteString(JsonUtils::SerializeSingleValueJsonObject("text", a_Mob.GetCustomName()));
> ```

Sounds good, i did not know about that.

But thinking about it abit, i belive the correct course of action might be to change the internal structures to `CompositeChat` (or something similar) aswell. Especially Mobnames, that might be styled to something special by some plugins.

I have already added that change for [item names](https://github.com/mjagdis/cuberite/pull/1) in a diffrent branch. The change for player names is also in this branch (it caused nether portals to crash the client, because of the way the client handles the respawn).

I have posted a discussion question about the structure of  `CompositeChat` in the discords dev-discussion channel. Is that appropriate there, or should i repost it somewhere else?

mjagdis:
It's worth considering. But:

1. The core only needs the stringified-JSON not the JSON tree
2. The custom names are expected to only change rarely so don't need to be re-stringified every time they are sent
3. If you expose too much to Lua you have the question of how you coordinate changes to particular elements of the JSON tree across different Lua mods which might have **very** different ideas of what JSON structure should be used.

So I would go for making Lua mods take ownership of, and set, the complete custom name using a minimally parseable JSON string. Keep the core simple. Except for maybe exposing some JSON handling lib to the Lua side.

Change related discussions should be on PRs or issues where one exists since it's all part of the VCS history. Besides, I don't use Discord. Then again I'm not one of the core devs so no need to feel beholden to me!

src/Protocol/Packetizer.cpp
```diff
@@ -6,6 +6,7 @@
 #include "Globals.h"
 #include "Packetizer.h"
 #include "../UUID.h"
+#include "../CompositeChat.h"
```
12xx12, line 9: Please use imports relative to `src/`

we agreed to change that some time ago but didn't want create a giant commit just to do this

12xx12 (changes requested):
I just have a minor comment.

I wait for a second person to review before merging

PR #5582, opened 2024-09-01
Support for 1.15-1.21.4

MightyFilipns:
Basic connection support for 1.15 - 1.21.4
Basic support for reading and writing the 1.13+ world format
New world format reader needs more testing
In 1.20.2+ and configuration phase has been added. Some changes were made to accommodate this.
Also, in 1.20.2+ NBT data sent over the network was slightly changed but critically [changed](https://wiki.vg/NBT#Network_NBT_(Java_Edition)). A constructor was added to FastNBTWriter to facilitate this change, however it should be reworked. Because of this in 1.20.2+ there are many packets that will disconnect the client because the NBT isn't in properly serializd.
In 1.16+ a new NBT data structure was added to The GameJoin packet, and in 1.20.2+ it was moved to its own packet. This structure specifies many things about each dimension. Currently, all data is hardcoded as there is no system to handle this.
Implemented CommandTree packet. This packet allows clients to autocomplete commands. However, only the command name can be autocomplete using the old system. 
Because of this, I started implementing a new command system.
The old command system is still used because the new one is incomplete as most argument types are missing and few commands are implemented using it.
Implemented the sample (a.k.a player list) field in status packets.
In 1.19 significant changes to chat were made. This mainly consists of introducing a bunch of cryptography.
Most of these changes aren't implemented. All chat messages are currently sent via the GameMessage packet.
In versions 1.19 - 1.19.3 when setting up encryption nounces may be sent either encrypted or signed, when they are signed they aren't checked.
Implements player tab list in 1.19+.
Implements channels in CustomPayload packets using the old system.
Chunk Serializers in 1.15+ only serialize blocks. Biomes are hard coded
Item metadata serialization is mostly implemented up to 1.20.5.
In 1.20.5 items are serialized using the new ItemComponent system. Currently this is not handled at all.
MapData packets are partially implemented.
Implements LongArray NBT tag.
A bunch of small and somewhat bigger changes everywhere. :)
Uses the new block system #5151


MightyFilipns:
For now don't plan on making more major changes. As I need feedback on everything I have done so far.
Specifically:

- The new command system. Can be found in src/Commands folder. All commands are created in PluginManger.cpp SetupNewCommands Function.
- World Reader/Writer for the new chunk format. Reader in WSSAnvil.cpp. Writer in NBTChunkSerializer.cpp

Also I am unsure how to implement some things:

- How do we deal with the [new item component system](https://wiki.vg/Slot_Data)  ?
- How should the [RegistryData](https://wiki.vg/Protocol#Registry_Data) be implemented. Just hardcode everything for each version or preferably have some dynamic system.
- Should we bother with the new chat system or just send everything via GameMessage packets

12xx12:
here we go! more progess :D

I haven't worked on my branch since ages and as far as I know it doesn't work properly :(

MightyFilipns:
It does mostly. Most of the changes I had to make to your branch to update it to the newest versions were to update some blocks that had a waterlogged block state added. Apart from that the worst bug I found was end portals causing stack corruption when trying to determine if the portal is complete.

bearbin:
I'll try and get back in the flow and review this soon, lots of familiarity with the codebase to build back up first :) Seems like lots of good work in here though!

MightyFilipns:
Finally, everything builds and all tests pass.
Jenkins fails because of apicheck command.
Clang crashes when trying to save the world for some reason but I can't reproduce it on my machine.

MightyFilipns:
Theres a glaring issue I kind of expected in this PR and thats the amount of auto-generated code.
Compilation times have balloned.
Clang and GCC both take over 10mins on jenkins. CheckBasicStyle takes 1min and 20 sec instead of just 20sec. Most of this code is in Protocol/Palette/, but theres some in Protocol/Procotol_1_XX.cpp files.
All of this code should be removed and the data loaded from files.

MightyFilipns:
Finally clang does not crash

NiLSPACE:
> CheckBasicStyle takes 1min and 20 sec instead of just 20sec.

If it's due to auto generated code that's increasing this I wouldn't be against ignoring those files. We currently do the same with the auto generated Lua bindings among others: https://github.com/cuberite/cuberite/blob/master/src/CheckBasicStyle.lua#L42

MightyFilipns:
Pretty sure it is. If you look in my other PR #5599 where the palette files do not have auto generated code CheckBasicStyle takes only about 30s
Also can the jenkins build job be force restarted? GCC timed out, and the log is somehow over 2 GiB so I can't open it.

NiLSPACE:
@bearbin did you have access to the jenkins jobs?

MightyFilipns:
Here's hoping this gets merged in 2025.

bearbin:
Most odd, I don't know why Jenkins messed up. Seems to be working now though, if it happens again I'll investigate some more.

We could probably improve the CheckBasicStyle speed by only processing files modified by a PR, but ultimately I think we'll probably get rid of it, a slightly longer build time is fine in the meantime.

I am minded to merge this into a feature branch in the cuberite/cuberite repo (not master) and do some more testing and a bit of refactoring there - probably distribute a beta build as well before it gets into master. But it's generally really good work and once it does get in it'll be a big change for the better. @NiLSPACE @madmaxoft @12xx12 any objections to merging into a feature branch?

madmaxoft:
On one hand, a feature branch is safe; on the other, even master is not much used anyway, so I'm not leaning towards either.

MightyFilipns:
A feature branch is probably better because this PR breaks some things (e.x block states aren't saved at all) and needs overall cleanup.

NiLSPACE:
I wouldn't mind putting it in a feature branch first 👍

madmaxoft:
A feature branch has pretty much no difference to a PR - it is still a branch which is NOT used at all. Seeing how this project is almost dead, no-one's gonna bother looking at a feature branch :D

src/WorldStorage/FastNBT.cpp
```diff
@@ -478,6 +497,22 @@ cFastNBTWriter::cFastNBTWriter(const AString & a_RootTagName) :
 
 
 
+cFastNBTWriter::cFastNBTWriter(bool Network1_21) :
```
MightyFilipns, line 500: This constructor is supposed to be temporary and needs to go. It prevents the other one from being used because some compilers, for some reason, implicitly convert strings to bool instead of using the other constructor. This caused the SchematicFileSerializer test to fail previously. Also, cParsedNBT needs to support this new format. No important C2S packets use it so I have not implemented it.

MightyFilipns:
So, are we going to merge this into a feature branch for further testing or go straight into master and then fix bugs along the way?

MightyFilipns:
1.8 - 1.12 should work but I haven't tested. I tried to avoid breaking them where I could. However, some constants have changed through the versions for which we don't have a way for per protocol mapping. E.g. sound names for note blocks (no mapping until 1.19), values for SoundParticleEffect packet, etc.

There are many breaking changes for plugins, mainly related to the Block Handling overhaul. Lua bindings are in a messy state. There are still some legacy functions that don't work anymore and I'm not sure if all the new functions for dealing with blocks are properly exported to Lua.
I just finished fixing the Lua docs, so the tests pass, but they need more work.

The old (1.12 and older) worlds are no longer supported and won't be loaded. 1.13+ worlds can be loaded; however, there have been many small changes in the file format throughout the versions, so some might fail to load. Worlds are always saved in 1.21.4 format.
Generally, the world reader/writer needs more testing.

MightyFilipns:
I have rebased this PR onto master so there aren't merge commits

mathiascode:
Thanks for the update. Since all our builds are rolling, I would prefer a separate branch for now, until we have a good enough migration path for old users.

mathiascode:
We can probably start dropping support for a bunch of old Minecraft versions as well in this branch. Trying to support all versions and their incompatibilities will just drive people mad.

MightyFilipns:
I suggest dropping support for 1.12 and below.
This would allow us to get rid of the old ID system completely, as keeping the new one compatible with the old one is kind of a nightmare.

12xx12:
> I am thinking of dropping support for 1.12 and below. This would allow us to get rid of the old ID system completely, as keeping the new one compatible with the old one is kind of a nightmare.

Yes please!!!

https://tenor.com/4yBa.gif

madmaxoft:
Since there is no other branch at this moment, just renaming this branch to whatever we want to call the new one is fine. Still, it might be good to keep cleaning up this PR before that - it has over 250 commits, many of them "fix ...", those should probably be squashed into whatever commit they are fixing. On the other hand, dropping the bomb and squashing this entire PR into a single commit is probably not a good idea either.

Your thoughts?

MightyFilipns:
I have cleaned up the git commit history quite a bit. I squashed over 100 commits.

MightyFilipns:
I have squashed all "state ..." commits

Q2297045667:
As a Chinese,I want to use a Chinese phrase to praise you.
Chinese：NB。

mathiascode:
> I am thinking of dropping support for 1.12 and below. This would allow us to get rid of the old ID system completely, as keeping the new one compatible with the old one is kind of a nightmare.

I would bump the minimum version even higher if it helps. I know people used to play versions like 1.8 for combat, but are there any reasons to play older versions after 1.12?

MightyFilipns:
I can't think of a reason to keep versions after 1.12. I doubt many people would play on those versions at all. So now the question becomes which version would become the new minimum. Do we go with a policy of supporting the latest version and a few before? The version that I'd go for is 1.18 which is because of world height changes.

If we try to support pre-1.18 we can't implement the world height changes.

Q2297045667:
> > I am thinking of dropping support for 1.12 and below. This would allow us to get rid of the old ID system completely, as keeping the new one compatible with the old one is kind of a nightmare.
> 
> I would bump the minimum version even higher if it helps. I know people used to play versions like 1.8 for combat, but are there any reasons to play older versions after 1.12?

As a non-developer,I think it's also necessary for me to point out that most players playing the Minecraft 1.8 version probably won't use third-party servers like Cuberite.I believe it's time to drop support for 1.12.Don't you think that the lower versions from 1.8 to 1.12 are a hindrance to the development of Cuberite?

Helcostr:
> I can't think of a reason to keep versions after 1.12. I doubt many people would play on those versions at all. So now the question becomes which version would become the new minimum. Do we go with a policy of supporting the latest version and a few before? The version that I'd go for is 1.18 which is because of world height changes.
> 
> If we try to support pre-1.18 we can't implement the world height changes.

If we have a post 1.18 world, we send world height changes, and block clients trying to enter (if we can send a protocol msg related to why they can't enter, that we can tell the client why they got blocked).

If we have a pre 1.18 world, all players can ender?

I think I'm of the opinion to default support new versions, with older versions getting support if old / unchanged mechanics exist, and not new breaking mechanics.

That being said... i wonder if we can trick a pre 1.18 client to run height changes by writing custom culling code that sends only the chunks they are interested in within the play area they can exist in (and teleporting the player to that necessary chunk).

MightyFilipns:
If the world is post-1.18. We should just block pre-1.18 players.
If the world is pre-1.18 all players can enter. In 1.18+ the world height is specified by the server, and the only limitations are that it must be 2048 or less and a multiple of 16.
As for the culling trick. It can probably be done but it will be most likely very buggy and not worth the effort necessary.

Helcostr:
> If the world is post-1.18. We should just block pre-1.18 players.

If the world height matches pre-1.18, why block them? (Assuming we dont have that fancy Wishlist culling trick that is probably never going to happen, but is a fun exercise).

MightyFilipns:
If the world height is 256 (pre-1.18 height) we can let all players join.

Q2297045667:
> If the world height is 256 (pre-1.18 height) we can let all players join.

Then Cuberite would have no appeal at all.If that's the case,why not just play version 1.15 directly?

Q2297045667:
> If the world height is 256 (pre-1.18 height) we can let all players join.

Many new biomes and terrains were added in the versions after 1.18.I think that if the world height before 1.18 is maintained,it will cause unnecessary trouble.

MightyFilipns:
World height would be user-configurable on a per-dimension basis. Only if it is 256 can a pre-1.18 player join.

Q2297045667:
> World height would be user-configurable on a per-dimension basis. Only if it is 256 can a pre-1.18 player join.

Very good

Helcostr:
> > If the world height is 256 (pre-1.18 height) we can let all players join.
> 
> Then Cuberite would have no appeal at all.If that's the case,why not just play version 1.15 directly?

I am not understanding this line of thinking:

If the config file that the server owner uses is set to 256, we can let the pre-1.18 in. If it's any other number than that,  pre-1.18 can't join.

The appeal of Cuberite was (at least how I discovered it) was "A lightweight, fast and extensible game server for Minecraft". By allowing lower versions to play (or not play), I do not believe it is ruining the core Cuberite experience. Especially if we figure out how to make the non-256 height avaible to pre 1.18: that would be crazy stupid fun.

Q2297045667:
> > > If the world height is 256 (pre-1.18 height) we can let all players join.
> > 
> > 
> > Then Cuberite would have no appeal at all.If that's the case,why not just play version 1.15 directly?
> 
> I am not understanding this line of thinking:
> 
> If the config file that the server owner uses is set to 256, we can let the pre-1.18 in. If it's any other number than that, pre-1.18 can't join.
> 
> The appeal of Cuberite was (at least how I discovered it) was "A lightweight, fast and extensible game server for Minecraft". By allowing lower versions to play (or not play), I do not believe it is ruining the core Cuberite experience. Especially if we figure out how to make the non-256 height avaible to pre 1.18: that would be crazy stupid fun.

Sorry,I'm just curious about this decision.At least for me personally,the appeal of Cuberite shouldn't be limited to high performance and scalability.Maybe we could add some features that vanilla doesn't have?Perhaps that would make it more interesting?

Helcostr:
> > > > If the world height is 256 (pre-1.18 height) we can let all players join.
> > > 
> > > 
> > > Then Cuberite would have no appeal at all.If that's the case,why not just play version 1.15 directly?
> > 
> > 
> > I am not understanding this line of thinking:
> > If the config file that the server owner uses is set to 256, we can let the pre-1.18 in. If it's any other number than that, pre-1.18 can't join.
> > The appeal of Cuberite was (at least how I discovered it) was "A lightweight, fast and extensible game server for Minecraft". By allowing lower versions to play (or not play), I do not believe it is ruining the core Cuberite experience. Especially if we figure out how to make the non-256 height avaible to pre 1.18: that would be crazy stupid fun.
> 
> Sorry,I'm just curious about this decision.At least for me personally,the appeal of Cuberite shouldn't be limited to high performance and scalability.Maybe we could add some features that vanilla doesn't have?Perhaps that would make it more interesting?

I think extra features that vanilla doesn't have is where mods/plugins come in. Cuberite about states the [following](https://cuberite.org/about/):

> Cuberite is designed with performance, configurability, and extensibility in mind, and also aims to accurately recreate most Minecraft features.

So if we as developers write bad code and cause the server to lag (relative to a Java server), but we get the features in, that would technically not be a good experience for the players, even though we got the shiny new thing in.

Tangent: right now, I'm personally working on EnderEye to make sure it works. Without EnderEye fully functioning, we don't have a way to access the end.

Q2297045667:
> > Sorry,I'm just curious about this decision.At least for me personally,the appeal of Cuberite shouldn't be limited to high performance and scalability.Maybe we could add some features that vanilla doesn't have?Perhaps that would make it more interesting?
> 
> I think extra features that vanilla doesn't have is where mods/plugins come in. Cuberite about states the [following](https://cuberite.org/about/):
> 
> > Cuberite is designed with performance, configurability, and extensibility in mind, and also aims to accurately recreate most Minecraft features.
> 
> So if we as developers write bad code and cause the server to lag (relative to a Java server), but we get the features in, that would technically not be a good experience for the players, even though we got the shiny new thing in.
> 
> Tangent: right now, I'm personally working on EnderEye to make sure it works. Without EnderEye fully functioning, we don't have a way to access the end.

Sorry,it seems that I might have overthought it.However,I'm curious about how you handle the addition of new version items.For example,if a player uses the 1.15 client to join the server,how will they be able to use the items added in 1.21.4?

Perhaps it's that after the server is set to a specific version,only clients of specific players are allowed to join?

Helcostr:
Couple methods:

The current favorable method of having a server in 1.21.4 mode is "block anyone who doesn't understand this item".

We could also display another item masquerading as the missing item, and maybe even supply a resource pack to bridge the gap (i am curious how far back this method can go).

Or we could just hide the item entirely / prevent it from being interacted with.

Those are realistically the only two options of allowing a lower version client interact with a higher version item.

Q2297045667:
> Couple methods:
> 
> The current favorable method of having a server in 1.21.4 mode is "block anyone who doesn't understand this item".
> 
> We could also display another item masquerading as the missing item, and maybe even supply a resource pack to bridge the gap (i am curious how far back this method can go).
> 
> Or we could just hide the item entirely / prevent it from being interacted with.
> 
> Those are realistically the only two options of allowing a lower version client interact with a higher version item.

Thank you for your reply.

MightyFilipns:
Has anyone been testing this branch?
If you want to test then do so on 1.21.3
I am currently looking for packet errors, pre-1.13 items and blocks not working as they should if they previously did and the Redstone simulator causing deadlocks. 
Anything related to entities (elytra won't close, minecarts not working, etc...). I don't plan on fixing these as the current system for handling entities is a mess. I will only fix bad entity metadata packets if they cause a disconnect on new clients. Old clients ignore bad metadata for the most part.
Also, known Bugs:
- Wrong chunks are sent. The client prints a warning in the log for this. We send chunks in a square around the player, while the new clients expect chunks to be sent in a circle.
- Sometimes, Cuberite won't complain about too short packets when parsing. This can happen when the last field from a packet is removed. The remaining fields are properly read; however, when it tries to read the removed field, it sees that there's no more data left and returns from the packet handling function. Usually, when packet reading fails some data is left and causes an error but not in this case. I figured this out because it happened with the use item packet.
- Door hinges not being determined properly.
- End portals sometimes won't complete
- Redstone wires sometimes won't get powered even tho they are directly connected to a wire. This is because somehow wires don't end in the WireStates array in the chunk struct so it is not updated. I don't know why this happens.
- Redstone is generally broken
- When loading a world from a file and placing a hopper under an already existing hopper causes a crash. This is because block entities are not loaded at all. There are probably similar bugs to do with block entities.
- Blocks are breaking too slowly in survival. This is because the new item component system isn't implemented and because block tags aren't implemented. This is the reason why many items don't work.
- Beacons don't work for the most part. Because effect IDs changed throughout versions.
- Player permissions aren't checked when executing commands. I enabled this temporarily for easier debugging.
- Players are replaced by other entities in 1.20.2+. Results from the problematic entity system.

mathiascode:
> I don't plan on fixing these as the current system for handling entities is a mess.

What are the current issues with the system, and do you have any ideas on how the system could be fixed? I think there were similar discussions before, but it's been years, and I've forgotten everything since...

MightyFilipns:
In pre-1.16 there were 6 different packets for spawning entities (entity, living entity, exp orb, player, painting and weather(lightning)). In 1.16 the weather one was merged with one of the other. In 1.19 entity, living entity and painting were merged into one. In 1.20.2 the player packet merged with the previous 3. And the [new snapshots](https://github.com/derklaro/mc-protocol/compare/1.21.4...25w02a#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R113) the exp orb was merged with the rest leaving only a single packet. Internally in cuberite entity handling was made around these these entity packet being separate. So when they were merged into one sending the new spawn packet proved to be a bit tricky. Espacially since there was no pre procotol mapping for player and painting entity ids.

The second issue is the way entity entity metadata is serilaziation is done. Currently to implement serialization for a new version the old function from the previous version must be entirely copy pasted even if only a single field on some entity has changed. This leads to a large amount of effectively duplicate code.

The third issue arose in 1.19 when a long field type was added and instead of it's id being the next unused one it was assigned id 2 meaning it pushed the ids of all other fields by one. This probably happened in other places but it hasn't caused any problems for now. This is probably the easiest to fix but it is annoying.

I am currently refactoring the entity handling system. I have merged all entity enums into and made it so the entity metadata writing function don't have to be copy pasted entirely. It's mostly finished, I am testing it currently. I am not sure whenever I push the commit to this PR or put it in another.

src/Protocol/Protocol_1_20.cpp
```diff
@@ -0,0 +1,2889 @@
...
+		Pkt.WriteBEUInt8(static_cast<UInt8>(eGameStateReason::InitialChunksComing));
+		Pkt.WriteBEFloat(0);  // Unused
+	}
+}
+
+
+
+
+
+void cProtocol_1_20_3::WriteEntityMetadata(cPacketizer & a_Pkt, const EntityMetadata a_Metadata, const EntityMetadataType a_FieldType) const
```
MightyFilipns, line 1707: The hack used for entity serialization

PR #5585, opened 2024-11-04
Use real block count for chunk data serialisation in 1.14

bearbin:
Extends #5560 to use the actual count of non-air entities in the chunk section.

Needs some testing to make sure this doesn't introduce bugs, and that the performance is acceptable.


src/Protocol/ChunkDataSerializer.cpp
```diff
@@ -502,7 +502,21 @@ inline void cChunkDataSerializer::Serialize477(const int a_ChunkX, const int a_C
...
+		Int16 BlockCount = 0;
+		if (Blocks != nullptr)
+		{
+			for (size_t i = 0; i != ChunkBlockData::SectionBlockCount; i++)
+			{
+				const BLOCKTYPE BlockType = (*Blocks)[i];
+				if (BlockType != E_BLOCK_AIR)
+				{
+					BlockCount++;
+				}
```
12xx12, line 515: Brute forcing this hurts a bit. But I don't have any other ideas on how to do this faster

12xx12 (approved):

src/Protocol/ChunkDataSerializer.cpp
bearbin: Vectorise! Would be easy to do a vector comparison,mask and popcount but would add some more complexity to the code/builds. Or keep an extra account of the block count alongside the main data (definitely sounds like a source of bugs to me!).

Honestly this naive version probably wouldn't actually have much effect on perf anyway, this block should be memory-throughput bound streaming the data into the packet anyway and an extra iteration would probably get lost in the noise.

MightyFilipns:
As this field exists in all 1.14+ versions we should have this counter as a separate function so we don't have duplicate code for all the different versions.

src/Protocol/ChunkDataSerializer.cpp
12xx12: I see, this is fine

PR #5586, opened 2024-11-04
Remove simple template-id from constructors for Vector3 template.

bearbin:
This was removed from the C++ standard in C++20; annoying warnings are produced by modern C++ compilers.

https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#2237


12xx12:
Why is the change only delivered as a diff o.O

bearbin:
Eh, it's legible... I guess I could just have linked the relevant section of the new standard but this one had the right amount of context in my head!

12xx12:
It does build. So we're good to go?

bearbin:
I'd say so - the standard change explicitly doesn't change functionality (the template-id was useless where we were using it) this PR just needed a look over to make sure I wasn't making some silly mistake.

12xx12 (approved):

PR #5589, opened 2024-11-10
Manually specify UTF-8 encoding for Cuberite.rc

### LINKED ISSUE #5588: The Cuberite.rc file is encoded incorrectly
Compiling Cuberite in vs2022 gives me an RC2104 error, which is suspected to be caused by the incorrect encoding of the rc file, I tried to change the file encoding from utf8 to gbk and successfully compiled Cuberite

#5072

bearbin:
Fixes #5588

I'm not a Windows user so can't test this, but this blog post suggests this is the correct solution: https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569

@66hh can you test that this compiles for your environment?

### DIFF
--- a/src/Resources/Cuberite.rc
+++ b/src/Resources/Cuberite.rc
@@ -1,3 +1,5 @@
+#pragma code_page(65001) // UTF-8
+
 #include "winres.h"
 
 Favicon ICON "icon.ico"


66hh:
> Fixes #5588
> 
> I'm not a Windows user so can't test this, but this blog post suggests this is the correct solution: https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569
> 
> @66hh can you test that this compiles for your environment?

Thanks, this change works in my environment

PR #5596, opened 2024-12-09
Update wiki links

MightyFilipns:
Recently wiki.vg was shut down.
So, the wiki.vg was merged into the Minecraft wiki
I replaced all wiki.vg links with the new ones however, some links pointed to pages that no longer exist


src/Mobs/Monster.h
```diff
@@ -42,7 +42,7 @@ class cMonster:
 
 	/** Creates the mob object.
 	If a_ConfigName is not empty, the configuration is loaded using GetMonsterConfig()
-	a_MobType is the type of the mob (also used in the protocol ( http://wiki.vg/Entities#Mobs 2012_12_22))
+	a_MobType is the type of the mob (also used in the protocol ( http://wiki.vg/Entities#Mobs 2012_12_22)) Page no longer exists 09.12.2024
```
Janmm14, line 45: As per https://web.archive.org/web/20241130182632/http://wiki.vg/Entities this redirects to Entity metadata: https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Entity_metadata

Due to changes in the protocol in more recent versions, mobs/monsters are no longer spawned different from other entities, so the current version of that page does not have a "Mobs" section anymore.
I'd say: Either link to some old version of Entity_metdata page (edit history is still available!) or find suitable other page.

src/Protocol/Authenticator.h
```diff
@@ -3,7 +3,7 @@
 
 // Interfaces to the cAuthenticator class representing the thread that authenticates users against the official Mojang servers
 // Authentication prevents "hackers" from joining with an arbitrary username (possibly impersonating the server admins)
-// For more info, see http://wiki.vg/Session
+// For more info, see http://wiki.vg/Session -- page no longer exists
```
Janmm14, line 6: similar to my first comment, the sessino page redirects to `Legacy_Mojang_Authentication`, but I think a link to `Protocol_Encryption#Authentication` mgiht be more suitable here.

src/Protocol/Protocol_1_8.cpp
```diff
@@ -2576,7 +2576,7 @@ void cProtocol_1_8_0::HandlePacketPlayerPosLook(cByteBuffer & a_ByteBuffer)
 
 void cProtocol_1_8_0::HandlePacketPluginMessage(cByteBuffer & a_ByteBuffer)
 {
-	// https://wiki.vg/index.php?title=Plugin_channels&oldid=14089#MC.7CAdvCmd
+	// https://wiki.vg/index.php?title=Plugin_channels&oldid=14089#MC.7CAdvCmd -- page probably no longer exists
```
Janmm14, line 2579: https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge/Plugin_channels?oldid=2768100 matched via web archive + looking up date of modify 2 July 2018 in the new wiki edit history.

MightyFilipns: I couldn't find this one on the new wiki

Janmm14: yeah if you look for it with the old oldid, it'll show you another page

bearbin:
Linking to pages in the Projects namespace in the minecraft wiki makes no sense, they will be deleted or moved back into the main namespace in due course, I would expect. 

The best thing to do is to link the current main namespace wiki pages if they exist, and if they don't then link Wayback Machine copies of the old page.

MightyFilipns:
Very few have been moved to the main namespace, and none of those are mentioned in this PR, as far as I know.

src/ChatColor.h
```diff
@@ -32,7 +32,7 @@ class cChatColor
 	static const char * White;
 
 	// Styles
-	// source: http://wiki.vg/Chat
+	// source: https://minecraft.wiki/w/Text_format
```
bearbin, line 35: don't link to a disambiguation page
```suggestion
	// source: https://minecraft.wiki/w/Formatting_codes
```

bearbin:
Yep, it's still early days for the migration project, so best to use the wayback machine instead (or just put a hold on this PR and wait for the pages to be moved across).

PR #5603, opened 2025-01-18
Add bug issue form

### LINKED ISSUE #5427: Update Issue Template
I saw on some othere repos that there is a way to create an actual form for bug reports and link some places.

see https://github.com/home-assistant/core/issues/new/choose

MightyFilipns:
Add the bug issue form
Identical to the existing issue template
Issue forms create an interactive web interface instead of just a big markdown input box
Fixes #5427


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,80 @@
...
+  - type: dropdown
+    id: OS
+    attributes:
+      label: On which OS does the issue appear?
+      multiple: true
+      options:
+        - Windows
+        - Linux
+        - macOS
+        - BSD
```
12xx12, line 72: Can we add "other" to open up a text box?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,80 @@
...
+        - 1.12.1
+        - 1.12
+        - 1.11.1 (1.11.2)
+        - 1.11
+        - 1.10.X
+        - 1.9.4 (1.9.3)
+        - 1.9.2
+        - 1.9.1
+        - 1.9
+        - 1.8.X
```
12xx12, line 59: Can we unify the version layout?

either `1.12.x` or `1.12`, `1.12.1`, `1.12.2` and so on

Or is there a reason to do it this way?

12xx12 (changes requested):
Great idea!

I have two notes and a question: Is there a way to preview this?

MightyFilipns:
To preview click View file on the bug_report.yml

.github/ISSUE_TEMPLATE/bug_report.yml
MightyFilipns: It looks like its not possible

.github/ISSUE_TEMPLATE/bug_report.yml
MightyFilipns: The reason for 1.10.X and 1.8.X is that all minor versions share the same protocol version. Meaning that a bug that affects one should affect all of them.

.github/ISSUE_TEMPLATE/bug_report.yml
12xx12: hm. any ideas on how to circumvent this?

Some people try to run cuberite on weird OS. like their routers xD

.github/ISSUE_TEMPLATE/bug_report.yml
12xx12: I see.

maybe add `(ProtocolVersion xxx) ` at the end so people in the future know the reason

MightyFilipns: I have added protocol version numbers to avoid future confusion

.github/ISSUE_TEMPLATE/bug_report.yml
MightyFilipns: The best we can do is to add an optional input box with a label that says to only fulfil if Other is chosen as the OS option

PR #5604, opened 2025-01-18
Update submodules

MightyFilipns:
Updates submodules
It looks like there are breaking changes in fmt
Most logging statements produce a compile error but I'm unsure how to fix them.
Does not build
Initially, I wanted to update all submodules in an attempt to compile with C++20, but it looks like some of the errors are still present. However, submodules should be updated regardless.


MightyFilipns:
Superseded by #5599
Closing

PR #5606, opened 2025-01-21
Sound Enums

Helcostr:
# Summary
Begins a basic SoundEvent enum, to be grown off of the current list of sounds in the modern list.

# Description
- Added the VS2017-x64 folder to the .gitignore for easier git management (hopefully this doesn't brick any CDI pipeline)
- Create the SoundEvent enums to help consolidate modern sound names into old sound identifiers (strings in 1.8)

# Added
- `SoundEvent.h`
- `PSounds_1_8.cpp` (file for implementing sounds for `cProtocol_1_8_0`)

### DIFF
--- a/src/Registries/SoundEvent.h
+++ b/src/Registries/SoundEvent.h
@@ -1,0 +1,8 @@
+# pragma once
+enum class SoundEvent
+{
+	// Ender Eye
+	EnderEyeCustomSurvive,
+    EnderEyeDeath,
+	EnderEyeLaunch,
+};
--- a/src/World.h
+++ b/src/World.h
@@ -192,6 +192,7 @@ public:
 	virtual void BroadcastScoreUpdate                (const AString & a_Objective, const AString & a_Player, cObjective::Score a_Score, Byte a_Mode) override;
 	virtual void BroadcastDisplayObjective           (const AString & a_Objective, cScoreboard::eDisplaySlot a_Display) override;
 	virtual void BroadcastSoundEffect                (const AString & a_SoundName, Vector3d a_Position, float a_Volume, float a_Pitch, const cClientHandle * a_Exclude = nullptr) override;  // Exported in ManualBindings_World.cpp
+	virtual void BroadcastSoundEffect                (const SoundEvent a_SoundEvent, Vector3d a_Position, float a_Volume, float a_Pitch, const cClientHandle * a_Exclude = nullptr) override;  // Exported in ManualBindings_World.cpp 
 	virtual void BroadcastSoundParticleEffect        (const EffectID a_EffectID, Vector3i a_SrcPos, int a_Data, const cClientHandle * a_Exclude = nullptr) override;  // Exported in ManualBindings_World.cpp
 	virtual void BroadcastSpawnEntity                (cEntity & a_Entity, const cClientHandle * a_Exclude = nullptr) override;
 	virtual void BroadcastThunderbolt                (Vector3i a_BlockPos, const cClientHandle * a_Exclude = nullptr) override;
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,9 @@ src/AllFiles.lst
 *.idb
 .vs/
 
+# win build
+VS2017-x64/*
+
 # cmake output folders and files
 ZERO_CHECK.dir/
 Debug/
--- a/src/Blocks/BroadcastInterface.h
+++ b/src/Blocks/BroadcastInterface.h
@@ -3,6 +3,7 @@
 
 #include "../Defines.h"
 #include "../Scoreboard.h"
+#include "../Registries/SoundEvent.h"
 
 // fwd:
 class cClientHandle;
@@ -58,6 +59,7 @@ public:
 	virtual void BroadcastScoreUpdate                (const AString & a_Objective, const AString & a_PlayerName, cObjective::Score a_Score, Byte a_Mode) = 0;
 	virtual void BroadcastDisplayObjective           (const AString & a_Objective, cScoreboard::eDisplaySlot a_Display) = 0;
 	virtual void BroadcastSoundEffect                (const AString & a_SoundName, Vector3d a_Position, float a_Volume, float a_Pitch, const cClientHandle * a_Exclude = nullptr) = 0;
+	virtual void BroadcastSoundEffect                (const SoundEvent a_SoundEvent, Vector3d a_Position, float a_Volume, float a_Pitch, const cClientHandle * a_Exclude = nullptr) = 0;
 	virtual void BroadcastSoundParticleEffect        (const EffectID a_EffectID, Vector3i a_SrcPos, int a_Data, const cClientHandle * a_Exclude = nullptr) = 0;
 	virtual void BroadcastSpawnEntity                (cEntity & a_Entity, const cClientHandle * a_Exclude = nullptr) = 0;
 	virtual void BroadcastThunderbolt                (Vector3i a_BlockPos, const cClientHandle * a_Exclude = nullptr) = 0;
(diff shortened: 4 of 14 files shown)


src/Registries/SoundEvent.h
MightyFilipns, line 1: If you want to auto-generate this file you can use the function ListSoundsEnum in Misc.cs in my [CuberiteScripts](https://github.com/MightyFilipns/CuberiteScripts/tree/main) repo. You may need to modify the script to run on your machine.

Helcostr: I do like leaving it manually generated, so we can slowly identify which sounds are missing, and fill them in. What do you think?

MightyFilipns: We can do it like that but it will be a lot of work considering how many sounds are in the game.

Helcostr: Well, this was my work process:

1. eye of ender identified with 2 sounds (a 3rd custom sound maybe)
1. generate enums by hand based on the name found in the wiki (or the export from the server json... whichever)
1. plug in the string related to the enum (to be used in the protocol)
1. use the enum in the spot i need the sound to emanate

idk if it's too much work, but it does provide careful steps.

MightyFilipns:
Doing a quick string search through the repo using the `SendSoundEffect("` as the search string reveals that it is only called twice using a hardcoded string. All other references are some kind of handling code.
Doing the same with `BroadcastSoundEffect("` reveals about 40 references. You should replace all of these with the new functions.

Helcostr:
> Doing a quick string search through the repo using the `SendSoundEffect("` as the search string reveals that it is only called twice using a hardcoded string. All other references are some kind of handling code. Doing the same with `BroadcastSoundEffect("` reveals about 40 references. You should replace all of these with the new functions.

Converting to Draft to implement the 40 ish references. WIP.

Helcostr:
1.8.9 Checklist (Verify sounds are intact):
- [ ] Blaze
- [ ] Horse
- [ ] Ghast
- [ ] ItemHoe
- [ ] Creeper
- [ ] BlockChest
- [ ] SplashPotion
- [ ] Firecharge
- [ ] Painting
- [ ] LeashKnot
- [ ] ChorusFruit
- [ ] ZombieVillager
- [ ] ArrowEntity
- [ ] MinecartChest
- [ ] ExperienceOrb
- [ ] Sheep
- [ ] Pickup
- [ ] Lever
- [ ] Dispenser
- [ ] Piston
- [ ] WoodenBtn
- [ ] Fishing bobber
- [ ] Thunder
- [ ] Tnt primed (+minecart)
- [ ] BrewingStand
- [ ] ArrowShoot

Helcostr:
Note to self: it's not as easy as replacing all of `BroadcastSoundEffect("`.

```
cBlaze::cBlaze(void) :
	Super("Blaze", mtBlaze, "entity.blaze.hurt", "entity.blaze.death", "entity.blaze.ambient", 0.6f, 1.8f),
	m_IsCharging(false),
	m_ChargeTimer(0)
{
	SetGravity(-8.0f);
	SetAirDrag(0.05f);
}
```

Helcostr:
@MightyFilipns i am waiting for a buddy to help me test 1.8 reception of non 1.8 sounds. but how are u liking the structure of everything? (ignore the non-alpha order of anything, since it's pretty much first come first ordered... ish)

src/World.h
```diff
@@ -192,6 +192,7 @@ class cWorld  // tolua_export
 	virtual void BroadcastScoreUpdate                (const AString & a_Objective, const AString & a_Player, cObjective::Score a_Score, Byte a_Mode) override;
 	virtual void BroadcastDisplayObjective           (const AString & a_Objective, cScoreboard::eDisplaySlot a_Display) override;
 	virtual void BroadcastSoundEffect                (const AString & a_SoundName, Vector3d a_Position, float a_Volume, float a_Pitch, const cClientHandle * a_Exclude = nullptr) override;  // Exported in ManualBindings_World.cpp
+	virtual void BroadcastSoundEffect                (const eSoundEvent a_SoundEvent, Vector3d a_Position, float a_Volume, float a_Pitch, const cClientHandle * a_Exclude = nullptr) override;  // Exported in ManualBindings_World.cpp
```
MightyFilipns, line 195: You should delete the comment here as you haven't exported it.

Helcostr: what is exporting? i just copied the line

MightyFilipns:
Overall good
You have only implemented the new SendSoundEffect 1.8 but you also need to do the same for 1.9. After that, the packet is unchanged until 1.19.
Also by the .gitignore changes, I assume you are using VS2017. May I ask why using such an old IDE?
Also, you should export the new enum and functions to Lua.

src/World.h
MightyFilipns: If you export the function it is will be accessible by lua plugins

Helcostr:
> Overall good
> You have only implemented the new SendSoundEffect 1.8 but you also need to do the same for 1.9. After that, the packet is unchanged until 1.19.
> Also by the .gitignore changes, I assume you are using VS2017. May I ask why using such an old IDE?
> Also, you should export the new enum and functions to Lua.

Thanks. I am not using VS2017, but rather the tutorial for getting started told me to make the folder named VS2017.

Also, i am not sure how to export the new enum/functions to lua. What library are we using so i can study the documentation?

MightyFilipns:
Well, I am not too well-versed in how the Lua plugin system works. But it appears to be that we are using tolua++ as the library which was archived 4.5 years ago... #4281 is a PR to replace the library partially. Also I found this which should help:
https://xoft.cz/cuberite/docs/ExportingAPI.html

Helcostr:
exporting seems complicated: i will unmark the comments that i "did" export, and leave it as an issue someone else can tackle later down the line (maybe me if i wrap my head around it)

the only thing i will probably cover is the 1.9 cover of the existing 1.8 sounds.

what is the final version that i will have to test btw? how do i know what's our current max coverage?

MightyFilipns:
Yeah, I also find that dealing with the Lua system is quite difficult. You should make sure it works in 1.12 and 1.14 I guess.

Helcostr:
ok, i will check 1.9, 1.10, 1.11, 1.12, 1.13, 1.14

MightyFilipns:
@bearbin Can you look into why Jenkins is falling?

aidanvalero:
Hey @Helcostr! This approach looks solid - having proper enums for sounds instead of magic strings is way better. I just fixed the 1.8 pickup sound bug (#5628) by doing translation in `SendSoundEffect`, so I'm familiar with this pain.

A few thoughts after reading through the comments:

**On the Lua export complexity** - Yeah, tolua++ is rough. If you want to punt on that for now and let someone else tackle it, that's totally reasonable. Getting the C++ side working first is more important than Lua bindings IMO.

**Version testing question** - MightyFilipns mentioned testing 1.12 and 1.14. From what I've seen working on protocol bugs, the packet format for sounds changes at:
- 1.8 (string-based, needs translation from modern names)
- 1.9-1.12 (still string-based but slightly different packet structure)
- 1.13+ (sound IDs instead of strings)
- 1.19+ (another packet format change)

So yeah, 1.9, 1.12, 1.13, and 1.14 coverage would catch the main transitions.

**Quick note on your approach** - The `PSounds_*.cpp` files for each protocol version make sense, but have you considered just having the enum-to-string mapping be version-aware instead of separate files? Something like:

```cpp
AString SoundEventToString(SoundEvent event, int protocolVersion) {
    if (protocolVersion < 340) { // 1.12 and below
        // return old format
    }
    // return modern format
}
```

Might be cleaner than maintaining 4 separate translation files? Just a thought.

Also saw you're hitting the Jenkins failure - that's been broken for months, ignore it. GitHub Actions is what actually matters now. There's an open issue (#5634) to remove Jenkins entirely.

Overall this is good work and will make sound handling way more maintainable. Let me know if you want help testing any of the protocol versions - I've got a Windows setup and can spin up clients pretty easily.

PR #5611, opened 2025-07-29
Add GitHub actions

MightyFilipns:
- StyleCheck.yml:  Runs CheckBasicStyle.lua and CheckBindingsDependencies
- cmake-multi-platform:

  - Runs GCC and Clang for linux_x86_x64
  - Run MSVC for Windows
  - Server files are put into a ZIP file
  - Runs tests for all platforms
  - Checks docs on non-Windows builds
  - If everything succeeds, the ZIP file is uploaded as a GitHub release

The name of the release will be "Release - (branch name)"

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -133,7 +133,8 @@ jobs:
         name: Cuberite-${{ matrix.os }}-${{ matrix.c_compiler }}-${{ matrix.build_type }}
         path: Cuberite-${{ matrix.os }}-${{ matrix.c_compiler }}-${{ matrix.build_type }}.zip
   upload_release:
-    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/experimental'
+    # Disable the job until a proper release process is made
+    if: (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/experimental') && false
     permissions:
       contents: write
     needs: build
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -27,6 +27,9 @@ jobs:
       # 3. <Linux, Release, latest Clang compiler toolchain on the default runner image, default generator>
       #
       # To add more build types (Release, Debug, RelWithDebInfo, etc.) customize the build_type list.
+
+      # The excluded ones are needed so some configs don't run.
+      # The included ones are needed because the cpp_compiler field is not set otherwise, leading CMake to fail.
       matrix:
         os: [ubuntu-latest, windows-latest]
         build_type: [Release]
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -1,11 +1,20 @@
 # This starter workflow is for a CMake project running on multiple platforms. There is a different starter workflow if you just want a single platform.
 # See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-single-platform.yml
 name: Build
-
-on: [push, pull_request]
+on:
+  push:
+    tags-ignore:
+      - release-*
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
 
 jobs:
   build:
+    timeout-minutes: 60
+    permissions: {}
     runs-on: ${{ matrix.os }}
 
     strategy:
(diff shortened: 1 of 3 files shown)


madmaxoft (approved):

.github/workflows/StyleCheck.yml
```diff
@@ -0,0 +1,25 @@
+name: Style Check
```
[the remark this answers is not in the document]
MightyFilipns: Fixed

.github/workflows/cmake-multi-platform.yml
```diff
@@ -0,0 +1,128 @@
+# This starter workflow is for a CMake project running on multiple platforms. There is a different starter workflow if you just want a single platform.
+# See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-single-platform.yml
+name: CMake on multiple platforms
+
+on:
+  push:
+    branches: [ "master", "experimental" ]
+  pull_request:
+    branches: [ "master", "experimental" ]
```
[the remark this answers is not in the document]
MightyFilipns: A complete build takes about 25 minutes. I have set it so that builds are run for all branches.
We could make it so that on non master/experimental branches, tests and tools aren't built, and docs aren't checked.
This should cut down the time to about 8 minutes.

.github/workflows/cmake-multi-platform.yml
```diff
@@ -0,0 +1,128 @@
...
+            if [ -f ./apiCheckFailed.flag ]; then
+                echo "ERROR: API check has failed with an unknown error"
+                exit 1
+            fi
+    - name: Extract branch name
+      shell: bash
+      run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+      id: extract_branch
+
+    - name: Upload Binary
```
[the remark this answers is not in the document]
MightyFilipns: I have split the uploading into a separate job.

.github/workflows/Build.yml
```diff
@@ -0,0 +1,181 @@
...
+          - os: ubuntu-latest
+            c_compiler: clang
+            cpp_compiler: clang++
+        exclude:
+          - os: windows-latest
+            c_compiler: gcc
+          - os: windows-latest
+            c_compiler: clang
+          - os: ubuntu-latest
+            c_compiler: cl
```
bearbin, line 50: Why do we need both include and exclude, seems a bit duplicative but maybe that's just how GHA does things?

.github/workflows/Build.yml
```diff
@@ -0,0 +1,181 @@
...
+
+    - name: Set reusable strings
+      # Turn repeated input strings (such as the build output directory) into step outputs. These step outputs can be used throughout the workflow file.
+      id: strings
+      shell: bash
+      run: |
+        echo "build-output-dir=${{ github.workspace }}/build" >> "$GITHUB_OUTPUT"
+
+    - name: Setup Deps
+      run: git submodule update --init
```
bearbin, line 63: This should be done by the actions/checkout when the right flag is passed in there.

.github/workflows/Build.yml
```diff
@@ -0,0 +1,181 @@
...
+    steps:
+      - name: Extract branch name
+        shell: bash
+        run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
+        id: extract_branch
+      - name: Download Artifacts
+        uses: actions/download-artifact@v4
+        with:
+          merge-multiple: true
+      - if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/experimental'
```
bearbin, line 148: We should move this up earlier so we're not running the release step for other branches.

bearbin (changes requested):
Few minor changes, otherwise looks basically ready to merge.

If it was entirely up to me, I would hold off enabling the upload_release job for now and try to get a proper release process worked out; our current solution of not having one doesn't really work when github demands releases have tags/"releases" - even a nightly or incrementing version number would be good to nail down.

.github/workflows/Build.yml
MightyFilipns: Fixed

MightyFilipns:
The upload_release job can be disabled for the time being, while we figure out a release process.

.github/workflows/Build.yml
madmaxoft: This should be then documented, if the file format supports comments

bearbin:
Doesn't look like the release has actually been turned off?

MightyFilipns:
> Doesn't look like the release has actually been turned off?

Should be now

MightyFilipns:
It looks like GH actions have to be explicitly enabled by an admin in the repo settings.

PR #5627, opened 2026-01-23
Fix custom name JSON serialization crash in 1.13+ clients

### LINKED ISSUE #5577: Custom names crash 1.13+ clients

aidanvalero:
Fixed client crashes when using name tags on mobs in Minecraft 1.13+. Custom names were sent as plain strings instead of properly formatted JSON chat components, causing invalid JSON error crashes.

## Changes

- Protocol_1_13.cpp: Serialize custom names as JSON chat components
- Protocol_1_14.cpp: Serialize custom names as JSON chat components  
- Protocol_1_15.cpp: Serialize custom names as JSON chat components
- Protocol_1_19.cpp: Fix cProtocol_1_19_4 custom name serialization

All affected protocols now use JsonUtils::SerializeSingleValueJsonObject() to ensure proper JSON formatting with correct escaping of special characters (quotes, backslashes, newlines, etc.).

Fixes #5577

### DIFF
--- a/src/Protocol/Protocol_1_13.cpp
+++ b/src/Protocol/Protocol_1_13.cpp
@@ -1107,7 +1107,7 @@ void cProtocol_1_13::WriteEntityMetadata(cPacketizer & a_Pkt, const cEntity & a_
 			{
 				WriteEntityMetadata(a_Pkt, EntityMetadata::EntityCustomName, EntityMetadataType::OptChat);
 				a_Pkt.WriteBool(true);
-				a_Pkt.WriteString(a_Mob.GetCustomName());
+				a_Pkt.WriteString(JsonUtils::SerializeSingleValueJsonObject("text", a_Mob.GetCustomName()));
 
 				WriteEntityMetadata(a_Pkt, EntityMetadata::EntityNameVisible, EntityMetadataType::Boolean);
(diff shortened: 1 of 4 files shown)


aidanvalero:
Runtime-verified this end to end. Built this branch (gcc Release, Linux) plus its base commit without the fix, ran an offline-mode server, spawned a zombie with custom name PR5627Test via a small test plugin, and captured what a client actually receives in entity metadata key 2 (custom name, OptChat) using node-minecraft-protocol.

Without this PR, protocol 1.19.4 and 1.14.4 both receive the raw string:

    value = "PR5627Test"

With this PR, both receive a valid chat component:

    value = {"text":"PR5627Test"}

Vanilla 1.13+ clients JSON-parse that field, which is why the raw string breaks them. The test plugin and client script are ~40 lines total, can post them if useful.

aidanvalero:
Closing as a duplicate of #5641, which is the same fix on a current branch.

PR #5629, opened 2026-01-23
Fix bedrock mining exploit with packet spamming

### LINKED ISSUE #5528: Mining bedrock with packetmine drops bedrock.
Client version: 1.12.2
Server OS: Linux
Cuberite Commit id: 770d838f377b82e9fd8c490a83eee4506fedfedb

### Expected behavior
Using PacketMine on bedrock shouldn't drop bedrock.

### Actual behavior
Using PacketMine on bedrock allows bedrock to be dropped. You can also mine pretty much any block instantly with FastBreak on.

Use FastBreak on lambda with PacketMine enabled

### Server log
Server logs nothing about this.

aidanvalero:
Fixed bedrock mining exploit where modified clients (FastBreak/PacketMine) could instantly break unbreakable blocks by spamming packet finish packets to the server.

## Root Cause

- `HandleBlockDigFinished()` had no server-side validation for unbreakable blocks
- Server trusted the client's "finished digging" packet without checking if the block can actually be mined
- FastBreak clients spam finish packets to bypass the anti-cheat timer

## Security Impact

**Before:** Players could obtain bedrock, barriers, and command blocks using packet-spamming clients  
**After:** Server validates block hardness and rejects mining attempts on unbreakable blocks

## Changes

- `ClientHandle.cpp`: Added hardness check in `HandleBlockDigFinished()`
- Blocks with negative hardness (bedrock, barriers) are now rejected
- Exploit attempts trigger block resync and position correction

## Testing

- Test with vanilla client: Normal mining still works ✓
- Test with FastBreak: Bedrock mining blocked ✓
- Test creative mode: Still can't break bedrock ✓

Fixes #5528

### DIFF
--- a/src/ClientHandle.cpp
+++ b/src/ClientHandle.cpp
@@ -1337,6 +1337,16 @@ void cClientHandle::HandleBlockDigFinished(Vector3i a_BlockPos, eBlockFace a_Blo
 		return;
 	}
 
+	// FIX FOR #5528: Prevent mining unbreakable blocks (bedrock, barriers, etc.)
+	// Blocks with negative hardness cannot be broken even in creative mode
+	if (cBlockInfo::GetHardness(DugBlock) < 0.0f)
+	{
+		LOGD("Player %s attempted to mine unbreakable block at %s", m_Player->GetName().c_str(), a_BlockPos);
+		m_Player->SendBlocksAround(a_BlockPos, 2);
+		SendPlayerPosition();
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


aidanvalero:
Closing - the fix is wrong. It blocks bedrock breaking in creative mode too, which vanilla allows, and the LOGD call passes a Vector3i to a printf-style %s so it would not compile. If I redo this it will be gamemode-aware and actually tested first.

PR #5632, opened 2026-01-23
Fix API crash when passing invalid eBlockFace to AddFaceDirection

### LINKED ISSUE #5441: API: Calling function AddFaceDirection with invalid eBlockFace crashes the server

aidanvalero:
Fixed server crash when Lua plugins pass invalid block face values to `AddFaceDirection()`. Previously, calling this with `BLOCK_FACE_NONE` (-1) or other invalid values would crash the server.

## Root Cause

- `AddFaceDirection()` had an `ASSERT` on `BLOCK_FACE_NONE` case
- Invalid block face values triggered crash via `UNREACHABLE()` macro
- Lua API allows passing any integer, including invalid values
- `BLOCK_FACE_NONE` is actually a valid enum value but wasn't handled gracefully

## Security Impact

**Before:** Malicious plugins could crash the server with `AddFaceDirection(Vector3i(1,1,1), -1, false)`  
**After:** Invalid block faces return position unchanged with warning log

## Changes

- `Defines.cpp`: Removed `ASSERT` from `BLOCK_FACE_NONE` case
- `BLOCK_FACE_NONE` now returns position unchanged (logical behavior for "no face")
- Invalid block face values log warning and return original position
- Replaced `UNREACHABLE()` with graceful fallback

## Testing

- Valid block faces: Still work correctly
- BLOCK_FACE_NONE (-1): Returns position unchanged
- Other invalid values: Log warning, return position unchanged
- Lua API fuzzing: No longer triggers crash

Fixes #5441

### DIFF
--- a/src/DataComponents/DataComponents.h
+++ b/src/DataComponents/DataComponents.h
@@ -73,14 +73,25 @@ namespace DataComponents
 		UInt32 RepairCost;
 	};
 
+	struct MapIdComponent
+	{
+		MapIdComponent() = default;
+		MapIdComponent(UInt32 a_MapID)
+		{
+			MapID = a_MapID;
+		}
+		UInt32 MapID;
+	};
+
 	COMP_EQU_OP(MaxDamageComponent, lhs.MaxDamage == rhs.MaxDamage);
 	COMP_EQU_OP(DamageComponent, lhs.Damage == rhs.Damage);
 	COMP_EQU_OP(MaxStackSizeComponent, lhs.maxStackSize == rhs.maxStackSize);
 	COMP_EQU_OP(UnbreakableComponent, lhs.unbreakable == rhs.unbreakable);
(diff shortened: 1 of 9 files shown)


aidanvalero:
Closing this PR - it was created without the actual fix code. Will resubmit properly if needed.

PR #5637, opened 2026-02-19
Fix entity chunk loading, improve fire simulation, and add world config comments

krutoisigma23:
1. Entity & Chunk Loading Safety (Chunk.cpp, ChunkMap.cpp)

Changed ConstructChunk to GetChunk when an entity moves to or spawns in a non-existent chunk.

Reason: ConstructChunk creates an empty chunk structure immediately, whereas GetChunk ensures the chunk is properly queued for loading or generation. This prevents entities from getting stuck in empty chunks.

2. Fire Simulator Improvements (FireSimulator.cpp)

Implemented random jitter for fire burn step timing (addressing a TODO).

Reason: Prevents the lockstep effect where large areas of fire update exactly at the same time, making fire behavior look more natural.

3. World Configuration (World.cpp)

Added descriptive comments for sections in the world.ini file (General, SpawnPosition, Broadcasting, etc.).

Reason: Makes the configuration file more user-friendly for server admins.

4. Block Ticking Safety (World.cpp)

Added a check in TickQueuedBlocks to verify if the chunk is still valid/loaded before ticking a block.

Reason: Prevents crashes or undefined behavior when trying to process queued ticks (e.g., redstone) in chunks that have already been unloaded.

5. Sign Handling Optimization (World.cpp, World.h)

Added an r-value reference overload for cWorld::SetSignLines to avoid unnecessary string copies.

Added explicit checks (IsBlockWallSign / IsBlockSignPost) inside SetSignLines to ensure the target block entity is actually a sign before casting.

### DIFF
--- a/src/World.cpp
+++ b/src/World.cpp
@@ -255,16 +255,16 @@ cWorld::cWorld(
 			IniFile.AddKeyComment(a_SectionName, a_Comment);
 		};
 
-		AddSectionComment("General",      "General world settings: dimension, time, weather, tick age, daylight cycle, etc.");
-		AddSectionComment("SpawnPosition","Spawn coordinates and pregeneration settings.");
-		AddSectionComment("Broadcasting", "Controls whether deaths and achievements are broadcast to all players.");
-		AddSectionComment("LinkedWorlds", "Portal linkage settings between worlds. Leave a value blank to disable that portal type.");
-		AddSectionComment("Storage",      "Chunk storage settings (schema and compression).");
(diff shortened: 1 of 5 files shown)


src/World.cpp
```diff
@@ -2948,7 +3002,15 @@ void cWorld::TickQueuedBlocks(void)
 		if (Block->TicksToWait <= 0)
 		{
 			// TODO: Handle the case when the chunk is already unloaded
```
MightyFilipns: Remove the TODO comment

MightyFilipns:
LGTM merging

PR #5638, opened 2026-02-20
Implement backpressure for chunk saving, prevent NBT corruption, and add effective light calculation

krutoisigma23:
1. Stability & Memory Management

Bounded chunk data queue (World.cpp): Implemented backpressure for m_SetChunkDataQueue. If the queue reaches 512 chunks, the thread now sleeps (yields) to allow the tick thread to drain it.

Reason: Prevents unbounded memory growth (OOM crashes) during massive world generation, pregeneration, or heavy chunk loading events.

NBT string length safety (FastNBT.cpp): Added a check to ensure strings do not exceed the 16-bit unsigned integer limit (65535) before saving. Strings exceeding this limit are now safely truncated.

Reason: NBT format uses a 16-bit length prefix. Previously, writing a string longer than 65535 bytes would overflow the length field, resulting in silently corrupted NBT files (e.g., corrupted chunk data).

2. World Logic

Actual light calculation (World.cpp, World.h): Implemented GetBlockActualLight (resolving an old TODO).

Reason: Properly calculates the effective light level at a specific block position by combining block light and sky light, while correctly accounting for the current m_SkyDarkness (day/night cycle).

Note for reviewers: The math is straightforward and follows standard Minecraft mechanics (max(EffectiveSky, BlockLight)), but I haven't done extensive in game behavioral testing for this specific function yet. Please double check the logic!

3. Quality of Life 

Anti-spam for failed chunk saves (WSSAnvil.cpp): Added a static set ReportedSerializationFailures to keep track of chunks that fail to serialize.

Reason: If a chunk gets corrupted in memory and cannot be saved, the server previously spammed the console with errors every single save interval. Now, it only logs the failure once per chunk per server run.

Better blockItem converter logging (BlockItemConverter.cpp): Replaced integer casting with NamespaceSerializer::From(a_Block) in the error log.

Reason: Makes debugging much easier by printing the actual block string name instead of a magic number when an unintended block conversion occurs.

### DIFF
--- a/src/Registries/BlockItemConverter.cpp
+++ b/src/Registries/BlockItemConverter.cpp
@@ -1,6 +1,7 @@
 
 #include "Globals.h"
 #include "BlockItemConverter.h"
+#include "BlockType.h"
 
 namespace BlockItemConverter
 {
@@ -963,7 +964,7 @@ namespace BlockItemConverter
 			case Item::ZombieHead:                         return BlockType::ZombieHead;
 			default:
 			{
-				LOGERROR("{}: Got unintended item to convert to block: {}", __FUNCTION__, static_cast<UInt32>(a_Item));
+				LOGERROR("{}: Got unintended item to convert to block: {}", __FUNCTION__, ItemTypeToString(a_Item));
 				UNREACHABLE("Converting unintended item to block");
 			}
 		}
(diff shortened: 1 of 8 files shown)


src/Registries/BlockItemConverter.cpp
```diff
@@ -963,7 +963,7 @@ namespace BlockItemConverter
 			case Item::ZombieHead:                         return BlockType::ZombieHead;
 			default:
 			{
-				LOGERROR("{}: Got unintended item to convert to block: {}", __FUNCTION__, static_cast<UInt32>(a_Item));  // TODO: print string instead of int
```
MightyFilipns, line 966: You just removed a TODO?

MightyFilipns (changes requested):

MightyFilipns:
CI/CD fails with a linking error

MightyFilipns (approved):

PR #5643, opened 2026-08-05
Migrate builds to GitHub Actions

mathiascode:
Builds Cuberite for Linux, Android, Windows and macOS, and creates a release containing the artifacts on every commit to master. This is intended to replace AppVeyor and the Jenkins build server.

x86 (32-bit) and armhf builds are not included, since I'm not sure how many people still use those architectures, and I didn't want to invest time in them before we know. Caching is also not set up right now, and the Windows/Android builds use older OS versions for now due to some build issues.

An example of how the release looks: https://github.com/mathiascode/cuberite/releases/tag/cuberite-latest

@bearbin @MightyFilipns


bearbin:
General opinion - let's do it. Will review properly later today.

mathiascode:
Thanks! There's some room for improvement, e.g. parallel Android builds and changing the artifact names in the scripts instead of renaming them in the build steps. I didn't want to touch the scripts more than necessary yet, though.

A bunch of build warnings that should ideally be fixed too, if someone wants to dive into the code and dependencies again at some point.

.github/workflows/build.yml
```diff
@@ -0,0 +1,256 @@
...
+      CUBERITE_BUILD_COMMIT_ID: ${{ github.sha }}
+      CUBERITE_BUILD_DATETIME: ${{ github.event.repository.updated_at }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: 'recursive'
+
+      - name: Build
+        run: |
+          cmake -B build -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DSELF_TEST=Yes -DBUILD_TOOLS=Yes
```
MightyFilipns, line 98: Disable tests as some are broken on windows, they aren't being run anyways. We could disable tools as well.

MightyFilipns (changes requested):

MightyFilipns:
While we are at it we could add RISC-V builds as well?

mathiascode:
> While we are at it we could add RISC-V builds as well?

I'd have to look into what options there are, since GH actions doesn't support RISC-V natively. Might be possible with QEMU.

mathiascode:
I learned some new tricks while doing CI work in other projects recently, so I'll add some small cleanups to this PR soon.

bearbin (approved):
Going to give my approval to this pending the change to the windows tests and any enhancements that you wanted to make.

